Issue #928, opened 2015-10-08
ACME Test fails with python3.4 - jose/interfaces_test.py", line 94, in test_json_dumps_pretty


mrueg:
https://github.com/letsencrypt/letsencrypt/commit/3dac62f20e8dbfacea32df6b1748d5f560b559e3
```
...............F.....................................................................................................................................................................................................................................................................................................................................
======================================================================
FAIL: test_json_dumps_pretty (acme.jose.interfaces_test.JSONDeSerializableTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/lib/acme/jose/interfaces_test.py", line 94, in test_json_dumps_pretty
    self.seq.json_dumps_pretty(), '[\n    "foo1", \n    "foo2"\n]')
AssertionError: '[\n    "foo1",\n    "foo2"\n]' != '[\n    "foo1", \n    "foo2"\n]'
  [
-     "foo1",
+     "foo1", 
?            +
      "foo2"
  ]

----------------------------------------------------------------------
Ran 341 tests in 2.725s

FAILED (failures=1)
```

It works with python 2.7

kuba:
I hate that the tests for Python 3 have been disabled :( Sorry about that issue!

PR #957, opened 2015-10-12
Python 2.6 and 3.3+ support

kuba:
I really don't think 13 line change should stop us from supporting Python 2.6 throughout and 3.3+ in `acme`. Note that those have been disabled quite long time ago and the code is almost 100% compliant. Previous efforts to make code base portable were just too much work and they shouldn't have been discarded that easily.

It's clear that people expect `acme` to work with Python 3 (c.f. #928) and it's a real pain that Travis doesn't check such silly small things. Apparently current setup was meant to make things easier for me, but it's the opposite. Turning Py3 switch in `tox` shouldn't affect anyone else, apart from me.

Travis builds are parallel, so it does not increase the total build time, contrary to https://github.com/letsencrypt/letsencrypt/pull/630#issuecomment-140599598.

If it really slows down the development of anyone in the team, just ignore Travis Py 2.6 build failure, and I'll fix stuff myself, but at least let's keep a track of the compatibility status.

This PR will also allow me to release py3 wheels to PyPI.


kuba:
Coincidentally, py2.6 job was affected by some Travis scheduling bug:
```
There was an error with the connection to the VM.
Your job will be requeued shortly...............
```

@bmw, could you kick https://travis-ci.org/letsencrypt/letsencrypt/jobs/84869803?

bmw:
kicked

kuba:
@bmw, thank you very much :)

Travis green.

kuba:
Hah, see also #962.

pde:
This will be merged at GA time, but we're going to leave it out until then to save developers time fixing python2isms and waiting longer for travis.

kuba:
Ready to merge.

bmw:
LGTM

PR #963, opened 2015-10-13
Added pip and virtualenv installation steps to Mac's bootstrap script.

douglasjose:


bootstrap/mac.sh
```diff
@@ -6,3 +6,13 @@ fi
 
 brew install augeas
 brew install dialog
+
+if ! hash pip 2>/dev/null; then
+    echo "pip Not Installed\nInstalling python from Homebrew..."
+    brew install python
```
kuba: Doesn't `virtualenv` install `pip` automatically? I.e. is the `brew install python` necessary? We don't need global `pip`, we just care about `./venv/bin/pip`. Actually, I'd prefer not to have global `pip` as this might end up in users typing `sudo pip`.

kuba:
@BKreisel, is that PR even necessary?

bootstrap/mac.sh
BKreisel: `virtualenv` will generate its own localized versions of `pip`. I solely used pip with virtualenv currently.

BKreisel:
@kuba. I'm not sure it is. I currently use the python 2.7.10 which ships with OS X and personally wouldn't want another homebrew copy of 2.7 installed from homebrew.

bmw:
Can confirm installing a 2nd copy of Python isn't necessary ([link](https://docs.python.org/2/using/mac.html#getting-and-installing-macpython)). @BKreisel, what about `virtualenv`?

BKreisel:
@bmw. virtualenv doesn't come pre-installed as far as I know.

bmw:
:frowning: So maybe we do need something like this as `virtualenv` must be installed in order for the client to work.

Not sure about the best way to go about this. Created #968 to track the issue.

douglasjose:
Hi all,

Sorry about the lack of context for this pull request. My objective is to cover the requirements I had to install manually after running `bootstrap/mac.sh`, according to the [prerequisite](https://letsencrypt.readthedocs.org/en/latest/using.html#prerequisites) instructions.

My approach was to install python using Homebrew in order to get a `pip` installation, which I could then use to get `virtualenv`. I did not had to `sudo` during any time of the process.

I'd be happy to rework my changes if you believe there is a better way to install those dependencies.

Thank you,

Douglas

bmw:
Thank you for the PR, @douglasjose. Unfortunately, none of the core developers working on the client use OS X so it's support was less than optimal. Recently, @BKreisel went through and cleaned up most of the issues so it's currently in a much better state.

As you can see by me opening a new issue in response to this PR, I actually was unaware that `virtualenv` wasn't being installed in `mac.sh`. I'm unsure of the best way to go about doing this, but I think something along these lines should be added to the script. I'm personally not a fan of installing a second copy of Python and `pip`, but if no one has any better alternatives, I think this is good to merge as this approach appears to be the most common.

kuba:
Can we not just install http://brewformulas.org/PyenvVirtualenv and use it's `virtualenv` to proceed with th installation?

douglasjose:
@kuba , I tried Homebrew's `pyenv-virtualenv` formula instead of `pip` + `virtualenv`, and here's what I observed:

* Once the `pyenv` and `pyenv-virtualenv` formulas are installed, they instruct the user to change their shell init scripts to [add initialization commands](https://github.com/yyuu/pyenv-virtualenv#installing-with-homebrew-for-os-x-users) to them. I don't know if it would be possible to (or even if we should) automate this configuration step on behalf of the user. 
* The `pyenv-virtualenv` command should be used in lieu of `virtualenv`. This would require changes in `dev/_venv_common.sh`, or that we instruct users to create an alias. 
* This approach does not install `pip`, which is invoked in `dev/_venv_common.sh`.

kuba:
@douglasjose I would like to stress that only `./venv/bin/pip` is used throughout the process, global one is not necessary. `dev/_venv_common.sh` does `source ./venv/bin/activate` which puts `./venv/bin/pip` on `PATH`, this is why it's called without `./venv/bin` prefix.

Creating new environment with `virtualenv` automatically installs `pip` inside. I'm pretty sure that this should also be the case for `pyenv-virtualenv`.

It sucks, though, that they need to be called as `pyvenv virtualenv`, which would make our scripts more complicated. Hm... any other suggestions?

bootstrap/mac.sh
kuba: @BKreisel, how did you get `virtualenv` in the first place then?

BKreisel: @kuba, by `sudo easy_install virtualenv` ( uses the easy_install module default to OS X)

bmw:
Interesting. So if `easy_install` comes preinstalled on OS X, that is definitely the preferred approach in my opinion. Can anyone find anything confirming or denying this? I cannot seem to find anything.

BKreisel:
I was able to locate setuptools in `/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python` which was last updated `Aug 22, 2015`. This matches all of the other system file datestamps for the `El Capitan` release so I'm fairly confident it ships as part of the default Apple-branded Python 2.7

bmw:
Then unless anyone has any objections, I say we use the `easy_install` approach. I think the change should be as simple as appending `easy_install virtualenv` to the current version of `mac.sh` in master. @douglasjose, if you'd like to make the change in this PR, please do so and I'll happily merge and close out #968.

douglasjose:
@bmw works for me; I should be able to update the pull request by tomorrow.

kuba:
`sudo easy_install` pollutes OS-space, something I've been trying hard to avoid :(

bmw:
Do you have a preferred alternative @kuba?

kuba:
Second copy of Python, i.e. merging this PR.

bmw:
So after talking it through with @pde and @kuba, I'm merging this PR as is @douglasjose. Sorry for the indecisiveness and thanks again for writing it up.

bootstrap/mac.sh
```diff
@@ -6,3 +6,13 @@ fi
...
+fi
+
+if ! hash virtualenv 2>/dev/null; then
+    echo "virtualenv Not Installed\nInstalling with pip"
+    pip install virtualenv
```
arkarkark: on my machine this needed to be run with sudo

pde: Was it the brew step, or the pip step, that needed sudo?

arkarkark: This pip step. brew got uppity when I ran the whole script with sudo (rightly so).

PR #967, opened 2015-10-13
Only test CLI for nginx plugin if it is present

### LINKED ISSUE #919: CLITest.test_help assumes nginx plugin is installed
Our unittests fail if `nginx` plugin is not installed:

```
FAIL: test_help (letsencrypt.tests.cli_test.CLITest)
 File "/home/edh/Projects/abs/letsencrypt-git/src/letsencrypt-0.0.0.dev20151006/letsencrypt/tests/cli_test.py", line 78, in test_help
   self.assertTrue("--nginx-ctl" in out)
AssertionError: False is not true
```

We must not depend on the `nginx` plugin being installed!

https://github.com/letsencrypt/letsencrypt/blob/412e6acef48ad2ecf6d42726c1280a88dc3ddc5f/letsencrypt/tests/cli_test.py#L76-L78

Thanks to @Edenhofer for reporting that!

pde:
Fixes #919

### DIFF
--- a/letsencrypt/tests/cli_test.py
+++ b/letsencrypt/tests/cli_test.py
@@ -75,7 +75,10 @@ class CLITest(unittest.TestCase):
             output.truncate(0)
             self.assertRaises(SystemExit, self._call_stdout, ['-h', 'nginx'])
             out = output.getvalue()
-            self.assertTrue("--nginx-ctl" in out)
+            from letsencrypt.plugins import disco
+            if "nginx" in disco.PluginsRegistry.find_all():
+                # may be false while building distributions without plugins
+                self.assertTrue("--nginx-ctl" in out)
             self.assertTrue("--manual-test-mode" not in out)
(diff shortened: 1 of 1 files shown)


bmw:
LGTM. Tested locally without `letsencrypt-nginx` and did not encounter the problem.

kuba:
This test should be totally evicted from the suite and we should look for some plugin args that are automatically generated by plugins, e.g. `--manual-test-mode` instead. Otherwise we just reduced the coverage.

letsencrypt/tests/cli_test.py
```diff
@@ -75,7 +75,10 @@ def test_help(self):
             output.truncate(0)
             self.assertRaises(SystemExit, self._call_stdout, ['-h', 'nginx'])
             out = output.getvalue()
-            self.assertTrue("--nginx-ctl" in out)
+            from letsencrypt.plugins import disco
```
kuba: this should not be imported here

Issue #964, opened 2015-10-13
Search key_dir when performing Proof of Possession


bmw:
`proof_of_possession.py` currently relies only on the `installer` to determine the location of potential private keys that could be used to complete the challenge. While the `installer` should continue to be used, the code should also check the `key_dir` of the main configuration. This is where the client stores all private keys it creates when generating a CSR.

This will require a little refactoring, but `continuity_auth.py` has access to the configuration and should simply forward it along to `proof_of_possession.py`

bmw:
Significant progress has been towards this in the `search_key_dir_in_pop` branch.

No one else should try to work on this until PoP is supported in the client again.

schoen:
It seems like this concept has been totally removed from the ACME work (for example, it's not present at all in https://www.rfc-editor.org/internet-drafts/draft-ietf-acme-acme-18.txt, which is about to become RFC 8555). So I don't think we need to keep issues open about it; I'm not aware that there are any current WG proposals to reinstate this mechanism.

PR #970, opened 2015-10-14
Storage paranoia

### LINKED ISSUE #564: Sensible behavior if storage.py is interrupted while updating a cert lineage
We should have some kind of lock or status flag that indicates which lineage, if any, is currently being updated.  Then if we get interrupted by an exception or signal during the process, we should check that lineage for consistency and delete it (if it's new) or roll it back (if it already existed) if the current state is inconsistent.

Right now it's very bad if a cert lineage ends up in an inconsistent state; it can cause the renewer to crash.

bmw:
Fixes #564. It was correctly pointed out that if `storage.py` was interrupted while updating a lineage, bad things happened to good people. This PR fixes that.

The only functions where the lineage stored on disk is updated are `update_all_links_to`/`_update_link_to`, `new_lineage`, `save_successor`, and the new `_fix_symlinks`. I'll briefly examine each in turn and explain how the problem has been solved.

#### `update_all_links_to`
Since `_update_link_to` is a protected function and only called by `update_all_links_to`, I include it in here. The problem occurs if `storage.py` is interrupted before all symlinks are updated. this can leave the different elements of the lineage at different versions, with broken symlinks, with missing symlinks, etc.

The solution is that before modifying any symlinks, the current target of symlink being updated is stored in a file in the same directory as the symlink. For example, the file used for certs is `cert.lock`. Perhaps this is a bad name as these aren't really lockfiles, so if anyone has a better suggestion, please let me know.

After creating the files, symlinks are updated as normal and the "lockfiles" are deleted if no problem occurs. The recovery process runs if an exception/signal occurs and when a new `RenewableCert` is created. If recovery is necessary, one of two things happens:

1. If all lockfiles still exist, we may not have finished updating all of the symlinks. The contents of the lockfiles are read and all changes are reverted.
2. If not all of the lockfiles exist, we either haven't started or already finished modifying the symlinks so these files can be deleted.

#### _fix_symlinks
It's not a problem if this function is interrupted. The lockfiles are not deleted until all symlinks have been fixed. The symlinks will be updated and the lockfiles will be deleted as necessary the next time a `RenewableCert` is created for that lineage.

#### `save_successor`
Keep in mind that this function does not update any existing symlinks. The first thing this function does is determine the version number for the successor. This is determined by examining directories the symlinks point to, looking for files of the form `cert2.pem`, `cert3.pem`, etc.

If we crash before writing anything to disk, it is not a problem. We already saved the key [here](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/client.py#L210) so we have it for proof of possession challenges (see #964). If we crash before writing everything to disk, the code will automatically use a higher version number in the future to avoid conflicts. The user will have to request another cert, but it is impossible to ensure all necessary files are stored to disk.

#### `new_lineage`
The first thing this function does that is specific to the lineage is call `le_util.unique_lineage_name`. This returns an open file descriptor for the renewal config. At this point, this lineage is represented on disk. In the future, when new lineages are created, a different lineage name will be chosen. 

The last thing this function does is write the contents of the renewal_config to disk. When a `RenewableCert` is created, it examines it's configuration. If it either cannot parse the configuration or any of the necessary elements are missing, a `CertStorageError` is raised which is correctly handled everywhere a `RenewableCert` is created directly ([here](https://github.com/letsencrypt/letsencrypt/compare/storage_paranoia?expand=1#diff-54b2f468d57ef7436008bd031bc89711R182) and [here](https://github.com/letsencrypt/letsencrypt/compare/storage_paranoia?expand=1#diff-88e9d12789ce14158dde8ef58d5f0e16R181)). Therefore, it's possible that a crash will leave extraneous files in `/etc/letsencrypt`, but all relevant code will handle the situation appropriately.

cc @schoen


bmw:
Incorporated @schoen's suggestions. Ready for re-review.

schoen:
Looks good to me!

bmw:
Merging with @schoen's out of band review,

PR #974, opened 2015-10-14
Add rateLimited error type.

jsha:


kuba:
LGTM. If you merge it now it will land in next dev release :)

Issue #880, opened 2015-10-02
Determine how the renewer will run given different installation methods


bmw:
We need to figure out how the renewer will be configured to run automatically regardless of how the user installs Let's Encrypt. When prepared for use with an operating system package manager, this setup should be done by the package maintainer.

If installed through something like PyPI, we should help the user with this configuration. This could be anything from providing documentation to attempting to automate the process for the user (after getting their approval). Another idea is to include crontabs, systemd timers, etc. with Let's Encrypt to help the user through the process. Regardless of how this done, we should help the user with this instead of leaving them to figure it out on their own.

HLFH:
Cool. I would love systemd timers support. I don't use cron.

GreatEmerald:
+1 for including a timer unit. The additional nicety of that is that the user needs to explicitly enable timers by default, as opposed to crontabs that run by default and are rather difficult to disable without confusing package managers.

sandys:
hi,
is there a systemd timer documented anywhere (even if it is not used in the package). We are starting to use letsencrypt and we are baking this in our Docker images. Would love to have this in already.

bmw:
`letsencrypt-renewer` (a separate executable from `letsencrypt`) is still under heavy development and is not ready for use. There are no cron tabs or systemd timers right now.

pde:
Proposed solution for 0.4.0: print a URL to renewal documentation :)

pde:
(Or print the documentation)

pde:
Actually maybe we do still want some automation for this, but it's 0.5.0.

schoen:
@bmw, do we still consider this a separate issue from #3522?

bmw:
I don't think so.

Our supported methods of installation are OS packages, certbot-auto, and Docker.

Configuring renewal in OS packages is outside of our control, #3522 handles certbot-auto, and #6657 handles Docker.

Issue #875, opened 2015-10-01
Properly report renewal status to the user


bmw:
When `obtain_and_enroll_certificate` is called in `client.py`, it examines the renewal configuration and tells the user if automatic renewal/deployment has been enabled for their certificate. These messages should only be displayed if the renewer is actually configured to run automatically. Otherwise, these messages will confuse the user.

Additionally, renewal status should also be reported if the new certificate is considered a renewal of an existing certificate.

bmw:
Always determining correctly in the client whether or not the renewer is configured to run is a hard problem. Perhaps these messages should simply say that automated renewal/deployment will occur if the renewer is scheduled to run your system.

SwartzCr:
should this be closed now that #978 is merged?

pde:
Closed by #2392

Issue #428, opened 2015-05-18
Deployment of renewed certs needs to call .restart()


schoen:
Currently the renewer script doesn't attempt to instantiate the installer and hence can't invokve the installer's `.restart()` method to restart the web server. Restarting the web server would be expected behavior as part of a deployment because it's necessary in order to make the server actually start using the new cert.

(The file links are successfully updated as part of the deployment step, but most server software wouldn't know to re-read them.)

The process of instantiating the installer ought to be exactly parallel to the existing process of instantiating the authenticator.

schoen:
I think this looks something like

```
def deploy(cert, desired_version):
    """Change the current version of the referenced cert.

    :param class:`letsencrypt.storage.RenewableCert` cert: the certificate
        lineage to affect.
    :param int desired_version: the version of the certificate lineage
        that should become current."""

    cert.update_all_links_to(desired_version)

    # Try to restart server
    if "renewalparams" not in cert.configfile:
        # TODO: notify user?
        return
    renewalparams = cert.configfile["renewalparams"]
    if "installer" not in renewalparams:
        # TODO: notify user?
        return
    # Instantiate the appropriate authenticator
    plugins = plugins_disco.PluginsRegistry.find_all()
    config = configuration.NamespaceConfig(AttrDict(renewalparams))
    # XXX: this loses type data
    try:
        installer = plugins[renewalparams["installer"]]
    except KeyError:
        # TODO: Notify user? (installer could not be found)
        return
    installer = installer.init(config)
    installer.restart()
```

based on the existing authenticator instantiation. This is currently hard to test due to client/server interoperability problems, so we should try it out once the client and server are happy talking to each other again.

marcan:
It would be nice if the plugin could be passed the name of the domain (or list of domains) as part of the restart() method, or have some other function invoked to notify of the changed certs. I'm writing a simple plugin that just shells out to a shell script for authentication requests, so that you can easily integrate it with custom server configs. This includes multiple server types, e.g. the shell script could know how to update certs for nginx, icecast, and postfix, for example.

Without knowing which certs were updated, the plugin would either have to restart all servers every time any certificate is renewed, or scrape the mtime of certificate files to figure out which ones changed.

pde:
@bmw has code for this in the `origin/autodeploy` branch, but it needs to be separated from some other changes.

bmw:
Confirmed fixed in #2338.

Created #2387 for tracking the messages printed to the user.

PR #978, opened 2015-10-14
Removed misleading renewal messages

bmw:
Due to problems like #428, #875, #880, I'm removing all renewer messages for the time being as to not confuse users.


kuba:
LGTM

letsencrypt/tests/account_test.py
```diff
@@ -70,12 +70,12 @@ def _call(self):
         from letsencrypt.account import report_new_account
         report_new_account(self.acc, self.config)
 
-    @mock.patch("letsencrypt.client.zope.component.queryUtility")
+    @mock.patch("letsencrypt.account.zope.component.queryUtility")
```
kuba: oh wow, nice catch

Issue #414, opened 2015-05-13
Client should strongly encourage use of an email address


jsha:
If the CA has many (/any) users with no email on file, it's going to cause a lot of hassles. We're still deciding if we want it to be official policy that everyone has to provide an email, but in the meantime we can work on the client side of this. Proposal: The interactive version of the client should ask for an email, and not proceed without one, unless it was invoked with a --allow-unsafe-registration. The help documentation for --allow-unsafe-registration would say:

"Specifying this flag enables registering an account with no email address. This is strongly discouraged, because in the event of key loss or account compromise you will irrevocably lose access to your account. You will also be unable to receive notice about impending expiration of revocation of your certificates. Updates to the Subscriber Agreement will still affect you, and the CA will make a best-effort attempt to notify you about them based on information about the affected domains, such as WHOIS data and common administrative addresses."

kuba:
duplicate of #410?

jsha:
One idea on this: We want it to be more convenient to provide an email than to not provide one. I suspect the most common unintended no-email case would be copy-pasted command lines.

Proposal: By default, the UI and command line require an email address. To register with no email address, you must pass the client an --allow-unsafe-registration and use the UI (not the command line). The UI would prompt you if you leave the email address empty with a description of all the ways in which it's risky (no expiration notices, no account recovery (which can mean inability to issue future certs), and no notice of subscriber agreement changes).

miquelruiz:
FYI I'm working on this. I hope I'll have a PR ready this weekend.

bmw:
Fantastic! We appreciate the help @miquelruiz. Welcome to the project.

jsha:
Thanks @miquelruiz! Note: in my original post, I included language "and the CA will make a best-effort attempt to notify you about them based on information about the affected domains, such as WHOIS data and common administrative addresses." That's no longer true: for people who don't provide email addresses, updates to the Subscriber agreement are effective N days after posting an update to the web site.

miquelruiz:
Is there a definitive value for "N days" to be included in the help?

miquelruiz:
Pull request attached. Please feel free to suggest any improvement :)

PR #979, opened 2015-10-14
Explain email recovery better

### LINKED ISSUE #659: email address issues
1.) There is no explanation what the email is for. You should say something like "Please enter a valid email address for recovery purposes and notifications. We will not add it to your certificate nor will we share it. You can also leave this field empty, but that makes recovery harder."

2. ) Right now when an invalid email address is given, the error message says 'an unexpected error occured, please check the logfile'. This should be changed to something along the lines of "You gave an invalid email address. Please enter a valid email address for recovery purposes and notifications, or leave the field empty to give no email address".

pde:
Closes #659.

More work will be required for #414.

### DIFF
--- a/letsencrypt/display/ops.py
+++ b/letsencrypt/display/ops.py
@@ -124,7 +124,7 @@ def get_email():
     """
     while True:
         code, email = zope.component.getUtility(interfaces.IDisplay).input(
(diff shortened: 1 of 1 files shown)


bmw:
LGTM

PR #980, opened 2015-10-15
letsencrypt-auto implements bootstrapping, venv install, and pip autoupdate before running the client

pde:
You can just run `./letsencrypt-auto` ... this will handle everything most users should need in a single command.

### DIFF
--- a/bootstrap/venv.sh
+++ b/bootstrap/venv.sh
@@ -1,0 +1,35 @@
+#!/bin/sh -e
+#
+# Installs and updates letencrypt virtualenv
+#
+# USAGE: source ./dev/venv.sh
+
+
+XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
+VENV_NAME="letsencrypt"
+VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
+
+# virtualenv call is not idempotent: it overwrites pip upgraded in
+# later steps, causing "ImportError: cannot import name unpack_url"
+if [ ! -d $VENV_PATH ]
+then
+  virtualenv --no-site-packages --python python2 $VENV_PATH
+fi
+
+. $VENV_PATH/bin/activate
+pip install -U setuptools
+pip install -U pip
+
+# TODO: install apache and nginx plugins by default?
+# --pre is not necessary for dev releases in more recent pip versions
+pip install -U letsencrypt letsencrypt-apache letsencrypt-nginx
+
+echo
+echo "Congratulations, Let's Encrypt has been successfully installed/updated!"
+echo
+echo -n "Your prompt should now be prepended with ($VENV_NAME). Next "
+echo -n "time, if the prompt is different, 'source' this script again "
+echo -n "before running 'letsencrypt'."
+echo
+echo
+echo "You can now run 'letsencrypt --help'."
--- a/bootstrap/venv.sh
+++ b/bootstrap/venv.sh
@@ -20,9 +20,7 @@ fi
 pip install -U setuptools
 pip install -U pip
 
-# TODO: install apache and nginx plugins by default?
-# --pre is not necessary for dev releases in more recent pip versions
-pip install -U letsencrypt letsencrypt-apache letsencrypt-nginx
+pip install -U letsencrypt letsencrypt-apache # letsencrypt-nginx
 
 echo
 echo "Congratulations, Let's Encrypt has been successfully installed/updated!"
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -371,24 +371,3 @@ i.e. it does not use ports.
 FreeBSD by default uses ``tcsh``. In order to activate virtulenv (see
 below), you will need a compatbile shell, e.g. ``pkg install bash &&
 bash``.
-
-
-Running with Docker
-===================
-
-Docker_ is another way to quickly obtaintesting certs. From the
-server that the domain your requesting a cert for resolves to,
-`install Docker`_, issue the following command:
-
-.. code-block:: shell
-
-   sudo docker auth -it --rm -p 443:443 --name letsencrypt \
-               -v "/etc/letsencrypt:/etc/letsencrypt" \
-               -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
-               quay.io/letsencrypt/letsencrypt:latest
-
-and follow the instructions. Your new cert will be available in
-``/etc/letsencrypt/certs``.
-
-.. _Docker: https://docker.com
-.. _`install Docker`: https://docs.docker.com/docker/userguide/
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -312,6 +312,8 @@ In general:
 * ``virtualenv`` and ``pip`` are used for managing other python library
   dependencies
 
+.. _Augeas: http://augeas.net/
+.. _Virtualenv: https://virtualenv.pypa.io
 
 Ubuntu
 ------
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -297,6 +297,8 @@ commands:
 This should generate documentation in the ``docs/_build/html``
 directory.
 
+.. _prerequisites:
+
 Notes on OS depedencies
 =======================
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -296,3 +296,99 @@ commands:
 
 This should generate documentation in the ``docs/_build/html``
 directory.
+
+Notes on OS depedencies
+=======================
+
+OS level dependencies are managed by scripts in ``bootstrap``.  Some notes
+are provided here mainly for the :ref:`developers <hacking>` reference.
+
+In general:
+
+* ``sudo`` is required as a suggested way of running privileged process
+* `Augeas`_ is required for the Python bindings
+* ``virtualenv`` and ``pip`` are used for managing other python library
+  dependencies
+
+
+Ubuntu
+------
+
+.. code-block:: shell
+
+   sudo ./bootstrap/ubuntu.sh
+
+
+Debian
+------
+
+.. code-block:: shell
+
+   sudo ./bootstrap/debian.sh
+
+For squeeze you will need to:
+
+- Use ``virtualenv --no-site-packages -p python`` instead of ``-p python2``.
+
+
+.. _`#280`: https://github.com/letsencrypt/letsencrypt/issues/280
+
+
+Mac OSX
+-------
+
+.. code-block:: shell
+
+   ./bootstrap/mac.sh
+
+
+Fedora
+------
+
+.. code-block:: shell
+
+   sudo ./bootstrap/fedora.sh
+
+
+Centos 7
+--------
+
+.. code-block:: shell
+
+   sudo ./bootstrap/centos.sh
+
+
+FreeBSD
+-------
+
+.. code-block:: shell
+
+   sudo ./bootstrap/freebsd.sh
+
+Bootstrap script for FreeBSD uses ``pkg`` for package installation,
+i.e. it does not use ports.
+
+FreeBSD by default uses ``tcsh``. In order to activate virtulenv (see
+below), you will need a compatbile shell, e.g. ``pkg install bash &&
+bash``.
+
+
+Running with Docker
+===================
+
+Docker_ is another way to quickly obtaintesting certs. From the
+server that the domain your requesting a cert for resolves to,
+`install Docker`_, issue the following command:
+
+.. code-block:: shell
+
+   sudo docker auth -it --rm -p 443:443 --name letsencrypt \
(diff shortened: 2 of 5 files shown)


pde:
This is on top of kuba's work on #976.

pde:
Tested on Debian unstable, Ubuntu 14.04.02, Ubuntu 15.01, FreeBSD, and Mac OS X.

On all of these systems it seemed to do its job, though on FreeBSD and Ubuntu Vivid there were problems unrelated to this script which I may file separate tickets about.

pde:
We could probably use a test on at least one RedHat based system.

pde:
Future question: should we really be using the pypi server to provide letsencypt?  Since we use git to fetch the tree,  should just pip install locally (and use git to provide our auto-updating for us?)

Alternatively, we could make the letsencypt-auto script a standalone thing (though it would need to incorporate or be bundled with the bootstrap scripts).

bmw:
Other than `docs/using.rst` needing to be updated to reflect your changes, this LGTM!

docs/using.rst
```diff
@@ -44,17 +24,62 @@ above method instead.
...
+
+To install and run the client you just need to type:
+
+.. code-block:: shell
+
+   ./letsencrypt-auto
+
+.. warning:: Please do **not** use ``python setup.py install``.  That mode of
+             operation might corrupt your operating system and is **not supported**
+             by the Let's Encrypt team!
```
kuba: I very much preferred previous warning message text. It explicitly disallowed certain commands that I've been seeing a lot while supporting clients on IRC. This is rather a step back.

docs/using.rst
```diff
@@ -44,17 +24,62 @@ above method instead.
 
 .. _prerequisites:
```
kuba: you removed section title, but not reflink - this must have broken something

kuba:
DigitalOcean droplet:
```
# ./letsencrypt-auto 
Bootstrapping dependencies for Debian-based OSes...
./letsencrypt-auto: 30: ./letsencrypt-auto: sudo: not found
```
BTW, I'm running this as root.

bmw:
I got the same problem.

letsencrypt-auto
```diff
@@ -0,0 +1,74 @@
...
+# without requiring specific versions of its dependencies from the operating
+# system.
+
+XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
+VENV_NAME="letsencrypt"
+VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
+VENV_BIN=${VENV_PATH}/bin
+
+for arg in "$@" ; do 
+  if [ "$arg" = "-v" ] || [ "$arg" = "--verbose" ] ; then
```
kuba: This doesn't work with multiple v's (`-vvvvv`), something that I always ask for when debugging stuff for clients.

letsencrypt-auto
```diff
@@ -0,0 +1,74 @@
...
+
+echo "Updating letsencrypt and virtual environment dependencies..."
+if [ "$VERBOSE" = 1 ]  ; then
+  $VENV_BIN/pip install -U setuptools
+  $VENV_BIN/pip install -U pip
+  $VENV_BIN/pip install -U letsencrypt letsencrypt-apache letsencrypt-nginx
+else
+  $VENV_BIN/pip install -U setuptools > /dev/null
+  $VENV_BIN/pip install -U pip > /dev/null
+  $VENV_BIN/pip install -U letsencrypt letsencrypt-apache letsencrypt-nginx > /dev/null
```
kuba: Can we dedup above 7 lines? Maybe put all `pip install` in a function, and redirect `>/dev/null` function call?

Same for `virtualenv --no-site-packages`.

docs/using.rst
```diff
@@ -44,17 +24,62 @@ above method instead.
...
+  set).
+
+.. keep it up to date with constants.py
+
+
+.. _Augeas: http://augeas.net/
+.. _Virtualenv: https://virtualenv.pypa.io
+
+Notes on OS depedencies
+=======================
```
kuba: This section should not be duplicated. Please remove from this file and leave it only in `contributing.rst`

docs/using.rst
```diff
@@ -44,17 +24,62 @@ above method instead.
...
-Debian**. Therefore, prerequisites for other platforms listed below
-are provided mainly for the :ref:`developers <hacking>` reference.
+OS level dependencies are managed by scripts in ``bootstrap``.  Some notes
+are provided here mainly for the :ref:`developers <hacking>` reference.
 
 In general:
 
 * ``sudo`` is required as a suggested way of running privileged process
 * `Augeas`_ is required for the Python bindings
+* ``virtualenv`` and ``pip`` are used for managing other python library
```
kuba: Listing `pip` is likely to be very confusing: c.f. https://github.com/letsencrypt/letsencrypt/pull/963#issuecomment-148278310. `pip` is installed automatically on `virtualenv` invocation - we don't use global `pip`.

letsencrypt-auto
kuba: Actually, you don't even need function.

```bash
if [ "$VERBOSE" = 1 ]  ; then
  , ./bootstrap/venv.sh
else
  . ./bootstrap/venv.sh >/dev/null
fi

# Explain what's about to happen, for the benefit of those getting sudo
# password prompts...
echo "Running with virtualenv:" sudo $(which letsencrypt) "$@"
letsencrypt $@
```

Just remove all `echo` lines from `./dev/venv.sh`.

BTW this is what I meant in our call

docs/using.rst
```diff
@@ -119,64 +144,26 @@ below), you will need a compatbile shell, e.g. ``pkg install bash &&
...
-handy:
-
-.. include:: ../examples/dev-cli.ini
-   :code: ini
-
-By default, the following locations are searched:
+   sudo docker auth -it --rm -p 443:443 --name letsencrypt \
+               -v "/etc/letsencrypt:/etc/letsencrypt" \
+               -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
+               quay.io/letsencrypt/letsencrypt:latest
```
kuba: Add `auth` command at the end. Otherwise `Docker` doesn't make sense.

docs/using.rst
pde: My thinking was that we should remove all of the parts of the documentation that tell people to explicitly do anything with venv and pip, and then they won't end up issuing mistaken commands involving those tools.

letsencrypt-auto
```diff
@@ -0,0 +1,74 @@
...
+fi
+
+# Explain what's about to happen, for the benefit of those getting sudo
+# password prompts...
+echo "Running with virtualenv:" sudo $VENV_BIN/letsencrypt "$@"
+sudo $VENV_BIN/letsencrypt $@
```
kuba: shouldn't this be `sudo $VENV_BIN/letsencrypt "$@"` (with quotes)?

docs/using.rst
kuba: I see, but in my long IRC support experience, I've seen a lot of them jumping straight to `pip install .` (without reading much docs), which is becoming more standard in Python world.

Also, you removed the part about forbidden use of `sudo` or `root`. While it's still true for all previously mentioned cases, this also applies to the wrapper: `sudo ./letsencrypt-auto` should not be called.

kuba:
After installing `sudo`, the script seems to work in general, though quite slow. Handling of `-vvvvv...` should definitely be fixed, otherwise we'll have hard time debugging.

kuba:
Another caveat:
```
# ./letsencrypt-auto help
Updating letsencrypt and virtual environment dependencies...
Running with virtualenv: sudo /root/.local/share/letsencrypt/bin/letsencrypt help
usage: 

the cert.  
letsencrypt: error: argument SUBCOMMAND: invalid choice: 'help' (choose from 'run', 'auth', 'install', 'revoke', 'rollback', 'config_changes', 'plugins')
```

CLI should be using `sys.argv[0]` to detect name by which it was called for the "usage" message and error, and other cases that I can't think of now.

kuba:
We should also install `letshelp-letsencrypt` PyPI package.

I forgot about it in my own script. BTW, another reason for dedup - we need to update at least 3 places now to include `letshelp-letsencrypt`.

docs/contributing.rst
```diff
@@ -296,3 +296,78 @@ commands:
...
+=======================
+
+OS level dependencies are managed by scripts in ``bootstrap``.  Some notes
+are provided here mainly for the :ref:`developers <hacking>` reference.
+
+In general:
+
+* ``sudo`` is required as a suggested way of running privileged process
+* `Augeas`_ is required for the Python bindings
+* ``virtualenv`` and ``pip`` are used for managing other python library
```
kuba: my comment about listing `pip` is still true in this file

letsencrypt-auto
```diff
@@ -0,0 +1,83 @@
...
+  $VENV_BIN/pip install -U setuptools > /dev/null
+  $VENV_BIN/pip install -U pip > /dev/null
+  # nginx is buggy / disabled for now...
+  $VENV_BIN/pip install -U letsencrypt letsencrypt-apache > /dev/null
+fi
+
+# Explain what's about to happen, for the benefit of those getting sudo
+# password prompts...
+echo "Running with virtualenv:" $SUDO $VENV_BIN/letsencrypt "$@"
+$SUDO $VENV_BIN/letsencrypt $@
```
kuba: you didn't address quotes comment: `$SUDO $VENV_BIN/letsencrypt "$@"`

docs/using.rst
```diff
@@ -42,126 +22,35 @@ above method instead.
    https://github.com/letsencrypt/letsencrypt/archive/master.zip
 
 
-.. _prerequisites:
```
kuba: this reflink is still referenced in other places in the documentation. It should not be removed

letsencrypt-auto
```diff
@@ -0,0 +1,89 @@
...
+    $SUDO $BOOTSTRAP/freebsd.sh
+  elif uname | grep -iq Darwin ; then
+    echo "Bootstrapping dependencies for Mac OS X..."
+    echo "WARNING: Mac support is very experimental at present..."
+    $BOOTSTRAP/mac.sh
+  else
+    echo "Sorry, I don't know how to bootstrap Let's Encrypt on your operating system!"
+    echo
+    echo "You will need to bootstrap, configure virtualenv, and run a pip install manually"
+    echo "Please see https://letsencrypt.readthedocs.org/en/latest/using.html#prerequisites"
```
kuba: Instructions have been moved to contributing.html`, right?

bootstrap/venv.sh
```diff
@@ -0,0 +1,35 @@
...
+  virtualenv --no-site-packages --python python2 $VENV_PATH
+fi
+
+. $VENV_PATH/bin/activate
+pip install -U setuptools
+pip install -U pip
+
+# TODO: install apache and nginx plugins by default?
+# --pre is not necessary for dev releases in more recent pip versions
+pip install -U letsencrypt letsencrypt-apache letsencrypt-nginx
```
kuba: we decided not to install `letsencrypt-nginx`

Are you going to `source` this script from the `letsencrypt-auto` wrapper?

docs/using.rst
pde: Where is it referenced?  I can only find a reference in the letsencrypt-auto script :)

kuba:
Please run `make -C docs clean html` and check for warnings/errors and the view generated html files in `docs/_build/html`.

```
/home/jakub/dev/letsencrypt/letsencrypt/docs/contributing.rst:309: ERROR: Unknown target name: "augeas".
```

docs/using.rst
kuba:
```
$ git grep prerequisites
docs/contributing.rst:Start by :ref:`installing Let's Encrypt prerequisites
docs/contributing.rst:<prerequisites>`. Then run:
docs/contributing.rst:encouraged) make sure to install prerequisites as listed in
letsencrypt-auto:    echo "Please see https://letsencrypt.readthedocs.org/en/latest/using.html#prerequisites"
```

pde:
OK, merging.  Anything else is going in another issue / PR :)

kuba:
Please don't merge without an LGTM from at least one other team member. It's not the first time you're doing it. That's not the policy team has been using. It's also proven that your unreviewed PRs tend to bring errors that don't get fixed for a long time after PR is merged and you often have to be chased about it (e.g. see #919). Moreover, I thought we explicitly agreed on merging this around now, after another review, not 10 hours ago, which makes this behavior even worse.

A couple of comments were totally ignored:
- https://github.com/letsencrypt/letsencrypt/pull/980#discussion_r42083277 and https://github.com/letsencrypt/letsencrypt/pull/980#discussion_r42087107
- https://github.com/letsencrypt/letsencrypt/pull/980#discussion_r42082641
- calling out `source ./bootstrap/venv.sh`

PR #983, opened 2015-10-15
Quick renewal status fix

bmw:
Unfortunately, I wasn't quite able to get all of my renewer changes finished in time (see my progress on the autodeploy branch). Throwing this up here as a quick fix for reporting certificate expiration to the user. It basically tells the user they're on their own for renewal and doesn't say anything if you run the client with your own CSR.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -303,9 +303,9 @@ def _auth_from_domains(le_client, config, domains, plugins):
     _report_new_cert(lineage.cert)
     reporter_util = zope.component.getUtility(interfaces.IReporter)
     reporter_util.add_message(
-        "Your certificate will expire at {0}. To obtain a new version of the "
+        "Your certificate will expire on {0}. To obtain a new version of the "
         "certificate in the future, simply run this client again.".format(
-            lineage.notafter().ctime()),
+            lineage.notafter().date()),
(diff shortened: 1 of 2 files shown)


letsencrypt/cli.py
```diff
@@ -301,6 +301,12 @@ def _auth_from_domains(le_client, config, domains, plugins):
             raise errors.Error("Certificate could not be obtained")
 
     _report_new_cert(lineage.cert)
+    reporter_util = zope.component.getUtility(interfaces.IReporter)
+    reporter_util.add_message(
+        "Your certificate will expire at {0}. To obtain a new version of the "
+        "certificate in the future, simply run this client again.".format(
+            lineage.notafter().ctime()),
```
kuba: is that UTC time? we should indicate timezone as I'm confused myself

kuba: FTR I'm getting "Your certificate will expire at Wed Jan 13 04:12:00 2016. To obtain a new version of the certificate in the future, simply run this client again."

kuba:
Tested on my VPS. Would LGTM it, but'd rather have the time(zone) confusion fixed somehow if it's possible.

kuba:
quikfix is good, tested on DO; LGTM

PR #986, opened 2015-10-15
Autodeployment

### LINKED ISSUE #428: Deployment of renewed certs needs to call .restart()

bmw:
Fixes #428. This was the code I wanted to submit earlier when I submitted #983. This PR supersedes those changes.

This PR also does a better job of informing users about the renewal. The basic idea is that if the user ran the client with an installer or configurator, it will inform them about the renewer. Otherwise, it will simply tell them when their certificates expire and that they can run the client again to renew them.

I tested this on a server with Apache and it worked properly. I briefly tried to get this to work in integration tests, but I encountered the problem of the renewer trying to autorenew certificates that used manual. This is related to this PR only in that maybe we shouldn't be pointing people towards the renewer until problems like this are fixed.

While the changes I made here are relatively straight forward, I'm not certain whether we want this in the next PyPI release. I wrote all of this in a bit of a hurry today and it hasn't been thoroughly tested.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -342,12 +342,6 @@ def _auth_from_domains(le_client, config, domains, plugins):
             raise errors.Error("Certificate could not be obtained")
 
     _report_new_cert(lineage.cert)
-    reporter_util = zope.component.getUtility(interfaces.IReporter)
-    reporter_util.add_message(
-        "Your certificate will expire on {0}. To obtain a new version of the "
-        "certificate in the future, simply run this client again.".format(
-            lineage.notafter().date()),
-        reporter_util.MEDIUM_PRIORITY)
 
     return lineage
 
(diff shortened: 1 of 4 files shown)


bmw:
Two more reasons we don't want this right now are #988 and #989.

bmw:
Closing this as the renewer/plugins aren't ready for this behavior. I'll factor out additional changes and submit a new PR for fixes to #428.

PR #895, opened 2015-10-04
Standalone 2.0

kuba:
Standalone plugin rewritten! Now supports both `DVSNI` and `SimpleHTTP` challenges! :)

May I please ask volunteers to test it? Check out this branch/PR, `rm -rf letsencrypt/plugins/standalone` to remove old `*.pyc` files, `pip uninstall letsencrypt` and install it once again (`pip install .`, or `pip install -e .` for devs). Normally type of the challenge is picked at random, but you can force one by using `--standalone-supported-challenges dvsni` or `--standalone-supported-challenges simpleHttp`.

Major differences in the code?
- use of threads instead of processes, which allows easier IPC (removing problematic signals handling)
- `acme.standalone` exposed for other client implementations
- built around stdlib `SocketServer`, `BaseHTTPServer` etc.
- generic `acme.crypto_util.SSLSocket`
- tons of other rewrites to make code more modular

Biggest caveat: Simple HTTP TLS works using an internal CPython  2.* API... but recent spec changes drop Simple HTTP TLS as a whole, so we should be fine once new spec is implemented.

This is an early preview, bugs are anticipated and your feedback appreciated!

PR #994, opened 2015-10-15
Integration tests: SimpleHTTP on port 5002.

kuba:
Should be pulled by @jsha in lockstep with https://github.com/letsencrypt/boulder/commit/6b8f1c27003bb43f33ffcfa9a5f63eb4154c6df9.

This will be in conflict with #895 and any other outstanding PRs that don't merge with it.

### DIFF
--- a/tests/integration/_common.sh
+++ b/tests/integration/_common.sh
@@ -16,7 +16,7 @@ letsencrypt_test () {
         --server "${SERVER:-http://localhost:4000/directory}" \
         --no-verify-ssl \
         --dvsni-port 5001 \
-        --simple-http-port 5001 \
+        --simple-http-port 5002 \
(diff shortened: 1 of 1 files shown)


jsha:
@kuba: Let's actually do this a slightly more gradual way that doesn't break tests at all. Can you say:

```
--simple-http-port ${SIMPLE_HTTP_PORT:-5001}
```

And then in the Boulder tests I'll set SIMPLE_HTTP_PORT=5002 in tandem with the change. We can then remove the code when we are done.

kuba:
@jsha, fixed

jsha:
Goose fetch failed. I restarted the Travis run.

jsha:
LGTM

PR #995, opened 2015-10-15
Get precompiled Goose from GithHub.

jsha:
Rather than fetching from bitbucket and building. Bitbucket is often down, and
building from scratch is slow. Github is sometimes down, but at least now we
have our eggs in one basket.


bmw:
LGTM! Thanks @jsha.

PR #997, opened 2015-10-16
Make the script POSIX compliant so it works with dash (Fixes #977)

### LINKED ISSUE #977: Boulder integration test: unexpected operator
https://travis-ci.org/letsencrypt/letsencrypt/letsencrypt/jobs/85367670#L1045

```
+uname
+[ Linux == Darwin ]
./tests/boulder-integration.sh: 17: [: Linux: unexpected operator
```

Interestingly, despite `-x -e`, scripts proceeds to a next step.

ddworken:
Looks like the problem was introduced [here](https://github.com/letsencrypt/letsencrypt/commit/c9e28309ed2d3efb3c2c21301f88796fe87169ab) likely due to @BKreisel having /usr/bin/sh set to bash (which this works on) rather than dash (which this fails on). 

So the fix is pretty simple, just switch to ```=``` instead of ```==``` so as to make the script POSIX compliant to fix #977.

### DIFF
--- a/tests/boulder-integration.sh
+++ b/tests/boulder-integration.sh
@@ -14,7 +14,7 @@ export PATH="/usr/sbin:$PATH"  # /usr/sbin/nginx
 export GOPATH="${GOPATH:-/tmp/go}"
 export PATH="$GOPATH/bin:$PATH"
 
-if [ `uname`  == 'Darwin' ]; then
+if [ `uname` = "Darwin" ];then
   readlink="greadlink"
 else
   readlink="readlink"


ddworken:
[No more error](https://travis-ci.org/letsencrypt/letsencrypt/jobs/85645573#L1074)

bmw:
LGTM. Thanks @ddworken.

tests/boulder-integration.sh
```diff
@@ -14,7 +14,7 @@ export PATH="/usr/sbin:$PATH"  # /usr/sbin/nginx
 export GOPATH="${GOPATH:-/tmp/go}"
 export PATH="$GOPATH/bin:$PATH"
 
-if [ `uname`  == 'Darwin' ]; then
+if [ `uname` = "Darwin" ];then
```
kuba: you removed whitespace :(

ddworken: Sorry about that. Are we following a specific style guide for shell code? I looked up the [Google Style Guide for shell scripts](http://google.github.io/styleguide/shell.xml) and I can tell we aren't following that since it says to always use bash (though it does agree with you that it should be ```; then```). Just grepping for ```;then``` and ```; then``` on my system, I see plenty of both (in packages provided by Ubuntu).

PR #1002, opened 2015-10-16
Detect downstream deps in tests

kuba:
This is meant to detect issues such as #919. It makes sure we run tests without any downstream dependencies installed, e.g. `letsencrypt_apache` is not installed when `letsencrypt` is being tested.

It also improves verbosity of the tests, so it's a lot easier to detect problems (especially busy waiting loops) when investigating Travis build logs output.


kuba:
Merged with `master`, ready to pull.

bmw:
LGTM

PR #1010, opened 2015-10-17
Integration tests: force le3.wtf over DVSNI.

kuba:
Becuse of https://github.com/letsencrypt/boulder/issues/985, we need to force standalone to one particular type of challenge at a time. #895 didn't fix it properly - it forgot about issuance of le3.wtf (CSR-initiated), which results in Travis builds failing 50% of a time, e.g. https://travis-ci.org/letsencrypt/letsencrypt/jobs/85875483#L3318.

Please merge before anything else to reduce number of failed builds :)

PR #1011, opened 2015-10-17
Fix dev release script: don't install from PyPI in tests

kuba:
There are a couple of fixes addressed by this PR:

1. updating `pip`/`setuptools` to latest: this matches rest of our codebase and lets us drop archaic `--pre`; just this change would also fix (2) below due to way newer versions of `pip` work...

2. script was confusing package names (with hyphens) and module names (with underscores): we were calling `pip instal letsencrypt_apache`, which failed to be found in the temporary index created by `python -m SimpleHTTPServer 1234`, but PyPI happily converted the name to correct `letsencrypt-apache` and served `0.0.0.dev20151008` instead of the locally created `0.0.0.dev20151017`

3. creating `virtualenv` for the dev-release: this installs necessary `wheel` and... `virtualenv` (!) - this is to make sure that virtualenv created later inherits latest versions of `pip`/`setuptools`/`wheel` (this achieves the same affect as (1), but caused a hell of confusion during last release)

This PR blocks next dev release.

### DIFF
--- a/tools/dev-release.sh
+++ b/tools/dev-release.sh
@@ -74,9 +74,10 @@ python -m SimpleHTTPServer $PORT &
 # installed from local PyPI rather than current directory (repo root)
 virtualenv --no-site-packages ../venv
 . ../venv/bin/activate
-# Now, use our local PyPI. --pre allows installation of pre-release (incl. dev)
+pip install -U setuptools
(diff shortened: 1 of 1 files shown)


kuba:
Build affected by #1010.

PR #1013, opened 2015-10-17
Use PEM for Account.id generation.

### LINKED ISSUE #1012: cryptography==0.8 breaks client
Follow-up from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=203405#c17. It seems that our `cryptography>=0.8` requirement is too weak. After `pip install cryptography==0.8` the following error is thrown:

```
  File "/root/letsencrypt/venv/bin/letsencrypt", line 9, in <module>
  File "/root/letsencrypt/letsencrypt/cli.py", line 1002, in main
  File "/root/letsencrypt/letsencrypt/cli.py", line 376, in auth
    le_client = _init_le_client(args, config, authenticator, installer)
  File "/root/letsencrypt/letsencrypt/cli.py", line 160, in _init_le_client
    acc, acme = _determine_account(args, config)
  File "/root/letsencrypt/letsencrypt/cli.py", line 147, in _determine_account
    config, account_storage, tos_cb=_tos_cb)
  File "/root/letsencrypt/letsencrypt/client.py", line 98, in register
    acc = account.Account(regr, key)
  File "/root/letsencrypt/letsencrypt/account.py", line 60, in __init__
    format=serialization.PublicFormat.SubjectPublicKeyInfo)
  File "/root/letsencrypt/venv/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 614, in public_bytes
    self._rsa_cdata
  File "/root/letsencrypt/venv/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1199, in _public_key_bytes
    raise ValueError("Only PEM encoding is supported by this backend")
ValueError: Only PEM encoding is supported by this backend
```

kuba:
This is a breaking change, already existing /etc/letsencrypt/accounts will cause the client raise `errors.AccountStorageError` ("Account ids mismatch ..."). Simple fix: start over by `rm -rf /etc/letsencrypt`. In a long run it's a better solution that bumping to `cryptography>=0.9`...

@pde, this is one of the examples that shows problem in sharing `/etc/letsencrypt` between our package and e.g. Debian packaging. Even if we had a migration script, if user went back to `letsencrypt` provided by Debian, they would have a really hard time.

Fixes #1012

### DIFF
--- a/letsencrypt/account.py
+++ b/letsencrypt/account.py
@@ -56,7 +56,7 @@ class Account(object):  # pylint: disable=too-few-public-methods
 
         self.id = hashlib.md5(
             self.key.key.public_key().public_bytes(
-                encoding=serialization.Encoding.DER,
+                encoding=serialization.Encoding.PEM,
                 format=serialization.PublicFormat.SubjectPublicKeyInfo)
         ).hexdigest()
         # Implementation note: Email? Multiple accounts can have the
--- a/letsencrypt/tests/account_test.py
+++ b/letsencrypt/tests/account_test.py
@@ -45,16 +45,16 @@ class AccountTest(unittest.TestCase):
 
     def test_id(self):
         self.assertEqual(
-            self.acc.id, "2ba35a3bdf380ed76a5ac9e740568395")
+            self.acc.id, "bca5889f66457d5b62fbba7b25f9ab6f")
 
     def test_slug(self):
         self.assertEqual(
-            self.acc.slug, "test.letsencrypt.org@2015-07-04T14:04:10Z (2ba3)")
+            self.acc.slug, "test.letsencrypt.org@2015-07-04T14:04:10Z (bca5)")
 
     def test_repr(self):
         self.assertEqual(
             repr(self.acc),
-            "<Account(2ba35a3bdf380ed76a5ac9e740568395)>")
+            "<Account(bca5889f66457d5b62fbba7b25f9ab6f)>")
 
 
 class ReportNewAccountTest(unittest.TestCase):


kuba:
Build affected by #1010.

bmw:
Why is this a better solution than bumping to `cryptography>=0.9`? The bump seems like a simpler fix and doesn't require an account wipe for people that already received certs from production.

kuba:
Because that makes our dependencies tighter, and hence the entire project is likely harder to package (some systems might not have `cryptography==0.9` packaged, but they might do have `cryptography==0.8`). If we can change our code to relax deps, then that's better for us. And we are still before launch so impact on users is low...

pde:
Because we have a growing userbase, including some people with soon-to-be-real certs, we should no longer land changes that require an account wipe.

Before it can land, this PR needs a DER-to-PEM converter added to it, so that even with 0.8 installed the client uses the current on-disk format.  It's fine if that converter is implemented in a hackish manner (Popen `openssl x509 -inform PEM -outform DER` for instance), since the entire purpose of this is compatibility during the period when OSes have a too-old version of python-cryptography.

kuba:
Correction after review of the code: customers will not see any errors and wipe is not really necessary - the only thing that changes is more DEBUG logs: https://github.com/letsencrypt/letsencrypt/blob/3885a40c421636805a77a0f1c573d1727169eddb/letsencrypt/account.py#L162. Client will happily create new account with new id. Problems might happen if using `--account ID`. FYI certificates are not bound to a particular account on the client side.

kuba:
Another correction: `letsencrypt-renewer` does tie account to a lineage.

pde:
OK, we can probably live with those debug logs and some extra-account-creation amongst the current userbase.  But we should test that before shipping a release.

jsha:
I think bumping the cryptogrpahy dep to 0.9 is a good short-term fix, and we can re-add cryptography 0.8 support later if we need.

Note that if you want to calculate the SPKI hash without depending on cryptogrpahy 0.9 you can do it by shelling out to openssl: https://raymii.org/s/articles/HTTP_Public_Key_Pinning_Extension_HPKP.html

``
openssl x509 -noout -in certificate.pem -pubkey | \
openssl asn1parse -noout -inform pem -out public.key;
openssl dgst -sha256 -hex public.key
```

jsha:
Another way to maintain compatibility would be to take the PEM output and run it through base64 decode to get the DER, then hash that. I really don't like the idea of moving to a key id method that depends on hashing the PEM.

That said, if we're going to break compatibility, can we also switch the hash to something other than md5?

kuba:
I mostly thinking about packaging now. E.g. `jessie-backports` has only `0.8`. FreeBSD packagers also had issues https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=203405#c17. Shelling out will cause us even more troubles.

kuba:
Point about `md5` is that it is short (it's a filename, accessible by users); collisions are very unlikely to happen anyway and we don't need any other crypto property for it.

pde:
Hmmm, what about:

https://docs.python.org/2/library/ssl.html#ssl.PEM_cert_to_DER_cert

pde:
Will that work with a key, or not?

pde:
(Seems the answer is no)

pde:
Overall: @jsha I think you have a somewhat reasonable point about hashing PEM, but I don't think it's a good use of our time today to engineer something else :/.

jsha:
Alright, given that we can write code to migrate people to whatever directory layout we want in the future, I'm fine with landing this as a quick fix. Another thought, perhaps for future tweaks: why not use the registration id, or even a locally incrementing number, as the name for this directory?

pde:
OK, landing.

PR #1020, opened 2015-10-17
Report expiration when using a CSR

bmw:
As noted in #983, cert expiration was not reported when using a user specified CSR. This fixes that.


kuba:
Affected by #1021.

pde:
LGTM.

PR #1025, opened 2015-10-19
Adjusted Docker user guide link

schumaml:
The current URL leads to a 404 error


bmw:
Thanks for the patch @schumaml! LGTM

PR #1034, opened 2015-10-20
Updated TODOs

### LINKED ISSUE #1033: Possibility of infinite loop?
It seems unlikely to be a problem, but this [comment](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/auth_handler.py#L156) suggests the possibility of an infinite loop while solving challenges. I haven't had the time to look into the issue personally, but if this is true, it should be fixed.

bmw:
* Fixes #1033. Turns out that `XXX` comment was outdated. All challenges are properly removed in `_cleanup_challenges`.
* The full chain is now included in `deploy_cert` so that comment can also be removed.
* The comment in `renewer.py` was incorrect. By passing the key to `save_successor`, the code actually assumes there was a key change. This behavior is fine.
* Assigning to `rc_config.filename` is not a dirty hack. It's [documented](https://configobj.readthedocs.org/en/latest/configobj.html#writing-a-config-file) functionality of a `configobj.ConfigObj`.

### DIFF
--- a/letsencrypt/auth_handler.py
+++ b/letsencrypt/auth_handler.py
@@ -153,10 +153,8 @@ class AuthHandler(object):
         """
         active_achalls = []
         for achall, resp in itertools.izip(achalls, resps):
-            # XXX: make sure that all achalls, including those
-            # corresponding to None or False returned from
-            # Authenticator are removed from the queue and thus avoid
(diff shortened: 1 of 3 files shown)


pde:
Looks good!

PR #1038, opened 2015-10-20
Implement --standalone

### LINKED ISSUE #1036: Broken letsencrypt --standalone
```
(venv)root@le:~/letsencrypt# git rev-parse HEAD
cb2ac71d1ee48205e6756fa3264551ae24b6eae2
(venv)root@le:~/letsencrypt# letsencrypt --help

the cert.  Major SUBCOMMANDS are:

  (default) everything Obtain & install a cert in your current webserver
  auth                 Authenticate & obtain cert, but do not install it

Choice of server for authentication/installation:

  --apache          Use the Apache plugin for authentication & installation
  --nginx           Use the Nginx plugin for authentication & installation
  --standalone      Run a standalone HTTPS server (for authentication only)
  OR:
  --authenticator standalone --installer nginx

More detailed help:

  -h, --help [topic]    print this message, or detailed help on a topic;
                        the available topics are:

   all, apache, automation, manual, nginx, paths, security, testing, or any of
   the subcommands

(venv)root@le:~/letsencrypt# letsencrypt --standalone
usage: 

the cert.  
letsencrypt: error: argument --standalone-supported-challenges: Unrecognized challenges: run
```

pde:
Closes #1036

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -72,7 +72,7 @@ Choice of server for authentication/installation:
   --apache          Use the Apache plugin for authentication & installation
   --nginx           Use the Nginx plugin for authentication & installation
   --standalone      Run a standalone webserver for authentication
-    OR:
+  OR:
   --authenticator standalone --installer nginx
 
 More detailed help:
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -60,7 +60,7 @@ the cert.  """
 # altogether
 USAGE = SHORT_USAGE + """Major SUBCOMMANDS are:
 
-  (default) everything Obtain & install a cert in your current webserver
+  (default) run        Obtain & install a cert in your current webserver
   auth                 Authenticate & obtain cert, but do not install it
   install              Install a previously obtained cert in a server
   revoke               Revoke a previously obtained certificate
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -750,7 +750,7 @@ def create_parser(plugins, args):
                 help="Obtain and install certs using Apache")
     helpful.add(None, "--nginx", action="store_true",
                 help="Obtain and install certs using Nginx")
-    helpful.add(None, "--standalone`", action="store_true",
+    helpful.add(None, "--standalone", action="store_true",
                 help='Obtain certs using a "standalone" webserver on port 443.')
     # positional arg shadows --domains, instead of appending, and
     # --domains is useful, because it can be stored in config
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -923,7 +923,7 @@ def _plugins_parsing(helpful, plugins):
     helpful.add("plugins", "--nginx", action="store_true",
                 help="Obtain and install certs using Nginx")
     helpful.add("plugins", "--standalone", action="store_true",
-                help="Obtain certs using a "standalone" webserver.")
+                help='Obtain certs using a "standalone" webserver.')
 
     # things should not be reorder past/pre this comment:
     # plugins_group should be displayed in --help before plugin
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -382,6 +382,8 @@ def choose_configurator_plugins(args, config, plugins, verb):
     if args.apache:
         req_inst = set_configurator(req_inst, "apache")
         req_auth = set_configurator(req_auth, "apache")
+    if args.standalone:
+        req_auth = set_configurator(req_auth, "standalone")
     logger.debug("Requested authenticator %s and installer %s", req_auth, req_inst)
 
     # Try to meet the user's request and/or ask them to pick plugins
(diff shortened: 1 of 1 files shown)


letsencrypt/cli.py
```diff
@@ -744,6 +750,8 @@ def create_parser(plugins, args):
                 help="Obtain and install certs using Apache")
     helpful.add(None, "--nginx", action="store_true",
                 help="Obtain and install certs using Nginx")
+    helpful.add(None, "--standalone", action="store_true",
+                help='Obtain certs using a "standalone" webserver on port 443.')
```
bmw: Nit: Standalone now supports both SimpleHTTP and DVSNI. Perhaps just change this to `Obtain certs using a "standalone" webserver.`?

letsencrypt/cli.py
```diff
@@ -636,8 +638,12 @@ def preprocess_args(self, args):
                 self.verb = token
                 return reordered
 
-        self.verb = "run"
-        return args + ["run"]
+        if "--standalone" in args and "--installer" not in args and "-i" not in args:
```
bmw: I don't think we should be changing the verb based on the plugin flags given. `run` is the default except for when it's not? That's confusing.

I think the behavior without these 6 lines of code is acceptable. If you have an installer available, it'll be used with `standalone` in run (this actually works if your server isn't currently running). If you don't, you'll get one of the messages like "No installers are available on your OS yet; try running "letsencrypt-auto auth" to get a cert you can install manually" which I think is perfectly acceptable and even preferred behavior.

pde: How about if the docs explain that the default verb becomes "auth"?

bmw: So we make an exception for `--standalone`, are there going to be other exceptions in the future?

I just don't like the added complexity and prefer the approach described above. `run` is the default, full stop. If you run `letsencrypt --standalone`, we try to install if there is an installer available. If another webserver is running it won't work but as you pointed out to me in another PR, this is true regardless of whether `auth` or `run` is used. If, however, there is no installer available we error out quickly, explaining the problem and pointing them to the `auth` command. This behavior is fairly self documenting and I think will be more easily understood by our users than your approach.

We made the decision to make `run` the default and I think we should stick to that decision regardless of the flags given on the command line.

pde: I think that `letsencrypt --standalone` is a pretty simple command that should just do what it seems like it should do.  I don't think there's a conceptual problem with the internal semantics of a CLI tool being slightly different when it receives a flag, if that's what makes logical sense.

That's especially true when we're talking about the *default* selection of subcommands, because the user will often not have much of a model of what the subcommands exactly are and what they exactly do.  And if they do have that model, the help is [very clear](https://github.com/letsencrypt/letsencrypt/pull/1038/files#diff-54b2f468d57ef7436008bd031bc89711R74).

pde: You could probably make the argument that we should remove my logic which keeps the verb as `run` if `-i INSTALLER` is supplied.  I'm fine either way there...

bmw: It is my opinion that this is a bad decision. How much gain does the user really get here? The three cases I see are:

1. No installers are available. This is the most likely scenario given the current state of installers. This change saves them a quickly failed run telling them to run `auth`.
2. An installer is available and a webserver is running on the desired port. With or without this change, they get a message saying the port is bound.
3. An installer is available and the port is free. Without this change, they get the full functionality provided by Let's Encrypt of automatically installing the cert and configuring their installation. With this change, the user gets reduced functionality.

So the only benefit I see in case 1, which is hardly worth discussing in my opinion.

Also, what about `-a standalone`and `--authenticator standalone`? Why don't those have special behavior?

kuba: I strongly agree with @bmw.

bmw: If you'd like to make another **exception** and rush this PR without tests to get it into PyPI, I'm fine in doing so but I want this functionality removed and the help for `--standalone` updated. If you really feel strongly about it, you can open another PR about the issue later. Not having this functionality is not a showstopper.

pde: @bmw I think the version I implemented is a bit more elegant, but I don't feel very strongly about it.  Will pull out the default-changing logic.

letsencrypt/cli.py
```diff
@@ -71,7 +71,8 @@
 
   --apache          Use the Apache plugin for authentication & installation
   --nginx           Use the Nginx plugin for authentication & installation
-  --standalone      Run a standalone HTTPS server (for authentication only)
+  --standalone      Run a standalone webserver for authentication
+
```
bmw: Nittiest nit: I don't like this space. It breaks up the block of plugin options.

letsencrypt/cli.py
```diff
@@ -921,6 +919,15 @@ def _plugins_parsing(helpful, plugins):
         "plugins", "--configurator", help="Name of the plugin that is "
         "both an authenticator and an installer. Should not be used "
         "together with --authenticator or --installer.")
+    helpful.add("plugins", "--apache", action="store_true",
+                help="Obtain and install certs using Apache")
+    helpful.add("plugins", "--nginx", action="store_true",
+                help="Obtain and install certs using Nginx")
+    helpful.add("plugins", "--standalone", action="store_true",
+                help=('Obtain certs using a "standalone" webserver. '
+                      'Changes the default "run" command to "auth".'))
```
bmw: Help still needs to be updated.

letsencrypt/cli.py
```diff
@@ -921,6 +919,15 @@ def _plugins_parsing(helpful, plugins):
...
+
+
```
bmw: Another nitty nit: Whitespace is excessive.

bmw:
LGTM. Merging when Travis is happy.

PR #1039, opened 2015-10-20
Fix docker run command in using.rst

gionn:


kuba:
LGTM

That's another regression introduced by #980

pde:
Someone (@bmw ?) seemed pretty sure that the `run` verb wouldn't work inside docker because it wouldn't be able to see and reconfigure systemwide webservers.

kuba:
You are mistaking "run" (earlier in the command) passed to Docker, with "run" (later in the command) passed to our binary. Configuration that doesn't work because it doesn't make sense for client to configure webserver inside Docker: (run, run). Configuration that doesn't work because i'ts not a valid Docker command: (auth, *).

kuba:
@pde, btw should you wait for a review before merging, this wouldn't be a problem!

bmw:
LGTM. Thanks @gionn!

PR #1040, opened 2015-10-20
Do not reinstall what's already installed

WebReflection:
There's no need to reinstall all packages. Few of  them are quite heavy (gcc) and most  of the time, when it comes to developers, already available. You need git, for instance, to clone the repository itself, that's going to be there and configured already too.

Please consider this change so that bootstrapping is going to be even faster. Best Regards

### DIFF
--- a/bootstrap/archlinux.sh
+++ b/bootstrap/archlinux.sh
@@ -3,7 +3,7 @@
 # "python-virtualenv" is Python3, but "python2-virtualenv" provides
 # only "virtualenv2" binary, not "virtualenv" necessary in
 # ./bootstrap/dev/_common_venv.sh
-pacman -S \
+pacman -S --needed \
   git \
   python2 \
   python-virtualenv \


pde:
Once operating systems package let's encrypt, you won't need to reinstall anything.  For the time being, we use virtualenv and pip to manage dependencies, which does imply a lot of reinstalling, but also ensures that you have a sufficiently recent version of everything the client needs.

kuba:
That seemed like a valid PR - it just made sure Arch packages are not reinstalled over and over again.... Why did you close it, @pde?!

WebReflection:
Exactly. That doesn't mean version  conflicts, it means if already there and it satisfies the dependency don't reinstall the package. Very odd way to handle this PR

WebReflection:
Following the description of the option I've added via `man pacman`

screenshot from 2015-10-20 10-34-10

It installs software only if not up to date so please re-consider this PR, thanks.

pde:
Sorry @WebReflection!  I misread this as an issue rather than a PR...

PR #1041, opened 2015-10-20
Added needed flag to pacman invocation to only install packages that …

Faerbit:
…are not already installed.

This makes for a better user experience on Arch.


kuba:
"Dupe" of #1040. I just hope this one won't be trashed as its predecessor...

LGTM

cc @WebReflection

WebReflection:
yeah, apparently @pde didn't know about this flag so instead of asking more he closed the PR without any valid reason.

I'll update the meaning of that here too so that after two PRs maybe we'll have it in.

what does that flag mean

bmw:
We certainly want this. LGTM.

Sorry about #1040 @WebReflection :(

WebReflection:
@bmw you've done the right thing and there's nothing *you* should personally be sorry about.  Thanks for merging it.

PR #1049, opened 2015-10-20
Small logging improvements (fixes #1037)

### LINKED ISSUE #1037: Store version string in logs
It will help us in debugging, in particular `letsencrypt-auto` users that run versioned PyPI releases. That's probably good candidate for `logging.INFO` level.

kuba:


kuba:
Forgot to tag it in commits, but it fixes https://github.com/letsencrypt/letsencrypt/issues/1037.

bmw:
LGTM. Merging when Travis has a prettier color.

acme/acme/standalone.py
```diff
@@ -55,6 +55,7 @@ def __init__(self):
 
     def serve_forever2(self):
         """Serve forever, until other thread calls `shutdown2`."""
+        logger.debug("Starting server at %s:%d...", *self.socket.getsockname())
```
pde: Will this work if the return value is an IPv6 4-tupple?

kuba: It will not work. Good catch!

kuba:
Fixed `getsockname` and Travis is now green...

bmw:
Green is a prettier color. LGTM.

PR #1058, opened 2015-10-21
Simple fix to hanging standalone

bmw:
I was able to recreate this problem a couple times locally today, but I wasn't able to track down the interleaving of events that caused the issue. With that said, I believe this is a simple fix to the problem.

In `ACMEServerMixin`, [self.timeout](https://docs.python.org/2/library/socketserver.html#SocketServer.BaseServer.timeout) gives us away to have `handle_request` return after `self.timeout` seconds if it has not received a request. If `shutdown2` has not been called, `handle_request` will be called again, but this allows the code to break out of the `handle_request` function and to check the value of `self._stopped`.

@kuba, I defer to you here as you know more about this code than I do, but I believe this is a trivial fix to the problem.

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -64,7 +64,7 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
         """Shutdown server loop from `serve_forever2`."""
         self._stopped = True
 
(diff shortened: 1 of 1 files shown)


bmw:
Real fix is more complicated. Closing.

Issue #960, opened 2015-10-12
100% test coverage for letsencrypt module


bmw:
The `acme`, `letsencrypt-apache`, and `letshelp-letsencrypt` modules all have 100% test coverage. We should strive towards bringing the `letsencrypt` module towards 100% coverage as well.

bmw:
We probably care about this less once we're using `mypy` (see #4244).

PR #1059, opened 2015-10-21
Add  a test, skip a test

mrwacky42:
Partial fix for #960.

### DIFF
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -130,10 +130,7 @@ binary for temporary key/certificate generation.""".replace("\n", "")
         if self.conf("test-mode"):
             logger.debug("Test mode. Executing the manual command: %s", command)
             # sh shipped with OS X does't support echo -n
-            if sys.platform == "darwin":
-                executable = "/bin/bash"
-            else:
-                executable = None
+            executable = "/bin/bash" if sys.platform == "darwin" else None
             try:
                 self._httpd = subprocess.Popen(
                     command,
--- a/letsencrypt/plugins/util_test.py
+++ b/letsencrypt/plugins/util_test.py
@@ -98,6 +98,10 @@ class AlreadyListeningTest(unittest.TestCase):
         self.assertEqual(mock_get_utility.call_count, 1)
         mock_process.assert_called_once_with(4420)
 
+    @mock.patch("letsencrypt.plugins.util.psutil.net_connections")
+    def test_access_denied_exception(self, mock_net):
+        mock_net.side_effect = psutil.AccessDenied("")
+        self.assertFalse(self._call(12345))
 
 if __name__ == "__main__":
     unittest.main()  # pragma: no cover


letsencrypt/plugins/manual.py
```diff
@@ -130,7 +130,7 @@ def _perform_single(self, achall):
         if self.conf("test-mode"):
             logger.debug("Test mode. Executing the manual command: %s", command)
             # sh shipped with OS X does't support echo -n
-            if sys.platform == "darwin":
+            if sys.platform == "darwin":   # pragma: no cover
                 executable = "/bin/bash"
             else:
                 executable = None
```
kuba: could we use `executable = "/bin/bash" if sys.platform == "darwin" else None` instead of `pragma: no cover`? Expression fits on one line, and we're sure darwin branch doesn't have a terrible typo.

letsencrypt/plugins/util_test.py
```diff
@@ -98,6 +98,11 @@ def test_listening_ipv6(self, mock_get_utility, mock_process, mock_net):
         self.assertEqual(mock_get_utility.call_count, 1)
         mock_process.assert_called_once_with(4420)
 
+    @mock.patch("letsencrypt.plugins.util.psutil.net_connections")
+    def test_access_denied_exception(self, mock_net):
+        mock_net.return_value = False
```
kuba: I don't think this line is necessary, `side_effect` should be enough

letsencrypt/plugins/util_test.py
```diff
@@ -98,6 +98,11 @@ def test_listening_ipv6(self, mock_get_utility, mock_process, mock_net):
         self.assertEqual(mock_get_utility.call_count, 1)
         mock_process.assert_called_once_with(4420)
 
+    @mock.patch("letsencrypt.plugins.util.psutil.net_connections")
+    def test_access_denied_exception(self, mock_net):
```
kuba: that's not useless, that's great test, thanks!

kuba:
Welcome to the project, and thanks for your contributions! :)

LGTM, but I'd be much happier if 2 above nits were fixed (pre or post merge).

letsencrypt/plugins/manual.py
mrwacky42: Doesn't that just feel like cheating the code coverage detector?
Better would probably be to copy a test somewhere and mock `sys.platform`, I suppose.

kuba: It is cheating, indeed, but comment about typo is still true: Darwin branch could say `exceutable = ...` and we could not notice the typo with this setup. Mocking `sys.platform` is ideal - do you have time to implement it? :)

letsencrypt/plugins/util_test.py
mrwacky42: Eh well, the exception logs and returns False, so I figured I'd cover all the bases.

kuba: I'm not sure what you mean. If the side effect raises exception, return will never be reached. So whatever you set to `mock_net.return_value` doesn't matter at all.

mrwacky42:
Rebased.
I couldn't satisfactorily mock sys.platform, so I went with your original suggestion.

kuba:
LGTM, thanks for fixes! This will have to be merged by someone else than me.

For future: we don't have a policy of rebased PRs only. If you find it too troublesome, you are just free to merge with `master`. Likewise, you can add new commits with fixes on top of existing PR fixes. In fact this strategy works better with GitHub, where comments can easily get lost when rebasing. IMHO messing with git history causes too much pain to by used in any circumstances.

mrwacky42:
Yeah, I figured for such a small change it would be nice to have less history associated with it.

PR #1070, opened 2015-10-21
Make debian bootstrap script more agnostic about package names

### LINKED ISSUE #1061: Debian 7, build with letsencrypt-auto fails -- virtualenv not found
Here is the log:

```
0 upgraded, 0 newly installed, 0 to remove and 228 not upgraded.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libaugeas0 is already the newest version.
ca-certificates is already the newest version.
dialog is already the newest version.
gcc is already the newest version.
git-core is already the newest version.
libffi-dev is already the newest version.
libssl-dev is already the newest version.
python is already the newest version.
python-dev is already the newest version.
python-virtualenv is already the newest version.
The following package was automatically installed and is no longer required:
  libgssglue1
Use 'apt-get autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 228 not upgraded.
./letsencrypt-auto: 63: ./letsencrypt-auto: virtualenv: not found
```

Also on annoying side -- it does not need to attempt to reinstall packages if they are already present in system...

pde:
Should fix #1061

### DIFF
--- a/bootstrap/_deb_common.sh
+++ b/bootstrap/_deb_common.sh
@@ -1,16 +1,22 @@
 #!/bin/sh
 
-# Tested with:
-#   - Ubuntu:
-#     - 12.04 (x64, Travis)
-#     - 14.04 (x64, Vagrant)
-#     - 14.10 (x64)
-#   - Debian:
-#     - 6.0.10 "squeeze" (x64)
-#     - 7.8 "wheezy" (x64)
-#     - 8.0 "jessie" (x64)
-#   - Raspbian:
-#     - 7.8 (armhf)
+# Current version tested with:
+#
+# - Ubuntu
+#     - 14.04 (x64)
+#     - 15.04 (x64)
+# - Debian
+#     - 7.9 "wheezy" (x64)
+#     - sid (2015-10-21) (x64)
+
+# Past versions tested with:
+#
+# - Debian 8.0 "jessie" (x64)
+# - Raspbian 7.8 (armhf)
+
+# Believed not to work:
+#
+# - Debian 6.0.10 "squeeze" (x64)
 
 apt-get update
 
--- a/bootstrap/_deb_common.sh
+++ b/bootstrap/_deb_common.sh
@@ -12,46 +12,33 @@
 #   - Raspbian:
 #     - 7.8 (armhf)
 
+apt-get update
 
 # virtualenv binary can be found in different packages depending on
 # distro version (#346)
-newer () {
-  apt-get install -y lsb-release --no-install-recommends
-  distro=$(lsb_release -si)
-  # 6.0.10 => 60, 14.04 => 1404
-  # TODO: in sid version==unstable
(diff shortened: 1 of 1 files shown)


pde:
Tested on:
* debian sid
* debian 7.9
* ubuntu 15.04
* ubuntu 14.04

bmw:
Tested successfully on Debian stable. I also tested on Debian squeeze which did not work, but the current bootstrap script didn't work for me either :frowning:

@pde made the argument to me that the current version of this script makes a number of assumptions about `lsb_release`. These assumptions likely hold true when using the official Debian and Ubuntu repositories. If the user is using a custom repositories list, then all bets are off regarding the `lsb_release` package.

In my mind, the script with this PR is simpler and makes less assumptions about the users system than the current version as I mentioned above. The only potential downside I see to this script is installing both `virtualenv` and `python-virtualenv` may not be necessary on some systems and we're installing unneeded packages. I feel that this is a fairly minor point, however.

This LGTM with the exception that we should probably update the list of supported distros at the top of the script. Debian squeeze did not work for me for either script and this has not been tested against Raspbian.

bmw:
[Log](https://gist.github.com/ac4f1364062b0f4af8ef) of failure with current bootstrap script (without this PR). Tested on DO instance

pde:
Brad that issue is now #1078.

bmw:
@kuba, another hanging standalone [log](https://gist.github.com/46a6c009f3784216a2bd).

bootstrap/_deb_common.sh
```diff
@@ -1,57 +1,50 @@
...
-  then
-    return 0;
-  elif [ "$distro" = "Debian" -a "$version" -ge 80 ]
-  then
-    return 0;
-  else
-    return 1;
-  fi
-}
+# Believed not to work:
```
kuba: That's not true.

bootstrap/_deb_common.sh
```diff
@@ -1,57 +1,50 @@
...
+# Current version tested with:
+#
+# - Ubuntu
+#     - 14.04 (x64)
+#     - 15.04 (x64)
+# - Debian
+#     - 7.9 "wheezy" (x64)
+#     - sid (2015-10-21) (x64)
 
+# Past versions tested with:
```
kuba: That's why I didn't like this change. Because it simply wasn't tested.

bootstrap/_deb_common.sh
```diff
@@ -1,57 +1,50 @@
...
+  $virtualenv \
   gcc \
   dialog \
   libaugeas0 \
   libssl-dev \
   libffi-dev \
   ca-certificates \
+
+if ! which virtualenv > /dev/null ; then
+  echo Failed to install a working \"virtualenv\" command, exiting
```
kuba: That shouldn't be part of this script.

Issue #1072, opened 2015-10-21
letsencrypt-auto / pip versioning errors if you have ever installed the nginx plugin


pde:
I've seen this on two systems that were used for pre-release dev work:

```
pde@damoclid:~/letsencrypt/letsencrypt$ ./letsencrypt-auto --help
Updating letsencrypt and virtual environment dependencies......
Running with virtualenv: sudo /home/pde/.local/share/letsencrypt/bin/letsencrypt --help
An unexpected error occurred.
VersionConflict: (acme 0.0.0.dev20151020 (/home/pde/.local/share/letsencrypt/lib/python2.7/site-packages), Requirement.parse('acme==0.0.0.dev20151008'))
Please see the logfile 'letsencrypt.log' for more details.
```

```
pde@damoclid:~/letsencrypt/letsencrypt$ cat letsencrypt.log 
Traceback (most recent call last):
  File "/home/pde/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/cli.py", line 1030, in main
    plugins = plugins_disco.PluginsRegistry.find_all()
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/plugins/disco.py", line 161, in find_all
    plugin_ep = PluginEntryPoint(entry_point)
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/plugins/disco.py", line 31, in __init__
    self.plugin_cls = entry_point.load()
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2354, in load
    self.require(*args, **kwargs)
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2371, in require
    items = working_set.resolve(reqs, env, installer)
  File "/home/pde/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 844, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
VersionConflict: (acme 0.0.0.dev20151020 (/home/pde/.local/share/letsencrypt/lib/python2.7/site-packages), Requirement.parse('acme==0.0.0.dev20151008'))

```
[...]

PR #1073, opened 2015-10-21
Fix pip / nginx heisenbug

pde:
Code to install letsencrypt-nginx was still active, but only in verbose mode.

Closes: #1072

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -69,8 +69,12 @@ if [ "$VERBOSE" = 1 ]  ; then
   echo
   $VENV_BIN/pip install -U setuptools
   $VENV_BIN/pip install -U pip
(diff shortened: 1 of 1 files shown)


kuba:
Please start using functions to make the code more modular, deduplicated and thus easier to review.

pde:
Agreed in principle, but that will be a future PR.

bmw:
Yep. I agree functions are nice, especially when we are telling users to run a script with `sudo` privileges that automagically installs, updates, and runs `letsencrypt` on their system. I bet many users will read the script (I know I would) and functions make it easier for them to do so.

With that said, this PR works. I tested the code without hiccups.

bmw:
Merging. Talked to @pde out of band about making the code easier to read.

Issue #1074, opened 2015-10-21
Point users at fullchain.pem rather than cert.pem?


pde:
Users who run standalone or manual get a message pointing them at a cert.pem file.  But actually, most users running modern webservers want fullchain.pem.

bmw:
A fairly large exception here is Apache (at least in the way we are configuring it). It uses `cert.pem`, `privkey.pem`, and `chain.pem`, but not `fullchain.pem`.

PR #1077, opened 2015-10-21
Point the user to fullchain.pem, not cert.pem

pde:
Closes: #1074


pde:
I'm not very happy about this PR; I started by trying to tweak an output message, and then tugging threads lead me to the fact that save_certificate and storage.py are totally different paths.

All of that should be refactored, but not for 1.0.

letsencrypt/cli.py
```diff
@@ -267,16 +267,30 @@ def _treat_as_renewal(config, domains):
     return None
 
 
-def _report_new_cert(cert_path):
-    """Reports the creation of a new certificate to the user."""
+def _report_new_cert(cert_path, fullchain_path):
+    """
+    Reports the creation of a new certificate to the user.
+    :param string cert_path: path to cert
```
bmw: Need a blank line before `:param ...` or Sphinx gets confused.

letsencrypt/client.py
```diff
@@ -258,8 +258,8 @@ def obtain_and_enroll_certificate(self, domains, plugins):
             params, config, cli_config)
         return lineage
 
-    def save_certificate(self, certr, chain_cert, cert_path, chain_path):
-        # pylint: disable=no-self-use
+    def save_certificate(self, certr, chain_cert, cert_path, chain_path, fullchain_path):
```
bmw: Long line. Consider breaking.

letsencrypt/client.py
```diff
@@ -268,20 +268,22 @@ def save_certificate(self, certr, chain_cert, cert_path, chain_path):
         :param list chain_cert:
         :param str cert_path: Candidate path to a certificate.
         :param str chain_path: Candidate path to a certificate chain.
+        :param str fullchain_path: Candidate path to a full cert chain.
 
-        :returns: cert_path, chain_path (absolute paths to the actual files)
+        :returns: cert_path, chain_path, fullchain_path (absolute paths to the actual files)
```
bmw: Long line. You could break or remove/shorten the comment about paths being absolute.

letsencrypt/client.py
```diff
@@ -293,9 +295,11 @@ def save_certificate(self, certr, chain_cert, cert_path, chain_path):
         logger.info("Server issued certificate; certificate written to %s",
                     act_cert_path)
 
+        # TODO too long, refactor... either split this into another function
```
bmw: Yep. Should be refactored. I hate leaving around TODOs because they get neglected. What about something like:
```
def _save_chain(chain_pem, chain_path):
    """Saves chain_pem at a unique path based on chain path.

    :param str chain_pem: certificate chain in PEM format
    :param str chain_path: candidate path for the cert chain 

    :returns: absolute path to saved cert chain
    :rtype: str

    """
    chain_file, act_chain_path = le_util.unique_file(chain_path, 0o644)
    try:
        chain_file.write(chain_pem)
    finally:
        chain_file.close()

    logger.info("Cert chain written to %s", act_chain_path)

    # This expects a valid chain file
    return os.path.abspath(act_chain_path)
```

Then update save_certificate to:
```
    ...

    if chain_cert:
        chain_pem = crypto_util.dump_pyopenssl_chain(chain_cert)
        cert_chain_abspath = _save_chain(chain_pem, chain_path)
        fullchain_abspath = _save_chain(cert_pem + chain_pem, fullchain_path)

    ...

    return os.path.abspath(act_cert_path), cert_chain_abspath, fullchain_abspath
```

Note: This code is not tested

bmw: This also might allow you to remove the `too-many-locals` pylint message.

letsencrypt/cli.py
bmw: And nit: I think `:param string` should be `:param str` to keep with the rest of the codebase.

bmw:
Yep. It's a bit of an ugly situation. Support for user supplied CSRs was hacked into `auth` and now relies on old code that isn't used anywhere else.

If we could unify `storage.py` and `save_certificate`, it'd be forward progress on removing differences between `run` vs `auth` and `install` as well. This is a pure refactor though and isn't necessary for launch.

pde:
LGTM, field tested on one ubuntu VPS, merging.

Issue #664, opened 2015-08-07
"Standalone" needs to be default authenticator in place of "manual" and both options need explanations


yaseminacar:
1.)  To enable users to just press enter for the easiest option, 'standard' needs to be default (pre-selected) instead of manual.
2.) The options right now are "Standard(standard)" and "Manual(manual)". Either the brackets need to go or a meaningful explanation needs to be added (the latter would be better).

jdkasten:
*typo standard -> Standalone

bmw:
I agree we need to a better job of differentiating `manual` from `standalone`.

pde:
Giving this to @swartzcr.

bmw:
Closed in #1079.

PR #1079, opened 2015-10-22
Fixed nits with plugin choice

SwartzCr:
Changed order of plugins listed - they'll now be reverse alphabetical
Made the descriptions of plugins not mimic their names
Made more_info for standalone work the same way as manual
Made more_info strings not have missing spaces due to stripped \n formatting by adding leading spaces

### DIFF
--- a/letsencrypt/display/ops.py
+++ b/letsencrypt/display/ops.py
@@ -74,7 +74,7 @@ def pick_plugin(config, default, plugins, question, ifaces):
 
     if len(prepared) > 1:
         logger.debug("Multiple candidate plugins: %s", prepared)
-        plugin_ep = choose_plugin(prepared.values(), question)
+        plugin_ep = choose_plugin(prepared.values()[::-1], question)
         if plugin_ep is None:
             return None
         else:
--- a/letsencrypt/display/ops.py
+++ b/letsencrypt/display/ops.py
@@ -74,7 +74,7 @@ def pick_plugin(config, default, plugins, question, ifaces):
 
     if len(prepared) > 1:
         logger.debug("Multiple candidate plugins: %s", prepared)
-        plugin_ep = choose_plugin(prepared.values()[::-1], question)
+        plugin_ep = choose_plugin(prepared.values(), question)
         if plugin_ep is None:
             return None
         else:
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -32,7 +32,7 @@ class Authenticator(common.Plugin):
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
 
-    description = "Manual Authenticator"
+    description = "Manually Edit Your Configuration"
 
     MESSAGE_TEMPLATE = """\
 Make sure your web server displays the following content at
@@ -82,10 +82,10 @@ s.serve_forever()" """
     def more_info(self):  # pylint: disable=missing-docstring,no-self-use
         return """\
 This plugin requires user's manual intervention in setting up a HTTP
-server for solving SimpleHTTP challenges and thus does not need to be
-run as a privilidged process. Alternatively shows instructions on how
-to use Python's built-in HTTP server and, in case of HTTPS, openssl
-binary for temporary key/certificate generation.""".replace("\n", "")
+ server for solving SimpleHTTP challenges and thus does not need to be
+ run as a privilidged process. Alternatively shows instructions on how
+ to use Python's built-in HTTP server and, in case of HTTPS, openssl
+ binary for temporary key/certificate generation.""".replace("\n", "")
 
     def get_chall_pref(self, domain):
         # pylint: disable=missing-docstring,no-self-use,unused-argument
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -26,17 +26,13 @@ logger = logging.getLogger(__name__)
 class Authenticator(common.Plugin):
     """Manual Authenticator.
 
-    This plugin requires user's manual intervention in setting up a HTTP
-    server for solving SimpleHTTP challenges and thus does not need to be
-    run as a privilidged process. Alternatively shows instructions on how
-    to use Python's built-in HTTP server and, in case of HTTPS, openssl
-    binary for temporary key/certificate generation.
+    .. todo:: Support for `~.challenges.DVSNI`.
     """
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
-#    hidden = True
+    hidden = True
 
-    description = "Manually Edit Your Configuration"
+    description = "Manually Configure And Run A Simple Python Webserver"
 
     MESSAGE_TEMPLATE = """\
 Make sure your web server displays the following content at
@@ -84,7 +80,11 @@ s.serve_forever()" """
         pass  # pragma: no cover
 
     def more_info(self):  # pylint: disable=missing-docstring,no-self-use
-        return self.__doc__.replace("\n    ", " ")
+        return ("This plugin requires user's manual intervention in setting "
+                "up a HTTP server for solving SimpleHTTP challenges and thus "
+                "does not need to be run as a privilidged process. "
+                "Alternatively shows instructions on how to use Python's "
+                "built-in HTTP server.")
 
     def get_chall_pref(self, domain):
         # pylint: disable=missing-docstring,no-self-use,unused-argument
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -26,13 +26,18 @@ logger = logging.getLogger(__name__)
 class Authenticator(common.Plugin):
     """Manual Authenticator.
 
+       This plugin requires user's manual intervention in setting up a HTTP
+       server for solving SimpleHTTP challenges and thus does not need to be
+       run as a privilidged process. Alternatively shows instructions on how
+       to use Python's built-in HTTP server.
+
     .. todo:: Support for `~.challenges.DVSNI`.
     """
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
     hidden = True
 
-    description = "Manually Configure And Run A Simple Python Webserver"
+    description = "Manually configure and run a simple Python webserver"
 
     MESSAGE_TEMPLATE = """\
 Make sure your web server displays the following content at
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -26,12 +26,15 @@ logger = logging.getLogger(__name__)
 class Authenticator(common.Plugin):
     """Manual Authenticator.
 
-    .. todo:: Support for `~.challenges.DVSNI`.
-
+    This plugin requires user's manual intervention in setting up a HTTP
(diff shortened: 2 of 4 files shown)


letsencrypt/plugins/standalone.py
```diff
@@ -134,18 +134,10 @@ def supported_challenges_validator(data):
...
 
 class Authenticator(common.Plugin):
-    """Standalone Authenticator.
-
-    This authenticator creates its own ephemeral TCP listener on the
-    necessary port in order to respond to incoming DVSNI and SimpleHTTP
-    challenges from the certificate authority. Therefore, it does not
-    rely on any existing server program.
-
-    """
```
kuba: Please don't remove docstrings, they are authoritative source of documentation.

letsencrypt/display/ops.py
```diff
@@ -74,7 +74,7 @@ def pick_plugin(config, default, plugins, question, ifaces):
 
     if len(prepared) > 1:
         logger.debug("Multiple candidate plugins: %s", prepared)
-        plugin_ep = choose_plugin(prepared.values(), question)
+        plugin_ep = choose_plugin(prepared.values()[::-1], question)
```
kuba: Wait, what? :O Why do you want plugins to be reversed alphabetical order?

BTW, `prepared.values()` returns stuff in random order, so this makes it reverse random order.

Also, if we really really really want to do reverse alphabetical order (which doesn't make any sense to me), unit tests are missing.

SwartzCr:
now also addressing #1087 and #664

letsencrypt/plugins/standalone.py
L1ghtn1ng: @SwartzCr you are still missing this text from the doc string 
```   This authenticator creates its own ephemeral TCP listener on the
necessary port in order to respond to incoming DVSNI and SimpleHTTP
challenges from the certificate authority. Therefore, it does not
rely on any existing server program.```

**Comment by SwartzCr** [comment]
@L1ghtn1ng I'm doing that on purpose. I want the layout of standalone.py to mimic manual.py which has a minimal docstring and then a longer explanation in more_info. The reason I'm doing this is to enforce spacing because otherwise the fixed width of the display box messes up the formatting

**Comment by L1ghtn1ng** [comment]
@SwartzCr Fair enough, just thought you missed it

File: letsencrypt/plugins/manual.py [hunk 3]
```diff
@@ -82,10 +83,10 @@ def prepare(self):  # pylint: disable=missing-docstring,no-self-use
...
         return """\
 This plugin requires user's manual intervention in setting up a HTTP
-server for solving SimpleHTTP challenges and thus does not need to be
-run as a privilidged process. Alternatively shows instructions on how
-to use Python's built-in HTTP server and, in case of HTTPS, openssl
-binary for temporary key/certificate generation.""".replace("\n", "")
+ server for solving SimpleHTTP challenges and thus does not need to be
+ run as a privilidged process. Alternatively shows instructions on how
+ to use Python's built-in HTTP server and, in case of HTTPS, openssl
+ binary for temporary key/certificate generation.""".replace("\n", "")
```
kuba: Since we are hiding this plugin from the UI, user's will likely never see this message, so it's probably better to move it to the docstring.

Could we also add information that this plugin can be run on laptop, while the code is executed on a remote web server?

letsencrypt/plugins/standalone.py
```diff
@@ -182,7 +176,11 @@ def supported_challenges(self):
                    self.conf("supported-challenges").split(","))
 
     def more_info(self):  # pylint: disable=missing-docstring
-        return self.__doc__
+        return """\
+This authenticator creates its own ephemeral TCP listener on the
+ necessary port in order to respond to incoming DVSNI and SimpleHTTP
+ challenges from the certificate authority. Therefore, it does not
+ rely on any existing server program.""".replace("\n", "")
```
kuba: Maybe `replace("\n", " ")` and get rid of the 1 char indentation throughout the text? This will make text more readable and is easier not to accidentally mess up.

letsencrypt/plugins/manual.py
pde: Hmmm.  Perhaps instead we should print these strings if the user types `letsencrypt --help $PLUGIN`?

pde: Opened a ticket for that: #1101 .

letsencrypt/plugins/standalone.py
```diff
@@ -134,18 +134,12 @@ def supported_challenges_validator(data):
...
-    necessary port in order to respond to incoming DVSNI and SimpleHTTP
-    challenges from the certificate authority. Therefore, it does not
-    rely on any existing server program.
-
-    """
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
 
-    description = "Standalone Authenticator"
+    description = "LE Will Automatically Test Your Domain"
```
pde: This string should perhaps be, "Run a standalone webserver to prove that you control domains"?  Or "Use a standalone webserver to prove control of domains to the CA"?

SwartzCr: The problem with that is it's too long and cuts off the (standalone) text.
lelong

SwartzCr:
I've moved the info back to the docstrings and am now stripping the indentation out in the more_info calls

letsencrypt/plugins/standalone.py
pde: That's not *terrible* because the name is already there. `It could be Run a (standalone) webserver to prove that you control domains`...

SwartzCr: The problem is that the 'do x and stuff' (name) format comes from letsencrypt/plugins/disco.py
 49     def description_with_name(self):                                             
 50         """Description with name. Handy for UI."""                               
 51         return "{0} ({1})".format(self.description, self.name)
which would make it "Run a (standalone) weberver to probe that you control domains (....)
though I suppose we could change the display code to call 
 44     def description(self):                                                       
 45         """Description of the plugin."""                                         
 46         return self.plugin_cls.description      

thoughts?

letsencrypt/plugins/manual.py
```diff
@@ -26,13 +26,17 @@
 class Authenticator(common.Plugin):
     """Manual Authenticator.
 
-    .. todo:: Support for `~.challenges.DVSNI`.
```
bmw: So it appears there has been a little debate on this, but things like this are why I don't think we should be using the docstrings in `more_info`. It forces docstrings (which are primarily used by programmers) to be identical to the return value of `more_info` (which is seen by end users). There is no reason these things need to be coupled.

Personally, I'd just include whatever text you want shown to the user as a standalone string in `more_info`. If they're largely/completely duplicated with the docstring, so be it, but it allows us to more easily edit these strings in the future without unwanted side effects. Doing it this way, you can keep the `todo` comment.

letsencrypt/plugins/manual.py
```diff
@@ -26,13 +26,17 @@
...
-
+    This plugin requires user's manual intervention in setting up a HTTP
+    server for solving SimpleHTTP challenges and thus does not need to be
+    run as a privilidged process. Alternatively shows instructions on how
+    to use Python's built-in HTTP server and, in case of HTTPS, openssl
+    binary for temporary key/certificate generation.
     """
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
+#    hidden = True
```
bmw: I understand this was done so you could test the UI, but make sure this line isn't commented before we merge.

letsencrypt/plugins/manual.py
```diff
@@ -26,13 +26,17 @@
...
 
-    description = "Manual Authenticator"
+    description = "Manually Edit Your Configuration"
```
bmw: This is likely to be the most controversial part of this PR, but personally, I don't like this description. I would prefer something like "Manually configure and run a simple Python webserver".

letsencrypt/plugins/standalone.py
bmw: What about "Automatically configure and run a simple webserver"?

letsencrypt/plugins/manual.py
```diff
@@ -80,12 +84,7 @@ def prepare(self):  # pylint: disable=missing-docstring,no-self-use
         pass  # pragma: no cover
 
     def more_info(self):  # pylint: disable=missing-docstring,no-self-use
-        return """\
-This plugin requires user's manual intervention in setting up a HTTP
-server for solving SimpleHTTP challenges and thus does not need to be
-run as a privilidged process. Alternatively shows instructions on how
-to use Python's built-in HTTP server and, in case of HTTPS, openssl
-binary for temporary key/certificate generation.""".replace("\n", "")
+        return self.__doc__.replace("\n    ", " ")
```
bmw: If you take my advice about not using the docstrings here, I think we can do this more cleanly. I'd use the fact that Python [automatically concatenates string literals](https://docs.python.org/2/reference/lexical_analysis.html#string-literal-concatenation). This way, you can cleanly break everything on its own line but have total control over whitespace. How you choose to format it is up to you, but  you could do something like:

```
        return ("This plugin requires user's manual intervention in setting "
                "up a HTTP server for solving SimpleHTTP challenges and thus "
                "does not need to be run as a privilidged process. "
                "Alternatively shows instructions on how to use Python's "
                "built-in HTTP server.")
```

letsencrypt/plugins/manual.py
```diff
@@ -26,13 +26,17 @@
 class Authenticator(common.Plugin):
     """Manual Authenticator.
 
-    .. todo:: Support for `~.challenges.DVSNI`.
-
+    This plugin requires user's manual intervention in setting up a HTTP
+    server for solving SimpleHTTP challenges and thus does not need to be
+    run as a privilidged process. Alternatively shows instructions on how
+    to use Python's built-in HTTP server and, in case of HTTPS, openssl
```
bmw: Info about HTTPS is outdated. You can delete everything starting with "and, in the case of HTTPS..."

bmw:
Added my comments to this. As I said in a line comment, I expect the descriptions to be one of the more controversial parts of this PR. I added my suggestions. Thoughts @pde, @kuba?

bmw:
LGTM. Thanks for the help @SwartzCr.

letsencrypt/plugins/manual.py
```diff
@@ -26,13 +26,18 @@
...
+       to use Python's built-in HTTP server.
 
+    .. todo:: Support for `~.challenges.DVSNI`.
     """
     zope.interface.implements(interfaces.IAuthenticator)
     zope.interface.classProvides(interfaces.IPluginFactory)
+    hidden = True
 
-    description = "Manual Authenticator"
+    description = "Manually configure and run a simple Python webserver"
```
kuba: I think the description is confusing as it does not have to be a Python webserver. Someone could just use their existing webserver, as it is described in the information message.

letsencrypt/plugins/manual.py
kuba: That's a good reason to decouple in general, yes. However, since Manual is not displayed in the UI, it doesn't matter much in this case.

letsencrypt/plugins/manual.py
kuba: And this won't be even displayed in the UI, because the plugin is hidden. Which means that only technically inclined people will read it (those that go into the source code). They will be confused by this statement.

bmw: It's displayed with `letsencrypt -h manual` as well.

PR #1089, opened 2015-10-23
Implement ./letsencrypt-dev and simplify developer documentation

pde:
This should place much less cognitive load on new prospective developers!

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -7,40 +7,33 @@ Contributing
 Hacking
 =======
 
-All changes in your pull request **must** have 100% unit test coverage, pass
-our `integration`_ tests, **and** be compliant with the
-:ref:`coding style <coding-style>`.
+Running a local copy of the client
+----------------------------------
 
-
-Bootstrap
----------
-
-Start by :ref:`installing Let's Encrypt prerequisites
-<prerequisites>`. Then run:
+Running the client in developer mode from your local tree is very similar to
+running the released versions of the client; just replace ``letsencrypt-auto``
+with ``letsencrypt-dev``:
 
 .. code-block:: shell
 
-   ./bootstrap/dev/venv.sh
+   git clone https://github.com/letsencrypt/letsencrypt
+   cd letsencrypt
+   ./letsencrypt-dev
 
-Activate the virtualenv:
 
-.. code-block:: shell
+Find issues to work on
+----------------------
 
-   source ./$VENV_NAME/bin/activate
+You can find the open issues in the `github issue tracker`_.  If you're
+starting work on something, post a comment to let others know and seek
+feedback on your plan where appropriate.
 
-This step should prepend you prompt with ``($VENV_NAME)`` and save you
-from typing ``./$VENV_NAME/bin/...``. It is also required to run some
-of the `testing`_ tools. Virtualenv can be disabled at any time by
-typing ``deactivate``. More information can be found in `virtualenv
-documentation`_.
+Once you've got a working branch, you can open a pull request.  All changes in
+your pull request must have thorough unit test coverage, pass our
+`integration`_ tests, and be compliant with the :ref:`coding style
+<coding-style>`.
 
-Note that packages are installed in so called *editable mode*, in
-which any source code changes in the current working directory are
-"live" and no further ``./bootstrap/dev/venv.sh`` or ``pip install
-...`` invocations are necessary while developing.
-
-.. _`virtualenv documentation`: https://virtualenv.pypa.io
-
+.. _github issue tracker: https://github.com/letsencrypt/letsencrypt/issues
 
 Testing
 -------
@@ -64,8 +57,12 @@ The following tools are there to help you:
   but you won't get TAB completion...
 
 
-Integration
-~~~~~~~~~~~
+Integration testing with the boulder CA
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Generally it is sufficient to open a pull request and let Github and Travis run
+integration tests for you.
+
 Mac OS X users: Run `./tests/mac-bootstrap.sh` instead of `boulder-start.sh` to
 install dependencies, configure the environment, and start boulder.
 
@@ -89,55 +86,6 @@ the integration tests suite.
 
 .. _Boulder: https://github.com/letsencrypt/boulder
 .. _Go: https://golang.org
-
-
-Vagrant
--------
-
-If you are a Vagrant user, Let's Encrypt comes with a Vagrantfile that
-automates setting up a development environment in an Ubuntu 14.04
-LTS VM. To set it up, simply run ``vagrant up``. The repository is
-synced to ``/vagrant``, so you can get started with:
-
-.. code-block:: shell
-
-  vagrant ssh
-  cd /vagrant
-  ./venv/bin/pip install -r requirements.txt .[dev,docs,testing]
-  sudo ./venv/bin/letsencrypt
-
-Support for other Linux distributions coming soon.
-
-.. note::
-   Unfortunately, Python distutils and, by extension, setup.py and
-   tox, use hard linking quite extensively. Hard linking is not
-   supported by the default sync filesystem in Vagrant. As a result,
-   all actions with these commands are *significantly slower* in
-   Vagrant. One potential fix is to `use NFS`_ (`related issue`_).
-
-.. _use NFS: http://docs.vagrantup.com/v2/synced-folders/nfs.html
-.. _related issue: https://github.com/ClusterHQ/flocker/issues/516
-
-
-Docker
-------
-
-OSX users will probably find it easiest to set up a Docker container for
-development. Let's Encrypt comes with a Dockerfile (``Dockerfile-dev``)
-for doing so. To use Docker on OSX, install and setup docker-machine using the
-instructions at https://docs.docker.com/installation/mac/.
-
-To build the development Docker image::
-
-  docker build -t letsencrypt -f Dockerfile-dev .
-
-Now run tests inside the Docker image:
-
-.. code-block:: shell
-
-  docker run -it letsencrypt bash
-  cd src
-  tox -e py27
 
 
 Code components and layout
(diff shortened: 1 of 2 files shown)


docs/contributing.rst
```diff
@@ -195,17 +144,23 @@ in a separate branch).
...
 ---------
 
-Installers classes exist to actually setup the certificate and be able
-to enhance the configuration. (Turn on HSTS, redirect to HTTPS,
-etc). You can indicate your abilities through the
-:meth:`~.IInstaller.supported_enhancements` call. We currently only
+Installers classes exist to actually setup the certificate in a server,
+possibly tweak the security configuration to make it more correct and secure
+(Fix some mixed content problems, turn on HSTS, redirect to HTTPS, etc).
+Installer plugins tell the main client about their abilities to do the latter
```
bmw: I think the wording is confusing. Is there a difference between "Installer classes" and "Installer plugins"?

docs/contributing.rst
```diff
@@ -195,17 +144,23 @@ in a separate branch).
...
+via the :meth:`~.IInstaller.supported_enhancements` call. We currently only
 have one Installer written (still developing), `~.ApacheConfigurator`.
```
bmw: Quite a bit of development has gone into the `nginx` plugin. Should we say something about it here? Doing so might encourage people to work on it.

docs/contributing.rst
```diff
@@ -7,40 +7,33 @@ Contributing
...
-
-Note that packages are installed in so called *editable mode*, in
-which any source code changes in the current working directory are
-"live" and no further ``./bootstrap/dev/venv.sh`` or ``pip install
-...`` invocations are necessary while developing.
-
-.. _`virtualenv documentation`: https://virtualenv.pypa.io
+Once you've got a working branch, you can open a pull request.  All changes in
+your pull request must have thorough unit test coverage, pass our
+`integration`_ tests, and be compliant with the :ref:`coding style
```
bmw: Integration link doesn't work.

docs/contributing.rst
```diff
@@ -7,40 +7,33 @@ Contributing
...
-Activate the virtualenv:
 
-.. code-block:: shell
+Find issues to work on
+----------------------
 
-   source ./$VENV_NAME/bin/activate
+You can find the open issues in the `github issue tracker`_.  If you're
+starting work on something, post a comment to let others know and seek
+feedback on your plan where appropriate.
```
bmw: Perhaps mention the "Good Volunteer Task" label?

docs/contributing.rst
```diff
@@ -7,40 +7,33 @@ Contributing
...
 
-This step should prepend you prompt with ``($VENV_NAME)`` and save you
```
bmw: So much of this documentation relies on having having activated the `virtualenv`. For instance, without doing this, to run `tox` users with this setup will have to run `$XDG_DATA_HOME/letsencrypt-dev/bin/tox` (and this is if they are lucky enough to have the environment variable set). Even worse, this is not explained in the documentation below so people will follow the instructions perfectly and get `command not found: tox`.

docs/contributing.rst
```diff
@@ -91,55 +88,6 @@ the integration tests suite.
...
-   Vagrant. One potential fix is to `use NFS`_ (`related issue`_).
-
-.. _use NFS: http://docs.vagrantup.com/v2/synced-folders/nfs.html
-.. _related issue: https://github.com/ClusterHQ/flocker/issues/516
-
-
-Docker
-------
-
-OSX users will probably find it easiest to set up a Docker container for
```
bmw: Does `letsencrypt` and all of it's tests now work on OS X without Docker? At the very least, integration tests do not work for @jmhodges.

docs/contributing.rst
```diff
@@ -293,6 +248,94 @@ directory.
...
+====================================
+
+Lower level venv scripts
+------------------------
+
+You can get slightly lower level exposure to virtualenv by using these
+scripts as an alternative to ``letsencrypt-dev``.
+
+these
+by :ref:`installing Let's Encrypt prerequisites <prerequisites>`. Then run:
```
bmw: Bad link.

letsencrypt-auto
```diff
@@ -3,16 +3,29 @@
 # A script to run the latest release version of the Let's Encrypt in a
 # virtual environment
 #
+# ./letsencrypt-auto     --      run the latest public release from PyPI
+#
+# ./lesencrypt-dev       --      run the copy from the local tree, including
```
bmw: I strongly feel that these should be separate scripts. As I've noted before, I believe a non-negligible fraction of our beta users will be reading this script to see what the hell we are doing to their servers when we are automatically downloading, installing, upgrading, and running the client. This script should be as small, concise, and readable as possible and making this script double as a script for developers moves in the opposite direction.

letsencrypt-auto
```diff
@@ -3,16 +3,29 @@
...
+#                                experimental plugins
+#
 # Installs and updates the letencrypt virtualenv, and runs letsencrypt
 # using that virtual environment.  This allows the client to function decently
 # without requiring specific versions of its dependencies from the operating
 # system.
 
 XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
-VENV_NAME="letsencrypt"
+if [ "`basename \"$0\"`" = "letsencrypt-dev" ] ; then
```
bmw: So this isn't included in this PR but how were you going to do this. Symlink?

letsencrypt-auto
```diff
@@ -65,10 +79,27 @@ then
...
+  cd $LE_ROOT
+  $VENV_BIN/pip install \
+    -r requirements.txt \
+    -e acme[testing] \
+    -e .[dev,docs,testing] \
+    -e letsencrypt-apache \
+    -e letsencrypt-nginx \
+    -e letshelp-letsencrypt \
+    -e letsencrypt-compatibility-test > /dev/null
+elif [ "$VERBOSE" = 1 ]  ; then
```
bmw: Why can't developers have verbose output?

docs/contributing.rst
```diff
@@ -64,8 +57,12 @@ The following tools are there to help you:
   but you won't get TAB completion...
 
 
-Integration
-~~~~~~~~~~~
+Integration testing with the boulder CA
```
bmw: This change broke other links that the one I commented on above.

bmw:
So being completely honest, I don't know if we want this. Sure, this script combines the step of bootstrapping and running `bootstrap/dev/venv.sh` but I think that's the only real gain here. Bootstrapping is only done once and will not even be needed if they tried out the client using `letsencrypt-auto` before contributing.

One could make an argument that this helps a new contributor in the case that the entry points change and the client needs to be installed again. This is true only in the case that the developer runs `letsencrypt-dev` before running tests. If they run `tox` first, they'll get an error. Additionally, while this may be worth considering in the long term, it is not in the short term as there are very few open issues that require that this be done and none of them are tagged "1.0 for launch".

This script places all testing executables in a directory that is likely to be in a very different location than the Python files the user is editing. What's the workflow here? In my personal configuration this means I edit my files in `/home/bmw/Code/letsencrypt` and then run all my tests in `/home/bmw/.local/share/letsencrypt-dev/bin`. This seems suboptimal.

@pde, if you want to make it easier for new contributors, I'd recommend the following:
* Suggest they try the client using `letsencrypt-auto`, saying this will install all necessary dependencies. Maybe include one line pointing them to the bootstrap scripts if they'd like to do this step manually.
* Have them run `bootstrap/dev/venv.sh` to setup the development environment. This script can be moved to a different location if you'd prefer.
* Shorten the documentation about activating the `virtualenv` and not having to reinstall after making changes. We can debate the approach of using terminal specific state, but I think activating the `virtualenv` simplifies the process of running the client, `tox`, and integration tests. If we don't include this in the documentation, then we need to explain how the user goes about running the programs without activating `virtualenv`.

As a very busy LE developer, I very much appreciate trying to make it easier for new contributors to join the project, I'm just not sure if this is the right approach. If there are other benefits to the `letsencrypt-dev` approach that I'm not thinking of, please let me know.

letsencrypt-auto
kuba: :+1:

docs/contributing.rst
pde: I run tox without activating venv; it seems to work as long as you're in the checked out tree...

kuba:
First of all, I though we agreed that this is a non-GA task! Frankly, I don't think we should even spend time on reviewing this, because we have plenty of other stuff to do. After having a quick look, there are more issues than Brad pointed, but I don't feel its best use of my engineering time to elaborate now.

Secondly, our current documentation for developers is pretty concise and requires 3 commands to be copied and pasted. I strongly beleive we shouldn't hide those 3 commands under a single magical script. In fact, I think that all developers should be well aware how `virtualenv` works/should be set up, and why do we use it. Hiding this kind of stuff is a really bad design to will lead to many misunderstandings.

I pretty much agree with all what Brad said.

Strong no-LGTM from me.

letsencrypt-auto
pde: Yes, forgot to git add the link.  I'm fine factoring this out into another script, though...

docs/contributing.rst
kuba: That's because you're using OS-wide `tox` and you shouldn't do that, because it causes strange side-effects that are hard to debug!

In fact most of the people won't even have OS-wide `tox` installed.

pde:
@kuba I thought we could put this off, but Josh is keen to recruit more client contributors, which I think means we need to make contributing.rst comprehensible to people who have never worked with virtualenv before.  I'm fine with other ways of doing this -- Brad's proposal is workable -- I just went with this one because it provides a maximally consistent experience for regular and developer users.

docs/contributing.rst
bmw: Right. None of our installation scripts install OS-wide `tox`.

pde:
Brad persuaded me to go with documenting the "activate" process rather than using a new script.  Will open a new PR for that.

Issue #892, opened 2015-10-04
Bad flag error handling in CLI


kuba:
Misspelling flag gives very confusing error message relating to the subcommand:
```
$ letsencrypt --asd qwe auth
usage: 
  letsencrypt [SUBCOMMAND] [options] [domains]

The Let's Encrypt agent can obtain and install HTTPS/TLS/SSL certificates.  By
default, it will attempt to use a webserver both for obtaining and installing
the cert.  
letsencrypt: error: argument SUBCOMMAND: invalid choice: 'qwe' (choose from 'run', 'auth', 'install', 'revoke', 'rollback', 'config_changes', 'plugins')
```

It should rather say: "`--asd` flag was not recognized". Optionally it should suggest autocorrections.

pde:
I don't think this is a 1.0 bug.  While it would be nice to have, argparse doesn't support it and we'd have to implement our own comprehensive option scanner to make it work.

kuba:
I think this is just a consequence of hacking argparse and is just one evidence that it's done wrong, a hint that more things will be failing. I'd suggest to revert magical "run-is-default-subcommand" functionality.

bmw:
I have a fix for this that I discussed with @pde that I could implement and test in no more than a couple hours.

PR #1099, opened 2015-10-23
Remove CLI subparsers

### LINKED ISSUE #640: User supplied cli.ini files don't seem to work
As suggested by the documentation, I tried this, which failed:

$ ./venv/bin/letsencrypt --config examples/dev-cli.ini auth
usage: 

the cert.  
letsencrypt: error: unrecognized arguments: --config-dir /tmp/le/conf --work-dir /tmp/le/conf --logs-dir /tmp/le/logs --email foo@example.com --domains example.com --text --agree-eula --debug --verbose --authenticator standalone

bmw:
Subparsers in `cli.py` have been an unfortunate source of problems for us (most of which are due to problems upstream). This PR removes them completely and fixes #640 and #892.

cc @pde

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -655,7 +655,11 @@ class HelpfulArgumentParser(object):
         return parsed_args
 
     def determine_verb(self):
-        """Determines the verb/subcommand provided by the user."""
+        """Determines the verb/subcommand provided by the user.
+
+        This function works around some of the limitations of argparse.
+
+        """
         if "-h" in self.args or "--help" in self.args:
             # all verbs double as help arguments; don't get them confused
             self.verb = "help"
@@ -756,7 +760,7 @@ class HelpfulArgumentParser(object):
             return dict([(t, t == chosen_topic) for t in self.help_topics])
 
 
-def parse_args(plugins, args):
+def prepare_and_parse_args(plugins, args):
     """Returns parsed command line arguments.
 
     :param .PluginsRegistry plugins: available plugins
@@ -1040,7 +1044,7 @@ def main(cli_args=sys.argv[1:]):
 
     # note: arg parser internally handles --help (and exits afterwards)
     plugins = plugins_disco.PluginsRegistry.find_all()
-    args = parse_args(plugins, cli_args)
+    args = prepare_and_parse_args(plugins, cli_args)
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -341,7 +341,7 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
 
     :param str cfg_type: either "installer" or "authenticator"
     :param str requested: the plugin that was requested
-    :param PluginRegistry plugins: available plugins
+    :param .PluginsRegistry plugins: available plugins
 
     :raises error.PluginSelectionError: if there was a problem
     """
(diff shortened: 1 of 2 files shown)


letsencrypt/cli.py
```diff
@@ -631,25 +642,32 @@ def __init__(self, args, plugins):
...
-        subcommand / verb comes last.
         """
-        if "-h" in args or "--help" in args:
+        parsed_args = self.parser.parse_args(self.args)
+        parsed_args.func = self.VERBS[self.verb]
+
+        return parsed_args
+
+    def determine_verb(self):
+        """Determines the verb/subcommand provided by the user."""
```
pde: I recommend leaving in a comment or docstring here to say that we're preprocessing to work around argparse limiations. Otherwise that might be quite confusing to future readers.

letsencrypt/cli.py
```diff
@@ -631,25 +642,32 @@ def __init__(self, args, plugins):
...
+        """Parses command line arguments and returns the result.
+
+        :returns: parsed command line arguments
+        :rtype: argparse.Namespace
 
-        Currently: add the default verb "run" as a default, and ensure that the
-        subcommand / verb comes last.
         """
-        if "-h" in args or "--help" in args:
+        parsed_args = self.parser.parse_args(self.args)
```
pde: Nit: can you move the call to `self.determine_verb()` to the line above this without breaking anything?  Since it's part of the parsing process, it makes sense to call here.

letsencrypt/cli.py
```diff
@@ -738,8 +756,16 @@ def determine_help_topics(self, chosen_topic):
             return dict([(t, t == chosen_topic) for t in self.help_topics])
 
 
-def create_parser(plugins, args):
-    """Create parser."""
+def parse_args(plugins, args):
```
pde: I don't really like having two functions / methods in the same file with the same name.  Maybe call this one `prepare_and_parse_args`?

letsencrypt/cli.py
```diff
@@ -1040,8 +1040,7 @@ def main(cli_args=sys.argv[1:]):
 
     # note: arg parser internally handles --help (and exits afterwards)
     plugins = plugins_disco.PluginsRegistry.find_all()
-    parser, tweaked_cli_args = create_parser(plugins, cli_args)
-    args = parser.parse_args(tweaked_cli_args)
```
pde: The variable `tweaked_cli_args` was used to preserve the original value of `args` because something about the modified version caused problems for configuration.NamespaceConfig.  I can't remember what it was, though.  Unsure if you'll need to reintroduce that preservation.

letsencrypt/cli.py
bmw: Doing so breaks `_paths_parser` which relies on knowing the verb.

letsencrypt/cli.py
bmw: I'm not sure if I understand. `tweaked_cli_args` wasn't used anywhere other than the call to `parser.parse_args(tweaked_cli_args)` and the value of `args` that `configuration.NamespaceConfig(args)` uses should be identical to the previous version (if it's not, I introduced a bug). Am I not understanding something?

bmw:
Ready for rereview.

letsencrypt/cli.py
pde: I was worried that there might have been a subtle bug of that sort (some call needing access to the command line args before they were reordered and a default verb introduced), but I think I was mistaken.  Tweaked_cli_args was introduced in 44d7ac2dce2691116da0ba4fa88d37d7fde89134 because of the opposite problem -- the Namespace being made from the unedited arguments.

pde:
LGTM; merging.

Issue #971, opened 2015-10-14
Confusing error message when manually specifying authenticator


jsha:
Steps to reproduce:

```
TMP=`mktemp -d`
letsencrypt --email foo@example.com --server http://localhost:4000/directory  --text --agree-dev-preview --agree-tos -d foo.com --config-dir $TMP --work-dir $TMP --logs-dir  $TMP --dvsni-port 5001 --nginx-server-root $TMP -a nginx  auth 
```

Expected result:

"Failed to initialize Nginx plugin: Could not find nginx.conf"

Actual result:

"Authenticator could not be determined"

pde:
Partially addressed by #1001

pde:
So this has improved; we now say: "The nginx plugin is not working; there may be problems with your existing configuration".  We should mention the logfile or better yet, surface the error.

Issue #1092, opened 2015-10-23
Improve NoInstallationError message


kuba:
https://github.com/letsencrypt/letsencrypt/blob/v0.0.0.dev20151021/letsencrypt-apache/letsencrypt_apache/configurator.py#L145
1. We should never do `raise Exception`, always `raise Exception()`
2. I just had a report from RHEL user that he couldn't get Apache plugin to work. Turns out that default server root is `/etc/httpd`. Can we please do `raise errors.NoInstallationError("Configs not found at {0}".format(self.conf("server-root")))` so that the explicit error message is found in the logs? Now it's only:
   
   ```
   2015-10-23 14:40:49,632:DEBUG:letsencrypt.plugins.disco:No installation (PluginEntryPoint#apache):
   Traceback (most recent call last):
   File "/usr/local/lib/python2.7/site-packages/letsencrypt/plugins/disco.py", line 103, in prepare
     self._initialized.prepare()
   File "/usr/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 145, in prepare
     raise errors.NoInstallationError
   NoInstallationError
   ```

Could you look into that @swartzcr, please? :)

pde:
Should be fixed these days.

PR #1100, opened 2015-10-24
Pass plugin errors back out to the user.

pde:
Closes: #971

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -351,8 +351,8 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
             msg = "The requested {0} plugin does not appear to be installed".format(requested)
         else:
             msg = ("The {0} plugin is not working; there may be problems with "
-                   "your existing configuration.\nThe error was: {1}"
-                   .format(requested, `plugins[requested].problem`))
+                   "your existing configuration.\nThe error was: {1!r}"
+                   .format(requested, plugins[requested].problem))
     elif cfg_type == "installer":
         if os.path.exists("/etc/debian_version"):
             # Debian... installers are at least possible
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -351,7 +351,8 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
             msg = "The requested {0} plugin does not appear to be installed".format(requested)
         else:
             msg = ("The {0} plugin is not working; there may be problems with "
-                   "your existing configuration").format(requested)
+                   "your existing configuration.\nThe error was: {1}"
+                   .format(requested, plugins[requested].problem))
     elif cfg_type == "installer":
         if os.path.exists("/etc/debian_version"):
             # Debian... installers are at least possible
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -352,7 +352,7 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
         else:
             msg = ("The {0} plugin is not working; there may be problems with "
                    "your existing configuration.\nThe error was: {1}"
-                   .format(requested, plugins[requested].problem))
+                   .format(requested, `plugins[requested].problem`))
     elif cfg_type == "installer":
         if os.path.exists("/etc/debian_version"):
             # Debian... installers are at least possible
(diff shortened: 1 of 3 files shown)


letsencrypt/cli.py
```diff
@@ -351,7 +351,8 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
             msg = "The requested {0} plugin does not appear to be installed".format(requested)
         else:
             msg = ("The {0} plugin is not working; there may be problems with "
-                   "your existing configuration").format(requested)
+                   "your existing configuration.\nThe error was: {1}"
+                   .format(requested, plugins[requested].problem))
```
bmw: I think we should also be displaying the exception type, at least while there are issues like #1092 around. Otherwise, people will get the following confusing behavior:
```
$ export TMP=$(mktemp -d)
$ letsencrypt --config-dir $TMP/config --work-dir $TMP/work --logs-dir $TMP/logs --apache -t --agree-dev-preview
2015-10-23 18:01:16,865:WARNING:letsencrypt.cli:Root (sudo) is required to run most of letsencrypt functionality.
The error was: 
$
```
It's a UX question I don't know the answer to whether we should continue doing this once #1092 is fixed, but I think we should at least be doing it for the time being.

bmw: It'd also be nice if these lines had test coverage.

letsencrypt/plugins/disco.py
```diff
@@ -121,6 +121,12 @@ def misconfigured(self):
         return isinstance(self._prepared, errors.MisconfigurationError)
 
     @property
+    def problem(self):
+        """Return the Exception raised during plugin setup, or None if all is well"""
+        if isinstance(self._prepared, Exception):
+            return self._prepared
```
kuba: Please explicitly `return None` in the `else` branch.

Unittest missing!

kuba:
Please fix Brad's comment, add unittests and ask for re-review.

pde:
We've already established that unit testing of the CLI around misconfigured plugins is a technical rabbit hole, because the unit tests run without the plugins being present.  Given our tight schedule, the best approach is to do some live testing (which I've done) and then perhaps to add tests for the these code paths to the integration test suite, where we can ensure that the plugins we're testing are actually installed.

letsencrypt/cli.py
pde: I thought about displaying the exception type, but when I tried it out it produced a somewhat confusing result.  For instance with jsha's example of the problem from #971, we get:

```
pde@damoclid:~/letsencrypt/letsencrypt$ sudo venv/bin/letsencrypt  --email foo@example.com --server http://localhost:4000/directory  --text --agree-dev-preview --agree-tos -d foo.com --config-dir $TMP --work-dir $TMP --logs-dir  $TMP --dvsni-port 5001 --nginx-server-root $TMP -a nginx  auth 
The nginx plugin is not working; there may be problems with your existing configuration.
The error was: NoInstallationError('Could not find configuration root',)
```
Arguably that's not really a "NoInstallationError".  But your example suggests that it might be worth the price, so I'll make the change.

kuba:
diagnose_configurator_problem doesn't even call PluginsRegistry, so you can easily test it without any "technical rabbit hole" like issues. You should also add test for new `PluginEntryPoint.problem`

letsencrypt/cli.py
```diff
@@ -351,7 +351,8 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
             msg = "The requested {0} plugin does not appear to be installed".format(requested)
         else:
             msg = ("The {0} plugin is not working; there may be problems with "
-                   "your existing configuration").format(requested)
+                   "your existing configuration.\nThe error was: {1}"
+                   .format(requested, `plugins[requested].problem`))
```
kuba: Please do not use backticks! `{1!r}` should do it.

pde:
Working on the `PluginEntryPoint.problem` test right now.  I don't think that unit testing `diagnose_configurator_problem` in complete isolation from all of the logic around it is the best use of our limited time before launch. But I've opened #1122 to track the task of getting more meaningful tests written for it.

letsencrypt/cli.py
pde: What's wrong with backticks?

kuba: They are completely unreadable and deprecated. Also, run `pep8` on your code ;)

pde:
All issues aside from the unresolvable debate about unit testing `diagnose_configurator_problem` addressed.  Merging.

PR #1103, opened 2015-10-24
Use better example config

### LINKED ISSUE #1045: Dev config example in user's guide might be confusing
User reported that all his certs landed in `/tmp/le/conf`... Root cause? Blindly copying config file presented at http://letsencrypt.readthedocs.org/en/latest/using.html#configuration-file. Maybe it would be better not to show **dev** config as an example over there?

bmw:
Fixes  #1045.

### DIFF
--- a/examples/cli.ini
+++ b/examples/cli.ini
@@ -1,16 +1,16 @@
 # This is an example of the kind of things you can do in a configuration file
 
+# Use a 4096 bit RSA key instead of 2048
+rsa-key-size = 4096
+
+# Always use the staging/testing server
+server = https://acme-staging.api.letsencrypt.org/directory
+
 # Uncomment and update to register with the specified e-mail address
 # email = foo@example.com
 
 # Uncomment to use a text interface instead of ncurses
 # text = True
-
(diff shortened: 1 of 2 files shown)


examples/cli.ini
```diff
@@ -0,0 +1,22 @@
+# This is an example of the kind of things you can do in a configuration file
+
+# Use a 4096 bit RSA key instead of 2048
+rsa-key-size = 4096
+
+# Always use the staging/testing server
+server = https://acme-staging.api.letsencrypt.org/directory
```
kuba: FYI this will not stop people from overriding using `--server` on the CLI, if that's what you had in mind.

examples/cli.ini
```diff
@@ -0,0 +1,22 @@
+# This is an example of the kind of things you can do in a configuration file
```
kuba: I would add a comment that you can set (at least (1)) all flags that are visible through `--help all`.

(1) since manual is hidden, `--manual-*` is hidden from `--help all` even

kuba:
LGTM

bmw:
Ready for rereview.

kuba:
LGTM

PR #1105, opened 2015-10-24
fixed the git path

joubin:
the path was missing a `.git` which forces github to require a password to clone.

### DIFF
--- a/bootstrap/install-deps.sh
+++ b/bootstrap/install-deps.sh
@@ -9,7 +9,7 @@ else
   SUDO=
 fi
 
-BOOTSTRAP=`dirname $0`/bootstrap
+BOOTSTRAP=`dirname $0`/
 if [ ! -f $BOOTSTRAP/debian.sh ] ; then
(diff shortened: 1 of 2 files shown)


kuba:
I never ever had to type in my password using the already existing command. Password is usually required when using `ssh` transport...

bmw:
```
$ git clone https://github.com/letsencrypt/letsencrypt
Cloning into 'letsencrypt'...
remote: Counting objects: 21589, done.
remote: Compressing objects: 100% (29/29), done.
remote: Total 21589 (delta 13), reused 0 (delta 0), pack-reused 21560
Receiving objects: 100% (21589/21589), 5.59 MiB | 2.59 MiB/s, done.
Resolving deltas: 100% (14972/14972), done.
Checking connectivity... done.
$
```
Closing as this doesn't appear to be necessary. Like kuba said, a password is required if you used `git@github.com:letsencrypt/letsencrypt`, but shouldn't be if you use a `https` version.

Regardless, thanks for the PR @joubin.

PR #1107, opened 2015-10-24
Made Stuff more clearly worded

akbapu14:
I got confused as I was running through...

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -11,8 +11,7 @@ Running a local copy of the client
 ----------------------------------
 
 Running the client in developer mode from your local tree is a little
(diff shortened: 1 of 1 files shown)


kuba:
I personally find the wording of this PR even more confusing :(

bmw:
I agree. @pde felt it was important to point out which steps you only need to do once. Furthermore, you don't "activate shells with `tox`".

With that said @akbapu14, we'd love your feedback as we're trying to make the instructions as friendly as possible to new developers. Where did you have trouble?

bmw:
Closing this, but we'd still love your feedback @akbapu14 if you're willing to provide it.

PR #1108, opened 2015-10-24
Fix more_info messages with proper spacing

dannysu:
Makes things easier to read


kuba:
@swartzcr is already working on a conflicting change in #1079, where I left some negative comments about using a similar approach

bmw:
Thanks for the PR @dannysu but these problems were fixed in #1079 and #1146.

Issue #1091, opened 2015-10-23
/var/log/letsencrypt/letsencrypt.log missing?


cedw93:
I am attempting to generate a certificate for `*.test.local.com`

I am on Mac OS X Yosemite (10.10.4) and I am running the following:

`./letsencrypt-auto auth`

and I follow the prompts on screen and after I enter the domain `*.test.local.com` the following error pops up:

```
 An unexpected error occurred.
 Error: unauthorized :: The client lacks sufficient authorization :: Error creating new authz :: Syntax    error
 Please see the logfiles in /var/log/letsencrypt for more details.
```

But the contents of `/var/log/letsencrypt' is empty. 

I thought it might be a permission problem so I attempted to the as `sudo` however I got errors similar to the following:

```
The directory '/Users/intilery/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
```

So I added `-H` to the command and `brew` refused to run the application:

```
Bootstrapping dependencies for Mac OS X...
WARNING: Mac support is very experimental at present...
Error: Cowardly refusing to `sudo brew install`
You can use brew with sudo, but only if the brew executable is owned by root.
However, this is both not recommended and completely unsupported so do so at
your own risk.
```

Is this a known issue, or is there a solution?

bmw:
The syntax error is due to Let's Encrypt not currently supporting wildcard certificates. From the [FAQ](https://community.letsencrypt.org/t/frequently-asked-questions-faq/26):

> Will Let’s Encrypt issue wildcard certificates?
>
> We currently have no plans to do so, but it is a possibility in the future. Hopefully wildcards aren’t necessary for the vast majority of our potential subscribers because it should be easy to get and manage certificates for all subdomains.

As for the logs, the client should have errored out with a message like `IOError: [Errno 13] Permission denied: '/var/log/letsencrypt/letsencrypt.log'`. If that directory is empty, you could try deleting it and running the client again. A temporary workaround is to specify a different location for logfiles. To do this, simply include `--logs-dir /path/to/wherever` on the command line.

There is no need to run `letsencrypt-auto` as `sudo` (in fact, you shouldn't do it). The script knows which commands it executes require `sudo` and prompt you accordingly.

[...]

PR #1111, opened 2015-10-24
Revert old install warning (fixes #1044)

### LINKED ISSUE #1044: Revert old warning about unsupported installation modes
@pde ignored my comment about reverting old warning about unsupported installation modes before merging #980. As anticipated this resulted in user's confusion: https://news.ycombinator.com/item?id=10417542. That's a tracking bug for me to revert old message.

kuba:


kuba:
Affected by https://github.com/letsencrypt/boulder/issues/756, restarting Travis job.

docs/using.rst
```diff
@@ -35,16 +35,20 @@ To install and run the client you just need to type:
...
 
 .. code-block:: shell
 
    ./letsencrypt-auto --help
 
+.. warning:: Please do **not** use ``python setup.py install`` or
+   ``python pip install .``. Please do **not** attempt the
+   installation commands as superuser/root and/or without virtual
+   environment, e.g. ``sudo python setup.py install``, ``sudo pip
+   install``, ``sudo ./venv/bin/...``. These modes of operation might
```
bmw: I wonder if we should specifically mention not running `sudo ./letsencrypt-auto` (see #1091).

bmw: Although that's slightly confusing since `letsencrypt-auto` uses `sudo` internally and will prompt you for your `sudo` password. Thoughts @kuba, @pde?

kuba: Hm, `sudo letsencrypt-auto` will install stuff to `~root/.local/share` instead of `~/.local/share`. I think it should be explicitly disallowed as well (although noting bad would actually happen).

bmw: Bad things happen on OS X.

bmw:
After thinking about it, I think this is fine. LGTM. Thanks @kuba.

PR #1109, opened 2015-10-24
removed extra path

joubin:
```
root@localhost:~# cd letsencrypt/
root@localhost:~/letsencrypt# ./bootstrap/install-deps.sh 
Cannot find the letsencrypt bootstrap scripts in ./bootstrap/bootstrap

```

However, with the proposed change

```
root@localhost:~/letsencrypt# ./bootstrap/install-deps.sh 
Bootstrapping dependencies for Debian-based OSes...
Ign http://mirrors.linode.com jessie InRelease
Hit http://mirrors.linode.com jessie-updates InRelease  
```

PR #1116, opened 2015-10-24
Fix BOOTSTRAP path.

titusjose:
Running ./bootstrap/install-deps.sh returns error:
Cannot find the letsencrypt bootstrap scripts in ./bootstrap/bootstrap


bmw:
@titusjose, I actually just merged #1109 which solved the same problem. Thanks for the PR though!

PR #1120, opened 2015-10-24
Fix DST caused 1 hour offset in parse_time_interval() function

devnsec-com:
The `parse_time_interval()` function was using `time.localtime()` and `time.mktime()` functions, which will use local time zone and DST settings, in some circumstances it will result in 1 hour offset.

This PR will fix this bug by using GMT (aka UTC) time instead of local time.

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -1,4 +1,5 @@
 """Renewable certificates storage."""
+import calendar
 import datetime
 import os
 import re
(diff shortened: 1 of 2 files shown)


pde:
LGTM

kuba:
Is it possible to write unit test that would catch an 1 hour offset error? To be honest, I don't think that either old or new code make much sense to me :(

devnsec-com:
@kuba I've added the unit test for it.

BTW, I agree with you, the current implementation of this function still looks wrong for me.
For an example, if today is Jan 1, 2000, and the input of the function is `"1 year"`, what's the expected output? I would say `366`, but actually it will give me `365`, because it always use the time offset based on `1970-01-01T00:00:00+0000`.

devnsec-com:
I've created a new PR #1144 to replace this one.

schoen:
@devnsec-com, I'm going to merge #1144 -- thanks!

Issue #1000, opened 2015-10-16
Docker documentation


bmw:
Running the client in Docker is an amazingly simple way to obtain a certificate. Unfortunately, running the client in Docker removes the ability to automatically install, renew, and deploy the certificate in all but the most complicated of configurations. These limitations need to be explained in the documentation.

kuba:
1000, congrats!

Issue #951, opened 2015-10-11
Docs for automatic renewal


kuba:
This is a follow-up for https://github.com/letsencrypt/letsencrypt/pull/912#issuecomment-147235477. We should document how to set up crontab to use `letsencrypt-renewer`. We need instructions for both end-users (running `pip install`) and distro package maintainers.

Issue #608, opened 2015-07-13
Missing Docs for cert.pem, chain.pem, fullchain.pem, etc


coolaj86:
When this goes live, **where will the intermediate certificates be located**?

Searching the docs for `chain.pem` yields [no results](https://letsencrypt.readthedocs.org/en/latest/search.html?q=chain.pem&check_keywords=yes&area=default).

I just [asked on the mailing list](See https://groups.google.com/a/letsencrypt.org/forum/#!topic/client-dev/jE5uK4lPx5g), but it seems it should be an issue on here as well.

Currently we have these files:

* privkey.pem
* cert.pem
* chain.pem
* fullchain.pem

But we don't know what goes in them. With the current dummy certificate it puts the **Root CA** in the `chain.pem` (which seems wrong) and `fullchain.pem` and that the **Server Certificate** goes into `cert.pem` and `fullchain.pem`, but we don't know where the **Intermediate Certificates** go.

This is extremely important because some webservers are lax, others are strict, others barf if you give too much info, others silently fail if you give to little.

I've already written some documentation and demo code based on assumption, but I already know that that's going to blow up in my face if I don't get the details correct.

In order for greatest compatibility with various servers it would seem that we need a file that bundles the server + intermediate certificates, and *excludes* the root certificate. I'm hoping that's what `cert.pem` is intended to be.

Once I know the technical details as to what these files are supposed to represent, I may open another issue to discuss the need for additional files.

coolaj86:
> At the moment we don't have any intermediate chains deployed on our test systems.  When those exist (for launch) they will be present in chain.pem and fullchain.pem, and should take the place of the root that's currently in those files.  At that stage, if you want to get a copy of the Let's Encrypt root, that would be an out of band operation (or could be a separate command line flag).

See https://groups.google.com/a/letsencrypt.org/d/msg/client-dev/jE5uK4lPx5g/mbufIV-2OMYJ

asquelt:
i'd propose `combined.pem` as sum of `privkey.pem` and `fullchain.pem`. haproxy reverse proxy requires this format, stunnel also accepts it.

https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#5.1-crt

coolaj86:
+1

I just set this up in haproxy myself yesterday.

```bash
cat cert.pem chain.pem privkey.pem > combined.pem
```

jdkasten:
I think this may being taken care of starting with #762 (multi-step process)

[...]

PR #1121, opened 2015-10-24
Reorganize and extend docs

kuba:
1. Separated docs into 3 logical sections:
   - User Guide
   - Developer Guide
   - Packaging Guide

2. Moved plugins.rst stuff to Developer Guide as it only described how to write new plugins.

3. Added Plugins table with basic notes and stability status to the User Guide. Linked to https://github.com/letsencrypt/letsencrypt/wiki/Plugins.

4. Make it clear what `letsencrypt-auto` is. There were comments and concerns throughout IRC / GitHub about that. Logically separate other installation methods (Docker, from distro, from source).


kuba:
You can view this branch generated at http://letsencryptkuba.readthedocs.org/en/docs/.

pde:
Will this break links to the existing pages on readthedocs?

kuba:
If those pages were linked from outside repo / documentation: then yes. But it's the same breakage as in all previous docs commits... Otherwise all other references have been fixed.

pde:
We have a bunch of links that we've given to people in community.letsencrypt.org, on the mailing list, etc, explaining things like how to find the plugin API and where to get started on development.  For that reason I think we should keep the existing filenames (contributing.html, using.html).

kuba:
@pde, reverted to `contributing.html`, `using.html`.

kuba:
This PR has now been extended with more commits, including fix for 3 issues: #608, #951, #1000. See git log for more info. http://letsencryptkuba.readthedocs.org/en/docs/ is up to date.

docs/using.rst
```diff
@@ -1,45 +1,109 @@
...
-Getting the code
-================
+.. _installation:
 
-Please `install Git`_ and run the following commands:
+Installation
+============
+
+Unless you have a very specific requirements, we kindly ask you to use
+the letsencrypt-auto_ method described below. It's the fastest, the
```
pde: Please remove this paragraph.  Rather than opening with a discussion of different installation methods, we should open with simple instructions for the best one.  If you really want to put this in the help, put it in the section where we talk about other ways of installing / running the client.

docs/using.rst
```diff
@@ -50,8 +114,128 @@ The ``letsencrypt`` commandline tool has a builtin help:
...
+               (force by ``standalone-supported-challenges dvsni``).
+webroot    Y N Works with already running webserver, by writing necessary files
+               to the disk (``--webroot-path`` should be pointed to your
+               ``public_html``). Currently, when multiple domains are specified
+               (`-d`), they must all use the same web root path.
+manual     Y N Hidden from standard UI, use with ``--a manual``. Requires to
+               copy and paste commands into a new terminal session. Allows to
+               run client on machine different than target webserver, e.g. your
+               laptop.
+apache     Y Y Alpha - might break stuff, so be careful. Support for
```
pde: "might break stuff, so be careful" -> "automates Apache installation, works fairly well but Debian only".  Move Apache up to position 2 I think.

pde:
Looks good aside from the two things flagged in comments.

docs/using.rst
kuba: Fixed.

docs/using.rst
kuba: I don't actually agree with this comment. I think narrative introduction is necessary for reader to digest following material properly. However, I don't want to hold off this PR anymore, so I move this paragraph to a section at the bottom: "Comparison of different methods".

kuba:
Thanks for review, @pde. All comments were addressed, ready to merge.

bmw:
@kuba, this is great. Good job. @pde's comments have been resolved and this LGTM. Merging.

PR #1124, opened 2015-10-24
Ask for email unless --allow-unsafe-registration

### LINKED ISSUE #414: Client should strongly encourage use of an email address

miquelruiz:
Add new option that explicitly allows to not provide an email.
Fixes #414

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -127,7 +127,7 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None and not args.allow_unsafe_registration:
+            if args.email is None and not args.register_unsafely_without_email:
                 args.email = display_ops.get_email()
             else:
                 args.email = None
@@ -783,7 +783,7 @@ def prepare_and_parse_args(plugins, args):
         None, "-t", "--text", dest="text_mode", action="store_true",
         help="Use the text output instead of the curses UI.")
     helpful.add(
-        None, "--allow-unsafe-registration", action="store_true",
+        None, "--register-unsafely-without-email", action="store_true",
         help="Specifying this flag enables registering an account with no "
              "email address. This is strongly discouraged, because in the "
              "event of key loss or account compromise you will irrevocably "
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -127,9 +127,9 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None:
+            if args.email is None and not args.allow_unsafe_registration:
                 args.email = display_ops.get_email()
-            if not args.email:  # get_email might return ""
+            else:
                 args.email = None
 
             def _tos_cb(regr):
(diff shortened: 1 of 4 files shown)


letsencrypt/cli.py
```diff
@@ -127,9 +127,9 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None:
+            if args.email is None and not args.allow_unsafe_registration:
                 args.email = display_ops.get_email()
-            if not args.email:  # get_email might return ""
```
kuba: i think that removing this line is no good - see the comment

letsencrypt/cli.py
```diff
@@ -127,9 +127,9 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None:
+            if args.email is None and not args.allow_unsafe_registration:
```
kuba: so basically `--alow-unsafe-registration` will never prompt for email? that doesn't sound good

pde: I think this is fine, though perhaps the flag should have a more accurate name, such as `--register-unsafely-without-email`.  The purpose of the flag is either use by folks who've decided they really don't want to provide an email, or automation in environments contact-based recovery is not in use.

pde:
Thanks for the PR Miquel!  It looks like there are some unit test failures to be fixed.

Also, the ncurses UI looks a little awkward with the double parentheses and variable spacing; perhaps leave the previous question and parentheses then two newlines and say "If you really want to skip this, run the client with --register-unsafely-without-email".

miquelruiz:
I've renamed the option and added some checks to ensure it exits if no email is provided.
I've run the unit tests in master and they seem to be failing too. Is there anything I can do?
I could try to fix them, but doesn't look like something easy for me.
.

pde:
One reason the tests are failing is that the following command is failling:

```
sudo letsencrypt -t --email foo@gmail.com --agree-dev-preview -vvvv --debug -d example.com
```
I get:
```
...
Error: No email was provided and --register-unsafely-without-email was not present.
2015-10-29 19:10:06,697:DEBUG:letsencrypt.cli:Exiting abnormally:
  File "venv/bin/letsencrypt", line 9, in <module>
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/letsencrypt/cli.py", line 1113, in main
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/letsencrypt/cli.py", line 434, in run
    le_client = _init_le_client(args, config, authenticator, installer)
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/letsencrypt/cli.py", line 160, in _init_le_client
    acc, acme = _determine_account(args, config)
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/letsencrypt/cli.py", line 151, in _determine_account
    "Unable to register an account with ACME server")
Error: Unable to register an account with ACME server
...
```

bmw:
Updating title with `[needs revision]`. Please remove when this is ready for review @miquelruiz.

pde:
This has now become #1524.

PR #1130, opened 2015-10-25
Remove confusing short flags

kuba:


bmw:
What? You don't want to make the next [tar](https://xkcd.com/1168/)?

I agree. Merging.

PR #1131, opened 2015-10-25
Point out required dependencies

Duesentrieb-Incorporation:
Since this is used in the beta-program only, a short hint in the documentation should be sufficient.
Otherwise the dependencies should be checked by the script.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -25,6 +25,12 @@ above method instead.
 Installation and Usage
 ======================
 
+Make sure you have the following dependencies installed on your system:
+
+- virtualenv
+- python2-libaugeas
+- dialog
+
 To install and run the client you just need to type:
 
 .. code-block:: shell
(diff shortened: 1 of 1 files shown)


kuba:
All dependencies are already being installed in `letsencrypt-auto` script. Did you follow instructions at http://letsencrypt.readthedocs.org/en/latest/using.html?

You also reverted https://github.com/letsencrypt/letsencrypt/pull/1111, which should take place.

This also conflicts with #1121.

Duesentrieb-Incorporation:
Yes, I followed the instructions. I am using Manjaro Linux (an Arch-based distribution), but it's missing the /etc/arch-release file, so `[ -f /etc/arch-release ]` did not work. Don't know why I missed the error-dialog.

Since Manjaro is based on Arch Linux, it's possible to add another if-clause `if uname -a | grep -i "Manjaro"` and execute the existing $BOOTSTRAP/archlinux.sh. Should Manjaro be supported in the installer, too?

PR #1133, opened 2015-10-25
Added support for Manjaro Linux

Duesentrieb-Incorporation:
The gcc package can conflict with gcc-multilib (same for the Arch-Installer). But there is probably no automatic solution to this.


letsencrypt-auto
```diff
@@ -41,6 +41,9 @@ then
   elif [ -f /etc/arch-release ] ; then
     echo "Bootstrapping dependencies for Archlinux..."
     $SUDO $BOOTSTRAP/archlinux.sh
+  elif [ -f /etc/manjaro-release ] ; then
+    echo "Bootstrapping dependencies for Manjaro linux..."
+    $SUDO $BOOTSTRAP/archlinux.sh
```
kuba: So... you created `manjaro.sh`, but you don't use it?

BTW I'd suggest that you follow the same approach as we have for Debian-derivatives: `debian.sh` and `ubuntu.sh` are symlinks to `_deb_common.sh`. In case something diverges we'll separate the files

Duesentrieb-Incorporation:
Oh man... thanks for being so patient -- I am not very experienced using GitHub yet.

Your approach seems better. I will do my best...

bmw:
Very nice @Duesentrieb-Incorporation. Thanks for the PR. LGTM.

PR #1135, opened 2015-10-25
Remove AAAA record check suggestion

### LINKED ISSUE #1134: On verification failure don't suggest to check AAAA
Boulder doesn't support IPv6 validation and hence any suggestions to check AAAA should probably be removed not to confuse people.

Thanks @rolandshoemaker for reporting!

ArchimedesPi:
Boulder doesn't support ipv6 validation, so don't confuse people
by asking them to check their AAAA record(s). Fixes #1134.


bmw:
LGTM

Issue #657, opened 2015-08-07
Rename the auth command


sfahl:
Rename the auth command to something more meaningful. Auth might make users expect to have to create an account or to login to a service provider. Make it certonly.

Issue #932, opened 2015-10-09
CLI help expects apache and nginx plugins to be installed


kuba:
In a similar vein to #919, `--help` should not hardcode information about Apache and nginx plugins, as these might not be installed at all. You could detect presence of plugins using `pkg_resources`, @pde.

kuba:
Actually, no need to go as deep as `pkg_resources`, just look at the, already initialized, `PluginsRegistry`

PR #1139, opened 2015-10-26
New verbs && plugin help changes if the plugins are not installed.

pde:
In #657 UX researchers noted that the "auth" command is deeply confusing; systems administrators expected it to be some kind of login command rather than the verb that gets you a cert.  So this PR makes "certonly" an alias for it, and makes that form the one that is suggested to users in the help.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -377,7 +377,7 @@ def choose_configurator_plugins(args, config, plugins, verb):
 
     # Which plugins do we need?
     need_inst = need_auth = (verb == "run")
-    if verb in ("auth","certonly"):
+    if verb in ("auth", "certonly"):
         need_auth = True
     if verb == "install":
         need_inst = True
@@ -669,7 +669,12 @@ class HelpfulArgumentParser(object):
 
         for i, token in enumerate(self.args):
             if token in self.VERBS:
-                self.verb = token
+                verb = token
+                if verb == "certonly":
+                    verb = "auth"
+                if verb == "everything":
+                    verb = "run"
+                self.verb = verb
                 self.args.pop(i)
                 return
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -87,7 +87,6 @@ More detailed help:
 
 def usage_strings(plugins):
     """Make usage strings late so that plugins can be initialised late"""
-    print plugins
     if "nginx" in plugins:
         nginx_doc = "--nginx           Use the Nginx plugin for authentication & installation"
     else:
@@ -391,7 +390,7 @@ def choose_configurator_plugins(args, config, plugins, verb):
 
     # Which plugins do we need?
     need_inst = need_auth = (verb == "run")
-    if verb == "auth":
+    if verb == "certonly":
         need_auth = True
     if verb == "install":
         need_inst = True
@@ -461,7 +460,7 @@ def run(args, config, plugins):  # pylint: disable=too-many-branches,too-many-lo
         display_ops.success_renewal(domains)
 
 
-def auth(args, config, plugins):
+def obtaincert(args, config, plugins):
     """Authenticate & obtain cert, but do not install it."""
 
     if args.domains is not None and args.csr is not None:
@@ -471,7 +470,7 @@ def auth(args, config, plugins):
 
     try:
         # installers are used in auth mode to determine domain names
-        installer, authenticator = choose_configurator_plugins(args, config, plugins, "auth")
+        installer, authenticator = choose_configurator_plugins(args, config, plugins, "certonly")
     except PluginSelectionError, e:
         return e.message
 
@@ -495,7 +494,7 @@ def install(args, config, plugins):
     # XXX: Update for renewer/RenewableCert
 
     try:
-        installer, _ = choose_configurator_plugins(args, config, plugins, "auth")
+        installer, _ = choose_configurator_plugins(args, config, plugins, "certonly")
     except PluginSelectionError, e:
         return e.message
 
@@ -623,7 +622,7 @@ class HelpfulArgumentParser(object):
     """
 
     # Maps verbs/subcommands to the functions that implement them
-    VERBS = {"auth": auth, "certonly": auth, "config_changes": config_changes,
+    VERBS = {"auth": obtaincert, "certonly": obtaincert, "config_changes": config_changes,
              "install": install, "plugins": plugins_cmd,
              "revoke": revoke, "rollback": rollback, "run": run}
 
@@ -685,8 +684,8 @@ class HelpfulArgumentParser(object):
         for i, token in enumerate(self.args):
             if token in self.VERBS:
                 verb = token
-                if verb == "certonly":
-                    verb = "auth"
+                if verb == "auth":
+                    verb = "certonly"
                 if verb == "everything":
                     verb = "run"
                 self.verb = verb
@@ -774,8 +773,8 @@ class HelpfulArgumentParser(object):
         """
         # topics maps each topic to whether it should be documented by
         # argparse on the command line
-        if chosen_topic == "certonly":
-            chosen_topic = "auth"
+        if chosen_topic == "auth":
+            chosen_topic = "certonly"
         if chosen_topic == "everything":
             chosen_topic = "run"
         if chosen_topic == "all":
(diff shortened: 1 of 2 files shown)


pde:
This now also has a fix for #932 , which involves further tweaking of plugin-related help.

letsencrypt/cli.py
```diff
@@ -1050,7 +1072,7 @@ def main(cli_args=sys.argv[1:]):
     sys.excepthook = functools.partial(_handle_exception, args=None)
 
     # note: arg parser internally handles --help (and exits afterwards)
-    plugins = plugins_disco.PluginsRegistry.find_all()
+    global plugins
```
kuba: Please don't use `global`, just skip this line, it's not necessary.

letsencrypt/cli.py
```diff
@@ -64,25 +64,38 @@
 
 """
 
+plugins = plugins_disco.PluginsRegistry.find_all()
```
kuba: This is going to be really hard to mock in the unittests... that are missing.

letsencrypt/cli.py
```diff
@@ -669,7 +682,12 @@ def determine_verb(self):
 
         for i, token in enumerate(self.args):
             if token in self.VERBS:
-                self.verb = token
+                verb = token
+                if verb == "certonly":
+                    verb = "auth"
```
kuba: I don't like having aliases. Let's just use `certonly` and get rid of `auth` altogether. It makes the code complicated, and is harder to grasp from UX point of view.

letsencrypt/cli.py
kuba: Also, there should be a `TODO` comment somewhere in the existing code, that says we're missing logging for plugins... because they are initialized before logging is setup. Moving this even higher in the execution makes it even harder to fix the problem.

letsencrypt/cli.py
pde: I'm fine with that plan in the medium term, I just wanted this to be a minimal code change before launch rather than a far-reaching one.

letsencrypt/cli.py
```diff
@@ -64,26 +64,40 @@
...
 
-   all, apache, automation, manual, nginx, paths, security, testing, or any of
-   the subcommands
+   all, automation, paths, security, testing, or any of the subcommands or
+   plugins (certonly, install, nginx, apache, standalone, etc)
 """
 
+def usage_strings(plugins):
+    """Make usage strings late so that plugins can be initialised late"""
+    print plugins
```
bmw: I'm assuming this is a debugging statement that should be removed.

letsencrypt/cli.py
bmw: I personally don't have a strong preference on aliases. (I secretly want to make a verb "allthethings" so we can say `letsencrypt allthethings`).

With that said, how much extra work would it be @pde to flip this logic here and have "certonly" be the main version of the verb? I could be wrong, but I don't think it'd be that hard and it'd get rid of awkward behavior like:
```
$ letsencrypt --help certonly       

usage: 

optional arguments:
                        config file path (default: None)

auth:
  Options for modifying how a cert is obtained

  --csr CSR             Path to a Certificate Signing Request (CSR) in DER
                        format; note that the .csr file *must* contain a
                        Subject Alternative Name field for each domain you
                        want certified. (default: None)
```

Where's my help on `certonly`? :(

pde:
OK, I've switched things around so that "certonly" is the real verb and "auth" is an alias for that.

acme/acme/challenges_test.py
```diff
@@ -359,9 +359,12 @@ def test_simple_verify(self, mock_verify_cert):
             cert=mock.sentinel.cert))
         mock_verify_cert.assert_called_once_with(self.msg, mock.sentinel.cert)
 
-    def test_simple_verify_false_on_probe_error(self):
+    @mock.patch('acme.challenges.socket.gethostbyname')
+    def test_simple_verify_false_on_probe_error(self, mock_gethostbyname):
         chall = mock.Mock()
         chall.probe_cert.side_effect = errors.Error
+
+        mock_gethostbyname.return_value = '127.0.0.2'
```
kuba: passing `host='127.0.0.2'` to `simple_verify` is easier than mocking :)

kuba: That being said, I don't understand why you need this code. `probe_cert` is the only function in the entire module that does `gethostbyname` and it's already mocked in `chall.probe_cert`...

bmw: Checking out the code locally and undoing all changes in this file, tests work fine locally. Why were these changes made?

letsencrypt/cli.py
```diff
@@ -609,7 +622,7 @@ class HelpfulArgumentParser(object):
     """
 
     # Maps verbs/subcommands to the functions that implement them
-    VERBS = {"auth": auth, "config_changes": config_changes,
+    VERBS = {"auth": obtaincert, "certonly": obtaincert, "config_changes": config_changes,
```
bmw: nit: The alias `auth` is included here but not `everything`. I think we should have the same behavior for both aliases. The only behavior that I'm aware of that is affected here is whether you can get help using the alias (e.g. `letsencrypt --help auth`).

I don't have a strong opinion, but I think aliases should not be included. The aliases are not shown in the help output and the distinction between `letsencrypt --help all` and `letsencrypt --help everything` is somewhat confusing.

letsencrypt/cli.py
```diff
@@ -56,7 +56,7 @@
 the cert. Major SUBCOMMANDS are:
 
   (default) run        Obtain & install a cert in your current webserver
-  auth                 Authenticate & obtain cert, but do not install it
+  certonly             Obtain cert, but do not install it (aka "auth")
```
bmw: UI/UX question I don't know the answer to: Do we want to include `auth` here anymore?

letsencrypt/cli.py
pde: Thanks for catching that.

letsencrypt/cli.py
pde: Yes, for at least a while, because a lot of people have heard about "auth" and should experience some continuity.

acme/acme/challenges_test.py
pde: Trying to run the unit tests while offline, I got the following error:
```
ERROR: test_simple_verify_false_on_probe_error (acme.challenges_test.DVSNIResponseTest)
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/acme/acme/challenges_test.py", line 366, in test_simple_verify_false_on_probe_error
    self.chall, self.domain, self.key.public_key()))
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/acme/acme/challenges.py", line 443, in simple_verify
    cert = self.probe_cert(domain=domain, **kwargs)
  File "/home/pde/eff/chocolate/chocolate/letsencrypt/acme/acme/challenges.py", line 383, in probe_cert
    host = socket.gethostbyname(domain)
gaierror: [Errno -3] Temporary failure in name resolution
-------------------- >> begin captured logging << --------------------
```
The above change fixed that issue.

pde: I'm fine factoring it out into another PR if either of you would like that :)

bmw: I think another PR is a better approach, but I honestly don't really care too much since only a one line change is necessary. See my comments on #1183 or Kuba's original comment in this thread.

kuba: please remove it from this PR and let's continue the thread on #1183

kuba: I also explain why the proposed change is wrong in https://github.com/letsencrypt/letsencrypt/pull/1183#discussion_r43352442

pde: OK, rebasing to remove that commit from this branch.

bmw:
LGTM.

I don't like how new functions like `usage_strings` don't have full test coverage, however, I tested by hand all changed lines without cover. This was a time suck, but I'd hate for us to make a PyPI release and have someone get something like a syntax error. I'm happy to figure out a good way to mock out the `plugin_disco` stuff sometime.

I also don't like the approach of including in the help that a plugin is not installed. I would personally prefer to see no output about a plugin if I haven't installed it. As the number of plugins grows this will be more important and I'm unsure that we should be doing anything special for the Apache and Nginx plugins that we don't do for 3rd party plugins.

PR #1144, opened 2015-10-26
use more accurate method to calculate time interval

devnsec-com:
Change the `should_autodeploy()` function and `should_autorenew()` function to use more accurate method to calculate time interval, considering of timezone, DST, leap year, and different month lengths, etc.

This PR replaces #1120

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -2,7 +2,6 @@
 import datetime
 import os
 import re
-import time
 
 import configobj
 import parsedatetime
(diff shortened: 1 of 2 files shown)


devnsec-com:
I think there's something wrong with the Travis CI, I'm closing and re-opening this PR to trigger the rebuild.

pde:
This looks good to me.  @schoen since this is your code, do you want to take a look and merge?

schoen:
@devnsec-com, thank you for addressing this! I was aware that this kind of inconsistency could be a problem in some cases but I think your approach is a good internally-consistent one. (There might be some users who would prefer a different interpretation, but maybe we'll find out from reports from the field. "No, that's not what I meant by 'month'!")

schoen:
I wonder if people's intuition would prefer `<=` instead of `<` (the `<` is from my original code). For example, if we say that the renewal interval is "1 week" and the renewer is run precisely 1 week prior to expiry, should renewal happen?

I believe with both versions of the code this can only happen if the code is run precisely at the second in question, which would probably only happen routinely as a result of a user scheduling an `at` job to run the renewal, or simply by very bad luck. :-) If we say "1 month" and it's run at any other time during the day, hour, or minute that falls 1 month prior to expiry, it will choose to renew.

PR #1145, opened 2015-10-26
 platform support for debian gnukfreebsd

jepler:
'psutil' is not available on debian kfreebsd.  Therefore, do not
require it there and deal gracefully when it is missing at runtime.
Apparently, this only prevents letsencrypt from diagnosing certain
"port already in use" type errors.

Note: I did run nox, but only on debian gnukfreebsd, not GNU/Linux.

Signed-off-by: Jeff Epler <jepler@unpythonic.net>

### DIFF
--- a/letsencrypt/plugins/util.py
+++ b/letsencrypt/plugins/util.py
@@ -2,7 +2,14 @@
 import logging
 import socket
 
-import psutil
+try:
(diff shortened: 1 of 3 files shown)


jepler:
clearly this patch needs additional work.

PR #1152, opened 2015-10-27
Remove duplicated 'that' in the Plugins section

venyii:


bmw:
LGTM. Thanks @venyii.

PR #1159, opened 2015-10-27
Respect pre-installed gcc package on Arch Linux

pritambaral:
Both `gcc-multilib` and `gcc` packages provide 64-bit gcc on x86_64 installations, so there is no harm if `gcc-multilib` is already installed. Making pacman install `gcc` when `gcc-multilib` is already installed prompts pacman to ask and uninstall `gcc-multilib` so it can be replaced by `gcc`.


pde:
Looks OK.

PR #1160, opened 2015-10-27
Standalone verifies ports for supported challenges only (fixes #1149).

### LINKED ISSUE #1149: Standalone should verify ports for suported challenges only
I ran the following commands and both complained that nginx was already using port 80.
`./letsencrypt-auto --standalone --dvsni-port 63443 --server https://acme-v01.api.letsencrypt.org/directory auth`
`./letsencrypt-auto --standalone --dvsni-port 63443 --server https://acme-v01.api.letsencrypt.org/directory --standalone-supported-challenges dvsni auth`

I was under the impression that using `--dvsni-port` would force dvsni only.

I had to supply a fake port with `--simple-http-port 9999` with the second commad to get a certificate.

I'm using haproxy in the front to redirect *.acme.invalid to port 63443.

kuba:


pde:
LGTM. Though arguably given boulder's current behaviour we would do better by seeing if *any* of the supported ports are available, rather than all of them, and then pruning the set of challenges we support to match the ports.

pde:
(That's a nice for 1.0 enhancement though...)

bmw:
I agree with @pde that choosing the challenge based on the available ports would be a nice enhancement, but is not necessary for launch. I created #1164 to track the issue.

LGTM. Thanks for fixing this so quickly after it was reported @kuba.

PR #1163, opened 2015-10-27
Add gitattributes file to mark bat file as CRLF

thomwiggers:
Gitattributes files can be used to mark files as crlf, which is useful for the `.bat` files in case people use eol=lf in their git config.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+* text=auto eol=lf
+
+# special files
+*.bat text eol=crlf
+*.jpeg binary
+*.jpg binary
+*.png binary


pde:
LGTM

pde:
@thomwiggers Interstingly, this pull request seems to have permanently broken git bisection in our tree :(

Fortunately I've been able to `git bisect` by rebasing an alternative history where it never happened.

thomwiggers:
Ah, shit, I should've dos2unixed `docs/make.bat` when I proposed this change. Weird, when I got a checkout of this repo on one of my machines git already had started complaining about changes in the bat file even though I hadn't touched it: this is why I assumed it was already using CRLF endlines...

`git update-index --assume-unchanged docs/make.bat` should allow to work around it for purposes such as `bisect`.

PR #1167, opened 2015-10-28
Always use absolute paths [needs revision]

### LINKED ISSUE #836: CLI directory options should take absolute path
I was attempting to determine if #443, was still a problem.

I ran letsencrypt --work-dir ../Downloads -vvvv run

I was left with the following error...
```
AH00526: Syntax error on line 10 of /etc/apache2/le_dvsni_cert_challenge.conf:
SSLCertificateFile: file '/etc/Downloads/OwYVJOOHVxWaHWNpNUgNDy4TIafggji3RdP0SymKozo.crt' does not exist or is empty
Action 'configtest' failed.
```

I thought it to be a peculiar place to be looking for a cert.

The config file is populated with...
```
def get_cert_path(self, achall):
        """Returns standardized name for challenge certificate.
        :param achall: Annotated DVSNI challenge.
        :type achall: :class:`letsencrypt.achallenges.DVSNI`
        :returns: certificate file name
        :rtype: str
        """
        return os.path.join(self.configurator.config.work_dir,
                            achall.chall.encode("token") + ".crt")
```

Thus, leading me to the conclusion that we do not handle our directories appropriately from the command-line.  I think it would be much easier if all of the work_dir, config_dir, and logs_dir were guaranteed to be an absolute path.

Providing the full path off the command-line did solve the problem.

Can we find a place to normalize all paths?

bmw:
fixes #836

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -466,8 +466,9 @@ def auth(args, config, plugins):
 
     # This is a special case; cert and chain are simply saved
     if args.csr is not None:
+        csr = read_file(args.csr)
         certr, chain = le_client.obtain_certificate_from_csr(le_util.CSR(
-            file=args.csr[0], data=args.csr[1], form="der"))
+            file=csr[0], data=csr[1], form="der"))
         cert_path, _, cert_fullchain = le_client.save_certificate(
             certr, chain, args.cert_path, args.chain_path, args.fullchain_path)
         _report_new_cert(cert_path, cert_fullchain)
@@ -497,18 +498,20 @@ def install(args, config, plugins):
 
 def revoke(args, config, unused_plugins):  # TODO: coop with renewal config
     """Revoke a previously obtained certificate."""
+    cert = read_file(args.cert_path)
     if args.key_path is not None:  # revocation by cert key
+        key = read_file(args.key_path)
         logger.debug("Revoking %s using cert key %s",
-                     args.cert_path[0], args.key_path[0])
+                     cert[0], key[0])
         acme = acme_client.Client(
-            config.server, key=jose.JWK.load(args.key_path[1]))
+            config.server, key=jose.JWK.load(key[1]))
     else:  # revocation by account key
-        logger.debug("Revoking %s using Account Key", args.cert_path[0])
+        logger.debug("Revoking %s using Account Key", cert[0])
         acc, _ = _determine_account(args, config)
         # pylint: disable=protected-access
         acme = client._acme_from_config_key(config, acc.key)
     acme.revoke(jose.ComparableX509(crypto_util.pyopenssl_load_certificate(
-        args.cert_path[1])[0]))
+        cert[1])[0]))
 
 
 def rollback(args, config, plugins):
@@ -560,13 +563,10 @@ def read_file(filename, mode="rb"):
     :returns: A tuple of filename and its contents
     :rtype: tuple
 
-    :raises argparse.ArgumentTypeError: File does not exist or is not readable.
+    :raises IOError: File does not exist or is not readable.
 
     """
-    try:
-        return filename, open(filename, mode).read()
-    except IOError as exc:
-        raise argparse.ArgumentTypeError(exc.strerror)
+    return filename, open(filename, mode).read()
 
 
 def flag_default(name):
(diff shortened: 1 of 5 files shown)


bmw:
Blocked on #1168.

letsencrypt/cli.py
```diff
@@ -867,7 +867,7 @@ def _create_subparsers(helpful):
     helpful.add_group("plugins", description="Plugin options")
 
     helpful.add("auth",
-                "--csr", type=read_file,
+                "--csr",
```
kuba: i like this change

letsencrypt/cli.py
```diff
@@ -896,14 +896,12 @@ def _paths_parser(helpful):
...
-        add("paths", "--cert-path", type=read_file, required=True, help=cph)
     else:
-        add("paths", "--cert-path", help=cph, required=(verb == "install"))
+        add("paths", "--cert-path", help=cph,
+            required=(verb in ["install", "revoke"]))
 
     # revoke --key-path reads a file, install --key-path takes a string
-    add("paths", "--key-path", type=((verb == "revoke" and read_file) or str),
-        required=(verb == "install"),
+    add("paths", "--key-path", required=(verb == "install"),
```
kuba: oh wow, thanks for fixing the convoluted code

kuba: @pde `X and Y or Z` trick should not be used in Python: we have `Y if X else Z` if you really need it

letsencrypt/cli.py
```diff
@@ -466,8 +466,9 @@ def auth(args, config, plugins):
 
     # This is a special case; cert and chain are simply saved
     if args.csr is not None:
+        csr = read_file(args.csr)
```
kuba: exception is not caught, here and in other instances

kuba: Also: can we use `csr_path, csr_contents = read_file(...)` to make code more readable?

letsencrypt/cli.py
```diff
@@ -560,13 +563,10 @@ def read_file(filename, mode="rb"):
...
 
-    :raises argparse.ArgumentTypeError: File does not exist or is not readable.
+    :raises IOError: File does not exist or is not readable.
 
     """
-    try:
-        return filename, open(filename, mode).read()
-    except IOError as exc:
-        raise argparse.ArgumentTypeError(exc.strerror)
+    return filename, open(filename, mode).read()
```
kuba: While we're at it, I think we should do `os.path.abspath(filename)`

letsencrypt/cli.py
```diff
@@ -560,13 +563,10 @@ def read_file(filename, mode="rb"):
     :returns: A tuple of filename and its contents
     :rtype: tuple
 
-    :raises argparse.ArgumentTypeError: File does not exist or is not readable.
+    :raises IOError: File does not exist or is not readable.
```
kuba: can we add unittest for `read_file`, including check for `IOError`?

letsencrypt/interfaces.py
```diff
@@ -199,23 +199,25 @@ class IConfig(zope.interface.Interface):
         "Email used for registration and recovery contact.")
     rsa_key_size = zope.interface.Attribute("Size of the RSA key.")
 
-    config_dir = zope.interface.Attribute("Configuration directory.")
-    work_dir = zope.interface.Attribute("Working directory.")
+    config_dir = zope.interface.Attribute(
+        "Absolute path to the configuration directory.")
+    work_dir = zope.interface.Attribute(
+        "Absolute path to the working directory.")
```
kuba: Slight problem here. While those descriptions are perfectly OK within this file, they are also used for CLI help generation. In CLI, however, user does not have to use absolute paths. I think that we should stop using `flag_help` in case of all those "absolute path" flags.

letsencrypt/cli.py
bmw: While I agree with `csr_path, csr_contents = read_file(...) `, I am not sure if the exception should be caught. In almost all cases, if we don't have these files we cannot continue and have to exit with an error. We could wrap these errors in our own error type or have the verb functions return the error message, but printing an error message and exiting is the proper behavior here.

letsencrypt/cli.py
bmw: I actually want to go in the opposite direction. If do the work upfront to ensure absolute paths are used, we can go through and remove some of the 26 calls to `os.path.abspath` between the `letsencrypt`, `letsencrypt-apache`, `letsencrypt-nginx` modules.

bmw:
Superseded by #1483.

PR #1168, opened 2015-10-28
Install listenbuddy

bmw:
Boulder added a dependency on small program named `listenbuddy` for testing in letsencrypt/boulder#1033. This PR adds the installation of this dependency on the client side.


kuba:
LGTM

Issue #1085, opened 2015-10-22
Hanging threads (acme.standalone)


kuba:

kuba:
Steps to reproduce: `while python acme/acme/standalone_test.py; do echo -n .; done` and wait for hang. Interestingly, stracing to the child thread unblocks it:

```
jakub@iks:~/dev/letsencrypt/letsencrypt$ strace -p 12897
Process 12897 attached
futex(0x1ef17a0, FUTEX_WAIT_PRIVATE, 0, NULL^CProcess 12897 detached
 <detached ...>
jakub@iks:~/dev/letsencrypt/letsencrypt$ strace -p 12903
Process 12903 attached
select(4, [3], [], [], NULL)            = -1 EBADF (Bad file descriptor)
stat("/usr/lib/python2.7/threading.py", {st_mode=S_IFREG|0644, st_size=47336, ...}) = 0
open("/usr/lib/python2.7/threading.py", O_RDONLY) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=47336, ...}) = 0
fstat(3, {st_mode=S_IFREG|0644, st_size=47336, ...}) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f65d9f98000
read(3, "\"\"\"Thread module emulating a sub"..., 8192) = 8192
read(3, "ore threads to wait until they a"..., 4096) = 4096
read(3, "                endtime = _time("..., 4096) = 4096
read(3, " with blocking set to false, do "..., 4096) = 4096
read(3, "f.__cond.__init__(Lock())\n\n    d"..., 4096) = 4096
read(3, "initialized = True\n        # sys"..., 4096) = 4096
read(3, "       if __debug__:\n           "..., 4096) = 4096
read(3, "rwrites the (if any) thread curr"..., 4096) = 4096
read(3, "def ident(self):\n        \"\"\"Thre"..., 4096) = 4096
read(3, "# Dummy thread class to represen"..., 4096) = 4096
read(3, "f-test code\n\ndef _test():\n\n    c"..., 4096) = 2280
read(3, "", 4096)                       = 0
close(3)                                = 0
munmap(0x7f65d9f98000, 4096)            = 0
stat("/usr/lib/python2.7/threading.py", {st_mode=S_IFREG|0644, st_size=47336, ...}) = 0
[...]

PR #1170, opened 2015-10-28
No newlines in JWK thumbprint (fixes #1165)

### LINKED ISSUE #1165: jwk thumbprint is mismade
The JWK.thumbprint method isn't quite right. 

I think the problem is that the `json.dumps` the thumbprint method uses can mis-order the keys in the JSON hash. There is an order that [RFC7638](https://tools.ietf.org/html/rfc7638#section-3.3) says the keys must be in. The go-jose code does that with this line of code: https://github.com/square/go-jose/blob/4ddd71883fa547d37fbf598071f04512d8bafee3/jwk.go#L155

Going off the go-jose test vectors, this RSA public key JWK: 

    {
        "kty": "RSA",
        "kid": "bilbo.baggins@hobbiton.example",
        "use": "sig",
        "n": "n4EPtAOCc9AlkeQHPzHStgAbgs7bTZLwUBZdR8_KuKPEHLd4rHVTeT-O-XV2jRojdNhxJWTDvNd7nqQ0VEiZQHz_AJmSCpMaJMRBSFKrKb2wqVwGU_NsYOYL-QtiWN2lbzcEe6XC0dApr5ydQLrHqkHHig3RBordaZ6Aj-oBHqFEHYpPe7Tpe-OfVfHd1E6cS6M1FZcD1NNLYD5lFHpPI9bTwJlsde3uhGqC0ZCuEHg8lhzwOHrtIQbS0FVbb9k3-tVTU4fg_3L_vniUFAKwuCLqKnS2BYwdq_mzSnbLY7h_qixoR7jig3__kRhuaxwUkRz5iaiQkqgc5gHdrNP5zw",
     "e": "AQAB"
    }

must be thumbprinted into `f63838e96077ad1fc01c3f8405774dedc0641f558ebb4b40dccf5f9b6d66a932`.

In case I fat-fingered the key or its hash, here they are in the go-jose tests: https://github.com/square/go-jose/blob/4ddd71883fa547d37fbf598071f04512d8bafee3/jwk_test.go#L331-L344 and https://github.com/square/go-jose/blob/4ddd71883fa547d37fbf598071f04512d8bafee3/jwk_test.go#L384

kuba:


kuba:
Hit by #1085 (https://gist.github.com/kuba/62d8a00f5bb7371405a1).

bmw:
LGTM. I tested this with the example given in [RFC 7638](https://tools.ietf.org/html/draft-ietf-jose-jwk-thumbprint-08#section-3.1) and got the correct result.

Issue #1140, opened 2015-10-26
Fix intersphinx links


kuba:
`make -C docs clean html` spits out the following:

```
WARNING: intersphinx inventory 'https://acme-python.readthedocs.org/objects.inv' not fetchable due to <class 'urllib2.HTTPError'>: HTTP Error 404: OK
```

We need to update `conf.py` (for all packages) to use correct links.

cowlicks:
I think this is closeable. I get a few warnings when I build the docs, but not the ones mentioned.
```

/home/blake/github/letsencrypt/letsencrypt/client.py:docstring of letsencrypt.client.Client.apply_enhancement:4: WARNING: Field list ends without a blank line; unexpected unindent.
/home/blake/github/letsencrypt/letsencrypt/client.py:docstring of letsencrypt.client.Client.apply_enhancement:7: WARNING: Field list ends without a blank line; unexpected unindent.
/home/blake/github/letsencrypt/letsencrypt/client.py:docstring of letsencrypt.client.Client.apply_enhancement:10: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/blake/github/letsencrypt/letsencrypt/client.py:docstring of letsencrypt.client.Client.apply_enhancement:11: WARNING: Field list ends without a blank line; unexpected unindent.
../README.rst:96: WARNING: nonlocal image URI found: https://travis-ci.org/letsencrypt/letsencrypt.svg?branch=master
../README.rst:96: WARNING: nonlocal image URI found: https://coveralls.io/repos/letsencrypt/letsencrypt/badge.svg?branch=master
../README.rst:96: WARNING: nonlocal image URI found: https://readthedocs.org/projects/letsencrypt/badge/
../README.rst:96: WARNING: nonlocal image URI found: https://quay.io/repository/letsencrypt/letsencrypt/status
../README.rst:100: WARNING: nonlocal image URI found: https://travis-ci.org/letsencrypt/letsencrypt.svg?branch=master
../README.rst:104: WARNING: nonlocal image URI found: https://coveralls.io/repos/letsencrypt/letsencrypt/badge.svg?branch=master
../README.rst:108: WARNING: nonlocal image URI found: https://readthedocs.org/projects/letsencrypt/badge/
../README.rst:112: WARNING: nonlocal image URI found: https://quay.io/repository/letsencrypt/letsencrypt/status
```

bmw:
[...]

PR #1171, opened 2015-10-28
Fix readthedocs Intersphinx URLs

chk1:
This fixes the Warning about missing/incorrect Intersphinx links from #1140 

Fixed according to the example here: http://read-the-docs.readthedocs.org/en/slumber_refactor/features.html#intersphinx-support

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -315,5 +315,5 @@ texinfo_documents = [
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/', None),
-    'acme': ('https://acme-python.readthedocs.org', None),
+    'acme': ('https://acme-python.readthedocs.org/en/latest/', None),
 }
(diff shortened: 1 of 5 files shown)


bmw:
Hit by letsencrypt/boulder#1033. Kicking Travis.

kuba:
Thanks and welcome to the project, @chk1 :)

While it's a clear improvement, one question still remains (and that's why bug should not be closed upon merging): how do do intersphinx for releases (we don't want to link to latest, but to stable).

LGTM, but please make sure to reopen #1140 after merging.

chk1:
From a quick look at the manual it seems like readthedocs tags releases as 'stable' and a documentation is created accordingly, e.g. http://letsencrypt.readthedocs.org/en/stable/ (currently not all Let's Encrypt projects listed on RTD have this stable tag)

So you could change all `/latest` to `/stable`, but I'm not sure if that's applicable. Maybe someone with a more experience with this can help.

kuba:
Related: https://github.com/letsencrypt/letsencrypt/issues/1097 (problems with "stable" on RTD)

bmw:
LGTM

PR #1172, opened 2015-10-28
Update letsencrypt-auto with CentOS6 SCL for python 2.7

### LINKED ISSUE #1046: letsencrypt-auto doesn't work with Python 2.6
When running letsencrypt-auto I get the following:

```
Updating letsencrypt and virtual environment dependencies...../root/.local/share/letsencrypt/lib/python2.6/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-TVlyY0/ConfigArgParse
/root/.local/share/letsencrypt/lib/python2.6/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
```
I think this relates to the ConfigArgParse problem that @kuba was trying to fix.

linickx:
The attached PR fixes #1046 for CentOS 6 users by setting up python 2.7 using the Software Collection Library that is part of CentOS. (Therefore not relying on 3rd party repo's to upgrade python)

Feedback/comments/suggestions welcome.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -12,6 +12,7 @@ XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
 VENV_NAME="letsencrypt"
 VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
+PYTHON_VER="python2"
 
 if test "`id -u`" -ne "0" ; then
   SUDO=sudo
@@ -46,7 +47,16 @@ then
     $SUDO $BOOTSTRAP/manjaro.sh
   elif [ -f /etc/redhat-release ] ; then
     echo "Bootstrapping dependencies for RedHat-based OSes..."
-    $SUDO $BOOTSTRAP/_rpm_common.sh
+    if [ `lsb_release -is` == "CentOS" ]; then
+        $SUDO $BOOTSTRAP/centos.sh
+	if [ `lsb_release -rs | cut -f1 -d.` -eq 6 ]; then
+		source /opt/rh/python27/enable
+		PYTHON_VER="python2.7"
+		SCL="scl enable python27 "
+	fi
+    else
+        $SUDO $BOOTSTRAP/_rpm_common.sh
+    fi
   elif uname | grep -iq FreeBSD ; then
     echo "Bootstrapping dependencies for FreeBSD..."
     $SUDO $BOOTSTRAP/freebsd.sh
@@ -62,11 +72,12 @@ then
     echo "for more info"
   fi
 
-  echo "Creating virtual environment..."
+
+  echo "Creating virtual environment...$VENV_PATH"
   if [ "$VERBOSE" = 1 ] ; then
-    virtualenv --no-site-packages --python python2 $VENV_PATH
+    virtualenv --no-site-packages --python $PYTHON_VER $VENV_PATH
   else
-    virtualenv --no-site-packages --python python2 $VENV_PATH > /dev/null
+    virtualenv --no-site-packages --python $PYTHON_VER $VENV_PATH > /dev/null
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


letsencrypt-auto
```diff
@@ -99,5 +110,9 @@ fi
 
 # Explain what's about to happen, for the benefit of those getting sudo
 # password prompts...
-echo "Running with virtualenv:" $SUDO $VENV_BIN/letsencrypt "$@"
-$SUDO $VENV_BIN/letsencrypt "$@"
+if [ -z "$SCL" ] ; then
+    echo "Running with virtualenv:" $SUDO $VENV_BIN/letsencrypt "$@"
+    $SUDO $VENV_BIN/letsencrypt "$@"
+else
+    echo "Now run: $SUDO $SCL \"$VENV_BIN/letsencrypt $@\""
```
kuba: letsencrypt-auto should call the wrapped command rather than asking user to copy paste it and run themselves

linickx: Yeah. I agree, however I couldn't figure it out.

My original go had `$SUDO $SCL $VENV_BIN/letsencrypt "$@"` (i.e. without the echo) but it fails. I tried escaping the quotes but to no avail. `$SCL` sets up the correct python ENV variables but something goes amiss when called in the script with SUDO... works fine if the user copy/paste.

Any idea on what I'm missing?

kuba:
What is this SLC. Is that like Debian backports? I think some people might not like the fact we enable this repo for them. I consider this a main showstopper for this PR.

linickx:
> What is this SLC. Is that like Debian backports?
> 

SLC info can be found here:

: https://wiki.centos.org/AdditionalResources/Repositories/SCL <https://wiki.centos.org/AdditionalResources/Repositories/SCL>

… basically allows newer languages to be installed along side the stock ones, in a non-breaking vendor supported way.

There is some upstream scl packaging documentation is here:

: https://jnovy.fedorapeople.org/scl-utils/scl.pdf <https://jnovy.fedorapeople.org/scl-utils/scl.pdf>

> I think some people might not like the fact we enable this repo for them. I consider this a main showstopper for this PR.
> 

Probably.

centminmod:
i'd make it optional or add a check if /usr/bin/python2.7 already exists as some folks like myself use IUS Community repo for python 2.7 and may conflict https://community.letsencrypt.org/t/letsencrypt-webroot-authentication-tested-on-beta-invited-whitelisted-domain/2276/3

centminmod:
@kuba 

>  I think some people might not like the fact we enable this repo for them. I consider this a main showstopper for this PR

indeed... it's probably better to add the various ways centos/rhel 6 users can install python 2.7 via SCL or IUS Community repo in the official documentation and manual/wikis

linickx:
Good suggestions :+1:  I've committed an update. I can squash the commits, if the maintainers/collaborators do think this PR will be accepted.

centminmod:
cheers @linickx maybe related too https://github.com/letsencrypt/letsencrypt/pull/1232

kuba:
You shouldn't need to install Python 2.7 at all. This will be fixed in #1257.

letsencrypt-auto
pde: Fun times.  Try this:

$SUDO $SCL \\"$VENV_BIN/letsencrypt $@\\"
```
The theory is that the double escapes protect the quotes through the current shell, and the sudo layer, reconstructing them for the scl command.  I don't have a Centos system to test on, but I tested in with a shell script that tries to run an analogous command through `sudo bash -c \\"$@\\"`

pde: Maybe if people use `"$@"` rather than `$@` it should be `$SUDO $SCL \\"$VENV_BIN/letsencrypt "$@"\\"`, though I think that just defers expansion of shell arguments until further down the stack?

pde:
@kuba are you saying that landing #1257 will make this PR unnecessary as a whole?  If so I guess there's no point in revising.  Although things like #1278 suggest it might be worth doing anyway?

pde:
@kuba @linickx should this PR be landed, or is it obsolete now?

centminmod:
related https://community.letsencrypt.org/t/install-fails-on-amazon-linux/3708/2

bmw:
Since Python 2.6 support was added to `letsencrypt-auto`, I believe this PR is obsolete. If you disagree, yell at me and I'll reconsider.

PR #1175, opened 2015-10-28
Started working on the OpenBSD port. [needs revision]

smutt:
Bootstrap file seems to work.
Halted work because psutil does not yet support OpenBSD.
https://github.com/giampaolo/psutil/issues/562

### DIFF
--- a/bootstrap/openbsd.sh
+++ b/bootstrap/openbsd.sh
@@ -1,0 +1,9 @@
+#!/bin/sh -xe
+
+pkg_add -aI \
+  git \
+  python-2.7.9p0 \
+  py-virtualenv \
+  augeas \
+  libffi \
+
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -12,6 +12,7 @@ XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
 VENV_NAME="letsencrypt"
 VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
+PYTHON_EXE=python2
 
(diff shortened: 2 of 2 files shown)


kuba:
I don't think it should be pulled it (yet) for a couple of reasons, including:
- list of installed dependencies is incomparably lower that for other systems which suggests that it wouldn't work even if `psutil` problem (NB already reported at https://github.com/letsencrypt/letsencrypt/issues/1174) is fixed
- it will not work due to `psutil` problem, and by pulling this in we commit to OpenBSD support in `letsencrypt-auto`

I would suggest you try to amend developer instructions and scripts (https://github.com/letsencrypt/letsencrypt/blob/v0.0.0.dev20151024/bootstrap/install-deps.sh) instead of user-facing `letsencrypt-auto`.

pde:
Great, let us know when you and @sboyettedh have made enough progress that you'd like us to consider merging things.

bmw:
Adding `[needs revision]` to the title to track this is not ready to be merged. When this changes, update the title and we'll take a look @smutt.

hobarrera:
> Bootstrap file seems to work.

Bootstrap failed for me, and linked to a page that states `sudo` is requiered. An OpenBSD port should use `doas(1)`, since `sudo` has been deprecated by it on OpenBSD (the former will usually be installed, but not the latter).

mpasternacki:
The `sudo` / `doas` depends on version, `doas` appeared in OpenBSD 5.8 which is relatively fresh. My 5.7 system still has `sudo`.

bmw:
Since we have official OpenBSD packaging, I don't think it's necessary to add support for OpenBSD to `letsencrypt-auto` and am closing this issue.

If anyone disagrees, (nicely) yell at me and I'll reconsider.

PR #1177, opened 2015-10-28
Begin wrangling letsencrypt-auto to be FHS-compliant

whyhellobradley:


kuba:
@pde seems to be working on that in https://github.com/letsencrypt/letsencrypt/tree/fhs branch. Why do you submit not your code?

PR #1178, opened 2015-10-28
fix acme.standalone hanging in unit tests

devnsec-com:
This PR fixes issue #1085 

I'm not sure if this is the best approach, but it solves the problem for me.

All suggestions are welcomed.

### DIFF
--- a/acme/acme/standalone_test.py
+++ b/acme/acme/standalone_test.py
@@ -88,8 +88,8 @@ class DVSNIServerTest(unittest.TestCase):
         }
         from acme.standalone import DVSNIServer
         self.server = DVSNIServer(("", 0), certs=self.certs)
-        # pylint: disable=no-member
-        self.thread = threading.Thread(target=self.server.handle_request)
+        # use serve_forever2() instead of handle_request() to avoid hanging
+        self.thread = threading.Thread(target=self.server.serve_forever2)
         self.thread.start()
 
     def tearDown(self):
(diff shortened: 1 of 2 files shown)


acme/acme/standalone_test.py
```diff
@@ -88,8 +88,8 @@ def setUp(self):
         }
         from acme.standalone import DVSNIServer
         self.server = DVSNIServer(("", 0), certs=self.certs)
-        # pylint: disable=no-member
-        self.thread = threading.Thread(target=self.server.handle_request)
+        # use serve_forever2() instead of handle_request() to avoid hanging
```
kuba: how does it help? Moving from one-off (handle_request) to a loop (serve_forever)?

devnsec-com: Because the `serve_forever2()` has a timeout for `handle_request()`, and this will make sure the function `handle_request()` always returns to its caller. The caller `serve_forever2()` will then check the `self._stopped` flag, and decide whether to call `handle_request()` again.

For the test case, it has to set the timeout to make sure it does not hang if it uses `handle_request()`. In the case `handle_request()` returned, it will need to check whether the test case has done or `handle_request()` timed-out, in the case it timed-out, the test case will need to call the `handle_request()` again. This behaviour exactly matches what `serve_forever2()` function does.

devnsec-com:
PR #1200 is definitely a better approach of this issue, the repeated unit tests have been running on my box for 10 minutes by now, not showing any sign of hanging.

kuba:
Okay, allow me then to close this PR to clean up GitHub space a bit. Thanks a lot for testing! Can you comment about your findings on #1200 directly?

Yell at me if you think this should be reopened.

PR #1183, opened 2015-10-29
Offline unittests

pde:
Ensure that unit tests can be run even if you don't have Internet access.

### DIFF
--- a/acme/acme/challenges_test.py
+++ b/acme/acme/challenges_test.py
@@ -359,9 +359,12 @@ class DVSNIResponseTest(unittest.TestCase):
             cert=mock.sentinel.cert))
         mock_verify_cert.assert_called_once_with(self.msg, mock.sentinel.cert)
 
-    def test_simple_verify_false_on_probe_error(self):
+    @mock.patch('acme.challenges.socket.gethostbyname')
+    def test_simple_verify_false_on_probe_error(self, mock_gethostbyname):
         chall = mock.Mock()
         chall.probe_cert.side_effect = errors.Error
+
+        mock_gethostbyname.return_value = '127.0.0.2'
         self.assertFalse(self.msg.simple_verify(
             self.chall, self.domain, self.key.public_key()))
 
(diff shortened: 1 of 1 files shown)


bmw:
I can confirm that unittests currently fail on `master` if you don't have internet access. I can also confirm that simply adding `host='127.0.0.1'` is all that is needed to make the tests work. No new `mock` objects are necessary.

acme/acme/challenges_test.py
```diff
@@ -359,11 +359,14 @@ def test_simple_verify(self, mock_verify_cert):
...
-    def test_simple_verify_false_on_probe_error(self):
+    @mock.patch('acme.challenges.socket.gethostbyname')
+    def test_simple_verify_false_on_probe_error(self, mock_gethostbyname):
         chall = mock.Mock()
         chall.probe_cert.side_effect = errors.Error
+
+        mock_gethostbyname.return_value = '127.0.0.2'
         self.assertFalse(self.msg.simple_verify(
-            self.chall, self.domain, self.key.public_key()))
+            self.chall, self.domain, self.key.public_key(), host="127.0.0.1"))
```
kuba: `probe_cert` is the only function in the entire module that does `gethostbyname` and it should already be mocked in `chall.probe_cert.side_effect = errors.Error` line. If it doesn't work then the whole test suite seems to be broken. So you're not fixing the root cause.

Indeed, we call `self.msg.simple_verify(self.chall, ...)` instead `self.msg.simple_verify(chall, ...)`.

kuba: In summary the only change in this PR should be: `self.chall` to `chall`. `host=...` should also be removed.

kuba:
So, this did decrease coverage. I'm happy to finish up this PR, @pde.

kuba:
FTR: `acme/challenges.py               215      5    98%   442-446` which corresponds to https://github.com/letsencrypt/letsencrypt/blob/890a76ca41d3187601f96599774267c93db5a823/acme/acme/challenges.py#L441-L446.

kuba:
@pde: fixed in #1187.

kuba:
#1187 has been pulled in, so allow me to close this one.

PR #1186, opened 2015-10-29
Use git instead of git-core in bootstrapping on Debian and friends

### LINKED ISSUE #1179: bootstrap script installs transitional package git-core
The correct package for Debian systems is "git".

xtaran:
Fixes #1179 (bootstrap script installs transitional package git-core)


kuba:
LGTM, thanks!

PR #1188, opened 2015-10-29
Add -t to apache2ctl -D DUMP_RUN_CFG

neffs:
Calling apache2ctl -D DUMP_RUN_CFG on Debian Wheezy (Apache 2.2) will
open a socket which breaks standalone auth. letsencrypt still won’t
work with apache 2.2 because it only returns „Syntax OK“.
Apache 2.4 works the same with or without -t.

(Checked with Apache 2.4 on Debian Jessie)


jdkasten:
This looks good to me.  This call is only used to get Apache variables.

I hadn't tested it on Apache 2.2 for some time and had missed this.  Thanks for your help!

PR #1191, opened 2015-10-29
su support (fixes #1148) [needs revision]

### LINKED ISSUE #1148: Using su if sudo not found
For security reasons, not all servers have sudo command.

TheNavigat:
This is my first PR here so I'm not expecting it to be merged before a dozen modifications are made. I'd really appreciate some guiding as I'm still trying to get hold of the project.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -14,7 +14,11 @@ VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
 
 if test "`id -u`" -ne "0" ; then
-  SUDO=sudo
+  if ! type "sudo" > /dev/null; then
+    SUDO="su -c"
+  else
+    SUDO=sudo
+  fi
 else
   SUDO=
 fi
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -103,5 +103,10 @@ fi
 
 # Explain what's about to happen, for the benefit of those getting sudo
 # password prompts...
-echo "Running with virtualenv:" $SUDO $VENV_BIN/letsencrypt "$@"
-$SUDO $VENV_BIN/letsencrypt "$@"
+if [ "$SUDO" = "su -c" ] ; then
+  echo "Running with virtualenv:" $SUDO "$VENV_BIN/letsencrypt \"$@\""
+  $SUDO "$VENV_BIN/letsencrypt \"$@\""
+else
+  echo "Running with virtualenv:" $SUDO $VENV_BIN/letsencrypt "$@"
+  $SUDO $VENV_BIN/letsencrypt "$@"
+fi


letsencrypt-auto
```diff
@@ -14,7 +14,12 @@ VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
 
 if test "`id -u`" -ne "0" ; then
-  SUDO=sudo
+  if ! type "sudo" > /dev/null; then
+    su
```
kuba: This will open new shell session, none of the commands following it would be executed in that session - in act none of the following commands would be executed :(

TheNavigat: Mmmm... Didn't expect that one to be that much trouble ._.

According to that link, http://stackoverflow.com/a/17758312/1256113, if I'd like this to work then I'll have to save the whole following block as a string and then execute it. I tried using ```"su [user] -c "[command here]"``` and it asked for the password every single time, which isn't really favourable, I assume.

What are your thoughts?

TheNavigat: On second thought, environments that don't support sudo (I guess Android is a good example) wouldn't ask for passwords every single time a command is added, as applications such as SuperSU manage that automatically, and they provide the application with root access whenever requested later, although I'm not really sure who'd possibly use Let's Encrypt on an Android device...

However, prepending the commands with ```su  -c "``` and appending them with an extra ```"``` doesn't seem like a really bad idea. Is there such a system where ```sudo``` is not available, apart from Android, so that one can test something like that?

pde:
You should be able to uninstall sudo on any linux system in order to test code with `su -c`.

TheNavigat:
Did that with Ubuntu, apparently it asks for the password every single time :/

I'll try to think of something other than the huge string thing. If someone can throw in a good idea/hint, I would really appreciate the help.

TheNavigat:
I had a semi-recursive idea a few minutes ago. I wrote a small script as a proof-of-concept, and it worked.

```
if test "`id -u`" -ne "0" ; then
  echo "not root"
  su -c ./$0
  exit $?
else
  echo "successful"
fi

echo "end"
```

The sample script worked. I updated the commit, hoping that it'll work as well.

TheNavigat:
Tested ```letsencrypt-auto``` a few minutes ago on Ubuntu 14.04. Worked perfectly.

letsencrypt-auto
```diff
@@ -14,7 +14,12 @@ VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
 
 if test "`id -u`" -ne "0" ; then
-  SUDO=sudo
+  if ! type "sudo" > /dev/null; then
+    su -c ./$0
```
kuba: That seems wrong to me. As I understand it, this tries to restart entire script under `su`.

The problem is, that we don't want all of the commands (e.g. virtualenv) to be run as root. Only OS dependencies bootstrapping and call to `letsencrypt` binary should be privileged.

pde:
Yeah, I think it's better to tolerate the multiple prompts for a password.  I believe there will be two of those the first time the script is run, and one thereafter?

TheNavigat:
I think so.

Prepend with ```su -c " and append with an extra "``` it is, then. I'll post another comment here when I'm done.

pde:
Adding the quotes may require some shell wizardry...

TheNavigat:
The first thing I thought of was adding another variable, analogous to what we do with sudo. I didn't start yet so I'm not really sure if it'll work, I guess I can use some advice.

TheNavigat:
Tested on Ubuntu 14.04.1 with no sudo. Works perfectly!

devnsec-com:
I've created another PR #1282 which properly escape command arguments.

TheNavigat:
The only line where we need sanitization for such characters is line 107. We don't really need to escape `*`, `\`, `(` or `)` since those already tend to corrupt console commands and whoever uses them should sanitize them in advance, we shouldn't be doing this job for them.

Locally, it looks like we only need to add quotes to line 107. Let's see what other kind of trouble is that going to bring upon us.

TheNavigat:
FWIW, no, if the arguments include the same character we're using to contain the string (in my case it's going to be `"`, writing the code write now), it won't cause a problem. The internal strings are processed properly and the resulting command works perfectly.

TheNavigat:
As usual, I was wrong.

```
Running with virtualenv: su -c " /home/local/.local/share/letsencrypt/bin/letsencrypt -v "
su: invalid option -- 'v'
```

TheNavigat:
I've pushed the commit so that people can see what I'm trying to do, because peculiarly the command actually works on the terminal but not through the bash script. I'm thinking maybe the echo command itself reads the ```-v``` incorrectly, but I'll have to check that a bit later.

TheNavigat:
Gotcha!

For some reason when ```"``` is added in a variable, it isn't actually processed as inverted commas that include something, it's processed as a normal character. I'm not sure why that happens, but basically the solution should be something like that.

```
if $SUDO == "su -c":
    $SUDO "$VENV_BIN/letsencrypt "$@""
else:
    $SUDO $VENV_BIN/letsencrypt "$@"
```

Because ```sudo "ls -v"``` doesn't work, apparently.

Will code that in a while, unless someone pointed out early enough that this still won't work.

devnsec-com:
@TheNavigat 
> Will code that in a while, unless someone pointed out early enough that this still won't work.

Sadly to say, but it still doesn't work properly.

For an example, if `$@` is `("a", "b")`, `$SUDO "$VENV_BIN/letsencrypt "$@""` will be `su -c "/path/to/letsencrtpt "a b""`, which equality to `su -c "/path/to/letsencrtpt a" b`, the `b` become user name now.

pde:
#1282 has been merged.  Thanks for your work on this people!

PR #1193, opened 2015-10-29
`echo -n` works in bash, but not in sh

meehow:


letsencrypt-auto
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/bash -e
```
kuba: Note that `bash` is not installed on all systems by default (for instance FreeBSD AFAIR). I'd much prefer if we didn't use `-n` at all. Also, on my system, where `/bin/sh` is `dash`, `echo -n` works perfectly OK.

devnsec-com:
I think the best approach of this is to use `printf` instead of `echo -n`.

kuba:
Interesting idea, @devnsec-com. Is `printf` generally cross-OS compatible? If so, we should also use this approach in https://github.com/letsencrypt/letsencrypt/blob/v0.0.0.dev20151024/letsencrypt/plugins/manual.py instead of using `bash` on Mac OS X.

devnsec-com:
@kuba See: http://www.in-ulm.de/~mascheck/various/echo+printf/ for the availability of `printf` on many platforms.

kuba:
Cool, thanks for link. You're approach SGTM.

bmw:
Since we want to support more shells than `bash`, I don't think we want this. If you implement another solution to the problem (e.g. `printf`), submit another PR and we'd be happy to take a look.

kuba:
Follow up in https://github.com/letsencrypt/letsencrypt/issues/1236.

pde:
Yeah, I actually tried @meehow 's approach while writing `letsencrypt-auto`; it failed on either FreeBSD or OS X because bash was in /usr/local/bin/ or somesuch...

meehow:
@devnsec-com solved it better in d5ccbdbcd29d9b79ba11d42f15c9661d7d60ab24
It's hard to assume that somebody has given shell installed in given location. Doesn't matter if it's bash or dash.

PR #1194, opened 2015-10-29
Update README.rst

freakomaniac:


kuba:
I really don't think it's necessary. It also likely breaks docs generation due to wrong whitespacing (it touches header below).

bmw:
I don't think this is necessary either. Like most public projects on Github, we happily accept PRs from the community. Telling people to contribute isn't needed in my opinion.

PR #1197, opened 2015-10-29
Update manual description

bmw:
@Kuba pointed out to me that the description I suggested for the manual plugin wasn't great. Here's another shot at it.

cc @pde


kuba:
I would suggest "Manually configure webserver". "your" kind of doesn't work in the most typical scenario (`python -m SimpleHTTPServer`), nor do you "edit" and configuration (... files).

kuba:
Also **web**server is a strong indicator of what's happening.

kuba:
LGTM, let's merge

PR #1200, opened 2015-10-29
Remove serve_forever2/shutdown2 (reduces probability of #1085).

kuba:
I'm not even sure why `serve_forever2` and `shutdown2` were introduced in the first place... It probably follows from my misconception about the SocketServer module. After having studied the module again, I come to the conclusion that we can get rid of my crap, simultaneously reducing probability of #1085 (hopefully down to 0)!

`server_forever` is used throughout tests instead of `handle_request`, because `shutdown`, following docs, "must be called while serve_forever() is running in another thread, or it will deadlock", and our `probe_sni` HTTP request is already enough to kill single `handle_request`.

We don't need to use any busy waiting block or `sleep` between serve and shutdown; studying CPython source code leads to the conclusion that the following construction is non-deadlocking:

```python
import threading, SocketServer
s = SocketServer.TCPServer(("", 0), None)
t = threading.Thread(target=s.shutdown)
t.start()
s.serve_forever()  # returns immediately
t.join()  # returns immediately
```

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -4,7 +4,6 @@ import collections
 import functools
 import logging
 import os
-import socket
 import sys
 
 import six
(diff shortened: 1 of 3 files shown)


devnsec-com:
I have tested this PR on my development environment, I repeated ran acme/standalone_test.py for at least 16 minutes, it did not show any sign of hanging.

bmw:
So after looking at this and playing with for a bit, I've reached the same conclusion as @kuba. This reduces the probability of the hang, potentially to 0, but I can't say anything more definitive than that.

The changes that were made here make sense and perform empirically better than `master`. I ran `acme.acme.standalone_test` and `letsencrypt.plugins.standalone_test` in a loop while I was working today. They ran for hours and never hung once. While I'm not certain this solves the problem, I am confident in saying that we want this. Merging.

PR #1210, opened 2015-10-30
bootstrap: use a proper dependency test for Arch

grawity:
`pacman -T` exists for this exact purpose; it respects provides without having to manually code them into the script.

### DIFF
--- a/bootstrap/_arch_common.sh
+++ b/bootstrap/_arch_common.sh
@@ -1,29 +1,27 @@
 #!/bin/sh
 
 # Tested with:
-#   - Manjaro 15.09 (x86_64)
 #   - ArchLinux (x86_64)
-
-# Both "gcc-multilib" and "gcc" packages provide gcc. If user already has
-# "gcc-multilib" installed, let's stick to their choice
-if pacman -Qc gcc-multilib &>/dev/null
-then
-	GCC_PACKAGE="gcc-multilib";
-else
-	GCC_PACKAGE="gcc";
-fi
-
+#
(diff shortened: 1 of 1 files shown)


pde:
1. If I understand it correctly, the previous code would install the latest version of each dependency; the new version will only install things that are completely missing. Though each approach can cause problems, I think it's probably safer to upgrade dependencies to their latest version.

2. @pritambaral are you okay with this change to the way gcc is handled? Superficially it looks like it could break your previous change, unless `gcc` is some kind of virtual package provided by `gcc-multilib`

pde:
@MartijnBraam any thoughts?

grawity:
`gcc` is a real package, but the 'provides' metadata of `gcc-multilib` lets it satisfy the `gcc` dependency.

Regarding updates – overall, in Arch, selective upgrades are not supported at all (since most packages don't have versioned dependencies), so IMHO both approaches can be equally risky (to overall system health).

IMHO, using `-T` is more flexible, since it allows other kinds of "alternative" packages, e.g. libressl providing openssl, or whiptail (libnewt) providing dialog.

pritambaral:
@pde Yes, this seems like a better option. `gcc-multilib` provides `gcc`

@grawity Thanks for enlightening me about pacman's `-T` option

bmw:
LGTM. Thanks for the PR @grawity.

MartijnBraam:
@grawity The `-T` option is great! never knew about it. now I have to rewrite some scripts.

PR #1221, opened 2015-10-30
Gentoo bootstrapping

cheba:
Includes support for all three major package managers.

### DIFF
--- a/bootstrap/_gentoo_common.sh
+++ b/bootstrap/_gentoo_common.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+PACKAGES="dev-vcs/git
+  dev-lang/python:2.7
+  dev-python/virtualenv
+  dev-util/dialog
+  app-admin/augeas
+  dev-libs/openssl
+  dev-libs/libffi
+  app-misc/ca-certificates
+  virtual/pkgconfig"
+
+case "$PACKAGE_MANAGER" in
+  (paludis)
+    cave resolve --keep-targets if-possible $PACKAGES -x
+    ;;
+  (pkgcore)
+    pmerge --noreplace $PACKAGES
+    ;;
+  (portage|*)
+    emerge --noreplace $PACKAGES
+    ;;
+esac
--- a/bootstrap/gentoo.sh
+++ b/bootstrap/gentoo.sh
@@ -1,0 +1,1 @@
+_gentoo_common.sh
--- a/bootstrap/install-deps.sh
+++ b/bootstrap/install-deps.sh
@@ -23,6 +23,9 @@ elif [ -f /etc/arch-release ] ; then
 elif [ -f /etc/redhat-release ] ; then
   echo "Bootstrapping dependencies for RedHat-based OSes..."
   $SUDO $BOOTSTRAP/_rpm_common.sh
+elif [ -f /etc/gentoo-release ] ; then
+  echo "Bootstrapping dependencies for Gentoo-based OSes..."
+  $SUDO $BOOTSTRAP/_gentoo_common.sh
 elif uname | grep -iq FreeBSD ; then
   echo "Bootstrapping dependencies for FreeBSD..."
   $SUDO $BOOTSTRAP/freebsd.sh
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -47,6 +47,9 @@ then
   elif [ -f /etc/redhat-release ] ; then
     echo "Bootstrapping dependencies for RedHat-based OSes..."
     $SUDO $BOOTSTRAP/_rpm_common.sh
+  elif [ -f /etc/gentoo-release ] ; then
+    echo "Bootstrapping dependencies for Gentoo-based OSes..."
+    $SUDO $BOOTSTRAP/_gentoo_common.sh
   elif uname | grep -iq FreeBSD ; then
     echo "Bootstrapping dependencies for FreeBSD..."
     $SUDO $BOOTSTRAP/freebsd.sh


bootstrap/_gentoo_common.sh
```diff
@@ -0,0 +1,12 @@
+PACKAGES="dev-vcs/git dev-lang/python:2.7 dev-python/virtualenv dev-util/dialog app-admin/augeas dev-libs/openssl dev-libs/libffi app-misc/ca-certificates virtual/pkgconfig"
```
kuba: pls split packages: one per line

kuba: also, missing shebang: `#!/bin/sh`

bootstrap/_gentoo_common.sh
```diff
@@ -0,0 +1,12 @@
...
+case "$PACKAGE_MANAGER" in
```
kuba: Name too generic: please use `LETSENCRYPT_GENTOO_PACKAGE_MANAGER`.

kuba:
That's absolutely cool. Welcome to the project!

Before we merge support for `letsencrypt-auto` I would like to make absolute sure, that this is working as intended, because users will expect `letsencrypt-auto` to work and we will effectively need to provide support for it.

1. Have you tested that `./letsencrypt-auto` works as expected? Can you please pastebin log from `script -c "letsencrypt-auto --agree-dev-preview plugins" log`?

2. Have you successfully managed to run standalone plugin? Some logs for confirmation, please?

3. Have you tried the Apache plugin? Again, logs?

4. Other plugins? `webroot`, `manual`?

If you don't feel like `letsencrypt-auto` is ready for end-users, you can remove that part and have the dev setup merged at least. Thanks in advance for you help :)

Also, would you be willing to maintain Gentoo bootstrap scripts and providing support for users?

bootstrap/_gentoo_common.sh
cheba: This is a system var.

Gentoo has a tool called `eselect` it let's user set preferred package manager. This var is used by other tools that need to interact with package manager. This is exactly that case.

kuba: OK, then SGTM

kuba:
Rebase just caused the comments to be hidden :(

cheba:
I tried is with the following command:

```shell
./letsencrypt-auto --agree-dev-preview --server https://***.api.letsencrypt.org/directory auth
```

And it displayed:
```
Bootstrapping dependencies for Gentoo-based OSes...
```

And proceeded installing missing packages. Asking questions and finally installing a certificate.

I'm not sure about any plugins. I'm totally ignorant of the overall architecture. The bootstrapping though looked pretty straightforward so I took a chance to add Gentoo support.

There doesn't seem to be anything relevant in the logs. It seem to only log what's going on in the Python part of it.

bootstrap/_gentoo_common.sh
cheba: Addressed both comments. Please see updated code.

pde:
Works for me on a `Pygoscelis-Papua_Gentoo_64-bit-EBS-2015-03-03-01-55-16` AMI.  Merging.

infectormp:
don't work for me.

i run ```./letsencrypt-auto --agree-dev-preview --debug --server https://***.api.letsencrypt.org/directory certonly```

```Bootstrapping dependencies for Gentoo...```
```./letsencrypt-auto: line 36: : command not found```

**Comment by cheba** [comment]
@infectormp What package manager do you usually use? Also please show output of the following commands:
```shell
eselect package-manager show
echo "${PACKAGE_MANAGER}"
```

infectormp:
@cheba i use Portage-2.2.24.

```eselect package-manager show```
```!!! Error: Can't load module package-manager
exiting```

```echo "${PACKAGE_MANAGER}"```
give empty answer 

need ````emerge app-eselect/eselect-package-manager``` ??

**Comment by cheba** [comment]
@infectormp No, you shouldn't need that. The bootstrap script should default to portage just fine. I just wondered if it might've been set to some other package manager that you don't use.

Could you please try something?

```bash
cd letsencrypt/bootstrap
bash -c "set -x; source ./install-deps.sh"
```

And post the output here.

infectormp:
```
bash -c "set -x; source ./install-deps.sh"
+ source ./install-deps.sh
+++ id -u
++ test 0 -ne 0
++ SUDO=
+++ dirname bash
++ BOOTSTRAP=.
++ '[' '!' -f ./debian.sh ']'
++ '[' -f /etc/debian_version ']'
++ '[' -f /etc/arch-release ']'
++ '[' -f /etc/redhat-release ']'
++ '[' -f /etc/gentoo-release ']'
++ echo 'Bootstrapping dependencies for Gentoo-based OSes...'
Bootstrapping dependencies for Gentoo-based OSes...
++ ./_gentoo_common.sh
```
```
 * IMPORTANT: 6 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.

Calculating dependencies... done!
>>> Jobs: 0 of 0 complete                           Load avg: 0.09, 0.08, 0.05
>>> Auto-cleaning packages...

>>> No outdated packages were found on your system.
```

cheba:
All right. So this doesn't look like a bootstrapping issue. This is something else happening elsewhere.

A bit of digging in the code suggests that you may have no `sudo` installed. Could you please check?

```bash
which sudo
```

infectormp:
```
which sudo
/usr/bin/sudo
```

infectormp:
i run letsencrypt-auto from ```root```

cheba:
Hm… Could you please try running `letsencrypt-auto` with `--debug` flag?

infectormp:
@cheba if you look to my first post you can see ```--debug``` flag.

infectormp:
without ```-debug``` flag i have message:
```
WARNING: Gentoo support is very experimental at present...
```

cheba:
Indeed, my bad.

Then could you please check if `bootstrap/_gentoo_common.sh` file is present and has executable permissions on it?

```bash
ls -la bootstrap/_gentoo_common.sh
```

infectormp:
```
ls -la bootstrap/_gentoo_common.sh
-rwxr-xr-x 1 root root 416 Nov  6 16:42 bootstrap/_gentoo_common.sh*
```

Haasje:
I've had the same problem this weekend when i tried to run letsencrypt-auto:

>./letsencrypt-auto: line 36: : command not found

I've fixed the problem by changing the line to:

>      echo "Bootstrapping dependencies for $1..."
>      if [ "$3" != "" ] ; then
>        "$3" "$BOOTSTRAP/$2"
>      else
>        "$BOOTSTRAP/$2"
>      fi

The problem seems to be in an empty $3. For some reason this does not work.
Also, I changed the _gentoo_common.sh to include the -v flag for emerge as the verbose boolean is not yet propagated into _gentoo_common.sh.

Haasje

cheba:
I'm running out of ideas…

Looks like your `sh` doesn't like empty commands. But then, Why did the script work for me? I ran it too under `root`.

That line, probably, needs to look like this:
```bash
sh -c "$3 '${BOOTSTRAP}/$2'"
```

But that's unrelated issue to this PR. Probably needs it's own issue.

infectormp:
@cheba your solution work fine

PR #1226, opened 2015-10-30
Apache with more than one vhost (fixes #809)

### LINKED ISSUE #809: Apache install fails with two vhosts
When I run `letsencrypt install --cert-path /etc/letsencrypt/live/example.com/cert.pem --key-path /etc/letsencrypt/live/example.com/privkey.pem --chain-path /etc/letsencrypt/live/example.com/chain.pem` with the following configuration https://drive.google.com/file/d/0B9CrBQYK660GRm1EVVFoQzdIRFE/view?usp=sharing, installation fails. The reason for this is the dummy directives including things like `insert_cert_file_path` and `insert_key_file_path` are not removed before restarting Apache, causing it to fail.

SwartzCr:
Added interstitial saves to apache changes which fixed but when multiple vhosts were specified

The final save still gets called an sets a checkpoint

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -219,6 +219,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                 self.parser.add_dir(
                     vhost.path, "SSLCertificateChainFile", chain_path)
             else:
+                # TODO: THIS???
                 self.aug.set(path["chain_path"][-1], chain_path)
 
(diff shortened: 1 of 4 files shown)


SwartzCr:
fixes problem in issue #809

pde:
Nice. @jdkasten can you sign off on this?

bmw:
I can confirm this fixes the issue I reported in #809. LGTM.

jdkasten:
Thanks @SwartzCr and @bmw!  This is the solution we thought would work yesterday, thanks for implementing it and testing it out.

PR #1229, opened 2015-10-30
Fix install

bmw:


letsencrypt/cli.py
```diff
@@ -494,7 +495,8 @@ def install(args, config, plugins):
     # XXX: Update for renewer/RenewableCert
 
     try:
-        installer, _ = choose_configurator_plugins(args, config, plugins, "certonly")
```
pde: This looks like this bug has been here for a while.  I wonder if we should in fact be calling `pick_installer` here?

pde:
LGTM

PR #1233, opened 2015-10-30
Ensure that mandatory flags are displayed under the relevant verb help topics

### LINKED ISSUE #996: Common subcommand flags help
Currently, when you run commands like `letsencrypt --help install`, you get the following output:
```
$ letsencrypt --help install  
usage: 

the cert.  

positional arguments:
  SUBCOMMAND
    run                 Obtain a certificate and install.
    auth                Authenticate & obtain cert, but do not install it.
    install             Install a previously obtained cert in a server.
    revoke              Revoke a previously obtained certificate.
    rollback            Rollback server configuration changes made during
                        install.
    config_changes      Show changes made to server config during installation
    plugins             List server software plugins.

optional arguments:
                        config file path (default: None)

install:
  Options for modifying how a cert is deployed
```
This to me doesn't seem very helpful. The problem is arguments shared between subcommands. Most of these common flags are now in the `paths` argument subgroup. I think ideally, the common arguments should not be duplicated when running `letsencrypt --help all`, but should be shown when running help for a specific subcommand. I do not believe this behavior is achievable with `argparse` and adds a fair amount of complexity, so we shouldn't worry about doing anything like this for now.

Instead, a simple thing we could do is update the description for the relevant subgroups to point users towards the `paths` group. I'm not sure what other simple fixes could be done here, but I don't think the help for commands like `install` and `revoke` does much to help the user right now.

pde:
Closes #996

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -38,7 +38,7 @@ load-plugins=linter_plugin
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=fixme,locally-disabled,abstract-class-not-used,bad-continuation,too-few-public-methods,no-self-use,invalid-name
+disable=fixme,locally-disabled,abstract-class-not-used,bad-continuation,too-few-public-methods,no-self-use,invalid-name,too-many-instance-attributes
 # abstract-class-not-used cannot be disabled locally (at least in pylint 1.4.1)
 
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -650,12 +650,12 @@ class HelpfulArgumentParser(object):
         help1 = self.prescan_for_flag("-h", self.help_topics)
         help2 = self.prescan_for_flag("--help", self.help_topics)
         assert max(True, "a") == "a", "Gravity changed direction"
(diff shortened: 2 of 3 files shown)


.pylintrc
```diff
@@ -38,7 +38,7 @@ load-plugins=linter_plugin
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=fixme,locally-disabled,abstract-class-not-used,bad-continuation,too-few-public-methods,no-self-use,invalid-name
+disable=fixme,locally-disabled,abstract-class-not-used,bad-continuation,too-few-public-methods,no-self-use,invalid-name,too-many-instance-attributes
```
kuba: `too-many-instance-attributes` is a very good indicator that something is wrong with the code - please fix the code, or disable locally if you **must**

kuba:
Tests missing :(

.pylintrc
pde: We've had to disable that pylint warning 14 times already in our code.  We have plenty of intelligent humans to pay attention to the elegance or otherwise of our code; we don't need foolish robots as well :)

bmw:
LGTM

Issue #1236, opened 2015-10-30
Use printf instead of echo -n


kuba:
See discussion in https://github.com/letsencrypt/letsencrypt/pull/1193. We want this both in bootstrap scripts and https://github.com/letsencrypt/letsencrypt/blob/v0.0.0.dev20151024/letsencrypt/plugins/manual.py.

devnsec-com:
I've created PR #1238 for this.

kuba:
Fixed as of #1238.

BTW, @devnsec-com, you can include "fixes #NUM" or "closes #NUM" in your commits in order for bugs to be automatically closed

PR #1237, opened 2015-10-30
Use printf in place of echo -n

pde:
Closes: #1236

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -70,7 +70,7 @@ then
   fi
 fi
 
-echo -n "Updating letsencrypt and virtual environment dependencies..."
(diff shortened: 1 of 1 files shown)


kuba:
It doesn't close #1236, because:
- it doesn't fix `boootstrap/venv.sh` stuff
- it doesn't fix Manual plugin

kuba:
Otherwise it's good to merge, but please remember to reopen ticket if you don't fix both problems above!

devnsec-com:
Oops, I've created PR #1238 seconds after you created this PR, but I have also fixed `boootstrap/venv.sh` and `letsencrypt/plugins/manual.py` as @kuba said.

pde:
OK, I'll close this one and #1238 can land when it's ready.

PR #1238, opened 2015-10-30
Use `printf` instead of `echo -n` for better portability

devnsec-com:
Related issue: #1236

### DIFF
--- a/bootstrap/venv.sh
+++ b/bootstrap/venv.sh
@@ -25,9 +25,9 @@ pip install -U letsencrypt letsencrypt-apache # letsencrypt-nginx
 echo
 echo "Congratulations, Let's Encrypt has been successfully installed/updated!"
 echo
-echo -n "Your prompt should now be prepended with ($VENV_NAME). Next "
-echo -n "time, if the prompt is different, 'source' this script again "
-echo -n "before running 'letsencrypt'."
+printf "%s" "Your prompt should now be prepended with ($VENV_NAME). Next "
+printf "time, if the prompt is different, 'source' this script again "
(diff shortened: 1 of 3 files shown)


bootstrap/venv.sh
```diff
@@ -25,9 +25,9 @@ pip install -U letsencrypt letsencrypt-apache # letsencrypt-nginx
 echo
 echo "Congratulations, Let's Encrypt has been successfully installed/updated!"
 echo
-echo -n "Your prompt should now be prepended with ($VENV_NAME). Next "
-echo -n "time, if the prompt is different, 'source' this script again "
-echo -n "before running 'letsencrypt'."
+printf "%s" "Your prompt should now be prepended with ($VENV_NAME). Next "
```
kuba: why do we need "%s" in this case but not others?

kuba:
It seems like `printf` drop quotes: https://travis-ci.org/letsencrypt/letsencrypt/jobs/88451445#L2880

bootstrap/venv.sh
devnsec-com: Becaus the $VENV_NAME could contains symbols which will cause issues, but others are plain texts.

devnsec-com:
> It seems like printf drop quotes: https://travis-ci.org/letsencrypt/letsencrypt/jobs/88451445#L2880

I'm looking into this issue.

devnsec-com:
Seems like I'm quoting a quoted string which lead to drop quotes. I'm going to push a fixed version soon.

devnsec-com:
@kuba OK, that problem has solved.

kuba:
LGTM

PR #1232, opened 2015-10-30
Bootstrap CentOS 6 Compatibility Improvement

devnsec-com:
Fix #1204 missing `virtualenv` issue, now the bootstrap will try to install `python-virtualenv` via `yum` or `dnf` first, and then try to use `pip install virtualenv` if previous attempt failed.

Also removed duplicates `python-devel`. Have two `python-devel` in the `yum install` or `dnf install` packages list make no sense.

PR #1241, opened 2015-10-30
Add documentation tell users to enable EPEL on RedHat/CentOS 6

### LINKED ISSUE #1204: letsencrypt-auto Error "virtualenv: command not found" On Fresh Installed CentOS 6
Logs:
```
# ./letsencrypt-auto -v
Bootstrapping dependencies for RedHat-based OSes...
yum is /usr/bin/yum
Loaded plugins: fastestmirror
Setting up Install Process
Loading mirror speeds from cached hostfile
 * base: centos.muzzy.org.uk
 * extras: mirrors.vooservers.com
 * updates: mirrors.ukfast.co.uk
Package git-1.7.1-3.el6_4.1.x86_64 already installed and latest version
Package python-2.6.6-64.el6.x86_64 already installed and latest version
Package python-devel-2.6.6-64.el6.x86_64 already installed and latest version
No package python-virtualenv available.
Package python-devel-2.6.6-64.el6.x86_64 already installed and latest version
Package gcc-4.4.7-16.el6.x86_64 already installed and latest version
Package dialog-1.1-9.20080819.1.el6.x86_64 already installed and latest version
Package augeas-libs-1.0.0-10.el6.x86_64 already installed and latest version
Package openssl-devel-1.0.1e-42.el6.x86_64 already installed and latest version
Package libffi-devel-3.0.5-3.2.el6.x86_64 already installed and latest version
Package ca-certificates-2015.2.4-65.0.1.el6_6.noarch already installed and latest version
Nothing to do
./letsencrypt-auto: line 67: virtualenv: command not found
```

devnsec-com:
This PR replaces PR #1232 to fix #1204

### DIFF
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -21,7 +21,6 @@ $tool install -y \
   python \
   python-devel \
   python-virtualenv \
-  python-devel \
   gcc \
   dialog \
   augeas-libs \
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -42,6 +42,11 @@ To install and run the client you just need to type:
 
    ./letsencrypt-auto
 
+.. note:: On RedHat/CentOS 6 you will need to enable the EPEL_
+   repository before install.
+
+.. _EPEL: http://fedoraproject.org/wiki/EPEL
+
 Throughout the documentation, whenever you see references to
 ``letsencrypt`` script/binary, you can substitute in
 ``letsencrypt-auto``. For example, to get the help you would type:


pde:
Can we just do this for the user, instead of telling them to do it?  Or failing that, detect its absence and give them a command to run?

devnsec-com:
Because we definitely do not want to enable EPEL for an user who has Python 2.7 installed (not necessarily via `rpm`, maybe they compiled it themselves), because the `python-virtualenv` provided in EPEL is for Python 2.6

pde:
OK. I'd still prefer to automate detection of those circumstances if possible (after all, the sys admin may  naively follow our instructions even if Python 2.7 is installed).  But will merge this if you don't think an automated approach is feasible.

kuba:
Why do you need to install this EPEL repo? A couple of people managed to get away without EPEL when using #1257.

bootstrap/_rpm_common.sh
```diff
@@ -21,7 +21,6 @@ $tool install -y \
   python \
   python-devel \
   python-virtualenv \
-  python-devel \
```
kuba: @devnsec-com: since we might not merge this PR at all, can you submit another with this change only?

pde:
@devnsec-com did #1257 being merged resolve the need for this?  Or should we still figure out how to land it?

devnsec-com:
@pde I don't think #1257 solves this issue(but I didn't test it), because the CentOS 6 does not have the `virtualenv` in the offcial repos, without EPEL the bootstrap script will not be able to install `virtualenv`.

centminmod:
yes centos 6 doesn't have yum virtualenv package hence requires EPEL https://community.letsencrypt.org/t/redhat-centos-6-x-users-need-python-2-7/2190/7

pde:
OK, merging in that case.

PR #1242, opened 2015-10-30
Added comment about installer.save()

bmw:
In response to #1226, I just wanted to add a comment about why the call to `installer.save()` is necessary so it's not arbitrarily deleted in the future.


pde:
Merge the comment.  Perhaps also put James's full explanation in the code?

jdkasten:
Yep, I think the comment is good.  Calling save more than once without changes should not cause any problems.  (It is essentially a NOP in Apache)  Thanks!

PR #1244, opened 2015-10-31
Search key_dir when performing Proof of Possession

christophermanning:
Fixes https://github.com/letsencrypt/letsencrypt/issues/964

This is my first technical PR for this repo so any code review is especially welcome and appreciated.

### DIFF
--- a/letsencrypt/continuity_auth.py
+++ b/letsencrypt/continuity_auth.py
@@ -31,7 +31,8 @@ class ContinuityAuthenticator(object):
         :type installer: :class:`letsencrypt.interfaces.IInstaller`
 
         """
-        self.proof_of_pos = proof_of_possession.ProofOfPossession(installer)
+        self.proof_of_pos = proof_of_possession.ProofOfPossession(
+            installer, config)
 
     def get_chall_pref(self, unused_domain):  # pylint: disable=no-self-use
         """Return list of challenge preferences."""
--- a/letsencrypt/continuity_auth.py
+++ b/letsencrypt/continuity_auth.py
@@ -32,7 +32,7 @@ class ContinuityAuthenticator(object):
 
         """
         self.proof_of_pos = proof_of_possession.ProofOfPossession(
-            installer, config)
+            config, installer)
 
     def get_chall_pref(self, unused_domain):  # pylint: disable=no-self-use
         """Return list of challenge preferences."""
--- a/letsencrypt/proof_of_possession.py
+++ b/letsencrypt/proof_of_possession.py
@@ -26,8 +26,9 @@ class ProofOfPossession(object):  # pylint: disable=too-few-public-methods
     :type installer: :class:`~letsencrypt.interfaces.IInstaller`
 
     """
-    def __init__(self, installer):
+    def __init__(self, installer, config):
         self.installer = installer
+        self.config = config
 
     def perform(self, achall):
         """Perform the Proof of Possession Challenge.
@@ -60,6 +61,13 @@ class ProofOfPossession(object):  # pylint: disable=too-few-public-methods
             cert_key = achall.alg.kty(key=cert_obj.public_key())
             if cert_key == achall.hints.jwk:
                 return self._gen_response(achall, key)
+
+        if os.path.isdir(str(self.config.key_dir)):
+            for key_file in os.listdir(self.config.key_dir):
+                full_path = os.path.join(self.config.key_dir, key_file)
+                response = self._gen_response(achall, full_path)
+                if response is not False:
+                    return response
 
         # Is there are different prompt we should give the user?
         code, key = zope.component.getUtility(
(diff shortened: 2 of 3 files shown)


letsencrypt/continuity_auth.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, config, installer):  # pylint: disable=unused-argument
         :type installer: :class:`letsencrypt.interfaces.IInstaller`
 
         """
-        self.proof_of_pos = proof_of_possession.ProofOfPossession(installer)
+        self.proof_of_pos = proof_of_possession.ProofOfPossession(
+            installer, config)
```
bmw: Nit: I'd flip the order so that `config` is the first argument and `installer` is the second. This is more consistent with the rest of the code base.

letsencrypt/proof_of_possession.py
```diff
@@ -61,6 +62,13 @@ def perform(self, achall):
             if cert_key == achall.hints.jwk:
                 return self._gen_response(achall, key)
 
+        if os.path.isdir(str(self.config.key_dir)):
```
bmw: `str` shouldn't be necessary. To fix unit tests in `proof_of_possession_test`, add a value for `key_dir` to the mocked configuration created on line 31. This can be accomplished with something like `self.config = mock.MagicMock(key_dir="/nonexistent/key/dir")`

letsencrypt/proof_of_possession.py
```diff
@@ -61,6 +62,13 @@ def perform(self, achall):
...
+            for key_file in os.listdir(self.config.key_dir):
+                full_path = os.path.join(self.config.key_dir, key_file)
+                response = self._gen_response(achall, full_path)
+                if response is not False:
```
bmw: So this doesn't quite work. The idea behind proof of possession is that the server gives the client a public key and the client has to sign a message using the corresponding private key.

`self._gen_response` simply takes a key and signs the message with no check that the correct private key was used. The result is that the code currently grabs the first private key it finds in `key_dir` and uses it to sign the message. What you need to do is load each private key as a JSON web key, get the corresponding public key, and compare it with the key given by the server. Only when you find a match can you generate a response and return it.

To do this nicely will require a bit of refactoring. Basically what you want to do is factor out the code that loads the key in `_gen_response` into a new function. You can then use this function both in `_gen_response` and in the code above to return a JWK object. You can then call `.public_key()` on that object to return the public key. After that, you need to compare it `achall.hints.jwk`. If there is a match, use the already loaded in key to generate the response.

Hopefully that all makes sense. Please follow up if you have any questions.

letsencrypt/tests/proof_of_possession_test.py
```diff
@@ -78,6 +80,15 @@ def test_perform_with_input(self, mock_input):
         self.assertFalse(self.proof_of_pos.perform(self.achall))
         self.assertTrue(self.proof_of_pos.perform(self.achall).verify())
 
+    def test_perform_with_key_dir(self):
+        # Remove the matching certificate
+        self.installer.get_all_certs_keys.return_value.pop()
+
+        self.config.key_dir = tempfile.mkdtemp('key_dir')
+        crypto_util.init_save_key(2048, self.config.key_dir)
```
bmw: There's no need to generate a new key here. `CERT2_KEY_PATH` and `CERT3_KEY_PATH` are paths to already generated private keys. `CERT3_KEY_PATH` is the one that matches the public key for the challenge used in these tests.

bmw:
Thanks for the PR @christophermanning! I'm appending `[needs revision]` to the title of this PR for the time being to denote that this isn't ready to be merged in yet. Just update the title when it's ready for re-review.

christophermanning:
Thank you for the detailed code review. I've revised my changes based on your guidance.

Please let me know if there is anything else you'd like me to change for this PR.

bmw:
Awesome! Thanks again for the PR @christophermanning.

Since we're rapidly approaching general availability, the client team has actually stopped reviewing PRs for issues not in the "1.0 for launch" [milestone](https://github.com/letsencrypt/letsencrypt/issues?q=is%3Aopen+is%3Aissue+milestone%3A%221.0+for+launch%22). We certainly appreciate the help though and I don't want to diminish the importance of this PR. We will take a look at this as soon as we've closed our launch critical issues. Thanks again for your help.

bmw:
Since #964 was opened, a number of things have changed with proof of possession:

1. Neither the server nor the client has any plans to support proof of possession in the immediate future.
2. There are plans to redesign how the proof of possession challenge works.
3. Support for proof of possession has been completely removed from `acme` and `certbot` for the time being.

With this in mind, I created a new branch on this repo called `use_key_dir_in_pop` containing all of your commits and made a comment about it in #964.

If this proves useful for PoP in the future, we'll encourage others to build off your work. Thanks again for the PR @christophermanning.

Issue #1216, opened 2015-10-30
Warning about port in use should appear earlier


chk1:
`The program nginx (process ID 12345) is already listening on TCP port 80. This will prevent us from binding to that port. Please stop the nginx program temporarily and then try again.`

From a usability standpoint I think it would be better to display this warning earlier in the process when running `letsencrypt-auto`, as I have to re-enter my domain names and go through all the steps. Maybe a kind of list or auto-suggestion feature for domains that I have already used would be nice.

centminmod:
+1 this would be much better and clearer for end users :)

wjdavis5:
+1 agreed this one got me for sure.

pde:
I'd have marked this **2.0** but since several of you noted it as problematic, it's **nice for 1.0**.

shrikrishnaholla:
I am planning on taking this up. Seems easy enough to begin with :+1:

pde:
The right moment for this test is *probably* immediately after the plugins have been selected, if and only if the user is using the `standalone` authenticator:

https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/cli.py#L427

metachris:
This PR is my suggestion of a minimal intrusive fix without code duplication: https://github.com/letsencrypt/letsencrypt/pull/1474.

bmw:
Catching this error earlier in Certbot's execution requires checking that ports are available before we know what challenges we need to perform. For example, with the current state of the ACME protocol, Certbot, and Let's Encrypt, if you use the standalone plugin without specifying `--standalone-supported-challenges` or `--preferred-challenges` on the command line, only the `TLS-SNI-01` challenge is performed and we only check that this port is available.

To move this check any earlier than we do now we either need to:

1. Code in assumptions that we can't safely make about the ACME protocol and Let's Encrypt's behavior about what challenges will be performed.
2. Check that every port that is possibly required is available.

We feel that either of these options is worse than what we're currently doing so I'm marking this issue as wontfix.

PR #1251, opened 2015-10-31
Add check in letsencrypt-auto to ensure port 80 not in use

shrikrishnaholla:
Ref #1216 .

Checks whether port 80 is in use, before doing anything else. If it is, exits with a message describing the issue. Continues with the setup otherwise.

Tested with bash. Not sure about other shell environments. Let me know if the approach needs to be changed.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -19,7 +19,13 @@ else
   SUDO=
 fi
 
-for arg in "$@" ; do 
+if timeout 1 sh -c 'cat < /dev/null > /dev/tcp/127.0.0.1/80' >/dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


kuba:
Unfortunately, this check happens waaay to early: it doesn't even allow you to call `./letsencrypt-auto --help` if port 80 is taken... I wouldn't like to discourage you too much but fix for #1216 is a lot more complicated and might not be even feasible in the current plugins architecture :( I'm gonna close this PR for now, as is it's not close to the solution, sorry! If you'd like to work on that, you need to take a closer look at the Python code - fix to the problem is not within `letsencrypt-auto` scope.

Issue #1016, opened 2015-10-17
Webroot plugin should support SANs (multiple public_html paths)


kuba:

centminmod:
hopefully soon https://community.letsencrypt.org/t/multi-domain-san-certificate-and-webroot-auth/2493

Maybe this suggestion https://github.com/letsencrypt/letsencrypt/pull/757#issuecomment-1425074155

> for multiple names, maybe allow multiple --webroot-path in an array and then if multiple -d nameXX are specified cycle through the --webroot-path array to check against the -d name XX values ?

kuba:
FYI I have created https://github.com/kuba/simp_le that is basically equivalent to webroot, but a lot simpler, with less deps and with renewal support. @centminmod, check it out :)

kuba:
And, of course, with multiple SAN support.

pde:
Currently I think a better CLI design for this may be, `letsencrypt --webroot -w /var/www/example -d example.com -d www.example.com -w /var/www/thingimy -d thingimy.net -d www.thingimy.net`.

bmw:
@centminmod, I'm not sure if I understand your suggestion for this. Can you give me an example CLI stating which webroot would be applied for which domains?

bmw:
Lately, I've been thinking a lot about multiple webroots and today I've been thinking about how to best specify them on the command line. The interface proposed in #1594 is pretty good, but I'm wondering if that's the best approach. Sure, we could always change the interface later, but I'd rather take a few extra days to get the interface right the first time instead of causing a breaking change later.

Unfortunately, I haven't been able to come up with a CLI I really like. The main problem I'm having is mapping webroots to domains. The way I see it, there are a couple approaches:

#### Include mapping in webroot-path flag
This looks something like `--webroot-path /var/www/html:example.com`. My main problem here is duplication. If you're specifying everything on the command line, you must include `--domains` meaning that you have to specify each domain in both `--webroot-path` and in `--domains`. This is excessive.

We can optimize this approach, however. The best I have come up with is using a range in `--webroot-path`, similar to how you specify lines in `git blame`. For example, 
```
letsencrypt --webroot --domains example.com,www.example.com,example.org,www.example.org,example.net,www.example.net --webroot-path /var/www/html:example.com,www.example.org --webroot-path /usr/share/nginx/html:example.net,www.example.net
```

`/var/www/html` will be used for: `example.com,www.example.com,example.org,www.example.org`
[...]

PR #1259, opened 2015-10-31
added: --webroot-domain-path domain:path [needs revision]

grubberr:
one of possible solutions for this #1016

added new option 

```
--webroot-domain-path domain:path
```

for example:

```
./bin/letsencrypt auth ... --authenticator webroot \
  --domains domain1.com --webroot-domain-path domain1.com:/home/user/public_html \
  --domains domain2.com --webroot-domain-path domain2.com:/home/user/public_html/domain2
```

old invocation also works

```
./bin/letsencrypt auth ... --authenticator webroot \
  --domains domain1.com --webroot-path /home/user/public_html
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,6 @@ addons:
   mariadb: "10.0"
   apt:
     packages:  # keep in sync with bootstrap/ubuntu.sh and Boulder
-    - lsb-release
     - python
     - python-dev
     - python-virtualenv
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -58,6 +58,7 @@ setup(
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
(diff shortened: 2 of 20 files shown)


kuba:
I would rather prefer `--webroot-path "default/path/if/nothing/specific/given, domain1.com:specific/path/to/domain1, domain2:specific/path/to/domain2"` syntax:
- config file doesn't allow value to be specified multiple times
- no need to create a brand new flag, `--webroot-path` can be overloaded
- `--webroot-path foo` still has the same meaning
- you can specify default value for domains that share common `public_html`

Good way to approach it is to decode comma/colon-separated string inside `manual.Autenticator.paths`, by defining new `@property` that operates over `self.conf("path")`.

@grubberr, do you think you could work towards this solution?

grubberr:
@kuba ok I will re-implement according to your requirements

kuba:
@grubberr, thank you very much! Welcome to the project! :)

kuba:
Please remove `[needs revision]` from the title when you're done.

grubberr:
closed because of #1284

Issue #1123, opened 2015-10-24
Implement flags to control: security parameters in general, ciphersuites in particular, and autoupdating of those things


pde:
At the moment, we tweak ciphersuites during `install` but not afterwards.  Arguably the renewal script, when it runs on a daily basis, should check for changes to best practices and auto deploy them.

Users should be able to control several aspects of this:
- Do they get ciphersuites optimised for compatibility, for security, a tradeoff betewen those, or in accordance with a specific project's recommendations (Mozilla/Bulletproof TLS/etc)
- Do they get other security features optimised for compatibility, for security, or a tradeoff?
- Are changes to these things automatically applied after a client update, or only at manual sys admin request?

I believe the default for each of these things should be tradeoff, tradeoff, and automatic.  And those defaults should be documented.

pde:
We do need some level of notification to the user that automation is the default.  I think the right place to do that is in a post-execution note, probably the same one that tells them that autorenewal is active.

pde:
See also #624.

rugk:
See also the thread about this in the LE community: [Does the LE client (automatically) update the TLS config/cipher list if a vulnerability is found?](https://community.letsencrypt.org/t/does-the-le-client-automatically-update-the-tls-config-cipher-list-if-a-vulnerability-is-found/1981)

schoen:
According to my PR #1261 we should have 3 options at launch (`mozilla-secure`, `mozilla-intermediate` [default], and `mozilla-old`) with respect to the ciphersuites, and we should check periodically whether the strings in our code are up-to-date with Mozilla's recommendations. People can also lobby us to add other ciphersuite setting strings if they can make a case that there is a reasonably large population that would use them.

jvehent:
@schoen: Would it help if we provided the mozilla recommendations in JSON format published at a stable location? We're talked about it a few times, and letsencrypt would be a strong enough incentive to act on it :)

pde:
@jvehent If the JSON format mapped to Nginx / Apache configuration options, I think that would be quite useful.  We could then land a unit test to ensure that what we have matches what Mozilla currently publishes.

pde:
When closing this ticket, read #1401.

jvehent:
As a follow-up, v4 of the Mozilla guidelines is now available in JSON format: https://statics.tls.security.mozilla.org/server-side-tls-conf-4.0.json

[...]

PR #1261, opened 2015-10-31
Ciphers

### LINKED ISSUE #555: Review all cryptographic algorithm and parameter defaults
There are many places where the Let's Encrypt client will generate cryptographic keys or parameters, or set or use cryptographic parameter or algorithm defaults (including defining key sizes and ciphersuite defaults for servers in which we enable TLS!).

We need a thorough review of whether all of our choices are good, not least because they may become the TLS configuration defaults for millions of Internet servers.

I started a page to collect references to what other people have had to say about ciphersuites at

https://github.com/letsencrypt/letsencrypt/wiki/Ciphersuite-guidance

That mostly doesn't include information about choices for other kinds of parameter and keylength options, but in a few places it does (and maybe we can put more of that kind of material in there).

Cc @jhalderm.

schoen:
Fixes #555 (in terms of providing and documenting a course of action, not implementing it!).

The theory is that we have decided to track Mozilla's ciphersuite recommendations, which Mozilla has extensively considered for web client compatibility issues and continues to update. We are also open to providing options to track other sources of recommendations, recognizing that different sites have different needs. This patch documents the issues involved in general terms and sets out what we have decided to do. As @pde described, we still need to actually implement these things, which is tracked at #1123.

### DIFF
--- a/docs/ciphers.rst
+++ b/docs/ciphers.rst
@@ -192,6 +192,20 @@ or
 
 to disable automatic configuration updates. These features have not yet
 been implemented and this syntax may change then they are implemented.
-The status of this feature is tracked as issue #1123 in our bug tracker.
+
+
+TODO
+----
+
+The status of this feature is tracked as part of issue #1123 in our
+bug tracker.
 
 https://github.com/letsencrypt/letsencrypt/issues/1123
+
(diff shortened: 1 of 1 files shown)


pde:
Great.

PR #1265, opened 2015-10-31
Only works for me with port 80 and 443 forwarded

mbrugger:
For the docker client to work correctly I needed to forward both ports 443 and 80. I will take a closer look on how this is supposed to work.


kuba:
Yes, that's correct. Standalone client randomly picks between two available challenges that run on different ports: 80 and 443. Manual and webroot need 80. Apache needs 443.

LGTM

Issue #1137, opened 2015-10-26
Narrative documentation for plugins


bmw:
While changes proposed in #1079 help, it is very hard to tell from the names `manual`, `standalone`, and `webroot` what our plugins actually do. We need to add narrative documentation for each plugin to help people understand the purpose of each plugin and when they should be used.

bmw:
@kuba, @pde: Do you feel this has been fixed enough by #1121 to either close this issue or kick it out of the "1.0 for launch" milestone?

kuba:
I think we should still work on improving docs (including narrative plugins explanation) before launch.

kuba:
https://github.com/letsencrypt/letsencrypt/issues/1190 is one of the examples why clear narrative docs is necessary.

kuba:
Since I personally find it very high priority, I will work on this during weekend.

pde:
From the latest Beta Program email:

# How to use the client

The Let's Encrypt client supports a number of different "plugins" that can be used to obtain and/or install certificates.  A few examples of the options are included below:

If you're running Apache on a recent Debian-based OS, you can try the Apache plugin, which automates both obtaining and installing certs:

  ./letsencrypt-auto --apache --server https://acme-v01.api.letsencrypt.org/directory --agree-dev-preview 

To obtain a cert using a "standalone" webserver (you may need to temporarily stop your exising webserver) for example.com and www.example.com:

  ./letsencrypt-auto certonly -a standalone -d example.com -d www.example.com --server https://acme-v01.api.letsencrypt.org/directory --agree-dev-preview 

To obtain a cert using the "webroot" plugin, which can work with the webroot of any webserver software:

  ./letsencrypt-auto certonly -a webroot --webroot-path /var/www/example -d example.com -d www.example.com -server https://acme-v01.api.letsencrypt.org/directory --agree-dev-preview 

NB: curretly the webroot plugin can only obtain certs for several domains simultaneously if they share a webroot.

bmw:
@kuba, are you still working on this? If not, I'll take it over.

PR #1268, opened 2015-10-31
Contet-Type snippets for webroot

kuba:
I'll probably move those docs around after #1137.

cc @centminmod @chriscroome

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -1,4 +1,43 @@
-"""Webroot plugin."""
+"""Webroot plugin.
+
+Content-Type
+------------
+
+This plugin requires your webserver to use a specific `Content-Type`
+header in the HTTP response.
+
+Apache2
+~~~~~~~
+
+.. note:: Instructions written and tested for Debian Jessie. Other
+   operating systems might use something very similar, but you might
+   still need to readjust some commands.
+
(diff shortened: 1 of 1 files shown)


centminmod:
looks good on nginx front :)

letsencrypt/plugins/webroot.py
```diff
@@ -1,4 +1,43 @@
...
+Content-Type
+------------
+
+This plugin requires your webserver to use a specific `Content-Type`
+header in the HTTP response.
+
+Apache2
+~~~~~~~
+
+.. note:: Instructions written and tested for Debian Jessie. Other
```
pde: You could probably go for more abstraction.  Instead of a2enmod (which is Debian-specific), say "enable the headers module".  Then instead of naming the conf-available dir, say, "make this is applied".

For extra fanciness, print the Debian-specific help when `/etc/debian_version` exists, and a more generic version otherwise :)

pde:
Merged, though this could be improved per the above note at some point.

kuba:
Yay, https://letsencrypt.readthedocs.org/en/latest/api/plugins/webroot.html#content-type :)

chriscroome:
This:

```bash
a2enmod headers; a2enconf letsencrypt
```

Should be:

```bash
a2enmod headers; a2enconf letsencrypt-simplehttp
```

If the config file to be symlinked is `/etc/apache2/conf-available/letsencrypt-simplehttp.conf`

PR #1284, opened 2015-11-02
feature: --webroot-path 'default-path, domain1:path1, domain2:path2'

grubberr:
This is continuation of #1259 which I closed because of code mess in my old repo.
I have re-implemented like we discussed in #1259
now only one original command line option --webroot-path
which was improved to allow getting more complex path

1)
```
--webroot-path some-path   # old format still works
```

2)
```
--webroot-path 'default-path, domain1:path1, domain2:path2'
--webroot-path 'domain1:path1, domain2:path2'
```

new possible formats which allow to specify more specific domain path
and ( or ) default-path

#### Complex parsing in _parse_path_argument(self, path):
I am not sure about _parse_path_argument function
probably it looks pretty over engineered
the main idea why this so complex
I wanted to allow possibility to pass webroot pathes which contain 'spaces' or ':', etc

for example

```
--webroot-path 'homelinux.ga:/home/user/public_html/a\ b' # domain path which contains 'space' symbol
--webroot-path '/home/user/public_html/a\\:c' # default path which contains colon
```

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -41,6 +41,7 @@ and reload your daemon.
 import errno
 import logging
 import os
+import shlex
 
 import zope.interface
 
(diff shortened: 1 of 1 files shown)


kuba:
Sorry for not getting to you back earlier and unfortunately I still don't have time to review it. At first glance, though, it looks massively over-engineered :(. There are no unit tests either

kuba:
For anyone else interested in webroot: you might want to try https://github.com/kuba/simp_le instead. It supports multiple SANs and renewal.

pde:
Hi @grubberr I think what @kuba really means when he says "over engineered" is that because neither `:` nor `,` are valid DNS characters, it should be safe to do something along the lines of:

```
entries = ",".split(config.domains)
for e in entires:
    domain, webroot = ":".split(e)
```
etc.  (Though it's worth noting that if we ever want to support certs for IPv6 addresses, ":" will be an issue.)

pde:
More fundamentally, there's a question of how to structure these arguments.  Kuba's approach in `simp_le` is better than trying to squeeze both the default and a bunch of domain <-> webroot mappings into a single command line flag.  But I actually think we can do better still, if we can get argparse to tell us the relative positions of multiple webroot arguments.  Something like this:

`letsencrypt --webroot -w /var/www/example -d example.com -d www.example.com -w /var/www/preposterous -d preposterous.net -d www.preposterous.net`.

That will save the user a lot of typing.

pde:
I've implemented that alternative in #1594 , though I was able to helpfully crib from @grubberr's work in a few places.

bmw:
Thanks for the PR @grubberr, but since #1594 landed, this is no longer necessary.

PR #1288, opened 2015-11-02
Add Exherbo bootstrapping [needs revision]

tgurr:
Pretty much the same like the one which has been merged for Gentoo: https://github.com/letsencrypt/letsencrypt/pull/1221

I currently don't know of any derivatives of Exherbo, but since we're a source-based distribution like Gentoo this is possible.

As a side note, I've also created an exheres for direct installation via our package manager:
https://git.exherbo.org/python.git/tree/packages/dev-python/letsencrypt/letsencrypt-0.0.0.20151024-r1.exheres-0
without the need of having to run the bootstrapping and installing it in a virtualenv.


pde:
LGTM.

bootstrap/exherbo.sh
```diff
@@ -0,0 +1 @@
+_exherbo_common.sh
```
kuba: that symlink is redunand

tgurr: I followed the example in the mentioned pull request from https://github.com/letsencrypt/letsencrypt/pull/1221 and e.g. fedora.sh is also currently unused. Just tell me which one you prefer and I'll update my pull request accordingly if required.

1. leave it as it is
2. leave the symlink and refer to it in install-deps.sh/letsencrypt-auto instead of _exherbo_common.sh
3. rename _exherbo_common.sh to exherbo.sh and refer to that in install-deps.sh/letsencrypt-auto
4. just delete the redundant symlink

kuba: Sorry, #1221 gave a bad example. Please use option 3. As a bonus you could fix Gentoo script! :)

tgurr: I've made the requested changes.

For Gentoo however it is more likely that the _gentoo_common.sh will be required in the near future since there exist some derivatives like e.g. Funtoo and Sabayon.

bootstrap/gentoo.sh
```diff
@@ -1 +0,0 @@
-_gentoo_common.sh
```
cheba: @tgurr Why did you remove the symlink? It looks like a common pattern in the codebase.

There are a lot Gentoo-based distros that this code would work for. For example, Sabayon. I just didn't have it installed to actually test the code. Also the change seem unrelated to Exherbo.

tgurr: @cheba Please see the comment from @kuba in the outdated diff above and also my comment mentioning the same like you did. I've split the changes into 2 commits so they can pick what they want. Personally I just care for adding support for Exherbo.

bmw:
@tgurr, sorry for the delay in looking at this. The client team has been very busy and hasn't been reviewing issues outside of our "1.0 for launch" milestone.

With that said, `letsencrypt-auto` has changed a bit since this PR was written. Basically, we're happy to take community bootstrap packages, but we are putting them behind a call to the `ExperimentalBootstrap` function as you can see in master. If you do this for this Exherbo as well, I'll merge.

bmw:
Appending "[needs revision]" to the title. Please remove when you'd like us to review this again.

bmw:
This PR has sat dead now for over a month so I am closing it. @tgurr, if you'd like this to be merged in, please make another PR putting Exherbo behind a call to `ExperimentalBootstrap`.

Issue #981, opened 2015-10-15
Fails to handle ssl.conf on (some?) Ubuntu vivid systems


pde:
`sudo venv/bin/letsencrypt -t --debug -vvvvv`
...
2015-10-14 19:36:29,059:DEBUG:letsencrypt.plugins.disco:Other error:(PluginEntryPoint#apache): ('There has been an error in parsing the file (%s): %s', u'/etc/apache2/mods-enabled/ssl.conf', u'Syntax error')
...
I'm not sure ssl.conf has been modified on this system.  Anyway:
`cat /etc/apache2/mods-enabled/ssl.conf`

```
IfModule mod_ssl.c>

        # Pseudo Random Number Generator (PRNG):
        # Configure one or more sources to seed the PRNG of the SSL library.
        # The seed data should be of good random quality.
        # WARNING! On some platforms /dev/random blocks if not enough entropy
        # is available. This means you then cannot use the /dev/random device
        # because it would lead to very long connection times (as long as
        # it requires to make more entropy available). But usually those
        # platforms additionally provide a /dev/urandom device which doesn't
        # block. So, if available, use this one instead. Read the mod_ssl User
        # Manual for more details.
        #
        SSLRandomSeed startup builtin
        SSLRandomSeed startup file:/dev/urandom 512
        SSLRandomSeed connect builtin
        SSLRandomSeed connect file:/dev/urandom 512

        ##
        ##  SSL Global Context
        ##
        ##  All SSL configuration in this context applies both to
        ##  the main server and all SSL-enabled virtual hosts.
        ##

        #
        #   Some MIME-types for downloading Certificates and CRLs
        #
        AddType application/x-x509-ca-cert .crt
        AddType application/x-pkcs7-crl .crl

        #   Pass Phrase Dialog:
        #   Configure the pass phrase gathering process.
        #   The filtering dialog program (`builtin' is a internal
        #   terminal dialog) has to provide the pass phrase on stdout.
        SSLPassPhraseDialog  exec:/usr/share/apache2/ask-for-passphrase

        #   Inter-Process Session Cache:
        #   Configure the SSL Session Cache: First the mechanism 
        #   to use and second the expiring timeout (in seconds).
        #   (The mechanism dbm has known memory leaks and should not be used).
        #SSLSessionCache                 dbm:${APACHE_RUN_DIR}/ssl_scache
        SSLSessionCache         shmcb:${APACHE_RUN_DIR}/ssl_scache(512000)
[...]

PR #1291, opened 2015-11-03
[needs revision] Embed Augeas Httpd lens

SwartzCr:
add httpd.aug (from v1.4) to the letsencrypt-apache directory and load it when initializing augeas


SwartzCr:
deals with issue #981

kuba:
Travis is all red :(

SwartzCr:
I know. None of the tests fail locally, I'm trying to debug :/

letsencrypt-apache/letsencrypt_apache/augeas_configurator.py
```diff
@@ -30,7 +33,8 @@ def __init__(self, *args, **kwargs):
         # Set Augeas flags to not save backup (we do it ourselves)
         # Set Augeas to not load anything by default
         my_flags = augeas.Augeas.NONE | augeas.Augeas.NO_MODL_AUTOLOAD
-        self.aug = augeas.Augeas(flags=my_flags)
+        my_dir = os.path.dirname(constants.AUGEAS_HTTPD_LENS)
```
kuba: I'm not sure if it's safe to call `dirname` on the `resource_filename` result... `resource_filename` might create a temporary file (copy it from the distribution to some location on the disk), especially when `zip_safe == True`

kuba: It seems like Augeas really needs directory. https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction: "If the named resource is a directory, then all resources within that directory (including subdirectories) are also extracted". So, please move the lens to a subdirectory (`letsencrypt-apache/letsencrypt_apache/augeas_lenses/httpd.aug`), and use `AUGEAS_LENSES = pkg_resources.resource_filename('letsencrypt_apache', 'lenses')`

kuba: also, please leave a comment about it somewhere in the repo, maybe `letsencrypt-apache/letsencrypt_apache/augeas_lenses/README`?

SwartzCr:
I'm still blocked on this, testing it as a local branch here: https://github.com/SwartzCr/letsencrypt/pull/3
@kuba do you have any ideas?

pde:
I can confirm that the unit tests pass for me locally.

kuba:
I just spun up DO droplet with 12.04 x64. It fails with the same errors. It installs `libauges0==0.10.0-0ubuntu4`, so I guess that is the problem.

kuba:
@SwartzCr, if you still plan on working on this PR, I updated it to use the "right" `pkg_resources` commands at: https://github.com/kuba/letsencrypt/tree/embed-lens. I also merged with `master` - you branched from a veryyyyy old commit - that could cause a lot of problems, so try not to do it in the feature and merge before with `master` without debugging further.

kuba:
https://gist.github.com/kuba/e873180b96de276c04c1

letsencrypt-apache/letsencrypt_apache/augeas_configurator.py
```diff
@@ -1,5 +1,6 @@
 """Class of Augeas Configurators."""
 import logging
+import os
```
pde: This import is causing the lint error; os isn't used after kuba's commits.

kuba:
Does it work on precise? It's not clear, because of the move to trusty (#1358)...

pde:
The apache plugin doesn't work with it on precise, but the client should be okay.  Will double check before merging.

kuba:
So are we effectively dropping support for Ubuntu 12.04 Long Term Support? That does not sound good :(

domcleal:
Precise appears to have Augeas 0.10.0, so that new lens from 1.4.0 won't work with that version of the library.  There was a change/fix to the square lenses in Augeas 1.0.0 so certain newer lenses (Httpd being one) can't be run on 0.x.

pde:
@kuba People on Ubuntu 12.04 will need to use standalone or webroot for now. That's not ideal, but it's better than not being able to [handle the stock config in 15.04+]https://github.com/letsencrypt/letsencrypt/issues/981()

pde:
Did some extra testing.  This seems solid on platforms where augeas 1.2+ is available, and seems to degrade gracefully (no apache plugin) on older ones.

PR #1336, opened 2015-11-04
Content-Type=text/plain for webroot snippets

kuba:


pde:
Explanation for review?

kuba:
We recently switched to `http-01` that uses different `Content-Type` by default. I forgot to change the snippets to reflect it.

bmw:
LGTM

PR #1344, opened 2015-11-04
Remove .pyc files when developer venv is (re)created

### LINKED ISSUE #1342: Persistence of .pyc files causes periodic local tox failures
I just wasted a bunch of time chasing down another instance of this. We should probably rm them in the `dev/venv.sh` script, at the very least.

pde:
Fixes #1342, at least for users who recreate their venv/ dir

### DIFF
--- a/bootstrap/dev/venv.sh
+++ b/bootstrap/dev/venv.sh
@@ -12,4 +12,8 @@ export VENV_ARGS="--python python2"
   -e letshelp-letsencrypt \
   -e letsencrypt-compatibility-test
 
-find . -iname '*.pyc' -exec rm '{}' ';'
+# Workaround for https://github.com/letsencrypt/letsencrypt/issues/1342
+# Ensure we are at the top of a letsencrypt developer tree first
+if grep -q github.com/letsencrypt/letsencrypt .git/config ; then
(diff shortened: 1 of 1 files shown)


kuba:
note: users shouldn't recreate their venv; why would they rerun the script?

bootstrap/dev/venv.sh
```diff
@@ -11,3 +11,5 @@ export VENV_ARGS="--python python2"
   -e letsencrypt-nginx \
   -e letshelp-letsencrypt \
   -e letsencrypt-compatibility-test
+
+find . -iname '*.pyc' -exec rm '{}' ';'
```
devnsec-com: What if someone (accidentally) run `/path/to/bootstrap/dev/venv.sh` from a directory which is not a part of LE but contains `*.pyc` files?

pde: The previous command would error, causing the script to exit due to -e.  But I can double-mitigate this.

devnsec-com: I think `-e` in the shellbang is not really safe enough for this. For an example, if the user types in `bash /path/to/bootstrap/dev/venv.sh`

pde:
Postponing until after **1.0 for launch** is done.

Issue #1270, opened 2015-11-01
Letsencrypt deinstalls postfix on FreeBSD


mfechner:
Install letsencrypt and start it on FreeBSD with:
./letsencrypt-auto --agree-dev-preview --server https://acme-v01.api.letsencrypt.org/directory auth

It killed my complete mailsetup as it starts to install packages using pkg on FreeBSD and does not use the already installed packages.
It uninstalled postfix and installed sendmail, this can be seen as absolutely critical, this needs really to be fixed, as the complete mailsetup on this system is now destroyed and I have to reinstall all manually now.

Suggestion: If you start to install software on the computer with root privileges, ask the user before.
Here the user should get a prompt and endless time to read what he is doing!
I do not even have an idea how letsencrypt can do this, as I was logged in as normal user.

barshinoff:
I don't know why this bug wasn't taken more seriously.

Deinstalling packages seems like a extremely serious bug. All this attention has been given to making sure conf file changes can be undone when you use letsencrypt but you might just break your webserver when you install letsencrypt in the first place and the only comment was have it require --debug.

Break a few webservers and nobody will trust this software.

ohemorange:
There's an official package for FreeBSD, so I'm closing this.

Issue #1196, opened 2015-10-29
Mac install failure (Python errors)


mzsanford:
I have a Mac running OS X 10.10.5 with Homebrew installed already. I attempted install via `git clone` and the `./letsencrypt-auto` with the following results:

```
$ ./letsencrypt-auto
Bootstrapping dependencies for Mac OS X...
WARNING: Mac support is very experimental at present...
Warning: augeas-1.4.0 already installed
Warning: dialog-1.2-20150528 already installed
virtualenv Not Installed
Installing with pip
Requirement already satisfied (use --upgrade to upgrade): virtualenv in /Users/matt/Library/Python/2.7/lib/python/site-packages
Cleaning up...
Creating virtual environment...
./letsencrypt-auto: line 69: virtualenv: command not found
```

This was on `master` at the time, SHA 8e7c8ad31a82db03fc643151de6dd5a80b99eb76. The `bootstrap/mac.sh` script uses `hash virtualenv` to test for the existence of the `virtualenv` package, which returns:

```
$ hash virtualenv
-bash: hash: virtualenv: not found
```

But then attempts an install via `pip` that returns:

```
$ pip install virtualenv
Requirement already satisfied (use --upgrade to upgrade): virtualenv in /Users/matt/Library/Python/2.7/lib/python/site-packages
Cleaning up...
```

I upgraded `virtualenv` as the command suggested (`sudo pip install --upgrade  virtualenv`) and then started getting:

```
$ ./letsencrypt-auto
Bootstrapping dependencies for Mac OS X...
WARNING: Mac support is very experimental at present...
Warning: augeas-1.4.0 already installed
Warning: dialog-1.2-20150528 already installed
Creating virtual environment...
Traceback (most recent call last):
  File "/bin/virtualenv", line 7, in <module>
    from virtualenv import main
ImportError: No module named virtualenv
```

I assume this is somehow related to Python path issues (System vs Homebrew installations). As a non-Python developer attempt to use this tool for my non-Python project I'm not sure how to proceed. I also figured this were one of the kinks that would need to be ironed out before general release, hence filing the issue.

I'll move on to running it one Docker, but on the Mac that means via boot2docker and virtualization so that's hardly a wide audience solution.

pde:
Looks like #963 didn't pan out perfectly. @douglasjose do you think you can help with this?

pde:
Oh, actually, I have one idea: is there by any chance a `python-virtualenv` package that you can install?

[...]

PR #1346, opened 2015-11-04
Require --debug for letsencrypt-auto on platforms thare aren't regularly tested

pde:
Mitigates: #1270 #1196


letsencrypt-auto
```diff
@@ -24,8 +24,24 @@ for arg in "$@" ; do
   if [ "$arg" = "-v" ] || [ "$arg" = "--verbose" ] || echo "$arg" | grep -E -- "-v+$" ; then
     VERBOSE=1
   fi
+  if [ "$arg" = "--debug" ] ; then
```
devnsec-com: Could we use `elif` instead of `fi` + `if`?

pde:
Closing in favour of #1361.

PR #1358, opened 2015-11-05
Test enabling builds on beta Trusty infrastructure [needs revision]

ArchimedesPi:
For sanity's sake, **DO NOT MERGE** without *way* more testing.


pde:
Nice, thank you!

pde:
(We're going to test it... by running our tests with it... and revert it if it actually causes problems).

PR #1360, opened 2015-11-05
Remove another reference to simplehttp

cure:
Since the command `a2enconf letsencrypt` is listed on this page, make sure the config file is named accordingly.


pde:
Thanks.

PR #1257, opened 2015-10-31
auto: 2.6, squeeze (fixes #640, #1046, #1078)

kuba:
Tested on DO squeeze droplet.

PR #1361, opened 2015-11-05
Require --debug for many things, including python2.6 platforms

pde:
This merges #1346 and #1257, and uses a --debug requirement as a protection from too many python2.6-related explosions.


pde:
Tested on Debian wheezy and sid; Ubuntu 12.04 and 15.04; RHEL 7.1.

pde:
One thing I think we should fix: we shouldn't be installing a copy of ConfigArgParse directly from @kuba .  If we need to vendorize it, let's vendorize it.

kuba:
I would much rather prefer if those changes were separated, rather than merged. This makes the review process a lot harder and longer, #1257 was already in the review for past 5 days... It's also hard for me to distinguish my own changes from yours. Moreover, previous review comments from both PRs are "harder to find". Please, may I ask you not do that in future?

That being said, I'm not very in favor of adding another ugly hack such as `--debug`. This scripts grows into some unmanageable shell scripting voodoo :(. Why don't we just remove support for those systems entirely from the `letsencrypt-auto`? Also, the name is super imprecise, it suggests that entire code will be run in debug mode - how about `--experimental-os`?

pde:
@kuba Your point about flags is a good one.  My philosophy thus far has been to only use flags that have well-defined matching semantics in the main python client, so that all of the voodoo is at least coherent.  I considered using `--agree-dev-preview` but then I'd want to change the text to capture the possibility that you're running a totally experimental version of the client.  Perhaps a good solution would be to document this in the --debug cli.py help, so that it says "enable full tracebacks, and allow running the client on experimental platofrms".

As for why this is now one PR, it's because I was okay landing #1257 with these combined changes, but not without them.

bmw:
LGTM

mayasl:
Please devs take a look at this : http://pastebin.com/Gn70tyKK

Issue #942, opened 2015-10-10
Nginx parser needs add_or_set_server_directives


jsha:
Right now Nginx's parser has an add_server_directives method, with a `replace` parameter. If `replace` is true, the directive must already exist in the config, and it will be replaced. If `replace` is false, the directive is added regardless of what exists in the config.

Since most directives are not allowed repeatedly, what we want in many cases is an `add_or_set_server_directives`, which will replace a directive if it's present*, and add it if it's not present.

*To be really tidy, we should say: If the directive is present with the correct value, leave it alone, i.e. don't change its location in the config.

jsha:
Hey @diracdeltas, what would you think if I just changed the semantics of `replace=True` in [parser._add_directives](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-nginx/letsencrypt_nginx/parser.py#L469) so it doesn't error out if there's nothing to replace, and instead just adds the directive?

jsha:
Actually thinking about this some more: since duplicate directives are almost always an error, perhaps instead we should change the semantics of replace=False so that if the directive exists and its value is equal to the one being set, no directive is added; otherwise error.

diracdeltas:
sgtm

jsha:
This was not actually fixed by #1388. That pull request bails out only if there's an exact match for the directive (name and value), but having a match only on name is sufficient to break the config.

Also, side note: Nginx expects to be able to insert error_log and access_log directives, for testing, but there are very likely to cause conflicts with existing configs.

PR #1372, opened 2015-11-05
nginx: add redirect for HTTP traffic

bit:
Once issue #942 is resolved, this should no longer be adding this condition again if it exists.

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/configurator.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/configurator.py
@@ -93,7 +93,7 @@ class NginxConfigurator(common.Plugin):
         # These will be set in the prepare function
         self.parser = None
         self.version = version
-        self._enhance_func = {}  # TODO: Support at least redirects
+        self._enhance_func = {"redirect": self._enable_redirect}
 
         # Set up reverter
         self.reverter = reverter.Reverter(self.config)
@@ -344,7 +344,7 @@ class NginxConfigurator(common.Plugin):
     ##################################
     def supported_enhancements(self):  # pylint: disable=no-self-use
         """Returns currently supported enhancements."""
-        return []
+        return ['redirect']
 
     def enhance(self, domain, enhancement, options=None):
         """Enhance configuration.
@@ -365,6 +365,26 @@ class NginxConfigurator(common.Plugin):
                 "Unsupported enhancement: {0}".format(enhancement))
         except errors.PluginError:
             logger.warn("Failed %s for %s", enhancement, domain)
+
+    def _enable_redirect(self, vhost, unused_options):
+        """Redirect all equivalent HTTP traffic to ssl_vhost.
+
+        Add rewrite directive to non https traffic
+
+        .. note:: This function saves the configuration
+
+        :param vhost: Destination of traffic, an ssl enabled vhost
+        :type vhost: :class:`~letsencrypt_nginx.obj.VirtualHost`
+
+        :param unused_options: Not currently used
+        :type unused_options: Not Available
+        """
+        redirect_block = [[['if', '($scheme != "https")'],
+            [['return', '301 https://$host$request_uri']]
+        ]]
+        self.parser.add_server_directives(vhost.filep, vhost.names,
+            redirect_block)
+        logger.info("Redirecting all traffic to ssl in %s", vhost.filep)
 
     ######################################
     # Nginx server management (IInstaller)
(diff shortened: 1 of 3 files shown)


pde:
The core client team won't be able to review this until we've closed out the [1.0 for launch](https://github.com/letsencrypt/letsencrypt/issues?q=is%3Aopen+is%3Aissue+milestone%3A%221.0+for+launch%22) milestone.  But perhaps @diracdeltas or @erikrose might be able to get started on reviewing this in the meantime?

letsencrypt-nginx/letsencrypt_nginx/parser.py
```diff
@@ -215,7 +215,7 @@ def filedump(self, ext='tmp'):
             try:
                 with open(filename, 'w') as _file:
                     nginxparser.dump(tree, _file)
-            except IOError:
+            except IOError: # pragma: no cover
```
erikrose: It seems a shame to no-cover these. We haven't done that for exception handlers elsewhere in the codebase. Perhaps we can find a way to exercise them without too much painful work.

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -362,6 +362,26 @@ def enhance(self, domain, enhancement, options=None):
         except errors.PluginError:
             logger.warn("Failed %s for %s", enhancement, domain)
 
+    def _enable_redirect(self, vhost, unused_options):
+        """Redirect all equivalent HTTP traffic to ssl_vhost.
+
+        Add rewrite directove to non https traffic
```
erikrose: direct*i*ve :-)

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -362,6 +362,26 @@ def enhance(self, domain, enhancement, options=None):
...
+
+        .. note:: This function saves the configuration
+
+        :param vhost: Destination of traffic, an ssl enabled vhost
+        :type vhost: :class:`~letsencrypt_apache.obj.VirtualHost`
```
erikrose: This probably isn't an apache vhost. :-)

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -362,6 +362,26 @@ def enhance(self, domain, enhancement, options=None):
...
+
+        :param unused_options: Not currently used
+        :type unused_options: Not Available
+        """
+        redirect_block = [[['if', '($scheme != "https")'],
```
erikrose: This works fine, but a more efficient way would be `return 301 https://$host$uri;`. That avoids firing up the comparatively expensive regex machinery (and, for my reference, has been supported since 0.7.51, released in 2009).

We might be able to do even better in the future if we make a fresh vhost for the HTTP side:
```
server {
    server_name example.com;
}
```
That eliminates even the `if` but would require a bit of redesign.

erikrose:
Thanks for the patch! If you have a chance to touch up those few spots, I look forward to merging this in. nginx fans will thank you.

bit:
thanks for the feedback. 
Fixed typos and changed block to a `return 301`.

Looks like test coverage is ok without ignoring the IOError block,
so leaving that out from the first patch.

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -366,6 +366,26 @@ def enhance(self, domain, enhancement, options=None):
...
+        .. note:: This function saves the configuration
+
+        :param vhost: Destination of traffic, an ssl enabled vhost
+        :type vhost: :class:`~letsencrypt_nginx.obj.VirtualHost`
+
+        :param unused_options: Not currently used
+        :type unused_options: Not Available
+        """
+        redirect_block = [[['if', '($scheme != "https")'],
+            [['return', '301 https://$host$uri']]
```
erikrose: Any reason you went with $uri rather than $request_uri? $uri is normalized, changing `/foo/../a` to `/a` (which doesn't matter so much, since browsers do that themselves) but also having some [vaguely documented differences around index files](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_uri) (which might matter, but I can't find a case where it does). My leaning is that $request_uri will have fewer surprises in store for us. (Their additional difference with regard to internal redirects shouldn't matter to us.)

bit: $request_uri seams to be the better choice here.

erikrose:
@pde +1 for merge!

pde:
Great! Thanks @bit & @erikrose !

PR #1375, opened 2015-11-05
docker: Use full filename when copying symlink

### LINKED ISSUE #1374: Dockerfile is broken due to symlink bug in docker
Due to what appears to be a [bug in Docker](https://github.com/docker/docker/issues/17730) affecting either newer versions of Docker or perhaps related to certain storage configurations, the letsencrypt Dockerfile fails to build.

It fails due to an attempt to copy the `ubuntu.sh` symlink into a directory in the container image.

This is a problem that should be fixed upstream by Docker, but a simple workaround is to just specify the full destination filename. Pull request incoming.

johnl:
Works around an upstream bug in Docker:

https://github.com/docker/docker/issues/17730

Fixes #1374

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,8 @@ WORKDIR /opt/letsencrypt
 # If <dest> doesn't exist, it is created along with all missing
 # directories in its path.
 
-COPY bootstrap/ubuntu.sh /opt/letsencrypt/src/
+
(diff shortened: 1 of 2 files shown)


bmw:
I can confirm this is an issue in `docker` 1.9 and is resolved by this PR. Merging.

Thanks for not only reporting the problem but fixing it @johnl!

Issue #1224, opened 2015-10-30
Can't automatically configure Apache configs when using mod_macro


codecat:
I use mod_macro to configure my vhosts, so that I can do this:

```
Use VHost example.com
```

But this is not recognized.

pde:
There are probably going to be lots of cases like this where some Apache module requires special handling.  Step one here would be to provide a clear error message if this situation applies.

Beyond that, actually emulating mod_macro in special cases or all cases sounds like a fun project for someone with strong CS skillz :)

pde:
The error message would be a **Nice for 1.0** ticket.  Emulating mod_macro is definitely a wishlist ticket for now.

AngeloG:
Would be very nice to have! Is there any way to manually set up my Apache config for now, though? (Or maybe it's possible to run the standalone plugin on a different port than 80 or 443?)

joohoi:
I'm working on this currently, it seems to be a good entry issue into the codebase.

pde:
Having landed #1385, I think the remainder of the ticket (actually parsing and editing user macros) is a wishlist ticket.

Hinnack:
Maybe this is starting point:
https://github.com/IMSoP/mod-macro-simulator

cylott:
I'm using mod_macro and currently can't renew my certificates. I am getting the error:
"Unfortunately mod_macro is not yet supported"

jsoref:
I wish there was an `apachectl -S` parser, since at least then one could get the domain for the file.

I tripped on this today. (Commenting because I don't like stale bot)

leonardosahon:
just as @cylott stated, "Unfortunately mod_macro is not yet supported" is what i receive when i try to run certbot for my multiple vhosts, please is there a way around this?

codecat:
This is in no way the answer for everyone, but personally I switched to Caddy for my webserver needs, which does certificates automatically, and you still get that nice mod_macro configuration feel. It's ultimately a lot easier to configure and set up with automatic certificates than Apache is.

Perhaps it's a good alternative for some people in this thread.

PR #1385, opened 2015-11-06
Ignore mod_macro vhosts, and display notification

joohoi:
Work towards #1224 

* Ignore mod_macro vhosts
* Display notification

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -340,12 +340,13 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         """Return all non mod_macro vhosts
 
         :param vhosts: List of VirtualHosts
-        :type vhosts: (:class:`list` of :class:`~letsencrypt_apache.obj.VirtualHost`)
+        :type vhosts: :class:`list` of
+                :class:`~letsencrypt_apache.obj.VirtualHost`
 
         :returns: List of VirtualHosts without mod_macro
-        :rtype: (:class:`list` of :class:`~letsencrypt_apache.obj.VirtualHost`)
+        :rtype: :class:`list` of :class:`~letsencrypt_apache.obj.VirtualHost`
         """
-        return [vh for vh in vhosts if vh.modmacro == False]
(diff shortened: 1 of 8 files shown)


pde:
Great! I'll try to review this today.

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -326,12 +328,26 @@ def _find_best_vhost(self, target_name):
         # No winners here... is there only one reasonable vhost?
         if best_candidate is None:
             # reasonable == Not all _default_ addrs
-            reasonable_vhosts = self._non_default_vhosts()
+            # remove mod_macro hosts from reasonable vhosts
+            reasonable_vhosts = self._without_modmacro(
```
pde: Maybe just do this as a one-liner rather than defining a method? `_without_modmacro` is trivial, and you already have a comment here explaining what it does.

joohoi:
You are right. I'll do that. Less unnecessary lines in both code and tests.

pde:
LGTM.

pde:
I did some testing in the hope of merging this, but I think I've found a probable bug.  I tweaked a port 80 vhost on my system to be macro-ified:

```
<Macro Vhost $host $port $dir>
        <VirtualHost *:$port>
                ServerName $host

                DocumentRoot $dir

                # error, crit, alert, emerg.
                # modules, e.g.
                #LogLevel info ssl:warn

        </VirtualHost>
</Macro>
Use Vhost goxogle.com 80 /var/www/goxogle/

```

That vhost remains in the list of vhosts to be selected, but it now has a blank name:

```
venv/bin/letsencrypt --agree-dev-preview --apache -t 

1:
2: www.isnot.org
3: isnot.org
4: is.isnot.org
Select the appropriate numbers separated by commas and/or spaces (Enter 'c' to
cancel):
```
(previously, entry 1 in the list was `goxogle.com`)

joohoi:
Unfortunately I seem to be unable to reproduce the problem with your exact apache conf example. Did you see an mod_macro notification prior to the domain selection screen at all?

pde:
Hmmm.  I think the error was mine.  This looks good; merging.

PR #1388, opened 2015-11-06
nginx: Only insert directive if its not already in block

### LINKED ISSUE #942: Nginx parser needs add_or_set_server_directives

bit:
fixes #942

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/parser.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/parser.py
@@ -257,6 +257,8 @@ class NginxParser(object):
 
         ..note :: If replace is True, this raises a misconfiguration error
         if the directive does not already exist.
+        ..note :: If replace is False nothing gets added if an identical
+        block exists already.
 
         ..todo :: Doesn't match server blocks whose server_name directives are
             split across multiple conf files.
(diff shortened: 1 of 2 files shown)


pde:
@erikrose this looks like a big win for two lines of code, but I haven't read the nginx codebase so I'm going to wait for your +1 to merge.

erikrose:
Having a look now. (Need to bone up on nginx syntax; it's been a long time.)

erikrose:
I don't see anything that could go wrong because of it, though I'd like to see the invisible "else" case covered with a test and the docstring updated.

bit:
ok, added modified test to check that adding existing blocks works and updated docstring.

erikrose:
Thanks! +1 for merge, @pde.

pde:
Great! Merging.

PR #1282, opened 2015-11-02
[needs revision] Use su if sudo is not available

devnsec-com:
This PR fixes #1148

PR #1392, opened 2015-11-06
Do not assume sudo is installed.

leitzler:
Do not assume sudo is installed.


devnsec-com:
As we discussed before, the `letsencrypt-auto` script should _**NOT**_ run as root.
See: #1148, #1282 and #1191 for more information.

PR #1395, opened 2015-11-06
apache: add general http-header enhacement [needs revision]

### LINKED ISSUE #582: Provide options for HSTS
It would be great to have an option to automatically enable **Http Strict Transport Securiy Option**

e.g. for nginx:
```
add_header Strict-Transport-Security max-age=31536000;
```

### LINKED ISSUE #812: Add Content-Security-Policy
We should really support CSP, because they are great:
- we can avoid mixed content
- we can automatically upgrade insecure requests
- we could avoid click-jacking
- webmasters could use even more restrictive rules

sagi:
Closes #582
Mostly closes #812 (though --uir isn't on or prompted by default) and #1611 (but no HPKP for now)


pde:
Thanks @sagi!  Preliminarily: can you structure this so that the enhancement is called "http-header", and it's up to the client to call it with `Strict-Transport-Security <blah>` as an argument?  That way we'll get other features like `Upgrade-Insecure-Requests` at the same time.

sagi:
Sure. I'm on it.

sagi:
Now we've a general http-header enhancement.  There's a new dict in constants.py where for each header_name (i.e. Strict-Transport-Security, Upgrade-Insecure-Requests and maybe in the future HPKP) there is a list of header arguments.

letsencrypt-apache/letsencrypt_apache/constants.py
```diff
@@ -27,3 +27,15 @@
 REWRITE_HTTPS_ARGS = [
     "^", "https://%{SERVER_NAME}%{REQUEST_URI}", "[L,QSA,R=permanent]"]
 """Apache rewrite rule arguments used for redirections to https vhost"""
+
+
+HSTS_ARGS = ["always", "set", "Strict-Transport-Security",
+    "\"max-age=31536000; includeSubDomains\""]
```
pde: Is there any reason for this level of detail to be inside the plugins themselves?  In the long run my thought for the "right" way to do HSTS enablement in interactive mode would be to start it off with a very short TTL, and gradually increase it if the admin has left the header in place.  But freedom to implement that kind of functionality would require the "details" of the headers to be determined in the main client, not the plugins.

(I'm being a little picky about this because once we've picked a particular set of semantics for then "http-header" enhancement that essentially becomes an API between the client and plugins).

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -731,6 +732,67 @@ def enhance(self, domain, enhancement, options=None):
...
+        self.save()
+        logger.info("Adding %s header to ssl vhost in %s", header_name,
+                ssl_vhost.filep)
+
+    def _verify_no_http_header(self, ssl_vhost, header_name):
+        """Checks to see if existing header_name header is in place.
+
+        Checks to see if virtualhost already contains a header_name header
+
+        :param vhost: vhost to check
```
pde: Nit `param` should be `ssl_vhost`?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -731,6 +732,67 @@ def enhance(self, domain, enhancement, options=None):
...
+
+        self.save_notes += ("Adding %s header to ssl vhost in %s\n" %
+                (header_name, ssl_vhost.filep))
+
+        self.save()
+        logger.info("Adding %s header to ssl vhost in %s", header_name,
+                ssl_vhost.filep)
+
+    def _verify_no_http_header(self, ssl_vhost, header_name):
+        """Checks to see if existing header_name header is in place.
```
pde: I'm not actually sure that `header_name` is the right description of the variable you're working with here.  If you have a header like this:

`Content-Security-Policy: script-src 'self' ajax.googleapis.com facebook.net ; upgrade-insecure-requests`

I think that standard terminology is that `Content-Security-Policy` is the header name, `script-src 'self' ajax.googleapis.com facebook.net ; upgrade-insecure-requests` is the header value, and `script-src` and `upgrade-insecure-requests` are either directives or substrings in that header.

So this variable might be `header_substring`, or `header_or_directive`, `apache_header_field`, or something like that?  And the function might be better named `verify_no_matching_http_header`?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -731,6 +732,67 @@ def enhance(self, domain, enhancement, options=None):
...
+        :rtype: (bool)
+
+        :raises errors.PluginError: When header header_name exists
+
+        """
+        header_path = self.parser.find_dir("Header", None, start=ssl_vhost.path)
+        if header_path:
+            # "Existing Header directive for virtualhost"
+            for match in header_path:
+                if self.aug.get(match).lower() == header_name.lower():
```
pde: The use of equality as the test here is slightly dangerous.  For instance, if I have an vhost entry like this:
```
Header add "Content-Security-Policy" "upgrade-insecure-requests"
```
then this test will return false, because it's comparing the strings `"upgrade-insecure-requests"` and `upgrade-insecure-requests`.  So you wind up with this in your conf file after `letsencrypt install --uir`:

```
Header add "Content-Security-Policy" "upgrade-insecure-requests"
Header always set Content-Security-Policy upgrade-insecure-requests
```
I don't know what a more correct alternative is, though.  Substring matching doesn't work either because there could be something like this in the conf file:
```
Header append Content-Security-Policy "script-src 'self' upgrade-insecure-requests.info"
``` 
We don't want to have to write parsers for every type of header we support, or at least not yet, so some zen philosophy might be required in working around this issue.

letsencrypt/client.py
```diff
@@ -354,57 +354,82 @@ def deploy_certificate(self, domains, privkey_path,
...
+        uir = config.uir # Upgrade Insecure Requests
+
         if redirect is None:
             redirect = enhancements.ask("redirect")
 
-        # When support for more enhancements are added, the call to the
-        # plugin's `enhance` function should be wrapped by an ErrorHandler
         if redirect:
-            self.redirect_to_ssl(domains)
+            self.apply_enhancement(domains, "redirect")
```
pde: Each of these calls to `apply_enhancement` should be wrapped in a `try: catch PluginError` block that suppresses the exception that is raised if the enhancement is already in place.

Otherwise, if a domain already redirects from HTTP to HTTPS, and the user runs `letsencrypt install --redirect --uir --hsts`, this call to `apply_enhancement` will raise `PluginError: Let's Encrypt has already enabled redirection` and not continue to apply the other enhancements.

pde: Or perhaps stop raising exceptions in those cases, and just pass a note to the user saying that we found the thing in place and aren't doing it again.

letsencrypt-apache/letsencrypt_apache/configurator.py
pde: If we have to do something quick-and-dirty for now, options include:
* the current test (equality)
* equality after stripping any known Apache quotations and escapes
* substring match
* some kind of clever regexp substring match, like `([ "]|^)upgrade-insecure-requests([ "]|$)` (ie upgrade-insecure-requests is a substring that's separated from the things around it by whitespace, quotes, or the ends of the string)

pde:
Overall, this is great!  Most of my comments should be easy fixes; the question about header matching is tricky; I'd like to hear your thoughts about the least-incorrect way to do it for now.

letsencrypt-apache/letsencrypt_apache/configurator.py
sagi: I agree. Changed `header_name` to `header_substring` and  `verify_no_http_header` to `verify_no_matching_http_header`.

letsencrypt-apache/letsencrypt_apache/configurator.py
sagi: Good catch. I used the following pattern: 
`pat = '(?:[ "]|^)(%s)(?:[ "]|$)' % (header_substring.lower())`

letsencrypt-apache/letsencrypt_apache/constants.py
sagi: We discussed it on `#letsencrypt-dev`: We'll create an API wrapper on top of augeas for a more versatile http header placements (for instance: adaptive HSTS max-age / HPKP). In the mean time we'll keep the simple http-header placement method that is contained in the PR.

letsencrypt/client.py
pde: Actually, this is probably the wrong place to do this, because if we're enhancing N domains, we want to try the later ones even if the enhancement has already been applied to the first.

letsencrypt/client.py
```diff
@@ -383,57 +383,82 @@ def deploy_certificate(self, domains, privkey_path,
...
-               "however, we successfully installed your certificate.")
+        msg = ("We were unable to set up enhancement %s for your server, "
+               "however, we successfully installed your certificate."
+               % (enhancement))
         with error_handler.ErrorHandler(self._recovery_routine_with_msg, msg):
             for dom in domains:
                 try:
-                    self.installer.enhance(dom, "redirect")
+                    self.installer.enhance(dom, enhancement, options)
                 except errors.PluginError:
```
pde: something like this here?
```
except PluginError, err:
    if "already enabled redirection" in repr(err) or re.search("Existing.*header", repr(e)):
        logger.info("Enhancement %s for %s already in place", enhancement, dom)
    else:
        # current error handling code including `raise`
```

pde:
I think we should rename the enhancement from "http-header" to "ensure-http-header", both to capture its semantics and to leave room for the day when we deprecate it.

Other than that and the exception handling code I suggested above, this is ready to merge.

letsencrypt/client.py
pde: Per IRC:
```
<sagi__> pdeee: I'll change it to ensure-http-header. Regarding the not crashing if enhancements are already in place -
<sagi__> It's not that easy
<pdeee> not least, because of bmw's great error handling code :)
<pdeee> for your new header calls, I think we know that no recovery routine is required
<sagi__> yes, the problem is with the redirect
<pdeee> we need to check in the case of redirect...
<pdeee> so if redirect needs real recovery, what we'd do is:
<pdeee> move the "with error_handler.ErrorHandler" logic inside of the for loop
<pdeee> create a new error handler type, which knows about these non-fatal enhancement errors
<pdeee> and cleans up if required but doesn't exit out when they happen
```

bmw:
@pde, I can take a look at this tomorrow morning EST if you'd like.

pde:
Seems to be working well in tests.  Merging!

Issue #1400, opened 2015-11-07
Convince ourselves that temp_install() isn't really a security vuln, or fix it


pde:
There's an extremely dramatic security warning comment in the function [temp_install](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/configurator.py#L1252) in the apache plugin.

This function performs a filesystem copy operation.  AFAICT the source file is controlled by `setuptools` and the destination is within the apache configuration directory controlled by the command line and/or a Let's Encrypt configuration file.

Since Let's Encrypt is not designed to perform privileged operations for unprivileged users, command lines or configuration files, I believe this is an utter non-issue.  But I'd like someone to check my reasoning on this.

pde:
Actually I have a PR for this, but I'd still like James's input if possible.

PR #1390, opened 2015-11-06
Remove nonsensical Apache config directives

pde:
- AcceptPathInfo Off was breaking things (fixes: #1274)
- Enabling DEFLATE compression for HTML and XML may make the site vulnerable to BREACH!

PR #1401, opened 2015-11-07
Downgrade a very excitable security comment to an XXX

pde:
Resolves: #1400

Also note that this function will need to become more sophisticated as we
support autoupdating of config options (eg those changed in #1390)

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -162,7 +162,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         # Get all of the available vhosts
         self.vhosts = self.get_virtual_hosts()
 
-        temp_install(self.mod_ssl_conf)
+        install_ssl_options_conf(self.mod_ssl_conf)
 
     def deploy_cert(self, domain, cert_path, key_path,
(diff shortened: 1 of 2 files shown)


bmw:
I'm merging this since it's only really an aesthetic/comment change and has no effect on program execution.

@jdkasten, we'd love your thoughts here to know if there is a real security vuln here that we're not thinking of and these changes should be reverted.

jdkasten:
This is performing the duty of a package manager.

It is not a problem as we no longer check for privilege escalation. (That
is exactly what this is)

Apache configs are Turing complete... so whoever has access to the file has
complete root access to your system.

You are free to do what you want with the info though.

bmw:
It's true that if someone modified `options-ssl-apache.conf`, `letsencrypt` will blindly copy and include that file in Apache. It's also true that `options-ssl-apache.conf` has the same permissions as our all of our Python files so an attacker could just modify our code to do whatever they want. This is at least the case when cloning this repo or when the files are installed using `letsencrypt-auto`.

In general, I think if an attacker is able to modify any files installed with `letsencrypt`, the user is pretty much screwed and I don't think there is much we can do to defend against that. Let me know if I'm thinking about this the wrong way.

jdkasten:
Yep, you are correct, the current installation path already assumes that to
be the case.

The comment was outdated and should be removed.

Issue #1397, opened 2015-11-06
User-Agent: should allow computation of client success rates by platform and plugin.


pde:
In order to make strategic decisions about plugin prioritisation (engineering prioritisation, whether to ship or not ship each plugin in `letsencrypt-auto`, and which ones to prefer in the UI), we need to know which ones are succeeding for people.

That means we should send a small amount of plugin and platform information by default in the client's User Agent string, so that we can compute aggregate success rates on the server side.  Probably it should be which plugins are in use (`Authenticator: Standalone; Installer: None`) and platform major version (`Debian 8`). We should have a flag to disable it, for people who prefer not to reveal anything at all about their server software.

pde:
Flagging this urgent because the sooner it's shipped, the sooner we can prioritise 1.0 tickets for Apache vs webroot (vs standalone?).

pde:
In terms of implementation:
* we  learn the plugins [here](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/cli.py#L426). You get their names by doing something like `authenticator.name`, provided authenticator is not `None`.
* we do something clever to determine the platform and major version
* we can pass them over to the `acme` module as optional keyword arguments, so as not to break uses of that module that don't pass along User Agent information.
* we ensure that acme's [_send_requests](https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/client.py#L561) helper uses the User Agent we want.

pde:
@yaseminacar @sfahl

kuba:
Duplicate of #858.

xamnesiax-sx:
@pde https://github.com/xamnesiax-sx/letsencrypt/commit/9de2cc0f2480f1e22458377d126c32a22bafafd8 I've mostly implemented it, need some clarification on a few things. I'll hit you up on IRC tomorrow

pde:
Here is what the final User Agent should perhaps look like:

```
User-Agent: LetsEncryptPythonClient/0.1.0.dev0 (Debian 8) Authenticator/standalone Installer/none
```

pde:
The client version ("0.1.0dev0" in my example above) can be found [here](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/__init__.py).

pde:
Closed in #1455 .  Thanks for your help on this @xamnesiax-sx and @bmw !

Issue #858, opened 2015-09-29
Add custom useragent


hlieberman:
Hi @Kuba,

It would be very useful for us to be able to track what versions of the clients people are making requests with.  I propose that we override the requests UA to something of the format "Let's Encrypt Agent v$(version) plugin $(apache,nginx,standalone)".

If we do that, I'll make sure that we override that code somehow to pass the Debian version in so that we can track that separately.  We can just have the version number in a single file and incr it there, I think.

FYI @jsha

jmhodges:
This would be good to prioritize so that we can make sure folks are using a letsencrypt install with the latest challenges. We're going to remove the simpleHTTP and DVSNI ones soon and having this data sooner rather than later will be helpful.

centminmod:
what about https://community.letsencrypt.org/t/standalone-manual-webroot-stats/2535/7 ? related ?

bmw:
Fixed in #1455

PR #1404, opened 2015-11-07
User-Agent support in acme (default: `acme-python`, fixes #1351).

### LINKED ISSUE #1351: Don't use the default UA when verifying challenges
While related to #858, I feel like this is a slightly different issue. We have had reports of people running into problems with the User-Agent when using the `manual` plugin. `manual` uses `acme` to validate that the challenge has been set up correctly before responding to the server. This is done using the `requests` library whose default UA as of writing this e-mail is `python-requests/2.8.1`. Some webservers are rejecting requests with this UA, causing self validation to fail, and `letsencrypt` to exit.

kuba:
In order to override the default (`acme-python`), clients (including Let's Encrypt: #858, #1397) should create a custom `acme.client.ClientNetwork` object and pass it to `acme.client.Client.__init__`.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -481,11 +481,13 @@ class ClientNetwork(object):
     JSON_ERROR_CONTENT_TYPE = 'application/problem+json'
     REPLAY_NONCE_HEADER = 'Replay-Nonce'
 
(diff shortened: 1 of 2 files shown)


pde:
Great! Thanks Kuba.

Issue #914, opened 2015-10-06
Implement latest ACME spec changes for challenges


kuba:
Soon to be pulled in https://github.com/letsencrypt/boulder/pull/774 implements the following ACME spec changes:
- https://github.com/ietf-wg-acme/acme/pull/6
- https://github.com/ietf-wg-acme/acme/pull/7
- https://github.com/ietf-wg-acme/acme/pull/8
- https://github.com/ietf-wg-acme/acme/pull/13

https://github.com/letsencrypt/boulder/pull/774 doesn't break integration tests, but follow-up PRs that disable old challenges will break.

I'll work on `acme` and `letsencrypt`, using #895 as base (please review!). Volunteers are necessary for `letsencrypt-apache` and `letsencrypt-nginx` follow-up PRs.

bmw:
If @jdkasten doesn't have the time, I'm happy to make the changes for `letsencrypt-apache`. As for `letsencrypt-nginx`, if we're able to invest the development time necessary to make the plugin ready for launch, I'd be happy to makes these changes as well. If we're not, I'm not sure if getting these changes incorporated to Nginx plugin is a high priority.

jdkasten:
I can make the changes for Apache.  It should be fairly easy.

kuba:
Opened a follow-up https://github.com/letsencrypt/letsencrypt/issues/1417.

PR #1411, opened 2015-11-07
tls-sni-01

kuba:
Another step towards #914.


acme/acme/challenges.py
```diff
@@ -318,89 +319,36 @@ def validation(self, account_key):
...
+    typ = "tls-sni-01"
 
     DOMAIN_SUFFIX = b".acme.invalid"
     """Domain name suffix."""
 
-    PORT = DVSNI.PORT
-    """Port to perform DVSNI challenge."""
-
-    validation = jose.Field("validation", decoder=jose.JWS.from_json)
+    PORT = 443
```
pde: Is hardcoding this going to mess with inegration tests?

acme/acme/challenges.py
```diff
@@ -318,89 +319,36 @@ def validation(self, account_key):
...
-
-        """
-        # Instance of 'Field' has no 'signature' member
-        # pylint: disable=no-member
-        return hashlib.sha256(self.validation.signature.encode(
-            "signature").encode("utf-8")).hexdigest().encode()
+    def z(self):
+        """``z`` value used for verification."""
+        return hashlib.sha256(
+            self.key_authorization.encode("utf-8")).hexdigest().encode()
```
pde: Note: we assume `hexidgest` is always lowercase, which is probably fine but may deserve a comment.

pde:
Otherwise, LGTM.

acme/acme/challenges.py
kuba: No, this the default ACME port. You can pass `port` to `simple_verify` to override it. Added an inline comment.

acme/acme/challenges.py
kuba: Interestingly, Python docs doesn't guarantee it to be lower case, so I've added `.lower()`.

https://docs.python.org/2/library/hashlib.html#hashlib.hash.hexdigest

kuba:
Fixed comments. Ready to merge.

PR #1413, opened 2015-11-07
Custom acme.messages.Error (fixes #946).

### LINKED ISSUE #946: Give friendlier errors on new error types from server
Currently, if the client receives an unrecognized error type from the server (e.g. rate limited), it gives an unfriendly error:

```
An unexpected error occurred.
ClientError: (<Response [429]>, DeserializationError("Could not decode 'typ' (u'urn:acme:error:rateLimited'): Deserialization error: Error type not recognized",))
```

Instead, the client should expect that error types will be added that it does not recognize, and report them to the user similarly to other errors, including the error detail message.

I think this is fairly high priority to fix, because otherwise it prohibits Boulder from using meaningful error types.

kuba:


kuba:
@pde, please merge? :)

bmw:
The one downside to this approach is now all of the ACME error messages we show users are prefixed with `urn:acme:error`. Is this the behavior we want?

bmw:
I think a better approach is if `urn:acme:error` is present, it should be removed. Otherwise, the type should not be modified.

kuba:
I think removing it when it's present is a confusing behavior, especially when someone starts sending `unauthorized` alongside `urn:acme:error:unauthorized`... You can always fix the display code in the core to remove the prefix, but I absolutely see no idea why this PR hasn't been pulled for the last 3 weeks, especially given that it fixes compliance issu. Can we finally merge it, please?

pde:
Merging.  Working around the verbosity in PR...

PR #1415, opened 2015-11-08
Fix ExperimentalBootstrap on Mac

pde:
Fixes: #1408

PR #1418, opened 2015-11-08
do not use brew to install pip; use python instead

mitnk:
Mac OS X already shipped python (2.7) with it at the most of time.
Use brew version python may cause a lot of other issues.

This PR needs #1415 merged first to be functional.

### DIFF
--- a/bootstrap/mac.sh
+++ b/bootstrap/mac.sh
@@ -8,11 +8,15 @@ brew install augeas
 brew install dialog
 
 if ! hash pip 2>/dev/null; then
-    echo "pip Not Installed\nInstalling python from Homebrew..."
-    brew install python
+    echo "pip Not Installed. Downloading..."
+    curl -so get-pip.py 'https://bootstrap.pypa.io/get-pip.py'
+    echo sudo python get-pip.py
+    sudo python get-pip.py
+    rm get-pip.py
 fi
 
 if ! hash virtualenv 2>/dev/null; then
     echo "virtualenv Not Installed\nInstalling with pip"
-    pip install virtualenv
+    echo sudo pip install virtualenv
+    sudo pip install virtualenv
 fi


kuba:
`sudo python` / `sudo pip` messes with OS directories. We should not do that.

mitnk:
Yeah. But I think install `pip/virtualenv` on system-wide if user did not is still better to install the whole python things by ignoring the existing python ;)

I'm ok to close this PR if we have to use brew.

kuba:
`letsencrypt-auto` is a temporary solution before we get real OS packaging. However, we don't want to mess with system too much ourselves. BTW what is standard packaging for Mac?

mitnk:
There's no standard official packaging tool for Mac. `brew` is the most popular one through developers.

kuba:
Would you like to work on a brew recipe for letsencrypt?

mitnk:
I'm not familiar with brew. And letsencrypt is not like `wget` or `openssl` as a package, it's more like a env needing python, pip, virtualenv, and building a virtual env out. So I'm not sure it's suitable to be a brew package.

yumitsu:
@kuba I can take it, if you don't mind.

kuba:
Not at all, please do! :)

yumitsu:
@kuba It seems somebody else already did it, but PR is still on hold (https://github.com/Homebrew/homebrew/pull/44033). The problem is, `letsencrypt` is still behind any 'stable' version (even `0.0.1`), and such formulas should be in `homebrew-devel-only`.
I can proceed on formula if you're OK with that.

bmw:
@yumitsu I think that's fine, I'm not very familiar with brew, but if `bootstrap/mac.sh` is still able to use the formula if it is in `homebrew-devel-only`, it shouldn't be a problem.

I'm closing this issue since the decision to install a second copy of Python as opposed to system-wide `pip` was discussed and made before merging #963.

PR #1419, opened 2015-11-08
Extend webroot plugin to support rsync [needs revision]

tomhughes:
Allow the specified webroot path to be an rsync URL and cause the webroot files to be published to that URL as necessary.

This also provides a general framework for publishing the webroot content to a remote system.

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -2,7 +2,12 @@
 import errno
 import logging
 import os
+import shutil
+import subprocess
+import tempfile
 from collections import defaultdict
+from urlparse import urlparse
+from itertools import groupby
 
 import zope.interface
 import six
@@ -45,6 +50,8 @@ to serve all files under specified web root ({0})."""
 
     def __init__(self, *args, **kwargs):
         super(Authenticator, self).__init__(*args, **kwargs)
+        self.paths = {}
+        self.temp_dirs = {}
         self.full_roots = {}
         self.performed = defaultdict(set)
 
@@ -57,8 +64,17 @@ to serve all files under specified web root ({0})."""
                 "--webroot-path and --domains, or --webroot-map. Run with "
                 " --help webroot for examples.")
         for name, path in path_map.items():
-            if not os.path.isdir(path):
-                raise errors.PluginError(path + " does not exist or is not a directory")
+            self.paths[name] = urlparse(path)
+            if self.paths[name].scheme == "" or self.paths[name].scheme == "file":
+                if not os.path.isdir(path):
+                    raise errors.PluginError(path + " does not exist or is not a directory")
+            elif self.paths[name].scheme == "rsync":
+                if path not in self.temp_dirs:
+                    self.temp_dirs[path] = tempfile.mkdtemp()
+                path = self.temp_dirs[path]
+            else:
+                raise errors.PluginError(self.paths[name].scheme + " URLs not supported")
+
             self.full_roots[name] = os.path.join(path, challenges.HTTP01.URI_ROOT_PATH)
 
             logger.debug("Creating root challenges validation dir at %s",
@@ -98,7 +114,9 @@ to serve all files under specified web root ({0})."""
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
         assert self.full_roots, "Webroot plugin appears to be missing webroot map"
-        return [self._perform_single(achall) for achall in achalls]
+        responses = [self._perform_single(achall) for achall in achalls]
+        self._sync_root(achalls)
+        return responses
 
     def _get_root_path(self, achall):
         try:
@@ -142,8 +160,8 @@ to serve all files under specified web root ({0})."""
             os.remove(validation_path)
             self.performed[root_path].remove(achall)
 
-        for root_path, achalls in six.iteritems(self.performed):
-            if not achalls:
+        for root_path, path_achalls in six.iteritems(self.performed):
+            if not path_achalls:
                 try:
                     os.rmdir(root_path)
                     logger.debug("All challenges cleaned up, removing %s",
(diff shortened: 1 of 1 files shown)


kuba:
`webroot`, originally named `SimpleFS`, was by design meant to interact with the file system rather than calling arbitrary command. If you want to copy stuff over `SSH`, you can use `sshfs`.

If we really want this functionality (calling arbitrary command) in the core, I think it should be a separate plugin.

tomhughes:
A separate plugin was my original thought, and I'm happy to go that way if people prefer, though there would obviously be a degree of duplication.

Equally I considered various remote filesystem options, but rejected NFS as unsuitable for non-local operation and sshfs as hard to secure compared to an rsync module that can easily be configured to operate as an unprivileged, chrooted user restricted to just allow writing to the web root by a limited set of clients.

The goal here, as you have likely guessed, is to be able to automate renewal of certificates on machines that are not running web servers, or at least not publically reachable web servers.

bmw:
First off, thanks for the PR @tomhughes. I think functionality like this is something we'd want at some point. I'm also leaning towards another plugin (that perhaps shares a common base with webroot), but I could be convinced after reviewing the code that this approach is better.

With that said, due to our launch schedule, we probably will not consider this until after we've fixed launch critical issues. I'm labeling this PR as such and we'll take a look at it then.

centminmod:
@bmw i had a idea for much later in the future https://github.com/letsencrypt/letsencrypt/issues/1223 heh

iveoles:
Is it possible be could add multiple paths to copy to? Being able to rsync to 2-3 folders at the same time would fix issues I currently have trying to automate on a cluster.

tomhughes:
Well my game plan was to only have one server handling ACME challenges - basically I created an `acme.example.com` domain for that and had everything else redirect ACME challenges to that domain which is where the webroot I publish to will be.

Basically for any site that I want to be able to issue certificates for I add to my apache config:

```
RedirectPermanent /.well-known/acme-challenge http://acme.example.com/.well-known/acme-challenge
```

iveoles:
Doesn't that mean you need to set up a new domain alias on all sites and new DNS A Record?

If so it doesn't quite fit with the level of automation the rest of the project is aiming at.

tomhughes:
Well for my personal stauff I created one new A record and configured a vhost for it on a web server and then added a top level redirect (outside of any vhost) to the two web servers involved that will apply to any domain hosted on those servers.

I am likely to do something similar for the openstreetmap domains, except in that case there are dozens of web servers but I can roll the redirect out to them all using our existing chef automation.

iveoles:
@tomhughes you are a steely-eyed missile man! 

Thanks to your tips I have a fully automated solution for my cluster.

pde:
I've actually been thinking that we should review and consider a version of this PR as part of webroot.  It seems to be fundamentally part of that plugin's ambit: I have a webroot directory here, it just happens to be shared with some other systems.

What I'm not sure about is whether we should stay specific to rsync, or consider a more general case, which is, "once the webroot is configured to pass challenges, run command X to push it to deployment".  That could be rsync, but it could also conceivably be other operations, such as clearing a cache, tweaking a firewall, etc.  So we `--webroot` could have flags like `--webroot-pre-validate` and `--webroot-post-validate` that take scripts as arguments, and are run before and after DV.  A `--rsync` flag could exist in addition, and use most of the same mechanics internally (a crafted `subprocess.Popen` rather than a raw call to a script).

@tomhughes what do you think, and are you still interested in working on this?

kuba:
This idea might make sense, but I don't think it makes sense to amend `webroot` - you should create new plugin instead (they could share common code though).

tomhughes:
@pde I made it use URL syntax precisely because I envisioned that people might want other deployment methods that could use other URL types - ftp, scp, http or whatever. I'm happy to have another look - I know it needs some work because of the changes for multiple webroots.

@kuba So you're thinking abstract some of the current web root code into a base class that both can inherit from?

pde:
I'm pretty agnostic on whether this is a separate plugin that inherits from webroot, or whether it's it's own plugin.  I think that's a code style and legibility question, and arguments can be made either way.  The main considerations are how the user encounters this functionality:

* Is the CLI clean and consistent for users across those cases?  Answer would probably be yes in either case because as it turns out we moved `--webroot-path` / `-w` parsing into the main cli.py file for other reasons, and `webroot-map` can be shared between multiple webroot-y plugins.
* Will the user have to at some hypothetical point choose between "webroot" and "rsync" plugins, or add a `--rsync` or `--pre-validation-script` flag to the the webroot plugin?  I think I like the idea of making `--pre-validation-script` a general webroot option, and that weakly implies doing this without inheriting from webroot?
* Probably we should also `--rsync $TARGET` which both picks the plugin and sets the rsync destination.  We can have that UI with either underlying implementation stategy, I think.

pde:
@tomhughes please proceed with whichever implementation strategy you think is cleaner.  @kuba may have some further arguments in favour of inheritance, but he's going to be offline for a week or so, so you may not want to block on them.

tomhughes:
As a starting point I have rebased my patch on current master, adding support for multiple webroots.

So for now this is still a patch to the existing plugin rather than a new one.

bmw:
So after talking about this a lot with the client team, we decided against adding an explicit `rsync` plugin or adding `rsync` functionality to `webroot`. Instead, we're planning to rework the `manual` plugin, allowing the user to provide scripts that can be used for the purposes of solving and cleaning up challenges (see #2782). Using this, the equivalent functionality of a `rsync` plugin can be implemented in a couple lines of `bash`.

Regardless, thanks for the PR @tomhughes and I'm sorry it sat around so long.

bmw:
I wanted to add to this that although we're not taking this plugin into our tree, we certainly would like to support the development of third party plugins for Certbot. If you'd like to host this on your own repo, we'll happily link to it in our [list of third party plugins](https://certbot.eff.org/docs/using.html#third-party-plugins).

PR #1423, opened 2015-11-08
Appropriately set and update Apache certificate directives

### LINKED ISSUE #1213: letsencrypt-auto install didn't remove SSLCertificateChainFile statement
I'm using Apache on Debian. I already had TLS configured, and thus had `SSLCertificateFile`, `SSLCertificateKeyFile` and `SSLCertificateChainFile` statements in my Apache site config.

letsencrypt-auto correctly replaced the values of the first two, but failed to change `SSLCertificateChainFile`, resulting in breakage.

Invocation:

```
./letsencrypt-auto install --apache -d example.com -d www.example.com --cert-path /etc/letsencrypt/live/example.com/fullchain.pem --key-path /etc/letsencrypt/keys/0000_key-letsencrypt.pem --agree-dev-preview 
```

Site config before invocation:

```
<VirtualHost *:80>
        ServerName example.com
        ServerAlias www.example.com
        DocumentRoot ...

        ErrorLog ...
        CustomLog ...
</VirtualHost>

<VirtualHost *:443>
        ServerName example.com
        ServerAlias www.example.com
        DocumentRoot ...

        ErrorLog ...
        CustomLog ...

        SSLEngine on
        SSLCertificateFile ...
        SSLCertificateKeyFile ...
        SSLCertificateChainFile ...
</VirtualHost>
```

Apart from that, it worked nicely, and so we now finally have what CACert has kept promising for "next year" for the past 10 years or so. So thanks for all the work, letsencrypt is amazing! ❤

### LINKED ISSUE #1266: Unable to install certficiate to Apache
I tried to obtain and install a certificate for Apache today and while I was able to obtain one, I wasn't able to install it. Apart from the fact that the command in the beta invite e-mail was using auth which is afaik not installing the certificate and you can't simply say “okay now install it”.

Anyway I got the following error:

    Cannot find a cert or key directive in
    /files/etc/apache2/sites-available/ssl.conf/virtualhost.
    VirtualHost was not modified

And I'm wondering what exactly that means. Now looking at the [source code](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/configurator.py#L203) it seems that `ssl.conf` already needs to contain `SSLCertificateFile` and `SSLCertificateKeyFile` which seem odd because if they are already there… I wouldn't need to install it.

Now for full disclosure, I already had tried to use SSL so the file already had `SSLCertificateFile` defined (but not the `SSLCertificateKeyFile`). So maybe that's the issue?

ArchimedesPi:
Also use fullchain.pem on Apache versions that support it.  From this IRC log:
```
2015-11-02 16:31:29	@pdeee	for >= 2.4.8:
2015-11-02 16:32:23	@pdeee	add new SSLCertificateFile pointing to fullchain.pem
2015-11-02 16:33:10	@pdeee	remove all preexisting SSLCertificateFile, SSLCertificateChainFile, SSLCACertificatePath, and possibly other fields subject to careful research :)
2015-11-02 16:33:21	@pdeee	for < 2.4.8:
2015-11-02 16:34:03	@pdeee	add SSLCertificateFile pointing to cert.pem
2015-11-02 16:34:42	@pdeee	and SSLCertificateChainFile pointing to chain.pem
2015-11-02 16:34:50	xamnesiax	gotcha
2015-11-02 16:34:55	@pdeee	remove all preexisting/conflicting entries
2015-11-02 16:35:19	xamnesiax	Am I correct to assume that this can all be done from deploy_certs in the apache configurator?
2015-11-02 16:36:32	xamnesiax	deploy_cert *
2015-11-02 16:36:48	@pdeee	I think so
2015-11-02 16:36:59	@pdeee	again, jdkasten may wish to say more
```

Will fix #1213, hopefully will fix #1266.

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -213,21 +213,10 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         # Assign the final directives; order is maintained in find_dir
         if self.version >= (2, 4, 8):
-            logger.debug("Apache version (%s) is >= 2.4.8",
-                         ".".join(map(str,self.version)))
-            for directive in ["SSLCertificateKeyFile", "SSLCertificateChainFile",
-                              "SSLCACertificatePath"]:
-                logging.debug("Trying to delete directive '%s'", directive)
-                directive_tree = self.parser.find_dir(directive, None, vhost.path)
-                logging.debug(directive_tree)
-                if directive_tree:
-                    logger.debug("Removing directive %s", directive)
-                    self.aug.remove(re.sub(r"/\w*$", "", directive_tree[-1]))
-            logging.debug("fullchain path: %s", fullchain_path)
             self.aug.set(path["cert_path"][-1], fullchain_path)
         elif self.version < (2, 4, 8):
-            logger.debug("Apache version (%s) is < 2.4.8",
-                         ".".join(map(str,self.version)))
+            self.aug.set(path["cert_path"][-1], cert_path)
+            self.aug.set(path["chain_path"][-1], chain_path)
 
         # Save notes about the transaction that took place
         self.save_notes += ("Changed vhost at %s with addresses of %s\n"
@@ -583,6 +572,10 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         # Update Addresses
         self._update_ssl_vhosts_addrs(vh_p)
 
+        # Remove existing SSL directives
+        logging.info("Removing existing SSL directives")
+        self._remove_existing_ssl_directives(vh_p)
+
         # Add directives
         self._add_dummy_ssl_directives(vh_p)
 
@@ -650,6 +643,16 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             ssl_addrs.add(ssl_addr)
 
         return ssl_addrs
+
+    def _remove_existing_ssl_directives(self, vh_path):
+        for directive in ["SSLCertificateKeyFile", "SSLCertificateChainFile",
+                          "SSLCACertificatePath", "SSLCertificateFile"]:
+            logger.debug("Trying to delete directive '%s'", directive)
+            directive_tree = self.parser.find_dir(directive, None, vh_path)
+            logger.debug("Parser found %s", directive_tree)
+            if directive_tree:
+                    logger.debug("Removing directive %s", directive)
+                    self.aug.remove(re.sub(r"/\w*$", "", directive_tree[-1]))
 
     def _add_dummy_ssl_directives(self, vh_path):
         self.parser.add_dir(vh_path, "SSLCertificateFile",
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -214,7 +214,8 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                      ".".join(str(i) for i in self.version))
 
         if self.version < (2, 4, 8) or (chain_path and not fullchain_path):
-            # install SSLCertificateFile, SSLCertificateKeyFile, and SSLCertificateChainFile directives
+            # install SSLCertificateFile, SSLCertificateKeyFile,
+            # and SSLCertificateChainFile directives
             set_cert_path = cert_path
             self.aug.set(path["cert_path"][-1], cert_path)
             self.aug.set(path["cert_key"][-1], key_path)
@@ -660,7 +661,8 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         return ssl_addrs
 
-    def _remove_existing_ssl_directives(self, vh_path, minus={}):
+    def _remove_existing_ssl_directives(self, vh_path, minus=None):
+        minus = minus or {}
         directives_to_remove = list({"SSLCertificateKeyFile", "SSLCertificateChainFile",
                                      "SSLCertificateFile"} - set(minus))
         for directive in directives_to_remove:
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -192,20 +192,22 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         path["cert_path"] = self.parser.find_dir(
             "SSLCertificateFile", None, vhost.path)
+        path["cert_key"] = self.parser.find_dir(
+            "SSLCertificateKeyFile", None, vhost.path)
 
         # Only include if a certificate chain is specified
         if chain_path is not None:
             path["chain_path"] = self.parser.find_dir(
                 "SSLCertificateChainFile", None, vhost.path)
 
-        if not path["cert_path"]:
+        if not path["cert_path"] or not path["cert_key"]:
             # Throw some can't find all of the directives error"
             logger.warn(
-                "Cannot find a cert directive in %s. "
+                "Cannot find a cert or key directive in %s. "
                 "VirtualHost was not modified", vhost.path)
             # Presumably break here so that the virtualhost is not modified
             raise errors.PluginError(
-                "Unable to find cert directive")
+                "Unable to find cert and/or key directives")
 
         logger.info("Deploying Certificate to VirtualHost %s", vhost.filep)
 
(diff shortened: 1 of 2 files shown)


ArchimedesPi:
@pde: This isn't really working. Apache barfs on using just `SSLCertificateChain /etc/letsencrypt/live/maybe.isnot.org/fullchain.pem` with
```
[Sun Nov 08 23:38:18.039573 2015] [ssl:error] [pid 13001:tid 140086295459648] AH02579: Init: Private key not found
[Sun Nov 08 23:38:18.039711 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D0680A8:asn1 encoding routines:ASN1_CHECK_TLEN:wrong tag
[Sun Nov 08 23:38:18.039731 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D08303A:asn1 encoding routines:ASN1_TEMPLATE_NOEXP_D2I:nested asn1 error
[Sun Nov 08 23:38:18.039747 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D0680A8:asn1 encoding routines:ASN1_CHECK_TLEN:wrong tag
[Sun Nov 08 23:38:18.039763 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D07803A:asn1 encoding routines:ASN1_ITEM_EX_D2I:nested asn1 error (Type=RSA)
[Sun Nov 08 23:38:18.039778 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:04093004:rsa routines:OLD_RSA_PRIV_DECODE:RSA lib
[Sun Nov 08 23:38:18.039793 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D0680A8:asn1 encoding routines:ASN1_CHECK_TLEN:wrong tag
[Sun Nov 08 23:38:18.039808 2015] [ssl:error] [pid 13001:tid 140086295459648] SSL Library Error: error:0D07803A:asn1 encoding routines:ASN1_ITEM_EX_D2I:nested asn1 error (Type=PKCS8_PRIV_KEY_INFO)
[Sun Nov 08 23:38:18.039818 2015] [ssl:emerg] [pid 13001:tid 140086295459648] AH02311: Fatal error initialising mod_ssl, exiting. See /var/log/apache2/error.log for more information
[Sun Nov 08 23:38:18.039826 2015] [ssl:emerg] [pid 13001:tid 140086295459648] AH02564: Failed to configure encrypted (?) private key maybe.isnot.org:443:0, check /etc/letsencrypt/live/maybe.isnot.org/fullchain.pem
AH00016: Configuration Failed
```
(ubuntu 12.04, Apache 2.4.16)

ArchimedesPi:
Working on apache 2.4.16!

pde:
Looks like this just needs a couple of tests...

ArchimedesPi:
@pde I'm working on them. Everything should be green by tomorrow evening, if there are no hangups.

ArchimedesPi:
Tests are done. Can someone review?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -643,6 +656,16 @@ def _update_ssl_vhosts_addrs(self, vh_path):
 
         return ssl_addrs
 
+    def _remove_existing_ssl_directives(self, vh_path):
+        for directive in ["SSLCertificateKeyFile", "SSLCertificateChainFile",
+                          "SSLCACertificatePath", "SSLCertificateFile"]:
```
pde: I don't think we shouldn't remove `SSLCACertificatePath`; that is used for authenticating client certificates.

pde: That's kind of subtle because of the insane cross-entanglement between client cert verification and chain building as documented [here](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatechainfile).  "Hopefully" nobody has `SSLCACertificatePath` if they don't use client certs, but in any case leaving that directive around shouldn't be a cause of breakage.  I hope.

pde:
This LGTM.  Ready to merge with one of the following:

* A second reviewer saying +1
* @ArchimedesPi or someone else saying they have done a bunch of tests in deployment (in addition to the unit tests)

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -212,12 +212,21 @@ def deploy_cert(self, domain, cert_path, key_path,
         logger.info("Deploying Certificate to VirtualHost %s", vhost.filep)
 
         # Assign the final directives; order is maintained in find_dir
-        self.aug.set(path["cert_path"][-1], cert_path)
-        self.aug.set(path["cert_key"][-1], key_path)
-        if chain_path is not None:
+        if self.version >= (2, 4, 8):
```
pde: An additional aspect to this test will need to be, "do we know the fullchain path?"  That will always be the case for `letsencrypt run`, but currently it won't be true for `letsencrypt install`, where we only know the path to `cert.pem` and `chain.pem`.  The best way to handle that is just to fall back to the deprecated case I think.

pde: (`letsencrypt install` needs to start using storage.py, but currently it does not).

ArchimedesPi:
@pde Tested and working on a polluted SSL vhost with configuration *very* similar to #1213.

pde:
Also closes: #1541

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -210,16 +210,28 @@ def deploy_cert(self, domain, cert_path, key_path,
...
-        if chain_path is not None:
-            if not path["chain_path"]:
-                self.parser.add_dir(
-                    vhost.path, "SSLCertificateChainFile", chain_path)
-            else:
-                self.aug.set(path["chain_path"][-1], chain_path)
+        logger.debug("Apache version is %s",
+                     ".".join(str(i) for i in self.version))
+
+        if self.version < (2, 4, 8):
```
pde: `or (cert_path and not fullchain_path)`?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -210,16 +210,28 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+            # install SSLCertificateFile, SSLCertificateKeyFile,
+            # and SSLCertificateChainFile directives
+            set_cert_path = cert_path
+            self.aug.set(path["cert_path"][-1], cert_path)
+            self.aug.set(path["cert_key"][-1], key_path)
+            if chain_path is not None:
```
pde: shouldn't this have an `else` branch that raises an error asking for `--chain-path`?

letsencrypt-apache/letsencrypt_apache/configurator.py
pde: I guess one could argue that we should nudge people towards doing things the "right" way for Apache 2.4.  But conversely we could argue that if someone sets up automated scripts calling `letsencrypt install` on an Apache 2.2 system, they shouldn't break just because Apache was upgraded to 2.4.  So I think add `or (cert_path and not fullchain_path)`

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -262,7 +276,15 @@ def choose_vhost(self, target_name):
...
             return vhost
 
-        return self._choose_vhost_from_list(target_name)
+        vhost = self._choose_vhost_from_list(target_name)
+        if vhost.ssl:
+            # remove duplicated or conflicting ssl directives
+            self._deduplicate_directives(vhost.path,
+                ["SSLCertificateFile", "SSLCertificateKeyFile"])
+            # remove all problematic directives
+            self._remove_directives(vhost.path, ["SSLCertificateChainFile"])
```
pde: In my testing, this `_remove_directives` call appears to not be working.  I created a test situation where my apache config contained these lines:
```
...
SSLCertificateFile /tmp/cert.pem
SSLCertificateKeyFile /tmp/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
SSLCertificateChainFile /tmp/chain.pem
...
```
Running the code from this branch (with master merged in), I got this resulting file:
```
<IfModule mod_ssl.c>
<VirtualHost *:443>
        ServerName isnot.org

        DocumentRoot /var/www/isnot

        # error, crit, alert, emerg.
        # modules, e.g.
        #LogLevel info ssl:warn

SSLCertificateFile /etc/letsencrypt/live/isnot.org-0001/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/isnot.org-0001/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
SSLCertificateChainFile /tmp/chain.pem
</VirtualHost>
</IfModule>
```
(And the resulting site is transvalid :( ).
```

pde:
Ok, this LGTM.  Currently cli.py doesn't match this PR in terms of what install flags it accepts, but I'll fix that in a separate PR.

PR #1436, opened 2015-11-09
Implemented useragent debugging code for #1397

xamnesiax-sx:
Information regarding the Authenticator and Installer, as well as basic Operating System info (OS / Dist + version) is now sent in the ACME HTTP request, via the useragent. New flag has been added (--user-agent), which allows for both the customization of the user agent, and the disabling of data collection via the user agent by Let's Encrypt's servers.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -615,7 +615,6 @@ class ClientNetwork(object):
         else:
             raise errors.MissingNonce(response)
 
-
     def _get_nonce(self, url):
         if not self._nonces:
             logging.debug('Requesting fresh nonce')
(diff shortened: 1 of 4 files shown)


acme/acme/client.py
```diff
@@ -626,3 +627,43 @@ def post(self, url, obj, content_type=JSON_CONTENT_TYPE, **kwargs):
         response = self._send_request('POST', url, data=data, **kwargs)
         self._add_nonce(response)
         return self._check_response(response, content_type=content_type)
+
+    def set_user_agent(self, keyval_map):
```
kuba: please move it to `letsencrypt` module, together with `user_agent_object`. It's not suitable for general purpose ACME library.

PR #1443, opened 2015-11-10
Help text for -d flag, and refactoring the long name from plural to singular form

### LINKED ISSUE #1442: Help text for -d (--domains) command line flag and the plural form of the long name
Seems like the short help isn't descriptive enough to point the user to the right direction when using multiple domain names, as seen here #1383 . 

Also the plural form of the long name is a bit deceptive, as it actually takes only one domain as an argument.

Touches also the subject of #1376 , any suggestions for the actual help text are appreciated.

joohoi:
* Added help text for -d flag
* Refactored the long name --domains to --domain to be semantically correct

Fixes #1442 

Still uses the old key "domains" in the internal code, to not to break existing configuration files etc.

### DIFF
--- a/letsencrypt-nginx/tests/boulder-integration.sh
+++ b/letsencrypt-nginx/tests/boulder-integration.sh
@@ -18,7 +18,7 @@ letsencrypt_test_nginx () {
         "$@"
 }
 
-letsencrypt_test_nginx --domains nginx.wtf run
+letsencrypt_test_nginx --domain nginx.wtf run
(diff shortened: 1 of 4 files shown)


kuba:
LGTM, good to merge.

However, what we should really do is to allow `--domains domain1.com,domain2.com` (i.e. parse it ourselves instead of relying on `action='append'`). This would e.g. allow multiple domains to be specified in the config file.

bmw:
LGTM. Thanks for another PR @joohoi.

PR #1453, opened 2015-11-10
Update Dockerfile

thomaschristory:
On docker 1.9, the copy of the ubuntu.sh file fails if the full path of destination is not specified


bmw:
This issue was fixed in #1375 which I just merged. Thanks for the PR though @thomaschristory!

PR #1455, opened 2015-11-11
Add a User Agent string for client analytics

pde:
Closes #1436 #858 
Building on @xamnesiax-sx 's work in #1436.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -169,6 +169,7 @@ def _determine_account(args, config):
 
 
 def _init_le_client(args, config, authenticator, installer):
+    config.deterimine_user_agent(authenticator, installer)
     if authenticator is not None:
         # if authenticator was given, then we will need account...
         acc, acme = _determine_account(args, config)
@@ -521,6 +522,7 @@ def revoke(args, config, unused_plugins):  # TODO: coop with renewal config
         logger.debug("Revoking %s using Account Key", args.cert_path[0])
         acc, _ = _determine_account(args, config)
         # pylint: disable=protected-access
+        config.determine_user_agent(None, None)
         acme = client._acme_from_config_key(config, acc.key)
     acme.revoke(jose.ComparableX509(crypto_util.pyopenssl_load_certificate(
         args.cert_path[1])[0]))
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -557,6 +557,7 @@ def plugins_cmd(args, config, plugins):  # TODO: Use IDisplay rather than print
     logger.debug("Filtered plugins: %r", filtered)
 
     if not args.init and not args.prepare:
+        print str(filtered)
         return
 
     filtered.init(config)
@@ -564,17 +565,19 @@ def plugins_cmd(args, config, plugins):  # TODO: Use IDisplay rather than print
     logger.debug("Verified plugins: %r", verified)
 
     if not args.prepare:
+        print str(verified)
         return
 
     verified.prepare()
     available = verified.available()
     logger.debug("Prepared plugins: %s", available)
+    print str(available)
 
 
 def read_file(filename, mode="rb"):
     """Returns the given file's contents.
 
-    :param str filename: Filename
+    :param str filename: filename as an absolute path
     :param str mode: open mode (see `open`)
 
     :returns: A tuple of filename and its contents
@@ -584,6 +587,7 @@ def read_file(filename, mode="rb"):
 
     """
     try:
+        filename = os.path.abspath(filename)
         return filename, open(filename, mode).read()
     except IOError as exc:
         raise argparse.ArgumentTypeError(exc.strerror)
@@ -676,7 +680,31 @@ class HelpfulArgumentParser(object):
         parsed_args = self.parser.parse_args(self.args)
         parsed_args.func = self.VERBS[self.verb]
 
+        parsed_args.domains = self._parse_domains(parsed_args.domains)
         return parsed_args
+
+    def _parse_domains(self, domains):
+        """Helper function for parse_args() that parses domains from a
+        (possibly) comma separated list and returns list of unique domains.
+
+        :param domains: List of domain flags
+        :type domains: `list` of `string`
+
+        :returns: List of unique domains
+        :rtype: `list` of `string`
+
+        """
+
+        uniqd = None
+
+        if domains:
+            dlist = []
+            for domain in domains:
+                dlist.extend([d.strip() for d in domain.split(",")])
+            # Make sure we don't have duplicates
+            uniqd = [d for i, d in enumerate(dlist) if d not in dlist[:i]]
+
+        return uniqd
 
     def determine_verb(self):
         """Determines the verb/subcommand provided by the user.
@@ -819,7 +847,11 @@ def prepare_and_parse_args(plugins, args):
     # --domains is useful, because it can be stored in config
     #for subparser in parser_run, parser_auth, parser_install:
     #    subparser.add_argument("domains", nargs="*", metavar="domain")
-    helpful.add(None, "-d", "--domains", metavar="DOMAIN", action="append")
+    helpful.add(None, "-d", "--domains", dest="domains",
+                metavar="DOMAIN", action="append",
+                help="Domain names to apply. For multiple domains you can use "
+                "multiple -d flags or enter a comma separated list of domains"
+                "as a parameter.")
     helpful.add(
         None, "--duplicate", dest="duplicate", action="store_true",
         help="Allow getting a certificate that duplicates an existing one")
@@ -888,7 +920,6 @@ def prepare_and_parse_args(plugins, args):
     # parser (--help should display plugin-specific options last)
     _plugins_parsing(helpful, plugins)
 
-
     return helpful.parse_args()
 
 
@@ -939,26 +970,28 @@ def _paths_parser(helpful):
     if verb in ("install", "revoke", "certonly"):
         section = verb
     if verb == "certonly":
-        add(section, "--cert-path", default=flag_default("auth_cert_path"), help=cph)
+        add(section, "--cert-path", type=os.path.abspath,
+            default=flag_default("auth_cert_path"), help=cph)
     elif verb == "revoke":
         add(section, "--cert-path", type=read_file, required=True, help=cph)
     else:
-        add(section, "--cert-path", help=cph, required=(verb == "install"))
+        add(section, "--cert-path", type=os.path.abspath,
+            help=cph, required=(verb == "install"))
 
     section = "paths"
     if verb in ("install", "revoke"):
         section = verb
     # revoke --key-path reads a file, install --key-path takes a string
-    add(section, "--key-path", type=((verb == "revoke" and read_file) or str),
-        required=(verb == "install"),
+    add(section, "--key-path", required=(verb == "install"),
+        type=((verb == "revoke" and read_file) or os.path.abspath),
         help="Path to private key for cert creation or revocation (if account key is missing)")
 
     default_cp = None
     if verb == "certonly":
         default_cp = flag_default("auth_chain_path")
-    add("paths", "--fullchain-path", default=default_cp,
+    add("paths", "--fullchain-path", default=default_cp, type=os.path.abspath,
         help="Accompanying path to a full certificate chain (cert plus chain).")
-    add("paths", "--chain-path", default=default_cp,
+    add("paths", "--chain-path", default=default_cp, type=os.path.abspath,
         help="Accompanying path to a certificate chain.")
     add("paths", "--config-dir", default=flag_default("config_dir"),
         help=config_help("config_dir"))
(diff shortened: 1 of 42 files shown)


pde:
(For now I've done things without touching `acme/` per @kuba 's request, although I think that API is a little inelegant).

letsencrypt/cli.py
```diff
@@ -890,7 +892,13 @@ def _create_subparsers(helpful):
     helpful.add_group("revoke", description="Options for revocation of certs")
     helpful.add_group("rollback", description="Options for reverting config changes")
     helpful.add_group("plugins", description="Plugin options")
-
+    helpful.add(
+        "certonly", "--user-agent", type=str, default=None,
```
bmw: Question: Why is this only for `certonly`?

letsencrypt/configuration.py
```diff
@@ -3,6 +3,7 @@
 import urlparse
 import re
 
+import letsencrypt
```
bmw: nit: Our normal import order is modules from the standard library, modules from our dependencies, followed by our own modules. With this in mind, I'd put this right before all of the `from letsencrypt...`, preserving the space between the `from acme import challenges` line above it.

letsencrypt/configuration.py
```diff
@@ -82,6 +84,21 @@ def http01_port(self):  # pylint: disable=missing-docstring
         else:
             return challenges.HTTP01Response.PORT
 
+    def determine_user_agent(self, authenticator, installer):
```
bmw: I wonder if this is the best place for this function. In general, this class is just a wrapper around `argparse.Namespace`.

What do you think of the approach of modifying this function so it returns the UA and put it in `client.py`. When `letsencrypt.client.Client` is initiated, it has the `authenticator` and `installer`. `__init__` can call this function and pass it into `_acme_from_config_key` which should take the UA as an optional argument. Thoughts?

letsencrypt/le_util.py
```diff
@@ -202,6 +203,37 @@ def safely_remove(path):
             raise
 
 
+def get_os_info():
+    """
+    Get Operating System type/distribution and major version
```
bmw: We should document what this function returns.

letsencrypt/le_util.py
```diff
@@ -202,6 +203,37 @@ def safely_remove(path):
...
+    """
+    info = platform.system_alias(
+        platform.system(),
+        platform.release(),
+        platform.version()
+    )
+    os_type, os_ver, _ = info
+    os_type = os_type.lower()
+    if os_type.startswith('linux'):
+        info = platform.linux_distribution()
```
bmw: FYI, `platform.linux_distribution()` returns `('', '', '')` on my Arch linux laptop.

letsencrypt/tests/cli_test.py
```diff
@@ -30,19 +32,19 @@ def setUp(self):
...
-        from letsencrypt import cli
-        args = ['--text', '--config-dir', self.config_dir,
-                '--work-dir', self.work_dir, '--logs-dir', self.logs_dir,
-                '--agree-dev-preview'] + args
+        args = self.standard_args + args
         with mock.patch('letsencrypt.cli.sys.stdout') as stdout:
             with mock.patch('letsencrypt.cli.sys.stderr') as stderr:
                 with mock.patch('letsencrypt.cli.client') as client:
-                    ret = cli.main(args)
+                    ret = cli.main(args[:]) # NOTE: parser can alter its args!
```
bmw: Is this necessary? `args` is a local variable.

letsencrypt/tests/cli_test.py
```diff
@@ -112,9 +111,37 @@ def test_help(self):
...
+                "-d", "example.com", '--agree-tos'] + self.standard_args
+        det.return_value = mock.MagicMock(), None
+        with mock.patch('letsencrypt.cli.client.acme_client.ClientNetwork') as acme_net:
+            cli.main(args[:])  # Protect args from alteration
+            os_ver = " ".join(le_util.get_os_info())
+            ua = acme_net.call_args[1]["user_agent"]
+            self.assertTrue(os_ver in ua)
+            import platform
+            plat = platform.platform()
+            if "linux" in plat.lower():
```
bmw: One minor downside to code like this is the coverage of `cli_test.py` varies depending on the platform it's running on.

letsencrypt/tests/cli_test.py
```diff
@@ -10,8 +10,10 @@
 import mock
 
 from letsencrypt import account
+from letsencrypt import cli
```
bmw: nit: In the rest of the repo, we only import the module we're testing in the specific functions where we need to do it.

letsencrypt/tests/cli_test.py
```diff
@@ -112,9 +111,37 @@ def test_help(self):
...
         self.assertTrue(cli.usage_strings(plugins)[0] in out)
 
+
+    @mock.patch('letsencrypt.cli.sys.stdout')
+    @mock.patch('letsencrypt.cli.sys.stderr')
+    @mock.patch('letsencrypt.cli.client.acme_client.Client')
+    @mock.patch('letsencrypt.cli._determine_account')
+    @mock.patch('letsencrypt.cli.client.Client.obtain_and_enroll_certificate')
+    @mock.patch('letsencrypt.cli._auth_from_domains')
+    def test_user_agent(self, _afd, _obt, det, _client, _err, _out):
```
bmw: nitty nit that you probably shouldn't take the time to fix: You could create a version of `_call` that doesn't mock `letsencrypt.cli.client` and have `_call` use that version internally.

letsencrypt/cli.py
pde: It's not only for certonly, but it should be displayed when the user types `--help certonly`.  Arguably it could be one of those arguments that moves around, though I hate those.  Or maybe we should institutionalise them, and have the HelpfulArgumentParser know about flags that are in many topics...

pde: OK I've opened #1460 for that.

pde: In the meantime I think this PR is sort of okay as is. `letsencrypt --help run` is currently pretty unhelpful, but that's larger than this PR, and we need #1460 to solve it I think.

letsencrypt/tests/cli_test.py
pde: I guess it is unnecessary because [this line](https://github.com/letsencrypt/letsencrypt/pull/1455/files#diff-1333bfbf9a96ab6aa2587c1ca0cbd2b3R43) protects the caller's copy of the list from being modified in place.  I still find that a terrifying side effect...

letsencrypt/tests/cli_test.py
pde: Yup.  I was coding for travis, which I think is good enough for now.

letsencrypt/tests/cli_test.py
pde: Is there a reason to do it that way?

There were a ridiculous number copies of `from letsencrypt import cli` in this file, and I was also unsure about the effect of mock decorators running *before* the imports that they were mocking.  But I guess that happens in other place without breaking.

letsencrypt/tests/cli_test.py
pde: Yeah, not doing that now :)

letsencrypt/tests/cli_test.py
pde: Leaving it there just in case anyone ever refactors the `args` logic and needs to be protected from the consequences.

pde:
I think I've addressed everything except possibly the `from letsencrypt import cli` thing; lmk what you think about that (or just merge :))

letsencrypt/configuration.py
pde: I think you're probably right, that would be more elegant.  Not sure if we should block on it, though...

pde: Why don't we say, this is on my list of things to try to do today, but if it happens after merge/release it's a refactor ticket for the renewal milestone?

letsencrypt/cli.py
bmw: Why should it be shown for `certonly`? It's not a big deal, but I feel like the best place for it is the ungrouped optional arguments such as `--email`.

letsencrypt/le_util.py
bmw: This doesn't quite work. If nothing else, we need a new line before `:returns:` or the documentation isn't formatted properly.

Additionally, `grep` the project for `rtype: tuple` to see how other functions with similar return values have been documented.

bmw:
Added some follow up comments.

letsencrypt/cli.py
```diff
@@ -425,14 +423,23 @@ def choose_configurator_plugins(args, config, plugins, verb):
...
         diagnose_configurator_problem("authenticator", req_auth, plugins)
 
+    record_chosen_plugins(config, plugins, authenticator, installer)
     return installer, authenticator
 
 
+def record_chosen_plugins(config, plugins, auth, inst):
+    "Update the config entries to reflect the plugins we actually selected."
+    cn = config.namespace
+    cn.authenticator = plugins.find_init(auth).name if auth else "none"
```
bmw: I think you should only set the value if a plugin has been selected. This is what the `renewer` expects.

letsencrypt/cli.py
```diff
@@ -512,18 +519,19 @@ def install(args, config, plugins):
 
 def revoke(args, config, unused_plugins):  # TODO: coop with renewal config
     """Revoke a previously obtained certificate."""
+    # For user-agent construction
+    config.namespace.installer = config.namespace.authenticator = "none"
```
bmw: Change above would allow you to remove this.

letsencrypt/cli.py
```diff
@@ -921,7 +929,13 @@ def _create_subparsers(helpful):
     helpful.add_group("revoke", description="Options for revocation of certs")
     helpful.add_group("rollback", description="Options for reverting config changes")
     helpful.add_group("plugins", description="Plugin options")
-
+    helpful.add(
+        None, "--user-agent", type=str, default=None,
```
bmw: Why specify `type=str`? This is the default ([source](https://docs.python.org/2/library/argparse.html#type)).

letsencrypt/client.py
```diff
@@ -31,10 +33,29 @@
 logger = logging.getLogger(__name__)
 
 
-def _acme_from_config_key(config, key):
+def acme_from_config_key(config, key):
+    "Wrange ACME client construction"
```
bmw: Typo

letsencrypt/client.py
```diff
@@ -31,10 +33,29 @@
...
+                                    user_agent=_determine_user_agent(config))
+    return acme_client.Client(directory=config.server, key=key, net=net)
+
+
+def _determine_user_agent(config):
+    """
+    Set a user_agent string in the config based on the choice of plugins.
+    (this wasn't knowable at construction time)
+
+    :rtype: `str`
```
bmw: Nit: I'd add something like `:returns: user-agent` above this line.

letsencrypt/tests/cli_test.py
```diff
@@ -113,9 +112,38 @@ def test_help(self):
...
+            if "linux" in plat.lower():
+                self.assertTrue(platform.linux_distribution()[0] in ua)
+
+        with mock.patch('letsencrypt.cli.client.acme_client.ClientNetwork') as acme_net:
+            ua = "bandersnatch"
+            args += ["--user-agent", ua]
+            cli.main(args)
+            acme_net.assert_called_once_with(mock.ANY, verify_ssl=True, user_agent=ua)
+
+        self._call(['install', '--domain', 'foo.bar', '--cert-path', 'cert',
```
bmw: Looks like the merge got messed up. Tests are currently failing because of this line.

letsencrypt/cli.py
```diff
@@ -512,18 +519,19 @@ def install(args, config, plugins):
 
 def revoke(args, config, unused_plugins):  # TODO: coop with renewal config
     """Revoke a previously obtained certificate."""
+    # For user-agent construction
```
bmw: Tests for this function would be nice since the majority of the lines changed. I had some on my old absolute paths [PR](https://github.com/letsencrypt/letsencrypt/blob/273e08275b910f31a1967d91e77bf80df2513788/letsencrypt/tests/cli_test.py#L235) you should be able to use with little to no modification.

bmw:
I added some additional comments, but they're all fairly minor changes that should be easy to fix. In general, like this approach much better.

letsencrypt/cli.py
pde: The problem is that `--installer` and `--authenticator` also set things inside that namespace.  Perhaps the new value can never be `None` unless the old one was, but I'd need to convince myself of that...

bmw: Good point! I checked the `renewer` code and it correctly handles the case where authenticator/installer are absent and when they have an unexpected value (e.g. `"none"`). With that in mind, I'm fine with this approach.

bmw:
LGTM

PR #1467, opened 2015-11-11
Improve comments for letsencrypt-auto

pde:
I think some people who read `letsencrypt-auto` are unfamiliar with the `VAR=${VAR:-NEWVALUE}` syntax, and conclude that it is less customisable than it is.  So document that.


pde:
Merging a zero-risk PR.

PR #1468, opened 2015-11-11
Fix misplaced verbosity from pip

pde:
When configargparse is [not] being upgraded, `letsencrypt-auto` prints a bunch of ugly messages about it.


pde:
Merging an epsilon-risk PR.

PR #1470, opened 2015-11-11
Reenable containerised Travis infrastructure

ArchimedesPi:
We're having builds error on the beta trusty infrastructure, so we're trying switching back.


jmhodges:
This made my branch work, as well! It's nice.

PR #1471, opened 2015-11-12
Move the nginxparser copyright statement to letsencrypt-nginx

fmarier:
I couldn't find anything related to the nginxparser in the base letsencrypt package. I assume this copyright notice is only relevant to the letsencrypt-nginx package.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -2,6 +2,13 @@ Let's Encrypt Python Client
 Copyright (c) Electronic Frontier Foundation and others
 Licensed Apache Version 2.0
 
+The nginx plugin incorporates code from nginxparser
+Copyright (c) 2014 Fatih Erikli
+Licensed MIT
+
+
+Text of Apache License
+======================
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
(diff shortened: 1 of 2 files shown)


pde:
Does Debian use the root-level LICENSE file?  I'd prefer to keep that one comprehensive for the whole tree, though I'm fine with making the subsidiary ones more specific.

pde:
`git cherry-pick 2a820a391fef8e3c2a12afbd87cbc16e8bfa0399 ` will fix the travis failure.

fmarier:
> Does Debian use the root-level LICENSE file? I'd prefer to keep that one comprehensive for the whole tree, though I'm fine with making the subsidiary ones more specific.

No, we have 4 different packages now, so we use whatever is in each tarball on PyPI.

Currently the "root" LICENSE file is part of the letsencrypt tarball. The LICENSE file in the letsencrypt-nginx package doesn't mention the nginxparser.

pde:
So let's keep the additive portion of this PR, but revert the piece that removes text from the root level license (it'd be fine to clarify that the MIT/nginxparser piece is specifically letsencrypt-nginx).

fmarier:
@pde I've restored the text in the top-level file but clarified that it only applies to the nginx plugin.

pde:
Thanks!

PR #1474, opened 2015-11-12
Check if ports are available at an earlier point in time (fixes #1216)

### LINKED ISSUE #1216: Warning about port in use should appear earlier

metachris:
Minor modification of `cli.py` to show an error asap if the user chose StandaloneAuthenticator but the ports are not available.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -37,6 +37,7 @@ from letsencrypt import storage
 from letsencrypt.display import util as display_util
 from letsencrypt.display import ops as display_ops
 from letsencrypt.plugins import disco as plugins_disco
+from letsencrypt.plugins.standalone import Authenticator as StandaloneAuthenticator
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


metachris:
Any thoughts on this?

bmw:
Thanks for the PR @metachris!  Currently, the client team is really only reviewing PRs that resolve issues in the "1.0 for launch" milestone. For that reason, I'm adding a label to this PR and we'll take a look when more critical issues are resolved.

I didn't do a full review, but taking a quick look at your PR, it looks like your approach will work. With that said, I think we can do better.

I believe the fix should be as simple as moving the code that checks for available ports out of the `perform` function and into the `prepare` function. No modifications to `cli.py` should be necessary. You'll have to modify the tests a bit and I haven't tried this approach, but I believe this approach will get the desired behavior with less modifications and fits better with the rest of the code base.

bmw:
We've decided not to make this change so I'm closing this PR. For an explanation of why, see #1216.

Thanks for writing this up though and I'm sorry it sat around so long.

PR #1454, opened 2015-11-10
Check for nginx binary before running letsencrypt-nginx

pritambaral:
Pretty simple really.

PR #1477, opened 2015-11-12
Fix test broken by #1454 if nginx installation not present in system

joohoi:
le_util.exe_exists had to be mocked in test, which failed if nginx installation was not present in the installation after changes from #1454


pde:
LGTM.

PR #1478, opened 2015-11-12
Storage logging

TheNavigat:
Apparently, the PR is not complete. I'm creating it so that people can observe the work on this. Apparently, any help/comments/criticise is greatly appreciated.

This PR addresses issue https://github.com/letsencrypt/letsencrypt/issues/905


letsencrypt/storage.py
```diff
@@ -95,6 +100,7 @@ def __init__(self, config_filename, cli_config):
         """
         self.cli_config = cli_config
         if not config_filename.endswith(".conf"):
+            logger.debug("Renewal config file name must end in .conf")
```
TheNavigat: I'm not really sure what I should be including inside ```debug()```. Taking this line as an example, is what I'm doing here correct? Or should I add ```errors.CertStorageError``` as an argument instead?

TheNavigat:
I think I've finished adding logging to ```storage.py```, however, some of the messages are not informative enough/specific/clear. Would be nice if someone can take a look at this and approve/disapprove the messages.

letsencrypt/storage.py
```diff
@@ -275,11 +289,13 @@ def current_version(self, kind):
         pattern = re.compile(r"^{0}([0-9]+)\.pem$".format(kind))
         target = self.current_target(kind)
         if target is None or not os.path.exists(target):
+            logger.debug("File does not exist")
```
pde: You should make these log messages maximally meaningful.  Something like this:
`logger.debug("Current-version target for %s does not exist at %s", kind, target)`

pde:
Probably that comment is true for most of these log messages.  Think about what you would want to read in logs about such events, and put the necessary details (filenames, etc) in the log message.

TheNavigat:
@pde, Looks better, now?

letsencrypt/storage.py
```diff
@@ -617,15 +644,19 @@ def new_lineage(cls, lineagename, cert, privkey, chain,
             os.symlink(os.path.join(relative_archive, kind + "1.pem"),
                        target[kind])
         with open(target["cert"], "w") as f:
+            logger.debug("Writing certificate.")
```
pde: maybe add the paths (`target["cert"]`) to these messages?

pde:
That looks much better.  Probably we should have the paths for where the certs, chains etc are being written, but aside from that it looks ready to merge.

TheNavigat:
Ummm... I think boulder needs a kick...

bmw:
[log](https://gist.github.com/684c8859c343cc281c44) = letsencrypt/boulder#756

letsencrypt/storage.py
```diff
@@ -7,12 +7,17 @@
 import parsedatetime
 import pytz
 
+import logging
```
bmw: nit: This line should be moved between `import datetime` and `import os`.

The general pattern for this repo is:

```
standard library imports

dependency imports

LE code imports
```
Each of the three groups should be in alphabetical order. Sometimes our own packages are further split up if there is a lot of imports. See `letsencrypt/client.py` for an example of that.

letsencrypt/storage.py
```diff
@@ -7,12 +7,17 @@
...
+import logging.handlers
```
bmw: Unused import that should be deleted.

letsencrypt/storage.py
```diff
@@ -136,14 +141,17 @@ def _consistent(self):
 
         """
         # Each element must be referenced with an absolute path
-        if any(not os.path.isabs(x) for x in
-               (self.cert, self.privkey, self.chain, self.fullchain)):
-            return False
+        for x in (self.cert, self.privkey, self.chain, self.fullchain):
+            if not os.path.isabs(x):
+                logger.debug("Element %s is not referenced with an "
+                             "absolute file.", x)
```
bmw: nit: I think "Element %s is not referenced with an absolute **path**" is more accurate.

letsencrypt/storage.py
```diff
@@ -157,16 +165,24 @@ def _consistent(self):
                 self.cli_config.archive_dir, self.lineagename)
             if not os.path.samefile(os.path.dirname(target),
                                     desired_directory):
+                logger.debug("Element's link does not point within the "
+                             "cert lineage's directory within the "
+                             "official archive directory. Link: %s, "
+                             "archive directory: %s.",
+                             os.path.dirname(target), desired_directory)
```
bmw: I believe `os.path.dirname(target)` should be replaced with `link`?

letsencrypt/storage.py
```diff
@@ -157,16 +165,24 @@ def _consistent(self):
...
                 return False
 
             # The link must point to a file that exists
             if not os.path.exists(target):
+                logger.debug("Link %s points to a file that does not exist.", target)
```
bmw: `target` should be `link`

letsencrypt/storage.py
```diff
@@ -157,16 +165,24 @@ def _consistent(self):
...
                 return False
 
             # The link must point to a file that follows the archive
             # naming convention
             pattern = re.compile(r"^{0}([0-9]+)\.pem$".format(kind))
             if not pattern.match(os.path.basename(target)):
+                logger.debug("Link %s does not follow the archive naming "
+                             "convention.", os.path.basename(target))
```
bmw: This should probably be `logger.debug("%s does not follow the archive naming convention.", target)`.

letsencrypt/storage.py
```diff
@@ -251,6 +267,7 @@ def current_target(self, kind):
             raise errors.CertStorageError("unknown kind of item")
         link = getattr(self, kind)
         if not os.path.exists(link):
+            logger.debug("Target %s of kind %s does not exist.", link, kind)
```
bmw: Should be something like `logger.debug("Expected symlink %s does not exist", link)`.

letsencrypt/storage.py
```diff
@@ -537,6 +558,9 @@ def should_autorenew(self):
                 "cert", self.latest_common_version()))
             now = pytz.UTC.fromutc(datetime.datetime.utcnow())
             if expiry < add_time_interval(now, interval):
+                logger.debug("Should renew, certificate "
+                             "has expired since %s.",
```
bmw: Should be "has been expired since" or "certificate expired at".

letsencrypt/storage.py
```diff
@@ -588,6 +612,7 @@ def new_lineage(cls, lineagename, cert, privkey, chain,
                   cli_config.live_dir):
             if not os.path.exists(i):
                 os.makedirs(i, 0700)
+                logger.debug("Creating CLI config directory %s.", i)
```
bmw: The name of the variable `cli_config` is a bit misleading (it's the CLI command line arguments/parsed configuration file).

Probably should just be `"Creating directory %s"`

bmw:
I made a number of small suggestions but in general this looks good! Thanks for another PR @TheNavigat!

letsencrypt/storage.py
TheNavigat: What variable name do you think is more suitable? It would be nice to refactor the code a bit along with the logging.

bmw:
LGTM. Merging.

PR #1483, opened 2015-11-13
Absolute paths (fixes #836)

### LINKED ISSUE #836: CLI directory options should take absolute path

bmw:
After letting #1167 sit while I worked on more pressing tasks, I realized that I don't really like that code :/

Here's my 2nd try.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -931,26 +931,28 @@ def _paths_parser(helpful):
     if verb in ("install", "revoke", "certonly"):
         section = verb
     if verb == "certonly":
-        add(section, "--cert-path", default=flag_default("auth_cert_path"), help=cph)
+        add(section, "--cert-path", type=os.path.abspath,
+            default=flag_default("auth_cert_path"), help=cph)
     elif verb == "revoke":
         add(section, "--cert-path", type=read_file, required=True, help=cph)
     else:
-        add(section, "--cert-path", help=cph, required=(verb == "install"))
+        add(section, "--cert-path", type=os.path.abspath,
+            help=cph, required=(verb == "install"))
 
     section = "paths"
     if verb in ("install", "revoke"):
         section = verb
     # revoke --key-path reads a file, install --key-path takes a string
-    add(section, "--key-path", type=((verb == "revoke" and read_file) or str),
-        required=(verb == "install"),
+    add(section, "--key-path", required=(verb == "install"),
+        type=((verb == "revoke" and read_file) or os.path.abspath),
         help="Path to private key for cert creation or revocation (if account key is missing)")
 
     default_cp = None
     if verb == "certonly":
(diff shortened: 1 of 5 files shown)


pde:
LGTM.

letsencrypt/cli.py
```diff
@@ -569,7 +569,7 @@ def plugins_cmd(args, config, plugins):  # TODO: Use IDisplay rather than print
 def read_file(filename, mode="rb"):
     """Returns the given file's contents.
 
-    :param str filename: Filename
+    :param str filename: filename as an absolute path
```
kuba: input doesn't have to be absolute, @bmw

letsencrypt/configuration.py
```diff
@@ -37,6 +37,11 @@ class NamespaceConfig(object):
 
     def __init__(self, namespace):
         self.namespace = namespace
+
+        self.namespace.config_dir = os.path.abspath(self.namespace.config_dir)
```
kuba: i liked previous solution better :(

bmw: Really? Why's that?

Sure, the previous solution simplified some of the code in `cli.py`, but that was outside the scope of #836. I think the changes to `configuration.py` are a strict improvement on the old PR.

Previously, I added properties for `cert_path`, `key_path`, `chain_path`, and `fullchain_path`. These should not be properties of the configuration and they are correctly never used that way in the code.

With those changes removed, the only paths that need to be made absolute in `configuration.py` are `config_dir`, `work_dir`, and `logs_dir`. These values do not change during runtime so converting them to an absolute path on every access is unnecessary. Instead, just convert them to an absolute path upfront and all paths based on them will work without further changes.

PR #1486, opened 2015-11-13
Update _rpm_common.sh

malublu:
Add the required package which


pde:
Actually, we should use `command -v` instead, it's somewhat POSIX compliant...

PR #1487, opened 2015-11-13
Replace "which" with "command -v"

pde:
Since the latter is at least semi-POSIX compliant:

http://pubs.opengroup.org/onlinepubs/009696899/utilities/command.html

Avoids the need for #1486


bmw:
LGTM

Issue #1409, opened 2015-11-07
Simple-verify for all plugins. Log, do not fail.


kuba:
Let's use `logger.warning` instead of stopping the client when `self_verify` has failed. Also, let's do it for all plugins, not just manual authenticator. It might be opt-out, but I don't see a clear reason why it should be.

kuba:
@bmw, I believe "Log, do not fail" part of this issue should be bumped to 1.0 milestone. It causes a lot of problems for `manual` users.

bmw:
The main reason it's causing people problems though is the recently removed restrictions on Content-Type correct? Once that is fixed, I suspect this issue will have a much lower priority.

bmw:
I'm breaking this into two issues: #1585 and #1586.

PR #1488, opened 2015-11-13
Additional hint on cmd vs copy&paste

Xosofox:
I spent a lot of time figuring out why the challenge failed - my file was 86 bytes (pasting to vim) instead of 87 when created with the printf

So to make users aware of issue when pasting only the text, this might help others spending that time again

### DIFF
--- a/letsencrypt/plugins/manual.py
+++ b/letsencrypt/plugins/manual.py
@@ -47,6 +47,7 @@ Make sure your web server displays the following content at
 {validation}
 
 Content-Type header MUST be set to {ct}.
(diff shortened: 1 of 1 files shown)


bmw:
Even though it's not specified in the ACME spec, it appears that Boulder allows a new line at the end of the file and the problem is the client's self verification is too strict. Instead of this, we should fix the problem in the client.

I'm bumping the priority of #1409. With this fixed, the client will show a warning that the client failed to successfully verify your file, but will contact boulder who will verify the file successfully. I'm closing this PR, but please follow up if you think I should reconsider and solving #1409 isn't an adequate fix.

PR #1494, opened 2015-11-13
use boulder's integration-test.py

jmhodges:
This prevents the integration tests from getting run before the boulder processes have finished booting in most cases. There's still some small races with debug ports going up before RPC ports, but this flushes the big ones (specifically, the WFE ports), and the boulder devs are going to fix the rest in integration-test.py over time.

This also makes boulder-start.sh a blocking operation. Now that the TravisCI integration tests no longer requires boulder-start.sh, we can let the other priority of being easier for users to control (that is, basically,
make it easy to Ctrl-C) take over. That plus the idea that self-daemonizing code is tricky to get right, especially over multiple platforms led me to not trying to get start.py to make itself asynchronous.

Most of this change is code movement in order to allow developers to run boulder-start.sh once and boulder-integration.sh many times while also not duplicating that code in order to run the tests in TravisCI.

I'm not a huge fan of both the letsencrypt's shell scripts and boulder's integration-test.py having hard-coded file dependencies in the other's repo.

This, however, seemed like the smallest path to code that would spuriously break less. All the designs I was able to come up that were maybe smaller changes either had the "starts tests before the servers are up" problem or with a "each repo uses another repo's test code file" problem. Those problem on top of the "it's a bigger change" problem led me here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,8 +60,7 @@ addons:
     - rsyslog
 
 install: "travis_retry pip install tox coveralls"
-before_script: '[ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/boulder-start.sh'
-script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || (source .tox/$TOXENV/bin/activate && ./tests/boulder-integration.sh))'
+script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/travis-integration.sh)'
 
 after_success: '[ "$TOXENV" == "cover" ] && coveralls'
 
--- a/tests/boulder-start.sh
+++ b/tests/boulder-start.sh
@@ -1,43 +1,9 @@
 #!/bin/bash
-# Download and run Boulder instance for integration testing
-
-
-# ugh, go version output is like:
-# go version go1.4.2 linux/amd64
-GOVER=`go version | cut -d" " -f3 | cut -do -f2`
-
-# version comparison
-function verlte {
-  #OS X doesn't support version sorting; emulate with sed
-  if [ `uname` == 'Darwin' ]; then
-    [ "$1" = "`echo -e \"$1\n$2\" |  sed 's/\b\([0-9]\)\b/0\1/g' \
-      | sort | sed 's/\b0\([0-9]\)/\1/g' | head -n1`" ]
-  else
-    [  "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
-  fi
-}
-
-if ! verlte 1.5 "$GOVER" ; then
-  echo "We require go version 1.5 or later; you have... $GOVER"
-  exit 1
-fi
-
-set -xe
 
 export GOPATH="${GOPATH:-/tmp/go}"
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -60,8 +60,7 @@ addons:
     - rsyslog
 
 install: "travis_retry pip install tox coveralls"
-before_script: '[ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/boulder-start.sh'
-script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || (source .tox/$TOXENV/bin/activate && ./tests/boulder-integration.sh))'
+script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/travis-integration.sh)'
```
kuba: slight downside is that previously we bootstraped boulder while unit tests were running, so total build time might increase with this PR

tests/boulder-start.sh
```diff
@@ -1,43 +1,9 @@
...
-wget https://github.com/jsha/boulder-tools/raw/master/goose.gz && \
-  mkdir $GOPATH/bin && \
-  zcat goose.gz > $GOPATH/bin/goose && \
-  chmod +x $GOPATH/bin/goose
-./test/create_db.sh
-# listenbuddy is needed for ./start.py
-go get github.com/jsha/listenbuddy
-./start.py &
-# Hopefully start.py bootstraps before integration test is started...
+./start.py
```
kuba: missing `&` -> `constributing.rst` might need to be updated ("... open new terminal window...")

jmhodges: Ah, the docs actually say "You may now run (in a separate terminal):" before the boulder-integration.sh command example.

.travis.yml
jmhodges: They look pretty comparable to me! https://travis-ci.org/letsencrypt/letsencrypt/builds/91021463 (this branch) vs https://travis-ci.org/letsencrypt/letsencrypt/builds/90990967 (master)

jmhodges: (But I'd think it was worth it anyway)

tests/boulder-start.sh
kuba: ok, then that should be good, thanks for checking

.travis.yml
kuba: nice, thanks for checking

kuba:
I don't see any other blockers for the PR, but I haven't looked very thoroughly either. Travis is green, so I guess it's ok... :)

bmw:
@jmhodges, when I looked at the Travis build for this PR, I saw that there was boulder output from the integration tests (you can see the same thing looking at older PRs). I was curious, so I  kicked integration tests and now there is no output from boulder. Has something changed in the boulder repo in the last 5 days that would have caused this?

Regardless of this new problem, it clearly isn't being caused by this PR and this is an improvement on current integration tests. LGTM

jmhodges:
Yeah, that was on boulder. We accidentally committed a config change that quieted things. We just fixed it. The PR that fixed it is here https://github.com/letsencrypt/boulder/pull/1164

bmw:
Great. Thanks for looking into the issue and for writing this PR!

PR #1496, opened 2015-11-13
Update key_path help

bmw:


pde:
OK. It's a little odd that we still have the flag, dangling into nowhere.  But I think that implementing key-based revocation might require some boulder work...

PR #1508, opened 2015-11-14
Check for python2.7 before python2 in letsencrypt-auto

samlanning:
There are cases where more than one version of python2 is installed, and where
the default is not 2.7. For example in CentOS 6, it is common for both 2.6 and
2.7 to be installed, as yum requires 2.6 apparently for some reason.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -83,10 +83,10 @@ ExperimentalBootstrap() {
 }
 
 DeterminePythonVersion() {
-  if command -v python2 > /dev/null ; then
-    export LE_PYTHON=${LE_PYTHON:-python2}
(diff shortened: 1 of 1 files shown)


mitnk:
I like this PR. I encountered this issue once when I have `python2.7` but doesn't have `python2`.

samlanning:
Looks like there have been other people encountering this problem: https://community.letsencrypt.org/t/install-fails-on-amazon-linux/3708

kuba:
LGTM, but someone else will have to merge this :)

pde:
Tested on recent Ubuntus and Debians.  Merging.

pde:
(And one RHEL and one older Debian).

PR #1509, opened 2015-11-14
Add missing 'a' before 'safe'

mattbostock:
Add an 'a' before the word 'safe' so that the sentences about protecting `privkey.pem` makes sense.


pde:
Tx!

PR #1512, opened 2015-11-15
Include Arch Linux packages in User Guide

felixonmars:
The line is long :/ Let me know if it has to be shorter...

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -94,6 +94,13 @@ Operating System Packages
   * Port: ``cd /usr/ports/security/py-letsencrypt && make install clean``
   * Package: ``pkg install py27-letsencrypt``
 
+**Arch Linux**
+
+.. code-block:: shell
+
+   sudo pacman -S letsencrypt letsencrypt-nginx letsencrypt-apache \
+                  letshelp-letsencrypt
+
 **Other Operating Systems**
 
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -94,6 +94,10 @@ Operating System Packages
   * Port: ``cd /usr/ports/security/py-letsencrypt && make install clean``
   * Package: ``pkg install py27-letsencrypt``
 
+**Arch Linux**
+
+  * Packages: ``pacman -S letsencrypt letsencrypt-nginx letsencrypt-apache letshelp-letsencrypt``
```
kuba: 1. Can you use `.. code-block:: shell` as in other places of the documentation?
2. "Packages" should be omitted IMHO
3. Have you actually tested nginx and apache plugins? We officially support only Apache, and only on (some) Debian based distros; there are known problems in other distros...
4. Yup, lines should be limited to 79 chars, and 72 for long blocks of text. Use `\` to indicate wrapped line.

kuba:
Thank you very much for moving the packages to `community`! :)

felixonmars:
Updated accordingly (I was trying to keep it the same as FreeBSD instructions). I have tried the nginx plugin but no success yet. I'll probably do more tests and add some feedbacks.

pde:
Thanks!

PR #1513, opened 2015-11-15
Fallback if setuptools does not exist

DenisCarriere:


kuba:
Sorry, but `setuptools` is necessary to run the project, e.g. because of `pkg_resources`.

DenisCarriere:
Oh ok, sorry for that.
Thanks for the reply!

PR #1518, opened 2015-11-16
Fixes failed Redhat bootstrapping

Bob131:
Currently on Fedora systems, the bootstrap will fail with the error message `No package python27-virtualenv available`. This patch tests whether the system is an Amazon system before appending the problem package to the dependency list. Additionally, there is a commit which changes the yum/dnf testing order as dnf is the new standard, and if it's available it should be the preference over yum (which in Fedora 22 currently is just a warning that yum is deprecated and that you should be using dnf).

### DIFF
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -16,16 +16,20 @@ else
 fi
 
 # "git-core" seems to be an alias for "git" in CentOS 7 (yum search fails)
+DEPS="git-core
+    python
+    python-devel
+    python-virtualenv
+    gcc
+    dialog
+    augeas-libs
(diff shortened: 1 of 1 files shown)


Aeyoun:
I feel stupid for duplicating your efforts. I didn't see this pull request before suggesting my own solution. But I do believe my patch #1520 solves this better.  The return code in my patch is 0 when everything works or 1 when it fails. Functionality this patch does not preserve.

There may be future conflicts with naming and my patch sets up a loose structure for solving these conflicts. (Plus #1519. I stupidly made two requests.)

pde:
I've merged #1520.  Please let us know if that isn't working for everyone!

PR #1519, opened 2015-11-16
Test dnf before yum

Aeyoun:
yum may still be installed (by default in recent Fedoras) and will display a deprecation and migration message. On the other hand, dnf either is or isn't installed and the test will proceed as intended. (dnf is the modern replacement for yum.)

### DIFF
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -4,12 +4,13 @@
 #   - Fedora 22 (x64)
 #   - Centos 7 (x64: on AWS EC2 t2.micro, DigitalOcean droplet)
 
-if type yum 2>/dev/null
-then
-  tool=yum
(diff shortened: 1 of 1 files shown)


pde:
I think the original bug we had here applied only to systems without yum, because yum is happy if you ask it to install things that don't exist:

```
udo yum install jaberwocky git && echo success
Loaded plugins: amazon-id, rhui-lb
No package jaberwocky available.
Package git-1.8.3.1-4.el7.x86_64 already installed and latest version
Nothing to do
success
```

levskaya:
For this PR merged with master a7377fc48 , letsencrypt-auto works with CentOS7, RHEL7, Fedora23, and Amazon Linux 2015.09.1

pde:
Great, merging.

PR #1520, opened 2015-11-16
Support more distros with different RPM names

Aeyoun:
* Support both python and python27 naming conventions (varies by distro)
* Test every install command and give feedback if they don't succeed

The new install structure also allows for future conflicts/package naming inconsistencies to be resolved in a similar way. Should allow for wider RPM distro support with more testing and feedback.

### DIFF
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 # Tested with:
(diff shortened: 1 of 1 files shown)


pde:
Also tested on RHEL 7.1.

I think this is okay, though if we see reports of persistent failures we might need to `rm -rf` broken copies of `~/.local/share/letsencrypt` to recover.

PR #1524, opened 2015-11-17
Better UI when asking for email

pde:
This finishes off @miquelruiz's branch  #1124.

* fixes a bug or two
* the note about `--register-unsafely-without-email` is only displayed if the user tries to give a blank email
* serverside email rejection produces a retry, not erroring out
* fix broken test cases
* add new ones

Closes: #414

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -139,7 +139,7 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None and not args.allow_unsafe_registration:
+            if args.email is None and not args.register_unsafely_without_email:
                 args.email = display_ops.get_email()
             else:
                 args.email = None
@@ -843,7 +843,7 @@ def prepare_and_parse_args(plugins, args):
         None, "-t", "--text", dest="text_mode", action="store_true",
         help="Use the text output instead of the curses UI.")
     helpful.add(
-        None, "--allow-unsafe-registration", action="store_true",
+        None, "--register-unsafely-without-email", action="store_true",
         help="Specifying this flag enables registering an account with no "
              "email address. This is strongly discouraged, because in the "
              "event of key loss or account compromise you will irrevocably "
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -141,8 +141,6 @@ def _determine_account(args, config):
         else:  # no account registered yet
             if args.email is None and not args.register_unsafely_without_email:
                 args.email = display_ops.get_email()
-            else:
-                args.email = None
 
             def _tos_cb(regr):
                 if args.tos:
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -139,9 +139,9 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None:
+            if args.email is None and not args.allow_unsafe_registration:
                 args.email = display_ops.get_email()
-            if not args.email:  # get_email might return ""
+            else:
                 args.email = None
 
             def _tos_cb(regr):
@@ -842,6 +842,16 @@ def prepare_and_parse_args(plugins, args):
     helpful.add(
         None, "-t", "--text", dest="text_mode", action="store_true",
         help="Use the text output instead of the curses UI.")
+    helpful.add(
+        None, "--allow-unsafe-registration", action="store_true",
+        help="Specifying this flag enables registering an account with no "
+             "email address. This is strongly discouraged, because in the "
+             "event of key loss or account compromise you will irrevocably "
+             "lose access to your account. You will also be unable to receive "
+             "notice about impending expiration of revocation of your "
+             "certificates. Updates to the Subscriber Agreement will still "
+             "affect you, and will be effective N days after posting an "
+             "update to the web site.")
     helpful.add(None, "-m", "--email", help=config_help("email"))
     # positional arg shadows --domains, instead of appending, and
     # --domains is useful, because it can be stored in config
--- a/letsencrypt/client.py
+++ b/letsencrypt/client.py
@@ -140,7 +140,8 @@ def perform_registration(acme, config):
     try:
         return acme.register(messages.NewRegistration.from_data(email=config.email))
     except messages.Error, e:
-        if "MX record" in repr(e):
+        err = repr(e)
+        if "MX record" in err or "Validation of contact mailto" in err:
             config.namespace.email = display_ops.get_email(more=True, invalid=True)
             return perform_registration(acme, config)
         else:
--- a/letsencrypt/client.py
+++ b/letsencrypt/client.py
@@ -135,7 +135,13 @@ def perform_registration(acme, config):
     Actually register new account, trying repeatedly if there are email
     problems
 
-    :returns: the same value as acme.register
+    :param .IConfig config: Client configuration.
+    :param acme.client.Client client: ACME client object.
+
+    :returns: Registration Resource.
+    :rtype: `acme.messages.RegistrationResource`
+
+    :raises .UnexpectedUpdate:
     """
     try:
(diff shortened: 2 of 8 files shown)


letsencrypt/client.py
```diff
@@ -125,6 +130,21 @@ def register(config, account_storage, tos_cb=None):
 
     return acc, acme
 
+def perform_registration(acme, config):
+    """
+    Actually register new account, trying repeatedly if there are email
```
bmw: nit: I think more info should be added to this docstring. We could document the `acme` and `config` parameters as well as the return value. The return value may be the same as `acme.register`, but why not document `:returns:` and `:rtype:` here so developers don't have to go searching through our code.

tests/boulder-integration.sh
```diff
@@ -24,6 +24,7 @@ common() {
     letsencrypt_test \
         --authenticator standalone \
         --installer null \
+        --register-unsafely-without-email \
```
bmw: This should be moved to `tests/integration/_common.sh` and have it replace the `--email ""` hack.

letsencrypt/cli.py
```diff
@@ -139,9 +139,10 @@ def _determine_account(args, config):
         elif len(accounts) == 1:
             acc = accounts[0]
         else:  # no account registered yet
-            if args.email is None:
+            if args.email is None and not args.register_unsafely_without_email:
                 args.email = display_ops.get_email()
-            if not args.email:  # get_email might return ""
+            if not args.email:     # may be "" in integration tests?
```
bmw: With the change I suggested to the integration test scripts, this `if` statement and its body can be removed.

letsencrypt/client.py
```diff
@@ -125,6 +130,21 @@ def register(config, account_storage, tos_cb=None):
...
+    problems
+
+    :returns: the same value as acme.register
+    """
+    try:
+        return acme.register(messages.NewRegistration.from_data(email=config.email))
+    except messages.Error, e:
+        if "MX record" in repr(e):
+            config.namespace.email = display_ops.get_email(more=True, invalid=True)
+            return perform_registration(acme, config)
```
bmw: I don't really care if you fix this here and in `get_email` or not, but I think we should avoid recursion in Python since it doesn't support tail call elimination. See Guido's [blog](http://neopythonic.blogspot.com.au/2009/04/final-words-on-tail-calls.html) for more info on the issue.

letsencrypt/client.py
```diff
@@ -100,6 +100,11 @@ def register(config, account_storage, tos_cb=None):
     if account_storage.find_all():
         logger.info("There are already existing accounts for %s", config.server)
     if config.email is None:
+        if not config.register_unsafely_without_email:
```
bmw: There is no test coverage here.

bmw:
A more general question about this approach. As pointed out [here](https://github.com/letsencrypt/letsencrypt/issues/414#issuecomment-104711849), one unfortunate case of people registering without e-mail will be copy and pasted command lines. The name of the flag should help mitigate this, but this will likely still be a problem for some people.

Have you given any thought to this problem? We could make the user click through a warning prompt as suggested in the comment above. Alternatively, we could still prompt for an e-mail, but allow users to cancel to skip the prompt like they can currently. One downside to both approaches is we'd still need a way to automate through these prompts for integration tests.

I'm not sure we need to make a change here, but I was curious what your thoughts were on this issue.

letsencrypt/display/ops.py
```diff
@@ -114,23 +114,36 @@ def pick_configurator(
...
-    while True:
-        code, email = zope.component.getUtility(interfaces.IDisplay).input(
-            "Enter email address (used for urgent notices and lost key recovery)")
+    msg = "Enter email address (used for urgent notices and lost key recovery)"
+    if invalid:
+        msg = "There seem to be problems with that address. " + msg
+    if more:
+        msg += ('\n\nIf you really want to skip this, you can run the client with '
+                '--register-unsafely-without-email but make sure you backup your '
+                'account key from /etc/letsencrypt/accounts\n\n')
```
pde: Fixed by adding a height parameter to `NcursesDisplay.input`.

letsencrypt/client.py
pde: Done.

pde:
Agree the possibility of copying and pasting `--register-unsafely-without-email` is bad, but I can't see any solution to it; we do want that behaviour to be available via CLI flags for people who need it, I think.

pde:
All comments addressed.

pde:
@bmw If you +1 this I may clean up and rebase a bit before merging.

letsencrypt/display/ops.py
```diff
@@ -114,23 +114,39 @@ def pick_configurator(
...
+    height = 8
+    if invalid:
+        msg = "There seem to be problems with that address. " + msg
+        height += 1
+    if more:
+        msg += ('\n\nIf you really want to skip this, you can run the client with '
+                '--register-unsafely-without-email but make sure you backup your '
+                'account key from /etc/letsencrypt/accounts.')
+        height += 4
+    code, email = zope.component.getUtility(interfaces.IDisplay).input(msg, height=height)
```
bmw: This breaks the text interface. We should update both `IDisplay.input` in `letsencrypt/interfaces` and `FileDisplay.input` in `letsencrypt/display/ops.py` to take a height parameter. This is done for other `IDisplay` functions and I see no need to use `kwargs`.

bmw: Follow up: Don't forget to update docstrings! :smile:

letsencrypt/client.py
```diff
@@ -125,6 +130,28 @@ def register(config, account_storage, tos_cb=None):
...
+    :param acme.client.Client client: ACME client object.
+
+    :returns: Registration Resource.
+    :rtype: `acme.messages.RegistrationResource`
+
+    :raises .UnexpectedUpdate:
+    """
+    try:
+        return acme.register(messages.NewRegistration.from_data(email=config.email))
+    except messages.Error, e:
```
bmw: I'm sorry. I missed this before. None of the body of this `except` statement has test coverage.

pde: Are you sure?  I tried to implement some [here](https://github.com/letsencrypt/letsencrypt/pull/1524/files#diff-eedefe06400da5ee1616093685c1b14aR57).

pde:
Instead of [changing the iDisplay interface](https://github.com/letsencrypt/letsencrypt/pull/1524#discussion_r45146070) I decided instead to make the dialog plugin a bit smarter¹ about choosing heights.

¹ by "smarter" I mean it seems to handle this case correctly

letsencrypt/client.py
bmw: Yep. I'm sure. [Proof](https://travis-ci.org/letsencrypt/letsencrypt/jobs/92367746#L751).

I'll comment on the tests about the problem.

letsencrypt/display/ops.py
```diff
@@ -114,23 +114,36 @@ def pick_configurator(
         config, default, plugins, question,
         (interfaces.IAuthenticator, interfaces.IInstaller))
 
-
-def get_email():
+def get_email(more=False, invalid=False):
     """Prompt for valid email address.
 
+    :param bool more: -- explain why the email is strongly advisable, but how to
```
bmw: You should remove the `--`. A hyphen is added for every parameter when the docs are formatted so this looks odd.

letsencrypt/display/ops.py
```diff
@@ -114,23 +114,36 @@ def pick_configurator(
...
+        skip it
+    "param bool invalid: -- true if the user just typed something, but it wasn't
```
bmw: Ditto with above.

Also, typo: s/"/:

letsencrypt/client.py
bmw: With my commit, we should have test coverage on everything but line 155. I think that would also be nice to have, but I'm not as worried about it.

bmw:
Other than my comments about comments, this LGTM.

pde:
Fixing comment nits, rebasing away commits for avenues that were explored and then abandoned, and then merging.

Issue #1510, opened 2015-11-15
letsencrypt-auto: Install letsencrypt packages using pacman for Arch


felixonmars:
Hello,

I am packaging letsencrypt packages for Arch, and as a rolling release we are always following the latest letsencrypt releases. I understand that letsencrypt-auto is meant to be a script for installing into a virtualenv currently, but as more howtos and experience came out it's somewhat treated as the recommended setup now. So I would like to suggest to add the ability to use our system package manager to install letsencrypt packages (and even by default, if possible).

Thank you!

kuba:
So, I don't think that `letsencrypt-auto` should actually install packages from `pacman` (just like it shouldn't for any other distro). However, I would suggest to drop support for Arch in `letsencrypt-auto` and simply instruct users to use `pacman` instead. Ideally, we should get rid of `letsencrypt-auto` altogether...

felixonmars:
I see, I totally agree with you.

pde:
Feel free to send a PR that has `letsencrypt-auto` suggest Arch packages rather than running on Arch.  If there's any risk of Arch falling behind official client releases, it should still be possible to run `letsencrypt-auto` on Arch with `--debug`.  If Arch is expected to always track upstream releases very swiftly, we could also remove Arch support from `letsencrypt-auto` altogether.

PR #1526, opened 2015-11-17
letsencrypt-auto: Add instructions to use pacman on Arch

felixonmars:
According to #1510, let's recommend to use pacman by default (which is expected to always follow the latest releases of letsencrypt), and still allow for virtualenv installation via `--debug`.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -126,8 +126,17 @@ then
     echo "Bootstrapping dependencies for openSUSE-based OSes..."
     $SUDO $BOOTSTRAP/_suse_common.sh
   elif [ -f /etc/arch-release ] ; then
-    echo "Bootstrapping dependencies for Archlinux..."
-    $SUDO $BOOTSTRAP/archlinux.sh
+    if [ "$DEBUG" = 1 ] ; then
+      echo "Bootstrapping dependencies for Archlinux..."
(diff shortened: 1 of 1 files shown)


letsencrypt-auto
```diff
@@ -126,8 +126,17 @@ then
...
     $SUDO $BOOTSTRAP/_suse_common.sh
   elif [ -f /etc/arch-release ] ; then
-    echo "Bootstrapping dependencies for Archlinux..."
-    $SUDO $BOOTSTRAP/archlinux.sh
+    if [ "$DEBUG" = 1 ] ; then
+      echo "Bootstrapping dependencies for Archlinux..."
+      $SUDO $BOOTSTRAP/archlinux.sh
+    else
+      echo "Please use pacman to install letsencrypt packages:"
+      echo "# pacman -S letsencrypt letsencrypt-nginx letsencrypt-apache letshelp-letsencrypt"
```
bmw: This should be updated to only include `letsencrypt` and `letsencrypt-apache`.

bmw:
LGTM other than my comment above. With that said, I'd like @pde's approval before merging this.

felixonmars:
Updated.

pde:
Great, thanks!

PR #1527, opened 2015-11-17
Add missing RPM requirement

Aeyoun:
`redhat-rpm-config` provides the required `/usr/lib/rpm/redhat/redhat-hardened-cc1`.

At least available in Fedora and CentOS. No idea about other RPM distros in this case.

### DIFF
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -40,6 +40,7 @@ if ! $tool install -y \
        augeas-libs \
        openssl-devel \
        libffi-devel \
+       redhat-rpm-config \
        ca-certificates
 then
     echo "Could not install additional dependencies. Aborting bootstrap!"


kuba:
why is this necessary?

pde:
Are we sure that this dependency isn't missing on some older systems?  We don't want it to become a reason for calls to `dnf` to return error codes.  Perhaps we should systematically prevent that by iterating through the packages we're installing with `dnf`, and making a call to ensure that each is available before trying to install it.

(We don't have the same problem with yum, since it returns true in any case... and we don't really want to do ` || true` because we want to bail in case of "real" yum/dnf errors)

Aeyoun:
The package, `redhat-rpm-config`, is available as far back as Fedora 6 according to the [builds list](https://apps.fedoraproject.org/packages/redhat-rpm-config/). It's available in current CentOS too, but I don't know how to check older versions. Nor which other RPM distros are important to check.  I'd say it's a good guess to assume it will be available pretty much everywhere given how long it's been in Fedora. (I'm a recent RPM-based-distro convert so I may just be optimistic.)

We could just ignore the output if `$tool install -y $package`? That isn't really a solution. For the installer to give a consistently good experience it shouldn't let users carry on with missing dependencies as that will leave them with less informative errors than they'll get from their package managers.

pde:
The problem is that characterising exactly what the dependencies are on all the variants of Fedora, RHEL, Centos and Amazon Linux is non trivial.  I'm somewhat inclined to move in the direction of the approach taken by the [debian bootstrap script](https://github.com/letsencrypt/letsencrypt/blob/master/bootstrap/_deb_common.sh#L27).

pde:
@levskaya if you tell me this runs on all of your RPM-based AMIs (including an Amazon Linux one), we can merge.

levskaya:
@pde letsencrypt-auto in certonly standalone mode works fine on RHEL7, Fedora23, CentOS7, and Amazon Linux with this PR.

pde:
Great.  Merging!

PR #1534, opened 2015-11-17
Allow public IP logging

TheNavigat:


TheNavigat:
The test failure is really confusing for me...

I tested the script locally, and the exception is raised correctly if ```--manual-allow-public-ip-logging``` is not passed. However, the testing script reports that the script continue running even after disagreeing with logging the public IP. I can't seem to be able to understand why does that happen.

TheNavigat:
I added a couple of print statements before the ```if not self.conf("public-ip-logging-ok")``` condition, turns out that for some reason MagicMock sets the argument to some value.

```
Public IP Logging:
<MagicMock name='mock.manual_public_ip_logging_ok' id='4407338960'>
```

letsencrypt/plugins/manual.py
```diff
@@ -90,6 +90,8 @@ def __init__(self, *args, **kwargs):
     def add_parser_arguments(cls, add):
         add("test-mode", action="store_true",
             help="Test mode. Executes the manual command in subprocess.")
+        add("public-ip-logging-ok", action="store_true",
```
bmw: Small suggestion: This becomes the flag `--manual-public-ip-logging-ok`. This is very verbose. What about `--manual-ip-logging-ok` or `--manual-public-log-ok`?

bmw:
Other than my very small suggestion, this LGTM.

TheNavigat:
As the whole PR is a solution to @kuba's issue, that's where I got the name from.

I'm siding with the more verbose one. A long flag is better than a short but an ambiguous one.

bmw:
Very true. After thinking about it, I think the flag is fine.

LGTM. Thanks @TheNavigat.

PR #1536, opened 2015-11-17
Fix for TLDs greater than 6 characters in length

SuperMatt:
Following bug https://github.com/letsencrypt/letsencrypt/issues/1529 which does not recognise .digital and .graphics, which are both TLDs which I use, I have increased the TLD length to 12 to encapsulate most of the new TLDs.

### DIFF
--- a/letsencrypt/configuration.py
+++ b/letsencrypt/configuration.py
@@ -155,8 +155,8 @@ def _check_config_domain_sanity(domains):
             "Punycode domains are not supported")
     # FQDN checks from
     # http://www.mkyong.com/regular-expressions/domain-name-regular-expression-example/
-    #  Characters used, domain parts < 63 chars, tld > 1 < 13 chars
+    #  Characters used, domain parts < 63 chars, tld > 1 char
     #  first and last char is not "-"
-    fqdn = re.compile("^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,12}$")
+    fqdn = re.compile("^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,}$")
     if any(True for d in domains if not fqdn.match(d)):
         raise errors.ConfigurationError("Requested domain is not a FQDN")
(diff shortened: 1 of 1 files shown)


letsencrypt/configuration.py
```diff
@@ -157,6 +157,6 @@ def _check_config_domain_sanity(domains):
     # http://www.mkyong.com/regular-expressions/domain-name-regular-expression-example/
     #  Characters used, domain parts < 63 chars, tld > 1 < 7 chars
```
sjerdo: This comment should also be changed. (7 => 13)

SuperMatt:
I've updated my master branch with the comment changes, so this could now be added to the let's encrypt master.

mrak:
Might be work renaming this PR to describe the problem being solved for searchability.

Also, at this point, it may make more sense to just ensure that the TLD portion is at least 2 characters.

pde:
@SuperMatt do you have a methodology for selecting a length of 12?  I suspect @mrak may be right, a length limit in that section might not work.  There are some [utterly ridiculous](https://www.iana.org/domains/root/db/sandvikcoromant.html) TLDs being issued now :(

SuperMatt:
I've updated the PR so that it just has {2,} at the end, with no upper limit. I picked 12 because it was double the number that was there before, but after seeing that 15 character tld, I don't think we can begin to consider an upper limit.

rolandshoemaker:
The hard limit is the DNS label length, which is defined to have a maximum of 63 characters, to quote the [RFC](https://www.ietf.org/rfc/rfc1034.txt) 'Labels must be 63 characters or less.'

SuperMatt:
Ok, so the max tld length should be no higher than 63.

XN--VERMGENSBERATUNG-PWB is the current longest TLD at 24 characters long, according to https://data.iana.org/TLD/tlds-alpha-by-domain.txt. What is the world coming to?

robertsky:
XN--* has something to do with internationalised tlds.
i.e. xn--yfro4i67o is .新加坡 (which is the Chinese tld for Singapore).

pde:
LGTM.

buryni:
I can not renew my domain colindar.es (spanish domain)

> Requested domain  is not a FQDN

SuperMatt:
This would be a separate issue because this one only covers domains which were 6 characters or over, and the code has progressed a lot since this bug was raised that the patch I submitted has been completely overwritten.

bmw:
@buryni, Certbot doesn't seem to have a problem with that name (typed completely in ASCII). If you're still having the problem, please open another issue.

PR #1558, opened 2015-11-18
Check that the user is actually allowed to run sudo.

eldering:
This is no complete check that using sudo will work, but at least
catches the case where there is nothing configured for this user,
and allows graceful degradation to using su.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -33,7 +33,7 @@ done
 # this script *can* be run as root (not recommended), or fall back to using
 # `su`
 if test "`id -u`" -ne "0" ; then
-  if command -v sudo 1>/dev/null 2>&1; then
+  if command -v sudo 1>/dev/null 2>&1 && sudo -vn 1>/dev/null 2>&1 ; then
     SUDO=sudo
   else
     echo \"sudo\" is not available, will use \"su\" for installation steps...


pde:
Unfortunately, `-v` doesn't mean verify in an versions of sudo :(.

On debian 6.0 (squeeze), man sudo shows this:

```
       -v          If given the -v (validate) option, sudo will update the user's time stamp, prompting for the user's password if necessary.  This extends the
                   sudo timeout for another 15 minutes (or whatever the timeout is set to in sudoers) but does not run a command.
```

`sudo -vn` errors out on that system :(

pde:
I guess it's actually the same flag.  But the side-effect semantics of testing whether you can run sudo seems to be missing...

pde:
Interestingly, the reason it's missing is that there's a bug where `-n` is not accepted with `-v` on that system (despite the man page saying it's allowed).  `sudo -v` works on its own...

eldering:
Hmmm... that's annoying. Would it be acceptable to add an option to override the sudo detection and use su? Maybe I'm one of the relatively few, but I suppose there are more users out there who have sudo installed (by default) but not configured for use, or at least not blanket root use. Now the only way around that is modifying the source code.

pde:
Yes, it would be fine to allow the SUDO variable (probably it should be called something less collision-prone) to be set by the environment, in the same way that our path variables are.

Closing this PR, feel free to open another one implementing that!

PR #1563, opened 2015-11-18
Always use the specified GPG keyid for signing everything.

pde:
This line would follow git's idea what of the default PGP key was, which doesn't even match gpg's idea of the default.


bmw:
letsencrypt/boulder#756 caused [log](https://gist.github.com/836c1fc739ca0e7e5647).

PR #1577, opened 2015-11-20
#1576 Clearer error messages

### LINKED ISSUE #1576: "No installers" error message not clear
If running anything but (Currently) Apache on Debian, letsencrypt command will give one of the errors in cli.py:
```msg = ('No installers seem to be present and working on your system; '
                   'fix that or try running letsencrypt with the "certonly" command')```

```msg = ('No installers are available on your OS yet; try running '
                   '"letsencrypt-auto certonly" to get a cert you can install manually')```

I think the message is not clear as it is using the term "Installers" which might not be immediately apparent to be a cert installer. I'm not the only one who got confused by it, for example:
https://community.letsencrypt.org/t/no-installers-seem-to-be-present-and-working-on-your-system/1791

I suggest changing it to something like:
"No certificate installers were found on your system. Fix that or run letsencrypt with the "certonly" command to get a certificate and install in manually."

Also the second message says "letsencrypt-auto" which as far as I understand is the "pre release" name? Feel free to correct me on this and I'll update the Pull Request.

Sveder:
Clearer error messages.

Closes #1576

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -173,10 +173,11 @@ Renewal
 In order to renew certificates simply call the ``letsencrypt`` (or
 letsencrypt-auto_) again, and use the same values when prompted. You
 can automate it slightly by passing necessary flags on the CLI (see
-`--help all`), or even further using the :ref:`config-file`. If you're
-sure that UI doesn't prompt for any details you can add the command to
-``crontab`` (make it less than every 90 days to avoid problems, say
-every month).
+`--help all`), or even further using the :ref:`config-file`. The 
+``--renew-by-default`` flag may be helpful for automating renewal. If 
+you're sure that UI doesn't prompt for any details you can add the 
+command to ``crontab`` (make it less than every 90 days to avoid 
+problems, say every month).
 
 Please note that the CA will send notification emails to the address
 you provide if you do not renew certificates that are about to expire.


bmw:
letsencrypt/boulder#756 go away, come again another day (or don't). [Log](https://gist.github.com/17967148acf079922e3e).

Sveder:
I was wondering whether I should redo the PR to trigger a new build :)

pde:
Thanks @Sveder .  For the first message, I'm mulling the wording you have, "No certificate installers were found on your system", and an alternative like "No installers (plugins to install certificates in server software) were found on your system".

I think that alternative is a bit more detailed and helps to explain our internal concept of an "installer" to the user better, but it is also wordier which is a minus.  Which one do you think we should use?

Sveder:
That is an interesting question. I just wanted to make sure users will not confuse cert installers with other installers (as the word already means things, especially to people coming from the Windows world).

My 2 cents: I'd use my version and add a link in parenthesis to a web page that explains more about the issue which can than be kept up to date and made as deep and as hand holding as needed. I'm new to let's encrypt, so I'm not sure if you have some kind of wiki, which would be perfect for this. 

I'm not dogmatic on it though. Feel free to change to your version (or ask me to change), as it is better than what is there currently, and that is a great first step :)

bmw:
My own two cents; I've been torn for a while if we should be trying to explain to the average user what an `authenticator` and `installer` is in the context of Let's Encrypt. My hunch is that the average user doesn't need to know. We certainly shouldn't be trying to hide information from the user, but I think if we can avoid introducing new terminology, we should do so.

Sveder:
@bmw I agree that adding new terminology is confusing, but using the wrong terminology is way worse. Installers are already a term computer users use to mean a very specific kind of software. "Cert installers" might be installing something, but they are not installers. 

This is probably one of the most common error messages users will encounter, so this is indeed something to think and debate over. My personal position on error messages is that over explaining is good.

TheNavigat:
Let's throw in my two cents, I believe wordier is better than vague.

Issue #808, opened 2015-09-22
Apache plugin's cleanup() doesn't remove SSL vhosts created during auth


bmw:
If you run `letsencrypt auth` for an Apache virtual host that doesn't have SSL enabled, the Apache plugin creates a new vhost `*-le-ssl.conf`. After performing the challenges, the plugin's `cleanup` function is called but it doesn't remove the created vhosts. Instead, the user has to run `letsencrypt` again and the files will be deleted up when `recovery_routine` is called. These temporary files should be removed during the `cleanup` function.

jdkasten:
Note: The files aren't actually included in the final configuration.  The files only exist in sites-available.

pde:
Reportedly this is hairy because plugin has to handle two cases (`run` and `certonly`) and do different checkpointing in each case.  So removing `Good Volunteer Task`.

pde:
@bmw @SwartzCr closing this, right?

bmw:
Yes. This was correctly closed.

Issue #1328, opened 2015-11-04
LetsEncrypt's Apache plugin uses dummy placeholders in the generated apache conf 


clarkeandrew:
When using letsencrypt-auto, the generated apache vhost file is created correctly but the location of the certs is not specified in the file : 

i.e.

```
SSLCertificateFile insert_cert_file_path
SSLCertificateKeyFile insert_key_file_path
```

It looks like this is quite intentional in the code but im not sure why (https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/configurator.py#L579)

Furthermore just specifying the cert and key doesn't appear to be enough (I got browser security errors with this setup). Using the SSLCertificateChainFile directive fixes this : 

```
SLCertificateFile /etc/letsencrypt/live/domain.com/cert.pem
SSLCertificateKeyFile /etc/letsencrypt/live/domain.com/privkey.pem
SSLCertificateChainFile /etc/letsencrypt/live/domain.com/fullchain.pem
```

clarkeandrew:
I realise now that this may be the default behaviour for the 'auth' mode of letsencrypt. Is this the case?

Cactucs:
The source is here: https://github.com/letsencrypt/letsencrypt/blob/510da48263a4f19caa24495af6d318840eff9f29/letsencrypt-apache/letsencrypt_apache/configurator.py#L670-L673

pde:
I suspect #1582 will fix this; if anyone is motivated to check that, let us know!

romanseidl:
I have the same error on a current master or on 0.2.0. Hasn't that been fixed and merged into master?

mfairchild365:
I am having this issue too.

mfairchild365:
Ah, it looks like the problem was that `mod_ssl` was not enabled. I was able to install it with the command `sudo a2enmod ssl` and everything started to work as expected.

joohoi:
@mfairchild365 What distribution are you running? LE should actually check that mod_ssl is enabled when run on Ubuntu / Debian.

mfairchild365:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.4 LTS
Release:	14.04
Codename:	trusty
```
And for Apache: `Apache/2.4.7 (Ubuntu)`

joohoi:
I'll have to run some tests and try to reproduce the problem, but it certainly should be supported.

iarp:
[...]

PR #1582, opened 2015-11-20
prevent files from being left in apache sites-available issue 808

SwartzCr:
see issue #808 
by not writing a new file to get port specified IP we take the IP from the old config directly and append ssl ports to it to do the test. A file gets written when make_vhost_ssl gets called inside of choose_vhost which is entirely unnecessary.


bmw:
letsencrypt/boulder#756 rears its ugly head: [log](https://gist.github.com/e28b1cffb6b4471d0809)

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -230,7 +230,7 @@ def deploy_cert(self, domain, cert_path, key_path,
         if not vhost.enabled:
             self.enable_site(vhost)
 
-    def choose_vhost(self, target_name):
+    def choose_vhost(self, target_name, temp=False):
```
bmw: You should update the docstring below to say what `temp` does with respect to ssl vhosts and caching.

bmw:
Other than my comment about docstrings, this LGTM. I tested on both a VPS and using `letsencrypt-compatibility-test` and can confirm this fixes #808. Who knew it was so easy?!

pde:
I suspect this will also fix #1328...

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -230,13 +230,14 @@ def deploy_cert(self, domain, cert_path, key_path,
...
         """Chooses a virtual host based on the given domain name.
 
         If there is no clear virtual host to be selected, the user is prompted
         with all available choices.
 
         :param str target_name: domain name
+        :param bool temp: whether or not self.make_vhost_ssl shouldn't be called
```
bmw: I think this can be improved. It doesn't reference caching and to me it's not clear if `make_vhost_ssl` is called if `temp` is `True` or `False`. Perhaps we should be explicit and say what happens if the value is `True`?

SwartzCr: feel free to suggest a wordign

bmw:
After looking at this code again, I found another potential cause of #808 that this PR doesn't fix.

It occurs if `return self._choose_vhost_from_list(target_name)` is executed in the bottom of `choose_vhost`. If the selected vhost doesn't support TLS, a new vhost is created and cached. This leaves the file behind if the client was ran using `certonly`.

It's pretty easy to recreate this problem. Starting from the default Apache configuration, create another vhost in `sites-available` that is identical to the default vhost, but listens on a different port. After doing this, run the client using `certonly`.

To fix this problem, we should either just pass the `temp` parameter down into `_choose_vhost_from_list` or combine `_choose_vhost_from_list` into `choose_vhost` to avoid code duplication.

letsencrypt-apache/letsencrypt_apache/configurator.py
bmw: What about:

```
        """Chooses a virtual host based on the given domain name.

        If there is no clear virtual host to be selected, the user is
        prompted with all available choices.

        The returned vhost is guaranteed to have TLS enabled unless temp
        is True. If temp is True, there is no such guarantee and the
        is not cached.

        :param str target_name: domain name
        :param bool temp: whether the vhost is only used temporarily

        :rtype: :class:`~letsencrypt_apache.obj.VirtualHost`

        :raises .errors.PluginError: If no vhost is available or chosen

        """
```

SwartzCr:
@bmw I opted to pass temp to _choose_vhost_from_list
I also updated the docstring as per your recommendation

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -272,7 +279,8 @@ def _choose_vhost_from_list(self, target_name):
         elif not vhost.ssl:
             addrs = self._get_proposed_addrs(vhost, "443")
             # TODO: Conflicts is too conservative
-            if not any(vhost.enabled and vhost.conflicts(addrs) for vhost in self.vhosts):
+            if not any(vhost.enabled and vhost.conflicts(addrs) for vhost in self.vhosts)\
+                    and not temp:
```
bmw: So this actually doesn't quite work. The problem is if `temp` is `True`, the result is cached which will cause problems when an ssl vhost is required in the future (such as when using `run`). This is an easy fix though.

SwartzCr:
@bmw can you provide a little more info on this (so I can do it before you get in tomorrow)

letsencrypt-apache/letsencrypt_apache/configurator.py
bmw: To elaborate on this, the problem is in `_choose_vhost_from_list`. If `temp` is `True`, the function doesn't return early like it does in `choose_vhost`. As a result, the line `self.assoc[target_name] = vhost` is executed. This means that a potentially non-ssl vhost is cached, which will be a problem when a ssl vhost is required in the future.

The way I'd fix this is revert the changes to the `if not any (vhost.enabled ...` line and before `elif not vhost.ssl:` line add:
```
elif temp:
    return vhost
```

Hopefully this helps! If not, feel free to ping me here or on IRC.

bmw:
LGTM and tested successfully. Merging. Thanks @SwartzCr for doing what I deemed was "too hard".

Issue #1583, opened 2015-11-20
Missing space in help section - "domainsas" instead of "domains as"


Jonher937:
Line is missing space at the end of "domains"
Help message shows "domainsas" instead of "domains as"

`Domain names to apply. For multiple domains you can
 use multiple -d flags or enter a comma separated list
 of domainsas a parameter. (default: None)`

https://github.com/letsencrypt/letsencrypt/blob/457be44f1520d8904427a84a3568d891084369f4/letsencrypt/cli.py#L858

![image](https://cloud.githubusercontent.com/assets/1125285/11313517/b961284a-8fdd-11e5-8fa3-641cff3dc7f6.png)

Jonher937:
Merge done, closing this

PR #1584, opened 2015-11-20
Added space to avoid printing "domainsas" instead of the correct "domains as"

Jonher937:
Added space to avoid printing "domainsas" instead of the correct "domains as"

Fixes issue #1583


bmw:
LGTM. Thanks for noticing and fixing this @Jonher937.

PR #1590, opened 2015-11-21
Trim trailing whitespace during challenge self-verification

### LINKED ISSUE #1322: http-01 provisioned resources and newlines
For some reason the Litespeed Webserver for my shared hosting appears to add new-lines to text/plain files meaning that the acme challenge fails due to there being a "\n" at the end, it'd be great if LE accepted the \n or knew to remove the \n as I can't find a way to stop it having a \n in the HTTP response (I tried nano -L and echo).
It was returned as a text/plain file (ForceType text/plain in .htaccess) which I checked with HttpFox.
[Relevant section of log](http://pastebin.com/9g0MNYSS)

patf:
From ACME:
> Verify that the body of the response is well-formed key authorization. **The server SHOULD ignore whitespace characters at the end of the body.**

This was recently implemented in boulder here:
https://github.com/letsencrypt/boulder/pull/1142

**Note:** My implementation differs a bit from boulder in that I also include `\r`in the whitespace character list since `\r\n` is how Windows usually writes line breaks. Let me know if there's some reason why it shouldn't be included, otherwise I will create a PR for boulder to get this in sync.

fixes #1322

### DIFF
--- a/acme/acme/challenges_test.py
+++ b/acme/acme/challenges_test.py
@@ -127,6 +127,16 @@ class HTTP01ResponseTest(unittest.TestCase):
             self.chall, "local", KEY.public_key()))
 
     @mock.patch("acme.challenges.requests.get")
+    def test_simple_verify_whitespace_validation(self, mock_get):
+        from acme.challenges import HTTP01Response
+        mock_get.return_value = mock.MagicMock(
+            text=(self.chall.validation(KEY) +
+                  HTTP01Response.WHITESPACE_CUTSET), headers=self.good_headers)
+        self.assertTrue(self.response.simple_verify(
+            self.chall, "local", KEY.public_key()))
+        mock_get.assert_called_once_with(self.chall.uri("local"))
+
+    @mock.patch("acme.challenges.requests.get")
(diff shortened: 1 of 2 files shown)


acme/acme/challenges_test.py
```diff
@@ -127,6 +127,17 @@ def test_simple_verify_bad_validation(self, mock_get):
             self.chall, "local", KEY.public_key()))
 
     @mock.patch("acme.challenges.requests.get")
+    def test_simple_verify_whitespace_validation(self, mock_get):
+        from acme.challenges import HTTP01Response
+        validation = (self.chall.validation(KEY) +
+                      HTTP01Response.WHITESPACE_CUTSET)
+        mock_get.return_value = mock.MagicMock(
+            text=validation, headers=self.good_headers)
```
kuba: personally I would inline `validation`

kuba:
LGTM, thanks

Ref to the spec change: https://github.com/ietf-wg-acme/acme/pull/40/files

patf:
updated based on feedback from @kuba, hope this looks better.

pde:
@kuba feel free to merge this if you like the final rebased version.

PR #1591, opened 2015-11-21
Log, not fail (fixes #1585)

### LINKED ISSUE #1585: Log when simple-verify fails
Half #1409. When we fail to self verify a challenge, we should log the problem and continue instead of exiting.

bmw:


pde:
A PR that slightly simplifies our code :) LGTM!

Issue #1564, opened 2015-11-19
./letsencrypt-auto fails on update because of missing package python2.7-dev


dune73:
Problem:
```bash
$>cat /etc/issue
Ubuntu 14.04.3 LTS \n \l
$> git show
commit 6ea51360f9fcbd40e1c754bbc63a57224b73fb92
$> ./letsencrypt-auto certonly -a webroot --webroot-path /apache/htdocs/ -d example.com --server https://acme-v01.api.letsencrypt.org/directory --agree-dev-preview
Updating lUpdating letsencrypt and virtual environment dependencies......Command "/home/dune73/.local/share/letsencrypt/bin/python2 -c "import setuptools, tokenize;__file__='/tmp/pip-build-JtvkAg/cffi/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-yrfsWF-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/dune73/.local/share/letsencrypt/include/site/python2.7/cffi" failed with error code 1 in /tmp/pip-build-JtvkAg/cffietsencrypt and virtual environment dependencies......Command "/home/dune73/.local/share/letsencrypt/bin/python2 -c "import setuptools, tokenize;__file__='/tmp/pip-build-JtvkAg/cffi/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-yrfsWF-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/dune73/.local/share/letsencrypt/include/site/python2.7/cffi" failed with error code 1 in /tmp/pip-build-JtvkAg/cffi
```
The root-cause:
```bash
$> sudo pip install --upgrade cffi
...
    c/_cffi_backend.c:2:20: fatal error: Python.h: No such file or directory
     #include <Python.h>
                        ^
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

The Solution:

```bash
$> sudo apt-get install pythond2.7-dev
```

pde:
What OS was this on?

dune73:
Ubuntu 14.04.3 LTS \n \l

bmw:
So I just tested this on a fresh install of Ubuntu 14.04 and did not encounter the problem.

[...]

PR #1592, opened 2015-11-21
Be doubly sure to install python2.7-dev

pde:
Probably solves #1564, whether or not it's a real issue!

### DIFF
--- a/bootstrap/_deb_common.sh
+++ b/bootstrap/_deb_common.sh
@@ -32,10 +32,19 @@ if apt-cache show python-virtualenv > /dev/null ; then
   virtualenv="$virtualenv python-virtualenv"
 fi
 
+# This is likely to be unecessary in almost all cases,
+# but if python-dev actually depends some non-2.7 version,
+# but python2.7 is installed, we'll want this dev package
+# https://github.com/letsencrypt/letsencrypt/issues/1564
+pydev=python-dev
+if apt-cache show python2.7-dev > /dev/null ; then
+  pydev="$pydev python2.7-dev"
+fi
+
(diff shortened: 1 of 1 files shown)


bmw:
So I'm not convinced this shouldn't go in, but I'm not convinced it should either. The downsides I see with this are:

1. Increases code complexity with little/no gain. We cannot reproduce the problem and we've had only one report of the issue with 35k+ PyPI downloads in the past month.
2. Shuns Python 2.6 support. We've put a non-trivial amount of effort into keeping this project working on Python 2.6. Should we prefer Python 2.7 so strongly as to explicitly install Python 2.7 packages? I think the answer is no.
3. Duplicates the role of a package manager. We need `python-dev`. The user's package manager should be able to determine whether `python2.6-dev` or `python2.7-dev` should be installed. We should avoid building a mini-package manager into Let's Encrypt.

Thoughts?

pde:
Those seem like reasonable concerns.  Re: forcing py27, we do [agressively prefer py27](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-auto#L85) if it's present; it would probably be "most correct" to only try this if we are running with python2.7 or python27 as our executable.  But that's taking us even further in the direction of complexity.

bmw:
We can (and probably should) prefer Python 2.7, but we shouldn't force people to use it.

Personally, I'd like to close this PR and not merge it in for now. If in the future, we continue to have reports about missing `python2.7-dev`, we can either reopen or add the installation of `python2.7-dev` to `letsencrypt-auto` which knows the version of Python you're running.

pde:
SGTM.

PR #1593, opened 2015-11-21
Adding Python 2.6+ note to docs

TheNavigat:
Partially solves https://github.com/letsencrypt/letsencrypt/issues/1589

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -365,10 +365,12 @@ are provided here mainly for the :ref:`developers <hacking>` reference.
 In general:
 
 * ``sudo`` is required as a suggested way of running privileged process
+* `Python`_ 2.6/2.7 is required
 * `Augeas`_ is required for the Python bindings
 * ``virtualenv`` and ``pip`` are used for managing other python library
   dependencies
 
+.. _Python: https://wiki.python.org/moin/BeginnersGuide/Download
 .. _Augeas: http://augeas.net/
 .. _Virtualenv: https://virtualenv.pypa.io
 


docs/using.rst
```diff
@@ -42,6 +42,11 @@ To install and run the client you just need to type:
 
    ./letsencrypt-auto
 
+.. note:: Let's Encrypt requires Python 2.6+. Check Python_'s website
```
bmw: This actually isn't quite accurate. Currently, we only support  Python 2.6 and Python 2.7.

The `acme` module supports Python 2.6+, including Python 3+, but Python 3 is not currently supported by the rest of the client,.

TheNavigat:
@bmw Updated. Thanks for noting that!

TheNavigat:
I guess Travis needs a kick...

bmw:
letsencrypt/boulder#756 - [log](https://gist.github.com/b18e05e13b1fc30b155b)

bmw:
So I've been thinking about this a bit lately. The main things that have been bugging me are:

1) In the `letsencrypt-auto` section, it points users towards instructions helping them install Python. `letsencrypt-auto` does this for you so this isn't necessary.
2) Even more generally, I don't think the Python requirements should be in the `letsencrypt-auto` section as these restrictions are true regardless of what installation method you are using.

With this in mind, I'm leaning towards listing them in the README under the "About the Let's Encrypt Client" which will be shown both on Github and [here](https://letsencrypt.readthedocs.org/en/latest/intro.html#about-the-let-s-encrypt-client).

I'm not sold on this approach though. Do you have any thoughts @pde?

TheNavigat:
What about moving the note to just below the "Installation" header?

TheNavigat:
https://github.com/letsencrypt/letsencrypt/pull/1629, created to add Python 2.6 and 2.7 note to README.rst

TheNavigat:
Ping @bmw

bmw:
Sorry this PR has been sitting for so long @TheNavigat.

What about just listing the Python requirements [here](https://letsencrypt.readthedocs.org/en/latest/contributing.html#notes-on-os-dependencies)? The thought is that all installation methods other than "From source" handle the dependencies for you. "From source" directs you to the developer guide and the section I linked is where we have been putting OS dependencies.

TheNavigat:
This is a more sound idea.

I'll do that when I get the chance.

bmw:
Great! Thanks @TheNavigat.

I'm closing this PR to try and clean up our list of open PRs. Just yell at me if you'd like me to reconsider.

TheNavigat:
I was going to update the original commit because the PR itself is still relevant but the commit isn't perfect. I'd rather update this one than create a new one :/

TheNavigat:
@bmw Better?

TheNavigat:
Ping @bmw

bmw:
LGTM. Sorry for the delay in reviewing this @TheNavigat.

Issue #1393, opened 2015-11-06
mutliple-SAN UI for webroot


pde:
Once #1016 is done, it would need interactive UI if it is going to work for the maximum number of users.  It should call out to IDisplay once for each domain name, to figure out what the webroot is.  Ideally if the domains are `example.com` and `www.example.com` the text box for the second of them can be pre-populated with the preceding value?

bmw:
I'm not sold on anything, but I had a few ideas for this.

In the long run I think we want to use #445 for this so we can help users input their paths correctly. We certainly don't have to block this on #445, but I think #445 provides a much nicer UI/UX.

Also, I was thinking about the following interface. For each domain, we do the following:
1. We populate a list of all webroots input so far. The first option is to input a new webroot. If no webroots have been entered, we skip this step (which I believe is automatically done by the display code if there is only one option).
2. If the user requested to enter a new webroot, we prompt for the user input one.

I think this approach minimizes typing and provides a pretty good workflow. Unfortunately, I think the pathological case is likely to occur for someone. In this case, no two domains share the same webroot. The menu is useless and provides unnecessary work for the user. Luckily, this work is only repeatedly choosing the first option, but it certainly isn't ideal.

bmw:
This should remain a high priority, but I don't believe it is reasonable to expect this to be done by Thursday.

pde:
Flagging this as urgent because for many people, webroot is a better answer than standalone.

pde:
This was #2758

PR #1594, opened 2015-11-21
Support multiple webroot paths

pde:
Closes: #1016.  Themost minimal way to use it is to interleave --webroot-path / -w flags with domain flags:

`letsencrypt --webroot -w /var/www/thing -d thing.com -d www.thing.com -w /var/www/stuff -d stuff.io -d www.stuff.io`

But the user can also pass in a map of domains to webroots:

`letsencrypt --webroot --webroot-map '{ "thing.com" : "/var/www/thing", "www.thing.com" : "/var/www/thing", "stuff.io" : "/var/www/stuff", "www.stuff.io" : "/var/www/stuff" }' -d thing.com -d stuff.io`

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -3,10 +3,12 @@
 import argparse
 import atexit
 import functools
+import json
 import logging
 import logging.handlers
 import os
 import pkg_resources
+import string
 import sys
 import time
 import traceback
@@ -670,7 +672,6 @@ class HelpfulArgumentParser(object):
             print usage
             sys.exit(0)
         self.visible_topics = self.determine_help_topics(self.help_arg)
-        #print self.visible_topics
         self.groups = {}  # elements are added by .add_group()
 
     def parse_args(self):
@@ -1023,19 +1024,21 @@ def _plugins_parsing(helpful, plugins):
     helpful.add("plugins", "--webroot", action="store_true",
                 help='Obtain certs by placing files in a webroot directory.')
 
-    # This would normally be a flag within the webroot plugin,  the webroot
-    # plugin, but because it is parsed in conjunction with --domains, it lives
-    # here
-    helpful.add("webroot", "-w", "--webroot-path", action=WebrootPathProcessor,
-                help="public_html / webroot path")
-    #helpful.add("plugins", "-w", action=WebrootAction,
-    #            help='Obtain certs by placing files in a webroot directory.')
-
     # things should not be reorder past/pre this comment:
     # plugins_group should be displayed in --help before plugin
     # specific groups (so that plugins_group.description makes sense)
 
     helpful.add_plugin_args(plugins)
+
+    # These would normally be a flag within the webroot plugin, but because
+    # they are parsed in conjunction with --domains, they live here for
+    # legibiility. helpful.add_plugin_ags must be called first to add the
+    # "webroot" topic
+    helpful.add("webroot", "-w", "--webroot-path", action=WebrootPathProcessor,
+                help="public_html / webroot path")
+    parse_dict = lambda s : dict(json.loads(s))
+    helpful.add("webroot", "--webroot-map", default={}, type=parse_dict,
+                help="Mapping from domains to webroot paths")
 
 
 class WebrootPathProcessor(argparse.Action): # pylint: disable=missing-docstring
@@ -1062,15 +1065,15 @@ class DomainFlagProcessor(argparse.Action): # pylint: disable=missing-docstring
         Process a new -d flag, helping the webroot plugin construct a map of
         {domain : webrootpath} if -w / --webroot-path is in use
         """
-        if not config.domains: config.domains = []
-        new_domains = [d.strip() for d in domain_arg.split(",")
-                                  if d not in config.domains]
-        config.domains.extend(new_domains)
+        if not config.domains:
+            config.domains = []
 
-        if config.webroot_path:
-            # Each domain has a webroot_path of the most recent -w flag
-            for d in new_domains:
-                config.webroot_map[d] = config.webroot_path[-1]
+        for d in map(string.strip, domain_arg.split(",")):
+            if d not in config.domains:
+                config.domains.append(d)
+                # Each domain has a webroot_path of the most recent -w flag
+                if config.webroot_path:
+                    config.webroot_map[d] = config.webroot_path[-1]
 
 
 def setup_log_file_handler(args, logfile, fmt):
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1,5 +1,7 @@
 """Let's Encrypt CLI."""
 # TODO: Sanity check all input.  Be sure to avoid shell code etc...
+# pylint: disable=too-many-lines
+# (TODO: split this file into main.py and cli.py)
 import argparse
 import atexit
 import functools
@@ -384,7 +386,6 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
 def choose_configurator_plugins(args, config, plugins, verb):  # pylint: disable=too-many-branches
     """
     Figure out which configurator we're going to use
-
     :raises error.PluginSelectionError if there was a problem
     """
 
@@ -802,7 +803,6 @@ class HelpfulArgumentParser(object):
             return dict([(t, t == chosen_topic) for t in self.help_topics])
 
 
-
 def prepare_and_parse_args(plugins, args):
     """Returns parsed command line arguments.
 
@@ -946,8 +946,7 @@ def _create_subparsers(helpful):
         help="Set a custom user agent string for the client. User agent strings allow "
              "the CA to collect high level statistics about success rates by OS and "
              "plugin. If you wish to hide your server OS version from the Let's "
-             'Encrypt server, set this to "".'
-    )
+             'Encrypt server, set this to "".')
     helpful.add("certonly",
                 "--csr", type=read_file,
                 help="Path to a Certificate Signing Request (CSR) in DER"
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1062,8 +1062,7 @@ class DomainFlagProcessor(argparse.Action): # pylint: disable=missing-docstring
         Process a new -d flag, helping the webroot plugin construct a map of
         {domain : webrootpath} if -w / --webroot-path is in use
         """
-        if not config.domains:
-            config.domains = []
+        if not config.domains: config.domains = []
         new_domains = [d.strip() for d in domain_arg.split(",")
                                   if d not in config.domains]
         config.domains.extend(new_domains)
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1043,6 +1043,10 @@ def _plugins_parsing(helpful, plugins):
 
 
 class WebrootPathProcessor(argparse.Action): # pylint: disable=missing-docstring
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


letsencrypt/cli.py
```diff
@@ -1051,6 +1030,51 @@ def _plugins_parsing(helpful, plugins):
 
     helpful.add_plugin_args(plugins)
 
+    # These would normally be a flag within the webroot plugin, but because
+    # they are parsed in conjunction with --domains, they live here for
+    # legibiility. helpful.add_plugin_ags must be called first to add the
+    # "webroot" topic
+    helpful.add("webroot", "-w", "--webroot-path", action=WebrootPathProcessor,
```
bmw: I disagree with moving these flags outside of `webroot.py`. We have a very clear infrastructure for adding plugin specific flags. The plugin defines `add_parser_arguments` and all arguments needed by the plugin are defined there.

This completely disregards that infrastructure and creates an unnecessary exception to the pattern for `webroot`.

letsencrypt/cli.py
```diff
@@ -1051,6 +1030,51 @@ def _plugins_parsing(helpful, plugins):
...
+                help="public_html / webroot path")
+    parse_dict = lambda s: dict(json.loads(s))
+    helpful.add("webroot", "--webroot-map", default={}, type=parse_dict,
```
bmw: I feel that having the user write JSON on the command line is awkward, especially when they still have to specify all of their domains again with `-d/--domains`. To me, this feels like an unnecessarily exposed implementation detail that we should hide from the client.

pde: I started working on some code to make `-d` unecessary when `--webroot-map` is present, but the tests turned out to be tricky, so I'm not going to spend more time on it for now.  Will hide --webroot-map instead.

letsencrypt/cli.py
```diff
@@ -1051,6 +1030,51 @@ def _plugins_parsing(helpful, plugins):
...
+            # apply that webroot path to those; subsequent entries in
+            # config.webroot_map are filled in by cli.DomainFlagProcessor
+            if config.domains:
+                config.webroot_map = dict([(d, webroot) for d in config.domains])
+            else:
+                config.webroot_map = {}
+        config.webroot_path.append(webroot)
+
+
+class DomainFlagProcessor(argparse.Action): # pylint: disable=missing-docstring
```
bmw: As I said briefly in IRC, I don't think that the argparsing for domains should do anything with `webroot-path`. This also goes against our architecture of having the plugins do their argument parsing.

I think that changes to domain parsing can be reverted and logic similar to below can be used in `webroot.py` to achieve the same effect. There are a couple ways to do this. Basically, you need to keep track of the previous `webroot-path` flag. Then when you encounter a new `webroot-path` flag, you know you will use that webroot for all new domains found in `config.domains`. The very last `webroot-path` flag will require special treatment and will have to be applied outside of argument parsing, but I believe this can be done nicely using a function to avoid duplicated code.

This relies on `argparse` parsing arguments in the order they appear on the command line, but so does the current code in this PR.

bmw:
So I only looked at this PR on a high level but, I think we can do a better job of having the code fit with our current architecture instead of making a special case out of `--webroot-path` and `--webroot-map`.

pde:
Agree we could probably reorganise things in that way.  It's pretty awkward, though, because the last event that occurs is typically addition of domains with `-d`, and none of webroot's argparse machinery will be called after the last stanza of domains is added.

Webroot could call `WebrootPathProcessor` one last time during its `prepare()` call, but I think the fact that config is changing after construction is side-effectual and dangerous.  This is exacerbated by the fact that `prepare()` is called via a number of different paths.

pde:
It also strikes me that webroot's `prepare()` semantics is going to get an overhaul if we do #1393 for launch.  So my strong preference would be to land this without spending a lot of time chasing `prepare()` side effects in test cases, and open a refactor ticket for making webroot parse without any help from outside the normal plugins API.

pde:
Revised to not accept -d first if there are multiple instances of -w, since in our hallway testing that seemed to be a confusing case to potential users.

letsencrypt/cli.py
```diff
@@ -382,7 +386,6 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
 def choose_configurator_plugins(args, config, plugins, verb):  # pylint: disable=too-many-branches
     """
     Figure out which configurator we're going to use
-
```
bmw: Deleting this line causes documentation formatting problems.

letsencrypt/cli.py
```diff
@@ -1070,6 +1048,63 @@ def _plugins_parsing(helpful, plugins):
 
     helpful.add_plugin_args(plugins)
 
+    # These would normally be a flag within the webroot plugin, but because
+    # they are parsed in conjunction with --domains, they live here for
+    # legibiility. helpful.add_plugin_ags must be called first to add the
+    # "webroot" topic
+    helpful.add("webroot", "-w", "--webroot-path", action=WebrootPathProcessor,
+                help="public_html / webroot path. This can be specified multiple times to "
```
bmw: Question: Should we also state the behavior when there's only one webroot?

letsencrypt/cli.py
```diff
@@ -1070,6 +1048,63 @@ def _plugins_parsing(helpful, plugins):
...
+    def __init__(self, *args, **kwargs):
+        self.domain_before_webroot = False
+        argparse.Action.__init__(self, *args, **kwargs)
+
+    def __call__(self, parser, config, webroot, option_string=None):
+        """
+        Keep a record of --webroot-path / -w flags during processing, so that
+        we know which apply to which -d flags
+        """
+        if not config.webroot_path:
```
bmw: If `--webroot-map` is already set, the body of the `if` statement overwrites it.

letsencrypt/cli.py
```diff
@@ -1070,6 +1048,63 @@ def _plugins_parsing(helpful, plugins):
...
+class DomainFlagProcessor(argparse.Action): # pylint: disable=missing-docstring
+    def __call__(self, parser, config, domain_arg, option_string=None):
+        """
+        Process a new -d flag, helping the webroot plugin construct a map of
+        {domain : webrootpath} if -w / --webroot-path is in use
+        """
+        if not config.domains:
+            config.domains = []
+
+        for d in map(string.strip, domain_arg.split(",")): # pylint: disable=bad-builtin
```
bmw: nit: Changing the line to this:
```
for d in (domain.strip() for domain in domain_arg.split(",")):
```
allows you to remove `import string` and delete the comment about "bad-builtin".

letsencrypt/cli.py
bmw: nit: This line in particular should be `if config.webroot_path is None:`

letsencrypt/cli.py
```diff
@@ -1070,6 +1048,63 @@ def _plugins_parsing(helpful, plugins):
...
+        config.webroot_path.append(webroot)
+
+
+class DomainFlagProcessor(argparse.Action): # pylint: disable=missing-docstring
+    def __call__(self, parser, config, domain_arg, option_string=None):
+        """
+        Process a new -d flag, helping the webroot plugin construct a map of
+        {domain : webrootpath} if -w / --webroot-path is in use
+        """
+        if not config.domains:
```
bmw: `if config.domains is None:`

letsencrypt/plugins/webroot.py
```diff
@@ -41,34 +44,39 @@ def get_chall_pref(self, domain):  # pragma: no cover
...
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
-        assert self.full_root is not None
+        assert self.full_roots, "Webroot plugin appears to be missing webroot map"
         return [self._perform_single(achall) for achall in achalls]
 
     def _path_for_achall(self, achall):
-        return os.path.join(self.full_root, achall.chall.encode("token"))
+        path = self.full_roots[achall.domain]
+        if not path:
```
bmw: What is this `if` statement supposed to be checking?

letsencrypt/plugins/webroot_test.py
```diff
@@ -31,7 +31,8 @@ def setUp(self):
         self.validation_path = os.path.join(
             self.path, ".well-known", "acme-challenge",
             "ZXZhR3hmQURzNnBTUmIyTEF2OUlaZjE3RHQzanV4R0orUEN0OTJ3citvQQ")
-        self.config = mock.MagicMock(webroot_path=self.path)
+        self.config = mock.MagicMock(webroot_path=self.path,
```
bmw: This should be `...webroot_path=[self.path],`

letsencrypt/tests/cli_test.py
```diff
@@ -261,10 +263,10 @@ def test_plugins_init(self, mock_disco):
         stdout.write.called_once_with(str(verified))
 
     @mock.patch('letsencrypt.cli.plugins_disco')
-    def test_plugins_prepare(self, mock_disco):
+    @mock.patch('letsencrypt.cli.HelpfulArgumentParser.determine_help_topics')
+    def test_plugins_prepare(self, _det, mock_disco):
         ifaces = []
         plugins = mock_disco.PluginsRegistry.find_all()
-
```
bmw: Why?

bmw:
Writing argument parsers is obnoxious:
```
(venv)root@letsencrypt:~/letsencrypt# git checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.
(venv)root@letsencrypt:~/letsencrypt# letsencrypt -t -webroot
usage: 

letsencrypt: error: unrecognized arguments: -webroot
(venv)root@letsencrypt:~/letsencrypt# git checkout webroot
Switched to branch 'webroot'
Your branch is up-to-date with 'origin/webroot'.
(venv)root@letsencrypt:~/letsencrypt# letsencrypt -t -webroot

This is a PREVIEW RELEASE of a client application for the Let's Encrypt
certificate authority and other services using the ACME protocol. The Let's
Encrypt certificate authority is NOT YET ISSUING CERTIFICATES TO THE PUBLIC.

Until publicly-trusted certificates can be issued by Let's Encrypt, this
software CANNOT OBTAIN A PUBLICLY-TRUSTED CERTIFICATE FOR YOUR WEB SERVER.  You
should only use this program if you are a developer interested in experimenting
with the ACME protocol or in helping to improve this software.  If you want to
configure your web site with HTTPS in the meantime, please obtain a certificate
from a different authority.

For updates on the status of Let's Encrypt, please visit the Let's Encrypt home
page at https://letsencrypt.org/.
(A)gree/(C)ancel: C
Must agree to TOS
(venv)root@letsencrypt:~/letsencrypt#
```

bmw:
Just to reiterate my concerns publicly, I still don't like the implementation of this and I think we can do better.

With that said, I'm fine merging this PR for now once my questions/comments have been addressed.

letsencrypt/cli.py
pde: I considered that version, but found it longer and uglier!

bmw: It's certainly more verbose, but that's not necessarily a bad thing!

bmw:
Merging. In playing with this more I found some new subtle bugs with `--webroot-map`. With that said, as long as we don't tell the world about the hidden flag in our open source project, I think it'll be okay :smile: Problems were:

1. If `--webroot-map` is specified after some `-w`/`-d` flags, the currently parsed map is overwritten.

2. If the domains specified in `--webroot-map` are given to `-d` after `-w` has been used, their webroot is overwritten. Is this the behavior we want?

Issue #1389, opened 2015-11-06
Webroot plugin should override permisions of files / dirs created


Zash:
Running with --webroot-path and umask 0027 creates .well-known/acme-challenge/... with no global read permissions, causing the verification to fail because the HTTP server can't read the files (at least not in my setup).

Either overriding permissions during/after creation or making the directories and challenge file have similar permissions as the webroot itself should probably work in most cases.

lbeltrame:
Indeed. Where in the code is this handled? I may take a look at this.

bmw:
@lbeltrame, the code for this is in `letsencrypt/plugins/webroot.py` in the `prepare` function.

pde:
@lbeltrame if you're getting started on this now, I suggest working on top of the [webroot](https://github.com/letsencrypt/letsencrypt/tree/webroot) branch being discussed over in #1594, since it somewhat alters that `prepare` function.

lbeltrame:
I started doing some work https://github.com/lbeltrame/letsencrypt/tree/webroot, however GH is not allowing me to do a PR yet...

lbeltrame:
As you can see, a WIP PR landed. Please let me know how I can improve it.

pde:
Thanks for fixing this @lbeltrame !

yaccin:
This issue seems to have reappeared in certbot 0.14.0

PR #1596, opened 2015-11-22
[WIP] Fix webroot permissions

lbeltrame:
It marked this as WIP as I'm not sure I fixed everything I might have to fix.  This should address  issue #1389.

I added unit tests, but I can't run them at the moment, unfortunately.

PR #1597, opened 2015-11-22
Fix webroot permissions [mergeable, but enhancements also requested]

lbeltrame:
Take them from the parent directory where the webroot is. Should fix issue #1389.

This is a revised PR from #1596, with a cleaner history.

NOTE: I've been unable to run the unit tests for now.

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -100,8 +100,15 @@ to serve all files under specified web root ({0})."""
             try:
                 os.makedirs(self.full_roots[name])
                 # Set permissions as parent directory (GH #1389)
-                filemode = stat.S_IMODE(os.stat(path).st_mode)
+                # We don't use the parameters in makedirs because it
+                # may not always work
+                # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
+                stat_path = os.stat(path)
+                filemode = stat.S_IMODE(stat_path.st_mode)
                 os.chmod(self.full_roots[name], filemode)
+                # Set owner and group, too
+                os.chown(self.full_roots[name], stat_path.st_uid,
+                          stat_path.st_gid)
 
             except OSError as exception:
                 if exception.errno != errno.EEXIST:
@@ -129,9 +136,11 @@ to serve all files under specified web root ({0})."""
 
         # Set permissions as parent directory (GH #1389)
         parent_path = self.full_roots[achall.domain]
-        filemode = stat.S_IMODE(os.stat(parent_path).st_mode)
+        stat_parent_path = os.stat(parent_path)
+        filemode = stat.S_IMODE(stat_parent_path.st_mode)
         # Remove execution bit (not needed for this file)
         os.chmod(path, filemode & ~stat.S_IEXEC)
+        os.chown(path, stat_parent_path.st_uid, stat_parent_path.st_gid)
 
         return response
 
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -41,6 +41,7 @@ and reload your daemon.
 import errno
 import logging
 import os
+import stat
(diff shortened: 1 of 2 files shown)


letsencrypt/plugins/webroot.py
```diff
@@ -98,6 +99,16 @@ def prepare(self):  # pylint: disable=missing-docstring
                          self.full_roots[name])
             try:
                 os.makedirs(self.full_roots[name])
+                # Set permissions as parent directory (GH #1389)
+                filemode = stat.S_IMODE(os.stat(path).st_mode)
+                os.chmod(self.full_roots[name])
```
pde: `os.chmod` takes two arguments, not one.

pde:
Were you unable to run the unit tests for reasons unrelated to that bug? `tox -e py27` (to run all unit tests) or `venv/bin/python letsencrypt/plugins/webroot_test.py` (for just this test file) work for me, and should in general if you have [followed the developer instructions](https://letsencrypt.readthedocs.org/en/latest/contributing.html#running-a-local-copy-of-the-client).

pde:
`tox -e lint` is also pretty handy and will catch several types of bugs.

lbeltrame:
Thanks for the hints. Now I have fixed the issues, and added a proper unit test for the permission handling (passes).

bmw:
Travis isn't running on this PR. I'm not sure why...

pde:
@bmw because it's not against master.

lbeltrame:
In any case, I took care of running the webroot test manually (including my 
changes, which test for permissions) and passes now.

letsencrypt/plugins/webroot.py
```diff
@@ -98,6 +99,10 @@ def prepare(self):  # pylint: disable=missing-docstring
                          self.full_roots[name])
             try:
                 os.makedirs(self.full_roots[name])
+                # Set permissions as parent directory (GH #1389)
+                filemode = stat.S_IMODE(os.stat(path).st_mode)
+                os.chmod(self.full_roots[name], filemode)
```
pde: Question: is the reason to call `os.makedirs(file)` then `os.chmod(file, filemode)` rather than just `os.makedirs(file, filemode)` because `makedirs` is affected by umask, while `chmod` is not?  If yes, maybe comment that; if not, just pass filemode to `makedirs`.

pde:
This is looking good! I wonder if we should also be calling `os.fchown` to match the parent directory's uid and gid?  I think in some cases the failure to do so will also cause the challenge files to be unreadable by the webserver.

letsencrypt/plugins/webroot.py
lbeltrame: > Question: is the reason to call `os.makedirs(file)` then `os.chmod(file,
> filemode)` rather than just `os.makedirs(file, filemode)` because

According to https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python os.makedirs's mode parameter is 
not always honored. If that's OK I can add a comment with a link to that.

lbeltrame:
> This is looking good! I wonder if we should also be calling `os.fchown` to
> match the parent directory's uid and gid?  I think in some cases the

Good idea! I'll do that next.

letsencrypt/plugins/webroot.py
pde: SGTM.

lbeltrame:
Added chown() calls too, let me know if it's OK now.

lbeltrame:
Ping? Anything else that needs to be done? (Yes, I'm being pushy because I actually need this enhancement in)

pde:
Sorry, this was simply blocked on #1594 being merged.  Landing it now.

pde:
It's possible that some users who have executable-but-not-readable $WEBROOT/.well-known/ directories are getting breakage because of this...

lbeltrame:
Hm, how could those be handled?

PR #1600, opened 2015-11-23
Update configuration.py domain regex pattern error

ClintFoo:
The domain regex pattern was failing to recognize a valid domain and returning error "Requested domain is not a FQDN" when attempting to renew via cli commands. I removed extra backslash in front of the period to fix the pattern.


ClintFoo:
Pylint syntax errors indicate that the double backslash was needed to properly escape the pattern. Now I am not sure why it worked at all. Good news is that my client is now automatically renewing from command line on CentOS. Bad news is that I am not sure why. Back to the drawing board. Please pardon my pull.

PR #1603, opened 2015-11-24
Use reload in Apache (fixes #954) [needs minor extension]

### LINKED ISSUE #954: Apache and Nginx plugins should use reload rather than restart
letsencrypt-apache/letsencrypt_apache/configurator.py
>     .. todo:: Try to use reload instead. (This caused timing problems before)                                                                                                    

Our challenge methods are designed to be used without taking down a web server, so we should make sure we actually don't take down the web server to restart.

Note that Nginx has this issue too, noted in https://github.com/letsencrypt/letsencrypt/pull/950 and worked around.

We need a generic wait_for_challenge method that can be used for both plugins, which will repeatedly attempt to connect to the appropriate port on localhost and ask for the appropriate servername. If the loop times out after, say, five seconds, it should error out and say the configuration to add the challenge failed.

bmw:
This is the simple fix for #954. I really don't like this code. This is not the correct way to solve this problem. It is inefficient and is likely to fail for someone, somewhere.

With that said, I expect it will work fairly well. I tested it on a VPS while pounding the server with other connections and it never failed.

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -96,7 +96,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             help="Path to the Apache 'a2enmod' binary.")
         add("init-script", default=constants.CLI_DEFAULTS["init_script"],
             help="Path to the Apache init script (used for server "
-            "reload/restart).")
+            "reload).")
         add("le-vhost-ext", default=constants.CLI_DEFAULTS["le_vhost_ext"],
             help="SSL vhost configuration extension.")
         add("server-root", default=constants.CLI_DEFAULTS["server_root"],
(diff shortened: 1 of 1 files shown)


pde:
This is broadly enough for Public Beta launch, I think.  Is there anything quick we can do to keep the user reassured as they wait for a cert?  Perhaps dislpaying a series of dots as progress notifications, like we do in `letsencrypt-auto`?

bmw:
We could reduce the logging level to `info` and add calls to the logger with more information. I think this fits the best with the current structure of the code, but has the downside of still formatting the output as logging output. An example log would look like:

```
2015-11-24 08:43:39,937:INFO:letsencrypt.cli:Setting up challenges for domain validation
```

If you're okay with this approach, I'm happy to do the work. I'll do it in another PR. We can block merging this on the logging info PR if you'd like.

pde:
OK, let's do that in another PR.  If there's a quick way to take the "info" level logs and hide their timestamps in dialog / terminal output, that would be ideal, because I worry they'll overflow the display width and make things fairly unreadable.

PR #1608, opened 2015-11-24
Detect SSL vhosts by port

patf:
`SSLEngine on` can be set outside of `<VirtualHost>`. Treat any vhost using port 443 as a SSL vhost.

**Note:** In order to test this, I had to add a new vhost to the two_vhost_80 apache test config. This required some changes to assert statements in various test cases (to fix counters and such).


pde:
Thanks!

PR #1609, opened 2015-11-24
Remove Content-Type checks from http-01

patf:
Content-Type type restrictions were removed in ACME, see
https://github.com/ietf-wg-acme/acme/commit/69ac2baade014796e5258a077e7600921cd1879d

Boulder does not restrict Content-Type currently, see https://github.com/letsencrypt/boulder/issues/1089 (which is obsolete).

Manual test against staging succeeded (\w manual and standalone plugin).

### DIFF
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -269,13 +269,6 @@ class HTTP01Response(KeyAuthorizationChallengeResponse):
         logger.debug("Received %s: %s. Headers: %s", http_response,
                      http_response.text, http_response.headers)
 
-        found_ct = http_response.headers.get(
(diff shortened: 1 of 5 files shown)


patf:
I guess it's one of those flaky integration tests? Will try triggering Travis again.

bmw:
LGTM. Thanks a lot for the PR @patf. Fixing this was a pretty high priority.

We may want to consider adding a `Content-Type` header to `standalone` at some point just so the response has one, but it's not necessary, especially for this PR.

kuba:
Please mark all PR/bugs that touch `acme` with `acme` label, thanks!

Issue #1362, opened 2015-11-05
Suppress insecure platform warnings from python versions that lack SNI client support


pde:
These confuse people a lot.  They exist for a good reason, but it isn't relevant to the Let's Encrypt client.

ArchimedesPi:
@pde I'd be happy to take this one. How do I replicate the warnings?

devnsec-com:
@ArchimedesPi run `letsencrypt` on a platform that has Python<2.7.9 installed (and `pyOpenSSL` not installed) will reproduce this warning.

bmw:
I think this issue has been resolved since Certbot requires `pyOpenSSL` which is used to avoid these warnings[[1]](https://github.com/certbot/certbot/blob/cf29e89366c7eb57dd48a99a06fe05ceaa9057fa/acme/acme/client.py#L35).

I'm closing this issue but if someone is still hitting this on an up-to-date version of Certbot, please comment or open a new issue.

PR #1623, opened 2015-11-26
Suppress InsecurePlatformWarning during letsencrypt-auto venv setup

pde:
It's not good that pip is using an old version of requests / urllib3, but since we've installed the latest version available on this OS, it's not our job to yell at the user about it further.

Once letsencrypt is running inside the venv, warnings are reactivated, and at that point we have a secure python platform.

Closes: #1362

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -179,7 +179,12 @@ if [ "$VERBOSE" = 1 ]  ; then
     $VENV_BIN/pip install -U letsencrypt letsencrypt-nginx
   fi
 else
-  $VENV_BIN/pip install -U setuptools > /dev/null
+  # Python's InsecurePlatformWarning is about a number of flaws in old python
+  # libraries. The fact that it's raised by these commands is a result of the
+  # system's copy of pip not having access to a good python TLS binding. But
+  # we've tried our best, and within our venv letsencrypt itself will be safe.
+  export PYTHONWARNINGS=ignore
+  $VENV_BIN/pip install -U setuptools 2>&1 > /dev/null
   printf .
   $VENV_BIN/pip install -U pip > /dev/null
   printf .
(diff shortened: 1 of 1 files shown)


bmw:
Testing this on Debian Squeeze, I still see quite a few InsecurePlatformWarnings.

I also don't think we should be squelching all warnings. Sure, InsecurePlatformWarnings are annoying and unnecessarily worry some people, but there may be larger problems on their system which we are hiding.

pde:
When I first opened #1362 I thought this warning was just about the lack of SNI support, which is a terrible screw-up on Python's part, but not something we have to worry about.

Actually it seems that InsecurePlatformWarning may be evidence of [more serious problems](https://github.com/shazow/urllib3/issues/602#issuecomment-97054951) that affect us at bootstrapping time.  I still think we need to suppress it, because it's confusing.  People will think it's Let's Encrypt that's the problem.  We should figure out how to detect it, and print a warning that users need an OS upgrade...

devster31:
I think the warning on pip install is unavoidable unless you suppress warnings for the installation, other errors however should be fixable, for example in `acme/acme/client.py` there's this:
```python
# Prior to Python 2.7.9 the stdlib SSL module did not allow a user to configure
# many important security related options. On these platforms we use PyOpenSSL
    requests.packages.urllib3.contrib.pyopenssl.inject_into_urllib3()
```
that should remove the warning by using PyOpenSSL and therefore removing potential vulnerabilities. There are other files that are using requests that don't have the above code in it, such as `acme/acme/challenges.py` and `letsencrypt/validator.py`.

erikrose:
1. pip output is hidden as of #1665 unless there's an actual error.
2. We now check hashes on all Python package downloads, so HTTPS cert checking is superfluous.

Thus, I don't think we need this anymore.

PR #1627, opened 2015-11-27
Prioritise dnf over yum

ajhaydock:
Perhaps it is better to prioritise dnf over yum here, like this?

dnf officially replaces yum, so it's probably safe to assume that it's the preferred package manager on any system with it installed


pde:
Duplicate of #1519 .

Issue #1630, opened 2015-11-27
letsencrypt-auto cant open 'py26reqs.txt' when called from a script


MikeSkril:
Tested on Ubuntu 14.04.

    /opt/letsencrypt/letsencrypt-auto \
    --agree-dev-preview \
    --renew-by-default \
    -a webroot \
    --webroot-path /var/www \
    --email postmaster@xyz.xyz \
    --verbose \
    --text \
    -d xyz.xyz \
    certonly

This scrip will result in:
 ```Updating letsencrypt and virtual environment dependencies...
   Requirement already up-to-date: setuptools in /root/.local/share/letsencrypt/lib/python2.7/site   packages
   Requirement already up-to-date: pip in /root/.local/share/letsencrypt/lib/python2.7/site-packages
   Could not open requirements file: [Errno 2] No such file or directory: 'py26reqs.txt'```
   
I have to move into the LE directory to run it successfully.
   
    cd  /opt/letsencrypt/
    .letsencrypt-auto \
     --agree-dev-preview \
     --renew-by-default \
     -a webroot \
     --webroot-path /var/www \
     --email postmaster@xyz.xyz \
     --verbose \
     --text \
     -d xyz.xyz \
     certonly

pde:
Funny, I noticed this for the first time yesterday too...

MikeSkril:
Thanks for taking care! :-)

the-real-tokai:
I encountered this too today, when setting up my automatisation scripts:

```
...
Updating letsencrypt and virtual environment dependencies.....Could not open requirements file: [Errno 2] No such file or directory: 'py26reqs.txt'
...
```

An easy workaround is to change into the letsencrypt git checkout directory before executing the command, e.g. in a bash script it could look like this:

```
...
( cd /some/path/letsencrypt && ./letsencrypt-auto \
	-a webroot --webroot-path /sshfsmounts/example.org \
	-d example.org -d www.example.org \
	--server "https://acme-v01.api.letsencrypt.org/directory" \
	--email "foobar@example.org" \
	--agree-dev-preview --agree-tos
)
...
```

PR #1631, opened 2015-11-27
py26reqs.txt needs to be path-relative

pde:
Fixes: #1630


pde:
@levskaya this would be a good example of a pull request to run through multi-OS tests.

bmw:
I tested everything locally and this LGTM.

razerrazer:
Is an easy fix, just go in to letsencrypt folder (/etc/letsencrypt) then you just type : touch py26reqs.txt
Then the problem is solved..

PR #1635, opened 2015-11-29
poll_and_ri: handle STATUS_INVALID, add max_attempts (fixes #1634)

### LINKED ISSUE #1634: poll_and_request_issuance loops forever on failure
In https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/client.py#L326, we only check for STATUS_VALID. We should also check for STATUS_INVALID and return once that happens, or there will be many clients looping forever making requests to the server.

Similarly, that method should have a timeout.

kuba:
Fixes #1634

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -246,9 +246,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
     def retry_after(cls, response, default):
         """Compute next `poll` time based on response ``Retry-After`` header.
 
-        :param response: Response from `poll`.
-        :type response: `requests.Response`
-
+        :param requests.Response response: Response from `poll`.
         :param int default: Default value (in seconds), used when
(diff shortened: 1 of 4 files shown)


kuba:
cc: @jsha

jsha:
LGTM. I'll leave it to a more active client developer to do the final review and merge. Thanks for fixing, @kuba!

jsha:
BTW: reminder to put the "fixes #XXX" in the body rather than the title. This makes it clickable, and makes GitHub auto-resolve the issue on merge. I've edited this PR to move it.

kuba:
@jsha, my experience shows that GitHub recognizes any of those three:
- PR title
- PR body
- commit description (first line or body)

bmw:
LGTM. Thanks @kuba!

PR #1639, opened 2015-12-01
Don't use Apache init scripts

### LINKED ISSUE #1638: letsencrypt stops apache, then fails
Relevant log except:

```
2015-11-30 21:18:48,589:INFO:letsencrypt.reporter:Reporting to user: The following 'urn:acme:error:connection' errors were reported by the server
:

Domains: roundcube.singpolyma.net
Error: The server could not connect to the client for DV
2015-11-30 21:18:48,614:INFO:letsencrypt.auth_handler:Cleaning up challenges
2015-11-30 21:18:48,782:ERROR:letsencrypt_apache.configurator:Apache Reload Failed!
Reloading web server: apache2 failed!
Apache2 is not running ... (warning).
```

After which apache is not running.  It was running before I ran letsencrypt (running from master, latest) and starting it again works fine.  It seems that letsencrypt is stopping it and then complaining that it is stopped...

bmw:
Partially resolves #1638. PR removes reliance on Debian specific Apache init scripts, restores the ability to run `letsencrypt --apache` if Apache isn't currently running, and adds utility function for registering deprecated command line flags.

### DIFF
--- a/letsencrypt/le_util.py
+++ b/letsencrypt/le_util.py
@@ -1,4 +1,5 @@
 """Utilities for all Let's Encrypt."""
+import argparse
 import collections
 import errno
 import logging
@@ -255,3 +256,23 @@ def safe_email(email):
     else:
         logger.warn("Invalid email address: %s.", email)
         return False
+
+
+def add_deprecated_argument(add_argument, argument_name):
+    """Adds a deprecated argument with the name argument_name.
+
+    Deprecated arguments are not shown in the help. If they are used on
+    the command line, a warning is shown stating that the argument is
+    deprecated and no other action is taken.
+
+    :param callable add_argument: Function that adds arguments to an
+        argument parser/group.
+    :param str argument_name: Name of deprecated argument.
+
+    """
+    class ShowWarning(argparse.Action):
+        """Action to log a warning when an argument is used."""
+        def __call__(self, unused1, unused2, unused3, option_string=None):
+            print "Use of {0} is deprecated".format(option_string)
+
+    add_argument(argument_name, action=ShowWarning, help=argparse.SUPPRESS)
--- a/letsencrypt/le_util.py
+++ b/letsencrypt/le_util.py
@@ -276,7 +276,7 @@ def add_deprecated_argument(add_argument, argument_name, nargs):
         """Action to log a warning when an argument is used."""
         def __call__(self, unused1, unused2, unused3, option_string=None):
             sys.stderr.write(
-                "Use of {0} is deprecated\n".format(option_string))
+                "Use of {0} is deprecated.\n".format(option_string))
 
     add_argument(argument_name, action=ShowWarning,
                  help=argparse.SUPPRESS, nargs=nargs)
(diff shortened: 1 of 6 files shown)


bmw:
Test failures should be fixed by letsencrypt/boulder#1214.

letsencrypt/le_util.py
```diff
@@ -255,3 +257,26 @@ def safe_email(email):
     else:
         logger.warn("Invalid email address: %s.", email)
         return False
+
+
+def add_deprecated_argument(add_argument, argument_name, nargs):
```
pde: It feels like this should be part of HelpfulArgumentParser or at least cli.py, rather than le_util.

pde: (though I'm fine throwing that task in the "refactor" pile for later)

pde:
While testing this on Ubuntu 14.04:

```
  File "venv/bin/letsencrypt", line 9, in <module>
  File "/home/ubuntu/letsencrypt/letsencrypt/cli.py", line 1225, in main
  File "/home/ubuntu/letsencrypt/letsencrypt/cli.py", line 451, in run
    installer, authenticator = choose_configurator_plugins(args, config, plugins, "run")
  File "/home/ubuntu/letsencrypt/letsencrypt/cli.py", line 421, in choose_configurator_plugins
    authenticator = installer = display_ops.pick_configurator(config, req_inst, plugins)
  File "/home/ubuntu/letsencrypt/letsencrypt/display/ops.py", line 123, in pick_configurator
  File "/home/ubuntu/letsencrypt/letsencrypt/display/ops.py", line 80, in pick_plugin
    verified.prepare()
  File "/home/ubuntu/letsencrypt/letsencrypt/plugins/disco.py", line 213, in prepare
    return [plugin_ep.prepare() for plugin_ep in self._plugins.itervalues()]
  File "/home/ubuntu/letsencrypt/letsencrypt/plugins/disco.py", line 103, in prepare
    self._initialized.prepare()
  File "/home/ubuntu/letsencrypt/letsencrypt-apache/letsencrypt_apache/configurator.py", line 142, in prepare
    if not le_util.exe_exists(exe):
  File "/home/ubuntu/letsencrypt/letsencrypt/le_util.py", line 74, in exe_exists
    path, _ = os.path.split(exe)
  File "/home/ubuntu/letsencrypt/venv/lib/python2.7/posixpath.py", line 92, in split
    i = p.rfind('/') + 1
AttributeError: 'NoneType' object has no attribute 'rfind'
```

letsencrypt/le_util.py
bmw: I'd argue that it should not go in `HelpfulArgumentParser` as then you'd need access to the parser object to be able to call the function. This approach gives 3rd party plugins a function they can use to deprecate their own arguments.

If we split `cli.py` into `main.py` and `cli.py` as you have suggested, I think putting it in `cli.py` makes sense. I personally like it best in `le_util.py` based on the current structure though.

bmw:
Oh man. Thanks for catching that. My last commit before submitting this PR was supposed to only modify the unit tests but one of the changes I made broke everything.

Problem has been fixed. Ready for re-review.

pde:
Now LGTM.

PR #1640, opened 2015-12-01
Run RabbitMQ setup during test setup.

jsha:
This was recently introduced on the Boulder side.

Note: long-term we want to have the client tests run the same setup steps as
Boulder does, with the same script. This is a quick fix to unbreak the build.


bmw:
LGTM. Merging when Travis is happy.

kuba:
@jsha, if your planning a shared script could you please make sure it's reusable by other LE clients, including https://github.com/kuba/simp_le?

jsha:
That should be possible, see https://github.com/letsencrypt/boulder/pull/1215.

PR #1641, opened 2015-12-01
Start a library of apache2 conf files for tests

pde:
With examples of passing and failing conf files sourced from our github
tickets.


bmw:
LGTM

PR #1651, opened 2015-12-02
Prepare for public beta

### LINKED ISSUE #887: Remove or edit --dev-preview for Public Beta
What's the plan with these flags? From the help it looks like --agree-eula applies only to the "Developer Preview." Are we planning to remove that one once we have a cross-signature in place and change the default server URL to point at prod?

bmw:
Fixes #887. Please review this PR but don't merge it yet. We want to get at least one, if not two, PyPI releases out before this is merged.

### DIFF
--- a/DISCLAIMER
+++ b/DISCLAIMER
@@ -1,1 +1,0 @@
-letsencrypt/DISCLAIMER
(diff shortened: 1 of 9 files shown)


pde:
Should we add `--staging` at the same time?

bmw:
The changes made in the current version of this PR are necessary for public beta. `--staging` is not so I'd prefer not to block on it.

pde:
It's showtime.

Issue #1589, opened 2015-11-21
Document supported Python versions


bmw:
People have tried running Python on older OSs and have seen:
```
You have an ancient version of Python entombed in your operating system...
```
We should document what versions of Python we support in both the README and readthedocs.

cowlicks:
Anyone know what 3.x versions are supported? 3.3+?

bmw:
@cowlicks, the client does not support Python 3.x at this time. We plan to add Python 3.x support after we launch.

Closed by #1593.

PR #1653, opened 2015-12-02
Make the ancient python error more friendly

pde:
Somewhat fixing #1589


pde:
Self-merging this trivial one-liner.

PR #1661, opened 2015-12-02
Remove py26reqs.txt

mgedmin:
ConfigArgParse 0.10 from PyPI supports Python 2.6, so there's no more
need to install a fixed version directly from a git branch.

### DIFF
--- a/bootstrap/README
+++ b/bootstrap/README
@@ -2,6 +2,5 @@ This directory contains scripts that install necessary OS-specific
 prerequisite dependencies (see docs/using.rst).
 
 General dependencies:
-- git-core: git+https://*
 - ca-certificates: communication with demo ACMO server at
-  https://www.letsencrypt-demo.org, git+https://*
+  https://www.letsencrypt-demo.org
--- a/bootstrap/README
+++ b/bootstrap/README
@@ -2,6 +2,6 @@ This directory contains scripts that install necessary OS-specific
 prerequisite dependencies (see docs/using.rst).
 
 General dependencies:
-- git-core: py26reqs.txt git+https://*
+- git-core: git+https://*
 - ca-certificates: communication with demo ACMO server at
-  https://www.letsencrypt-demo.org, py26reqs.txt git+https://*
+  https://www.letsencrypt-demo.org, git+https://*
--- a/setup.py
+++ b/setup.py
@@ -52,13 +52,13 @@ if sys.version_info < (2, 7):
     install_requires.extend([
         # only some distros recognize stdlib argparse as already satisfying
         'argparse',
-        'mock<1.1.0',
         'ConfigArgParse>=0.10.0',  # python2.6 support, upstream #17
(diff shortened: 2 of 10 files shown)


bmw:
LGTM. Thanks for noticing this @mgedmin. I didn't realize that @kuba's changes had landed in the official tree.

I'll leave this up to @pde on whether we want to make this change before public beta.

kuba:
0.10 was released without fixing https://github.com/bw2/ConfigArgParse/issues/26? :( (Un)luckily, we stopped using subparsers, so we should not be affected.

`setup.py` should be updated to `argaprse>0.10` in case of py26

bootstrap/README
```diff
@@ -2,6 +2,6 @@ This directory contains scripts that install necessary OS-specific
 prerequisite dependencies (see docs/using.rst).
 
 General dependencies:
-- git-core: py26reqs.txt git+https://*
```
kuba: if this is removed, then `git-core` should be removed as well, or the comment should be updated if, some other projects are downloaded using `pip install git+http...`

mgedmin:
> setup.py should be updated to argaprse>0.10 in case of py26

I assume you meant ConfigArgParse>=0.10.0?

bootstrap/README
kuba: FYI, I partially meant removal from deps in bootstrap scripts

setup.py
```diff
@@ -32,7 +32,7 @@ def read_file(filename, encoding='utf8'):
 
 install_requires = [
     'acme=={0}'.format(version),
-    'ConfigArgParse',
+    'ConfigArgParse>=0.10.0',  # python2.6 support, upstream #17
```
kuba: Version pin should be moved to else branch of sys.version_info < (2, 7) condition checked below.

mgedmin: Hmm.  Dynamically-computed `install_requires` causes problems for users of pip >= 7.0, which has wheel caching enabled by default: https://github.com/pypa/pip/issues/3025.  Short version:
- I pip install letsencrypt on Python 2.7, pip runs setup.py and builds a wheel (marked -py2-none-any.whl) and puts it in ~/.cache/pip/
- I pip install letsencrypt on Python 2.6, pip discovers a valid wheel and installs it, using `install_requires` computed earlier for a different Python version.

What's the harm of having everyone use a recent version of ConfigArgParse?

mgedmin:
Rebased to resolve merge conflict in Dockerfile-dev.

setup.py
kuba: Well, if that causes the problem, then please open a separate bug to discuss changing our strategy. But this PR is not the best place to discuss is, or deviate from already established standards.

One potential harm that instantly comes to my mind is that it makes OS packaging a bit harder (maintainers, such as @hlieberman, will have to patch `setup.py` unnecessarily).

mgedmin:
#1908 supersedes this.

PR #1663, opened 2015-12-02
Fixed some spelling errors.

the-real-tokai:
Resolves a few harmless "doubling" issues in the client's help text.


bmw:
LGTM. Thanks for catching these @the-real-tokai.

Issue #1256, opened 2015-10-31
bootstrap/venv.sh


kuba:
As suggested multiple times before, this file should probably be included from `letsencrypt-auto` or deleted otherwise. Currently it's not used anywhere and it's not mentioned in the docs. Maintaining both `letsencrypt-auto` and `bootstrap/venv.sh` is a hassle.

devnsec-com:
I think just simply delete the `bootstrap/venv.sh` file will do. I did a quick `grep`, didn't find anything references to this file.

bmw:
I personally think the opposite and we should call out to `bootstrap/venv.sh`. It makes the code more modular and easier to understand. Since we are currently having `letsencrypt-auto` users clone our entire repo for OS bootstrap scripts, `letsencrypt-auto` calling out to `bootstrap/venv.sh` is not a problem. Furthermore, if we adopt either approach proposed in #1258, calling out to `bootstrap/venv.sh` will still work.

PR #1665, opened 2015-12-02
Make letsencrypt-auto release-suitable, self-upgrading, and self-contained.

### LINKED ISSUE #1572: Make letsencrypt-auto more paranoid.
Make letsencrypt-auto use PyPI (where we'll put our dev releases from now on), a signed requirements file, and peep.

Prereq: #1571.

* letsencrypt-auto will go figure out what the latest version of LE is ([from PyPI](https://pypi.python.org/pypi/letsencrypt/json) (look at the `version` key in the first dict) or GH--over HTTPS to make it harder to artificially keep LE out of date) then wget that version's reqs.txt and signature off GH.
* It'll verify the sig with `openssl` (already in the bootstrap scripts and easy to airgap sigs for).
* Then it'll `peep install` that reqs.txt into the venv.

erikrose:
Fix #1572.

Here's [the comment that describes the design](https://github.com/letsencrypt/letsencrypt/issues/1572#issuecomment-159827232). Here's what remains to be done:

- [x] [Gather requirement hashes.](https://github.com/letsencrypt/letsencrypt/issues/1571)
- [x] Make le-auto not upgrade letsencrypt unless necessary.
- [x] Make more easily testable.
- [x] Test.
- [x] Review.
- [x] Add building and signing le-auto to release process.
- [x] Make peep compatible with pip 7.
- [x] Add an option that skips the self-upgrade, for people who have audited this and want to run only this version.
- [x] Figure out under what conditions to install dependencies needed by only Python 2.6 (e.g. https://github.com/letsencrypt/letsencrypt/pull/1914/files).


letsencrypt_auto/letsencrypt-auto.template
```diff
@@ -141,65 +81,146 @@ then
...
+      su root -c "$args"
+    }
+    SUDO=su_sudo
   fi
 else
-  DeterminePythonVersion
+  SUDO=
 fi
 
+if [ "$1" = "--os-packages-only" ]; then
```
pde: This flag needs to be at least accepted by the main `letsencrypt` Python client, and probably documented in its --help.

letsencrypt_auto/letsencrypt-auto.template
```diff
@@ -141,65 +81,146 @@ then
...
+    Bootstrap
+    echo "Creating virtual environment..."
+    rm -rf "$VENV_PATH"
+    DeterminePythonVersion
+    if [ "$VERBOSE" = 1 ] ; then
+      virtualenv --no-site-packages --python $LE_PYTHON $VENV_PATH
+    else
+      virtualenv --no-site-packages --python $LE_PYTHON $VENV_PATH > /dev/null
+    fi
+
```
pde: Perhaps add a big visual separator line here, because we're about to change languages and universes :)

letsencrypt_auto/letsencrypt-auto.template
```diff
@@ -141,65 +81,146 @@ then
...
+else  # --_skip-to-install was passed.
+  # Install Python dependencies with peep, then run letsencrypt.
+  echo "Installing Python package dependencies..."
+  TEMP_DIR="$2"
+  shift 2
+  cat << "UNLIKELY_EOF" > $TEMP_DIR/letsencrypt-auto-requirements.txt
+{{ letsencrypt-auto-requirements.txt }}
+UNLIKELY_EOF
+
+  cat << "UNLIKELY_EOF" > $TEMP_DIR/peep.py
```
pde: Perhaps add a big visual separator line here, because we're about to change languages and universes :)

letsencrypt_auto/letsencrypt-auto.template
```diff
@@ -141,65 +81,146 @@ then
...
+{{ peep.py }}
+UNLIKELY_EOF
```
pde: And another one here as we return to shell?

letsencrypt_auto/letsencrypt-auto.template
pde: Once that's done we can move all flag detection to the loop on line 105, so that the arguments are ordering-agnostic.

letsencrypt_auto/letsencrypt-auto.template
```diff
@@ -141,65 +81,146 @@ then
...
+    SUDO=su_sudo
   fi
 else
-  DeterminePythonVersion
+  SUDO=
 fi
 
+if [ "$1" = "--os-packages-only" ]; then
+  Bootstrap
+elif [ "$1" != "--_skip-to-install" ]; then
```
pde: Is this flag semantically equivalent to "do not autoupdate"?  If so, we should definitely document it and let users supply it :)

docs/contributing.rst
```diff
@@ -22,7 +22,7 @@ once:
 
    git clone https://github.com/letsencrypt/letsencrypt
    cd letsencrypt
-   ./bootstrap/install-deps.sh
+   ./letsencrypt-auto/letsencrypt-auto --os-packages-only
```
pde: Once we have `--no-autoupdate` or something, we should probably suggest it on this line?

letsencrypt_auto/letsencrypt-auto.template
erikrose: It makes sense that the options to letsencrypt should be tolerated (passed along, really) by letsencrypt-auto, but why the other way around?

letsencrypt_auto/letsencrypt-auto.template
erikrose: No; if someone manually passes this flag, it could cause le-auto to try installing Python packages without a virtualenv made.

erikrose: Now it is. Rename the arg and made it public in be6c34de32eec35b861decc08fedf94a0d3a566e. Moved the venv creation bit to make it work.

erikrose:
`letsencrypt-auto --no-self-upgrade --debug` just worked great on my Mac, as of 02255fa.

letsencrypt_auto/build.py
```diff
@@ -0,0 +1,50 @@
...
+        if token in special_replacements:
+            return special_replacements[token]
+        else:
+            return file_contents(join(dir, 'pieces', token))
+
+    result = re.sub(r'{{\s*([A-Za-z0-9_./-]+)\s*}}',
+                    replacer,
+                    file_contents(join(dir, 'letsencrypt-auto.template')))
+    with open(join(dir, 'letsencrypt-auto'), 'w') as out:
+        out.write(result)
```
pde: We also need to `os.chmod()` the resulting script to be executable.

erikrose: Good idea. (If the old one were +x, the new one would be too due to cp's behavior, but if they ran "sh letsencrypt-auto", it might not have been +x.)

erikrose:
I have a crazy HTTPS-serving test harness starting to take shape locally (fake CA, signed certs, the whole bit), but it's not useful yet.

pde:
Closes: #1256

pde:
Fixes: #2107

erikrose:
On merging, restore the old le-auto script and the bootstrap dir to the root dir. Delete them after we're satisfied the new le-auto is working.

pde:
Do we know why the tests are failing in this branch?

erikrose:
@pde: It looks like something related to MySQL connections in the Boulder tests, probably a side effect of changing to `sudo: required`. Ideas are welcome. I haven't dug in much yet.

pde:
I'd say that for now, let's run all nosetests that require sudo manually outside of travis?  That's what I'm doing with `tox -e apacheconftest`.

acme/.pylintrc
```diff
@@ -60,7 +60,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=fixme,locally-disabled,abstract-class-not-used
```
kuba: that doesn't seem to belong to this PR, please revert

acme/setup.py
```diff
@@ -22,15 +22,14 @@
 ]
 
 # env markers in extras_require cause problems with older pip: #517
+# Keep in sync with conditional_requirements.py.
 if sys.version_info < (2, 7):
     install_requires.extend([
         # only some distros recognize stdlib argparse as already satisfying
         'argparse',
-        'mock<1.1.0',
```
kuba: I'm really not convinced about moving this out of `install_requires`

acme/setup.py
```diff
@@ -75,6 +74,7 @@
     packages=find_packages(),
     include_package_data=True,
     install_requires=install_requires,
+    tests_require='mock<1.1.0' if sys.version_info < (2, 7) else 'mock',
```
kuba: if we're really going with this approach (see https://github.com/letsencrypt/letsencrypt/pull/1665/files#r49645236), then please use a `tests_require` variable just as we do for all other package lists

docs/contributing.rst
bmw: Incorrect path.

docs/contributing.rst
```diff
@@ -365,10 +365,13 @@ Now run tests inside the Docker image:
...
 ========================
 
-OS level dependencies are managed by scripts in ``bootstrap``.  Some notes
-are provided here mainly for the :ref:`developers <hacking>` reference.
+OS-level dependencies can be installed like so:
 
-In general:
+.. code-block:: shell
+
+    letsencrypt-auto/letsencrypt-auto --os-packages-only
```
bmw: Ditto

acme/setup.py
```diff
@@ -22,15 +22,14 @@
 ]
 
 # env markers in extras_require cause problems with older pip: #517
+# Keep in sync with conditional_requirements.py.
```
kuba: What is the plan for this when acme gets into it's own repo? it will become pretty much infeasible to keep those in sync. Looking at the problem from different perspective, how do you make sure in `conditional_requirements.py` that it pulls conditional deps from our own dependencies that we don't have control over?

.travis.yml
```diff
@@ -45,14 +47,13 @@ branches:
     - master
     - /^test-.*$/
 
-# container-based infrastructure
-sudo: false
+sudo: required
```
kuba: That setting was pushed back and forth a lot of times... Why is it necessary this time and how does it relate to what this PR is trying to solve?

.travis.yml
```diff
@@ -2,13 +2,14 @@ language: python
...
   - rabbitmq
-  - mariadb
   # apacheconftest
   #- apache2
 
 # http://docs.travis-ci.com/user/ci-environment/#CI-environment-OS
 # gimme has to be kept in sync with Boulder's Go version setting in .travis.yml
 before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq mysql-server-5.6 mysql-client-5.6 mysql-client-core-5.6
```
kuba: Again, this PR is about `letsencrypt-auto` - why is `mysql` touched here?

Dockerfile
```diff
@@ -22,8 +22,8 @@ WORKDIR /opt/letsencrypt
 # directories in its path.
 
 
-COPY bootstrap/ubuntu.sh /opt/letsencrypt/src/ubuntu.sh
-RUN /opt/letsencrypt/src/ubuntu.sh && \
+COPY letsencrypt-auto-source/letsencrypt-auto /opt/letsencrypt/src/letsencrypt-auto
```
bmw: Since we're not instructing users to make the shift to the new `letsencrypt-auto` yet, I don't think we should be making the changes to the Dockerfile now.

Vagrantfile
```diff
@@ -7,7 +7,7 @@ VAGRANTFILE_API_VERSION = "2"
 # Setup instructions from docs/contributing.rst
 $ubuntu_setup_script = <<SETUP_SCRIPT
 cd /vagrant
-./bootstrap/install-deps.sh
+./letsencrypt-auto/letsencrypt-auto --os-packages-only
```
bmw: Bad path

docs/contributing.rst
bmw: Actually, I think we should just make it simpler and revert all changes to this file for now.

acme/.pylintrc
erikrose: I'll roll this back with the Travis config rollback. Its relevance lay in the Travis config, in that the linter mysteriously ignored the too-many-attrs "problem" on master but not on this branch.

acme/setup.py
erikrose: I think we resolved this to our mutual satisfaction in IRC. Kuba's concern was that test_utils routines in acme might be called from future third-party acme clients and that test_utils-only dependencies would thus not get installed. We will solve this, if it arises, by keeping a [testing] extra in acme that includes mock and having the third-party client's `test_requires='acme[testing']`.

docs/contributing.rst
erikrose: I think we can be confident in the le-auto-based bootstrapping at this point. It's textually almost identical to the old scripts, and we've vetted it on the test farm and in the community. Furthermore, we're talking to devs here, and they should be using the latest, not least because they know how to file bugs at us if it *doesn't* work. :-)

Dockerfile
erikrose: Again, this is a dev-only thing, and I think we're pretty confident it works. Do you really want to revert it?

.travis.yml
erikrose: It allows us to run the le-auto integration tests, which need sudo almost by definition, on Travis. (I isolate them in a Docker container so our scribbling doesn't interfere with other tests and so they can be run locally.)

.travis.yml
erikrose: Please see the commit message.

acme/setup.py
kuba: Unfortunately, **we haven't been able to reach a consensus** here yet. In fact, after more extensive IRC discussion, I'm pretty sure **this should not be merged** here.

In general, I strongly believe that:

1. `le-auto` should not impose any constraints on `setup.py`. By the way, how would you impose such constraints on `setup.py` files found external dependencies?

2. Up so far, we have kept our tests together with the main source code. As long as we do it, we should also have test deps in `install_requires`, because that code is shipped to end-users and importable. Rationale for such setup is not really in the scope of this PR, but I'm open to discussion elsewhere. Moreover, messing with this is likely introduce more bugs. That particular concern Erik brought up, was the quickest argument I could come up with, in that particular context.

An obvious solution here is to have `le-auto` pin `"mock==1.1.0"` as this would satisfy `acme`'s setup.py constraints under both Python 2.6 and 2.7. Implementing this plan is about adding one new line to pin mock in `conditional_requirements.py` and one (or more) to add it's hash (anyway it must be easy - it was meant to be this way, because we're likely often to update those pins and their hashes - that's the whole point of this PR). Touching `setup.py` is really the wrong hammer...

Apologies if you think this comes out late in review, @erikrose . This PR was hanging in the queue as WIP almost forever now, lastests commits are still being added, and it was never labelled with `acme` - hence I didn't realize it's touching `acme/` subproject. However, I've hinted you about my concerns much earlier than the comment above.

kuba: FTR Remark 1. goes together with my generic comment about `conditional_requirements.py` from https://github.com/letsencrypt/letsencrypt/pull/1665/files#r49646113.

acme/setup.py
erikrose: That, unfortunately, is a price we pay for deeply pinning dependencies: more security, less dynamism. I haven't audited our third-party dependencies for conditional subdependencies. That said, we haven't received any test reports about failures spawned therefrom.

acme/setup.py
erikrose: Upon further discussion, I don't think we've reached agreement yet after all.

pde: I don't really agree that we need to split tests into separate packages before removing `mock` as a dependency (I think it's not insane to have a package that functions with core functionality without some extra package X, but which can do more -- such as run tests -- if X is present).

But I do agree that there's some risk of introducing bugs this way.  So my vote would be pin `mock==1.1.0` for this release, then take more time to consider possible breakage cases, perhaps wrap `import mock`s in an error handler, and then revert the `mock==1.1.0` pin for a future release?

pde: After much further discussion on IRC and at EFF, our revised conclusion is: we'd like to move the mock dependency out of the main setup.py, but we aren't going to rush shipping that way, so 0.2.0 is going out without this change and we can do it in a more relaxed manner for 0.2.1 on Monday.

kuba: Would you mind actually addressing my concerns and suggested solution instead of ignoring it? I don't see any relevant discussion in IRC either.

To make it explicit: please don't remove `mock` dependency (at least as a part of this PR). Thanks!

pde: @kuba absolutely yes!  We're not going to land the mock removal in a hurry and without your agreement.  Regarding your questions:

1. Obviously we can't modify the `setup.py` of third party dependencies unless we fork of vendorise those projects, though in some cases (ConfigArgParse) we've already been tempted to do that, and it's not totally out of the question.  We'd always prefer to work with those projects to keep their dependency sets minimal if possible.

2. How would you feel about a separate PR against `acme` which (1) removes mock from the main dependency set; (2) raises a planned exception if the mock-dependent test code is called without mock installed; (3) documents this behaviour as part of the API?

pde:
LGTM! @bmw want to give this a final look over and merge?

acme/setup.py
kuba: This sounds a bit scary. Even simplest changes to `setup.py` tend to cause problems as we've seen with #1914. I hope you're ready to receive a bunch of bug reports from affected users. Or maybe it's better to perform such audit? In fact, you may want to have an automatic script that does it (before merging this PR), as any dep version bump might cause troubles.

tox.ini
```diff
@@ -8,23 +8,20 @@
...
-# packages installed separately to ensure that dowstream deps problems
+# packages installed separately to ensure that downstream deps problems
 # are detected, c.f. #1002
 commands =
-    pip install -e acme[testing]
-    nosetests -v acme
-    pip install -e .[testing]
-    nosetests -v letsencrypt
+    pip install -e acme
+    python acme/setup.py test
```
kuba: I believe all lines in `tox.ini` should be reverted, including above 2, due to the fact we changed handling of install_requires and tests_requires. Especially given the fact that you removed `tests_require=install_requires` lines throughout.

tox.ini
```diff
@@ -8,23 +8,20 @@
 skipsdist = true
 envlist = py{26,27,33,34,35},py{26,27}-oldest,cover,lint
 
-# nosetest -v => more verbose output, allows to detect busy waiting
-# loops, especially on Travis
```
kuba: so why is `nosetest` actually removed? If it is removed here, why do we keep `'nose'` in `testing_extras` throughout?

erikrose: Replacing `nosetests` with `python setup.py test` allowed us to move mock to `test_requires`; setuptools then installs it on demand, when tests are run. That was the immediate practical benefit. However, since you've argued for not moving mock yet, there's less reason to do this at the moment. If you like, I can bundle it with the later PR that moves mock.

acme/setup.py
erikrose: It is a bit scary. My favored approach (please see https://github.com/letsencrypt/letsencrypt/pull/2182#issuecomment-172099220 and the following comment) is to do away with conditional dependencies.

tox.ini
```diff
@@ -12,19 +12,19 @@ envlist = py{26,27,33,34,35},py{26,27}-oldest,cover,lint
 # loops, especially on Travis
 
 [testenv]
-# packages installed separately to ensure that dowstream deps problems
+# packages installed separately to ensure that downstream deps problems
 # are detected, c.f. #1002
 commands =
-    pip install -e acme[testing]
+    pip install -e acme
```
kuba: `pip install` lines should also be reverted, I believe

tox.ini
kuba: Please do, thanks!

tox.ini
erikrose: Good catch.

pde:
Looks good to me as long as `setup.cfg` doesn't need that trailing newline.

erikrose:
@pde It doesn't, but here it is anyway.

bmw:
Fixes #1572

LGTM && merging

Issue #1007, opened 2015-10-16
Suggest donating to Let's Encrypt after successful issuance


bdaehlie:
We should ask people to donate to Let's Encrypt after successful issuance of a certificate through our client. Maybe present 1-2 sentences and a link to where they can donate. It should be easy to disable this via command-line flag (or some other mechanism).

pde:
I think this is something we should prioritise as soon as we feel like we're free of showstoppers, which will hopefully be before GA :)

PR #1666, opened 2015-12-02
Add some suggested donation links upon success

pde:
Closes: #1007


sjerdo:
@pde shouldn't the documentation of this function be changed?

pde:
@sjerdo indeed it should :)

pde:
Merging for launch!

TomiBelan:
`If like Let's Encrypt` should probably be `If you like Let's Encrypt`.

PR #1670, opened 2015-12-03
Stop randomizing standalone challenge preferences

ranweiler:
For: letsencrypt/letsencrypt/issues/1658.

- [x] Sharpen test expectation
- [x] Remove preference shuffling
- [x] Update help text, any documentation

### DIFF
--- a/letsencrypt/plugins/standalone.py
+++ b/letsencrypt/plugins/standalone.py
@@ -197,8 +197,7 @@ class Authenticator(common.Plugin):
 
     def get_chall_pref(self, domain):
         # pylint: disable=unused-argument,missing-docstring
-        return [chall for chall in SUPPORTED_CHALLENGES
-                if chall in self.supported_challenges]
+        return self.supported_challenges
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
         if any(util.already_listening(port) for port in self._necessary_ports):
--- a/letsencrypt/plugins/standalone.py
+++ b/letsencrypt/plugins/standalone.py
@@ -198,9 +198,7 @@ class Authenticator(common.Plugin):
 
     def get_chall_pref(self, domain):
         # pylint: disable=unused-argument,missing-docstring
-        chall_pref = list(self.supported_challenges)
-        random.shuffle(chall_pref)  # 50% for each challenge
-        return chall_pref
+        return SUPPORTED_CHALLENGES
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
         if any(util.already_listening(port) for port in self._necessary_ports):
--- a/letsencrypt/plugins/standalone.py
+++ b/letsencrypt/plugins/standalone.py
@@ -173,8 +173,8 @@ class Authenticator(common.Plugin):
     @property
     def supported_challenges(self):
         """Challenges supported by this plugin."""
-        return set(challenges.Challenge.TYPES[name] for name in
-                   self.conf("supported-challenges").split(","))
+        return [challenges.Challenge.TYPES[name] for name in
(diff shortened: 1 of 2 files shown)


ranweiler:
Questions:

1. Do we actually want `tls-sni-01` to be preferred over `http-01` by default, as is implemented?
1. Do we want to try to infer a preference ordering from the user configuration? Right now I believe the only way a user can express preference for `http-01` is by omitting `tls-sni-01`.

letsencrypt/plugins/standalone.py
```diff
@@ -166,10 +165,10 @@ def __init__(self, *args, **kwargs):
...
-            "order preferences are randomly chosen.",
-            type=supported_challenges_validator, default=",".join(
-                sorted(chall.typ for chall in SUPPORTED_CHALLENGES)))
+        add("supported-challenges",
+            help="Supported challenges. Prefers tls-sni-01.",
+            type=supported_challenges_validator,
+            default=",".join(chall.typ for chall in SUPPORTED_CHALLENGES))
 
     @property
     def supported_challenges(self):
```
bmw: What if we changed this function to return a list instead of a set so that order will be preserved?

letsencrypt/plugins/standalone.py
```diff
@@ -166,10 +165,10 @@ def __init__(self, *args, **kwargs):
 
     @classmethod
     def add_parser_arguments(cls, add):
-        add("supported-challenges", help="Supported challenges, "
-            "order preferences are randomly chosen.",
-            type=supported_challenges_validator, default=",".join(
-                sorted(chall.typ for chall in SUPPORTED_CHALLENGES)))
+        add("supported-challenges",
+            help="Supported challenges. Prefers tls-sni-01.",
```
bmw: If the suggested change is made on `supported_challenges`, we could update the help to say something like "Preferred in the order they are listed."

letsencrypt/plugins/standalone.py
```diff
@@ -198,9 +197,8 @@ def prepare(self):  # pylint: disable=missing-docstring
 
     def get_chall_pref(self, domain):
         # pylint: disable=unused-argument,missing-docstring
-        chall_pref = list(self.supported_challenges)
-        random.shuffle(chall_pref)  # 50% for each challenge
-        return chall_pref
+        return [chall for chall in SUPPORTED_CHALLENGES
```
bmw: If the suggested change is made on `supported_challenges`, this function can simply be:
```
return self.supported_challenges
```

bmw:
Thank you very much for the PR @ranweiler. I was actually going to write this PR myself before public beta but I ran out of time.

This PR looks good! Really the only thing I'd like to see changed is have `supported_challenges` use a list instead of a set. All other changes I suggested are really just in response to that one change. You will also have to modify tests so that this change works.

I will be appending "[needs revision]" to the title of the PR to designate it has been reviewed. Please remove it when you'd like another one.

To answer your questions:

1. Yes. I believe we should prefer `tls-sni-01` over `http-01`. `tls-sni-01` requires port 443 to be free and `http-01` requires port 80. My reasoning is that if you're running `letsencrypt`, you're more likely to have port 443 free than port 80. This obviously isn't always true, but I expect it to be true in general.
2. Yes. I believe my comments on the code implements this.

letsencrypt/plugins/standalone.py
ranweiler: That'd be great! Due to the specificity of the existing `test_supported_challenges` test, I assumed it needed to be a set.

ranweiler:
@bmw, revisions pushed!

bmw:
LGTM! Thanks again for the contribution @ranweiler.

PR #1671, opened 2015-12-03
fix sudo issue on amazon linux instance with letsencrypt-auto

levskaya:
the letsencrypt-auto script was missing the sudo parameter on call to ExperimentalBootstrap for amazon linux. 
also added comment to mac entry to clarify why it lacks the parameter


pde:
Nice catch.

jmhodges:
Peter, this is exactly the kind of change we agreed should not go in until
after the release. Please stop.

jmhodges:
And by "release", I mean the Public Beta. This is irresponsible and unkind to your teammates.

PR #1679, opened 2015-12-03
Unbreak master

kuba:


jsha:
LGTM

jmhodges:
I saw mention of tag pushes in some PRs. Is there additional clean up we need to do around those?

PR #1681, opened 2015-12-03
Remove ! at end of url (fixes open url in gnome-terminal)

bit:
The ! at the end of the url is parsed as part of
the url if one uses "Open Link" in gnome-terminal.

### DIFF
--- a/letsencrypt/display/ops.py
+++ b/letsencrypt/display/ops.py
@@ -245,7 +245,7 @@ def success_installation(domains):
 
     """
     util(interfaces.IDisplay).notification(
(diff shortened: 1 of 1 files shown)


letsencrypt/display/ops.py
```diff
@@ -245,7 +245,7 @@ def success_installation(domains):
 
     """
     util(interfaces.IDisplay).notification(
-        "Congratulations! You have successfully enabled {0}!{1}{1}"
+        "Congratulations! You have successfully enabled {0} {1}{1}"
```
kuba: space after `{0}` should be removed, `{1}` get turned into `\n` FYI

Issue #1531, opened 2015-11-17
Augeas fails to parse RewriteRule and FilesMatch regexps (incl. Drupal .htaccess files)


kristofferwiklund:
I am trying running the following command:
./letsencrypt-auto --server https://acme-v01.api.letsencrypt.org/directory --help

And I get:
Updating letsencrypt and virtual environment dependencies.......
Running with virtualenv: sudo /home/koffe/.local/share/letsencrypt/bin/letsencrypt --apache --server https://acme-v01.api.letsencrypt.org/directory --agree-dev-preview
The apache plugin is not working; there may be problems with your existing configuration.
The error was: PluginError(('There has been an error in parsing the file (%s): %s', u'/var/aegir/platforms/20151024/.htaccess', u'Syntax error'),)

With a verbose output:
Requested authenticator apache and installer apache
Other error:(PluginEntryPoint#apache): ('There has been an error in parsing the file (%s): %s', u'/var/aegir/platforms/20151024/.htaccess', u'Syntax error')
Traceback (most recent call last):
File "/home/koffe/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/plugins/disco.py", line 103, in prepare self._initialized.prepare()
File "/home/koffe/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 152, in prepare self.check_parsing_errors("httpd.aug")
File "/home/koffe/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/augeas_configurator.py", line 68, in check_parsing_errors raise errors.PluginError(msg)
PluginError: ('There has been an error in parsing the file (%s): %s', u'/var/aegir/platforms/20151024/.htaccess', u'Syntax error')  

Here is the .htaccess file
[htaccess.txt](https://github.com/letsencrypt/letsencrypt/files/36394/htaccess.txt)

anarcat:
this is likely #1294 - i had the same issue and in there they say it was fixed in master. yet i have tried to run the dev version as explained in:

https://letsencrypt.readthedocs.org/en/latest/contributing.html#hacking

.. but i got the same error. either the bug is not fixed or the install instructions are too complicated or it's time to make a release (or all of the above).

anarcat:
yep - i tried flushing the venv and re-running the above instructions, i still get the bug:

http://paste.debian.net/333539/

```
$ git show
[...]

PR #1692, opened 2015-12-03
Merge augeas lens fix for backslashes at the start of regexps

pde:
From: hercules-team/augeas#325
Fixes: #1531


domcleal:
There's a further bug report at https://github.com/letsencrypt/letsencrypt/issues/1294#issuecomment-161805063 which I've fixed in https://github.com/hercules-team/augeas/commit/ba10b23e3561f7fbac224fe0d1b21d5047192f7b.  Not sure if you want to bring that fix into this PR or file another?

pde:
@domcleal @jdkasten it seems that your lens patch breaks our tests because we declare that [this file](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/tests/testdata/debian_apache_2_4/two_vhost_80/apache2/conf-available/bad_conf_file.conf) should error out.  In a quick test, Apache seems to run with that file, but won't server things from it even if I add a DocumentRoot?  That's kind of weird behaviour.  Do others get the same result?

pde:
In retrospect, wish I'd made another PR rather making two lense changes in this one.  In fact, I'm going to do that...

domcleal:
Yep, sorry for the confusion.

:+1: to this PR.

PR #1696, opened 2015-12-03
ownership and permission fixes to webroot plugin generated dirs/files

piru:
Fix ownership and permissions of the .well-known directory. Without this the .well-known might end up owned by root. Also both .well-known directory and the validation file would end up with wrong permissions when system umask was stricter than 022.

Without this fix and 077 umask the validation would always fail with 403 error (AH00132: file permissions deny server access)

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -72,6 +72,11 @@ to serve all files under specified web root ({0})."""
                 os.chown(self.full_roots[name], stat_path.st_uid,
                           stat_path.st_gid)
 
+                # Also fix the parent directory (.well-known) permissions
+                os.chmod(os.path.dirname(self.full_roots[name]), filemode)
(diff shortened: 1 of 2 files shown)


pde:
@lbeltrame may be able to help get this passing tests and reviewed.

lbeltrame:
Can you modify the test so the proper permissions are tested? Just need to do the same thing as your PR, but in the test code.

piru:
Right, oops. Corrected the unit tests.

I also fixed the code to remove the file X permission for 'group' and 'other' too, not just 'user'.

lbeltrame:
LGTM. @pde Your thoughts?

pde:
@bmw is looking at the three webroot permissions PRs we have, and trying to figure out which (combination) of them make sense.

pde:
Conclusion from @SwartzCr is that of the two PRs for this bug we should take #1807 for the imminent 0.1.1 release rather than this PR, so closing here.

PR #1702, opened 2015-12-03
Add another apache conf test case

pde:


domcleal:
You can apply https://github.com/hercules-team/augeas/commit/f44a7a55cc7162beced99659234eb078a8d20e1d#diff-8f4883422ccaa870b2414a76707a6641 to the lens which should make it parse.

Issue #1473, opened 2015-11-12
archive/domain/privkeyN.pem is set to 0644 instead of 0600 (or 0440)


ZaiLynch:
The private keys in /etc/letsencrypt/archive/_domain_/privkey<N>.pem are currently set to 0644. 
I guess it's better to have only root to be able to read the private key, hence set them to 0600.

pde:
Hrm. What permissions do you see on the directory `/etc/letsencrypt/keys`?  Can you actually read a privkey as an unprivileged user?

pde:
Probably the keys should be 0640 and set the group that runs the webserver, if it isn't running as root?  And the directory should be 0750, assuming such a group exists?

pde:
This is going to be hard to get right from `letsencrypt-auto`.  But we should have control of these variables concisely exported for OS packagers.  @hlieberman @fmarier.

pde:
This is part of #420.

ZaiLynch:
Hm, you're right.
/etc/letsencrypt/archive is set to 700, so a regular user shouldn't be able to reach the keys...

pde:
One point made on community.letsencrypt.org is that users may move the certs out of /etc/letsencrypt.  Of course they should be symlinking in rather than moving / copying them out, but it will happen.  So bumping this ticket up to **nice for 1.0**.

pde:
Actually, the point about not moving keys makes me think that in fact the permissions on everything in the directory should be 0440, not 0640.

lbeltrame:
A related matter would be to honor the group permissions set by the directory the certs reside in. I use a "ssl" group that's the only one allowed to touch the certs aside root, for example.

thomaszbz:
For symlinks from /etc/ssl/..., something like this would be nice:

    -rw-r--r-- 1 root root     2.1K Jan 17 17:54 cert1.pem
    -rw-r--r-- 1 root root     1.1K Jan 17 17:54 chain1.pem
    -rw-r--r-- 1 root root     3.2K Jan 17 17:54 fullchain1.pem
    -rw-r----- 1 root ssl-cert 3.2K Jan 17 17:54 privkey1.pem

These settings should be fine for e.g. Debian 8 / Apache.

I guess that users have different requirements (e.g. according to distro), so I think the permissions should be configurable via ``cli.ini`` and maybe via parameters (overwriting ``cli.ini`` settings).

[...]

PR #1705, opened 2015-12-04
Set permissions on private keys. [needs revision]

daztucker:


schoen:
Thanks for looking into this issue -- I also saw on the forum that it was a topic of concern for people because private keys are indeed among the most sensitive files that may exist on the server.

@hlieberman, @pde suggested that you might have an idea about the Debian policy aspects of this (or already know of a plan for what our file permissions should be). In particular I'm wondering whether there are many systems where private keys ought to be group-readable, not just owner-readable, because the web server is running as e.g. user `httpd` group `httpd` (or `www`), while the `privkey.pem` might be owned by root. In that case it ought to be group-readable to allow the daemon to read it at startup.

@pde also suggested `440` instead of either `660` or `600` as a default (because there's normally no need to write to the privkey and because it would be helpful to warn people who are trying to delete it!) -- and finding a way to make the file modes configurable in addition to changing the default. (I don't know if there's work toward that in another PR yet.)

schoen:
Tracked at #1473 which also shows @colincross is working on this problem too.

hlieberman:
Debian policy doesn't specifically call out what the proper modes are for keys, if that's what you mean.

That being said, there's an example which we can probably steal wholesale, if not reuse: the snakeoil cert.

ssl-cert creates /etc/ssl/private, which is root:ssl-cert and 710, and the snakeoil cert inside is root:ssl-cert, 640.  That seems a reasonable choice to me to allow people to have daemons be able to read the certificates and keys.

Another option would be, for the nginx and apache plugins, to simply use whatever user that the thing is going to be running as, since we are parsing the config files anyway.  Not sure if I like that option more or less.

colincross:
Sorry for the noise, I didn't realize github would put a note on the issue when I pushed to my local fork.  I'll put the issue in the pull request instead of the commit message next time.

Sendmail requires private keys to be at least 0600, which is what got me started on this:
STARTTLS=server: file /etc/letsencrypt/live/<domain>/privkey.pem unsafe: Group readable file

schoen:
If someone else reviews this, note also the parallel PR #1760 for the same purpose.

thomaszbz:
For symlinks from /etc/ssl/..., something like this would be nice:

    -rw-r--r-- 1 root root     2.1K Jan 17 17:54 cert1.pem
    -rw-r--r-- 1 root root     1.1K Jan 17 17:54 chain1.pem
    -rw-r--r-- 1 root root     3.2K Jan 17 17:54 fullchain1.pem
    -rw-r----- 1 root ssl-cert 3.2K Jan 17 17:54 privkey1.pem

These settings should be fine for e.g. Debian 8 / Apache.

I guess that users have different requirements (e.g. according to distro), so I think the permissions should be configurable via ``cli.ini`` and maybe via parameters (overwriting cli.ini settings).

I also want to note that wrong permissions (in the sense of user's requirements) cannot necessarily be healed by a later user interaction: Once the private key is readable after being written to disk with weak permissions, it could theoretically be stolen if the admin user changes permissions too late. Therefore, an _emtpy_ file with correct file permissions must be written to disk _before_ key material is written into the file.

I suggest to introduce three config values for cli.ini (including example settings):

    private-key-file-owner = root
    private-key-file-group = ssl-cert
    private-key-file-permisson = 640

Optionally, this could also be done for public key material (just for completeness).

jsha:
This PR and #1760 both block on a design decision (#1473) about how exactly we want configuration of ownership and permissions to work. Closing for now; we can make or solicit a new PR once we've reached a conclusion on #1473.

Issue #1713, opened 2015-12-04
Use of third-party domains in examples considered harmful


rwetzlmayr:
`README.rst` contains a few example shell commands using third-party domains as examples, e.g. 

> `./letsencrypt-auto certonly --standalone --email admin@thing.com -d thing.com -d www.thing.com -d otherthing.net`

I think that using [example.com](https://en.wikipedia.org/wiki/Example.com) and its siblings would be equally educative while it would not force the real owner of thing.com to deal with unsolicited mail arriving at `admin@thing.com` as people copy/paste this line into their shell.

PR #1716, opened 2015-12-04
 Use reserved example domains in examples of shell commands. [needs revision]

rwetzlmayr:
Refs #1713.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,11 +51,11 @@ client will guide you through the process of obtaining and installing certs
 interactively.
 
 You can also tell it exactly what you want it to do from the command line.
-For instance, if you want to obtain a cert for ``thing.com``,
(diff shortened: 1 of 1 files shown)


pde:
One downside of this is that it becomes less immediately clear to the reader that these are different domains.  It's a pity the RFCs don't seem to include any reserved examples that look very different from each other (there are some for IDNs but that's not really helpful).

chriscroome:
Use some sub-domains, `thing.example.com`, `otherthing.example.com`, `yetanotherthing.example.net`?

schoen:
I think @chriscroome's suggestion would make this more useful to the reader because it addresses @pde's objection.

bmw:
fixed in #2452

PR #1717, opened 2015-12-04
Check for pip-installed virtualenv (ie. not pyenv shims)

impressiver:
I ran into a small hiccup when trying to run `letsencrypt-auto` on OS X. Because `pyenv-virtualenv` installs shims for the real `virtualenv`, running `hash virtualenv` will always succeed, even if it's not really installed for the currently active pyenv (`system`, in this case).

This changes the `mac.sh` script to ensure `virtualenv` is a pip-installed package. I also added a check to only install brew packages that aren't already installed (`augeas`, `dialog`).

### DIFF
--- a/bootstrap/mac.sh
+++ b/bootstrap/mac.sh
@@ -4,15 +4,15 @@ if ! hash brew 2>/dev/null; then
     ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 fi
 
-brew install augeas
-brew install dialog
+brew --prefix augeas &>/dev/null || brew install augeas
+brew --prefix dialog &>/dev/null || brew install dialog
 
 if ! hash pip 2>/dev/null; then
     echo "pip Not Installed\nInstalling python from Homebrew..."
     brew install python
 fi
 
-if ! hash virtualenv 2>/dev/null; then
+if ! pip list | grep 'virtualenv\s' &>/dev/null then
     echo "virtualenv Not Installed\nInstalling with pip"
     pip install virtualenv
 fi


bmw:
@TheNavigat, would you like to take a look at this?

TheNavigat:
IMHO the proposed code makes a lot of sense, but I've had difficulty reproducing the error because of the fact that pyenv itself is quite unstable. Running something like `pyenv shell` prints `pyenv: no such command `shell'`.

On the other hand, the code makes sense because checking for the installation of virtualenv properly is something I'd favour. However, as @bmw mentioned on the IRC, `pip list` isn't always trustable.

impressiver:
As far a second I know, `pip list` has been available since pip 1.2.1.post1. The only other way I can think of where `pip list` wouldn't be a reliable way to test for virtualenv is if virtualenv was installed manually somewhere else in the path. Even if it were, installing again with pip *shouldn't* break anything.

Further, the script already assumes pip is the defacto for installing virtualenv (if it isn't already), so it doesn't seem outrageous to use pip to see if it's already installed.

bmw:
Yeah. My concern about using `pip list` is due to the possibility that `virtualenv` is in your path and `pip` doesn't find it. While I haven't tested this on OS X, this situation is very easy to recreate on linux systems as often packages installed with the OS package manager aren't found by `pip`. We'd like to avoid installing another version of `pip` if possible.

jsha:
@impressiver perhaps you could test for the existence of virtualenv on the path by running

`if ! virtualenv --version &>/dev/null ; then`

Does `pyenv-virtualenv`'s shim return successfully if you run it like that, or does it exit with a nonzero status code?

bmw:
This PR hasn't seen any progress in a while and we want to rethink how we're supporting macOS in `certbot-auto`. Closing this for now until more progress can be made.

PR #1735, opened 2015-12-04
fix(setup.py): require mock only for tests

asfaltboy:
This may just work or require more code cleanup, lets see what travis says.


kuba:
I'm afraid that do want `mock` to be part of `install_requires` - keeping tests alongside the code is an intentional decision and hence we also need a dep to run them. That's a no-go for me.

asfaltboy:
roger that

PR #1751, opened 2015-12-04
Various shelly fixes to letsencrypt-auto

Arthur2e5:
Fixed:
* weird logic in DeterminePythonVersion.

Other changes:
* Let's just use `case` for option parsing. Here `-v*` should be a good-enough approximation.
  * P.S. original regex is just a regex -- `-E` looks unnecessary.
    * Do you mean `^-v+$`? It looks weird to take something like `dont-be-verbose` as an option switch.

See the commit messages for more info.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -87,17 +87,13 @@ ExperimentalBootstrap() {
 }
 
 DeterminePythonVersion() {
-  if command -v python2.7 > /dev/null ; then
-    export LE_PYTHON=${LE_PYTHON:-python2.7}
-  elif command -v python27 > /dev/null ; then
-    export LE_PYTHON=${LE_PYTHON:-python27}
-  elif command -v python2 > /dev/null ; then
-    export LE_PYTHON=${LE_PYTHON:-python2}
-  elif command -v python > /dev/null ; then
-    export LE_PYTHON=${LE_PYTHON:-python}
-  else
+  for LE_PYTHON in "$LE_PYTHON" python2.7 python27 python2 python; do
+    # break (while keeping the LE_PYTHON value) if found
+    command -v "$LE_PYTHON" > /dev/null && break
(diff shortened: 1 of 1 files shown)


Arthur2e5:
Mixed something up in the second commit msg. fixed.

Maybe someday I will also make all those error reports `>&2`'d.

pde:
We should definitely take the regexp fix before we add any --flags-that-are-very-strange.

The stylistic improvements would probably be deferred for the time being (don't touch it if it's not broken) but @erikrose feel free to take them for your branch, which will need a lot of testing energy put into it anyway.

Arthur2e5:
Feel free to cherry-pick the commits then.

Well, actually both can be considered as somehow logical fixes, since:

* DeterminePythonVersion doesn't change nor verify a pre-existing `$LE_PYTHON` value after it found a working python. It exports the same value.
* The regex accpets weird values just like what I have said in the 0th issue comment.

P.S. It's possible to match the exact opposite of `^-v+$` using the pattern `[!-]*|*[!v]*|-` (anything that does not start with a `-` or contains anything that's not a `v` or just a `-`). This slightly clumsy and obfuscated segment does the match:
```sh
case "$arg" in  # (
  --verbose)
    VERBOSE=1;; # (
  --debug)
    DEBUG=1;;   # (
  [!-]*|*[!v]*|-) # anything that's not what we want
    ;; # do nothing (
  *) # Now the requirement of at-least-one-v is also back.
    VERBOSE=1;;
esac
```

erikrose:
> It's possible to match the exact opposite
That's really clever and seems to work perfectly. I'll take it. :-)

erikrose:
Ported these to the new le-auto in #2466. Can I hit you up for a review, @Arthur2e5?

Arthur2e5:
Sure.

PR #1758, opened 2015-12-05
lens update from augeas upstream

pde:
From: https://github.com/hercules-team/augeas/pull/329
 Fixes: https://github.com/letsencrypt/letsencrypt/issues/1294#issuecomment-161805063

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/augeas_lens/httpd.aug
+++ b/letsencrypt-apache/letsencrypt_apache/augeas_lens/httpd.aug
@@ -51,7 +51,7 @@ let sep_osp             = Sep.opt_space
 let sep_eq              = del /[ \t]*=[ \t]*/ "="
 
 let nmtoken             = /[a-zA-Z:_][a-zA-Z0-9:_.-]*/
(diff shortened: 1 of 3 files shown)


pde:
@domcleal @jdkasten it seems that your lens patch breaks our tests because we declare that [this file](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/tests/testdata/debian_apache_2_4/two_vhost_80/apache2/conf-available/bad_conf_file.conf) should error out.  In a quick test, Apache seems to run with that file, but won't server things from it even if I add a DocumentRoot?  That's kind of weird behaviour.  Do others get the same result?

domcleal:
The file works for me under EL7's httpd-2.4.6, and trying to change the case of other section tags in Apache configs seems to make no difference, it accepts them.

It might just need a different test to keep aecb7b71 working, perhaps a missing end tag?

bmw:
It took a bit of work, but I was able to get this file to work with Apache 2.4.10 on Debian 8. Made the modification suggested by @domcleal to fix the test.

pde:
Merging!

PR #1759, opened 2015-12-05
Added a descriptive error if domain list includes a Unicode-encoded IDN

rainest:
The current error for IDNs passed in Unicode form is incorrect and does not describe the actual problem:

```
$ letsencrypt --manual -d example.com -d ёжикв.сайт
Requested domain is not a FQDN
```

This change checks for any domains that cannot be encoded as ASCII, and if one is present:

```
$ letsencrypt --manual -d example.com -d ёжикв.сайт
Internationalized domain names are not supported
```

### DIFF
--- a/letsencrypt/configuration.py
+++ b/letsencrypt/configuration.py
@@ -148,7 +148,7 @@ def _check_config_domain_sanity(domains):
     # Unicode
     try:
         for domain in domains:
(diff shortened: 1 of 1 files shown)


schoen:
Thanks, that's a useful error message that will help users better understand what's going on.

Issue #1504, opened 2015-11-14
Run script on renewal of certificate


bwesterb:
The preferred way to use letsencrypt certificates, is pointing applications directly into `/etc/letsencrypt/live`.  However, for several reasons (see below), this might not work and some custom steps need to be taken to install a certificate for an application.  An admin should be able to install a script which is run on the renewal of a certificate to install it.

A hypothetical implementation could work as follows

    letsencrypt -d mydomain --on-renewal=/usr/local/bin/install-cert-for-lighttpd

With the script `/usr/local/bin/install-cert-for-lighttpd` containing, for instance

    cat "/etc/letsencrypt/live/$1/privkey.pem" "/etc/letsencrypt/live/$1/cert.pem" > /etc/lighttpd/ssl/$1.pem
    chown www-data:www-data /etc/lighttpd/ssl/$1.pem
    service lighttpd reload

Some reasons I can think of for a script are:

1. The applications requires that the certificate and key in a modified format.   See  #1201
2. The application does not run as root.  Thus, we need to make a copy of the certificate only readable by the user as which the application is running.  See e.g. #1425

PatrickHeppler:
Why not just putting the letsencrypt call into install-cert-for-lighttpd?
I also wrote a python script, which stores the current hpkp pin, runs letsencrypt, retrieves the new hpkp pin and write both to a config file. It also hits my dns api to update TLSA records.

NEOatNHNG:
@PatrickHeppler Probably because the install-cert-for-lighttpd script has no idea when the renewal will happen.

@bwesterb I think this is rather inelegant. The main selling point of Let's Encrypt is to make the life of admins easier. One way it does that is by automating the installation process for standard software. So there should not be a script that you have to give on the command line to make renewal work on standard software. #1425 is easily solvable by setting the group permissions of the certificates right – no script needed. For the automation on cert renewal you are probably right: We need some kind of hook that gets executed. But that should be a module in the letsencrypt client, not a script you have to write yourself and it should take into account that multiple applications use the certificate and therefore need to have their hooks executed. As the format is needed for multiple pieces of software it should be either generated unconditionally (which would make custom configurations for software where we do not have an installer module easier) or it should use a dependency based mechanism to generate the formats that are required by an installer module.

PatrickHeppler:
@NEOatNHNG That's why I wrote he should put the letsencrypt call into his script ;)

bwesterb:
[...]

PR #1760, opened 2015-12-05
Set permissions of privkey.pem to 0600

colincross:
Sendmail requires privkey.pem to be unreadable by anything except
root.  Create privkey.pem as 0600, and make the permissions of the
certificate files explictly 0644.

Writing through the live symlink to the archive directory prevents
using O_EXCL | O_CREAT, so write directly to the archive instead.

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -624,40 +624,41 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
         # lineagename will now potentially be modified based on which
         # renewal configuration file could actually be created
         lineagename = os.path.basename(config_filename)[:-len(".conf")]
-        archive = os.path.join(cli_config.archive_dir, lineagename)
+        archive_dir = os.path.join(cli_config.archive_dir, lineagename)
         live_dir = os.path.join(cli_config.live_dir, lineagename)
-        if os.path.exists(archive):
+        if os.path.exists(archive_dir):
             raise errors.CertStorageError(
                 "archive directory exists for " + lineagename)
         if os.path.exists(live_dir):
             raise errors.CertStorageError(
                 "live directory exists for " + lineagename)
-        os.mkdir(archive)
+        os.mkdir(archive_dir)
         os.mkdir(live_dir)
         logger.debug("Archive directory %s and live "
-                     "directory %s created.", archive, live_dir)
-        relative_archive = os.path.join("..", "..", "archive", lineagename)
+                     "directory %s created.", archive_dir, live_dir)
+        relative_archive_dir = os.path.join("..", "..", "archive", lineagename)
 
         # Put the data into the appropriate files on disk
         target = dict([(kind, os.path.join(live_dir, kind + ".pem"))
                        for kind in ALL_FOUR])
+        archive = dict([(kind, os.path.join(archive_dir, kind + "1.pem"))
+                       for kind in ALL_FOUR])
         for kind in ALL_FOUR:
-            os.symlink(os.path.join(relative_archive, kind + "1.pem"),
+            os.symlink(os.path.join(relative_archive_dir, kind + "1.pem"),
                        target[kind])
-        with open(target["cert"], "w") as f:
-            logger.debug("Writing certificate to %s.", target["cert"])
+        with le_util.safe_open(archive["cert"], "w", chmod=0o644) as f:
+            logger.debug("Writing certificate to %s.", archive["cert"])
             f.write(cert)
-        with open(target["privkey"], "w") as f:
-            logger.debug("Writing private key to %s.", target["privkey"])
+        with le_util.safe_open(archive["privkey"], "w", chmod=0o600) as f:
+            logger.debug("Writing private key to %s.", archive["privkey"])
             f.write(privkey)
-            # XXX: Let's make sure to get the file permissions right here
-        with open(target["chain"], "w") as f:
-            logger.debug("Writing chain to %s.", target["chain"])
+        with le_util.safe_open(archive["chain"], "w", chmod=0o644) as f:
+            logger.debug("Writing chain to %s.", archive["chain"])
             f.write(chain)
-        with open(target["fullchain"], "w") as f:
+        with le_util.safe_open(archive["fullchain"], "w", chmod=0o644) as f:
             # assumes that OpenSSL.crypto.dump_certificate includes
             # ending newline character
-            logger.debug("Writing full chain to %s.", target["fullchain"])
+            logger.debug("Writing full chain to %s.", archive["fullchain"])
             f.write(cert + chain)
 
         # Document what we've done in a new renewal config file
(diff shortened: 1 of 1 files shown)


colincross:
Fixes #1473

olivluca:
That would break, e.g., cyrus-imapd, since it doesn't run as root (note that it doesn't work by default either unless I manually chgrp the certificates to mail and change the permission/group of the containing directories).
Maybe the owner/group/permissions shouldn't be hardcoded but configurable?

kousu:
The same problem breaks opensmtpd:
```
kousu@comms:~$ sudo /usr/sbin/smtpd -d
warn:  /etc/letsencrypt/live/comms.kousu.ca/privkey.pem: insecure permissions: must be at most rwx------ 
fatal: load_pki_tree: failed to load key file
```

There's something else strange here: the symlink's permissions are 777, even though the actual file isn't. OpenSMTPd is checking permissions on the symlink, not the actual file. The permissions of the symlinks should match the permissions of the files!

But that doesn't solve this problem in general. Some programs are going to be more or less picky about permissions than others, and I suspect some will flat out contradict (e.g. if a program assumes the certs are owned by its user, or difference of opinion about if certs should be read pre or post priv-dropping). #1504 argues that each specific piece of software should get a different adaptor script. I don't know what's best.

Incidentally, changing the symlink to be *more* restrictive than the actual file fixes OpenSMTPd:
```
root@comms:/etc/letsencrypt/live/comms.kousu.ca# chmod go-rwx privkey.pem 
root@comms:/etc/letsencrypt/live/comms.kousu.ca# smtpd -d
info: OpenSMTPD 5.4.2p1 starting
info: startup
```

schoen:
It seems we will want it to be configurable, but it's still not clear what the best default is. (When using a specific installer, the installer could maybe have a default that's appropriate for the software that the installer is integrating with, but we're not always using a specific installer because people also want certs for external software that we don't directly integrate with or that they don't want us to configure for them.)

schoen:
If someone else reviews this, note also the parallel PR #1705 for the same purpose.

convissor:
I'm surprised by [Cyrus' behavior](http://cyrusimap.org/docs/cyrus-imapd/2.5.7/install-configure.php#openssl).  Most other daemons (eg postfix, apache) start up as root to get the keys, then drop to a non-privileged user.

kousu:
:+1: to avoiding knobs in letsencrypt. Daemons are not consistent and that is hard, but the more configuration in core letsencrypt the more chance for something to go wrong.

Maybe there should be a letsencrypt group (`letsencrypt_private_key`?) that daemon users can be placed in. Like, set the privkeys `0440` `root:letsencrypt_private_key`. Daemons that drop from root will keep working, other ones will need to have the install script check that their user is in that group.

thomaszbz:
Using Debian 8, there's a group ``ssl-cert`` for use with private keys. Maybe LE could go for this (or equivalent) as a default? Plus, configurable permissions would be nice (via ``cli.ini``). Also see my comment at #1473.

Zash:
Configurable permissions and ownership for certificates and files would be nice.

Another example: [Prosody](https://prosody.im) drops root long before it even starts. Even a hidden option somewhere to chown/chmod certificates and keys meant for it to a different user would make using Let's Encrypt with it much easier.

kousu:
I just renewed my certs and this issue bit me again. I cannot set up auto-renew because opensmtpd is prissy: I *must* go in and manually `chmod g-o` the keyfile or else it refuses to run, since it is designed to run briefly as root and then drop privs. But that `chmod` breaks other daemons that assume they will be run as the user they are supposed to be. I don't know what to do. OpenSMTPD is doing the more secure thing---certainly we won't convince them to change---by forcing the more restrictive permissions u+r go-r, but daemons that don't do priv-dropping instead require g+r and group configuration.

I see that since I last checked in, `letsencrypt` now sets `privkey$N.pem` to 0640 (rw-r-----), supporting the latter case.  I like that, I think, but I also like opensmtpd.

I could use hard-links to effectively lie to smtpd about the permissions:
```
$ N=3  #which edition this cert is
$ cd /etc/letsencrypt/archive/comms.kousu.ca/
$ sudo ln privkey$N.pem privkey$N_.pem
$ sudo chmod 0600 privkey$N_.pem 
```

and arrange for smtpd to use the stable address of `/etc/letsencrypt/live/comms.kousu.ca/privkey_.pem` 
```
$ cd /etc/letsencrypt/archive/comms.kousu.ca/
$ sudo ln -s ../../archive/comms.kousu.ca/privkey$N_.pem privkey_.pem
```
but this requires me to redo it every time I renew.
Does anyone have any better ideas?

kousu:
Oh wait scratch that I don't know how hardlinks work: permissions are on the inode, not the directory entry, so hardlinks share permissions. Which I guess makes sense.

So *instead* on `ln` I need to do this:
```
$ sudo cp privkey${N}.pem privkey${N}.paranoid.pem
$ sudo chmod g-r *paranoid.pem
```

FelixSchwarz:
@convissor: As noted in this thread there are many daemons not reading key material with "root". Exim is another notable example. So we need a solution for them.

@kousu: While less knobs in letsencrypt are generally a good idea I think we need some here. The ability to screw up can be mitigated by a safe default which is "mode=0600" with uid/gid root/root. Hence most people will be safe by default and admins can opt in to set an unsafe default.

FelixSchwarz:
> but this requires me to redo it every time I renew and for each domain vhosted.
> Does anyone have any better ideas?

@kousu: renew hooks were merged today (3d260485cd833571a2e8d4ac1baaa3dd228a292c). This might be best workaround to set permissions/ownership short of a specialized config option (which I still prefer, seems to be harder to mess up on the admin side + doesn't require admins to write the same logic over and over again).

LukeShu:
Setting the permissions as a post hook is a terrible, broken idea; that means that the permissions are temporarily too permissive.

A better workaround would be to just set the umask to something more restrictive.

blendmaster:
Friendly bump. Also just got burned by opensmtpd on this:

>warn:  /etc/letsencrypt/live/example.com/privkey.pem: insecure permissions: must be at most rwxr-----

LukeShu:
FWIW, I think that the best solution is just to configure the process' umask appropriately.

If running certbot from a systemd service, stick the line `UMask=0077` into the .service file, which would make the permissions be `0600`.

At the end of the day, setting the process' umask is the standard, configurable, way to control the permissions files created by a process.  It's not a hack, it's the standard Unix administration way of doing it.

Meanwhile, if it set the permissions to 0600 explicitly in the source, then I would have to resort to hacks to get a nice 0640 permission allowing services (nginx, smtpd, et c.) never need root, but just to be a member of the special cert-reading group.

On the other hand, having them be 0644 by default is antagonistic to certbot's goal of being as close to magical and zero-setup as possible, as 0644 is pretty wrong.  Perhaps the best thing to do is to set the umask explicitly at program startup, and add a --umask flag that changes the umask away from the default 0077.

kousu:
:+1: to @LukeShu

blendmaster:
Re: umask on the process, won't that affect the operation of `--webroot`, which presumably has to write files readable by the web server? I'll test it out next time I get a chance.

I think I agree in principle that umask 0077 is a good default, but considering all the automagical stuff that `certbot-auto` does inside of its black box (like webroot), a very specific config flag for permissions on privkey.pem seems like the most immediate way to hack around the "services don't agree on the appropriate privkey permissions" problem. Certainly a little depressing, but not all programs can be elegant paragons of the Unix Philosophy™.

LukeShu:
@blendmaster You are right, 0077 might interfere with webroot.

On the other hand, you could use FACLs on the certs directory to set a default permission on the directory itself (essentially setting a per-directory umask, instead of a per-process umask).  But yes, a flag is probably the best option.

blendmaster:
>On the other hand, you could use FACLs on the certs directory to set a default permission on the directory itself (essentially setting a per-directory umask, instead of a per-process umask). But yes, a flag is probably the best option.

Thanks for this suggestion. I did:

    setfacl -d -m 'group: :0' -m 'other: :0' /etc/letsencrypt/archive/example.com

and renewed certs do get created at 600. The public key and chain are unfortunately also only root-accessible now, but at least for me, that's fine.

jsha:
This PR and #1705 both block on a design decision (#1473) about how exactly we want configuration of ownership and permissions to work. Closing for now; we can make or solicit a new PR once we've reached a conclusion on #1473.

PR #1768, opened 2015-12-05
Allow to symlink the letsencrypt-auto script [needs revision]

sttts:
As a user I want to checkout the git repository and symlink the letsencrypt-auto
script into some directory in the path like /usr/local/bin.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -16,7 +16,7 @@ VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
 VENV_BIN=${VENV_PATH}/bin
 # The path to the letsencrypt-auto script.  Everything that uses these might
 # at some point be inlined...
-LEA_PATH=`dirname "$0"`
+LEA_PATH="$(LEA="$0"; while test -L "$LEA"; do LEA=$(readlink "$LEA"); done; echo $(dirname "$LEA"))"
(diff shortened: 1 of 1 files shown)


pde:
Thanks for the patch @stts!

At this point in the script we aren't strictly guaranteed to have python available; it may be pulled in by the boostrapping process at some points; we should do this with just the POSIX shell, or do it after bootstrapping.

Secondly, an operation like this is a potential source of breakage, so my instinct was to leave LEA_PATH as it is unless `test -L $LEA_PATH` returns true.

stts:
NAVER - http://www.naver.com/
--------------------------------------------

keliong@naver.com 님께 보내신 메일 <Re: [letsencrypt] Allow to symlink the letsencrypt-auto script [needs revision] (#1768)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

sttts:
Usually you use "readlink -f" for that, but that doesn't work on OSX due to BSD userland. The `test -L` might work.

sttts:
@pde have replaced the python with some portable bash code.

erikrose:
With the new letsencrypt-auto, landed in #1665, you no longer need a git checkout, so you can place the actual script anywhere you like. Cheers!

PR #1773, opened 2015-12-05
Clarify error messages with acronym DV

### LINKED ISSUE #1772: Error messages use confusing acronym DV
Two error messages use the acronym "DV" to mean "Domain Verification" which may be confusing to the user (it was to me).

gene1wood:
Fixes #1772

### DIFF
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -22,12 +22,14 @@ class Error(jose.JSONObjectWithFields, errors.Error):
         ('urn:acme:error:' + name, description) for name, description in (
             ('badCSR', 'The CSR is unacceptable (e.g., due to a short key)'),
             ('badNonce', 'The client sent an unacceptable anti-replay nonce'),
-            ('connection', 'The server could not connect to the client for DV'),
+            ('connection', 'The server could not connect to the client to '
+                'verify the domain'),
             ('dnssec', 'The server could not validate a DNSSEC signed domain'),
             ('malformed', 'The request message was malformed'),
             ('rateLimited', 'There were too many requests of a given type'),
             ('serverInternal', 'The server experienced an internal error'),
-            ('tls', 'The server experienced a TLS error during DV'),
+            ('tls', 'The server experienced a TLS error during domain '
+                'verification'),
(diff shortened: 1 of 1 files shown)


acme/acme/messages.py
```diff
@@ -22,12 +22,14 @@ class Error(jose.JSONObjectWithFields, errors.Error):
...
-            ('connection', 'The server could not connect to the client for DV'),
+            ('connection', 'The server could not connect to the client to '
+                'verify the domain'),
             ('dnssec', 'The server could not validate a DNSSEC signed domain'),
             ('malformed', 'The request message was malformed'),
             ('rateLimited', 'There were too many requests of a given type'),
             ('serverInternal', 'The server experienced an internal error'),
-            ('tls', 'The server experienced a TLS error during DV'),
+            ('tls', 'The server experienced a TLS error during domain '
+                'verification'),
```
pde: Normally DV is expanded as "domain validation", though both are used and "domain verification" actually seems slightly more self-explanatory.

kuba:
I don't think `acme` library is the right place to edit messages for better UX. Those messages were copied verbatim from the ACME spec (https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#errors) and they should IMO stay this way, e.g. to reduce the maintenance burden. If it needs overwriting, please do so in the core package, or update the spec. Thanks!

acme/acme/messages.py
```diff
@@ -22,12 +22,14 @@ class Error(jose.JSONObjectWithFields, errors.Error):
         ('urn:acme:error:' + name, description) for name, description in (
             ('badCSR', 'The CSR is unacceptable (e.g., due to a short key)'),
             ('badNonce', 'The client sent an unacceptable anti-replay nonce'),
-            ('connection', 'The server could not connect to the client for DV'),
+            ('connection', 'The server could not connect to the client to '
+                'verify the domain'),
```
kuba: indentation is broken here and I wonder why `pylint` didn't complain...

which error check was disabled this time? :(

kuba:
Revert done in #1785.

PR #1780, opened 2015-12-05
[Test] Use demandimport to load modules on demand

bwesterb:
Uses Mercurial's on-demand module loading to speed up start-up times, which should make letsencrypt more snappy.  This PR isn't ready to be merged yet: it's here to let Travis have a go at it.

### DIFF
--- a/letsencrypt/_demandimport.py
+++ b/letsencrypt/_demandimport.py
@@ -4,7 +4,9 @@
     might cause problems as some `ImportError's aren't raised immediately. """
 
 import demandimport
-demandimport.enable()
 
-demandimport.ignore('PyICU')      # parsedatetime
-demandimport.ignore('simplejson') # requests.compat
+def enable():
+    demandimport.enable()
+
+    demandimport.ignore('PyICU')      # parsedatetime
+    demandimport.ignore('simplejson') # requests.compat
--- a/letsencrypt/_demandimport.py
+++ b/letsencrypt/_demandimport.py
@@ -6,4 +6,5 @@
 import demandimport
 demandimport.enable()
 
-demandimport.ignore('PyICU')
+demandimport.ignore('PyICU')      # parsedatetime
+demandimport.ignore('simplejson') # requests.compat
--- a/letsencrypt/_demandimport.py
+++ b/letsencrypt/_demandimport.py
@@ -1,0 +1,9 @@
+""" Use demandimport to postpone actually loading a module until it is
+    first used.  This will decrease start-up time a lot, as most imported
+    modules are not used at all --- however, some third-party modules
+    might cause problems as some `ImportError's aren't raised immediately. """
+
+import demandimport
+demandimport.enable()
+
+demandimport.ignore('PyICU')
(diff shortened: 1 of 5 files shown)


pde:
I feel that the reason letsencrypt-auto is slow has much to do with the way it updates venv too enthusiastically (every time it's run), and not much to do with module load times -- everything is pretty fast if you run via the [developer instructions](https://letsencrypt.readthedocs.org/en/latest/contributing.html#running-a-local-copy-of-the-client), or use a packaged client.  So I fear that this PR might actually hurt us, performance wise, by adding one more module for pip to fetch and update.

I believe that @erikrose has started trying to improve this in his work in #1665, though I haven't yet gotten it to run to do some latency measurements.

pde:
(But I'm open to data showing that in fact this makes things faster)

bwesterb:
I didn't intend to directly speed  up `letsencrypt-auto` --- but rather `letsencrypt` itself.  As you say, it's quite quick already, but it can be faster.

For me, the wall-times for the following non-interactive commands are:

* `letsencrypt --version`
 * without `demandimport`: 0.27
 * with `demandimport`:  0.11
* `letsencrypt certonly -d notfqdn`
 * without `demandimport`: 0.28
 * with `demandimport`: 0.12
* `letsencrypt --help`
 * without `demandimport`: 0.27
 * with `demandimport`:  0.11

For these commands that immediately return, the change is quite noticeable.

I also ran `letsencrypt` to request certificates and the wall-times were quite high and variable because of the communication with the server.  I list here the user-space CPU time:

* `letsencrypt certainly -d properdomain -a webroot --webroot-path /my/webroot`
 * without `demandimport`: 0.42
 * with `demandimport`: 0.28

So here, with the high wall-time of 4 seconds, you don't really notice the difference.

So the advantage lies with commands that exit immediately (without communicating with a server) (or go into a prompt immediately, although I did not measure this).

The disadvantage is that it may cause surprising bugs for someone who is not aware of `demandimport` running.  For instance, if a third-party module uses something like

```python
try:
    import somemodule as somepurpose
except ImportError:
    import someothermodule as somepurpose
```

Then even if `somemodule` does not exist, it will still assume it does as the `ImportError` isn't raised immediately.  Later on, on first use of `somepurpose` the `ImportError` will be raised.  The solution is to add `demandimport.ignore('some module')` after `demandimport.enable()`.

erikrose:
As @pde says, the new letsencrypt-auto will certainly fix the overzealous venv recreation. But that's another subject.

On this subject, it's nifty that we can get such a speed gain for our quicker commands, but I do fear the potential development cost in debugging confusion. (My own vague ideas toward a lighter-weight LE involve evicting the Zope Component Architecture, which I don't think we're making a lot of use of beyond a few registered utilities. It's no accident that we (me being an old Zope guy) pioneered the use of Python lazy imports; those modules are huge. ;-))

pde:
I don't know if we have any `try: import ; except: fallback` paths, but we probably *should* have some because we have too many dependencies right now :(

It sounds like we're coming down with this as a "won't merge for now", though once we're more widely packaged and `letsencrypt-auto` is no longer our main distribution channel, we can consider adding this with a `suggests: python-demandimport` in OS packages.

PR #1784, opened 2015-12-06
Mention the --renew-by-default flag

devvmh:
I was going crazy looking for this flag - I think it's worth a mention in the Renewal section


pde:
Yes.

PR #1785, opened 2015-12-06
Keep error messages in sync with ACME spec

kuba:
This reverts #1773 and should probably be followed with re-opening #1772 (and better fix). Reasons mentioned in https://github.com/letsencrypt/letsencrypt/pull/1773#issuecomment-162296740, commit description, and inline comment. Also, from the library client perspective, `DV` is perfectly understandable, so this is a clearly a problem in the `letsencrypt` core.


pde:
Kuba my understanding is that the text in the IETF spec has not been carefully optimized in any way -- and certainly not for end users.  Nor does it constrain what clients should say to their users, who are of course a different audience.

IMO the acme library should provide minimally-confusing default text to everything that uses it; that way every client will benefit from the UI work that's been done via the library.  WDYT?

pde:
BTW I hear your argument about maintenance burden over in #1773, but we can provide resources to help you keep those messages up-to-date if it becomes a problem.

kuba:
We are discussing about UX here - that's a clear indication that this kind of problems should be fixed in the core, rather than `acme`. Again, if spec is not OK, then please update it. I want to copy and paste what is in spec.

kuba:
Given the direction in #2010, those error messages will not be even displayed to the `letsencrypt` end-user. Please merge, it's been a month now.

kuba:
Merged with `master` fixing conflicts due to recent changes.

@pde, would you mind finally looking at this and merging?

bmw:
Due this now being an EFF project and @pde being strongly against these changes, I'm closing this PR.

PR #1787, opened 2015-12-06
Change all /bin/sh and /bin/bash to /usr/bin/env

jeroenseegers:


Pyrohh:
I understand the rationale for using `env` with `bash`, but I've never used a system where `/bin/sh` was not a POSIX shell. Do such systems exist?

pde:
Several people have filed bugs because they're trying to use non-POSIX shells like BusyBox and `fish` as their `/bin/sh`.  At present we're not supporting those.

Although `env` is supposedly standard, I believe I actually tried this `#!/usr/bin/env sh` thing and found a system that it didn't work on. I can't remember what it was... possibly FreeBSD?

jeroenseegers:
Hmm, I've personally never encountered a system where it didn't work.

bmw:
Other than a few people who aren't using POSIX compatible shells as `/bin/sh`, the current approach has worked very well and is the most common shebang line used for shell scripts.

With this in mind, I'm going to close this PR and keep things as is. Regardless, thanks for writing the PR @jeroenseegers.

PR #1791, opened 2015-12-06
Adding instructions for staging server in documentation

cdancette:
Explains how to use the staging server to test letsencrypt


bmw:
I believe this was solved by #1848. If you disagree and don't think that change is adequate, let me know and I'll reopen.

PR #1805, opened 2015-12-07
Use print_function not print_statement

nakato:
Change the print statements used into print functions.

The print statement is not valid in Python 3, however the print function
is valid in at least 2.6+. __future__ provides support for older versions
as well as being required for the linter.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1,4 +1,6 @@
 """Let's Encrypt CLI."""
+from __future__ import print_function
+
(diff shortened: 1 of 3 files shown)


schoen:
LGTM

I think this is useful for future Python 3 compatibility and won't harm compatibility with any existing Python version.

bmw:
LGTM as well. Thanks for the PR @nakato.

kuba:
You should be using `six` instead. Have a look at `acme` code, that compatbile with Python 2.6, 2.7, 3.3, 3.4 and 3.5 :)

Issue #1795, opened 2015-12-06
Webroot breaks for non-root due to chown


SunMar:
Hi,

Today I'm trying out letsencrypt and stumbled upon that using webroot breaks when not using root due to a chown on line 72 of letsencrypt/plugins/webroot.py:

```
 49     def prepare(self):  # pylint: disable=missing-docstring
 50         path_map = self.conf("map")
 51
 52         if not path_map:
 53             raise errors.PluginError("--{0} must be set".format(
 54                 self.option_name("path")))
 55         for name, path in path_map.items():
 56             if not os.path.isdir(path):
 57                 raise errors.PluginError(path + " does not exist or is not a directory")
 58             self.full_roots[name] = os.path.join(path, challenges.HTTP01.URI_ROOT_PATH)
 59
 60             logger.debug("Creating root challenges validation dir at %s",
 61                          self.full_roots[name])
 62             try:
 63                 os.makedirs(self.full_roots[name])
 64                 # Set permissions as parent directory (GH #1389)
 65                 # We don't use the parameters in makedirs because it
 66                 # may not always work
 67                 # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
 68                 stat_path = os.stat(path)
 69                 filemode = stat.S_IMODE(stat_path.st_mode)
 70                 os.chmod(self.full_roots[name], filemode)
 71                 # Set owner and group, too
 72                 os.chown(self.full_roots[name], stat_path.st_uid,
 73                           stat_path.st_gid)
 74
 75             except OSError as exception:
 76                 if exception.errno != errno.EEXIST:
 77                     raise errors.PluginError(
 78                         "Couldn't create root for {0} http-01 "
 79                         "challenge responses: {1}", name, exception)
```

It tries to set the ownership of the directory to the owner of the parent. So currently to bypass this issue either the webroot must be chown'd to the letsencrypt user, or the script needs to be run as root. Neither of which are satisfactory solutions. Is chown at all neccesary when letsencrypt is not run as root?

pde:
Probably the right answer is that `chown` should be wrapped in a try: except: pass branch.  @lbeltrame, are you interested in taking this?

kuba:
That's really sad news :( In the meantime you can use https://github.com/kuba/simp_le - it's a client for webroot only, without any root requirements.

lbeltrame:
[...]

PR #1807, opened 2015-12-07
Make webroot usable also when running as non-root (GH #1795) - WIP

lbeltrame:
Thanks to @aburch's suggestions, the logic has been changed:

- Set umask before creating folders and files
- Leverage os.makedirs' mode option in conjunction with umask

The program still tries to change owner / group, but in case of errors
it continues gracefully.

Tests have been updated, and they pass.

This does not implement all the suggestions offered, and it's posted for review.

This should fix issue #1795.

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -60,23 +60,38 @@ to serve all files under specified web root ({0})."""
             logger.debug("Creating root challenges validation dir at %s",
                          self.full_roots[name])
             try:
-                os.makedirs(self.full_roots[name])
-                # Set permissions as parent directory (GH #1389)
-                # We don't use the parameters in makedirs because it
-                # may not always work
+                # Change the permissiosn to be writable (GH #1389)
+                # We also set umask because os.makedirs's mode parameter does
+                # not always work:
                 # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
+                # We set the umask instead of going the chmod route to ensure the client
+                # can also run as non-root (GH #1795)
+
                 stat_path = os.stat(path)
-                filemode = stat.S_IMODE(stat_path.st_mode)
-                os.chmod(self.full_roots[name], filemode)
-                # Set owner and group, too
-                os.chown(self.full_roots[name], stat_path.st_uid,
-                          stat_path.st_gid)
+                old_umask = os.umask(0o022)
+                os.makedirs(self.full_roots[name], 0o0755)
+
+                # Set owner as parent directory if possible
+
+                try:
+                    stat_path = os.stat(path)
+                    os.chown(self.full_roots[name], stat_path.st_uid,
+                             stat_path.st_gid)
+                except OSError as exception:
+                    if exception.errno == errno.EACCES:
+                        logger.debug("Insufficient permissions to change owner and uid - ignoring")
+                    else:
+                        raise errors.PluginError(
+                            "Couldn't create root for {0} http-01 "
+                            "challenge responses: {1}", name, exception)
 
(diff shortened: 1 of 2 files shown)


letsencrypt/plugins/webroot.py
```diff
@@ -60,23 +60,38 @@ def prepare(self):  # pylint: disable=missing-docstring
...
+                # We set the umask instead of going the chmod route to ensure the client
+                # can also run as non-root (GH #1795)
+
                 stat_path = os.stat(path)
-                filemode = stat.S_IMODE(stat_path.st_mode)
-                os.chmod(self.full_roots[name], filemode)
-                # Set owner and group, too
-                os.chown(self.full_roots[name], stat_path.st_uid,
-                          stat_path.st_gid)
+                old_umask = os.umask(0o022)
```
aburch: `old_umask` should be obtained outside the try block.  If anything up to and including this line throws an exception, `os.umask(old_umask)` in the `finally` block is otherwise called with an undefined value.

letsencrypt/plugins/webroot.py
```diff
@@ -95,18 +110,18 @@ def _path_for_achall(self, achall):
...
-        filemode = stat.S_IMODE(stat_parent_path.st_mode)
-        # Remove execution bit (not needed for this file)
-        os.chmod(path, filemode & ~stat.S_IEXEC)
-        os.chown(path, stat_parent_path.st_uid, stat_parent_path.st_gid)
+
+        old_umask = os.umask(0o022)
+
+        try:
+            with open(path, "w") as validation_file:
+                # Change permissions to be world-readable, owner-writable (GH #1795)
```
aburch: I think the comment about the permissions should be moved to the `os.umask` call.

SwartzCr:
It looks like you're still failing a linting test, should be easy to fix, just need to remove an unused variable and import. See: https://travis-ci.org/letsencrypt/letsencrypt/jobs/95318686 for details or run ./venv/bin/tox -e lint from your Let's Encrypt root

letsencrypt/plugins/webroot.py
```diff
@@ -59,24 +59,42 @@ def prepare(self):  # pylint: disable=missing-docstring
...
+
                 stat_path = os.stat(path)
-                filemode = stat.S_IMODE(stat_path.st_mode)
-                os.chmod(self.full_roots[name], filemode)
-                # Set owner and group, too
-                os.chown(self.full_roots[name], stat_path.st_uid,
-                          stat_path.st_gid)
+                # This is coupled with the "umask" call above because
+                # os.makedirs's "mode" parameter may not always work:
+                # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
```
SwartzCr: according to the python documentation the umask value is only used if the mode value is used: https://docs.python.org/2/library/os.html

letsencrypt/plugins/webroot.py
```diff
@@ -59,24 +59,42 @@ def prepare(self):  # pylint: disable=missing-docstring
...
+
+                os.makedirs(self.full_roots[name], 0o0755)
```
SwartzCr: is there a reason that this is below os.stat(path) now?

SwartzCr:
Besides these notes (and the failed linting) I think this is preferable to #1696 since it makes the files world readable rather than editing the permissions on the parent directory, as well as not needing root to run umask as called out in #1795 and mentioned in the comment on lines 64-65

lbeltrame:
Sorry for not answering sooner. Should I open a new PR with the requested changes?

bmw:
Sure! If you think you can clean it up a bit, another PR is appreciated.

PR #1808, opened 2015-12-07
Docker do not need sudo

pierreozoux:
(if well configured, but this is not the topic of this doc)

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -323,7 +323,7 @@ to, `install Docker`_, then issue the following command:
 
 .. code-block:: shell
 
-   sudo docker run -it --rm -p 443:443 -p 80:80 --name letsencrypt \
+   docker run -it --rm -p 443:443 -p 80:80 --name letsencrypt \
                -v "/etc/letsencrypt:/etc/letsencrypt" \
                -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
                quay.io/letsencrypt/letsencrypt:latest auth


PythonProdigy:
What if their Docker implementation is not configured properly? Are we leaving them out to dry with this change? I think it's important that Let's Encrypt does not force best practices upon anyone. Imagine a piece of software that required it's users to configure their system properly in order to use it! Linus Torvalds once said "Fix your f*cking 'compliance tool', because it is obviously broken. And fix your approach to kernel programming." Please be aware that these opinions are my own and in no way, shape or form reflect those of my employer.

pierreozoux:
I don't like Linus language, but this is another debate.

Docker is not the recommend way, and it is targeted at power users.
 https://github.com/pierreozoux/letsencrypt/blob/master/docs/using.rst#running-with-docker

And power user should configure well their docker.
https://docs.docker.com/engine/installation/ubuntulinux/#create-a-docker-group

but well I'll not debate. I found it strange, I PRed, but I'll not debate hours on that. Feel free to merge or close please, but no debate, it is not worth it.

PythonProdigy:
Pierre I appreciate your honesty and I see where you are coming from. I agree that users should configure their software properly but this is simply not the world we live in. If the United States Government used Docker do you think they would bother to follow the best practices guide and secure it properly? Many government sites don't even use SSL yet! Let's hope that let's encrypt will help start the movement to change that.

bmw:
Thanks for the PR @pierreozoux! I was actually on the fence about this issue for a while. After thinking about it a bit. I don't think we want this.

My reasoning is:
1) As pointed out by @PythonProdigy, running `docker` without `sudo` will not work for some users so we shouldn't been giving them invalid instructions. Taking this a step further, running `docker` with `sudo` should always work so let's prefer the more reliable instructions.
2) I'm not sure if we should be encouraging people to run `docker` without `sudo` as there are serious security implications of doing so (read [this](https://docs.docker.com/engine/articles/security/#docker-daemon-attack-surface) if you're not familiar). With this in mind, if we are encouraging anything, I think we should encourage the more secure method which is to run `docker` with `sudo` instead of adding your user to the `docker` group.

Regardless, thanks again pierreozoux!

PR #631, opened 2015-07-25
Apache Overhaul

jdkasten:
Sorry for the size of this PR, but it fixes most of the issues within Apache and raises test coverage to 100% for the Apache plugin.
Fixes #571 #603 #479 #394 and implements #17

In addition to the above issues... Here are some of the larger changes..
Parsing now incorporates current plugins and variable declarations and correctly applies them to IfModule and IfDefine directives.

Apache variables are interpreted correctly.

Apache works cleanly on the Debian default installation.

Better error handling in general.

Known list of things still left to do.
1.  Support ports other than 443 for https
2.  Support other Apache layouts (non-debian)
3.  Include IfModule <rewrite_module> around rewrite directives.
4.  Smarter auto-selection for vhosts.  It will prompt you if you have more than one non-default addressed VirtualHost.  (The command now runs seamlessly on the default Apache configuration)

PR #1811, opened 2015-12-07
Support apache plugin in different operating systems

### LINKED ISSUE #788: Support for multi-OS Apache
The Apache plugin has been designed and tested on the Ubuntu/Debian layout.  It shouldn't be that difficult to adapt the code to work with other OS configurations easily, but some work needs to be done.

Some initial ideas...

https://community.letsencrypt.org/t/installer-throwing-error-message/1120/9?u=jdkasten

Also, failure should be graceful for unsupported layouts...

joohoi:
Making it easy to add support for different operating systems and directory layouts by defining the OS specific apache layout and features in constants.py, users are able to override the values via command line if they have custom layout in place.

Fixes #788 

Tested on: Fedora 23, CentOS 7, Ubuntu 14.04, Debian 8
Defaults to the (original) Debian layout


pde:
@levskaya feel like firing this at the test farm?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -86,18 +86,29 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
...
+        add("ctl", default=constants.os_constant("ctl"),
             help="Path to the 'apache2ctl' binary, used for 'configtest', "
                  "retrieving the Apache2 version number, and initialization "
                  "parameters.")
-        add("enmod", default=constants.CLI_DEFAULTS["enmod"],
+        add("enmod", default=constants.os_constant("enmod"),
             help="Path to the Apache 'a2enmod' binary.")
-        add("dismod", default=constants.CLI_DEFAULTS["dismod"],
+        add("dismod", default=constants.os_constant("dismod"),
             help="Path to the Apache 'a2enmod' binary.")
```
sagi: Change a2enmod to a2dismod.

joohoi: Oh, wow. That's actually in master, but how I could have missed it while editing around it all the time. This is a good place to fix it though. Good find.

levskaya:
One last major issue I see is that 'sites-enabled' directory is still hard-coded in the configurator.py file, at lines 1118, 1148 on this version.  If one is on a centos7 system where things are put into the 'conf.d' directory, as opposed to sites-enabled, this causes the apache plugin to fail spuriously, however just creating the sites-enabled directory with nothing in it allows it to proceed.

To get this to work with red hat enterprise linux server, add
```
"red hat enterprise linux server": CLI_DEFAULTS_CENTOS
```
to the CLI_DEFAULTS dict in constants.py

lastly: the python platform.linux_distribution() isn't a great way to detect distros, for instance, amazon linux, a popular centos variant on EC2, reports ('','','')

joohoi:
I had missed the hardcoded values. They are still there, but the methods (is_site_enabled and enable_site) shouldn't be ever accessed in distributions that don't know the concept of enable / disable sites. So I made the checks to respect the constants value handle_sites which is used to control this behavior.

I'm going to do a few tests and add sane error messages if those are encountered (if user forces handle_sites = True in an environment that doesn't support it)

About the os detection: it's up for discussion, is there something better that we could use? I'm a bit hesitant to start manual fingerprinting, at least at larger scale, maybe on the corner cases ie. returned ('','','') by platform etc.

letsencrypt-apache/letsencrypt_apache/parser.py
```diff
@@ -56,9 +58,11 @@ def __init__(self, aug, root, ctl):
         # Set up rest of locations
         self.loc.update(self._set_locations())
 
-        # Must also attempt to parse sites-available or equivalent
-        # Sites-available is not included naturally in configuration
-        self._parse_file(os.path.join(self.root, "sites-available") + "/*")
+        # Take the CentOS layout into account, httpd.conf not in httpd root
+        self._parse_file(os.path.join(self.root, "conf") + "/httpd.conf")
```
sagi: The above line is redundant.  In line `45` the `self._parse_file(self.loc["root"])` call already parses `conf/httpd.conf` if run under CentOS. (See line `553`).

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -458,7 +471,10 @@ def _create_vhost(self, path):
                 is_ssl = True
 
         filename = get_file_path(path)
-        is_enabled = self.is_site_enabled(filename)
+        if self.conf("handle-sites"):
+            is_enabled = self.is_site_enabled(filename)
+        else:
+            is_enabled = True
```
sagi: If  `conf("handle-sites") == False` it does not mean that the site is enabled. If  `conf("handle-sites") == True`  it means that we are allowed to `a2enmode` a site, which we don't do here, here we only check if the site is enabled. Therefore, I think that `474-477` should be reverted to the original `is_enabled = self.is_site_enabled(filename)`. Unless I'm missing something.

joohoi: Outside of debian based distros there's no concept of sites-available/sites-enabled, everything that's found in the path is parsed (and hence enabled). 

If a2ensite is available in OS, handle-sites shouls be set true on constants for it, and vhost-root pointed to a directory where a2ensite picks it up. If your platform doesn't support enabling, handle-sites should be set false and vhost-root pointed to a directory that's parsed by apache.

letsencrypt-apache/letsencrypt_apache/parser.py
joohoi: You are right. I'll remove this line.

letsencrypt-apache/letsencrypt_apache/configurator.py
sagi: Thanks for explaining. 

Is the following scenario possible ?
`letsencrypt-auto` is ran under a debian-like distro, and the user specifies `--apache-handle-sites False` (or something like that). So a vhost may not be enabled, and `conf("handle-sites")` is False. (A user may just want to see the vhost config generated by letsencrypt, before enabling it).

Semantically, I think that the check for `conf("handle-sites")` (or other checks that signifies a non debian-like underlying distro) should be encapsulated in `self._is_site_enabled(...)`.

joohoi: The described scenario is possible, and it will also enable users to request certificates for sites that aren't currently enabled (temporarily disabled for maintenance reasons or whatnots), but exist in the sites-available.

My view of the semantics differs from yours; the site enabling concept is not in vanilla apache, and even though the apache plugin was originally created towards Debian based distributions I think they are the kind of corner case here. Hence semantically it wouldn't make sense to call a method regarding enabling in the context that doesn't know anything about enabling.

joohoi:
As @levskaya noted, platform.linux_distribution() isn't adequate way to discover what flavor of OS we're running. Does someone have a decent solution to that in mind? We could implement the same checks that are done by letsencrypt-auto for example... Discuss.

osirisinferi:
As [requested](): *a* Gentoo fingerprint:

    osiris@server ~ $ python -c "import platform;print(platform.system_alias(platform.system(), platform.release(), platform.version()))"
    ('Linux', '4.0.5-gentoo', '#2 SMP Thu Aug 27 16:33:32 CEST 2015')
    osiris@server ~ $

Furthermore, to add to the discussion: linux_distribution is going to be [deprecated](http://bugs.python.org/issue1322#msg207427) apparently.. I haven't fully read the discussion, but perhaps it gives an alternative in the referenced Python issue?

joohoi:
They don't seem to have any real answer to that, and are moving dist and linux_distribution out of platform package. The consensus seems to be that detecting distribution flavor is too much of a moving target for Python stdlib, which I agree completely of. There doesn't seem to be any potential alternatives aside from this: https://github.com/nir0s/ld , which is very new and stated experimental & not production ready.

So it seems to me that the only reasonable way forward is to bake our own fingerprinting method in. The implementation is still up for discussion. To me it seems like the easiest way forward would be using the same method than bootstrapper is currently using - searching for the distribution specific release files for most, and uname for FreeBSD for example. Additionally we could parse /etc/os-release to support distributions with systemd more easily.

pde:
@joohoi I've made #1925 into a separate ticket so that we don't have to block this PR on a solution to it.

pde:
One problem I'm seeing when I try to test the current version of this branch is that it seems to break detection of domain names on at least some systems (`lel` is my script for running my local git copy of letsencrypt)
```
$ lel --apache -t

1: dee.isnot.org
2: tweedle.isnot.org
Select the appropriate numbers separated by commas and/or spaces (Enter 'c' to
cancel):c
Please specify --domains, or --installer that will help in domain names autodiscovery
$ git checkout apache-portability # contains a merge of joohoi/multios_apache
Switched to branch 'apache-portability'
$ lel --apache -t

No names were found in your configuration files.
You should specify ServerNames in your config files in order to allow for
accurate installation of your certificate.
If you do use the default vhost, you may specify the name manually. Would you
like to continue?
(Y)es/(N)o:
```
I have the following conf files in `/etc/apache2/sites-available`, symlinked from `/etc/apache2/sites-enabled`:

```
$ cat /etc/apache2/sites-enabled/001-tweedle.conf 
<VirtualHost *:80>
        ServerName tweedle.isnot.org

	DocumentRoot /var/www/tweedle
	<Directory />
		Options FollowSymLinks
		AllowOverride None
	</Directory>
	<Directory /var/www/>
		Options Indexes FollowSymLinks MultiViews
		AllowOverride None
		Order allow,deny
		allow from all
	</Directory>

	ScriptAlias /cgi-bin/ /usr/lib/cgi-bin/
	<Directory "/usr/lib/cgi-bin">
		AllowOverride None
		Options +ExecCGI -MultiViews +SymLinksIfOwnerMatch
		Order allow,deny
		Allow from all
	</Directory>

	# Possible values include: debug, info, notice, warn, error, crit,
	# alert, emerg.
	LogLevel warn

</VirtualHost>
```
and
```
$ cat /etc/apache2/sites-enabled/001-dee.conf 
<VirtualHost *:80>
        ServerName dee.isnot.org

	DocumentRoot /var/www/tweedle
	<Directory />
		Options FollowSymLinks
		AllowOverride None
	</Directory>
	<Directory /var/www/>
		Options Indexes FollowSymLinks MultiViews
		AllowOverride None
		Order allow,deny
		allow from all
	</Directory>

	ScriptAlias /cgi-bin/ /usr/lib/cgi-bin/
	<Directory "/usr/lib/cgi-bin">
		AllowOverride None
		Options +ExecCGI -MultiViews +SymLinksIfOwnerMatch
		Order allow,deny
		Allow from all
	</Directory>

	# Possible values include: debug, info, notice, warn, error, crit,
	# alert, emerg.
	LogLevel warn

</VirtualHost>

```

pde:
This is Debian unstable. `constants` in `os_constant()` look correct for debian: `{'handle_mods': True, 'vhost_root': '/etc/apache2/sites-available', 'le_vhost_ext': '-le-ssl.conf', 'challenge_location': '/etc/apache2', 'enmod': 'a2enmod', 'server_root': '/etc/apache2', 'dismod': 'a2dismod', 'ctl': 'apache2ctl', 'handle_sites': True}` (though my [attempts to add logging in that function](https://github.com/letsencrypt/letsencrypt/commit/64044c9ecae38c22fee8d91c079a0674bd3bedf7) have been failing..)

joohoi:
That sounds rather weird. The only change that affects the virtualhost detection should be this: https://github.com/letsencrypt/letsencrypt/pull/1811/files#diff-35a5744d3537602a402f3dd9aaedd34bL483

I'll fire up an debian unstable vagrant box to try and weed it out.

joohoi:
I could not reproduce this. I booted up a up-to-date debian unstable, used your exact configuration files (including the file names), and it seems to work correctly in my env.

```
(venv)joona@debian:~/letsencrypt$ git branch
  master
* multios_apache
(venv)joona@debian:~/letsencrypt$ sudo venv/bin/letsencrypt --apache -t

1: dee.isnot.org
2: tweedle.isnot.org
Select the appropriate numbers separated by commas and/or spaces (Enter 'c' to
cancel):
```

joohoi:
I tested it out also with apache-portability branch from the main repo. I'd guess you have something funky going on with the lel - script, as it worked fine for me with that branch as well. After I commented out 

raise ValueError, "bang" 

from constants.py. 

```
(venv)joona@debian:~/letsencrypt$ git branch
* apache-portability
  master
(venv)joona@debian:~/letsencrypt$ sudo venv/bin/letsencrypt --apache -t
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>
Got logger <logging.Logger object at 0x7f7b8714b590>

1: dee.isnot.org
2: tweedle.isnot.org
Select the appropriate numbers separated by commas and/or spaces (Enter 'c' to
cancel):
```

osirisinferi:
OK guys, we've got a Gentoo "problem":

I've edited constants.py so I could test my Gentoo server for you (from the `apache-portability` branch), so it wouldn't default to the Debian configuration. But it seems that `apache2ctl` isn't what it looks like:

> 2015-12-17 15:29:36,667:DEBUG:letsencrypt.le_util:/usr/sbin/apache2ctl
> WARNING:letsencrypt_apache.parser:Error in checking parameter list: /etc/init.d/apache2: invalid option -- 't'
> 
> 2015-12-17 15:29:36,735:WARNING:letsencrypt_apache.parser:Error in checking parameter list: /etc/init.d/apache2: invalid option -- 't'
> 
> DEBUG:letsencrypt.plugins.disco:Misconfigured PluginEntryPoint#apache: Apache is unable to check whether or not the module is loaded because Apache is misconfigured.
>   File "/usr/lib64/python2.7/site-packages/letsencrypt/plugins/disco.py", line 103, in prepare
>     self._initialized.prepare()
>   File "/usr/lib64/python2.7/site-packages/letsencrypt_apache/configurator.py", line 161, in prepare
>     self.aug, self.conf("server-root"), self.conf("vhost-root"), self.conf("ctl"))
>   File "/usr/lib64/python2.7/site-packages/letsencrypt_apache/parser.py", line 39, in __init__
>     self.update_runtime_variables(ctl)
>   File "/usr/lib64/python2.7/site-packages/letsencrypt_apache/parser.py", line 98, in update_runtime_variables
>     stdout = self._get_runtime_cfg(ctl)
>   File "/usr/lib64/python2.7/site-packages/letsencrypt_apache/parser.py", line 140, in _get_runtime_cfg
>     "Apache is unable to check whether or not the module is "
> MisconfigurationError: Apache is unable to check whether or not the module is loaded because Apache is misconfigured.

So I took a look at `apache2ctl`:

> server letsencrypt-apache # cat /usr/sbin/apache2ctl 
> #!/bin/sh
> exec /etc/init.d/apache2 "$@"
> server letsencrypt-apache # 

I don't think that's how it looks on other distro's? :stuck_out_tongue: Anyway, the `-t` switch isn't gonna work for the Gentoo init script I'm afraid...

I'm assuming we really need those runtime variables? Anyway to skip that part (with a big disclaimer to Gentoo users) and hope for the best? :worried:

joohoi:
TBH the functionality that _get_runtime_cfg actually provides is a bit of a mystery to me. I haven't hit any case in any of the environments where I could get any meaningful data out of apache2ctl -D DUMP_RUN_CFG, and the matches stays unpopulated. This functionality was introduced in #631 can you help us out & shed some light on this @jdkasten ?

jdkasten:
It is meant to parse Apache variables appropriately.
http://httpd.apache.org/docs/2.4/mod/core.html#define

That way you can do things like

Define config_path "/etc/..."
Include ${config_path}

You can also do things like...

Define TEST

<IfDefine TEST>
  Include additional_config.conf
</IfDefine>

Where the directives will only be seen if the "variables" are defined within the configuration.

I believe the tests are most indicative of the intended performance.
Tests: https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/tests/complex_parsing_test.py

Test Data: https://github.com/letsencrypt/letsencrypt/tree/master/letsencrypt-apache/letsencrypt_apache/tests/testdata/complex_parsing

https://github.com/letsencrypt/letsencrypt/issues/1208#issuecomment-152839686
https://github.com/letsencrypt/letsencrypt/issues/1208#issuecomment-157689550

jdkasten:
The binary was chosen to avoid issues with Apache's parsing order.

Apache only makes  a single pass through the configuration so the ordering of includes like
```Include conf/*```
are problematic.  The parser's representation has to match Apache's.  (Variables can be defined and then depended later on in the configuration.  More problematic for IfDefine scenarios)

The binary seemed like the easiest way to precisely get which variables Apache knew about.

(Although somewhere in the tree there is an implementation using configuration directives.)

Best,
James

**Comment by osirisinferi** [comment]
> server letsencrypt-apache # /usr/sbin/apache2 -h     
> Usage: /usr/sbin/apache2 [-D name] [-d directory] [-f file]
>                          [-C "directive"] [-c "directive"]
>                          [-k start|restart|graceful|graceful-stop|stop]
>                          [-v] [-V] [-h] [-l] [-L] [-t] [-T] [-S] [-X]
> Options:
>   -D name            : define a name for use in <IfDefine name> directives
>   -d directory       : specify an alternate initial ServerRoot
>   -f file            : specify an alternate ServerConfigFile
>   -C "directive"     : process directive before reading config files
>   -c "directive"     : process directive after reading config files
>   -e level           : show startup errors of level (see LogLevel)
>   -E file            : log startup errors to file
>   -v                 : show version number
>   -V                 : show compile settings
>   -h                 : list available command line options (this page)
>   -l                 : list compiled in modules
>   -L                 : list available configuration directives
>   -t -D DUMP_VHOSTS  : show parsed vhost settings
>   -t -D DUMP_RUN_CFG : show parsed run settings
>   -S                 : a synonym for -t -D DUMP_VHOSTS -D DUMP_RUN_CFG
>   -t -D DUMP_MODULES : show all loaded modules 
>   -M                 : a synonym for -t -D DUMP_MODULES
>   -t                 : run syntax check for config files
>   -T                 : start without DocumentRoot(s) check
>   -X                 : debug mode (only one worker, do not detach)
> server letsencrypt-apache # 

I have no clue what the `apache2ctl` binary should be doing, but it seems Gentoo's `/usr/sbin/apache2` binary can perform the same `-t` as `apache2ctl`?

**Comment by joohoi** [comment]
Thanks for explanation @jdkasten ! Seems like we need a new constant for this as the apachectl function usage fluctuates so much between the distributions. Do you feel like we should let it graciously in hopes of supporting versions of distributions not explicitly defined n constants?

**Comment by joohoi** [comment]
@osirisinferi yeah, -t -D DUMP_RUN_CFG is the exact thing it does. Actually it seems to support other things that apache2ctl is used on other distributions. You could try replacing the apache2ctl in the constants with the path to the /usr/sbin/apache2 , that might be the right way on gentoo.

**Comment by osirisinferi** [comment]
I'm guessing Gentoo support for the Apache plugin is going to be a pain in the #*&$*#...

It seems Gentoo is a little bit strange in terms of its [configuration](https://wiki.gentoo.org/wiki/Apache#Configuration)..

httpd.conf is filled with constructions like the following:

> <IfDefine SSL>
> LoadModule ssl_module modules/mod_ssl.so
> </IfDefine>

And in the `conf.d` file for the init.d script, all defines are specified like `-D SSL` for the command line.

So, even if you change `apache2ctl` to `apache2`, one runs into the problem of missing all those extra `-D` switches when [configurator.py wants to do its configtest](https://github.com/letsencrypt/letsencrypt/blob/apache-portability/letsencrypt-apache/letsencrypt_apache/configurator.py#L1316):

Well, in that specific case, `configtest` is non-existant in `apache2`, but when we change that to `-t`, the Gentoo equivalent, it gives the following error:

> ERROR:letsencrypt.le_util:Error while running apache2 -t.
> 
> AH00526: Syntax error on line 15 of /etc/apache2/vhosts.d/vhosts/example.com.conf:
> Invalid command 'SSLEngine', perhaps misspelled or defined by a module not included in the server configuration

This because `-D SSL` is missing from the command. A solution would be to extract the variable `APACHE2_OPTS` out of `/etc/conf.d/apache2` and append it to every `apache2` command.

**Comment by joohoi** [comment]
I introduced a new constant (but not cli variable) for the command to check apache2 version, and got the gentoo fingerprint right. Could you test again against the latest changes @osirisinferi ? I think I got gentoo sorted out too.

**Comment by pde** [comment]
Regarding the dump runtime config things: over in his fix for #1208, @SwartzCr has implemented some code that searches for Define statements and errors out if (1) `apache2ctl` won't dump the runtime config, and (2) Define statements are present.  But of course having a way to actually get the Defined variables is better.

**Comment by joohoi** [comment]
For apache 2.2 there's no way to get the define statements in a sane way. @SwartzCr implemented a check that errors out with a a good error message if defines are found in configuration files, and we planned to implement that functionality on this one.

**Comment by osirisinferi** [comment]
@joohoi Gentoo is still failing. I'm still seeing a reference to `apache2ctl`. Remember, this is just a shell script indirectly linking `apache2ctl` to the init-script. And most commands tried by Let's Encrypt aren't meant for the init-script :wink: 
Besides, I'm missing the whole "Gentoo has filled all the config files with <IfDefine>'s and puts all the -D switches in the /etc/conf.d/apache2 file and we need to add all these -D switches to every single call of `/usr/sbin/apache2`, otherwise every configfile with any mod_ssl directive (or any other optional module) will fail if the -D SSL isn't on the CLI call" part?

**Comment by joohoi** [comment]
Yeah, apache2ctl is still used it's used for restarts etc. Are you running gentoo stable? Because stable provides apache 2.2 and that's what I was testing against.

**Comment by osirisinferi** [comment]
Aaahh, good point!

No, I'm running Apache testing, currently 2.4.18.

No clue why it's testing on Gentoo; if there's a distro with a motto spelling "Living on the edge" it's Gentoo :stuck_out_tongue_closed_eyes:

**Comment by joohoi** [comment]
@osirisinferi how about now, after the last change? It works fine on my gentoo/testing as well as gentoo/stable platform currently, with the default Gentoo constants.

**Comment by osirisinferi** [comment]
Hmm, no cigar.. I'm wondering if my Apache configuration is something ancient from the past, which I've always updated with new stuff here and there, but never really to a version which is currently in use?

Let me check for myself with a VirtualBox Gentoo system, because I'm finding this very strange.. :confused: Why my Apache configuration is so different from yours :stuck_out_tongue:

**Comment by joohoi** [comment]
@osirisinferi what error do you get? Or does LE just simply not find your vhosts?

**Comment by osirisinferi** [comment]
The error is:

> 2015-12-22 14:24:55,653:WARNING:letsencrypt_apache.parser:Error in checking parameter list: > AH00526: Syntax error on line 15 of /etc/apache2/vhosts.d/vhosts/example.com.conf:
Invalid command 'SSLEngine', perhaps misspelled or defined by a module not included in the server configuration

This is because mod_ssl is not loaded, because `-D SSL` is missing from the command line (which is specified in a separate runscript variable in `/etc/conf.d/apache2`. But I understand that this/my configuration is apparently not standard Gentoo (anymore?), so I'm probably not the appropriate person to help you with this :disappointed: Currently, I'm installing a fresh Gentoo installation in VirtualBox too, so I'll let you know if I know anything further.

**Comment by joohoi** [comment]
For Gentoo this is very hard to manage, as there are so many variables in the game. We would need to make sure the user has "ssl" USE-flag set (and if it isn't, enable it and recompile \@world, which takes from two hours to forever). Additionally we'd have to make sure the -D SSL exists in the init script configuration in /etc/conf.d/apache2 . For now we'll have to settle with being able to auto-detect vhosts & setup the configuration. 

This kind of distribution based extra care may be feasible in distant future, if we implement something like way extended bootstrapping or OS based preinstall hooks.

**Comment by joohoi** [comment]
The USE-flag and /etc/conf.d/apache2 modifications are more in the realm of OS packaging anyway.

**Comment by osirisinferi** [comment]
Gentoo has a script called `equery`, which allows you to check the current USE flags for an installed package, like this: `equery u www-servers/apache | grep ssl`:

> server ~ # equery u www-servers/apache | grep ssl
> -libressl
> +ssl
> server ~ # 

Unfortunately, this is provided by the package `app-portage/gentoolkit` which isn't installed by default :neutral_face: 

But indeed, Gentoo is quite the distro which is heavily customizable. Perhaps it's wise for Let's Encrypt not to mess too much there. It would be nice though, if it would be possible to have the Apache plugin working to get it all automated.

**Comment by joohoi** [comment]
Well, the prerequisites are fulfilled by package dependencies when and if it's packaged. We are way better off leaving that to the right realm in such a volatile environment.

File: letsencrypt-apache/letsencrypt_apache/configurator.py [hunk 4]
```diff
@@ -1308,7 +1342,8 @@ def get_version(self):
 
         """
         try:
-            stdout, _ = le_util.run_script([self.conf("ctl"), "-v"])
+            stdout, _ = le_util.run_script(
+                constants.os_constant("version_cmd").split(" "))
```
pde: Parsing the command in this way is slightly hackish; it might create headaches for us if we ever want to support paths with spaces in them.  But I'm not sure that I have a better idea (we don't really want to make people type python lists on the command line); perhaps if anyone ever asks for that, we can add some kind of escaping to this parser line?

pde:
Code LGTM.  Working on some test farm tests...

osirisinferi:
If that line of code is just neede for Gentoo... One might consider just dropping it, as the setup on Gentoo is quite a hassle. The fresh install on VirtualBox has all the same `-D SSL` switches in `/etc/conf.d/apache2` as on my own server, so I guess there's no way around that silly setup.

letsencrypt-apache/letsencrypt_apache/configurator.py
joohoi: Actually these constants for commands aren't available as command line parameters to users. So we're pretty much free to write the python lists in the constants file. And you are right, this isn't the right place to parse the string or generate a list out of it. I'll move the lists to constants directly.

bmw:
Results of test farm test:
```
0 ubuntu15.10 pass
1 ubuntu15.04LTS pass
2 ubuntu14.04LTS pass
3 ubuntu14.04LTS_32bit pass
4 ubuntu12.04LTS fail
5 debian8.1 pass
6 debian7.8.aws.1 fail
7 debian7.8.aws.1_32bit fail
8 amazonlinux-2015.09.1 fail
9 amazonlinux-2015.03.1 fail
10 RHEL7 pass
11 fedora23 pass
12 centos7 pass
13 centos6 fail
```

joohoi:
@osirisinferi the commands are for general inter-OS operability. Do you still have problems besides the bootstrapping? If you have -D SSL in the /etc/conf.d/apache2 - file already you should be good to go with default configuration from LE.

osirisinferi:
Well.. Remembering a topic on the Community forum it "hit" me: I never put `<IfDefine SSL>` (or `<IfModule mod_ssl.c>`) statements around my `<VirtualHost *:443>` statements, because I never run my Apache without `-D SSL` and its not some Fortune 500 company production server of the sorts.. But if I **do** put those IfDefines around it, at least I get to the domain selection part. Progress!
But I think by that you're just bypassing the SSL VirtualHosts. Now it only detects the non-SSL VirtualHosts, even if the running Apache server does recognise the SSL hosts.. I have no clue if that's a bad thing or if it doesn't matter.

Now I'm running into another problem, perhaps because of those `<IfDefine SSL>`s, maybe not.. It doesn't update my configurations with an "Cannot find a cert or key directive in /bla/bla.conf/IfModule/VirtualHost. VirtualHost was not modified" error.. But that's beyond the scope of this PR I recon ;)

bmw:
As requested by @joohoi, here's a few logs of the failed tests:

[ubuntu12.04LTS](https://gist.github.com/4cb97098f153386b82b1)
[debian7.8.aws.1](https://gist.github.com/14ca263a595dce981d09)
[debian7.8.aws.1_32bit](https://gist.github.com/38b186ef4dab66d5fced)
[centos6](https://gist.github.com/daab431c0dfc827fc7a5)

pde:
Overall those test results are a big improvement on what we had prior to this PR, so my instinct is let's land this and open new issues for further work.  In fact I've made #2012 for the Debian 7 / Ubuntu 12.04 stuff.  @bmw if you agree, want to go ahead and hit merge?

pde:
Sorry, one thing: I want to double check the bootstrap changes.  I tried to do that yesterday but my test farm drifted off into the æther.  Trying again now.

pde:
(Turns out that test farm test is broken, will try to get it fixed if I can...)

pde:
letsencrypt-auto tests passed

pde:
leauto tests are passsing.  test_apach2.sh is still showing a significant improvement over master:

```
0 ubuntu15.10 pass
1 ubuntu15.04LTS pass
2 ubuntu14.04LTS pass
3 ubuntu14.04LTS_32bit pass
4 ubuntu12.04LTS fail
5 debian8.1 pass
6 debian7.8.aws.1 fail
7 debian7.8.aws.1_32bit fail
8 amazonlinux-2015.09.1 fail
9 amazonlinux-2015.03.1 fail
10 RHEL7 pass
11 fedora23 pass
12 centos7 pass
13 centos6 fail
```

mkormendy:
I realize this is closed .. but this is missing OSX.

PR #1814, opened 2015-12-07
Check domain list from plugin get_all_names() and remove wildcard, punycode etc. [needs tweak]

### LINKED ISSUE #1664: nginx plugin returns localhost and other non-public names from get_all_names()
Probably these should be pruned to ensure that the client doesn't ask for things it can't have!

### LINKED ISSUE #1682: Don't try to auto-configure local names
In https://community.letsencrypt.org/t/dns-name-does-not-have-enough-labels/5036/2, we have a couple of people who ran into trouble because they had a virtualhost for a local name, i.e. one with no dots. We should omit these from the list of names we try to auto-configure.

More generally, we could try to reproduce some of the validity checks that happen in boulder, so if you have "foo.corp", the client could see that that doesn't end in a public suffix and ignore it.

Alternately, we could be a little friendlier about presenting "these names failed to validate. do you want to issue a cert for the other names anyhow, or keep trying to validate these names?"

### LINKED ISSUE #1708: Should not select domains which can't be used with letsencrypt
Neither international (names starting with `xn--`) nor wildcard domains are currently supported by letsencrypt. In the "Which names would you like to activate HTTPS for?" dialog these domains should not be selectable.

image

joohoi:
Using the same rules than command line -d checks, remove wildcard, punycode, non-FQDN and names including unicode characters before the domain selection list.

Fixes #1708 
Fixes #1664
Fixes #1682


pde:
@joohoi are there reasons to do this in the Apache plugin rather than in the main client code?  If we move the check up to the main client, can we solve #1664 at the same time?

joohoi:
The check itself still resides in the main client code (le_utils to be accurate) and as such, can be called from the nginx plugin as well. The pruning is done in between parsing the configuration and displaying the options to user, both of which are done in the plugin code realm so I think it should stay there.

ArchimedesPi:
Reviewed, LGTM.

letsencrypt/le_util.py
```diff
@@ -280,3 +280,37 @@ def __call__(self, unused1, unused2, unused3, option_string=None):
...
+
+    """
+    # Check if there's a wildcard domain
+    if domain.startswith("*."):
+        raise errors.ConfigurationError(
+            "Wildcard domains are not supported")
+    # Punycode
+    if "xn--" in domain:
+        raise errors.ConfigurationError(
+            "Punycode domains are not supported")
```
pde: I think this and the internationalized errors should say "not presently supported".

pde:
Looks good to me on a cursory read too, will merge after that one wording change.

pde:
Great!

PR #1815, opened 2015-12-07
Add warning about rate limit and how to use test server [needs revision]

arnaudmm:


schoen:
I'm a little reluctant to put this as the "first thing to know", but I think I would support having this information somewhere in the document.

schoen:
I'd say that the existence of the rate limit can be mentioned as soon as the end of the first paragraph, but the specific details about the testing server are not relevant to *most* users, and so should be moved somewhere like the end of the section (we don't want people to think that they should do that by default, because they will get invalid certs, which I think ought to be reiterated specifically).

arnaudmm:
Edited as adviced

README.rst
```diff
@@ -48,7 +48,8 @@ How to run the client
 
 In many cases, you can just run ``letsencrypt-auto`` or ``letsencrypt``, and the
 client will guide you through the process of obtaining and installing certs
-interactively.
+interactively. If you want to test your workflow, you may be interested by using
+a test server because production server is rate limited (More informations below)
```
jimrollenhagen: nit: s/informations/information

jimrollenhagen:
It would be nice to include the actual limits here, but that could also become out of date fairly quickly. Worth the tradeoff?

bmw:
Let's Encrypt's rate limits have been increased significantly since this was written and there are far fewer complaints about rate limits so I no longer think this is necessary.

Regardless, thanks for the PR!

Issue #1766, opened 2015-12-05
Apache config parse failure with SSL_CLIENT_S_DN_CN and an array


philvdb:
Hello (especially @domcleal , trying out LE and *loving* it so far.

Here's my apache parsing trouble right now (PluginError(('There has been an error in parsing the file (%s): %s', u'/etc/apache2/sites-enabled/xy-le-ssl.conf', u'Syntax error')), which hinges on the following line in my apache configuration:

SSLRequire %{SSL_CLIENT_S_DN_CN} in {"foo@bar.com", "bar@foo.com"}

If you need the entire config file, let me know. But commenting out this line leads to LE running normally.

domcleal:
Thanks for the bug report.  I've fixed this in Augeas in commit https://github.com/hercules-team/augeas/commit/f86a28d03a5c42a6c58293667a95d7794e30a42f.

There's a conflicting change being made to the LE Augeas module right now, but as soon as that's merged I'll submit the change back to LE.

philvdb:
Awesome. Thank you!

Issue #1755, opened 2015-12-04
Apache plugin fails; syntax problem with mod_deflate.conf


seezee:
Command:
user@server:~/letsencrypt$ ./letsencrypt-auto run --apache -d domain.com -d www.domain.com

Result:
Updating letsencrypt and virtual environment dependencies.......
Running with virtualenv: sudo /home/user/.local/share/letsencrypt/bin/letsencrypt run --apache -d domain.com -d www.domain.com
The apache plugin is not working; there may be problems with your existing configuration.
The error was: PluginError(('There has been an error in parsing the file (%s): %s', u'/etc/apache2/conf.d/mod_deflate.conf', u'Syntax error'),)

Contents of mod_deflate.conf:
# Enable mod_deflate (gzip)
    SetOutputFilter DEFLATE
    # Do not attempt to compress the following extensions
    SetEnvIfNoCase Request_URI \
    \.(?:gif|jpe?g|png|swf|flv|zip|gz|tar|mp3|mp4|m4v)$ no-gzip dont-vary

pde:
This is the same bug as #1531.  Has now been fixed in master.

pde:
(Fix was #1692)

pde:
I was wrong, this isn't fixed by #1692.

pde:
But it does appear to be fixed by #1827 :)

domcleal:
Seems to be #1692 as you first said I think.

MarcelWaldvogel:
On a fresh `git clone`, `letsencrypt-auto --apache -d example.com -d www.example.com` still fails on
```SetEnvIfNoCase Request_URI \.(?:exe|t?gz|zip|iso|Z|bz2|sit|rar|png|jpg|gif|jpeg|flv|swf|mp3)$ no-gzip dont-vary```

domcleal:
We're discussing why this doesn't seem to be fixed at https://github.com/letsencrypt/letsencrypt/issues/1531#issuecomment-162809832 and subsequent comments.  I suspect letsencrypt-auto isn't updating the lens in ~/.local.

PR #1827, opened 2015-12-08
Merge Augeas lens fix for SSLRequire wordlists

domcleal:
Augeas fails to parse the wordlist (args inside braces) in the
SSLRequire directive.

From hercules-team/augeas@f86a28d

Closes: #1766


pde:
Fixes: #1755

pde:
I continue to be terrified by these regexps.  But this does okay on the (still inadequate) tests we have, and appears to also fix a bug it wasn't explicitly designed for.

domcleal:
Yeah, they can be terrifying!  We rely heavily on the test harness in Augeas core to maintain them, as it not only tests that certain snippets can be parsed, but also typechecks the lens to prevent ambiguities and other subtle bugs.  As long as we keep what's in LE in sync with upstream Augeas, it should be fine - if we deviate then we would need to run those tests in LE too.

PR #1828, opened 2015-12-08
Abort when no Pythons are found

mgedmin:
It seems ill-advised to continue without setting the LE_PYTHON variable, when the very next command tries to use it.


pde:
LGTM but I want a second +1 since this merge goes straight to users.

bmw:
:+1:

Issue #1699, opened 2015-12-03
Apache ports.conf parsing too strict for Listen 443


Maescool:
In my configuration, I only have certain IP's bound to the 443
so it could be `Listen 10.1.1.1:443` (given private IP as an example, IRL it's a public one)
it seems to only look for `Listen 443`
cause other the other IP's bound to this machine are used by different services.
Snippits of log:
````
2015-12-03 22:40:32,796:DEBUG:letsencrypt_apache.configurator:No Listen 443 directive found. Setting the Apache Server to Listen on port 443
2015-12-03 22:40:33,178:DEBUG:letsencrypt.reverter:Creating backup of /etc/apache2/ports.conf
2015-12-03 22:40:33,178:DEBUG:letsencrypt.reverter:Creating backup of /etc/apache2/apache2.conf
2015-12-03 22:40:33,289:ERROR:letsencrypt.le_util:Error while running apache2ctl configtest.
Action 'configtest' failed.
The Apache error log may have more information.

AH00526: Syntax error on line 2 of /etc/apache2/le_tls_sni_01_cert_challenge.conf:
The address or port is invalid

2015-12-03 22:40:33,295:DEBUG:letsencrypt.error_handler:Encountered exception:
Traceback (most recent call last):
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/auth_handler.py", line 115, in _solve_challenges
    dv_resp = self.dv_auth.perform(self.dv_c)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 1289, in perform
    self.restart()
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 1199, in restart
    self.config_test()
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 1222, in config_test
    raise errors.MisconfigurationError(str(err))
MisconfigurationError: Error while running apache2ctl configtest.
Action 'configtest' failed.
The Apache error log may have more information.

AH00526: Syntax error on line 2 of /etc/apache2/le_tls_sni_01_cert_challenge.conf:
The address or port is invalid
2015-12-03 22:40:33,296:DEBUG:letsencrypt.error_handler:Calling registered functions
[...]

PR #1831, opened 2015-12-08
Apache configurator respects Listen ip:port statements

### LINKED ISSUE #1739: Not possible to issue a certificate if Apache does not bind to all IPs
Firstly, very pleased with this project, and how far it has some in such a sort space of time!

On saying that, I do have an issue regarding Apache.  My current `ports.conf` is this, obviously replacing `(IP)` with my actual IP;
```
# If you just change the port or add more ports here, you will likely also
# have to change the VirtualHost statement in
# /etc/apache2/sites-enabled/000-default.conf

Listen (IP):80

<IfModule ssl_module>
        Listen (IP):443
</IfModule>

<IfModule mod_gnutls.c>
        Listen (IP):443
</IfModule>

```
However, this means certificate issuance fails with the following error;
```
IMPORTANT NOTES:
 - The following 'urn:acme:error:connection' errors were reported by
   the server:

   Domains: minersrealm.net, www.minersrealm.net
   Error: The server could not connect to the client for DV
```
...this means I am forced to use this configuration...;
```
# If you just change the port or add more ports here, you will likely also
# have to change the VirtualHost statement in
# /etc/apache2/sites-enabled/000-default.conf

Listen 80

<IfModule ssl_module>
        Listen 443
</IfModule>

<IfModule mod_gnutls.c>
        Listen 443
</IfModule>

```
...which then means Apache is binding to **all** IPs on the server, something which I do not want.  My main problem with this is I wish to automate the SSL renewal once every two months, something which I cannot do if I need to manually update the configuration everytime I wish to renew a certificate.

watercrossing:
Fixes #1739. 

The current `prepare_server_https` method fails if Apache is set up to bind to specific IPs only. It believes that there is no `Listen` statement at all, and adds a generic `Listen 443`. If a specific `Listen 127.0.0.1:443` already exists, apache fails to start up as it tries to bind to an already occupied port. (`/var/log/apache2/error.log`: `Address already in use: AH00072: make_sock: could not bind to address 0.0.0.0:443`. 

The fix in this pull requests now iterates through all Listen statements, and checks if they are on the correct port and if not, adds the appropriate Listen statement. Its not perfect yet, as it creates a new Listen statement for every listen statement which is not the target ssl port, no matter if a vhost for which we want to create a certificate actually needs this IP. But this is quite the edge case and I can't think of a neat way of integrating this restriction.

I have added some additional tests which give full coverage.

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -545,21 +545,43 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         # Check for Listen <port>
         # Note: This could be made to also look for ip:443 combo
-        if not self.parser.find_dir("Listen", port):
-            logger.debug("No Listen %s directive found. Setting the "
-                         "Apache Server to Listen on port %s", port, port)
-
-            if port == "443":
-                args = [port]
+        listens = [self.parser.get_arg(x).split()[0] for x in self.parser.find_dir("Listen")]
+        # In case no Listens are set (which really is a broken apache config)
+        if not listens:
+            listens = ["80"]
+        for listen in listens:
+            # For any listen statement, check if the machine also listens on Port 443.
+            # If not, add such a listen statement.
+            if len(listen.split(":")) == 1:
+                # Its listening to all interfaces
+                if port not in listens:
+                    if port == "443":
+                        args = [port]
+                    else:
+                        # Non-standard ports should specify https protocol
+                        args = [port, "https"]
+                    self.parser.add_dir_to_ifmodssl(
+                        parser.get_aug_path(
+                            self.parser.loc["listen"]), "Listen", args)
+                    self.save_notes += "Added Listen %s directive to %s\n" % (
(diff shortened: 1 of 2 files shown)


pde:
Addresses #1699 #1831

joohoi:
Needs some pep8 - love

watercrossing:
I have fixed the pep8 formatting of my previous commit. `pep8.travis.sh` does highlight quite a lot in the rest of letsencrypt...

joohoi:
Code LGTM, tested and works. Handles ipv4 & ipv6.

joohoi:
Thanks for fixing the pep8 stuff, it's not followed fanatically, but it's good to implement where reasonable. There sure are more fixes to do, but considering the LOCs in letsencrypt client, the violations are very rare.

joohoi:
Code LGTM, tested and works. Handles ipv4 & ipv6.

pde:
Great, landing!

koehn:
I had to shut down other services that had listeners on 443 and 80 on IPs not used by Apache, even when Apache was shut down, since letsencrypt tries to open  `*:80` and `*:443`. FYI.

PR #1846, opened 2015-12-09
[TYPO] Fix help string for --apache-dismod

remram44:


pde:
This is also fixed in #1811, but we may as well take this for 0.1.1, rather than waiting for that large PR to land in 0.2.0.

Issue #1545, opened 2015-11-17
Provide helpful error messages in case of rate limits


pde:
- Ensure the user isn't reissuing a cert when they should just be reinstalling.
- Suggest the staging API if the user is testing, or running a private infrastructure
- Provide a place to go to request overrides

centminmod:
what about `--staging` flag which overrides `--server` url to the staging location ?

PR #1848, opened 2015-12-09
Add staging server hint to avoid rate limit issues

patf:
Users accidentally running into rate limit is the most common issue on the community forum and on IRC.

This adds a more obvious hint encouraging users to use the staging server initially.

I moved the EPEL note up a bit so those two boxes aren't right next to each other, and because it makes more sense to put it before the command IMO. Happy to change that though, comes down to personal preference.

fixes parts of #1545

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -31,16 +31,21 @@ Firstly, please `install Git`_ and run the following commands:
 
 .. _`install Git`: https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
 
+.. note:: On RedHat/CentOS 6 you will need to enable the EPEL_
+   repository before install.
+
+.. _EPEL: http://fedoraproject.org/wiki/EPEL
+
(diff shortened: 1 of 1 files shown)


patf:
switched to the new `--staging` flag implemented in #1850

pde:
LGTM, but since doc changes go live immediately, will merge this once 0.1.1 has actually shipped and people have that flag available to them.

osirisinferi:
Per [3b58109](https://github.com/letsencrypt/letsencrypt/commit/3b5810995ddc161c5263cd77d044d2a0be14b5c2) the flag is called `--test-cert` now.

patf:
updated flag and wording since "test" is probably more obvious to most users.

bmw:
LGTM. Merging. Thanks for another PR @patf.

pde:
This was really a 0.1.1 documentation fix.

PR #1850, opened 2015-12-10
Implement --staging

### LINKED ISSUE #1667: Implement --staging
This should point the client at the staging server, but also take steps to ensure that lineages with valid certs aren't accidentally updated with happy hacker fake CA ones.

pde:
- Fixes #1667
- Should help with rate limit problems

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -696,6 +696,15 @@ class HelpfulArgumentParser(object):
         parsed_args = self.parser.parse_args(self.args)
         parsed_args.func = self.VERBS[self.verb]
 
+        # Do any post-parsing homework here
+
+        # argparse seemingly isn't flexible enough to give us this behaviour easily...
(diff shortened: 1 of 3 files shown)


letsencrypt/cli.py
```diff
@@ -696,6 +696,15 @@ def parse_args(self):
         parsed_args = self.parser.parse_args(self.args)
         parsed_args.func = self.VERBS[self.verb]
 
+        # Do any post-parsing homework here
+
+        # argparse seemingly isn't flexible enough to give us this behaviour easily...
+        staging_uri = 'https://acme-staging.api.letsencrypt.org/directory'
```
bmw: This should be moved to `constants.py` and all other references to the URL should be changed to use the constant. This URL could change and we don't want to create new bugs/problems/work for us in the future.

bmw:
Other than my complaint about hard coding the URL in 3 different places, this LGTM.

pde:
Fixed.

bmw:
LGTM

Issue #1546, opened 2015-11-17
Existing Certificate <Replace> or <Cancel> question should actually default to <Reinstall>


pde:
Currently if there is an existing cert for the names you requested with `run`, you get this message:

`You have an existing certificate that contains exactly the same domains you requested (ref: /etc/letsencrypt/renewal/isnot.org.conf) Do you want to renew and replace this certificate with a newly-issued one?`

These are the wrong options.  Prior to the renewal window, the default option should be **\<Reinstall\>**.  After the typical renewal window has opened, the **\<Replace\>** default should become **\<Renew\>**.

Also we should convince ourselves that it's actually a good idea to ask these questions at all; and ensure that we have CLI flags to control the answer.

pde:
Until we fix this, we believe that some people will make themselves many copies of the same certificate unintentionally.  So flagging this as relatively urgent amongst **nice for 1.0** milestones.

pde:
Temporarily putting this in **1.0 for launch**, to just see if we can do it.

bmw:
I think this issue is actually little tricky. Just to make sure everyone is on the same page including myself, this question is asked when requesting a new certificate for a subset of domains found in a lineage. This can be a strict subset or identical names to an existing lineage.

This is actually quite subtle. Imaging you have a lineage for example.com, example.org, and example.net and then run the client requesting a cert for example.com and example.org (excluding example.net). What should the client do? Should it update the existing lineage? If so, it may break a server for example.net. If it doesn't update the existing lineage, the configurations for example.com and example.net need to be updated.

With that said, I think there is value in adding some of the functionality @pde proposed above. The one choice I'm not sure I agree with is **\<Reinstall\>**. The problem is this question comes up during authentication, not installation. I believe we have correctly been trying to make `run` the equivalent of `certonly` followed by `install` (see #771). This moves in the opposite direction as allowing the user to choose **\<Reinstall\>** when running `certonly` doesn't make sense.

The one thing we might be able to do is provide something like a **\<Keep existing cert\>** option (hopefully with better wording). Probably a little early to jump into implementation details, but the way I see this working is if **\<Keep existing cert\>** is chosen, `_auth_from_domains()` in `cli.py` simply returns the existing lineage without any modification. If this is done, `certonly` will return with no further effect and `run` will give that certificate to the `installer`.

bmw:
Given the complications in the short novel I just wrote, I think getting this done properly by Thursday is a little optimistic.

[...]

PR #1860, opened 2015-12-11
Implement <Reinstall> option in addition to <Renew> or <Cancel>

schoen:
This is missing some documentation and has no updated tests. Also, `_treat_as_renewal` could usefully be refactored (I think some branches might even be dead now because the duplicative-cert case tries to return earlier), but I *think* the overall logic is correct for each case that's relevant to #1546.

You can use `--duplicate` or `--treat-as-renewal` to trigger the two behaviors in question (that is, getting a completely separate lineage regardless of the amount of overlap, and replacing the chosen lineage automatically with a new cert without being prompted). I guess there is still no distinctive CLI flag for "just try to reinstall without trying to get a new cert, and without interactively asking whether that's what the user wanted", although there's a whole separate CLI verb that was meant to address that case, right?


letsencrypt/cli.py
```diff
@@ -232,11 +248,44 @@ def _treat_as_renewal(config, domains):
...
         question = None
         if ident_names_cert is not None:
+            # TODO: I bet this question is confusing to people who don't know
+            #       how the backend repreentation of certificates work. The
+            #       distinction is renewal updates existing lineage with new
+            #       cert (eventually maybe preserving the privkey), while
+            #       newcert creates a new lineage. And reinstall doesn't cause
+            #       a new issuance at all.
+            if config.renew_by_default:
+                return "renew", ident_names_cert
```
pde: Also default to "renew" if the cert is within its renewal window.

letsencrypt/cli.py
```diff
@@ -232,11 +248,44 @@ def _treat_as_renewal(config, domains):
         # configuration file.
         question = None
         if ident_names_cert is not None:
```
pde: Split this entire question into another function?

letsencrypt/cli.py
pde: We should also default to "renew" if the cert is within its renewal window, I think.

pde:
I started lookin at the tests. ` test_certonly_new_request_failure` seems to be broken because `_treat_as_renewal()` is somehow returning `None`.

pde:
@schoen this is looking close, it just needs the near-to-expiry check that I researched in IRC, and a bit of cleanup to split that unreadably large function.

pde:
I've given this a fairly thorough field testing, and believe it is of acceptable risk to deploy to canaries now.  It should have a few unit tests added soon, though.

pde:
@bmw if the CLI flags and UX work for you, can you merge this for the canaries?

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
 
 def _treat_as_renewal(config, domains):
-    """Determine whether or not the call should be treated as a renewal.
+    """Determine whether there are duplicated names and how to handle them.
 
-    :returns: RenewableCert or None if renewal shouldn't occur.
-    :rtype: :class:`.storage.RenewableCert`
+    :returns: Two-element tuple containing desired new-certificate behavior as
+              a string token ("reinstall", "renew", or "newcert"), plus either
+              a RenewableCert instance or None if renewal shouldn't occur.
```
bmw: nit: Should add `:rtype: tuple` below this (I don't really care if we fix this before merging because we don't generate documentation for `cli.py` and you said it was a tuple in `:returns:`).

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
-
-        if renewal:
-            return ident_names_cert if ident_names_cert is not None else subset_names_cert
-
-    return None
+    if config.duplicate:
+        return "newcert", None
+    # TODO: Also address superset case
+    ident_names_cert, subset_names_cert = _find_duplicative_certs(config, domains)
+    # XXX ^ schoen is not sure whether that correctly reads the systemwide
```
bmw: If this refers to the systemwide renewal configuration file, I can tell you that it is not read.

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
+    # configuration file.
+    if not (ident_names_cert or subset_names_cert):
```
bmw: Should be `if ident_names_cert is None or subset_names_cert is None:`

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
+        return _handle_identical_cert_request(config, ident_names_cert)
+    elif subset_names_cert is not None:
+        return _handle_subset_cert_request(config, domains, subset_names_cert)
+
+def _handle_identical_cert_request(config, cert):
+    """Figure out what to do if a cert has the same names as a perviously obtained one
+
+    :param storage.RenewableCert cert:
+
+    :returns: Tuple of (string, cert_or_None) as per _treat_as_renewal
```
bmw: Same nit as above.

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
+        "{br}(ref: {0}){br}{br}What would you like to do?"
+    ).format(cert.configfile.filename, br=os.linesep)
+
+    if config.verb == "run":
+        keep_opt = "Attempt to reinstall this existing certificate"
+    elif config.verb == "certonly":
+        keep_opt = "Keep the existing certificate for now"
+    choices = [keep_opt,
+               "Renew & replace the cert (limit ~5 per 7 days)",
+               "Cancel this operation and do nothing"]
```
bmw: Why have a 2nd cancel option? Both the CLI and dialog menus have a built in way to cancel.

letsencrypt/cli.py
```diff
@@ -878,9 +963,11 @@ def prepare_and_parse_args(plugins, args):
                 "multiple -d flags or enter a comma separated list of domains "
                 "as a parameter.")
     helpful.add(
-        None, "--duplicate", dest="duplicate", action="store_true",
-        help="Allow getting a certificate that duplicates an existing one")
-
+        None, "--keep-until-expiring", "--keep", "--reinstall",
```
bmw: What is the benefit of these aliases? I personally feel that it complicates the command line with little to no gain.

bmw: Why doesn't this go in the `automation` group like other similar flags?

letsencrypt/cli.py
```diff
@@ -889,15 +976,24 @@ def prepare_and_parse_args(plugins, args):
         version="%(prog)s {0}".format(letsencrypt.__version__),
         help="show program's version number and exit")
     helpful.add(
+        "automation", "--expand", "--expand-existing-certs", "--replace", action="store_true",
```
bmw: Same comment about aliases.

letsencrypt/cli.py
```diff
@@ -889,15 +976,24 @@ def prepare_and_parse_args(plugins, args):
...
+             "previously attained cert (often --keep-until-expiring is "
+             "more appropriate). Implies --expand.")
     helpful.add(
         "automation", "--agree-tos", dest="tos", action="store_true",
         help="Agree to the Let's Encrypt Subscriber Agreement")
     helpful.add(
         "automation", "--account", metavar="ACCOUNT_ID",
         help="Account ID to use")
+    helpful.add(
+        "automation", "--duplicate", dest="duplicate", action="store_true",
```
bmw: `HelpfulArgumentParser` doesn't support this, but perhaps we should make some of these `automation` flags mutually exclusive. It makes the code and interaction between the flags simpler.

letsencrypt/cli.py
```diff
@@ -314,10 +367,21 @@ def _suggest_donate():
...
+    # (which results in treating the request as a renewal), or newcert
+    # (which results in treating the request as a new certificate request).
+
+    action, lineage = _treat_as_renewal(config, domains)
+    if action == "reinstall":
+        # The lineage already exists; allow the caller to try installing
+        # it without getting a new certificate at all.
+        return lineage
+    elif action == "renew":
+        original_server = lineage.configuration["renewalparams"]["server"]
```
bmw: This is outside the scope of this PR, but I think we should carefully define what elements of the renewal configuration can be modified by the user. For instance, this code assumes the `server` parameter was not deleted.

letsencrypt/cli.py
pde: One reason for the multiple names is to basically always have a name that matches the button the user presses ("Reinstall" for run mode, "Keep" for certonly); the other reason is to have a longform default that is fairly self-explanation, and a short abrreviation.  Maybe that can be -k.

letsencrypt/cli.py
pde: Agreed.

letsencrypt/cli.py
pde: I think Seth added this because he thought the one in the corner was easy to miss.  I think that may be true, but I don't have data either way.

letsencrypt/cli.py
```diff
@@ -341,6 +405,27 @@ def _auth_from_domains(le_client, config, domains):
 
     return lineage
 
+def _avoid_invalidating_lineage(config, lineage, original_server):
+    "Do not renew a valid cert with one from a staging server!"
```
bmw: I think the code in this function is a bit hacky, but it works.

I'm curious if we can validate the certificate in this function. We require the `ca-certificates` package to be installed which I believe is used by programs like `pip`. If there is a way in Python for us to find the location of the certificates bundle, I believe we can use `fullchain.pem` and `PyOpenSSL` to validate the certificate. I think that might be a better and more reliable test than what we have here.

I don't think we have to block on this though.

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
+
+    :param storage.RenewableCert cert:
+
+    :returns: Tuple of (string, cert_or_None) as per _treat_as_renewal
+
+    """
+    if config.renew_by_default:
+        logger.info("Auto-renewal forced with --renew-by-default...")
+        return "renew", cert
+    if cert.should_autorenew(interactive=True):
```
bmw: Question: Is setting this new `interactive` flag to `True` the behavior you want? I think it's somewhat counter intuitive as the user manually set the `autorenew` to `False` but when he ran the client, it autorenewed his cert.

We could revert changes to `storage.py` and then if `cert.should_autorenew()` returns `False`, we check if `cert.autorenewal_is_enabled`. We can then either fall through and prompt the user or ask them to modify the config file or run the client again with `--renew-by-default`.

I'm not sure there is a problem with this, I just suspect this behavior may surprise some users.

letsencrypt/cli.py
pde: I don't remember why I didn't put it in automation, though there was a reason of sorts.  Moving it there now.

letsencrypt/cli.py
```diff
@@ -205,76 +205,129 @@ def _find_duplicative_certs(config, domains):
...
+        "You have an existing certificate that contains a portion of "
+        "the domains you requested (ref: {0}){br}{br}It contains these "
+        "names: {1}{br}{br}You requested these names for the new "
+        "certificate: {2}.{br}{br}Do you want to replace this existing "
+        "certificate with the new certificate?"
+    ).format(cert.configfile.filename,
+             existing,
+             ", ".join(domains),
+             br=os.linesep)
+    if config.expand or config.renew_by_default or zope.component.getUtility(
```
bmw: Question: Should `--renew-by-default` trigger this? I feel this is also slightly unexpected as this is not strictly a renewal.

If we do want this, I feel like having both `--renew-by-default` and `--expand` is redundant. Maybe we could just use one flag, perhaps with a name change.

pde: I agree, `--renew-by-default` probably shouldn't trigger it, but that's what it did previously so I feel that we now need to be cautious about changing its behaviour.

letsencrypt/cli.py
bmw: Hm. Maybe it is easy to miss but this is the default behavior we have on all of our menu prompts. I'd recommend removing the option from the list and if we have a problem with our dialog menu, we fix it for all of our menus, not just this one.

letsencrypt/cli.py
bmw: If we want to keep the same behavior then, I'd recommend removing the `--expand` flag.

letsencrypt/cli.py
pde: The real reason we added that flag is that right now, renewal.conf files have been generated without any `autorenew` entry in them, and [this code](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/storage.py#L520) treats that as a "no".

Perhaps what `interactive=True` should do is actually change the default from `False` to `True`.

pde: I totally misread that line of storage.py.  The defaults it has are okay.

pde:
Testing renewal by adding something like `renew_before_expiry = 89 days` to my renewal conf files, seems to work.  However, it produces some ridiculous log entries along the way:

```
2015-12-13 01:17:19,037:WARNING:letsencrypt.cli:Renewal configuration file /etc/letsencrypt/renewal/.isnot.org.conf.swp is broken. Skipping.
2015-12-13 01:17:19,042:DEBUG:letsencrypt.storage:Should renew, certificate has been expired since 2016-02-29 21:52:00 UTC.
2015-12-13 01:17:19,042:INFO:letsencrypt.cli:Cert is due for renewal, auto-renewing...
```

pde:
It'll now say:
```
2015-12-13 01:26:14,177:DEBUG:letsencrypt.storage:Should renew, less than 80 days before certificate expiry 2016-02-29 21:52:00 UTC.
2015-12-13 01:26:14,177:INFO:letsencrypt.cli:Cert is due for renewal, auto-renewing...
```

pde:
Sending to the canaries on master!

Issue #1861, opened 2015-12-11
Versioned dependency on PyOpenSSL interferes with packaging


pde:
Currently acme depends on PyOpenSSL >= 0.15.  The comment in setup.py says this is for `X509req.get_extensions()`, but the only place where that's called appears to be [wrapped in a fallback](https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/crypto_util.py#L171).

If we can relax this version dependency, that will greatly assist packaging for RHEL 6/7 and Debian 7.

pde:
I tried testing this on a Debian squeeze system (I couldn't get PyOpenSSL==0.13 to pip install on newer systems):

`sudo venv/bin/letsencrypt --webroot -w /var/www/tweedle/ -d tweedle.isnot.org -w /var/www/dee -d dee.isnot.org certonly -t -vv --debug`

I see this error:
```
Traceback (most recent call last):
  File "venv/bin/letsencrypt", line 9, in <module>
    load_entry_point('letsencrypt', 'console_scripts', 'letsencrypt')()
  File "/home/admin/letsencrypt/letsencrypt/cli.py", line 1283, in main
    return args.func(args, config, plugins)
  File "/home/admin/letsencrypt/letsencrypt/cli.py", line 512, in obtain_cert
    _auth_from_domains(le_client, config, domains)
  File "/home/admin/letsencrypt/letsencrypt/cli.py", line 336, in _auth_from_domains
    lineage = le_client.obtain_and_enroll_certificate(domains)
  File "/home/admin/letsencrypt/letsencrypt/client.py", line 283, in obtain_and_enroll_certificate
    certr, chain, key, _ = self.obtain_certificate(domains)
  File "/home/admin/letsencrypt/letsencrypt/client.py", line 266, in obtain_certificate
    return self._obtain_certificate(domains, csr) + (key, csr)
  File "/home/admin/letsencrypt/letsencrypt/client.py", line 228, in _obtain_certificate
    authzr)
  File "/home/admin/letsencrypt/acme/acme/client.py", line 310, in request_issuance
    headers={'Accept': content_type})
  File "/home/admin/letsencrypt/acme/acme/client.py", line 631, in post
    data = self._wrap_in_jws(obj, self._get_nonce(url))
  File "/home/admin/letsencrypt/acme/acme/client.py", line 508, in _wrap_in_jws
    jobj = obj.json_dumps().encode()
[...]

PR #1863, opened 2015-12-11
Reducing the PyOpenSSL version dep [do not merge]

pde:
Partially addresses #1861, facilitaes packaging on Debian Jessie and possibly Fedora 22, though not [Debian Wheezy](packages.debian.org/python-openssl) or [CentOS 6/7](https://www.rpmfind.net/linux/rpm2html/search.php?query=pyOpenSSL), which have 0.13.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -12,8 +12,11 @@ install_requires = [
     'cryptography>=0.8',
     'ndg-httpsclient',  # urllib3 InsecurePlatformWarning (#304)
     'pyasn1',  # urllib3 InsecurePlatformWarning (#304)
-    # Connection.set_tlsext_host_name (>=0.13), X509Req.get_extensions (>=0.15)
-    'PyOpenSSL>=0.15',
+    # PyOpenSSL version dependencies:
+    # Connection.set_tlsext_host_name (>=0.13)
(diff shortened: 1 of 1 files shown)


pde:
This was tested on Debian wheezy.

pde:
I tested this with `venv/bin/pip install pyopenssl==0.14` on a debian wheezy system, and it succeeded in obtaining and replacing certs, which by my reading should have triggered the get_extensions code path.

However I don't think all the unit tests pass with pyopenssl==0.14, so this isn't ready to merge.

pde:
We didn't get a solution to this for 0.1.1, so it will have to catch the 0.2.0 train.

kuba:
Package maintainers can always patch `setup.py` as necessary. Please don't change those version constraints, especially pinning (`==`) to specific version.

pde:
Just to be clear the pin is there to get tox and travis to test that version, and will not be in this PR when it's ready for merging!

itamarjp:
It's interesting to have letsencrypt working with pyopenssl => 0.13 ( to have it packaged for centos 6),  but just changing setup.py isn't the solution.

pde:
Closing this since bmw is doing the real work for this in another branch.

PR #1869, opened 2015-12-11
Apache conf library

pde:
More conf files.  And a very hackish script to run each of them through the client, so you can see that it does what you expect (though in some cases, what you expect is breakage because of the nature of the test).


pde:
No user-facing changes, so merging.

PR #1873, opened 2015-12-12
Don't specify nargs when deprecating args

bmw:
Removes the parameter having you specify the number of args in `add_deprecated_argument`. Instead, `*` is used for `nargs` which consumes all arguments after the flag being deprecated up to the next flag.

[Documentation](https://docs.python.org/2/library/argparse.html#nargs)

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -98,7 +98,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             help="SSL vhost configuration extension.")
         add("server-root", default=constants.CLI_DEFAULTS["server_root"],
             help="Apache server root directory.")
-        le_util.add_deprecated_argument(add, "init-script", 1)
+        le_util.add_deprecated_argument(add, "init-script")
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


pde:
Do we have any flags that take more than 1 argument?  If not, it seems that this PR gives us some risk without reward.  If yes, LGTM.

bmw:
No. We do not currently have any flags that take more than one argument. I think what this buys us is ease of use and code cleanliness. Instead of having to pass around whether the flag took 0 or 1 arguments, it'll just work.

Since `*` only consumes arguments up to the next flag. I think there is little/no risk here. There is some negative behavior though. Currently on `master`, you get the following behavior:
```
$ letsencrypt --agree-dev-preview 42
Use of --agree-dev-preview is deprecated.
usage: 

letsencrypt: error: unrecognized arguments: 42
```

With this PR, you still get the deprecation message, but the client will run normally because the parser consumed 42 as an argument to `--agree-dev-preview`.
```
$ letsencrypt --agree-dev-preview 42
Use of --agree-dev-preview is deprecated.
```
With this in mind, maybe we don't want this?

bmw:
After thinking about this more, I don't think we want this. I could change it to use `nargs=?` which consumes no more than one argument, but that still allows things that don't make sense like `--agree-dev-preview 42`.

Let's not make our command line sloppy.

PR #1875, opened 2015-12-12
Webroot permissions

### LINKED ISSUE #1795: Webroot breaks for non-root due to chown

bmw:
Supersedes #1807 and fixes #1795.

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -60,23 +60,38 @@ to serve all files under specified web root ({0})."""
             logger.debug("Creating root challenges validation dir at %s",
                          self.full_roots[name])
             try:
-                os.makedirs(self.full_roots[name])
-                # Set permissions as parent directory (GH #1389)
-                # We don't use the parameters in makedirs because it
-                # may not always work
+                # Change the permissiosn to be writable (GH #1389)
+                # We also set umask because os.makedirs's mode parameter does
+                # not always work:
                 # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
+                # We set the umask instead of going the chmod route to ensure the client
+                # can also run as non-root (GH #1795)
+
                 stat_path = os.stat(path)
-                filemode = stat.S_IMODE(stat_path.st_mode)
-                os.chmod(self.full_roots[name], filemode)
-                # Set owner and group, too
-                os.chown(self.full_roots[name], stat_path.st_uid,
-                          stat_path.st_gid)
+                old_umask = os.umask(0o022)
+                os.makedirs(self.full_roots[name], 0o0755)
+
+                # Set owner as parent directory if possible
+
+                try:
+                    stat_path = os.stat(path)
+                    os.chown(self.full_roots[name], stat_path.st_uid,
+                             stat_path.st_gid)
+                except OSError as exception:
+                    if exception.errno == errno.EACCES:
+                        logger.debug("Insufficient permissions to change owner and uid - ignoring")
+                    else:
+                        raise errors.PluginError(
+                            "Couldn't create root for {0} http-01 "
+                            "challenge responses: {1}", name, exception)
 
             except OSError as exception:
                 if exception.errno != errno.EEXIST:
                     raise errors.PluginError(
                         "Couldn't create root for {0} http-01 "
                         "challenge responses: {1}", name, exception)
+            finally:
+                os.umask(old_umask)
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
         assert self.full_roots, "Webroot plugin appears to be missing webroot map"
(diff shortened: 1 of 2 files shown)


lucasRolff:
After this was merged, it means that servers that prevents reading files owned by other users (also known as symlink attacks) - they'll not be able to validate anymore due to removed code here: https://github.com/letsencrypt/letsencrypt/pull/1875/files#diff-522ab130649a0ce14df40114d4ccd0b5L109 

Maybe it should be added back (in a try/except block) - or we should be able to pass an argument with the uid/gid we want to use for the challenge files, else this prevents us for issuing let's encrypt certificates on all our shared hosting boxes.

kuba:
If you're on shared hosting box how do you get root? Which servers "that prevent reading (...)" do you mean? Web servers? Which attacks do you mean?

lucasRolff:
@kuba - I as a system administrator - that wants to offer letsencrypt to my customers.

Basically if you run multiple users on a server, you often want to prevent things from happening, you can either do it by using software like CloudLinux which creates a virtual environment for each user, or you can do other things like installing some extra code in Apache to prevent e.g. a symlink vulnerability in Apache.

Basically in Apache if you enable `SymLinksIfOwnerMatch` and `FollowSymlinks` in Apache you become vulerable to a race condition through symlinks, which allows a user (attacker) to create a symlink to another users file, and he will be able to read it.

To prevent this you can patch apache so it cannot read other users files (which in a shared environment makes sense anyway), since the challenge files are created by root and is owned by root, it means the challenge validation will fail because apache (or any other web-server or server software preventing the webserver) from reading files in the www-folder to read files not owned by the vhost user, or apache/nginx/www-data user will simply fail.

Likewise if you would run with selinux which is very strict about ownership and permissions on folders in /var/www it could cause the webserver being prevented from reading the files.

So to be able (as a shared host) or people protecting in various ways, we would need the possibility to be able to decide who the files should be owned as, or inherit from the parent directly where the file are located.

ZeusF1:
@bmw 
Why did validation files no longer set chown?

I have an issue on Centos6 with with, when permissions are set incorrectly, and ownership for files is root.

Directories have correct permissions, but not a files.

I made short test how permissions works:

` cat test3.py
import os
old_mask = os.umask (022)
fh3 = open ("test3.txt", "w")
os.umask (0)
fh4 = open ("test4.txt", "w")
os.umask (old_mask)
fh3.close()
fh4.close()`

`cat test.py
import os
oldmask = os.umask (022)
fh1 = os.open ("test1.txt", os.O_CREAT, 0o777)
fh2 = os.open ("test2.txt", os.O_CREAT, 0o022)
os.umask (oldmask)
os.close (fh1)
os.close (fh2)`

Result will be:

```
-rwxr-x---+  1 root         root            0 Nov  2 12:47 test1.txt
----------+  1 root         root            0 Nov  2 12:47 test2.txt
-rw-r-----+  1 root         root            0 Nov  2 12:46 test3.txt
-rw-r-----+  1 root         root            0 Nov  2 12:46 test4.txt
```

Just as same as for https://community.letsencrypt.org/t/webroot-plugin-wrong-permission-on-validation-files-nginx-403-error/21728

Any ideas how to fix that or set ownership? ?

PR #1876, opened 2015-12-12
Simple verify

TheNavigat:
Addresses issue https://github.com/letsencrypt/letsencrypt/issues/1586


kuba:
It's pretty daunting task to verify challenges at the plugin level. This should be verified at the core instead.

TheNavigat:
Seems like a better idea.

Where should I start, though?

kuba:
Look where `perform()` is called and check it's return value. Based on the type, call `simple_verify` with appropriate arguments.

TheNavigat:
I checked ```continuity_auth.py```, ```def perform```, where ```responses.append(self.proof_of_pos.perform(achall))``` is. However, the responses here are in text.

Same goes for ```_solve_challenges``` in ```auth_handler.py```. ```self.cont_c``` and ```self.dv_c``` are text arrays are well.

I'm stuck, I guess.

TheNavigat:
Ping @kuba. Any help?

kuba:
https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt/interfaces.py#L171 defines what plugins return in `perform()`. At https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt/auth_handler.py#L115  `dv_resp` is a list of `ChallengeResponse` and `dv_c` is a list of corresponding challenges.

TheNavigat:
I've almost triple checked, and I've made sure that dv_resp is actually a list of strings, not a list of ```ChallengeResponse```. That's at least what happens when it's called using ```tox -e cover```.

I've tried to trace the code, that's roughly what I've ended up with.

_solve_challenges (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt/auth_handler.py#L106) in auth_handler.py:

```dv_resp = self.dv_auth.perform(self.dv_c)```

```dv_resp``` is array of strings.

```self.dv_auth``` is set in line 46 (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt/auth_handler.py#L46), ```self.dv_auth = dv_auth```

That is elated to auth_handler_test.py, ```self.mock_dv_auth = mock.MagicMock(name="ApacheConfigurator”)``` (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt/tests/auth_handler_test.py#L70)
Type of ```dv_auth``` is ApacheConfigurator, figured that out by printing the type and raising an error/forcing an exit.

Parent class is ```Configurator```. ```perform``` method contains ```apache_dvsni = dvsni.ApacheDvsni(self)``` (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt-apache/letsencrypt_apache/configurator.py#L1131)

```dvsni``` is a class, dvsni.py (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt-apache/letsencrypt_apache/dvsni.py). ```perform``` method contains ```responses.append(self._setup_challenge_cert(achall))```, (https://github.com/letsencrypt/letsencrypt/blob/d209b5ea20c2f5f35eadc304dbe5cbb85c233f4d/letsencrypt-apache/letsencrypt_apache/dvsni.py#L71)

```_setup_challenge_cert``` is defined in common.py, https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/plugins/common.py#L180

```response, (cert, key) = achall.response_and_validation(cert_key=cert_key)```, https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/plugins/common.py#L189

Which leads to letsencrypt/achallenges.py, ```def response_and_validation(self, *args, **kwargs):```, https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/achallenges.py#L52

And that leads to letsencrypt/acme/acme/challenges.py, ```def response_and_validation(self, account_key, *args, **kwargs)```, https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/challenges.py#L204, which calls response in the same class, https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/challenges.py#L177. That function calls ```response_cls```, which is set to not implemented, https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/challenges.py#L162, and I'm not sure what's the next step to take from here.

The most confusing thing is that by printing ```dv_resp```, it appears that the contents of the list are strings. Enumerating over each of them and printing the type returns string as well.

**Comment by kuba** [comment]
I would advise not to check the types available during tests (such as `tox`), as we often mock stuff (e.g. by using strings). Try to run the client against staging environment and check the types there.

**Comment by TheNavigat** [comment]
Okay so, one step closer.

The current checks for ```dv_resp``` do work but they won't pass the tests, because apparently string responses are generated in ```auth_handler_test.py```, https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/tests/auth_handler_test.py#L482

I guess what needs to be done right now is returning valid responses that pass simple_verify.

**Comment by TheNavigat** [comment]
Also, any advice on how to test continuity authentication, since apparently the unit testing won't work?

File: letsencrypt/auth_handler.py [hunk 1]
```diff
@@ -113,6 +113,13 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for response in dv_resp:
+                        for achall in self.dv_c:
+                            if not response.simple_verify(
```
kuba: That will not work, because `simple_verify` takes different arguments depending on the response class. You have to condition on `isinstance`

kuba:
Yes, you will need to change existing tests and a new ones.

TheNavigat:
Any hints on how to generate a proper response, though?

kuba:
Either `mock.(Magic)Mock` or construct some fake responses for test purposes only.

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,15 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for response in dv_resp:
+                        for achall in self.dv_c:
+                            print(type(response))
+                            print(response)
```
kuba: remove debug statements

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,15 @@ def _solve_challenges(self):
...
+                            if not response.simple_verify(
```
kuba: this is still not fixed, use `isinstance`

letsencrypt/tests/auth_handler_test.py
```diff
@@ -479,8 +479,11 @@ def test_different_errors_and_domains(self, mock_zope):
 
 def gen_auth_resp(chall_list):
     """Generate a dummy authorization response."""
-    return ["%s%s" % (chall.__class__.__name__, chall.domain)
+    from acme.challenges import HTTP01Response
+    return [HTTP01Response(key_authorization=u'foo')
```
kuba: docstring says "dummy", you're doing something wrong

TheNavigat: Why's that? The returned HTTP01Response object is still not the actual response. I'm not exactly sure how does something classify as "dummy" when it comes to uniy testing.

kuba: If someone starts debugging failing tests he will be surprised that he sees `HTTP01Response` in some random places.

In general, you shouldn't touch this function. You should rather rewrite tests for the function in question.

TheNavigat: The test will fail, though, because str objects have no ```simple_verify``` method :/

kuba: I have already told you that you need to use `isinstance` and use different calls to `simple_verify` depending on which response class is returned. If you do so, then you need to write separate tests: one for response class - to check if correct arguments are passed to `simple_verify`. You won't be using `gen_auth_resp` anymore, so you don't need to touch it.

TheNavigat:
I think this should work for HTTP01 responses, however, I had to comment that line for the tests to pass. https://github.com/letsencrypt/letsencrypt/pull/1876/files#diff-9960178edda4b54481c1c2db41631748R173

I triple checked that ```logger.warning``` is actually called. It even appeared in the console when I commented the ```logging.disable(logging.CRITICAL)```, but the test always fails. I'll trying to figure out why does that happen.

Also, for the PyLint errors, I'm working on it.

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,16 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for response in dv_resp:
+                        if isinstance(response, challenges.HTTP01Response):
```
kuba: what about `tls-sni-01`? it should be implemented as well

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,16 @@ def _solve_challenges(self):
...
+                            for achall in self.dv_c:
+                                if(isinstance(achall.challb.chall, challenges.HTTP01) or
+                                    isinstance(achall.challb.chall, challenges.TLSSNI01)):
```
kuba: That's wrong logic, because:
- it tries to simple verify (almost) all possible pairs of challenges and responses
- tries to match `http-01` response with `tls-sni-01` challenge

You should rather use `for chall, resp in zip(challenges, responses)`. `interfaces.py` documentation should be updated to make it explicit that plugins are required to return responses in the same order as challenges.

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,30 @@ def test_perform_failure(self):
...
+            challb=acme_util.TLSSNI01_P, domain=domain, account_key=key)
+
+        dv_c = [http_01, tls_sni_01]
+
+        self.handler.dv_c = dv_c
+
+        self.mock_dv_auth.perform.side_effect = gen_http01_resp
+
+        mock_verify.return_value = False
+        # with mock.patch("letsencrypt.plugins.manual.logger") as mock_logger:
```
kuba: don't try to check if logger was called enough times. check instead that `simp_le` verify was called enough times AND with correct arguments

TheNavigat:
https://github.com/letsencrypt/letsencrypt/pull/1876/files#diff-9960178edda4b54481c1c2db41631748R172

Do I need to use ```get_authorizations``` instead? Or would it be fine to tell PyLint to ignore accessing protected methods from testing scripts?

Also, I'm not exactly sure why Travis Boulder tests fail. It looks like something's wrong with the TLS challenge, but I can't be sure enough.

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,18 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        if isinstance(response, challenges.HTTP01Response):
+                            if not response.simple_verify(
+                                achall.chall, achall.domain,
+                                achall.account_key.public_key(), response.PORT):
```
kuba: `self.config.http01_port` should be used, as it was previously

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,18 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        if isinstance(response, challenges.HTTP01Response):
+                            if not response.simple_verify(
+                                achall.chall, achall.domain,
```
kuba: Continuation lines in the `if` condition should be indented by additional 4 spaces. Same below (in total 4 lines have wrong indentation). Otherwise they are indented to the same level as `logger.warning` statement...

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,18 @@ def _solve_challenges(self):
...
+                                achall.account_key.public_key(), response.PORT):
+                                logger.warning("Self-verify of challenge failed.")
+                        elif isinstance(response, challenges.TLSSNI01Response):
+                            if not response.simple_verify(
+                                achall.chall, achall.domain,
+                                achall.account_key.public_key()):
+                                logger.warning("Self-verify of challenge failed.")
```
kuba: Please include some kind of challenge ID in the log message - otherwise it will be hard to guess which challenge actually failed if there are multiple being solved.

Same for `http-01`. Though this logging line could be deduplicated, instead of copying it over between all difference subclass branches.

letsencrypt/auth_handler.py
```diff
@@ -113,6 +113,18 @@ def _solve_challenges(self):
...
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        if isinstance(response, challenges.HTTP01Response):
+                            if not response.simple_verify(
+                                achall.chall, achall.domain,
+                                achall.account_key.public_key(), response.PORT):
+                                logger.warning("Self-verify of challenge failed.")
+                        elif isinstance(response, challenges.TLSSNI01Response):
+                            if not response.simple_verify(
+                                achall.chall, achall.domain,
+                                achall.account_key.public_key()):
```
kuba: you should pass `port=self.config.tls_sni_01_port` for the challenge

letsencrypt/auth_handler.py
kuba: @pde, I feel like a robot here - would you mind turning back on relevant settings in the pylint configuration?

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
         self.assertRaises(
             errors.AuthorizationError, self.handler.get_authorizations, ["0"])
 
+    @mock.patch("acme.challenges.HTTP01Response.simple_verify")
+    @mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
+    def test_simple_verify(self, mock_http01_verify, mock_tlssni01_verify):
+        from acme import jose
+        from letsencrypt.tests import test_util
```
kuba: import both things at the top...

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+        responses = responses_http + responses_tls
+
+        self.handler.dv_c = dv_c_http + dv_c_tls
+        self.mock_dv_auth.perform.side_effect = [responses]
+        self.handler._solve_challenges()
+
+        mock_http01_verify.return_value = False
+        for achall, response in zip(dv_c_http, responses_http):
+            response.simple_verify.assert_called_with(achall.chall,
+                                                           achall.domain,
```
kuba: broken indentation

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+        responses_tls = gen_tls_resp(dv_c_tls)
+
+        responses = responses_http + responses_tls
+
+        self.handler.dv_c = dv_c_http + dv_c_tls
+        self.mock_dv_auth.perform.side_effect = [responses]
+        self.handler._solve_challenges()
+
+        mock_http01_verify.return_value = False
+        for achall, response in zip(dv_c_http, responses_http):
```
kuba: you have only one challenge under test, please remove the `for` loop and access `[0]` explicitly

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+        from letsencrypt.tests import test_util
+
+        key = jose.JWK.load(test_util.load_vector('rsa512_key.pem'))
+        http_01 = achallenges.KeyAuthorizationAnnotatedChallenge(
+            challb=acme_util.HTTP01_P, domain=b'localhost', account_key=key)
+        tls_sni_01 = achallenges.KeyAuthorizationAnnotatedChallenge(
+            challb=acme_util.TLSSNI01_P, domain=b'localhost', account_key=key)
+
+        dv_c_http = [http_01]
+        responses_http = gen_http01_resp(dv_c_http)
```
kuba: again, it's a bit of an overkill to use a function that will return a one-element response list, please just inline the responses here

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+        dv_c_tls = [tls_sni_01]
+        responses_tls = gen_tls_resp(dv_c_tls)
```
kuba: same as above

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+
+        responses = responses_http + responses_tls
+
+        self.handler.dv_c = dv_c_http + dv_c_tls
```
kuba: I suggest:
```python
self.handler.dv_c = [http_01, tls_sni_01]
```

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +148,42 @@ def test_perform_failure(self):
...
+        self.mock_dv_auth.perform.side_effect = [responses]
```
kuba: just use
```python
self.mock_dv_auth.perform.return_value = [http_01_response, tls_sni_01_response]
```

letsencrypt/tests/auth_handler_test.py
```diff
@@ -477,6 +513,16 @@ def test_different_errors_and_domains(self, mock_zope):
...
 
+def gen_http01_resp(chall_list):
+    """Returns a dummy HTTP01 response"""
+    from acme.challenges import HTTP01Response
+    return [HTTP01Response(key_authorization=u'foo')] * len(chall_list)
+
+def gen_tls_resp(chall_list):
+    """Returns a dummy TLSSNI01 response"""
+    from acme.challenges import TLSSNI01Response
+    return [TLSSNI01Response(key_authorization=u'foo')] * len(chall_list)
```
kuba: As mentioned above, I don't think these 2 functions add any real value. Just inline what's necessary.

letsencrypt/tests/auth_handler_test.py
kuba: actually if you follow my other suggestions then you will have variables `http_01` and `http_01_response` etc. which is even cleaner / more readable.

kuba:
TLS is failing in tests because you're not using `IConfig.tls_sni_01_port`.

In general, accessing protected methods from the corresponding testcase and adding local pylint ignore is OK. I'm not sure if it's the best way here, though.

letsencrypt/tests/auth_handler_test.py
TheNavigat: I followed the pattern of a few other methods in the same file, wasn't sure which is more correct.

kuba: The rule is: if it's the module that your testing then import in the function (in this file only `from letsencrypt import auth_handler`), otherwise import at the top.

letsencrypt/tests/auth_handler_test.py
TheNavigat: ```side_effect``` has precedence over ```return_value```. ```self.mock_dv_auth.perform.side_effect``` has to be set to ```None``` first, or to be removed from the ```setUp``` function. Or maybe set inside ```test_name1_tls_sni_01_1``` and ```test_name3_tls_sni_01_3_rectok_3```.

Simply changing ```side_effect``` to ```return_value``` won't work.

TheNavigat:
> TLS is failing in tests because you're not using IConfig.tls_sni_01_port.

Can you please elaborate?

.travis.yml
```diff
@@ -39,6 +39,12 @@ addons:
   # make sure simplehttp simple verification works (custom /etc/hosts)
   hosts:
     - le.wtf
+    # For boulder-integration.sh
```
kuba: This comment is misleading - `le.wtf` from above is also used in `boulder-inteagration.sh` - i would recommend to amend comment about `simplehttp` from above and add info that it also necessary for simple_verify for `http-01` and `tls-sni-01` (BTW, it shouldn't be called "simplehttp" any more).

I believe there should be a doc for developers that needs to be updated with the list of necessary mocked domains, but I don't see it being updated in this PR.

letsencrypt/auth_handler.py
```diff
@@ -42,7 +42,7 @@ class AuthHandler(object):
         form of :class:`letsencrypt.achallenges.AnnotatedChallenge`
 
     """
-    def __init__(self, dv_auth, cont_auth, acme, account):
+    def __init__(self, dv_auth, cont_auth, acme, account, config=None):
```
kuba: `config=None` should be changed to just `config` (i.e. without default value) - this class won't work without valid `IConfig` - moreover, you should update the docstring for the class

letsencrypt/auth_handler.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, dv_auth, cont_auth, acme, account):
         self.account = account
         self.authzr = dict()
 
+        self.config = config
```
kuba: nit: this is randomly placed and uses random whitespace around - please move above, under `self.acme = acme`

letsencrypt/auth_handler.py
```diff
@@ -113,6 +115,22 @@ def _solve_challenges(self):
...
+                                    achall.chall, achall.domain,
+                                    achall.account_key.public_key(),
+                                    self.config.http01_port)
+                        elif isinstance(response, challenges.TLSSNI01Response):
+                            response.simple_verify(
+                                    achall.chall, achall.domain,
+                                    achall.account_key.public_key())
+                        if not simple_verify_return:
+                            logger.warning("Self-verify of %s "
+                                    "challenge for domain %s failed.",
```
kuba: - indentation is broken
- can we please refer to the `achall.uri` in the log message? there might be multiple challenges of the same type...

letsencrypt/auth_handler.py
kuba: also, I would actually move it to be the first argument

letsencrypt/tests/auth_handler_test.py
```diff
@@ -7,14 +7,17 @@
...
 from acme import client as acme_client
+from acme import jose
 from acme import messages
 
 from letsencrypt import achallenges
 from letsencrypt import errors
 from letsencrypt import le_util
 
-from letsencrypt.tests import acme_util
+from letsencrypt.configuration import NamespaceConfig
```
kuba: please do `from letsencrypt import configuration`

letsencrypt/tests/auth_handler_test.py
```diff
@@ -80,9 +83,13 @@ def setUp(self):
         self.mock_account = mock.Mock(key=le_util.Key("file_path", "PEM"))
         self.mock_net = mock.MagicMock(spec=acme_client.Client)
 
+        namespace = mock.MagicMock(
+            tls_sni_01_port=1234, http01_port=4321)
+        self.config = NamespaceConfig(namespace)
```
kuba: it's not necessary to wrap `MagicMock` in `NamespaceConfig` - please remove

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +155,37 @@ def test_perform_failure(self):
...
+            challb=acme_util.TLSSNI01_P, domain=b'localhost', account_key=key)
+
+        mock_http01_verify.return_value = False
+        mock_tlssni01_verify.return_value = False
+
+        response_http_01 = challenges.HTTP01Response(key_authorization=u'foo')
+        response_tls_sni_01 = challenges.TLSSNI01Response(key_authorization=u'foo')
+
+        self.handler.dv_c = [challenge_http_01, challenge_tls_sni_01]
+        self.mock_dv_auth.perform.side_effect = [[response_http_01, response_tls_sni_01]]
```
kuba: I still don't understand why you cannot use `.return_value = [response_http_01, response_tls_sni_01]`...

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +155,37 @@ def test_perform_failure(self):
...
+        response_http_01.simple_verify.assert_called_with(
+            challenge_http_01.chall,
+            challenge_http_01.domain,
+            challenge_http_01.account_key.public_key(),
+            self.config.http01_port)
+
+        response_tls_sni_01.simple_verify.assert_called_with(
+            challenge_tls_sni_01.chall,
+            challenge_tls_sni_01.domain,
+            challenge_tls_sni_01.account_key.public_key()) #None accounts for cert
```
kuba: `#None ...` comments seems out of place here - should be removed

letsencrypt/auth_handler.py
```diff
@@ -113,6 +115,22 @@ def _solve_challenges(self):
...
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        simple_verify_return = True
+                        if isinstance(response, challenges.HTTP01Response):
+                            simple_verify_return = response.simple_verify(
+                                    achall.chall, achall.domain,
+                                    achall.account_key.public_key(),
+                                    self.config.http01_port)
+                        elif isinstance(response, challenges.TLSSNI01Response):
+                            response.simple_verify(
```
kuba: You don't set `simple_verify_return` here... It's a bit disappointing that tests pass despite such bug - please also amend them to make sure such problems are caught

letsencrypt/auth_handler.py
```diff
@@ -113,6 +115,22 @@ def _solve_challenges(self):
...
+                                    achall.chall, achall.domain,
+                                    achall.account_key.public_key())
```
kuba: You don't seem to use `self.config.tls_sni_01_port` which doesn't sound good..

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +155,37 @@ def test_perform_failure(self):
         self.assertRaises(
             errors.AuthorizationError, self.handler.get_authorizations, ["0"])
 
+    @mock.patch("acme.challenges.HTTP01Response.simple_verify")
+    @mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
+    def test_simple_verify(self, mock_http01_verify, mock_tlssni01_verify):
+        # pylint: disable=protected-access
+        key = jose.JWK.load(test_util.load_vector('rsa512_key.pem'))
```
kuba: I don't see why would you need to load an actual key if you mock the calls to `simple_verify`...

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +155,37 @@ def test_perform_failure(self):
         self.assertRaises(
             errors.AuthorizationError, self.handler.get_authorizations, ["0"])
 
+    @mock.patch("acme.challenges.HTTP01Response.simple_verify")
+    @mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
+    def test_simple_verify(self, mock_http01_verify, mock_tlssni01_verify):
+        # pylint: disable=protected-access
```
kuba: which protected method are you actually using? the location of this comment is not very helpful - please move it next to the actually call

letsencrypt/auth_handler.py
TheNavigat: What is the guidelines by which arguments should be ordered ordered?

letsencrypt/tests/auth_handler_test.py
TheNavigat: I think we can remove the entire import line if ```self.config``` will work if it's simply mocked.

letsencrypt/tests/auth_handler_test.py
TheNavigat: https://github.com/letsencrypt/letsencrypt/pull/1876#discussion_r48681156

letsencrypt/auth_handler.py
TheNavigat: The tests pass because we don't assert whether logs are printed or not, we assert the calls to ```simple_verify```. Should both be done?

letsencrypt/tests/auth_handler_test.py
TheNavigat: ```self.handler._solve_challenges()```

letsencrypt/auth_handler.py
TheNavigat: Also, I avoided doing that because many tests call ```AuthHandler``` without ```config```, which would be broken by requiring ```config```. Should I go on with that change and fix the tests?

TheNavigat:
https://github.com/letsencrypt/letsencrypt/pull/1876#discussion_r49012201

(Just commenting this here to make sure I don't miss it)

letsencrypt/tests/auth_handler_test.py
kuba: ACK, replying over there

letsencrypt/tests/auth_handler_test.py
kuba: Ah, `self.mock_dv_auth.perform.side_effect` is set somewhere else, in `setUp`, before this point is reached. Thanks for clarification.

letsencrypt/tests/auth_handler_test.py
TheNavigat: This seems a little bit too confusing for me. The calls to ```simple_verify``` are not mocked.

letsencrypt/tests/auth_handler_test.py
kuba: Still, please move the comment next to the `_solve_challenges` line, or even make it `self.handler._solve_challenges()  # pylint: disable=protected-access` if it is under 80 chars.

letsencrypt/tests/auth_handler_test.py
kuba: You mock the `simple_verify`...

```python
@mock.patch("acme.challenges.HTTP01Response.simple_verify")
@mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
```

letsencrypt/tests/auth_handler_test.py
TheNavigat: Got it, will do

letsencrypt/tests/auth_handler_test.py
TheNavigat: Oh, I see.

I'll read a bit about how mock.patch works and then add another comment, or replace the key with a dummy one.

Thanks for pointing it out! :)

TheNavigat:
(Posting those here to make sure I don't miss them as well)

https://github.com/letsencrypt/letsencrypt/pull/1876#discussion_r49011076

Docs should be updated

https://github.com/letsencrypt/letsencrypt/pull/1876#discussion_r49012201

Tests should be added to make sure ```logger.warning``` is called

letsencrypt/auth_handler.py
kuba: You should fix all tests. Config is typically passed first everywhere else in the code.

kuba:
I'm not sure why you keep on copying the comments into maint thread. As long as you always **merge** with `master`, rather than *rebase*, you should still see old comments here, collapsed ("kuba commented on an outdated diff 4 days ago").

TheNavigat:
I post them here again because they get collapsed, and I tend to have a bad memory :(

TheNavigat:
I'm having a little bit of difficulty implementing tests for the logger.

I've followed the ```manual_test.py``` example here, https://github.com/letsencrypt/letsencrypt/commit/23ec8ea3a4140de8a8b676dc82faa5b9c7102b1d#diff-93f7d0b7aefce1c7d9d763588f67f80dR64, as the way ```logger``` is used in both flies is quite close, if I understand correctly.

However, adding the following to ```auth_test.py``` always fails. I even tried adding ```logger.warning("test")``` to ```_solve_challenges``` method just under ```dv_resp = []```, and the test still failed.

```python
        mock_http01_verify.return_value = False
        mock_tlssni01_verify.return_value = False
        with mock.patch("letsencrypt.plugins.manual.logger") as mock_logger:
            self.handler._solve_challenges() # pylint: disable=protected-access

            response_http_01.simple_verify.assert_called_with(
                challenge_http_01.chall,
                challenge_http_01.domain,
                challenge_http_01.account_key.public_key(),
                self.config.http01_port)

            response_tls_sni_01.simple_verify.assert_called_with(
                challenge_tls_sni_01.chall,
                challenge_tls_sni_01.domain,
                challenge_tls_sni_01.account_key.public_key(),
                port=self.config.tls_sni_01_port)

            mock_logger.warning.assert_called_with("Self-verify of %s "
                                                   "challenge for domain %s failed. "
                                                   "Challenge URI: %s",
                                                   challenge_http_01.chall.__class__.__name__,
                                                   challenge_http_01.domain,
                                                   challenge_http_01.uri)

            mock_logger.warning.assert_called_with("Self-verify of %s "
                                                   "challenge for domain %s failed. "
                                                   "Challenge URI: %s",
                                                   challenge_tls_sni_01.chall.__class__.__name__,
                                                   challenge_tls_sni_01.domain,
                                                   challenge_tls_sni_01.uri)
```

The assertion fails with "Not Called".

I tried my best to figure this one out, but I'm just stuck :/

kuba:
Well, you're mocking the wrong thing: `mock.patch("letsencrypt.plugins.manual.logger")` instead of `mock.patch("letsencrypt.plugins.auth_handler.logger")`

letsencrypt/auth_handler.py
```diff
@@ -113,6 +114,26 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        simple_verify_return = True
```
kuba: BTW, it feels like all this stuff within `for` block should be moved to a separate function: `def simple_verify(self, achall, response)`. Testing should be easier.

TheNavigat: Sounds like a better idea than the current state.

Will do that.

TheNavigat:
I reviewed the code twice and took a quick look at Travis's log, everything looks fine from perfectly fine from here.

@kuba Can you please take one more look?

TheNavigat:
Ping @kuba

.travis.yml
```diff
@@ -36,9 +36,14 @@ branches:
 sudo: false
 
 addons:
-  # make sure simplehttp simple verification works (custom /etc/hosts)
+  # Custom /etc/hosts required for SimpleHTTP simple verification,
```
kuba: `SimpleHTTP` is not a thing anymore, it has been renamed to `http-01`, please adjust comment

letsencrypt/auth_handler.py
```diff
@@ -42,10 +42,11 @@ class AuthHandler(object):
         form of :class:`letsencrypt.achallenges.AnnotatedChallenge`
 
     """
-    def __init__(self, dv_auth, cont_auth, acme, account):
+    def __init__(self, config, dv_auth, cont_auth, acme, account):
         self.dv_auth = dv_auth
         self.cont_auth = cont_auth
         self.acme = acme
+        self.config = config
```
kuba: move it up, order the same as in function args

letsencrypt/auth_handler.py
```diff
@@ -103,6 +104,30 @@ def _choose_challenges(self, domains):
             self.dv_c.extend(dom_dv_c)
             self.cont_c.extend(dom_cont_c)
 
+    def _simple_verify(self, achall, response):
+        """Self-verifies a response with the corresponding challenge"""
+        return_value = True
+
+        if isinstance(response, challenges.HTTP01Response):
```
kuba: I'd add `else` stmt that calls `logger.debug("Self-verify of challenge $class is not implemented, ignoring $challenge_url")

letsencrypt/auth_handler.py
```diff
@@ -103,6 +104,30 @@ def _choose_challenges(self, domains):
...
+                    achall.account_key.public_key(),
+                    self.config.http01_port)
+        elif isinstance(response, challenges.TLSSNI01Response):
+            return_value = response.simple_verify(
+                    achall.chall, achall.domain,
+                    achall.account_key.public_key(),
+                    port=self.config.tls_sni_01_port)
+        if not return_value:
+            logger.warning("Self-verify of %s "
+                           "challenge for domain %s failed. "
```
kuba: whitespace is really strange here, please make it more compact, e.g:

```python
logger.warning(
    "Self-verify of %s challenge (%s) for domain %s failed.",
    achall.chall.typ, achall.uri, achall.domain)
```

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +153,49 @@ def test_perform_failure(self):
...
+
+        self.handler.dv_c = [challenge_http_01, challenge_tls_sni_01]
+        self.mock_dv_auth.perform.side_effect = [[response_http_01, response_tls_sni_01]]
+
+        mock_http01_verify.return_value = False
+        mock_tlssni01_verify.return_value = False
+
+        self.handler._solve_challenges() # pylint: disable=protected-access
+
+        response_http_01.simple_verify.assert_called_with(
```
kuba: `assert_called_once_with` is a bit safer

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +153,49 @@ def test_perform_failure(self):
         self.assertRaises(
             errors.AuthorizationError, self.handler.get_authorizations, ["0"])
 
+    @mock.patch("acme.challenges.HTTP01Response.simple_verify")
+    @mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
+    @mock.patch("letsencrypt.auth_handler.logger")
+    def test_simple_verify(self, mock_logger, mock_tlssni01_verify, mock_http01_verify):
+        challenge_http_01 = achallenges.KeyAuthorizationAnnotatedChallenge(
+            challb=acme_util.HTTP01_P, domain=b'localhost', account_key=mock.Mock(key="mock_key"))
```
kuba: How about:
```python
account_key=mock.Mock()
challenge_http_01 = ...account_key=account_key...
challenge_tls_sni_01 = ...account_key=account_key...
```

`key="mock_key"` is not necessary I think

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +153,49 @@ def test_perform_failure(self):
...
+        response_http_01 = challenges.HTTP01Response(key_authorization=u'foo')
+        response_tls_sni_01 = challenges.TLSSNI01Response(key_authorization=u'foo')
+
+        self.handler.dv_c = [challenge_http_01, challenge_tls_sni_01]
+        self.mock_dv_auth.perform.side_effect = [[response_http_01, response_tls_sni_01]]
+
+        mock_http01_verify.return_value = False
+        mock_tlssni01_verify.return_value = False
+
+        self.handler._solve_challenges() # pylint: disable=protected-access
```
kuba: nit: there should be two spaces before `#`

letsencrypt/tests/auth_handler_test.py
```diff
@@ -167,10 +215,12 @@ def setUp(self):
         from letsencrypt.auth_handler import challb_to_achall
         from letsencrypt.auth_handler import AuthHandler
 
-        # Account and network are mocked...
+        # Account, config and network are mocked
         self.mock_net = mock.MagicMock()
+        self.config = mock.MagicMock(
```
kuba: why override `self.config` if the same has been initialized in `__init__`?

kuba: meant `setUp` not `__init__`

letsencrypt/tests/auth_handler_test.py
```diff
@@ -22,9 +22,11 @@ class ChallengeFactoryTest(unittest.TestCase):
     def setUp(self):
         from letsencrypt.auth_handler import AuthHandler
 
-        # Account is mocked...
+        # Account and config are mocked
+        self.config = mock.MagicMock(
+            tls_sni_01_port=1234, http01_port=4321)
```
kuba: `self.config` is not used anywhere in functions tested in this class, right? why not `self.config = mock.Mock()` then?

letsencrypt/tests/auth_handler_test.py
kuba: ah, this is actually a line within a different `setUp`. Okay, so I have the same comment as in https://github.com/letsencrypt/letsencrypt/pull/1876/files#r49932087. There is no point in specifying port values, just use `self.config = mock.Mock()`

TheNavigat:
Fixed stuff, reviewed my own code again (twice), and quickly skimmed Travis's logs.

@kuba Can you please check the PR again?

TheNavigat:
PIng @kuba

TheNavigat:
Another ping, @kuba

letsencrypt/auth_handler.py
```diff
@@ -103,6 +104,28 @@ def _choose_challenges(self, domains):
...
+                    self.config.http01_port)
+        elif isinstance(response, challenges.TLSSNI01Response):
+            return_value = response.simple_verify(
+                    achall.chall, achall.domain,
+                    achall.account_key.public_key(),
+                    port=self.config.tls_sni_01_port)
+        else:
+            logger.debug("Self-verify of challenge %s is not "
+                         "implemented, ignoring %s",
+                         achall.chall.__class__.__name__,
```
kuba: `achall.typ` (`achall.chall.typ`) should give `http-01`, `dns-01`, etc.

letsencrypt/auth_handler.py
```diff
@@ -103,6 +104,28 @@ def _choose_challenges(self, domains):
             self.dv_c.extend(dom_dv_c)
             self.cont_c.extend(dom_cont_c)
 
+    def _simple_verify(self, achall, response):
+        """Self-verifies a response with the corresponding challenge"""
+        return_value = True
+
+        if isinstance(response, challenges.HTTP01Response):
+            return_value = response.simple_verify(
```
kuba: `return response.simple_verify` here and below, `return True` in the `else` branch

letsencrypt/auth_handler.py
```diff
@@ -113,6 +136,14 @@ def _solve_challenges(self):
                     cont_resp = self.cont_auth.perform(self.cont_c)
                 if self.dv_c:
                     dv_resp = self.dv_auth.perform(self.dv_c)
+                    for achall, response in zip(self.dv_c, dv_resp):
+                        if not self._simple_verify(achall, response):
+                            logger.warning("Self-verify of %s challenge (%s) "
+                                           "for domain %s failed.",
+                                           achall.chall.__class__.__name__,
+                                           achall.domain,
+                                           achall.uri)
```
kuba: Please make the line wrapping optimally compact by:
- putting more arguments on the same line, e.g. `achall.domain, achall.uri` instead of `achall.domain,\nachall.uri`
- you can try to put a newline after `logger.warning(` and then continue with 4 space indent.

letsencrypt/interfaces.py
```diff
@@ -169,7 +169,9 @@ def perform(achalls):
               Authenticator will never be able to perform (error).
 
         :rtype: :class:`list` of
-            :class:`acme.challenges.ChallengeResponse`
+            :class:`acme.challenges.ChallengeResponse`,
+            where responses are required to be returned
+            in the same order as challenges
```
kuba: "as corresponding input challenges" might be less confusing

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +152,63 @@ def test_perform_failure(self):
...
+            challenge_http_01.account_key.public_key(),
+            self.config.http01_port)
+
+        response_tls_sni_01.simple_verify.assert_called_once_with(
+            challenge_tls_sni_01.chall,
+            challenge_tls_sni_01.domain,
+            challenge_tls_sni_01.account_key.public_key(),
+            port=self.config.tls_sni_01_port)
+
+        logger_calls = [mock.call(mock.ANY,
```
kuba: Again, line wrapping is quite bad here, please add a newline after opening `(` to make it more readable

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +152,63 @@ def test_perform_failure(self):
...
+                                  challenge_http_01.domain,
+                                  challenge_http_01.uri),
+                        mock.call(mock.ANY,
+                                  challenge_tls_sni_01.chall.__class__.__name__,
+                                  challenge_tls_sni_01.domain,
+                                  challenge_tls_sni_01.uri)]
+
+        mock_logger.warning.assert_has_calls(logger_calls)
+
+        mock_logger.debug.assert_called_with(mock.ANY,
```
kuba: wrapping

kuba:
Sorry, @TheNavigat, but my time resources have been severely reduced recently. I didn't find anything obviously wrong with the code (apart from style issues), but please find another reviewer before getting this merged - I'm not able to give you proper feedback in foreseeable future and this PR has been rotting in the tracker forever already...

TheNavigat:
I noticed. Thanks for your help so far, Kuba :)

@swartzcr, @bmw, @schoen: Can one of you please review this PR?

schoen:
@TheNavigat, thanks for doing this! I would like to take a look at this after the 0.4.0 release is out; could you please remind me about that?

TheNavigat:
[needs another reviewer] tag removed. Thanks @schoen, will do!

TheNavigat:
@schoen Ping :D

moshevds:
Hi,

I'm looking into simply_verify because it touches my PR #2461.

First of all, I think it is good that the responsibility to call response.simple_verify() is moved away from the plugins with this patch.

I have one comment on the code:
Would it be possible to remove the use of isinstance() to check the challenge response type? This breaks polymorphism while the only difference is the port argument. I would also note that TLSSNI01Response does not use the port argument at all.

However, I also have a more fundamental question about self-verification:
The nature the acme protocol is that it is up to the server to verify if a challenge succeeds, and it is sometimes difficult (or impossible) for the client to know that verdict in advance. https://github.com/letsencrypt/letsencrypt/issues/1409#issuecomment-157308493 is related to a case of self-verification being too strict, but the other way around is also possible. Self-verification can also easily fail or pass incorrectly because of local network configuration. And in the case of dns-01, the server verification checks DNSSEC, but the user might not have a validating resolver available for the self-verification to use. How should these issues be dealt with?

TheNavigat:
> Would it be possible to remove the use of isinstance() to check the challenge response type? This breaks polymorphism while the only difference is the port argument. I would also note that TLSSNI01Response does not use the port argument at all.

+1. ``port`` is set to ``None`` by default for ``HTTP01Response``. Using the ``port=`` bit to send the parameter makes the code more consistent. I'll try to implement that, hopefully I won't have to face any walls in my way.

>However, I also have a more fundamental question about self-verification:
>The nature the acme protocol is that it is up to the server to verify if a challenge succeeds, and it is sometimes difficult (or impossible) for the client to know that verdict in advance. #1409 (comment) is related to a case of self-verification being too strict, but the other way around is also possible. Self-verification can also easily fail or pass incorrectly because of local network configuration. And in the case of dns-01, the server verification checks DNSSEC, but the user might not have a validating resolver available for the self-verification to use. How should these issues be dealt with?

I don't possess enough knowledge to have an opinion of value, because I can only understand the tip of the iceberg when it comes to how verifying challenges work. If I have to say something, is there a way to make sure the client knows that verdict in advance? If so, we should pursue that. If not, why?

moshevds:
I see this change as an improvement, and would like to advocate for it. But it might need some more review and although I am new to Let's Encrypt, perhaps I can give some more suggestions.

First, some included changes seem not directly related:
 - 74982a65 (and the related change in 1a755e771) can stand on its own as a documentation improvement, right? If so, you could create a separate pull request for it.
 - The same goes for that /etc/hosts stuff in 1d441ae0, 98f2d329 and f59dcc4b. All those commits have other changes also, so separating would require some fiddling, but likely worth it as a pull request on its own.

I think both might have originated from the discussion here, but that's no reason to contain them to this PR and it might speed up reviews.

I have also spotted some issues that I will comment on in the diff.

letsencrypt/plugins/manual_test.py
```diff
@@ -54,17 +54,10 @@ def test_perform(self, mock_raw_input, mock_verify, mock_stdout, mock_interactio
         resp = self.achalls[0].response(KEY)
         self.assertEqual([resp], self.auth.perform(self.achalls))
         self.assertEqual(1, mock_raw_input.call_count)
-        mock_verify.assert_called_with(
-            self.achalls[0].challb.chall, "foo.com", KEY.public_key(), 8080)
```
moshevds: I think mock_verify is no longer used an can be removed from the test

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +152,66 @@ def test_perform_failure(self):
         self.assertRaises(
             errors.AuthorizationError, self.handler.get_authorizations, ["0"])
 
+    @mock.patch("acme.challenges.HTTP01Response.simple_verify")
+    @mock.patch("acme.challenges.TLSSNI01Response.simple_verify")
+    @mock.patch("letsencrypt.auth_handler.logger")
+    def test_simple_verify(self, mock_logger, mock_tlssni01_verify, mock_http01_verify):
```
moshevds: This test is very long and seems to do multiple things. Each path of execution is touched by this only test, but usually it is much better to have one test for each path. That way one test tests one thing.

letsencrypt/tests/auth_handler_test.py
```diff
@@ -148,6 +152,66 @@ def test_perform_failure(self):
...
+        self.handler.dv_c = [challenge_http_01,
+                             challenge_tls_sni_01,
+                             challenge_dummy]
+
+        self.mock_dv_auth.perform.side_effect = [[response_http_01,
+                                                  response_tls_sni_01,
+                                                  response_dummy]]
+
+        mock_http01_verify.return_value = False
+        mock_tlssni01_verify.return_value = False
```
moshevds: This test only tests failure for these challenges. This is poor for coverage, both success and failure branches should be covered.

letsencrypt/plugins/manual_test.py
TheNavigat: It is removed from manual_test.py already, which is indicated by the ``-`` sign. The whole PR addresses Kuba's comment, https://github.com/letsencrypt/letsencrypt/pull/1876#issuecomment-164149879, which indicates that instead of testing on the plugin level, we should test on a more generic one.

letsencrypt/tests/auth_handler_test.py
TheNavigat: That's not how tests work, actually.

Each test function is called only once. The execution of ``simple_verify`` does not trigger this test.

letsencrypt/tests/auth_handler_test.py
TheNavigat: Well, that is both true and false.

If ``simple_verify`` returns ``True``, there's nothing we can solidly assert. Maybe insure that logger isn't called, but that would be counter-productive. If someone added some other ``logger.warning`` call that uses the same parameters, the test would fail.

I'm not currently able to think of a better idea.

TheNavigat:
>74982a6 (and the related change in 1a755e7) can stand on its own as a documentation improvement, right? If so, you could create a separate pull request for it.

For https://github.com/letsencrypt/letsencrypt/commit/74982a655266d111f19e8226ccbac07775a58056, that is true, this can be possibly separated into its own PR. However, I'd argue that it's too trivial, so it wouldn't actually speed up the review that much.

https://github.com/letsencrypt/letsencrypt/commit/1a755e771644f5413a2742a85f23418471e3d6d4 is updates made to code written in the PR itself, so that one is relevant. Updated styling is for code that hasn't been there in prior, but one that I have written in the context of this PR.

>The same goes for that /etc/hosts stuff in 1d441ae, 98f2d32 and f59dcc4. All those commits have other changes also, so separating would require some fiddling, but likely worth it as a pull request on its own.

https://github.com/letsencrypt/letsencrypt/commit/1d441ae056e611031b5d76b0f50f6f671f5d0565 is related, too. ``simple_verify`` itself requires the domains to be owned by the user so that a false negative isn't triggered. Previously, ``simple_verify`` wasn't consistently tested, and that's why we didn't really need these domains to be added to ``/etc/hosts``.

letsencrypt/tests/auth_handler_test.py
TheNavigat: Another reason is that the code tests the functionality of ``simple_verify`` itself, not how other functions react to its return values, so I think we're fine with not including the case where it returns ``True``.

letsencrypt/plugins/manual_test.py
moshevds: I meant the variable mock_verify that you removed here. Some code for it is still on line 48, 50 and 51, right? Those can now be removed

letsencrypt/tests/auth_handler_test.py
moshevds: I think you misunderstood my comment. A unit-test is a safeguard to check that the code you wrote does what you meant to do and keeps doing that. Usually this means having one test for each "path of code branches" (or "path of execution") and asserting that it did the right thing. This test triggers essentially all paths in one go and asserts the result of that.

That means it is not testing much, consider this implementation:

```python
    def _solve_challenges(self):
        """Get Responses for challenges from authenticators."""
        cont_resp = []
        dv_resp = []
        with error_handler.ErrorHandler(self._cleanup_challenges):
            try:
                if self.cont_c:
                    cont_resp = self.cont_auth.perform(self.cont_c)
                if self.dv_c:
                    dv_resp = self.dv_auth.perform(self.dv_c)

            except errors.AuthorizationError:
                logger.critical("Failure in setting up challenges.")
                logger.info("Attempting to clean up outstanding challenges...")
                raise

        if len(dv_resp) == 3 and hasattr(dv_resp[0], 'simple_verify'):
            dv_resp[0].simple_verify(self.dv_c[0].chall, self.dv_c[0].domain,
                    self.dv_c[0].account_key.public_key(), self.config.http01_port)
            dv_resp[1].simple_verify(self.dv_c[1].chall, self.dv_c[1].domain,
                    self.dv_c[1].account_key.public_key(), port=self.config.tls_sni_01_port)
            logger.warning("", "HTTP01", "localhost", "http-01_uri")
            logger.warning("", "TLSSNI01", "localhost", "tls-sni-01_uri")
            logger.debug("", self.dv_c[2].typ, self.dv_c[2].uri)

        assert len(cont_resp) == len(self.cont_c)
        assert len(dv_resp) == len(self.dv_c)

        return cont_resp, dv_resp
```

This bogus code passes unit testing. Mostly because you only have one test and also because you're not at all asserting if the warning comes from verification failure or that each challenge is verified.
A good way would be to create multiple tests each testing one thing the code has to do.

letsencrypt/plugins/manual_test.py
TheNavigat: Ah, good catch! :) Thanks, removed the variable.

letsencrypt/tests/auth_handler_test.py
TheNavigat: /me is confused.

I agree that this test basically "touches" parts of the execution and ensures that it is ran correctly. That is because ``simple_verify`` itself doesn't alter the response of other functions using it, including ``_solve_challenges`` (I'm not sure if it should). Thus, I can't understand how we can have an implementation where everything is tested in a single run. Issue https://github.com/letsencrypt/letsencrypt/issues/1409 called for the change from stopping the client on ``simple_verify``'s failure to simply logging a warning.

The sample code you've used would be extremely difficult to exist by mere mistake. Conspicuous intent would be required to fool the tests for something like that to exist. In that sense, most tests can be fooled one way or another, and the only way we can stop that is by having detailed reviews before a PR is merged.

letsencrypt/tests/auth_handler_test.py
moshevds: My bogus implementation above might also clarify this a bit hopefully. Success or failure of the verify call is the most important part of this code so it makes sense to test both situations.

You mention that if someone adds a new `logger.warning` it might break if you test the successful verification. This is true and should be avoided by checking for absence of the warning about verification only.

You also mention that the functionality of `simple_verify` is tested but I assume you mean `_simple_verify`? That is (partly) what should be tested. But your changes to `_solve_challenges` should also be tested and right now you are not really testing the return value of `_simple_verify`.

Lets look at the code under test again:
```python
                    for achall, response in zip(self.dv_c, dv_resp):
                        if not self._simple_verify(achall, response):
                            logger.warning(
                                "Self-verify of %s challenge (%s) "
                                "for domain %s failed.",
                                achall.chall.__class__.__name__,
                                achall.domain, achall.uri)
```
This is equivalent to:
```python
                    for achall, response in zip(self.dv_c, dv_resp):
                        retval = self._simple_verify(achall, response)
                        if not retval:
                            logger.warning(
                                "Self-verify of %s challenge (%s) "
                                "for domain %s failed.",
                                achall.chall.__class__.__name__,
                                achall.domain, achall.uri)
```

In this last bit, the line `if not retval:` is at the core of how the code is supposed to work, but it is not properly tested. It can be replaced with many other lines that are wrong but pass this test, for example `if achall.domain == 'localhost':` or `if hasattr(response, 'simple_verify'):` etc.

TheNavigat: Ah, you're completely right regarding the fact that we don't test the return value of ``_simple_verify`` (and yes I meant ``_simple_verify`` above as well). I'll work on that.

I understand the other problems you've mentioned but I'm not sure what would be the best way to fix such problems. I really appreciate the code snippets, though, really helped to resolve the confusion :)

letsencrypt/tests/auth_handler_test.py
moshevds: Not everything can be tested in a single run. That's exactly why I think you need separate tests.

The bogus code I wrote can't really exist by mistake, but it is an extreme to show everything that isn't tested currently. I wasn't trying to fool the test inconspicuously, I was trying to fool the test easily. A better test is probably harder to fool (and thus also harder to fool unintentionally). I'm also not saying every possibility has to be tested, but that it is desirable to better safeguard the intended behavior here.

For example you could create the following tests:
 - **test_simple_verify_one_success_http01** Call the code with 1 mock_http01 that returns True. Assert that it does not log the warning.
 - **test_simple_verify_one_success_tlssni01** Call the code with 1 mock_tlssni01 that returns True. Assert that it does not log the warning.
 - **test_simple_verify_one_failure** Call the code with 1 mock that returns False. Assert that it logs the warning.
 - **test_simple_verify_one_unimplemented** Call the code with the dummy. Assert that it logs the debug message.
 - **test_simple_verify_many** Call the code with multiple challenges. Assert that the correct number of calls are made.

That's a lot of tests but this is only an example and more or less tests could also be fine. The above tests could be short and simple and hopefully easy to read. If I would know the Lets Encrypt code a bit better I would be able te better predict what @schoen or @kuba is going to expect :-)

TheNavigat: Makes sense :) However,

I'm thinking we should rename ``test_simple_verify`` to ``test_solve_challenges`` and then create a test that sends all those examples to ``_simple_verify`` and asserts the return value.

Afterwards we can possibly alter the ``_solve_challenges`` to test fewer cases, where one of them logs a warning, and another doesn't.

Does it seem logical for you?

moshevds: It would indeed be easier that way. This would also mean you're going to mock internal methods (starting with _), and that normally impairs flexibility and refactorability, but is probably okay.

TheNavigat: "but it is probably okay" :D

moshevds:
About the two PRs I suggested, I have PRed them to your fork with the git commands I used to create them as the comment.

They are indeed trivial themselves, but I hope it'll make reviewing the changes for this PR a tiny bit easier and therefore merged faster. Not sure if it helps but I don't see it hurting either.

Here they are: https://github.com/TheNavigat/letsencrypt/pull/4 https://github.com/TheNavigat/letsencrypt/pull/5

TheNavigat:
Ummmm, why to my fork? :/

TheNavigat:
Closing per https://github.com/letsencrypt/letsencrypt/issues/1586#issuecomment-186020863

PR #1878, opened 2015-12-12
Echoing error and exiting on failure of apt-get

TheNavigat:
Fixes https://github.com/letsencrypt/letsencrypt/issues/1786


bootstrap/_deb_common.sh
```diff
@@ -44,6 +44,10 @@ apt-get install -y --no-install-recommends \
   libffi-dev \
   ca-certificates \
 
+if [[ $? > 0 ]] ; then
```
pde: Where is the closing branch of this "if" statement?

pde: Does this cause an exit in the case where one of the packages was simply not available on some Debian variant?  If it does, we will need to be extremely careful about the change, because we've seen incredible diversity in what some packages are called, and erroring out because of it will require a lot more careful logic to figure out the right names for virtualenv and python packages especially.

TheNavigat: Sorry about the closing branch, I tend to make some silly mistakes from time to time.

And yes it does. I think that should be the correct behaviour though since these packages are supposedly required packages, and failure to install them should exit the installation.

TheNavigat:
Ping @pde

TheNavigat:
@pde Another ping

TheNavigat:
Per IRC chat,

```
03:31 <+pdeee> the other problem is, all changes to letsencrypt-auto are going to be made obsolete by #1665
03:32 <+pdeee> so really we're going to have to do a sweep of all the PRs, and decide which ones make sense and need to be ported to the new version
03:33 < TheNavigat> pdeee: Oh :/
03:34 < TheNavigat> Okay, guess I should close the PR, then?
03:54 <+pdeee> probably
03:54 <+pdeee> we're going to have to close, port, or ask submitters to port a lot of letsencrypt-auto PRs :(
```

Thus putting this PR on hold until further notice.

PR #1879, opened 2015-12-12
Fixed an inaccurate comment

alex:
While it's true that older Pythons do not do (critical) TLS validation by default, that's not what this warning is about.


kuba:
LGTM, thanks

PR #1881, opened 2015-12-12
Fixed a type in a docstring

alex:


alex:
You can see the misrendering this causes here: https://letsencrypt.readthedocs.org/projects/acme/en/latest/api/client.html#acme.client.Client.poll_and_request_issuance

kuba:
LGTM, thanks

PR #1887, opened 2015-12-13
Update Debian using instructions

hlieberman:
This updates the Debian using instructions to reflect the upload of Debian packages to sid.


Nightfirecat:
Slight nitpick on the commit name, but it should probably read "Update Debian *usage* instructions."

pde:
Merging, though I might tweak the wording to say "Debian unstable".

hlieberman:
Yeah, I did that intentionally because it's going to be in testing soon.

PR #1901, opened 2015-12-15
Support for Synology-like systems with busybox [revision requested]

kenkendk:
Added support for detecting system where "command -v" fails, but "which" works instead (i.e. Synology)


pde:
This is not sufficient to support busybox's non-POSIX `sh`.  You will need to replace all instances of `VAR={$VAR:-$DEFAULTVALUE}` with something like `if [ "$VAR" = "" ] ; then VAR="$DEFAULTVALUE" ; fi`.  Please work on top of the pending changes in #1665 if you're planning to attempt that.

kenkendk:
I guess that depends on what you need to support. 
The current version works on Synology, but if you would like a version that does not use the `{$VAR:-$DEFAULT}` syntax, I can make that.

pde:
@kenkendk thanks.  In my testing `busybox sh` would mangle environment variables when it hit those expressions.

pde:
We will review and test this after #1665 has landed.

erikrose:
#1665 has now landed, so this is ready to be rebased on the latest master. There will also be some porting working involved.

kenkendk:
I have re-applied the patch.
The space-in-path handling fixes were part of master, so I just added the `which` check and the `VarOrDefault` from my patch.

I have not tested it on my box yet, will do asap.

kenkendk:
I have tested now, and it works on my box.

The Travis error is unrelated to the fix, it has a 500 error from the server in one of the build configs. If you can resubmit the job, I am sure it will pass. I can also make a bogus commit to trigger the Travis if you prefer.

bmw:
I'm sorry this sat around so long @kenkendk. If you want to bring this PR up to date so that it works with busybox, I'll happily review.

kenkendk:
Ok, I will close this and create a new PR where all the history is gone, as it is really a simple fix.

PR #1906, opened 2015-12-15
Revert "Add staging server hint to avoid rate limit issues"

bmw:
Reverts letsencrypt/letsencrypt#1848


bmw:
Ugh. I forgot that `--test-cert` hasn't landed in PyPI and merged this. Will remerge after `0.1.1` lands.

PR #1907, opened 2015-12-16
Candidate 0.1.1

pde:
0.1.1 has been released; this branch contains its tag (as well as finalised release scripts)


tools/release.sh
```diff
@@ -97,16 +135,44 @@ pip install \
...
+done
+deactivate
+
+cd ..
+echo Now in $PWD
+name=${root_without_le%.*}
+ext="${root_without_le##*.}"
+rev="$(git rev-parse --short HEAD)"
+echo tar cJvf $name.$rev.tar.xz $name.$rev
+echo gpg -U $RELEASE_GPG_KEY --detach-sign --armor $name.$rev.tar.xz
```
bmw: These two commands are the ones that are automated by `dev-release2.sh`, but not this script.

tools/release.sh
```diff
@@ -1,13 +1,43 @@
-#!/bin/sh -xe
+#!/bin/bash -xe
 # Release dev packages to PyPI
 
+Usage() {
+    echo Usage:
+    echo "$0 [ --production ]"
```
bmw: I think we should explain what that if `--production` is included, you must also specify the current and next versions, but we obviously don't have to block this PR on that.

bmw:
This also doesn't include the tag (see [this](https://stackoverflow.com/questions/12278660/adding-tags-to-a-pull-request)).

Also, is this supposed to include the version bump or is that in another PR?

pde:
I accidentally pushed this branch from a working checkout, rather than the `releases/le-0.1.1.XX/` subdirectory.  Correcting that now gets us the signed release and version bump; I've also pushed the tag.

bmw:
Looking at the changes, it looks like a version bump wasn't even necessary. The version was already set at `0.2.0.dev0`.

It also looks like the changes made to the version number by the release scripts are incorrect.

kuba:
I would like to note that "this branch contains its tag" statement doesn't make much sense. Both branches and tags are git refs. You could say that this branch contains a commit happens to be tagged, but pulling in the branch doesn't mean that the tag is pulled in. They are separate things!

acme/setup.py
```diff
@@ -4,7 +4,7 @@
 from setuptools import find_packages
 
 
-version = '0.2.0.dev0'
+version = '0.2.0dev0'
```
kuba: why do you remove the dot? `version` has been kept so far in canonical form, with dot

letshelp-letsencrypt/setup.py
```diff
@@ -4,7 +4,7 @@
 from setuptools import find_packages
 
 
-version = '0.2.0.dev0'
+version = '0.2.0'
```
kuba: that's wrong!

tools/half-sign.c
```diff
@@ -0,0 +1,117 @@
+#include <unistd.h>
```
kuba: oh wow, when did we start using C? Where is a doc for this file? what does it do?

tools/release.sh
```diff
@@ -1,13 +1,43 @@
-#!/bin/sh -xe
+#!/bin/bash -xe
```
kuba: why change from POSIX to bash-only?

pde: To allow the use of bash string editing features to keep our release script as sane as possible.

acme/setup.py
pde: Thanks for catching that.  Human error!

tools/half-sign.c
pde: Explanation added. This isn't in use yet, but we're likely to add airgapped signatures from that code in the future.

bmw:
LGTM

tools/half-sign.c
kuba: if it's not used anywhere then remove it...

PR #1913, opened 2015-12-16
Check an enhancement is supported before applying (fixes #1432)

### LINKED ISSUE #1432: Asks about HTTPS redirect despite installer not supporting it
Hi,

I'm writing an installer plugin that will not support enhancements, and thus returns none:

    def supported_enhancements(self):
        return []

Nevertheless, on letsencrypt install (with this plugin), I get the choice “Please choose whether HTTPS access is required or optional.”, and then I can choose between Easy or Secure. Shouldn't it check with the installer if it can do the “redirect” enhancement before it asks? It just calls enhance() with no more ifs and buts, which fails.

xgin:


bmw:
LGTM. Thanks @xgin.

PR #1914, opened 2015-12-16
Guarantee a true SSLContext object with Python 2

Natim:


setup.py
```diff
@@ -55,6 +55,10 @@ def read_file(filename, encoding='utf8'):
         'argparse',
         'mock<1.1.0',
     ])
+elif sys.version_info < (2, 8):
```
pde: Shouldn't this be `sys.version_info < (2, 7, 9)`?

https://urllib3.readthedocs.org/en/latest/security.html#installing-urllib3-with-sni-support-and-certificates

pde:
Closes: #1926

pde:
You should also move this change into `acme/setup.py`, which already depends on `python-ndg-httpsclient`.

Natim:
> You should also move this change into acme/setup.py, which already depends on python-ndg-httpsclient.

Actually acme seems to already have all the dependencies.
Is it supposed to be installed when running ``letsencrypt-auto``?
In that case why running ``letsencrypt-auto`` still display ``InvalidProtocolWarnings``?

Natim:
I leaved it in both places because both places have got the requests dependency.

pde:
Integration test heisenbug.  Kicking Travis.

Natim:
The bug on py26 seems real I will have a look tomorrow.

acme/setup.py
```diff
@@ -29,6 +27,10 @@
         'argparse',
         'mock<1.1.0',
     ])
+elif sys.version_info < (2, 7, 9):
```
kuba: that matches Python 2.6, so the previously existing code for py2.6 is never run!

Natim:
Seems to be fixed :) Thank you for your feedbacks.

Natim:
I am wondering why acme and letsencrypt have so closed setup.py files? Shouldn't one depends on the other? Acme from letsencrypt or letsencrypt from acme?

pde:
@Natim there isn't necessarily a good answer to that.  Perhaps they can be rationalised.

pde:
Tested on a couple of systems, LGTM!

setup.py
```diff
@@ -58,6 +58,11 @@ def read_file(filename, encoding='utf8'):
 else:
     install_requires.append('mock')
 
+if sys.version_info < (2, 7, 9):
```
kuba: this should not have been included in this PR; `ndg-httpsclient` and `pyasn1` were not previously explicit dependencies, and I don't think they should be. @Natim, could you please revert?

kuba:
It's not true there isn't good answer for that, @pde. Dependencies are listed explicitly if they are used in a given subproject (e.g. if there is explicit `import foo` in the core package, then `foo` must be listed in main `setup.py`), even if they are already a transitive dependency. That's handy in cases where upstream dep removes one of its dependencies hence removing transitive dep.

In this case I don't think `ndg-httpsclient` or `pysan1` are explicitly necessary for the core package - they will be pulled in together with `acme` package. This leads me to belief that this PR didn't fix the problem you were trying to fix: `InvalidProtocolWarnings`. Please revert changes to main `setup.py` or provide some evidence that it changed something.

setup.py
Natim: Actually I added them because they were missing.

Since requests is a dependency here and that it is used to call HTTPS URLs we need to add this in order to use a secure communication (and remove the InvalidPlatformWarning)

Natim:
> This leads me to belief that this PR didn't fix the problem you were trying to fix: InvalidProtocolWarnings.

Actually it is taking care of by requests. I agree with you that requests should probably add this code itself not sure why they didn't do it.

setup.py
Natim: Here is the rationale about it: http://urllib3.readthedocs.org/en/latest/security.html#installing-urllib3-with-sni-support-and-certificates

kuba: They were already being pulled in `acme`, so by including them here you do not change anything!

So while still you're totally missing my point I did check it myself:

```
jakub@iks:~/dev/letsencrypt/letsencrypt/letsencrypt$ git grep 'import requests'
tests/validator_test.py:import requests
validator.py:import requests
jakub@iks:~/dev/letsencrypt/letsencrypt/letsencrypt$ grep 'requests.' validator.py
            response = requests.get(url, headers=headers, allow_redirects=False)
            response = requests.get(url, allow_redirects=False)
        headers = requests.get("https://" + name).headers
```

So I'm fine with keeping changes to the main  `setup.py`, but for reasons other than you mention. (AFAIK `validator.py` is a dead code - it's not used anywhere)

Natim: Ok let's do that.

Natim: I opened https://github.com/letsencrypt/letsencrypt/issues/1997 to address this.

PR #1921, opened 2015-12-16
Removed duplicate cancel

bmw:
As I stated before it was merged, I think the duplicate cancel when dealing with lineages is unnecessary and clutters the menu. With my changes, the menus now look like:
##### Text
```
You have an existing certificate that contains exactly the same domains you requested and isn't close to expiry.
(ref: /etc/letsencrypt/renewal/example.com.conf)

What would you like to do?
1: Keep the existing certificate for now
2: Renew & replace the cert (limit ~5 per 7 days)
Select the appropriate number [1-2] then [enter] (press 'c' to cancel):
```
##### Dialog
http://imagebin.ca/v/2QC7kYuyy1i6

If the cancel option is not prominent enough, we should correct it for all of our menus, not just this one prompt.


pde:
@schoen what do you think?

schoen:
Peter Eckersley writes:

> @schoen what do you think?

I don't have a very clear intuition about how readily users will notice
the cancel button.  I originally had the duplicate cancel option because
I wanted the do-nothing option to be parallel to the other choices, but
I can also sympathize with the criticism that it doesn't make sense to
have two different ways to cancel in the same dialog.

I'd be willing to have this change happen; it seems like something that
could benefit from user testing that we wouldn't have the luxury of
doing right now.

pde:
Good point.  Let's hallway test it :)

bmw:
Sure! User testing is a good idea.

To make this point again though, if there is a problem with our menu, we should fix it in the general case not just for this single prompt. I think these changes should go in regardless of what we find from user testing (maybe with additional changes fixing problem(s)).

bmw:
ping @schoen

I still would like this to land. Just addressed merge conflicts.

schoen:
Hi @bmw, I'm happy to merge it, but for clarity I think you need to remove the 2nd condition in the `if` statement on line 326, because that's referring to the condition you're removing from the menu!

letsencrypt/cli.py
```diff
@@ -319,8 +319,7 @@ def _handle_identical_cert_request(config, cert):
     elif config.verb == "certonly":
         keep_opt = "Keep the existing certificate for now"
     choices = [keep_opt,
-               "Renew & replace the cert (limit ~5 per 7 days)",
-               "Cancel this operation and do nothing"]
+               "Renew & replace the cert (limit ~5 per 7 days)"]
 
     display = zope.component.getUtility(interfaces.IDisplay)
     response = display.menu(question, choices, "OK", "Cancel", default=0)
```
schoen: GitHub won't let me attach a line note to line 326 because it's too far away from your change (!) but the `if` statement just below here is no longer correct after removing this item from the menu. I believe it should change to

```
    if response[0] == "cancel":
```

bmw:
Oh wow! Thanks for catching that @schoen. I actually had it correct in my initial commit but screwed it up when I merged with master.

Should be fixed now.

schoen:
Keen!

PR #1924, opened 2015-12-16
Import the letstest test farming code into the main client repo

pde:
@levskaya made a nice test farm codebase for us.  This PR uses some [historical revisionism](https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history/15135004#15135004) to pull that into `tests/letstest` while keeping its history visible.

(We could also `git mv` to its new home and then merge that, but every line of every file would have no visible history past the `git mv`).

### DIFF
--- a/tests/letstest/README.md
+++ b/tests/letstest/README.md
@@ -1,0 +1,44 @@
+# letstest
+simple aws testfarm scripts for letsencrypt client testing
+
+- Configures (canned) boulder server
+- Launches EC2 instances with a given list of AMIs for different distros
+- Copies letsencrypt repo and puts it on the instances
+- Runs letsencrypt tests (bash scripts) on all of these
+- Logs execution and success/fail for debugging
+
+## Notes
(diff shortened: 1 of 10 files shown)


bmw:
@pde, I'll be honest and say that I don't know enough `git` fanciness to know how dangerous this is. I will say, however, that I saw @kuba's request for you not to do this and even the warning given in the stack overflow post you linked. To quote the most relevant part:

> You could be rewriting a shared history, which is the most important DON'T while using Git. If someone else has cloned the repository, you'll break it doing this.

As of writing this comment, about 40k people have cloned this repo according to GitHub. I do not think preserving history is worth potentially breaking their setup if they ever do a `git` pull. I think you should consider just bringing the files in without the history.

pde:
I think in this case it isn't particularly dangerous but I also think I was wrong in fearing that `git mv` would render history prior to the move hard to find.  So I'm going to close this and reopen with a `git mv`-based  PR.

PR #1927, opened 2015-12-17
hackish tests using the Apache conf library

pde:
Not as good as true compatibility tests, but also easy to implement... we just drop the conf file into the apache conf and see what happens.


pde:
This PR is now obsolete; will open a new / clean one.

PR #1929, opened 2015-12-17
fix issue with parsing renewal confs

SwartzCr:
previously it only passed the name of the config file which prevented the following code from opening it. Now we pass the full path
split on multiple lines for linting


bmw:
LGTM. Thanks @SwartzCr

PR #1936, opened 2015-12-17
letsencrypt-auto: prevent recording of deps in Gentoo's world set

osirisinferi:
Dependencies shouldn't be recorded into Gentoos [world set](https://wiki.gentoo.org/wiki/World_set_(Portage)). The current single switch to `emerge` "--noreplace" makes sure already installed dependencies are explicitly recorded into the world set, which isn't desirable. Adding "--oneshot" prevents this behaviour.
Cave: I have absolutely no clue how Paludis works. It's [manual](http://paludis.exherbo.org//clients/cave-resolve.html) and an entry in their [news](http://paludis.exherbo.org/news.html) (0.26.0_pre3) suggests --preserve-world should do the trick.

(Credits: [till.schaefer](https://community.letsencrypt.org/t/gentoo-letsencrypt-should-not-automatically-record-packages-to-the-world-config-file/7252))

By the way: Gentoo has a [native package](https://packages.gentoo.org/packages/app-crypt/letsencrypt) since Dec 3.


osirisinferi:
Uch, GitHubs `command line instructions` for the conflict resolving went horribly wrong :sob: 

Guess that was *not* a very good way to correctly merge letsencrypt:master into my own fork's branch :disappointed:

PR #1938, opened 2015-12-17
Ability to suppress warning "Root (sudo) is required" in plugins [needs revision]

xgin:
Some plugins do not require root privileges.
The warning is misleading, e.g. https://github.com/plesk/letsencrypt-plesk/issues/35
Proposed solution is to expose is_root_required property in order to check root privileges on demand.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -549,7 +549,11 @@ def record_chosen_plugins(config, plugins, auth, inst):
 
 def check_root_privileges(plugin):
     """Warns user about insufficient privileges."""
-    if plugin.is_root_required and not os.geteuid() == 0:
+    if not hasattr(plugin, 'is_root_required'):
+        return
+    if not plugin.is_root_required:
+        return
+    if not os.geteuid() == 0:
         logger.warning(
             "The plugin %s requires root privileges. "
             "Since you are not root (not using sudo) it may encounter errors.",
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -531,6 +531,11 @@ def choose_configurator_plugins(args, config, plugins, verb):  # pylint: disable
     if need_auth and not authenticator:
         diagnose_configurator_problem("authenticator", req_auth, plugins)
 
+    if need_inst:
+        check_root_privileges(installer)
+    if need_auth:
+        check_root_privileges(authenticator)
+
     record_chosen_plugins(config, plugins, authenticator, installer)
     return installer, authenticator
 
@@ -540,6 +545,15 @@ def record_chosen_plugins(config, plugins, auth, inst):
     cn = config.namespace
     cn.authenticator = plugins.find_init(auth).name if auth else "none"
     cn.installer = plugins.find_init(inst).name if inst else "none"
+
+
+def check_root_privileges(plugin):
+    """Warns user about insufficient privileges."""
+    if plugin.is_root_required and not os.geteuid() == 0:
+        logger.warning(
+            "The plugin %s requires root privileges. "
+            "Since you are not root (not using sudo) it may encounter errors.",
+            plugin.name)
 
 
 # TODO: Make run as close to auth + install as possible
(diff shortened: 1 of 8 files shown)


bmw:
Thanks for the PR @xgin. I know these are the changes proposed by @pde but I personally do not like this approach. The reason is that this changes the plugin API and will break plugins that don't have a `is_root_required` property (there is no requirement that they inherit from `letsencrypt.plugins.common.Plugin`).

I'm still in favor of the solution I discussed in IRC of simply removing the warning. To post my reasoning here, the "Root (sudo) is required to run most of letsencrypt functionality" message is only a warning. It does not stop execution so plugins still have to handle the case where they have insufficient permissions. Removing the warning stops suggesting people run the client with root privileges when it may not be necessary (such as when using `manual`, `webroot`, or `plesk` plugins). This approach also goes with the idea that it's easier to ask for forgiveness than permission.

If we did this, I think we should make sure sane error messages are thrown when we have insufficient permissions, but I suspect this will be a pretty straight forward fix.

If pde wants to use this approach, he's certainly able to merge this PR, but I personally think we can do better.

xgin:
@bmw Good point, I have not thought plugins could not be inherited from the `Plugin` class.
We may check the property exists  (`hasattr` or even `try-except`), otherwise treat it as either True or False.
That will not break backward compatibility.

I also agree that drop of the code is always a good solution :)
BTW, I have checked the error messages in plugins:
* standalone
```
Could not bind TCP port 443 because you don't have the appropriate permissions (for example, you aren't running this program as root).
```
* webroot
```
The webroot plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('/root/test does not exist or is not a directory',)
```
```
The webroot plugin is not working; there may be problems with your existing configuration.
The error was: PluginError("Couldn't create root for {0} http-01 challenge responses: {1}", 'ek-local.ru', OSError(13, 'Permission denied'))
```
* apache
```
The selected plugin encountered an error while parsing your server configuration
and cannot be used. The error was:

Error while running apache2ctl configtest.
Action 'configtest' failed.

apache2: Syntax error on line 216 of /etc/apache2/apache2.conf: Could not open
configuration file /etc/apache2/conf-enabled/zz010_psa_httpd.conf: Permission
denied
```

pde:
To be clear, I wasn't suggesting a breaking change to the plugin API for this -- either a warning printed by the plugins that need root themselves, or an API attribute if but only if we can make that property optional for the plugin to implement.

xgin:
@pde, Sure. The interface is not even touched.
Plugins inherited `letsencrypt.plugins.common.Plugin` respond to `is_root_required` with True by default.
I've added case with plugin having no attribute for complete backward compatibility.

letsencrypt/cli.py
```diff
@@ -542,6 +547,19 @@ def record_chosen_plugins(config, plugins, auth, inst):
     cn.installer = plugins.find_init(inst).name if inst else "none"
 
 
+def check_root_privileges(plugin):
+    """Warns user about insufficient privileges."""
+    if not hasattr(plugin, 'is_root_required'):
+        return
+    if not plugin.is_root_required:
+        return
+    if not os.geteuid() == 0:
```
kuba: `x != 0` instead of `not x == 0`

letsencrypt/cli.py
```diff
@@ -542,6 +547,19 @@ def record_chosen_plugins(config, plugins, auth, inst):
     cn.installer = plugins.find_init(inst).name if inst else "none"
 
 
+def check_root_privileges(plugin):
+    """Warns user about insufficient privileges."""
+    if not hasattr(plugin, 'is_root_required'):
+        return
+    if not plugin.is_root_required:
+        return
```
kuba: it' be a lot easier to just use `getattr(plugin, 'is_root_required', False)`

letsencrypt/plugins/common.py
```diff
@@ -90,6 +90,16 @@ def add_parser_arguments(cls, add):
 
         """
 
+    @property
```
kuba: Why make a property of such uncomplicated thing? Just say `is_root_required = True`

kuba:
That PR is very hackish :(

`interfaces.py` should be updated accordingly

kuba:
That being said, I would find it a lot better to just remove the warning as `bmw` says. Those hacks are not worth it.

pde:
OK, I'm going to try to take some time to figure out whether we want to clean up this PR, or take another simpler but (perhaps) less user-helpful route.

bmw:
Despite this change being made in conjunction with our largest 3rd party plugin, I think changing the plugin interface is a bad idea and unnecessary. With this in mind, I'm closing this PR since #1962 has been merged. If you'd like me to reconsider, please let me know!

PR #1943, opened 2015-12-18
Require PyOpenSSL >= 0.13

### LINKED ISSUE #1861: Versioned dependency on PyOpenSSL interferes with packaging

bmw:
Fixes #1861.

I had to remove use of:
* `OpenSSL.crypto.X509Extension._prefixes[OpenSSL.crypto._lib.GEN_DNS]` (it's a constant in PyOpenSSL anyway as of writing this PR)
* `OpenSSL.crypto.X509._subjectAltNameString()` and `OpenSSL.crypto.X509Req._subjectAltNameString()`
* Passing `acme.jose.ComparableX509` objects to `OpenSSL.crypto.load_certificate` and `OpenSSL.crypto.load_certificate_request`
* Keyword argument `bits` with `generate_key`

I tested this fairly thoroughly on a VPS, but this definitely needs more testing (including canary time).

@kuba, if you have time to take a look at this, I'd appreciate your feedback.


bmw:
Hit by letsencrypt/boulder#1237: [log](https://gist.github.com/c16ccf3d60fe24c8a61f)

acme/acme/challenges_test.py
```diff
@@ -264,7 +264,9 @@ def test_gen_verify_cert_gen_key(self):
 
     def test_verify_bad_cert(self):
         self.assertFalse(self.response.verify_cert(
-            test_util.load_cert('cert.pem')))
+            OpenSSL.crypto.load_certificate(
```
kuba: why don't you just use `test_util.load_cert()`?

bmw: Because `verify_cert` calls `crypto_util._pyopenssl_cert_or_req_san` which doesn't work with a `ComparableX509` object anymore.

I'm happy to change `_pyopenssl_cert_or_req_san` to have it work with both if you'd like. I'd just have to special case that type.

pde:
We didn't have a test farm script to run the unit tests, so I've [added one](https://github.com/pde/letstest/blob/upgrade-test/scripts/test_tox.sh).  That `test_tox.sh` test farm test currently succeeds on most but not all of our AMIs; this `pyopenssl` branch succeeds on the same subset of AMIs that `master` does (ie I haven't found a regression):

```
0 ubuntu15.10 pass
1 ubuntu15.04LTS pass
2 ubuntu14.04LTS pass
3 ubuntu14.04LTS_32bit pass
4 ubuntu12.04LTS fail
5 debian8.1 pass
6 debian7.8.aws.1 fail
7 debian7.8.aws.1_32bit fail
8 amazonlinux-2015.09.1 pass
9 amazonlinux-2015.03.1 pass
10 RHEL7 pass
11 fedora23 pass
12 centos7 pass
13 centos6 pass
```

acme/acme/jose/util.py
```diff
@@ -43,8 +43,17 @@ def __init__(self, wrapped):
     def __getattr__(self, name):
         return getattr(self._wrapped, name)
 
-    def _dump(self, filetype=OpenSSL.crypto.FILETYPE_ASN1):
-        # pylint: disable=missing-docstring,protected-access
+    def dump(self, filetype=OpenSSL.crypto.FILETYPE_ASN1):
```
kuba: I don't think this API should be exposed on `ComparableX509`. For instance it leads to abuses such as the one from this very PR:
```python
text = jose.ComparableX509(cert_or_req).dump(OpenSSL.crypto.FILETYPE_TEXT)
```

Moreover, if we want to really expose this API (not saying here if we should), then `filetype` argument shall not have default value as it leads to less readable code: `.dump()` - is that PEM, DER or maybe text representation?

acme/acme/challenges_test.py
kuba: You've uncovered an interesting bug over there (passing ComparableX509 to pyOpenSSL functions seems to work in later versions of the latter). We should audit the codebase and fix docstrings throughout. We probably also want `test_util.load_cert` and `test_util.load_comparable_cert`.

Also, how do you plan to test for regressions due to problems with `pyOpenSSL==0.13`. Current tests install latest version which would not catch all such errors. Maybe tests should be fixed to oldest version (of any package)?

acme/acme/crypto_util.py
```diff
@@ -161,31 +160,22 @@ def _pyopenssl_cert_or_req_san(cert_or_req):
...
-    # pylint: disable=protected-access,no-member
-    label = OpenSSL.crypto.X509Extension._prefixes[OpenSSL.crypto._lib.GEN_DNS]
-    assert parts_separator not in label
     prefix = label + part_separator
+    title = "X509v3 Subject Alternative Name:"
 
-    san_extensions = [
-        ext._subjectAltNameString().split(parts_separator)
-        for ext in extensions if ext.get_short_name() == extension_short_name]
+    text = jose.ComparableX509(cert_or_req).dump(OpenSSL.crypto.FILETYPE_TEXT)
```
kuba: Dumping to `FILETYPE_TEXT` might in the end be less hacky way of achieving the API of this function... However, I don't feel confident that this solution is good enough without testing it more carefully. I think we should add some test certs that would uncover problems, e.g.:
- very huge SAN list
- IDN

acme/acme/crypto_util.py
```diff
@@ -161,31 +160,22 @@ def _pyopenssl_cert_or_req_san(cert_or_req):
...
+    lines = iter(text.decode("utf-8").splitlines())
+    sans = [next(lines).split(parts_separator)
+            for line in lines if title in line]
```
kuba: Code above is overly complicated (`iter` & `next`). What's happening here?

acme/acme/crypto_util.py
```diff
@@ -161,31 +160,22 @@ def _pyopenssl_cert_or_req_san(cert_or_req):
...
     # WARNING: this function assumes that no SAN can include
     # parts_separator, hence the split!
 
-    return [part.split(part_separator)[1] for parts in san_extensions
-            for part in parts if part.startswith(prefix)]
+    return [part.split(part_separator)[1] for parts in sans
+            for part in parts if part.lstrip().startswith(prefix)]
```
kuba: it's not obvious to me why `lstrip` is necessary, which leads me to belief that there should be some inline comments with example data

acme/acme/challenges_test.py
pde: Pinning old versions of deps in our main tests is a bad idea because we'd miss regressions in the latest versions of our dependencies :)  We could instead write a [test farm test](https://github.com/letsencrypt/letsencrypt/tree/testfarm/tests/letstest/scripts) that pins the versions in old-but-popular OSes and runs the tests on those.

pde: Alternatively, we could just check in with Debian, Redhat and other folks making backport packages to ensure they run full tox tests on the things they build before uploading them to backports.  I believe that is already true for debian but @hlieberman should confirm.

bmw:
I think this is ready for another review. @kuba, I made all the changes you suggested:

* `dump` is a protected method of `ComparableX509`. You can access with wrapped object with `.wrapped`.
* Rewrote `_pyopenssl_cert_or_req_san`. I personally like the new version much better.
* Tests:
  * Unit tests for IDNs and large SANs lists.
  * For dependencies where we require a specific version (e.g. `PyOpenSSL>=0.13`), I added `tox` tests for Python 2.x testing the oldest allowed version. These tests now run in Travis along with `boulder` integration tests.

TODOs:
* I ran into some problems pinning the version of `pyparsing` specified by the nginx plugin. I stopped using the specified version in the tests and created #2106 to track the issue.
* Determine if we want to keep these new tests in Travis. Tests for both py26 and py27 with integration tests seems a bit excessive, but I'd like to keep them in place for now. If they begin causing problems we can reevaluate.
* We can definitely do more with testing older versions of our dependencies. The list packages/versions to use is hard coded and we have no such tests for Python dependencies where we don't specify a version. I believe solving both of these problems are outside of the scope of this PR though.

acme/acme/crypto_util_test.py
```diff
@@ -69,6 +71,14 @@ def _call(cls, loader, name):
         from acme.crypto_util import _pyopenssl_cert_or_req_san
         return _pyopenssl_cert_or_req_san(loader(name))
 
+    @classmethod
+    def _get_idn_names(cls):
```
kuba: How was this function determined? There needs to be some kind of inline comment explanation or if it was used to generate the certs, maybe it's better to regenerate those certs during tests rather than dropping static files?

acme/acme/crypto_util_test.py
```diff
@@ -69,6 +71,14 @@ def _call(cls, loader, name):
...
+        chars = [six.unichr(i) for i in itertools.chain(range(0x3c3, 0x400),
```
kuba: you should use `from six.moves import range` for those range functions

kuba:
`py26-oldest`, `py27-oldest` - that's a very interesting idea, nice! Can you please add a comment to `setup.py` to remind everyone to keep the oldest dep list in sync?

I personally find any kind of new tests relatively cheap (additional O(minutes) for testing < O(dozens of hours) of debugging and cumulative time of bugged end-users). Agreed that any sync-like solutions are out of scope for the PR.

letsencrypt/plugins/standalone.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, *args, **kwargs):
 
         # one self-signed key for all tls-sni-01 certificates
         self.key = OpenSSL.crypto.PKey()
-        self.key.generate_key(OpenSSL.crypto.TYPE_RSA, bits=2048)
+        self.key.generate_key(OpenSSL.crypto.TYPE_RSA, 2048)
```
kuba: I find `bits=2048` more readable

acme/acme/crypto_util.py
```diff
@@ -161,31 +160,23 @@ def _pyopenssl_cert_or_req_san(cert_or_req):
     :rtype: `list` of `unicode`
 
     """
-    # constants based on implementation of
-    # OpenSSL.crypto.X509Error._subjectAltNameString
-    parts_separator = ", "
+    # constants based on PyOpenSSL certificate/CSR text dump
```
kuba: let's explain here why we use text dump - reference relevant issues, mention dep version requirements, maybe this PR - so that in future someone doesn't try to waste time to rewrite it

kuba:
Nice job, this looks very good, @bmw! Only a couple of small nits.

letsencrypt/plugins/standalone.py
bmw: I agree but it doesn't work with `PyOpenSSL` version 0.13:
```
TypeError: generate_key() takes no keyword arguments
```

kuba: right :(

bmw:
Ready for another review!

acme/acme/crypto_util.py
kuba: reminder: this wasn't addressed in the latest commits push

bmw: There's [this](https://github.com/letsencrypt/letsencrypt/pull/1943/files#diff-f437033746ea336ab13f6ba632e01807R172). Inadequate?

bmw: On second thought, that comment is inadequate. Just added a more verbose version to the top of the function.

bmw:
Failed Travis log linked in #2118.

kuba:
Was this tested under Python 3? I don't see it being part of the build matrix...

pde:
There was a `.travis.yml` merge conflict preventing py3 testing.

kuba:
LGTM, merging

Issue #1934, opened 2015-12-17
Error in parsing apache config file /etc/apache2/sites-enabled/graphite-web.conf


ser:
```
# ./letsencrypt-auto --apache -d domain.com --email email@domain.com --agree-tos
Updating letsencrypt and virtual environment dependencies.......
Running with virtualenv: /root/.local/share/letsencrypt/bin/letsencrypt --apache -d domain.com --email email@domain.com --agree-tos
The apache plugin is not working; there may be problems with your existing configuration.
The error was: PluginError(('There has been an error in parsing the file (%s): %s', u'/etc/apache2/sites-enabled/graphite-web.conf', u'Syntax error'),)
```
```
# cat /etc/apache2/sites-enabled/graphite-web.conf
<VirtualHost *:80>

        WSGIDaemonProcess _graphite processes=5 threads=5 display-name='%{GROUP}' inactivity-timeout=120 user=_graphite group=_graphite
        WSGIProcessGroup _graphite
        WSGIImportScript /usr/share/graphite-web/graphite.wsgi process-group=_graphite application-group=%{GLOBAL}
        WSGIScriptAlias / /usr/share/graphite-web/graphite.wsgi

        Alias /content/ /usr/share/graphite-web/static/
        <Location "/content/">
                SetHandler None
        </Location>

        ErrorLog ${APACHE_LOG_DIR}/graphite-web_error.log

        # Possible values include: debug, info, notice, warn, error, crit,
        # alert, emerg.
        LogLevel warn

        CustomLog ${APACHE_LOG_DIR}/graphite-web_access.log combined

</VirtualHost>
```

pde:
@domcleal this looks like an Augeas issue

domcleal:
Thanks for the heads up @pde, PR's up at #1945.

PR #1945, opened 2015-12-18
Merge Augeas lens fix for quotes in directive arguments

domcleal:
Augeas fails to parse a directive argument with a quote inside (expecting
either fully quoted or unquoted values).

From hercules-team/augeas@d4d7ea9

Closes: #1934


pde:
Passes hackish Apache tests, merging for canary testing.

PR #1953, opened 2015-12-19
Suppress spurious output

cure:
Suppress spurious output while testing for the presence of the virtualenv or python-virtualenv package.


SwartzCr:
LGTM - unless we need those to appear in the logs

pde:
We shouldn't need them, no. In cases where we need to infer which virtualenv packages were installed, that should be visible from the apt-get command's output below.

pde:
Also LGTM.  Two +1s for a letsencrypt-auto PR, so merging.

PR #1959, opened 2015-12-19
Keep storage.names() from passing None to open()

convissor:
Fixes exiting abnormally with:
    TypeError: coercing to Unicode: need string or buffer, NoneType found

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -450,12 +450,15 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
         :param int version: the desired version number
         :returns: the subject names
         :rtype: `list` of `str`
+        :raises .CertStorageError: if could not find cert file.
 
         """
         if version is None:
             target = self.current_target("cert")
         else:
             target = self.version("cert", version)
+        if target is None:
+            raise errors.CertStorageError("could not find cert file")
         with open(target) as f:
             return crypto_util.get_sans_from_cert(f.read())
 


convissor:
The Travis CI failure was due to one of the four test jobs being unable to reach the test server.

pde:
(Travis issue fixed)

Hi @convissor were you seeing this error occur in practice?  If so, under what circumstances?

convissor:
Funny you ask.  Last night while falling asleep I thought, "hey, I should explain how to reproduce that." :)

Pick an existing certificate to renew.  Before renewing, rename or delete the `/etc/letsencrypt/live/<domain>` directory.  Then call letsencrypt.  I did it via `letsencrypt-auto certonly --webroot ...`, though I suspect it will happen under other plugins as well.

bmw:
Just a thought, perhaps `RenewableCert` objects should have checks added to `__init__` to verify that the symlinks and their targets exist to prevent this and similar problems occurring in other functions as well.

convissor:
That can be an additional check.  The check I added should stay because `current_target()` can return `None`.

bmw:
I believe if we add the checks I suggested, `current_target` cannot return `None`.

bmw:
This check doesn't hurt though. With that in mind, I'm merging this PR and opening a new issue to fix the problem in `__init__`.

Thanks for writing this up @convissor!

pde:
Ah, that makes sense.  Probably in cases where there is a renewal configuration file that points to nonexistent things in `/etc/letsencrypt/live`, we should print a warning; if the user really, really wants to move stuff out of /etc/letsencrypt/live, maybe they want to either update or delete the renewal configuration file.

PR #1962, opened 2015-12-20
Remove warning "Root (sudo) is required"

xgin:
Pending the PR #1938 
Remove misleading warning because some plugins do not require root privileges (e.g. https://github.com/plesk/letsencrypt-plesk/issues/35)

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1384,17 +1384,6 @@ def main(cli_args=sys.argv[1:]):
     zope.component.provideUtility(report)
     atexit.register(report.atexit_print_messages)
 
-    if not os.geteuid() == 0:
-        logger.warning(
-            "Root (sudo) is required to run most of letsencrypt functionality.")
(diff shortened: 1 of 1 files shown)


Natim:
I agree that root shouldn't be required to run letsencrypt.
However it doesn't seems possible to configure letsencrypt webroot to store SSL cerficates at another place than /etc/letsencrypt.

Doesn't your plugin store them there?

xgin:
@Natim there is no such problem:
```
    --config-dir "${LE_HOME}/etc" \
    --work-dir "${LE_HOME}/lib" \
    --logs-dir "${LE_HOME}/logs" \
```

bmw:
@pde, you said you were going to think about this, but I'd like to take this PR and clean up the relevant error messages.

bmw:
ping @pde. I'd like to merge this and close #1938. Thoughts?

bmw:
Sorry we sat on this so long. lgtm and thanks for the PR @xgin. Merging

PR #1971, opened 2015-12-21
Import letstest test farm (with less git magic)

pde:
Using a `git mv` inside letstest this time.

### DIFF
--- a/tests/letstest/README.md
+++ b/tests/letstest/README.md
@@ -1,0 +1,44 @@
+# letstest
+simple aws testfarm scripts for letsencrypt client testing
+
+- Configures (canned) boulder server
+- Launches EC2 instances with a given list of AMIs for different distros
+- Copies letsencrypt repo and puts it on the instances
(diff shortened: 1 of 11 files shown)


kuba:
1. Urm, how did you use `git mv` between two disjoint repos? How did the commits show up here? Something fishy here.
2. I'll repeat once again: I don't think it should be merged into this (already giant) repo. This is something that can leave completely on its own...

pde:
1. The answer is that you add one repo as a remote of the other, and merge a branch from it.  Afterwards, we see some commits that add the tests to an empty tree, then a commit to move them to the correct location, then a merge that incorporates that strand of history with the main repo's strands of history.
2. Discussion in the IRC channel has a several active developers agreeing that we'd like these tests in the tree, because there are interactions between the scripts they run and other parts of the codebase, especially the other tests.

pde:
The `git mv` in particular occurred in a letstest branch; it's this commit: 87dfe8c

bmw:
LGTM

Issue #1966, opened 2015-12-20
letsencrypt-0.1.1 but it requires acme=0.2.0.dev0


willysr:
Hi

It seems in letsencrypt-0.1.1, you missed this file
"letsencrypt/__ init __ .py" (i added space otherwise it will be edited by the editor)

it still point to 0.2.0.dev0
i suggest to make a quick release of 0.1.2 to fix this problem as the problem will not be spotted until the client runs letsencrypt command

kuba:
Tarball from PyPI seems to have correct version string (https://pypi.python.org/packages/source/l/letsencrypt/letsencrypt-0.1.1.tar.gz#md5=ddd393563f3d5835d678b7478148e3d9). However, https://github.com/letsencrypt/letsencrypt/blob/v0.1.1/letsencrypt/__init__.py is wrong, which suggests that wrong commit has been tagged.

willysr:
Yes, i used the one from github since it contains all the code, not just the acme protocol

pde:
Looks like the bug is that the release scripts [set the version correctly](https://github.com/letsencrypt/letsencrypt/blob/master/tools/release.sh#L87) but [fail to stage that file](https://github.com/letsencrypt/letsencrypt/blob/master/tools/release.sh#L89) for committing and tagging.

willysr:
good thing it can be found and hopefully gets fixed for the next round of release

PR #1978, opened 2015-12-22
release.sh: stage version changes to letsencrypt/ !

pde:
Fixes: #1966

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -86,7 +86,7 @@ SetVersion() {
     done
     sed -i "s/^__version.*/__version__ = '$ver'/" letsencrypt/__init__.py
 
-    git add -p $SUBPKGS # interactive user input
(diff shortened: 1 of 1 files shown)


tools/release.sh
```diff
@@ -86,7 +86,7 @@ SetVersion() {
     done
     sed -i "s/^__version.*/__version__ = '$ver'/" letsencrypt/__init__.py
 
-    git add -p $SUBPKGS # interactive user input
+    git add -p letsencrypt $SUBPKGS # interactive user input
```
Natim: Actually letsencrypt is already present here: https://github.com/letsencrypt/letsencrypt/pull/1978/files#diff-b91eeb35e4e3920b1a96d2d03212320aR45

Natim: Oh sorry it isn't...

bmw:
LGTM

Issue #1979, opened 2015-12-22
webroot unit tests fail when run as root


pde:
Because root can read files even when they are chmod'd to be unreadable, this test is erroring out as root:

https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/plugins/webroot_test.py#L67

pde:
(This has caused utter havoc for debian packaging)

PR #1980, opened 2015-12-22
When permissions don't work (maybe you're root), disable a unit test that assumes permissions work

pde:
Fixes: #1979

### DIFF
--- a/letsencrypt/plugins/webroot_test.py
+++ b/letsencrypt/plugins/webroot_test.py
@@ -73,7 +73,7 @@ class AuthenticatorTest(unittest.TestCase):
         os.chmod(self.path, 0o000)
         try:
             open(permission_canary, "r")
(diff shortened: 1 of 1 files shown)


joohoi:
Are we actually checking if we're root (title of PR), or if we have too broad privileges to begin with?

If we're supposed to check only if we're root, shouldn't os.geteuid() do the trick?

pde:
I guess I didn't want to test for the hypothesised cause of the problem, but for evidence that it was always occurring.  Will re-title the PR :)

pde:
For instance, I don't fully understand how debian's `fakeroot` tool works and how it would interact with `os.geteuid()`

joohoi:
I personally like to work with files using 'with open as', but that's only preference. LGTM

pde:
Taking @joohoi's suggestion, and merging.

PR #1998, opened 2015-12-23
Updated nginx information for cert.pem.

Dygear:
Including a link to the nginx documentation regarding the ssl_trusted_certificate directive.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -230,7 +230,9 @@ The following files are available:
   Server certificate only.
 
   This is what Apache < 2.4.8 needs for `SSLCertificateFile
-  <https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile>`_.
+  <https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile>`_,
+  and nginx for `ssl_trusted_certificate
+  <http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate>`_.
 
 ``chain.pem``
   All certificates that need to be served by the browser **excluding**


kuba:
This seems to be in conflict with what's said for `fullchain.pem` (see comment about nginx).

Dygear:
So the information is omitted because we don't want people using both directives?

kuba:
By providing 2 alternative ways of achieving the same thing we would confuse the user. Unless, that directive does something different?

Dygear:
I'm using nginx's [`ssl_certificate`][1] directive for my `fullchain.pem` file.

> Syntax:	`ssl_certificate file;`
> Default:	`-`
> Context: `http, server`

> Specifies a file with the certificate in the PEM format for the given virtual server. If intermediate certificates should be specified in addition to a primary certificate, they should be specified in the same file in the following order: the primary certificate comes first, then the intermediate certificates. A secret key in the PEM format may be placed in the same file.

> It should be kept in mind that due to the HTTPS protocol limitations virtual servers should listen on different IP addresses:

> `server {
> `    listen          192.168.1.1:443;`
> `    server_name     one.example.com;`
> `    ssl_certificate /usr/local/nginx/conf/one.example.com.cert;`
> `    ...`
> `}`

> `server {`
> `    listen          192.168.1.2:443;`
> `    server_name     two.example.com;`
> `    ssl_certificate /usr/local/nginx/conf/two.example.com.cert;`
> `    ...`
> `}`

> otherwise the [first server’s certificate](http://nginx.org/en/docs/http/configuring_https_servers.html#name_based_https_servers) will be issued for the second site.

I'm using nginx's [`ssl_certificate_key`][2] directive for my `privkey.pem` file.

> Syntax: `ssl_certificate_key file;`
> Default: `-`
> Context: `http, server`

> Specifies a `file` with the secret key in the PEM format for the given virtual server.

> The value `engine:name:id` can be specified instead of the `file` (1.7.9), which loads a secret key with a specified `id` from the OpenSSL engine `name`.

I'm using nginx's [`ssl_trusted_certificate`][3] directive for my `chain.pem` file, **because I am using [`ssl_stapling`][4].**

> Syntax: `ssl_trusted_certificate file;`
> Default:	`-`
> Context: `http, server`
> This directive appeared in version 1.3.7.

> Specifies a `file` with trusted CA certificates in the PEM format used to [verify](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_verify_client) client certificates and OCSP responses if [ssl_stapling][4] is enabled.

> In contrast to the certificate set by [ssl_client_certificate](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_client_certificate), the list of these certificates will not be sent to clients.

However I do not have a currently working client. As I'm reading through the documentation I'm trying to update it with information that I find to be missing.

[1]: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_certificate
[2]: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_certificate_key
[3]: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate
[4]: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling

bmw:
An equivalent PR landed in #2094 so I'm closing this one. Regardless, thanks taking the time to write this up for us.

PR #2004, opened 2015-12-23
--hsts should not use includeSubDomains

### LINKED ISSUE #1728: don't add includeSubDomains for Strict-Transport-Security header
Hi,

I think the letsencrypt client shouldn’t add `includeSubDomains` to the Strict-Transport-Security header by default.

If you let the letsencrypt client modify the configuration for a webserver which hosts example.com, this would make it only possible to reach subdomain hosts (foo.example.com) with https. If you forgot only one webserver which serves a subdomain by http only, you can't reach this one any more.
And as browsers actually don't have an option to clear the HSTS cache, It won't help you to remove the includeSubDomains later from the webserver config.

includeSubDomains is something, that an admin should only add if he truly knows, that all subdomains are reachable by https. So I think this should never be added by an automatic tool, like the letsencrypt client.

RFC 6797 section 11.1 [1] also describes some implications of includeSubDomains

Chris

[1] https://tools.ietf.org/html/rfc6797#section-11.4

pde:
Fixes #1728

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/constants.py
+++ b/letsencrypt-apache/letsencrypt_apache/constants.py
@@ -33,7 +33,7 @@ REWRITE_HTTPS_ARGS_WITH_END = [
     https vhost"""
 
 HSTS_ARGS = ["always", "set", "Strict-Transport-Security",
-    "\"max-age=31536000; includeSubDomains\""]
+    "\"max-age=31536000\""]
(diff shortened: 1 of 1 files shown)


pde:
Tested on https://is.isnot.org

bmw:
LGTM. Tested on a VPS.

sagi:
LGTM.

Issue #1997, opened 2015-12-23
letsencrypt.validator.Validator is dead code. Remove it as well as its dependencies.


Natim:
As per https://github.com/letsencrypt/letsencrypt/pull/1914#discussion_r48341824

kuba:
```
jakub@iks:~/dev/letsencrypt/letsencrypt$ git grep 'from letsencrypt import validator'
letsencrypt-compatibility-test/letsencrypt_compatibility_test/test_driver.py:from letsencrypt import validator
letsencrypt/tests/validator_test.py:from letsencrypt import validator
```

So it seems like `letsencrypt-compatibility-test` is still using that. In turn, I'm not sure if that package is still in use.

Natim:
Yes it is what I was going to ask :)

Natim:
Apparently it is used to test plugins:

> Tests Let's Encrypt plugins against different server configuratons. It is
> assumed that Docker is already installed. If no test types is specified, all
> tests that the plugin supports are performed.

Natim:
Should we move the validator code in here?

bmw:
I think moving the code into `letsencrypt-compatibility-test`, at least for now, makes sense.

pde:
Yes @Natim, we'd love a PR for that.  What dependencies would it free us from?

Natim:
Apparently requests as well as l the one for secure SSL connection with
python < 2.7.9

Natim:
Tentative of fixing this here: https://github.com/letsencrypt/letsencrypt/pull/2007

pde:
Thanks @Natim !

PR #2007, opened 2015-12-23
Move validator to compatibility-test — Refs #1997

Natim:


Natim:
I don't know if tests in letsencrypt-compatibility-tests actually runs with travis but at least the linter is happy.

pde:
They don't.  It's something we should fix...

PR #2008, opened 2015-12-23
Include the python version being used in the user agent for ACME requ…

alex:
…ests.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -2,6 +2,7 @@
 import datetime
 import heapq
 import logging
+import platform
 import time
 
 import six
@@ -483,6 +484,8 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
                 'Successful revocation must return HTTP OK status')
 
 
+DEFAULT_USER_AGENT = 'acme-python Python/%s' % platform.python_version()
+
 class ClientNetwork(object):
     """Client network."""
     JSON_CONTENT_TYPE = 'application/json'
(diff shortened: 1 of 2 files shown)


alex:
cc: @reaperhulk

pde:
Hi @alex; is there a specific reason why we need to collect this?

alex:
@pde the local impetus is that pyca/cryptography is interested in deprecating Python 2.6 support, letsencrypt is the downstream most likely to be affected, and we'd like to collect data to see how many folks would be affected.

The more general reason is that it's a useful thing to have data on :-)

acme/acme/client.py
```diff
@@ -483,6 +484,8 @@ def revoke(self, cert):
                 'Successful revocation must return HTTP OK status')
 
 
+DEFAULT_USER_AGENT = 'acme-python Python/%s' % platform.python_version()
+
```
kuba: there should be 2 separating newlines between constant and class here

kuba:
I'm not sure if we should disclose user's Python version, at least in the reusable `acme` library by default. That gives an attacker more info about the target operating system. Core client already discloses a ton of data, so maybe let's keep the change scope to `letsencrypt/` dir and revert `acme`?

alex:
@kuba done.

kuba:
Thanks, @alex :)

pde:
@alex [these](https://docs.google.com/spreadsheets/d/1i0O9EO61NL-YyAHp0OlOXk-5GliMwfO1qrEr5Gx01ek/edit#gid=1767154651) are the top OSes the client is running on at present.

I think this implies that a bit over 10% of clients are running on platforms where Python 2.6 is the default.  Does that sufficiently answer your question?

pde:
Actually maybe more like 13-15%, a large fraction of "Other" is older OSes including Centos 6 variants and  Debian 6.

alex:
@pde thanks!

pde:
Closing this for now then.

PR #2010, opened 2015-12-24
Show error detail to the user.

jsha:
Previously this code would show the user a hardcoded message based on the error
type, but it's much more useful to show the error detail field, which often
helps debug the problem.

### DIFF
--- a/letsencrypt/auth_handler.py
+++ b/letsencrypt/auth_handler.py
@@ -540,11 +540,11 @@ def _generate_failed_chall_msg(failed_achalls):
 
     """
     typ = failed_achalls[0].error.typ
-    msg = [
-        "The following errors were reported by the server:".format(typ)]
+    msg = ["The following errors were reported by the server:"]
 
     for achall in failed_achalls:
-        msg.append("\n\nDomain: %s\nType:   %s\nDetail: %s" % (achall.domain, achall.error.typ, achall.error.detail))
+        msg.append("\n\nDomain: %s\nType:   %s\nDetail: %s" % (
+            achall.domain, achall.error.typ, achall.error.detail))
 
     if typ in _ERROR_HELP:
         msg.append("\n\n")
--- a/letsencrypt/auth_handler.py
+++ b/letsencrypt/auth_handler.py
@@ -543,13 +543,8 @@ def _generate_failed_chall_msg(failed_achalls):
     msg = [
         "The following '{0}' errors were reported by the server:".format(typ)]
 
-    problems = dict()
     for achall in failed_achalls:
-        problems.setdefault(achall.error.description, set()).add(achall.domain)
-    for problem in problems:
-        msg.append("\n\nDomains: ")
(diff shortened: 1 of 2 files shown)


kuba:
I would personally display both, i.e. `achall.error.detail` and `achall.error.description` if available.

jsha:
I disagree. `achall.error.description`, as I understand it, is copied from the `semantic` column in the Errors section of the spec (https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#errors), but those values aren't intended to be user-facing error messages, and in fact are quite bad for that purpose. For instance, the error type that almost always comes up for challenges is `connection`, which gets printed as `The server could not connect to the client for DV`. That's meaningless at best-- most people don't know what DV means-- and incorrect at worst, since it makes DNS errors sound like HTTP errors.

letsencrypt/auth_handler.py
```diff
@@ -543,13 +543,8 @@ def _generate_failed_chall_msg(failed_achalls):
...
         "The following '{0}' errors were reported by the server:".format(typ)]
 
-    problems = dict()
     for achall in failed_achalls:
-        problems.setdefault(achall.error.description, set()).add(achall.domain)
-    for problem in problems:
-        msg.append("\n\nDomains: ")
-        msg.append(", ".join(sorted(problems[problem])))
-        msg.append("\nError: {0}".format(problem))
+        msg.append("\n\nDomain: %s\nDetail: %s\n" % (achall.domain, achall.error.detail))
```
pde: This creates three newlines between errors; two would be more appropriate.

pde:
Here's a typical output from the code in this branch as is:
```
 - The following 'urn:acme:error:connection' errors were reported by
   the server:

   Domain: tweedle.isnot.org
   Detail: Failed to connect to host for DVSNI challenge


   Domain: dee.isnot.org
   Detail: Failed to connect to host for DVSNI challenge
```

Since we're landing this patch, consider fixing up the very strange `'urn:acme:error:connection' errors` language.  How about stripping so that it reads:

`The following connection errors were reported by the server`.

Or even:

`The following ACME connection errors were reported by the server`?

pde:
Otherwise LGTM

jsha:
Good points. I changed it to put the Type field in the error body rather than at the top, and fixed the newline issue. Please re-review and merge.

pde:
LGTM.  Fix the lint error and merge!

letsencrypt/auth_handler.py
```diff
@@ -541,15 +541,10 @@ def _generate_failed_chall_msg(failed_achalls):
     """
     typ = failed_achalls[0].error.typ
     msg = [
-        "The following '{0}' errors were reported by the server:".format(typ)]
+        "The following errors were reported by the server:".format(typ)]
```
sjerdo: Why is the `.format(typ)` still there?

jsha:
Fixed lint error and removed left-over `.format(typ)`, thanks! Will merge after tests finish.

PR #2013, opened 2015-12-24
Fixes to test farm scripts

pde:


pde:
Self-merging, since this is not user-facing code.

PR #2014, opened 2015-12-24
Allow issuing certificates without authz-handshake

valhallasw:
I'm interested in a situation where there are two front-end webservers without shared filesystem, that each should have ssl certificates. Only one is serving at one time, the other is a hot spare. If I read the spec correctly, this should be possible by 
 - having the agent private key on both servers (one-time manual setup),
 - letting the active webserver run authentication for a given domain (new-authz)
 - have both webservers call new-cert

This patch adds a parameter --skip-authz which skips the authz
handshake. The intended use is a situation where the client
authorization and certificate generation happen on different machines.

To allow for this, some changes in acme were also necessary:
 - Client.request_issuance now directly takes new_cert_uri instead of
   authzrs
 - messages.CertificateResource no longer contains authzrs, as these
   do not necessarily exist without authz-handshake. As far as I could
   see, this parameter was unused.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -286,19 +286,18 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         # TODO: check and raise UnexpectedUpdate
         return updated_authzr, response
 
-    def request_issuance(self, csr, authzrs):
+    def request_issuance(self, csr, new_cert_uri):
         """Request issuance.
 
         :param csr: CSR
         :type csr: `OpenSSL.crypto.X509Req` wrapped in `.ComparableX509`
 
-        :param authzrs: `list` of `.AuthorizationResource`
+        :param new_cert_uri: new-cert URI
 
         :returns: Issued certificate
         :rtype: `.messages.CertificateResource`
 
         """
-        assert authzrs, "Authorizations list is empty"
         logger.debug("Requesting issuance...")
 
         # TODO: assert len(authzrs) == number of SANs
@@ -306,7 +305,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
         content_type = self.DER_CONTENT_TYPE  # TODO: add 'cert_type 'argument
         response = self.net.post(
-            authzrs[0].new_cert_uri,  # TODO: acme-spec #90
+            new_cert_uri,
             req,
             content_type=content_type,
             headers={'Accept': content_type})
@@ -319,7 +318,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             raise errors.ClientError('"Location" Header missing')
 
         return messages.CertificateResource(
-            uri=uri, authzrs=authzrs, cert_chain_uri=cert_chain_uri,
+            uri=uri, cert_chain_uri=cert_chain_uri,
             body=jose.ComparableX509(OpenSSL.crypto.load_certificate(
                 OpenSSL.crypto.FILETYPE_ASN1, response.content)))
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -998,6 +998,10 @@ def prepare_and_parse_args(plugins, args):
         "automation", "--duplicate", dest="duplicate", action="store_true",
         help="Allow making a certificate lineage that duplicates an existing one "
              "(both can be renewed in parallel)")
+    helpful.add(
+        "automation", "--skip-authz", dest="skip_authz", action="store_true",
+        help="Skip domain authentication. This requires the domains to have been "
+             "authenticated in a previous run.")
 
     helpful.add_group(
         "testing", description="The following flags are meant for "
--- a/letsencrypt/client.py
+++ b/letsencrypt/client.py
@@ -221,11 +221,16 @@ class Client(object):
 
         logger.debug("CSR: %s, domains: %s", csr, domains)
 
-        authzr = self.auth_handler.get_authorizations(domains)
+        if not self.config.skip_authz:
+            authzrs = self.auth_handler.get_authorizations(domains)
+            certr_url = authzrs[0].new_cert_uri
+        else:
+            certr_url = self.acme.directory['new-cert']
+
         certr = self.acme.request_issuance(
             jose.ComparableX509(OpenSSL.crypto.load_certificate_request(
                 OpenSSL.crypto.FILETYPE_ASN1, csr.data)),
-            authzr)
+            certr_url)
         return certr, self.acme.fetch_chain(certr)
 
     def obtain_certificate_from_csr(self, csr):
--- a/letsencrypt/tests/client_test.py
+++ b/letsencrypt/tests/client_test.py
@@ -79,7 +79,8 @@ class ClientTest(unittest.TestCase):
 
     def setUp(self):
         self.config = mock.MagicMock(
-            no_verify_ssl=False, config_dir="/etc/letsencrypt")
+            no_verify_ssl=False, config_dir="/etc/letsencrypt",
+            skip_authz=False)
         # pylint: disable=star-args
         self.account = mock.MagicMock(**{"key.pem": KEY})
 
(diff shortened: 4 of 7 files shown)


kuba:
FTR, Travis build fails with `FailedChallenges: Failed authorization procedure. nginx.wtf (tls-sni-01): urn:acme:error:connection :: The server could not connect to the client to verify the domain :: DNS query timed out` (https://travis-ci.org/letsencrypt/letsencrypt/jobs/98766322#L4433), which seems like a transient error unrelated to the changes in this PR.

acme/acme/client.py
```diff
@@ -286,19 +286,18 @@ def poll(self, authzr):
         # TODO: check and raise UnexpectedUpdate
         return updated_authzr, response
 
-    def request_issuance(self, csr, authzrs):
```
kuba: There are a couple of other places where this function is used but wasn't updated (`git grep request_issuance\(`):
- [ ] `poll_and_request_issuance`
- [ ] `example_client.py`

acme/acme/client.py
```diff
@@ -286,19 +286,18 @@ def poll(self, authzr):
...
+    def request_issuance(self, csr, new_cert_uri):
         """Request issuance.
 
         :param csr: CSR
         :type csr: `OpenSSL.crypto.X509Req` wrapped in `.ComparableX509`
 
-        :param authzrs: `list` of `.AuthorizationResource`
+        :param new_cert_uri: new-cert URI
```
kuba: Please leave a hint on how to get this URI: reference `AuhtorizationResource.new_cert_uri`

letsencrypt/cli.py
```diff
@@ -998,6 +998,10 @@ def prepare_and_parse_args(plugins, args):
         "automation", "--duplicate", dest="duplicate", action="store_true",
         help="Allow making a certificate lineage that duplicates an existing one "
              "(both can be renewed in parallel)")
+    helpful.add(
+        "automation", "--skip-authz", dest="skip_authz", action="store_true",
+        help="Skip domain authentication. This requires the domains to have been "
+             "authenticated in a previous run.")
```
kuba: I'm pretty sure indentation here is wrong, it should be aligned to letter "h"

letsencrypt/client.py
```diff
@@ -221,11 +221,16 @@ def _obtain_certificate(self, domains, csr):
 
         logger.debug("CSR: %s, domains: %s", csr, domains)
 
-        authzr = self.auth_handler.get_authorizations(domains)
+        if not self.config.skip_authz:
+            authzrs = self.auth_handler.get_authorizations(domains)
+            certr_url = authzrs[0].new_cert_uri
```
kuba: The `# TODO: acme-spec #90` comment was removed. While the referenced issue was closed by https://github.com/letsencrypt/acme-spec/commit/7aac185b6c05b3665d682f9efbbd8bac70effc96, I still think we should leave a comment here on why it's enough to look at `[0]`. In fact I would actually check that all authorizations have the same URI and `logger.warn` or even raise an error otherwise.

Please use `new_cert_uri` name.

You can inline `authzrs`: `self.auth_handler.get_authorizations(domains)[0].new_cert_uri`

letsencrypt/tests/client_test.py
```diff
@@ -100,22 +101,35 @@ def _mock_obtain_certificate(self):
         self.acme.request_issuance.return_value = mock.sentinel.certr
         self.acme.fetch_chain.return_value = mock.sentinel.chain
 
-    def _check_obtain_certificate(self):
-        self.client.auth_handler.get_authorizations.assert_called_once_with(
-            ["example.com", "www.example.com"])
+    def _check_obtain_certificate(self, skip_authz=False):
+        if skip_authz:
+            self.client.auth_handler.get_authorizations.assert_not_called()
+            uri = self.acme.directory['cert-uri']
```
kuba: `cert-uri` should be changed to `new-cert`, `uri` should be change to `new_cert_uri`

kuba: That being said, I'm a bit worried that tests didn't catch this error at all

letsencrypt/tests/client_test.py
```diff
@@ -100,22 +101,35 @@ def _mock_obtain_certificate(self):
...
+        else:
+            self.client.auth_handler.get_authorizations.assert_called_once_with(
+                ["example.com", "www.example.com"])
+            uri = self.client.auth_handler.get_authorizations()[0].new_cert_uri
+
         self.acme.request_issuance.assert_called_once_with(
             jose.ComparableX509(OpenSSL.crypto.load_certificate_request(
                 OpenSSL.crypto.FILETYPE_ASN1, CSR_SAN)),
-            self.client.auth_handler.get_authorizations())
+            uri)
```
kuba: this would fit on the previous line :)

letsencrypt/tests/client_test.py
```diff
@@ -100,22 +101,35 @@ def _mock_obtain_certificate(self):
...
         self._mock_obtain_certificate()
         self.assertEqual(
             (mock.sentinel.certr, mock.sentinel.chain),
             self.client.obtain_certificate_from_csr(le_util.CSR(
                 form="der", file=None, data=CSR_SAN)))
-        self._check_obtain_certificate()
+        self._check_obtain_certificate(skip_authz)
+
+    def test_obtain_certificate_from_csr_with_authz(self):
+        self._test_obtain_certificate_from_csr(False)
```
kuba: please use named kwargs: `skip_authz=False`

letsencrypt/tests/client_test.py
```diff
@@ -100,22 +101,35 @@ def _mock_obtain_certificate(self):
...
+
+    def test_obtain_certificate_from_csr_without_authz(self):
+        self._test_obtain_certificate_from_csr(True)
```
kuba: same as above

letsencrypt/tests/client_test.py
```diff
@@ -100,22 +101,35 @@ def _mock_obtain_certificate(self):
         self.acme.request_issuance.return_value = mock.sentinel.certr
         self.acme.fetch_chain.return_value = mock.sentinel.chain
 
-    def _check_obtain_certificate(self):
-        self.client.auth_handler.get_authorizations.assert_called_once_with(
-            ["example.com", "www.example.com"])
+    def _check_obtain_certificate(self, skip_authz=False):
+        if skip_authz:
+            self.client.auth_handler.get_authorizations.assert_not_called()
```
kuba: `assert_not_called` requires `mock>=1.3.0`, so it's a NOP e.g. in Python 2.6. Please fix

kuba: Apologies, meant `1.1.0`: https://github.com/testing-cabal/mock/commit/c7d6c7dcafd4482626194528b24f174dd15f5f5e

kuba:
To be honest, I would be very hesitant to include this change. Your "2 spares" workflow would only differ in the certificate retrieval step: instead of copying from active spare you get the cert from the CA! Why not just copy the cert between machines yourself? NB, you already have to set up those 2 machines differently, as one will have to be run with `--skip-authz` so this means you need to wrap all of this workflow in some automation scripts. It seems pretty ridiculous for such generic client to support a feature that can easily be implemented outside the scope of this software. More features leads to more bugs. At this stage `--skip-authz` is a no-go for me.

On the other hand, the code for `acme` library is not that bad and I would consider merging it even if changes to the core client are not submitted. However, it does break compatibility with existing clients such as https://github.com/kuba/simp_le.

No matter what happens, code needs to be improved as outlined in the comments above.

valhallasw:
Hi Jakub,

Thank you for your extensive comments. My motivation for this change comes from -- in retrospect -- an incorrect reading of the spec. I imagined the following workflow:

1) (manually) copy the user certificate to all servers,
2) (manually) authorize a given domain for that user on the currently serving server,
3) (automatically) generate certificates for that domain on all servers.

However, after re-reading the spec, I noticed the letsencrypt server can request re-authorization at any time, which means this workflow won't actually work reliably. So: I agree -- the client part isn't a good idea.

As for the `acme` part, I'd be happy to clean it up and add a proper deprecation. I'm not entirely sure what the cleanest option is -- the most obvious option is to make both authzrs and new_cert_uri optional in the function definition, a la

``` python
def request_issuance(self, csr, authzrs=None, new_cert_uri=None):
    if authzrs is None and new_cert_uri is None:
        raise TypeError('request_issuance() requires new_cert_uri to be specified')
    if authzrs is not None and new_cert_uri is not None:
        raise TypeError('request_issuance() requires either authzrs or new_cert_uri to be specified')
    if (authzrs is not None):
        new_cert_uri = authzrs[0].new_cert_uri
        warnings.warn("request_issuance(..., authzrs) is deprecated. Use request_issuance(..., new_cert_uri=...) instead.")
```

which does feel like a lot of code for a minor improvement. The other option is to have

``` python
def request_issuance_by_uri(self, csr, new_cert_uri):
    (...)

def request_issuance_by_authzrs(self, csr, authzrs):
    new_cert_uri = authzrs[0].new_cert_uri
    return self.request_issuance_by_uri(csr, new_cert_uri)

def request_issuance(self, csr, authzrs):
    warnings.warn("request_issuance(..., authzrs) is deprecated. Use request_issuance_by_authzrs or request_issuance_by_uri instead.")
    return request_issuance_by_authzrs(self, csr, authzrs)
```

Lastly, the function could 'magically' accept both strings and authz objects, but I'm not a big fan of that option. I'd prefer the last option, but please let me know which one would fit best in the existing code base. I'm also OK with dropping the change altogether.

kuba:
Closing this in favour of #2474.

PR #2018, opened 2015-12-25
[letstest] create and reuse a persistent boulder AMI

### LINKED ISSUE #2017: letstest should use a persistent boulder instance
That should reduce test farm run times by ~5 minutes

pde:
Closes #2017


pde:
Merging this, since it isn't user-facing, and having it unmerged is making some testing trickier.

PR #2020, opened 2015-12-25
The actual 0.1.1 release

pde:
Due to bug #1966, the previous v0.1.1 tag was missing a version change to
letsencrypt/__init__.py this commit and the v0.1.1-corrected tag are the code
that was signed and uploaded to PyPI.

### DIFF
--- a/letsencrypt-compatibility-test/setup.py
+++ b/letsencrypt-compatibility-test/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 from setuptools import find_packages
 
 
-version = '0.2.0.dev0'
+version = '0.1.1'
 
 install_requires = [
     'letsencrypt=={0}'.format(version),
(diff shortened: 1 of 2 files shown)


pde:
Not sure what's going on with these test failures; the tests are passing for me locally on both of these commits, and at this point I do not see the string 0.2.0 anywhere in the tree...

kuba:
So are you telling me that you uploaded some random tag (that you haven't even tested) rather than the actual code that was used in the release? :( https://github.com/letsencrypt/letsencrypt/releases/tag/v0.1.1-corrected points at edd20a8... I think I just lost any trust in the release process of this software, I'm afraid.

pde:
No I tested the tag locally before uploading, and I currently believe it is fine (when I download the tarball of the tag from github, tests pass).  Not sure what travis is doing, yet...

kuba:
Tag that you uploaded does not contain 9d50c3e from this PR...

kuba:
https://github.com/letsencrypt/letsencrypt/issues/2019#issuecomment-167263396

PR #2021, opened 2015-12-25
Revert "Issue 2002" (logging function was actually calling an interactive iDisplay notification)

pde:
Reverts letsencrypt/letsencrypt#2009


pde:
Reverting to unbreak test farm tests (and also potentially automation by dev branch users, though that's a bit less likely)

PR #2030, opened 2015-12-27
Set python version in the travis matrix

evilaliv3:
This commit enforce travis to test against python version 2.7.

In fact currently even if the file travis.yml specify to use python language, not specifying the version in the test matrix causes the environment to be not initialized/optimized for python testing as shown in the following screenshot:
screenshot from 2015-12-27 12 54 09

I've also removed ```python``` from the manually installed packages, probably the same could be done for ```python-dev``` and ```python-virtualenv```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,16 @@ env:
   global:
     - GOPATH=/tmp/go
     - PATH=$GOPATH/bin:$PATH
-  matrix:
-    - TOXENV=py26 BOULDER_INTEGRATION=1
-    - TOXENV=py27 BOULDER_INTEGRATION=1
-    - TOXENV=lint
-    - TOXENV=cover
+matrix:
+  include:
+    - python: "2.6"
+      env: TOXENV=py26 BOULDER_INTEGRATION=1
+    - python: "2.7"
+      env: TOXENV=py27 BOULDER_INTEGRATION=1
+    - python: "2.7"
+      env: TOXENV=lint
+    - python: "2.7"
+      env: TOXENV=cover
 
 
 # Only build pushes to the master branch, PRs, and branches beginning with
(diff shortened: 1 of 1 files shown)


evilaliv3:
@pde: if i'm not wrong you would like to test both on python 2.6 and 2.7, that by the way was not done properly ans your were still testing python2.6 code on python2.7.

with this commit this is done properly; you can see the difference in this screenshot:
screenshot from 2015-12-27 22 06 11

in addition a thing that you should consider is that travis allows the possibility to speparate tests in two categories:

1. tests that would not allow failures
2. tests that would allow failures

separating among those two categoriees makes possbile to get the build status be finalized (success/failure) immediatly after the category 1 has succeeded. this way also one can start working on improvals without impacting the global status of the main tests

kuba:
You are wrong, `TOXENV=py26` enforces Python 2.6 when tox runs tests. Screenshots that you pasted, or rather green vs red, is likely unrelated to the Python version in build matrix... Python 2.7 is the default, so this PR is completely a NOP. tox probably always runs itself using 2.7 interpreter (I guess it comes from OS package, and is not related to Python version in build matrix).

alex:
This is a no-op, except it makes the travis matrix use accurate icons.

evilaliv3:
i see, i'm not so into Tox so you both should be right.

by the way it makes certain that the python binary and few symlinks links to the right python version (2.6 againt 2.7) and using this specific environments the build should be a little faster instead of apt-get install python that it is currently done.

pde:
@alex how does it change the icons in use? Are you saying "+1 to merge"?

alex:
@pde sorry its not the icon, just the text. it changes that text because travis uses the python key to determine that. it's a noop because tox respects the TOXENV var. we use the same tactic on pyca/cryptography, +1 from me.

PR #2031, opened 2015-12-27
Enable TravisCI apt packages cache

evilaliv3:
This commit enable travisCI cachin for the apt packages so to speed up build timing.

Reference: https://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ branches:
 # container-based infrastructure
 sudo: false
 
+# enable apt packages caching in order to speed up builds
+cache: apt
+
 addons:
   # make sure simplehttp simple verification works (custom /etc/hosts)
   hosts:


watercrossing:
Why not also cache pip packages?

evilaliv3:
thanks @watercrossing, i already thought of this but i was thinking that this can be a little tricky.

tox creates a new virtualenv for each environment and i think the package directory may not be the same and use different package versions for each python version.

bmw:
@evilaliv3, checking out the link you provided, it appears that APT caching is only available for private repositories and for non-container based infrastructure. Since we're neither of those things, I'm not sure this will work for us.

evilaliv3:
things are a little confused in the documentation of travis, but the cache as by [1] should be available since you use the container-based infrastructure.

[1] https://docs.travis-ci.com/user/ci-environment/

evilaliv3:
let me say that i was experimenting the Apt cache with this pull request while since now i've only used the pip one that is for sure available to every repository that is using the container-based infrastructure so it's possible that you are right :)

on of you that has privileges on the repository could check at https://travis-ci.org/letsencrypt/letsencrypt/jobs/98963224 and if you go where there is "Settings" you should see a new voice "Cache" if it is working.

bmw:
Doesn't seem like it's working?

https://ibin.co/2TG3gn50gZcm

bmw:
Since this doesn't seem to work with our current infrastructure, I'm closing this PR.

If this functionality becomes available for us in the future, don't hesitate to reopen.

PR #2038, opened 2015-12-28
Abstract the remaining OS specific constants from the code

### LINKED ISSUE #2034: Need to be able to define the files parsed in Apache vhost directory
While in some distributions all the configuration files should have .conf - extension, which is the default after #1811 - some distributions use filenames without extension. However for example CentOS has non-configuration files in the same directory as vhost .conf  - files, and augeas errors out on them.

We will need to add the option to define files to be parsed per-os basis.

joohoi:
Fix to enable us to support broader range of different configuration schemes - some of which are OS defaults, and abstract remaining hardcoded but still OS dependent values.

Fixes #2034


pde:
LGTM!

PR #2039, opened 2015-12-29
Allow old versions to fail gracefully

denmat:
Context:
Old versions of python fail with bash test errors because `python --version` is not available for old versions of python.

```
./letsencrypt-auto: line 104: [: too many arguments
./letsencrypt-auto: line 106: [: too many arguments
```

Change:
Old and new versions of python support `-V` instead of `--version`. This allow them to pass the version tests and fail with 'ancient version' version message.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -100,7 +100,7 @@ DeterminePythonVersion() {
     exit 1
   fi
 
-  PYVER=`$LE_PYTHON --version 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//'`
+  PYVER=`$LE_PYTHON -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//'`
   if [ $PYVER -eq 26 ] ; then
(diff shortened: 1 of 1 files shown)


alex:
When you say old, how old are you talking about?

denmat:
```
$ python -V
Python 2.4.3
$ cat /etc/redhat-release
CentOS release 5.8 (Final)
```

sigmavirus24:
Does Red Hat not have software collections for Cent such that you could use a more recent and secure version of Python?

denmat:
Oh, it's not the oldness that worried me, it is just the bash failure. `2.6` is installed on the host but `/usr/bin/python` is linked to `2.4`. There is a message to say 'hey you've got an old version', but you never get there.

alex:
This makes me a bit sad, because `-V` is slightly less readable than `--version`, and damn 2.4 is old, but I think it makes sense and is worth landing.

denmat:
Yeah, `-1` for clarity of code, `+1` for running cleanly. :)

erikrose:
@denmat Thanks for the catch!

PR #2040, opened 2015-12-29
update using.rst show example for crontab renewal. [needs revision]

peacetara:
Show an example for renewal using command line only.


bmw:
The problem with this is the client can still prompt the user causing the script to fail. An example of how this can occur is if the Apache plugin cannot determine which vhost to use.

Currently, there's not a great way to automate the renewal process, but after releasing `0.2.0` in the next couple days, automatic renewal will be the primary focus of the client team.

bmw:
We're currently working on improving our renewal documentation right now and this is largely outdated since the release of `letsencrypt renew`. Closing this PR. If you think I should reconsider, please let me know!

PR #2048, opened 2015-12-31
Add bootstrap for OpenBSD

delan:


pde:
@smutt @sboyettedh would either of you like to review this?  Some fraction of OpenBSD users now have packages, but it's probably fine to offer experimental letsencrpt-auto support too.

pde:
Also, if this is good, a reminder that we should close #1175.

smutt:
I no longer have any domains that need https so I can't test this. I ended up using letsencrypt-nosudo.

Thanks very much for doing this work. Your efforts will increase adoption of HTTPS.

erikrose:
We now have official OpenBSD packages, which can integrate with the system better than le-auto. If there's a compelling reason to add le-auto support regardless, please reopen.

PR #2052, opened 2015-12-31
Update DNS challenge code for the latest version of the ACME spec.

alex:
This now implements the DNS-01 challenge and response. Thus far I've removed all the old tests and not added any new ones.

Would like some feedback on whether this looks remotely correct, and then I'll go about adding tests for this code.


alex:
Being able to use this in practice is blocked on https://github.com/letsencrypt/boulder/pull/1295

alex:
Closing in favor of #2061.

Issue #1958, opened 2015-12-19
Move acme to own repository


trestoa:
I would suggest to move the acme part of the project to its own repository.
Doing so would make it much easier to use the python acme implementation for other projects.

alex:
It's released independently on PyPI, so it's easy for projects to just `pip install acme` without the rest of `letsencrypt`. (I dont' have an opinion on your proposal, just want to point out that it's not really a blocker)

kleiinnn:
Ah, I did not know that, but it's good to know. Then it actually makes no difference from the viewpoint of an consumer of the library.

kuba:
Working on it :)

As @alex said, though, `acme` is already packaged separately with tarballs uploaded to https://pypi.python.org/pypi/acme. Distros should package is separately as well (that's what is being done for Debian for instance).

Issue #1301, opened 2015-11-03
Too many dependencies


bmw:
Users/developers have reported that our project has too many Python dependencies. This is related to things being overly complex.

Here is pde's attempt to characterize what our dependencies currently are: 

> 1. Purely for certbot-auto and the things it needs to build, not for
>    the python code itself (these no longer apply on recent OSes where we have packages):
>    
>    ~~git~~
>    gcc 
>    python-dev
>    libffi-dev
>    libssl-dev
>    setuptools
>    virtualenv
> 2. Essential OS packages
>    python
>    openssl
> 3. Semi-essential OS packages:
>    ~~dialog~~
> 4. Needed for specific plugins:
>    
>    ~~python-psutil (standalone -- see what has bound :80 or :443)~~ (now only "recommended")
>    libaugeas0 (apache -- parse & edit conf files)
>    python-augeas
> 5. Python dependencies:
>    
>    ConfigArgParse       (for reading config files)
>    configobj            (for writing & reading renewal conf files)
>    python-cryptography (>= 0.8)
>    PyOpenSSL  ~~(>= 0.15)~~ (>= 0.13)
>    ~~pythondialog~~
>    ~~ndg-httpsclient  (for working around TLS insecurities in older Python versions)~~
>    requests
>    six 
> 6. Python dependencies specifically related to time processing. It seems
>    absurd that there are four of these:
>    
>    ~~werkzeug~~
>    pytz
>    parsedatetime
>    pyicu (needed by parsedatetime)
>    pyrfc3339
> 7. Python dependencies for python-cryptography:
>    idna
>    enum34
>    ipaddress
>    cffi
>    pycparser (build dep only, but affects letsencrypt-auto)
>    pyasn1
> 8. These were used to define plugin interfaces.  Not sure if that was wise,
>    but it's probably annoying to third parties if we ditch them:
>    
>    zope.components
>    zope.interface
>    zope.event
> 9. Development and doc-building dependencies.  For reasons I don't
>    yet understand, pip always fetches some of these too:
>    
>    mock
>    funcsigs (mock depends on this)
>    pbr (mock depends on this)
>    sphinxdocs

pde:
We should be able to remove some of them.  At some point we could go from needing argparse (which isn't a non-Python dependency), configargparse, and configobj, to needing zero or one of those.

Perhaps we one day we can get rid of PyOpenSSL.

[...]

PR #2061, opened 2016-01-02
Add dns-01 challenge support to the ACME client

wteiken:
Tested against a local dev version of the non-spec server, should also work against the soon-to-be-updated staging server.

### DIFF
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -207,7 +207,7 @@ class KeyAuthorizationChallenge(_TokenChallenge):
 
 @ChallengeResponse.register
 class DNS01Response(KeyAuthorizationChallengeResponse):
-    """ACME "dns-01" challenge response."""
+    """ACME dns-01 challenge response."""
     typ = "dns-01"
 
     def simple_verify(self, chall, domain, account_public_key):
@@ -215,7 +215,7 @@ class DNS01Response(KeyAuthorizationChallengeResponse):
 
         :param challenges.DNS01 chall: Corresponding challenge.
         :param unicode domain: Domain name being verified.
-        :param account_public_key: Public key for the key pair
+        :param JWK account_public_key: Public key for the key pair
             being authorized.
 
         :returns: ``True`` iff validation with the TXT records resolved from a
@@ -247,7 +247,7 @@ class DNS01Response(KeyAuthorizationChallengeResponse):
 
 @Challenge.register  # pylint: disable=too-many-ancestors
 class DNS01(KeyAuthorizationChallenge):
-    """ACME "dns-01" challenge."""
+    """ACME dns-01 challenge."""
     response_cls = DNS01Response
     typ = response_cls.typ
 
@@ -298,7 +298,7 @@ class HTTP01Response(KeyAuthorizationChallengeResponse):
             being authorized.
         :param int port: Port used in the validation.
 
-        :returns: ``True`` iff validation of the files currently server by the
+        :returns: ``True`` iff validation with the files currently served by the
             HTTP server is successful.
         :rtype: bool
 
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -14,7 +14,6 @@ from acme import crypto_util
 from acme import fields
 from acme import jose
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -207,6 +206,74 @@ class KeyAuthorizationChallenge(_TokenChallenge):
 
 
 @ChallengeResponse.register
+class DNS01Response(KeyAuthorizationChallengeResponse):
+    """ACME "dns-01" challenge response."""
+    typ = "dns-01"
+
+    def simple_verify(self, chall, domain, account_public_key):
+        """Simple verify.
+
+        :param challenges.DNS01 chall: Corresponding challenge.
+        :param unicode domain: Domain name being verified.
+        :param account_public_key: Public key for the key pair
+            being authorized.
+
+        :returns: ``True`` iff validation with the TXT records resolved from a
+            DNS server is successful.
+        :rtype: bool
+
+        """
+        if not self.verify(chall, account_public_key):
+            logger.debug("Verification of key authorization in response failed")
+            return False
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+
+        try:
+            from acme import dns_resolver
+        except ImportError:  # pragma: no cover
+            raise errors.Error("Local validation for 'dns-01' challenges "
+                               "requires 'dnspython'")
+        txt_records = dns_resolver.txt_records_for_name(validation_domain_name)
+        exists = validation in txt_records
+        if not exists:
+            logger.debug("Key authorization from response (%r) doesn't match "
+                         "any DNS response in %r", self.key_authorization,
+                         txt_records)
+        return exists
+
+
+@Challenge.register  # pylint: disable=too-many-ancestors
+class DNS01(KeyAuthorizationChallenge):
+    """ACME "dns-01" challenge."""
+    response_cls = DNS01Response
+    typ = response_cls.typ
+
+    LABEL = "_acme-challenge"
+    """Label clients prepend to the domain name being validated."""
+
+    def validation(self, account_key, **unused_kwargs):
+        """Generate validation.
+
+        :param JWK account_key:
+        :rtype: unicode
+
+        """
+        return jose.b64encode(hashlib.sha256(self.key_authorization(
+            account_key).encode("utf-8")).digest()).decode()
+
+    def validation_domain_name(self, name):
+        """Domain name for TXT validation record.
+
+        :param unicode name: Domain name being validated.
+
+        """
+        return "{0}.{1}".format(self.LABEL, name)
+
+
+@ChallengeResponse.register
 class HTTP01Response(KeyAuthorizationChallengeResponse):
     """ACME http-01 challenge response."""
     typ = "http-01"
@@ -231,8 +298,8 @@ class HTTP01Response(KeyAuthorizationChallengeResponse):
             being authorized.
         :param int port: Port used in the validation.
 
-        :returns: ``True`` iff validation is successful, ``False``
-            otherwise.
+        :returns: ``True`` iff validation of the files currently server by the
+            HTTP server is successful.
         :rtype: bool
 
         """
@@ -410,7 +477,7 @@ class TLSSNI01Response(KeyAuthorizationChallengeResponse):
 
 
         :returns: ``True`` iff client's control of the domain has been
-            verified, ``False`` otherwise.
+            verified.
         :rtype: bool
 
         """
--- a/acme/acme/challenges_test.py
+++ b/acme/acme/challenges_test.py
@@ -77,6 +77,93 @@ class KeyAuthorizationChallengeResponseTest(unittest.TestCase):
         self.assertFalse(response.verify(self.chall, KEY.public_key()))
 
 
+class DNS01ResponseTest(unittest.TestCase):
+    # pylint: disable=too-many-instance-attributes
+
+    def setUp(self):
+        from acme.challenges import DNS01Response
+        self.msg = DNS01Response(key_authorization=u'foo')
+        self.jmsg = {
+            'resource': 'challenge',
+            'type': 'dns-01',
+            'keyAuthorization': u'foo',
+        }
+
+        from acme.challenges import DNS01
+        self.chall = DNS01(token=(b'x' * 16))
+        self.response = self.chall.response(KEY)
+
+    def test_to_partial_json(self):
+        self.assertEqual(self.jmsg, self.msg.to_partial_json())
+
+    def test_from_json(self):
+        from acme.challenges import DNS01Response
+        self.assertEqual(self.msg, DNS01Response.from_json(self.jmsg))
+
+    def test_from_json_hashable(self):
+        from acme.challenges import DNS01Response
+        hash(DNS01Response.from_json(self.jmsg))
+
+    def test_simple_verify_bad_key_authorization(self):
+        key2 = jose.JWKRSA.load(test_util.load_vector('rsa256_key.pem'))
+        self.response.simple_verify(self.chall, "local", key2.public_key())
+
+    @mock.patch("acme.dns_resolver.txt_records_for_name")
+    def test_simple_verify_good_validation(self, mock_resolver):
+        mock_resolver.return_value = [self.chall.validation(KEY.public_key())]
+        self.assertTrue(self.response.simple_verify(
+            self.chall, "local", KEY.public_key()))
+        mock_resolver.assert_called_once_with(
+            self.chall.validation_domain_name("local"))
+
+    @mock.patch("acme.dns_resolver.txt_records_for_name")
+    def test_simple_verify_good_validation_multiple_txts(self, mock_resolver):
+        mock_resolver.return_value = [
+            "!", self.chall.validation(KEY.public_key())]
+        self.assertTrue(self.response.simple_verify(
+            self.chall, "local", KEY.public_key()))
+        mock_resolver.assert_called_once_with(
+            self.chall.validation_domain_name("local"))
+
+    @mock.patch("acme.dns_resolver.txt_records_for_name")
+    def test_simple_verify_bad_validation(self, mock_dns):
+        mock_dns.return_value = ["!"]
+        self.assertFalse(self.response.simple_verify(
+            self.chall, "local", KEY.public_key()))
+
+
+class DNS01Test(unittest.TestCase):
+
+    def setUp(self):
+        from acme.challenges import DNS01
+        self.msg = DNS01(token=jose.decode_b64jose(
+            'evaGxfADs6pSRb2LAv9IZf17Dt3juxGJ+PCt92wr+oA'))
+        self.jmsg = {
+            'type': 'dns-01',
+            'token': 'evaGxfADs6pSRb2LAv9IZf17Dt3juxGJ-PCt92wr-oA',
+        }
+
+    def test_validation_domain_name(self):
+        self.assertEqual('_acme-challenge.www.example.com',
+                         self.msg.validation_domain_name('www.example.com'))
+
+    def test_validation(self):
+        self.assertEqual(
+            "rAa7iIg4K2y63fvUhCfy8dP1Xl7wEhmQq0oChTcE3Zk",
+            self.msg.validation(KEY))
+
+    def test_to_partial_json(self):
+        self.assertEqual(self.jmsg, self.msg.to_partial_json())
+
+    def test_from_json(self):
+        from acme.challenges import DNS01
+        self.assertEqual(self.msg, DNS01.from_json(self.jmsg))
+
+    def test_from_json_hashable(self):
+        from acme.challenges import DNS01
+        hash(DNS01.from_json(self.jmsg))
+
+
 class HTTP01ResponseTest(unittest.TestCase):
     # pylint: disable=too-many-instance-attributes
 
--- a/acme/acme/dns_resolver.py
+++ b/acme/acme/dns_resolver.py
@@ -1,0 +1,28 @@
+"""DNS Resolver for ACME client.
+Required only for local validation of 'dns-01' challenges.
+"""
+import logging
+
+import dns.resolver
+import dns.exception
+
+logger = logging.getLogger(__name__)
+
+
+def txt_records_for_name(name):
+    """Resolve the name and return the TXT records.
+
+    :param unicode name: Domain name being verified.
+
+    :returns: A list of txt records, if empty the name could not be resolved
+    :rtype: list of unicode
+
+    """
+    try:
+        dns_response = dns.resolver.query(name, 'TXT')
+    except dns.resolver.NXDOMAIN as error:
+        return []
+    except dns.exception.DNSException as error:
+        logger.error("Error resolving %s: %s", name, str(error))
+        return []
+    return [txt_rec for rdata in dns_response for txt_rec in rdata.strings]
--- a/acme/acme/dns_resolver.py
+++ b/acme/acme/dns_resolver.py
@@ -25,4 +25,6 @@ def txt_records_for_name(name):
     except dns.exception.DNSException as error:
         logger.error("Error resolving %s: %s", name, str(error))
         return []
-    return [txt_rec for rdata in dns_response for txt_rec in rdata.strings]
+
+    return [txt_rec.decode("utf-8") for rdata in dns_response
+            for txt_rec in rdata.strings]
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -35,6 +35,15 @@ if sys.version_info < (2, 7):
 else:
     install_requires.append('mock')
 
+if sys.version_info < (3, 0):
+    dns_extras = [
+        'dnspython',
+    ]
+else:
+    dns_extras = [
+        'dnspython3',
+    ]
+
 dev_extras = [
     'nose',
     'pep8',
@@ -76,6 +85,7 @@ setup(
     include_package_data=True,
     install_requires=install_requires,
     extras_require={
+        'dns': dns_extras,
         'dev': dev_extras,
         'docs': docs_extras,
     },
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -35,14 +35,10 @@ if sys.version_info < (2, 7):
 else:
     install_requires.append('mock')
 
-if sys.version_info < (3, 0):
-    dns_extras = [
-        'dnspython',
-    ]
-else:
-    dns_extras = [
-        'dnspython3',
-    ]
+# dnspython 1.12 is required to support both Python 2 and Python 3.
+dns_extras = [
+    'dnspython>=1.12',
+]
 
 dev_extras = [
     'nose',
--- a/acme/acme/dns_resolver_test.py
+++ b/acme/acme/dns_resolver_test.py
@@ -1,0 +1,53 @@
+"""Tests for acme.dns_resolver."""
+import unittest
+import mock
+
+from acme import dns_resolver
+
+try:
+    import dns
+except ImportError:  # pragma: no cover
+    dns = None
+
+
+def create_txt_response(name, txt_records):
+    """
+    Returns an RRSet containing the 'txt_records' as the result of a DNS
+    query for 'name'.
+
+    This takes advantage of the fact that an Answer object mostly behaves
+    like an RRset.
+    """
+    return dns.rrset.from_text_list(name, 60, "IN", "TXT", txt_records)
+
+
+class TxtRecordsForNameTest(unittest.TestCase):
+
+    @mock.patch("acme.dns_resolver.dns.resolver.query")
+    def test_txt_records_for_name_with_single_response(self, mock_dns):
+        mock_dns.return_value = create_txt_response('name', ['response'])
+        self.assertEqual(['response'],
+                         dns_resolver.txt_records_for_name('name'))
+
+    @mock.patch("acme.dns_resolver.dns.resolver.query")
+    def test_txt_records_for_name_with_multiple_responses(self, mock_dns):
+        mock_dns.return_value = create_txt_response(
+            'name', ['response1', 'response2'])
+        self.assertEqual(['response1', 'response2'],
+                         dns_resolver.txt_records_for_name('name'))
+
+    @mock.patch("acme.dns_resolver.dns.resolver.query")
+    def test_txt_records_for_name_domain_not_found(self, mock_dns):
+        mock_dns.side_effect = dns.resolver.NXDOMAIN
+        self.assertEquals([], dns_resolver.txt_records_for_name('name'))
+
+    @mock.patch("acme.dns_resolver.dns.resolver.query")
+    def test_txt_records_for_name_domain_other_error(self, mock_dns):
+        mock_dns.side_effect = dns.exception.DNSException
+        self.assertEquals([], dns_resolver.txt_records_for_name('name'))
+
+    def run(self, result=None):
+        if dns is None:  # pragma: no cover
+            print(self, "... SKIPPING, no dnspython available")
+            return
+        super(TxtRecordsForNameTest, self).run(result)
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@ envlist = py{26,33,34,35},cover,lint
 # packages installed separately to ensure that downstream deps problems
 # are detected, c.f. #1002
 commands =
-    pip install -e acme[dev]
+    pip install -e acme[dns,dev]
     nosetests -v acme
     pip install -e .[dev]
     nosetests -v certbot
@@ -38,23 +38,23 @@ deps =
 
 [testenv:py33]
 commands =
-    pip install -e acme[dev]
+    pip install -e acme[dns,dev]
     nosetests -v acme
 
 [testenv:py34]
 commands =
-    pip install -e acme[dev]
+    pip install -e acme[dns,dev]
     nosetests -v acme
 
 [testenv:py35]
 commands =
-    pip install -e acme[dev]
+    pip install -e acme[dns,dev]
     nosetests -v acme
 
 [testenv:cover]
 basepython = python2.7
 commands =
-    pip install -e acme[dev] -e .[dev] -e certbot-apache -e certbot-nginx -e letshelp-certbot
+    pip install -e acme[dns,dev] -e .[dev] -e certbot-apache -e certbot-nginx -e letshelp-certbot
     ./tox.cover.sh
 
 [testenv:lint]
@@ -64,7 +64,7 @@ basepython = python2.7
 # duplicate code checking; if one of the commands fails, others will
 # continue, but tox return code will reflect previous error
 commands =
-    pip install -q -e acme[dev] -e .[dev] -e certbot-apache -e certbot-nginx -e certbot-compatibility-test -e letshelp-certbot
+    pip install -q -e acme[dns,dev] -e .[dev] -e certbot-apache -e certbot-nginx -e certbot-compatibility-test -e letshelp-certbot
     ./pep8.travis.sh
     pylint --reports=n --rcfile=.pylintrc certbot
     pylint --reports=n --rcfile=acme/.pylintrc acme/acme


acme/acme/challenges.py
```diff
@@ -216,6 +219,81 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+            dns_response = dns.resolver.query(validation_domain_name, 'TXT')
+            txt_records = sum([rdata.strings for rdata in dns_response], [])
+        except dns.exception.DNSException as error:
+            logger.error("Unable to resolve %s: %s", validation_domain_name,
+                         error)
+            return False
+
+        for txt_record in txt_records:
+            if txt_record == validation:
+                return True
```
alex: I think this code can be simplified to:

```py
try:
    dns_response = dns.resolver.query(validation_domain_name, 'TXT')
except dns.exception.DNSException as error:
    logger.error(...)
    return False

for rdata in dns_response:
    for txt_record in rdata.strings:
        if txt_record == validation:
            return True
```

acme/acme/challenges.py
```diff
@@ -216,6 +219,81 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        """Generate validation.
+
+        :param JWK account_key:
+        :rtype: unicode
+
+        """
+        key_authorization = self.key_authorization(account_key)
+        if dns01_hexdigit_response:
+            return hashlib.sha256(key_authorization).hexdigest()
+        return base64.urlsafe_b64encode(hashlib.sha256(key_authorization).digest())
```
alex: On reading the discussion on https://github.com/letsencrypt/boulder/pull/1295 I think this maybe should be `jose.b64encode`? What do you think?

alex:
Not sure if it's a bug here or a bug in boulder or a bug in my code, but running https://github.com/alex/letsencrypt-aws/ I pass `simple_verify` but get an error back from the staging server.

wteiken:
For dns-01 validations?   The staging server does not support that yet, it’s on an older branch without DNS-01 support.  I had to run my tests against a local server to verify it works.

alex:
Ah, for some reason I thought staging had DNS-01 deployed. Sorry for the
noise, thanks for working on this!

On Sat, Jan 2, 2016 at 12:31 PM, Wilfried Teiken <notifications@github.com>
wrote:

> For dns-01 validations? The staging server does not support that yet, it’s
> on an older branch without DNS-01 support. I had to run my tests against a
> local server to verify it works.
>
>
> > On Jan 2, 2016, at 11:29 AM, Alex Gaynor <notifications@github.com>
> wrote:
> >
> > Not sure if it's a bug here or a bug in boulder or a bug in my code, but
> running https://github.com/alex/letsencrypt-aws/ I pass simple_verify but
> get an error back from the staging server.
>
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/letsencrypt/letsencrypt/pull/2061#issuecomment-168411126>
> .
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

wteiken:
I actually ran into the same problem and was pulling out my hair before I realized where the problem was.

Just in case you are not aware of it:  https://acme-staging.api.letsencrypt.org/build tells you the running build.

acme/acme/challenges.py
wteiken: I think it makes sense to use one common encoding, but the spec needs to be updated for that (the last drat I looked at did not even specify urlsafe base64).

The discussion cites passages related to JSON, but this is not a JSON based field so that does not clarify the issue.  It is an implementation detail that the server does a byte-to-byte comparison of the encoded values, so it is not necessary (although preferred) that the DNS record and the JSON objects use identical encoding.

acme/acme/challenges.py
wteiken: I factored out the dns query, I'd rather contain the dnspython objects as much as I can.  And I think it gives some extra structure that helps.

acme/acme/challenges.py
```diff
@@ -216,6 +219,89 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+
+        """
+        if not self.verify(chall, account_public_key):
+            logger.debug("Verification of key authorization in response failed")
+            return False
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+        for txt_record in self.txt_records_for_domain(validation_domain_name):
```
alex: If there's an error in DNS querying now it'll fail with a `TypeError: NoneType is not iterable`.

wteiken: Yep, just pushed the fixed version.  I ran the wrong test command.

acme/acme/challenges.py
```diff
@@ -216,6 +219,93 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        """
+        if not self.verify(chall, account_public_key):
+            logger.debug("Verification of key authorization in response failed")
+            return False
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+        txt_records = self.txt_records_for_name(validation_domain_name)
+        if txt_records == None:
```
alex: `txt_records is None` is more idiomatic (and I think `pylint` might complain :-))

acme/acme/challenges.py
```diff
@@ -216,6 +219,93 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        """Simple verify.
+
+        :param challenges.DNS01 chall: Corresponding challenge.
+        :param unicode domain: Domain name being verified.
+        :param account_public_key: Public key for the key pair
+            being authorized. If ``None`` key verification is not
+            performed!
+        :param JWK account_public_key:
+
+        :returns: ``True`` iff validation is successful, ``False``
```
onovy: iff -> if

wteiken: I'd rather remove the superfluous "False otherwise" part.  Same in rest of doc.

acme/acme/challenges.py
janeczku: @wteiken See this issue on the acme draft repository https://github.com/ietf-wg-acme/acme/issues/64

acme/acme/challenges.py
```diff
@@ -216,6 +219,93 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+    def validation(self, account_key, dns01_hexdigit_response=True, **unused_kwargs):
+        """Generate validation.
+
+        :param JWK account_key:
+        :rtype: unicode
+
+        """
+        key_authorization = self.key_authorization(account_key)
+        if dns01_hexdigit_response:
+            return hashlib.sha256(key_authorization).hexdigest()
```
alex: The code was landed in `boulder` to fix the hex stuff, so I think this fallback can be removed.

jmhodges: Very much agreed that we remove that fallback! Boulder will be fixed Real Soon Now and abiding by Postel's Law is a quick way to build insecure systems. There's also a ticket to [clarify the spec](https://github.com/ietf-wg-acme/acme/issues/64) on this matter.

wteiken: Yes, it was only in there in case a server gets released with the old implementation.   I will remove the various references (and replace the encoding with jose.b64encode).

jmhodges: (To be clear, the DNS challenge in boulder is fixed in master and will soon be deployed to staging.)

wteiken: Done

acme/setup.py
```diff
@@ -12,6 +12,7 @@
     'cryptography>=0.8',
     # Connection.set_tlsext_host_name (>=0.13), X509Req.get_extensions (>=0.15)
     'PyOpenSSL>=0.15',
+    'dnspython',
```
kuba: Unfortunately, this adds additional dependency to the project. Moreover, some of the clients using this library (e.g. https://github.com/kuba/simp_le) will never use DNS challenge. Unless there are some technical arguments against, I would strongly prefer to move `dnspython` to `extras_require` (say `dns`) and make sure the rest of the code is importable when `dnspython` is not installed (i.e. catching `ImportError` or alternatively importing at the non-module scope - I'm not sure what's better here).

acme/acme/challenges.py
```diff
@@ -1,5 +1,7 @@
 """ACME Identifier Validation Challenges."""
 import abc
+import dns.resolver
+import dns.exception
```
kuba: Those 2 imports should be separated from stdlib imports group.

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        :returns: A list of txt records, or None if the name could not be resolved
+        :rtype: list of unicode
+
+        """
+        try:
+            dns_response = dns.resolver.query(name, 'TXT')
+        except dns.exception.DNSException as error:
+            logger.error("Unable to resolve %s: %s", name, error)
+            return None
+        return sum([rdata.strings for rdata in dns_response], [])
```
kuba: Why do you use `sum()` here? This should rather be just `return [rdata.strings for rdata in dsn_response`

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
 
 
 @ChallengeResponse.register
+class DNS01Response(KeyAuthorizationChallengeResponse):
+    """ACME "dns-01" challenge response."""
+    typ = "dns-01"
+
+    def txt_records_for_name(self, name):
```
kuba: `self` is not used :( if this passes tests then that means that `pylint` config has been messed up

@pde I still feel like a robot :(

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+
+        :returns: A list of txt records, or None if the name could not be resolved
+        :rtype: list of unicode
+
+        """
+        try:
+            dns_response = dns.resolver.query(name, 'TXT')
+        except dns.exception.DNSException as error:
+            logger.error("Unable to resolve %s: %s", name, error)
+            return None
```
kuba: it'd be a lot easier if you returned `[]` here

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        """
+        if not self.verify(chall, account_public_key):
+            logger.debug("Verification of key authorization in response failed")
+            return False
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+        txt_records = self.txt_records_for_name(validation_domain_name)
+        if txt_records is None:
```
kuba: I would drop this test in favour of returning `[]` from `txt_records_for_name`

acme/acme/challenges.py
wteiken: Yep, it passes the tests.  Will change and move it out of the class.

acme/acme/challenges.py
wteiken: this is a list of lists (multiple rdata objects with possibly multiple strings), so this flattens that structure.

acme/setup.py
wteiken: I'll have a look at that (new to python).

I see your point, but I feel it's a little odd to have something covered by the core standard require an "extras" library.  And given the core standard includes DNS queries some DNS library dependency is to be expected.

acme/acme/challenges.py
kuba: Please use `[rec for rdata in dns_response for rec in rdata.strings]`. More readable and more efficient.

acme/acme/challenges.py
kuba: Agreed about containing `dnspython`. However, too many loops for my taste. This would be more Pythonic and likely more efficient:

```python
exists = validation in txt_records
if not exists:
  logger.debug(...)
return exists
```

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        return sum([rdata.strings for rdata in dns_response], [])
+
+    def simple_verify(self, chall, domain, account_public_key):
+        """Simple verify.
+
+        :param challenges.DNS01 chall: Corresponding challenge.
+        :param unicode domain: Domain name being verified.
+        :param account_public_key: Public key for the key pair
+            being authorized. If ``None`` key verification is not
+            performed!
```
kuba: Comment about `None` is not true, I believe. (I will fix it in a separate PR for other occurrences)

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+                return True
+
+        logger.debug("Key authorization from response (%r) doesn't match any "
+                     "DNS response in %r", self.key_authorization, txt_records)
+        return False
+
+@Challenge.register  # pylint: disable=too-many-ancestors
+class DNS01(KeyAuthorizationChallenge):
+    """ACME "dns-01" challenge."""
+
```
kuba: pls drop whitespace here

acme/acme/challenges.py
```diff
@@ -216,6 +218,90 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+    def validation(self, account_key, **unused_kwargs):
+        """Generate validation.
+
+        :param JWK account_key:
+        :rtype: unicode
+
+        """
+        key_authorization = self.key_authorization(account_key)
+        return jose.b64encode(hashlib.sha256(key_authorization).digest())
```
kuba: let's inline `key_authorization`

acme/setup.py
kuba: Basic DNS queries can by performed by stdlib. On the other hand, official client (rightfully) gets a lot of critique due to too many dependencies (see #1301). :(

acme/acme/challenges_test.py
```diff
@@ -76,6 +77,113 @@ def test_verify_wrong_form(self):
             key_authorization='.foo.oKGqedy-b-acd5eoybm2f-NVFxvyOoET5CNy3xnv8WY')
         self.assertFalse(response.verify(self.chall, KEY.public_key()))
 
+class DNS01ResponseTest(unittest.TestCase):
```
kuba: nit: please make sure there are 2 separating lines between classes

acme/acme/challenges_test.py
```diff
@@ -76,6 +77,113 @@ def test_verify_wrong_form(self):
...
+        mock_dns.side_effect = dns.exception.DNSException
+        self.assertFalse(self.response.simple_verify(
+            self.chall, "local", KEY.public_key()))
+
+class DNS01Test(unittest.TestCase):
+
+    def setUp(self):
+        from acme.challenges import DNS01
+        self.msg = DNS01(
+            token=jose.decode_b64jose(
```
kuba: nit: you can squeeze this into the previous line

kuba:
Nice job, thanks! :) Couple of nits to fix, couple of idiomatic improvements and possibly-tricky-to-implement `extras_require` approach of declaring `dnspython` dependency before next round of review.

acme/acme/challenges.py
wteiken: Done

acme/acme/challenges.py
wteiken: Done

acme/acme/challenges.py
wteiken: Done

acme/acme/challenges.py
wteiken: Done

acme/acme/challenges_test.py
wteiken: Done

acme/acme/challenges_test.py
wteiken: Done

wteiken:
Successfully tested the new library against the "Boulder=( +34c1b83 Tue Jan  5 22:16:57 UTC 2016)" staging server.

wteiken:
I have made some progress isolating dnspython, it is now in a separate module which means challenges and challenges_test have no direct dependency on dnspython anymore (in fact, only calling simple_verify will have any issues without it).  I moved the dependency to extras_require (via testing_extras).

The functional tests on travis-ci succeed, but lint and coverage fail (as they don't install the extras_require for 'testing'), see https://travis-ci.org/wteiken/letsencrypt/builds/100522564.  So problem one will be to make these two get the right dependency.

Another question is how the tests for developers running the tests locally, both as part of the client install and as part of using the ACME client e.g., via PIP.  I don't know enough about that part to make any predictions...

kuba:
It seems that you haven't pushed commits with isolated `dnspython`. Intended?

For Travis, you will need to update `tox.ini` to call `pip install -e acme[dns,testing]` instead of just `pip install -e acme[testing]`.

Yeah, we should have some kind of integration testing for the DNS challenge. All the other tests are the the core client (`letsencrypt`) level, rather than `acme` library (this might change in future, in order to get `acme` library separated from this repo, cf. #1958), so I just assumed you were going to create a follow-up PR with new DNS `IPlugin` and integration tests including deployment of local temporary DNS server.

acme/acme/challenges_test.py
kuba: Forgot to push commits? To be clear, I meant: `self.msg = DNS01(token=jose.decode_b64jose(\n`

wteiken:
Yes, I am trying to have it pass checks first :-)  I assume otherwise the chatter will be a lot higher for you.
I do the travis-cl tests from my fork, once that is done I will merge the change to the branch linked to the PR.

acme/acme/challenges_test.py
wteiken: Hah, I accidentally fixed another line instead that was similar.  Will be fixed in the next merge.

wteiken:
K, this should work.

I was more wondering how the tests are run outside of travis.  The new dns_resolver has a unit test that will of course fail if dnsresolver is not installed....  so there is probably some configuration to translate the option (dns) into which tests to include.

acme/acme/dns_resolver.py
```diff
@@ -0,0 +1,28 @@
...
+
+    :param unicode name: Domain name being verified.
+
+    :returns: A list of txt records, if empty the name could not be resolved
+    :rtype: list of unicode
+
+    """
+    try:
+        dns_response = dns.resolver.query(name, 'TXT')
+    except ImportError as error:  # pragma: no cover
```
kuba: `dns.resolver.query` won't raise an import error - that error would happen at the top of the file (`import dns.resolver`)

acme/acme/dns_resolver.py
```diff
@@ -0,0 +1,28 @@
...
+        raise ImportError("Local validation for 'dns-01' challenges requires "
```
kuba: I wouldn't necessarily hide one import error using another import error, `import dns` (even if unused) will give a good guidance on which module is missing

kuba:
NB One of the standard ways of dealing with optional deps I've seen around in other Python projects is:
```python
try:
  import foo
except ImportError:
  # optionally logger.warning('all function depending on foo will not work because it's not importable')
  foo = None
...
def bar():
  if foo is None:
    raise Error('foo is necessary for bar to work, please install it')
```

Does anyone else have any ideas, suggestions?

acme/acme/dns_resolver.py
wteiken: Yep, leftover from an earlier version.  It seems the import must be global so the mocking works properly.

acme/acme/dns_resolver.py
wteiken: Thinking about it it seems that errors.Error is more appropriate.

acme/acme/challenges.py
```diff
@@ -216,6 +215,76 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+
+        try:
+            from acme import dns_resolver
+            txt_records = dns_resolver.txt_records_for_name(
+                validation_domain_name)
+        except ImportError:  # pragma: no cover
```
kuba: In general, please restrict `try` block to minimum, ie. only `from acme import dns_resolver` line.

However, it's still not clear whether `dns_resolver` module approach is best to handle this PR (see main thread for my comment)

wteiken: Assuming in the long run the DNS interaction may be extended (e.g., checking DANE records for some future validation, or explicitly validating DNSSEC) I figured we might as well move the DNS specific code into a separate module anyway.  I am more wondering if a little more should be pushed to the DNS specific file.

The other option of course would be to modularize this code and have challenge-specific files.  That way the whole challenge handling could be loaded via an options file.  But even then the DNS dependency only applies to simple_verify, which is not strictly needed even if DNS challenges are supported by the client.

acme/acme/challenges.py
```diff
@@ -216,6 +215,75 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+    typ = "dns-01"
+
+    def simple_verify(self, chall, domain, account_public_key):
+        """Simple verify.
+
+        :param challenges.DNS01 chall: Corresponding challenge.
+        :param unicode domain: Domain name being verified.
+        :param account_public_key: Public key for the key pair
+            being authorized.
+        :param JWK account_public_key:
```
kubek2k: I am not really a pythonist myself - but isn't this line excessive (no JWK arg on the list) ?

kuba:
BTW, if you follow the method I suggested above (catching `ImportError`), and mock `Answer` using `mock.Mock` rather than `dns.rrset` it should be possible to run tests in environments where `dnspython` is missing - which answers your question from https://github.com/letsencrypt/letsencrypt/pull/2061#issuecomment-169253648 and https://github.com/letsencrypt/letsencrypt/pull/2061#issuecomment-169264687

acme/acme/challenges.py
kuba: Fixed in #2132.

kuba:
Given DNS has landed in production (https://twitter.com/letsencrypt/status/689919523164721152) we probably want to have this merged ASAP. Will look into that after I come back from FOSDEM.

CyrilPeponnet:
Works like a charm for me. Can't wait to have it merged in master :)

moshevds:
I didn't find this PR before. I'm writing a plugin that needs this challenge, and started working on the ACME challenge myself in PR #2461. I'll test this code with my plugin and close my PR as it isn't needed.

acme/acme/challenges.py
```diff
@@ -216,6 +215,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+            return False
+
+        validation_domain_name = chall.validation_domain_name(domain)
+        validation = chall.validation(account_public_key)
+        logger.debug("Verifying %s at %s...", chall.typ, validation_domain_name)
+
+        try:
+            from acme import dns_resolver
+        except ImportError:  # pragma: no cover
+            raise errors.Error("Local validation for 'dns-01' challenges "
```
moshevds: I'm not sure this should raise an error.
A failed self-verification produces a warning only, so perhaps this should also be a warning, followed by return False?

acme/acme/challenges.py
```diff
@@ -216,6 +215,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+                               "requires 'dnspython'")
+        txt_records = dns_resolver.txt_records_for_name(validation_domain_name)
+        exists = validation in txt_records
+        if not exists:
+            logger.debug("Key authorization from response (%r) doesn't match "
+                         "any DNS response in %r", self.key_authorization,
```
moshevds: self.key_authorization should be validation here

acme/acme/challenges.py
```diff
@@ -216,6 +215,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+
+    def validation(self, account_key, **unused_kwargs):
+        """Generate validation.
+
+        :param JWK account_key:
+        :rtype: unicode
+
+        """
+        return jose.b64encode(hashlib.sha256(self.key_authorization(
+            account_key).encode("utf-8")).digest()).decode()
```
moshevds: Why return unicode here? DNS TXT records are not encoding aware so I think a string would be better.

acme/acme/challenges.py
wteiken: Not sure.   My idea was that if you do call self-verification you should have the extra library installed (i.e., code using the client should ensure the lib is installed).  Returning false means "I tried the verification and it failed".   That is not the case if the library is missing, in that case a precondition for the call is missing.

A "typical" client would have to abort trying to answer the challenge (assuming it will fail).  And the client would not be able to determine if it is a problem that is solvable by creating a new challenge and trying to verify it.

Thoughts?

acme/acme/dns_resolver.py
```diff
@@ -0,0 +1,25 @@
...
+    :param unicode name: Domain name being verified.
+
+    :returns: A list of txt records, if empty the name could not be resolved
+    :rtype: list of unicode
+
+    """
+    try:
+        dns_response = dns.resolver.query(name, 'TXT')
+    except dns.exception.DNSException as error:
+        logger.error("Unable to resolve %s: %s", name, str(error))
```
moshevds: If the subdomain does not exist NXDOMAIN is raised here, and that's not really an error, just the signal for "no result". Maybe add "except dns.resolver.NXDOMAIN: return []"?

acme/acme/challenges.py
wteiken: I see your argument, but this matches all the other responses in the module.  So I'd rather keep it this way (and maybe change it uniformly later, because I do agree that it would be helpful to see the actual expected string).

acme/acme/challenges.py
moshevds: I agree that "failed to do verification" is seperate from "verification failed". I'm not sure how to best indicate this...

moshevds:
There is a small merge conflict with 86d6d270 but this code works great. Thanks.

I reviewed the code and it looks good to me. I hope this is merged soon :-)

acme/acme/challenges.py
wteiken: Mostly for compatibility with the HTTP version.

acme/acme/dns_resolver.py
wteiken: Sounds reasonable. Done.

acme/acme/challenges.py
moshevds: The return value from this method becomes available in Authenticator plugins, and this way they will always need to encode it back to a string.

My thought that a string might be better mostly comes from the docstring of the abstract method:
>        Subclasses must implement this method, but they are likely to
>        return completely different data structures, depending on what's
>        necessary to complete the challenge. Interepretation of that
>        return value must be known to the caller.

I am tempted to say that even the validation_domain_name and other constants defined by the ACME specification could be returned here. (Thats what I did myself in 83fe5e87)

moshevds:
I uploaded my plugin that uses this pull request to https://github.com/moshevds/le_dnsupdate.

@kuba When do you think you will be able to look at this pullrequest?
Also, is there any interest in having a dnsupdate/RFC2136 plugin directly in the letsencrypt repository?

wteiken:
@kuba Any idea when you'll have a chance to take another look to get this merged?

pde:
@wteiken @kuba has been fairly busy with his day job; if we don't hear from him next week we'll see if someone else can land this...

mitchellrj:
One I noticed while playing around with #2236 was that a dependency need adding in `setup.py` for `dnspython`.

moshevds:
Hi @mitchellrj,

The dnspython dependency is actually already added (you can see it at d842f268e5414f418e69913f6ac4fadffb327fb1). It is only used for the simple_verify method, but there is no need to use that method in the client itself. #2505 is actually created to remove the one place the client currently uses it (the http manual plugin, your new dns manual plugin does not need to use simple_verify either).

siccegge:
Hi!
@pde Any news on this?

kuba:
I'm currently writing integration tests with Boulder to get this merged soon.

kuba:
FYI my WIP branch which builds off @wteiken's code: https://github.com/kuba/letsencrypt/tree/dns-01.

kuba:
Couple of PRs and bug fixing later, bd4a07a2b8f8305189050233c69a02884ff3d9c4 has all green (https://travis-ci.org/kuba/letsencrypt/builds/122090126) integration testing for dns-01. At least I know this PR is working... Will look at the code once again and send a new PR soon, stay tuned.

acme/acme/challenges.py
kuba: raising an error here looks good to me, we might want to have a different class of errors, like `MissingDependencyError`

acme/acme/challenges.py
kuba: Agreed with @wteiken, we could potentially fix this in follow-up PR, but not here.

NB there is no security issue here as the challenge response is verified at the beginning of the function, so `self.key_authorization` at this point of the execution must be the same as in the validation

zh99998:
how will it handle hooks to dns provider?
does it compatible to letsencrypt.sh or https://github.com/AnalogJ/lexicon ?

siccegge:
This is not at all about any client suport but only the python librarly. If you program your tooling against the acme python library you can integrate with your DNS in any way you like.

zh99998:
It's really bad, consider implement it in CLI ?

if need a "standard" dns update method, here is a "nsupdate" program. although most of DNS service providers don't support RFC2136, There will be nsupdate-compatible program to do it, like AnalogJ/lexicon . It's much better than let every dns-01 user develop them own client.

moshevds:
Hi @zh99998,

I have already implemented a RFC2136 plugin that works with this patch: https://github.com/moshevds/le_dnsupdate

When this DNS-01 support is merged I intend to create a pull request to get RFC2136 support into the standard client. So what you are suggesting is already underway. I have also seen efforts to create a "manual" plugin that works with DNS-01.

devth:
What happened with this PR? Is something blocking this from being merged?

HLFH:
cc @bmw

wteiken:
Not exactly sure.  Just waiting for somebody to merge...

cixelsyd:
Definitely want to know when this is released.

bmw:
Kuba almost single-handedly wrote the `acme` module so he was reviewing and building off this PR, however, he hasn't pushed any commits to his branch in over two months.

While most of the rest of the client team is focusing on `nginx` support. We'll see if we can get this reviewed for the next release.

Sorry for the delay everyone!

JakeDEvans:
can this even happen without DNSSEC?  Otherwise if you figure out what IP is requesting the TXT record you could spoof it.

wteiken:
You would have to compromise the DNS server used by the machine that does the verification so you control the required response, which is possible.

But you could make the same argument for any of the controls. If you can compromise the DNS then you can have the domain resolve to your own server, and you can have your URL based verification request answered.

pde:
@bmw we talked about whether adding a `dnspython` dependency would undermine our medium-term goal of having `certbot-auto` not require `gcc` and compilation steps.  It appears that `dnspython` is purely Pythonic, so that's not a concern for merging this PR.

dropje86:
It might be a good idea to adjust the `simple_verify()` method to query the authoritative nameserver(s) of a domain, instead of the system configured ones. If there's for some reason delay in publishing the newly added records (due to zone transfers for example) the authoritative nameserver will respond with a NXDOMAIN which will be cached.

dropje86:
@moshevds totally agree that it shouldn't block the PR. Was just mentioning the downside of the current verification method implementation. For now I've packaged a local version with this PR merged, as it is explicitly what I need.

cixelsyd:
Any updated ETA for the official DNS support this PR represents? As noted, many await with baited breath!


mobile

pde:
@cixelsyd yes it's scheduled for [Certbot 0.9.0](https://github.com/certbot/certbot/milestone/22).

acme/acme/challenges.py
```diff
@@ -207,6 +206,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
 
 
 @ChallengeResponse.register
+class DNS01Response(KeyAuthorizationChallengeResponse):
+    """ACME "dns-01" challenge response."""
```
bmw: nit: Maybe we should remove the quotes around `"dns-01" to be consistent with the other challenge types.

acme/acme/challenges.py
```diff
@@ -207,6 +206,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+        if not exists:
+            logger.debug("Key authorization from response (%r) doesn't match "
+                         "any DNS response in %r", self.key_authorization,
+                         txt_records)
+        return exists
+
+
+@Challenge.register  # pylint: disable=too-many-ancestors
+class DNS01(KeyAuthorizationChallenge):
+    """ACME "dns-01" challenge."""
```
bmw: nit: Maybe we should remove the quotes around `"dns-01" to be consistent with the other challenge types.

acme/acme/challenges.py
```diff
@@ -207,6 +206,74 @@ def response_and_validation(self, account_key, *args, **kwargs):
...
+class DNS01Response(KeyAuthorizationChallengeResponse):
+    """ACME "dns-01" challenge response."""
+    typ = "dns-01"
+
+    def simple_verify(self, chall, domain, account_public_key):
+        """Simple verify.
+
+        :param challenges.DNS01 chall: Corresponding challenge.
+        :param unicode domain: Domain name being verified.
+        :param account_public_key: Public key for the key pair
```
bmw: This should be `:param JWK account_public_key:`.

acme/setup.py
```diff
@@ -35,6 +35,15 @@
 else:
     install_requires.append('mock')
 
+if sys.version_info < (3, 0):
```
bmw: This kind of code makes me nervous. My understanding is that we'd have to start building two wheels of our packages, one for Python 2 and one for Python 3. It'd be nice to not have to do that if possible.

Looks like we're in luck though. Since version 1.12, `dnspython` is supports both Python 2 and Python 3. This version of `dnspython` seems to be available everywhere that we are packaged. With this in mind, I think we should change this to depend on `dnspython>=1.12`, allowing us to continue building a single wheel that should work everywhere. We should add a comment saying that version 1.12 is required to provide both Python 2 and Python 3 support.

bmw:
Other than a few tiny nits above, this LGTM. Looks there's a merge conflict that needs to be fixed as well, but once these changes are made, I'll merge. Sorry again for the delay!

wteiken:
Thanks for the review, I'll try to address the open items this weekend.

wteiken:
The latest update should address the comments and resolve the merge conflict.

bmw:
LGTM. Thanks for this PR @wteiken and sorry again it sat around so long.

hobarrera:
Is it possible that documentation is missing for this PR? I haven't been able to figure out how to use `certbot` with `dns-01`challenge, either from the docs, or looking at this PR.

brianjmurrell:
Hopefully I am not going to try too many people's patience with this question, but is this the patch that will give the {{certbot}} client the ability to request certificates using the ```dns-01``` challenge method automatically?  i.e. by populating DNS with the challenge?

acme/acme/challenges.py
```diff
@@ -231,8 +298,8 @@ def simple_verify(self, chall, domain, account_public_key, port=None):
             being authorized.
         :param int port: Port used in the validation.
 
-        :returns: ``True`` iff validation is successful, ``False``
-            otherwise.
+        :returns: ``True`` iff validation with the files currently served by the
```
ad-m: Typo ```iff```. It should be ```if```.

bmw: https://en.wikipedia.org/wiki/If_and_only_if

FelixSchwarz:
> Typo |iff|. It should be |if|.

https://en.wikipedia.org/wiki/If_and_only_if

acme/acme/challenges.py
ad-m: @bmw , i'm sorry. Thanks for the link!

bmw: No worries. Thanks for commenting about a potential typo.

hobarrera: In that case, the `otherwise` sounds somewhat redundant.

PR #2062, opened 2016-01-02
Better Nginx error handling.

jsha:
Raise MisconfigurationError on restart failure, so we don't attempt to continue
with an authorization we know will fail.

Log at debug level the config files that are about to be written out, so it's
easier to debug restart failures.

Fix https://github.com/letsencrypt/letsencrypt/issues/942:
Error out if adding a conflicting directive.

Remove unnecessarily-inserted access_log and error_log directives. These were
added to make integration testing easier but are no longer needed. Incidentally
this makes the plugin work with some configs where it wouldn't have worked
previously.

Change the semantics of add_server_directives with replace=True so only the
first instance of a given directive is replaced, not all of them. This works
fine with the one place in the code that calls add_server_directives with
replace=True, because all of the involved directives aren't allowed to be
duplicated in a given block.

Make add_http_directives do inserts into the structure itself, since its needs
were significantly different than the more general add_server_directives. This
also allows us to narrow the scope of the `block.insert(0, directive)` hack that
we inserted to work around https://trac.nginx.org/nginx/ticket/810, since it's
only necessary for http blocks.

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/parser.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/parser.py
@@ -518,8 +518,12 @@ def _add_directive(block, directive, replace):
         # Append directive. Fail if the name is not a repeatable directive name,
         # and there is already a copy of that directive with a different value
         # in the config file.
-        if location != -1 and directive[0].__str__() not in repeatable_directives:
-            if block[location][1] == directive[1]:
+        directive_name = directive[0]
+        directive_value = directive[1]
+        if location != -1 and directive_name.__str__() not in repeatable_directives:
+            if block[location][1] == directive_value:
+                # There's a conflict, but the existing value matches the one we
+                # want to insert, so it's fine.
                 pass
             else:
                 raise errors.MisconfigurationError(
--- a/letsencrypt-nginx/letsencrypt_nginx/parser.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/parser.py
@@ -213,6 +213,7 @@ class NginxParser(object):
             if ext:
                 filename = filename + os.path.extsep + ext
             try:
+                logger.debug('Dumping to %s:\n%s', filename, nginxparser.dumps(tree))
                 with open(filename, 'w') as _file:
                     nginxparser.dump(tree, _file)
             except IOError:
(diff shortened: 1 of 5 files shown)


letsencrypt-nginx/letsencrypt_nginx/parser.py
```diff
@@ -478,21 +491,39 @@ def _add_directives(block, directives, replace=False):
...
-            if len(directive) == 0:
-                continue
-            for index, line in enumerate(block):
-                if len(line) > 0 and line[0] == directive[0]:
-                    block[index] = directive
-                    changed = True
-            if not changed:
+        _add_directive(block, directive, replace)
+
+repeatable_directives = set(['server_name', 'listen', 'include'])
```
SwartzCr: should this maybe be inside of the _add_directive function? It's a little strange to me to see it hanging out between defs

letsencrypt-nginx/letsencrypt_nginx/parser.py
```diff
@@ -478,21 +491,39 @@ def _add_directives(block, directives, replace=False):
...
+            raise errors.MisconfigurationError(
+                'expected directive for %s in the Nginx '
+                'config but did not find it.' % directive[0])
+        block[location] = directive
+    else:
+        # Append directive. Fail if the name is not a repeatable directive name,
+        # and there is already a copy of that directive with a different value
+        # in the config file.
+        if location != -1 and directive[0].__str__() not in repeatable_directives:
+            if block[location][1] == directive[1]:
```
SwartzCr: I don't understand this line, wouldn't directive[1] be different than directive[0]?

letsencrypt-nginx/letsencrypt_nginx/parser.py
jsha: It only needs to be constructed once, so I put it outside the function. I'm not sure what the best Python idiom is here. Should I move it to the top of the file?

SwartzCr: that might be the right idiom? In the fun all caps mode?

SwartzCr:
Can't say much for code quality of tests as they still baffle me, but all the rest of it looks really good to me

SwartzCr:
updated variable wording looks good to me

SwartzCr:
+1 to merge

PR #1707, opened 2015-12-04
Fix Redirect Verification

sagi:
Closes #1620.

PR #2075, opened 2016-01-03
Copy only relevant lines from http vhost to ssl vhost skeleton [needs revision]

### LINKED ISSUE #1996: the apache plugin adds an unconditional redirect inside the :443 vhost
The apache plugin adds an unconditional redirect inside the :443 vhost

I had to comment out these lines or else the webaddress just kept looping

RewriteEngine on
RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [L,QSA,R=permanent]

sagi:
Closes #1996. See #1996 (comments) for reasoning.


pde:
What if there's a rewrite from `/path/1` to `script.php?command=path&param=1`?  Wouldn't failing to copy the rewrite cause breakage in that case?

pde:
Unless we can land a fixed-up version of this, it may be best to revert #1707 for 0.2.0 :(

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -697,6 +697,47 @@ def _get_ssl_vhost_path(self, non_ssl_vh_fp):
...
+        Copying the above line to the ssl vhost would cause a redirection loop.
+
+        :param str line: a line extracted from the http vhost config file.
+
+        :returns: True - don't copy line from http vhost to SSL vhost.
+        :rtype: (bool)
+        """
+
+        rewrite_rule = "RewriteRule"
+        if line.lstrip()[:len(rewrite_rule)] != rewrite_rule:
```
pde: `if not line.lstrip.startswith(rewrite_rule):` is nicer

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -697,6 +697,47 @@ def _get_ssl_vhost_path(self, non_ssl_vh_fp):
...
+        # line starts with RewriteRule.
+
+        # According to: http://httpd.apache.org/docs/2.4/rewrite/flags.html
+        # The syntax of a RewriteRule is:
+        # RewriteRule pattern target [Flag1,Flag2,Flag3]
+        # i.e. target is required, so it must exist.
+        target = line.split()[2].strip()
+
+        # target may be surrounded with double quotes
+        if len(target) > 0 and target[0] == target[-1] == "\"":
```
bmw: What other characters can be used here? At the very least, using single quotes seems to work.

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -697,6 +697,47 @@ def _get_ssl_vhost_path(self, non_ssl_vh_fp):
...
+            target = target[1:-1]
+
+        https_prefix = "https://"
+        if len(target) < len(https_prefix):
```
pde:
```
if target.startswith(https_prefix):
    return True
return False
```

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -697,6 +697,47 @@ def _get_ssl_vhost_path(self, non_ssl_vh_fp):
...
+            return False
+
+        if target[:len(https_prefix)] == https_prefix:
```
bmw: What if someone has a redirect to another HTTPS site? For example, what if someone wants to have both `http://example.com` and `https://example.com` redirect to `https://www.example.com`?

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -714,9 +755,27 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
             with open(avail_fp, "r") as orig_file:
                 with open(ssl_fp, "w") as new_file:
                     new_file.write("<IfModule mod_ssl.c>\n")
+                    sift = False
+
                     for line in orig_file:
-                        new_file.write(line)
+                        if self._sift_line(line):
+                            if not sift:
+                                new_file.write("# The following rewrite rules "
```
bmw: nit: From this wording, I'd expect all disabled rewrite rules to be found directly under this comment. This is only the case, however, if all disabled rewrite rules are adjacent in the file.

pde: "One or more of the following rewrite rules"?

bmw:
@sagi, just so you're aware, we'd like this to land in `0.2.0` so I'm currently working on a PR based on this to make the small revisions pde and I suggested.

PR #2078, opened 2016-01-04
Implement non-interactive mode for the client.

### LINKED ISSUE #804: Offer a well-defined non-interactive mode.
For various purposes (such as scripting or our own integration tests!) people will sometimes want to run the client in non-interactive mode.  In such situations, it is a mistake to block on user input of any sort.  We should offer a clear cli flag that causes the client to stop asking the user questions.  

In non-interactive mode it may be acceptable to error out if some instruction is missing, though if at all possible that should be deterministic and happen immediately when the client is run, not partway through execution.

pde:
Closes #804.


pde:
Let's get started on reviewing this.  It might be good for 0.2.1, if we can do it.

letsencrypt/display/util.py
```diff
@@ -21,6 +21,20 @@
 HELP = "help"
 """Display exit code when for when the user requests more help."""
 
+def _wrap_lines(msg):  # pylint: disable=no-self-use
```
bmw: pylint disable can be removed.

letsencrypt/display/util.py
```diff
@@ -61,10 +75,11 @@ def menu(self, message, choices,
 
         :param str ok_label: label of the OK button
         :param str help_label: label of the help button
+        :param dict _kwargs: absorbs default / cli_args
 
-        :returns: tuple of the form (`code`, `tag`) where
```
bmw: nit: Why were the backticks removed? They provide formatting to the Sphinx documentation. This removal was done in a number of places.

letsencrypt/display/util.py
```diff
@@ -403,6 +411,117 @@ def _get_valid_int_ans(self, max_):
 
         return OK, selection
 
+class NoninteractiveDisplay(object):
+    """File-based display."""
```
bmw: Incorrect docstring

letsencrypt/display/util.py
```diff
@@ -403,6 +411,117 @@ def _get_valid_int_ans(self, max_):
 
         return OK, selection
 
+class NoninteractiveDisplay(object):
```
bmw: Question: Should this inherit from FileDisplay?

letsencrypt/display/util.py
```diff
@@ -197,10 +215,12 @@ def menu(self, message, choices,
         :param choices: Menu lines, len must be > 0
         :type choices: list of tuples (tag, item) or
             list of descriptions (tags will be enumerated)
+        :param dict _kwargs: absorbs default / cli_args
+
+        :returns: tuple of (`code`, `index`) where
```
bmw: Incorrect returns documentation. Compare it with the returns from the other `menu` functions.

letsencrypt/cli.py
```diff
@@ -44,6 +44,11 @@
 
 logger = logging.getLogger(__name__)
 
+# For help strings, figure out how the user ran us.
+# When invoked from letsencrypt-auto, sys.argv[0] is something like:
+# /home/user/.local/share/letsencrypt/bin/letsencrypt"
+fragment = os.path.join(".local", "share", "letsencrypt")
+cli_command = "letsencrypt-auto" if fragment in sys.argv[0] else "letsencrypt"
```
bmw: I don't like this design for a few reasons:
1) While the path to `letsencrypt` when installed from `letsencrypt-auto` will be `~/.local/share/letsencrypt/bin/letsencrypt` in most cases, it certainly won't be in all cases meaning this won't reliably work.
2) `letsencrypt-auto` is supposed to be a temporary solution until we get real packaging which we are getting at an astonishing pace. With this in mind, why do we keep baking more le-auto stuff into letsencrypt?

letsencrypt/cli.py
```diff
@@ -315,7 +322,8 @@ def _handle_subset_cert_request(config, domains, cert):
              ", ".join(domains),
              br=os.linesep)
     if config.expand or config.renew_by_default or zope.component.getUtility(
-            interfaces.IDisplay).yesno(question, "Expand", "Cancel"):
+            interfaces.IDisplay).yesno(question, "Expand", "Cancel",
+                                       default=True):
```
bmw: I'm not sure if this is a fair default?

letsencrypt/cli.py
```diff
@@ -612,14 +647,17 @@ def install(args, config, plugins):
...
+        domains = _find_domains(args, installer)
+        le_client = _init_le_client(
+            args, config, authenticator=None, installer=installer)
+        assert args.cert_path is not None  # required=True in the subparser
+        le_client.deploy_certificate(
+            domains, args.key_path, args.cert_path, args.chain_path,
+            args.fullchain_path)
+        le_client.enhance_config(domains, config)
+    except errors.MissingCommandlineFlag, e:
+        return e.message
```
bmw: What is the point of this? If I remember how our exception handling works (no guarantees) we get this exact behavior without any changes to this function. [source](https://github.com/letsencrypt/letsencrypt/pull/2078/files#diff-54b2f468d57ef7436008bd031bc89711R1363)

letsencrypt/interfaces.py
```diff
@@ -377,14 +377,19 @@ def menu(message, choices,
...
             `code` - str display exit code
             `index` - int index of the user's selection
 
+        :raises errors.MissingCommandlineFlag: if called in non-interactive
+            mode without a default set
+
         """
 
-    def input(message):
+    def input(message, default=None, cli_args=None):
```
bmw: Needs `:raises errors.MissingCommandLineFlag:...`

letsencrypt/display/util.py
```diff
@@ -403,6 +411,117 @@ def _get_valid_int_ans(self, max_):
...
+        :param bool pause: The NoninteractiveDisplay waits for no keyboard
+
+        """
+        side_frame = "-" * 79
+        message = _wrap_lines(message)
+        self.outfile.write(
+            "{line}{frame}{line}{msg}{line}{frame}{line}".format(
+                line=os.linesep, frame=side_frame, msg=message))
+
+    def menu(self, message, choices, default=None, cli_flag=None, **kwargs):
```
bmw: I don't think we should use `**kwargs` here. 3rd party plugins could follow IDisplay perfectly, but if they didn't use keyword arguments, bad things can happen to good people.

letsencrypt/cli.py
pde: I think that minimising the number of lineages is almost always what the user wants, so I'm inclined to go with that as the default.  The alternatives are erroring or duplicating, I suppose, but I don't like the non-determinism that introduces.

letsencrypt/display/util.py
```diff
@@ -403,6 +411,117 @@ def _get_valid_int_ans(self, max_):
...
+        "Error out in case of an attempt to interact in noninteractive mode"
+        msg = "Missing command line flag or config entry for this setting:\n"
+        msg += message
+        if extra:
+            msg += "\n" + extra
+        if cli_flag:
+            msg += "\n\n(You can set this with the {0} flag)".format(cli_flag)
+        raise errors.MissingCommandlineFlag, msg
+
+    def notification(self, message, height=10, pause=False):
```
bmw: nit: No cover.

letsencrypt/display/util.py
```diff
@@ -403,6 +411,117 @@ def _get_valid_int_ans(self, max_):
...
+        :returns: tuple of (`code`, `tags`) where
+            `code` - str display exit code
+            `tags` - list of selected tags
+        :rtype: tuple
+
+        """
+        if default is None:
+            self._interaction_fail(message, cli_flag, "? ".join(tags))
+        else:
+            return OK, default
```
bmw: nit: no cover

letsencrypt/cli.py
pde: IIRC without that code the error was still raised, but in a different enough way that [this test](https://github.com/letsencrypt/letsencrypt/pull/2078/files#diff-1333bfbf9a96ab6aa2587c1ca0cbd2b3R135) failed.

letsencrypt/display/util.py
pde: What if we recommend implementation that way?  I bet people will tend to crib from our implementations...

letsencrypt/display/util.py
pde: I'm not sure that they were intentionally removed, I think there were multiple variants of this method's docstring in the code, some with and some without backticks.  I'll try to standardise on including them.

pde: (Ref: https://github.com/letsencrypt/letsencrypt/pull/2078/files#diff-1333bfbf9a96ab6aa2587c1ca0cbd2b3R135)

letsencrypt/display/util.py
```diff
@@ -178,15 +196,15 @@ def notification(self, message, height=10, pause=True):
...
         self.outfile.write(
             "{line}{frame}{line}{msg}{line}{frame}{line}".format(
                 line=os.linesep, frame=side_frame, msg=message))
         if pause:
             raw_input("Press Enter to Continue")
 
-    def menu(self, message, choices,
-             ok_label="", cancel_label="", help_label=""):
+    def menu(self, message, choices, ok_label="", cancel_label="",
+             help_label="", **_kwargs):
```
bmw: Nit: unused_kwargs is probably nicer.

letsencrypt/display/util.py
pde: I think this one is right and the other ones are wrong.

letsencrypt/display/util.py
pde: It could, and reuse `notification` that way.  Or it could just call that `notification` function.  Would save about 4 lines of code, not sure it's worth the compexity.

letsencrypt/cli.py
bmw: Talked about this out of band and I believe we agreed we should error out. Ping me if you reached a different conclusion.

letsencrypt/tests/cli_test.py
```diff
@@ -130,6 +131,28 @@ def test_help(self):
...
         self.assertTrue(cli.usage_strings(plugins)[0] in out)
 
+
+    def _cli_missing_flag(self, args, message):
+        "Ensure that a particular error raises a missing cli flag error containing message"
+        exc = None
+        try:
+            with mock.patch('letsencrypt.cli.sys.stderr'):
+                out = cli.main(self.standard_args + args[:])  # NOTE: parser can alter its args!
+            print out
```
bmw: Stray print statement

letsencrypt/cli.py
bmw: Just tried without the try/catch and it seemed to work for me. If not, I believe the behavior to the user should be the same and we should just fix the test.

letsencrypt/display/util.py
bmw: You're right. I wrote this down as a thing to look into when I was reviewing the code and then commented on the wrong function.

letsencrypt/display/util.py
bmw: Maybe? But it's a pretty easy change on our end and saves potential headaches. The arguments to be eaten by `kwargs` can just be `unused_*`.

bmw:
I think this is a pretty good approach. We need to be cognizant of this feature when adding prompts in the future, but in general I think we should merge this after a bit of cleanup.

letsencrypt/display/util.py
pde: I guess my view is that plugins **should** use some version of `**unused` when implementing these interfaces, so that they are extensible in a backwards-compatible way.  That's more true of third party plugins that it is of our own internal implementations, but since people will tend to read our implementations, I'm inclined to lead the way.

bmw: How often are we planning on changing these interfaces? My thought is that since we have the ability to not make this a breaking API change (simply by removing `kwargs` and adding two arguments to this single function...), we should take advantage of that.

Furthermore, I think it's just as likely (if not more so) that authors of 3rd party plugins will look at the interfaces we've defined in `interfaces.py` to determine how to interact with the base `letsencrypt` module. This function as it stands does not properly implement this interface.

letsencrypt/display/util.py
pde: Fixed.

letsencrypt/display/util.py
pde: Fixed

letsencrypt/cli.py
pde: I looked into doing this in a more exhaustive / thorough way, and it's probably impossible.  By default, `sudo` doesn't preserve environment variables (you need to pass in `-E` for that), so we can't tell if the user had `VENV_PATH` or `XDG_DATA_HOME` set.  But I think it's going to be very rare for those to be in play, so as long as we're cautious about what we hang on this hook, it should be okay.  I'll add a comment to that effect...

pde:
I think I've addressed everything in your review; LMK if you want the inheritance from `FileDisplay`; I'm not sure it's worth the effort to save a 3-4 lines of code.

pde:
Lint passes for me locally, travis *should* turn green now :)

bmw:
LGTM

Soullivaneuh:
@pde Looks good. Where can we find documentation about that? Thanks.

PR #2082, opened 2016-01-04
Fix boulder-start.sh

TheNavigat:


jsha:
I'd prefer to see this shell file start with `cd $(dirname $0)`, and leave the invocation the same. That way it can work regardless of what directory it's invoked from. Also please use `git rebase` to squash these commits into one.

jmhodges:
I actually disagree with both this change and the idea above! Isn't this fun.

cd'ing and dirname don't work well across platforms with their different handling of symlinks. I'd rather we error out if we're not in the same directory as boulder-fetch.sh. This is similar to `go/src/make.bash` and other commands in the Go project.

TheNavigat:
Because of the way this is run, @jmhodges's argument makes sense.

However... wouldn't it be too evil to force the users to be in the tests directory in order to run boulder tests?

jmhodges:
So, in boulder previously, we've assumed folks were running commands in the top-level directory. Which is why that code used that file path in the first place.

The reference to the go source code I was making was to things like https://github.com/golang/go/blob/4577cfcb0b861b70e5d503ffb14dba1ac2a66314/src/make.bash#L50

That one assumes folks are in the same dir but is just a prepend of "./test/" away from being like what I want.

TheNavigat:
Sounds good to me.

Should I code this, or should I wait for @jsha to add his thoughts?

jsha:
Can you tell me more about the cd and dirname issues across platforms? It's a a pattern I use a lot and would like to know more about why it's broken.

Also, to rewind a bit: In what context is boulder-start.sh broken? The docs currently say to run it as `./tests/boulder-start.sh`, which would break under your change, I think.

TheNavigat:
Oh, you're correct regarding that bit, didn't notice the docs.

I instinctively thought that boulder-start.sh should be able to run without errors if executed from the same directory. Maybe we should echo an error to make it more clear?

TheNavigat:
Ping @jsha @jmhodges

bmw:
While I'm not jsha or jmhodges, I'll chime in and say I don't see the need for this.

A common pattern in our documentation is running things from the root directory of the repo. This invalidates our documentation and breaks the script when run from the project root.

TheNavigat:
Makes sense. Closing.

Issue #829, opened 2015-09-26
Recover registration URI


kuba:
As per [ACME spec](https://letsencrypt.github.io/acme-spec/#registration):

"If the server already has a registration object with the provided account key, then it MUST return a 409 (Conflict) response and provide the URI of that registration in a Location header field. This allows a client that has an account key but not the corresponding registration URI to recover the registration URI."

This is not implemented by the `acme` module

kevgliss:
Has this been worked on? It's pretty key to some integration I am currently writing. If no one has taken a stab at this I will submit a PR.

kuba:
@kevgliss, I have some code to support this bug in my local branch. Maybe I could try to upload it... Though New Year's Eve is close, so it might not happen until first days of January :)

kevgliss:
@kuba sounds good, I took a look at look at the client yesterday but could not find a particularly clean way of doing this. Will look forward to seeing your changes and maybe I can iterate on them.

PR #2085, opened 2016-01-04
Add a "KeyAlreadyRegistered" exception for Client.register [needs revision]

wteiken:
Add a specific KeyAlreadyRegistered exception that contains the previous registration URL.

This includes adding the response object to the ClientError exception.  It also includes returning the mesages.Error parsed from a non-success response in a field of ClientError instead of as a separate exception.

### DIFF
--- a/acme/acme/errors.py
+++ b/acme/acme/errors.py
@@ -12,15 +12,29 @@ class SchemaValidationError(jose_errors.DeserializationError):
 
 class ClientError(Error):
     """Network error or unexpected client response."""
-    def __init__(self, response, error=None, *args, **kwargs):
+    def __init__(self, response, *args, **kwargs):
+        """
+        :param requests.Response details: The response object that caused the
+            exception to be thrown.
+        """
         super(ClientError, self).__init__(*args, **kwargs)
-        self.error = error
         self.response = response
 
+
+class ClientErrorWithDetails(ClientError):
+    """Network error."""
+    def __init__(self, response, details, *args, **kwargs):
+        """
+        :param requests.Response details: The response object that caused the
+            exception to be thrown.
+        :param acme.messages.Error details: The details of the error as transmitted
+            by the server.
+        """
+        super(ClientErrorWithDetails, self).__init__(response, *args, **kwargs)
+        self.details = details
+
     def __str__(self):
-        if self.error is None:
-            return super(ClientError, self).__str__()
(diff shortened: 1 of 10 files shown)


pde:
Integration tests failed due to a heisenbug we'd been wrestling with; restarting them.

wteiken:
There are also pieces missing to adopt to this in the rest of the client (although no integration test seems to fail on my test side).   I am not sure if this is the right way to go, somebody who has a better idea where the API is supposed to head should chime in...

My motivation is that while it is nice to have the parsed error object it loses a lot of the context and due to the nature of message.Error it can't easily be added there.  So it seemed natural to make the message.Error a field of the ClientError and uniformly return that (and probably also no longer make message.Error a subclass of Exception as the ClientError will be caught).  It may be nicer to throw a subclass of ClientError (and not have to check if error is set).  I am pretty much already planning to do that as it will simplify code.

There are a few references to message.Error in auth_handler that I am not sure yet how they are handled best...  so this pull is more an attempt to start the discussion, not a final solution.

There are a few sites in the client where the message.Error

kuba:
This PR likely fixes #829 but stands in conflict with my local changes... It would be better to sync before duplicating each other's work. Will attempt to review some time this week.

acme/acme/errors.py
```diff
@@ -11,26 +11,62 @@ class SchemaValidationError(jose_errors.DeserializationError):
 
 
 class ClientError(Error):
+    """Network error or unexpected client response."""
+    def __init__(self, response, *args, **kwargs):
+        """
+        :param requests.Response details: The response object that caused the
```
kuba: Move docstring to `ClientError.__doc__` (using `ivar`), rename `details` to `response`

pde:
@wteiken I believe the main places to look at where your exceptions could be caught are `letsencrypt/cli.py` (especially `_handle_exception` and `_determine_account`) and `letsencrypt/client.py` (especially `perform_registration`).

siccegge:
Hi! what's the state of this?

bmw:
@siccegge, it appears Kuba left some comments on the PR that were never addressed and there are merge conflicts that need to be fixed. If these changes are made, I will review.

jsha:
I have a narrower fix for this, introducing a ConflictError in #4254. Since this PR is relatively old and has a lot of conflicts, I'm closing it for now.

Issue #2054, opened 2016-01-01
OSError: [Errno 22] Invalid argument


janbuecker:
hi,

my renewal does not work. I'm running Ubuntu 14.04 with python 2.7.6 on my machine. My config file in `/etc/letsencrypt/cli.ini` looks as follows (I changed the domain names):

```
rsa-key-size = 4096
server = https://acme-staging.api.letsencrypt.org/directory
email = mail@mydomain.com
domains = mydomain.com
text = True
authenticator = webroot
webroot-path = /var/www/html
agree-tos = True
renew-by-default = True
```

If I now run (as root) the command `/opt/letsencrypt/letsencrypt-auto certonly --config /etc/letsencrypt/cli.ini` I get the following error:
```
Updating letsencrypt and virtual environment dependencies......
Requesting root privileges to run with virtualenv: /root/.local/share/letsencrypt/bin/letsencrypt certonly -a webroot --config /etc/letsencrypt/cli.ini
An unexpected error occurred:
OSError: [Errno 22] Invalid argument: '/etc/letsencrypt/live/mydomain.com/cert.pem'
Please see the logfiles in /var/log/letsencrypt for more details.
```

**/var/log/letsencrypt/letsencrypt.log**
```
2016-01-01 18:28:39,820:DEBUG:letsencrypt.cli:Root logging level set at 30
2016-01-01 18:28:39,820:INFO:letsencrypt.cli:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2016-01-01 18:28:39,820:DEBUG:letsencrypt.cli:letsencrypt version: 0.1.1
2016-01-01 18:28:39,820:DEBUG:letsencrypt.cli:Arguments: ['-a', 'webroot']
2016-01-01 18:28:39,820:DEBUG:letsencrypt.cli:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2016-01-01 18:28:39,821:DEBUG:letsencrypt.cli:Requested authenticator webroot and installer None
2016-01-01 18:28:39,823:DEBUG:letsencrypt.plugins.webroot:Creating root challenges validation dir at /usr/share/nginx/html/.well-known/acme-challenge
[...]

PR #2088, opened 2016-01-05
Don't error if the current_target is a hard file [needs revision]

kevinburke:
In at least some scenarios when you run

```
letsencrypt certonly --webroot -w <directory> -d <domain>
```

The client exits with the error "OSError: [Errno 22] Invalid argument:
'/etc/letsencrypt/live/<other-domain>/cert.pem'". Particularly confusing is the
fact that this error message is from letsencrypt 1) attempting to read a file
that exists and 2) attempting to read a domain that's not related to the domain
being validated.

The error comes from the letsencrypt client attempting to find duplicate
certs, and resolve symlinks by calling `os.readlink`. However, in some cases
letsencrypt attempts to call `os.readlink` on a hard file, which throws an
error in Python.

Since `current_target` is only attempting to retrieve the path to the given
certificate, it seems safe enough to check if a file is a symlink before
attempting to dereference it with `os.readlink`.

I verified that this patch fixes the issue on my own server - without the
patch, I get the error in #2054 and with the patch, I'm able to get a
certificate.

Fixes https://github.com/letsencrypt/letsencrypt/issues/2054.

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -1,5 +1,6 @@
 """Renewable certificates storage."""
 import datetime
+import errno
 import logging
 import os
 import re
@@ -273,10 +274,12 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
         try:
             target = os.readlink(link)
         except OSError as e:
-            logger.debug('Expected %s to be a symlink to a file in %s, was a'
-                         'hard file. Did you modify the letsencrypt directory?',
-                         link, self.cli_config.archive_dir)
-            raise
+            if not e.errno is errno.EINVAL:
+                raise
+            raise errors.CertStorageError(
+                'Expected %s to be a symlink to a file in %s, was a '
+                'hard file. Did you modify the letsencrypt directory?' %
+                (link, self.cli_config.archive_dir))
 
         if not os.path.isabs(target):
             target = os.path.join(os.path.dirname(link), target)
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -270,7 +270,10 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
             logger.debug("Expected symlink %s for %s does not exist.",
                          link, kind)
             return None
-        target = os.readlink(link)
+        if os.path.islink(link):
+            target = os.readlink(link)
+        else:
+            target = link
         if not os.path.isabs(target):
             target = os.path.join(os.path.dirname(link), target)
         return os.path.abspath(target)
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -274,7 +274,7 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
         try:
             target = os.readlink(link)
         except OSError as e:
-            if not e.errno is errno.EINVAL:
+            if not e.errno == errno.EINVAL:
                 raise
             raise errors.CertStorageError(
                 'Expected %s to be a symlink to a file in %s, was a '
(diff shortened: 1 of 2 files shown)


letsencrypt/storage.py
```diff
@@ -270,7 +270,10 @@ def current_target(self, kind):
             logger.debug("Expected symlink %s for %s does not exist.",
                          link, kind)
             return None
-        target = os.readlink(link)
+        if os.path.islink(link):
+            target = os.readlink(link)
```
kevinburke: There's a slight race here, in theory the file could be deleted in between the check and the `readlink` call.

I could change this to do

```python
try:
    target = os.readlink(link)
except OSError:
    # Could add a check for more specific error here
    target = link
```

pde:
My first instinct is, "this shouldn't be necessary; all of the files in `/etc/letsencrypt/live/*/` should be symlinks (to real certs/chains/keys in `/etc/letsencrypt/archive`)".  Are there some systems where that has turned out not to be the case?  Can we figure out how it happened?

kevinburke:
>  Can we figure out how it happened?

Ah, I can tell you exactly how; I'm moving a bunch of my websites over to an Ansible configuration, and I plan on frequently wiping/rebooting the server to ensure the box comes up ok. I didn't want to hit LE's rate limits, so I copied the certs locally after generation (scp remote:/etc/letsencrypt/live local/certs) and then scp'd them back to the box when I booted a new one...

kevinburke:
@pde What if I changed this to just make the error message clearer? Something like "we expected %s to be a symlink to the %s directory but it was a hard file".

pde:
@kevinburke sorry I missed that question; yes, changing the error message sounds great.

kevinburke:
@pde I pushed 7cb059b which adds a debug log line here.

I tried to modify the `OSError` but it's difficult to change the error message that gets returned, since it's generated from the `errno`, the filename and `os.strerror`. Debug logs also seem to be in line with the rest of the file

kevinburke:
If this looks OK, I can rebase it down into one commit. Also happy to work on the wording a little bit, the easiest thing would be to kill the second sentence

pde:
I think you should be able to modify the OS error, and that's probably going to lead to a better user experience.  How about something like:

```
if "[Errno 22]" in str(e):
   raise errors.CertStorageError, "Expected %s to be a symlink to a file in %s...".format(etc)
```
If the underlying error actually contained useful information, you could append it to the string for the new error you're synthesising (but in my testing it doesn't seem to contain useful information)

I worry that `logger.debug` buries things where they're hard to find; it's appropriate for diagnostics that may not be fatal or which may confuse some people, but in this case you have a good error message that should be accessible to most if not all users.

kevinburke:
> I worry that logger.debug buries things where they're hard to find; it's appropriate for diagnostics that may not be fatal or which may confuse some people, but in this case you have a good error message that should be accessible to most if not all users.

I'm on board with this as well! I wasn't sure if it was feasible and didn't think about reraising a LE error type.

I've pushed a0eb8c2 and a test as well; if it looks good I can rebase down to a single, better commit!

letsencrypt/storage.py
```diff
@@ -270,7 +271,16 @@ def current_target(self, kind):
             logger.debug("Expected symlink %s for %s does not exist.",
                          link, kind)
             return None
-        target = os.readlink(link)
+        try:
+            target = os.readlink(link)
+        except OSError as e:
+            if not e.errno is errno.EINVAL:
```
kevinburke: Crap, Ive been doing too much ruby lately. This is not right

bmw:
Thanks for your work on this @kevinburke. Sorry it took us so long to look at it.

Since you wrote this, the code in `storage.py` has changed a bit and we currently are checking our symlinks when the object is initially created to prevent problems later in the program. I wrote up #2454 which should resolve the issue.

Sorry again we didn't get this reviewed before the file changed!

bmw:
fixed in #2454

PR #2089, opened 2016-01-05
Update new_authzr_uri when querying or updating a registration

wteiken:
Sets new_authzr_uri  in _reg_from_response if the link is available in the response.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -66,15 +66,13 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
     @classmethod
     def _regr_from_response(cls, response, uri=None, new_authzr_uri=None,
                             terms_of_service=None):
-        terms_of_service = (
-            response.links['terms-of-service']['url']
-            if 'terms-of-service' in response.links else terms_of_service)
+        if 'terms-of-service' in response.links:
+            terms_of_service = response.links['terms-of-service']['url']
+        if 'next' in response.links:
+            new_authzr_uri = response.links['next']['url']
 
         if new_authzr_uri is None:
-            try:
-                new_authzr_uri = response.links['next']['url']
-            except KeyError:
-                raise errors.ClientError('"next" link missing')
+            raise errors.ClientError(response, 'missing "new_authrz_uri"')
 
         return messages.RegistrationResource(
             body=messages.Registration.from_json(response.json()),
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -72,7 +72,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             new_authzr_uri = response.links['next']['url']
 
         if new_authzr_uri is None:
-            raise errors.ClientError(response, 'missing "new_authrz_uri"')
+            raise errors.ClientError('"next" link missing')
 
         return messages.RegistrationResource(
             body=messages.Registration.from_json(response.json()),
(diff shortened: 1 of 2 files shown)


acme/acme/client.py
```diff
@@ -66,15 +66,13 @@ def __init__(self, directory, key, alg=jose.RS256, verify_ssl=True,
...
+            terms_of_service = response.links['terms-of-service']['url']
+        if 'next' in response.links:
+            new_authzr_uri = response.links['next']['url']
 
         if new_authzr_uri is None:
-            try:
-                new_authzr_uri = response.links['next']['url']
-            except KeyError:
-                raise errors.ClientError('"next" link missing')
+            raise errors.ClientError(response, 'missing "new_authrz_uri"')
```
kuba: - please remove `response` as first arg - we can discuss it in another PR (I've seen #2085 does a lot of similar stuff)
- can we use the same message as previously? I find reference to `next` link a it more readable than internal `new_authtzr_uri` name

acme/acme/client_test.py
```diff
@@ -127,6 +127,13 @@ def test_query_registration(self):
         self.response.json.return_value = self.regr.body.to_json()
         self.assertEqual(self.regr, self.client.query_registration(self.regr))
 
+    def test_query_registration_updates_new_authzr_uri(self):
```
kuba: I think this test would hold true without changes from this PR. What you want to test is the `_regr_from_response(response, new_authzr_uri='a').new_authzr_uri == 'UPDATED'` case. Or maybe you want to keep the test and revert changes to the `_regr_from_response`?

kuba:
I'm slightly confused by what this PR is trying to achieve... Maybe provide more context?

acme/acme/client.py
wteiken: What is the issue with the response argument?  I figured having a bit more context in the exception is helpful (e.g., to have the response headers).  That is why I added the explicit field in 2085.

I am happy with changing the message, but from my point of view an API user is actually more familiar with new_authrz_uri (as it is the field name in the object) than with the next/url notation (as the API hides most of the http call details).  So just trying to figure out the underlying principe for messaging for the future.

acme/acme/client_test.py
wteiken: Query registration passes in the previous value and the existing code does not update the link (i.e. it ignores a new value sent by the server and uses the current one if it is not None).  Unless of course the current value in the regr object is None (which would then be the right thing to change for this test).

acme/acme/client.py
kuba: I'm not saying `response` argument is wrong in general. I'm saying it's not in the scope for this PR :) Let's discuss it elsewhere and fix all occurrences at once instead of deviating in one case.

Other error messages in this file (and entire library) reference primitives from the spec rather than API itself. In fact there is another instance of `"next" link is missing` in this file. Again, let's keep things consitent.

acme/acme/client_test.py
kuba: Ah, I was wrong, indeed - that test wouldn't pass in the previous version of the code because `_send_recv_regr` takes data from supplied `regr`.

Resolved.

kuba:
Okay, I now see what's going on. Code was confusing because of special handling of `next` link (c.f. inline comment about https://github.com/letsencrypt/acme-spec/issues/94). LGTM, once `raise errors.ClientError` is fixed (https://github.com/letsencrypt/letsencrypt/pull/2089#discussion_r48820003).

acme/acme/client.py
wteiken: Done.

wteiken:
Done

kuba:
LGTM, thanks

bmw:
LGTM. Thanks for the PR @wteiken.

PR #2092, opened 2016-01-05
Select broader wildcard matching pattern for augeas paths if overlapping

### LINKED ISSUE #1043: apache parser searching sites-available, should be sites-enabled
for obvious reasons..

joohoi:
If multiple augeas paths are added for same FS directory under /augeas/load/Httpd/incl out of which one is a wildcard path, eg:

```
/etc/apache2/sites-enabled/*.conf
/etc/apache2/sites-enabled/*
```

augeas doesn't parse the files under the directory at all. This PR works around the problem by checking if a path with (full) wildcard already exists for the directory in question, and skips adding new, more specific ones if so. Also if a more specific ones are already found, and wildcard is being added, remove the existing more specific matching paths with only the wildcard path.

Fixes #1043

Examples before the modifications:

Does not work:
```
/etc/apache2/sites-enabled/*.conf
/etc/apache2/sites-enabled/*
```
Works:
```
/etc/apache2/sites-enabled/*.conf
/etc/apache2/sites-enabled/*.other
```

Ping @domcleal , I don't know if this is intended behaviour.

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/parser.py
+++ b/letsencrypt-apache/letsencrypt_apache/parser.py
@@ -500,18 +500,19 @@ class ApacheParser(object):
         """
 
         try:
-            use_new = False
-            remove_old = False
             new_file_match = os.path.basename(filepath)
             existing_match = self.parser_paths[os.path.dirname(filepath)]
-            if existing_match == new_file_match:
-                # True here to let augeas verify that the path is parsed
+            if existing_match == "*":
+                use_new = False
+            else:
                 use_new = True
-            elif new_file_match == "*":
-                use_new = True
+            if new_file_match == "*":
                 remove_old = True
+            else:
+                remove_old = False
         except KeyError:
             use_new = True
+            remove_old = False
         return use_new, remove_old
 
     def _remove_httpd_transform(self, filepath):
--- a/letsencrypt-apache/letsencrypt_apache/parser.py
+++ b/letsencrypt-apache/letsencrypt_apache/parser.py
@@ -35,6 +35,7 @@ class ApacheParser(object):
         # https://httpd.apache.org/docs/2.4/mod/core.html#define
         # https://httpd.apache.org/docs/2.4/mod/core.html#ifdefine
         # This only handles invocation parameters and Define directives!
+        self.parser_paths = {}
         self.variables = {}
         if version >= (2, 4):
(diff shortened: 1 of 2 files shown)


domcleal:
> Ping @domcleal , I don't know if this is intended behaviour.

No, I don't think it is, it looks very much like an Augeas library bug to me.  Seems that any overlapping globs will cause the issue, not just `*` and `*.conf`, so avoid adding anything that can cause the same file to be added/matched multiple times.

pde:
I'm seeing this fail to solve 1043 with this conf file: https://paste.debian.net/hidden/b8fa839c/

pde:
https://paste.debian.net/hidden/af235e19/

joohoi:
You should be able to get them parsed by running with parameter --apache-vhost-root /etc/apache2/sites-enabled

If the vhost isn't found in sites-available (the default) the vhost root can be overridden with that parameter. However the if the apache is configured correctly, the vhosts should be found in the default sites-available - path also. If in the other hand it's not, you can use the above parameter to override, but in that case you most likely want to have --apache-handle-sites False as well, as the new HTTPS vhost will be written (following the users broken configuration scheme) in sites-enabled and hence can't be "enabled" by the a2ensite.

pde:
So `--apache-vhost-root /etc/apache2/sites-enabled` does indeed work, but I guess what I'm wondering is whether there's a sane way to include *both* of those directories in the set of vhosts that's searched?

If we just include one, would it make more sense for `/sites-enabled/` to be the Debian default?

joohoi:
There is a way, by making vhost_root - constant variable a list, but that would disable any easy way for user to define their own vhost root directory, additionally we'd have to introduce a new constant - vhost_write_directory, to define the directory to write the vhosts separately.

While sites-enabled is indeed the debian default *parsed* vhost directory, the actual configuration files should exist in sites-available, to conform with debian / ubuntu apache configuration standards. So in my view any vhosts existing directly, or symlinked from custom locations (ie. non-sites-available) to sites-enabled is a configuration anomaly in debian/ubuntu, but can still be worked with by using the command line parameter.

pde:
@domcleal, @ArchimedesPi would either of you like to +1 this?  I don't feel I know augeas well enough to do so efficiently.

pde:
LGTM. Merging for canary testing.

Issue #1391, opened 2015-11-06
Install command should find existing certs the same way that run/certonly do


pde:
If the user ran "letsencrypt certonly" and then runs "letsencrypt install", it's very counter-intuitive that they need to find their cert in `/etc/letsencrypt/live`, know that they should use `fullchain.pem` rather than `cert.pem`, and then find the matching key in `/etc/letsencrypt/keys`.  We use all of the storage.py mechanisms by default.

pde:
The ways now include (1) matching existing certs via domain names, or (2) accepting a lineage specified with `--cert-name`.

PR #2094, opened 2016-01-05
Warn the user when --chain-path or --fullchain-path are missing but are required [minor revision requested]

juniorz:
Both options are not required as per `letsencrypt --help install`. The nginx plugin does not fail when they are required but missing, and generates a broken nginx configuration file instead.

See https://github.com/twstrike/letsencrypt/issues/3 and https://github.com/twstrike/letsencrypt/issues/5 for the original discussion.

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/configurator.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/configurator.py
@@ -122,7 +122,7 @@ class NginxConfigurator(common.Plugin):
 
     # Entry point in main.py for installing cert
     def deploy_cert(self, domain, cert_path, key_path,
-                    chain_path, fullchain_path):
+                    chain_path=None, fullchain_path=None):
         # pylint: disable=unused-argument
         """Deploys certificate to specified virtual host.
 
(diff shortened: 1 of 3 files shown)


letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -136,7 +136,14 @@ def deploy_cert(self, domain, cert_path, key_path,
 
         .. note:: This doesn't save the config files!
 
+        :raises errors.PluginError: When unable to deploy certificate due to
+            a lack of directives or configuration
+
         """
+        if not fullchain_path:
+            raise errors.PluginError(
+                "--fullchain-path is required for nginx plugin.")
```
pde: "The nginx plugin currently requires --fullchain-path to install a cert".  That requirement will go away when we resolve #1391; in most cases `letsencrypt install` should be able to find previously obtained certs.

juniorz: Sure.

juniorz:
Updated the error message.

pde:
LGTM

Issue #925, opened 2015-10-08
server blocks added to nginx.conf


jsha:
Steps to reproduce:

(on Ubuntu)
```
letsencrypt --server SERVER --installer nginx --authenticator nginx --text --agree-eula --agree-tos --no-verify-ssl -d 4.le-test.hoffman-andrews.com run
```

Expected result:

A new file named 4.le-test.hoffman-andrews.com is created under /etc/nginx/sites-available and symlinked into /etc/nginx/sites-enabled/.

Actual result:

A server block is added to /etc/nginx/nginx.conf:

```
    server {
        server_name 4.le-test.hoffman-andrews.com;
        listen 443 ssl;
        access_log /var/lib/letsencrypt/access.log;
        error_log /var/lib/letsencrypt/error.log;
        ssl_certificate /etc/letsencrypt/live/4.le-test.hoffman-andrews.com/cert.pem;
        ssl_certificate_key /etc/letsencrypt/live/4.le-test.hoffman-andrews.com/privkey.pem;
        include /etc/letsencrypt/options-ssl-nginx.conf;
    }
```

I know this is probably a hard problem to solve nicely, because they layout of where each virtualhost goes is a distribution-dependent thing. If we don't want to have special logic for the distribution's layouts of virtualhosts, maybe we could just default to putting them in a separate file that gets included in, e.g /etc/nginx/letsencrypt-virtualhosts.conf, or /etc/letsencrypt/nginx-virtualhosts.conf.

tdruiva:
Based on the conversation in this PR #2097 , I'll going to create a PR  to remove all vhost creation.
@joohoi @pde Thoughts?

jsha:
Makes sense to me.

pde:
Over in Apache land, we've gradually developed a [collection of machinery](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/constants.py) for knowing how each host OS will want to store its vhosts/server blocks. That's probably the right way to do things, if we are going to keep making new server blocks in the nginx plugin (in at least some cases).

bmw:
Fixed in #3528.

Issue #718, opened 2015-08-27
letsencrypt client and dealing with non-standard Nginx configurations


centminmod:
With the letsencrypt client I see you can pass specific Nginx `--nginx-server-root` and `--nginx-ctl1 parameters but I assume the tool has also made specific assumptions about where nginx vhost files will be located etc and the format they take ? For non-standard Nginx configurations what would I need to do ? Can I tell the letsscrypt client to look for my specific custom Nginx paths and locations ? or would I need to use standalone method and then script it to place the obtained SSL certs and files in my custom locations ?

For instance on my Centmin Mod LEMP stack and example `newdomain2.com`, my nginx SSL configuration is located at `/usr/local/nginx/conf/conf.d/newdomain2.com.ssl.conf` with install for Nginx at `/usr/local/nginx` and SSL certificates end up in directory at `/usr/local/nginx/conf/ssl/newdomain2.com`. Example of [how newdomain2.com Nginx vhost was generated](http://centminmod.com/http-public-key-pinning.html).

Basically, I am looking to integrate letsencrypt into my Nginx vhost creation routine :)

thanks

George

help

```
letsencrypt --help nginx
usage: 
  letsencrypt [SUBCOMMAND] [options] [domains]

The Let's Encrypt agent can obtain and install HTTPS/TLS/SSL certificates.  By
default, it will attempt to use a webserver both for obtaining and installing
the cert.  

positional arguments:
  SUBCOMMAND
    run                 Obtain a certificate and install.
    auth                Authenticate & obtain cert, but do not install it.
    install             Install a previously obtained cert in a server.
    revoke              Revoke a previously obtained certificate.
    rollback            Rollback server configuration changes made during
                        install.
    config_changes      Show changes made to server config during installation
    plugins             List server software plugins.

optional arguments:
  -h, --help            show this help message and exit
  -c CONFIG_FILE, --config CONFIG_FILE
                        config file path (default: None)

nginx:
  Nginx Web Server

  --nginx-server-root NGINX_SERVER_ROOT
                        Nginx server root directory. (default: /etc/nginx)
  --nginx-ctl NGINX_CTL
                        Path to the 'nginx' binary, used for 'configtest' and
                        retrieving nginx version number. (default: nginx)
```

centminmod:
not working or incorrect syntax ? on centos 6.7 with python 2.7

[...]

PR #2097, opened 2016-01-05
Auto creates a vhost configuration file if possible [needs revision]

juniorz:
When the target vhost is not configured in a separate file, the config is written to the root config file (`nginx.conf`) which usually contains only general nginx configuration.

The preferable approach is storing virtual host configuration in a separate directory whose location is distribution-dependent.

This change replaces this behavior by auto-detecting the vhost configuration directory based on the existing vhost configuration files. This way the vhost configuration files is created in a separate directory rather than the root config file.

Related issues: #718 and #925.


pde:
This is a rather different approach to the one that joohoi took in [making the Apache pluign aware of various config directories](https://github.com/letsencrypt/letsencrypt/pull/1811).  But that might be okay? Or we could have the same set of OS default directories that the Apache plugin has, but use this algorithm as a fallback?  @joohoi, what do you think?

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -217,6 +218,38 @@ def choose_vhost(self, target_name):
...
+                continue
+
+            path = os.path.dirname(vhost.filep)
+            paths_freq.setdefault(path, 0)
+            paths_freq[path] += 1
+
+        if len(paths_freq) == 0:
+            return None
+
+        conf_dir = max(paths_freq, key=(lambda k: paths_freq[k]))
```
pde: Perhaps at this point, if there is more than one path in use, and you're picking the most frequently used, add a `logger.info` to tell the user that is happening?  Otherwise this behaviour could be totally mysterious?

pde: Or even an `iDisplay.notification` call, this is fairly important information.

juniorz:
Comparing this PR with the behavior introduced by #1811:

The apache plugin does not create a vhost configuration file unless it can find an existing configuration that matches the target domain. Is this an intentional design decision?

The current nginx plugin implementation adds the configuration regardless of having a matching configuration for the target domain. For this reason this PR will create a new vhost configuration file in this case. It also assumes the newly created file will be imported because it's in the same directory as other available vhosts conf files (which were recursively loaded by the parser).

joohoi:
As a precaution, I'm not familiar enough with the nginx configuration variation between different distributions, so the below comments are based on the information I have.

In apache land, the OS based conventions are historically very opinionated, and differ greatly (have a lot of added functionality) compared to the vanilla apache. Not only the vhost configuration directory layout, but the whole control script / module handling shebang. This requires us to have broader default configuration arrays for apache. Nginx doesn't have the historical burden, and has a lot clearer execution flow, without the additional distribution specific control scripts as the default way of handling things. 

For nginx, this looks like a robust way to find out the best location for the vhost files, if I'm not mistaken of the common configuration structures. 

It should be noted, that this will ignore the possible sites-available / sites-enabled - paradigm completely, and just write the new files to the sites-enabled (or whatever directory is parsed by the main configuration) instead. But at the same time, it allows the usage in different environments without having to guess defaults by LE client itself.

That being said, and if my assumptions about the nginx configuration variances are correct, this looks good to me.

letsencrypt-nginx/letsencrypt_nginx/configurator.py
juniorz: That's fair. It would not interrupt the flow for fresh/standard nginx installations while still giving users with custom configuration an opportunity to fix wrong guesses.

pde:
OK, so this looks like a good strategy.  Add the `logger.info` and/or `iDisplay.notification` calls to notify the user and we can merge.

joohoi:
After the brief discussion in IRC, I'd suggest not merging before we have a clear consensus of the responsibilities of plugins. Creating new vhosts jumps into realm of the Nginx itself and the OS based packaging. I don't think we can make good enough guesses, and asking the user to provide documentroot, log file locations etc. is clearly out of scope of LE itself.

If we decide not to create new vhosts out of the blue, which I strongly advice for, this code will be dead, as we'll be able to just create the vhost config files to the location they were read from.

juniorz:
> If we decide not to create new vhosts out of the blue, which I strongly advice for, this code will be dead, as we'll be able to just create the vhost config files to the location they were read from.

Not only this but also the plugins needs to be changed to do not create a new vhost in `nginx.conf`: https://github.com/letsencrypt/letsencrypt/blob/8f984bd24f2779490bf526d47f7fbe14633fcfb1/letsencrypt-nginx/letsencrypt_nginx/configurator.py#L200

joohoi:
@juniorz yes, you are correct. That's the culprit.

pde:
On the call today I think we concluded that *in theory* it could be okay to make new vhosts, but only if the user asks for it with a flag and provides the minimum necessary metadata.

pde:
(and that's by no means an urgent thing to support...)

joohoi:
Ok, that sounds good. I still fear that configuring (any) httpd from ground up will open up a can of worms that is the configuration variance between operating systems and user expectations in general. But that will be the discussion of the time when/if we decide to implement such behaviour.

So for now, I suggest not merging this PR

pde:
@joohoi although if my reading of the code is correct, the plugin is already creating vhosts under some circumstances; this PR just moves them.  So we should either land it, or remove the vhost creation case altogether?

joohoi:
@pde I suggest removing it completely until we can comply with our own standards regarding it (the outlines you discussed on the call). As I see it, having it in half assed would be a blocker when it comes to shipping.

pde:
@juniorz does that make sense to you? If so, we should close this PR and open a separate ticket to remove all vhost creation cases for now?

(And I gather, implicitly, that nginx configs are such that we should always always be able to modify an existing vhost to be able to serve TLS, rather than making new copies of existing ones as occurs with Apache).

pde:
@joohoi: having talked to @olabini and read the nginx code a bit more closely, I do think there are circumstances where vhost creation can make sense.  Those are essentially when vhosts for a new (to nginx) domain will be created on a transient basis and then reverted/deleted after a cert has been obtained.  That can make sense if the nginx plugin is used with `certonly`, or if it is used with a different plugin (like a future mail server plugin) as the installer.  Such vhosts can be sensibly created during `perform`, so long as they are deleted during `cleanup`.

I do think that supporting this case is lower than Nginx MVP priority, and there's an argument for possibly even requiring an extra flag for it `--nginx-novel-domains` or something, since at least some of the time this case will arise because the admin is confused, not because they want to obtain a cert for a domain they don't intend to have nginx answer for.

pde:
Superceded by #3528

PR #2105, opened 2016-01-06
Added some standard apache2 modules that make tests fail.

phispi:
letsencrypt failed on our server when parsing our apache2.4 config (see below). It turned out that the files where the error happens are standard apache2 modules. Therefore we added the module configuration that triggers the failure. Now many tests fail with "AttributeError: 'NoneType' object has no attribute 'lower'" in letsencrypt_apache/parser.py", line 311.
Thanks for providing this great project!

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/tests/testdata/debian_apache_2_4/two_vhost_80/apache2/mods-available/authz_svn.load
+++ b/letsencrypt-apache/letsencrypt_apache/tests/testdata/debian_apache_2_4/two_vhost_80/apache2/mods-available/authz_svn.load
@@ -1,0 +1,5 @@
+# Depends: dav_svn
+<IfModule !mod_dav_svn.c>
+    Include mods-enabled/dav_svn.load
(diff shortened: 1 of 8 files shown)


phispi:
Reported as issue #2247

pde:
Sorry this PR was a bit confusing, it tends to get filtered out of PR lists because of the red test X but actually that was its intention :)

pde:
We can merge this into whatever fix we make for #2247.

phispi:
Sorry for "pushing hard" by creating an additional bug report. However, we (the friend of mine and I who work together on this) realized that the PR might be confusing: (a) who wants to merge a foreign code that makes tests fail? ;-) and (b) other users searching in the bug list for this issue might not find it as PR and (c) you might not like our PR and reject it but still might want to have the issue reported ;-)
Thanks!

bmw:
LGTM. Merging.

Thanks a lot @phispi for both reporting and providing tests for this issue.

PR #2109, opened 2016-01-07
Makes NginxParser aware of `ssl on` directive

### LINKED ISSUE #2059: Doesn't understand `ssl on` directive
In Nginx there are two ways to turn on SSL for a server:

```
server {
  listen 443 ssl;
}
```

```
server {
  listen 443;
  ssl on;
}
```

Our Nginx plugin doesn't understand the latter. In parser.py's `get_vhosts`, the `VirtualHost` object will be created with `ssl = False`, leading later code to attempt to "enable" SSL on that vhost, creating a fatal conflict for the ssl_certificate directive. One possible easy fix: Look for `ssl_certificate` to determine whether a vhost has ssl enabled, instead of looking at the `listen` or `ssl` directives.

juniorz:
Fixes #2059

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/parser.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/parser.py
@@ -113,7 +113,7 @@ class NginxParser(object):
         for filename in servers:
             for server in servers[filename]:
                 # Parse the server block into a VirtualHost object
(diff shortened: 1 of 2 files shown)


pde:
LGTM, thanks!

Issue #2012, opened 2015-12-24
Old packages in Ubunutu 12.04 cause apache conf breakage


pde:
Despite landing #1961, some people have continued to report parser failures with it...

ehsaan:
Take a look. Let me know if you want me to try out a change/...: https://goo.gl/bIE9T4

rroseselavy42:
can also confirm

cimm:
Same problem here. Tried Let's Encrypt a few weeks back and ran into issue #1961. Tried it again today with the latest master (ffd0d5a) but still failing with the `PluginError('Unable to parse runtime variables',)` error. Running Apache/2.2.22 on Debian 7.

Not running `mod_define` if that's important. Here are the loaded modules:

```
 core_module (static)
 log_config_module (static)
 logio_module (static)
 version_module (static)
 mpm_prefork_module (static)
 http_module (static)
 so_module (static)
 alias_module (shared)
 auth_basic_module (shared)
 authn_file_module (shared)
 authz_default_module (shared)
 authz_groupfile_module (shared)
 authz_host_module (shared)
 authz_user_module (shared)
 autoindex_module (shared)
 cgi_module (shared)
 deflate_module (shared)
 dir_module (shared)
 env_module (shared)
 headers_module (shared)
 mime_module (shared)
 negotiation_module (shared)
 php5_module (shared)
 reqtimeout_module (shared)
 rewrite_module (shared)
 setenvif_module (shared)
 ssl_module (shared)
 status_module (shared)
```

grantnhoward:
Also confirmed, still unresolved with latest master.
Running 2.2.22 on Ubuntu 12.04
ran ```./letsencrypt-auto --apache ```
received:

```
The apache plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('Unable to parse runtime variables',)
```

letsencrypt.log
```
2015-12-27 22:06:57,200:DEBUG:letsencrypt.cli:letsencrypt version: 0.1.1
2015-12-27 22:06:57,200:DEBUG:letsencrypt.cli:Arguments: ['--apache', '-v']
2015-12-27 22:06:57,200:DEBUG:letsencrypt.cli:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
[...]

PR #2110, opened 2016-01-07
Precise bootstrap

pde:
This addresses the libaugeas0 version issues that are part (or perhaps all) of #2012.

### DIFF
--- a/bootstrap/_deb_common.sh
+++ b/bootstrap/_deb_common.sh
@@ -35,25 +35,39 @@ fi
 augeas_pkg=libaugeas0
 AUGVERSION=`apt-cache show --no-all-versions libaugeas0 | grep ^Version: | cut -d" " -f2`
 
+AddBackportRepo() {
+    # ARGS:
+    BACKPORT_NAME="$0"
+    BACKPORT_SOURCELINE="$1"
+    if ! grep -v -e ' *#' /etc/apt/sources.list | grep -q "$BACKPORT_NAME" ; then
+        # This can theoretically error if sources.list.d is empty, but in that case we don't care.
+        if ! grep -v -e ' *#' /etc/apt/sources.list.d/* 2>/dev/null | grep -q "$BACKPORT_NAME"; then
+            /bin/echo -n "Installing augeas from wheezy-backports in 3 seconds..."
+            sleep 1s
+            /bin/echo -ne "\e[0K\rInstalling augeas from wheezy-backports in 2 seconds..."
+            sleep 1s
+            /bin/echo -e "\e[0K\rInstalling augeas from wheezy-backports in 1 second ..."
+            sleep 1s
+            if echo $BACKPORT_NAME | grep -q wheezy ; then
+                /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
+            fi
+
+            echo $BACKPORT_SOURCELINE >> /etc/apt/sources.list.d/"$BACKPORT_NAME".list
+            apt-get update
+        fi
+    fi
+
+}
+
+
 if dpkg --compare-versions 1.0 gt "$AUGVERSION" ; then
     if lsb_release -a | grep -q wheezy ; then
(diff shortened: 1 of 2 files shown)


pde:
This will need two +1s because it's a letsencrypt-auto PR.  Perhaps @ArchimedesPi can also review?

ArchimedesPi:
LGTM.

joohoi:
The code itself looks good to me, however nosetests -v letsencrypt-apache still throws a lot of errors:
```
Ran 137 tests in 13.406s
FAILED (errors=17, failures=16)
```
in my vagrant box running precise64, also the configuration example from @ixevix still doesn't find the vhost that is found in Debian 7.

The vagrant box is the LE repo root Vagrantfile with trusty64 substituted with precise64, the offical image: http://pastebin.com/GXkNqtS2

joohoi:
Ah, the Vagrant image didn't provide lsb_release. Rechecking.

bootstrap/_deb_common.sh
```diff
@@ -35,25 +35,39 @@ fi
...
+            if echo $BACKPORT_NAME | grep -q wheezy ; then
                 /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
-
-                echo deb http://http.debian.net/debian wheezy-backports main >> /etc/apt/sources.list.d/wheezy-backports.list
-                apt-get update
             fi
+
+            echo $BACKPORT_SOURCELINE >> /etc/apt/sources.list.d/"$BACKPORT_NAME".list
+            apt-get update
+            apt-get install -y --no-install-recommends -t "$BACKPORT_NAME" libaugeas0
```
joohoi: This will never happen if user has backports already enabled, so the backports package won't get installed. This is the case on the default vagrant box.

joohoi:
With libaugeas0 from backports (Version: 1.2.0-0ubuntu1.1~ubuntu12.04.1) the nosetests -v letsencrypt-apache returns with FAILED (errors=21, failures=27)

ArchimedesPi:
@joohoi weird, I ran the tests against that backported version a couple months ago, I don't know what's changed.

joohoi:
@ArchimedesPi it could be some other package too, but the libaugeas0 is the most probable one.

ArchimedesPi:
Wait - IIRC, you have to install augeas-lenses from backports as well.

pde:
@ArchimedesPi I think we ship our own copy of the very latest httpd lens, so we shouldn't need that package...

ArchimedesPi:
After testing, I've found it *is* necessary to update `augeas-lenses`.

bootstrap/_deb_common.sh
```diff
@@ -35,25 +35,39 @@ fi
...
+            if echo $BACKPORT_NAME | grep -q wheezy ; then
                 /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
-
-                echo deb http://http.debian.net/debian wheezy-backports main >> /etc/apt/sources.list.d/wheezy-backports.list
-                apt-get update
             fi
+
+            echo $BACKPORT_SOURCELINE >> /etc/apt/sources.list.d/"$BACKPORT_NAME".list
+            apt-get update
+            apt-get install -y --no-install-recommends -t "$BACKPORT_NAME" libaugeas0 augeas-lenses
```
joohoi: The installing is still done inside if (backports repo not found in sources.list) - directive. So if user already has backports enabled, also the package installation from backports repo is skipped.

joohoi:
I installed the packages manually because of the problem pointed above, and can confirm - augeas-lenses package did the trick. All tests now pass, and the problematic example configuration gets parsed correctly!

bootstrap/_deb_common.sh
pde: I could swear that this is the third time I've had to fix that bug in this script :(

pde:
Passed basic unit tests for non-breakage of letsencrypt-auto.

Issue #1719, opened 2015-12-04
max_attempts in poll_and_ri should be per-authorization


kuba:
`poll_and_request_issuance([authz1, authz2], max_attempt=1)` is the simplest example, where the code fails, because `max_attempts` counts total number of attempts for all authorizations, instead of counting per authorization. With default limit of 10, this causes problems in practice - usually when people are using more than 10 domains.

Core is not affected, because it doesn't use this retry loop. https://github.com/kuba/simp_le, however, is (and any other client using our library).

kuba:
Fix is pending in #2111.

PR #2111, opened 2016-01-07
Fix "global" max_attempt bug (#1719)

kuba:
Honest disclaimer: this breaks the API (`PollError.waiting` list of 2-tuples turns into `PollError.exhausted` set). :)

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -1,4 +1,5 @@
 """ACME client API."""
+import collections
 import datetime
 import heapq
 import logging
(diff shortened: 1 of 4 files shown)


kuba:
@pde, any chance to squeeze this into upcoming [0.2.0](https://github.com/letsencrypt/letsencrypt/milestones/0.2.0)? :)

pde:
That didn't happen, but we can get it in for 0.2.1 next week.

pde:
LGTM.

pde:
Hmmm, looks like this may have broken travis tests on master, so reverting.

PR #2115, opened 2016-01-08
Docker option for production [needs revision]

goacid:
By default docker image point to stagging url. Add a command line for dokcer on production url.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -328,12 +328,27 @@ If you're still not convinced and have decided to use this method,
 from the server that the domain you're requesting a cert for resolves
 to, `install Docker`_, then issue the following command:
 
+
+
+Stagging :  
+Cert from Fake CA
+
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


kuba:
Docker image uses defaults of the CLI client, which currently is the production API endpoint (see https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/constants.py#L19). To be explicit, you only need a custom `--server` if you want to use Docker and staging API endpoint. So this is the other way round as suggested by you and your PR.

bmw:
You can also simply use `--staging` to use the staging URL.

bmw:
@goacid, there have been review comments on this for a month now and we haven't heard back from you. I'm closing this PR. Feel free to open a new one if you'd like us to rereview. Thanks!

goacid:
Hi
That is ok for me. I needed to update my image so that is was ok.

Issue #2002, opened 2015-12-23
log changes to Apache conf files right before performing .reload()


pde:
This should make it much easier to see what's going wrong in cases like #1994 where the changes we think we should have made to conf files apparently aren't quite what's happened.

SwartzCr:
@pde how does https://github.com/SwartzCr/letsencrypt/commit/c9b9b0edda27c3be3173fdcda0e98bcbd4995b3c look?

pde:
Done.

pde:
Had to revert that PR because of unintended iDisplay.notification() interactivity, so reopening this ticket.

SwartzCr:
can you give more info on the error and how to repro it?

SwartzCr:
Oh I see it's in the PR ( #2009 )
That's garbage, I wonder if there's a way to get the logger from elsewhere so it doesn't hit that code in reverter.
On the other hand, it seems like we're making good progress with the bug that this was meant to help debug, so maybe it's not worth spending too much time working on.
I'm going to focus on #2012 for now

PR #2116, opened 2016-01-08
don't iDisplay if logging

SwartzCr:
add a flag variable to view_config_changes which returns the changes as a string if we're logging it.
I'm unsure how to test since the comments left in https://github.com/letsencrypt/letsencrypt/pull/2009 don't specify where this issue was arising


SwartzCr:
to fix #2002

joohoi:
TBH, I'd simplify it all by just logging.debug:ing the changes every time in view_config_changes().

SwartzCr:
so you mean, log then iDisplay? Will that get called when tls_sni_01.py gets called?

joohoi:
Ah, you are right. Sorry, was reviewing in a hurry. 

LGTM

pde:
Hrm, spotted after merging:

It's a little bit dangerous to have the `for_logging` argument on one implementation of `view_config_changes`, but not in the [interface it's part of](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/interfaces.py#L333) and all the other implementations.  Currently it's only safely usable if the call site knows for sure it's calling this method on the apache plugin, and not a different plugin.

So we should ask ourselves, "is this the semantics we want from the interface?" If so, add it to the interface and also implement it in the other plugins.  Prrrobably it's safe to leave this merged and just open a ticket to make that decision.

PR #2117, opened 2016-01-08
letsencrypt-auto: prevent recording of deps in Gentoo's world set

osirisinferi:
(This is a second attempt of a previous PR I made and consequently f*cked up.. I closed that one.. :stuck_out_tongue: This one is with `git` on my local machine, so hopefully any conflicts/upstream master merging/whatever won't FUBAR it anymore.. :disappointed: )

Dependencies shouldn't be recorded into Gentoos [world set](https://wiki.gentoo.org/wiki/World_set_(Portage)). The current single switch to `emerge` "--noreplace" makes sure already installed dependencies are explicitly recorded into the world set, which isn't desirable. Adding "--oneshot" prevents this behaviour.
Cave: I have absolutely no clue how Paludis works. It's [manual](http://paludis.exherbo.org//clients/cave-resolve.html) and an entry in their [news](http://paludis.exherbo.org/news.html) (0.26.0_pre3) suggests --preserve-world should do the trick.

(Credits: [till.schaefer](https://community.letsencrypt.org/t/gentoo-letsencrypt-should-not-automatically-record-packages-to-the-world-config-file/7252))

By the way: Gentoo has a [native package](https://packages.gentoo.org/packages/app-crypt/letsencrypt) since Dec 3.

### DIFF
--- a/bootstrap/_gentoo_common.sh
+++ b/bootstrap/_gentoo_common.sh
@@ -12,12 +12,12 @@ PACKAGES="
 
 case "$PACKAGE_MANAGER" in
   (paludis)
-    cave resolve --keep-targets if-possible $PACKAGES -x
+    cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
     ;;
   (pkgcore)
-    pmerge --noreplace $PACKAGES
+    pmerge --noreplace --oneshot $PACKAGES
     ;;
   (portage|*)
-    emerge --noreplace $PACKAGES
+    emerge --noreplace --oneshot $PACKAGES
     ;;
 esac
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -378,20 +378,49 @@ Packages for Debian Jessie are coming in the next few weeks.
 
 **Gentoo**
 
+The official Let's Encrypt client is available in Gentoo Portage. If you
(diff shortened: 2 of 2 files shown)


pde:
Hi @osirisinferi.  Thanks for your help, and apologies for being slow with this PR. We don't currently have any of our `lestencrypt-auto` and apache test farm tests running on gentoo, mostly because all the compilation makes them much, much slower than deb and rpm based systems.  However I'm okay merging your work here if @erikrose agrees and you're willing to do followup and troubleshooting.

osirisinferi:
Hey @pde 
I've got no problem doing the Gentoo follow up and troubleshooting. I'm not going to change distros any time soon and would like to make the LE client as flawless as it possibly can be on Gentoo ;)
That said, I've promised @joohoi in https://github.com/letsencrypt/letsencrypt/issues/2216 I'd add some more documentation about getting the client to work properly on Gentoos not-so-standard configuration, but I haven't made time for it yet. I could always open another PR though, but it might clutter things more.

erikrose:
@osirisinferi Thanks for the good patch! I took the liberty of porting it to the latest le-auto infra so we could merge it in. Sorry for the delay!

osirisinferi:
No problem obviously ;)

PR #2120, opened 2016-01-08
Don't use cryptography version 1.2

bmw:
PyOpenSSL fails when using this version on Ubuntu 12.04.5 (Travis). I modified `setup.py` to install an older version of cryptography until pyca/pyopenssl#407 is resolved. Once it is, these changes should be reverted.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -9,7 +9,7 @@ version = '0.2.0.dev0'
 install_requires = [
     # load_pem_private/public_key (>=0.6)
     # rsa_recover_prime_factors (>=0.8)
-    'cryptography>=0.8',
+    'cryptography>=0.8,<1.2',
     # Connection.set_tlsext_host_name (>=0.13), X509Req.get_extensions (>=0.15)
     'PyOpenSSL>=0.15',
     'pyrfc3339',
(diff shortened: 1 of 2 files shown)


erikrose:
lgtm. If we were sure they were going to fix it in 1.2.1, we could say `!=1.2`, but we're not, so we shouldn't.

alex:
It looks like we are fix is 1.2.1, @reaperhulk or I will send a PR changing this then.

alex:
Sorrya bout the breakages :-(

erikrose:
Thanks, @alex!

reaperhulk:
@erikrose @bmw we've released 1.2.1 to correct this now. I've also filed an issue (https://github.com/pyca/cryptography/issues/2656) to add a builder to prevent this situation from occurring again in the future.

eric-tucker:
ImportError: No module named cryptography.hazmat.bindings.openssl.binding

Did this merge cause this?

bmw:
Thanks for such a quick response @alex @reaperhulk. Just created the PR to revert this one.

@eric-tucker, it shouldn't have (plus it's about to be reverted). Can you provide some context on where the error occurred, potentially opening a new issue to track the problem?

PR #2121, opened 2016-01-08
Revert "Don't use cryptography version 1.2"

bmw:
Reverts letsencrypt/letsencrypt#2120


pde:
LGTM.

eric-tucker:
I'm in favour of this as I think it broke setup / execution on redhat.

PR #2129, opened 2016-01-09
Python 3 Travis testing for acme.

kuba:
Despite its description, https://github.com/letsencrypt/letsencrypt/pull/630, removed not only Python 2.6 support, but also Travis tests against Python 3. ACME library supports Python 3 and Travis should tests it.

This must be merged before any pending PRs against acme library.


pde:
LGTM

PR #2130, opened 2016-01-09
Default: renew 30 days before expiry, rather than 10

pde:
- gives more time for various fallback strategies if renewal doesn't work the
   first time

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -565,7 +565,8 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
                 return True
 
             # Renews some period before expiry time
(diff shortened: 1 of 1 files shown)


eric-tucker:
Slightly off topic, but is there any documentation on how renewal currently works?

pde:
@eric-tucker not good docs yet but we should have some up in a few days.  In short, run the client like this `letsencrypt certonly --keep -d example1.com,example2.com`, plus the flags for whatever plugin you're using.  It'll renew the cert N days before expiry.

pde:
@kuba yes I know, but I wanted to not break the link to the logs in IRC.

bmw:
LGTM

PR #2136, opened 2016-01-10
.gitattributes: EOL=LF overrides auto

tboegi:
The line
* text=auto eol=lf
in .gitattributes does not work as expected:
setting eol=lf overrides text=auto, and works as
* eol=lf
or
* text eol=lf

This is not what is intended, as binary files like *.gz go through
a CRLF -> LF conversion at commit time.

Use
* crlf=auto
instead (which is even understood by older Git versions, which are still in use)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,15 @@
-* text=auto eol=lf
+#Default, normalize CRLF into LF in non-binary files
+# Files identified as binary by Git are not changed
+* crlf=auto
 
 # special files
+*.sh crlf=input
+*.py crlf=input
+
 *.bat text eol=crlf
+
+*.der binary
+*.gz binary
 *.jpeg binary
 *.jpg binary
 *.png binary


kuba:
Changes around `.gitattributes` CR/LF have proven to be problematic in the past (e.g. broken `git blame`). This should be reviewed / merged with double caution!

tboegi:
Yes, one problem was introduced in 
commit e9a2180d16b:
The file docs/make.bat had originally be commit with CRLF into the repo.
After adding the .gitattributes file, the repo should have been normalized,
as described here:

https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html
$ echo "* text=auto" >>.gitattributes
$ rm .git/index     # Remove the index to force Git to
$ git reset         # re-scan the working directory
$ git status        # Show files that will be normalized
$ git add -u
$ git add .gitattributes
$ git commit -m "Introduce end-of-line normalization"

(I don't know why this didn't happen, hints how to improve the Git documentation are welcome)

In any case, the current master is normallized.

bmw:
LGTM.

Sorry this took so long to review, but thanks for the PR @tboegi.

PR #2146, opened 2016-01-12
Copy only relevant lines from http vhost to ssl vhost skeleton v2

bmw:
Makes minor fixes to #2075.

TODO (I will create a new issue when this is merged): Cleanup reporter message. It has the potential to spam the user if rewrite rules are stripped from many vhosts.

ping @pde


pde:
LGTM!

PR #2151, opened 2016-01-12
added support for authentication via FTP and SFTP

ntamas:
I have created a plugin for my own purposes that allows the user to proceed through the authentication by FTP-ing files to a remote server.

This is useful for shared web hosting providers where the hosting company does not provide full SSH access but allows FTP and/or SFTP (like mine). The plugin is similar to the `webroot` plugin in the sense that the user must define "web roots" for the domains and the plugin will create files there to make the server pass the authentication step. I know that this could probably have been solved using FUSE with `curlftpfs`, but the advantage of this method is that there is no need to mount a file system on the local machine (for which the user might not have privileges anyway), and it works out-of-the-box on a vanilla Mac where FUSE is uncommon.

The plugin is not feature-complete yet and there are no tests specific to the FTP/SFTP functionality, but I thought I'd share it anyway to gauge whether there's any interest for such a plugin. If there is, I am happy to keep on working on the code until it is of acceptable quality.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -510,6 +510,8 @@ def choose_configurator_plugins(args, config, plugins, verb):  # pylint: disable
         req_auth = set_configurator(req_auth, "webroot")
     if args.manual:
         req_auth = set_configurator(req_auth, "manual")
+    if args.ftp:
+        req_auth = set_configurator(req_auth, "ftp")
     logger.debug("Requested authenticator %s and installer %s", req_auth, req_inst)
 
     # Try to meet the user's request and/or ask them to pick plugins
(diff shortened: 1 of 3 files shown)


pde:
See also my recent comment on #1419

bmw:
So after talking about this a lot with the client team, we decided against adding an explicit FTP/SFTP plugin or adding such functionality to webroot. Instead, we're planning to rework the manual plugin, allowing the user to provide scripts that can be used for the purposes of solving and cleaning up challenges (see #2782). Using this, the equivalent functionality of a FTP plugin can be implemented in a couple lines of bash.

Regardless, thanks for the PR @ntamas  and I'm sorry it sat around so long.

bmw:
I wanted to add to this that although we're not taking this plugin into our tree, we certainly would like to support the development of third party plugins for Certbot. If you'd like to host this on your own repo, we'll happily link to it in our [list of third party plugins](https://certbot.eff.org/docs/using.html#third-party-plugins).

PR #2161, opened 2016-01-13
Cache Python packages

bmw:


kuba:
Does `tox` use this directory, i.e. would `tox`-used packages be cached as well?

pde:
@kuba it appears to do so for constructing things under `.tox/`, yes.

pde:
LGTM

PR #2169, opened 2016-01-13
Check that recent enough version of libaugeas is available

### LINKED ISSUE #2143: The apache plugin shouldn't report itself as available if augeas is too old
We'll need this to ensure that as Apache 2.2 support is added, people who bootstrapped on wheezy, precise, or some other old OS, aren't getting a plugin that doesn't work.

joohoi:
Print error message if the augeas version doesn't support case insensitive regexes.

Fixes #2143


pde:
This looks good, pending travis passing.

pde:
Woohoo!

PR #2174, opened 2016-01-14
Adding support for boulder integration testing to Vagrantfile

TheNavigat:
Script works but needs some cleanup.


TheNavigat:
This is getting slightly debatable, so here's the arguments we've had over the IRC laid down.

I believe adding executing ```boulder-start.sh``` on boot would make it way easier since you'd only need to execute ```./tests/boulder-integration.sh```, instead of running ```./tests/boulder-start.sh```, waiting for it to finish, . That's the way I personally have it set up, makes it way easier.

However, as @bmw noted, that would conflict with the docs. Specifically, ```docs/contributing.rst``` which specify that ```boulder-start``` should be executed before ```boulder-integration```.

As ErikRose (without an @ before he asked not to be added to the thread), it would be nice to have Vagrant running as a box where the only thing required to execute the tests would be simply running them.

He also mentioned (and I quote) "Typically, my contribution docs say "Use the vagrant box. Alternately, if you really must, here's a long list of instructions to set it up on your own machine." - which I personally agree with.

Therefore, I'd think it'd be really nice to add running ```./tests/boulder-start.sh``` on boot, and modifying the docs to reflect that.

Vagrantfile
```diff
@@ -5,10 +5,18 @@
...
 
 # Setup instructions from docs/contributing.rst
+# Script installs dependencies for tox and boulder integration
 $ubuntu_setup_script = <<SETUP_SCRIPT
 cd /vagrant
 ./bootstrap/install-deps.sh
 ./bootstrap/dev/venv.sh
+wget https://storage.googleapis.com/golang/go1.5.3.linux-amd64.tar.gz -P /tmp/
+sudo tar -C /usr/local -xzf /tmp/go1.5.3.linux-amd64.tar.gz
+if ! grep -Fxq "export GOROOT=/usr/local/go" ~/.profile ; then echo "export GOROOT=/usr/local/go" >> ~/.profile; fi
```
bmw: You should be able to successfully run integration tests as `vagrant` user so these should be appended to `/home/vagrant/.profile` also/instead.

Vagrantfile
```diff
@@ -5,10 +5,18 @@
...
+if ! grep -Fxq "export PATH=\\$GOROOT/bin:\\$PATH" ~/.profile ; then echo "export PATH=\\$GOROOT/bin:\\$PATH" >> ~/.profile; fi
+if ! grep -Fxq "export GOPATH=\\$HOME/go" ~/.profile ; then echo "export GOPATH=\\$HOME/go" >> ~/.profile; fi
+export DEBIAN_FRONTEND=noninteractive
+sudo -E apt-get -q -y install git mysql-server-5.5 libltdl-dev rabbitmq-server make nginx
```
bmw: A few things here:

1. Why do we deviate from the installation instructions found [here](https://github.com/letsencrypt/boulder#boulder---an-acme-ca)? We shouldn't be doing this unless there is a reason to do so.
2. nit: We can get away with `nginx-light` package to make setup marginally faster.

bmw:
I agree we should run `boulder-start.sh` on boot as long as we update the documentation accordingly.

This is a good work in progress @TheNavigat. Thanks for coding it up!

Vagrantfile
TheNavigat: 1. The instructions on Boulder's README page is outdated. Boulder requires MySQL, nor MariaDB.
2. Will get to that in a bit.

TheNavigat: 1. Turned out to be my own mistake. MariaDB is a fork of MySQL, it looks.
2. Works, will add that!

TheNavigat:
Added [needs revision] since a conflict needs to be resolved. Will resolve and remove the tag.

TheNavigat:
Conflict fixed, removed [needs revision]

TheNavigat:
Ping @bmw

docs/contributing.rst
```diff
@@ -96,6 +96,14 @@ Integration testing with the boulder CA
 Generally it is sufficient to open a pull request and let Github and Travis run
 integration tests for you.
 
+However, if you prefer to run tests, you can use Vagrant, using the Vagrantfile
+in Let's Encrypt's repository. To execute the tests on a Vagrant box, the only
+command you are required to run is::
```
bmw: `s/::/:`

bmw:
Other than the small typo in the documentation, this lgtm.

Merging when fixed.

docs/contributing.rst
TheNavigat: I want the command to me parsed correctly as a literal block, though :/

bmw: Ah. Fair enough. I didn't realize `::` was a thing. Merging now.

PR #2181, opened 2016-01-14
Tell user letsencrypt was  installed correctly. 

### LINKED ISSUE #1621: After successful install, print an "Install appears to have been successful" message

cowlicks:
closes #1621

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -198,6 +198,8 @@ else
   echo
 fi
 
+echo "Install appears to have been succesful."
+
 # Explain what's about to happen, for the benefit of those getting sudo
 # password prompts...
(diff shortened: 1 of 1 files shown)


cowlicks:
Is this in the right place?

pde:
If you do it in that spot, it's fairly spammy because it gets printed on every single run.  It might be better to do at the end of the bootstrapping process, so that it only happens when install happens.

Having said all of that, `letsencrypt-auto` is getting a major overhaul in #1655 and so probably you should wait until that lands, or make another PR against the branch for #1665 itself.

pde:
(Closing this version because I think we should only land important bugfixes against the old leauto)

Issue #2175, opened 2016-01-14
Conditional python-ndg-httpsclient dependency doesn't work on Ubuntu 14.04


pde:
During release testing:
```
export PIP_EXTRA_INDEX_URL=https://isnot.org/pip
./letsencrypt-auto --version -v
Bootstrapping dependencies for Debian-based OSes...
Ign http://us-east-1.ec2.archive.ubuntu.com trusty InRelease
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates InRelease           
Hit http://us-east-1.ec2.archive.ubuntu.com trusty Release.gpg                 
Hit http://us-east-1.ec2.archive.ubuntu.com trusty Release                     
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/main Sources        
Hit http://security.ubuntu.com trusty-security InRelease                       
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/universe Sources    
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/main amd64 Packages 
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/universe amd64 Packages
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/main Translation-en 
Hit http://us-east-1.ec2.archive.ubuntu.com trusty-updates/universe Translation-en
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/main Sources                
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/universe Sources            
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/main amd64 Packages         
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/universe amd64 Packages     
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/main Translation-en         
Hit http://us-east-1.ec2.archive.ubuntu.com trusty/universe Translation-en     
Hit http://security.ubuntu.com trusty-security/main Sources                    
Hit http://security.ubuntu.com trusty-security/universe Sources                
Hit http://security.ubuntu.com trusty-security/main amd64 Packages             
Ign http://us-east-1.ec2.archive.ubuntu.com trusty/main Translation-en_US      
Ign http://us-east-1.ec2.archive.ubuntu.com trusty/universe Translation-en_US  
Hit http://security.ubuntu.com trusty-security/universe amd64 Packages
[...]

PR #2182, opened 2016-01-14
Always install ndg-httpsclient and pyasn1

bmw:
Solution to #2175 for `0.2.0`. We can reconsider making these dependencies conditional in the next release after more testing.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -11,6 +11,8 @@ install_requires = [
     # load_pem_private/public_key (>=0.6)
     # rsa_recover_prime_factors (>=0.8)
     'cryptography>=0.8',
+    'ndg-httpsclient',  # urllib3 InsecurePlatformWarning (#304)
+    'pyasn1',  # urllib3 InsecurePlatformWarning (#304)
     # Connection.set_tlsext_host_name (>=0.13)
     'PyOpenSSL>=0.13',
     'pyrfc3339',
@@ -30,11 +32,6 @@ if sys.version_info < (2, 7):
     ])
 else:
     install_requires.append('mock')
-
-if sys.version_info < (2, 7, 9):
-    # For secure SSL connection with Python 2.7 (InsecurePlatformWarning)
-    install_requires.append('ndg-httpsclient')
-    install_requires.append('pyasn1')
 
 docs_extras = [
     'Sphinx>=1.0',  # autodoc_member_order = 'bysource', autodoc_default_flags
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ version = meta['version']
 # Please update tox.ini when modifying dependency version requirements
 install_requires = [
     'acme=={0}'.format(version),
+    'ConfigArgParse>=0.10.0',  # python2.6 support, upstream #17
     'configobj',
     'cryptography>=0.7',  # load_pem_x509_certificate
     'parsedatetime',
@@ -52,12 +53,10 @@ if sys.version_info < (2, 7):
     install_requires.extend([
         # only some distros recognize stdlib argparse as already satisfying
         'argparse',
-        'ConfigArgParse>=0.10.0',  # python2.6 support, upstream #17
         'mock<1.1.0',
     ])
 else:
     install_requires.extend([
-        'ConfigArgParse',
         'mock',
     ])
 


bmw:
Me appending `[needs revision]` was a false alarm.

Natim:
This sounds good :+1:

erikrose:
Aren't we going to have similar problems with the conditional argparse dependency?

Natim:
Yes tox is doing it like that: https://bitbucket.org/hpk42/tox/src/2d5d0e7584cc4cc35cc7e0519ce9610dd52b7a62/setup.py?fileviewer=file-view-default#setup.py-42

erikrose:
We should be able to get away with leaving the mock ones in there (which have to be <1.1 or >=1.1.4 for 2.6) since mock isn't used at runtime.

pde:
Actually it's sort of a bug but we don't have any test farm scripts that simulate letsencrypt-auto, without actually having a signed release.  That should be pretty doable, especially after #1655 lands, but in the mean time, we'll have to test this after deploying a build of it to isnot.org

pde:
I gather this needs revision to revert argparse dependency changes?

erikrose:
@pde That's how I see it.

bmw:
LGTM

pde:
Landing!

kuba:
Urm, what is the real plan of handling that whole mess? Submitting random PRs doesn't sound good to me.

erikrose:
I spent some time today really digging into this, plowing through old release notes and talking to the PyPA people. (I owe thanks to dstufft for his oracular knowledge about packaging corner cases.) Here's what I found.

As already established, our main problem is wheels. Wheels install faster and more deterministically, but, consequently, they don't run our setup.py conditional logic; in fact, they contain no setup.py. While one can build a wheel specific to a Python version, this goes only as far as tenths-place versions; we cannot ship a wheel targeted at >=2.7.10, for instance.

Environment markers offer a way around this, but support for them is determined by the versions of 2 tools: setuptools and pip.

The setuptools version matters whenever setup.py is run. For sdists, this is at install time. For wheels, this is at build time. It must be >=17.1 (released June 2015) to support < and > operators; == is supported as early as 0.7 (mid-2013 but mysteriously missing from PyPI). As packagers of LE, we control the latter but not the former. There is a workaround to support env markers earlier: list the requirements, with markers, explicitly in setup.cfg, as [in](https://github.com/pypa/twine/blob/master/setup.cfg#L9-L15) [twine](https://github.com/pypa/twine/blob/master/setup.py#L28-L31), but that requires entirely duplicating the install_requires list and keeping it in sync, since bdist_wheel ignores the install_requires in setup.py whenever it finds a requires-dist option in setup.cfg. Because of the opportunity for error, I disregard this option below.

The second variable, the pip version, matters only for wheels. pip >=6 (December 2014) understands and obeys all environment markers laid down in wheel metadata.

All of the above matter only to people installing LE packages by hand: generally, I would hope, developers. The old le-auto solves these by updating setuptools and pip as a first step; the new, by pinning the entire dependency tree.

Combining all that, the reasonable options are...
1. Use the full-on < environment markers, and require pip >=6 and setuptools >=17.1. (Advantage: only the minimal set of requirements is installed.)
2. Use only == env markers, and require pip >= 6 and setuptools >= 0.7.
3. Use only unconditional dependencies. Conveniently, this is a workable option with our codebase. (Advantages: fewer different code paths run on different platforms, so we're likely to miss fewer bugs. Also, fewer code paths are needed in the new le-auto.)
4. Don't ship wheels at all, so our setup.py logic always has a chance to run. (Advantage: support the oldest pips and setuptools.)

erikrose:
Now I'll add my value judgements and pick favorites:

(1) and (2) will probably generate some tech-support issues for us as LE devs use out-of-date tools and non-LE-dev Python people, in an effort to stick to their accustomed tooling, manually install LE into virtualenvs.

(4) combines the worst of all worlds: lots of branches, lowest performance for end users, and the appearance to devs of living in the past.

(3) is my favorite. Its only concrete disadvantage is an extra 200K to download beyond optimal for Python>2.7.9, and it makes our system simpler rather than more complex.

dstufft:
To be clear, for *wheel* metadata pip has supported it for as long as it's supported wheels (pip 1.4, but there were some unrelated bugs with pip 1.4 and wheels). pip 6+ is only needed for environment markers inside of an *sdist*.

PR #2183, opened 2016-01-14
Include requested domain in error message

stapelberg:
This is especially helpful when providing multiple domains.


bmw:
Closed an reopened hoping Travis would decide to run but no luck!

bmw:
After looking at this closer, it looks like equivalent code was landed in #2309, so I'm closing this again.

Thanks for the PR @stapelberg. I'm sorry we didn't get around to reviewing it before someone else had the idea.

PR #2185, opened 2016-01-14
More testfarm

pde:
Add gentoo 2.2 and freebsd 10 to the test farm.

* those were our most widely used untested platforms.  Darwin 10 is next.


pde:
This is not ready to fly.

PR #2187, opened 2016-01-15
Fix tests on letsencrypt-apache packages

bmw:


pde:
Good to merge if tests pass.

PR #2193, opened 2016-01-15
Don't specify mock-version to be more flexible

weberhofer:


kuba:
This is a lot more complicated, I'm afraid. There are a couple of floating PRs that are conflicting with this change.

pde:
We can't land something like this unless we deprecate py26, and at the moment, that's 10-15% of our userbase.

PR #2199, opened 2016-01-15
Include 0.2.0 release in master's history, and bump version

pde:


kuba:
I think the bump should be `0.3.0.dev0`, last component is meant to be a bugfix release only and not a planned in advance new version. At least this is how I understand semantic versioning - and I believe we agreed earlier on such scheme.

pde:
We are planning an 0.2.1.dev0 release for this week, so that's what the bump is to.

pde:
If significant features wind up landing and we change to 0.3.0, then we change to 0.3.0.

kuba:
Never release `*dev*` version! I hope you meant `0.2.1`.

According to semantic versioning, last component is for bugfixes only (also, what will you do if you need in future to fix a bug in 0.2.0 release?). FYI there are also some API changes that went into ACME module (`acme.errors`). For that reason, please release 0.3.0 instead.

bmw:
For the time being, pde and I decided to always bump the patch version while doing releases. This prevents us from deciding on the next version number before the release is ready and aims to be more conservative (instead of creating 1.0.0.dev0 and then releasing 0.3.0.dev0).

Kuba, if there's a problem with this logic, please let me know. LGTM. Merging.

kuba:
Come on... we agreed (and voted!) on something else during our conversation at IRC. :( Moreover I don't see any further conversation (on IRC) after I went to bed. So I can't really address you concerns, and I don't see anyone actually addressing mine. `1.0.0.dev0` comment doesn't make much sense to me as we've always done `0.X.0.dev0`.

Also, does this mean you will release `0.3.0` as the next one, or `0.2.1`?

PR #2211, opened 2016-01-17
add basic support for OpenBSD bootstrap

phreakocious:


erikrose:
Looks sane to me, but it needs some updating: the bootstrap scripts now live in letsencrypt-auto-source/pieces/bootstrappers and have changed structure a little.

bmw:
I actually just closed #1175 since OpenBSD has official packaging. With this in mind, do people still want this?

erikrose:
There's also #2048.

I'm not sure. I lean toward "no"; the only reason to use le-auto on a platform that has its own packages is if said packages don't keep up.

bmw:
I have the same inclination. If the community really wants it, I could be swayed, but in general I think we should be encouraging people with native packing to use it as opposed to `letsencrypt-auto`.

erikrose:
Good. Our official position is "use the packages", until somebody makes a compelling case otherwise.

PR #2213, opened 2016-01-17
Add support for optional trailing period in FQDNs

### LINKED ISSUE #2065: FQDN passed with a trailing period is not valid.
```
Requesting root privileges to run with virtualenv: ~/.local/share/letsencrypt/bin/letsencrypt run --apache --email ...@....... --redirect --agree-tos -d mysubdomain.mydomain.io.
Requested domain is not a FQDN
```

It is valid to refer to a domain using the trailing period, and I had expected letsencrypt would handle it as such too.

filipochnik:
Fixes #2065

### DIFF
--- a/letsencrypt/le_util.py
+++ b/letsencrypt/le_util.py
@@ -314,7 +314,7 @@ def check_domain_sanity(domain):
     # FQDN checks from
     # http://www.mkyong.com/regular-expressions/domain-name-regular-expression-example/
     #  Characters used, domain parts < 63 chars, tld > 1 < 64 chars
-    #  first and last char is not "-"
-    fqdn = re.compile("^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,63}$")
+    #  first and last char is not "-", optional trailing period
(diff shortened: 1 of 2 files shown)


kuba:
Have you actually tested it against Boulder (e.g. testing instance)?

filipochnik:
Right, my bad. Can you tell me how to do that/point to the docs?

kuba:
Easiest way to do that is to use staging API endpoint by using `--staging`

filipochnik:
Thanks! Turns out it's not supported on the Boulder side.
`letsencrypt certonly --test-cert --text --webroot --webroot-path /tmp --email mail@test.com -d www.test.com. --staging`
fails with
`The request message was malformed :: Error creating new authz :: DNS label is too short`

I'll file an issue on the Boulder's repo.

pde:
I doubt that one actually wants a cert with a trailing period in its names; there are bound to be client implementations out there that choke on such certs.  Would anything go wrong if the client just stripped the trailing dot to re-normalise the name?

pde:
The Decentralized SSL Observatory has seen about 13,000 certs with names that end in a dot.  However it's only ever seen one such certificate that's valid in browsers, a cert for `webmail.` that was signed by Calvados and was valid from 2012-2013.

filipochnik:
Makes sense. I'm closing this PR then, I'll submit a new one with proper fix.

PR #2214, opened 2016-01-17
Print message about successful installation #1621

msoedov:
#1621 Added `echo` to the end of the bootstrapping process in `letsencrypt-auto` script


pde:
This is better than #2181 because it solves the spaminess problem. +1 from me, we'll need another +1 for a le-auto PR.

Atrolantra:
Looks good. +1 from me too.

erikrose:
Ported this to the new le-auto. Thanks for pointing out a need!

PR #2223, opened 2016-01-18
Revert "enable config_test in configurator prepare"

pde:
Does reverting letsencrypt/letsencrypt#2165 fix travis?


pde:
This wasn't the thing that needed reverting.

PR #2226, opened 2016-01-18
Reorder travis tests for performance

pde:
- lint is slow, so get started on it earlier
 - py33 and py34 are fast, so it's fine if they start very late


pde:
I think this reordering was provided by the other changes to .travis.yml that just landed.

PR #2237, opened 2016-01-19
Bump version number

bmw:
Bumps version number to `0.2.1.dev0`. For the time being, pde and I decided to always bump the patch version while doing releases. This prevents us from deciding on the next version number before the release is ready and aims to be more conservative (instead of creating `1.0.0.dev0` and then releasing `0.3.0.dev0`).

I believe this should be preferred over #2199 because we have never included the non-dev version number in git history before (at least Kuba or I never did). Instead, the non-dev version simply exists in a tag as a reference.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 from setuptools import find_packages
 
 
-version = '0.2.0.dev0'
+version = '0.2.1.dev0'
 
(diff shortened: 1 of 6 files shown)


pde:
I think we should include the non-dev version number in master's git history, because that allows folks to easily see where the release happened relative to other developments.

bmw:
Good point. Closing this PR and merging yours.

PR #2238, opened 2016-01-19
Enable pip cache and set Python version

bmw:
Builds off #2160 (which builds off #2030). My theory for why this didn't work before is Travis didn't know Python version our tests were running, causing it to use the wrong `pip` cache.

Building off of #2160 should fix this. I will run Travis tests multiple times and verify the cache is used in each test to prevent the problems that occurred before.


bmw:
Yay it works! In the log for job in the latest Travis build for this PR you'll see:
```
Setting up build cache
$ export CASHER_DIR=$HOME/.casher
$ Installing caching utilities
attempting to download cache archive
fetching PR.2238/cache--python-{2.6,2.7,3.3,3.4,3.5}.tgz
found cache
adding /home/travis/.cache/pip to cache
```
where the version of the tarball is the Python version of the test being run.

pde:
LGTM, although I note that the travis runtime in this PR isn't especially low :(.  Go ahead and merge the dependencies and this branch too?

bmw:
Travis build times seem to be varying wildly. With that said, comparing the job times here to my version bump PR, this is about 70 seconds faster on average assuming I didn't screw up my math.

Merging.

Issue #1668, opened 2015-12-02
Several code paths (conf files, interactive domains, --csr) do not set webroot_map entries correctly


tobbez:
This may or may not be intended.

If you set `webroot-path` in your config and don't specify your domains, letsencrypt fails with the following error:
```
The webroot plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('--webroot-path must be set',)
```
This is pretty confusing, both because `webroot-path` is set, and also because the code in question actually uses `webroot-map`, which is autogenerated in cli.py if both webroot and domains are specified upon invocation.

Previously this use pattern worked.

These changes were introduced in e1f0fcc and ffe6226e.

pde:
This sounds like a bug.  Perhaps when webroot-path is set via the config file, [WebrootPathProcessor](https://github.com/letsencrypt/letsencrypt/pull/1594/files#diff-54b2f468d57ef7436008bd031bc89711R1065) isn't called?  Or when the fields come from the config file, some of the ordering assumptions that underpin [these changes](https://github.com/letsencrypt/letsencrypt/pull/1594/files) might be violated?

In the mean time, you can set `webroot-map` in your config file.  It has to look like this IIRC:
```
webroot-map = {"domain.com,www.domain.com" : "/var/www/domain", "eg.is" : "/var/www/is"}
```
Unfortunately, you can't have newlines in there ATM because ConfigArgParse can't cope with them.  For now domains must be specified in addition to webroot-map, though in the future the latter may imply the former.

tobbez:
`WebrootPathProcessor` is called. However, if you haven't specified any domains before `webroot-path` (either in the config file or on the command line), the `WebrootPathProcessor` won't do anything, since `config.domains` is empty.

Later, when the `prepare` method of Authenticator (in [letsencrypt/plugins/webroot.py](https://github.com/letsencrypt/letsencrypt/blob/a191daf3eedcabc175a2de71abb0d29989a5c343/letsencrypt/plugins/webroot.py#L49)) is called, it only uses `webroot-map`, which is still empty since it was never populated.

pde:
@tobbez want to send us a PR?  That will let us land a fix more quickly :)

tobbez:
I'm just not sure what the right way to fix it is.

You could call `display_ops.choose_names` if `config.domains` is empty in `WebrootPathProcessor`, but this might lead to confusing/inconsistent behaviour if the `webroot- path` is specified before domains.

[...]

PR #2241, opened 2016-01-20
Support trailing period in domain names

filipochnik:
Removes trailing period if present in the domain name #2065
Should the same be done for domain names passed in webroot_map? It's not really documented so I wasn't sure how to handle it.


pde:
The integration between `domains` and `webroot-map` is imperfect and needs work (#1668) but the aim is to have everything work the same way there.  So I'd say mirror behaviour in both cases, otherwise document as extra work being created in #1668.

filipochnik:
OK, now they both strip trailing period. `webroot-map` is still buggy though, I'll submit a separate PR for that.

pde:
This looks good, but rather than merging it directly I've included it as part of #2309.  @filipochnik would you like to review that overall PR?

PR #2248, opened 2016-01-20
Downgrade declared ConfigArgParse requirement. Fix #2243.

### LINKED ISSUE #2243: VersionConflict: ConfigArgParse 0.9.3
Hello,

I tried renewing our certificates today, and I got an error from letsencrypt-auto. The error message was:

VersionConflict: (ConfigArgParse 0.9.3 (/home/bogdanb/.local/share/letsencrypt/lib/python2.7/site-packages), Requirement.parse('ConfigArgParse>=0.10.0'))

Please see the two attached files for the logs:
[execution.log.txt](https://github.com/letsencrypt/letsencrypt/files/97381/execution.log.txt)
[letencrypt.log.txt](https://github.com/letsencrypt/letsencrypt/files/97383/letencrypt.log.txt)

This is after I had deleted the ~/.local folder where letsencrypt stores its downloads. (I had the same exception before.) This is on an Ubuntu 15.04 machine:

$ uname -a
Linux qotilabs-dime 2.6.32-042stab112.15 #1 SMP Tue Oct 20 17:22:56 MSK 2015 x86_64 x86_64 x86_64 GNU/Linux

erikrose:


erikrose:
Let's see if the 2.6 tests fail. If they do, I'd rather fudge Kuba's python2.6-0.9.3 tag instead, replacing it with a copy of 0.10.0.

erikrose:
I'd actually rather fudge the tag regardless, since it leaves the weirdness in the past, in a mislabeled git branch, rather than dragging it into the indeterminate future with us in setup.py.

erikrose:
Fudging the tag instead: https://github.com/kuba/ConfigArgParse/pull/1

pde:
Because Kuba didn't land that before going on vacation, I think we should reopen, merge and ship this PR.  I believe we can revert later, once kuba's ConfigArgParse is 0.10.0

pde:
LGTM, btw.  We don't yet have tests that simulate a release, so unless we write some really fast, this is going to get its real testing after the release is prepared but before it's uploaded to PyPi.

PR #2249, opened 2016-01-20
Make argparse dependency unconditional.

erikrose:
The primary motivation is to avoid a branch, giving bugs one fewer place to hide. But, as a bonus, more people get a more bugfixed version of argparse. (To use the example from the argparse docs, people stuck on Python 3.2.3 can get bugfixes that made it into the stdlib only in 3.2.4.)


kuba:
We used to run it like that and we had problems AFAIR. If I remember correctly, Arch users had problems. Please refrain from merging until this is well investigated (including finding old threads about the issues).

erikrose:
I don't see any relevant mentions of argparse on community.letsencrypt.org or https://groups.google.com/a/letsencrypt.org/. Any idea where I should search for those threads? I'll try running a test on Arch.

erikrose:
I'm having trouble getting Arch set up; I'll keep trying. Incidentally, it looks like we have Arch packages for LE (and le-auto requires `--debug` to run the old arch bootstrap steps), so, FWIW, only developers even use the declared dependencies.

bmw:
The only thing I could find was #613 which contains no reference to what the problems were.

bmw:
Running `bootstrap/dev/venv.sh` on Arch with this PR works.

erikrose:
Thank, @bmw!

erikrose:
Unless someone knows where [those old threads](https://github.com/letsencrypt/letsencrypt/pull/2249#issuecomment-173676838) are, I move to merge this, for these reasons:

* bmw's installation on Arch works
* LE packages exist for Arch, and only users of `--debug` can even invoke le-auto on Arch.
* Only people manually installing the Python packages with pip (likely LE devs) will even use these dependency declarations.

If I get a +1, I'll merge. If a -1, I'll move it out of the milestone.

pde:
Probably we should ask @kuba if he knows what was going on in #613?  Which means we should bump this to 0.4.0?  Is there a reason to try to ship it quickly?

erikrose:
Nope.

kuba:
The problem was about Arch linux packaging. In fact, there is a comment indicating this: "# only some distros recognize stdlib argparse as already satisfying". Please talk to package maintainers to work this out.

erikrose:
Another point in favor of this: [pip < 8.0 is buggy](https://github.com/pypa/pip/pull/3225) and will cause problems on machines in which LE is installed first on 2.7 and later on 2.6—mostly developers, I would hope, but still a waste of time and a disincentive for contribs. This is the problem we hit in #2343 as well. Bugs love to hide under branches. :-)

Arfrever:
Standard library is in `sys.path` before site-packages directory.
So if I install external argparse, then still:
```
$ python2.7 -c 'import argparse; print(argparse)'
<module 'argparse' from '/usr/lib64/python2.7/argparse.pyc'>
```

With ugly manipulation of `sys.path`:
```
$ python2.7 -c 'import site, sys; sys.path = site.getsitepackages() + sys.path; import argparse; print(argparse)'
<module 'argparse' from '/usr/lib64/python2.7/site-packages/argparse.pyc'>
```

And actually not all fixes from https://hg.python.org/cpython/log/2.7/Lib/argparse.py or https://hg.python.org/cpython/log/default/Lib/argparse.py from before 2015-09-12 (date of release of external argparse 1.4.0) have been backported to external argparse 1.4.0.

bmw:
I'd love to merge this as it's progress on #3455, however, Kuba's memory is correct. On Ubuntu:
```
$ virtualenv --no-site-packages venv
Running virtualenv with interpreter /usr/bin/python2
New python executable in /tmp/tmp.NbmBQDTvMf/venv/bin/python2
Also creating executable in /tmp/tmp.NbmBQDTvMf/venv/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
$ . ./venv/bin/activate
(venv) $ pip install argparse
Requirement already satisfied (use --upgrade to upgrade): argparse in /usr/lib/python2.7
(venv) $
```

On Arch Linux:
```
$ virtualenv2 --no-site-packages venv
New python executable in /tmp/tmp.VsrqAN0pGm/venv/bin/python2
Also creating executable in /tmp/tmp.VsrqAN0pGm/venv/bin/python
Installing setuptools, pip, wheel...done.
$ . ./venv/bin/activate
(venv) $ pip install argparse
Collecting argparse
  Using cached argparse-1.4.0-py2.py3-none-any.whl
Installing collected packages: argparse
Successfully installed argparse-1.4.0
(venv) $
```
Thoughts @erikrose?

erikrose:
So one motivation for this PR, "more people get a more bugfixed version of argparse", is null thanks to @Arfrever's findings.

The other—to avoid branches on principle—still stands except on Ubuntu, presumably. It is somewhat amplified by our installing argparse unconditionally in certbot-auto (where I don't think it's ever caused any problems). However, there will always be plentiful differences in dependency resolution between the loose setup.py declarations and the pinned requirements.txt ones, so that shouldn't be the sole decider. @bmw, what was the problem we were talking about last week that motivated us to take another look at being unconditional? Maybe that can push me over the edge. Absent that, I think we ought to just close this.

bmw:
@erikrose, the discussion I remember in the last week or two about dependency problems was about optional dependencies, which we currently have but will probably remove in the next release (see #3650). I don't think this isn't particularly relevant here.

My main motivation for wanting this PR is #3455 which I linked above. Removing conditional dependencies is the best way I am aware of to fix problems in our wheels. We can easily make our `mock` dependency unconditional, so `argparse` is the only dependency stopping us from removing conditional dependencies entirely. Alternatively, we could start building a separate wheel for Python 2.6, for both the `acme` and `certbot` packages, but I'd like to avoid this if possible.

I'm also not sure what distros are affected. Based on the conversation here, I thought it was an Arch Linux problem. Unless you know the answer here, perhaps I/we should do more testing. If it is just an Arch problem, we could reach out to them and see if they're willing to solve the problem on their end.

erikrose:
Ah, wheel-building. So yes, our chocies are…

1. eliminating conditional dependencies,
2. using [environment markers](https://www.python.org/dev/peps/pep-0496/) (which work only in pip >= 6.0), or
3. building separate wheels for 2.6 and >2.6.

I don't know offhand how various distros handle their Python metadata, so you could test if you like. But I'm not optimistic about fixing this in the short term through the distros, given the slowness with which they move.

With the context you added, my leaning is to depend on argparse unconditionally—to merge this patch. I'm don't think "Requirement already satisfied" qualifies as an "error", nor does the favoring of the stdlib version on certain platforms—since argparse is a stdlib module, it never really breaks API compatibility, and in fact the stdlib version may contain newer fixes, as Arfrever says. To bolster our confidence that none of those fixes is vital to us, we have the evidence of all certbot-auto deployments, which install argparse unconditionally and haven't created a rain of argparse-related bug reports. So I'm for merging. Broken wheels is worse than any effect I can think of.

bmw:
My only concern about depending on `argparse` unconditionally is problems for OS Certbot packages. For example, we saw above that Arch Linux doesn't recognize stdlib `argparse` as satisfying the `argparse` requirement. [Arch also doesn't have an argparse package](https://www.archlinux.org/packages/?q=argparse) available (since it's in the standard library). Will this requirement cause headaches for our package maintainers? Will they have to modify `setup.py` downstream to prevent Certbot erroring out with a message about our `argparse` dependency not being satisfied?

If we're confident this won't happen, I'll happily merge this PR.

erikrose:
> Will they have to modify setup.py downstream to prevent Certbot erroring out with a message about our argparse dependency not being satisfied?

It depends on whether their distro's Python packaging tooling automatically emits a dependency on a nonexistent argparse distro package. If it does, then yes, they'll have to tell it not to. At least in the case of debs, that's not a big deal and can be done one time (by including a patchfile against setup.py) and apply automatically thereafter.

bmw:
@hlieberman, @hogarthj, @felixonmars, do you know how much of a headache this will be for you downstream? Basically, we want to unconditionally depend on `argparse` in `setup.py` for both the `acme` and `certbot` modules as it allows us to build a universal wheel that should work on Python 2.6+. There is an [argparse package](https://pypi.python.org/pypi/argparse) available on PyPI, however, most distros don't have it packaged because `argparse` is available in the Python standard library starting with Python 2.7. If we do this, will you have to patch `setup.py` on your end and how much of a pain will it be to maintain this patch as our `setup.py` files change over time?

bmw:
After talking to a Fedora package maintainer, it we may be okay there. The Python package in all RPM based distros where we're packaged provides `python-argparse`.

There are two potential concerns I can think of here:

1. The [Python3 package](http://koji.fedoraproject.org/koji/rpminfo?rpmID=8081794) does not do that and we're almost Python 3 compatible. If they start shipping a Python 3 version of Certbot, this may be a problem.
2. What run time checks does Python do to verify the necessary dependencies are available? I'm not sure if this is sufficient testing, but after adding `argparse` to `/usr/lib/python2.7/site-packages/certbot-0.8.1-py2.7.egg-info/requires.txt`, Certbot still works with no issues.

felixonmars:
It would be okay to patch setup.py downstream here for Arch Linux. I currently do that to other packages like obfsproxy, too. A simple `sed` would be enough :)

For the second question, the runtime check done by pkg_resources will check that `requires.txt` and give errors when it's not satisfied. Since the metadata is generated by setuptools, patching setup.py is enough to avoid argparse in that file. The Arch Linux packaging problem mentioned above should have been fixed quite long ago, and it's definitely a bug in packaging.

Arfrever:
`sed` for `setup.py` would be enough, although [pip 6.0](https://pypi.python.org/pypi/pip/6.0) is almost 2 years old and there was at least [one](https://www.cve.mitre.org/cgi-bin/cvename.cgi?name=2015-2296) security fix since pip <6.0 (in 6.1.0) ([Changelog](https://pip.pypa.io/en/stable/news/)), so nobody sane should use old pip.

erikrose:
Within the last year or so, one of the top clients in PyPI's web logs was pip 1.5.4, since Debian was shipping it.

jsha:
So it sounds like the consensus is: Not really a problem for Arch anymore, so go ahead and merge. Is that accurate?

@erikrose, could you resolve the conflicts?

erikrose:
Yep.

erikrose:
Tried out GitHub's built-in conflict resolution UI. Not bad. Crappy, unchangeable commit messages, though.

bmw:
I used GH's conflict resolution UI for the first time recently to resolve merge conflicts on a PR from a fork I didn't have write access to. Nice that the tool exists, but I wish there was a way to do the same thing using `git` locally. I like my editor better.

erikrose:
> Nice that the tool exists, but I wish there was a way to do the same thing using gitlocally. I like my editor better.

Try "git mergetool". And then I like to put this in my .gitconfig (Mac):

```
[merge]
	tool = opendiff
```

That way, you get to use FileMerge, which is still my favorite diff-er after 25 years (it's from NeXTStep).

osirisinferi:
@erikrose This merge broke `certbot` on Gentoo...

hogarthj:
@bmw just went to do the 0.13.0 builds on Fedora/EPEL7 and saw this had been merged.

Although rpm/yum has the provides in place /usr/bin/python2 setup.py test did not recognise argparse as present and tried to obtain it:

https://kojipkgs.fedoraproject.org//work/tasks/6209/19086209/build.log

I'm going to patch it out of our build - but this will probably have an affect on anyone using virtualenv rather than my packages.

bmw:
OK. Thanks for the update.

I talked about this change with @NoodlesNZ in IRC a long time ago before it was merged and left the conversation under the impression this wouldn't be a problem in Fedora/EPEL because the `python` package provides `python-argparse`. This was likely due to me misunderstanding or poorly explaining what we were doing. Sorry for the trouble.

If users are running `pip install` in a virtualenv, they should be OK because there is an `argparse` package on [PyPI](https://pypi.python.org/pypi/argparse/1.4.0). If you see a problem here that I don't, please let me know. Additionally, I'm planning on reverting this PR for our next release assuming I can get everyone on board so this should only be a problem for this release ([source](https://github.com/certbot/certbot/pull/4379#issuecomment-294939992)).

PR #2250, opened 2016-01-20
[test_leauto_upgrades] actually use everything from v0.1.0 for setup

pde:
This would catch issues like #2243 in test farm testing.


bmw:
LGTM. Free to merge when green.

PR #2251, opened 2016-01-21
Do not say we've renewed a cert if it was reinstalled

pde:


schoen:
The

`"Your existing certificate has been successfully {3}ed, and the "`

reminds me: what's our internationalization plan for the client? :-)

schoen:
I filed a new issue (#2261) on the ability to internationalize the client.

pde:
I agree this will need to be tweaked for internationalisation, though I feel that's a pretty distant wishlist item at the moment.  Want to +1 the PR or ask for revisions?

schoen:
LGTM.

Conceptually, the criterion

`    if len(lineage.available_versions("cert")) == 1:`

which this code relies on as part of deciding what to tell the user (but doesn't introduce) might not be the best one because it's looking at the end result in the filesystem rather than at what we just attempted to do. As a contrived example, if some other code (external to the client) deleted the old cert immediately after the new cert was installed, the lineage would contain only 1 available version and the resulting notification would be for a fresh installation rather than for a renewal. (Again, that code was already there and isn't introduced by this PR.)

bmw:
LGTM. Merging. Perhaps we should open an issue for the fragility `len(lineage.available_versions("cert")) == 1`, assuming we have a better way to do it?

PR #2255, opened 2016-01-21
fix "grep: /etc/os-release: No such file or directory"

aaronzirbes:
Removes the error message `grep: /etc/os-release: No such file or directory`
when running `./letsencrypt-auto --help`


aaronzirbes:
Previous output:
```
$ ./letsencrypt-auto --help
grep: /etc/os-release: No such file or directory
WARNING: Mac OS X support is very experimental at present...
```

pde:
LGTM, but will need another +1

erikrose:
LGTM too as far as it goes, though we will need to edit the template, too, of course. I'll take care of it.

PR #2257, opened 2016-01-22
Upgrade peep to 3.0.

erikrose:
This will avoid crashing when used with pip 8.x, which was released today and is already the [3rd most used client](https://caremad.io/s/kEiD84jkTz/) against PyPI. (7.1.2 and 1.5.4 take spots 1 and 2, respectively. They're all about 2M hits a day.)


erikrose:
Thought about leaving this till 0.4.0, but it's a very quick review, it'll save us tech support, and we don't have to worry about it breaking le-auto since it's not replacing the old script yet.

pde:
LGTM, though I'm not a peep expert.  If you think it's safe to merge, go ahead -- this isn't instantly exposed to mainstream users, anyway.

Issue #543, opened 2015-06-24
Nginx Restart Failed!


N00D13:
every Time i do `letsencrypt -d www.domain.de run`
i got                            Nginx Restart Failed!                                                │  
                              │ nginx: [emerg] the size 1048576 of shared memory zone "SSL"          │  
                              │ conflicts with already declared size 5242880 in                      │  
                              │ /etc/letsencrypt/options-ssl-nginx.conf:1

PatrickHeppler:
Check your nginx config for already defined ssl_session_cache

pde:
@diracdeltas want to take this?

diracdeltas:
@N00D13 to confirm: do you have ssl_session_cache defined in your nginx config file somewhere? Quick fix is to delete that line, run letsencrypt, then manually edit options-ssl-nginx.conf with your original ssl_session_cache.

I can work on a real fix but it might take a week or so (busy with work).

PatrickHeppler:
Deleting the line i a workaround but not a fix ;)

Sent from Nine

Von: Yan Zhu <notifications@github.com>
Gesendet: 30.06.2015 19:32
An: letsencrypt/letsencrypt
Cc: PatrickHeppler
Betreff: Re: [letsencrypt] Nginx Restart Failed! (#543) 

@N00D13 to confirm: do you have ssl_session_cache defined in your nginx 
config file somewhere? Quick fix is to delete that line, run letsencrypt, 
then manually edit options-ssl-nginx.conf with your original 
ssl_session_cache.

I can work on a real fix but it might take a week or so (busy with work).

---

Japh:
I came up on this error too, so I grepped the nginx directory to find where the other entry was and changed my letsencrypt config so the times matched.

That is, in my letsencrypt version I had 
```
    ssl_session_cache shared:SSL:50m;
```

But a line in another config had `20m`, so I changed it to `20m` and all good!

SwartzCr:
I'm going to close this - since it's probably been fixed by the nginx work that @ohemorange has been doing. Feel free to re-open it if I'm mistaken.

ohemorange:
This was actually fixed by #2262 and #3383, but yes, it's fixed. We only modify non-ssl server blocks, and have a specific fix for the `ssl_session_cache` issue.

Issue #1743, opened 2015-12-04
nginx adapter not including options-ssl-nginx.conf with existing HTTPS server


aprescott:
Found this while using letsencrypt 0.1.0.

I have `/etc/nginx/nginx.conf` which has these sections in its `http {}` (among other configuration):

```
ssl_prefer_server_ciphers   on;
ssl_ciphers  DEFAULT:!aNULL:!eNULL:!ADH:!RC4:!RC2:!DES:+HIGH:+MEDIUM:!LOW:!SSLv2:!EXP:!MD5:!DSS:!PSK;
ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
```

(Note this existing configuration is a default in a `http {}` block in `/etc/nginx/nginx.conf`, not specific to any individual `server {}`.)

It then uses

```
include sites-enabled/*.conf;
```

to pull in `/etc/nginx/sites-enabled/example.com.conf`, which defines an existing SSL-enabled server:

```
server {
  server_name www.example.com;
  rewrite ^ https://example.com$request_uri? permanent;
}

server {
  # ...

  listen 443 ssl spdy;
  server_name example.com;

  ssl                  on;
  ssl_certificate      ssl/foo-cert.pem;
  ssl_certificate_key  ssl/foo-key.key;

  ssl_session_cache    shared:SSL:5m;
  ssl_session_timeout  5m;
  # ...
}
```

After accounting for #921 (which causes problems because of the `ssl_session_cache`) I found that `letsencrypt --nginx -d example.com` will correctly update the `example.com.conf` file to have the correct `ssl_certificate` and `ssl_certificate_key` values. It _doesn't_, however, seem to include `options-ssl-nginx.conf` anywhere. (Before touching anything to do with `letsencrypt`, I did a `git init .` in my `/etc/nginx` directory, which has let me `git diff` on the file changes.)

I'm not sure what I was expecting, though. I guess I thought that the `server {}` block in `example.com.conf` would get the `include options-ssl-nginx.conf`. Does that override the SSL configuration for _only_ `https://example.com`, and leave the `/etc/nginx/nginx.conf` default? Is that actually expected because of `-d example.com`?

olabini:
The options-ssl-nginx.conf file is only used when turning a vhost from a non-https to an https based vhost, it will never be added when changing an existing SSL configuration.
I'm currently working on a patch that will not use the conf-file at all, because of things like #921

olabini:
In my opinion, this ticket can be closed

[...]

Issue #921, opened 2015-10-08
Nginx fails if ssl_session_cache already defined


jsha:
Steps to reproduce:
1. Put `ssl_session_cache shared:SSL:10m;` in your Nginx configs somewhere.
2. Run letsencrypt with the nginx authenticator.

Expected results: works

Actual results:

`2015-10-08 00:30:17,856:ERROR:letsencrypt_nginx.configurator:Nginx Restart Failed!

nginx: [emerg] the size 1048576 of shared memory zone "SSL" conflicts with already declared size 10485760 in /etc/letsencrypt/options-ssl-nginx.conf:1`

Contents of /etc/letsencrypt/options-ssl-nginx.conf:

```
ssl_session_cache shared:SSL:1m;
ssl_session_timeout 1440m;

ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
ssl_prefer_server_ciphers on;

# Using list of ciphers from "Bulletproof SSL and TLS"
ssl_ciphers "ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-ECDSA-AES128-SHA ECDHE-ECDSA-AES256-SHA ECDHE-ECDSA-AES128-SHA256 ECDHE-ECDSA-AES256-SHA384 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-RSA-AES128-SHA ECDHE-RSA-AES128-SHA256 ECDHE-RSA-AES256-SHA384 DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384 DHE-RSA-AES128-SHA DHE-RSA-AES256-SHA DHE-RSA-AES128-SHA256 DHE-RSA-AES256-SHA256 EDH-RSA-DES-CBC3-SHA";
```

jsha:
Correction: It looks like the `ssl_session_cache` directive is placed in both /etc/nginx/conf.d/ssl.conf and  /etc/letsencrypt/options-ssl-nginx.conf by letsencrypt, which causes the conflict. I noticed this because I tried editing the line out of ssl.conf and re-running, but found that ssl.conf was restored to its previous value.

jsha:
[...]

PR #2262, opened 2016-01-22
No conflicting declarations

### LINKED ISSUE #574: Option to exclude options-ssl-nginx.conf
If one has already setup SSL the letsencrypt options-ssl-nginx.conf will prevent nginx restart.
In my case it's the ssl_session_cache which is already defined in my config.

olabini:
This pull request does two things - it stops using the options-ssl-nginx.conf and instead reads it and adds the declarations directly. The reason for doing this is to stop potential conflicting declarations, and make it easier to deal with other requirements, such as providing the possibility of upgrading cipher suites.

It also changes the name of the session cache used. That means that even if an ssl_session_cache is defined at the http{} level, this cache will not conflict with it.

This should fix #574, #921 and #1743

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,6 +21,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       # Cannot allocate memory" when running
       # letsencrypt.client.tests.display.util_test.NcursesDisplayTest
       v.memory = 1024
+
+      # Handle cases when the host is behind a private network by making the 
+      # NAT engine use the host's resolver mechanisms to handle DNS requests.
+      v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
     end
   end
 
--- a/bootstrap/_rpm_common.sh
+++ b/bootstrap/_rpm_common.sh
@@ -3,6 +3,7 @@
 # Tested with:
 #   - Fedora 22, 23 (x64)
 #   - Centos 7 (x64: on DigitalOcean droplet)
+#   - CentOS 7 Minimal install in a Hyper-V VM
 
 if type dnf 2>/dev/null
 then
@@ -21,12 +22,16 @@ fi
 if ! $tool install -y \
        python \
        python-devel \
-       python-virtualenv
+       python-virtualenv \
+       python-tools \
+       python-pip
 then
   if ! $tool install -y \
          python27 \
          python27-devel \
-         python27-virtualenv
+         python27-virtualenv \
+         python27-tools \
+         python27-pip
   then
     echo "Could not install Python dependencies. Aborting bootstrap!"
     exit 1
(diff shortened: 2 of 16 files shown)


pde:
Interesting approach.  This is rather different from what we do in Apache, but it might be workable.  Some questions:

* Do you defer to existing directives, override them, or offer a switch between the two?
* If you defer, there's probably no elegant way for us to update the SSL options we've set, which seems bad.
* Overriding is a better default, though we should think about how to let the user know, and give them an option to cancel (I think we can improve the current Easy / Secure UI step to achieve that).

Related: #1123. CC'ing @joohoi and @sagi for their thoughts.

pde:
The other option to consider is commenting out existing conflicting declarations, but continuing to keep our options in a separate file, so that people can see where they came from and we can update the file under appropriate circumstances.

olabini:
Yeah, the main reason I went with this approach was because it felt like if we want support two modes of ssl_ciphers, one "low" and one "high" - dealing with having them all in a shared include file seemed like a bad idea.

olabini:
Incidentally, @pde - have you considered doing something along the lines of marking each line that LE is responsible for with a comment of some kind? Something like

    ssl_prefer_server_ciphers on; #LE-line - please don't remove this comment

olabini:
In terms of your questions:
    Do you defer to existing directives, override them, or offer a switch between the two?
Currently, the code uses the existing method of adding directives, which uses add_server_directives with replace=False - thus, right now we defer to existing ones.

    Overriding is a better default, though we should think about how to let the user know, and give them an option to cancel (I think we can improve the current Easy / Secure UI step to achieve that).
Sure, that's an easy change if that's what we want to do.

pde:
OK, here are a few approaches that I think make sense:

* Keep "existing directives take precedence" as the default behaviour, but add a comment like `# Set by LetsEncryptPythonClient" to each directive that we add, so they can be updated later.
* Switch to "new directives take precedence", and then add a UI option (we can do that in another PR) to disable that if the user wants to.  Directive comments are still nice in this case, but less necessary.
* Continue putting all of our security directives in our own separate file, and deal with the issue by commenting out any directives that conflict with ours.  If we want to support high-security and high-compatibility ciphersuite options, we can make two different config files, and include the one that the user requested.

I think any of these paths are workable, and I'm willing to merge whichever one the nginx team thinks is most elegant.

pde:
@diracdeltas is mostly swamped with other work at the moment but might also have thoughts on this.

olabini:
I think all three of these are workable. The final one I think might be trickier, since then we have to keep track of what we commented out, etc.

For now, I'll update this PR with approach number one.

pde:
@olabini why would we need to keep track of what we commented out in case 3?  (I guess we could need to do that for reverting changes, but I assume in that case we just keep an old copy of the entire configuration).

olabini:
True, that might be easier.

olabini:
This one is now waiting for the new parser.
This one also fixes #543 and #939, once the new parser is in, and I can add comments.

PR #1824, opened 2015-12-08
CentOS add check for python tools and pip

sorvani:
fixes #1823
Add check for python-tools and python-pip

PR #2266, opened 2016-01-22
Add installation of python-tools and python-pip for CentOS

erikrose:
This replaces #1824, adding the necessary porting now that #1665 has landed.


pde:
Passes testfarm `test_letsencrypt_auto_certonly_standalone.sh`.

pde:
And the old (not entirely accurate) `test_leauto_upgrades.sh`.  The new one is still broken...

pde:
LGTM.  Merge with a second +1!

PR #2280, opened 2016-01-26
Check for the existence of symlinks

TheNavigat:
Addresses https://github.com/letsencrypt/letsencrypt/issues/2000

Uses code from here, https://github.com/cowlicks/letsencrypt/blob/gh-2000/letsencrypt/storage.py#L131-L139. Credits go to @cowlicks. Thanks for the snippet!


cowlicks:
Thanks for wrapping this up @TheNavigat 

Just a heads up, in the future you might want to use commits from other people's branches rather than copy/pasting the code. I don't mind, so feel free to keep this as is. But other people in the same situation might complain about attribution.

TheNavigat:
I apologize, @cowlicks, your commit is correctly added :)

SwartzCr:
Before this commit LE errors out with the print of `could not find cert file`
Now we get errors saying the conf in `/etc/letsencrypt/renewer/blah.conf` is broken. Skipping.
and then we move on to renewal. This seems fine to me - though obviously it doesn't fix the existing broken symlinks (unless the new one is deployed)

SwartzCr:
That said this fixes the issue in #1959, so I'm +1 to merge

Issue #2279, opened 2016-01-26
Centos httpd doesn't support SSLCompression directive


SwartzCr:
it's served in /etc/letsencrypt/options-ssl-apache.conf on line 9, but centos chokes on it. It'd be nice (if it's not essential) to remove it from the conf file on old CentOS systems

PR #2284, opened 2016-01-26
make different options ssl conf for centos

SwartzCr:
A proposed fix for #2279 
Could also fix it using augeas, but that'd require checking OS again in a different place


pde:
@joohoi, does this look like the right approach to you?

joohoi:
Yes, LGTM. The use of constants is the best way to go, rather than hiding the OS differentiation inside the actual code. It's going to make addition of different configuration files for other distributions / OSes pretty easy.

Issue #432, opened 2015-05-19
nginxparser from PyPI instead of vendoring


kuba:
@diracdeltas, do you think we could work with the upstream (I guess that's https://github.com/fatiherikli/nginxparser) to contribute back our changes and make sure `nginxparser` is released to PyPI (not there yet: https://pypi.python.org/pypi?%3Aaction=search&term=Nginxparser&submit=search), rather than embedding the code directly within our source tree? I'm happy to help (with the packaging).

diracdeltas:
That would probably be good, though the upstream maintainer is not super responsive. (Ex: I don't think he's addressed https://github.com/fatiherikli/nginxparser/pull/14#issuecomment-86639221 so tests probably aren't passing.)

OTOH the code is not very complex, and we've probably used it more extensively than anyone else at this point.

GitHubGeek:
Could be done with a fork of `nginxparser` and use the `git+` syntax, e.g.:

```
# requirements.txt
# ...
git+https://github.com/letsencrypt/nginxparser.git
```

pde:
For the time being I think we're heading off into fork-land for Nginxparser.  Perhaps we should try to preserve APIs and upstream our changes, but it's not a top priority ATM.

Issue #938, opened 2015-10-10
Nginx configurator should preserve whitespace on output


jsha:
When the configurator edits a config file, it normalizes all the whitespace in the file it outputs. This is confusing if you are used to editing your configs by hand, because it changes the visual appearance of the file. When I first ran the configurator, I was convinced that letsencrypt had completely overwritten my nginx.conf.

aprescott:
Having just encountered this, I'm mostly okay with the normalized indentation whitespace, since it neatened up my inconsistent config files. My only real complaint is that blank lines were all removed, which made things harder to read. For instance, groupings like

```ruby
x 1;
y 2;
z 3;

m 10;
n 20;

foo 1000;
bar 2000;
```

were all "ungrouped" since the blank lines were deleted.

olabini:
This is being worked on by @iapazmino

olabini:
The linked PR fixes this issue, so once merged we can close this ticket.

pde:
@sagi is going to take a closer look at pyparsing to figure out whether there's any way to persuade/modify it to do this for us

Issue #2106, opened 2016-01-07
Required pyparsing version


bmw:
I've been experimenting with writing tests using the oldest allowed versions of our Python dependencies. `setup.py` for `letsencrypt-nginx` says it requires `pyparsing>=1.5.5` but when I pin version 1.5.5, I encounter problems. You can see Travis logs of the issue [here](https://travis-ci.org/letsencrypt/letsencrypt/jobs/100739657) and [here](https://travis-ci.org/letsencrypt/letsencrypt/jobs/100739658).

We should determine what version we require and update `setup.py` accordingly.

PR #2285, opened 2016-01-26
New nginx parser

olabini:
This PR implements a new parser for the nginx format - it's slightly looser, but has the capability to retain more information. The current PR also retains newlines, thus fixing #938 - it also allows for retaining comments on the same line as directives. This is something we want, in order to mark the directives we have added.


olabini:
This PR can also allow us to fix #432 and #2106

letsencrypt-nginx/letsencrypt_nginx/parser.py
```diff
@@ -55,12 +54,14 @@ def _parse_recursively(self, filepath):
                 if _is_include_directive(entry):
                     # Parse the top-level included file
                     self._parse_recursively(entry[1])
-                elif entry[0] == ['http'] or entry[0] == ['server']:
+                elif len(entry) > 0 and (entry[0] == ['http'] or entry[0] == ['server']):
```
erikrose: We can say just `entry` rather than `len(entry) > 0`.

letsencrypt-nginx/letsencrypt_nginx/tls_sni_01.py
```diff
@@ -99,11 +99,14 @@ def _mod_config(self, ll_addrs):
...
+        for block in main:
+            key = block[0] if len(block) > 0 else None
+            body = block[1] if len(block) > 1 else None
+
             if key == ['http']:
                 found_bucket = False
-                for k, _ in body:
-                    if k == bucket_directive[0]:
+                for b in body:
+                    if len(b) > 0 and b[0] == bucket_directive[0]:
```
erikrose: Again (and above in the same file) the idiomatic `b` can be substituted for `len(b) > 0`.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
-    Optional, OneOrMore, Regex, ZeroOrMore)
-from pyparsing import stringEnd
-from pyparsing import restOfLine
+import StringIO
 
+class ParseException(Exception):
+    """A parsing exception happened."""
+    def __init__(self, value):
+        self.value = value
+        super(ParseException, self).__init__()
```
erikrose: There's no sense in calling the superclass's constructor if you're not passing anything to it. All it does is set `self.args`. I wouldn't bother making this call at all, since you clearly mean to use the `value` attr instead.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+        return self.parse()
 
+    def parse(self):
+        """Returns the parsed tree."""
+        return RawNginxParser.Internal(self.source).parse_file()
+
+    class Internal(object):
+        """An internal class to keep track of parsing state."""
+        def __init__(self, source):
+            self.src = StringIO.StringIO(source)
```
erikrose: Is there a reason to bend the logic into file IO terms? It would be simpler to keep a position counter. Then we wouldn't need StringIO, and _peek() and _read() could be expressed more simply:

```python
def _peek(self):
    try:
        return self.src[self.pos]
    except IndexError:
        self.eof = True
        return None


def _read(self):
    ret = self._peek()
    self.pos += 1
    return ret
```

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
     def as_list(self):
         """Returns the parsed tree as a list."""
-        return self.parse().asList()
+        return self.parse()
 
+    def parse(self):
+        """Returns the parsed tree."""
+        return RawNginxParser.Internal(self.source).parse_file()
+
+    class Internal(object):
```
erikrose: Why split this off as a separate class? It's actually increasing the amount of state, since both the external and the internal class need to hang onto `source`/`src`. There's no additional state in the outer class.

In fact, the outer class should probably go away altogther. (A good rule of thumb is that a class with only a single public method should just be a function. RawNginxParser has 2, but one just calls the other. (Why?)) Why force callers to construct a class and call a method on it when there's nothing else useful you can do with the class? We could simplify the API to something like…

```python
foo = parse_nginx('some config code')
```

If you want to design for subclassing, that could even be a class method.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+
+        def _parse_block(self):
+            result = []
+            while True:
+                res = self._parse_next()
+                if res is None:
+                    return result
+                result.append(res)
+
+        def _is_whitespace(self, val, includeNL):
```
erikrose: Let's be consistent with `_read_newline` and `_is_at_newline` here and call this `include_newline`.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+            if self._is_whitespace(self._peek(), includeNL):
+                return self._read()
+            return None
+
+        def _is_at_newline(self):
+            return self._peek() in ['\n', '\r']
+
+        def _read_newline(self):
+            if self._read() == '\r' and self._peek() == '\n':
+                self._expect('\n')
```
erikrose: We just checked that _peep() == '\n', so what is the point of calling _expect() over simply _read()?

letsencrypt-nginx/letsencrypt_nginx/parser.py
```diff
@@ -55,12 +54,13 @@ def _parse_recursively(self, filepath):
                 if _is_include_directive(entry):
                     # Parse the top-level included file
                     self._parse_recursively(entry[1])
-                elif entry[0] == ['http'] or entry[0] == ['server']:
+                elif entry and (entry[0] == ['http'] or entry[0] == ['server']):
```
erikrose: `entry and entry[0] in ['http', 'server']` might be a little more readable.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+            result = [key]
+            if key == 'location':
+                mod = self._parse_modifier()
+                if mod is not None:
+                    result.append(mod)
+            val = self._parse_value()
+            nx = self._peek()
+
+            if nx == ';':
+                return self._finish_assignment(result, val)
```
erikrose: It seems to me that _parse_assignment_or_block() returns a list whose elements' purposes are distinguished not only by their placement but by the length of the list itself. For example, element 2 could be a modifier or something else, and there's no good way to tell which. (Correct me if I'm missing something.)

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+            while not self._is_at_newline():
+                result = result + self._read()
+
+            self._read_newline()
+
+            return result
+
+        def _parse_comment(self):
+            if self._peek() == '#':
+                self._expect('#')
```
erikrose: Here's that strange peek-then-expect pattern again (where the expect can't possibly fail and so its param is pointless). Since it's becoming a pattern, perhaps we should factor it out into an `if self._next_is('#')` which consumes the next char iff it's as specified.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+            if self._is_at_newline():
+                self._read_newline()
+                return []
+            return None
+
+        def _collect_until_end_of_line(self):
+            result = ''
+
+            while not self._is_at_newline():
+                result = result + self._read()
```
erikrose: We try to avoid string concatenation in tight loops, as it's got quadratic performance in some Python implementations. Better is appending to an array and `join`ing.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+            while True:
+                val = self._peek()
+                if val is None or not self._is_key_char(val):
+                    if result == '':
+                        return None
+                    return result
+                result = result + self._read()
+
+        def _parse_string(self):
+            startVal = self._read()
```
erikrose: startVal uses different case conventions than the rest of the symbols.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+                    raise ParseException("Invalid configuration file, unfinished string literal")
+                if val == startVal:
+                    return result
+
+        def _parse_value(self):
+            self._pass_whitespace(True)
+            result = ''
+            while True:
+                val = self._peek()
+                if val is None or val in ['{', ';']:
```
erikrose: You could just say `val in '{;'`, since a string is a sequence of chars. (And it'd likely get interned, saving a list construction, but that's probably academic.)

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
erikrose: Yes, it looks like this is the case. This leads to a lot of awkward length tests in the callers; representing the emitted tokens as attributes of a class or keys of a dict would make for shorter, more readable callsites.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
```diff
@@ -1,56 +1,241 @@
...
+                    result = result.rstrip()
```
erikrose: Note to self: continue reviewing here.

erikrose:
olabini is to be commended for attacking this problem, but I don't think this approach is t a good tradeoff: it ends up harder to understand and, due to sheer length, more bug-prone than the PyParsing-based method. It's been awhile since I've used PyParsing, but I'm fairly certain it can be persuaded to retain comments and whitespace. Have you looked at its leaveWhitespace() handler? If it truly can't be bent to this purpose, I think Parsimonious is the way to go: it'll give us clarity, attribute-based access, and probably better speed as a bonus. It's pure Python and has no subdependencies.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
olabini: Actually, I only put it in since your lint configuration errors if it isn't there. I'm fine with leaving it out though.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
olabini: Actually, it's a separate class because I wanted to keep the parser immutable. The previous parser was also immutable, and I couldn't really foresee the consequences of changing that invariant. 

I completely disagree with your rule of thumb, btw. I find extracting a method into a separate class one of the more useful refactorings to make, especially when it's manipulating a lot of mutable data.

Now, the Internal class is only used internally, it's not for external consumption, and if you feel making the overall parser stateful and one-shot, I can definitely remove it.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
olabini: Yeah, this is probably a good idea. I did it initially because I assumed the parser would want to work on file objects - what happens if we encounter a larger configuration file? But maybe just using the string directly is fine.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
olabini: Sorry, old habit from writing many parsers. Using _expect() in this way is a bit of a tradition. It's not strictly necessary in this case, since we know it will always return the same thing.

letsencrypt-nginx/letsencrypt_nginx/nginxparser.py
olabini: Yes, I noted this, but I didn't want to change it in this PR since some of that awkwardness was there from before (several type-checks that actually differentiate different directive types). I would be fine with fixing it, but it probably fits better as a self contained refactoring.

olabini:
Before even embarking on doing this, me and another person spent half a day trying to convince PyParsing to do the right thing, including leaveWhitespace() and several of the other hooks. We didn't succeed, and the inbetween we had was brittle and failed in unpredictable ways. I can take a timeboxed stab at Parsimonious tomorrow, but then I'll really have to give up on this.

olabini:
Correct me if I'm wrong, but also, doesn't PEG parsers usually have exponential memory usage?

olabini:
Correct me if I'm wrong, but also, doesn't PEG parsers usually have exponential memory usage?

olabini:
Incidentally, I have also moved to using 'in' on a string instead of an array, but as you can see that causes its own ugliness since you can't say None in "abc".

olabini:
OK, this is how far I got with parsimonious. This parser covers maybe half the logic necessary - it doesn't have anything for values, including string values. It also doesn't do anything about comments, especially same line comments. And most importantly, it doesn't contain the code for actually turning the output of this into something that is useful for the rest of the package. The logic for that would have to contain enough info to do some parsing on its own in order to figure out which whitespace should be retained and which shouldn't. My evaluation is that this is not a good idea, and it's not worth the time finishing it. Especially not since with what @pde  mentioned yesterday, that Parsimonious isn't packaged for Debian yet.

`
    space            = " " / "\t"
    spaces           = space*
    newline          = "\n" "\r"?
    spaceOrNewline   = space / newline
    spacesOrNewlines = spaceOrNewline*

    values           = ~".*"
    block            = "{" spacesOrNewlines section* spacesOrNewlines "}"
    modifier         = "=" / "~*" / "~" / "^~"
    location         = "location" spaces modifier? spaces values spacesOrNewlines block
    simpleDecl       = key spacesOrNewlines values spacesOrNewlines ";" spacesOrNewlines

    section          = spaces (newline / location / simple)
    sections         = section+
`

erikrose:
> doesn't PEG parsers usually have exponential memory usage?

They take O(grammar size * text length) memory.

erikrose:
@olabini Maybe we could set aside some time to chat about this. I'd like to have a better grip on what you're trying to accomplish, and then we can come up with the best way to proceed. Do you have an hour or so next week?

olabini:
I'm currently traveling, so doing an hour long meeting will be hard. Basically, what I want to achieve is this: retain newlines in the places where the original had newlines, so we can output a config file that looks structurally similar to the input, and second, to retain comments on the same line as a directive or statement, so we can output that as well.

erikrose:
Which whitespacedo you *not* care about? (For example, whitespace between the end of a line and its comment?)

olabini:
All of that is encoded in the new parser. Fundamentally, the whitespace we mostly care about is newlines between declarations.

pde:
With the Certbot name transition complete, and many other parts of the client looking stable, the core team is going to be turning its attention to Nginx next.  Which means sorting out this parser question is probably our top priority.

For my research, I've started by [looking at how Nginx does its own parsing](https://github.com/nginx/nginx/blob/master/src/core/ngx_conf_file.c#L101). It uses a hand-rolled explicit state machine, which is going to make verifying the grammar hard.  But it's something of an argument for landing @olabini 's parser, since we could potentially walk through the state machine code and ensure the two match?

Some other obvious options are (1) complicate our pyparsing to explicitly parse and prserve the whitespace as a part of the grammar; (2) switch to augeas.

bmw:
After diving into this a bit myself, I am most optimistic about the `pyparsing` approach. I think we should take another stab at trying to get the parser to preserve whitespace and reevaluate afterwards.

pde:
Closing this in favour of the pyparsing approach via #3184

Issue #1991, opened 2015-12-22
When installing a cert for a system that only has a *:80 model vhost, add ServerName / ServerAlias to the new vhost


pde:
The failure to do this is one of the major reasons we hit the "please tell me which vhost to install on" dialog that underlies #1944 and other issues.

SwartzCr:
Fix should be live in #2286 with tests passing, probably still needs some farm testing or personal vm testing.
No new tests were written for it, which I'd be happy to do before a merge

bmw:
Fixed in #2286.

PR #2286, opened 2016-01-26
write servername/serveralias if copying a generic vhost

SwartzCr:
added code to input the server name to a file if it doesn't have one.
two parts of this could use outside opinions:
1) using _find_best_vhost to determine if it's a generic vhost, we could also use augeas in make_vhost_ssl
2) passing these two new flags to make_vhost_ssl, I'm a bit worried about flag inflation

corresponds to #1991

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -298,12 +298,12 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             return self.assoc[target_name]
 
         # Try to find a reasonable vhost
-        vhost, is_generic_host = self._find_best_vhost(target_name)
+        vhost = self._find_best_vhost(target_name)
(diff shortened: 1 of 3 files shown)


letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -859,6 +863,10 @@ def _add_dummy_ssl_directives(self, vh_path):
                             "insert_key_file_path")
         self.parser.add_dir(vh_path, "Include", self.mod_ssl_conf)
 
+    def _add_servername(self, servername, vh_path):
+        self.parser.add_dir(vh_path, "ServerName", servername)
```
pde: I haven't followed this whole patch yet, but instinctively what I'm expecting around here is "if the vhost doesn't have a ServerName, add one with the name `example.com`; if it already has a ServerName, add a ServerAlias with the name `example.com`".

bmw:
LGTM

PR #2290, opened 2016-01-27
Allow trailing period in fqdn as per #2065

hexaclock:
Updated `check_domain_sanity()` regex to allow for trailing period.

Fixes: https://github.com/letsencrypt/letsencrypt/issues/2065


pde:
As noted [here](https://github.com/letsencrypt/letsencrypt/pull/2213#issuecomment-172691600) we don't actually want to have certs with trailing dots.  However we are going to accept domains with trailing dots and strip them in the client when #2309 lands.

Issue #2200, opened 2016-01-15
Update pinnings and hashes of LE packages for le-auto


erikrose:
We'll need to update the pinnings and hashes of the LE packages to point to the 0.3.0 values before release if we want to publicly test le-auto. Ultimately, this should become part of the release script.

erikrose:
Make sure you do an install under 2.6 as well and catch those conditional dependencies. (ConfigArgParse has some, even though we expunged our direct conditional dependencies.)

erikrose:
Done in the above and cf218dd7f15d03414e6b850729492884c4676a28.

PR #2296, opened 2016-01-27
Add ordereddict, a conditional dependency of ConfigArgParse under Python 2.6. Ref #2200.

erikrose:
It doesn't hurt under 2.7.


bmw:
LGTM

PR #2309, opened 2016-01-29
Set webroot-map from diverse flags, user interaction, and config files correctly

### LINKED ISSUE #1668: Several code paths (conf files, interactive domains, --csr) do not set webroot_map entries correctly

pde:
Closes #1668


pde:
I've incorporated @filipochnik's PR #2241 into this one, since it touches many of the same webroot-map processing mechanisms.

letsencrypt/cli.py
```diff
@@ -1282,19 +1294,36 @@ def __call__(self, parser, config, webroot, option_string=None):
...
+            # unless it was explicitly included in webroot_map
+            if webroot_path:
+                config.webroot_map.setdefault(domain, webroot_path[-1])
+
+
+class WebrootMapProcessor(argparse.Action): # pylint: disable=missing-docstring
+    def __call__(self, parser, config, webroot_map_arg, option_string=None):
+        webroot_map = json.loads(webroot_map_arg)
+        for domains, webroot_path in webroot_map.iteritems():
+            _process_domain(config, str(domains), [str(webroot_path)])
```
filipochnik: Right now something like `-d żąćł.com` results in a nice error "Internationalized domain names are not presently supported". `--webroot-map "{'żąćł.com': '/tmp'}"` on the other hand raises `UnicodeEncodeError`. Would be nice to have consistency between these two options.
Maybe moving `str(domains)` to `_process_domain` and wrapping it in a try clause?

`str(webroot_path)` raises an uncaught exception here as well, but utf-8 characters in webroot_path seem to be unsupported now anyway?

filipochnik:
Other than this unicode thing LGTM :)

pde:
Unicode issue fixed; merging.

filipochnik:
@pde Does it work for you?
```
$ sudo letsencrypt certonly --staging --test-cert --text --webroot --email a@a.com --webroot-map '{"test.com": "/tmp"}'
An unexpected error occurred:
TypeError: initializer for ctype 'char *' must be a str or list or tuple, not unicode
```
Traceback from logs:
```
2016-01-30 13:11:23,525:DEBUG:letsencrypt.cli:Exiting abnormally:
  File "/home/filip/SSD/letsencrypt/venv/bin/letsencrypt", line 9, in <module>
    load_entry_point('letsencrypt', 'console_scripts', 'letsencrypt')()
  File "/home/filip/SSD/letsencrypt/letsencrypt/cli.py", line 1484, in main
  File "/home/filip/SSD/letsencrypt/letsencrypt/cli.py", line 644, in obtain_cert
    _auth_from_domains(le_client, config, domains)
  File "/home/filip/SSD/letsencrypt/letsencrypt/cli.py", line 422, in _auth_from_domains
  File "/home/filip/SSD/letsencrypt/letsencrypt/client.py", line 283, in obtain_and_enroll_certificate
  File "/home/filip/SSD/letsencrypt/letsencrypt/client.py", line 264, in obtain_certificate
    csr = crypto_util.init_save_csr(key, domains, self.config.csr_dir)
  File "/home/filip/SSD/letsencrypt/letsencrypt/crypto_util.py", line 78, in init_save_csr
    csr_pem, csr_der = make_csr(privkey.pem, names)
  File "/home/filip/SSD/letsencrypt/letsencrypt/crypto_util.py", line 118, in make_csr
    value=", ".join("DNS:%s" % d for d in domains)
  File "/home/filip/SSD/letsencrypt/venv/lib/python2.7/site-packages/OpenSSL/crypto.py", line 651, in __init__
    extension = _lib.X509V3_EXT_nconf(_ffi.NULL, ctx, type_name, value)
TypeError: initializer for ctype 'char *' must be a str or list or tuple, not unicode
```
Seems like pyOpenSSL expects str.

pde:
@filipochnik sorry just saw this note; yes, current git master works fine for me with a command like that.

Aside: `--staging` and `--test-cert` are synonyms, you don't need both!

filipochnik:
Yeah, the issue was fixed by #2336 

Didn't know that about `--staging` and `--test-cert`, thanks.

PR #2312, opened 2016-01-29
If Nginx does not found vhost info, does not create it

tdruiva:
For issue like #925 was created the PR #2097, but looks like the dir structure for each platform is different :disappointed: 

So, for now if we didn't find vhost in the config files, weren't create it.

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/configurator.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/configurator.py
@@ -219,17 +219,13 @@ class NginxConfigurator(common.Plugin):
         :rtype: :class:`~letsencrypt_nginx.obj.VirtualHost`
 
         """
-        vhost = None
-
         matches = self._get_ranked_matches(target_name)
         if not matches:
-            # No matches. Create a new vhost with this name in nginx.conf.
-            filep = self.parser.loc["root"]
-            new_block = [['server'], [['server_name', target_name]]]
-            self.parser.add_http_directives(filep, new_block)
-            vhost = obj.VirtualHost(filep, set([]), False, True,
-                                    set([target_name]), list(new_block[1]))
-        elif matches[0]['rank'] in xrange(2, 6):
+            logger.info("Virtual Host not found for %s", target_name)
+            return None
+
+
+        if matches[0]['rank'] in xrange(2, 6):
             # Wildcard match - need to find the longest one
             rank = matches[0]['rank']
             wildcards = [x for x in matches if x['rank'] == rank]
@@ -237,9 +233,8 @@ class NginxConfigurator(common.Plugin):
         else:
             vhost = matches[0]['vhost']
 
-        if vhost is not None:
-            if not vhost.ssl:
-                self._make_server_ssl(vhost)
+        if not vhost.ssl:
+            self._make_server_ssl(vhost)
 
         return vhost
 
--- a/letsencrypt-nginx/letsencrypt_nginx/configurator.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/configurator.py
@@ -612,15 +612,20 @@ class NginxConfigurator(common.Plugin):
         outstanding challenges will have to be designed better.
 
         """
-        self._chall_out += len(achalls)
-        responses = [None] * len(achalls)
         chall_doer = tls_sni_01.NginxTlsSni01(self)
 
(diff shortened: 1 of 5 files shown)


tdruiva:
After run letsencript, was added this lines to conf file where was found the vhost based on domain

```
server {
    .
    .
    .
    listen 443 ssl;
    ssl_certificate /etc/letsencrypt/live/letsencrypt.you/fullchain.pem;
    ssl_certificate_key /etc/letsencrypt/live/letsencrypt.you/privkey.pem;
    include /etc/letsencrypt/options-ssl-nginx.conf;
    ssl_trusted_certificate /etc/letsencrypt/live/letsencrypt.you/chain.pem;
    ssl_stapling on;
    ssl_stapling_verify on;

// If you choose to redirect to HTTPS, is added this one
    if ($scheme != "https") {
    }
}

```

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -572,15 +573,24 @@ def perform(self, achalls):
...
             # challenge. This helps to put all of the responses back together
             # when they are all complete.
-            chall_doer.add_chall(achall, i)
+            # As we does not create vhost when it isn't found in the current configuration,
+            # we will filter out this domains before to perform TLS challenge
+            if self.choose_vhost(achall.domain) is not None:
+                chall_doer.add_chall(achall, i)
+            else:
+              # TODO: Handle feedback about domain without vhost
+                logger.info("As was not found a virtual host for domain %s, "
```
olabini: Can we change the string here to: "As we couldn't find a virtual host for domain %s, it will be ignored"

letsencrypt-nginx/letsencrypt_nginx/tls_sni_01.py
```diff
@@ -65,8 +65,8 @@ def perform(self):
                 if addr.default:
                     addresses.append([obj.Addr.fromstring(default_addr)])
                     break
-            else:
-                addresses.append(list(vhost.addrs))
+                else:
```
olabini: Is this change actually correct?

tdruiva: Nope, is not correct. Thank you for highlight it. 
I was thinking that this `else` was related with the `if` inside of the last for loop but is not, it's related with the if in the L:57, also this else is unnecessary because we're returning if not found a vhost.

letsencrypt-nginx/letsencrypt_nginx/configurator.py
tdruiva: :+1:

letsencrypt-nginx/letsencrypt_nginx/configurator.py
```diff
@@ -213,17 +216,13 @@ def choose_vhost(self, target_name):
...
         matches = self._get_ranked_matches(target_name)
         if not matches:
-            # No matches. Create a new vhost with this name in nginx.conf.
-            filep = self.parser.loc["root"]
-            new_block = [['server'], [['server_name', target_name]]]
-            self.parser.add_http_directives(filep, new_block)
-            vhost = obj.VirtualHost(filep, set([]), False, True,
-                                    set([target_name]), list(new_block[1]))
-        elif matches[0]['rank'] in xrange(2, 6):
+            logger.info("Virtual Host not found for %s", target_name)
```
pde: We could expand this to say `"Virtual Host not found for %s. Are you sure you're really serving that domain name?"`  It could in the future make sense to create transient vhosts in this case, for instance to support a case like `letsencrypt --authenticator nginx --installer exim -d mail.example.com`, even if nginx only answers for `www.example.com`.  But that may be a low urgency case?

pde:
Aside from the one tiny change I suggested (an extra sentence in that string), this looks good to me.

letsencrypt-nginx/letsencrypt_nginx/configurator.py
tdruiva: :+1: 
Make sense for me also not looking to the future, the phrase "Are you sure you're really serving that domain name?" for me, provide a better user feedback :)

ilanbiala:
What's the status on this?

pde:
A solution for this was included in the final Nginx patches that landed for the Certbot 0.9.0 release.

PR #2313, opened 2016-01-29
Fix multiple snakeoil certs for nginx

### LINKED ISSUE #923: With multiple domains, key mismatch for snakeoil cert
```
letsencrypt --server http://localhost:4000/directory --no-verify-ssl --dvsni-port 5001 --simple-http-port 5001           --text         --agree-eula         --agree-tos         --email "" --authenticator nginx -d yo.wtf -d foo.wtf auth  

...
2015-10-07 22:20:07,914:ERROR:letsencrypt_nginx.configurator:Nginx Restart Failed!
nginx: [emerg] SSL_CTX_use_PrivateKey_file("/var/lib/letsencrypt/snakeoil/0037_key.pem") failed (SSL: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch)
```

This doesn't happen with a single domain. Poking around, it looks like multiple snakeoil keys are created but only one cert, which is probably the reason for the mismatch.

zauguin:
Fixes #923 by using le_util.unique_file to avoid overwriting existing certificates.

### DIFF
--- a/letsencrypt-nginx/letsencrypt_nginx/configurator.py
+++ b/letsencrypt-nginx/letsencrypt_nginx/configurator.py
@@ -310,11 +310,11 @@ class NginxConfigurator(common.Plugin):
         key = OpenSSL.crypto.load_privatekey(
             OpenSSL.crypto.FILETYPE_PEM, le_key.pem)
         cert = acme_crypto_util.gen_ss_cert(key, domains=[socket.gethostname()])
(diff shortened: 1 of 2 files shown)


olabini:
I feel a bit comfortable about opening a file and then closing it without guarding against IO-exception etc. Except for that I think it's fine.

pde:
Great, thanks!

Issue #989, opened 2015-10-15
Plugin prompts when run with the Renewer


bmw:
What happens when a plugin wants to prompt the user when run with the `renewer`? Currently, the plugin has no idea it is not being run interactively and may attempt to do so. I just experienced this with the Apache plugin. I had only the default Debian Apache setup and the plugin tried to prompt me which vhost to use. This will cause renewal to fail.

Have we thought about this problem before?

pde:
Between having landed `--non-interactive`, and our plan to implement a test mode, this should be solved for 0.4.0.

Issue #2315, opened 2016-01-29
Reporter doesn't respect on_crash parameter


bmw:
`letsencrypt.reporter.Reporter.add_message` has a `on_crash` parameter that is supposed determine whether the registered message should be shown if the program exits with an unhandled exception.

Due to the way we suppress all exceptions, this does not work properly and needs  to be fixed.

Marking this as low priority since there is currently only one place in the code where we add a message with `on_crash=False` and it is the last thing done before the program exits.

bmw:
Error handling on exit has been rewritten since this issue was created and we have plans to completely remove the reporter so I'm closing this issue.

PR #2316, opened 2016-01-29
--dry-run

### LINKED ISSUE #2303: Add --dry-run
We should have a `--test-renew` flag that sets defaults appropriate to testing whether renewal would work properly (including selecting the staging server and not saving the resulting certificate).

bmw:
Fixes #2303

In writing this I uncovered #2315, but I don't think is a high priority.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -404,12 +404,13 @@ def _auth_from_domains(le_client, config, domains):
         # https://github.com/letsencrypt/letsencrypt/pull/777/files#r40498574
         new_certr, new_chain, new_key, _ = le_client.obtain_certificate(domains)
         # TODO: Check whether it worked! <- or make sure errors are thrown (jdk)
-        lineage.save_successor(
-            lineage.latest_common_version(), OpenSSL.crypto.dump_certificate(
-                OpenSSL.crypto.FILETYPE_PEM, new_certr.body.wrapped),
-            new_key.pem, crypto_util.dump_pyopenssl_chain(new_chain))
+        if not config.dry_run:
+            lineage.save_successor(
+                lineage.latest_common_version(), OpenSSL.crypto.dump_certificate(
+                    OpenSSL.crypto.FILETYPE_PEM, new_certr.body.wrapped),
+                new_key.pem, crypto_util.dump_pyopenssl_chain(new_chain))
 
-        lineage.update_all_links_to(lineage.latest_common_version())
+            lineage.update_all_links_to(lineage.latest_common_version())
         # TODO: Check return value of save_successor
         # TODO: Also update lineage renewal config with any relevant
         #       configuration values from this attempt? <- Absolutely (jdkasten)
@@ -419,7 +420,7 @@ def _auth_from_domains(le_client, config, domains):
         if lineage is False:
             raise errors.Error("Certificate could not be obtained")
 
-    if lineage is not None:
+    if not config.dry_run:
         _report_new_cert(lineage.cert, lineage.fullchain)
 
     return lineage, action
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -405,20 +405,11 @@ def _auth_from_domains(le_client, config, domains):
     # (which results in treating the request as a new certificate request).
 
     action, lineage = _treat_as_renewal(config, domains)
-    if action == "reinstall" and not config.dry_run:
+    if action == "reinstall":
         # The lineage already exists; allow the caller to try installing
         # it without getting a new certificate at all.
         return lineage, "reinstall"
-    elif action == "newcert":
-        # TREAT AS NEW REQUEST
-        lineage = le_client.obtain_and_enroll_certificate(domains)
-        if lineage is False:
-            raise errors.Error("Certificate could not be obtained")
-    else:
-        assert action == "renew" or config.dry_run, "invalid auth command"
-        if config.dry_run and action == "reinstall":
-            logger.info("Cert not due for renewal, but "
-                        "simulating renewal for dry run")
+    elif action == "renew":
         original_server = lineage.configuration["renewalparams"]["server"]
         _avoid_invalidating_lineage(config, lineage, original_server)
         # TODO: schoen wishes to reuse key - discussion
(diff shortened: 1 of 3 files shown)


letsencrypt/cli.py
```diff
@@ -383,6 +383,12 @@ def _suggest_donate():
     reporter_util.add_message(msg, reporter_util.LOW_PRIORITY)
 
 
+def _report_successful_dry_run():
+    reporter_util = zope.component.getUtility(interfaces.IReporter)
+    reporter_util.add_message("The dry run was successful.",
```
pde: Maybe elaborate a little here?  "Using these options, we appear to be able to pass ACME Challenges and obtain a cert for <names>".

letsencrypt/cli.py
```diff
@@ -404,25 +410,28 @@ def _auth_from_domains(le_client, config, domains):
...
         # TODO: Check whether it worked! <- or make sure errors are thrown (jdk)
-        lineage.save_successor(
-            lineage.latest_common_version(), OpenSSL.crypto.dump_certificate(
-                OpenSSL.crypto.FILETYPE_PEM, new_certr.body.wrapped),
-            new_key.pem, crypto_util.dump_pyopenssl_chain(new_chain))
+        if not config.dry_run:
+            lineage.save_successor(
+                lineage.latest_common_version(), OpenSSL.crypto.dump_certificate(
+                    OpenSSL.crypto.FILETYPE_PEM, new_certr.body.wrapped),
+                new_key.pem, crypto_util.dump_pyopenssl_chain(new_chain))
```
pde: Perhaps:
```
else:
    logger.info("Dry run: skipping saving cert to lineage {0}".format(lineage.RELEVANT PROPERTY)
```

pde:
I think that [this logic](https://github.com/letsencrypt/letsencrypt/pull/2316/files#diff-54b2f468d57ef7436008bd031bc89711R402) may need to be changed? The point of a dry run is to see what will happen when renewal is due, not to skip renewal because it's not due yet.  So if we end up in the `keep / reinstall` case, we should print a note saying something like "cert not due for renewal yet, but simulating renewal for dry run" and then set `action` to `"renew"`...

letsencrypt/cli.py
```diff
@@ -631,14 +640,18 @@ def obtain_cert(args, config, plugins):
     if args.csr is not None:
         certr, chain = le_client.obtain_certificate_from_csr(le_util.CSR(
             file=args.csr[0], data=args.csr[1], form="der"))
-        cert_path, _, cert_fullchain = le_client.save_certificate(
-            certr, chain, args.cert_path, args.chain_path, args.fullchain_path)
-        _report_new_cert(cert_path, cert_fullchain)
+        if not args.dry_run:
+            cert_path, _, cert_fullchain = le_client.save_certificate(
+                certr, chain, args.cert_path, args.chain_path, args.fullchain_path)
+            _report_new_cert(cert_path, cert_fullchain)
```
pde: again, perhaps log that you're skipping this

letsencrypt/cli.py
```diff
@@ -631,14 +640,18 @@ def obtain_cert(args, config, plugins):
...
     else:
         domains = _find_domains(args, installer)
         _auth_from_domains(le_client, config, domains)
 
-    _suggest_donate()
+    if args.dry_run:
+        _report_successful_dry_run()
+    else:
+        _suggest_donate()
```
pde: There are probably a bunch more cases where donation shouldn't be suggested (eg `--staging`).  I'm not sure if those are best added in the subcommand implementations, or moved inside `_suggest_donation()` itself (it could become `_maybe_suggest_donation()`.

letsencrypt/client.py
```diff
@@ -298,12 +298,14 @@ def obtain_and_enroll_certificate(self, domains):
                 "Non-standard path(s), might not work with crontab installed "
                 "by your operating system package manager")
 
-        lineage = storage.RenewableCert.new_lineage(
-            domains[0], OpenSSL.crypto.dump_certificate(
-                OpenSSL.crypto.FILETYPE_PEM, certr.body.wrapped),
-            key.pem, crypto_util.dump_pyopenssl_chain(chain),
-            params, config, cli_config)
-        return lineage
+        if cli_config.dry_run:
```
pde: log non-creation of new linege here

letsencrypt/cli.py
```diff
@@ -827,15 +840,24 @@ def parse_args(self):
         parsed_args.verb = self.verb
 
         # Do any post-parsing homework here
+        if parsed_args.staging or parsed_args.dry_run:
+            if (parsed_args.server not in
+                    (flag_default("server"), constants.STAGING_URI)):
+                conflicts = ["--staging"] if parsed_args.staging else []
+                if parsed_args.dry_run:
```
pde: `conflicts += ["--dry-run"] if parsed_args.dry_run else []`

pde: (Makes it much clearer the two flags are being treated equivalently)

pde:
LGTM otherwise.

letsencrypt/client.py
bmw: fixed

bmw:
I believe I made all suggested changes.

bmw:
Also, while it doesn't completely eliminate the issue, this + #2286 + #2078 mitigate #989 enough for me to say:

fixes #989

letsencrypt/cli.py
```diff
@@ -612,7 +630,7 @@ def run(args, config, plugins):  # pylint: disable=too-many-branches,too-many-lo
     else:
         display_ops.success_renewal(domains, action)
```
pde: When `--dry-run` is expanded to support `run`, leaving `action == "reinstall"` will have some consequences that may be incorrect here.  But I don't think it matters for this PR.

letsencrypt/cli.py
```diff
@@ -397,36 +405,46 @@ def _auth_from_domains(le_client, config, domains):
     # (which results in treating the request as a new certificate request).
 
     action, lineage = _treat_as_renewal(config, domains)
```
pde: But I do think you can just simplify a bunch of cases by doing this here:
```
if action == "reinstall" and config.dry_run:
    logger.info("Cert not due for renewal, but simulating renewal for dry run")
    action = "renew"
```

letsencrypt/cli.py
pde: My suggestion above would make this a future non-issue.

pde:
LGTM otherwise.

letsencrypt/cli.py
bmw: done

letsencrypt/cli.py
bmw: and done

pde:
Woohoo!

PR #2318, opened 2016-01-29
Run le-auto tests on Travis.

erikrose:
Run le-auto tests on Travis.

* We choose a different Travis infra for one of the jobs, as in https://github.com/numpy/numpy/blob/master/.travis.yml#L49.
* We keep the language as "python" so the installation of packages (like tox, which we need) doesn't fail.
* Override the before_install to disable the dpkg stuff the other jobs need.
* adduser is redundant with `--groups sudo` above, so we delete it.


SwartzCr:
looks reasonable to me

PR #2322, opened 2016-01-31
Make all except/raises Python 3 compatible.

thanatos:
Made all the `except` & `raise` statements use the newer form that Python 3 supports. letsencrypt already uses this in some places prior to this change, so this isn't breaking compatibility with ancient versions of Python, as that's already done.

Additionally, I removed a double space in front of a URL that I came across.


pde:
LGTM, thanks.

Issue #318, opened 2015-03-25
Add a flag to support "best effort" M-of-N success when requesting N names


pde:
This will be required for renewal.

jdkasten:
Related to #300

pde:
Possible names for this option: <tt>--best-effort</tt>, <tt>--allow-partial-success</tt>, <tt>--allow-subsets-of-names</tt>.

jdkasten:
This should be taken on after full compatibility with Boulder is achieved.  Authorization handling changes in v02 of the spec and the current dummy server doesn't support multiple names in a certificate.

jdkasten:
This is mostly implemented... but I realized that obtain_certificate should take into account this information before requesting a certificate.

SwartzCr:
closing now that #2327 is merged

PR #2327, opened 2016-02-01
Adding --allow-subsets-of-names flag

TheNavigat:
Addresses issue #318 

[do not merge added] because running ``letsencrypt certonly --allow-subsets-of-names -d example.com,example2.com`` fails with the following.

```
An unexpected error occurred:
The client lacks sufficient authorization :: Error creating new cert :: Authorizations for these names not found or expired: example2.com
```


TheNavigat:
I trying to understand how the thing works, and I originally thought that ``_response`` is responsible for returning the validity of the domains. However, I made a mistake thinking that not returning the domain form ``_response`` means that it won't be set to boulder. Little did I know...

It seems to me that ``client.py`` handles the process of issuing a certificate. The problem lies in the fact that a CSR is created for all the domains. https://github.com/TheNavigat/letsencrypt/blob/4f3f2c5507395f1fab25bfb0d2f01f521c35cd0b/letsencrypt/client.py#L247-L266

The CSR then, including the domains, is sent to boulder here. https://github.com/TheNavigat/letsencrypt/blob/4f3f2c5507395f1fab25bfb0d2f01f521c35cd0b/letsencrypt/client.py#L198-L229

As shown above, the CSR is created before the domain is authorized. The result is that to prevent the domain from being sent to boulder (and thus prevent the error), we have to send a different CSR.

So, first choice, we recreate the CSR if there are domains failing.
Second choice, we issue a single CSR for each domain. Which is... yep, you guessed it! Insane.
Third choice, we restructure/recode/redesign ``client.py`` so that the CSR is generated after the domains are authorized and not the other way around.
And there's possibly a fourth choice I'm not considering.

I'm inclined towards the third option, because it sounds reasonable and because it shouldn't be too difficult. I'm not a CSR expert so I apologize if there are any mistakes caused by limited knowledge of how the whole certificate issuance process works, but it looks like CSRs are not used for authorization. A simple split of the functions, and calling authorization from ``obtain_certificate`` (and not ``_obtain_certificate``) should fix it.

TheNavigat:
So, it works!

Apparently the code is a mess and tests are failing, but it's working.

Next step is fixing the tests and cleaning up the code! :)

TheNavigat:
Nope, it doesn't.

``The program letsencrypt (process ID 26764) is already listening on TCP port 443. This will prevent us from binding to that port. Please stop the letsencrypt program temporarily and then try again``

TheNavigat:
Fixed that error. Next problem is. the actual type of the domains parameter is not ``set``, it's ``list``.

Should update code to reflect that.

TheNavigat:
For those who are going to review this, apparently the way this is structured is really, really bad. The PR misses tests and the code isn't really clean because I'm expecting that the solution will need to be altered/redesigned.

I'm having some difficulty coming up with a better solution. I thought of adding a class variable that contains the failed domains, but although this might be better, it's still bad. I also thought of raising an exception and catching it when failed domains exist, instead of returning them, which sounds better. Should I implement the latter?

letsencrypt/auth_handler.py
```diff
@@ -139,11 +154,13 @@ def _respond(self, cont_resp, dv_resp, best_effort):
 
         # Check for updated status...
         try:
-            self._poll_challenges(chall_update, best_effort)
+            result = self._poll_challenges(chall_update, best_effort)
         finally:
             # This removes challenges from self.dv_c and self.cont_c
             self._cleanup_challenges(active_achalls)
 
+        return result
```
SwartzCr: I'd update the doc string to mention that you've added a return value to this function

letsencrypt/cli.py
```diff
@@ -694,7 +694,7 @@ def obtain_cert(config, plugins, lineage=None):
     if config.csr is not None:
         assert lineage is None, "Did not expect a CSR with a RenewableCert"
         csr, typ = config.actual_csr
-        certr, chain = le_client.obtain_certificate_from_csr(config.domains, csr, typ)
+        certr, chain = le_client.obtain_certificate_from_csr(config.domains, csr, False, typ)
```
SwartzCr: I might explicitly make this authzr=False

letsencrypt/cli.py
```diff
@@ -1603,6 +1603,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         "security", "--strict-permissions", action="store_true",
         help="Require that all configuration files are owned by the current "
              "user; only needed if your config is somewhere unsafe like /tmp/")
+    helpful.add(
+        "automation", "--allow-subset-of-names", dest="allow_subset_of_names",
+        action="store_true", default=False,
+        help="Allow subsets of domain names to fail validation without exiting.")
```
SwartzCr: I might say: "Allow subsets of domain names in a single liniage to fail validation without exiting"

letsencrypt/auth_handler.py
```diff
@@ -75,19 +75,34 @@ def get_authorizations(self, domains, best_effort=False):
...
             # Send all Responses - this modifies dv_c and cont_c
-            self._respond(cont_resp, dv_resp, best_effort)
+            response = self._respond(cont_resp, dv_resp, best_effort)
+
+            if response:
+                failed_domains = failed_domains.union(response)
+
+        my_authzr = self.authzr
+
+        logger.debug("authzr: %s", my_authzr)
```
SwartzCr: does this debug statement need to be here

SwartzCr: and if it doesn't should we return my_authzr to self.authzr

letsencrypt/auth_handler.py
```diff
@@ -75,19 +75,34 @@ def get_authorizations(self, domains, best_effort=False):
...
+
+        returnDomains = []
+        #Remove failing domains if best_effort is true
+        for domain in domains:
+            if not domain in failed_domains:
+                returnDomains.append(domain)
```
SwartzCr: this can be simplified as a list comprehension of the form
`returnDomains = [domain for domain in domains if domain not in failed_domains]`

letsencrypt/auth_handler.py
```diff
@@ -75,19 +75,34 @@ def get_authorizations(self, domains, best_effort=False):
...
 
         # Just make sure all decisions are complete.
         self.verify_authzr_complete()
         # Only return valid authorizations
-        return [authzr for authzr in self.authzr.values()
-                if authzr.body.status == messages.STATUS_VALID]
+        return ([authzr for authzr in my_authzr.values()
```
SwartzCr: I'm not sure you need parens around this

letsencrypt/auth_handler.py
```diff
@@ -139,11 +154,13 @@ def _respond(self, cont_resp, dv_resp, best_effort):
 
         # Check for updated status...
         try:
-            self._poll_challenges(chall_update, best_effort)
+            result = self._poll_challenges(chall_update, best_effort)
```
SwartzCr: should we call this failed_domain_list or failed_domains or something? It's a bit oblique at this point what result is expected to be

SwartzCr:
can you elaborate on why you feel that returning a list of failed domains is a bad solution?

TheNavigat:
The point is that ``_respond`` is a function that should return whether it succeeded or failed. It's not "respond and return the failed functions", neither is ``_poll_challenges``, returning failed domains doesn't make sense. Raising an exception kind of does, though, which is why I suggested it.

SwartzCr:
If the intention is to submit a CSR with only the passing domains how do you plan to do it without passing information about the domains that are fialing?

TheNavigat:
What if instead of returning the failed domains from ``_poll_challenges``, we return the successful ones? That makes more sense to me.

TheNavigat:
https://github.com/letsencrypt/letsencrypt/pull/2327#discussion_r54157964

Skipped that bit because I'm still not sure if I should go on with this or switch to returning successful ones, or raising exceptions.

TheNavigat:
Boulder's failing with "Internal Server Error". Here's a log in case we somehow lost the build link. https://www.dropbox.com/s/72wmj8asaekvxy2/log.txt?dl=0

TheNavigat:
Ping @SwartzCr

TheNavigat:
So, apparently Travis needs a kick...

SwartzCr:
kicked, I'll take a look at this today, how are you feeling about it @TheNavigat ?

SwartzCr:
Looks mostly good, but I don't see any way that the failed domains will be shown to the user, do you think it'd be reasonable to log or print something to that effect?

TheNavigat:
Regarding the log, ah, true. I'll add a ``logger.debug`` statement in a few hours, then.

I still have a slightly bad feeling about the code and I think the fact that this hasn't been done in a cleaner way will bite us later, but I also think that we should go ahead with it anyway and get to finish the other tasks rotting on the list, so yea, if it looks good to you then please go ahead and merge after the ``logger.debug`` bit is added.

letsencrypt/cli.py
```diff
@@ -1620,6 +1620,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         "security", "--strict-permissions", action="store_true",
         help="Require that all configuration files are owned by the current "
              "user; only needed if your config is somewhere unsafe like /tmp/")
+    helpful.add(
+        "automation", "--allow-subset-of-names", dest="allow_subset_of_names",
```
bmw: Use of `dest` is unnecessary. `argparse` will automatically set the destination to the value you've set explicitly.

letsencrypt/cli.py
```diff
@@ -1620,6 +1620,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        action="store_true", default=False,
```
bmw: Similarly, using `default=False` is unnecessary when using `action="store_true". The default value is automatically `False`.

letsencrypt/cli.py
```diff
@@ -1620,6 +1620,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        help="Allow subsets of domain names in a single lineage to fail "
+             "validation without exiting.")
```
bmw: To me, this help isn't very clear. I think we should err of writing too much help rather than two little. I'm not sold on this, but what about something like:
> When performing domain validation, do not consider it a failure if authorizations can not be obtained for a strict subset of the requested domains. This option cannot be used with --csr.

With this in mind, perhaps we should enforce that `--csr` and `--allow-subset-of-domains` aren't used together. Due to the arguments being in different groups, we can't use [argparse's mutual exclusion](https://docs.python.org/2.7/library/argparse.html#mutual-exclusion), but perhaps we should check for this early somewhere in `cli.py` and error out with a useful error message explaining the problem.

letsencrypt/cli.py
```diff
@@ -694,7 +694,7 @@ def obtain_cert(config, plugins, lineage=None):
     if config.csr is not None:
         assert lineage is None, "Did not expect a CSR with a RenewableCert"
         csr, typ = config.actual_csr
-        certr, chain = le_client.obtain_certificate_from_csr(config.domains, csr, typ)
+        certr, chain = le_client.obtain_certificate_from_csr(config.domains, csr, typ, authzr=False)
```
bmw: Since the `authzr` argument has a default and using `--allow-subset-of-names` cannot be used with `--csr`, I think this change should be undone.

letsencrypt/client.py
```diff
@@ -215,14 +215,17 @@ def obtain_certificate_from_csr(self, domains, csr,
 
         logger.debug("CSR: %s, domains: %s", csr, domains)
 
-        authzr = self.auth_handler.get_authorizations(domains)
+        if authzr is False:
+            authzr, _ = self.auth_handler.get_authorizations(
+                            domains,
+                            self.config.allow_subset_of_names)
```
bmw: We shouldn't pass in `self.config.allow_subset_of_names` here. If we do and we are unable to obtain an authorization for a requested domain, `boulder` will give us an error because we don't have authorizations for all of the names in the CSR.

letsencrypt/client.py
```diff
@@ -189,7 +189,7 @@ def __init__(self, config, account_, auth, installer, acme=None):
             self.auth_handler = None
 
     def obtain_certificate_from_csr(self, domains, csr,
-        typ=OpenSSL.crypto.FILETYPE_ASN1):
+        typ=OpenSSL.crypto.FILETYPE_ASN1, authzr=False):
```
bmw: We should use `authzr=None` here. Additionally, this option should be documented in the docstring.

letsencrypt/auth_handler.py
```diff
@@ -175,9 +190,8 @@ def _poll_challenges(
                         chall_update[domain].remove(achall)
                 # We failed some challenges... damage control
                 else:
-                    # Right now... just assume a loss and carry on...
                     if best_effort:
-                        comp_domains.add(domain)
+                        failed_domains.add(domain)
```
bmw: To be honest, I think changes in this file are more complicated than they need to be. I think every change in this file can be reverted, except we should log that a challenge was failed on this line. See my comments in `client.py` for how to obtain the list of domains from `authzr`.

letsencrypt/client.py
```diff
@@ -237,12 +240,15 @@ def obtain_certificate(self, domains):
         :rtype: tuple
 
         """
+        authzr, domains = self.auth_handler.get_authorizations(domains,
+                                     self.config.allow_subset_of_names)
+
```
bmw: Something like:
```
domains = [a.body.identifier.value for a in authzr]
```
can be used to obtain the list of domains from `authzr`.

letsencrypt/auth_handler.py
```diff
@@ -66,19 +66,31 @@ def get_authorizations(self, domains, best_effort=False):
...
+
+        returnDomains = [domain for domain in domains
+                        if domain not in failed_domains]
 
         # Just make sure all decisions are complete.
         self.verify_authzr_complete()
         # Only return valid authorizations
         return [authzr for authzr in self.authzr.values()
-                if authzr.body.status == messages.STATUS_VALID]
+                if authzr.body.status == messages.STATUS_VALID], returnDomains
```
bmw: We should check that authorizations succeeded for at least one of the requested names. Otherwise, we get errors like this later:
```
AssertionError: Must provide one or more hostnames for the CSR.
```
To do this, you can store the result of the list comprehension in a local variable and raise an exception if it is empty.

bmw:
Other than a few tricky cases such as `--csr` or when the client fails all attempted challenges, this seems to work well. One additional comment is you should try to make sure that all modified lines have unit test coverage.

Per my comments above, I think we can clean up the code a bit but this looks good. Thanks for yet another PR @TheNavigat!

letsencrypt/auth_handler.py
```diff
@@ -76,9 +76,16 @@ def get_authorizations(self, domains, best_effort=False):
...
         # Just make sure all decisions are complete.
         self.verify_authzr_complete()
+
         # Only return valid authorizations
-        return [authzr for authzr in self.authzr.values()
-                if authzr.body.status == messages.STATUS_VALID]
+        retVal = [authzr for authzr in self.authzr.values()
+                  if authzr.body.status == messages.STATUS_VALID]
+
+        if len(retVal) <= 0:
```
bmw: This can simply be:
```
if not retVal:
```

Isn't Python great?

letsencrypt/auth_handler.py
```diff
@@ -76,9 +76,16 @@ def get_authorizations(self, domains, best_effort=False):
```
bmw: GitHub won't let me comment on the exact line, but the return type of this function should be updated. Looks like it's been wrong for a while...

letsencrypt/client.py
```diff
@@ -199,6 +199,8 @@ def obtain_certificate_from_csr(self, domains, csr,
         :param .le_util.CSR csr: DER-encoded Certificate Signing
             Request. The key used to generate this CSR can be different
             than `authkey`.
+        :param dict authzr: ACME Authorization Resource dict where keys are
```
bmw: `authzr` isn't a `dict`, it's a `list`.

letsencrypt/client.py
```diff
@@ -237,12 +240,20 @@ def obtain_certificate(self, domains):
         :rtype: tuple
 
         """
+        authzr = self.auth_handler.get_authorizations(
+                domains,
+                self.config.allow_subset_of_names)
+
+        domains = [a.body.identifier.value.encode('ascii', 'ignore')
```
bmw: I don't think `'ignore'` should be passed to `encode`. It should be impossible for this to fail because we encode every domain as `ascii` [here](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/le_util.py#L313). If it does fail, it's a problem and we should error out.

Out of curiousity, did using `encode('ascii')`fail on you or were you just trying to lookout for errors in the future?

letsencrypt/tests/client_test.py
```diff
@@ -115,15 +115,22 @@ def test_init_acme_verify_ssl(self):
 
     def _mock_obtain_certificate(self):
         self.client.auth_handler = mock.MagicMock()
+        self.client.auth_handler.get_authorizations.return_value = (None, None)
```
bmw: This return type is wrong.

letsencrypt/client.py
```diff
@@ -215,14 +217,15 @@ def obtain_certificate_from_csr(self, domains, csr,
 
         logger.debug("CSR: %s, domains: %s", csr, domains)
 
-        authzr = self.auth_handler.get_authorizations(domains)
+        if authzr is None:
+            authzr = self.auth_handler.get_authorizations(domains)
```
bmw: Looks like this line is missing coverage. It'd be nice to add a test here.

bmw:
I had a few additional nits, mostly about documentation, but otherwise this looks good!

letsencrypt/client.py
TheNavigat: encode('ascii') itself didn't fail, but I had to encode the domains since they were originally encoded in unicode. Boulder integration test failed as a result.

letsencrypt/client.py
```diff
@@ -199,7 +199,7 @@ def obtain_certificate_from_csr(self, domains, csr,
         :param .le_util.CSR csr: DER-encoded Certificate Signing
             Request. The key used to generate this CSR can be different
             than `authkey`.
-        :param dict authzr: ACME Authorization Resource dict where keys are
+        :param list authzr: ACME Authorization Resource dict where keys are
```
bmw: Looks like the description of `authzr` needs to be updated.

> ACME Authorization Resource dict where keys are domains and values are :class:`acme.messages.AuthorizationResource

isn't quite right.

bmw:

PR #2333, opened 2016-02-02
Avoid dangerous and mysterious behaviour if someone tries to modify a config [needs revision]

pde:


bmw:
This doesn't seem to work so I ripped this commit out of #2338.

PR #2335, opened 2016-02-02
Stop leaving socache_shmcb enabled

### LINKED ISSUE #2168: apache reverter can fail due to module dependency order
Running on Ubuntu 12 with the wrong augeas version, and a only default vhosts present, I hit this interesting error:
```
No vhost exists with servername or alias of: maybe.isnot.org. No     │  
                                               │ vhost was selected. Please specify servernames in the Apache config  │  
                                               │ Error while running a2dismod socache_shmcb.                          │  
                                               │ ERROR: The following modules depend on socache_shmcb and need to be  │  
                                               │ disabled first: ssl                                                  │  
                                               │ Unable to run undo command: a2dismod socache_shmcb                   │  
                                               │ Error while running a2dismod mime.                                   │  
                                               │ WARNING: The following essential module will be disabled.            │  
                                               │ This might result in unexpected behavior and should NOT be done      │  
                                               │ unless you know exactly what you are doing!                          │  
                                               │  mime                                                                │  
                                               │ To continue type in the phrase 'Yes, do as I say!' or retry by       │  
                                               │ passing '-f': Aborting                                               │  
                                               │ Unable to run undo command: a2dismod mime 
```
Probably we shouldn't freak out so much in this case...

bmw:
fixes #2168

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -644,11 +644,11 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         """
 
         if self.conf("handle-modules"):
-            if "ssl_module" not in self.parser.modules:
-                self.enable_mod("ssl", temp=temp)
             if self.version >= (2, 4) and ("socache_shmcb_module" not in
                                            self.parser.modules):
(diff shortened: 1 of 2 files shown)


pde:
LGTM.

PR #2336, opened 2016-02-02
Name sanity

schoen:
This attempts to centralize enforcement of domain name sanity checking and also guarantees that names are in ASCII format (from Python 2.7's point of view, `str` objects), which OpenSSL will want.

I think in Python 3 the result will be that domain names are represented as `bytes` objects, which may cause some nuisances: they probably need to be so for passing to OpenSSL, but may need *not* to be in other contexts!


pde:
LGTM.

pde:
This fixes some unfiled bugs with unicode actually causing problems if `webroot-map` is passed in on the command line.

PR #2338, opened 2016-02-02
Renew verb [needs revision]

schoen:
Do not merge yet (depends on name_sanity, and also may fail tests and needs additional test coverage, plus some solution to the restart issue)

This is work in progress on adding the renew verb and this version has succeeded in renewing a cert.


letsencrypt/cli.py
```diff
@@ -693,6 +706,94 @@ def install(args, config, plugins):
     le_client.enhance_config(domains, config)
 
 
+def renew(args, cli_config, plugins):
+    """Renew previously-obtained certificates."""
+    cli_config = configuration.RenewerConfiguration(cli_config)
+    configs_dir = cli_config.renewal_configs_dir
+    for renewal_file in reversed(os.listdir(configs_dir)):
```
pde: Rather than being inline, this loop should probably call subroutines, one to handle the config merge and (if it's more than ~5 LOC) one to invoke `obtain_cert` for that config.

letsencrypt/cli.py
```diff
@@ -693,6 +706,94 @@ def install(args, config, plugins):
...
+        renewalparams = renewal_candidate.configuration["renewalparams"]
+        if "authenticator" not in renewalparams:
+            logger.warning("Renewal configuration file %s does not specify "
+                           "an authenticator. Skipping.", full_path)
+            continue
+    # ?? config = configuration.NamespaceConfig(_AttrDict(renewalparams))
+        # XXX: also need: webroot_map
+        # XXX: also need: nginx_, apache_, and plesk_ items
+        # string-valued items to add if they're present
+        for config_item in ["config_dir", "log_dir", "work_dir", "user_agent",
```
pde: set these lists as constants outside this function (though in this file is fine).

letsencrypt/cli.py
```diff
@@ -693,6 +706,94 @@ def install(args, config, plugins):
...
+                    value = int(renewalparams[config_item])
+                    config.__setattr__(config_item, value)
+                except ValueError:
+                    logger.warning("Renewal configuration file %s specifies "
+                                   "a non-numeric value for %s. Skipping.",
+                                   full_path, config_item)
+                    continue
+        # XXX: ensure that each call here replaces the previous one
+        zope.component.provideUtility(config)
+        # try:
```
pde: remove comment...

letsencrypt/cli.py
```diff
@@ -695,6 +730,116 @@ def install(config, plugins):
     le_client.enhance_config(domains, config)
 
 
+def renew(cli_config, plugins):
+    """Renew previously-obtained certificates."""
+    cli_config = configuration.RenewerConfiguration(cli_config)
+    if cli_config.domains != []:
```
pde: This turns out to make it impossible to pass `--webroot-map` to the renewer for the purposes of changing the directories of webroot entries, which is a pity.

letsencrypt/cli.py
```diff
@@ -208,15 +224,12 @@ def _find_duplicative_certs(config, domains):
...
+    for renewal_file in _renewal_conf_files(cli_config):
         try:
-            full_path = os.path.join(configs_dir, renewal_file)
-            candidate_lineage = storage.RenewableCert(full_path, cli_config)
+            candidate_lineage = storage.RenewableCert(renewal_file, cli_config)
         except (errors.CertStorageError, IOError):
-            logger.warning("Renewal configuration file %s is broken. "
-                           "Skipping.", full_path)
+            logger.warning("Renewal conf file %s is broken. Skipping.", renewal_file)
+            logger.info("Traceback was:\n%s", traceback.format_exc())
```
bmw: nit: Maybe this should be `logger.debug`? Seems like a traceback is really only used for debugging purposes.

letsencrypt/cli.py
```diff
@@ -235,7 +248,8 @@ def _find_duplicative_certs(config, domains):
 
 
 def _treat_as_renewal(config, domains):
-    """Determine whether there are duplicated names and how to handle them.
+    """Determine whether there are duplicated names and how to handle them
+    (renew, reinstall, newcert, or no action).
```
bmw: Question: Where do we return no action?

letsencrypt/cli.py
```diff
@@ -657,6 +686,7 @@ def obtain_cert(config, plugins):
 
     # This is a special case; cert and chain are simply saved
     if config.csr is not None:
+        assert lineage is None, "Did not expect a CSR with a RenewableCert"
```
bmw: Scenarios where this could occur:
* Someone modified their renewal config file (highly unlikely)
* They provided a CSR on the command line when running `letsencrypt renew`. This is certainly possible.

Perhaps we should check for the latter case in the `renew` function and raise an error that the `renew` subcommand does not (currently) support `--csr` and point the user towards `certonly`.

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+                raise
+
+def _restore_plugin_configs(config, renewalparams):
+    # Now use parser to get plugin-prefixed items with correct types
+    # XXX: the current approach of extracting only prefixed items
+    #      related to the actually-used installer and authenticator
+    #      works as long as plugins don't need to read plugin-specific
+    #      variables set by someone else (e.g., assuming Apache
+    #      configurator doesn't need to read webroot_ variables).
+    # XXX: is it true that an item will end up in _parser._actions even
```
bmw: Perhaps we should investigate this before `0.4.0`?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+    #      when no action was explicitly specified?
+    plugin_prefixes = [renewalparams["authenticator"]]
+    if "installer" in renewalparams and renewalparams["installer"] != None:
```
bmw: nit: Should be `renewalparams.get("installer", None) != None:`

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+        plugin_prefixes.append(renewalparams["installer"])
+    for plugin_prefix in set(renewalparams):
```
bmw: Question: Why do we use a set here?

bmw: Additionally, shouldn't this be iterating through `plugin_prefixes`?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+        for config_item in renewalparams.keys():
```
bmw: nit: Couldn't we use `iteritems()` here instead of looking up the value again?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+            if renewalparams[config_item] == "None":
+                # Avoid confusion when, for example, "csr = None" (avoid
+                # trying to read the file called "None")
+                # Should we omit the item entirely rather than setting
+                # its value to None?
+                setattr(config.namespace, config_item, None)
+                continue
+            if config_item.startswith(plugin_prefix + "_"):
```
bmw: The logic of these two lines can be simplified from:
```
continue
if ...
```
to
```
elif ...
```

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+                for action in _parser.parser._actions: # pylint: disable=protected-access
+                    if action.type is not None and action.dest == config_item:
+                        setattr(config.namespace, config_item,
+                                action.type(renewalparams[config_item]))
+                        break
+                else:
+                    setattr(config.namespace, config_item,
+                            str(renewalparams[config_item]))
+    return True
```
bmw: Unnecessary return.

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+            try:
+                value = int(renewalparams[config_item])
+                setattr(config.namespace, config_item, value)
+            except ValueError:
+                logger.warning("Renewal configuration file %s specifies "
+                               "a non-numeric value for %s. Skipping.",
+                               full_path, config_item)
+                raise
+
+def _restore_plugin_configs(config, renewalparams):
```
bmw: Suggestion: Couldn't we add `_parser` to `config` to avoid making it a global? Additionally, could we only store the parser's actions because that's all we care about?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+    #      variables set by someone else (e.g., assuming Apache
+    #      configurator doesn't need to read webroot_ variables).
+    # XXX: is it true that an item will end up in _parser._actions even
+    #      when no action was explicitly specified?
+    plugin_prefixes = [renewalparams["authenticator"]]
+    if "installer" in renewalparams and renewalparams["installer"] != None:
+        plugin_prefixes.append(renewalparams["installer"])
+    for plugin_prefix in set(renewalparams):
+        for config_item in renewalparams.keys():
+            if renewalparams[config_item] == "None":
```
bmw: Why do we store variables with the value of `None` even if they don't start with the correct prefix?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+                # Avoid confusion when, for example, "csr = None" (avoid
+                # trying to read the file called "None")
+                # Should we omit the item entirely rather than setting
+                # its value to None?
+                setattr(config.namespace, config_item, None)
+                continue
+            if config_item.startswith(plugin_prefix + "_"):
+                for action in _parser.parser._actions: # pylint: disable=protected-access
```
bmw: This will have to be changed if the behavior of `None` values is inappropriate, but this `for` loop and the accompanying `else` can be replaced with these two lines of code:
```
typ = next((a.type for a in _parser.parser._actions
            if a.type is not None and a.dest == config_item), str)
setattr(config.namespace, config_item, typ(renewalparams[config_item]))
```

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+
+    :returns: the RenewableCert object or None if a fatal error occurred
+    :rtype: `storage.RenewableCert` or NoneType
+    """
+
+    try:
+        renewal_candidate = storage.RenewableCert(full_path, config)
+    except (errors.CertStorageError, IOError):
+        logger.warning("Renewal configuration file %s is broken. Skipping.", full_path)
+        logger.info("Traceback was:\n%s", traceback.format_exc())
```
bmw: s/info/debug

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+    if "authenticator" not in renewalparams:
+        logger.warning("Renewal configuration file %s does not specify "
+                       "an authenticator. Skipping.", full_path)
+        return None
+    # Now restore specific values along with their data types, if
+    # those elements are present.
+    try:
+        _restore_required_config_elements(full_path, config, renewalparams)
+        _restore_plugin_configs(config, renewalparams)
+    except ValueError:
```
bmw: Question: Can anything else be raised here?

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+
+def _renewal_conf_files(config):
+    """Return /path/to/*.conf in the renewal conf directory"""
+    return glob.glob(os.path.join(config.renewal_configs_dir, "*.conf"))
+
+
+def renew(cli_config, unused_plugins):
+    """Renew previously-obtained certificates."""
+    cli_config = configuration.RenewerConfiguration(cli_config)
+    if cli_config.domains != []:
```
bmw: `if not cli_config.domains:`

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+        raise errors.Error("Currently, the renew verb is only capable of "
+                           "renewing all installed certificates that are due "
+                           "to be renewed; individual domains cannot be "
+                           "specified with this action. If you would like to "
+                           "renew specific certificates, use the certonly "
+                           "command. The renew verb may provide other options "
+                           "for selecting certificates to renew in the future.")
+    for renewal_file in _renewal_conf_files(cli_config):
+        if not renewal_file.endswith(".conf"):
```
bmw: Unnecessary given the implementation of `_renewal_conf_files`

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+            continue
+        print("Processing " + renewal_file)
+        # XXX: does this succeed in making a fully independent config object
+        #      each time?
+        config = configuration.RenewerConfiguration(copy.deepcopy(cli_config))
+        config.noninteractive_mode = True
```
bmw: This doesn't work. We have to check that `renew` was used in `main` and set up the correct interface is setup with `zope`.

cc @SwartzCr

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+
+        # Note that this modifies config (to add back the configuration
+        # elements from within the renewal configuration file).
+        try:
+            renewal_candidate = _reconstitute(renewal_file, config)
+        except Exception as e: # pylint: disable=broad-except
+            # reconstitute encountered an unanticipated problem.
+            logger.warning("Renewal configuration file %s produced an "
+                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.info("Traceback was:\n%s", traceback.format_exc())
```
bmw: `s/info/debug`

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+        # XXX: does this succeed in making a fully independent config object
+        #      each time?
+        config = configuration.RenewerConfiguration(copy.deepcopy(cli_config))
+        config.noninteractive_mode = True
+
+        # Note that this modifies config (to add back the configuration
+        # elements from within the renewal configuration file).
+        try:
+            renewal_candidate = _reconstitute(renewal_file, config)
+        except Exception as e: # pylint: disable=broad-except
```
bmw: In an ideal world, I think we'd modify `_reconstitute` and the functions it calls to make sure it only raises an error derived from `errors.Error` in the places we think an error might reasonably occur. Then here, we'd only catch `errors.Error` and any unexpected exceptions would halt the program.

I don't think we should do this now though.

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+            # reconstitute encountered an unanticipated problem.
+            logger.warning("Renewal configuration file %s produced an "
+                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.info("Traceback was:\n%s", traceback.format_exc())
+            continue
+
+        if renewal_candidate is None:
```
bmw: nit: Rather than using continue, we could just change the logic to `if renewal_candidate is not None` and put the next three lines under that `if` statement.

tests/boulder-integration.sh
```diff
@@ -44,6 +44,22 @@ common --domains le3.wtf install \
        --cert-path "${root}/csr/cert.pem" \
        --key-path "${root}/csr/key.pem"
 
+# This won't renew (because it's not time yet)
+common renew
+
+# This will renew
+sed -i "4arenew_before_expiry = 10 years" "$root/conf/renewal/le1.wtf.conf"
+common renew
```
bmw: We should actually verify that the original `common renew` call didn't renew the certificates and the second one did by examining the configuration directory. I'm fine with just making an issue about this for later though.

bmw:
Finished first review of this. I think it can be cleaned up a bit but that is to be expected with the conditions under which this PR was written.

I didn't see any major problems with this. In general, more {unit, test farm, integration} testing would be nice, but I think this is on the right track.

letsencrypt/cli.py
```diff
@@ -695,6 +725,168 @@ def install(config, plugins):
...
+                           "renew specific certificates, use the certonly "
+                           "command. The renew verb may provide other options "
+                           "for selecting certificates to renew in the future.")
+    for renewal_file in _renewal_conf_files(cli_config):
+        if not renewal_file.endswith(".conf"):
+            continue
+        print("Processing " + renewal_file)
+        # XXX: does this succeed in making a fully independent config object
+        #      each time?
+        config = configuration.RenewerConfiguration(copy.deepcopy(cli_config))
```
pde: Why do you make a `RenewerConfiguration` from this thing?  Isn't it already one?

tests/boulder-integration.sh
pde: I think we should do it before 0.4.0 (perhaps in a separate PR).  It should be easy: just count the files in the `archive/domain/` config subdirectory.

We should also verify that some configuration state has persisted across these runs via the renewal config file, and/but that supply arguments to a renewal run causes them to be respected.

tests/boulder-integration.sh
```diff
@@ -44,6 +44,22 @@ common --domains le3.wtf install \
...
+
+# letsencrypt-renewer $store_flags
+# dir="$root/conf/archive/le1.wtf"
```
pde: This commented code is obsolete; delete it.

letsencrypt/cli.py
```diff
@@ -1506,6 +1705,8 @@ def main(cli_args=sys.argv[1:]):
         displayer = display_util.NoninteractiveDisplay(sys.stdout)
     elif config.text_mode:
         displayer = display_util.FileDisplay(sys.stdout)
+    elif config.verb == "renew":
```
bmw: We should set `config.namespace.noninteractive_mode = True` here so things like https://github.com/letsencrypt/letsencrypt/pull/2338/files#diff-9c91e384e95127e1ff7a8b0e8bc9d521R94 work correctly.

letsencrypt/cli.py
pde: There is a default; it's `store`: https://docs.python.org/3/library/argparse.html#action
I find `argparse.py` fairly hard to disentangle, so I can't say for sure that this will never be `[]`.  But it looks like Seth's code handles that case okay?  So perhaps we can remove this comment?

letsencrypt/cli.py
pde: Done.

letsencrypt/cli.py
schoen: The idea is that only the plugins that are going to be used are relevant. And it's a set because installer and authenticator might be the same or might be different. If you can think of a more concise way of taking two things that could be the same or could be different and making an iterable with the unique ones, I'm happy to use that!

bmw: That makes sense. Using a set to eliminate duplicates is fine.

It still seems like we should be iterating over `plugin_prefixes` though. We currently build `plugin_prefixes` but never make use of it.

letsencrypt/cli.py
schoen: We now check for the latter case in the `renew` function.

letsencrypt/cli.py
```diff
@@ -44,6 +46,19 @@
...
 
+# This is global scope in order to be able to extract type information from
+# it later
+_parser = None
+
+# These are the items which get pulled out of a renewal configuration
+# file's renewalparams and actually used in the client configuration
+# during the renewal process. We have to record their types here because
+# the renewal configuration process loses this information.
+STR_CONFIG_ITEMS = ["config_dir", "log_dir", "work_dir", "user_agent",
```
SwartzCr: log_dir isn't a thing returned by prepare_and_parse_args, I believe this should be logs_dir

letsencrypt/cli.py
```diff
@@ -638,7 +669,7 @@ def run(config, plugins):  # pylint: disable=too-many-branches,too-many-locals
     _suggest_donation_if_appropriate(config)
 
 
-def obtain_cert(config, plugins):
+def obtain_cert(config, plugins, lineage=None):
```
SwartzCr: this doesn't seem like it can a) be called without a lineage supplied, or b) called with a lineage of None, as the only place this is called is on line 912 inside of renew.
at worst this is sloppy looking, but not actually harmful

letsencrypt/cli.py
```diff
@@ -695,6 +727,198 @@ def install(config, plugins):
...
+            full_path, configuration.RenewerConfiguration(config))
+    except (errors.CertStorageError, IOError):
+        logger.warning("Renewal configuration file %s is broken. Skipping.", full_path)
+        logger.debug("Traceback was:\n%s", traceback.format_exc())
+        return None
+    if "renewalparams" not in renewal_candidate.configuration:
+        logger.warning("Renewal configuration file %s lacks "
+                       "renewalparams. Skipping.", full_path)
+        return None
+    renewalparams = renewal_candidate.configuration["renewalparams"]
```
SwartzCr: this implies that we won't be able to specify authenticator on the command line, since storage.RenewableCert doesn't take cli_config into account when making self.configuration. Are we okay with that

letsencrypt/cli.py
pde: main calls it; it gets a reference via the VERBS table.

SwartzCr:
LGTM +1 to merge

PR #2343, opened 2016-02-02
Add 2.6 dependencies that were missing from le-auto. Fix #2334.

### LINKED ISSUE #2334: pkg_resources.DistributionNotFound: unittest2
When running `letsencrypt-auto-source/letsencrypt-auto --debug` on Debian Squeeze, I get the following:
```
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 5, in <module>
  File "/root/.local/share/letsencrypt/lib/python2.6/site-packages/distribute-0.6.10-py2.6.egg/pkg_resources.py", line 2655, in <module>
    working_set.require(__requires__)
  File "/root/.local/share/letsencrypt/lib/python2.6/site-packages/distribute-0.6.10-py2.6.egg/pkg_resources.py", line 648, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/root/.local/share/letsencrypt/lib/python2.6/site-packages/distribute-0.6.10-py2.6.egg/pkg_resources.py", line 546, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: unittest2
```
Using the old le-auto, the script runs successfully.

cc @erikrose

erikrose:
ConfigArgParse has a conditional dependency for Pythons < 2.7. On my local machine, I had a cached ConfigArgParse wheel built under 2.7, so it didn't carry those dependencies, and the pip freeze I used to determine the le-auto requirements thus missed it. From now on, we'll do those passes with --no-cache-dir.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -438,8 +438,8 @@ if [ "$NO_SELF_UPGRADE" = 1 ]; then
     # -------------------------------------------------------------------------
     cat << "UNLIKELY_EOF" > "$TEMP_DIR/letsencrypt-auto-requirements.txt"
 # This is the flattened list of packages letsencrypt-auto installs. To generate
-# this, do `pip install -e acme -e . -e letsencrypt-apache`, `pip freeze`,
-# and then gather the hashes.
+# this, do `pip install --no-cache-dir -e acme -e . -e letsencrypt-apache`, and
+# then use `hashin` or a more secure method to gather the hashes.
 
 # sha256: wxZH7baf09RlqEfqMVfTe-0flfGXYLEaR6qRwEtmYxQ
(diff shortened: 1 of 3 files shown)


erikrose:
@bmw Want to give this a try? The way to try it is to get back to your broken state, then peep install the letsencrypt-auto-requirements.txt, and try running `letsencrypt` again. (Running `letsencrypt-auto` won't work, because it'll pull down and replace itself with the 0.3.0 release, which still has the bug.)

erikrose:
Incidentally, this is considered a pip bug, [fixed in 8.0](https://github.com/pypa/pip/pull/3225), but the fix doesn't retroactively clean up bad wheels in your cache directory, so you can continue to enjoy the bug into the indefinite future until you wipe out your cache dir and swear off older versions of pip forever. :-)

bmw:
I can confirm this fixes #2334. Thanks for writing a PR so quickly @erikrose!

PR #2345, opened 2016-02-02
Ignore venv in letstest dir

bmw:
Convenience for those running test farm tests from inside a venv.


bmw:
Merging this since it's not user facing and makes my life easier.

Issue #2347, opened 2016-02-03
letsencrypt renew should accept command line arguments


bmw:
Arguments provided on the command line should override those found in the renewer config files.

pde:
We accept some arguments at the moment, but not all of them. For instance, if I get a cert with `--standalone` and then try to call `letsencrypt renew --webroot`, I get:

`Too many flags setting configurators/installers/authenticators 'standalone' -> 'webroot'`

pde:
`--rsa-key-size 4096` is accepted, but does not actually cause renewal to use the larger key size.

schoen:
This was fixed by #2388. Ability to change webroot map might still be missing so I will make a separate non-0.4.0 ticket to confirm that.

PR #2358, opened 2016-02-03
accept command line arguments when renewing

SwartzCr:
corresponds to #2347

The logic is if the value supplied on the command line is different from the default value that prepare_and_parse_args provides then we shouldn't overwrite that value with the value from the renewalconfig file.

Note that domains being specified is blocked by an earlier code check that exits if the -d flag is found


bmw:
Should this be against master?

PR #2359, opened 2016-02-03
Remove renewer.py

bmw:
If we ever want this, we can look at the old release tags.


pde:
LGTM.

marcan:
What's the replacement for this? I've been using it (and working around its quirks) on my servers for a while, and just deployed LE on another one today and had to roll back my tree to keep using it.

schoen:
Sorry for the change; the replacement is meant to be

https://github.com/letsencrypt/letsencrypt/pull/2338

in the upcoming 0.4.0 release. This is integrated into the client and
handles preservation of renewal configuration options better.

marcan:
I had a quick look and it looks good. I'll stick with an older commit for now, until that's merged in.

PR #2361, opened 2016-02-04
Update config in save_successor

### LINKED ISSUE #1029: Update renewal config in save_sucessor
Currently, the client never modifies a renewal configuration once it's created. Suppose the client obtains a renewal manually, specifying different values on the command line. The renewal config is not updated with these new values.

bmw:
Fixes #1029 and builds off of #2359.


pde:
This looks okay to me, though I'm not very familiar with `storage.py`, so probably it should have another review from @schoen.

pde:
Also someone should post here saying they've applied careful manual tests...

bmw:
I'm posting here saying that I've applied manual tests.

schoen:
I looked at this yesterday and we talked about it on IRC. I mentioned that we don't currently preserve comments (if the user has edited the configuration file); @bmw agrees and pointed out that the renewal configuration files aren't documented or mentioned anywhere so it's unlikely that many people are successfully manually editing them today. In the future we can also try to use ConfigObj features to preserve users' comments.

PR #2370, opened 2016-02-04
updating readme with free memory requirement [revised]

rorosaurus:
There is a non-obvious requirement for memory to be available for the crypto libraries on the client.  The readme's System Requirements section didn't reflect this.  On low-memory systems this crashes the client with an extremely verbose but not terribly useful error. 
See https://github.com/letsencrypt/letsencrypt/issues/1883#issuecomment-167276122

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,18 +121,24 @@ email to client-dev+subscribe@letsencrypt.org)
 System Requirements
 ===================
 
-The Let's Encrypt Client presently only runs on Unix-ish OSes that include 
-Python 2.6 or 2.7; Python 3.x support will be added after the Public Beta 
-launch. The client may require at least 512MB of free memory in order to 
-properly initalize the cryptography libraries. The client requires root access 
-in order to write to ``/etc/letsencrypt``, ``/var/log/letsencrypt``, 
-``/var/lib/letsencrypt``; to bind to ports 80 and 443 (if you use the 
-``standalone`` plugin) and to read and modify webserver configurations (if you 
-use the ``apache`` or ``nginx`` plugins).  If none of these apply to you, it is 
-theoretically possible to run without root privileges, but for most users who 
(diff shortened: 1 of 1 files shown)


pde:
I agree we should mention that in the readme, but the RAM isn't a requirement of the client; it's a requirement of the installation approach taken by `letsencrypt-auto`.

If there are client packages available for your OS, those will work on low-memory systems.  And if your OS packages a recent copy of `python-cryptography` but not the letsencrypt client, you may be able to do `pip install letsencrypt letsencrypt-apache` and have that work.

Can you update the PR to reflect those nuances in some way?

rorosaurus:
Excellent points.  How does this new wording sound?  Does that capture the details adequately?

Can you confirm I correctly stated this last bit? 
> if your OS packages a recent copy of ``python-cryptography``, you
may be able to run ``letsencrypt`` without the ``letsencrypt-auto`` wrapper,
avoiding the compilation.

pde:
Hi @rorosaurus, I've made [a PR](https://github.com/rorosaurus/letsencrypt/pull/1) to get that wording to be clearer; however it starts to seem like too much detail on this specific issue for the main README.  Should we perhaps put these workaround in docs/using.rst and just link to that section from the top level README?

rorosaurus:
Yeah, I was thinking the same.  It now seems too specific for the main README.  
My original reason for wanting to mention the memory complications here was just to spread awareness.  Maybe a brief message, then a link to that section in docs/using.rst would be perfect?

seanthewebber:
@pde @rorosaurus Can we consider using `letsencrypt` instead of `letsencrypt-auto` an official solution or is this a temporary workaround?

pde:
@seanthewebber yes that's definitely the plan; in the long run we aspire to retiring `letsencrypt-auto` in favour of native packages.

jsha:
Thanks for the PR! I've wrapped the advice into #4305, which also updates some more of the documentation.

PR #2378, opened 2016-02-05
Fixing parameter type for get_authorizations domains parameter

TheNavigat:
https://github.com/letsencrypt/letsencrypt/pull/2367 follow-up.


schoen:
Thanks for the documentation fix.

PR #2380, opened 2016-02-05
don't override commandline flags when renewing

SwartzCr:
refers to #2347


pde:
Closing this, so we can update and reopen against master.

PR #2381, opened 2016-02-05
Document RAM limits

pde:
Building on @rorosaurus's PR...


pde:
Actually, I'm going to make this against the other PR... more to come.

PR #2383, opened 2016-02-05
Make the new letsencrypt-auto script the main one. [requires minor revision]

erikrose:
Remove the old bootstrap scripts, which have been subsumed into letsencrypt-auto-source/pieces/bootstrappers. They no longer need to be dispatched among manually: everyone can just run letsencrypt-auto --os-packages-only, regardless of OS.

Make the root-level le-auto a symlink to the canonical version. It should thus still work for people running le-auto from a git checkout.


erikrose:
Now that there's nothing left in the bootstrap folder, we should move the remaining couple of venv scripts somewhere. Any ideas? `tools/venv`?

pde:
Yeah, or just `tools/`.  Can you also grep the `docs/` for references to these things in their current locations, and update accordingly?

erikrose:
Yep, I already grepped the docs for "bootstrap", of course. I'll move the venv stuff and update the docs for those as well.

erikrose:
@pde Done.

pde:
Merging!

PR #2388, opened 2016-02-06
Ensure that command line arguments still work with `renew`

pde:
Quite a lot of hairiness is involved in knowing whether a given attribute of `config` was set by the user, or came as a default...


letsencrypt/cli.py
```diff
@@ -850,7 +876,7 @@ def _reconstitute(config, full_path):
     # webroot_map is, uniquely, a dict, and the general-purpose
     # configuration restoring logic is not able to correctly parse it
     # from the serialized form.
-    if "webroot_map" in renewalparams:
+    if "webroot_map" in renewalparams and not _set_by_cli("webroot_map"):
```
SwartzCr: since add passes on webroot_map won't this never evaluate to True && False?

letsencrypt/cli.py
```diff
@@ -862,9 +888,10 @@ def _reconstitute(config, full_path):
                        "invalid. Skipping.", full_path)
         return None
 
-    setattr(config.namespace, "domains", domains)
-    return renewal_candidate
+    if not _set_by_cli("domains"):
```
SwartzCr: since domains defaults to [] and the new add passes on it, could this be replaced with if not getattr(config.namespace, "domains")
though I could see why you might not want to do that for uniformity

SwartzCr:
The code looks good to me
my comments mention some superfluous logic, but not logic that will cause errors +1 to merge dependent on solid testing

letsencrypt/cli.py
```diff
@@ -1166,6 +1200,23 @@ def add(self, topic, *args, **kwargs):
...
+            # These are config elements which cannot tolerate being set to ""
+            # during parsing; that's fine as long as their defaults evalute to
+            # boolean false.
+            if not any(exception in args for exception in ["--webroot-map", "-d", "-w", "-v"]):
+                arg_type = kwargs.get("type", None)
+                if arg_type == int:
+                    kwargs["default"] = 0
+                elif arg_type == read_file or "-c" in args:
+                    kwargs["default"] = ""
+                    kwargs["type"] = str
```
SwartzCr: could you document why you're changing the type as well as setting the default?

letsencrypt/cli.py
```diff
@@ -733,6 +732,32 @@ def install(config, plugins):
...
+        # Setup on first run: `detector` is a weird version of config in which
+        # the default value of every attribute is wrangled to be boolean-false
+        plugins = plugins_disco.PluginsRegistry.find_all()
+        # reconstructed_args == sys.argv[1:], or whatever was passed to main()
+        reconstructed_args = _parser.args + [_parser.verb]
+        default_args = prepare_and_parse_args(plugins, reconstructed_args, empty_defaults=True)
+        _set_by_cli.detector = configuration.NamespaceConfig(default_args, fake=True)
+    try:
+        # Is detector.variable something that isn't false?
+        if _set_by_cli.detector.__getattr__(variable):
```
SwartzCr: Will this work for arguments that we're setting to false, such as --no-hsts or --no-uir?

SwartzCr:
I'm getting a bug that no matter what the command line flag is things that default already to falsey values are getting set to false no matter what the commandline flag

steps to reproduce:
call renew with --hsts --redirect --renew-by-default
call renew again
note: despite renew by default being true in the conf file it doesn't get renewed
call renew again with --renew-by-default
note that hsts = False and redirect = None

letsencrypt/cli.py
```diff
@@ -1166,6 +1200,23 @@ def add(self, topic, *args, **kwargs):
         it, but can be None for `always documented'.
 
         """
+
+        if self.empty_defaults:
```
SwartzCr: this should be moved to a separate function

letsencrypt/cli.py
pde: It's safe to pass it on because its default value is `{}`, which is false.  If it's set by the user, it's likely to become true.

pde:
@SwartzCr addressed all your comments, including handling the `store_false` variables, in the anticipation that some of them might one day apply to the renew verb.

SwartzCr:
lgtm

pde:
Review from schoen:
> As I mentioned on IRC, testing this (with `letsencrypt renew`) didn't work properly for me; I got a ton of errors like
>
> `2016-02-09 04:10:15,727:WARNING:letsencrypt.cli:Missing default analysis for 'tls_sni_01_port'`
> 
> and it seems that `--rsa-key-size` is ignored (it doesn't affect the key size chosen for renewal and doesn't change the renewal configuration file).

CHANGES.rst
```diff
@@ -5,23 +5,7 @@ Please note:
 the change log will only get updated after first release - for now please use the
 `commit log <https://github.com/letsencrypt/letsencrypt/commits/master>`_.
 
+To see the changes in a given release, inspect the github milestone for the
```
bmw: nit: Why are these changes in this PR?

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+    """
+    detector = _set_by_cli.detector
+    if detector is None:
+        # Setup on first run: `detector` is a weird version of config in which
+        # the default value of every attribute is wrangled to be boolean-false
+        plugins = plugins_disco.PluginsRegistry.find_all()
+        # reconstructed_args == sys.argv[1:], or whatever was passed to main()
+        reconstructed_args = _parser.args + [_parser.verb]
+        default_args = prepare_and_parse_args(plugins, reconstructed_args, detect_defaults=True)
+        detector = _set_by_cli.detector = configuration.NamespaceConfig(default_args, fake=True)
```
bmw: What is the benefit of creating a `NamespaceConfig` and the accompanying changes versus just using the return value of `prepare_and_parse args`? Seems like unnecessary wrapping/complexity.

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+        # propagate plugin requests: eg --standalone modifies config.authenticator
+        auth, inst = cli_plugin_requests(detector)
+        if auth:
+            detector.namespace.__setattr__("authenticator", auth)
```
bmw: `detector.namespace.authenticator = auth`

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+        if inst:
+            detector.namespace.__setattr__("installer", inst)
```
bmw: ditto

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+        logger.debug("Default Detector is %r", detector.namespace)
+
+    try:
+        # Is detector.var something that isn't false?
+        change_detected = detector.__getattr__(var)
+    except AttributeError:
+        logger.warning("Missing default analysis for %r", var)
+        return False
+
+    if change_detected:
```
bmw: The idea of checking for "falsy" values seems very fragile to me. With your special casing of `store_false`, I feel fairly confident this approach works now, at least for the built-in plugins, but I think it is likely to break in the future when values can legitimately take on false values.

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+        return True
+    # Special case: we actually want account to be set to "" if the server
+    # the account was on has changed
+    elif var == "account" and (detector.server or detector.dry_run or detector.staging):
```
bmw: Hardcoding relationships between arguments like this for `_set_by_cli` seems fragile and is likely to break in the long term.

letsencrypt/cli.py
```diff
@@ -740,6 +738,52 @@ def install(config, plugins):
...
+        return True
+    # Special case: vars like --no-redirect that get set True -> False
+    # default to None; False means they were set
+    elif var in detector.store_false_vars and change_detected is not None:
```
bmw: nit: You can just return the result of this comparison here

letsencrypt/cli.py
```diff
@@ -870,9 +914,10 @@ def _reconstitute(config, full_path):
                        "was: %s. Skipping.", full_path, error)
         return None
 
-    setattr(config.namespace, "domains", domains)
-    return renewal_candidate
+    if not _set_by_cli("domains"):
+        setattr(config.namespace, "domains", domains)
```
bmw: nit: `config.namespace.domains = domains`

letsencrypt/constants.py
```diff
@@ -22,6 +22,7 @@
     config_dir="/etc/letsencrypt",
     work_dir="/var/lib/letsencrypt",
     logs_dir="/var/log/letsencrypt",
+    renewal_glob="*.conf",
```
bmw: Unnecessary. I believe this is a remnant of your CLI option you were using for testing.

letsencrypt/cli.py
```diff
@@ -1112,6 +1158,13 @@ def __init__(self, args, plugins):
         self.parser._add_config_file_help = False  # pylint: disable=protected-access
         self.silent_parser = SilentParser(self.parser)
 
+        # This setting attempts to force all default values to None; it
```
bmw: Incorrect comment. It doesn't force all default values to `None`, it forces them to falsy values.

letsencrypt/configuration.py
```diff
@@ -34,7 +34,7 @@ class NamespaceConfig(object):
     """
     zope.interface.implements(interfaces.IConfig)
 
-    def __init__(self, namespace):
+    def __init__(self, namespace, fake=False):
```
bmw: As stated above, I think these changes should be reverted.

CHANGES.rst
pde: Hah, I wanted to trigger a travis rerun of this branch without erasing the previous failure.  I intended make a one character change to a consequence-free file, but ended up realising the file needed love :)

bmw:
In general, I think/hope the approach here will work in the short term. I am not aware of any major problems with it and no new problems have been found since commits stopped landing a couple hours ago.

With that said, a couple hours of manual testing isn't very exhaustive and this feels very fragile, hacky, and difficult to maintain. As new flags and interactions between flags are added to the command line, we will have to remember to update this code and add more special cases.

I was originally told on Friday that this code should take 10 - 15 minutes to write and we just finished tracking down problems less than 48 hours before the release. The code's complexity and lack of testing makes me nervous.

In the short term, we may want this. Out of the other possibilities we suggested, this is unfortunately the only one that is even implemented 24 hours before we're planning on releasing 0.4.0. In the long term, however, I think we should readdress this problem and come up with a solution that doesn't rely on us remembering to maintain it.

letsencrypt/cli.py
pde: Yes, I actually removed that wrapper for a while, but it seemed as though maybe the derived property of `NamespaceConfig` were relevant.  It's also possible that I misattributed other bugs to the lack of a `NamespaceConfig` layer.

letsencrypt/cli.py
pde: Yes, though I'd prefer to keep the special case totally separate from the `return False` base case for legibility.

letsencrypt/cli.py
pde: Yes, it will require some (fairly minimal) maintenance until we can find a way to refactor it into something cleaner, though short of adding extra internal plumbing to argparse / configarparse, I'm not sure what a cleaner answer is.

pde: I'm adding a note to `interfaces.py` so that plugin authors at least have documentation of this situation.

pde: Hmmm, turns out that part of the plugins API is actually defined in `common/plugins.py`, so updating the docs to try to lead people to it...

pde:
I agree with all your cautions. Mulling this overnight, I think there a couple of things we could do to reduce fragility here:

1. Every time `_set_by_cli` returns `False`, it could double check that the argument flag isn't present in `sys.argv`.  That isn't a 100% reliable detector of special cases we missed (since `"-v" in ["-vvv", "certonly"]` returns `False`) but it will print a warning for us to catch fairly quickly.

2. We remove the `account` special case by no longer importing `account` from the renewal conf file. It really isn't clear that we need to do that.

If we like these solutions, we could land them for the next release.

PR #2392, opened 2016-02-06
Reporting output from the renew verb

schoen:
This isn't necessarily exactly the right format for this information, particularly to make it as conveniently machine-readable as possible (because a goal of outputting this is to let shell scripts grep for it in the `letsencrypt renew` output). I did a renewal with both `webroot` and `apache` certs in the same tree and confirmed that it printed factually correct notifications for each of them, so I believe the logic is right and the question is just what the output message needs to look like.

I also noticed that there is a generally high level of inconsistency in the kind of output that the renew verb can give about different situations. In my experiment I had `webroot`, `apache`, and `standalone` certs all present on the same system and attempted to renew all three. The result is: debug prints about attempting each renewal, this PR's notifications for success of `webroot` and `apache` certs, a very differently-formatted set of errors about the `standalone` renewal failure (due to a port in use), and then finally the "Congratulations" messages for the success of the first two. I fear that a user would have to spend a fair amount of time puzzling over this to understand that the upshot is "your first two worked and the third one didn't and the first two had links updated and Apache was reloaded".


pde:
OK, it sounds like we should turn off the "Congratulations" notification, and instead have `renew` do its own notification, which is something like 
* `"Congratulations, all of your renewals succeeded" (list of fullchain.pem paths & domains)` or
* `"Some renewals succeed" (list of fullchain.pem paths & domains) "\n but others failed" (list of fullchain.pem paths & domains)"` or
*  `"all renewal attempts failed" (list of fullchain.pem paths & domains)`.

schoen:
Currently trying to deal with the distinction between `_auth_from_domains` deciding not to renew at all and a renewal success.

schoen:
That seems to work, so I'm now working on making the output more machine-readable.

letsencrypt/cli.py
```diff
@@ -704,12 +706,18 @@ def obtain_cert(config, plugins, lineage=None):
...
+        if installer is None:
+            # Tell the user that the server was not restarted.
+            print("new certificate deployed without reload, fullchain is",
+                  lineage.fullchain)
+        else:
+            # In case of a renewal, reload server to pick up new certificate.
+            # In principle we could have a configuration option to inhibit this
+            # from happening.
+            installer.restart()
+            print("new certificate deployed with reload of plugin",
```
pde: Perhaps `"new certificate deployed with reload of ", config.installer.name, " server; fullchain is ", lineage.fullchain`

letsencrypt/cli.py
```diff
@@ -871,6 +879,30 @@ def _renewal_conf_files(config):
...
 
 
+def _renew_describe_results(renew_successes, renew_failures, parse_failures):
+    print()
+    if not renew_successes and not renew_failures:
+        print("No renewals were attempted.")
+    elif renew_successes and not renew_failures:
+        print("Congratulations, all renewals succeeded. The following certs "
+              "have been renewed:")
+        print("\t" + "\n\t".join(x + " (success)" for x in renew_successes))
```
pde: Make these complex, repeated calls into a lambda or helper function

letsencrypt/cli.py
```diff
@@ -901,28 +936,42 @@ def renew(config, unused_plugins):
...
-                # (or successful).
-                obtain_cert(lineage_config,
-                            plugins_disco.PluginsRegistry.find_all(),
-                            renewal_candidate)
+                # Although obtain_cert itself also indirectly decides
+                # whether to renew or not, we need to check at this
+                # stage in order to avoid claiming that renewal
+                # succeeded when it wasn't even attempted (since
+                # obtain_cert wouldn't raise an error in that case).
+                if _should_renew(lineage_config, renewal_candidate):
```
pde: This is a bad way to do this; attempting to perform the check twice creates the possibility that you'll somehow reach a different decision about whether to renew from the two call sites.  Can't we just keep records within `obtain_cert`?

pde: Since you're actually not calling `obtain_cert` in the case where `_should_renew` returns `False`, we shoouuld at least be consistent.  But why not then remove the call to `_should_renew` inside `obtain_cert` in the `lineage != None` case, and simply set `action = renew` there?

letsencrypt/cli.py
```diff
@@ -871,6 +879,30 @@ def _renewal_conf_files(config):
     return glob.glob(os.path.join(config.renewal_configs_dir, "*.conf"))
 
 
+def _renew_describe_results(renew_successes, renew_failures, parse_failures):
```
pde: This function completely controls the output of `letsencrypt renew --dry-run`, so it should print slightly different messages in that case.

letsencrypt/cli.py
```diff
@@ -871,6 +879,30 @@ def _renewal_conf_files(config):
...
+    elif renew_failures and not renew_successes:
+        print("All renewal attempts failed. The following certs could not be "
+              "renewed:")
+        print("\t" + "\n\t".join(x + " (failure)" for x in renew_failures))
+    elif renew_failures and renew_successes:
+        print("The following certs were successfully renewed:")
+        print("\t" + "\n\t".join(x + " (success)" for x in renew_successes))
+        print("\nThe following certs could not be renewed:")
+        print("\t" + "\n\t".join(x + " (failure)" for x in renew_failures))
+
```
pde: Also print a batch of "not yet due for renewal" outputs.

letsencrypt/cli.py
```diff
@@ -886,6 +918,9 @@ def renew(config, unused_plugins):
                            "specifying a CSR file. Please try the certonly "
                            "command instead.")
     renewer_config = configuration.RenewerConfiguration(config)
+    renew_successes = []
+    renew_failures = []
+    parse_failures = []
     for renewal_file in _renewal_conf_files(renewer_config):
         print("Processing " + renewal_file)
         # XXX: does this succeed in making a fully independent config object
```
pde: Yes, it does, you can remove this comment.

pde:
Address comments, fix tests?

letsencrypt/cli.py
```diff
@@ -418,7 +419,8 @@ def _suggest_donation_if_appropriate(config):
 
 def _report_successful_dry_run():
     reporter_util = zope.component.getUtility(interfaces.IReporter)
-    reporter_util.add_message("The dry run was successful.",
+    reporter_util.add_message("A test certificate requested in dry run was "
```
pde: Will this message be printed many times if `renew` renews many certs?

schoen: > Will this message be printed many times if `renew` renews many certs?

Yes (it already is being as a result of the dry_run PR).

letsencrypt/cli.py
```diff
@@ -672,11 +676,18 @@ def run(config, plugins):  # pylint: disable=too-many-branches,too-many-locals
...
+        # supplied, check if CSR matches given domains?
+        return "--domains and --csr are mutually exclusive"
+
     try:
         # installers are used in auth mode to determine domain names
         installer, authenticator = choose_configurator_plugins(config, plugins, "certonly")
     except errors.PluginSelectionError as e:
-        return e.message
+        logger.info(
+            "Could not choose appropriate plugin: %s", e)
```
pde: What is this change for?

schoen: > What is this change for?

To allow this to raise exceptions instead of returning errors as strings.

letsencrypt/tests/cli_test.py
```diff
@@ -222,13 +222,16 @@ def test_configurator_selection(self, mock_exe_exists):
         if "nginx" in real_plugins:
             # Sending nginx a non-existent conf dir will simulate misconfiguration
             # (we can only do that if letsencrypt-nginx is actually present)
-            ret, _, _, _ = self._call(args)
-            self.assertTrue("The nginx plugin is not working" in ret)
-            self.assertTrue("MisconfigurationError" in ret)
+            self._call(args)
+            # XXX: This probably now raises an exception (when nginx is
+            #      present, but I don't know which one!)
+            # self.assertTrue("The nginx plugin is not working" in ret)
```
pde: It's `errors.PluginSelectionError`. That was pretty easy to determine by doing `git grep "plugin is not working"`, and reading that code.

pde: but you can also map these to:
```
try:
    self._call(args)
except errors.Error as error:
    self.assertTrue("Test string" in e.message)
```

schoen: I tried locally changing it to `self.assertRaises(errors.PluginSectionError, ...)` and it didn't pass `tox`. (Nor did `errors.MisconfigurationError`, for that matter.) I think I tried that before with the same result, and that's why I didn't know what error it should be.

pde: I suspect that's because it's not actually raising an error, which is a pretty serious violation of our assumptions :)

pde:
Merged after a re-read and test with Seth.

Issue #2377, opened 2016-02-05
Wrong error message when not specifying a domain for certonly webroot 


johnl:
When I specify a webroot path but don't specify a domain, the error message says I didn't specify a webroot, which is misleading:

```
ubuntu@srv-1gybf:~/letsencrypt$ ./letsencrypt-auto certonly --webroot -w /home/ubuntu/app/current/public 
Updating letsencrypt and virtual environment dependencies......
Requesting root privileges to run with virtualenv: sudo /home/ubuntu/.local/share/letsencrypt/bin/letsencrypt certonly --webroot -w /home/ubuntu/app/current/public
The webroot plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('--webroot-path must be set',)
```

When I specify the domain, it works of course:
```
ubuntu@srv-1gybf:~/letsencrypt$ ./letsencrypt-auto certonly --webroot -w /home/ubuntu/app/current/public  -d example.com
Updating letsencrypt and virtual environment dependencies......
Requesting root privileges to run with virtualenv: sudo /home/ubuntu/.local/share/letsencrypt/bin/letsencrypt certonly --webroot -w /home/ubuntu/app/current/public -d example.com

IMPORTANT NOTES:
 - Congratulations! Your certificate and chain have been saved at
   /etc/letsencrypt/live/example.com/fullchain.pem. Your cert will expire
   on 2016-05-05. To obtain a new version of the certificate in the
   future, simply run Let's Encrypt again.
 - If you like Let's Encrypt, please consider supporting our work by:

   Donating to ISRG / Let's Encrypt:   https://letsencrypt.org/donate
   Donating to EFF:                    https://eff.org/donate-le
```

This is with master branch, commit 70ab18203d5681c07ef9c0d64238b6a74bbf31fb

bmw:
Related #1668 and #2309.

PR #2393, opened 2016-02-06
Allow webroot-map and --csr to exist together.

pde:
Further addressing: #1668.


pde:
Also closes: #2377

pde:
Merging with a "/looks/ ok" in IRC from @SwartzCr plus at least one manual invocation succeeding :)

PR #2394, opened 2016-02-06
Pass the SAN list as a string object, not a unicode object

thefinn93:
I was running into [this](https://github.com/letsencrypt/letsencrypt/pull/2309#issuecomment-177172587) issue, so I converted it to a string as the issue suggested and it renewed successfully. I think i tested everything sufficiently, but not 100% sure?

### DIFF
--- a/letsencrypt/crypto_util.py
+++ b/letsencrypt/crypto_util.py
@@ -115,7 +115,7 @@ def make_csr(key_str, domains):
         OpenSSL.crypto.X509Extension(
             "subjectAltName",
             critical=False,
-            value=", ".join("DNS:%s" % d for d in domains)
+            value=str(", ".join("DNS:%s" % d for d in domains))
         ),
     ])
     req.set_pubkey(pkey)


pde:
Hi! Are you sure this is still needed?  Five days ago we merged [this PR](https://github.com/letsencrypt/letsencrypt/pull/2336/files#diff-6f1cbade75baf6af63a7abf922aa9b40R311) into git master which should have made a second conversion to ASCII unnecessary.

If it is still necessary, perhaps there's somewhere we should be calling `enforce_domain_sanity()` but aren't?

thefinn93:
Ah, perhaps not! I've was testing off whatever's in pip, maybe that just hasn't been updated?

pde:
@thefinn93 yeah pip has the 0.3.0 release, you can [follow these instructions](https://letsencrypt.readthedocs.org/en/latest/contributing.html#running-a-local-copy-of-the-client) to run the code from git master.

Please reopen if this (or more calls to `ensure_domain_sanity`) are still needed.

PR #2395, opened 2016-02-07
Use pythondialog if on Python 3 instead of python2-pythondialog.

thanatos:
Since the latter is a backport of the former, for Python 2 only.

(with this change, letsencrypt cleanly installs for me. Doesn't work yet, of course, but one step at a time.)


bmw:
`certbot` currently does not support Python 3 (see [system requirements](https://github.com/certbot/certbot#system-requirements)). While I'm personally interested in changing this in the future, the code base isn't Python 3 compliant so a change like this is a little premature. With this in mind, I'm closing the PR.

Regardless, thanks for writing it up!

PR #2397, opened 2016-02-07
Fix import ordering s.t. future versions of pylint won't warn on it.

thanatos:
pylint gets more strict about import order in the future, so adjust it now to make upgrading smoother. pylint is following the order from PEP-8:

> Imports should be grouped in the following order:
>
> 1. standard library imports
> 2. related third party imports
> 3. local application/library specific imports
>
> You should put a blank line between each group of imports.

### DIFF
--- a/letsencrypt/plugins/common.py
+++ b/letsencrypt/plugins/common.py
@@ -1,11 +1,11 @@
 """Plugin common functions."""
 import os
-import pkg_resources
 import re
 import shutil
 import tempfile
 
 import OpenSSL
+import pkg_resources
 import zope.interface
 
 from acme.jose import util as jose_util
--- a/letsencrypt/plugins/disco_test.py
+++ b/letsencrypt/plugins/disco_test.py
@@ -1,8 +1,8 @@
 """Tests for letsencrypt.plugins.disco."""
-import pkg_resources
 import unittest
 
 import mock
+import pkg_resources
 import zope.interface
 
 from letsencrypt import errors
--- a/letsencrypt/tests/storage_test.py
+++ b/letsencrypt/tests/storage_test.py
@@ -1,13 +1,13 @@
 """Tests for letsencrypt.storage."""
 import datetime
-import pytz
 import os
-import tempfile
 import shutil
+import tempfile
 import unittest
 
 import configobj
 import mock
+import pytz
 
 from letsencrypt import configuration
 from letsencrypt import errors


kuba:
Erm, that doesn't look good to me. Does pylint really does not respect import groups?

thanatos:
I'm not sure what you mean by "does not respect import groups".

Perhaps see [this test output](https://travis-ci.org/letsencrypt/letsencrypt/jobs/107641632#L538) from another PR where I tried upgrading pylint. It seems like presently the project operates with a zero lint errors policy, so I was trying to stick that that, and this was easy to fix.

Specifically, in that linked output, pylint is warning that the imports, as written in those files, don't match PEP-8. (And it is correct; the imports do not.)

kuba:
We bunch together imports into the following groups:
- stdlib, such as `tempfile` or `shutil`
- third party, such as `OpenSSL`, `mock`
- our own project: `from letsencrypt....`

Each group is separated by a whitespace. Sometimes we add additional groups.

That being said, `pkg_resources` is actually a part of setuptools, so we might want to move it to third party import group across the codebase...

thanatos:
> We bunch together imports into the following groups:

This is the order used by PEP-8 and `pylint`

> That being said, `pkg_resources` is actually a part of setuptools

Which is why — I think — the future version of `pylint` nits on it. (I consider it a third-party import for the contexts of PEP-8 myself, as well.)

> so we might want to move it to third party import group across the codebase...

Hmm. My initial assumption was that I had them all. (I presumed that if I was missing any, `pylint` would have errored on them.) Grepping the codebase seems to indicate that this wasn't the case. I can attempt a more thorough sweep, if you want.

Perhaps for context: I was originally trying to make some sections of the code Python 3 compatible. To this end, I needed something from `six`; it seems like the present version of `pylint` errors on `from six.moves import <thing>`, because it isn't able to introspect into `moves`. (I guess it statically analyzes it, because actual introspection should work; due to its nature, `six.moves` isn't a "real" module.) At any rate, this seems to be "fixed" (via exception) in the later pylints, so I attempted to upgrade pylint. The newer pylint is more strict, and thus builds fail. This PR is meant to correct a subset of those newly introduced failures.

The rest of the codebase seesm to just `# pylint: disable-warning=that-import-error`, and I ended up doing that in the relevant PR, but figured this change wasn't bad either, though admittedly cosmetic.

pde:
OK!

PR #2401, opened 2016-02-08
Fix ACME error message help

### LINKED ISSUE #2365: webroot failure gives a not-very-useful error
If webroot renewal (or probably initial auth) fails due to the lack of a running web server that's pointed at that webroot, the error message returned to the user is very generic and doesn't really mention the possibility that a suitable web server was unavailable. Users might find it hard to diagnose this problem on this basis.

bmw:
Fixes #2365 as well as the regression of never showing the extra help text which apparently has existed for quite a while :frowning:

### DIFF
--- a/letsencrypt/auth_handler.py
+++ b/letsencrypt/auth_handler.py
@@ -480,6 +480,9 @@ def is_preferred(offered_challb, satisfied,
     return True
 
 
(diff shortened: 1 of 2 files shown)


pde:
LGTM.

PR #2402, opened 2016-02-09
Revert "Allow webroot-map and --csr to exist together."

pde:
Will revert letsencrypt/letsencrypt#2393 if I can't fix it faster than travis tests this revert...


schoen:
The original merge broke some test stuff related to `letsencrypt renew` success/failure notifications (in #2392) so I'm happy to see it revised if possible (although I'm not sure that's the exact problem that caused @pde to want to revert it).

PR #2403, opened 2016-02-09
Webroot map and flags

pde:
Further fixes to this branch, as an alternative to reverting it..

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1114,23 +1114,23 @@ class HelpfulArgumentParser(object):
 
     def handle_csr(self, parsed_args):
         """
-        Process a --csr flag. This needs to happen early enought that the
+        Process a --csr flag. This needs to happen early enough that the
         webroot plugin can know about the calls to _process_domain
         """
         try:
             csr = le_util.CSR(file=parsed_args.csr[0], data=parsed_args.csr[1], form="der")
             typ = OpenSSL.crypto.FILETYPE_ASN1
             domains = crypto_util.get_sans_from_csr(csr.data, OpenSSL.crypto.FILETYPE_ASN1)
-        except:
+        except OpenSSL.crypto.Error:
             try:
                 e1 = traceback.format_exc()
                 typ = OpenSSL.crypto.FILETYPE_PEM
                 csr = le_util.CSR(file=parsed_args.csr[0], data=parsed_args.csr[1], form="pem")
                 domains = crypto_util.get_sans_from_csr(csr.data, typ)
-            except:
+            except OpenSSL.crypto.Error:
                 logger.debug("DER CSR parse error %s", e1)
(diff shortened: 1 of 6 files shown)


letsencrypt/cli.py
```diff
@@ -1106,8 +1107,47 @@ def parse_args(self):
...
                 parsed_args.break_my_certs = parsed_args.staging = True
 
+        if parsed_args.csr:
+            self.handle_csr(parsed_args)
+
         return parsed_args
 
+    def handle_csr(self, parsed_args):
+        """
+        Process a --csr flag. This needs to happen early enought that the
```
schoen: should be "enough"

letsencrypt/cli.py
```diff
@@ -1106,8 +1107,47 @@ def parse_args(self):
...
+        webroot plugin can know about the calls to _process_domain
+        """
+        try:
+            csr = le_util.CSR(file=parsed_args.csr[0], data=parsed_args.csr[1], form="der")
+            typ = OpenSSL.crypto.FILETYPE_ASN1
+            domains = crypto_util.get_sans_from_csr(csr.data, OpenSSL.crypto.FILETYPE_ASN1)
+        except:
```
schoen: pylint doesn't like this broad except and needs a pylint ignore if you're sure of it. Same thing with the except at line 1130. I wonder whether this could produce a "parse error" when the CSR file simply can't be read or was wrongly specified? Like if you say `--csr my_crs.pem` instead of `--csr my_csr.pem`?

letsencrypt/cli.py
```diff
@@ -1106,8 +1107,47 @@ def parse_args(self):
...
+            raise errors.Error(
+                "Unfortunately, your CSR %s needs to have a SubjectAltName for every domain"
+                % parsed_args.csr[0])
+        for d in domains:
+            _process_domain(parsed_args, d)
+        parsed_args.actual_csr = (csr, typ)
+        csr_domains, config_domains = set(domains), set(parsed_args.domains)
+        if csr_domains != config_domains:
+            raise errors.ConfigurationError(
+                "Inconsistent domain requests:\ncsr: {0}\ncli config: {1}"
```
schoen: The user might not know what `cli config` means here (even though it's fairly clear from context); maybe refer to it another way? "Domains specified on command line"?

schoen:
I'm not really familiar with the problem that this PR aims to solve, but it looks good to me apart from the things that I mentioned.

schoen:
As I mentioned on IRC, testing this (with `letsencrypt renew`) didn't work properly for me; I got a ton of errors like

`2016-02-09 04:10:15,727:WARNING:letsencrypt.cli:Missing default analysis for 'tls_sni_01_port'`

and it seems that `--rsa-key-size` is ignored (it doesn't affect the key size chosen for renewal and doesn't change the renewal configuration file).

PR #2398, opened 2016-02-08
better integration testing for renewer (builds on #2388)

pde:

PR #2404, opened 2016-02-09
Renew integration tests v2

bmw:
Uses pieces from #2398 but builds off of master instead of #2388.

### DIFF
--- a/tests/boulder-integration.sh
+++ b/tests/boulder-integration.sh
@@ -51,8 +51,6 @@ common --domains le3.wtf install \
        --cert-path "${root}/csr/cert.pem" \
        --key-path "${root}/csr/key.pem"
 
-echo round 1
-
 CheckCertCount() {
     CERTCOUNT=`ls "${root}/conf/archive/le.wtf/"* | wc -l`
     if [ "$CERTCOUNT" -ne "$1" ] ; then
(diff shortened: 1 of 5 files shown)


tests/boulder-integration.sh
```diff
@@ -51,21 +50,27 @@ common --domains le3.wtf install \
...
-ls "$root/conf/archive/le1.wtf"
-# dir="$root/conf/archive/le1.wtf"
-# for x in cert chain fullchain privkey;
-# do
-#     latest="$(ls -1t $dir/ | grep -e "^${x}" | head -n1)"
-#     live="$($readlink -f "$root/conf/live/le1.wtf/${x}.pem")"
-#     [ "${dir}/${latest}" = "$live" ]  # renewer fails this test
-# done
+# This will renew because the expiry is less than 10 years from now
+sed -i "4arenew_before_expiry = 10 years" "$root/conf/renewal/le.wtf.conf"
```
pde: If this branch is genuinely working, I'd suggest going back to `--renew-by-default` instead of editing the renewal config file; that causes all the lineages to renew and generally tests things more.

pde:
Otherwise LGTM.

tests/boulder-integration.sh
schoen: > If this branch is genuinely working, I'd suggest going back to `--renew-by-default` instead of editing the renewal config file; that causes all the lineages to renew and generally tests things more.

I was hoping that there would be some test of the ability to get renewal
intervals out of the configuration file (it doesn't have to be the only
test!).

bmw: Lines 67/68 use `--renew-by-default` and these lines modify the renewal config file. With this in mind, I think leaving these lines as is a better test than using `--renew-by-default` again.

Perhaps we should extend `CheckCertCount` and have it check other lineages. Then we can verify all certs are renewed with `--renew-by-default` and only `le.wtf` is renewed here.

PR #2412, opened 2016-02-09
added a missing space that was causing an error in letsencrypt-auto [needs revision]

gicappa:
I'm a n00b and I just tried to launch the help script on OSX with zsh. 
The output was an error because of this missing space.
HTH


pde:
Thanks @gicappa! Can you also make this change to `letsencrypt-auto-source/letsencrypt-auto.template`?

gicappa:
done!

pde:
Thanks!

PR #2417, opened 2016-02-09
Deprecate --apache-ctl

### LINKED ISSUE #2414: Deprecate apache-ctl argument
`--apache-ctl` argument was removed. This argument should still be added as a deprecated argument so it still works for people.

bmw:
fixes #2414


pde:
Looks good.

Issue #2411, opened 2016-02-09
Test #2388 in integration tests


bmw:
If #2388 lands, we should probably test it in integration tests to make sure it works and prevent regressions.

pde:
There was some stuff in #2398 to do that; we could fish it out...

bmw:
It looked like [these lines](https://github.com/letsencrypt/letsencrypt/pull/2398/files#diff-c74d3ad4d86d0804f64e9354fa460b94R84) started to test that, but since a different key size was never set anywhere, I figured it was never finished.

If I'm missing something, feel free to reopen #2398 if you feel that's appropriate.

pde:
The test [was there](https://github.com/letsencrypt/letsencrypt/commit/f37908eadc61d6ab90a25174119cbb4a31fa290a) but was pulled out while trying to identify the source of the -oldest breakage.

PR #2418, opened 2016-02-10
Test cli flag setting within renewal integration tests

pde:
Closes: #2411


bmw:
lgtm. merging when green

pde:
This is failing integration tests.  I think maybe that `NamespaceConfig` wrapper is necessary, because re-adding it locally causes integration tests to succeed.  Trying in travis.

bmw:
That was my fault. I missed another change when removing the namespace. Just tested this on successfully on a VPS.

PR #2421, opened 2016-02-10
Better command-line docs for renew.

pde:


letsencrypt/cli.py
```diff
@@ -1490,6 +1495,16 @@ def prepare_and_parse_args(plugins, args):
         help="Require that all configuration files are owned by the current "
              "user; only needed if your config is somewhere unsafe like /tmp/")
 
+    helpful.add_group(
+        "renew", description="The 'renew' subcommand will attempt to renew all "
```
schoen: Depending on the meaning of "attempt to renew", this isn't quite right because it only actually tries to renew ones that are near expiry. I think this distinction could confuse people!

How about something like "you have previously obtained and that are near to expiry"?

schoen:
LGTM.

PR #2423, opened 2016-02-10
Offline sigs are actually made with sha256

pde:


schoen:
Keen!

schoen:
I merged without Travis check because this was a local-only change.

PR #2428, opened 2016-02-11
Include centos-options-ssl-apache.conf in letsencrypt-apache

bmw:


bmw:
I build test packages and can confirm this fixes the problem.

PR #2429, opened 2016-02-11
Fix broken reference in venv script to old bootstrap directory

MinnSoe:
The `venv` scripts were still pointing to the old bootstrap directory. This has been fixed in this PR. Seems to have been missed out in PR #2383.


pde:
Thanks!

PR #2435, opened 2016-02-11
letstest: work with a local repo, even if it isn't called letsencrypt

pde:
* allows test farm tests to be run from release branches


pde:
Not user-facing, safe to self-merge.

PR #2436, opened 2016-02-11
Fix venv.sh for the bootstrap/less future

pde:


schoen:
Duplicates #2429? (That one also modifies `venv3.sh` in the same way.)

Under our review rules you could probably merge #2429 yourself because it isn't your PR. :-)

PR #2438, opened 2016-02-11
Some preliminary documentation updates to mention renew verb

schoen:
This isn't really detailed but @bmw pointed out that we really do want to point to the renew verb in all of our documentation now that it's there, because people will start looking! So this is a first cut at doing that so that something useful will be there when people look.


bmw:
[Log](http://paste.ubuntu.com/15014636/) of integration test failure. Kicking Travis.

@jsha, we get these boulder errors from time to time. It's pretty rare, but would you like us to send logs of this problem your way?

jsha:
Yeah, please do! Thanks.

bmw:
lgtm

PR #2449, opened 2016-02-11
Support system-default Apache on OS X. Tested on Yosemite (10.10).

nneonneo:
This patch adds a configuration for the Apache plugin to support OS X's built-in Apache web server.

Note that it still requires the user to enable SSL (including httpd-ssl.conf) and generate a snakeoil CA cert to support the DVSNI check. In addition, the user has to add a vhost configuration to `/etc/apache2/other`.


pde:
Thanks, LGTM!

PR #2450, opened 2016-02-11
If le-auto's installation fails, delete the venv.

### LINKED ISSUE #2332: Can't import _io (requires deleting old virtualenv)
Someone on the community forums came across a problem with Python refusing to import `_io` a while ago:

https://community.letsencrypt.org/t/cant-use-letsencrypt-auto-importerror-no-module-named--io/2345/3

The solution required deleting and recreating the virtualenv. Unfortunately, people are still running into this; someone reported #2329 today which was exactly the same problem (with the same solution). Why does this happen, and how can we make it so that people don't have to recreate their virtualenvs?

erikrose:
Fix #2332.

Leaving broken venvs around can, if it got as far as installing the venv/bin/letsencrypt script, wreck future le-auto runs, since the presence of that script means "a working LE is installed" to it. (It was pretty unlikely this would happen, as LE is among the last things installed by peep, but it was possible.) Waiting until a new version of le-auto comes out and running it would recover, but this lets re-running the same version recover as well.


pde:
LGTM, but will need a manual merge.

PR #2452, opened 2016-02-11
Use reserved examples

### LINKED ISSUE #1713: Use of third-party domains in examples considered harmful

bmw:
fixes #1713 and supersedes #1716 by incorporating @chriscroome's comments

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,11 +51,11 @@ client will guide you through the process of obtaining and installing certs
 interactively.
 
 You can also tell it exactly what you want it to do from the command line.
-For instance, if you want to obtain a cert for ``thing.com``,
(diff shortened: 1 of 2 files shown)


bmw:
@jsha: [log](http://paste.ubuntu.com/15020531/) of boulder integration test failure.

Kicking travis.

pde:
OK

PR #2453, opened 2016-02-12
Remove werkzeug by parsing Retry-After ourselves

### LINKED ISSUE #2409: Yum install fails with RHEL v7.2
yum installation of letsencrypt.noarch and python2-letsencrypt.noarch fails when using RHEL's EPEL repository. The issue is the result of python-werkzeug - which is a dependency of both letsencrypt packages, but is not provided by EPEL. Manually installing the `python-werkzeug-0.9.1-2.el7.noarch.rpm` from the CentOS 7 Extras repository fixes the issue; installing Werkzeug by compiling from source or using pip does not resolve the issue, as yum will still list it as a failed dependency unless it sees the rpm has been installed.

Here is a list of all installed repos on the server where this fails:

Repo-id      : epel/x86_64
Repo-name    : Extra Packages for Enterprise Linux 7 - x86_64
Repo-revision: 1454890725
Repo-tags    : binary-x86_64
Repo-updated : Sun Feb  7 20:49:57 2016
Repo-pkgs    : 9,397
Repo-size    : 7.4 G
Repo-metalink: https://mirrors.fedoraproject.org/metalink?repo=epel-7&arch=x86_64
  Updated    : Sun Feb  7 20:49:57 2016
Repo-baseurl : http://s3-mirror-us-east-1.fedoraproject.org/pub/epel/7/x86_64/ (34 more)
Repo-expire  : 21,600 second(s) (last: Tue Feb  9 04:06:34 2016)
Repo-filename: /etc/yum.repos.d/epel.repo

Repo-id      : rhui-REGION-client-config-server-7/x86_64
Repo-name    : Red Hat Update Infrastructure 2.0 Client Configuration Server 7
Repo-revision: 1439271503
Repo-updated : Tue Aug 11 01:38:23 2015
Repo-pkgs    : 6
Repo-size    : 299 k
Repo-mirrors : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/mirror/rhui-client-config/rhel/server/7/x86_64/os
Repo-baseurl : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/repos//rhui-client-config/rhel/server/7/x86_64/os/ (1 more)
Repo-expire  : 21,600 second(s) (last: Tue Feb  9 04:06:34 2016)
Repo-filename: /etc/yum.repos.d/redhat-rhui-client-config.repo

Repo-id      : rhui-REGION-rhel-server-optional/7Server/x86_64
Repo-name    : Red Hat Enterprise Linux Server 7 Optional (RPMs)
Repo-revision: 1454007492
Repo-updated : Thu Jan 28 13:59:28 2016
Repo-pkgs    : 8,213
Repo-size    : 9.6 G
Repo-mirrors : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/mirror/content/dist/rhel/rhui/server/7/7Server/x86_64/optional/os
Repo-baseurl : https://rhui2-cds02.us-east-1.aws.ce.redhat.com/pulp/repos//content/dist/rhel/rhui/server/7/7Server/x86_64/optional/os/ (1 more)
Repo-expire  : 21,600 second(s) (last: Tue Feb  9 04:06:34 2016)
Repo-filename: /etc/yum.repos.d/redhat-rhui.repo

Repo-id      : rhui-REGION-rhel-server-releases/7Server/x86_64
Repo-name    : Red Hat Enterprise Linux Server 7 (RPMs)
Repo-revision: 1454005863
Repo-updated : Thu Jan 28 13:33:14 2016
Repo-pkgs    : 10,152
Repo-size    : 12 G
Repo-mirrors : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/mirror/content/dist/rhel/rhui/server/7/7Server/x86_64/os
Repo-baseurl : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/repos//content/dist/rhel/rhui/server/7/7Server/x86_64/os/ (1 more)
Repo-expire  : 21,600 second(s) (last: Tue Feb  9 04:06:34 2016)
Repo-filename: /etc/yum.repos.d/redhat-rhui.repo

Repo-id      : rhui-REGION-rhel-server-rh-common/7Server/x86_64
Repo-name    : Red Hat Enterprise Linux Server 7 RH Common (RPMs)
Repo-revision: 1447968073
Repo-updated : Thu Nov 19 16:21:13 2015
Repo-pkgs    : 168
Repo-size    : 3.2 G
Repo-mirrors : https://rhui2-cds01.us-east-1.aws.ce.redhat.com/pulp/mirror/content/dist/rhel/rhui/server/7/7Server/x86_64/rh-common/os
Repo-baseurl : https://rhui2-cds02.us-east-1.aws.ce.redhat.com/pulp/repos//content/dist/rhel/rhui/server/7/7Server/x86_64/rh-common/os/ (1 more)
Repo-expire  : 21,600 second(s) (last: Tue Feb  9 04:06:34 2016)
Repo-filename: /etc/yum.repos.d/redhat-rhui.repo

pde:
Fixes #2409
Progress on #1301

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -264,10 +264,10 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             when = parsedate_tz(retry_after)
             try:
                 year = when[0] # raises TypeError if t is None
-                # Handle two-digit years -- but any webserver that thinks
+                # py26: Handle two-digit years -- but any server that thinks
                 # "retry after 99" means "come back after 1999" is.. deprecated
                 if year >= 0 and year < 100:
-                    when = [year + 2000] + list(when[1:])
+                    when = [year + 2000] + list(when[1:]) #  pragma: no cover
                 tzone = when[-1] if when[-1] else 0
                 # raises ValueError/OverflowError
                 return datetime.datetime(*when[:7]) - datetime.timedelta(tzone)
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -258,7 +258,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         :rtype: `datetime.datetime`
 
         """
-        retry_after = response.headers.get('Retry-After', str(default)).strip()
+        retry_after = response.headers.get('Retry-After', str(default))
         try:
             seconds = int(retry_after)
         except ValueError:
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -4,7 +4,7 @@ import heapq
 import logging
 import time
 
-from datetime import datetime, timedelta
+import datetime
 
 import six
 from six.moves import http_client  # pylint: disable=import-error
@@ -21,7 +21,8 @@ from acme import messages
 try:
     from email.utils import parsedate_tz
 except ImportError: # pragma: no cover
-    from email.Utils import parsedate_tz
+    # pylint: disable=import-error,no-name-in-module
+    from email.Utils import parsedate_tz 
 
 logger = logging.getLogger(__name__)
 
@@ -257,7 +258,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             ``Retry-After`` header is not present or invalid.
 
         :returns: Time point when next `poll` should be performed.
-        :rtype: `datetime.datetime`
+        :rtype: `datetime.datetime.datetime`
 
         """
         retry_after = response.headers.get('Retry-After', str(default)).strip()
@@ -265,20 +266,20 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             seconds = int(retry_after)
         except ValueError:
             # pylint: disable=no-member
-            t = list(parsedate_tz(retry_after))
+            when = parsedate_tz(retry_after)
             try:
-                year = t[0] # raises TypeError if t is None
+                year = when[0] # raises TypeError if t is None
                 # Handle two-digit years -- but any webserver that thinks
                 # "retry after 99" means "come back after 1999" is.. deprecated
                 if year >= 0 and year < 100:
-                    t[0] = year + 2000
-                tz = t[-1] if t[-1] else 0
+                    when = [year + 2000] + when[1:]
+                tzone = when[-1] if when[-1] else 0
                 # raises ValueError/OverflowError
-                return datetime(*t) - timedelta(tz)
+                return datetime.datetime(*when[:7]) - datetime.timedelta(tzone)
             except (TypeError, ValueError, OverflowError):
                 seconds = default
 
-        return datetime.now() + timedelta(seconds=seconds)
+        return datetime.datetime.now() + datetime.timedelta(seconds=seconds)
 
     def poll(self, authzr):
         """Poll Authorization Resource for status.
@@ -368,9 +369,9 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         attempts = collections.defaultdict(int)
         exhausted = set()
 
-        # priority queue with datetime (based on Retry-After) as key,
+        # priority queue with datetime.datetime (based on Retry-After) as key,
         # and original Authorization Resource as value
-        waiting = [(datetime.now(), authzr) for authzr in authzrs]
+        waiting = [(datetime.datetime.now(), authzr) for authzr in authzrs]
         # mapping between original Authorization Resource and the most
         # recently updated one
         updated = dict((authzr, authzr) for authzr in authzrs)
@@ -378,7 +379,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         while waiting:
             # find the smallest Retry-After, and sleep if necessary
             when, authzr = heapq.heappop(waiting)
-            now = datetime.now()
+            now = datetime.datetime.now()
             if when > now:
                 seconds = (when - now).seconds
                 logger.debug('Sleeping for %d seconds', seconds)
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -1,10 +1,9 @@
 """ACME client API."""
 import collections
+import datetime
 import heapq
 import logging
 import time
-
-import datetime
 
 import six
 from six.moves import http_client  # pylint: disable=import-error
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -11,13 +11,16 @@ from six.moves import http_client  # pylint: disable=import-error
 import OpenSSL
 import requests
 import sys
-import werkzeug
 
 from acme import errors
 from acme import jose
 from acme import jws
 from acme import messages
 
+try:
+    from email.utils import parsedate_tz
+except ImportError: # pragma: no cover
+    from email.Utils import parsedate_tz
 
 logger = logging.getLogger(__name__)
 
@@ -245,7 +248,8 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
     @classmethod
     def retry_after(cls, response, default):
-        """Compute next `poll` time based on response ``Retry-After`` header.
+        """Compute next `poll` time based on response ``Retry-After`` header,
+        per https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37
 
         :param requests.Response response: Response from `poll`.
         :param int default: Default value (in seconds), used when
@@ -256,17 +260,26 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
         """
         retry_after = response.headers.get('Retry-After', str(default))
+        now = datetime.datetime.now()
+        year_now = now[0]
         try:
             seconds = int(retry_after)
         except ValueError:
             # pylint: disable=no-member
-            decoded = werkzeug.parse_date(retry_after)  # RFC1123
-            if decoded is None:
+            t = parsedate_tz(value.strip())
+            try:
+                year = t[0] # raises TypeError if t is None
+                # Handle two-digit years -- but any webserver that thinks
+                # "retry after 99" means "come back after 1999" is.. deprecated
+                if year >= 0 and year < 100:
+                    year += 2000
+                t_corrected = datetime(*([year] + t[1:7])) # raises ValueError
+                tz = t[-1] if t[-1] else 0
+                return t_corrected - timedelta(tz)      # raises OverflowError
+            except (TypeError, ValueError, OverflowError):
                 seconds = default
-            else:
-                return decoded
 
-        return datetime.datetime.now() + datetime.timedelta(seconds=seconds)
+        return now + datetime.timedelta(seconds=seconds)
 
     def poll(self, authzr):
         """Poll Authorization Resource for status.
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -262,6 +262,8 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         try:
             seconds = int(retry_after)
         except ValueError:
+            # The RFC 2822 parser handles all of RFC 2616's cases in modern
+            # environments (primarily HTTP 1.1+ but also py27+)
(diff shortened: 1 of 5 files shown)


pde:
Hmmm, my scripts for rsyncing + testing were not working for `acme/`; will reopen when this is actually ready...

pde:
May work now...

acme/acme/client.py
```diff
@@ -245,26 +245,34 @@ def answer_challenge(self, challb, response):
 
     @classmethod
     def retry_after(cls, response, default):
-        """Compute next `poll` time based on response ``Retry-After`` header.
+        """Compute next `poll` time based on response ``Retry-After`` header,
+        per https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37
```
kuba: First docstring paragraph should be limited to one line only, please move the link somewhere down.

acme/acme/client.py
```diff
@@ -245,26 +245,34 @@ def answer_challenge(self, challb, response):
...
 
         :param requests.Response response: Response from `poll`.
         :param int default: Default value (in seconds), used when
             ``Retry-After`` header is not present or invalid.
 
         :returns: Time point when next `poll` should be performed.
-        :rtype: `datetime.datetime`
+        :rtype: `datetime.datetime.datetime`
```
kuba: That's incorrect

kuba:
Rather than inlining all decoding code please write a drop-in replacement function for `werkzeug.parse_date` (with the same signature). Then write all necessary unit test cases for such function.

acme/acme/client.py
```diff
@@ -245,26 +245,34 @@ def answer_challenge(self, challb, response):
...
             # pylint: disable=no-member
-            decoded = werkzeug.parse_date(retry_after)  # RFC1123
-            if decoded is None:
+            when = parsedate_tz(retry_after)
+            try:
+                year = when[0] # raises TypeError if t is None
+                # py26: Handle two-digit years -- but any server that thinks
+                # "retry after 99" means "come back after 1999" is.. deprecated
+                if year >= 0 and year < 100:
+                    when = [year + 2000] + list(when[1:]) #  pragma: no cover
```
kuba: urm, slice operator (`when[1:]`) already produces list, please don't cast explicitly

pde: The cast is required for python 2.6.

kuba: No, cast is not required for Python 2.6. The problem here is that you're trying to concatenate list and tuple which won't work with any version of Python. You should probably use `when = (year + 2000,) + when[1:]`.

acme/acme/client.py
```diff
@@ -255,16 +258,23 @@ def retry_after(cls, response, default):
...
         """
-        retry_after = response.headers.get('Retry-After', str(default))
+        retry_after = response.headers.get('Retry-After', str(default)).strip()
         try:
             seconds = int(retry_after)
         except ValueError:
             # pylint: disable=no-member
-            decoded = werkzeug.parse_date(retry_after)  # RFC1123
-            if decoded is None:
+            when = parsedate_tz(retry_after)
```
kuba: According to the docs, that's using RFC 2822 (with some hacks), not RFC 2616.

acme/acme/client.py
```diff
@@ -255,16 +258,23 @@ def retry_after(cls, response, default):
...
+            try:
+                year = when[0] # raises TypeError if t is None
+                # py26: Handle two-digit years -- but any server that thinks
+                # "retry after 99" means "come back after 1999" is.. deprecated
+                if year >= 0 and year < 100:
+                    when = [year + 2000] + list(when[1:]) #  pragma: no cover
+                tzone = when[-1] if when[-1] else 0
+                # raises ValueError/OverflowError
```
kuba: try/except block should be limited to the smallest possible block of code, in this case line below

acme/acme/client.py
kuba: Also, why is this not covered? Please test it or write further inline explanation. Also, you should use `  # pragma: no cover` to satisfy linter (note double whitespace before hashtag, one after - opposite what you've done)

acme/acme/client.py
```diff
@@ -255,16 +258,23 @@ def retry_after(cls, response, default):
         :rtype: `datetime.datetime`
 
         """
-        retry_after = response.headers.get('Retry-After', str(default))
+        retry_after = response.headers.get('Retry-After', str(default)).strip()
```
kuba: if you care about stripping, then please add an appropriate test case

acme/acme/client.py
pde: It's covered in py26, but in py27 `parsedate_tz` never returns two-digit years.

acme/acme/client.py
pde: I haven't read the RFCs closely to figure out their relationships with each other. But `werkzeug.parse_date` is a thin wrapper on `email.utils.parsedate_tz`.

pde: In any case, `parsedate_tz` handles [each of the 2616 cases](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3.1).  However I'm going to remove any special casing in our code for the RFC 850 subcase.  No ACME servers will ever speak HTTP/1.0 or issue RFC 850 dates, so we don't need extra code to sure up python 2.6's implementation of that deprecated format.

acme/acme/client.py
pde: Done; I'm not sure if requests passes along trailing whitespace but the HTTP spec does seem to allow it.

pde:
This should be ready for re-review.

bmw:
@kuba, are you able to rereview this? pde really wants this in the next release. If not, I can take a crack at it but I'd much prefer you did the review if you're able/interested.

kuba:
Yup, I'll review this. When is the next release?

BTW, why is the milestone `0.4.1`? I thought we already agreed on the versioning scheme.

acme/acme/client.py
```diff
@@ -11,13 +11,13 @@
...
 import requests
 import sys
-import werkzeug
 
 from acme import errors
 from acme import jose
 from acme import jws
 from acme import messages
 
+from email.utils import parsedate_tz
```
kuba: stdlib imports should happen in the top-most import group

kuba:
My [comment about separate drop-in function and relevant unit tests](https://github.com/letsencrypt/letsencrypt/pull/2453#issuecomment-183904010) was never fixed. I'll review this once it's fixed.

acme/acme/client_test.py
```diff
@@ -205,6 +205,13 @@ def test_retry_after_invalid(self, dt_mock):
             datetime.datetime(2015, 3, 27, 0, 0, 10),
             self.client.retry_after(response=self.response, default=10))
 
+        # wrong date -> ValueError
```
kuba: move to separate test

acme/acme/client.py
```diff
@@ -259,12 +262,16 @@ def retry_after(cls, response, default):
...
-            else:
-                return decoded
+            # The RFC 2822 parser handles all of RFC 2616's cases in modern
+            # environments (primarily HTTP 1.1+ but also py27+)
+            when = parsedate_tz(retry_after)
+            if when is not None:
+                try:
+                    tz_secs = datetime.timedelta(when[-1] if when[-1] else 0)
+                    return datetime.datetime(*when[:7]) - tz_secs
+                except (ValueError, OverflowError):
```
kuba: again, this looks like too broad `try/except`

bmw:
@kuba, the next release is sometime this week. Preferably earlier rather than later.

And we did agree on a versioning scheme. We voted and decided to follow the semantic versioning scheme as you suggested. Initially, it was looking like this release would simply fix bugs in 0.4.0, meaning it should be called 0.4.1. If any new functionality is added, we'll call it 0.5.0.

pde:
@kuba, werkzeug's `parse_date` function only does two things on top of `email.utils.parsedate_tz`:

1. cast the timezone from seconds into something `datetime` understands
2. play weird games -- of increasingly questionable relevance and correctness -- with two-digit years for long-deprecated RFC 822/850 support.

I don't believe we should have (2) in our tree, and I don't currently believe that (1) is enough to warrant a separate function, though I'm open to arguments on either point!

acme/acme/client.py
pde: `OverflowError` can be raised by either the `timedelta` call or the subtract operation on the `datetime` objects.  `ValueError` can only be raised by the second line, but I believe the code as written is cleaner than two nested try blocks.

bmw:
Again, I'd prefer not to review this and leave that to @kuba.

With that said, @pde, looking quickly at kuba's review comments, I see [this](https://github.com/letsencrypt/letsencrypt/pull/2453#issuecomment-186565114) which was never fixed if I'm understanding the comment correctly.

pde:
@bmw my answer [here](https://github.com/letsencrypt/letsencrypt/pull/2453#issuecomment-187462269) was: I'm not sure it makes sense to have a drop-in identical replacement, but if someone has a good argument in favour I'm happy to do it.

acme/acme/client.py
```diff
@@ -4,6 +4,7 @@
 import heapq
 import logging
 import time
+from email.utils import parsedate_tz
```
bmw: nit: Our inputs are alphabetized.

acme/setup.py
```diff
@@ -20,7 +20,6 @@
     'requests',
     'setuptools',  # pkg_resources
     'six',
-    'werkzeug',
```
bmw: With this change we need to update `letsencrypt-auto-sources/pieces/letsencrypt-auto-requirements.txt` and remove this package. Otherwise, it will continue to be installed when people use `letsencrypt-auto`.

If you'd like this to be a separate PR, that's fine but you should open an issue/PR to track the issue to make sure it makes it into 0.4.1.

acme/acme/client.py
```diff
@@ -259,12 +262,16 @@ def retry_after(cls, response, default):
...
         except ValueError:
-            # pylint: disable=no-member
-            decoded = werkzeug.parse_date(retry_after)  # RFC1123
-            if decoded is None:
-                seconds = default
-            else:
-                return decoded
+            # The RFC 2822 parser handles all of RFC 2616's cases in modern
+            # environments (primarily HTTP 1.1+ but also py27+)
+            when = parsedate_tz(retry_after)
```
bmw: `werkzeug` and you in a previous version of this PR call `strip` on `retry_after` ([source](https://github.com/letsencrypt/letsencrypt/blob/6fd3dba737a8de2c2f2d43d52d932418762762bb/acme/acme/client.py#L259)). I think we should be doing this unless we are **certain** it isn't necessary.

acme/setup.py
pde: Thanks for spotting that!

acme/acme/client.py
pde: I'm certain.  In py26 and py27:
```
>>> int(" 3 ")
3
>>> from email.utils import parsedate_tz
>>> parsedate_tz(" Tue, 16 Feb 2016 11:50:00 MST ")
(2016, 2, 16, 11, 50, 0, 0, 1, -1, -25200)
```

bmw:
lgtm

kuba:
Indeed, my concerns were never addressed, although I pointed them out at
least 3/4 times (here and on IRC). I find it ridiculous to having justify
the need of good tests or good coding practices. My time is extremely
constrained recently, so apologies for slow or sparse reviews - I prefer to
save some time by reviewing only when all changes have been implemented,
which didn't happen here. I would also prefer being pinged at least (via
Hangouts) rather than skipping LGTM. That being said, i m almost sure this
change will cause troubles for clients... If I only had time I would
request for this change to be reverted and reimplemented properly...

PR #2461, opened 2016-02-12
dns-01 challenge and response

moshevds:
I am creating a dns-01 Authenticator plugin, and these DNS01+DNS01Response classes are needed for that. (With this patch my current plugin prototype works.)

So far I have omitted tests, DNS01Response.simple_verify() and the removal of DNS (old code).

Some questions:
 - Why do HTTP01 and TLSSNI01 have the .simple_verify() method? I see what they do, but only the manual HTTP01 authenticator uses it, and then only to possibly generate a warning. Creating this method for DNS01 would introduce a new dependency (dnspython seems best suited here), so I omited the method entirely.
 - Am I correct in assuming the DNS and DNSResponse classes are dead code? As best as I can tell, the type "dns" is removed from the spec and the way it worked is no longer used anyway.
 - I know unit tests are desired, but since I'm mostly focusing on the plugin itself, I would like a reaction before writing tests that have to be reworked later. I'll update the PR when I add tests.
 - My plugin uses DNS UPDATE, just like the ddns_auth branch by @ThomasWaldmann, but implemented using the dnspython library. Would there be any interest in having it in the main branch? As there is currently no officially supported DNS Authenticator, including it in this repository might make sense.

I would gladly hear comments on the code and these questions.

### DIFF
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -641,3 +641,32 @@ class DNSResponse(ChallengeResponse):
 
         """
         return chall.check_validation(self.validation, account_public_key)
+
+
+@ChallengeResponse.register
+class DNS01Response(KeyAuthorizationChallengeResponse):
+    """ACME dns-01 challenge response."""
+    typ = "dns-01"
+
+
+@Challenge.register  # pylint: disable=too-many-ancestors
+class DNS01(KeyAuthorizationChallenge):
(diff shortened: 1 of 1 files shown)


moshevds:
I'm closing this PR because PR #2061 seems to address the same issue. My questions are also relevant with that code in mind, but i'll just ask them in another location.

PR #1899, opened 2015-12-15
Quote bash variable expansion.

yig:
Variable expansion should be quoted.

PR #2464, opened 2016-02-12
Quote the remaining variable expansions in le-auto. Refs #1899.

erikrose:
...except for $SUDO, which is always either "sudo", "su_sudo", or "", never having a quote-needing char in it.

It's unlikely that $PYVER would have a space in it, but it doesn't hurt.


pde:
LGTM.

PR #2467, opened 2016-02-12
Reorder travis matrix by typical runtime

pde:
Since that makes most efficient use of travis's parallelism


pde:
This should be safe, self-merging.

PR #2141, opened 2016-01-10
Clean up dev/testing extras messup (fixes #2140)

kuba:

PR #2138, opened 2016-01-10
Enforce PEP8 checking in ACME [needs revision]

kuba:

PR #2469, opened 2016-02-13
Enforce PEP8 checking in ACME v2

bmw:
Simply merges master into #2138 so that tests will past now that #2141 has been merged.


bmw:
Closing since it looks like `acme` no longer passes `pep8` tests :(

PR #2477, opened 2016-02-15
[test farm] Fix use of the newleauto in test_letsencrypt_auto_certonl…

pde:
…y_standalone.sh


pde:
Not user facing, merging.

Issue #1143, opened 2015-10-26
Handle IPv6 adresses in Apache vhost entries


NoodlesNZ:
When running ```./letsencrypt-auto --agree-dev-preview --server https://acme-v01.api.letsencrypt.org/directory auth``` it crashes out with this:

```
Error while restarting Apache:
Restarting apache2 (via systemctl): apache2.service failed!

Job for apache2.service failed. See 'systemctl status apache2.service' and 'journalctl -xn' for details.
```
Upon further investigation via ```journalctl -xn``` I get this:

```
-- Logs begin at Tue 2015-09-08 20:40:09 CDT, end at Mon 2015-10-26 17:16:13 CDT. --
Oct 26 17:16:12 dag apache2[581]: AH00526: Syntax error on line 2 of /etc/apache2/le_dvsni_cert_challenge.conf:
Oct 26 17:16:12 dag apache2[581]: The address or port is invalid
Oct 26 17:16:12 dag apache2[581]: Action 'configtest' failed.
Oct 26 17:16:12 dag apache2[581]: The Apache error log may have more information.
Oct 26 17:16:12 dag systemd[1]: apache2.service: control process exited, code=exited status=1
Oct 26 17:16:12 dag systemd[1]: Failed to start LSB: Apache2 web server.
```

Unfortunately the file in question /etc/apache2/le_dvsni_cert_challenge.conf is not available after letsencrypt-auto crashes out.

Running on Debian jessie x86_64

jdkasten:
Yes, the configuration changes and challenge files are cleaned up/reverted in the result of a failure.  (Note for future work: We should have an available debug flag to avoid this and make debugging easier)

Looking at the template of the dvsni challenge file found here... https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/dvsni.py#L32

It looks like the syntax error probably occurred when trying to include the vhost address.  Given the ```address or port is invalid``` error.

Could you provide the VirtualHost block of the domain you selected to upgrade?

The challenge configuration file attempts to copy over the addresses of the selected VirtualHost while changing the port to 443.

NoodlesNZ:
This is the vhost I was trying to add let's encrypt to:
```
<VirtualHost 173.192.30.7:80 [2607:f0d0:1005:99::3:1337]:80>
DocumentRoot /xxxx/
ServerName noodles.net.nz
ServerAlias www.noodles.net.nz
[...]

PR #2478, opened 2016-02-15
parse IPv6 and IPv4 virtual host entries correctly (issue #1143) [needs tests]

TheBoegl:
d48c560 should fix the issue #1143 
b5bd330 adds a few missing test cases and moves them to the passing test-config folder. I hope this behaves as expected.


joohoi:
This needs tests, there's currently no unit tests hitting this.

letsencrypt/plugins/common.py
```diff
@@ -110,8 +110,17 @@ def __init__(self, tup):
...
     def fromstring(cls, str_addr):
         """Initialize Addr from string."""
-        tup = str_addr.partition(':')
-        return cls((tup[0], tup[2]))
+        if str_addr.startswith('['):
+            # ipv6 addresses starts with [
+            endIndex = str_addr.rfind(']')
+            host = str_addr[:endIndex + 1]
+            port = ''
+            if len(str_addr) > endIndex + 3 and str_addr[endIndex + 2] == ':':
```
joohoi: str_addr[endIndex] + 2] should never be ":", if the port is set, that'll be the first digit of the port nr.

TheBoegl:
I've corrected the index misstep and added a few unit tests like the ipv4 ones.

joohoi:
Some of the tests are failing because of the change. Please follow the steps on https://letsencrypt.readthedocs.org/en/latest/contributing.html#submitting-a-pull-request

TheBoegl:
@joohoi running a clean `tox` with #2518 results in passing tests. Can you confirm this?

joohoi:
Unfortunately I cannot. Two of letsencrypt-apache tests fail. You can run nosetests -v letsencrypt-apache within venv, here's the error output from my env:

```
ERROR: Make sure all vhosts are being properly found.
  File "/vagrant/letsencrypt-apache/letsencrypt_apache/tests/configurator_test.py", line 136, in test_get_virtual_hosts
    raise Exception("Missed: %s" % vhost)  # pragma: no cover
Exception: Missed: File: /tmp/tmpLfbStltemp/debian_apache_2_4/two_vhost_80/apache2/sites-available/000-default.conf
Vhost path: /files/tmp/tmpLfbStltemp/debian_apache_2_4/two_vhost_80/apache2/sites-available/000-default.conf/VirtualHost
Addresses: [2607:f0d0:1005:99::3:1337]:80, *:80
Name: ip-172-30-0-17
Aliases:
TLS Enabled: No
Site Enabled: Yes
mod_macro Vhost: No

FAIL: test_find_best_vhost_default (letsencrypt_apache.tests.configurator_test.TwoVhost80Test)
  File "/vagrant/letsencrypt-apache/letsencrypt_apache/tests/configurator_test.py", line 217, in test_find_best_vhost_default
    self.config._find_best_vhost("example.demo"), self.vh_truth[2])
AssertionError: <letsencrypt_apache.obj.VirtualHost object at 0x7f76c1417390> != <letsencrypt_apache.obj.VirtualHost object at 0x7f76c13f94d0>

Ran 141 tests in 25.071s

FAILED (errors=1, failures=1)
ERROR: InvocationError: '/vagrant/.tox/py27/bin/nosetests -v letsencrypt_apache'
```

joohoi:
The test fails actually indicate that we might have to do some modifications in find_best_vhost algo as well to land this.

joohoi:
I'm very sorry, the fails were because of my dirty test environment. 

However we'll still have to make sure the whole shebang works as intended even before we have IPv6 support as whole if we wish to land this PR before we have it. 

So I think we'll need to make sure no IPv6 only vhosts are picked up for the time being. Preferably wrapping that code in a way that's easy to revert when we have the actual support.

joohoi:
That being said, the code looks good to me, but we'll have to confirm the correct behaviour before we'll be able to support IPv6 for validation.

TheBoegl:
I'll take a look into that.
Furthermore, abbreviated IPv6 addresses should be saved in their exploded representation form for a correct equals-comparision.

joohoi:
@pde @bmw I'd love to hear your opinions on this IPv6 issue. I'm a bit on the edge where we should error out from validating IPv6 only vhost. This is mostly about the code health in the long run. If we do the checks in client, we have multiple options where to error out:

1) Do not list the vhost as available at all - needs a lot of checking in code all around, which can be cumbersome to remove when we have IPv6 support
2) Error out before challenge - this will interrupt the flow, and is harmful in a case where user wants to validate multiple domains, one of which being IPv6 only
3) Let the server fail the challenge - same downsides as above, and additionally generates unnecessary load on the servers.

Another option would be to ignore the IPv6 definitions altogether, which would then result into broken SSL vhosts generated by apache plugin.

pde:
@joohoi I don't have a categorical answer for you, but some ideas about each of your cases:

> 1) Do not list the vhost as available at all - needs a lot of checking in code all around, which can be cumbersome to remove when we have IPv6 support

One way to do this is to make a variable like `server_supports_ipv6`, and set it to `False` for now.

> 2) Error out before challenge - this will interrupt the flow, and is harmful in a case where user wants to validate multiple domains, one of which being IPv6 only

What do you mean by "interrupt the flow"?  If an error is coming anyway, the flow will be interrupted.  I also don't know how the user is going to be able to validate multiple domains if one is IPv6 only... unless you mean that IPv4 actually works for that domain, but Apache isn't configured to answer for it?

> 3) Let the server fail the challenge - same downsides as above, and additionally generates unnecessary load on the servers.

Don't worry about the load, I think boulder can handle it (or ask us to change if it's a problem).

pde:
Overall, I'd say that:

* If you think the IPv6-only vhosts will always cause errors right now, let's go with 1
* If it's possible that IPv6-only vhosts will succeed (because IPv4 actually exists and we can make a temporary IPv4 vhost and succeed in authentication) let's go with 3.

joohoi:
Thanks for the insights, I like the server_supports_ipv6 approach. Even though we'll have to clean up the code after we get IPv6 supported, it'll enable us to a) identify the relevant parts of the code, and b) gives us some time to do so.

With interrupting the flow I was thinking about an situation where user goes through 50 or so vhosts in the system, selects 20 of them, and proceeds. The first 16 vhosts are validated without problem, and then the 17th is one with only IPv6 configuration. The whole process fails, and user goes back to the beginning, forcing him/her to start the selection process from the beginning.

joohoi:
@TheBoegl I took the liberty to continue fixing the issue based on your work in #2690 

There are some use cases of IPv6 only vhosts resolving to IPv4 addresses, and hence being able to get the verification process right. For example using sniproxy / HAproxy endpoint in IPv4 address, that proxies the connection to IPv6 only host. This being the case, I am using the fail-at-boulder route in the continued work. This also makes us IPv6 ready in the client side.

TheBoegl:
@joohoi I'll check some of the test cases I added for this PR against yours and comment if some of them fail.

Issue #1003, opened 2015-10-16
Webroot plugin should delete .well-known/acme-challenge on cleanup


kuba:
`cleanup()` should delete `.well-known/acme-challenge` when all previously perfomed challenges have been cleaned up. This requires us to track the status of all incoming challenges, just like standalone2 does.

pde:
This might have been a somewhat involved task for a first PR, but since there's an implementation to work from, it shouldn't be too bad.

filipochnik:
I started working on this.
IMO `.well-known` should be removed as well if it wasn't present before. We want to leave everything as it was. Is that correct?

hlieberman:
I think it should be removed regardless of whether it was present before.  We're going to blindly overwrite what was there anyway, so there doesn't seem to be much point to keeping it around even if it was present before.

filipochnik:
Maybe I'm not seeing something, but aren't we writing only to `.well-known/acme-challenge`? Seem like all other files and directories in `.well-known` are left untouched.

hlieberman:
On February 16, 2016 12:35:44 AM EST, Filip Ochnik <notifications@github.com> wrote:
>Maybe I'm not seeing something, but aren't we writing only to
>`.well-known/acme-challenge`? Seem like all other files and directories
>in `.well-known` are left untouched.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/letsencrypt/letsencrypt/issues/1003#issuecomment-184527930

Agreed. This is just about the acme-challenge.
-- 
Harlan Lieberman-Berg
~hlieberman

alanhogan:
It was confusing when I used Dreamhost's one-click Let’s Encrypt integration and then was left trying to figure out who left me a greeting card named `.well-known`. The entire `.well-known` folder should be deleted, right? Assuming it only contains files for "let's encrypt"

hlieberman:
On March 25, 2016 5:54:38 PM EDT, Alan Hogan <notifications@github.com> wrote:
>It was confusing when I used Dreamhost's one-click Let’s Encrypt
>integration and then was left trying to figure out who left me a
>greeting card named `.well-known`. The entire `.well-known` folder
>should be deleted, right? Assuming it only contains files for "let's
>encrypt"
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/letsencrypt/letsencrypt/issues/1003#issuecomment-201532994


[...]

PR #2480, opened 2016-02-16
Remove acme-challenge after cleaning up all challenges

filipochnik:
Solves #1003

### DIFF
--- a/letsencrypt/plugins/webroot.py
+++ b/letsencrypt/plugins/webroot.py
@@ -97,7 +97,7 @@ to serve all files under specified web root ({0})."""
         assert self.full_roots, "Webroot plugin appears to be missing webroot map"
         return [self._perform_single(achall) for achall in achalls]
 
-    def _path_for_achall(self, achall):
+    def _get_root_path(self, achall):
         try:
             path = self.full_roots[achall.domain]
         except KeyError:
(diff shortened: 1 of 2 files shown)


pde:
That was a [strange travis failure](https://gist.github.com/pde/329abca79929b34ce9bb) ("Unable to find authorization") on py26-oldest integration, retrying it.

letsencrypt/plugins/webroot.py
```diff
@@ -106,27 +109,41 @@ def _path_for_achall(self, achall):
...
+            validation_path = self._get_validation_path(root_path, achall)
+            logger.debug("Removing %s", validation_path)
+            os.remove(validation_path)
+            self.performed[root_path].remove(achall)
+
+        for root_path, achalls in six.iteritems(self.performed):
+            if not achalls:
+                logger.debug("All challenges cleaned up, removing %s",
+                             root_path)
+                os.rmdir(root_path)
```
pde: This may raise an exception and crash if that directory is not empty (for instance, if the user had run the webroot plugin with a previous version of the client).

pde: Perhaps:
```
try:
    os.rmdir(root_path)
    logger.debug("All challenges cleaned up, removing %s", root_path)
except OSerror:
    logger.debug("Challenges cleaned up but %s not empty, root_path)
```

filipochnik:
@pde I used your snippet, just added checking for the error code.

BTW same error message as https://github.com/letsencrypt/boulder/issues/652, not sure if that's related.

pde:
Great, thanks!

Jalle19:
It seems the directory doesn't get removed if the certificate request fails for some reason. I noticed this because we use Capistrano for deployments, and in an old release that was about to be pruned there was a `.well-known/` folder left in the web root, owned by root, which causes all new deployments to fail since Capistrano can't delete the oldest release anymore.

filipochnik:
@Jalle19 If I understood your problem correctly, leaving `.well-known` untouched is the desired behavior as per discussion here https://github.com/letsencrypt/letsencrypt/issues/1003 

Can you post your comment there?

bmw:
I'm working on a major webroot rewrite. In this I'm fixing permission problems, allowing for interactively choosing webroot paths, and I also plan to work on this issue. What I'm planning on doing is deleting `.well-known` if the client created it and leaving it untouched otherwise.

Jalle19:
@bmw another option that would work for me would be to create `.well-known` using the same permissions as the web root itself. This way the user that owns the web root wouldn't have any problems deleting it even though Letsencrypt leaves files in it.

alasdairdc:
@bmw Hows the rewrite going on?

bmw:
I did the rewrite in #2758, but I unfortunately ran out of time to add this feature.

PR #2482, opened 2016-02-16
wrap csr in ComparableX509

pfeyz:
This causes the example client to fail for the correct reason instead of

    AttributeError: 'X509Req' object has no attribute 'wrapped'


kuba:
LGTM, thanks

PR #2484, opened 2016-02-16
Explain IP addresses are specifically forbidden

schoen:
People still sometimes ask on the forums and issue tracker whether we *really* meant to forbid them to ask for bare IP addresses in certs. This PR adds a new, distinct error message in this case to make clear that yes, we really did specifically intend that.

### DIFF
--- a/letsencrypt/le_util.py
+++ b/letsencrypt/le_util.py
@@ -319,7 +319,7 @@ def enforce_domain_sanity(domain):
     domain = domain[:-1] if domain.endswith('.') else domain
 
     # Explain separately that IP addresses aren't allowed (apart from not
-    # being FQDNs) because hope springs eternal on this point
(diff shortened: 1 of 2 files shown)


schoen:
The failure for 16761bc seemed to be a spurious integration test failure (from the boulder side).

bmw:
Yep. Travis failure is a known, spurious `boulder` problem.

Making a trivial change is a fine way to rerun tests, but alternatively you can:

1. Restart the specific test that failed through Travis. This has the benefit of not running all tests again because one failed. If for some reason you can't/don't want to login to Travis to do this, ping me and I'll happily do it for you.
2. Close and reopen the PR. I usually wait roughly 10 seconds with the PR closed though. I've done it quickly in the past and the build didn't start again!

pde:
LGTM.

PR #2485, opened 2016-02-16
Fix minor bootstrap problems

bmw:
I was trying to run `letsencrypt-auto --os-packages-only --debug` on an Arch system and had the following problems.

1) When running `letsencrypt-auto` as root, `"$SUDO"` becomes an empty string leading to:
```
++ '' pacman -T python2 python-virtualenv gcc dialog augeas openssl libffi ca-certificates pkg-config
./letsencrypt-auto: line 310: : command not found
```

To fix this, I removed quotes around `$SUDO` from `arch_common.sh` and the other bootstrap scripts that had them.

2) `pacman -T example` exits with code 127 if the `example` package is not installed. Because `set -e` is at the top of `letsencrypt-auto`, this means that the script will exit if it needs to install any OS packages.

I fixed this by appending `|| true` to the end of the line containing `pacman -T`.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -328,19 +328,19 @@ BootstrapGentooCommon() {
 
   case "$PACKAGE_MANAGER" in
     (paludis)
-      "$SUDO" cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
+      $SUDO cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
       ;;
     (pkgcore)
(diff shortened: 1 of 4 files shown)


pde:
LGTM but needs test farm tests.

erikrose:
LGTM as well.

bmw:
Merging this, while I didn't run test farm tests (none of the affected systems are tested against in the test farm), I manually tested the changes on Arch Linux, FreeBSD, and Gentoo. On Gentoo, I only tested change with `emerge`, but I think this is okay since this is Gentoo's default package manager, the Gentoo bootstrap scripts are behind a `--debug` flag, and Gentoo has an official package.

PR #2486, opened 2016-02-16
Upgrade peep to 3.1.1.

erikrose:
Fix bad LE experience reported at https://github.com/erikrose/peep/issues/119.


bmw:
LGTM

edmorley:
Please can we review/merge this asap? :-)

We're having many issues incorrectly filed against the peep repo, due to the confusing error output that this PR will resolve (since the updated version of peep includes my change from erikrose/peep#118).

PR #2314, opened 2016-01-29
Small Apache 2.2 fixes

bmw:
Fixes:
* Trying to modify the wrong config file in CentOS6.
* Not adding `NameVirtualHost` causing `_default_` and `*` conflicts.

PR #2487, opened 2016-02-16
Fixes #2470

### LINKED ISSUE #2470: NameError: free variable 'addrs' referenced before assignment in enclosing scope
When using `./letsencrypt-auto run` on Ubuntu 14.04 I run into this error:

`An unexpected error occurred:
NameError: free variable 'addrs' referenced before assignment in enclosing scope`.

From letsencrypt.log:
```
2016-02-13 10:13:29,799:DEBUG:letsencrypt.error_handler:Calling registered functions
2016-02-13 10:13:29,802:DEBUG:letsencrypt.cli:Exiting abnormally:
  File "/home/aaa/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/cli.py", line 1987, in main
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/cli.py", line 667, in run
    lineage.chain, lineage.fullchain)
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/client.py", line 355, in deploy_certificate
    fullchain_path=fullchain_path)
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 214, in deploy_cert
    vhost = self.choose_vhost(domain)
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 308, in choose_vhost
    vhost = self.make_vhost_ssl(vhost)
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 733, in make_vhost_ssl
    self._add_name_vhost_if_necessary(ssl_vhost)
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 905, in _add_name_vhost_if_necessary
    test_addr in test_vh.addrs) and
  File "/home/aaa/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/configurator.py", line 905, in <genexpr>
    test_addr in test_vh.addrs) and
NameError: free variable 'addrs' referenced before assignment in enclosing scope
```

I'm using version 45ee1eb889c86f582c53b1743fdfaf9187c4658b.

bmw:


bmw:
Closed and reopened because Travis tests didn't start for some reason...

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -895,9 +895,10 @@ def _add_name_vhost_if_necessary(self, vhost):
         for addr in vhost.addrs:
             # In Apache 2.2, when a NameVirtualHost directive is not
             # set, "*" and "_default_" will conflict when sharing a port
+            addrs = set((addr,))
```
pde: Trying to restate what this does for sanity checking:

In addition to ensuring that `addrs` is always defined, this change will cause `add_name_vhost` to be called once for every other vhost with the same address.  Previously it would only be called if there was a `"*"` or `"_default_"` vhost.

pde:
Do we want to call `break` after line 910?

pde:
Or maybe the call to `is_name_vhost` in the `if` statement makes that unnecessary?

letsencrypt-apache/letsencrypt_apache/configurator.py
bmw: Almost.

> In addition to ensuring that addrs is always defined, this change will cause add_name_vhost to be called once for every other vhost with the same address **if the address does not already have a name vhost directive.**

> Previously it would only be called if there was a "*" or "_default_" vhost.
Yes, this was also the case before the last commit on #2314, including in `master` before the PR was merged.

bmw:
No, we do not want to `break` after line 910 because we may need to add virtual host directives for more than one address.

pde:
Hmmm, but isn't `addr` a constant through the entire [inner for loop](https://github.com/letsencrypt/letsencrypt/pull/2487/files#diff-35a5744d3537602a402f3dd9aaedd34bL902)?  So there's no reason to continue looping once you've enabled name based vhosting for a given `addr`?  (I'm not saying that this is a bug, I think the call to `is_name_vhost` probably protects us, just that things might be clearer with a `break`).

bmw:
Ah you're right, both in that `addr` is a constant in the inner loop and that `is_name_vhost` protects us. Regardless, I added the `break`.

pde:
LGTM.

PR #2489, opened 2016-02-17
Fix apache 2.2 redirect problems (Issue 2455) [needs minor revision]

### LINKED ISSUE #2455: redirect fails on apache
in apache if there is a default-ssl conf around then redirect fail
on apache 2.2 and 2.4 it gives this error:
```
2016-02-11 20:31:40,023:WARNING:letsencrypt_apache.configurator:Failed redirect for noah2.isnot.org
2016-02-11 20:31:40,023:WARNING:letsencrypt.client:Unable to set enhancement redirect for noah2.isnot.org
Unable to find corresponding HTTP vhost; Unable to create one as intended addresses conflict; Current configuration does not support automated redirection

IMPORTANT NOTES:
 - If you lose your account credentials, you can recover through
   e-mails sent to noah@eff.org.
 - We were unable to set up enhancement redirect for your server,
   however, we successfully installed your certificate.
   /etc/letsencrypt/live/noah2.isnot.org/fullchain.pem. Your cert will
   expire on 2016-05-12. To obtain a new version of the certificate in
   the future, simply run Let's Encrypt again.
 - Your account credentials have been saved in your Let's Encrypt
   configuration directory at /etc/letsencrypt. You should make a
   secure backup of this folder now. This configuration directory will
   also contain certificates and private keys obtained by Let's
   Encrypt so making regular backups of this folder is ideal.
```
according to @bmw it should work on 2.4 but not 2.2 because of how _default_:443 is handled, but we don't seem to have code to deal with this

relevant code is configurator.py lin 1050

SwartzCr:
an attempt to fix #2455 
pull in either a vhost with the same names or a generic one (one with no names) on the same addr/port and add a conditional redirect to it using all of the server names in the newly created ssl vhost

more info on the apache syntax here: https://httpd.apache.org/docs/2.2/mod/mod_rewrite.html#rewritecond

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/obj.py
+++ b/letsencrypt-apache/letsencrypt_apache/obj.py
@@ -194,6 +194,8 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
 
         Used in redirection - indicates whether or not the two virtual hosts
         serve on the exact same IP combinations, but different ports.
+        The generic flag indicates that that we're trying to match to a
+        default or generic vhost
 
         .. todo:: Handle _default_
 
@@ -207,8 +209,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
             if self.name is not None or self.aliases:
                 return True
         # If we're looking for a generic vhost, don't return one with a ServerName
-        else:
-            if self.name:
+        elif self.name:
                 return False
 
         # Both sets of names are empty.
--- a/letsencrypt-apache/letsencrypt_apache/obj.py
+++ b/letsencrypt-apache/letsencrypt_apache/obj.py
@@ -189,7 +189,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
                     return True
         return False
 
-    def same_server(self, vhost):
+    def same_server(self, vhost, generic=False):
         """Determines if the vhost is the same 'server'.
 
         Used in redirection - indicates whether or not the two virtual hosts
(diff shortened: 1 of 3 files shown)


letsencrypt-apache/letsencrypt_apache/obj.py
```diff
@@ -189,7 +189,7 @@ def conflicts(self, addrs):
                     return True
         return False
 
-    def same_server(self, vhost):
+    def same_server(self, vhost, generic=False):
```
bmw: We should update the docstring to say what generic does.

While slightly outside the scope of this PR, it'd be nice to update the documentation for all parameters/return values.

letsencrypt-apache/letsencrypt_apache/obj.py
```diff
@@ -199,12 +199,17 @@ def same_server(self, vhost):
 
         """
 
-        if vhost.get_names() != self.get_names():
-            return False
+        if not generic:
```
bmw: I think we can make this logic a little simpler. I think this line should be `if generic and self.name:`

letsencrypt-apache/letsencrypt_apache/obj.py
```diff
@@ -199,12 +199,17 @@ def same_server(self, vhost):
...
+                return False
 
-        # If equal and set is not empty... assume same server
-        if self.name is not None or self.aliases:
-            return True
+            # If equal and set is not empty... assume same server
+            if self.name is not None or self.aliases:
+                return True
+        # If we're looking for a generic vhost, don't return one with a ServerName
+        else:
```
bmw: nit:
```
else:
    if:
```

Can just be `elif`

bmw:
Other than minor documentation/stylistic nits, this lgtm.

bmw:
LGTM. Merging.

Issue #2490, opened 2016-02-17
automatic upgrading in letsencrypt-auto downgrades to incompatible version


tsuriga:
Tested in latest master, https://github.com/letsencrypt/letsencrypt/commit/1a77c7385768e73851b2e4e685073fa39d4a2298

**Steps:**

1. git clone https://github.com/letsencrypt/letsencrypt.git
2. cd letsencrypt
3. ./letsencrypt-auto

`letsencrypt: error: unrecognized arguments: --le-auto-phase2`

Cause: **letsencrypt-auto-source/letsencrypt-auto**
```bash
REMOTE_VERSION=`"$LE_PYTHON" "$TEMP_DIR/fetch.py" --latest-version`
if [ "$LE_AUTO_VERSION" != "$REMOTE_VERSION" ]; then
    echo "Upgrading letsencrypt-auto $LE_AUTO_VERSION to $REMOTE_VERSION..."
    ...
fi
"$0" --le-auto-phase2 "$@"
```
will fetch the version from https://pypi.python.org/pypi/letsencrypt/json. Right now that version is "0.4.0". Since the check above only checks if the versions are different, the auto script will "upgrade" version "0.5.0.dev0" to "0.4.0", which does not know about the new argument, and attempt to run with that.

The changes were introduced in https://github.com/letsencrypt/letsencrypt/commit/e8f6d4ff39e0d7cd4331a99a8941fc3771c5f07c

bmw:
Fixed in #2492

PR #2451, opened 2016-02-11
Let --no-self-upgrade bootstrap OS packages. Fix #2432.

erikrose:
--no-self-upgrade metamorphosed from a private flag to a public one, so add a new private flag, --le-auto-phase2, to take its original role of marking the division between phases. This flag must come first and, consequently, can be stripped off the arg list before calling through to letsencrypt, which means the client doesn't need to know about it.

The downside is that anyone still (deprecatedly) running le-auto out of the root of a (recently updated) master checkout will get a "Hey, the current release version le-auto I just self-upgraded to doesn't understand the --le-auto-phase2 flag" error from when we merge this until the next release is made, but that's better than a documented option not working right.

Also, remove a needless folder creation from the Dockerfile.

PR #2492, opened 2016-02-17
Revert "Let --no-self-upgrade bootstrap OS packages. Fix #2432."

### LINKED ISSUE #2432: letsencrypt-auto --no-self-upgrade does not bootstrap
From a test farm test:
```
[Process-2 : client 0 ami-26d5af4c ubuntu15.10]
server ec2.Instance(id='i-2085f6a0') at 54.173.22.73
ubuntu@54.173.22.73
[ubuntu@54.173.22.73] put: letest-1455160422/le.tar.gz -> /home/ubuntu/le.tar.gz
[ubuntu@54.173.22.73] run: tar xzf le.tar.gz
[ubuntu@54.173.22.73] out: To run a command as administrator (user "root"), use "sudo <command>".
[ubuntu@54.173.22.73] out: See "man sudo_root" for details.
[ubuntu@54.173.22.73] out: 
[ubuntu@54.173.22.73] out: 

[ubuntu@54.173.22.73] put: scripts/test_letsencrypt_auto_certonly_standalone.sh -> /home/ubuntu/test_letsencrypt_auto_certonly_standalone.sh
[ubuntu@54.173.22.73] run: ./test_letsencrypt_auto_certonly_standalone.sh 
[ubuntu@54.173.22.73] out: To run a command as administrator (user "root"), use "sudo <command>".
[ubuntu@54.173.22.73] out: See "man sudo_root" for details.
[ubuntu@54.173.22.73] out: 
[ubuntu@54.173.22.73] out: + cd letsencrypt
[ubuntu@54.173.22.73] out: + ./letsencrypt-auto certonly --no-self-upgrade -v --standalone --debug --text --agree-dev-preview --agree-tos --renew-by-default --redirect --register-unsafely-without-email --domain ec2-54-173-22-73.compute-1.amazonaws.com --server http://172.31.54.176:4000/directory
[ubuntu@54.173.22.73] out: Creating virtual environment...
[ubuntu@54.173.22.73] out: ./letsencrypt-auto: 428: ./letsencrypt-auto: virtualenv: not found
[ubuntu@54.173.22.73] out: 

ami-26d5af4c - ubuntu15.10 FAIL


letsencrypt.log
--------------------------------------------------------------------------------

[ubuntu@54.173.22.73] sudo: if [ -f /var/log/letsencrypt/letsencrypt.log ]; then     cat /var/log/letsencrypt/letsencrypt.log; else echo "[novarlog]"; fi
[ubuntu@54.173.22.73] out: [novarlog]
[ubuntu@54.173.22.73] out: 

[ubuntu@54.173.22.73] sudo: if [ -f ./letsencrypt.log ]; then     cat ./letsencrypt.log; else echo "[nolocallog]"; fi
[ubuntu@54.173.22.73] out: [nolocallog]
[ubuntu@54.173.22.73] out: 
```

bmw:
This reverts commit 6eb2d60166f142d489f70a2e6076d2fe7c3b3769.

### DIFF
--- a/letsencrypt-auto-source/Dockerfile
+++ b/letsencrypt-auto-source/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && \
     apt-get clean
 RUN pip install nose
 
-RUN mkdir -p /home/lea/letsencrypt
+RUN mkdir -p /home/lea/letsencrypt/letsencrypt
 
 # Install fake testing CA:
 COPY ./tests/certs/ca/my-root-ca.crt.pem /usr/local/share/ca-certificates/
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -422,10 +422,9 @@ TempDir() {
 
 
 
-if [ "$1" = "--le-auto-phase2" ]; then
+if [ "$NO_SELF_UPGRADE" = 1 ]; then
   # Phase 2: Create venv, install LE, and run.
 
-  shift 1  # the --le-auto-phase2 arg
   if [ -f "$VENV_BIN/letsencrypt" ]; then
     INSTALLED_VERSION=$("$VENV_BIN/letsencrypt" --version 2>&1 | cut -d " " -f 2)
   else
@@ -1666,11 +1665,10 @@ else
     exit 0
   fi
 
-  if [ "$NO_SELF_UPGRADE" != 1 ]; then
-    echo "Checking for new version..."
-    TEMP_DIR=$(TempDir)
-    # ---------------------------------------------------------------------------
-    cat << "UNLIKELY_EOF" > "$TEMP_DIR/fetch.py"
+  echo "Checking for new version..."
+  TEMP_DIR=$(TempDir)
+  # ---------------------------------------------------------------------------
+  cat << "UNLIKELY_EOF" > "$TEMP_DIR/fetch.py"
 """Do downloading and JSON parsing without additional dependencies. ::
 
     # Print latest released version of LE to stdout:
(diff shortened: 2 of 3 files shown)


bmw:
fixes #2490

bmw:
So you guys can do what you want with this, but I think this should be merged as I don't understand why #2451 was.

First of all, when did we stop requiring two reviews for `letsencrypt-auto` changes? They affect our users immediately so we need to be extra diligent so breakages don't occur.

Second, it appears that @erikrose was aware of this problem in #2490.
> The downside is that anyone still (deprecatedly) running le-auto out of the root of a (recently updated) master checkout will get a "Hey, the current release version le-auto I just self-upgraded to doesn't understand the --le-auto-phase2 flag" error from when we merge this until the next release is made, but that's better than a documented option not working right.

I hardly think this is deprecated. https://letsencrypt.org/howitworks/, https://letsencrypt.readthedocs.org/en/latest/intro.html, and https://github.com/letsencrypt/letsencrypt#installation all tell people to clone our repo and run `letsencrypt-auto`. The only place I found that didn't say this was https://letsencrypt.readthedocs.org/en/latest/using.html which provides no instructions on how to obtain `letsencrypt-auto`!

With this in mind, we have to choose between the following two scenarios until 0.4.1/0.5.0.

1. Have anyone who obtains a new copy of `letsencrypt-auto` following our official instructions experience #2490. This will only occur on their first run of the client, but nonetheless, their first experience with `letsencrypt` will be a fatal error with a message about a flag they didn't specify.
2. Let people who checkout `master`, use `--no-self-upgrade`. If they ever forget, `--no-self-upgrade` will be broken and they'll experience the problem in option 1.

In my opinion, the vast majority of people will not include `--no-self-upgrade` on every run of the client so this "fix" that causes confusing behavior/breakage for everyone is not worth it. If you agree, please merge this PR.

cc @schoen, @erikrose, @pde

erikrose:
* I never intended --no-self-upgrade to be public anyhow.
* It's not vital, but it's nice to be able to reproduce an installation of a certain version of LE. Aside from OS-specific packages, there aren't any other easy ways of doing that. We could probably have our cake and eat it too by dubbing --no-self-upgrade the "internal" flag and introducing a new one to act as the public one (and then changing the docs).
* I wasn't aware of the docs elsewhere; I just did a find and replace on the ones in the source tree. We should update those, no? My understanding was that we were trying to get git out of the process for non-devs, if for no other reason than that it makes us have to tiptoe around updating le-auto on master.

bmw:
> I never intended --no-self-upgrade to be public anyhow.

That's fair. If we want, we could add it as a hidden flag to `letsencrypt` that has no effect. It wouldn't be shown in the help, but it won't yell at you if `letsencrypt-auto` passes it to `letsencrypt`.

> It's not vital, but it's nice to be able to reproduce an installation of a certain version of LE. Aside from OS-specific packages, there aren't any other easy ways of doing that. We could probably have our cake and eat it too by dubbing --no-self-upgrade the "internal" flag and introducing a new one to act as the public one (and then changing the docs). What do we think of that?

I agree it'd be nice to have a flag that has the functionality of `--no-self-upgrade` in #2451. I personally don't have too strong of feeling about how that's implemented, other than it shouldn't cause breakages for other users.

> I wasn't aware of the docs elsewhere; I just did a find and replace on the ones in the source tree. We should the other docs, no? My understanding was that we were trying to get git out of the process for non-devs, if for no other reason than that it makes us have to tiptoe around updating le-auto on master.

The first link I posted above is a separate repo, but the other three are in the source tree. The first two correspond to `README.rst` and the last one with no instructions on how to obtain le-auto is `docs/using.rst`.

We should definitely stop having non-devs using `git`, but it looks like we never pulled the trigger on that. During the call tomorrow I think we should discuss our plans for making changes to le-auto and how we expect users to obtain it.

PR #2494, opened 2016-02-17
Removed a mention of SimpleHTTP from comments

moshevds:
I didn't notice that I missed this before, but noticed it now. Sorry for the noise.


pde:
Thanks!

Issue #2206, opened 2016-01-16
--csr option without certonly


kroeckx:
The --csr option seems to be doing strange things when used without certonly.  It's asking me if I want to expand the certificate to have more hostnames and things like that.  It seems to have generated it's own keypair again, it's own csr, and so on.  That clearly wasn't what I wanted.

Maybe you're not supposed to run it in that mode?

pde:
Yes, I believe that `--csr` is ignored in "run" mode.  We should document that now, and fix it at some point.

pde:
Ok, I've made a PR (#2218) to fix the doc issue, and this ticket can be for the longer term wishlist item of supporting --csr in all cases.

schoen:
I have an additional PR (#2504) to add an explicit error message for this case to reduce users' confusion. (It doesn't fix the issue, if we regard the issue as tracking the desire to be able to use `--csr` with `run` in the long run.)

bmw:
While this issue came first, I believe it has been superseded by #2373 so I'm closing this issue.

PR #2504, opened 2016-02-19
Explicit error message for #2206

schoen:
This does not solve #2206 but does cause a clear error message to be generated when supplying `--csr` with `run` (previously, it would just be ignored, which would confuse users who expected to be able to do `letsencrypt --csr foo.pem`, such as the user who filed #2206). Now, it will be an error that stops the execution of the client and directs users to use `certonly` instead.


schoen:
Closing and re-opening in the hope of triggering a test.

pde:
Probably this should go in `handle_csr`, which does all of our other sanity checking for `--csr` invocations.  You could do `if config.verb == "run"` there.

schoen:
> Probably this should go in `handle_csr`, which does all of our other sanity checking for `--csr` invocations.  You could do `if config.verb == "run"` there.

I've added a commit that moves it there.

pde:
LGTM

PR #2509, opened 2016-02-19
Turn the root-level letsencrypt-auto symlink into a regular file. Close #2501.

### LINKED ISSUE #2501: Switch le-auto symlink to copy instead.
Turn the letsencrypt-auto symlink at the root of the project into a copy of letsencrypt-auto as of the latest release. Add a line to the release script to copy the latest build to the root.

Our instructions used to tell people to `git pull` and then run le-auto from the root. This will keep them from running a bleeding-edge master version and allow us to tiptoe less around changes to le-auto (double reviews, etc.).

erikrose:
It will always be a copy of the latest release version, 0.4 in this case. (Modify the release script to make that so.) This way, people using the old method of running le-auto from a git checkout will not end up using a bleeding-edge version, letting us work on the tip-of-tree version more freely.

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -1,1 +1,1809 @@
-letsencrypt-auto-source/letsencrypt-auto
+#!/bin/sh
+#
+# Download and run the latest release version of the Let's Encrypt client.
+#
+# NOTE: THIS SCRIPT IS AUTO-GENERATED AND SELF-UPDATING
+#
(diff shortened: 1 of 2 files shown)


bmw:
LGTM. We just need one additional review to end the idea of needing additional reviews.

bmw:
ping @pde

pde:
LGTM too.

Issue #1977, opened 2015-12-22
Audit nginx plugin for guaranteed config reversion in case of error


pde:
This is the nginx equivalent of #1394

olabini:
This is being worked on by @tcz001

tcz001:
Just need some confirm, is this issue only related with all `try catch` containing file dump?
If yes, the pull request https://github.com/letsencrypt/letsencrypt/pull/1457 fixed most places we throw any error related with file dump process.

The missing part is [here](https://github.com/twstrike/letsencrypt/commit/cc3ab2142ef531bc1f6b0198745a272b52be324f), most revert functions are missing error handling in nginx_configurator (which is taken in apache_augeas_configurator)

pde:
In the Apache case, we had to think about config file writing as well as symlink generation in `/etc/apache/sites-enabled`, module enablement and a few things.  I'm not sure if nginx has all of those cases, but it may have the symlinks on some systems?

We had to think about `try except` wrappers for everything, including cases like control-c triggering an exit.

pde:
Actually another thing that was important in the Apache case was making sure that the checkpoint we rolled back to was the correct one.  That was tricky because with the `run` subcommand (the default) we would make temporary config changes for authentication, and then permanent config changes for installation, and in some cases we had bugs where we would roll back to the wrong one.

tcz001:
Yes you're right, Nginx also uses symlinks in site-enabled as best
practice.

2016年1月26日 下午1:07，"Peter Eckersley" <notifications@github.com>写道：
>
> In the Apache case, we had to think about config file writing as well as
symlink generation in /etc/apache/sites-enabled, module enablement and a
few things. I'm not sure if nginx has all of those cases, but it may have
the symlinks on some systems?
>
> We had to think about try except wrappers for everything, including cases
like control-c triggering an exit.

tcz001:
AFAICS the temporary checkpoint is working as expectation. The installation
step sometimes make two permanent checkpoints while choosing second strict
mode. I can try to write some test to see if that's a problem when catching
error.
2016年1月26日 下午1:17，"蒋颿" <fan.torchz@gmail.com>写道：

tcz001:
This [PR](https://github.com/letsencrypt/letsencrypt/pull/2282) should fix most of problem and show the OSError and IOError message in a Plugin Error,

bmw:
I'm sorry I wasn't around to answer questions about this.

[...]

PR #2510, opened 2016-02-19
Temporary saves only

bmw:
This is related to the problem with `certonly` and the Nginx plugin I mentioned in #1977. This will be a long post for a simple change, but I want to explain context for the reviewer as well link to this post in #1977.

##### Temporary changes

In the context of the Apache/Nginx plugins, these changes are made when passing `temporary=True` to the `save` method. These changes will be reverted next time `letsencrypt` is run or when `cleanup`/`recovery_routine` is called.

##### In Progress changes

These changes are made when calling the `save` method with `temporary=False` (which is the default) and no title. These changes will be reverted next time `letsencrypt` is run or when `recovery_routine` is called.

##### Checkpointed changes

These changes are made when calling the `save` method with `temporary=False` and a title. The changes include not only those made since the last call to `save`, but all in progress changes as well. They are reverted only when rolling back checkpoints.

#### Why shouldn't authenticators save in progress changes?

An authenticator should only be making temporary changes to the user's configuration. It is not expected that an authenticator should be making permanent changes to the user's configuration in any way.

Per `interfaces.py`, the only reversion function that an authenticator has is `cleanup`. This should revert all changes the plugin has made. Currently, `cleanup` only reverts temporary changes for both the Apache and Nginx plugins, however, the line of code I changed here previously saved in progress changes.

There are a number of fixes to this problem:

1. Have `cleanup` also revert in progress changes. While this works, I think it makes the most sense for all authenticator functions to only make temporary changes since its changes should only exist while performing the challenge.
2. Delete the save if possible. I believe this is possible for the Apache plugin. I am pretty confident Apache never makes any changes before this save (not 100% confident because Augeas continues to surprise me). Nginx, however, does make changes before this save causing the problem I described [here](https://github.com/letsencrypt/letsencrypt/issues/1977#issuecomment-185493304).
3. Make this save temporary so it is reverted properly. This is the approach taken by this PR.

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/tls_sni_01.py
+++ b/letsencrypt-apache/letsencrypt_apache/tls_sni_01.py
@@ -62,7 +62,7 @@ class ApacheTlsSni01(common.TLSSNI01):
             return []
         # Save any changes to the configuration as a precaution
         # About to make temporary changes to the config
-        self.configurator.save()
+        self.configurator.save("Changes before challenge setup", True)
 
         # Prepare the server for HTTPS
         self.configurator.prepare_server_https(


bmw:
Adding this to 0.4.1, because it's a small change and fixes a potential (but unlikely) reversion problem.

SwartzCr:
I agree with brad's logic here and am +1 to this change

PR #2515, opened 2016-02-20
Class advice impossible

thanatos:
The uses of these two `zope` items causes exceptions in Python 3. The exception suggests this use instead, and exception is definitely being raised because of the former syntax being used in Python 3:

Here's an example,

```
----> 1 import letsencrypt.configuration

…/letsencrypt/configuration.py in <module>()
     12 
     13 
---> 14 class NamespaceConfig(object):
     15     """Configuration wrapper around :class:`argparse.Namespace`.
     16 

…/zope/interface/declarations.py in implements(*interfaces)
    410     # the coverage for this block there. :(
    411     if PYTHON3: #pragma NO COVER
--> 412         raise TypeError(_ADVICE_ERROR % 'implementer')
    413     _implements("implements", interfaces, classImplements)
    414 

TypeError: Class advice impossible in Python3.  Use the @implementer class decorator instead.
```

Attempting to follow the advice given by the exception.

### DIFF
--- a/examples/plugins/letsencrypt_example_plugins.py
+++ b/examples/plugins/letsencrypt_example_plugins.py
@@ -9,9 +9,9 @@ from letsencrypt import interfaces
 from letsencrypt.plugins import common
 
 
+@zope.interface.implementer(interfaces.IAuthenticator)
 class Authenticator(common.Plugin):
     """Example Authenticator."""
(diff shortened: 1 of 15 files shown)


pde:
kicking travis

pde:
This syntax is nicer, too.  Thanks!

PR #2518, opened 2016-02-21
PyOpenSSL 0.13 can not be compiled with OpenSSL 1.0.2

TheBoegl:
This sets the PyOpenSSL version for py26-oldest and py27-oldest to 0.14. Increasing it further or not capping it at all result in a coverage failure within the acme package regarding the standalone_test and crypto_util.

Running `tox` with OpenSSL >=1.0.2a resulted in the following build error which is known in pyca/pyopenssl#276:
```
OpenSSL/crypto/crl.c:6:23: error: static declaration of ‘X509_REVOKED_dup’ follows non-static declaration
   static X509_REVOKED * X509_REVOKED_dup(X509_REVOKED *orig) {
                         ^
  In file included from /usr/include/openssl/ssl.h:156:0,
                   from OpenSSL/crypto/x509.h:17,
                   from OpenSSL/crypto/crypto.h:30,
                   from OpenSSL/crypto/crl.c:3:
  /usr/include/openssl/x509.h:751:15: note: previous declaration of ‘X509_REVOKED_dup’ was here
   X509_REVOKED *X509_REVOKED_dup(X509_REVOKED *rev);
                 ^
  error: command 'gcc' failed with exit status 1
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps =
     py{26,27}-oldest: cryptography==0.8
     py{26,27}-oldest: configargparse==0.10.0
     py{26,27}-oldest: psutil==2.1.0
-    py{26,27}-oldest: PyOpenSSL==0.13
+    py{26,27}-oldest: PyOpenSSL==0.14
     py{26,27}-oldest: python2-pythondialog==3.2.2rc1
 
 [testenv:py33]


bmw:
I can't run the `*-oldest` tests on my system either, however, we specifically want to test against `PyOpenSSL 0.13` as that is the latest version available in some of the distributions where we are packaged. The tests run correctly in Travis though.

Perhaps we could just add a comment about this in the developer documentation.

TheBoegl:
A comment should be enough, though another option would be to specify a range like `PyOpenSSL>=0.13,<0.15` which would compile on newer OpenSSL versions as well as on 'older' distributions.

bmw:
I believe that would result in 0.14 always being installed when installing our dependencies from PyPI since `pip` prefers the latest available version. This would affect both Travis and everyone using our developer instructions.

pde:
This PR defeats the purpose of the `-oldest` test, which is to check that the client is running correctly on rather old versions of various OSes.  @TheBoegl is there a way you can meet your needs by simply avoiding running the `-oldest` tests?

TheBoegl:
@pde Updating the developers guide regarding the section 'how to submit a pull request' point 5 should be sufficient.

bmw:
Opened #2556 for tracking the issue.

PR #2529, opened 2016-02-23
Set CSR version in make_csr

### LINKED ISSUE #2528: Version must be set on CSR before marshaling
If you happen to have a very old version of OpenSSL (looking at you Travis) and you don't set the CSR Version an empty integer (0 length byte slice) will be inserted into the version field which causes Boulder to choke as this is an invalid DER encoding. On newer versions if the version isn't set OpenSSL will sneakily just insert a `0` value and continue, meaning we hadn't stumbled across this before.

The version should be set using `req.set_version(2)` (2 = X509v3... because) in [`make_csr`](https://github.com/letsencrypt/letsencrypt/blob/fd1ebb2d9f6a0030307028aad92ceee586fc06c5/letsencrypt/crypto_util.py#L95). Fixing this will allow us to unblock moving Boulder to Golang 1.6 (which we should build a Boulder instance from and run a test-farm thingies against just to check!)

rolandshoemaker:
Fixes #2528.

### DIFF
--- a/letsencrypt/crypto_util.py
+++ b/letsencrypt/crypto_util.py
@@ -118,6 +118,7 @@ def make_csr(key_str, domains):
             value=", ".join("DNS:%s" % d for d in domains)
         ),
     ])
+    req.set_version(2)
     req.set_pubkey(pkey)
     req.sign(pkey, "sha256")
     return tuple(OpenSSL.crypto.dump_certificate_request(method, req)


bmw:
LGTM. Thanks for tracking this down @rolandshoemaker. Sorry I wasn't more useful in helping with this yesterday.

PR #2533, opened 2016-02-23
Try to finally prevent dangling AWS volumes

pde:


pde:
Semi-tested, somewhat safe to self-merge.

Issue #2508, opened 2016-02-19
Renewer should report failures via exit status


marcan:
The `renew` subcommand should cause the process to exit with a nonzero exit status if renewal failed for any certificate. This makes it easier for an automated renewal script to determine if something went wrong or not. Right now, it seems to exit with status 0 anyway (at least in dry run mode).

Relatedly, I would also suggest adding a "quiet" option that skips the "Processing" lines and does not print the final summary if the status of all certs is "skipped". This makes it easier to just run as a cronjob, since it will avoid producing any output if nothing was done, and therefore won't send an email.

pde:
Eeep, I actually think this is a bug because we've given examples of scripts that assume this return code is available.

bmw:
Created #2512 so the `--quiet` issue isn't lost with the issue about the `letsencrypt renew` exit code.

PR #2534, opened 2016-02-23
Return an error code if any renewals fail

pde:
Closes: #2508

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -1030,6 +1030,12 @@ def renew(config, unused_plugins):
     _renew_describe_results(config, renew_successes, renew_failures,
                             renew_skipped, parse_failures)
 
+    if renew_failures or parse_failures:
+        raise errors.Error("{0} renew failure(s), {1} parse failure(s)".format(
+            len(renew_failures), len(parse_failures)))
+    else:
+        logger.debug("no renewal failures")
+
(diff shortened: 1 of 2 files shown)


letsencrypt/cli.py
```diff
@@ -1030,6 +1030,12 @@ def renew(config, unused_plugins):
     _renew_describe_results(config, renew_successes, renew_failures,
                             renew_skipped, parse_failures)
 
+    if renew_failures or parse_failures:
+        raise errors.Error("{0} renew failure(s), {1} parse failure(s)".format(
```
schoen: I looked around at the call chain and it seems that this error will be uncaught and ultimately cause the interpreter to exit with status code 1, but it also seems that the API would allow `renew` to return a status which could be used as the exit code -- is there a reason to prefer this over the other case?

Do we have a strategy for eventually returning different exit codes to indicate different kinds of problems?

schoen:
LGTM apart from my one note concerning line 1034.

letsencrypt/cli.py
pde: Conclusion: it's caught by `_handle_exception` and will lead to an exit(1) without a traceback.

I think we should consider changing the convention that verbs return strings, and have exceptions be the only way that they return non-zero error codes.

Issue #2240, opened 2016-01-20
Handle wildcards when trying to match the best server name


ehsan:
Currently if you have an apache vhost file with a `VirtualHost` containing:

```
  ServerName site.com
  ServerAlias *.site.com
```

And pass `-d site.com,www.site.com` to the Let's Encrypt client, you get a "We were unable to find a vhost with a ServerName or Address of www.site.com" error message and even worse get a selection table.

This is because [_find_best_vhost()] (https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-apache/letsencrypt_apache/configurator.py#L341) is completely unaware of aliases.

patf:
[get_names()](https://github.com/letsencrypt/letsencrypt/blob/6fbcebd4ab8cb356de92c80c976c9bf0fefb13b4/letsencrypt-apache/letsencrypt_apache/obj.py#L130) seems to handle aliases. I think the issue here is that the client doesn't match wildcards in any way.

I suppose the right question to ask is whether it makes sense to support this given the fact that Let's Encrypt itself doesn't support wildcard certificates. Users would have to enumerate all domains when they call the client anyway, so expecting the same from your apache configuration is fine IMO. This would also further complicate the matching logic, as wildcard aliases could intersect with regular names and thus require additional weighing logic.

ehsan:
> get_names() seems to handle aliases. I think the issue here is that the client doesn't match wildcards in any way.

You're right!

> I suppose the right question to ask is whether it makes sense to support this given the fact that Let's Encrypt itself doesn't support wildcard certificates. Users would have to enumerate all domains when they call the client anyway, so expecting the same from your apache configuration is fine IMO. This would also further complicate the matching logic, as wildcard aliases could intersect with regular names and thus require additional weighing logic.

Well, this will force users to have bloated config files, like:

```
  ServerAlias *.site.com
  ServerAlias every.site.com
  ServerAlias single.site.com
  ServerAlias prefix.site.com
```

It's also surprising since if you know how ServerAlias works in Apache, it's not that obvious why the client complains about some but not all of your domain names.  It took me ~half an hour to figure out what was wrong and that was only after I ended up looking at the code.  :-)

pde:
I think it's probably correct to match the wildcards when deciding if a vhost is the one we want to work with, but we might want to add those ServerAliases ourselves as we install certs (along the same lines as #1991).

SwartzCr:
[...]

PR #2537, opened 2016-02-24
Apache: handle wildcards when matching server names (Issue #2240)

SwartzCr:
relating to #2240 for @bmw to review

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/configurator.py
+++ b/letsencrypt-apache/letsencrypt_apache/configurator.py
@@ -344,9 +344,16 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
     def included_in_wildcard(self, names, target_name):
         """Helper function to see if alias is covered by wildcard"""
-        wildcards = [domain for domain in names if domain.startswith("*")]
+        target_name = target_name.split(".")[::-1]
+        wildcards = [domain.split(".")[1:] for domain in names if domain.startswith("*")]
         for wildcard in wildcards:
-            if wildcard.split(".")[1] == target_name.split(".")[1]:
+            if len(wildcard) > len(target_name):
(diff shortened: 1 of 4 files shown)


letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -342,6 +342,21 @@ def _choose_vhost_from_list(self, target_name, temp=False):
         self.assoc[target_name] = vhost
         return vhost
 
+    def included_in_wildcard(self, names, target_name):
+        """Helper function to see if alias is covered by wildcard"""
```
bmw: When looking into this, I found that `ServerAlias` directives can also include a protocol/port. For example, `https://*example.com:443` is a valid server alias. While this code won't blow up if it encounters one of these directives, they aren't correctly parsed as a `ServerAlias`.

letsencrypt-apache/letsencrypt_apache/configurator.py
```diff
@@ -351,16 +366,21 @@ def _find_best_vhost(self, target_name):
         :returns: VHost or None
 
         """
-        # Points 4 - Servername SSL
-        # Points 3 - Address name with SSL
-        # Points 2 - Servername no SSL
+        # Points 6 - Servername SSL
```
bmw: I'm not sure it really matters, but Apache gives no preference to exact `ServerName/ServerAlias` vs those with wildcards ([source](https://httpd.apache.org/docs/current/mod/core.html#serveralias)). For this reason, I'm not sure we should either and perhaps they should have the same priority.

bmw:
SwartzCr and I decided there were some additional nice things we could do here, but we don't need to block on them. I'll open additional issues after merging this PR.

lgtm

PR #2548, opened 2016-02-25
Expanding tests for le-auto, adding Ubuntu test suite

TheNavigat:
Addresses issue #2443 

Added Ubuntu Precise (12.04), and Trusty (14.04)


TheNavigat:
For some reason the tests fail on Ubuntu Wily (15.10). I traced the problem and I found out that for some reason the ``fetch.py`` script created by ``letsencrypt-auto`` fails with the following

```
Couldn't download https://localhost:4443/letsencrypt/json. <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>
```

Which is returned by ``HttpsGetter``'s ``get`` function.

Not sure why that happens, though. Will keep digging.

TheNavigat:
Ah, found it.

Ubuntu 15.10 uses Python 2.7.10 as the default 2.7 version, while Ubuntu 14.04 uses 2.7.6. The behaviour of the HTTPS URL opener has changed between those versions, and while it skips certificate verification in the older version, the newer one doesn't.

Looks like we'll have to edit the Python script a little.

TheNavigat:
Aaaaaand what fixes the test on 15.10 leads it to fail on 14.04 because ``build_opener`` didn't accept context as a parameter until Python 2.7.9.

I'll probably force Ubuntu 14.04 to install an updated version of Python, although that is /probably/ a really bad thing to do.

TheNavigat:
For reference, here's the exact error the script prints.

``
Couldn't download https://localhost:4443/letsencrypt/json. <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>
``

TheNavigat:
Aaaand now I'm stuck with the following.

``
Couldn't download https://localhost:4443/v/letsencrypt-auto. HTTP Error 404: Path "v/letsencrypt-auto" not found in self.resources
``

TheNavigat:
Was my own fault. First test now passes, but script fails at the second test with ``No distributions at all found for setuptools==20.2.2``

https://github.com/letsencrypt/letsencrypt/pull/2608 will probably affect the issue, and therefore I'll wait for the PR to be merged.

TheNavigat:
And Travis fails on both ``le_auto`` and ``le_auto_wily`` and I have absolutely no idea why.

Tagging the PR with [need help]

TheNavigat:
I'm restoring the old tag since it looks like Travis is doing something wrong. I'll wait for a while, and if the problem persisted I'll ask for help.

TheNavigat:
I think Travis needs a kick. It acted weird for a few minutes.

TheNavigat:
@bmw Can you please take a look at this?

tox.ini
```diff
@@ -127,3 +116,24 @@ commands =
 whitelist_externals =
     docker
 passenv = DOCKER_*
+
+[testenv:le_auto_precise]
+# At the moment, this tests under Python 2.7 only, as only that version is
+# readily available on the Trusty Docker image.
```
bmw: s/Trusty/Precise

letsencrypt-auto-source/Dockerfile.precise
```diff
@@ -0,0 +1,31 @@
...
+    apt-get -q -y install python-pip sudo openssl && \
+    apt-get clean
+RUN pip install nose
+
+# Let that user sudo:
+RUN sed -i.bkp -e \
+      's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' \
+      /etc/sudoers
+
+RUN mkdir -p /home/lea/letsencrypt/letsencrypt
```
bmw: Unless I'm missing something, we don't need and never use the inner `letsencrypt` directory. Additionally, this should probably be renamed `certbot` like the Trusty Dockerfile was. (I know you wrote this before the rename.)

TheNavigat:
@bmw There, fixed :)

PR #2553, opened 2016-02-25
Cast webroot-path from str to [str] if needed

### LINKED ISSUE #2542: Renewal conf files from pre-0.1.0 releases contain webroot-path as a string, not a list
I tried to use the new `renew` subcommand, but it first failed with #2468, which I fixed by manually replacing the port, and then with the following error message:

```
2016-02-24 19:23:18,395:WARNING:letsencrypt.cli:Attempting to renew cert from /etc/letsencrypt/renewal/[redacted].conf produced an unexpected error: The webroot plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('t does not exist or is not a directory',). Skipping.
```

All certificates produced exactly the same error. I looked through the code and it appears it's failing to find the correct webroot path. The path is correct in the config file LE uses to renew the certificate.

pde:
- for compatibility with pre-public-beta renewal conf files
 - fixes #2542


bmw:
lgtm

PR #2565, opened 2016-02-29
Work around leap day bug in parsedatetime

riking:
Leap years are every 4 years, so this workaround will break in 2096.

Hopefully the bug's fixed upstream by then.


schoen:
Thanks for the fix, which was rolled into #2566.

PR #2568, opened 2016-02-29
Pin parsedatetime version < 2.0

bmw:


bmw:
Issue on parsedatetime reported at bear/parsedatetime#156.

schoen:
Hopefully upstream will fix this and we won't need this change. Maybe they can even fix it before Travis finishes! :-)

bmw:
Looks like Travis won the race, although I don't think we can blame upstream too much about that.

schoen:
Sure enough!

schoen:
Good luck to upstream in any case - maybe we can revert this pin soon.

PR #2577, opened 2016-03-01
Simple release help

bmw:
Adds `eff-pubkey.pem` to the tree and stops tracking the releases folder which is created by `release.sh`.


bmw:
@jsha: [boulder log](https://gist.github.com/e9616c17be8640c7fcc5) (error occurred while running `letsencrypt renew` so the client continued running after the error so you'll have to scroll/grep a bit)

pde:
LGTM.

PR #2579, opened 2016-03-01
Legacy http01 port

### LINKED ISSUE #2468: Expected a numeric value for http01_port
Forum user Randall describes this problem when using `letsencrypt renew`:

> ...The error was `Expected a numeric value for http01_port. ...` 
> Editing the file /etc/letsencrypt/renewal/[my.domain].conf file solved this problem. I changed `http01_port = None`  to  `http01_port = 80`

https://community.letsencrypt.org/t/help-us-test-renewal-with-letsencrypt-renew/10562/5

@bmw said that this would have been fixed (in terms of renewal config files correctly specifying `http01_port`) back in November, but people who have older lineages than that may have this problem when renewing with `standalone`. That could conceivably be a pretty large number of users!

bmw:
fixes #2468

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -807,6 +807,7 @@ def _restore_required_config_elements(config, renewalparams):
     for config_item in INT_CONFIG_ITEMS:
         if config_item in renewalparams and not _set_by_cli(config_item):
             config_value = renewalparams[config_item]
+            # the default value for http01_port was None during private beta
             if config_item == "http01_port" and config_value == "None":
                 logger.info("updating legacy http01_port value")
                 int_value = flag_default("http01_port")
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -806,12 +806,17 @@ def _restore_required_config_elements(config, renewalparams):
     # int-valued items to add if they're present
     for config_item in INT_CONFIG_ITEMS:
         if config_item in renewalparams and not _set_by_cli(config_item):
-            try:
-                value = int(renewalparams[config_item])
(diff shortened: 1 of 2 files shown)


letsencrypt/cli.py
```diff
@@ -806,12 +806,18 @@ def _restore_required_config_elements(config, renewalparams):
...
         if config_item in renewalparams and not _set_by_cli(config_item):
-            try:
-                value = int(renewalparams[config_item])
-                setattr(config.namespace, config_item, value)
-            except ValueError:
-                raise errors.Error(
-                    "Expected a numeric value for {0}".format(config_item))
+            config_value = renewalparams[config_item]
+            # the default value for http01_port was None during private beta
+            if config_item == "http01_port" and config_value == "None":
```
schoen: Are you sure that other code couldn't convert this to `None` instead of `"None"` before we get to this point? (Did you check with a real renewal config that had `http01_port` set to `None`)?

schoen:
LGTM, just wondering about the type question noted above (is it `None` or `"None"` at that point in the code?).

schoen:
And thanks for adding the comment we discussed earlier.

letsencrypt/cli.py
bmw: I did check this with a real renewal config. The value was `"None"` and not `None`. I believe this is similar to what's going on [here](https://github.com/letsencrypt/letsencrypt/pull/2579/files#diff-54b2f468d57ef7436008bd031bc89711L803).

schoen:
OK, that makes sense then.

PR #2569, opened 2016-03-01
Release 0.4.1

bmw:

PR #2581, opened 2016-03-01
Bump source le-auto version

bmw:
fix problem @pde identified in #2569.


pde:
I think this change needs to be in the template file too!

bmw:
The version isn't in the template ([source](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt-auto-source/letsencrypt-auto.template#L22)). `letsencrypt-auto-source/build.py` grabs it from `letsencrypt/__init__.py`.

pde:
OK, looks good.

PR #2582, opened 2016-03-01
Autobuild le-auto with dev version

bmw:
Automatically fixes the problem for future releases that was manually fixed in #2581.


bmw:
@jsha: [log](https://gist.github.com/dfdc689f302707bd8655) of spurious boulder problem

bmw:
merging without review because not user facing

PR #2583, opened 2016-03-01
More release enhancements

bmw:
Automatically removes `SIGFILEBALL` after creating signature in `offline-sigrequest.sh`.


pde:
I tend to think we should leave that around as a record!

bmw:
But should we leave it in the tree? This is the first release tag that has contained that file.

If we want it in the tree, this PR can be closed.

If we don't want it in the tree, I believe something like this should be merged. If you just want the file locally, maybe we should just update `.gitignore.`

pde:
I'd say it should be in .gitignore.

pde:
Great!

Issue #1304, opened 2015-11-03
Continuity Authenticator


bmw:
This is a relic from the old spec, over complicate things, and should be removed.

PR #2585, opened 2016-03-01
Progress on #1304

bmw:
This PR stops continuity challenges/authenticator objects from being created when `letsencrypt` runs. They are a relic from an old version of the ACME spec and need to be removed.

This also removes code for the proof of possession challenge which is outdated, not in use, and likely to change dramatically. Despite the fact that proof of possession will almost certainly be implemented in the future, keeping an obsolete version around does more harm than good by complicating our tree and encouraging future design around the crufty code. This latter problem is the cause of many of the issues with the `possible refactor objective?` label.

With that said, the current code **MAY** be useful when writing proof of possession in the future. In this case, we can access that code using past release tags or the `proof-of-possession` branch I just created.

I'll be submitting subsequent PRs that remove PoP as well as completely removing continuity challenge types from both the letsencrypt and ACME modules.


pde:
LGTM

PR #2587, opened 2016-03-02
More progress on #1304

bmw:
This builds on top of #2585. I expected this to be a larger PR, but it's really not too bad. With this in mind, feel free to either review this PR directly or review and merge #2585 first to make things easier.

This PR removes all remnants of the distinction between continuity and DV challenges from the `letsencrypt` module. Next I plan on doing the same from the ACME module which should finish work on #1304.


kuba:
Please use descriptive bug titles when opening - makes email manual filtering life a lot easier, thanks! :)

pde:
LGTM

Issue #2588, opened 2016-03-02
Peep install error when running letsencrypt-auto


singlow:
Just cloned the repo and tried to run in an Ubuntu 14.04 VPS. Using master f205d15ba254

```
peep: (3, 1, 1)
python: '2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2]'
pip: '1.5.4'
Command line:  ['/tmp/tmp.zrzkeazK61/peep.py', 'install', '-r', '/tmp/tmp.zrzkeazK61/letsencrypt-auto-requirements.txt']
Traceback (most recent call last):
  File "/tmp/tmp.zrzkeazK61/peep.py", line 967, in <module>
    exit(main())
  File "/tmp/tmp.zrzkeazK61/peep.py", line 939, in main
    return commands[argv[1]](argv[2:])
  File "/tmp/tmp.zrzkeazK61/peep.py", line 884, in peep_install
    req.install()
  File "/tmp/tmp.zrzkeazK61/peep.py", line 652, in install
    run_pip(['install'] + other_args + ['--no-deps', '-U', archive_path])
  File "/tmp/tmp.zrzkeazK61/peep.py", line 206, in run_pip
    status_code = pip.main(initial_args)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pip/__init__.py", line 185, in main
    return command.main(cmd_args)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pip/basecommand.py", line 161, in main
    text = '
'.join(complete_log)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 79: ordinal not in range(128)```

**Comment by singlow** [comment]
More info. Looks like it may be related to a recent openssl update in Ubuntu. I see that openssl updated today on my all my systems and this brand new vm has the latest openssl. The message below hints that it is related to openssl. So this probably isn't anything you can do anything about.

```
generating cffi module 'build/temp.linux-x86_64-2.7/_constant_time.c'

generating cffi module 'build/temp.linux-x86_64-2.7/_openssl.c'

building '_openssl' extension

creating build/temp.linux-x86_64-2.7/build

creating build/temp.linux-x86_64-2.7/build/temp.linux-x86_64-2.7

[...]

PR #2590, opened 2016-03-02
Update to cryptography 1.2.3

ccovey:
fixes https://github.com/letsencrypt/letsencrypt/issues/2588


letsencrypt-auto
```diff
@@ -500,7 +500,8 @@ configobj==5.0.6
 # sha256: 26GA8xrb5xi6qdbPirY0hJSwlLK4GAL_8zvVDSfRPnM
 # sha256: 5RinlLjzjoOC9_B3kUGBPOtIE6z9MRVBwNsOGJ69eN4
 # sha256: f1FFn4TWcERCdeYVg59FQsk1R6Euk4oKSQba_l994VM
```
erikrose: Just need the rest of these hashes updated

bmw:
Thanks for getting started on this @ccovey. I've been working on this in #2594. Due to the number of OSs we support with `letsencrypt-auto`, this change has actually turned out to be nontrivial. We'll be pushing a fix and doing a 0.4.2 release for #2588 as soon as possible.

PR #2594, opened 2016-03-02
upgrade cryptography version in le-auto

### LINKED ISSUE #2588: Peep install error when running letsencrypt-auto

bmw:
fixes #2588

Will update PR with results of test farm tests.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -431,7 +431,6 @@ InstallRequirements() {
       echo "Had a problem while downloading and verifying Python packages:"
       echo "$PEEP_OUT"
       rm -rf "$VENV_PATH"
-      rm -rf "$TEMP_DIR"
       exit 1
     fi
 }
@@ -458,6 +457,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
 
     echo "Installing Python packages..."
     TEMP_DIR=$(TempDir)
+    trap "rm -rf $TEMP_DIR" EXIT
     # There is no $ interpolation due to quotes on starting heredoc delimiter.
     # -------------------------------------------------------------------------
     cat << "UNLIKELY_EOF" > "$TEMP_DIR/setuptools-requirements.txt"
@@ -1666,7 +1666,6 @@ UNLIKELY_EOF
     # -------------------------------------------------------------------------
     InstallRequirements "setuptools-requirements.txt"
     InstallRequirements "letsencrypt-auto-requirements.txt"
-    rm -rf "$TEMP_DIR"
     echo "Installation succeeded."
   fi
   echo "Requesting root privileges to run letsencrypt..."
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -451,12 +451,6 @@ if [ "$1" = "--le-auto-phase2" ]; then
 # this, do `pip install --no-cache-dir -e acme -e . -e letsencrypt-apache`, and
 # then use `hashin` or a more secure method to gather the hashes.
 
-# cryptography requires a more modern version of setuptools
-# sha256: _ANFf7h6utSdwJ-cMTOGNpPn3bbKgrtQpzmnc3nOWpo
-# sha256: JPz8FTZKn-CaIg830tztyEl5Xj3j5LOT7piOZqnL2Fo
-# sha256: gJaELiTE8ddN_xKr6Qwm0S8F0NmlbtXgb8qm-qHkC2o
-setuptools==20.2.2
-
 # sha256: wxZH7baf09RlqEfqMVfTe-0flfGXYLEaR6qRwEtmYxQ
 # sha256: YrCJpVvh2JSc0rx-DfC9254Cj678jDIDjMhIYq791uQ
 argparse==1.4.0
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -457,7 +457,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
 
     echo "Installing Python packages..."
     TEMP_DIR=$(TempDir)
-    trap "rm -rf $TEMP_DIR" EXIT
+    trap "rm -rf '$TEMP_DIR'" EXIT
     # There is no $ interpolation due to quotes on starting heredoc delimiter.
     # -------------------------------------------------------------------------
     cat << "UNLIKELY_EOF" > "$TEMP_DIR/setuptools-requirements.txt"
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -461,7 +461,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
     # There is no $ interpolation due to quotes on starting heredoc delimiter.
     # -------------------------------------------------------------------------
     cat << "UNLIKELY_EOF" > "$TEMP_DIR/setuptools-requirements.txt"
-# cryptography requires a more modern version of setuptools
+# cryptography requires a more modern version of setuptools.
 # sha256: _ANFf7h6utSdwJ-cMTOGNpPn3bbKgrtQpzmnc3nOWpo
 # sha256: JPz8FTZKn-CaIg830tztyEl5Xj3j5LOT7piOZqnL2Fo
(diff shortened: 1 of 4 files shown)


bmw:
Adding `setuptools` to requirements just changed the [problem](https://gist.github.com/830ef32edec5ef5eeb46).

bmw:
To provide all the necessary context to this PR for the reviewer, I've been running [this](https://gist.github.com/4a5b6f89ccc210b83d02) test farm test to test this PR. This is simply a modified `test_letsencrypt_auto_certonly_standalone.sh` test, except it uses the local copy of `letsencrypt-auto`.

Doing the simple update of `letsencrypt-auto-requirements.txt` to `cryptography` v1.2.3 causes failures on most Red Hat systems. The problem is `cryptography` v1.2.3 requires `setuptools>=1.0` ([source](https://github.com/pyca/cryptography/blob/1.2.3/setup.py#L40)) which wasn't required in our previously pinned version of `cryptography` ([source](https://github.com/pyca/cryptography/blob/1.1.2/setup.py#L40)). A log of test farm failures can be found [here](https://gist.github.com/f44592b1054ea99c719b).

Next I tried pinning `setuptools` to the top of `letsencrypt-auto-requirements.txt` and got errors like [this](https://gist.github.com/830ef32edec5ef5eeb46) on some systems.

Finally, I split `setuptools` into its own requirements file and all tests finally pass. This change is ugly, but it successfully fixes the problem. I would very much like @erikrose's opinion of this PR and to know whether we should ship this as a quick fix for #2588 or if we should wait and use `pipstrap`.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -446,6 +460,15 @@ if [ "$1" = "--le-auto-phase2" ]; then
     TEMP_DIR=$(TempDir)
     # There is no $ interpolation due to quotes on starting heredoc delimiter.
     # -------------------------------------------------------------------------
+    cat << "UNLIKELY_EOF" > "$TEMP_DIR/setuptools-requirements.txt"
+# cryptography requires a more modern version of setuptools
```
erikrose: An independent clause gets a period. :-)

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -1641,18 +1664,9 @@ if __name__ == '__main__':
 
 UNLIKELY_EOF
     # -------------------------------------------------------------------------
-    set +e
-    PEEP_OUT=`"$VENV_BIN/python" "$TEMP_DIR/peep.py" install -r "$TEMP_DIR/letsencrypt-auto-requirements.txt"`
-    PEEP_STATUS=$?
-    set -e
+    InstallRequirements "setuptools-requirements.txt"
```
erikrose: Yay. I like your factoring.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -421,6 +421,20 @@ TempDir() {
...
+    set +e
+    PEEP_OUT=`"$VENV_BIN/python" "$TEMP_DIR/peep.py" install -r "$TEMP_DIR/$1"`
+    PEEP_STATUS=$?
+    set -e
+    if [ "$PEEP_STATUS" != 0 ]; then
+      # Report error. (Otherwise, be quiet.)
+      echo "Had a problem while downloading and verifying Python packages:"
+      echo "$PEEP_OUT"
+      rm -rf "$VENV_PATH"
+      rm -rf "$TEMP_DIR"
```
erikrose: Is there a way we can make deleting the temp dir the responsibility of the one who created it (i.e. the caller)?

letsencrypt-auto-source/letsencrypt-auto
erikrose: Btw, I realized why you need to break this up into 2 files: peep runs all its invocations of pip in a single process, calling out to `pip.main()` directly. I did that so I wouldn't accidentally invoke some other venv's (or the system's) copy of pip, no matter how screwed up `$PATH` happened to be.

bmw:
@jsha: [Error creating new cert](https://gist.github.com/d52b1af2a5d62b2a148b)

bmw:
All test farm tests passed. I also tested this successfully on FreeBSD.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -444,8 +457,18 @@ if [ "$1" = "--le-auto-phase2" ]; then
 
     echo "Installing Python packages..."
     TEMP_DIR=$(TempDir)
+    trap "rm -rf $TEMP_DIR" EXIT
```
erikrose: Should be fine except if there's a space in $TEMP_DIR. I think we could put single quotes around $TEMP_DIR to fix it.

erikrose:
Add those single quotes, and it looks good to me.

letsencrypt-auto-source/letsencrypt-auto
erikrose: Actually, we shouldn't use the [bash-specific EXIT pseudo signal](http://redsymbol.net/articles/bash-exit-traps/); it won't work on plain Bourne shells, which `sh` still is on some platforms.

bmw: I'll fix the single quotes, but I don't think EXIT is bash specific ([source](http://www.unix.com/man-page/POSIX/1posix/trap/)).

I tested the script on FreeBSD and the directory was deleted.

erikrose: lgtm, then.

bmw:
Merging with [Erik's LGTM](https://github.com/letsencrypt/letsencrypt/pull/2594#discussion_r54943750).

BrettTrotter:
I hate to be debbie downer and maybe I'm the only one but on Ubuntu 14.04 LTS with all packages up to date and letsencrypt updated to this revision, i still get the issue....

erikrose:
@BrettTrotter Which issue do you get? The compilation one or the Unicode one?

BrettTrotter:
Sorry I tried to delete it before anyone read it.. I wasn't sure which ticket to put things in and i figured someone else would hit the error too or I was doing something wrong.

I'm still getting the ascii / unicode.

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 79: ordinal not in range(128)

Also I installed pip 8 and pip --version shows the correct version but the letsencrypt-auto still shows pip 1.5.4

erikrose:
@BrettTrotter It may be that virtualenv is installing the version of pip that came bundled with it. Upgrading virtualenv might fix the problem. Or just wait: I'm working on #2596, which should fix it.

BrettTrotter:
Take your time, I still have 15 days! :)
:+1: 
Thank you for your efforts.

BrettTrotter:
upgrading virtualenv fixed it for me. I didn't even think of that prior to your comment, just assumed letsencrypt would upgrade what it needed.

erikrose:
It will soon. :-) We didn't anticipate quite so much ancient infrastructure floating around out there.

PR #2599, opened 2016-03-03
Require setuptools>=1.0 in all our packages.

erikrose:
When pip-installing any of our packages, pip hits our permissive, any-version "setuptools" dependency first and then ignores all subsequent, more constrained ones, like cryptography's "setuptools>=1.0". See https://github.com/pypa/pip/issues/988. It thus, on a box with setuptools 0.9.8, sticks with that version. Then, at runtime, letsencrypt crashes because pkg_resources can't satisfy cryptography's setuptools>=1.0 requirement.

This change lets us pip-install our packages and have it work. We'll need to make sure our direct requirements (all of them) satisfy the more constrained requirements of our dependencies. Yes, it is disgusting.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -18,7 +18,9 @@ install_requires = [
     'pyrfc3339',
     'pytz',
     'requests',
-    'setuptools',  # pkg_resources
+    # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
+    # will tolerate; see #2599:
+    'setuptools>=1.0',
     'six',
 ]
 
--- a/letsencrypt-apache/setup.py
+++ b/letsencrypt-apache/setup.py
@@ -11,7 +11,9 @@ install_requires = [
     'acme=={0}'.format(version),
     'letsencrypt=={0}'.format(version),
     'python-augeas',
-    'setuptools',  # pkg_resources
+    # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
+    # will tolerate; see #2599:
+    'setuptools>=1.0',
     'zope.component',
     'zope.interface',
 ]
--- a/letsencrypt-nginx/setup.py
+++ b/letsencrypt-nginx/setup.py
@@ -12,7 +12,9 @@ install_requires = [
     'letsencrypt=={0}'.format(version),
     'PyOpenSSL',
     'pyparsing>=1.5.5',  # Python3 support; perhaps unnecessary?
-    'setuptools',  # pkg_resources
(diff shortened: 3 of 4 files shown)


bmw:
I can confirm this fixes the problem, but I think we should add a comment to each `setup.py` explaining why we require `setuptools>=1.0`.

letshelp-letsencrypt/setup.py
```diff
@@ -7,7 +7,7 @@
 version = '0.5.0.dev0'
 
 install_requires = [
-    'setuptools',  # pkg_resources
+    'setuptools>=1.0',  # pkg_resources
```
bmw: We don't need to specify a version here because `letshelp-letsencrypt` doesn't use `cryptography`.

bmw:

rbu:
This new requirement now breaks the RHEL7 EPEL package unfortunately. Believe it or not, we actually still have setuptools 0.9.8 in there, and an old version of cryptography which is fine with that.

The dependency to setuptools >= 1.0 was only introduced in cryptography 1.2. In addition, current master of cryptography even requires setuptools 11, ref https://github.com/pyca/cryptography/commit/30e199e0fb220a17b3089508a59d29a2924f10e4

erikrose:
@rbu It doesn't break it instantly, does it? This sort of thing came up when we were deciding what to do, and we decided that the package maintainers can edit out that requirement when they package. Otherwise, LE fails to be pip-installable on every platform, not just RHEL. It's unfortunate that pip's lack of real dependency resolution causes this sort of fallout. :-(

rbu:
@erikrose, I've noticed the breakage during acme's "setup.py test", see the log at https://kojipkgs.fedoraproject.org//work/tasks/6955/13226955/build.log

If it helps general installability, I'm happy if you leave it as is and we downgrade the setuptools dependency during packaging via this patch: http://pkgs.fedoraproject.org/cgit/rpms/python-acme.git/commit/?h=epel7&id=0f7ca9bedf530a08592925edb66f1f84b50e9950

My main concern would be you starting to actually require the higher version in your own code (and not just in dependencies) and we'd not notice this during our testing. Given that this is rather unlikely, with setuptools issues usually being rather visible, I'm good. Thanks for the feedback.

erikrose:
Sure, thanks for bringing it up. Your patch looks good to me, and it's unlikely we would actually start requiring a recent setuptools directly. In any case, if you run our tests as part of your packaging procedure, you should be protected against regressions.

Issue #2229, opened 2016-01-19
Please, make Let's Encrypt client less intrusive


bviktor:
I think we can agree on that reducing the attack surface should be a principle everyone should follow. That means install the least amount of packages that still lets you get things done. Now if I do a mere

    # letsencrypt-auto --help

It installs the following packages:

    python-devel python-virtualenv gcc libffi-devel openssl-devel redhat-rpm-config cpp dwz glibc-devel glibc-headers kernel-headers keyutils-libs-devel krb5-devel libcom_err-devel libmpc libselinux-devel libsepol-devel libverto-devel mpfr pcre-devel perl-srpm-macros zip zlib-devel

**without even asking**. `yum -y` all the way. If I do this as a sudoer _user_, it still becomes root in the blink of an eye, everything's run with `sudo`.

If you think it's acceptable to install compilers and a lot of libs and headers on each and every web server for... running a Python script, that's totally up to you. But please, at least ask us if we really want that.

Now, this wouldn't bother me this much if there was a way to set up a single, isolated VM which I could use to request certificates for all our servers. Except I, of course, can't do that. The script needs to be run on the server that'll serve those domains.

I personnaly erased all these packages and reverted to our Comodo certs for now and by the look of things, it'll stay that way. I want to make it clear that I'm not _complaing_ - it wouldn't even make any sense since you're giving away certs for free, which is awesome. This is merely a suggestion that you may or may not consider.

I do believe that most admins would greatly appreciate if letsencrypt came in the way of

a) a single binary - IIRC there are tools for doing that in Python but FIXME
b) a self-contained Python script - e.g. does serving HTTP requests, amongst other, mostly simple things really require kernel headers and a C compiler in Python? I doubt it.

iinstead of a script that

a) installs a lot of bloat on web servers
b) sets up a Python virtual environment... somewhere, that may be trivial for you python guys but not for the average nginx admin and honestly, I really don't have the interest or desire to deal with it
c) installs packages via pip

Thanks in advance.

kuba:
FYI #334 and https://github.com/kuba/letsencrypt/tree/pyinstaller.

bviktor:
Hm, so if it's that simple, why can't I download a binary from the letsencrypt project? Could this be on the TODO list?

pde:
[...]

PR #2605, opened 2016-03-04
Do not install mod_ssl [needs revisions]

### LINKED ISSUE #2589: Do not install mod_ssl - it can cause an outage
Hi,

I have been having random outages on a development environment recently, usually during a configuration converge or (most often) after a reboot, where Apache would all of a sudden steal port 443 and mean Nginx couldn't bind to it anymore when it was restarted.

I found the cause was the `mod_ssl` package kept being installed (it's a CentOS system). It seemed really odd that it was being installed as it was completely unwanted. Initially I thought I had messed up somewhere and accidentally installed it, but it happened another two times.

I finally found the problem was Let's Encrypt - letsencrypt-auto seems to install `mod_ssl` onto the server because it has `httpd` package installed. It's just sods law that this completely breaks my configuration. :cry: 

Is there a specific reason this is being installed? Python packages and dependencies I can understand, but `mod_ssl` drops in configurations and example SSL configurations which means it enables port 443 and I can't understand why this would be needed by a certificate tool.

I can raise a PR to remove it if there's no reason for it to be installed - maybe it's a remnant of a previous era :smile:

driskell:
Fixes #2589 

Installing mod_ssl on some servers triggers a change in Apache behaviour - it starts listening on port 443. This is not a desired change for me and I don't believe it would be for other people.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -262,6 +262,14 @@ BootstrapRpmCommon() {
       echo "Could not install additional dependencies. Aborting bootstrap!"
       exit 1
   fi
+
+
+  if $SUDO $tool list installed "httpd" >/dev/null 2>&1; then
+    if ! $SUDO $tool install -y mod_ssl
+    then
+      echo "Apache found, but mod_ssl could not be installed."
+    fi
+  fi
 }
 
 BootstrapSuseCommon() {
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -262,14 +262,6 @@ BootstrapRpmCommon() {
       echo "Could not install additional dependencies. Aborting bootstrap!"
       exit 1
   fi
-
-
-  if $SUDO $tool list installed "httpd" >/dev/null 2>&1; then
-    if ! $SUDO $tool install -y mod_ssl
-    then
-      echo "Apache found, but mod_ssl could not be installed."
-    fi
-  fi
 }
 
 BootstrapSuseCommon() {
--- a/letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
+++ b/letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
@@ -50,12 +50,4 @@ BootstrapRpmCommon() {
       echo "Could not install additional dependencies. Aborting bootstrap!"
       exit 1
   fi
-
-
-  if $SUDO $tool list installed "httpd" >/dev/null 2>&1; then
-    if ! $SUDO $tool install -y mod_ssl
-    then
-      echo "Apache found, but mod_ssl could not be installed."
-    fi
-  fi
 }


bmw:
Travis failures are unrelated to this PR. Merging `master` into this PR or rebasing will fix the problem.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -262,14 +262,6 @@ BootstrapRpmCommon() {
       echo "Could not install additional dependencies. Aborting bootstrap!"
       exit 1
   fi
-
-
-  if $SUDO $tool list installed "httpd" >/dev/null 2>&1; then
```
pde: This change needs to be made in the file `letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh` (running `build.py` in `letsencrypt-auto-source/` will propagate the change).

pde:
My suspicion is that this will break `letsencrypt-auto` for a lot of people who are trying to use it with Apache.  One Correct™ solution is probably to have a native package for `letsencrypt-apache` that's separate from the main `letsencrypt` package.  The Debian packaging folks have taken that approach, though the Fedora team has not thus far.

We could add a flag to prevent `mod_ssl` installation (which is pretty hacky).  Or you could try `pip install letsencrypt` rather than `letsencrypt-auto` as a workaround?

FelixSchwarz:
@pde: I don't think the Fedora/EPEL rpm has anything to do with it. As far as I can see EPEL rpms do not have any dependencies on mod_ssl.

driskell:
@pde I've updated the PR to remove from correct place.

I don't understand how this will break `letsencrypt-auto` for a lot of people. Anyone who is already using it will have had `mod_ssl` installed automatically already if it was not originally. For new users I don't believe it would be unexpected to find out they have to install `mod_ssl` if using Apache.

For those automating platforms with already written scripts that aren't properly installing `mod_ssl` along with `httpd` packages - I guess it would affect those and I'm sure there's a way to prevent that. Possibly by only installing `mod_ssl` if actually using Let's Encrypt to manage Apache.

Further examination of all the other bootstrappers leads me to believe this is purely RPM specific behaviour. For all other platforms, including Debian, I can see no code in their bootstrappers that triggers `mod_ssl` to be installed.

I think `letsencrypt-auto` should only be installing dependencies for Let's Encrypt to run. It should not install dependencies for the web server to run as IMHO that's not its job. Especially in my case, installing those dependencies directly impacts the running configuration on the server due to the introduction of `/etc/httpd/conf.d/ssl.conf`.

@FelixSchwarz is correct - I am not using Let's Encrypt packages from EPEL or anywhere. I installed it via `git clone`. This is specifically an issue with the `letsencrypt-auto` script

driskell:
Is there any interest in continuing this PR?
I'll rebase onto latest if there's a chance it will be accepted.
Thanks!

bmw:
@pde should also chime in on this as he's likely to have an opinion, but what about solution similar to what we use for [enabling backports on Debian](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh#L49) or [enabling EPEL on CentOS](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh#L38)? Unfortunately, to many people's frustration, when we first wrote `letsencrypt-auto`, it never prompted the user before calling `letsencrypt/certbot` (see #2229). We've fixed this in `certbot-auto`, however, we've preserved this behavior in `letsencrypt-auto` because it is what people are used to and they have built integrations around it.

In the current `*-auto` scripts, the environment variable `$ASSUME_YES` is set to 1 if the user provides `-n/--noninteractive/--non-interactive` on the command line or the name of the script is `letsencrypt-auto`. We can use this variable to do something similar for `mod_ssl`. If `$ASSUME_YES` is 1, we install `mod_ssl` without prompting the user. If it isn't, we prompt the user explaining that `mod_ssl` is needed to use the Apache plugin and ask for permission to install it.

I don't have a strong preference whether we ask the question using `read` and then provide `-y` to `yum/dnf` or simply provide more context and allow `yum/dnf` to ask the question like we do for EPEL on CentOS. Additionally, we should make sure the Apache plugin fails somewhat gracefully when `mod_ssl` isn't installed.

jsha:
@driskell, am I correct in understanding that you want to install Certbot on a machine that happens to have Apache, but you don't want to use it with Apaache, you want to use it for some other piece of software? And it's important to you that Apache not listen on port 443?

While I agree that that's a totally valid use case, my feeling is that `certbot-auto` is geared towards a very easy "it just works" process, especially for users trying to deploy web servers. If users on RedHat got an error message on their first run saying they need to separately install mod_ssl, I think it would break that sense of "it just works."

You mention that the other bootstrappers don't have code to install mod_ssl. At least on Debian and Ubuntu, mod_ssl ships with the default apache2 package (but is not enabled by default).

So my vote is to close, though I appreciate the work @driskell has put into this PR!

driskell:
I apologise if I'm repeating what's known here, but the issue is not specifically that mod_ssl is installed - the issue is that mod_ssl is enabled on install.

I understand the "it just works" concerns but I'm sure collectively we could produce a better solution here that works for everyone with just a little thought. My first thought is this particular distribution has a special case where if conf.d/ssl.conf (or whatever the default mod_ssl config file is called) is created as an empty file if it doesn't exist (touch it for example) before installing mod_ssl. This explicitly ensures the configuration does not activate on install (rpm will not overwrite configs).

I can see only one issue - that mod_ssl must be enabled for things to work - I am sure something could be worked out here though.

Thanks!

driskell:
> am I correct in understanding that you want to install Certbot on a machine that happens to have Apache, but you don't want to use it with Apaache, you want to use it for some other piece of software? And it's important to you that Apache not listen on port 443?

This is correct yes. It's more a case of some packages I use require Apache and run in Apache. So I run Apache on say, 8888. Then my main Web stack runs in Nginx. By installing mod_ssl it default enables in Apache and steals port 443 on reboot. This generates a prolonged outage. I've worked around it for now with an empty ssl.conf file in all images.

jsha:
Yep, agreed that we can come up with a better solution, and thanks for sticking with the issue to help find it!

For your specific case, I think it would work to add a check for a port 443 listener during the bootstrap script, and not install mod_ssl if something else already has port 443, since we know that would cause a conflict. What do you think?

bmw:
@driskell, I'm sorry this PR has sat around so long. If you're still interested, if you implement what I described [here](https://github.com/certbot/certbot/pull/2605#issuecomment-250330222), I'll review and merge the PR.

bmw:
I haven't seen a response in a month+ so I'm closing this PR. With that said, if you'd like to continue working on this, please open a new PR after addressing feedback and we'll happily take a look.

PR #2608, opened 2016-03-04
Use pip 8 (installed via pipstrap) instead of peep. Close #2596.

### LINKED ISSUE #2596: Replace peep with pipstrap and pip 8
A [unicode error](https://github.com/letsencrypt/letsencrypt/issues/2588#issuecomment-191472841) is biting a lot of people who have the several-year-old pip 1.5.4. It's likely a bug in that version of pip and, slightly less likely, in an old version of setuptools. Since pip 8 is now out, let's replace peep with it (via embedding pipstrap), gain a more maintained installer, and dodge these bugs all in one fell swoop. Pipstrap installs a modern pip *and* setuptools, so we're covered either way.

erikrose:
This should solve our unicode error (https://github.com/erikrose/peep/issues/125) as well as many other errors caused by pip, setuptools, or wheel being really, really old. It also means I don't have to maintain peep anymore for LE's sake.

Revert the patch that added the InstallRequirements function, since we're back to needing only 1 requirements file.


bmw:
LGTM.

I also ran test farm tests and everything passed :)

petelewis:
Hi,

I'm still getting the following, even after this. I'm on Centos 6.7, trying to use letsencrypt-auto.

```
peep: (3, 1, 1)
python: '2.7.11 (default, Dec  7 2015, 11:22:37) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)]'
pip: '1.4.1'
Command line:  ['/tmp/tmp.4a30r5X3qA/peep.py', 'install', '-r', '/tmp/tmp.4a30r5X3qA/letsencrypt-auto-requirements.txt']
  File "/tmp/tmp.4a30r5X3qA/peep.py", line 967, in <module>
    exit(main())
  File "/tmp/tmp.4a30r5X3qA/peep.py", line 939, in main
    return commands[argv[1]](argv[2:])
  File "/tmp/tmp.4a30r5X3qA/peep.py", line 884, in peep_install
    req.install()
  File "/tmp/tmp.4a30r5X3qA/peep.py", line 652, in install
    run_pip(['install'] + other_args + ['--no-deps', '-U', archive_path])
  File "/tmp/tmp.4a30r5X3qA/peep.py", line 206, in run_pip
    status_code = pip.main(initial_args)
  File "/home/lewisp/.local/share/letsencrypt/lib/python2.7/site-packages/pip/__init__.py", line 148, in main
    return command.main(args[1:], options)
  File "/home/lewisp/.local/share/letsencrypt/lib/python2.7/site-packages/pip/basecommand.py", line 169, in main
    text = '\n'.join(complete_log)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 36: ordinal not in range(128)
```

But it sounds like from the title of this conversation, that peep shouldn't even be being used any more.

I'm new to letsencrypt, so might be doing something wrong.

erikrose:
While we've merged this branch, we haven't yet done a release that includes the new code. Hang in there! Or, if you want to try some bleeding-edge stuff, download the latest letsencrypt-auto from within our letsencrypt-auto-source folder, and give it a try.

petelewis:
Ah, ha! Thanks for the clarification. I was under the impression that since the version of letsencrypt-auto I was using was from the latest git and that included these commits, then it would have this. No problem :-)

erikrose:
It's tricky: there's a copy at the root of the repo; that gets updated only on release, because it's the copy that the le-auto script downloads to self-update. The bleeding-edge version is in letsencrypt-auto-source/.

PR #2612, opened 2016-03-05
Documentation--webroot configuration text--fix format

wyhitcs:
We found a display error in letsencrypt document at website "Read the Docs",  at this part: http://letsencrypt.readthedocs.org/en/latest/using.html?highlight=mail#webroot

``letsencrypt certonly --webroot -w /var/www/example/ -d www.example.com -d example.com -w /var/www/other -d other.example.net -d another.other.example.net``

Exceed the border of the browser window, I checked at Chrome, Firefox and Safari at Mac OSX and Ubuntu, so I change the way to display it to make it visible: 
 
I imitate the way you layout the instruction in the same file (shown below):

    66.133.109.36 - - [05/Jan/2016:20:11:24 -0500] "GET /.well-known/acme-challenge/HGr8U1IeTW4kY_Z6UIyaakzOkyQgPr_7ArlLgtZE8SX HTTP/1.1" 200 87 "-" "Mozilla/5.0 (compatible; Let's Encrypt validation server; +https://www.letsencrypt.org)"

This text has a horizontal scroll, and it's a good way to display a long text (both in rst and html), so we add a scroll for the text which exceed the browser window mentioned above. Now it looks like this:

    letsencrypt certonly --webroot -w /var/www/example/ -d www.example.com -d example.com -w /var/www/other -d other.example.net -d another.other.example.net


bmw:
Travis failures are unrelated to this PR. Merging `master` into this PR or rebasing will fix the problem.

@jsha: I'm not too worried about one off integration test failures, but here's a [log](https://gist.github.com/b9b2b0068fee78639d3c) where `create_db.sh` failed to run to completion.

PR #2623, opened 2016-03-08
Document some known third-party plugins

pde:


pde:
Self-merging doc change for tomorrow's post

PR #2634, opened 2016-03-10
Further plugin documentation cleanups

pde:


pde:
Self-merging doc change

PR #2635, opened 2016-03-10
Expanding tests for le-auto, adding Debian test suite

TheNavigat:
Addresses issue #2443 

Added Debian Wheezy (7) and Jessie (8)


.travis.yml
```diff
@@ -44,6 +44,16 @@ matrix:
       services: docker
       before_install:
       addons:
+    - sudo: required
+      env: TOXENV=le_auto_wheezy
+      services: docker
+      before_install:
+      addons:
+    - sudo: required
```
bmw: After chatting about this with people around the EFF office, we decided we only want these tests for distros where we recommend `certbot-auto` on certbot.eff.org. The reason for this is some devs often find themselves waiting around for Travis to finish so we only want to add jobs for distros where we are actually telling users to run `certbot-auto`.

With this in mind, the Jessie test and the associated Dockerfile should be deleted.

letsencrypt-auto-source/Dockerfile.wheezy
```diff
@@ -0,0 +1,31 @@
...
+    apt-get -q -y install python-pip sudo openssl && \
+    apt-get clean
+RUN pip install nose
+
+# Let that user sudo:
+RUN sed -i.bkp -e \
+      's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' \
+      /etc/sudoers
+
+RUN mkdir -p /home/lea/letsencrypt
```
bmw: nit: Change `letsencrypt` to `certbot`.

tox.ini
```diff
@@ -92,3 +92,23 @@ commands =
 whitelist_externals =
     docker
 passenv = DOCKER_*
+
+[testenv:le_auto_wheezy]
+# At the moment, this tests under Python 2.7 only, as only that version is
+# readily available on the Trusty Docker image.
```
bmw: s/Trusty/Wheezy

bmw:
Looks like there's merge conflicts that need to be resolved. After that, I'll merge.

PR #2649, opened 2016-03-11
Split main.py out of cli.py

pde:
The first part of a solution to #2442.

`renewer.py` and `plugins/selection.py` yet to come.

PR #2650, opened 2016-03-11
Split renew.py out of cli.py

pde:
Builds on #2649.  Further progress toward #2442


bmw:
This looks good! My only suggestion is that perhaps `renew()` should be moved into `main.py`.

This allows us to stop hacking around cyclic import problems and allows `main` and all subcommands to exist in `main.py`.

`renew.py` can continue to contain functions that provide much of the renewal functionality (in actuality, this code is spread throughout the project in places such as `storage.py`). `renew()` could be refactored into additional functions that exist in `renew.py`. This refactoring could be a bit tricky, however.

Alternatively, `renew()` could stay as is and just call relevant functions from `renew.py` which is already imported and used in `main.py`.

pde:
Putting `renew()` in `main.py` won't really simplify cyclic deps since `cli.py` will still import from `main`.  Given that the two imports are in exactly the same place, we don't seem to loose much elegance from the second one :)

I also thought about splitting `renew()` while writing the patch, and concluded there wasn't a nice, quick way to do it.

Perhaps parts of `storage.py` could be refactored into `renew.py` too; I don't have an opinion on that yet.

bmw:
> Putting renew() in main.py won't really simplify cyclic deps since cli.py will still import from main. Given that the two imports are in exactly the same place, we don't seem to loose much elegance from the second one :)

Putting `renew()` in `main.py` breaks `cli.py`'s dependence on `renew.py` and `renew.py`'s dependence on `main.py`. While this simple change does not eliminate our cyclic dependency problem, it reduces it to `cli.py <-> main.py`. Instead of having 3 cyclic dependency problems, we have 1.

> I also thought about splitting renew() while writing the patch, and concluded there wasn't a nice, quick way to do it.

Assuming you don't want to spend much more time on this, what about simply taking `renew()` as is and putting it in `main.py`? I think having all subcommands in one place as well as removing the majority of our cyclic dependencies is worth this simple change.

> Perhaps parts of storage.py could be refactored into renew.py too; I don't have an opinion on that yet.

Maybe, but I don't think that should happen in this PR.

pde:
All of your points are fair. While I agree that `renew()` could conceptually live in either file, I feel pretty dirty having a function in `main.py` that calls all of these things:

* `renew.renewal_conf_files`
* `renew._reconstitute`
* `renew.should_renew`
* `renew._renew_describe_results`

And has semantics that are deeply interwoven with them (especially `_renew_describe_results`).  I'd prefer to leave this as a further refactoring ticket (including addressing `storage.py`).

pde:
@bmw, I added the `main.renew()` as discussed.  That caused me to rename `renew.py` to `renewal.py`, avoiding an ugly namespace conflict within `main`'s module scope.

bmw:
LGTM

PR #2652, opened 2016-03-12
Move plugin selection from cli.py and display/ops.py into plugins/selection.py

### LINKED ISSUE #2442: Split cli.py into smaller pieces

pde:
Split plugin.selection our of cli.py

Builds on: #2650 #2649 

This closes #2442


letsencrypt/plugins/selection.py
```diff
@@ -0,0 +1,275 @@
...
+            if prepared[index].misconfigured:
+                msg = "Reported Error: %s" % prepared[index].prepare()
+            else:
+                msg = prepared[index].init().more_info()
+            z_util(interfaces.IDisplay).notification(
+                msg, height=display_util.HEIGHT)
+        else:
+            return None
+
+logger = logging.getLogger(__name__)
```
bmw: You already did this earlier in the file: https://github.com/letsencrypt/letsencrypt/pull/2652/files#diff-80aef1fff7bcdc6abeb7ebdc5d995bddR14

letsencrypt/plugins/selection.py
```diff
@@ -0,0 +1,275 @@
...
+    "Update the config entries to reflect the plugins we actually selected."
+    cn = config.namespace
+    cn.authenticator = plugins.find_init(auth).name if auth else "none"
+    cn.installer = plugins.find_init(inst).name if inst else "none"
+
+
+def choose_configurator_plugins(config, plugins, verb):
+    """
+    Figure out which configurator we're going to use, modifies
+    config.authenticator and config.istaller strings to reflect that choice if
```
bmw: It looks like this wasn't created in this PR, but...

s/istaller/installer

letsencrypt/main.py
```diff
@@ -32,6 +32,7 @@
 
 from letsencrypt.display import util as display_util, ops as display_ops
 from letsencrypt.plugins import disco as plugins_disco
+from letsencrypt.plugins import selection as ps
```
bmw: `ps` isn't clear and while it shortens some of the longer lines in this file, I don't think it's worth using naming conventions that provide little to no information about the purpose of the module.

pde: Because `choose_configurator_plugins` is such a self-explanatory (but long) function name, I actually tried to import it from the module, but hit some test issues.  I guess I really dislike wrapped lines if they can be avoided :)

pde:
Noting that doing #2716 would slightly improve some of this inelegance, but is vastly more work than it's worth right now.

Issue #2600, opened 2016-03-03
Failed to run Apache plugin non-interactively


schoen:
As described at 

https://community.letsencrypt.org/t/trouble-automatically-renewing-with-a-script/11894

a hand-written `default-ssl.conf` containing various VirtualHosts is giving 

```
Failed to run Apache plugin non-interactively
Missing command line flag or config entry for this setting:
We were unable to find a vhost with a ServerName or Address of subdomain.client1.com.
Which virtual host would you like to choose?
```

when running `letsencrypt renew`. This looks like a problem with the `apache` plugin to me.

misterch0c:
+1

bmw:
Looking at the post on the community forum, I believe this is a dupe of #2320. Multiple virtual hosts per file is not currently supported by the Apache plugin (see #1042), however, the plugin fails to recognize the problem.

mgedmin:
I'm getting this problem when I have just two `<VirtualHost>` directives in the .conf file for a particular vhost (one for port *:80, one for port *:443, both with the same ServerName).

I'm pretty sure this used to work with older LetsEncrypt versions.

Can I ask the Apache authenticator plugin with a command-line option or something to please stop trying to verify that my Apache config is correct and just proceed with the TLS-SNI-01 challenge?

mgedmin:
I've been reading the source code again, trying to understand why the Apache *authenticator* plugin needs to parse virtual host files when it doesn't need to modify them.

The call is in tls_sni_01.py, in _get_addrs(): the plugin wants to know what IP addresses are used for the domains you're verifying.

pde:
Should have been fixed by #2906

frozen-solid:
Still having a problem with this bug as of today, though I'm wondering if my issue might be related to having a hyphen in the domain name. I have 4 hosts all setup with exactly the same configs. The only difference is the directory, document root, and server names. However, the domain with a hyphen keeps throwing the error still.

bmw:
@frozen-solid, how are you running the client? The fix in #2906 hasn't been packaged yet, so unless you're doing something like our [developer instructions](https://letsencrypt.readthedocs.io/en/latest/contributing.html), you won't have the fix yet. With that said, we're planning on doing a new client release, including that PR, later this week.

frozen-solid:
Oh, well in that case you're right I wasn't using the dev instructions. I
wrongly assumed running from the git source would be enough. Thanks for the
information, and sorry for cluttering the issue report.

Issue #2320, opened 2016-01-30
Client parses files with multiple vhosts when it should skip them


dwilson390:
The client does not correctly recognise what sites the Apache .conf (under sites-available) corresponds to.  Here is an example;

Screenshot; https://dyw.pm/i/T7gEMb.png

Actual contents;

dyw.pm.conf (LE claims multiple sites are hosted in this file);
```
<VirtualHost *:80>
	ServerName dyw.pm
	ServerAlias www.dyw.pm
	ServerAdmin contact@dwilson.me.uk
	DocumentRoot /var/www/dyw.pm/www/public/
	ErrorLog ${APACHE_LOG_DIR}/error.log
	CustomLog ${APACHE_LOG_DIR}/access.log combined
	RewriteEngine on
	RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [L,QSA,R=permanent]
  <ifmodule mpm_itk_module>
    AssignUserID dywpm dywpm
  </ifmodule>
</VirtualHost>

<IfModule mod_ssl.c>            
	<VirtualHost *:443>
		ServerName dyw.pm
		ServerAlias www.dyw.pm
		ServerAdmin contact@dwilson.me.uk
		DocumentRoot /var/www/dyw.pm/www/public/
		ErrorLog ${APACHE_LOG_DIR}/error.log
		CustomLog ${APACHE_LOG_DIR}/access.log combined
		SSLEngine on
[...]

PR #2654, opened 2016-03-12
don't include files that have multiple vhosts

SwartzCr:
should prevent files with multiple vhosts from getting parsed and offered as options for ssl enabled vhosts
this is because if augeas notices more than one vhost in a file it will list them as VirtualHost[1], VirtualHost[2], and so on:
realted to #2320 and #2600

### DIFF
--- a/letsencrypt-apache/letsencrypt_apache/display_ops.py
+++ b/letsencrypt-apache/letsencrypt_apache/display_ops.py
@@ -83,7 +83,8 @@ def _vhost_menu(domain, vhosts):
         code, tag = zope.component.getUtility(interfaces.IDisplay).menu(
             "We were unable to find a vhost with a ServerName "
             "or Address of {0}.{1}Which virtual host would you "
-            "like to choose?".format(domain, os.linesep),
+            "like to choose?\n(note: conf files with multiple "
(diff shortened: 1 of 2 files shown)


letsencrypt-apache/letsencrypt_apache/display_ops.py
```diff
@@ -83,7 +83,8 @@ def _vhost_menu(domain, vhosts):
         code, tag = zope.component.getUtility(interfaces.IDisplay).menu(
             "We were unable to find a vhost with a ServerName "
             "or Address of {0}.{1}Which virtual host would you "
-            "like to choose?".format(domain, os.linesep),
+            "like to choose?\n(note: conf files with multiple "
```
pde: Rather than including this case in the select vhost / missing flag message, I think we should raise a separate error.  There isn't a flag the user can supply that will solve things here, right?

pde: Maybe "not yet supported"? Otherwise this is actually.

pde: I shouldn't comment while doing other things at the same time. "Actually fine".

PR #2657, opened 2016-03-12
Update README.rst

YourDaddyIsHere:
Hello, I am a new comer, just start digest the codes of letsencrypt.
I am not so familiar with ACME, that makes it very hard to digest the codes. I guest some other new comers are facing similar  situation with me.
I notice that there are some links in README.rst aim to help new comer understand the architecture of letsencrypt, including links to: 
**Documentation:** https://letsencrypt.readthedocs.org
**Software project:**https://github.com/letsencrypt/letsencrypt
etc. However, there is no link to ACME specification, 
so I propose to add a link to the **ACME spec:** https://letsencrypt.github.io/acme-spec/    in **README.rst**, that will help new comers who are not familiar with ACME.

PR #2658, opened 2016-03-13
Fix the incorrect link in previous PR

YourDaddyIsHere:
I proposed to add a link to ACME spec README.rst in previous PR (#2657), that will make it easier for new comers who are not familiar with ACME.

But the ACME spec I linked to is not the Active Draft, now I fix that mistake and link to:
https://tools.ietf.org/pdf/draft-ietf-acme-acme-01.pdf

That is the document found in "Active Internet-Drafts" in IETF Datatracker: https://datatracker.ietf.org/doc/active/
I think that is the correct document


kuba:
The correct link is https://github.com/ietf-wg-acme/acme or http://ietf-wg-acme.github.io/acme.

YourDaddyIsHere:
................................................................................Alright, thank you, I will correct it...

YourDaddyIsHere:
@kuba 
Thank you for your help,  Travis check is still in progress, I will start a new PR and this one will be closed.

Issue #2632, opened 2016-03-09
Add an option to configure dual RSA/ECDSA certificates


jsha:
Related to #2625: In addition to / alternatively to explicitly choosing a key type, we should have a flag to automatically get both an RSA and an ECDSA cert, and offer both depending on what the client accepts. Eventually (after experimentation) we might want to make this the default.

Here's info on how to do it in Apache: https://blog.joelj.org/dual-rsaecdsa-certificates-in-apache-2-4/

nitmir:
:+1:

J0WI:
This is currently also planned for nginx 1.11:
https://trac.nginx.org/nginx/roadmap

jonnybarnes:
Indeed dual cert support was added to nginx v1.11.0: http://nginx.org/en/CHANGES

Yamakaky:
At least it should be supported by `certonly`.

jonnybarnes:
Do we have any more of an idea of the timescale for this?

STonEDSTonE:
Any update on this matter? This feature would be awesome!

udf2457:
Yeah guys.... where is it ?

jookk:
Here :O https://github.com/Neilpang/acme.sh
I am now using two clients.

sanmai:
This issue is still relevant.

udf2457:
Agree with @sanmai .  The only thing stale are the certbot developers. ;-)

bmw:
Since this issue was opened, I think the landscape has changed a bit. For instance, [Mozilla's TLS recommendations](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29) recommend an ECDSA cert over a RSA cert and using ECDSA has more compatibility with old clients than RSA with the cipher suites they recommend.

If anyone has a use case for this other than supporting clients older than [Mozilla supports in their intermediate config](https://wiki.mozilla.org/Security/Server_Side_TLS#Recommended_configurations), I'd love to hear about it.

jsha:
The proposal is to implement ECDSA, but without a "dual cert" option? That does seem like a nice way to unblock progress on #2625, since it would simplify a bunch of things. If anyone has concrete data on specific clients or OS' that break with ECDSA-only certs, that would be helpful.

rhymeswithmogul:
[...]

Issue #2625, opened 2016-03-09
Add --key-type flag


jsha:
The client should support a `--key-type` flag, values `rsa`, `p256`, or `p384`, to make it easy to generate ECDSA certificates.

nitmir:
:+1:

ageis:
Great idea; I love it, but don't forget that we presently default to SHA256 to my knowledge... so why not have an option allowing users to increase that SHA384 or SHA512 too?

Speaking of which, when will [BLAKE2b](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2) go mainstream!

Shouldn't Brainpool be in the curve list too? Or `ed25519`, and `sect###(k|r)1` since the others are @usNISTgov cuves? I still get confused sometimes between the cipher algorithms, key exchange, (ECDHE w/ RSA not bad) and HMACs/hashes/digests, sadly. :sob:

jsha:
FYI, see https://github.com/certbot/certbot/pull/2660; implementation is somewhat stalled out on decisions about storage.

Right now, the listed curves are the only ones valid for issuance under the BRs, but ed25519 will probably come down the pipe soonish.

ChadTaljaardt:
@jsha Has there been any updated regarding this?

lnicola:
I think there's still a lot of interest in ECDSA support.

TaaviE:
Boulder supports [ECDSA](https://github.com/letsencrypt/boulder/pull/1357) for a while now. What's blocking implementing this? PRs are going stale, roadmaps have been changed multiple times already.

ali-alsabbah:
Hello, I’d like to work on this.

TaaviE:
https://github.com/certbot/certbot/pull/7247 There's work being done on this already

ali-alsabbah:
> @ali-alsabbah there's work being done on this already here #7247 .

I see. Do you know of any issues where I can contribute? Should I just browse through the help-wanted label? I did join the Mattermost channel also.

TaaviE:
@ali-alsabbah 
Check out https://github.com/certbot/certbot/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy-fix

bmw:
This was added in https://github.com/certbot/certbot/pull/8444 and will be included in our next release.

PR #2660, opened 2016-03-13
ECDSA subject key support

### LINKED ISSUE #2163: Support ECDSA subject keys
Boulder now supports [ECDSA subject keys](https://github.com/letsencrypt/boulder/pull/1298), and although it isn't deployed just yet it will be soon.  So we may want to add acme and cli support for making P384 and P521 subject keys.

osirisinferi:
I've made a stub for ECDSA key support. It probably isn't the nicest work of coding, but it does work. :stuck_out_tongue: Fixes #2163

All comments are welcome, so please do :smile:


pde:
Tagging this for 0.6.0, although we can merge it sooner if it's ready sooner.

letsencrypt/cli.py
```diff
@@ -1583,6 +1584,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         "security", "--rsa-key-size", type=int, metavar="N",
         default=flag_default("rsa_key_size"), help=config_help("rsa_key_size"))
     helpful.add(
+        "security", "--ecdsa_curve", default=flag_default("ecdsa_curve"))
```
jsha: This should be --ecdsa-curve to match the other flags.

jsha: Also, it needs help text, and should list the available options.

jsha: And should validate its input against the available options. :-)

osirisinferi: Help text needs to be added indeed. Good point about the switch style.

For now, the options are hardcoded, but it should be configurable methinks.

letsencrypt/cli.py
```diff
@@ -1583,6 +1584,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+    helpful.add(
+        "security", "--privkey_signature_algorithm", default=flag_default("privkey_signature_algorithm"))
```
jsha: I would instead call this flag `--key-types` and make it a list of key types. Same comment above about listing options and validating them applies.

The reason to make it a list is that I expect that for people using the autoconfiguration features, they will most likely want to issue both an RSA cert and and ECDSA cert, and configure both of them in their web server, to be offered based on client handshake.

There's some additional groundwork to be laid for having multiple "live" certs per name, which I don't think has to happen in this change. The main thing is that if we're introducing a new flag, I'd like to have its name and semantics be what we want to use for the long term. For now, the validator can reject any list of key types with length greater than one.

letsencrypt/crypto_util.py
```diff
@@ -18,12 +18,16 @@
...
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec
 
 logger = logging.getLogger(__name__)
 
 
 # High level functions
-def init_save_key(key_size, key_dir, keyname="key-letsencrypt.pem"):
+def init_save_key(key_algo, rsa_key_size, ecdsa_curve, key_dir, keyname="key-letsencrypt.pem"):
```
jsha: This function now has a weird signature, in that you're required to specify `rsa_key_size` even if you're creating an ECDSA key. It probably makes more sense to change this function to `save_key(key_pem, key_dir...)` and have the code in client.py call `make_key_rsa` or `make_key_ecdsa` depending on the config, and then call `save_key` on the result.

letsencrypt/crypto_util.py
```diff
@@ -56,7 +68,10 @@ def init_save_key(key_size, key_dir, keyname="key-letsencrypt.pem"):
     with key_f:
         key_f.write(key_pem)
 
-    logger.info("Generating key (%d bits): %s", key_size, key_path)
+    if key_algo == "RSA":
+            logger.info("Generating RSA key (%d bits): %s", rsa_key_size, key_path)
```
jsha: I think this log function was originally in the wrong place, since it comes after the key is already generated. These should happen before the key is generated (and outside of this function, which is now about saving keys).

osirisinferi:
The tests need work.. But first I've got to learn more about tests in total :stuck_out_tongue:

pde:
@osirisinferi writing mock tests is often hard.  Feel free to come and ask us for help on IRC if you're stuck on things!

osirisinferi:
Uch, `tox` locally wasn't complaining, but Travis is.. :stuck_out_tongue:

osirisinferi:
I think the `test_success` and `test_key_failure` are broken, because `init_save_key()` is now split between `client.py` and `crypto_util.py`, correct?

`test_key_failure` is looking for a ValueError exception, which isn't in `crypto_util.py` anymore, but in `client.py`..

osirisinferi:
Well, at least all tests are succeeding now, but that's because I didn't include a test for `test_key_failure` yet.. Any help is much appreciated! (I'm idling in #letsencrypt :smile: )

Just the `lint` and `coverage` left..

bmw:
By splitting `make_key_rsa` out of `save_key`, `test_key_failure` isn't really necessary anymore.

To fix `cover`, you should write tests for functions like `make_key_ecdsa`. In general, we like for pull requests to have full test coverage for all modified lines.

As for `lint`, fixing the errors below [this line](https://travis-ci.org/letsencrypt/letsencrypt/jobs/117256314#L691) will solve the problem.

osirisinferi:
[AttributeError: 'MakeKeyECDSATest' object has no attribute 'assertIsInstance'](https://travis-ci.org/letsencrypt/letsencrypt/jobs/117562113#L1434) FFS :stuck_out_tongue_winking_eye: That attribute is new in 2.7......

bmw:
Great!

Yeah all of those tests sound good to me! If functions like `make_key_rsa` and `make_key_ecdsa` already have good test coverage, you can [mock](https://docs.python.org/dev/library/unittest.mock.html) out those functions in `obtain_certificate()` for relevant tests if it makes your life easier.

letsencrypt/cli.py
```diff
@@ -1583,6 +1584,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         "security", "--rsa-key-size", type=int, metavar="N",
         default=flag_default("rsa_key_size"), help=config_help("rsa_key_size"))
     helpful.add(
+        "security", "--ecdsa-curve", default=flag_default("ecdsa_curve"),
+        help="Set the certificate ECDSA curve. Current possible curves: "
+        "prime256v1 (default) or secp384r1")
```
jsha: Let's use the more easily remembered [FIPS 186-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf) names for these curves: P-256 and P-384. Also, please extract them to a constant and use it in both a validator for the flag and in obtain_certificate.

Also, should this be `config_help("ecdsa_curve")`?

jsha:
After thinking about this some more, I think that all files related to ECDSA certificates (and any new certificate types, like ed25519 in the future), should have a suffix so they can coexist in the same directory with RSA certificates. So they would be like `cert-ecdsa-1.pem`, `chain-ecdsa-1.pem`, `fullchain-ecdsa-1.pem`, `key-ecdsa-1.pem`. And they would be symlinked into the live directory under `cert-ecdsa.pem`, `chain-ecdsa.pem`, etc.

Ideally RSA keys would be named in the same pattern, but we're more or less locked into our existing naming scheme for them now, so we shouldn't change those.

osirisinferi:
The file naming part is gonna take some time I think.. Those filenames are everywhere and nowhere, at least when I `grepped` `fullchain` in the `/letsencrypt/` directory.. :confounded:

certbot/tests/client_test.py
```diff
@@ -214,12 +289,72 @@ def test_obtain_certificate(self, mock_crypto_util):
...
+                mock.MagicMock(
+                    body=mock.MagicMock(
+                        identifier=mock.MagicMock(
+                            value=domain))))
+
+        self.client.auth_handler.get_authorizations.return_value = authzr
+
+        self.assertRaises(errors.Error,
+            self.client.obtain_certificate(domains),
+            (mock.sentinel.certr, mock.sentinel.chain, mock.sentinel.key, csr))
```
mgedmin: Here you're calling `self.client.obtain_certificate(domains)`, and then passing the result to `self.assertRaises()`. If `obtain_certificate()` raises an exception, which it does, then `self.assertRaises()` will never be called.

You want
```python
with self.assertRaises(errors.Error):
    self.client.obtain_certificate(domains)
```
or
```
self.assertRaises(errors.Error, self.client.obtain_certificate, domains)
```

I'm not sure what `(mock.sentinel.certr, ...)` are supposed to be -- the assertion message?  You cannot pass a custom assertion message to `assertRaises`, the API [doesn't allow that](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertRaises).

pde:
Looks like this is probably going to miss the 0.6.0 train? Not sure if it's close or if @mgedmin or @jsha might pop up with a last minute "LGTM".

jsha:
This will miss 0.6.0. We shouldn't ship without the filename fixes or it will be harder to ship the right thing later.

pde:
Kicking travis.

J0WI:
It seems that this PR fixes #2163 as well as #2625.
But I wonder if we really need both. Can't we just use a syntax like `--key-type rsa-4096` `--key-type ecdsa-p256` or `--key-type rsa-4096 ecsda-p256` (for generating both, see #2632).
This would also be more open for whatever is coming in future.
For backwards compatibility `--rsa-key-size` could be kind of an alias for `--key-type rsa-N`.

jsha:
Technically RSA isn't limited to 2048 or 4096 - Boulder will sign, e.g. a 2096-bit RSA key. That means that `--key-type rsa-2096` would involve parsing the `rsa-2096` string as a microformat, which I'd prefer to avoid.

Another way to put it: RSA is a key type, and comes in different sizes. But `P256` and `P384` are considered distinct key types in the standards because they use different curves.

osirisinferi:
Well, I wouldn't call P256/P384 a key type. The "key type" is ECDSA, P256/P384 is the parameter. The curves can also be used in ECDH and aren't limited to ECDSA.

jsha:
Good clarification, thanks!

J0WI:
> Technically RSA isn't limited to 2048 or 4096 - Boulder will sign, e.g. a 2096-bit RSA key.

Do we want to support exotic key lengths?
<2048 isn't allowed anyway.

> Well, I wouldn't call P256/P384 a key type. The "key type" is ECDSA, P256/P384 is the parameter. The curves can also be used in ECDH and aren't limited to ECDSA.

Okay, that make sense.

But what about my suggestion to support multiple arguments on `--key-type`?

jsha:
There's discussion at https://github.com/letsencrypt/boulder/issues/1599#issuecomment-195587275 of why it's useful to allow "interesting" RSA key lengths.

jsha:
@osirisinferi: Any progress on getting the file naming sorted out? @schoen are you able to help @osirisinferi work that out?

schoen:
@jsha, I think that @pde and @bmw and I are going to need to have a further discussion to reach consensus about how the keys should be represented (in particular, the combined-lineage vs. split-lineage issue). We talked about this a bit during our call yesterday and didn't immediately agree about how it should work.

pde:
Kicking this to 0.10.0 since it's blocked on proper lineage support in `storage.py`, which @schoen is going to work on, but not until after we have nginx tests working really well in 0.9.0.

pde:
Kicking to 0.11.0 since although we may have the lineage stuff sorted out for 0.10.0, it's going to be landing at the last minute...

bmw:
Kicking this to 0.11.0. There's a design discussion that needs to happen about how lineages work while supporting multiple key types.

jonnybarnes:
Is there an issue with public discussion about the design decision regarding lineages?

jsha:
@jonnybarnes Not previously, but I'll summarize it here:

Right now, we store four files in `/etc/letsencrypt/archive/example.com/`: `certN.pem`, `chainN.pem`, `fullchainN.pem`, `privkeyN.pem`, where N increases over time. We also symlink the most recent of each of these files into `/etc/letsencrypt/live/example.com`, with the N removed, so server software can be configured with a reliable filename. This symlinking step is considered "atomic"[1], so that all four files in `live/` match each other at any given time.

The question is: If you have identical certificates for multiple key types, should we store them in separate lineages (i.e. `archive/example.com` for RSA certs and `archive/example.com-ecdsa` for ECDSA certs), or should we consider them part of the same lineage, adding new filenames to the collection (i.e. `archive/example.com/ecdsa-certN.pem`, `archive/example.com/ecdsa-privkeyN.pem`).

Mostly this comes down to a question of what we want to do if renewal fails for one key type, but succeeds for the other. Do we consider both renewals to have failed, and retry both later? Or do take the successful one and configure it, and retry only the failed one later? The former implies the "same lineage" approach. The latter implies the "separate lineages" approach. I think we are slowly arriving at consensus on the "separate lineages" approach. Is that your feeling too, @bmw, @ohemorange, @ynasser?

[1] Note that the symlinking is not actually atomic in the Unix operation sense, and a poorly-timed crash could result in an inconsistent state, but that is a totally different issue.

J0WI:
This big question about the naming convention affect only the scenario of #2632. This bug is a blocker of #2632, but it does not depend on it. If you just want to use ECDSA, you can continue using the same naming convention and change it later, together with dual cert support. Another usecase is, if you using Certbot just to get an ECDSA cert, which you put manually to your preferred location.
Other ACME clients already support ECDSA certs. So I think it would be helpful, when Certbot at least have the switch for issuing ECDSA certs.

jsha:
@J0WI: I agree that I'd like to get this landed soon. But I do think the storage decision is a blocker. We'd like to store ECDSA certificates in a consistent place, so we don't change assumptions on people in the future, after they've come to depend on certificates getting stored in a certain location. Note that of course you can still use the `--csr` flag to manually issue ECDSA certs with Certbot, even though it's a pain (and doesn't support auto-install).

bmw:
FWIW, when @schoen and I specced this out between the two of us we settled on separate lineages but @pde was strongly in favor of putting things in a single lineage to more easily offer things like configuring your webserver to use both RSA and ECDSA certificates.

jonnybarnes:
@bmw well, looking at nginx, you just need to specify `ssl_certificate` and `ssl_certificate_key` twice, one pointing to the rsa version, and one pointing to the ecdsa version. It doesn’t really matter wther there in seprate folders so long as its consistent.

bmw:
Yep. We can do it either way, it's just easier due to Certbot's architecture to have them in the same lineage. For example, currently at most one lineage is created during a run and is then passed around through Certbot code. If we put them in separate lineages, we'd have to change this or have people run Certbot twice with some kind of flag/prompt to distinguish between people who want to replace the certificates for a domain and those who want to add additional ones. Certainly not an unsolvable problem, but I believe @pde's position was putting them in the same lineage could save us a lot of work.

Responding to earlier in this thread though, I think we may be able to kick this decision to a later date. Regardless of how support this long term, I suspect we'll want people to be able to request a lineage containing only an RSA or ECDSA certificate. If this is the case, we could support ECDSA certs in their own lineage for now and potentially offer dual lineages later. I'd definitely have to get more of the Certbot team on board with this decision though.

bmw:
To keep this thread up to date, here's my IRC messages about the current status of this:
```
2017-03-03 07:46:53	+bmw	OsirisInferi: my inclination wasn't to make a distinguished name for the lineage, but I brought it up the rest of the Certbot team and got disagreement from some people
2017-03-03 07:49:03	+bmw	their concern was if we want to support a single lineage with both key types, we'll need to change file names with the current thought to make the ECDSA files "cert-ecdsa.pem", "privkey-ecdsa.pem", etc.
2017-03-03 07:49:24	+bmw	and if we're going to do that, let's be consistent and always do that
2017-03-03 07:49:35	+bmw	so unfortunately, we're back to blocking on the larger decision on how to support both :(
2017-03-03 07:49:41	+bmw	I'll try to push the decision along though
```

jgillula:
@bmw should we keep this PR open?

bmw:
I think @ohemorange should make the call here as she wanted to own this task when she finds the time. Personally, I don't care too much whether or not this PR stays open, but if we close it, we should keep this code around in a branch somewhere to build off of.

osirisinferi:
It has been a whole year since the last "we still need to make a decision" bit.

In the mean time, this soccer ball has become a nightmare (for me) to manage with all the frequent updates and subsequent conflicts. I've got very little spare time nowadays and I mostly made these changes so I could generate ECDSA certificates easily on my server.

Therefore, I've made [a new ECDSA branch on my own fork](https://github.com/osirisinferi/certbot/tree/ecdsa-v2), forgo all the tests and just made it working for me.

Personally, I don't care either if you keep this PR open or not. *I* have ECDSA support for myself.. I just want to mention I don't think it's very polite if people take the effort in helping to improve `certbot` and subsequently just closing the PR without much gratitude shown. Wouldn't be the [first](https://github.com/letsencrypt/website/pull/116) [time](https://github.com/letsencrypt/website/pull/211).

Further more, on the Community, people are frequently asking about ECDSA support. I understand Let's Encrypt/EFF certbot team is small and has many pressing issues to resolve, but one might think this choice might have been resolved within a year.

bmw:
@osirisinferi, I'm sorry we haven't been able to land this yet and maintenance has been time consuming for you. I'm also sorry that we and the Let's Encrypt team haven't seemed appreciative to you for the time you've spent contributing on the project.

I can assure you that we are grateful for your work and we have every intention of using this PR. If we close this, I want us to add your code to a branch on the Certbot repo so it can be used as a base to build off to finish our ECDSA support. I probably should have made it clearer in my recent comment that the reason I want this if we close this PR is because the work that's been done here is valuable addition to the project that should be used in the future.

I won't try and estimate when this feature will land or justify why it hasn't landed yet, but we are certainly still planning on adding this to Certbot. While we've prioritized different pieces of the project, I'm glad you've found a solution that works for you in the meantime and I'm hoping we can finally get around to adding this feature sooner rather than later.

gbdlin:
Is there any decision about if this should live in the same lineage or different?

gbdlin:
Actually, isn't this just not an issue at all? We can use `--cert-name` to save ecdsa certificate in separate lineage, so they're separated. That way you can have both RSA and ECDSA and it implies minimal amount of changes.

Issue #2661, opened 2016-03-13
none ascii domains, error catching failed


puggan:
This code should catch UnicodeDecodeError?
```
    # Unicode
    try:
        domain = domain.encode('ascii').lower()
    except UnicodeDecodeError:
        raise errors.ConfigurationError(
            "Internationalized domain names are not presently supported: {0}"
            .format(domain))
```
So why is my log saying:

> 
> 2016-03-13 18:57:09,064:DEBUG:letsencrypt.cli:Exiting abnormally:
> Traceback (most recent call last):
>   File "./letsencrypt", line 11, in <module>
>     sys.exit(main())
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/cli.py", line 1993, in main
>     return config.func(config, plugins)
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/cli.py", line 658, in run
>     domains = _find_domains(config, installer)
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/cli.py", line 133, in _find_domains
>     domains = display_ops.choose_names(installer)
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/display/ops.py", line 208, in choose_names
>     names = get_valid_domains(domains)
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/display/ops.py", line 242, in get_valid_domains
>     valid_domains.append(le_util.enforce_domain_sanity(domain))
>   File "/root/.local/share/letsencrypt/lib64/python2.7/site-packages/letsencrypt/le_util.py", line 312, in enforce_domain_sanity
>     domain = domain.encode('ascii').lower()
> UnicodeEncodeError: 'ascii' codec can't encode character u'\xe5' in position 1: ordinal not in range(128)
> 

row 312 is the `domain = domain.encode('ascii').lower()` row, LE version 0.4.2

nitmir:
UnicodeDecodeError != UnicodeEncodeError
u"unicode string".encode() --> UnicodeEncodeError
b"byte string".decode() --> UnicodeDecodeError

puggan:
oh, missed that one, so guess it should be pathced to "except UnicodeEncodeError:",
tried to replace it with just "except:", but then it just moved the errormessage to  the ".format(domain))" row.

[...]

PR #2662, opened 2016-03-13
fixed error catching in enforce_domain_sanity(), github issue #2661 [needs revision] [needs tests]

puggan:
Python isn't my main language, but I think this will fix it.
Hope someone with python skills can review it.


pde:
Thanks for the review @nitmir .  We'll wait for that change before reviewing further.  This will also need tests; @puggan if you haven't written Python before, I think we should get someone else to help with that; working with `mock` definitely adds an extra degree of magic and weirdness to Python.

pde:
We wound up with two PRs for this... merging #2665 since it's ready without extra work.

Issue #2277, opened 2016-01-26
Do not save default values to renewal configuration files


schoen:
If a client configuration parameter is set to its default value and wasn't specified by the user in some way, it shouldn't be saved into the renewal configuration at all. If it was derived from another configuration file, we can allow it to be derived from that file again rather than saved in a duplicative way.

schoen:
Related to #2503.

pde:
It's probably a good idea to do something like this:

if you have a default value for a config entry, and the RCF is new or does not contain the string for that entry anywhere, add a comment to the RCF showing that that setting is available.  Eg:

# DEFAULT VALUES
# rsa-key-size = 2048

bmw:
Also see #421 for more context.

bmw:
Closed in #2663.

Issue #2651, opened 2016-03-11
_set_by_cli spuriously flags all plugin variables to be set by cli


schoen:
In new code intended to fix #2503, I use `_set_by_cli` as one criterion for deciding to save something in the renewal config file. I found that it sometimes returns `True`, resulting in saving items, even when I did not actually set those things from the CLI at all. The items in question that I observed so far were these:

```
apache_enmod = a2enmod
apache_server_root = /etc/apache2
apache_dismod = a2dismod
apache_handle_sites = True
apache_handle_modules = True
standalone_supported_challenges = "tls-sni-01,http-01"
apache_le_vhost_ext = -le-ssl.conf
apache_challenge_location = /etc/apache2
nginx_server_root = /etc/nginx
nginx_ctl = nginx
apache_vhost_root = /etc/apache2/sites-available
```

There could be others, but these were ones I saw so far.

Cc @pde

pde:
Noah expressed interest in this...

SwartzCr:
The issue is that plugin_arguments get added through a call to `add_plugin_args` which in turn pulls plugin parsers from `six` and calls `add_argument`
This presents a few problems.

1. The `modify_arg_for_default_detection` line is in HelpfulArgParser's `add` command, not in `add_argument`
2. The parsers used for this are `SilentParsers` not helpful
3. These parsers don't necessarily have detect_default set

it'd be nice to have SilentParser have the same implementation of HelpfulArgParses modify_arg_for_default_detection, and then we could add a detect_default flag to it before adding plugin args - that would solve this problem

bmw:
I'm going to take this.

PR #2663, opened 2016-03-14
Don't save irrelevant items into renewal configuration

### LINKED ISSUE #2503: Do not save irrelevant values to renewal conf files
Only things that would actually be read back should be written out.

schoen:
Fix #2503.

#2651 slightly limits the awesomeness of this fix.

I think the basic functionality is correct now. Not ready to merge yet because tests still need updating.


schoen:
Also fixes #2277 because the logic here checks for both reasons not to save something into the renewal config.

schoen:
Let's see how this does with Travis. I think I've now avoided making test coverage worse, although there are more meaningful tests that we could now add to directly confirm what does and doesn't get saved in renewal config files. I think those tests might be best added after #2651 is fixed.

letsencrypt/cli.py
```diff
@@ -537,6 +537,20 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
     raise errors.PluginSelectionError(msg)
 
 
+def _relevant(option):
```
pde: Since it's being called from other modules, this function should not be private (ie not start with underscore).

bmw:
#2669 should fix integration test failures

letsencrypt/storage.py
```diff
@@ -127,6 +122,59 @@ def get_link_target(link):
...
 
+def _relevant(option):
+    """
+    Is this option one that could be restored for future renewal purposes?
+    :param str option: the name of the option
+
+    :rtype: bool
+    """
+    # The list() here produces a list of the plugin names as strings.
+    from letsencrypt import cli
```
SwartzCr: is it bad that these are imported in a helper function? Will that cause them to get imported multiple times per call to relevant_vaules or does python do some optimazation for this

SwartzCr:
besides the one mentioned import issue this code looks good (and straightforward) to me

letsencrypt/storage.py
nitmir: Python will import module only once independently of the number or `import module`. To reimport a module, one explicitly must call `reload(module)`.
But it is considered bad practice to not do import at the top of the file: https://www.python.org/dev/peps/pep-0008/#imports.

bmw:
Merging with review from @SwartzCr.

PR #2664, opened 2016-03-14
Don't suggest --duplicate; it's likely to confuse people

### LINKED ISSUE #2602: Why do we recommend --duplicate as a way of expanding a cert?
In `cli.py` at line 363 we say

https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/cli.py#L363

"--expand (or, in some cases, --duplicate)".

These two options have opposite behaviors with respect to controlling the creation of new lineages. What are the "some cases" in which `--duplicate` would be appropriate to "expand and replace th[e] existing certificate with the new certificate"? Could we remove the parenthesis there? I think it's even leading people astray to try `--duplicate` when it's not useful to them at all.

schoen:
Fixes #2602.

(I saw on the forum that someone was confused by the existing message into thinking that `--duplicate` was a synonym for `--expand`, when in fact they are opposites!)

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -360,7 +360,7 @@ def _handle_subset_cert_request(config, domains, cert):
              br=os.linesep)
     if config.expand or config.renew_by_default or zope.component.getUtility(
             interfaces.IDisplay).yesno(question, "Expand", "Cancel",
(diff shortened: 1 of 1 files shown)


pde:
Great

PR #2667, opened 2016-03-15
Add --must-staple flag

### LINKED ISSUE #2626: Add --must-staple flag
We should have a flag that adds the Must Staple extension in CSRs, per https://tools.ietf.org/html/rfc7633.

jsha:
Fixes #2626


pde:
Reading the help for this flag, one might naively expect it to integrate with the Apache (and Nginx?) plugins to actually do the stapling.  If not it should be clearly documented that it simply changes a property of the obtained cert and that (for now) the user will have to set things up manually.

pde:
@sagi would you be interested in implementing the plugin-side of a must staple enhancement?

jsha:
Clarified the help text to indicate it doesn't do the autoconfiguration.

sagi:
@pde yes! thanks. I'll start working on it this weekend :)

jsha:
The relevant issue for the enhancement part is at https://github.com/letsencrypt/letsencrypt/issues/930. Leaving this PR open for now in case it requires modification as part of that work.

pde:
@sagi have you started looking at this yet? Any thoughts on whether we should merge this PR as is, or wait for possible tweaks to match an Apache OCSP stapling component?

sagi:
@pde yes. Lets call the extension **X**, the minimum apache version that supports OCSP Stapling **V**, Letsencrypt client **L**, the `--must-staple` flag **F** and a flag that disables **X** by **NX**

My thoughts are:

The gist is that **X** should be enabled by default and that **X** adds both the `must stable` to the CSR and defines the apache config for OCSP Stapling (see [this](https://github.com/letsencrypt/letsencrypt/issues/930#issuecomment-199927627) for more context).

pseudocode:
```
If apache >= V:
    if NX is off:
        L disregards F and applies X.
    else:
        L regards F and doesn't apply X.
else:
    L disregards NX, regards F and doesn't apply X.
```

What are your thoughts @jsha, @pde ?

jsha:
Strongly disagree. Must Staple is still very new, and many servers (including Apache) are likely to have problems. I think the Must Staple extension should only be by request from adventurous users, but Stapling can be turned on by default because it's very low risk.

sagi:
I [erred](https://github.com/letsencrypt/letsencrypt/issues/930#issuecomment-199997173). I agree with @jsha.

jsha:
Now that 0.5.0 is released, can we land this?

pde:
This never got a "LGTM" from Sagi.  Sagi, should we merge it?

pde:
And @jsha, want to merge master?

sagi:
LGTM.

PR #2670, opened 2016-03-16
handle permission denied during cleanup

SuperTux88:
if the .well-known/acme-challenge was created before and the user has no
permissions to delete it, it failed at cleanup.


bmw:
LGTM. Thanks for the PR @SuperTux88.

Issue #2674, opened 2016-03-16
-v should imply --text


jsha:
If you specify the -v flag without --text, the verbose logging gets sent into an ncurses text box, and is impossible to read. We should automatically assume --text when -v is specified.

pde:
This should have been fixed (#2338 / #2389).  Under what circumstances are you still seeing it?

pde:
Citation: https://github.com/letsencrypt/letsencrypt/pull/2338/files#diff-54b2f468d57ef7436008bd031bc89711R1742

bmw:
@pde, I believe what you're talking about only applies when using `renew/--noninteractive`. When running `letsencrypt` with neither of those flags, logging is sent to `dialog` (in addition to log files) regardless of the level of verbosity.

pde:
(Didn't make the cut in our 0.5.0 triage meeting)

bmw:
Fixed in #3064.

PR #2684, opened 2016-03-17
Refactor HelpfulArgumentParser homework (was: renew implies noninteractive earlier) [brought in with tests in 3046]

pde:
#2674 got me looking at how we establish noninteractivity, and it seems that it's cleaner to do this earlier, when we construct our configuration.


pde:
This has now turned into a more general refactor of `HelpfulArgumentParser.parse_args`.

letsencrypt/cli.py
```diff
@@ -314,58 +312,55 @@ def parse_args(self):
...
-                                   "cannot be used with --csr")
             self.handle_csr(parsed_args)
 
         hooks.validate_hooks(parsed_args)
 
         return parsed_args
 
+
+    def set_test_server(self, parsed_args):
+        "We have --staging/--dry-run; perform sanity check and set config.server"
```
bmw: nit: `"""` should be used for docstrings.

letsencrypt/crypto_util.py
```diff
@@ -171,6 +172,35 @@ def csr_matches_pubkey(csr, privkey):
...
+
+    :returns: (le_util.CSR object representing the CSR,
+               OpenSSL FILETYPE_ representing DER or PEM,
+               list of domains requested in the CSR)
+    """
+    try:
+        csr = le_util.CSR(file=csrfile, data=contents, form="der")
+        typ = OpenSSL.crypto.FILETYPE_ASN1
+        domains = get_sans_from_csr(csr.data, OpenSSL.crypto.FILETYPE_ASN1)
+    except OpenSSL.crypto.Error:
```
bmw: The body of this `except` statement doesn't have coverage. You should write a unit test for it. It should be pretty easy to write test(s) for the three cases:

1. CSR in DER format
2. CSR in PEM format
3. Malformed CSR

letsencrypt/crypto_util.py
```diff
@@ -171,6 +172,35 @@ def csr_matches_pubkey(csr, privkey):
         return False
 
 
+def import_csr_file(csrfile, contents):
+    """Import a CSR file, which can be either PEM or DER.
+
+    :param str csrfile: CSR filename
+    :param str contents: contens of the CSR file
+
+    :rtype: tuple
```
bmw: The returns documentation can be cleaned up a bit:

1. Everywhere else, we put `:returns:` before `:rtype:`. I think we should continue doing this for consistence of our code and documentation.
2. The order of the elements in the `returns` tuple is wrong.
3. We get errors when building the docs due to this line:

```
OpenSSL FILETYPE_ representing DER or PEM,
```

Error was:
```
letsencrypt/crypto_util.py:docstring of letsencrypt.crypto_util.import_csr_file:8: ERROR: Unknown target name: "filetype".
```

Replacing that line with something like [this](https://github.com/letsencrypt/letsencrypt/pull/2684/files#diff-31d015f64b83c8ff85723c569cec870fR268) would fix the problem.

bmw:
Other than a few documentation, testing, and stylistic complaints. I think this looks good!

PR #2690, opened 2016-03-20
IPv6 support for Addr and correct parsing of IPv6 vhosts

### LINKED ISSUE #1143: Handle IPv6 adresses in Apache vhost entries

joohoi:
Continued work on @TheBoegl PR #2478 
Added IPv6 address comparison based on exploded addresses and Apache plugin test cases.

Fixes #1143


pde:
@SwartzCr want to give this a final review and merge?

letsencrypt-apache/letsencrypt_apache/tests/apache-conf-files/passing/ipv6-1143b.conf
```diff
@@ -0,0 +1,18 @@
...
+</Directory>
+
+  SSLEngine on
+
+  SSLHonorCipherOrder On
+  SSLProtocol all -SSLv2 -SSLv3
+  SSLCipherSuite "EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 EECDH EDH +aRSA RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS"
+
+   SSLCertificateFile /etc/ssl/certs/ssl-cert-snakeoil.pem
+   SSLCertificateKeyFile /etc/ssl/private/ssl-cert-snakeoil.key
```
SwartzCr: Just checking that this line was removed on purpose:
Header set Strict-Transport-Security "max-age=31536000; preload"

letsencrypt/plugins/common.py
```diff
@@ -104,14 +104,24 @@ class Addr(object):
...
     def fromstring(cls, str_addr):
         """Initialize Addr from string."""
-        tup = str_addr.partition(':')
-        return cls((tup[0], tup[2]))
+        if str_addr.startswith('['):
+            # ipv6 addresses starts with [
+            endIndex = str_addr.rfind(']')
+            host = str_addr[:endIndex + 1]
+            port = ''
+            if len(str_addr) > endIndex + 2 and str_addr[endIndex + 1] == ':':
```
SwartzCr: is it true that the port needs to be greater than a single character?

letsencrypt/plugins/common.py
```diff
@@ -136,7 +155,41 @@ def get_port(self):
...
+
+    def get_ipv6_exploded(self):
+        """Return IPv6 in normalized form"""
+        if self.ipv6:
+            return ":".join(self._normalize_ipv6(self.tup[0]))
+        return ""
+
+    def _explode_ipv6(self, addr):
+        """Explode IPv6 address for comparison"""
+        result = ['0', '0', '0', '0', '0', '0', '0', '0']
```
SwartzCr: nit: it might be easier to read this as `result = ["0"]*8` though I'm not sure how python 3 will like this

letsencrypt/plugins/common.py
```diff
@@ -136,7 +155,41 @@ def get_port(self):
...
+        addr_list = addr.split(":")
+        append_to_end = False
+        for i in range(0, len(addr_list)):
+            block = addr_list[i]
+            if len(block) == 0:
+                # encountered ::, so rest of the blocks should be
+                # appended to the end
+                append_to_end = True
+                continue
+            elif len(block) > 1:
```
SwartzCr: I don't see why this can't just be an if

letsencrypt/plugins/common.py
```diff
@@ -136,7 +155,41 @@ def get_port(self):
...
+                # remove leading zeros
+                block = block.lstrip("0")
+            if not append_to_end:
+                result[i] = str(block)
+            else:
+                # count the location from the end using negative indices
+                result[i-len(addr_list)] = str(block)
```
SwartzCr: I'm always suspect when len(...) is used in list indexing, is this right or should it be like len(...) - 1 or range (1, len(addr_list)), etc.

letsencrypt/plugins/common.py
```diff
@@ -104,14 +104,24 @@ class Addr(object):
...
-        tup = str_addr.partition(':')
-        return cls((tup[0], tup[2]))
+        if str_addr.startswith('['):
+            # ipv6 addresses starts with [
+            endIndex = str_addr.rfind(']')
+            host = str_addr[:endIndex + 1]
+            port = ''
+            if len(str_addr) > endIndex + 2 and str_addr[endIndex + 1] == ':':
+                port = str_addr[endIndex + 2:]
+            return cls((host, port), True)
```
SwartzCr: nit: personally I'd write this as cls((host, port), ipv6=True)

SwartzCr:
A few nits, and a few simple questions, but overall this looks good

letsencrypt-apache/letsencrypt_apache/tests/apache-conf-files/passing/ipv6-1143b.conf
joohoi: Yeah, it was, Travis doesn't have mod_headers, so it kept failing ( https://travis-ci.org/letsencrypt/letsencrypt/jobs/117270310). Also mod_headers is outside of the scope of this PR (and test).

letsencrypt/plugins/common.py
joohoi: It can, but I prefer using elif when doing something switch-like.

letsencrypt/plugins/common.py
joohoi: No it's actually not. I'd suspect ports < 10 are failing currently. Going to fix that, good catch.

letsencrypt/plugins/common.py
joohoi: Yes, that's clearer, fixing it.

letsencrypt/plugins/common.py
joohoi: It's fine with Python 3. However I myself prefer it this way, as it pretty much visually selfdocuments the inner workings of this method.

letsencrypt/plugins/common.py
joohoi: It is correct, however I'm not currently checking if len(addr_list) <= len(result) in this method which - if false, would lead to IndexError. Going to add that check.

SwartzCr:
looks good to me! merging

PR #2691, opened 2016-03-20
Changed testdata directory and test class name to a better fitting one

joohoi:
This is about TwoVhost80Test.

Janitor duty, changing test names to a better fitting ones, as they have been out of sync with reality for a long time.


pde:
LGTM

PR #2693, opened 2016-03-21
Fix symlink webroot paths

krukas:
When using symlinks for webroot folders and requesting SSL for same websites with symlink domains.
The challenges files and acme-challenge directory are removed on the first domain and on second domain the
acme-challenge directory is already removed and tries to remove it again.


pde:
Let's review this for landing in 0.6.0.

PR #2700, opened 2016-03-22
letsencrypt-auto fix of the --help parameter so nothing is installed.

nyxz:


pde:
This isn't a good fix for this problem.  Here are some alternatives:

1. Special case help in `letsencrypt-auto`, so that `--help` runs without a full install (I don't think that's a very good plan, but some version of it might work)
2. Update the docs so that the first command is something other than `--help`, such that the installation isn't suprising.
3.  Maybe add a new flag to `letsencrypt-auto` to explicitly request OS bootstrapping and venv setup.  Something like `letsencrypt-auto --setup` or `letsencrypt-auto --install-software`.  One thing that's tricky is that we already have an "install" command, which confusingly collides with the most self-explanatory flag name you would use. This would go well with 2.

bmw:
4) If `-h` or `--help` is provided on the command line and `letsencrypt` is not installed (e.g. there isn't a venv), continue execution but print a message saying that help for the `letsencrypt` program cannot be provided until it is installed. Furthermore, we stop forcing yes when installing OS packages (e.g. no more `apt-get install -y`) which will allow people who don't want `letsencrypt --help` to install packages to stop the program.

nitmir:
It could also be nice the have a Yes/No prompt before the first installation of the venv as it is possible that all packages from OS bootstrap are already installed.

pde:
@nitmir we can't really add a new (Y/N) prompt without breaking existing integrations people have built.  But we will consider how to get the UI for this better as part of our client renaming task (since the new command line name can behave slightly differently).

pde:
ah but yes, perhaps in the case of --help we can add a (Y/N), because nobody should be calling --help from a script...

bmw:
@pde, people have built integrations around the first run of `letsencrypt-auto` that installs OS packages and sets up the venv? Have you seen any of these?

PR #2702, opened 2016-03-22
Mention that we're talking about Unix OSes

schoen:
One-word documentation edit to implement a suggestion from the forums to clarify that this documentation is talking about Unix OSes (a Windows user spent a long time after reading this unsure what OS-level packaging referred to).


bmw:
LGTM

PR #2707, opened 2016-03-23
Fixed Gentoo define command

### LINKED ISSUE #2701: letsencrypt-apache on Gentoo should read /etc/conf.d/apache
In Gentoo, Define's are passed by the initscript as -D arguments read in /etc/conf.d/apache

LE seems to ignore that. As a result, since the "Listen 443" directive is inside IfDefine blocks, it is systematically overlooked by LE since it doesn't know about the active Define directives.

LE will therefore add a "Listen 443" temporary directive, which will cause apache to fail with a "could not bind to address 0.0.0.0:443" error. LE in turn will fail with "urn:acme:error:connection" since apache is not running during the challenge.

joohoi:
Found the way in Gentoo to get the active Define statements.

Fixes #2701


joohoi:
@osirisinferi It would be great if you have time to test this out!

osirisinferi:
Well, actually, `letsencrypt:master` isn't as broken anymore as it used to be on Gentoo, but your branch works perfectly indeed. Nice!

I'll test it on my VirtualBox Gentoo too.

pde:
OK, merging based on @osirisinferi's feedback.

Issue #1157, opened 2015-10-27
Webroot plugin not visible from menu


kuba:
This is because it requires `--webroot-path` to be set.

bmw:
With me taking #2651, this is probably going to get kicked out of the 0.5.0 milestone. I'll leave it here for now, but I doubt I'll get both this and #2651 done in time for the next release.

PR #2708, opened 2016-03-23
Add directory_select method to IDisplay

### LINKED ISSUE #445: File selection display
When prompting a user for a path (e.g. when the key cannot be found in Proof of Possession), the user is currently given a simple text box. It would be better if we gave the user a dialog such as http://pythondialog.sourceforge.net/doc/widgets.html#file-or-directory-selection to help them find and type the desired path correctly.

bmw:
Fixes #445 (the first issue I filed against LE during my first day at EFF).

Progress towards #1157.

### DIFF
--- a/letsencrypt/display/completer.py
+++ b/letsencrypt/display/completer.py
@@ -1,22 +1,22 @@
-"""Provides tab autocompletion when prompting users for a path."""
+"""Provides tab completion when prompting users for a path."""
 import glob
 import readline
 
 
 class Completer(object):
-    """Provides tab autocompletion when prompting users for a path.
+    """Provides tab completion when prompting users for a path.
 
     This class is meant to be used with readline to provide tab
-    autocompletion for users entering paths. The complete method can
-    be passed to readline.set_completer directly, however, this function
+    completion for users entering paths. The complete method can be
+    passed to readline.set_completer directly, however, this function
     works best as a context manager. For example:
 
     with Completer():
         raw_input()
 
-    In this example, tab autocompletion will be available during
-    the call to raw_input above, however, readline will be restored to
-    its previous state when exiting the body of the with statement.
+    In this example, tab completion will be available during the call to
+    raw_input above, however, readline will be restored to its previous
+    state when exiting the body of the with statement.
 
     """
 
@@ -24,7 +24,7 @@ class Completer(object):
         self._completer = self._delims = self._iter = None
 
     def complete(self, text, state):
-        """Provides path autocompletion for use with readline.
+        """Provides path completion for use with readline.
 
         :param str text: text to offer completions for
         :param int state: which completion to return
--- a/letsencrypt/display/completer.py
+++ b/letsencrypt/display/completer.py
@@ -1,4 +1,4 @@
-"""Provides tab completion when prompting users for a path."""
+"""Provides Tab completion when prompting users for a path."""
 import glob
 # readline module is not available on all systems
 try:
@@ -8,9 +8,9 @@ except ImportError:
 
 
 class Completer(object):
-    """Provides tab completion when prompting users for a path.
+    """Provides Tab completion when prompting users for a path.
 
-    This class is meant to be used with readline to provide tab
+    This class is meant to be used with readline to provide Tab
     completion for users entering paths. The complete method can be
     passed to readline.set_completer directly, however, this function
     works best as a context manager. For example:
@@ -18,14 +18,14 @@ class Completer(object):
     with Completer():
         raw_input()
 
-    In this example, tab completion will be available during the call to
+    In this example, Tab completion will be available during the call to
     raw_input above, however, readline will be restored to its previous
     state when exiting the body of the with statement.
 
     """
 
     def __init__(self):
-        self._completer = self._delims = self._iter = None
+        self._iter = self._original_completer = self._original_delims = None
 
     def complete(self, text, state):
         """Provides path completion for use with readline.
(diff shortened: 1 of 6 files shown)


bmw:
So I actually wasn't able to test this against the `libedit` implementation of readline. Can someone with an OS X machine test this?

If you've checked out this branch and activated the venv, you can test this by executing a file with these contents:
```
from letsencrypt.display import completer
with completer.Completer():
    raw_input("Partially type a path and press tab please: ")
```
Paths should be tab completed.

cc @erikrose

erikrose:
I get "readline is not using libedit :(".

erikrose:
Using the system python, it uses libedit and tab-completes successfully.

bmw:
Great! Thanks @erikrose.

schoen:
@bmw, are you anticipating making any changes as a result of @erikrose's experiment?

bmw:
@schoen: No. Everything worked as expected

letsencrypt/display/completer.py
```diff
@@ -0,0 +1,61 @@
...
+        :param int state: which completion to return
+
+        :returns: possible completion for text or ``None`` if all
+            completions have been returned
+        :rtype: str
+
+        """
+        if state == 0:
+            self._iter = glob.iglob(text + '*')
+        return next(self._iter, None)
```
schoen: Strictly speaking, this assumes that the completes get requested in order, right? If a called requested 7 and then 5, it would not actually go into the glob data and get item 7 and then item 5, but just return the next two in the currently-ongoing sequential order.

(I don't think this is a practical problem in terms of how readline will call this.)

letsencrypt/display/completer.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __enter__(self):
+        self._completer = readline.get_completer()
```
schoen: Should this maybe be called `_old_completer` or `_original_completer` or `_previous_completer` for clarity?

letsencrypt/display/completer.py
```diff
@@ -0,0 +1,61 @@
...
+        self._delims = readline.get_completer_delims()
```
schoen: Similarly, maybe `_previous_delims` or something?

letsencrypt/display/util.py
```diff
@@ -7,10 +7,18 @@
 
 from letsencrypt import interfaces
 from letsencrypt import errors
+from letsencrypt.display import completer
 
 WIDTH = 72
 HEIGHT = 20
 
+DSELECT_HELP = (
+    "Use the arrow keys or tab to move between window elements. Space can be "
```
schoen: I tend to write "Tab" instead of "tab", but maybe that's not a universal style?

letsencrypt/display/completer.py
bmw: Yes. I made the assumption the completions will be requested in order based on this [documentation](https://docs.python.org/2/library/readline.html#readline.set_completer). If you think it's a problem, I could just use `glob.glob` and index into the resulting `list`.

bmw:
Other than the suggestion I commented above, I agree all suggested changes should be made. Unless you'd like me to do otherwise, I'll wait until you've finished reviewing this to actually make the changes to this PR.

letsencrypt/display/util.py
```diff
@@ -174,6 +182,21 @@ def checklist(self, message, tags, default_status=True, **unused_kwargs):
...
+        """Display a directory selection screen.
+
+        :param str message: prompt to give the user
+
+        :returns: tuple of the form (`code`, `string`) where
+            `code` - int display exit code
+            `string` - input entered by the user
+
+        """
+        root_directory = os.path.abspath(os.sep)
```
schoen: Is this a standard idiom and is it portable to Windows (assuming it's meant to be)?

letsencrypt/display/util.py
```diff
@@ -520,6 +555,23 @@ def checklist(self, message, tags, default=None, cli_flag=None, **kwargs):
...
 
+    def directory_select(self, message, default=None, cli_flag=None):
+        """Simulate prompting the user for a directory.
+
+        This function returns default if it is not ``None``, otherwise,
+        an exception is raised.
+
+        :param str message: prompt to give the user
+        :param default: default value to return (if one exists)
+        :param str cli_flag: option used to set this value with the CLI
```
schoen: I didn't understand how `cli_flag` is used (although it already exists in the code elsewhere). Is it possible to give a slightly more detailed explanation in the docstring?

letsencrypt/display/completer.py
schoen: That documentation does seem to allow you to rely on that. :smile:

schoen:
LGTM apart from questions noted.

letsencrypt/display/util.py
pde: @schoen if the client is running in non-interactive mode, and a `MissingCommandLineFlag` error occurs because the client doesn't know the value of some config variable that it needs, the `cli_flag` is there to give the user a hint about what command line flag would set the missing variable.

letsencrypt/display/util.py
bmw: I got it from the most upvoted response here; https://stackoverflow.com/questions/12041525/a-system-independent-way-using-python-to-get-the-root-directory-drive-on-which-p

bmw:
Other than the two things I commented on, I addressed all review comments. Let me know if you'd like me to change anything else!

schoen:
Neat!

PR #2711, opened 2016-03-24
none should be None

### LINKED ISSUE #2709: authenticator/installer is set to "none"
This gets put into the renewal config file which will not be restored correctly. The correct value is "None".

bmw:
Fixes #2709


schoen:
That is indeed a very simple PR! :smile:

Is there any known issue with code trying to reinstantiate plugins on the basis of renewal configs that somehow does *better* with `none` than with `None`?

bmw:
> Is there any known issue with code trying to reinstantiate plugins on the basis of renewal configs that somehow does better with none than with None?

Not that I'm aware of.

schoen:
OK, hopefully we'll become aware of them if there turn out to be any!

schoen:
LGTM.

PR #2715, opened 2016-03-24
Fix LEA for Python 2.6

TheNavigat:
This PR fixes pipstrap for Python 2.6, and removes the redundant ``check_output`` import in ``auto_test.py`` which breaks the test when ran using Python 2.6.

As this link shows, https://docs.python.org/2/library/subprocess.html#subprocess.CalledProcessError, the ``output=output`` parameter was introduced with ``check_output`` itself, which isn't supported in Python 2.6.


bmw:
Did you report/fix this issue over at https://github.com/erikrose/pipstrap?

TheNavigat:
Ah, sorry I missed that.

There you go. https://github.com/erikrose/pipstrap/pull/6

bmw:
@TheNavigat, despite you being first, I took #2759 to get the version bump so we keep our copy of `pipstrap` up to date with Erik's repo.

TheNavigat:
I forgot that this PR contains fixes for both files ``auto_test.py`` and the ``pipstrap`` script, until I hit that wall again while working on another PR.

We still need to fix ``auto_test.py``.

PR #2717, opened 2016-03-25
Update Dockerfile so that it uses the pip package

### LINKED ISSUE #2712: Version Conflict in latest docker image
I just pulled the latest docker image today, and I get the following error when I try to run it:

```
# docker run -it --rm --name letsencrypt -p 10080:80 -p 10443:443 -v `pwd`/etc:/etc/letsencrypt -v `pwd`/var:/var/lib/letsencrypt quay.io/letsencrypt/letsencrypt:latest     
  File "/opt/letsencrypt/venv/bin/letsencrypt", line 5, in <module>
  File "/opt/letsencrypt/venv/local/lib/python2.7/site-packages/pkg_resources.py", line 2720, in <module>
    parse_requirements(__requires__), Environment()
  File "/opt/letsencrypt/venv/local/lib/python2.7/site-packages/pkg_resources.py", line 592, in resolve
    raise VersionConflict(dist,req) # XXX put more info here
pkg_resources.VersionConflict: (setuptools 2.2 (/opt/letsencrypt/venv/lib/python2.7/site-packages), Requirement.parse('setuptools>=11.3'))
```

TheNavigat:
Fixes #2712

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,59 +12,37 @@ EXPOSE 443
 # TODO: make sure --config-dir and --work-dir cannot be changed
 # through the CLI (certbot-docker wrapper that uses standalone
 # authenticator and text mode only?)
-VOLUME /etc/letsencrypt /var/lib/letsencrypt
 
-WORKDIR /opt/certbot
+WORKDIR ~/certbot
 
-# no need to mkdir anything:
+# No need to mkdir anything:
 # https://docs.docker.com/reference/builder/#copy
 # If <dest> doesn't exist, it is created along with all missing
 # directories in its path.
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-COPY letsencrypt-auto-source/letsencrypt-auto /opt/certbot/src/letsencrypt-auto-source/letsencrypt-auto
-RUN /opt/certbot/src/letsencrypt-auto-source/letsencrypt-auto --os-packages-only && \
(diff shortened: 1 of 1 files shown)


bmw:
Sorry for the false start here, but I don't think we want this. I just created #3782 to explain what I think we should do with our Dockerfile, but to provide more context, I think we have three options:

1. Install from OS packages
2. Install from PyPI packages
3. Install from git

`git` has the significant benefit that we can easily automate builds on quay.io and DockerHub. When we create a new release tag, we can configure those sites to automatically build a new docker image. My plan is to have Docker tag for each release and have `latest` be for the most recent release. We could potentially include `master` tag for people who want it.

This isn't as feasible when using PyPI packages because in addition to being unable to provide a `master` tag, due to how `certbot-auto` works, we must create the GitHub tag before uploading to PyPI which creates a race condition. Furthermore, I think having a Dockerfile in our tree that doesn't use the files in that tree is a bit awkward and has no real benefit over simply using the local files.

The benefit of OS packages is we don't need any dependencies for compilation. I proposed in #3762 to try removing these dependencies after installing Certbot. The downside of using OS packages in our Dockerfile is we can't automate builds as they depend on when the packages are uploaded in the distro, we can't provide a `master` tag, and we still have the awkward behavior of having a Dockerfile in the tree that doesn't use the build context.

I hope this makes sense. If you have any questions, comments, concerns, etc. please follow up here or in #3762. I'm closing this issue, but if my current plan falls through for some reason, we can revisit this PR.

PR #2722, opened 2016-03-27
Revise the concatenation order for fullchain.pem

chkpnt:
Let's encrypt does a [correct](http://blog.edgecloud.com/post/19519955133/ssl-certificate-chain-order-matters) concatenation of `certs.pem` and `chain.pem` for `fullchain.pem`. But in the docs, the order is interchanged.


schoen:
Thanks, this is a useful correction to our documentation.

PR #2724, opened 2016-03-28
Fix DNS challenge name: dns -> dns-01

jsha:
This fixes a debug log entry that says "dns-01 was not recognized."


moshevds:
Hi @jsha,

This doesn't seem correct to me. The dns and dns-01 challenges are actually different, and the code you are changing should, I think, best be removed. I am actually eagerly awaiting the DNS-01 updates, because I have an authenticator plugin that I want to propose when it arrives (implementing RFC 2136).

There is a lingering PR (#2061) that implements this acme challenge that is, I think, currently waiting to find a reviewer. I am not the author of that PR, but if there is anything I can do to get support for dns-01 into the codebase, please let me know.

jsha:
Good point, @moshevds, thanks! I'll close this PR.

Issue #2676, opened 2016-03-17
Renewer needs to reload webserver. 


plutocrat:
Hi, 
I just noticed my certs had renewed successfully, but the old certs were still showing on the website. This was not a big deal immediately, but after 29 days they would have expired, and that would cause problems. 
So, for a fully automated solution, the letsencrypt renew command should reload the webserver config. In my case this was a matter of 'service nginx reload' and the new certs appeared. This command should only be run on a successful renewal. i.e. when the cert has less than 30 days to run. 
Perhaps this could be addressed by a line in the relevant /etc/letsencrypt/renew/domain.conf file. Something like
run_on_successful_renewal = 'service nginx reload'

nitmir:
Having a `run_on_successful_renewal ` option would be nice. It could also allow for automatic TLSA record publishing.
```
run_on_successful_renewal = 'publish_tlsa; sleep 3600; service nginx reload'
```

FelixSchwarz:
I like the approach with a generic command being executed. This also helps for more complex setups where you probably just want to "register" the task ("restarting webserver") so these can be processed during the regular maintenance windows.

osirisinferi:
@nitmir: Shouldn't you also add a old-TLSA-record removal command after the `sleep`?

nitmir:
@osirisinferi  in fact my script (`make_tlsa`) scan certificates on 3 hosts, generate records pour every valid certificate (event old one in `/etc/letsencrypt/archives`) and SSFP records for every ssh pub key in `/etc/ssh/` in a file by dns zone. This files are included in bind dns zones files(`$INCLUDE /etc/bind/generated/db.tlsa.example.com example.com.`). So just calling `make_tlsa` will add and removes records, update the dns serial and reload the dns zones if needed.

Old certificates get remove then they are expired (`make_tlsa` is also called on a daily basis).

But it was just an example.

pde:
The current plan for this is via #2331, where I'm basically planning to make the "callback in case of renewal" be a function that's defined in a script in /etc.  It'll be like the [second of these two scripts](https://letsencrypt.org/getting-started/#writing-your-own-renewal-script), but will grep for [this message](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/cli.py#L714) in the client's output, and call a renewal function in that case. 

[...]

Issue #1303, opened 2015-11-03
Renewer needs cleanup


bmw:
The `renewer` is over-engineered in places and things such as what can be configured are not well defined.

pde:
#2304 is going to partially address this for 0.4.0.  Future work on `renewer.py` is an open question.

pde:
I think this is largely done!

PR #2729, opened 2016-03-30
Implement renew hooks

pde:
Implement user-specified shell hooks that can be run:

1. before obtaining any certs
2. after obtainaing all certs
3. after a specific lineage is renewed

Also does some renewal refactoring along the way.

Closes: #1504 #2676 
Finishes: #1303

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -558,7 +558,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         None, "--dry-run", action="store_true", dest="dry_run",
         help="Perform a test run of the client, obtaining test (invalid) certs"
              " but not saving them to disk. This can currently only be used"
-             " with the 'certonly' subcommand.")
+             " with the 'certonly' and 'renew' subcommands. \nNote: Although --dry-run"
+             " tries to avoid making any persistent changes on a system, it "
+             " is not completely side-effect free: if used with webserver authenticator plugins"
+             " like apache and nginx, it makes and then reverts temporary config changes"
+             " in order to obtain test certs, and reloads webservers to deploy and then"
+             " roll back those changes.  It also calls --pre-hook and --post-hook commands"
+             " if they are defined because they may be necessary to accurately simulate"
+             " renewal. --renew-hook commands are not called.")
     helpful.add(
         None, "--register-unsafely-without-email", action="store_true",
         help="Specifying this flag enables registering an account with no "
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -692,7 +692,23 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         " used to create obtain or most recently successfully renew each"
         " certificate lineage. You can try it with `--dry-run` first. For"
         " more fine-grained control, you can renew individual lineages with"
-        " the `certonly` subcommand.")
+        " the `certonly` subcommand. Hooks are available to run commands "
+        " before and after renewal; see XXX for more information on these.")
+
+    helpful.add(
+        "renew", "--pre-hook",
+        help="Command to be run in a shell before obtaining any certificates. Intended"
+        " primarily for renewal, where it can be used to temporarily shut down a"
+        " webserver that might conflict with the standalone plugin. This will "
+        " only be called if a certificate is actually to be obtained/renewed. ")
+    helpful.add(
+        "renew", "--post-hook",
+        help="Command to be run in a shell after attempting to obtain/renew "
+        " certificates. Can be used to deploy renewed certificates, or to restart"
+        " any servers that were stopped by --pre-hook.")
+    helpful.add(
+        "renew", "--renew-hook",
+        help="Command to be run in a shell once for each renewed certificate")
 
     helpful.add_deprecated_argument("--agree-dev-preview", 0)
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -711,7 +711,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         " any servers that were stopped by --pre-hook.")
     helpful.add(
         "renew", "--renew-hook",
-        help="Command to be run in a shell once for each renewed certificate."
+        help="Command to be run in a shell once for each successfully renewed certificate."
         "For this command, the shell variable $RENEWED_LINEAGE will point to the"
         "config live subdirectory containing the new certs and keys; the shell variable "
         "$RENEWED_DOMAINS will conatain a space-delimited list of renewed cert domains")
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -708,7 +708,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         " any servers that were stopped by --pre-hook.")
     helpful.add(
         "renew", "--renew-hook",
-        help="Command to be run in a shell once for each renewed certificate")
+        help="Command to be run in a shell once for each renewed certificate."
+        "For this command, the shell variable $RENEWED_LINEAGE will point to the"
+        "config live subdirectory containing the new certs and keys; the shell variable "
+        "$RENEWED_DOMAINS will conatain a space-delimited list of renewed cert domains")
 
     helpful.add_deprecated_argument("--agree-dev-preview", 0)
 
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -18,6 +18,7 @@ import letsencrypt
 from letsencrypt import constants
 from letsencrypt import crypto_util
 from letsencrypt import errors
+from letsencrypt import hooks
 from letsencrypt import interfaces
 from letsencrypt import le_util
 
@@ -305,6 +306,8 @@ class HelpfulArgumentParser(object):
 
         if self.detect_defaults:  # plumbing
             parsed_args.store_false_vars = self.store_false_vars
+
+        hooks.validate_hooks(parsed_args)
 
         return parsed_args
 
--- a/letsencrypt/hooks.py
+++ b/letsencrypt/hooks.py
@@ -58,9 +58,12 @@ def post_hook(config, final=False):
 def renew_hook(config, domains, lineage_path):
     "Run post-renewal hook if defined."
     if config.renew_hook:
-        os.environ["RENEWED_DOMAINS"] = " ".join(domains)
-        os.environ["RENEWED_LINEAGE"] = lineage_path
-        _run_hook(config.renew_hook)
+        if not config.dry_run:
+            os.environ["RENEWED_DOMAINS"] = " ".join(domains)
+            os.environ["RENEWED_LINEAGE"] = lineage_path
+            _run_hook(config.renew_hook)
+        else:
+            print("Dry run: skipping renewal hook command: {0}".format(config.renew_hook))
 
 def _run_hook(shell_cmd):
     """Run a hook command.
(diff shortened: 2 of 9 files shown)


pde:
This has been manually tested, and should be ready for first review (though it will still need unit tests).

letsencrypt/cli.py
```diff
@@ -692,7 +695,26 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        " webserver that might conflict with the standalone plugin. This will "
+        " only be called if a certificate is actually to be obtained/renewed. ")
+    helpful.add(
+        "renew", "--post-hook",
+        help="Command to be run in a shell after attempting to obtain/renew "
+        " certificates. Can be used to deploy renewed certificates, or to restart"
+        " any servers that were stopped by --pre-hook.")
+    helpful.add(
+        "renew", "--renew-hook",
+        help="Command to be run in a shell once for each renewed certificate."
```
schoen: It might be clearer to say "for each successfully-renewed certificate" (although I think someone looking at this closely will eventually understand that that's already implied).

letsencrypt/renewal.py
```diff
@@ -199,6 +205,50 @@ def should_renew(config, lineage):
...
+    def _is_staging(srv):
+        return srv == constants.STAGING_URI or "staging" in srv
+
+    # Some lineages may have begun with --staging, but then had production certs
+    # added to them
+    latest_cert = OpenSSL.crypto.load_certificate(
+        OpenSSL.crypto.FILETYPE_PEM, open(lineage.cert).read())
+    # all our test certs are from happy hacker fake CA, though maybe one day
+    # we should test more methodically
+    now_valid = "fake" not in repr(latest_cert.get_issuer()).lower()
```
schoen: I thought this was a problem due to https://community.letsencrypt.org/t/upcoming-intermediate-changes/13106 but `.lower()` should mean this is still OK.

letsencrypt/main.py
```diff
@@ -142,7 +108,8 @@ def _handle_subset_cert_request(config, domains, cert):
 
     :param storage.RenewableCert cert:
 
-    :returns: Tuple of (string, cert_or_None) as per _treat_as_renewal
+    :returns: Tuple of (stringa action, cert_or_None) as per _treat_as_renewal
```
schoen: I think the `stringa` is a typo.

letsencrypt/main.py
```diff
@@ -183,7 +150,8 @@ def _handle_identical_cert_request(config, cert):
 
     :param storage.RenewableCert cert:
 
-    :returns: Tuple of (string, cert_or_None) as per _treat_as_renewal
+    :returns: Tuple of (string action, cert_or_None) as per _treat_as_renewal
```
schoen: maybe `str` instead of `string` anyway

letsencrypt/main.py
```diff
@@ -507,41 +475,53 @@ def run(config, plugins):  # pylint: disable=too-many-branches,too-many-locals
...
-        certr, chain = le_client.obtain_certificate_from_csr(config.domains, csr, typ)
-        if config.dry_run:
-            logger.info(
-                "Dry run: skipping saving certificate to %s", config.cert_path)
-        else:
-            cert_path, _, cert_fullchain = le_client.save_certificate(
-                certr, chain, config.cert_path, config.chain_path, config.fullchain_path)
-            _report_new_cert(cert_path, cert_fullchain)
-    else:
+    # SHOWTIME: Possibly obtain/renew a cert, and set action to renew | newcert | reinstall
```
schoen: Did you ever play the Addams Family Pinball?

(for reference: https://www.youtube.com/watch?v=pkoP3jLjH3U)

letsencrypt/renewal.py
```diff
@@ -199,6 +205,50 @@ def should_renew(config, lineage):
...
+    else:
+        prior_version = lineage.latest_common_version()
+        new_cert = OpenSSL.crypto.dump_certificate(
+            OpenSSL.crypto.FILETYPE_PEM, new_certr.body.wrapped)
+        new_chain = crypto_util.dump_pyopenssl_chain(new_chain)
+        renewal_conf = configuration.RenewerConfiguration(config.namespace)
+        lineage.save_successor(prior_version, new_cert, new_key.pem, new_chain, renewal_conf)
+        lineage.update_all_links_to(lineage.latest_common_version())
+
+    hooks.renew_hook(config, domains, lineage.live_dir)
```
schoen: What is the most intuitively correct behavior for the interaction of `dry_run` with `renew_hook`? The PR takes the position that the `renew_hook` should be run, which makes sense but I'm not positive it's right, because it means the dry run will have potentially significant side effects.

letsencrypt/storage.py
```diff
@@ -252,6 +252,7 @@ def __init__(self, config_filename, cli_config):
         self.privkey = self.configuration["privkey"]
         self.chain = self.configuration["chain"]
         self.fullchain = self.configuration["fullchain"]
+        self.live_dir = os.path.dirname(self.cert)
```
schoen: We'll have a lot of thinking to do about what this means if we end up reorganizing the cert storage so that the four PEM objects are not necessarily in the same directory (as I believe djb is going to advocate for security/privilege separation reasons). I agree that this construction of `live_dir` is totally correct in our status quo.

letsencrypt/cli.py
```diff
@@ -692,7 +695,26 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        "For this command, the shell variable $RENEWED_LINEAGE will point to the"
+        "config live subdirectory containing the new certs and keys; the shell variable "
```
schoen: Is there a possibility that someone's script would want references to any other storage objects? I agree that this is the most useful one overall; I'm just wondering if we should also supply any others (like a reference to the renewal config file, maybe?).

schoen: As a further concern, I mentioned elsewhere the possibility that after the storage reorganization there will no longer be a single live dir containing all four PEM objects. If that happens, people's existing hook scripts could break, especially if they try to reconstruct the four paths by just concatenating `something.pem` to the `RENEWED_LINEAGE` variable. Although this is super-annoying, I think it might be good future-proofing to instead provide four separate variables (e.g. `RENEWED_LINEAGE_CERT`, `RENEWED_LINEAGE_CHAIN`, `RENEWED_LINEAGE_FULLCHAIN`, and `RENEWED_LINEAGE_PRIVKEY`), because this can still be correct and the variables can still be used in the same way following the reorganization. This might encourage/force users to write safer hook scripts, in the sense of more reorganization-tolerant ones.

letsencrypt/cli.py
```diff
@@ -692,7 +695,26 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        " only be called if a certificate is actually to be obtained/renewed. ")
+    helpful.add(
+        "renew", "--post-hook",
+        help="Command to be run in a shell after attempting to obtain/renew "
+        " certificates. Can be used to deploy renewed certificates, or to restart"
+        " any servers that were stopped by --pre-hook.")
+    helpful.add(
+        "renew", "--renew-hook",
+        help="Command to be run in a shell once for each renewed certificate."
+        "For this command, the shell variable $RENEWED_LINEAGE will point to the"
```
schoen: I don't think the dollar sign is the most correct documentation style in the name of a shell variable (for instance, I think it's more correct to write "the `PATH` variable" rather than "the `$PATH` variable", although both forms are common). `bash(1)` has many examples of this.

letsencrypt/cli.py
```diff
@@ -692,7 +695,26 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
+        "config live subdirectory containing the new certs and keys; the shell variable "
+        "$RENEWED_DOMAINS will conatain a space-delimited list of renewed cert domains")
```
schoen: Similarly, I'd say `RENEWED_DOMAINS` instead of `$RENEWED_DOMAINS`.

letsencrypt/hooks.py
```diff
@@ -0,0 +1,99 @@
...
+def post_hook(config, final=False):
+    """Run post hook if defined.
+
+    If the verb is renew, we might have more certs to renew, so we wait until
+    we're called with final=True before actually doing anything.
+    """
+    if config.post_hook:
+        if final or config.verb != "renew":
+            logger.info("Running post-hook command: %s", config.post_hook)
+            _run_hook(config.post_hook)
```
schoen: If I were writing a post-hook script, I can imagine that I would want more programmatic access to information about what happened, for example a `RENEW_SUCCESS` variable or something that indicates whether renewals succeeded. It's a little hard for me to make super-specific suggestions about what information to provide because I haven't had any actual need to write these scripts, but maybe we should ask on the forum?

letsencrypt/hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    def _is_exe(fpath):
+        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
+
+    fpath, _fname = os.path.split(program)
+    if fpath:
+        if _is_exe(program):
+            return program
+    else:
+        for path in os.environ["PATH"].split(os.pathsep):
+            path = path.strip('"')
```
schoen: Why would the PATH contain quote marks? If it did, why would it be correct to remove them?

schoen:
LGTM apart from comments noted. I didn't look very closely at the reorganizational aspects as opposed to the renew hook aspects. The PR contains a bit of reorganization alongside the new renew hook functionality.

letsencrypt/main.py
pde: I might have.  But I have no recollection of that phenomenon.  Probably its cultural roots go even deeper?

In fact, [both things seem to be true](https://books.google.com/ngrams/graph?content=showtime%2CAddams+Family&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cshowtime%3B%2Cc0%3B.t1%3B%2CAddams%20Family%3B%2Cc0).

letsencrypt/hooks.py
pde: Hmmm. I agree that doesn't make sense.

letsencrypt/hooks.py
pde: I believe this should only be run if the renewal succeeded.  I believe an exception should be raised before the call in `renewal.renew_cert()` if something goes wrong.

letsencrypt/storage.py
pde: I don't think you should change the structure of the `live_dir` layout; my sense is that the things djb is asking for would all involve reorganising `archive/` but leaving the links in `live/` as they currently are (though maybe more kinds of file should be added there...)

letsencrypt/renewal.py
pde: Actually the PR *doesn't* run the renew_hook because of [this test](https://github.com/letsencrypt/letsencrypt/pull/2729/files#diff-b94a39e90e97569ae241441be8dfd482R60).  But it does run the `pre-hook` and `post-hook` commands, because the pre-hook may be necessary for being able to accurately test whether renewal will succeed and the post-hook may be necessary for cleanup.

I don't really like this inconsistency in behaviour.  One option is just to document it under `--dry-run`.

pde: Ok, I've at least documented this behaviour.

schoen:
I think all my comments have been addressed, although I haven't looked at any of the new code.

I'll try posting on the forum about whether people would like programmatic access to any other information for their scripts. (We don't necessarily have to wait for their answers before merging this because people who do want more information in environment variables can update to a future release that provides that information.)

letsencrypt/hooks.py
schoen: I did ask about this on the forum and I'll try to file an issue if anyone proposes anything.

schoen:
@pde, I think this is fine now but don't know whether you were planning to do anything else on it before merge.

pde:
OK, merging!

PR #2733, opened 2016-03-31
Preserve comments in renewal conf files

### LINKED ISSUE #2368: Preserve comments when updating renewal configs
The approach proposed by #2361 doesn't preserve comments in renewal config files. When these files are documented, comments MUST (where MUST is defined in RFC 2119) be preserved.

schoen:
Changing the method of updating and rewriting renewal config files to use the same ConfigObj instance rather than a new one, and change individual renewalparams items individually rather than replacing the entire renewalparams dict with a new dict (which may also cause a loss of associated comment data).

Fixes #2368.

### DIFF
--- a/letsencrypt/storage.py
+++ b/letsencrypt/storage.py
@@ -50,10 +50,11 @@ def add_time_interval(base_time, interval, textparser=parsedatetime.Calendar()):
     return textparser.parseDT(interval, base_time, tzinfo=tzinfo)[0]
 
 
-def write_renewal_config(filename, target, relevant_data):
+def write_renewal_config(o_filename, n_filename, target, relevant_data):
     """Writes a renewal config file with the specified name and values.
 
-    :param str filename: Absolute path to the config file
+    :param str o_filename: Absolute path to the previous version of config file
+    :param str n_filename: Absolute path to the new destination of config file
     :param dict target: Maps ALL_FOUR to their symlink paths
     :param dict relevant_data: Renewal configuration options to save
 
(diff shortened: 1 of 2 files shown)


schoen:
Still needs updated tests, and I have to confirm that it does the right thing in both the new-lineage and the renewal cases.

schoen:
Basic functionality appears to work correctly (and fix #2368) for both initial creation and renewal cases. The PR could still use test coverage, though.

letsencrypt/storage.py
```diff
@@ -50,10 +50,11 @@ def add_time_interval(base_time, interval, textparser=parsedatetime.Calendar()):
     return textparser.parseDT(interval, base_time, tzinfo=tzinfo)[0]
 
 
-def write_renewal_config(filename, target, relevant_data):
+def write_renewal_config(o_filename, n_filename, target, relevant_data):
```
schoen: (The reason for specifying `o_filename` and `n_filename` is to avoid disturbing the use of `os.rename` in `update_configuration`, given that we are now creating a renewal config based on the old one rather than simply blindly writing the values from scratch.)

letsencrypt/storage.py
```diff
@@ -61,21 +62,27 @@ def write_renewal_config(filename, target, relevant_data):
...
+        config["renewalparams"] = {}
         config.comments["renewalparams"] = ["",
                                             "Options used in "
                                             "the renewal process"]
 
+    config["renewalparams"].update(relevant_data)
+
+    for k in config["renewalparams"].keys():
+        if k not in relevant_data:
+            del config["renewalparams"][k]
```
bmw: One thing I noticed is that by doing this after the line `config["renewalparams"].update(relevant_data)`, comments are preserved even if the value is modified. For example, if I have a renewal config file with this line:
```
authenticator = apache # my authenticator is apache
```
After running:
```
letsencrypt renew --force-renewal --webroot -w /var/www/html
```
The line becomes:
```
authenticator = webroot# my authenticator is apache
```
Is this the behavior we want? It might be, but I thought I'd ask the question.

bmw:
I have a question about preserving comments on lines with modified values, but it's a pretty minor point and this PR doesn't need to block on it.

LGTM. Nice work @schoen.

PR #2737, opened 2016-03-31
Add MariaDB debugging to Travis build

### LINKED ISSUE #2681: Add MariaDB debugging to Travis build
In Boulder, we just landed this change to our Travis config: https://github.com/letsencrypt/boulder/pull/1625. It sends mysql.err to stdout in the event of failure. Hopefully this will give us more insight into the spurious i/o timeout failures.

Unfortunately to be able to do this, I had to set sudo:true (temporarily), which puts us on the slower, non-container infrastructure. I don't yet have an idea of how much slower.

Note that since the client gets more pull requests than Boulder, it tends to trigger the bug more often. You may want to consider making the same change in the client .travis.yml.

TheNavigat:
Closes #2681


bmw:
LGTM. Thanks @TheNavigat!

PR #2741, opened 2016-04-01
Fix plugin default detection

### LINKED ISSUE #2651: _set_by_cli spuriously flags all plugin variables to be set by cli

bmw:
Fixes #2651.

### DIFF
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -86,6 +86,17 @@ More detailed help:
 """
 
 
+# These sets are used when to help detect options set by the user.
+ARGPARSE_PARAMS_TO_REMOVE = set(("const", "nargs", "type",))
+
+
+EXIT_ACTIONS = set(("help", "version",))
+
+
+ZERO_ARG_ACTIONS = set(("store_const", "store_true",
+                        "store_false", "append_const", "count",))
+
+
 def usage_strings(plugins):
     """Make usage strings late so that plugins can be initialised late"""
     if "nginx" in plugins:
@@ -100,7 +111,19 @@ def usage_strings(plugins):
 
 
 class _Default(object):
-    """Trivial class used to detect if an option was set by the user."""
+    """A class to use as a default to detect if a value is set by a user"""
+
+    def __bool__(self):
+        return False
+
+    def __eq__(self, other):
+        return isinstance(other, _Default)
+
+    def __hash__(self):
+        return id(_Default)
+
+    def __nonzero__(self):
+        return self.__bool__()
 
 
 def set_by_cli(var):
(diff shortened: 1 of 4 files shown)


bmw:
@jsha, `./bin/boulder-sa` exited early: [log1](https://gist.github.com/86c2096c408c715ccaee86d18af1f604), [log2](https://gist.github.com/238fe522ace44b629b18491eaa856655).

letsencrypt/cli.py
```diff
@@ -87,6 +87,48 @@
...
+# These argparse parameters should be removed when detecting defaults.
+ARGPARSE_PARAMS_TO_REMOVE = ("const", "nargs", "type",)
+
+
+# These sets are used when to help detect options set by the user.
+EXIT_ACTIONS = set(("help", "version",))
+
+
+ZERO_ARG_ACTIONS = set(("store_const", "store_true",
+                        "store_false", "append_const", "count",))
```
pde: Stylistically, we should probably move all of this default detection machinery to a different file (though that will require inheriting from `HelpfulArgumentParser` rather than tweaking it) or at least a clearly marked section of this file, because it's probably confusing that these variables are scattered amongst the regular CLI parsing case.

But that can be another PR rather than this one.

letsencrypt/renewal.py
```diff
@@ -102,16 +102,14 @@ def _restore_webroot_config(config, renewalparams):
     form.
     """
     if "webroot_map" in renewalparams:
-        # if the user does anything that would create a new webroot map on the
-        # CLI, don't use the old one
-        if not (cli.set_by_cli("webroot_map") or cli.set_by_cli("webroot_path")):
-            setattr(config.namespace, "webroot_map", renewalparams["webroot_map"])
+        if not cli.set_by_cli("webroot_map"):
```
pde: I think the previous semantics might be less confusing to people here.  Suppose on day 1 you did:
```
letsencrypt certonly --webroot -w /var/www/ -d example.com,www.example.com
```
And then you did:
```
letsencrypt renew -w /var/www/moved
```
I believe the change you've made here would cause the webroot to remain `/var/www/` during renewal, because the client would construct and save a `webroot_map` during first execution.

pde: Oh wait, I'm sorry, I see you've handled this case via `VAR_MODIFIERS`.

pde:
LGTM!

PR #2743, opened 2016-04-01
IDisplay calls should also produce log entries [needs tests]

pde:
all iDisplay implementations now have a `.log()` method which wraps them in a `DisplayLogger` shim.


bmw:
This hasn't seen any progress in quite a while.

Closing the PR to help with my OCD, but feel free to open again when it's ready for review.

Issue #2512, opened 2016-02-20
Implement --quiet / -q


bmw:
As suggested in #2508, we should add `--quiet` so the client can more easily be used with cron.

bmw:
Perhaps this should provide output if an error occurs. On the other hand, all relevant output will be in the log files (with the exception of the renewal summary which was recently added).

osirisinferi:
But a "no output if all is OK, do give output when something is wrong" setting would be very helpfull in a `cron` situation.

nitmir:
Or put OK output to stdout and error output to stderr (I do not know, it can already be the case).
So in a `cron` one can do `> /dev/null` and still receive errors.

osirisinferi:
That's even better IMHO.

pde:
This will be a necessary item if we do #2676 instead of #2331

PR #2745, opened 2016-04-02
Implement --quiet / -q

pde:
Which silences all non-error output.

### DIFF
--- a/letsencrypt/main.py
+++ b/letsencrypt/main.py
@@ -687,7 +687,7 @@ def main(cli_args=sys.argv[1:]):
     zope.component.provideUtility(displayer)
 
     # Reporter
-    report = reporter.Reporter()
+    report = reporter.Reporter(config)
     zope.component.provideUtility(report)
     atexit.register(report.atexit_print_messages)
 
--- a/letsencrypt/main.py
+++ b/letsencrypt/main.py
@@ -34,7 +34,6 @@ from letsencrypt.display import util as display_util, ops as display_ops
 from letsencrypt.plugins import disco as plugins_disco
 from letsencrypt.plugins import selection as plug_sel
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -537,20 +536,21 @@ def obtain_cert(config, plugins, lineage=None):
         domains = _find_domains(config, installer)
         _, action = _auth_from_domains(le_client, config, domains, lineage)
 
+    notify = zope.component.getUtility(interfaces.IDisplay).notification
     if config.dry_run:
         _report_successful_dry_run(config)
     elif config.verb == "renew":
         if installer is None:
             # Tell the user that the server was not restarted.
-            print("new certificate deployed without reload, fullchain is",
-                  lineage.fullchain)
+            notify("new certificate deployed without reload, fullchain is {0}".format(
+                   lineage.fullchain), pause=False)
         else:
             # In case of a renewal, reload server to pick up new certificate.
             # In principle we could have a configuration option to inhibit this
             # from happening.
             installer.restart()
-            print("new certificate deployed with reload of",
-                  config.installer, "server; fullchain is", lineage.fullchain)
+            notify("new certificate deployed with reload of {0} server; fullchain is {1}".format(
+                   config.installer, lineage.fullchain), pause=False)
     _suggest_donation_if_appropriate(config, action)
 
 def renew(config, unused_plugins):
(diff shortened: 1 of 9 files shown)


pde:
Closes #2512.

pde:
@bmw this should be ready for review.

bmw:
@pde, travis is red.

pde:
@bmw it should have been green, I think I had just confused Travis.

letsencrypt/main.py
```diff
@@ -672,7 +672,10 @@ def main(cli_args=sys.argv[1:]):
     sys.excepthook = functools.partial(_handle_exception, config=config)
 
     # Displayer
-    if config.noninteractive_mode:
+    if config.quiet:
+        config.noninteractive_mode = True
+        displayer = display_util.NoninteractiveDisplay(open("/dev/null", "w"))
```
bmw: nit: We should probably use `os.devnull` here.

letsencrypt/reporter.py
```diff
@@ -76,9 +77,10 @@ def print_messages(self):
         if not self.messages.empty():
             no_exception = sys.exc_info()[0] is None
             bold_on = sys.stdout.isatty()
-            if bold_on:
-                print(le_util.ANSI_SGR_BOLD)
-            print('IMPORTANT NOTES:')
+            if not self.config.quiet:
```
bmw: This logic isn't quite right.

Before, `bold_on` was set and `le_util.ANSI_SGR_BOLD` was written to the terminal depending on the value of `bold_on`. Now, `bold_on` can be set to `True` but we never actually write `le_util.ANSI_SGR_BOLD` to the terminal, causing us to unnecessarily write `le_util.ANSI_SGR_RESET` later.

This isn't a big deal, but it should probably be written correctly.

bmw:
My biggest comment on this PR is I think using `notification` for `_renew_describe_results` looks pretty bad. Here's example output:
```
# letsencrypt renew --force-renewal

Processing /etc/letsencrypt/renewal/example.com.conf

new certificate deployed with reload of apache server; fullchain is
/etc/letsencrypt/live/example.com/fullchain.pem

Congratulations, all renewals succeeded. The following certs have been renewed:
  /etc/letsencrypt/live/example.com/fullchain.pem (success)
#
```
If you're happy with this, I suppose we can ignore this problem and move forward with this PR.

pde:
OK, I've tweaked things further so that the `.notification()` call only happens, typical output now looks like:
```
lel renew  --dry-run 

Processing /etc/letsencrypt/renewal/isnot.org.conf
2016-04-02 19:29:48,156:WARNING:letsencrypt.renewal:Attempting to renew cert from /etc/letsencrypt/renewal/isnot.org.conf produced an unexpected error: 'server'. Skipping.

Processing /etc/letsencrypt/renewal/goxogle.com.conf

Processing /etc/letsencrypt/renewal/demo.reworld.org.conf
2016-04-02 19:29:57,768:WARNING:letsencrypt.renewal:Attempting to renew cert from /etc/letsencrypt/renewal/demo.reworld.org.conf produced an unexpected error: 'server'. Skipping.

Processing /etc/letsencrypt/renewal/isnot.org-0001.conf
** DRY RUN: simulating 'letsencrypt renew' close to cert expiry

The following certs were successfully renewed:
  /etc/letsencrypt/live/goxogle.com/fullchain.pem (success)
  /etc/letsencrypt/live/isnot.org-0001/fullchain.pem (success)

The following certs could not be renewed:
  /etc/letsencrypt/live/isnot.org/fullchain.pem (failure)
  /etc/letsencrypt/live/demo.reworld.org/fullchain.pem (failure)
** DRY RUN: simulating 'letsencrypt renew' close to cert expiry
2 renew failure(s), 0 parse failure(s)
```

I think having the "heading" around the start of each renewed lineage's logs is probably nicer than what we have in `master` at the moment.

bmw:
@jsha: [Error creating new cert](https://api.travis-ci.org/jobs/120326815/log.txt)

letsencrypt/reporter.py
bmw: This doesn't seem to have been addressed.

letsencrypt/reporter.py
```diff
@@ -86,6 +88,11 @@ def print_messages(self):
...
         while not self.messages.empty():
             msg = self.messages.get()
+            if self.config.quiet:
+                # In --quiet mode, we only print high priority messages that
+                # are flagged for crash cases
+                if not (msg.priority == self.HIGH_PRIORITY and msg.on_crash):
+                    continue
             if no_exception or msg.on_crash:
                 if bold_on and msg.priority > self.HIGH_PRIORITY:
                     sys.stdout.write(le_util.ANSI_SGR_RESET)
```
bmw: That's still not right due to this line.

pde: Thanks.  I was reasoning incorrectly about that state machine!

bmw:
Haha. I think a better way to have approached the problem is only set `bold_on` to `True` if we print `le_util.ANSI_SGR_BOLD`, but this works.

That is assuming you don't want to use bold text for high priority messages when using `--quiet`. I don't really see a reason for not using bold text if it's available (stdout is a tty), but I don't have a problem if this is the behavior you want.

If this is the desired behavior, this LGTM if it looks good to Travis.

pde:
I don't think it matters very much; people running with `--quiet` are presumably not looking for nicely formatted output :)

PR #2761, opened 2016-04-04
Fix --dry-run interactions

### LINKED ISSUE #2748: renew: letsencrypt.client:Registering without email!
I succesfully set up a certificate for https://turbolab.it/ (CentOS 6.7 with Apache 2.2) like this:

````
git clone https://github.com/letsencrypt/letsencrypt /opt/letsencrypt
cd /opt/letsencrypt
./letsencrypt-auto certonly --webroot -w /full/site/path/ -d turbolab.it -d www.turbolab.it
````

I provided my email + agreed to TOS from the *graphic* (blue) UI.

**Now I'm trying to stage the cert renewal but I'm facing issue**. When I try to renew with `./letsencrypt-auto renew --dry-run` I get:

```
[root@skycracker letsencrypt]# ./letsencrypt-auto renew --dry-run
Checking for new version...
Installing Python packages...
Installation succeeded.
Requesting root privileges to run letsencrypt...
   /root/.local/share/letsencrypt/bin/letsencrypt renew --dry-run
/root/.local/share/letsencrypt/lib/python2.6/site-packages/cryptography/__init__.py:26: DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of cryptography will drop support for Python 2.6
  DeprecationWarning
Processing /etc/letsencrypt/renewal/turbolab.it.conf
2016-04-02 13:49:45,115:WARNING:letsencrypt.client:Registering without email!
2016-04-02 13:49:45,910:WARNING:letsencrypt.cli:Attempting to renew cert from /etc/letsencrypt/renewal/turbolab.it.conf produced an unexpected error: Missing command line flag or config entry for this setting:
Please read the Terms of Service at https://letsencrypt.org/documents/LE-SA-v1.0.1-July-27-2015.pdf. You must agree in order to register with the ACME server at https://acme-staging.api.letsencrypt.org/directory

(You can set this with the --agree-tos flag). Skipping.
** DRY RUN: simulating 'letsencrypt renew' close to cert expiry

All renewal attempts failed. The following certs could not be renewed:
  /etc/letsencrypt/live/turbolab.it/fullchain.pem (failure)
** DRY RUN: simulating 'letsencrypt renew' close to cert expiry
1 renew failure(s), 0 parse failure(s)
```

The `DeprecationWarning` makes me very uncomfortable, but I understood shouldn't be an issue.

The failure is the real issue here. I understood that the `renew` parameter should fire a renew with the same options used at install-time, so why the failure?

Thanks!

Update: I can confirm that `/etc/letsencrypt/renewal/turbolab.it.conf` holds the correct setting `
email = info@ MY SITE`

### PR DESCRIPTION
**Pull request by bmw**
Fixes #2748.

### DIFF
--- a/letsencrypt/client.py
+++ b/letsencrypt/client.py
@@ -105,7 +105,8 @@ def register(config, account_storage, tos_cb=None):
                    "--register-unsafely-without-email was not present.")
             logger.warn(msg)
             raise errors.Error(msg)
-        logger.warn("Registering without email!")
+        if not config.dry_run:
+            logger.warn("Registering without email!")
 
     # Each new registration shall use a fresh new key
     key = jose.JWKRSA(key=jose.ComparableRSAKey(
--- a/letsencrypt/cli.py
+++ b/letsencrypt/cli.py
@@ -339,7 +339,7 @@ class HelpfulArgumentParser(object):
                 if glob.glob(os.path.join(parsed_args.config_dir, constants.ACCOUNTS_DIR, "*")):
                     # The user has a prod account, but might not have a staging
                     # one; we don't want to start trying to perform interactive registration
-                    parsed_args.agree_tos = True
+                    parsed_args.tos = True
                     parsed_args.register_unsafely_without_email = True
 
         if parsed_args.csr:
--- a/letsencrypt/tests/cli_test.py
+++ b/letsencrypt/tests/cli_test.py
@@ -427,20 +427,40 @@ class CLITest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         short_args += '--server example.com'.split()
         self._check_server_conflict_message(short_args, '--staging')
 
-    def _assert_dry_run_flag_worked(self, namespace):
+    def _assert_dry_run_flag_worked(self, namespace, existing_account):
         self.assertTrue(namespace.dry_run)
         self.assertTrue(namespace.break_my_certs)
         self.assertTrue(namespace.staging)
         self.assertEqual(namespace.server, constants.STAGING_URI)
 
+        if existing_account:
+            self.assertTrue(namespace.tos)
+            self.assertTrue(namespace.register_unsafely_without_email)
+        else:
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by bmw** [comment]
@jsha: holy lookup errors batman - [log1](https://gist.github.com/102d8883cd784d38dcaa967694d108d2), [log2](https://gist.github.com/991f50875145ad073d44c2e58550c8d8), [log3](https://gist.github.com/a18d3fcf46cebc2e40196857280d7d55)

**Comment by jsha** [comment]
Sorry, this is my bad - I merged a Boulder change that broke your build. PR to unbreak build in 3, 2, 1...

File: letsencrypt/client.py [hunk 1]
```diff
@@ -105,7 +105,8 @@ def register(config, account_storage, tos_cb=None):
                    "--register-unsafely-without-email was not present.")
             logger.warn(msg)
             raise errors.Error(msg)
-        logger.warn("Registering without email!")
+        if not config.dry_run:
```
pde: Channeling Seth a little here, there are subtleties to this:
* Should we also suppress this message in staging mode?  Or do we expect people to know about their staging accounts?
* If not, should we note when what we're registering for is a staging account?

pde:
Overall, LGTM, merge when green.

bmw:
Ah great! Thanks for the quick PR jsha.

letsencrypt/client.py
bmw: > Should we also suppress this message in staging mode? Or do we expect people to know about their staging accounts?

I don't have a strong opinion on this either way. If we want to suppress this message when using the staging server, we can simply change this line to `if not config.dry_run` (and fix the test that will break).

> If not, should we note when what we're registering for is a staging account?

I don't think so. Treating servers differently in places other than `cli.py` seems clunky and unnecessary. The only way we're going to ask a user to register for a staging account is if they specifically requested it with the CLI/config file. Given this, adding a special prompt here doesn't seem like a big gain.

PR #2762, opened 2016-04-04
Minor cleanup

bmw:
This is a trivial PR, but I noticed the deleted code is redundant because of [these lines](https://github.com/letsencrypt/letsencrypt/blob/23dfb5b99ea206a38f96cd099b865df4cebab819/letsencrypt/cli.py#L313) and I couldn't help myself.


pde:
LGTM

PR #2763, opened 2016-04-04
Add boulder-mysql and boulder-rabbitmq hosts.

jsha:
Fixes build


bmw:
LGTM. Merging when green.

PR #2767, opened 2016-04-05
NcursesDisplay.menu: treat ESC as cancel

pde:
Currently it will fire a weird traceback like:

```
File "/home/ubuntu/letsencrypt/letsencrypt/plugins/selection.py", line 113, in choose_plugin
    code, index = disp.menu(question, opts, help_label="More Info")
File "/home/ubuntu/letsencrypt/letsencrypt/display/util.py", line 129, in menu
    return code, int(index) - 1
ValueError: invalid literal for int() with base 10: ''
```


bmw:
In general, I think the best approach here is to have all of the `NCursesDisplay` `IDisplay` methods to check whether dialog's exit code is `OK` or `HELP`. If not, it should return `CANCEL` as the exit code.

bmw:
This hasn't seen any progress in quite a while.

Closing the PR to help with my OCD, but feel free to open again when it's ready for another review.

pde:
Reopening this to work on it

pde:
Reopening for real now.

certbot/display/util.py
```diff
@@ -29,7 +29,6 @@
 HELP = "help"
 """Display exit code when for when the user requests more help."""
 
-
```
bmw: nit: This does nothing but cause `pep8` errors.

certbot/display/util.py
```diff
@@ -51,6 +50,21 @@ def _wrap_lines(msg):
 
     return os.linesep.join(fixed_l)
 
+
+def _clean(dialog_result):
+    """Work around inconsistent return codes from python-dialog.
```
bmw: nit: The codes aren't really inconsistent. They're all [documented](http://pythondialog.sourceforge.net/doc/Dialog_class_overview.html), it's just we only use and expect a subset of them. Maybe this comment should reflect that more accurately?

certbot/display/util.py
```diff
@@ -126,9 +140,9 @@ def menu(self, message, choices, ok_label="OK", cancel_label="Cancel",
                 (str(i), choice) for i, choice in enumerate(choices, 1)
             ]
             # pylint: disable=star-args
-            code, index = self.dialog.menu(message, **menu_options)
+            code, index = _clean(self.dialog.menu(message, **menu_options))
 
-            if code == CANCEL:
+            if code == CANCEL or index == "":
```
bmw: Question: Is this necessary after checking the exit code? Is it possible to have a OK or HELP exit code and an empty index string?

certbot/display/util.py
```diff
@@ -164,6 +178,7 @@ def yesno(self, message, yes_label="Yes", no_label="No", **unused_kwargs):
         :rtype: bool
 
         """
+        assert OK == self.dialog.DIALOG_OK, "What kind of absurdity is this?"
```
bmw: nit: If we want to leave this in, maybe we should check it at the top of the file or in `__init__` to catch the problem if we actually think there might be one early.

bmw:
What about the `checklist` and `directory_select` methods?

certbot/display/util.py
pde: I can't remember why I added this now.  Perhaps I saw this for real, or perhaps I was just being defensive?

certbot/display/util.py
pde: The reason this is here is the same as the reason we have these janky constants in the first place: they're not available from the `dialog` module, only from `dialog.Dialog` objects.  But maybe we should just make a Dialog object at import time...

bmw: But why not put in `__init__` where we have a `dialog.Dialog` object?

bmw:
Other than moving the assertion up and the logging level, this LGTM.

bmw:
LGTM

PR #2770, opened 2016-04-05
Reuse venv with le-auto

bmw:
With #2759 we forgot to run `build.py`.

Additionally, `letsencrypt-auto` recreates the venv on every run with Python 2.6. This behavior seems to exist in every 0.4.x release.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -700,7 +700,7 @@ except ImportError:
             cmd = kwargs.get("args")
             if cmd is None:
                 cmd = popenargs[0]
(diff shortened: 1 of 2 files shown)


bmw:
For clarity, the problem is on Python 2.6, the output of `letsencrypt --version` is:
```
/root/.local/share/letsencrypt/lib/python2.6/site-packages/cryptography/__init__.py:26: DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of cryptography will drop support for Python 2.6
  DeprecationWarning
letsencrypt 0.5.0
```
This does not equal `0.5.0` when ran through `cut -d " " -f 2` causing us to always recreate the venv.

pde:
LGTM

PR #2775, opened 2016-04-06
Add Directory.meta (fixes #2768)

### LINKED ISSUE #2768: Directory deserialization should allow 'meta' section
The spec has added support for a [`meta` section in the `directory` object](https://ietf-wg-acme.github.io/acme/#directory) which allows for a number of pieces of information that are not necessarily related to the ACME protocol. Currently the client will fail if these fields are returned which is causing letsencrypt/boulder#1692 to fail.

Also in general it would be good to be slightly less strict about JSON deserialization (i.e. ignoring unexpected fields) since we have other better ways of testing if the server/client start drifting out of sync.

kuba:


bmw:
LGTM. Thanks for fixing this so quickly @kuba.

PR #2785, opened 2016-04-08
Renewal interval in renewal configuration file

### LINKED ISSUE #422: Allude to renewal interval defaults in renewal config file
The renewal config file should somehow mention or explain the default renewal intervals even if none are explicitly set there (presumably in a comment created when the lineage was first created).

TheNavigat:
Closes #422

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -78,6 +78,9 @@ def write_renewal_config(o_filename, n_filename, target, relevant_data):
         if k not in relevant_data:
             del config["renewalparams"][k]
 
+    config["renew_before_expiry"] = constants.RENEWER_DEFAULTS["renew_before_expiry"]
+    config.comments["renew_before_expiry"] = "Renewal interval"
+
     # TODO: add human-readable comments explaining other available
     #       parameters
     logger.debug("Writing new config %s.", n_filename)
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -79,9 +79,8 @@ def write_renewal_config(o_filename, n_filename, target, relevant_data):
             del config["renewalparams"][k]
 
     if "renew_before_expiry" not in config["renewalparams"]:
-        config["renewalparams"]["renew_before_expiry"] = (
+        config["renewalparams"].comments["renew_before_expiry"] = (
             constants.RENEWER_DEFAULTS["renew_before_expiry"])
-    config["renewalparams"].comments["renew_before_expiry"] = ["Renewal interval"]
 
     # TODO: add human-readable comments explaining other available
     #       parameters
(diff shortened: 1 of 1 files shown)


schoen:
Thanks, @TheNavigat! The way that I understood #422 was that we might have a line like

`# renew_before_expiry = 30 days`

rather than

renew_before_expiry = 30 days # renewal default

That is, we would document the interval by having a commented-out version of a line showing the way to change it, rather than by actually causing the default to be recorded in the file. The former way, I thought, would make it easier if we wanted to change the underlying default elsewhere (because it wouldn't be overriden by the more-specific setting in the individual renewal config file).

Do you or other people think it's clearer the other way?

TheNavigat:
Yep, this does make sense, actually.

I've updated the code to reflect that. Thanks for the correction, @schoen.

schoen:
I don't think your change has the effect that you intended. I think it will attach a comment that says `30 days` or something instead of `renew_before_expiry = 30 days`. Or is there some reason that setting `comment[x] = "y"` can produce a renewal file entry that reads `# x = y`? That doesn't make too much sense to me.

TheNavigat:
Ah, I apologize for the screw up. I fixed that.

This might also seem a bit hazy, but I wasn't able to find a way to add a comment to a section, where that comment is not associated with other members of the section and such. ``final_comment`` seemed most logical, so I used it.

schoen:
I tested the new version and it put the comment in `[renewalparams]`. But, maybe confusingly, `renew_before_expiry` isn't one of the renewalparams at all -- it needs to exist outside of that section at the top level of the renewal config file!

TheNavigat:
Ah, I see!

The naming of the section is quite confusing. If ``renew_before_expiry`` doesn't belong to ``renewalparams``, then what does?

I've checked sample configuration files and adjusted the code accordingly, before testing it. I hope I didn't miss anything else.

certbot/storage.py
```diff
@@ -78,6 +78,9 @@ def write_renewal_config(o_filename, n_filename, target, relevant_data):
         if k not in relevant_data:
             del config["renewalparams"][k]
 
+    if "renew_before_expiry" not in config:
+        config.initial_comment = ["renew_before_expiry = 30 days"]
+
```
schoen: This worked properly for me and did what I expected. I think it might be more correct in principle to use the `constants.py` value rather than hard-coding it, which would look like
```
    if "renew_before_expiry" not in config:
        default_interval = constants.RENEWER_DEFAULTS["renew_before_expiry"]
        config.initial_comment = ["renew_before_expiry = " + default_interval]
```

TheNavigat: That popped in my head as well, but I thought that since it's a comment, it doesn't make much of a difference.

Your comment solves the dilemma. I'll add a commit in a few.

schoen:
Thanks for your help with this!

PR #2788, opened 2016-04-08
ignore EPERM when chown()ing dir for http-01

sthen:
The webroot plugin ignores EACCES when chowning the acme-challenge directory fails; OpenBSD returns EPERM in this case, so ignore it too.


bmw:
Thanks for the PR @sthen, but we decided to go with the simpler approach in #3063 of catching all `OSErrors` when `chown`ing or removing the challenge directory. #3063 should fix the problem you're having though and it will be in the `certbot` release later this week.

PR #2790, opened 2016-04-08
letsencrypt-auto --help now avoids bootstrapping

Flimm:
Users do not expect sudo to be invoked or for anything to be changed when passing the `--help` flag. This commit ensures that this expectation is met.

This fixes this issue: https://github.com/letsencrypt/letsencrypt/issues/1903

Let me know if there is anything wrong with this pull request, I'm happy to make any needed changes and I won't be offended if someone makes changes to my changes, I just want to fix this issue!


pde:
An equivalent PR shipped with Certbot 0.6.0, so this is now deprecated.

Flimm:
Do you mind linking to that PR? I couldn't find it when I searched for it.

bmw:
#2885

Issue #1243, opened 2015-10-31
apache install - Unable to finalize checkpoint


charlesay:
after generating the certificates I was unable to make them operational, first by entering the paths in the apache-config files.

running `./letsencrypt-auto install --apache --cert-path /etc/letsencrypt/live/www.eiserneketten.de/cert.pem --key-path /etc/letsencrypt/live/www.eiserneketten.de/privkey.pem` results in the error 
more extensive excerpt from the log, can't paste it here in full
> Unable to finalize checkpoint 
>2015-10-30 23:26:49,074:DEBUG:letsencrypt.error_handler:Encountered exception:
>Traceback (most recent call last):
  >File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/client.py", line 330, in deploy_certificate
>    self.installer.save("Deployed Let's Encrypt Certificate")
  >File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache/augeas_configurator.py", line 125, in save
>    raise errors.PluginError(str(err))
>PluginError: Unable to add title
>
>2015-10-30 23:26:49,074:DEBUG:letsencrypt.error_handler:Calling registered functions
>2015-10-30 23:26:49,179:DEBUG:letsencrypt.cli:Exiting abnormally:
>Traceback (most recent call last):
  >File "/home/server/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  >File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/cli.py", line 1131, in main
  >  return args.func(args, config, plugins)
  >File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/cli.py", line 507, in install
>    args.fullchain_pa2015-10-30 23:26:49,074:DEBUG:letsencrypt.error_handler:Encountered exception:
>Traceback (most recent call last):
  >File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt /client.py", line 330, in deploy_certificate
>    self.installer.save("Deployed Let's Encrypt Certificate")
>  File "/home/server/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt_apache /augeas_configurator.py", line 125, in save
>    raise errors.PluginError(str(err))
>PluginError: Unable to add title

[...]

PR #2793, opened 2016-04-08
Reverter.py: clock change protection, and debugging for #1243

### LINKED ISSUE #2792: Reverter could be confused by system clock changes
If the client is run with a system clock at time t1, and then run with an earlier time t0, `reverter.rollback_checkpoints` would wind up broken.  That would be especially sever in case of an installer error, but would also affect use of the `revert` verb.

pde:
Closes #2792


bmw:
LGTM

PR #2795, opened 2016-04-09
Remove text wrapping

### LINKED ISSUE #2754: Disable line wrapping in --text mode
When I run with `--text`, I get these lines as output:

```
Congratulations! You have successfully enabled https://02e645.le-test.hoffman-
andrews.com

You should test your configuration at:
https://www.ssllabs.com/ssltest/analyze.html?d=02e645.le-test.hoffman-
andrews.com
```

These lines are artificially wrapped at 80 characters, even though my terminal is much wider than that. This breaks the links and makes them (a) unclickable and (b) hard to paste into a browser.

I think with `--text`, we shouldn't wrap lines, but leave it to the user's terminal.

TheNavigat:
Closes #2754


jsha:
Looks like you remove wrapping in the ncurses display too? Doesn't that break things?

TheNavigat:
I didn't remove wrapping in ``NcursesDisplay``, nope.

Github changes view is tricky, it skipped around 200 lines between a couple of changes, making it look like I removed wrapping from ``NcursesDisplay`` instead of ``FileDisplay``. As you can see here, https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/display/util.py, ``NcursesDisplay`` doesn't even use wrapping.

I didn't rigorously test the changes, though. I relied heavily on the cover tests and the boulder integration tests, so I'd be happy to test the changes by actually running ``letsencrypt-auto`` if that is needed.

jsha:
Yes, please do test using a full run in both ncurses mode and in text mode. Thanks!

TheNavigat:
Just tested both and they work perfectly. Wrapping successfully removed when ``--text`` is used, and ``NcursesDisplay`` displays messages correctly.

jsha:
LGTM. @pde, want to give a final review and merge?

jsha:
Actually, un-LGTM. I pulled this down and tested against a local Boulder, and I still see unnecessary wrapping in the IMPORTANT NOTES output:

```
$ letsencrypt_test -d $(openssl rand -hex 3).le-test.hoffman-andrews.com -a standalone certonly
...
IMPORTANT NOTES:
   /tmp/leitCjo6/conf/live/a3b21e.le-test.hoffman-
   andrews.com/fullchain.pem. Your cert will expire on 2016-07-10. To
   obtain a new version of the certificate in the future, simply run
   Let's Encrypt again.
```

TheNavigat:
The ``IMPORTANT NOTES`` section works a bit differently. As you might have noticed, the text is printed to the console when ``NcursesDisplay`` is used as well.

https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/main.py#L51 is where the message is passed to the reporter (yep, reporter, not the display), and https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/reporter.py#L69-L107 is where the text is wrapped.

That is a little bit different from @pde 's comment here, https://github.com/letsencrypt/letsencrypt/issues/2754#issuecomment-206592938. Apparently it's quite easy to remove. However, I'm not sure would it affect other sections of Let's Encrypt.

Should I proceed to remove wrapping from ``reporter.py``?

jsha:
Yes, I think that's the right move. I think also moving the "Your cert will expire on " note to its own bullet might make things read more nicely on wide terminals after that change. Also interested in what the client team has to say here?

TheNavigat:
@jsha Done. Does that look better?

letsencrypt/reporter.py
```diff
@@ -16,6 +16,34 @@
 
 logger = logging.getLogger(__name__)
 
+def _get_terminal_size():
+    """Returns a list of terminal width and height"""
```
jsha: Why do you need this?

TheNavigat: For the indentation to work properly without wrapping the text, ``textwrap`` needs to be aware of the width of the terminal. Unfortunately passing 0 does not make it detect the width by itself, so I had to use that.

TheNavigat:
Ping @jsha

jsha:
I think this is the wrong approach. Why not just bypass the text wrapping entirely for non-ncurses output?

TheNavigat:
The indentation would look bad, basically

bmw:
I also think this is the wrong approach for the reporter. Instead, I think we can simply prepend every note with " - " to indent it and make every message visually distinct.

TheNavigat:
Fixed and tested. Does this look better for you?

bmw:
@jsha: Adding new lines to keep the same appearance, this PR causes output like the following on my laptop with a full screen terminal:
```
IMPORTANT NOTES:
 - Congratulations! Your certificate and chain have been saved at /etc/letsencrypt/live/example.com/fullchain.pem.
 - Your account credentials have been saved in your Certbot configuration directory at /etc/letsencrypt. You should make a secure backup of this folder now. This con
figuration directory will also contain certificates and private keys obtained by Certbot so making regular backups of this folder is ideal.
 - Your cert will expire on 2016-08-22. To obtain a new or tweaked version of this certificate in the future, simply run certbot again. To non-interactively renew *a
ll* of your ceriticates, run "certbot renew"
```
Is this the behavior you wanted? I personally think this looks worse, however, if you think it's fine, I'm happy to merge this PR for our next release. The code LGTM.

bmw:
So after talking to jsha out of band, we decided that the output above does look pretty bad and we have another solution. Sorry for changing our position on this, but since you've identified all places we use text wrapping in order to write this PR, implemented the new solution should be pretty easy if you're still interested.

The main problem identified in #2754 is we break things like URLs that would normally be clickable in a terminal. `textwrap` has options [break_long_words](https://docs.python.org/2/library/textwrap.html#textwrap.TextWrapper.break_long_words) and [break_on_hyphens](https://docs.python.org/2/library/textwrap.html#textwrap.TextWrapper.break_on_hyphens) which solve this problem. Using the examples, provided [here](https://github.com/certbot/certbot/issues/2754#issue-145527636), output becomes:

```
Congratulations! You have successfully enabled
https://02e645.le-test.hoffman-andrews.com

You should test your configuration at:
https://www.ssllabs.com/ssltest/analyze.html?d=02e645.le-test.hoffman-andrews.com
```
which solves the problem and still allows for text wrap to make the output look a little better.

@TheNavigat, this can be implemented in another PR or built on this one if you're still interested. You may also be able to write a common function passing the `break_long_words` and `break_on_hyphens` values to `textwrap` avoiding duplication. If only `util.py` uses this function, it should be placed there, otherwise it should probably go into `le_util.py`. Let me know if you have any questions!

bmw:
What about code that uses `_wrap_lines` and `reporter.py`?

TheNavigat:
As far as I understand, those don't include URLs

bmw:
Here's [one example](https://github.com/certbot/certbot/blob/master/certbot/main.py#L48) that does.

Even if that wasn't there though, I think we should solve the problem correctly now for all IDisplay/IReporter output because they may include URLs in the future causing this problem again.

certbot/display/util.py
```diff
@@ -39,12 +39,19 @@ def _wrap_lines(msg):
     :rtype: str
 
     """
+    wrapper = textwrap.TextWrapper(
```
bmw: In case you're unaware, you can pass these values directly to `textwrap.fill` if you want instead of creating a `TextWrapper` object. Either approach is fine though.

bmw:
What about the `input` and `_print_menu` methods of `FileDisplay` that call `textwrap.fill` directly?

TheNavigat:
Thanks for the note, fixed! :)

bmw:
LGTM. Will kick Travis and merge when integration tests are fixed.

PR #2797, opened 2016-04-09
ACME: omitempty Error.detail, Error.type (fixes #2289)

### LINKED ISSUE #2289: Don't fail when Detail field is empty
The Detail field of a [Problem Detail](https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-03) is not required, and may in particular be absent when it is redundant with the status field. Right now the client throws a nasty exception when it receives a Problem Detail without a Detail field. Instead it should just print the type.

kuba:


pde:
Fixes: #2955 #2289

PR #2809, opened 2016-04-11
Preserve domain ordering in CSR

### LINKED ISSUE #2798: Domain order not preserved in generated CSR
4d6a1ee7ff23b29d4416ee91610a7d8a58173292 introduced a change which causes the domain order passed to the client not to be preserved.

The previous implementation in `get_authorizations` would iterate over the original domain list, preserving the original order, whereas the new implementation uses `authzr.values()`, which does not follow the same sorting order.

As a side-effect, the Common Name field is not guaranteed to be the first domain passed via `-d`, causing issues for users who are relying on this after upgrading to 0.5.0.

bmw:
Fixes #2798.


pde:
This probably deserves a test case to protect against regressions.

bmw:
You're right. I modified a test so we'll catch this problem in the future.

jsha:
LGTM. FYI, on Boulder side all we care about is what's in the CN. You could in theory do this a little more simply by pulling out the first element before de-duping and always using that as CN, but I think the current version is fine.

Also, there are conflicts to be resolved.

certbot/tests/client_test.py
```diff
@@ -201,7 +201,8 @@ def test_obtain_certificate(self, mock_crypto_util):
 
         authzr = []
 
-        for domain in domains:
+        # domain ordering should not be affected by authorization order
+        for domain in reversed(domains):
```
SwartzCr: I don't understand why we're reversing domains here

schoen: Reversing the domains is meant to check that the order of the authorizations doesn't affect the subsequent order of the domains in the CSR.

schoen:
LGTM!

mccharlet:
H doesn't work. I have a strange behavior

I want the subject of my certificate are srv3.jxxxx.ch. I run this command (i take srv3.jxxxx.ch in the fist domain)

/root/letsencrypt/letsencrypt-auto certonly -a webroot --webroot-path /var/www/letsencrypt -d srv3.jxxxx.ch -d dexxxx.ch -d www.dexxxx.ch -d sxxxch -d www.sxxx.ch -d aaxxxx.ch -d www.aaxxxx.ch --server https://acme-v01.api.letsencrypt.org/directory --renew-by-default

The result in my CSR  (and the certificate) is the dexxxx.ch domain is the subject and not srv3.jxxxx.ch

Certificate Request:
    Data:
        Version: 2 (0x2)
        Subject: CN=dexxxx.ch
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption

Thanks

patf:
This patch isn't part of the released version of the client yet. It will be included in `0.6.0` once that is released. You can try your luck running from master [as described in the documentation](https://letsencrypt.readthedocs.io/en/latest/contributing.html#running-a-local-copy-of-the-client), but things might randomly break.

PR #2818, opened 2016-04-12
Systemd /etc/os-release parsing for distribution fingerprinting

joohoi:
Read /etc/os-release to fingerprint the distribution in use. Gives us larger fingerprint base, being able to identify Amazon Linux for example. Also platform.linux_distribution() is going to be removed from standard lib in 3.7, so we're prepared, and building on solid ground.

Still uses platform module as fallback, for older distribution versions not on systemd yet.

Out of major distributions only Slackware and Gentoo are not migrating / have not migrated to systemd, so we have a pretty good base covered here. 

(Not implemented yet) This also enables us to match minor distributions forked from bigger ones to their parents by using ID_LIKE field in the future. Check: https://www.freedesktop.org/software/systemd/man/os-release.html

Also fixed os version string for OS X, it was pulling only the first part of the version string (eg. resulting "10" for version string "10.10.5"). The first part has been 10 for the entirety of OS X lifetime, version 9 was still MacOS. Added missing tests for Darwin while I was at it.


joohoi:
If we still have the test farm for multiple distributions, would like to see it running through this PR.

bmw:
It looks like we have introduced new Apache test farm failures since our 0.5.0 release. I created #2837 to track that issue, but this PR introduces no new failures. Results:

```
0 ubuntu15.10 pass
1 ubuntu15.04LTS pass
2 ubuntu14.04LTS pass
3 ubuntu14.04LTS_32bit pass
4 ubuntu12.04LTS pass
5 debian8.1 fail
6 debian7.8.aws.1 fail
7 debian7.8.aws.1_32bit fail
8 amazonlinux-2015.09.1 fail
9 amazonlinux-2015.03.1 fail
10 RHEL7 pass
11 fedora23 pass
12 centos7 pass
```

[logfiles](https://drive.google.com/file/d/0B5w-1Ta1RduuLVBOMkdMOThYQ28/view?usp=sharing)

certbot/le_util.py
```diff
@@ -209,9 +209,75 @@ def safely_remove(path):
...
+    :rtype: `str`
+    """
+
+    var_string = varname+"="
+    if not os.path.isfile(filepath):
+        return ""
+    with open(filepath, 'r') as fh:
+        contents = fh.readlines()
+
+    for line in contents:
```
pde: Personally I find this idiom a bit nicer:
```
_, _, value = line.strip().partition(var_string)
if value:
   return _normalize_string(value)
```
Though neither approach is really elegant.

pde:
Trying this on my debian unstable system, I see:
```
In [1]: from certbot import le_util

In [2]: le_util.get_os_info()
Out[2]: ('debian', '')
```
Compared to current master:
```
In [1]: from certbot import le_util

In [2]: le_util.get_os_info()
Out[2]: ('debian', 'stretch/sid')
```

pde:
My `/etc/os-release/` is:

```
PRETTY_NAME="Debian GNU/Linux stretch/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

certbot/le_util.py
joohoi: I do like it better too, however .startswith() makes it more readable in my opinion, if someone has to look into / touch that in the future. Also the comment is in place to account for the unreadable part. Being able to quickly spot why we're getting in to that if clause makes up for it imo.

joohoi:
That's pretty interesting. Thanks for pointing it out. I think I'll have to reach out to debian devs to clarify the use (and/or lack of) of os-release. I'm hoping this only applies to unstable. However I think we're forced to move in to this detection method one way or another in the future, so better round the rough corners at this point.

joohoi:
Ok, so I talked to base-files package maintainer, and he cleared things up for me. Because for Debian, testing and unstable use a rolling release model, they are deliberately leaving the VERSION and VERSION_ID - fields off the os-release file, because they are meant for an actual release pinpointing.

Now leutil.get_os_info() is currently only used in apache plugin to determine the distribution name from the first part of the tuple it returns, and as UA - string in client.py. So I'd leave get_os_info() as it is, and write a new method for UA string, reading the NAME - variable from os-release. That is present in all of the (systemd) distributions after all, and the ID - field is enough to figure out the configuration paradigm for other use cases.

PR #2827, opened 2016-04-13
Fix Travis

bmw:
Fixes client integration tests in response to letsencrypt/boulder#1647.


bmw:
Self merging because it's not user facing and is blocking running Travis successfully in all other PRs.

PR #2836, opened 2016-04-14
Fix testing PRs with test farm

bmw:


bmw:
Merging because not user facing

PR #2849, opened 2016-04-15
Create LE shims

### LINKED ISSUE #2647: Python modules - possible/needed to rename?
from meeting notes:
- Python module names: do we keep letsencrypt.*, or change it?
- Renaming will be fine except where plugins import from us?
- Can we shim so that letsencrypt.* temporarily mirrors newname.*?

bmw:
Fixes #2647.

I tested this locally and the packages seem to work. Additionally, I also successfully tested that the underlying `certbot` package is updated when doing `pip install -U letsencrypt` without changing the `letsencrypt` package.

### DIFF
--- a/letsencrypt-apache/LICENSE.txt
+++ b/letsencrypt-apache/LICENSE.txt
@@ -1,0 +1,190 @@
+   Copyright 2015 Electronic Frontier Foundation and others
+
+   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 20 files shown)


SwartzCr:
looks reasonable to me, though I'd appreciate another +1
maybe from @hlieberman or @erikrose ?

hlieberman:
Yeah, I think this will work.

PR #2852, opened 2016-04-15
Discover all plugins

### LINKED ISSUE #2646: Ensure rename doesn't break third party plugins
There are a lot of third party plugins for Let's Encrypt (listed here: https://letsencrypt.readthedocs.org/en/latest/using.html#plugins)
It's possible that they reference things in such a way that changing the name or package of LE will break their compatibility with the client - we should ensure that they are unbroken

bmw:
Fixes #2646.

With the rename, we changed the group id where plugins are registered ([source](https://github.com/letsencrypt/letsencrypt/blob/master/setup.py#L132)).

We could arguably not do this. `letsencrypt.plugins` is an arbitrary string. It's not actually loading the module `letsencrypt.plugins` so there is no requirement we actually make this change. With that said, I suspect given a long enough time frame, having new plugins continue to use the `letsencrypt` name is a bit awkward.

This PR causes our plugin discovery code to search both `letsencrypt.plugins` and `certbot.plugins`. I tested this code with #2849 and `letsencrypt-plesk` everything worked.

### DIFF
--- a/certbot/constants.py
+++ b/certbot/constants.py
@@ -8,6 +8,9 @@ from acme import challenges
 SETUPTOOLS_PLUGINS_ENTRY_POINT = "certbot.plugins"
 """Setuptools entry point group name for plugins."""
 
+OLD_SETUPTOOLS_PLUGINS_ENTRY_POINT = "letsencrypt.plugins"
+"""Plugins Setuptools entry point before rename."""
+
 CLI_DEFAULTS = dict(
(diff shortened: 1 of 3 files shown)


SwartzCr:
I am in favor of supporting both in the hope that we can migrate someday
We can either pressure people directly (in the case of plugins) or at some point we'll have better user prompts

PR #2853, opened 2016-04-15
Don't pause when reporting renewal isn't necessary

### LINKED ISSUE #2831: --keep-until-expiring requires interaction if no certificate update
Running a command of `./letsencrypt-auto --config /opt/letsencrypt.ini certonly --keep-until-expiring --agree-tos -d my.domain.com ` with configuration file contents that includes the option `text = True` results in prompted input if certificates are not yet due for removal, breaking automated scripts.

## STR:
1. Request certificates using let's encrypt (using the options provided below).
2. Request certificates again using `--keep-until-expiring` flag set and a configuration file that specifies `text = True`, the same that should have been used for the first request:
  2. An example command would be `./letsencrypt-auto --config /opt/letsencrypt.ini certonly --keep-until-expiring --agree-tos -d my.domain.com`
  2. with a configuration file of:
  ```
# Let's Encrypt Configuration with defaults
# All flags used by the client can be configured here. Run Let's Encrypt with
# "--help" to learn more about the available options.

# Use a 4096 bit RSA key instead of 2048
rsa-key-size = 4096

# Uncomment and update to register with the specified e-mail address
email = EMAIL_HERE@EMAIL_HERE

# Uncomment and update to generate certificates for the specified
# domains.
# domains = example.com, www.example.com
# Always provide domains via env. See letsencrypt-run.sh

# Uncomment to use a text interface instead of ncurses
text = True

# Uncomment to use the standalone authenticator on port 443
# authenticator = standalone
# standalone-supported-challenges = tls-sni-01

# Uncomment to use the webroot authenticator. Replace webroot-path with the
# path to the public_html / webroot folder being served by your web server.
authenticator = webroot
webroot-path = /var/www/challenges
```

## Expected:
1. Let's Encrypt client prints:
```
Certificate not yet due for renewal; no action taken.                           
```
and does not prompt for input afterward, allowing the client to exit with a code of `0`.

## Actual:
1. Let's Encrypt client prints:
```
Certificate not yet due for renewal; no action taken.                           
Press Enter to Continue
```
and requires interaction from a tty before exiting cleanly. 

## Environment
- OS = Ubuntu 14.04.2 
- Webserver = nginx, but using certonly since acme-challenges are successfully handled through an existing nginx server. The `letsencrypt.ini` file specifies the correct challenge directory for the client by providing the options 
  - 
  ```
authenticator = webroot
webroot-path = /var/www/challenges
```

### Notes
Let's Encrypt was installed by running and then calling:
```
git clone https://github.com/letsencrypt/letsencrypt
./letsencrypt-auto --config /opt/letsencrypt.ini certonly --keep-until-expiring --agree-tos -d my.domain.com
```

bmw:
Fixes #2831.

### DIFF
--- a/certbot/tests/main_test.py
+++ b/certbot/tests/main_test.py
@@ -13,6 +13,14 @@ from certbot.plugins import disco as plugins_disco
 class ObtainCertTest(unittest.TestCase):
     """Tests for certbot.main.obtain_cert."""
 
+    def setUp(self):
+        self.get_utility_patch = mock.patch(
+            'certbot.main.zope.component.getUtility')
+        self.mock_get_utility = self.get_utility_patch.start()
+
+    def tearDown(self):
+        self.get_utility_patch.stop()
+
     def _call(self, args):
         plugins = plugins_disco.PluginsRegistry.find_all()
         config = configuration.NamespaceConfig(
@@ -26,9 +34,14 @@ class ObtainCertTest(unittest.TestCase):
 
     @mock.patch('certbot.main._auth_from_domains')
     def test_no_reinstall_text_pause(self, mock_auth):
+        mock_notification = self.mock_get_utility().notification
+        mock_notification.side_effect = self._assert_no_pause
         mock_auth.return_value = (mock.ANY, 'reinstall')
-        # This hangs if the reinstallation notification pauses
         self._call('certonly --webroot -d example.com -t'.split())
+
+    def _assert_no_pause(self, message, height=42, pause=True):
+        # pylint: disable=unused-argument
+        self.assertFalse(pause)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


certbot/tests/main_test.py
```diff
@@ -0,0 +1,35 @@
...
+        from certbot import main
+        with mock.patch('certbot.main._init_le_client') as mock_init:
+            main.obtain_cert(config, plugins)
+
+        return mock_init()  # returns the client
+
+    @mock.patch('certbot.main._auth_from_domains')
+    def test_no_reinstall_text_pause(self, mock_auth):
+        mock_auth.return_value = (mock.ANY, 'reinstall')
+        # This hangs if the reinstallation notification pauses
```
SwartzCr: is there a way to alert on this? will the test fail silently in the future if a pause is added back?

bmw: Previously, I was trying to minimize `mock` usage and had the idea that the tests would just hang. There were a number of problems with that though.

The new version mocks out `getUtility` and asserts that `pause` always equals `False`.

PR #2855, opened 2016-04-16
Reuse HTTP connections.

### LINKED ISSUE #2778: Connectionpool not pooling
When I run with `-v`, I see entries like this repeatedly:

2016-04-07 05:23:55,283:INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): acme-v01.api.letsencrypt.org
2016-04-07 05:23:58,509:INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): acme-v01.api.letsencrypt.org

This probably means that the connectionpool is not actually reusing connections properly, which makes the process slower than it should be.

jsha:
Fixes #2778


joohoi:
IIRC requests.session() leaves the open connection hanging and its use in classes is best accompanied by self.session.close() in magic method __del__

jsha:
I fixed the tests and this is now ready for review again.

acme/acme/client_test.py
```diff
@@ -528,40 +528,48 @@ def test_check_response_jobj(self):
...
+        self.net._send_request('GET', 'http://example.com/',
+                               headers={'User-Agent': 'foo2'})
+        self.net.session.request.assert_called_with(
+            'GET', 'http://example.com/',
+            verify=mock.ANY, headers={'User-Agent': 'foo2'})
+
+    def test_del(self):
+        self.net.session = mock.MagicMock()
+        self.net.__del__()
+        self.net.session.close.assert_called_once()
```
joohoi: something along the lines of:
```
          sess = mock.MagicMock()
          self.net.session = sess
          del self.net
          sess.close.assert_called_once()
```
would be more "true" to the actual functionality.

joohoi:
LGTM despite the nit above

jsha:
Fixed the nit.

Issue #2822, opened 2016-04-13
Renewal: append() takes exactly one argument (2 given)


sebix:
Using letsencrypt 0.5.0, I got this error when renewing my certificates (subcommand renew):
```shell
2016-04-13 07:29:24,428:INFO:letsencrypt.renewal:Cert not yet due for renewal
2016-04-13 07:29:24,429:DEBUG:letsencrypt.main:Exiting abnormally:
Traceback (most recent call last):
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/main.py", line 692, in main
    return config.func(config, plugins)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/main.py", line 538, in renew
    renewal.renew_all_lineages(config)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/renewal.py", line 355, in renew_all_lineages
    renew_skipped, parse_failures)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/letsencrypt/renewal.py", line 290, in _renew_describe_results
    notify(parse_failures, "parsefail")
TypeError: append() takes exactly one argument (2 given)
```

kevinburke:
Happy to write a test here but I'm not seeing any tests for `renew_all_lineages` or `renew_describe_results` or Python tests that mention `renew`, maybe I'm not looking in the right place?

bmw:
Fixed in #2860.

PR #2860, opened 2016-04-18
Pass a single argument to append()

kevinburke:
The other notify() calls in this block all pass the output of report(), but
this one attempts to pass two arguments, which results in the stack trace
described in #2822.

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -287,7 +287,7 @@ def _renew_describe_results(config, renew_successes, renew_failures,
     if parse_failures:
         notify("\nAdditionally, the following renewal configuration files "
                "were invalid: ")
-        notify(parse_failures, "parsefail")
+        notify(report(parse_failures, "parsefail"))
 
     if config.dry_run:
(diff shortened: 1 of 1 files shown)


kevinburke:
Test failure, looks like the system shut down? I don't know much about the tests but am doubtful that my change caused it :(

kevinburke:
Latest master also had a build fail with "stalled": https://travis-ci.org/letsencrypt/letsencrypt/jobs/123479998

bmw:
I believe the system shutdown was caused by some problems on Travis' end over the weekend, however, now builds are failing due a change in `boulder`. If you merge `master` into this PR, it should fix the problem.

kevinburkeshyp:
Rebased against master, still failing with a "stalled build" message

bmw:
@jsha: tests hung while running `create_db.sh`

Kicking the build again. Sorry for the trouble!

bmw:
LGTM. Thanks for the PR @kevinburke.

PR #2864, opened 2016-04-18
Fix problem with godep and vendor directories

bmw:
`boulder` team told me this should fix our integration tests.


bmw:
Self merging since this change is not user facing.

PR #2866, opened 2016-04-19
Remove beta software disclaimer

krikmo:
*Let's Encrypt*'s blog [explains](https://letsencrypt.org/2016/04/12/leaving-beta-new-sponsors.html) that it is no longer Beta software, therefore that [disclaimer](https://github.com/letsencrypt/letsencrypt#disclaimer) is no longer necessary.

The FAQ link has value and should come back somewhere in the *[About the Let's Encrypt Client](https://github.com/letsencrypt/letsencrypt#about-the-lets-encrypt-client)* section, I think.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,16 +1,5 @@
 .. notice for github users
 
-Disclaimer
-==========
-
-The Let's Encrypt Client is **BETA SOFTWARE**. It contains plenty of bugs and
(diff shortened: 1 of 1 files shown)


kuba:
Let's Encrypt is not beta, but Certbot might be ;)

krikmo:
Oh, ok then :)

You folks are doing an awesome work here!

PR #2867, opened 2016-04-19
Proof-of-concept Heroku support

brentdax:
I've been working on a reasonably full-featured Heroku configurator plugin. This plugin needs to be run in the working directory of a Heroku project and provided with the name of a remote that pushes to a Heroku app. When run, it can:
- Use the Heroku API to discover the domains associated with the app
- Use git to commit, push, and validate HTTP-01 challenges with a live app
- Use the Heroku API to install an issued certificate onto an app's existing SSL Endpoint add-on

**This work is not ready to merge.** It is currently just barely working; there are no tests, it has a number of outright hacks in it, and I've only tried it in the very simple configurations I need for my projects. I'll most likely continue working on and improving it over the next week or so.

However, I'm wondering if there is interest in eventually merging it once it's in a more complete state. So far I have been developing it in a fork of this repository. If you think you'd eventually be interested in merging it, I'll continue to develop it that way; otherwise, I'll turn it into a standalone plugin so that users who need it can install it themselves.

I'm also wondering if you would be interested in feedback on the `IInstaller` interface and on installation more generally in this unusual environment. `IInstaller` seems to assume that certificates, keys, and configuration are all written into files and then the server is restarted to read them, but Heroku uses a REST API and automatically restarts services when the relevant configuration changes, so that model is quite awkward for this plugin. A few operations appear to be impossible to implement—there is no documented endpoint for retrieving an SSL Endpoint's private key, for example. Even on the UI end of things, the Heroku plugin has to perform many relatively slow network operations, and would greatly benefit from some kind of cancelable "doing X, please wait" call in the display module.

And, of course, I would welcome feedback on how to make this fit into the project better. I have no background in this project—and in fact, no background in Python at all—so making this work took a lot of trial and error and not a few dumb hacks. (For instance, the `git` and `heroku` command-line tools have to be run as the original user to work properly, so when I'm running as root, I `sudo` back to the original user to run them.) There are probably some truly awful things in here which someone with even a little bit of experience could improve.

Finally, there's a philosophical issue I would like feedback on: I could most likely add an Enhancement to set up a new SSL Endpoint add-on if there isn't already one configured. However, the add-on costs money (US$20 per month). I'm not sure how to present this option to the user or even if it should be included at all.

Thanks for reading!


pde:
This is awesome! (And sorry we've been a little slow to reply, the Certbot transition has been a little distracting).

In the long run we're definitely interested in either merging, or supporting your work in a third-party repo by providing installation instructions for your plugin via the newly launched https://certbot.eff.org/ UI (which is created here: https://github.com/certbot/website/tree/master/_scripts/instruction-widget) and this [section of the docs](https://github.com/certbot/certbot/blob/master/docs/using.rst#plugins).

The main advantage of the third party approach is that you wouldn't be blocked by our release schedule and limited resources for reviewing PRs; that's the approach that the plesk, gandi and s3front plugins have taken.

pde:
Regarding the `IInstaller` model: does it make sense for you to use the `IInstaller.restart` event as a trigger to push the current values of things to the REST API? Or would we need to change the model entirely to make things non-inelegant?

pde:
Regarding the question of actions that cause costs for users: you have the option of calling `IDisplay` methods (the Apache plugin does [quite a bit of that](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/display_ops.py)) to get permission from the user before incurring costs.

There's currently no across-the-board method for triggering `enhance` calls from the client; we're probably going to add some infrastructure of that sort as we work on #1123, though I'm not sure it makes sense to pick up arbitrary enhancements from plugins and offer them to the user directly via UI.

brchristian:
Starting today, it appears that SSL will be free on Heroku: https://devcenter.heroku.com/articles/ssl-beta

This may remove the philosophical issue about presenting this option to users!

pde:
@brentdax are you planning to do more with this?

bmw:
We haven't heard back on this PR in over 4 months. If you're still interested in continuing work on this, please open another PR when you'd like us to review.

PR #2868, opened 2016-04-19
letsencrypt-auto: don't use --oneshot when emerging dependencies on Gentoo

5lava:
Installed dependencies won't be _suddenly_ removed by `emerge --depclean`. This behaviour is also on par with what's being done on other Linux distros (`pkg install`, `yum install`, `brew install`, `port install`, etc.).

`man portage`:
> WARNING: This option should only be used for packages that are reachable from the `@world` package set (those that would not be removed by `--depclean`)...

### DIFF
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -328,13 +328,13 @@ BootstrapGentooCommon() {
 
   case "$PACKAGE_MANAGER" in
     (paludis)
-      $SUDO cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
+      $SUDO cave resolve --keep-targets if-possible $PACKAGES -x
       ;;
     (pkgcore)
-      $SUDO pmerge --noreplace --oneshot $PACKAGES
+      $SUDO pmerge --noreplace $PACKAGES
       ;;
     (portage|*)
-      $SUDO emerge --noreplace --oneshot $PACKAGES
+      $SUDO emerge --noreplace $PACKAGES
       ;;
   esac
 }
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -328,13 +328,13 @@ BootstrapGentooCommon() {
 
   case "$PACKAGE_MANAGER" in
     (paludis)
-      $SUDO cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
+      $SUDO cave resolve --keep-targets if-possible $PACKAGES -x
       ;;
     (pkgcore)
-      $SUDO pmerge --noreplace --oneshot $PACKAGES
+      $SUDO pmerge --noreplace $PACKAGES
       ;;
     (portage|*)
-      $SUDO emerge --noreplace --oneshot $PACKAGES
+      $SUDO emerge --noreplace $PACKAGES
       ;;
   esac
 }
--- a/letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
+++ b/letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
@@ -11,13 +11,13 @@ BootstrapGentooCommon() {
 
   case "$PACKAGE_MANAGER" in
     (paludis)
-      $SUDO cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
+      $SUDO cave resolve --keep-targets if-possible $PACKAGES -x
       ;;
     (pkgcore)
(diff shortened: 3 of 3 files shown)


osirisinferi:
The problem without `--oneshot` is a lot of **dependencies** in your world file, which is bad too.

Ideally, one should use [the official Gentoo package](https://packages.gentoo.org/packages/app-crypt/letsencrypt) which handles all the deps without adding them to the world list ánd preventing them from removal with `--depclean`..

In essence, `letsencrypt-auto` is a poor work-around (in my humble opinion) for the lack of official packages, but polluting your `world` file is not the way to go IMHO.

5lava:
Well, if I had to choose between few extra records in my `world` and a _suddenly_ broken program that is crucial to all my websites, I'd prefer first.

Besides, for most Gentoo systems "polluting" actually comes down to just three packages out of [eight](https://github.com/letsencrypt/letsencrypt/blob/4d4cfb414f2847f1814cfe2949c6e76f112e4ade/letsencrypt-auto#L319) — `dev-python/virtualenv`, `app-admin/augeas` and `dev-util/dialog`, because the remaining five are already installed:

- `dev-libs/libffi` (via `virtual/libffi`) — is required by python;
- python (`dev-lang/python:2.7`) is required by portage itself;
- `virtual/pkgconfig` — by gcc, glibc and dozens of other major packages;
- `dev-libs/openssl` and `app-misc/ca-certificates` — ahem, obviously.

That being said, the list of packages `emerge`'d by `letsencrypt-auto` can (and should IMO) be reduced to 3.

> Ideally, one should use the official Gentoo package

I tried to, but quickly realized that `-auto` is totally enough for my needs. And, speaking of pollution, `-auto` adds just 3 packages to my system, while the official package wants to add 43. Oh, and 9 lines to my `/etc/portage/package.keywords`. No, thank you.

osirisinferi:
`letsencrypt-auto` shouldn't be suddenly broken as it should install missing deps every time it's ran.

And about the difference in dependencies between the apparently short list of `letsencrypt-auto` and `app-crypt/letsencrypt` you're mentioning: did you see the [huge list of deps `letsencrypt-auto` installs using `pip` áfter your mentioned short list](https://github.com/letsencrypt/letsencrypt/blob/v0.5.0/letsencrypt-auto#L467)? The ebuild isn't listing those deps because Manuel thinks that's funny :wink: Using `letsencrypt-auto` you're installing a ton of deps outside of the package manager of Gentoo.. No thank you :wink:

pde:
@5lava what were the things that `letsencrypt-auto` was doing for you that the Gentoo packages weren't?

(One advantage of `letsencrypt-auto` is that you're always running the latest version of the client. But the downside is that we're exposed to all of the fragility that `pip` and `virtualenv` and `python-cryptogrpahy` compilation seem to introduce, and it leaves a lot of problems up to the small letsencrypt team, rather than people who know each platform best -- which is hopefully the distributions themselves...)

osirisinferi:
> One advantage of letsencrypt-auto is that you're always running the latest version of the client.

As Gentoo is mainly based on non-binary packages which need to be compiled, many packages have a "9999" version number in the repository, [just like Let's Encrypt does](https://packages.gentoo.org/packages/app-crypt/letsencrypt). These packages will pull the latest CVS/Git/et c. files and compile it. So "re-merging" those packages will get you the most recent, bleeding egde software, just like `letsencrypt-auto`. Ofcourse it's not 100% comparible with `letsencrypt-auto`, because Gentoo will obviously not get the most recent files on running `letsencrypt`, but will need a manual remerge of the package, but it's very much possible to get the most recent software without the need for `letsencrypt-auto` :wink:

5lava:
Hey all,
Apologies for late response.

@osirisinferi
> huge list of deps

Indeed, I haven't noticed that! Thank you for pointing that out.

> letsencrypt-auto shouldn't be suddenly broken as it should install missing deps every time it's ran.

Right you are! I should've kept `--debug` for each invocation, not only the first one.

@pde
> the things that letsencrypt-auto was doing for you that the Gentoo packages weren't

1. Auto-update. Very useful considering the software was in beta. And it's much faster than emerging version 9999.
2. I noticed that `-auto` installs less packages so I thought it's kinda lighter (especially considering that I'm not using GUI, so a bunch of python stuff isn't needed for sure). But as @osirisinferi pointed out — I was wrong, packages are [still installed](https://github.com/letsencrypt/letsencrypt/blob/v0.5.0/letsencrypt-auto#L467).

And btw, I believe the PR can be closed, I agree that `--oneshot` makes sense.

PR #2873, opened 2016-04-19
Fix test farm tests

bmw:
Looks like I didn't change all references to this function when I did my rename PR, breaking test farm tests. This PR fixes them.


bmw:
Self merging PR as it modifies no user facing code.

PR #2841, opened 2016-04-14
Module rename

bmw:
Fixes #2739 and does everything but the shim for #2647.

PR #2875, opened 2016-04-20
Deprecation warning for non-updating letsencrypt-auto

### LINKED ISSUE #2502: Print a deprecation warning for users running the old letsencrypt-auto
Parts:
* wrangle sudo to pass a variable in to the client
* if the client detects leauto but neithert that variable nor `--no-self-upgrade`, its the old non-self-updating version, so print a warning

pde:
Closes #2502

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -649,7 +649,6 @@ def main(cli_args=sys.argv[1:]):
     args = cli.prepare_and_parse_args(plugins, cli_args)
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
-    cli.possible_deprecation_warning(config)
 
     # Setup logging ASAP, otherwise "No handlers could be found for
     # logger ..." TODO: this should be done before plugins discovery
@@ -662,6 +661,7 @@ def main(cli_args=sys.argv[1:]):
     le_util.make_or_verify_dir(
         config.logs_dir, 0o700, os.geteuid(), "--strict-permissions" in cli_args)
     setup_logging(config, _cli_log_handler, logfile='letsencrypt.log')
+    cli.possible_deprecation_warning(config)
 
     logger.debug("certbot version: %s", certbot.__version__)
     # do not log `config`, as it contains sensitive data (e.g. revoke --key)!
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -649,6 +649,7 @@ def main(cli_args=sys.argv[1:]):
     args = cli.prepare_and_parse_args(plugins, cli_args)
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
+    cli.possible_deprecation_warning(config)
 
     # Setup logging ASAP, otherwise "No handlers could be found for
     # logger ..." TODO: this should be done before plugins discovery
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -37,8 +37,9 @@ helpful_parser = None
 # should only be used for purposes where inability to detect letsencrypt-auto
 # fails safely
 
+LEAUTO = "letsencrypt-auto"
 fragment = os.path.join(".local", "share", "certbot")
-cli_command = "letsencrypt-auto" if fragment in sys.argv[0] else "certbot"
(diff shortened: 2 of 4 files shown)


pde:
I don't have a good strategy for testing this.  Once we've released it, we could write an integration test that invokes an ancient copy of `letsencrypt-auto` and checks for the warning, and invokes a recent copy and checks for its absence.

Maybe we should make that test a todo item after 0.6.0 is signed but before it's pushed to PyPI?

pde:
Actually maybe this is better for erik to review?

bmw:
This looks pretty good. A couple things though:

1) There's a bunch of changes to `letsencrypt-auto` which don't show up in `letsencrypt-auto.template`, meaning they will lost the next time `build.py` is run (which should be done anyway before merging this PR). With that said, I looked through all of the changes that would be lost and they are just `letsencrypt` -> `certbot` string changes and I believe I changed all necessary strings in #2885.

2) As discussed out of band, one thing we should do is change the `cli_command` logic to be `letsencrypt-auto` or `certbot-auto` dependent on the `CERTBOT_AUTO` variable. I said this could be in another PR, but we could just do it in this PR and then the logic for `possible_deprecation_warning` would simply be:
```
if cli_command == LE_AUTO and not config.no_self_upgrade:
    logger.warning("blah")
```
This to me seems like a worthwhile change, however, if you disagree we could put it in another PR.

3) Test farm tests should be run on this PR once it is updated to reflect review comments. I'm happy to do it myself if you'd prefer.

pde:
I think the stray `letsencrypt-auto` changes were possibly a result of someone else committing something without having re-run build.py?  If I merge master, and rebuild, I get no diff, at any rate.

bmw:
Ah you're right. Looks like that was me in #2841.

pde:
OK, I've changed the LEAUTO behaviour as you suggested. I've tested this a fair bit interactively, including these cases:

* Running with `sudo` and `su`
* Running from `letsencrypt-auto` and `certbot-auto`
* Running without a `-auto` wrapper

However my tests required a lot of leauto editing, so I'm not sure how to make them part of the test suite just yet.

certbot/main.py
```diff
@@ -649,6 +649,7 @@ def main(cli_args=sys.argv[1:]):
     args = cli.prepare_and_parse_args(plugins, cli_args)
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
+    cli.possible_deprecation_warning(config)
```
bmw: This should be moved after the call to `setup_logging`. Otherwise, the only output is `No handlers could be found for logger "certbot.cli"`.

tools/venv.sh
```diff
@@ -15,5 +15,7 @@ fi
   -e .[dev,docs] \
   -e certbot-apache \
   -e certbot-nginx \
+  -e letsencrypt \
+  -e letsencrypt-apache \
```
bmw: Why is this here?

letsencrypt-auto
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Download and run the latest release version of the Let's Encrypt client.
+# Download and run the latest release version of the Certbot client.
```
bmw: All changes to this file should be reverted. It should remain the same as the file from our last release.

bmw:
It looks like something went wrong with the merge in 3a32ab8. Additionally, I didn't have time to write/run test farm tests for this last night.

Unless we can merge this early today, I think this PR should just miss the 0.6.0 train and we can include it in our next release.

pde:
Ooops, sorry, I think I had some unstaged changes in my tree that I was keeping out of the branch, but which got pulled in by `git merge` demanding a `commit -a`.

pde:
Hopefully this is fixed now. I don't think it's critical that it catch this train, but it's also probably of value to some users.

pde:
@bmw up to you whether to test & merge or kick. This change wouldn't be canary-tested in any case.

pde:
oops, the venv change is still there...

pde:
Nope, ghosts in my github.

bmw:
Successfully ran test farm tests. LGTM

PR #2879, opened 2016-04-21
Add privacy information to the Subscriber Agreement question

pde:
Currently the Subscriber Agreement agreement interaction in the client does not address privacy issues directly (though they are part of Let's Encrypt's Subscriber Agreement).  This PR attempts to make Certbot's privacy stance clear in the same screen.  Not sure yet if this is too wordy, but it may be the best way forward.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -37,11 +37,11 @@ def acme_from_config_key(config, key):
     "Wrangle ACME client construction"
     # TODO: Allow for other alg types besides RS256
     net = acme_client.ClientNetwork(key, verify_ssl=(not config.no_verify_ssl),
-                                    user_agent=_determine_user_agent(config))
+                                    user_agent=determine_user_agent(config))
     return acme_client.Client(config.server, key=key, net=net)
 
 
-def _determine_user_agent(config):
+def determine_user_agent(config):
     """
     Set a user_agent string in the config based on the choice of plugins.
     (this wasn't knowable at construction time)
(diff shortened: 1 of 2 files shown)


pde:
Here are some "before" screenshots of the current interaction:

pre
pre2

pde:
And here are some "after" screenshots from this branch:

post
post2

schoen:
I think the biggest privacy "surprise" for users is not necessarily the information that the CA collects, but rather the way the cert contents are automatically made public through CT. (Maybe people who've gotten certs from other non-CT-using CAs in the past would be especially surprised here.) If there's a way that we could usefully convey that to Certbot users, I would definitely advocate for it.

I haven't seen a huge amount of complaining about this issue, but I feel like a handful of people have mentioned it now and then, or at least been super-secretive about their cert contents as though they didn't know that that information had already been irrevocably published to the whole world!

schoen:
Is there a reason to mention that the requested domain names are sent, but not the subject public key?

pde:
Yes will add the public key.  I wonder if this should be broken down by "information that's inherently required for your public cert" (names, keys, server IP) and "information that is any sense optional" (email address and user agent)

SwartzCr:
Looks like tests are failing due to linting.
I'm happy with this added wording, though would also be fine with a mention of CT logging

pde:
I m now leaning against making this change; since we've documented all of this in the Certbot privacy policy, this would just make that screen much wordier and more confusing.

Issue #2884, opened 2016-04-22
Modify GitHub URL in letsencrypt-auto


bmw:
There is no rush to do this. I verified for #2639 that `letsencrypt-auto` will still receive updates after the repo rename. Changing the URL in `letsencrypt-auto` should only be done for aesthetic.

If we don't rename the repo until shortly before the rename, I believe we should wait to do this until 0.7.0 to avoid making last minute changes.

SwartzCr:
I'm in favor of moving this to 0.7.0 along with the pypi change, so long as there's no breakage from the gh redirects

bmw:
We decided to move this to 0.7.0.

Issue #2877, opened 2016-04-20
Determine if *-auto should be a symlink


bmw:
Unfortunately, the way that GitHub handles symlinks isn't ideal for `*-auto`. For example, here's a [symlink to a directory](https://github.com/certbot/testrepo2/blob/master/amul2). Symlinks to files are treated the same way. The raw contents of the symlink is simply the target's path, not the contents.

With this in mind I propose we leave `letsencrypt-auto-source` largely unmodified. Moving this directory or making it a symlink to `certbot-auto-source` will break `letsencrypt-auto`. I also think keeping both `letsencrypt-auto` and `certbot-auto` in the `letsencrypt-auto-source` directory is unnecessary as they will be the same file. This means both `certbot-auto` and `letsencrypt-auto` will be downloading `letsencrypt-auto-source/letsencrypt-auto` for updates, but this is an implementation detail and not a big deal in my opinion.

The other question we have to consider is about `*-auto`s in the root of our repo. `letsencrypt-auto` must remain in our repo in some form. All of our installation instructions use it and have done so since the beginning of public beta.

Do we want a `certbot-auto` in the root of our repo? I think we do as having `letsencrypt-auto` be the only `-auto` script in our repo when `certbot-auto` is the preferred form is a bit awkward. We could make one of the scripts a symlink to the other. The problem is this breaks anyone doing `wget https://raw.githubusercontent.com/letsencrypt/letsencrypt/master/letsencrypt-auto` (or with `certbot-auto`). While I haven't seen anyone do this, we can avoid this potential problem by having full copies of both `certbot-auto` and `letsencrypt-auto` in the root directory. I think I prefer this approach over using symlinks.

Thoughts?

SwartzCr:
There was some discussion on the call today about whether we should include two scripts or to just have a symlink from le-auto to cb-auto. The concern is that this would break things for people who wget le-auto and try to run it.
I don't think anyone has scripted LE to work this way, and I'm mildly in favor of it because it means that people who are following old documentation will know that they've done something wrong when they wget le-auto and don't get a file back - especially since we'll no longer be telling people to wget the le-auto script.

SwartzCr:
maybe @pde should give input as well

SwartzCr:
We have decided to ship two identical files

bmw:
I'll make a PR that copies `certbot-auto` to the root of the tree during the release.

Issue #2848, opened 2016-04-15
certbot-auto


bmw:
We need to make a `certbot-auto` that doesn't install OS packages without prompting the user.

Question: Do we need/want to do this for the `0.6.0` release?

Issue #2839, opened 2016-04-14
Point letsencrypt-auto to the new package on PyPI


bmw:

bmw:
We have two options here:

1. Release a 0.5.0 version of `certbot` containing shim packages around `letsencrypt`. This is a fair amount of extra work to create temporary shim packages so `letsencrypt-auto` continues to work and pass tests until the 0.6.0 release.
2. Wait to do this until the 0.7.0 release. Have `certbot-auto/letsencrypt-auto` use the `letsencrypt` shim packages during 0.6.0. This means `*-auto` has to download two extra packages (`letsencrypt` and `letsencrypt-apache`) and we'll have to do an `0.7.0` release of the shim packages to allow `*-auto` to update to the new version of itself that points to the `certbot` packages.

I'm in favor of option 2. Either approach involves doing another release of shim packages and we already have the `letsencrypt` shim packages. Additionally, there is still plenty of work to be done for 0.6.0 and there is less of a deadline for 0.7.0

cc @SwartzCr

SwartzCr:
This problem exists because as soon as we make a merge pointing away fro pypi's le-auto all of our tests break since there isn't a cb package in pypi since we haven't done a release

@hlieberman has suggested doing the merge in private, pushing it to pip and then do the release in public.
I think we're all in agreement that there's not much scary with doing this in two seperate releases since the 'extra packages' that people will have to download are just the shim packages - and if we trust them a little it's fine to trust them for everyone (?) but also we'll update cb-auto in the 0.7.0 release to stop pointing at them

I'm in favor of Harlan's rogue plan - as long as people are in the office to do it, or for option 2. But we can decide closer to the release since neither really take much additional coding now

pde:
I'm in favour of the plan that's less work here.  It could be 0.7.0, or even just 0.6.1.  In fact, we could release 0.6.0 and 0.6.1 at basically the same time.

SwartzCr:
that was discussed, the thought was that that'd be the same amount of release work as releasing 0.6.0 and 0.7.0 - but if you're fine with doing two back to back releases I think that'd be fine

bmw:
We decided to do this in a later release.

bmw:
After doing this, we should continue to install the `letsencrypt` package as well to support people reaching into the venv.

bmw:
Fixed in #3026.

PR #2885, opened 2016-04-22
certbot-auto

### LINKED ISSUE #1337: ./letsencrypt-auto --help should not require root
This overlaps with item #1286 but comes at things from a different angle.

According to https://letsencrypt.github.io/acme-spec/, the raison d'être for ACME is to provide a good user interface, to provide good usability.

From a UI point of view, it is not reasonable to require root privileges for ./letsencrypt-auto --help

It is irrelevant and pointless to say that because of the way it's written, it requires root privileges.  The relevant point is, it shouldn't be written that way.  A good UI should be written in such a way that it can provide an informative message no matter what.  One way (not the only way) of doing this would be to incorporate a helpful message within the letsencrypt-auto shell script ... and to check for the --help option before checking for root privileges.

bmw:
Fixes #1337, #1903, #2821, and #2848.

`certbot-auto` and `letsencrypt-auto` are the same file by request so that we only have to sign one script during releases. The script changes its behavior based upon the value of `$0`. If `basename $0` is `letsencrypt-auto`, the old behavior is preserved.

I only modified the bootstrap scripts that are officially supported (the ones that are not behind a call to `ExperimentalBootstrap`). It is difficult to test many of the 3rd party bootstrap scripts so I left their behavior the same instead of potentially breaking them. Perhaps we can add support for `--yes` into these scripts at a later date. Since the only way to access them is by using a `--debug` flag, there is less of a promise of the stability.

I unfortunately did not have time to write automated tests for these changes. I was also struggling to run test farm tests, however, I tested every change using docker.

I also did not modify the `letsencrypt-auto-source` directory layout (#2877), point the script to the `certbot` PyPI packages (#2839), or modify the embedded GitHub URL (#2884). I think these changes should occur later, in a subsequent PR.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -650,9 +650,6 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         help="(letsencrypt-auto only) prevent the letsencrypt-auto script from"
              " upgrading itself to newer released versions")
     helpful.add(
-        "automation", "--yes", action="store_true",
-        help="(letsencrypt-auto only) assume yes is the answer to all prompts")
-    helpful.add(
         "automation", "-q", "--quiet", dest="quiet", action="store_true",
         help="Silence all output except errors. Useful for automation via cron."
              " Implies --non-interactive.")
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -650,6 +650,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         help="(letsencrypt-auto only) prevent the letsencrypt-auto script from"
              " upgrading itself to newer released versions")
     helpful.add(
+        "automation", "--yes", action="store_true",
+        help="(letsencrypt-auto only) assume yes is the answer to all prompts")
+    helpful.add(
         "automation", "-q", "--quiet", dest="quiet", action="store_true",
         help="Silence all output except errors. Useful for automation via cron."
              " Implies --non-interactive.")
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -307,7 +307,12 @@ BootstrapRpmCommon() {
 BootstrapSuseCommon() {
   # SLE12 don't have python-virtualenv
 
-  $SUDO zypper -nq in -l \
+  if [ "$ASSUME_YES" = 1 ]; then
+    zypper_flags="-nq"
+    install_flags="-l"
+  fi
+
+  $SUDO zypper $zypper_flags in $install_flags \
     python \
     python-devel \
     python-virtualenv \
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -183,26 +183,34 @@ BootstrapDebCommon() {
       # ARGS:
       BACKPORT_NAME="$1"
       BACKPORT_SOURCELINE="$2"
+      echo "To use the Apache Certbot plugin, augeas needs to be installed from $BACKPORT_NAME."
       if ! grep -v -e ' *#' /etc/apt/sources.list | grep -q "$BACKPORT_NAME" ; then
           # This can theoretically error if sources.list.d is empty, but in that case we don't care.
           if ! grep -v -e ' *#' /etc/apt/sources.list.d/* 2>/dev/null | grep -q "$BACKPORT_NAME"; then
-              /bin/echo -n "Installing augeas from $BACKPORT_NAME in 3 seconds..."
-              sleep 1s
-              /bin/echo -ne "\e[0K\rInstalling augeas from $BACKPORT_NAME in 2 seconds..."
-              sleep 1s
-              /bin/echo -e "\e[0K\rInstalling augeas from $BACKPORT_NAME in 1 second ..."
-              sleep 1s
               if echo $BACKPORT_NAME | grep -q wheezy ; then
-                  /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
+                  echo 'Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports"'
               fi
-
-              $SUDO sh -c "echo $BACKPORT_SOURCELINE >> /etc/apt/sources.list.d/$BACKPORT_NAME.list"
-              $SUDO apt-get update
+              if [ "$ASSUME_YES" = 1 ]; then
+                  add_backports=1
+              else
+                  read -p "Would you like to enable the $BACKPORT_NAME repository [Y/n]? " response
+                  case $response in
+                      [yY][eE][sS]|[yY]|"")
+                          add_backports=1;;
+                      *)
+                          add_backports=0;;
+                  esac
+              fi
+              if [ "$add_backports" = 1 ]; then
+                  $SUDO sh -c "echo $BACKPORT_SOURCELINE >> /etc/apt/sources.list.d/$BACKPORT_NAME.list"
+                  $SUDO apt-get update
+              fi
           fi
       fi
-      $SUDO apt-get install $YES_FLAG --no-install-recommends -t "$BACKPORT_NAME" $augeas_pkg
-      augeas_pkg=
-
+      if [ "$add_backports" != 0 ]; then
+          $SUDO apt-get install $YES_FLAG --no-install-recommends -t "$BACKPORT_NAME" $augeas_pkg
+          augeas_pkg=
+      fi
   }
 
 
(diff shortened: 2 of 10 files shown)


SwartzCr:
This code looks good to me, but I don't seem to see a certbot-auto file anywhere nor an indication of how to call the shared file as certbot-auto to allow for the --help.
When I rename it to certbot-auto and call it with --help it install's system packages, which if I'm to understand the $BASENAME code correctly it shouldn't do since $0 should be certbot-auto so help should = 1 wand not get overwritten by the line 102 logic

```root@cb-test:~/letsencrypt# ./certbot-auto --help
Bootstrapping dependencies for Debian-based OSes...
Hit http://security.ubuntu.com trusty-security InRelease
Hit http://security.ubuntu.com trusty-security/main Sources```

**Comment by bmw** [comment]
> This code looks good to me, but I don't seem to see a certbot-auto file anywhere nor an indication of how to call the shared file as certbot-auto to allow for the --help.

I didn't make this change until #2877 is resolved.

> When I rename it to certbot-auto and call it with --help it install's system packages, which if I'm to understand the $BASENAME code correctly it shouldn't do since $0 should be certbot-auto so help should = 1 wand not get overwritten by the line 102 logic
root@cb-test:~/letsencrypt# ./certbot-auto --help Bootstrapping dependencies for Debian-based OSes... Hit http://security.ubuntu.com trusty-security InRelease Hit http://security.ubuntu.com trusty-security/main Sources

Which `letsencrypt-auto` did you use? The `letsencrypt-auto` in the root of the tree or `letsencrypt-auto-source/letsencrypt-auto`? The copy in the root of the tree is to remain the same as the last release tag to help people using an ancient version of `letsencrypt-auto` who must `git pull` manually to receive updates.

When testing, I get the following behavior:

```
bmw@ubuntu:~/Code/letsencrypt$ git rev-parse HEAD
23baf225a4ee9fb9382381290781f9d9c124ac20
bmw@ubuntu:~/Code/letsencrypt$ cp letsencrypt-auto-source/letsencrypt-auto certbot-auto
bmw@ubuntu:~/Code/letsencrypt$ ./certbot-auto --help
Usage: certbot-auto [OPTIONS]
A self-updating wrapper script for the Certbot ACME client. When run, updates
to both this script and certbot will be downloaded and installed. After
ensuring you have the latest versions installed, certbot will be invoked with
all arguments you have provided.

Help for certbot itself cannot be provided until it is installed.

  --debug               attempt installation on experimental platforms
  --help                print this help
  --no-self-upgrade     do not download updates for certbot or certbot-auto
  --os-packages-only    install OS dependencies and exit
  -v, --verbose         provide more output
  --yes                 assume yes is the answer to all prompts

All arguments are accepted and forwarded to the Certbot client when run.
```

With this said, I'm not sure this should be merged yet. I removed the 3 second countdown for installing packages from backports which should be unchanged when the script is named `letsencrypt-auto`.

**Comment by bmw** [comment]
Closing and reopening so Travis will start...

File: certbot/cli.py [hunk 1]
```diff
@@ -650,6 +650,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         help="(letsencrypt-auto only) prevent the letsencrypt-auto script from"
              " upgrading itself to newer released versions")
     helpful.add(
+        "automation", "--yes", action="store_true",
+        help="(letsencrypt-auto only) assume yes is the answer to all prompts")
```
pde: "all prompts about OS dependency and package installation"?

certbot/cli.py
```diff
@@ -650,6 +650,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         help="(letsencrypt-auto only) prevent the letsencrypt-auto script from"
              " upgrading itself to newer released versions")
     helpful.add(
+        "automation", "--yes", action="store_true",
```
pde: Should we also support `-y`?

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -20,10 +20,24 @@ VENV_NAME="letsencrypt"
...
+all arguments you have provided.
+
+Help for certbot itself cannot be provided until it is installed.
+
+  --debug               attempt installation on experimental platforms
+  --help                print this help
+  --no-self-upgrade     do not download updates for certbot or certbot-auto
+  --os-packages-only    install OS dependencies and exit
+  -v, --verbose         provide more output
+  --yes                 assume yes is the answer to all prompts
```
pde: tweak as above

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -151,30 +175,48 @@ BootstrapDebCommon() {
   augeas_pkg="libaugeas0 augeas-lenses"
   AUGVERSION=`apt-cache show --no-all-versions libaugeas0 | grep ^Version: | cut -d" " -f2`
 
+  if [ "$ASSUME_YES" = 1 ]; then
+    YES_FLAG="-y"
```
pde: Consistency with `apt` and `dnf` is another reason to accept `-y`.

pde: (aside from this being something that might be commonly wanted)

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -34,6 +48,10 @@ for arg in "$@" ; do
       # Do not upgrade this script (also prevents client upgrades, because each
       # copy of the script pins a hash of the python client)
       NO_SELF_UPGRADE=1;;
+    --help)
+      HELP=1;;
+    --yes)
+      ASSUME_YES=1;;
```
pde: `--noninteractive`, `--non-interactive`, and `-n` should imply `--yes`.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -34,30 +34,48 @@ BootstrapDebCommon() {
...
-              /bin/echo -ne "\e[0K\rInstalling augeas from $BACKPORT_NAME in 2 seconds..."
-              sleep 1s
-              /bin/echo -e "\e[0K\rInstalling augeas from $BACKPORT_NAME in 1 second ..."
-              sleep 1s
               if echo $BACKPORT_NAME | grep -q wheezy ; then
-                  /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
+                  echo 'Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports"'
+              fi
+              if [ "$ASSUME_YES" = 1 ]; then
+                  /bin/echo -n "Installing augeas from $BACKPORT_NAME in 3 seconds..."
```
pde: We can probably remove this countdown stuff, since it will only fire if the user supplied `-y` or `-n`?

bmw: It will fire if the user used `--yes` or they're using `letsencrypt-auto`. If you read my original post on this PR, I left it out at first, but added it back in to prevent making `letsencrypt-auto` more aggressive.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -34,30 +34,48 @@ BootstrapDebCommon() {
...
           if ! grep -v -e ' *#' /etc/apt/sources.list.d/* 2>/dev/null | grep -q "$BACKPORT_NAME"; then
-              /bin/echo -n "Installing augeas from $BACKPORT_NAME in 3 seconds..."
-              sleep 1s
-              /bin/echo -ne "\e[0K\rInstalling augeas from $BACKPORT_NAME in 2 seconds..."
-              sleep 1s
-              /bin/echo -e "\e[0K\rInstalling augeas from $BACKPORT_NAME in 1 second ..."
-              sleep 1s
               if echo $BACKPORT_NAME | grep -q wheezy ; then
-                  /bin/echo '(Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports")'
+                  echo 'Backports are only installed if explicitly requested via "apt-get install -t wheezy-backports"'
```
pde: Note that previously this sentence came after the countdown.  Now it might make more sense to say something like this:
`certbot-auto would like to enable backports for you; note that packages from that repository are only installed if explicitly request it`?

bmw:
I ran test farm tests on the most recent commit using this test script:
```
#!/bin/bash -xe

cd letsencrypt/letsencrypt-auto-source
cp -p letsencrypt-auto certbot-auto

if ! ./certbot-auto -nhv | grep "experimental installation" ; then
    echo "Not very helpful"
    exit 1
fi

if ! ./certbot-auto --no-self-upgrade --help | grep "experimental installation" ; then
    echo "Not very helpful"
    exit 1
fi

if ! ./certbot-auto -vn --debug --version 2>&1 | grep 0.5.0 ; then
    echo "Couldn't run LE"
    exit 1
fi

rm -rf ~/.local/share/letsencrypt

if ! ./letsencrypt-auto --debug --help | grep "Choice of server" ; then
    echo "Couldn't run LE"
    exit 1
fi

echo "We work!"
```
and all 13 tests passed.

pde:
LGTM

PR #2895, opened 2016-04-26
Joyent SmartOS Zone Bootstrap Support [revised]

dietsche:
Initial support for bootstrapping on Joyent SmartOS Zones.

Testing with image: 088b97b0-e1a1-11e5-b895-9baa2086eb33 base-64-lts 15.4.1 smartos zone-dataset 2016-03-04

Also, a small bug fix for Amazon EC2 detection. /etc/issue doesn't exist on some machines. Prior to this patch, smartos zones would print an error message from the grep command.


pde:
Please make these changes to the template files and bootstrap scripts under `letsencrypt-auto-source/`, and then run `build.py` in that subdirectory.

The copy of `letsencrypt-auto` in the root of the tree should only be updated when a new release is made.

dietsche:
@pde thanks for the feedback. I've updated my patch set. Should I also commit the output from build.py?

pde:
Merging this without having tested it on that OS, since it's appropriately marked as experimental.

PR #2901, opened 2016-04-28
Update using.rst

1resu:
jessie packages in backports


pde:
Thanks!  Looks like someone else landed a similar change slightly ahead of this PR...

Issue #1042, opened 2015-10-20
support multiple vhosts per file


abstratt:
My Apache configuration declares multiple vhosts in one site configuration file. letsencrypt [currently does not handle that](https://community.letsencrypt.org/t/error-should-only-be-one-vhost-in-etc-apache2/).

```
Error: should only be one vhost in /etc/apache2/sites-available/000-default.conf
```

pde:
I think this is a dupe of #809 ?

bmw:
@pde, this is not a dupe.

This refers to the fact that the Apache plugin currently doesn't allow multiple vhosts per file. #809 refers to a specific problem with multiple vhosts when using `install`. The example I gave in #809 has each vhost in a separate file.

sjau:
This is not the problem I have.

As you can see, the vhost of apache has 4 entries. One entry each for ipv4/6 and for non-/ssl. The first time I let letsencrypt create a certificate for such domain it works.

When I then remove all letsencrypt data files and stuff and re-try to create new cert. Then it fails.

SwartzCr:
I'm going to take a look at this tomorrow, @jdkasten do you have any advice for me?

pde:
(if we aren't getting this done by launch, we could kick this ticket out of the milestone with a PR to make the error message more friendly).

bmw:
I just made the simple change to improve this error message while we're working on this. If we decide we need to kick this out of the 1.0 milestone, we can do so without further changes to the code.

pde:
I think yes, let's kick it out of the milestone.  We have a bunch of compelling apache tickets, it's not really clear this is the most important of them.

hrstoyanov:
This is a showstopper for folks who do the right thing - place each domain into a separate conf file. Will come back to letsencrypt when this is solved, but no go for me at this point, sigh ...

bmw:
More information (as well as a sample vhost) provided in #2574.

samwilson:
+1

antoinemichea:
+1

jayshields:
+1

acdha:
If you've previously obtained a cert and followed the last couple decades of Apache practice by having a simple redirect block in the same file, both `certonly` or `--standalone` also fail in addition to the default process. Even if there's a harder general problem handling complex files, it should be possible to write an renewed certificate either way.

```
<VirtualHost *:80>
    ServerName example.org
    RedirectPermanent / https://example.org/
</VirtualHost>
[...]

PR #2906, opened 2016-05-02
Don't fail authentication when vhost cannot be found

### LINKED ISSUE #677: Support "new config file" when no vhost / servername found
Using the Apache configurator on our acme-client test machine, I got a cert for one of my hostnames that didn't already exist in any of the Apache configs. At one point I got a screen that said "We were unable to find a vhost with a ServerName or Address of abcde.crud.net. Which virtual host would you like to choose?"

None of the virtual hosts was really suitable, since I hadn't yet set one up for that hostname on that machine. Ideally we should offer a "new vhost" option that would create a file in the appropriate directory. I realize this is probably quite fraught: How do we know if the new file will be included into the main config; what do we set its documentroot to.

As a stopgap, to make the messaging clearer, maybe we should add language like: "If you haven't yet set up Apache for <host> on this machine, we suggest doing that now and running letsencrypt again."

mgedmin:
It's rather annoying when you cannot renew a certificate LetsEncrypt generated
for you, because the newer version is pickier about parsing Apache config files
and fails to understand typical setups like
```apache
<VirtualHost *:80>
    ServerName example.com
    Redirect permanent / https://example.com
</VirtualHost>
<VirtualHost *:433>
    ServerName example.com
    SSLEngine on
    SSLCertificateFile /etc/letsencrypt/live/example.com/fullchain.pem
    SSLCertificateKeyFile /etc/letsencrypt/live/example.com/privkey.pem
    ...
</VirtualHost>
```
because there are multiple VirtualHost sections in one config file (#1042).

This PR attempts to work around the problem by allowing the Apache
authenticator to proceed without understanding the existing virtual hosts, by
falling back to a sensible default (listen on *:443) when it cannot figure out
what ports were used for the virtual hosts you're trying to authenticate.

Should fix #677 and #2600.

~~I don't expect this PR to be merged as-is, since it lacks tests, but I hope to
push the two bugs mentioned above closer to resolution.  (I can try to add the
tests, with some help perhaps, if the approach seems acceptable to the
LetsEncrypt maintainers.)~~  There are even tests now ;)

### DIFF
--- a/certbot-apache/certbot_apache/tests/tls_sni_01_test.py
+++ b/certbot-apache/certbot_apache/tests/tls_sni_01_test.py
@@ -4,6 +4,7 @@ import shutil
 
 import mock
 
+from certbot import errors
 from certbot.plugins import common_test
 
 from certbot_apache import obj
@@ -131,6 +132,16 @@ class TlsSniPerformTest(util.ApacheTest):
                 set([obj.Addr.fromstring("_default_:443")]),
                 False, False)
         )
+
+        # pylint: disable=protected-access
+        self.assertEqual(
+            set([obj.Addr.fromstring("*:443")]),
+            self.sni._get_addrs(self.achalls[0]))
+
+    def test_get_addrs_no_vhost_found(self):
+        self.sni.configurator.choose_vhost = mock.Mock(
+            side_effect=errors.MissingCommandlineFlag(
+                "Failed to run Apache plugin non-interactively"))
 
         # pylint: disable=protected-access
         self.assertEqual(
--- a/certbot-apache/certbot_apache/tls_sni_01.py
+++ b/certbot-apache/certbot_apache/tls_sni_01.py
@@ -4,6 +4,7 @@ import os
 import logging
 
 from certbot.plugins import common
+from certbot.errors import PluginError
 
 from certbot_apache import obj
 from certbot_apache import parser
@@ -116,11 +117,20 @@ class ApacheTlsSni01(common.TLSSNI01):
 
     def _get_addrs(self, achall):
         """Return the Apache addresses needed for TLS-SNI-01."""
-        vhost = self.configurator.choose_vhost(achall.domain, temp=True)
         # TODO: Checkout _default_ rules.
         addrs = set()
         default_addr = obj.Addr(("*", str(
             self.configurator.config.tls_sni_01_port)))
+
+        try:
+            vhost = self.configurator.choose_vhost(achall.domain, temp=True)
+        except PluginError:
+            # We couldn't find the virtualhost for this domain, possibly
+            # because it's a new vhost that's not configured yet (GH #677),
+            # or perhaps because there were multiple <VirtualHost> sections
+            # in the config file (GH #1042).  See also GH #2600.
+            addrs.add(default_addr)
+            return addrs
 
         for addr in vhost.addrs:
             if "_default_" == addr.get_addr():
--- a/certbot-apache/certbot_apache/tls_sni_01.py
+++ b/certbot-apache/certbot_apache/tls_sni_01.py
@@ -4,7 +4,7 @@ import os
 import logging
 
 from certbot.plugins import common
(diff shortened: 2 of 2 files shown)


mgedmin:
I've tested a backported version of this patch in production, against letsencrypt-apache 0.5.0, and can confirm that it fixed issue #2600 for me:

```diff
--- /opt/letsencrypt/lib/python2.7/site-packages/letsencrypt_apache/tls_sni_01.py.orig	2016-05-02 09:37:32.116940649 +0300
+++ /opt/letsencrypt/lib/python2.7/site-packages/letsencrypt_apache/tls_sni_01.py	2016-05-02 09:43:04.247984037 +0300
@@ -4,6 +4,7 @@
 import logging
 
 from letsencrypt.plugins import common
+from letsencrypt.errors import PluginError, MissingCommandlineFlag
 
 from letsencrypt_apache import obj
 from letsencrypt_apache import parser
@@ -116,12 +117,18 @@
 
     def _get_addrs(self, achall):
         """Return the Apache addresses needed for TLS-SNI-01."""
-        vhost = self.configurator.choose_vhost(achall.domain, temp=True)
         # TODO: Checkout _default_ rules.
         addrs = set()
         default_addr = obj.Addr(("*", str(
             self.configurator.config.tls_sni_01_port)))
 
+        try:
+            vhost = self.configurator.choose_vhost(achall.domain, temp=True)
+        except (PluginError, MissingCommandlineFlag):
+            # https://github.com/letsencrypt/letsencrypt/pull/2906
+            addrs.add(default_addr)
+            return addrs
+    
         for addr in vhost.addrs:
             if "_default_" == addr.get_addr():
                 addrs.add(default_addr)

```

pde:
@joohoi would you like to review this?

joohoi:
Yes, I actually have a few thoughts that I'd need input on from @pdee @bmw 

I think this is a good PR, but maybe should be behind an CLI argument. While I know that adding new arguments to already large list isn't something we should default to, I think this should be optional. We actually need to decide which route to take with configuration files that we aren't able to parse. 

With this PR we lose the ability to notify the user, which would make her/him able to adjust the configuration to suite longer term configuration management by our client, and make the installation part less confusing for the unintiated ones.

mgedmin:
I'm sympathetic to that argument, but I think that understanding Apache config files should be confined to the *installer* part of the Apache plugin.  After all, the other authenticator plugins (webroot, standalone) don't go out of their way to verify that you actually have a web server serving the domains you're authenticating.

pde:
If this workaround can be strictly scoped to the authentication step, I think that makes it safer wrt @joohoi's concerns.  For the IInstaller, the options might include:

* actually implement enough augeas logic to work with the multiple vhosts in a given file
* offer to split the vhosts out into separate files for the admin, so that we can handle them more elegantly

mgedmin:
It *is* strictly scoped to the authentication step.  `tls_sni_01.py`, the only non-test module changed by this PR, deals solely with authentication.

joohoi:
The functionality is, but I think we will need to add a pre-flight check if apache installer is used. This is needed to prevent situation where authentication succeeds, but installing fails and leaves uninitiated user with something else than expected, but certs counted towards limits.

pde:
As I noted in IRC, I believe the fact that "reinstall" is now the default choice in a menu if you got a cert during your first attempt, means that naive users won't normally exhaust their rate limits just because they're fighting to get the installer to work.

reinstall

pde:
(So I support @joohoi's suggestion of a pre-check, but I don't think this PR necessarily needs to block on it)

joohoi:
@pde good point! With this information, the pre-check isn't needed at this point, but let's keep it in mind if the need arises.

This PR is good to go, LGTM.

frozen-solid:
Still having a problem with this bug as of today, though I'm wondering if my issue might be related to having a hyphen in the domain name. I have 4 hosts all setup with exactly the same configs. The only difference is the directory, document root, and server names. However, the domain with a hyphen keeps throwing the error still.

PR #2909, opened 2016-05-02
Randomize serial numbers of DVSNI challenge certificates. [needs revision]

chrismarget:
Background here:
https://community.letsencrypt.org/t/serial-numbers-of-dvsni-challenge-certificates-are-a-problem/14317/5

### DIFF
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -203,7 +203,7 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-    cert.set_serial_number(1337)
+    cert.set_serial_number(OpenSSL.rand.bytes(16))
     cert.set_version(2)
 
     extensions = [


kuba:
How about adding a test that checks if two calls return certs with different serials?

chrismarget:
On Mon, May 2, 2016 at 5:32 PM, Jakub Warmuz <notifications@github.com>
wrote:

> How about adding a test that checks if two calls return certs with
> different serials?
>
Beyond my ability. If I could have, I'd have incremented a counter, rather
than looking for collisions in random numbers.

The proposed change introduces the possibilities of collisions in a 128 bit
space, but that certainly beats the *guaranteed collision* that results
from assigning every certificate serial number 1337 :)

acme/acme/crypto_util.py
```diff
@@ -203,7 +203,7 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-    cert.set_serial_number(1337)
+    cert.set_serial_number(OpenSSL.rand.bytes(16))
```
schoen: Unfortunately, this produces

`TypeError: serial must be an integer`

`OpenSSL.rand.bytes()` returns a `str` (probably `bytes` in Python 3), but `set_serial_number` wants an `int`. 

How about

`cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))`

?

schoen:
@chrismarget, I think this would work properly if you make the change I suggested. Then we could try to deal with @kuba's test-related suggestion.

PR #2922, opened 2016-05-05
venv is still named letsencrypt

### LINKED ISSUE #2911: Auto venv details
In #2885 I did not change the name of the venv created by `*-auto`. The reason for this is I believe we expressed some concern about people reaching into the venv to run `letsencrypt` directly, e.g. `~/.local/share/letsencrypt/bin/letsencrypt`.

Do we want to rename the venv? If not, we need to revert the change made on this [line](https://github.com/letsencrypt/letsencrypt/blob/master/certbot/cli.py#L40).

bmw:
Fixes #2911.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -37,7 +37,7 @@ helpful_parser = None
 # should only be used for purposes where inability to detect letsencrypt-auto
 # fails safely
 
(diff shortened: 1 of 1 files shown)


SwartzCr:
lgtm

PR #2923, opened 2016-05-05
Release script prep

bmw:
This release script can be used for the next two releases. After that, we should be able to stop building and releasing shim packages.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -188,9 +188,10 @@ while ! openssl dgst -sha256 -verify $RELEASE_OPENSSL_PUBKEY -signature \
 done
 
 # copy leauto to the root, overwriting the previous release version
+cp -p letsencrypt-auto-source/letsencrypt-auto certbot-auto
 cp -p letsencrypt-auto-source/letsencrypt-auto letsencrypt-auto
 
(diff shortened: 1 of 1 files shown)


tools/release.sh
```diff
@@ -45,7 +45,7 @@ export GPG_TTY=$(tty)
 PORT=${PORT:-1234}
 
 # subpackages to be released
-SUBPKGS=${SUBPKGS:-"acme certbot-apache certbot-nginx letshelp-certbot"}
+SUBPKGS=${SUBPKGS:-"acme certbot-apache certbot-nginx letshelp-certbot letsencrypt letsencrypt-apache letsencrypt-nginx letshelp-letsencrypt"}
```
pde: Do we need to keep `letshelp-letsencrypt`?

pde:
Otherwise LGTM.

tools/release.sh
bmw: I think so, at least for the next release. The package is currently useless, however, we have been packaging it every release and we want to make sure people are able to upgrade if the package ever becomes useful.

With that said, after the next release, perhaps we should stop packaging `certbot-nginx` and `letshelp-certbot` until the packages are ready.

PR #2925, opened 2016-05-05
Adding --dialog argument

### LINKED ISSUE #2351: Way to specify dialog/gui interface
With #2350, there isn't a way to run `letsencrypr renew` with dialog. To change this, we need a `--dialog` or `--gui` flag.

TheNavigat:
Closes #2351


certbot/cli.py
```diff
@@ -570,6 +570,8 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
               "additional command line flags; the client will try to explain "
               "which ones are required if it finds one missing")
     helpful.add(
+        None, "--gui", dest="gui", action="store_true", help="Run using GUI")
```
bmw: This should probably be `--dialog` not `--gui`. `dialog` is questionably a GUI and this allows us to add another UI later if we want to.

certbot/main.py
```diff
@@ -678,7 +678,8 @@ def main(cli_args=sys.argv[1:]):
     elif config.text_mode:
         displayer = display_util.FileDisplay(sys.stdout)
     elif config.verb == "renew":
-        config.noninteractive_mode = True
+        if not config.gui:
```
bmw: This shouldn't just apply to `renew`. While redundant, people should be able `--dialog` for other subcommands as well.

certbot/cli.py
bmw: nit: Perhaps we should also set `dest=dialog_mode` to match the other display options.

bmw:
Perhaps we should check for conflicting values and thrown an error. Something like:

We could start the block of `if` statements about the Display with something like:
```
if config.dialog_mode:
    if config.quiet or config.noninteractive_mode or config.text_mode:
        raise errors.Error("Conflicting values for displayer")
```

It may also be nice to say specifically which two (or more) flags conflict.

certbot/main.py
```diff
@@ -677,8 +687,9 @@ def main(cli_args=sys.argv[1:]):
         displayer = display_util.NoninteractiveDisplay(sys.stdout)
     elif config.text_mode:
         displayer = display_util.FileDisplay(sys.stdout)
+    elif config.dialog_mode:
+        displayer = display_util.NcursesDisplay()
     elif config.verb == "renew":
-        config.noninteractive_mode = True
```
bmw: This line shouldn't be deleted.

certbot/tests/main_test.py
```diff
@@ -1,48 +1,68 @@
...
+import six
 
 from certbot import cli
 from certbot import configuration
+from certbot import errors
+from certbot import main
 from certbot.plugins import disco as plugins_disco
 
 
 class ObtainCertTest(unittest.TestCase):
```
bmw: > 18:24 < TheNavigat> The fact that someone decided that the default method in 
                    main_test's _call should be obtain_cert is sad

This is because these tests are for `ObtainCert`.

You should make a new `MainTest` class instead of modifying this one.

TheNavigat: Hmmmm... I actually didn't notice that.

Alright, working on it.

TheNavigat:
So, the lint test is failing because some of the code in `main_test.py` and `cli_test.py` is identical. I'm not sure how should that be fixed.

certbot/main.py
```diff
@@ -677,6 +687,8 @@ def main(cli_args=sys.argv[1:]):
         displayer = display_util.NoninteractiveDisplay(sys.stdout)
     elif config.text_mode:
         displayer = display_util.FileDisplay(sys.stdout)
+    elif config.dialog_mode:
```
bmw: Merge `master` into this PR. The logic for choosing the display for renew moved [here](https://github.com/certbot/certbot/blob/master/certbot/cli.py#L337). This PR shouldn't have any changes on these lines and instead make the linked to code something like:
```
if self.verb == "renew" and not self.dialog_mode:
```

certbot/main.py
```diff
@@ -669,6 +669,16 @@ def main(cli_args=sys.argv[1:]):
...
 
+    # Avoid conflicting args
+    conficting_args = ["quiet", "noninteractive_mode", "text_mode"]
+    if config.dialog_mode:
+        for arg in conficting_args:
+            if getattr(config, arg):
+                raise errors.Error(
+                    ("Conflicting values for displayer."
+                    " {0} conflicts with dialog_mode").format(arg)
+                )
```
bmw: Sorry I didn't catch this last night, but this code is really just part of command line parsing. We should probably move it to `parse_args` (or a function/method called by `parse_args`) in `cli.py` instead of `main.py`. This allows you to move your test back in to `cli.py`, mitigating your code duplication problem.

bmw:
LGTM. Will kick Travis and merge when Travis is green.

bmw:
LGTM

PR #2930, opened 2016-05-05
Changing some "will happen"s to "hopefully will happen"s

jgillula:
Made some changes to the documentation to avoid promising things that we're gonna try to do, but can't absolutely guarantee, for purposes of legal review.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,8 +128,8 @@ System Requirements
 ===================
 
 The Let's Encrypt Client presently only runs on Unix-ish OSes that include
-Python 2.6 or 2.7; Python 3.x support will be added after the Public Beta
-launch. The client requires root access in order to write to
+Python 2.6 or 2.7; Python 3.x support will hopefully be added after the Public
+Beta launch. The client requires root access in order to write to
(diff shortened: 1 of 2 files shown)


SwartzCr:
We actually have Jessie backports already
So it should say something like
Packages exist for Debian Jessie via backports. First you'll have to follow the instructions here[backports.debian.org/instructions] to enable the Jessie backports repo, if you have not already done so. Then run:
`apt-get install certbot python-certbot-apache -t jessie-backports`

README.rst
```diff
@@ -128,8 +128,8 @@ System Requirements
 ===================
 
 The Let's Encrypt Client presently only runs on Unix-ish OSes that include
-Python 2.6 or 2.7; Python 3.x support will be added after the Public Beta
-launch. The client requires root access in order to write to
+Python 2.6 or 2.7; Python 3.x support will hopefully be added after the Public
+Beta launch. The client requires root access in order to write to
```
schoen: It's a while after the Public Beta launch already, so how about making this say "hopefully be added in the future"?

SwartzCr:
lgtm

schoen:
I agree.

Issue #2896, opened 2016-04-26
./tools/venv.sh and ./tools/venv3.sh fail on Ubuntu 16.04, Python 2.7.11


TheNavigat:
Python version is 2.7.11

Here's everything printed.

```
+ command -v python2
/usr/bin/python2
+ export VENV_ARGS=--python python2
+ ./tools/_venv_common.sh -e acme[dev] -e .[dev,docs] -e certbot-apache -e certbot-nginx -e letshelp-certbot -e certbot-compatibility-test
+ VENV_NAME=venv
+ rm -rf *.egg-info
+ date +%s
+ mv venv venv.1461692822.bak
+ virtualenv --no-site-packages venv --python python2
Running virtualenv with interpreter /usr/bin/python2
New python executable in /root/letsencrypt/venv/bin/python2
Also creating executable in /root/letsencrypt/venv/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
+ . ./venv/bin/activate
+ deactivate nondestructive
+ unset -f pydoc
+ [ -z  ]
+ [ -z  ]
+ [ -n  ]
+ [ -n  ]
+ [ -z  ]
+ unset VIRTUAL_ENV
+ [ ! nondestructive = nondestructive ]
+ VIRTUAL_ENV=/root/letsencrypt/venv
+ export VIRTUAL_ENV
+ _OLD_VIRTUAL_PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
+ PATH=/root/letsencrypt/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
+ export PATH
+ [ -z  ]
+ [ -z  ]
+ _OLD_VIRTUAL_PS1=# 
+ [ x != x ]
+ basename /root/letsencrypt/venv
+ PS1=(venv) # 
+ export PS1
+ alias pydoc
+ [ -n  ]
+ [ -n  ]
+ pip install -U setuptools
Requirement already up-to-date: setuptools in ./venv/lib/python2.7/site-packages
+ pip install -U pip
Requirement already up-to-date: pip in ./venv/lib/python2.7/site-packages
+ pip install -e acme[dev] -e .[dev,docs] -e certbot-apache -e certbot-nginx -e letshelp-certbot -e certbot-compatibility-test
[...]

PR #2936, opened 2016-05-06
Use --force-reinstall to fix bad virtualenv package

bmw:
Mitigates #2896 until the package is fixed upstream.


bmw:
Self-merging because change is not user facing.

PR #2940, opened 2016-05-07
s/--letsencrypt/--certbot

bmw:
Should fix integration test failures.


bmw:
Self merging as this only affects tests.

Issue #2890, opened 2016-04-24
Problematic symlink creation in mac.sh bootstrapper


MichalMMac:
I've just started experimenting with letsencrypt on OS X. I am also using MacPorts. When trying to run `letsencrypt-auto` I got following import error `raise ImportError("Unable to import lib%s!" % args[0])`. Similiar problem is described in this community [thread](https://community.letsencrypt.org/t/unable-to-import-libaugeas-on-osx-el-capitan/5794/4).

In file `letsencrypt-auto-source/pieces/bootstrappers/mac.sh` there is code responsible for creating `libaugeas.dylib` symlink:

```
# Workaround for _dlopen not finding augeas on OS X
  if [ "$pkgman" = "port" ] && ! [ -e "/usr/local/lib/libaugeas.dylib" ] && [ -e "/opt/local/lib/libaugeas.dylib" ]; then
    echo "Applying augeas workaround"
    $SUDO ln -s /opt/local/lib/libaugeas.dylib /usr/local/lib
  fi
```

`ln` command is making assumption that `/usr/local/lib` directory exists. If it does **not** exist bogus symlink `/usr/local/lib -> /opt/local/lib/libaugeas.dylib` is created.

1. Script should make sure `/usr/local/lib` directory exists (if not create it)
2. $SUDO ln -s /opt/local/lib/libaugeas.dylib /usr/local/lib/

Possible fix:

```
then
    echo "Applying augeas workaround"
    $SUDO mkdir -p /usr/local/lib/
    $SUDO ln -s /opt/local/lib/libaugeas.dylib /usr/local/lib/
  fi
```

I would submit PR but since I am new to this project It is probably better for someone with more insigh to look at it.

MichalMMac:
Perhaps @nneonneo or @erikrose could look at this.

nneonneo:
That's a good fix. I'm a bit concerned that we could fail to make `/usr/local/lib` on certain El Capitan installs, but that's a matter for a different PR. 

Make a PR and submit it: I would sign off on it.

MichalMMac:
Ok PR #2943 created.

@nneonneo By El Capitan installs you mean possible problem with System Integrity Protection? /usrl/local content is not under SIP. If for some reason /usr/local itself does not exists that would be a problem but you would have to disable SIP and delete it somehow in first place.

nneonneo:
Aha, understood. I knew that /usr/local was excluded from SIP but didn't
[...]

PR #2943, opened 2016-05-08
Ensure /usr/local/lib/ exists in mac.sh bootstraper

MichalMMac:
Solves problem outlined in #2890
/usr/local/lib directory does not exist on fresh install of OS X.
We need to create it with root proviledges.


nneonneo:
Per my earlier comment on #2890, +1 from me. This is something that I would've included in my original patch had I been aware of the problem.

pde:
LGTM. Sorry this didn't make 0.6.0!

PR #2945, opened 2016-05-09
Update git instructors from letsencrypt to certbot

dandv:


pde:
These are going to be changed to slightly different instructions, which use `wget` rather than `git clone`.  Stand by...

Issue #2840, opened 2016-04-14
change download instructions


SwartzCr:
When we move over the client to EFF we should no longer instruct people to git clone the repo, instead they should be wgetting a URL from eff.org and checking a signature. The instructions should reflect this.
Peter will work with tech-ops to make sure we have this set up.

PR #2950, opened 2016-05-09
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### GitHub Corrected URLs 
Was | Now 
--- | --- 
https://github.com/letsencrypt/letsencrypt | https://github.com/certbot/certbot


pde:
We'll have an update to the instructions shortly; #2840

PR #2954, opened 2016-05-10
New download instructions

### LINKED ISSUE #2840: change download instructions

pde:
Will close #2840.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,7 @@ from your OS and puts others in a python virtual environment::
 .. hint:: The certbot-auto download is protected by HTTPS, which is pretty good, but if you'd like to
           double check the integrity of the ``certbot-auto`` script, you can use these steps for verification before running it::
 
-            user@server:~$ wget https://dl.eff.org/certbot-auto.asc
+            user@server:~$ wget -N https://dl.eff.org/certbot-auto.asc
             user@server:~$ gpg2 --recv-key A2CFB51FA275A7286234E7B24D17C995CD9775F2
             user@server:~$ gpg2 --trusted-key 4D17C995CD9775F2 --verify certbot-auto.asc certbot-auto
 
--- a/README.rst
+++ b/README.rst
@@ -30,9 +30,15 @@ systems have packages yet, we provide a temporary solution via the
 ``letsencrypt-auto`` wrapper script, which obtains some dependencies
 from your OS and puts others in a python virtual environment::
 
-  user@webserver:~$ git clone https://github.com/letsencrypt/letsencrypt
-  user@webserver:~$ cd letsencrypt
-  user@webserver:~/letsencrypt$ ./letsencrypt-auto --help
+  user@webserver:~$ wget https://dl.eff.org/certbot-auto
+  user@webserver:~$ chmod a+x ./certbot-auto
+  user@webserver:~$ ./certbot-auto --help
+
+.. hint:: For stronger security, you can use these steps for extra verification before running the script:
+
+  user@webserver:~$ wget https://dl.eff.org/certbot-auto.sig
+  user@webserver:~$ gpg2 --recv-key A2CFB51FA275A7286234E7B24D17C995CD9775F2
+  user@webserver:~$ gpg2 --trusted-key 4D17C995CD9775F2 --verify certbot-auto.sig certbot-auto
 
 Or for full command line help, type::
 
--- a/README.rst
+++ b/README.rst
@@ -37,9 +37,9 @@ from your OS and puts others in a python virtual environment::
 .. hint:: The certbot-auto download is protected by HTTPS, which is pretty good, but if you'd like to
           double check the integrity of the ``certbot-auto`` script, you can use these steps for verification before running it::
 
-            user@server:~$ wget https://dl.eff.org/certbot-auto.sig
+            user@server:~$ wget https://dl.eff.org/certbot-auto.asc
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -30,11 +30,18 @@ systems have packages yet, we provide a temporary solution via the
...
 
-  user@webserver:~$ git clone https://github.com/letsencrypt/letsencrypt
-  user@webserver:~$ cd letsencrypt
-  user@webserver:~/letsencrypt$ ./letsencrypt-auto --help
+  user@webserver:~$ wget https://dl.eff.org/certbot-auto
+  user@webserver:~$ chmod a+x ./certbot-auto
+  user@webserver:~$ ./certbot-auto --help
 
-Or for full command line help, type::
+.. hint:: The certbot-auto download is protected by HTTPS, which is pretty good, but if you'd like to
```
bmw: This hint block doesn't seem to render properly, at least when clicking "View" above.

README.rst
```diff
@@ -30,11 +30,18 @@ systems have packages yet, we provide a temporary solution via the
...
+          double check the integrity of the ``certbot-auto`` script, you can use these steps for verification before running it::
+
+            user@server:~$ wget https://dl.eff.org/certbot-auto.sig
```
bmw: nit: Should we change "@server" to "@webserver" to keep the hostname consistent with the one used above?

README.rst
```diff
@@ -30,11 +30,18 @@ systems have packages yet, we provide a temporary solution via the
...
+            user@server:~$ gpg2 --recv-key A2CFB51FA275A7286234E7B24D17C995CD9775F2
+            user@server:~$ gpg2 --trusted-key 4D17C995CD9775F2 --verify certbot-auto.sig certbot-auto
```
bmw: When running this on a Wheezy docker image:
```
gpg2 --trusted-key 4D17C995CD9775F2 --verify letsencrypt-auto-source/letsencrypt-auto.sig letsencrypt-auto
```

I get this error:
```
gpg: no valid OpenPGP data found.
gpg: the signature could not be verified.
Please remember that the signature file (.sig or .asc)
should be the first file given on the command line.
```

If you've tested this and you're sure this is correct or if we're changing our signature procedure for 0.6.0, this is fine with me.

README.rst
pde: It caused the code block to grow horizontal scrollbars for me, so I shortened to "server".

README.rst
pde: You're right, gpg and openssl/smime have different signature formats. We could keep the openssl one the releases currently use, and that's probably the best idea, but it'll prevent people from using the PGP web of trust as a second verification method.  It's probably good enough to give them a URL on some independent domain (letsencrypt.org, perhaps) to fetch the openssl key from.

README.rst
bmw: Perhaps we should use "@server" above then as well for consistency?

README.rst
bmw: I agree continuing to use OpenSSL signatures is probably the best idea, at least in the short term, however, I bet some of our users don't completely trust the CA system. Other options include

1. Switch le-auto to gpg signatures. Most systems le-auto supports come with gpg by default, however, we'd certainly end up having to do extra work on some systems (e.g. OS X).
2. Do both an openssl and gpg signature during the release process :frowning: 
3. Convert the gpg key to an openssl key for the signature check. I'm not sure how user friendly we can make this or how well we can exploit the PGP web of trust for additional security if we do this.

pde: Option 3 is so nuts we should definitely not do it.  But I'm just curious so I'm going to go and see what it would look like...

README.rst
pde: When I build the html docs I get reasonable output for that block.

tools/release.sh
```diff
@@ -187,6 +187,9 @@ while ! openssl dgst -sha256 -verify $RELEASE_OPENSSL_PUBKEY -signature \
    read -p "Please correctly sign letsencrypt-auto with offline-signrequest.sh"
 done
 
+# This signature is not quite as strong, but easier for people to verify out of band
+gpg -u "$RELEASE_GPG_KEY" --detach-sign --armor --sign letsencrypt-auto-source/letsencrypt-auto
+
 # copy leauto to the root, overwriting the previous release version
 cp -p letsencrypt-auto-source/letsencrypt-auto certbot-auto
 cp -p letsencrypt-auto-source/letsencrypt-auto letsencrypt-auto
```
pde: As the tree and this PR stand, we'll be making the following signatures:

`letsencrypt-auto-source/letsencrypt-auto.sig` (openssl)
`letsencrypt-auto-source/letsencrypt-auto.asc` (gpg)

I guess we can't rename the former, because old `letsencrypt-auto`s will fetch it.  But we could probably rename the latter...

pde: Fixed. We now have:

`letsencrypt-auto-source/letsencrypt-auto.sig` (openssl)
`letsencrypt-auto-source/certbot-auto.asc` (gpg)

Though arguably we could put `certbot-auto.asc` in the root?

README.rst
bmw: You're right. It still doesn't display right on GH, but I assume that's just a bug on their end.

bmw:
Only nit is I think we should use a consistent hostname for instructions (i.e. `webserver` or `server`) but not both. Other than that, LGTM.

PR #2960, opened 2016-05-11
Fix mime.types parsing for nginx

zauguin:
Added characters to key parsing rule which appear in the mime type in mime.types.


pde:
Hmmm, I think this is actually safe to merge for 0.6.0 since we don't ship nginx...

PR #2962, opened 2016-05-11
Randomize serial numbers of DVSNI challenge certificates.

chrismarget:
Setting all DVSNI challenge certificates to the same serial number (1337) requires the TLS speaking application to load multiple instances of the _same certificate_ (issuer + s/n should uniquely identify a certificate). [Some TLS speakers](http://www.cisco.com/c/en/us/products/security/asa-firepower-services/index.html) aren't happy about this. Discussion on the client forum [here](https://community.letsencrypt.org/t/serial-numbers-of-dvsni-challenge-certificates-are-a-problem).

### DIFF
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -203,12 +203,10 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-
     try:
         cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))
     except AttributeError:
-        cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16),'big'))
-
+        cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16), 'big'))
     cert.set_version(2)
 
     extensions = [
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -203,7 +203,7 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-    cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))
+    cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16),'big'))
     cert.set_version(2)
 
     extensions = [
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -206,6 +206,7 @@ def gen_ss_cert(key, domains, not_before=None,
     try:
         cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))
     except AttributeError:
+        # pylint: disable=E1101
         cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16), 'big'))
     cert.set_version(2)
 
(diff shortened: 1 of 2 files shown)


chrismarget:
Not Python 3 friendly.

schoen:
I can see that each of these methods works in either Python 2 or Python 3, but this kind of exception-catching alternative doesn't look great to me. It would be nice to find a single way of doing it that works in both Python versions.

One option which isn't that nice is `import binascii` and then `int(binascii.hexlify(OpenSSL.rand.bytes(16)), 16)`.

chrismarget:
Thank you, schoen. Do you prefer the not-that-nice `binascii` approach? I headed down this road briefly, then got hung up on the fact that `OpenSSL.rand.bytes` returns `str` in Python 2 and `bytes` in Python 3.

schoen:
What's more, the Python 2 `str` type is the same as the Python 3 `bytes` type (Python 2 `unicode` corresponds to Python 3 `str`). So it's not necessarily different in the larger scheme of things even though the type is called something else.

I think this is a good solution. I agree with the suggestion that we should add a test to check that the serial numbers of two independently-generated certs via `gen_ss_cert()` are different from one another. I can help do that if you don't know how.

chrismarget:
> Python 2 `str` type is the same as the Python 3 `bytes` type

Yup, I've finally grokked that. I initially abandoned the `binascii` conversion because of a misunderstanding about this, but finally realized that the random function (which returns this type) and the hexlify function (which expects it) _both_ changed with the transition between Python versions.

Checking for collisions certainly seems sane to me, but leaves me wondering: Why even do random serial numbers if we truly care about collisions (I appear to be the only person experiencing a problem with it)? If we detect a collision, we'll need to go about rectifying it... Why not just use an incrementing counter starting with `\x00`? 

> I can help do that if you don't know how.

I'll take you up on that! I would have gone for a counter if I'd known what I was doing, but I couldn't find my way from the `challenges.TLSSNI01` reference in a plugin to the creation of the certificate in `crypto_utils`. I'm sure you've noticed from my trail of breadcrumbs that I'm in a pretty steep part of the learning curve here.

Thank you.

schoen:
Hi @chrismarget, I didn't mean run-time tests, but rather unit tests, which will help make sure that this code does what's intended and that other people's subsequent changes to the project's code don't break its functionality in some way. I believe @kuba was concerned with making sure that there are associated unit tests for this new functionality.

chrismarget:
Hi @schoen. You're talking about @kuba's comment [here](https://github.com/certbot/certbot/pull/2909#issuecomment-216370705)?

> How about adding a test that checks if two calls return certs with different serials?

I'd interpreted that as being concerned with run-time serial number collisions. If we're talking about a *unit test*, then I might be able to figure that out.

Still, I worry about the collision situation. In fact, I worry about it *even more* with a unit test: A run-time collision will break my plugin, but just for that one run, which wasn't guaranteed to succeed anyway (Internet outage, Letsencrypt outage, etc...) It'll probably be fine when the client runs the next time around. A collision during the unit test will break the CI process. Is it reasonable to leave the CI pass/fail up to a (128-bit) coin toss?

pde:
Ooops sorry I shouldn't have assigned schoen here, since the ball's in @chrismarget's court.

1. in terms of a unit test, it seems you just want to make a bunch of these and check they don't collide?

chrismarget:
Okay, [this](http://pastebin.com/raw/JdwJspYE) is where I'm headed. I'm thinking of merging that into `acme/acme/crypto_util_test.py`

Am I on the right track, @schoen ?

Thank you.

chrismarget:
I've written the unit test. It creates 5 certificates and checks to make sure we got >1 unique serial numbers. The rationale here being that travis will "never" randomly land on the same 128-bit number 5 times in a row.

Is this the sort of thing you're looking for, @schoen and @pde? Does it look okay? It seems to have sneaked by Travis without any problem. Is there anything else I should do?

Thanks very much!

schoen:
LGTM, thanks!

Issue #2446, opened 2016-02-11
Add a "register" verb


pde:

bmw:
To provide more information to this issue, @jvanasco added in #2980:
>I'm not sure if this is a dupe of #2446 , which has no description

>The intent of this ticket is that the certbot program would have a switch that will (only) create a new account (with TOS registration, etc).

>There are at least 2 reasons for this:

>a user may want to create an account / accept the TOS on "machine 1", but run certbot on "machine 2"
a user may want to use another client, but that client does not offer the account creation/management - and expects the account to be created on the official client.

>full disclosure: i wrote a client that I'm opensourcing and we're the second case.

PR #2967, opened 2016-05-11
Add a register verb, with a way to update registrations

### LINKED ISSUE #1215: Change the e-mail address associated with an account
#1214 was originally discovered as a result of a user wanting to change the e-mail address associated with an account and not knowing how to do that. I don't think there's currently a documented way to do this, but there should be. This probably also requires changes on the boulder side and maybe an additional API...?

schoen:
Fixes #1215


certbot/cli.py
```diff
@@ -591,6 +592,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
              "certificates. Updates to the Subscriber Agreement will still "
              "affect you, and will be effective 14 days after posting an "
              "update to the web site.")
+    helpful.add(
+        None, "--update-registration", action="store_true",
```
pde: You should make a "register" help topic, and use that for this argument, rather than `None`.  Search for "paths" or "automation" to see how to do that.

certbot/main.py
```diff
@@ -363,6 +364,33 @@ def _init_le_client(config, authenticator, installer):
     return client.Client(config, acc, authenticator, installer, acme=acme)
 
 
+def register(config, unused_plugins):
+    """Create or modify accounts on the server."""
+
+    # Currently, only --update-registration is implemented. Issue #2446
+    # calls for a fuller register verb, to allow better separation of
+    # account management from obtaining certs.
+    if not config.update_registration:
```
pde: Since we have more time now, we may as well get the register verb to actually register :).  I think it should be very easy to do so; just grab the relevant portion from the beginning of `run` or `obtain_cert`.

certbot/cli.py
```diff
@@ -265,8 +265,9 @@ def __init__(self, args, plugins, detect_defaults=False):
         self.VERBS = {"auth": main.obtain_cert, "certonly": main.obtain_cert,
                       "config_changes": main.config_changes, "run": main.run,
                       "install": main.install, "plugins": main.plugins_cmd,
-                      "renew": main.renew, "revoke": main.revoke,
-                      "rollback": main.rollback, "everything": main.run}
+                      "register": main.register, "renew": main.renew,
```
pde: Also add the new subcommand to the usage strings at the top of cli.py?

certbot/main.py
schoen: So, these functions use `determine_account()` to choose what account the user wants to use, and create one if no account currently exists.

I think it would make sense to simply called `determine_account()` here as the default `register` action, but a possible problem is that if accounts already exist, `determine_account()` tries to choose one instead of creating a new one. It's not totally clear to me what the right behavior of the `register` verb ought to be in the case where an account already exists.

* Create a new account no matter what?
* Ask the user whether to create a new account or not?
* Never create a new account?

In any of these cases, is it reasonable for me to modify `determine_account()` so that it has an additional parameter choosing that behavior instead of its default behavior?

schoen: @pde, I'd be curious what behavior you think is appropriate in this case.

schoen:
Fixes #2446

certbot/main.py
```diff
@@ -366,6 +367,50 @@ def _init_le_client(config, authenticator, installer):
...
+                "required\n(hint: --email)")
+    acc, acme = _determine_account(config)
+    acme_client = client.Client(config, acc, None, None, acme=acme)
+    try:
+        updated_reg = client.messages.Registration.from_data(email=config.email)
+        acme_client.acme.update_registration(acme_client.account.regr,
+                                             updated_reg)
+    except acme_errors.UnexpectedUpdate:
+        # We expect the unexpected update!
+        pass
```
bmw: The fact we have to catch and raise an exception here indicates to me that we're not going about this the best way. For example,  [this code](https://github.com/certbot/certbot/blob/v0.6.0/acme/acme/client.py#L163) uses `update_registration` without raising an exception.

I played with it a bit and was able to update a registration using this code:
```
acme_client.acme.update_registration(
    acc.regr.update(body=acc.regr.body.update(contact=('mailto:' + config.email,))))
```
While absolutely horrendous, it works and seems to be the way that `update_registration` was designed to work. Additionally, this function returns the updated registration so we don't need the additional call and trip to Boulder with `query_registration`.

certbot/main.py
```diff
@@ -366,6 +367,50 @@ def _init_le_client(config, authenticator, installer):
...
+    query_data = acme_client.acme.query_registration(acme_client.account.regr)
+    # We rely on an ACME exception to interrupt this process if it didn't work.
+    print("Registration change succeeded. New registration data:\n")
+    print(query_data)
```
bmw: Do we really want to print the entire registration resource here? It's kind of an ugly mess.

Since users can only use this to update their e-mail address, perhaps we can just say that their email address updated to `config.email`.

Also, in most of the client we don't call `print` directly in favor of using `IDisplay`/`IReporter`. Perhaps we should just register a message with `IReporter` about the update?

bmw:
This PR doesn't update the `regr.json` file on disk. While this doesn't seem to matter (right now), perhaps it should be done for the sake of correctness?

While a great start on the issue, I personally think this could use a bit more polishing, however, many people really want the ability to update their e-mail address. If @pde or @jsha want to +1 this PR, I'm happy to merge it and we can clean up things later.

certbot/main.py
```diff
@@ -366,6 +366,45 @@ def _init_le_client(config, authenticator, installer):
...
+                    "unsupported.")
+        # _determine_account will register an account
+        _determine_account(config)
+        return
+
+    # --update-registration
+    if len(accounts) == 0:
+        return "Could not find an existing account to update."
+    if config.email is None:
+        return ("Currently, --update-registration can only change the e-mail "
```
pde: Rather than raising this error, just call `display.ops.get_email()`.

pde:
Looks good aside from that one correction.

certbot/main.py
bmw: Good idea!

bmw:
Fixes #3075.

Phew. The suggested one or two line change to call `display_ops.get_email` turned into quite the task. Here's what's going on here:

While I didn't have a lot of time to work with the boulder team on this, it appears that boulder currently won't let you remove your e-mail address from your account (see letsencrypt/boulder#1846). I believe this will only be temporary, but Certbot has to work around it for the time being.

With this in mind, telling people to run the client again with `--register-unsafely-without-email` when prompting for an e-mail doesn't make sense so I had to refactor `get_email`. In doing this, I created an `optional` flag, which I believe can be removed completely once letsencrypt/boulder#1846 is resolved.

In refactoring `get_email`, I found #3075. While a very silly bug, I fixed it in writing this PR.

@schoen, @pde, @SwartzCr, can one of you do a final review this?

bmw:
Kicking this to 0.8.0 since it never got LGTM'd in time for the release.

schoen:
@bmw, would you be able to resolve the conflicts that crept in on this branch? I'm not sure what the right resolutions are.

bmw:
Sure. Just pushed a commit solving all conflicts.

schoen:
Thanks!

schoen:
@bmw, that sure did complicate things to deal with all of those registration cases...

LGTM!

PR #2968, opened 2016-05-11
Randomize DVSNI challenge certificate serial number, now for python 3.3.

chrismarget:
Today's earlier attempt wasn't python 3.3 friendly. Is a new pull request the right way to handle such a revision? I've no idea. I'm sorry for the clutter, especially if it's the wrong way to go.

Setting all DVSNI challenge certificates to the same serial number (1337) requires the TLS speaking application to load multiple instances of the same certificate (issuer + s/n should uniquely identify a certificate). Some TLS speakers aren't happy about this. Discussion on the client forum [here](https://community.letsencrypt.org/t/serial-numbers-of-dvsni-challenge-certificates-are-a-problem/).

### DIFF
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -203,7 +203,7 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-    cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))
+    cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16),'big'))
     cert.set_version(2)
 
     extensions = [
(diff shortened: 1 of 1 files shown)


acme/acme/crypto_util.py
```diff
@@ -203,7 +203,7 @@ def gen_ss_cert(key, domains, not_before=None,
     """
     assert domains, "Must provide one or more hostnames for the cert."
     cert = OpenSSL.crypto.X509()
-    cert.set_serial_number(1337)
+    cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16),'big'))
```
schoen: Unfortunately, Python 2 doesn't have `int.from_bytes`, so this can't be used in Python 2-compatible code.

schoen:
I think it would be preferable to revise the prior PR instead of making a new one. I also pointed out that, unfortunately, `int.from_bytes` doesn't work in Python 2.

acme/acme/crypto_util.py
chrismarget: I'm learning slowly - thank you for your patience.

Currently leaning towards:

`
    try:
        cert.set_serial_number(int(OpenSSL.rand.bytes(16).encode("hex"), 16))
    except AttributeError:
        cert.set_serial_number(int.from_bytes(OpenSSL.rand.bytes(16),'big'))
`

Does that look reasonable?

PR #2976, opened 2016-05-12
FIX certbot update path :  letsencrypt doesn't exist anymore on github !

mrgngautier:
https://raw.githubusercontent.com/letsencrypt/letsencrypt/ was in 404 so I update it !


bmw:
This was actually due to the 0.6.0 tag not being pushed yet. It should be fixed now.

PR #2994, opened 2016-05-15
system python path has changed on el capitan

tapple:
on both my mac machines with fresh el capitan installed, "which python" gives /usr/bin/python


pde:
LGTM

PR #2998, opened 2016-05-15
Use kill instead of killpg. [needs revision]

jsha:
Killpg kills an entire process group and produces unexpected results when run as
part of integration tests running in a noninteractive shell. Specifically, it
kills the entire test rather than just the httpd process.

This was causing problems running Boulder integration tests inside of Docker
with an unprivileged user, because `su buser -c ...` creates a noninteractive
shell. So the integration test would always exit with status 143 after printing
"Terminating manual command process."

### DIFF
--- a/certbot/plugins/manual.py
+++ b/certbot/plugins/manual.py
@@ -196,7 +196,7 @@ s.serve_forever()" """
                 "cleanup() must be called after perform()")
             if self._httpd.poll() is None:
                 logger.debug("Terminating manual command process")
-                os.killpg(self._httpd.pid, signal.SIGTERM)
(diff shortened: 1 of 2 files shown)


bmw:
Also not sure if you're still interested in this, but this PR seems to cause other problems in [Travis](https://travis-ci.org/certbot/certbot/jobs/130422559#L2707).

Alternatively, we could also try using `self._httpd.terminate()` which is provided by [subprocess](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.terminate).

pde:
Deprecated by #3750?

bmw:
Closing this in favor of #3750.

You beat me.

PR #2999, opened 2016-05-15
Make Docker changes to allow use in tests.

jsha:
For Boulder we're starting to move towards using Docker images to run our tests.
The certbot Docker image almost has what we need, but we need the test directory
too.


pde:
@prefer since you're sending us Docker PRs, are you also interested in reviewing them?

bmw:
@jsha are you still interested in seeing this merged?

If so, I think `Dockerfile-dev` is probably a better place for this since `Dockerfile` is meant to be downloaded by end users wanting to use `certbot certonly`. The less we put in that image, the better.

jsha:
I'll close this for now. We currently use the packaged certbot (actually letsencrypt 0.5.0) in our Boulder tests. We may want to revisit if we want to use a Certbot docker image instead.

PR #3005, opened 2016-05-16
Don't call os.pid() the default kwargs to atexit_print_messages

lamby:
Whilst it has no side-effects, this results in non-reproducible output when
generating the certbot documentation:

    http://i.imgur.com/Q6VGi9S.jpg

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>

### DIFF
--- a/certbot/reporter.py
+++ b/certbot/reporter.py
@@ -16,6 +16,11 @@ from certbot import le_util
 
 logger = logging.getLogger(__name__)
 
+# Store the pid of the process that first imported this module so that
+# atexit_print_messages side-effects such as error reporting can be limited to
+# this process and not any fork()'d children.
+INITIAL_PID = os.getpid()
+
 
 @zope.interface.implementer(interfaces.IReporter)
 class Reporter(object):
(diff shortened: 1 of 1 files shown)


bmw:
So this actually doesn't work, at least not in the way it was intended to.

This code was intended to fix the problem if the client forks. While it doesn't fork anymore (but might in the future), the problem we were having is we registered `atexit_print_messages` to run when the client exits. If the client forks, however, you'd get the messages printed multiple times because the messages would be printed for each exiting process.

Default arguments are evaluated when the function is defined. This allowed us to set the default value of `pid` to the pid of the process that first imported the module. This way, even though all exiting processes called `atexit_print_messages`, only the call that had the matching pid would actually print any messages.

The code proposed in this PR removes this protection though and messages will be printed for every exiting process. There are other solutions though. We could create a global variable or have `atexit_print_messages` register the call with `atexit` itself. In the latter approach, it could set internal state and use it to determine which process causes the messages to printed.

Hopefully this makes sense! Let me know if you have any questions.

lamby:
> this allowed us to set the default value of pid to the pid of the process that first imported the module. 

Ahhhhh. :)  Apologies for not spotting this but, on the other hand, it's the sort of thing that would definitely deserve a large/scary comment, if only to prevent someone helpfully following the instructions of a linter to remove dangerous-looking default arguments.

@bmw I've updated this PR to match and documented this behaviour explicitly.

bmw:
Yeah. I agree there should have been a comment so thanks for adding one.

LGTM. Thanks @lamby.

PR #3010, opened 2016-05-16
Fix *-auto command line parsing

### LINKED ISSUE #2987: *-auto arg parsing problems
If you run `certbot-auto --os-packages-only`, `getopts` parses the `n`, thinks you mean `-n` and forces yes in your package manager.

To do what we want in a portable way, I think we'll have to do argument parsing by hand.

bmw:
Fixes #2987. This has been excessively tested with the test farm, `dash`, and `tcsh` (FreeBSD).

Changes to OS X bootstrap come from changes to *-auto files where `build.py` wasn't run before the merge.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -38,17 +38,6 @@ Help for certbot itself cannot be provided until it is installed.
 
 All arguments are accepted and forwarded to the Certbot client when run."
 
-while getopts ":hnv" arg; do
-    case $arg in
-        h)
-            HELP=1;;
-        n)
-            ASSUME_YES=1;;
-        v)
-            VERBOSE=1;;
-    esac
-done
-
 for arg in "$@" ; do
   case "$arg" in
     --debug)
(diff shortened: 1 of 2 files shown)


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -65,6 +54,17 @@ for arg in "$@" ; do
...
+        case "$short_arg" in
+          h)
+            HELP=1;;
+          n)
+            ASSUME_YES=1;;
+          v)
+            VERBOSE=1;;
+        esac
+      done;;
   esac
```
pde: do we want to add a `-- break` case?

pde:
LGTM.  Do you want to do `--quiet` in another PR?

letsencrypt-auto-source/letsencrypt-auto.template
bmw: What's the benefit of doing that?

bmw:
Yes, I plan to add `--quiet` in another PR built off this one.

PR #3011, opened 2016-05-17
Allow modifying webroot-map during renewal

### LINKED ISSUE #2977: --dry-run on renew didn't work
Hello,

I just try to test the renew like this:
./letsencrypt-auto renew -v --dry-run
and the log says:

```
2016-05-12 21:03:20,463:DEBUG:certbot.main:Root logging level set at 20
2016-05-12 21:03:20,464:INFO:certbot.main:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2016-05-12 21:03:20,464:DEBUG:certbot.main:certbot version: 0.6.0
2016-05-12 21:03:20,464:DEBUG:certbot.main:Arguments: ['-v', '--dry-run']
2016-05-12 21:03:20,464:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2016-05-12 21:03:20,465:DEBUG:certbot.main:Exiting abnormally:
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/certbot/main.py", line 693, in main
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/certbot/main.py", line 538, in renew
    renewal.renew_all_lineages(config)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/certbot/renewal.py", line 305, in renew_all_lineages
    raise errors.Error("Currently, the renew verb is only capable of "
Error: Currently, the renew verb is only capable of renewing all installed certificates that are due to be renewed; individual domains cannot be specified with this action. If you would like to renew specific certificates, use the certonly command. The renew verb may provide other options for selecting certificates to renew in the future.
```

I use a cli.ini with 
authenticator = webroot
webroot-map = {....some domains which are correctly installed with certonly command }

Any idea?

Thanks, Alex

bmw:
Fixes #2977.

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -301,7 +301,8 @@ def _renew_describe_results(config, renew_successes, renew_failures,
 def renew_all_lineages(config):
     """Examine each lineage; renew if due and report results"""
 
-    if config.domains != []:
+    if (config.domains != [] and
+            set(config.domains) != six.viewkeys(config.webroot_map)):
         raise errors.Error("Currently, the renew verb is only capable of "
                            "renewing all installed certificates that are due "
                            "to be renewed; individual domains cannot be "
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -301,6 +301,8 @@ def _renew_describe_results(config, renew_successes, renew_failures,
 def renew_all_lineages(config):
     """Examine each lineage; renew if due and report results"""
 
+    # If more plugins start using cli.add_domains,
+    # we may want to only log a warning here
     if (config.domains != [] and
             set(config.domains) != six.viewkeys(config.webroot_map)):
         raise errors.Error("Currently, the renew verb is only capable of "
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -301,10 +301,10 @@ def _renew_describe_results(config, renew_successes, renew_failures,
 def renew_all_lineages(config):
     """Examine each lineage; renew if due and report results"""
 
-    # If more plugins start using cli.add_domains,
-    # we may want to only log a warning here
-    if (config.domains != [] and
-            set(config.domains) != six.viewkeys(config.webroot_map)):
+    # This is trivially False if config.domains is empty
+    if any(domain not in config.webroot_map for domain in config.domains):
+        # If more plugins start using cli.add_domains,
+        # we may want to only log a warning here
         raise errors.Error("Currently, the renew verb is only capable of "
                            "renewing all installed certificates that are due "
                            "to be renewed; individual domains cannot be "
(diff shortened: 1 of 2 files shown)


certbot/renewal.py
```diff
@@ -301,7 +301,10 @@ def _renew_describe_results(config, renew_successes, renew_failures,
 def renew_all_lineages(config):
     """Examine each lineage; renew if due and report results"""
 
-    if config.domains != []:
+    # This is trivially False if config.domains is empty
+    if any(domain not in config.webroot_map for domain in config.domains):
```
pde: This is probably correct, though we should convince ourselves that there are no other ways that `config.domains` might get set, aside from the webroot plugin.  For instance, if there's a cli.ini file with `domains` set to we want to be erroring out here?

pde:
If we're sure that that's the best approach for now (as opposed to allowing any `domains` setting method as long as it matches) then LGTM.

pde:
Ooops sorry caught the wrong username in a dropdown menu :)

certbot/renewal.py
bmw: I think so. This is what we have been doing in the past and in general we cannot differentiate between values set on the command line and those in a CLI config file.

bmw:
> as opposed to allowing any domains setting method as long as it matches

Do you mean using domains as a selector for choosing which lineages to renew? I personally don't think we should do this right now. Cert/lineage management has proven to be quite difficult and I think we should think carefully about how we want to implement it (James' cert manager?) before doing anything like this. Otherwise, we we'll have multiple ways to specify the same thing, which may conflict with each other in the future.

pde:
@bmw no I just mean being tolerant if someone has `domains = blah` in their `cli.ini` file, and the domains in that file match the certs being renewed.

pde:
So basically, don't error if `domains` exactly matches the set of domains being renewed.

Intuitively I think that would be polite behaviour, but it might be a little tricky to implement.

For several reasons I've been thinking that the `renew.renew_all_lineages` might need to be refactored to walk all the renewal conf files before commencing any actual renewal work; this might be another reason to do that.

bmw:
Ah I understand. I think adding something like that would be fine, however, it'll require significant refactoring and should be thoroughly tested given the importance of `renew`. Do you think something like this needs to be in the 0.7.0 release?

I personally would prefer not to block this PR on this feature though. Perhaps we should just open a new issue?

pde:
I got a report of that sort of breakage (perhaps from Seth?) yesterday afternoon, actually.  But yes let's make it a different ticket.

PR #3014, opened 2016-05-17
Detect RewriteEngine  directives that originates in a VirtualHost directive  

### LINKED ISSUE #3003: Missing 'RewriteEngine on' from vhost auto-configuration in http2https redirection
When redirect all http connections to https setting enabled in vhost autoconfiguration. Installer adds `RewritCond` and `RewriteRule`, but redirection doesn't work, because of missing `RewriteEngine on` -line.

Problem solved by adding this line manually. So please add it to the script. It was all good in former letsencrypt-script.

sagi:
Fixes #3003. 

- An Augeas path of a RewriteEngine directive that originates in an .htaccess file:
`[u'/files/var/www/dumpbits/public_html/.htaccess/directive[1]/arg']`

- An Augeas path of a RewriteEngine directive that originates in a VirtualHost:
`[u'/files/etc/apache2/sites-available/dumpbits.conf/VirtualHost/directive[5]/arg']`

The fix detects only RewriteEngine directives that originates in a VirtualHost.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1177,7 +1177,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         :type vhost: :class:`~certbot_apache.obj.VirtualHost`
 
         """
-        rewrite_engine_path_list= self.parser.find_dir("RewriteEngine", "on",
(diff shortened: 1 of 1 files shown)


SwartzCr:
looks good to me

PR #3015, opened 2016-05-18
Make letsencrypt-auto respect --quiet

bmw:
Fixes #2990.

PR #3016, opened 2016-05-18
Make certbot-auto quieter without implementing --quiet

### LINKED ISSUE #2990: certbot-auto doesn't respect --quiet
It prints, amongst other things:
```
Requesting root privileges to run certbot...
   sudo CERTBOT_AUTO=./certbot-auto /home/ubuntu/.local/share/letsencrypt/bin/letsencrypt --quiet renew
```

bmw:
Fixes #2990. I think this PR should be preferred to one that adds `certbot-auto` support for `--quiet`.

Initially, I planned to provide `--quiet` support for the whole script, including bootstrapping and setting up the `virtualenv`. The problem with this is these steps don't happen very often and the output is often something the user should care about.

With this in mind, I set out to only implement `--quiet` support for the non-upgrading case. I did this in #3015 and then realized there are only two places where we output anything.

1. We output "Checking for new version..." on every single run. This seems very excessive. I believe we are very clear that `certbot-auto` is a self updating script and this output is just clutter. I removed it entirely.
2. We output "Requesting root privileges to run certbot..." on every run, even if we already have root. After fixing the bug of always showing this prompt, I hid it when using `--quiet`, however, if we're about use su/sudo to run `certbot`, we will likely be prompting you to enter a password. Asking for a `sudo` password with no context is a bad idea and isn't likely to be useful in the renewal case either. Therefore, to fix this problem, I only print these messages if we're about to use su/sudo.

And...that fixes the problem! The fact we were able to do this without further complicated our temporary `certbot-auto` script is a win in my opinion and there is little to nothing to be gained by implementing `--quiet` on top of this PR.

Please let me know if you disagree!

EDIT: I didn't list it explicitly, but we also output if we're going to use `su` instead of `sudo`. Again, if we're using `su`, we're likely going to prompt so I didn't see this as a problem.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -899,13 +899,12 @@ UNLIKELY_EOF
     echo "Installation succeeded."
   fi
   echo "Requesting root privileges to run certbot..."
+  echo "  $VENV_BIN/letsencrypt" "$@"
   if [ -z "$SUDO_ENV" ] ; then
     # SUDO is su wrapper / noop
-    echo "  " $SUDO "$VENV_BIN/letsencrypt" "$@"
     $SUDO "$VENV_BIN/letsencrypt" "$@"
   else
     # sudo
-    echo "  " $SUDO "$SUDO_ENV" "$VENV_BIN/letsencrypt" "$@"
     $SUDO "$SUDO_ENV" "$VENV_BIN/letsencrypt" "$@"
   fi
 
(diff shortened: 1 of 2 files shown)


pde:
LGTM.

solarissmoke:
@bmw I'm not sure if we're using this wrong, but we currently run renewal tasks through cron, with the following set in the **root** crontab (Ubuntu 14.04):

`53 06 * * * /path/to/certbot-auto renew --quiet`

This still prints the "Requesting root privileges to run certbot..." message every time it runs. Is this intended behaviour?

From our perspective we would like the script to be completely quiet unless there is an error.

bmw:
@solarissmoke the code from this PR hasn't be included in a release yet. It'll be in the 0.7.0 release later this week.

PR #3018, opened 2016-05-18
Remove dangling footnote

mgedmin:
This footnote has no references!

I checked by searching for `#venv`, since [the ReST docs](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#footnotes) tell me a reference ought to look like `[#venv]_`.


pde:
We can either delete this, or move it into README.rst, where there is at least a sensible place to reference it?

bmw:
I think at the very least we should remove the dangling reference. If we'd like to add it to `README.rst` we can do so in a follow up PR.

Issue #3020, opened 2016-05-18
--post-hook validation too strict


mgedmin:
Here's the command I intend to run from cron.weekly:
```sh
#!/bin/sh

[ -x /opt/letsencrypt/bin/certbot ] || exit 0

/opt/letsencrypt/bin/certbot renew --quiet --post-hook '

if [ -x /usr/sbin/apache2 ]; then
    service apache2 reload
fi

if [ -d /etc/pov/post-cert-renewal.d ]; then
    run-parts /etc/pov/post-cert-renewal.d
fi
'
```

This fails with

```
Unable to find post-hook command 

if in the PATH.
(PATH is /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
```

mgedmin:
Workaround for now:
```sh
# 'true #' on 1st line is a workaround for https://github.com/certbot/certbot/issues/3020
/opt/letsencrypt/bin/certbot renew --quiet --post-hook 'true #

if ...
...
'
```

pde:
I don't think there's going to be any sensible way to validate the correctness of a hook in all of these different cases, and I think that the validation step is necessary [because of PATH-related problems](https://github.com/certbot/certbot/issues/2676#issuecomment-198087749) with hooks and `sudo`.

So one option is that we document the "begin with `true`" hack either in the CLI docs or the error message.  The other (perhaps slightly nicer) option is that we add a `--disable-hook-validation` flag.

mgedmin:
Would it be possible to do the $PATH check only if the script invocation fails?

mgedmin:
Ah, I see from https://github.com/certbot/certbot/issues/2676#issuecomment-198090013 that you want to notify the admin about possible errors even if no certificates were renewed during this run.  That sounds sensible.

In that case I'm in favour for a `--disable-hook-validation` flag.

pde:
@mgedmin want to send a PR for that?

mgedmin:
@pde: #3037.

PR #3021, opened 2016-05-18
Hook validation: skip leading spaces/newlines

mgedmin:
Improves the situation with #3020 a bit.

Does nothing about other valid shell commands that the current validation would reject:

  - shell builtins like `--post-hook 'if [ -x /my/script ]; then /my/script; fi'`
  - variable assignments like `--post-hook 'ENV_VAR=value command'`
  - comments
  - redirections like `--post-hook '<infile command'`
  - I'm sure I'm missing many other exciting possibilities

Still, I believe this is a step in the right direction and therefore ask you to please merge it (assuming I haven't broken any tests).


pde:
Thanks, LGTM!

PR #3027, opened 2016-05-18
Fixes Certbot for OS X

TheNavigat:
This commit removes the condition that checks for Python's path. The hard-coded path is incorrect since in OS X 10.11 Python is installed in /usr/bin/python. Removing the condition makes sense because any of the software we use to install Python should return a "Already installed" message if Python is installed.


TheNavigat:
Already fixed by https://github.com/certbot/certbot/commit/4cb35eaeb346e11963c59fd4aa5170a7e01f7190. Closing.

PR #3029, opened 2016-05-19
Remove reference to letsencrypt-auto in error message [revision requested]

voltagex:
Ran into this on FreeBSD. Quick fix although it'd probably be better to set it up so that it uses the name of the script being run rather than hardcoding a name.


pde:
Thanks for this PR! An even better solution would be to `import cli` and then use the string `cli.cli_command`, which will mirror whatever command the user has on their system (`certbot`, `certbot-auto`, or older `letsencrypt` commands)

bmw:
Since this PR was opened, it has been superseded by other PRs that incorporated pde's suggestions. I'm closing this PR, but thanks for taking the time to write this up regardless.

PR #3032, opened 2016-05-19
Always save server value in renewal config

### LINKED ISSUE #3023: Always save server value in renewal config
Currently, if people use configuration files from 0.5.0+ and then use them with a version before 0.5.0, renewal fails because it expects the server value to be there. We should always save the server value in the renewal config to help mitigate this problem.

bmw:
Fixes #3023.

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -146,9 +146,9 @@ def _relevant(option):
     from certbot import renewal
     from certbot.plugins import disco as plugins_disco
     plugins = list(plugins_disco.PluginsRegistry.find_all())
-    return (option in renewal.STR_CONFIG_ITEMS
-            or option in renewal.INT_CONFIG_ITEMS
-            or any(option.startswith(x + "_") for x in plugins))
+    return (option in renewal.STR_CONFIG_ITEMS or
+            option in renewal.INT_CONFIG_ITEMS or
+            any(option.startswith(x + "_") for x in plugins))
 
 
 def _is_cli_default(option, value):
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -7,7 +7,9 @@ import re
 import configobj
 import parsedatetime
 import pytz
+import six
 
+from certbot import cli
 from certbot import constants
 from certbot import crypto_util
 from certbot import errors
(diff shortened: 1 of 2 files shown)


schoen:
I'm sad about this method of fixing the problem because we deliberately didn't save `server` in order to allow using the new default on renewal if the CA endpoint changes. In this case I'm concerned about what our strategy will be to change the CA endpoint if (say) there's a new version of ACME, if all lineages begin to hard-code the endpoint that was used to obtain their certs. Are we going to have a special case that says

```
if config.server == "https://acme-v01.api.letsencrypt.org/directory":
    # This protocol version has been superseded.
    config.server = "https://acme-v02.api.letsencrypt.org/directory"
```

or something?

certbot/storage.py
```diff
@@ -144,9 +146,51 @@ def _relevant(option):
...
+
+    :param str option: name of the option to be considered
+    :param value: the value of option
+
+    :returns: True if option should be stored, otherwise False
+    :rtype: bool
+
+    """
+    # Try to find reasons to store this item in the renewal config.  It
+    # can be stored if it is relevant and (it is the server option,
```
schoen: Maybe the comment should say why the `server` option is always stored, because this represents a kind of unusual special case.

schoen: Oh, it does already say that, sorry!

schoen:
LGTM.

bmw:
That's a very good point and honestly something I hadn't thought about.

I think we could reliably update the server. While very much pseudocode, we could do something like this:

```
import acme

if config.server == "https://acme-v01.api.letsencrypt.org/directory" and acme.__version__ >= "1.2.3":
    config.server = "https://acme-v02.api.letsencrypt.org/directory"
```

One of the problems we'll have here is maybe people manually set the server value to the v01 endpoint and don't want it changed (although that'd likely break their client).

Additionally, if we do something like this, it may cause similar breakage to what this PR is trying to fix later down the road. If people use their renewal config files with an old version of the client, it will incorrectly try to use the v02 endpoint and fail.

With this in mind, I could be convinced we don't want this, however, I suspect quite a few people will have been using `letsencrypt-auto` on an older version of Ubuntu and will be upgrading to Xenial ([source](https://www.eff.org/files/lepc-oses-jan-2016.png)). If we don't merge this PR, I think we should try to act somewhat quickly to mitigate the problem for these users.

schoen:
Let's ask @pde!

bmw:
(Alternatively, we could continue to store the v01 version, but use the proper URL based on the client version. This is a horrendous hack though).

bmw:
After talking to @pde, we decided we probably shouldn't do this.

PR #3033, opened 2016-05-19
Fix warning about already enabled redirect in Apache

### LINKED ISSUE #2736: Invalid warning about Apache redirect
With a fresh install of Apache on Debian Jessie, I created this vhost:
```
<VirtualHost *:80>
        ServerName example.com
        ServerAlias www.example.com

        DocumentRoot /var/www/html

        # error, crit, alert, emerg.
        # modules, e.g.
        #LogLevel info ssl:warn

</VirtualHost>
```
This is the only vhost I had enabled. When running `letsencrypt run --redirect` I get the error message:
```
WARNING:letsencrypt.client:Enhancement redirect was already set.
```
Despite this error message, the redirect was correctly enabled. Are we trying to set up the redirect twice?

bmw:
Fixes #2736. Sagi and I worked on this together yesterday.

@SwartzCr would you be able to take a look at this? Basically, the problem was the core client asks for security enhancements per domain but Apache applies them for each vhost. Therefore, if you have two domains in one vhost and ask for a redirect, you'll get a warning on the second one saying you already have a redirect in the vhost.

We fixed this problem by keeping track of the enhancements that were enabled during the current run.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -124,6 +124,8 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         self.assoc = dict()
         # Outstanding challenges
         self._chall_out = set()
+        # Maps enhancements to vhosts we've enabled the enhancement for
+        self._enhanced_vhosts = defaultdict(set)
 
         # These will be set in the prepare function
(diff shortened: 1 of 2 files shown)


SwartzCr:
lgtm

PR #3034, opened 2016-05-19
Typo: too many self's

chbrown:
The extra self will push along the arguments, resulting in the accurate but not very helpful error message: "AttributeError: 'JWKRSA' object has no attribute 'kty'"


pde:
LGTM

Issue #2978, opened 2016-05-13
OS X Tests


bmw:
Travis supports [OS X Build Environments](https://docs.travis-ci.com/user/osx-ci-environment/). We should set one of these up for fixing/preventing regressions on OS X.

bmw:
@TheNavigat might be interested in this

TheNavigat:
I think we need to fix https://github.com/certbot/certbot/issues/1535 first.

TheNavigat:
Fixed. Working on this issue.

yan12125:
There's a blocker: https://github.com/travis-ci/cpython-builder/pull/3

As Python 3 support is almost there, I think it's good to test both Python 2 and 3 on macOS. However, the aforementioned issue should be resolved first, or the common ```python: "x.y"``` pattern in ```.travis.yml``` won't work.

bmw:
I agree testing with multiple Python versions on macOS is valuable, but could we test using a single Python version on macOS until that issue is resolved?

yan12125:
It's possible to use the system Python on Travis CI. More work should be done so that all tests pass on macOS.

For those who are interested, here's my previous try: https://travis-ci.org/yan12125/certbot/builds/227099560

bmw:
Fixed in #4697.

PR #3035, opened 2016-05-20
Adding OS X 10.11 test to Travis [revision requested]

TheNavigat:
The tests currently don't work because the exception raised here https://github.com/certbot/certbot/blob/master/acme/acme/crypto_util.py#L134 and the code written here https://github.com/certbot/certbot/blob/master/acme/acme/standalone_test.py#L146 isn't executed.

Adding tests that provoke an error never work for some reason. I'm testing on OS X, and after adding the following after the `test_it` function in `standalone_test.py`, the tests kept running forever.

``` python
    def test_failure(self):
        self.assertRaises(errors.Error, crypto_util.probe_sni, b'localhost', b'0.0.0.0', 1235)
```

I have absolutely no idea why does that happen. Even setting the port to the default 1234 doesn't work, and the tests keep running forever.

Furthermore, the tests fail on order versions of OS X because of the outdated `setuptools` package. That's why I'm forced the use of OS X 10.11 for now at least.

PR addresses issue #2978

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,7 +99,7 @@ matrix:
       env: TOXENV=cover
       before_install:
         - "travis_retry brew update"
-        - "travis_retry brew install python augeas dialog"
+        - "travis_retry brew install augeas dialog"
 
 # Only build pushes to the master branch, PRs, and branches beginning with
 # `test-`. This reduces the number of simultaneous Travis runs, which speeds
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,10 +92,14 @@ matrix:
     - python: "2.7"
       env: TOXENV=nginxroundtrip
     - os: osx
+      # Cryptography no longer supports OpenSSL 0.9.8, which
+      # is the versionprovided by the default OS X image
       osx_image: xcode7.3
       language: generic
       env: TOXENV=cover
-      before_install: "travis_retry brew install python augeas dialog"
+      before_install:
+        - "travis_retry brew update"
+        - "travis_retry brew install python augeas dialog"
 
 # Only build pushes to the master branch, PRs, and branches beginning with
 # `test-`. This reduces the number of simultaneous Travis runs, which speeds
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,12 +93,9 @@ matrix:
       env: TOXENV=nginxroundtrip
     - os: osx
       osx_image: xcode7.3
-      sudo: required
       language: generic
       env: TOXENV=cover
       before_install: "travis_retry brew install python augeas dialog"
-      install: "travis_retry pip install tox coveralls"
-      script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/travis-integration.sh)'
 
 # Only build pushes to the master branch, PRs, and branches beginning with
 # `test-`. This reduces the number of simultaneous Travis runs, which speeds
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,14 @@ matrix:
       env: TOXENV=py36
     - python: "2.7"
       env: TOXENV=nginxroundtrip
-
+    - os: osx
+      osx_image: xcode7.3
+      sudo: required
+      language: generic
+      env: TOXENV=cover
+      before_install: "travis_retry brew install python augeas dialog"
+      install: "travis_retry pip install tox coveralls"
+      script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/travis-integration.sh)'
 
 # Only build pushes to the master branch, PRs, and branches beginning with
 # `test-`. This reduces the number of simultaneous Travis runs, which speeds
--- a/acme/acme/standalone_test.py
+++ b/acme/acme/standalone_test.py
@@ -137,7 +137,6 @@ class TestSimpleTLSSNI01Server(unittest.TestCase):
         )
         self.old_cwd = os.getcwd()
         os.chdir(self.test_cwd)
-        self.thread.start()
 
     def tearDown(self):
         os.chdir(self.old_cwd)
@@ -146,6 +145,8 @@ class TestSimpleTLSSNI01Server(unittest.TestCase):
 
     def test_it(self):
         max_attempts = 5
+        started = False
+
         while max_attempts:
             max_attempts -= 1
             try:
@@ -153,7 +154,12 @@ class TestSimpleTLSSNI01Server(unittest.TestCase):
                     b'localhost', b'0.0.0.0', self.port)
             except errors.Error:
                 self.assertTrue(max_attempts > 0, "Timeout!")
-                time.sleep(1)  # wait until thread starts
+                if started:
+                    # Wait until thread starts
+                    time.sleep(1)
+                else:
+                    self.thread.start()
+                    started = True
             else:
                 self.assertEqual(jose.ComparableX509(cert),
                                  test_util.load_comparable_cert(
--- a/acme/acme/standalone_test.py
+++ b/acme/acme/standalone_test.py
@@ -154,12 +154,12 @@ class TestSimpleTLSSNI01Server(unittest.TestCase):
                     b'localhost', b'0.0.0.0', self.port)
             except errors.Error:
                 self.assertTrue(max_attempts > 0, "Timeout!")
-                if started:
-                    # Wait until thread starts
-                    time.sleep(1)
-                else:
+                if not started:
                     self.thread.start()
                     started = True
+                # Wait until thread starts. This line should remain outside
+                # the if condition to avoid coverage missing line false positive
+                time.sleep(1)
             else:
                 self.assertEqual(jose.ComparableX509(cert),
                                  test_util.load_comparable_cert(
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -589,14 +589,14 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                 if realpath not in vhost_paths.keys():
                     vhs.append(new_vhost)
                     vhost_paths[realpath] = new_vhost.filep
-                elif realpath == new_vhost.filep:
+                elif not os.path.islink(new_vhost.filep):
                     # Prefer "real" vhost paths instead of symlinked ones
                     # ex: sites-enabled/vh.conf -> sites-available/vh.conf
 
                     # remove old (most likely) symlinked one
                     vhs = [v for v in vhs if v.filep != vhost_paths[realpath]]
                     vhs.append(new_vhost)
-                    vhost_paths[realpath] = realpath
+                    vhost_paths[realpath] = new_vhost.filep
 
         return vhs
 


bmw:
> The tests currently don't work because the exception raised here https://github.com/certbot/certbot/blob/master/acme/acme/crypto_util.py#L134 and the code written here https://github.com/certbot/certbot/blob/master/acme/acme/standalone_test.py#L146 isn't executed.

That's because `test_it` is a pretty fragile test. Assuming I'm reading it correctly, we start a server in a thread in `setUp` and attempt to connect to it in `test_it`. The assumption in `test_it` is we fail to connect to it at least once.

To fix this problem, we could move `self.thread.start()` into the `except` block of `test_it`. We can create a bool named `started` before the `while` loop and set it to `False`. Then in the `except` block, if `started` is `False`, we only do `self.thread.start()` and set `started` to `True`, otherwise, we sleep.

> I have absolutely no idea why does that happen. Even setting the port to the default 1234 doesn't work, and the tests keep running forever.

Me either. Can you reproduce locally? I'd be curious to know where the code is hanging.

> Furthermore, the tests fail on order versions of OS X because of the outdated setuptools package. That's why I'm forced the use of OS X 10.11 for now at least.

Which version of `setuptools` is available? We require `setuptools>=1.0` for `certbot`.

acme/acme/standalone_test.py
```diff
@@ -137,12 +136,18 @@ def tearDown(self):
...
         max_attempts = 5
+        started = False
+
         while max_attempts:
             max_attempts -= 1
             try:
                 cert = crypto_util.probe_sni(
                     b'localhost', b'0.0.0.0', self.port)
             except errors.Error:
+                if not started:
```
bmw: nit: I personally this should be rewritten as:

```
self.assertTrue(max_attempts > 0, "Timeout!")
if started:
    time.sleep(1)
else:
    self.thread.start()
    started = True
```

You could put the assertion in the `if` block if you want, but `continue` isn't necessary here.

bmw:
I fixed #3172 in `master` this morning so I kicked Travis to see what was going on. Great to see we're passing ACME tests!

Looks there are some issues with Apache now which might be a good introduction for you to the Apache plugin. Alternatively, we could just use `py27` here for tests to avoid these issues, but using `cover` is certainly a better test to help make sure we're getting good test coverage on OS X as well.

TheNavigat:
Currently apache tests fail because this line always is false, https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L569

This happens because while `new_vhost.filep` points to a path that starts with `/var/`, `realpath` points to a path that starts with `/private/var/`. This happens because in OS X `/var/` is actually a symlink to `/private/var/`.

When I tried to fix this by changing these lines https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L569 so that the `parser_paths` key is the realpath, other tests broke. This line, https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L558, fails when the path is prefixed by `/private/`, even though, as explained earlier, they're actually the exact same path.

I'm stuck with the augeas bit, I'm not sure why it fails, and I'm not sure where to start.

TheNavigat:
One more comment to make things clear.

I used augtool to check. `ls /files/etc/apache2` prints the files correctly, `ls /files/private/etc/apache2` prints nothing. Running augtool using sudo makes no difference.

And I did the following, too.

```
Python 2.7.11 (default, Jan 22 2016, 08:29:18) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.abspath("/etc/apache2")
'/etc/apache2'
>>> os.path.realpath("/etc/apache2")
'/private/etc/apache2'
>>> 
```

Just to confirm my findings.

TheNavigat:
https://github.com/TheNavigat/letsencrypt/blob/1e68d614bc5aff86bcc6371c9386da00b1c20141/certbot-apache/certbot_apache/tests/configurator_test.py#L760

Mysteriously, this test fails locally, yet it passes successfully on Travis.

TheNavigat:
Turns out I my branch wasn't updated. I think the PR is ready now! :)

.travis.yml
```diff
@@ -69,6 +69,14 @@ matrix:
       env: TOXENV=py34
     - python: "3.5"
       env: TOXENV=py35
+    - os: osx
+      osx_image: xcode7.3
```
bmw: I'm curious if we can just use the default image? Have you tried it? If not, you should. If you have, you should add a comment saying why it's necessary.

.travis.yml
```diff
@@ -69,6 +69,14 @@ matrix:
...
+      sudo: required
```
bmw: Why do we need `sudo`?

.travis.yml
TheNavigat: I did. `setuptools` is too old. I'll check again and let you know the exact version.

.travis.yml
```diff
@@ -69,6 +69,14 @@ matrix:
...
+      language: generic
+      env: TOXENV=cover
+      before_install: "travis_retry brew install python augeas dialog"
```
bmw: nit: Saw [this](https://docs.travis-ci.com/user/osx-ci-environment/#Homebrew). Maybe we should change this to
```
before_install:
    - travis_retry brew update
    - travis_retry brew install python augeas dialog
```

.travis.yml
```diff
@@ -69,6 +69,14 @@ matrix:
...
+      install: "travis_retry pip install tox coveralls"
+      script: 'travis_retry tox && ([ "xxx$BOULDER_INTEGRATION" = "xxx" ] || ./tests/travis-integration.sh)'
```
bmw: Are either of `install` or `script` necessary here?

bmw:
Other than my comments on `.travis.yml` this LGTM. Thanks for this PR @TheNavigat. I'm very happy to have these tests.

.travis.yml
TheNavigat: Hmmmm... Not sure where did I get that `setuptools` idea.

Anyway,

`RuntimeError: You are linking against OpenSSL 0.9.8, which is no longer support by the OpenSSL project. You need to upgrade to a newer version of OpenSSL.`. Here's the full Travis log. https://api.travis-ci.org/jobs/141770652/log.txt?deansi=true

bmw: Ah. That makes sense. I think it's fine to leave this as is as long as we add a comment saying something like `cryptography no longer supports OpenSSL 0.9.8`.

TheNavigat: What about upgrading OpenSSL through brew though if we really need to run it on the default image?

I'm with running the latest for other reasons, because it's the most commonly used, but if we need the default image maybe we can make it work.

bmw: I think the benefits are minimal/nonexistent and it's likely to slow down our build. I was just curious why/if we had to specify the xcode version. Makes sense now.

bmw: I'd still like to see a comment like one I suggested above added to `.travis.yml`.

.travis.yml
bmw: GitHub seems to think that this comment is from an outdated diff, however, this still hasn't been addressed. If you disagree with me, please let me know why and we can talk about it, but otherwise, I'd like to see this change made before merging.

DomT4: Coming from a Homebrew maintainer point of view: Please do run `brew update` when using Travis. Some Travis boxes are quite severely out of date & there's not much we can do to help if you hit issues running outdated versions of `brew`.

bmw:
Travis seems to freeze like this from time to time. Closing the PR, waiting 10 seconds, and then reopening usually resolves the issue.

.travis.yml
TheNavigat: I'm aware, the problem is that `brew update` broke earlier builds for some reason. I was waiting for Travis builds to run to be able to determine whether it broke because of `brew update` or not.

I'm not ignoring @bmw, don't worry. Thanks a lot for the note! :)

TheNavigat:
Augeas 1.5.0 upgrade causes tests to fail. There are too many failing ones so I decided to pick just one, which is `test_bad_parse`, https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/tests/augeas_configurator_test.py#L30. This test fails because `lens_path` in https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/augeas_configurator.py#L69 is always false.

Additionally, running `self.aug.get(path)` there returns the following.

```
u'/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:55.26-.38:Undefined variable Util.comment
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:56.26-.37:Undefined variable Util.doseol
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:57.26-.40:Undefined variable Util.empty_dos
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:58.26-.37:Undefined variable Util.indent
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:90.20-.32:Undefined variable Util.del_str
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:99.26-.47:Undefined variable arg_wordlist
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:106.34-.48:Undefined variable comment
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:112.21-.27:Undefined variable indent
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:117.18-.25:Undefined variable section
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:117.26-.44:Undefined variable directive
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:120.10-.18:Undefined variable content
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:133.13-.25:Undefined variable Util.stdexcl
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:135.20-.23:Undefined variable lns
/Users/TheNavigat/Documents/letsencrypt/certbot-apache/certbot_apache/augeas_lens/httpd.aug:135.24-.30:Undefined variable filter'
```

Which is quite peculiar, I believe.

I think I'm stuck.

bmw:
Are you still interested and have the time to work on this? If so, can you bring this PR up to date and see if you're still having issues? If you are, I'm happy to help you, but because I don't have a OS X machine available, you will have to track down some of these problems on your own.

TheNavigat:
I've lost my hackintosh as I switched to a newer laptop, but I'm still interested. I'll bring this PR up to date and then see what I can do. Worst case scenario I can borrow someone's Mac for a bit.

bmw:
This PR hasn't seen any progress in almost a year. If you continue working on it, feel free to open a new PR.

PR #3037, opened 2016-05-20
Add --disable-hook-validation

mgedmin:
As discussed in #3020.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -356,7 +356,8 @@ class HelpfulArgumentParser(object):
                         " {0} conflicts with dialog_mode").format(arg)
                     )
 
-        hooks.validate_hooks(parsed_args)
+        if parsed_args.validate_hooks:
+            hooks.validate_hooks(parsed_args)
 
         return parsed_args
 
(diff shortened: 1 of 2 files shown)


mgedmin:
Two things I'm not sure about:
- I haven't added any tests
- The option name (disable-hook-validation) doesn't match the attribute name (config.validate_hooks), because I dislike double negatives ("if not config.disable_hook_validation: ...")

pde:
Tests should be easy to add! Just repeat the [explodinator case](https://github.com/certbot/certbot/blob/master/certbot/tests/hook_test.py#L20) in the existing tests with `config.validate_hooks = False`, but instead of `assertRaises` you just call the function (if an exception occurs, the test will fail).

pde:
You also have a [pylint problem](https://travis-ci.org/certbot/certbot/jobs/131611058#L1446); I think you can just disable that pylint check for that function though.

mgedmin:
Rebased on master, resolved merge conflicts.

Will see about the rest a bit later.

mgedmin:
> Tests should be easy to add! Just repeat the explodinator case in the existing tests with config.validate_hooks = False, but instead of assertRaises you just call the function (if an exception occurs, the test will fail).

Well, that test is calling the `validate_hooks()` function directly, but I didn't change it.  The check I added is outside.

I either need to move the check inside `validate_hooks()` (turning it into a NOP when the config says not to validate), or find a different way to test the change.

mgedmin:
The pylint error went away on its own, after I rebased on master.

mgedmin:
Ok, tests are here!

PR #3041, opened 2016-05-20
Fixing question in non-interactive mode

### LINKED ISSUE #3039: Asking for input despite --noninteractive
I am using [this script](https://github.com/tozd/docker-nginx-proxy/blob/master/letsencrypt/dockergen.sh) to automatically generate certificates for my domains. I have the following command:

```
/letsencrypt/data/letsencrypt/bin/letsencrypt --no-self-upgrade --noninteractive --quiet --agree-tos --email "${LETSENCRYPT_EMAIL}" \
 --config-dir /ssl/letsencrypt certonly --webroot --keep-until-expiring --rsa-key-size 4096 \
 --webroot-map "${LIST_JSON}"
```

As you see, I use `--noninteractive`. But despite this, my automatic process has been stopped by (I see in the log) the following prompt:

```
You have an existing certificate that contains a portion of the domains you requested (ref: /ssl/letsencrypt/renewal/example.com.conf)

It contains these names: example.com

You requested these names for the new certificate: example.com, test.example.com.

Do you want to expand and replace this existing certificate with the new certificate?

(You can set this with the --expand flag)
```

I do not want to expand it the domain, and I would assume that `--noninteractive` should choose the default: not to expand and just do it.

TheNavigat:
Closes #3039

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -123,7 +123,9 @@ def _handle_subset_cert_request(config, domains, cert):
              existing,
              ", ".join(domains),
              br=os.linesep)
-    if config.expand or config.renew_by_default or zope.component.getUtility(
+    if config.noninteractive_mode:
+        return "newcert", None
+    elif config.expand or config.renew_by_default or zope.component.getUtility(
             interfaces.IDisplay).yesno(question, "Expand", "Cancel",
                                        cli_flag="--expand"):
         return "renew", cert


mitar:
But this will expand by default, no?

TheNavigat:
You are correct. I apologize for the mistake.

Can you please check the code again? If I'm not mistaken, this should work correctly.

mitar:
Not sure. This is definitely something I would like to be the result, personally. That it just creates a new cert. But by looking at the code this seems to not be allowed? That if you in interactive mode say "no", that then you have to use `--duplicate` mode? I am not sure if this is really required or why it would be required. Why you cannot have two certs, one for `example.com` and one for `test.example.com`.

TheNavigat:
I'm not following.

You have 2 options, either to overwrite the existing cert files, or to have duplicates. I think the logged error is just to make sure that the user doesn't unknowingly create duplicate certs, which does make sense.

In that sense actually it makes more sense that expanding should be the default action, and not duplication, but a second opinion is always helpful, and in this case yours made sense. If most other people agree, then we should go for duplication as the default action.

mitar:
But then you should check that `--duplicate` flag was provided? So it should be that if there is noninteractive flag, that you display the error, and if there is a duplicate flag, then you issue it.

But again, why is `example.com` and `test.example.com` considered duplicate?

bmw:
I believe this can be closed. See my comment [here](https://github.com/certbot/certbot/issues/3039#issuecomment-220707150).

PR #3042, opened 2016-05-20
Limiting tox envlist to really needed tests

### LINKED ISSUE #2556: Update tox/instructions
As pointed out by @TheBoegl, our [instructions for submitting a PR](https://letsencrypt.readthedocs.org/en/latest/contributing.html#submitting-a-pull-request) are misleading. Step 5 says:

> Run `tox` to run the entire test suite including coverage. Fix any errors.

Currently, running `tox` runs the following tests: py{26,27,33,34,35},py{26,27}-oldest,cover,lint.

Many of these tests such as tests for 5 python versions as well of the *-oldest tests which require an older version OpenSSL are not likely to run successfully on a user's system.

We should either:

1. Update `envlist` in `tox.ini` so that only a subset of the current tests are run. In most cases, I think cover, lint, and maybe a py3x test would be sufficient. We should, however, continue running all of our tests in Travis.
2. Update the developer instructions to explain that `ERROR: InterpreterNotFound:` for the py* tests or compilation failures for the *-oldest tests do not indicate a problem with your PR and you can continue with submitting your PR.

TheNavigat:
Closes #2556

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -71,6 +71,9 @@ The following tools are there to help you:
   experimental, non-production Apache2 install on them.  ``tox -e
   apacheconftest`` can be used to run those specific Apache conf tests.
 
+- ``tox --skip-missing-interpreters`` runs tox while ignoring missing versions
+  of Python needed for running the tests.
+
 - ``tox -e py27``, ``tox -e py26`` etc, run unit tests for specific Python
   versions.
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -317,7 +317,9 @@ Steps:
 3. Run ``./pep8.travis.sh`` to do a cursory check of your code style.
    Fix any errors.
 4. Run ``tox -e lint`` to check for pylint errors. Fix any errors.
-5. Run ``tox`` to run the entire test suite including coverage. Fix any errors.
+5. Run ``tox --skip-missing-interpreters`` to run the entire test suite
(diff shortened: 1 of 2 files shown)


pde:
I'm not sure that it's a good idea to remove the `-oldest` tests unless we have a good plan for running them before release. But perhaps it could be sensible to stop them from slowing down every PR we land? @bmw what are your thoughts there?

bmw:
So this PR just changes what tests are run when running `tox` locally. It does not affect the tests run in Travis.

With this in mind, I think we can remove the oldest tests. As the name suggests, they only work on old systems and generate really nasty errors when they fail.

I also think we can remove `py27`. `py27` is the exact same thing as `cover`, except that `cover` also checks test coverage. We could even remove `py27` from Travis, however, if we do this, we should also run integration tests when running `cover`.

Beyond that though, what about keeping `py{26,27,33,34,35},cover,lint` in the envlist, but updating our developer instructions telling people to run `tox --skip-missing-interpreters`. This will have `tox` run all the tests it can, but if it can't find the specified Python version, it's not considered an error. If we wanted, we can maybe add a sentence explaining the use of this flag.

bmw:
nit: I'd also change step 5 of [submitting a PR](https://github.com/certbot/certbot/blob/master/docs/contributing.rst#submitting-a-pull-request) to use `--skip-missing-interpreters` as that line was the motivation for creating the issue fixed by this PR.

Other than that though, LGTM.

bmw:
LGTM.

PR #3054, opened 2016-05-23
Pin old pkginfo version

bmw:


bmw:
Self merging because it only affects developer deps and this fixes integration tests.

PR #3058, opened 2016-05-24
Run Boulder via docker-compose in tests.

jsha:
This removes a lot of setup code we used to need in order to get Boulder to run,
and should reduce brittleness of tests based on Boulder changes.

This also unblocks Boulder from upgrading to MariaDB 10.1 in integration tests,
since changing to 10.1 syntax for user creation would break the current certbot
integration tests (which run 10.0).

Note: the apacheconftest task was failing until I uncommented the .travis.yml section that installs apache. Is that correct?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
 # test has failed
 env:
   global:
-    - BOULDERPATH=$GOPATH/src/github.com/letsencrypt/boulder/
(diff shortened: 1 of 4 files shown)


bmw:
There's a few small things I'd like to change here, but for the most part, this LGTM. I'll go ahead and merge this PR to unlock Boulder and send other changes in a subsequent PR.

Thanks @jsha.

Issue #2687, opened 2016-03-19
Please check FQDN check string


Mirocow:
In this line https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/le_util.py#L341

this domain ws.dev.somedomain.com don`t check by regular string

```
^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,63}$
```

but such it work

```
^((?!-)[[A-Za-z0-9-]{1,63}(?<!-)\.)+[A-Za-z]{2,6}$
```

joohoi:
What version of python are you running? It should match:

```
$ python -c 'import re;print(re.compile("^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,63}$").match("ws.dev.somedomain.com").string)'
ws.dev.somedomain.com
```

Mirocow:
2.7

joohoi:
That's strange, it matches fine in 2.7 for me too. Please provide steps to reproduce.

ohemorange:
Seems that this was fixed in #3061.

PR #3061, opened 2016-05-24
Fix FQDN checks, closes #3057 and #3056 [needs minor revision]

### LINKED ISSUE #3056: Installer thinks bare TLD is not a valid FQDN
If you attempt to enter either "[ai](http://ai.)" or "[ai.](http://ai.)" (same with [to](http://to.)), the installer complains about it not being a valid FQDN. It is a valid FQDN, and the installer should allow it to proceed with validation. Oddly enough, the installer *does* treat `user@tld` properly for email entry.

Other TLDs such as IO used to also have websites hosted directly on their TLD. [ai](http://ai.) and [to](http://to.) are just two such examples of websites being hosted directly on a TLD.

### LINKED ISSUE #3057: Bug in FQDN detection: installer wrongly interprets _
## Problem

I'm attempting to get a certificate for a new domain with a following format:

`wkoszek_nc2.x.mycompany.com`

However the client seems to get confused about `_`. Upon removing the `_` from the domain name, it works fine. For example, this works OK:

`wkoszeknc2.x.mycompany.com`

## To reproduce the problem:

Use this command to reproduce:

`# ./certbot-auto certonly --standalone --email wkoszek@mycompany.com -d wkoszek_nc2.x.mycompany.com
`

I'm getting this:

```
Checking for new version...
Requesting root privileges to run certbot...
   /root/.local/share/letsencrypt/bin/letsencrypt certonly --standalone --email wkoszek@mycompany.com -d wkoszek_nc2.x.mycompany.com
Requested domain wkoszek_nc2.x.mycompany.com is not a FQDN
```

cowlicks:
To test this fixes #3057 try
```
# ./certbot-auto certonly --standalone --email wkoszek@mycompany.com -d wkoszek_nc2.x.mycompany.com 
```

To test this fixes #3056 try
```
# ./certbot-auto certonly --standalone --email wkoszek@mycompany.com -d ai
```

These no longer raise configuration errors.

To fix this I changed the `certbot.le_util.enforce_domain_sanity`  check to match what [RFC 2181](https://tools.ietf.org/html/rfc2181#section-11) specifies as a valid domain name.

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -431,7 +431,7 @@ def enforce_domain_sanity(domain):
     for l in labels:
         if not 0 < len(l) < 64:
             raise errors.ConfigurationError(msg + "label {0} is too long.".format(l))
-    if len(domain) > 256:
+    if len(domain) > 255:
         raise errors.ConfigurationError(msg + "it is too long.")
 
     return domain
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -423,14 +423,17 @@ def enforce_domain_sanity(domain):
         # It wasn't an IP address, so that's good
         pass
 
-    # FQDN checks from
-    # http://www.mkyong.com/regular-expressions/domain-name-regular-expression-example/
-    #  Characters used, domain parts < 63 chars, tld > 1 < 64 chars
-    #  first and last char is not "-"
-    fqdn = re.compile("^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,63}$")
-    if not fqdn.match(domain):
-        raise errors.ConfigurationError("Requested domain {0} is not a FQDN"
-                                        .format(domain))
+    # FQDN checks according to RFC 2181: domain name should be less than 255
+    # octets (inclusive). And each label is 1 - 63 octets (inclusive).
+    # https://tools.ietf.org/html/rfc2181#section-11
+    msg = "Requested domain {0} is not a FQDN because ".format(domain)
+    labels = domain.split('.')
+    for l in labels:
+        if not 0 < len(l) < 64:
+            raise errors.ConfigurationError(msg + "label {0} is too long.".format(l))
+    if len(domain) > 256:
+        raise errors.ConfigurationError(msg + "it is too long.")
+
     return domain
 
 
--- a/certbot/tests/cli_test.py
+++ b/certbot/tests/cli_test.py
@@ -342,11 +342,11 @@ class CLITest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         # FQDN
         self.assertRaises(errors.ConfigurationError,
                           self._call,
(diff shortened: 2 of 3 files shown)


bmw:
The boulder docker image we use for tests is currently broken so integration tests are going to fail. I'll kick Travis when its working again though.

cowlicks:
@bmw thank you

Note that this does not allow non-ascii characters, those are already caught [here](https://github.com/certbot/certbot/pull/3061/commits/f3f1f9111260d74e43544d277104009cd1b6f282#diff-a2038cb40ab8c4b6f5e1accd20a2ea37R313). It permits all ascii, instead of just `-, a-z, A-Z, 0-9`.

bmw:
So the fix required changing our `.travis.yml` so kicking Travis won't work. If you merge `master` into this PR though, tests should pass.

Sorry for the trouble!

cowlicks:
Rebased

cowlicks:
This would also close #2687.

However it seems to be the LE CA policy to *reject* FQDN with characters outside of `[a-zA-Z0-9-]` ([link](https://github.com/letsencrypt/boulder/pull/1437#issuecomment-175963839)). But it is allowed by other CA's.

Is certbot just supposed to work with the LE CA? If so, this should be closed (or maybe refactored to just fix #3056).

certbot/util.py
```diff
@@ -423,14 +423,17 @@ def enforce_domain_sanity(domain):
...
-                                        .format(domain))
+    # FQDN checks according to RFC 2181: domain name should be less than 255
+    # octets (inclusive). And each label is 1 - 63 octets (inclusive).
+    # https://tools.ietf.org/html/rfc2181#section-11
+    msg = "Requested domain {0} is not a FQDN because ".format(domain)
+    labels = domain.split('.')
+    for l in labels:
+        if not 0 < len(l) < 64:
+            raise errors.ConfigurationError(msg + "label {0} is too long.".format(l))
+    if len(domain) > 256:
```
bmw: nit: Shouldn't this be `len(domain) > 255`?

bmw:
Sorry this sat around for a bit @cowlicks. It actually generated some discussion around the EFF office about what's best to do here.

`certbot` is designed to work with any server who supports ACME, not just Let's Encrypt. With this in mind, I think we should move forward on this PR. I mentioned a small change I'd like to see above, but other than that, this LGTM.

certbot/util.py
cowlicks: [yup](https://tools.ietf.org/html/rfc2181#section-11). 

```
A full domain name is limited to 255 octets (including the separators).
```

cowlicks:
@bmw Thanks! Fixed.

If y'all are looking for something else to discuss in the office, I sent an application in a few week ago ;)

bmw:
That's great! I'll keep that in mind at the watercooler :)

Thanks for another PR. LGTM.

PR #3062, opened 2016-05-24
Always remove temporary directories in le-auto

### LINKED ISSUE #3060: certbot-auto is not cleaning up temporary directory
The certbot-auto (letsencrypt-auto) script leaves behind a temporary directory every time it runs.  This looks like a regression as of release v0.5.0 (commit 4d4cfb414f2847f1814cfe2949c6e76f112e4ade) which removes the `trap "rm -rf '$TEMP_DIR'" EXIT` statement.

Assuming removing the `trap` was intentional (it may not have been; I can't find an issue or commit that refers specifically to that change, and it also temporarily disappeared in the "bad" pip8 branch merge), here's where the problem is:

```sh
if [ "$NO_SELF_UPGRADE" != 1 ]; then
  echo "Checking for new version..."
  TEMP_DIR=$(TempDir)

  ...

  if ! REMOTE_VERSION=`"$LE_PYTHON" "$TEMP_DIR/fetch.py" --latest-version` ; then
    echo "WARNING: unable to check for updates."
  elif [ "$LE_AUTO_VERSION" != "$REMOTE_VERSION" ]; then

    ...

    # TODO: Clean up temp dir safely, even if it has quotes in its path.
    rm -rf "$TEMP_DIR"
  fi  # A newer version is available.

  >>>$TEMP_DIR cleanup should be here instead<<<
fi  # Self-upgrading is allowed.
```

bmw:
Fixes #3060.

Tested on the test farm, `dash`, and FreeBSD.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -532,6 +532,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
 
     echo "Installing Python packages..."
     TEMP_DIR=$(TempDir)
+    trap 'rm -rf "$TEMP_DIR"' EXIT
     # There is no $ interpolation due to quotes on starting heredoc delimiter.
     # -------------------------------------------------------------------------
     cat << "UNLIKELY_EOF" > "$TEMP_DIR/letsencrypt-auto-requirements.txt"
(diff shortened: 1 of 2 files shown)


SwartzCr:
lgtm

PR #3064, opened 2016-05-24
-v implies --text

TheNavigat:
Closes issue #2674

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -392,6 +392,11 @@ class HelpfulArgumentParser(object):
                         ("Conflicting values for displayer."
                         " {0} conflicts with dialog_mode").format(arg)
                     )
+        else:
+            # -v should imply --text
+            if (parsed_args.verbose_count > flag_default("verbose_count") and
+                not parsed_args.dialog_mode):
+                parsed_args.text_mode = True
 
         if parsed_args.validate_hooks:
             hooks.validate_hooks(parsed_args)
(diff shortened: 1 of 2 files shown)


certbot/main.py
```diff
@@ -653,6 +653,10 @@ def main(cli_args=sys.argv[1:]):
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
 
+    # -v should imply --text
+    if config.verbose_count > cli.flag_default("verbose_count"):
+        config.text_mode = True
```
bmw: This should be moved to `HelpfulArgumentParser.parse_args` and should not be set if `dialog_mode` is set.

certbot/cli.py
```diff
@@ -392,6 +392,11 @@ def parse_args(self):
                         ("Conflicting values for displayer."
                         " {0} conflicts with dialog_mode").format(arg)
                     )
+        else:
+            # -v should imply --text
+            if (parsed_args.verbose_count > flag_default("verbose_count") and
```
bmw: nit:
```
else:
    if:
```

should just be `elif:`.

certbot/cli.py
```diff
@@ -392,6 +392,11 @@ def parse_args(self):
...
+                not parsed_args.dialog_mode):
```
bmw: Checking `not parsed_args.dialog_mode` isn't necessary since we're in the `else` branch of `if parsed_args.dialog_mode`.

bmw:
Other than my nits, I think we should also add a test that this actually works to help prevent regressions.

bmw:
LGTM

PR #3065, opened 2016-05-24
Check out Boulder master instead of branch.

jsha:
We temporarily had the integration tests checking out a branch of Boulder while we waited to land some code. Now that code is in master and we can remove the branch flag.


bmw:
LGTM.

Can merge when Travis is green.

PR #3069, opened 2016-05-25
Autoconfigure OCSP Stapling with --must-staple

sagi:


bmw:
So I think we can simplify this a bit. In [parse_args](https://github.com/certbot/certbot/blob/master/certbot/cli.py#L321), we do any necessary post-processing like setting interactions between flags.

How about we just add two lines like:
```
if parsed_args.must_staple:
    parsed_args.staple = True
```

After that, all changes to the client and client tests can be reverted and we can write a simple test like the one found in the first few lines of [this test](https://github.com/certbot/certbot/blob/master/certbot/tests/cli_test.py#L425).

certbot/interfaces.py
```diff
@@ -201,9 +201,9 @@ class IConfig(zope.interface.Interface):
         "Email used for registration and recovery contact.")
     rsa_key_size = zope.interface.Attribute("Size of the RSA key.")
     must_staple = zope.interface.Attribute(
-        "Whether to request the OCSP Must Staple certificate extension. "
-        "Additional setup may be required after issuance. This does not "
-        "currently autoconfigure web servers for OCSP stapling. ")
+        "Adds the OCSP Must Staple extension to the certificate."
```
bmw: Missing space at the end of this line. Resulting help output becomes:
```
  --must-staple         Adds the OCSP Must Staple extension to the
                        certificate.Autoconfigures OCSP Stapling for supported
                        setups (Apache version >= 2.3.3 ). (default: False)
```

bmw:
LGTM. Will merge when Travis is happy.

PR #3073, opened 2016-05-25
Rename certbot.le_util to certbot.util

cowlicks:
This renames the `le_util.py` module to `util.py` because the `le_` prefix was a holdover from the letsencrypt name. I decided `util.py` was better than `cb_util.py` or something because its path is `certbot/util.py` so `certbot/cb_util.py` seemed redundant, and using `util` didn't cause any name collisions. 

Also rename `certbot/tests/le_util_test.py` to `certbot/tests/util_test.py`


cowlicks:
All the boulder tests failed. I'm not sure what that is, but I just found the docs for it. brb

pde:
We might want to do a quick grep of the major third party plugin codebases to see if any of them would be affected by this. If so, we may want to create a transitional `le_util` wrapper around the new `certbot.util` module :/

cowlicks:
@pde Would we just do that in the letsencrypt dummy package ([here](https://github.com/certbot/certbot/tree/master/letsencrypt/letsencrypt))? Or would it need to happen in certbot too?

I was assuming we could be pretty blasé about api breaking since this is beta and it just had a huge renaming. If it is better to hold off on this until an api breaking release, that is fine. It will be easier later having this PR to rebase off of.

bmw:
Looking through all 3rd party plugins in the [wiki](https://github.com/certbot/certbot/wiki/Plugins) and [documentation](https://certbot.eff.org/docs/using.html#third-party-plugins), it appears none of them use `le_util` so I'm not sure this dummy package is necessary.

pde:
Ok, LGTM.

PR #3076, opened 2016-05-26
0.7.0 release script prep

bmw:
Turns out not much had to be done. This makes the release script a little more convenient, stop packaging `letshelp-*`, and properly update `letsencrypt-auto-requirements.txt`.

@pde, it's probably best if you can review this, but the changes are very simple if anyone else wants to take a crack at it.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -176,7 +176,7 @@ if ! wc -l /tmp/hashes.$$ | grep -qE "^\s*15 " ; then
 fi
 
 # perform hideous surgery on requirements.txt...
-head -n -9 letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt > /tmp/req.$$
+head -n -15 letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt > /tmp/req.$$
 cat /tmp/hashes.$$ >> /tmp/req.$$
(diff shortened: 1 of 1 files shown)


bmw:
Travis isn't starting...

Closing and reopening.

tools/release.sh
```diff
@@ -162,21 +162,21 @@ for module in certbot $subpkgs_modules ; do
...
     nosetests $module
 done
-deactivate
 
 # pin pip hashes of the things we just built
 for pkg in acme certbot certbot-apache letsencrypt letsencrypt-apache ; do
     echo $pkg==$version \\
     pip hash dist."$version/$pkg"/*.{whl,gz} | grep "^--hash" | python2 -c 'from sys import stdin; input = stdin.read(); print "   ", input.replace("\n--hash", " \\\n    --hash"),'
 done > /tmp/hashes.$$
+deactivate
```
pde: Is this being moved so that `pip` is used from the venv rather than the OS?  If that's the reason, this PR LGTM.

bmw: Yep! That's the reason. This allows the script to work even if you don't have `pip` installed as well as making sure you're using the latest `pip` (we update `pip` in the venv).

PR #3086, opened 2016-05-27
Add code layout to contributing.rst [has suggestion]

pde:


bmw:
This looks like a good start to the issue. Thinking about this though, this information changes over time as modules are created, removed, and modified. What if instead of doing all of this, we provide a section like the one found in this PR laying out the different packages in the project (`acme`, `certbot`, `certbot-apache`, etc.). In each section, we describe the basic usage and link to the API documentation for that package.

We'll want to make sure all submodules are included (e.g. `cli.py` is not) and ideally add better docstrings for some of them (`certbot.client` is "Certbot client API" which tells the user literally no new information). Using this approach allows the user to dive into the lowest level of detail they want and helps ensure the documentation stays up to date as the project develops.

bmw:
I made what I think was a good suggestion on how we can improve this and never heard back. Closing as part of my mission to clean up our PR queue, but reopen when you'd like to continue work on this.

Issue #3091, opened 2016-05-29
certbot-apache broken on Gentoo: incorrect os info parsing?


LeCoyote:
In certbot-apache's constants.py, the client expects `"gentoo base system"` as key for the OS-specific defaults. However, `os_info[0]` is `"gentoo"`. As a result, certbot-apache fails with "Unable to find Apache version" since it relies on the non-working debian defaults.

I am not sure, but I feel that "gentoo base system" is a leftover from letsencrypt-auto which tested for distribution-specific files, such as `/etc/gentoo-release` which indeed contains that string. In contrast, certbot seems to rely on `/etc/os-release`, which only mentions "gentoo".

Therefore, I suggest that "gentoo base system" be changed to simply "gentoo". At least it fixes the problem for me.

LeCoyote:
Proposed pull-request: #3092

LeCoyote:
Actually the old behaviour was to use `platform.linux_distribution()`, which indeed yields "Gentoo Base System" as first element of the array. The problem is still the same, though.

PR #3092, opened 2016-05-29
Changed Gentoo os_info string

LeCoyote:
Gentoo's os_info is not what is expected in constants.py, resulting in certbot-apache systematic failure.
See bug #3091

### DIFF
--- a/certbot-apache/certbot_apache/constants.py
+++ b/certbot-apache/certbot_apache/constants.py
@@ -81,6 +81,7 @@ CLI_DEFAULTS = {
     "rhel": CLI_DEFAULTS_CENTOS,
     "amazon": CLI_DEFAULTS_CENTOS,
     "gentoo": CLI_DEFAULTS_GENTOO,
+    "gentoo base system": CLI_DEFAULTS_GENTOO,
     "darwin": CLI_DEFAULTS_DARWIN,
 }
 """CLI defaults."""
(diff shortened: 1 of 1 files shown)


pde:
@joohoi, want to review?

joohoi:
This is more complicated, as Gentoo enables user to choose wheter to use systemd or not. In this @LeCoyote must be running systemd, while the string in constants.py certainly is for platform.linux_distrubution().

certbot-apache/certbot_apache/constants.py
```diff
@@ -80,7 +80,7 @@
     "red hat enterprise linux server": CLI_DEFAULTS_CENTOS,
     "rhel": CLI_DEFAULTS_CENTOS,
     "amazon": CLI_DEFAULTS_CENTOS,
-    "gentoo base system": CLI_DEFAULTS_GENTOO,
```
joohoi: Do not delete this line, just add the new one for the systemd /etc/os-release string.

LeCoyote:
Actually, I am **not** using systemd but openrc.
Just to clarify, since /etc/os-release exists as part of Gentoo's baselayout, its contents are used even if systemd is **not** installed (which is my case). In other words, on a non-systemd Gentoo, certbot does **not** use platform.linux_distribution() but just reads /etc/os-release. Hence my suggestion to amend the string.

joohoi:
You are right. Gentoo seems to have adopted the systemd os-release configuration file. That's good. However I still think we should leave the platform.linux_distribution() string in place just in case.

LeCoyote:
So you're saying we should have two lines, one for each string, that will result in using CLI_DEFUALTS_GENTOO? Just to clarify :)

joohoi:
Yes, exactly.

LeCoyote:
There you go :)

joohoi:
LGTM

bmw:
LGTM as well.

PR #3101, opened 2016-05-31
Bump minimum psutil version

### LINKED ISSUE #1080: ValueError: ambiguos inode with multiple PIDs references
I got some trouble with the standalone authentication plugin.

Apparantly, the authenticator tries to check whether certain ports which it needs to listen on are already taken or if it can open them. On my machine, this check causes the client to crash, no matter whether my web server is running or not.

      2015-10-22 06:34:12,046:DEBUG:letsencrypt.cli:Exiting abnormally:
	File "/usr/bin/letsencrypt", line 9, in <module>
	  load_entry_point('letsencrypt==0.1', 'console_scripts', 'letsencrypt')()
	File "/usr/lib/python2.7/dist-packages/letsencrypt/cli.py", line 926, in main
	File "/usr/lib/python2.7/dist-packages/letsencrypt/cli.py", line 335, in auth
	  domains, authenticator, installer, plugins):
	File "/usr/lib/python2.7/dist-packages/letsencrypt/client.py", line 240, in obtain_and_enroll_certificate
	File "/usr/lib/python2.7/dist-packages/letsencrypt/client.py", line 216, in obtain_certificate
	  return self._obtain_certificate(domains, csr) + (key, csr)
	File "/usr/lib/python2.7/dist-packages/letsencrypt/client.py", line 174, in _obtain_certificate
	  authzr = self.auth_handler.get_authorizations(domains)
	File "/usr/lib/python2.7/dist-packages/letsencrypt/auth_handler.py", line 80, in get_authorizations
	  cont_resp, dv_resp = self._solve_challenges()
	File "/usr/lib/python2.7/dist-packages/letsencrypt/auth_handler.py", line 115, in _solve_challenges
	  dv_resp = self.dv_auth.perform(self.dv_c)
	File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/standalone/authenticator.py", line 386, in perform
	  if self.already_listening(self.config.dvsni_port):
	File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/standalone/authenticator.py", line 304, in already_listening
	  listeners = [conn.pid for conn in psutil.net_connections()
	File "/usr/lib/python2.7/dist-packages/psutil/__init__.py", line 1790, in net_connections
	  return _psplatform.net_connections(kind)
	File "/usr/lib/python2.7/dist-packages/psutil/_pslinux.py", line 572, in net_connections
	  return _connections.retrieve(kind)
	File "/usr/lib/python2.7/dist-packages/psutil/_pslinux.py", line 556, in retrieve
	  for fd, family, type_, laddr, raddr, status, bound_pid in ls:
	File "/usr/lib/python2.7/dist-packages/psutil/_pslinux.py", line 490, in process_inet
	  raise ValueError("ambiguos inode with multiple "
      ValueError: ambiguos inode with multiple PIDs references

I think that this issue might be related to my slightly unusual N/W configuration which involves several interfaces and some strange routing arrangements; However, I still don't get what is happening here.

bmw:
Fixes #1080.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ install_requires = [
     'configobj',
     'cryptography>=0.7',  # load_pem_x509_certificate
     'parsedatetime>=1.3',  # Calendar.parseDT
-    'psutil>=2.1.0',  # net_connections introduced in 2.1.0
+    'psutil>=2.2.1',  # 2.1.0 for net_connections and 2.2.1 resolves #1080
     'PyOpenSSL',
     'pyrfc3339',
     'python2-pythondialog>=3.2.2rc1',  # Debian squeeze support, cf. #280


pde:
LGTM with the proviso that we should only land this if we're sure that there isn't a backwards-compatible workaround.

bmw:
The corresponding issues on the `psutil` repo are giampaolo/psutil#497 and giampaolo/psutil#572. The fix for this was to just comment out the code raising the exception as you can see in their current [master](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L680) branch. The code causing problems is in every release from 2.1.0 to 2.2.0. `psutil` 2.2.1 is not in `jessie-backports`, however, it is already in EPEL 7.

As for other workarounds, we could:

1. We could remove the use of `psutil`. This means we can't provide as much information about the process using the port we want, but it allows us to drop the dependency altogether.
2. We could potentially look into seeing if there's another function we can call other than `net_connections`. Maybe there is and it will work great. However, there's no guarantees we won't see the same or more by changing the function here.
3. Catch the exception raised by `psutil` and don't use it in the message about not being able to bind to port 80/443.

pde:
I think probably we should do 3, and print a note if the exception saying `(your version of python-psutil is buggy; Certbot can't determine what process is using port 80/443)`.

pde:
Update from discussion: we're going to try #3106.

PR #3102, opened 2016-06-01
docs: Fix generation of manpage 'certbot.1'.

lfam:
* docs/man/certbot.rst: Fix path to resource file.

Without this change, the manpage build process cannot find 'docs/cli-help.txt',
and the resulting certbot.1 includes only the header and footer, but no actual
documentation.


pde:
Thanks!

PR #3109, opened 2016-06-02
Stop packaging shim packages

bmw:
This PR stops packaging shim packages but continues to have `certbot-auto` install the `letsencrypt` package for backwards compatibility with people running `letsencrypt` manually from the venv.


pde:
LGTM

PR #3111, opened 2016-06-02
Exit if cannot bootstrap in certbot-auto

### LINKED ISSUE #2540: letsencrypt-auto gives "sudo" is not available
Hello guys,

Look at the below, I am trying to set it up.

>
> bash-3.2$ ./letsencrypt-auto —help
> “sudo” is not available, will use “su” for installation steps…
> Sorry, I don’t know how to bootstrap Let’s Encrypt on your operating system!
> 
> You will need to bootstrap, configure virtualenv, and run a peep install manually.
> Please see https://letsencrypt.readthedocs.org/en/latest/contributing.html#prerequisites
> for more info.
> Checking for new version…
> ./letsencrypt-auto: line 108: [: too many arguments
>   File “/tmp/tmp.jEueeZ1320/fetch.py”, line 57
>     except (HTTPError, IOError) as exc:
>                                  ^
> SyntaxError: invalid syntax
> bash-3.2$ 
>

I am hosted on TSOHOST.COM on Apache -- any guess on what is going wrong here?

bmw:
Fixes #2540.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -500,6 +500,7 @@ Bootstrap() {
     echo "You will need to bootstrap, configure virtualenv, and run pip install manually."
     echo "Please see https://letsencrypt.readthedocs.org/en/latest/contributing.html#prerequisites"
     echo "for more info."
+    exit 1
   fi
 }
 
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -197,6 +197,7 @@ Bootstrap() {
     echo "You will need to bootstrap, configure virtualenv, and run pip install manually."
     echo "Please see https://letsencrypt.readthedocs.org/en/latest/contributing.html#prerequisites"
     echo "for more info."
+    exit 1
   fi
 }
 


letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -500,6 +500,7 @@ Bootstrap() {
     echo "You will need to bootstrap, configure virtualenv, and run pip install manually."
     echo "Please see https://letsencrypt.readthedocs.org/en/latest/contributing.html#prerequisites"
```
pde: I wonder if we should also improve that error message?  Right now that link is (1) to readthedocs; (2) is a remnant of the ancient era when using.rst walked the user through venv creation.  Maybe https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client would be a better link?

bmw: That's outside the scope of this problem this PR is trying to fix (which continuing `certbot-auto` execution after failing to bootstrap, resulting in awful errors when installing/using the Python packages), however, given the size of this PR, I'm fine addressing it here if you prefer.

With that said, I don't think we actually have a good link in this case. The instructions you linked to tell the user to run `certbot-auto` in order to install OS packages (which just failed for them).

Perhaps we should try to put together a list of OS dependencies somewhere, however, this list varies based on your OS.

pde: I guess the reason I think they're related is that (in my original thinking when writing this code without an `exit`) if the user is going to have to hack things together manually, it's most helpful if the script keeps running and hits a real error that tells them what dependency they're missing, rather than a generic "bootstrap failed" one.  But that could well be wrong.

bmw: I like that approach in theory, but I'm not sure how well it works in practice. I haven't seen a lot of instances of it, but the comment that caused me to create this PR can be found [here](https://github.com/certbot/certbot/issues/2540#issue-136091238). The cause of the problem is not very clear.

pde:
OK

PR #3112, opened 2016-06-02
Fix ValueError: Invalid header value on OS X

federicobond:
The OS X user agent header contained a `\n` which didn't pass the validation:
````
'CertbotACMEClient/0.8.0 (darwin 10.11.5\n) Authenticator/standalone Installer/None'
````


federicobond:
A similar fix has been merged in #3118.

PR #3118, opened 2016-06-03
Strip "\n" from end of OS version string for OS X.

mrstanwell:
If you don't, it ends up in the UserAgent header and you get an error
like:
Invalid header value 'CertbotACMEClient/0.8.0 (darwin 10.10.5\n)...'

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -325,7 +325,7 @@ def get_python_os_info():
         os_ver = subprocess.Popen(
             ["sw_vers", "-productVersion"],
             stdout=subprocess.PIPE
-        ).communicate()[0]
+        ).communicate()[0].rstrip('\n')
     elif os_type.startswith('freebsd'):
         # eg "9.3-RC3-p1"
         os_ver = os_ver.partition("-")[0]


deltaepsilon:
@mrstanwell I've made thsi edit to ```certbot/util.py```, but I'm still getting the bug. I have no idea how to build Python... is there something that I need to do after editing ```certbot/util.py```?

The command that's failing for me is ```./certbot-auto certonly --manual --debug```.

mrstanwell:
@deltaepsilon I only know enough to be dangerous.  But `certbot-auto` updates itself and recreates a virtual env when you run it.  If you look at whatever `util.py` file you changed, I bet you'll see that your change isn't there anymore.  When I was trying to fix the file, I could prevent `certbot` from updating it by adding `--no-self-upgrade` to the cmd.

You could also use the instructions for ["Hacking" certbot](https://certbot.eff.org/docs/contributing.html#hacking) to check out [my fix branch](https://github.com/mrstanwell/certbot/tree/darwin-useragent-fix) and run `certbot` from it.  That also sidesteps the auto-upgrading.

deltaepsilon:
Hooray! I got it working. I had to check out your branch and use the --no-self-upgrade flag, and it magically worked. Just editing the file didn't work for me, and I never saw certbot revert my change... it just ignored it.

schoen:
I just looked into this bug and concluded that this exact fix was necessary, only to discover that there is already a PR that makes it.

Fixes #3120.

Thanks, @mrstanwell.

shallad:
I am still getting this error in Mac OSX 10.11.4.
I tried it today for the first time and was following https://in.godaddy.com/help/install-a-lets-encrypt-ssl-apache-20245. 
Also , I verified the certbot/util.py which had the fix.

schoen:
@shallad, did you previously install a version of `certbot` or `letsencrypt` in some other way?

shallad:
@schoen This is the first time I was trying to install it.

mrstanwell:
@shallad: If an earlier run of `certbot` installed a venv under `~/.local/share/letsencrypt`, that might be what it's picking up.  Or, running `certbot` without the `--no-self-upgrade` might be causing it to pull down the tagged code from github, not `master` (where the fix now is).

You might try removing  `~/.local/share/letsencrypt` if it exists.  Then check out `master` and try running it following the [hacking certbot instructions](https://certbot.eff.org/docs/contributing.html#hacking), which should ensure you're running out of the latest files.

mrstanwell:
Yeah, I just verified that something like this works for me:
```
    $ rm -fr ~/.local/share/letsencrypt
    $ cd /tmp
    $ git clone https://github.com/certbot/certbot.git
    $ cd certbot
    $ ./certbot-auto --no-self-upgrade --debug renew --dry-run
```
Now, I have existing certs that I could do a renew on, so that last command might not be applicable to you.  But using `certbot-auto` with `--no-self-upgrade --debug` should get it to work with whatever action you're trying to perform.  (The `--debug` is required on OS X.)

Andersos:
I still couldn't get it working so I ended up doing the following:
Open this file `~/.local/share/letsencrypt`
Change line 55 from:
`ua = "CertbotACMEClient/{0} ({1}) Authenticator/{2} Installer/{3}"`
to
`ua = "CertbotACMEClient/{0} () Authenticator/{2} Installer/{3}"`

mdridley:
Which file Andersos?

Andersos:
You are right. The file was: `lib/python2.7/site-packages/certbot/client.py`.
So full path would be if I'm not mistaken: `~/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/client.py`

PR #3126, opened 2016-06-05
Improve debug logs.

jsha:
Print request and response bodies with newlines, rather than all on one line.
Remove "Omitted empty field" log, which gets logged meaninglessly for every JSON
serialization.
Remove duplicated logging of responses.
Log the base64 version of the nonce, rather than turning it into bytes and
logging the backslash-escaped version of those bytes.
Only pass -vv in tests.


jsha:
Ping?

jsha:
BTW, as a side question: In Boulder's test.js we color-code debug output into request, response, and payload categories. Is that potentially interesting for certbot, or do you want to keep the debug log free of escape codes?

Also, the JWS logging tends to be less interesting than the payload logging. Wondering if we should omit it or bump it to a higher verbosity level.

Neither of the above should be considered a blocker on this change though. We can do them as an add-on.

pde:
Ooops sorry wrong ticket, Brad should probably take this one and I'll pass take the other.

jsha:
Looks like I accidentally closed this back in June. @bmw @ohemorange could one of you review?

ohemorange:
I see no problems, could you merge in master since this is an older commit?

jsha:
Done. Also as an FYI, this results in logging the DER-encoded certs as raw bytes, which shows up in `less` as seen here:

screenshot from 2016-10-24 17-35-35

IMO, this is okay. However, if you feel like we should have 7-bit clean ASCII in the logs, I can add some extra code to base64 bodies of certain Content-Types before logging them.

ohemorange:
Is there any reason not to base64 this bit?

acme/acme/client.py
```diff
@@ -631,7 +635,7 @@ def _add_nonce(self, response):
                 decoded_nonce = jws.Header._fields['nonce'].decode(nonce)
             except jose.DeserializationError as error:
                 raise errors.BadNonce(nonce, error)
-            logger.debug('Storing nonce: %r', decoded_nonce)
+            logger.debug('Storing nonce: %s', nonce)
```
jsha: The nonce isn't really an "object" to be decoded. It's just a string of random bytes, base64 encoded. So the decoded values are not useful.

jsha:
> Is there any reason not to base64 this bit?

Half laziness, half not being sure how best to express to the user that those weren't the literal bytes returned from the server. I'll just go ahead and base64 it though.

acme/acme/client.py
```diff
@@ -521,10 +523,11 @@ def _wrap_in_jws(self, obj, nonce):
         :rtype: `.JWS`
 
         """
-        jobj = obj.json_dumps().encode()
-        logger.debug('Serialized JSON: %s', jobj)
+        jobj = obj.json_dumps(indent=2).encode()
+        logger.debug('JWS payload:\n%s', jobj)
```
ohemorange: for extra fancy points: `os.linesep`

we do use both in the codebase though so feel free to ignore this.

jsha: https://docs.python.org/2/library/os.html
> Do not use os.linesep as a line terminator when writing files opened in text mode (the default); use a single '\n' instead, on all platforms.

Issue #1173, opened 2015-10-28
Install/compile fails of letsencrypt-auto on Smartos/Illumos


fire:
The auto method fails because a script was not written.

Can't seem to find Augeas in the package manager. It may be possible to install letsencrypt client manually.

Seems avaliable since Smartos pkgsrc 2015Q1

Will try to target 2015Q3

kuba:
Augeas is only necessary for `apache` plugin - maybe we could simply not install apache plugin in case of you platform?

fire:
How do you disable the plugin?

fire:
I was able to get the client working after I disabled the Apache plugin and the tests. 

The python on Smartos 14.4.0 is named python2.7. So it had to be renamed from python2.

kuba:
With #1257 you will be able to call `LE_PYTHON=python2.7 letsencrypt-auto`.

If you can provide autodetection for your OS in `letsencrypt-auto` script and set `LE_PYTHON` appropriately, that would be awesome.

fire:
Is that patch in the master branch yet?

kuba:
#1257 has not yet been merged to `master` (it's still green, rather than purple)

fire:
Is there a better way of disabling the apache plugin than just disabling the lines? in letsencrypt-auto?

kuba:
What do you mean by disabling it? Why do you want to do that?

fire:
> Augeas is only necessary for apache plugin - maybe we could simply not install apache plugin in case of your platform?

Is there a non invasive way of doing this?

kuba:
Oh, I'm really sorry; lost context a bit! Does running `letsencrypt-auto` fail if Augeas (and other apache deps) are not installed?

fire:
> Oh, I'm really sorry; lost context a bit! Does running letsencrypt-auto fail if Augeas (and other apache deps) are not installed?

Yes.

Edgewise:
I got this working on my smartos server running an nginx reverse proxy. I have python2.7 installed and the script recognized it correctly (I just started/pulled today). You might try pulling the latest source and trying again.

I used the following steps:
* commented out the two lines in the letsencrypt-auto file that contain 'letsencrypt-apache'. 
* installed virtualenv and pip using 'pkgin in py27-virtualenv-1.11.6 py27-pip-1.5.6'.
* used the standalone option from the beta invite email.
* updated my nginx conf file with the new certificates

[...]

PR #3128, opened 2016-06-06
Initialize Augeas in a different method to be able to react to ImportError

joohoi:
Handle augeas initialization in a separate method to be able to call it from inside prepare(), and react to the possible ImportError by reporting Apache plugin unavailable and letting user to choose out of the other possibilities.

Mitigates #1173


SwartzCr:
looks good to me

PR #1742, opened 2015-12-04
mageia bootstrap [needs revision]

olivluca:
I added a simple script in bootstrap to use letsencrypt-auto with mageia.
I tested it in mageia 5 with the webroot plugin.

PR #3133, opened 2016-06-08
Mageia Bootstrap

bmw:
Brings in #1742 and puts the bootstrap script behind a call to `ExperimentalBootstrap`. There were a number of +1s on #1742 and I tested this successfully in Docker and on the test farm.


pde:
(Merging an ExperimentalBootstrap()er on the basis of various +1s)

PR #3138, opened 2016-06-08
Remove proxy in certbot-compatibility-test

bmw:
This is the work @schoen and I did on `certbot-compatibility-test`. The tests now work completely inside of Docker and we were able to get all tests to pass using the configurations provided in `certbot-compatibility-test/certbot_compatibility_test/testdata/config.tar.gz`.

Future steps:
* Write a Dockerfile to set up the docker image to run these tests.
* Get the tests running in Travis.
* Add `nginx` support.

Opening this PR to try and keep PRs on the tests small and easily reviewable.


schoen:
It's quite possible that there are other bugs in the compatibility test that we'll find when we start making more use of it, but I think this is a strict improvement and makes it appear to work as far as we know right now. I'm planning to use some of this in trying to get the same thing to work for nginx.

PR #3143, opened 2016-06-10
Preserve common name during renewal

### LINKED ISSUE #3127: Renew changes common name
I created a multi-domain certificate using certbot certonly -d mydomain -d a.mydomain -d b.mydomain which had the common name mydomain. After running certbot renew it renewed the certificate but made a.mydomain the common name.

I don't know if this is a blocker for any use case, it certainly works for me. But I would like the renewed certificates to keep the old common name. It especially makes it easier to work with multiple certificates if you can map the common name of the certificate to the entry in /etc/letsencrypt/live/.

bmw:
Fixes #3127.


schoen:
LGTM.

PR #3144, opened 2016-06-10
Adding sensible UI logging for typical user

### LINKED ISSUE #599: Sensible UI logging for typical user
There currently isn't any output that provides reasonable detail about the underlying process.

-v is the closest, but it suffers from informing the user about every single HTTPS connection and could use some general cleanup.

It would be nice to have a mode that was similar to the display seen in the Ncurses demo.  https://youtu.be/Gas_sSB-5SU?t=51

TheNavigat:
Closes #599 

There are 2 rules I followed when adjusting the logging level of certain messages.

First, I refrained from changing messages from info to warning (although some messages seemed like they belong more to warning, but that could be done in another PR).

And second, I tried to make it as possible to what a normal user would understand, without the messages being pedantic.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -282,7 +282,7 @@ class Client(object):
                 "by your operating system package manager")
 
         if self.config.dry_run:
-            logger.info("Dry run: Skipping creating new lineage for %s",
+            logger.debug("Dry run: Skipping creating new lineage for %s",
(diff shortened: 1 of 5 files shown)


bmw:
LGTM.

Thanks for the PR @TheNavigat.

Issue #3154, opened 2016-06-11
certbot-auto not quiet when used with regular user


gdoucet:
When certbot-auto is call from a regular user crontab, certbot-auto is not quiet even when used with --quiet.

The following is print to the script even when with the --quiet argument:
$ ./certbot-auto renew --quiet
Requesting root privileges to run certbot...
  /home/gdoucet/.local/share/letsencrypt/bin/letsencrypt renew --quiet

bmw:
fixed in #3155

PR #3155, opened 2016-06-11
Added the argument --quiet and -q so then when used with a regular user there is no output to the screen.

gdoucet:
Pull request to fix fix issue #3154.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -34,6 +34,7 @@ Help for certbot itself cannot be provided until it is installed.
   -n, --non-interactive, --noninteractive   run without asking for user input
   --no-self-upgrade                         do not download updates
   --os-packages-only                        install OS dependencies and exit
+  -q, --quiet                               provide only update/error output
   -v, --verbose                             provide more output
 
 All arguments are accepted and forwarded to the Certbot client when run."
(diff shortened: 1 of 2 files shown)


bmw:
What's the use case here? You've configured `sudo`/`su` to not prompt for a password? What's the benefit of doing this vs. configuring `cron` to run `certbot-auto` as `root`?

gdoucet:
I know it is possible to use run certbot-auto as root, but on some system the application support team don't have access to root or the crontab from root.

pde:
We should either merge this or [fix our renewal instructions](https://certbot.eff.org/#ubuntutrusty-apache) to specify that that has to be in root's crontab.  I think merging is preferable.

gdoucet:
Their is really no technical reason that certbot-auto need to run as root, because certbot-auto automatically detect it is a regular user and execute the proper command with sudo.

certbot-auto
```diff
@@ -34,7 +34,8 @@ Help for certbot itself cannot be provided until it is installed.
   -n, --non-interactive, --noninteractive   run without asking for user input
   --no-self-upgrade                         do not download updates
   --os-packages-only                        install OS dependencies and exit
-  -v, --verbose                             provide more output
+  -q, --quiet                               provide no output
```
bmw: Output is still provided when `certbot-auto` upgrades itself. What if we changed the flags description to something like "provide only update/error output" (`certbot` itself will provide error output when given `-q`)?

bmw:
After talking to pde about this out of band, I'm fine adding this. I previously implemented this exact thing in #3015 but decided against it and instead wrote up #3016.

With that said, I'm fine moving forward this PR. I made a small suggestion about the help text for the `--quiet` flag.

Additionally, while perhaps a little overly complex, these changes should actually be made to `letsencrypt-auto-source/letsencrypt-auto.template`. After making changes to this file, `build.py` should be run in the same directory which will modify `letsencrypt-auto-source/letsencrypt-auto`. We only make changes to `certbot-auto` and `letsencrypt-auto` during releases so all changes to these files should be removed.

After making these changes, remove the "[needs revision]" from the title and we'll review again. Thanks @gdoucet for the PR and holding your ground convincing us it's a good idea.

gdoucet:
@bmw thanks for the instructions. I have change my commit and also put your usage suggestion for the --quiet argument.

bmw:
LGTM. Thanks for the PR @gdoucet.

PR #3160, opened 2016-06-13
Add additional warning with actual exception message during renewal

lwolf:
Log and show warning with real exception message to make it more clear what exactly happened.
Currently it shows `config is broken` when in fact we have broken symlinks in live folder.

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -60,7 +60,8 @@ def _reconstitute(config, full_path):
     try:
         renewal_candidate = storage.RenewableCert(
             full_path, configuration.RenewerConfiguration(config))
-    except (errors.CertStorageError, IOError):
+    except (errors.CertStorageError, IOError) as exc:
+        logger.warning(exc)
         logger.warning("Renewal configuration file %s is broken. Skipping.", full_path)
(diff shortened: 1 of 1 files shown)


certbot/renewal.py
```diff
@@ -60,7 +60,8 @@ def _reconstitute(config, full_path):
     try:
         renewal_candidate = storage.RenewableCert(
             full_path, configuration.RenewerConfiguration(config))
-    except (errors.CertStorageError, IOError):
+    except (errors.CertStorageError, IOError) as exc:
+        logger.warning(exc)
         logger.warning("Renewal configuration file %s is broken. Skipping.", full_path)
         logger.debug("Traceback was:\n%s", traceback.format_exc())
```
pde: The error is already being logged at `debug` level below, which means you can find the traceback in `/var/log/letsencrypt.log` or by adding `-vv`.  I'm reluctant to show raw Python tracebacks at the default logging level, because they will confuse and intimidate some users...

pde:
Hmmm, I guess my previous comment was wrong; logging the exception will actually produce a human-readable message rather than a full traceback.

PR #3164, opened 2016-06-13
Provide nonroot guidance when logging gets EACCES.

### LINKED ISSUE #2306: Error that results when run without root is unclear
When running `letsencrypt` without root, I get

```
An unexpected error occurred:
OSError: [Errno 13] Permission denied: '/etc/letsencrypt'
Please see the logfile 'letsencrypt.log' for more details.
```

The underlying reason is that, to run without root, I need to specify `--config-dir`, `--work-dir`, and `--logs-dir`. However, the client doesn't give me any suggestion about this or about the possibility of doing anything (even reading help) without root access.

While a sysadmin would probably conclude that the reason for the `OSError` is simply that the client needs root to run, it would be great to have a clearer and more useful error message in this case.

jsha:
Fixes #2306

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -1,6 +1,7 @@
 """Certbot main entry point."""
 from __future__ import print_function
 import atexit
+import errno
 import functools
 import logging.handlers
 import os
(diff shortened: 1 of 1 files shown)


pde:
LGTM.

PR #3168, opened 2016-06-13
Fix default detection

### LINKED ISSUE #3153: Interactive webroot values not stored in renewal config file
This is due to using mutable values as the default in `argparse` such as `list` and `dict` and then reaching inside `argparse` to check the defaults in `storage.py`. The result is modifications to these values also modify the default, meaning they always compare equal to the default value and `storage.py` thinks they should not be saved.

bmw:
Fixes #3153.

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -162,30 +162,16 @@ def relevant_values(all_values):
 
     from certbot import cli
 
-    def _is_cli_default(option, value):
-        # Look through the CLI parser defaults and see if this option is
-        # both present and equal to the specified value. If not, return
-        # False.
-        # pylint: disable=protected-access
-        for x in cli.helpful_parser.parser._actions:
-            if x.dest == option:
-                if x.default == value:
-                    return True
-                else:
-                    break
-        return False
-
     values = dict()
     for option, value in all_values.iteritems():
         # Try to find reasons to store this item in the
         # renewal config.  It can be stored if it is relevant and
-        # (it is set_by_cli() or flag_default() is different
-        # from the value or flag_default() doesn't exist).
+        # (it is set_by_cli(), we don't know the default value, or
+        # the current value differs from the default value).
         if _relevant(option):
-            if (cli.set_by_cli(option)
-                or not _is_cli_default(option, value)):
-#                or option not in constants.CLI_DEFAULTS
-#                or constants.CLI_DEFAULTS[option] != value):
+            if (cli.set_by_cli(option) or
+                    option not in cli.DEFAULTS or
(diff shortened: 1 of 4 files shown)


pde:
Should `set_by_cli` return true for things that are actually set interactively? It seems that it possibly should (and should perhaps be renamed accordingly?).

bmw:
So I could move the logic that I added to `storage.py` to `set_by_cli`, but something like the `DEFAULTS` mechanism is still necessary since the dummy arguments used by `set_by_cli` aren't passed through the entire program.

pde:
I understand that we'll still need `DEFAULTS`, it just seems that we should aim to have one code path for detecting whether variables have their default values, rather than (if I'm reading it correctly) two variants of this logic.

certbot/storage.py
```diff
@@ -162,30 +162,16 @@ def relevant_values(all_values):
...
+        # (it is set_by_cli(), we don't know the default value, or
+        # the current value differs from the default value).
         if _relevant(option):
-            if (cli.set_by_cli(option)
-                or not _is_cli_default(option, value)):
-#                or option not in constants.CLI_DEFAULTS
-#                or constants.CLI_DEFAULTS[option] != value):
+            if (cli.set_by_cli(option) or
+                    option not in cli.DEFAULTS or
+                    cli.DEFAULTS[option] != value):
```
pde: I think it would be fine to wrap these two boolean tests in a simple `cli.is_default` function and merge.

bmw: I may have misunderstood you. I thought you wanted a common function in `cli.py`. Either way, I'm almost done.

certbot/tests/storage_test.py
```diff
@@ -533,10 +533,9 @@ def test_relevant_values_default(self, mock_parser):
         """Test that relevant_values() can reject a default value."""
         # pylint: disable=protected-access
         from certbot import storage
-        mock_parser.verb = "certonly"
         mock_parser.args = ["--standalone"]
-        mock_action = mock.Mock(dest="rsa_key_size", default=2048)
-        mock_parser.parser._actions = [mock_action]
+        mock_parser.defaults = {"rsa_key_size": 2048}
```
TheNavigat: What about this? https://github.com/certbot/certbot/blob/f9d5ecaf6fc6776988ddf5ea9dbb010ad45eeb7f/certbot/tests/storage_test.py#L548

bmw: Nice catch.

I rewrote all `relevant_values` test in my latest commit.

TheNavigat:
LGTM

pde:
LGTM

PR #3174, opened 2016-06-15
Log new cert and cert renewal

### LINKED ISSUE #3166: Log whether renewing or obtaining a new certificate
It appears we don't log whether we're renewing an existing certificate or obtaining a new one. We should do this as it makes our logs more useful for debugging.

jsachs:
Closes #3166

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -88,9 +88,11 @@ def _auth_from_domains(le_client, config, domains, lineage=None):
     hooks.pre_hook(config)
     try:
         if action == "renew":
+            logger.info("Renewing an existing certificate")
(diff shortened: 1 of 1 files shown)


bmw:
On line 86 where we return if `action == "reinstall"` it'd be nice to log the decision there as well. We should probably use wording like "Keeping the existing certificate" because this code is used in both the `certonly` and `run` case and the certificate is not installed when using `certonly`.

Other than that though, this LGTM. Thanks for the PR @jsachs. I'm adding "[needs revision]" to the title of this PR. Please remove it when you'd like another review.

bmw:
LGTM. Thanks for the PR @jsachs.

PR #3176, opened 2016-06-15
Printing pip output to terminal when -v is used

### LINKED ISSUE #2835: pip and verbosity
A user reported that `letsencrypt-auto` hang when reaching the "Installing Python packages...` step. Currently, there is no way for them to see the output of `pip`, even when `-v/--verbose` is provided on the command line.

`-v/--verbose` should allow users to see `pip` output for more information/easier debugging.

TheNavigat:
Closes #2835

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -247,13 +247,19 @@ UNLIKELY_EOF
     # Set PATH so pipstrap upgrades the right (v)env:
     PATH="$VENV_BIN:$PATH" "$VENV_BIN/python" "$TEMP_DIR/pipstrap.py"
     set +e
-    PIP_OUT=`"$VENV_BIN/pip" install --no-cache-dir --require-hashes -r "$TEMP_DIR/letsencrypt-auto-requirements.txt" 2>&1`
+    if [ "$VERBOSE" = 1 ]; then
+      "$VENV_BIN/pip" install --no-cache-dir --require-hashes -r "$TEMP_DIR/letsencrypt-auto-requirements.txt" 2>&1
+    else
+      PIP_OUT=`"$VENV_BIN/pip" install --no-cache-dir --require-hashes -r "$TEMP_DIR/letsencrypt-auto-requirements.txt" 2>&1`
+    fi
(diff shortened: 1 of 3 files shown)


bmw:
@pde or @schoen can you take a look at this? I'm not sure how portable the `tee /dev/tty` approach is and I think both of you are better at bash hackery than I am.

schoen:
@bmw, `tee /dev/tty` is relatively portable, but where is the output going otherwise that prevents the user from seeing it? How about just `>&2` or simply not doing `2>&1` in this case?

bmw:
Good point @schoen!

@TheNavigat, with that in mind I propose the following:

Only use `PIP_OUT` in the non-verbose case. Then in the `if [ "$PIP_STATUS" != 0 ]` block that directly following this code, replace the colon at the end of this line "Had a problem while installing Python packages:" with a period. Then add another if block, only echoing PIP_OUT if verbose wasn't set.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -247,13 +247,19 @@ UNLIKELY_EOF
     # Set PATH so pipstrap upgrades the right (v)env:
     PATH="$VENV_BIN:$PATH" "$VENV_BIN/python" "$TEMP_DIR/pipstrap.py"
     set +e
-    PIP_OUT=`"$VENV_BIN/pip" install --no-cache-dir --require-hashes -r "$TEMP_DIR/letsencrypt-auto-requirements.txt" 2>&1`
+    if [ "$VERBOSE" = 1 ]; then
+      "$VENV_BIN/pip" install --no-cache-dir --require-hashes -r "$TEMP_DIR/letsencrypt-auto-requirements.txt" 2>&1
```
bmw: We probably should remove `2>&1` here.

bmw:
Other than my nit about redirection, only other comment is to make sure to run `build.py` after changing the template so the change gets included in the `letsencrypt-auto-source/letsencrypt-auto` script

bmw:

bmw:
Looks like the `letsencrypt-auto` tests failed.

bmw:
LGTM.

Issue #2813, opened 2016-04-11
Dialog error messages


bmw:
I haven't taken the time to dig into this, but `letsencrypt` loses the content of dialog error messages. I caused an error by making my terminal too small to display the `dialog` widget (see #2812). When doing this, the output to my terminal was:

```
An unexpected error occurred:
DialogError
```

The contents of my logfile were:

```
2016-04-11 18:31:36,104:DEBUG:letsencrypt.main:Root logging level set at 30
2016-04-11 18:31:36,105:INFO:letsencrypt.main:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2016-04-11 18:31:36,105:DEBUG:letsencrypt.main:letsencrypt version: 0.6.0.dev0
2016-04-11 18:31:36,105:DEBUG:letsencrypt.main:Arguments: []
2016-04-11 18:31:36,106:DEBUG:letsencrypt.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#nginx,PluginEntryPoint#standalone,PluginEntryPoint#manual,PluginEntryPoint#webroot,PluginEntryPoint#apache,PluginEntryPoint#null)
2016-04-11 18:31:36,109:DEBUG:letsencrypt.plugins.selection:Requested authenticator None and installer None
2016-04-11 18:31:36,430:DEBUG:letsencrypt.plugins.disco:No installation (PluginEntryPoint#nginx): 
Traceback (most recent call last):
  File "/root/letsencrypt/letsencrypt/plugins/disco.py", line 104, in prepare
    self._initialized.prepare()
  File "/root/letsencrypt/letsencrypt-nginx/letsencrypt_nginx/configurator.py", line 115, in prepare
    raise errors.NoInstallationError
NoInstallationError
2016-04-11 18:31:36,430:DEBUG:letsencrypt.plugins.selection:Single candidate plugin: * apache
Description: Apache Web Server - Alpha
Interfaces: IAuthenticator, IInstaller, IPlugin
Entry point: apache = letsencrypt_apache.configurator:ApacheConfigurator
Initialized: <letsencrypt_apache.configurator.ApacheConfigurator object at 0x7fc63206bc10>
Prep: True
[...]

PR #3177, opened 2016-06-15
Set dialog widgets to use autowidgetsize

### LINKED ISSUE #2812: Using dialog in small terminals
We have hard coded that the dialog window be 24 characters tall and 72 characters wide in `letsencrypt/display/util.py`. If the terminal is smaller than that, `dialog` raises an exception. This exception may be unnecessary because the requested `dialog` widget may fit in a terminal smaller than 24 x 72.

`dialog` has the ability to [autosize widgets](http://pythondialog.sourceforge.net/doc/Dialog_class_overview.html#autowidgetsize) which we might want to use. If we take this route, we should do the following:

1. Make sure this works on all system with both new and old version of `dialog`.
2. Make sure we're happy with how the autosized widgets look.

jsachs:
Closes #2812 and #2813

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -57,7 +57,7 @@ class NcursesDisplay(object):
 
     def __init__(self, width=WIDTH, height=HEIGHT):
         super(NcursesDisplay, self).__init__()
-        self.dialog = dialog.Dialog()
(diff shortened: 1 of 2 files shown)


pde:
I haven't been able to reproduce either of these underlying bugs; for me the `master` branch seems to be happy running in a tiny terminal window...

PR #3182, opened 2016-06-16
Handle commas in apache filenames (Issue 2983)

### LINKED ISSUE #2983: certbot-apache fails to parse files with comma in the filename
Original issue was: #2949

Apache plugin errors out when trying to read files with comma in the filename.

```
letest@letest:/etc/apache2/sites-available$ ls -la
total 48
drwxr-xr-x 2 root root 4096 May 13 11:35 .
drwxr-xr-x 8 root root 4096 May 13 11:29 ..
-rw-r--r-- 1 root root 1878 May 10 13:36 000-default-le-ssl.conf
-rw-r--r-- 1 root root 1472 May 10 00:30 000-default.conf
-rw-r--r-- 1 root root  255 May 10 13:51 1.conf,OLD.conf
```

Running ```letsencrypt --apache --staging``` results in:

```
letest@letest:~$ sudo letsencrypt --apache --staging
An unexpected error occurred:
RuntimeError: Error during match procedure!
```
With stack trace in the log file:
```
2016-05-13 08:35:32,962:DEBUG:letsencrypt.main:Root logging level set at 30
2016-05-13 08:35:32,963:INFO:letsencrypt.main:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2016-05-13 08:35:32,963:DEBUG:letsencrypt.main:letsencrypt version: 0.5.0
2016-05-13 08:35:32,963:DEBUG:letsencrypt.main:Arguments: ['--apache', '--staging']
2016-05-13 08:35:32,964:DEBUG:letsencrypt.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2016-05-13 08:35:32,967:DEBUG:letsencrypt.plugins.selection:Requested authenticator apache and installer apache
2016-05-13 08:35:33,292:DEBUG:letsencrypt.main:Exiting abnormally:
  File "/usr/bin/letsencrypt", line 9, in <module>
    load_entry_point('letsencrypt==0.5.0', 'console_scripts', 'letsencrypt')()
  File "/usr/lib/python2.7/dist-packages/letsencrypt/main.py", line 692, in main
  File "/usr/lib/python2.7/dist-packages/letsencrypt/main.py", line 448, in run
    installer, authenticator = plug_sel.choose_configurator_plugins(config, plugins, "run")
  File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/selection.py", line 184, in choose_configurator_plugins
    authenticator = installer = pick_configurator(config, req_inst, plugins)
  File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/selection.py", line 24, in pick_configurator
  File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/selection.py", line 76, in pick_plugin
    verified.prepare()
  File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/disco.py", line 214, in prepare
    return [plugin_ep.prepare() for plugin_ep in self._plugins.itervalues()]
  File "/usr/lib/python2.7/dist-packages/letsencrypt/plugins/disco.py", line 104, in prepare
    self._initialized.prepare()
  File "/usr/lib/python2.7/dist-packages/letsencrypt_apache/configurator.py", line 177, in prepare
    self.vhosts = self.get_virtual_hosts()
  File "/usr/lib/python2.7/dist-packages/letsencrypt_apache/configurator.py", line 552, in get_virtual_hosts
    new_vhost = self._create_vhost(path)
  File "/usr/lib/python2.7/dist-packages/letsencrypt_apache/configurator.py", line 505, in _create_vhost
    args = self.aug.match(path + "/arg")
  File "/usr/lib/python2.7/dist-packages/augeas.py", line 312, in match
    raise RuntimeError("Error during match procedure!")
RuntimeError: Error during match procedure!
```

This looks like a bug in Augeas, so pinging @domcleal

SwartzCr:
fixes #2983


certbot-apache/certbot_apache/configurator.py
```diff
@@ -511,7 +511,11 @@ def _create_vhost(self, path):
 
         """
         addrs = set()
-        args = self.aug.match(path + "/arg")
+        try:
+            args = self.aug.match(path + "/arg")
+        except RuntimeError:
```
joohoi: I think this could be triggered in various different occasions, so maybe more generic error message would be in order. Also it'd be best to print out the path that caused the error for the user to notify exactly what caused the breakage.

joohoi:
Possibly printing the error with problematic filename etc. to logs and continuing the execution would be better in terms of UX instead of erroring out.

domcleal:
Does `get_file_path` a few lines lower still work with escaped paths if the filename contains a comma (on a new version of Augeas)?  I think the path needs unescaping with:

    filename = get_file_path(self.aug.aug_get("/augeas%s/path" % path))

to get the real filesystem path.

SwartzCr:
@joohoi I'd be happy to add a print of the path to the logs, what addition to the error try should I add
@domcleal I'd be happy to test that, do you know what systems have augeas 1.4?

domcleal:
@SwartzCr EL7, Ubuntu 16.04+, Fedora 21+ are the main ones I know of.

joohoi:
Something along the lines of:

logger.warn("Encountered a problem while parsing file: %s, skipping" % path)

The message would then cover also the RuntimeErrors we haven't possibly encountered yet. Also letting the user know the specific file, and that it has been skipped due to the error is beneficial.

SwartzCr:
exciting!
```
We were unable to find a vhost with a ServerName or Address of whatdo.info.
Which virtual host would you like to choose?
(note: conf files with multiple vhosts are not yet supported)
1: old\,default.conf              |                       |       |
2: 000-default.conf               |                       |       | Enabled
3: default-ssl.conf               |                       | HTTPS |
Select the appropriate number [1-3] then [enter] (press 'c' to cancel): 1

We were unable to find a vhost with a ServerName or Address of whatdo.info.
Which virtual host would you like to choose?
(note: conf files with multiple vhosts are not yet supported)
1: old\,default.conf              |                       |       |
2: 000-default.conf               |                       |       | Enabled
3: default-ssl.conf               |                       | HTTPS |
Select the appropriate number [1-3] then [enter] (press 'c' to cancel): 1
2016-06-21 16:17:23,829:CRITICAL:certbot_apache.configurator:Error writing/reading to file in make_vhost_ssl
2016-06-21 16:17:23,831:WARNING:certbot.reverter:File: 
 - Could not be found to be deleted /etc/apache2/sites-available/old\,default-le-ssl.conf - LE probably shut down unexpectedly
Unable to write/read in make_vhost_ssl

IMPORTANT NOTES:
   /etc/letsencrypt/live/whatdo.info/fullchain.pem. Your cert will
   expire on 2016-09-19. To obtain a new or tweaked version of this
   certificate in the future, simply run certbot again with the
   "certonly" option. To non-interactively renew *all* of your
   certificates, run "certbot renew"
 - If you lose your account credentials, you can recover through
   e-mails sent to n@eff.org.
 - Your account credentials have been saved in your Certbot
   configuration directory at /etc/letsencrypt. You should make a
   secure backup of this folder now. This configuration directory will
   also contain certificates and private keys obtained by Certbot so
   making regular backups of this folder is ideal.

```

SwartzCr:
@domcleal is there a good way get the path back in an unescaped way? Currently I'm having the problem where python is choking on the file paths because the commas don't need escaping in python.  So for example it writes the backslashes into the apache conf file and then conftest fails because it can't find that file.
But it seems that augeas still needs the backslashes for matching later - is there an easy way to interoperate, or should I just strip backslashes whenever it goes into python land?

domcleal:
@SwartzCr yes, please see my comment at https://github.com/certbot/certbot/pull/3182#issuecomment-227066198 - you can retrieve the unescaped path with `aug.get('/augeas/files/etc/escaped_thing/path')`.

SwartzCr:
hmm that's totally not returning anything for me - calling it with this invocation in the debugger and getting a blank back:
`self.aug.get('/augeas/files/etc/apache2/sites-enabled/000-default.conf/VirtualHost/path')`

domcleal:
Ah, sorry - the get_file_path and aug.get calls might be the other way around.

> self.aug.get('/augeas/files/etc/apache2/sites-enabled/000-default.conf/VirtualHost/path')

This shouldn't have the /VirtualHost bit, it should just be the filename portion, so: `self.aug.get('/augeas/files/etc/apache2/sites-enabled/000-default.conf/path')`.  If you call get_file_path first then it'll probably give you the right path to look up in Augeas.

SwartzCr:
hmm coverage is failing on travis but passing locally @bmw any idea what's going on?

bmw:
Really? Tests fail locally for me. It also seems correct that they would fail. Almost all of the tests have to do with the number of vhosts the plugin finds and you're adding one to the directory.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -948,6 +954,12 @@ def _add_name_vhost_if_necessary(self, vhost):
         if need_to_save:
             self.save()
 
+    def _unescape(self, fp):
+        return fp.replace("\\", "")
```
domcleal: Why use this instead of getting the path from `/augeas`?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -948,6 +954,12 @@ def _add_name_vhost_if_necessary(self, vhost):
...
+
+    def _escape(self, fp):
+        return fp.replace(",", "\\,")
```
domcleal: It may be worth leaving a comment here that this is (partially) equivalent to aug_escape_name, which is in 1.4.0 but not yet added to python-augeas. In future it could be swapped out, which'd make it reliable and correct for some other characters (such as `[]`).

bmw:
So I'm a bit worried about whether or not we're escaping/unescaping everything properly. My understanding is we want the escaped path when calling augeas methods, but otherwise, always want the unescaped path.

If this is correct, perhaps we should try and enforce this. Can we create a clear division to where things are escaped and where they aren't? Perhaps this could be done using wrappers around augeas methods where inputs are escaped and outputs are unescaped. Maybe this isn't possible and some configurator methods also have to deal with escaping/unescaping the path.

This will likely be cumbersome, but I'm worried we'll fail to (un)escape a path somewhere and everything will blow up.

bmw:
Steps to reproduce one example of not unescaping correctly:

1. On a system with Augeas 1.4+, create an HTTP vhost with a comma in the file name.
2. Use Certbot using the `run` subcommand (e.g. not `certonly`).
3. When prompted, select the vhost with a comma in the filename.
4. Get told "we only support configurations with one vhost per file".

Why does this happen? In [make_vhost_ssl](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L750), avail_fp and ssl_vp are unescaped paths. We then pass this path back into Augeas, `match` returns an empty list, and since the length of the list isn't 1, we assume it's greater than 1, and output the "...vhost per file" error message.

I'm unfortunately strongly in favor of reading through the entire plugin, reasoning about where paths should and should not be escaped, and creating a system to ensure this happens.

SwartzCr:
@domcleal So I understand that I can get an unescaped path from augeas, but is there a way to pass back an unescaped path to augeas and have it not blow up when looking for that file. We're passing things back and forth between augeas and python and are having the problem that python wants them to be unescaped (to do things like write them into apache configs) and augeas wants them to be escaped, (to do things like find or match them)
any ideas?

SwartzCr:
@domcleal or maybe one option would be to add aug_escape_name to the python implementation of augeas and wrap things leaving augeas with self.aug.get(get_file_path('/augeas/files%s/path' % path)) and aug_escape_name(vhost.filep) - does that sound right?
@bmw do you think it'd be okay to ship out own version of python augeas?

bmw:
@SwartzCr, I don't have a problem with it, but I think this is a good question for Harlan and James on the call tomorrow.

domcleal:
> So I understand that I can get an unescaped path from augeas, but is there a way to pass back an unescaped path to augeas and have it not blow up when looking for that file.

You can perhaps rewrite queries to use an expression like: `aug.match('/files/etc/apache/sites-available/*[label()="a,b.conf"]')` but that's quite a bit of effort as you'd have to split the path up and construct a complex-looking query.

You can keep and use your existing `_escape` implementation at https://github.com/certbot/certbot/pull/3182/files#diff-bdd584c1dc4f4d6316df485617bec545R1003, which I think is your best option.  Additional special characters are `][|/=()!` (`name_follow[]`, https://github.com/hercules-team/augeas/blob/master/src/pathx.c#L126-L133).

> or maybe one option would be to add aug_escape_name to the python implementation of augeas and wrap things leaving augeas with self.aug.get(get_file_path('/augeas/files%s/path' % path)) and aug_escape_name(vhost.filep) - does that sound right?

Yes, nearly - aug_escape_name will also escape `/` according to https://github.com/hercules-team/augeas/blob/master/src/augeas.h#L373-L388, so you'd split the path by `/` and then pass each component through aug_escape_name().

I'd just stick with your existing _escape implementation and use that when you want to re-escape a path to use in an Augeas lookup.

SwartzCr:
@domcleal that's really good feedback. Do you think it's better to just strip out \'s from the escaped augeas path or to do an aug.get query?

domcleal:
Interesting question! I don't think there's going to be any real difference, I can't think of an example right now that would be a problem, but I would tend towards using aug.get to retrieve that /path node.  The only disadvantage is that it only works with a file path (`/files/etc/foo.conf`), and not say, a VirtualHost node beneath the path (`/files/etc/foo.conf/VirtualHost/bar`), where that suffix has to be stripped first.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -511,7 +511,11 @@ def _create_vhost(self, path):
 
         """
         addrs = set()
-        args = self.aug.match(path + "/arg")
+        try:
+            args = self.aug.match(path + "/arg")
+        except RuntimeError:
+            logger.warn("Encountered a problem while parsing file: %s, skipping", path)
```
bmw: All `logger.warning` messages are shown to the user so it'd be nice if we could show them the unescaped path here. Is there an easy way to do this?

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=too-many-public-methods
+# pylint: disable=too-many-public-methods, protected-access
```
bmw: nit: We shouldn't do this at the top level of the file. I think we should either:

1. Make a `_call_create_vhost` method of `AugeasVhostTest` that just forwards all args to `_create_vhost`. In the body of this method, you can have the disabling comment.
2. Place the comment in the body of the `AugeasVhostTest` class like so:

```
class blah(object):
   """docstring"""
    # pylint: disable=protected-access
```

certbot-apache/certbot_apache/tests/testdata/debian_apache_2_4/augeas_vhosts/apache2/sites-enabled/000-default.conf
```diff
@@ -0,0 +1 @@
+../sites-available/000-default.conf
```
bmw: We can probably remove all of these broken symlinks.

bmw:
While looking through this code, it looks like we missed some escaping in `_enable_ocsp_stapling`.

SwartzCr:
ping @bmw  so that it shows up in his updates

bmw:
LGTM

PR #3184, opened 2016-06-17
Nginx space preservation

pde:
Implemented by tweaking `nginxparser` and using PyParsing's resources, rather than a less Pythonic approach.

Closes: #938


certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -152,17 +152,17 @@ def deploy_cert(self, domain, cert_path, key_path,
                 "install a cert.")
 
         vhost = self.choose_vhost(domain)
-        cert_directives = [['ssl_certificate', fullchain_path],
-                           ['ssl_certificate_key', key_path]]
+        cert_directives = [['\n', 'ssl_certificate', ' ', fullchain_path],
```
pde: It's not really clear to me whether it was better to allow this change the semantics of `parser._add_directive` that required spaces to be explicit like this, or whether I should have implicitly prepended a `\n` and joined with `" "` every time it's called.  The current approach is perhaps less magical, but it has the downside of exposing both structures for the parse tree up here in `configurator.py`.

schoen:
@pde, this branch unfortunately now has conflicts with master.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -26,20 +29,24 @@ class RawNginxParser(object):
...
-    assignment = (key + Optional(space + value, default=None) + semicolon)
-    location_statement = Optional(space + modifier) + Optional(space + location)
-    if_statement = Literal("if") + space + Regex(r"\(.+\)") + space
-    map_statement = Literal("map") + space + Regex(r"\S+") + space + Regex(r"\$\S+") + space
+    comment = space + Literal('#') + restOfLine()
+
+    assignment = space + key + Optional(space + value, default=None) + semicolon
+    location_statement = space + Optional(modifier) + Optional(space + location + space)
+    if_statement = space + Literal("if") + space + Regex(r"\(.+\)") + space
+    map_statement = space + Literal("map") + space + Regex(r"\S+") + space + Regex(r"\$\S+") + space
```
schoen: This predates this set of changes, but why it is reasonable to use a `Regex` instead of defining a specific object that represents that part of the grammar?

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -4,7 +4,7 @@
 import string
 
 from pyparsing import (
-    Literal, White, Word, alphanums, CharsNotIn, Forward, Group,
+    Literal, White, Word, alphanums, CharsNotIn, Combine, Forward, Group,
```
bryongloden: @pde. Is this funny (like should we laugh emoji?) because the relationship between these pyparsing 'words' is non-obvious

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -115,24 +134,111 @@ def load(_file):
...
+        item, spaced_item = self._coerce(x)
+        self.spaced.insert(i + self._spaces_before(i), spaced_item)
+        list.insert(self, i, item)
+
+    def append(self, x):
+        item, spaced_item = self._coerce(x)
+        self.spaced.append(spaced_item)
+        list.append(self, item)
+
+    def extend(self, x):
```
schoen: If you instead have `regular_list.extend(unspaced_list)` rather than this case, it seems to lose the spacing information because the the regular_list's extend method is used instead of the unspaced_list's version doesn't know it should treat the contents of the unspaced_list differently! (I'm not sure whether this issue can arise from any the uses to which UnspacedList is put in this code.) The same thing is true of trying to `.__add__` with a conventional list on the left rather than on the right.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -115,24 +134,111 @@ def load(_file):
...
+        self.spaced.__delitem__(i + self._spaces_before(i))
+        list.__delitem__(self, i)
+
+    def __deepcopy__(self, unused_memo):
+        l = UnspacedList(self[:])
+        l.spaced = copy.deepcopy(self.spaced)
+        return l
+
+
+    def _spaces_before(self, idx):
```
schoen: Can this ever be called with `idx < -1`? (In that case, it would never terminate.)

certbot-nginx/certbot_nginx/tests/nginxparser_test.py
```diff
@@ -168,5 +175,58 @@ def test_issue_518(self):
              [['set', '$webp "true"']]]
         ])
 
+class TestUnspacedList(unittest.TestCase):
+    """Test the raw low-level Nginx config parser."""
```
schoen: Shouldn't this comment say `"""Test the UnspacedList class"""`? That class is only one aspect of the low-level parser and this test doesn't test other parts of the parser.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -1,23 +1,29 @@
...
 from pyparsing import restOfLine
 
+logger = logging.getLogger(__name__)
 
 class RawNginxParser(object):
     # pylint: disable=expression-not-assigned
     """A class that parses nginx configuration with pyparsing."""
 
     # constants
+    space = Optional(White())
```
schoen: Is there some way that we can check that this grammar is equivalent to the grammar used by the real configuration parser inside of nginx itself? (I know that this grammar is currently passing all known test cases, which is certainly a good sign, but is there a practical way to compare code-against-code and say "apparently these grammars define the same language"?)

certbot-nginx/certbot_nginx/nginxparser.py
pde: You're right, this is buggy when used with constructions like `unsigned_list[-2] = thing`, though the symptom seems to be crashing rather than infinite looping.

certbot-nginx/certbot_nginx/nginxparser.py
pde: I don't think that's a problem.  If you're adding or extending a regular list with an unspaced one, you can do `a.extend(b.spaced)` or `a + b.spaced` if you want the other behaviour.

certbot-nginx/certbot_nginx/nginxparser.py
pde: Not to my knowledge, no.  The nginx grammar does not appear to be formally documented anywhere, and the canonical implementation is a hand-rolled parser in C.  I guess if we'd taken @olabini 's approach we could possibly have walked through and compared our hand-rolled parser with nginx's.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -115,24 +131,120 @@ def load(_file):
...
+
+    def __delitem__(self, i):
+        self.spaced.__delitem__(self._spaced_position(i))
+        list.__delitem__(self, i)
+
+    def __deepcopy__(self, unused_memo):
+        l = UnspacedList(self[:])
+        l.spaced = copy.deepcopy(self.spaced)
+        return l
+
```
schoen: I looked at the other methods defined by Python list objects and several are missing and will fall back to the list implementation. For example, the lack of `__mul__()` means that `UnspacedList(["foo", " ", "bar"])*3` is just the list `["foo", "bar", "foo", "bar", "foo", "bar"]` without a `.spaced` attribute.

In some cases, this is desirable; for example, for `__getitem__`, the behavior of the underlying list is exactly correct. It is arguably also correct for `__eq__` and other comparisons. In other cases, it may result in some confusion.

I see these list methods as not specifically implemented here: `__class__`, `__contains__`, `__delattr__`, `__delslice__`, `__doc__`, `__eq__`, `__format__`, `__ge__`, `__getattribute__`, `__getitem__`, `__getslice__`, `__gt__`, `__hash__`, `__iadd__`, `__imul__`, `__init__`, `__iter__`, `__le__`, `__len__`, `__lt__`, `__mul__`, `__ne__`, `__new__`, `__reduce__`, `__reduce_ex__`, `__repr__`, `__reversed__`, `__rmul__`, `__setattr__`, `__sizeof__`, `__str__`, `__subclasshook__`, `count`, `index`, `pop`, `remove`, `reverse`, `sort`.

Many of these are arguably right to use the underlying list implementation (for example, things like `index`, `count`, and `__len__` are credible). I am still concerned about at least `pop`, `remove`, `reverse`, and `sort`, which all modify the list. I tested trying to `reverse` an `UnspacedList`, and it caused the `.spaced` version to become severely desynchronized with the user-visible content.

I think the right answer is probably "anything that generates a new list with the entire content of the old list, or that modifies the existing object, should either be reimplemented or raise a `NotImplementedError`". So I have 4 specific examples of things I think need one or the other treatment. I think this is prudent even if the existing code doesn't call any of those methods, to avoid hard-to-debug cases if the parser is modified in the future.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -115,24 +131,120 @@ def load(_file):
...
+
+    def _spaced_position(self, idx):
+        "Convert from indexes in the unspaced list to positions in the spaced one"
+        pos = spaces = 0
+        # Normalize indexes like list[-1] etc, and save the result
+        if idx < 0:
+            idx = len(self) + idx
```
schoen: If you want to be really paranoid about this fix, Python's behavior is that `L[-i]` and `L.__getitem__(-i)` raise `IndexError` whenever `i>len(L)`.

pde: Added tests asserting that we get this right (by letting line 245 raise the exception for us)

certbot-nginx/certbot_nginx/nginxparser.py
pde: Now raising `NotImplementedError` for the problematic things in this list...

schoen:
LGTM now. I'm pretty paranoid about the correctness of the grammar (not the space preservation PR, but the underlying parser).

I opened an issue based on an upstream nginxparser issue:

https://github.com/certbot/certbot/issues/3206

I think this is just one of many ways where the grammar accepted by the actual Nginx software diverges from the grammar accepted by the parser we use, and I'm worried about how we can identify more of these and what their consequences will be. Also, the space preservation logic is fragile because it makes several assumptions about where the grammar will allow spaces to occur, which can be violated if the grammar changes even slightly in the future. But as we discussed, fragile ≠ incorrect. :-)

Issue #289, opened 2015-03-18
Pylint 1.4.3 linter_plugin.py


jdkasten:
Did a bit of digging on the pylint 1.4.3 [problem] (https://travis-ci.org/letsencrypt/lets-encrypt-preview/jobs/54677169) There are a few issues.... first of all it appears that for the classes where slots really matter... (classes derived from ```acme.util.ImmutableMap```) do not pass the recently added new-class style test.

https://bitbucket.org/logilab/astroid/diff/astroid/scoped_nodes.py?diff1=ca726f4c9aad&diff2=46deabf86bd2ff4cf2bb3c5017240f56dc48d4c1&at=default

The major errors happen when the scoped_node(_Environ) class gets run which throws the exception..
astroid.exceptions.AstroidError: Could not find <Import() l.398 [os] at 0x7f0feffce150> in <TryExcept() l.392 [os] at 0x7f0ff00bbfd0>'s children

Obviously, we should limit the plugin to run only on the classes derived from ImmutableMap, but the objects passed to the transform function are of ```<class 'astroid.scoped_nodes.Class>```.  
I did not find an easy way to access the underlying represented class... ie. classes like ```letsencrypt.acme.challenges.*``` where something like ```issubclass``` could be run.  Even if we do find a nice mechanism... the linter_plugin still won't work because the classes are not being recognized as having the ```newstyle``` attribute.

The changes causing the issues are extremely recent.  I might temporarily pin pylint to 1.4.1 or 1.4.2 until we can figure out what is going on or the issues are solved upstream.

kuba:
Fixed in 2a2e4d2. Please close, @jdkasten.

PR #3185, opened 2016-06-17
Cleanup dev setup

### LINKED ISSUE #3172: nosetests doesn't print coverage missing lines anymore
For some reason, nosetests stopped showing the coverage missing lines.

After doing some research, I figured that we can fix this by using a coverage config file and setting the `show_missing` parameter to true as shown here. https://coverage.readthedocs.io/en/coverage-4.1/config.html#report

This doesn't work because in order to do this, we need to use the `--cover-config-file` argument that is supposedly supported by nosetests. However, nosetests crashes with `nosetests version 1.3.7`.

This issue is reported here. https://github.com/nose-devs/nose/issues/1001

bmw:
The addition of `.coveragerc` fixes #3172 ([source](https://stackoverflow.com/questions/37733194/python-nosetests-with-coverage-no-longer-shows-missing-lines)).

`nosexcover` hasn't been used since dfd7e142c143e8f3f92541dfd02349592caf4de3.

~~PyCQA/pylint#248 has been resolved.~~

~~Lastly, the specific problem identified with using `astroid` in #289 appears to have been resolved. I didn't find the specific issue/PR on their repo, but `astroid` is only used for testing and all tests pass without pinning `astroid`.~~

EDIT: Turns out the `astroid`/`pylint` problems only shows up on some systems including Travis. I removed those changes.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+# show lines missing coverage in output
+show_missing = True
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,6 @@ dev_extras = [
     'astroid==1.3.5',
     'coverage',
     'nose',
-    'nosexcover',
     'pep8',
     'pylint==1.4.2',  # upstream #248
     'tox',


bmw:
Split this in two PRs so I can self-merge the cover change in good conscience.

PR #3187, opened 2016-06-17
Show lines missing test coverage in test output

### LINKED ISSUE #3172: nosetests doesn't print coverage missing lines anymore

bmw:
Fixes #3172

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+# show lines missing coverage in output
+show_missing = True


bmw:
Self-merging as this only affects tests and isn't installed by any of our users.

PR #3196, opened 2016-06-21
Explain the most likely cause of a missing replay nonce error

pde:
People have gotten this error because Akamai keeps answering during boulder outages, and it's pretty mysterious.

### DIFF
--- a/acme/acme/errors.py
+++ b/acme/acme/errors.py
@@ -49,7 +49,7 @@ class MissingNonce(NonceError):
 
     def __str__(self):
         return ('Server {0} response did not include a replay '
-                'nonce, headers: {1}\n(This may be a service outage)'.format(
+                'nonce, headers: {1} (This may be a service outage)'.format(
                     self.response.request.method, self.response.headers))
 
 
--- a/acme/acme/errors.py
+++ b/acme/acme/errors.py
@@ -49,7 +49,7 @@ class MissingNonce(NonceError):
 
     def __str__(self):
         return ('Server {0} response did not include a replay '
-                'nonce, headers: {1}'.format(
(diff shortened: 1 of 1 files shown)


acme/acme/errors.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, response, *args, **kwargs):
 
     def __str__(self):
         return ('Server {0} response did not include a replay '
-                'nonce, headers: {1}'.format(
+                'nonce, headers: {1}\n(This may be a service outage)'.format(
```
bmw: nit: I think we should remove the newline. None of the other exceptions in the `acme` library contain a newline and I think it is a bit odd to have it there. We don't really want the newline in log files for example.

Also, we could potentially catch `MissingNonce` errors at a higher level in the client and provide a better error message. Ideally, we'd do this at the lowest level in the client where we can catch the exceptions, however, its thrown at multiple places and we're worried about all of them (maybe we only really care about the first time we expect a nonce from the server?) then we could handle this in the `_handle_exception` function in `main.py`.

pde:
I've removed the newline. I guess it could be better to catch these elsewhere and display a crafted UI response, but I'm not sure enough that `MissingNonce` is always going to be the symptom of outages to think it's worth the effort.

bmw:
That's fine. LGTM.

bryongloden:
Greetings @bmw. How are you? Do you have the the meaning of LGTM? I've seen it used a few times in here

bmw:
I'm great @bryongloden. Hope you're doing good as well.

LGTM stands for "looks good to me." In this context, its often used to designate approval of the PR.

bryongloden:
@bmw SGTM. Sounds good to me

PR #3199, opened 2016-06-21
Remove warning about nginx options file

pde:
- This would be a security issue if we ran as setuid root at the request of
   non-privileged users, but we don't


pde:
CC: @diracdeltas unless I'm missing something.

diracdeltas:
I believe that warning was simply inherited from the Apache plugin, so if Apache is fixed then this is also good to go.

diracdeltas:
However, I no longer have write access to certbot so I can't merge this. :)

bmw:
LGTM. The PR where this was changed in Apache is #1401.

Issue #767, opened 2015-09-11
nginx leaves dirty/modified config files


marcan:
When letsencrypt-renewer is run, it makes temporary changes to the web server configs, reformatting everything in the process (at least for nginx). Those changes are not fully cleaned up after the renewer is done. Running 'letsencrypt config_changes' after as a workaround implicitly performs that cleanup (`rev.recovery_routine()`) and puts back the original config files. The renewer should be doing this itself.

Note: my use case is letting letsencrypt manage config changes for domain validation, but making the persistent changes to deploy the new certs myself; thus, I don't expect letsencrypt to ever make lasting changes to my nginx config files.

bmw:
This is not technically a renewer problem, but a Nginx problem. The renewer modifies files in the Let's Encrypt configuration directory, but relies on the plugins to make modifications to the web server configuration if necessary. Removing the `renewer` label.

tcz001:
Any instruction to reproduce this?

pde:
@marcan I realise this report was a while ago, but there's a team now polishing the nginx plugin for release; is there any chance you could still provide repro instructions?

bmw:
Fixed in #3204.

PR #3204, opened 2016-06-23
Fix Nginx reversion

### LINKED ISSUE #1977: Audit nginx plugin for guaranteed config reversion in case of error

bmw:
Fixes #1977.

After auditing the code for creating/modifying/removing files, proper reverter use, etc., it turns out this was trivial. The problem occurred when the Nginx plugin was performing a challenge. Before setting up the challenge, it would create a permanent checkpoint and write out the Nginx config. It is not expected that a checkpoint is created here so when rolling back, the client always rolled back one too few checkpoints.

It turns out that no changes are made before this save point so I deleted it (although if you break after the save in `master` you'll likely see a lot of changes, but that's only because your configuration has been normalized). The reason it was even there in the first place is the Nginx plugin is modeled after the Apache plugin which makes a temporary save at this point (the save used to be permanent in Apache).

Using this PR, you should ALWAYS be able to revert your Nginx configuration to a previous state. This can be accomplished using `certbot rollback`. Furthermore, Certbot should revert in progress changes if it is interrupted. This includes hitting ctrl+c at any point during the program.

NOTE: If it's interrupted when enabling a redirect but already successfully installed your cert, only the redirect changes are reverted. This is the desired behavior and is mentioned in the program output after handling this case. You can rollback all the way using `certbot rollback`.

The big caveat here is it's possible to interrupt the code rolling back your changes. For example, if you hit ctrl+c once to trigger the error handling code and then hit ctrl+c again, you'll interrupt the error handler and your changes might not be reverted. With that said, even if you interrupt the error handler, `certbot rollback` will still work.

With all that said, I encourage you to try and break this PR. I tried and couldn't but this will certainly benefit from additional testing, including the certbot-compatibility-test.

### DIFF
--- a/certbot-nginx/certbot_nginx/tests/tls_sni_01_test.py
+++ b/certbot-nginx/certbot_nginx/tests/tls_sni_01_test.py
@@ -80,7 +80,7 @@ class TlsSniPerformTest(util.NginxTest):
 
         mock_setup_cert.assert_called_once_with(self.achalls[0])
         self.assertEqual([response], responses)
-        self.assertEqual(mock_save.call_count, 2)
+        self.assertEqual(mock_save.call_count, 1)
 
         # Make sure challenge config is included in main config
         http = self.sni.configurator.parser.parsed[
--- a/certbot-nginx/certbot_nginx/tls_sni_01.py
+++ b/certbot-nginx/certbot_nginx/tls_sni_01.py
@@ -46,8 +46,6 @@ class NginxTlsSni01(common.TLSSNI01):
         if not self.achalls:
             return []
 
-        self.configurator.save()
-
         addresses = []
         default_addr = "{0} default_server ssl".format(
             self.configurator.config.tls_sni_01_port)


pde:
One concern I have is this `except` without a `raise`, which theoretically suggest that the Nginx plugin could be tripped up if a file became unwritable partway through a series of config changes: https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/parser.py#L219

However I guess (though I haven't checked closely) that it doesn't matter, because operations like cleanup after challenge completion are actually performed by the reverter, and not by the plugin itself.  And not having the `raise` is valuable until we fix #3205.

pde:
Otherwise, LGTM.

pde:
(Probably) fixes: #767

bmw:
> One concern I have is this except without a raise, which theoretically suggest that the Nginx plugin could be tripped up if a file became unwritable partway through a series of config changes: https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/parser.py#L219

> However I guess (though I haven't checked closely) that it doesn't matter, because operations like cleanup after challenge completion are actually performed by the reverter, and not by the plugin itself. And not having the raise is valuable until we fix #3205.

Exactly, The reverter works by making a copy of all files that are going to be modified. This should be (and is) done before the files are modified. Even if the modification fails, we have valid backups we can restore to.

In this case, the only function that calls `filedump` is `configurator.NginxConfigurator.save`. Looking at [that method](https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/configurator.py#L511), you'll see we add files that are about to be modified to the checkpoint before calling `filedump`.

In this theoretical case, if the file is unwriteable we might not be able to restore the backup, but I'm not sure what else we could reasonably do here. At the very least, there will be a valid backup on disk that can be used once the file is made writeable again.

bmw:
For those that want some bedtime reading about how Certbot helps guarantee successful reversion when using plugins, here's a ton of information I've written over the past few months. I'll probably try and add some of this information to our docs/code if it's not already present. In the meantime:

If you're not already familiar with [plugin interface](https://github.com/certbot/certbot/blob/master/certbot/interfaces.py#L99), read that first. Relevant interfaces are IPlugin, IAuthenticator, and IInstaller.

Next, [this comment](https://github.com/certbot/certbot/issues/1977#issuecomment-185493304) describes how Certbot uses these methods to restore your config.

Furthermore, [this comment](https://github.com/certbot/certbot/pull/2510#issue-134968581) talks a bit about how the reverter works and a little about a design decision around `perform` and `cleanup`.

Finally, here are my personal notes I used when writing this PR. I cleaned them up and expanded them a bit so hopefully they're more understandable:

```
Client assumptions:
    - recovery_routine() reverts all unfinalized changes
    - cleanup() reverts all changes made by perform(), even if perform is
      interupted
    - save() finalizes exactly one checkpoint when given a title and temporary
      is set to False. Furthermore, a call to this function finalizes at most
      one checkpoint and this is the only plugin method that finalizes
      checkpoints. If an exception was thrown during save, a checkpoint was not 
      finalized

Where does Nginx make changes?
Below is a tree that starts with direct modifiers to the Nginx config and 
branches based on callers. Branch ends when only callers are outside of the 
plugin.

- nginxparser.dump
    - parser.NginxParser.filedump
        - configurator.NginxConfigurator.save
            - tls_sni_01.perform * 2 
                - configurator.perform
    - tls_sni_01.NginxTlsSni01.perform
- reverter.Reverter.recovery_routine
    - configurator.recovery_routine
    - configurator.__init__
- reverter.Reverter.revert_temporary_config
    - configurator.revert_challenge_config
        - configurator.cleanup
- reverter.Reverter.rollback_checkpoints

Where does Nginx register changes with the reverter?
- All changes through configurator.NginxConfigurator.save are backed up through
  add_to_checkpoint and add_to_temp_checkpoint.
- Change made in perform is backed up through call to register_file_creation.

Problem:
- We finalize a checkpoint in perform() violating our assumption that only save
  finalizes checkpoints.
- Additionally, because cleanup() doesn't revert the checkpoint, cleanup()
  violates the assumption that it correctly reverts changes made by
  perform()

Solution:
- Stop finalizing checkpoints in perform(). This solves both problems
  identified above.
```

If you read all of this and you're not asleep yet, I recommend sleeping pills.

PR #3208, opened 2016-06-24
Better document plugins and reversion

bmw:


pde:
Woohoo, documentation!

Issue #2734, opened 2016-03-31
More useful error when running as non-root?


schoen:
I tried running the client as non-root by mistake and it failed with

```An unexpected error occurred:
IOError: [Errno 13] Permission denied: '/var/log/letsencrypt/letsencrypt.log'
Please see the logfile 'letsencrypt.log' for more details.```

Maybe it could say something more useful about needing to be run as root in this case.

bmw:
I agree. I noticed this when preparing my RightsCon demo. I was very careful to always run the client with `sudo` during the demo to avoid this embarrassingly bad error message.

whitenoiseoss:
I'm going to have a look at this this weekend.

bmw:
Fixed in #3265.

PR #3214, opened 2016-06-25
Improve non root error message [revision requested]

DownMoney:
This PR aims to add a more verbose message when running certbot without root privileges, as mentioned in #2734.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -712,6 +712,11 @@ def main(cli_args=sys.argv[1:]):
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
 
+    # Check if running as root
+    if os.geteuid() != 0:
+        raise errors.Error(
+            "You need to have root privileges to run certbot.\nPlease try again, this time using 'sudo'. Exiting.")
+
     # Setup logging ASAP, otherwise "No handlers could be found for
     # logger ..." TODO: this should be done before plugins discovery
     for directory in config.config_dir, config.work_dir:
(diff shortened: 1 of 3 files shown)


pde:
@jsha recently made a change related to this; assigning him for review

jsha:
Thanks! Under what conditions do you run into this issue? I think the change in https://github.com/certbot/certbot/pull/3164/files should cover the most common case, since we try to write to /var/log/letsencrypt.log very early on.

I think it definitely makes sense to have a similar message in multiple places. Can you pull out the message in #3164 to a shared variable to we can return the same message here?

pde:
Since @DownMoney seems to have disappeared, assigning this to Brad who's doing a non-root error sweep for 0.9.0.

DownMoney:
Sorry I was just moving house and only just now got the internet connection. I get the error message when running certbot on Mac without sudo If you still wan tI can update the error message.

bmw:
Sorry this has sat around so long.

The common cases of error messages resulting from running Certbot without root were addressed in #3265. In general, we shouldn't be saying that Certbot needs root as it's possible to run the client without those privileges (see [this](https://certbot.eff.org/faq/#does-certbot-require-root-administrator-privileges)). Additionally, we shouldn't be telling people to use `sudo` which isn't present on all systems.

If you'd like to help further improve these messages, I wrote up what I think should be done in #3429. If you make the changes described there in this or another PR, I'd be happy to review.

DownMoney:
Brad,

Ok sure I will have a look over the weekend and get back to you.

jsha:
@DownMoney: Any update? @bmw: Is this still a live issue or has it been fixed elsewhere?

bmw:
@jsha: see #3429

bmw:
Closing as this PR hasn't seen any progress in quite some time. If you'd like to continue work on #3429, please open another PR when it's ready for review.

Issue #3205, opened 2016-06-23
Nginx plugin shouldn't write files it hasn't changed


pde:
This is especially bad because of #938, but conceptually, we shouldn't be messing with things we don't need to, and it may have undesirable consequences such as modifying `mtime`, ownership, etc.

PR #3218, opened 2016-06-27
Only write nginx config files if they've been modified

pde:
We keep a dirty bit for each node in the parse tree, and walk it to implement a `.is_dirty()` method.  That's not very efficient, but is no more inefficient than having to perform elaborate the deepcopying that a `.top` attribute on each node of the tree would require.

Fixes: #3205


schoen:
LGTM. :penguin:

Issue #3215, opened 2016-06-27
Signals during error handling


bmw:
For those not familiar, `certbot/error_handler.py` can be used to execute cleanup code in case an exception or signal is encountered. Currently, however, the error handler can still be interrupted. For example, suppose the user hits ctrl+c to interrupt Certbot. If the user hits ctrl+c again while the error handler is running, it is interrupted and exits without performing all of the necessary cleanup code.

We could potentially ignore such signals during error handling. If we want to do this, I think the best way is before running any cleanup functions, the `ErrorHandler` uses `signal` to ignore signals such as `SIGINT`. After running the cleanup code, the original signal handlers should be restored.

PR #3223, opened 2016-06-28
Fix Unix signal handling in certbot.error_handler.ErrorHandler [minor revision]

### LINKED ISSUE #3219: Resuming error handling functions after a signal
Despite me writing `certbot/error_handler.py` many months ago, it just surprised me. What surprised me is more or less documented below:

> Every registered function is attempted to be run to completion exactly once. If a registered function raises an exception, it is logged and the next function is called. If a (different) handled signal occurs while calling a registered function, it is attempted to be called again by the next signal handler.

I think not calling a registered function again if it just raised an exception makes sense. If it errored the first time, it's likely to error again. I'm not sure if calling the function again if it is interrupted by a signal, however, is the right behavior. Registered functions can change state and an interrupted function may leave something in an invalid state such that calling it again causes incorrect behavior.

For a real life example, `client.py` registers the method `_restart_and_rollback` with the `ErrorHandler`. This method rolls back one checkpoint and restarts the plugin. What if, however, a signal interrupts the method after rolling back the checkpoint. Calling it again will attempt to rollback another checkpoint which is incorrect.

I think we should call try and call a registered function at most once to avoid this problem.

cowlicks:
Closes #3219 and #3215 

This tweaks how Unix signals are managed by `certbot.error_handler.ErrorHandler`.

Previously, signals sent while `ErrorHandler`'s cleanup functions were running would cause the functions to get run again.

Now all signals are deferred until after the cleanup functions are run.

### DIFF
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -112,7 +112,8 @@ class ErrorHandler(object):
         self.prev_handlers.clear()
 
     def _signal_handler(self, signum, unused_frame):
-        """Stores the recieved signal.
+        """Stores the recieved signal. If we are executing the code block in
+        the body of the context manager, stop by raising signal exit.
 
         :param int signum: number of current signal
 
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -115,7 +115,9 @@ class ErrorHandler(object):
         self.prev_handlers.clear()
 
     def _signal_handler(self, signum, unused_frame):
-        """Stores the recieved signal. If we are executing the code block in
+        """Replacement function for handling recieved signals.
+
+        Store the recieved signal. If we are executing the code block in
         the body of the context manager, stop by raising signal exit.
 
         :param int signum: number of current signal
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -75,10 +75,10 @@ class ErrorHandler(object):
                     traceback.format_exception(exec_type, exec_value, trace)))
 
             self._call_registered()
-            self._call_signals()
             return retval
         finally:
-            self._reset_signal_handlers()
+            prev_handlers = self._reset_signal_handlers()
+            self._call_signals(prev_handlers)
 
     def register(self, func, *args, **kwargs):
         """Sets func to be called with *args and **kwargs during cleanup
@@ -112,7 +112,9 @@ class ErrorHandler(object):
         """Resets signal handlers for signals in _SIGNALS."""
         for signum in self.prev_handlers:
             signal.signal(signum, self.prev_handlers[signum])
+        out = dict((k, v) for k, v in self.prev_handlers.items())
         self.prev_handlers.clear()
+        return out
 
     def _signal_handler(self, signum, unused_frame):
         """Replacement function for handling recieved signals.
@@ -127,7 +129,7 @@ class ErrorHandler(object):
         if not self.body_executed:
             raise errors.SignalExit
 
-    def _call_signals(self):
+    def _call_signals(self, prev_handlers):
         """Finally call the deferred signals.
 
         :param int signum: signal number
@@ -135,5 +137,5 @@ class ErrorHandler(object):
         """
         for signum in self.received_signals:
             logger.debug("Calling signal %s", signum)
-            signal.signal(signum, self.prev_handlers[signum])
+            signal.signal(signum, prev_handlers[signum])
             os.kill(os.getpid(), signum)
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -19,6 +19,14 @@ _SIGNALS = ([signal.SIGTERM] if os.name == "nt" else
              signal.SIGXCPU, signal.SIGXFSZ])
 
 
+class SignalExit(Exception):
+    """This exception is used stop of execution of the code in the body of the
+    ErrorHandler context manager.
+
+    """
+    pass
+
+
 class ErrorHandler(object):
     """Registers functions to be called if an exception or signal occurs.
 
@@ -34,16 +42,17 @@ class ErrorHandler(object):
     if a signal is encountered, cleanup_func is called followed by the
     previously registered signal handler.
 
-    Every registered function is attempted to be run to completion
-    exactly once. If a registered function raises an exception, it is
-    logged and the next function is called. If a (different) handled
-    signal occurs while calling a registered function, it is attempted
-    to be called again by the next signal handler.
+    Each registered cleanup function is called exactly once. If a registered
+    function raises an exception, it is logged and the next function is called.
+    Signals received while the registered functions are executing are
+    deferred until they finish.
 
     """
     def __init__(self, func=None, *args, **kwargs):
+        self.body_executed = False
         self.funcs = []
         self.prev_handlers = {}
+        self.received_signals = []
         if func is not None:
             self.register(func, *args, **kwargs)
 
@@ -51,12 +60,22 @@ class ErrorHandler(object):
         self.set_signal_handlers()
 
     def __exit__(self, exec_type, exec_value, trace):
+        self.body_executed = True
+        retval = False
+        if exec_type is SignalExit:
+            logger.debug("Encountered signals: %s", self.received_signals)
+            self.call_registered()
+            for signum in self.received_signals:
+                self.call_signal(signum)
+            retval = True
         # SystemExit is ignored to properly handle forks that don't exec
-        if exec_type not in (None, SystemExit):
+        elif exec_type not in (None, SystemExit):
             logger.debug("Encountered exception:\n%s", "".join(
                 traceback.format_exception(exec_type, exec_value, trace)))
             self.call_registered()
+
         self.reset_signal_handlers()
+        return retval
 
     def register(self, func, *args, **kwargs):
         """Sets func to be called with *args and **kwargs during cleanup
@@ -93,12 +112,21 @@ class ErrorHandler(object):
         self.prev_handlers.clear()
 
     def _signal_handler(self, signum, unused_frame):
-        """Calls registered functions and the previous signal handler.
+        """Stores the recieved signal.
 
         :param int signum: number of current signal
 
         """
-        logger.debug("Singal %s encountered", signum)
-        self.call_registered()
+        self.received_signals.append(signum)
+        if not self.body_executed:
+            raise SignalExit
+
+    def call_signal(self, signum):
+        """Calls the signal given by signum.
+
+        :param int signum: signal number
+
+        """
+        logger.debug("Calling signal %s", signum)
         signal.signal(signum, self.prev_handlers[signum])
         os.kill(os.getpid(), signum)
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -5,6 +5,7 @@ import os
 import signal
 import traceback
 
+from certbot import errors
 
 logger = logging.getLogger(__name__)
 
@@ -17,14 +18,6 @@ logger = logging.getLogger(__name__)
 _SIGNALS = ([signal.SIGTERM] if os.name == "nt" else
             [signal.SIGTERM, signal.SIGHUP, signal.SIGQUIT,
              signal.SIGXCPU, signal.SIGXFSZ])
-
-
-class SignalExit(Exception):
-    """This exception is used stop of execution of the code in the body of the
-    ErrorHandler context manager.
-
-    """
-    pass
 
 
 class ErrorHandler(object):
@@ -62,7 +55,7 @@ class ErrorHandler(object):
     def __exit__(self, exec_type, exec_value, trace):
         self.body_executed = True
         retval = False
-        if exec_type is SignalExit:
+        if exec_type is errors.SignalExit:
             logger.debug("Encountered signals: %s", self.received_signals)
             self.call_registered()
             for signum in self.received_signals:
@@ -120,7 +113,7 @@ class ErrorHandler(object):
         """
         self.received_signals.append(signum)
         if not self.body_executed:
-            raise SignalExit
+            raise errors.SignalExit
 
     def call_signal(self, signum):
         """Calls the signal given by signum.
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -50,6 +50,7 @@ class ErrorHandler(object):
             self.register(func, *args, **kwargs)
 
     def __enter__(self):
+        self.body_executed = False
         self.set_signal_handlers()
 
     def __exit__(self, exec_type, exec_value, trace):
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -61,22 +61,24 @@ class ErrorHandler(object):
         self._set_signal_handlers()
 
     def __exit__(self, exec_type, exec_value, trace):
-        self.body_executed = True
-        retval = False
-        if exec_type is errors.SignalExit:
-            logger.debug("Encountered signals: %s", self.received_signals)
-            self._call_registered()
-            for signum in self.received_signals:
-                self._call_signal(signum)
-            retval = True
-        # SystemExit is ignored to properly handle forks that don't exec
-        elif exec_type not in (None, SystemExit):
-            logger.debug("Encountered exception:\n%s", "".join(
-                traceback.format_exception(exec_type, exec_value, trace)))
-            self._call_registered()
+        try:
+            self.body_executed = True
+            retval = False
+            # SystemExit is ignored to properly handle forks that don't exec
+            if exec_type in (None, SystemExit):
+                return retval
+            elif exec_type is errors.SignalExit:
+                logger.debug("Encountered signals: %s", self.received_signals)
+                retval = True
+            else:
+                logger.debug("Encountered exception:\n%s", "".join(
+                    traceback.format_exception(exec_type, exec_value, trace)))
 
-        self._reset_signal_handlers()
-        return retval
+            self._call_registered()
+            self._call_signals()
+            return retval
+        finally:
+            self._reset_signal_handlers()
 
     def register(self, func, *args, **kwargs):
         """Sets func to be called with *args and **kwargs during cleanup
@@ -123,12 +125,13 @@ class ErrorHandler(object):
         if not self.body_executed:
             raise errors.SignalExit
 
-    def _call_signal(self, signum):
+    def _call_signals(self):
         """Calls the signal given by signum.
 
         :param int signum: signal number
 
         """
-        logger.debug("Calling signal %s", signum)
-        signal.signal(signum, self.prev_handlers[signum])
-        os.kill(os.getpid(), signum)
+        for signum in self.received_signals:
+            logger.debug("Calling signal %s", signum)
+            signal.signal(signum, self.prev_handlers[signum])
+            os.kill(os.getpid(), signum)
--- a/certbot/error_handler.py
+++ b/certbot/error_handler.py
@@ -77,8 +77,8 @@ class ErrorHandler(object):
             self._call_registered()
             return retval
         finally:
-            prev_handlers = self._reset_signal_handlers()
-            self._call_signals(prev_handlers)
+            self._reset_signal_handlers()
+            self._call_signals()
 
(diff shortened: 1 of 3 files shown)


pde:
Hah, I think @bmw might have started work on this yesterday.  Assigning him for review.

cowlicks:
Travis failed because a one line closure doesn't have a docstring >____<
 
I don't think my solution is particularly elegant so if @bmw has nicer approach free to close this.

bmw:
@cowlicks, I do have an approach I like better, but I haven't started implementing it yet. If you'd like to continue working on this, we'd appreciate the contribution and I'm happy to describe what I had in mind.

I was planning on creating a second signal handler method in the `ErrorHandler` class. This method would simply log the signum and frame it received and append the signum to a list. Then, in `__exit__`, if there is an exception, we set the handler for all signums in `_SIGNALS` to this new signal handler. Next, we run the registered functions as normal (although I'd pop the function from the list before calling it in `call_registered` to make sure #3219 is resolved). Finally, after running the registered functions and restoring the original signal handlers, loop through all signums that were added to the list and send the signal to ourselves.

And I think that should do it! Let me know if you have any questions and whether or not you'd like to continue work on this.

cowlicks:
@bmw I *think* my PR has the same behavior as what you're describing. However my implementation is different.

I only use *one* signal handler function. It always logs signals for later. But if there is a signal while the code block in the context manager is being executed, the signal handler raises `SignalExit` to hand things off to `__exit__`.  Then in `__exit__`, we run the registered cleanup functions, then call/send the logged signals.

Having the signal handler work in both cases avoids the complexity of having to set, reset, and unset handlers for all the signums. And since the signal handler just logs the signals and does not call `call_registered`, it avoids the issues with receiving signals causing `call_registered` to be called twice (as in #3219).

Does this sound reasonable? If you see a discrepancy with the behavior in my implementation, and what you would like, can you write a quick pseudo-code test to illustrate it?

bmw:
@cowlicks, you're right. What you proposed is quite reasonable. Sorry I didn't look at it more closely before commenting.

I still have some suggestions on things I'd like to see changed, but in many ways, what you implemented here is better than my planned approach. Nice work!

cowlicks:
@bmw ping. Suggestions?

bmw:
Sorry. I somewhat got caught up in other things but mainly have been thinking about how to best handle what we're trying to do here. This PR is basically a strict improvement over the current error handler, however, there are still a lot of cases where things can go wrong.

For example, it's still possible for a signal to interrupt the error handler as described in #3215. Imagine we're in `__exit__` before the `self.body_executed = True` line and get a signal. In this case, we'll raise `SignalExit`, breaking out of the function before performing any cleanup code or resetting the signal handlers.

I have some ideas on how to help mitigate this and other potential corner cases, however, it adds quite a bit of complexity that might not be worth it (they involve locks and inspecting the frame argument to the signal handler). I'm still not sure exactly what we should do here, but I think your general approach is the right one. I'll try to make comments on minor things I'd like to see changed (hopefully today) and create follow up issues/PRs if necessary.

Sorry again for the delay!

certbot/error_handler.py
```diff
@@ -19,6 +19,14 @@
              signal.SIGXCPU, signal.SIGXFSZ])
 
 
+class SignalExit(Exception):
```
bmw: This class should be added to `errors.py` instead of this file since it can propagate out to the user.

certbot/error_handler.py
```diff
@@ -19,6 +19,14 @@
...
+    """This exception is used stop of execution of the code in the body of the
+    ErrorHandler context manager.
```
bmw: nit: The docstring here should have the following form, according to [our style guide](https://certbot.eff.org/docs/contributing.html#coding-style) (and more specifically [pep8](https://www.python.org/dev/peps/pep-0008/) and [pep257](https://www.python.org/dev/peps/pep-0257/)):

```
"""one-line-docstring"""
```

or

```
"""one-line-doc-string

more-info

"""
```

Where each line is no more than 72 characters in length.

certbot/error_handler.py
```diff
@@ -19,6 +19,14 @@
...
+
+    """
+    pass
```
bmw: nit: pass isn't necessary due to the docstring

certbot/error_handler.py
```diff
@@ -19,6 +19,14 @@
...
+
+
 class ErrorHandler(object):
     """Registers functions to be called if an exception or signal occurs.
```
bmw: With the change of raising an exception in the signal handler, this class can really only be used as a context manager. I think this is the right approach, however, we should update the docstring below to say that and make all methods besides `register` "protected" by adding a leading underscore.

certbot/error_handler.py
```diff
@@ -34,29 +42,40 @@ class works best as a context manager. For example:
...
     def __init__(self, func=None, *args, **kwargs):
+        self.body_executed = False
         self.funcs = []
         self.prev_handlers = {}
+        self.received_signals = []
         if func is not None:
             self.register(func, *args, **kwargs)
 
     def __enter__(self):
         self.set_signal_handlers()
```
bmw: nit: While not necessary based on our current usage, we never reset `body_executed` meaning this class can only be used once. We can easily fix the problem by setting also the `body_executed` in the beginning of `__enter__` or the of `__exit__`. I don't really care, but I prefer the beginning of `__enter__`.

certbot/error_handler.py
```diff
@@ -34,29 +42,40 @@ class works best as a context manager. For example:
...
 
     def __exit__(self, exec_type, exec_value, trace):
+        self.body_executed = True
```
bmw: Signals that occur if `__exit__` is triggered due to an exception other than `SignalExit` are completely ignored. This should be fixed.

A nice way to do this I think is to refactor this function a bit to have a form like the following:
```
self.body_executed = True
retval = False
if exec_type in (None, SystemExit):
    return retval
elif exec_type is SignalExit:
    log signals
    retVal = True
else:
    log exception

call_registered
call_signals
reset_signals
return retVal
```

certbot/error_handler.py
```diff
@@ -93,12 +112,22 @@ def reset_signal_handlers(self):
         self.prev_handlers.clear()
 
     def _signal_handler(self, signum, unused_frame):
-        """Calls registered functions and the previous signal handler.
+        """Stores the recieved signal. If we are executing the code block in
+        the body of the context manager, stop by raising signal exit.
```
bmw: nit: Same comment as the one on `SignalExit` about docstring style.

certbot/error_handler.py
```diff
@@ -93,12 +112,22 @@ def reset_signal_handlers(self):
...
 
         """
-        logger.debug("Singal %s encountered", signum)
-        self.call_registered()
+        self.received_signals.append(signum)
+        if not self.body_executed:
+            raise SignalExit
+
+    def call_signal(self, signum):
+        """Calls the signal given by signum.
```
bmw: Perhaps we should make this clear that this uses the previous signal handler to handle the signal?

Also, we could perhaps move the looping over the signals to this method? This allows `__exit__` to simply call `self._call_signals()` rather than looping itself.

certbot/error_handler.py
```diff
@@ -34,29 +42,40 @@ class works best as a context manager. For example:
...
         self.set_signal_handlers()
 
     def __exit__(self, exec_type, exec_value, trace):
+        self.body_executed = True
+        retval = False
+        if exec_type is SignalExit:
+            logger.debug("Encountered signals: %s", self.received_signals)
+            self.call_registered()
+            for signum in self.received_signals:
+                self.call_signal(signum)
```
bmw: I think we should call `_reset_signal_handlers()` before doing this. There are two benefits to this approach:

1. We're not looping over a list that may be modified if we get another signal.
2. We ensure no signals are incorrectly ignored (what currently happens if we get a signal after this loop but before the call to `_reset_signal_handlers`?).

certbot/error_handler.py
bmw: Feel free to ignore this comment about the docstring if you take my suggestion about calling `_reset_signal_handlers` first.

certbot/tests/error_handler_test.py
```diff
@@ -1,11 +1,31 @@
 """Tests for certbot.error_handler."""
+import os
 import signal
 import sys
 import unittest
+from contextlib import contextmanager
```
bmw: nit: In general, we just import the module rather than specific methods, decorators, etc. So here, you'd want to `import contextlib` (as the first import) and change line 11 to `@contextlib.contextmanager`.

certbot/tests/error_handler_test.py
```diff
@@ -30,25 +50,20 @@ def test_context_manager(self):
...
-        for signum in self.signals:
-            mock_signal.signal.assert_any_call(signum, signal_handler)
-
-        signum = self.signals[0]
-        signal_handler(signum, None)
+    def test_context_manager_with_signal(self):
+        with signal_receiver(self.signals) as signals_received:
+            with self.handler:
+                should_be_42 = 42
+                send_signal(signal.SIGTERM)
```
bmw: I think we should avoid hardcoding specific signals in these tests. Why?

1. When changing the signals list in `error_handler.py` we have to remember to update these as well.
2. In the case of `SIGHUP` below, we don't actually use that signal on all systems

Instead, I think we should use `self.signals[0]`, `self.signals[-1]`, etc.

certbot/tests/error_handler_test.py
```diff
@@ -30,25 +50,20 @@ def test_context_manager(self):
...
+                should_be_42 *= 10
+
+        # check exectuion stoped when the signal was sent
+        assert 42 == should_be_42
+        # assert signals were caught
+        assert [signal.SIGTERM] == signals_received
```
bmw: nit: you should use `self.assertTrue` rather than `assert`.

certbot/tests/error_handler_test.py
```diff
@@ -58,6 +73,18 @@ def test_bad_recovery(self):
...
         bad_func.assert_called_once_with()
 
+    def test_bad_recovery_with_signal(self):
+        bad_func = mock.MagicMock(
+                side_effect=lambda: send_signal(signal.SIGHUP))
+        self.handler.register(bad_func)
+        with signal_receiver(self.signals) as signals_received:
+            with self.handler:
+                send_signal(signal.SIGTERM)
+        assert [signal.SIGTERM, signal.SIGHUP] == signals_received
```
bmw: `self.assertTrue` instead of `assert`

certbot/tests/error_handler_test.py
```diff
@@ -30,25 +50,20 @@ def test_context_manager(self):
...
+        # assert signals were caught
+        assert [signal.SIGTERM] == signals_received
+        # assert the error handling function was just called once
         self.init_func.assert_called_once_with(*self.init_args,
                                                **self.init_kwargs)
-        mock_os.kill.assert_called_once_with(mock_os.getpid(), signum)
-
-        self.handler.reset_signal_handlers()
-        for signum in self.signals:
-            mock_signal.signal.assert_any_call(signum, signal.SIG_DFL)
```
bmw: Removing this, there is now no test that we restore signals correctly. There are a number of ways to do this, but I think in general, we should basically duplicate what the `ErrorHandler` does in a test. We should use `signal.getsignal` to record all signal handlers in use before using the `ErrorHandler` and use `self.assertEqual` to check that the signal was restored properly after using the `ErrorHandler`.

bmw:
Most of my comments are simple, stylistic changes. Like I said before, I think this approach is an improvement over the current `ErrorHandler`. For example, apparently the current class can [deadlock](https://docs.python.org/2/library/logging.html#thread-safety) and this problem is resolved by this PR.

cowlicks:
Thanks for the review! I'll take a look at these tonight.

cowlicks:
@bmw I think a simple way to (almost) ensure the signal handlers are reset is to use a try/finally block like:

```python
class ErrorHandler:
    def __exit__(...):
        try:
            ...
        finally:
            self.reset_signal_handlers()
            return retval
````

**Comment by cowlicks** [comment]
@bmw ready for review

File: certbot/error_handler.py [hunk 15]
```diff
@@ -20,43 +21,64 @@
...
+            elif exec_type is errors.SignalExit:
+                logger.debug("Encountered signals: %s", self.received_signals)
+                retval = True
+            else:
+                logger.debug("Encountered exception:\n%s", "".join(
+                    traceback.format_exception(exec_type, exec_value, trace)))
+
+            self._call_registered()
+            return retval
+        finally:
```
bmw: Question: What is the purpose of the `try`/`finally`? Where are we worried about an exception being raised? Is it in case we get a signal before the `self.body_executed = True` line?

certbot/error_handler.py
```diff
@@ -86,19 +108,34 @@ def set_signal_handlers(self):
...
                 signal.signal(signum, self._signal_handler)
 
-    def reset_signal_handlers(self):
+    def _reset_signal_handlers(self):
         """Resets signal handlers for signals in _SIGNALS."""
         for signum in self.prev_handlers:
             signal.signal(signum, self.prev_handlers[signum])
+        out = dict((k, v) for k, v in self.prev_handlers.items())
         self.prev_handlers.clear()
+        return out
```
bmw: `out` seems unnecessary. See my comments on `_call_signals` for why.

certbot/error_handler.py
```diff
@@ -86,19 +108,34 @@ def set_signal_handlers(self):
...
+
+    def _call_signals(self, prev_handlers):
+        """Finally call the deferred signals.
+
+        :param int signum: signal number
+
+        """
+        for signum in self.received_signals:
+            logger.debug("Calling signal %s", signum)
+            signal.signal(signum, prev_handlers[signum])
```
bmw: Because we're calling `_reset_signal_handlers` first, we don't need to do this now. The call to `signal.signal` in `_reset_signal_handlers` already restored the previous signal handlers.

certbot/error_handler.py
```diff
@@ -86,19 +108,34 @@ def set_signal_handlers(self):
...
-        signal.signal(signum, self.prev_handlers[signum])
-        os.kill(os.getpid(), signum)
+        self.received_signals.append(signum)
+        if not self.body_executed:
+            raise errors.SignalExit
+
+    def _call_signals(self, prev_handlers):
+        """Finally call the deferred signals.
+
+        :param int signum: signal number
```
bmw: This parameter has been removed. If you take my suggestion to remove `prev_handlers`, we can just make this a one line docstring.

certbot/tests/error_handler_test.py
```diff
@@ -26,34 +46,51 @@ def test_context_manager(self):
             with self.handler:
                 raise ValueError
         except ValueError:
-            pass
+            exception_raised = True
+
+        self.assertTrue(exception_raised)
```
bmw: This is probably a good addition, but what happens if `ValueError` is not raised? This line will blow up because `exception_raised` isn't defined. I'd add a `exception_raised = False` line to the beginning of this method.

certbot/tests/error_handler_test.py
```diff
@@ -26,34 +46,51 @@ def test_context_manager(self):
...
+        # assert the error handling function was just called once
         self.init_func.assert_called_once_with(*self.init_args,
                                                **self.init_kwargs)
-        mock_os.kill.assert_called_once_with(mock_os.getpid(), signum)
-
-        self.handler.reset_signal_handlers()
         for signum in self.signals:
-            mock_signal.signal.assert_any_call(signum, signal.SIG_DFL)
+            sig = signal.getsignal(signum)
+            self.assertTrue((sig == signal.SIG_DFL) or (sig == signal.SIG_IGN))
```
bmw: So I'm a bit worried about this. A lot of our package maintainers use our tests to make sure Certbot is working properly and signal handling can vary on different OSs. Will the signal always been `SIG_DFL` or `SIG_IGN` when running these tests? I have no idea.

With this in mind, I'd actually save the signal handlers in use at the start of this method and test that they've been restored to those values here. You could potentially do this by breaking some of the code from `signal_receiver` into a separate function that returns a `dict` of the signal handlers currently in use. The best way to do this in my opinion is to call the separate function and then loop over `signums` again to set the `lambda` signal handler.

bmw:
A couple small suggestions and one question, but other than that, this looks good! Thanks again for working on this @cowlicks.

certbot/tests/error_handler_test.py
cowlicks: That sounds like a good solution.

I did this little hack because `getsignal(SIGXFSZ)` returns `SIG_IGN`. But before `getsignal` was mocked to return `SIG_DFL`.

certbot/error_handler.py
cowlicks: Good catch! Thanks.

certbot/tests/error_handler_test.py
bmw: Yeah I encountered the same problem and it surprised me. I saw this behavior on Ubuntu. Not sure what other systems do.

certbot/error_handler.py
cowlicks: Yes this was to defend against that. But I don't know how effective it is.

cowlicks:
The travis failure seems to be unrelated. It is happening for other recent PR's.

bmw:
Travis failures were unrelated to this PR so I kicked the build to make them pass. Sorry for the trouble!

cowlicks:
@bmw I think I addressed all your comments. Anything else?

certbot/error_handler.py
bmw: So my understanding is that with this change, we increase the likelihood that the code currently in the `finally` block gets called, but decrease the probability that the code in the `try` block gets called. This decrease is basically due to adding more code in `__exit__` that must run before we set `body_executed`.

With this in mind, I personally think we should remove the `try`/`finally` at least for now. After this is merged I'm going to make a followup issue in the wishlist milestone to discuss things we could theoretically do to make the `ErrorHandler` even more fail proof, but I'd like to worry about things like that later as this PR is already a significant improvement over the current code and I'd like to see it merged.

certbot/tests/error_handler_test.py
```diff
@@ -1,10 +1,37 @@
...
 import mock
 
+def get_signals(signums):
+    """Get the handlers for an iterable of signums."""
+    return dict((s, signal.getsignal(s)) for s in signums)
+
+
+def set_signals(sig_handler_dict):
+    """Set the signal (keys) with the handler (values) from the input dict."""
+    tuple(signal.signal(s, h) for (s, h) in sig_handler_dict.items())
```
bmw: We shouldn't use the comprehension just to do this in one line. New lines bytes are cheap! This should just be a simple for loop that calls `signal.signal` on each key/value pair.

certbot/tests/error_handler_test.py
```diff
@@ -1,10 +1,37 @@
...
+
+@contextlib.contextmanager
+def signal_receiver(signums):
+    """Context manager to catch signals"""
+    signals = []
+    prev_handlers = {}
+    prev_handlers = get_signals(signums)
+    set_signals(dict((s, lambda s, _: signals.append(s)) for s in signums))
+    yield signals
+    set_signals(dict((s, prev_handlers[s]) for s in signums))
```
bmw: We should be able to just pass in `prev_handlers` to `set_signals` instead of building a new dict.

bmw:
Sorry for the delay @cowlicks. I've had a pretty busy week.

I made 3 comments above, all of which are minor changes. If you make those changes I'll merge this PR. When you want a final review, remove "[has suggested revision]" from the title and I'll take a look.

certbot/tests/error_handler_test.py
cowlicks: Addressed https://github.com/certbot/certbot/pull/3223/files#diff-4a585967926ff0bb7ba8f3918dc9d91cR52

bmw:
I think the only remaining comment is the one about the `try`/`finally`.

cowlicks:
@bmw Whoops, I missed that sorry. Got it now.

bmw:
LGTM. Thanks for this @cowlicks!

PR #3226, opened 2016-06-29
FreeBSD port is now py-certbot

bmw:
https://www.freebsd.org/cgi/ports.cgi?query=certbot&stype=all

Although it appears the package is still named `letsencrypt` (and is stuck on 0.4.1).


pde:
LGTM

PR #3230, opened 2016-06-30
Permissive parsing of nginx map blocks

### LINKED ISSUE #3206: Nginx parser apparently can't parse "map"
I was looking at issues on the nginxparser project and found this one:

https://github.com/fatiherikli/nginxparser/issues/19

It appears to me that we can't parse configurations containing this. I got something like

`ParseException: Expected stringEnd (at char 100), (line:5, col:1)`

when trying to `nginxparser.loads()` a file that included that map directive.

cc @pde

pde:
Closes #3206, at least in a good-enough-for-now sense.


pde:
This still fails on [this](https://www.nginx.com/resources/wiki/start/topics/recipes/wordpress/#) map statement:
```
map $uri $blogname{
    ~^(?P<blogpath>/[^/]+/)files/(.*)       $blogpath ;
}
```

pde:
(The problem is the lack of whitespace between `$blogname` and `{`.)

pde:
@schoen fixed. Please use this branch for evaluating the present correctness of the parser; although we know it's a little janky, that is even more true of the parser in `master` :)

pde:
Kicking travis.

bmw:
Travis failures were unrelated to this PR so I kicked the build to make them pass.

schoen:
We've discussed the issue that the parser is still not correct in general and that we're tending to get our notion of the grammar from test cases rather than from a formal specification. So we know, for example, that this solution is not necessarily completely correct.

It does seem to be an improvement. For the roundtrip testing, it passes 2 more test cases than the parser without it. I also don't know of any new problems it would introduce.

We definitely need to keep an eye on ways to improve and check the correctness of the parser in the future, though.

schoen:
I've added a test case for the

```
map $uri $blogname{
    ~^(?P<blogpath>/[^/]+/)files/(.*)       $blogpath ;
}
```

thing (above at https://github.com/certbot/certbot/pull/3230#issuecomment-231546594).

Issue #2782, opened 2016-04-07
Implement a "script" plugin and/or fix manual for scriptability


aug-riedinger:
I guess I'm not the first one to raise the issue, but I haven't found anything yet by searching.

I'm trying to automate VM deployment with a letsencrypt certificate. So I have tried to remove user interaction as much as possible and reached:

```
sudo /opt/letsencrypt/letsencrypt-auto certonly --manual -m contact@example.com -d example.com -t --manual-public-ip-logging-ok --agree-tos
```

But if I add the `-n` option, I get:

> `PluginError('Running manual mode non-interactively is not supported',)`

I'm guessing the point is that manual mode _does_ require user interaction hence cannot be fully done in non-interactive mode. At least during the file-validation step.

Which is why I'd like to discuss splitting the command in two steps:

```
sudo /opt/letsencrypt/letsencrypt-auto certonly --manual -m contact@example.com -d example.com -t --manual-public-ip-logging-ok --agree-tos --generate-file /home/user
```

which would create the file `/home/user/.well-known/acme-challenge/buCgmexKDQypJfWE4-XmUfP1cpILzsBR4UWRE_o8iuc` with the expected content and then running

```
sudo /opt/letsencrypt/letsencrypt-auto certonly --manual -m contact@example.com -d example.com -t --manual-public-ip-logging-ok --agree-tos --validate
```

would finish the process.

Would this make sense? Did I miss something?

Another option would be to build allow a script to be executed between the file generation step and the verification step:

```
sudo /opt/letsencrypt/letsencrypt-auto certonly --manual -m contact@example.com -d example.com -t --manual-public-ip-logging-ok --agree-tos --deploy-verification ./deploy-verification.sh
```

bmw:
This certainly does make sense. #2610 also suggests the script approach, but there are benefits to having two different client invocations as well. We certainly want to implement something like this in the future.

pde:
@joohoi is actually working on this, in part to make support for DNS challenges practical.  I don't think he's decided on a a design yet, though some of the options under discussion include:

* Run a "Step 1" command that gives you a challenge, and then a "Step 2" command when you've deployed it.  Challenge: how to serialise and save state so that the client knows where to pick up from.
* Call the client with two hook functions, one of which is to be run when a challenge is ready, the other when it has succeeded.  Challenge: does this mean that the user has to write three separate scripts, one for each hook and one to call the client with the hooks?

Another option that @bmw and I talked about recently is:

[...]

PR #3236, opened 2016-07-03
add -D/--webroot-delay-auth option

NiklasRosenstein:
With this new option enabled, the "webroot" plugin will wait for user
input before the challanges will be verified.

This is useful for creating certificates for shared hosting plans
without root (or even no SSH access at all): The challenges can be
proposed from a foreign machine, and with this option, the user has
the time to copy the validation files to the actual webserver before
the challanges are verified.

### DIFF
--- a/certbot/plugins/webroot.py
+++ b/certbot/plugins/webroot.py
@@ -29,6 +29,12 @@ class Authenticator(common.Plugin):
 
     description = "Place files in webroot directory"
 
+    MSG_AUTH_DELAYED = """\
+Authentication delayed. Deploy the {0}/.well-known directory to \
+your webroot now.
+
+Press ENTER to continue to the validation process."""
+
     MORE_INFO = """\
 Authenticator plugin that performs http-01 challenge by saving
 necessary validation resources to appropriate paths on the file
@@ -82,7 +88,9 @@ to serve all files under specified web root ({0})."""
         responses = [self._perform_single(achall) for achall in achalls]
 
         if self.conf("delay-auth"):
-            six.moves.input("Press ENTER to continue authentication...")
+            path = self.conf("path")[0]
+            display = zope.component.getUtility(interfaces.IDisplay)
+            display.notification(self.MSG_AUTH_DELAYED.format(path))
 
         return responses
 
--- a/certbot/plugins/webroot.py
+++ b/certbot/plugins/webroot.py
@@ -55,6 +55,12 @@ to serve all files under specified web root ({0})."""
                  "-d entries. At present, if you put webroot-map in a config "
                  "file, it needs to be on a single line, like: webroot-map = "
                  '{"example.com":"/var/www"}.')
+        add("delay-auth", "-D", action="store_true",
+            help="Wait until the user presses ENTER before the "
+                 "authentication process kicks in. This is useful when "
+                 "proposing the challenges from a foreign machine to be "
+                 "able to copy the files to the real webserver before "
+                 "the challenges are verified.")
 
     def get_chall_pref(self, domain):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


bmw:
Have you seen the [manual plugin](https://certbot.eff.org/docs/using.html#manual)? While we need to improve the UI, this provides the user with instructions for how to setup their webserver to complete the challenge and waits for user input before continuing.

What benefit does the functionality added in this PR have over `manual`?

NiklasRosenstein:
Hi! Yes, I've know about the manual plugin. The problem is this:

>  **https://certbot.eff.org/docs/using.html#manual**
> This requires you to copy and paste commands into another terminal session, which may be on a different computer.

Manually copy & pasting the verification URL and content is much too tedious. The `webroot` plugin automatically creates the files in the right folder structure. It's just a matter of uploading the `.well-known` folder to your `public_html/` directory.

Of course, the same could be achieved by adding an option to the `manual` plugin that would create this folder structure instead of prompting you to copy & paste the verification codes, though I figured it is much easier implemented this way.

NiklasRosenstein:
Also, I was thinking about adding an option that allows you to run a script before the validation process kicks in. This would allow you to completely automate the certificate generation from a remote machine.

    certbot-auto certonly -a webroot --webroot-uploader the_upload_script.sh -d mydomain.com

What do you think?

certbot/plugins/webroot.py
```diff
@@ -73,7 +79,12 @@ def perform(self, achalls):  # pylint: disable=missing-docstring
 
         self._create_challenge_dirs()
 
-        return [self._perform_single(achall) for achall in achalls]
+        responses = [self._perform_single(achall) for achall in achalls]
+
+        if self.conf("delay-auth"):
+            six.moves.input("Press ENTER to continue authentication...")
```
pde: Please the [IDisplay](https://github.com/certbot/certbot/blob/master/certbot/interfaces.py#L375) interface for this task.  An example of how to do that is [here](https://github.com/certbot/certbot/blob/master/certbot/plugins/webroot.py#L152) but you will want to set `pause=True`.

NiklasRosenstein: Thanks! Will change that as soon as I find the time.

bmw:
> Also, I was thinking about adding an option that allows you to run a script before the validation process kicks in. This would allow you to completely automate the certificate generation from a remote machine.

We are actually planning on making a more general `script` plugin (or make `manual` scriptable) as described in #2782. @joohoi, was going to work on this, but if you're interested, talk to him and you may be able to (help) work on it.

NiklasRosenstein:
@pde I've made the changes that you suggested in commit fb15222. It looks much nicer now, too. :)

image

@bmw Thanks for linking the other issue. A separate `script` plugin sounds perfectly valid to me as well. It can probably reuse a lot of the `webroot` plugin's code for creating the directory structure and files. I'd be happy to help.

For now, IMHO, this PR makes things a lot more easy especially for websites on shared hosting plans, even more so with multiple domains/subdomains. With the `manual` plugin, you have to do the validation process for each and every domain manually, while the `webroot` plugin creates *all* verification files before the validation.

NiklasRosenstein:
Currently the test cases fail because there is no user input. Even if you're considering not to merge this PR, I'd appreciate if you could instruct me on how to resolve this issue correctly. Solutions range from testing whether a TTY is attached to updating the test cases. :)

bmw:
Sorry this PR has sat around so long. Rather than taking this, I think we're going to go with the script plugin as mentioned before (whose PR is #3521) as well as a more general flag for pausing Certbot after the challenges have been set up as proposed in #3679. If you'd like to work on this, we'll happily take a look at another PR.

Issue #3237, opened 2016-07-05
Cleanup "Other methods of installation"


bmw:
We should clean up https://certbot.eff.org/docs/using.html#other-methods-of-installation.

In particular, we currently list instructions for OS packages in two places which makes it harder for them to stay up to date/in sync (especially since they are in two different repos). Additionally, I think some of this information should be changed. For example, we say

> Unless you have a very specific requirements, we kindly suggest that you use the certbot-auto method.

I think this is incorrect. In almost all cases, OS packages should be preferred over `certbot-auto`.

pconrad-fb:
I'll put this on my list.

bmw:
Great! Please let me know if you have any questions.

pconrad-fb:
Questions:

1. Running with Docker: 
You can't install a cert that you get via Docker because the plugins can't reach into the Docker container to get it. We should add a sentence that starts with "If you use Docker, you'll have to..." and then explains what the docker run command below will do. What does it do?
Here, we're also recommending certbot-auto. Should we be recommending OS package instructions instead? In what cases should we recommend certbot-auto?

2. Operating System Packages:
I understand that on Debian, Ubuntu, and other Linux distros, the package exists in one of the default repositories and you can just apt-get or yum install or whatever. On FreeBSD and OpenBSD, does the /usr/ports/security/py-letsencrypt directory exist right out of the box?

3. Proposed change based on your comments:
Change "Unless you have very specific requirements" to "If a package does not exist for your operating system".

bmw:
1) Running with Docker:

> You can't install a cert that you get via Docker because the plugins can't reach into the Docker container to get it.

Actually, all of Certbot, including the plugins, are running inside of Docker. The problem is the plugins can't reach outside of the container. So, for example, if you're running an Apache webserver and then use Certbot, Certbot is unable to reach outside of Docker to modify Apache.

> We should add a sentence that starts with "If you use Docker, you'll have to..." and then explains what the docker run command below will do. What does it do?

Using the long `sudo docker run ...` command given in our documentation, Certbot will obtain a certificate and place it in the directories under `/etc/letsencrypt` on your system. The [Where are my certificates?](https://certbot.eff.org/docs/using.html#where-certs) linked in to in our Docker documentation further explains the layout of this directory. After running Certbot, the user will have to manually install their new certificates (because as explained above, Certbot is unable to do so while running inside of Docker).

[...]

PR #3238, opened 2016-07-05
Update FreeBSD package name

bmw:
Other half of certbot/website#144. I also created #3237 for tracking the issue of needing a PR on two different repos to update OS instructions.


pde:
LGTM.

Issue #3244, opened 2016-07-06
Unexpected error: NoInstallationError()


bigretromike:
Hello, 
been using letsencrypt for some time now, after rename I deleted /opt/letsencrypt and moved to apt-get install certbot.
Debian 8.5

from time to time im getting email (on root):
WARNING:certbot.renewal:Attempting to renew cert from /etc/letsencrypt/renewal/DOMAIN.conf produced an unexpected error: The apache plugin is not working; there may be problems with your exisitng configuration. The error was: NoInstallationError(). Skipping. The following certs are not sue for renewal yet:
/etc/letsencrypt/live/domain2/fullchain.pem (skipped)
/etc/letsencrypt/live/domain3/fullchain.pem (skipped)

All renewal attempts failed. The following certs could not be renewed:
/etc/letsencrypt/live/DOMAIN/fullchain.pem (failure)

conf is:
```
cert = /etc/letsencrypt/live/DOMAIN/cert.pem
privkey = /etc/letsencrypt/live/DOMAIN/privkey.pem
chain = /etc/letsencrypt/live/DOMAIN/chain.pem
fullchain = /etc/letsencrypt/live/DOMAIN/fullchain.pem

# Options used in the renewal process
[renewalparams]
authenticator = apache
installer = apache
account = account_hash
```

Fun part is that running 'dry' or by hand resolve this problem until the time it hits it again.

```
DRY RUN: simulating 'certbot renew' close to cert expiry
          (The test certificates below have not been saved.)

Congratulations, all renewals succeeded. The following certs have been renewed:
  /etc/letsencrypt/live/domain2/fullchain.pem (success)
  /etc/letsencrypt/live/DOMAIN/fullchain.pem (success)
  /etc/letsencrypt/live/domain3/fullchain.pem (success)
```

Package: certbot
Source: python-certbot
Version: 0.8.0-1~bpo8+2

Package: python-certbot-apache
Version: 0.8.0-1~bpo8+1

pde:
This is very odd. It's really hard to see why you would get this error from the cron job in `/etc/cron.daily/`, but not while running interactively.  Although one possibility is that there are still some remnants of a previous install somewhere (did you possibly use `pip` to install?) and that differences in your environment variables are causing [parts of] that old copy to be run from cron but not from the shell.

pde:
[...]

Issue #1833, opened 2015-12-08
NoInstallationError() from Apache plugin within renewal cron jobs due to /usr/sbin not being in the PATH


linuxlurak:
Im running a cron job that executes a script with the following content:
```
#!/bin/sh
echo "" >> /var/log/letsencrypt/$1.log
echo "$(date): Certificate Renewal of $1">> /var/log/letsencrypt/$1.log
echo "-----------------------------------------------------------------"
/root/letsencrypt/letsencrypt-auto certonly --apache --rsa-key-size 4096 --renew-by-default --hsts --redirect --agree-tos -m EMAIL.TLD -d $1 --text >> /var/log/letsencrypt/$1.log 2>&1
```

I get this Error:
```
Running with virtualenv: /root/.local/share/letsencrypt/bin/letsencrypt certonly --apache --rsa-key-size 4096 --renew-by-default --hsts --redirect --agree-tos -m EMAIL.TLD -d DOMAIN.TLD --text
The apache plugin is not working; there may be problems with your existing configuration.
The error was: NoInstallationError()"
```

Anything i can do?

pde:
Run interactively with -vvv and --debug, to get more detail on what the problem is.  I suspect but am not sure that the apache plugin is crashing on one of your conf files; if you can figure out which one, can you paste it?

linuxlurak:
Debug log:
```
2015-12-09 18:57:29,688:DEBUG:letsencrypt.cli:Root logging level set at 0
2015-12-09 18:57:29,689:INFO:letsencrypt.cli:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2015-12-09 18:57:29,689:DEBUG:letsencrypt.cli:letsencrypt version: 0.1.0
2015-12-09 18:57:29,690:DEBUG:letsencrypt.cli:Arguments: ['--debug', '-vvv', '--apache', '--rsa-key-size', '4096', '--renew-by-default', '--hsts', '--redirect', '--agree-tos', '-m', 'REMOVED', '-d', 'REMOVED', '--text']
2015-12-09 18:57:29,690:DEBUG:letsencrypt.cli:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2015-12-09 18:57:29,692:DEBUG:letsencrypt.cli:Requested authenticator apache and installer apache
[...]

PR #3249, opened 2016-07-06
Explain why Apache [appears] not to be installed

pde:
Would help debug #3244


bmw:
LGTM. This also makes progress on #1833.

Issue #1460, opened 2015-11-11
HelpfulArgumentParser should know about flags that are relevant to several topics


pde:
We currently have a bunch of hairy special cases where flags are documented in different places depending on which help topics are being displayed.  These are bad to maintain, so we should either have HelpfulArgumentParser know about and organise them automatically (alternatively, we chould completely rearchitect cli.py to give good concise topic-based help in some better way, such as writing our own argument parser, forking [config]argparse in the tree, or something else..).

PR #3254, opened 2016-07-07
Multi-topic help listings

pde:
Closes: #1460


pde:
@bmw after fighting Travis a lot, this should be turning green any second now...

certbot/cli.py
```diff
@@ -699,14 +732,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         version="%(prog)s {0}".format(certbot.__version__),
         help="show program's version number and exit")
     helpful.add(
-        "automation", "--force-renewal", "--renew-by-default",
+        ["automation", "renew"],
```
bmw: This should probably be under `certonly` too.

certbot/cli.py
```diff
@@ -699,14 +732,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+        "--force-renewal", "--renew-by-default",
         action="store_true", dest="renew_by_default", help="If a certificate "
              "already exists for the requested domains, renew it now, "
              "regardless of whether it is near expiry. (Often "
              "--keep-until-expiring is more appropriate). Also implies "
              "--expand.")
     helpful.add(
-        "automation", "--allow-subset-of-names", action="store_true",
+        ["automation", "renew"],
```
bmw: Perhaps this should also be under `certonly`. The flag works with `certonly` and we recommend `certonly` as an [alternative to renew](https://certbot.eff.org/docs/using.html#renewal).

certbot/cli.py
```diff
@@ -730,14 +765,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="(certbot-auto only) prevent the certbot-auto script from"
              " upgrading itself to newer released versions")
     helpful.add(
-        "automation", "-q", "--quiet", dest="quiet", action="store_true",
+        ["automation", "renew"],
```
bmw: Maybe `certonly`. I care less about this one, but there are a number of people still using `certonly` for renewal. We should try to stop people from doing that, but in the meantime, perhaps it should be listed here?

certbot/cli.py
```diff
@@ -944,19 +953,17 @@ def _paths_parser(helpful):
...
     add("paths", "--server", default=flag_default("server"),
         help=config_help("server"))
-    # overwrites server, handled in HelpfulArgumentParser.parse_args()
-    add("testing", "--test-cert", "--staging", action='store_true', dest='staging',
-        help='Use the staging server to obtain test (invalid) certs; equivalent'
-             ' to --server ' + constants.STAGING_URI)
 
 
 def _plugins_parsing(helpful, plugins):
+    # It's nuts, but there are two "plugins" topics.  Somehow this works
```
bmw: That is nuts. Looks like we do this in `master` too which I didn't realize. Perhaps we should change the name of the `plugins` subcommand?

certbot/cli.py
```diff
@@ -500,16 +499,23 @@ def prescan_for_flag(self, flag, possible_arguments):
             pass
         return True
 
-    def add(self, topic, *args, **kwargs):
+    def add(self, topics, *args, **kwargs):
```
bmw: nit: My understanding is if `--help all` is used and multiple topics are specified, the flag will show up in the first listed topic. Perhaps this should be documented?

certbot/cli.py
```diff
@@ -500,16 +499,23 @@ def prescan_for_flag(self, flag, possible_arguments):
...
         """Add a new command line argument.
 
-        :param str: help topic this should be listed under, can be None for
-                    "always documented"
+        :param topics: str or [str] help topic(s) this should be listed under,
+                       or None for "always documented"
```
bmw: Question: This isn't new, but how is this supposed to work? It seems that the flags that are "always documented" only show up under all help output.

bmw:
In general, I'm really happy with this PR! Very simple and seems to work well.

Other than a few nits/questions above, I think can add flags to more help topics. As examples:

* In general, perhaps every `renew` flag should be under `certonly`.
* `-n` should probably be under automation and maybe even renewal and certonly.
* `--register-unsafely-without-email` and `--update-registration` should be under register. `--register-unsafely-without-email` perhaps should be under automation?

Not all of this has to be done now and if you disagree on certain points feel free to ignore them, however, I think we can make our help output even more helpful!

certbot/cli.py
bmw: To clarify, I don't think this should be done in this PR. It's just a bit awkward that `-h plugins` gives you help for two different things. If you agree, please open an issue or comment here and I'll do so.

certbot/cli.py
pde: I think my prior idea was that the topics are really deep dives if you want to know about stuff that's specific to that invocation of the client, rather than (say) exhaustively listing things that are relevant to each subcommand.  Though perhaps there could also be a case for making the topic help into something more like the git manpages for subcommands, which list every possible flag for each of them?

certbot/cli.py
pde: I think my prior idea was that the topics are really deep dives if you want to know about stuff that's specific to that invocation of the client, rather than (say) exhaustively listing things that are relevant to each subcommand.  Though perhaps there could also be a case for making the topic help into something more like the git manpages for subcommands, which list every possible flag for each of them?

certbot/cli.py
pde: I skipped this because I feel that it's rarely the correct thing for anyone to be using...

certbot/cli.py
pde: I'm indifferent. I really don't think there are many use cases for the `plugins` subcommand at all. I've used it for debugging during plugin development occasionally, I guess. I imagine someone might use it to see what plugins they have, but I've never done that...

pde: When I try that, it hides `manual` from the list, and also prints spurious ncurses output.  So it's not getting a lot of love at the moment : )

bmw:
Looks like there's still a merge conflict due to #3397 being merged.

Other than that, this LGTM.

PR #3255, opened 2016-07-07
Reference certbot-auto in CLI help

rbu:
Do these commands really only apply to the old letsencrypt-auto? If not, the help text should be updated as proposed.


pde:
Thanks!

PR #3258, opened 2016-07-08
Clarifications to Docker, certbot-auto content reflecting first three…

pconrad-fb:
Edits reflect answers to the first three questions in my comment https://github.com/certbot/certbot/pull/3232#issuecomment-231154320
provided by Brad Warren

Known deficiencies:

- Link to certbot-auto just goes to the "name of the certbot command" section, which doesn't tell what certbot-auto is or how to use it. We need to add a cookbook of sorts somewhere.
- I haven't locally tested the RST with Sphinx (but it should be OK (but test there))

I also see more comments on my comments, so I'll try to make more progress tomorrow or next week.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -9,9 +9,10 @@ Getting Certbot
 ===============
 
 To get specific instructions for installing Certbot on your OS,
-visit certbot.eff.org_. This is the easiest way to install Certbot. 
+visit certbot.eff.org_. This is the easiest way to learn how to get
+Certbot up and running on your system. 
 
-If you're offline, or if your webserver or OS are not in the menu, you can find some general
+If you're offline, you can find some general
 instructions `in the README / Introduction <intro.html#installation>`__
 
 __ installation_
@@ -398,7 +399,7 @@ certificate. However, this mode of operation is unable to install
 certificates or configure your webserver, because our installer
 plugins cannot reach your webserver from inside the Docker container.
 
-Most users should use the operating system packages (available from 
+Most users should use the operating system packages (see instructions at 
 certbot.eff.org_) or, as a fallback, ``certbot-auto``. You should only 
 use Docker if you are sure you know what you are doing and have a
 good reason to do so.
@@ -420,12 +421,12 @@ to, `install Docker`_, then issue the following command:
    sudo docker run -it --rm -p 443:443 -p 80:80 --name certbot \
                -v "/etc/letsencrypt:/etc/letsencrypt" \
                -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
-               quay.io/letsencrypt/letsencrypt:latest auth
+               quay.io/letsencrypt/letsencrypt:latest certonly
 
 Certbot will obtain a certificate and place it in the directory
-``/etc/letsencrypt/live`` on your system and display further instructions 
-for installing the certificates. You must use the ``auth`` command
-to install the certificates instead of plug-ins for this method. 
+``/etc/letsencrypt/live`` on your system. 
+You must use the ``certonly`` command
+to install the certificate. 
 
 For more information about the layout 
 of the ``/etc/letsencrypt`` directory, see :ref:`where-certs`. 
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -8,14 +8,16 @@ User Guide
 Getting Certbot
 ===============
 
-To get specific instructions for installing Certbot on your OS, we recommend
-visiting certbot.eff.org_. If you're offline, you can find some general
+To get specific instructions for installing Certbot on your OS,
+visit certbot.eff.org_. This is the easiest way to install Certbot. 
+
+If you're offline, or if your webserver or OS are not in the menu, you can find some general
 instructions `in the README / Introduction <intro.html#installation>`__
 
 __ installation_
 .. _certbot.eff.org: https://certbot.eff.org
 
-.. _certbot-auto:
+.. _certbot-auto: https://certbot.eff.org/docs/using.html#certbot-auto
 
 The name of the certbot command
 -------------------------------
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -8,14 +8,16 @@ User Guide
 Getting Certbot
 ===============
 
-To get specific instructions for installing Certbot on your OS, we recommend
-visiting certbot.eff.org_. If you're offline, you can find some general
+To get specific instructions for installing Certbot on your OS,
+visit certbot.eff.org_. This is the easiest way to install Certbot.
```
bmw: I wonder if this should be something like "This is the easiest way to learn how to get Certbot up and running on your system." I'm suggesting this because visiting the website only gives you instructions on how to do it and doesn't actually install the software. If you think clarifying this sentence isn't necessary, feel free to ignore this comment.

docs/using.rst
```diff
@@ -8,14 +8,16 @@ User Guide
 Getting Certbot
 ===============
 
-To get specific instructions for installing Certbot on your OS, we recommend
-visiting certbot.eff.org_. If you're offline, you can find some general
+To get specific instructions for installing Certbot on your OS,
+visit certbot.eff.org_. This is the easiest way to install Certbot. 
+
+If you're offline, or if your webserver or OS are not in the menu, you can find some general
```
bmw: Ideally, everyone's webserver or OS should be in the menu. We have a catch all for webserver titled "None of the above". While we have "Other UNIX", we don't say anything about non-UNIX systems on the site. I created certbot/website#147 to track that issue.

With this in mind, I think this line should remain "If you're offline, you can find some general..."

docs/using.rst
```diff
@@ -394,7 +396,12 @@ Running with Docker
 Docker_ is an amazingly simple and quick way to obtain a
 certificate. However, this mode of operation is unable to install
 certificates or configure your webserver, because our installer
-plugins cannot reach it from inside the Docker container.
+plugins cannot reach your webserver from inside the Docker container.
+
+Most users should use the operating system packages (available from 
+certbot.eff.org_) or, as a fallback, ``certbot-auto``. You should only
```
bmw: So the packages aren't actually available on https://certbot.eff.org. However, that website gives you instructions how to obtain those packages from your operating system.

docs/using.rst
```diff
@@ -394,7 +396,12 @@ Running with Docker
 Docker_ is an amazingly simple and quick way to obtain a
 certificate. However, this mode of operation is unable to install
 certificates or configure your webserver, because our installer
-plugins cannot reach it from inside the Docker container.
+plugins cannot reach your webserver from inside the Docker container.
+
+Most users should use the operating system packages (available from 
+certbot.eff.org_) or, as a fallback, ``certbot-auto``. You should only 
+use Docker if you are sure you know what you are doing and have a
+good reason to do so.
```
bmw: Perfect!

docs/using.rst
```diff
@@ -415,9 +422,13 @@ to, `install Docker`_, then issue the following command:
                -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
                quay.io/letsencrypt/letsencrypt:latest auth
 
-and follow the instructions (note that ``auth`` command is explicitly
-used - no installer plugins involved). Your new cert will be available
-in ``/etc/letsencrypt/live`` on the host.
+Certbot will obtain a certificate and place it in the directory
+``/etc/letsencrypt/live`` on your system and display further instructions 
+for installing the certificates. You must use the ``auth`` command
+to install the certificates instead of plug-ins for this method.
```
bmw: Few thoughts on this section.

1. We don't actually display any information about how to install the certificates. Doing so is extremely webserver specific. We have a bit of that information under [where are my certificates](https://certbot.eff.org/docs/using.html#where-are-my-certificates).
2. `auth` doesn't install certificates, it simply obtains them. Actually, `auth` is simply an alias for the `certonly`  command which we switched to after talking to some usability researchers (but kept the old command for backwards compatibility). We should actually change references of `auth` here to `certonly`. We can do this in a separate PR if you prefer.
3. I think we should be consistent with our use of plugins vs plug-ins. We normally use plugins, but we could potentially switch if you think there's value in doing so.
4. Certbot still actually uses the plugin system here. It just doesn't use any installer plugins. Take a look at our [plugin documentation](https://certbot.eff.org/docs/using.html#plugins) to see the distinction.

bmw:
I added a few comments with clarifications and things I'd like to see changed. Please let me know if you have any questions.

docs/using.rst
```diff
@@ -413,11 +421,15 @@ to, `install Docker`_, then issue the following command:
    sudo docker run -it --rm -p 443:443 -p 80:80 --name certbot \
                -v "/etc/letsencrypt:/etc/letsencrypt" \
                -v "/var/lib/letsencrypt:/var/lib/letsencrypt" \
-               quay.io/letsencrypt/letsencrypt:latest auth
+               quay.io/letsencrypt/letsencrypt:latest certonly
+
+Certbot will obtain a certificate and place it in the directory
+``/etc/letsencrypt/live`` on your system. 
+You must use the ``certonly`` command
+to install the certificate.
```
bmw: Like `auth`, `certonly` doesn't install your certificates to your web server.

To flesh out this distinction a little more, the default behavior of Certbot is the `run` subcommand. This is used if the user types `certbot run` or even just `certbot` because the default subcommand is `run`. What `run` does is prove to the Let's Encrypt certificate authority that you own the domain name(s) you requested, downloads the certificates issued by Let's Encrypt, and then installs them in your webserver.

`certonly` on the other hand omits this last step. `certbot certonly` will prove ownership and get certificates from Let's Encrypt and save them on your computer. After doing that though, the user has to install them to their webserver software themselves.

Assuming you're wanting to explain the use of `certonly`, I think we may be able to just combine this and the previous sentence to say:

"Running Certbot with the ``certonly`` subcommand will obtain a certificate and place it in the directory ``/etc/letsencrypt/live`` on your system."

I think this is kind of nice because it flows into the next sentence about providing more information about `/etc/letsencrypt/live.` An alternative more like what you have here would be:

"...on your system. You must use the ``certonly`` command to get your certificate because as explained above, Certbot is unable to install your certificate from inside Docker."

bmw:
I have one comment about the use of `certonly` with Docker, but other than that is looks good!

pconrad-fb:
Cool. I'll take a look in a bit.

pconrad-fb:
How about this:

> Running Certbot with the ``certonly`` command will obtain a certificate and place it in the directory
``/etc/letsencrypt/live`` on your system. Because Certonly cannot install the certificate from 
within Docker, you must install the certificate manually according to the procedure
recommended by the provider of your webserver.

bmw:
That sounds fantastic. If you make that change, I'll merge this PR.

bmw:
LGTM. I'll merge this PR once our tests finish.

Thanks for writing this @pconrad-fb!

pconrad-fb:
Awesome.

PR #3259, opened 2016-07-08
Automatically enable EPEL after prompting users

### LINKED ISSUE #2158: Enable EPEL in RPM bootstrapper
[Some letsencrypt-auto runs on CentOS fail](https://community.letsencrypt.org/t/help-test-the-new-letsencrypt-auto-script/8544/4) due to the EPEL repository being disabled, which [is the default for CentOS and perhaps others](https://community.letsencrypt.org/t/help-test-the-new-letsencrypt-auto-script/8544/4). We should probably enable EPEL in our bootstrap script, since we enable Debian Backports in our Debian one.

bmw:
Fixes #2158.

I tested all branches of this on CentOS 6 and tested the script on CentOS 7 (EPEL doesn't need to be enabled on CentOS 7 so there's not much to test). I also tested this on the test farm, however, it appears EPEL is already enabled so I think the value of testing there is fairly minimal.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -285,7 +285,7 @@ BootstrapRpmCommon() {
     yes_flag="-y"
   fi
 
-  if ! $SUDO $tool list *virtualenv > /dev/null 2>&1; then
+  if ! $SUDO $tool list *virtualenv >/dev/null 2>&1; then
     echo "To use Certbot, packages from the EPEL repository need to be installed."
     if [ "$ASSUME_YES" = 1 ]; then
         /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -287,6 +287,10 @@ BootstrapRpmCommon() {
 
   if ! $SUDO $tool list *virtualenv >/dev/null 2>&1; then
     echo "To use Certbot, packages from the EPEL repository need to be installed."
+    if ! $SUDO $tool list epel-release >/dev/null 2>&1; then
+      echo "Please enable this repository and try running Certbot again."
+      exit 1
+    fi
     if [ "$ASSUME_YES" = 1 ]; then
         /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
         sleep 1s
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -281,6 +281,26 @@ BootstrapRpmCommon() {
     exit 1
   fi
 
+  if [ "$ASSUME_YES" = 1 ]; then
+    yes_flag="-y"
+  fi
+
+  if ! $SUDO $tool list *virtualenv > /dev/null 2>&1; then
+    echo "To use Certbot, packages from the EPEL repository need to be installed."
+    if [ "$ASSUME_YES" = 1 ]; then
+        /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
+        sleep 1s
+        /bin/echo -ne "\e[0K\rEnabling the EPEL repository in 2 seconds..."
+        sleep 1s
+        /bin/echo -e "\e[0K\rEnabling the EPEL repository in 1 seconds..."
+        sleep 1s
+    fi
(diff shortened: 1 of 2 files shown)


pde:
@NoodlesNZ or @hogarthj would either of you (or someone else from your team) be interested in reviewing this?  This is realistically going to be the way that CentOS 6 users get to install Certbot for the rest of time...

bmw:
Just noticed that I had some changes locally that weren't included here. I've updated this PR.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -281,6 +281,26 @@ BootstrapRpmCommon() {
...
 
+  if [ "$ASSUME_YES" = 1 ]; then
+    yes_flag="-y"
+  fi
+
+  if ! $SUDO $tool list *virtualenv > /dev/null 2>&1; then
+    echo "To use Certbot, packages from the EPEL repository need to be installed."
+    if [ "$ASSUME_YES" = 1 ]; then
+        /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
+        sleep 1s
```
erikrose: Is the `s` an actual thing? I can't find it documented anywhere, and OS X, at least, just ignores the s. You can `sleep 1q` with equal results.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
```diff
@@ -17,6 +17,26 @@ BootstrapRpmCommon() {
...
+    echo "To use Certbot, packages from the EPEL repository need to be installed."
+    if [ "$ASSUME_YES" = 1 ]; then
+        /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
+        sleep 1s
+        /bin/echo -ne "\e[0K\rEnabling the EPEL repository in 2 seconds..."
+        sleep 1s
+        /bin/echo -e "\e[0K\rEnabling the EPEL repository in 1 seconds..."
+        sleep 1s
+    fi
+    if ! $SUDO $tool install $yes_flag epel-release; then
```
erikrose: Have you considered that RHN users [need to enable additional repos](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F) in order for EPEL to work?

erikrose:
That's all I see. Otherwise, it looks fine.

letsencrypt-auto-source/letsencrypt-auto
pde: `sleep` on Fedora etc is GNU sleep, which supports the `s` (as well as `m` and `h`).

pde:
@bmw I think this is back in your court to consider enabling the "optional" channel for RHN users.

pde:
@bmw it may be that we can kick the RHN piece into a future ticket. If you agree, go ahead and merge this!

bmw:
I'm really struggling testing this on a Red Hat machine with RHN. Does anyone else want to take a crack at it?

bmw:
So it appears that you can only `yum install epel-release` on CentOS. RHEL users are told to `wget` a package and install it directly through `rpm`.

Due to this oversight on my part, we will error out on RHEL systems (or CentOS systems that don't include the `extras` repository which is [enabled by default](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F)) for now with a message asking the user to enable EPEL. This is a better user experience anyway than erroring out later with an error message like `virtualenv: command not found`. I created #3433 to track the issue of automatically enabling EPEL on RHEL.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -287,6 +287,10 @@ BootstrapRpmCommon() {
 
   if ! $SUDO $tool list *virtualenv >/dev/null 2>&1; then
     echo "To use Certbot, packages from the EPEL repository need to be installed."
+    if ! $SUDO $tool list epel-release >/dev/null 2>&1; then
```
pde: As long as we're sure that this package will *always* be visible on RPM-based systems where `certbot-auto` works.

pde: If we're not, I guess we can do something like make this a warning?

pde: Actually I guess this is fine, because we're inside the `if` about virtualenv. So overall, LGTM.

Issue #3148, opened 2016-06-10
No handlers could be found for logger "certbot.crypto_util"


PatrickHeppler:
I'm using this python script for a while to renew certs and update TLSA records in DNS:
https://gist.github.com/PatrickHeppler/064f6463d14fcf7d7a99f18f68a10684

But now certbot errors out: No handlers could be found for logger "certbot.crypto_util" which "triggers" my error handling. The certificates are signed, that's not the problem. But my script sees an error from letsencrypt and stops processing. I could disable the check, but that's a bad idea.

How to fix this? Missing package maybe?

bmw:
The issue is in our processing in our processing of the `--csr` flag. While processing the value, we attempt to log information but in this case, the logging hasn't been set up yet.

We can fix the problem on our end, but in the mean time you may want to make your error handling a little more lenient until this is fixed.

PatrickHeppler:
Okay, I have changed my error handling to check for this specific error and ignore it.
Could you please leave a comment here when it's fixed?

bmw:
Will do. We'll leave this issue open and close it once it's resolved.

pombredanne:
Using the latest from Pypi  https://pypi.python.org/pypi/certbot/0.8.1 I get `No handlers could be found for logger "certbot.plugins.disco"` ... Would this be related? I am not using certbot as a lib but from the command line. I am quite familiar with Python logging though, so I can chip in alright.

bmw:
It's almost certainly related. In general, the problem is that Certbot sometimes attempts to log information while processing and post-processing its arguments, however, the logging handlers cannot be setup until we've processed the arguments that affect logging settings.

There are at least three solutions to the problem:
1. Stop logging during argument processing/post-processing
2. Move setting up logging to be the first thing that happens during post-argument processing.
3. Have a simple argument parser that just parses the logging arguments. After this parser has run, we can set up logging and then do normal argument parsing.

I personally don't think 1 is a good idea. There's actually lot of work that goes on while processing/post-processing our arguments and losing logging info there would be unfortunate. An argument could be made that this work should be moved elsewhere which might be a good idea in the long term, however, I don't think this is necessary to fix such a simple problem.

[...]

PR #3261, opened 2016-07-08
Apache plugin PATH fallback

pde:
Work around broken PATH values in some cron jobs, by adding the two most common paths to `apachectl`: `/usr/sbin/` and `/usr/local/bin`.

Fixes: #1833 (aka #3244)

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -141,9 +141,13 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         return os.path.join(self.config.config_dir,
                             constants.MOD_SSL_CONF_DEST)
 
-    def _path_surgery(self):
-        """Mitigate https://github.com/certbot/certbot/issues/1833"""
-        dirs = ("/usr/sbin/", "/usr/local/bin/", "/usr/local/sbin/")
+    def _path_surgery(self, restart_cmd):
+        """Mitigate https://github.com/certbot/certbot/issues/1833
+
+        :returns: " expanded" if an expansion of the PATH occurred;
+                  "" otherwise
+        """
+        dirs = ("/usr/sbin", "/usr/local/bin", "/usr/local/sbin")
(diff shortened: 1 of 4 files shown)


joohoi:
After a quick readtrough few questions came to mind. This doesn't seem to address the problem "globally". We'd be unable to actually restart the services and to gather DEFINE - information for example. Maybe adding full paths to constants.py would be more robust solution, taken that distribution we're fixing it does have the same path throughout different versions.

bmw:
Another question I had is whether or not the logger is always set up when `prepare` is called. Trying to avoid another problem like the one in #3148.

pde:
@joohoi I thought it should address the problem globally if `prepare()` is called before other things, since the change to `os.environ["PATH"]` should persist?

pde:
@bmw I don't think we call `.prepare()` on plugins until [this point](https://github.com/certbot/certbot/blob/master/certbot/plugins/selection.py#L76)?  So `logger` calls shouldn't be a problem here....

pde:
@joohoi it appears to work, too:

```
root@isnot:~/certbot# export PATH=/tmp/
root@isnot:~/certbot# venv/bin/certbot --apache --dry-run certonly -d isnot.org -t
Starting new HTTPS connection (1): acme-staging.api.letsencrypt.org
tls-sni-01 challenge for isnot.org
Waiting for verification...
Cleaning up challenges
Generating key (2048 bits): /etc/letsencrypt/keys/0007_key-certbot.pem
Creating CSR: /etc/letsencrypt/csr/0007_csr-certbot.pem

IMPORTANT NOTES:
 - The dry run was successful.
```

bmw:
@pde, great. You're correct about the `logger` calls.

joohoi:
You are correct about the long term effects of the survery too. Looks fine.

linas:
thanks!

hlieberman:
FYI this is fixed in the Debian packages as of 0.8.1-2.

PR #3265, opened 2016-07-08
Errors when run without root

### LINKED ISSUE #2306: Error that results when run without root is unclear

bmw:
Fixes #2306, #2734 

All `OSError` and `IOError` exceptions that occur in the relevant sections will include a message about root. I did this due to the variety of errno types that occur across different systems, like we saw with the `webroot` directory.

An example error message looks like this:
```
The following error was encountered:
[Errno 13] Permission denied: '/etc/letsencrypt'
If running as non-root, set --config-dir, --logs-dir, and --work-dir to writeable paths.
```

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -36,9 +36,10 @@ from certbot.plugins import disco as plugins_disco
 from certbot.plugins import selection as plug_sel
 
 
-_PERM_ERR_FMT = ("An error occurred while trying to create or modify {0}. To "
-                 "run as non-root, set --config-dir, --logs-dir, and "
-                 "--work-dir to writeable paths.")
+_PERM_ERR_FMT = os.linesep.join((
+    "The following error was encountered:", "{0}",
+    "If running as non-root, set --config-dir, "
+    "--logs-dir, and --work-dir to writeable paths."))
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


schoen:
Will these also apply if the apache or nginx plugin is unable to modify the web server configuration? (Does it raise one of these exceptions in this case, and do they get far enough in the call stack to trigger these messages?)

bmw:
Unfortunately, no. None of the permissions errors that can occur in any of the plugins are affected.

pde:
LGTM.  Seth I made #3266 for the issue you identified.

pde:
While testing, I managed to find an `OSError` that you aren't catching:
```
pde@damoclid:~/certbot/certbot$ venv/bin/certbot --logs-dir /tmp/log certonly --standalone -t
Saving debug log to /tmp/log/letsencrypt.log
An unexpected error occurred:
OSError: [Errno 13] Permission denied: '/etc/letsencrypt/accounts/acme-v01.api.letsencrypt.org'
Please see the logfiles in /tmp/log for more details.
```

bmw:
@pde, yeah. I was specifically trying to fix the error messages reported in the two issues I claimed to be fixing. They are the most common error messages for people when they run the client without root.

In general, such errors could occur all over the place. Any time we access a file or grab a port, there's a chance we could have insufficient privileges. If we wanted to catch all `IOError` and `OSError` exceptions that could ever occur within the program, I could add another case to `_handle_exception` and add the message there.

I think the problem with doing this though is in many contexts, the message doesn't make sense. If you're trying to use the standalone, Nginx, or Apache plugins without root, you're going to have a bad time. Changing the default directories isn't going to fix that. If we want to do more than what's done in this PR, I think we have to look at each case individually and write an error message that describes that specific problem.

pde:
Fair point. Merging, then!

Issue #3179, opened 2016-06-15
Python 3 Support


bmw:
Just opening this issue for tracking purposes because it appears we don't have one. Ideally, Python 3 support should be added to `certbot` and its official plugins.

jacnoc:
This really needs to be high priority. I just looked it up and Python 3.4 is already TWO years old. Adding support for Python 3 will also encourage people to stop using older versions of Python (2.6/2.7)

Super-Baleine:
Yea, i agree. But it requires to adapt the code in version 3, so it requires to change many lines, it will take a long time, isn't it ?

bmw:
Yes, The `acme` module (a library for speaking the ACME protocol with Let's Encrypt) already works on Python 2.6, 2.7, and 3.3+. Adding support to the other parts of the project requires rewriting a significant portion of code.

We'll likely do it at some point (and if any of you would like to help with the effort, we'd certainly accept good PRs), however, support for Python 2.6 and 2.7 cannot be dropped due to the number of older systems using Certbot.

Super-Baleine:
oh yea, there are many old systems which don't have python3...

bmw:
See #4264 for a specific problem someone had trying to run Certbot with Python 3.

ynasser:
Here is a summary (mostly written for a Pycon-affiliated sprint, but applicable to all) if someone wants to get started working on this issue:

There's lots of investigating to be done for this issue. Almost all our unit tests currently pass with Python 3 (except `certbot-nginx`), but certbot doesn't actually work with Python 3. Ideas on where to start:


- run certbot with Python 3 with the staging environment, and try to fix it when it breaks
- run the integration tests with Python 3 and figure out how to fix them when they break
- look through the issues referenced above and try to figure out what’s wrong with them individually/how to fix them
- a fairly self-contained issue would be to try and make the nginx unit tests pass with Python 3, as they’re currently the only ones which don’t
- add Python 3 support for a specific subcommand (kind of like this: https://github.com/certbot/certbot/pull/3269) 

Here’s a guide on how to support versions both 2 and 3 simultaneously: https://docs.python.org/3/howto/pyporting.html (pay particular attention to the “text vs binary data” section, as that’s where most projects run into problems.)

[...]

PR #3269, opened 2016-07-10
Python 3 support for certonly

yan12125:
Ref: #3179

It's far from complete, just a quick try. Only these command is tested:
```
$ certbot --help
$ certbot --help all
$ certbot --logs-dir logs --config-dir config --work-dir certbot-workdir certonly --manual --domains my.domain.name --email my.email@some.company.com --agree-tos --manual-public-ip-logging-ok
```

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -343,8 +343,10 @@ class HelpfulArgumentParser(object):
         self.determine_verb()
         help1 = self.prescan_for_flag("-h", self.help_topics)
         help2 = self.prescan_for_flag("--help", self.help_topics)
-        assert max(True, "a") == "a", "Gravity changed direction"
-        self.help_arg = max(help1, help2)
+        if isinstance(help1, bool) and isinstance(help2, bool):
+            self.help_arg = help1 or help2
+        else:
+            self.help_arg = help1 if isinstance(help1, str) else help2
         if self.help_arg is True:
             # just --help with no topic; avoid argparse altogether
             print(usage)
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,6 @@ install_requires = [
     'psutil>=2.2.1',  # 2.1.0 for net_connections and 2.2.1 resolves #1080
     'PyOpenSSL',
     'pyrfc3339',
-    'python2-pythondialog>=3.2.2rc1',  # Debian squeeze support, cf. #280
     'pytz',
     # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
     # will tolerate; see #2599:
@@ -52,6 +51,12 @@ install_requires = [
     'zope.component',
     'zope.interface',
 ]
+
+# Debian squeeze support, cf. #280
+if sys.version_info[0] == 2:
+    install_requires.append('python2-pythondialog>=3.2.2rc1')
+else:
+    install_requires.append('pythondialog>=3.2.2rc1')
 
 # env markers in extras_require cause problems with older pip: #517
 # Keep in sync with conditional_requirements.py.
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -1,8 +1,8 @@
 """ACME AuthHandler."""
-import itertools
 import logging
 import time
 
+import six
 import zope.component
 
 from acme import challenges
@@ -141,7 +141,7 @@ class AuthHandler(object):
 
         """
         active_achalls = []
-        for achall, resp in itertools.izip(achalls, resps):
+        for achall, resp in six.moves.zip(achalls, resps):
             # This line needs to be outside of the if block below to
             # ensure failed challenges are cleaned up correctly
             active_achalls.append(achall)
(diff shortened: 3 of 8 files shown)


cowlicks:
The pylint failure is from a problem in master.

bmw:
Travis failures were unrelated to this PR so I kicked the build to make them pass. Sorry for the trouble!

yan12125:
Thanks. I know the magic number 1250 :)

Here's a question: I see some Python 3.x tests in Travis CI. Apparently those tests don't cover such command line usages. Should I add some tests for my changes?

pde:
@yan12125 the tests that currently run for Py 3.x are unit tests; what we don't run are any of the [integration tests](https://github.com/certbot/certbot/blob/master/tests/boulder-integration.sh). It may be that you want to tweak those and enable them for one of the Py 3.x travis matrix entries.

pde:
(You may also want to add more unit tests; I'm actually a little surprised that there were so many Py 3 errors that were not exposed by the current unit test suite).

bmw:
@pde, please let me know if you disagree, but I don't think adding new tests is necessary (assuming the new code added is covered by tests). I suppose we could write a new test suite that tests a subset of `certbot` using Python 3, but I think what we really want in the long run is to build on PRs like these and use our current test suite with both Python 2 and 3 (like we do with `acme`).

pde:
@bmw no I think we agree. I was just mystified by the fact that there were aspects of the code that were passing Py3 unit tests but blowing up in fundamental (py3-incompatible) ways when invoked from the command line; that seemed to imply gaps in our unit test coverage. But I haven't looked closely.

setup.py
```diff
@@ -64,6 +63,12 @@ def read_file(filename, encoding='utf8'):
 else:
     install_requires.append('mock')
 
+# Debian squeeze support, cf. #280
+if sys.version_info[0] == 2:
```
bmw: Just noting this now for when this is given a full review, but my understanding is changes like this will require us to build multiple wheels of `certbot` when pushing the packages to PyPI. We'll need a Python 2 wheel and a Python 3 wheel.

I just encountered a similar problem [here](https://github.com/certbot/certbot/pull/2061#discussion_r71929407) but luckily there was a workaround. I'm not aware of a single `dialog` package we can use here so this may be inevitable (unless we want to write our own `dialog` compatibility library or drop support for `dialog`, although at that point making and testing two wheels is probably easier) but I wanted to make sure we kept this in mind before merging.

yan12125:
I'd like to add an functional test. Is there a test server that I can upload ```.well-known/acme-challenge/*``` files?

bmw:
@yan12125, the way we run tests like that is by running a local copy of Let's Encrypt server software where the DNS for certain names resolves to `localhost`. You can see examples of this in our integration tests. The code for setting up these tests is already written and can be reused.

It might make sense to do that in another PR building off this one so the features can be reviewed individually.

certbot/cli.py
```diff
@@ -342,8 +342,13 @@ def __init__(self, args, plugins, detect_defaults=False):
         self.determine_verb()
         help1 = self.prescan_for_flag("-h", self.help_topics)
         help2 = self.prescan_for_flag("--help", self.help_topics)
-        assert max(True, "a") == "a", "Gravity changed direction"
-        self.help_arg = max(help1, help2)
```
pde: I'm a little sad that this hack stopped working in Python 3 :)

certbot/cli.py
```diff
@@ -342,8 +342,13 @@ def __init__(self, args, plugins, detect_defaults=False):
...
+        if isinstance(help1, bool) and isinstance(help2, bool):
+            self.help_arg = help1 or help2
+        else:
+            if isinstance(help1, str):
```
pde: It's probably nicer to do:
```
self.help_arg = help1 if isinstance(help1, str) else help2
```

pde:
Aside from that one suggestion, this looks okay to me. @bmw should we try to deal with the wheels and release engineering implications in this PR, or make a separate issue for actually shipping Python 3 support?

pde:
So my instinct is to just land this one PR as incremental progress (though @yan12125 if you can add integration tests that would be awesome and will prevent regressions from your work), and once someone has dealt with testing and fixing the other subcommands and plugins in Python 3, we can consider the question of how to release correctly for Python 3.

yan12125:
I was trying to add integration tests. However, existing tests fail and I can't run integration tests locally (got the same error as letsencrypt/boulder#1913). So I just skip adding new tests.

yan12125:
I guess I need to change the title back :)

pde:
@yan12125 do you want to come and work with us in the #letsencrypt-dev Freenode IRC channel, or post in that boulder issue, to get help running integration tests locally?

You can also probably get Travis to run the tests for you, though that may be best done in a different pull request that doesn't interfere with this one.

pde:
LGTM

yan12125:
I guess my first step is trying to make unit tests pass on Python 3. Integration tests sound more like the final step. Anyway thanks you all for the kind help!

yan12125:
Thanks!

PR #3272, opened 2016-07-11
Try to fix travis-ci lint failure

cowlicks:


bmw:
LGTM. Thanks for being so proactive on this @cowlicks.

I created #3273 to track potentially cleaning up the tests so the `pylint: disable` isn't necessary and #3274 to track investigating how `master` became broken in the first place.

cowlicks:
The failure is due to `certbot-apache/certbot_apache/tests/configurator_test.py` exceeding the number of lines allowed by us in our `.pylintrc` file (1250). I'm not sure how this change got into master without being tested.

I disabled the `too-many-lines` check for this module. This is arguably the wrong solution. The module should be less than 1250 lines since that is the limit we set in our pylintrc. but this is the easiest thing to to do because.... And I'm going to stop writing here since @bmw already merged this. Cheers!

Issue #3227, opened 2016-06-29
add nginx round-trip tests to tox/travis


pde:
@schoen has some tests to ensure that the nginxparser `load`/`dump` operations do not modify a wide range of nginx conf files.  Let's get them into our test suite!

pde:
These tests are now in the tree, but not yet in tox / Travis (they're not passing yet).

bmw:
Fixed in #3393.

PR #3275, opened 2016-07-11
Initial version of nginx parser roundtrip test

schoen:
This can be used to check whether the parser succeeds in parsing a variety of nginx config files and whether the reserialized versions then match the original contents.


pde:
Fixes: #3227

pde:
LGTM

PR #3280, opened 2016-07-12
Use simple socket test for port availability if psutil not found

### LINKED ISSUE #3106: Replace psutils dependency
It's compiled from C.  It's causing problems (#1080 / #3101).  If there's something equally effective that we can do in pure Python, let's do it.

joohoi:
Use stdlib socket to check port availability if psutil dependency is not found. Display slightly different notification to user (because we don't have all the information available).

Remove psutil dependency.

Fixes #3106


certbot/plugins/util.py
```diff
@@ -24,6 +35,50 @@ def already_listening(port, renewer=False):
     :returns: True or False.
 
     """
+
+    if USE_PSUTIL:
+        return already_listening_psutil(port, renewer=renewer)
```
cowlicks: Since `renewer` is just getting passed through, maybe `already_listening` should just take `**kwargs` and pass them to `already_listening_*`. That way if we want to change the default value of renewer, we just change `already_listening_socket/psutil` and don't need to change it in `already_listening` too (where we might forget).

If this change is made the `renewer` kwarg should be added to the docstring.

cowlicks: Err on second thought, we want `already_listening_socket/psutil` to always behave the same, so changing their behavior should happen in one place: `already_listening`. So nvm lol.

pde:
Progress towards: #1301

cowlicks:
Is there any behavior non-psutil function doesn't duplicate? If not, is there any reason to keep the psutil function around?

joohoi:
Yeah, We don't get the process name and pid out of it (easily). Plan was to make it optional.

pde:
@joohoi this looks stuck waiting for you to merge master into it. Also, you should now have permissions to push to branches on the `origin` repo, which would allow others to make those kinds of fixing commits to your branches.

joohoi:
Left outside of the scope of this PR: remove the actual dependencies in setup.py etc.

certbot/plugins/util.py
```diff
@@ -3,15 +3,27 @@
...
+    USE_PSUTIL = True
+except ImportError:
+    USE_PSUTIL = False
 
 logger = logging.getLogger(__name__)
 
+RENEWER_EXTRA_MSG = (
+    " For automated renewal, you may want to use a script that stops"
+    " and starts your webserver. You can find an example at"
+    " https://letsencrypt.org/howitworks/#writing-your-own-renewal-script"
```
pde: This URL should probably be `https://certbot.eff.org/docs/using.html#renewal`.

pde:
We can probably remove the `psutil` dependency in this PR too.

pde:
Otherwise LGTM.

pde:
Great, thanks!

PR #3291, opened 2016-07-14
Update various package names in using.rst from "letsencrypt" to "certbot" [needs revision]

pde:


docs/using.rst
```diff
@@ -460,13 +460,19 @@ repo, if you have not already done so. Then run:
...
 .. code-block:: shell
 
-    sudo dnf install letsencrypt
+    sudo dnf install certbot
+
+if you want to use the Apache plugin:
+
+.. code-block:: shell
+   
+   sudo dnf install python-certbot-apache
```
bmw: This appears to be called `python2-certbot-apache`. I created certbot/website#155 to fix the issue on the website.

docs/using.rst
```diff
@@ -491,7 +497,7 @@ the Nginx plugin is ready.
    emerge -av app-portage/layman
    layman -S
    layman -a mrueg
-   emerge -av app-crypt/letsencrypt-nginx
+   emerge -av app-crypt/certbot-nginx
```
bmw: I think everything about using layman here should be deleted. We have very little information about it and I'm not sure if a `certbot-nginx` package exists. We removed all info about it from the website before it launched.

bmw:
In addition to my few comments above, I'd personally like to remove all OS specific installation instructions from our docs in the future (see #3237). We can still move forward with this PR if you'd like though.

docs/using.rst
bmw: This may not be a problem. I commented on the website issue.

bmw:
Closing in favor of #3747.

Issue #1741, opened 2015-12-04
FAQ: Do I need root (superuser) to use letsencrypt?


hickford:
My website is hosted on a shared server. I don't have root on it. Can I use letsencrypt? 

I tried installing letsencrypt-auto following https://letsencrypt.readthedocs.org/en/latest/using.html#installation 

    ham@pandora:~/letsencrypt$ ./letsencrypt-auto
    Bootstrapping dependencies for Debian-based OSes...
    [sudo] password for ham:
    ham is not in the sudoers file.  This incident will be reported.

The user guide is confusing:

> Please do not attempt the installation commands as superuser/root

spede:
Try:
https://github.com/kuba/simp_le
Or:
https://github.com/diafygi/letsencrypt-nosudo

The confusing bit means you shouldn't run letsencrypt-auto as superuser, because the program will invoke sudo when it needs to automatically.

mikehaertl:
I also ran into some problems as I started installation as root (`sudo -s`). I think the above notice should also be added here: https://letsencrypt.readthedocs.org/en/latest/using.html#letsencrypt-auto

kolAflash:
In many scenarios it's the most useful to run letsencrypt-auto with "--webroot-path" (on a webroot with proper permissions) or "--manual". If the user is also sure, that everything needed is installed via apt-get, zypper, yum, ... there should be absolutely no reason to require root!

So please, ask only for root if it's needed!

For installing packages, either check first if the needed packages are already installed. E.g.:
rpm -ql | grep PACKAGE_YOU_LOOK_FOR
dpkg -l | grep PACKAGE_YOU_LOOK_FOR
Or provide an option "--dont-install-packages" which just skips this step.

Always using root is a bad idea, because:
- Not everyone has root on a server
- Bad security
Additionally, probably many users don't want software-packages being installed without being asked before.

tobiasmcnulty:
Agreed, it would be nice to be able to automate certificate renewal even if root is not available (e.g., if the cert is going to be uploaded via an API).

pde:
Whether root is required to run Certbot or not depends on how you intend to use it.  In some cases, typically when using the `webroot` and `manual` plugins, Certbot can be run without root if you ensure that `/etc/letsencrypt/`, `/var/log/letsencrypt`, and `/var/lib/letsencrypt` are writeable by the user that wants to run it (you can also change these locations with `--config-dir`, `--log-dir` and `--work-dir`).

[...]

PR #3294, opened 2016-07-15
New docs structure and introduction

bmw:
In addition to making sure the new structure/text is correct, we need to make sure this won't break packaging. `README.rst` is included in the `certbot` package and is shown on [PyPI](https://pypi.python.org/pypi/certbot/0.8.1). Will including other files break this? I doubt it since `certbot` also includes the entire `docs` directory, but we should make sure this won't cause problems before this is merged.


bmw:
Welp. Just saw github/markup#172. We'll have to make some changes here.

bmw:
Closing and reopening to start tests

bmw:
Well, Travis is being weird so I'll keep playing with that, but I think this is ready for review.

This incorporates @ptychomancer's work as well as the structural doc changes we wanted. The way I accomplished the doc structure with respect to the `README` is a little hacky, but it's the best solution I could come up with.

Including other files in the `README` causes problems like the one I linked above so I moved the text we wanted back into the file (except `CHANGES.rst` but I'll mention that at the end). To avoid duplication, I include parts of the `README` in other files. You can do this either based on line numbers or by searching for certain text. To try and make this robust as possible, I added comments in the `README` which specify tags that begin and end certain sections. Using these tags, we can include the section we want in other files without relying on either the line number or the visible wording of a specific section which may change in the future.

As I mentioned, I didn't do anything with `CHANGES.rst`. We still want this file and we don't want to duplicate it in the `README` so I left it as is. If we want `CHANGES.rst` mentioned in the `README`, perhaps we can link to it, but I'm not sure how well this will work on PyPI, GitHub, our packages, etc. `CHANGES.rst` is still linked to in the User Guide like we discussed last week so it's still referenced for our users.

README.rst
```diff
@@ -1,169 +1,19 @@
...
 
-Certbot (previously, the Let's Encrypt client) is **BETA SOFTWARE**. It
-contains plenty of bugs and rough edges, and should be tested thoroughly in
-staging environments before use on production systems.
+Anyone who has gone through the trouble of setting up a secure website knows what a hassle getting and maintaining a certificate is. Certbot and Let’s Encrypt can automate away the pain and let you turn on and manage HTTPS with simple commands. Using Certbot and Let's Encrypt is free, so there’s no need to arrange payment.
 
-For more information regarding the status of the project, please see
-https://letsencrypt.org. Be sure to checkout the
-`Frequently Asked Questions (FAQ) <https://community.letsencrypt.org/t/frequently-asked-questions-faq/26#topic-title>`_.
+How you use Certbot depends on the configuration of your web server. The best way to get started is to use our `interactive guide <https://certbot.eff.org>`_. It generates instructions based on your configuration settings. You’ll need root or administrator access to your web server to run Certbot.
```
pde: Perhaps "In most cases, you'll need root or administrator access"... with a link to an appropriate FAQ? We sort of have the answers over in #1741.

pde:
One thing that we should address is the use case where a developer clones the git repo and opens README.rst to understand how to use the repo they just cloned.  Currently there are no pointers or references in the README to the relevant parts of `contributing.rst`.  I guess there are multiple ways of addressing that (moving the "running a local copy" section into the README but then disabling its display by default? Just including a one-sentence reference to it? other ways I haven't thought of?).

pde:
Other than those things, LGTM.

bmw:
This should be ready for another review.

PR #3301, opened 2016-07-19
Deprecation fixes

scop:


pde:
I'm a little mystified by the test failures here; superficially it looks like one of the places where logging is mocked in the tests might need to be updated, but that isn't consistent with the `py27-oldest` tox test having passed...

bmw:
It's worse than that. `cover` passed but `py27` didn't. `cover` doesn't pass for me locally with this branch.

What's going on Travis...?

bmw:
So I'm not sure what's going on here (I suspect some strangeness with `mock`), however, @scop, in [one of our tests](https://github.com/certbot/certbot/blob/master/certbot/tests/reverter_test.py#L359), we mock out `logger.warn` and changing the code to `logger.warning` is causing failures. If you fix this test to use `logger.warning` as well, I'm happy to review this PR.

I'm adding "[needs revision]" to the title of this PR. Please remove it when you'd like another review.

scop:
The issue was that changing `_checkpoint_timestamp` to use `warning` instead of `warn` resulted in another "new" `warning` call from `_setup_three_checkpoints`, and that affected the expected call count in `test_rollback_too_many`. Changed so that the call count is zeroed only after setting up the test checkpoints so that the test tests only warnings from `rollback_checkpoints`.

bmw:
Right, but the `py2*` and `cover` tests all run `test_rollback_too_many` yet only some of them failed. Not sure what's going on there, but I'm not too worried about it.

LGTM. Thanks for the PR @scop.

scop:
Maybe the tests that didn't fail don't provoke the warning from `_setup_three_checkpoints` for some reason.

PR #3314, opened 2016-07-21
Change fatal warning to a fatal message

### LINKED ISSUE #3311: Fatal warnings
As reported by @Mechazawa [here](https://github.com/certbot/certbot/commit/d2f1ae9c3870289b52fb1c0ab5130c8fae52aa72#commitcomment-18318527), if `*-auto` needs the `--debug` flag on your system, it will display a warning and then exits abnormally.

Since it causes the program to exit, it's not really a warning so we should not report it as such.

Mechazawa:
Fixes #3311


bmw:
LGTM. Thanks for the PR @Mechazawa.

PR #3335, opened 2016-07-27
Gradually remove psutil dependency, bugfix [URGENT]

### LINKED ISSUE #3334: psutil fails to install because hash is missing when running certbot-auto
certbot-auto pulls packages from pypi, and the latest version still has psutil dependency from setup.py, but in the git master version has the hashes already removed. Dependency needs to be restored, and removed gradually.

Community forum thread: https://community.letsencrypt.org/t/had-a-problem-while-installing-python-packages/18333

joohoi:
Return psutil hashes until we get the next release out and pypi package has the psutil dependency removed.

Fixes #3334


bmw:
Thanks for making this PR so quickly @joohoi. Merging as soon as Travis finishes.

PR #3342, opened 2016-07-27
Adding modification check against the current /letsencrypt-auto

### LINKED ISSUE #3337: prevent modifications to certbot-auto and letsencrypt-auto
`certbot-auto` and `letsencrypt-auto` in the root of our repo should only be modified during our releases, however, this is not enforced by any of our tests. It should be enforced by something running in Travis.

Probably the easiest way to do that is to use `letsencrypt-auto-source/pieces/fetch.py` to download the latest stable copy of the script and diff that against `certbot-auto` and `letsencrypt-auto` in the root. If they differ, the test should fail.

TheNavigat:
Closes #3337

### DIFF
--- a/tests/modification-check.sh
+++ b/tests/modification-check.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+temp_dir=`mktemp -d`
+
+# Script should be run from Certbot's root directory
+cp letsencrypt-auto ${temp_dir}/to-be-checked
+cp letsencrypt-auto-source/pieces/fetch.py ${temp_dir}/fetch.py
+cd ${temp_dir}
+
+LATEST_VERSION=`python fetch.py --latest-version`
+python fetch.py --le-auto-script v${LATEST_VERSION}
+
+cmp -s letsencrypt-auto to-be-checked
+
+if [ $? != 0 ]; then
+	echo "Root letsencrypt-auto has changed."
+	rm -rf temp_dir
+	exit 1
+fi
+
+rm -rf temp_dir
--- a/tests/modification-check.sh
+++ b/tests/modification-check.sh
@@ -3,19 +3,33 @@
 temp_dir=`mktemp -d`
 
 # Script should be run from Certbot's root directory
-cp letsencrypt-auto ${temp_dir}/to-be-checked
+cp letsencrypt-auto ${temp_dir}/letsencrypt-to-be-checked
+cp certbot-auto ${temp_dir}/certbot-to-be-checked
+
 cp letsencrypt-auto-source/pieces/fetch.py ${temp_dir}/fetch.py
 cd ${temp_dir}
 
 LATEST_VERSION=`python fetch.py --latest-version`
 python fetch.py --le-auto-script v${LATEST_VERSION}
 
-cmp -s letsencrypt-auto to-be-checked
+cmp -s letsencrypt-auto letsencrypt-to-be-checked
 
 if [ $? != 0 ]; then
 	echo "Root letsencrypt-auto has changed."
 	rm -rf temp_dir
 	exit 1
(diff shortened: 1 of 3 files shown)


TheNavigat:
```
19:09 < TheNavigat> bmw: Regarding your note, that will fail because fetch.py grabs the le-auto inside the le-auto-source dir and not the one in the root
19:10 < TheNavigat> I think creating a new script that kind of mimics what fetch.py does and diffs the end result isn't a bad idea
19:41 <+bmw> TheNavigat: but it gets the script from the specified release tag
19:42 <+bmw> so it should work fine
19:42 <+bmw> and certbot-auto should be checked too
19:52 < TheNavigat> bmw: We'll have to update the script every time we release, then
19:52 < TheNavigat> If we're going to use fetch.py
19:54 <+bmw> TheNavigat: nope
19:54 <+bmw> look at how *-auto uses fetch.py
19:54 <+bmw> we want to do that
19:54 <+bmw> (minus comparing the remote version to our own version number)
```

@bmw: I rethought those ideas and I have a comment. Here's a basic summary of my idea.

Compare this PRs version to master's. If they're different, fail.

This should work because unless the commit is made directly on master, the files shouldn't be different. However, ```le-auto```'s technique will fail when someone releases a new version that isn't release to PyPi yet. ```le-auto``` specifies the latest version to be able to retrieve the file, and the release commit will definitely have a different file from the one le-auto will grab using the ```PyPi``` version. See https://github.com/certbot/certbot/blob/master/letsencrypt-auto#L1094-L1119

bmw:
@TheNavigat, I don't think that would work. We never make commits directly on `master`. In fact, we've disabled pushing to `master` on GitHub through [protected branches](https://help.github.com/articles/about-protected-branches/). All changes must go through a PR that has passed Travis tests.

Our general release process is as follows:
1. Push a [tag](https://github.com/certbot/certbot/tags) containing the version bump and the new `certbot-auto` and `letsencrypt-auto`s.
2. Upload Certbot packages to PyPI.
3. Make a PR to merge those changes into `master`.

With the system you're proposing, I believe it would be impossible to ever change `*-auto` unless we removed our protections on `master` which we're not going to do. If someone makes a PR trying to change `*-auto` without having already created a tag and released PyPI packages, I think it is correct behavior that the tests fail. The released versions of the scripts have not changed, so changing them in `master` should not be allowed.

TheNavigat:
@bmw, I apologize, I wasn't aware of the release plan and the fact that PyPI packages are released before merging the new tag to master, and not the other way around.

Please review the code again. Thanks! :)

tests/modification-check.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+temp_dir=`mktemp -d`
+
+# Script should be run from Certbot's root directory
+cp letsencrypt-auto ${temp_dir}/to-be-checked
```
bmw: You should also check `certbot-auto` in the root.

tests/modification-check.sh
```diff
@@ -0,0 +1,21 @@
...
+
+cmp -s letsencrypt-auto to-be-checked
+
+if [ $? != 0 ]; then
+	echo "Root letsencrypt-auto has changed."
+	rm -rf temp_dir
+	exit 1
+fi
+
+rm -rf temp_dir
```
bmw: While not strictly necessary, I'd like to see an `echo` stating that the root `{certbot,letsencrypt}-auto` files are unchanged. I'm happy checking the return value locally, but I'd like to see something in the Travis log.

bmw:
Other than my tiny suggested revisions, this LGTM.

bmw:
LGTM. Thanks @TheNavigat.

Issue #3333, opened 2016-07-26
Failure to start Nginx after configuring redirect


bmw:
This seems to happen to any Nginx config running Certbot off `master`.

Error is:
```nginx: [emerg] unknown directive "if($scheme"```

This is a regression. Certbot 0.8.1 did not have this problem.

PR #3345, opened 2016-07-27
Fix spacing of nginx redirect blocks

pde:
Closes: #3333

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -401,8 +401,8 @@ class NginxConfigurator(common.Plugin):
         :type unused_options: Not Available
         """
         redirect_block = [[
-            ['if', '($scheme != "https")'],
-            [['return', '301 https://$host$request_uri']]
+            ['\n', 'if', ' ', '($scheme != "https")'],
+            [['\n    ', 'return', ' ', '301 https://$host$request_uri']]
         ]]
         self.parser.add_server_directives(
             vhost.filep, vhost.names, redirect_block, replace=False)
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -160,9 +160,9 @@ class NginxConfigurator(common.Plugin):
         stapling_directives = []
         if self.version >= (1, 3, 7):
             stapling_directives = [
-                ['\n', 'ssl_trusted_certificate', ' ', chain_path],
-                ['\n', 'ssl_stapling', ' ', 'on'],
-                ['\n', 'ssl_stapling_verify', ' ', 'on'], ['\n']]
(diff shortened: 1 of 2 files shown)


certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -160,9 +160,9 @@ def deploy_cert(self, domain, cert_path, key_path,
         stapling_directives = []
         if self.version >= (1, 3, 7):
             stapling_directives = [
-                ['\n', 'ssl_trusted_certificate', ' ', chain_path],
-                ['\n', 'ssl_stapling', ' ', 'on'],
-                ['\n', 'ssl_stapling_verify', ' ', 'on'], ['\n']]
+                ['\n    ', 'ssl_trusted_certificate', ' ', chain_path],
+                ['\n    ', 'ssl_stapling', ' ', 'on'],
+                ['\n    ', 'ssl_stapling_verify', ' ', 'on'], ['\n']]
```
bmw: Question: Can you explain to me the reason for the change here? I'm assuming it's to indent these directives. If so, a couple follow up questions:

1. Will this work if the block is has more or less indentation than what is seen here?
2. Can the parser not handle this for us when adding new directives?

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -401,9 +401,10 @@ def _enable_redirect(self, vhost, unused_options):
         :type unused_options: Not Available
         """
         redirect_block = [[
-            ['if', '($scheme != "https")'],
-            [['return', '301 https://$host$request_uri']]
-        ]]
+            ['\n    ', 'if', ' ', '($scheme != "https") '],
+            [['\n        ', 'return', ' ', '301 https://$host$request_uri'],
+             '\n    ']
+        ], ['\n']]
```
bmw: Question: Is there a difference as far as our parser is concerned between `['\n']` and `\n` at the end of a block?

bmw:
I have a couple of questions, mainly about manually including indentation here which seems cumbersome and error prone, but other than that, this LGTM.

certbot-nginx/certbot_nginx/configurator.py
pde: This is the arbitrary-but-common case. Sometimes the indentation will be wrong. Adding indentation matching to `_add_directive` is future work; I have a branch that starts to do that sort of stuff (and makes all of this whitespace in new directives obsolete) but it's going to be a lot of work, and not land in 0.9.0.

certbot-nginx/certbot_nginx/configurator.py
pde: Probably not!

pde: But rather than messing with it, I'm inclined to move on for now and come back later to automate all of this indentation :)

PR #3353, opened 2016-07-29
Clarification to confusing success message on manual re-installation [minor revision requested]

chrismarget:
Detail in commit comments.

### DIFF
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -218,7 +218,7 @@ def success_installation(domains, action="install"):
             _gen_https_names(domains),
             os.linesep,
             os.linesep.join(_gen_ssl_lab_urls(domains)),
-            action,)
+            action),
         height=(10 + len(domains)),
         pause=False)
 
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -203,7 +203,7 @@ def _choose_names_manually():
     return []
 
 
-def success_installation(domains):
+def success_installation(domains, action="installed"):
     """Display a box confirming the installation of HTTPS.
 
     .. todo:: This should be centered on the screen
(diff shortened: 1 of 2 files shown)


certbot/display/ops.py
```diff
@@ -212,11 +212,13 @@ def success_installation(domains):
 
     """
     z_util(interfaces.IDisplay).notification(
-        "Congratulations! You have successfully enabled {0}{1}{1}"
+        "Congratulations! You have successfully {3}ed a certificate for{1}"
```
pde: How about "enabled HTTPS" or "installed a certificate and enabled HTTPS"?

Since turning on HTTPS may be the conceptually important point for many users.

chrismarget:
@pde - The original message text "you have enabled <URL>" is a little bit confusing to me. The message is displayed regardless of the plugin in use. So, in an authenticator-only scenario, what's been "enabled" exactly?

My perspective: The "action" being passed here (renew/reinstall/newcert) better represents what's actually happened, but even it is wrong: "reinstall" is a big assumption, because the cert may just be sitting on disk, unused.

I went with the default action of "install" on line 206 because I was following that trend, but it feels like "receive" better describes what's going on here.

With that in mind, I'm not sure that "installed a certificate and enabled HTTPS" is something that can be reasonably claimed with the information available in this function.

At any rate, my intention here was to make the message somewhat more sensible, while sticking to the confines of the existing structure.

certbot/display/ops.py
bmw: You also should probably remove the newline here after "certificate for". The notification method being called here will handle creating newlines when necessary.

chrismarget:
The issue that prompted me to poke around at this message is what happens on reinstallation of an existing certificate:

`        Your existing certificate has been successfully reinstalled, and the new
        certificate has been installed.
`

There is no `new certificate` here, so what's it referring to exactly?

The overall structure of the message changed in my commit only because of the inputs to the function which creates the message. I have no opinion on "installed/re-installed" vs "enabled/re-enabled". I was only trying to fix the reference to the non-existent "new certificate" when running the installer only.

bmw:
> There is no new certificate here, so what's it referring to exactly?

Yeah I saw this in your commit message. It's a bug, fixed by this PR, and I appreciate you noticing it and taking the time to write up a fix. Comments made by pde and I are just ways we feel that the fix can be slightly improved.

If you not available or interested in continuing to work on this, I'm happy to build off this PR to get it merged. Let me know what you'd like to do.

bmw:
Thanks for your PR! Mine is at #3784.

PR #3354, opened 2016-07-30
Fix README typo

bmw:


bmw:
Self merging because I'm a rebel (and because it's a one character change to our documentation fixing an obvious typo).

Issue #1466, opened 2015-11-11
Standalone mode does not bind to ipv6


rewbycraft:
As far as I can tell, the letsencrypt client does not bind to ipv6 in standalone mode nor do the commands given in manual mode. While this may not seem too significant, but for people with ipv6 only machines (like me) this is quite a big problem. My suggestion would be to somehow bind to both ipv4 and ipv6 at the same time.

Either way, what you're doing is a great service to the global internet community, keep on rocking!

pde:
At the moment the server won't connect over IPv6 to perform validation, either.  So this feature request is blocked (or at least not useful) until [this boulder ticket](https://github.com/letsencrypt/boulder/issues/593) is resolved.

pde:
Subticket of #180

rewbycraft:
While I do see where you are coming from, I have to disagree with you. Having the client listen on ipv6 is relevant *right now* because of NAT64.

What is NAT64, you may ask?
Basically, the network admin can set aside a block of ipv6 (the RFC range is 64:ff9b::/96) which will be 1:1 mapped to the ipv4 internet (basically, if you try to connect to the nat64 encoded version of an ipv4 it will automatically route it through to the public ipv4 (and a reverse path exists in my case)).
I am in fact in posession of an ipv4 address, it is however not directly attached to the server in question (it's behind the NAT64). Any connections to my ipv4 arrive as ipv6 on my machine. I've been able to work around the bug by using the manual option and doing some shenanigans to the given python code to make it listen on ipv6. (I can't use my web server directly due to the machine in question functioning primarily as a load balancer, as such it does not have the capability to serve a directory using an ipv6 enabled server like nginx or apache.)

At the very least, I do appreciate that you guys may look at this eventually.

kuba:
Have you tried running the client under your setup? I don't think there is anything in the code that stops us from "binding" to IPv6. If there is, then please paste the traceback.

rewbycraft:
There is no error, it's just not binding/listening on ipv6. Basically, it seems to be listening on 0.0.0.0 but not :: (aka, it's listening on all ipv4 ips, but not all ipv6 ips) (I gathered this information with netstat). Now, from taking a cursory glance at the code I can spot why it does this.
[...]

PR #3369, opened 2016-08-04
Bind to IPv6, fix the problem of ipv6 site cannot generate / renew certificate [revision requested]

palxex:
Follow the guide in https://github.com/certbot/certbot/issues/1466.

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -5,6 +5,7 @@ import functools
 import logging
 import os
 import sys
+import socket
 
 from six.moves import BaseHTTPServer  # pylint: disable=import-error
 from six.moves import http_client  # pylint: disable=import-error
@@ -51,6 +52,7 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
 
 class TLSSNI01Server(TLSServer, ACMEServerMixin):
     """TLSSNI01 Server."""
+    address_family = socket.AF_INET6
 
     def __init__(self, server_address, certs):
         TLSServer.__init__(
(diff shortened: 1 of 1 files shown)


pde:
This is failing unit tests, so will obviously need some more work...

nomaster:
The patch works for me. Thank you!

bmw:
This patch will work on systems with IPv6 support, but will fail on those without. Since this solution isn't quite enough and it's failing tests, I'm closing this PR in favor of the more complete #3687.

jsha:
@bmw are you sure this will fail on systems without IPv6? I just tested this patch on my IPv4-only system and found that it successfully bound to a port that I could then connect to:

```
~/certbot/acme/examples/standalone $ python ~/certbot/acme/acme/standalone.py
INFO:__main__:Serving at https://:::33510...
```

Or is the issue systems that don't understand the IPv6 address family at all (regardless of whether they have an address)?

bmw:
Interesting! It's not clear to me why that works if the underlying OS doesn't have IPv6 support. Other problems with an approach like this though are:

1. Python can be compiled without IPv6 support.
2. Some systems such as the BSDs don't map IPv4 traffic to IPv6. A patch like this would make standalone unusable for people on these systems relying on IPv4.

jsha:
To be clear, I haven't tested it on an OS without IPv6 support; I tested it on a host with IPv6 support but no IPv6 address.

cmcqueen:
You would need to test on a kernel compiled without IPv6 support. Perhaps you would need to compile your own kernel for testing. For example, I've got an embedded Linux BeagleBone Black based system with IPv6 disabled in the kernel.

cmcqueen:
A more cross-platform solution would be to open separate IPv4 and IPv6 listeners, and to abort only if both of them fail to open.

higuita:
Just to help people until this is really fixed:

If you have ipv6 and certbot is failing , use this patch (https://github.com/certbot/certbot/pull/3369.diff) and it will work.

This is not the long term solution, but fix the problem until #3687 finished and applied

Issue #3343, opened 2016-07-27
Rationalise challenge and port selection flags


pde:
We need to choose challenge types for:
* manual
* the (currently nonexistent) script plugin
* standalone

We need to choose ports for:
* standalone
* testing deployments
* things that are running behind proxies

The flags for doing all of these things should probably be cleaned up and rationalised.

pde:
Needed for: #2782

pde:
Needed for turning #2061 into an end user feature.

bmw:
To add more info here, we want a flag (or flags) in the core of the client (not in any of the plugins) that can be used for challenge selection. Flags like `--standalone-supported-challenges` and `--manual-supported-challenges` should be removed.

Probably the best way for this to work is to have a flag like `--preferred-challenges` or `--challenge-preferences` that can be used to provide a delimited list of the preferred challenges you want Certbot to perform, listed in the order they are preferred. This is like the `--standalone-support-challenges` flag.

Normally, Certbot determines that challenges that will be performed with the `gen_challenge_path` function in `certbot/auth_handler.py`. `get_chall_pref` is called on the selected plugin and is passed as the `preferences` argument to this function, which then uses these preferences and the challenges required by the server to determine which challenge to perform.

With the addition of the flag described above, the `preferences` parameter to this function should be the intersection of the challenges provided by the user in the new flag and the return value of `get_chall_pref`, with the order of the challenges being determined the user in the new flag.

For example, if the user provides `--preferred-challenges dns-01,tls-sni-01,http-01` and `get_chall_pref` returns `[tls-sni-01,dns-01]`, the value for `preferences` should be `[dns-01,tls-sni-01]`. Note: `gen_challenge_path` expects `acme.challenges.Challenge` classes, not simple strings as shown in my example.

If the intersection between the challenges requested by the user and those preferred by the plugin is empty, Certbot should error out (this may happen automatically, but this is the desired end result as there is no way to satisfy the CA and the user).

This issue also requests the ability to specify ports which is available through the `--http-01-port` and `--tls-sni-01-port` flags. Not much (or anything?) has to be done here, except maybe move them out of the testing section of help and fix up their documentation.

This was quite the brain dump, so if you have any questions, don't hesitate to ask.

SwartzCr:
PR at #3457

Issue #2505, opened 2016-02-19
Stop using simple_verify in manual plugin


bmw:
We can't use simple_verify reliably so we shouldn't be using it in the manual plugin.

Issue #1586, opened 2015-11-20
Simple-verify in all plugins


bmw:
We should attempt to verify our challenges for correctness in all plugins. Doing this would simplify debugging for us and our users.

TheNavigat:
I find this relatively confusing. I can't find any other plugins that use simple-verify.

bmw:
`manual` is the only plugin that uses `simple_verify`. This issue proposed that all of our plugins should use it.

TheNavigat:
I'm having a problem with adding ```simple_verify``` to ```standalone.py``` plugin.

```python
    def perform2(self, achalls):
        """Perform achallenges without IDisplay interaction."""
        responses = []

        for achall in achalls:
            if isinstance(achall.chall, challenges.HTTP01):
                server = self.servers.run(
                    self.config.http01_port, challenges.HTTP01)
                response, validation = achall.response_and_validation()
                self.http_01_resources.add(
                    acme_standalone.HTTP01RequestHandler.HTTP01Resource(
                        chall=achall.chall, response=response,
                        validation=validation))
            else:  # tls-sni-01
                server = self.servers.run(
                    self.config.tls_sni_01_port, challenges.TLSSNI01)
                response, (cert, _) = achall.response_and_validation(
                    cert_key=self.key)
                self.certs[response.z_domain] = (self.key, cert)
            self.served[server].add(achall)
            responses.append(response)

            if not response.simple_verify(
                    achall.chall, achall.domain,
                    achall.account_key.public_key(), self.config.http01_port):
                logger.warning("Self-verify of challenge failed.")

        return responses
```

That code fails with these errors.

```
======================================================================
ERROR: test_perform2 (letsencrypt.plugins.standalone_test.AuthenticatorTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/nav/Documents/letsencrypt/letsencrypt/plugins/standalone_test.py", line 177, in test_perform2
    responses = self.auth.perform2([http_01, tls_sni_01])
  File "/Users/nav/Documents/letsencrypt/letsencrypt/plugins/standalone.py", line 252, in perform2
    achall.account_key.public_key(), self.config.http01_port):
  File "/Users/nav/Documents/letsencrypt/acme/acme/challenges.py", line 432, in simple_verify
    return self.verify_cert(cert)
  File "/Users/nav/Documents/letsencrypt/acme/acme/challenges.py", line 396, in verify_cert
[...]

PR #3379, opened 2016-08-08
Handle dns01 challenge into the manual plugin [see #3466]

mlhamel:
Following changes made by @bmw at https://github.com/certbot/certbot/pull/2061

I really needed to be able to use the `dns-01` challenge using the manual mode.

Since `dns-01` challenge is now supported by the `acme` client of `certbot` this PR is making use of the challenge by adding a new command line argument called `supported-challenges`.

### DIFF
--- a/certbot/plugins/manual.py
+++ b/certbot/plugins/manual.py
@@ -195,7 +195,7 @@ s.serve_forever()" """
                                               uri=uri,
                                               command=command)
 
-            self._ip_logging_permission(response, formated_message)
+            self._ip_logging_permission(formated_message)
 
         if not response.simple_verify(
                 achall.chall, achall.domain,
@@ -211,7 +211,7 @@ s.serve_forever()" """
             formated_message = message.format(validation=validation,
                                               domain=achall.domain,
                                               response=response)
-            self._ip_logging_permission(response, formated_message)
+            self._ip_logging_permission(formated_message)
 
         if not response.simple_verify(
                 achall.chall, achall.domain,
@@ -241,7 +241,7 @@ s.serve_forever()" """
         sys.stdout.write(message)
         six.moves.input("Press ENTER to continue")
 
-    def _ip_logging_permission(self, response, formated_message):
+    def _ip_logging_permission(self, formated_message):
         # pylint: disable=missing-docstring
         if not self.conf("public-ip-logging-ok"):
             if not zope.component.getUtility(interfaces.IDisplay).yesno(
--- a/certbot/plugins/manual.py
+++ b/certbot/plugins/manual.py
@@ -4,24 +4,28 @@ import logging
 import pipes
 import shutil
 import signal
-import socket
 import subprocess
 import sys
 import tempfile
-import time
 
 import six
 import zope.component
 import zope.interface
 
+from functools import partial
+
 from acme import challenges
 
 from certbot import errors
 from certbot import interfaces
-from certbot.plugins import common
+from certbot.plugins import common, util
+from certbot.util import busy_wait
 
 
 logger = logging.getLogger(__name__)
+
+
+SUPPORTED_CHALLENGES = [challenges.HTTP01, challenges.DNS01]
 
 
 @zope.interface.implementer(interfaces.IAuthenticator)
@@ -41,7 +45,16 @@ class Authenticator(common.Plugin):
 
     description = "Manually configure an HTTP server"
 
-    MESSAGE_TEMPLATE = """\
+    MESSAGE_TEMPLATE = {
+        "dns-01": """\
+Make sure your dns configuration content the following key before continuing:
+
+{validation}
+
+if you didn't, make sure to add the validation as TXT record into your domain
+configuration.
+""",
+        "http-01": """\
 Make sure your web server displays the following content at
 {uri} before continuing:
 
@@ -51,7 +64,7 @@ If you don't have HTTP server configured, you can run the following
 command on the target server (as root):
 
 {command}
-"""
+"""}
 
     # a disclaimer about your current IP being transmitted to Let's Encrypt's servers.
     IP_DISCLAIMER = """\
@@ -86,10 +99,23 @@ s.serve_forever()" """
 
     @classmethod
     def add_parser_arguments(cls, add):
+        validator = partial(util.supported_challenges_validator,
+                            supported=SUPPORTED_CHALLENGES)
+
         add("test-mode", action="store_true",
             help="Test mode. Executes the manual command in subprocess.")
         add("public-ip-logging-ok", action="store_true",
             help="Automatically allows public IP logging.")
+        add("supported-challenges",
+            help="Supported challenges. Preferred in the order they are listed.",
+            type=validator,
+            default="http-01")
+
+    @property
+    def supported_challenges(self):
+        """Challenges supported by this plugin."""
+        return [challenges.Challenge.TYPES[name] for name in
+                self.conf("supported-challenges").split(",")]
 
     def prepare(self):  # pylint: disable=missing-docstring,no-self-use
         if self.config.noninteractive_mode and not self.conf("test-mode"):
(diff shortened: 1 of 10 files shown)


mlhamel:
Sorry for the noise with the @coveralls bot. I was fighting with linter and was not sure how to run it locally.

I'm sure also how increase the coverage. If you've got any idea...

pde:
Thanks for sending this! We'll get it reviewed shortly, and landed for the next release.

certbot/plugins/manual.py
```diff
@@ -41,7 +45,16 @@ class Authenticator(common.Plugin):
 
     description = "Manually configure an HTTP server"
 
-    MESSAGE_TEMPLATE = """\
+    MESSAGE_TEMPLATE = {
+        "dns-01": """\
```
pde: Change this template to be:
```
To prove control of the domain {domain}, please deploy a DNS TXT record with the following value:

{validation}

Once this is deployed, 
```
(the next line will be `Press ENTER to continue`).  You'll need to supply `achall.domain` as a parameter to the template.

certbot/plugins/manual.py
```diff
@@ -183,15 +211,30 @@ def _perform_single(self, achall):
...
-        # TODO: IDisplay wraps messages, breaking the command
-        #answer = zope.component.getUtility(interfaces.IDisplay).notification(
-        #    message=message, height=25, pause=True)
-        sys.stdout.write(message)
-        six.moves.input("Press ENTER to continue")
+    def _perform_dns01_challenge(self, achall):
+        response, validation = achall.response_and_validation()
+        if not self.conf("test-mode"):
+            if not self.conf("public-ip-logging-ok"):
+                if not zope.component.getUtility(interfaces.IDisplay).yesno(
```
pde: Slightly  unrelated to this PR, but If I run this with two domains, it asks for permission twice. Maybe we should just add an `else` case, like this:
```
else:
    self.config.namespace.manual_public_ip_logging_ok = True
```
So that it only asks once?

certbot/plugins/manual.py
```diff
@@ -183,15 +211,30 @@ def _perform_single(self, achall):
...
-    def _notify_and_wait(self, message):  # pylint: disable=no-self-use
-        # TODO: IDisplay wraps messages, breaking the command
-        #answer = zope.component.getUtility(interfaces.IDisplay).notification(
-        #    message=message, height=25, pause=True)
-        sys.stdout.write(message)
-        six.moves.input("Press ENTER to continue")
+    def _perform_dns01_challenge(self, achall):
+        response, validation = achall.response_and_validation()
+        if not self.conf("test-mode"):
+            if not self.conf("public-ip-logging-ok"):
```
pde: This same set of logic is in `_perform_http01_challenge`, it should be factored out in to its own utility method, something like `_ip_logging_permission`.

pde:
I note that if you try to run this without installing `dnspython`, it errors out:

```
  File "venv/bin/certbot", line 9, in <module>
  File "/home/pde/certbot/certbot/certbot/main.py", line 770, in main
  File "/home/pde/certbot/certbot/certbot/main.py", line 565, in obtain_cert
  File "/home/pde/certbot/certbot/certbot/main.py", line 104, in _auth_from_domains
  File "/home/pde/certbot/certbot/certbot/client.py", line 276, in obtain_and_enroll_certificate
  File "/home/pde/certbot/certbot/certbot/client.py", line 247, in obtain_certificate
    self.config.allow_subset_of_names)
  File "/home/pde/certbot/certbot/certbot/auth_handler.py", line 70, in get_authorizations
    resp = self._solve_challenges()
  File "/home/pde/certbot/certbot/certbot/auth_handler.py", line 108, in _solve_challenges
    resp = self.auth.perform(self.achalls)
  File "/home/pde/certbot/certbot/certbot/plugins/manual.py", line 147, in perform
    responses.append(mapping[achall.typ](achall))
  File "/home/pde/certbot/certbot/certbot/plugins/manual.py", line 222, in _perform_dns01_challenge
    if not zope.component.getUtility(interfaces.IDisplay).yesno(
  File "/home/pde/certbot/certbot/acme/acme/challenges.py", line 237, in simple_verify
    raise errors.Error("Local validation for 'dns-01' challenges "
Error: Local validation for 'dns-01' challenges requires 'dnspython'
```

Would it maybe make more sense to make that a call to `logger.warning` rather than `raise errors.Error`, so that people running `certbot-auto` don't need to try to figure out how to install `dnspython`?  Or do we really really want dnspython installed?

pde:
(Obviously the actual `dnspython` based checks would also need to be moved inside that `try:` block)

mlhamel:
Thanks,

Regarding `dnspython` I don't really see how we could remove this dependency. In fact, i'm wondering why it never became a first class citizen dependency since it's needed by `acme` in a way.

The only other way would be:

a) Change `acme` to use something like `nslookup` or `dig`
b) Find a way to perform a dns query by hand to extract the `TXT` entry of the domain.

Both solution are kind of anoying..

I would say the best way for now might to catch the exception and as you are suggesting, maybe just display a warning message and then exit. What do you think?

pde:
We could make it a first class dependency, but we've been fighting to reduce the number of dependencies we have (#1301) so adding more for non-universal use cases feels like going backwards.  Initially DNS01 is going to be a corner case, so it's probably okay to leave the current error in place and have people install dnspython if they're using us.  Our debian/redhat packages can list it as `Recommends`.  When we land fancier UI for this in the future (#2782) that might change...

mlhamel:
@pde If you don't mind i'm creating a new PR only to deal with missing dependency of `dnspython` cause it my imply changes into the ACME client also... What do you think?

bmw:
While you're welcome to create another PR, can you explain to me why you think `acme` needs to be changed? The library only requires `dnspython` if you're using both:

1.The DNS challenge.
2. The `simple_verify` method of the DNS challenge (which the `acme` module itself doesn't use).

As discussed in #2061, there are other ACME clients that use are `acme` module that don't satisfy the criteria above. Additionally, we've been moving away from using `simple_verify` at all (see #1586 and #2505). This may change in the future for the DNS challenge as we can more easily verify it than the `HTTP-01` and `TLS-SNI-01` challenges, but in the mean time, I don't think we should introduce `dnspython` as a required dependency when we've been actively trying to reduce the number of dependencies we have as @pde mentioned above.

mlhamel:
Yeah I totally agree with you, requiring `dnspython` is probably a bad idea. 

What I'm proposing regarding ACME and how to deal properly with this dependency is the following:

Right now at https://github.com/certbot/certbot/blob/master/acme/acme/challenges.py#L237 we are raising a simple `errors.Error` whenever someone try to use dns challenge without having `dnspython`. That's difficult to handle cause `Error` is super generic. We could then:

a) make use of Python 3 `raise from` using the `raise_from` syntax from `six` and chain the Error exception the ImportError caused by `dnspython` and check if there is ImportError in the traceback of the exception.

b) Create a custom exception only for missing dependencies that would inherit from `errors.Error`

c) Just catch `errors.Error` when using `simple_very` and interpret those as if the dependency was missing no matter what

Any preference?

bmw:
Ah okay. I totally agree with you that the error is too generic. This rules out option c. I think I have a slight preference for option b over c as it's a little simpler and consistent with errors in the rest of the `acme` module, but if you prefer a, I wouldn't have a problem with it.

mlhamel:
I've open another PR with my proposal of a new kind of exception https://github.com/certbot/certbot/pull/3439

mlhamel:
@bmw @pde do you have other concerns? What do you think of merging this PR and then I move to work on https://github.com/certbot/certbot/pull/3439 ?

bmw:
Sorry for the delay @mlhamel. EFF's Certbot team has been somewhat preoccupied with other things lately. Additionally, we've wanted to address #3343 which affects your PR. Instead of pushing all of this work on you, we were working on closing that issue which would then merge with your PR ourselves. I have done this work in #3466 which brings in both this PR and #3439.

PR #3383, opened 2016-08-08
Avoid importing conflicting security policy directives

pde:
Builds on and lands: #2262
Fixes: #574, #921, #1743

* Only add security options directives if they aren't already present in a `server` block
* Comment them as `# managed by Certbot` they're added, so that we can come back later and know they're safe to change


pde:
Power cycling to try the new travis tests...

bmw:
Currently testing on a VPS to make sure:

1. I didn't break anything.
2. This actually fixes all the issues we think it does.

bmw:
I think this is ready for another review. Most of the changes I made were stylistic or making the code more defensive. Additionally, I fixed a bug where we try to parse `/etc/letsencrypt/options-ssl-nginx.conf` before we copied the file to that location. I added a test to prevent regressions here.

One thing this PR does not do is use the `# managed by Certbot` comment as a signal that we can overwrite this directive. I think this is fine and we can add that functionality later, but I want to make sure this is known up front.

bmw:
closing and reopening to make travis run

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -506,8 +520,30 @@ def _add_directives(block, directives, replace):
...
+REPEATABLE_DIRECTIVES = set(['server_name', 'listen', 'include'])
+COMMENT = ' managed by Certbot'
+COMMENT_BLOCK = [' ', '#', COMMENT]
+
+
+def _comment_directive(block, location):
+    """Add a comment to the end of the line at location."""
+    next_entry = block[location + 1] if location + 1 < len(block) else None
+    if isinstance(next_entry, list) and next_entry:
+        if COMMENT in next_entry[-1]:
```
pde: Probably for maximum defensiveness we might also test that `next_entry` is longer than that and that `next_entry[-2] == "#"` or something?

bmw: That's not a bad idea. If we did that, we'd also have to check that `len(next_entry) >= 2`.

pde: Done. I could easily get tempted by further perfectionism (the code that lined all the "managed by Certbot" comments over at the right hand side of 80 columns got lost, for instance). But I think we should probably just merge this.

bmw:
LGTM

Issue #3377, opened 2016-08-07
`master` branch gives TypeError and quits


osirisinferi:
Recently, I updated my `certbot` to the latest Github master branch (Gentoo makes this very easy). Unfortunately, `certbot` stopped working. I confirmed the error on a separate Gentoo box in VirtualBox, which is as clean as it can be, so I'm doubting some weird setting I'm having on my "production" server causes the error.

```
Root logging level set at 0
Saving debug log to /var/log/letsencrypt/letsencrypt.log
certbot version: 0.9.0.dev0
Arguments: ['-debug', '-tvv', '-i', 'apache', '-a', 'webroot', '-d', 'le-test-12.example.com', '-w', '/var/www/vhosts/le-test-12.example.com/htdocs/', '--staging']
Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
Requested authenticator webroot and installer apache
Exiting abnormally:
Traceback (most recent call last):
  File "/usr/lib/python-exec/python2.7/certbot", line 9, in <module>
    load_entry_point('certbot==0.9.0.dev0', 'console_scripts', 'certbot')()
  File "/usr/lib64/python2.7/site-packages/certbot/main.py", line 770, in main
    return config.func(config, plugins)
  File "/usr/lib64/python2.7/site-packages/certbot/main.py", line 504, in run
    installer, authenticator = plug_sel.choose_configurator_plugins(config, plugins, "run")
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/selection.py", line 187, in choose_configurator_plugins
    installer = pick_installer(config, req_inst, plugins)
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/selection.py", line 31, in pick_installer
    config, default, plugins, question, (interfaces.IInstaller,))
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/selection.py", line 76, in pick_plugin
    verified.prepare()
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/disco.py", line 220, in prepare
    return [plugin_ep.prepare() for plugin_ep in six.itervalues(self._plugins)]
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/disco.py", line 106, in prepare
    self._initialized.prepare()
[...]

Issue #3319, opened 2016-07-23
Certbot apache plugin fails with TypeError: 'NoneType' object has no attribute '__getitem__'


j4m35:
I've just installed Certbot on my Gentoo system; using the following packages so that the GIT version will be pulled:
```
=app-crypt/certbot-apache-9999
=app-crypt/certbot-9999
=app-crypt/acme-9999
```

When I run certbot:
```
# certbot --apache
An unexpected error occurred:
TypeError: 'NoneType' object has no attribute '__getitem__'
Please see the logfiles in /var/log/letsencrypt for more details.
```

And the log is as follows:
```
# cat letsencrypt.log
2016-07-23 15:30:43,953:DEBUG:certbot.main:Root logging level set at 20
2016-07-23 15:30:43,954:INFO:certbot.main:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2016-07-23 15:30:43,962:DEBUG:certbot.main:certbot version: 0.9.0.dev0
2016-07-23 15:30:43,962:DEBUG:certbot.main:Arguments: ['--apache']
2016-07-23 15:30:43,963:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2016-07-23 15:30:43,965:DEBUG:certbot.plugins.selection:Requested authenticator apache and installer apache
2016-07-23 15:30:44,307:DEBUG:certbot.main:Exiting abnormally:
Traceback (most recent call last):
  File "/usr/lib/python-exec/python2.7/certbot", line 9, in <module>
    load_entry_point('certbot==0.9.0.dev0', 'console_scripts', 'certbot')()
  File "/usr/lib64/python2.7/site-packages/certbot/main.py", line 770, in main
    return config.func(config, plugins)
  File "/usr/lib64/python2.7/site-packages/certbot/main.py", line 504, in run
    installer, authenticator = plug_sel.choose_configurator_plugins(config, plugins, "run")
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/selection.py", line 184, in choose_configurator_plugins
    authenticator = installer = pick_configurator(config, req_inst, plugins)
  File "/usr/lib64/python2.7/site-packages/certbot/plugins/selection.py", line 24, in pick_configurator
    (interfaces.IAuthenticator, interfaces.IInstaller))
[...]

PR #3385, opened 2016-08-09
Refactored get_file_path

joohoi:
Instead of fiddling with predefined apache statement strings, do an actual file path verification for augeas paths. 

This eliminates the problem of having to predefine all possible Apache configuration file categories.

Also even though nobody(?) has hit it before, allows people to use vhost config file names starting with strings "ifmodule", "virtualhost" and "macro".

Fixes: #3377 #3319

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1801,25 +1801,17 @@ def get_file_path(vhost_path):
     :rtype: str
 
     """
-    # Strip off /files
-    avail_fp = vhost_path[6:]
-    # This can be optimized...
-    while True:
-        # Cast all to lowercase to be case insensitive
-        find_if = avail_fp.lower().find("/ifmodule")
-        if find_if != -1:
-            avail_fp = avail_fp[:find_if]
-            continue
-        find_vh = avail_fp.lower().find("/virtualhost")
-        if find_vh != -1:
-            avail_fp = avail_fp[:find_vh]
-            continue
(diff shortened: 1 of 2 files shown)


SwartzCr:
This looks good - boulder integration is failing, but it might be a fluke so I re-ran it.
I will note that you don't add back the leading slash so the `self.aug.get("/augeas/files%s/path" % get_file_path(path))` command will result in something like "/augeas/filesetc/apache2 ..." instead of "/augeas/files/etc/apache2..." - so that should be fixed
Otherwise I'm fine with this method, though still would be happy to see a check for .startswith("/files/" at the beginning of get_file_path

joohoi:
The leading slash is re-added here on the first pass of the loop: https://github.com/certbot/certbot/blob/filepath_refactor/certbot-apache/certbot_apache/configurator.py#L1815

I actually made sure that the method returns the exact same values (as per tests) for the working preconditions than the previous implementation. (Had originally both methods in, and compared the results).

joohoi:
The boulder integration tests are failing because of:
```
./tests/boulder-fetch.sh: line 9: docker-compose: command not found
```
https://travis-ci.org/certbot/certbot/jobs/151073313#L1838

That's something outside of this PR.

SwartzCr:
Ah yes, looks good to me, let's see if it fixes the problem for people - but I'm happy to merge once the tests pass

pde:
This was stuck needing a merge with master. Weirdly, I didn't see any conflicts when I did that manually?

SwartzCr:
looks like travis is fixed! merging

PR #3390, opened 2016-08-10
Add more ignored files to gitignore.

jsha:


bmw:
Assuming this was just to play with Travis while we were having issues? If you actually wanted to see this PR merged, let me know and I'll reopen.

jsha:
Nope, these are actual files that exist in my certbot directory that I wanted to ignore. Vim creates multiple swap files as foo.swp, foo.swo, foo.swn, etc. And tags is from exuberant ctags.

bmw:
Ah gotcha! Sorry about that. I'll merge when Travis is green.

Issue #3380, opened 2016-08-08
Plugins with hyphens do not receive their args during renewal


code-tree:
When filtering config parameters from renewal config files, certbot checks their prefix to see if they are specific to a plugin. However, the prefix it tests with has not had hyphens replaced with underscores, where as the config parameters it tests against *have*.

Example: https://github.com/dlapiduz/letsencrypt-s3front
Solution: https://github.com/code-tree/certbot/commit/74166b6d6e37dd344d64bffe741fb6fea4be3335

Patch fixes it for me, though there may be a cleaner solution to this?

PR #3398, opened 2016-08-12
Replace '-' with '_' before filtering plugin settings

pde:
This bug notably occurs when renewing certs for with the plugin `letsencrypt-s3front`

Fixes: #3380


pde:
Thanks @code-tree

PR #3409, opened 2016-08-15
Revert "Set dialog widgets to use autowidgetsize"

SwartzCr:
Reverts certbot/certbot#3268


SwartzCr:
merged before doing appropriate tests - going to revert and try on other systems

Issue #3408, opened 2016-08-15
Experimental support on gentoo auto-installs packages


SwartzCr:
reportd by @mstrache in #3319 - emerge doesn't use the --ask flag when installing system packages:

> Btw, nice job automatically calling emerge to install missing dependencies. Only one thing: it would be nice if you run emerge --ask ... to give the user the chance to review dependencies and cancel the process. As a gentoo user I like to have control over what packages are installed (it's just a gentoo thing).

@mstrache would you be willing to update the code to use the right flags? The gentoo bootstrap script is defined here: https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
and should use the yes flag as specified here: https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/letsencrypt-auto#L191

mstrache:
Change committed. I generally would recommend to use `local` to declare variables inside functions. It prevents nasty side effects from recurring variable names.

SwartzCr:
Not only did we merge a change to ask before bootstrapping, but we're now installable via emerge, so I am closing this

PR #3410, opened 2016-08-15
#3408: Made Gentoo bootstrapping asking before performing any changes

mstrache:
Made Gentoo bootstrapping asking the user before performing any changes. This works for emerge and pmerge. Cave has been left out, since it has no equivalent option (or at least I do not know any). Overriding this behaviour using the `--non-interactive` option is possible.

### DIFF
--- a/letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
+++ b/letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
@@ -9,15 +9,20 @@ BootstrapGentooCommon() {
     app-misc/ca-certificates
     virtual/pkgconfig"
 
+  local ASK_OPTION="--ask"
+  if [ "$ASSUME_YES" = 1 ]; then
+    ASK_OPTION=""
+  fi
+
   case "$PACKAGE_MANAGER" in
     (paludis)
       $SUDO cave resolve --preserve-world --keep-targets if-possible $PACKAGES -x
       ;;
     (pkgcore)
-      $SUDO pmerge --noreplace --oneshot $PACKAGES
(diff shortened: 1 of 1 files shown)


pde:
I don't know much about the universe of gentoo package management, but if `cave` is often or even sometimes going to be used, you could ask the [same way that the Debian bootstrapper does it](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh#L58).

That can probably be work for a future PR, unless you're inclined to refactor everything to use a single shell `read` statement instead of `--ask` for all three tools.

pde:
Aside from that: which gentoo variants has this been tested on? Which ones *should* it be tested on? We currently don't have gentoo support in our test farm tests :(.

SwartzCr:
related to issue #3408

mstrache:
From what I know, Cave isn't that common. Emerge is pretty good and therefore used by most of the users.

There are no distinguished Gentoo variants, since each installation is completely custom made. My setup contained systemd and apache 2.4.23. There are some other Gentoo based distributions, like Sabayon, but if it works on a recent gentoo, it should work with those as well.

Implement the ask feature myself would be a solution, true. I'll add it, when I find some time.

letsencrypt-auto-source/pieces/bootstrappers/gentoo_common.sh
```diff
@@ -9,15 +9,20 @@ BootstrapGentooCommon() {
     app-misc/ca-certificates
     virtual/pkgconfig"
 
+  local ASK_OPTION="--ask"
```
bmw: We run `certbot-auto` with `/bin/sh` and `local` isn't POSIX compliant so it should be removed.

bmw:
LGTM. Thanks @mstrache.

PR #3415, opened 2016-08-16
clarify invalid email error in non-interactive

### LINKED ISSUE #3282: Unclear error about invalid email in non-interactive mode
If an invalid email address is specified in the non-interactive mode, it is not clear from the error message what's wrong:
```
# certbot --non-interactive --agree-tos --email=qwe@qwe.qwe register
You should register before running non-interactively, or provide --agree-tos and --email <email_address> flags
```
Increase the verbosity with `-vvv` makes sense:
```
2016-07-13 13:21:01,448:DEBUG:acme.client:Received response <Response [400]> (headers: {'Content-Length': '123', 'Boulder-Request-Id': 'jSUozaFbKZV7fCbfIJvfLG6-J9WxYEcl3osWRXnKS0E', 'Expires': 'Wed, 13 Jul 2016 07:21:01 GMT', 'Server': 'nginx', 'Connection': 'close', 'Pragma': 'no-cache', 'Cache-Control': 'max-age=0, no-cache, no-store', 'Date': 'Wed, 13 Jul 2016 07:21:01 GMT', 'Content-Type': 'application/problem+json', 'Replay-Nonce': 'wDx_-VvU_PzjfRWovewncePYqQ4frdUw3dC29B5LKbc'}): '{\n  "type": "urn:acme:error:invalidEmail",\n  "detail": "DNS problem: NXDOMAIN looking up MX for qwe.qwe",\n  "status": 400\n}'
2016-07-13 13:21:01,450:DEBUG:certbot.main:Exiting abnormally:
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/bin/certbot", line 11, in <module>
    sys.exit(main())
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/main.py", line 744, in main
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/main.py", line 388, in register
    _determine_account(config)
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/main.py", line 345, in _determine_account
    config, account_storage, tos_cb=_tos_cb)
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/client.py", line 120, in register
    regr = perform_registration(acme, config)
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/client.py", line 153, in perform_registration
    config.namespace.email = display_ops.get_email(invalid=True)
  File "/usr/local/psa/var/modules/letsencrypt/venv.X5yGo/lib/python2.7/site-packages/certbot/display/ops.py", line 52, in get_email
    raise errors.MissingCommandlineFlag(msg)
MissingCommandlineFlag: You should register before running non-interactively, or provide --agree-tos and --email <email_address> flags
```
In the interactive mode an email address is asked again:
```
There seem to be problems with that address. Enter email address (used for urgent notices and lost key recovery)
```
This is much better, but I would expect the detailed error `DNS problem: NXDOMAIN looking up MX for qwe.qwe` to be displayed.

SwartzCr:
fixes #3282

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -154,7 +154,7 @@ def perform_registration(acme, config):
                 msg = ("The email you specified was unable to be verified "
                        "by acme. Please ensure it is a valid email and "
                        "attempt registration again.")
-                raise erros.MissingCommandlineFlag(msg)
+                raise errors.MissingCommandlineFlag(msg)
             else:
                 config.namespace.email = display_ops.get_email(invalid=True)
                 return perform_registration(acme, config)
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -150,8 +150,14 @@ def perform_registration(acme, config):
         return acme.register(messages.NewRegistration.from_data(email=config.email))
     except messages.Error as e:
         if e.typ == "urn:acme:error:invalidEmail":
-            config.namespace.email = display_ops.get_email(invalid=True)
-            return perform_registration(acme, config)
+            if config.noninteractive_mode:
+                msg = ("The email you specified was unable to be verified "
+                       "by acme. Please ensure it is a valid email and "
+                       "attempt registration again.")
+                raise erros.MissingCommandlineFlag(msg)
+            else:
+                config.namespace.email = display_ops.get_email(invalid=True)
+                return perform_registration(acme, config)
         else:
             raise
 
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -151,10 +151,10 @@ def perform_registration(acme, config):
     except messages.Error as e:
         if e.typ == "urn:acme:error:invalidEmail":
             if config.noninteractive_mode:
-                msg = ("The email you specified was unable to be verified "
-                       "by acme. Please ensure it is a valid email and "
-                       "attempt registration again.")
-                raise errors.MissingCommandlineFlag(msg)
+                msg = ("The ACME server believes %s is an invalid email address. "
+                       "Please ensure it is a valid email and attempt "
+                       "registration again." % config.email)
+                raise errors.Error(msg)
             else:
                 config.namespace.email = display_ops.get_email(invalid=True)
                 return perform_registration(acme, config)
--- a/certbot/tests/client_test.py
+++ b/certbot/tests/client_test.py
@@ -61,13 +61,13 @@ class RegisterTest(unittest.TestCase):
 
     @mock.patch("certbot.account.report_new_account")
     @mock.patch("certbot.client.display_ops.get_email")
-    def test_email_retry(self, _rep, mock_get_email):
+    def test_email_invalid_noninteractive(self, _rep, mock_get_email):
         from acme import messages
         msg = "DNS problem: NXDOMAIN looking up MX for example.com"
         mx_err = messages.Error(detail=msg, typ="urn:acme:error:invalidEmail")
         with mock.patch("certbot.client.acme_client.Client") as mock_client:
             mock_client().register.side_effect = [mx_err, mock.MagicMock()]
-            self.assertRaises(errors.MissingCommandlineFlag, self._call)
+            self.assertRaises(errors.Error, self._call)
 
     def test_needs_email(self):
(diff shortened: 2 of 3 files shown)


SwartzCr:
So unfortunately there's not a really good way to pass the error from the failed registration without modding the acme register function. Since we only want that error back in this specific case it seems unwise to always pass back the error details - just to print them here. Since line 153 of client.py is the only place we call `display_ops.get_email(invalid=True)` I think it's fine to just add to this hard coded error, since it will only trigger in the situation where you've entered an invalid email with noninteractive mode specified.

pde:
Rather than raising an error message with many possible explanations in it, I think it would be better to detect `--non-interactive` mode [here](https://github.com/certbot/certbot/blob/master/certbot/client.py#L153) and print an appropriate error rather than trying to re-request input?

SwartzCr:
good point, I've moved the error out to client.py, checking config.noninteractive_mode

certbot/client.py
```diff
@@ -150,8 +150,14 @@ def perform_registration(acme, config):
         return acme.register(messages.NewRegistration.from_data(email=config.email))
     except messages.Error as e:
         if e.typ == "urn:acme:error:invalidEmail":
-            config.namespace.email = display_ops.get_email(invalid=True)
-            return perform_registration(acme, config)
+            if config.noninteractive_mode:
+                msg = ("The email you specified was unable to be verified "
```
pde: "Unable to be verified" seems a bit confusing... Maybe a first sentence like 'The ACME server believes {email} is an invalid email address'.

certbot/client.py
```diff
@@ -150,8 +150,14 @@ def perform_registration(acme, config):
...
+                       "by acme. Please ensure it is a valid email and "
+                       "attempt registration again.")
+                raise errors.MissingCommandlineFlag(msg)
```
pde: Just raise `errors.Error`, I think?

pde: (since the problem isn't that the command line flag is missing, but that it has an invalid value)

certbot/tests/client_test.py
```diff
@@ -67,8 +67,7 @@ def test_email_retry(self, _rep, mock_get_email):
         mx_err = messages.Error(detail=msg, typ="urn:acme:error:invalidEmail")
         with mock.patch("certbot.client.acme_client.Client") as mock_client:
             mock_client().register.side_effect = [mx_err, mock.MagicMock()]
-            self._call()
-            self.assertEqual(mock_get_email.call_count, 1)
+            self.assertRaises(errors.MissingCommandlineFlag, self._call)
```
pde: Hmmm. I think you want to test this error-raising property (in the case where you're in non-interactive mode), but keep the previous test in place (in the case where you're not in non-interactive mode).

If I had to guess, the test probably failed because we were defaulting to being in `non-interactive` mode due to `mock`ing things?

SwartzCr: yeah - hmm open to suggestions on how to preserve that without mock spinning forever (because it'll just keep prompting)

certbot/tests/client_test.py
```diff
@@ -60,15 +60,13 @@ def test_it(self):
                 self._call()
 
     @mock.patch("certbot.account.report_new_account")
-    @mock.patch("certbot.client.display_ops.get_email")
-    def test_email_retry(self, _rep, mock_get_email):
```
pde: This is almost ready! Except you should put `test_email_retry` back in the test suite, in addition to the new `test_email_invalid_noninteractive` :).  You'll probably need to explicitly set `self.config.noninteractive_mode = False` to get it to work.

SwartzCr:
@pde to merge

pde:
LGTM!

Issue #3422, opened 2016-08-17
Broken using.html#plugins fragment link


pde:
https://certbot.eff.org/docs/using.html#plugins

PR #3423, opened 2016-08-17
Unbreak using.html#plugins links

pde:
Fixes: #3422


pde:
Self-merging a trivial doc fix

Issue #3416, opened 2016-08-16
Missing links in README


jsha:
The terms client-dev, Freenode, and OFTC in the README are supposed to be named links, but the references, which used to be at the bottom of the README, got deleted at some point. They should be restored.

bmw:
Related: Looks like `resources.rst` and the README both contain these links, however, `resources.rst` doesn't have this problem. Ideally, we should fix up the README and remove the duplication in `resources.rst`.

PR #3425, opened 2016-08-17
Fix the links for #3416 and align `README.rst` & `docs/resources.rst`

### LINKED ISSUE #3341: Remove psutil from requirements.txt
We need to do this during the release process for 0.9.0

jpluimers:
Fix the links for #3416 and align content of in `README.rst` and `docs/resources.rst` so it's easier to later de-dupe 

(I've not done this now as in `README.rst` does too much `tag:` fiddling and I'm not sure how that will work out if the fiddling is not aware of `.. include::`.

Dang: seems github includes merges after you've fired the pull request.

Anyway: this then also fixes #3341


jpluimers:
Abandon. I managed to get unrelated commits in this pull-request.

PR #3430, opened 2016-08-17
Add the Nginx plugin to certbot-auto

bmw:
and prep the release script for this change.

I verified `pyparsing`'s hash on 4 different systems, but it doesn't hurt to have the reviewer check the hash as well.


pde:
Great!

PR #3432, opened 2016-08-17
Hide the Nginx plugin

### LINKED ISSUE #3427: Nginx plugin selection
We want to slowly roll out the Nginx plugin to make sure it's working for people. To do this we should:
* Mark the nginx plugin as hidden (we did this for the manual plugin)
* Disable automatic selection if it's the only available configurator (if not already the case, we should disable automatic selection of hidden plugins).
* Make sure the description of the Nginx plugin is sane (it used to say it was broken).

bmw:
Fixes #3427. Turns out this was trivial.


pde:
LGTM!

PR #3435, opened 2016-08-18
Fix the links for #3416 and align `README.rst` & `docs/resources.rst`

jpluimers:
Fix the links for #3416 and align content of in README.rst and docs/resources.rst so it's easier to later de-dupe

(I've not done this now as in README.rst does too much tag: fiddling and I'm not sure how that will work out if the fiddling is not aware of .. include::.


bmw:
LGTM. Thanks for the PR @jpluimers.

PR #3436, opened 2016-08-18
Resolve "Remove psutil from requirements.txt #3341" [revisions requested]

### LINKED ISSUE #3341: Remove psutil from requirements.txt

jpluimers:
The change is actually in `./letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt` and fixes #3341


jpluimers:
Via IRC: also update these:

```
./certbot-auto
./letsencrypt-auto
./letsencrypt-auto-source/letsencrypt-auto
```

Background from IRC (thanks @joohoi ):

| file                                           | comment                                                                                     |
| -----------------------------------------------|---------------------------------------------------------------------------------------------|
| ./certbot/plugins/util.py                      | can work without psutil                                                                     |
| ./certbot/plugins/util_test.py                 | can work without psutil                                                                     |
| ./certbot-auto                                 | creates own `$TEMP_DIR/letsencrypt-auto-requirements.txt` with reference to `psutil==3.3.0` |
| ./letsencrypt-auto                             | creates own `$TEMP_DIR/letsencrypt-auto-requirements.txt` with reference to `psutil==3.3.0` |
| ./letsencrypt-auto-source/letsencrypt-auto     | creates own `$TEMP_DIR/letsencrypt-auto-requirements.txt` with reference to `psutil==3.3.0` |
| ./setup.py                                     | `psutil` referenced only in `dev_extras` so I think it's safe                               |
| ./tox.ini                                      | `deps` has reference `py{26,27}-oldest: psutil==2.1.0`                                      |

> so it should be removed from certbot-auto, and both letsencrypt-auto scripts leaving setup.py, tox.ini and util files intact
> this was the original PR that handled making previously required psutil dependency as optional: #3280 
> but I made that exact mistake - removing all the dependencies prematurely, causing the breakage. this was the fixing PR: #3335

bmw:
The way we may changes to `{certbot,letsencrypt}-auto` is kind of complicated/convoluted. Changes to `certbot-auto` and `letsencrypt-auto` in the root directory should only happen during releases (see #3337). The way to modify files in `letsencrypt-auto-source` is to not modify `letsencrypt-auto-source/letsencrypt-auto` directly, but to modify the other files contained in that directory and run `build.py` which will combine all the individual pieces to create `letsencrypt-auto-source/letsencrypt-auto`.

To fix this PR, you should revert the changes to `{certbot,letsencrypt}-auto` in the root. Additionally, if you want to remove the unnecessary whitespace you found, you should modify `letsencrypt-auto-source/pieces/fetch.py`. After doing that, run `build.py` and update the PR containing the resulting changes to all three files.

Like your other PR, I'm appending "[revisions requested]" to the title. Please remove this when you'd like another review.

bmw:
Revisions were never made on this PR and we're planning on releasing 0.9.0 tomorrow. Given making this change in `master` without a release will break things, I will just make this change during the release tomorrow. (What I proposed above would only break `letsencrypt-auto-source/letsencrypt-auto` which isn't a big deal, but it'd obviously be nice to avoid it.)

Thanks for taking a look at this though @jpluimers.

jpluimers:
Sorry, I had hoped to get back to it sooner but then reality overtook me on the right side. So thanks for taking this out of my hands.

bmw:
No worries! I appreciate you working on this and I completely understand getting busy with real life. I just wanted to close this issue to clean up our PR queue a bit.

PR #3437, opened 2016-08-18
Make return type of certbot.interfaces.IInstaller.get_all_keys_certs() an iterator

SnijderC:
We're developing a plugin for certbot called certbot-haproxy, for HAProxy as the name implies. There are 2 likely scenario's we are considering for the development of the plugin. 
1. HAProxy is used as a load-balancer/high availability proxy for a single or a few large websites with multiple back-ends.
2. HAProxy is used as a load-balancer/high availability proxy for many websites (up to thousands of websites) with multiple back-ends.

Our use case is the latter. Because we are dealing with bigger numbers we need to be careful not to create stuff that does not perform well when run repeatedly like in for loops.

We have a working proof of concept, that implements the interfaces specified by Certbot and is capable of authorisation and installation of certificates. It looks like it will scale pretty well too. 

There is however 1 thing that does not perform well when the amount of certificates goes beyond a thousand or so. It's `cerbot.interfaces.IInstaller.get_all_certs_keys()` ([link](https://github.com/certbot/certbot/blob/5d8744bf7cd62d999ed130f912bf50c72d2c2de1/certbot/interfaces.py#L300)), which is supposed to return a `list` of `tuple`'s containing the path of the certificate, key and configuration file.

``` python
def get_all_certs_keys():
    """Retrieve all certs and keys set in configuration.
    :returns: tuples with form `[(cert, key, path)]`, where:
    - `cert` - str path to certificate file
    - `key` - str path to associated key file
    - `path` - file path to configuration file
    :rtype: list
    """
```

HAProxy does know the concept of "VirtualHost" (Apache) or "server" (nginx), instead we rely on ACL's to pass the requests for each website to the correct back-end. These ACL's do not relate to the configured certificates in the configuration file. Which means we can not implement this function the conventional way i.e.: find all VirtualHost's and parse them to find paths to certificates. We have to inspect our certificate directory instead, parsing each certificate to check that they are in fact certificates issued by Let's Encrypt. This has to be done for every certificate every time the method is called since we need to return a `list` of `tuple`'s. Performance is still mostly linear per call, but decreases with every issued certificate, which is not a best practice for software development.

We therefore want to propose a change in the interface:

``` python
def get_all_certs_keys():
    """Retrieve all certs and keys set in configuration.

    :returns: An iterator that returns `(cert, key, path)` for every
        iteration, where:
        - `cert` - str path to certificate file
        - `key` - str path to associated key file
        - `path` - file path to configuration file

    :rtype: iterator

    """
```

**Pro's**
- Performance is slightly increased regardless of the plugin because: 
  - Memory use is decreased, even if only **very slightly** in most cases.
  - If the first or one of the first keys is the one you are looking for, the others don't even need to be parsed.

**Cons**
- The interface is already defined and implementations already exist, maybe we can come up with a workaround that allows both a `list` or an `iterator`? But this would depend on the purpose of this method, which at this point is not very clear to me.
  - It seems a non-issue since it's only used in unittests.
  - I have updated all unittests and `get_all_certs_keys()` implementations that are in certbot-apache and certbot-nginx in my pull-request.

**Questions**
- We aren't aware if there is any specific reasoning for this to return a `list` instead of an `iterator`, is there?
- We couldn't really find any call to this method other than in the unittests, is this a future use kind of thing?

I ran the tox tests according to the contribution guidelines, <s>it doesn't seem any issues arose from my changes</s>. Had a few issues with Python 2.6 compatibility, fixed that along with some obvious errors, now all tests seem to pass.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1503,7 +1503,6 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         :rtype: list
 
         """
-        c_k = set()
 
         for vhost in self.vhosts:
             if vhost.ssl:
@@ -1517,11 +1516,10 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                 if cert_path and key_path:
                     cert = os.path.abspath(self.parser.get_arg(cert_path[-1]))
                     key = os.path.abspath(self.parser.get_arg(key_path[-1]))
-                    c_k.add((cert, key, get_file_path(cert_path[-1])))
+                    yield (cert, key, get_file_path(cert_path[-1]))
                 else:
                     logger.warning(
                         "Invalid VirtualHost configuration - %s", vhost.filep)
-        return c_k
 
     def is_site_enabled(self, avail_fp):
         """Checks to see if the given site is enabled.
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1496,11 +1496,11 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         Retrieve all certs and keys set in VirtualHosts on the Apache server
 
-        :returns: list of tuples with form [(cert, key, path)]
+        :returns: iterator that returns tuples with form (cert, key, path)
             cert - str path to certificate file
             key - str path to associated key file
             path - File path to configuration file.
-        :rtype: list
+        :rtype: iterator
 
         """
 
--- a/certbot-apache/certbot_apache/tests/configurator_test.py
+++ b/certbot-apache/certbot_apache/tests/configurator_test.py
@@ -786,9 +786,9 @@ class MultipleVhostsTest(util.ApacheTest):
     def test_get_all_certs_keys_malformed_conf(self):
         self.config.parser.find_dir = mock.Mock(
             side_effect=[["path"], [], ["path"], [], ["path"], []])
-        c_k = self.config.get_all_certs_keys()
+        c_k = set(self.config.get_all_certs_keys())
 
-        self.assertIsNone(c_k)
+        self.assertEqual(set(c_k), set([]))
 
     def test_more_info(self):
         self.assertTrue(self.config.more_info())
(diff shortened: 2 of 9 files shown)


cowlicks:
Since this is an API breaking change it might be better to just add another function instead. Like `get_cert_key_iterator`. Alternatively you could add a `iterator` keyword argument that defaults to `False`.

SnijderC:
@cowlicks Well I would make it another method instead but that would be besides the point because I think it would be better if nobody would call this method the way it is currently described in the interface.

Adding a keyword could work, or maybe wherever this function will be called, in the future (as I don't see it called anywhere right now), check that you're dealing with an iterator or a list, and converting lists to iterators first, e.g.:

```
all_certs_keys = installer.get_all_certs_keys()
if isinstance(all_certs_keys, list):
    all_certs_keys = iter(all_certs_keys)
# do action x on all_certs_keys
```

Though most likely, whatever is called on `all_certs_keys` will be `for cert, key, config in all_certs_keys: [...]` or map(), filter() etc. which also take iterable objects.

BTW, can anyone please answer the questions, what this function is intended for? And where it is used?

pde:
Beginning by trying to answer a few of the questions here!

1. What is the purpose of `get_all_certs_keys()`? That's a great question, because the folks who added it to the interface are not presently active on the project :).  It isn't called by Certbot at the moment. So I'm trying to think of what it might be needed for in the future:
  * For automatically managing security enhancements (a la #1123 ) , ask the installer which certs it's currently using so that we know which ones might need security setting tweaks (but perhaps the `RenewableCert` / lineage objects can do that for you anyway? The advantage of `get_all_certs_keys()` is that it would reflect any changes made to config files after Certbot last worked with them.)
  * For asking installers about certs that came from somewhere other than Certbot/Let's Encrypt, in cases where there's something we should do with or to them, like auditing for weaknesses in the future?
2. What should we do with it? It seems that either accepting a `list` or an `iterator` is least likely to cause breakage. Or if we don't find any compelling use cases (the ones above seem very speculative) perhaps we should just remove it from the interface.

pde:
@bmw thoughts?

bmw:
So I just made two issues for what I'd like to see here which are #3748 and #3749. If you'd like to continue work on this, please send another PR and I'll happily review.

PR #3440, opened 2016-08-18
Allow tests to run when psutil isn't available

### LINKED ISSUE #3438: Tests fail on Certbot package due to missing psutil dependency
With the change of moving `psutil` as a `dev` extras dependency, our unit tests now fail when `psutil` isn't installed. This will likely be a problem for package maintainers/users who use our unit tests to make sure the package is working correctly on their system. We should either:

1. Move `psutil` back (temporarily) as a required dependency.
2. Fix our unit tests so they don't depend on the package.

bmw:
Fixes #3438.

This adds a dependency of `unittest2` on Python 2.6 systems which already exists indirectly due to `configargparse` ([source1](https://github.com/bw2/ConfigArgParse/blob/0.10.0/setup.py#L76) and [source2](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt#L161)).

### DIFF
--- a/certbot/plugins/util_test.py
+++ b/certbot/plugins/util_test.py
@@ -4,13 +4,7 @@ import unittest
 import sys
 
 import mock
-
-try:
-    # Python 3.5+
-    from importlib import reload as refresh  # pylint: disable=no-name-in-module
-except ImportError:
-    # Python 2-3.4
-    from imp import reload as refresh
+from six.moves import reload_module  # pylint: disable=import-error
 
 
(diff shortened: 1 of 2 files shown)


bmw:
Unless we're rushing to get 0.9.0 out the door, we should hold off on merging this. While this doesn't introduce conditional behavior based on Python version, I'm seeing some weird behavior in our wheels on PyPI due to this.

bmw:
Closing in favor of #3450.

PR #3445, opened 2016-08-19
fix #3347 - quote domain name [needs revision]

### LINKED ISSUE #3347: Error message for empty domain name is incorrect
In a recent commit that [Fix FQDN checks](https://github.com/certbot/certbot/commit/6156b452bcf7c924f12c334e7094ff5a9049ee37),
 the error message fails to address the case where domain name is an empty string. See below.

`--domain` is a CSV string, but the 2nd DNS name is parsed to be an empty string, which caused 
and error message `Requested domain  is not a FQDN because label  is too long.`. 

It would have been more logically to say `Requested domain "" is not a FQDN because label  is either empty or too long.`

The check it failed is here: https://github.com/certbot/certbot/blob/6156b452bcf7c924f12c334e7094ff5a9049ee37/certbot/util.py#L432

``` bash
$ docker run --rm -it --entrypoint bash quay.io/letsencrypt/letsencrypt:latest
root@4b80271cdd6c:/opt/certbot# certbot certonly \
>     --webroot \
>     --non-interactive \
>     --agree-tos \
>     --email test@test.com \
>     --webroot-path /var/lib/letsencrypt \
>     --domain www.example.com,,test1.example.com \
>     --no-verify-ssl \
>     --debug \
>    --test-cert
Requested domain  is not a FQDN because label  is too long.
```

jpluimers:
This makes it more clear when passing an empty domain name leads to the error:

```
Requested domain  is not a FQDN because label is either empty or too long.
```

now has become

```
Requested domain "" is not a FQDN because label is either empty or too long.
```

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -451,7 +451,8 @@ def enforce_domain_sanity(domain):
     # FQDN checks according to RFC 2181: domain name should be less than 255
     # octets (inclusive). And each label is 1 - 63 octets (inclusive).
     # https://tools.ietf.org/html/rfc2181#section-11
-    msg = "Requested domain {0} is not a FQDN because ".format(domain)
+    # quoted domain just in case the domain is empty - see issue #3347
+    msg = "Requested domain \"{0}\" is not a FQDN because ".format(domain)
     labels = domain.split('.')
     for l in labels:
         if not 0 < len(l) < 64:


bmw:
I get:
```
Requested domain "" is not a FQDN because label  is too long.
```
Maybe this PR is missing a commit you have locally. If so, we should probably also put quotes around the specific label in the domain name that's causing problems to avoid the double space as shown in my error message.

Perhaps a better approach, however, is to move some of this error handling logic around. After the `labels = domain.split('.')` line, we can check if the entire domain is too short or long. If the domain is empty, raise an error saying so. You can move the `if len(domain) > 255` before the `for` loop.

In the `for` loop, we could have separate branches for the label being empty or too long. This allows us to simply say things like `Requested domain ".example.com" is not a FQDN because it contains an empty label.` rather than including `""` and a less helpful error message.

I personally think this route is better, but if you decide to go this way, please also add tests for the new branches to https://github.com/certbot/certbot/blob/master/certbot/tests/util_test.py#L333 (looks like we're already missing some test coverage there). Due to lack of static checks in Python, I think simple tests like this are important to avoid shipping code with silly syntax errors (see #2822).

I'm appending "[revision requested]" to the title of this PR. Please remove it when you'd like another review. Thanks for the contribution @jpluimers!

bmw:
Since we haven't seen any progress on this, I'm closing the PR to clean up our PR queue a bit. If you'd like to continue work on this, just open a new PR when it's ready for another review and I'll happily take a look.

PR #3448, opened 2016-08-21
Fix apache logs dir for centos

DanCld:
CentOS stores Apache log files in /var/log/httpd instead of /var/log/apache2 . Running certbot on "full auto mode" will prevent httpd from starting with the new configuration.


joohoi:
LGTM

bmw:
I ran this against our test farm as well with no issues.

Thanks for the PR @DanCld.

PR #3450, opened 2016-08-23
Psutil tests

### LINKED ISSUE #3438: Tests fail on Certbot package due to missing psutil dependency

bmw:
Fixes #3438.

I like this better than #3440 because it doesn't add an optional dependency for us which it turns out doing properly is a real pain in Python packaging (e.g. we don't do it right).

@SwartzCr or @joohoi, if one of you could review this I'd appreciate it. This needs to be merged before we can do a release off of `master`. Currently, our release script errors out.


certbot/plugins/util_test.py
```diff
@@ -81,6 +75,42 @@ def test_ports_blocked(self, mock_getutil):
         self.assertEqual(mock_getutil.call_count, 2)
 
 
+def psutil_available():
+    """Checks if psutil can be imported.
+
+    :rtype: bool
+    :returns: True iff psutil is installed and can be imported
```
joohoi: iff

certbot/plugins/util_test.py
```diff
@@ -81,6 +75,42 @@ def test_ports_blocked(self, mock_getutil):
...
+    return True
+
+
+def skipUnless(condition, reason):
+    """Skip tests unless a condition holds.
+
+    This implements the basic functionality of unittest.skipUnless
+    which is only available on Python 2.7+.
+
+    :param bool condition: skip the test iff condition is False
```
joohoi: iff

joohoi:
LGTM, I think this is the most elegant way to handle the situation, and leaves us with a mechanism to reuse in the future iff (pun intended) - and most likely: when we're going to need it.

bmw:
iff is sometimes used as shorthand for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if), however, using it here really doesn't do much other than making our documentation harder to understand.

Thanks for reviewing this so quickly @joohoi!

PR #3452, opened 2016-08-24
changing issue.net to system-release [needs tweaking]

shawn-sterling:
If someone has a custom issue, amazon detection will fail. This moves it to use the /etc/system-release instead.


pde:
@shawn-sterling thanks for the PR! Can you tweak it to make the change against `letsencrypt-auto-source/letsencrypt-auto.template`, and then run `letsencrypt-auto-source/build.py`, but leave the copy of `certbot-auto` in the root unchanged? It will be updated automatically during the next release.

bmw:
It's been over two months and we haven't seen any more progress on this. Closing this in favor of cleaning up our PR queue, but please open another PR if you address review comments and we'll take another look.

Issue #2558, opened 2016-02-27
Don't add extra ServerName/ServerAlias directives


bmw:
With #2537, we correctly detect wildcard `ServerAlias` directives when choosing a vhost, however, we still add a specific `ServerName/ServerAlias` even though it's not necessary.

For example if we have:

```
ServerAlias *.example.com
```

After we install the certifiicate we'll have:

```
ServerAlias *.example.com
ServerAlias www.example.com
```

We shouldn't do this as the change is unnecessary and (slightly) complicates the resulting vhost.

PR #3457, opened 2016-08-26
Rationalise challenge and port selection flags #3343 [see #3466])

SwartzCr:
resolves issue #3343


pde:
It isn't totally obvious how this fixes an Apache bug, but I guess it'll get reviewed as part of #3466...

SwartzCr:
Oh hmm this totally doesn't fix that bug, not sure why it has that title
What it does do is allow you to choose which challenge you want outside of plugin specific challenge logic. Though brad and I opted to leave the standalone parser in there but to print a deprecation warning

SwartzCr:
Ahh I see what happened, I pushed this from a different (not yet) in progress branch and brad copied the issue heading. I still have to do #2558 - but this doesn't fix it

PR #3458, opened 2016-08-26
Support both invalidEmail and invalidContact errors

rolandshoemaker:
With the switch to the proper ACME error coming sometime in the future (letsencrypt/boulder#2140) Certbot should support both types in the interim.

### DIFF
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -27,6 +27,8 @@ class Error(jose.JSONObjectWithFields, errors.Error):
             ('dnssec', 'The server could not validate a DNSSEC signed domain'),
             ('invalidEmail',
              'The provided email for a registration was invalid'),
+            ('invalidContact',
+             'The provided email for a registration was invalid'),
             ('malformed', 'The request message was malformed'),
             ('rateLimited', 'There were too many requests of a given type'),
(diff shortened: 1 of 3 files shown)


acme/acme/messages.py
```diff
@@ -27,6 +27,8 @@ class Error(jose.JSONObjectWithFields, errors.Error):
             ('dnssec', 'The server could not validate a DNSSEC signed domain'),
             ('invalidEmail',
              'The provided email for a registration was invalid'),
+            ('invalidContact',
+             'The provided email for a registration was invalid'),
```
bmw: nit: perhaps the description here should be more general to match the error type. Kuba wanted to keep these error messages in sync with the ACME, but that was a bit contentious (see #1785). With this in mind I'd like to see this being something like either:
```
The contact URI for a registration was invalid
```
or
```
The provided contact information for a registration was invalid
```
I have a slight preference for the former as just including the string from the ACME spec is in line with most of error messages here, but if you prefer the latter, that's fine with me.

bmw:
Other than my minor nit, this LGTM. Thanks for the PR @rolandshoemaker.

bmw:

PR #3463, opened 2016-08-28
Fixed hash_bucket_size detection for nginx

Gordin:
bucket_directive[0] is '\n', bucket_directive[1] is 'server_names_hash_bucket_size'

The code shouldn't search for \n when trying to find the bucket_size option.


pde:
Thanks for catching this. Perhaps we should have wrapped those lists in `nginxparser.UnspacedList()` constructors instead, but this should work.

PR #3466, opened 2016-08-30
DNS challenge support in the manual plugin and general purpose --preferred-challenges flag

### LINKED ISSUE #2558: Don't add extra ServerName/ServerAlias directives

### LINKED ISSUE #3343: Rationalise challenge and port selection flags

bmw:
Builds on and brings in #3379, #3439, and #3457. Fixes #3343.

This PR adds a `--preferred-challenges` flag to the Certbot core instead of having near identical flags in each plugin that supports multiple challenge types. The `--http-01-port` and `--tls-sni-01-port` flags are cleaned up to make them more user friendly. Mostly due to the work of @mlhamel, this PR also adds support for the `dns-01` challenge to manual plugin.

#### -01 suffix
We currently require the user to specify `-01` on every challenge type. This is true for selection both ports and challenge type. Is this the behavior we want?

I think this is fairly tricky and we should think about it now before introducing the `--preferred-challenges` flag and making the port flags more visible. Allowing the user to omit the version of the challenge they want may simplify Certbot upgrading to new challenge types in the future. On the other hand, it's possible an ACME CA supports multiple types of the same challenge and the user has a preference. In fact, a different version of the challenge may not work on a user's system. For example, a change in the path used for the HTTP challenge may cause webroot not to work if their web server is not configured to serve files out of a different (hidden) directory.

While it's ugly and verbose, I think I'm in favor of keeping the `-01` suffix. We could perhaps write special logic that allows the user to also specify `--preferred-challenges http,tls-sni,dns` which would select either all or the most recent challenge of each type. This may not be worth the effort and could result in confusing behavior for users who don't understand the difference in between specifying challenges with and without the version number

I'd like to hear other people's opinions here.

#### DNS self verification

One of the changes I made from @mlhamel's PR is a softer error message when `dnspython` is not available. Due to the way the dependency is specified in `acme/setup.py`, I suspect most Certbot installations will not have this package. If we want to make an additional change here, we have a few options:

1. Make `dnspython` a required dependency in the `certbot` package. This is nicer than making it in required in `acme` as it keeps the library more flexible.
2. Remove the self verification entirely.
3. Remove the warning when `dnspython` is not available.

I don't have a strong preference here, but I think I prefer any the three options above over the approach in this PR. I appreciate any feedback.


certbot/plugins/manual.py
```diff
@@ -97,21 +107,29 @@ def prepare(self):  # pylint: disable=missing-docstring,no-self-use
...
+                "an HTTP server. Alternatively, instructions are shown on how "
+                "to use Python's built-in HTTP server. The user is "
+                "responsible for configuration of a domain's DNS when solving "
+                "dns-01 challenges. The type of challenges used can be "
+                "controlled through the --preferred-challenges flag.")
 
     def get_chall_pref(self, domain):
         # pylint: disable=missing-docstring,no-self-use,unused-argument
-        return [challenges.HTTP01]
+        return [challenges.HTTP01, challenges.DNS01]
```
mlhamel: Sorry my bad. I've forgot it was the list of possible values :)

certbot/auth_handler.py
```diff
@@ -33,14 +33,18 @@ class AuthHandler(object):
         and values are :class:`acme.messages.AuthorizationResource`
     :ivar list achalls: DV challenges in the form of
```
mlhamel: I'm wondering if it is still true cause `achalls` doesn't seems to be a parameter anymore.

certbot/auth_handler.py
```diff
@@ -33,14 +33,18 @@ class AuthHandler(object):
...
         :class:`certbot.achallenges.AnnotatedChallenge`
+    :ivar list pref_challs: sorted user specified preferred challenges
+        in the form of subclasses of :class:`acme.challenges.Challenge`
+        with the most preferred challenge listed first
 
     """
-    def __init__(self, auth, acme, account):
+    def __init__(self, auth, acme, account, pref_challs):
```
mlhamel: maybe we could only used `challs` or simply `challenges` in place of `pref_challs`?

mlhamel:
I really your changes 👍 That's amazing !!

I just had a few questions, firstly regarding the code, a bit a everywhere you are using `prefered` `challenges`, why not use simply `challenges`?

Then also, regarding the different options

Do you mean removing the dns challenge, as if it didn't exists when `dnspython` is missing?

> Remove the self verification entirely.

I'm not sure I understand what you mean when saying:

> Remove the warning when dnspython is not available.

hobarrera:
I tried using this branch to obtain a certificate, but it crashed when writing the certificate:

```
2016-08-31 17:42:22,237:DEBUG:certbot.main:Exiting abnormally:
  File "/home/hugo/.local/share/virtualenvs/certbot/bin/certbot", line 11, in <module>
    load_entry_point('certbot==0.9.0.dev0', 'console_scripts', 'certbot')()
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/main.py", line 770, in main
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/main.py", line 565, in obtain_cert
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/main.py", line 104, in _auth_from_domains
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/client.py", line 281, in obtain_and_enroll_certificate
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/client.py", line 260, in obtain_certificate
    self.config.rsa_key_size, self.config.key_dir)
  File "/home/hugo/.local/share/virtualenvs/certbot/lib/python3.5/site-packages/certbot-0.9.0.dev0-py3.5.egg/certbot/crypto_util.py", line 58, in init_save_key
    key_f.write(key_pem)
TypeError: write() argument must be str, not bytes
```

mlhamel:
@hobarrera which command/test have you ran?

hobarrera:
> certbot certonly --preferred-challenges dns-01 --authenticator manual --domain hugo.barrera.io --manual-public-ip-logging-ok

I then set the DNS record appropriately, and checked it before confirming.

asfaltboy:
tl;dr; [python 3.5 is not supported][2], try running with python 2.7.

A quick exception dissection: looks like the exception was raised in `init_save_key` that calls [`make_key`][1], which is supposed to return a string but, in this case, returns a byte string. We then try to write this byte-string into a file object opened with 'w', which means it expects an encoded string (e.g unicode), and fails as expected.

[1]: https://github.com/certbot/certbot/blob/master/certbot/crypto_util.py#L207
[2]: https://github.com/certbot/certbot#system-requirements

SwartzCr:
I didn't see any glaring errors when looking through it

certbot/auth_handler.py
bmw: Yeah it's still true. `ivar` stands for instance variable and `achalls` is still a variable of the `AuthHandler` class.

bmw:
@mlhamel,

> I just had a few questions, firstly regarding the code, a bit a everywhere you are using prefered challenges, why not use simply challenges?

I personally like preferred challenges as the name because it's more descriptive. If you look through our code, there are all types of challenges being passed around. There's challenge bodies (challbs), challenges (challs), annotated challenges (achalls), etc. With all of this going on, I think it's worth the few extra characters in the variable name to better describe what that list of challenges is.

> Do you mean removing the dns challenge, as if it didn't exists when dnspython is missing?

No. When I say remove the self verification, I'm referring to removing the call to `simple_verify` for the DNS challenge. All this function does is attempt to check client side that the challenge is set up correctly. It is not a necessary check. It can be useful as it can report problems earlier to the user, however, if it's going to fail for most people due to the missing `dnspython` dependency, I question if it's worth the hassle.

> I'm not sure I understand what you mean when saying:
Remove the warning when dnspython is not available.

Currently, we call `logger.warning` with the following message when `dnspython` is not available:
"Self verification requires optional dependency `dnspython` to be installed."

Due to other pieces of the project, this message is shown in bright red on the user's terminal. Since this message will confuse/worry many users, I'm suggesting we remove this message.

@hobarrera,

This Python 3 error seems unrelated to this PR. As @asfaltboy points out, Python 3 is not yet supported by Certbot (although it is very close!). Feel free to create a new bug report/submit a PR.

bmw:
@pde, Noah wanted you to take a look over this PR as he wasn't familiar with everything. I'd love your feedback on my questions in the original post.

mlhamel:
@bmw then I have nothing else to say. Good job :)

I would be in for keeping this PR as is and not change anything regarding `DNS self verification` and this might be done in a separated PR. I fear if we are still waiting this will become out-of-date....

certbot/auth_handler.py
pde: I presume you were suggesting this because the combination of `pref_challs` and `chall_prefs` is confusing. I agree, though there are several call sites that pass in `pref_challs` from elsewhere in the codebase, and it's probably clearer to have an indication of semantics at those call sites.

Maybe we should rename `chall_prefs`? Or kick this to a refactor ticket?

bmw:
@jledford-devops, there's not a specific date no. Unfortunately, the best I can say at this point is it will be released soon. It should be out sometime within the next 3 weeks or so.

If you want to try the DNS challenge in the meantime, depending on your setup and your willingness to take risk by running bleeding edge software, you can run the client from git following our [developer instructions](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client) and checkout this branch.

pde:
OK, LGTM. If this is green @bmw you should check the last few commits I pushed, then hit merge!

mlhamel:
@bmw is there anything else to work on to have something top notch and 100% nice? This is currently looking great.

bmw:
Fixed a few nits I had around the `--preferred-challenges` flag. Also, it doesn't appear that https://certbot.eff.org/docs/using.html#plugins works. I believe this is simply due to the lack of a website update.

@SwartzCr, can you push an update to certbot.eff.org when you get a chance?

pde:
Looks great. Merging!

Issue #3468, opened 2016-08-31
Update errors to match latest ACME version.


cowlicks:
This is the sister of [le/boulder#2068](https://github.com/letsencrypt/boulder/issues/2068).

In the ACME 03-draft errors change from `use urn:acme:error:` to `urn:ietf:params:acme:error:`.

I'm working on a pr for this.

TheNavigat:
I think this issue should be closed, @bmw

bmw:
Thanks @TheNavigat.

Fixed in #3469.

PR #3469, opened 2016-08-31
Update ACME error namespace to match the new draft.

cowlicks:
Previously ACME errors were prefixed with `urn:acme:error:`, now they are prefixed with `urn:ietf:params:acme:error:`. More about this [here](https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#errors).

This is the sister of a PR in boulder [here](https://github.com/letsencrypt/boulder/pull/2147).

This PR shouldn't break existing code. I have a few questions below

### DIFF
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -30,7 +30,7 @@ ERROR_CODES = {
 ERROR_TYPE_DESCRIPTIONS = dict(
     (ERROR_PREFIX + name, desc) for name, desc in ERROR_CODES.items())
 
-ERROR_TYPE_DESCRIPTIONS.update(dict(  # add errors with old prefix, depricate me
+ERROR_TYPE_DESCRIPTIONS.update(dict(  # add errors with old prefix, deprecate me
     (OLD_ERROR_PREFIX + name, desc) for name, desc in ERROR_CODES.items()))
 
 
@@ -62,7 +62,7 @@ class Error(jose.JSONObjectWithFields, errors.Error):
 
         """
         if code not in ERROR_CODES:
-            raise ValueError("The supplied code: %s is not an ACME error"
+            raise ValueError("The supplied code: %s is not a known ACME error"
                              " code" % code)
         typ = ERROR_PREFIX + code
         return cls(typ=typ, **kwargs)
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -57,8 +57,8 @@ class Error(jose.JSONObjectWithFields, errors.Error):
     def with_code(cls, code, **kwargs):
         """Create an Error instance with an ACME Error code.
 
-        :??? unicode code:
-        :??? kwargs:
+        :unicode code: An ACME error code, like 'dnssec'.
+        :kwargs: kwargs to pass to Error.
 
         """
         if code not in ERROR_CODES:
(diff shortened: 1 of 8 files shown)


certbot/tests/cli_test.py
```diff
@@ -908,7 +908,7 @@ def test_handle_exception(self, mock_sys):
             mock_sys.exit.assert_any_call(''.join(
                 traceback.format_exception_only(errors.Error, error)))
 
-        exception = messages.Error(detail='alpha', typ='urn:acme:error:triffid',
+        exception = messages.Error(detail='alpha', typ='urn:acme:error:triffid',  # triffid ?????????????????????????????????????
```
cowlicks: Not sure what's going on here.

acme/acme/messages.py
```diff
@@ -18,31 +46,23 @@ class Error(jose.JSONObjectWithFields, errors.Error):
...
-
     typ = jose.Field('type', omitempty=True, default='about:blank')
     title = jose.Field('title', omitempty=True)
     detail = jose.Field('detail', omitempty=True)
 
+    @classmethod
+    def with_code(cls, code, **kwargs):
+        """Create an Error instance with an ACME Error code.
+
+        :??? unicode code:
```
cowlicks: What goes here? And for kwargs? Idk what the `ivar` in the other places means.

cowlicks:
@kuba it looks like you might be the most familiar with this code.

pde:
Thanks @cowlicks! Most of the Certbot team has been on vacation for the last couple of weeks, but we should be back on deck and able to review this next week.

cowlicks:
Closes #3468

acme/acme/messages.py
```diff
@@ -18,31 +49,24 @@ class Error(jose.JSONObjectWithFields, errors.Error):
...
+    @classmethod
+    def with_code(cls, code, **kwargs):
+        """Create an Error instance with an ACME Error code.
+
+        :unicode code: An ACME error code, like 'dnssec'.
+        :kwargs: kwargs to pass to Error.
+
+        """
+        if code not in ERROR_CODES:
+            raise ValueError("The supplied code: %s is not an ACME error"
```
pde: Perhaps "not a known ACME error code", since sometimes people will be running an old Certbot in a newer universe.

acme/acme/messages.py
```diff
@@ -7,6 +7,37 @@
...
+    'serverInternal': 'The server experienced an internal error',
+    'tls': 'The server experienced a TLS error during domain verification',
+    'unauthorized': 'The client lacks sufficient authorization',
+    'unknownHost': 'The server could not resolve a domain name',
+}
+
+ERROR_TYPE_DESCRIPTIONS = dict(
+    (ERROR_PREFIX + name, desc) for name, desc in ERROR_CODES.items())
+
+ERROR_TYPE_DESCRIPTIONS.update(dict(  # add errors with old prefix, depricate me
```
pde: "deprecate" :)

bmw:
I have not tested or reviewed this PR, but I'm happy to do so if you'd like. Quickly skimming through, I think it'll be OK. This PR seems to cause `acme.messages.Error` to support both prefixes as well as add some convenience methods. Support for both prefixes seems correct and is a fairly minor change. Based on what I've seen, I don't expect this PR to cause any breakage.

acme/acme/messages.py
cowlicks: whoops thanks you

acme/acme/messages.py
cowlicks: got it

cowlicks:
@bmw @pde indeed this works with both old and new errors. So it shouldn't break anything.

PR #3475, opened 2016-09-02
When getopts is called multiple time we need to reset OPTIND.

tcottier-enalean:
Not resetting OPTIND between each call of getopts skips all short args except the first one.

It fixes this automated command: 
`./certbot-auto certonly --webroot -w /tmp -d example.com  --agree-tos --email contact@example.com -n`

Where "-w"  was parsed by getopts and not "-n"

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -58,6 +58,7 @@ for arg in "$@" ; do
     --verbose)
       VERBOSE=1;;
     -[!-]*)
+      OPTIND=1
       while getopts ":hnvq" short_arg $arg; do
         case "$short_arg" in
           h)
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -58,7 +58,6 @@ for arg in "$@" ; do
     --verbose)
       VERBOSE=1;;
     -[!-]*)
-      OPTIND=1
       while getopts ":hnvq" short_arg $arg; do
         case "$short_arg" in
           h)
--- a/certbot-auto
+++ b/certbot-auto
@@ -55,6 +55,7 @@ for arg in "$@" ; do
     --verbose)
       VERBOSE=1;;
     -[!-]*)
+      OPTIND=1
       while getopts ":hnv" short_arg $arg; do
         case "$short_arg" in
           h)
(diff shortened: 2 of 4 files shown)


tcottier-enalean:
Some help would be appreciated on how to modify [certbot|letsencrypt]-auto without breaking the ci :)

erikrose:
IIRC, the root-level letsencrypt-auto is deprecated and should never change; it exists only to keep some ancient instructions for early adopters working. If you revert that one file, things might clear up.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -58,6 +58,7 @@ for arg in "$@" ; do
     --verbose)
       VERBOSE=1;;
     -[!-]*)
+      OPTIND=1
```
erikrose: Now you're changing only the template; you should also be committing the built version. (Run build.py.)

pde: Probably it's survivable to just change the template since a rebuild will happen automatically at release, but best practice is to do as Erik suggests.

We may need to expand the comment at the top of certbot-auto to explain this more extensively :)

tcottier-enalean:
Did anything changed between today and yesterday? 
I've just tested locally on a vagrant box and I have the same error on my previous commit (that the CI validated).

erikrose:
Yikes! IHNI what that is, but I don't think it's your fault.

vietnamxinhdep:
i don't understand the error message here. is this bug now fixed?

erikrose:
@pde Can you try restarting the Travis job? I "can't", according to Travis.

bmw:
Sorry about the Travis failures. They were unrelated to this PR and have been fixed. Rerunning your Travis build.

vietnamxinhdep:
woohoo, i am so happy that it finally got overtaken! i will test it this weekend on a freshly installed linux box and will let you guys know the result afterwards! thank you mister tcottier and the rest!

vietnamxinhdep:
ok, i still have the same problem. --quiet does work as it did before but the behaviour on apt-get is still the same. if a package is missing, it gets installed by certbot-auto but apt-get is not launched with either the -qq flag (recommended, because --quiet) nor with the -y flag and thus needs ("y" + enter) manual input. for me it would even suffice to just add -y (which would violate quieteness) but at least it doesn't need user input anymore. i consider user interaction with the quiet parameter quite dangerous since it can block a whole systems, waiting for completion of this task. could anyone please add the -qq/-y flag to possible apt-get(s) please? thank you very very much.

PR #3496, opened 2016-09-14
Fixes cffi errors in Travis during oldest tests

bmw:


bmw:
Self merging as this only affects tests and fixes our repo.

PR #3498, opened 2016-09-15
Server alias [revision requested]

### LINKED ISSUE #2558: Don't add extra ServerName/ServerAlias directives

SwartzCr:
fixes #2558

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1012,6 +1012,19 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             self.parser.find_dir("ServerAlias", target_name,
                                  start=vh_path, exclude=False)):
             return
+        # TODO find all wildcards
+        names = self.parser.find_dir("ServerName|ServerAlias", start=vh_path, exclude=False)
+        potential_wildcards = [server for server in names if server.startswith("*")]
+        # TODO reverse search through them to see if the target name is covered
+        target_split = target_name.split(".")[::-1]
+        matches = []
+        for wildcard in potential_wildcards:
+            for idx, part in enumerate(wildcard.split(".")[:1:-1]):
+                if target_split[idx] != part:
+                    break
+            matches.append(wildcard)
+        if wildcard:
+            return
         if not self.parser.find_dir("ServerName", None,
                                     start=vh_path, exclude=False):
             self.parser.add_dir(vh_path, "ServerName", target_name)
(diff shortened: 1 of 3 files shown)


certbot-apache/certbot_apache/configurator.py
```diff
@@ -1012,13 +1012,34 @@ def _add_servername_alias(self, target_name, vhost):
...
                                     start=vh_path, exclude=False):
             self.parser.add_dir(vh_path, "ServerName", target_name)
         else:
             self.parser.add_dir(vh_path, "ServerAlias", target_name)
         self._add_servernames(vhost)
 
+    def _find_matching_wildcards(self, vh_path, target_name):
+        # find all wildcards
+        names = [self.aug.get(card) for card
+                 in self.parser.find_dir("ServerName", start=vh_path, exclude=False)]
```
bmw: It appears that `ServerName` directives can't contain wildcards.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1012,13 +1012,34 @@ def _add_servername_alias(self, target_name, vhost):
...
+        aliases = [self.aug.get(card) for card
+                   in self.parser.find_dir("ServerAlias", start=vh_path, exclude=False)]
+        potential_wildcards = [server for server in names if server.startswith("*")]
+        potential_wildcards.extend([server for server in aliases if server.startswith("*")])
+        # reverse search through them to see if the target name is covered
+        target_split = target_name.split(".")[::-1]
+        matches = []
+        for wildcard in potential_wildcards:
+            for idx, part in enumerate(wildcard.split(".")[:0:-1]):
```
bmw: I think we can just do something like:
```
if target_name.endswith(wildcard[1:]):
    matches.append(wildcard)
```

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1012,13 +1012,34 @@ def _add_servername_alias(self, target_name, vhost):
...
+                if target_split[idx] != part:
+                    break
```
bmw: This only breaks out of the inner `for` loop, meaning that we always append to `matches` if a wildcard exists, even if it doesn't match.

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -1293,5 +1293,10 @@ def test_get_vhost_continue(self, mock_vhost):
         vhs = self.config.get_virtual_hosts()
         self.assertEqual([], vhs)
 
+    def test_wildcard_alias(self):
+        target_name = "banana.com"
+        chosen_vhost = self.config.choose_vhost(target_name)
+        self.assertTrue("*.banana.com" in chosen_vhost.aliases)
```
bmw: This only tests that we didn't delete the wildcard `ServerAlias`. Really we should be testing that we didn't add one.

bmw:
I built on this in #3917. Closing this in favor of that PR.

PR #3504, opened 2016-09-16
Improve error message for "no installer plugin."

### LINKED ISSUE #3321: "No installers present" message is confusing.
```
No installers seem to be present and working on your system; fix that or try running certbot with the "certonly" command
```

This is one of the most common messages, received when running certbot on an Nginx installation, but it is confusing to a new user. What's an installer? Is it my mistake or certbot's that there are no installers? What does "certonly" do?

Instead how about:

```
Certbot doesn't know how to automatically configure the web server on this system. However, it can still get a certificate for you. Please run "certbot[-auto] certonly" to do so. You'll need to manually configure your web server to use the resulting certificate.
```

jsha:
Fixes #3321.

### DIFF
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -261,9 +261,10 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
                    "your existing configuration.\nThe error was: {1!r}"
                    .format(requested, plugins[requested].problem))
     elif cfg_type == "installer":
-        msg = ('No installer plugins seem to be present and working on your system; '
-               'fix that or try running certbot with the "certonly" command to obtain'
-               ' a certificate you can install manually')
+        msg = ('Certbot doesn\'t know how to automatically configure the web '
+          'server on this system. However, it can still get a certificate for '
+          'you. Please run "certbot[-auto] certonly" to do so. You\'ll need to '
+          'manually configure your web server to use the resulting certificate.')
     else:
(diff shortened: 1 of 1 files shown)


certbot/plugins/selection.py
```diff
@@ -261,9 +261,10 @@ def diagnose_configurator_problem(cfg_type, requested, plugins):
                    "your existing configuration.\nThe error was: {1!r}"
                    .format(requested, plugins[requested].problem))
     elif cfg_type == "installer":
-        msg = ('No installer plugins seem to be present and working on your system; '
-               'fix that or try running certbot with the "certonly" command to obtain'
-               ' a certificate you can install manually')
+        msg = ('Certbot doesn\'t know how to automatically configure the web '
+          'server on this system. However, it can still get a certificate for '
+          'you. Please run "certbot[-auto] certonly" to do so. You\'ll need to '
```
bmw: You should use `certbot.cli.cli_command` instead of "certbot[-auto]" here which is set to the name used to invoke the program. The only ugly caveat with this is because `certbot/cli.py` also includes this file, you'll need to include it in this function rather than at the top of the file to avoid cyclic imports. You can include it wherever you think is prettiest. You can find another example of this [here](https://github.com/certbot/certbot/blob/master/certbot/plugins/selection.py#L168).

jsha: Done.

bmw:
LGTM

Thanks @jsha.

Issue #3472, opened 2016-09-02
Seperate the virtual environment creation step and the actual task


timdream:
We got `--os-packages-only` and I recommend we create an option `--venv-only`.

There are times it make sense to create venv first without doing anything, for example, get the environment to set up in automation so that human can do things quickly later.

Good idea?

timdream:
I am refer to `certbot-auto` here so please add that label for me.

bmw:
That might be a good idea. `certbot-auto` is simply meant to be a temporary workaround for OS packaging while we wait for it to catch up. For this reason, I personally try to avoid adding unnecessary complexity or features to it because it will be going away.

While it's certainly less elegant, people can simply run something like `certbot-auto --version` for this purpose. I'm inclined not to add a `--venv-only` flag, but I could be convinced.

timdream:
> While it's certainly less elegant, people can simply run something like certbot-auto --version for this purpose. I'm inclined not to add a --venv-only flag, but I could be convinced.

@bmw What if I propose a doc change `--version` instead? I was reading the code and the doc for quite some time but didn't figure out `--version` could do what you said.

bmw:
I think that's fine. To be clear, the main purpose of `--version` is to print the current version of Certbot, however, it will also work for your purposes because Certbot will simply print its version string and exit.

timdream:
#3512 proposed.

bmw:
I've actually changed my mind about this. We should provide a flag like `--install` that will install Certbot without running it. This flag should never be passed to Certbot as `certbot-auto` should exit before calling into Certbot.

Providing this feature helps cases like #3691.

timdream:
@bmw Should I close #3529 then? I might not be able to produce a patch for `--install` soon-ish.

bmw:
@timdream, yes we should probably close that PR. Sorry I lead you down the wrong path and we didn't end up taking the PR. I hope to have this feature soon though.

bmw:
This was done with the flag `--install-only`.

PR #3512, opened 2016-09-17
Mention the Python venv in cli-help doc for --version argument

timdream:


bmw:
For context, see #3472 where I argued against adding a flag for this purpose.

We could add a flag or we could word it differently in the `certbot-auto` help or documentation. If we want to do it in the flag's documentation, I think we should only do so for `certbot-auto` and say something like:
```
--version                                 install certbot and print version
````

timdream:
@bmw You are right, this only make sense for `certbot-auto` and not `certbot`. For the purpose of record keeping, let me submit another pull request and do what you said.

Issue #3355, opened 2016-07-30
Warning message for low memory servers 


benileo:
I've ran into it a couple times where I'm installing a certificate on a tiny server. In my case it was Ubuntu 14.04, which still does not have the letsencrypt package (not sure if that is coming). 

It would be nice if `certbot-auto` printed a warning message before trying to install the Python packages (in this case it is cryptography that is failing to build). 

The message could be something like:
`Warning your system only contains 0.5 GB of memory which is less the recommended 2.0 GB, you may no be able to build the proper dependencies needed`. 

Just an idea..

pde:
This is a good idea!

surya-v:
Implemented the suggested check.

pde:
Fixed by #3473

PR #3473, opened 2016-09-02
certbot-auto: Print link to doc on debugging pip install error [revision requested]

timdream:
Also, update the doc to teach the user to workaround problem on a low
memory system.

This is a document- and log-only fix.

With instruction here and #3472, we would be able to allow the user to
create automation scripts to do that exact workaround.

PR #3513, opened 2016-09-19
#3355: Updated certbot-auto source files to check for free memory

surya-v:
-> Ran "tox"  : Result Successful


benileo:
would a better option be to create enough swap space so that `cryptography` can be installed? we are running as root, but maybe that is too risky.

surya-v:
i am not sure if clearing swap space would be a good idea or not ? need more insights on this.

pde:
Fixes: #3355 Obsoletes? #2370

pde:
Sorry, was confused here but this should be fixed by #3473

Issue #3510, opened 2016-09-17
Screen on GUI should indicate how to deselect options


ohemorange:
Specifically, the "Which names would you like to activate HTTPS for?" screen should have a note to the effect of "Use spacebar to select and deselect options".

bmw:
With `dialog` being removed, this can be closed.

PR #3519, opened 2016-09-20
#3510: Updating GUI with a note to use spacebar for select/deselect

surya-v:


certbot/display/ops.py
```diff
@@ -152,7 +152,8 @@ def _filter_names(names):
 
     """
     code, names = z_util(interfaces.IDisplay).checklist(
-        "Which names would you like to activate HTTPS for?",
+        "Which names would you like to activate HTTPS for? "
+        "(Use spacebar to select/deselect options)",
```
pde: You should only add this extra text if the `IDisplay` is an instance of `display_util.NcursesDisplay`.

Something like this:
```
import 
msg = "Which names..."
disp = z_util(interfaces.IDisplay)
if isinstance(disp, display_util.NcursesDisplay)
    msg + = " (Use spacebar..."
disp.checklist(msg, ..)
```
To see why, try running certbot with `-t` :).

bmw:
Thanks for working on this, however, we've recently completely removed `dialog` from the project so this PR is no longer necessary. Thanks for taking the time to write it up though!

Issue #3147, opened 2016-06-10
Hook scripts directory


ebourg:
Hi,

I'd like to suggest the addition of a directory for storing hook scripts executed after a certificate renewal (for example `/etc/letsencrypt/hooks.d`). This would allow third party packages to easily integrate with certbot without overloading the `--post-hook` parameter in the cron file. For example a tomcat package could install a script in this directory to create a PKCS12 keystore in `/etc/tomcat` from the pem files and restart the Tomcat connector after each renewal.

came88:
Related to the discussion in #3394, these hooks should be global or per lineage?

ebourg:
Per lineage hooks might be a good idea, but the hook I had in mind was a global one (with an extra parameter indicating which certificate was renewed).

pde:
Related: #2782

pde:
(Giving this to @joohoi since he's also working on the script plugin)

edwardotis:
Hi @joohoi, any update on this hook scripts feature?

joohoi:
Hi, the functionality was split off from the script plugin in the process after all. I believe @oerdnj is working on this issue currently.

edwardotis:
Gotcha, thanks. @oerdnj, any news?

oerdnj:
@edwardotis I mentioned I might look into this if I find some free time. No continuous free time has popped up yet. So if there are any takers, feel free to implement this before me.

edwardotis:
@oerdnj thanks for the update

joohoi:
This issue will be tracked in #3868 , it has been assigned to milestone for v0.11

bmw:
Our packaging maintainers are continuing to get requests about this feature. To prevent them from caving and making a change like this themselves, I promised we'd ship this in 0.18.0 or 0.19.0.

Putting this in 0.18.0 for now to try and land it sooner rather than later.

bmw:
In Certbot's next minor release, we are planning on adding hook directories to Certbot to allow users and package maintainers to add scripts to be run during certificate renewal by simply placing an executable file in the appropriate directory. We wanted to post the proposed design of this system publicly to get feedback on it and to make sure it covers the majority of the desired use cases. If you are not already familiar with hooks in Certbot, see the documentation about the --pre-hook, --deploy-hook, and --post-hook flags found at https://certbot.eff.org/docs/using.html#renewing-certificates.

[...]

PR #3521, opened 2016-09-21
Script plugin

joohoi:
Script plugin, ready for review. I left the original description below for context on later comments.

Features:
- Supports HTTP-01 and DNS-01 challenges.
- Enables user to provide path in command line for script to use in authentication and another to use in cleanup after the authentication.
- Sets values needed by the user script to environment variables (more on this below)

New cli parameters:
- -script-auth=/path/to/authentication_script.sh
- -script-cleanup=/path/to/cleanup_script.sh

Also uses --preferred-challenges to define which challenge to answer to. (Plugin defaults to HTTP-01 if not defined)

Environment variables exposed to different scripts:
### Authenticator script
- CERTBOT_DOMAIN: The domain being authenticated
- CERTBOT_VALIDATION: The validation string
- CERTBOT_TOKEN: Resource name part of HTTP-01 challenge (HTTP-01 only)

Additionally for cleanup:
- CERTBOT_AUTH_OUTPUT: Whatever auth script wrote to stdout

~~The plugin uses script bundles, which reside in /etc/letsencrypt/script.d/$BUNDLENAME~~

~~The bundle can have the following files:~~
- ~~pre.sh  - Script to execute before attempting the challenges~~
- ~~authenticator.sh - Script to handle the authentication~~
- ~~cleanup.sh - Script to execute after authentication~~
- ~~config.sh - Configuration file for static configuration variables (eg. which challenge the script bundle answers to)~~

~~The different scripts have different environment variables they are able to access:~~
- ~~pre.sh - CERTBOT_DOMAIN~~
- ~~authenticator.sh - CERTBOT_DOMAIN, CERTBOT_VALIDATION, CERTBOT_TOKEN (HTTP-01 only)~~
- ~~cleanup.sh - CERTBOT_DOMAIN, CERTBOT_VALIDATION, CERTBOT_TOKEN (HTTP-01 only)~~
#### HTTP-01

Example usage:

```
certbot certonly --script                                 \
--preferred-challenges=http                               \
--script-auth=/path/to/http/authenticator.sh              \
--script-cleanup=/path/to/http/cleanup                    \
-d secure.example.com

```

/path/to/http/authenticator.sh

``` bash
#!/bin/bash
echo $CERTBOT_VALIDATION > /var/www/htdocs/.well-known/acme-challenge/$CERTBOT_TOKEN
```

/path/to/http/cleanup.sh

``` bash
#!/bin/bash
rm -f /var/www/htdocs/.well-known/acme-challenge/$CERTBOT_TOKEN
```
#### DNS-01 (Cloudflare API v4)

Example usage:

```
certbot certonly --script                                 \
--preferred-challenges=dns                                \
--script-auth=/path/to/dns/authenticator.sh               \
--script-cleanup=/path/to/dns/cleanup.sh                  \
-d secure.example.com

```

/path/to/dns/authenticator.sh

``` bash
#!/bin/bash

# Get your API key from https://www.cloudflare.com/a/account/my-account
API_KEY="your-api-key"
EMAIL="your.email@example.com"

# Strip only the top domain to get the zone id
DOMAIN=$(expr match "$CERTBOT_DOMAIN" '.*\.\(.*\..*\)')

# Get the Cloudflare zone id
ZONE_EXTRA_PARAMS="status=active&page=1&per_page=20&order=status&direction=desc&match=all"
ZONE_ID=$(curl -s -X GET "https://api.cloudflare.com/client/v4/zones?name=$DOMAIN&$ZONE_EXTRA_PARAMS" \
     -H     "X-Auth-Email: $EMAIL" \
     -H     "X-Auth-Key: $API_KEY" \
     -H     "Content-Type: application/json" | python -c "import sys,json;print(json.load(sys.stdin)['result'][0]['id'])")


# Create TXT record
CREATE_DOMAIN="_acme-challenge.$CERTBOT_DOMAIN"
RECORD_ID=$(curl -s -X POST "https://api.cloudflare.com/client/v4/zones/$ZONE_ID/dns_records" \
     -H     "X-Auth-Email: $EMAIL" \
     -H     "X-Auth-Key: $API_KEY" \
     -H     "Content-Type: application/json" \
     --data '{"type":"TXT","name":"'"$CREATE_DOMAIN"'","content":"'"$CERTBOT_VALIDATION"'","ttl":120}' \
        | python -c "import sys,json;print(json.load(sys.stdin)['result']['id'])")

# Save info for cleanup
if [ ! -d /tmp/CERTBOT_$CERTBOT_DOMAIN ];then
        mkdir -m 0700 /tmp/CERTBOT_$CERTBOT_DOMAIN
fi
echo $ZONE_ID > /tmp/CERTBOT_$CERTBOT_DOMAIN/ZONE_ID
echo $RECORD_ID > /tmp/CERTBOT_$CERTBOT_DOMAIN/RECORD_ID

# Sleep to make sure the change has time to propagate over to DNS
sleep 25
```

/path/to/dns/cleanup.sh

``` bash
#!/bin/bash

# Get your API key from https://www.cloudflare.com/a/account/my-account
API_KEY="your-api-key"
EMAIL="your.email@example.com"

if [ -f /tmp/CERTBOT_$CERTBOT_DOMAIN/ZONE_ID ]; then
        ZONE_ID=$(cat /tmp/CERTBOT_$CERTBOT_DOMAIN/ZONE_ID)
        rm -f /tmp/CERTBOT_$CERTBOT_DOMAIN/ZONE_ID
fi

if [ -f /tmp/CERTBOT_$CERTBOT_DOMAIN/RECORD_ID ]; then
        RECORD_ID=$(cat /tmp/CERTBOT_$CERTBOT_DOMAIN/RECORD_ID)
        rm -f /tmp/CERTBOT_$CERTBOT_DOMAIN/RECORD_ID
fi

# Remove the challenge TXT record from the zone
if [ -n "${ZONE_ID}" ]; then
    if [ -n "${RECORD_ID}" ]; then
        curl -s -X DELETE "https://api.cloudflare.com/client/v4/zones/$ZONE_ID/dns_records/$RECORD_ID" \
                -H "X-Auth-Email: $EMAIL" \
                -H "X-Auth-Key: $API_KEY" \
                -H "Content-Type: application/json"
    fi
fi
```


joohoi:
Update:
- Supports now DNS-01 challenge also
- Renamed posthook to cleanup to avoid mixup
- Removed prehook (do we actually need an preflight check?)
- DNS-01 challenge using Cloudflare API v4 tested (and works), using crude hacked-together bash script

TODO:
- Make it possible for user to select between using script bundle (the current functionality, scripts living under /etc/letsencrypt/script.d/) or a single user defined authentication / cleanup script residing anywhere on the filesystem.
- Tests

joohoi:
Comments on this pile of random thoughts and rambling @pde @bmw @jsha ?

About the design and how to continue from here:
- Trashing the concept of script bundles and introducing cli parameters for auth script, cleanup script and using the global preferred-challenges to define the challenge to use.
-  Possibly refactoring hooks depending of the outcome of this discussion 

I thought about the correct place for authenticator and cleanup scripts, and contrary to what we discussed on the call, I think the plugin itself is the correct place for authenticator and cleanup scripts after all (and renew hook should be moved to renewer.py). The common code these share (validation, exec, env var control etc) should be moved to a util module. 

Some of the reasoning behind this:

- If we're implementing prehooks.d and posthooks.d as per #3147 . All the hooks will be "blindly" executed regardless of the lineage / domain, while the authentication & cleanup and renew hooks are very specific to a lineage, and in case of script plugin even more so, because both auth and cleanup need to run for each domain (challenge) individually.
- The scopes for pre- and posthooks are currently on a more general level than renew hook and  even further for auth & cleanup scripts which are specific to a lineage and an individual domain in lineage respectively.
- Example of the execution order in renewal hitting two lineages using script plugin: - Prehook(s) - lineage1( auth + cleanup for each domain in lineage1) - lineage1 renew hook - lineage2( auth + cleanup for each domain in lineage2) - lineage2 renew hook - Posthook(s) -

pde:
@joohoi I presume I shouldn't review this further until you've pushed your code that's revamped to work without bundles?

joohoi:
The original description modified to reflect the latest changes.

joohoi:
Discussion on IRC regarding the state (only the relevant parts):

> 14:08 < Peng> Would it be possible for the script plugin to store a small amount of state from the auth script and pass it to the cleanup script?
> 14:08 < Peng> It's hacky that your CloudFlare example has to save state in /tmp
> 14:09 < Peng> It would be cleaner -- for the scripts, if not for certbot -- if certbot could handle the state itself.
> 14:17 < joohoi> Peng: if you have a good idea how to return a generally meaningful and abstract enough piece of state from script to certbot, I'm all ears. :)
> 14:19 < Peng> joohoi: Good question. I don't have a great idea. It could just use stdin/stdout.
> 14:22 < joohoi> Peng: passing the state by stdout back to certbot, writing it to env var to be read in cleanup could work.

srvrco:
The cloudflare script doesn't work if you don't have control of the highest level domain.   For example if you own something.co.uk whereas you don't own co.uk of course ;)     It would be possible to get a complete list of your domains from cloudflare though, and then select the "best match". ( an example of a similar script at https://github.com/srvrco/getssl/blob/master/dns_scripts/dns_add_cloudflare if you want to use some of it). 

Also, do you need to save the Zone / Record ID ?  The cleanup script could follow the same process as the initial script to find the Zone and record ID for removing it.

joohoi:
Oh, the script is meant to be a proof of concept instead of anything for people to actually use. Maybe I should make that more clear in the original description.

srvrco:
No problem at all.    I'm interested in the script plugin for certbot as that will enable a lot of things that were not possible before ( such as placing tokens on remote servers via ssh) which was the reason I wrote getssl in the first place.   I'm hopeful that in the future it can all be done with certbot and the script plugin :)

pde:
In your docs in this ticket you have `--script-auth=/path/to/http/authenticator.sh` but I think that should maybe be `--script-auth /path/to/http/authenticator.sh`?

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
...
+        :param str script_path: Path of script to check
+        :raises errors.PluginError: If script is not valid
+        :returns `boolean`: If script is valid"""
+
+        if os.path.exists(script_path) and os.path.isfile(script_path):
+            if os.access(script_path, os.X_OK):
+                return True
+            else:
+                raise errors.PluginError(
+                    "Script {0} isn't readable by Certbot".format(script_path))
```
pde: Current this error fires if a script is not executable, but it's the wrong error message.

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
+"""Standalone Authenticator."""
```
joohoi: Fixed

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
...
+@zope.interface.provider(interfaces.IPluginFactory)
+class Authenticator(common.Plugin):
+    """Script authenticator
+
+    calls user defined script to perform authentication and
+    optionally cleanup.
+
+    """
+
+    description = "Authenticate using user provided script(s)"
```
joohoi: Unfortunately more_info() can't be used, as PluginEntryPoint can't access class methods, only class variables. I implemented long_description variable, and PluginEntryPoint property with the same name, that uses long_description if available, falling back to description if not.  Also put it to use in CLI.

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
...
+            help="path to the authentication script")
+        add("cleanup", default=None, required=False,
+            help="path to the cleanup script")
+
+    @property
+    def supported_challenges(self):
+        """Challenges supported by this plugin."""
+        return self.challenges
+
+    def check_script_validity(self, script_path):
```
joohoi: Now using hooks.validate_hook, had to do a bit of refactoring to hooks to accomplish this.

certbot/plugins/script.py
joohoi: It's gone now as we're using validation from hooks.

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
...
+        if isinstance(value, bytes):
+            value = value.decode(sys.getdefaultencoding())
+        return str(value)
+
+    def _write_env(self, env_vars):
+        """Write environment variables"""
+        for k in env_vars.keys():
+            os.environ[k] = env_vars[k]
+
+    def execute(self, shell_cmd):
```
joohoi: Had to do minor refactoring to _run_hook, but this is merged now.

certbot/plugins/script.py
```diff
@@ -0,0 +1,172 @@
...
+    def _normalize_string(self, value):
+        """Return string instead of bytestring for Python3.
+        Helper function for _write_env, as os.environ needs
+        str"""
+
+        if isinstance(value, bytes):
+            value = value.decode(sys.getdefaultencoding())
+        return str(value)
+
+    def _write_env(self, env_vars):
```
joohoi: Totally! Nice catch.

bmw:
@joohoi, when using the script plugin with multiple domains `CERTBOT_AUTH_OUTPUT` only contains the output from the auth script that ran last because we overwrite it each time. Perhaps we should create a `dict` that maps achalls to auth output and set the variable to the matching output when performing cleanup. The one downside of this approach is if someone wanted to pass messages between script invocations used for different domains, it wouldn't work. Alternatively, we could just constantly append to the environment variable or remove it entirely. Thoughts?

joohoi:
Oh wow, something I missed completely. You are right. I can't figure out a use case where data had to be passed in between of auths / cleanups, but the cleanup sure should be run for each auth. I think a list of `dict` approach is good though. Appending `dict` of `CERTBOT_DOMAIN` `CERTBOT_TOKEN`, `CERTBOT_VALIDATION` and  `CERTBOT_AUTH_OUTPUT`to a list in auth, and looping it for the cleanup should be sufficient.

bmw:
I can't think of a real use case for having to pass data between scripts for different challenges. Even if there was one, it can be worked around because as you know `CERTBOT_AUTH_OUTPUT` is just a convenience.

And thanks for being explicit about needing to store all four environment variables. I actually didn't notice that when I wrote my other post.

grinapo:
As for the documentation. ;-)

That's `auth.sh` I have used for pdns:
```
#!/bin/sh
SUBDOM=$CERTBOT_DOMAIN
DOM=`echo "$SUBDOM" | cut -d. -f 2-`
RR="_acme-challenge.$SUBDOM"
TYPE='TXT'
VAL=$CERTBOT_VALIDATION
TTL=10
SQL="INSERT INTO records (domain_id,name,type,content,ttl)
        SELECT id,'$RR','$TYPE','$VAL',$TTL FROM domains WHERE name='$DOM'"
psql -h dnsserver.example.com -U pdns -p 5432 dns -c "$SQL" > /tmp/auth.sql.out 2>&1
sleep 5
```

using
```
certbot -c ~/certbot/grin.ini \
        --agree-tos  --manual-public-ip-logging-ok \
        certonly --manual --preferred-challenges=dns \
        --manual-auth-hook ~/certbot/m-auth.sh \
        --manual-cleanup-hook ~/certbot/m-clean.sh \
        -d blackhole.example.com
```

where grin.ini contains
```
config-dir      = /home/grin/work/ssl/certbot/conf
logs-dir        = /home/grin/work/ssl/certbot/log
work-dir        = /home/grin/work/ssl/certbot/work
staging = True
text = True
email = grin@example.com
expand  = True
renew-by-default = True
rsa-key-size = 4096
```

(Obviously this is only for experiments, without any proper error handling, and the login is handled through ~/.pgpass which isn't quite the state-of-the-art either. Works™.)

PR #3526, opened 2016-09-22
Remove pointless question

### LINKED ISSUE #3509: Fix Nginx prompt
Saw the following prompt when using Nginx
```
No names were found in your configuration files.
You should specify ServerNames in your config files in order to allow for
accurate installation of your certificate.
If you do use the default vhost, you may specify the name manually. Would you
like to continue?
(Y)es/(N)o: 
```
`ServerName` is Apache specific and based on #3508, we don't want to allow people to use the default vhost.

bmw:
Fixes #3509.

I feel that the best approach here is to simply remove this question. First, I thought we wanted Nginx to error out if `get_all_names` returns an empty list but that's not correct. For instance, the user could have
```
server_name *.example.org;
```
We will not suggest this domain (it won't be in the return value for `get_all_names`), but if the user manually requests `www.example.org`, the Nginx plugin will/should work.

Additionally, I feel that the question itself was a little silly. You may not have to specify `ServerName` directives in your config, for instance, the Apache plugin will do this for you now. It also discusses default vhosts which may not be relevant (see above example) and may be more detail about server configuration than is needed for most users.

If there's a problem later, we should error out with the appropriate error message, but I feel like this question doesn't add much/any value for users.

### DIFF
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -103,12 +103,8 @@ def choose_names(installer):
     names = get_valid_domains(domains)
 
     if not names:
-        server = ""
-        if hasattr(installer, "name") and installer.name:
-            server = "{0} ".format(installer.name)
-            server = server[0].upper() + server[1:] # capitalise server name
         return _choose_names_manually(
-            "No names were found in your {0}configuration files. ".format(server))
+            "No names were found in your configuration files. ")
 
     code, names = _filter_names(names)
     if code == display_util.OK and names:
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -103,7 +103,8 @@ def choose_names(installer):
     names = get_valid_domains(domains)
 
     if not names:
-        return _choose_names_manually()
+        return _choose_names_manually(
+            "No names were found in your configuration files. ")
 
     code, names = _filter_names(names)
     if code == display_util.OK and names:
@@ -146,10 +147,17 @@ def _filter_names(names):
     return code, [str(s) for s in names]
 
 
-def _choose_names_manually():
-    """Manually input names for those without an installer."""
+def _choose_names_manually(prompt_prefix=""):
+    """Manually input names for those without an installer.
 
+    :param str prompt_prefix: string to prepend to prompt for domains
+
+    :returns: list of provided names
+    :rtype: `list` of `str`
+
+    """
     code, input_ = z_util(interfaces.IDisplay).input(
+        prompt_prefix +
         "Please enter in your domain name(s) (comma and/or space separated) ",
         cli_flag="--domains")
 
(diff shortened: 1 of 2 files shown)


bmw:
I'm not necessarily against adding such a prefix, however, I'd like to know what you think the value is in doing so. An accurate prompt would be something like "No names that we feel comfortable suggesting were found in your configuration files.". This is due to #3507 and my `*.example.org` example above. (I'm not actually suggesting we use this prompt.)

Furthermore, why should the user care? You could argue the value is that they can go add the names themselves, but why should they have to? The Apache plugin can do this for you. I see no reason the Nginx plugin shouldn't do so in the future. In the meantime with Nginx, we can avoid sharing this potentially inaccurate detail and we have [this error message](https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/configurator.py#L201) (which we can improve if we so desire). (To see why this message may be inaccurate, see my `*.example.org` example above.) Given our ability to handle this case, why should we share this unnecessary detail with the user?

pde:
I think the reason to add this string is to provide a little bit more transparency and context. Is Certbot asking for names because it doesn't have a way to learn them (ie, you aren't running an installer plugin)? Or because something has gone wrong with its attempts to learn them (ie your installer plugin isn't working).

I know at least for my own purposes I'd like to know which of these states Certbot is in, though I may be an unusual kind of Certbot user :)

bmw:
I suspect you're in the 5% of Certbot users who could draw such conclusions from this prefix, but I went ahead and added it back in.

Let me know if you have any other review comments.

certbot/display/ops.py
```diff
@@ -103,18 +103,8 @@ def choose_names(installer):
...
-            "If you do use the default vhost, you may specify the name "
-            "manually. Would you like to continue?{0}".format(os.linesep),
-            default=True)
-
-        if manual:
-            return _choose_names_manually()
-        else:
-            return []
+        return _choose_names_manually(
+            "No names were found in your configuration files. ")
```
pde: This could even be the name of the installer plugin: "No names were found in your Nginx configuration files," "no names were found in your Postfix configuration files".

bmw:
Don't merge this yet. It's possible this will look ugly for 3rd party plugins. I want to test it.

bmw:
My suspicions were correct. For example, this is the output when using the plesk plugin:
```
No names were found in your letsencrypt-plesk:plesk configuration files. Please
enter in your domain name(s) (comma and/or space separated)  (Enter 'c' to
cancel):
```
`.name` is the name of the plugin, not the name of the server supported by the plugin. I vote to revert my two recent commits and just say "...your configuration files.".

pde:
Oops, I totally didn't see those comments, and coded this. Uncoding it :(

Issue #3514, opened 2016-09-19
Make parser.py: add_server_directives documentation consistent with functionality


ohemorange:
Documentation and at least one calling function assume that add_server_directives performs: "Add or replace directives in the first server block with names."

Instead, it add or replaces directives in all server blocks with names

ohemorange:
This manifests itself when there is an 80 and a 443 server block with the same name. `choose_vhost` will act as documented, but after choosing, when modifying, `add_server_directives` will attempt to modify both.

ohemorange:
Also, "first server block with names" throws out the logic of `choose_vhost`. This only makes sense if there the set of names in a server block uniquely identify a server block, which is not the case.

ohemorange:
On the other hand, if we go with the implemented behavior rather than the documented, this makes switching over to "modify all matching sslish blocks" behavior much more straightforward.

ohemorange:
This also causes a bug if there is more than one plaintextish matching server block in the initial config - it will make all of them sslish, and then add the cert information to all of them.

ohemorange:
This probably means that _get_ranked_matches is broken is multiple server blocks have the same set of names, too.

bmw:
This only causes matching server blocks in the same file to be modified correct?

bmw:
Fixed in #3528.

PR #3528, opened 2016-09-22
Restructure how Nginx parser re-finds vhosts, and disable creating new server blocks.

### LINKED ISSUE #3508: Make Nginx error out if no matching server block is found
This includes the default. We can add logic for this case later.

ohemorange:
Fixes #3508, #925, #3514.

### DIFF
--- a/certbot-nginx/certbot_nginx/obj.py
+++ b/certbot-nginx/certbot_nginx/obj.py
@@ -110,7 +110,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
 
     """
 
-    def __init__(self, filep, addrs, ssl, enabled, names, raw):
+    def __init__(self, filep, addrs, ssl, enabled, names, raw, path):
         # pylint: disable=too-many-arguments
         """Initialize a VH."""
         self.filep = filep
@@ -119,6 +119,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
         self.ssl = ssl
         self.enabled = enabled
         self.raw = raw
+        self.path = path
 
     def __str__(self):
         addr_str = ", ".join(str(addr) for addr in self.addrs)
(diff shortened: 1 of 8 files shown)


certbot-nginx/certbot_nginx/obj.py
```diff
@@ -110,7 +110,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
 
     """
 
-    def __init__(self, filep, addrs, ssl, enabled, names, raw):
+    def __init__(self, filep, addrs, ssl, enabled, names, raw, path):
```
ohemorange: Yup, that's what I intended.

Barring pathological cases of server blocks inside server blocks? I don't think that's possible though, and even then, it will only choose the top-most one, which I think we can probably be fine with.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -240,42 +241,10 @@ def filedump(self, ext='tmp', lazy=True):
...
-                return False
-
-            if len(item) > 0 and item[0] == 'server_name':
-                server_names.update(_get_servernames(item[1]))
-
-        return server_names == names
-
-    def add_server_directives(self, filename, names, directives,
+    def add_server_directives(self, vhost, directives,
                               replace):
```
ohemorange: Fixed.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -240,42 +241,10 @@ def filedump(self, ext='tmp', lazy=True):
...
-        """Add or replace directives in the first server block with names.
+        """Add or replace directives in the first server block with names,
+           and update vhost with the new directives.
```
ohemorange: Fixed.

pde:
Woohoo!

Issue #3497, opened 2016-09-14
should_autorenew ignores symlinks


bmw:
Imagine the following scenario (which happened to a user). You've successfully completed your challenges from Let's Encrypt and you're renewing your certificates. After saving all four of `cert.pem`, `privkey.pem`, `chain.pem`, and `fullchain.pem` to `/etc/letsencrypt/archive`, you hit ctrl+C. Your symlinks in `/etc/letsencrypt/live` were never updated.

When you run `certbot renew` again, Certbot will tell your certificate is not due for renewal. This is because we currently check the expiration of the newest entry in `/etc/letsencrypt/archive` that has all four of the files listed above ([source](https://github.com/certbot/certbot/blob/master/certbot/storage.py#L706)). We do not check the symlink.

There are a number of ways to fix this, but this is a pretty bad bug and should be resolved before we release 0.9.0.

pde:
I guess under the current semantics of `storage.py`, the problem is that deployment is overdue, not that renewal hasn't happened...

bmw:
My understanding is renewal has occurred and `should_autorenew` will return `False` because the updated versions of the files were written to `/etc/letsencrypt/archive`. `should_autorenew` is the only metric we currently use for whether or not to renew certificates and `should_autodeploy` is never called.

pde:
Agreed. I'm thinking about addressing this by adding a call to `should_autodeploy`. That's a little frightening, because it's never been used before and I'll need to ensure that it's sane. But that will prevent our solution from breaking the hidden `deploy_before_expiry` feature.

bmw:
I'm personally not sure if that alone will work. Depending on how long ago the user encountered the error in this PR, `should_autodeploy` could return `True` or `False`. My understanding is the default `deploy_before_expiry` value is 5 days. You could potentially fix this problem by changing the default value to be the same as `renew_before_expiry`.

Also, I could be wrong, but I think other work would have to be done to actually respect `deploy_before_expiry`. For instance, as we just discussed, we choose whether or not to renew on the basis of `should_autorenew`. During renewal we do [this](https://github.com/certbot/certbot/blob/master/certbot/renewal.py#L249) which immediately deploys the cert, regardless of the `deploy_before_expiry` value.

pde:
> You could potentially fix this problem by changing the default value to be the same as renew_before_expiry.

I'm planning to do that, yes.

pde:
[...]

Issue #3366, opened 2016-08-02
certificate symlinks not fixed up on upgrade.


charlescurley:
I installed and used letsencrypt via git. The last commit was:

commit 4d4cfb414f2847f1814cfe2949c6e76f112e4ade
Author: Brad Warren <bradmwarren@gmail.com>
Date:   Tue Apr 5 17:55:18 2016 -0700

    Release 0.5.0

This on Debian Jessie:

root@chaffee:~/letsencrypt# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.5 (jessie)
Release:	8.5
Codename:	jessie
root@chaffee:~/letsencrypt#

I was able to obtain a certificate and all went well.

At the end of July, I installed certbot from the debian repo:

root@chaffee:~/letsencrypt# pre certbot
certbot	0.8.1-2~bpo8+1		all
python-certbot	0.8.1-2~bpo8+1		all
python-certbot-apache	0.8.1-1~bpo8+1		all
root@chaffee:~/letsencrypt# 

(pre is a tool for seeing what packages I have installed.)

The certificate updated, but Apache did not serve up the new certificate. I narrowed the problem down to the symlinks in /etc/letsencrypt/live/wli.ddns.info/: they were not updated.

root@chaffee:/etc/apache2# ll /etc/letsencrypt/live/wli.ddns.info/*
lrwxrwxrwx 1 root root 37 May  3 09:30 /etc/letsencrypt/live/wli.ddns.info/cert.pem -> ../../archive/wli.ddns.info/cert2.pem
lrwxrwxrwx 1 root root 38 May  3 09:30 /etc/letsencrypt/live/wli.ddns.info/chain.pem -> ../../archive/wli.ddns.info/chain2.pem
lrwxrwxrwx 1 root root 42 May  3 09:30 /etc/letsencrypt/live/wli.ddns.info/fullchain.pem -> ../../archive/wli.ddns.info/fullchain2.pem
lrwxrwxrwx 1 root root 40 May  3 09:30 /etc/letsencrypt/live/wli.ddns.info/privkey.pem -> ../../archive/wli.ddns.info/privkey2.pem
root@chaffee:/etc/apache2# ll /etc/letsencrypt/archive/wli.ddns.info/*
-rw-r--r-- 1 root root 1793 Feb  6 10:22 /etc/letsencrypt/archive/wli.ddns.info/cert1.pem
[...]

PR #3533, opened 2016-09-23
If lineages are in an inconsistent (non-deployed) state, deploy them

pde:
Fixes:  #3497. Probably also fixes: #3366.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -166,9 +166,9 @@ def _handle_identical_cert_request(config, lineage):
 
     """
     if lineage.has_pending_deployment():
-        lineage.update_all_links_to(lineage.latest_common_version())
         logger.warn("Found a new cert /archive/ that was not linked to in /live/; "
                     "fixing and reinstalling..")
+        lineage.update_all_links_to(lineage.latest_common_version())
         return "reinstall", lineage
     if renewal.should_renew(config, lineage):
         return "renew", lineage
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -155,34 +155,39 @@ def _handle_subset_cert_request(config, domains, cert):
             "reinvoke the client.")
 
 
-def _handle_identical_cert_request(config, cert):
-    """Figure out what to do if a cert has the same names as a previously obtained one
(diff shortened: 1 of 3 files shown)


pde:
Still needs tests...

pde:
Obviously if we support `deploy_before_expiry` in the future this will need to get slightly more complicated, but I think that would just be an extra test in this spot.

certbot/main.py
```diff
@@ -155,34 +155,39 @@ def _handle_subset_cert_request(config, domains, cert):
...
+    :param storage.RenewableCert lineage:
 
     :returns: Tuple of (str action, cert_or_None) as per _treat_as_renewal
               action can be: "newcert" | "renew" | "reinstall"
     :rtype: tuple
 
     """
-    if renewal.should_renew(config, cert):
-        return "renew", cert
+    if lineage.has_pending_deployment():
```
schoen: I think this is the right way to do this!

schoen: I'm not sure that it's ideal that there's absolutely no way to override this behavior with a command-line flag, although I appreciate the argument that typically users would not want to do so. While this is a good default even in respect to our UI, maybe we should consider having a way to be more explicit about this (although that risks a cascade of `--no-really` flags).

certbot/main.py
```diff
@@ -155,34 +155,39 @@ def _handle_subset_cert_request(config, domains, cert):
...
-    ).format(cert.configfile.filename, br=os.linesep)
+    ).format(lineage.configfile.filename, br=os.linesep)
 
     if config.verb == "run":
         keep_opt = "Attempt to reinstall this existing certificate"
     elif config.verb == "certonly":
         keep_opt = "Keep the existing certificate for now"
     choices = [keep_opt,
-               "Renew & replace the cert (limit ~5 per 7 days)"]
+               "Renew & replace the lineage (limit ~5 per 7 days)"]
```
schoen: @pde now says this should be "cert" for UI reasons.

schoen:
LGTM but @pde has something to change for UI reasons. Should be fine to merge after that.

Issue #3518, opened 2016-09-20
Nginx server block selection: Handle non-80/443 ports


ohemorange:
If we have to make a non-ssl block into an ssl block and it's not 80, ignore it.

If there are already ssl blocks that aren't 443, either ignore them, or install them, or otherwise factor this into ranking for choosing a block.

ohemorange:
See further commentary at #3546.

bmw:
Fixed in #3546.

Issue #1825, opened 2015-12-08
nginx http redirect


vasilevich:
Apologize in advance, I would edit it myself but lack the knowledge in python.

the HTTP redirect is implemented wrong:
example of current implementation

SERVER{
if(its http)
   {redirect to HTTP}
 listen 443 ssl;
}
problem: http traffic will never reach this block due to listen 443.
instead you should make an entirely new block
with the exact same server names:
server{
listen 80; #is default anyway
   server names exactly like in the edited block...
return ...what's in the if;
}

vasilevich:
apologizes again forgot to add, the issuse should be at the file 

letsencrypt-nginx/letsencrypt_nginx/configurator.py

lines: 359-388

Queeq:
AFAIK the best practice is this:

```
server {
        listen *:80;
        server_name example.com;
        # Redirect to https
        return 301 https://$server_name$request_uri;
}
```
`listen` and `server_name` should be taken from original configuration, definitely.

vasilevich:
I added this issuse from my phone.
here are some more details and the procedure to reproduce the error:
Nginx settings before running the script:
server {
    include /etc/nginx/subextensions/*;
    server_name www.duckgamer.com;
    server_name duckgamer.com;
    root $Websites/duckgamer;

    location / {

        }
}

, after running the script(in arch linux with default setup of pacman -S letsencrypt letsencrypt-nginx):
letsencrypt -d duckgamer.com -d www.duckgamer.com --standalone
in the easy/secure window, selected:
secure make all requests reditect to secure HTTPS access
new nginx config:
server {

    if ($scheme != "https") {
        return 301 https://$host$request_uri;
    }
    ssl_stapling_verify on;
    ssl_stapling on;
    ssl_trusted_certificate /etc/letsencrypt/live/duckgamer.com/chain.pem;
    include /etc/letsencrypt/options-ssl-nginx.conf;
    ssl_certificate_key /etc/letsencrypt/live/duckgamer.com/privkey.pem;
    ssl_certificate /etc/letsencrypt/live/duckgamer.com/fullchain.pem;
    error_log /var/lib/letsencrypt/error.log;
    access_log /var/lib/letsencrypt/access.log;
    listen 443 ssl;
    include /etc/nginx/subextensions/*;
    server_name www.duckgamer.com;
    server_name duckgamer.com;
    root $Websites/duckgamer;

    location / {
    }
}

The problem in this config is the combination of:
 if ($scheme != "https") {
[...]

PR #3546, opened 2016-09-27
Update Nginx redirect enhancement process to modify appropriate blocks

### LINKED ISSUE #3505: Implement Nginx server block selection
* NGINX server block treatment sync
  * Installing certificates
    * If >= 1 matching sslish block, add cert info to all matching sslish blocks
    * Else if >= 1 matching non-sslish block, make all matching non-sslish blocks sslish, and add cert info
    * For 0.9.0, if there are no matching blocks, error out and say “sorry you need to give me a relevant server block before I can install a cert for you”.
  * Redirect enhancement
    * At redirect time, there will be at least one matching sslish block.
    * If there are only matching sslish blocks, add listen on 80 and if statement redirect to all matching sslish blocks.
    * Else if there are both matching sslish and matching plaintextish blocks, add the if statement redirect to all matching plaintextish blocks.

ohemorange:
Fixes #3505, #3518, and #1825.


certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+                return test_port == self.DEFAULT_PORT
+            return test_port == matching_port
+
+        def _vhost_matches(vhost, port):
+            found_matching_port = False
+            for addr in vhost.addrs:
+                if _port_matches(port, addr.get_port()) and addr.ssl == False:
+                    found_matching_port = True
+            # if port is DEFAULT_PORT, check to see if there are no listen directives at all
+            if port == self.DEFAULT_PORT and len(vhost.addrs) == 0:
```
pde: *Maybe* easier to read if you do something like
```
if len(vhost.addrs) == 0:
  if port ==...:
else:
  for addr in vhost.addrs
```
to make it clear these are two alternative paths.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        """Get all ports that all vhosts are listening on, that match domain.
+
+        :param str domain: domain to find ports of
+        :returns: set of strings representing the port numbers
+        :rtype: set
+        """
+        matching_vhosts = [item['vhost'] for item in self._get_ranked_matches(domain)]
+        out = set([addr.get_port() for vhost in matching_vhosts for addr in vhost.addrs])
+        # also check the default port, 80
+        out.add(self.DEFAULT_PORT)
```
pde: (or if we always want 80 included for some semantic reason, explain that in the docstring)

ohemorange: That's not quite what this does - it's compensating for the fact that an empty port number defaults to 80. Theoretically, we could also just add every port number to `out` and scan that way, this is just probably faster than doing that.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+
+        """
+        vhost = None
+
+        matches = self._get_redirect_ranked_matches(target_name, port)
+        if not matches:
+            # No matches for this port. That's fine.
+            return None
+        elif matches[0]['rank'] in xrange(2, 6):
+            # Wildcard match - need to find the longest one
```
ohemorange: Yup, that's true. This comment came over from the old choose_vhosts, which originally only supported wildcard matches.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        ssl on. Rank by how well these match target_name.
+
+        :param str target_name: The name to match
+        :param str port: port number or "\*" or ""
+        :returns: list of dicts containing the vhost, the matching name, and
+            the numerical rank
+        :rtype: list
+
+        """
+        # Nginx chooses a matching server name for a request with precedence:
```
ohemorange: They're actually `{0,2,4,6}`, yes, this is a holdover to match the ranking system when we choose from vhosts for deploying and ocsp. No real reason not to rank that way, then update `xrange(2,6)` above, I can change it to that to make it more clear.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Yup, I like that.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -58,6 +58,8 @@ class NginxConfigurator(common.Plugin):
 
     hidden = True
 
+    DEFAULT_PORT = '80'
```
bmw: I didn't realize that in Nginx, 80 was literally the default port in that the lack of a port number defaults to 80. In that case, you can ignore this comment if you want.

certbot-nginx/certbot_nginx/configurator.py
bmw: If that's the case, should we maybe add it to the `if "" in out` block? If we want to treat port 80 specially, maybe we should either document it in this function or add it to the set returned by this function when it's relevant such as in `_enable_redirect`.

ohemorange: Well, maybe, unless as per the comment below we rip out this piece and only modify http servers listening on 80.

But yeah I agree that if we're keeping it, that's more elegant.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -241,6 +241,24 @@ def filedump(self, ext='tmp', lazy=True):
...
+
+        :param :class:`~certbot_nginx.obj.VirtualHost` vhost: The vhost in question
+
+        :returns: True if 'ssl on' directive is included
+        :rtype: bool
+
+        """
+        server = vhost.raw
+        for directive in server:
+            if not directive:
```
ohemorange: Great I like it! Then we should probably also fix this in parse_server, in parser.py (in a different PR).

certbot-nginx/certbot_nginx/configurator.py
ohemorange: I can rename it to something more descriptive to avoid future confusion.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
 
+    def all_vhost_ports(self, domain):
+        """Get all ports that all vhosts are listening on, that match domain.
+
+        :param str domain: domain to find ports of
+        :returns: set of strings representing the port numbers
+        :rtype: set
+        """
+        matching_vhosts = [item['vhost'] for item in self._get_ranked_matches(domain)]
+        out = set([addr.get_port() for vhost in matching_vhosts for addr in vhost.addrs])
```
ohemorange: That is a fun fact! If it doesn't break things no reason not to make the change, right? I don't think I have ever used Python generators so if they have weird gotchas I don't know them.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        # also check the default port, 80
+        out.add(self.DEFAULT_PORT)
+        if "" in out:
+            out.remove("")
+        if r"*" in out:
```
ohemorange: We are using raw string literals because I got a linter error at some point so I over-rawed to compensate. But I'll take out unnecessary r's.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+            out.remove(r"*")
+        return out
+
+    def choose_redirect_vhost(self, target_name, port):
+        r"""Chooses a virtual host for redirect enhancement.
```
ohemorange: That's interesting, I did the docstring that way because a different docstring that referred to `"\*"` did it that way. I imagine this is because that author also wanted the linter error to go away. I will find the best way to write this. How do I generate docs locally?

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+
+        Filter all hosts for those that include listen port but do not include
+        ssl on. Of these, find the most matching one based on target_name.
```
ohemorange: Yup, that description is more accurate.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        :param str target_name: domain name
+        :param str port: port number or "\*" or ""
+        :returns: vhost associated with name
+        :rtype: :class:`~certbot_nginx.obj.VirtualHost`
+
+        """
+        vhost = None
+
+        matches = self._get_redirect_ranked_matches(target_name, port)
+        if not matches:
```
ohemorange: I went back and forth on this one - I think it should be clear that they are separate pathways, but I agree that a future optimization could be to share some logical code at the lowest level. Except that's a bit more of a refactor, because the ranking is embedded pretty deeply. But it's definitely doable.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+            rank = matches[0]['rank']
+            wildcards = [x for x in matches if x['rank'] == rank]
+            vhost = max(wildcards, key=lambda x: len(x['name']))['vhost']
+        else:
+            vhost = matches[0]['vhost']
+
+        return vhost
+
+    def _get_redirect_ranked_matches(self, target_name, port):
+        r"""Gets a ranked list of plaintextish port-listening vhosts matching target_name
```
ohemorange: ACK

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+
+        Filter all hosts for those that include listen port but do not include
+        ssl on. Rank by how well these match target_name.
```
ohemorange: ACK

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        # Nginx chooses a matching server name for a request with precedence:
+        # 1. exact name match
+        # 2. longest wildcard name starting with *
+        # 3. longest wildcard name ending with *
+        # 4. first matching regex in order of appearance in the file
+        matches = []
+        all_vhosts = self.parser.get_vhosts()
+        def _port_matches(test_port, matching_port):
+            # test_port is a number, matching is a number or * or ""
+            if matching_port == r"\*":
```
ohemorange: oops this is a bug, we want to say True the string contains a single star.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        matches = self._get_redirect_ranked_matches(target_name, port)
+        if not matches:
+            # No matches for this port. That's fine.
+            return None
+        elif matches[0]['rank'] in xrange(2, 6):
+            # Wildcard match - need to find the longest one
+            rank = matches[0]['rank']
+            wildcards = [x for x in matches if x['rank'] == rank]
+            vhost = max(wildcards, key=lambda x: len(x['name']))['vhost']
+        else:
```
pde: Stylistic nit: it might be cleaner if this case came first,
```
if matches[0]['rank'] == 0:    # exactly named server block
    vhost = matches[0]['vhost']   
else:
    # wildcard or regex case
```

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+        # 2. longest wildcard name starting with *
+        # 3. longest wildcard name ending with *
+        # 4. first matching regex in order of appearance in the file
+        matches = []
+        all_vhosts = self.parser.get_vhosts()
+        def _port_matches(test_port, matching_port):
+            # test_port is a number, matching is a number or * or ""
+            if matching_port == r"\*":
+                return True
+            if matching_port == "":
```
ohemorange: TIL Python thinks empty strings are False.

But actually `if not matching_port` will also catch `None`, which is probably more correct just in case anyway.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Sounds good, no opinion here, again we may want to change this in a future PR in `choose_vhost`.

pde:
@bmw I do see your point about serving redirects on non-80 HTTP ports. Adding such redirects to server blocks is a little dangerous -- you might add one to a server that actually lives behind a TLS-terminating load balancer for instance. Although there are counter arguments: that could also be true on port 80, and the user shouldn't have requested a redirect if they didn't mean it.

I don't think it hugely matters for 0.9.0. No answer seems perfect. Maybe we should give the user choice about such questions, or maybe users who have complicated cases like this need to do manual work, so there's no point trying to polish everything for them.

bmw:
@pde what do you think about listening on port 80 for the purposes of setting up a redirect if we weren't already?

ohemorange:
@bmw I'm opposed to listening on port 80 to redirect if we weren't already. If you're not listening on 80 at all, there's no problem if someone tries to access a non-existent port 80 version of your site.

ohemorange:
Also, please ignore #3518. It's an artifact of going back and forth about which server blocks to modify how, and is honestly no longer relevant (back then, the steps were modify all blocks --> modify only those port blocks --> later on, figure out the best way to modify all port blocks).

If we want to modify only 80-blocks, do we also want to only modify 443-blocks?

certbot-nginx/certbot_nginx/configurator.py
bmw: Making this change doesn't break anything. It's certainly not a necessary change, but it should be a little faster.

You probably have used generators without even knowing it! `range` returns a list. `xrange` returns a generator. No real gotchas. Closest thing to a gotcha is you can only iterate forward through generator expressions and you can only go through them once. None of that matters here.

ohemorange:
We could handle this by printing a more specific success message, such as "modified all server blocks listening for plaintext traffic" in one scenario or "modified all server blocks listening for plaintext traffic on port 80 to redirect to https" in the other.

pde:
@ohemorange I think the questions of redirects on non-80 HTTP blocks, and of installing certs to non-443 TLS blocks, are largely orthogonal.

I'm pretty sure that installing the certs is mostly a good idea (though one day we may want a flag to control that behaviour.

I really don't know the best answer for redirects. I can see a case for: ship what we have, but open a wishlist ticket to implement a flag to control this in the future; change to be more conservative per @bmw's suggestion, with or without the future flag.  Probably I lean towards doing whatever will let us ship 0.9.0 for now :)

pde:
As for listening on 80 in order to serve a redirect? I think that might make sense *if* there are literally no HTTP server blocks. It's the kind of thing where asking questions and adding flags could be the right approach.

And again, my instinct is "go with what we have coded up right now, tweak and optimise this later".

certbot-nginx/certbot_nginx/configurator.py
bmw: Don't worry too much about getting the docs to display perfectly. It was just something small I noticed, however, here's how to see the problem.

I saw it using `IPython`. If you've [activated the virtual environment](https://certbot.eff.org/docs/contributing.html#getting-started) and have `IPython` installed, you can do:
```
$ ipython
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: from certbot_nginx.configurator import NginxConfigurator

In [2]: NginxConfigurator.choose_redirect_vhost?
Signature: NginxConfigurator.choose_redirect_vhost(self, target_name, port)
Docstring:
Chooses a virtual host for redirect enhancement.

Filter all hosts for those that include listen port but do not include
ssl on. Of these, find the most matching one based on target_name.

.. todo:: This should maybe return list if no obvious answer
    is presented.

.. todo:: The special name "$hostname" corresponds to the machine's
    hostname. Currently we just ignore this.

:param str target_name: domain name
:param str port: port number or "\*" or ""
:rtype: :class:`~certbot_nginx.obj.VirtualHost`
File:      ~/Code/certbot/certbot/certbot-nginx/certbot_nginx/configurator.py
Type:      instancemethod
```

Alternatively, you can do the same thing in the default Python interpreter (also after activating the venv):
```
$ python
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from certbot_nginx.configurator import NginxConfigurator
>>> help(NginxConfigurator.choose_redirect_vhost)
```

certbot-nginx/certbot_nginx/configurator.py
bmw: Yeah. I felt pretty confident that `matching_port` will always be a string. If you think this might not be the case and another [falsey](https://docs.python.org/2/library/stdtypes.html#truth-value-testing) value is possible and could cause problems for us, feel free to leave this as is.

ohemorange:
Update: we will only add redirects to server blocks listening plaintextishly on 80 (or empty string, which in Nginx means 80). We will then notify the user that we have installed redirects on all and only 80-blocks, or that we failed to find any 80-blocks.

Question (@bmw, @pde): Do servers listening on * count as 80-blocks or not 80-blocks? I'm leaning towards 80-blocks, because traffic coming in on 80 will reach those blocks.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Actually, the most descriptive and correct thing to do might be `if matching_port == "" or matching_port is None`. I think it's likely this line will disappear though.

bmw:
Some of [our code says otherwise](https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/obj.py#L58) but when I try it, Nginx yells at me:
```
nginx: [emerg] invalid port in "127.0.0.1:*" of the "listen" directive in /etc/nginx/sites-enabled/default:18
```

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Thanks! Perhaps we might want to fix this in `obj.py:Addr`'s documentation as well, later...

ohemorange:
Well that certainly simplifies this question.

We may (again, future.) want to update the various places in the code that don't know that.

ohemorange:
Ok, so if Nginx sees two server blocks listening on the same address with the same server_name match-level, it warns about the second one and only takes the first. Again, to satisfy our function of "if you get non-https traffic to this domain we'll make sure it redirects", I think it makes sense to only modify the first matching server we see. Which is what the code was doing before (per port), but I said something slightly different above.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -193,6 +198,8 @@ def choose_vhost(self, target_name):
         :rtype: :class:`~certbot_nginx.obj.VirtualHost`
 
         """
+        if self.vhost is not None:
```
ohemorange: This was supposed to get set to vhost at the end of this function. Instead, I set it to None.

ohemorange: The goal is to make sure that we're modifying the same block when we deploy the cert and when we add ocsp.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Well, this code's gone anyway. Next time for generators!

certbot-nginx/certbot_nginx/configurator.py
ohemorange: This code's gone, hopefully it's better documented in the next version.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Actually, the best way to fix all this `r"\*"` stuff is to take out all code referencing it, because it's wrong! Woohoo!

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Actually, no it's not - what we want is "Chooses the vhost most closely matching target_name that is listening to port without using ssl."

certbot-nginx/certbot_nginx/configurator.py
schoen: Another awesome thing is that newer Pythons also have set comprehensions, like

        out = {addr.get_port() for vhost in matching_vhosts for addr in vhost.addrs}

Sadly, that's incompatible with Python 2.6 so I don't think we would be allowed to use it. :frowning:

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Actually, looking at this again, this looks weird and might be a bug both here and when choosing which vhost to deploy to. TODO(ebp): take a look after this set of changes.

ohemorange: Ah ok no, that's not true, this is only wildcard because 6 is excluded. I'm going to change the numbering system to make this more clear. Phew.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: As per previous, no longer agree because this else catches regex or exact. I'll put in a comment to that effect.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Also amending this one. Ranks are `{0,1,2,3,4,5,6,7}`. That's just an ordered list on top, leaving out the "also rank by ssl" part. I think it makes more sense in the amended version, so leaving this.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: luckily this no longer matters!

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -250,6 +259,121 @@ def _get_ranked_matches(self, target_name):
...
+                found_matching_port = True
+
+            if found_matching_port:
+                # make sure we don't have an 'ssl on' directive
+                return not self.parser.has_ssl_on_directive(vhost)
+            else:
+                return False
+        matching_vhosts = [vhost for vhost in all_vhosts if _vhost_matches(vhost, port)]
+
+        for vhost in matching_vhosts:
```
ohemorange: Done.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Anyway, I realize that if we're doing this it should be cached by nameset or something. It's probably best to not do this particular thing for now, and maybe add some other type of caching as a feature later because it will require more careful thought.

bmw:
I think that's fine. If a user has two server blocks on the same address, port, and match level, I think one could argue their config is broken/ambiguous and I wouldn't worry about it too much for 0.9.0. (I understand Nginx does have ordering rules here on which one it uses, but why would you place the burden of understanding that on yourself/others?)

certbot-nginx/certbot_nginx/configurator.py
bmw: I think that's fine. We've done a few similar things in Apache. We can probably just use a dictionary that maps domain names to their SSLish server block(s).

certbot-nginx/certbot_nginx/configurator.py
bmw: You're right! English is hard.

ohemorange:
It doesn't even have ordering rules for that situation, it warns and ignores the second one it encounters:
`nginx: [warn] conflicting server name "*.functorkitten.xyz" on 0.0.0.0:443, ignored`

So it's definitely broken anyway.

bmw:
Even better!

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Luckily, we're using Python.

ohemorange:
Oh, noticed something interesting from the compatibility tests (which I assume were also failing before this change? @schoen ?)

```
server {
    server_name test.org;
}
```

When we deploy, we add `listen 443` here. Before, we would add the if statement to the block, which was definitely wrong. Now, we don't add it to this block at all. We should really be adding both a `listen 80` and the if statement here during redirect enhancement.

A hacky way to fix this is when we add `listen 443` to blocks with no listen directives, also add `listen 80`. I don't think this would be wrong? Because the statement was technically there, just implicit before?

The least hacky way would be to set up that mapping system, note that it previously had no listen directives, and then add this vhost into the list of vhosts we test for matchingness (which it will almost 100% certainly win I think so actually we could then just keep a note to enhance this one if we're asked later).

bmw:
From the [Nginx documentation on the listen directive](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen):
> If the directive is not present then either *:80 is used if nginx runs with the superuser privileges, or *:8000 otherwise.

That's fun. Being corner casey, I'm not worried about this for 0.9.0, but I'll throw out solutions which feel a little more intuitive to me:

1. If we don't want to assume the lack of a `listen` directive means `listen *:80;` then I really think we can kick this case WAY down the road. Coming up with a platform independent way of determining if `nginx` has superuser privileges sounds awful. Maybe we should detect this case and error out because we know we might not do everything properly. This error detection isn't necessary for 0.9.0 either in my opinion.
2. If we want to assume the lack of a `listen` directive means `listen *:80;`, then I think the `server` block you initially posted had an implicit `listen *:80;`. When we messed with their config, we broke that implicit directive and now`nginx` is (probably) responding differently to requests on port 80 before we touched it. The user only asked us to set up HTTPS, not to change other behavior. If we see this case and have to add a `listen 443 ssl;` directive, we should just add a `listen 80;` directive right then and there to preserve the behavior the user currently set up. This is simpler to implement than adding it only if the user asks for a redirect and properly preserves their server's behavior if they don't.

ohemorange:
Your solution 2 is the same as the first thing I had above, right? Or is there something I'm missing?

bmw:
Whoops. I somehow missed your "hacky" solution. +1 to that. Personally, that feels more correct than the "least hacky" solution because we're changing the current (implicit) behavior of nginx if the user doesn't ask for a redirect (assuming I'm understanding your suggestion).

certbot-nginx/certbot_nginx/tests/configurator_test.py
```diff
@@ -401,6 +437,21 @@ def test_redirect_enhance(self):
...
+        # Test that we successfully add a redirect when there is
+        # no listen directive
+        migration_conf = self.config.parser.abs_path('sites-enabled/migration.com')
+        self.config.enhance("migration.com", "redirect")
+
+        generated_conf = self.config.parser.parsed[migration_conf]
+        self.assertTrue(util.contains_at_depth(generated_conf, expected, 2))
+
+    def test_redirect_dont_enhance(self):
+        with LogCapture() as l:
```
bmw: While this is definitely nicer for the developer than what I'm about to propose, I think we should remove `testfixtures` from the project. Many of our package maintainers use our unit tests to check if Certbot is working properly on their system. When they run these tests, they do so without our `dev` dependencies meaning this test would fail for them. We could add `testfixtures` as a regular dependency but that is a step backwards on #1301 and requires all OS's to have a `testfixtures` package available.

Instead, I think we should just use `mock` here. You can probably register logging handler and check output that way, but for this test you can do something like:
```
        with mock.patch("certbot_nginx.configurator.logger") as mock_logger:
            self.config.enhance("geese.com", "redirect")
        self.assertEqual(mock_logger.info.call_args[0][0],
                'No matching insecure server blocks listening on port %s found.')
```
Note the need to use `%s` rather than the port number.

setup.py
```diff
@@ -76,6 +76,7 @@ def read_file(filename, encoding='utf8'):
     'pep8',
     'psutil>=2.2.1',  # for tests, optional
     'pylint==1.4.2',  # upstream #248
+    'testfixtures',
```
bmw: Don't forget to remove this.

certbot-nginx/certbot_nginx/tests/configurator_test.py
ohemorange: Good idea, fewer dependencies sounds good to me.

bmw:
Everything seemed to work as expected! Merging this when Travis is happy.

PR #3551, opened 2016-09-28
Mention that the domain is used to choose filename [revision requested]

dehnert:
The cert filename is chosen based on the first domain listed. With certs with overlapping domains or where some domains are less canonical, it's therefore useful to put the most canonical/unique domain first. This updates the help text to inform users of this fact.

When I used certbot for the first time, I didn't realize this, and ended up with a somewhat poorly named cert directory.


dehnert:
A certbot subcommand for moving and updating all the relevant files/directories to change a cert's short name might also be nice, or documentation on how to do that, but I'm not sure it's worth maintaining such a thing.

bmw:
I made the requested changes in #4021 and included this commit. Thanks for the PR!

PR #3560, opened 2016-09-29
Renew symlink safety

pde:
Re-do the fix for #3497 to ensure it works in all cases.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -161,10 +161,7 @@ def _handle_identical_cert_request(config, lineage):
     :rtype: tuple
 
     """
-    if lineage.has_pending_deployment():
-        logger.warn("Found a new cert /archive/ that was not linked to in /live/; "
-                    "fixing and reinstalling..")
-        lineage.update_all_links_to(lineage.latest_common_version())
+    if not lineage.ensure_deployed():
         return "reinstall", lineage
     if renewal.should_renew(config, lineage):
         return "renew", lineage
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -166,9 +166,9 @@ def _handle_identical_cert_request(config, lineage):
 
     """
     if lineage.has_pending_deployment():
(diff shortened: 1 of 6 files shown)


certbot/main.py
```diff
@@ -87,11 +83,17 @@ def _auth_from_domains(le_client, config, domains, lineage=None):
         # interested in
         action = "renew"
 
+    if action == "renew" and lineage.has_pending_deployment():
```
bmw: Why should we only do this if `action == "renew"`? Why not do it for all actions? Since the current behavior is to always deploy the certificate, the only way `has_pending_deployment` is `True` is if Certbot crashed. Let's give the user their more or less hidden certificate, help them avoid rate limits, etc.

bmw:
Also after chatting about this with you, this doesn't technically fix #3497 either. In the `certbot renew` case, the `should_renew` function determines whether or not to even call into `obtain_cert` and `_auth_from_domains`. For the problem described in #3497, `should_renew` will return `False` and `obtain_cert` is never called.

certbot/storage.py
```diff
@@ -520,6 +520,22 @@ def next_free_version(self):
...
+        """Make sure we've deployed the latest version.
+
+        :returns: False if a change was needed, True otherwise
+        :rtype: bool
+
+        May need to recover from rare interrupted / crashed states."""
+
+        if self.has_pending_deployment():
+            logger.warn("Found a new cert /archive/ that was not linked to in /live/; "
+                        "fixing and reinstalling..")
```
bmw: "reinstalling" is only true in the `run` case. When using `renew` or `certonly`, the cert is not reinstalled. For this reason, I think we should remove this word. If we really want, we can add logic to the `run` function to print a message about reinstalling the existing certificate if the action is "reinstall".

pde: OK done.

PR #3561, opened 2016-09-29
Allow validation of cross-domain redirects

ohemorange:
Update compatibility validator to pass redirect check when redirecting to a different domain, whether http or https.

### DIFF
--- a/certbot-compatibility-test/certbot_compatibility_test/validator.py
+++ b/certbot-compatibility-test/certbot_compatibility_test/validator.py
@@ -41,7 +41,11 @@ class Validator(object):
 
         redirect_location = response.headers.get("location", "")
         if not redirect_location.startswith("https://"):
-            return False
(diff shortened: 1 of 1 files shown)


certbot-compatibility-test/certbot_compatibility_test/validator.py
```diff
@@ -41,7 +41,11 @@ def redirect(self, name, port=80, headers=None):
 
         redirect_location = response.headers.get("location", "")
         if not redirect_location.startswith("https://"):
-            return False
+            if not redirect_location.startswith("http://"):
```
pde: Maybe add a comment saying something like, "it's okay for some server configuration to redirect to an http URL, as long as it's on some other domain"?

ohemorange: Done.

PR #3562, opened 2016-09-29
Tweak for Travis performance

pde:
- merge `cover` and `py27 BOULDER_INTEGRATION` into one matrix entry
 - re-order to put the fastest environments last, improving average
   case parallelism
 - but ensure that the most likely-to-break tests (cover, lint) start first to fail early.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,12 +32,7 @@ matrix:
         - sudo cat /var/log/mysql/error.log
         - ps aux | grep mysql
     - python: "2.7"
-      env: TOXENV=apacheconftest
-      sudo: required
-    - python: "2.7"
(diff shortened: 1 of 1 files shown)


pde:
In an unscientific n=1 comparison, this reduces Travis runtime from [12:57](https://travis-ci.org/certbot/certbot/builds/163852000) to [12:13](https://travis-ci.org/certbot/certbot/builds/163855927), saving 45 seconds.

pde:
The theory was: generally order by descending run time, to avoid having one thread working on a long job after everything else is done. `lint` gets to go early so that it can fail early, because it fails so often. It's also moderately long-running, too.

PR #3565, opened 2016-09-29
Nginx compatibility test [revision requested]

schoen:
Updates to nginx compatibility test to include lots more test cases, and also run with -ai instead of the default -aie due to a small number of enhancement (redirect) failures


bmw:
Travis killed the build because the Nginx compatibility test took too long! We'll have to rethink our approach here.

ohemorange:
Some potential slowdowns:
- Generating snakeoil certs on each run (instead, ship a pre-generated cert)
- sleep(1) when restarting server (restart server after making a batch of changes)

bmw:
We should try to speed up the compatibility tests to get this merged. If that doesn't work, we may need to reduce the test suite, possibly by splitting the files between different Travis jobs.

@schoen are you planning on looking at this?

bmw:
For the record, after taking this branch, merging in `master`, and running the tests, 328 (1028 bit) RSA keys are generated.

jsha:
@schoen, ping on this?

bmw:
We should update this PR so it isn't run in Travis. I'm planning on setting up these tests to run automatically every night and report the results to us.

bmw:
Closed in favor of #4863.

Issue #3555, opened 2016-09-28
Missing Content-Type 'application/json' in POST requests


niklas-studer:
When certbot posts a request to any of the ACME endpoints it does not include the Content-Type 'application/json' but leaves that fields empty event though it is posting JSON data

PR #3566, opened 2016-09-30
Use correct Content-Types in headers.

cowlicks:
This closes https://github.com/certbot/certbot/issues/3555

The content type for all posts requests is now `application/jose+json`, which is in compliance with the latest ACME spec. https://tools.ietf.org/html/draft-ietf-acme-acme-03


cowlicks:
This is ready for review.

pde:
Looks good to me for 0.10.0?

PR #3571, opened 2016-10-02
Impelment account deactivation

### LINKED ISSUE #3483: Implement account deactivation
Boulder recently merged code to allow for [ACME draft-03 style](https://tools.ietf.org/html/draft-ietf-acme-acme-03#section-6.2.2) account deactivation.

The process is extremely simple, to deactivate their account a user must POST a signed payload containing `{"status":"deactivated"}` to the `/acme/reg/{id}` endpoint. The server will respond with 200 and the contents of the deactivated registration if successful.

cowlicks:
Closes #3483

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -154,7 +154,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         new_regr = self._regr_from_response(
             response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
             terms_of_service=regr.terms_of_service)
-        if response.status_code not in (200, 202):
+        if response.status_code != 200:
             msg = "Server returned bad status code: %s"
             raise errors.DeactivationError(msg % (response.status_code,))
         elif new_regr != regr:
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -148,20 +148,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         :rtype: `.RegistrationResource`
 
         """
-        update = {'status': 'deactivated'}
-        body = messages.UpdateRegistration(**dict(update))
-        response = self.net.post(regr.uri, body)
-        new_regr = self._regr_from_response(
-            response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
-            terms_of_service=regr.terms_of_service)
-        if response.status_code != 200:
-            msg = "Server returned bad status code: %s"
-            raise errors.DeactivationError(msg % (response.status_code,))
-        elif new_regr != regr:
-            m = "Server returen bad registration object"
-            raise errors.DeactivationError(m)
-
-        return new_regr
+        return self.update_registration(regr, update={'status': 'deactivated'})
 
     def query_registration(self, regr):
         """Query server about registration.
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -152,7 +152,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             m = "Server returen bad registration object"
             raise errors.DeactivationError(m)
 
-        return self.update_registration(regr, update=update)
+        return new_regr
 
     def query_registration(self, regr):
         """Query server about registration.
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -139,7 +139,19 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         return updated_regr
 
     def deactivate(self, regr):
-        update = {'status': 'deactivate'}
+        update = {'status': 'deactivated'}
+        body = messages.UpdateRegistration(**dict(update))
+        response = self.net.post(regr.uri, body)
+        new_regr = self._regr_from_response(
+                response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
+                terms_of_service=regr.terms_of_service)
+        if response.status_code != 200:
+            m = "Server returned bad status code: %s" % (response.status_code,)
+            raise errors.DeactivationError(m)
+        elif new_regr != regr:
+            m = "Server returen bad registration object"
+            raise errors.DeactivationError(m)
+
         return self.update_registration(regr, update=update)
 
     def query_registration(self, regr):
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -132,11 +132,15 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
         """
         update = regr.body if update is None else update
-        updated_regr = self._send_recv_regr(
-            regr, body=messages.UpdateRegistration(**dict(update)))
+        body = messages.UpdateRegistration(**dict(update))
+        updated_regr = self._send_recv_regr(regr, body=body)
         if updated_regr != regr:
             raise errors.UnexpectedUpdate(regr)
         return updated_regr
+
+    def deactivate(self, regr):
+        update = {'status': 'deactivate'}
+        return self.update_registration(regr, update=update)
 
     def query_registration(self, regr):
         """Query server about registration.
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -138,7 +138,16 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             raise errors.UnexpectedUpdate(regr)
         return updated_regr
 
-    def deactivate(self, regr):
+    def deactivate_registration(self, regr):
+        """Deactivate registration.
+
+        :param messages.RegistrationResource regr: The Registration Resource
+            to be deactivated.
+
+        :returns: The Registration resource that was deactivated.
+        :rtype: `.RegistrationResource`
+
+        """
         update = {'status': 'deactivated'}
         body = messages.UpdateRegistration(**dict(update))
         response = self.net.post(regr.uri, body)
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -143,11 +143,11 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         body = messages.UpdateRegistration(**dict(update))
         response = self.net.post(regr.uri, body)
         new_regr = self._regr_from_response(
-                response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
-                terms_of_service=regr.terms_of_service)
-        if response.status_code != 200:
-            m = "Server returned bad status code: %s" % (response.status_code,)
-            raise errors.DeactivationError(m)
+            response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
+            terms_of_service=regr.terms_of_service)
+        if response.status_code not in (200, 202):
+            msg = "Server returned bad status code: %s"
+            raise errors.DeactivationError(msg % (response.status_code,))
         elif new_regr != regr:
             m = "Server returen bad registration object"
             raise errors.DeactivationError(m)
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -130,19 +130,11 @@ class ClientTest(unittest.TestCase):
         self.assertEqual(self.regr,
                          self.client.deactivate_registration(self.regr))
 
-    def test_deactivate_account_bad_status(self):
-        self.response.headers['Location'] = self.regr.uri
-        self.response.json.return_value = self.regr.body.to_json()
-        self.response.status_code = 500  # something that isn't 200 or 202
-        self.assertRaises(errors.DeactivationError,
-                          self.client.deactivate_registration,
-                          self.regr)
-
     def test_deactivate_account_bad_registration_returned(self):
         self.response.headers['Location'] = self.regr.uri
         self.response.json.return_value = "some wrong registration thing"
         self.assertRaises(
-            errors.DeactivationError,
+            errors.UnexpectedUpdate,
             self.client.deactivate_registration,
             self.regr)
 
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -124,6 +124,11 @@ class ClientTest(unittest.TestCase):
         self.assertRaises(
             errors.UnexpectedUpdate, self.client.update_registration, self.regr)
 
+    def test_deactivate_account(self):
+        self.response.headers['Location'] = self.regr.uri
+        self.response.json.return_value = self.regr.body.to_json()
+        self.assertEqual(self.regr, self.client.deactivate(self.regr))
+
     def test_query_registration(self):
         self.response.json.return_value = self.regr.body.to_json()
         self.assertEqual(self.regr, self.client.query_registration(self.regr))
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -127,7 +127,8 @@ class ClientTest(unittest.TestCase):
     def test_deactivate_account(self):
         self.response.headers['Location'] = self.regr.uri
         self.response.json.return_value = self.regr.body.to_json()
-        self.assertEqual(self.regr, self.client.deactivate_registration(self.regr))
+        self.assertEqual(self.regr,
+                         self.client.deactivate_registration(self.regr))
 
     def test_deactivate_account_bad_status(self):
         self.response.headers['Location'] = self.regr.uri
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -129,6 +129,21 @@ class ClientTest(unittest.TestCase):
         self.response.json.return_value = self.regr.body.to_json()
         self.assertEqual(self.regr, self.client.deactivate(self.regr))
 
+    def test_deactivate_account_bad_status(self):
+        self.response.headers['Location'] = self.regr.uri
+        self.response.json.return_value = self.regr.body.to_json()
+        self.response.status_code = 202
+        self.assertRaises(
+            errors.DeactivationError,
+            lambda: self.client.deactivate(self.regr))
+
+    def test_deactivate_account_bad_registration_returned(self):
+        self.response.headers['Location'] = self.regr.uri
+        self.response.json.return_value = "some wrong registration thing"
+        self.assertRaises(
+            errors.DeactivationError,
+            lambda: self.client.deactivate(self.regr))
+
     def test_query_registration(self):
         self.response.json.return_value = self.regr.body.to_json()
         self.assertEqual(self.regr, self.client.query_registration(self.regr))
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -133,16 +133,17 @@ class ClientTest(unittest.TestCase):
         self.response.headers['Location'] = self.regr.uri
         self.response.json.return_value = self.regr.body.to_json()
         self.response.status_code = 500  # something that isn't 200 or 202
-        self.assertRaises(
-            errors.DeactivationError,
-            lambda: self.client.deactivate_registration(self.regr))
+        self.assertRaises(errors.DeactivationError,
+                          self.client.deactivate_registration,
+                          self.regr)
 
     def test_deactivate_account_bad_registration_returned(self):
         self.response.headers['Location'] = self.regr.uri
         self.response.json.return_value = "some wrong registration thing"
         self.assertRaises(
             errors.DeactivationError,
-            lambda: self.client.deactivate_registration(self.regr))
+            self.client.deactivate_registration,
+            self.regr)
 
     def test_query_registration(self):
         self.response.json.return_value = self.regr.body.to_json()
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -127,7 +127,7 @@ class ClientTest(unittest.TestCase):
     def test_deactivate_account(self):
         self.response.headers['Location'] = self.regr.uri
         self.response.json.return_value = self.regr.body.to_json()
-        self.assertEqual(self.regr, self.client.deactivate(self.regr))
+        self.assertEqual(self.regr, self.client.deactivate_registration(self.regr))
 
     def test_deactivate_account_bad_status(self):
         self.response.headers['Location'] = self.regr.uri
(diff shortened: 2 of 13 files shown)


cowlicks:
Functionally this does it. But there are a few things.

* The spelling `certbot registration --deactivate` is awkward. Suggestions welcome.
* The code in `main.py` needs some refactoring so adding this is less awkward.
* I tested this a local boulder server and it seems to work. But I need to add integration tests.

certbot/client.py
```diff
@@ -163,7 +163,7 @@ def perform_registration(acme, config):
 
 
 class Client(object):
-    """ACME protocol client.
+    """Certbot's client.
```
cowlicks: This makes more sense right?

cowlicks:
@rolandshoemaker I'm testing this against boulder master, with this patch 

```
commit 16084872c6427097fb99aca51a5b746f1c039c8a
Author: Blake Griffith <blake.a.griffith@gmail.com>
Date:   Sun Oct 2 20:43:36 2016 -0700

    Enable account deactivation.

diff --git a/wfe/wfe.go b/wfe/wfe.go
index 521d7ff..108f607 100644
--- a/wfe/wfe.go
+++ b/wfe/wfe.go
@@ -1126,7 +1126,7 @@ func (wfe *WebFrontEndImpl) Registration(ctx context.Context, logEvent *requestE
 	// If a user tries to send both a deactivation request and an update to their
 	// contacts or subscriber agreement URL the deactivation will take place and
 	// return before an update would be performed.
-	if features.Enabled(features.AllowAccountDeactivation) && (update.Status != currReg.Status) {
+	if update.Status != currReg.Status {
 		if update.Status != core.StatusDeactivated {
 			wfe.sendError(response, logEvent, probs.Malformed("Invalid value provided for status field"), nil)
 			return
```

But when I try to create an account and deactivate it
```
(venv) ➜  letsencrypt git:(gh-3483) ✗ source tests/integration/_common.sh 
Root integration tests directory: /tmp/leitgJnz
(venv) ➜  letsencrypt git:(gh-3483) ✗ certbot_test certonly -a standalone -d example.com      
...
(venv) ➜  letsencrypt git:(gh-3483) ✗ certbot_test register --deactivate 
...
Error: urn:acme:error:malformed :: The request message was malformed :: Error deactivating registration :: Only vaid registrations can be deactivated

  File "/home/blake/github/letsencrypt/venv/bin/certbot", line 11, in <module>
  File "/home/blake/github/letsencrypt/certbot/main.py", line 796, in main
  File "/home/blake/github/letsencrypt/certbot/main.py", line 409, in register
    acme_client.acme.deactivate(acc.regr)
  File "/home/blake/github/letsencrypt/acme/acme/client.py", line 138, in deactivate
    response = self.net.post(regr.uri, body)
  File "/home/blake/github/letsencrypt/acme/acme/client.py", line 665, in post
    return self._check_response(response, content_type=content_type)
  File "/home/blake/github/letsencrypt/acme/acme/client.py", line 581, in _check_response
    raise messages.Error.from_json(jobj)
Error: urn:acme:error:malformed :: The request message was malformed :: Error deactivating registration :: Only vaid registrations can be deactivated

```

The full output is [here](https://gist.github.com/cowlicks/a871dc791c34329cc5c9078b413348b3). It seems like the account registration object is not being created with the `Status` field set to `core.StatusValid`. 


Note I had to apply a small [tweak to the _common.sh script](https://gist.github.com/cowlicks/866e2007c4f2f04504a46d2607985734) to get it to work for me.

cowlicks:
This is blocked by https://github.com/letsencrypt/boulder/issues/2224

cowlicks:
This is working and almost ready if anyone would like to look at it.

@bmw I think you mentioned something about how this should be released?

certbot/cli.py
```diff
@@ -723,6 +723,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="With the register verb, indicates that details associated "
              "with an existing registration, such as the e-mail address, "
              "should be updated, rather than registering a new account.")
+    helpful.add(
+        "register", "--deactivate", action="store_true",
+        help="Irrecvocably deactivate your account")
```
cowlicks: This needs to be more descriptive.

cowlicks:
The behavior associated with account deactivation is specific to the certificate authority (according to the ACME spec). Boulder/LE allows users to revoke their certs after their account is deactivated.

bmw:
Is this ready for review @cowlicks? IIRC, the last I talked to you about it you wanted to continue working on this PR before it was reviewed.

cowlicks:
@bmw Yeah, I wanted to add more tests though. I've been busy with a contract lately.

tests/boulder-integration.sh
```diff
@@ -119,6 +119,8 @@ common revoke --cert-path "$root/conf/live/le1.wtf/cert.pem"
 common revoke --cert-path "$root/conf/live/le2.wtf/cert.pem" \
        --key-path "$root/conf/live/le2.wtf/privkey.pem"
 
+common register --deactivate
```
cowlicks: This isn't actually deactivating the account because with the `--non-interactive`the defaults to `False` for deactivation.

Is there a way around using the default in `--non-interactive` mode? Should account deactivation not default to `False`?

acme/acme/client.py
```diff
@@ -150,7 +150,7 @@ def deactivate(self, regr):
         new_regr = self._regr_from_response(
             response, uri=regr.uri, new_authzr_uri=regr.new_authzr_uri,
             terms_of_service=regr.terms_of_service)
-        if response.status_code != 202:
+        if response.status_code not in (200, 202):
```
cowlicks: Boulder currently returns 200 but it is supposed to return 202. See [boulder #2313](https://github.com/letsencrypt/boulder/issues/2313).

certbot/cli.py
```diff
@@ -709,6 +709,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="With the register verb, indicates that details associated "
              "with an existing registration, such as the e-mail address, "
              "should be updated, rather than registering a new account.")
+    helpful.add(
+        "register", "--deactivate", action="store_true",
+        help="Irrevocably deactivate your account. Certificates associated "
+             "with your account can still be revoked, but they can not be "
+             "renewed.")
```
bmw: While Boulder doesn't do this, in the ACME spec it says:
```
   Once an account is deactivated... It is up to server policy...
   whether to revoke certificates issued by that account...
```
I think this potential side effect should be mentioned here.

I'm also not sure about the phrase "but they can not be renewed". From a Certbot perspective, if the user runs `certbot certonly` once to associate the lineages with a new account, they can still renew their certs. I think either this phrase should be removed or changed to something like "but they can not be renewed without first obtaining certificates from a new account".

certbot/client.py
bmw: Probably. This class is still an ACME client in the sense that it uses the ACME protocol, however, it does so at a higher level and contains a lot of Certbot specific code. I'm fine with either docstring.

certbot/main.py
```diff
@@ -382,6 +382,28 @@ def register(config, unused_plugins):
     # exist or not.
     account_storage = account.AccountFileStorage(config)
     accounts = account_storage.find_all()
+    reporter_util = zope.component.getUtility(interfaces.IReporter)
+    add_msg = lambda m: reporter_util.add_message(m, reporter_util.MEDIUM_PRIORITY)
+
+    if config.deactivate:
+        if len(accounts) == 0:
+            add_msg("Could not find existing account to deactivate.")
```
bmw: I think we should just `return` this string here. This will cause this message to be printed to stderr and exit with a nonzero status. We do the same thing below when updating an account and it cannot be found. I think this is an error condition because the user requested action could not be completed.

certbot/main.py
```diff
@@ -382,6 +382,28 @@ def register(config, unused_plugins):
...
+            prompt = ("Are you SURE you would like to irrevocably deactivate "
+                      "your account? You will lose access to the domains "
+                      "associated with it.")
+            wants_deactivate = yesno(prompt, yes_label='Deactivate',
+                                     no_label='Abort', cli_flag='--deactivate',
+                                     default=False)
+            if wants_deactivate:
+                acc, acme = _determine_account(config)
+                acme_client = client.Client(config, acc, None, None, acme=acme)
+                acme_client.acme.deactivate(acc.regr)
```
bmw: After doing this, I think we should either delete the account or update the data on disk to reflect deactivation. I think I have a slight preference for deleting the account, but we can probably take either approach here.

certbot/main.py
```diff
@@ -382,6 +382,28 @@ def register(config, unused_plugins):
...
+                add_msg("Account deactivated.")
+            else:
+                add_msg("Deactivation aborted.")
```
bmw: Like above, I think we should just return this string here. The user requested action could not be completed and we effectively do this in [other places](https://github.com/certbot/certbot/blob/master/certbot/main.py#L187) if the user cancels their request.

certbot/tests/cli_test.py
```diff
@@ -1042,34 +1042,73 @@ def test_update_registration_unsafely(self):
                 "--register-unsafely-without-email".split())
             self.assertTrue("--register-unsafely-without-email" in x[0])
 
+    @mock.patch('certbot.main._determine_account')
```
bmw: Normally I would encourage you to put these tests in `main.py` to avoid adding to the insanity of this file when you're testing something in `main.py`, however, since you told me you're working on this in another PR, I don't care in which PR you make this change.

tests/boulder-integration.sh
bmw: > Is there a way around using the default in --non-interactive mode? Should account deactivation not default to False?

There should definitely be a way for users who run certbot with `--non-interactive` to deactivate their account. The two ways we've solved this problem in the past are to either change the default or make the user provide an extra flag like `--mass-revocation-ok`. I think I prefer just changing the default, but either option (or potentially another approach you can think of) is fine with me.

certbot/main.py
```diff
@@ -382,6 +382,28 @@ def register(config, unused_plugins):
     # exist or not.
```
bmw: The `register` function here is getting a little long and should probably be broken up into smaller functions. While I think this should definitely at some point, this can be done in another PR if you prefer.

certbot/main.py
```diff
@@ -382,6 +382,28 @@ def register(config, unused_plugins):
...
+    add_msg = lambda m: reporter_util.add_message(m, reporter_util.MEDIUM_PRIORITY)
+
+    if config.deactivate:
+        if len(accounts) == 0:
+            add_msg("Could not find existing account to deactivate.")
+        else:
+            yesno = zope.component.getUtility(interfaces.IDisplay).yesno
+            prompt = ("Are you SURE you would like to irrevocably deactivate "
+                      "your account? You will lose access to the domains "
+                      "associated with it.")
```
bmw: I think the sentence "You will lose access to the domains associated with it." is a bit ambiguous and overstates the problem a bit. Instead, I think you can mention some of the other concerns from the help text for `--deactivate` and the comments I posted on that.

cowlicks:
@bmw I'm testing against boulder master running locally with it turned on ([by setting this](https://github.com/letsencrypt/boulder/blob/ce679bad41dc291bd8989ba7ce285d9c9c1c707b/features/features.go#L25)). This works for me.

cowlicks:
@bmw I'm testing against boulder master running locally with it turned on ([by setting this](https://github.com/letsencrypt/boulder/blob/ce679bad41dc291bd8989ba7ce285d9c9c1c707b/features/features.go#L25)). This works for me.

I think @cpu reported it was turned on in production at some point in IRC. I don't recall if I tested against production though. It could have been turned off by now for some reason.

acme/acme/client_test.py
```diff
@@ -121,6 +121,26 @@ def test_update_registration(self):
...
+        self.response.json.return_value = self.regr.body.to_json()
+        self.assertEqual(self.regr, self.client.deactivate(self.regr))
+
+    def test_deactivate_account_bad_status(self):
+        self.response.headers['Location'] = self.regr.uri
+        self.response.json.return_value = self.regr.body.to_json()
+        self.response.status_code = 500  # something that isn't 200 or 202
+        self.assertRaises(
+            errors.DeactivationError,
+            lambda: self.client.deactivate(self.regr))
```
cowlicks: I usually prefer lambda's since they seem more explicit. But your style is indeed more consistent with the codebase.

certbot/cli.py
cowlicks: Strange, I didn't know boulder would allow you to do that. If you can still revoke you certs and still renew you certs, then what is even the point of account deactivation?  I thought the point was to block someone from renewing your certs if your account was compromised.

certbot/tests/cli_test.py
cowlicks: Yeah, I'll include that in a seprerate PR. The work is [here](https://github.com/cowlicks/letsencrypt/blob/gh-2716/certbot/tests/not_cli_test.py).

cowlicks:
@bmw how does alias'ing usually work? I don't see it in cli.py

certbot/main.py
cowlicks: done

tests/boulder-integration.sh
cowlicks: I decided to switch the default to True. So the account is deactivated.

certbot/client.py
cowlicks: I prefer "Certbot's client" just to disambiguate it from the acme package's client.

certbot/cli.py
bmw: Authorization reuse is the biggest reason.

After successfully performing a DV challenge, the ACME server has a valid authorization object saying that it may issue a certificate to the owner of the account whose request created the authorization. This authorization does not expire immediately, even after the certificate issued. This means that the owner of an account key can issue certificates for that domain without performing DV challenges until the authorization expires. Because of this, if someone steals your account key, they can issue a certificate for your domain.

Let's Encrypt does this proactively, reusing authorizations even if you request a new one ([source](https://community.letsencrypt.org/t/upcoming-change-valid-authz-reuse/16982)). This has lead to some [interesting conversations](https://news.ycombinator.com/item?id=12787735), but behavior like this is common practice for CAs and is completely valid in CAB forum requirements.

bmw:
> I think cpu reported it was turned on in production at some point in IRC. I don't recall if I tested against production though. It could have been turned off by now for some reason.

It has been turned off. The boulder team told me yesterday they expect to have it live in a couple weeks.

> bmw how does alias'ing usually work? I don't see it in cli.py

You're referring to creating an alias for a subcommand right? The two we have are `auth` and `everything`. If you search for "everything" in `cli.py` (since `auth` has a lot more unrelated matches), you should see how to make this work. Let me know if you have any more questions.

cowlicks:
I think aliasing "register" for "account" is a little out of scope for this PR. So I made another issue here #3816

bmw:
I thought you were still working on this due to the Travis failures (which weren't your fault) and "[revision requested]" in the title. Is this ready for another review?

cowlicks:
@bmw Yeah, this is ready. I was waiting for boulder to turn on the feature. I'll rebase it tonight.

cowlicks:
It looks like I clobbered the unit test that was in `cli_test.py` when I rebased. I don't have a copy of it anywhere so I'll just rewrite it later.

cowlicks:
@bmw ping. This is ready. `git reflog` ftw.

certbot/account.py
```diff
@@ -197,6 +198,18 @@ def save_regr(self, account):
...
 
+    def delete(self, account_id):
+        """Delete registration info from disk
+
+        :param account_id: id of account which should be deleted
+
+        """
+        account_dir_path = self._account_dir_path(account_id)
+        if not os.path.isdir(account_dir_path):
+            raise errors.AccountNotFound(
```
bmw: nit: This line doesn't have test coverage.

certbot/cli.py
```diff
@@ -852,6 +852,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              "should be updated, rather than registering a new account.")
     helpful.add(["register", "automation"], "-m", "--email", help=config_help("email"))
     helpful.add(
+        "register", "--deactivate", action="store_true",
+        help="Irrecvocably deactivate your account")
```
bmw: s/Irrecvocably/Irrevocably

certbot/main.py
```diff
@@ -394,13 +394,44 @@ def _init_le_client(config, authenticator, installer):
     return client.Client(config, acc, authenticator, installer, acme=acme)
 
 
+def _deactivate(config, accounts, messenger):
+    if len(accounts) == 0:
```
bmw: stylistic nit: `if not accounts:` should be used according to our [style guides](https://certbot.eff.org/docs/contributing.html#coding-style).

certbot/main.py
```diff
@@ -394,13 +394,44 @@ def _init_le_client(config, authenticator, installer):
...
+        return "Could not find existing account to deactivate."
+    yesno = zope.component.getUtility(interfaces.IDisplay).yesno
+    prompt = ("Are you sure you would like to irrevocably deactivate "
+              "your account?")
+    wants_deactivate = yesno(prompt, yes_label='Deactivate',
+                             no_label='Abort', cli_flag='--deactivate',
+                             default=True)
+
+    if not wants_deactivate:
+        return "Deactivation aborted."
```
bmw: nit: This line is missing test coverage.

cowlicks:
@bmw I addressed the coverage and style comments. I'll change this to just be an `unregister` verb in the next few days.

cowlicks:
@bmw ready

certbot/tests/main_test.py
```diff
@@ -1,4 +1,5 @@
 """Tests for certbot.main."""
+# pylint: disable=too-many-lines
```
bmw: I created #4069 which should help this significantly.

tests/boulder-integration.sh
```diff
@@ -155,13 +155,6 @@ openssl x509 -in "${root}/csr/cert-p384.pem" -text | grep 'ASN1 OID: secp384r1'
 common auth --must-staple --domains "must-staple.le.wtf"
 openssl x509 -in "${root}/conf/live/must-staple.le.wtf/cert.pem" -text | grep '1.3.6.1.5.5.7.1.24'
 
-# revoke by account key
```
bmw: Were these lines modified by accident? I don't think they should be deleted.

certbot/tests/main_test.py
```diff
@@ -1164,6 +1165,69 @@ def test_update_registration_with_email(self, mock_utility, mock_email):
                             email in mock_utility().add_message.call_args[0][0])
 
 
+class UnregisterTest(unittest.TestCase):
+    def setUp(self):
+        self.patchers = {
```
bmw: In the future, we may want to avoid using `mock.patch` like this. See the comments [here](https://github.com/certbot/certbot/pull/3784#issuecomment-262337637). I don't think this has to change though. We do it in plenty of other places.

certbot/tests/main_test.py
```diff
@@ -1164,6 +1165,69 @@ def test_update_registration_with_email(self, mock_utility, mock_email):
...
+            '_determine_account': mock.patch('certbot.main._determine_account'),
+            'account': mock.patch('certbot.main.account'),
+            'client': mock.patch('certbot.main.client'),
+            'getUtility': mock.patch('certbot.main.zope.component.getUtility')}
```
bmw: This is new and you had no way of knowing, but `certbot.test.util.patch_get_utility` should be used here. It does additional error checking on methods like `yesno` to try and catch bugs.

`test_abort_unregister` will have to be modified slightly. It uncovered a bug which you can see in #4071. For now, don't set the `getUtility` return value. You can set properties of it, but don't change the object itself.

Stuff like this is OK:
```
display = self.mocks['getUtility'].return_value
display.yesno.return_value = False
self.mocks['getUtility'].return_value.foo.return_value = False
```
At least for now, anything like this isn't:
```
self.mocks['getUtility'].return_value = ...
```

certbot/tests/main_test.py
```diff
@@ -1164,6 +1165,69 @@ def test_update_registration_with_email(self, mock_utility, mock_email):
...
+
+        acme_client = mock.MagicMock()
+        self.mocks['client'].Client.return_value = acme_client
+
+        config = mock.MagicMock()
+        unused_plugins = mock.MagicMock()
+
+        res = main.unregister(config, unused_plugins)
+        m = "Could not find existing account to deactivate."
+        self.assertTrue(res == m)
```
bmw: This should probably be `self.assertEqual(res, m)`.

certbot/tests/main_test.py
```diff
@@ -1164,6 +1165,69 @@ def test_update_registration_with_email(self, mock_utility, mock_email):
...
+
+    def test_abort_unregister(self):
+        self.mocks['account'].AccountFileStorage.return_value = mock.Mock()
+
+        util_mock = mock.Mock()
+        util_mock.yesno.return_value = False
+        self.mocks['getUtility'].return_value = util_mock
+
+        config = mock.Mock()
+        unused_plugins = mock.Mock()
```
bmw: nit: I think we should either pass this object into `unregister` or delete it.

certbot/tests/main_test.py
cowlicks: 👍

certbot/main.py
```diff
@@ -406,13 +406,44 @@ def _init_le_client(config, authenticator, installer):
...
+
+    acc, acme = _determine_account(config)
+    acme_client = client.Client(config, acc, None, None, acme=acme)
+
+    # delete on boulder
+    acme_client.acme.deactivate_registration(acc.regr)
+    account_files = account.AccountFileStorage(config)
```
cowlicks: I don't think so. I think I need to use `_determine_account` first to make sure I have the right account.

certbot/tests/main_test.py
cowlicks: I don't like it either, but the alternative was dumping these tests in `TestMain`.

tests/boulder-integration.sh
cowlicks: Yes! I think I got confused in a rebase

certbot/main.py
bmw: Yeah you have to call `_determine_account` first to find the account, but after this line you have two `AccountFileStorage` objects: `account_storage` and `account_files`. What's wrong with the first one?

cowlicks: Ah okay, I did this because `_determine_account` mutates `config` so `account_storage` and `account_files` could possibly be different. I think this might happen when we have multiple accounts, but I'm not sure.

Moving `_determine_account` up to consolidate the variables does not work in the case where there is no account. And fixing `_determine_account` seemed like a tangential rabbit hole.

bmw: Gotcha. Looking in `account.py`, nothing modified in `_determine_account` should matter, however, who knows when that will change in the future. Keeping things as they are now is more robust.

PR #3575, opened 2016-10-03
disallow binary (wheel) install for pycparser

minrk:
pycparser has uploaded a wheel, which is currently causing certbot-auto [to fail](https://community.letsencrypt.org/t/certbot-auto-fails-while-setting-up-virtual-environment-complains-about-package-hashes/20529) due to a hash mismatch:

```
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE
```

because it's downloading the wheel instead of the tarball whose hash is checked. pycparser appears to also have [an issue](https://github.com/eliben/pycparser/issues/147) with the wheel, so rather than add the wheel's hash to the accepted hashes, disallow it, forcing download of the tarball.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -625,7 +625,8 @@ argparse==1.4.0 \
 # This comes before cffi because cffi will otherwise install an unchecked
 # version via setup_requires.
 pycparser==2.14 \
-    --hash=sha256:7959b4a74abdc27b312fed1c21e6caf9309ce0b29ea86b591fd2e99ecdf27f73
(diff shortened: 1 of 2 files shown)


minrk:
I believe the CI failure is due to the broken wheel (https://github.com/eliben/pycparser/issues/148) being pulled in at an earlier point in the CI process.

jmcbri:
+ Much needed.

bmw:
@minrk, thanks for updating the PR.

Test failures aren't your fault. One of our dependencies did a release last night that broke some of our tests. Once this is fixed, I'll kick Travis and merge your PR. Sorry for the trouble!

PR #3581, opened 2016-10-04
Allow tests to pass without dnspython

bmw:
With the addition of using another port in the release script, it allows us to successfully build packages that pass our unit tests. Yay!

### DIFF
--- a/acme/acme/dns_resolver_test.py
+++ b/acme/acme/dns_resolver_test.py
@@ -1,13 +1,17 @@
 """Tests for acme.dns_resolver."""
+import sys
 import unittest
+
 import mock
 
-from acme import dns_resolver
+from acme import test_util
+
 
 try:
     import dns
+    DNS_AVAILABLE = True  # pragma: no cover
 except ImportError:  # pragma: no cover
-    dns = None
+    DNS_AVAILABLE = False
 
 
 def create_txt_response(name, txt_records):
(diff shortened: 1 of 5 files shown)


acme/acme/dns_resolver_test.py
```diff
@@ -21,33 +25,56 @@ def create_txt_response(name, txt_records):
...
 
 
-class TxtRecordsForNameTest(unittest.TestCase):
+@test_util.skip_unless(DNS_AVAILABLE,
+                       "optional dependency dnspython is not available")
+class DnsResolverTestWithDns(unittest.TestCase):
+    """Tests for acme.dns_resolver when dns is available."""
+    @classmethod
+    def _call(cls, name):
+        from acme import dns_resolver
```
ohemorange: Is there a reason to put this in `_call` rather than in `setUp`, or just at the top of the class?

bmw: Running the function we're testing like this is a fairly common thing in this project. The basic idea is to import whatever you're testing in the smallest scope possible to minimize other actions having side effects on what you're testing.

I don't think putting it in `setUp` wouldn't work unless we did something like:
```
def setUp(self):
    from acme import dns_resolver
    self.txt_records_for_name = dns_resolver.txt_records_for_name
```
And then called the function through `self` for all the tests.

We could potentially try and import the name as a class attribute, but I went with `_call` as it's more common in the code base and has the minor benefit I mentioned above.

ohemorange: This `_call` syntax violates that scope rule though. We're using it because every function in here needs `from acme import dns_resolver`, which means the import should probably be at the class level anyway. If you want to shorten the name of the function we're testing, that's fine to leave like this, but re-importing for every function seems weird.

ohemorange: Ok sg I'm convinced

PR #3585, opened 2016-10-05
Fix requirements.txt surgery in response to shipping certbot-nginx

### LINKED ISSUE #3426: Add certbot-nginx to certbot-auto
This requires:
* Adding the Nginx plugin's dependencies to the requirements file
* Fixing up the release script so the hash of the Nginx plugin gets added to the requirements file.

### LINKED ISSUE #3431: Fix requirements file surgery for 0.10.0 release
In response to #3426, this line in `tools/release.sh`:
```
head -n -9 letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt > /tmp/req.$$
```
needs to become
```
head -n -12 letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt > /tmp/req.$$
```

bmw:
Fixes #3431.

While a trivial change, this needs to be merged before we do another release, otherwise, `certbot-auto` will depend on both `certbot-nginx` 0.9.0 and x.y.z for our next release which will obviously cause problems. Wanted to get this out of the way before we forget in case we end up doing a point release.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -176,7 +176,7 @@ if ! wc -l /tmp/hashes.$$ | grep -qE "^\s*12 " ; then
 fi
 
 # perform hideous surgery on requirements.txt...
-head -n -9 letsencrypt-auto-source/pieces/letsencrypt-auto-requirements.txt > /tmp/req.$$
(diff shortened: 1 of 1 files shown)


pde:
LGTM.

Issue #2071, opened 2016-01-03
Allow removing SAN from multidomain certificate when renewing


dusek:
When one has a multidomain certificate with many SANs (Subject Alternative Names), there can come a situation where one wants to remove certain SANs from the certificate when renewing (e.g. the removed domain is no longer used - "retired", or it has become relevant in another multidomain certificate instead; in both cases, to keep things clean, I want to remove the no longer needed SAN from the multidomain certificate).

The problem is when renewing with the reduced set of SANs, letsencrypt does not modify the existing certificate "branch", but creates a new one, usually with the `-0001` suffix. That is quite understandable, as it is a potentially destructive action; however in many cases (e.g. the one I am describing above), it is not what the user wants/expects.

Example:

Multidomain certificate created and repeatedly renewed with `--domains domain.tld,s1.domain.tld,s2.domain.tld,s3.domain.tld` exists in `/etc/letsencrypt/live/domain.tld`.

Now I want to remove e.g. `s3.domain.tld`, so I run with `--domains domain.tld,s1.domain.tld,s2.domain.tld`. Since there is no way currently to specify the intention to modify the SAN list in the existing `/etc/letsencrypt/live/domain.tld` "branch" (is it called "lineage"?), I end up with `/etc/letsencrypt/live/domain.tld` (and same in `archive` and `renewal`) untouched, and instead a new "branch" ("lineage"?) created in `/etc/letsencrypt/live/domain.tld-0001` with the requested domain list.

This presents a few problems:
- I have to modify server config to point to the new certificate path (i.e. `/etc/letsencrypt/live/domain.tld` --> `/etc/letsencrypt/live/domain.tld-0001`)
- I really don't like the "-0001" stuff optically 😊
- it leaves `/etc/letsencrypt/{live,archive,renewal}/domain.tld` on disk with no further purpose

A workaround could be to (neither was tested):
- delete `/etc/letsencrypt/{live,archive,renewal}/domain.tld` before doing the renewal that removes the SAN – that is a little bit disconcerting, as there will be a period where no cert will be on disk (possible downtime, if the renewal should fail, I have to restore working certificates from archive)
- do as above (i.e. let it create the "branch" with `-0001` suffix), and then somehow "rename" `domain.tld-0001` to `domain.tld` (again, possible downtime, possible to mess up, waste of time...)
[...]

Issue #2128, opened 2016-01-09
Explicit specify cert storage location(remove use of automatic -0001 suffixes)


hultqvist:
# Scenario

I start with generating a single cert for "example.com, www.example.com, test.example.com", this is stored in `/etc/letsencrypt/archive/example.com`

Later I generate a new cert with "example.com, www.example.com"
This will be stored in `/etc/letsencrypt/archive/example.com-0001`
# Problems

At this stage it is unclear whether the user intended to remove `test.example.com` or to generate a new cert.

The risks are that one might not notice the suffix and assume that the new certs are in the original location.
I'm not suggesting silently replacing the cert in the original location.

**The location of the generated certs should be known before running the client.**
# Suggestion

With this issue I suggest removal of the automatic suffix generation.

The algorithm that triggers the use of prefix should still be in use but once triggered I suggest the following approach.

When triggered an error will occur.
The error message will instruct that one of the following approaches is required.
1. Indicate that you want to modify the domains in the current cert. This feature is suggested in #2071 
2. Choose a new alias for the storage using an option like `--alias`

With option 2 the use of `--alias example2` would store the certs in `/etc/letsencrypt/archive/example2`

These two options can work independently and combined.
# Examples

If your intention was to remove `test.example.com` then you use 1

If you intention was to generate another different cert then use 2

If you want to further modify the domain list for the cert in the new location you must use both 1 and 2.

bmw:
I actually think this is a duplicate of #1882. Both address the problem of requesting a certificate whose names are a subset of those an existing certificate. With that said, I appreciate you taking the time to write up such a detailed post with both suggestions and examples.

The approach we're likely to take to solve #1882 mirrors how we handled the opposite situation. This is where you are requesting a certificate whose names are a superset of those in an existing certificate. In this case, assuming you provided nothing on the command line, you'll see the following:
```
-------------------------------------------------------------------------------
You have an existing certificate that contains a portion of the domains you
requested (ref: /etc/letsencrypt/renewal/example.com.conf)

It contains these names: example.com, www.example.com

You requested these names for the new certificate: example.com,
www.example.com, test.example.com

Do you want to expand and replace this existing certificate with the new
certificate?
-------------------------------------------------------------------------------
(E)xpand/(C)ancel: 
```

You can also skip this prompt by providing command line flags. If you provide `--expand`, on the command line, the symlinks for the existing certificate will be updated to point to the certificate with the new name.

[...]

PR #3587, opened 2016-10-05
Add --lineage command line option for nicer SAN management.

cowlicks:
cc @shoen, @PaulSD, @yaegashi 

This PR is from some work by @PaulSD mention in #2071  as an alternative way managing SANs on a cert (instead of --expand & --shrink). I have not tested this to see if it works yet.

From the commit:

This allows the user to explicitly select a certificate lineage for
renewal, which may be necessary when removing a domain from a
multi-domain certificate.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -689,6 +689,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Run using interactive dialog menus")
     helpful.add(
         [None, "run", "certonly"],
+        "--lineage", metavar="LINEAGE", dest="lineage", default=None,
+        help="File name of a config file under CONFIG_DIR/renewal/ whose associated certificate "
+             "lineage should be renewed by the 'certonly' or 'run' subcommands.  If not specified, "
+             "the relevant certificate lineage will be automatically determined based on the "
+             "specified domain names and the presence or absence of the --expand or --duplicate "
+             "flags.")
+    helpful.add(
+        [None, "run", "certonly"],
         "-d", "--domains", "--domain", dest="domains",
         metavar="DOMAIN", action=_DomainsAction, default=[],
         help="Domain names to apply. For multiple domains you can use "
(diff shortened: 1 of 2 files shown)


peterthomassen:
While this is suitable to prescribe a lineage for renewal, it does not allow prescribing a lineage name when creating a certificate (because the renewal configuration file does not exist at creation time). However, this would be desirable in some circumstances.

To generalize this so that it also applies at creation time, it would be sufficient if the new command line flag took a lineage name, instead of a configuration filename, i.e. one of the directory names under `{live,archive}`. Usually, this will be a domain name, or a domain name with a `-N` suffix.

This should not create any ambiguities, as the name of the renewal configuration file just the lineage name and the `conf` suffix, i.e. the generalized version is able to provide all the functionality of the current implementation where the filename is passed.

A very similar proposal is `--alias`, see #2128 and https://github.com/certbot/certbot/issues/2128#issuecomment-252109083 therein.

schoen:
@cowlicks, thanks! (there's a "c" in my name, by the way)

@cowlicks @PaulSD, I have a branch where I was working on the same functionality (and even with the same option name) in response to #2071. I was trying to get a few more features in beyond this, such as allowing a wider range of ways to specify a lineage and being more paranoid about what combinations of options makes sense. If I push my work in progress on that branch, would either of you be interested in reviewing it and/or working on it with me?

cowlicks:
@schoen Sure! But I won't have time to contribute until the weekend.

cowlicks:
Also, feel free to close this.

ohemorange:
@schoen is there an issue somewhere discussing your plans for the --lineage option branch?

schoen:
@ohemorange, currently not; it was based on #2071 and some in-person conversations with @pde and @bmw. As we've seen this is a subtle issue, I would be happy to have an issue where we discuss what the solution should look like (and I'm also happy to change existing stuff in my branch if we decide on a different strategy). Would you like to start an issue to track it?

PR #3588, opened 2016-10-05
Document the Nginx plugin release

pde:


docs/using.rst
```diff
@@ -179,13 +179,9 @@ want to use the Apache plugin, it has to be installed separately:
...
-Currently, only the Apache plugin is included in Portage. However, if you
-Warning!
-You can use Layman to add the mrueg overlay which does include a package for the
-Certbot Nginx plugin, however, this plugin is known to be buggy and should only
-be used with caution after creating a backup up your Nginx configuration.
-We strongly recommend you use the app-crypt/letsencrypt package instead until
-the Nginx plugin is ready.
+Currently, only the Apache plugin is included in Portage. However, you
+can use Layman to add the mrueg overlay which does include a package for the
+Certbot Nginx plugin.
```
bmw: I'm not sure we want to make this change. Unless you have information about it that I don't, we have very little about the `letsencrypt-nginx` package included in the mrueg overlay. Has the package been renamed to Certbot? Is it actively maintained? We removed all instructions about doing this from the instruction generator before we launched. I think we should consider doing the same here.

docs/using.rst
```diff
@@ -381,11 +378,15 @@ be on a different computer.
 Nginx
 -----
 
-In the future, if you're running Nginx you will hopefully be able to use this
-plugin to automatically obtain and install your certificate. The Nginx plugin is
-still experimental, however, and is not installed with certbot-auto_. If
-installed, you can select this plugin on the command line by including
-``--nginx``.
+The Nginx plugin is included in Certbot 0.9.0. It should
```
bmw: nit: To me this is a little misleading. The Nginx plugin is not included in version 0.9.0 of the Certbot package. I'm fine leaving this alone if you don't think it's a problem.

If you decide to make the change, you used the same wording in the table above.

docs/using.rst
```diff
@@ -381,11 +378,15 @@ be on a different computer.
...
+work for most configurations. Because it is alpha code, we recommend backing up Nginx
+configurations before using it (though you can also revert changes to
+configurations with ``certbot --nginx rollback``). You can use it by providing
+the ``--nginx`` flag on the commandline.
```
bmw: nit: Did you mean to write "command line" or did you purposefully make it one word? I think two words is more common, however, I have no preference. I was just checking to see if this was a typo.

bmw:
@pde, I pushed a commit addressing a nit I missed before. Previously, the order of plugin descriptions was the same as the order in the table and I think having a consistent ordering here is kind of nice.

If you're fine with this change, merge and I will update the website.

PR #3589, opened 2016-10-06
Match psutil port open checking behavior to that of socket test, and update tests.

ohemorange:


certbot/plugins/standalone_test.py
```diff
@@ -106,13 +106,22 @@ def test_dvsni(self):
         self.assertEqual("tls-sni-01,http-01", self._call("dvsni,http-01"))
 
 
+def get_open_port():
+    """Gets an open port number from the OS."""
+    open_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
```
pde: We could extra defensive and call `open_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)` to reduce the risk that something randomly connects to this test socket, and causes it to be unavailable for the test later. In order to benefit from this, we would [also]( http://man7.org/linux/man-pages/man7/socket.7.html#NOTES) need to set `SO_REUSEADDR` elsewhere in our tests.

pde: I may or may not gotten my understanding of what's going on here right, or explained it right; we should probably sync up our understanding of all of the cases in person. And from a high level this is all pretty corner case-y and not worth doing extra work for. So merging as is :)

PR #3599, opened 2016-10-07
acme: Make it work with python3

mrueg:
str.decode method is not available in python3

```
ERROR: test_txt_records_for_name_with_multiple_responses (acme.dns_resolver_test.DnsResolverTestWithDns)
  File "/usr/lib64/python3.4/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/var/tmp/portage/app-crypt/acme-0.9.1/work/certbot-0.9.1/acme/acme/dns_resolver_test.py", line 46, in test_txt_records_for_name_with_multiple_responses
    self.assertEqual(['response1', 'response2'], self._call('name'))
  File "/var/tmp/portage/app-crypt/acme-0.9.1/work/certbot-0.9.1/acme/acme/dns_resolver_test.py", line 35, in _call
    return dns_resolver.txt_records_for_name(name)
  File "/var/tmp/portage/app-crypt/acme-0.9.1/work/certbot-0.9.1/acme/acme/dns_resolver.py", line 29, in txt_records_for_name
    return [txt_rec.decode("utf-8") for rdata in dns_response
  File "/var/tmp/portage/app-crypt/acme-0.9.1/work/certbot-0.9.1/acme/acme/dns_resolver.py", line 30, in <listcomp>
    for txt_rec in rdata.strings]
AttributeError: 'str' object has no attribute 'decode'
```

### DIFF
--- a/acme/acme/dns_resolver.py
+++ b/acme/acme/dns_resolver.py
@@ -26,5 +26,8 @@ def txt_records_for_name(name):
         logger.error("Error resolving %s: %s", name, str(error))
         return []
 
-    return [txt_rec.decode("utf-8") for rdata in dns_response
-            for txt_rec in rdata.strings]
+    try:
+        return [txt_rec.decode("utf-8") for rdata in dns_response
+                for txt_rec in rdata.strings]
+    except AttributeError:  # Python3.3 or later
+        return [txt_rec for rdata in dns_response for txt_rec in rdata.strings]


bmw:
@mrueg, we run our unit tests on Python 3.3, 3.4, and 3.5 and `txt_rec` here is always `bytes`. Can you tell me what other Python packages you're using? I'm specifically interested in `dnspython`. I'd like like to know if this a problem with our production code or just our tests so we can correctly fix the problem.

mrueg:
@bmw This test is running on python3.4 using http://www.dnspython.org/kits3/1.12.0/dnspython3-1.12.0.tar.gz
Applying the suggested patch fixed the issue for me, I haven't thought about an issue with the dnspython dep.

bmw:
Ah. `acme` requires `dnspython`, not `dnspython3` to use DNS verification ([source](https://github.com/certbot/certbot/blob/master/acme/setup.py#L40)) . The two packages are both imported with `import dns` though, so perhaps we should do something on our end.

The packages have supposedly shared the [same source since 1.12](https://github.com/rthalley/dnspython#new-since-1120), however, there are [problems using dnspython 1.12 with acme on Python 3](https://github.com/certbot/certbot/issues/3603). There are the problems you are describing using `dnspython3 1.12`. Newer versions of either package work correctly though.

mrueg:
@bmw Thanks for further investigation. I would suggest to increase the version of dnspython to 1.14.0 in https://github.com/certbot/certbot/blob/master/acme/setup.py#L40 as https://pypi.python.org/pypi/dnspython3/1.14.0 is a transitional package (while 1.12.0 actually provides the functionality for python3)

bmw:
I described our hesitation to do that and what we might do in #3603.

The problem you had here was resolved in #3618. `dnspython` is an optional dependency that is only used to attempt to validate DNS challenges client side before sending the challenge response to Let's Encrypt. With #3618, we correctly recognize that `dnspython3` doesn't satisfy this optional dependency and properly behave both during testing and production.

The fix will be included in the 0.9.2 release in the next couple days. Thanks for reporting the problem!

Issue #3632, opened 2016-10-13
Don't assume you're not using certbot-auto during tests


bmw:
[This line](https://github.com/certbot/certbot/blob/v0.9.2/certbot/tests/cli_test.py#L110) assumes that if you're running tests, you're not using `letsencrypt-auto`. This is an unnecessary assumption and should be removed, ideally accompanied by more tests that `cli.cli_command` has the correct value.

PR #3633, opened 2016-10-13
Yo dawg, I heard you liked tests so I created a test script for running unit tests on your test farm

bmw:
I had to use `VENV_PATH` because our tests assume that `certbot` was not installed with `certbot-auto`. We shouldn't be making this assumption, however, we've done this in our tests for over 10 months and I don't think fixing this should block our release. I created #3632 to track that issue.


pde:
LGTM

PR #3634, opened 2016-10-14
Release 0.9.3 Option 1 (see #3635)

bmw:
This is what we've done in the other `candidate-0.9.x` branches. After merging this, `master` will include the tag, however, there will be duplicate commits for the changes that were released in `0.9.x`.


pde:
(Taking the other route)

Issue #3638, opened 2016-10-14
Be more conservative when suggesting domains in Apache


bmw:
I did this in Nginx by filtering all names through `certbot.util.enforce_le_validity`. I think we should do the same thing here, otherwise, we suggest the user obtains a certificate for things like wildcards.

dashaxiong:
The `certbot_apache.tests.MultipleVhostsTest.test_get_all_names` expects to find a wildcard domain for "*.blue.purple.com". This test would need to be altered once the names are filtered.

dashaxiong:
The `certbot.util.enforce_le_validity` also rejects domain names which have less than two labels. For instance "ip-172-30-0-17" which is set in `certbot_apache.tests.util.py:get_vh_truth`. This means that I also had to change the number of domains expected in the `certbot_apache.tests.MultipleVhostsTest.test_get_all_names_addrs` once the names are filtered.

dashaxiong:
Please ignore my commit above, I didn't realise this issue was already being dealt with.

PR #3639, opened 2016-10-15
Adding checking name validity to the Apache plugin

TheNavigat:
Closes https://github.com/certbot/certbot/issues/3638

I think it would be cleaner to move the `_get_filtered_names` method to `util.py` or some other file where common functions are available. Please let me know your thoughts on this. If you agree, I'll do that :)

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -219,6 +219,25 @@ def safely_remove(path):
             raise
 
 
+def get_filtered_names(all_names):
+    """Removes names that aren't considered valid by Let's Encrypt.
+
+    :param set all_names: all names found in the configuration
+
+    :returns: all found names that are considered valid by LE
+    :rtype: set
+
+    """
+    filtered_names = set()
+    for name in all_names:
+        try:
+            filtered_names.add(enforce_le_validity(name))
+        except errors.ConfigurationError as error:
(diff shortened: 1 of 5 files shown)


TheNavigat:
I think I'm stuck. I've tried tracing the problem, but the `domains` list isn't empty, and I'm not sure why that error occurs. @bmw, can you please help me with that?

bmw:
@TheNavigat, sorry for the delay in responding here. Holidays coupled with trying to get Certbot 0.10.0 out the door is time consuming.

#4026 should solve problems for you here.

TheNavigat:
@bmw Seems like it's working! Thanks a lot! Can you please review this PR one more time?

certbot/util.py
```diff
@@ -219,6 +219,25 @@ def safely_remove(path):
             raise
 
 
+def get_filtered_names(all_names):
+    """Removes names that aren't considered valid by Let's Encrypt.
+
+    :param set all_names: all names found in the Nginx configuration
```
bmw: nit: With this being a general method now, we shouldn't reference Nginx.

TheNavigat:
@bmw Can you please give this another review? Hopefully a final one :)

bmw:
LGTM. Thanks @TheNavigat.

PR #3652, opened 2016-10-17
Allow certbot to get a cert for default_servers

### LINKED ISSUE #3642: [nginx] Cert for two domains in one virtaulhost fails
When trying to obtain or renew a cert for two domains which is listed in `server_name` in one virtualhost the operation fails.

### Fails
**00-default.conf**
`server_name   _ example.com sub.example.com;`

```
# ./certbot-auto renew --dry-run

Processing /etc/letsencrypt/renewal/example.com.conf
Starting new HTTPS connection (1): acme-staging.api.letsencrypt.org
tls-sni-01 challenge for example.com
tls-sni-01 challenge for sub.example.com
Cleaning up challenges
Attempting to renew cert from /etc/letsencrypt/renewal/example.com.conf produced an unexpected error: nginx restart failed:

nginx: [emerg] a duplicate default server for 0.0.0.0:443 in /etc/letsencrypt/le_tls_sni_01_cert_challenge.conf:36
. Skipping.
```

### Works
**00-default.conf**
`server_name   _ example.com;`

**sub.conf**
`server_name   sub.example.com;`

### Info
```
# ./certbot-auto --version
letsencrypt 0.9.3
# cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)
```

ohemorange:
Fixes #3642.


ohemorange:
We should add a couple test cases to the compatibility tests. One is in the issue linked above. Another is:

```
server {
  server_name test.example.org;
  ssl on;
  ssl_certificate snakeoil.cert;
  ssl_certificate_key snakeoil.key;
}
```

certbot-nginx/certbot_nginx/tls_sni_01.py
```diff
@@ -141,7 +140,11 @@ def _make_server_block(self, achall, addrs):
         document_root = os.path.join(
             self.configurator.config.work_dir, "tls_sni_01_page")
 
-        block = [['listen', ' ', str(addr)] for addr in addrs]
+        no_default_addrs = [copy.deepcopy(addr) for addr in addrs]
```
ohemorange: It seemed unpythonic to have a separate to_string method... I suppose JSON has this though.

bmw: I agree it's a bit unpythonic. Googling around, I couldn't find a common pattern used for controlling the output of `str`. This seems a bit silly to me as it doesn't seem that unusual to need display an object in different ways depending on the context.

I think the benefits of writing a `<insert_alternative_string_method_name_here>` method over the current solution is:

* Less code
* Faster
* Can easily be reused

With all that said, I don't think it's clear what's best here and if you want to use your current approach...

Looks like you coded it up before I could respond :smile:

Issue #3615, opened 2016-10-11
Cert Management Improvement Project (C-MIP)


ohemorange:
Based on previous discussions (#3587, #2071, #338), we should have an improved overall strategy for managing certs. This looks something like:

Step 1. list-certs command
Step 2. Specify canonical `{archive, live}` directories in conf file
Step 3. --use-lineage-name <lineage-name>
Step 4. --set-lineage-name <lineage-name>
Step 5. --use-lineage-name <wrong name or domain names or whatever>, "did you mean" based on filenames and such if ambiguous

Discussion items:
A. What should we call the canonical selector that I have temporarily termed "lineage-name"?
B. Where does this canonical name live?
Options for this include:
- as a line item in the conf file, whose name does not matter
- as the filename of the conf file, and document this choice.

I am personally for the latter, based on @schoen's observation that the file system then ensures uniqueness.

cc: @pde @bmw @cowlicks @PaulSD @yaegashi

schoen:
I like calling it a lineage name and having it be the name of the file in `/etc/letsencrypt/renewal`, less the `.conf` suffix. (But I think that's the status quo, as nearly as we have one.)

I was hoping `--use-lineage-name` could just be called `--lineage`, although as @pde and I mentioned earlier today, it might be extra-confusing that the existing `-d` can mean both "apply this action to the lineage that has (at least) these domains" and "make this lineage have these domains". So it may be helpful to be sure `--use-lineage-name` and `--set-lineage-name` are very clearly different from one another.

ohemorange:
I would like to amend step 2 to read:

Specify archive directory along with existing `{cert, privkey, chain, fullchain}` files in conf file.

This will remove dependence on lineage name in filepaths while maximizing the flexibility users have for specifying locations of files they plan to serve.

ohemorange:
Update to step 2: Print a helpful error message if we are having trouble with the archive directory.

From @pde:

> 5:43:41 PM - pdeee: […]  think all of these cases can be handled by this algorithm:
> 5:43:49 PM - pdeee: if your live directory doesn't exist, make it
> 5:44:00 PM - pdeee: if your live directory has broken or outdated links, fix them
> 5:44:52 PM - pdeee: (using the relpath function Brad just linked to)
> 5:47:07 PM - pdeee: Probably if the archive directory in the renewal conf file has been edited, and the thing it points to doesn't look like an archive, we should just error out rather than trying to fix things
[...]

PR #3661, opened 2016-10-19
Specify archive directory in renewal configuration file

ohemorange:
Step 2. of C-MIP (#3615)


schoen:
In the C-MIP issue #3615, step 2 is given as "Specify canonical `{archive, live}` directories in conf file", but it looks like this PR only changes the treatment of `archive` and not `live`. Could you explain a little bit about how `archive` and `live` are handled before and after this PR?

ohemorange:
@schoen that is correct, as per this comment:https://github.com/certbot/certbot/issues/3615#issuecomment-253083708

Before, archive was inferred as a relative location from the renewal conf file. Now, archive is written out to the renewal conf file, read and used if it is there, and if it is not there, inferred from the renewal conf file as before.

certbot/main.py
```diff
@@ -16,6 +16,7 @@
 import certbot
 
 from certbot import account
+from certbot import cert_manager
```
schoen: In the past we used "cert manager" to refer to @jdkasten's tool including a UI for cert management tasks. I'm just wondering about C-MIP nomenclature for a library of methods that can perform such tasks vs. a UI for invoking them.

certbot/storage.py
```diff
@@ -204,7 +207,7 @@ class RenewableCert(object):  # pylint: disable=too-many-instance-attributes
         renewal configuration file and/or systemwide defaults.
 
     """
-    def __init__(self, config_filename, cli_config):
+    def __init__(self, config_filename, cli_config, update_symlinks=False):
```
schoen: As above, how come `update_symlinks` belongs in the constructor? Is there something particular about instantiating `RenewableCert` that should make this action sometimes get invoked here as opposed to be (perhaps immediately) calling a method on the resulting instantiated object? (I'm just wary of complicating the API.)

certbot/storage.py
```diff
@@ -759,7 +783,7 @@ def new_lineage(cls, lineagename, cert, privkey, chain, cli_config):
         """
 
         # Examine the configuration and find the new lineage's name
-        for i in (cli_config.renewal_configs_dir, cli_config.archive_dir,
+        for i in (cli_config.renewal_configs_dir, cli_config.default_archive_dir,
```
schoen: Does this make it impossible to specify an archive directory when creating a new lineage? So would the only path to creating a new lineage with a specified archive directory be to first created it and then update the symlinks with `_update_symlinks()` after the lineage is finished being created?

schoen:
@ohemorange, and the live directory is taken from the path of the (currently four) live objects that are specified in the renewal configuration file's main section, so doesn't need to be specified explicitly, right?

For renewal there is an assumption that those four are archived in the same archive directory, in order to find the latest one; overall, you intend to maintain that assumption while allowing the location of that directory to be specified, right?

certbot/cert_manager.py
```diff
@@ -0,0 +1,22 @@
...
+
+    :param config: Configuration.
+    :type config: :class:`certbot.interfaces.IConfig`
+
+    """
+    renewer_config = configuration.RenewerConfiguration(config)
+    for renewal_file in renewal.renewal_conf_files(renewer_config):
+        storage.RenewableCert(renewal_file,
+            configuration.RenewerConfiguration(renewer_config),
+            update_symlinks=True)
```
ohemorange: It's in the constructor because the constructor checks that the symlinks referenced in the config file are valid. So, the two options are to either put it in the constructor or to check if the constructor fails, call a class method to update the symlinks, then call the constructor again. Or, to always preface a constructor call with a class method call, which is possibly a more complex interface.

certbot/main.py
ohemorange: Well, we don't have a UI anymore, so maybe the point is moot? My idea was that these are the tasks that let the use perform certificate management. Perhaps they could be selected from a list of possible tasks, if we really want a UI. Any good suggestions for alternate naming schemes?

certbot/storage.py
ohemorange: Being required to call a method immediately after instantiating an object seems like a more complicated API. The idea of putting it in the constructor is that sometimes you're asked to construct an object from a renewal file that is currently not in a valid state, and to properly construct it you'll have to go in and fix that state.

certbot/storage.py
ohemorange: No, it doesn't make it impossible - it was never possible to begin with. We could additionally add a command line flag to specify `archive_dir`, but otherwise, yes, that is how one would currently do it.

ohemorange:
> @ohemorange, and the live directory is taken from the path of the (currently four) live objects that are specified in the renewal configuration file's main section, so doesn't need to be specified explicitly, right?

Oof. Kind of. So, it's defined in `configuration.py` alongside `archive_dir`, now `default_archive_dir`. But that one is only used in `new_lineage`. Then we use the path of the cert to get the object's `live_dir`, which we don't use ourselves at all but do pass to `renew_hook`. This is maybe problematic but I don't think I want to be touching the `renew_hooks` api until something comes up.

> For renewal there is an assumption that those four are archived in the same archive directory, in order to find the latest one; overall, you intend to maintain that assumption while allowing the location of that directory to be specified, right?

Yes, exactly.

bmw:
I just want to add to the conversation here that I don't think removing `dialog` means we can't have a certificate manager as part of Certbot. It's not like Certbot is non-interactive now. We just have to use a text interface rather than Ncurses which I think is a minimal loss in usability.

Issue #3324, opened 2016-07-24
DialogError should come with --text instructions


jsha:
During my workshop today, someone got a DialogError. It wasn't immediately clear what was wrong, but the message was pretty cryptic. We should give instructions to run with `--text` whenever a DialogError is thrown.

Alternately, maybe we should remove the dialog dependency entirely?

bmw:
Removing dialog fixes errors like #3511, simplifies our code, makes shipping Certbot for Python 3 easier, etc. The only downside is a potential usability loss. I know @pde and @schoen feel that dialog provides a significantly better UI for the user in some scenarios. For example:

Dialog:
![screenshot from 2016-09-19 12-33-59](https://cloud.githubusercontent.com/assets/6504915/18646217/6a8701fa-7e65-11e6-969a-8ec3df2e8992.png)

Text:
```
Which names would you like to activate HTTPS for?
-------------------------------------------------------------------------------
1: example.org
2: example.net
3: example.com
4: www.example.org
5: www.example.net
6: www.example.com
-------------------------------------------------------------------------------
Select the appropriate numbers separated by commas and/or spaces (Enter 'c' to
cancel):
```
Do you guys feel that benefits like this are worth keeping dialog around?

bmw:
Another argument for removing dialog. #2025 has been around for a while and we don't know the cause or how to fix it (it is almost certainly a longstanding bug in `dialog`).

jsha:
I think if the name selection interface is the only argument for keeping dialog, it is not a strong enough argument. We can make the `--text` version more convenient, for instance by treating "enter" as "select all."

ohemorange:
Hey, I think the time has come to remove dialog. Anyone opposed?

bmw:
@ohemorange, as I mentioned above, @schoen and @pde were the main people opposed when this was discussed in person. If you work on this, you may want to:

1. Make sure they're okay with such a change being merged.
2. Chat with me about the best way to do this. I think we can dramatically simplify our code by removing `dialog`. All of this work could be done in different PRs, but I think removing `dialog` could remove ~1000 LOC from the project.

jsha:
I believe when we talked to @pde in person about this a few weeks ago he was convinced that it made sense as a post-0.9.0 change. But probably worthwhile to get a final thumbs-up. :-)

pde:
Yes we can do this for 0.10.0 if we make "all" the default answer to the domain selection question.

pde:
[...]

PR #3665, opened 2016-10-20
Remove the curses dialog, thereby deprecating the --help and --dialog command line options

ohemorange:
As per #3324.


ohemorange:
Coverage went down because I removed covered lines, and also because I fixed some of the testing code that runs when the tests break.

certbot/cli.py
```diff
@@ -677,7 +666,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         "e.g. -vvv.")
     helpful.add(
         None, "-t", "--text", dest="text_mode", action="store_true",
-        help="Use the text output instead of the curses UI.")
+        help="Use the text output.")
```
jsha: Shouldn't this also say "Deprecated - this has no effect?"

docs/contributing.rst
```diff
@@ -246,7 +246,7 @@ configuration checkpoints and rollback.
 Display
 ~~~~~~~
 
-We currently offer a pythondialog and "text" mode for displays. Display
+We currently offer a "text" mode for displays. Display
```
jsha: This should probably say "We currently only offer.."

certbot/cli.py
ohemorange: That it should!

docs/contributing.rst
ohemorange: Very true.

docs/cli-help.txt
```diff
@@ -22,14 +22,11 @@ optional arguments:
   -v, --verbose         This flag can be used multiple times to incrementally
                         increase the verbosity of output, e.g. -vvv. (default:
                         -2)
-  -t, --text            Use the text output instead of the curses UI.
```
bmw: This is another of our "only modify during releases" files, except this one doesn't have a test to catch the problem. The reason we don't modify this is we use this file to generate our documentation at https://certbot.eff.org/docs/using.html#command-line and we want to keep this information synced with the current release.

certbot/cli.py
ohemorange: Oh right now I remember, I left this in on the off chance we eventually implement something else for the `IDialog` interface. I'll make this clear though.

certbot/cli.py
```diff
@@ -686,7 +675,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
               "which ones are required if it finds one missing")
     helpful.add(
         None, "--dialog", dest="dialog_mode", action="store_true",
-        help="Run using interactive dialog menus")
+        help="Deprecated - this has no effect")
```
bmw: I think we should suppress help output here rather than continuing to print messages about a deprecated flag. We have a utility function for deprecating arguments called [add_deprecated_argument](https://github.com/certbot/certbot/blob/master/certbot/util.py#L369) that does this. I think the only reason we may not want to use this is it will cause a warning to be printed whenever the flag is used. This might be overkill for flags like `--text` that might be used in renewal scripts.

If we don't want to print a warning, we should use `argparse.SUPPRESS` here as it's done in `add_deprecated_argument`.

ohemorange: I wonder if literally anyone ever used the `--dialog` flag, I have no problem deprecating that one properly.

As for `--text`, we have a whole bunch of tests that use it, I think suppressing it is the right move.

docs/cli-help.txt
ohemorange: Ok it seems the correct choice here is to not generate the documentation on that website based on master, but instead to generate it based on the release branch. Is there already an issue for that, or a reason why we don't already do that?

ohemorange:
Re: `certbot/interfaces.py` - I'm fine keeping it in the interface and continuing to ignore it, on the off chance that this breaks some closed-source plugin.

docs/cli-help.txt
bmw: We often change the documentation for reasons other than a new release. We could delay publishing changes until a release has been done, but blocking all improvements to our documentation to keep this file in lockstep seems undesirable to me.

certbot/cli.py
bmw: I agree. Let's deprecate `--dialog` and suppress help output from `--text`.

bmw:
I personally don't think reverting changes to `certbot/interfaces.py` and the `notification` methods in `certbot/display/util.py` is necessary. I was just bringing that up as an example of the kind of thing we need be careful about. I looked into it though and it doesn't seem to be a problem.

While it's possible there is some closed source plugin out there, none of us have ever heard about it. I personally think we should take advantage of this and use it as a chance to clean up the API a bit. I don't have a strong opinion here though, so feel free to revert these changes if you wish.

docs/cli-help.txt
ohemorange: Are these documentation changes urgent enough that they can't wait for a new release? If we release regularly, this shouldn't be a problem.

bmw: They are rarely urgent, however, they do offer a better experience for the user. Why should we tie our hands and force ourselves to sit on these changes until we do a release?

And while I wouldn't call them urgent, there are definitely times we want to make changes outside of a release. For example, the stuff that I've been working on involves doc changes and needs to go live almost immediately so it is up before super secret project that I'm not sure I'm allowed to talk about publicly involving encryption, pipes, and Noah the hacker is released. Another example of changes like this is changes in OS packaging such as when a new version of a distro is released.

bmw: Yeah I think there are a number of things we could do to make this a little nicer for developers. I think it could get a bit tricky though.

And no need to make a separate PR. These changes happen automatically in our [release script](https://github.com/certbot/certbot/blob/master/tools/release.sh#L149).

ohemorange: Cool! I also modified `docs/contributing.rst`, should I make a separate PR for that one?

bmw: Oh sorry. I missed that one. Yeah I suppose you should make a separate PR that we can merge during the release.

ohemorange:
All requested changes made!

PR #3669, opened 2016-10-21
Add list-certs command

ohemorange:
Step 1 of C-MIP (#3615). Relies on #3661.


ohemorange:
Please note that this branch is based on #3661 - new commits start at 67c1889.

certbot/storage.py
```diff
@@ -256,8 +259,26 @@ def __init__(self, config_filename, cli_config):
...
+        if update_symlinks:
+            self._update_symlinks()
         self._check_symlinks()
 
+    def __str__(self):
+        """Returns a string representation of the object."""
+        # import ipdb; ipdb.set_trace()
+        outdict = {"Lineage": self.lineagename,
+                   "Domains": " ".join(self.names())}
+        return "RenewableCert(%s)" % str(outdict)
```
schoen: If this is used as the basis for the `_report()`, which apparently it is, the user probably won't need to see the text "RenewableCert" (which is more of a Certbot internal thing and would probably be confusing to the user). Conversely, I think it would be useful to include the expiration date in the output because that's something users commonly care about intensely.

ohemorange: I was trying to keep the representation pythonic, but that makes sense. I like the expiration date idea! I should mention more broadly that I'm looking for further suggestions of what to put in here.

pde: I think the most UNIX-y output data would be a table that people can parse with `cut` and/or `awk` by default.

schoen: I think it would be very valuable to have a machine-readable output and I think people would use it a lot. It could be tab-delimited and maybe include

`lineage	expiry	domain:domain:domain:domain`

or maybe also some representation of the issuer and subject keys?

There could be separate human-readable and machine-readable versions of the output.

For the version that you show above, I would get rid of the quotation marks and also provide a human readable summary of the time to expiry, maybe like

    Valid until: Jan 24 21:28:00 2017 GMT  (89 days from now)

It would also be helpful to have a version for certificates that have already expired, maybe like:

    Valid until: Jan 24 21:28:00 2016 GMT  (EXPIRED)

schoen: I'm also curious what date-formatting method uses "GMT" instead of "UTC" as the time zone designator!

ohemorange: grabbed the date formatting from `openssl x509 -inform pem -noout -text -in`

what's a possible use case for machine-readable? can we default to human-readable, and then also have a machine-readable flag?

I really like the (89 days from now / EXPIRED) idea!

schoen: > grabbed the date formatting from openssl x509 -inform pem -noout -text -in

Huh! That's wacky of the OpenSSL project...

> what's a possible use case for machine-readable?

People writing their own renewal scripts (maybe because they have criteria or associated actions that aren't well-covered by the behavior of `certbot renew`) or writing pre- and post-hooks for renewal that need to know more about what certificates exist, again because they want to take some action based on what happens or what exists. (For example, maybe something about notification to a human administrator, or something about synchronization between machines in a cluster -- like rsyncing only the certificates that changed in the course of the renewal process, perhaps?)

> can we default to human-readable, and then also have a machine-readable flag?

Yep, I think that would be a good choice.

> I really like the (89 days from now / EXPIRED) idea!

There are probably a lot of library approaches to doing this -- for example if you have a pair of `datetime` objects you can subtract them and then refer to the `.days` attribute of the resulting `timedelta` objects, but if you can get the expiry time as Unix time, you can say something like

```
diff = expiry - time.now()
if diff <= 0:
    return "EXPIRED"
elif diff < 86400:
    return "expiring today"
elif diff < 86400*2:
    return "1 day"
else:
    return "{0} days".format(diff % 86400)
```

schoen: ... actually "tomorrow" would probably be better than "1 day".

ohemorange: Actually, machine-readability is a larger change - this is currently being printed as a display notification, which seems correct, but if we do that then it imposes line breaks, which would likely be annoying to parse anyway. Maybe we leave off machine readability for now, and if user requests come in, we find a way to add it in.

schoen: If Certbot doesn't have the curses interface anymore, is there a clear reason why it has to retain the `IDisplay` abstraction? Alternatively, could `IDisplay` be extended with a new method to output unformatted text, and then have the machine-readable output call that method?

ohemorange: Plugins that we did not write could be using it.

I'd also prefer to hold off on the machine-readable part until we get outside requests/ use cases for it, unless you could point to any we've seen go by. The idea for this is to give users control over the interactions of their certs and their domains, both for selecting lineages to use and controlling which domains are in which cert. Which a human-readable list-certs will address, but a machine-readable won't.

bmw:
Looks like there's merge conflicts with `master` since #3661 was merged.

ohemorange:
@bmw no more conflicts!

bmw:
Great. I'll review today.

certbot/tests/cert_manager_test.py
```diff
@@ -8,7 +8,6 @@
 import configobj
 import mock
 
-from certbot import configuration
 from certbot.storage import ALL_FOUR
 
 class CertManagerTest(unittest.TestCase):
```
bmw: I don't care if you make this change, but you can also write a method like:

```
self._list_certs(*args, **kwargs):
    from certbot.cert_manager import list_certs
    return list_certs(*args, **kwargs)
```

So you don't have to import `cert_manager` at the top of every test. This is arguably even slightly better as you can't modify `cert_manager` other than through `mock` until the very last second before you call it. You don't have to fool around with `*args, **kwargs` if you don't want. If you make a separate class to test each function, you could even call this method `_call`.

certbot/cli.py
```diff
@@ -69,6 +69,7 @@
   config_changes       Show changes made to server config during installation
   update_symlinks      Update cert symlinks based on renewal config file
   plugins              Display information about installed plugins
+  list_certs           Display information about certificates that Certbot knows about
```
ohemorange: I like both of these suggestions! I'll go with `certificates`, and then "Display information about certs configured with Certbot".

certbot/storage.py
```diff
@@ -264,6 +264,11 @@ def __init__(self, config_filename, cli_config, update_symlinks=False):
         self._check_symlinks()
 
     @property
+    def target_expiry(self):
+        """Returns at datetime object representation of the expiry of the current target cert"""
```
ohemorange: (Y)

certbot/cert_manager.py
```diff
@@ -20,3 +30,83 @@ def update_live_symlinks(config):
...
+            expiration_text = "EXPIRED"
+        else:
+            diff = cert.target_expiry - now
+            if diff.days == 1:
+                expiration_text = "1 day"
+            elif diff.days < 1:
+                expiration_text = "under 1 day"
+            else:
+                expiration_text = "{0} days".format(diff.days)
+        valid_string = "{0} ({1})".format(cert.target_expiry, expiration_text)
```
ohemorange: This does print a timezone? `Valid Until: 2017-01-24 20:33:00+00:00`, the `+00:00` is the UTC part.

certbot/cert_manager.py
```diff
@@ -20,3 +30,83 @@ def update_live_symlinks(config):
...
+        if parsed_certs:
+            notify("Found the following certs:")
+            notify(_report_human_readable(parsed_certs))
+        if parse_failures:
+            notify("\nThe following renewal configuration files "
+               "were invalid:")
+            notify(_report_lines(parse_failures))
+
+    if config.quiet:
+        return
```
ohemorange: That's strange, I had this because that's how `renewal.py` has it. Does that include output sent to `IDisplay`?

certbot/cert_manager.py
```diff
@@ -20,3 +30,83 @@ def update_live_symlinks(config):
...
+            renewal_candidate = storage.RenewableCert(renewal_file,
+                configuration.RenewerConfiguration(lineage_config))
+        except Exception as e:  # pylint: disable=broad-except
+            logger.warning("Renewal configuration file %s produced an "
+                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.debug("Traceback was:\n%s", traceback.format_exc())
+            parse_failures.append(renewal_file)
+            continue
+
+        if renewal_candidate is None:
```
ohemorange: You're right, it can't be `None`.

certbot/cert_manager.py
bmw: Yes that includes `IDisplay`. If `-q/--quiet` is provided, `zope.component.getUtility(interfaces.IDisplay)` returns a `NoninteractiveIDisplay` constructed [here](https://github.com/certbot/certbot/blob/master/certbot/main.py#L761).

I think we do this in `renewal.py` because we're using `print` not `IDisplay`.

ohemorange: `--quiet` doesn't send `print` to `/dev/null` then? if so, we'll want to make sure we have this for a machine-readable version of printing this output.

ohemorange:
> One higher level thought I had is I wonder if we should also be printing the path to the fullchain as well. We tell users they have a cert, but don't tell them where it's at.
I like this.

Any personal opinions on what to call lineage externally? Over in #3615 I mentioned alias, certificate chain, and certificate group. We should decide in this PR.

certbot/cert_manager.py
bmw: You're right. Ignore me.

bmw:
I/we should bring this up in the tech projects meeting today. We're also going to talk about the unicode/punycode question that @schoen raised there as well.

I also thought about this a little more and I'm not sure what's best for us to provide the path for. Potential reasonable options:

* Fullchain (and key?)
* All four (cert, privkey, chain, and fullchain)
* Live directory (if there is still a requirement that all four symlinks are in the same directory)

certbot/cert_manager.py
```diff
@@ -20,3 +30,83 @@ def update_live_symlinks(config):
         storage.RenewableCert(renewal_file,
             configuration.RenewerConfiguration(renewer_config),
             update_symlinks=True)
+
+def _report_lines(msgs):
+    """Format a results report for a category of single-line renewal outcomes"""
+    certinfo = [str(msg) for msg in msgs]
```
ohemorange: Going to keep it anyway though actually.

certbot/cert_manager.py
bmw: That's fine. Appending to the list we made at the top of function really should never fail. If it does though, we'll incorrectly report the problem as the user's renewal conf file rather than a bug in our code or Python.

Like I said, this should never happen though, so I don't really have a problem if you take this approach.

ohemorange:
Updates from IRL conversation:

- `Lineage` --> `Certificate Name`, as noted in #3615
- Add a `Certificate Path` line

ohemorange:
Some Travis problems but maybe it'll work anyway; closing and reopening.

ohemorange:
@pde, @bmw, what are the blockers to merging this PR?

bmw:
LGTM. Nice work @ohemorange.

PR #3673, opened 2016-10-21
Add standalone option for running a TLS server when validating HTTP-01 challenges

### LINKED ISSUE #3645: Standalone option for running a TLS server when validating HTTP-01 challenges
Boulder will follow insecure redirects when validating a http-01 challenge. Hence, redirects to insecure TLS domains will be followed.

Since the `ServerManager` has all the certificates needed to run a TLS server already, there should be an option to use TLS when validating a HTTP-01 challenge.

benileo:
This PR, add an option to the CLI to run the standalone HTTP-01 server using TLS with a self signed certificate. The name of the option is `--http-01-use-tls`. 

Note for reviewers, I had to add a `Content-Length` header to all requests served by the `HTTP01ResponseHandler`. Without this header, http requests are malformed and TLS connections will not be terminated properly. 

Fixes #3645

### DIFF
--- a/certbot/interfaces.py
+++ b/certbot/interfaces.py
@@ -238,6 +238,12 @@ class IConfig(zope.interface.Interface):
         "This only affects the port Certbot listens on. "
         "A conforming ACME server will still attempt to connect on port 80.")
 
+    http01_use_tls = zope.interface.Attribute(
+        "Runs the standalone server with a self signed certificate. "
+        "Useful for when you need to validate an http-01 challenge with "
+        "an https endpoint or if you have an upstream proxy redirecting "
+        "all http traffic to https.")
+
 
 class IInstaller(IPlugin):
     """Generic Certbot Installer Interface.
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -319,7 +319,7 @@ class HTTP01Response(KeyAuthorizationChallengeResponse):
         uri = chall.uri(domain)
         logger.debug("Verifying %s at %s...", chall.typ, uri)
         try:
-            http_response = requests.get(uri)
+            http_response = requests.get(uri, verify=False)
         except requests.exceptions.RequestException as error:
             logger.error("Unable to reach %s: %s", uri, error)
             return False
--- a/acme/acme/challenges_test.py
+++ b/acme/acme/challenges_test.py
@@ -216,7 +216,7 @@ class HTTP01ResponseTest(unittest.TestCase):
         mock_get.return_value = mock.MagicMock(text=validation)
         self.assertTrue(self.response.simple_verify(
             self.chall, "local", KEY.public_key()))
-        mock_get.assert_called_once_with(self.chall.uri("local"))
+        mock_get.assert_called_once_with(self.chall.uri("local"), verify=False)
 
     @mock.patch("acme.challenges.requests.get")
     def test_simple_verify_bad_validation(self, mock_get):
@@ -232,7 +232,7 @@ class HTTP01ResponseTest(unittest.TestCase):
                   HTTP01Response.WHITESPACE_CUTSET))
         self.assertTrue(self.response.simple_verify(
             self.chall, "local", KEY.public_key()))
-        mock_get.assert_called_once_with(self.chall.uri("local"))
+        mock_get.assert_called_once_with(self.chall.uri("local"), verify=False)
 
     @mock.patch("acme.challenges.requests.get")
     def test_simple_verify_connection_error(self, mock_get):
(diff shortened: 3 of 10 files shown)


cowlicks:
Probably related to all recent service outages. You should try again once everything is back up.

certbot/interfaces.py
```diff
@@ -238,6 +238,12 @@ class IConfig(zope.interface.Interface):
         "This only affects the port Certbot listens on. "
         "A conforming ACME server will still attempt to connect on port 80.")
 
+    http01_use_tls = zope.interface.Attribute(
+        "Runs the standalone server with a self signed certificate. "
+        "Useful for when you need to validate an http-01 challenge with "
+        "an https endpoint or if you have an upstream proxy redirecting "
+        "all http traffic to https.")
```
patf: Oh, in that case, I might have misunderstood the purpose of this PR (or at least some of it). I assumed this was for something like a CDN that enforced TLS for backend connections, where the CDN itself still listens and responds on port 80.

`http-01` always works via HTTP on port 80, at least for the initial request. HTTP redirects are allowed. HTTPS-only hosts would fail validation. Here's the relevant spec section:

>    As a domain may resolve to multiple IPv4 and IPv6 addresses, the
   server will connect to at least one of the hosts found in A and AAAA
   records, at its discretion.  **Because many webservers allocate a
   default HTTPS virtual host to a particular low-privilege tenant user
   in a subtle and non-intuitive manner, the challenge must be completed
   over HTTP, not HTTPS.**
> https://tools.ietf.org/html/draft-ietf-acme-acme-01#section-7.2

benileo: Interesting. Thank you for the reference from the spec. In that case, I guess this PR is void. Boulder will follow http -> https redirects and will validate at an https endpoint. So perhaps boulder's implementation is flawed?

benileo: From line 198 in `va.go`
```
tr := &http.Transport{
		// We are talking to a client that does not yet have a certificate,
		// so we accept a temporary, invalid one.
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
		// We don't expect to make multiple requests to a client, so close
		// connection immediately.
		DisableKeepAlives: true,
		// Intercept Dial in order to connect to the IP address we
		// select.
		Dial: dialer.Dial,
	}
```

It says here that it will accept invalid certificates from the domain being validated...?

benileo: Right, and if I have a proxy that redirects all http to https traffic (which is not un-common), then there is no entry point for upstream https servers running the acme client to validate an http-01 challenge, unless they are validating the challenge on https. Since I'm also doing tls-sni based routing (and can't use tls-sni-01), I've expressed this on the acme mailing list - which I believe you gave some feedback on. There is no way for me to use the acme protocol to install certificates.. :face_with_thermometer:

patf: I'm not sure I understand this correctly. If you have a proxy that redirects http/80 to https/443 (via HTTP 301/302), you're good. The upstream HTTPS server may indeed operate as a HTTPS server, that's between you and the proxy. I'll try to write this as a sequence of requests:

1. ACME client asks ACME server to validate domain
2. ACME server connects to your (reverse) proxy on port 80, requesting `http://example.com/.well-known/acme-challenge/{token}`
3. Reverse proxy responds with HTTP 301 to `https://example.com/.well-known/acme-challenge/{token}`
4. ACME server connects to reverse proxy on on port 443 via TLS, requesting `https://example.com/.well-known/acme-challenge/{token}`
5. Reverse proxy connects to upstream/backend server, requesting `/.well-known/acme-challenge/{token}`. This can be either HTTP or HTTPS, on any port - the ACME server wouldn't known or care.
6. Upstream server responds with challenge token, reverse proxy forwards that response, challenge passes, all is well.

This is how e.g. people use Let's Encrypt behind CloudFlare today. Is this what you were referring to, or are we talking about different things? 😇

benileo: You are right on point and we are talking about the same thing. I'll start from where I differ a little bit, although I think your solution might be the right one, thank you.

Step 5: Reverse proxy (Haproxy) uses TLS-SNI based routing to route requests based on server name indication to the correct upstream/backend server. The upstream/backend servers are running nginx, but they also have the ACME client installed. I have a bootstrap script that calls the ACME client before it attempts to start nginx. I use the standalone option with http-01 to try and obtain these certificates and once I have them, I start nginx with the obtained certificates. 

I think you have solved my problem at @patf, I need to first start nginx running as an https reverse proxy so that when the https request arrives on the server, I simply proxy it to the certbot client running locally on port 80 serving the http-01 challenge. 

That was a good exploration into the `certbot` code base and I'm glad there is a solution to what I'm trying to do. Thanks again.

Issue #3683, opened 2016-10-24
Cryptography dependency makes installation/testing of certbot non-trivial, documentation should address this


ClifHouck:
Unfortunately, the `cryptography` pip module places some burden on the user to configure their system properly before it will build/install cleanly via `pip install`. I ran into this while trying to run automated tests via `tox` on a fresh system.

Would be nice to call this out in installation/contributing documentation.

SwartzCr:
I'm going to close this much like @jsha closed the associated PR - but feel free to reopen or create a new issue if you think our documentation is still lacking

Issue #2406, opened 2016-02-09
letsencrypt-auto --help should not run "apt-get install -y ..." automatically


pmorch:
I just had a machine trashed.

Running `./letsencrypt-auto --help` ran `apt-get install -y ...` automatically, which, among other things updated `libc`, trashing the entire machine to a point where re-installation was the easiest path. Sure the cause was a misconfiguration of `/etc/apt/sources` performed by us months ago, but still, `./letsencrypt-auto --help` shouldn't trash the machine because of that.

I suggest removing the `-y` switch so that the user is prompted for confirmation before apt-get does its thing, and/or a `letsencrypt-auto --yes-to-apt` switch or something. If one were to argue that the '-auto' in `letsencrypt-auto` implies automatic behaviour, then please provide a `letsencrypt-non-auto` or something *and use __that__ as the default in the documentation*.

Please don't perform an operation that can trash my entire machine by running `./letsencrypt-auto --help`. Nobody expects that. 

Or at the very least check that this doesn't install any packages you didn't anticipate. Perhaps running `apt-get install -s $packages` and testing the output. If only packages from an expected white-list show up, then run `apt-get install -y $packages` otherwise omit the -y and do `apt-get install $packages`, so the user is prompted for confirmation.

bmw:
I'm very sorry that you just had a machine trashed. `letsencrypt-auto` is supposed to be a temporary solution while we work on distro packing. With that said, the script can certainly be improved.

I feel that this is largely a duplicate of #1903, focusing more on the issue of using `apt-get install -y` in the script. If, however, you feel that this is a different issue, I'm happy to keep both open until the issue is resolved.

pmorch:
I agree, this issue is a subset of #1903. I'll move some comments to #1903 instead.

pmorch:
> I'm very sorry that you just had a machine trashed.

Thanks. I understand that nobody did this on purpose and that this is new software. It's fine.

PR #3684, opened 2016-10-24
Add note regarding cryptography dependency in contribution docs [revision requested]

ClifHouck:
Installation of the cryptography pip module requires some admin work, and is
required to run tox tests.

Partial #3683

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -90,6 +90,12 @@ The following tools are there to help you:
   code. Alternatively, you can use Python's standard library `pdb`,
   but you won't get TAB completion...
 
+.. important::
+   To run Certbot's automated Python tests via ``tox`` you must be able to
+   install ``cryptography`` via ``pip``. See `Cryptography's documentation`_
+   for more information on its requirements for installation.
+
+.. _Cryptography's documentation: https://cryptography.io/en/latest/installation/#installation
 
 .. _integration:
 


bmw:
Can I ask what admin work you had to do? Installing OS dependencies is covered under [running a local copy of the client](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client). Are you using a version of OpenSSL other than what is packaged with your operating system?

ClifHouck:
I didn't read that part of the documentation, my mistake. I'm used to being able to to run 'tox' on a freshly cloned repo and have unit tests run without needing additional boot-strapping.

I wonder if there's value in running ``./letsencrypt-auto-source/letsencrypt-auto --os-packages-only`` as an initial step of creating pyXX environments?

bmw:
I think running `{certbot,letsencrypt}-auto` in `tox` would surprise users and is more dangerous than just having tests fail. For example, see https://github.com/certbot/certbot/issues/2406#issue-132355287

I would rather either:

1. Change our documentation to make this more clear.
2. Modify `tox.ini` to catch this problem up front or provide a better error message if installation fails.

This problem is faced by any Python project that has OS dependencies (that aren't covered by `tox`). I'm not sure if there's a common practice here.

ClifHouck:
Oh man, I didn't think about it being able to hose a system like that.

Maybe I should just change this PR to talk about how running *-auto can install dependencies required to run certbot/tests, but that its potentially dangerous?

bmw:
So `certbot-auto` (as opposed to `letsencrypt-auto`) prompts users before installing any OS packages. Because of this behavior, I don't think it's dangerous anymore, however, I also think it's a strange practice to get a `sudo` prompt to run your package manager from inside `tox`.

Perhaps you can just add a section to the beginning of the "Testing" saying like:

"Before running tests, Certbot's OS dependencies need to be installed. You can do this by running ./certbot-auto --os-packages-only` as described in Running a local copy of the client" where "Running a local copy of the client" is a link.

While we're at it, we should update the "Running a local copy of the client" to use `./certbot-auto` rather than `letsencrypt-auto`.

jsha:
This PR is fairly old, and I think is already covered by a different part of the documentation. Closing for now, but if you'd like to propose further improvement, please go ahead!

PR #3687, opened 2016-10-24
Add IPv6 support to ACME standalone mode

### LINKED ISSUE #1466: Standalone mode does not bind to ipv6

jayofdoom:
```
Added support for using IPv6 sockets if possible. As part of this,
I also had to fix crypto_util.probe_sni() to use ('', 0) as the
default source address, instructing python to use the system default
source. The previous value ('0', 0) was interpreted as the IPv4 address
0.0.0.0.

Strategy for detecting IPv6 support: check socket.has_ipv6 to see if
python was compiled with IPv6 support. If it was, attempt to bind to the
port we are trying to standup a server on on IPv6 localhost. If it works,
IPv6 is enabled.
```

Closes #1466

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -4,6 +4,7 @@ import collections
 import functools
 import logging
 import os
+import socket
 import sys
 
 from six.moves import BaseHTTPServer  # pylint: disable=import-error
@@ -22,10 +23,41 @@ logger = logging.getLogger(__name__)
 # pylint: disable=too-few-public-methods,no-init
 
 
+def ipv6_enabled(port):
+    """Method to see if IPv6 is functional.
+
+    This method only verifies IPv6 is enabled in underlying OS; this method
+    makes nor attempts no claims that the IPv6 internet is reachable.
+
+    :param port: Port to attempt bogus IPv6 connection
+    :returns: bool indicating if IPv6 is enabled in underlying system
+    """
+    if not socket.has_ipv6:  # pragma: no cover
+        # Python compiled without IPv6 support
+        return False
+
+    try:
+        testsock = socket.socket(socket.AF_INET6,
+                                 socket.SOCK_STREAM,
+                                 socket.IPPROTO_IP)
+        testsock.bind(('::1', port))
+        testsock.close()
+    except socket.error:
+        return False
+
+    # We were able to bind to an IPv6 address.
+    return True
+
+
 class TLSServer(socketserver.TCPServer):
     """Generic TLS Server."""
 
     def __init__(self, *args, **kwargs):
+        if ipv6_enabled(args[0][1]):
+            self.address_family = socket.AF_INET6
+        else:
+            self.address_family = socket.AF_INET
+
         self.certs = kwargs.pop("certs", {})
         self.method = kwargs.pop(
             # pylint: disable=protected-access
@@ -74,6 +106,11 @@ class HTTP01Server(BaseHTTPServer.HTTPServer, ACMEServerMixin):
     """HTTP01 Server."""
 
     def __init__(self, server_address, resources):
+        if ipv6_enabled(server_address[1]):
+            self.address_family = socket.AF_INET6
+        else:
+            self.address_family = socket.AF_INET
+
         BaseHTTPServer.HTTPServer.__init__(
             self, server_address, HTTP01RequestHandler.partial_init(
                 simple_http_resources=resources))
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -103,7 +103,7 @@ class SSLSocket(object):  # pylint: disable=too-few-public-methods
 
 
 def probe_sni(name, host, port=443, timeout=300,
-              method=_DEFAULT_TLSSNI01_SSL_METHOD, source_address=('0', 0)):
+              method=_DEFAULT_TLSSNI01_SSL_METHOD, source_address=('', 0)):
     """Probe SNI server for SSL certificate.
 
     :param bytes name: Byte string to send as the server name in the
--- a/acme/acme/standalone_test.py
+++ b/acme/acme/standalone_test.py
@@ -1,6 +1,8 @@
 """Tests for acme.standalone."""
+import mock
 import os
 import shutil
+import socket
 import threading
 import tempfile
 import time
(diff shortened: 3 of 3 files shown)


acme/acme/standalone.py
```diff
@@ -22,10 +23,40 @@
 # pylint: disable=too-few-public-methods,no-init
 
 
+def _ipv6_enabled(port):
```
CrashenX: Given that listening on "::" will still work with subsystem disabled in Linux (at least for local connections over ipv4), this is probably not strictly necessary. However, listening on "::1" would bomb, and this ensures that ipv6 is enabled, so I think it's good.

jayofdoom:
Coverage check failing, but want to make sure reviewers agree with approach before I write test coverage for all this code.

jayofdoom:
Wrote tests to fix coverage issues; one test was nearly-impossible to write; something about socket.has_ipv6 makes it incredibly difficult to mock. @bmw said in IRC he'd try to figure it out, if he does will notify here and I'll update the PR. As-is, I'm just disabling coverage for that one line.

joohoi:
I can see a small problem with this; if your underlying OS support IPv6 (pretty much everything current out there atm), but you don't actually have a working IPv6 route to your machine (also way too common), this will break standalone plugin, as it does not have a IPv4 fallback. There should also be a way for user to define which one they prefer (know that works).

jayofdoom:
I don’t know why this isn’t showing up in GitHub on my view; but I think you misread the patch. This adds IPv6 support to the /server side/, not the client side. All we require is the ability to listen on :: rather than 0.0.0.0, whether or not IPv6 actually works on the server is irrelevant as long as the OS/python support is there.

joohoi:
I acknowledged that it's a server side (or actually client side, as in acme client context ;), but I realized the very thing you pointed out - still supporting v4, through IPv4 mapping.

bmw:
@jayofdoom, thanks for the PR and sorry it sat around so long.

While we were initial excited by the possibility of simply using an IPv6 socket when possible to solve #1466, I don't think that approach will work. When reviewing this PR, I came across the following stack exchange [post](http://security.stackexchange.com/questions/92081/is-receiving-ipv4-connections-on-af-inet6-sockets-insecure). The tl;dr version is FreeBSD/OpenBSD do not map IPv4 traffic to IPv6 sockets and I have verified this is at least the case on modern FreeBSD. Since we support both of these systems with Certbot, this approach won't work.

Ideally, we can handle both IPv4 and IPv6 traffic transparently for the user. The best way I am aware of to try and do this (which is recommended in the stack exchange post) is to use two sockets, one for IPv4 and one for IPv6. The IPv6 socket has to have `IPV6_V6ONLY` set to prevent errors when binding the second socket on systems that do map IPv4 traffic to IPv6 sockets.

If you're still interested in and available to work on this, we'd love to see more from you here. If you're not, let us know and we can have someone else take a crack at it.

jayofdoom:
I attempted this morning to fix the patch up, but found a few problems:

It appears that the entire interactions between certbot/acme and spinning up servers assumes one server being passed around. I'm not sure how to refactor it so that multiple servers (separate sockets/servers for IPv6 + IPv4) could work, and it seems super invasive.

So, I changed to option 2: adding the ability to explicitly request IPv6 support via the command line. Problem is, I had significant trouble getting certbot to run at all on my Fedora 24 system, so any testing was out of the question. (Issue #3838)

I have a little more time to contribute to certbot, either in troubleshooting my issue or cleaning up this patch, but making more progress will require additional help. If you can provide it I can take another stab at it over Thanksgiving holiday.

bmw:
Sorry I didn't respond to this sooner @jayofdoom.

> It appears that the entire interactions between certbot/acme and spinning up servers assumes one server being passed around. I'm not sure how to refactor it so that multiple servers (separate sockets/servers for IPv6 + IPv4) could work, and it seems super invasive.
>
> So, I changed to option 2: adding the ability to explicitly request IPv6 support via the command line. Problem is, I had significant trouble getting certbot to run at all on my Fedora 24 system, so any testing was out of the question. (Issue #3818)

I think automatically handling IPv4/IPv6 for the user is definitely the best option. Not only is it simpler for the user, but with the current architecture of Certbot, it allows you to do things you couldn't do things that you otherwise could not. While perhaps somewhat unlikely, imagine a user has ipv4.example.org and ipv6.example.org, which each of the domains only being accessible over the specified protocol. If there is a command line flag to specify either IPv4 or IPv6, a user cannot obtain a single certificate for both domains. (In actuality, they might be able to, but it's not sustainable for things like automatic renewal.)

As for how to implement this, I think the best approach is to create a class in `acme/standalone.py` that spins up both an IPv4 and an IPv6 server. Failures from only one server would not raise exceptions to the user of the class. Alternatively, you could make `ServerManager` in `certbot/plugins/standalone.py` create two server instances for each challenge, however, I prefer putting this functionality in ACME so other clients can benefit.

This is not a trivial refactor. If it appears we're not able to take this approach soon, we may want to consider creating a command line flag. This will allow users to start using the standalone plugin to get a cert on an IPv6 only server. The downside of this approach is complicates our command line interface, adding options we know we won't need in the future which we'll have to slowly deprecate to avoid breaking people's Certbot integrations.

I certainly have a preference for writing the automatic IPv4/IPv6 code. If you don't have time to do this, someone else could build on what you have so far. I may be able to do this myself, but probably not until January.

bmw:
Is anyone working on this? People were interested, but if people are uninterested/unavailable now, one of the core developers can take over.

nick-2017:
please can this be added, I need a certificate for my ipv6 only hosts

jayofdoom:
@bmw I don't intend on working on this patch further, and would love to see it picked up by someone else.

bmw:
OK. Thanks for the update and the work you've done so far. I'll probably pick up this issue myself for our 0.12.0 release unless someone else starts working on it. I'm closing this PR, but I've put the code here in our [ipv6-standalone branch](https://github.com/certbot/certbot/tree/ipv6-standalone) in case it proves useful for resolving #1466.

Issue #3641, opened 2016-10-16
[nginx] --hsts and --uir flags not working?


emilstahl:
Am I the only one who experience that the `--hsts` and `--uir` flags isn't adding any Strict-Transport-Security nor CSP headers?

I'm running the following and isn't seeing the headers added to any of the virtualhosts.

```
./certbot-auto --nginx --rsa-key-size 4096 --hsts --redirect --uir --staple-ocsp --must-staple -m hostmaster@domain.dk -d domain.dk -d somesub.domain.dk -d other.domain.dk
```

 The other relevant certbot stuff are however added.

```
# ./certbot-auto --version
letsencrypt 0.9.3
# cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)
```

bmw:
The current Nginx plugin doesn't support enabling HSTS or UIR. This isn't at all clear to the user though. We should either:

1. Check that requested enhancement isn't supported by the selected plugin early and error out with a helpful error message.
2. Print a warning saying that the requested enhancement isn't supported by the selected plugin and continue execution.

I also created #3648 and #3649.

emilstahl:
👍 

I think 2 is great, as it is non breaking issues.

bmw:
Fixed in #3641

haarts:
You mean 'Fixed in https://github.com/certbot/certbot/pull/3688' ;)

PR #3688, opened 2016-10-24
Warn user if client is trying to use unsupported config option

supermari0:
…ugin

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -196,6 +196,11 @@ class Client(object):
         else:
             self.auth_handler = None
 
+        # Warn if the client is using unsupported config options with an
+        # installer
+        if self.installer is not None:
+            self._verify_all_config_options_supported(config)
+
     def obtain_certificate_from_csr(self, domains, csr,
         typ=OpenSSL.crypto.FILETYPE_ASN1, authzr=None):
         """Obtain certificate.
@@ -411,9 +416,10 @@ class Client(object):
             raise errors.Error("No config available")
 
         supported = self.installer.supported_enhancements()
+
         redirect = config.redirect if "redirect" in supported else False
         hsts = config.hsts if "ensure-http-header" in supported else False
-        uir = config.uir if "ensure-http-header"  in supported else False
+        uir = config.uir if "ensure-http-header" in supported else False
         staple = config.staple if "staple-ocsp" in supported else False
 
         if redirect is None:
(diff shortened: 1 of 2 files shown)


supermari0:
This should address #3641

certbot/client.py
```diff
@@ -410,11 +410,13 @@ def enhance_config(self, domains, config, chain_path):
...
 
-        supported = self.installer.supported_enhancements()
-        redirect = config.redirect if "redirect" in supported else False
-        hsts = config.hsts if "ensure-http-header" in supported else False
-        uir = config.uir if "ensure-http-header"  in supported else False
-        staple = config.staple if "staple-ocsp" in supported else False
+
+        self._verify_all_config_options_supported(config)
+
+        redirect = getattr(config, "redirect", None)
```
bmw: This should only be necessary during unit tests. With this in mind, I'd rather us fix the problem in the tests than using this in production code.

certbot/client.py
```diff
@@ -410,11 +410,13 @@ def enhance_config(self, domains, config, chain_path):
             logger.warning("No config is specified.")
             raise errors.Error("No config available")
 
-        supported = self.installer.supported_enhancements()
-        redirect = config.redirect if "redirect" in supported else False
-        hsts = config.hsts if "ensure-http-header" in supported else False
-        uir = config.uir if "ensure-http-header"  in supported else False
-        staple = config.staple if "staple-ocsp" in supported else False
+
+        self._verify_all_config_options_supported(config)
```
bmw: This method raises an error crashing Certbot if the requested enhancement is not supported. I do not think we should do this here. This code will not run until near the very end of Certbot's execution so unfortunately this doesn't make Certbot fail fast.

Instead, I think we should:

1. Change `raise error.Error(msg)` in `_verify_all_config_options_supported` to `logger.warning(msg)`.
2. Call `_very_all_config_options_supported` in this class's `__init__` method. This will make Certbot fail fast as this happens before Certbot has begun any communication with Let's Encrypt

supermari0: Done, and added a new unit test to check that the logging happens in the constructor

certbot/client.py
supermari0: Done. Reverted this to what it was before and am doing everything in the constructor now

supermari0:
Removed [revision requested] tag and updated title to reflect the fact that this warns instead of fails now

bmw:
LGTM. Thanks @supermari0!

Issue #3629, opened 2016-10-13
Add a cli option for "all domains my installer sees"


pde:
There should be a way to say, "the domains I want to select are all of the domains my installer plugin believes to be available".  This could look like `certbot --domains all`, or `certbot --all-domains`.

pde:
Of course the local network ones would need to be filtered out, which we could get by calling `util.enforce_le_validity()`.

bmw:
I'm assuming you created this ticket in response to potentially removing `dialog`. To me, it seems like a bit of a bad idea to run Certbot blindly without knowing what domains you're requesting. You could have missed domains, included domains you didn't want (see #3545), etc. What about instead we did something like:
```
Which names would you like to activate HTTPS for?
-------------------------------------------------------------------------------
1: example.org
2: www.example.org
-------------------------------------------------------------------------------
Select the appropriate numbers separated by commas and/or spaces or
simply press enter to select all above domains (Enter 'c' to cancel):
```
This makes it just as easy to select all domains when using the text interface.

TheNavigat:
@bmw @pde This issue should probably be closed now since #3693 is merged.

bmw:
Fixed in #3693.

PR #3693, opened 2016-10-26
Allow user to select all domains by typing empty string at checklist

ohemorange:
Addresses #3629, as a follow-up to #3665.

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -179,9 +179,12 @@ class FileDisplay(object):
             self._print_menu(message, tags)
 
             code, ans = self.input("Select the appropriate numbers separated "
-                                   "by commas and/or spaces")
+                                   "by commas and/or spaces, or press enter to "
+                                   "select all domains shown")
 
             if code == OK:
+                if len(ans) == 0:
+                    ans = " ".join([str(x) for x in range(1, len(tags)+1)])
                 indices = separate_list_input(ans)
                 selected_tags = self._scrub_checklist_input(indices, tags)
                 if selected_tags:
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -179,12 +179,12 @@ class FileDisplay(object):
             self._print_menu(message, tags)
 
             code, ans = self.input("Select the appropriate numbers separated "
-                                   "by commas and/or spaces, or press enter to "
-                                   "select all domains shown")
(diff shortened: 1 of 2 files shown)


certbot/display/util.py
```diff
@@ -179,9 +179,12 @@ def checklist(self, message, tags, default_status=True, **unused_kwargs):
             self._print_menu(message, tags)
 
             code, ans = self.input("Select the appropriate numbers separated "
-                                   "by commas and/or spaces")
+                                   "by commas and/or spaces, or press enter to "
+                                   "select all domains shown")
```
bmw: We probably shouldn't say "domains" here as this is supposed to be a general method for displaying a checklist. Perhaps we should change it to "options"?

Also, nit:

Despite this being largely the wording I proposed, I'm not sure I like "press enter to select all". You also have to press enter to select a specific options after entering the desired numbers. While this is probably unlikely to confuse people, perhaps we should say "simply/just press enter" or something more verbose like "press enter without selecting any numbers to select all".

certbot/display/util.py
```diff
@@ -179,9 +179,12 @@ def checklist(self, message, tags, default_status=True, **unused_kwargs):
...
 
             if code == OK:
+                if len(ans) == 0:
```
bmw: nit: It might be nice to change this to:
```
if len(ans.strip()) == 0:
```
so the method still works if the user accidentally provides whitespace.

certbot/display/util.py
```diff
@@ -179,9 +179,12 @@ def checklist(self, message, tags, default_status=True, **unused_kwargs):
...
+                    ans = " ".join([str(x) for x in range(1, len(tags)+1)])
```
bmw: Bonus points: Remove square brackets for a generator expression.

ohemorange: weird, I swear this wasn't working when I tried it before.

ohemorange: (works now)

certbot/display/util.py
bmw: I think "leave input blank..." sounds good.

bmw:
The Travis instance had a problem trying to install sudo ([log](https://paste.ubuntu.com/23385355/)). Kicking Travis to try running tests again.

ohemorange:
@bmw :joy:

PR #3695, opened 2016-10-26
Fix travis tests?

bmw:


ohemorange:
Why is this just failing now? How are our tests passing with out?

ohemorange:
*without it?

bmw:
> Why is this just failing now?

Travis probably changed something upstream. If I remember correctly, we're using an undocumented feature of Travis where different tests run on different environments. The environments that need `sudo` are slower, but we need it in some cases and wanted to avoid using the slower environments for all builds.

> How are our tests passing without it?

Because we specify `sudo` for each job that needs it ([source](https://github.com/certbot/certbot/blob/master/.travis.yml#L48)). It's my guess we never needed `sudo` here. For example, `boulder` uses `sudo` in all their builds, yet don't depend on the `sudo` package like we were doing before this PR ([source](https://github.com/letsencrypt/boulder/blob/master/.travis.yml#L13)).

Issue #3539, opened 2016-09-26
Create leaf.pem instead of cert.pem for new lineages


jsha:
We see a fairly large number of sites misconfigured to serve only the end-entity (leaf) certificate, not the entire chain. I believe part of this is that we create a file `cert.pem` in addition to `fullchain.pem`. For someone looking to manually install a certificate, `cert.pem` looks like the natural choice. We can help people configure their sites correctly by renaming `cert.pem` to something else, like leaf.pem. This should be accompanied by a README in the certificates directory explaining what each file is.

We don't want to disrupt existing lineages where people are depending on certain naming conventions, but can can have Certbot take this new behavior for new lineages.

I realize this is a fairly disruptive change, but since part of our goal is to remove the human error and pain points in deploying certificates, it seems worthwhile.

bmw:
Related: #2880

How do you suggest such a change is implemented? If a lineage already exists using `cert.pem`, continue using that convention (and maybe add an extra symlink for `leaf.pem`), but if we're creating a new lineage, only create `leaf.pem`?

jsha:
Yep, exactly.

schoen:
This might cause backwards compatibility problems with scripts and tools that integrate with Certbot -- for example, if there are scripts to integrate with control panels or something that automatically pull in `cert.pem` under that name. I'm concerned that we'd trigger a lot of integration bugs as a result, even though I think it would become clearer for people who use Certbot directly on the command line.

ohemorange:
Maybe we could symlink fullchain.pem to something like the-thing-you-want.pem (but better named).

schoen:
Well, at least creating a README could help because maybe then some people would look at it!

schoen:
I'd also like to get @pde's view of the backwards-compatibility issue. I agree with @jsha's underlying concern, so if it weren't for the backwards-compatibility problem (that is, third-party scripts and tools that may wrap Certbot and assume the presence of `cert.pem`), I'd favor making this change.

pde:
`leaf.pem` might have been a better choice, but I don't think we should change it now. We can do #1819 instead...

pde:
@ohemorange unfortunately the-thing-you-want depends on what server you're using. Most often it's `fullchain.pem`, sometimes it's `cert.pem` and `chain.pem` separately, sometimes it's the concatenation of `fullchain.pem` and `privkey.pem`.  See #1250 and #1201.

cowlicks:
[...]

Issue #1819, opened 2015-12-07
Add a README file to each live directory explaining its contents


schoen:
Individually tracking the suggestion at

https://github.com/letsencrypt/letsencrypt/issues/1201#issuecomment-155652280

to have a README file to explain to users what's going on.

bmw:
Related to #608.

schoen:
Does anyone want to propose what this paragraph might say?

pde:
Perhaps steal some of the text that you use for #3232?

schoen:
(about the explanation of lineages)

pde:
I'd do this: write just a paragraph about what files are in the lineage, and how to work with them (eg, always refer `/etc/letsencrypt/live` rather than the `archive` files etc).

Then link to your new docs that provide more narrative about flags for working with lineages.

bmw:
Related #3615

PR #3696, opened 2016-10-26
Add README file to each live directory explaining its contents.

ohemorange:
Addresses #1819. Related #3539.


certbot/storage.py
```diff
@@ -810,6 +810,19 @@ def new_lineage(cls, lineagename, cert, privkey, chain, cli_config):
...
+        with open(readme_path, "w") as f:
+            logger.debug("Writing README to %s.", readme_path)
+            f.write("The files in this directory are symlinks to the latest "
+                    "versions of your keys and certificates.\n\n"
+                    "`privkey.pem` is the private key for your certificate.\n"
+                    "`fullchain.pem` contains all the information your certificate needs, and "
+                    "is likely the certificate file you will need.\n"
+                    "`chain.pem` is used for OCSP stapling in Nginx >=1.3.7.\n\n"
+                    "For more information, see the Certbot User Guide "
+                    "at https://certbot.eff.org/docs/using.html.\n")
```
schoen: It might be helpful to break this text out into a variable that's set elsewhere so that it will be slightly easier to find and modify.

I'd propose something like this:

```
The files in this directory were created by the Certbot software, which may also be able to renew this certificate automatically. These files are symbolic links to the latest versions of your key and certificate data and are automatically updated when the certificate is renewed. These files, rather than the link targets, should be referred to in server configurations.

privkey.pem  is the private key for your certificate.

fullchain.pem  contains the certificate together with the associated chain certificate, which is needed by some software that connects to your site in order to confirm that the certificate is valid. This is likely the certificate file that you should use with server software.

chain.pem  contains only the chain certificate. This is used in specialized circumstances, such as for HTTPS with Apache versions before 2.4, and OCSP stapling in Nginx versions starting with 1.3.7.

cert.pem  contains only the certificate, which is not sufficient by itself and should not be used in most software. It can be used in specialized circumstances together with chain.pem, such as in Apache versions before 2.4. Most users need to use fullchain.pem instead of cert.pem.

If you have OpenSSL installed on this system, you can see detailed information about the current certificate's content by running

    openssl x509 -in cert.pem -text -noout

For more information, please see the Certbot User Guide at https://certbot.eff.org/docs/using.html; assistance may also be available from the Let's Encrypt Community Forum at https://community.letsencrypt.org/.
```

ohemorange: I'm good with breaking it out, but my goal here was to keep it as short as possible and point people to the online docs, because I don't want to have to be keeping this up to date in sync with the online docs.

schoen: Also, I'm fine with this PR apart from the question of exactly what the README should say.

ohemorange:
Updates from IRL conversation:

- Add "we recommend not moving these files, for more info see the docs" instead of symlink implementation detail.
- Instead of "Nginx >= whatever", say "modern server software".

ohemorange:
Some Travis problems but maybe it'll work anyway; closing and reopening.

ohemorange:
README now reads:

```
This directory contains your keys and certificates.

`privkey.pem` is the private key for your certificate.
`fullchain.pem` is the certificate file used in modern server software.
`chain.pem` is used for OCSP stapling in Nginx >=1.3.7.

We recommend not moving these files. For more information, see the Certbot User Guide at https://certbot.eff.org/docs/using.html.
```

certbot/storage.py
```diff
@@ -833,6 +833,18 @@ def new_lineage(cls, lineagename, cert, privkey, chain, cli_config):
...
+        readme_path = os.path.join(live_dir, "README")
+        with open(readme_path, "w") as f:
+            logger.debug("Writing README to %s.", readme_path)
+            f.write("This directory contains your keys and certificates.\n\n"
+                    "`privkey.pem` is the private key for your certificate.\n"
+                    "`fullchain.pem` is the certificate file used in modern server software.\n"
+                    "`chain.pem` is used for OCSP stapling in Nginx >=1.3.7.\n\n"
+                    "We recommend not moving these files. "
+                    "For more information, see the Certbot User Guide "
+                    "at https://certbot.eff.org/docs/using.html.\n")
```
ohemorange: I like adding the fragment.

I'd rather not even mention the wrong thing to use? That is, instead of saying, "use this not that", I'd rather say "use this".

I'd definitely prefer to not mention renew hook here.

pde: Concluded that we do need *some* docs for `cert.pem`, not least because some modern server software probably still needs it (though not Nginx and Apache).

ohemorange:
Current copy, formatted:

```
This directory contains your keys and certificates.

`privkey.pem`  : the private key for your certificate.
`fullchain.pem`: the certificate file used in most server software.
`chain.pem`    : used for OCSP stapling in Nginx >=1.3.7.
`cert.pem`     : will break many server configurations, and should not be used
                 without reading further documentation (see link below).

We recommend not moving these files. For more information, see the Certbot
User Guide at https://certbot.eff.org/docs/using.html#where-are-my-certificates.
```

certbot/storage.py
schoen: This seems all to have been addressed by further discussion now.

PR #2671, opened 2016-03-16
Expanding tests for le-auto, adding CentOS test suite

TheNavigat:
Addresses issue #2443

Requires PR #2926

PR #3701, opened 2016-10-27
Point people towards OS packages [revision requested]

surya-v:
Edited deb_common.sh and rpm_common.sh to check if certbot is already packaged with the OS

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -152,6 +152,8 @@ DeterminePythonVersion() {
   fi
 }
 
+#!/usr/bin/env bash
+#!/usr/bin/env bash
 BootstrapDebCommon() {
   # Current version tested with:
   #
@@ -172,11 +174,18 @@ BootstrapDebCommon() {
   # - Debian 6.0.10 "squeeze" (x64)
 
   $SUDO apt-get update || echo apt-get update hit problems but continuing anyway...
+  
+  # checking if certbot/letsencrypt is already packaged with OS
+  if [ "$QUIET" != 1 ]; then
+      if dpkg --get-selections | grep -v letsencrypt.sh | grep -c letsencrypt -gt 0; then
+        echo "Certbot is already packaged with your OS."
+        dpkg --get-selections | grep -v letsencrypt.sh | grep letsencrypt
+      fi
+  fi
 
   # virtualenv binary can be found in different packages depending on
   # distro version (#346)
-
-  virtualenv=
+    virtualenv=
   if apt-cache show virtualenv > /dev/null 2>&1 && ! apt-cache --quiet=0 show virtualenv 2>&1 | grep -q 'No packages found'; then
     virtualenv="virtualenv"
   fi
@@ -278,6 +287,14 @@ BootstrapRpmCommon() {
   else
     echo "Neither yum nor dnf found. Aborting bootstrap!"
     exit 1
+  fi
+
+  # checking if certbot is already packaged with the os
+  if [ "$QUIET" != 1 ]; then
+      if rpm -qa | grep -c certbot -gt 0; then
+         echo "Certbot is already packaged with you OS."
+         rpm -qa | grep certbot
+      fi
   fi
 
   if [ "$ASSUME_YES" = 1 ]; then
--- a/letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
+++ b/letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+#!/usr/bin/env bash
 BootstrapDebCommon() {
   # Current version tested with:
   #
@@ -18,11 +20,18 @@ BootstrapDebCommon() {
   # - Debian 6.0.10 "squeeze" (x64)
 
   $SUDO apt-get update || echo apt-get update hit problems but continuing anyway...
+  
+  # checking if certbot/letsencrypt is already packaged with OS
+  if [ "$QUIET" != 1 ]; then
+      if dpkg --get-selections | grep -v letsencrypt.sh | grep -c letsencrypt -gt 0; then
+        echo "Certbot is already packaged with your OS."
+        dpkg --get-selections | grep -v letsencrypt.sh | grep letsencrypt
+      fi
+  fi
 
   # virtualenv binary can be found in different packages depending on
   # distro version (#346)
-
-  virtualenv=
+    virtualenv=
   if apt-cache show virtualenv > /dev/null 2>&1 && ! apt-cache --quiet=0 show virtualenv 2>&1 | grep -q 'No packages found'; then
     virtualenv="virtualenv"
   fi
--- a/letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
+++ b/letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
@@ -17,6 +17,14 @@ BootstrapRpmCommon() {
     exit 1
   fi
 
+  # checking if certbot is already packaged with the os
+  if [ "$QUIET" != 1 ]; then
+      if rpm -qa | grep -c certbot -gt 0; then
+         echo "Certbot is already packaged with you OS."
+         rpm -qa | grep certbot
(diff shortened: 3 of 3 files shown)


letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+#!/usr/bin/env bash
```
TheNavigat: Why are these duplicated, and why are they needed in the first place?

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -18,11 +20,18 @@ BootstrapDebCommon() {
...
+        echo "Certbot is already packaged with your OS."
+        dpkg --get-selections | grep -v letsencrypt.sh | grep letsencrypt
+      fi
+  fi
 
   # virtualenv binary can be found in different packages depending on
   # distro version (#346)
-
-  virtualenv=
+    virtualenv=
```
TheNavigat: Is the extra indentation a mistake?

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -18,11 +20,18 @@ BootstrapDebCommon() {
   # - Debian 6.0.10 "squeeze" (x64)
 
   $SUDO apt-get update || echo apt-get update hit problems but continuing anyway...
+  
+  # checking if certbot/letsencrypt is already packaged with OS
+  if [ "$QUIET" != 1 ]; then
+      if dpkg --get-selections | grep -v letsencrypt.sh | grep -c letsencrypt -gt 0; then
```
TheNavigat: I'm not a bash master, but that if condition doesn't work properly. Apparently the -gt 0 bit doesn't work as intended.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
surya-v: They got added by mistake. Removed them now.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
surya-v: Yes it was unintentional. I have reverted this change. I would check the diff properly before creating a pull request from now on.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
surya-v: i am not sure in which cases this fails. But any how i have replaced "-gt" with ">".

TheNavigat:
First of all, you need to implement the fixes suggested to proceed with this (unless you think they're wrong, you're very welcome to do that :) ).

Second, regarding your comment, https://github.com/certbot/certbot/pull/3701/files#r86008033

You should add your tests to https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/tests/auto_test.py. In this case, I _think_ we should basically install the package and check the output in both cases, if quiet is enabled, and if it's not.

I'm not 100% sure this is the best way to implement the test (as in by installing the package itself), so feel free to suggest a different method, or to ask anyone else. Also, be aware that for the time being, LEA tests are ran only on Ubuntu (there's a PR that adds CentOS but it was never merged, https://github.com/certbot/certbot/pull/2671), so the RPM test will never be ensured automatically. However, try to make your test as OS-generic as possible so that it will automatically be triggered whenever CentOS LEA testing is added.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
pde: Where you currently have the code `grep -c letsencrypt -gt 0`, any of the following would do what you actually want:

```
grep -q letsencrypt                # grep returns true or false based on whether it has any matches`
test `grep -c letsencrypt` -gt 0   # run grep as a subcommand that counts lines, compare the answer to 0
[ `grep -c letsencrypt` -gt 0 ]    # in the unix shell putting something in [ ] is the same as using the "test" command
```
However only the first of these three is really idiomatic bash; I've never seen anyone pipe the output of something into `test` before.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -172,11 +174,18 @@ BootstrapDebCommon() {
   # - Debian 6.0.10 "squeeze" (x64)
 
   $SUDO apt-get update || echo apt-get update hit problems but continuing anyway...
+  
+  # checking if certbot/letsencrypt is already packaged with OS
+  if [ "$QUIET" != 1 ]; then
+      if dpkg --get-selections | grep -v letsencrypt.sh | grep -c letsencrypt -gt 0; then
```
pde: I don't think we want to check for `letsencrypt` packages; although they're present on Xenial, they're increasingly old and crufty and we want to deprecate them.  Let's just check for `certbot` packages.

letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
pde: My previous comment was about how to do what you were trying to do in a shell script. But what we should probably actually do here is this:

```
if apt-cache show certbot > /dev/null 2>&1 ; then
   echo "NOTE: Cerbot is packaged for your OS; you may want use those packages rather than cerbot-auto"
fi

Comment (pde) [inline letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh]: (That's more consistent with the method we already use to [test package availability](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh#L30) on debian)

File: letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh (hunk 1 above)
Comment (jsha) [inline letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh]: It looks like you never pushed the change removing these `#!/usr/bin/env bash` lines. Could you push it?

File: letsencrypt-auto-source/letsencrypt-auto [hunk 5]
```diff
@@ -152,6 +152,8 @@ DeterminePythonVersion() {
   fi
 }
 
+#!/usr/bin/env bash
+#!/usr/bin/env bash
```
jsha: These lines are unnecessary and should be removed.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -280,6 +289,14 @@ BootstrapRpmCommon() {
     exit 1
   fi
 
+  # checking if certbot is already packaged with the os
+  if [ "$QUIET" != 1 ]; then
+      if rpm -qa | grep -c certbot -gt 0; then
+         echo "Certbot is already packaged with you OS."
```
jsha: s/you/your/.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -280,6 +289,14 @@ BootstrapRpmCommon() {
...
+         rpm -qa | grep certbot
```
jsha: Why do you run rpm again here? I think this is a mistake.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -280,6 +289,14 @@ BootstrapRpmCommon() {
     exit 1
   fi
 
+  # checking if certbot is already packaged with the os
+  if [ "$QUIET" != 1 ]; then
+      if rpm -qa | grep -c certbot -gt 0; then
```
jsha: You don't need the `-gt 0` here. You also don't need `-c`. Just `if rpm -qa | grep certbot; then`.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -280,6 +289,14 @@ BootstrapRpmCommon() {
     exit 1
   fi
 
+  # checking if certbot is already packaged with the os
```
jsha: Capitalize OS.

bmw:
This hasn't seen any progress since it was initially reviewed in November. If you'd like to continue working on it, address review feedback and open a new PR.

PR #3704, opened 2016-10-28
Allowing modification check to run using "tox"

### LINKED ISSUE #3339: Run simple certbot-auto tests with `tox`
#3337 and #3338 should ideally run when the user type `tox`. This allows them to catch the problem locally before they make a PR.

TheNavigat:
Closes #3339

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 
 [tox]
 skipsdist = true
-envlist = py{26,33,34,35},cover,lint
+envlist = modification,py{26,33,34,35},cover,lint
 
 # nosetest -v => more verbose output, allows to detect busy waiting
 # loops, especially on Travis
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -93,11 +93,16 @@ commands =
...
+# allow users to run the modification check by running `tox`
+[testenv:modification]
+commands =
+    {toxinidir}/tests/modification-check.sh
+
 [testenv:le_auto]
 # At the moment, this tests under Python 2.7 only, as only that version is
 # readily available on the Trusty Docker image.
 commands =
-    {toxinidir}/tests/modification-check.sh
```
bmw: Was this deleted by mistake or did we have a misunderstanding? I don't think we should delete this line.

TheNavigat: My bad, it's a mistake indeed

PR #3721, opened 2016-10-31
Remove some domain name checks.

### LINKED ISSUE #3700: Don't crash when U-label IDN provided on command line
Steps to reproduce:

```
. venv/bin/activate
. tests/integration/_common.sh
certbot_test -d ウェブ.crud.net
```

Expected result: friendly error.

Actual result: Print newline, exit 1. No logs are written.

jsha:
These checks are done on the server side, so it's redundant to do them here, and
it causes some issues. In particular the Unicode check causes a crash, which
gets swallowed so that it isn't printed or written to logs.

Fixes #3700.


jsha:
Closing in favor of a smaller change, #3727.

Issue #3394, opened 2016-08-11
Hooks do not get stored in renewal config file


maddes-b:
OS: Debian 8 "Jessie" with all current updates (as of 2016-08-10)
Package: certbot 0.8.1-2~bpo8+ from jessie-backports

When getting a new cert via certonly then the hook definitions are not executed and stored in the renewal config file.
Used certonly with --config for domain specific stuff and the hook definitions did not get executed and are missing in the renewal config file.
Also checked with a fresh setup and defining the hook on the command line with "--post-hook 'systemctl reload apache2'", but no difference.
Expected behaviour after reading the docs was that the post hook gets executed on certonly and also stored in the renewal config file.
Getting the certifcate and renewal itself works fine.

My setup:
- File /etc/letsencrypt/cli.ini contains just registration config
- File /home/root/setup_files/files_letsencrypt/letsencrypt/letsencrypt.www.example.ltd.conf contains the domain config for the initial certonly call
- Executed certbot certonly --noninteractive --test-cert --config /home/root/setup_files/files_letsencrypt/letsencrypt/letsencrypt.www.example.ltd.conf
- Renewal file /etc/letsencrypt/renewal/www.example.ltd.conf was created

Following is the content of those file. The content was changed to only have example.ltd instead of the original multiple domains and webroots.

1) /etc/letsencrypt/cli.ini

```
email = someone@example.ltd
agree-tos
update-registration
```

2) /home/root/setup_files/files_letsencrypt/letsencrypt/letsencrypt.www.example.ltd.conf

```
authenticator = webroot

post-hook = systemctl reload apache2

### results will be in /etc/letsencrypt/live/<1st domain>
### here: /etc/letsencrypt/live/www.example.ltd/[cert|fullchain].pem and privkey.pem

### use domain to define the first domain (only works 100% on initial request)
domain = www.example.ltd

### in config file have to use webroot-map for multiple webroot-paths
### webroot-map must be a single line (as of 0.8.1)
webroot-map = {"www.example.ltd,example.ltd":"/home/maddes/www/example.ltd/httpdocs/" }

### ATTENTION!!!
### csr will make the cert not renewable via renew command, but can still be renewed via certonly
#csr = /etc/ssl/private/services-4096.www.example.ltd.sha512.csr
### paths are ignored when *not* called with csr
#cert-path = /etc/ssl/certs/services-4096.www.example.ltd.sha256.letsencrypt.crt
[...]

PR #3724, opened 2016-11-01
Add renew_hook to options stored in the renewal config, #3394

gilou:
renew_hook are called for each lineage, so it makes sense to store it and run it for each. Discussion on the issue probably still applies for pre/post hooks though. Discussion was in  #3394


pde:
We talked a bit about the idea of going further than this, and having `certbot renew --renew-hook x` run *in addition* to any renew hooks that were saved during cert creation. But although there are some logical reasons to want that behaviour, for now I think we concluded that the implementation and documentation complexity it would introduce might outweight the benefits.

So merging this as-is.

PR #3675, opened 2016-10-21
I restructured Installation and Using a bit

pconrad-fb:
I restructured install.rst and using.rst (and changed a link in README.rst) to reflect Option C as we discussed at https://github.com/certbot/website/issues/158#issuecomment-248759478 . I have not yet made a pass through the Packages section of the Alternate Installation Methods, but wanted to create this pull request to make sure the restructuring is sane.

NOTE:
I have a new work laptop and therefore I don't have Sphinx to do a test build. Please build before accepting my changes! I'll try to remedy the situation in my copious free time... :/

PR #3725, opened 2016-11-01
Handle merge conflicts in #3675

bmw:
Fixes merge conflict preventing #3675 from being merged, caused by #3651.

@ohemorange, @pde, or @schoen, I've reviewed #3675 and approved changes, but I also need a review from one of you to merge this PR (since we've turned on GitHub enforcing PRs are reviewed by another org member before they can be merged). I'd like this sooner rather than later because I want to make some follow up commits on the website repo.


ohemorange:
Any chance this commit could go on the end of #3675 so that when we squash, master is always in a valid state?

bmw:
I don't understand the concern. Why wouldn't `master` be in a valid state after merging this PR?

ohemorange:
Oh, I meant under the assumption that the merge order goes the other way. Isn't this PR replicating most of the work from the other one? Why is it over here separately?

bmw:
This PR is built on #3675. All commits in that PR are also here. I created this because I can't push to that PR and I didn't want to make the author deal with the problem.

GitHub **should** be smart enough to recognize this is the case and will automatically close #3765 when this PR is merged. If not, we can do that manually.

ohemorange:
Ah, didn't realize you can't push to that PR (I find this strange?). But in that case, that sounds good!

PR #3727, opened 2016-11-01
Fix non-ASCII domain check.

jsha:
Previously, the code would convert to utf-8, check for non-ASCII, and then try
to use .format() to interpolate the result into an error message. This would
generate a second error that would cause the whole message to get dropped, and
the program to silently exit. The problem can be succinctly observed like so:

```
$ python
>>> "{0}".format("ウェブ.crud.net")
'\xe3\x82\xa6\xe3\x82\xa7\xe3\x83\x96.crud.net'
>>> "{0}".format(u"ウェブ.crud.net")
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```

Note for the curious: This problem only seems to happen with .format():

```
>>> "%s" % ("ウェブ.crud.net")
'\xe3\x82\xa6\xe3\x82\xa7\xe3\x83\x96.crud.net'
>>> "%s" % (u"ウェブ.crud.net")
u'\u30a6\u30a7\u30d6.crud.net'
```

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -447,7 +447,10 @@ def enforce_domain_sanity(domain):
 
     # Unicode
     try:
-        domain.decode('utf-8').encode('ascii')
+        if isinstance(domain, six.binary_type):
+            domain.encode('ascii')
+        else:
+            domain.decode('utf-8').encode('ascii')
     except UnicodeError:
         raise errors.ConfigurationError("Non-ASCII domain names not supported. "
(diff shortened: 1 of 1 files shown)


bmw:
I don't know what's going on with Travis logs, but assuming the tests actually ran, I suspect the problem is string types don't have a `decode` method in Python 3.

PR #3728, opened 2016-11-02
Allow notification interface to not wrap text

ohemorange:
This leaves existing prints in place but allows new code to not wrap text.


ohemorange:
This is a roughly 10-line PR can we get it in?

PR #3737, opened 2016-11-03
Do we need trusty in Travis?

bmw:
~~We probably shouldn't merge this yet. While it works, I haven't yet convinced myself we should be happy with all the consequences.~~

Fixes Travis failures in `master.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ cache:
     directories:
         - $HOME/.cache/pip
 
-# This makes sure we get a host with docker-compose present.
-dist: trusty
-
 before_install:
   - 'dpkg -s libaugeas0'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,30 +18,34 @@ matrix:
   include:
     - python: "2.7"
       env: TOXENV=cover BOULDER_INTEGRATION=1
-      sudo: true
+      sudo: required
       after_failure:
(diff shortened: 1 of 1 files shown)


bmw:
Turns out, no, we don't need Trusty in Travis solving current build failures.

The effect of this PR is we change our [build environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) from Ubuntu 14.04 Trusty containers to Ubuntu 12.04 Precise containers for tests that do not require `sudo`. Ubuntu Trusty is still considered in its beta phase for Travis and they have significantly fewer packages whitelisted which was causing [the errors](https://travis-ci.org/certbot/certbot/jobs/173107329#L147) we were seeing. [Travis' advice](https://docs.travis-ci.com/user/trusty-ci-environment/#Container-based-with-sudo%3A-false) for dealing with dependencies that are not whitelisted is to either require `sudo` for the build (which significantly increases the build's startup time) or use Precise. Alternatively, we can ask for these packages to be whitelisted, however, I see no real benefit in choosing this route over what I'm proposing here.

This change has no effect on the build environments for tests requiring `sudo` as can be seen in the most recent Travis logs or this [blog post](https://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/). If you're using `docker`, Travis runs the test with Ubuntu Trusty.

Why did we suddenly started seeing failures now? Absolutely no idea.

Issue #3252, opened 2016-07-06
Use appropriate caution when handling configurations that have complex rewrite logic


pde:
If the user has a config file with complex redirects (eg conditional redirects), it's in general impossible for us to add HTTP -> HTTPS redirects safely.  It's even hard to create new vhosts safely, which is necessary in Apache (cf #3210 ).  Where possible, we should try to detect the sorts of conditional redirecting logic that create this danger ([here's an nginx example](http://redant.com.au/ruby-on-rails-devops/manage-ssl-redirection-in-nginx-using-maps-and-save-the-universe/); [here's an Apache example](http://serverfault.com/a/609348); but of course these will come in many flavours).

If we detect this sort of conditionality, we should at minimum warn the user before enabling any redirects, or possibly refuse to do it without some sort of override.

bmw:
After talking about this on the call, we determined the safest thing we can do here is just refuse to enable the redirect if there are already rewrite rules for the vhost. We proposed we could:

1. Add rewrite directives as a comment, telling the user they can uncomment these lines if they are confident it won't create a redirect loop.
2. Dump the directives required to the terminal (this might be a bit uglier when rewrites couldn't be enabled for multiple vhosts).

pde:
Discussed on today's call; we've decided we want this in for 0.10.0 as part of marking the nginx plugin as beta and cautiously encouraging people to use it.

PR #3751, opened 2016-11-05
Don't re-add redirects if one exists

### LINKED ISSUE #3743: Don't re-append Nginx redirect directive
When the user selects Reinstall > Secure and there's already a certbot-managed redirect directive, we just append another one anyway. Don't do that.

ohemorange:
If we've already added a redirect block, don't add it again. If we didn't add it but there's one in there anyway, add a commented version of the block we add. Fixes #3743 and #3252.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -45,6 +45,21 @@ TEST_REDIRECT_BLOCK = [
     ['#', ' managed by Certbot']
 ]
 
+REDIRECT_COMMENT_BLOCK = [
+    ['\n    ', '#', ' Redirect non-https traffic to https'],
+    ['\n    ', '#', ' if ($scheme != "https") {'],
+    ['\n    ', '#', "     return 301 https://$host$request_uri;"],
+    ['\n    ', '#', " } # managed by Certbot"],
+    ['\n']
+]
+
+TEST_REDIRECT_COMMENT_BLOCK = [
+    ['#', ' Redirect non-https traffic to https'],
+    ['#', ' if ($scheme != "https") {'],
+    ['#', "     return 301 https://$host$request_uri;"],
+    ['#', " } # managed by Certbot"],
+]
+
 @zope.interface.implementer(interfaces.IAuthenticator, interfaces.IInstaller)
 @zope.interface.provider(interfaces.IPluginFactory)
 class NginxConfigurator(common.Plugin):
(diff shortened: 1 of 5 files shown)


certbot-nginx/certbot_nginx/obj.py
```diff
@@ -149,3 +150,36 @@ def __eq__(self, other):
...
+        """
+        for i in xrange(0, len(self.raw) - len(test)):
+            if self.raw[i:i + len(test)] == test:
+                return True
+        return False
+
+def _find_directive(directives, directive_name):
+    """Find a directive of type directive_name in directives
+    """
+    if len(directive_name) == 0 or not directives or \
```
pde: What is that first `len(directive_name) == 0` test for? It seems as though it will never occur at present? Would the semantics if called with `directive_name=""` matter?

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -149,3 +150,36 @@ def __eq__(self, other):
...
+        isinstance(directives, str) or len(directives) == 0:
+        return None
+    else:
+        if directives[0] == directive_name:
+            return directives
+        else:
+            for line in directives:
```
pde: if you want to be fancy you could do this:

`return next(_find_directive(line, directive_name) for line in directives), None)`

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -504,17 +539,19 @@ def _enable_redirect(self, domain, unused_options):
...
-
-            self.parser.add_server_directives(
-                vhost, redirect_block, replace=False)
-            logger.info("Redirecting all traffic on port %s to ssl in %s",
-                self.DEFAULT_LISTEN_PORT, vhost.filep)
+            if self._has_certbot_redirect(vhost):
+                logger.info("Traffic on port %s already redirecting to ssl in %s",
+                    self.DEFAULT_LISTEN_PORT, vhost.filep)
+            elif vhost.has_redirect():
+                self._add_redirect_block(vhost, active=False)
```
ohemorange: Not if we were the first ones to ever put a redirect in there. But that's a good point, it should probably also only add a maximum of one of these commented-out guys.

certbot-nginx/certbot_nginx/obj.py
ohemorange: I can't remember why I put that check there. It's possible a holdover from a previous iteration. I'll take it out and if I remember differently put it back in.

certbot-nginx/certbot_nginx/obj.py
ohemorange: I don't understand how this would do the same thing. I want to DFS `directives`, using `None` as an indicator of failure to find a match in a subtree.

pde: Revised suggestion:
```
matches = (_find_directive(line, directive_name) for line in directives)
return next(m for m in matches if m is not None, None)
```

PR #3757, opened 2016-11-06
Fix writing pem files with Python3

kaiyou:
Currently certbot seems broken when using Python3 due to ``bytes`` and ``str`` confusion when manipulating OpenSSL pem dumps.

These commits make sure that OpenSSL dumps are handled as ``bytes`` and written to files using binary write mode. Also, arguments in the ``util``module were prone to confusion (``mode`` used both as open mode and file mode in various functions), switching to ``chmod`` to describe permissions mode and sticking to ``mode`` for file open mode.


kaiyou:
Well, still some more work apparently. Sorry, I will run full coverage tests locally before I update the PR.

pde:
Thanks!

PR #3769, opened 2016-11-08
Remove letsencrypt-auto*

### LINKED ISSUE #3763: Remove letsencrypt* directories from our tree
Unless dramatic changes take place in `pip` or PyPI, we should never need to mess with these packages again. Let's remove them from our tree to clean things up a bit. If we decide we need them in the future, we always have our git history.

nickfong:
Closes #3763


nickfong:
Removing the `letsencrypt-auto-source` directory (and the `letsencrypt-auto`shell script) breaks running of the `modification-check.sh` script, as evidenced by the failing CI.  

I'm updating this script to remove references to `letsencrypt-auto` so that the testing suite will pass locally and (hopefully) in CI.  Keeping the commits separate since I'm only 90% sure that this is what should be done.

pde:
`letsencrypt-auto` needs to stay around (probably forever) for backwards compatibility; `letsencrypt-auto-source` contains the source code for `certbot-auto`; we can't rename it because old versions of `letsencrypt-auto` [hard-code it as a place to check for updates](https://github.com/certbot/certbot/blob/fe0c9c8ff128cc64c67d2f675bb75d6f999ce13a/letsencrypt-auto#L1787):

nickfong:
@pde Thanks for the clarification.  

Reset my fork to the upstream `HEAD` instead of reverting my commit to preserve history (i.e. I want to make sure`blame` and `bisect` still work properly) for `letsencrypt-auto` and `letsencrypt-auto-source/*`.  

Removed the `letsencrypt`, `letsencrypt-apache`, and `letsencrypt-nginx` directories.

PR #3779, opened 2016-11-10
Add workaround for platforms where injecting pyopenssl to urllib3 fails with requests package (Ubuntu Trusty) [revision requested]

oerdnj:
There's a failure on Ubuntu Trusty even with recent requests, urllib3 and pyopenssl that fails to inject pyopenssl into urllib3 via requests package (end with AttributeError).

This adds workaround that will use alternative method for injecting pyopenssl into urllib3 (the upstream recommended method).


acme/acme/client.py
```diff
@@ -27,7 +27,11 @@
 # for SSL, which does allow these options to be configured.
 # https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
 if sys.version_info < (2, 7, 9):  # pragma: no cover
-    requests.packages.urllib3.contrib.pyopenssl.inject_into_urllib3()
+    try:
+        requests.packages.urllib3.contrib.pyopenssl.inject_into_urllib3()
+    except AttributeError:
+        import urllib3.contrib.pyopenssl
```
pde: Will need `# pylint: disable=import-error` here, and maybe also in the clause above.

pde: Possibly also `pragma: no cover` comments similar to the ones we use for other [try-import patterns](https://github.com/certbot/certbot/blob/master/acme/acme/dns_resolver.py#L11)

bmw:
I'm appending `[revision requested]` to the title of this PR so we can see at a glance it's been reviewed. Please remove this when you'd like us to take another look.

PR #3784, opened 2016-11-11
Fix reinstall message

bmw:
Built on and supersedes #3353. This PR stops us from printing messages like:
```
        Your existing certificate has been successfully reinstalled, and the new
        certificate has been installed.
```
which makes a whole lot of no sense.

Also, high level unit tests are silly.


ohemorange:
I just noticed these changes - we might not want to combine patches like this. I think I ran into some problem with addCleanup (maybe python version?) the last time I tried this.

"If you use this technique you must ensure that the patching is “undone” by calling stop. This can be fiddlier than you might think, because if an exception is raised in the setUp then tearDown is not called. unittest.TestCase.addCleanup() makes this easier"
from https://docs.python.org/3.5/library/unittest.mock-examples.html

bmw:
@ohemorange, thanks for mentioning this! I had no idea. This probably isn't a major problem because while failing to remove a `patch` might cause strange behavior, it will only happen if tests are already failing (because an exception was raised). If you want, feel free to open an issue for going through and changing how we use `patch` to avoid this problem. I know this isn't the only place we do it.

bmw:
Adding another comment here to expand on the `mock.patch` discussion: we can't use `addCleanup` as suggested by mock's documentation because it's new in Python 2.7

PR #3785, opened 2016-11-11
Implement the --cert-name flag to select a lineage by its name.

ohemorange:
Step 3 of #3615.


ohemorange:
Pushing to github to see missing comment updates and test coverage, not ready for review.

ohemorange:
Update: this also contains a `--new-cert-name` flag to go along with a `rename` command. I'm starting to think that it may be appropriate to move verbs like `rename`, `certificates`, `update_symlinks`, and `revoke` to an interactive cert manager, but that is definitely not appropriate for this PR.

Once coverage and tests are happy I expect this to be ready for review.

schoen:
Test 1: I have an `--apache` lineage called `tomarbanhoeruim.loyalty.org` and I ran

`certbot certonly --cert-name tomarbanhoeruim.loyalty.org --force-renew`

But I was interactively prompted for an authenticator, which I didn't expect.

I then tried

`certbot run --cert-name tomarbanhoeruim.loyalty.org --force-renew`

In this case the renewal appeared to work, but I got mysterious behavior on the install attempt:

```Renewing an existing certificate
tls-sni-01 challenge for tomarbanhoeruim.loyalty.org
Waiting for verification...
Cleaning up challenges
Generating key (2048 bits): /etc/letsencrypt/keys/0016_key-certbot.pem
Creating CSR: /etc/letsencrypt/csr/0016_csr-certbot.pem
An unexpected error occurred:
TypeError: 'NoneType' object is not iterable

IMPORTANT NOTES:
   /etc/letsencrypt/live/tomarbanhoeruim.loyalty.org/fullchain.pem.
```

(Also not sure why I can get "Congratulations" with the error: I guess because auth worked but install didn't?)

```
2016-11-30 19:40:39,374:DEBUG:certbot.main:Exiting abnormally:
  File "/tmp/certbot/venv/bin/certbot", line 11, in <module>
  File "/tmp/certbot/certbot/main.py", line 806, in main
  File "/tmp/certbot/certbot/main.py", line 555, in run
    lineage.chain, lineage.fullchain)
  File "/tmp/certbot/certbot/client.py", line 371, in deploy_certificate
    for dom in domains:
TypeError: 'NoneType' object is not iterable
```

Can any of these behaviors be attributed to this PR? I can try to run the same commands with an older branch for comparison.

certbot/main.py
```diff
@@ -230,51 +229,66 @@ def _treat_as_renewal(config, domains):
...
+    if config.renew_with_new_domains:
+        return
+    msg = ("Confirm that you intend to update certificate {0} "
+           "to include domains {1}. Note that it previously "
+           "contained domains {2}.".format(
+               certname,
+               new_domains,
+               old_domains))
+    obj = zope.component.getUtility(interfaces.IDisplay)
+    if not obj.yesno(msg, "Update cert", "Cancel"):
```
pde: Thanks to @PaulSD for noticing this: you should add `default=True` to this `obj.yesno()` call, so that this path can be run non-interactively.

certbot/cert_manager.py
```diff
@@ -30,6 +33,29 @@ def update_live_symlinks(config):
...
+
+    :param config: Configuration.
+    :type config: :class:`certbot.interfaces.IConfig`
+
+    """
+    if not config.certname:
+        raise errors.ConfigurationError("Specify a certificate name with "
+            "with flag --cert-name.")
+    if not config.new_certname:
+        raise errors.ConfigurationError("Specify a new name for certificate {0} "
```
pde: This should be `errors.MissingCommandlineFlag`.

But even better would be to do something like calling `disp.input("Enter the new name for certificate {0}".format(config.certname), flag="--new-cert-name")`. That will raise a nicely formatted `errors.MissingCommandlineFlag` for you.

certbot/cert_manager.py
```diff
@@ -104,3 +130,69 @@ def certificates(config):
...
+            logger.debug("Renewal conf file %s is broken. Skipping.", renewal_file)
+            logger.debug("Traceback was:\n%s", traceback.format_exc())
+            continue
+        rv = func(candidate_lineage, rv)
+    return rv
+
+def lineage_for_certname(config, certname):
+    """Find a lineage object with name certname.
+    """
+    def func(candidate_lineage, rv):
```
pde: Style nit: Maybe nicer to call this `matches_certname` or `certname_match_fn` or something?

certbot/cli.py
```diff
@@ -687,6 +688,19 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              "multiple -d flags or enter a comma separated list of domains "
              "as a parameter.")
     helpful.add(
+        [None, "run", "certonly"],
+        "--cert-name", dest="certname",
+        metavar="CERTNAME", default=None,
+        help="Certificate name to apply. Only one certificate name can be used "
+             "per Certbot run. Show certificate names by running certificates "
```
pde: Yes and maybe also better to say something like `by running "certbot certificates"` or ` by running cerbot with the "certificates" subcommand`?

certbot/main.py
```diff
@@ -68,17 +70,17 @@ def _report_successful_dry_run(config):
                                   reporter_util.HIGH_PRIORITY, on_crash=False)
 
 
-def _auth_from_domains(le_client, config, domains, lineage=None):
+def _auth_from_available(le_client, config, domains=None, certname=None, lineage=None):
     """Authenticate and enroll certificate.
```
pde: Maybe we should document what is going on here a little bit more?

certbot/main.py
```diff
@@ -172,7 +172,7 @@ def _handle_identical_cert_request(config, lineage):
         return "reinstall", lineage
     question = (
         "You have an existing certificate that contains exactly the same "
-        "domains you requested and isn't close to expiry."
+        "domains or certificate name you requested and isn't close to expiry."
```
pde: If you're feeling fancy, provide a question whose wording varies depending on whether the user actually provided domains or a certificate name (or if you agree that's a good idea but don't want to do it in this PR, open a new ticket for it).

certbot/storage.py
```diff
@@ -96,6 +96,25 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
...
+    """Rename's cli_config.certname's config to cli_config.new_certname.
+
+    :param .RenewerConfiguration cli_config: parsed command line
+        arguments
+    """
+    prev_filename = os.path.join(
+        cli_config.renewal_configs_dir, prev_name) + ".conf"
+    new_filename = os.path.join(
+        cli_config.renewal_configs_dir, new_name) + ".conf"
+    if os.path.isfile(new_filename):
```
pde: I think this should probably be `os.path.exists(new_filename)`, so that it errors on directories, links, etc.

ohemorange:
@schoen could you try running the weird behaviors above on master? I don't have apache set up.

certbot/cert_manager.py
ohemorange: ah! I didn't know we had that but I like it!

certbot/cert_manager.py
```diff
@@ -104,3 +130,69 @@ def certificates(config):
...
+    rv = initial_rv
+    for renewal_file in renewal.renewal_conf_files(cli_config):
+        try:
+            candidate_lineage = storage.RenewableCert(renewal_file, cli_config)
+        except (errors.CertStorageError, IOError):
+            logger.debug("Renewal conf file %s is broken. Skipping.", renewal_file)
+            logger.debug("Traceback was:\n%s", traceback.format_exc())
+            continue
+        rv = func(candidate_lineage, rv)
+    return rv
```
ohemorange: Flexibility is my goal! I'll put this in the docstring to make it clear.

certbot/cert_manager.py
ohemorange: This function returns a lineage object for a given certname. Is it that the name is too long?

certbot/cli.py
ohemorange: Good call. I like `To see certificate names, run "certbot certificates".`

certbot/cli.py
```diff
@@ -687,6 +688,19 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+             "command. If there is no existing certificate with this name and "
+             "domains are requested, create a new certificate with this name.")
+    helpful.add(
+        "rename",
+        "--new-cert-name", dest="new_certname",
+        metavar="NEW_CERTNAME", default=None,
+        help="New name for the certificate. Must be a valid filename.")
```
ohemorange: Oh good point. Maybe `--updated-cert-name` is better? Or just `--new-certname` and then also call it `--certname` in other places? I don't want to make this flag for anything other than setting a new cert name.

certbot/cli.py
```diff
@@ -739,6 +753,12 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              "--keep-until-expiring is more appropriate). Also implies "
              "--expand.")
     helpful.add(
+        "automation", "--renew-with-new-domains",
+        action="store_true", dest="renew_with_new_domains", help="If a "
+             "certificate already exists for the requested certificate name "
+             "but does not match the request domains, renew it now, "
```
ohemorange: So it should.

certbot/main.py
ohemorange: Ha yes. Especially given that this is the primary `main` authentication method, as far as I can tell.

certbot/main.py
```diff
@@ -230,51 +229,66 @@ def _treat_as_renewal(config, domains):
...
+                raise errors.ConfigurationError("No certificate with name {0} found. "
+                    "Use -d to specify domains, or run certbot --certificates to see "
+                    "possible certificate names.".format(certname))
+
+def _ask_user_to_confirm_new_names(config, new_domains, certname, old_domains):
+    """Ask user to confirm update cert certname to contain new_domains.
+    """
+    if config.renew_with_new_domains:
+        return
+    msg = ("Confirm that you intend to update certificate {0} "
```
ohemorange: I have no opinion about this. Let me know if you want me to change it.

certbot/main.py
ohemorange: I had added a cli flag because I didn't know about this `default=True` thing. It seems like we usually have both a noninteractive option and a cli flag, is that right? Or should I remove the cli flag given that it can be run noninteractively?

certbot/storage.py
```diff
@@ -96,6 +96,25 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
         config.write(outfile=f)
     return config
 
+def rename_renewal_config(prev_name, new_name, cli_config):
+    """Rename's cli_config.certname's config to cli_config.new_certname.
```
ohemorange: lol

certbot/cert_manager.py
ohemorange: In addition, the previous check can probably provide a list of cert names.

ohemorange: ok it now does that.

schoen:
@ohemorange, yes, I'll try getting a new lineage with master, then running the closest equivalents to those, then running those with this branch, and see if the problem recurs or not.

certbot/main.py
schoen: I guess I'm also not sure, so I won't suggest a change.

certbot/tests/cert_manager_test.py
```diff
@@ -178,5 +187,219 @@ def test_report_human_readable(self):
...
+        domains = ["example.com", "example.org"]
+        mock_match.names.return_value = domains
+        mock_renewable_cert.return_value = mock_match
+        from certbot import cert_manager
+        self.assertEqual(cert_manager.domains_for_certname(None, "other.com"), None)
+        self.assertTrue(mock_make_or_verify_dir.called)
+        self.assertTrue(mock_renewer_config)
+
+
+class RenameLineageTest(storage_test.BaseRenewableCertTest):
```
ohemorange: There's a `BaseCertManagerTest` that will work here, why not use that?

ohemorange: What a good idea, I should do that!

certbot/cert_manager.py
pde: My nit was suggesting something slightly more descriptive instead of `func`. But feel free to ignore this suggestion!

certbot/main.py
pde: Yup; it was already mysterious, and is becoming more so :)

ohemorange: How do you feel about the current documentation?

certbot/cert_manager.py
ohemorange: Oh that makes more sense! I can do that.

certbot/main.py
ohemorange: Let's keep both, because a user might not know about non-interactive mode and just look for the specific command-line flag they want.

peterthomassen: While I don't feel strongly, I think it is the user's responsibility to look at the documentation of the available flags more closely if he doesn't immediately find what (s)he needs.

Furthermore, if you're running in interactive mode, I don't think it is so bad to get an extra question; in other words, I don't think there needs to be a switch to mute specific aspects of the interactive mode. (The number of questions is low anyways.) It's a trade-off with code simplicity, which I also consider valuable.

ohemorange:
Make coveralls report back?

certbot/main.py
pde: LGTM

PR #3805, opened 2016-11-16
Take advantage of urllib3 pyopenssl rewrite

bmw:
Built on #3803 so tests pass.

As I described in #3803, `pyasn1` and `ndg-httpsclient` are no longer required in newer versions of `requests`. Let's take advantage of that. I bumped the version of `requests` installed with `certbot-auto` so we don't need to install these packages.

Additionally, this PR removes our direct dependency on these two packages in `acme` and instead has us depend on `requests` "security" setuptools extras. This allows us to depend on `pyasn1` and `ndg-httpsclient` when they are needed and not depend on them when they're not. This would fix the theoretical problems identified in #3803, but [pip doesn't offer real dependency resolution](https://github.com/pypa/pip/issues/988).

I added the minimum version requirement to `requests` because the security extras are not available in older versions. The only packaging efforts this may affect is the Trusty PPA as the version of `requests` available in the Trusty repos does not have these setuptools extras.

@oerdnj, will this be a problem for you? The version of `requests` available in the Trusty repos will work just fine, however, you'll need to make sure the `acme` package depends on `pyasn1` and `ndg-httpsclient` and you may need to modify `acme/setup.py` to prevent errors about missing dependencies.


oerdnj:
@bmw I have requests 2.11.1 in the Ubuntu Trusty backports PPA and @hlieberman has at least 	2.8.1 in the Debian Jessie backports, so I don't think this will cause any disturbance.

erikrose:
lgtm!

bmw:
Fixed merge conflicts.

With Erik's LGTM, @ohemorange or @pde, can you add an approving review to this so it can be merged?

pde:
@noodlesnz @hogarthj does/can EPEL have a new enough version of requests to live with this?

pde:
This is progress on #1301

NoodlesNZ:
@pde EPEL7 has python-requests 2.12.1, does this meet the requirements here?

bmw:
Fixed merge conflicts.

Yes, the version of `python-requests` in EPEL is new enough. Any version 2.4.1+ will work just fine.

Issue #3761, opened 2016-11-07
certbot-auto Alpine support


willfarrell:

pde:
Send us a pull request!

bmw:
Reason for closing is explained [here](https://github.com/certbot/certbot/pull/3929#issuecomment-294992843).

PR #3815, opened 2016-11-22
Add bootstrapping of alpine deps

willfarrell:
Closes: #3761

- [x] Update letsencrypt-auto source
- [x] Add Travis-CI Alpine Docker Test
- [x] Pass: Make sure we stop if the openssl signature check fails.
- [ ] Pass: Make sure pip stops us if there is a hash mismatch.
- [ ] Pass: Exercise most branches of letsencrypt-auto.
- [ ] Docs if needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,11 @@ matrix:
       services: docker
       before_install:
       addons:
+    - sudo: required
+      env: TOXENV=le_auto_alpine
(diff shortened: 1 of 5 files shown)


willfarrell:
oops, looks like I was a little too aggressive with the rebase. Will re-PR. Sorry.

Issue #3788, opened 2016-11-12
please let me reuse a key during cert renewal


dkg:
I have a cert that needs to keep the same public key over a couple years at least.

I like the short cert lifetimes, i just want to renew with the same public key.

But certbot automatically generates a new key at each renewal event.  Please give "certbot renew" the option of re-using the existing key.

pde:
There is already a flag `--key-path`, that should probably have these semantics.

pde:
The CLI entry for `key-path` is currently way too complicated;

```
    section = "paths"
    if verb in ("install", "revoke"):
        section = verb
    # revoke --key-path reads a file, install --key-path takes a string
    add(section, "--key-path", required=(verb == "install"),
        type=((verb == "revoke" and read_file) or os.path.abspath),
        help="Path to private key for cert installation "
             "or revocation (if account key is missing)")
```

The `section` logic is obsolete since #3254 landed; we can just make the section be `["paths", "install", "revoke"]`. Then the place to implement the actual key reuse logic would probably be to take an alternative path [here](https://github.com/certbot/certbot/blob/master/certbot/client.py#L259) if `self.config.key_path` is set.

pde:
@dkg send us a pull request!

thomaszbz:
There's also a technical use case for this (be it `--key-path` or `--reuse-key` like suggested by @schoen in the PR): DNSSEC/DANE, thereby not pinning the CA but the public key itself. Due to the lack of the corresponding private key, even LE could not issue a fraudulent cert for that (at least if a client validates it...)

ArchangeGabriel:
I agree with @thomaszbz.

noqu:
Unless I am much mistaken, users of HPKP (HTTP Public Key Pinning) would also greatly benefit from `--reuse-key`.

With HPKP, in order to migrate from one public key to another, it is necessary to adapt the server's secondary `pin-sha256` header well in advance of the actual migration, so that clients will accept the (future) new key as genuine.

This is definitely not something you would want to do every three months.

ArchangeGabriel:
Yes indeed, but HPKP is dying and DNSSEC/DANE is what people should be focusing on. ;)

vbertola:
[...]

PR #3819, opened 2016-11-23
Add ability to reuse key during cert renewal (#3788)

supermari0:


supermari0:
This should address #3788 

I don't have a server with a cert that needs renewal, so I haven't tested whether this actually works - hoping that someone can pull this down and try it in their environment or add an integration test for this before merging

pde:
@supermari0 We'll try to take a look at this soon but in the mean time, in order to test your work you can trigger renewal with `--force-renew` :)

pde:
You can also use `--test-cert` to get your certs from the staging server, so you never hit rate limits while playing around with this sort of thing.

certbot/client.py
```diff
@@ -255,9 +255,22 @@ def obtain_certificate(self, domains):
...
+        if self.config.key_path is None:
+            key = crypto_util.init_save_key(
+                self.config.rsa_key_size, self.config.key_dir)
+        else:
+            try:
+                key_pem = open(self.config.key_path, 'r').read()
+                if not crypto_util.valid_privkey(key_pem):
+                    raise errors.Error("Invalid private key specified")
+            except Exception:
+                msg = ("Unable to load private key from specified key file %s."
```
supermari0: I need to fill in the string interpolation variable here

certbot/tests/client_test.py
```diff
@@ -178,8 +179,7 @@ def test_obtain_certificate_from_csr(self, mock_logger):
             test_csr)
         mock_logger.warning.assert_called_once_with(mock.ANY)
 
-    @mock.patch("certbot.client.crypto_util")
-    def test_obtain_certificate(self, mock_crypto_util):
+    def _test_obtain_certificate(self, mock_crypto_util, key_to_check=None):
```
supermari0: This can probably be refactored further to use with other test_obtain_cert.* methods, but I will leave it for now if there are no objections

schoen:
Hi @supermari0, thanks for working on this. It looks like you're creating a `--key-path` to explicitly specify a private key to be used; for the re-use case, wouldn't it be simpler for the user if we just had `--reuse-key` that would use the most recent private key that exists in the lineage?

pde:
@supermari0 do you want to make whatever changes you think are necessary, and then ping schoen for review when you're happy with the state of the PR?

supermari0:
@pde sounds good to me. If you all need this change sooner, feel free to take this as a base and change it however is needed. Thanks

supermari0:
I'll be working on this change again today, so hopefully I will be able to either add the more useful ``--reuse-key`` button or both.

certbot/client.py
```diff
@@ -255,11 +257,41 @@ def obtain_certificate(self, domains):
...
+                       " File must be a valid RSA private key in PEM format." %
+                       str(self.config.key_path))
+                logger.warning(msg)
+                raise
+            key = util.Key(self.config.key_path, key_pem)
 
+        csr = crypto_util.init_save_csr(key, domains, self.config.csr_dir)
+        # TODO(mariojv) Should we wrap the potential exception thrown by this?
```
supermari0: this should be resolved before merge

certbot/tests/client_test.py
```diff
@@ -107,7 +107,11 @@ class ClientTest(unittest.TestCase):
 
     def setUp(self):
         self.config = mock.MagicMock(
-            no_verify_ssl=False, config_dir="/etc/letsencrypt", allow_subset_of_names=False)
+            no_verify_ssl=False,
+            config_dir="/etc/letsencrypt",
+            work_dir="/var/lib/letsencrypt",
```
supermari0: this additional parameter (work_dir) is probably pointless

certbot/tests/client_test.py
```diff
@@ -203,16 +206,60 @@ def test_obtain_certificate(self, mock_crypto_util):
...
+                mock.sentinel.key, domains, self.config.csr_dir)
+        else:
+            fake_key = util.Key(self.config.key_path, key_to_check)
+            self.assertEqual(
+                self.client.obtain_certificate(domains),
+                (mock.sentinel.certr, mock.sentinel.chain, fake_key,
+                 csr))
+            self.assertFalse(mock_crypto_util.init_save_key.called)
+            mock_crypto_util.init_save_csr.assert_called_once_with(
+                fake_key, domains, self.config.csr_dir)
```
supermari0: new test case for --renew-key needs to be added here

supermari0:
I've switched from only having the ``--key-path`` option and now made this PR support both ``--key-path`` and ``--reuse-key``. I still need to:

1. Add unit tests for the ``--reuse-key`` path
2. Figure out how to call the CSR + key validation method properly, or how to mock out that method appropriately
3. Test on a real web server against LE staging endpoint

I'll ping the reviewers when this is done, but I've put up my work so far for initial review.

certbot/constants.py
```diff
@@ -32,6 +32,7 @@
     auth_cert_path="./cert.pem",
     auth_chain_path="./chain.pem",
     strict_permissions=False,
+    reuse_key=False,
```
hlieberman: For the record, though I don't think anyone has proposed this seriously, I would vote to reject any change of reuse becoming the default.

supermari0: +1

This is a new option, so setting it to False to explicitly preserve existing behavior is the best course of action IMHO.

Perhaps I should add a comment to that effect above this line?

jsha:
Sorry to chime in after @schoen already gave feedback and you implemented `--reuse-key`, but I think we should offer only `--key-path`. In general we should strive to keep the number of flags to a minimum, and `--key-path` is strictly more powerful than `--reuse-key`. It can meet a wider variety of needs.

The only reason to additionally offer `--reuse-key` is, as Seth said, to make this particular use case easier. However, the only use cases where retaining the same key matters are already advanced options, like using DANE TLSA with certain configurations, or using HPKP. I think it's acceptable to expect users of those advanced options to specify a full path to a key. Especially because the key path will be saved in the renewal config, so the path doesn't need to be specified again on renewal. @schoen, @bmw, what do you think?

bmw:
I'd like to see Seth's comments on this since he's the one that purposed `--reuse-key`.

With that said, I think there is a use case where `--key-path` is significantly more cumbersome than `--reuse-key`. If you want Certbot to generate your key for you on your first run and then reuse the same key thereafter, you'll either manually have to modify the renewal configuration file (which we currently don't encourage or really tell people how to do) to add `key_path` or renew your cert once manually using `certonly`/`run` including the `--key-path` flag so it's added to your configuration file for you. It's certainly doable, but to me it feels a little harder than it should be.

For my personal opinion, pretending for a minute this PR doesn't exist, I think I'd prefer to initially only support `--key-path` and see if something like `--reuse-key` is a commonly requested feature. Since this feature has already been implemented though at the request of a Certbot developer and it improves the UX in some cases, I'm inclined to take this PR with `--reuse-key`.

jsha:
>  If you want Certbot to generate your key for you on your first run and then reuse the same key
> thereafter, you'll either manually have to modify the renewal configuration file (which we currently don't
> encourage or really tell people how to do) to add key_path or renew your cert once manually using
> certonly/run including the --key-path flag so it's added to your configuration file for you.

The current PR doesn't fix that, since it defines `--reuse-key` as a flag to the renew subcommand rather than to the issuance subcommands. Certainly it could be changed to fix that, but given that the current state of `--reuse-key` in this PR is not quite right, I'd rather move it in the direction of fewer new flags, and as you say, see if there's still a lot of demand for `--reuse-key` once `--key-path` works.

Also, again, all appreciation and apologies to @supermari0 for the mixed messages here. Thanks for your patience while we sort things out!

bmw:
> The current PR doesn't fix that, since it defines --reuse-key as a flag to the renew subcommand rather than to the issuance subcommands. Certainly it could be changed to fix that, but given that the current state of --reuse-key in this PR is not quite right, I'd rather move it in the direction of fewer new flags, and as you say, see if there's still a lot of demand for --reuse-key once --key-path works.

You're correct that it doesn't work as is. If your workflow is to generate a cert once using `certbot certonly/run --reuse-key` and then use `certbot renew`, `--reuse-key` doesn't have to be respected for `certonly/run`. It just needs to be saved for renewal (which to make happen is a one line change to `renewal.py`). With that said, `--reuse-key` probably should be respected when updating lineages with `certonly/run` as well.

I think jsha is right here. If we already had `--reuse-key` implemented in the way we'd want to support it, I'd be more inclined to take the PR since some people want that feature, but since this PR needs a little more work, I think we should just remove `--reuse-key` for now and just use `--key-path`. As jsha said, `--key-path` is strictly more powerful than `--reuse-key`. Anything you could do with `--reuse-key` you can do with `--key-path`.

schoen:
I would like to have both `--reuse-key` and `--key-path`, but I'm doing some work on my own branch now that uses only `--key-path` for the time being.

jsha:
@schoen does that mean you are taking over this work in your own branch? Or are you building on top of this? Should @supermari0 continue work on this branch?

schoen:
I'm going to take this over in my own branch (although maybe we can use some of the code from this PR later on when adding --reuse-key, which I still personally hope we can do as well).

sydneyli:
Closing this PR in favor of #4610

PR #3828, opened 2016-11-29
[changes requested] Refactor cli_test.py and main_test.py

### LINKED ISSUE #2716: Split cli_tst.py?
The incentive to do this is fairly low compared to the crazy amount of work involved, but `cli_tst.py` should probably be split into pieces to test `main`, `cli` and `plugins.selection` separately.

cowlicks:
This moves the tests in `cli_test.py` that used the `certbot.main.main` entrypoint to main_test.py. It also moves a lot of the other extraneous tests out of there.

It does not fully close #2716 because many of the tests now in `main_test.py` still use the same   monolithic entrypoint, but it is a good step toward that.


pde:
This looks like heroic work!

certbot/tests/error_handler_test.py
```diff
@@ -112,6 +113,59 @@ def test_sysexit_ignored(self):
             pass
         self.assertFalse(self.init_func.called)
 
+    @mock.patch('certbot.main.sys')
+    def test_handle_exception(self, mock_sys):
```
ohemorange: It's confusing to me to have `_handle_exception()` defined in `main.py` but tested in `error_handler_test.py`. I don't know the history of why it's in `main`, but could we move it to `error_handler.py` to match the test?

certbot/tests/main_test.py
```diff
@@ -194,5 +211,833 @@ def test_failure(self, mock_make_or_verify):
...
+        self.assertEqual(result[0], None)
+        self.assertTrue(result[1].configfile.filename.endswith('example.org.conf'))
+
+        # Partial overlap doesn't count
+        result = _find_duplicative_certs(
+            self.cli_config, ['example.com', 'something.new'])
+        self.assertEqual(result, (None, None))
+
+
+class MainTest(unittest.TestCase):  # pylint: disable=too-many-public-methods
```
ohemorange: How would you feel about having a BaseMainTest class that the different verb testers inherit from?

certbot/tests/renewal_test.py
```diff
@@ -0,0 +1,30 @@
+"""Tests for certbot.renewal"""
+import os
+import mock
+import unittest
+import tempfile
+
+from certbot import configuration
+from certbot import renewal
```
ohemorange: Can you move this into the method, just before calling `_restore_webroot_config`?

certbot/tests/test_util.py
```diff
@@ -111,3 +115,46 @@ def skip_unless(condition, reason):  # pragma: no cover
         return lambda cls: cls
     else:
         return lambda cls: None
+
+
+def make_lineage(self, testfile):
```
ohemorange: With this method in `test_util`, does `DuplicativeCertsTest` have to inherit from `BaseRenewableCertTest`?

ohemorange:
Oh also this is going to conflict with #3785, so let's try to get this one in.

cowlicks:
@pde @ohemorange I think I should rename `test_util.py` to `util.py` to avoid confusing with `util_test.py`. And the name information is already there in the import path `certbot.tests.test_util`. Any objections?

ohemorange:
Only reason I see to keep it is that we might then get some `import certbot.tests.util as test_util` to avoid conflicts with the other `util`. No strong opinion either way.

cowlicks:
@ohemorange cool, I'll probably have time to address your comments tomorrow morningish CST.

certbot/tests/error_handler_test.py
cowlicks: I looks like the purpose of `error_handler.ErrorHandler` is to handle clean up of stateful operations when there is an exception. But `main._handle_exception` is for logging tracebacks and displaying them to the user properly. So they are separate but sound similar.

I think it would be better to move this test into its own class in main_test.py.

certbot/tests/main_test.py
cowlicks: That sounds like a good strategy. But I think I will save breaking this into separate verb tests for another PR. This PR just gets the low hanging fruit, like the parsing tests.

certbot/tests/test_util.py
cowlicks: Yeah, it looks like `make_lineage` can be used instead of inheriting from `BaseRenewableCertTest`. I'll try that this afternoon.

certbot/tests/main_test.py
ohemorange: Sounds good! I'll make you an issue for it.

certbot/tests/error_handler_test.py
ohemorange: Either way works for me, as long as they match.

certbot/tests/renewal_test.py
cowlicks: Got it

certbot/tests/test_util.py
cowlicks: I didn't finish this today, I'll take another look tomorrow.

pde:
Kicking Travis

certbot/tests/test_util.py
cowlicks: @ohemorange, a few things for posterity, `make_lineage` creates as lineage based on any of the configuration files in `certbot/tests/testdata/*.conf`. While `BaseRenewableCertTest` makes a lineage (correct word?) on the fly, based on some parameters in its class definition. It would be nice to remove the dependency of `BaseRenewableCertTest` since it is imported from another test file.

After a few attempts I could not get the `DuplicativeCertsTest` tests to work using  `make_lineage`.  So I'll punt for now and make another issue.

certbot/tests/main_test.py
```diff
@@ -194,5 +211,887 @@ def test_failure(self, mock_make_or_verify):
...
+        self.assertRaises(
+            errors.Error, self._call,
+            'certonly -d example.org --csr {0}'.format(CSR).split())
+
+    def _get_argument_parser(self):
+        plugins = disco.PluginsRegistry.find_all()
+        return functools.partial(cli.prepare_and_parse_args, plugins)
+
+    # why can't I remove this?
+    def test_option_was_set(self):
```
cowlicks: A lot of this refactoring expects that `prepare_and_parse_args` is stateless, but it actually has some global side effects which are tested here. Is there a better way to handle this state?

cowlicks:
Rebased

certbot/tests/main_test.py
cowlicks: Fixed. I just manually reset the state of `set_by_cli` in each test. This is way cheaper and faster than loading all the plugins and rebuilding the parser for each test.

cowlicks:
@ohemorange I think I got everything. Anything else?

Issue #3729, opened 2016-11-02
Remove sphinxcontrib-programoutput dependency?


bmw:
It's not maintained and the owner chose to delete the project because no one offered to take it over. It was reuploaded to PyPI when everyone's builds broke, but we might want to look into removing this dependency entirely to prevent more problems in the future.

See lunaryorn/sphinxcontrib-programoutput#10 for more information.

bmw:
Supposedly we don't even use this dependency (any more). If this is the case, this is a very easy fix.

bmw:
It looks like we do use this in https://github.com/certbot/certbot/blob/master/acme/docs/man/jws.rst

pde:
OK, if you still want to remove the dependency you can do the same thing I did to create the certbot man page [here](https://github.com/certbot/certbot/blob/master/tools/release.sh#L149) and [here](https://github.com/certbot/certbot/blob/master/docs/man/certbot.rst).

bmw:
Fixed in #3830.

PR #3830, opened 2016-11-30
Remove the sphinxcontrib.programout [docs]dependency

pde:
- By making a static text file at release time, and including that
   instead.


pde:
Closes: #3729

PR #3836, opened 2016-12-01
Parallalelise nosetests from tox

pde:
It turn out that [travis containers have two cores](https://docs.travis-ci.com/user/migrating-from-legacy/#More-available-resources) and we can use them by just passing some flags to `nosetests`.

However we have some tests that are no thread-safe. For instance there are a couple of Reporter tests that seem to be broken by this; we could either fix them or use some of [these control parameters](https://nose.readthedocs.io/en/latest/doc_tests/test_multiprocess/multiprocess.html) to contorl/disable parallelism in those places.


pde:
I'm not seeing quite as much speedup here as I'd expected -- it's about 15% faster when running `nosetests --processes=12` on a six core machine. It's possible we need to [mark some of these fixtures as reentrant](https://nose.readthedocs.io/en/latest/plugins/multiprocess.html) to get this to work really well.

pde:
Turns out you have to set `_multiprocess_can_split_ = True` inside the largest test suites to really benefit.

pde:
OK, this is pretty good. It speeds things up more than 2x on obs7, and probably by 20-30 seconds for a full Travis run.

pde:
Trying to kick coveralls?

certbot/tests/reporter_test.py
```diff
@@ -19,8 +20,10 @@ def setUp(self):
     def tearDown(self):
         sys.stdout = self.old_stdout
 
-    def test_multiline_message(self):
+    @mock.patch('certbot.reporter.os.getpid')
+    def test_multiline_message(self, mock_getpid):
         self.reporter.add_message("Line 1\nLine 2", self.reporter.LOW_PRIORITY)
+        mock_getpid.return_value = self.rep_module.INITIAL_PID
         self.reporter.atexit_print_messages()
```
bmw: We can just use `self.reporter.print_messages()` here and undo all other changes to this test.

certbot/tests/reporter_test.py
```diff
@@ -39,8 +42,10 @@ def test_no_tty_print_empty(self):
             self.reporter.print_messages()
         self.assertEqual(sys.stdout.getvalue(), "")
 
-    def test_atexit_print_messages(self):
+    @mock.patch('certbot.reporter.os.getpid')
+    def test_atexit_print_messages(self, mock_getpid):
         self._add_messages()
+        mock_getpid.return_value = self.rep_module.INITIAL_PID
         self.reporter.atexit_print_messages()
```
bmw: Instead of setting `self.rep_module` to the `reporter` module in `setUp` (which is specifically imported that way to minimize the amount of the tests that have access to the module) you can also just mock out `INITIAL_PID` so you can control both values. You can do this with a context manager like this:
```
with mock.patch('certbot.reporter.INITIAL_PID', 5):
```
or as a decorator like this:
```
    @mock.patch('certbot.reporter.INITIAL_PID', 5)
    @mock.patch('certbot.reporter.os.getpid')
    def test_atexit_print_messages(self, mock_getpid):
```
Notice in the latter example, an additional argument isn't added to the function.

tox.cover.sh
```diff
@@ -37,7 +42,7 @@ cover () {
   # specific package directory; --cover-tests makes sure every tests
   # is run (c.f. #403)
   nosetests -c /dev/null --with-cover --cover-tests --cover-package  \
-            "$1" --cover-min-percentage="$min" "$1"
+            "$1" --cover-min-percentage="$min" "$1" $proc
```
bmw: nit: Since all packages use the same number of processes, why not just put that down here instead of setting it to the same thing in each if block? In the if block, you can just set a `timeout` value, which defaults to 10 for `nosetests`.

pde:
I haven't seen any travis failures on this, but when I run it on a much faster machine with many more cores, I sometimes see this intermittent failure:

```======================================================================
ERROR: test__do_for_subarray (certbot_nginx.tests.parser_test.NginxParserTest)
  File "/home/pde/certbot/certbot-nginx/certbot_nginx/tests/parser_test.py", line 20, in setUp
    super(NginxParserTest, self).setUp()
  File "/home/pde/certbot/certbot-nginx/certbot_nginx/tests/util.py", line 29, in setUp
    "etc_nginx", "certbot_nginx.tests")
  File "/home/pde/certbot/certbot/plugins/common.py", line 290, in dir_setup
    test_configs, os.path.join(temp_dir, test_dir), symlinks=True)
  File "/usr/lib/python2.7/shutil.py", line 208, in copytree
    raise Error, errors
Error: [('/home/pde/certbot/certbot-nginx/certbot_nginx/tests/testdata/etc_nginx/minimalistic_comments.new.conf', '/tmp/tmpy_E3Jttemp/etc_nginx/minimalistic_comments.new.conf', "[Errno 2] No such file or directory: '/home/pde/certbot/certbot-nginx/certbot_nginx/tests/testdata/etc_nginx/minimalistic_comments.new.conf'")]
```

certbot/tests/reporter_test.py
```diff
@@ -42,11 +39,9 @@ def test_no_tty_print_empty(self):
             self.reporter.print_messages()
         self.assertEqual(sys.stdout.getvalue(), "")
 
-    @mock.patch('certbot.reporter.os.getpid')
-    def test_atexit_print_messages(self, mock_getpid):
+    def test_atexit_print_messages(self):
         self._add_messages()
-        mock_getpid.return_value = self.rep_module.INITIAL_PID
-        self.reporter.atexit_print_messages()
+        self.reporter.print_messages()
```
bmw: The problem with this is it reduces test coverage and this test was specifically written to test `atexit_print_messages` as suggested by the name.

bmw:
I suspect intermittent failures are caused by lines like https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/tests/nginxparser_test.py#L131.

There (and in a few other places) we modify `testdata` while tests are running. If another process is reading from the directory at the same time trying to copy it, I could see there being problems. We should fix the test to stop doing that.

pde:
Your changes to the nginx tests LGTM. I can no longer reproduce those failures.

PR #3843, opened 2016-12-02
Adopt consistent linting practices for the entire tree

pde:
For historical reasons we have different, stricter linting standards for the `acme/` module. Since the maintainer who advocated strongly for those is no longer active, it's probably saner for us to have consistent standards across the entire tree.

As a side benefit, this allows further parallelisation of our linting tasks. On my system that reduces the `tox -e lint` run time from 69 seconds to 50 seconds (that's on top of the benefits of #3836, upon which this PR builds).


pde:
Kicking travis

Issue #3844, opened 2016-12-03
Renewal: Preserve 'OCSP Must Staple' (option --must-staple)


thomaszbz:
### Description:

Via #2626, Certbot creates a certificate with `OCSP Must Staple` [TLS extension](https://tools.ietf.org/html/rfc7633) when called with option `--must-staple`. Please note that this TLS extension is contained in the generated certificate file itself.

However, the option `--must-staple` is not stored in `/etc/letsencrypt/renewal/www.example.com.conf`. In my case:

```
[renewalparams]
installer = apache
authenticator = apache
rsa_key_size = 4096
account = abc
```

There also seems not to be an entry for `OCSP Must Staple` in the [test file](https://github.com/certbot/certbot/blob/master/certbot/tests/testdata/sample-renewal.conf#L8).

When executing `certbot renew` for renewal, certbot renews certificates _without_ `OCSP Must Staple`, extension respectively.

### Suggested Fix:

- For renewal, certbot should be aware that the old certificate was generated with  `--must-staple`. That information should be stored in the `[renewalparams]` section, respectively.
- For `certbot renew`, the new certificate should be generated with `--must-staple`, in case the old certificate was generated with that extension.

### Use case:

- I have some certificates which I want to renew with `--must-staple`, because e.g. apache supports that.
- I have other certificates which I use with other software (e.g. an ancient mailserver) which does not support OCSP at all.
- For `certbot renew`, certbot should be aware which one is which, using `[renewalparams]`. Otherwise, `certbot renew` would not qualify to be used for automation in my case.

### Version:
```
certbot --version
certbot 0.9.3
```

OS: Debian 8, with certbot installed via official debian repository.

pde:
Thanks for catching this!

pde:
We should think about whether any other security enhancements should also be stored in renewal conf files -- most of them don't need to be, because they're one-off calls to the `IInstaller`.

thomaszbz:
related: #3394

thomaszbz:
Now that PR #3948 got merged and includes my commits, I think this issue should be solved and can be closed.

PR #3849, opened 2016-12-03
Preserve --must-staple in configuration for renewal (#3844)

thomaszbz:
See #3844: Preserves `--must-staple` for renewal.

This patch is completely untested! Review carefully (my first commit to certbot, and my first loc in python).

All I can say is that the testsuite still runs in docker and that I found a few places that somehow qualify to be patched for #3844.

If someone knows how to easily simulate/bypass the whole domain checking process in Docker (with e.g. an untrusted cert, of course), I could eventually test it on my dev system. `--dry-run` and `--staging` still did domain checking.

To be tested:

- [ ]  when `certbot` is executed, a new entry `must_staple` must occur in configuration. It must be `True` when called with `--must-staple` and `False` otherwise.
- [ ] `certbot renew` must respect the config variable on basis of a certificate.


pde:
@schoen do you have time to review this?

schoen:
@thomaszbz, I liked this (and it's exactly parallel to the existing types), except that I questioned the use of `eval`. Please see what you think!

Thanks.

thomaszbz:
@schoen I agree, I did not like this as well. It's did it similar to a few lines above, for consistency.

To avoid the eval(), I can just return the result of a comparison ("True" or "False" are checked before).

schoen:
@thomaszbz, I think that would be preferable overall.

thomaszbz:
@schoen Could you please revisit this PR? Someone still must test if the change actually does what it is supposed to. It's untested by me so far.

certbot/renewal.py
```diff
@@ -190,6 +191,15 @@ def _restore_required_config_elements(config, renewalparams):
                     raise errors.Error(
                         "Expected a numeric value for {0}".format(config_item))
             setattr(config.namespace, config_item, int_value)
+    # bool-valued items to add if they're present
```
pde: The test coverage of `renewal.py` is a mess for historical reasons (they're mostly in `certbot/tests/main_test.py`, and are more like functional tests than unit tests), but we should probably use each PR landing there as an opportunity to migrate towards consistent unit test coverage. So let's try to add a simple unit test for at least for the new `BOOL_CONFIG_ITEMS` logic, and maybe for all of `_restore_required_config_elements` to `certbot/tests/renewal_test.py` as part of this PR.

@thomaszbz, let us know whether you'd be comfortable tackling that, or might need guidance or help in writing unit tests.

thomaszbz: Sorry, I currently can't even test if it runs without a unit test ;-) Nope, I can't do that without learning a lot of python specific stuff. Which I currently don't want to.

schoen:
@thomaszbz, I'm sorry to say I was on holiday during most of the relevant time so I didn't look at this, and as you might have seen, @bmw ended up duplicating much of this PR (plus some other relevant stuff that we need). We just merged a version of his work on top of this PR and are planning to merge it shortly... sorry for the duplication of effort involved, and thank you for your work on this.

thomaszbz:
@schoen All my commits got merged via @bmw 's PR #3948. That's great to see!

Thanks for integrating this so fast and even improving it a lot!

Issue #3734, opened 2016-11-03
Cli help is sometimes wrong about what the default for something is


pde:
Because `argparse.ArgumentDefaultsHelpFormatter` really doesn't understand the default behaviour of the program, or the semantics of different values.

A few examples:

```
  --disable-hook-validation
                        Ordinarily the commands specified for --pre-hook
                        /--post-hook/--renew-hook will be checked for
                        validity, to see if the programs being run are in the
                        $PATH, so that mistakes can be caught early, even when
                        the hooks aren't being run just yet. The validation is
                        rather simplistic and fails if you use more advanced
                        shell constructs, so you can use this switch to
                        disable it. (default: True)
```
As far as the user is concerned, the real default is `False` (but argparse doesn't understand its own `store_false` semantics.)

```
  --dialog              Run using interactive dialog menus (default: False)
```
This is a flag that shadows the default behaviour of the code (though it's now deprecated).

```
  -d DOMAIN, --domains DOMAIN, --domain DOMAIN
                        Domain names to apply. For multiple domains you can
                        use multiple -d flags or enter a comma separated list
                        of domains as a parameter. (default: [])

  --expand              If an existing cert covers some subset of the
                        requested names, always expand and replace it with the
                        additional names. (default: False)
```
The real default in these cases is "Ask" :)

```
  -c CONFIG_FILE, --config CONFIG_FILE
                        config file path (default: None)
```
The real default here is loaded from `certbot/constants.py`, and is `/etc/letsencrypt/cli.ini`.

pde:
I have a branch that's starting to fix this by creating ridiculous classes that are pythonically `False` or `True`, but which render as the correct string in the argparse output. This is skeezy but seems to work...

pde:
@bitdancer I see you wontfixed [the argparse bug](https://bugs.python.org/issue27153) for the case of `store_false` flags. I think you might have made the wrong decision there, because you're reasoning about the default value of an internal Python variable, whereas a user reading commandline help has no idea about Python variables, and is thinking about the default for whether a *flag* is in effect or not.

But a general question: is there a sane way to stop `argparse` from printing an incorrect default for a few flags, but keep it printing defaults for all the other ones? That would clearly be useful for projects like this one that have dozens or hundreds of flags, most but not all of which are correctly documented by `argparse`.

bitdancer:
On Fri, 04 Nov 2016 13:12:17 -0700, Peter Eckersley <notifications@github.com> wrote:
[...]

PR #3862, opened 2016-12-06
Fix incorrect documentation of some CLI defaults by argparse

pde:
Fixes: #3734 

Involves vendoring a copy of `argparse`, per @erikrose's suggestion...

PR #3863, opened 2016-12-06
Document defaults

pde:
Re-try #3862, which experienced a bunch of github outage weirdness.

Fixes: #3734

~~Involves vendoring a copy of argparse, per @erikrose's suggestion...~~


pde:
About to reboot this PR again, since I noticed I could do it without vendoring `argparse` at all.

pde:
@ohemorange I am not initially persuaded that your first worry is a real problem, though I might be persuaded if you could provide a realistic example of an arugment help string that could accidentally contain the substring `(default:` without that implying that it documented its own default. Also I point out further that the consequence of this bug would be failure to document a default for one of our arguments, which would not be terrible.

pde:
Regarding your second worry: it's possible that bugs will be fixed in upstream's `ArgumentDefaultsHelpFormatter`, but it's quite short and the code for it has not changed in the six years between python 2.7.0's release and python 3.6 development. If it lands, my proposed patch would be the first such change.

certbot/cli.py
```diff
@@ -898,12 +913,21 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
 def _create_subparsers(helpful):
     helpful.add("config_changes", "--num", type=int,
                 help="How many past revisions you want to be displayed")
+
+    class DummyConfig(object):
```
ohemorange: This takes advantage of `determine_user_agent`'s implementation details. If we wanted to change those, we have to change this here, which is weird. Could this be a new method in `client.py` called `demo_user_agent`, or something to that effect that keeps the details of what goes into a user agent in `client.py`?

docs/cli-help.txt
```diff
@@ -13,27 +13,27 @@ cert. Major SUBCOMMANDS are:
   register             Perform tasks related to registering with the CA
   rollback             Rollback server configuration changes made during install
   config_changes       Show changes made to server config during installation
+  update_symlinks      Update cert symlinks based on renewal config file
   plugins              Display information about installed plugins
+  certificates         Display information about certs configured with Certbot
```
ohemorange: Does this get generated from `cli.py`? Where did `rename` go? Is `docs` the folder that we don't update until release?

docs/cli-help.txt
```diff
@@ -81,7 +81,9 @@ optional arguments:
                         agent strings allow the CA to collect high level
                         statistics about success rates by OS and plugin. If
                         you wish to hide your server OS version from the Let's
-                        Encrypt server, set this to "". (default: None)
+                        Encrypt server, set this to "". (default:
+                        CertbotACMEClient/0.10.0.dev0 (Debian GNU/Linux
```
ohemorange: Is this what we want default to say? Presumably this will be regenerated once we update the version number so it doesn't say `0.10.0.dev0`?

docs/cli-help.txt
pde: Yes; I did an update of that file to show how this PR was changing the output of `certbot --help all`; `rename` is missing because that landed after I did the manual update; we usually update `docs/` in real time but maybe there have been one or two exceptions in the past?

ohemorange: Which is the one that we keep in sync with the current release because the docs pulls it straight from master then?

Issue #2305, opened 2016-01-28
Ways of specifying what to renew


schoen:
# 2304 provides for a new `renew` verb. Its default behavior is probably to renew everything possible. However, we also need to figure out possible ways of being more specific about what to renew, like how to interpret `-d` in this context.

bmw:
We want this, but we think that `letsencrypt renew` should renew all certs by default and are pushing this functionality to a later release.

bmw:
See #3615.

pde:
Will be closed as part of #3615

PR #3880, opened 2016-12-09
Make renew command respect the --cert-name flag

### LINKED ISSUE #2071: Allow removing SAN from multidomain certificate when renewing

ohemorange:
Fixes #2071, #2305. Part of C-MIP (#3615).

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -38,6 +38,9 @@ def renewal_conf_files(config):
     """Return /path/to/*.conf in the renewal conf directory"""
     return glob.glob(os.path.join(config.renewal_configs_dir, "*.conf"))
 
+def renewal_file_for_certname(config, certname):
+    """Return /path/to/certname.conf in the renewal conf directory"""
+    return os.path.join(config.renewal_configs_dir, "{0}.conf".format(certname))
 
 def _reconstitute(config, full_path):
     """Try to instantiate a RenewableCert, updating config with relevant items.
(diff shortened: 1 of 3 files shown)


certbot/renewal.py
```diff
@@ -38,6 +38,9 @@ def renewal_conf_files(config):
     """Return /path/to/*.conf in the renewal conf directory"""
     return glob.glob(os.path.join(config.renewal_configs_dir, "*.conf"))
 
+def renewal_file_for_certname(config, certname):
+    """Return /path/to/certname.conf in the renewal conf directory"""
+    return os.path.join(config.renewal_configs_dir, "{0}.conf".format(certname))
```
pde: This should check whether the file exists and is readable, and return a sensible error if it doesn't / isn't.  Currently, if you try to renew a non-existent name, you get:

```
certbot renew --cert-name nonexistentname

Processing /etc/letsencrypt/renewal/nonexistentname.conf
renewal config file {} is missing a required file reference
Renewal configuration file /etc/letsencrypt/renewal/nonexistentname.conf is broken. Skipping.

No renewals were attempted.

Additionally, the following renewal configuration files were invalid: 
  /etc/letsencrypt/renewal/nonexistentname.conf (parsefail)
0 renew failure(s), 1 parse failure(s)
```

When perhaps a more succinct error would do?

ohemorange: I don't think certbot cares about being readable (I assume you mean in the permissions sense?). I like the checking for the filename existing though, especially because it'll have nice symmetry with the renew all lineages version.

PR #3883, opened 2016-12-09
Implement our fancy new --help output

pde:
The top level help is now:
```

cert. The most common SUBCOMMANDS and flags are:

obtain, install, and renew certificates:
    (default) run   Obtain & install a cert in your current webserver
    certonly        Obtain or renew a cert, but do not install it
    renew           Renew all previously obtained certs that are near expiry
   -d DOMAINS       Comma-separated list of domains to obtain a cert for

  --apache          Use the Apache plugin for authentication & installation
  --standalone      Run a standalone webserver for authentication
  --nginx           Use the Nginx plugin for authentication & installation
  --webroot         Place files in a server's webroot folder for authentication
  --manual          Obtain certs interactively, or using shell script hoooks

   -n               Run non-interactively
  --test-cert       Obtain a test cert from a staging server
  --dry-run         Test "renew" or "certonly" without saving any certs to disk

manage certificates:
    certificates    Display information about certs you have from Certbot
    revoke          Revoke a certificate (supply --cert-path)

manage your account with Let's Encrypt:
    register        Create a Let's Encrypt ACME account
  --agree-tos       Agree to the ACME server's Subscriber Agreement
   -m EMAIL         Email address for important account notifications

More detailed help:

  -h, --help [TOPIC]    print this message, or detailed help on a topic;
                        the available TOPICS are:

   all, automation, commands, paths, security, testing, or any of the
   subcommands or plugins (certonly, renew, install, register, nginx,
   apache, standalone, webroot, script, etc.)
```

There is also a `cerbot -h commands`, which produces:
```

cert. The full list of available SUBCOMMANDS is:

run (default)      Obtain/renew a certificate, and install it
certonly           Obtain or renew a certificate, but do not install it
renew              Renew all certificates (or one specifed with --cert-name)
certificates       List all certificates managed by Certbot
revoke             Revoke a certificate specified with --cert-path
revoke             Change a certificate's name (for management purposes)
register           Register for account with Let's Encrypt / other ACME server
install            Install an arbitrary cert in a server
config_changes     Show changes that Certbot has made to server configurations
rollback           Roll back server conf changes made during cert installation
plugins            List plugins that are installed and available on your system
update_symlinks    Recreate symlinks in your /live/ directory

You can get more help on a specific subcommand with --help SUBCOMMAND
```
It's also possible to customise the help more for any given subcommand; I've done that for `run`, `certonly`, and `renew`.  For instance, `certbot -h renew` now produces:
```
usage: 

  certbot renew [--cert-name NAME] [options]

optional arguments:
                        and ~/.config/letsencrypt/cli.ini) (default: None)

renew:
  The 'renew' subcommand will attempt to renew all certificates (or more
  precisely, certificate lineages) you have previously obtained if they are
  close to expiry, and print a summary of the results. By default, 'renew'
  will reuse the options used to create obtain or most recently successfully
  renew each certificate lineage. You can try it with `--dry-run` first. For
  more fine-grained control, you can renew individual lineages with the
  `certonly` subcommand. Hooks are available to run commands before and
  after renewal; see https://certbot.eff.org/docs/using.html#renewal for
  more information on these.

  --dry-run             Perform a test run of the client, obtaining test
                        (invalid) certs but not saving them to disk. This can
                        currently only be used with the 'certonly' and 'renew'
                        subcommands. Note: Although --dry-run tries to avoid
                        making any persistent changes on a system, it is not
                        completely side-effect free: if used with webserver
                        authenticator plugins like apache and nginx, it makes
                        and then reverts temporary config changes in order to
                        obtain test certs, and reloads webservers to deploy
                        and then roll back those changes. It also calls --pre-
                        hook and --post-hook commands if they are defined
                        because they may be necessary to accurately simulate
                        renewal. --renew-hook commands are not called.
                        (default: False)
  --force-renewal, --renew-by-default
                        If a certificate already exists for the requested
                        domains, renew it now, regardless of whether it is
                        near expiry. (Often --keep-until-expiring is more
                        appropriate). Also implies --expand. (default: False)
  --allow-subset-of-names
                        When performing domain validation, do not consider it
                        a failure if authorizations can not be obtained for a
                        strict subset of the requested domains. This may be
                        useful for allowing renewals for multiple domains to
                        succeed even if some domains no longer point at this
                        system. This option cannot be used with --csr.
                        (default: False)
  -q, --quiet           Silence all output except errors. Useful for
                        automation via cron. Implies --non-interactive.
                        (default: False)
  --preferred-challenges PREF_CHALLS
                        A sorted, comma delimited list of the preferred
                        challenge to use during authorization with the most
                        preferred challenge listed first (Eg, "dns" or "tls-
                        sni-01,http,dns"). Not all plugins support all
                        challenges. See
                        https://certbot.eff.org/docs/using.html#plugins for
                        details. ACME Challenges are versioned, but if you
                        pick "http" rather than "http-01", Certbot will select
                        the latest version automatically. (default: [])
  --pre-hook PRE_HOOK   Command to be run in a shell before obtaining any
                        certificates. Intended primarily for renewal, where it
                        can be used to temporarily shut down a webserver that
                        might conflict with the standalone plugin. This will
                        only be called if a certificate is actually to be
                        obtained/renewed. (default: None)
  --post-hook POST_HOOK
                        Command to be run in a shell after attempting to
                        obtain/renew certificates. Can be used to deploy
                        renewed certificates, or to restart any servers that
                        were stopped by --pre-hook. This is only run if an
                        attempt was made to obtain/renew a certificate.
                        (default: None)
  --renew-hook RENEW_HOOK
                        Command to be run in a shell once for each
                        successfully renewed certificate. For this command,
                        the shell variable $RENEWED_LINEAGE will point to the
                        config live subdirectory containing the new certs and
                        keys; the shell variable $RENEWED_DOMAINS will contain
                        a space-delimited list of renewed cert domains
                        (default: None)
  --disable-hook-validation
                        Ordinarily the commands specified for --pre-hook
                        /--post-hook/--renew-hook will be checked for
                        validity, to see if the programs being run are in the
                        $PATH, so that mistakes can be caught early, even when
                        the hooks aren't being run just yet. The validation is
                        rather simplistic and fails if you use more advanced
                        shell constructs, so you can use this switch to
                        disable it. (default: True)
```


pde:
This includes #3877 only because I was using it while writing this branch, but it's going to be hard to rebase out due to subsequent merges...

certbot/cli.py
```diff
@@ -309,6 +299,82 @@ def add_argument(self, *args, **kwargs):
...
+        "short": "Roll back server conf changes made during cert installation",
+        "opts": "Options for rolling back server configuration changes"
+    }),
+    ("plugins", {
+        "short": "List plugins that are installed and available on your system",
+        "opts": 'Options for for the "plugins" subcommand'
+    }),
+    ("update_symlinks", {
+        "short": "Recreate symlinks in your /live/ directory",
+        "opts": ("Recreates cert and key symlinks in {0}, if you changed them by hand, "
```
ohemorange: nit: these commas do not belong.

pde:
`rename` landed in master halfway through my work on this branch, so adding it now. `manage` should be added once it exists?

ohemorange:
I thought we were just going to have `manage` as a help topic and not as a subcommand for now?

pde:
You're totally right...

certbot/cli.py
```diff
@@ -51,51 +51,54 @@
...
 
-OR use different plugins to obtain (authenticate) the cert and then install it:
+   -n               Run non-interactively
+  --test-cert       Obtain a test cert from a staging server
+  --dry-run         Test "renew" or "certonly" without saving any certs to disk
 
-  --authenticator standalone --installer apache
+manage certificates:
+    certificates    Display information about certs you have from Certbot
+    revoke          Revoke a certificate (supply --cert-path)
```
pde: Added

pde:
Ok I added a `manage` topic:

```
certbot --help manage
usage: 

cert. 

optional arguments:
                        and ~/.config/letsencrypt/cli.ini) (default: None)

manage:
  Various subcommands and flags are available for managing your
  certificates:

  certificates          List all certificates managed by Certbot
  renew                 Renew all certificates (or one specifed with --cert-
                        name)
  revoke                Revoke a certificate specified with --cert-path
  rename                Change a certificate's name (for management purposes)
  --cert-name CERTNAME  Certificate name to apply. Only one certificate name
                        can be used per Certbot run. To see certificate names,
                        run 'certbot certificates'.If there is no existing
                        certificate with this name and domains are requested,
                        create a new certificate with this name. (default:
                        None)
  --updated-cert-name NEW_CERTNAME
                        New name for the certificate. Must be a valid
                        filename. (default: None)
  --cert-path CERT_PATH
                        Path to where cert is saved (with auth --csr),
                        installed from or revoked. (default: None)
```

certbot/cli.py
```diff
@@ -788,15 +795,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              "multiple -d flags or enter a comma separated list of domains "
              "as a parameter.")
     helpful.add(
-        [None, "run", "certonly"],
+        [None, "run", "certonly", "manage"],
         "--cert-name", dest="certname",
         metavar="CERTNAME", default=None,
         help="Certificate name to apply. Only one certificate name can be used "
              "per Certbot run. To see certificate names, run 'certbot certificates'."
```
ohemorange: Missing a space after the period

certbot/cli.py
```diff
@@ -1051,9 +1058,7 @@ def _paths_parser(helpful):
         verb = helpful.help_arg
 
     cph = "Path to where cert is saved (with auth --csr), installed from or revoked."
```
ohemorange: Is that a lack of oxford comma I see? Probably because other parts are hogging all the commas.

PR #3886, opened 2016-12-09
Fix certbox-nginx address equality check

### LINKED ISSUE #3855: Fully check for Nginx address equality
`"0.0.0.0"`, `"*"`, and `""` are equivalent to Nginx but not to the parser. This will manifest as a failure to correctly label server blocks as `ssl`, or anywhere else address equality is tested. This should probably be fixed in the Nginx `Addr` object when testing equality.

ClifHouck:
0.0.0.0, *, and '' are equivalent hosts to nginx. Changes Addr object's
equality testing to treat them as equal.

Fixes #3855

### DIFF
--- a/certbot-nginx/certbot_nginx/obj.py
+++ b/certbot-nginx/certbot_nginx/obj.py
@@ -29,10 +29,14 @@ class Addr(common.Addr):
     :param bool default: Whether the directive includes 'default_server'
 
     """
+    UNSPECIFIED_IPV4_ADDRESSES = ('', '*', '0.0.0.0')
+    CANONICAL_UNSPECIFIED_ADDRESS = UNSPECIFIED_IPV4_ADDRESSES[0]
+
     def __init__(self, host, port, ssl, default):
         super(Addr, self).__init__((host, port))
         self.ssl = ssl
         self.default = default
+        self.unspecified_address = host in self.UNSPECIFIED_IPV4_ADDRESSES
 
     @classmethod
     def fromstring(cls, str_addr):
@@ -96,6 +100,13 @@ class Addr(common.Addr):
     def super_eq(self, other):
         """Check ip/port equality, with IPv6 support.
         """
+        # If both addresses got an unspecified address, then make sure the
+        # host representation in each match when doing the comparison.
+        if self.unspecified_address and other.unspecified_address:
+            return common.Addr((self.CANONICAL_UNSPECIFIED_ADDRESS,
+                                self.tup[1]), self.ipv6) == \
+                   common.Addr((other.CANONICAL_UNSPECIFIED_ADDRESS,
+                                other.tup[1]), other.ipv6)
         # Nginx plugin currently doesn't support IPv6 but this will
         # future-proof it
         return super(Addr, self).__eq__(other)
--- a/certbot-nginx/certbot_nginx/tests/obj_test.py
+++ b/certbot-nginx/certbot_nginx/tests/obj_test.py
@@ -1,5 +1,6 @@
 """Test the helper objects in certbot_nginx.obj."""
 import unittest
+import itertools
 
 
 class AddrTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


ClifHouck:
@ohemorange and @joohoi apologies if this steps on one or both of your toes, but this looked like something I could tackle.

pde:
Thanks for the PR @ClifHouck! I don't think either of them had started work on this.

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -96,6 +96,11 @@ def __repr__(self):
     def super_eq(self, other):
         """Check ip/port equality, with IPv6 support.
         """
+        # Check if both objects have wildcard equivalent addresses.
+        # If so, we're done.
+        any_addresses = ('0.0.0.0', '*', '')
+        if self.tup[0] in any_addresses and other.tup[0] in any_addresses:
+            return True
```
ohemorange: This isn't quite right - the ports have to match as well, and the purpose of the `super` call below is that it handles ipv6 addresses. Handling ipv6 addresses might be a nice-to-have rather than necessary right now, but at minimum this method should return false for testing, say, `0.0.0.0:80` against `*:443`.

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -96,6 +96,11 @@ def __repr__(self):
     def super_eq(self, other):
```
ohemorange: Previously, this method just called the equal method of this class' parent, which is why it was glossed `super_eq`. If the address check is going here, could you rename it to "raw address eq" or something of the sort?

certbot-nginx/certbot_nginx/obj.py
ClifHouck: Oh true, duh. I'll fix the port checking at least and add a unit test. Maybe add v6 checking in a follow-up?

ohemorange: That's one way. Another way would be to decide on a canonical internal representation of these three addresses, set that at init based on which we have, and test equality using the internal representation. Then we don't have to duplicate the port checking in `super`.

ClifHouck:
@ohemorange Thanks for the review, I think I've addressed your comments. Please let me know what you think.

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -96,6 +100,13 @@ def __repr__(self):
...
         """Check ip/port equality, with IPv6 support.
         """
+        # If both addresses got an unspecified address, then make sure the
+        # host representation in each match when doing the comparison.
+        if self.unspecified_address and other.unspecified_address:
+            return common.Addr((self.CANONICAL_UNSPECIFIED_ADDRESS,
+                                self.tup[1]), self.ipv6).__eq__(
```
ohemorange: You can use `==` here since you're calling the class' built-in `__eq__` method.

ohemorange: You can use `==` here since you're calling the class' built-in `__eq__` method.

certbot-nginx/certbot_nginx/tests/obj_test.py
```diff
@@ -72,6 +73,18 @@ def test_eq(self):
...
+    def test_equivalent_any_addresses(self):
+        from certbot_nginx.obj import Addr
+        any_addresses = ("0.0.0.0:80 default_server ssl",
+                         "80 default_server ssl",
+                         "*:80 default_server ssl")
+        for first, second in itertools.combinations(any_addresses, 2):
+            self.assertEqual(Addr.fromstring(first), Addr.fromstring(second))
+
+        # Also, make sure ports are checked.
+        self.assertNotEqual(Addr.fromstring(any_addresses[0]),
```
ohemorange: Could you check that `any_addresses[1]` isn't equal to some non-unspecified address listening on the same port?

ClifHouck: Yes, added a check for all three any_addresses

certbot-nginx/certbot_nginx/obj.py
ClifHouck: Sure, I'll do that.

ClifHouck:
Again, I believe this addresses your review @ohemorange

ClifHouck:
By the way, is there an easy way to directly run the certbot_nginx unit tests without running all the other test suites? I've just been using `tox`.

PR #3889, opened 2016-12-09
Refactor exception handler

pde:
* Take the crazy logic in `main._handle_exception` and unfurl it for clarity
* Should preserve behaviour, except we don't bother writing `certbot.log` if the user provided `--debug` and the code crashes before a `config` exists.
* Also parallelise the refactored `main_test.py`, for good measure.


certbot/main.py
```diff
@@ -705,42 +705,40 @@ def _handle_exception(exc_type, exc_value, trace, config):
...
+    elif (config is not None and config.debug) or (config is None and "--debug" in sys.argv):
         sys.exit(tb_str)
+    elif issubclass(exc_type, errors.Error):
+        sys.exit(exc_value)
+
+    if config is None:
+        logfile = "certbot.log"
+        try:
+            with open(logfile, "w") as logfd:
+                traceback.print_exception(
```
bmw: We could just do something like the following here since we already have a formatted exception:
```
logfd.write(tb_str + "\n")
```
I think this is a little easier to read, but I don't really care if you make this change.

certbot/main.py
```diff
@@ -705,42 +705,40 @@ def _handle_exception(exc_type, exc_value, trace, config):
...
-            else:
-                msg += "logfiles in {0} for more details.".format(config.logs_dir)
-            sys.exit(msg)
-    else:
+    if not issubclass(exc_type, Exception):
+        sys.exit(tb_str)
+    elif (config is not None and config.debug) or (config is None and "--debug" in sys.argv):
         sys.exit(tb_str)
+    elif issubclass(exc_type, errors.Error):
+        sys.exit(exc_value)
```
pde: Good question! This was the logic we had previously (since  13913fd8e!), so I just preserved it. It could be wrong, but I've found this case (a `certbot.errors.Error` before arguments are parsed) to only happen when I make mistakes during development.  And then I just add `--debug` to see what really happened. We could take the opposite gambit and always traceback, at the risk of ugly errors for some users sometimes?

bmw:
This will be obsoleted by my upcoming PR.

Issue #3767, opened 2016-11-08
Further merge --script-* with --*-hook


pde:
Today @jsha made an argument that we should work a bit harder to make the script plugin feel and operate consistently with the existing hooks.  We reached some conclusions about this:

* `--script-auth` and `--script-cleanup` could become `--auth-hook` and `--cleanup-hook`; that would require moving them out of a plugin but that's doable.
* Support from them could be moved from their own plugin to the manual plugin, with the advantage that existing manual users could told about them interactively and onboarded into the brave new scripting future
* We disagreed on whether `--auth-hook` should always be run or whether it should be specific to `--manual`, but concluded that this would be the prerogative of whoever implements this refactor.

bmw:
I was going to take this unless someone else wanted to.

joohoi:
It's a hard question. There certainly are benefits to the both approaches, however I would suggest leaving it as is - to script plugin. This is not up to me, but here are my thoughts about the subject to help you to pick the way to proceed.

Things that would promote merging the functionality with hooks:
 - Code simplicity
 - Being able to possibly use auth and cleanup hooks from inside other plugins also if needed

However things that I see beneficial for having the functionality inside a plugin, be it merged with manual plugin, or left to script plugin:
 - Keeping the authenticator / installer functionality inside plugin, and staying consistent with current design. I think this is also intuitive from the user perspective, as the current hooks are pretty much "one shot" scripts, that are run outside of the actual auth loop, ie: outside of the "magic".
 - In contrary to hooks, plugins are for the actual "heavy lifting" process, while hooks are more like helpers in my view. This is effectively what the authenticator part is doing.

Things promoting keeping it in script plugin instead of merging with manual:
 - The current manual plugin users have either scripted around it or have some kind of process around it. I don't think there are actual conversion factors.
 - Omitting the current manual plugin pause-to-wait-for-challenge-deployment functionality would break scripts for users that have scripted around the manual plugin altogether.

bmw:
I wanted to comment on some of the things that have been said here.

> We disagreed on whether --auth-hook should always be run or whether it should be specific to --manual, but concluded that this would be the prerogative of whoever implements this refactor.

[...]

Issue #3650, opened 2016-10-17
Remove optional dependencies


bmw:
I think we should consider this for our next release as it will dramatically simplify our code. If we remove `already_listening` as #3623 suggests, our optional dependencies have little to no benefit.

The biggest change this would cause is the `simple_verify` method in ACME for the DNS challenge either needs to:
1. Raise `NotImplementedError`
2. Always return `True`.

If any third party plugins/clients that use ACME use this method, we should probably just return `True`. Otherwise, we should probably do the former and consider removing the `simple_verify` method from all challenge objects.

bmw:
Fixed in #4088.

Issue #1777, opened 2015-12-05
Fix UI flow for the manual plugin


pde:
Currently it confuses people greatly with its mix of iDisplay and non-iDisplay output, and it doesn't do a good job of explaining the narrative flow it's creating ("in another window, open a shell to your server and do X").

http://mojzis.com/software/letsencryptshell/

might be one source of ideas for doing a better job.

pde:
One review of the manual plugin from 1775 was:

> Fanky cli experience, step-by-step:
> 1) Manual mode presents a direction of what should be done to server on a domain to prove, and asks  to press return when done
> 2) Upon pressing a return after some 20 minutes (in which time server is properly configured), a screen  with color-in-cli appears showing something about error in validation, as a single line header.
> 3) A few seconds later, new line appear in cli (non-gui) style, saying that everything is OK, and pointing to places with certs.
> Given that step 3 shows success, warning display at step 2 just "puts hair on fire"

bmw:
With the removal of `dialog` and the manual plugin now always using `IDisplay`, this can be closed.

PR #3890, opened 2016-12-10
Merge the manual and script plugins

### LINKED ISSUE #2505: Stop using simple_verify in manual plugin

bmw:
Fixes #2505, #3767 

This merges the manual and script plugins into one. While this was largely just refactoring existing code, I changed a few things about how the plugins used to work. I wanted to bring these things to the reviewer's attention to explain myself and to make sure they saw them.

#### --manual-test-mode

I have removed this flag entirely. It existed before to give us a way to test the manual plugin in integration tests. Now that the manual plugin accepts hooks, this is unnecessary. I did some searching to make sure that no one was using this flag and I couldn't find any references is to it other than output from `certbot --help` simply showing it exists. Because of this and the fact the flag was explicitly listed for testing, I think it is safe to simply remove this flag.

#### Stop using `simple_verify`

This fixes #2505. If you're not familiar, the `simple_verify` methods are client side checks that the challenge is set up correctly. I think doing this is bad for a number of reasons:

1. Code complexity for little/no benefit. Why does it matter if the client or the server tells you there is a problem? We don't do anything extra for the user if we detect a problem.
2. It's unreliable. Due to things like firewall rules, we've had `simple_verify` tell the user the challenge was set up incorrectly, yet it passes when Let's Encrypt checks. Here's [one example](https://github.com/certbot/certbot/issues/1777#issuecomment-162240317) of this.
3. For the DNS challenge, it's progress on #3650 because we're no longer using `dnspython` in Certbot.

#### Challenge selection for the script plugin

The plugin doesn't actually have to do anything here. The relevant code for this is [here](https://github.com/certbot/certbot/blob/master/certbot/auth_handler.py#L255). The challenges are preferred in the order given by the user with any challenges not supported by the selected plugin removed.


bmw:
Unfortunately I didn't get a chance to track down why the new integration tests are failing. I'll take a look next week.

bmw:
I couldn't get the DNS auth hook to work during integration tests. I created an issue to track it in #3901.

I don't believe that should be a blocker for this PR though. I've successfully tested the plugin with DNS challenges manually and with hooks.

certbot/plugins/manual.py
```diff
@@ -59,203 +52,113 @@ class Authenticator(common.Plugin):
...
+        add('auth-hook',
+            help='Path or command to execute for the authentication script')
+        add('cleanup-hook',
+            help='Path or command to execute for the cleanup script')
+        add('public-ip-logging-ok', action='store_true',
+            help='Automatically allows public IP logging')
+
+    def prepare(self):  # pylint: disable=missing-docstring
+        if self.config.noninteractive_mode and not self.conf('auth-hook'):
+            raise errors.PluginError(
```
pde: Maybe use `errors.MissingCommandLineFlag`?  Or even make `NoninteractiveDisplay._interaction_fail` a public method, and call it?

certbot/plugins/manual.py
```diff
@@ -59,203 +52,113 @@ class Authenticator(common.Plugin):
...
     def add_parser_arguments(cls, add):
-        add("test-mode", action="store_true",
-            help="Test mode. Executes the manual command in subprocess.")
-        add("public-ip-logging-ok", action="store_true",
-            help="Automatically allows public IP logging.")
-
-    def prepare(self):  # pylint: disable=missing-docstring,no-self-use
-        if self.config.noninteractive_mode and not self.conf("test-mode"):
-            raise errors.PluginError("Running manual mode non-interactively is not supported")
+        add('auth-hook',
```
bmw: According to your comment [here](https://github.com/certbot/certbot/issues/3767#issue-187879645):
> We disagreed on whether --auth-hook should always be run or whether it should be specific to --manual, but concluded that this would be the prerogative of whoever implements this refactor.

Both Joona and I argued against doing this in #3767.

certbot/plugins/manual.py
bmw: `prepare` has a very specific interface as defined in `interfaces.py` and code like [this](https://github.com/certbot/certbot/blob/master/certbot/plugins/disco.py#L114) only handles those specific error cases. I'd be open to raising `errors.MisconfigurationError` if you think it's more fitting though.

pde: Would we break things by adding new exception types to the interface?

certbot/plugins/manual.py
pde: OK I do see meeting minutes noting that it was up to whoever implemented this to choose between `--auth-hook` and `--manual-auth-hook` :)

certbot/plugins/manual.py
bmw: No, but we'd need to document them in `interfaces.py` and add code to `plugins/disco.py` to handle the new cases. It is my understanding that the code was originally written this way to differentiate between cases where the plugin was simply unavailable and those where there was a real problem and Certbot should crash and notify the user.

Do you think making these changes is valuable? Should we be removing the feature of differentiating between exception types? Do you think these changes should be made in this PR?

My personal opinion is the message raised here is nicer than the one raised out of `_interaction_fail` because it's customized to this specific problem. I also care very little about the exception type, however, it's not immediately clear to me we should be ripping out the feature I described above just to use a different error type.

certbot/plugins/manual.py
```diff
@@ -59,203 +52,113 @@ class Authenticator(common.Plugin):
...
-            time.sleep(1)
-            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            try:
-                sock.connect(("localhost", port))
-            except socket.error:  # pragma: no cover
-                pass
+    def _verify_ip_logging_ok(self):
+        if not self.conf('public-ip-logging-ok'):
+            cli_flag = '--{0}'.format(self.option_name('public-ip-logging-ok'))
+            msg = ('NOTE: The IP of this machine will be publicly logged as '
```
bmw: I did this because I prefer single quotes for strings, however, the lines using double quotes include single quote characters. Rather than escaping them, I just used double quotes which is consistent with [our](https://certbot.eff.org/docs/contributing.html#coding-style) [style](https://www.python.org/dev/peps/pep-0008/#string-quotes) [guides](https://google.github.io/styleguide/pyguide.html?showone=Strings#Strings).

bmw:
This is ready for another review. I've fixed or replied to all review comments.

In fixing the incorrect error message for hooks, I was able to delete a lot of code. The one additional behavioral change I made here to simplify the implementation is we no longer include the CLI flag with the invalid command in the raised error message. I don't think this really hurts usability but if the reviewer does, I'll either add code to catch and reraise the error in `validate_hooks` or write an [argparse.Action](https://docs.python.org/2.7/library/argparse.html#argparse.Action) class in `util` that all hooks can use so we can stop passing around hook prefixes to be included in potential error messages (the `argparse.Action` object already has access to the CLI flag that was used).

certbot/plugins/manual.py
```diff
@@ -59,203 +52,111 @@ class Authenticator(common.Plugin):
...
-            help="Test mode. Executes the manual command in subprocess.")
-        add("public-ip-logging-ok", action="store_true",
-            help="Automatically allows public IP logging. (default: Ask)")
-
-    def prepare(self):  # pylint: disable=missing-docstring,no-self-use
-        if self.config.noninteractive_mode and not self.conf("test-mode"):
-            raise errors.PluginError("Running manual mode non-interactively is not supported")
+        add('auth-hook',
+            help='Path or command to execute for the authentication script')
+        add('cleanup-hook',
```
pde: If this is landing after #3918, please add some checks for these hook to [the new hook integration tests](https://github.com/certbot/certbot/pull/3918/files#diff-c74d3ad4d86d0804f64e9354fa460b94R55).

bmw: Ugh. I just realized my hook error message improvements and refactoring conflicts pretty badly with that PR.

bmw: Also, do you think this is necessary? I already wrote [integration tests for the manual hooks](https://github.com/certbot/certbot/pull/3890/files#diff-c74d3ad4d86d0804f64e9354fa460b94R47) that are actually run to perform the challenge.

bmw:
I went ahead and removed the hook improvement that I implemented. I think making these changes is outside of the scope of this PR and I don't want the merge conflicts it creates with #3918 to prevent this from being merged when it is one of the last large PRs for 0.10.0. I created #3939 to track hook error messages as a separate issue.

I also added another commit to make the name of the hook used in the error messages thrown in `validate_hooks` consistent with other hooks.

joohoi:
LGTM!

certbot/plugins/manual.py
joohoi: This is ok, though not my personal preference. Have to appreciate underlining the legibility with three different sources ;)

PR #3892, opened 2016-12-10
Sort the names by domain (then subdomain) before showing them

Frankkkkk:
Hi,

An easy commit. I think it's quite nice when you have multiple (**a.b**).domain.tld to sort the list first by domain.tld and then, if many, by **b**.subdomain.tld, then by **a**.b.subdomain.tld, etc..

Makes it easier to read.

Cheers

### DIFF
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -142,9 +142,12 @@ def _filter_names(names):
     :rtype: tuple
 
     """
+    #Sort by domain first, and then by subdomain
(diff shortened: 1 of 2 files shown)


Frankkkkk:
Should I also test `_filter_names` or is it enough to only test `_sort_names` ?

Thanks !

pde:
Those tests are pretty elaborate! I would probably have just called the function with a couple of literal examples and checked that it behaved as expected :).  In any case, I think this is okay to merge.

Issue #3242, opened 2016-07-06
Allow specifying the revocation reason


rolandshoemaker:
The ACME spec has added support for [specifying the reason for revoking a certificate](https://ietf-wg-acme.github.io/acme/#certificate-revocation) (one of the `reasonCodes` defined in [RFC5280](https://tools.ietf.org/html/rfc5280#section-5.3.1)). Since the Boulder fix for this is extremely simple wanted to let you know this may appear pretty soon.

bmw:
I moved this into our next milestone because I wanted to make sure that this change in Boulder wouldn't cause any problems for clients that didn't support this feature. After double checking the spec, this shouldn't be a problem, so I'm bumping this to a later milestone.

dashaxiong:
Hi. The commit above (2efa1d) addresses this issue. I have opted to omit the reason field from the revocation request, if the user does not specify a reason. I have a feeling it would be better simply to send 'unspecified' in that case, do you agree? Also, when running `certbot revoke` with the new `reason` option against a local `boulder` instance I find that the audit log doesn't reflect the reason. Would appreciate a second set of eyes to look over this to confirm that the `JWS` payload is formed correctly.

pde:
@dashaxiong yes I think it's fine to say `unspecifed` if the user didn't provide a reason.

I guess we could also interactively ask for the reason (grab the `IDisplay` interface from zope and call `.menu()` on it with `default="unspecified"`.)

dashaxiong:
@pde I have made the default reason to be `unspecified`. The acme protocol and boulder server don't require any reason to be sent, so I initially opted for leaving it out in that case to minimise changes to existing tests that didn't give a reason for revocation. This was possible because of the `omitempty` flag in the `jose.Field`. But it seems cleaner to always send a reason, even if it is `unspecified`. 

Given that prompting for a reason code may cause a workflow regression, it seems wise to wait for #3878 to be resolved prior to adding the interactivity as a separate issue. Also, as the reason code is optional, is it desirable to prompt for it if it is missing?

pde:
@dashaxiong I think that's fine although we just decided to include #3878 in the 0.10.0 milestone to protect against other possible workflow regressions that will be in that release.

dashaxiong:
@rolandshoemaker While doing integration testing against a local Boulder instance the audit log doesn't reflect the reason code sent through. Here's the relevant output of `certbot/tests/boulder-integration.sh`:

```
[...]

PR #3903, opened 2016-12-13
Allow specify revocation reason

dashaxiong:
This adds a feature to specify the reason for certificate revocation (#3242).

The reasons are the same as the user allowed reasons in `letsencrypt.boulder.revocation.reasons.go` which are a subset of the reasons specified in the ACME protocol according to RFC 5280 section 5.3.1. The only difference is that the command line input of the reasons is all lower case. This can easily be changed without consequence if a different style is preferred, as the reason is encoded as an integer code by the `cli` parser.


pde:
@rolandshoemaker interested in reviewing this?

pde:
Putting this in 1.0.0 tentatively because that's where the ticket was but I suspect this will land much sooner.

dashaxiong:
FYI the first two commits (b083db0, 2efa1db) are also by me, but I wrote my address as 'craig at plannd.com.au', which means that github doesn't recognise it.

rolandshoemaker:
Generally looks good to me, you may want to do a downcase on the reason arg before checking in `REVOCATION_REASONS` though so people can use either the all lowercase or IETF style reason names (i.e. `keyCompromise`).

dashaxiong:
@rolandshoemaker it will now ignore the case of reason arguments. It's a little bit hacky. Might be better to decide on just using IETF style. 

The trouble is, it ought to have been straightforward to use the `type` argument of `argparse.add_argument` to convert the input string to lower case before the `choices` list is checked (that is the document behaviour of `argparse`). For some reason this didn't transform the input string before checking against the `choices` list, so I have made a `CaseInsensitiveList` to pass to `choices`.

dashaxiong:
@rolandshoemaker I looked into the command line argument handling. It turns out that `configargparse.ArgParser` is called a first time, in which the `type_func` is applied to the input prior to being compared to the `choices` array. It is also called a second time, by `cli.set_by_cli` and the `reconstructed_args` are passed to it, but on the second run the `type_func` is not applied. I have commented the CaseInsensitiveList to explain briefly why it is necessary for use with `choices` parameter. Ready for review.

dashaxiong:
I'm closing this pull request as I am not happy with the commit history, and I am uncertain what effect `git filter-branch` will have given that I have merged in from master several times. I have opened another PR (#3988) with a cleaner commit history.

PR #3913, opened 2016-12-14
Implement delete command

### LINKED ISSUE #2551: adding -delete option to remove the cert files
Hello,

Do you think you can add an -delete option to let's encrypt?

My context is this: 
From a hosting control panel, users will be able to create an owncloud instance.
When deleting their owncloud instance, I should delete also the certificates files in /etc/letsencrypt folder (I'm not talking about revocation as I can let the certs expire).

Deleting them seems important as by not deleting myself the files in /etc/letsencrypt/{archives,live,renewal}, if I understand correctly, the renewal of the certs will renew all the certs, included the certs for (sub)domains that don't exist anymore.

I must say I'm not sure how the renewal system works, if it tests the domain validity or the host file presence before renewing it. But still, you can have a host file for a domain name that doesn't exist anymore (not very good, but it can happen) and you can have a catchall for subdomains that show a generic page or use the (sub)domain for another purpose.

The only post I found on that topic (https://community.letsencrypt.org/t/correct-way-to-completely-remove-issued-certificate-s-for-a-domain/7409/4) doesn't have an answer on this problem.

If I manage the deletion by myself, deleting all the files /etc/letsencrypt/ subdirs, what about if something changes in the way let's encrypt manage the files (names / placements / ..).
The script will be broken.

Maybe there is another way:
If, by using the revoke option, the renewal feature won't renew the revoked certs, I can live with that.
But in that case, if later, a user want to re-use the same (sub)domain, is it possible to reactivate the certs? (didn't see anything about that in the doc. The renew feature doesn't say anything about this case)

So, is it possible to add such a feature?

And by the way, let's encrypt is a wonderful project. Thanks to you all.

ohemorange:
Fixes #2551. Deletes a certificate. Does not check for revocation. Part of #3615 (C-MIP).

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -233,11 +233,9 @@ def delete_files(config, certname):
         raise errors.CertStorageError(
             "error parsing {0}".format(renewal_filename))
     finally:
-        # we couldn't read it, but let's at least try to delete it
-        try:
-            os.remove(renewal_filename)
-        except OSError:
-            pass
+        # we couldn't read it, but let's at least delete it
+        # if this was going to fail, it already would have.
+        os.remove(renewal_filename)
 
     # cert files and (hopefully) live directory
     # it's not guaranteed that the files are in our default storage
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -20,6 +20,7 @@ from certbot import util
 logger = logging.getLogger(__name__)
 
 ALL_FOUR = ("cert", "privkey", "chain", "fullchain")
+README = "README"
 CURRENT_VERSION = util.get_strict_version(certbot.__version__)
 
 
@@ -102,10 +103,8 @@ def rename_renewal_config(prev_name, new_name, cli_config):
     :param .RenewerConfiguration cli_config: parsed command line
         arguments
     """
-    prev_filename = os.path.join(
-        cli_config.renewal_configs_dir, prev_name) + ".conf"
-    new_filename = os.path.join(
-        cli_config.renewal_configs_dir, new_name) + ".conf"
+    prev_filename = renewal_filename_for_lineagename(cli_config, prev_name)
+    new_filename = renewal_filename_for_lineagename(cli_config, new_name)
     if os.path.exists(new_filename):
         raise errors.ConfigurationError("The new certificate name "
             "is already in use.")
@@ -129,8 +128,7 @@ def update_configuration(lineagename, archive_dir, target, cli_config):
     :rtype: configobj.ConfigObj
 
     """
-    config_filename = os.path.join(
-        cli_config.renewal_configs_dir, lineagename) + ".conf"
+    config_filename = renewal_filename_for_lineagename(cli_config, lineagename)
     temp_filename = config_filename + ".new"
 
     # If an existing tempfile exists, delete it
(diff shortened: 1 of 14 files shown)


certbot/storage.py
```diff
@@ -198,6 +209,70 @@ def lineagename_for_filename(config_filename):
...
+def delete_files(config, certname):
+    """Delete all files related to the certificate.
+
+    If some files are not found, ignore them and continue.
+    """
+    renewal_filename = renewal_file_for_certname(config, certname)
+    try:
+        renewal_config = configobj.ConfigObj(renewal_filename)
+    except configobj.ConfigObjError:
+        raise errors.CertStorageError(
```
pde: If the renewal file is corrupt, we'd have the option of assuming the user has not renamed the lineage and also renamed another lineage to the same name,  and trying to delete live and archive directories at the names we expect them to have.  That would be the right thing to do 99% of the time.

Alternatively, we could print a log message suggesting that the user might want to consider deleting those directories by hand.

certbot/storage.py
```diff
@@ -198,6 +209,70 @@ def lineagename_for_filename(config_filename):
...
+            "error parsing {0}".format(renewal_filename))
+    finally:
+        # we couldn't read it, but let's at least delete it
+        # if this was going to fail, it already would have.
+        os.remove(renewal_filename)
```
pde: I think we should log that we're doing this (and the other deletion steps below).  I don't know what the correct verbosity is, but certainly this should be in `/var/log/letsencrypt`, and probably it should be echoed to the terminal by default.

certbot/storage.py
```diff
@@ -20,9 +22,21 @@
...
 
+def renewal_conf_files(config):
+    """Return /path/to/*.conf in the renewal conf directory"""
+    return glob.glob(os.path.join(config.renewal_configs_dir, "*.conf"))
+
+def renewal_file_for_certname(config, certname):
+    """Return /path/to/certname.conf in the renewal conf directory"""
+    path = os.path.join(config.renewal_configs_dir, "{0}.conf".format(certname))
+    if not os.path.exists(path):
+        raise errors.CertStorageError("No certificate found with name {0}.".format(certname))
```
pde: Possibly append `({1} is missing)` with `path` as `{1}`?

ohemorange: I like appending the path, I don't want people to have to know about the manual stuff. If the renewal file isn't there, we won't touch it, which is acceptable.

ohemorange: Also, this function is used in other places.

certbot/storage.py
ohemorange: Oh good point!

certbot/storage.py
ohemorange: That might be the right thing to do 99% of the time, but it will severely mess up people's configs the other 1%.

To rename another lineage to the same name, they would have had to do it manually rather than through certbot rename. Which means they're already messaging around with files manually, which makes it more likely that this will mess things up.

I'd be good with printing a message saying they may want to delete the default files here.

pde:
Merge if you're happy with this @ohemorange !

ohemorange:
travis

PR #3916, opened 2016-12-15
Mitigate problems for people who run without -n

### LINKED ISSUE #3906: Use isatty checks before asking new questions

bmw:
Fixes #3906

### DIFF
--- a/certbot/constants.py
+++ b/certbot/constants.py
@@ -95,3 +95,6 @@ RENEWAL_CONFIGS_DIR = "renewal"
 
 RENEWER_CONFIG_FILENAME = "renewer.conf"
 """Renewer config file name (relative to `IConfig.config_dir`)."""
+
+FORCE_INTERACTIVE_FLAG = "--force-interactive"
+"""Flag to disable TTY checking in IDisplay."""
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -51,9 +51,10 @@ def _wrap_lines(msg):
 class FileDisplay(object):
     """File-based display."""
 
-    def __init__(self, outfile):
+    def __init__(self, outfile, force_interactive):
         super(FileDisplay, self).__init__()
         self.outfile = outfile
+        self.force_interactive = force_interactive
 
     def notification(self, message, pause=True, wrap=True):
         # pylint: disable=unused-argument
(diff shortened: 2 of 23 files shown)


certbot/constants.py
```diff
@@ -95,3 +95,6 @@
 
 RENEWER_CONFIG_FILENAME = "renewer.conf"
 """Renewer config file name (relative to `IConfig.config_dir`)."""
+
+FORCE_INTERACTIVE_FLAG = "--force-interactive"
```
pde: This pattern is okay, but I'd note that we haven't used it in other situations where there are string literals around the code that contain things like this (`git grep cli_flag`)

certbot/main.py
```diff
@@ -275,7 +277,8 @@ def _ask_user_to_confirm_new_names(config, new_domains, certname, old_domains):
                new_domains,
                old_domains))
     obj = zope.component.getUtility(interfaces.IDisplay)
-    if not obj.yesno(msg, "Update cert", "Cancel", default=True):
+    if not obj.yesno(msg, "Update cert", "Cancel",
```
pde: This is landing in 0.10.0; I'm not sure whether it should have `force_interactive=True`...

certbot/display/util.py
```diff
@@ -197,18 +231,59 @@ def checklist(self, message, tags, default_status=True, **unused_kwargs):
...
+            IDisplay call
+
+        :returns: True if the display can interact with the user
+        :rtype: bool
+
+        """
+        if self.force_interactive or force_interactive:
+            return True
+        else:
+            return sys.stdin.isatty() and self.outfile.isatty()
```
pde: When this happens, should we debug log that we're suppressing an interactive question because there's no TTY?

certbot/constants.py
bmw: Yeah not using a literal here is probably fine because we shouldn't be changing CLI flags on people, but this only adds one line of code and adds a bit of extra flexibility. I'd prefer to leave this as is, but I can change it if you really want it to be consistent with the rest of the code.

certbot/main.py
bmw: You're right! I thought I excluded all of the new prompts in previously existing code paths but looks like I added it to this one by mistake. Thanks for catching it.

certbot/display/util.py
bmw: Definitely. In addition, how about the first time we suppress a question, we log a warning saying what we're doing, that the user probably wants to be using `-n`, and if they don't they can use `--force-interactive`?

bmw:
This should be ready for another review.

PR #3917, opened 2016-12-15
Don't add ServerAlias directives when the domain is already covered by a wildcard

### LINKED ISSUE #2558: Don't add extra ServerName/ServerAlias directives

bmw:
Fixes #2558. Built on #3498.


certbot-apache/certbot_apache/configurator.py
```diff
@@ -1012,13 +1012,34 @@ def _add_servername_alias(self, target_name, vhost):
...
+        :returns: True if there is a wildcard covering target_name in
+            the vhost in vhost_path, otherwise, False
+        :rtype: bool
+
+        """
+        matches = self.parser.find_dir(
+            "ServerAlias", start=vh_path, exclude=False)
+        for match in matches:
+            alias = self.aug.get(match)
+            if alias.startswith("*.") and target_name.endswith(alias[1:]):
```
joohoi: The asterisk wildcard can exist in any position of the string, so I think we'll need to broaden the search.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1012,13 +1012,34 @@ def _add_servername_alias(self, target_name, vhost):
...
+        if self._has_matching_wildcard(vh_path, target_name):
+            return
         if not self.parser.find_dir("ServerName", None,
                                     start=vh_path, exclude=False):
             self.parser.add_dir(vh_path, "ServerName", target_name)
         else:
             self.parser.add_dir(vh_path, "ServerAlias", target_name)
         self._add_servernames(vhost)
 
+    def _has_matching_wildcard(self, vh_path, target_name):
```
joohoi: Apache2 supports both '*' and '?' wildcards, to claim that we support them also, we'll need to match '?' wildcards too.

bmw:
This is ready for another review.

bmw:
I actually didn't realize we already had a method for detecting wildcards when I started working on this PR. I consolidated the two methods and added some more tests. I also removed a broken test and a bug where we were considering `*.example.net` a wildcard match for `example.net`.

This is ready for another review.

bmw:
New tests LGTM. Thanks for the review!

PR #3918, opened 2016-12-16
Save --pre and --post hooks in renewal conf files, and run them in a sophisticated way

pde:
This runs pre hooks as the lineages they're associated with are renewed (but won't run them again if they are literally the same string), and then will save up all post hooks for renewed lineages and run them at the end.

Closes: #3394


certbot/tests/util_test.py
```diff
@@ -45,34 +45,19 @@ def test_failure(self, mock_popen):
...
-        self.assertTrue(self._call("exe"))
-
-    @mock.patch("certbot.util.os.path.isfile")
-    @mock.patch("certbot.util.os.access")
-    def test_not_found(self, mock_access, mock_isfile):
-        mock_access.return_value = False
-        mock_isfile.return_value = True
-        self.assertFalse(self._call("exe"))
+class WhichTest(unittest.TestCase):
+    @mock.patch('certbot.util._is_exe')
```
bmw: nit: It's nice to be consistent for strings and stick to either single or double quotes. This is specified by both of our [style](https://www.python.org/dev/peps/pep-0008/#string-quotes) [guides](https://google.github.io/styleguide/pyguide.html?showone=Strings#Strings).

certbot/util.py
```diff
@@ -66,28 +66,36 @@ def run_script(params):
...
-                return True
+            exe_file = os.path.join(path, program)
+            if _is_exe(exe_file):
+                return exe_file
+
+    return None
 
-    return False
+# Either name makes sense, and we have a lot of call sites!
+exe_exists = which
```
bmw: What about simply using the existing `exe_exists` in your code? There are two call sites to `which` with one `mock.patch` with 21 calls/mocks for `exe_exists`. Why have two functions with the same name? You also only ever use the result of `which` as a `bool` except when you call `_prog` [here](https://github.com/certbot/certbot/pull/3918/files#diff-fee68073ff44ffe30e9303db0210c070R115).

I think this call should be changed. If `which` fails due to known problems with hook validation, the user will get a message like "Error output from None". If `which` succeeds, you pass it in to `os.path.basename` which I believe will return the same thing whether or not you called `which` (`os.path.basename("/bin/echo") == os.path.basename("echo")`).

tests/boulder-integration.sh
```diff
@@ -37,14 +37,41 @@ common() {
...
+
+CheckHooks() {
+    EXPECTED="/tmp/expected$$"
+    echo "wtf.pre" > "$EXPECTED"
+    echo "wtf2.pre" >> "$EXPECTED"
+    echo "renew" >> "$EXPECTED"
+    echo "renew" >> "$EXPECTED"
+    echo "wtf.post" > "$EXPECTED"
+    echo "wtf2.post" >> "$EXPECTED"
+    if cmp --quiet "$EXPECTED" "$HOOK_TEST" ; then
```
bmw: I'm a bit worried about this test because according to the [glob documentation](https://docs.python.org/2/library/glob.html) which [we use for finding renewal configuration files](https://github.com/certbot/certbot/blob/master/certbot/storage.py#L29), the order returned is arbitrary meaning whether `wtf.pre` or `wtf2.pre` comes first isn't well defined.

certbot/main.py
```diff
@@ -782,6 +782,20 @@ def set_displayer(config):
         displayer = display_util.FileDisplay(sys.stdout)
     zope.component.provideUtility(displayer)
 
+def _post_logging_setup(config, plugins, cli_args):
+    "Perform any setup or configuration tasks that require a logger."
```
pde: I thought our plan was to adopt a strict style guide in combination with an autoformatter so that human developers wouldn't spend their time on them ;)

tests/boulder-integration.sh
pde: Three options:
1. We live with that risk, on the theory that we're probably the only people running these unit tests, and we'll fix thing if a future version of Python changes this
2. We wrap the `glob.glob` call in `sorted()`, causing our renewal order to be deterministic but subtly risking breaking anyone who assumed that our renewal order wouldn't change ;)
3. Remove the integration test, or write it in some crazy order-agnostic way.

certbot/hooks.py
```diff
@@ -36,35 +44,51 @@ def validate_hook(shell_cmd, hook_name):
...
+        _run_hook(cmd)
+
+post_hook.eventually = []
+
+def run_saved_post_hooks():
+    """Run any post hooks that were saved up in the course of the 'renew' verb"""
+    for cmd in post_hook.eventually:
+        logger.info("Running post-hook command: %s", cmd)
+        _run_hook(cmd)
+    if len(post_hook.eventually) == 0:
```
bmw: Posting this again here for tracking purposes because the relevant lines changed on the old comment.

nit: From your favorite document https://www.python.org/dev/peps/pep-0008/#id49
```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes: if not seq:
     if seq:

No: if len(seq):
    if not len(seq):
```

certbot/main.py
bmw: We discussed this as a possibility, however, we still need to come up with that style and make sure there is an autoformatter we're happy with. In the meantime, I personally don't think that means that existing style/conventions go out the window.

tests/boulder-integration.sh
bmw: I was leaning towards the latter half of 3:
> ...write it in some crazy order-agnostic way.

My concern with 1 is there are at least two regular outside contributors who run the boulder integration tests locally. Even for us, if this breaks six months down the line, will we immediately know why?

After looking at the first line of `$HOOK_TEST`, you know what the rest of the file should look like. It's certainly more code, but it's more robust and I don't think it's that difficult to implement.

bmw:
#3890 landed so I'm supposed to add the task mentioned [here](https://github.com/certbot/certbot/pull/3890#discussion_r93125904) to this PR. After discussing it with you, I agree that these tests are valuable, but I personally would be fine not blocking this PR on them and instead adding that as a task for 0.11.0.

bmw:
This is ready for another review. You may want to merge rather than squash here to preserve authorship.

pde:
Brad's changes LGTM.

Issue #3264, opened 2016-07-08
Speed up travis tests (split out integration tests?)


pde:
Travis latency is a pretty major drag on Certbot developers.  Here are some steps could help:
- Get rid of `py27`, since `cover` covers it
- Split integration tests out of the unit test matrix entries, which would allow them to be run in parallel
- Pick some things that we don't expect to fail more than once in blue moon, and only run those during releases and/or on `obs7`.  The most likely candidate would be the `-oldest` matrix elements.

pde:
Addressed by #3836 #3843

pde:
We could still do better wrt integration tests, though

pde:
Was also addressed by getting more travis jobs

pde:
Changing ticket title to reflect the one remaining major task.

pde:
Alternatively, figure out ways to use two-core parallelism to speed up the integration tests. Though that may be tricky because Certbot isn't generally designed to be run multiple times simultaneously.

pde:
We've probably gotten these to be as fast as they're going to be!

PR #3919, opened 2016-12-16
Serialize coverage tests

bmw:
I noticed today that coverage is now skipping files. For example, looking at the coverage run from #3916, `certbot/notify.py` and `certbot/plugins/script.py` aren't reporting coverage at all. While looking into this I saw this quote from https://nose.readthedocs.io/en/latest/doc_tests/test_multiprocess/multiprocess.html:

> Plugin interaction warning

> The multiprocess plugin does not work well with other plugins that expect to wrap or gain control of the test-running process. Examples from nose’s builtin plugins include **coverage** and profiling: a test run using both multiprocess and either of those is likely to fail in some confusing and spectacular way.

While I like the speed increase, I don't think it's worth our tests being buggy when there are documented problems here. This only disables running our coverage tests in parallel. Our normal unit tests and lint are unaffected by this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ matrix:
       after_failure:
         - sudo cat /var/log/mysql/error.log
         - ps aux | grep mysql
+      services: docker
     - sudo: required
       env: TOXENV=apache_compat
       services: docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,7 @@ env:
 matrix:
   include:
     - python: "2.7"
-      env: TOXENV=cover BOULDER_INTEGRATION=1
-      sudo: required
-      after_failure:
-        - sudo cat /var/log/mysql/error.log
-        - ps aux | grep mysql
-      services: docker
+      env: TOXENV=cover
     - python: "2.7"
       env: TOXENV=lint
     - python: "2.7"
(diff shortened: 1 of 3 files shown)


pde:
Maybe we should even go further an split all the integration tests into their own travis containers (finishing off #3264), although I have a feeling that might not speed us up so much because we're close to using our 10 jobs at full capacity throughout each Travis run.

bmw:
Actually, this could have been a fluke but the coverage tests here ran faster than any of the other latest 5 PRs. Running py27 integration tests in their own environment wouldn't be that hard though. If you'd like to wait to merge this PR until we've split py27 integration tests in their own environment, I'll try to send follow up commits next week,

I'm not sure splitting all integration tests into their own environment is a good idea though. Ideally, someone should implement and test it, but due to all of the duplicate work of starting a Travis environment and installing APT and Python packages, I'm not sure at first glance it'd actually be faster.

I think a bigger win would be to start boulder  while the unit tests are running. I believe we used to do that.

bmw:
Ready for another review.

bmw:
@pde, to give you more incentive to merge this, Travis tests here ran 7-30 seconds faster than the top 5 open PRs with passing tests.

pde:
Nice! We should probably this for the other integration tests too, but I'll merge this now.

Issue #3922, opened 2016-12-16
Certbot revoke gets confused by staging certificates


pde:
When the user has made certs with `--test-cert` (aka `--staging`), and tries to revoke them, `certbot revoke` isn't smart enough to figure out where the revocation request should go.

We can probably hack this back together, but maybe we should have put the ACME directory endpoint in the cert after all...

```
certbot --config-dir /tmp/xyz/ certificates
Saving debug log to /var/log/letsencrypt/letsencrypt.log

-------------------------------------------------------------------------------
Found the following certs:
  Certificate Name: isnot.org
    Domains: isnot.org
    Expiry Date: 2017-03-15 23:46:00+00:00 (INVALID: TEST CERT)
    Certificate Path: /tmp/xyz/live/isnot.org/fullchain.pem
    Private Key Path: /tmp/xyz/live/isnot.org/privkey.pem
  Certificate Name: demo.reworld.org
    Domains: demo.reworld.org
    Expiry Date: 2017-03-15 23:47:00+00:00 (INVALID: TEST CERT)
    Certificate Path: /tmp/xyz/live/demo.reworld.org/fullchain.pem
    Private Key Path: /tmp/xyz/live/demo.reworld.org/privkey.pem
  Certificate Name: www.isnot.org
    Domains: www.isnot.org
    Expiry Date: 2017-03-15 23:47:00+00:00 (INVALID: TEST CERT)
    Certificate Path: /tmp/xyz/live/www.isnot.org/fullchain.pem
    Private Key Path: /tmp/xyz/live/www.isnot.org/privkey.pem
-------------------------------------------------------------------------------
ubuntu@isnot:~$ 
ubuntu@isnot:~$ certbot --config-dir /tmp/xyz/ revoke --cert-path /tmp/xyz/live/isnot.org/fullchain.pem
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Enter email address (used for urgent renewal and security notices) (Enter 'c' to
cancel):c
An e-mail address or --register-unsafely-without-email must be provided.
ubuntu@isnot:~$ lel --config-dir /tmp/xyz/ revoke --cert-path /tmp/xyz/live/isnot.org/fullchain.pem --test-cert
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Starting new HTTPS connection (1): acme-staging.api.letsencrypt.org

-------------------------------------------------------------------------------
Congratulations! You have successfully revoked the certificate that was located
[...]

PR #3923, opened 2016-12-16
Document some particularities of the revoke subcommand

pde:
Give revoke better cli help. Also hint at the weirdness of #3922.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -913,7 +913,8 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Silence all output except errors. Useful for automation via cron."
              " Implies --non-interactive.")
     # overwrites server, handled in HelpfulArgumentParser.parse_args()
-    helpful.add(["testing", "revoke"], "--test-cert", "--staging", action='store_true', dest='staging',
+    helpful.add(["testing", "revoke"], "--test-cert", "--staging",
+        action='store_true', dest='staging',
         help='Use the staging server to obtain or revoke test (invalid) certs; equivalent'
              ' to --server ' + constants.STAGING_URI)
     helpful.add(
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -354,7 +354,8 @@ VERB_HELP = [
     }),
     ("revoke", {
         "short": "Revoke a certificate specified with --cert-path",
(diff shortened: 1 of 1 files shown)


certbot/cli.py
```diff
@@ -912,8 +913,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Silence all output except errors. Useful for automation via cron."
              " Implies --non-interactive.")
     # overwrites server, handled in HelpfulArgumentParser.parse_args()
-    helpful.add("testing", "--test-cert", "--staging", action='store_true', dest='staging',
-        help='Use the staging server to obtain test (invalid) certs; equivalent'
+    helpful.add(["testing", "revoke"], "--test-cert", "--staging",
```
ohemorange: If we're adding this to "revoke", should it also be in "certonly", "run", "renew", etc?

pde: It's a matter of judiciously selecting the most relevant options to be documented in those topics; `certonly` already has `--dry-run` which is probably better for most use cases, but I can add `--test-cert` to run.

pde:
The error from `revoke` looks like this:
```
--config-dir /tmp/xyz/ revoke --cert-path /tmp/xyz/live/isnot.org/fullchain.pem --agree-t --regi
Registering without email!
An unexpected error occurred:
The request message was malformed :: No such certificate

IMPORTANT NOTES:
 - Your account credentials have been saved in your Certbot
   configuration directory at /tmp/xyz. You should make a secure
   backup of this folder now. This configuration directory will also
   contain certificates and private keys obtained by Certbot so making
   regular backups of this folder is ideal.
```
Note the creation of a production account in what was previously a staging-only setup.  If you already have a production account it looks like this:

```
certbot --config-dir /tmp/xyz/ revoke --cert-path /tmp/xyz/live/isnot.org/fullchain.pem
An unexpected error occurred:
The request message was malformed :: No such certificate
```
The `No such certificate` error is generated by boulder, so we can't directly change it. For the record I don't think this is a true fix for #3922 (which would really be to examine the `Issuer` of a cert to set a sensible default revocation `server` value).

ohemorange:
Do all these commits belong in this PR?

pde:
I was wondering where those commits had gone :).  Rebased/fixed.

ohemorange:
travis?

PR #3941, opened 2016-12-20
OCSP checking and other cool "certbot certificates" features

### LINKED ISSUE #3907: "certbot certificates" should support filtering of output
For instance,

`certbot certificates -d example.com` should only list certs for that domain (though question: if the user specifies two domains, should it be intersection or union?)

`certbot certificates --cert-name NAME` should only provide information about that specific lineage.

pde:
* Brings in some OCSP code from @jdkasten's ancient `certificate_manager` branch, allowing `certbot certificates` to list revocation statuses. Fixes: #3908 
* Allows the user to filter `certbot certificates` output with `-d` or `--cert-name`. Fixes #3907

### DIFF
--- a/certbot/cert_manager.py
+++ b/certbot/cert_manager.py
@@ -172,12 +172,12 @@ def _report_human_readable(parsed_certs):
     for cert in parsed_certs:
         now = pytz.UTC.fromutc(datetime.datetime.utcnow())
         expiration_text = ""
+        revoked = ocsp.revoked_status(cert.cert, cert.chain)
         if cert.is_test_cert:
             expiration_text = "INVALID: TEST CERT"
         elif cert.target_expiry <= now:
             expiration_text = "INVALID: EXPIRED"
         else:
-            revoked = ocsp.revoked_status(cert.cert, cert.chain)
             if revoked:
                 expiration_text = "INVALID: " + revoked
 
--- a/certbot/cert_manager.py
+++ b/certbot/cert_manager.py
@@ -81,7 +81,7 @@ def certificates(config):
             parse_failures.append(renewal_file)
 
     # Describe all the certs
-    _describe_certs(parsed_certs, parse_failures)
+    _describe_certs(config, parsed_certs, parse_failures)
 
 def delete(config):
     """Delete Certbot files associated with a certificate lineage."""
@@ -166,34 +166,30 @@ def _report_lines(msgs):
     """Format a results report for a category of single-line renewal outcomes"""
     return "  " + "\n  ".join(str(msg) for msg in msgs)
 
-def _report_human_readable(parsed_certs):
+def _report_human_readable(config, parsed_certs):
     """Format a results report for a parsed cert"""
     certinfo = []
-    checker = ocsp.RevocationChecker()
+    checker = ocsp.RevocationChecker(config)
     for cert in parsed_certs:
         now = pytz.UTC.fromutc(datetime.datetime.utcnow())
-        expiration_text = ""
-        revoked = checker.check_ocsp(cert.cert, cert.chain)
-        if revoked:
-            expiration_text = "INVALID: " + revoked
-        elif cert.is_test_cert:
-            expiration_text = "INVALID: TEST CERT"
-        elif cert.target_expiry <= now:
-            expiration_text = "INVALID: EXPIRED"
+        status = ""
+        if cert.is_test_cert:
+            status = "INVALID: TEST_CERT"
+        if cert.target_expiry <= now:
+            status = status + ",EXPIRED" if status else "INVALID: EXPIRED"
         else:
-            if revoked:
-                expiration_text = "INVALID: " + revoked
+            status = checker.ocsp_status(cert.cert, cert.chain, status)
 
-        if not expiration_text:
+        if not status:
             diff = cert.target_expiry - now
             if diff.days == 1:
-                expiration_text = "VALID: 1 day"
+                status = "VALID: 1 day"
             elif diff.days < 1:
-                expiration_text = "VALID: {0} hour(s)".format(diff.seconds // 3600)
+                status = "VALID: {0} hour(s)".format(diff.seconds // 3600)
             else:
-                expiration_text = "VALID: {0} days".format(diff.days)
+                status = "VALID: {0} days".format(diff.days)
 
-        valid_string = "{0} ({1})".format(cert.target_expiry, expiration_text)
+        valid_string = "{0} ({1})".format(cert.target_expiry, status)
         certinfo.append("  Certificate Name: {0}\n"
                         "    Domains: {1}\n"
                         "    Expiry Date: {2}\n"
@@ -206,7 +202,7 @@ def _report_human_readable(parsed_certs):
                             cert.privkey))
     return "\n".join(certinfo)
 
-def _describe_certs(parsed_certs, parse_failures):
+def _describe_certs(config, parsed_certs, parse_failures):
     """Print information about the certs we know about"""
     out = []
 
@@ -217,7 +213,7 @@ def _describe_certs(parsed_certs, parse_failures):
     else:
         if parsed_certs:
             notify("Found the following certs:")
-            notify(_report_human_readable(parsed_certs))
+            notify(_report_human_readable(config, parsed_certs))
         if parse_failures:
             notify("\nThe following renewal configuration files "
                "were invalid:")
--- a/certbot/cert_manager.py
+++ b/certbot/cert_manager.py
@@ -169,11 +169,14 @@ def _report_lines(msgs):
 def _report_human_readable(parsed_certs):
     """Format a results report for a parsed cert"""
     certinfo = []
+    checker = ocsp.RevocationChecker()
     for cert in parsed_certs:
         now = pytz.UTC.fromutc(datetime.datetime.utcnow())
         expiration_text = ""
-        revoked = ocsp.revoked_status(cert.cert, cert.chain)
-        if cert.is_test_cert:
+        revoked = checker.check_ocsp(cert.cert, cert.chain)
+        if revoked:
+            expiration_text = "INVALID: " + revoked
+        elif cert.is_test_cert:
             expiration_text = "INVALID: TEST CERT"
         elif cert.target_expiry <= now:
             expiration_text = "INVALID: EXPIRED"
--- a/certbot/cert_manager.py
+++ b/certbot/cert_manager.py
@@ -171,6 +171,10 @@ def _report_human_readable(config, parsed_certs):
     certinfo = []
     checker = ocsp.RevocationChecker(config)
     for cert in parsed_certs:
+        if config.certname and cert.lineagename != config.certname:
+            continue
+        if config.domains and not set(config.domains).issubset(cert.names()):
+            continue
         now = pytz.UTC.fromutc(datetime.datetime.utcnow())
         status = ""
         if cert.is_test_cert:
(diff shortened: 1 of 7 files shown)


pde:
Example output:
```
certbot -h certificates
usage: 

  certbot certificates [options] ...

Print information about the status of certificates managed by Certbot.

optional arguments:
                        and ~/.config/letsencrypt/cli.ini)

certificates:
  List certificates managed by Certbot

  -d DOMAIN, --domains DOMAIN, --domain DOMAIN
                        Domain names to apply. For multiple domains you can
                        use multiple -d flags or enter a comma separated list
                        of domains as a parameter. (default: Ask)
  --cert-name CERTNAME  Certificate name to apply. Only one certificate name
                        can be used per Certbot run. To see certificate names,
                        run 'certbot certificates'. When creating a new
                        certificate, specifies the new certificate's name.
                        (default: None)
  --check-ocsp CHECK_OCSP
                        Whether to check OCSP for listed certs. Can be set to
                        "never", "always", or "lazy" (ie, only for certs that
                        are otherwise valid). (default: always)
```
```
certbot certificates
OCSP check failed for /etc/letsencrypt/live/www.isnot.org/cert.pem (are we offline?)

Found the following certs:
  Certificate Name: isnot.org
    Domains: isnot.org goxogle.com is.isnot.org www.isnot.org
    Expiry Date: 2017-03-20 17:57:00+00:00 (VALID: 89 days)
    Certificate Path: /etc/letsencrypt/live/isnot.org/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/isnot.org/privkey.pem
  Certificate Name: goxogle.com
    Domains: goxogle.com www.isnot.org
    Expiry Date: 2016-04-07 23:39:00+00:00 (INVALID: TEST_CERT,EXPIRED)
  Certificate Name: demo.reworld.org
    Domains: demo.reworld.org
    Expiry Date: 2017-03-14 22:41:00+00:00 (VALID: 83 days)
    Certificate Path: /etc/letsencrypt/live/demo.reworld.org/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/demo.reworld.org/privkey.pem
  Certificate Name: www.isnot.org
    Domains: www.isnot.org
    Expiry Date: 2017-02-28 23:41:00+00:00 (INVALID: TEST_CERT)
    Certificate Path: /etc/letsencrypt/live/www.isnot.org/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/www.isnot.org/privkey.pem
  Certificate Name: isnot.org-0001
    Domains: isnot.org
    Expiry Date: 2017-03-14 22:41:00+00:00 (INVALID: REVOKED)
    Certificate Path: /etc/letsencrypt/live/isnot.org-0001/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/isnot.org-0001/privkey.pem
```
```
certbot certificates -d isnot.org

Found the following matching certs:
  Certificate Name: isnot.org
    Domains: isnot.org goxogle.com is.isnot.org www.isnot.org
    Expiry Date: 2017-03-20 17:57:00+00:00 (VALID: 89 days)
    Certificate Path: /etc/letsencrypt/live/isnot.org/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/isnot.org/privkey.pem
  Certificate Name: isnot.org-0001
    Domains: isnot.org
    Expiry Date: 2017-03-14 22:41:00+00:00 (INVALID: REVOKED)
    Certificate Path: /etc/letsencrypt/live/isnot.org-0001/fullchain.pem
    Private Key Path: /etc/letsencrypt/live/isnot.org-0001/privkey.pem
```

pde:
This still needs unit tests, but review can probably start in the mean time.

certbot/cert_manager.py
```diff
@@ -165,24 +166,34 @@ def _report_lines(msgs):
...
+        status = ""
         if cert.is_test_cert:
-            expiration_text = "INVALID: TEST CERT"
-        elif cert.target_expiry <= now:
-            expiration_text = "INVALID: EXPIRED"
+            status = "INVALID: TEST_CERT"
+        if cert.target_expiry <= now:
+            status = status + ",EXPIRED" if status else "INVALID: EXPIRED"
         else:
+            status = checker.ocsp_status(cert.cert, cert.chain, status)
```
ohemorange: cleaner:
```
reasons = []
if cert.is_test_cert:
  reasons.append('TEST CERT')
if cert.target_expiry <= now:
  reasons.append('EXPIRED')
if checker.check_ocsp(...):
  reasons.append('REVOKED')
if reasons:
  status = "INVALID: " + ", ".join(reasons) 
```

certbot/ocsp.py
```diff
@@ -0,0 +1,136 @@
...
+        :param str cert_path: path to a cert to check
+        :param str chain_path: issuing intermediate for the cert
+        :param str status_in: a string that is either empty, if the cert is otherwise
+                              believed to be valid, or 'INVALID: $REASON'.
+
+        :returns: a new status including revocation, if the cert is revoked."""
+
+        if self.config.check_ocsp.lower() == "never":
+            return status_in
+        elif self.config.check_ocsp.lower() == "lazy" and status_in:
```
ohemorange: this check can go in `cert_manager` if you really want it to exist, then I think we don't need this function at all.

certbot/cert_manager.py
pde: You're probably right. Of course, I read this right after I finished writing tests for `ocsp_status` :)

certbot/cli.py
```diff
@@ -944,6 +946,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help=config_help("no_verify_ssl"),
         default=flag_default("no_verify_ssl"))
     helpful.add(
+        ["certificates", "testing"], "--check-ocsp", default="always",
```
pde: `dest`s default to the flag name, with `-` replaced by `_`.

pde: The `never` option is there in case people don't want `certbot certificates` to be hitting the network.  The `lazy` option claims to be there in case people have a really slow and terrible network, and they don't want to wait for it unnecessarily. But it's not clear that we really need lazy?

certbot/tests/cert_manager_test.py
```diff
@@ -179,7 +179,9 @@ def test_certificates_no_files(self, mock_utility, mock_logger):
         self.assertTrue(mock_utility.called)
         shutil.rmtree(tempdir)
 
-    def test_report_human_readable(self):
+    @mock.patch('certbot.cert_manager.ocsp.RevocationChecker.ocsp_status')
```
pde: Yeah, that form works too; it walks in via `cert_manager`'s reference to `ocsp`.

certbot/cert_manager.py
```diff
@@ -165,24 +166,37 @@ def _report_lines(msgs):
     """Format a results report for a category of single-line renewal outcomes"""
     return "  " + "\n  ".join(str(msg) for msg in msgs)
 
-def _report_human_readable(parsed_certs):
+def _report_human_readable(config, parsed_certs):
     """Format a results report for a parsed cert"""
     certinfo = []
+    checker = ocsp.RevocationChecker()
     for cert in parsed_certs:
+        if config.certname and cert.lineagename != config.certname:
```
ohemorange: Please add tests for these filters.

certbot/ocsp.py
```diff
@@ -0,0 +1,109 @@
...
+               "-no_nonce",
+               "-issuer", chain_path,
+               "-cert", cert_path,
+               "-url", url,
+               "-CAfile", chain_path,
+               "-verify_other", chain_path,
+               "-header"] + self.host_args(host)
+        try:
+            output, err = util.run_script(cmd, log=logging.debug)
+        except errors.SubprocessError as e:
```
ohemorange: 62-65 are missing coverage. We're so close to full coverage!

certbot/ocsp.py
```diff
@@ -0,0 +1,109 @@
...
+            logger.info("Cannot process OCSP host from URL (%s) in cert at %s", url, cert_path)
+            return None, None
+
+def _translate_ocsp_query(cert_path, ocsp_output, ocsp_errors):
+    """Parse openssl's weird output to work out what it means."""
+
+    if not "Response verify OK" in ocsp_errors:
+        logger.info("Revocation status for %s is unknown", cert_path)
+        logger.debug("Uncertain ouput:\n%s\nstderr:\n%s", ocsp_output, ocsp_errors)
+        return None
```
ohemorange: Having both `return None` and `return False` seems error-prone - the `None`s can raise a specific error instead, and have the caller catch it. I think that's the pythonic thing anyway.

certbot/tests/ocsp_test.py
```diff
@@ -0,0 +1,135 @@
...
+
+openssl_revoked = ("blah.pem", """
+blah.pem: revoked
+	This Update: Dec 20 01:00:00 2016 GMT
+	Next Update: Dec 27 01:00:00 2016 GMT
+	Revocation Time: Dec 20 01:46:34 2016 GMT
+""",
+"""Response verify OK""")
+
+openssl_broken = ("", "tentacles", "Response verify OK")
```
ohemorange: XD

certbot/ocsp.py
pde: This was by design: the calling code should treat either `None` or `False` as "move along, we don't believe this is revoked", but it has the option of looking at the value to learn a bit more, if desired.

pde: I currently can't imagine situations where Certbot will want to treat OCSP query failures as fatal; if we ever do, we can refactor at that point.

pde:
OK, extra tests added. This PR should be merged rather than squashed to give James credit for having written the first version of it!

certbot/ocsp.py
ohemorange: It shouldn't treat them as fatal, but nor should it make it easy to treat a query failure in precisely the same way as a definitely unrevoked cert. It's python, exceptions aren't fatal. https://jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/

Written this way, in idiomatic checks, this function will check if the cert is revoked or unknown, not revoked or unrevoked. Which works for how it's called now, but would require refactoring were we ever to want to use it again. That is, people will call it as:

```
if ocsp_revoked:
  do_something()
```
which works fine, but
```
if not ocsp_revoked():
  do_something_else()
```
is not analogous and might lead to future buggy code.

certbot/cert_manager.py
```diff
@@ -165,24 +166,37 @@ def _report_lines(msgs):
...
+        reasons = []
         if cert.is_test_cert:
-            expiration_text = "INVALID: TEST CERT"
-        elif cert.target_expiry <= now:
-            expiration_text = "INVALID: EXPIRED"
+            reasons.append('TEST_CERT')
+        if cert.target_expiry <= now:
+            reasons.append('EXPIRED')
+        if checker.ocsp_revoked(cert.cert, cert.chain):
+            reasons.append('REVOKED')
```
ohemorange: This line is missing coverage.

certbot/ocsp.py
pde: Why do you think that second example might lead to bugs? AFAICT in both of your cases the inability to check revocation is treated the same way as nonrevocation by default (though of course there are some log messags about the problem). That's what I suspect we'll always want in Certbot.

If OTOH the calling site wants to fail closed on OCSP checking failures it can do:
```
revoked = ocsp_revoked()
if revoked is None:
    hard_fail()
elif revoked:
    do_something()
```
or
```
revoked = ocsp_revoked()
if revoked is None:
    hard_fail()
elif not_revoked:
    do_something_else()
```

PR #3948, opened 2016-12-22
Save allow_subset_of_names in renewal conf files

### LINKED ISSUE #3888: --allow-subset-of-names should probably be a renewalparam

bmw:
Fixes #3888.


bmw:
@schoen, it looks like there is some duplicate work between this PR and #3849 which I unfortunately didn't notice before writing this. Since I believe you were planning on fixing up #3849, would you like to choose between these two approaches (or implement your own thing)? With a way to save/restore bools, adding a new relevant config item should be as simple as adding an element to `BOOL_CONFIG_ITEMS` and adding a test.

thomaszbz:
This PR includes my commits for #3844 which is why it also

Fixes #3844

PR #3949, opened 2016-12-22
Start boulder while unit tests are running

bmw:


bmw:
I think this is a bit of a failed experiment. Integration tests were slightly faster, however, it causes boulder start up output to be interspersed with output from our unit tests. I personally find this annoying enough to suggest we don't take this PR.

I think the biggest speed increase to our integration tests would be to use a prebuilt boulder docker image on DockerHub or Quay.io instead of building it ourselves each time. This is especially true since the branch of boulder we use hasn't changed since May and we rebuild it for every integration test!

bmw:
Thought about this a bit more. I think the biggest speed improvements for us in order of desirability are:

1. Use a remote `boulder` server for integration tests. The reason we can't do this now is we don't have valid domain(s) to use in Travis.
2. Use a prebuilt `boulder` docker image as I mentioned above.
3. Land a PR like this.

Obviously doing both 2 and 3 is better than just doing one or the other.

PR #3972, opened 2017-01-04
Enable Py36 Tests

bmw:
[Python 3.6 was released](https://www.python.org/downloads/release/python-360/).


bmw:
Oh man. Looking more closely at the Travis logs, there's a lot of tests that need changed here. I'll work more on this in the next few days. I thought this would be a trivial PR!

bmw:
Almost all required changes were due to us passing mock objects into `os.path.abspath`.

bmw:
I created #4078 to track creating utility classes.

PR #3975, opened 2017-01-05
added basic manual instructions

SwartzCr:
In progress
pulled from joohoi's PR - feel free to comment @joohoi & @bmw

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -55,7 +55,8 @@ standalone_ Y    N    | Uses a "standalone" webserver to obtain a cert. Requires
                       | with no webserver, or when direct integration with the local
                       | webserver is not supported or not desired.
 manual_     Y    N    | Helps you obtain a cert by giving you instructions to perform http-01_ (80) or
-                      | domain validation yourself.                                   dns-01_ (53)
+                      | domain validation yourself. Additionally allows you to        dns-01_ (53)
+                      | specify scripts to be run before and after validation.
 =========== ==== ==== =============================================================== =============================
 
 Under the hood, plugins use one of several ACME protocol "Challenges_" to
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -55,7 +55,8 @@ standalone_ Y    N    | Uses a "standalone" webserver to obtain a cert. Requires
                       | with no webserver, or when direct integration with the local
                       | webserver is not supported or not desired.
 manual_     Y    N    | Helps you obtain a cert by giving you instructions to perform http-01_ (80) or
-                      | domain validation yourself.                                   dns-01_ (53)
+                      | domain validation yourself. Additionally allows you to        dns-01_ (53)
+                      | specify scripts to be run before and after validation.
```
pde: Not just "scripts to be run before and after validation", but "scripts to let you automate the validation task in a customized way", or something to that effect?

docs/using.rst
```diff
@@ -168,6 +169,10 @@ the UI, you can use the plugin to obtain a cert by specifying
 to copy and paste commands into another terminal session, which may
 be on a different computer.
 
+Additionally you can specify scripts to run before and after verification
```
pde: I think people might be confused by this before/after notion. It's true that the scripts run before/after boulder does its validation, but from the user's perspective these scripts actually perform the authentication/validation task, and clean up from it?

pde: I guess either "perform authentication" or "prepare for validation" could be better ways of putting it?

docs/using.rst
```diff
@@ -416,6 +423,129 @@ The following files are available:
    could convert using ``openssl``. You can automate that with
    ``--renew-hook`` if you're using automatic renewal_.
 
+.. _hooks:
+
+Pre and Post Validation Hooks
+=============================
+
+Certbot allows for the specification fo pre and post validation hooks when run
```
jsoref: fo => of

Issue #3976, opened 2017-01-05
Refactor renewal.py


bmw:
I had to add `# pylint: disable=too-many-branches` to `handle_renewal_request()` today. We should refactor functions like this to make them more manageable, testable, and easier to read.

bmw:
I ended up being able to remove this comment, but I think this is still a worthwhile task.

pde:
One way to do this would be to make a small class that could contain the lists `renew_successes`, `renew_failures`, `renew_skipped` and `parse_failures` and provide methods for adding to and reporting based on them. All the logging and output could be side effects of adding to the lists.

pde:
Have been slightly tempted to do that for a long time.

bmw:
Yeah I remember you talking about that.

An approach like that could work. If this is done, the class should probably be placed in it's own file and have it's own unit tests.

bmw:
Honestly, this file doesn't seem too bad anymore. I think `handle_renewal_request` is a little long and could be stand to be broken up a bit, but the file itself isn't that long and most functions are reasonably short.

lmeerwood:
If you don't mind I could try implement a new class to handle those lists. I'm assuming that the "easy-fix" tag is for beginner contributors like me?

adferrand:
Yes I am still not very satisfied by the code I wrote in `handle_renewal_request` even if I think we agree that is better than before. The two things that bother me are that error handling is way too visible, and the loop that feeds 4 arrays is super weird.

You can give a try, I will gladly review it. Be really careful about respecting the logic, in particular about when it waits, and when and how it errors out, because this is really sensible.

lmeerwood:
Sorry, my circumstances have changed and I no longer have the free time to attempt this.

PR #3977, opened 2017-01-05
Correctly report when we skip hooks during renewal

bmw:


bmw:
I created #3976 for the lint check I had to disable.

ohemorange:
Why is it closed?

bmw:
Because I'm indecisive. One sec.

certbot/tests/hook_test.py
```diff
@@ -62,10 +62,7 @@ def _test_renew_post_hooks(self, expected_count):
             with mock.patch('certbot.hooks._run_hook') as mock_run:
                 hooks.run_saved_post_hooks()
                 self.assertEqual(mock_run.call_count, expected_count)
-                if expected_count:
-                    self.assertEqual(mock_info.call_count, expected_count)
-                else:
-                    self.assertEqual(mock_info.call_count, 1)
+                self.assertEqual(mock_info.call_count, expected_count)
```
ohemorange: nice

certbot/renewal.py
```diff
@@ -304,6 +304,8 @@ def _renew_describe_results(config, renew_successes, renew_failures,
         notify(report(renew_skipped, "skipped"))
     if not renew_successes and not renew_failures:
         notify("No renewals were attempted.")
+        if config.pre_hook is not None or config.post_hook is not None:
```
ohemorange: Should we also check for renew hooks?

bmw: In my mind they are a little different, but that's probably just me. I'll send a follow up commit so we're consistent here.

ohemorange: Why are they different?

bmw: Renew hooks are specifically run after each cert is renewed. With this understanding, I personally wouldn't expect a renew hook to be run if no certs are renewed.

Pre and post hooks on the other hand are respectively run before or after attempting renewal. To me, it isn't as clear whether or not I should expect Certbot to run them if nothing needs to be renewed.

I'm fine making this change though. Just waiting on tests to finish locally.

ohemorange: You're right that that one should be kind of obvious. I'm cool with not checking, or with checking.

bmw:
I don't think it hurts anything to print the message for renew hook. The message is still accurate and may help someone if they don't totally understand how `--renew-hook` works.

PR #3988, opened 2017-01-09
Add option to specify revocation reason (#3242)

dashaxiong:
Addresses issue #3242. 
This includes two new tests in the integration test script to check that
boulder gets the correct code. The encoding is specified in RFC5280
5.3.1. The codes that boulder will accept are a subset of that,
specified in `boulder.revocation.reasons.go`. 

Supersedes #3903 to clean up the commit history (wrong email address on 
earlier commits). That branch had incorporated merges from master on 
several occasions, so I was uncertain of the effect of `git filter-branch ...` and
thought it better to apply the diff to a fresh branch from master.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -477,17 +477,21 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
                 "Recursion limit reached. Didn't get {0}".format(uri))
         return chain
 
-    def revoke(self, cert):
+    def revoke(self, cert, rsn):
         """Revoke certificate.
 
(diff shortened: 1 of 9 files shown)


dashaxiong:
A couple of notes about this PR. 
1. It doesn't prompt for a reason if none is given. I can come back to this and add that if necessary.
2. The option choices are presented as all lower case, but are actually case insensitive. I can come back to this and try to refine that further so that the options choices are presented in camel case but still case insensitive.

pde:
@rolandshoemaker were you happy with this? If so will Github let you approve-review it?

PR #3990, opened 2017-01-09
pyopenssl inject workaround

oerdnj:
This adds workaround for platforms where injecting pyopenssl to urllib3 fails with requests package (Ubuntu Trusty) originally started as #3779.

The whole block is covered by `#pragma: no cover` and I've added `# pylint: disable=import-error` as requested in original PR.

`tox -e lint` is now fine.


pde:
Actually just merging this, I think it's safe.

PR #4004, opened 2017-01-09
Attempt at updating configuration from existing lineage

schoen:


schoen:
Should fix #3997.

schoen:
@ohemorange @pde I don't actually have a setup to check if this succeeds in affecting #3997; can anyone opine on that either from a theoretical or empirical point of view?

schoen:
@bmw perhaps you can also see if you think this is correct :slightly_smiling_face:

ohemorange:
Tested it, this correctly applies a saved-but-not-cli `--staging` command that was not previously being applied.

certbot/main.py
```diff
@@ -85,6 +85,18 @@ def _auth_from_available(le_client, config, domains=None, certname=None, lineage
...
+        if action in ("renew", "reinstall"):
+            # Found existing lineage; update config based on its configuration.
+            try:
+                # Some of the steps of reconstitute(), but without copying
+                # domains and without as much sanity checking.
+                renewalparams = lineage.configuration["renewalparams"]
+                renewal.restore_required_config_elements(config, renewalparams)
+                renewal.restore_plugin_configs(config, renewalparams)
+            except (KeyError, ValueError):
+                # This doesn't make things worse... right?
```
pde: I'd at least `log.info()` the error and `log.debug()` the traceback here rather than silently ignoring it!

certbot/main.py
```diff
@@ -85,6 +85,18 @@ def _auth_from_available(le_client, config, domains=None, certname=None, lineage
     if lineage is None:
         # This will find a relevant matching lineage that exists
         action, lineage = _find_lineage_for_domains_and_certname(config, domains, certname)
+        if action in ("renew", "reinstall"):
+            # Found existing lineage; update config based on its configuration.
+            try:
+                # Some of the steps of reconstitute(), but without copying
```
pde: Aesthetically I now wonder if maybe we shouldn't have added a flag `include_domains=True` to `_reconstitute()` and just called that.  But I'm fine either way, I think!

pde:
Checklist to self before merging:
* [ ] walk through all restored variables and convince self that they will behave as expected in interactions with UI
* [ ] maybe get a second review from @ohemorange

pde:
Walking through the restored variables:
 * `config_dir` is in there. It's not clear that should be something a renewal conf file can change, but under normal circumstances it would be hard for it to be set to anything different than whatever value the user set with `--config-dir` on the second run.
* `allow_subset_of_names` is a bit weird to persist (it certainly makes sense for `certbot renew` and non-interactive `certonly` runs, but one could imagine popping an interactive question rather than assuming it should still be true). But I don't think that's a showstopper.

pde:
More problematically, the attempt to restore `account`, `server`, and `staging` variables is ineffectual because we've already called `_init_le_client` by the time this happens. A band-aid for that could be to push that call down the stack and make it inside `_auth_from_available`, right before we first call methods on `le_client`. We would then pass the client back up as an additional return variable.

A cleaner solution would be to split `_auth_from_available` into a "figure out what we're doing" function and then an "actually do stuff" function, with the client being initialised in between.  I'm going to take a look at @ohemorange's refactor to see if that's already in the works :)

bmw:
What's the status of this?

ohemorange:
Update! The main feedback from the community forum is that no one really expects anything here, because it's un-unix-y that we store command line arguments at all. So no one expects it to be reconstituted so they just always use all the flags they'd want. Which is important design information to keep in mind for other issues as well, but for the short term I'm going to close this to conserve bandwidth. Thanks for the work @schoen!

PR #4008, opened 2017-01-10
Add update_symlinks to "--help manage"

### LINKED ISSUE #3995: Put update_symlinks in certbot --help manage

pde:
Fixes #3995


ohemorange:
Make sure to remove `rename` when fixing the merge conflict, otherwise LGTM.

ohemorange:
But I think we should find a way to write a non-unit test that fails without this change and passes with it before merging this, and I still think we should do that after we release 0.10.0.

PR #4010, opened 2017-01-10
Stop IDisplay AssertionErrors

### LINKED ISSUE #3996: certbot delete and rename evoke IDisplay bug

bmw:
Fixes #3996.

I'm pretty confident this PR solves the problem. I've audited all calls to `IDisplay` methods and the assertions done in `certbot.display.util` are now done in all our unit tests.

With that said, it wouldn't hurt to have someone else double check I didn't miss anything. The easiest way to do this is to `grep` for `IDisplay` in our code and ensure all calls to `IDisplay` methods are valid. This means every method call other than `notification` (because a `notification` call is always OK) either provides a value for `default` or `force_interactive`. This is defined in `interfaces.py`.

I've also been considering removing the assertion that's been causing us trouble here from our release. The only argument I have for not doing so is it may hinder 3rd party plugin development. When they use `IDisplay`, they have the same problem as we do with prompting users without a TTY. Not keeping this assertion in makes it more likely they won't notice the issue and Certbot will crash on an unsuspecting user.

With that said, none of our known 3rd party plugins use `IDisplay` at all.

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -253,7 +253,7 @@ class FileDisplay(object):
         :rtype: bool
 
         """
-        assert_valid_call(prompt, default, cli_flag, force_interactive)
+        # assert_valid_call(prompt, default, cli_flag, force_interactive)
         if self._can_interact(force_interactive):
             return False
         elif default is None:
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -256,6 +256,13 @@ class FileDisplay(object):
         assert_valid_call(prompt, default, cli_flag, force_interactive)
         if self._can_interact(force_interactive):
             return False
+        elif default is None:
+            msg = "Unable to get an answer for the question:\n{0}".format(prompt)
+            if cli_flag:
+                msg += (
+                    "\nYou can provide an answer on the "
+                    "command line with the {0} flag.".format(cli_flag))
+            raise errors.Error(msg)
         else:
             logger.debug(
                 "Falling back to default %s for the prompt:\n%s",
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -253,12 +253,7 @@ class FileDisplay(object):
         :rtype: bool
 
         """
-        msg = "Invalid IDisplay call for this prompt:\n{0}".format(prompt)
-        if cli_flag:
-            msg += ("\nYou can set an answer to "
-                    "this prompt with the {0} flag".format(cli_flag))
-        assert default is not None or force_interactive, msg
-
+        assert_valid_call(prompt, default, cli_flag, force_interactive)
         if self._can_interact(force_interactive):
             return False
         else:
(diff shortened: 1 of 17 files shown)


ohemorange:
```
    def _test_perform_bind_errors(self, errno, achalls, mock_get_utility):
        port = get_open_port()
        def _perform2(unused_achalls):
            raise errors.StandaloneBindError(mock.Mock(errno=errno), port)

        self.auth.perform2 = mock.MagicMock(side_effect=_perform2)
        self.auth.perform(achalls)
        mock_get_utility.assert_called_once_with(interfaces.IDisplay)
        notification = mock_get_utility.return_value.notification
        self.assertEqual(1, notification.call_count)
        self.assertTrue(str(port) in notification.call_args[0][0])
```
Can we get weird interactions here/cases like this, where it does or doesn't pass things on to the internal mock object in unexpected ways?

certbot/tests/util.py
```diff
@@ -158,3 +162,87 @@ def make_lineage(self, testfile):
...
+            return setattr(self._mock, name, value)
+        elif name != '_frozen_set':
+            self._frozen_set.add(name)
+        return object.__setattr__(self, name, value)
+
+
+def _create_get_utility_mock():
+    display = FreezableMock()
+    for name in interfaces.IDisplay.names():  # pylint: disable=no-member
+        if name != 'notification':
```
ohemorange: Are we sure we want to exclude notification? What if an existing method that we somehow magically missed *really* wants to force a pause?

certbot/tests/util.py
```diff
@@ -158,3 +162,87 @@ def make_lineage(self, testfile):
...
+            except AttributeError:
+                return False
+        elif name == '_frozen_set' or name in self._frozen_set:
+            return object.__getattribute__(self, name)
+        else:
+            return getattr(object.__getattribute__(self, '_mock'), name)
+
+    def __setattr__(self, name, value):
+        if self._frozen:
+            return setattr(self._mock, name, value)
```
ohemorange: Any chance that setting a frozen attribute silently in the mock could blow up? Probably not yeah?

ohemorange:
Oh guess I should wait to approve until the follow-up commit.

bmw:
In this case, everything works as I intended (I know because initially I was improperly patching this test and it failed :smile: ). Every access of `mock_get_utility` attributes in this function is passed to the underlying `mock`.

certbot/display/util.py
```diff
@@ -253,9 +253,16 @@ def _return_default(self, prompt, default, cli_flag, force_interactive):
         :rtype: bool
 
         """
-        assert_valid_call(prompt, default, cli_flag, force_interactive)
+        # assert_valid_call(prompt, default, cli_flag, force_interactive)
         if self._can_interact(force_interactive):
             return False
+        elif default is None:
```
ohemorange: Alternatively, try to interact anyway? I'm worried about false negatives on `_can_interact` where there isn't a cli flag, which will totally break the feature.

bmw: Hmm. If we try and interact and the user doesn't have a TTY, `certbot` will indefinitely hang. Perhaps I should mention `--force-interactive` in the error message?

bmw: Although info about `--force-interactive` is given in the warning in `_can_interact`.

ohemorange: Oh, if it's given there that's fine.

PR #4020, opened 2017-01-11
Added Heroku to the list of third-party plugins

gboudreau:
I published a new plugin for Heroku. Simply adding it to the list of third-party plugins.


gboudreau:
Quick Q: my `certbot-auto` upgraded to 0.10.0 this morning, and my plugin disappeared. I think it's because `certbot-auto` deletes the previous virtual env. and creates a new one. Is there something I can do in my plugin to make `certbot-auto` reinstall it when upgrading, or should I open a new issue (in this repository?) to report this bug in `certbot-auto`?

bmw:
I created #4037 to track the issue and mentioned you as the one who reported the issue. Thanks for letting us know. This certainly isn't ideal.

PR #4021, opened 2017-01-11
Mention first domain = name in --help output

bmw:
Supersedes #3551.

This is worth mentioning because it surprises people, but I'm not sure we should be going into the full complexity of:
> The first provided domain will be used to decide where to store the new certificate, unless --cert-name is provided or an existing certificate is found with identical domains.

We'd also have to remember to keep help text like that up to date with new Certbot changes, while "otherwise specified' takes care of itself.


bmw:
Don't merge this yet. I want to bring in the original commit from #3551.

pde:
I changed this a bit to do a better job of explaining what cert names are for and where they come from. Probably best for @bmw to review/merge now.

certbot/cli.py
```diff
@@ -831,14 +831,19 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         metavar="DOMAIN", action=_DomainsAction, default=[],
         help="Domain names to apply. For multiple domains you can use "
              "multiple -d flags or enter a comma separated list of domains "
-             "as a parameter. (default: Ask)")
+             "as a parameter. (default: Ask) Note: the first provided domain "
+             "will be most prominent in filenames and some software user "
+             "interfaces")
```
bmw: I think this information is inaccurate now. The first domain will not be most prominent if you use `--cert-name` or you have a matching certificate with the same names and didn't include `--duplicate`.

I also personally think the extra information after `(default: Ask)` looks bad because it breaks the convention used by all other flags, but that's a stylistic preference. If you're okay with that, we can leave it as is.

certbot/cli.py
```diff
@@ -831,14 +831,19 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
         metavar="CERTNAME", default=None,
-        help="Certificate name to apply. Only one certificate name can be used "
-             "per Certbot run. To see certificate names, run 'certbot certificates'. "
-             "When creating a new certificate, specifies the new certificate's name.")
+        help="Certificate name to apply. This name is used by Certbot for housekeeping "
+             "and in file paths; it doesn't affect the content of the certificate itself. "
+             "To see certificate names, run 'certbot certificates'. "
+             "When creating a new certificate, specifies the new certificate's name. "
+             " (default: use the first of the domain names provided with -d or entered "
+             "interactively)")
```
bmw: The default isn't correct if we find an existing certificate with the same domains, but perhaps that's OK?

bmw:
Closing in favor of #4896.

PR #3929, opened 2016-12-16
Add bootstrapping of alpine deps

willfarrell:
Closes: #3761

- [x] Update letsencrypt-auto source
- [x] Add Travis-CI Alpine Docker Test
- [x] Pass: Make sure we stop if the openssl signature check fails.
- [x] Pass: Make sure pip stops us if there is a hash mismatch.
- [ ] Pass: Exercise most branches of letsencrypt-auto.
- [ ] Docs if needed

Issue #492, opened 2015-06-11
Docker Hub


kuba:
@JensErat asked in https://github.com/JensErat/docker-lets-encrypt-preview/issues/2#issuecomment-111083470 if we could upload Docker images not only to Quay, but also to Docker Hub.

JensErat:
Maybe I'm overlooking something, but I don't think this involves any further but giving the "official" Docker Hub permissions to install an on-commit trigger.

In the end, the official Docker Hub will be what the vast majority of Docker users is using, and most of them will strongly dislike uploaded images and prefer automated builds.

DoubleMalt:
I agree. It is dead easy to set up an automated build and it inspires more confidence as it is built directly from the published code.

ArchimedesPi:
@kuba: it's very easy to set up, [here are instructions](https://docs.docker.com/docker-hub/github/).
I think it would help a lot with testing.

ludovicc:
Quay.io uses an old, insecure registry and it prevents Ansible docker to execute the letsencrypt image

TASK: [letencrypt-nginx | Run Letencrypt in Docker] *************************** 
failed: ... => {"changes": ["{\"status\":\"Pulling repository quay.io/letsencrypt/letsencrypt\"}\r\n",  ... "{\"status\":\"Status: Image is up to date for quay.io/letsencrypt/letsencrypt:latest\"}\r\n", "{\"status\":\"quay.io/letsencrypt/letsencrypt: this image was pulled from a legacy registry.  Important: This registry version will not be supported in future versions of docker.\"}\r\n"], "failed": true, "item": {"auto_dir": "/tmp/letencrypt-auto", "domain": "hbps1.chuv.ch"}, "status": "quay.io/letsencrypt/letsencrypt: this image was pulled from a legacy registry.  Important: This registry version will not be supported in future versions of docker."}
msg: Unrecognized status from pull.

czerasz:
:+1: 

This would be great to have an official Let's Encrypt Docker Image in the Docker Hub.

Now I have to build it manually (Ansible):
```
- name: Build the Let's Encrypt Docker image
  docker_image:
    path: "github.com/letsencrypt/letsencrypt.git"
    name: "my-user/letsencrypt"
    state: present
```

There is no easy update/version control over it :-(.

programmerq:
I'd love to see a letsencrypt official image. `docker run letsencrypt ...`

https://github.com/docker-library/official-images#contributing-to-the-standard-library

[...]

PR #4032, opened 2017-01-12
Dockerfile refactored to use Alpine.

### LINKED ISSUE #3782: Rewrite our Dockerfile

Yajo:
Fix #3782, and hopefully help #492. Idea from #3929.

@Tecnativa

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,70 +1,30 @@
-# https://github.com/letsencrypt/letsencrypt/pull/431#issuecomment-103659297
-# it is more likely developers will already have ubuntu:trusty rather
-# than e.g. debian:jessie and image size differences are negligible
-FROM ubuntu:trusty
+FROM python:2-alpine
 MAINTAINER Jakub Warmuz <jakub@warmuz.org>
 MAINTAINER William Budington <bill@eff.org>
 
-# Note: this only exposes the port to other docker containers. You
-# still have to bind to 443@host at runtime, as per the ACME spec.
-EXPOSE 443
-
-# TODO: make sure --config-dir and --work-dir cannot be changed
-# through the CLI (certbot-docker wrapper that uses standalone
-# authenticator and text mode only?)
+EXPOSE 80 443
 VOLUME /etc/letsencrypt /var/lib/letsencrypt
-
 WORKDIR /opt/certbot
-
-# no need to mkdir anything:
-# https://docs.docker.com/reference/builder/#copy
-# If <dest> doesn't exist, it is created along with all missing
-# directories in its path.
-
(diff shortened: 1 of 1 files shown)


willfarrell:
Leveraging what will be in the certbot-auto, you could simplify it even more.

```Dockerfile
FROM alpine:3.5

RUN apk --no-cache add \
        bash \
        openssl \
        wget

# certbot
RUN cd /usr/bin \
    && wget https://dl.eff.org/certbot-auto \
    && chmod a+x certbot-auto \
    && certbot-auto --version --non-interactive
...
```

Yajo:
Cool idea, although I did not want to make this PR dependant on #3929. After that one is merged, this one could change, but this way it is not a requirement.

hamburml:
Any news? Would like to have a smaller container :)

willfarrell:
I actually switched to using dehydrated, smaller and has dns-01 and ecdsa support build it. You can find the docker image here: https://hub.docker.com/r/willfarrell/letsencrypt/

hamburml:
@willfarrell Thanks. Would also love to use an official image with tags pushed to docker hub for business usage.

Dockerfile
```diff
@@ -1,70 +1,30 @@
...
-# install in editable mode (-e) to save space: it's not possible to
-# "rm -rf /opt/certbot/src" (it's stays in the underlaying image);
-# this might also help in debugging: you can "docker run --entrypoint
-# bash" and investigate, apply patches, etc.
-
 ENTRYPOINT [ "certbot" ]
+COPY . src
+
+RUN apk add --no-cache --virtual .certbot-deps \
+        dialog \
```
bmw: This dependency was removed in Certbot 0.10.0. Looks like we forgot to update our current Dockerfile.

Dockerfile
```diff
@@ -1,70 +1,30 @@
...
+        augeas-libs \
```
bmw: If `certbot-apache` is no longer installed, this package can also be removed.

Dockerfile
```diff
@@ -1,70 +1,30 @@
...
+        libffi \
+        libssl1.0 \
+        wget \
```
bmw: Why is this package needed?

Yajo:
Thanks for your comments @bmw, I agree on all of them. I just wanted to provide something similar to what you had before, but more Docker-ish and Alpine-based; indeed, I think it's best that you, with more certbot knowledge than me, take this base and build on it what you want.

Yajo:
Cool! It would be great to trigger an automated build in the docker hub and link it in the readme, don't you think?

Issue #3915, opened 2016-12-15
Refactor IDisplay interface


bmw:
As of writing this, no third party plugins are using it. It's getting a little gross with tons of arguments being passed in, many of which aren't relevant anymore with the removal of `dialog` and lots of `*args` and `**kwargs` due to differences in `FileDisplay` and `NoninteractiveDisplay`. We should clean all this up while we have the chance.

bmw:
@zjs, ideally when we do this we won't break any 3rd party plugins. (We still technically haven't released 1.0 so anything can change according to semantic versioning, but that's a bad experience for 3rd party developers.) You can find the list of 3rd party plugins we know about [here](https://certbot.eff.org/docs/using.html#third-party-plugins). As of a few months ago, the only usage of `IDisplay` out of those plugins was [this](https://github.com/gboudreau/certbot-heroku/blob/2837a5bdfe527853311e57923557ec77b78c3317/certbot_heroku/configurator.py#L101). It's worth double checking this hasn't changed though.

zjs:
Additional uses of `IDisplay`:
 * https://github.com/EnigmaBridge/certbot-external-auth/blob/03dc577c33a0abcfbfbdf6d907e68142dee78904/certbot_external_auth/plugin.py#L802
 * https://github.com/thelabnyc/certbot-django/blob/a922536cc4a3db384b63fa0e3edb2d3d5e7acc13/src/certbot_django/coordinator/authenticator.py#L121
 * https://github.com/thelabnyc/certbot-openshift/blob/ee364bbe0e8ef8aae6fd71374e7497048e84a320/src/certbot_openshift/installer.py#L192

zjs:
Additional 3rd party plugins which don't use `IDisplay`:
 * https://github.com/robin-thoni/certbot-pdns
 * https://github.com/kontalk/certbot-tigase (WIP)
 * https://github.com/oGGy990/certbot-dns-inwx
 * https://github.com/kento1218/aws-lambda-letsencrypt
 * https://github.com/adepertat/certbot-cwebroot
 * https://github.com/hostinger/letsencrypt-redis
 * https://github.com/chrismarget/certbot-asa
 * https://github.com/nickbreen/docker-letsencrypt
 * https://github.com/sjerdo/letsencrypt-directadmin

bmw:
Woah. I had no idea there were that many. We should probably add them to our docs. I created #4828 to track that.

zjs:
[...]

PR #4072, opened 2017-01-18
Uncomment IDisplay Assertion

bmw:
This will hopefully help us catch bugs during development.


pde:
Is the plan to comment this in during development and then out for releases?

bmw:
For the meantime, yes. The idea is system tests and devs will run into it if a problem is introduced.

Hopefully a better plan will develop out of #3915 though.

PR #4088, opened 2017-01-19
Remove optional dependencies

### LINKED ISSUE #595: Standalone Authenticator should prompt repeatedly until 443 is available.
When calling standalone authenticator while nginx or apache is running it should automatically stop it and restart it after finishing auth instead of throwing an error.

```
letsencrypt -d example.com -t auth

How would you like to authenticate with Let's Encrypt CA?
1: Nginx Web Server (nginx)
2: Manual Authenticator (manual)
3: Standalone Authenticator (standalone)
Select the appropriate number [1-3] then [enter] (press 'c' to cancel): 3

The program nginx (process ID 19803) is already listening on TCP port 443. This
will prevent us from binding to that port. Please stop the nginx program
temporarily and then try again.
Press Enter to Continue
An unexpected error occurred. Please see the logfiles in /var/log/letsencrypt for more details.
```

bmw:
Fixes #595 and #3650.

`simple_verify` wasn't removed entirely or changed to raise `NotImplementedError` because at least one [other script](https://github.com/alex/letsencrypt-aws/blob/84d56c17800868f64d64564c3dd4c3e6de2e924a/letsencrypt-aws.py#L306) uses it. After removing DNS extras from `acme/setup.py`, `pip install acme[dns]` doesn't error out and `acme` is installed without the non-existent DNS extras. I tested this on both new and ancient versions of `pip`.


ohemorange:
Otherwise, the code is looking good! Love the red PR and breaking out functionality into smaller methods!

bmw:
Those calls to actually use other challenges `simple_verify` methods ([HTTP](https://github.com/certbot/certbot/blob/master/acme/acme/challenges.py#L293) and [TLS-SNI](https://github.com/certbot/certbot/blob/master/acme/acme/challenges.py#L464) respectively) so they shouldn't be affected by this PR.

Issue #2189, opened 2016-01-15
Run test farm tests automatically


bmw:
We're getting some decent test farm tests written, but it does us no good if they're not run. We should set up a system to run these tests regularly off of `master` (every 24 hours?) and e-mail the results (or regressions) to the client dev team.

pde:
Hey look, we already have this ticket!

pde:
Maybe we also want to do #2427

bmw:
Hoping to resolve this before the next release.

bmw:
I always remember talking about how we can't do this because the tests are too flakey, but this isn't a big deal for our nightly tests so maybe we should finally set this up.

bmw:
This was done in https://github.com/certbot/certbot/pull/7072.

PR #4089, opened 2017-01-20
Cleanup test farm tests and add test_sdists

### LINKED ISSUE #4050: Prevent regressions of problems like #4044
The cause of #4044 is that although `setUpClass` wasn't added until Python 2.7, [nose manually runs this method for us](https://github.com/nose-devs/nose/blob/master/nose/suite.py#L308). We should make sure our tests catch problems like this in the future.

Options:

1. Stop using `nose` and rely either on standard Python scripts for running tests or use another program to run our tests.
2. Use another test (perhaps like my hacky script [here](https://github.com/certbot/certbot/pull/4043#issuecomment-272350209)) and make sure it runs regularly on a Python 2.6 system.
3. ???

bmw:
Fixes #4050 (at least once #2189 is also resolved).


pde:
Reasons for deleting the OCSP test farm test rather than getting it working or at least leaving it around for future repair?

pde:
Otherwise LGTM.

bmw:
> Reasons for deleting the OCSP test farm test rather than getting it working or at least leaving it around for future repair?

The test is still around in `tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh`. If you remember, we just wrote `test_ocsp_experimental.sh ` because the feature wasn't testable using `certbot-auto` until we released 0.10.0.

pde:
Ah yes.

PR #4091, opened 2017-01-20
Fix modification-check.sh

bmw:
The version of `modification-check.sh` in `master` modifies `letsencrypt-auto-source/letsencrypt-auto` by running `build.py`. Since Travis reruns our tests up to three times if they fail, if `build.py` needs to be run, `modification-check.sh` will fail the first time and succeed afterwards.

This fixes this problem so changes are properly attributed and prevents confusion for newer contributors when `letsencrypt-auto-source/letsencrypt-auto` suddenly has changes they didn't make.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -560,7 +560,7 @@ BootstrapMageiaCommon() {
       exit 1
   fi
 
-  if ! $SUDO urpmi --force \
+  if ! $SUDO urpmi --force $QUIET_FLAG \
       git \
       gcc \
       python-augeas \
--- a/tests/modification-check.sh
+++ b/tests/modification-check.sh
@@ -61,7 +61,7 @@ else
 build.py."
 fi
 
-rm -rf temp_dir
+rm -rf $temp_dir
 
 if $FLAG ; then
 	exit 1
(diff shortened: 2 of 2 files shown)


tests/modification-check.sh
```diff
@@ -60,8 +61,8 @@ else
 build.py."
 fi
 
+rm -rf temp_dir
```
nickfong: Shouldn't this be `$temp_dir`?

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -560,7 +560,7 @@ BootstrapMageiaCommon() {
       exit 1
   fi
 
-  if ! $SUDO urpmi --force \
+  if ! $SUDO urpmi --force $QUIET_FLAG \
```
nickfong: I was just about to open a PR to address this.  Thanks!

tests/modification-check.sh
bmw: It definitely should! Thanks @nickfong.

letsencrypt-auto-source/letsencrypt-auto
bmw: You still can if you want so you get credit for it in our git history, but I don't think it's a big deal either way.

nickfong: This doesn't really matter to me either.  Not having to open another PR saves me some typing and clicking a few more buttons :smile:

Issue #4016, opened 2017-01-11
Document lineages in our user guide


bmw:
We used this concept a lot when developing Certbot and it's often a useful abstraction for thinking about certificates. We've never properly explained this term, however, so we can't really use it in our documentation.

We may or may not want to explain how lineages work to the average user, but at the very least they should be documented somewhere.

pconrad-fb:
Looking at this from a documentation standpoint, I think it might actually be useful to add a "What Are Certificates?" section. This would accomplish two things:

- It would avoid having to jam a "Lineages" section into some other random place
- It would provide a place for the naive user to get started

We could link to a Lineages subsection in the What Are Certificates section from a few places in the doc, and then we're free to use the term "lineages" anywhere we want without worry.

If this proposal is too heavyweight, then we could add a short "Certificates and Lineages" section to either the Introduction or "Getting Certificates" in the User Guide. I'm okay with any of these options, so please pick whichever one you like.

SwartzCr:
+1 to a "What are Certificates" section, with a lineages sub section
+1 to short thing in the introduction. It could say something along the lines of "What the client is doing is generating a ssl certificate for you. To find out more about what those are and how Certbot keeps them up to date, look here." Or something along those lines.
Looking forward to seeing a draft!

SwartzCr:
just so I don't lose it here's seth explaining part of it: https://community.letsencrypt.org/t/link-point-to-wrong-certs/28441

SwartzCr:
@pconrad-fb any progress?

pconrad-fb:
I haven't jumped into lineages. I expect to do that this Wednesday or Thursday. In the mean time, "Certificates" is stalled. Can I get more directive feedback on this thread: https://github.com/certbot/certbot/pull/4370#pullrequestreview-27701860

SwartzCr:
This should likely be closed once #4948 gets merged

PR #4092, opened 2017-01-20
Add information about cert management to the docs

ohemorange:


docs/using.rst
```diff
@@ -534,8 +534,90 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not
...
+replace that set entirely::
+
+  certbot certonly --cert-name example.com -d example.org,www.example.org
+
+If a certificate is requested with ``run`` or ``certonly`` with a name that does not already
+exist, the new certificate created will be assigned the name specified.
+
+If your account key has been compromised or you otherwise need to revoke a certificate,
+use the revoke command to do so. Note that the revoke command is passed the certificate path
+(ending in ``cert.pem``), not a certificate name or domain. Additionally, if a certificate
```
pde: This is one of those places where explaining lineages would really help people. After all, if you revoke a cert but don't delete it, your renewal cron job will come along later and reissue it for you (!).

docs/using.rst
```diff
@@ -534,8 +534,90 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not
...
+Once a certificate is revoked (or for other cert management tasks), all of a certificate's
+relevant files can be removed from the system with the ``delete`` subcommand::
+
+  certbot delete --cert-name example.com
+
+For advanced certificate management tasks, it is possible to manually modify the certificate's
+renewal configuration file, located at ``/etc/letsencrypt/renewal/CERTNAME``.
+
+.. warning:: Modifying any files in ``/etc/letsencrypt`` can make it so Certbot can no longer
+   properly manage its certificates, and we do not recommend doing so for most users.
```
pde: Perhaps add: "For most tasks, it is safest to limit yourself to pointing symlinks at the files there, or using `--renew-hook` to copy / make new files based upon those files, if your operational situation requires it (for instance, combining certs and keys in different way, or having copies of things with different specific permissions that are demanded by other programs)."

pde: (or something like that)

docs/using.rst
ohemorange: Great point! @pconrad-fb is taking a pass on this; I can make these changes now or block on that, whichever.

pde: I don't have a strong opinion on the documentation side; I suspect that we should actually add a little polish to this in the UI though (either offering to delete the cert, or telling the user that unless they delete the cert it will be renewed at the next renewal event).

pde: I made #4108 to track that UX issue.

pde:
Feel free to either merge as is or make the changes I suggested and then merge.

docs/using.rst
pconrad-fb: I've got using.rst from the Managing Docs branch open locally right now. I can talk about lineages in this pass through the doc—but I don't know what lineages are. Where can I learn about them?

ohemorange: @pconrad-fb unfortunately, #4016.

Quickly: a lineage is what is elsewhere just called a certificate. We use the term internally to create the illusion of consistence across renewals. There is not really any such thing as renewing a cert, to a CA, only getting a new one that might have related/ the same domains in it. Within Certbot, each lineage is attached to a single renewal configuration file, and its related information is stored in the archive and live directories. Then the links in the live directories point to the updated file in the appropriate archive directory whenever you renew.

pconrad-fb: So, to close the loop with @pde, the takeaway is that Certbot makes it a no-brainer to get and maintain certificates; and therefore it's useful to think of them as "lineages" of certificates that have the same domain; and that if you want to end a lineage, so to speak, you really have to explicitly delete them and not just revoke them. Right?

ohemorange: Mostly! They don't always necessarily have the same domain (you could always include more domains on a certificate with `--expand`, and now you can change them entirely by using `--cert-name`) although they often do have that in common. The last half is definitely true and important to note!

ohemorange:
@SwartzCr, @pde - moving back to this PR from #4137. This has the updated version of my changes, @pconrad-fb's changes, and incorporated comments.

Issue #3791, opened 2016-11-13
Certbot and --apache modules looks for vhost conf files in sites-available causing missconfigurations in edge cases


ophiocus:
This will cause the configuration to be written to the wrong conf file 

>>>if more than one vhost file has the same servername specified<<<

repro steps

- in apache2 create 2 vhosts*80 files in sites-available, give theme the same server name, enable the one that is alpahabetiacally last ( Avhost.conf and Bvhost.conf, enable Bvhost)

- run certbot; (sudo letsencrypt --apache -d commonname.com -d www.commonname.com ) 

configurations will be written to the first conf, which is dissabled and the resulting -le-ssl conf file in sites enabled is name based off the dissabled configuration (Avhost.conf has redirect rules, and Avhost-le-ssl.conf is now created as a new vhost and enabled as a new site )

Potential fix-

only look for vhosts conf files present in sites-enabled when using the apache plugin unless otherwise specified

NOTE: i only noticed this because the force secure connection was not working, this led me to find the configuration for the redirect in Avhost which is a dissabled vhost

pde:
@joohoi, thoughts?

joohoi:
Thanks for the report @ophiocus  . It boils down to the question: do we want to search and suggest vhosts in currently active configuration or all possible in the system. 

I think I might agree with @ophiocus . People are prepared to see only the vhosts they have active listed by Certbot, and expect the configuration to be the one that's actually active. I think it's pretty rare to hit this bug however.

The fix itself is merely a OS constants fix.

joohoi:
We discussed this issue, and agreed to modify the process so that we'll search vhosts from sites-enabled explicitly, with a possibility to override the default directory using a cli parameter. This will require some refactoring of the code, but we'll track the process with this issue#

jenlampton:
I'm not using a version of apache that has the `sites-available` directory, and all my virtual hosts are defined in the bottom of my `httpd.conf` file.  I often add a duplicate virtual host entry at the very bottom to catch redirect any errant traffic that didn't match any of the defined virtual hosts, and 301 redirect it back to my primary site, like so:

```
# Redirect all other subdomains to www.
#<VirtualHost *:8080>
#    ServerName www.example.com
#    ServerAlias *.example.com
#    Redirect 301 / http://www.example.com/
#</VirtualHost>\
```

I've been having trouble with the Certbot being able to identify my primary domain, and after seeing this comment: 

> if more than one vhost file has the same servername specified

[...]

Issue #3545, opened 2016-09-27
apache plugin parses files from /etc/apache2/sites-available


anarcat:
sample certbot run here:

```
# certbot run --apache

Which names would you like to activate HTTPS for?
-------------------------------------------------------------------------------
1: foobar.example.com        
[...]
```

that domain is not actually enabled in my apache configuration: it is not symlinked to the `sites-enabled` directory:

```
root@marcos:/etc/letsencrypt# grep -R foobar.example.com /etc/apache2
/etc/apache2/sites-available/ikisite-foobar.example.com:        ServerName foobar.example.com:80
```

therefore it should not be detected by the apache plugin.

this is similar to #2949, except that one was treated as a problem parsing files with commas, whereas there is the larger problems of disabled configs here.

bmw:
This was a design decision made when writing the plugin. Currently, in addition to parsing any files included by Apache, we also parse the `vhost_root` as it is defined in [this file](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/constants.py) for your particular OS. This decision was explained a bit [here](https://github.com/certbot/certbot/issues/2949#issuecomment-218992638), but just because your virtual host isn't enabled doesn't mean that Certbot can't help you set up HTTPS for that name.

anarcat:
i see. well, in my case it means way more domains than I expected to see, and it is really confusing because there's pretty old stuff in there. granted, I should cleanup that directory, but it seems to me that, at least, it should be made clear some directories are disabled.

i am also wondering how the apache plugin will configure LE on those vhosts if they are not configured. will they be temporarily enabled?

maybe i misunderstand how the plugin works.

thanks for the prompt response!

bmw:
> it should be made clear some directories are disabled.

Certbot should understand that the virtual host is disabled and enable it if necessary. More on that below.

> i am also wondering how the apache plugin will configure LE on those vhosts if they are not configured. will they be temporarily enabled?

It depends. If you have a disabled SSL vhost in `sites-available` and you select the domain name found in that vhost, Certbot will enable it and install certificates to the vhost. If you select a disabled HTTP vhost, Certbot will make a copy of the vhost, replace port 80 with port 443, install any certificates, and enable the vhost. The old port 80 vhost will be enabled if you ask for a redirect from HTTP to HTTPS.

anarcat:
On 2016-09-27 20:41:57, Brad Warren wrote:
>> it should be made clear some directories are disabled.
>
[...]

PR #4104, opened 2017-01-24
Do not parse disabled configuration files from under sites-available on Debian / Ubuntu

joohoi:
This changes the apache plugin behaviour to only parse enabled configuration files, ~while still writing new files to sites-available~, and respecting the --apache-vhost-root CLI parameter for new SSL vhost creation. If --apache-vhost-root isn't defined, or doesn't exist, the SSL vhost will be created to originating non-SSL vhost directory.

This PR also implements actual check for vhost enabled state, and makes sure parser.parse_file() does not discard changes in Augeas DOM, by doing an autosave.

Also handles enabling the new SSL vhost, if it's on a path that's not parsed by Apache.

Fixes: #1328 
Fixes: #3545
Fixes: #3791
Fixes: #4523
Fixes: #4837
Fixes: #4905


bmw:
I ran Apache test farm tests on this (with the change from #4126 merged in) and everything passed. If no one beats me to it, I'm going to try and review this first thing tomorrow morning.

certbot-apache/certbot_apache/parser.py
```diff
@@ -60,9 +60,12 @@ def __init__(self, aug, root, vhostroot, version=(2, 4)):
         # Set up rest of locations
         self.loc.update(self._set_locations())
 
-        # Must also attempt to parse virtual host root
-        self._parse_file(self.vhostroot + "/" +
-                         constants.os_constant("vhost_files"))
+        # Must also attempt to parse virtual host root, however
+        # omit sites-available for debian / ubuntu
+        if (not self.vhostroot.endswith("sites-available") and
+                not constants.os_constant("handle_sites")):
```
bmw: Should we be calling `parse_file()` here at all anymore? If I understand correctly, Augeas already automatically parses any files included in an Apache configuration. With this in mind, why do we need to parse anything here manually?

If we need this call for some reason, is the logic here correct? Assuming `constants.py` doesn't change, we now don't parse the vhost directory if it ends in `*sites-available` or we're on Debian. What if a user has modified their main Apache config file to include `sites-available/*`? Should we be parsing the file then?

bmw:
Another thing I forgot to take a look at during my review is whether calling `parse_file()` causes us to lose any changes due to hercules-team/python-augeas#15. `parse_file()` conditionally calls `match` meaning that any unsaved changes in Augeas may be lost. Did you look to see if this is a potential problem here?

If it's OK with you, I'd like to leave this out of 0.11.0 and merge it for 0.12.0. I'm not sure there's any problems here, but I have some questions and want to make sure this behaves as we expect.

joohoi:
is_site_enabled method has been removed, as well as the parse_file call for constants.vhost_root. Would love a new test farm run.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -822,6 +810,8 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
         # We know the length is one because of the assertion above
         # Create the Vhost object
         ssl_vhost = self._create_vhost(vh_p)
+        if self.conf("handle-sites"):
+            ssl_vhost.enabled = False
```
bmw: Is this the right logic? I think it works in the conventional case, but what if someone has a file in `sites-enabled` rather than a symlink?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -833,10 +823,14 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
         # This is for compliance with versions of Apache < 2.4
         self._add_name_vhost_if_necessary(ssl_vhost)
 
+
         return ssl_vhost
 
     def _get_ssl_vhost_path(self, non_ssl_vh_fp):
         # Get filepath of new ssl_vhost
+        # Make sure we use the realpath (eg. sites-available instead of
+        # sites-enabled
```
bmw: nit: missing closing parenthesis

certbot-apache/certbot_apache/parser.py
```diff
@@ -60,10 +60,6 @@ def __init__(self, aug, root, vhostroot, version=(2, 4)):
         # Set up rest of locations
         self.loc.update(self._set_locations())
 
-        # Must also attempt to parse virtual host root
-        self._parse_file(self.vhostroot + "/" +
-                         constants.os_constant("vhost_files"))
```
bmw: When we talked about this on the call I think we expressed some concerns about whether or not we can safely do this. In the conventional case, this works fine but what about people who have been manually setting `--apache-vhost-root`? These people may not exist, but this might break things for them correct?

certbot-apache/certbot_apache/configurator.py
joohoi: This is always newly generated vhost, and this is the only situation a vhost can actually be in disabled state in the current vhost object lifecycle (we "handle-sites" only for debian based distros). 

However maybe we should instead do the handling logic in vhost object init and populate this value instead. Currently, vhosts are always created with enabled = True.

certbot-apache/certbot_apache/parser.py
joohoi: You are right, I brought it back with a condition that we're running a distro with no "handle-sites" (anything but Debian / Ubuntu basically), or have defined a path other than the one in constants. Effectively letting users define the value, but ignoring sites-available.

certbot-apache/certbot_apache/configurator.py
bmw: Yeah I agree this is the only case the vhost can be in an enabled state, but I think it's possible it's already enabled. While it's maybe a bit of a silly example, I was able to get this PR to fail because it tries to enable an already enabled vhost. Starting with the default Apache2 config on Ubuntu 16.04, the changes I made were:
```
sudo a2dissite 000-default.conf
sudo cp /etc/apache2/sites-available/000-default.conf /etc/apache2/sites-enabled
sudo rm /etc/apache2/sites-available/*
```
In the end, your only vhost is a real file in `sites-enabled` which when we run Certbot it creates a vhost in `sites-enabled` which is marked as not enabled.

certbot-apache/certbot_apache/parser.py
```diff
@@ -28,7 +28,7 @@ class ApacheParser(object):
     arg_var_interpreter = re.compile(r"\$\{[^ \}]*}")
     fnmatch_chars = set(["*", "?", "\\", "[", "]"])
 
-    def __init__(self, aug, root, vhostroot, version=(2, 4)):
+    def __init__(self, aug, root, vhostroot, version=(2, 4), extra_path=False):
```
bmw: nit: Since `vhostroot` is only used now if `extra_path` is `True`, can we just make `vhostroot` an optional parameter and remove `extra_path`?

joohoi: Both of these are now fixed, I also added test coverage for the non-symlinked files in sites-enabled.

joohoi:
The requested changes are now in. Due to the large merge and modifications, I'd like to see test farm test rerun.

bmw:
Test farm tests pass.

joohoi:
> So I just wanted to check in on a higher level what our plan is here. This PR breaks some use cases. People could have automated deployments with certbot-apache on Debian/Ubuntu that were relying on us to enable their vhosts. Are we fine with breaking this use case?

I think so, having such deployment flow seems a bit of an edge case that I haven't stumbled upon. The common way is to have a vanilla configuration and have Certbot to copy over the default vhost and enable HTTPS for it. After all, having Certbot parse over all files in sites-available seems to cause more issues than it would solve.

> If we are, perhaps we should go a little farther. For example, do we want to enable sites at all any more? If not, do want to continue to support vhost-root? If your vhost is included in your Apache config we should find it and if it's not I'm not sure how much we can do with it.

We can do these steps after we have a reliable way to step in the relevant IfDefine statements. Some distributions have rather exotic ways and locations to define the Defines so we aren't there yet, and need to provide users a way to specify a path to use if we aren't able to find it.

> These changes certainly don't have to happen in this PR, but I just wanted a little more context for what we actually want to support here especially when all of these values are configurable (e.g. --apache-handle-sites nonempty-string).

Regardless of what we're doing here, I think we should keep using the handle-sites option to comply with the strategy a distribution has agreed upon (ie. creating the actual file under sites-available, and enabling it by symlinking to sites-enabled vs. having a static directory for the actual files in others)

bmw:
> Regardless of what we're doing here, I think we should keep using the handle-sites option to comply with the strategy a distribution has agreed upon (ie. creating the actual file under sites-available, and enabling it by symlinking to sites-enabled vs. having a static directory for the actual files in others).

What if instead we used logic like you added to `_get_ssl_vhost_path` of using `os.path.realpath` and symlinked the file as necessary? It seems like this would allow us to support whatever convention is used, regardless of the convention normally used by the underlying distro.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -577,17 +581,12 @@ def _create_vhost(self, path):
         if filename is None:
             return None
 
-        if self.conf("handle-sites"):
-            is_enabled = self.is_site_enabled(filename)
```
joohoi: This is now addressed along with the fix for creating vhosts in a non-parsed directory (eg. original vhost in Apache root configuration). 

However we'll prefer vhost-root setting for newly created SSL vhosts from now on if the path exists, enabling (or adding Include in root config) if needed. This should make it usable in even more exotic configurations.

joohoi:
This is ready for a review. I changed the PR description to match the current status as well.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1707,21 +1723,32 @@ def enable_site(self, vhost):
...
+        except OSError as err:
+            if os.path.islink(enabled_path) and os.path.realpath(
+               enabled_path) == vhost.filep:
+                # Already in shape
+                vhost.enabled = True
+                return
+            else:
+                logger.warning(
+                    "Could not symlink %s to %s, got error: %s", enabled_path,
+                    vhost.filep, err.strerror)
```
bmw: I never knew about `strerror`.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -972,6 +1005,10 @@ def _copy_create_ssl_vhost_skeleton(self, vhost, ssl_fp):
                 # The content does not include the closing tag, so add it
                 new_file.write("</VirtualHost>\n")
                 new_file.write("</IfModule>\n")
+            # Add new file to augeas paths if we're supposed to handle
+            # activation (it's not included as default)
+            if self.conf("handle-sites"):
+                self.parser.parse_file(ssl_fp)
```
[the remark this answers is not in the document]
joohoi: Before we parsed sites-available contents (added manually to augeas parse tree using value in constants.py). It's not included until the vhost is enabled by creating a symlink from sites-enabled. We need this temporarily to find the newly created vhost in the certificate configuration step.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -857,8 +854,21 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
         vh_p = self._get_new_vh_path(orig_matches, new_matches)
 
         if not vh_p:
-            raise errors.PluginError(
-                "Could not reverse map the HTTPS VirtualHost to the original")
+            # The vhost was not found on the currently parsed paths
+            if not self.conf("handle-sites"):
```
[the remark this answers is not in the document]
joohoi: If handle-sites is true, we're handling the include using distribution specific manner (eg, enabling the site) in deploy_cert method instead.

certbot-apache/certbot_apache/parser.py
```diff
@@ -46,9 +46,7 @@ def __init__(self, aug, root, vhostroot, version=(2, 4)):
         # Find configuration root and make sure augeas can parse it.
         self.root = os.path.abspath(root)
         self.loc = {"root": self._find_config_root()}
-        self._parse_file(self.loc["root"])
-
-        self.vhostroot = os.path.abspath(vhostroot)
```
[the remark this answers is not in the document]
joohoi: You are right, absolute path will be used.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -581,20 +586,34 @@ def _create_vhost(self, path):
...
+                parser.case_i("Include"), inc_path)) == 0:
+            logger.debug("Adding Include %s to %s",
+                         inc_path, parser.get_aug_path(main_config))
+            self.parser.add_dir(
+                parser.get_aug_path(main_config),
+                "Include", inc_path)
+            # Add the new file to augeas DOM
+            # self.parser.parse_file(inc_path)
+            # Reload augeas
+            # self.aug.load()
```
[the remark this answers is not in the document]
joohoi: Yeah, removed the lines altogether

certbot-apache/certbot_apache/configurator.py
```diff
@@ -625,28 +644,6 @@ def get_virtual_hosts(self):
                     file_paths[realpath] = new_vhost.filep
                     internal_paths[realpath].add(internal_path)
                     vhs.append(new_vhost)
-                elif (realpath == new_vhost.filep and
-                      realpath != file_paths[realpath]):
```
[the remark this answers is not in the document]
joohoi: You are right, the removal has been undone.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -903,11 +918,29 @@ def _get_new_vh_path(self, orig_matches, new_matches):
...
-        if non_ssl_vh_fp.endswith(".conf"):
-            return non_ssl_vh_fp[:-(len(".conf"))] + self.conf("le_vhost_ext")
+        """ Get a file path for SSL vhost, uses user defined path as priority,
+        but if the value is invalid or not defined, will fall back to non-ssl
+        vhost filepath.
+
+        :param str non_ssl_vh_fp: File path of non-SSL vhost
+
+        :returns: File path for SSL vhost
+        :rtype: string
```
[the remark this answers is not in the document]
joohoi: Fixed

certbot-apache/certbot_apache/configurator.py
```diff
@@ -903,11 +918,29 @@ def _get_new_vh_path(self, orig_matches, new_matches):
...
+        """
+
+        if self.conf("vhost-root") and os.path.exists(self.conf("vhost-root")):
+            # Defined by user on CLI
+
+            fp = (os.path.realpath(self.conf("vhost-root")) +
+                  "/" + os.path.basename(non_ssl_vh_fp))
```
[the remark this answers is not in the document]
joohoi: Fixed

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1612,7 +1649,7 @@ def _write_out_redirect(self, ssl_vhost, text):
             if len(ssl_vhost.name) < (255 - (len(redirect_filename) + 1)):
                 redirect_filename = "le-redirect-%s.conf" % ssl_vhost.name
 
-        redirect_filepath = os.path.join(self.conf("vhost-root"),
+        redirect_filepath = os.path.join(self.vhostroot,
```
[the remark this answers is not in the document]
joohoi: Yeah, agreed.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1623,6 +1660,12 @@ def _write_out_redirect(self, ssl_vhost, text):
         # Write out file
         with open(redirect_filepath, "w") as redirect_file:
             redirect_file.write(text)
+
+        # Add new file to augeas paths if we're supposed to handle
+        # activation (it's not included as default)
+        if self.conf("handle-sites"):
+            self.parser.parse_file(redirect_filepath)
```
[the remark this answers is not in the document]
joohoi: We handle the actual include later in the process.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -875,6 +885,10 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
         # Create the Vhost object
         ssl_vhost = self._create_vhost(vh_p)
         ssl_vhost.ancestor = nonssl_vhost
+
+        if self.conf("handle-sites") and "/sites-enabled/" not in ssl_vhost.filep:
+            # Set vhost as disabled for enabling later on
+            ssl_vhost.enabled = False
```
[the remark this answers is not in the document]
joohoi: This will be addressed in the distro specific reliability work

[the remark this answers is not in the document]
bmw: I think we may want to address it now. On Debian based systems, we're incorrectly marking vhosts as enabled which can cause problems later in execution.

For example, while it's not trivial to repro, starting with the default Apache config on Ubuntu 16.04, do the following:
```
sudo a2dissite 000-default.conf
sudo mkdir /etc/apache2/sites-available2
sudo cp /etc/apache2/sites-available/000-default.conf /etc/apache2/sites-available2/
# Add a ServerName directive to /etc/apache2/sites-available2/000-default.conf that you can obtain a cert for
sudo certbot -d foo.science --apache --no-redirect --apache-vhost-root /etc/apache2/sites-available2/
# Add a ServerAlias directive to /etc/apache2/sites-available2/000-default.conf
sudo certbot -d foo.science --apache --redirect --apache-vhost-root /etc/apache2/sites-available2/
```
On this branch, this last command fails because it things the disabled vhost in `/etc/apache2/sites-available2` would conflict with a redirect vhost. If you checkout `master`, this problem goes away.

certbot-apache/certbot_apache/configurator.py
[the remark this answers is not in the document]
bmw: Calls to `parse_file` causing changes to be lost should certainly be addressed, however, if it is, perhaps we should just call this unconditionally here, especially if the logic in `make_vhost_ssl` I commented on is changed. Rather than parsing the file in two locations depending on whether or not `sites-enabled` is set, does it hurt anything to always ensure Augeas has parsed the file?

certbot-apache/certbot_apache/configurator.py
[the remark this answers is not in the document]
bmw: Similar to my other comment, if problems with `parse_file` (and specifically `match`) can be addressed, perhaps we should just do this unconditionally?

joohoi:
This PR contains a lot of things I'm building on, and it's a nice overall cleanup of the configurator, hence: Ready for yet another round of review :)

I cleaned up the AugeasConfigurator.save() method and implemented a check for unsaved changes and autosave functionality when doing destructive operations in Augeas DOM tree. I'm not very happy about the way that I had to implement it (passing the ApacheConfigurator instance reference to parser), but I toyed around with options at hand and this seemed like the only way to do this without duplicating too much code. This is result of how the augeas_configurator, configurator and parser relate to each other, it works and seems pretty robust whatsoever.

I also implemented an actual check for vhost enabled state in the way we discussed on the last call, so it should reflect the actual state now.

All the requested issues should have been solved.

bmw:
I'm planning to hold off on reviewing this until the problem causing test failures is fixed, but let me know if you'd like me to review it sooner.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1660,31 +1722,37 @@ def _get_proposed_addrs(self, vhost, port="80"):
 
         return redirects
 
-    def is_site_enabled(self, avail_fp):
-        """Checks to see if the given site is enabled.
-
-        .. todo:: fix hardcoded sites-enabled, check os.path.samefile
+    def parsed_in_current(self, filep):
```
bmw: Feel free to ignore this comment if you disagree, but I think these three methods make more sense in `parser.py`.

certbot-apache/certbot_apache/parser.py
```diff
@@ -62,9 +67,12 @@ def __init__(self, aug, root, vhostroot, version=(2, 4)):
         # Set up rest of locations
         self.loc.update(self._set_locations())
 
-        # Must also attempt to parse virtual host root
-        self._parse_file(self.vhostroot + "/" +
-                         constants.os_constant("vhost_files"))
+        self.existing_paths = copy.deepcopy(self.parser_paths)
```
bmw: One thing that worries me a bit is this is never updated if an `Include` directive is added later. What if we moved the `add_include` function into the parser and keep this value up to date?

certbot-apache/certbot_apache/configurator.py
joohoi: I do agree. These are now moved to parser.py

certbot-apache/certbot_apache/parser.py
joohoi: Nice catch! This functionality is now added to the PR

joohoi:
Suggested improvements have been done, good suggestions regarding code health, thanks!

While adding new include paths to the `parser.existing_paths`-variable I stumbled upon a bug in site enabling functionality: when `vhost.enabled` was `False` in the later check if a enable_site should be run (for Debian / Ubuntu symlinking), the new dynamic check for if the vhost was enabled was correctly reporting `True`, hence symlink was not added. I moved the add_include call to enable_site function, so it gets run instead of the symlinking if needed.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1705,21 +1721,39 @@ def enable_site(self, vhost):
...
         """
-        if self.is_site_enabled(vhost.filep):
+        if vhost.enabled:
+            return
+
+        # Handle non-debian systems
+        if not self.conf("handle-sites"):
+            if not self.parser.parsed_in_original(vhost.filep):
+                # Add direct include to root conf
+                self.parser.add_include(self.parser.loc["default"], vhost.filep)
```
bmw: We should probably set `vhost.enabled` to `True` after this to make sure the value stays up to date.

joohoi: Oh, of course. Thanks!

PR #4113, opened 2017-01-25
Automatically retry after getting badNonce error

### LINKED ISSUE #2244: Should auto-recover from bad nonces
I recently got this error trying to issue a cert:

`The client sent an unacceptable anti-replay nonce :: JWS has invalid anti-replay nonce`

This happens from time to time - the server loses track of outstanding nonces. Since the reply always includes a fresh nonce, the client should auto-retry the request.

### LINKED ISSUE #4099: The client sent an unacceptable anti-replay nonce :: JWS has invalid anti-replay nonce
Hey guys, so i am trying to renew some ssl certs that will expire in about 2 weeks but i keep getting the stuff below. i googled and i saw that you have to retry but i did it about 5 times today and i am still getting this. i was using certbot yum rpm on centos7 , then i replaced with the version from git. still the same thing.

> An unexpected error occurred:
> The client sent an unacceptable anti-replay nonce :: JWS has invalid anti-replay nonce tEHB789Ajg4v2yGpSPSIwwrjp4viAWY7k4zjjJwQtGo
> Please see the logfiles in /var/log/letsencrypt for more details.
> 

any ideas on what can i do ? someone said that he got it to work by sending the same nonce but i didn't find any info on how to do that.

thanks in advance.

bmw:
Fixes #2244 and fixes #4099.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -676,8 +676,16 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
 
     def post(self, url, obj, content_type=JOSE_CONTENT_TYPE, **kwargs):
         """POST object wrapped in `.JWS` and check response."""
-        data = self._wrap_in_jws(obj, self._get_nonce(url))
-        kwargs.setdefault('headers', {'Content-Type': content_type})
-        response = self._send_request('POST', url, data=data, **kwargs)
-        self._add_nonce(response)
-        return self._check_response(response, content_type=content_type)
+        MAX_ATTEMPTS = 3
+        for attempt in range(MAX_ATTEMPTS+1):
+            try:
+                data = self._wrap_in_jws(obj, self._get_nonce(url))
+                kwargs.setdefault('headers', {'Content-Type': content_type})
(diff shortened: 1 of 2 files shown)


bmw:
This should be merged not squashed to give credit to the patch from #2244.

acme/acme/client.py
```diff
@@ -674,8 +674,25 @@ def _get_nonce(self, url):
...
+
+        """
+        should_retry = True
+        while True:
+            try:
+                return self._post_once(*args, **kwargs)
+            except messages.Error as error:
+                if should_retry and error.code == 'badNonce':
+                    logger.debug('Retrying request after error:\n%s', error)
+                    should_retry = False
```
ohemorange: if we're only trying once, we cooould just put `self._post_once` right in the except. this should be fine though.

ynasser:
lgtm!

bmw:
Awesome. Thanks for the review guys. I agree with you on the just call `self._post_once` twice comment, but after talking to Erica, I'm not going to block the release on this (making the change also required changing tests).

PR #4127, opened 2017-01-27
Refactor main

ohemorange:
Simplify certificate logic, including separating taking action and reporting.

Potential future steps that do not belong in this PR:
- kill _init_le_client
- put _find_cert, _get_and_save_cert, _install_cert somewhere better
- make the reporting less strangely conditional
- (aspirational) kill should_get_cert/reinstall


ohemorange:
I know I'm missing some comments, and I'd like to add some integration tests to make sure there are no regressions, but otherwise this is roughly the amount of refactoring appropriate to include in this PR.

ohemorange:
Timed out exception??

ohemorange:
Maybe tests will be happier today?

certbot/main.py
```diff
@@ -651,34 +661,24 @@ def obtain_cert(config, plugins, lineage=None):
...
-        action = "newcert"
+        _report_new_cert(config, print_fullchain_path=False)
+        _suggest_donation(config)
+        return
 
-    # POSTPRODUCTION: Cleanup, deployment & reporting
-    notify = zope.component.getUtility(interfaces.IDisplay).notification
-    if config.dry_run:
-        _report_successful_dry_run(config)
-    elif config.verb == "renew":
```
bmw: With the logic in this `if` statement removed (and not put in `renew_cert`), we now no longer reload user's webservers upon renewal. If they used the Apache or Nginx plugin, their servers will continue to use the old certificates upon renewal.

certbot/tests/main_test.py
```diff
@@ -114,15 +121,20 @@ def _call(self, args):
...
+    @mock.patch('certbot.main._find_cert')
+    @mock.patch('certbot.main._get_and_save_cert')
+    @mock.patch('certbot.main._report_new_cert')
+    def test_no_reinstall_text_pause(self, unused_report, mock_auth,
+        mock_find_cert):
         mock_notification = self.mock_get_utility().notification
         mock_notification.side_effect = self._assert_no_pause
-        mock_auth.return_value = ('reinstall', mock.ANY)
+        mock_auth.return_value = mock.Mock()
+        mock_find_cert.return_value = None, None
```
bmw: nit: `_find_cert` will never return `None` here. `False` should be used instead.

certbot/tests/main_test.py
```diff
@@ -495,22 +507,23 @@ def test_noninteractive(self):
...
                 self._cli_missing_flag(args, "--agree-tos")
 
+    @mock.patch('certbot.main._report_new_cert')
     @mock.patch('certbot.main.client.acme_client.Client')
     @mock.patch('certbot.main._determine_account')
     @mock.patch('certbot.main.client.Client.obtain_and_enroll_certificate')
-    @mock.patch('certbot.main._auth_from_available')
-    def test_user_agent(self, afa, _obt, det, _client):
+    @mock.patch('certbot.main._get_and_save_cert')
+    def test_user_agent(self, afa, _obt, det, _client, unused_report):
```
bmw: nit: We should probably update the name `afa` so the name is still relevant.

certbot/tests/main_test.py
```diff
@@ -562,13 +576,13 @@ def test_configurator_selection(self, mock_exe_exists):
         self._cli_missing_flag(["--standalone"], "With the standalone plugin, you probably")
 
         with mock.patch("certbot.main._init_le_client") as mock_init:
-            with mock.patch("certbot.main._auth_from_available") as mock_afa:
-                mock_afa.return_value = (mock.MagicMock(), mock.MagicMock())
+            with mock.patch("certbot.main._get_and_save_cert") as mock_afa:
```
bmw: Me in 2018: What's an `afa`?

certbot/tests/main_test.py
```diff
@@ -656,7 +670,7 @@ def test_certonly_abspath(self):
         chain = 'chain'
         fullchain = 'fullchain'
 
-        with mock.patch('certbot.main.obtain_cert') as mock_obtaincert:
+        with mock.patch('certbot.main.certonly') as mock_obtaincert:
```
bmw: This should probably be `mock_certonly`, although I care less than I do about `afa` because this name is much better.

certbot/tests/main_test.py
```diff
@@ -1000,7 +1018,7 @@ def test_renew_obtain_cert_error(self):
             mock_rc.return_value = mock_lineage
             mock_lineage.configuration = {
                 'renewalparams': {'authenticator': 'webroot'}}
-            with mock.patch('certbot.main.obtain_cert') as mock_obtain_cert:
+            with mock.patch('certbot.main.renew_cert') as mock_obtain_cert:
```
bmw: This should probably be `mock_obtain_cert`, although I care less than I do about `afa` because this name is much better.

certbot/main.py
```diff
@@ -48,72 +48,52 @@
 logger = logging.getLogger(__name__)
 
 
-def _suggest_donation_if_appropriate(config, action):
+def _suggest_donation(config):
```
ohemorange: Before it was conditional based on action and verb, now it's only conditional based on "are we a test case". Which is less conditional but still conditional -- I can change it back.

certbot/main.py
```diff
@@ -235,6 +215,13 @@ def _find_lineage_for_domains(config, domains):
     elif subset_names_cert is not None:
         return _handle_subset_cert_request(config, domains, subset_names_cert)
 
+def _find_cert(config, domains, certname):
+    """Returns a tuple of (should_get_cert, lineage or None)"""
```
ohemorange: Ok so I'll definitely mention the type (boolean). Beyond that - "Finds an existing certificate object given domains and/or a certificate name. `should_get_cert` indicates if this function should be followed by a call to fetch a certificate from the server."

certbot/main.py
```diff
@@ -306,16 +293,24 @@ def _find_domains_or_certname(config, installer):
     return domains, certname
 
 
-def _report_new_cert(config, cert_path, fullchain_path):
+def _report_new_cert(config, lineage=None, print_fullchain_path=True):
     """Reports the creation of a new certificate to the user.
```
ohemorange: actually I'm probably going to switch it back to the old version, see other comment.

certbot/main.py
```diff
@@ -632,15 +632,25 @@ def _csr_obtain_cert(config, le_client):
         logger.debug(
             "Dry run: skipping saving certificate to %s", config.cert_path)
     else:
-        cert_path, _, cert_fullchain = le_client.save_certificate(
+        le_client.save_certificate(
```
ohemorange: ack, I thought that logic happened before here. going back to the old _report_new_cert, then.

how are we giving you the path to cert.pem?

certbot/main.py
ohemorange: oversight, thanks - meant to put this in renew_cert.

certbot/main.py
bmw: Good question. After digging, it turns out we aren't. See #4182.

certbot/main.py
ohemorange: Also, pretty sure we don't actually need `print_fullchain_path` or the conditional below, as per https://github.com/certbot/certbot/issues/4182#issuecomment-278158360.

ohemorange:
"The command "sudo -E apt-add-repository -y "ppa:raphink/augeas"" failed and exited with 1 during ."???

certbot/storage.py
```diff
@@ -392,6 +392,26 @@ def __init__(self, config_filename, cli_config, update_symlinks=False):
         self._check_symlinks()
 
     @property
+    def key_path(self):
```
bmw: Now that we're no longer using a configuration and lineage in places interchangeably, are these properties needed? Do we need/want to be able to access the same attribute of a lineage inmultiple ways?

certbot/main.py
```diff
@@ -600,25 +593,29 @@ def run(config, plugins):  # pylint: disable=too-many-branches,too-many-locals
...
+    if lineage is None or not should_get_cert:
         display_ops.success_installation(domains)
     else:
         display_ops.success_renewal(domains)
 
-    _suggest_donation_if_appropriate(config, action)
+    _suggest_donation_if_appropriate(config)
 
 
 def _csr_obtain_cert(config, le_client):
```
bmw: nit: If we're renaming `obtain_cert` to `certonly`, should this be renamed to `_csr_certonly`?

certbot/main.py
```diff
@@ -653,34 +673,26 @@ def obtain_cert(config, plugins, lineage=None):
...
-    if config.csr is None: # the common case
-        domains, certname = _find_domains_or_certname(config, installer)
-        action, _ = _auth_from_available(le_client, config, domains, certname, lineage)
-    else:
-        assert lineage is None, "Did not expect a CSR with a RenewableCert"
-        _csr_obtain_cert(config, le_client)
-        action = "newcert"
+    if config.csr:
+        cert_path, fullchain_path = _csr_obtain_cert(config, le_client)
+        _report_new_cert(config, cert_path, fullchain_path)
```
bmw: nit: I think it'd be better to put these two calls in `_csr_obtain_cert` since it pushes more work into the smaller function, let's `_csr_obtain_cert` handle all `certonly --csr` specific work, and allows us to stop returning `(None, None)` which will blow up if we ever start reporting new certs from a dry run.

If this change is made, I'd make the above line `return _csr_obtain_cert(config, le_client)`.

certbot/main.py
```diff
@@ -653,34 +673,26 @@ def obtain_cert(config, plugins, lineage=None):
...
+
+    if not should_get_cert:
+        notify = zope.component.getUtility(interfaces.IDisplay).notification
         notify("Certificate not yet due for renewal; no action taken.", pause=False)
-    _suggest_donation_if_appropriate(config, action)
+        return
+
+    lineage = _get_and_save_cert(le_client, config, domains, certname, lineage)
 
+    cert_path = lineage.cert_path if lineage else None
```
bmw: Can `lineage` ever be `None` (or `False`)? If so, `_report_new_cert` is going to blow up (unless we're also doing a dry run).

certbot/main.py
```diff
@@ -600,25 +593,29 @@ def run(config, plugins):  # pylint: disable=too-many-branches,too-many-locals
...
+
+    new_lineage = lineage
+    if should_get_cert:
+        new_lineage = _get_and_save_cert(le_client, config, domains,
+            certname, lineage)
 
-    le_client.deploy_certificate(
-        domains, lineage.privkey, lineage.cert,
-        lineage.chain, lineage.fullchain)
+    cert_path = new_lineage.cert_path if new_lineage else None
```
bmw: Can `new_lineage` ever be `None` (or `False`)? If so, `_report_new_cert` is going to blow up (unless we're also doing a dry run).

certbot/storage.py
ohemorange: I hate that they're not named consistently, but you're right, this is dead code and should be removed.

certbot/main.py
ohemorange: `new_lineage` cannot be `False`, because then we would have already errored out. If `new_lineage` is `None`, it means we've done a dry run. `new_lineage` should not be `None` unless we've done a dry run.

certbot/main.py
ohemorange: For both this and renaming `_csr_obtain_cert` to `_csr_certonly`, that depends on how we're conceptualizing `_csr_obtain_cert`. I was thinking of it as analogous to `_get_and_save_cert`, whereas it seems like you're thinking of it as more analogous to `certonly`. If we were going to move these lines into `_csr_obtain_cert`, I'd actually also move it above initializing `le_client` and just fully split the paths immediately. But instead of doing that, maybe we could rename to `_csr_get_and_save_cert`. Would that make it more clear? Or are you still worried about the `None` thing? I can also put a comment in `_report_new_cert` that `cert_path` and `fullchain_path` will be `None` if we're calling from a dry run.

certbot/main.py
ohemorange: same as above - only if we've done a dry run.

ok, how about I change `_report_new_cert` so that it has a comment about "these will be false if we're doing a dry run", and also change it to `if config.dry_run or not cert_path`.

certbot/storage.py
ohemorange: jk

certbot/main.py
ohemorange: adding assertion

certbot/main.py
ohemorange: renaming to _csr_get_and_save_cert

PR #4146, opened 2017-02-01
Upgrading Vagrantfile to Xenial

TheNavigat:


TheNavigat:
Based on this https://github.com/mitchellh/vagrant/issues/7155#issuecomment-228568200, this might not be a very good idea.

We have multiple choices here. We can use bento, we can use Debian 8, or we can move to other choices.

bmw:
We actually completely removed our Vagrantfile in #4211. Thanks for the PR though!

PR #4164, opened 2017-02-04
Make JWKRSA() generate an RSA private key.

jsha:
Currently, to use JWKRSA, you need to import
cryptography.hazmat.backends.openssl.rsa and generate your own key. Ideally acme
clients shouldn't have to touch hazmat. This provides a way for them to avoid
that.

Also, fix a typo: 'asymmetric'

### DIFF
--- a/acme/acme/jose/jwk.py
+++ b/acme/acme/jose/jwk.py
@@ -174,9 +174,10 @@ class JWKOct(JWK):
 class JWKRSA(JWK):
     """RSA JWK.
 
-    :ivar key: `cryptography.hazmat.primitives.rsa.RSAPrivateKey`
+    :ivar key: Optional `cryptography.hazmat.primitives.rsa.RSAPrivateKey`
         or `cryptography.hazmat.primitives.rsa.RSAPublicKey` wrapped
-        in `.ComparableRSAKey`
+        in `.ComparableRSAKey`. If not present, will generate an RSAPrivateKey
+        with suitable defaults.
 
     """
     typ = 'RSA'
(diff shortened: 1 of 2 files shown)


jsha:
The le_auto_trusty build is failing, not sure why.

bmw:
@jsha sorry for the confusion. Travis fails aren't you're fault. It's due to my changes in #4162 from our release not being merged. Once that happens, I'll kick your Travis build.

acme/acme/jose/jwk_test.py
```diff
@@ -114,6 +115,12 @@ def test_init_auto_comparable(self):
             self.jwk256_not_comparable.key, util.ComparableRSAKey))
         self.assertEqual(self.jwk256, self.jwk256_not_comparable)
 
+    def test_init_default_generate(self):
+        from acme.jose.jwk import JWKRSA
+        jwk = JWKRSA()
+        # pylint: disable=protected-access
+        self.assertTrue(isinstance(jwk.key._wrapped, rsa._RSAPrivateKey))
```
alex: You want to check if it's an instance of `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey`, which is the interface; you don't want to check against the private implementation.

acme/acme/jose/jwk.py
```diff
@@ -174,9 +174,10 @@ def public_key(self):
 class JWKRSA(JWK):
     """RSA JWK.
 
-    :ivar key: `cryptography.hazmat.primitives.rsa.RSAPrivateKey`
+    :ivar key: Optional `cryptography.hazmat.primitives.rsa.RSAPrivateKey`
```
bmw: nit: This documentation is for the `key` instance variable of a `JWKRSA` object and it doesn't really make sense to call it optional. To document you don't have to pass a key to the constructor, you should either add documentation to the `__init__` method or remove "Optional" at the beginning and make the behavior clear in the other sentence you added here.

acme/acme/jose/jwk_test.py
```diff
@@ -3,6 +3,7 @@
 import unittest
 
 from acme import test_util
+from cryptography.hazmat.primitives.asymmetric import rsa
```
bmw: nit: According to our style guide, this should be changed to:
```
...
import unittest

from cryptography.hazmat.primitives.asymmetric import rsa

from acme.jose import errors
...
```

jsha:
I'm now starting to have second thoughts about this change: I think it may be at the wrong level of abstraction. Instead, what if we make the `key` argument to `Account` optional, and provide load/dump methods on Account? That way the library can decide for itself what type of key to use, rather than requiring clients to decide. This will allow us to store more data on an account than just the key, if necessary.

jsha:
Closing while I work on the better account-based approach.

PR #4165, opened 2017-02-04
Move make_csr into acme.crypto_util

jsha:
It's relatively finicky to make a CSR appropriate to pass to poll_and_request_issuance. I think most users want to be able to give a list of domains and a private key, and get back a CSR. This branch adds that functionality to crypto_util.

This change also adds a simple method to generate a private key. Right now the function does not accept a key type parameter. Since this is a convenience function I'm fine defaulting to RSA 2048, and adding an optional key type field later.

Note that the two new functions take arguments, and return values, as PEM-encoded buffers. This is a departure from some existing ACME interfaces that take PyOpenSSL types. I've discussed with the Certbot team, and we agree that this is broadly the direction the ACME API should take, so that users of the module don't need to import PyOpenSSL themselves, or use its primitives.


jsha:
Another thought here: make_csr could take a private key in PEM format (as a buffer), since folks have been asking for a long time to be able to easily issue with a specific private key.

acme/acme/crypto_util_test.py
```diff
@@ -151,6 +151,55 @@ def test_sn_collisions(self):
             self.serial_num.append(cert.get_serial_number())
         self.assertTrue(len(set(self.serial_num)) > 1)
 
+class FunctionTest(unittest.TestCase):
```
bmw: If `make_private_key` is removed we can change this class to `TestMakeCSR` to match the convention followed in most of the rest of the repo where a test class tests a single function/class/etc.

acme/acme/crypto_util_test.py
```diff
@@ -151,6 +151,55 @@ def test_sn_collisions(self):
...
+    """Test for standalone functions."""
+
+    def test_make_private_key(self):
+        from acme.crypto_util import make_private_key
+        pem = make_private_key()
+        self.assertTrue(b'--BEGIN PRIVATE KEY--' in pem)
+        self.assertTrue(b'--END PRIVATE KEY--' in pem)
+
+    def test_make_csr(self):
+        from acme.crypto_util import make_private_key, make_csr
```
bmw: Another common testing convention followed in this repo is to import whatever you're testing in the smallest scope possible. The way I'd do this is to make a method like:

```
@classmethod
_call_with_key(cls, *args, **kwargs):
```
that generates a key, imports `make_csr`, and returns its return value.

certbot/tests/crypto_util_test.py
```diff
@@ -299,36 +227,6 @@ def test_common_name_sans_order(self):
...
-            self._call(test_util.load_vector('csr-san.pem'))))
-
-    def test_extract_six_sans(self):
-        self.assertEqual(
-            set(self._call(test_util.load_vector('csr-6sans.pem'))),
-            set(("example.com", "example.org", "example.net",
-                 "example.info", "subdomain.example.com",
-                 "other.subdomain.example.com",)))
-
-    def test_parse_non_csr(self):
```
bmw: Deleting this test removed coverage of the exception handling in `certbot.crypto_util._load_cert_or_req`. It'd be nice if a test was added to `GetNamesFromCertTest` so we don't lose coverage here.

jsha:
Review feedback fixed.

PR #4179, opened 2017-02-06
Do not try to do TLS activation tricks on unsupported systems

### LINKED ISSUE #3364: opensuse Tumbleweed: Each time I run certbot, a line with `Listen 443` gets added to `/etc/apache2/httpd.conf`
On openSUSE Tumbleweed, each time I run certbot, a line with `Listen 443` gets added to `/etc/apache2/httpd.conf`.

See this diff:

```
# git diff /etc/apache2/httpd.conf | more
diff --git a/apache2/httpd.conf b/apache2/httpd.conf
index eb9bd25..11e1962 100644
--- a/apache2/httpd.conf
+++ b/apache2/httpd.conf
@@ -229,4 +229,6 @@ IncludeOptional /etc/apache2/vhosts.d/*.conf
 <IfModule mod_ssl.c>
 Listen 443
 Listen 443
+Listen 443
+Listen 443
 </IfModule>
```

### LINKED ISSUE #3967: Breaking extra "Listen 0.0.0.0:443" created in httpd.conf during renewal in certbot's apache plugin
I've run into this several times now on OpenSUSE 42.1, using apache 2.4. httpd.conf doesn't include `Listen` directives on its own but rather `Include /etc/apache2/listen.conf`.

What happens is that certbot, seemingly unaware of this, ends up adding this at the end of httpd.conf:
```
<IfModule mod_ssl.c>
Listen 0.0.0.0:443
Listen 0.0.0.0:443
</IfModule>
```

Note that it happens twice too (perhaps because I'm renewing a cert for 2 domains). Regardless, this breaks apache because it can't bind to port 443 more than once, so it doesn't start at all.

Sample command used:
`certbot --test-cert --break-my-certs --webroot --installer apache --webroot-path /srv/www/htdocs/site.com -d dev.site.com -d dev2.site.com --post-hook "/usr/sbin/rcapache2 reload; /usr/sbin/rcnginx reload"`

I believe this also happened in much simpler renewal commands too when I was testing, but, interestingly, not always - for example, it didn't happen during the initial certificate installation.

joohoi:
Skip the prepare_server_https() altogether instead of previous behaviour where we just skipped the mod_ssl activation parts, but still tried to add Listen statement. This failed because we couldn't reliably detect if we should jump into IfDefined statements in some of the distros.

Fixes #3967
Fixes #3364

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -233,7 +233,9 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
         # This is done first so that ssl module is enabled and cert_path,
         # cert_key... can all be parsed appropriately
-        self.prepare_server_https("443")
+
+        if self.conf("handle-modules"):
+            self.prepare_server_https("443")
 
         path = {"cert_path": self.parser.find_dir("SSLCertificateFile",
                                                   None, vhost.path),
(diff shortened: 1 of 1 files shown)


joohoi:
With the concerns raised by @bmw I'm starting to think this PR as a lazy man's fix for the underlying problem. I'll close this one out, and look for a good way to actually support as many of the known distributions way of Apache Defines, module installation etc.

PR #4180, opened 2017-02-06
Use latest Boulder for integration tests.

jsha:
I'm open to discussion about whether we want to always run against latest Boulder (with potential flakiness), or just bring the fixed branch up to a more modern commit.


cpu:
@pde Does this seem mergeable to you? I'd like to close https://github.com/letsencrypt/boulder/issues/2030 on our side if it won't cause you folks hardship.

Issue #4143, opened 2017-02-01
The 'ndg-httpsclient' distribution was not found


kierun:
I run the following from cron every month, and got this error:

```
/opt/letsencrypt/letsencrypt-auto certonly --quiet --standalone --renew-by-default -d example.org -d www.example.org  
Creating virtual environment...
Installing Python packages...     
Installation succeeded.               
An unexpected error occurred:             
DistributionNotFound: The 'ndg-httpsclient' distribution was not found and is required by requests                               
Please see the logfile 'certbot.log' for more details.
```

Clearly, "example" is not my domain, I used the correct one. The certbot.log does not offer any more details:

```
Traceback (most recent call last):
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/opt/letsencrypt/certbot/main.py", line 858, in main
    plugins = plugins_disco.PluginsRegistry.find_all()
  File "/opt/letsencrypt/certbot/plugins/disco.py", line 183, in find_all
    plugin_ep = PluginEntryPoint(entry_point)
  File "/opt/letsencrypt/certbot/plugins/disco.py", line 34, in __init__
    self.plugin_cls = entry_point.load()
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2201, in load
    self.require(*args, **kwargs)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2218, in require
    items = working_set.resolve(reqs, env, installer)
  File "/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 830, in resolve
    raise DistributionNotFound(req, requirers)
DistributionNotFound: The 'ndg-httpsclient' distribution was not found and is required by requests
```

What am I missing?

System is Ubuntu 14.04.5 LTS (Trusty Tahr) if that matters and certbot is the lastest master as of today.

bmw:
That's an interesting error. Can you tell me the version of your `letsencrypt-auto` script? Probably the easiest way for you to do this is `grep -m1 LE_AUTO_VERSION /opt/letsencrypt/letsencrypt-auto`. If there's no output from `grep`, you have a very old `letsencrypt-auto` and I encourage you to download a new one from https://dl.eff.org/certbot-auto.

[...]

PR #4185, opened 2017-02-08
Docs for custom Python environment interference.

kierun:
Documenting an edge case where a user's Python environment might interfere with the correct setup stages.  See [this issue report](https://github.com/certbot/certbot/issues/4143) for more details.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -76,6 +76,9 @@ For full command line help, you can type::
 Problems with Python virtual environment
 ----------------------------------------
 
+Low memory system
+^^^^^^^^^^^^^^^^^
+
 On a low memory system such as VPS with only 256MB of RAM, the required dependencies of Certbot will failed to build.
 This can be identified if the pip outputs contains something like ``internal compiler error: Killed (program cc1)``.
 You can workaround this restriction by creating a temporary swapfile::
@@ -89,6 +92,11 @@ Disable and remove the swapfile once the virtual enviroment is constructed::
 
   user@webserver:~$ sudo swapoff /tmp/swapfile
   user@webserver:~$ sudo rm /tmp/swapfile
+
+Custom Python environment
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Note that if you have configured any custom Python environment variables on your system, make sure they are properly (un)set when running `certbot-auto` and `sudo`. This could happen if you run `sudo zsh` as your user's `PYTHONPATH` will be preserved which might cause problems.
 
 Running with Docker
 -------------------


bmw:
@pde, you were looking into code/documentation changes around low memory problems with `certbot-auto` with PRs like #2370 and #3473. Are you still working on this or would you like Noah or I to take over?

jsha:
@kierun can you explain in more detail what sort of problems arose for you with a custom PYTHONPATH? Ideally we should fix or detect those problems in certbot-auto.

bmw:
Whoops. I didn't mean to assign this PR to Peter. Also, the reason for this PR was explained in #4143.

docs/install.rst
```diff
@@ -90,6 +93,11 @@ Disable and remove the swapfile once the virtual enviroment is constructed::
   user@webserver:~$ sudo swapoff /tmp/swapfile
   user@webserver:~$ sudo rm /tmp/swapfile
 
+Custom Python environment
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Note that if you have configured any custom Python environment variables on your system, make sure they are properly (un)set when running `certbot-auto` and `sudo`. This could happen if you run `sudo zsh` as your user's `PYTHONPATH` will be preserved which might cause problems.
```
bmw: It might be nice to link to make "Python environment variables" link to https://docs.python.org/2/using/cmdline.html#environment-variables. You can do this like:

```
...custom `Python environment variables <https://docs.python.org/2/using/cmdline.html#environment-variables>`_ on..
```

Also, please correct me if I'm wrong, but I don't think it's the default behavior of either `sudo` or `zsh` to preserve environment variables like `PYTHONPATH`. If this is the case, I'd remove the second sentence entirely. You can add "(such as `PYTHONPATH`)" to the first sentence if you want to bring more attention to this variable.

jsha: In general descendant processes inherit the environment of their parent. If you run `sudo su`, you get a login shell that clears the environment and loads root's default environment (through the various mechanisms that do so). However, if you run `sudo bash` or `sudo zsh`, most of the environment gets kept (though this may differ depending on sudo permissions).

kierun: @bmw Your changes are fine.  However, you can set up virtual environments to link to all the system wide libraries as well. It is not the default, but it is possible.

kierun:
What happens next?

jsha:
Let's do a little more investigation at https://github.com/certbot/certbot/issues/4143 so we can be sure the warning we put in the docs, if any, is very specific.

PR #4188, opened 2017-02-08
Fix Python installation when using MacPorts

### LINKED ISSUE #3479: OS X MacPorts problem when installing python
OS X 10.11.6 with latest Macports installed.
Cerbot scripts are unable to install python via MacPorts. 

```
echo "Installing python..."
Error: Port python not found
```

This concerns following files:

```
certbot-auto:   
letsencrypt-auto:   
letsencrypt-auto-source/letsencrypt-auto
letsencrypt-auto-source/pieces/bootstrappers/mac.sh
```

This happens because `sudo port install python` returns `Error: Port python not found`. My workaround is to replace `$pkgcmd python` with `$pkgcmd python27`

I am not sure if `sudo port install python` is suposed to work and something is broken on MacPorts side.

MichalMMac:
This pull requrest fixes #3479 where I desribed that curret `BootstrapMac` function in unable to install Python from MacPorts because port name differes from HomeBrew.

Simple `if` condition decides if python is to be installed by MacPorts or HomeBrew.

Homebrew case: `brew install python`
MacPorts case: `sudo port install python27`

### DIFF
--- a/certbot-auto
+++ b/certbot-auto
@@ -518,7 +518,11 @@ BootstrapMac() {
     # We want to avoid using the system Python because it requires root to use pip.
     # python.org, MacPorts or HomeBrew Python installations should all be OK.
     echo "Installing python..."
-    $pkgcmd python
+    if [ "$pkgman" == "brew" ]; then
+      $pkgcmd python
+    elif [ "$pkgman" == "port" ]; then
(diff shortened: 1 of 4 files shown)


MichalMMac:
Sorry. I pushed something by mistake.

PR #4190, opened 2017-02-08
macOS Fix Python installation when using MacPorts

### LINKED ISSUE #3479: OS X MacPorts problem when installing python

MichalMMac:
This pull requrest fixes #3479 where I desribe that current `BootstrapMac` function in unable to install Python from MacPorts because port name differes from HomeBrew's.

Simple if condition decides if python is to be installed by MacPorts or HomeBrew.

Homebrew case: brew install python
MacPorts case: sudo port install python27

### DIFF
--- a/certbot-auto
+++ b/certbot-auto
@@ -518,7 +518,11 @@ BootstrapMac() {
     # We want to avoid using the system Python because it requires root to use pip.
     # python.org, MacPorts or HomeBrew Python installations should all be OK.
     echo "Installing python..."
-    $pkgcmd python
+    if [ "$pkgman" == "brew" ]; then
+      $pkgcmd python
+    elif [ "$pkgman" == "port" ]; then
+      $pkgcmd python27
+    fi
   fi
 
   # Workaround for _dlopen not finding augeas on macOS
--- a/letsencrypt-auto
+++ b/letsencrypt-auto
@@ -518,7 +518,11 @@ BootstrapMac() {
     # We want to avoid using the system Python because it requires root to use pip.
     # python.org, MacPorts or HomeBrew Python installations should all be OK.
     echo "Installing python..."
-    $pkgcmd python
+    if [ "$pkgman" == "brew" ]; then
+      $pkgcmd python
+    elif [ "$pkgman" == "port" ]; then
+      $pkgcmd python27
+    fi
   fi
 
   # Workaround for _dlopen not finding augeas on macOS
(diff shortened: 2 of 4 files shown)


bmw:
Thanks for the PRs @MichalMMac. Unfortunately, the core Certbot team had a design meeting recently where we decided to deprecate and remove support for experimental platforms in `certbot-auto` where there are better options available such as Homebrew or MacPorts (see #4520). `certbot-auto` was simply meant to be a temporary solution until Certbot was properly packaged. This work has certainly be done on macOS and we don't want `certbot-auto` used on the platform because it is much less stable than other options.

We're adding the deprecation messages this release and won't be accepting any additional changes until support is ultimately removed. If things worked for people, great, but we're actively discouraging people from using the platform and don't want to make/maintain any changes that may cause issues for existing users during the transition period. I hope you understand and I'm sorry this PR sat around so long.

If you're interested in doing additional work on Certbot surrounding macOS, we've been trying to get macOS tests to run in Travis without much luck and would certainly appreciate the help (see #3035).

MichalMMac:
Oh well time to look into `acme.sh` I guess. Not sure why I missed certbot in MacPort repo.

To be honest certbot works OK on macOS. True hell is Server.app config handling after importing certificates :-)

bmw:
Is your reason for going to `acme.sh` due to these changes not landing in `certbot-auto`? Is there something that `certbot-auto` gets you that Certbot from Homebrew/MacPorts does not? Don't get me wrong. `acme.sh` is great and if it works better for you please use it, but if there's a use case we're missing we'd like to know about it.

Haha hopefully Certbot (or another ACME client) can better support Server.app someday. I'll be honest and say I've never tried it personally. We'd certainly be open to doing this in Certbot, but there are many servers and systems out there and only so many developers and hours in a day!

MichalMMac:
Sorry. I should have phrase it more clearly. I am going to look at MacPort certbot package first.

I think that would be Apple's job to do. I would not recommend macOS Server to anyone for production use but a lot of our clients use it so we have to deal with it.

PR #4194, opened 2017-02-09
Remove Link rel=next for authzs and new-certs.

jsha:
An early version of the spec indicated that clients should process issuance
sequentially, following Link rel=next from an account URL to an authz URL, to a
new-cert URL. However, the spec has long since moved to putting these URLs in
the directory.

Certbot nominally supports either; This change consolidates on always using the
directory, simplifying things and making the transition to the latest ACME spec
easier.


acme/acme/client_test.py
```diff
@@ -40,6 +40,8 @@ def setUp(self):
                 'https://www.letsencrypt-demo.org/acme/revoke-cert',
             messages.NewAuthorization:
                 'https://www.letsencrypt-demo.org/acme/new-authz',
+            messages.CertificateRequest:
```
cpu: Was there a reason to add this as part of this PR? I don't see an explicit reference anywhere later on but I might be missing something.

ohemorange: It should be used implicitly in tests for `request_issuance`.

cpu: Aha! Thanks for the clarification

PR #4197, opened 2017-02-09
Remove UnexpectedUpdate exceptions.

jsha:
These exceptions trigger when the server sends the client back an object with a
field that doesn't exactly match what the client previously sent.

This causes unnecessary breakage in various cases, doesn't prevent any problems,
and isn't required by spec.


jsha:
Are those properties in the json, or are they pulled from the file path and file timestamp? Because I don't see either property in an example regr.json I have here.

ohemorange:
Ah yeah nope looks like they're only generated programmatically from `now` and `socket.getfqdn()` - you can pass them in, but looks like that's only used for reloading existing accounts, not creating new ones. In that case it's maybe probably fine? Still worried that I'm missing some case somewhere that I just don't know about, but I can't find any obvious immediate pitfalls.

PR #4199, opened 2017-02-09
Add renew deploy cert method for plugins()

### LINKED ISSUE #4046: renew function does not call installer.deploy_cert()
The `renew` function does not call `installer.deploy_cert()` on every renewal. While this is expected behavior, there are also use-cases where it is desired the installer does handle/process the certificate on renewal. Use-cases are coping certs into a java keystore or push them to some other location/database where they are used. Not all application can work with (sym)linking to the certificates/keystore certbot provides itself.

Of course this kind or processing can also be handled by a wrapper script and/or by using the hooks available in `renew` command. But it would be nicer if this is possible within the plugin infrastructure itself, as it is very likely this code would be very similar to the deploy_cert() code when a new certificate is requested and it is better to keep similar processing in the same place.

We propose to introduce a `renew_deploy_cert()` in the `IInstaller` interface, and call this once a certificate is renewed. Standard the implementation should be empty, to keep the same behavior as in the current situation. But now it's up to the plugin developer to decide if they do or don't want to handle the new certificate in some manner. 

- tested against 0.9.3
- tested on 64-bit/debian jessie

(patch pending, testing against 0.11.0.dev.0 pending)

frozen-sky:
Closes #4046

Implemented renew_deploy_cert() function for plugins. To be backward compatible, the interface.py is not changed, but the code checks if the renew_deploy_cert() method is implemented in the plugin. If so, it uses this in a similar way as deploy_cert(). If not, nothing happens.
I made the choice to not reuse the existing deploy_cert() function, as in my opinion, this can have unexpected behavior for users. If you for example make manual changes in config files which are reverted by deploy_cert(), this is unexpected. Also a plugin developer now can handle certs different in a renew scenario then in a install scenario.

If a different implementation is desired please let me know.


ohemorange:
Thanks for the PR!

@bmw, could we add optional methods to our plugin interface?

certbot/client.py
```diff
@@ -362,6 +362,48 @@ def save_certificate(self, certr, chain_cert,
...
+                           "the certificate")
+            raise errors.Error("No installer available")
+
+        chain_path = None if chain_path is None else os.path.abspath(chain_path)
+
+        with error_handler.ErrorHandler(self.installer.recovery_routine):
+            for dom in domains:
+                # this is an optional method. be sure it exists
+                # before executing
+                if (hasattr(self.installer, 'renew_deploy_cert') and
```
ohemorange: https://jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/

Or, in short, it's more Pythonic to write this as `try: except:` than to test if it has the method before calling.

bmw:
I'm not familiar with a way off the top of my head to specify optional methods with `zope.interface`. If we've verified there isn't one, we should just document the method as a comment in the `IInstaller` interface.

ohemorange:
We may also want to call `renew_deploy_certificate` from `run`. This would mean either adding a `_reinstall_cert` to match `_install_cert`, or adding a flag to `_install_cert`.

ohemorange:
Hm, doesn't look like there is a canonical way. I'm good with adding a comment.

ohemorange:
Oh, but then for `_reinstall_cert`, it should probably fall back to `_install_cert` if `renew_deploy_certificate` doesn't work. Which means that `renew_deploy_certificate` should either throw an error or return a failure value.

bmw:
> We may also want to call renew_deploy_certificate from run.

I'd have a preference for not doing this but making this behavior clear. This way, plugins avoid doing duplicate work that they may do in both `deploy_certificate` and `renew_deploy_certificate`.

ohemorange:
For calling it from `run`, it would be specifically for the case where we're reinstalling an existing cert. I suppose that could be called after calling `certonly`. Either way, we're going to be duplicating code in some manner. In practice, it will probably end up looking like moving all/some of the existing code into a helper function, but the `renew_deploy` has some sort of flag to blow away the existing managed lines.

So like for ciphersuites, once we've updated it once, we'll always be setting that in `renew_deploy`, because it's not like we're going to check for the old version. If we eventually end up updating everything, `renew_deploy` will end up always updating everything anyway.

Which, this removes the "doing less work" part of the reason for structuring it this way. I'm still good with having a separate method, because you can explicitly say "get rid of all this old stuff" rather than guessing about what you should do.

bmw:
> For calling it from run, it would be specifically for the case where we're reinstalling an existing cert. I suppose that could be called after calling certonly.

I don't think this works at least with the current form of "reinstall" for `certonly`/`run`. All that's saying is we don't need to obtain a new certificate, you just need to install it. You could run `certbot certonly --standalone` and then later run again with `run` and Apache or Nginx and this would be considered a "reinstall" if your cert hasn't expired.

I think for both our use case and the one that motivated this PR, `deploy` and `renew_deploy` are likely to basically be identical. For us with Nginx, if we see comment "managed by Certbot", you know these lines were created by us on a previous run and we can change them to satisfy the request. For the use case that motivated this PR, maybe there is less work to be done with `renew_deploy`, but that's purely speculation and both have to upload the new certificate and configure the remote webserver to use it.

If the comments we add in Nginx aren't sufficient for some reason, we could get around this by changing our behavior based on whether or not the `installer` from the renewal config is the same, but this seems more complex than it has to be. We added the "managed by Certbot" comments specifically so we can differentiate between values set by us and those set by the user.

ohemorange:
> I think for both our use case and the one that motivated this PR, `deploy` and `renew_deploy` are likely to basically be identical.

Ok, then why not just add a call to `deploy` in the relevant place in `renew`, fix up the `deploy` code to overwrite existing `# managed by Certbot` lines, and call it a day?

bmw:
I proposed that [here](https://github.com/certbot/certbot/issues/4046#issuecomment-273246927) and I kind of like that approach better because it's cleaner, but I think it has the potential to break 3rd party plugins. With that said, if their plugin works calling `run` twice for the same domain, it should™ work here as well.

ohemorange:
Yeah, if it's already breaking when you call it twice something has gone terrible wrong.

So... taking that approach instead that?

ohemorange:
Note: we should definitely check about interactivity here, because currently `install` is pretty interactive. I think `renew` sets `--non-interactive`, but it may only do that implicitly. This might require something like a `--no-enhancements` flag.

bmw:
@frozen-sky, just talked about this at length with ohemorange. While not initially what you had in mind, how would you feel about uploading certificates during the call to `restart()`? In the non-renew case, you can batch updates and perform them all at once when `restart()` is called.

During renewal, you can access certificate storage using `self.config.live_dir` which will point to `/etc/letsencrypt/live` by default. You can then find certificates there and upload them as needed. I'm not sure how you're storing structuring certificate storage remotely, but this might be very trivial for you if your mimicking the structure of the `/etc/letsencrypt/live` directories. If not, you can walk the directories and determine the domains in the certificates by inspecting them.

Let me know what you think! We definitely want to work with you to make Certbot work for your use case but are a little hesitant to start adding more methods to the plugin API.

bmw:
Since we haven't seen a response in roughly a month, I'm closing this PR for now. We can consider reopening if my suggestion above doesn't work for your use case.

blyork:
@bmw

The reason I don't like the proposed workaround is that it expects the plugin to crawl the certificates and determine which certificates have been recently renewed. Why do this when the renew process already knows? It also overloads restart. To top it all off, because this integration would be at the plugin level, some plugins might not implement this features, some might not allow for it to be turned off, or some might allow it to be optional but don't have a convention for what the flag would be called.

Is there a reason that there is not just a flag that will cause renew to call deploy_cert on all renewed certs?

This would avoid breaking expectations for existing installs since they will not be using the flag. This would also avoid needing to add new methods to the plugin interface. Since this flag could be saved in the renewal information, it would be per certificate instead of system wide. Also, all existing plugins would able to take advantage of the feature without modification.

blyork:
Another option would be to add a renew-and-deploy subcommand but then you would lose the certificate-level granularity,

bmw:
@blyork, the discussion here and related issue are outdated. See my comment at https://github.com/certbot/certbot/issues/4046#issuecomment-406349485.

PR #4201, opened 2017-02-09
created an issue template

### LINKED ISSUE #3195: Create issue template
We should create an issue template for this repo to help make sure we get relevant information from a user reporting a new issue. Potential questions we could include in the template are:
- Which OS are you using?
- How did you install Certbot?
- What command(s) did you run?
- What did you expect?
- What was the result?
- Can you provide a log of the problem?

We want to be careful not to discourage users from opening new issues, but getting this information up front allows us to better solve a user's problem.

SwartzCr:
fixes #3195

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,18 @@
+Thanks for using Certbot - if you're experiencing issues with using Certbot
+we'd appreciate it if you could answer a few questions about your environment so
+that we can better assist you:
+
+## My operating system is (include version):
+
+
+## My web server is (include version):
+
+
+## How did you install Certbot:
+
+
+## What command did you run? What output did it produce?
+### Please provide the Certbot error log if you can. It is stored by default in `/var/log/letsencrypt`
+
+
+## Many problems can occur due to your server or DNS records being misconfigured. If you share the domain you're attempting to validate with us we can often debug the issue by accessing the URL and getting some information about this configuration. If you have concerns about publishing this domain publicly you contact one of the maintainers to share it over email.
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,7 +1,3 @@
-Thanks for using Certbot - if you're experiencing issues with using Certbot
-we'd appreciate it if you could answer a few questions about your environment so
-that we can better assist you:
-
 ## My operating system is (include version):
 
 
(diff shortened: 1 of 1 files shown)


SwartzCr:
@bmw @schoen wanna look this over?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+Thanks for using Certbot - if you're experiencing issues with using Certbot
```
bmw: I personally don't think any intro is necessary and we can just start with questions. It makes it less to read and is the format I've seen on most other repos. Feel free to leave it if you disagree though.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+
+## How did you install Certbot:
+
+
+## What command did you run? What output did it produce?
+### Please provide the Certbot error log if you can. It is stored by default in `/var/log/letsencrypt`
+
+
+## Many problems can occur due to your server or DNS records being misconfigured. If you share the domain you're attempting to validate with us we can often debug the issue by accessing the URL and getting some information about this configuration. If you have concerns about publishing this domain publicly you contact one of the maintainers to share it over email.
```
bmw: After thinking about this more, I don't think we should ask for this. I think we should restrict issues on this repo to be issues with Certbot and not more general problems configuring their server. I've found the domain is really useful for helping people debug a webserver they manually misconfigured with certificates from Certbot, but almost never needed to debug a problem with Certbot itself. I asked about this in IRC and everyone who responded agreed.

This coupled with the fact that people are very hesitant to give up their domain makes me think we should nix this question entirely.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+## My operating system is (include version):
+
+
+## My web server is (include version):
+
+
+## How did you install Certbot:
+
+
+## What command did you run? What output did it produce?
```
bmw: [supernit](http://www.baseballnationtx.com/wp-content/uploads/2016/12/2017_SUPER-NIT_banner.jpg): I personally like the wording "What command did you run and what output did it produce?" better. If you like it better this way, we can leave it.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+### Please provide the Certbot error log if you can. It is stored by default in `/var/log/letsencrypt`
```
bmw: I'd make this it's own question like "Can you provide a Certbot log showing the issue?".

If you agree with me about not asking for the domain, I think we should also include wording like the following somewhere:

> Feel free to redact domain names and e-mail and IP addresses as you feel appropriate.

We should also include the sentence about where logs are stored.

I'm not sure of the best way to format this though. If you look at the rendered markdown, the domain question looks really weird with a small paragraph in bold text. I guess this line looks better in slightly softer text under the main question, but still looks a bit off to me. If you can find something you're happy with though then that works for me.

SwartzCr:
@bmw wanna look at it again?

PR #4204, opened 2017-02-10
Change QSA to NE in HTTPS redirection

### LINKED ISSUE #3243: certbot http -> https rewrite rules messes up GET params
Hello,

An issue I encountered today: the rewrite rules added to apache config by certbot-auto to coerce http requests to https doesn't seem to be bulletproof.

In particular, GET params that are url-encoded, eg ?email=testest%2Byhbugfixer%40gmail.com, seem to end up getting encoded twice so that my application receives 'testest%2Byhbugfixer%40gmail.com' for the param value instead of the intended 'testest+yhbugfixer@gmail.com', if I make the initial request in http://

Notably, this does not happen if in place of the rewrite rules added by certbot-auto, I simply create a new virtual host like so:

 &lt;VirtualHost *:80&gt;
        ServerName mydomain.com
        Redirect / https://mydomain.com/
 &lt;/VirtualHost&gt;

sagi:
Fixes #3243 (Prevents double URL encoding).

I've tested it on an Ubuntu 14.04 instance with Apache/2.4.7.
~~~
QSA                                                                             
===                                                                                
http://x.com/t.php?e=Alice%40example.com                                           
->                                                                                 
https://x.com/t.php?e=Alice%2540example.com                                        
                                                                                   
NE                                                                                 
==                                                                                 
http://x.com/t.php?e=Alice%40example.com                                           
->                                                                                 
https://x.com/t.php?e=Alice%40example.com
~~~

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -1315,18 +1315,15 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             #     even with save() and load()
             if not self._is_rewrite_engine_on(general_vh):
                 self.parser.add_dir(general_vh.path, "RewriteEngine", "on")
+
             names = ssl_vhost.get_names()
             for idx, name in enumerate(names):
                 args = ["%{SERVER_NAME}", "={0}".format(name), "[OR]"]
                 if idx == len(names) - 1:
                     args.pop()
                 self.parser.add_dir(general_vh.path, "RewriteCond", args)
-            if self.get_version() >= (2, 3, 9):
-                self.parser.add_dir(general_vh.path, "RewriteRule",
-                                    constants.REWRITE_HTTPS_ARGS_WITH_END)
-            else:
-                self.parser.add_dir(general_vh.path, "RewriteRule",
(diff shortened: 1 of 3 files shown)


sagi:
Thanks for your inputs once again!

Assume one has the following HTTPS redirection rule:
```
RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [END,QSA,R=permanent] 
```
And lets say he updates his cert and hits the Secure option. Then the redirection rule would change to:

```
RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [END,NE,R=permanent] 
```

The only problem is that if there are two (or more) names per VirtualHost (e.g. `example.com` and `www.example.com`, `Certbot` will output:
```
Enhancement redirect was already set.
Enhancement redirect was already set.
```
Which is its exact behavior that occurs without the PR (that is, Choosing "Secure" on a VirtualHost that `Certbot` already enabled https redirection and has two names). I believe this "bug" (?) should be addressed in a different PR.

jsha:
Looks like a solid change to me! Will leave it to a full-time Certbot maintainer for final review though. Thanks for posting.

PR #4230, opened 2017-02-16
Remove reference to #certbot on OFTC

jsha:
The #letsencrypt channel on Freenode is much more active, and is the defacto place for questions about Certbot. Users posting questions on #certbot on OFTC are not getting prompt answers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,7 @@ Main Website: https://certbot.eff.org
 
 Let's Encrypt Website: https://letsencrypt.org
 
-IRC Channel: #letsencrypt on `Freenode`_ or #certbot on `OFTC`_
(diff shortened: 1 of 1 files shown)


bmw:
@SwartzCr, do you have an opinion on this? It's kind of nice for Certbot to have it's own IRC channel to talk about the project, but I think I agree with Jacob; people definitely get faster answers on #letsencrypt since that channel is so much larger.

jsha:
BTW, to expand on my thoughts a little: Right now the Let's Encrypt community forums are the defacto place to get async help on Certbot (and many other clients), and that's worked quite well. I think similarly #letsencrypt has become the defacto real-time place to get help on Certbot, regardless of the name. And it seems to be working fine so far.

PR #4234, opened 2017-02-20
Added documentation about renew exit status #Fixes #4090

### LINKED ISSUE #4090: Document certbot's exit status codes
**Recommendation:** certbot's command line exit status codes should be added to the user documentation.

**Use case:** I have a shell script that does something when certbot updates the certificates that are about to expire. The easiest way I could think to do that would be to run the `cerbot renew` command, and check its exit status.

I observe that the exit code is 1 when the certbot did *not* need to update the certs. I assume the status would be 0 when it does update the certs. Users could possibly also add contingencies to their scripts and tools in case something went wrong and there were other status codes.

yonjah:


docs/using.rst
```diff
@@ -409,6 +409,13 @@ Certbot is working hard to improve the renewal process, and we
 apologize for any inconvenience you encounter in integrating these
 commands into your individual environment.
 
+.. note:: ``certbot renew`` exit status will only be 1 if renewal attempt failed.
```
bmw: `s/if renewal/if a renewal`

docs/using.rst
```diff
@@ -409,6 +409,13 @@ Certbot is working hard to improve the renewal process, and we
...
+  This means ``certbot renew`` exit status will be 0 if no cert needs to be updated.
+  If you write a custom script and expect to run a command only after a cert was actually renewed
+  you will need to use the ``--renew-hook`` since the exit status will be 0 both on successful renewal
+  and when renewal is not necessary
```
bmw: Missing closing period.

docs/using.rst
```diff
@@ -409,6 +409,13 @@ Certbot is working hard to improve the renewal process, and we
 apologize for any inconvenience you encounter in integrating these
 commands into your individual environment.
 
+.. note:: ``certbot renew`` exit status will only be 1 if renewal attempt failed.
+  This means ``certbot renew`` exit status will be 0 if no cert needs to be updated.
+  If you write a custom script and expect to run a command only after a cert was actually renewed
+  you will need to use the ``--renew-hook`` since the exit status will be 0 both on successful renewal
```
bmw: I think we probably want to recommend `--post-hook` here. There are a couple differences:

1. `--renew-hook` will be run once for each cert is renewed. This means if `certbot renew` renews five certificates in one run, the `--renew-hook` is run five times as well.
2. `--renew-hook` is also saved for future renewals. If you renew a cert using `certbot renew --renew-hook "service nginx reload"`, future renewals can just use `certbot renew` and "service nginx reload" will be run whenever that cert is renewed.

PR #4236, opened 2017-02-21
Fix direct usages of the root logger

damz:
A couple of places use `logging.info` and `logging.debug` instead of `logger.info` and `logger.debug`, making it impossible to silence these messages.


damz:
This passed in Travis CI but never reported back. https://travis-ci.org/certbot/certbot/builds/204112129

PR #4237, opened 2017-02-21
acme: Make the network timeout configurable

damz:
Network requests currently have no timeouts, which under certain network conditions can mean the program will hang indefinitely.

The documentation of the Requests library is very clear on how important this is:

> Nearly all production code should use this parameter
> in nearly all requests. Failure to do so can cause
> your program to hang indefinitely.

http://docs.python-requests.org/en/master/user/quickstart/#timeouts

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -33,6 +33,8 @@ if sys.version_info < (2, 7, 9):  # pragma: no cover
         import urllib3.contrib.pyopenssl  # pylint: disable=import-error
         urllib3.contrib.pyopenssl.inject_into_urllib3()
 
+DEFAULT_NETWORK_TIMEOUT = 45
+
 DER_CONTENT_TYPE = 'application/pkix-cert'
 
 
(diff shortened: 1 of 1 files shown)


cpu:
I think this is a duplicate of https://github.com/certbot/certbot/pull/4217 from @jsha.

bmw:
Yes this is a duplicate of #4217 which was just merged. Thanks for the PR though @damz!

damz:
Could we please make that configurable? The default 45s is _really_ high. I'll resend the PR.

joohoi:
It is configurable in #4217 , the 45 is just a default value if none is provided

damz:
@joohoi No, it's not. This is deep inside a private method of `ClientNetwork`, that is only called from inside that class.

joohoi:
But the kwargs are passed through from every public method that calls the private method in question.

jsha:
Agreed this should be configurable, though I think 45s is a good default. It's slightly above an internal maximum in the Let's Encrypt servers. We generally aim to complete requests much, much faster than that, but I don't want a situation where the servers are super slow under heavy load, but still doing work and returning eventually, to clients that have already timed out the request.

@damz, can you merge master, keeping the 45 second default, but otherwise plumbing through the configurability?

jsha:
Also, looks like tests are failing. I recently pushed some updated documentation on running the tests that provides a much easier and faster way to iterate on them locally: https://github.com/certbot/certbot/blob/5c2ee64f3de28746bee60753161a51500c0b947e/docs/contributing.rst#id4.

damz:
@jsha Sorry for the delay. Rebased on top of `master`. `pylint` was complaining about the number of arguments to `Client.__init__`, which sounds reasonable, so I ended up only adding the timeout argument to `ClientNetwork`.

jsha:
Excellent, thanks!

PR #4254, opened 2017-02-24
Add an account deactivate utility script.

jsha:
This is handy if you created an account with a tool other than Certbot, and want
to deactivate the account.


jsha:
I'll put it in tools just in case someone needs it in the future, if that sounds good to you. I agree it would be better to solve this more generically, but I think that's unlikely to happen soon, and we have at least one user who needed this.

bmw:
Closing this since it's had unaddressed review comments for roughly a month now. Please reopen when it's ready for another review.

jsha:
Pushed a comment about how to set the server; reopening.

PR #4256, opened 2017-02-24
Polish issue template

bmw:
I removed the web server question because it's irrelevant/obvious for most issues. I think it's more important when helping people configure their server manually.

I also added the question asking how Certbot's behavior differed from what you expected. Sometimes this is obvious (e.g. "I didn't expect Certbot to crash"), however, actual/expected results is a common and useful template here and often shows differences between how Certbot works and what the user intended.

Finally, I made the wording/style of the questions consistent.

@SwartzCr, can you take a look at this?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,16 +1,14 @@
 ## My operating system is (include version):
 
 
-## My web server is (include version):
+## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
 
 
-## How did you install Certbot:
+## I ran this command and it produced this output:
 
 
-## What command did you run and what output did it produce?
+## Certbot's behavior differed from what I expected because:
 
 
-## Can you provide a Certbot error log showing the issue?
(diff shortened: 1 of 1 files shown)


SwartzCr:
looks good - though I'm curious if you think that we should be asking people what version of Certbot they're running.
I know in some situations this will be obvious (packaged and cb-auto) but I'm guessing sometimes it wont? Feel free to add that or just merge as is

bmw:
Good question. The version is included early in the Certbot log so I personally don't think it's really necessary. Scenarios where we may ask for this information are:

* User doesn't provide Certbot logs (although we may just ask them again to provide a log if we need more information).
* Certbot crashes before logging the version number (which happens but is extremely rare).
* `certbot-auto` crashes before starting Certbot. One of the first things `certbot-auto` does is upgrade itself so the user should be on the latest version. There could be a problem with the upgrade process or the user could be providing `--no-self-upgrade` on the command line though.

So there certainly are scenarios where we wouldn't have this information, but I'm not sure if they're common enough that it warrants asking the question to all users. If you'd like to see this question added though, feel free to send a PR and I'll approve it.

PR #4263, opened 2017-02-25
Fix python3 incompatibility in certbot-auto

sedrubal:
Use printfunction from `__future__` in order to get letsencrypt installed on a python3 only system.


bmw:
Thanks for the PR! Looks like tests are failing, but they can be fixed following the instructions [here](https://certbot.eff.org/docs/contributing.html#updating-certbot-auto-and-letsencrypt-auto).

sedrubal:
There was only one way to fail and I failed... 🙈  Thanks 👍

PR #4268, opened 2017-02-27
Make mod-check more flexible

### LINKED ISSUE #4166: Make modification_check.sh more flexible
Currently Travis tests will fail if `{certbot,letsencrypt}-auto` in the root of our repo isn't the same as the version from the most recent release tag. The result of this is Travis tests will fail from the time the release is pushed live until the PR incorporating `certbot-auto` changes is merged. Instead, we should allow `certbot-auto` to either be the version from the most recent release tag or the version currently in `master`.

bmw:
After the last release, Travis tests on all PRs failed for a week until #4162 was merged. This PR fixes #4166 making our tests more flexible to prevent this problem in the future.

@ohemorange, would you feel comfortable reviewing this? I'd like to get this in by Wednesday to prevent everyone's PRs from breaking again.

EDIT: I forgot to mention I tested this against the `0.10.x` branch as you can see [here](https://travis-ci.org/certbot/certbot/jobs/205909523).

### DIFF
--- a/tests/modification-check.sh
+++ b/tests/modification-check.sh
@@ -3,10 +3,8 @@
 temp_dir=`mktemp -d`
 trap "rm -rf $temp_dir" EXIT
 
-# Script should be run from Certbot's root directory
-
-SCRIPT_PATH=`dirname $0`
-SCRIPT_PATH=`readlink -f $SCRIPT_PATH`
+# cd to repo root
+cd $(dirname $(dirname $(readlink -f $0)))
 FLAG=false
 
 if ! cmp -s certbot-auto letsencrypt-auto; then
--- a/tests/modification-check.sh
+++ b/tests/modification-check.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 temp_dir=`mktemp -d`
+trap "rm -rf $temp_dir" EXIT
 
 # Script should be run from Certbot's root directory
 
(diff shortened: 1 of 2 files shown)


tests/modification-check.sh
```diff
@@ -1,46 +1,39 @@
...
-
-SCRIPT_PATH=`dirname $0`
-SCRIPT_PATH=`readlink -f $SCRIPT_PATH`
+# cd to repo root
+cd $(dirname $(dirname $(readlink -f $0)))
 FLAG=false
 
-# Compare root letsencrypt-auto and certbot-auto with published versions
-
-cp letsencrypt-auto ${temp_dir}/letsencrypt-to-be-checked
```
ohemorange: is there some escaping reason that we had `{}` here?

tests/modification-check.sh
```diff
@@ -1,46 +1,39 @@
...
+    cp certbot-auto "$temp_dir/local-auto"
+    cp letsencrypt-auto-source/pieces/fetch.py "$temp_dir/fetch.py"
+    cd $temp_dir
+
+    # Compare file against current version in the target branch
+    BRANCH=${TRAVIS_BRANCH:-master}
+    URL="https://raw.githubusercontent.com/certbot/certbot/$BRANCH/certbot-auto"
+    curl -sS $URL > certbot-auto
+    if cmp -s certbot-auto local-auto; then
+        echo "Root *-auto were unchanged."
```
ohemorange: This is just checking that root certbot-auto is unchanged, right? or is it implicitly checking that letsencrypt-auto was unchanged, somewhere?

tests/modification-check.sh
```diff
@@ -1,46 +1,39 @@
...
+        # Compare file against the latest released version
+        python fetch.py --le-auto-script "v$(python fetch.py --latest-version)"
+        if cmp -s letsencrypt-auto local-auto; then
+            echo "Root *-auto were updated to the latest version."
+        else
+            echo "Root *-auto have unexpected changes."
+            FLAG=true
+        fi
+    fi
+    cd ~-
```
ohemorange: is there a reason to use `~-` vs just `-`?

tests/modification-check.sh
bmw: It's implicitly checking that `letsencrypt-auto` was unchanged. The [first comparison](https://github.com/certbot/certbot/pull/4268/files#diff-b534c5f8ac7b0d2c8b5e9eb514cab83fR10) was checking that `certbot-auto` and `letsencrypt-auto` were equal. If they're not, the test already failed and we don't execute this code.

tests/modification-check.sh
bmw: `~-` doesn't print the path you're `cd`ing to:
```
bmw@localhost:/tmp/tmp.5OdoAuteiN/dir1$ cd ../dir2/
bmw@localhost:/tmp/tmp.5OdoAuteiN/dir2$ cd -
/tmp/tmp.5OdoAuteiN/dir1
bmw@localhost:/tmp/tmp.5OdoAuteiN/dir1$ cd ../dir2/
bmw@localhost:/tmp/tmp.5OdoAuteiN/dir2$ cd ~-
bmw@localhost:/tmp/tmp.5OdoAuteiN/dir1$
```

ohemorange: fancy

tests/modification-check.sh
ohemorange: "In certain contexts, only the less ambiguous ${parameter} form works." lol. probably not a problem...

Issue #3798, opened 2016-11-15
nginxparser fails for given test case


joystein:
Versoin of certbot (installed as root with `pip install --user certbot-nginx`):

```bash
# pip list --user | grep certbot
certbot (0.9.3)
certbot-nginx (0.9.3)
```

Os: debian wheezy (7.11)

Command (run through fabric from workstation to remote server):
```
~/.local/bin/certbot certonly --nginx --email username@domain.no -d sub.domain.no # (real domain redacted)
```

Actual output:
```
Cannot find a VirtualHost matching domain player.fabelapp.no.
```

Expected: that certbot would have found the server block in the nginx config files.

I should note that i have a possibly unusual structure for nginx config files:
```
|-> nginx 
|        |-> sites-{available,enabled}
|        |        |-> domain1.no
|        |        |       |-> www.conf
|        |        |       |-> api.conf
|        |        |-> domain2.no
|        |        |       |-> www.conf
|        |        |       |-> api.conf
|        |        |-> domain3.no
|        |        |       |-> www.conf
|        |        |       |-> api.conf
```

This is matched with the following include statements in the main nginx config:
```
    include /etc/nginx/conf.d/*.conf;
    include /etc/nginx/sites-enabled/*.conf;
    include /etc/nginx/sites-enabled/*/*.conf;
```

I have however tried moving the config files i have been testing to directly under `sites-{available,enabled}` and i still get the same error.

In the end, i downgraded to `pip install ---user 'certbot-nginx < 0.9.0'` and finally the command ran, and i could go on with the installation step.

bmw:
@ohemorange, any ideas here?

ohemorange:
Hi @joystein, would you feel comfortable posting an example config file? Feel free to redact any information you'd like.

joystein:
It should be pretty standard stuff, i believe. The initial file looked something like (only domain name altered):

```
/etc/nginx/sites-{available,enabled}/domain1.no/api.conf:

server {
    listen 80;
    server_name api.domain1.no;
    access_log /var/log/nginx/api_domain1_no.access.log;
    error_log /var/log/nginx/api_domain1_no.error.log;
    gzip on;
    gzip_types application/json;

    location / {
        uwsgi_pass uwsgi-router-local;
        include uwsgi_params;
        uwsgi_param HTTP_X_FORWARDED_PROTO     $scheme;
    }
}
```

After running `certbot-nginx < 0.9.0` it looks something like this:

```
server {
[...]

PR #4272, opened 2017-02-28
Fix nginx map parsing problem

ohemorange:
Fixes https://github.com/certbot/certbot/issues/3798#issuecomment-278267350/


ohemorange:
Case that was failing before:
```
http {
    map $http_upgrade $connection_upgrade {
      default upgrade;
      ''      close;
      "~Opera Mini" 1;
      *.example.com 1;
    }
    one;
}
```
The problem was when a map block was directly followed by an assignment. It didn't like our use of `nonspaced` for some reason.

I'm not convinced that the rest of the map parsing code is accurate, and would feel more comfortable with more extensive testing of what nginx allows vs. what the parser matches on.

I plan to add this as a test case in another commit before merging.

ohemorange:
Anyway I think the problem was that `\S` eats newline.

ohemorange:
Closing in favor of #4296.

Issue #2750, opened 2016-04-02
Error with renewal conf files that no longer have a "server" entry


pde:
I have some renewal conf files that have started erroring since we've cleaned up renewal config entries. Here's one such file:
```
cert = /etc/letsencrypt/live/isnot.org/cert.pem
privkey = /etc/letsencrypt/live/isnot.org/privkey.pem
chain = /etc/letsencrypt/live/isnot.org/chain.pem
fullchain = /etc/letsencrypt/live/isnot.org/fullchain.pem

# Options used in the renewal process
[renewalparams]
apache_enmod = a2enmod
standalone_supported_challenges = "tls-sni-01,http-01"
account = a015e247a80f161ed2a1918624eecadc
apache_handle_modules = True
apache_server_root = /etc/apache2
apache_dismod = a2dismod
apache_handle_sites = True
authenticator = apache
apache_le_vhost_ext = -le-ssl.conf
apache_challenge_location = /etc/apache2
nginx_server_root = /etc/nginx
installer = apache
nginx_ctl = nginx
apache_vhost_root = /etc/apache2/sites-available
```
The error is:
```
2016-04-02 19:31:38,725:WARNING:letsencrypt.renewal:Attempting to renew cert from /etc/letsencrypt/renewal/isnot.org.conf produced an unexpected error: 'server'. Skipping.
2016-04-02 19:31:38,726:DEBUG:letsencrypt.renewal:Traceback was:
Traceback (most recent call last):
  File "/home/ubuntu/letsencrypt/letsencrypt/renewal.py", line 344, in renew_all_lineages
    main.obtain_cert(lineage_config, plugins, renewal_candidate)
  File "/home/ubuntu/letsencrypt/letsencrypt/main.py", line 513, in obtain_cert
    _, action = _auth_from_domains(le_client, config, domains, lineage)
  File "/home/ubuntu/letsencrypt/letsencrypt/main.py", line 90, in _auth_from_domains
    renewal.renew_cert(config, domains, le_client, lineage)
  File "/home/ubuntu/letsencrypt/letsencrypt/renewal.py", line 236, in renew_cert
    original_server = lineage.configuration["renewalparams"]["server"]
  File "/home/ubuntu/letsencrypt/venv/local/lib/python2.7/site-packages/configobj.py", line 554, in __getitem__
    val = dict.__getitem__(self, key)
KeyError: 'server'

```

pde:
[...]

PR #4277, opened 2017-03-01
Revert "Remove Link rel=next for authzs and new-certs."

bmw:
Reverts certbot/certbot#4194. #4194 broke forward compatibility with Certbot accounts. Previous versions of Certbot expected and required `new_authzr_uri` to be present in `regr.json` and this field was removed in this PR. This breakage may be OK as we've broken forward compatibility in the past (see #2750), but it was unexpected behavior found during the release process. Let's think about whether or not this is what we want to do.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -71,13 +71,20 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             self.directory = directory
 
     @classmethod
-    def _regr_from_response(cls, response, uri=None, terms_of_service=None):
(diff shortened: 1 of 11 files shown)


bmw:
I think if we do anything to keep forward compatibility here, we should keep the scope of the change as small as possible. My thought was to add the current `new-authz` URI to the json [here](https://github.com/certbot/certbot/blob/v0.11.1/certbot/account.py#L219G). This value will be dropped when loading the registration resource using new versions of the client, however, old versions of the client will continue to work. This would require passing the `acme.client.Client` object, directory object, or `new-authz` URI into the `save()` method.

PR #4280, opened 2017-03-02
Store new_authzr_uri in AccountFileStorage.

jsha:
This is an attempt to make switching to older Certbot versions work, fixing the issue discussed in #4277. Note that I'm still fine landing #4277 in order to unblock the release, but wanted to get this out there for discussion. It's quite hacky, so let me know what you think!

### DIFF
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -117,6 +117,18 @@ class AccountMemoryStorage(interfaces.AccountStorage):
         except KeyError:
             raise errors.AccountNotFound(account_id)
 
+class RegistrationResourceWithNewAuthzrURI(messages.RegistrationResource):
+    """A backwards-compatible RegistrationResource with a new-authz URI.
+
+       Hack: Certbot versions pre-0.11.1 expect to load
+       new_authzr_uri as part of the account. Because people
+       sometimes switch between old and new versions, we will
+       continue to write out this field for some time so older
+       clients don't crash in that scenario. Because this is a
+       temporary hack, just hardcode the known new-authz URI for
+       Let's Encrypt.
+    """
+    new_authzr_uri = jose.Field('new_authzr_uri')
 
 class AccountFileStorage(interfaces.AccountStorage):
     """Accounts file storage.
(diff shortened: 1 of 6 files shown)


ohemorange:
I assume this should be on top of a commit reverting the revert?

certbot/account.py
```diff
@@ -117,6 +117,18 @@ def load(self, account_id):
...
+
+       Hack: Certbot versions pre-0.11.1 expect to load
+       new_authzr_uri as part of the account. Because people
+       sometimes switch between old and new versions, we will
+       continue to write out this field for some time so older
+       clients don't crash in that scenario. Because this is a
+       temporary hack, just hardcode the known new-authz URI for
+       Let's Encrypt.
+    """
+    new_authzr_uri = jose.Field('new_authzr_uri')
```
ohemorange: Will old clients ever need to actually use the saved value, or does it just crash when it tries to load it but it actually uses a different method? Because if the former, this is temporary in the sense of "until we get updated versions into all distros we support", which might happen some time after a potential change to the url.

jsha: Old clients will indeed use the saved value. I was being lazy here because I think the URL is very unlikely to change during the lifetime of the v1 API. But I've pushed some changes to do the more rigorous thing. If you think they look good, I'll rebase this on top of a revert of the revert.

jsha:
Getting this error in the py33+ builds. Note sure what's causing it?

```
FAIL: test_find_all (certbot.tests.account_test.AccountFileStorageTest)
  File "/home/travis/build/certbot/certbot/certbot/tests/account_test.py", line 149, in test_find_all
    self.assertEqual([self.acc], self.storage.find_all())
AssertionError: Lists differ: [<Account(bca5889f66457d5b62fb... != [<Account(bca5889f66457d5b62fb...
First differing element 0:
<Account(bca5889f66457d5b62fbba7b25f9ab6f)>
<Account(bca5889f66457d5b62fbba7b25f9ab6f)>
  [<Account(bca5889f66457d5b62fbba7b25f9ab6f)>]
```

jsha:
Closing in favor of #4303, which passes tests.

PR #4281, opened 2017-03-02
Save hyphenated plugin params for renewal

### LINKED ISSUE #4219: Plugin renewalparams not saved in renewal config files
I'm using the s3front plugin and I set various args via the commandline (e.g. --certbot-s3front:auth-s3-bucket). However, these args are not saved in the renewal config file. So when certbot detects the cert needs renewing it fails as no args for the plugin were given.

Any ideas why this is?

(using 0.11.1)

bmw:
Fixes #4219.

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -19,6 +19,9 @@ from certbot import errors
 from certbot import error_handler
 from certbot import util
 
+from certbot.plugins import common as plugins_common
+from certbot.plugins import disco as plugins_disco
+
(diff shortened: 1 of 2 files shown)


bmw:
@schoen do you have the spare cycles to review this?

ohemorange:
Code looks good, have you tested live with a plugin with a dash in the title?

bmw:
Yeah. I tested it with `certbot-s3front`.

Issue #4163, opened 2017-02-02
certbot-auto -q prints messages on upgrade


jsha:
This may be "working as intended," or "too much hassle to fix, since it would involve overlapping arg parsers," but I figured I'd post for discussion:

I have a cron job that runs `certbot-auto -q renew`. Normally it produces no output, but on the most recent run, it printed (and therefore emailed me):

```
Upgrading certbot-auto 0.10.2 to 0.11.0...
Replacing certbot-auto...
Creating virtual environment...
Installing Python packages...
Installation succeeded.
```

Expected results: No output on self-upgrade when `-q` is present.

jsha:
Oops, just caught up with the release notes and I see this is fixed in the latest release. Thanks!

bmw:
This output is actually still printed (and you'll probably get e-mail about updating to 0.11.1 if you haven't already). In the latest release, we just silenced output from your OS package manager which should only happen when Certbot first runs.

This should be pretty easy to fix though. Basically we just have to make sure that all `echo` statements that aren't errors are only printed if `-q/--quiet` isn't provided.

jsha:
I think this is fixed now; is that right?

bmw:
Yep! Thanks for catching this.

PR #4292, opened 2017-03-03
Make it easier to honor --quiet

jsoref:
This should help address #4163.

This doesn't cover passing quiet into the various platform bits, but at least it should make certbot-auto itself quieter...

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -104,6 +104,10 @@ say() {
     fi
 }
 
+error() {
+    echo "$@"
+}
+
 # Support for busybox and others where there is no "command",
 # but "which" instead
 if command -v command > /dev/null 2>&1 ; then
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -115,7 +115,7 @@ if command -v command > /dev/null 2>&1 ; then
 elif which which > /dev/null 2>&1 ; then
   export EXISTS="which"
 else
-  echo "Cannot find command nor which... please install one!"
+  error "Cannot find command nor which... please install one!"
   exit 1
 fi
 
@@ -160,7 +160,7 @@ if [ -n "${LE_AUTO_SUDO+x}" ]; then
       ;;
     '') ;; # Nothing to do for plain root method.
     *)
-      echo "Error: unknown root authorization mechanism '$LE_AUTO_SUDO'."
+      error "Error: unknown root authorization mechanism '$LE_AUTO_SUDO'."
       exit 1
   esac
   say "Using preset root authorization mechanism '$LE_AUTO_SUDO'."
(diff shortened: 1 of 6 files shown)


letsencrypt-auto-source/pieces/bootstrappers/deb_common.sh
```diff
@@ -21,7 +21,7 @@ BootstrapDebCommon() {
     QUIET_FLAG='-qq'
   fi
 
-  $SUDO apt-get $QUIET_FLAG update || echo apt-get update hit problems but continuing anyway...
+  $SUDO apt-get $QUIET_FLAG update || say apt-get update hit problems but continuing anyway...
```
bmw: I think this should be `error` so the user is notified that we encountered problems.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
```diff
@@ -25,9 +25,9 @@ BootstrapRpmCommon() {
   fi
 
   if ! $SUDO $tool list *virtualenv >/dev/null 2>&1; then
-    echo "To use Certbot, packages from the EPEL repository need to be installed."
+    say "To use Certbot, packages from the EPEL repository need to be installed."
```
bmw: I'm a little unsure we should be using `say` here as well. This code should only be executed the first time `certbot-auto` runs on a system. In this case, if the user runs `certbot-auto --quiet`, they'll get a `sudo` prompt with no context. I'm not sure if this is the behavior we want.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -309,42 +319,40 @@ UNLIKELY_EOF
...
+    say "Installation succeeded."
   fi
   if [ -n "$SUDO" ]; then
     # SUDO is su wrapper or sudo
-    if [  "$QUIET" != 1 ]; then
-      echo "Requesting root privileges to run certbot..."
-      echo "  $VENV_BIN/letsencrypt" "$@"
-    fi
+    say "Requesting root privileges to run certbot..."
+    say "  $VENV_BIN/letsencrypt" "$@"
```
bmw: I have the same comment here about the promptless `sudo` call when run interactively. Unless you have thoughts about how to work around this, I recommend leaving these as `echo` for now and we'll continue work on this.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -385,9 +393,9 @@ UNLIKELY_EOF
     # ---------------------------------------------------------------------------
     DeterminePythonVersion
     if ! REMOTE_VERSION=`"$LE_PYTHON" "$TEMP_DIR/fetch.py" --latest-version` ; then
-      echo "WARNING: unable to check for updates."
+      say "WARNING: unable to check for updates."
```
bmw: I think this should be `error` as well so the user is notified `certbot-auto` encountered a problem.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
jsoref: Well, we might be able to do `sudo -A ...` and use the `-A` argument to spit out stuff we should have printed earlier... I'm not sure how portable `-A` is...

jsoref: I think we could probably try running `$tool list ...`, capture and process the output and then decide to spit out the message and run `sudo $tool ...`.

bmw: We could also wrap `sudo` the way we do for `su` to echo the command we're about to run, but many users only have to type in their `sudo` password the first time meaning we'll be printing a lot of unnecessary output. Ideally we find a way to only print the message if we know we're going to prompt for a password. We could do this by capturing the output, but I want to make sure whatever we do isn't too complex and is robust on the wide variety of systems `certbot-auto` is run on.

PR #4295, opened 2017-03-03
Rewrite our Dockerfile

### LINKED ISSUE #3782: Rewrite our Dockerfile

bmw:
Fixes #3782 and builds on #4032.

This PR renames our current Dockerfile to `Dockerfile-old` and changes `Dockerfile` to be an Alpine image. I've turned off automatic builds of https://quay.io/repository/letsencrypt/letsencrypt for now during this transition.

To accomplish the deprecation warnings suggested by Joona [here](https://github.com/certbot/certbot/issues/3782#issuecomment-283319093), I wrote a shell script to wrap Certbot. Initially I planned on simply changing `Dockerfile-old`'s `ENTRYPOINT` to this new script, but this won't give deprecation warnings to [people who overwrote the entrypoint](https://github.com/czerasz/docker-letsencrypt-webroot). On the other hand, this approach won't give warnings to people who call `certbot` or `letsencrypt` using a direct path (`/opt/certbot/venv/bin/certbot` or `venv/bin/certbot`), but I'm not sure it's worth the effort to change this.

After this is merged, I plan on resolving #492 and then resuming automatic builds on quay.io.

@joohoi, could you take a look at this?

EDIT: This PR should be merged, not squashed.


jsha:
This looks great! I assume by requesting merge rather than squash, your intention is to preserve credit to the original author? If so I might suggest rebasing some of your smaller, less descriptive commits, like "do less" into a larger, more descriptive commit.

joohoi:
Alpine image looks good, it appears to shave 70% off of the container image size as well. For the old image: could we print the deprecation message after the certbot run (as well)? It goes easily unnoticed if ran before certbot itself.

bmw:
The biggest challenge with printing the warning after Certbot runs is we run `exec certbot $@` so Certbot receives and is able to handle any Unix signals. Certbot uses its own signal handlers in a few places, but most relevant here is when modifying your certificates and symlinks to try to ensure they're left in a consistent state.

With this in mind, we have a few options:

1. Don't worry about Certbot not reacting to signals and stop using `exec`.
2. Add signal handlers to `docker-warning.sh` that send the signals to Certbot.
3. Just leave the deprecation messages before Certbot.

I'm personally leaning towards option 3 as I'm not sure the extra complexity in `docker-warning.sh` is worth it, but I'd like to hear what others think. I'm happy to add signal handlers to `docker-warning.sh` if others think that's the right approach.

jsha:
I'm inclined towards #3 also.

joohoi:
I agree, didn't consider signals. 3 is fine.

bmw:
Closing and reopening due to GH trouble.

bmw:
Anyone understand why this isn't mergable?

jsha:
GitHub says "Required statuses must pass before merging," but clearly the status checks have indeed passed.

I would not be surprised if there is a bug with handling the `++` in the branch name, which is likely to get canonicalized in funny ways if it shows up in a URL. Probably best to open a new branch?

bmw:
Interesting. I've had similar branch names before (see #4268), but something may have changed. I'll try opening a new branch.

bmw:
I filed a help ticket with GH about status checks.

PR #4296, opened 2017-03-04
Fix nginx parser

### LINKED ISSUE #3798: nginxparser fails for given test case

ohemorange:
This PR rewrites the main logic of `nginxparser.py` to allow the same syntax that nginx does. This fixes #3798, and supersedes #4272.


ohemorange:
Many thanks to http://pyparsing.wikispaces.com/HowToUsePyparsing

ohemorange:
No thanks to https://github.com/nginx/nginx/blob/master/src/core/ngx_conf_file.c

ohemorange:
I'm actually not super convinced that nginx even allows things like `try_files /typo3temp/tx_ncstaticfilecache/$host${request_uri}index.html @nocache`, but it's in the compatibility test so let's parse it. We allow `$variable`, `${variable}`, `${}`, and `$` for maximum permissibility.

I'm choosing to do that because we don't actually use most of the things we parse, so if people have strange stuff in there, it's probably better to accidentally parse something that isn't valid rather than the reverse.

ohemorange:
Note: "sdfasdf"asdfasdf is considered invalid by nginx, so we're not going to bother trying to parse that correctly (`token = squoted | dquoted | unquoted` isn't `token = Combine(squoted | dquoted | unquoted)`).

ohemorange:
Note (II): This PR changes how we parse assignments. Previously, `listen 80 ssl;` would be parsed as `['listen', '80 ssl']`, but will now be parsed as `['listen', '80', 'ssl']`. This makes things cleaner in a bunch of places, and only required a couple small changes at one abstraction layer up.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -2,85 +2,53 @@
...
-    value = Combine(ZeroOrMore(dquoted | squoted | varsub | nonspecial))
-
-    location = CharsNotIn("{};," + string.whitespace)
-    # modifier for location uri [ = | ~ | ~* | ^~ ]
-    modifier = Literal("=") | Literal("~*") | Literal("~") | Literal("^~")
-
-    # rules
-    comment = space + Literal('#') + restOfLine
+    tokenchars = Regex(r"[^\{\};\s\$]+")
+    variable = Regex(r"(\$\{\w*\})") | Regex(r"(\$\w*)")
```
schoen: Should this perhaps be `\w+` instead of `\w*` in both cases? Is the name permitted to be empty?

schoen: The old version used `\w+` in the `varsub` context, which does seem to be an argument for the `+`, unless you've found that the grammar allows empty variable identifiers (??).

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -468,6 +468,17 @@ def _is_include_directive(entry):
             len(entry) == 2 and entry[0] == 'include' and
             isinstance(entry[1], str))
 
+def _is_ssl_on_directive(entry):
+    """Checks if an nginx parsed entry is an 'include' directive.
```
schoen: I think the first line of this docstring was wrongly left unchanged when pasting from elsewhere? Should be "is an 'ssl on' directive", matching the `returns` value right below.

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -2,85 +2,53 @@
...
     semicolon = Literal(";").suppress()
-    key = Word(alphanums + "_/+-.")
-    dollar_var = Combine(Literal('$') + Regex(r"[^\{\};,\s]+"))
-    condition = Regex(r"\(.+\)")
-    # Matches anything that is not a special character, and ${SHELL_VARS}, AND
-    # any chars in single or double quotes
-    # All of these COULD be upgraded to something like
-    # https://stackoverflow.com/a/16130746
     dquoted = Regex(r'(\".*\")')
     squoted = Regex(r"(\'.*\')")
```
schoen: I know these two definitely are unchanged from before, but could this be inappropriately greedy where the quoted string includes other things in the middle, like

`"hello"; # now you don't notice this is a comment "heh heh"`

or something? (`dquoted` would match that entire line, if it's allowed to.) I know the parser has ways of handling this through prioritization of the things that are built up out of these elements, but was the context in which `dquoted` and `squoted` get used correct in this regard before, and is it still correct now?

A possibly safer alternative would be

`dquoted = Regex(r'(\"[^"]*\")')`

`squoted = Regex(r"(\'[^']*\')")`

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -2,85 +2,53 @@
...
 
-    block = Forward()
+    whitespace_token_group = space + OneOrMore(token + space)
+    assignment = whitespace_token_group + semicolon
 
-    # key could for instance be "server" or "http", or "location" (in which case
-    # location_statement needs to have a non-empty location)
+    # whitespace_token_group will parse this fine, but we need to know about
+    # if statements for redirects, so let's explicitly parse this.
+    condition = Regex(r"\(.+\)")
```
schoen: Also a little nervous about `.+\)`, which could include a closing `\)` inside the `.+`; is there something in context that effectively guarantees that the parentheses are balanced?

ohemorange: The only special treatment of `)` in the actual parsing module is that it's a little like whitespace, in that it allows a token to end.

Fun fact! `one'test'` is allowed, but `'test'one` isn't!

Yeah! So `)` is mostly normal, except it's the only character allowed right after a quoted token finishes. Fun!

You can only begin a quote at the beginning of a token. In the middle or the end, it doesn't count as starting a quote. Isn't this game great?

ohemorange: I rewrote things to reflect these changes. The tests show what is and isn't allowed.

certbot-nginx/certbot_nginx/parser.py
ohemorange: That is true!

ohemorange:
@schoen, lots of comments, and also I changed things in a fairly significant way again. Woohoo!

schoen:
@ohemorange has explained a lot of this to me and it looks good, though I didn't try to verify the correspondence to the nginx parser!

Noting that we should check what happens when newlines happen in the middle of a directive (from the example of a `server_name` directive that has a list of names with intervening newlines). Would we pick up all the list items that the original parser does? Does their grammar permit escaping a newline with `\` to remove its special meaning, if any, like

```
server_name foo.example.com bar.example.com \
            baz.example.com qux.example.com;
```

?

ohemorange:
Note: do not merge this yet!

ohemorange:
@schoen it works with and without the `\`! Both in our parser and in theirs. Test added.

ohemorange:
hahahahaa this has to be a bug:

`directive ${variablename};` errors out, but `directive $variablename;` is fine. (It's because the variable-state-and-brace-escaping should also be in the `last_space` section but it's only in the `else` section.)

ohemorange:
@schoen:

in one of the test files, we have:
```
 location ~ ^/(images|javascript|js|css|flash|media|static)/ {
   root ${PROJECTBASE}/${PROJECTNAME}/static;
 }
```
was this something that we wrote/modified or that was submitted? because I'm pretty certain that this is invalid nginx syntax.

schoen:
@ohemorange,

It's derived from the example at

https://www.nginx.com/resources/wiki/start/topics/examples/simplepythonfcgi/

It's possible that there are variable substitutions that have to be made by a human (not by nginx) in order for it to be valid nginx. If so, sorry for the mix-up!

ohemorange:
Ah, yup. Changing that now. Anywhere else you know of that might be analogously not actually a valid config file?

schoen:
I'm afraid I didn't notice this phenomenon when extracting these from the nginx site. If you do get any parse failures on other examples, I'd be happy to double-check the context of where they were taken from. Sorry for the inconvenience about this. I think it's a clear sign that you're making the parser more accurate!

ohemorange:
Not seeing other parse failures currently! If anything, I'm more worried about things that we didn't used to be able to parse that were manually "fixed" so that we could parse them, and losing those test cases. Do you know of any places where we did that?

schoen:
> Not seeing other parse failures currently! If anything, I'm more worried about things that we didn't used to be able to parse that were manually "fixed" so that we could parse them, and losing those test cases. Do you know of any places where we did that?

Not that I can think of at the moment! At least not for the roundtrip testdata.

Issue #4078, opened 2017-01-18
Dedupe test code


bmw:
As pointed out in the discussion of #3972, we have some common patterns in our tests.

#### Pattern 1
`setUp()` creates a temporary directory and `tearDown()` destroys it.

#### Pattern 2
`setUp()` creates a temporary directory and creates/sets `config_dir`, `work_dir`, and/or `logs_dir` paths to be somewhere inside it. `tearDown()` destroys all directories.

Let's dedupe this code. I think the best approach is to make a class in `certbot.tests.util` called something like `TempDirTestCase` for pattern 1 that is a subclass of `unittest.TestCase` and make another class for pattern 2 that is a subclass of `TempDirTestCase`.

bmw:
Fixed in #4297.

PR #4297, opened 2017-03-04
tests: deduplicate temporary directory code (#4078)

zjs:
Introduce a test class to deduplicate temporary directory setup and teardown in
testing code and update existing test code to use this new class.

Note: #4078 mentions two patterns, but I didn't see duplicated code involving `config_dir`, `work_dir`, `logs_dir` paths. (All three are used in `MainTest`, `config_dir` is used in `RenewalTest`, and `logs_dir` is used as the _only_ temp directory in `SetupLogFileHandlerTest` and `SetupLoggingTest`.)

### DIFF
--- a/certbot/tests/cert_manager_test.py
+++ b/certbot/tests/cert_manager_test.py
@@ -18,11 +18,14 @@ from certbot.storage import ALL_FOUR
 from certbot.tests import storage_test
 from certbot.tests import util as test_util
 
-class BaseCertManagerTest(unittest.TestCase):
+from certbot.tests.util import TempDirTestCase
+
+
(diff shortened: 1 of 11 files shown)


certbot/tests/cert_manager_test.py
```diff
@@ -69,7 +72,7 @@ def _set_up_config(self, domain, custom_archive):
         return config
 
     def tearDown(self):
-        shutil.rmtree(self.tempdir)
+        super(BaseCertManagerTest, self).tearDown()
```
bmw: I think we can just delete the definition of `tearDown` here as it's defined in the base class.

PR #4303, opened 2017-03-06
Unrevert new_authzr_uri removal

jsha:
In #4277 we reverted the removal of new_authzr_uri because it would have broken people who use both old and new clients on the same machine. This change unreverts that change, and adds an additional change to store the new_authz URL from the directory, so older clients can work with recently stored accounts.

### DIFF
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -320,7 +320,7 @@ class RegistrationResource(ResourceWithURI):
 
     """
     body = jose.Field('body', decoder=Registration.from_json)
-    new_authzr_uri = jose.Field('new_authzr_uri')
+    new_authzr_uri = jose.Field('new_authzr_uri', omitempty=True)
     terms_of_service = jose.Field('terms_of_service', omitempty=True)
 
 
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -429,7 +429,7 @@ class AuthorizationResource(ResourceWithURI):
 
     """
     body = jose.Field('body', decoder=Authorization.from_json)
-    new_cert_uri = jose.Field('new_cert_uri')
+    new_cert_uri = jose.Field('new_cert_uri', omitempty=True)
 
 
 @Directory.register
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -315,10 +315,12 @@ class RegistrationResource(ResourceWithURI):
     """Registration Resource.
 
     :ivar acme.messages.Registration body:
+    :ivar unicode new_authzr_uri: Deprecated. Do not use.
     :ivar unicode terms_of_service: URL for the CA TOS.
 
     """
     body = jose.Field('body', decoder=Registration.from_json)
+    new_authzr_uri = jose.Field('new_authzr_uri')
     terms_of_service = jose.Field('terms_of_service', omitempty=True)
 
 
@@ -423,9 +425,11 @@ class AuthorizationResource(ResourceWithURI):
     """Authorization Resource.
 
     :ivar acme.messages.Authorization body:
+    :ivar unicode new_cert_uri: Deprecated. Do not use.
 
     """
     body = jose.Field('body', decoder=Authorization.from_json)
+    new_cert_uri = jose.Field('new_cert_uri')
 
 
 @Directory.register
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -191,7 +191,7 @@ class Directory(jose.JSONDeSerializable):
         try:
             return self[name.replace('_', '-')]
         except KeyError as error:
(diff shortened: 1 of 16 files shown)


ohemorange:
Is `acme` being passed around to make the tests happy?

jsha:
No, acme is passed around so that AccountFileStorage can ask for the new_authz URL. This felt cleaner than passing around the specific URL.

ohemorange:
Is there any clean way to test the using-old-version thing? Like making sure that the old loading code that doesn't exist reads things correctly that are written out by the new version?

jsha:
I added a unittest that parses and directly looks for that specific field. What do you think? Seems like the closest we'll get without copying back in all the old code.

ohemorange:
I think ideally I'd like to have the old `RegistrationResource` around and create that object, but yeah, that's probably too much and this is good. I'm good with this, but I think @bmw should probably confirm because I also thought it was good last time.

jsha:
We are inevitably going to have to change the public API of `acme` for the next spec version. I've been meaning to post an issue discussing what that should look like. But I agree that it's probably better to batch up the changes in a bigger chunk.

> As for the public methods that take new_authzr_uri, I think we should keep them as is. If you provide the value we'll use it, otherwise, we'll use the value in the directory. Alternatively, we could ignore it, but I prefer the current approach.

I think the two resource types should retain the `new_*_uri` fields for backwards compatibility, but mark them as deprecated / ignored.

The place where we actually use these URIs is in Client.request_challenges and Client.request_issuance. I think those spots should treat the directory URIs as canonical (they are), and ignore the `new_*_uri` fields on authzs and regs. Thoughts?

bmw:
>We are inevitably going to have to change the public API of acme for the next spec version. I've been meaning to post an issue discussing what that should look like. But I agree that it's probably better to batch up the changes in a bigger chunk.

Yeah I think there's a lot of discussion to be had there and I agree breaking changes should be shipped in larger chunks. Honestly, we're currently in a bit of an awkward spot. Although we're still in 0.x.y so our [API can change at any time](http://semver.org/#spec-item-4) this isn't how the Certbot team or the world has been treating the project.

> I think the two resource types should retain the new_*_uri fields for backwards compatibility, but mark them as deprecated / ignored.
> The place where we actually use these URIs is in Client.request_challenges and Client.request_issuance. I think those spots should treat the directory URIs as canonical (they are), and ignore the new_*_uri fields on authzs and regs. Thoughts?

I agree. What do you think we should do with `new_authzr_uri` values passed into `request_challenges()` and `request_domain_challenges()`?

jsha:
> I agree. What do you think we should do with new_authzr_uri values passed into request_challenges() and request_domain_challenges()?

Ignore them; possibly log a deprecation warning.

bmw:
Works for me. Logging a debug level warning seems like a good idea.

certbot/main.py
```diff
@@ -466,7 +466,7 @@ def register(config, unused_plugins):
     # We rely on an exception to interrupt this process if it didn't work.
     acc.regr = acme_client.acme.update_registration(acc.regr.update(
         body=acc.regr.body.update(contact=('mailto:' + config.email,))))
-    account_storage.save_regr(acc)
+    account_storage.save_regr(acc, acme_client)
```
bmw: TIL `acme_client` isn't an `acme.client`. It's `certbot.client` and the `acme` client is currently `acme_client.acme` :/

This code currently blows up because `acme_client` doesn't have a directory object. We should change the name of `acme_client`, but it's up to you whether or not we do this in this PR. If you want, you can just use `acme_client.acme` here and I'll make sure this gets fixed.

acme/acme/messages.py
```diff
@@ -315,7 +315,7 @@ class RegistrationResource(ResourceWithURI):
     """Registration Resource.
 
     :ivar acme.messages.Registration body:
-    :ivar unicode new_authzr_uri: URI found in the 'next' ``Link`` header
+    :ivar unicode new_authzr_uri: Deprecated. Do not use.
```
bmw: I don't think this matters, however, leaving the `new_*_uri` fields unchanged means they're still included in the serialized JSON. For instance:
```
{'body': {'key': ..., 'new_authzr_uri': None, 'terms_of_service': ...
```
I'm not sure if this is a problem since I don't believe we ever send the full registration resource or authorization resource to the ACME server, but perhaps it should be fixed anyway. This can be done by setting `default=None` and `omitempty=True` when defining the field as described at https://github.com/certbot/certbot/blob/master/acme/acme/jose/json_util.py#L41.

I'll defer to your judgement of whether or not you think we should do this.

jsha:
Nope, I missed `new_cert_uri`. Thanks for pointing it out; fixed.

Issue #3455, opened 2016-08-25
Build our wheels properly


bmw:
It turns out we're not building our wheels for PyPI correctly. Due to the way we're adding conditional dependencies, the dependencies that get added to the wheel are only those that were required on the system that built the wheel. For example, all of our `setup.py` files contains code like this:

```
if sys.version_info < (2, 7): 
    install_requires.extend([
        'argparse',
        'mock<1.1.0',
    ])  
else:
    install_requires.append('mock')
```

We have always built our wheels on Python 2.7, however,  which means our wheels will never install `argparse` or `mock<1.1.0`. This problem doesn't affect `certbot-auto`, but affects people installing through `pip`.

Fixing this problem without breaking people on old versions of `setuptools`/`pip` will be tricky (or impossible).

erikrose:
For completeness, I'll mention [environment markers](https://www.python.org/dev/peps/pep-0496/), which would solve the problem in the event we are unable to dispense with the conditonals. But, as you hint, those work only in pip >= 6.0.

bmw:
Another option I've seen is to use `setup.cfg` and `metadata`. Although this is technically depreciated, it is still used by a number of very popular and maintained Python packages such as [twine](https://github.com/pypa/twine).

erikrose:
> Another option I've seen is to use setup.cfg and metadata.

Yes, those are another place to put environment markers. They can also go right in setup.py. The user would need pip>=6.0 [or setuptools>=20.6.0](https://setuptools.readthedocs.io/en/latest/history.html#v20-6-8) to respect them.

PR #4313, opened 2017-03-09
Build wheels correctly for Python 2.6

### LINKED ISSUE #2176: Dynamically computed install_requires: investigate potential problems
In https://github.com/letsencrypt/letsencrypt/pull/1661#discussion_r46522959 @mgedmin said:

> Hmm. Dynamically-computed install_requires causes problems for users of pip >= 7.0, which has wheel caching enabled by default: pypa/pip#3025. Short version:
> - I pip install letsencrypt on Python 2.7, pip runs setup.py and builds a wheel (marked -py2-none-any.whl) and puts it in ~/.cache/pip/
> - I pip install letsencrypt on Python 2.6, pip discovers a valid wheel and installs it, using install_requires computed earlier for a different Python version."
#2175 might be related

bmw:
With #2249 merged, this fixes #2176 and #3455. We pinned <1.1.0 during the brief period between `mock 1.1.0` and `mock 1.1.4` where they didn't support Python 2.6 (see their [changelog](https://mock.readthedocs.io/en/latest/changelog.html#id4)).

@erikrose, can you review this?

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -13,6 +13,7 @@ install_requires = [
     # rsa_recover_prime_factors (>=0.8)
     'cryptography>=0.8',
     # Connection.set_tlsext_host_name (>=0.13)
+    'mock',
     'PyOpenSSL>=0.13',
     'pyrfc3339',
     'pytz',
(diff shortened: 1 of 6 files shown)


bmw:
@ohemorange, can you give this the final +1 it needs to be merged?

PR #4316, opened 2017-03-09
Rewrite our Dockerfile pt 2

### LINKED ISSUE #3782: Rewrite our Dockerfile

bmw:
Fixes #3782 and builds on #4032.

Let's see if GH lets me merge this PR. Original PR was #4295. I'll need another approving review.


bmw:
I filed a help ticket with GH about status checks.

Issue #3998, opened 2017-01-09
cerbot rename should rename all relevant files


ohemorange:

ohemorange:
Oh yeah, just remembering why we didn't do this in the first place:

If someone's manually gone in and renamed the appropriate files and updated the renewal conf file and updated the symlinks, do we blow that all away by matching all the names to the new cert name?

I think we should.

bmw:
In the normal case, we should rename the renewal conf file, the directory in archive, and the directory in live. I don't have a strong opinion on what we do in the case when things have inconsistent names. Some options are;

1. Change all the things.
2. Change only renewal conf file (in which case reinstallation doesn't have to happen).
3. Allow the user to choose with a prompt and/or command line flags.

To me, 3 feels the most correct™, but I'm fine with the simpler approach of 1 or 2 for now if you prefer.

ohemorange:
Or, probably best but most annoying:

4. Just change them, but then print "we changed all the things! run <command> to undo these changes."

bmw:
5. "We only changed your renewal conf file, run again with `--rename-all` to rename all the things"

ohemorange:
This really isn't actually a commonly requested feature, and honestly doesn't seem worth the work it would take to get it working correctly.

PR #4320, opened 2017-03-10
More thoroughly rename during `certbot rename`.

ohemorange:
Addresses #3998.


ohemorange:
Fun fact! Because the plugin selector checks that the config files are valid, we now get a lineage duplicator for free. Congratulations us!

ohemorange:
Ready for review! It probably only needs one person -- @ynasser, @bmw, do either of you have time?

ynasser:
I can take a look within the next few days, but am not sure when. Definitely before end of day on Monday.

ohemorange:
Hm, just noticed that coveralls is confused about what changed in this PR, so I'm ignoring it yelling at me for bad coverage.

PR #4322, opened 2017-03-10
IChallengeProvider interface added - Domain challenge serialization

ph4r05:
[Feedback required]
AuthHandler asks ACME directly for domain challenges. With this new interface developers can augment objects implementing `IAuthenticator` so the authenticator plugin is able to restore serialised domain challenges. In this way Certbot can validate domains across multiple Certbot invocations which is suitable e.g., for Ansible (e.g. DNS domain validation with Ansible DNS modules. 2 Certbot invocations are needed. See linked issue for more info).

We are developing external auth plugin for the Certbot which can really use this plugin extension in the Certbot.

If the PR is merged, the idea with our plugin is the following:

 - Execute `./certbot <arguments> --configurator plugin --to-session /tmp/mydomain.json`
   - The first phase of the validation. 
   - Requests challenges from ACME, stores them to `/tmp/mydomain.json` - serialized state
   - Outputs JSON to Stdout so Ansible can parse the output
 - Update DNS in Ansible task
 - Execute `./certbot <arguments> --configurator plugin --from-session /tmp/mydomain.json`
   - The second phase of the validation.
   - Deserializes domain challenges from the json file instead of generating new ones on ACME.
   - Finishes the process.

We tested this domain challenge serialization/deserialization approach and it works smoothly. The only thing is we need a hook in the Certbot to either use ACME or the auth plugin to load challenges.

Update: Another way to implement this with lower change footprint is to have only one method in the `IChallengeProvider`.  If Authenticator implements also `IChallengeProvider` the `AuthHandler` would call the plugin to obtain challenges. The downside is the plugin would need to call ACME plugin as a fallback if it cannot find the domain challenge in the storage.

Thanks for considering!

### DIFF
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -5,6 +5,9 @@ import time
 import six
 import zope.component
 
+from zope.interface.exceptions import DoesNotImplement
+from zope.interface.verify import verifyObject
+
 from acme import challenges
 from acme import messages
 
@@ -62,9 +65,7 @@ class AuthHandler(object):
             authorizations
 
         """
-        for domain in domains:
-            self.authzr[domain] = self.acme.request_domain_challenges(domain)
-
+        self._request_challenges(domains)
         self._choose_challenges(domains)
 
         # While there are still challenges remaining...
@@ -87,6 +88,44 @@ class AuthHandler(object):
                 "Challenges failed for all domains")
 
         return retVal
+
+    def _request_challenges(self, domains):
+        """Request challenges for domains, uses acme or the plugin"""
+
+        # If plugin does not implement challenge provider, ask ACME directly.
+        try:
+            return self._request_challenges_authenticator(domains)
+        except DoesNotImplement:
+            return self._request_challenges_directly(domains)
+
+    def _request_challenges_authenticator(self, domains):
+        """Uses auth plugin to load domain challenges.
+
+        :raises DoesNotImplement if self.auth does not implement
+        IChallengeProvider
+        """
+        verifyObject(interfaces.IChallengeProvider, self.auth)
+
+        # Ask provider for challenge, if there is none (NoChallengeError)
+        # ask ACME, then inform challenge provider
+        for domain in domains:
+            try:
+                self.authzr[domain] = self.auth.request_domain_challenges(
+                    domain, self.acme, self.account)
+
+            except errors.NoChallengeError:
+                self.authzr[domain] = self._request_domain_challenge(domain)
+                self.auth.on_domain_challenge_loaded(
+                    domain, self.authzr[domain])
+
+    def _request_challenges_directly(self, domains):
+        """Uses ACME client to request domain challenges"""
+        for domain in domains:
+            self.authzr[domain] = self._request_domain_challenge(domain)
+
+    def _request_domain_challenge(self, domain):
+        """Asks ACME for domain challenge"""
+        return self.acme.request_domain_challenges(domain)
 
     def _choose_challenges(self, domains):
         """Retrieve necessary challenges to satisfy server."""
--- a/certbot/errors.py
+++ b/certbot/errors.py
@@ -81,6 +81,10 @@ class NotSupportedError(PluginError):
     """Certbot Plugin function not supported error."""
 
 
+class NoChallengeError(PluginError):
+    """Certbot Plugin function could not restore challenge."""
+
+
 class StandaloneBindError(Error):
     """Standalone plugin bind error."""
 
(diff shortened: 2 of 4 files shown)


dancvrcek:
Hi @ohemorange,

I wonder if you've decided yet whether to accept this PR or not. 

Cheers
  Dan

ohemorange:
Hello! Glad to see the excitement to integrate Certbot.

Reading through the use case (but not having used Ansible myself), this sounds to me very similar to what the manual plugin allows: https://certbot.eff.org/docs/using.html#hooks. It's only a few months old, so I don't know if you've seen it before? It sounds like a script calling Ansible would be a perfect candidate for `--manual-auth-hook`. This might not make sense for the way Andible works, but as mentioned in the documentation, we should provide all of the information a DNS-updating script might need, and then everything could be run in a single step.

Alternatively, if there's a fully-featured plugin, why can't it call into Ansible to make the DNS updates, then wait for propagation? As in #4522, #4448, #4423, and #4360 (which are still in-progress, but demonstrate how a plugin could update DNS).

ph4r05:
Thanks for getting back!

We are aware of manual plugin, it works pretty well in one pass scenario. We also implemented something like this. But regarding usage with Ansible we got this issue ticket:
https://github.com/EnigmaBridge/certbot-external-auth/issues/3

It basically says Ansible already has a lot of plugins for DNS auth so the idea was to use this Ansible capability to perform DNS validation. As I understand it, it cannot be done in one pass as certbot cannot simply invoke Ansible DNS plugin. Thats the reason we would need to serialise challenges and finish the validation later. The issue author suggested something like this:

```
- shell: certbot -d example.com <more args>
  register: result
  set_fact:
      certbot_out_json: "{{ result.stdout | from_json }}"

- your_cloudprovider:
      dns:
          domain: _certbot-challenge.example.com
          type: txt
          value: "{{ certbot_out_json.token }}"

- shell: dig -t txt _certbot-challenge.example.com
  retry: 5
  delay: 60

- shell: certbot --resume --session-id {{ certbot_out_json.session_id }} <more args>
```

ohemorange:
Ah, I see. If the goal is solely to be able to use Ansible DNS plugins, that won't be needed, as Certbot is soon to release its own DNS plugins (see PRs linked above). Given that the functionality will already be covered, it will be unnecessary to support this additional complexity, and users will be able to call a single pass of Certbot that does the DNS validation internally.

dancvrcek:
It's an excellent news that you decided to extend certbot with a number of DNS plugins as it certainly improves usability.

It is your decision and yours only and we respect that but please if you can spare a few minutes of your time, try to consider the following train of thoughts.

I wonder why you want to take on the responsibility for secure handling of API keys and passwords for access to DNS servers. These can already be armoured in specialized processes or even servers or access-restricted, subject to 4-eyes / dual-control approvals. 

I understand that access to DNS records can be severely restricted to prevent brutal attacks against company networks. I'm glad this is not my decision as I would be thinking long and hard if I really want users lock-in into this.

The change, which @ph4r05 proposes basically just handles the challenge generation exception - thus allowing to use certbot in a 2-step process. While it looks like a lot of lines, most of that is for aesthetic reasons and the core change is about 5 lines of code if written economically. (lines 117-119 in auth_handler.py)

This seems to be one of those cases when usability v security go head-to-head. I just think there doesn't have to be only one winner.

ohemorange:
You make a great point! We of course think that security is important.

But we don't want to commit to maintaining code that is not widely used, so I'm hesitant to include something like this in Certbot unless it's something a large part of our community will make use of. Allowing multiple runs of Certbot for a single issuance is outside of our current model.

The current hooks should allow a lightweight DNS client to talk to a separate process, if that's a feature you'd like to include. If writing a specialized DNS plugin doesn't give the level of customization necessary, I might recommend taking a look at the list of alternate ACME clients: https://letsencrypt.org/docs/client-options/

dancvrcek:
We can't promise wide use, unfortunately. Although, @ph4r05 is getting some traction at Serverfault with our DNS Certbot plugin (https://serverfault.com/questions/750902/how-to-use-lets-encrypt-dns-challenge-validation - the accepted answer there).

I personally like the fact that DNS validation doesn't interfere with the configuration of an actual server.  The downside is that DNS is a rather big hammer.

Well, please let us know your decision at your convenience so we can amend our plugin setup.

ohemorange:
We are unfortunately unable to support these changes in core Certbot. If the upcoming DNS plugins don't meet your needs, I'd recommend achieving the desired functionality by writing a specialized DNS management plugin using the existing plugin APIs.

PR #4325, opened 2017-03-10
Improve Account.repr()

jsha:
The old version had two problems:
 - It printed `<Account(bca5889f66457d5b62fbba7b25f9ab6f)>`, which showed the id, but looked like it was showing a pointer of some sort.
 - It failed to print the meta and regr fields.

This meant that, for instance, when a test failed because two Accounts differed, the test output would not contain sufficient information to see where they differed.

### DIFF
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -75,7 +75,7 @@ class Account(object):  # pylint: disable=too-few-public-methods
 
     def __repr__(self):
         return "<{0}({1}, {2}, {3})>".format(
-            self.__class__.__name__, self.id, self.regr, self.meta)
+            self.__class__.__name__, self.regr, self.id, self.meta)
 
     def __eq__(self, other):
         return (isinstance(other, self.__class__) and
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -74,7 +74,8 @@ class Account(object):  # pylint: disable=too-few-public-methods
             self.meta.creation_dt), self.meta.creation_host, self.id[:4])
 
     def __repr__(self):
-        return "<{0}({1})>".format(self.__class__.__name__, self.id)
+        return "<{0}({1}, {2}, {3})>".format(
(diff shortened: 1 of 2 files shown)


certbot/account.py
```diff
@@ -74,7 +74,8 @@ def slug(self):
             self.meta.creation_dt), self.meta.creation_host, self.id[:4])
 
     def __repr__(self):
-        return "<{0}({1})>".format(self.__class__.__name__, self.id)
+        return "<{0}({1}, {2}, {3})>".format(
+            self.__class__.__name__, self.id, self.regr, self.meta)
```
bmw: nit: As you probably already know, normally it's good practice to make the `repr` of an object be as close to the constructor call necessary to recreate that object as possible. While we don't want to do exactly that here because the `repr` would include the private key, we can keep the order the same as the constructor by swapping `id` and `regr` here.

certbot/tests/account_test.py
```diff
@@ -52,10 +52,7 @@ def test_slug(self):
             self.acc.slug, "test.certbot.org@2015-07-04T14:04:10Z (bca5)")
 
     def test_repr(self):
-        self.assertEqual(
-            repr(self.acc),
-            "<Account(bca5889f66457d5b62fbba7b25f9ab6f)>")
-
+        self.assertTrue("<Account(bca5889f66457d5b62fbba7b25f9ab6f" in repr(self.acc))
```
bmw: If you make the other requested change, this will have to be changed slightly.

bmw:
Looks like lint is failing.

bmw:
I think my comment earlier about lint masked another problem here. Python 3.6 tests are also failing because the order of output of `repr(Meta)` changed.

I don't care too much how you fix this but you can either just check some of the `repr()` output rather than the exact thing, change the expected output based on Python version (if it's consistent), also mock out `meta.__repr__`, or modify the `Meta` class so `repr` output is consistent.

PR #4327, opened 2017-03-11
Improve macOS contributor instructions

bmw:


docs/contributing.rst
```diff
@@ -343,56 +352,20 @@ This should generate documentation in the ``docs/_build/html``
...
-
-.. code-block:: shell
+macOS users will probably find it easiest to use Docker Compose for
+development. To install Docker Compose on macOS, follow the instructions at
+https://docs.docker.com/docker-for-mac/install/.
 
-  docker run -it certbot bash
-  cd src
-  tox -e py27
+Now you can develop locally and run Certbot and test your changes in Docker.
```
ohemorange: Now you can develop on your host machine, but run Certbot and test your changes in Docker.

docs/contributing.rst
```diff
@@ -343,56 +352,20 @@ This should generate documentation in the ``docs/_build/html``
...
+For example, from anywhere inside your clone of the Certbot repository you can
+run the following to check for linting errors::
 
+  docker-compose run --rm --service-ports development bash -c 'cd src;tox -e lint'
```
ohemorange: docker-compose creates a docker image that spits you out into a top-level where there's a subdirectory src? maybe we want to note this? does docker-compose have a reference to the repo on the host, or is copied somewhere when you create an image? so, could you run with run and exec to do things interactively?

ohemorange: And have we fixed `ERROR: In file './../docker-compose.yml' service 'version' doesn't have any configuration options. All top level keys in your docker-compose.yml must map to a dictionary of configuration options.`?

jsha: The second thing is really "your docker-compose is too old."

bmw: > docker-compose creates a docker image that spits you out into a top-level where there's a subdirectory src? maybe we want to note this?

I'd ideally just change it so you're automatically in `src` which Jacob and I discussed earlier in IRC. I'll make that one line change.

> does docker-compose have a reference to the repo on the host, or is copied somewhere when you create an image? so, could you run with run and exec to do things interactively?

The repo on the host is mounted inside of the Docker container so you can do things interactively. What would you like to see added here?

bmw: @ohemorange, to run this on your laptop, `pip install docker-compose` in your Certbot `venv` and then run the command with `sudo` (keeping in mind `venv/bin/activate` and `sudo` don't place nice).

ohemorange: If the version of docker-compose shipped from the os package manager gets this error, we should probably mention running `pip install`.

ohemorange: > The repo on the host is mounted inside of the Docker container so you can do things interactively. What would you like to see added here?

An example of the command to do this.

bmw: Hmm. I was treating this section as how to do things on macOS like it was before but I could expand on it and give Linux instructions.

bmw: > An example of the command to do this.

Basically just how to get a shell in the Docker container?

ohemorange: Running a shell wouldn't work from mac?

Yeah.

bmw: OK. Ready for another review. If you want to test that you're now in the source directory but have already ran `docker-compose` on your machine, you need to run `docker-compose build` to get the updates.

PR #4329, opened 2017-03-11
Improve plugin-writing docs.

jsha:
Move "Writing your own plugin" under Code components and layout, with the other
plugin docs.

Include instructions on how to install a plugin into a virtualenv and how to check
for its presence.

Document that users can install third-party plugins systemwide, but not with
certbot-auto.

Remove obsolete information from Authenticators section and make the section
more informative.

Remove IDisplay sub-section since it repeats information in the main "Plugin
architecture" section.


jsha:
Incidentally, in writing this, I came to wonder: Where does the generated documentation from all the inline pydoc stuff wind up? Do we have it on certbot.eff.org?

PR #4333, opened 2017-03-13
If we fail to reload, try starting Nginx once without catching output

### LINKED ISSUE #4324: Using the Nginx plugin when Nginx is turned off hangs forever?

Arch Linux

OS package manager

`certbot --nginx -vv -d domain.net -d blog.domain.net -d cf.domain.net -d mail.domain.net`

It hung for 5 - 10 minutes without printing any output.

http://pastebin.com/w29rj9F7

## Other info:
Nginx wasn't running when running Certbot. Nginx version is 1.11.10. We should investigate this issue before unhiding Nginx. If we struggle to reproduce, we can kick/close it.

cc @ohemorange

ohemorange:
Because a successful start with piped output hangs on Arch.

Fixes #4324.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -829,10 +829,8 @@ def nginx_restart(nginx_ctl, nginx_conf="/etc/nginx.conf"):
 
     """
     try:
-        proc = subprocess.Popen([nginx_ctl, "-c", nginx_conf, "-s", "reload"],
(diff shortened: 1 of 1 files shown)


ohemorange:
travis...?

bmw:
Turn it off and on again again

bmw:
:frowning_face:

In the past I've had waiting a minute between opening and closing the PR fix the problem. No luck here though.

ohemorange:
Confirmed that it fixes the bug on Arch.

PR #4340, opened 2017-03-14
Add url and kid to jws.

jsha:
This will be required in order to implement the latest ACME spec, which uses
these protected header fields.


jsha:
I think @zjs' suggestion is good, and plan to implement it.

bmw:
OK great. Ping me when you want another review.

jsha:
Alright, added.

PR #4347, opened 2017-03-15
Fixes nginxparser to allow multiline quoted strings

kernelpanek:
In the [current nginxparser module](https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/nginxparser.py#L32-L33), single and double quoted strings are parsed with:

```python
dquoted = Regex(r'(\".*\")')
squoted = Regex(r"(\'.*\')")
```

However, the parser will miss quoted strings that span several lines (e.g. when using Openresty directives and having inline Lua code, that code can span multiple lines for clarity.)  If the module parses quoted strings with Pyparsing's `QuotedString`, then parsing is able to handle more unwieldy (and sometimes auto-generated) nginx configuration files. This PR proposes a 2 line change to the nginxparser module:

```python
dquoted = QuotedString('"', multiline=True, unquoteResults=False)
squoted = QuotedString("'", multiline=True, unquoteResults=False)
```

I preemptively ran these changes through [TravisCI and everything passes](https://travis-ci.org/kernelpanek/certbot). I hope this is helpful!

### DIFF
--- a/certbot-nginx/certbot_nginx/nginxparser.py
+++ b/certbot-nginx/certbot_nginx/nginxparser.py
@@ -29,8 +29,8 @@ class RawNginxParser(object):
     # any chars in single or double quotes
     # All of these COULD be upgraded to something like
     # https://stackoverflow.com/a/16130746
-    dquoted = QuotedString('"', multiline=True)
-    squoted = QuotedString("'", multiline=True)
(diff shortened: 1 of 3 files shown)


ohemorange:
Great PR and tests, thanks! We're also doing a rewrite of the parser at the moment (#4296), so if the code looks different soon that's why, but definitely incorporating this!

Issue #4353, opened 2017-03-16
Make our user agent strings more informative


pde:
There are a few things that might be quite useful to have in our user agent:
* Is the user running `cerbot-auto` or not?
* Is this a renewal or not?
* What verb did the user pick?
* Python version (conceding that #2008 might have been a good idea, because we can't tell which users on old OSes have installed a newer python etc)

PR #4354, opened 2017-03-16
Add extra information to user-agent

pde:
* Are we auto or not?
* What verb?
* What python version?

Not yet added, because it may be too tricky:
* Are we renewing or not?


pde:
Closes: #4354

pde:
Closes: #4353

certbot/client.py
```diff
@@ -52,9 +53,12 @@ def determine_user_agent(config):
     """
 
     if config.user_agent is None:
-        ua = "CertbotACMEClient/{0} ({1}) Authenticator/{2} Installer/{3}"
+        auto = "no" if cli.cli_command == "certbot" else "yes"
```
hlieberman: This also leaves out people using certbot via the old letsencrypt symlink.

pde: @hlieberman I believe that `cli.cli_command` would be set to `"letsencrypt-auto"` for those people, so for them this variable would be `"yes"`.

certbot/client.py
```diff
@@ -52,9 +53,12 @@ def determine_user_agent(config):
...
+        ua = ("CertbotACMEClient/{0} ({1}) Authenticator/{2} Installer/{3} Verb/{4} "
```
zjs: The format for `User-Agent` strings is defined in [RFC 1945](https://www.ietf.org/rfc/rfc1945.txt) Section 10.15.

```
   The User-Agent request-header field contains information about the
   user agent originating the request. This is for statistical purposes,
   the tracing of protocol violations, and automated recognition of user
   agents for the sake of tailoring responses to avoid particular user
   agent limitations. Although it is not required, user agents should
   include this field with requests. The field can contain multiple
   product tokens (Section 3.7) and comments identifying the agent and
   any subproducts which form a significant part of the user agent. By
   convention, the product tokens are listed in order of their
   significance for identifying the application.

       User-Agent     = "User-Agent" ":" 1*( product | comment )
```

With `product` defined in Section 3.7.

```
   Product tokens are used to allow communicating applications to
   identify themselves via a simple product token, with an optional
   slash and version designator. Most fields using product tokens also
   allow subproducts which form a significant part of the application to
   be listed, separated by whitespace. By convention, the products are
   listed in order of their significance for identifying the
   application.

       product         = token ["/" product-version]
       product-version = token

   Examples:

       User-Agent: CERN-LineMode/2.15 libwww/2.17b3
```

And `comment` defined in Section 2.2.

```
   Comments may be included in some HTTP header fields by surrounding
   the comment text with parentheses. Comments are only allowed in
   fields containing "comment" as part of their field value definition.
   In all other fields, parentheses are considered part of the field
   value.

       comment        = "(" *( ctext | comment ) ")"
       ctext          = <any TEXT excluding "(" and ")">
```

While the authenticator and installer seem like they might the qualify as "subproducts", verb, renewing, and auto don't. It seems like comment(s) should be used for those instead.

For Python, we should look for existing conventions.

certbot/client.py
```diff
@@ -52,9 +53,12 @@ def determine_user_agent(config):
...
+              "Renewing/{5} Auto/{6} Py/{7}")
```
zjs: We should consider also including information about the OS.

certbot/client.py
zjs: As a comment on the authenticator, we should consider capturing the type (currently always `dns`, but that may change in the future).

certbot/client.py
pde: That should already be present (field `{1}`), unless you mean something else?

certbot/renewal.py
```diff
@@ -393,6 +393,7 @@ def handle_renewal_request(config):
         # elements from within the renewal configuration file).
         try:
             renewal_candidate = _reconstitute(lineage_config, renewal_file)
+            renewal_candidate.renewing = "yes"
```
zjs: This shouldn't be included, should it?

certbot/cli.py
```diff
@@ -532,6 +532,7 @@ def parse_args(self):
         parsed_args = self.parser.parse_args(self.args)
         parsed_args.func = self.VERBS[self.verb]
         parsed_args.verb = self.verb
+        parsed_args.renewing = "unknown"  # an important config property we don't know until later
```
zjs: Maybe mention that this is used for the UA string instead of just that it's important?

certbot/cli.py
```diff
@@ -1103,8 +1104,9 @@ def _create_subparsers(helpful):
     helpful.add(
         None, "--user-agent", default=None,
         help="Set a custom user agent string for the client. User agent strings allow "
-             "the CA to collect high level statistics about success rates by OS and "
-             "plugin. If you wish to hide your server OS version from the Let's "
+             "the CA to collect high level statistics about success rates by OS, "
+             "plugin and use case, and to know when to deprecate support for past Python "
+             "versions. If you wish to hide this information from the Let's "
```
zjs: This is worded kind of inaccurately.

The UA string does allow the CA to collect usage information, but it's the client developers who make decisions about things like which python versions to support (not the CAs).

certbot/client.py
```diff
@@ -53,9 +54,15 @@ def determine_user_agent(config):
     """
 
     if config.user_agent is None:
-        ua = "CertbotACMEClient/{0} ({1}) Authenticator/{2} Installer/{3}"
-        ua = ua.format(certbot.__version__, util.get_os_info_ua(),
-                       config.authenticator, config.installer)
+        if cli.cli_command in ["certbot", "letsencrypt", "certbot-auto", "letsencrypt-auto"]:
+            auto = cli.cli_command
+        else:
+            auto = "other"
```
zjs: Looking at the code which sets `cli_command` in `cli`, it seems like maybe looking for the presence of `CERTBOT_AUTO` in `os.environ` might help provide more information in the case where the `cli_command` isn't a value we expect (e.g., "other (auto)" vs. "other").

certbot/cli.py
pde: Let's Encrypt shares statistical data with EFF to help us improve Certbot and boulder, so in practice it's not inaccurate.

certbot/client.py
pde: If we want to improve this, we could do so in the definition of `cli.cli_command` itself. But that feels like it should be another PR?

pde: (I think the benefits would be positive but extremely small, essentially limited to cases where a `certbot-auto` user sets `VENV_NAME` or `XDG_DATA_HOME`)

certbot/cli.py
pde: I've explained what it's used for *and* removed the claim that it's important :)

pde:
@zjs this should be back in your court now

certbot/main.py
```diff
@@ -215,6 +215,10 @@ def _find_cert(config, domains, certname):
               RenewableCert instance or None.
     """
     action, lineage = _find_lineage_for_domains_and_certname(config, domains, certname)
+    if action == "renew":
```
ohemorange: `_find_cert` should not have the side effect of modifying `config`, towards the goal of reducing the statefulness of Certbot's architecture. If we want to know for sure whether this is a renewal or certonly in practice, we can remove the machinery that asks "looks like you have a cert with these domains already, want to renew instead?". And then we'll know for sure, because there will be a single supported way to renew vs get a new cert.

zjs: > we can remove the machinery that asks "looks like you have a cert with these domains already, want to renew instead?"

What if a user of `certbot-auto` is relying on this behavior for a command in their crontab?

pde: @ohemorange that logic was added because it greatly improved user experience; I don't think removing it is a good idea.

I agree it's good to minimise state, but not at the expense of functionality.

pde: @ohemorange my instinct is that we should land this as-is. We could consider whether it's sensible to have a wishlist refactor ticket for "move all stateful information out of `config` into another object", but the set of things we'd need to move would be larger than this one variable -- it would I think include every config variable that the user can set interactively and which persists for renewal (installer, authenticator, webroot map, etc)

certbot/main.py
```diff
@@ -690,6 +690,7 @@ def certonly(config, plugins):
 def renew(config, unused_plugins):
     """Renew previously-obtained certificates."""
     try:
+        config.renewing = "yes"
```
ohemorange: This method is called iff the `renew` verb is used, so this line is currently redundant and `config.renewing` can be removed entirely, and only `config.verb` inspected.

certbot/cli.py
```diff
@@ -532,6 +532,7 @@ def parse_args(self):
         parsed_args = self.parser.parse_args(self.args)
         parsed_args.func = self.VERBS[self.verb]
         parsed_args.verb = self.verb
+        parsed_args.renewing = "unknown"  # used for the User Agent; not known until later
```
ohemorange: This should probably be removed, since we don't currently use it.

certbot/client.py
ohemorange: Why throw out the data of what `cli.cli_command` is here, rather than in post-processing? Better to have the data then throw it out later, no?

pde: My thinking was just that if the user moves our command to some other weird name like `certbot-sucks` or whatever, maybe that's private? But I guess it's corner case-y, and documented in the dummy UA, and the code is simpler without this logic.

ohemorange:
@pde not sure if you saw my last round of comments?

pde:
@ohemorange I suggest a non-squashing merge for this to give Alex credit for starting this work :)

PR #4357, opened 2017-03-16
Call certbot client cb_client rather than acme_client

bmw:
In some sense, `certbot.client.Client` is an ACME client, but it's the not the client in the ACME library and this leads to confusion. Let's make what this is clear.


bmw:
Coveralls isn't responding. Closing and reopening...

Issue #1279, opened 2015-11-01
certbot-auto and additional plugins


kuba:
I would like to see something along the lines of `LE_PIP_PKGS="letsencrypt-nginx git+https://github.com/plesk/letsencrypt-plesk" ./letsencrypt-auto ---configurator plesk`

zjs:
I've taken a look at this ticket, PR #4047, tried to understand the current logic in `letsencrypt-auto.template`, and tried to think about use cases for plugins.

It seems like being able to nuke the `venv` is nice. Starting from a "clean slate" each time makes things easier to test, debug, and reason about. Less persistent state is, I think, preferable.

It also seems like a "one size fits all" solution for plugins may not be tractable, especially if we have a mix of first-party, third-party, and user-specific plugins.

Given this, I propose what I think would be a relatively small change:

Allow people to put scripts in a well-known location relative to the `certbot-auto` script (e.g., `certbot-auto.d/` or `certbot-auto-plugins/`). During phase2, between installing and running, run each script.

To be compatible with `certbot-auto`, plugin authors would simply need to define such a script to install the plugin into the `venv`. This could be anything from a call to `pip` to a command to copy files from elsewhere on the system into the `venv`, whatever is required for the plugin.

For consistency and simplicity, `certbot-auto` can ensure a set of a documented set of environment variables (e.g., `$VENV_BIN`) or even helper functions/aliases (e.g., `pip_install`) are defined when invoking the scripts. In this way, the work required of the plugin author is proportional to how special/unique their plugin's installation process is, without imposing restrictions on that process.

This process means that plugins behave similarly/follow a similar lifecycle to certbot itself when using `certbot-auto`.

This process also means that there is no point of contention when managing plugins; installing/uninstalling plugins is as simple as adding/removing files from a directory.

cc @bmw, @ohemorange

ohemorange:
cc @erikrose

bmw:
More info in #4037.

bmw:
@zjs, thanks for taking a look at this and I'm sorry I didn't respond sooner. I think we have a renewed reason to look into this with all of the DNS plugins. Do we want `certbot-auto` to include all of the DNS plugins by default? I'm happy to create a new issue for this, but the two problems are definitely related and I think it's worth discussing this now as it's much harder to change our decision after we've shipped with certain packages included.

[...]

Issue #334, opened 2015-04-01
pyinstaller


kuba:
@jezdez suggested using http://www.pyinstaller.org/ for binary distribution

kuba:
Blocker: https://github.com/pyca/cryptography/issues/2039

m1keil:
I would actually suggest you guys to try the [Omnibus](https://github.com/chef/omnibus) path instead, to create full stack installers. If I understand correctly, your client is pluggable, which is painful when you freezing Python stuff.

Another potential issue there is that your code might need to change in order to adapt to being in frozen state.

rmoriz:
:+1: for Omnibus. Also used by other Python users e.g. https://github.com/DataDog/dd-agent-omnibus

bmw:
We decided not to use pyinstaller, largely for the reasons mentioned above of having optionally installed plugins so I'm closing this issue.

PR #4360, opened 2017-03-16
plugin: Cloudflare DNS Authenticator

zjs:
Implement an Authenticator which can fulfill a dns-01 challenge using the
Cloudflare API. Applicable only for domains using Cloudflare for DNS.

Testing Done:
 * tox -e py27
 * tox -e lint
 * Manual testing:
    * Used `certbot certonly --dns-cloudflare -d`, specifying an API key
      and email as command line arguments. Verified that a certificate
      was successfully obtained without user interaction.
    * Used `certbot certonly --dns-cloudflare -d`, without specifying an
      API key and email as command line arguments. Verified that the user
      was prompted and that a certificate was successfully obtained.
      API key and email address after selecting cloudflare interactively
      and that a certificate was successfully obtained.
 * ./tox.cover.sh certbot
    * certbot/plugins/dns_cloudflare.py: 98% (Uncovered: 161, 180)
    * certbot/plugins/dns_cloudflare_test.py: 100%

### DIFF
--- a/certbot/plugins/dns_cloudflare.py
+++ b/certbot/plugins/dns_cloudflare.py
@@ -1,0 +1,344 @@
+"""DNS Authenticator for Cloudflare."""
+import logging
+
+import zope.interface
+
+from time import sleep
+
+from acme import challenges
+
+from certbot import errors
+from certbot import interfaces
+
+from certbot.display import util as display_util
+
+from certbot.plugins import common
+
+import CloudFlare
+
+logger = logging.getLogger(__name__)
+
+
+@zope.interface.implementer(interfaces.IAuthenticator)
+@zope.interface.provider(interfaces.IPluginFactory)
+class Authenticator(common.Plugin):
+    """DNS  Authenticator for Cloudflare
+
+    This Authenticator uses the Cloudflare API to fulfill a dns-01 challenge.
+    """
+
+    description = 'Obtain certs using a DNS TXT record (if you are using Cloudflare for DNS).'
+
+    _attempt_cleanup = False
+
+    def __init__(self, *args, **kwargs):
+        super(Authenticator, self).__init__(*args, **kwargs)
+
+        self.email = None
+        self.api_key = None
+
+    @classmethod
+    def add_parser_arguments(cls, add):
+        add('propagation-seconds',
+            default=10,
+            type=int,
+            help='The number of seconds to wait for DNS to propagate before asking the ACME server '
+                 'to verify the DNS record.')
+        add('email',
+            help='Email address associated with Cloudflare account.')
+        add('api-key',
+            help='API key for Cloudflare account. ' +
+                 '(Which can be obtained from https://www.cloudflare.com/a/account/my-account)')
+
+    def prepare(self): # pylint: disable=missing-docstring
+        pass
+
+    def more_info(self): # pylint: disable=missing-docstring,no-self-use
+        return 'This plugin configures a DNS TXT record to respond to a dns-01 challenge using ' + \
+               'the Cloudflare API.'
+
+    def get_chall_pref(self, unused_domain): # pylint: disable=missing-docstring,no-self-use
+        return [challenges.DNS01]
+
+    def perform(self, achalls): # pylint: disable=missing-docstring
+        self._setup_credentials()
+
+        self._attempt_cleanup = True
+
+        responses = []
+        for achall in achalls:
+            self._perform_achall(achall)
+            responses.append(achall.response(achall.account_key))
+
+        # DNS updates take time to propagate and checking to see if the update has occurred is not
+        # reliable (the machine this code is running on might be able to see an update before
+        # the ACME server). So: we sleep for a short amount of time we believe to be long enough.
+        sleep(self.conf('propagation-seconds'))
+
+        return responses
+
+    def _perform_achall(self, achall):
+        """
+        Performs a dns-01 challenge by creating a DNS TXT record.
+
+        :param `~certbot.achallenges.AnnotatedChallenge` achall: the challenge to perform
+        :raises errors.PluginError: If the challenge cannot be performed
+        """
+
+        domain = achall.domain
+        record_name = achall.validation_domain_name(domain)
+        record_content = achall.validation(achall.account_key)
+        ttl = 120
+
+        self._get_cloudflare_client().add_txt_record(domain, record_name, record_content, ttl)
+
+    def cleanup(self, achalls):  # pylint: disable=missing-docstring
+        if self._attempt_cleanup:
+            for achall in achalls:
+                self._cleanup_achall(achall)
+
+    def _cleanup_achall(self, achall):
+        """
+        Deletes the DNS TXT record which would have been created by `_perform_achall`.
+
+        Fails gracefully if no such record exists.
+
+        :param `~certbot.achallenge s.AnnotatedChallenge` achall: the challenge to clean up after
+        """
+
+        domain = achall.domain
+        record_name = achall.validation_domain_name(domain)
+        record_content = achall.validation(achall.account_key)
+
+        self._get_cloudflare_client().del_txt_record(domain, record_name, record_content)
+
+    def _setup_credentials(self):
+        """
+        Establish credentials, prompting if necessary.
+        """
+
+        # XXX: We could make these optional and let CloudFlare.CloudFlare attempt to read them from
+        #      .cloudflare.cfg or ~/.cloudflare.cfg or ~/.cloudflare/cloudflare.cfg
+        #
+        #      Marking them as required seems to provide for a better user experience.
+
+        configured_email = self.conf('email')
+        if configured_email:
+            self.email = configured_email
+        else:
+            self.email = self._prompt_for_data('email address')
+
+        if self.email:
+            setattr(self.config, self.dest('email'), self.email)
+        else:
+            raise errors.PluginError('Cloudflare account email address required to proceed.')
+
+        configured_api_key = self.conf('api-key')
+        if configured_api_key:
+            self.api_key = configured_api_key
+        else:
+            self.api_key = self._prompt_for_data('API key')
+
+        if self.api_key:
+            setattr(self.config, self.dest('api-key'), self.api_key)
+        else:
+            raise errors.PluginError('Cloudflare account API key required to proceed.')
+
+    @staticmethod
+    def _prompt_for_data(label):
+        """
+        Prompt the user for a piece of information.
+
+        :param string label: The user-friendly label for this piece of information.
+        :returns: The user's response (guaranteed non-empty).
+        :rtype: string
+        """
+
+        display = zope.component.getUtility(interfaces.IDisplay)
+
+        while True:
+            code, response = display.input(
+                'Input Cloudflare account {0}'.format(label),
+                force_interactive=True)
+            if code == display_util.HELP:
+                # Displaying help is not currently implemented
+                return None
+            elif code == display_util.CANCEL:
+                return None
+            else:  # code == display_util.OK
+                if not response:
+                    display.notification('Please enter an {0}.'.format(label), pause=False)
+                else:
+                    return response
+
+    def _get_cloudflare_client(self):
+        """
+        Helper method to construct a `_CloudflareClient`.
+
+        Uses configured credentials.
+
+        :return: a new
+        :rtype: `_CloudflareClient`
+        """
+
+        return _CloudflareClient(self.email, self.api_key)
+
+
+class _CloudflareClient(object):
+    """
+    Encapsulates all communication with the Cloudflare API.
+    """
+
+    def __init__(self, email, api_key):
+        self.cf = CloudFlare.CloudFlare(email, api_key)
+
+    def add_txt_record(self, domain, record_name, record_content, record_ttl):
+        """
+        Add a TXT record using the supplied information.
+
+        :param string domain: The domain to use to look up the Cloudflare zone.
+        :param string record_name: The record name (typically beginning with '_acme-challenge.').
+        :param string record_content: The record content (typically the challenge validation).
+        :param int record_ttl: The record TTL (number of seconds that the record may be cached).
+        :raises: errors.PluginError if an error occurs communicating with the Cloudflare API
+        """
+
+        zone_id = self._find_zone_id(domain)
+
+        data = {'type': 'TXT',
+                'name': record_name,
+                'content': record_content,
+                'ttl': record_ttl}
+
+        try:
+            logger.debug('Attempting to add record to zone %s: %s', zone_id, data)
+            self.cf.zones.dns_records.post(zone_id, data=data)  # zones | pylint: disable=no-member
+        except CloudFlare.exceptions.CloudFlareAPIError as e:
+            logger.error('Encountered CloudFlareAPIError adding TXT record: %d %s', e, e)
+            raise errors.PluginError('Error communicating with the Cloudflare API: {0}'.format(e))
+
+        record_id = self._find_txt_record_id(zone_id, record_name, record_content)
+        logger.debug('Successfully added TXT record with record_id: %s', record_id)
+
+    def del_txt_record(self, domain, record_name, record_content):
+        """
+        Delete a TXT record using the supplied information.
+
+        Note that both the record's name and content are used to ensure that similar records
+        created concurrently (e.g., due to concurrent invocations of this plugin) are not deleted.
+
+        Failures are logged, but not raised.
+
+        :param string domain: The domain to use to look up the Cloudflare zone.
+        :param string record_name: The record name (typically beginning with '_acme-challenge.').
+        :param string record_content: The record content (typically the challenge validation).
+        """
+
+        try:
+            zone_id = self._find_zone_id(domain)
+        except errors.PluginError as e:
+            logger.debug('Encountered error finding zone_id during deletion: %s', e)
+            return
+
+        if zone_id:
+            record_id = self._find_txt_record_id(zone_id, record_name, record_content)
+            if record_id:
+                try:
+                    # zones | pylint: disable=no-member
+                    self.cf.zones.dns_records.delete(zone_id, record_id)
+                    logger.debug('Successfully deleted TXT record.')
+                except CloudFlare.exceptions.CloudFlareAPIError as e:
+                    logger.warn('Encountered CloudFlareAPIError deleting TXT record: %s', e)
+            else:
+                logger.debug('TXT record not found; no cleanup needed.')
+        else:
+            logger.debug('Zone not found; no cleanup needed.')
+
+    def _find_zone_id(self, domain):
+        """
+        Find the zone_id for a given domain.
+
+        :param string domain: The domain for which to find the zone_id.
+        :returns: The zone_id, if found.
+        :rtype: string
+        :raises: errors.PluginError if no zone_id is found.
+        """
+
+        zone_name_guesses = self._zone_name_guesses(domain)
+
+        for zone_name in zone_name_guesses:
+            params = {'name': zone_name,
+                      'per_page': 1}
+
+            try:
+                zones = self.cf.zones.get(params=params)  # zones | pylint: disable=no-member
+            except CloudFlare.exceptions.CloudFlareAPIError as e:
+                code = int(e)
+                hint = None
+
+                if code == 6003:
+                    hint = 'Did you copy your entire API key?'
+                elif code == 9103:
+                    hint = 'Did you enter the correct email address?'
+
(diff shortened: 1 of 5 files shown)


zjs:
Thinking about the future: perhaps this should be named something like `dns-cloudflare` instead of just `cloudflare`. That would allow DNS plugins to be grouped together in the short term and, potentially, grouped together as children of some new `dns` plugin in the future, without affecting command line arguments.

certbot/plugins/cloudflare.py
```diff
@@ -0,0 +1,323 @@
...
+            logger.error('Encountered CloudFlareAPIError adding TXT record: %s', e)
+            raise errors.PluginError('Error communicating with the Cloudflare API: {0}'.format(e))
+
+        record_id = self._find_txt_record_id(zone_id, record_name, record_content)
+        logger.debug('Successfully added TXT record with record_id: %s', record_id)
+
+        # DNS updates take time to propagate and checking to see if the update has occurred is not
+        # reliable (the machine this code is running on might be able to see an update before
+        # the ACME server). So: we sleep for a short amount of time we believe to be long enough.
+        sleep(self.propagation_seconds)
```
zjs: This should probably be moved up to `perform`, so that it happens only once.

zjs:
Note that PR #4281 is a better way to address the issue addressed by cfb30ee54d22a457a112abe4557e56b144c02971. Once that is merged, I'll remove cfb30ee54d22a457a112abe4557e56b144c02971 from this PR.

bmw:
I want to think about how to best package this, but @joohoi if you want to review the plugin itself, that'd be great.

zjs:
Note to reviewers: #4423 will change the way much of the code from this PR is structured.

certbot/plugins/dns_cloudflare.py
```diff
@@ -0,0 +1,323 @@
...
+        :param int record_ttl: The record TTL (number of seconds that the record may be cached).
+        :raises: errors.PluginError if an error occurs communicating with the Cloudflare API
+        """
+
+        zone_id = self._find_zone_id(domain)
+
+        data = {'type': 'TXT',
+                'name': record_name,
+                'content': record_content,
+                'ttl': record_ttl}
```
joohoi: We have a lot of request boilerplate here. It makes me think if we could just skip depending on python-cloudflare and just use raw requests instead as Cloudflare API v4 is pretty simple and straightforward. We are using a very small subset of python-cloudflare anyway, and not depending on it would immensively ease the packaging and distribution pains.

certbot/plugins/dns_cloudflare.py
```diff
@@ -0,0 +1,323 @@
...
+@zope.interface.provider(interfaces.IPluginFactory)
+class Authenticator(common.Plugin):
+    """DNS  Authenticator for Cloudflare
+
+    This Authenticator uses the Cloudflare API to fulfill a dns-01 challenge.
+    """
+
+    description = 'Obtain certs using a DNS TXT record (if you are using Cloudflare for DNS).'
+
+    propagation_seconds = 10
```
zjs: 10 seconds was consistently enough for my testing, but we can easily increase this if we see it taking longer.

certbot/plugins/dns_cloudflare.py
```diff
@@ -0,0 +1,323 @@
...
+        return 'This plugin configures a DNS TXT record to respond to a dns-01 challenge using ' + \
+               'the Cloudflare API.'
+
+    def get_chall_pref(self, unused_domain): # pylint: disable=missing-docstring,no-self-use
+        return [challenges.DNS01]
+
+    def perform(self, achalls): # pylint: disable=missing-docstring
+        self._setup_credentials()
+
+        self._attempt_cleanup = True
```
zjs: `_attempt_cleanup` will be `False` if we encounter an error during `_setup_credentials`, in which case there's no point in attempting cleanup.

certbot/plugins/dns_cloudflare.py
zjs: Some DNS providers (like Cloudflare) have simple enough APIs that we could use raw requests, but because not all do (see, e.g., ##4448) we eventually need a good story/pattern for packaging and distribution of dependencies. I'd prefer consistently applying that. 

Generally, I also think there are some advantages to using a library/SDK:

- Someone else is responsible for testing that the library actually works with the API. (For the most part, we can focus on testing that we use the library as documented.) 
- Someone else is responsible for testing that the library continues to work when API changes are made. (It's likely that services will test popular libraries/SDKs when making changes.)
- Someone else takes care of things like authentication workflows and translating API errors.

joohoi:
If user provided API key is too short, the user is presented with rather ugly stack trace:
```
Encountered exception during recovery
Encountered CloudFlareAPIError determining zone_id: Invalid request headers
  File "/vagrant/certbot/error_handler.py", line 99, in _call_registered
    self.funcs[-1]()
  File "/vagrant/certbot/auth_handler.py", line 284, in _cleanup_challenges
    self.auth.cleanup(achalls)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 94, in cleanup
    self._cleanup_achall(achall)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 109, in _cleanup_achall
    self._get_cloudflare_client().del_txt_record(domain, record_name, record_content)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 233, in del_txt_record
    zone_id = self._find_zone_id(domain)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 268, in _find_zone_id
    .format(e))
PluginError: Encountered CloudFlareAPIError determining zone_id: Invalid request headers
```
And if the credentials are just wrong, the user gets a rather uninformative stack trace as well:
```
Encountered exception during recovery
Encountered CloudFlareAPIError determining zone_id: Unknown X-Auth-Key or X-Auth-Email
  File "/vagrant/certbot/error_handler.py", line 99, in _call_registered
    self.funcs[-1]()
  File "/vagrant/certbot/auth_handler.py", line 284, in _cleanup_challenges
    self.auth.cleanup(achalls)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 94, in cleanup
    self._cleanup_achall(achall)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 109, in _cleanup_achall
    self._get_cloudflare_client().del_txt_record(domain, record_name, record_content)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 233, in del_txt_record
    zone_id = self._find_zone_id(domain)
  File "/vagrant/certbot/plugins/dns_cloudflare.py", line 268, in _find_zone_id
    .format(e))
PluginError: Encountered CloudFlareAPIError determining zone_id: Unknown X-Auth-Key or X-Auth-Email
```

certbot/plugins/dns_cloudflare.py
joohoi: You are right, cleanup gets called despite raising PluginError in setup_credentials()

certbot/plugins/dns_cloudflare.py
joohoi: Good points. It might get cumbersome for distribution packagers when these land in the base package though.

joohoi:
For some reason the plugin (python-cloudflare tbh) returns "error 1004 - DNS Validation Error" while direct requests made with exactly the same API key works. 

```
$ sudo venv/bin/certbot certonly --staging --dns-cloudflare --dns-cloudflare-email #RETRACTED# --dns-cloudflare-api-key #RETRACTED# -d testing.1o.fi
Obtaining a new certificate
dns-01 challenge for testing.1o.fi
Encountered CloudFlareAPIError adding TXT record: 1004 DNS Validation Error
Cleaning up challenges
Error communicating with the Cloudflare API: DNS Validation Error
```

While running this works:

```
#!/bin/bash

# Get your API key from https://www.cloudflare.com/a/account/my-account
CERTBOT_DOMAIN="testing.1o.fi"
CERTBOT_VALIDATION="hello from auth script"
API_KEY="RETRACTED"
EMAIL="RETRACTED"

# Strip only the top domain to get the zone id
DOMAIN=$(expr match "$CERTBOT_DOMAIN" '.*\.\(.*\..*\)')

# Get the Cloudflare zone id
ZONE_EXTRA_PARAMS="status=active&page=1&per_page=20&order=status&direction=desc&match=all"
ZONE_ID=$(curl -s -X GET "https://api.cloudflare.com/client/v4/zones?name=$DOMAIN&$ZONE_EXTRA_PARAMS" \
     -H     "X-Auth-Email: $EMAIL" \
     -H     "X-Auth-Key: $API_KEY" \
     -H     "Content-Type: application/json" | python -c "import sys,json;print(json.load(sys.stdin)['result'][0]['id'])")

# Create TXT record
CREATE_DOMAIN="_acme-challenge.$CERTBOT_DOMAIN"
RECORD_ID=$(curl -s -X POST "https://api.cloudflare.com/client/v4/zones/$ZONE_ID/dns_records" \
     -H     "X-Auth-Email: $EMAIL" \
     -H     "X-Auth-Key: $API_KEY" \
     -H     "Content-Type: application/json" \
     --data '{"type":"TXT","name":"'"$CREATE_DOMAIN"'","content":"'"$CERTBOT_VALIDATION"'","ttl":120}' \
             | python -c "import sys,json;print(json.load(sys.stdin)['result']['id'])")
# Save info for cleanup
if [ ! -d /tmp/CERTBOT_$CERTBOT_DOMAIN ];then
        mkdir -m 0700 /tmp/CERTBOT_$CERTBOT_DOMAIN
fi
echo $ZONE_ID > /tmp/CERTBOT_$CERTBOT_DOMAIN/ZONE_ID
echo $RECORD_ID > /tmp/CERTBOT_$CERTBOT_DOMAIN/RECORD_ID
```

zjs:
I've attempted to clean up the error messages. Some examples:

Incorrect email address or well-formed-but-invalid API key:
> Error determining zone_id: 9103 Unknown X-Auth-Key or X-Auth-Email. Please confirm that you have supplied valid Cloudflare credentials. (Did you enter the correct email address?)

Entering an invalid API key (e.g., by removing characters or typing a password instead of an API key):
> Error determining zone_id: 6003 Invalid request headers. Please confirm that you have supplied valid Cloudflare API credentials. (Did you copy your entire API key?)

Valid credentials, but an invalid domain name:
> Unable to determine zone_id for example.com using zone names: [u'example.com', u'com']. Please confirm that the domain name has been entered correctly and is already associated with the supplied Cloudflare account.

certbot/plugins/dns_cloudflare.py
```diff
@@ -0,0 +1,323 @@
...
+        Performs a dns-01 challenge by creating a DNS TXT record.
+
+        :param `~certbot.achallenges.AnnotatedChallenge` achall: the challenge to perform
+        :raises errors.PluginError: If the challenge cannot be performed
+        """
+
+        domain = achall.domain
+        record_name = achall.validation_domain_name(domain)
+        record_content = achall.validation(achall.account_key)
+        ttl = 120
```
zjs: Apparently, 120 is an undocumented minimum; a lower TTL is what leads to the `1004 DNS Validation Error`.

certbot/plugins/dns_cloudflare.py
joohoi: Yes, this seems to be the case currently, 10 seconds is fine.

joohoi:
LGTM, one more question though:

Do you think we should make the propagation pause time configurable by user (for all the DNS auth plugins)? If Cloudflare (or some other provider) would make a change to their system that causes the propagation take more than the current wait time, the only thing we could do to make it work again would be to make a new release.

zjs:
> Do you think we should make the propagation pause time configurable by user (for all the DNS auth plugins)? If Cloudflare (or some other provider) would make a change to their system that causes the propagation take more than the current wait time, the only thing we could do to make it work again would be to make a new release.

I think this is a good idea. I'll add an optional `--dns-<provider>-propagation-time` argument.

certbot/plugins/dns_cloudflare.py
bmw: I wanted to provide a little background here about dependencies in Certbot. First, I'll say that too many dependencies is one of the most common complaints we have. Some people have small lightweight servers and hate having to install 35+ packages just to get a cert. This can be mitigated by making Certbot more modular but this increases the complexity for people getting the pieces of Certbot they need for their use case. It's a trade off and we should try to strike a reasonable balance here.

Additionally, our current approach to keeping Certbot reasonably up to date as we and the ACME protocol continue to change rapidly is to try and convince Linux distros to update us more quickly than they usually do. This usually involves a lot of compromise and one instance of this is we told Debian we wouldn't add any non-optional dependencies to Certbot. None of these things are total blockers to us adding dependencies and there are definite benefits to doing so but I wanted to bring them up so we can make the right design decision here.

One question I had is about the viability of using the [lexicon](https://github.com/AnalogJ/lexicon) project you found for many of our DNS plugins including this one. All of the DNS plugins that have been implemented so far are supposedly supported by lexicon with the exception of Google Cloud DNS which they claim to be open to accepting a PR for. If we go this route, I'd also be curious if restructuring our DNS plugins makes sense if there's a common base and level of abstraction from lexicon we can use to easily support all/many of these DNS providers.

If for some reason this doesn't work, I think we have two options for this plugin. We could do both of these options but I think at least one is necessary.

1. We split it out into its own package.
2. We vendorize the dependencies that aren't widely packaged by OSes such as `cloudflare` and `logger`. I haven't looked into exactly what we'd need to vendorize here but I'm happy to do so if we go this route.

Not doing one of these two things means not only did we go back on our promise to Debian, but we now depend on libraries that aren't packaged. We can talk to our package maintainer if we really want to go this route, but I think we'll have a bad time.

zjs: > One question I had is about the viability of using the lexicon project you found for many of our DNS plugins including this one. All of the DNS plugins that have been implemented so far are supposedly supported by lexicon with the exception of Google Cloud DNS which they claim to be open to accepting a PR for. If we go this route, I'd also be curious if restructuring our DNS plugins makes sense if there's a common base and level of abstraction from lexicon we can use to easily support all/many of these DNS providers.

In many cases, Lexicon does not provide the functionality we need. Namely: it does not provide functionality for determining the DNS zone which should be modified.

As background: DNS APIs expose the concept of a DNS zone in various ways. Essentially, a DNS zone is a subtree of the overall DNS record tree. The DNS zone for `_acme-challenge.foo.bar.example.com` may be rooted at `_acme-challenge.foo.bar.example.com`, `foo.bar.example.com`, `bar.example.com`, or `example.com` depending on way the provider and the way the user is using the provider. For all DNS APIs I've encountered so far, DNS record creation requests are scoped to a DNS zone. As a result, it is necessary to determine the zone id (called a domain id by some providers).

Most providers have a query-like process for determining the zone id, but that is not exposed by Lexicon.

In some cases, like CloudXNS (see #4522), guess-and-check is the best we can do with or with or without Lexicon. For those, I think using Lexicon makes sense.

For providers that offer a clear query API to determine the zone id, I don't think we should attempt to determine it though guess-and-check (which means not using Lexicon for those providers).

There are also other reasons not to use Lexicon when a better alternative is available:

1. Lexicon often suppresses details when something goes wrong. With `cloudflare`, we can provide distinct error messages for the following cases: invalid credentials, malformed credentials, the domain not being associated with the Cloudflare account, and other errors. With `lexicon`, we would not be able to distinguish those cases.
2. With few exceptions, Lexicon does not use existing API libraries. Using a provider's official library makes it much easier to go to them and ask for help if something breaks. 
3. Lexicon provides a minimum subset of functionality. This is convenient for providing a consistent CLI for managing DNS records. This is less convenient for our use cases; some providers' APIs include functionality to determine whether a record update has been saved or whether a record update has finished propagating (from the provider's point of view). We don't have access to that functionality through Lexicon.

> We split it out into its own package.

If we address #1279, I think this would be worth considering regardless of the dependency discussion; it seems like few users will ever use _all_ DNS plugins. (That said, I think there are arguments to be made in favor of a simple monolithic distribution model that provides a consistent experience for all users.)

> We vendorize the dependencies that aren't widely packaged by OSes such as `cloudflare` and `logger`. I haven't looked into exactly what we'd need to vendorize here but I'm happy to do so if we go this route.

`logging` is just a standard python library, but yes: we could vendorize `cloudflare`.

I don't know much about python dependency management, but after reading a [random article](http://bitprophet.org/blog/2012/06/07/on-vendorizing/), this seems like the right call.

zjs: > This usually involves a lot of compromise and one instance of this is we told Debian we wouldn't add any non-optional dependencies to Certbot.

One option that I don't _think_ I like, but is worth considering: we could make `cloudflare` an optional dependency and hide/disable the `dns-cloudflare` plugin if it is not present.

bmw: How close do you think Lexicon is to a state where we can use it for all/more of our DNS plugins? They're not packaged in any major distro either so we'd probably have to vendorize them as well or make the functionality only available to a smaller subset of Certbot users. With this in mind, if we could easily work with upstream or fork it to get the functionality we wanted, I think it may be worth investigating.

> If we address #1279, I think this would be worth considering regardless of the dependency discussion; it seems like few users will ever use all DNS plugins.

My instinct is to put it in the base Certbot package if there are a minimal number of dependencies added/vendorized. If we can add the functionality to the core package without significantly increasing bloat, it should be done to make it easy for users to get the functionality they need, otherwise, it should go into its own package. For example, we decided for route53 we don't want to add/vendorize `boto3` and will be going another route.

> logging is just a standard python library, but yes: we could vendorize cloudflare.

`logging` is but `cloudflare` depends on `requests`, `logger`, `pyyaml`, and `future` ([source](https://github.com/cloudflare/python-cloudflare/blob/master/setup.py#L35)). We already depend on `requests` but `logger`, `pyyaml`, and `future` would be new dependencies for us. `logger` isn't packaged anywhere so we'd probably have to vendorize it in addition to `cloudflare`. `pyyaml` is pretty thoroughly packaged, but it'd still be a new (vendorized) dependency. `future` is somewhere in the middle between `logger` and `pyyaml` about how widely a distro package is available.

I reached out to our Debian package maintainer to get his thoughts, but the need to add 4 new dependencies to support Cloudflare has me a little hesitant of taking this approach or putting it in the core Certbot package.

> One option that I don't think I like, but is worth considering: we could make cloudflare an optional dependency and hide/disable the dns-cloudflare plugin if it is not present.

Yeah that's definitely an option. Previously we've taken the approach of separate packages as you can see by `certbot-apache` and `certbot-nginx`, but we could change our approach here if we wanted.

zjs: > How close do you think Lexicon is to a state where we can use it for all/more of our DNS plugins? They're not packaged in any major distro either so we'd probably have to vendorize them as well or make the functionality only available to a smaller subset of Certbot users. With this in mind, if we could easily work with upstream or fork it to get the functionality we wanted, I think it may be worth investigating.

If I were maintaining Lexicon, I'm not sure I'd want to add the functionality Certbot needs; it's a fair bit of incremental work for each provider and we may be the only users who would want it. We can certainly ask though.

I think changing the way Certbot is packaged/distributed so that dependencies aren't a huge headache may be a more valuable investment. What do you think of #4537 (which turns out to be a duplicate of #334) as a way to address dependency problems?

bmw: > I think changing the way Certbot is packaged/distributed so that dependencies aren't a huge headache may be a more valuable investment. What do you think of #4537 (which turns out to be a duplicate of #334) as a way to address dependency problems?

I'd love to do that but some of us have been pushing for this for a year+ at this point and haven't been able to get approval. There's been a lot of renewed interest and frustration recently so it's possible things will change this time, but I'm not super confident this will happen in the short term. I'm a little hesitant to block this PR on those decisions, but if you want I can bring them up in the next week or two and see what we can accomplish.

What I'd like to do is reach out the Lexicon package maintainer, explain our situation and what we'd like to do, and see how receptive they are to the changes we'd like to make. This can help inform what we do here. I think it'd be good if you talked to them since you're the most familiar with the library but I can reach out first if you prefer.

Also, while I understand the inefficiency of the guess-and-check method of determining a DNS zone to modify with Lexicon, what are the other downsides to this approach? Is the code significantly more complex or error prone? How does doing this negatively impact the user? I just want to make sure I properly understand the problems here so we can make the right decision.

I talked about this with Erica for a bit and I think we should have a design meeting about packaging these plugins this week. My inclination right now is to put this plugin as an external package if we can reasonably move it into core Certbot if we want later without causing problems for users. I think this is doable but I want to talk and think about it to make sure. Please let me know what you think!

joohoi:
LGTM! The dependency discussion must get resolved to proceed, but the PR itself looks fine.

zjs:
Since the revised version, based on Tuesday's design discussion, is so different, I've created a new PR: #4567.

PR #4361, opened 2017-03-16
Save public-ip-logging-ok when set interactively

bmw:
If someone is using manual with hooks but answered this question interactively, Certbot is going to prompt when it tries to renew!


bmw:
#4362 also solves this problem, but we may want to do a point release for this (and/or file locking) and the scope of this change is much smaller.

pde:
I just merged #4362 so we probably don't need this unless we're going back and releasing 0.12.1?

Issue #4328, opened 2017-03-11
Allow private keys to be password protected


Juma7C9:
In some scenarios it would be desirable not to save unencrypted private keys on disk, for instance when they reside on untrusted locations (like VPSes without disk encryption support), and securing communications may be a concern.

For now I'm using a script than encrypts the saved key and deletes the original one, but I think a native solution would be preferable, also to avoid saving (even temporarily) plain keys on disk.

I also understand that automation is one of the main goals of LE, but considering than many servers allow for the key's password to be entered on startup (Apache is for instance one of them), such an option may be implemented as an optional switch that have to be explicitly activated when needed.

Juma7C9:
I added a PR that implements a proof-of-concept for the functionality, I hope it would be useful.

By the way, is there a reason why the account key is generated and stored using directly the `hazmat` primitive routines instead of the ones used for everything else?
It would be nice if those keys could be stored as PEMs and encrypted as well, to augment their security too.

SwartzCr:
Whoa, really interesting @Juma7C9 
I think @bmw @pde should look at this and make a decision on how we want to support this. The PR is really helpful! Thanks for thinking about this, and if you want to chat with us about it you should join #letsencrypt-dev on freenode!

pv2b:
This would help scratch a particular itch of mine. I'm currently working on a side project to automate certificate issuance for VMware NSX, and it would seem NSX does not like passphrase-less private keys. I'll end up just generating a passphrased copy of the private key on the fly before submitting the API call in my application, but if this functionality was in certbot, that'd be one less thing for me to do.

jsha:
Hi @Juma7C9! Thanks for the suggestion. I think this is probably not a good fit for Certbot, since it adds complexity and I am dubious about its benefit. For instance, if you don't consider your VPS' disk to be secure, why would you consider its memory to be secure?

For the VMware NSX case, I think it's reasonable for NSX users to take a separate step to encrypt their keys.

SwartzCr:
I'm going to close this since it seems that we're not going to support this.

PR #4365, opened 2017-03-16
Add private key encipherment functionality

Juma7C9:
Hi,
in reference to issue #4328, I implemented a little change to the code to allow an optional private key encipherment.

This PR adds a new command line switch `--encrypt-private-key [CIPHER]`, which accepts an optional cipher arguments (all those accepted by OpenSSL, i.e. those reported by `openssl rsa --help`, defaults to `aes256` if not specified), and asks for a passphrase, which is then used to encrypt the key using the standard pyOpenSSL routines.
The passphrase is then saved to a variable and made available to be used within the same invocation, for example to sign CSRs without prompting again the user for the same password.

As I am not particularly familiar with certbot's internals, this PR should be tested for a correct integrations with the rest of the package; in particular, I had to remove the `nargs` element from the `ARGPARSE_PARAMS_TO_REMOVE` tuple (from `certbot/cli.py`), as otherwise the `argparse`'s `nargs='?'` argument would not work - and it seems that no other option uses it.
I had also to use a `if not isinstance(cipher, str):` statement instead of `if cipher is None:` as otherwise the code would not pass the `py27` test (but would correctly pass the `py36` one, which I think suggests that the problem lies elsewhere in the package).

I hope you'll find this addition useful :)

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -104,7 +104,7 @@ More detailed help:
 
 
 # These argparse parameters should be removed when detecting defaults.
-ARGPARSE_PARAMS_TO_REMOVE = ("const", "nargs", "type",)
+ARGPARSE_PARAMS_TO_REMOVE = ("const", "type",)
 
 
 # These sets are used when to help detect options set by the user.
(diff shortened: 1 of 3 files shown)


bmw:
There's been a response from the Certbot team about this feature on #4328 for a while now that hasn't seen any response. With this in mind, I'm closing this PR for now until the concerns raised in the issue have been addressed.

PR #4369, opened 2017-03-17
Add a global lock file to Certbot

### LINKED ISSUE #933: Implement locking to ensure two clients do not run at once

bmw:
Fixes #933.

This PR implements a global lock file to prevent multiple versions of Certbot from running at once. It uses functionality implemented by the OS to put an advisory lock on a file that is tied to your process. If another process tries to grab an advisory lock on the same file, it will fail. The lock file is automatically released when your process exits. I don't have a machine to test it on, but this process should work on both Windows and any UNIX system.

#### Design

Initially I planned to put a lock file in some subset of `config_dir`, `work_dir`, and `logs_dir`. Technically, this isn't enough though. These directories can all be configured to use other paths and we also modify other files like your webserver's configuration. Rather than use lock files for all directories (and hope 3rd party plugins do the same) or leave holes for potential problems, I decided to make a global lock file at it's own path.

I chose `/tmp` because it's usually writeable by all users and is available on all UNIX systems. I made the path configurable so if someone really wants they can change it. I also made the inability to access the lock file a soft fail that just warns the user of the problem. If another process holds the lock, we crash, however, if we can't even access the file, we keep going with a warning. This prevents breaking automatic renewal in systems that have unconventional permissions on `/tmp`.

#### Why fasteners?

I picked `fasteners` because it is the most popular, non-deprecated, Python 2 and 3 compatible, Windows and UNIX compatible lock file module that is packaged on all relevant OSs. It uses `fcntl()`/`lockf()` on UNIX rather than `flock()` which works on systems like [Solaris](https://wiki.enterpriselab.ch/el/public:sysadmin:letsencrypt:install-solaris-11) and works on NFS mounts. It has one dependency on `monotonic`, but both `fasteners` and `monotonic` are relatively small.

One downside of this library (and almost every other library I looked at) is it never deletes the lockfile. It turns out doing this properly is tricky. I found a Python 2 and UNIX only library that does this and we could make work for our purposes, but we'd have to vendor it and deal with the bugs. I'd rather not go down this route just for aesthetics, but if you're curious, there is a branch on this repo containing this code called `lock`.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -8,6 +8,7 @@ import sys
 import time
 import traceback
 
+import fasteners
 import zope.component
 
 from acme import jose
@@ -893,7 +894,15 @@ def main(cli_args=sys.argv[1:]):
     zope.component.provideUtility(report)
     atexit.register(report.atexit_print_messages)
 
-    return config.func(config, plugins)
+    lock = fasteners.InterProcessLock(constants.LOCK_FILE)
+    logger.debug("Attempting to acquire lock file %s", constants.LOCK_FILE)
+    if not lock.acquire(blocking=False):
+        raise errors.Error("Another instance of Certbot is already running.")
+
+    try:
+        return config.func(config, plugins)
+    finally:
+        lock.release()
 
 
 if __name__ == "__main__":
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -867,6 +867,37 @@ def _post_logging_setup(config, plugins, cli_args):
     logger.debug("Discovered plugins: %r", plugins)
 
 
+def acquire_file_lock(lock_path):
+    """Obtain a lock on the file at the specified path.
+
+    :param str lock_path: path to the file to be locked
+
+    :returns: lock file object representing the acquired lock
+    :rtype: fasteners.InterProcessLock
+
+    :raises .Error: if the lock is held by another process
+
+    """
+    lock = fasteners.InterProcessLock(lock_path)
+    logger.debug("Attempting to acquire lock file %s", lock_path)
+
+    try:
+        lock.acquire(blocking=False)
+    except IOError as err:
+        logger.debug(err)
+        logger.warning(
+            "Unable to access lock file %s. You should set --lock-file "
+            "to a writeable path to ensure multiple instances of "
+            "Certbot don't attempt modify your configuration "
+            "simultaneously.", lock_path)
+    else:
+        if not lock.acquired:
+            raise errors.Error(
+                "Another instance of Certbot is already running.")
+
+    return lock
+
+
 def _run_subcommand(config, plugins):
     """Executes the Certbot subcommand specified in the configuration.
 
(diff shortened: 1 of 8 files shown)


certbot/main.py
```diff
@@ -866,6 +867,56 @@ def _post_logging_setup(config, plugins, cli_args):
     logger.debug("Discovered plugins: %r", plugins)
 
 
+def acquire_file_lock(lock_path):
```
pde: Maybe put this in `util.py` rather than `main.py`?

certbot/main.py
```diff
@@ -866,6 +867,56 @@ def _post_logging_setup(config, plugins, cli_args):
...
+        logger.debug(err)
+        logger.warning(
+            "Unable to access lock file %s. You should set --lock-file "
+            "to a writeable path to ensure multiple instances of "
+            "Certbot don't attempt modify your configuration "
+            "simultaneously.", lock_path)
+    else:
+        if not lock.acquired:
+            raise errors.Error(
+                "Another instance of Certbot is already running.")
```
pde: This creates some risk; if Certbot gets `kill -9`'d renewal scripts will hit this error forever, right?  I wonder if we should put in some mitigation for the common case (`ps` on linux) to try to prevent that?

certbot/main.py
```diff
@@ -866,6 +867,56 @@ def _post_logging_setup(config, plugins, cli_args):
...
+
+    :param str lock_path: path to the file to be locked
+
+    :returns: lock file object representing the acquired lock
+    :rtype: fasteners.InterProcessLock
+
+    :raises .Error: if the lock is held by another process
+
+    """
+    lock = fasteners.InterProcessLock(lock_path)
```
pde: Does this actually open `lock_path` and zero / write anything to it?  Is there an attack surface where an unprivileged user can symlink `/tmp/.certbot.log` to a sensitive system file?

certbot/main.py
bmw: I can if you prefer. I put it here because it's unlikely people will ever need it but I can move it if you like. If I do, I'll want to expand on it's limitations more with the biggest being you can't use it as a lock between threads.

certbot/main.py
bmw: [No](https://github.com/harlowja/fasteners/blob/master/fasteners/process_lock.py#L124). We open it in append mode (you need an open file description to use this lock technique), but we never write anything there.

certbot/main.py
bmw: Nope! I explained this in my wall of text above which is mainly for future Certbot maintainers to go back and know why we did things this way, but the OS handles binding the lock to the process. If the process exits, the lock goes away.

We technically don't even need to release the lock in our code, but doing so feels more correct™.

bmw:
So the easy fix for test failures is to set `--lock-file` to a path in the temporary directory. I'm not sure why this is happening though and I can't reproduce locally on py26 or py27.

certbot/main.py
pde: Not a huge issue. I was just watching the fact that main.py is getting pretty large...

pde:
I think we intentionally [ask Travis to run two tests at once](https://github.com/certbot/certbot/blob/master/tox.ini#L19). Travis VMs have two cores I believe.

bmw:
Yep. You can see it running the same test in multiple process in https://travis-ci.org/certbot/certbot/builds/212318783. Going to turn that off for py26.

bmw:
I'll file a bug about this with `nose` later, but I think we should just stop running Python 2.6 tests in parallel for now.

pde:
You'd prefer to do that rather than setting `--lock-file` everywhere?

pde:
Another option: instead of crashing, wait and try again when the lock is taken, with some timeout.

pde:
I'm fine with any of those, but like protecting our test execution speed ;)

bmw:
I just don't want someone else to hit this problem in the future, but I'll use different lock files for now.

pde:
Awesome!

bmw:
Fixed merge conflicts.

bmw:
Closing and reopening to make Travis run...

pde:
Great!

PR #4370, opened 2017-03-17
"What is a Certificate?" section first draft

pconrad-fb:
Added what.rst, edited index.rst; I'm not sure if anything else is required to put what.rst into the build and TOC.


docs/what.rst
```diff
@@ -0,0 +1,23 @@
+=====================
+What is a Certificate?
+=====================
+
+An SSL *certificate* contains a public key and a private key that work together to enable
```
schoen: I'm alarmed by a couple of things in this first sentence (!). First, it uses the popular term "SSL" which is obsolete and no longer technically correct. Second, the certificate doesn't "contain" the private key. (The certificate is public, but the private key is private, and in Certbot they're even kept in separate files.)

docs/what.rst
```diff
@@ -0,0 +1,23 @@
...
+a web browser to communicate with your server securely over an encrypted SSL (secure socket layer)
+connection. The certificate also identifies your website so that the browser (and the person behind it)
+can trust your identity. SSL certificates have mainly been used to provide secure,
```
schoen: I'm not sure this description is a great fit for our DV certificates because "trust your identity" might have a lot of conceptual baggage that doesn't apply here. For example, if someone read this and then we asked "hey, if there's a certificate for greatregionalbank.com, does that mean we can trust that it's Great Regional Bank's web site?" or "hey, if there's a certificate for awesomesoftwaredownloads.net, does that mean we can trust that awesomesoftwaredownloads.net is a legitimate web site?".

I've struggled with this issue a lot because I think that certificates were *always* presented this way when they were first introduced to the web, yet in fact in our DV service, we almost don't want users to get *any* information from the certificate, we just want it to help the browser detect MITM attacks and refuse to connect. Beyond that, we add just about no security value, and may aspire to a future where the browser doesn't even show any distinctive UI for HTTPS connections, but shows a security warning for HTTP. In this case the certificate really doesn't aspire to communicate anything to the user in the course of a typical interaction, but it does still protect them by defending against surveillance and manipulation by the network intermediaries.

schoen: But, I'm kind of at a loss about how to explain this!

Also, this is more accurate in a way for EV, but Let's Encrypt never does EV.

pconrad-fb: Yes, the generic "certificate" info I got was from the web, and I knew that it might not be exactly the way you want to present them—thence the phrase "first draft" in the commit message. I figured we'd need to iterate. 

So, from what you're saying, the main points of the cert are to enable the encryption to happen (with the public/private keys) and to ensure that you're connecting to the site that actually has the cert associated with it (though there's no guarantee that the site is run by nice people).  Is that closer to the truth? I want to state the benefits accurately.

pconrad-fb: @schoen Do you have any comments on my recent reply? I'd like to refine this draft.

docs/what.rst
SwartzCr: So maybe this should be re-worded as:
"A *certificate* is made up of a public key and a private key that work together to enable a web browser to communicate with your server securely over an encrypted SSL/TLS (transport layer security)
 connection."
I think mimicing the language in the HTTPS wikipedia article is probably a safe bet: https://en.wikipedia.org/wiki/HTTPS

St-Ranger: 1.  `s^is made up^consists^`

2. I'd say that the language is a bit better in this article: https://en.wikipedia.org/wiki/Public_key_certificate (compared to the article you referenced).

3. If I may:
_A public key certificate, or a digital certificate is a combination of a public key and a private key. It is used to be called "SSL certificate". The ultimate practical purpose of a certificate is to enable secure communication between a client program (web browser, e-mail client, etc.) and a server over an encrypted SSL/TLS (transport layer security) connection. The certificate is used to encrypt such a communication and to identify the computer (device)._

And that can be continued:
_The certificate includes information about the key, information about the identity of the computer and the digital signature of the issuer. If the issuer is trusted by the software initiating the communication and the signature is valid, the key can be used to communicate securely with the server identified by the certificate._

---
Note: Since the term "SSL certificate" is still being widely used, I think it would help people making the connection, if they've heard that term.

pconrad-fb: Excellent feedback; very helpful. I'll take a look at this today or next week. What about the Lineages section?

docs/what.rst
```diff
@@ -0,0 +1,23 @@
...
+
+Certificates and Lineages
+=========================
+
+Certbot introduces the concept of a *lineage,* which is a collection of all the versions of a certificate
+plus additional information that Certbot uses to maintain the configuration of your webserver from
+renewal to renewal. Whenever you renew a certificate, Certbot keeps the same configuration unless
+you explicitly change it. If you add domains, you can either add them to an existing lineage or create
+a new one. 
+
```
St-Ranger: First, let me share the reaction of a person who was learning certbot from scratch:  "lineage" is a rather confusing term.
(I saw it being discussed in one of the recent threads here, and I agree.)  It is _sort of_ understandable, but I would have used "set", "family", or, possibly wouldn't not introduce it at all. I would rather think of different versions of the same certificate, calling it collectively a "certificate".
(For comparison: you don't call the collection of different versions of the same file in the source control a _lineage_, right?) 


My biggest confusion is what is the difference between _`lineage`_ and `Certificate Name`. 
Can someone please explain the [practical] difference? So far, from the practical point of view, _Certificate Name_ is used to label the _lineage_ (and its components). If you are intending to keep "lineage" and describe it, I would suggest incorporating that explanation in this document.

Second, I think including the "additional information that Certbot uses to maintain the configuration of your webserver from renewal to renewal" into the notion of lineage is confusing. It is like claiming that birth certificates and county records of birth, marriage and death are a part of your family. I would call those lineage (or, more exactly certificate) configuration.

Yet another suggestion: While a big focus of certbot (and LetsEncrypt) is at the certificates for websites, these certificates are also used for other servers (IMAP/POP3, MTA/MSA, ... ). I think the documentation should be talking more generally about "servers" (as opposed to webservers).

pconrad-fb: Well, here's a radical idea: Should we purge the word "lineage" from the doc and the command options? Is there ever a time when a user has to touch previous versions of a certificate, or are they just kept around to preserve history and make it easy to automatically configure certs at renewal time?

SwartzCr: I'd be really curious to hear what @schoen has to say about this

docs/what.rst
```diff
@@ -0,0 +1,23 @@
+=====================
+What is a Certificate?
+=====================
+
+An SSL *certificate* contains a public key and a private key that work together to enable
+a web browser to communicate with your server securely over an encrypted SSL (secure socket layer)
+connection. The certificate also identifies your website so that the browser (and the person behind it)
+can trust your identity. SSL certificates have mainly been used to provide secure,
+encrypted communication for sensitive information like financial and health data, but
+we now recommend all websites run on SSL to prevent eavesdropping and content hijacking.
```
St-Ranger: In relation to this suggestion and a few other suggestions above that you made last night, - I am quoting from my comment above:

> Yet another suggestion: While a big focus of certbot (and LetsEncrypt) is certificates for websites, these certificates are also used for other servers (IMAP/POP3, MTA/MSA, ... ). I think the documentation should be talking more generally about "servers" (as opposed to webservers).

docs/what.rst
```diff
@@ -0,0 +1,29 @@
...
+content.
+
+You can use Certbot to easily obtain and configure a free certificate from Let's Encrypt, a
+joint project of EFF, Mozilla, and many other sponsors.
+
+Certificates and Lineages
+=========================
+
+Certbot introduces the concept of a *lineage,* which is a collection of all the versions of a certificate
+plus additional information that Certbot uses to maintain the configuration of your webserver from
```
St-Ranger: Should it be just "server" instead of "webserver"?
In principle, one can have certbot running on a server without any webserver, right?

Thinking more about that, I started questioning: is it really the configuration of the server that is saved, or rather the configuration of _certbot_? Yes, it may _reflect_ the server configuration, but only in the part (amount) and the way that is used for configuring _certbot_. 
Moreover, it is actually a configuration pertaining only to the particular certificate. (And in the next sentence, you are actually referring to this _Certbot ... configuration_
So, I am proposing "to maintain the configuration of certbot for that certificate".

Furthermore, we can simplify that to:
```
-plus additional information that Certbot uses to maintain the configuration of your webserver from
+plus configuration of certbot for that certificate maintained from
```

pconrad-fb: Made an edit similar to your suggestion in what.rst.

SwartzCr:
I'd like @schoen to have the final word on approving this

docs/what.rst
```diff
@@ -0,0 +1,29 @@
...
+=========================
+
+Certbot introduces the concept of a *lineage,* which is a collection of all the versions of a certificate
+plus Certbot configuration information maintained for that certificate from
+renewal to renewal. Whenever you renew a certificate, Certbot keeps the same configuration unless
+you explicitly change it. If you add domains, you can either add them to an existing lineage or create
+a new one. 
+
+See also:
+:ref:`updating_certs`
```
SwartzCr: @schoen suggests;

What is a Certificate?

A digital certificate (historically called an SSL certificate) enables secure communication between a client program (web browser, email client, etc.) and a server over an encrypted TLS (Transport Layer Security) connection, such as HTTPS for web connections. (TLS was previously known as SSL.)

Browsers and other software commonly require a certificate when making a secure connection to an Internet service. You can use Certbot to easily obtain and configure a free certificate from Let's Encrypt, a joint project of EFF, Mozilla, and many other sponsors. These certificates are accepted by mainstream browsers and other software.

The certificate contains a public key. The certificate's recipient—the operator of the site or service described by the certificate—also has the corresponding private key, which needs to be kept secret. The certificate is used to identify the server when establishing a secure connection. The certificate includes information about the public key, information about the server identity, and the digital signature of the certificate authority that issued the certificate. If the issuer is trusted by the software that initiates the communication, and the signature is valid, then the certificate confirms that the public key can be used to communicate securely with the server identified by the certificate. The main security benefit of using a certificate is the prevention of "man-in-the-middle" attacks, in which someone in between you and the server you think you are talking to is able to spy on you or insert their own (harmful) content.

Once a certificate has been obtained, it also needs to be installed in whatever server software is going to use it. This is generally done by editing that software's configuration files. Certbot can take care of this step for Apache and Nginx web servers.

Certificates and Lineages

Certificates eventually expire (in Let's Encrypt's case, every ninety days). "Renewing" a certificate really means getting a fresh certificate for the same domain names, which can be installed in place of the old one. In most configurations, Certbot can do this automatically using `certbot renew`.

Certbot introduces the concept of a lineage, which is a collection of all the versions of a certificate and its associated files from renewal to renewal, plus Certbot configuration information for that certificate. Whenever you renew a certificate, Certbot keeps the same configuration unless you explicitly change it. If you add domains, you can either add them to an existing lineage or create a new one.

You can view a list of lineages, and their names, by running `certbot certificates`; for most operations with Certbot, you can specify which lineage you want to act on using `--cert-name`.

See also: :ref:`updating_certs`

St-Ranger: That's a nicely detailed and well formulated description. May I add a few comments and suggestions? (I am using **bold** font to indicate added and edited portions of the text.

**Web** browsers and other software commonly require a certificate when making a secure connection to an Internet service. ...

[Comment: There are also image browsers, file browsers, etc. that are working locally.]

The certificate contains a public key. The certificate's recipient—the operator of the site or service described by the certificate—also has the corresponding private key, which needs to be kept secret. The certificate is used to identify the server when establishing a secure connection **and to encrypt the communication for this connection**. 

> The main security benefit of using a certificate is the prevention of "man-in-the-middle" attacks, in which someone in between you and the server you think you are talking to is able to spy on you or insert their own (harmful) content.

That's not quite right. That's only one security benefit. (And I wouldn't call it a "benefit", rather it is the purpose.)
But since the encryption of the secure channel relies on the key (certificate), having a certificate/key is a necessity.  Hence, I suggest something like this:

The certificate contains a public key. The certificate's recipient—the operator of the site or service described by the certificate—also has the corresponding private key, which needs to be kept secret. **There are two roles that a certificate plays in a secure connection:  to identify the server when establishing the connection and to encrypt the communication for the connection**. The certificate includes information about the public key, information about the server identity, and the digital signature of the certificate authority that issued the certificate. If the issuer is trusted by the software that initiates the communication, and the signature is valid, then the certificate confirms that the public key can be used to communicate securely with the server identified by the certificate. **One of the purposes of identifying the server is to prevent** "man-in-the-middle" attacks, in which someone **in-between** you and the server you think you are talking to is able to spy on you or insert their own (harmful) content.

[Note the dash in *in-between*, between the words ;-)  Ref: https://www.merriam-webster.com/dictionary/in-between ]

Would it make sense listing all components of the certificate explicitly and together -- somewhere around here?


> Certbot introduces the concept of a lineage, which is a collection of all the versions of a certificate and its associated files from renewal to renewal, plus Certbot configuration information for that certificate. 

What are the "associated files" that are neither parts of the certficate, not Certbot configuration information? Do you mean CSRs? Whatever you mean by them, why not list all components of a lineage explicitly?


> You can view a _list of lineages_, and their names, by running certbot _certificates_; for most operations with Certbot, you can _specify which lineage_ you want to act on using _--cert-name_.

I understand that you correctly described how Certbot works, but doesn't this sentence expose some awkwardness of the terminology used by Certbot? A naive question: if I want to view a list of *lineages* why don't I use a command `certbot lineages`? And if I want to specify which *lineage*, why don't I use --lineage-name? ;-)
I am not suggesting to introduce these commands; the existing commands are more intuitive. Rather, I am demonstrating why the notion of "lineage" is confusing. 

Actually, I'd suggest rephrasing this sentence to avoid this awkwardness as follows:

You can view a list of **certificate** lineages, and their names, by running `certbot certificates`. For most operations with Certbot, you can specify which **certificate** ~~lineage~~ you want to act on using `--cert-name`.

I think certbot is primarily acting on a certificate, not on a lineage. This is consistent with your description of the lineage and with certbot's commands:   you request, obtain, renew, modify, remove  a **certificate**, not a _lineage_. (You modify the lineage by modifying [replacing] the current certificate) You delete the _lineage_. (I've never done that yet, but I assume the entire lineage is deleted, right?)     And `certbot --help` output talks about acting on **certificates**, not on lineages.

SwartzCr:
@pconrad-fb are those suggestions useful? Do you have enough to respond to, or would you like feedback in a different format?

docs/what.rst
pconrad-fb: @SwartzCr This is great feedback. I have been totally slammed at work for a few days; sorry for the radio silence. I'll try to collate all the comments from this thread and any related threads into my PR and try to get this into pullable shape. Uh, hopefully over the next few days. Let me know if that works.

pconrad-fb: Also kudos to major feedbackers @St-Ranger and @schoen.

docs/using.rst
```diff
@@ -297,7 +297,9 @@ need to issue this command in normal circumstances.
 
 ``--expand`` tells Certbot to update an existing certificate with a new
 certificate that contains all of the old domains and one or more additional
-new domains.
+new domains. With the ``--expand`` option, use the ``-d`` option to specify
+the new domain, and also use ``-d`` and re-type each existing domain in the certificate.
+For example:
```
SwartzCr: is there supposed to be an example here? I don't see it

pconrad-fb: Yes, eventually. Still waiting on answer to syntax question here: https://github.com/certbot/certbot/issues/3664

I pushed this commit just to keep things moving along, but my real focus yesterday was the Challenges section. The --expand stuff is not done yet, but I can finish it once I understand the syntax sufficiently (and then I'll add an example).

docs/what.rst
```diff
@@ -0,0 +1,29 @@
+=====================
+What is a Certificate?
+=====================
+
+A public key or digital *certificate* (formerly called an SSL certificate) uses a public key 
+and a private key to enable secure communication between a client program (web browser, email client, 
+etc.) and a server over an encrypted SSL (secure socket layer) or TLS (transport layer security) connection.
+The certificate is used both to encrypt the communication and to identify the server. The certificate
```
St-Ranger: -The certificate is used both to encrypt the communication and to identify the server. The certificate
+The certificate is used both to encrypt _the initial stage of_ the communication _(secure key exchange)_ and to identify the server. The certificate


(I believe this is more accurate, isn't it? And I know that I proposed the previous version myself... But I just realized that it was somewhat inaccurate.)

docs/using.rst
```diff
@@ -296,7 +297,21 @@ need to issue this command in normal circumstances.
...
+
+Example:
+
+.. code-block::
+
+  certbot --expand --cert-name existing.com -d example.com,newdomain.com
+
+If you prefer, you can specify the domains individually like this:
+
+.. code-block::
```
bmw: When building our docs, these `code-block`s are the source of error messages and are completely absent from the built documentation. The syntax of these directives in general is described [here](http://www.sphinx-doc.org/en/stable/markup/code.html#directive-code-block), but you should make this `code-block` line and the one above:
```
.. code-block:: none
```

docs/what.rst
```diff
@@ -0,0 +1,29 @@
+=====================
+What is a Certificate?
+=====================
```
bmw: This is a minor point, but an open source contributor recently went through our docs and cleaned up all the error messages you get when you build our docs (see #4706). These lines introduce new ones though because the title overline and underlines are too short. You should add an extra `=` character to each line.

pconrad-fb:
I was hoping to get to this today, but things get crazy. Also there's a lot here to look at. It's going to end up happening next Wednesday or Thursday.

SwartzCr:
@pconrad-fb any progress?

pconrad-fb:
Not yet. I got as far as I could last week and as commented above I'm hoping to get some time to devote to this today or tomorrow... probably tomorrow, but I might get a chance to dig in after 3:30 today.

SwartzCr:
Okay great! Looking forward to it

docs/challenges.rst
```diff
@@ -1,159 +1,86 @@
...
+* Posting a specified file in a specified location on a web site (the HTTP-01 challenge)
+* Offering a specified temporary certificate on a web site (the TLS-SNI-01 challenge) 
+* Posting a specified DNS record in the domain name system (the DNS-01 challenge)
+
+It’s possible to complete each type of challenge *automatically* (Certbot directly makes the necessary 
+changes itself, or runs another program that does so), or *manually* (Certbot tells you to make a 
+certain change, and you edit a configuration file of some kind in order to accomplish it). Certbot's 
+design favors performing challenges automatically, and this is the normal case for most users of Certbot.
+
+Some plugins offer an *authenticator*, meaning that they can satisfy challenges:
```
pconrad-fb: I'm going to punt on this one temporarily in order to hit as much of the other LHF as possible and get a functioning version of challenges.rst into your repo. 

However, I agree that there's some overlap and these sections should be restructured. The genesis of the redundancy is that challenges.rst was developed separately. Once we get the content accurate, we can look at better integration.

docs/challenges.rst
```diff
@@ -1,159 +1,86 @@
...
+
+TLS-SNI-01 Challenge
+~~~~~~~~~~~~~~~~~~~~
+
+* The TLS-SNI-01 challenge doesn’t work with content delivery networks (CDNs) 
+  like CloudFlare and Akamai because the domain name is pointed at the CDN, not directly at your server.
+* Make sure port 443 is open, publicly reachable from the Internet, and not blocked by a router or firewall.
+* When using the Apache plugin, make sure you are running Apache and no other web server on port 443.
+* When using the NGINX plugin, make sure you are running NGINX and no other web server on port 443.
+* With either the Apache plugin or the NGINX plugin, check whether certbot is modifying your web
```
pconrad-fb: So, should I change it to "file a bug and use ``certonly`` for now"?

pconrad-fb:
Okay, I pushed changes that I think reflect all the feedback except for the fact that challenges.rst has some overlap with the Choosing a Plugin section. That's next on the docket, once we resolve and merge all the current stuff.

docs/challenges.rst
bmw: I'd personally opt to just remove this bullet point entirely.

cc @schoen @SwartzCr

docs/what.rst
bmw: @pconrad-fb, looks like this comment was missed.

docs/challenges.rst
SwartzCr: I'd re-word it in such a way as to not suggest users look at their configurations if certbot doesn't show an error. Instead I'd say something like:
"With either the Apache or NGINX plugin certbot will also automatically modify your web server configuration to serve the certificate. If you'd prefer this not to happen you can use the ``certonly`` command to just obtain the certificate, then install it manually. If you receive an InstallationError that means you have received a certificate but the plugin was unable to modify your web server configuration, meaning that you'll have to install it manually."
But maybe that's too wordy?
Essentially you should let people know that if it runs the challenge but gives an error on install then they still have a valid cert, but prompting people to read certbot's changes to their apache/nginx configs when things are fine seems like a bad idea.

docs/using.rst
```diff
@@ -296,7 +297,21 @@ need to issue this command in normal circumstances.
...
+.. code-block:: none
+
+  certbot --expand --cert-name existing.com -d example.com,newdomain.com
+
+If you prefer, you can specify the domains individually like this:
+
+.. code-block:: none
+
+  certbot --expand --cert-name existing.com -d example.com -d newdomain.com
+
```
SwartzCr: This doesn't seem to be an accurate representation of what @bmw said in #4660 
I think this section should be elaborated to either specify that expand and cert-name are different commands, or that cert-name should always be used instead of expand.
I think I prefer the latter, since --expand can lead to an unpredicted cert getting modified, where as cert-name leads to more control.

docs/what.rst
pconrad-fb: I'll take a look at this (and other comments) today or tomorrow.

docs/challenges.rst
bmw: I'm sorry I first read this after this line was updated again, but there's isn't an `InstallationError` in Certbot. Perhaps @SwartzCr knows what he was referring to here.

docs/using.rst
```diff
@@ -296,7 +297,24 @@ need to issue this command in normal circumstances.
...
 certificate that contains all of the old domains and one or more additional
-new domains.
+new domains. With the ``--expand`` option, use the ``-d`` option to specify
+all existing domains and one or more new domains.
+
+Example:
+
+.. code-block:: none
+
+  certbot --expand -d existing.com -d example.com,newdomain.com
```
bmw: Question: I think the original purpose of these two `--expand` examples was to show that you can use `-d <domain 1>,<domain 2>,<domain 3>` or `-d <domain 1> -d <domain 2> -d <domain 3>`. Now the first example uses both versions (which is also valid). Is this what you wanted?

pconrad-fb: Oh, my mistake. I will fix it.

SwartzCr:
Just to speed things up, I made commits on top of this and made a new PR #4948
I'm going to close this PR so that we can get all of @pconrad-fb's work in sooner rather than later

pconrad-fb:
Cool. Next steps: consolidating the two sections that talk about plug-ins? Or is there something more urgent to do first? @SwartzCr

SwartzCr:
@pconrad-fb I think that's  fine thing to work on, I'm going to take a crack at the configuration file stuff, and then maybe you can edit it once I have?

pconrad-fb:
Sounds like a plan. I'll start by doing a pull to make sure I'm synced with your head revisions, been a while since I did that.

PR #4372, opened 2017-03-17
Enhance display.util to support input validation

zjs:
Currently, each caller of `display.input` and `display.directory_select` needs to implement their own validation handling, if they want any.

The `dns-cloudflare` plugin introduces validation handling, and it seems as though each other DNS plugin will also need some.

To avoid (growing) duplication, this change allows callers to supply a `validator` to calls to `display.input` and `display.directory_select`.

`NoninteractiveDisplay` is intentionally omitted from these changes, as there is no way to re-prompt in a non-interactive workflow. This should also help ensure that new validation rules are not inadvertently applied to stored values now or in the future.

### DIFF
--- a/certbot/display/ops.py
+++ b/certbot/display/ops.py
@@ -289,3 +289,60 @@ def _gen_https_names(domains):
             domains[-1])
 
     return ""
+
+
+def _get_validated(method, validator, message, default=None, **kwargs):
+    if default is not None:
+        try:
+            validator(default)
+        except errors.Error as error:
+            logger.debug('Encountered invalid default value "%s" when prompting for "%s"',
+                         default,
+                         message,
+                         exc_info=True)
+            raise AssertionError('Invalid default "{0}"'.format(default))
+
+    while True:
+        code, raw = method(message, default=default, **kwargs)
+        if code == display_util.OK:
+            try:
+                validator(raw)
+                return code, raw
+            except errors.Error as error:
+                logger.debug('Validator rejected "%s" when prompting for "%s"',
+                             raw,
+                             message,
+                             exc_info=True)
+                zope.component.getUtility(interfaces.IDisplay).notification(str(error), pause=False)
+        else:
+            return code, raw
+
+
+def validated_input(validator, *args, **kwargs):
+    """Like `~certbot.interfaces.IDisplay.input`, but with validation.
+
+    :param callable validator: A method which will be called on the
+        supplied input. If the method raises a `errors.Error`, its
+        text will be displayed and the user will be re-prompted.
+    :param list *args: Arguments to be passed to `~certbot.interfaces.IDisplay.input`
+    :param dict **kwargs: Arguments to be passed to `~certbot.interfaces.IDisplay.input`
+    :return: as `~certbot.interfaces.IDisplay.input`
+    :rtype: tuple
+    """
+    return _get_validated(zope.component.getUtility(interfaces.IDisplay).input,
+                          validator, *args, **kwargs)
+
+
+def validated_directory(validator, *args, **kwargs):
+    """Like `~certbot.interfaces.IDisplay.directory_select`, but with validation.
+
+    :param callable validator: A method which will be called on the
+        supplied input. If the method raises a `errors.Error`, its
+        text will be displayed and the user will be re-prompted.
+    :param list *args: Arguments to be passed to `~certbot.interfaces.IDisplay.directory_select`
+    :param dict **kwargs: Arguments to be passed to `~certbot.interfaces.IDisplay.directory_select`
+    :return: as `~certbot.interfaces.IDisplay.directory_select`
+    :rtype: tuple
+    """
+    return _get_validated(zope.component.getUtility(interfaces.IDisplay).directory_select,
+                          validator, *args, **kwargs)
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -122,8 +122,8 @@ class FileDisplay(object):
         return code, selection - 1
 
     def input(self, message, default=None,
-              cli_flag=None, force_interactive=False, **unused_kwargs):
-        # pylint: disable=no-self-use
+              cli_flag=None, force_interactive=False, validator=None,
+              **unused_kwargs):
         """Accept input from the user.
 
         :param str message: message to display to the user
@@ -131,6 +131,9 @@ class FileDisplay(object):
         :param str cli_flag: option used to set this value with the CLI
         :param bool force_interactive: True if it's safe to prompt the user
             because it won't cause any workflow regressions
+        :param function validator: A method which will be called on the
+            supplied input. If the method raises a `errors.PluginError`, its
+            text will be displayed and the user will be re-prompted.
 
         :returns: tuple of (`code`, `input`) where
             `code` - str display exit code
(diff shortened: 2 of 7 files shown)


zjs:
Because the logic using the validator is `IDisplay`-implementation-dependent, I think this belongs in `IDisplay` itself.

For example, we don't want `NoninteractiveDisplay` to attempt to re-prompt since we can't interact with the user and the static value we're reading isn't going to change just because we try to read it again.

It seems like writing a utility method for this would require that method to base its behavior on which concrete implementation of `IDisplay` is being used, which seems... bad.

bmw:
Oh good point. I honestly didn't think about that.

We do have similar behavior in `FileDisplay` though where we will sometimes return the default without prompting. The way this is currently worked around in this PR is we simply don't validate the input if we don't prompt. Is this the correct behavior? While it should only stop a programmer from shooting himself in the foot, I think returning unvalidated input when a validator was given is a bit odd.

What do you think about an external function that does something like the following pseudocody Python:
```
if default is not None:
    assert validator(default) doesn't raise
while True:
    ans = IDisplay.input_method(*args, **kwargs)
    if ans is valid:
        return ans
    IDisplay.notification("whoops")
```
On the one hand this may slightly violate the boundary of `IDisplay` by assuming things about the behavior of
its implementations but on the other this behavior is clearly documented in `IDisplay`, it simplifies the interface, ensures input is validated as requested, and prevents programmers from shooting themselves in the foot.

zjs:
> The way this is currently worked around in this PR is we simply don't validate the input if we don't prompt. Is this the correct behavior? While it should only stop a programmer from shooting himself in the foot, I think returning unvalidated input when a validator was given is a bit odd.

I agree it's odd, but I think validating non-interactive input leads to tricky backwards compatibility questions.

> We do have similar behavior in `FileDisplay` though where we will sometimes return the default without prompting.

The behavior in `FileDisplay` to determine whether to return the default without prompting is substantially more complex than `if default is not None` (see `_can_interact`). I don't think it makes sense to try to duplicate/emulate/relocate that logic to avoid adding the notion of validation to `IDisplay`.

bmw:
> I agree it's odd, but I think validating non-interactive input leads to tricky backwards compatibility questions.

How so? Validation isn't done if one isn't provided.

> The behavior in FileDisplay to determine whether to return the default without prompting is substantially more complex than if default is not None (see _can_interact). I don't think it makes sense to try to duplicate/emulate/relocate that logic to avoid adding the notion of validation to IDisplay.

The idea isn't to move that logic, it's to ensure the provided default is valid whether it's going to be used or not. It's a basic sanity check for the programmer and allows us to workaround repeatedly trying to validate the same static string by making the (documented) assumption about default being returned when Certbot can't/won't interact with the user.

zjs:
> How so? Validation isn't done if a validator isn't provided.

I was worried about what would happen for non-interactive input as a result of the change I make to [webroot](https://github.com/certbot/certbot/pull/4372/files#diff-7393c62a285ef57dcedaeac0f94364d3) to supply a validator. I see now that `force_interactive=True` is used, so it seems like there's nothing to worry about.

> The idea isn't to move that logic, it's to ensure the provided default is valid whether it's going to be used or not. It's a basic sanity check for the programmer and allows us to workaround repeatedly trying to validate the same static string by making the (documented) assumption about default being returned when Certbot can't/won't interact with the user.

Ah, I see! The psuedocode makes much more sense now.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
 
     return ""
+
+
+def _get_validated(method, validator, message, default=None, **kwargs):
+    if default is not None:
+        try:
+            validator(default)
+        except errors.PluginError as error:
+            raise AssertionError("Invalid default {0}".format(default))
```
bmw: nits:

1. We should log the raised exception out of `validator`. It probably isn't useful, but it might be so lets throw it in a log. Easiest way to do this is `logger.debug("some msg", exc_info=True)`.
2. This probably will never matter, but you can turn off assertions in Python by passing `-O` to the interpreter, but this doesn't work if you explicitly raise an `AssertionError`. Instead, we could do `assert False, "Invalid default {}".format(default)` here.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
+
+    while True:
+        code, raw = method(message, default=default, **kwargs)
+        if code == display_util.OK:
+            try:
+                validator(raw)
+                return code, raw
+            except errors.PluginError as error:
```
bmw: Same comments about using `PluginError` and not logging the raised exception.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
+                z_util(interfaces.IDisplay).notification(str(error), pause=False)
+        else:
+            return code, raw
+
+
+def validated_input(validator, *args, **kwargs):
+    """Like ~certbot.interfaces.IDisplay.input, but with validation.
```
bmw: nit: If you surround `~certbot.interfaces.IDisplay.input` with backticks, it gets generated as a link in our documentation which is kind of nice. If you haven't seen it before, here's a [link](https://certbot.eff.org/docs/contributing.html#updating-the-documentation) to how to build our documentation. There's a number of places in this and `validated_directory`'s docstring where this can be done.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
+    
+    :param validator:
```
bmw: This should probably be updated to match the `validator` description below or at the very least the docstring of this function should refer to the other function.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
+    :return: as ~certbot.interfaces.IDisplay.input
+    :rtype: tuple
+    """
+    return _get_validated(z_util(interfaces.IDisplay).input, validator, *args, **kwargs)
+
+
+def validated_directory(validator, *args, **kwargs):
+    """Like ~certbot.interfaces.IDisplay.directory_select, but with validation.
+
+    :param function validator: A method which will be called on the
```
bmw: Using "callable" for the type is arguably more accurate and also causes a link to be generated to the Python documentation where "function" does not.

Also, just a reminder to update this docstring if you change how `validator` reports failure.

certbot/display/ops.py
```diff
@@ -289,3 +289,49 @@ def _gen_https_names(domains):
...
+    while True:
+        code, raw = method(message, default=default, **kwargs)
+        if code == display_util.OK:
+            try:
+                validator(raw)
+                return code, raw
+            except errors.PluginError as error:
+                z_util(interfaces.IDisplay).notification(str(error), pause=False)
+        else:
+            return code, raw
```
bmw: nit: This line has no test coverage.

certbot/display/ops.py
zjs: > This probably will never matter, but you can turn off assertions in Python by passing `-O` to the interpreter, but this doesn't work if you explicitly raise an `AssertionError`. Instead, we could do `assert False, "Invalid default {}".format(default)` here.

I'm not sure being able to turn this off would be good; it seems like that could lead to an infinite loop below.

bmw: That's fair. If you wanted we could also change the loop from `while True:` to `for _ in range(n)` where `n>=100` and error out after the loop.

zjs: It seems like `while True` is an established pattern in the codebase, whereas I can't find uses of `for _ in range`. I can see the advantages of the `range` pattern, but I think it'd be better to make a change like that in a cross-cutting way if it's what we want.

bmw: That's fair. It was just a thought to add some extra protections around the non-interactive case. We could set `n` high enough where we don't expect a user to ever hit it, but I don't expect this to be a problem.

PR #4373, opened 2017-03-17
Add Gentoo to list of official packages

osirisinferi:
Additidion of the `certbot` and related packages from the Gentoo repository to the Packaging documentation.

Also some other minor changes so every list of packages among the distributions has the same style.


SwartzCr:
lgtm - thanks for the update!

PR #4374, opened 2017-03-18
Py3 enable boulder

yan12125:
This pull request enables boulder integration tests for all Python 3 versions on Travis CI as well as fix bugs identified by integration tests.

Ref: #3179 
Ref: #4264. Commit 7a71779 ("certbot: Use bytes objects for all server-returned data") fixes an issue with a very similar backtrace as #4264. I believe they're the same bug.

PR #4376, opened 2017-03-18
Fix Docker IP detection with different ifconfig output formats

yan12125:
When trying out integration tests (#4374), I find that on Arch Linux IP detection doesn't work, leading to failures of integration tests. The cause is that new ```ifconfig``` uses a different layout. For example:
```
$ ifconfig docker0 
docker0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.17.0.1  netmask 255.255.0.0  broadcast 0.0.0.0
        inet6 fe80::42:80ff:fe80:f92  prefixlen 64  scopeid 0x20<link>
        ether 02:42:80:80:0f:92  txqueuelen 0  (Ethernet)
        RX packets 351742  bytes 19691534 (18.7 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 354247  bytes 63201066 (60.2 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
The upstream change can be found in https://sourceforge.net/p/net-tools/code/ci/eb04ef31571f6c707eacaba6846feeebfab518e6/

In this pull request I fix it up as well as report errors earlier so that debugging can be easier for others.

### DIFF
--- a/tests/boulder-fetch.sh
+++ b/tests/boulder-fetch.sh
@@ -12,5 +12,7 @@ fi
 
 cd ${BOULDERPATH}
 FAKE_DNS=$(ifconfig docker0 | grep "inet addr:" | cut -d: -f2 | awk '{ print $1}')
+[ -z "$FAKE_DNS" ] && FAKE_DNS=$(ifconfig docker0 | grep "inet " | xargs | cut -d ' ' -f 2)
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks!

Issue #2208, opened 2016-01-16
Install Python3 via EPEL on CentOS6 with letsencrypt-auto?


alex:
CentOS 6 ships with Python 2.6 by default, which isn't 100% stably supported by letsencrypt, and which the Python community at large is trying to drop support for. But there's good news! CentOS has support for Software Collections which make it possible to install Python 2.7 in parallel: https://www.softwarecollections.org/en/scls/rhscl/python27/

Would it make sense for `letsencrypt-auto` to just install a Python 2.7, in addition to all the other packages it provides, in order to a) ensure users always get a fully functional environment, b) help encourage folks to move to Python 2.7

pde:
@hogarthj what are your thoughts on this?

claque2000:
I got this need too (using the already installed scl repository and python27) and plan to made the necessaries changes. 
If you wanna to install scl and python during bootstrap, the determine python version function used in letsencrypt-auto will need another check to cover the specific python path

noizo:
I have similar issue.
But i have installed Python 3.5 from IUS repo, and it resides just next to standalone Python 2.6.

I think, it would be nice, if letsencrypt-auto would check, if there is another Python version installed, and if it is newer then deprecated, it would use it, instead of default one.

bciscato:
+1

bmw:
Doing this allows us to remove Python 2.6 support which fixes or helps make progress on a lot of issues like #3726, #4244, #4377, and #4379. @SwartzCr put me in touch with some people from Red Hat who told us how to reliably and safely do this programmaticly. Here's what I learned:

OK, so the first thing to do would be to get a copy of Python 2.7 onto the system. RHEL and CentOS 6 both provide a technology called "Software Collections" that allow users to install a version of Python 2.7 (along with many other useful tools) outside of the usual system paths.

On CentOS, this is accomplished by installing the `centos-release-scl` package (`yum install centos-release-scl`) followed by `yum install python27`.

On RHEL, this may depend on which subscriptions are available; it may be necessary to add the "Optional" set of repositories to the RHEL subscription in order to install the python27 package.
```
subscription-manager repos --enable rhel-6-server-optional-rpms
subscription-manager repos --enable rhel-server-rhscl-6-rpms
```
On some systems, `rhn-channel` must be used instead of subscription manager. You can determine which program must be used by running:
```
yum -C repolist |grep 'receiving updates from RHN Classic'
```

or
```
[...]

PR #4379, opened 2017-03-18
Don't install argparse on python >=3.2

cmskog:
On Python 3.5, I got an exception about argparse.ArgumentParser
not supporting allow_abbrev in __init__.
Turns out argparse is in stdlib on Python >=3.2, so no need to install
argparse there.
Worse, installing argparse from PyPi pulls in a stale version that does
not support the argument allow_abbrev.
ConfigArgParse.ArgumentParser, which uses as argparse.ArgumentParser
as base class, assumes 'allow_abbrev' is supported since we are on
Python 3.5(where this actual argument got supported).
And with the old class installed we get the exception.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -8,7 +8,7 @@ version = '0.13.0.dev0'
 
 # Please update tox.ini when modifying dependency version requirements
 install_requires = [
-    'argparse',
+    'argparse;python_version<"3.2"',
     # load_pem_private/public_key (>=0.6)
     # rsa_recover_prime_factors (>=0.8)
     'cryptography>=0.8',
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ version = meta['version']
 # https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
     'acme=={0}'.format(version),
-    'argparse',
+    'argparse;python_version<"3.2"',
     # We technically need ConfigArgParse 0.10.0 for Python 2.6 support, but
     # saying so here causes a runtime error against our temporary fork of 0.9.3
     # in which we added 2.6 support (see #2243), so we relax the requirement.


bmw:
We recently changed this to properly package our wheels in a way that works with both new and old versions of `pip` (without building multiple wheels). Thanks for bringing this up before we did a release.

@erikrose, seems like we'll have to rethink our approach here.

erikrose:
Some ideas:
1. Port to click, which we've been dallying with anyway and which is based on optparse.
2. Drop Python 3 support, which would quickly fix the problem.
3. Drop Python 2 support, which many major projects (like Django) are now doing.
4. Keep a conditional argparse dependency, but build separate py2 and py3 wheels.

Click porting is time-consuming and full of packaging unknowns, so it's a long-term plan.

Dropping Python 2 is prettiest but will probably leave users of ancient distros behind. Do they deserve it? Perhaps. Will it make a more secure internet? Well…maybe, if it encourages them to upgrade! ;-)

Dropping Python 3 is meh, because we'll be forced to deal with it eventually.

My favorite short-term fix is (4). Because we're dividing on major versions, we shouldn't have the wheel-caching problems of #2343. Also because we're dividing on major versions, we don't need to depend on environment markers, which ancient pips don't support. Long-term, (3) or (1) would be nice. ConfigArgParse has been a continual source of trouble.

bmw:
5) Use environment markers to install `argparse` on Python 2.6 systems (it's in the standard library in Python 2.7+)
6) Drop support for Python 2.6.

My favorite short term fix is 4 or 5, but I think I prefer 5 due to comments like [these](https://github.com/certbot/certbot/pull/2249#issuecomment-210087073). We won't work on ancient versions of `pip`, but we'll work on `certbot-auto` and installing Certbot yourself through `pip` is a bit tricky already due to OS dependencies.

 Long term, I think we should look into 1, 3, and/or 6.

erikrose:
(5)'s problem is that pips<6 don't support env markers, and, when last I checked, pip 1.5.4 was still a major part of the PyPI client pie chart. An old Debian ships with it. Of course, certbot-auto users get a new pip and won't have that problem. And, I suppose, arguably, developer types who install manually with an old pip get what they deserve. Yes, I could see using env markers now that we've guaranteed ourselves pip>=8.

(6) Python 2.6 will live FOREVER! :-)

bmw:
(6) I bet that's how the Python core team has felt about Python 2 in general over the past few years.

(7) This is a bit unconventional, but what if we built our wheel with the tag `py27.py3` and add our previous conditional `argparse` dependency back to `setup.py`? Then we can either not build a py26 wheel (my preference) or build a `py26` only wheel but our dependencies will be specified properly and work on every system.

erikrose:
For (7) we'd open the door to bug reports from people using the versions of pip with the buggy wheel caching who cached a 2.7 wheel at some point and then try to install it on 2.6 or vice versa. Those versions of pip are a lot newer than the 6.0 that (5) would require.

Remind me never to write client-side software in Python again.

bmw:
> Remind me never to write client-side software in Python again.

Seconded.

In that case, I'm leaning towards (5). Is your preference still towards (4)? As I see it, we either unnecessarily depend on `argparse` in Python 2.7 which will lead to an old version of the package being installed on some systems or stop working with old versions of `pip`.

cmskog:
Would like to pitch in my proverbial two cents....
Obviously, I have no insight whatsoever regarding earlier commitments("THIS distro absolutely must be supported !", etc) or the particular trials and tribulations you are facing, but from my point of view(admittedly a very comfy, "outside looking in" position) the rational would be to ONLY support officially Python core supported 3.x version(seems to be 3.2-3.6 at this time). A sane comprise would be officially supported Python core versions, i.e. 3.2-3.6 + 2.7.

erikrose:
I suppose I still lean toward (4) for this reason: with either (4) or (5), we have, effectively, a conditional dependency. But in (4), no matter how old your pip is, the right thing gets done—so, fewer bug reports. The nice thing about (5) is fewer wheels to build, but we should have that scripted anyway.

erikrose:
Though perhaps the most important take-away from this is: why didn't our test suite catch this?

bmw:
My perception of (4) is the right thing will not always get done on py27 which is our most popular platform (as we haven't released anything but `acme` for py3 yet). If we're going to install `argparse` in the py2 wheel, we either:

1. Build the wheel using py26.
2. Depend on `argparse` in our sdist on py27 systems as well.

Option 1 requires people doing releases to have py26 installed on their system (which isn't the end of the world but at least I'd prefer not to) and I think option 2 is a bit of a problem. Not all systems recognize stdlib `argparse` as satisfying the `argparse` requirement. Testing on Ubuntu, Debian, CentOS, Fedora, Arch, and FreeBSD, only Debian based distros seem to do so. This means when installing from `pip`, an old version of `argparse` gets installed. For OS packages, some surgery may be required as described [here](https://github.com/certbot/certbot/pull/2249#issuecomment-259044484).

While this is unnecessary, it may not be a big deal unless PyPI `argparse` is used instead of stdlib `argparse` when it's imported. During my testing, none of the systems I tried had this behavior, however, this is the problem described in the original post. @cmskog, what OS are you using and how did you install Certbot?

cmskog:
I installed with a plain old for python3: "sudo python3 setup.py install" for acme and certbot(the top directory).

erikrose:
bmw is going to see in what way old pips fail with env markers, and that'll inform our decision.

bmw:
I still haven't played with env markers yet, but yet another option here is implement #2208, drop Python 2.6 support (which was dropped by core like 5 years ago), and stop depending on `argparse` entirely.

bmw:
So I think the path forward we want here is to remove Python 2.6 support. I want to do this soon because this helps all sorts of issues, but it'll be a bit tricky and probably won't happen in this release.

In the meantime, I'd like to revert #2249 since it only negatively effects is `pip install` users (who we said we're not going to actively try and support) and stops causing issues with OSes and packaging systems like MacPorts, Arch, and Gentoo. Let's not cause them problems when there is no additional benefit for us.

bmw:
@erikrose, are you OK with the above plan? Sorry for ping but I want to add a dependency on `ordereddict` in Python 2.6 (which we already indirectly have) and want to know which approach to take.

bmw:
I suppose another approach which would work temporarily is environment markers (keeping in mind we're not too worried about people trying to install through `pip`). I actually think I like this option better because then our wheel isn't incorrect, it just only works with new versions of `pip`.

erikrose:
If we're not going to support `pip install` (which, for my own reference, leaves out our several non-Python dependencies like augeus), the next best thing (from the alternatives at https://github.com/certbot/certbot/pull/2249#issuecomment-258426502) is environment markers. Since we support only certbot-auto (which guarantees us a marker-honoring version of pip) and OS packaging (which ignores our dependency metadata altogether), we should be covered.

bmw:
Will environment markers cause problems for OS package maintainers if they're using an older versions of `pip`? If so, I think I take back what I said and prefer the dynamic `install_requires` which will still avoid issues for us but make it easier for package maintainers.

erikrose:
OS package maintainers should at least be taking a look at the source, not just installing and then trying to reverse engineer what happened. If not, they at least shouldn't be using a version of pip from before December 2014 (which is when the first env-marker-using pip came out). If someone is rolling OS packages for Python and doesn't know to look for env markers, shame on them. My 2¢. :-)

yan12125:
I have an idea: how about using get-pip.py on _all_ distributions? (e.g. stay away from system pip packages) The latest pip has dropped Python 2.6 support, while certbot can still use the 2.6 variant of get-pip.py at https://github.com/pypa/get-pip/blob/master/2.6/get-pip.py

erikrose:
> I have an idea: how about using get-pip.py on all distributions?

That's essentially what we do with pipstrap, which has the advantage of not being 2MB and of validating the hash of the downloaded pip.

yan12125:
Sorry I misunderstood the problem here.

> Will environment markers cause problems for OS package maintainers if they're using an older versions of pip?

My experience at Arch Linux is that package maintainers don't use pip to create python packages. In most cases ```python setup.py install``` is used. Seems Debian recommends a similar practice. [1] I guess other distributions are similar, so ancient pip won't be a problem here.

[1] https://wiki.debian.org/Python/AppStyleGuide

ArchangeGabriel:
Yes, using pip for packaging is not a good practice indeed. That’s the problem of modern packaging: every language/framework has it own package manager and trying to get around is hard for a lot of them (think ruby/gem, nodejs/npm, go, etc.). But python/pip is one we can work without quite well, so we do.

ArchangeGabriel:
BTW, in Arch: https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/certbot&id=132652fa4ea90b3037ca06a19eb5a76d7dded715

bmw:
I mentioned the wrong package earlier. I'm worried about the packager using an old version of `setuptools` that doesn't understand environment markers. As far as I know, correct support of this was only added in `setuptools 20.8.0` which barely came out a year ago ([source](https://setuptools.readthedocs.io/en/latest/history.html#v20-8-1)). For example, an older version of `setuptools` is in Ubuntu 16.04 ([source](http://packages.ubuntu.com/xenial/python-setuptools)). With this in mind, I'd like to use dynamic `install_requires` again which of course may not work for people installing with `pip`, but we already said we're not worried about this use case.

If people have a problem with this, I can ping package maintainers on slower moving distros and see if this will be an issue.

yan12125:
@bmw you're right - ancient setuptools is broken with environment markers. Here's the result from setuptools 0.9.8 on CentOS 7.0.1406:
```
[root@0810394879cc acme]# python setup.py build
error in acme setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
```

erikrose:
The nice thing is it fails loudly, so the maintainer won't silently get a wrong selection of packages.

On the subject of dynamic `install_requires`, it's not my favorite, but it's not the end of the world. However, it *is* a big problem if we ship wheels which do the wrong thing, and that's easy to do with such conditionals. So consider that as you move forward.

bmw:
To provide a little more context for anyone who's interested in this thread, `certbot-auto` uses [pipstrap](https://github.com/erikrose/pipstrap) to install newer versions of `pip`, `setuptools`, and `wheel` in a more secure way.

I just confirmed that the version of `setuptools` used here is too old to support environment markers (at least in the way I did it). Running an older system in Docker, I changed `acme/setup.py` from:
```
install_requires = [
    'argparse',
...
```
to
```
install_requires = [
    'argparse ; python_version < "2.7"',
```
I confirmed this works with up to date `pip` and `setuptools`, but after using `pipstrap`, I got:
```
DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
Obtaining file:///root/certbot/acme
Exception:
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/basecommand.py", line 209, in main
    status = self.run(options, args)
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/commands/install.py", line 310, in run
    wb.build(autobuilding=True)
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/wheel.py", line 747, in build
    self.requirement_set.prepare_files(self.finder)
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/req/req_set.py", line 359, in prepare_files
    ignore_dependencies=self.ignore_dependencies))
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/req/req_set.py", line 646, in _prepare_file
    set(req_to_install.extras) - set(dist.extras)
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/_vendor/pkg_resources/__init__.py", line 2810, in extras
    return [dep for dep in self._dep_map if dep]
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/_vendor/pkg_resources/__init__.py", line 2624, in _dep_map
    dm.setdefault(extra,[]).extend(parse_requirements(reqs))
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/_vendor/pkg_resources/__init__.py", line 2980, in parse_requirements
    "version spec")
  File "/root/certbot/venv/lib/python2.6/site-packages/pip/_vendor/pkg_resources/__init__.py", line 2945, in scan_list
    raise RequirementParseError(msg, line, "at", line[p:])
RequirementParseError: Expected version spec in argparse ; python_version < "2.7" at  ; python_version < "2.7"
You are using pip version 8.0.3, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```
With this in mind, I'm going to revert #2249 and opened #4553 so we can reconsider using environment markers at a later date.

Issue #2495, opened 2016-02-17
--dry-run creates persistent keys in /etc/letsencrypt/keys


pde:
It shouldn't... probably they should be deleted.  Maybe they should also be in /tmp though that reduces the realism of the dry run simulation.

bmw:
It also saves the staging account (which I believe you're aware of) and the CSR it creates to obtain the new certificate.

pde:
Yup.  I think keeping the staging account is a lesser-of-two-evils thing, since it's actually useful.  The keys and CSRs don't have a reason for persisting, really.

bmw:
We basically just need to add `if not config.dry_run` around the place we write the keys and CSRs to disk.

dhuang:
Grabbing this one!

PR #4380, opened 2017-03-18
Don't save keys/csr on dry run

dhuang:
Fixes https://github.com/certbot/certbot/issues/2495.

Just log that we generated them without saving them to file.


bmw:
We set up the config for compatibility tests in certbot-compatibility-test/certbot_compatibility_test/util.py

Issue #1684, opened 2015-12-03
Implement a --debug-challenge option


pde:
If the client fails DV, there should be a way to keep the challenges deployed so that the user can go and manually inspect them to see what's failing.

It could be --debug-challenge.  Or it could be the default behaviour if DV fails during --debug.  Or it could be that --debug implies --debug-challenge by default.

SwartzCr:
@bmw wanna add some context to this?

pde:
The most ambitious version of what this flag might do is:

* if a challenge fails, pause and print some information:
  * what the challenge was, and how it is supposed to work
  * information on your current state (eg, "Certbot has made temporary changes to your Apache config to serve this TLS-SNI-01 cert (perhaps printing the config diff or pointing to a file that contains it) when things are correct you should be able to see that cert by doing curl/wget/nmap command XYZ", or "Certbot has deployed an HTTP-01 challenge file at path XYZ, you should be able to inspect it with wget blah")
* offer a prompt to either unpause and quit, or retry the challenge

bmw:
I think instead we should just have a flag that always pauses after challenges are set up but before notifying the ACME server. I'm open to ideas on what this flag should be called.

dhuang:
Working on a less ambitious version of this. Pausing after challenges are setup and suggesting the user use "-v" to debug further.

pde:
Closing this as done. We can make a new ticket if we want the more ambitious versions!

PR #4385, opened 2017-03-19
Add --debug-challenges flag (#1684)

dhuang:
Initial version of https://github.com/certbot/certbot/issues/1684.

Pausing after challenges are setup and suggesting the user use "-v" to debug further.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -956,7 +956,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Show tracebacks in case of errors, and allow certbot-auto "
              "execution on experimental platforms")
     helpful.add(
-        ["certonly", "renew", "run"], "--debug-challenges", action="store_true",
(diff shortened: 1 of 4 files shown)


certbot/cli.py
```diff
@@ -956,6 +956,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Show tracebacks in case of errors, and allow certbot-auto "
              "execution on experimental platforms")
     helpful.add(
+        ["certonly", "renew", "run"], "--debug-challenges", action="store_true",
```
bmw: I'd add `None` as the first element in this list. This changes where `--debug-challenges` is displayed in the output of `--help all`. Currently, it's listed under `certonly`, but it's more of a general flag for a lot of what Certbot does, so I'd use `None` which puts it under "Optional arguments".

dhuang: Fixed

PR #3373, opened 2016-08-04
Multiple vhosts

SwartzCr:
for @bmw or @joohoi to take a look at

PR #4390, opened 2017-03-21
Multiple vhosts

### LINKED ISSUE #1042: support multiple vhosts per file

joohoi:
Support for multiple vhosts.

Outline of the changes:
 - ~VirtualHost object paths now always return the index number~
 - Fetching the ServerName and ServerAliases from raw path are now handled in a new private helper method  to avoid code duplication (for reverse mapping the newly created ssl vhost in _vhost_names_match): _get_vhost_names(self, path)
 - Appends to orig_filename-le-ssl.conf if it already exists (from previous certbot run on different VirtualHost in the same originating file)
 - Does not register reverter action for file creation if the -le-ssl.conf for the file pre-exists, but creates a checkpoint instead. This is to avoid removing the file if operation in cancelled or fails at some point later in the installer.
 - Extracting the VirtualHost from original configuration file is done using augeas span method that gives us the starting and ending byte of the VirtualHost directive. This applies to all VirtualHosts.
 - We are no longer copying the original file over, not even in case of single VirtualHost in a file.
 - Refactored _copy_create_ssl_vhost_skeleton

Built on #3373 and bmw_multiple_vhosts branch

Fixes #1042


bmw:
Hey @sagi, would you be interested in taking a look at this? I plan on doing so as well.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -138,6 +138,7 @@ def __init__(self, *args, **kwargs):
         self._enhance_func = {"redirect": self._enable_redirect,
                               "ensure-http-header": self._set_http_header,
                               "staple-ocsp": self._enable_ocsp_stapling}
+        self._skeletons = {}
```
bmw: `_skeletons` is left over from the PRs this was built on and is no really longer used outside of tests. We should probably remove it.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -962,6 +1069,8 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
                 "the potential to create redirection loops.".format(avail_fp,
                                                                     ssl_fp),
                 reporter.MEDIUM_PRIORITY)
+        self.aug.set("/augeas/files%s/mtime" %(self._escape(ssl_fp)), "0")
+        self.aug.set("/augeas/files%s/mtime" %(self._escape(avail_fp)), "0")
```
joohoi: This was commited by Noah in the original PR: https://github.com/certbot/certbot/pull/3373/commits/7133f0cf8c227b11436a47d5c8e3b055325540e3

I believe the underlying reason was augeas not flagging the file for reload after changes made in rapid succession or something alike.

certbot-apache/certbot_apache/configurator.py
joohoi: Ah, true. Removed now.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -792,24 +846,26 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
...
-            # This simplifies the process
-            vh_p = vh_p[0]
-
+        matches = self.aug.match("/files%s//* [label()=~regexp('%s')]" %
+                                 (self._escape(ssl_fp),
+                                  parser.case_i("VirtualHost")))
+        vh_p = None
+        for match in matches:
+            # Make sure we have the correct ssl vhost
+            if self._vhost_names_match(match, nonssl_vhost):
```
joohoi: This suggestion is now implemented.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -792,24 +846,26 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
...
+                vh_p = match
+
+        # Fallback to at least something
+        if not vh_p and matches:
+            vh_p = matches[0]
```
joohoi: Yeah, erroring out is the sane thing to do here. Implemented.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -875,7 +932,42 @@ def _sift_rewrite_rule(self, line):
...
+        vhost_num = vhost_num-1
+        blocks = [idx for idx, line in enumerate(orig_file_list)
+                     if line.lower().lstrip().startswith("<virtualhost")
+                     or line.lower().lstrip().startswith("</virtualhost")]
+        blocks = blocks[:vhost_num*2] + blocks[(vhost_num*2)+2:]
+        out = self._section_blocks(blocks)
+        return [line for idx, line in enumerate(orig_file_list)
+                          if idx not in out]
+
+    def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp, vhost_num): # pylint: disable=too-many-branches,too-many-locals,too-many-statements
```
joohoi: We should definitely do it as part of the refactoring (in my next reply).

certbot-apache/certbot_apache/configurator.py
```diff
@@ -886,70 +978,85 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
...
-                    for line in orig_file:
-                        A = line.lstrip().startswith("RewriteCond")
-                        B = line.lstrip().startswith("RewriteRule")
-
-                        if not (A or B):
-                            new_file.write(line)
-                            continue
+                orig_file_list = [line for line in orig_file]
+                if vhost_num != -1:
+                    orig_file_list = self._create_block_segments(orig_file_list, vhost_num)
```
joohoi: I can't believe I missed this! You are absolutely right. I played around with some ways of getting the correct vhost, and this suggestion seems to be the best way. However I would like to implement it in to VirtualHost object creation so that every VirtualHost object would read in, and carry their own configuration block through the execution. This will affect the _copy_create_ssl_vhost_skeleton rewrite as well, and will most likely simplify the flow when writing out the new configuration.

joohoi:
You are right, we should probably postpone this PR. The upcoming changes will further affect the vhost creation and surely should be tested throughout before shipping. More details in the inline comment replies above.

joohoi:
Ready for another review! The PR description updated to reflect the changes.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -959,9 +968,100 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
...
+    def _get_vhost_block(self, vhost):
+        """ Helper method to get VirtualHost contents from the original file.
+        This is done with help of augeas span, which returns the span start and
+        end positions
+
+        :returns: `list` of VirtualHost block content lines without closing tag
+        """
+
+        try:
+            span_val = self.aug.span(vhost.path)
```
bmw: I think there may be a problem here. This copies the vhost and ignores any other directives in the file. For example, a user may have:
```
<IfModule mod_macro.c>
    <VirtualHost *:80>
        ....
    </VirtualHost>
</IfModule>
```
When we copy the vhost, we remove these `<IfModule>` lines which seems incorrect. To be clear, this is the case for all directives that start a block not just `<IfModule>`.

One thing I'm not sure about is if directives contained in the block that contains the vhost should be copied. To provide a toy example because to be honest I'm not familiar enough with Apache directives to be aware of a real one:
```
<IfModule mod_magic.c>
    MagicSpell "foo"
    <VirtualHost *:80>
        ....
    </VirtualHost>
</IfModule>
```
Does `MagicSpell "foo"` apply to the Apache config globally or that specific block? If the former, we ideally shouldn't copy them but we probably should if it's the latter. Is it dependent on the directive?

I'm not sure how best to handle this. Perhaps we could copy everything in the highest level block that contains the vhost which shouldn't be too hard using our functions to manipulate Augeas paths and `span` but the two problems I see are:

1. Directives that apply globally we'd be duplicating. We've been doing this for years though by copying the entire file containing the vhost so maybe this isn't a problem?
2. There may be other vhosts in that block which we don't want to copy. We'd have to remove these somehow which will probably be a bit tricky.

I'm curious your thoughts on this.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -959,9 +968,100 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
...
+                    result.append('\n'.join(chunk))
+                    continue
+        return result, sift
+
+    def _get_vhost_block(self, vhost):
+        """ Helper method to get VirtualHost contents from the original file.
+        This is done with help of augeas span, which returns the span start and
+        end positions
+
+        :returns: `list` of VirtualHost block content lines without closing tag
```
joohoi: I think leaving it open will allow us to easily add directives at this point of execution if ever need arises, so I'd rather leave it as it is.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -959,9 +968,100 @@ def _copy_create_ssl_vhost_skeleton(self, avail_fp, ssl_fp):
...
                 "vhost for your HTTPS site located at {1} because they have "
-                "the potential to create redirection loops.".format(avail_fp,
-                                                                    ssl_fp),
-                reporter.MEDIUM_PRIORITY)
+                "the potential to create redirection loops.".format(
+                    vhost.filep, ssl_fp), reporter.MEDIUM_PRIORITY)
+        self.aug.set("/augeas/files%s/mtime" % (self._escape(ssl_fp)), "0")
+        self.aug.set("/augeas/files%s/mtime" % (self._escape(vhost.filep)), "0")
+
+    def _prepare_ssl_skeleton_contents(self, contents):
```
joohoi: I will rename it, `_sift_rewrite_rules` sounds accurate.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -792,24 +827,30 @@ def make_vhost_ssl(self, nonssl_vhost):  # pylint: disable=too-many-locals
...
-            # This simplifies the process
-            vh_p = vh_p[0]
+        new_matches = self.aug.match("/files%s//* [label()=~regexp('%s')]" %
+                                     (self._escape(ssl_fp),
+                                      parser.case_i("VirtualHost")))
+
+        vh_p = self._get_new_vh_path(orig_matches, new_matches)
+
+        # Raise error if we wrote something but can't map it back
+        if not vh_p and len(new_matches) != len(orig_matches):
```
joohoi: This is a remnant from the old, flaky parsing, I will remove it completely.

certbot-apache/certbot_apache/configurator.py
bmw: Works for me!

certbot-apache/certbot_apache/configurator.py
bmw: Let me know how you'd like to chat about this. Happy to do it here, IRC, or a video chat.

bmw: I just wanted to ping you again here. If you'd like to talk about this using something other than messages here, let me know when you're available and I can arrange something this weekend or early next week.

joohoi: The configuration directives inside blocks apply globally, so it's not that big of an issue. However people be overwriting some values in nested blocks containing VirtualHosts. Then again people could similarly have things like:
```
Listen 80
<VirtualHost *:80>
...
</VirtualHost>
```
as an example, which would result to apache failing to address already in use error. So it kind of goes both ways.

I think we should just copy the VirtualHost itself. Linux distributions provide examples for, and encourage people to use a structure where VirtualHosts are sitting on root level of the configuration file. The IfModules and IfDefines are pretty much just in place to prevent loading of certain (default) config blocks if modules are not enabled. 

Also I doubt people play around with the modules a lot. Basically, if you have your configuration running, you do have the modules in anyway, and the point where it would break (in configtest syntax check fail way of breaking) would be when you would disable a module manually later on.

If anything, I would say we'd need to put more effort into active module parsing and not peeking in to configuration structures under disabled IfModule statements, hence not offering the disabled domains as installation candidates by Certbot.

In a world where we wouldn't need to revisit the configuration ever again, or give user a clear indication of the new vhost file we created - we'd create the new VirtualHost in the original file, right beside the original one, hence leveraging all the nested config directives in place.

bmw: You make a good point with multiple `Listen` statements. My plan wasn't to copy these if they were in the outside of an enclosing block, but you could easily imagine someone doing:
```
<IfModule mod_ssl.c>
    Listen 443
    <VirtualHost *:443>
    ...
    </VirtualHost>
</IfModule>
```
Copying the entire block here causes wrong behavior. On the other hand, a user could have something like:
```
<Macro VHost $name $domain>
<VirtualHost *:80>
    ServerName $domain
    ServerAlias www.$domain

    DocumentRoot "/var/www/vhosts/$name"
    ErrorLog "/var/log/httpd/$name.error_log"
    CustomLog "/var/log/httpd/$name.access_log" combined
</VirtualHost>
</Macro>

Use VHost example example.com
Use VHost myhost hostname.org
Use VHost apache apache.org

UndefMacro VHost
```
In this case, only copying the vhost causes the wrong behavior and breaks a configuration we currently work for. I asked on the [community forum](https://community.letsencrypt.org/t/lets-help-certbots-apache-plugin/33194) for other directives like this. It's not immediately clear to me what the solution is here...

joohoi: mod_macro is the only directive I can think of that this would break. However we are specifically NOT supporting it at the moment, and supporting it would need a more fine tuned directive extraction anyway. 

From Apache perspective, the best way to handle this would be to put the newly created VirtualHosts right before or after the already existing ones in the original files, it would make these cases non-issues as well.. Distribution specific configuration schemes and us being as transparent and intuitive as possible disagree with this however.

bmw: `mod_macro` is the only case I could come up with as well and no one has responded to the forum post. It's worth noting that we don't completely ignore vhosts using `<Macro>` directives, as they're still added to our list of vhosts in `get_virtual_hosts()` and can be selected by the user interactively. This is probably a bug though as there are plenty of cases where us copying the vhost using `mod_macro` can go wrong regardless of whether we copy the entire file or just the vhost.

Since I'm not aware of any problems here, if you're confident this is the right approach, we can go forward with this. Thanks for walking me through your thinking.

certbot-apache/certbot_apache/obj.py
```diff
@@ -123,7 +123,7 @@ class VirtualHost(object):  # pylint: disable=too-few-public-methods
     strip_name = re.compile(r"^(?:.+://)?([^ :$]*)")
 
     def __init__(self, filep, path, addrs, ssl, enabled, name=None,
-                 aliases=None, modmacro=False):
+                 aliases=None, modmacro=False, ancestor=None):
```
joohoi: Documentation line is now added.

certbot-apache/certbot_apache/configurator.py
joohoi: Yes, I strongly believe that if this has effect on error ratio, it will be to the favorable direction.

joohoi:
Up for next round of review!

bmw:
This was merged in #4607.

PR #4395, opened 2017-03-22
Save hyphenated plugin params for renewal (#4281)

bmw:
* fix plugin namespace check

* Add test to prevent regressions

(cherry picked from commit b7d282309d9e28221ab557264966d42f4b383529)

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -19,6 +19,9 @@ from certbot import errors
 from certbot import error_handler
 from certbot import util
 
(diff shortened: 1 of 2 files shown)


ohemorange:
This commit has only been out for a few days. The bug has been around for roughly forever I think? I personally don't think it meets the bar for cherry picking, but I'm fine as long as you want to babysit it. I'll approve and let you decide.

bmw:
That's a fair point. Closing.

PR #4396, opened 2017-03-22
Ensure a SHA2 hash algorithm is used when signing releases (#4384)

bmw:
* use gpg2

* explictly use sha256

(cherry picked from commit bf45cea7cddd47e47194904d5f5b47757c622306)


ohemorange:
This doesn't go into the packaged code, right?

bmw:
It doesn't, but the alternative is set up my files like this locally before building the release. Since future commits to this branch containing the signatures are affected by changes here, I have a slight preference for seeing these merged so it is clear how the release was built.

PR #4398, opened 2017-03-22
Change QSA to NE in HTTPS redirection (#4204)

bmw:
* Change QSA to NE in HTTPS redirection

* Seamless transition to new HTTPS redirection RewriteRule

(cherry picked from commit 26a7023b8dd985674dbf9ff2d35d4065ca7cae9d)


bmw:
This will catch the next train.

PR #4409, opened 2017-03-24
Fix unorderable types error

Ngo-The-Trung:
Seen on 0.9.3 and 0.12.0 with python 3.4

I ran into this issue where the `poll_and_request_issuance()` threw an "unorderable type" error because it attempted to compare the AuthenticationResource objects. After inspecting, some elements of the `waiting` heap shows similar datetime.

If I'm not wrong the piece of code that this PR is modifying assumes the datetime of the tuples wouldn't collide. If the datetime collides, the comparator of heapq will move onto the AuthorizationResource value and throws an "unorderable type" error.

This adds an index value to the element tuple to ensure that they are always strictly ordered.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -358,14 +358,18 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
         # priority queue with datetime.datetime (based on Retry-After) as key,
         # and original Authorization Resource as value
-        waiting = [(datetime.datetime.now(), authzr) for authzr in authzrs]
+        waiting = [
+            (datetime.datetime.now(), index, authzr)
+            for index, authzr in enumerate(authzrs)
+        ]
+        heapq.heapify(waiting)
         # mapping between original Authorization Resource and the most
         # recently updated one
         updated = dict((authzr, authzr) for authzr in authzrs)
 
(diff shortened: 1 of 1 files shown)


Ngo-The-Trung:
Oh and generating `datetime.datetime.now()` in a loop was the source of collisions.

Ngo-The-Trung:
Anybody? 😅

acme/acme/client.py
```diff
@@ -358,14 +358,18 @@ def poll_and_request_issuance(
 
         # priority queue with datetime.datetime (based on Retry-After) as key,
         # and original Authorization Resource as value
-        waiting = [(datetime.datetime.now(), authzr) for authzr in authzrs]
+        waiting = [
+            (datetime.datetime.now(), index, authzrs[index])
+            for index in range(len(authzrs))
```
zjs: Might this be clearer with `enumerate`?

`[(datetime.datetime.now(), index, authrz) for index, authrz in enumerate(authzrs)]`

Ngo-The-Trung: Done

Ngo-The-Trung:
@bmw what is your thought on this?

PR #4412, opened 2017-03-24
Lock file part 2

### LINKED ISSUE #933: Implement locking to ensure two clients do not run at once

bmw:
Fixes #933 (again).

Much of the design is the same as what I described in #4369. The differences are:

#### Make the lock file world writeable

This allows the lock to be shared regardless of the permissions used to run Certbot. The way this works is if the path provided with `--lock-path` doesn't exist and the target is not a broken symlink, the file is created with 0666 permissions. If it does exist, its permissions are unchanged. If the target is a broken symlink, we'll refuse to create the file, display a warning, and continue without grabbing the lock. We could potentially create the file with more restricted permissions in this case, but I think this behavior is better than allowing an arbitrary user to create arbitrary files.

#### Switch to portalocker

This is mainly due the change above. `portalocker` allows you to handle creating the file yourself and pass it in to `portalocker` to lock it which now works better for us. `portalocker` is packaged everywhere relevant except EPEL and FreeBSD. I believe this can be easily changed on both places to package Certbot.

Also, `portalocker` uses `flock()` on UNIX and I explained a few reasons why `lockf()` works better for us in #4369, but:

1. Python emulates `flock()` on systems where it's not available ([source](https://docs.python.org/2/library/fcntl.html#fcntl.floc) and [source source](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Modules/fcntlmodule.c#L282)).
2. The `flock()` system call has worked properly over NFS (and by properly I mean the lock isn't limited to the local file system) on Linux since 2005 ([source](http://man7.org/linux/man-pages/man2/flock.2.html)).

~~1. I'm unaware of any UNIX system where `fcntl.flock()` doesn't work in Python. On some systems, this is emulated by a call to `fcntl.fcntl()` ([source](https://docs.python.org/2/library/fcntl.html#fcntl.flock)), but I couldn't find any reports of it not working at all. BSD and Linux have supported the real `flock()` system call since 1983 and no later than 1996 respectively ([source](http://man7.org/linux/man-pages/man2/flock.2.html)). I found reports of it working in Python on Solaris in 2002. The `flock()` system call has worked properly over NFS (and by properly I mean the lock isn't limited to the local file system) on Linux since 2005 ([source](http://man7.org/linux/man-pages/man2/flock.2.html)). While it's possible there's a super old or obscure UNIX system where `flock()` won't work for us, I couldn't find one and not for lack of trying. We will not have a problem here on any modern or popular system.
2. If there is a problem on an obscure system, I've talked to the `portalocker` maintainer and he's willing to add a fallback using `lockf` or `fcntl`. The blocker to doing this now is we have no idea how the error manifests (e.g. `IOError`, `flock()` not defined, etc.).~~

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -9,6 +9,7 @@ import time
 import traceback
 
 import fasteners
+import portalocker
 import zope.component
 
 from acme import jose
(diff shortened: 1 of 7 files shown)


bmw:
Ugh. Closing for now. We ideally want to set up the lock before setting up logging but then we can't use log messages to communicate with the user.

Issue #4253, opened 2017-02-24
Allow detemining whether certs need renewal via exit code


strugee:
Currently the script I use for managing Let's Encrypt renewals has to stop all active web servers because of #255. It'd be nice if I was able to add some logic that checked if any of the certs needed renewal and abort if they don't.

SwartzCr:
this sounds like a pretty useful enhancement.
Correct me if I'm wrong, but what you're looking for is a command that you can run that says "these scripts are within the renewal window and will trigger a challenge if you run certbot renew"
Then you could take the output of that and know whether to stop the servers on your computer and then run `certbot renew`
right? or were you thinking of a different workflow?

strugee:
@SwartzCr that'd be useful, but what's really important is the exit code so I can do e.g.:

```bash
#!/bin/sh

if certbot renew --check-renewal > /dev/null; then
    certbot renew
fi
```

That way you don't have to parse the output, it's just "is there any cert that can be renewed? Yes? Renew 'em all"

SwartzCr:
@strugee got it
@ohemorange wanna put this into a milestone? Seems like a plausible feature

bmw:
As explained in #3262, we decided not to use different exit codes for renewal in favor of a hooks system. For your case, you'd want something like:
```
certbot renew --pre-hook "service myserver stop" --post-hook "service myserver start"
```
These hooks will only run when a certificate has to be renewed. If I'm misunderstanding your use case and this won't work for some reason, please let me know and I'll reopen.

ohemorange:
I agree with Brad. Also, if you really want to do things this way, #4138 should let you answer this and other questions programmatically.

strugee:
Oh yeah, that seems to satisfy my use case. Do hooks do error handling? I.e. if a hook exits with an exit code of 1, will certbot fail accordingly? This isn't documented anywhere.

bmw:
If you cert is due for renewal and a hook fails, it will print the error to `stderr` but still try to renew the cert. A failing hook doesn't cause Certbot to exit with a non-zero error code.

SwartzCr:
@strugee any interest in documenting this? There's now a section for hooks in the documentation I think it'd be nice to mention error handling there. If you're up for writing something I'd be happy to review and merge it!

PR #4418, opened 2017-03-27
Document hook error handling

strugee:
Ref https://github.com/certbot/certbot/issues/4253#issuecomment-283733385; /cc @SwartzCr for review :)

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -378,10 +378,15 @@ then restart it after the plugin is finished. Example::
 
   certbot renew --pre-hook "service nginx stop" --post-hook "service nginx start"
 
-The hooks will only be
-run if a certificate is due for renewal, so you can run this command
-frequently without unnecessarily stopping your webserver. More
-information about renewal hooks can be found by running
+If a hook exits with a non-zero exit code, the error will be printed
+to ``stderr`` but renewal will be attempted anyway. A failing hook
+doesn't directly cause Certbot to exit with a non-zero exit code, but
+if it causes renewal to fail it will have the same effect. Hooks will
+only be run if a certificate is due for renewal, so you can run the
+above command frequently without unnecessarily stopping your
+webserver.
+
+ More information about renewal hooks can be found by running
 ``certbot --help renew``.
 
 If you're sure that this command executes successfully without human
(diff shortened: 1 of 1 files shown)


strugee:
Travis failures are unrelated

SwartzCr:
lgtm but I want @bmw to check off on it before merging

docs/using.rst
```diff
@@ -378,10 +378,15 @@ then restart it after the plugin is finished. Example::
...
   certbot renew --pre-hook "service nginx stop" --post-hook "service nginx start"
 
-The hooks will only be
-run if a certificate is due for renewal, so you can run this command
-frequently without unnecessarily stopping your webserver. More
-information about renewal hooks can be found by running
+If a hook exits with a non-zero exit code, the error will be printed
+to ``stderr`` but renewal will be attempted anyway. A failing hook
+doesn't directly cause Certbot to exit with a non-zero exit code, but
+if it causes renewal to fail it will have the same effect. Hooks will
```
zjs: > if it causes renewal to fail it will have the same effect.

As someone who doesn't know how hook error handling works, I find this phrasing to be a bit confusing.

What is this intended to communicate?

strugee: @zjs so let's say you have some certs that were retrieved using the standalone authenticator. If you have `systemctl stop apache2; false` as your pre-renewal hook, `false` will cause the hook's exit code to be `1` but the renewal will still succeed and Certbot's exit code will be `0`.

However, let's say you typo'd the first command so your pre-renewal hook was e.g. `systemctl stp apache2`. In that case, renewal would fail (because Apache wouldn't stop and the authenticator wouldn't be able to bind to the relevant port) and Certbot's exit code would be `1`.

zjs: I think my confusion is around "it will have the same effect". To me, that implies that Certbot would exit with the same exit code as the failed hook, which I don't believe to be true.

Maybe this could be rephrased to something like:

> If a hook exits with a non-zero exit code the error will be printed to ``stderr``, but renewal will be attempted anyway. The outcome of renewal, not the hook, will determine Certbot's exit code.

strugee: Ah yeah, that makes sense. Will do.

strugee:
Sorry, forgot about this one! Changed the wording around to address @zjs's feedback, plus rebased on `master`.

SwartzCr:
brad is out of the office right now, maybe @schoen wants to give this a second +1?

PR #4422, opened 2017-03-27
Pin python-augeas version to avoid error with 1.0.0

zjs:
When running ./tools/venv.sh with 1.0.0 (now the latest version), I encountered:
    build/temp.linux-x86_64-2.7/augeas.c:434:35: fatal error: augeas.h: No such file or directory

### DIFF
--- a/certbot-apache/setup.py
+++ b/certbot-apache/setup.py
@@ -11,7 +11,7 @@ install_requires = [
     'acme=={0}'.format(version),
     'certbot=={0}'.format(version),
     'mock',
-    'python-augeas',
+    'python-augeas<=0.5.0',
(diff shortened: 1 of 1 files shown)


certbot-apache/setup.py
```diff
@@ -11,7 +11,7 @@
     'acme=={0}'.format(version),
     'certbot=={0}'.format(version),
     'mock',
-    'python-augeas',
+    'python-augeas==0.5.0',
```
bmw: nit: We should probably do `python-augeas<=0.5.0` for the older distros that are using an even older version of the package.

zjs: Makes sense!

Issue #4152, opened 2017-02-02
Document configuration files


SwartzCr:
Certbot has a number of configuration files. Specifically we have a cli.ini file that handles global configuration & files in /etc/letsencrypyt/renewal that are lineage configuration files (relating to #4016)
This also relates to #4096 
These need to be documented!
In theory we only need to document cli.ini since we have other documentation about how to update the renewal configs, but if we decide that we want people to edit these by hand it would be prudent to document them.

SwartzCr:
It sounds like @pde is interested in working on this - via his work on #4096. So I'm tagging him here so that if he chooses to do more expansive documentation we can close this ticket

SwartzCr:
There's also #4407 and #4490 which should be mentioned (for people wanting to edit their own config files)

troy-lamerton:
Is anyone working on this? I could pick it up.

Is it a matter of rewriting the docs in https://certbot.eff.org/docs/api/cli.html for the `cli.ini` file?

bmw:
I think the biggest thing that needs to be done here is expanding https://certbot.eff.org/docs/using.html#configuration-file to describe how it works. The very brief version is setting options there is equivalent to setting them on the command line. If the same option is set both in the `cli.ini` file and the command line, I forget it's behavior.

Just a better description of what you can put in this file and how it works should be added to our documentation.

Issue #4096, opened 2017-01-21
`certbot renew` doesn't work when the /etc/letsencrypt/cli.ini file contains a list of domains.


terrycloth:
**My system setup:**

* Fedora 25 Server (64-bit)
* certbot 0.9.3
* openssl 1.0.2j

**What I did:**

Using the `domains` config line, I successfully generated a certificate for all my domains and subdomains at once, but it apparently causes a problem for renewing them. certbot complains that you can't specify individual domains, even though no domains are specified *on the command line*.

```
↪  sudo certbot renew  --pre-hook "systemctl stop nginx.service"  \
    --post-hook "systemctl start nginx.service"
Saving debug log to /var/log/letsencrypt/letsencrypt.log
No renewals attempted, so not running post-hook
Currently, the renew verb is only capable of renewing all installed
certificates that are due to be renewed; individual domains cannot
be specified with this action. If you would like to renew specific
certificates, use the certonly command. The renew verb may provide
other options for selecting certificates to renew in the future.

<exit code 1>
```

When I comment out the `domains` line in `cli.ini`, `certbot renew` behaves as expected.

```
↪  sudo certbot renew  --pre-hook "systemctl stop nginx.service"  \
    --post-hook "systemctl start nginx.service"
Saving debug log to /var/log/letsencrypt/letsencrypt.log

-----------------------------------------------------------------------
Processing /etc/letsencrypt/renewal/<MY_DOMAIN>.conf
-----------------------------------------------------------------------
Cert not yet due for renewal

The following certs are not due for renewal yet:
  /etc/letsencrypt/live/<MY_DOMAIN>/fullchain.pem (skipped)
No renewals were attempted.
No renewals attempted, so not running post-hook
```

bmw:
This is expected behavior and is a duplicate of #3824 which is now closed. With that said, we could try and improve the error message here to mention a configuration file as suggested in #3824.

terrycloth:
I read #3824, which refers to #2305, which refers to #3615. That last thread was a little confusing to me, but these all seem to drift further and further away from my original point. #3824 asks for a more specific error message, but was closed without getting one. #2305 asks for a way to specify which domains to renew, but was closed in favor of #3615, which does not even implement that, as far as I can tell...

[...]

PR #4427, opened 2017-03-27
No domains in cli.ini

schoen:
This is a small set of changes to the example `cli.ini` to mitigate a recurring bug where people list their domains there with `domains =` and then are unable to renew at all (due to the renewer rejecting `-d`). This tries to make clear that this is for systemwide defaults that apply to every cert on the system.


schoen:
@SwartzCr, I seem to remember that there was somewhere else that I was also supposed to change this in the documentation; do you remember where?

SwartzCr:
@schoen maybe you're referring to #4152 ? I'm not sure if that specifies anywhere else to share this info, but this PR should be mentioned in it

SwartzCr:
I'm going to merge this since I don't see any issues with it. In the future I do hope to write a documentation section about the cli.ini file

pde:
Trying to mitigate #4096 by documenting around it.

PR #4430, opened 2017-03-28
security: preserve permissions on renewal conf

### LINKED ISSUE #4575: Preserve permissions on renewal conf files during renewal
Ensure that permissions are preserved when renewal data is written to conf files. This allows users to limit access to the file, if they wish.

See #4430.

zjs:
Ensure that permissions are preserved when renewal data is written to
conf files. This allows users to limit access to the file, if they wish.

Testing done:
 * `tox -e py27`
 * `tox -e lint`
 * Manual Testing
    * Got a new certificate. Restricted the permissions on the renewal
      conf. Renewed the certificate. Verified that the new renewal conf
      permissions matched.

Resolves #4575.

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -4,6 +4,7 @@ import glob
 import logging
 import os
 import re
+import stat
 
 import configobj
 import parsedatetime
@@ -117,9 +118,19 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
     # TODO: add human-readable comments explaining other available
     #       parameters
     logger.debug("Writing new config %s.", n_filename)
+
+    # Ensure that the file exists
+    open(n_filename, 'a').close()
+
+    # Copy permissions from the old version of the file, if it exists.
+    if os.path.exists(o_filename):
+        current_permissions = stat.S_IMODE(os.lstat(o_filename).st_mode)
+        os.chmod(n_filename, current_permissions)
+
     with open(n_filename, "wb") as f:
         config.write(outfile=f)
     return config
+
 
 def rename_renewal_config(prev_name, new_name, cli_config):
     """Renames cli_config.certname's config to cli_config.new_certname.
--- a/certbot/tests/storage_test.py
+++ b/certbot/tests/storage_test.py
@@ -3,6 +3,7 @@
 import datetime
 import os
 import shutil
+import stat
 import unittest
 
 import configobj
@@ -758,13 +759,16 @@ class RenewableCertTests(BaseRenewableCertTest):
         with open(temp, "w") as f:
             f.write("[renewalparams]\nuseful = value # A useful value\n"
                     "useless = value # Not needed\n")
+        os.chmod(temp, 0o640)
         target = {}
         for x in ALL_FOUR:
             target[x] = "somewhere"
         archive_dir = "the_archive"
         relevant_data = {"useful": "new_value"}
+
         from certbot import storage
         storage.write_renewal_config(temp, temp2, archive_dir, target, relevant_data)
+
         with open(temp2, "r") as f:
             content = f.read()
         # useful value was updated
@@ -775,6 +779,9 @@ class RenewableCertTests(BaseRenewableCertTest):
         self.assertTrue("useless" not in content)
         # check version was stored
         self.assertTrue("version = {0}".format(certbot.__version__) in content)
+        # ensure permissions are copied
+        self.assertEqual(stat.S_IMODE(os.lstat(temp).st_mode),
+                         stat.S_IMODE(os.lstat(temp2).st_mode))
 
     def test_update_symlinks(self):
         from certbot import storage


ohemorange:
A thought: sometimes people switch from using a higher version to a lower one, generally due to things with where they're getting their packages from. This might break things for them, if they ran the current version as root but then later try to run as not root. This might be fine, but we should definitely consider it.

zjs:
> This might break things for them, if they ran the current version as root but then later try to run as not root.

It seems like there are other files/directories that are currently created with similar permissions (e.g., `config/keys`), which would also cause a problem when switching from running as root to running as not root.

In other words: I don't think this would be breaking things in a new way.

ohemorange:
Working on duplicating lineages for `rename`, which will mean copying the renewal conf file -- could definitely use a util function eventually, because permissions aren't copied.

ohemorange:
Actually, I can just call a separate function to copy permissions, ignore me.

bmw:
After thinking about this for a bit, I'm leaning towards us changing the permissions existing files when they are rewritten if they are being updated from an older version of Certbot. If we don't do anything, I'm worried we will be leaving people silently insecure if they switch an existing lineage from something like the `manual` plugin to one of the new DNS plugins. If we start printing a warning message, we either try to detect whether the file needs to be more secure in an error prone way or print the warning message to all users, almost all of whom will be unaffected by what we do here.

I'm leaning towards doing this based on the version in the renewal configuration file. If the version is less than the version we make this change in (or the version doesn't exist as was the case in much older renewal conf files), we `chmod` the file. The problem with this is approach is identified by Erica above where sometimes people downgrade their Certbot version when switching to OS packages. This can cause us to `chmod` a file multiple times which will be annoying for any users relying on this behavior. Alternatively, we can set another flag in the renewal configuration file but I'd rather not do this for a change we're not sure is going to negatively affect anyone. We've never documented the permissions of our files anywhere and it'd be a somewhat strange use case to be relying on this behavior.

With that said, I reached out to someone I know who has built a service around Certbot and asked @schoen who designed `/etc/letsencrypt` for his thoughts. I also think this PR should get two reviews due to it's potential to cause problems.

zjs:
Updated as suggested by @bmw in https://github.com/certbot/certbot/pull/4430#issuecomment-295473307.

The behavior is now:

- Use "safe" permissions for new files.
- Use "safe" permissions when updating a file written by a pre-0.14.0 version of certbot.
- Retain permissions when updating a file written by cerbot 0.14.0 or later.

certbot/storage.py
```diff
@@ -93,6 +94,13 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
 
     """
     config = configobj.ConfigObj(o_filename)
+    if not "version" in config \
```
bmw: nit: You could probably simplify this logic a bit by using `.get()` on `config` and providing an optional parameter of either `None` or a version number lower than 0.14.0.

This logic works as is though and I'll leave it up to you whether you make this change.

certbot/storage.py
```diff
@@ -117,10 +125,24 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
...
+    open(n_filename, 'a').close()
+
+    # Copy permissions from the old version of the file. If writing a new file, or if the file is
+    # old, ensure that the file can only be accessed by the user. (Equivalent to `chmod go-rwx`.)
+    current_permissions = stat.S_IMODE(os.lstat(o_filename).st_mode)
+    if restrict_permissions:
+        new_permissions = current_permissions & ~stat.S_IRWXG & ~stat.S_IRWXO
+    else:
+        new_permissions = current_permissions
+    os.chmod(n_filename, new_permissions)
```
bmw: nit: Any reason to always `chmod` the file rather than put this in the `if restrict_permissions` block?

certbot/storage.py
```diff
@@ -93,6 +94,13 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
 
     """
     config = configobj.ConfigObj(o_filename)
+    if not "version" in config \
+            or not config["version"] \
+            or util.get_strict_version(config["version"]) < util.get_strict_version("0.14.0"):
+        restrict_permissions = True
+    else:
+        restrict_permissions = False
```
bmw: nit: No test coverage.

certbot/storage.py
```diff
@@ -117,10 +125,24 @@ def write_renewal_config(o_filename, n_filename, archive_dir, target, relevant_d
...
+    # Ensure that the file exists
+    open(n_filename, 'a').close()
+
+    # Copy permissions from the old version of the file. If writing a new file, or if the file is
+    # old, ensure that the file can only be accessed by the user. (Equivalent to `chmod go-rwx`.)
+    current_permissions = stat.S_IMODE(os.lstat(o_filename).st_mode)
+    if restrict_permissions:
+        new_permissions = current_permissions & ~stat.S_IRWXG & ~stat.S_IRWXO
+    else:
+        new_permissions = current_permissions
```
bmw: nit: No test coverage.

certbot/storage.py
zjs: During each renewal, a new renewal conf file is created (and, later, renamed to have the same name as the existing file, so that updates are atomic).

Initially, the new file's permissions seem to be based on the `umask`.

The `else` case ensures that the old file's permissions are preserved (from the user's point of view) even though a new file is created as an implementation detail.

bmw: Ah makes sense! Thanks for the explanation.

bmw:
Proposal: Since we decided to keep renewal configuration files publicly shareable with our new approach to DNS plugins, I don't think it's worth the potential problems for users changing our permissions on these files. Let's keep the code that ensures permissions of current files are reused and add tests for it, but let's remove the rest of the code in this PR.

PR #4432, opened 2017-03-28
Document tests/integration/_common.sh.

jsha:
This makes it much easier to run certbot during development.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -39,17 +39,17 @@ Then in each shell where you're working on the client, do:
 .. code-block:: shell
 
    source ./venv/bin/activate
+   . tests/integration/_common.sh
 
 After that, your shell will be using the virtual environment, and you run the
-client by typing:
+client by typing `certbot` or `certbot_test`. The latter is an alias that
+includes several flags useful for testing. For instance, it sets various output
+directories to point to /tmp/, and uses non-privileged ports for challenges, so
+root privileges are not required.
 
-.. code-block:: shell
-
-   certbot
-
-Activating a shell in this way makes it easier to run unit tests
-with ``tox`` and integration tests, as described below. To reverse this, you
-can type ``deactivate``.  More information can be found in the `virtualenv docs`_.
+Activating a shell with `venv/bin/activate` sets environment variables so that
+Python pulls in the correct versions of various packages needed by Certbot.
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -39,17 +39,17 @@ Then in each shell where you're working on the client, do:
 .. code-block:: shell
 
    source ./venv/bin/activate
+   . tests/integration/_common.sh
```
zjs: Should this use `source` instead of `.` for consistency with the above line? (Or vice versa.)

jsha: Good point, fixed.

SwartzCr:
so would you recommend that @jsha remove that section?

jsha:
One option would be to add an `export SERVER=https://acme-staging.api.letsencrypt.org/directory` to that section.

FWIW, I actually think it's much nicer to develop against a local Boulder, but I do see how that's one more step and probably most contributors are not doing it.

zjs:
Maybe it would be worth having a sub-section about developing against a local Boulder?

That would give an "easy mode" (the instructions as they exist today) and an "advanced mode" (local Boulder, certbot_test).

That would allow developers to opt-in to the complexity.

I might normally be wary of having multiple ways of doing things, but it seems like there's value in having a low-friction process for first-time/one-time contributors and value in making repeat contributors as efficient as possible.

jsha:
> That would give an "easy mode" (the instructions as they exist today) and an "advanced mode" (local Boulder, certbot_test).

I *think* the instructions as they exist today are not "easy mode" because you have to figure out a big long set of flags to pass, right? Or do you just run as root?

My goal with this change is to promote certbot_test (or something similar) as an "easy mode." If that means making the default instructions use the staging server, I'm fine with that.

FWIW, there's already a bit of a section about running against Boulder: https://certbot.eff.org/docs/contributing.html#integration.

zjs:
> My goal with this change is to promote certbot_test (or something similar) as an "easy mode." If that means making the default instructions use the staging server, I'm fine with that.

I see. I think that makes sense.

> FWIW, there's already a bit of a section about running against Boulder: https://certbot.eff.org/docs/contributing.html#integration.

I think this is a good start that could be expanded to cover running manual tests against a local bounder as well.

PR #4443, opened 2017-03-30
Logging setup before argument parsing

### LINKED ISSUE #3148: No handlers could be found for logger "certbot.crypto_util"

bmw:
Fixes #3148.

This fixes an old problem with code logging messages before logging has been set up. How this works is explained in the docstring of certbot.log.pre_arg_setup. I apologize for the size of the PR, although a lot of it is testing and moving existing code to other modules.

@joohoi or @ohemorange, can you take a look at this? I'm planning on rebasing my locking code for 0.13.0 on top of this PR.


ohemorange:
I don't suppose you'd want to split this out into a refactor and a change? Both for logical consistency and for tracing what happened here.

bmw:
Fair point. I split this into a refactor PR and a behavior change PR based on that.

PR #4444, opened 2017-03-30
Refactoring for better logging

bmw:
First part of #4443. There should be no behavioral changes here but this does the refactoring to better support the other changes I made in #4443.

### DIFF
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -27,7 +27,7 @@ def pre_arg_parse_setup():
     sys.excepthook = functools.partial(except_hook, config=None)
 
 
-def post_arg_setup(config):
+def post_arg_parse_setup(config):
     """Setup logging after command line arguments are parsed.
 
     :param certbot.interface.IConfig config: Configuration object
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -1,8 +1,10 @@
 """Logging utilities for Certbot."""
 from __future__ import print_function
+import functools
 import logging
 import os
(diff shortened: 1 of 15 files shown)


certbot/log.py
```diff
@@ -0,0 +1,180 @@
...
+
+# Logging format
+CLI_FMT = "%(message)s"
+FILE_FMT = "%(asctime)s:%(levelname)s:%(name)s:%(message)s"
+
+
+logger = logging.getLogger(__name__)
+
+
+def pre_arg_setup():
```
ohemorange: nit: `pre_arg_parse_setup` and `post_arg_parse_setup`?

certbot/colored_logging.py
```diff
@@ -1,45 +0,0 @@
-"""A formatter and StreamHandler for colorizing logging output."""
-import logging
-import sys
-
-from certbot import util
-
-
-class StreamHandler(logging.StreamHandler):
```
ohemorange: Would you mind terribly if this stayed in its own file? We do like random bits of OOO sprinkled throughout, and it's nice for them to stay in their own files like a proper OOO paradigm.

bmw: In #4446, I define two more of these simple handler classes. Would you like to see them in their own file as well? If so, I'll make this change and fix up the other PR too.

ohemorange: oh then nope, this way is better.

PR #4446, opened 2017-03-30
Logging setup before argument parsing

### LINKED ISSUE #3148: No handlers could be found for logger "certbot.crypto_util"

bmw:
Second part of #4443. Built on #4444. Fixes #3148.

This fixes an old problem with code logging messages before logging has been set up. How this works is explained in the docstring of certbot.log.pre_arg_setup.

I'm planning on rebasing my locking code for 0.13.0 on top of this PR.

### DIFF
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -204,7 +204,7 @@ class MemoryHandler(logging.handlers.MemoryHandler):
         return False
 
 
-def except_hook(exc_type, exc_value, unused_trace, debug, log_path):
+def except_hook(exc_type, exc_value, trace, debug, log_path):
     """Logs fatal exceptions and reports them to the user.
 
     If debug is True, the full exception and traceback is shown to the
@@ -213,18 +213,20 @@ def except_hook(exc_type, exc_value, unused_trace, debug, log_path):
 
     :param type exc_type: type of the raised exception
     :param BaseException exc_value: raised exception
+    :param traceback trace: traceback of where the exception was raised
     :param bool debug: True if the traceback should be shown to the user
     :param str log_path: path to file or directory containing the log
 
     """
+    exc_info = (exc_type, exc_value, trace)
     # constants.QUIET_LOGGING_LEVEL or higher should be used to
     # display message the user, otherwise, a lower level like
     # logger.DEBUG should be used
     if debug or not issubclass(exc_type, Exception):
         assert constants.QUIET_LOGGING_LEVEL <= logging.ERROR
-        logger.exception('Exiting abnormally:')
+        logger.error('Exiting abnormally:', exc_info=exc_info)
     else:
-        logger.debug('Exiting abnormally:', exc_info=True)
+        logger.debug('Exiting abnormally:', exc_info=exc_info)
         if issubclass(exc_type, errors.Error):
             sys.exit(exc_value)
         print('An unexpected error occurred:', file=sys.stderr)
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -66,6 +66,8 @@ def pre_arg_parse_setup():
     root_logger.addHandler(memory_handler)
     root_logger.addHandler(stream_handler)
 
+    # logging.shutdown will flush the memory handler because flush() and
+    # close() are explicitly called
     util.atexit_register(logging.shutdown)
     sys.excepthook = functools.partial(
         except_hook, debug='--debug' in sys.argv, log_path=temp_handler.path)
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -75,7 +75,7 @@ def post_arg_parse_setup(config):
     logs_dir = os.path.dirname(file_path)
 
     root_logger = logging.getLogger()
-    assert len(root_logger.handlers) == 2, "Expected handlers not found!"
+    assert len(root_logger.handlers) == 2, "Unexpected handlers found!"
     # pylint: disable=unbalanced-tuple-unpacking
     if isinstance(root_logger.handlers[0], MemoryHandler):
         memory_handler, stderr_handler = root_logger.handlers
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -38,9 +38,8 @@ def pre_arg_parse_setup():
     to properly log/display fatal exceptions.
 
     """
-    temp_fd, temp_path = tempfile.mkstemp()
-    temp_fd = os.fdopen(temp_fd, 'w')
-    temp_handler = logging.StreamHandler(temp_fd)
+    temp_log = tempfile.NamedTemporaryFile('w', delete=False)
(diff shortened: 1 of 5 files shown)


certbot/tests/log_test.py
```diff
@@ -27,12 +26,33 @@ def _call(cls, *args, **kwargs):
         return pre_arg_parse_setup(*args, **kwargs)
 
     def test_it(self):
-        with mock.patch('certbot.log.except_hook') as mock_except_hook:
-            with mock.patch('certbot.log.sys') as mock_sys:
-                self._call()
-
+        with mock.patch('certbot.log.util.atexit_register') as mock_register:
```
ohemorange: Given that there's now four of these, decorate them? Also oh my god because we use the value of `mock_sys` outside of the with, what?

certbot/log.py
```diff
@@ -23,36 +24,82 @@
...
+    stream_handler = ColoredStreamHandler()
+    stream_handler.setFormatter(logging.Formatter(CLI_FMT))
+    stream_handler.setLevel(constants.QUIET_LOGGING_LEVEL)
+
+    root_logger = logging.getLogger()
+    root_logger.setLevel(logging.DEBUG)  # send all records to handlers
+    root_logger.addHandler(memory_handler)
+    root_logger.addHandler(stream_handler)
+
+    util.atexit_register(logging.shutdown)
```
ohemorange: Will this still automatically flush the memory handler even if `shouldFlush` is set to always be `False`?

certbot/log.py
```diff
@@ -127,54 +171,125 @@ def format(self, record):
...
+        :rtype: bool
 
-    Config is used to determine how to display exceptions to the user. In
-    general, if config.debug is True, then the full exception and traceback is
-    shown to the user, otherwise it is suppressed. If config itself is None,
-    then the traceback and exception is attempted to be written to a logfile.
-    If this is successful, the traceback is suppressed, otherwise it is shown
-    to the user. sys.exit is always called with a nonzero status.
+        """
+        return False
```
ohemorange: I mean, if we wanted to be fancy, we could auto-flush once we're in non-temp-handler-land. On the one hand, it's cleaner to not do inspection of the type of target we have. On the other hand, do we really want to be sitting there at the very end having to flush everything?

We could set a property of the object when we hand it off, and condition `shouldFlush` on that.

certbot/main.py
```diff
@@ -714,37 +714,24 @@ def set_displayer(config):
...
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
 
-    make_or_verify_needed_dirs(config)
-
     log.post_arg_parse_setup(config)
-
-    _post_logging_setup(config, plugins, cli_args)
-
+    make_or_verify_needed_dirs(config)
```
bmw: I moved creation of the log directory to the `post_arg_parse_setup`. I personally think `post_arg_parse_setup` should be called as soon as possible so we can stop buffering everything in memory and display lower level logging messages to the terminal if requested by the user.

certbot/log.py
```diff
@@ -23,36 +24,82 @@
 
 
 def pre_arg_parse_setup():
-    """Ensures fatal exceptions are logged and reported to the user."""
-    sys.excepthook = functools.partial(except_hook, config=None)
+    """Setup logging before command line arguments are parsed.
+
+    Terminal logging is setup using
```
bmw: Instead I added a more general description to the top of the module and made this docstring slightly more specific. I think we need to reference `post_arg_parse_setup` here though because the two functions work together closely and the location of the log file depends on whether or not `post_arg_parse_setup` is called.

certbot/log.py
bmw: Oh I wanted to get fancy. Otherwise, I believe the order of logging output gets all messed up. I added more to the docstring of the `MemoryHandler` class and inline comments to `*_arg_parse_setup` to better explain what's going on here.

certbot/log.py
```diff
@@ -127,54 +171,125 @@ def format(self, record):
...
+
+        """
+        self.acquire()
+        try:
+            self.stream.close()
+            if delete:
+                os.remove(self.path)
+            if sys.version_info < (2, 7):  # pragma: no cover
+                logging.StreamHandler.close(self)
+            else:
```
bmw: I added an inline comment about this, but apparently not close enough to your comment to make GH happy.

certbot/tests/log_test.py
bmw: Haha. I made this change, but yeah Python doesn't have a problem with it. The scope of variables is only limited to functions, classes, and modules. So for example, all of these things are valid:
```
def foo():
    if condition:
        var = True
    else:
        var = False
    print var 
```
```
with open('foo.txt', 'w') as f:
    f.write('hi')
print f
```
```
nums = [i for i in range(10)]
print i
```
This is strange behavior for people familiar with other languages like C/C++ though so we probably shouldn't do it.

ohemorange: I knew it was a thing for `if`s, but I thought the whole point of the `with` was to clean things up at the end, making this scope pattern rather absurd...

certbot/log.py
ohemorange: Super clear now, thanks!

certbot/log.py
```diff
@@ -138,13 +154,16 @@ class ColoredStreamHandler(logging.StreamHandler):
...
+    standard `logging.StreamHandler`. Default red_level is
+    `logging.WARNING`.
 
     :ivar bool colored: True if output should be colored
     :ivar bool red_level: The level at which to output
 
     """
     def __init__(self, stream=None):
+        # logging handlers use old style classes in Python 2.6 so
+        # super() cannot be used
```
ohemorange: lol fantastic.

certbot/tests/log_test.py
bmw: If you're REALLY curious you can read [PEP 343](https://www.python.org/dev/peps/pep-0343/) which describes the motivation behind the `with` statement and how it works, but it's really more like a `try ... finally` block. For the file example, it's equivalent to something like:
```
f = open('foo.txt', 'w')
try:
    # imagine I'm the body of the with statement
finally:
    f.close()
```
Note this isn't exactly correct and there's code that describes how to exactly how to mimic the behavior in the above PEP.

bmw:
Awesome. Thanks for the review!

PR #4451, opened 2017-03-31
Fix string test bug

cpbills:
${variable+x} is always going to evaluate to 'x', which will always be a
non-zero length string. I do not believe this was the intention of this
test.

### DIFF
--- a/certbot-auto
+++ b/certbot-auto
@@ -131,7 +131,7 @@ su_sudo() {
 
 SUDO_ENV=""
 export CERTBOT_AUTO="$0"
-if [ -n "${LE_AUTO_SUDO+x}" ]; then
+if [ -n "$LE_AUTO_SUDO" ]; then
   case "$LE_AUTO_SUDO" in
     su_sudo|su)
       SUDO=su_sudo


certbot-auto
```diff
@@ -131,7 +131,7 @@ su_sudo() {
 
 SUDO_ENV=""
 export CERTBOT_AUTO="$0"
-if [ -n "${LE_AUTO_SUDO+x}" ]; then
```
zjs: > ${variable+x} is always going to evaluate to 'x'

`${LE_AUTO_SUDO+x}` evaluates to `x` if and only if `$LE_AUTO_SUDO` is set [1].

That seems to be the intended behavior of this line; if `$LE_AUTO_SUDO` is set, we use its value to set `$SUDO`, otherwise we determine the appropriate value for `$SUDO`  based on the presence of `sudo`.

1 - For more information, see: www.tldp.org/LDP/abs/html/parameter-substitution.html#PARAMALTV

cpbills: No. I caught the error in my initial audit of the script.

cpbills: Thanks for the explanation. I only tested with a defined variable. Is that meant to support being called by some other script that has pre-set some variables? I'll have to dig around and look at the rest of this codebase.

cpbills: Is there a reason this isn't defined via a command-line argument? I'm very curious now, and resisting the urge to try to 'improve' things.

PR #4458, opened 2017-04-04
Sort OS-with-packages list and add NetBSD

truist:


SwartzCr:
Thanks for the update @truist !

PR #4459, opened 2017-04-04
Improved info for FreeBSD

St-Ranger:
1. Replace the outdated reference to "vritualenv", "see below" -> "see above".
2. Replace the awkward sentence (partially incorrect) about installing on FreeBSD. 
It can be installed via different ways: "pkg" is not exclusive. I am not aware of any reason why it can **not** be installed from ports. Some people prefer build and install everything from ports. (The previous version of the description implies that ports cannot be used.)

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -427,7 +427,7 @@ For squeeze you will need to:
 FreeBSD
 -------
 
-Packages can be installed FreeBSD using ``pkg``, 
+Packages can be installed on FreeBSD using ``pkg``, 
 or any other port-management tool (``portupgrade``, ``portmanager``, etc.) 
 from the pre-built package or can be built and installed from ports. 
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -427,8 +427,12 @@ For squeeze you will need to:
 FreeBSD
 -------
 
-Package installation for FreeBSD uses ``pkg``, not ports.
+Packages can be installed FreeBSD using ``pkg``,
```
SwartzCr: I think you're missing a word here - it should be 'can be installed [on|for] FreeBSD'
If you fix that I'd be happy to merge (sorry for the slow review)

St-Ranger: Sorry and thanks! Done.
As the rule says: "Check to see if you any words out."

SwartzCr:
Yay! Thanks!

PR #4464, opened 2017-04-05
Except Exception in rename instead of specific exceptions

### LINKED ISSUE #4469: Improve rename testing and error checking
- un-revert rename
- except Exception as e
- remove dead code
- stop if the existing cert doesn't exist
- what if there's no plugin? how to fail?
- what if it's not already installed?
- don't delete whatever's specified by the new_cert name if it already exists!

ohemorange:
Fixes #4469


ohemorange:
This is in-progress and should not be reviewed.

ohemorange:
Closing this because there's a bunch of changes that need to be made around renaming

PR #4466, opened 2017-04-05
Add Arch Linux constants for Apache

### LINKED ISSUE #4461: Cannot find Apache control command apache2ctl on Arch Linux
I have this problem when using certbot for apache following this guide http://letsencrypt.readthedocs.io/en/latest/using.html#apache

```
$ sudo certbot --apache
Failed to find apache2ctl in expanded PATH: /usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/sbin
The error was: NoInstallationError('Cannot find Apache control command apache2ctl',)
```
I am on Arch Linux, and it seems that there is no apache2ctl for Arch but certbot requires it?

I tried to installed it but no luck:

```
$ sudo pacman -S apache2
error: target not found: apache2
```

My Apache's version:

```
$ httpd -v
Server version: Apache/2.4.25 (Unix)
Server built:   Dec 20 2016 13:02:2
```

yan12125:
Fixes #4461.

CLI_DEFAULTS_ARCH is modified from CLI_DEFAULTS_CENTOS. CentOS is another distro that uses apachectl everywhere instead of apache2ctl. I'm not sure whether all fields are correct or not. The only information is that Arch does as little patch as possible when packaging, so it's likely they're valid. Arch's packaging script at https://git.archlinux.org/svntogit/packages.git/tree/trunk?h=packages/apache can help if there are questions. ```build()``` and ```package()``` in ```PKGBUILD``` are the most important parts.

Tested on 26 Feb on my own Arch machine. I don't have much time these days for verifying this patch again, sorry.

### DIFF
--- a/certbot-apache/certbot_apache/constants.py
+++ b/certbot-apache/certbot_apache/constants.py
@@ -110,6 +110,24 @@ CLI_DEFAULTS_SUSE = dict(
     MOD_SSL_CONF_SRC=pkg_resources.resource_filename(
         "certbot_apache", "options-ssl-apache.conf")
 )
+CLI_DEFAULTS_ARCH = dict(
+    server_root="/etc/httpd",
+    vhost_root="/etc/httpd/conf",
+    vhost_files="*.conf",
+    logs_root="/var/log/httpd",
+    version_cmd=['apachectl', '-v'],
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks!

PR #4479, opened 2017-04-08
"Renew" now ignores domains in cli.ini

### LINKED ISSUE #4096: `certbot renew` doesn't work when the /etc/letsencrypt/cli.ini file contains a list of domains.

pde:
Fixes #4096

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -517,6 +517,7 @@ class HelpfulArgumentParser(object):
 
     def remove_config_file_domains_for_renewal(self, parsed_args):
         """Make "certbot renew" safe if domains are set in cli.ini."""
+        # Works around https://github.com/certbot/certbot/issues/4096
         if self.verb == "renew":
             for source, flags in self.parser._source_to_settings.items(): # pylint: disable=protected-access
                 if source.startswith("config_file") and "domains" in flags:
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -515,6 +515,12 @@ class HelpfulArgumentParser(object):
 
         return usage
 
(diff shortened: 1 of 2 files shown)


pde:
@zjs I don't think there are any doc changes required. #4427 was inspired by the same bug as this PR, but this PR doesn't make any of it inaccurate.

Issue #4484, opened 2017-04-10
certbot-auto doesn't know that "renew" should be non-interactive


kiranmr:
## My operating system is (include version): 
OS: Ubuntu 14.04.2 LTS, Trusty Tahr
certbot: 0.13.0

## I installed Certbot with (certbot-auto, OS package manager, pip, etc): 
certbot-auto downloaded from https://dl.eff.org/certbot-auto

## I ran this command and it produced this output: 
Command: `certbot-auto renew --pre-hook "docker ps -q -f name=web|xargs docker stop"` from a cronjob
Output: Script got stuck in apt-get 
```
 \_ /bin/sh /usr/local/sbin/certbot-auto renew --pre-hook docker stop Controller_contro
     \_ apt-get install --no-install-recommends python python-dev python-virtualenv gcc
```

apt-get is ran without -y  option, so its stuck waiting for user input.

## Certbot's behavior differed from what I expected because:
I expected cert-bot to install required packages non-interactively and continue. `-auto` suffix should  have ensured that apt-get is called with `-y` option

## Root cause
I analysed code of script https://github.com/certbot/certbot/blob/master/certbot-auto
`ASSUME_YES` flag is set only if script name is `letsencrypt-auto`
```shell
if [ $BASENAME = "letsencrypt-auto" ]; then
  # letsencrypt-auto does not respect --help or --yes for backwards compatibility
  ASSUME_YES=1
  HELP=0
fi
```

I think above code should be changed to include `certbot-auto` also.

SwartzCr:
I'm not entirely sure that's the problem, but I agree that this seems like a bug. Maybe it has to do with the newly added NO_BOOTSTRAP option?
Tagging @bmw and @ohemorange to see what their thoughts are

ohemorange:
Hi @kiranmr. How did you get a cert without having those packages installed? We assume that all packages are properly installed before running certbot from from the crontab. We don't want to be automatically installing packages silently on the user's behalf.

SwartzCr:
I wonder if it's due to the dependency @erikrose added for argparse? https://github.com/certbot/certbot/pull/2249/files

ohemorange:
Looks like it's getting stuck on os package install, not python dependencies.

SwartzCr:
maybe @bmw can puzzle this out?
It was mentioned (in conversation between me and @ohemorange) that this means that we can never include a new OS level dependency, since certbot-auto doesn't default to yes

[...]

PR #4486, opened 2017-04-10
Set ASSUME_YES=1 when script name is certbot-auto

kiranmr:
Fix for issue https://github.com/certbot/certbot/issues/4484


SwartzCr:
I think we don't actually want to assume yes for people on Certbot-auto. I'm going to close this while we discuss it more in #4484

PR #4494, opened 2017-04-12
Pin Python package dependencies. Close #4433.

### LINKED ISSUE #4433: Should we pin dependency versions?
Currently we pin specific versions of all our of Python dependencies in `certbot-auto` but don't do this in our PyPI packages. Should we change this? I'd be very interested to see what other popular client side Python projects with OS dependencies do. (I'm not personally aware of any and over the past 2 years I've learned why.)

There was discussion of this in IRC between @erikrose, @ohemorange, and @zjs, but I didn't get a chance to respond while everyone was still in the channel. Thinking about this for a bit, I think I'm against pinning while we have our current distribution methods, but let's see where this helps us.

#### certbot-auto

We already pin specific dependency versions and check its hash.

#### OS packages

While we're currently relying and encouraging systems to package Certbot and for people to use those packages, pinning specific versions of our dependencies isn't feasible. While it'd be nice to only support only one version of our dependencies at a time, the packages available in different distros vary wildly and we need to support multiple versions. Pinning specific versions makes testing different package versions harder for us (see [tox.ini](https://github.com/certbot/certbot/blob/master/tox.ini#L37) and #4049) and makes OS packaging harder. I see no benefit here.

Pinning `dependency<next.major.version` is interesting. This makes it clear that we don't expect Certbot to work with new major versions of our dependencies. If Certbot or its tests are run with a new dependency version though, it'll definitely fail if we pin, but it may not if we don't. Which behavior do we want?

#### pip install

We never tell anyone to install this way and I am perfectly fine not supporting this installation method as was suggested in IRC. In general, I think Python does a poor job supporting `pip` installing packages with OS dependencies which we have many of. Unless you have an intimate understanding of the project and understand you need to install `libffi-dev`, `gcc`, `libaugeas`, and more (or luckily happen to have those packages installed), `pip install` isn't going to work.

#### developer setup

`tools/venv.sh` failing slows down developers, but:

1. If this is our primary motivation for this, we can fix it without affecting packaging.
2. It's beneficial for us to know when a dependency changes that breaks Certbot (although we could test this during CI or nightly testing).

erikrose:
This keeps backward-incompatible versions from sneaking in and causing crashes for (new) devs when using venv.sh.

Allowed newer versions (than in certbot-auto) of some things that are already published, figuring we have already accidentally tested them in the real world. We can always tighten later if we continue to get bug reports from devs.

I'm afraid to pin pytz because (1) their versioning scheme cannot possibly be construed as semver and (2) it would damn us to old timezone definitions, which would be bad. Libraries which essentially provide mutable-over-time datasets are Special, I posit.

Don't pin versions of the dev dependencies, since we don't import and use those; they're tools that are largely decoupled from assumptions in our codebase. I could hear arguments about nose.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,9 +35,9 @@ setenv =
 # https://github.com/shazow/urllib3/pull/930
 deps =
     py{26,27}-oldest: cffi<=1.7
-    py{26,27}-oldest: cryptography==0.8
+    py{26,27}-oldest: cryptography==1.8.1
     py{26,27}-oldest: configargparse==0.10.0
-    py{26,27}-oldest: PyOpenSSL==0.13
+    py{26,27}-oldest: PyOpenSSL==16.2.0
     py{26,27}-oldest: requests<=2.11.1
 
 [testenv:py27_install]
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -8,24 +8,24 @@ version = '0.14.0.dev0'
 
 # Please update tox.ini when modifying dependency version requirements
 install_requires = [
-    'argparse',
+    'argparse<2.0',
     # load_pem_private/public_key (>=0.6)
     # rsa_recover_prime_factors (>=0.8)
-    'cryptography>=0.8',
+    'cryptography>=0.8,<2.0',
     # Connection.set_tlsext_host_name (>=0.13)
-    'mock',
-    'PyOpenSSL>=0.13',
-    'pyrfc3339',
+    'mock<3.0',
+    'PyOpenSSL>=0.13,<17.0',
+    'pyrfc3339<2.0',
     'pytz',
     # requests>=2.10 is required to fix
     # https://github.com/shazow/urllib3/issues/556. This requirement can be
     # relaxed to 'requests[security]>=2.4.1', however, less useful errors
     # will be raised for some network/SSL errors.
-    'requests[security]>=2.10',
+    'requests[security]>=2.10,<3.0',
     # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
     # will tolerate; see #2599:
     'setuptools>=1.0',
-    'six',
+    'six<2.0',
 ]
 
 dev_extras = [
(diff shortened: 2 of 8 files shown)


erikrose:
If the first commit passes, I'll push another that updates tox.ini.

pde:
Saying this again so Github parses it.. closes: #4433

tox.ini
```diff
@@ -35,9 +35,9 @@ setenv =
 # https://github.com/shazow/urllib3/pull/930
 deps =
     py{26,27}-oldest: cffi<=1.7
-    py{26,27}-oldest: cryptography==0.8
+    py{26,27}-oldest: cryptography==1.8.1
```
ohemorange: Is there a reason these deps were older? They're already not equivalent to what we're pinning in certbot-auto (1.5.3, 16.2.0), which is what I would have expected them to be.

ohemorange:
If the entire point of this is for developers to not have everything be broken, why not also pin the dev dependencies? Because of the different times when people will start running into issues?

tox.ini
erikrose: (IHNI, so I'm interested in the answer, too.)

erikrose:
The dev dependencies I saw were mostly tooling: ipdb, coverage, tox, twine, wheel. That stuff is coupled to the Python runtime itself and not to our codebase—we don't even import it. So I judged it would be better for people to have newer versions that fix bugs than to stay in stasis. Now that I look at it again, I could be persuaded to pin the things that take config, like tox<3.0, in case they change config file format at some point. I could go either way.

zjs:
> Now that I look at it again, I could be persuaded to pin the things that take config, like tox<3.0, in case they change config file format at some point. I could go either way.

It doesn't seem like `tox` applies semver; there weren't compatibility-breaking changes in 2.0, they just seem to increment the minor version until they get to 9 and then bump the major version to avoid a ".10" release.

erikrose:
Many, many Python packages don't use semver. I was conservative. My thinking was that, if they ever *do* break backward compat, they'll at least bump the major version then. As we said in #4433, we release certbot often enough that we can bump the <X.0 versions in our pins if a major version of something is released that doesn't break us.

bmw:
Closing after conversation in #4433.

PR #4503, opened 2017-04-14
Set HOME if it's not defined in certbot-auto

### LINKED ISSUE #4489: certbot-auto stores configuration in /~/ directory when launched as systemd service

Debian 8 amd64

```
# wget -q -O certbot-auto https://dl.eff.org/certbot-auto
# chmod a+x certbot-auto 
# mv certbot-auto /usr/local/bin/certbot
```

I create initial certificates like this:
```
certbot certonly --non-interactive --agree-tos --no-self-upgrade --email info@domain.com --apache -d my.domain.com
```
Certbot creates `/root/.local/share/letsencryt`file tree:
```
# tree -a /root/.local/ | head
/root/.local/
└── share
    └── letsencrypt
        ├── bin
        │   ├── activate
        │   ├── activate.csh
        │   ├── activate.fish
        │   ├── activate_this.py
        │   ├── certbot
        │   ├── easy_install
```

BUT, 

After declaring systemd unit file `/etc/systemd/system/certbot.service` (to be usde by `certbot.timer`) like this:
```
[Unit]
Description=Certbot

[Service]
Type=oneshot
ExecStart=/usr/local/bin/certbot --no-self-upgrade -q renew
PrivateTmp=true
```
And execute `systemctl start certbot.service` (or by setting up `certbot.timer`),  Certbot creates different configuration file tree in system root, in `~` directory:
```
# tree -a /~/ | head                                                                                                                                     
/~/
└── .local
    └── share
        └── letsencrypt
            ├── bin
            │   ├── activate
            │   ├── activate.csh
            │   ├── activate.fish
            │   ├── activate_this.py
            │   ├── certbot

```

Certbot duplicates `.local/share/letsencrypt` in different path when executed as systemd service, compared to execution interactively in terminal. Seems like "~/" is not resolved when running as systemd service.

## Here is a Certbot log showing the issue (if available):
There is no relevant log entries.

bmw:
Fixes #4489

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -18,7 +18,7 @@ set -e  # Work even if somebody does "sh thisscript.sh".
 
 # HOME might not be defined when being run through something like systemd
 if [ -z "$HOME" ]; then
-  HOME=/root
+  HOME=~root
 fi
 if [ -z "$XDG_DATA_HOME" ]; then
   XDG_DATA_HOME=~/.local/share
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -15,6 +15,11 @@ set -e  # Work even if somebody does "sh thisscript.sh".
 
 # Note: you can set XDG_DATA_HOME or VENV_PATH before running this script,
 # if you want to change where the virtual environment will be installed
+
+# HOME might not be defined when being run through something like systemd
+if [ -z "$HOME" ]; then
+  HOME=/root
+fi
 if [ -z "$XDG_DATA_HOME" ]; then
   XDG_DATA_HOME=~/.local/share
 fi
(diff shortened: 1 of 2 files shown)


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -15,6 +15,11 @@ set -e  # Work even if somebody does "sh thisscript.sh".
 
 # Note: you can set XDG_DATA_HOME or VENV_PATH before running this script,
 # if you want to change where the virtual environment will be installed
+
+# HOME might not be defined when being run through something like systemd
+if [ -z "$HOME" ]; then
+  HOME=/root
```
zjs: Instead of assuming that the root user's home directory is `/root` (which it may not be), I think it would be safer to say `HOME=~root`.

bmw: Good point. Do you know if `~root` is POSIX compliant? `certbot-auto` gets run in all kinds of shells; bash, dash, busybox, csh, tcsh, etc.

zjs: It looks like it: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_01

zjs: It looks like it: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_01

bmw: Awesome. Thanks for the link. I've made the change and tested this on a variety of systems with no issues.

Issue #4153, opened 2017-02-02
Document all challenges and their relationship with Certbot


SwartzCr:
Right now we suggest users use manual auth hooks to allow them to do the DNS challenge, and we even go so far as to suggest example scripts for them to run, but at no point do we explain what that challenge is doing!

There is already some things written about this in the manual plugin, which gets written out to the user when they call `certbot certonly --manual` without an auth hook flag. This command can give you information about any of our challenges, and you can specify which one by specifying that challenge with the --preferred-challenge flag.

Note that like the webroot challenge, certbot will tell the user to put certain files in certain directories to be served as part of the challenge. These files and directories are generated randomly for each run and shouldn't be put verbatim into the documentation. You can also look at these suggestions directly in the code: https://github.com/certbot/certbot/blob/master/certbot/plugins/manual.py

schoen:
This issue should actually be about a new documentation section that talks about *all* of the challenges, what they require, and how Certbot can use each of them.

schoen:
I'm suggesting calling the documentation section "How Certbot proves your ownership of domains".

schoen:
@pconrad-fb, I might try to work on this but I'll let you know if I do or what progress I make.

SwartzCr:
fyi that @schoen has a draft of this that I'm looking over - hopefully we'll have something PRable by next week

pconrad-fb:
I'm looking at what.rst today, but as long as I'm in the neighborhood, is there a draft of the challenges content that you'd like me to look at @schoen @SwartzCr ?

SwartzCr:
@schoen mentioned he'd post it, but I'm not sure that he has yet

schoen:
Sorry this took me a while to put into a PR. Now available here for your editing and commenting as #4509.

pconrad-fb:
Okay, I took a look at this. It's good, and has plenty of information. I do think there is a lot I can do to compress and restructure it without losing information. If you merge it into master, I'm happy to pull it down and submit a PR. If there's another approach you'd like me to take, let me know.

SwartzCr:
@pconrad-fb I've merged it, so feel free to edit
I'll probably make edits on top of your edit? But as long as we don't add it to the index file it won't get published so we can edit it in master

pconrad-fb:
I've pushed up the initial edit. Might not be perfectly clean yet (may be a typo or two in there somewhere) as I was mainly focusing on restructuring the content.

Remaining questions:
[...]

PR #4509, opened 2017-04-14
Attempt to document challenge types

schoen:
This still needs some editing but I'm putting it into a PR so that other people can see and work on it.

### DIFF
--- a/docs/challenges.rst
+++ b/docs/challenges.rst
@@ -1,0 +1,156 @@
+Digital certificates can only be issued to people who are entitled to them. For example, assuming you don't run google.com, you're not entitled to a certificate for it. Nor is someone else entitled to receive a certificate for your web site.
+
+In order to receive a certificate from Let's Encrypt certificate authority (CA), you have to prove your control over each of the domain names that will be listed in the certificate. You can do so by making certain publicly-visible changes, proving that the person who's requested a particular certificate is the same person who controls the site(s) that the certificate will refer to.
+
+Let’s Encrypt specifies three different ways to prove your control over a domain (each of which Certbot may be able to do for you). These are called "challenges," because you are being challenged to perform tasks that only someone who controls the domain should be able to accomplish.
+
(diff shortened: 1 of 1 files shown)


schoen:
Fixes #4153.

SwartzCr:
@pconrad-fb feel free to add comments

docs/challenges.rst
```diff
@@ -0,0 +1,156 @@
...
+  program on your server listening to port 443
+
+* Your DNS records aren't valid
+  Try checking your DNS records with a tool like the DNSchecker at
+  http://www.dnsstuff.com/ to make sure there are no serious errors.
+  Sometimes a DNS error still allows your site to load in a web
+  browser, but prevents the certificate authority from issuing a
+  certificate.
+
+DNS-01 challenge:
```
zjs: Another possible issue with DNS-01 would be not waiting long enough for the DNS change to propagate.

schoen:
I don't think that is a possible reason because Certbot checks the authoritative DNS servers directly, rather than relying on any cache.  Have you found an instance where someone has claimed this was a reason for a DNS-01 validation failure?

zjs:
> Certbot checks the authoritative DNS servers directly

Perhaps I'm looking for the wrong strings, but I can't find any code in Certbot that queries DNS servers.

However, even if Certbot did query an authoritative DNS server directly, Certbot is not guaranteed to get the same view of the world as the ACME server; many providers (e.g., Google Cloud DNS) use Anycast to route requests to the nearest of a set of global-distributed eventually-consistent authoritative nodes.

Eventually, ACME may support some notion of retry (see, e.g., https://github.com/ietf-wg-acme/acme/pull/292, https://github.com/ietf-wg-acme/acme/pull/297) at which point it may be able to hide this complexity from users.

> Have you found an instance where someone has claimed this was a reason for a DNS-01 validation failure?

I'm not aware of user reports, but this is something I've experienced.

SwartzCr:
merging so that @pconrad-fb can provide edits

PR #4530, opened 2017-04-19
Fix minor certbot-auto verbosity issue

bmw:
[My comment](https://github.com/certbot/certbot/pull/4292#discussion_r112085601) on #4292 caused this. These lines were previously conditional on `if [ "$QUIET" = 1 ]` and my comment caused this behavior to be removed. This PR just fixes that rather than having the PR author have to change things back due to my mistake.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -1069,8 +1069,8 @@ UNLIKELY_EOF
   fi
   if [ -n "$SUDO" ]; then
     # SUDO is su wrapper or sudo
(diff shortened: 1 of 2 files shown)


ohemorange:
So the comments about promptless sudo when run interactively no longer apply?

bmw:
They do but it's not a behavior change. We were already not providing a prompt for `sudo` here.

I think this sucks and want to do better, but: I'm concerned about printing messages when run from something like `cron` if we leave this as is. There was some discussion of other things we could do in [this thread](https://github.com/certbot/certbot/pull/4292#discussion_r112104626), but I don't want to take the time to engineer one of those solutions right now.

Issue #4507, opened 2017-04-14
Finalize Python 3 support


bmw:
We should be Python 3 compatible now! Before the next release though, we should:

- [ ] Spend some time playing with it to see if we can find a corner case where it breaks. If so, it should be added to our integration tests.
- [x] Make `tools/venv3.sh` install Certbot and it's official plugins (fixed in #4532)
- [ ] Modify the relevant `setup.py` files listing support for Python 3.
- [ ] Update system requirements in `docs/install.rst` to mention support for Python 3 versions.

ahaw021:
Hi @bmw 

Awesome

how do we help out with testing?

Andrei

yan12125:
@ahaw021 A good way is packaging certbot for Python 3 and using it. Apparently you're already doing that :) https://github.com/ahaw021/CERTBOT-WINDOWS-BUILD

Here I'd like to advertise my packages, too. For Arch Linux users, python-certbot-git, python-certbot-apache-git and python-certbot-nginx-git on AUR can be used for testing. Note that I patch setup.py in python-certbot-apache-git so that it accepts python-augeas 1.x.

cclauss:
TODO: Add `'Programming Language :: Python :: 3'` to setup.py.

yan12125:
I have a question: I guess https://certbot.eff.org/docs/install.html#system-requirements should be updated, too? Currently it reads:

> Certbot currently requires Python 2.6 or 2.7

bmw:
Nice catch! I added updating `docs/install.rst` to my original post.

PR #4532, opened 2017-04-20
Extend tools/venv3.sh to support full certbot development

yan12125:
Ref: #4507

Ran unit tests in the virtual environment and it's all fine.


yan12125:
Wow! That's fast!

bmw:
Haha hopefully that becomes the norm for this repo. No one likes it when their PR sits around with no response.

Issue #4079, opened 2017-01-18
Make nginx a first class plugin!


pde:

ohemorange:
It may be ready, but it will take work and testing that we shouldn't be doing in this release.

ohemorange:
So, oddly enough, we will probably have to make nginx a second-class plugin to make it a first-class plugin.

That is, certbot-auto currently auto-installs nginx and apache. We need it to not do that, because then we change the workflow of `run` from automatically choosing apache to asking the user to interactively select between apache and nginx.

So, doing this will require:

- delete `hidden = True`
- update website instructions to recommend plugin
- change from Alpha to Beta
- have certbot-auto not install nginx

ohemorange:
\* have certbot-auto not install nginx, unless the `--nginx` flag is passed.

zjs:
> then we change the workflow of run from automatically choosing apache to asking the user to interactively select between apache and nginx

Why is adding a prompt unacceptable? (Is there a reason this backwards compatibility must be maintained?)

It seems like the critical use case for preserving backwards compatibility would be renewal, but in the case of renewal, wouldn't the renewal conf file have a reference to the apache plugin, allowing the prompt to be avoided?

pde:
@ohemorange so if I'm following you correctly, the case we're most worried about is users who:

1. Have both apache and nginx installed on a server
2. Are running some kind of renewal script that is currently non-interactive, but doesn't use `renew` or `-n`
3. Are using the Apache plugin successfully, but without providing `--apache`, and instead are having that plugin selected for them as a default?

ohemorange:
@zjs: both `certonly` and `run` work perfectly well for renewal, and people use those in their crontabs. if you do it that way, it doesn't read from the renewal conf file and instead uses only command line arguments.

ohemorange:
@pde 

1. Do you mean the servers, or the plugins? Because the servers don't matter, and the question is if we should be installing the nginx plugin in one case, and in another case yes, they've already installed both plugins.
2. Yup! This should probably only be `run` / no subcommand.
3. Yup!

zjs:
> @zjs: both certonly and run work perfectly well for renewal, and people use those in their crontabs. if you do it that way, it doesn't read from the renewal conf file and instead uses only command line arguments.

I see. That... complicates things.

ohemorange:
[...]

PR #4543, opened 2017-04-24
Be careful with new interaction from enabling nginx

pde:
This is an option for resolving the interactivity blockers for #4079.


bmw:
Closing in favor of #4969.

PR #4552, opened 2017-04-26
Add curl package

thiagotalma:
Without this package it is not possible to use the hooks that depend on http requests. Especially the example with Cloudflare.


bmw:
Closing this for now. Feel free to comment on this PR or #4556 if you think I should reconsider.

Issue #4524, opened 2017-04-19
Use include directive for Nginx added SSL directives


ohemorange:
To prevent future versions of #4284.

ohemorange:
Note: remember to test downgrading Certbot versions

ohemorange:
The include file should be put in as soon as you run, but ideally put in by packagers -- look at apache

bmw:
Fixed in #4557.

Issue #583, opened 2015-07-02
Recognize inherited ssl directives


HepplerDotNet:
WARNING:letsencrypt_nginx.configurator:Cannot find a cert or key directive in /etc/nginx/vhosts.d/gowide.de.conf for set(['www.gowide.de', 'aws.gowide.de', 'gowide.de']). VirtualHost was not modified.

My gowide.de.conf: https://pastebin.mozilla.org/8838346
Lines 5&6 point to a selfsigned cert.

Command i run:

```
letsencrypt -d www.gowide.de -d gowide.de -d aws.gowide.de -e -t run
```

jsha:
@diracdeltas can you look at this one?

diracdeltas:
Will look into it. The plugin doesn't expect the NGINX config to already have an SSL configuration defined, so this sort of error is expected (unfortunately).

PatrickHeppler:
It works if I have only one server name, even with already enabled SSL.
https://pastebin.mozilla.org/8839291
This config works if I run 
```
letsencrypt -d www.gowide.de -d gowide.de -d aws.gowide.de -e -t run
```
I don't get any errors.  ssl_certificate and ssl_certificate_key get replaced with correct path to the letsencrypt certs in all 3 server blocks.

bmw:
I'm unable to repo this issue. Unfortunately, the pastebins above have expired, but taking a simple configuration file and adding multiple server names, Certbot correctly modifies the file.

If you can still reproduce this issue, can you either create a new pastebin or paste the configuration file's contents into this issue?

pde:
Closing this as unreproducible for now :(

pde:
But pinging @PatrickHeppler in case he can still reproduce :/

johnaoss:
I believe I may have reproduced this. Doing a certonly with --dry-run works perfectly fine, yet it fails on installation, stating that it was unable to install the certificate. 

Command:
`./certbot-auto run --nginx -d johnoss.com -d www.johnoss.com -d video.johnoss.com`

Output:
[Link to Gist](https://gist.github.com/johnaoss/e36c706323c701bad4ed4e2313a5f22f)

I'm unsure as to why it fails during expansion, and only for the domain I currently have a cert for (['johnoss.com']).

ohemorange:
Could you repost your Nginx config file?

johnaoss:
@ohemorange Updated previous comment with other files.

ohemorange:
[...]

PR #4557, opened 2017-04-27
Switch to using include directive for Nginx constants

ohemorange:
Addresses #4524.

### DIFF
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -312,7 +312,7 @@ class NginxParser(object):
         except errors.MisconfigurationError as err:
             raise errors.MisconfigurationError("Problem in %s: %s" % (filename, str(err)))
 
-def _parse_flat_file(ssl_options):
+def _parse_ssl_options(ssl_options):
     if ssl_options is not None:
         try:
             with open(ssl_options) as _file:
@@ -515,8 +515,10 @@ def _add_directive(block, directive, replace):
 
     # Find the index of a config line where the name of the directive matches
     # the name of the directive we want to add. If no line exists, use None.
-    find_location = lambda direc: next((index for index, line in enumerate(block) \
-                     if line and line[0] == direc[0]), None)
+    def find_location(direc):
+        return next((index for index, line in enumerate(block) \
+            if line and line[0] == direc[0]), None)
+
     location = find_location(directive)
 
     if replace:
@@ -535,31 +537,30 @@ def _add_directive(block, directive, replace):
         # handle flat include files
 
         directive_name = directive[0]
-        can_append = lambda loc, dir_name: loc is None or (isinstance(dir_name, str) \
-                                and dir_name in REPEATABLE_DIRECTIVES)
+        def can_append(loc, dir_name):
+            return loc is None or (isinstance(dir_name, str) and dir_name in REPEATABLE_DIRECTIVES)
 
-        cannot_append_error = lambda direc, loc: errors.MisconfigurationError( \
-                'tried to insert directive "{0}" but found ' \
-                'conflicting "{1}".'.format(direc, block[loc]))
+        err_fmt = 'tried to insert directive "{0}" but found conflicting "{1}".'
 
         # Give a better error message about the specific directive than Nginx's "fail to restart"
         if directive_name == INCLUDE:
             # in theory, we might want to do this recursively, but in practice, that's really not
             # necessary because we know what file we're talking about (and if we don't recurse, we
             # just give a worse error message)
-            included_directives = _parse_flat_file(directive[1])
+            included_directives = _parse_ssl_options(directive[1])
 
             for included_directive in included_directives:
                 included_dir_loc = find_location(included_directive)
                 included_dir_name = included_directive[0]
                 if not can_append(included_dir_loc, included_dir_name):
-                    raise cannot_append_error(included_directive, included_dir_loc)
+                    raise errors.MisconfigurationError(err_fmt.format(included_directive,
+                        block[included_dir_loc]))
 
         if can_append(location, directive_name):
             block.append(directive)
             _comment_directive(block, len(block) - 1)
         elif block[location] != directive:
-            raise cannot_append_error(directive, location)
+            raise errors.MisconfigurationError(err_fmt.format(directive, block[location]))
 
 def _apply_global_addr_ssl(addr_to_ssl, parsed_server):
     """Apply global sslishness information to the parsed server block
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -534,7 +534,6 @@ def _add_directive(block, directive, replace):
         # and there is already a copy of that directive with a different value
         # in the config file.
 
-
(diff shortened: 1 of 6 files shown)


ohemorange:
This undoes parts of #2262 and #3383 (but leaves the parts about fixing the `ssl_session_cache`, for example).

We previously thought that it would be easier to upgrade ciphersuites if we kept them inline, but it turns out that's not the case.

Yes, this runs the risk of accidentally duplicating specific directives in a single server block. But, that's arguably safer now than it was when we made the original change. First, we only ever do this to non-ssl server blocks, and we more robustly detect which are and aren't server blocks now. So the only case where it's a problem is when there are already ssl directives in a non-ssl server block, at which point something else has probably also already gone even more terribly wrong.

Some people who used `/etc/letsencrypt/options-ssl-nginx.conf` in the past might have manually modified it, because it lives in `etc`. We should make sure that old versions are kept around when we add the updating code (which isn't in here just yet, and should probably just go in a separate PR). As a bonus for adding that, people who have the super-old version (from before we changed to `ssl_session_cache shared:le_nginx_SSL:1m;`) will get the update to fix that for free!

ohemorange:
There's a slight benefit to doing it this way.

Previously, we tried to insert the directive, and if we found it at all, we told them and didn't bother trying to insert it and rolled everything back.

Now, we just stick it in (effectively) and try to start nginx. Because sometimes it's just a warning, not a failure.

So it's optimistically try to stick it in and start nginx anyway vs. go parse the file, test its contents for duplicates in the server block, and only if it all checks out stick it in.

Which gives us a chance of it succeeding vs. a better error message.

I think the error message is worth it, I'll change it here.

ohemorange:
@bmw took some refactoring, but error message is back.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -344,6 +312,16 @@ def add_server_directives(self, vhost, directives, replace):
         except errors.MisconfigurationError as err:
             raise errors.MisconfigurationError("Problem in %s: %s" % (filename, str(err)))
 
+def _parse_flat_file(ssl_options):
```
bmw: nit: If you want to make this function be specific to our TLS options file, I'd update the function name to reflect that. I see a benefit in making this more general as we continue to use more include directives though so perhaps you should update variable names and error messages to make it more useful for parsing included files in the future.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -536,8 +515,10 @@ def _add_directive(block, directive, replace):
 
     # Find the index of a config line where the name of the directive matches
     # the name of the directive we want to add. If no line exists, use None.
-    location = next((index for index, line in enumerate(block)
-                     if line and line[0] == directive[0]), None)
+    find_location = lambda direc: next((index for index, line in enumerate(block) \
```
bmw: nit: perhaps you should make a small helper function for this as this is a bit complicated to read.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -549,15 +530,36 @@ def _add_directive(block, directive, replace):
...
         # and there is already a copy of that directive with a different value
         # in the config file.
+
+
+        # handle flat include files
+
         directive_name = directive[0]
-        if location is None or (isinstance(directive_name, str) and
-                                directive_name in REPEATABLE_DIRECTIVES):
+        can_append = lambda loc, dir_name: loc is None or (isinstance(dir_name, str) \
```
bmw: nit: consider splitting into small helper function rather than lambda.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -549,15 +530,36 @@ def _add_directive(block, directive, replace):
...
+                                and dir_name in REPEATABLE_DIRECTIVES)
+
+        cannot_append_error = lambda direc, loc: errors.MisconfigurationError( \
```
bmw: nit: This is a somewhat strange use of a lambda in my opinion. Maybe instead you should make an `ERR_FMT` string that you can use or a simple function called something like `_fail_add_directive()`.

ohemorange:
Oops meant to preserve behavior, I'll add that condition back in.

certbot-nginx/certbot_nginx/parser.py
ohemorange: I may have gone a little overboard on the lambdas.

certbot-nginx/certbot_nginx/parser.py
ohemorange: given that was already have a complicated general mechanism for parsing files, I'm going to return this to its original name and leave it as a special-use function.

If #583 ever gets resolved, that would be a good time to fix our general machinery for parsing files and included files and figuring out which directives apply where.

ohemorange:
Actually, there will be a behavior change here.

Before, if there was an exact duplicate, we wouldn't error out, but nor would we append it to have two copies of the same directive in the same block.

Now, it's probably most correct to include the entire file containing that directive and (as we currently do) just fall back to Nginx's own error checking to figure out whether that was ok or not on restart.

bmw:
I want to do some more testing here but after playing with this for a bit, it appears whether or not Nginx errors out if there a duplicate directive depends on the directive. This is fine:
```
ssl_session_cache shared:le_nginx_SSL:1m;
ssl_session_cache shared:le_nginx_SSL:1m;
```
but this is not:
```
ssl_prefer_server_ciphers on;
ssl_prefer_server_ciphers on;
```
This is an unfortunate side effect of using `include` because previously we just wouldn't include the directive if an identical one was already present. This means that switching to this approach will cause Certbot to fail for configurations where it previously worked.

Unfortunately, I have to sign off for the evening, but I'll play with and think about this more tomorrow.

ohemorange:
Yeah, that's right. We could comment out the original directive with a note saying "duplicated in <include file location>."On May 1, 2017 6:37 PM, Brad Warren <notifications@github.com> wrote:I want to do some more testing here but after playing with this for a bit, it appears whether or not Nginx errors out if there a duplicate directive depends on the directive. This is fine:
ssl_session_cache shared:le_nginx_SSL:1m;
ssl_session_cache shared:le_nginx_SSL:1m;

but this is not:
ssl_prefer_server_ciphers on;
ssl_prefer_server_ciphers on;

This is an unfortunate side effect of using include because previously we just wouldn't include the directive if an identical one was already present. This means that switching to this approach will cause Certbot to fail for configurations where it previously worked.
Unfortunately, I have to sign off for the evening, but I'll play with and think about this more tomorrow.

bmw:
What would you like to see with this PR? Try to get functionality like that added in time for 0.14.0? Skip this for 0.14.0? Ship this knowing it will no longer work on some existing configurations?

ohemorange:
That's like one extra line, I can do it in about 5 minutes. If you'd rather this wait, I'm always fine with that.

bmw:
If it's that simple, go for it! It won't work if the directive is duplicated outside of the server block, but we have that same problem currently so it certainly doesn't have to be fixed in this PR.

ohemorange:
Right yeah, if we want this whole shebang to be correct that would be more and different work.

PR #4566, opened 2017-04-28
Fixes the suggested python code to work with IPv4 & IPv6 installations

sgyurko:


sgyurko:
Perhaps a bit of background. The suggested code (original) binds only on the IPv4 address. I have a dual-stack web server and verification takes ages as the verifier tries the IPv6 address first which needs to time out first. The PR makes it that the script binds to both IPv4 and IPv6. If IPv6 is unavailable it still works just fine.

sgyurko:
Sure. I'll make it robust enough to cover those bits.

sgyurko:
How about the committed bit? That will work in all of the mentioned scenarios, with silent fallbacks. Tested on OSX (BSD like) and Linux.

sgyurko:
Hang on, one thing is still outstanding from your list. Python not having IPv6 can be a problem. That perhaps could be done if two scripts created and attempted to run. One might or might not fail and the other will do just fine.

That is assuming that the execution is aggregated with a small bash script. How likely is a python without IPv6? The OS without it is more likely, isn't it?

sgyurko:
Apparently socket.has_ipv6 is there to say if python has IPv6 built in (not the OS). That should cover all corners then.

bmw:
@ohemorange, since you're looking into how to implement/test automatic IPv4 and IPv6 support in Python for #1466, it'd be great if you could take a look at this so I don't have to duplicate your work. If you still have a preference that I review it though, let me know.

bmw:
@sgyurko, I'm sorry, but one of our other contributors suggested that we remove the lines you were changing here from our output entirely. You can see the issue for that at #4804. I tend to agree with him that it's usually visual noise and probably does more harm than good so I went ahead and merged his PR making the change.

With this in mind, this PR is now obsolete. I'm sorry it took us so long to take a look at it and that we ended up not taking it. We do support IPv6 in standalone though (see #1466) and if you're still interested in contributing to Certbot but aren't sure what to work on, please let me know.

sgyurko:
All right, no problem. I submitted the fix because when I needed it there were no other way to do IPv6 manual validation. If this is no longer the case, I'm fine.

PR #4567, opened 2017-04-28
Cloudflare DNS Authenticator

### LINKED ISSUE #4569: Implement a DNS authenticator for Cloudflare
Implement an Authenticator plugin which can fulfill a dns-01 challenge using the Cloudflare API.

See #4360 and #4567.

zjs:
Implement an Authenticator which can fulfill a dns-01 challenge using the
Cloudflare API. Applicable only for domains using Cloudflare for DNS.

Testing Done:
 * `tox -e py27`
 * `tox -e lint`
 * Manual testing:
    * Used `certbot certonly --dns-cloudflare -d`, specifying a
      credentials file as a command line argument. Verified that a
      certificate was successfully obtained without user interaction.
    * Used `certbot certonly --dns-cloudflare -d`, without specifying a
      credentials file as a command line argument. Verified that the user
      was prompted and that a certificate was successfully obtained.
      a credentials file after selecting cloudflare interactively and
      that a certificate was successfully obtained.
 * Negative testing:
    * Path to non-existent credentials file.
    * Credentials file with unsafe permissions (644).
    * Credentials file missing e-mail address.
    * Credentials file with blank API key.
    * Credentials file with incorrect e-mail address.
    * Credentials file with malformed API key.
    * Credentials file with invalid API key.
    * Domain name not registered to Cloudflare account.
 * `./tox.cover.sh`
    * certbot/plugins/dns_common.py                  100%
    * certbot/plugins/dns_common_test.py             100%
    * certbot/plugins/dns_test_common.py             100%
    * certbot_dns_cloudflare/dns_cloudflare.py        97%   61, 157, 159
    * certbot_dns_cloudflare/dns_cloudflare_test.py  100%

(Replaces #4360, resolves #4569)


bmw:
Just to make sure you saw it, it looks like tests are failing on Python 3.

bmw:
@joohoi, if you have time to take a look at this over the weekend, that'd be useful, but you should prioritize your own PRs. If you don't manage to get to it, it's not a problem and I'll review it on Monday.

certbot-dns-cloudflare/setup.py
```diff
@@ -0,0 +1,65 @@
...
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Environment :: Plugins',
+        'Intended Audience :: System Administrators',
+        'License :: OSI Approved :: Apache Software License',
+        'Operating System :: POSIX :: Linux',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
+        'Programming Language :: Python :: 2.7',
```
yan12125: How about adding Python 3 classifiers like 8fa12bef8e7c687400367cf45a022b43e8c92754?

zjs: Should I also be introducing a `setup.cfg` file with the following?

```
[bdist_wheel]
universal = 1 
```

bmw: Yes. That'd be great. It tells the Python packaging tools that it is safe to build a single wheel and mark it compatible with both Python 2 and Python 3. You can do this as long as your package doesn't contain any C extensions which must be compiled (and this plugin does not).

zjs:
*Testing*
> There are a few lines that don't have unit test coverage. In the cases where it's easy to add coverage, I think it should be done. Places where it'd be complicated/convoluted to get test coverage can be ignored if you'd like.

It seems like there are two uncovered lines. I don't see a way to write a meaningful test for either.

*Expected file format*
> I think we should do something to help users figure out the format we expect for their DNS credentials.

Agreed! I think it would be much better to include some instructions in the help output. However, I couldn't see an easy way to do that given that `add_parser_arguments` is a `@classmethod` and the plugin name, which is included in the property names (for good reasons, I think) and would therefore need to be included in the documentation, is an instance variable.

> In the mean time, I think it'd be good to add more information to the prompt and/or error message to help users know how to write this file.

I think the error messages are reasonably helpful, but I'm open to suggestions to improve those.

I'll look at making the prompt a bit more helpful without making it too verbose.

*Release prep*
> Making each of these plugins shippable comes with a bit of work on `certbot-auto` and our release script. It'd be beneficial to me if you could help with this, but I was planning on doing some work that would conflict with this. How about I quickly land the changes that would conflict and then show you how to do this so you can do it in the future?

I'm happy to do any of this sort of work -- just let me know when the conflicting changes are in.

certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+    def _perform(self, domain, validation_name, validation):
+        self._get_cloudflare_client().add_txt_record(domain, validation_name, validation, self.ttl)
+
+    def _cleanup(self, domain, validation_name, validation):
+        self._get_cloudflare_client().del_txt_record(domain, validation_name, validation)
+
+    def _get_cloudflare_client(self):
+        return _CloudflareClient(self.credentials.conf('email'), self.credentials.conf('api-key'))
```
zjs: There's little overhead to creating a new client and using a new one for each operation means we don't need to worry about the client (or the underlying library) storing state. If we _always_ create a new client, we know there's never state, making the code easier to reason about.

certbot/plugins/dns_common.py
```diff
@@ -0,0 +1,305 @@
...
+@zope.interface.provider(interfaces.IPluginFactory)
+class DNSAuthenticator(common.Plugin):
+    """Base class for DNS  Authenticators"""
+
+    _attempt_cleanup = False
+
+    @classmethod
+    def add_parser_arguments(cls, add):
+        add('propagation-seconds',
+            default=10,
```
zjs: For the first plugin where 10 seconds isn't enough, I introduce a pattern for plugins to define their own default ([diff](https://github.com/certbot/certbot/pull/4585/files#diff-e2ab6366a1edcd587391b637b7b58a10R34)). Essentially, 10 seconds because the "default default".

bmw:
> It seems like there are two uncovered lines. I don't see a way to write a meaningful test for either.

Coveralls counted 4. One in [selection.py](https://coveralls.io/builds/11337056/source?filename=certbot%2Fplugins%2Fselection.py#L241) and three in [dns_cloudflare.py](https://coveralls.io/builds/11337056/source?filename=certbot-dns-cloudflare%2Fcertbot_dns_cloudflare%2Fdns_cloudflare.py). If a test would be complicated and/or meaningless though, you can ignore this.

bmw:
Let me know when this is ready for another review. Looks like the remaining things are the [docs directory](https://github.com/certbot/certbot/pull/4567#discussion_r114452932), user prompt (which I can try and help with), and potentially additional testing.

zjs:
I think this is ready for another review.

The credentials file prompt still isn't perfect, but it now re-prompts when the expected values aren't properly set in the file instead of exiting. If we want a better experience than that, I think we should reconsider prompting the user from the credentials and creating the file for them.

certbot/plugins/dns_common.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def __init__(self, filename, mapper=lambda x: x):
+        """
+        :param str filename: A path to the configuration file.
+        :param callable mapper: A transformation to apply to configuration key names
+        :raises PluginError: If the file does not exist.
+        """
+        validate_file_permissions(filename)
+
+        self.confobj = configobj.ConfigObj(filename)
```
bmw: If the file isn't a valid ini file, Certbot crashes with a traceback from `configobj`. Perhaps `configobj` raising an error should be documented in the docstring here and a validator that creates this object can catch (and log) this exception and explain the problem to the user?

certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare.py
```diff
@@ -0,0 +1,199 @@
...
+        # TODO: Document format (without calling `self.dest`?!)
+
+    def more_info(self):  # pylint: disable=missing-docstring,no-self-use
+        return 'This plugin configures a DNS TXT record to respond to a dns-01 challenge using ' + \
+               'the Cloudflare API.'
+
+    def _setup_credentials(self):
+        self.credentials = self._configure_credentials(
+            'credentials',
+            'Cloudflare credentials file',
```
bmw: Perhaps we should say that this should be an ini file in the label here?

zjs:
> This is unfortunately a painfully manual process, but I'm very open to ideas to simplify/automate it.

7754240 introduces  script which invokes `sphinx-quickstart` and then uses `sed`/`echo` to make the changes from that long list of commits.

Future _changes_ will still be painful, but adding a docs directory for a new package will be easy.

I think this belongs in a separate commit from the Cloudflare work though, so let me know when you're happy with this PR, but before you merge it, and I'll rebase to squash everything else into a single commit.

zjs:
Addressed the last few comments and rebased/autosquashed.

I pushed to [zjs:topic/zjs/cloudflare-package-unsquashed](https://github.com/zjs/certbot/commits/topic/zjs/cloudflare-package-unsquashed) prior to squashing to allow for easier review.

PR #4568, opened 2017-04-28
Tell the world we're Python 3 compatible

### LINKED ISSUE #4507: Finalize Python 3 support

bmw:
Fixes #4507.

I did some manual testing of our Python 3 support and didn't find any problems with it. Let's let the world use it!


ohemorange:
Do our OS dependencies not count as compiled extensions? Or does `universal` just mean things that get installed with `pip`.

bmw:
No our OS dependencies don't count. The information [about C extensions](https://packaging.python.org/distributing/#universal-wheels) refers specifically to C extensions in our package itself, not in the entire dependency graph of our project. For what it's worth, we've been doing this in our `acme` module (which depends on `cryptography`, which depends on `libssl`) for a long time now (see b8690cd4).

PR #4581, opened 2017-04-29
Google Cloud DNS Authenticator

### LINKED ISSUE #4571: Implement a DNS authenticator for Google Cloud DNS
Implement an Authenticator plugin which can fulfill a dns-01 challenge using the Google Cloud DNS API.

See #4448.

zjs:
Implement an Authenticator which can fulfill a dns-01 challenge using
the Google Cloud DNS API. Applicable only for domains using Google Cloud
DNS for DNS.

Testing Done:
 * `tox -e py27`
 * `tox -e lint`
 * Manual testing:
    * Used `certbot certonly --dns-google -d`, specifying a credentials
      file as a command line argument. Verified that a certificate was
      successfully obtained without user interaction.
    * Used `certbot certonly --dns-google -d`, without specifying a
      credentials file as a command line argument. Verified that the
      user was prompted and that a certificate was successfully
      obtained.
      a credentials file after selecting google interactively and that
      a certificate was successfully obtained.
 * Negative testing:
    * Path to non-existent credentials file.
    * Credentials file with unsafe permissions (644).
    * Domain name not registered to Google Cloud Platform account.
 * `./tox.cover.sh`
    * certbot_dns_google/dns_google.py 99%   63
    * certbot_dns_google/dns_google_test.py 100%

(Replaces #4448, resolves #4571)


certbot-dns-google/setup.py
```diff
@@ -0,0 +1,70 @@
...
+from setuptools import find_packages
+
+
+version = '0.15.0.dev0'
+
+# Please update tox.ini when modifying dependency version requirements
+install_requires = [
+    'acme=={0}'.format(version),
+    'certbot=={0}'.format(version),
+    'fasteners',
```
bmw: I think this is here as a result of a bad merge as I don't see why this package depends on `fasteners`.

tox.cover.sh
```diff
@@ -25,6 +25,8 @@ cover () {
     min=98
   elif [ "$1" = "certbot_dns_digitalocean" ]; then
     min=98
+  elif [ "$1" = "certbot_dns_google" ]; then
+    min=98
```
bmw: I missed this for Digital Ocean but it looks like we could get away with `min=99` here to help ensure future changes to this plugin have test coverage.

certbot-dns-google/setup.py
```diff
@@ -0,0 +1,70 @@
...
+    'google-api-python-client',
+    'mock',
```
bmw: Since we import it directly, I think it's good practice to list `oauth2client` as a dependency.

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -0,0 +1,183 @@
...
+
+        changes = self.dns.changes()  # changes | pylint: disable=no-member
+
+        try:
+            request = changes.create(project=self.project_id, managedZone=zone_id, body=data)
+            response = request.execute()
+
+            status = response['status']
+            change = response['id']
+            while status == 'pending':
```
zjs: In the Google DNS API, record changes are not synchronous. Once `status` transitions from `pending` to `done`, it has begun propagating.

>  Why do we loop here but not in `del_txt_record` or `_find_managed_zone_id`?

In the case of `del_txt_record`: it doesn't matter if we report cleanup before the change has been made

In the case of `_find_managed_zone_id`: it's a read operation, so there's no status.

>  What is the benefit of calling `request.execute()` repeatedly?

We re-execute the request to get the updated `status` value from the server.

bmw: Makes sense! Thanks for the explanation.

PR #4583, opened 2017-04-30
Common code for Lexicon-based DNS authenticators

### LINKED ISSUE #4572: Define pattern for implementing Lexicon-based DNS Authenticators
Define abstract classes for implementing Lexicon-based DNS authenticator plugins and for implementing tests for such plugins.

zjs:
Note: This PR builds on #4581. (Once that has landed, I'll rebase this branch and remove this note.)

Introduce abstract classes to provide base functionality for Lexicon-based DNS Authenticator plugins and corresponding test cases.

(Resolves #4572)

### DIFF
--- a/certbot/plugins/dns_common_lexicon.py
+++ b/certbot/plugins/dns_common_lexicon.py
@@ -1,0 +1,94 @@
+"""Common code for DNS Authenticator Plugins built on Lexicon."""
+
+import logging
+
+from requests.exceptions import HTTPError, RequestException
+
+from certbot import errors
+from certbot.plugins import dns_common
+
+logger = logging.getLogger(__name__)
+
+
+class LexiconClient(object):
+    """
+    Encapsulates all communication with a DNS provider via Lexicon.
+    """
+
+    provider = None
+
+    def add_txt_record(self, domain, record_name, record_content):
+        """
+        Add a TXT record using the supplied information.
+
+        :param str domain: The domain to use to look up the managed zone.
+        :param str record_name: The record name (typically beginning with '_acme-challenge.').
+        :param str record_content: The record content (typically the challenge validation).
+        :raises errors.PluginError: if an error occurs communicating with the DNS Provider API
+        """
+        self._find_domain_id(domain)
+
+        try:
+            self.provider.create_record(type='TXT', name=record_name, content=record_content)
+        except RequestException as e:
+            logger.debug('Encountered error adding TXT record: %s', e, exc_info=True)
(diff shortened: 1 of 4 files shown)


certbot/plugins/dns_common_lexicon.py
```diff
@@ -0,0 +1,94 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class LexiconClient(object):
+    """
+    Encapsulates all communication with a DNS provider via Lexicon.
+    """
+
+    provider = None
```
bmw: While this is technically fine, I think this should be an instance variable rather than a class variable based on how it's used.

certbot/plugins/dns_common_lexicon.py
```diff
@@ -0,0 +1,94 @@
...
+            except Exception as e:  # pylint: disable=broad-except
+                result = self._handle_general_error(e, domain_name)
+
+                if result:
+                    raise result
+
+        raise errors.PluginError('Unable to determine zone identifier for {0} using zone names: {1}'
+                                 .format(domain, domain_name_guesses))
+
+    def _handle_http_error(self, e, unused_domain_name):
```
bmw: It looks like `unused_domain_name` isn't used in any of the plugins you've written so far. If you expect it to be used in the future, I think you should remove the `unused_` prefix so other classes can override the method without either changing the argument name or using the "unused" domain name parameter. The problem with changing the argument name is calls like `self._handle_http_error(e, unused_domain_name=domain_name)` fail.

If you don't expect it to be used, I think this argument should be removed.

I have the same feedback about the `_handle_general_error` method.

zjs: The parameter is used in the NS1 and DNS Made Easy implementations.

Having different argument names in subclasses does require callers to use positional arguments instead of keyword arguments when calling `_handle_http_error`, but I think that's okay; it's a private method.

bmw: Ah. I looked for it by searching for `unused_` in your PRs. It's obvious now why that didn't work.

My instinct is it's a bit of an unusual implementation detail to satisfy our linter. It's unlikely to be a problem, but as more plugins are built on this base class, some possibly by 3rd parties, I could see someone running into this issue.

I don't care that much and if you really don't think it's a good idea to make the change it's fine, but I don't like the idea of our linter forcing us to add (even relatively minor) complexities into our codebase that we wouldn't otherwise. The tool is supposed to work for us not the other way around.

Issue #3709, opened 2016-10-31
Add full support for using an existing private key


bdesham:
Certbot seems to generate a new private key each time it requests certificates. This makes it very difficult to use [SSL pinning](https://possiblemobile.com/2013/03/ssl-pinning-for-increased-app-security/) with Certbot: ordinarily you would distribute the server (public) key with your application and use the (private) key to obtain the server’s certificate. If the server’s certificate changes every three months then you would also have to update your application every three months (and risk users being locked out entirely if they happened not to upgrade).

The only way to bring your own server key to Certbot [seems to be](https://community.letsencrypt.org/t/how-to-use-existing-private-key-with-certbot/17050/2) to create your own CSR and pass this to Certbot using the `--csr` option. (In my understanding the key is “baked in” to the CSR in a way that makes it a suitable proxy for the private key.)

This has drawbacks, however. You have to get the CSR from somewhere but Certbot offers no relevant documentation. (I followed [this documentation](https://github.com/diafygi/letsencrypt-nosudo#how-to-use-the-signing-script) from letsencrypt-nosudo, another ACME client.) More significantly, when Certbot obtains the certificates from your CSR it doesn’t put them in `/etc/letsencrypt` but just in the current working directory. Part of the appeal of Certbot is that it usually puts certificates in a well-defined directory hierarchy; having symlinks to the latest versions in `/etc/letsencrypt/live` is also a boon for convenience. Neither of these is the case when you use your own CSR.

I am requesting that one of the following features be added to Certbot:

1. The ability to point Certbot to an existing server private key, presumably with a command-line flag. (This would be my preferred solution.)
2. Certificates that “came from” a user-provided CSR should be treated the same (i.e. placed in the same directories) as certificates that were obtained in the usual way.

daadu:
👍

lancedolan:
Big upvote on this!

Additionally, if your SSL termination happens on some hosted infrastructure or platform provider who doesn't use Let's Encrypt, but who will allow you to programmatically upload certs, then your certbot server will need to sign all certs with a private key provided to you by your hosting provider. That is our case now, and it's causing great difficulty, if not a total impasse. We might need to use some other janky ACME tool instead of certbot because of this.

lancedolan:
This documentation indicates that the feature already exists and that using the CSR is only optional... but I've found no documentation otherwise:

https://certbot.eff.org/faq/#can-i-use-an-existing-private-key-or-certificate-signing-request-csr-with-certbot

[...]

PR #4610, opened 2017-05-03
Allow reuse of private keys

### LINKED ISSUE #3788: please let me reuse a key during cert renewal

schoen:
This PR adds an option to allow the reuse of a specified existing private key when requesting a new certificate (other than via CSR, which is already an alternative which can achieve this function with or without direct access to the private key material).

Fixes #3788 and #3709.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -865,6 +865,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              "per Certbot run. To see certificate names, run 'certbot certificates'. "
              "When creating a new certificate, specifies the new certificate's name.")
     helpful.add(
+        [None, "run", "certonly"],
+        "--privkey-path", dest="privkey_path", default=None,
+        help="Location of an existing private key to use in this new certificate. "
+             "(If not specified, a new key pair will be created.)")
+    helpful.add(
         [None, "testing", "renew", "certonly"],
         "--dry-run", action="store_true", dest="dry_run",
         help="Perform a test run of the client, obtaining test (invalid) certs"
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -294,17 +294,18 @@ class Client(object):
 
         return certr, chain
 
-    def obtain_certificate(self, domains):
+    def obtain_certificate(self, domains, existing_key=None):
         """Obtains a certificate from the ACME server.
 
         `.register` must be called before `.obtain_certificate`
 
         :param list domains: domains to get a certificate
+        :param str existing_key: path to existing private key, or None to create one
 
         :returns: `.CertificateResource`, certificate chain (as
-            returned by `.fetch_chain`), and newly generated private key
-            (`.util.Key`) and DER-encoded Certificate Signing Request
-            (`.util.CSR`).
+            returned by `.fetch_chain`), and existing or newly generated
+            private key (`.util.Key`) and DER-encoded Certificate Signing
+            Request (`.util.CSR`).
         :rtype: tuple
 
         """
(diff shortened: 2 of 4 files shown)


schoen:
This does currently work to reuse a private key.

Four remaining problems or issues with this code (mostly all CLI design matters):

* Super-confusingly, there is already an option called `--key-path` which does something totally different! Hence I have called this one `--privkey-path`. Will this forever mire users in hopeless confusion between the two options? Any way out? Should we make `--privkey-path` have *either* meaning (even though they're quite different) depending on the context?

* Currently the CLI default `None` overrides the renewal config but it should not. Hence the option disappears on a subsequent renewal. I think I need to add a 

* There nonetheless needs to be a way to intentionally turn this off, which *should* override the renewal config and cause `reuse_key` to be removed from the renewal config. Should this be called `--no-privkey-path` or `--new-key` or `--new-privkey`? I assume we shouldn't try to get the parser to recognize the special string `None` as the Python value `None`, which would be ugly and confusing.

* It turns out that this functionality is still a bit confusing if you want to make a new key on first run but then reuse it afterward (for renewals). The reason is that the key-to-reuse doesn't exist yet on first run! So right now if you wanted to pin a key, you would either need to make it yourself with `openssl genrsa` and then keep your non-lineage copy around indefinitely, or you would need to make a new lineage and then immediately renew it with `--force-renew --privkey-path /etc/letsencrypt/live/example.com/privkey.pem` to give the renewal `privkey_path` something specific to refer to. Although some users who want to pin keys do have an existing copy of the key they want to pin, some people would prefer for Certbot to create the key on first run and then reuse it on renewal, and I'm not exactly sure what the UI for that common use case should look like.

bmw:
> Super-confusingly, there is already an option called --key-path which does something totally different! Hence I have called this one --privkey-path. Will this forever mire users in hopeless confusion between the two options? Any way out? Should we make --privkey-path have either meaning (even though they're quite different) depending on the context?

Do you think we could get away with just reusing `--key-path` here? Currently, the help for `--key-path` says:
```
  --key-path KEY_PATH   Path to private key for cert installation or
                        revocation (if account key is missing) (default: None)
```
Looking at the code, this seems accurate. This means the flag is only currently used for the `install` and `revoke` subcommands. The way you're using `--privkey-path` only affects the `certonly`, `run`, and `renew` subcommands. Since these different usages don't interact, I think we could get away with a single flag if we wanted. I personally like that approach better as it prevents people from accidentally using the wrong flag.

> Currently the CLI default None overrides the renewal config but it should not. Hence the option disappears on a subsequent renewal. I need to fix the lineage reconstitution to prevent this from happening.

Briefly playing with this, it didn't happen to me. Keep in mind that `renew` is the only subcommand that reuses values from a renewal configuration file. If you obtain a new cert using `certonly`, the renewal conf file is completely ignored and overwritten with the values used to obtain the cert you just got with `certonly`.

> There nonetheless needs to be a way to intentionally turn this off, which should override the renewal config and cause reuse_key to be removed from the renewal config. Should this be called --no-privkey-path or --new-key or --new-privkey? I assume we shouldn't try to get the parser to recognize the special string None as the Python value None, which would be ugly and confusing.

For better or worse, [we already treat None specially](https://github.com/certbot/certbot/blob/master/certbot/renewal.py#L251). The user can also work around this problem by just deleting the line from their renewal conf file.

Looking at the [values we currently save](https://github.com/certbot/certbot/blob/master/certbot/renewal.py#L30) in a renewal config, we have the problem with some of these as well. While this isn't the best UX, we can improve it in the future with a common approach to flags like this. My preference is to not add a `--no-*` flag and stick with the current approach for now.

> It turns out that this functionality is still a bit confusing if you want to make a new key on first run but then reuse it afterward (for renewals). The reason is that the key-to-reuse doesn't exist yet on first run! So right now if you wanted to pin a key, you would either need to make it yourself with openssl genrsa and then keep your non-lineage copy around indefinitely, or you would need to make a new lineage and then immediately renew it with --force-renew --privkey-path /etc/letsencrypt/live/example.com/privkey.pem to give the renewal privkey_path something specific to refer to. Although some users who want to pin keys do have an existing copy of the key they want to pin, some people would prefer for Certbot to create the key on first run and then reuse it on renewal, and I'm not exactly sure what the UI for that common use case should look like.

This is similar to the argument I initially made for `--reuse-key` [here](https://github.com/certbot/certbot/pull/3819#issuecomment-285234271). As I stated there, the user can also manually edit their renewal conf file to get this effect. While I agree it's not the best UX, I don't think we need to block this PR on that functionality. We can add it in the future as/if it's requested by users.

bmw:
@schoen, are you still planning on working on this?

schoen:
@bmw, yes, I am. Thanks for the reminder.

Nosxxx:
Any update?

voidzero:
@schoen could you please let us know how far along this is? I'd like to reuse my private key when renewing one of my certificates.

jschlyter:
@schoen, any updates on this issue?

schoen:
Planning to replace this with #5901, which does use a simpler approach different from what I talked about with @bmw last year (just `--reuse-key` rather than specifying `--key-path`). @ohemorange and I concluded that users will be much more confused by specifying `--key-path` from `/etc/letsencrypt/live` for this purpose if they don't already know what that is.

schoen:
Please merge #5901 instead of this PR. However, we should eventually also provide a way to use an existing non-Certbot-created key while also enrolling a cert for renewal (as even with the current version #5901, the only way to do that is with `--csr`, which doesn't enroll for renewal).

PR #4620, opened 2017-05-05
Allow boulder-fetch.sh run with `ip` from iproute2

yan12125:
On Arch Linux, `ifconfig`, which is part of the net-tools package, [2] is not within the base package group, so fresh Arch installations may come without it. This patch adds support for `ip` from the iproute2 package, [1] which is a member in the base group.

A sample output from `ip`:
```
$ ip addr show dev docker0 
3: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
    link/ether 02:42:22:95:fa:79 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 scope global docker0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:22ff:fe95:fa79/64 scope link 
       valid_lft forever preferred_lft forever
```

[1] https://www.archlinux.org/packages/core/x86_64/iproute2/
[2] https://www.archlinux.org/packages/core/x86_64/net-tools/


yan12125:
Thanks for reviewing!

PR #4623, opened 2017-05-05
update README

### LINKED ISSUE #4622: Update README.rst
This file hasn't really changed much in a long time and plenty of features have been added/removed. We should update it to reflect the current state of Certbot.

SwartzCr:
fixes #4622

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,17 +96,9 @@ ACME spec: http://ietf-wg-acme.github.io/acme/
 
 ACME working area in github: https://github.com/ietf-wg-acme/acme
 
-
-Mailing list: `client-dev`_ (to subscribe without a Google account, send an
-email to client-dev+subscribe@letsencrypt.org)
-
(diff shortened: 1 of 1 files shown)


SwartzCr:
Do you know if that's a public list that people should be joining? I don't want to suggest it as another place where people can send bug reports, since we want that to happen on IRC and GitHub, in what way to do you imagine that people will be using it?

bmw:
It is a public list but admins have to approve people joining/posting to the list. In my opinion, the purpose of the list should be Certbot development discussions (not bug reports), so with this in mind maybe we should list it in the developer guide rather than our README?

SwartzCr:
Yeah, I think people would object to people joining it before contributing in other ways to the project

bmw:
We've accepted people who were contributing for the first time by providing guidance on the list about things such as Debian updates, but I agree we don't want every Certbot user on the list.

Issue #4586, opened 2017-05-01
letshelp-certbot test failures on Arch Linux


bmw:
Running `tox -e py27` on 4ca702f6fbd5936f1b1e4ee794ebca09d5acd6fb (`master` at the time of this writing), `letshelp-certbot` unit tests fail with:
```
ERROR: test_copy_config (letshelp_certbot.apache_test.LetsHelpApacheTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/certbot/letshelp-certbot/letshelp_certbot/apache_test.py", line 79, in test_copy_config
    letshelp_le_apache.copy_config(server_root, tempdir)
  File "/root/certbot/letshelp-certbot/letshelp_certbot/apache.py", line 105, in copy_config
    elif safe_config_file(config_file_path):
  File "/root/certbot/letshelp-certbot/letshelp_certbot/apache.py", line 149, in safe_config_file
    universal_newlines=True)
  File "/usr/lib64/python2.7/subprocess.py", line 390, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/subprocess.py", line 1024, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

----------------------------------------------------------------------
```
We don't ship this package, but our tests should pass and this could cause problems for new developers. This is not a regression as tests also fail on v0.13.0 which was our last release.

yan12125:
Tests pass here. Did you run this in containers like docker? The ```file``` package is not critical for Arch Linux and I guess some docker image maintainers may remove it. If so ```pacman -S file``` should fix it.

bmw:
Yep! You figured it out. I ran this in `docker` and `file` wasn't available. Since this package is in the [base](https://www.archlinux.org/groups/x86_64/base/) group, this shouldn't be an issue for devs actually running Arch, but perhaps we should do something for other people who try to test in these minimal docker images. Probably the easiest approach is to check if the `file` command is available in `letshelp` and print a useful error message if it is not.

yan12125:
Good idea! letshelp is broken with Python 3. Let me fix it first :)

ArchangeGabriel:
Just to be clear, having the whole `base` group is not mandatory at all on Arch Linux. That being said, I still expect most people to have file installed. ;)

bmw:
[...]

Issue #4516, opened 2017-04-17
Build/verify the unicode sandwich


bmw:
Everything I've ever read about properly handling strings in Python says that you should use Unicode throughout your program and then convert them to/from the correct external encoding at the edges of your program. This means decoding input from a user, file, or network and converting it to Unicode as soon as possible and conversely encoding output right before it is sent. I've heard this described as the "Unicode sandwich". To do this, you should use `unicode/str` in Python 2 and 3 respectively throughout your program and convert to `str/bytes` respectively before output. `six.text_type` and `six.binary_type` help with this.

I don't think we currently do this correctly in Certbot. For example, see #4273. If someone is aware of a better way to handle text encoding, I'm open to other suggestions, but AFAIK this is the best way to do things. We should look through our code and make sure we a have a Unicode sandwich, correcting code as necessary.

@yan12125, you may be interested in this.

yan12125:
IMO it's better to have #2208 fixed first before going down this way. In my experience in Python 2.6 and earlier 2.7 versions, CPython's standard library doesn't support Unicode strings well. Here are some examples I saw in another Python project:

* https://bugs.python.org/issue2646: Unicode keyword arguments are not supported in Python < 2.6.5
* https://bugs.python.org/issue19099: struct.pack and struct.unpack don't accept Unicode as the format in Python < 2.7.7

If certbot aims to support as much Python versions as possible and switch to a "all-Unicode" model, I guess there will be lots of workarounds.

bmw:
Ugh good point. We shouldn't do this until we drop Python 2.6 support. Based on what I've read about this, despite the standard library having issues in some places, it's still worth doing to have a clear model and boundaries to prevent bugs like #4273. Again, I'm open to other ideas though.

Another thing we could do to help this is to work more quickly towards dropping Python 2 support all together. The Python community is heading this way and I'm not aware of any systems we support that don't have some version of Python 3 available (although it may be in an optional repository like EPEL on CentOS).. This will be a longer process though.

yan12125:
I've started some experiments by adding ```from __future__ import unicode_literals``` to all some files and checking unit tests. Apparently things are much better than I've thought. No tricky workarounds are necessary so far.

By switching all internal data to unicode, it's possible that third party plugins will get broken. Do you think certbot needs a transition plan? Or is it OK to force all plugins to use ```from __future__ import unicode_literals``` since some point?

bmw:
[...]

PR #4626, opened 2017-05-06
Use universal_newlines=True whereever the output is used

yan12125:
That fixes:

1. The letshelp command on Python 3. This is a preparation work for #4586
2. Most certbot commands on Python 3 on macOS

This topic should be revisited when #4516 is going to be implemented as ```universal_newlines=True``` implements a "bytes for Python 2 and unicode for Python 3" model.

Cross-ref: #3179


yan12125:
Thanks!

PR #4629, opened 2017-05-08
Pin dependency versions when using tools/venv.sh

### LINKED ISSUE #4433: Should we pin dependency versions?

bmw:
Fixes #4433.


letsencrypt-auto-source/build.py
```diff
@@ -21,14 +21,17 @@ def build(version=None, requirements=None):
...
-        contents of letsencrypt-auto-requirements.txt
+        contents of dependency-requirements.txt, letsencrypt-requirements.txt,
+        and certbot-requirements.txt
 
     """
     special_replacements = {
         'LE_AUTO_VERSION': version or certbot_version(DIR)
     }
     if requirements:
-        special_replacements['letsencrypt-auto-requirements.txt'] = requirements
```
ohemorange: What are these lines doing?

certbot-apache/setup.py
```diff
@@ -11,7 +11,7 @@
     'acme=={0}'.format(version),
     'certbot=={0}'.format(version),
     'mock',
-    'python-augeas<=0.5.0',
+    'python-augeas',
```
yan12125: python-augeas 1.x is currently broken and that affects certbot tests: https://github.com/hercules-team/python-augeas/issues/26. Can certbot keep 0.5.0 for now?

letsencrypt-auto-source/build.py
bmw: If the caller provides the optional `requirements` value it is included in `letsencrypt-auto` rather than our normal requirements. These lines cause the templates to be expanded using the contents of the dictionary rather than the normal file. I assign two values to an empty string so only the caller's requirements are used rather than mixing in ours.

This functionality is used in our [tests](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/tests/auto_test.py#L255).

certbot-apache/setup.py
bmw: This was discussed with a lot of detail in #4433. Basically our plan is not to specify the maximum version of our dependencies on PyPI as when new versions of our dependencies are released and packaged in OSes, Certbot package maintainers have to do extra work to remove these requirements.

Instead, the plan is to continue pinning specific versions of dependencies in `certbot-auto` and add this to our developer set up so temporary problems with new versions of upstream dependencies don't slow down Certbot contributors. We have other tests that test with the newest versions of all our dependencies so core developers will still be notified when changes from upstream break things.

letsencrypt-auto-source/build.py
ohemorange: Could any of these be the one set to `requirements`?

bmw: yep!

bmw:
This should be ready for a final review. I had to bump the version of `mock` we pin (which indirectly required we bump the version of `functools`) so tests pass.

(As a side note, the feature of dismissing stale reviews works!)

PR #4642, opened 2017-05-11
Make it easier to add new packages to the release script

bmw:
Prep work for packaging all of the DNS plugins. After this PR, the only change that needs to be made to the release script to package a plugin is to add it to the `SUBPKGS_NO_CERTBOT` string.

Merging this single commit PR rather than squashing it prevents the need for handling merge conflicts in ~~my incoming PR that builds off of this to package `certbot-dns-cloudflare`~~ #4644.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -45,7 +45,8 @@ export GPG_TTY=$(tty)
 PORT=${PORT:-1234}
 
 # subpackages to be released
-SUBPKGS=${SUBPKGS:-"acme certbot-apache certbot-nginx"}
+SUBPKGS_NO_CERTBOT="acme certbot-apache certbot-nginx"
+SUBPKGS="certbot $SUBPKGS_NO_CERTBOT"
 subpkgs_modules="$(echo $SUBPKGS | sed s/-/_/g)"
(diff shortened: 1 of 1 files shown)


tools/release.sh
```diff
@@ -45,7 +45,8 @@ export GPG_TTY=$(tty)
 PORT=${PORT:-1234}
 
 # subpackages to be released
-SUBPKGS=${SUBPKGS:-"acme certbot-apache certbot-nginx"}
+SUBPKGS_NO_CERTBOT="acme certbot-apache certbot-nginx"
+SUBPKGS="certbot $SUBPKGS_NO_CERTBOT"
```
bmw: I don't think it's necessary. It's nice in theory, but:

1. We've never used the feature. All of our packages are in lock step (e.g. Certbot x.y.z requires acme x.y.z) so releasing a subset of our packages doesn't make a lot of sense right now.
2. It isn't even respected by the current script in `master`. Certbot is always packaged, we expect there to be 12 lines of hash output, etc.

I think if we ever need this feature in the future, we can properly add it.

ohemorange: sgtm

Issue #4080, opened 2017-01-18
Refactor/rewrite release script


bmw:
We should sign releases only after they're full tested. This is more secure and saves me time so I don't have to keep signing releases only to throw them away.

We probably also don't need code for development releases anymore (and if we do, they're in our git history).

bmw:
This needs to be modified to easily release a subset of our packages.

PR #4644, opened 2017-05-11
certbot-dns-cloudflare packaging

### LINKED ISSUE #2208: Install Python3 via EPEL on CentOS6 with letsencrypt-auto?

bmw:
Built on #4642.

### How to package official plugins

These instructions currently only apply to packages that at least support Python 2.6 and 2.7. If the plugin doesn't support both of these Python versions, hold off on packaging while I resolve #2208 and #4080.

Packaging a new plugin requires collecting hashes of dependencies and adding the package to the release script. After #4642, the latter can be done by simply modifying the variable `SUBPKGS_NO_CERTBOT`.

Collecting hashes is a bit trickier. It is common for Python packages to have different dependencies depending on the Python version. The only released code where the hashes are used is `certbot-auto` which only works with Python 2.6 and Python 2.7. You need to gather hashes for both of these versions of Python.

Here's an in progress [script](https://github.com/certbot/certbot/files/994210/hash.txt) for helping with this. You need to place it somewhere inside the Certbot repo, but if you run it providing a space delimited list of [requirement specifiers](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers) on the command line, it will update `letsencrypt-auto-source/pieces/dependency-requirements.txt` to contain all of the new hashes needed to install those packages on this system. You should run the script once on a system where `python` is Python 2.7 and once where it is Python 2.6. (I personally use the CentOS 6 docker image for Python 2.6.) You'll want to include the combination of the hashes found by this script from both systems in `dependency-requirements.txt`.

When you know the packages that will be installed, it's not a bad idea to check the hashes from different networks if you're feeling paranoid as the only signature checking done by `pip` is HTTPS. Here is a much rougher [script](https://github.com/certbot/certbot/files/994253/hash2.txt) that I used here to get the hashes on different systems.

Once you're happy with `dependency-requirements.txt`, run `letsencrypt-auto-source/build.py` to update `letsencrypt-auto-source/letsencrypt-auto` as described [here](https://certbot.eff.org/docs/contributing.html#updating-certbot-auto-and-letsencrypt-auto). Let me know if you have any questions!


bmw:
Thanks for the review.

Going to wait for @ohemorange (or @pde) to review #4642 before I merge this. If they have a chance to look at this too that'd be great, but it's not the end of the world.

zjs:
Now that #4642 has been merged, I think this can be merged too.

bmw:
We still don't know how we're packaging these plugins so I'm closing this PR for now. I'm keeping the branch though because the changes may here may prove to be useful when we determine the path we're going to take.

Issue #4638, opened 2017-05-11
Support --certbot-dir?


pde:
There are a bunch of cases where for various reasons people want to change Certbot's directories, including running as non-root, testing things, and even [performance through parallelism](https://github.com/certbot/certbot/pull/4449#issuecomment-299802507). The UI for that is quite awkward, with three flags (`--config-dir`, `--work-dir` and `--logs-dir`) that the user has to provide.

Perhaps we should support a `--certbot-dir` flag that sets `--config-dir`, but also sets `--work-dir` to something like `$CONFIG_DIR/state` and `--logs-dir` to `$CONFIG_DIR/logs`, so the user only needs to set one flag.

bmw:
Related: #3079

bmw:
Closing in favor of the older #3079 which discusses a flag like this as well as other options.

Issue #3079, opened 2016-05-26
Add a --user flag for certbot to run as non-privileged


jsha:
Certbot currently needs root for these reasons:

All:
- Write to /var/log/letsencrypt/
- (Create and) write to /etc/letsencrypt/
- Write to /var/lib/letsencrypt

Apache:
- Write to /etc/apache2
- Run `apachectl`

Standalone:
- Bind to port 443

Manual:
- None

Lots of third-party Let's Encrypt clients were written because people want to use something like manual mode without root. Right now, certbot lets you do that, but it's complicated. You have to set `--work-dir`, `--conf-dir`, and `--log-dir`. I propose that we add a `--user` flag, or `--no-root` or `--unprivileged` or similar. That would default those three log files to point to `$HOME/certbot/work`, `$HOME/certbot/`, and `$HOME/certbot/logs`, respectively. This would also allow standalone mode to work as non-root, so long as the user was granted the privilege to bind 443.

Slight disadvantage: The keys wind up owned by non-root. But I think that's fine.

jsha:
Another possible flag name: `--nosudo`, in honor of diafygi/letsencrypt-nosudo.

pde:
Related: #2964

bmw:
@jsha, you and I also talked out of band about providing a flag like `--dir-prefix` which would cause `certbot` to create the config, work, and logs dir under the directory passed in `--dir-prefix`. Do you prefer `--user`/`--nosudo` to this? Do you want both?

jsha:
One or the other is better, not both. I've gone back and forth about which approach I prefer, and I'm definitely interested in your thoughts. I think `--prefix` is a flag that describes what it *does*, which is appealing to us as developers. I think `--user` (or `--nosudo`) is a flag that describe what the user wants to accomplish, which probably makes easier to understand and more appealing for end-users. So right now I'm leaning towards that.

bmw:
I think I agree with you that `--nosudo` is better for the user. It's a little less flexible than `--prefix`, however, all of this simply avoids users having to set `--config-dir`, `--work-dir`, and `--logs-dir` manually so people who want the extra control have it either way.

We'll have to figure out the best place for `--nosudo` though. #843 proposed to permanently move Certbot directories to those owned by the user and suggested following XDG Base Directory Spec.

bmw:
Related: #973

bmw:
[...]

PR #4645, opened 2017-05-11
--certbot-dir is easier to set that --config-dir, --logs-dir, --work-dir

pde:
Implements #4638

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -557,6 +557,19 @@ class HelpfulArgumentParser(object):
                 "Flag for non-interactive mode and {0} conflict".format(
                     constants.FORCE_INTERACTIVE_FLAG))
 
+        if parsed_args.certbot_dir:
+            new_dirs = (
+                   ("config_dir", parsed_args.certbot_dir),
+                   ("logs_dir", os.path.join(parsed_args.certbot_dir, constants.LOGS_SUBDIR)),
+                   ("work_dir", os.path.join(parsed_args.certbot_dir, constants.WORK_SUBDIR)))
+            for var, val in new_dirs:
+                if set_by_cli(var) and parsed_args.__getattribute__(var) != val:
+                    raise errors.Error("Conflicting values for --{0}: --certbot-dir "
+                        "implies {1} but --{0} says {2}".format(
+                        var.replace("_", "-"), val, parsed_args.__getattribute__(var)))
+                else:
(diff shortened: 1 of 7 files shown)


certbot/cli.py
```diff
@@ -1190,6 +1203,9 @@ def _paths_parser(helpful):
         help=config_help("work_dir"))
     add("paths", "--logs-dir", default=flag_default("logs_dir"),
         help="Logs directory.")
+    add("paths", "--certbot-dir", default=None,
+        help="Asks Certbot to keep all of its files below a given directory. --certbot-dir $DIR "
+             "is equivalent to --config-dir $DIR --work-dir $DIR/.internal/ --logs-dir $DIR/logs")
```
zjs: Can you reference `constants.WORK_SUBDIR` and `constants.LOGS_SUBDIR` in the help message to ensure the message is updated if the constants ever change?

certbot/lock.py
```diff
@@ -80,7 +80,9 @@ def _try_lock(self, fd):
                 logger.debug(
                     "A lock on %s is held by another process.", self._path)
                 raise errors.LockError(
-                    "Another instance of Certbot is already running.")
+                    "Another instance of Certbot is already running. If for "
+                    "some reason you really want to run two Certbots simultaneously, "
+                    "set --certbot-dir to a different value.")
```
zjs: This seems distinct enough from adding `--certbot-dir` that it probably belongs in a separate PR.

It also seems like @bmw's comment from #4655, https://github.com/certbot/certbot/pull/4655#pullrequestreview-37966820, still applies:

> Additionally, I think this is the wrong place for this error message. If we fail to acquire a lock for the Apache or Nginx plugins, this exception about `--config-dir` and friends gets raised. Instead, if you'd like to see an error message like this, I think it should be added to `certbot.util.set_up_core_dir` by having us catch the `LockError` and raise a new error message. We already do this there in the event of a permission problem so adding another case for a `LockError` fits nicely in my opinion.

bmw:
This has had unaddressed review feedback on it for a month and a half and when this was discussed in #3079, Jacob and I both agreed we liked a `--nosudo` or `--user` flag better than a flag like this as it's more friendly and less verbose to the average user. With this in mind, I'm closing this PR in favor of seeing one like we discussed in #3079 and/or review feedback being addressed.

Issue #4504, opened 2017-04-14
Allow a user to enable/disable autorenewal of certs


ohemorange:
From #338.

This should probably be done via a subcommand or flag to add a field to the renewal config file, which is tested during `renew`.

SwartzCr:
Here's a more full explanation from @zjs from that pr:

The "disable" operation would update the appropriate renewal configuration file with a property. When renew is identifying the certificates to renew, it would ignore/skip any which have this property.

From a user's point of view, this would allow them to:

Create several certificates.
Disable automatic renewal on one or more.
Put a call to certbot renew in their crontab to automatically renew all but those disabled in step 2.
This PR seems to do something else: it adds a flag to globally disable renewal. Do you have a use case in mind for this?

SwartzCr:
@ohemorange do you feel that this is still needed? Or do you think users could manage this with a renewal hook?

ohemorange:
I'm just the messenger here - I was just separating the previous PR into its component parts.

r5d:
I would like to work on this issue.

I've made a rough patch: https://github.com/certbot/certbot/compare/master...r5d:4504-disable-enable-renew-cert-patch-02 (want to know if I'm going in the right direction).

The enable/disable workflow:

The renewal conf for (test) cert `ara.ricketyspace.net` looks like
this initially:
```
  # renew_before_expiry = 30 days
   version = 0.22.0.dev0
  archive_dir = /tmp/leitfJ7Z/conf/archive/ara.ricketyspace.net
  cert = /tmp/leitfJ7Z/conf/live/ara.ricketyspace.net/cert.pem
  privkey = /tmp/leitfJ7Z/conf/live/ara.ricketyspace.net/privkey.pem
  chain = /tmp/leitfJ7Z/conf/live/ara.ricketyspace.net/chain.pem
  fullchain = /tmp/leitfJ7Z/conf/live/ara.ricketyspace.net/fullchain.pem

  # Options used in the renewal process
  [renewalparams]
  account = d9302309ddf75295fc278b68b5df89b4
  work_dir = /tmp/leitfJ7Z/work
  webroot_path = /home/s/pub,
  server = https://acme-staging.api.letsencrypt.org/directory
  authenticator = webroot
  logs_dir = /tmp/leitfJ7Z/logs
  installer = None
  config_dir = /tmp/leitfJ7Z/conf
  [[webroot_map]]
  ara.ricketyspace.net = /home/s/pub
```

After executing:
```bash
[...]

PR #4647, opened 2017-05-12
Adds --disable-renewal and --enable-renewal to allow a user to prevent the auto-renew of certs

2over12:
Implements #4504
with --enable-renewal and --disable-renewal


2over12:
Ah I miss interpreted what was wanted I can see that this probably would not be particularly useful so this PR should be discarded.

SwartzCr:
No worries @2over12 - let us know if you have any interest in giving it another try! We'd be happy to help you make sure you're implementing the right thing if you want to work on it!

PR #4646, opened 2017-05-12
Package certbot-dns-digitalocean

zjs:
Builds on #4644 .

* Add DigitalOcean dependency hash
* Add certbot-dns-digitalocean to the release script
* Rebuild letsencrypt-auto

PR #4651, opened 2017-05-12
 Package certbot-dns-google

zjs:
Builds on #4646. (The only new commit is ec35883.)

* Add Google dependency hashes (python 2.6 and 2.7)
* Add certbot-dns-google to the release script
* Rebuild letsencrypt-auto


bmw:
Since we still don't know how we're packaging these plugins but this PR may turn out to be useful, I pushed this work into the `zjs-google-cloud-dns-packaging` branch and am closing this PR for now.

PR #4650, opened 2017-05-12
support version 0.12.0 of configargparse

ryanpineo:
fixes #4648

PR #4652, opened 2017-05-13
Fix configargparse 0.12.0

bmw:
Built on #4650.

Because we define the `ShowWarning` class inside the function, we actually are adding multiple different `ShowWarning` classes to the object. This doesn't hurt anything besides performance and I started to fix it but we should only apply the minimal patch for now. This is not a behavior change.

### DIFF
--- a/certbot/tests/util_test.py
+++ b/certbot/tests/util_test.py
@@ -6,6 +6,7 @@ import shutil
 import stat
 import unittest
 
+import configargparse
 import mock
 import six
 from six.moves import reload_module  # pylint: disable=import-error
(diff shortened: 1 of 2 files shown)


bmw:
I modified this PR so tests will pass on the 0.14.x branch. I also added an additional commit to clean up the code here a bit, however, I don't plan on including it in 0.14.1.

bmw:
Can I get another review on this? I changed some things later on Friday and I've set up GitHub to dismiss stale reviews automatically.

certbot/tests/util_test.py
```diff
@@ -368,6 +368,29 @@ def test_help(self):
...
 
+    def test_set_constant(self):
+        """Test when ACTION_TYPES_THAT_DONT_NEED_A_VALUE is a set.
+
+        This variable is a set in configargparse versions < 0.12.0.
+
+        """
+        self._test_constant_common(set)
+
+    def test_tuple_constant(self):
```
ohemorange: the comment needs switching from `set` to `tuple`

bmw: fixed

ohemorange: one more

bmw: no more

PR #4654, opened 2017-05-13
Provide advice to people bitten by locking

pde:


Hoten:
I lost connectivity to my server while running `certbot certonly ...`. I tried running the command again when I reconnected, but got the error `Another instance of Certbot is already running.`. Waited a bit, tried again, same error.

I looked at how the locking mechanism worked and determined I should delete all the `.certbot.lock` files on my system: `sudo find / -name .certbot.lock -exec rm {} \;` My earlier commands worked after this.

Hopefully this helps others hurt by locking.

bmw:
I think in your case, Certbot was actually still running on your server, probably waiting for user input. Certbot uses [OS system calls](http://pubs.opengroup.org/onlinepubs/009695399/functions/lockf.html) to provide locking which should automatically be released when Certbot exits. This even happens if you cause Certbot to exit by running something like `kill -9`.

I'm glad you found a workaround, but in general the best way to work around this problem is to kill the currently running version of Certbot so a new instance can grab the lock.

Hoten:
That makes way more sense. Thanks for sharing your insight.

PR #4655, opened 2017-05-13
0.14.x: provide advice to people bitten by locking

pde:


certbot/lock.py
```diff
@@ -80,7 +80,9 @@ def _try_lock(self, fd):
                 logger.debug(
                     "A lock on %s is held by another process.", self._path)
                 raise errors.LockError(
-                    "Another instance of Certbot is already running.")
+                    "Another instance of Certbot is already running.\n\n"
+                    "If you really need to run two Certbot instances at once, provide a "
+                    "distinct --config-dir, --work-dir, and --logs-dir to each of them.")
```
zjs: Exception messages that include newlines can make logs harder to read/grep. It can be especially annoying when using log parsing/monitoring tools. I'm not sure it's an issue for certbot logs.

A different way to write this error message would be: "Another instance of Certbot is already running using the same --config-dir, work-dir, or --logs-dir." This isn't quite as helpful, but it's a lot more concise and I think it's enough to make the solution clear.

zjs:
> Instead, if you'd like to see an error message like this, I think it should be added to `certbot.util.set_up_core_dir` by having us catch the `LockError` and raise a new error message. We already do this there in the event of a permission problem so adding another case for a `LockError` fits nicely in my opinion.

:+1:

pde:
Closing this version of the PR...

PR #4689, opened 2017-05-18
Update options-ssl-nginx.conf in`prepare` if it hasn't been manually modified

### LINKED ISSUE #4559: Update options-ssl-nginx.conf
Add a mechanism for updating it, preferably taking into account that some people manually modify that file.

ohemorange:
Fixes #4559.


ohemorange:
Debating the best place to put `PREVIOUS_SSL_OPTIONS_HASHES`. Ideally we'd also put a message somewhere indicating that if `options-ssl-nginx.conf` is updated, so should that value be. Or, we could add logic to create and save the hash to the release script.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -867,3 +875,18 @@ def install_ssl_options_conf(options_ssl):
...
+    digest = crypto_util.sha256sum(options_ssl)
+    if digest in PREVIOUS_SSL_OPTIONS_HASHES: # safe to update
+        shutil.copyfile(constants.MOD_SSL_CONF_SRC, options_ssl)
+    elif digest == CURRENT_SSL_OPTIONS_HASH: # already up to date
+        return
+    else: # not safe to update
+        new_filename = options_ssl + ".new"
+        shutil.copyfile(constants.MOD_SSL_CONF_SRC, new_filename)
+        # print warning
+        logger.warning("%s has been manually modified; updated ssl configuration options "
```
bmw: One thing that's slightly awkward is if you run Certbot with the Nginx plugin once, modify `options-ssl-nginx.conf`, and then run it again, it tells you an updated file has been saved and you should update your modified version despite them being the same base file. If users are exposed to this behavior of Certbot telling you to update your file despite them being the same, I could see them adding `--nginx-no-tls-updates` just to get Certbot to shut up.

There are are number of ways to mitigate this, but basically I think we want to save `CURRENT_SSL_OPTIONS_HASH` (or the original file) somewhere and check if it's different than our current file before creating the `.new` file with a warning. `/var/lib/letsencrypt` would be nice, but then you have the slightly awkward fact that `options-ssl-nginx.conf` is stored in one configurable directory location and it's original hash/contents are stored in another. Perhaps we could just create a new (hidden) file/folder in `/etc/letsencrypt` to contain this?

This is definitely one downside to doing things at runtime that I didn't think of. If we feel this is too hard/complex to solve correctly, we could go back to your original plan and get things in a sane state for OS packages, `certbot-auto`, `docker`, (and `pip`?), but I still think that's more work and causes Certbot to have different behavior depending on how it's installed.

certbot-nginx/certbot_nginx/tests/configurator_test.py
```diff
@@ -537,6 +538,56 @@ def test_staple_ocsp(self):
...
+            self.config_path, self.config_dir, self.work_dir, self.logs_dir)
+
+    def _call(self):
+        from certbot_nginx.configurator import install_ssl_options_conf
+        install_ssl_options_conf(self.config.mod_ssl_conf)
+
+    def _assert_current_file(self):
+        self.assertTrue(os.path.isfile(self.config.mod_ssl_conf))
+        from certbot_nginx.configurator import CURRENT_SSL_OPTIONS_HASH
+        self.assertEqual(crypto_util.sha256sum(self.config.mod_ssl_conf), CURRENT_SSL_OPTIONS_HASH)
```
bmw: The nice thing about a function like this is our unit tests will fail if the file is updated without updating the hashes. It might be a good idea to put comments in here reminding the developer what needs to happen to the constants in `configurator.py` to properly fix test failures.

certbot/crypto_util.py
```diff
@@ -353,3 +354,11 @@ def _notAfterBefore(cert_path, method):
     if six.PY3:
         timestamp_str = timestamp_str.decode('ascii')
     return pyrfc3339.parse(timestamp_str)
+
+
+def sha256sum(filename):
+    """Compute a sha256sum of a file"""
```
bmw: Since this module is [publicly documented](https://certbot.eff.org/docs/api/crypto_util.html), it'd be nice to document parameters and return values like the rest of the functions in this module.

Our tests and coveralls also think this function doesn't have test coverage because it's only tests are in the Nginx plugin. I think it'd be good to add a simple test or two to `certbot/tests/crypto_util_test.py`.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -62,6 +62,14 @@
     ['#', " } # managed by Certbot"],
 ]
 
+PREVIOUS_SSL_OPTIONS_HASHES = [
+    '0f81093a1465e3d4eaa8b0c14e77b2a2e93568b0fc1351c2b87893a95f0de87c',
```
ohemorange: file was renamed without changing its contents in that commit

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -867,3 +875,18 @@ def install_ssl_options_conf(options_ssl):
...
         shutil.copyfile(constants.MOD_SSL_CONF_SRC, options_ssl)
+        return
+    # there's already a file there. if it exactly matches a previous file hash,
+    # we can update it. otherwise, create a .new and print a warning.
+    digest = crypto_util.sha256sum(options_ssl)
+    if digest in PREVIOUS_SSL_OPTIONS_HASHES: # safe to update
+        shutil.copyfile(constants.MOD_SSL_CONF_SRC, options_ssl)
+    elif digest == CURRENT_SSL_OPTIONS_HASH: # already up to date
+        return
+    else: # not safe to update
```
ohemorange: This one we can fix by also checking the contents of the `.new` file, and do nothing if it matches `CURRENT_SSL_OPTIONS_HASH`. Then this only happens once per time we update the file.

certbot-nginx/certbot_nginx/tests/configurator_test.py
ohemorange: Oh yeah, good point! This totally solves the "how to remind devs to update things" problem. Commenting is a great idea :+1:.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Ok, what about if we do what I said above (only print the warning if .new is different), but also create a .new when we create the original file?

ohemorange: `There are are number of ways to mitigate this, but basically I think we want to save CURRENT_SSL_OPTIONS_HASH (or the original file) somewhere and check if it's different than our current file before creating the .new file with a warning.`

Not sure what you mean; this is what we already do. Do you mean after they've edited it, save a copy somewhere else?

ohemorange: We could also do checking of certbot version -- only do things once per version. I hasn't wanted to do that because it can get messy, but if we don't always want to create `.new`, there's various ways we could work version number in.

certbot-nginx/certbot_nginx/configurator.py
bmw: Right, but I don't think you have a hash of that version of the file. It looks like the oldest hash of the file you have is from commit 6b639679e8df0c860e16f571dc831bc339132623 where we changed the name of the SSL session cache.

ohemorange: Hm, thought I did, I'll double check though.

certbot-nginx/certbot_nginx/configurator.py
bmw: I personally think it's nice that we continue to make noise about the problem until the user either fixes the file or tells us they aren't interested in updates.

certbot-nginx/certbot_nginx/configurator.py
bmw: > Not sure what you mean; this is what we already do. Do you mean after they've edited it, save a copy somewhere else?

I mean save a copy somewhere on disk whenever we write `options-ssl-nginx.conf`. This way we can tell what file we placed there and whether or not we really have a new version of the file that contains security updates.

> We could also do checking of certbot version -- only do things once per version. I hasn't wanted to do that because it can get messy, but if we don't always want to create .new, there's various ways we could work version number in.

I think a hash, copy of the file, or version number stored somewhere would mitigate this problem. I kind of like a hash best because it's not likely to change with every update but any of these approaches is fine.

In my opinion, ideally we'd like to:

1. Only create `.new` when we actually have a new versions of the file.
2. Make as much noise as we reasonably can about the issue until the user resolves it one way or another.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: that seems like a cruel experience to subject the user to.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Sticking a hash somewhere when we write and writing `.new` out immediately would have the same effect, I don't actually have a preference between them. I'm happy to write a hash out to some file somewhere.

Also, it occurs to me that we don't actually have to write out a `.new` -- we could just point them to `constants.MOD_SSL_CONF_SRC`.

I really don't think we should make lots of noise over and over until they tell us to shut up. Especially because they might update the contents of a field we changed, but have the file still be different from the updated version. Then, they'd have to tell us with a field or flag set somewhere, which I definitely don't want to be adding.

certbot-nginx/certbot_nginx/configurator.py
bmw: This is the current experience of this PR and cruel isn't how I look at it. We're telling the user there's a (potentially serious) problem that needs their attention the only way we really have. Next time they check their cron e-mail, look at their logs, or run Certbot manually, they'll see the message and can react accordingly. I think sending only one message basically limits people who will see it to those who have cron e-mail set up properly and pay attention to it.

certbot-nginx/certbot_nginx/configurator.py
bmw: > Sticking a hash somewhere when we write and writing .new out immediately would have the same effect, I don't actually have a preference between them. I'm happy to write a hash out to some file somewhere.

If you write out the `.new` file at the same time as you write the live file it has the same effect. I have a preference for either including a special value/version in the file itself or creating an effectively hidden file containing the original contents/hash than to create the publicly visible `.new` file initially. If we did a hidden file, it could be an actually hidden file in `/etc/letsencrypt` or something placed somewhere like `/var/lib/letsencrypt` (although the latter opens us up to other problems I identified above).

 > Also, it occurs to me that we don't actually have to write out a .new -- we could just point them to constants.MOD_SSL_CONF_SRC.

Yeah that'd be fine too.

> I really don't think we should make lots of noise over and over until they tell us to shut up. Especially because they might update the contents of a field we changed, but have the file still be different from the updated version. Then, they'd have to tell us with a field or flag set somewhere, which I definitely don't want to be adding.

Very good point. Forcing the user to either take our file as is or turn off updates entirely is a bad idea. Basically we'd want some way to know the user is happy with the current state of the file, but I don't think there's really a clean way to do that. `--nginx-no-tls-updates-until-next-time` is pretty ridiculous.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: figured it out. we used to merge commits, not squash them, so going to where it was changed then `reset`ing to get the hash from right before then is inaccurate.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -62,6 +62,15 @@
...
 ]
 
+PREVIOUS_SSL_OPTIONS_HASHES = [
+    '0f81093a1465e3d4eaa8b0c14e77b2a2e93568b0fc1351c2b87893a95f0de87c',
+    '9a7b32c49001fed4cff8ad24353329472a50e86ade1ef9b2b9e43566a619612e',
+    'a6d9f1c7d6b36749b52ba061fff1421f9a0a3d2cfdafbd63c05d06f65b990937',
+    '7f95624dd95cf5afc708b9f967ee83a24b8025dc7c8d9df2b556bbc64256b3ff',
+]
+
+CURRENT_SSL_OPTIONS_HASH = '394732f2bbe3e5e637c3fb5c6e980a1f1b90b01e2e8d6b7cff41dde16e2a756d'
```
bmw: nit: I'm sorry I didn't bring this up before, but it may be worth moving these values into `constants.py` given that's the purpose of the file and it's a lot smaller that configurator.py. That file also has docstrings for all of the constants which I think is nice.

You can ignore this comment if you want though.

certbot-nginx/certbot_nginx/constants.py
```diff
@@ -17,6 +17,9 @@
 """Path to the nginx mod_ssl config file found in the Certbot
 distribution."""
 
+UPDATED_MOD_SSL_CONF_DIGEST = ".updated-mod-ssl-conf-digest.txt"
```
bmw: We may want to change the name of the file here as I can see it conflicting with what we do for Apache.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: great point!

ohemorange:
I think I'll refactor and add apache in a follow-up PR, as long as we're good with the various choices made for this one that will be paralleled there.

certbot-nginx/certbot_nginx/constants.py
```diff
@@ -17,9 +17,21 @@
 """Path to the nginx mod_ssl config file found in the Certbot
 distribution."""
 
-UPDATED_MOD_SSL_CONF_DIGEST = ".updated-mod-ssl-conf-digest.txt"
+UPDATED_MOD_SSL_CONF_DIGEST = ".updated-mod-ssl-conf-nginx-digest.txt"
```
bmw: I should have noticed this before, but want to know something fun? `mod_ssl` refers to the SSL module in the Apache plugin. The Apache plugin was written first and the Nginx plugin very closely mimics that so the name `mod-ssl` is a bit of a misnomer.

Since we're basically stuck with this name forever (unless we want to do extra work in the future to migrate it), we may want to change it to something like `.updated-nginx-ssl-conf-digest.txt`.

At some point, we should probably change references to `mod_ssl` in the code too, but most weren't created by this PR and I think it makes sense to use consistent naming for those that were.

certbot-nginx/certbot_nginx/tests/configurator_test.py
```diff
@@ -537,6 +539,79 @@ def test_staple_ocsp(self):
...
+
+    def _call(self):
+        from certbot_nginx.configurator import install_ssl_options_conf
+        install_ssl_options_conf(self.config.mod_ssl_conf, self.config.updated_mod_ssl_conf_digest)
+
+    def _assert_current_file(self):
+        """If this is failing, remember that configurator.PREVIOUS_SSL_OPTIONS_HASHES and
+           configurator.CURRENT_SSL_OPTIONS_HASH must be updated when self.config.mod_ssl_conf
+           is updated. Add CURRENT_SSL_OPTIONS_HASH to PREVIOUS_SSL_OPTIONS_HASHES and set
+           CURRENT_SSL_OPTIONS_HASH to the hash of the updated self.config.mod_ssl_conf."""
```
bmw: You should update this comment in response to moving the values to `constants.py`.

ohemorange: ah, missed the couple in the comment

certbot-nginx/certbot_nginx/constants.py
ohemorange: Changing the filename. Since we (inaccurately) call the property `mod_ssl_conf`, the variable name should probably match. Not too worried about that in general, since we also call them vhosts and server blocks interchangeably.

bmw: Yeah I agree the variable name should not be changed.

PR #4690, opened 2017-05-18
client: allow callers to add information to the user agent

### LINKED ISSUE #4682: Allow callers to add information to the User Agent (without replacing it entirely)
(Based on IRC discussion with @nuxwin.)

Currently, there is not a good way for people re-packaging or wrapping Certbot to influence its user agent string. They may which to do this so that stats/metrics are available to boulder.

The `--user-agent` flag allows a caller to set the user agent, but that sets the string in its entirely, losing all of the information which is included by default.

For this use case, it is desirable to both the Certbot team and the party re-packaging Certbot that the custom user agent match Certbot user agent as closely as possible: this greatly simplifies the "parsing problem" and allows data for the re-packaged version to be collected along side _or_ separately from vanilla certbot.

Supporting an additional `--user-agent-comment` flag seems like an easy way to address this.

This comment, if set, could be inserted into the main (product-level) comment, following the OS information.

This argument and `--user-agent` would be mutually exclusive.

cc @pde

zjs:
This change introduces a new flag to allow callers to add information to the user agent without replacing it entirely.

This allows people re-packaging or wrapping Certbot to influence its user agent string. They may which to do this so that stats/metrics related to their distribution are available to boulder.

This is beneficial for both the Certbot team and the party re-packaging Certbot as it allows the custom user agent to match the Certbot user agent as closely as possible, allowing data about use of the re-packaged version to be collected along side or separately from vanilla certbot.

Resolves #4682


nuxwin:
@zjs 

I will fall in love with you soon ;) Thank for your commit there ;)

bmw:
@pde's the one who's spent the most time playing and tinkering with Certbot's user agent by far so I'd like to get his opinion if he's available to give this a review.

zjs:
@pde, any feedback?

nuxwin:
@pde 

Great ;)

PR #4691, opened 2017-05-18
Fix defaults on older systems

### LINKED ISSUE #3987: Certbot help showing default arguments improperly
When issuing:
certbot-auto --help certonly

I get output like:
[...]
  --manual              Provide laborious manual instructions for obtaining a
                        cert (default: <certbot.cli._Default object at
                        0x255e410>)
  --webroot             Obtain certs by placing files in a webroot directory.
                        (default: <certbot.cli._Default object at 0x255e490>)

Version numbers:
certbot-auto --version
letsencrypt 0.9.3
python --version
Python 2.7.3

This works properly with the same certbot, but with python 2.7.12.

bmw:
This fixes #3987 as the call to `set_by_default` can be removed entirely. Additionally, `logger.warning` can be used rather than writing to `stderr` directly because #3184 has been resolved and we're guaranteed to having logging setup. `supported_challenges_validator` was moved to `SupportedChallengesAction` so `argparse.ArgumentError` can be easily used to provide nice error output.

The easiest way to test this is on Debian 7.0 Wheezy. After installing Certbot using our developer instructions, if you run `certbot --help all` on `master`, you'll see defaults like:
```
  --nginx-server-root NGINX_SERVER_ROOT
                        Nginx server root directory. (default:
                        <certbot.cli._Default object at 0x3c89210>)
```
If you check out this branch, the problem goes away.

### DIFF
--- a/certbot/plugins/standalone.py
+++ b/certbot/plugins/standalone.py
@@ -146,6 +146,10 @@ class SupportedChallengesAction(argparse.Action):
 
         unrecognized = [name for name in challs
                         if name not in challenges.Challenge.TYPES]
+
+        # argparse.ArgumentErrors raised out of argparse.Action objects
+        # are caught by argparse which prints usage information and the
+        # error that occurred before calling sys.exit.
         if unrecognized:
             raise argparse.ArgumentError(
                 self,
--- a/certbot/plugins/standalone.py
+++ b/certbot/plugins/standalone.py
@@ -112,41 +112,6 @@ class ServerManager(object):
 SUPPORTED_CHALLENGES = [challenges.TLSSNI01, challenges.HTTP01]
 
 
-def supported_challenges_validator(data):
-    """Supported challenges validator for the `argparse`.
-
-    References to "dvsni" are automatically converted to "tls-sni-01".
-
-    :param str data: comma delimited list of challenge types
-
-    :returns: validated and converted list of challenge types
-    :rtype: str
-
-    """
-    challs = data.split(",")
-
-    # tls-sni-01 was dvsni during private beta
-    if "dvsni" in challs:
-        logger.info("Updating legacy standalone_supported_challenges value")
-        challs = [challenges.TLSSNI01.typ if chall == "dvsni" else chall
-                  for chall in challs]
-        data = ",".join(challs)
-
-    unrecognized = [name for name in challs
-                    if name not in challenges.Challenge.TYPES]
-    if unrecognized:
-        raise argparse.ArgumentTypeError(
-            "Unrecognized challenges: {0}".format(", ".join(unrecognized)))
-
-    choices = set(chall.typ for chall in SUPPORTED_CHALLENGES)
-    if not set(challs).issubset(choices):
-        raise argparse.ArgumentTypeError(
-            "Plugin does not support the following (valid) "
-            "challenges: {0}".format(", ".join(set(challs) - choices)))
-
-    return data
-
-
 class SupportedChallengesAction(argparse.Action):
     """Action class for parsing standalone_supported_challenges."""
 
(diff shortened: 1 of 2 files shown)


certbot/plugins/standalone_test.py
```diff
@@ -93,10 +91,10 @@ def test_correct(self):
 
     def test_unrecognized(self):
         assert "foo" not in challenges.Challenge.TYPES
-        self.assertRaises(argparse.ArgumentTypeError, self._call, "foo")
+        self.assertRaises(SystemExit, self._call, "foo")
```
ohemorange: why doesn't this raise an `ArgumentError`?

certbot/plugins/standalone.py
```diff
@@ -114,39 +112,53 @@ def running(self):
 SUPPORTED_CHALLENGES = [challenges.TLSSNI01, challenges.HTTP01]
 
 
-def supported_challenges_validator(data):
-    """Supported challenges validator for the `argparse`.
+class SupportedChallengesAction(argparse.Action):
+    """Action class for parsing standalone_supported_challenges."""
 
-    It should be passed as `type` argument to `add_argument`.
+    def __call__(self, parser, namespace, values, option_string=None):
```
ohemorange: Why was the `if` in there originally? How might this be called if not via cli?

certbot/plugins/standalone_test.py
bmw: `argparse` does a really poor job documenting things, but this is from the docstring at the top of the module:
```
    - ArgumentError -- The exception raised by ArgumentParser objects when
        there are errors with the parser's actions. Errors raised while
        parsing the command-line are caught by ArgumentParser and emitted
        as command-line messages.
```
It causes error messages to look like:
```
$ certbot --standalone-supported-challenges foo
usage: 

cert. 
certbot: error: argument --standalone-supported-challenges: Unrecognized challenges: foo
```
rather than something like this when using our own error type:
```
$ certbot --standalone-supported-challenges foo
Unrecognized challenges: foo
```

certbot/plugins/standalone.py
bmw: More `argparse` fun! As documented [here](https://docs.python.org/2.7/library/argparse.html#default), if the default is a `str` and you specify a `type` argument when adding the CLI flag to the parser, the `type` parameter gets called with the default before adding it to the namespace. As a result, sometimes when `type` was called we were just parsing the default rather than a user provided value so we had to use our `set_by_cli()` function to differentiate between these cases.

This avoids all that because the `action` argument is only called if the user provides a value.

bmw:
As for why this fixes the bug, I explained a little more in the issue this fixes (#3987), but to summarize and add to it, you're right that the problem was our usage of `set_by_cli`.

Before Python 2.7.4, the `type` argument of a CLI flag was called on the default even when `--help` was provided. As a result, if the user ran `certbot --help all`, `supported_challenges_validator` would be called, which would call `set_by_cli`, which would parse command line arguments again using dummy `certbot.cli._Default` values. This second parsing of our command line arguments would finish and `argparse` would print help with the dummy values and exit.

This fixes the problem by removing the need to use `set_by_cli`.

certbot/plugins/standalone.py
```diff
@@ -184,7 +200,7 @@ def __init__(self, *args, **kwargs):
     def add_parser_arguments(cls, add):
         add("supported-challenges",
```
ohemorange: Does this automatically get prefixed with `standalone_` somewhere, since we're calling it from the standalone plugin? Because I see that `namespace.standalone_supported_challenges` is set, instead of `supported_challenges`.

bmw: Yes. Plugin flags are prefixed with the name of the plugin [here](https://github.com/certbot/certbot/blob/master/certbot/plugins/common.py#L69).

certbot/plugins/standalone.py
ohemorange: Is there a reason we weren't just doing this in the first place? It seems to me that all instances of `set_by_cli` could be avoided by using actions rather than sticking logic into `type`.

bmw: This was the only instance of using `set_by_cli()` inside argument parsing. The other places it's used are trying to detect if a value is set by the user or not outside of argument parsing.

`set_by_cli()` was used originally because the validation here was written as an argument for `type` rather than `action` way before `set_by_cli()` was even written. `set_by_cli()` was a simple one line fix until we hit `argparse`'s legacy behavior here.

PR #4693, opened 2017-05-19
Update example CloudFlare authenticator script

mjmasn:
This change removes the fairly broken regex for getting the zone name from the domain/subdomain. 

The example will now work with root domains such as `example.com` (previously `DOMAIN` was a blank string).

For CNAME-d subdomains where `sub.example1.com` -> `another-sub.example2.com`, a possible caveat is that`DOMAIN` will be `example2.com` not `example1.com`. It appears that `certbot` checks both domains anyway (I think?) but I believe returning the SOA domain is preferable because it's likely to be the origin host generating this type of certificate, not the person who set up the CNAME (if they are different people).

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -653,8 +653,8 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not 
    API_KEY="your-api-key"
    EMAIL="your.email@example.com"
 
-   # Strip only the top domain to get the zone id
-   DOMAIN=$(expr match "$CERTBOT_DOMAIN" '.*\.\(.*\..*\)')
+   # Fetch the SOA domain for the domain/subdomain
+   DOMAIN=$(dig soa "$CERTBOT_DOMAIN" | grep -v ^\; | grep SOA | awk '{print $1}';)
 
    # Get the Cloudflare zone id
    ZONE_EXTRA_PARAMS="status=active&page=1&per_page=20&order=status&direction=desc&match=all"


mjmasn:
@zjs I didn't know there was a plugin being developed for this - in that case maybe this PR doesn't matter at all! 

Good point on the subdomain with different NS, seems like a tricky one to solve nicely (or rather an easier one for CloudFlare to solve by allowing searching for a zone by a fully-qualified subdomain). Looking at that plugin it just guesses less specific domain names until it finds the correct record which I guess is a fairly foolproof solution, if a little less efficient.

Feel free to close this.

zjs:
> in that case maybe this PR doesn't matter at all!

Given how long older versions of Certbot can remain in use, I think this is still a good example to include in the documentation (and therefore still a useful PR).

paulproteus:
Out of curiosity, is anything holding up a merge here? Just curious since I saw this PR while looking for things to work on at PyCon.

paulproteus:
cc: @bmw (since @SwartzCr says you can merge things!)

bmw:
@paulproteus, it's waiting for a second review as suggested by zjs.

@joohoi, as the original author of this script, do you have any spare cycles to look at this?

docs/using.rst
```diff
@@ -653,8 +653,8 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not
    API_KEY="your-api-key"
    EMAIL="your.email@example.com"
 
-   # Strip only the top domain to get the zone id
-   DOMAIN=$(expr match "$CERTBOT_DOMAIN" '.*\.\(.*\..*\)')
+   # Fetch the SOA domain for the domain/subdomain
+   DOMAIN=$(dig soa "$CERTBOT_DOMAIN" | grep -v ^\; | grep SOA | awk '{print $1}';)
```
joohoi: We are adding another dependency (even though curl was one already) here, so maybe a note should be added telling the user that dig should be made available. I don't think it's included in many default installations.

bmw:
@mjmasn, I just wanted to check in since it's been over a month since we've seen any activity from you here. Are you planning on addressing review comments yourself or would you like someone else to build on your work and get this merged?

mjmasn:
Hey @bmw, I think realistically I'm not going to have time to look at this in the near future so I'm more than happy for someone else to take this on. Sorry for abandoning a half finished PR.

bmw:
No problem! Thanks for submitting the PR in the first place. I created #4883 to finish this up.

bmw:
Closed based on the comment [here](https://github.com/certbot/certbot/pull/4883#issuecomment-318186804). Please leave a comment if you'd like me to reconsider.

Issue #4520, opened 2017-04-18
Print deprecation warning on experimental systems in certbot-auto


bmw:
We have a number of systems where we have "experimental" bootstrap support but we have OS packages available. We should print a deprecation warning for these systems telling users that support is going away including how to install the OS packages. We will be removing support completely for these systems in a few months. The current experimental systems are:

* Mageia
* Manjaro 
* Gentoo
* FreeBSD
* macOS
* Amazon Linux
* Joyent SmartOS Zone

At the very least, Gentoo, FreeBSD, and macOS have other options but we should check out other systems as well.

SwartzCr:
So here's what I've found:
It seems like we're packaged for Mageia - http://check.mageia.org/cauldron/updates.html
It's possible that this netbsd package works for manjaro - http://cdn.netbsd.org/pub/pkgsrc/current/pkgsrc/README-all.html
It's possible that the arch package of certbot works for Joyent Smart OS
Doubt there's anything for Amazon Linux (they never responded to me)

bmw:
Also, we should print this deprecation warning more often rather than just when we bootstrap the system.

bmw:
certbot-auto has been deprecated, will error out on new systems, and existing installations are no longer receiving updates so I'm closing this.

PR #4697, opened 2017-05-19
Enable unit tests on OS X

yan12125:
Another try for #2978. Currently only the latest Python 2 and 3 are tested. Travis CI workers are already installed with Python 2 from Homebrew and I use a similar trick to install Python 3. As I've mentioned in #2978, testing on older Python versions requires a missing feature on the Travis CI side. It's possible to build older Python versions via third party Python version managers like pyenv, but I'm reluctant to make .travis.yml even messy.

About the mess in pip_install.sh: by default pip uses the prebuilt wheel for cryptography, which brings segmentation fault when running acme tests. Here I decide to use Homebrew to install OpenSSL and build cryptography against it. Homebrew is the recommended way to install certbot on Mac OS X, and the Homebrew team already uses [its own OpenSSL](https://github.com/Homebrew/homebrew-core/blob/master/Formula/certbot.rb#L19) to build certbot. (certbot uses openssl@1.1 while the default openssl target is still 1.0 on Homebrew - I'm not sure which I should use here) A drawback is that certbot-auto won't support MacPorts anymore. Those who prefer MacPorts need to use the certbot package from MacPorts instead.

The --no-binary flag requires pip 0.7.0. As Homebrew updates packages very fast and this flag is used on OS X only, I don't think it can be a problem.

After this PR certbot-auto should be updated to install openssl.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
         - $HOME/.cache/pip
 
 before_install:
-  - 'dpkg -s libaugeas0'
+  - '[ $TRAVIS_OS_NAME == linux ] && dpkg -s libaugeas0 || brew install augeas'
 
 # using separate envs with different TOXENVs creates 4x1 Travis build
 # matrix, which allows us to clearly distinguish which component under
@@ -112,6 +112,9 @@ matrix:
       services: docker
     - python: "2.7"
       env: TOXENV=nginxroundtrip
+    - language: generic
+      env: TOXENV=py27
+      os: osx
 
 
 # Only build pushes to the master branch, PRs, and branches beginning with
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
         - $HOME/.cache/pip
 
 before_install:
-  - '[ $TRAVIS_OS_NAME == linux ] && dpkg -s libaugeas0 || brew install augeas'
+  - '[ $TRAVIS_OS_NAME == linux ] && dpkg -s libaugeas0 || brew install augeas python3'
 
 # using separate envs with different TOXENVs creates 4x1 Travis build
 # matrix, which allows us to clearly distinguish which component under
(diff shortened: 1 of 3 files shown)


yan12125:
Well, apparently Travis CI sets a lower-than-usual ulimit for OS X workers - fixed by forcing a common value in .travis.yml.

bmw:
> by default pip uses the prebuilt wheel for cryptography, which brings segmentation fault when running acme tests

Do you know if this issue been reported/fixed upstream? We currently are pinning an older wheel. If we updated to a newer one would it fix the problem?

> A drawback is that certbot-auto won't support MacPorts anymore. Those who prefer MacPorts need to use the certbot package from MacPorts instead.

There's no change in this PR that drops support for using MacPorts with `certbot-auto` right? It's just that the version of the wheel that's currently being used doesn't work.

This also isn't a huge deal as we're actively working on deprecating macOS support in `certbot-auto` (see #4520).

yan12125:
> Do you know if this issue been reported/fixed upstream? We currently are pinning an older wheel. If we updated to a newer one would it fix the problem?

Great point! I didn't notice the pinned version is an old one 😂 I tried cryptography 1.8.1 and the wheel works just fine on my Sierra machine. Let's see if El Capitan on Travis CI is satisfied or not. If bumping cryptography fixes the crash, is it better to open another issue/PR or I can just do them all in this one?

yan12125:
El Capitan on Travis CI is also satisfied. Now there are no longer certbot-auto and homebrew issues.

bmw:
Awesome! I'm glad updating the version of cryptography used worked. Unfortunately, updating our pinned dependencies is a real pain and it looks like some dependencies were missed. If you're curious, you can read #4644 where I describe the process of how we do this, but I went ahead and made #4755 which handles this for you.

I'll try and get a quick review on that PR and after it lands we can add the changes here and merge this PR.

bmw:
With #4755 merged, if you rebase on/merge those changes into your branch, I'll +1 this PR and merge.

yan12125:
Thanks! Rebased and looks good :-)

yan12125:
Thanks for reviewing!

PR #4701, opened 2017-05-21
RFC 2136 DNS Authenticator

bodgit:
I've been wanting to try out Let's Encrypt using dns-01 challenges but it's always seemed a bit fiddly to set up with chaining of various hook scripts so it was nice to see that lots of new DNS authenticators have recently been added however I couldn't see one for adding the challenge TXT record using RFC 2136 Dynamic Updates, (a.k.a. `nsupdate`) which is supported in BIND and maybe a few other DNS servers. So here's the beginnings of one. It's only dependency is the [http://www.dnspython.org](http://www.dnspython.org) library and you basically need to configure the following things:

1. Target DNS server
2. TSIG Key name
3. TSIG Key secret (Base64-encoded blob)
4. TSIG Key algorithm if you're using something other than HMAC-MD5 so this can be optional

I doubt the code works properly at the moment but I figured I'd get a PR started in case you have any feedback whilst I polish it.


zjs:
Thanks for the PR, @bodgit! This looks really useful.

I'll take a closer look when I have some time, but it looks like you're off to a good start.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -0,0 +1,157 @@
...
+               'RFC 2136 Dynamic Updates.'
+
+    def _setup_credentials(self):
+        self.credetials = self._configure_credentials(
+            'credentials',
+            'RFC 2136 credentials INI file',
+            {
+                'name': 'TSIG key name',
+                'secret': 'TSIG key secret',
+                'algorithm': 'TSIG key algorithm',
```
zjs: If `algorithm` is optional, you probably don't want to include it here. But you may want to include some logic later in `_setup_credentials` to ensure the value is valid if set.

bodgit: Do you have an example for that?

bodgit:
I'm waiting for my test domain to get propagated so I can test it fully but so far it does seem to be adding the TXT record correctly (and then removing it again because it gets `NXDOMAIN` as the domain isn't propagated fully). So apart from the algorithm setting I guess I just need to add some tests now.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
zjs: I think this will be the first DNS plugin with an optional parameter. I posted a very quick sketch of how I would imagine handling an optional, but validated, credential as a PR to your repo: https://github.com/bodgit/certbot/pull/1. This structure results in a much earlier error during the interactive workflow, allowing the user to adjust the file and retry without starting over from the beginning of the interactive workflow. Feel free to use or ignore.

bodgit:
Test domain was finally fixed up and I've verified the plugin now works correctly. I bumped the default propagation delay based on the google authenticator as even with notifies it can take a short while for any slave servers to perform a transfer to pick up the TXT record and 10 seconds was a bit too short as a default.

bodgit:
I've added some basic tests and the test coverage is about 51%, I'm not quite sure how to go about testing this so any insights would be appreciated.

I've also rebased it so there's currently no conflicts.

zjs:
This looks good so far!

> I've added some basic tests and the test coverage is about 51%, I'm not quite sure how to go about testing this so any insights would be appreciated.

Does https://github.com/bodgit/certbot/pull/2 help?

bodgit:
> Does bodgit#2 help?

Yes thanks, that helps a lot, I always get a bit confused with mocking things. Annoyingly I'm not going to be able to finish this off for a week but I'll get back on it when I get back.

zjs:
Hey @bodgit, thanks for your work on this so far! Is there anything I can do to help finish it off?

bodgit:
Sorry, I've had a hectic couple of weeks which stopped me looking at this. I'm going to try and finish this off in the next couple of days.

bodgit:
I've updated the tests and gotten it to 99% coverage with no lint warnings or errors, assuming the tests are satisfactory. I've rebased everything against current master and the Travis CI tests seem to mostly pass, I'm not sure if it's my code causing the failures.

Is there anything else I need to create/generate/commit?

bodgit:
I think the only failure is https://github.com/bodgit/certbot/blob/rfc2136/certbot/plugins/selection.py#L220 which me adding one more plugin has pushed it over the limit of 12 `if` statements.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -0,0 +1,215 @@
...
+            keyring=self.keyring,
+            keyalgorithm=self.algorithm)
+        update.add(rel, record_ttl, dns.rdatatype.TXT, record_content)
+
+        try:
+            response = dns.query.tcp(update, self.server)
+            rcode = response.rcode()
+
+            if rcode == dns.rcode.NOERROR:
+                logger.debug('Successfully added TXT record')
```
zjs: Should other `rcode` values be logged and treated as an error, resulting in a `errors.PluginError` (like in the `except` clause immediately following this)?

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -0,0 +1,215 @@
...
+    description = 'Obtain certs using a DNS TXT record (if you are using BIND for DNS).'
+    ttl = 120
+
+    def __init__(self, *args, **kwargs):
+        super(Authenticator, self).__init__(*args, **kwargs)
+        self.credentials = None
+
+    @classmethod
+    def add_parser_arguments(cls, add):  # pylint: disable=arguments-differ
+        super(Authenticator, cls).add_parser_arguments(add, default_propagation_seconds=60)
```
bodgit: I was testing with two public DNS servers; the master that accepts the dynamic updates and another that just slaves the whole zone from the master. Both are listed as authoritative.

Even with using notifies to minimise the delay between the two servers when there was an update it sometimes still took the slave a while before performing the actual zone transfer upon receipt of a notify. I think BIND puts pending zone transfers into a queue and performs them serially so if there are a lot of zones and/or updates it can sometimes take a bit longer to propagate.

It did work when I still had a propagation delay of 10 seconds but that was just pure luck as the ACME server was choosing the master DNS server for queries. 60 seconds seemed to always be sufficient for everything to settle but it's probably something that's fairly site-specific.

For testing, there's not too much that needs to be set up in BIND. You need to generate a TSIG key:
```
dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST keyname.
```
Then in your `named.conf` you need something like:
```
key "keyname." {
        algorithm hmac-sha512;
        secret "<the Base64-encoded string from the .key file generated above>";
};

zone "example.com." IN {
        type master;
        file "named.example.com";
        // You can either do this which allows the key full access to the zone:
        // allow-update { key "keyname."; };
        // Or you can grant the key access to only a very specific part of the zone:
        update-policy { grant keyname. name _acme-challenge.host.example.com. txt; };
};
```
If you need any help in setting up BIND let me know. You can also use PowerDNS which also supports the same update mechanism but I've not tried that.

certbot-dns-rfc2136/certbot_dns_rfc2136/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+``--dns-rfc2136-credentials``         RFC 2136 credentials_ INI file.
+                                      (Required)
+``--dns-rfc2136-propagation-seconds`` The number of seconds to wait for DNS
+                                      to propagate before asking the ACME
+                                      server to verify the DNS record.
+                                      (Default: 60)
+===================================== =====================================
+
+
+Credentials
```
zjs: I found the `update-policy { grant keyname. name _acme-challenge.host.example.com. txt; };` example you gave helpful when setting up BIND on my system.

Perhaps it would be useful to mention and/or link to something about that in a BIND-specific subsection of this Credentials section, so that users know the minimum necessary privileges they need to grant to the key?

bodgit: I was thinking about doing that until I had 4% battery left. I've added that now.

bodgit:
Rebased again.

zjs:
Thanks, @bodgit! This looks good to me, but lint is unhappy about the length of a couple of lines:
```
************* Module certbot_dns_rfc2136
C: 36, 0: Line too long (113/100) (line-too-long)
C: 82, 0: Line too long (104/100) (line-too-long)
```

bodgit:
My fault for using the most secure TSIG key size :smile:

renne:
Please support CNAMES in case the nameserver does not support RFC 2136/2845 itself and challenge records have to be mapped to another nameserver via CNAMES (different DynDNS zone than domain/validator zone).

noci2012:
See also https://github.com/certbot/certbot/issues/5663
for my experience... & proposed fix.

PR #4704, opened 2017-05-22
updated manual challenge prompt so last sentence is complete

anna-liao:


certbot/plugins/manual.py
```diff
@@ -42,7 +42,8 @@ class Authenticator(common.Plugin):
 
 {validation}
 
-Once this is deployed,"""
+Once this TXT record is deployed,
+Let's Encrypt will be able to check the validity of your domain."""
```
bmw: nit: The protocol that Certbot and Let's Encrypt uses (called ACME) is an open standard and there are other organizations starting to adopt it. Because of this, Certbot can be used with more servers than Let's Encrypt so I wouldn't mention them explicitly here. You could change this to "the ACME server", "the ACME CA", "the certificate authority", etc.

Personally though, I have a slight preference for a different approach. In our instructions for the HTTP challenge, we say:
```
Make sure your web server displays the following content at <url> before continuing:
```
We could do something similar here, delete the "Once this is deployed" line and change the beginning of the prompt to say "Before continuing, please deploy...". This is an aesthetic thing though and if you like your approach better, we can go with it as long as you remove the string "Let's Encrypt".

anna-liao: @bmw Thanks for reviewing this PR!  I updated the wording for a gentle alert to take action before proceeding.  What do you think?

certbot/plugins/manual.py
```diff
@@ -42,7 +42,9 @@ class Authenticator(common.Plugin):
 
 {validation}
 
-Once this is deployed,"""
+Before continuing, verify the record is deployed.
+
+The certificate authority will then be able to check the validity of your domain."""
```
bmw: I'm sorry I didn't notice this before and that I'm continuing to nitpick this, but this sentence still isn't quite right in my opinion. The CA isn't checking the validity of your domain in the sense that the domain name itself is valid or invalid. The name provided by the user may be an acceptable and recognized domain, but the CA is checking that they have control over the domain by verifying they were able to configure a DNS record as requested. Here's two links for seeing how this is normally described in this context:

* https://tools.ietf.org/html/draft-ietf-acme-acme-06#section-8
* https://en.wikipedia.org/wiki/Domain-validated_certificate

I hope this makes sense and feel free to either update this sentence or delete it entirely. After that, I'll merge this PR!

anna-liao: Hi Brad, that makes sense.  thanks for clarifying!

jsha:
Yay, thanks for fixing this!

PR #4706, opened 2017-05-22
Clean up Sphinx errors

tephyr:
Also minor copyedits, suppress unhelpful errors, mark files that should *not* be in the TOC as hidden.


tephyr:
@zjs Fixed line length problem; tox results are clean.

Issue #4244, opened 2017-02-22
Get static typing (using mypy?)


ohemorange:

ohemorange:
http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/

ynasser:
Since we need to support Python 2.6, we can only use the comment-style type annotations mentioned in the blog post above (i.e. the ones from the `typing` module). The type-hints-in-function-definitions syntax (such as `def my_function(numbers: List[int]) -> str:` is only available in Python 3.4+.

As for where to start annotating, the blog post above suggests that changes can be gradual and suggests starting with the "core library" first (see here for details: http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/#fullyannotatingalargecodebase). For certbot, this would probably mean starting with:
- `certbot/main.py`
- `certbot/configuration.py` (because `config`, which is an instance of `NamespaceConfig`, is passed around everywhere) 

Also, the blog post suggests standardizing how you're running mypy and sharing that information, as that will allow multiple people to work on this more smoothly!

dokazaki:
I'll start working on this.

toolness:
Oh, apparently some foundation was actually laid out for this in #4386, but when I run `tox -e mypy` on my system, I get tons of errors. Hmmmm.

bmw:
Glad you're interested in working on this! We're excited about getting mypy type annotations added to Certbot.

None of the core Certbot team has done much experimentation with mypy yet, however, dokazaki tried to work on this at another Python sprint and hit some issues. Basically, the problem is our continued support of Python 2.6. Certbot currently is supporting this old version of Python due to old Red Hat Enterprise Linux systems. Issue #2208 aims to remove our Python 2.6 support while still supporting old Red Hat systems, but there's some nontrivial engineering work to be done to make that happen.

mypy does not support Python 2.6. At the bare minimum, the `typing` module cannot be used. No one has done much additional probing to determine what other problems there are with using mypy on Python 2.6 (if there are any). If you'd still like to work on this knowing this hurdle, you have at least a few options:

1. Don't use `typing` at all. My understanding is this makes it next to impossible to provide meaningful type annotations.
2. When importing `typing`, put a `try` and `except` around each import.
3. Ignore problems with Python 2.6 and work on this in a separate branch. If you take this approach, tests on Travis will fail and we probably won't be able to merge the work into `master` for a few weeks, but it'll allow you to ignore the hurdles that should be going away soon.

[...]

PR #4707, opened 2017-05-22
Fix mypy failures

toolness:
So I started adding mypy support as per #4244, but partway through that I noticed that mypy support was actually added in #4386, it just wasn't mentioned anywhere in #4244.

However, it seems the tox `mypy` testenv isn't actually run as part of the tox `envlist`, nor is it mentioned anywhere in `.travis.yml`... so I think it might not be run under CI at all.

In any case, when I ran `tox -e mypy`, there were a number of breakages.

This PR attempts to fix the breakages.

### DIFF
--- a/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
+++ b/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
@@ -50,7 +50,7 @@ class AuthenticatorTest(test_util.TempDirTestCase, dns_test_common.BaseAuthentic
 
 
 class DigitalOceanClientTest(unittest.TestCase):
-    id = 1
+    _id = 1
     record_prefix = "_acme-challenge"
     record_name = record_prefix + "." + DOMAIN
     record_content = "bar"
@@ -70,7 +70,7 @@ class DigitalOceanClientTest(unittest.TestCase):
 
         domain_mock = mock.MagicMock()
         domain_mock.name = DOMAIN
-        domain_mock.create_new_domain_record.return_value = {'domain_record': {'id': self.id}}
+        domain_mock.create_new_domain_record.return_value = {'domain_record': {'id': self._id}}
 
         self.manager.get_all_domains.return_value = [wrong_domain_mock, domain_mock]
 
--- a/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
+++ b/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
@@ -50,7 +50,7 @@ class AuthenticatorTest(test_util.TempDirTestCase, dns_test_common.BaseAuthentic
 
 
 class DigitalOceanClientTest(unittest.TestCase):
-    _id = 1
+    id = 1  # type: ignore
     record_prefix = "_acme-challenge"
     record_name = record_prefix + "." + DOMAIN
     record_content = "bar"
@@ -70,7 +70,7 @@ class DigitalOceanClientTest(unittest.TestCase):
 
         domain_mock = mock.MagicMock()
         domain_mock.name = DOMAIN
-        domain_mock.create_new_domain_record.return_value = {'domain_record': {'id': self._id}}
+        domain_mock.create_new_domain_record.return_value = {'domain_record': {'id': self.id}}
 
         self.manager.get_all_domains.return_value = [wrong_domain_mock, domain_mock]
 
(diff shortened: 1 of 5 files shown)


certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
```diff
@@ -50,7 +50,7 @@ def test_cleanup(self):
 
 
 class DigitalOceanClientTest(unittest.TestCase):
-    id = 1
+    _id = 1
```
toolness: I *think* mypy didn't like this because we were re-assigning python's built-in `id` function, but I'm not sure. In any case, renaming this to `_id` seems to make mypy happy and not break the test.

certbot/plugins/disco.py
```diff
@@ -13,11 +13,11 @@
...
 
 try:
-    from collections import OrderedDict
+    from collections import OrderedDict  # type: ignore
 except ImportError:  # pragma: no cover
     # OrderedDict was added in Python 2.7
-    from ordereddict import OrderedDict  # pylint: disable=import-error
-
+    from ordereddict import OrderedDict as _OrderedDict  # pylint: disable=import-error
+    OrderedDict = _OrderedDict  # type: ignore
```
toolness: This is a *really* annoying way to work around https://github.com/python/mypy/issues/1393.  Ideally I wouldn't have to import the symbol as `_OrderedDict` on line 19, but I couldn't figure out how to comment that line with *both* a `# pylint:` *and* a `# type:` comment, so I had to do it in two lines. Bluggh.

tox.ini
```diff
@@ -115,7 +115,7 @@ commands =
 [testenv:mypy]
 basepython = python3.4
 commands =
-    {[base]pip_install} mypy
+    {[base]pip_install} mypy==0.501
```
toolness: I think part of the reason for the breakage was due to new, backwards-incompatible versions of mypy being released after #4386 landed. This pins the version to the one that was the latest as of the time of that PR's merging, [0.501](http://mypy-lang.blogspot.com/2017/03/mypy-0501-released.html).  (That said, we should definitely move to the latest version ASAP!)

toolness:
Oh, I should also note that one thing this PR does *not* do is add mypy testing to `.travis.yml` or the default `testenv` list in `tox.ini`... I think it *should* be added to travis jobs, I'm just not sure what the best way of doing that is.

certbot/plugins/disco.py
bmw: If putting both the `pylint: ...` and `type: ...` on the same line doesn't work for whatever reason (including excess line length), you should be able to just move the `pylint:` message to the line before `from ordereddict import OrderedDict`. This disables that warning for the entire block, but because the `import` whose error we want to ignore is the only statement in the block, this shouldn't be a problem.

certbot/util.py
```diff
@@ -23,11 +23,11 @@
 from certbot import lock
 
 try:
-    from collections import OrderedDict
+    from collections import OrderedDict  # type: ignore
 except ImportError:  # pragma: no cover
     # OrderedDict was added in Python 2.7
-    from ordereddict import OrderedDict  # pylint: disable=import-error
-
+    from ordereddict import OrderedDict  as _OrderedDict # pylint: disable=import-error
```
bmw: I have the same comment as I put in `certbot/plugins/disco.py` about what I think is a slightly cleaner way to handle this.

bmw:
> Oh, I should also note that one thing this PR does not do is add mypy testing to .travis.yml or the default testenv list in tox.ini... I think it should be added to travis jobs, I'm just not sure what the best way of doing that is.

My only hesitation with adding `mypy` to Travis is I'm not sure how much it's really helping us when we have so few type annotations (and the ones we have are almost exclusively `type: ignore`). This may largely be due to our unfamiliarity with mypy and if you think it's still doing a lot of work for us or can link me to documentation showing what `mypy` does without any type annotations, I'd be open to adding it. Doing this should be as simple as adding lines like:
```
- python: "3.4"
  env: TOXENV=mypy
```
in `.travis.yml` where our other jobs are specified.

As for adding it to the default `testenv`, we can do this if we stop specifying `basepython=python3.4` so it can be more flexible and work with the versions of Python likely to be on a user's system. This could be done by changing it to `python3` or `python2.7`.

bmw:
@toolness, I just wanted to check in as it has been a month since we've seen any activity here. Are you still planning on working on this or would you like someone else to take over your work here?

bmw:
I'm closing these PRs for now, but I've referenced them in the mypy issue so people can build on them in the future.

PR #4710, opened 2017-05-22
Add more useful usage instructions for all subcommands

### LINKED ISSUE #3875: Provide custom help headers for important subcommands
Currently for instance if you do `certbot --help revoke` you get:

```
usage: 

cert. 

optional arguments:
                        and ~/.config/letsencrypt/cli.ini)

revoke:
  Options for revocation of certs

  --cert-path CERT_PATH
                        Path to where cert is saved (with auth --csr),
                        installed from or revoked. (default: None)
  --key-path KEY_PATH   Path to private key for cert installation or
                        revocation (if account key is missing) (default: None)
```

The bottom of that output is good, but the top could be better.  We'd list a usage string like:
```
certbot revoke --cert-path CERT_PATH [--key-path KEY_PATH]
```
And a summary that was relevant.

We'd probably want to do this with a lookup table of usage headers.

tephyr:
Fixes #3875.


certbot/cli.py
```diff
@@ -366,33 +367,40 @@ def _get_help_string(self, action):
...
         "short": "Register for account with Let's Encrypt / other ACME server",
-        "opts": "Options for account registration & modification"
+        "opts": "Options for account registration & modification",
+        "usage": "\n\n  certbot register --email user@example.com [options]"
     }),
     ("unregister", {
         "short": "Irrevocably deactivate your account",
-        "opts": "Options for account deactivation."
+        "opts": "Options for account deactivation.",
+        "usage": "\n\n  certbot unregister --email user@example.com [options]"
```
bmw: Although it probably should work, using `--email` here has no effect. Instead, I think this should just be `certbot unregister [options]`.

certbot/cli.py
```diff
@@ -366,33 +367,40 @@ def _get_help_string(self, action):
...
     }),
     ("install", {
         "short": "Install an arbitrary cert in a server",
-        "opts": "Options for modifying how a cert is deployed"
+        "opts": "Options for modifying how a cert is deployed",
+        "usage": "\n\n  certbot install --cert-path /path/to/fullchain.pem [options]"
```
bmw: You also need a `--key-path` here.

certbot/cli.py
```diff
@@ -366,33 +367,40 @@ def _get_help_string(self, action):
...
         "short": "Show changes that Certbot has made to server configurations",
-        "opts": "Options for controlling which changes are displayed"
+        "opts": "Options for controlling which changes are displayed",
+        "usage": "\n\n  certbot config_changes --num NUM [options]"
     }),
     ("rollback", {
         "short": "Roll back server conf changes made during cert installation",
-        "opts": "Options for rolling back server configuration changes"
+        "opts": "Options for rolling back server configuration changes",
+        "usage": "\n\n  certbot rollback [options]"
```
bmw: I think it's a good idea to include `--checkpoints` here. This flag allows the user to specify the number of checkpoints Certbot should rollback instead of just simply rolling back a single checkpoint.

tephyr:
@bmw Updated code to incorporate your requested changes; all checks have passed.

certbot/cli.py
```diff
@@ -357,7 +357,8 @@ def _get_help_string(self, action):
     }),
     ("delete", {
         "short": "Clean up all files related to a certificate",
-        "opts": "Options for deleting a certificate"
+        "opts": "Options for deleting a certificate",
+        "usage": "\n\n  certbot delete --cert-name CERTNAME"
```
zjs: I'm not sure of the reasoning, but it seems that the existing `usage` lines begin _and_ end with `\n\n`. Perhaps the new ones should as well?

bmw: Nice catch. Testing with the versions of argparse I have locally, it doesn't seem to matter, but this is an undocumented implementation detail of `argparse` so I agree that we should be consistent here.

Issue #2516, opened 2016-02-20
no response "Installing Python packages"


joviqiao:
![20160220203209](https://cloud.githubusercontent.com/assets/13116366/13196841/65098e24-d815-11e5-9d67-8e34bf8e9d6c.png)
stop here, no response.
What's wrong?

bmw:
One thing to try is running `letsencrypt-auto` with `--verbose` which will provide you with more output to help identify the problem.

LzIManD:
long long long long time,then
```
Had a problem while downloading and verifying Python packages:
Downloading argparse==1.4.0 (23K)...
Downloading cffi==1.4.2 (365K)...
Downloading ConfigArgParse==0.10.0 (36K)...
Downloading configobj==5.0.6 (33K)...
Downloading cryptography==1.1.2 (349K)...
Downloading enum34==1.1.2 (46K)...
Downloading funcsigs==0.4 (14K)...
Downloading idna==2.0 (61K)...
Downloading ipaddress==1.0.16 (17K)...
Downloading linecache2==1.0.0 (12K)...
Downloading ndg-httpsclient==0.4.0 (24K)...
Downloading ordereddict==1.1 (2K)...
Downloading parsedatetime==1.5 (50K)...
Downloading pbr==1.8.1 (89K)...
Downloading psutil==3.3.0 (261K)...
Downloading pyasn1==0.1.9 (36K)...
Downloading pycparser==2.14 (223K)...
Downloading pyOpenSSL==0.15.1 (102K)...
Downloading pyRFC3339==1.0 (8K)...
Downloading python-augeas==0.5.0 (90K)...
Downloading python2-pythondialog==3.3.0 (1831K)...
Downloading pytz==2015.7 (476K)...
Downloading requests==2.9.1 (501K)...
Downloading six==1.10.0 (10K)...
Downloading traceback2==1.4.0 (16K)...
Downloading unittest2==1.1.0 (96K)...
Downloading Werkzeug==0.11.3 (305K)...
Downloading zope.component==4.2.2 (546K)...
Downloading zope.event==4.1.0 (476K)...
Downloading zope.interface==4.1.3 (141K)...
Downloading acme==0.4.0 (95K)...
Downloading letsencrypt==0.4.0 (197K)...
Downloading letsencrypt-apache==0.4.0 (93K)...
Downloading mock==1.0.1 (818K)...

[...]

PR #4711, opened 2017-05-22
Check DNS resolution of pypi before running pipstrap.

aaroncohen:
Change to check if pypi.python.org can be resolved in pipstrap.py. If it  cannot, give guidance to change your pypi index url. 

Looks like this: 
```Unable to resolve DNS name pypi.python.org. You may be experiencing internet connectivity problems, or it may be blocked due to regional filtering. If you suspect that to be the case, you may need to set a pip mirror in ~/.pip/pip.conf. See GitHub issue https://github.com/certbot/certbot/issues/2516```

### DIFF
--- a/letsencrypt-auto-source/pieces/pipstrap.py
+++ b/letsencrypt-auto-source/pieces/pipstrap.py
@@ -21,6 +21,8 @@ anything goes wrong, it will exit with a non-zero status code.
 # The above copyright notice and this permission notice shall be included in
 # all copies or substantial portions of the Software.
 from __future__ import print_function
+
+import socket
 from hashlib import sha256
 from os.path import join
 from pipes import quote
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: letsencrypt-auto-source/pieces/pipstrap.py [hunk 1]
```diff
@@ -123,7 +125,24 @@ def read_chunks(response, chunk_size):
...
+        return bool(socket.gethostbyname('pypi.python.org'))
+    except socket.error:
+        return False
+
+
 def main():
+    if not pypi_available():
+        print("Unable to resolve DNS name pypi.python.org. You may be experiencing internet connectivity problems, or "
+              "it may be blocked due to regional filtering. If you suspect that to be the case, you may need to set a "
+              "pip mirror in ~/.pip/pip.conf. See GitHub issue https://github.com/certbot/certbot/issues/2516")
```
zjs: I think it might be nice if this referenced some sort of documentation URL instead of just a GitHub issue. I think that would better allow for evolving content over time and providing clear, concise information to users who receive this error message.

The over-engineering-inclined part of my brain likes the idea of including the pipstrap version as a part of the URL (perhaps as a fragment or query parameter). For now it could be unused, but including it now allows for easy customization of the help page based on version if that's ever necessary (e.g., if we ever decide to improve the user experience here).

In line with @bmw's comment on the associated issue (https://github.com/certbot/certbot/issues/2516#issuecomment-303238072), I think the it might be best if the documentation were associated with pipstrap itself and not Certbot, since that project is the best place provide this documentation. Perhaps it could be added to the pipstrap repo as a `gh-page`. (Although I'd defer to @erikrose on that.)

erikrose:
I'm happy to entertain a change like this to pipstrap, but there's a bit more to do: right now, pypi.python.org is hard-coded in a couple of URLs inside pipstrap: part of the security of the tool is that it doesn't run all the complex, unpredictable code paths of pip itself. So we'll need to design a way to not only, as you say, persist the preference but use it as well.

bmw:
I'm closing this in favor of Erik's work on the [pipstrap repo](https://github.com/erikrose/pipstrap). Regardless, thanks for the PR!

PR #4720, opened 2017-05-23
Refactor nginx file update mechanism in preparation for working with apache plugin

ohemorange:
See commit titles for explanations.

### DIFF
--- a/certbot/plugins/common.py
+++ b/certbot/plugins/common.py
@@ -268,8 +268,10 @@ class TLSSNI01(object):
 # c.f. #383)
 
 def install_ssl_options_conf(options_ssl, options_ssl_digest, mod_ssl_conf_src,
-    current_ssl_options_hash, previous_ssl_options_hashes, logger): # pragma: no cover
+    previous_ssl_options_hashes, logger): # pragma: no cover
     """Copy Certbot's SSL options file into the system's config dir if required."""
+    current_ssl_options_hash = crypto_util.sha256sum(mod_ssl_conf_src)
+
     def _write_current_hash():
         with open(options_ssl_digest, "w") as f:
             f.write(current_ssl_options_hash)
--- a/certbot/plugins/common.py
+++ b/certbot/plugins/common.py
@@ -268,7 +268,7 @@ class TLSSNI01(object):
 # c.f. #383)
 
 def install_ssl_options_conf(options_ssl, options_ssl_digest, mod_ssl_conf_src,
-    current_ssl_options_hash, previous_ssl_options_hashes, logger):
+    current_ssl_options_hash, previous_ssl_options_hashes, logger): # pragma: no cover
     """Copy Certbot's SSL options file into the system's config dir if required."""
     def _write_current_hash():
         with open(options_ssl_digest, "w") as f:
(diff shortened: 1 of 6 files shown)


certbot/plugins/common.py
```diff
@@ -262,15 +263,49 @@ def _setup_challenge_cert(self, achall, cert_key=None):
...
 # "pragma: no cover") TODO: this might quickly lead to dead code (also
 # c.f. #383)
 
-def setup_ssl_options(config_dir, src, dest):  # pragma: no cover
-    """Move the ssl_options into position and return the path."""
-    option_path = os.path.join(config_dir, dest)
-    shutil.copyfile(src, option_path)
-    return option_path
+def install_ssl_options_conf(options_ssl, options_ssl_digest, mod_ssl_conf_src,
+    all_ssl_options_hashes, logger): # pragma: no cover
```
bmw: Instead of passing in the logger, one should just be created for this file by putting:
```
```
at the top of the file. The only thing this affects is the string saying where the logging message was generated from in our file logs.

certbot/plugins/common.py
```diff
@@ -262,15 +263,49 @@ def _setup_challenge_cert(self, achall, cert_key=None):
         return response
 
 
-# test utils used by certbot_apache/certbot_nginx (hence
```
bmw: Rather than just changing this to document production code also not having test coverage, I think it'd be good to write simple unit tests here like was done for the classes above.

certbot/plugins/common.py
```diff
@@ -262,15 +263,49 @@ def _setup_challenge_cert(self, achall, cert_key=None):
...
 # c.f. #383)
 
-def setup_ssl_options(config_dir, src, dest):  # pragma: no cover
-    """Move the ssl_options into position and return the path."""
-    option_path = os.path.join(config_dir, dest)
-    shutil.copyfile(src, option_path)
-    return option_path
+def install_ssl_options_conf(options_ssl, options_ssl_digest, mod_ssl_conf_src,
+    all_ssl_options_hashes, logger): # pragma: no cover
+    """Copy Certbot's SSL options file into the system's config dir if required."""
```
bmw: Since this module is [documented publicly](https://certbot.eff.org/docs/api/plugins/common.html) and other plugin authors may use it, it'd be good to document the parameters of this function.

certbot/plugins/common.py
```diff
@@ -262,17 +266,57 @@ def _setup_challenge_cert(self, achall, cert_key=None):
...
+    # Check to make sure options-ssl.conf is installed
+    if not os.path.isfile(options_ssl):
+        _install_current_file()
+        return
+    # there's already a file there. if it's up to date, do nothing. if it's not but
+    # it matches a known file hash, we can update it.
+    # otherwise, print a warning once per new version.
+    active_file_digest = crypto_util.sha256sum(options_ssl)
+    if active_file_digest == current_ssl_options_hash: # already up to date
+        return
```
bmw: This line doesn't have test coverage (according to our tests).

certbot/plugins/common.py
```diff
@@ -262,17 +266,57 @@ def _setup_challenge_cert(self, achall, cert_key=None):
...
+    elif active_file_digest in all_ssl_options_hashes: # safe to update
+        _install_current_file()
```
bmw: This line doesn't have test coverage (according to our tests).

PR #4721, opened 2017-05-23
Re-organize certbot-auto template code

birdsarah:
This should be a no-op change that re-organizes the templating code for certbot-auto.

It:
1 - moves more pieces out of letsencrypt-auto.template into sub-files
2 - updates build.py so it can handle nesting of templates
3 - makes phase1 and phase2 functions

This makes the template shorter and, I hope, easier to parse.

But the goal is to make it easy to do a follow-on PR that adds a branch for supporting windows.

Note that:
1 - the tests haven't been changed 
2 - the built `letsencrypt-auto` should not have meaningfully change - some code has moved around slightly, but that's all


letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -649,14 +652,8 @@ Bootstrap() {
...
 
-TempDir() {
-  mktemp -d 2>/dev/null || mktemp -d -t 'le'  # Linux || macOS
-}
-
-
-
 if [ "$1" = "--le-auto-phase2" ]; then
-  # Phase 2: Create venv, install LE, and run.
+    # Phase 2: Create venv, install LE, and run.
```
bmw: So it's not a big deal, but I think it'd be nice if we could handle indentation a little more gracefully. Previously, the only files that were included by the template were definitions of shell script functions or contents of files to be written to disk (or the version number). This is the first time that the template system is being used to include blocks of code that should be indented relative to the template. As a side effect of the way it currently works and the way the code was initially written, we have slightly off indentation like this and the need to indent the majority of entire files like we see in the `phase-?/main.sh` files.

I'm open to any ideas you have to solve this, but the current approach I have in mind is to stick with the previous approach of when including shell script code, only include function definitions. The way this would work is the current places in `letsencrypt-auto.template` that these files are included become a call to the function defined in that file rather than something to be expanded by `build.py`. Then the `main.sh` files would become standalone functions that could be included by the template anywhere before they are used.

Hopefully this makes sense and let me know if you have any better ideas or questions about what I'm proposing here!

letsencrypt-auto-source/pieces/bootstrap/main.sh
```diff
@@ -0,0 +1,81 @@
...
+    ExperimentalBootstrap "Joyent SmartOS Zone" BootstrapSmartOS
+  else
+    error "Sorry, I don't know how to bootstrap Certbot on your operating system!"
+    error
+    error "You will need to install OS dependencies, configure virtualenv, and run pip install manually."
+    error "Please see https://letsencrypt.readthedocs.org/en/latest/contributing.html#prerequisites"
+    error "for more info."
+    exit 1
+  fi
+}
```
bmw: nit: GitHub's pointing out that there's no newline at the end of the file here. There are a few other files in this PR like this too. Since this is often added automatically by people's editors, it might be a good idea to add them now so there aren't somewhat spurious changes to these files and `letsencrypt-auto` later.

bmw:
@birdsarah, I just wanted to check in since it's been a month since we've seen any activity from you here. Are you planning on addressing review comments yourself or would you like someone else to build on your work and get this merged?

birdsarah:
I dropped the ball here. How does this sound: 
- I will plan to pick this up this weekend.
- If it doesn't happen. I'm happy to hand it off.

Does that sound reasonable?

bmw:
Yep! That's totally reasonable. There's no rush or pressure, but I wanted to check in with you on what your plans were.

sydneyli:
@bmw Should we try to pick this up and get this merged? If so, I'll add it to the next milestone.

bmw:
Especially with the future of `certbot-auto` being up in the air as we start rethinking our approach to packaging, I don't think spending time on this ourselves is a good idea right now.

What I'd recommend for this and other PRs that have significant work put into them and we've shown interest in accepting but are waiting on additional code changes is:

1. Move the contents of the PR into a branch on the `certbot` repo.
2. Make sure there is an issue tracking the problem and reference the certbot branch and/or PR.
3. Close the PR.

bmw:
The code in this PR has been added to https://github.com/certbot/certbot/tree/birdsarah/template_out_phases.

Issue #4319, opened 2017-03-10
Rewrite our various `if os.path.exists`s atomically.


ohemorange:

bobstrecansky:
Working on this at PyCon 2017.

bobstrecansky:
@ohemorange - wanted to make sure this is what you want before I go on with the rest:
https://github.com/certbot/certbot/pull/4730

chefscott:
What should happen if the file exists? Throw an error?

ohemorange:
@chefscott The behavior should be the same, but with atomicity. So if the code block previously raised an error, it should be rewritten to continue to raise that error. There's further explanation at #4730 if you're looking to take this on.

bobstrecansky:
Picking this back up; my apologies on the delay; was working through some other tickets first.

fr0zenbits:
Is this still a valid task?  #4730 was closed yesterday.

sydneyli:
@fr0zenbits Yes, this task is still available. The contributor who was working on #4730 didn't get the chance to finish it. You're welcome to pick it back up if you like, and there's a lot of relevant context/comments on that PR thread. Let us know if you have any more questions!

camservo:
Looking for a first issue to help out on.  Was there a consensus on how these should be rewritten?

ohemorange:
@camservo see discussion here: https://github.com/certbot/certbot/pull/4730#issuecomment-303933896

ohemorange:
This is still important, and a good choice for people who want to contribute who don't have extensive knowledge of our codebase! Some information about how to implement this is available here: https://stackoverflow.com/questions/2333872/atomic-writing-to-file-with-python

martinraag:
Looking at this as a potential first issue. For me, the description leaves a lot to interpretation and is perhaps a bit confusing. Judging by a couple of misguided attempts at solving this, I'm not alone.

As I gather, what is required are atomic versions of operations that include `os.path.exists`, to avoid potential race conditions. Looking at the source code, these vary from a simple rename referenced in [#4730 (comment)](https://github.com/certbot/certbot/pull/4730#issuecomment-303933896), to creating a single directory, creating multiple directories, opening and writing to a file, copying permissions, deleting a file, creating symlinks and launching Docker in a subprocess.

For someone new to the codebase, or at least for me, it would be useful to have more context about which operations and modules are potentially problematic. Also, whether we're trying to protect against race conditions between multiple threads, or multiple processes of certbot.

[...]

PR #4730, opened 2017-05-24
Adding Atomic Hooks - Fixing Spacing

bobstrecansky:
Refactored after IRC talk.
@ohemorange - Just wanted to do one of these before I went though and fixed all of these to be atomic.  Is this what you were looking for?  The test suite passes, so i'd think this is ok, but i'll defer to you.

### DIFF
--- a/certbot/hooks.py
+++ b/certbot/hooks.py
@@ -46,7 +46,8 @@ def validate_hook(shell_cmd, hook_name):
         cmd = shell_cmd.split(None, 1)[0]
         if not _prog(cmd):
             path = os.environ["PATH"]
-            if os.path.exists(cmd):
+            validate_path = util.os_path_exists(path)
+            if validate_path:
                 msg = "{1}-hook command {0} exists, but is not executable.".format(cmd, hook_name)
             else:
                 msg = "Unable to find {2}-hook command {0} in the PATH.\n(PATH is {1})".format(
@@ -135,4 +136,5 @@ def execute(shell_cmd):
     if err:
         logger.error('Error output from %s:\n%s', base_cmd, err)
     return (err, out)
+
 
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -44,7 +44,8 @@ def renewal_conf_files(config):
 def renewal_file_for_certname(config, certname):
     """Return /path/to/certname.conf in the renewal conf directory"""
     path = os.path.join(config.renewal_configs_dir, "{0}.conf".format(certname))
-    if not os.path.exists(path):
+    valid_path = util.os_path_exists(path)
+    if valid_path:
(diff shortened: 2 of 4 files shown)


bobstrecansky:
I think this is what you want @ohemorange - can you take a look when you have a minute?

zjs:
I'm not @ohemorange, but I think the goal of #4319 is to ensure that operations involving `os.path.exists` work as expected even when other operations are happening concurrently.

Consider this excerpt from `storage.py`:

```135) def rename_renewal_config(prev_name, new_name, cli_config):
136)     """Renames cli_config.certname's config to cli_config.new_certname.
137) 
138)     :param .NamespaceConfig cli_config: parsed command line
139)         arguments
140)     """
141)     prev_filename = renewal_filename_for_lineagename(cli_config, prev_name)
142)     new_filename = renewal_filename_for_lineagename(cli_config, new_name)
143)     if os.path.exists(new_filename):
144)         raise errors.ConfigurationError("The new certificate name "
145)             "is already in use.")
146)     try:
147)         os.rename(prev_filename, new_filename)
148)     except OSError:
149)         raise errors.ConfigurationError("Please specify a valid filename "
150)             "for the new certificate name.")
```

Currently, we check whether the `new_filename` is in use on line 143 before renaming a file to that name on line 147. However, some concurrent operation could create a file with `new_filename` between when `os.path.exists` is evaluated on line 143 and when `os.rename` is evaluated on line 147, causing us to overwrite a file ― the very thing we're trying to protect against on line 143.

On newer versions of Python, I think `os.replace` might be helpful here. However, I'm not sure what the best practice would be for the set of versions we support.

ohemorange:
Yup, that's exactly right.

@bobstrecansky you may want to check out https://blogs.msdn.microsoft.com/pythonengineering/2016/06/29/idiomatic-python-eafp-versus-lbyl/ for further reading about this

@zjs the issue is that `os.replace` and `os.rename` will both happily rename `prev_filename` to `new_filename`, even if `new_filename` already exists.

zjs:
> @zjs the issue is that `os.replace` and `os.rename` will both happily rename `prev_filename` to `new_filename`, even if `new_filename` already exists.

You're right; I mis-remembered and thought `os.replace` raised an error if the target existed.

bobstrecansky:
Thanks for the pointers @zjs and @ohemorange - i'll get to this PR later on this weekend / next week.

bmw:
@bobstrecansky, I just wanted to check in since it's been over a month since we've seen any activity from you here. Are you still planning on working on this?

bobstrecansky:
@bmw - yup, was just focusing on my other PR before getting to this one.

bobstrecansky:
Picking this back up; apologies for the delay.

bobstrecansky:
Sorry @ohemorange - closed!

Issue #4688, opened 2017-05-18
Add certbot-route53 to our tree


bmw:
After talking to the maintainers we're going to make https://github.com/lifeonmarspt/certbot-route53 an official plugin and move it into our repo. Things we want with this transition are:

* Git history is preserved
* We have a plugin under the name certbot-dns-route53 in our tree
* We have a plugin or at least something to upload to the name certbot-route53 on PyPI so users under that name continue to get updates as the plugin is further developed. In theory, this could be the same package as `certbot-dns-route53` but we'd want to make sure the user experience is sane if both are installed.
* The certbot-route53 package should continue to work with the plugin name `certbot-route53:auth` due to the way we name 3rd party plugins (see #4351). Whether or not it also works with `certbot-route53` is an implementation detail in my opinion.

I don't really have a preference for how this is implemented. My instinct is it'll be easiest to make `certbot-route53` a shim package around `certbot-dns-route53`. We did something similar for the `letsencrypt` package in our 0.7.0 release. This can be left up to whoever takes this ticket though.

@zjs, if you have an interest in working on this, that'd be useful, otherwise, I can take it when I get some spare cycles.

bmw:
@zjs, what's the plan for uploading these packages to PyPI? Make `certbot-route53` a shim package around `certbot-dns-route53`? If so, do you have time to prepare a simple shim package? You can see an example of this at https://github.com/certbot/certbot/tree/v0.7.0/letsencrypt-apache.

zjs:
There's a change ready to go for that here: https://github.com/lifeonmarspt/certbot-route53/pull/5

bmw:
Oh wow. Thanks for taking care of that. That's super helpful.

bmw:
With lifeonmarspt/certbot-route53#5, I think this has been resolved. Thanks Zach!

PR #4731, opened 2017-05-24
route53: update setup.py to follow repo conventions

zjs:
This change updates the setup script for the route53 plugin to more closely match conventions from other packages in the repository.

Notable changes:
 * The version number is bumped to match the rest of Certbot.
 * The package now requires a matching version of ACME and core Certbot.
 * Contact information is updated.
 * Additional versions of Python are listed.

Part of #4688.

### DIFF
--- a/certbot-route53/setup.py
+++ b/certbot-route53/setup.py
@@ -20,7 +20,6 @@ setup(
     name='certbot-route53',
     version=version,
     description="Route53 DNS Authenticator plugin for Certbot",
-    url='https://github.com/lifeonmarspt/certbot-route53',
     url='https://github.com/certbot/certbot',
     author="Certbot Project",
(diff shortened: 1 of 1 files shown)


certbot-route53/setup.py
```diff
@@ -3,23 +3,28 @@
...
+    'setuptools>=1.0',
+    'zope.interface',
 ]
 
 setup(
     name='certbot-route53',
     version=version,
-    description="Route53 plugin for certbot",
+    description="Route53 DNS Authenticator plugin for Certbot",
     url='https://github.com/lifeonmarspt/certbot-route53',
```
bmw: I think this line should be deleted. I'm assuming this was an accident, but if you did it on purpose to try and give a nod to the original maintainers, I think we could do that in a different way.

Issue #4535, opened 2017-04-21
Unwrap "max retries exceeded" errors


jsha:
When there is some sort of network problem, or the server is unavailable, we get an error like:

> An unexpected error occurred: ConnectionError: HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org', port=443): Max retries exceeded with url: /directory (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f020eee70d0>: Failed to establish a new connection: [Errno -2] Name or service not known',))

This is confusing because it emphasizes the "retries" part, when the important part is the "Name or service not known" part. We should handle this type of error specially and just show the actual error, not the part about exceeding retries.

bobstrecansky:
Working on this one at PyCon2017

bobstrecansky:
Proposed starting to implementation: https://github.com/kennethreitz/requests/pull/4047

bobstrecansky:
Talked to the requests library maintainers; they mentioned that I could take this exception; wrap it as a string, and parse it.  Going to take a stab at that today.

bobstrecansky:
WIP: https://github.com/certbot/certbot/pull/4733

bmw:
@jsha, Noah said you might be interested in working on this?

PR #4733, opened 2017-05-24
[#4535] - Unwrap 'max retries exceeded' errors

bobstrecansky:
This should help out with https://github.com/certbot/certbot/issues/4535
utilizing regex to parse out the relevant error messages and displaying them back out to stdout if we fail to connect to the proper host.  This should pass back the urllib3 responses in a more readable fashion.  

Example test with acme routed to localhost:

```
[12:55:16] bstrecansky ~ tail -1 /etc/hosts
127.0.0.1	acme-v01.api.letsencrypt.org
```

```
venv)[12:52:29] bstrecansky certbot max_retries_exceeded ~ sudo certbot                                                       1 ↵
An unexpected error occurred:
Can't connect to acme-v01.api.letsencrypt.org
[Errno 65] - No route to host
```


bobstrecansky:
@jsha -
The requests maintainers would like for us to change this within our module rather than upstream in the requests module because they believe (rightly so) there are probably many clients that are expecting the multiple values downstream.  

I've added another commit with the regex explanation and examples.  The string that would need to be matched is:

```
HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org', port=443): Max retries exceeded with url: /directory (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x106be06d0>: Failed to establish a new connection: [Errno 65] No route to host',))
```
So I figured we could just have individual descriptions and examples (I felt that kept the code cleaner while giving a useful example).  Happy to include the whole string if you deem it necessary.

We are only not printing the traceback if it matches this particular case.  The associated else prints a traceback in all other cases.  I'll be happy to change that logic if you'd like as well.

jsha:
Great, thanks for the update! Does that mean https://github.com/kennethreitz/requests/issues/2876 should be closed?

One more question on this implementation: Why regexes rather than catching typed exceptions and interrogating their contents?

bobstrecansky:
@jsha - that was the feeling I got when I discussed the topic with the requests maintainers at PyCon.   We could catch the typed exceptions and interrogate their contents; the regexes just felt like a cleaner solution to me (I have a systems background, so always happy to do some regex'ing).  I'll be happy to change that to the typed exceptions method if you'd like me to.

bobstrecansky:
@jsha - just wanted to bump this.

certbot/log.py
```diff
@@ -299,7 +300,23 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
             _, _, exc_str = str(exc_value).partition(':: ')
             print(exc_str, file=sys.stderr)
         else:
-            traceback.print_exception(exc_type, exc_value, None)
+            """ Regex for parsing error messages"""
+            """ Regular expression groups"""
+            """ Match Group #1: Attempted connection host"""
+            """ Match Group #2: urllib3 Error Number"""
+            """ Match Group #3: urllib3 Error String"""
+            """ Example:"""
```
jsha: For the example, I was thinking a full example error message before parsing.

Also, triple quotes (`"""`) are generally used for multi-line comments in Python, so open with a triple quote at the top, then close with another triple quote at the bottom, rather than repeating the quotes on each line.

certbot/log.py
```diff
@@ -299,7 +300,23 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
...
+            """ Match Group #1: acme-v01.api.letsencrypt.org"""
+            """ Match Group #2: [Errno 65]"""
+            """ Match Group #3: No route to host"""
+            err_re = re.search(r"host='(\S*)'.*(\[Errno \d+\]\ )(.*)(',\)\))", str(exc_value))
```
jsha: You don't need to put the `.*` in a separate group, you can leave it combined with the Errno. Also, `err_re` isn't the right name for this variable, since it's the result of running a search. I'd call it `match` instead.

In this case I'd prefer `re.match` over `re.search`, with more anchoring at the beginning of the error message (e.g. including the bit about RetryError). This minimizes the chance that we accidentally transform an error message we didn't intend to.

Let's also extract the path (e.g. /directory) so we can compose the full URL for the error message.

certbot/log.py
```diff
@@ -299,7 +300,23 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
...
+            if err_re:
+                err_msg = "Can't connect to "+err_re.group(1)+\
```
jsha: Instead of referring by group name, you can do:

(host, error) = match.groups()

Then, when generating the output message, you can do:

new_error = "Problem requesting {1}: {2}".format(host, error)

bobstrecansky: @jsha - what parts of the message do you think are important to capture?  I want to make sure that the error messaging is sensible to the person reading the error message.  I'd propose:

hostname
[urllib error code] - urllib error message


So we would net an output something like so:

```

An unexpected error occurred:
Can't connect to acme-v01.api.letsencrypt.org
[Errno -2] Name or service not known

```

certbot/log.py
bobstrecansky: Made the change on the quotes.  Would you like me to  keep the match group descriptions and examples, or would you just like me to leave the initial string that we're parsing?

bobstrecansky:
@jsha - when you have a chance, I can finish up this PR with your clarifications I believe and move on to the next one in my queue (https://github.com/certbot/certbot/pull/4730)

certbot/log.py
jsha: It's unusual to separate errors onto multiple lines like that. Also, we can make things clearer by combining the host, path, and scheme. E.g.:

> Requesting https://acme-v01.api.letsencrypt.org/directory: [Errno -2] Name or service not known

jsha:
Also, I should have pointed this out earlier, but it would be nicer to put this in a more specific try/catch around where we make network requests, specifically in acme/acme/client.py, method _send_request, surrounding the

`        response = self.session.request(method, url, *args, **kwargs)`

acme/acme/client.py
```diff
@@ -18,7 +18,7 @@
 from acme import jose
 from acme import jws
 from acme import messages
-
+from certbot import log
```
jsha: We don't want to create a dependency from `acme` to `certbot`. That will create a dependency loop, where each depends on the other. Also, we want to keep the dependencies for `acme` relatively small.

Instead of calling `log.except_hook`, my recommendation is to move the matching code you implemented over there into this file. In other words, once you're done you'll wind up with no diffs in `certbot`, only in `acme`. You'll also want to remove the `prints` and raise a new exception with the message you want, e.g. `raise Exception("foo ...")`.

certbot/log.py
```diff
@@ -309,14 +309,11 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
...
-            if err_re:
-                err_msg = "Can't connect to "+err_re.group(1)+\
-                          "\n"+err_re.group(2)+\
-                          "- "+err_re.group(3)
-                print(err_msg, file=sys.stderr)
-            else:
-                traceback.print_exception(exc_type, exc_value, None)
+            print (exc_value, file=sys.stderr)
+            match = re.match(r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])((\s*\w*)*)", str(exc_value))
+            host, directory, error_no, error_msg = (match.group(1), match.group(2), match.group(3), match.group(4))
```
jsha: You can just use `match.groups()` here.

certbot/log.py
```diff
@@ -309,14 +309,11 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
...
             """ Match Group #3: No route to host"""
-            err_re = re.search(r"host='(\S*)'.*(\[Errno \d+\]\ )(.*)(',\)\))", str(exc_value))
-            if err_re:
-                err_msg = "Can't connect to "+err_re.group(1)+\
-                          "\n"+err_re.group(2)+\
-                          "- "+err_re.group(3)
-                print(err_msg, file=sys.stderr)
-            else:
-                traceback.print_exception(exc_type, exc_value, None)
+            print (exc_value, file=sys.stderr)
```
jsha: Please remove the extraneous print.

certbot/log.py
```diff
@@ -309,14 +309,11 @@ def except_hook(exc_type, exc_value, trace, debug, log_path):
...
+            match = re.match(r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])((\s*\w*)*)", str(exc_value))
```
jsha: I'd still like to see a comment with a full example of a real error message that we're matching against.

certbot/log.py
bobstrecansky: Sorry, still working through this.  Will remove.

certbot/log.py
bobstrecansky: I attempted to do that.  Will attempt it again.

acme/acme/client.py
bobstrecansky: sounds good.

certbot/log.py
bobstrecansky: Will add one in.

bobstrecansky:
@jsha - I think this should be all fixed up now, lmk.

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
         kwargs.setdefault('headers', {})
         kwargs['headers'].setdefault('User-Agent', self.user_agent)
         kwargs.setdefault('timeout', self._default_timeout)
-        response = self.session.request(method, url, *args, **kwargs)
+        try:
+            response = self.session.request(method, url, *args, **kwargs)
+        except requests.exceptions.RequestException as e:
+            """ Regex for parsing error messages"""
```
jsha: Instead of terminating each line, please use the ability of Python's triple-quote """ to span multiple lines, and remove the initial space on the first line:

```
"""First line
   Second line
"""
```

Also, good comments should explain "why" rather than "what" or "how." For instance, here we would say something like "The requests library emits exceptions with a lot of extra text. We parse them with a regexp to raise a more readable exceptions."

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            """ Regular expression groups"""
+            """ Match Group #1: Attempted connection host"""
```
jsha: I would leave out these descriptions of each match group. They are implicitly documented by the variable names that you assign below.

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            """ Match Group #2: urllib3 Error Number"""
+            """ Match Group #3: urllib3 Error String"""
+            """ Example:"""
+            """HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org', 
+            port=443): Max retries exceeded with url: /directory 
+            (Caused by NewConnectionError('
+            <requests.packages.urllib3.connection.VerifiedHTTPSConnection
+            object at 0x108356c50>: Failed to establish a new connection:
+            [Errno 65] No route to host',))"""
+            """ Match Group #1: acme-v01.api.letsencrypt.org"""
```
jsha: Similarly, you don't need to provide an example of each matched group. The single example error message, plus the names assigned to the match groups, are enough.

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            """ Match Group #2: /directory"""
+            """ Match Group #3: [Errno 65]"""
+            """ Match Group #4: No route to host"""
+            error_string = str(e.message)
```
jsha: Why does e.message need to be converted to a `str`? What is it naturally?

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            error_regex = r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])([A-Za-z ]*)"
+            host, directory, error_no, error_msg = re.match(error_regex, error_string).groups()
+            new_error = "Requesting {0}{1} : {2} {3}".format(host, directory, error_no, error_msg)
```
jsha: Please remove the space before the :.

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            <requests.packages.urllib3.connection.VerifiedHTTPSConnection
+            object at 0x108356c50>: Failed to establish a new connection:
+            [Errno 65] No route to host',))"""
+            """ Match Group #1: acme-v01.api.letsencrypt.org"""
+            """ Match Group #2: /directory"""
+            """ Match Group #3: [Errno 65]"""
+            """ Match Group #4: No route to host"""
+            error_string = str(e.message)
+            error_regex = r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])([A-Za-z ]*)"
+            host, directory, error_no, error_msg = re.match(error_regex, error_string).groups()
```
jsha: Please replace "directory" with "path", which matches HTTP URL terminology better.

acme/acme/client.py
```diff
@@ -624,7 +624,32 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            new_error = "Requesting {0}{1} : {2} {3}".format(host, directory, error_no, error_msg)
+            print new_error
```
jsha: Please remove this `print`.

acme/acme/client.py
bobstrecansky: Naturally it's a 
<class 'requests.packages.urllib3.exceptions.MaxRetryError'>

Converting it to a string allowed the regex.

acme/acme/client.py
bobstrecansky: We could also use "uri" rather than "path", in case any query strings come into play in the future.  I'll leave it as path for the time being like you're requesting.

bobstrecansky:
@jsha - i've made all the changes you've requested, as well as done some unit test digging.  I am not super versed in python unit tests; can you give me a little more information as to where / what I need to add for the unit test?  I also noticed that the test_requests_error_passthrough unit tests begins to fail with my code add.  Is that my responsibility to update as well?

acme/acme/client_test.py
```diff
@@ -21,7 +21,6 @@
 KEY = jose.JWKRSA.load(test_util.load_vector('rsa512_key.pem'))
 KEY2 = jose.JWKRSA.load(test_util.load_vector('rsa256_key.pem'))
 
-
```
jsha: Looks like you introduced a spurious diff here; could you restore the deleted line?

acme/acme/client_test.py
```diff
@@ -741,6 +740,18 @@ def test_head_get_post_error_passthrough(self):
...
                           self.net.post, 'uri', obj=self.obj)
 
+class ErrorTest(unittest.TestCase):
+    """Tests for urllib3 error conditions"""
+    # pylint: disable=too-many-instance-attributes,too-many-public-methods
+
+    def setUp(self):
+        self.response = mock.MagicMock(
+            ok=True, status_code=http_client.OK, headers={}, links={})
+        self.net = mock.MagicMock()
```
jsha: Mocks are used for the components of a system that are not being tested. In this case we're testing the `ClientNetwork` component, so we don't want to use a mock for it. Instead, you should instantiate a real ClientNetwork object. There should be examples elsewhere in this file.

acme/acme/client_test.py
```diff
@@ -741,6 +740,18 @@ def test_head_get_post_error_passthrough(self):
...
+        self.net.get.return_value = self.response
+
+    def test_urllib_error(self):
+         self.assertEqual(self.response, self.net.get("http://localhost:19123/nonexistent.txt"))
```
jsha: What we want to do here is assert that this get raises an exception. Elsewhere in this file, you'll see `assertRaises`, but that's not quite what we want: we don't want to test just the type of the exception, but also its content. So set up a try/catch block around the `get` call, and in the catch block, verify that the raised exception has the correct message.

bobstrecansky:
Alright, everything seems to be good now, minus the test_requests_error_passthrough mock test.  I think my test can actually replace that one, but i'd like your input on that.

jsha:
The failing test is actually showing you a real bug:

```
  File "/home/travis/build/certbot/certbot/.tox/py27-oldest/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/certbot/certbot/acme/acme/client_test.py", line 622, in test_requests_error_passthrough
    self.net._send_request, 'GET', 'uri')
  File "/opt/python/2.7.13/lib/python2.7/unittest/case.py", line 473, in assertRaises
    callableObj(*args, **kwargs)
  File "/home/travis/build/certbot/certbot/acme/acme/client.py", line 646, in _send_request
    host, path, error_no, error_msg = re.match(error_regex, str(e.message)).groups()
AttributeError: 'NoneType' object has no attribute 'groups'
```

In other words, you need to assign the result of `re.match` to something, then check whether's it's None before calling `.groups()`.

Also , the coverage results suggest that client.py line 629 is not getting exercised in tests: https://travis-ci.org/certbot/certbot/jobs/253236669

bobstrecansky:
Gah what a moron.  Fixing all of these now.

bobstrecansky:
@jsha - I know my tests are failing - just wanted to get your $0.02 before I continue going in the wrong direction here.

acme/acme/client.py
```diff
@@ -624,7 +625,31 @@ def _send_request(self, method, url, *args, **kwargs):
...
+            HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org',
+            port=443): Max retries exceeded with url: /directory
+            (Caused by NewConnectionError('
+            <requests.packages.urllib3.connection.VerifiedHTTPSConnection
+            object at 0x108356c50>: Failed to establish a new connection:
+            [Errno 65] No route to host',))"""
+
+            err_regex = r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])([A-Za-z ]*)"
+            err_str = str(e.message)
+            if err_str is not None:
```
jsha: I don't think you need to check if err_str is None - it should always be present in an exception. I meant you need to check the result of the re.match:

```
m = re.match(err_regex, err_tr)
if m is None:
 raise
else:
  host, path, err_no, err_msg = m.groups()
```

Also, you don't need the `res` object you created here.

bobstrecansky: The only reason I created it was because otherwise the function had too many variables and it didn't pass linting.  Should I not worry about that or should I do something else?

jsha: Our pylint settings are pretty extreme. I would be fine with disable=too-many-locals here. I suspect the underlying concept behind that warning is that at some point the code gets complex enough that you want to break it out into multiple functions. That might be a decent idea here, but I don't think it's required.

jsha:
One thing I forgot to mention on IRC - in order for Travis to build on your own repo, I believe you have to sign in to Travis and enable that repo for builds.

bobstrecansky:
@jsha - I did a lot of work on this today - figured out a couple things that are annoying:  Different versions of python use different requests, which therefore throw back different exceptions and different messages.  How should we handle this?  I also don't want to take code coverage away from acme for an exception error (which it seems as if we would have to do if we had to manage different general error exceptions for different versions of python).

jsha:
That's a good question. @bmw @ohemorange, how do we reach 100% test coverage when some code gets executed only by certain versions of Python? Do we run the "cover" task only for one version?

bmw:
Yes, we only run coverage tests for Python 2.7. I'd recommend adding `pragma: no cover` to lines that aren't run in Python 2.7 like [this](https://github.com/certbot/certbot/blob/a7a8e060e37fad21a9702ac0f143fd39609c8c2d/certbot/log.py#L208).

bobstrecansky:
@bmw - Thanks, that was extremely helpful
@jsha - I actually had to make 2 separate distinctions here - one for the python version, and another for the OS version.  I was able to encapsulate all of this in my tests.

jsha:
Looks like there were still some failures on Travis. :-(

bobstrecansky:
Yup; going to investigate tomorrow - will check against my test branch (passing now)

bobstrecansky:
okay @jsha - was able to get the tests to pass 👍

acme/acme/client_test.py
```diff
@@ -621,6 +622,26 @@ def test_requests_error_passthrough(self, mock_requests):
...
 
+    def test_urllib_error(self):
+        try:
+            # pylint: disable=protected-access
+            self.net._send_request('GET', "http://localhost:19123/nonexistent.txt")
+
+        except ValueError as y:
+            if "linux" in sys.platform:
+                self.assertEqual("Requesting localhost/nonexistent: "
+                                 "[Errno 111] Connection refused", str(y))
```
jsha: I thought we had agreed (via IRC) to leave out the Errno since it is different between platforms and doesn't add useful information.

acme/acme/client.py
```diff
@@ -624,7 +627,31 @@ def _send_request(self, method, url, *args, **kwargs):
         kwargs.setdefault('headers', {})
         kwargs['headers'].setdefault('User-Agent', self.user_agent)
         kwargs.setdefault('timeout', self._default_timeout)
-        response = self.session.request(method, url, *args, **kwargs)
+        try:
+            response = self.session.request(method, url, *args, **kwargs)
+        except requests.exceptions.RequestException as e:
+            #pylint: disable=pointless-string-statement
```
jsha: Local style is to always include a space after the `#` that begins a comment. There are a few other similar comments below that need fixing.

acme/acme/client.py
```diff
@@ -599,6 +600,8 @@ def _check_response(cls, response, content_type=None):
         return response
 
     def _send_request(self, method, url, *args, **kwargs):
+    #pylint: disable=too-many-locals
```
jsha: Please line this comment up with the beginning of the block quote, and remove the extra line below it.

acme/acme/client.py
```diff
@@ -624,7 +627,31 @@ def _send_request(self, method, url, *args, **kwargs):
...
+
+            Example:
+            HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org',
+            port=443): Max retries exceeded with url: /directory
+            (Caused by NewConnectionError('
+            <requests.packages.urllib3.connection.VerifiedHTTPSConnection
+            object at 0x108356c50>: Failed to establish a new connection:
+            [Errno 65] No route to host',))"""
+
+            err_regex = r".*host='(\S*)'.*url\: (\/\w*).*(\[Errno \d+\])([A-Za-z ]*)"
```
jsha: Let's add a little anchoring at the beginning here to make extra sure we don't match the wrong error:

```
err_regex = r"^HTTPSConnectionPool\(host='(\S*)
```

acme/acme/client_test.py
```diff
@@ -621,6 +622,26 @@ def test_requests_error_passthrough(self, mock_requests):
...
+            else: #pragma: no cover
+                self.assertEqual("Requesting localhost/nonexistent: "
+                                 "[Errno 61] Connection refused", str(y))
+
+        except requests.exceptions.ConnectionError as z: #pragma: no cover
```
jsha: It's not clear to me when we would expect to get a ConnectionError, and what we're testing for here. Could you document that here?

acme/acme/client_test.py
bobstrecansky: I thought about it a bit more and it might be helpful for the end user to see the error number - they may have seen that urllib3 error message before.  It's easy enough for me to remove if we decide that we should, maybe we should just decide that together.

acme/acme/client_test.py
bobstrecansky: I added documentation about python2 and python3, as well as linux and OSX.  I'm assuming that's sufficient?

acme/acme/client.py
bobstrecansky: We can't really do that, because that would encapsulate the outer exceptions that we are trying to pull out here.   Open to other suggestions though.

bobstrecansky:
@jsha - all I did was add comments and my PR won't pass unit tests now.  You can see the difference with:
`git diff 4f9dbcd 2cefd01`

bobstrecansky:
@jsha - I think this is ready for your review again.  I've addressed all your concerns either in a change or in a comment, and all the tests are passing again.  TIA!

acme/acme/client_test.py
jsha: Looking at the test makes me more convinced we should leave it out. Splitting up the test based on platform is pretty hacky, especially when we can simplify the error message and the test at the same time without removing useful information.

acme/acme/client_test.py
jsha: It looks like, on python3, hitting this error will print `('Connection aborted.', error(111, 'Connection refused')`, and you've made the test check that behavior. But I think that behavior is wrong. We'd like to make this print the same error on python3 that it prints on other Python versions. I think the best thing to do here is install python3 and figure out why you're not getting the expected results there.

acme/acme/client.py
jsha: I don't understand what you mean by this?

jsha:
By the way, it looks like you are rebasing your commit and pushing it fresh each time. That is good practice in some repos, but what we do here is keep all our commits separate, even if they are small or we aren't sure about them. Once the PR is ready, we merge it with a squash commit, so all those tiny commits get squashed into one. The advantage of this approach is that when you push individual commits I can see the diffs versus the last version of the PR I looked at.

acme/acme/client.py
bobstrecansky: We are unwrapping the wrapped error message with the regex, and I don't want to catch the outer limited regex. I can post examples if you'd like.

bobstrecansky:
Will do a squash & merge from here on out; just what i'm used to doing at work.

acme/acme/client.py
jsha: Yes, please. Some examples would be helpful.

jsha:
> Will do a squash & merge from here on out; just what i'm used to doing at work.

To be clear, what I was saying is that when we merge this, *we'll* do a squash. I'd like you do push unsquashed commits as you go so I can track the incremental changes between pushes better. Thanks!

acme/acme/client.py
bobstrecansky: I forgot that I had already posted one earlier in the PR.  Here it is again:

`HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org',` port=443): Max retries exceeded with url: /directory (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x106be06d0>: Failed to establish a new connection: [Errno 65] No route to host',))`

bobstrecansky: We may be able to anchor with HTTPSConnectionPool; this is the full error message that gets spit back:

`ConnectionError: HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org', port=443): Max retries exceeded with url: /directory (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x109dbb110>: Failed to establish a new connection: [Errno 65] No route to host',))`

jsha: Ah, yep, looks like we could anchor with HTTPSConnectionPool. I'd also like to expand the "url:" in the regex to "Max retries exceeded with url:".

bobstrecansky: If it's okay with you I think it would make sense to just use the .* rather than anchoring it with HTTPSConnectionPool.  If we were to anchor it that way, we'd have to change the exception handling and write more unit tests for it.  We aren't actually looking for that part of the error message, the host is the first important piece, so everything before it is sort of a moot point.  That might also help us in the future too if an error gets nested in more exceptions in the requests library.  I did add the Max retries exceeded with url, but that made my line too long, so I had to ignore it in pylint.

jsha: Ah, I probably wasn't clear enough. It's not important to me that you have a `^` immediately before HTTPSConnectionPool, just that the string be in the regex. That is, I originally proposed:

`err_regex = r"^HTTPSConnectionPool\(host='(\S*)`

But based on the actual message string you postedrecently, it should be:

`err_regex = r"HTTPSConnectionPool\(host='(\S*)`

(that is, remove the `^`). I think that shouldn't require changing exception handling or tests. If that's incorrect, please let me know what it would require changing. Thanks!

bobstrecansky: That value can actually change with the version of python and operating system as well.  Can we leave that part out of the regex match?  The first piece of pertinent data we get back is the host, everything else should be irrelevant for parsing this request.

jsha: Is the string "HTTPSConnectionPool" what changes with different Python or OS versions? Or is it the ordering of the fields within the parens? If it's the latter, it could be fixed with `HTTPSConnectionPool\(.*host`

bobstrecansky: It seems like a couple different things are at play.  Notably the method that requests nests exceptions seems to change here and there.  Sometimes the value gets called HTTPConnecitonPool, sometimes it gets called HTTPSConnection pool, sometimes it gets nested in deeper within the exception string.

jsha: Interesting! Okay, will consider this good enough then. Thanks!

jsha:
I think the only remaining action item is to remove errno from the output and simplify the test accordingly.

bobstrecansky:
Do we want to remove the error number?  I think I mentioned it in one of my rebases that got shoved under the rug - maybe it would be smart to leave it.  Those error messages can be associated with the underlying library and could be helpful for people to see a pattern.  I know I normally like having some sort of error number associated with a failure so I can associate it; that could just be my preference though.

jsha:
Yep, I replied to your comment at https://github.com/certbot/certbot/pull/4733#discussion_r130214385, but it probably got lost :-)

bobstrecansky:
Got it, will change it this weekend then we should be good to go.

bobstrecansky:
@jsha - thinking about this a bit more - will let you know when it's ready for review.

bobstrecansky:
@jsha did a bunch of refactoring to simplify my logic a bit. Happy to discuss if you'd like.

bmw:
@jsha, are you able to give this another look?

jsha:
Thanks for your patience! Looks good to me.

PR #4737, opened 2017-05-25
Fixes #4719

### LINKED ISSUE #4719: Temporal files and dirs not removed from /tmp
Since certbot-auto 0.14.0, every time `certbot-auto renew --quiet` runs on my cron job it left behind a tmp file in /tmp.

Just an example:

```
-rw-------  1 root root  361 May 22 12:17 tmpzdqsnC
-rw-------  1 root root  361 May 22 18:17 tmptUuhXk
-rw-------  1 root root  361 May 23 00:17 tmp1T_Xes
-rw-------  1 root root  361 May 23 06:17 tmp3vaADu
-rw-------  1 root root  361 May 23 12:17 tmpSuRRuU
-rw-------  1 root root  361 May 23 18:17 tmpgtxFOp
-rw-------  1 root root  361 May 24 00:17 tmpZF3Xp3
```
The content of these files are like this:
```
2017-05-22 12:17:01,586:DEBUG:certbot.main:certbot version: 0.14.1
2017-05-22 12:17:01,586:DEBUG:certbot.main:Arguments: ['--version']
2017-05-22 12:17:01,586:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#manual,PluginEntryPoint#nginx,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
```
Also there is a tmp dir,

`drwx------ 2 root root 4096 May 24 00:33 tmp.ccaHF7JIEe/`

Tha only have a file inside:

`-rw-r--r-- 1 root root 4614 May 19 12:28 fetch.py`

As I said, checking the tmp files, it is happening from version 0.14.0 and I think they should be removed automatically.

Here a community buddy asking the same question. https://community.letsencrypt.org/t/certbot-auto-recently-starting-to-leave-files-in-tmp/34789

bmw:
In #4446, we fixed issues with errors and dropped logging messages caused by parts of Certbot code trying to log messages before logging was fully configured. The reason we can't configure logging right away is we need to parse command line arguments to determine how logging should be set up.

The way we fixed this is we buffered messages in memory and once logging was setup, we sent them to our normal file handler. If for some reason we crashed before these file handlers were created, we sent them to a temporary log file as to not lose these messages.

The problem with this was if the user provided an invalid command line argument, `-h/--help`, or `--version` on the command line, Certbot would exit before logging was fully configured, create this temporary file, and not even tell them about it.

Not telling the user about it is a bug, but creating a temporary log file in these instances seems excessive. This is especially a problem for `certbot-auto` which runs `certbot --version` to determine if we needed to upgrade Certbot. This meant that for every run of `certbot-auto` a temporary file was created. This PR fixes the problem.

Note: See #4738 for a problem not fixed by this PR.

### DIFF
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -320,6 +320,10 @@ def pre_arg_parse_except_hook(memory_handler, *args, **kwargs):
     command line argument was invalid or -h, --help, or --version was
     provided on the command line.
 
+    :param MemoryHandler memory_handler: memory handler to flush
+    :param tuple args: args for post_arg_parse_except_hook
+    :param dict kwargs: kwargs for post_arg_parse_except_hook
+
     """
     try:
         post_arg_parse_except_hook(*args, **kwargs)
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -103,6 +103,7 @@ def post_arg_parse_setup(config):
     root_logger.removeHandler(memory_handler)
     temp_handler = memory_handler.target
     memory_handler.setTarget(file_handler)
+    memory_handler.flush(force=True)
     memory_handler.close()
     temp_handler.close()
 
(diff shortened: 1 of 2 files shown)


bmw:
Thanks for the quick review Zach!

PR #4739, opened 2017-05-25
Add regression/system test for #4719

bmw:


tests/boulder-integration.sh
```diff
@@ -80,6 +80,17 @@ CheckHooks() {
     rm "$HOOK_TEST"
 }
 
+# test for regressions of #4719
+get_num_tmp_files() {
+    ls -1 /tmp | wc -l
+}
+num_tmp_files=$(get_num_tmp_files)
+common --version
```
zjs: Maybe also run `--help` here to verify that neither regresses?

PR #4748, opened 2017-05-27
GoDaddy DNS Authenticator

bahadir:
Implement an Authenticator which can fulfill a dns-01 challenge using the Godaddy API. Most of the code copied from DigitalOcean DNS Authenticator.

Testing Done:
- tox -e py27
- tox -e lint
- Manual testing:
  - Used certbot certonly --dns-godaddy -d, specifying a credentials file as a command line argument. Verified that a certificate was successfully obtained without user interaction.
  - Used certbot certonly --dns-godaddy -d, without specifying a credentials file as a command line argument. Verified that the user was prompted and that a certificate was successfully obtained.
  - Used certbot certonly -d. Verified that the user was prompted for a credentials file after selecting Godaddy interactively and that a certificate was successfully obtained.
  - Used certbot renew --force-renewal. Verified that certificates were renewed without user interaction.
- Negative testing:
  - Path to non-existent credentials file.
  - Credentials file with unsafe permissions (644).
  - Credentials file missing token.
  - Credentials file with blank token.
  - Credentials file with incorrect token.
  - Domain name not registered to Godaddy account.
- /tox.cover.sh
  - certbot_dns_godaddy/dns_godaddy.py 99% 54
  - certbot_dns_godaddy/dns_godaddy_test.py 100%


certbot-dns-godaddy/certbot_dns_godaddy/dns_godaddy.py
```diff
@@ -0,0 +1,168 @@
...
+                                and record['name'] == self._compute_record_name(domain, record_name)
+                                and record['data'] == record_content]
+        except godaddypy.client.BadResponse as e:
+            logger.debug('Error getting DNS records using the Godaddy API: %s', e)
+            return
+
+        for record in matching_records:
+            try:
+                logger.debug('Removing TXT record with name: %s', record['name'])
+                self.client.delete_records(domain, name=record['name'], record_type='TXT')
```
zjs: It seems like there's a fundamental issue with the GoDaddy API functionality here (i.e., there's no delete operation, just a replace, which is inherently race-y), which we'll probably just have to live with.

It also seems like the GoDaddyPy library could be safer. It does a [full replace](https://github.com/eXamadeus/godaddypy/blob/master/godaddypy/client.py#L277) of all DNS records for the domain during deletion, which could cause data loss if other, potentially unrelated, DNS changes are being made while Certbot is running. Instead, it could do a [limited replace of records that match the type and name](https://developer.godaddy.com/doc#!/_v1_domains/recordReplaceTypeName), which would only impact other ACME challenges.

Better yet would be doing a limited replace _and_ checking the record content in addition to the name and type. Without that, users are likely to encounter issues if multiple instances of Certbot run concurrently.

Would you be interested in working with the library maintainer to see if these changes could be made upstream? If you'd prefer, I'd be happy to reach out to them.

certbot-dns-godaddy/certbot_dns_godaddy/dns_godaddy.py
```diff
@@ -0,0 +1,168 @@
...
+                logger.debug('Found base domain for %s using name %s', domain_name, guess)
+                return domain
+
+        raise errors.PluginError('Unable to determine base domain for {0} using names: {1}.'
+                                 .format(domain_name, domain_name_guesses))
+
+    @staticmethod
+    def _compute_record_name(domain, full_record_name):
+        # The domain, from Godaddy's point of view, is automatically appended.
+        return full_record_name.rpartition("." + domain)[0]
```
zjs: Now that two plugins need this logic, perhaps it would be good to move it to `dns_common` and avoid duplication.

certbot-dns-godaddy/setup.py
```diff
@@ -0,0 +1,69 @@
...
+    license='Apache License 2.0',
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Environment :: Plugins',
+        'Intended Audience :: System Administrators',
+        'License :: OSI Approved :: Apache Software License',
+        'Operating System :: POSIX :: Linux',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
```
zjs: It [seems](https://travis-ci.org/certbot/certbot/jobs/236741405) like the GoDaddyPy doesn't currently work with Python 2.6.

It may be worth asking them to add support (it looks like it would be a small change).

Alternatively, we could opt not to support Python 2.6 for this plugin, the main downside of which would be that we wouldn't be able to include this plugin with `certbot-auto` at this time. (More information in #4644.)

bahadir:
Thank you for your feedback. I'll make those changes to GoDaddyPy and contact maintainer asap, then I'll update this PR.

certbot-dns-godaddy/certbot_dns_godaddy/dns_godaddy.py
bahadir: Unfortunately, Godaddy doesn't provide a good API for this. **records** attribute for PUT request is mandatory, and empty list is not accepted.

If replaced record list empty (in our case, saved record list is going to be empty all the time), it simply fails with 400. It doesn't provide a DELETE request either.

bahadir: I've added two more tests to cover these API calls.

zjs: Thanks for investigating. That's... unfortunate.

For now, perhaps we can just document the potential concurrency issue.

bahadir:
How should I proceed?

zjs:
I think the outstanding issues are:

- [ ] Either remove Python 2.6 from `setup.py` (and adjust `tox.ini` not to test with Python 2.6) or get `GoDaddyPy` to support it (https://github.com/certbot/certbot/pull/4748#discussion_r118829970).
- [ ] Document the concurrency risk (in the code and somewhere that's user-visible) (https://github.com/certbot/certbot/pull/4748#discussion_r118848580)

zjs:
Hey @bahadir, thanks for your work on this so far! Is there anything I can do to help finish it off?

I'd be happy to submit a PR or two to your `dns-godaddy` branch to address the [work that's left](https://github.com/certbot/certbot/pull/4748#issuecomment-306293779), and then merge everything into `master` if it looks good to you.

bahadir:
I've opened a PR to GodaddyPy project to fix the Python 2.6 compatibility issue.

bahadir:
GodaddyPy seems to be Python 2.6 compatible now.

bmw:
That's great and thanks for taking the initiative to do the work here, however, the GoDaddyPy maintainer doesn't seem interested in maintaining Python 2.6 support or even claiming to be compatible with that version of Python (see eXamadeus/godaddypy#26). While it might be nice to have this plugin work on Python 2.6, it relying on libraries with questionable support for the platform seems like a bad idea if we want it to work solidly for people.

We're working towards dropping Python 2.6 support ourselves soon (see #2208). With this in mind, my preference would be we don't list this plugin as Python 2.6 compatible. We'd be making this change in the coming months anyway because the Certbot core (on which this plugin relies) will no longer work on Python 2.6.

Also, unless I missed it, you should also probably document the concurrency risk with this plugin as zjs suggested.

jgillula:
Hi @bahadir! Since we haven't seen any activity on this PR in a while, I wanted to check and see whether you wanted to continue working on it, or if you were OK with someone else picking up the work.

hurelhuyag:
Any progress?

sydneyli:
@bahadir @hurelhuyag We'd be happy to move forward with the GoDaddy DNS authenticator-- there seems to be another one open in #6135 which is possibly this branch but rebased on the most recent version of master. I'm closing this one out in favor of that one, though I'm happy to revert that decision (and keep this one open instead) if the authors of this PR would like to continue working on this.

Issue #4751, opened 2017-05-29
Certbot-auto virtual env setup fails with >1k lines of OpenSSL errors


pokrface:
## My operating system is (include version):
Ubuntu 16.04.02 LTS (4.4.0-72)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
Certbot-auto, currently using version 0.14.2

## I ran this command and it produced this output:
Running certbot-auto produces [this output](https://gist.github.com/pokrface/55a3f71e975098f7ebfd695560f5425f). There are lots of openssl references in there, and I believe my problem might be because I'm using OpenSSL 1.1.0f. However, I thought certbot supported OpenSSL 1.1.0 as of at least a few months ago (though I don't know).

## Certbot's behavior differed from what I expected because:
Certbot fails to bootstrap and will not start.

## Here is a Certbot log showing the issue (if available):
N/A - certbot doesn't get far enough to start logging and so the log contains no information

## Speculation follows
I'm using OpenSSL 1.1.0f and I suspect this is the problem.

SwartzCr:
Yes that is likely the problem. You might instead try to install it from the ubuntu ppa: https://launchpad.net/~certbot/+archive/ubuntu/certbot
Which also has 0.14.2 - though it might still break if you have the wrong version of openssl installed.
Let me know if that doesn't work?

SwartzCr:
Alternatively you might just need to rebuild certbot-auto. You could try deleting it and installing it again.
What is possibly happening here is that when you built certbot-auto it built a version of pythoncryptography to go along with your version of openssl - and made bindings to it's API. Now those API endpoints have changed and it has the wrong bindings. So if you rebuild pythoncryptography it might fix itself.
You can do this by having certbot-auto remake it's virtual environment. This happens typically only when we bump a version, but you can make it do this be deleting and re-installing it

pokrface:
Thanks, @SwartzCr - the errors occur during the virtual environment build, while certbot is compiling pythoncryptography. Based on the giant error dump, the failing command is:

```
Command 
"/home/lee/.local/share/letsencrypt/bin/python2.7 -u -c "import 
setuptools, 
tokenize;__file__='/tmp/pip-build-7qg1M5/cryptography/setup.py';exec(compile(getattr(tokenize,
 'open', open)(__file__).read().replace('

', '

'), __file__, 'exec'))" install --record 
/tmp/pip-SOmLff-record/install-record.txt 
[...]

PR #4755, opened 2017-05-31
Bump pinned cryptography version to 1.8.2

bmw:
This allows macOS tests to pass in Travis as seen in #4697. As a side effect of doing this:

* We have new pinned dependencies on `asn1crypto`, a newer version of `idna`, and `packaging`.
* We dropped our dependency on `pyasn1`.
* We no longer work with OpenSSL 1.0.0 as support was removed in cryptography 1.7.0. I failed to find any major OS that still uses this version of OpenSSL though.


yan12125:
Great! This also fixes ```./tools/venv3.sh``` on systems with Python 3.6 and OpenSSL 1.1.0f. (There's ```cryptography-1.5.3-cp36-cp36m-linux_x86_64.whl``` in my pip cache, but not on https://pypi.python.org/pypi/cryptography/1.5.3; I guess the cryptography team removed the py36 wheel for some reason) I believe #4751 can also be fixed.

PR #4760, opened 2017-05-31
route53: fix error handling

zjs:
Make error handling match other plugins:
 * Raise `PluginError` instead of errors from underlying libraries
 * Swallow errors during cleanup

Part of #4688.


zjs:
Good catch! (I didn't have a test for that case, so I missed it. Adding a test will require a bit of refactoring though, so I'll do that in a separate PR.)

I think just raising a `PluginError` there is easier than catching the `Exception` and re-raising.

zjs:
Closing and re-opening because coveralls seems stuck.

PR #4761, opened 2017-05-31
Improve bad preferred challenge error message

bmw:
Using ArgumentTypeError causes Certbot to report an unexpected error occurred while using ArgumentError causes argparse to print more usage information and call sys.exit().

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1295,7 +1295,7 @@ class _PrefChallAction(argparse.Action):
         try:
             challs = parse_preferred_challenges(pref_challs.split(","))
         except errors.Error as error:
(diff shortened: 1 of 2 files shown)


zjs:
Should the `ArgumentTypeError` on line 267 also be changed to `ArgumentError`?

bmw:
I don't think it should. That exception is raised in `read_file` which is only used as the `type` parameter with `argparse`. If `ArgumentTypeError` is raised there, `argparse` properly catches the exception, prints useful help output, and causes the program to exit. Additionally, we don't have a good value to provide as the first parameter to the `ArgumentError` constructor.

PR #4703, opened 2017-05-22
[#4382] Install git into Docker development file

kevinkle:
The first command listed in the docs at https://certbot.eff.org/docs/contributing.html#running-the-client-with-docker: `docker-compose run --rm --service-ports development bash -c 'tox -e lint'`
will fail without git installed.

```sh
 ❯ docker-compose run --rm --service-ports development bash -c 'tox -e lint'   [09:39:06]
lint create: /opt/certbot/src/.tox/lint
lint installed: appdirs==1.4.3,packaging==16.8,pyparsing==2.2.0,six==1.10.0
lint runtests: PYTHONHASHSEED='0'
lint runtests: commands[0] | /opt/certbot/src/tools/pip_install.sh -q -e acme[dev] -e .[dev] -e certbot-apache -e certbot-nginx -e certbot-dns-cloudflare -e certbot-dns-digitalocean -e certbot-dns-google -e certbot-dns-cloudxns -e certbot-compatibility-test -e letshelp-certbot
/opt/certbot/src/tools/pip_install.sh: 5: /opt/certbot/src/tools/pip_install.sh: git: not found
ERROR: InvocationError: '/opt/certbot/src/tools/pip_install.sh -q -e acme[dev] -e .[dev] -e certbot-apache -e certbot-nginx -e certbot-dns-cloudflare -e certbot-dns-digitalocean -e certbot-dns-google -e certbot-dns-cloudxns -e certbot-compatibility-test -e letshelp-certbot'
________________________________________ summary _________________________________________
ERROR:   lint: commands failed
```
I've added git to the Dockerfile-dev which addresses this. Linter runs fine now:

```sh
lint runtests: commands[1] | pylint --reports=n --rcfile=.pylintrc acme/acme certbot certbot-apache/certbot_apache certbot-nginx/certbot_nginx certbot-dns-cloudflare/certbot_dns_cloudflare certbot-dns-digitalocean/certbot_dns_digitalocean certbot-dns-google/certbot_dns_google certbot-dns-cloudxns/certbot_dns_cloudxns certbot-compatibility-test/certbot_compatibility_test letshelp-certbot/letshelp_certbot tests/lock_test.py
________________________________________ summary _________________________________________
  lint: commands succeeded
  congratulations :)
```

PR #4770, opened 2017-06-01
Remove dependency on git from pip_install.sh.

bmw:
Using git allowed this file to continue to work even if it was moved to another
directory. This slight increase in robustness wasn't worth it though as it
broke our development Dockerfile (see #4703), the certbot website's Dockerfile
(see certbot/website#226), and our test farm tests (see
certbot/tests/letstest/scripts/test_apache2.sh for an example that calls
tools/venv.sh without installing git). Rather than continuing to find and patch
these things, let's just allow this script to fail if it's moved rather than
propagating the git dependency all over the place.

### DIFF
--- a/tools/readlink.py
+++ b/tools/readlink.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+"""Canonicalizes a path and follows any symlinks.
+
+This is the equivalent of `readlink -f` on many Linux systems. This is
+useful as there are often differences in readlink on different
+platforms.
+
+"""
+from __future__ import print_function
+import os
+import sys
+
+print(os.path.realpath(sys.argv[1]))
--- a/tools/pip_install.sh
+++ b/tools/pip_install.sh
@@ -2,7 +2,8 @@
 # pip installs packages using Certbot's requirements file as constraints
 
 # get the root of the Certbot repo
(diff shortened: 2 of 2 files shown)


bmw:
While small and not even shipped with our code, I'd like this PR for 0.15.0. Without it, test farm tests will fail unless I modify the test code during the release process.

tools/readlink.py
```diff
@@ -0,0 +1,13 @@
...
+This is the equivalent of `readlink -f` on many Linux systems. This is
+useful as there are often differences in readlink on different
+platforms.
+
+"""
+from __future__ import print_function
+import os
+import sys
+
+print(os.path.realpath(sys.argv[1]))
```
ohemorange: Does this not just call `readlink -f` internally? Are we guaranteed a consistent result if we use the python version?

bmw: I don't think it calls `readlink -f` internally. Tests pass now and before they didn't.

Additionally, there are no restrictions on availability listed on [python.org](https://docs.python.org/2/library/os.path.html#os.path.realpath). At the very least, Python isn't documenting any differences in behavior for this function and (when compared to the rest of the module) is implicitly claiming it works on Windows where the shell command `readlink` isn't available.

PR #4694, opened 2017-05-19
Added options to set the listen address for standalone mode

jeffallen:
This might help with the problem mentioned in certbot/certbot#255.

PR #4773, opened 2017-06-02
Enable IPv6 support in standalone plugin

### LINKED ISSUE #1466: Standalone mode does not bind to ipv6

ohemorange:
Fixes #1466.

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -83,9 +83,10 @@ class BaseDualNetworkedServers(object):
                 kwargs["ipv6"] = ip_version
                 new_address = (address[0],) + (port,) + address[2:]
                 new_args = (new_address,) + args[1:]
-                server = ServerClass(*new_args, **kwargs)
+                server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
             except socket.error:
-                pass
+                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                    new_address[1], "IPv6" if ip_version else "IPv4")
             else:
                 self.servers.append(server)
                 # Always bind to the same port
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -26,6 +26,11 @@ class TLSServer(socketserver.TCPServer):
     """Generic TLS Server."""
 
     def __init__(self, *args, **kwargs):
+        self.ipv6 = kwargs.pop("ipv6", False)
+        if self.ipv6:
+            self.address_family = socket.AF_INET6
+        else:
+            self.address_family = socket.AF_INET
         self.certs = kwargs.pop("certs", {})
         self.method = kwargs.pop(
             # pylint: disable=protected-access
@@ -49,12 +54,62 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
     allow_reuse_address = True
 
 
+class TLSSNI01DualNetworkedServers(object):
+    """TLSSNI01 Server Wrapper. Tries everything for both. Failures for one don't
+       affect the other."""
+
+    def __init__(self, *args, **kwargs):
+        self.threads = []
+        self.servers = []
+
+        for ip_version in [True, False]:
+            try:
+                server = TLSSNI01Server(args, kwargs, ipv6=ip_version)
+            except socket.error as e:
+                pass
+            else:
+                self.servers.append(server)
+        if len(self.servers) == 0:
+            raise socket.error("Could not bind to IPv4 or IPv6.")
+
+    def serve_forever(self):
+        successful_servers = []
+        # Extra checks are probably not necessary here, since we would already
+        # have failed at bind in init
+        for server in self.servers:
+            if server is not None:
+                try:
+                    thread = threading.Thread(
+                        # pylint: disable=no-member
+                        target=server.serve_forever)
+                    thread.start()
+                except socket.error:
+                    pass
+                else:
+                    successful_servers.append(server)
+                    self.threads.append(thread)
+        self.servers = successful_servers
+        if len(self.servers) == 0:
+            raise socket.error("Could not serve on IPv4 or IPv6.")
+
+    def getsocknames(self):
+        return [server.socket.getsockname() for server in self.servers]
+
+    def shutdown_and_server_close(self):
+        for server in self.servers:
+            server.shutdown()
+            server.server_close()
+        for thread in self.threads:
+            thread.join()
+        self.threads = []
+
+
 class TLSSNI01Server(TLSServer, ACMEServerMixin):
     """TLSSNI01 Server."""
 
-    def __init__(self, server_address, certs):
+    def __init__(self, server_address, certs, ipv6=False):
         TLSServer.__init__(
-            self, server_address, BaseRequestHandlerWithLogging, certs=certs)
+            self, server_address, BaseRequestHandlerWithLogging, certs=certs, ipv6=ipv6)
 
 
 class BaseRequestHandlerWithLogging(socketserver.BaseRequestHandler):
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -89,7 +89,8 @@ class BaseDualNetworkedServers(object):
                     new_address[1], "IPv6" if ip_version else "IPv4")
             else:
                 self.servers.append(server)
-                # Always bind to the same port
+                # If two servers are set up and port 0 was passed in, ensure we always
+                # bind to the same port for both servers.
                 port = server.socket.getsockname()[1]
         if len(self.servers) == 0:
             raise socket.error("Could not bind to IPv4 or IPv6.")
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -54,21 +54,37 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
     allow_reuse_address = True
 
 
-class TLSSNI01DualNetworkedServers(object):
-    """TLSSNI01 Server Wrapper. Tries everything for both. Failures for one don't
-       affect the other."""
+class BaseDualNetworkedServers(object):
+    """Base class for a pair of IPv6 and IPv4 servers that tries to do everything
+       it's asked for both servers, but where failures in one server don't
+       affect the other.
 
-    def __init__(self, *args, **kwargs):
+       If two servers are instantiated, they will serve on the same port.
+       """
+
+    def __init__(self, address, unused_arg, **kwargs):
+        port = address[1]
+        ServerClass = kwargs.pop("server_class", object)
         self.threads = []
         self.servers = []
 
+        # Must try True first.
+        # Ubuntu, for example, will fail to bind to IPv4 if we've already bound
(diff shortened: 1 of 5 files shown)


ohemorange:
All of the assumptions in code here are based on things tested manually.

That is:

- If python is compiled without IPv6, binding to an IPv6 port will raise `socket.error`.
- On FreeBSD, IPv4 traffic will not forward to a socket bound on an IPv6 port, but you can ask for two different servers on the same address/port, with different protocols.
- On Ubuntu, IPv4 traffic will forward to a socket bound on an IPv6 port, but you cannot ask for two different servers on the same address/port, with different protocols.

ohemorange:
@bmw, I haven't added tests (nor have I tested the whole thing manually...), because I made a few design decisions here based on the discussion in #1466 and other places that might seem weirder now that they've actually been implemented. No need to do a full review just yet, but if you're good with the overall structure, I'll:

- add unit tests
- clean up the pieces of system test code that I've been using, and
- find some way to get them to run automatically on different flavors of unix.

ohemorange:
- check there's no race condition (things are deduped later) if a response comes in on both ports before we shut the server down

bmw:
For the purposes of trying to land everything we want in time for the release, I've asked @zjs to take a look at this, but one thing I wanted to mention:

> On Ubuntu, IPv4 traffic will forward to a socket bound on an IPv6 port, but you cannot ask for two different servers on the same address/port, with different protocols.

If desired I believe we can work around this using `IPV6_V6ONLY` as described [here](https://github.com/certbot/certbot/pull/3687#issuecomment-261113255).

acme/acme/standalone.py
```diff
@@ -49,12 +56,94 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
...
+        # If Python is compiled without IPv6, we'll error out but (probably) successfully
+        # create the IPv4 server.
+        for ip_version in [True, False]:
+            try:
+                kwargs["ipv6"] = ip_version
+                new_address = (address[0],) + (port,) + address[2:]
+                new_args = (new_address,) + args[1:]
+                server = ServerClass(*new_args, **kwargs)
+            except socket.error:
+                pass
```
zjs: nit: It seems like some logging here would be useful for debugging things if things don't work quite as we expect.

acme/acme/standalone_test.py
```diff
@@ -112,6 +118,96 @@ def test_http01_not_found(self):
...
+            if prev_port:
+                self.assertEqual(prev_port, port)
+            prev_port = port
+
+    def test_connect(self):
+        socknames = self.servers.getsocknames()
+        # connect to all addresses
+        for sockname in socknames:
+            host, port = sockname[:2]
+            cert = crypto_util.probe_sni(
```
zjs: This test fails for me locally with `Error: [Errno -9] Address family for hostname not supported` (which seems to match some of the Travis results.

acme/acme/standalone.py
```diff
@@ -49,12 +56,94 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
...
+        # socket. On the other hand, FreeBSD will successfully bind to IPv4 on the
+        # same port, which means that server will accept the IPv4 connections.
+        # If Python is compiled without IPv6, we'll error out but (probably) successfully
+        # create the IPv4 server.
+        for ip_version in [True, False]:
+            try:
+                kwargs["ipv6"] = ip_version
+                new_address = (address[0],) + (port,) + address[2:]
+                new_args = (new_address,) + args[1:]
+                server = ServerClass(*new_args, **kwargs)
```
zjs: nit: lint doesn't like this use of star args

ohemorange: ok but the linter thinks it's bad for readability but really we want it for functionality so I'm going to tell it to mind its own business.

acme/acme/standalone.py
ohemorange: It's expected behavior that this will fail, so setting the level to `DEBUG`.

acme/acme/standalone_test.py
zjs: 01a5ada solves this for me. The standalone plugin now binds to IPv6 on my machine (Ubuntu, Python 2.7).

ohemorange:
(so I remember after the meeting: for some reason passing 0 to the OS is giving out the same port twice)

ohemorange:
> If desired, I believe we can work around this and get the same behavior as the BSDs using IPV6_V6ONLY as briefly mentioned here.

Note: not easily, because you can't set that as `address_family`, it's a lower-level socket option.

ohemorange:
Anyway, I ended up mimicking FreeBSD-like conditions in the unit tests, and Travis runs things on Ubuntu, so I no longer feel as strongly that we need to automate system tests to get this in.

ohemorange:
Manual testing I've done:

- [x] Ubuntu with IPv6 enabled but no DNS record so it connects over IPv4
- [x] Ubuntu with IPv6 enabled + DNS
- [ ] FreeBSD with IPv6 + DNS

bmw:
I'm reviewing this.

acme/acme/standalone.py
```diff
@@ -49,12 +56,83 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
...
+                new_address = (address[0],) + (port,) + address[2:]
+                new_args = (new_address,) + args[1:]
+                server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
+            except socket.error:
+                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                    new_address[1], "IPv6" if ip_version else "IPv4")
+            else:
+                self.servers.append(server)
+                # Always bind to the same port
+                port = server.socket.getsockname()[1]
```
zjs: The purpose of this wasn't immediately obvious to me. I think it's for when `port` is initially `None`. Perhaps for future readers it might be helpful to expand the comment to explain the case(s) where this is is used?

ohemorange: pretty sure this is leftover from testing code and should be removed.

acme/acme/standalone.py
```diff
@@ -49,12 +56,84 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
...
+                new_args = (new_address,) + args[1:]
+                server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
+            except socket.error:
+                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                    new_address[1], "IPv6" if ip_version else "IPv4")
+            else:
+                self.servers.append(server)
+                # If two servers are set up and port 0 was passed in, ensure we always
+                # bind to the same port for both servers.
+                port = server.socket.getsockname()[1]
```
bmw: Nice catch.

acme/acme/standalone_test.py
```diff
@@ -1,6 +1,8 @@
 """Tests for acme.standalone."""
+import mock
```
bmw: super nit: Since this isn't part of the standard library, it should be moved down with the other 3rd party imports according to our [style](https://www.python.org/dev/peps/pep-0008/#imports) [guides](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting).

acme/acme/standalone_test.py
```diff
@@ -112,6 +120,134 @@ def test_http01_not_found(self):
...
+        def __init__(self, *args, **kwargs):
+            ipv6 = kwargs.pop("ipv6", False)
+            if ipv6:
+                self.address_family = socket.AF_INET6
+                kwargs["bind_and_activate"] = False
+            else:
+                self.address_family = socket.AF_INET
+            socketserver.TCPServer.__init__(self, *args, **kwargs)
+            if ipv6:
+                # pylint: disable=no-member
```
bmw: With this line, the three `disable=no-member` comments below can be removed.

ohemorange:
> Other than my nitpicky inline comments and the things we talked about out of band, one concern I have is the combination of the unreleased --<challenge-name>-address flag and this PR. What would you expect Certbot's behavior to be when using this flag and standalone? Is what we do correct/OK? Do we want to support it long term?

I built this PR on top of #4694, so I was intending it to work correctly with these changes. Manual testing indicates that's the case.

acme/acme/standalone.py
```diff
@@ -49,12 +56,83 @@ class ACMEServerMixin:  # pylint: disable=old-style-class
...
+       it's asked for both servers, but where failures in one server don't
+       affect the other.
+
+       If two servers are instantiated, they will serve on the same port.
+       """
+
+    def __init__(self, *args, **kwargs):
+        address = args[0]
+        port = address[1]
+        ServerClass = kwargs.pop("server_class", object)
```
ohemorange: It can't be `socketserver.TCPServer`, because that doesn't like when you pass in an `ipv6` kwarg. I think it's actually fine to make it a named argument, because you probably wouldn't directly use the Base class anyway, and our `TLSSNI01DualNetworkedServers` and `HTTP01DualNetworkedServers` set that argument for you.

bmw:
> I built this PR on top of #4694, so I was intending it to work correctly with these changes. Manual testing indicates that's the case.

It works, but it's a new flag that we have a chance to modify right now. Does it work the way we want long term. To be fair, this may not be the right place for this discussion as I don't think it should block this PR, however, it may cause us to block shipping #4694.

With that said, after thinking about it, I'm personally fine with things as they are now. Given I think we should support a flag like this, I think our options are:

1. Create both IPv4 and IPv6 versions of each flag.
2. Have one flag that can work for either address type.

Having separate flags makes Certbot uglier and increases complexity (what happens if you provide both IPv4 and IPv6 flags?). Currently, Certbot just works whether you provide an IPv4 or IPv6 address on the command line. As other plugins support this flag, we may have to write code that determines whether an address is IPv4 or IPv6 but then every plugin can benefit from it.

The only downside I see to this approach is you can't tell Certbot to listen on both IPv4 and IPv6 and specify the interface for each one. The only reason I could come up with where someone would want to do this is they have multiple domains that each require a different interface be used but they want all domains in the same cert. This is an extreme edge case though and I don't think it's unreasonable to have the user create two certs in this case.

I'd like someone else to make sure they agree with me here, but otherwise, I think the only thing left to worry about here are my obnoxious nits.

acme/acme/standalone_test.py
ohemorange: pylint scoping is strange.

ohemorange:
> The only downside I see to this approach is you can't tell Certbot to listen on both IPv4 and IPv6 and specify the interface for each one. The only reason I could come up with where someone would want to do this is they have multiple domains that each require a different interface be used but they want all domains in the same cert. This is an extreme edge case though and I don't think it's unreasonable to have the user create two certs in this case.

I think it's reasonable to say "create two certs" for that case.

PR #4794, opened 2017-06-07
route53: avoid listing the plugin twice in the UI

zjs:
Without this change, the Route53 plugin is listed twice when running
Certbot interactively (once under the old name, once under the new name).

This change ensures only the new name is shown, while maintaining hidden
backwards compatibility with the old name.


zjs:
Testing:

- [X]  Run Certbot interactively and verify the plugin is not shown under the old name.
- [X] Run Cerbot non-interactively, using the old name (`-a cerbot-route53:auth`)

PR #4796, opened 2017-06-07
Provide basic DNS authenticator documentation

zjs:
This commit provides documentation for the following DNS authenticator plugins:

* Cloudflare
* CloudXNS
* DigitalOcean
* DNSimple
* Google
* NS1

The documentation provides basic information about each of these plugins, information about the command line arguments for each of these plugins and the default values for those arguments, and information about the credentials used by each of these plugins.

The documentation is placed in the modules' `__init__.py` files so that the information is available from python contexts (e.g., from `help(certbot_dns_cloudflare)`) in addition to being included in the generated documentation by Sphinx.

### DIFF
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
@@ -1,17 +1,35 @@
-"""Cloudflare DNS Authenticator
+"""
+The `~certbot_dns_cloudflare.dns_cloudflare` plugin automates the process of
+completing a `dns-01` challenge (`~acme.challenges.DNS01`) using the Cloudflare
+API.
 
-This plugin automates the process of completing a dns-01 challenge
-(`~acme.challenges.DNS01`) using the Cloudflare API.
+
+Credentials
+-----------
 
 Use of this plugin requires a configuration file containing Cloudflare API
 credentials, obtained from your Cloudflare
 `account page <https://www.cloudflare.com/a/account/my-account>`_.
 
-Example:
-
 .. code-block:: ini
+   :name: credentials.ini
+   :caption: Example credentials file:
 
-  dns_cloudflare_email = cloudflare@example.com
-  dns_cloudflare_api_key = 0123456789abcdef0123456789abcdef01234567
+   # Cloudflare API credentials used by Certbot
+   dns_cloudflare_email = cloudflare@example.com
+   dns_cloudflare_api_key = 0123456789abcdef0123456789abcdef01234567
+
+The path to this file can be provided interactively or using the
+`--dns-cloudflare-credentials` command-line argument. Certbot records the path
+to this file for use during renewal, but does not store the file's contents.
+
+.. caution::
+   You should protect these API credentials as you would the password to your
+   Cloudflare account. Users who can read this file can use these credentials
+   to issue API calls on your behalf. Users who can cause Certbot to run using
+   these credentials can complete a `dns-01` challenge to acquire new
+   certificates or revoke existing certificates for associated domains.
+
+
 
 """
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
@@ -4,6 +4,20 @@ completing a `dns-01` challenge (`~acme.challenges.DNS01`) using the Cloudflare
 API.
 
 
+Named Arguments
+---------------
+
+======================================  =======  ==============================
+Argument                                Default  Description
+======================================  =======  ==============================
+`--dns-cloudflare-credentials`          None     Cloudflare credentials INI
+                                                 file. (See Credentials_.)
+`--dns-cloudflare-propagation-seconds`  10       The number of seconds to wait
+                                                 for DNS to propagate before
+                                                 asking the ACME server to
+                                                 verify the DNS record.
+======================================  =======  ==============================
+
 Credentials
 -----------
 
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
@@ -1,7 +1,7 @@
 """
 The `~certbot_dns_cloudflare.dns_cloudflare` plugin automates the process of
-completing a ``dns-01`` challenge (`~acme.challenges.DNS01`) using the
-Cloudflare API.
+completing a ``dns-01`` challenge (`~acme.challenges.DNS01`) by creating, and
+subsequently removing, TXT records using the Cloudflare API.
 
 
 Named Arguments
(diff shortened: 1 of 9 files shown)


zjs:
To facilitate review, here's a screenshot of what the rendered HTML documentation would look like:

image

zjs:
screen shot 2017-06-07 at 16 35 11-fullpage

certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
```diff
@@ -1 +1,76 @@
...
+   dns_cloudflare_api_key = 0123456789abcdef0123456789abcdef01234567
+
+The path to this file can be provided interactively or using the
+``--dns-cloudflare-credentials`` command-line argument. Certbot records the path
+to this file for use during renewal, but does not store the file's contents.
+
+.. caution::
+   You should protect these API credentials as you would the password to your
+   Cloudflare account. Users who can read this file can use these credentials
+   to issue API calls on your behalf. Users who can cause Certbot to run using
```
SwartzCr: you might want to specify that you can't necessarily scope what API calls the credentials can do, so I'd say something like "arbitrary API calls"

certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
```diff
@@ -1 +1,76 @@
...
+   these credentials can complete a ``dns-01`` challenge to acquire new
+   certificates or revoke existing certificates for associated domains.
```
SwartzCr: even if those domains aren't being managed by this server

certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
```diff
@@ -1 +1,76 @@
...
+
```
SwartzCr: should you provide an example of how to have these credentials be password protected but still allow certbot to automate renewal? (Is that a simple process?)

SwartzCr: should you mention what this challenge is doing?
People might worry that it's changing their DNS A records, so you should at least make a mention maybe?

zjs: We [warn](https://github.com/certbot/certbot/blob/master/certbot/plugins/dns_common.py#L304) if the file seems to have unsafe permissions, but beyond that I'm not sure what we should recommend at the moment.

One option would be to run certbot on a dedicated machine (with appropriate firewall rules, tightly controlled access, strict monitoring, etc.) and push certificates + private keys to the machines using the certificates via a post-renew hook, but that doesn't seem very elegant.

Once we can re-use private keys (#4610), the story gets a bit better, since private keys would no longer need to be sent over the wire following renewal.

zjs:
Note: this doesn't currently render a valid man page. (Filed https://github.com/sphinx-doc/sphinx/issues/3859.)

zjs:
I've added documentation for CloudXNS. It's very similar to the Cloudflare cloudflare documentation.

I'm open to ideas about how to deduplicate information between the two.

I worry that splitting the information across multiple pages (e.g., a page with common information for the DNS authenticators and then separate pages for the specifics) would greatly decrease readability/usability of the documentation though.

Probably, the best way to deduplicate information is to generate the documentation. Looking at Sphinx, that seems doable, but decidedly non-trivial. I'm not sure that this documentation will change [often enough for the initial investment to be worth it](https://xkcd.com/1205/). That also increases the barrier for someone less familiar with our stack to submit documentation changes/corrections.

certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
bmw: I agree a dedicated, locked down machine is probably the best option, but I'm not sure if that's what we should recommend to the average user (although perhaps we could mention in an offhand comment that's a way to provide additional security here).

For these files in general, I think we should provide instructions that the user should follow largely regardless of the machine they actually end up running Certbot on it. For example:

1. Minimal permissions on file containing API credentials (e.g. 600).
2. Minimal permissions with API credentials itself. In the rare cases it is possible, we should instruct the user to create the API credentials with the fewest permissions Certbot needs to operate.

I believe we decided when we had this conversation before we weren't going to password protect or encrypt the file as it doesn't provide much additional security. Is there anything else we should do?

To cc people who were interested in this issue before:

cc @joohoi, @jsha

zjs: >I agree a dedicated, locked down machine is probably the best option, but I'm not sure if that's what we should recommend to the average user (although perhaps we could mention in an offhand comment that's a way to provide additional security here).

Agreed that this isn't a good recommendation for the average user.

I think something like this probably belongs somewhere outside of the main documentation for these plugins.
 
> Minimal permissions on file containing API credentials (e.g. 600).

Certbot will warn about more generous permissions and I'll add a note about this to the documentation.

> Minimal permissions with API credentials themselves. In the rare cases it is possible, we should instruct the user to create the API credentials with the fewest permissions Certbot needs to operate.

The documentation for the Google plugin (just pushed) covers this. I'll do the same for AWS and any other providers where I can find a way to create API credentials with limited permissions.

jsha: I think the measures you suggest here are good. I agree that we should have documentation somewhere about how to do DNS-based validation and issuance on a separate machine and push certificates/keys. Then we can simply link that from the "average user" docs saying "For the most secure configuration, see X." But I don't think that should block these docs landing.

I also don't think push-based deployments are significantly blocked on key reuse. To be able to push new certificates and reload the web server, you need SSH or its moral equivalent. Sending along fresh keys via SSH is fine.

zjs:
## Cloudflare
screen shot 2017-06-08 at 16 23 35-fullpage

## CloudXNS
screen shot 2017-06-08 at 16 23 36-fullpage

## DigitalOcean
screen shot 2017-06-08 at 16 23 37-fullpage

## DNSimple
screen shot 2017-06-08 at 16 23 38-fullpage

## Google
screen shot 2017-06-08 at 16 23 40-fullpage

## NS1
screen shot 2017-06-08 at 16 23 41-fullpage

SwartzCr:
This looks good. Once these ship you should make a PR to add them to the documentation, maybe via a chapter in the users guide that links to all of them

aptalca:
@SwartzCr @zjs 

Are the read the docs pages (screenshot above) for the dns plugins publicly available? 

There is still very little info on these plugins anywhere. Certbot documentation merely references them, but even though it mentions the credentials ini files, it doesn't even say what they should contain. A user has to go through github to figure out how to use these plugins.

Thanks

SwartzCr:
@aptalca we plan to make them publically accessible once the plugins are packaged. Currently they aren't packaged for any distributions, so the only way to use them would be to build from source. We have plans to get them packaged in debian and ubuntu as well as certbot-auto, hopefully soon followed by other distros. Once that process begins we'll make these docs public on the certbot.eff.org documentation page

Issue #4789, opened 2017-06-06
prompts don't have a space after the :


SwartzCr:
you get things like 'Enter your email:noah@eff.org" which is ugly

SwartzCr:
fixed in that PR

PR #4813, opened 2017-06-09
Keep trailing space on prompt

### LINKED ISSUE #1919: bad text formatting in manual domain prompt
There are two problems going on with the manual prompts. One is that there are two spaces between the prompt determined by ops.py and the prompt from utils.py this is likely because it joins them at some point with a space. Additionally the space at the end (after the colon) is stripped. 
Below is an example of the text as just given to me on my server:

Please enter in your domain name(s) (comma and/or space separated)  (Enter 'c'
to cancel):x.example.com

bmw:
Fixes #1919 (I think).

This fixes the problem of lacking of a trailing space in our prompts (such as the prompt for an e-mail address). I'm not sure the problem Noah is describing in the first problem he lists in the issue. @SwartzCr?

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -176,12 +176,8 @@ class FileDisplay(object):
         if self._return_default(message, default, cli_flag, force_interactive):
             return OK, default
 
-        ans = input_with_timeout(
-            textwrap.fill(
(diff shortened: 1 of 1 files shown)


SwartzCr:
yup! this is what I wanted. This also fixes #4789

PR #4817, opened 2017-06-09
route53: add module-level documentation

zjs:
Add module-level documentation describing the use of `certbot-dns-route53` including discussion of credential management.

### DIFF
--- a/certbot-dns-route53/certbot_dns_route53/__init__.py
+++ b/certbot-dns-route53/certbot_dns_route53/__init__.py
@@ -1,1 +1,120 @@
-"""Certbot Route53 plugin."""
+"""
+The `~certbot_dns_route53.dns_route53` plugin automates the process of
+completing a ``dns-01`` challenge (`~acme.challenges.DNS01`) by creating, and
+subsequently removing, TXT records using the Amazon Web Services Route 53 API.
+
+
+Named Arguments
+---------------
+
+========================================  =====================================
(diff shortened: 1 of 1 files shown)


bmw:
@SwartzCr, would you be able to take a look at this as well?

certbot-dns-route53/certbot_dns_route53/__init__.py
```diff
@@ -1 +1,120 @@
...
+.. caution::
+   You should protect these API credentials as you would a password. Users who
+   can read this file can use these credentials to issue some types of API calls
+   on your behalf, limited by the permissions assigned to the account. Users who
+   can cause Certbot to run using these credentials can complete a ``dns-01``
+   challenge to acquire new certificates or revoke existing certificates for
+   domains these credentials are authorized to manage.
+
+
+Examples
```
SwartzCr: should you provide an example of how to set any of the environment variables here?

SwartzCr:
One small nit which I'm happy to take either way. After that I'm happy to land this

certbot-dns-route53/certbot_dns_route53/__init__.py
zjs: My thinking was that if users don't know how to set environment variables, it may be best to steer them towards the simplest solution. (And if they do know, I don't think they need an example.)

PR #4818, opened 2017-06-09
Low-impact cleanup of IDisplay

### LINKED ISSUE #4795: Remove old help code from IDisplay
Related to #3872, however, we decided we should just not support help and instead provide better prompts in the first place.

zjs:
Remove unused help-related display code. When NcursesDisplay was removed (d54cb76432a2eff43cc9cc3c1cc4d9136eac2221), help was deprecated. This change removes the remaining bits and pieces of code. This change does not attempt to re-incorporate messages which had been "orphaned" (under the assumption that if we managed to live without them for months, they probably weren't that important in the first place).

Remove unused escape-related display code. When NcursesDisplay was removed (d54cb76432a2eff43cc9cc3c1cc4d9136eac2221), escape was deprecated. This change removes the remaining bits and pieces of code.

Remove uses of unused menu parameters.

Remove unused default_status/default_state argument from checklist. (This seems safe because not only is it unused, the parameter has different names in the interface and implementation)

Resolves #4795.

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -24,10 +24,10 @@ CANCEL = "cancel"
 """Display exit code for a user canceling the display."""
 
 HELP = "help"
-"""Display exit code when for when the user requests more help."""
+"""Display exit code when for when the user requests more help. (UNUSED)"""
 
 ESC = "esc"
-"""Display exit code when the user hits Escape"""
+"""Display exit code when the user hits Escape (UNUSED)"""
 
 
 def _wrap_lines(msg):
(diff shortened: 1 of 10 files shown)


certbot/display/util.py
```diff
@@ -23,9 +23,6 @@
 CANCEL = "cancel"
 """Display exit code for a user canceling the display."""
 
-HELP = "help"
```
bmw: I'm not sure if we want to delete this due to its usage [here](https://github.com/gboudreau/certbot-heroku/blob/2837a5bdfe527853311e57923557ec77b78c3317/certbot_heroku/configurator.py#L107). Did you have any thoughts about my comments [here](https://github.com/certbot/certbot/issues/3915#issuecomment-307519556) and did you check if `IDisplay`/`HELP` is used anywhere else?

If there aren't many plugins using it, I think we can open an issue/PR against their repos to notify them of the upcoming change. Alternatively, we can keep this around and update the docstring to state that it's deprecated.

zjs: Ah, good catch! I saw your comment and was thinking about it in the context of #3915, but not #4795.

I'll revisit this with that in mind.

PR #4821, opened 2017-06-11
Fix digitalocean plugin tests for Python 3

felixonmars:
assertItemsEqual() doesn't exist in Python 3.x. Travis didn't fail because of some nose errors (so not all tests were run).

### DIFF
--- a/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
+++ b/certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
@@ -5,6 +5,7 @@ import unittest
 
 import digitalocean
 import mock
+import six
 
 from certbot import errors
 from certbot.plugins import dns_test_common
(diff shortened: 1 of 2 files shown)


zjs:
> Travis didn't fail because of some nose errors (so not all tests were run).

Filed #4825 for this.

certbot-dns-digitalocean/certbot_dns_digitalocean/dns_digitalocean_test.py
```diff
@@ -2,6 +2,7 @@
 
 import os
 import unittest
+import six
```
zjs: It seems like `six` should be added to `certbot-dns-digitalocean`'s `setup.py` file.

felixonmars: Updated :)

zjs: I hate to nit-pick, but it also seems like `six` should be listed in the next [group of imports](https://www.python.org/dev/peps/pep-0008/#imports), instead of with standard library imports?

felixonmars: As you wish :3

PR #4829, opened 2017-06-12
Get more information out of brew

bmw:
`brew`, why is the post-install step for OpenSSL failing?


DomT4:
This was fixed by the [new tag](https://github.com/Homebrew/brew/releases/tag/1.2.3) being cut earlier. Apologies if was briefly confusing!

PR #4833, opened 2017-06-14
Add compatible if modified server agent.

SunFulong:


bmw:
Thanks for the PR @SunFulong. In what cases might the output of `nginx -V` be modified? Are there Nginx configuration directives that do this?

SunFulong:
@bmw , version string can be modified by edit source, it's common in 3rd re-published nginx, like China's Tengine ( http://tengine.taobao.org/ ), or other cases end user modify it to hide real server agent to prevent attack.

bmw:
Hmm. @ohemorange, do you have an opinion here?

ohemorange:
related: https://github.com/certbot/certbot/issues/4919#issuecomment-314927218

ohemorange:
Thanks for the patch, @SunFulong!

I think the best thing to do here is to leave the regex as is and do the existing check if we find matches. But if we don't find a version string matching the official version format, we should print a warning saying that we don't officially support forks of Nginx, and continue. This is different from what we do now in that we currently error out if we can't find a version string.

ohemorange:
@SunFulong, do you have the bandwidth to make that change?

ohemorange:
Closed by #5467.

PR #4852, opened 2017-06-19
Add __main__.py

bmw:
Adding this file allows Certbot to easily be invoked from something like `coverage run certbot` which will look for `__main__.py` to know which file to run as described briefly [here](https://docs.python.org/3/library/__main__.html).

This also fixes us not actually logging the return value of `certbot.main.main` as our entry point was on `certbot.main:main` so the `if __name__ == '__main__'` code at the bottom of the file was silently ignored.

### DIFF
--- a/certbot/__main__.py
+++ b/certbot/__main__.py
@@ -23,4 +23,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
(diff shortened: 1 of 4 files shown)


bmw:
I don't really think we need this.

PR #4854, opened 2017-06-19
Report and enforce coverage on integration tests

bmw:


tests/boulder-integration.sh
```diff
@@ -217,3 +220,5 @@ if ${CI:-false} || type nginx;
 then
     . ./certbot-nginx/tests/boulder-integration.sh
 fi
+
+coverage report --fail-under 52 -m
```
pde: We agreed that the better solution is probably to merge the coverage and py27 integration travis tests, and get a single threshold for unit + integration tests that way. But this is fine for now.

pde: Though current coverage seems to be 59% so maybe we should threshold there?

bmw: Oh yeah. It was 52 when I tested locally but it didn't run the nginx tests because I didn't have nginx installed. I'll update this number when I see where the coverage is at including #4855.

PR #4855, opened 2017-06-19
Test additional subcommands in integration tests

bmw:


bmw:
I added a test for rollback with Nginx, but they failed due to bugs in our command line parsing. See #4858 for the bug and `test-rollback` for the test.

PR #4248, opened 2017-02-22
Bump requests requirement to >=2.10

bmw:
Fixes #3804.

PR #4857, opened 2017-06-19
Finish oldest tests

### LINKED ISSUE #4049: Flesh out oldest tests
We should find the oldest versions of all our Python dependencies used in OS packages and add them to the [oldest tests](https://github.com/certbot/certbot/blob/master/tox.ini#L36) in Travis. This will prevent bugs like #3098 and #4040 from slipping into a release.

The two distros I'd check here are CentOS 7 and Debian 8.

bmw:
Fixes #4049. Built on #4856.

### DIFF
--- a/certbot-dns-google/certbot_dns_google/dns_google_test.py
+++ b/certbot-dns-google/certbot_dns_google/dns_google_test.py
@@ -76,7 +76,7 @@ class GoogleClientTest(unittest.TestCase):
 
     @mock.patch('oauth2client.service_account.ServiceAccountCredentials.from_json_keyfile_name')
     @mock.patch('certbot_dns_google.dns_google.open',
-                mock.mock_open(read_data='{"project_id": "' + PROJECT_ID + '"}'))
+                mock.mock_open(read_data='{"project_id": "' + PROJECT_ID + '"}'), create=True)
     def test_add_txt_record(self, unused_credential_mock):
         client, changes = self._setUp_client_with_mock([{'managedZones': [{'id': self.zone}]}])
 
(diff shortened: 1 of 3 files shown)


bmw:
This PR reverts #4248 which pinned a newer version of `requests` that provided better error messages for `SSLErrors` that occur. Since our conversation in #4433 though where we decided that we're only actively supporting installations through OS packages or `certbot-auto` (or Docker) though, I don't think this pinning buys us anything. `certbot-auto` pins an earlier version and OS packages will use whatever they have available. Perhaps this pinning would have caused an OS to upgrade their `requests` version, however, there's no guarantee that will actually happen and removing it allows easy testing with older versions to prevent regressions in systems like CentOS/RHEL which I think is more important.

@erikrose, I'd like your +1 on this specific part of this PR when you get a few spare cycles.

pde:
(needs a merge, assigning to Brad)

bmw:
Merge conflict resolved, but it needs another review now that we have the repo configured to not accept stale reviews.

I'd also kind of like @erikrose's +1 on the `setup.py` changes before this is merged.

certbot-dns-google/certbot_dns_google/dns_google_test.py
```diff
@@ -76,7 +76,7 @@ def _setUp_client_with_mock(self, zone_request_side_effect):
 
     @mock.patch('oauth2client.service_account.ServiceAccountCredentials.from_json_keyfile_name')
     @mock.patch('certbot_dns_google.dns_google.open',
-                mock.mock_open(read_data='{"project_id": "' + PROJECT_ID + '"}'))
+                mock.mock_open(read_data='{"project_id": "' + PROJECT_ID + '"}'), create=True)
```
erikrose: What is this for?

erikrose: Can you help me understand how this addition related to the rest of the patch?

bmw: Without this change, tests fail like this https://travis-ci.org/certbot/certbot/jobs/245125937#L5394 due to the older version of `mock`. Rather than just patching `open()` globally, we can use `create=True` to have `mock.patch()` create an `open` attribute on the specific module we're testing.

erikrose: Gotcha.

PR #4863, opened 2017-06-21
Nginx compatibility test

### LINKED ISSUE #4087: Fix nginx compatibility tests
The nginx compatibility tests fail when taking #3565, merging in `master`, and running tests. All failures are due to redirects where the tests expect us to have set up a redirect, but we have not (correctly) done so. In all/most cases, we've added a commented out redirect in the nginx config and output a message like:
```
The appropriate server block is already redirecting traffic. To enable redirect anyway, uncomment the redirect lines in ...
```
I don't expect there to be a critical issue here. I think either:

1. We're being too conservative and not enabling redirects when we should.
2. The compatibility tests themselves are wrong.

Neither of these things is a big deal. Here is a [log](https://drive.google.com/file/d/0B5w-1Ta1Rduua0RVU3FJdlhRbTQ/view?usp=sharing) of the issue (after modifying `tox.ini` to skip authenticator tests which all currently pass). It'd be nice to get these tests working because they could easily be run automatically as part of #2189.

bmw:
Fixes #4087. Supersedes #3565.

Tests now pass and hopefully they are fast enough to be run in Travis at least occasionally.


bmw:

schoen:
Yes, I can take a look.

PR #4865, opened 2017-06-21
Turn on IRC notifications for Travis failures in master

bmw:
This should work according to instructions at https://docs.travis-ci.com/user/notifications/#Configuring-IRC-notifications. According to that section, pull request builds don't trigger failure messages so I'm unfortunately not sure how to test this without merging this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -180,4 +180,3 @@ notifications:
     on_success: never
     on_failure: always
     use_notice: true
-    skip_join: true


SwartzCr:
Have you removed the IRC channel's ban on external messages? Otherwise it seems that skip join will prevent it from posting

bmw:
This disables `skip join`. Now a bot will temporarily join the channel to notify us of build failures.

SwartzCr:
That's not my reading of the documentation page, are you sure you don't want it to be false?

bmw:
Did you find different documentation than the page I linked above? The only example of `skip join` given in the documentation I found is setting it to `true` to stop the bot from joining before sending messages. There are no examples of setting `skip join: false` there, but it probably works and we can try it if you prefer.

PR #4883, opened 2017-06-28
Finish manual Cloudflare documentation improvements

bmw:
Built on #4693. Addresses @joohoi's review comments.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -653,8 +653,8 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not 
    API_KEY="your-api-key"
    EMAIL="your.email@example.com"
 
-   # Strip only the top domain to get the zone id
-   DOMAIN=$(expr match "$CERTBOT_DOMAIN" '.*\.\(.*\..*\)')
+   # Fetch the SOA domain for the domain/subdomain
+   DOMAIN=$(dig soa "$CERTBOT_DOMAIN" | grep -v ^\; | grep SOA | awk '{print $1}';)
 
    # Get the Cloudflare zone id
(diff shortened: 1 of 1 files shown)


mjmasn:
@bmw hmm, slight issue with my approach in #4693 (either I was mistaken about LE checking SOA domain as well as the actual domain or something has changed):

We have domains `example.com` and `example.org`, `www.example.ORG `is CNAME'd to `subdomain.example.COM` so dig gives us `example.COM.` as the SOA. This was working previously, but appears LE is only checking `_acme-challenge.example.org` rather than `_acme-challenge.example.org.example.com`. That's probably fair enough, but it does complicate things again.

This is still an issue CloudFlare could fix by allowing searching for a zone by subdomain...

SwartzCr:
I'm fine with the text about dig and curl, but can't comment on the SOA issue. Maybe @zjs can?

zjs:
> I'm fine with the text about dig and curl, but can't comment on the SOA issue. Maybe @zjs can?

I couldn't find a reliable way to make use of the SOA record when implementing the Cloudflare plugin.

I ended up taking a naive approach which seemed to work: given a domain like foo.bar.baz.example.org, query Cloudflare for a zone matching progressively less specific portions of the FQDN until I got a "hit" (foo.bar.baz.example.org, then bar.baz.example.org, then baz.example.org, and finally example.org). Within that zone, you compose the record name using the original FQDN.

Regardless of CNAMEs, this seems to give the right behavior: Let's Encrypt expects a record for `_acme-challenge.foo.bar.baz.example.org` and Cloudflare expects that record to appear in the `example.org` zone.

bmw:
Thanks for the input Zach.

With this in mind, I'm leaning towards closing this and #4693 unless anyone has any objections.

PR #4889, opened 2017-06-29
Fix nginx --dry-run

### LINKED ISSUE #4846: Nginx fails when --dry-run is provided
This is because `certbot.crypto_util.init_save_key` doesn't actually save they key if `--dry-run` is provided as an attempt to resolve #2495.

cc @ohemorange

bmw:
Fixes #4846.

This reverts #4380 and moves conditionally saving CSRs and keys out of `init_save_*`. I think this makes sense as both of these functions take the directory where the file should be saved. Whether or not the file should be saved during a dry run really depends on the directory used, so let's push this decision into the caller and have the functions actually do what they say: create and save a CSR/key.


bmw:
Yes I did. Thanks for the review!

PR #4891, opened 2017-06-29
Hide exceptions that occur during session.close()

### LINKED ISSUE #4840: client.py broken for python 3.4
Gentoo 4.8.17-hardened-r2

https://certbot.eff.org/#gentoo-nginx

`certbot renew --dry-run`

```
Exception ignored in: <bound method ClientNetwork.__del__ of <acme.client.ClientNetwork object at 0x7faec2bf1f60>>
  File "/usr/lib64/python3.4/site-packages/acme/client.py", line 522, in __del__
    self.session.close()
  File "/usr/lib64/python3.4/site-packages/requests/sessions.py", line 689, in close
    for v in self.adapters.values():
  File "/usr/lib64/python3.4/_collections_abc.py", line 512, in __iter__
    for key in self._mapping:
  File "/usr/lib64/python3.4/collections/__init__.py", line 91, in __iter__
    curr = root.next

ReferenceError: weakly-referenced object no longer exists
```

There was an exception

The log `/var/log/letsencrypt/letsencrypt.log` does not show any errors.
```
user nginx nginx;
worker_processes 1;

error_log /var/log/nginx/error_log info;

events {
        worker_connections 1024;
        use epoll;
}

http {
        include /etc/nginx/mime.types;
        default_type application/octet-stream;

        log_format main
                '$remote_addr - $remote_user [$time_local] '
                '"$request" $status $bytes_sent '
                '"$http_referer" "$http_user_agent" '
                '"$gzip_ratio"';

        client_header_timeout 10m;
        client_body_timeout 10m;
        send_timeout 10m;

        connection_pool_size 256;
        client_header_buffer_size 1k;
        large_client_header_buffers 4 2k;
        request_pool_size 4k;

        gzip off;

        output_buffers 1 32k;
        postpone_output 1460;

        sendfile on;
        tcp_nopush on;
        tcp_nodelay on;

        keepalive_timeout 75 20;

        ignore_invalid_headers on;

        index index.html;

        server {
                listen 0.0.0.0;
                server_name <redacted>;

                access_log /var/log/nginx/<redacted>.access_log main;
                error_log /var/log/nginx/<redacted>.error_log info;

                root /var/www/www.<redacted>.com;

                if ($scheme != "https") {
                } # managed by Certbot


                # Redirect non-https traffic to https
                # if ($scheme != "https") {
                #     return 301 https://$host$request_uri;
                # } # managed by Certbot

        }

        # SSL example
        server {
                listen 0.0.0.0:443;
                server_name <redacted>;

                ssl on;
                ssl_certificate /etc/letsencrypt/live/<redacted>.com/fullchain.pem; # managed by Certbot
                ssl_certificate_key /etc/letsencrypt/live/<redacted>.com/privkey.pem; # managed by Certbot

                access_log /var/log/nginx/<redacted>.ssl_access_log main;
                error_log /var/log/nginx/<redacted>.ssl_error_log info;

                root /var/www/www.<redacted>.com;

        }
}
```

bmw:
This fixes #4840. Exceptions that are raised out of `__del__` methods are caught and printed to stderr. By catching any exceptions that occur, we now prevent this from happening.

Alternative solutions to this would have been either not calling session.close() at all or adding a close() method to acme.client.ClientNetwork, acme.client.Client, and certbot.client.Client and using certbot.client.Client in a context manager to ensure close() is called. The former means that users of the ACME library never properly close their connections until their program exits and the latter adds a lot of complexity and nesting of client code for little benefit.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -523,7 +523,7 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
         # user if the call to close() fails. See #4840.
         try:
             self.session.close()
-        except:  # pylint: disable=bare-except
+        except Exception:  # pylint: disable=broad-except
             pass
 
     def _wrap_in_jws(self, obj, nonce):
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -519,7 +519,12 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
         self._default_timeout = timeout
 
     def __del__(self):
-        self.session.close()
+        # Try to close the session, but don't show exceptions to the
+        # user if the call to close() fails. See #4840.
+        try:
+            self.session.close()
+        except:  # pylint: disable=bare-except
+            pass
 
     def _wrap_in_jws(self, obj, nonce):
         """Wrap `JSONDeSerializable` object in JWS.
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -600,11 +600,18 @@ class ClientNetworkTest(unittest.TestCase):
             mock.ANY, mock.ANY, verify=mock.ANY, headers=mock.ANY,
             timeout=45)
 
-    def test_del(self):
+    def test_del(self, close_exception=None):
         sess = mock.MagicMock()
+
+        if close_exception is not None:
+            sess.close.side_effect = close_exception
+
         self.net.session = sess
         del self.net
         sess.close.assert_called_once_with()
+
+    def test_del_error(self):
+        self.test_del(ReferenceError)
 
     @mock.patch('acme.client.requests')
     def test_requests_error_passthrough(self, mock_requests):


acme/acme/client.py
```diff
@@ -519,7 +519,12 @@ def __init__(self, key, alg=jose.RS256, verify_ssl=True,
         self._default_timeout = timeout
 
     def __del__(self):
-        self.session.close()
+        # Try to close the session, but don't show exceptions to the
+        # user if the call to close() fails. See #4840.
+        try:
+            self.session.close()
+        except:  # pylint: disable=bare-except
```
ohemorange: any reason not to specifically `except ReferenceError`?

bmw: My thought was we want to make a best effort attempt to properly close the session. Using `__del__` is very error prone and we don't want to print tracebacks for any strange errors that might occur. By using the bare except, we're protected from dealing with this problem again if other errors manifest in the future.

Perhaps we want to know what other exceptions are possible though and want the bug reports. I don't really have a strong preference though and created #4894 with this change. Feel free to approve whichever one you like better.

PR #4895, opened 2017-06-30
Deploy hook

### LINKED ISSUE #3947: --deploy-hook
`--renew--hook` should probably be supplanted by `--deploy-hook`, which should run when you first make a cert as well as when you renew one.

bmw:
Fixes #3947.

This PR hides `--renew-hook` and adds `--deploy-hook`. They work the same way except `--deploy-hook` is also run with the `certonly` and `run` subcommands.

Deploy hooks are still saved in the renewal configuration file as `renew_hook`. This is important for forwards compatibility as it is fairly common for people to downgrade their Certbot version when switching from `certbot-auto` to OS packages. Deploy hooks are not saved as `deploy_hook` as well to prevent confusion and mismatched values.

@ynasser, do you have time to review this or would you like me to ask Erica?

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1095,7 +1095,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " space-delimited list of renewed certificate domains (for example,"
         " \"example.com www.example.com\"")
     helpful.add(
-        "renew", "--deploy-hook",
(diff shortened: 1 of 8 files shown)


ynasser:
Yes, I can take a look by the end of today!

bmw:
Nice catch! I've updated the file.

PR #4897, opened 2017-06-30
Replace the easy v. secure prompt with more clear choices

### LINKED ISSUE #3328: Add more context to "Easy vs Secure" prompt
Right now you get this prompt for redirect

```
Please choose whether HTTPS access is required or optional.
1: Easy - Allow both HTTP and HTTPS access to these sites
2: Secure - Make all requests redirect to secure HTTPS access
```

One confusing thing: The title asks "required or optional," but those words don't appear in either option.

Also, we should provide more advice on how to choose. I.e. 

```
1: Easy - Allow both HTTP and HTTPS access to these sites. Choose this if you haven't yet tested that your sites function properly when visited on HTTPS.
2: Secure - Make all requests redirect to secure HTTPS access. Choose this for new sites, or if you're confident your site works on HTTPS. You can undo this change by editing your web server's configuration.
```

ohemorange:
Fixes #3328, though we should still make the larger change to a more expressive, expansive UI.

### DIFF
--- a/certbot/display/enhancements.py
+++ b/certbot/display/enhancements.py
@@ -49,7 +49,7 @@ def redirect_by_default():
     ]
 
     code, selection = util(interfaces.IDisplay).menu(
-        "Please choose whether or not to add a redirect removing HTTP access.",
+        "Please choose whether or not to redirect HTTP traffic to HTTPS, removing HTTP access.",
         choices, default=0,
         cli_flag="--redirect / --no-redirect", force_interactive=True)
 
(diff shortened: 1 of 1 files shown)


schoen:
I agree that this is a huge improvement over the status quo. Do you think the phrasing about "removing HTTP access" might confuse some people in terms of thinking that the port 80 listener is turned off?

ohemorange:
@schoen -- how about "Please choose whether or not to redirect HTTP traffic to HTTPS, removing HTTP access."

schoen:
Yeah, I think that may be a little clearer.

I've seen a fair amount of confusion about this kind of thing on the forums (in terms of people's different ideas about "turning off" or "disabling" HTTP, which for some of them just means redirecting, but for other people means turning off the port 80 listener).

PR #4515, opened 2017-04-17
Add support for Devuan-based OSes

linuxitux:
Tested on Devuan 1.0 Jessie.

PR #4903, opened 2017-07-05
First Devuan integration

maurorappa:
Devuan 1.0 (https://devuan.org/) is now officially released and it has with long term support.
It's time to add it to the supported OSes and being a Debian fork the integration effort is minimal.
It already ships the recent version of the certbot requirements:

root@devuan:~# lsb_release  -a
No LSB modules are available.
Distributor ID:	Devuan
Description:	Devuan GNU/Linux 1.0 (jessie)
Release:	1.0
Codename:	jessie
root@devuan:~# LC_ALL=C apt-cache show --no-all-versions libaugeas0 | grep ^Version: | cut -d" " -f2
1.2.0-0.2+deb8u1
root@devuan:~# LC_ALL=C apt-cache show --no-all-versions python | grep ^Version: | cut -d" " -f2
2.7.9-1

if you want to test it in AWS, there's an AMI ready (ami-9bafb8ff London region, ami-9bafb8ff N Virginia region, ami-e043e78f Frankfurt region)

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -228,6 +228,8 @@ BootstrapDebCommon() {
   # - Debian
   #     - 7.9 "wheezy" (x64)
   #     - sid (2015-10-21) (x64)
+  # - Devuan
+  #     - 1.0 "Jessie" (x64)
 
   # Past versions tested with:
   #
@@ -605,6 +607,9 @@ Bootstrap() {
       return
   elif [ -f /etc/debian_version ]; then
     BootstrapMessage "Debian-based OSes"
+    BootstrapDebCommon
+  elif [ -f /etc/devuan_version ]; then
+    BootstrapMessage "Devuan"
     BootstrapDebCommon
   elif [ -f /etc/mageia-release ]; then
     # Mageia has both /etc/mageia-release and /etc/redhat-release


maurorappa:
I am frankly astonished:  if I go here: https://certbot.eff.org/#debianother-nginx, it suggest to me to use certbot-auto!
Since it doesn't recognize Devuan, I wrote this small patch to make if work seamlessly.
I don't see an issue with it.

bmw:
That makes sense. We should update our instructions.

Really, "Debian (other)" shouldn't tell people to use `certbot-auto`. The only supported Debian system where we don't have packages available through `apt-get` is Debian Wheezy which is explicitly mentioned.

What if we updated "Debian (other)" to use OS packages and/or added an option in the dropdown menu to explicitly mention Devuan Jessie so people using that system get the correct advice?

maurorappa:
I think the dropdown addition seems easy and very helpful to guide people!

bmw:
Great! I created certbot/website#267 to add options for Devuan.

With this in mind, what would you like to see happen with this PR? Now that we (will) have instructions for Devuan telling them how to install Certbot on their system, do you think the problem here has been resolved or do you also want to see Devuan support added to `certbot-auto`?

The reason I rejected #4515 as I explained there is `certbot-auto` is disproportionately the cause of bug reports on this repo. It was written to be a temporary solution to packaging while distros caught up and the core Certbot team want to deprecate it as soon as we can. In general, OS packages provide a much better experience so attracting more people to the script and adding more systems which we must try and support with the script doesn't excite me too much.

maurorappa:
Fine, let's close it and forget about certbot-auto  which will be deprecated soon.
Thanks

bmw:
Sounds good to me. Regardless, thanks for pointing out the problems with our instructions for Devuan!

PR #4195, opened 2017-02-09
Add flags to configure log rotation

sc-perth:
Provides a quick, and possibly keep-able, fix to certbot logging issues. Primarily: certbot handling log rotation on its own (--disable-log-rotation) & certbot keeping an insane number of logs (--max-log-count). Further details in #3602 
Thank you,

Issue #3602, opened 2016-10-10
Change log rotation scheme, again


sc-perth:
Certbot's built in log rotation prevents logs from being handled in the manner defined by the system administrator. I see that in #3382 ([2bdf125](https://github.com/certbot/certbot/commit/2bdf1258ad9db1a3a0681a3b81614c93a1d5b96d)) we discovered one aspect of this issue and addressed that by increasing the number of logs kept by Certbot to 1000.

I object to this fix for a few reasons:
1) Systems not specifically configured to address Certbot's logs will end up with 1001 log files.
2) It is not typically in the utility's scope to address log rotation.
3) An administrator seeking to control Certbot's logging must now consider how to work around the existing log rotation, instead of simply addressing log size & count in the normal fashion.

For these reasons, please consider removing the built in (and hard coded) log rotation. That will bring Certbot's log file handling into line with other *nix utilities. If it remains necessary to call logger.handlers.RotatingFileHandler; please then set the size & count to 0; this will cause it to constantly write to the same file. 

Another note is that some/all log rotators can inform a process when a log is rotated; it may be desirable to enable this integration. Example: `path_to_pid_cmd_file` section in the [newsyslog.conf man page](http://www.unix.com/man-page/freebsd/5/newsyslog.conf/) & `postrotate` section in the [logrotate man page](https://linux.die.net/man/8/logrotate).

Lastly, adding a section to the documentation on Certbot's logging may be advised.

Thank you!

pde:
@jsha thoughts? Interested in taking this?

jsha:
I agree the current hard-coded log rotation is bad. I increased the number of logs to avoid the most pernicious problem, that logrotate + certbot rotation silently drops logs. That problem is still present, but only if logrotate is configured to retain more than 1000 files, which is rare.

I'm afraid I don't have the time for the more thorough fix, but I would encourage anyone to take it!

It's also worth thinking about how to configure "out of the box." The Debian / etc packages can install something in logrotate.d. However, if someone installs with certbot-auto on a machine with low disk, it's likely that we would eventually cause an out of disk error unless they do additional configuration. Maybe we just decide that our daily log growth, when run daily, is small enough not to worry about it?

sc-perth:
I have 2 questions that I think will help this move forward.
1. @jsha Will _certbot_ have issues if it simply constantly appends to the same log file (disable all built in log rotation)?
2. What were the design considerations that caused log rotation to be implemented in the first place?

[...]

PR #4907, opened 2017-07-06
Make Certbot's log rotation configurable

### LINKED ISSUE #2292: Add option to log to one file only
Feature request!

I'd like to have the option to log to one file only in the log_dir (or be able to specifically set a singular log and not have it rotate), this helps with docker logs ala nginx e.g. in Dockerfile

RUN ln -sf /dev/stdout /var/letsencrypt/letsencrypt.log

OR have the ability to print the errors that _would_ be in the log to stdout

bmw:
Fixes #2292 and #3602. Built on #4195.

This PR adds a single flag to allow the user to control Certbot's log rotation. This PR functionally implements most of the desired features discussed in #3602. We'd be open to well written PRs for the missing functionality.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -440,8 +440,8 @@ class HelpfulArgumentParser(object):
         }
 
         # List of topics for which additional help can be provided
-        HELP_TOPICS = ["all", "security", "paths", "automation", "testing"] + list(self.VERBS)
-        HELP_TOPICS += self.COMMANDS_TOPICS + ["manage"]
+        HELP_TOPICS = ["all", "security", "paths", "automation", "testing"]
+        HELP_TOPICS += list(self.VERBS) + self.COMMANDS_TOPICS + ["manage"]
 
         plugin_names = list(plugins)
         self.help_topics = HELP_TOPICS + plugin_names + [None]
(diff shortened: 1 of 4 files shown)


certbot/cli.py
```diff
@@ -440,8 +440,8 @@ def __init__(self, args, plugins, detect_defaults=False):
         }
 
         # List of topics for which additional help can be provided
-        HELP_TOPICS = ["all", "security", "paths", "automation", "testing"] + list(self.VERBS)
-        HELP_TOPICS += self.COMMANDS_TOPICS + ["manage"]
+        HELP_TOPICS = ["all", "security", "paths", "automation", "testing"]
+        HELP_TOPICS += list(self.VERBS) + self.COMMANDS_TOPICS + ["manage"]
```
ohemorange: did you do this to make the line lengths equal?

certbot/cli.py
```diff
@@ -850,6 +850,12 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         None, "-t", "--text", dest="text_mode", action="store_true",
         help=argparse.SUPPRESS)
     helpful.add(
+        None, "--log-backups", type=nonnegative_int, default=1000,
```
ohemorange: Naming is tough, I know, but a `-max-` might be prudent here, to make it a bit more clear that it's not how many copies of the log we're making.

certbot/cli.py
bmw: This was done in the PR by another author this PR is based on in [this commit](https://github.com/certbot/certbot/pull/4195/commits/93e814a85fb82e1cd201ce90d69828d89e52371d) because a new help topic was added in a prior commit (and later removed).

As an extremely minor positive side effect though, the line lengths are here now follow our [style](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) [guides](https://google.github.io/styleguide/pyguide.html#Line_length).

Issue #4805, opened 2017-06-08
Revoke by key should check keys


jsha:
You can revoke a certificate using the corresponding private key by passing `--key-path` to the `revoke` subcommand. However, Certbot doesn't check that the key matches the certificate. It should do so.

ynasser:
This can be accomplished using `crypto_util.verify_cert_matches_priv_key`.

Since this is a fairly self-contained task, I'm marking it as volunteer-friendly. If it needs to be done sooner for some reason, lmk and I'll take it.

Baeumla:
Since the function accepts a `storage.RenewableCert` object as a parameter, I assume that I'll have to create this object within the if statement `if config.key_path is not None:` to be able to call `crypto_util.verify_cert_matches_priv_key` to check if the key matches the certificate?

Just want to make sure that this is the right way to do it. I'm new to the project so I'm sorry if this is kind of a stupid question.

ynasser:
Hi @Baeumla 

When I originally wrote my previous comment, I wasn't paying super close attention to the types the functions accepted and didn't notice what you've pointed out. It is hypothetically possible to construct a `storage.RenewablwCert` object given only the `config` parameter, but I would not suggesting going down that path. It's much more complex than it seems :/

Instead, I think a better way to accomplish this is to modify `crypto_util.verify_cert_matches_priv_key` (and everywhere that calls it) to just accept a certificate path and private key path. That way you can call `crypto_util.verify_cert_matches_priv_key` in `main.revoke` (under the `if config.key_path is not None` branch) with just `crypto_util.verify_cert_matches_priv_key(config.cert_path[0], config.key_path[1])`.

Also, your question is not at all stupid, and please feel free to ask as many questions as you need! And sorry for the delayed response, we've all been on vacation for the past few days.

Thanks for working on this! Would it be OK if I assigned the issue to you?

Baeumla:
Okay I already thought about that but I assumed that a less-invasive method would be more appreciated. But it absolutely makes sense to do it that way. 

Yes you can assign this issue to me. I'd be happy to assist in this awesome project. I'll try to create the PR for this over the weekend.

ynasser:
OK, it turns out I can't assign the issue to you because of permissions-related reasons, but the issue is yours nonetheless :)

Thanks for helping out!

PR #4915, opened 2017-07-08
Check keys if revoke certificate by private key 

Baeumla:
Issue #4805

Revoke certificate by key now checks if the provided private key matches the certificate.

### DIFF
--- a/certbot/tests/main_test.py
+++ b/certbot/tests/main_test.py
@@ -1026,6 +1026,12 @@ class MainTest(test_util.TempDirTestCase):  # pylint: disable=too-many-public-me
                     jose.ComparableX509(cert),
                     mock.ANY)
 
+    def test_revoke_with_key_mismatch(self):
+        server = 'foo.bar'
+        self.assertRaises(errors.Error, self._call_no_clientmock,
+            ['--cert-path', CERT, '--key-path', KEY,
+                                 '--server', server, 'revoke'])
+
     @mock.patch('certbot.main._determine_account')
     def test_revoke_without_key(self, mock_determine_account):
         mock_determine_account.return_value = (mock.MagicMock(), None)
--- a/certbot/tests/main_test.py
+++ b/certbot/tests/main_test.py
@@ -37,6 +37,8 @@ CERT = test_util.vector_path('cert.pem')
 CSR = test_util.vector_path('csr.der')
 KEY = test_util.vector_path('rsa256_key.pem')
 JWK = jose.JWKRSA.load(test_util.load_vector("rsa512_key_2.pem"))
+RSA2048_KEY_PATH = test_util.vector_path('rsa2048_key.pem')
+SS_CERT_PATH = test_util.vector_path('self_signed_cert.pem')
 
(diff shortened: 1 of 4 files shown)


Baeumla:
@ynasser 

Since the original function threw the error `Details: [('SSL routines', 'SSL_CTX_check_private_key', 'no private key assigned')]` which was weird to begin with, I re-wrote the function `verify_cert_matches_priv_key` to call the OpenSSL function directly with the key and cert file path (default filetype PEM). 

Now, Travis throws this error, and I really don't know how that could happen or what I should do about it. Is it possible that the certificate really does not match the provided private key?
`Error: verifying the cert located at /home/travis/build/certbot/certbot/certbot/tests/testdata/cert.pem matches the                 private key located at /home/travis/build/certbot/certbot/certbot/tests/testdata/rsa256_key.pem has failed.                 Details: [('x509 certificate routines', 'X509_check_private_key', 'key values mismatch')]`

ynasser:
@Baeumla

Great job so far!

The problems you're experiencing stem from the fact that we have a horribly organized `certbot/tests/testdata/` directory, and it's not clear which of our testing certificates go with which keys. The reason why `test_revoke_with_key` in `certbot/tests/main_test.py` is failing is because the new verification code you wrote is in fact working properly. I don't think `cert.pem` was generated with `rsa256_key.pem`, as you've pointed out. I think that *previously* any values would have worked in that test, as the only things that mattered were reasonable-looking values existing. I do think we need some more thorough tests for revoking via a key, which is part of the issue you are working on :)

In tests which will trigger the changes you're making, I'd just copy `self_signed_cert.pem` and `rsa2048_key.pem` from `certbot/tests/crypto_util_test.py`, as for those only values which actually match can be used.

By the way, something useful to know which _may_ come up is knowing the differences between `cert`, `chain`, and `fullchain`, which is explained here: https://certbot.eff.org/docs/using.html#where-are-my-certificates.

ynasser:
FYI: I've made an issue for cleaning up + documenting `certbot/tests/testdata/` in: #4917.

certbot/tests/main_test.py
```diff
@@ -1012,12 +1014,12 @@ def test_certonly_csr_dry_run(self):
     @mock.patch('certbot.main.client.acme_client')
     def test_revoke_with_key(self, mock_acme_client):
```
ynasser: I'd add another test below this one named something like: `test_revoke_with_mismatched_key` and make sure the appropriate error is raised to make sure revoking by key is failing when it's supposed to. You can use the previous values of `CERT` and `KEY` to make it fail.

Baeumla: Yeah that makes sense. Should have thought about that before already. Sorry that it took me so long for the commit but I was having a stressful week at work.

ynasser:
This looks good to me!

@jsha since you originally requested this feature, do you want to take a quick look to make sure this is what you had in mind when you requested this feature?

ynasser:
Thanks for this @Baeumla!

Issue #4885, opened 2017-06-29
acme library: deprecate/remove phone number contact method


jaddison:
Boulder no longer supports phone number as a contact method - should `acme` be updated to remove the phone number functionality in `Registration` (ie. `NewRegistration` & `UpdateRegistration` classes)?

I ran into this while building an in-house client using `acme`. Because the documentation for `acme` is sparse, reading its source code, experimenting, and seeing how other clients (including certbot) do things is the only viable way to achieve results.

For completeness, here's the error returned from LetsEncrypt staging:
```
urn:acme:error:malformed :: The request message was malformed :: Unable to update registration :: contact method tel is not supported
```
Of course, I remove my references to 'phone' from my client implementation and things are working now - but still, it should probably be deprecated.

SwartzCr:
I bet @jsha has thoughts on this

jsha:
I think I proposed this myself a while ago, but people wanted to retain it against the possibility that other future ACME servers might want to support `tel:`. I don't feel super strongly about it right now.

SwartzCr:
I made a PR to remove this from certbot - but I don't think it's a huge issue - so I'm curious to hear what @bmw thinks

bmw:
Yeah Jacob proposed something similar in #4095 which I suggested we didn't take. While Let's Encrypt doesn't support `tel:`, it's clearly in the [ACME spec](https://tools.ietf.org/html/draft-ietf-acme-acme-07#section-7.1.2) so I think we should aim to support it.

@jaddison, if you have specific problems you're having with our [minimal acme documentation](https://acme-python.readthedocs.io/en/latest/), please open issues about it so we can work on improving it.

jaddison:
Sounds good - I was just implementing a client based on code inspection and came up against API errors because I made some assumptions. If there's a chance that phone support will be added in properly, then this is a non-issue.

PR #4924, opened 2017-07-12
remove phone number registration

SwartzCr:
addresses #4885


SwartzCr:
That's fine with me - I just made the PR because it was simple

Issue #4473, opened 2017-04-06
timeout on certbot-auto update check if pypy.python.org is unreachable


Findus23:
similar to #4074, #3313, #3636
## My operating system is (include version):
`Linux Debian-85-jessie-64-minimal 4.9.0-2-amd64 #1 SMP Debian 4.9.13-1 (2017-02-27) x86_64 GNU/Linux`

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
git clone (certbot-auto)

## I ran this command and it produced this output:
When I try to run certbot-auto nothing happens until I press CTRL+C and get:
```
Traceback (most recent call last):
  File "/tmp/tmp.bPi8RZRS1H/fetch.py", line 126, in <module>
    exit(main())
  File "/tmp/tmp.bPi8RZRS1H/fetch.py", line 114, in main
    print latest_stable_version(get)
  File "/tmp/tmp.bPi8RZRS1H/fetch.py", line 71, in latest_stable_version
    'https://pypi.python.org/pypi/certbot/json')))
  File "/tmp/tmp.bPi8RZRS1H/fetch.py", line 56, in get
    return self._opener.open(url).read()
  File "/usr/lib/python2.7/urllib2.py", line 429, in open
    response = self._open(req, data)
  File "/usr/lib/python2.7/urllib2.py", line 447, in _open
    '_open', req)
  File "/usr/lib/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/urllib2.py", line 1241, in https_open
    context=self._context)
  File "/usr/lib/python2.7/urllib2.py", line 1201, in do_open
    r = h.getresponse(buffering=True)
  File "/usr/lib/python2.7/httplib.py", line 1121, in getresponse
    response.begin()
  File "/usr/lib/python2.7/httplib.py", line 438, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python2.7/httplib.py", line 394, in _read_status
    line = self.fp.readline(_MAXLINE + 1)
  File "/usr/lib/python2.7/socket.py", line 480, in readline
    data = self._sock.recv(self._rbufsize)
  File "/usr/lib/python2.7/ssl.py", line 766, in recv
    return self.read(buflen)
  File "/usr/lib/python2.7/ssl.py", line 653, in read
    v = self._sslobj.read(len)
KeyboardInterrupt
```
[...]

PR #4929, opened 2017-07-14
Add timeout to certbot-auto HTTPS fetches #4473

cj-dev:


cj-dev:
Addresses [4473](https://github.com/certbot/certbot/issues/4473) by adding a timeout argument to `HttpsGetter.get` in `fetch.py`. The default value mirrors `OpenerDirector.open`'s signature with `socket._GLOBAL_DEFAULT_TIMEOUT`.

This timeout applies to connections and requests. I set the timeout to 10 seconds for the requests in `fetch.py` because it "felt" right.

bmw:
@erikrose, do you have time to review this?

erikrose:
In the next week or two I should. I have a few other certbot-related things in my queue atm.

pde:
@erikrose nudge :)

erikrose:
It's in the queue, but more important things have thus far taken precedence. :-)

cowlicks:
I'll look at this today @cj-dev

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -49,14 +50,14 @@ def __init__(self):
             if isinstance(handler, HTTPHandler):
                 self._opener.handlers.remove(handler)
 
-    def get(self, url):
+    def get(self, url, timeout=socket._GLOBAL_DEFAULT_TIMEOUT):
```
cowlicks: @cj-dev why not use socket.getdefaulttimeout()?

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -49,14 +50,14 @@ def __init__(self):
...
         """Return the document contents pointed to by an HTTPS URL.
 
         If something goes wrong (404, timeout, etc.), raise ExpectedError.
 
         """
         try:
-            return self._opener.open(url).read()
+            return self._opener.open(url, timeout = timeout).read()
         except (HTTPError, IOError) as exc:
```
cowlicks: I think you should add `urllib2.URLError` here.

letsencrypt-auto-source/pieces/fetch.py
cj-dev: I remember pondering this for a while, but I'm embarrassed I didn't take complete notes on my rationale. A few python issues seem to indicate this is a long-standing debacle [1][2][3] that leaves everyone a little dissatisfied. If I understand correctly, `_GLOBAL_DEFAULT_TIMEOUT` rose to prominence as a sentinel to distinguish between `None` meaning "revert to default" (old) and `None` meaning "no timeout" (present).

The python socket documentation states the default returned by getdefaulttimeout is always going to be `None`, presumably until it is set using setdefaulttimeout [4], which I think would be least useful.

So my justification boiled down to `HttpsGetter` being not much more than a fit-for-purpose `urllib2.urlopen`. I thought that the safest thing to do in my confusion would be to mimic the `urllib2.urlopen` and `OpenerDirector.open` method signatures. When I was looking for precedent to contradict my plan, I didn't see any python in the stdlib that actually uses `getdefaulttimeout`, outside of tests.

If you have a preference or think I missed the mark, just say the word.

1. https://bugs.python.org/issue18417
2. https://bugs.python.org/issue4079
3. https://bugs.python.org/issue14425
4. https://docs.python.org/2/library/socket.html?highlight=socket#socket.getdefaulttimeout

letsencrypt-auto-source/pieces/fetch.py
cowlicks: Suggesting because, when I do `.open(url, timeout=some_super_short_timeout)` it throws a URLError.

letsencrypt-auto-source/pieces/fetch.py
cowlicks: Nope! I think you are correct. That is certainly the default value for `timeout` kwarg of `OpenerDirector.open`.  cpython 2.7 source [here](https://github.com/python/cpython/blob/a72d15c97f9837f6c9c6bd476a62175c942cc588/Lib/urllib2.py#L411), master source [here](https://github.com/python/cpython/blob/a72d15c97f9837f6c9c6bd476a62175c942cc588/Lib/urllib2.py#L411).

I was worried this might break some non-cpython pythons, but pypy does it too, and idk if we even need to worry about non-cpython pythons. 👍

letsencrypt-auto-source/pieces/fetch.py
cj-dev: I agree. I just got pulled away from keyboard before pushing.

URLError is a subclass of IOError; Are you concerned about an uncaught exception or is the idea to simply catalogue all the likely exception classes?

erikrose: If it's a subclass, we should be all set.

erikrose:
Made a few tweaks and continued in #5118. Thanks for the start, @cj-dev! If you have thoughts on my changes, please share.

Issue #4866, opened 2017-06-21
Hide pip upgrade notices in certbot-auto


bmw:
Pretty sure this is as simple as including `--disable-pip-version-check` on the command line and reduces a lot of visual noise when debugging installation problems.

SwartzCr:
fixed in #4941

PR #4941, opened 2017-07-17
add disable pip version check flag

SwartzCr:
solves #4866


letsencrypt-auto-source/pieces/bootstrappers/mac.sh
```diff
@@ -39,6 +39,6 @@ BootstrapMac() {
   if ! hash virtualenv 2>/dev/null; then
     say "virtualenv not installed."
     say "Installing with pip..."
-    pip install virtualenv
+    pip install --disable-pip-version-check virtualenv
```
bmw: This change worries me a bit. It might be fine, however, we're assuming that the version of `pip` on the user's system is new enough to support `--disable-pip-version-check`. I don't think we can safely make this assumption and since we're deprecating our experimental `certbot-auto` support for macOS in #4943, I think we can just leave things as they've always been here.

SwartzCr: done

PR #4943, opened 2017-07-17
add deprecation warning

### LINKED ISSUE #4520: Print deprecation warning on experimental systems in certbot-auto

SwartzCr:
fixes #4520

### DIFF
--- a/certbot-auto
+++ b/certbot-auto
@@ -200,6 +200,24 @@ ExperimentalBootstrap() {
   fi
 }
 
+DeprecationBootstrap() {
+  # Arguments: Platform name, bootstrap function name
+  if [ "$DEBUG" = 1 ]; then
+    if [ "$2" != "" ]; then
+      BootstrapMessage $1
+      $2
+    fi
+  else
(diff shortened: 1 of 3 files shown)


SwartzCr:
I just listed the ones you mentioned in the issue, happy to do some investigation if you'd like

bmw:
I think the wording is pretty good. It's a little harsher than need be I think because we haven't needed to make changes to those bootstrap scripts for a long time now, but making people switch sooner rather than later isn't a bad thing.

It'd be good if we investigated the other OSes, but we can merge this as a start if you'd like. Just make sure #4520 isn't closed.

Issue #4946, opened 2017-07-18
Developer guide should mention extra installation instructions for Sphinx on MacOS


ynasser:
>  When I was generating the Sphinx documentation on my macOS with the `make -C docs clean html` man command, I was having import errors. I had to manually install `repoze.sphinx.autointerface` and `sphinx_rtd_theme`. I think this should be mentioned in the Developer Guide.
> 
> To install those packages I used the pip installer.
> ```
> pip install repoze.sphinx.autointerface
> pip install sphinx_rtd_theme 
> ```
> 

(Taken from https://github.com/certbot/certbot/issues/4917#issuecomment-315536077)

scottilee:
I verified this issue occurs on both a Mac running Sierra and an Ubuntu 14.04 VM. I have a pull request here: https://github.com/certbot/certbot/pull/4949.

ynasser:
Thanks @scottilee! One of us will take a look soon.

bmw:
See https://github.com/certbot/certbot/pull/4949#issuecomment-316783263.

PR #3985, opened 2017-01-08
Add Dependencies to generate Sphinx documentation

nickfong:
Specify that developers should install `Sphinx`, `sphinx_rtd_theme`, and `repoze.sphinx.autointerface` to build the documentation

See #3980

PR #4949, opened 2017-07-19
[#4946] Update Sphinx installation docs

scottilee:
I verified this issue occurs on both a Mac running Sierra and an Ubuntu 14.04 VM.

Generated the docs to see my changes as well.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -362,8 +362,14 @@ during the next release.
 Updating the documentation
 ==========================
 
-In order to generate the Sphinx documentation, run the following
-commands:
+In order to generate the Sphinx documentation, you will need to install some tools:
+
+.. code-block:: shell
+
+   pip install repoze.sphinx.autointerface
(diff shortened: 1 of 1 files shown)


bmw:
So I'm not sure if this is necessary. If you follow our [developer installation instructions](https://certbot.eff.org/docs/contributing.html#getting-started), these packages will be installed because [tools/venv.sh installs our docs extras](https://github.com/certbot/certbot/blob/bb6a22b9853cfab2a06cb020ddf57dd5a6ae8eba/tools/venv.sh#L15) which [includes these packages](https://github.com/certbot/certbot/blob/bb6a22b9853cfab2a06cb020ddf57dd5a6ae8eba/setup.py#L77).

With that said, people have hit this problem before (see #3985). Perhaps we should instead add a line saying that if you didn't follow our installation instructions in the "Getting Started" section above, make sure that you have installed Certbot's docs extras.

PR #4956, opened 2017-07-25
Change add_server_directives replace=True behavior to attempt to replace, but append on failure to find.

### LINKED ISSUE #4832: Interpret replace=True as "replace if possible, otherwise append"
https://community.letsencrypt.org/t/permissions-required-to-etc-letsencrypt-live-for-script/35645/6

In Nginx, say someone's done something like this:

```
server {
  server_name perfectly-good-name.com;
  listen 443 ssl;
  # ssl_certificate /etc/letsencrypt/live/example.org/fullchain.pem;
  # ssl_certificate_key /etc/letsencrypt/live/example.org/privkey.pem;
}
```

They know where the certs will live, so they've put in links to them, but Nginx needs real files there, so they comment them out.

But now the installer's going to fail, because we make the assumption that if someone has a server block with ssl enabled, we shouldn't add any ssl directives because we should try to leave things as much the same as possible.

I propose changing this assumption from an empirical perspective. People run into this all the time, so therefore it's the workflow that people expect should work. Therefore we should support it.

I'd love to hear any reasons why we didn't do this previously, or if we treat this differently in Apache.

ohemorange:
Fixes #4832.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -198,16 +198,10 @@ class NginxConfigurator(common.Plugin):
         cert_directives = [['\n', 'ssl_certificate', ' ', fullchain_path],
                            ['\n', 'ssl_certificate_key', ' ', key_path]]
 
-        try:
-            self.parser.add_server_directives(vhost,
-                                              cert_directives, replace=True)
-            logger.info("Deployed Certificate to VirtualHost %s for %s",
-                        vhost.filep, vhost.names)
-        except errors.MisconfigurationError as error:
-            logger.debug(error)
-            # Presumably break here so that the virtualhost is not modified
-            raise errors.PluginError("Cannot find a cert or key directive in {0} for {1}. "
-                "VirtualHost was not modified.".format(vhost.filep, vhost.names))
+        self.parser.add_server_directives(vhost,
+                                          cert_directives, replace=True)
+        logger.info("Deployed Certificate to VirtualHost %s for %s",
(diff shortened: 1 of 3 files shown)


ohemorange:
I think the error is still accurate -- it may not have been found for other reasons, like a malformed server block. `add_server_directives` still raises a `MisconfigurationError`, so we should still catch it and turn it into a `PluginError` for consistency.

bmw:
> I think the error is still accurate -- it may not have been found for other reasons, like a malformed server block.

I completely agree that other problems may occur, however, not finding the directives should no longer be the cause of a `MisconfigurationError` as we'll just try to add the missing directives. I think we should either:

1. Remove the `try`/`except` block. `MisconfigurationError` is a subclass of `PluginError` which we already make use of in the same method. We call [choose_vhost](https://github.com/certbot/certbot/blob/fc097de5ffc9294d59fab674d1f0a400a0d2b2d0/certbot-nginx/certbot_nginx/configurator.py#L197) without handling any exceptions which [raises a MisconfigurationError](https://github.com/certbot/certbot/blob/fc097de5ffc9294d59fab674d1f0a400a0d2b2d0/certbot-nginx/certbot_nginx/configurator.py#L246) if it can't find a vhost.
2. Replace the error message with something more generic like `Unable to set cert and key directive in {0} for {1}. Your server block was not modified.`

ohemorange:
Ah, didn't realize `MisconfigurationError` is a subclass of `PluginError`, that makes me much more comfortable about removing it.

Issue #4932, opened 2017-07-15
Print cert name in renewal errors


jsha:
Some errors during renewal print the path to a certificate, e.g.:

```
Attempting to renew cert from /etc/letsencrypt/renewal/jacob.hoffman-andrews.com-0001.conf produced an unexpected error: Account at /etc/letsencrypt/accounts/acme-staging.api.letsencrypt.org/directory/06c4ea2fe30100e185fea43f26817312 does not exist. Skipping.
```

It would be nice to also print the cert name, since that is the identifier used to delete certs through Certbot.

r5d:
Does this look like the right way to do it?

```diff
diff --git a/certbot/renewal.py b/certbot/renewal.py
index 5c669794..d5d4b2ce 100644
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -389,14 +389,17 @@ def handle_renewal_request(config):
         disp = zope.component.getUtility(interfaces.IDisplay)
         disp.notification("Processing " + renewal_file, pause=False)
         lineage_config = copy.deepcopy(config)
+        lineagename = storage.lineagename_for_filename(renewal_file)
 
         # Note that this modifies config (to add back the configuration
         # elements from within the renewal configuration file).
         try:
             renewal_candidate = _reconstitute(lineage_config, renewal_file)
         except Exception as e:  # pylint: disable=broad-except
-            logger.warning("Renewal configuration file %s produced an "
-                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.warning("Renewal configuration file %s (cert: %s) "
+                           "produced an "
+                           "unexpected error: %s. Skipping.", lineagename,
+                               renewal_file, e)
             logger.debug("Traceback was:\n%s", traceback.format_exc())
             parse_failures.append(renewal_file)
             continue
@@ -422,8 +425,9 @@ def handle_renewal_request(config):
                     renew_skipped.append(renewal_candidate.fullchain)
         except Exception as e:  # pylint: disable=broad-except
             # obtain_cert (presumably) encountered an unanticipated problem.
-            logger.warning("Attempting to renew cert from %s produced an "
-                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.warning("Attempting to renew cert (%s) from %s produced an "
+                           "unexpected error: %s. Skipping.", lineagename,
+                               renewal_file, e)
             logger.debug("Traceback was:\n%s", traceback.format_exc())
[...]

PR #4957, opened 2017-07-25
Add cert name in renewal error messages (#4932)

r5d:
Update `certbot.renewal.handle_renewal_request` to print cert name in error messages.

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -389,14 +389,16 @@ def handle_renewal_request(config):
         disp = zope.component.getUtility(interfaces.IDisplay)
         disp.notification("Processing " + renewal_file, pause=False)
         lineage_config = copy.deepcopy(config)
+        lineagename = storage.lineagename_for_filename(renewal_file)
 
         # Note that this modifies config (to add back the configuration
         # elements from within the renewal configuration file).
         try:
             renewal_candidate = _reconstitute(lineage_config, renewal_file)
         except Exception as e:  # pylint: disable=broad-except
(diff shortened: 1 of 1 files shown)


jsha:
@ohemorange can you take a look since you worked on the lineage stuff?

ohemorange:
What's the intended goal of this PR? The cert name is by definition the renewal config filename minus the `.conf`. Were you looking to add the domain names covered by the cert?

jsha:
See my reasoning in https://github.com/certbot/certbot/issues/4932. I wanted to delete a certificate because it was failing renewal. IMO it's not immediately clear that the certificate name is part of the filename.

certbot/renewal.py
```diff
@@ -389,14 +389,16 @@ def handle_renewal_request(config):
...
         # Note that this modifies config (to add back the configuration
         # elements from within the renewal configuration file).
         try:
             renewal_candidate = _reconstitute(lineage_config, renewal_file)
         except Exception as e:  # pylint: disable=broad-except
-            logger.warning("Renewal configuration file %s produced an "
-                           "unexpected error: %s. Skipping.", renewal_file, e)
+            logger.warning("Renewal configuration file %s (cert: %s) "
+                           "produced an unexpected error: %s. Skipping.",
+                           lineagename, renewal_file, e)
```
ohemorange: `lineagename` and `renewal_file` should be reversed

ohemorange:
Ah, that makes sense!

certbot/renewal.py
r5d: Done.

PR #4960, opened 2017-07-25
Explicitly advise against using easy_install.

bmw:
Created in response to #4934.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -234,11 +234,10 @@ Installing from source
 Installation from source is only supported for developers and the
 whole process is described in the :doc:`contributing`.
 
-.. warning:: Please do **not** use ``python setup.py install`` or
-   ``python pip install .``. Please do **not** attempt the
-   installation commands as superuser/root and/or without virtual
-   environment, e.g. ``sudo python setup.py install``, ``sudo pip
-   install``, ``sudo ./venv/bin/...``. These modes of operation might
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -234,11 +234,10 @@ Installing from source
...
 
-.. warning:: Please do **not** use ``python setup.py install`` or
-   ``python pip install .``. Please do **not** attempt the
-   installation commands as superuser/root and/or without virtual
-   environment, e.g. ``sudo python setup.py install``, ``sudo pip
-   install``, ``sudo ./venv/bin/...``. These modes of operation might
-   corrupt your operating system and are **not supported** by the
-   Certbot team!
+.. warning:: Please do **not** use ``python setup.py install``, ``python pip
+   install .``, or ``easy_install .``. Please do **not** attempt the
```
SwartzCr: github is showing that the word install isn't in the ``
I don't know if this merely a syntax hilighting issue but how sure are you that ``s work across line breaks?

bmw: Very sure. It works correctly when building the docs locally and [this section already has double backticks across multiple lines](https://github.com/certbot/certbot/pull/4960/files?diff=unified#diff-62c330cd8da31d476bb63b51bb8b1307L240) in `master` which is [displayed properly](https://certbot.eff.org/docs/install.html#installing-from-source).

Issue #4071, opened 2017-01-17
patch_get_utility problems
Search
￼


bmw:
```
In [2]: from certbot.tests.util import patch_get_utility

In [3]: with patch_get_utility():
   ...:     print zope.component.getUtility("foo")
   ...:     zope.component.getUtility.return_value = 42
   ...:     print zope.component.getUtility("foo")
   ...:     
<certbot.tests.util.FreezableMock object at 0x7f664a372450>
42
```
We shouldn't let people do that.

bmw:
The purpose of `patch_get_utility()` is described in more detail in #4010. To give a short summary, we have a lot of places in the code where we do things like [this](https://github.com/certbot/certbot/blob/v0.14.1/certbot/main.py#L263). We use `zope.component.getUtility` to get a display object and then do some kind of user interaction.

The problem is that during our unit tests, these utilities and displays aren't set up so we mock the call. This is fine, however, there are a number of optional parameters to these display objects and some combinations aren't valid. By mocking the call, these assertions never run and we've had code that fails assertions land in `master` because the display object was never properly called because it was mocked out during tests.

`patch_get_utility` mocks out the `getUtility` call, but also runs the assertions on the parameters. The problem identified in the original post of this issue is there is a way `patch_get_utility` causes all of this special functionality to be ignored. This should be fixed.

cj-dev:
I'm new here, so excuse me if I'm way off. As I understand the mock objects used for the display...

`tests.util.patch_get_utility` mocks the `getUtility` function by patching it with a call to `tests.util._create_get_utility_mock` which returns a `tests.util.FreezableMock`. The FreezableMock object itself doesn't have any assertions (its `_func` is None) but its attributes inherited from `interfaces.IDisplay` (other than 'notification') have their parameters checked by `display.util.assert_valid_call`, containing an assertion.

What you showed in this thread's original post, `zope.component.getUtility.return_value = 42`, is legitimate mock behavior, right? And from my understanding above, I'd not expect the assertion function to be called when assigning to `return_value`. Are you suggesting that the mocked-out `getUtility` shouldn't have a `return_value` attribute that can be set after-the-fact? Maybe I need to see examples of the other code you mention that fails assertions getting into `master`.

```
(Pdb) zope.component.getUtility('foo').yesno('yes or no', default='yes')
<MagicMock name='mock()' id='140186259122896'>
(Pdb) zope.component.getUtility('foo').yesno('yes or no')
[...]

PR #4962, opened 2017-07-26
[WIP] Addressing #4071 by preventing FreezableMock's return_value from being set

cj-dev:
First of two candidate solutions for #4071. This one is simply a throwaway `if` block.

This isn't perfect because while it will prevent changing the `return_value` attribute on the FreezableMock objects, it won't prevent changing the return value of the patched `zope.component.getUtility` because it is a `mock.MagicMock`.

So I consider this incomplete. I think I'll come back to this issue with fresher eyes, or maybe someone else with sicker Python skills can sweep it.


cj-dev:
Closing because better discussion and work will continue in #4963

PR #4963, opened 2017-07-26
#4071 Mixin to prevent setting return_value after initializing certain Mock objects

cj-dev:
Second of two candidate solutions to #4071. This involves an `ImmutableReturnMixin` to short-circuit attempts to set "return_value" on certain Mock objects which would otherwise bypass validity checks.

This covers more than the first candidate because the Mixin can be inherited alongside `mock.MagicMock` to create a true Mock object which could be used as the patched `getUtility`. This would cover the original case outlined in #4071 that PR #4962 would not.

However, inheriting from Mock or overriding is `__setattr__` gets weird fast and I've been trying to grok the unittest.Mock source to extend FreezableMock with limited success. One case of weird behavior is that calling a `FreezableMock`'s _mock attribute twice changes the returned value's `__str__` to an `ImmutableReturnMock`. I don't yet understand why.


cj-dev:
I had misunderstood the `FreezableMock` mechanics, but now that you mention it, it's clear that `FreezableMock` can replace the `MagicMock` returned by _create_get_utility_mock. I created a world of hurt for myself with that misunderstanding. How embarrassing.

The default in this PR is still to allow setting return_value because numerous tests take advantage of setting return_value on these objects, for better or for worse. I haven't inspected all the tests to find ways to update them for use with this patch.

https://github.com/certbot/certbot/blob/master/certbot/tests/main_test.py#L1141, https://github.com/certbot/certbot/blob/master/certbot/plugins/selection_test.py#L147, and ~25 others now fail due to uncaught AttributeError. Thought I'd maintain compatibility.

cj-dev:
> it's OK if the return value of the IDisplay methods is set by the user.
That makes all the difference in making implementation and compatibility with existing tests much easier. Tests were already taking advantage of changing `IDisplay` return values and I wouldn't know where to begin if I needed to refactor that.

I'm not in love with how I basically used a couple `if` blocks to make `FreezableMock` play the way we want it to. The other way I might have implemented this is by assigning return_value to the `FreezableMock` instance instead of the `_mock` and short-circuiting the `__call__`, returning `return_value` instead of calling the underlying `_mock`. I vetoed that approach because some tests both set `return_value` and assert that the mock object was called. I think it's neater to just pass that on down to the `_mock`. By the same token, I passed the `side_effect`s through. They would not occur if set on the `FreezableMock`.

certbot/tests/util.py
```diff
@@ -208,21 +208,36 @@ def __getattribute__(self, name):
...
             self._frozen_set.add(name)
         return object.__setattr__(self, name, value)
 
-
 def _create_get_utility_mock():
     display = FreezableMock()
     for name in interfaces.IDisplay.names():  # pylint: disable=no-member
         if name != 'notification':
-            frozen_mock = FreezableMock(frozen=True, func=_assert_valid_call)
+            frozen_mock = FreezableMock(frozen=False, func=_assert_valid_call)
```
bmw: This mock object is now no longer frozen when we return it. In practice, this probably doesn't matter because the only attribute we defined was `_func` and it'd be very unusual for a user of this function to change this value.

With that said, I see no reason it can't be frozen. What if we changed the definition of `__setattr__` above to something like:
```
def __setattr__(self, name, value):
    """ Set attributes on the underlying _mock if the FreezableMock is frozen.

    In cases of return_value and side_effect, these attributes are always passed
    through to the instance's _mock.

    """
    if self._frozen:
        if name in self._frozen_set:
            raise AttributeError('Cannot change frozen attribute ' + name)
        else:
            return setattr(self._mock, name, value)
   
    if name != '_frozen_set':
        self._frozen_set.add(name)

    if name in ('return_value', 'side_effect',):
        return setattr(self._mock, name, value)
    else:
        return object.__setattr__(self, name, value)
```
With this approach, the way a `FreezableMock` object works is still consistent. We do treat `return_value` and `side_effect` specially, but I personally think this is just making good use of the tools available to us to avoid reinventing the wheel than an ugly hack.

Note: With this approach, the `return_value` parameter to `__init__` would have to be modified slightly. I think this can be done by setting `self.return_value = return_value` before `self._frozen` is defined but after defining `self._mock`.

cj-dev:
I've incorporated the feedback you suggested. It's almost verbatim copy+paste. Sorry you ended up doing most of the critical thinking; I was slow on getting the spirit of the task and precedent, despite your explanations that someone else probably could have run with. Next time, I will ask more questions up-front.

certbot/tests/util.py
```diff
@@ -208,21 +209,35 @@ def __getattribute__(self, name):
...
+        else:
+            return object.__setattr__(self, name, value)
 
 
 def _create_get_utility_mock():
     display = FreezableMock()
     for name in interfaces.IDisplay.names():  # pylint: disable=no-member
         if name != 'notification':
-            frozen_mock = FreezableMock(frozen=True, func=_assert_valid_call)
+            frozen_mock = FreezableMock(frozen=False, func=_assert_valid_call)
```
bmw: We should change the `frozen` value back to `True`.

cj-dev:
I don't feel slighted at all; no need to apologize. Glad we can collaborate to get stuff done.

I think freezing the `IDisplay`s isn't right... You commented earlier that it's okay if the return_value on `IDisplay` objects is set by the user. If the `IDisplay`s and the mock `getUtility` are both `FreezableMock`, freezing must disallow changing the `return_value` of a `FreezableMock` to address the original intent of the PR, which was to prevent developers from reassigning the `getUtility` return_value to vanilla Mock objects, which would nullify any `_func`. Freezing the `getUtility` mock accomplishes the original intent. Meanwhile, the `IDisplay` objects must allow changing their `return_value` for meaningful tests*, so they must not be frozen at initialization.

Here's roughly what I'm doing to check the criteria:
```
import zope
from certbot.tests.util import patch_get_utility
from certbot import interfaces
import pdb
with patch_get_utility():
    get = zope.component.getUtility #Frozen Freezable
    foo = zope.component.getUtility("foo") #Frozen Freezable

    try: # Expect key error raised by _func
        print "foo.yesno(): %s" %(foo.yesno())
    except KeyError:
        print "Got expected KeyError invoking foo.yesno()"
    print "Correctly invoked foo.yesno('okay', default=True): %s" %(foo.yesno('okay', default=True))

    try: # Expect this to fail. This is the original case outlined in issue. get represents the mock getUtility
        get.return_value = 1
    except AttributeError:
        print "Got expected attributeError during: %s" %("get.return_value = 1")
    else:
        raise Exception("Succeeded setting return value on %s" %("get.return_value = 1"))

    try: # Expect this to fail too. foo is the display returned by getUtility which presents IDisplay objects to developer
        foo.return_value = 1
    except AttributeError:
        print "Got expected attributeError during: %s" %("foo.return_value = 1")
    else:
        raise Exception("Succeeded setting return value on %s" %("foo.return_value = 1"))

    try: # But changing the return_value on an IDisplay mock should work.
        foo.yesno.return_value = 1
    except AttributeError:
        raise Exception("Should not get an attributeError during: %s" %("foo.yesno.return_value = 1"))
    else:
        print "Succeeded setting return value on %s" %("foo.yesno.return_value = 1")
```

* examples
https://github.com/certbot/certbot/pull/4963/files#diff-09e9656fd46b31cbd3ba087797b33188R315
https://github.com/certbot/certbot/pull/4963/files#diff-9db063ec1d30d8d2352b0fd1c2a5c97bR136

bmw:
> Are you sure about freezing the IDisplays? You commented earlier that it's okay if the return_value on IDisplay objects is set by the user. Since freezing forbids changing return_value, doesn't it need to stay un-frozen for those cases*?

Ah. It wasn't my intent that freezing forbids changing the return value. Ideally what I'd like to happen is changing the return value is only forbidden if the `return_value` is set before the object is frozen.

In general, this is the way I initially wanted `FreezableMock` objects to work. After the object is frozen, all previously set properties cannot be changed while all other values can be changed. We almost have this behavior, but because we set `self.return_value` in `__init__`, `return_value` is added to the `_frozen_set` and then cannot be changed once the object is frozen. Instead, I think we should make the default value for `return_value` in `__init__` be `None` and only set the property on the `FreezableMock` if it is set by the caller.

This should largely fix the problem, however, one of your tests above fails with this approach:
```

    try: # Expect this to fail too. foo is the display returned by getUtility which presents IDisplay objects to developer
        foo.return_value = 1
    except AttributeError:
        print "Got expected attributeError during: %s" %("foo.return_value = 1")
    else:
        raise Exception("Succeeded setting return value on %s" %("foo.return_value = 1"))
```
This is because we never set a return value for a call to the display object. I think this is fine because such a call is meaningless and this is the normal behavior of `mock` objects, however, we could prevent it if you wanted by setting `side_effect` or `return_value` in `patch_get_utility`.

cj-dev:
I modified `__init__` to let us have our cake and eat it too. Instead of defaulting to `None`, continue to default to mock's sentinel because in this pattern we need to detect if the argument was changed or if the caller passed `None` as the argument. If the caller passed a return value deliberately, then `return_value` is added to the `_frozen_set` and set with the supplied value, otherwise it remains undefined and out of the set.

Another dirty sanity check to help you see what I'm thinking.
`
from certbot.tests.util import FreezableMock

return_value_set = FreezableMock(frozen=True, return_value=1)
print return_value_set._frozen_set # set(['_func', 'return_value', '_mock'])
try: # Expect this to fail
    return_value_set.return_value = 2
except AttributeError:
    print "Got expected attributeError during: %s" %("return_value_set.return_value = 2")
else:
    raise Exception("Succeeded setting return value on %s" %("return_value_set.return_value = 2"))

return_value_unset = FreezableMock(frozen=True)
print return_value_unset._frozen_set # set(['_func', '_mock'])
print return_value_unset() # mock obj

try: # Expect this to succeed
    return_value_unset.return_value = 2
except AttributeError:
    raise Exception("Should not get an AttributeError during: %s" %("return_value_unset.return_value = 2"))
else:
    print "Succeeded setting return value on %s" %("return_value_unset.return_value = 2")
try: # Expect this to succeed
    return_value_unset.return_value = 3
except AttributeError:
    raise Exception("Should not get an AttributeError during: %s" %("return_value_unset.return_value = 3"))
else:
    print "Succeeded setting return value on %s" %("return_value_unset.return_value = 3")

print return_value_unset() # 3
`

PR #4969, opened 2017-07-27
Unhide Nginx

### LINKED ISSUE #4079: Make nginx a first class plugin!

bmw:
Fixes #4079. This is #4543 with a couple of changes.

First, I removed our nginx plugin's `hidden` attribute so it is found during plugin discovery.

Second, this code is now only used when using `certbot-auto`. In my opinion, having to do this is at all a wart and I'd like to make its scope as small as possible. OS packages will get the normal experience which I believe is safe because for this to be a breaking change, they'd have to have both web servers installed, both of our plugins installed (which are always packaged separately), and be running `certbot` non-interactively without setting `-n`. I think it's highly unlikely (m)any of these users exist. If they do, we have timeouts on our interactions so Certbot will exit with a warning rather than hanging indefinitely on their server.

Using the `CERTBOT_AUTO` environment variable is a reliable way to check if we're being run from `certbot-auto` as it was added in our 0.6.0 release before we ever started shipping the nginx plugin. Old `letsencrypt-auto` users that don't receive updates won't have the nginx plugin installed.

### DIFF
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -112,7 +112,7 @@ def choose_plugin(prepared, question):
 
     while True:
         disp = z_util(interfaces.IDisplay)
-        if names == set(("apache", "nginx")):
+        if "CERTBOT_AUTO" in os.environ and names == set(("apache", "nginx")):
             # The possibility of being offered exactly apache and nginx here
             # is new interactivity brought by https://github.com/certbot/certbot/issues/4079,
             # so set apache as a default for those kinds of non-interactive use
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -108,11 +108,19 @@ def choose_plugin(prepared, question):
     opts = [plugin_ep.description_with_name +
             (" [Misconfigured]" if plugin_ep.misconfigured else "")
             for plugin_ep in prepared]
+    names = set(plugin_ep.name for plugin_ep in prepared)
 
     while True:
         disp = z_util(interfaces.IDisplay)
-        code, index = disp.menu(
-            question, opts, force_interactive=True)
+        if names == set(("apache", "nginx")):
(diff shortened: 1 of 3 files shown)


certbot/plugins/selection.py
```diff
@@ -108,11 +108,19 @@ def choose_plugin(prepared, question):
     opts = [plugin_ep.description_with_name +
             (" [Misconfigured]" if plugin_ep.misconfigured else "")
             for plugin_ep in prepared]
+    names = set(plugin_ep.name for plugin_ep in prepared)
 
     while True:
         disp = z_util(interfaces.IDisplay)
-        code, index = disp.menu(
-            question, opts, force_interactive=True)
+        if "CERTBOT_AUTO" in os.environ and names == set(("apache", "nginx")):
```
pde: It may be slightly better to use `certbot.cli.cli_command == certbot.cli.LEAUTO`, in case we're somehow in a really, really old non-updating `letsencrypt-auto` venv?

bmw: I just want to verify, did you read what I wrote in the original description of this PR?

Since the very old non-updating `letsencrypt-auto` doesn't install the nginx plugin, this case is almost the same as OS packages. The user would need both web servers installed, both plugins installed, and be running Certbot without setting `-n`. The one difference here is we already installed the Apache plugin in `letsencrypt-auto`. I think this case is fairly unlikely, however, I don't have a strong opinion and if you'd still like me to make this change I'm happy to do so.

pde: Nope, I missed that. I think perhaps there's an argument for having One Official Way™ to check if we're `certbot-auto` or not, but that should probably be a function not a magic test, and doesn't need to be in this PR.

pde:
Sorry for the slow turnaround here while I've been travelling. A pity this missed the 0.17 train :(

Issue #3522, opened 2016-09-21
certbot-auto-cron


bmw:
It'd be nice if we could help people set up cron or systemd timers to automatically run `certbot renew -q`. Ideally, for packaged versions of Certbot, this job should be left up to the package maintainer. For `certbot-auto` users, we could offer to try and set up the cronjob or systemd timer for them.

If we do this, we should probably only do so when installing OS packages (which should only happen once per system) and we should prompt the user to ask permissions (unless they included something like `--noninteractive` on the command line). We should avoid causing `certbot-auto` to be run automatically on systems where Certbot is already installed and there is an existing integration around `certbot-auto`.

Putting this behavior in `certbot-auto` has a couple of downsides:
1. It doesn't help people installing directly through `pip` (which we never tell anyone to do).
2. Some distributions may not create a crontab for us due to ideological reasons or because they don't know they're supposed to. This will leave these users to deal with their own cron/systemd setup.

With that said, I personally prefer putting this behavior in `certbot-auto` rather than the core Certbot client, but I want to make sure we know the limitations here up front.

Related: #880

pde:
Alternative option: offer to do this setup from within the client, after successfully obtaining a cert?

bmw:
Yeah. I mentioned that option above. I personally think it's cleaner to put it in `certbot-auto` as installing a standard crontab for a package is the role of a package manager, but we could theoretically help more people by putting it in the core client. I put in the list above the two groups of people we may miss by putting the functionality into `certbot-auto`.

pde:
My view is that it would be saner to implement this in Python code rather than in `certbot-auto`. Basically, what we would do at the end of `certonly` or `run` is walk through all of the known locations for Certbot cron and systemd configurations, and if all of them are missing, ask an interactive question offering to create one that looks correct for your system.  That would work for:

* All of the `certbot-auto` users
* The OSes that don't / won't have this included in their packaging

This shouldn't interfere with:
* People whose packages provide this now (or in the future, if packing is done well...)
* People who don't want it (they can say no)

ynasser:
Rough plan for this so far (all this code will be in `certbot-auto`, not the core client, as using bash makes more sense for this):

0. Add cron as a dependency to certbot-auto.

1. Ask user if we want to setup a cron job for renewal. If so, proceed to # 2. If not, exit with a message that contains a link to directions on how to setup up cron themselves.

[...]

PR #4970, opened 2017-07-27
Change certbot-auto's installation path to /opt

### LINKED ISSUE #3480: certbot-auto doesn't appear to respect HTTPS_PROXY env var
Hello, thanks for certbot! I'm attempting to use it behind an https proxy, and it hangs after asking me for my email address, eventually erroring out with this message:

```
ConnectionError: HTTPSConnectionPool(host='acme-v01.api.letsencrypt.org', port=443): Max retries exceeded with url: /directory (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f7f4f42f510>: Failed to establish a new connection: [Errno 110] Connection timed out',))
```

I'm fairly confident that this is due to it not using the https proxy set in both `https_proxy` and `HTTPS_PROXY` environment variables. Any ideas?

bmw:
In general, the purpose of this PR is to move `certbot-auto`'s install path to a single location to help people who have multiple installations and prep for #3522. This also fixes #3480.

I've broken design decisions into sections so reviewers or others in the future can read why we made specific decisions that they may be interested in. This PR should probably get 2 reviews due to its risk of breaking all the things.

#### Why /opt/eff.org/certbot/venv?

Per my understanding of the [Linux FHS](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), `/opt` is the best place for these files. This approach is also often used on other UNIX like systems.

The recommended layout of an installation in `/opt` is described [here](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html). I avoided going with `/opt/<package>` (`/opt/certbot`) to avoid conflicts with sysadmins who may have reasonably installed Certbot there.

Instead, I went with `/opt/<provider>`. While we don't have a LANANA provider name, [this page](http://www.lanana.org/lsbreg/providers/index.html) says we can use our fully qualified domain.

Finally, the layout within `/opt/eff.org` was chosen to allow EFF to package other software here in the future and `certbot/venv` was created per Erik's thought that we may want to put other things in `/opt/eff.org/certbot` in the future. The layout of an `/opt/<provider>` directory is not specified by FHS.

#### FHS deviations

The only one that I'm aware of is by installing ourselves in `/opt` we are supposed to use `/var/opt`, `/etc/opt`, etc., however, this is a major breaking change and I don't think we should do this (now).

#### Use of `sudo -E`

sudo usually preserves $HOME.

We allow people to place Certbot configuration files in their home directory (see https://certbot.eff.org/docs/using.html#configuration-file).

Certbot configuration files can contain hooks.

∴ certbot-auto already allows arbitrary code execution as root if an attacker can modify the files of an unprivileged user who runs certbot-auto using sudo so an attacker gains no new privileges by us preserving the environment.

I plan to document this in our user guide, however, since this behavior is not new, I don't think it should go in this large PR. I will document us preserving environment variables with `SUDO` (which `su` already does) in the same location.

#### Symlink from old venv path to new

I didn't do this if the old venv doesn't exist. This could cause problems for people for automated integrations installing `certbot-auto` and reaching into the venv directly. I'm not sure these people exist and I think this approach is cleaner, however, I don't have a strong preference and am happy to modify this PR so we always create the symlink.

#### certbot-auto test changes

A number of changes were necessary.

1. The tests used `venv_path` for both the path to the `letsencrypt-auto` script and the path to the virtual environment. Despite `letsencrypt-auto` deleting its venv, we were able to get away with this because we didn't actually set `VENV_PATH` but instead used `XDG_DATA_HOME`. Now that this variable isn't used in the same way, I split this to have separate paths to the venv and le-auto.
2. Because `certbot-auto` now does more as `root`, we sometimes can't delete directories used in testing when running as an unprivileged user. I fixed this by setting `ignore_errors=True` in the call to `rmtree` which shouldn't matter because we run these tests in Docker and they already do much more aggressive things like invoking your package manager.
3. In a similar vein, I had to split up one of the tests as the unprivileged tests don't have the permission to modify the venv owned as root.

#### Additional testing

I've done a lot of manual and hackily automated testing. In particular, I've tested the `su`, `sudo`, and `root` cases, both with and without existing Certbot installations on Debian, Red Hat, and FreeBSD systems. I've also successfully tested a modified `test_leauto_upgrades.sh` script with the following code:
```
if [ "$(tools/readlink.py ~/.local/share/letsencrypt)" != "/opt/eff.org/certbot/venv" ]; then
    echo symlink from old venv path not properly created!
    exit 1
fi
```
We should make this change after/during the release.

#### Warnings from pip

This should be fixed with erikrose/pipstrap#11. We already disable the cache when we invoke `pip` in certbot-auto.


letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -147,35 +152,38 @@ su_sudo() {
...
-else
-  if test "`id -u`" -ne "0" ; then
-    if $EXISTS sudo 1>/dev/null 2>&1; then
-      SUDO=sudo
-      SUDO_ENV="CERTBOT_AUTO=$0"
+if [ "$1" = "--cb-auto-has-root" ]; then
+  shift 1
+elif [ "$1" != "--le-auto-phase2" ]; then
+  # if $1 is --le-auto-phase2, we've executed this branch before
+  if [ -n "${LE_AUTO_SUDO+x}" ]; then
```
pde: Nit: put either this `SUDO` setting logic, and/or maybe all of this code block, in a named function for readability purposes?

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -147,35 +152,38 @@ su_sudo() {
...
+      '') ;; # Nothing to do for plain root method.
+      *)
+        error "Error: unknown root authorization mechanism '$LE_AUTO_SUDO'."
+        exit 1
+    esac
+    say "Using preset root authorization mechanism '$LE_AUTO_SUDO'."
+  else
+    if test "`id -u`" -ne "0" ; then
+      if $EXISTS sudo 1>/dev/null 2>&1; then
+        SUDO="sudo -E"
```
pde: Shouldn't we have `say "Requesting root privileges to run certbot..."` here, so that people aren't surprised by the prompt?

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -1124,12 +1127,14 @@ else
   # package). Phase 2 checks the version of the locally installed certbot.
 
   if [ ! -f "$VENV_BIN/letsencrypt" ]; then
-    if [ "$HELP" = 1 ]; then
-      echo "$USAGE"
-      exit 0
+    if [ -z "$OLD_VENV_PATH" -o ! -f "$OLD_VENV_PATH/bin/letsencrypt" ]; then
```
pde: Under what circumstances would `[ -z "$OLD_VENV_PATH" ]` be meaningful? Did you mean something like `! -d`?

letsencrypt-auto-source/letsencrypt-auto
bmw: Thanks for catching that. People have gotten understandably frustrated in the past when we ask for root without telling them why.

letsencrypt-auto-source/letsencrypt-auto
bmw: If the user sets `VENV_PATH`, `OLD_VENV_PATH` isn't defined. See [this](https://github.com/certbot/certbot/pull/4970/files#diff-a4881cc6ba1c9dcaeb3962d88a0c369dR26).

bmw:
Travis tests are expected to fail until #5004 is merged.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -152,10 +152,11 @@ su_sudo() {
...
 
-if [ "$1" = "--cb-auto-has-root" ]; then
-  shift 1
-elif [ "$1" != "--le-auto-phase2" ]; then
-  # if $1 is --le-auto-phase2, we've executed this branch before
+# Sets the environment variable SUDO to be the name of the program or function
+# to call to get root access. If this script already has root privleges, SUDO
+# is set to an empty string. The value in SUDO should be run with the command
+# to called with root privileges as arguments.
+set_root_auth_mechanism() {
```
erikrose: Maybe we could standardize the subroutine name casing conventions while we're in here. This one arguably introduces a new convention. Most seem to be UpperCamelCase, with su_sudo as an outlier and some single-word lowercase ones thrown in for good measure.

bmw: I didn't change `say` or `error` as they're used all over the place and otherwise are unaffected by this PR, but I believe all other names now use the same convention and I created #5012 to track finishing this up.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -152,10 +152,11 @@ su_sudo() {
...
 
-if [ "$1" = "--cb-auto-has-root" ]; then
-  shift 1
-elif [ "$1" != "--le-auto-phase2" ]; then
-  # if $1 is --le-auto-phase2, we've executed this branch before
+# Sets the environment variable SUDO to be the name of the program or function
+# to call to get root access. If this script already has root privleges, SUDO
+# is set to an empty string. The value in SUDO should be run with the command
+# to called with root privileges as arguments.
+set_root_auth_mechanism() {
```
pde: Suggestion: personally I like to CapitaliseBashFunctions so it's easier to tell when they're being called. But we've already acquired both idioms...

bmw:
Unless @erikrose wanted to take another look, I'm going to merge this PR.

erikrose:
Lgtm. Merge away!

PR #4971, opened 2017-07-27
certbot-auto OS dependency update system

### LINKED ISSUE #2208: Install Python3 via EPEL on CentOS6 with letsencrypt-auto?

bmw:
Start of the work for #2208. This allows us to add new OS dependencies to `certbot-auto`. Once this general mechanism has landed, we can use it resolve #2208.

One notable thing about this PR is we now differentiate between `-n` and `renew`. Since we tell people to run `certbot-auto renew` from `cron`, if we allowed this subcommand to force us to rebootstrap, we'd automatically install packages on people's system who followed our instructions which caused a lot of complaints before and largely makes this entire mechanism unnecessary.


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -358,7 +358,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
   # If the venv doesn't exist, we just bootstrapped
   if [ -d "$VENV_PATH" -a "$BOOTSTRAP_VERSION" != "$(PrevBootstrapVersion)" ]; then
     # if non-interactive mode or stdin and stdout are connected to a terminal
-    if [ "$NONINTERACTIVE" = 1 -o -t 0 -a -t 1]; then
+    if [ \( "$NONINTERACTIVE" = 1 \) -o \( \( -t 0 \) -a \( -t 1 \) \) ]; then
```
erikrose: Interesting. Are the parens necessary? I was going to ask you to put them in to help poor people like me who don't know the relative precedence of -a and -o by heart, but then I couldn't find any docs on parens, and "seems to work" is a lousy guide for portable software. :-)

bmw: The parentheses are necessary. Without them, I get a syntax error when using `dash`.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -319,6 +319,32 @@ else
   BOOTSTRAP_VERSION="Unknown 1"
 fi
 
+PrevBootstrapVersion() {
+  if [ -f $VENV_BOOTSTRAP_VERSION ]; then
+    cat $VENV_BOOTSTRAP_VERSION
+  elif grep -Fqx "$BOOTSTRAP_VERSION" << "UNLIKELY_EOF"
+NO_BOOTSTRAP 1
+BootstrapDebCommon 1
```
bmw: So the thought here was that we want our current bootstrap scripts to be treated specially. If there's no bootstrap version saved to disk but the bootstrap script we're considering running is from before we landed this PR, we probably don't need to rebootstrap with the thinking that the venv was from before we made this change. New bootstrap scripts we make in the future though should always have their version saved.

I very indirectly touched on this [here](https://github.com/certbot/certbot/pull/4971/files#diff-a4881cc6ba1c9dcaeb3962d88a0c369dR763), but I'm happy to expand that comment and/or move it to before this list.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -229,23 +229,40 @@ DeterminePythonVersion() {
...
-    # Mageia has both /etc/mageia-release and /etc/redhat-release
+  }
+  BOOTSTRAP_VERSION="BootstrapDebCommon $BOOTSTRAP_DEB_COMMON_VERSION"
+elif [ -f /etc/mageia-release ]; then
+  # Mageia has both /etc/mageia-release and /etc/redhat-release
+  Bootstrap() {
     ExperimentalBootstrap "Mageia" BootstrapMageiaCommon
-  elif [ -f /etc/redhat-release ]; then
+  }
+  BOOTSTRAP_VERSION="BootstrapMageiaCommon $BOOTSTRAP_MAGEIA_COMMON_VERSION"
```
bmw: The more I work on `certbot-auto`, the more I grow to dislike writing POSIX shell scripts with any complexity (especially those that have to use a subset of the POSIX shell to preserve compatibility with non-POSIX shells like busybox).

erikrose: Yes, I tend to branch off into Python at my earliest opportunity (as you can see here) to get some decent flow control. One crazy idea I considered for this situation was to store a dispatch table as text and use grep to pretend it's some kind of associative array:

```
operating_system_id    version    function_name
2                      1          bootstrapSuchAndSuch
5                      2          bootstrapAnotherThing
...
```

But I couldn't quite make it apply here.

letsencrypt-auto-source/letsencrypt-auto.template
erikrose: Yes, I grok that and agree. But I still feel like there could be a simpler way.

One might be to say roughly `if [ ! -f $VENV_BOOTSTRAP_VERSION ]; then echo "$BOOTSTRAP_VERSION" | awk '{print $1, 1}' > "$VENV_BOOTSTRAP_VERSION"; fi;`. Then PrevBootstrapVersion could just unconditionally check the file.

There might need to be changes to how NO_BOOTSTRAP is handled; you wouldn't want that getting written to a file. And you wouldn't want it to re-bootstrap unnecessarily if BOOTSTRAP_VERSION was "NO_BOOTSTRAP 1" (because --no-bootstrap was passed) and then control hit `if [ -d "$VENV_PATH" -a "$BOOTSTRAP_VERSION" != "$(PrevBootstrapVersion)" ]; then`.

letsencrypt-auto-source/letsencrypt-auto.template
bmw: > Yes, I tend to branch off into Python at my earliest opportunity (as you can see here) to get some decent flow control.

Yeah. Just to state this explicitly, we can't do that here because this code runs before bootstrapping meaning we can't rely on Python being installed.

The only approaches I have really just push things around and result in more code overall. We could do something like set `BOOTSTRAP_FUNC="BootstrapSuchAndSuch"` and then later for every OS run `"Get"$BootstrapSuchAndSuch"Version"` which could be a function that returns/sets the bootstrap version for that system but I'm personally not convinced that's any cleaner.

letsencrypt-auto-source/letsencrypt-auto.template
bmw: > One might be to say roughly if [ ! -f $VENV_BOOTSTRAP_VERSION ]; then echo "$BOOTSTRAP_VERSION" | awk '{print $1, 1}' > "$VENV_BOOTSTRAP_VERSION"; fi;. Then PrevBootstrapVersion could just unconditionally check the file.

I don't think this has the behavior I described above (unless we make a rule that all new bootstrap scripts start at version 2).

For example, my plan was to split Red Hat based systems where we're installing a newer Python from EPEL into a new bootstrap script rather than adding a conditional to the existing RH script. The reason for doing this is it minimizes the number of people who need to rebootstrap.

If we always write "BootstrapFunc 1" if no file exists, some Red Hat users that need to rebootstrap won't do so. `certbot-auto` won't have created the file, we'll see it's the first version of the script (despite it now being a different script) and assume there's nothing to do.

My plan was to never modify this list. The existing bootstrap scripts are special because they existed before we saved the bootstrap version.

> There might need to be changes to how NO_BOOTSTRAP is handled; you wouldn't want that getting written to a file. And you wouldn't want it to re-bootstrap unnecessarily if BOOTSTRAP_VERSION was "NO_BOOTSTRAP 1" (because --no-bootstrap was passed) and then control hit if [ -d "$VENV_PATH" -a "$BOOTSTRAP_VERSION" != "$(PrevBootstrapVersion)" ]; then.

That's a good point. I think `--no-bootstrap` should just sidestep this logic entirely. In fact, it might be a good idea to copy over the version from the last bootstrap script we did run for when the user (possibly accidentally) runs without `--no-bootstrap` in the future.

bmw: Let me know what you think of this idea so I can modify this PR accordingly.

erikrose: I was actually going to suggest giving all the existing bootstrap scripts version 0, since it's more "special" to human eyes and people will probably give version 1 to new scripts no matter what our documentation says. The fact that we arrived at the same solution makes me lean further in that direction. That, combined with --no-bootstrap genuinely sidestepping the bootstrapping logic, is my favored approach.

bmw: > I was actually going to suggest giving all the existing bootstrap scripts version 0, since it's more "special" to human eyes and people will probably give version 1 to new scripts no matter what our documentation says.

I personally have a preference of hard coding a list like this. Up until now, we've had no reason to change our bootstrap scripts and I'm only planning on adding a new one. With this in mind, most will continue to sit at whatever version we give them now for the foreseeable future. Rather than hoping/documenting that future devs should do versioning a specific way, I'd rather put this list here so bootstrap script authors don't need to know anything about what version strings are treated specially. Regardless of what they put, it should just work.

> That, combined with --no-bootstrap genuinely sidestepping the bootstrapping logic, is my favored approach.

Great I'll do this along with preserving the bootstrap version between venvs if `--no-bootstrap` is provided.

erikrose: In that case, be sure to leave a big, clear comment above your hard-coded list explaining what it's for so adders of new bootstrap scripts won't be tempted to add their script to the list just because the rest of them seemed to be there. :-)

bmw: OK I made this change. I treated the case where we don't know how to bootstrap on your system the same was as when we include `--no-bootstrap`. While this is unlikely to matter much, it stops us from writing "Unknown..." to disk and prevents us the following:

1. A user on an unsupported system runs Certbot with `--no-bootstrap`.
2. They later forget to include `--no-bootstrap`, we tell them they need to rebootstrap, and we exit with a warning saying we don't know how to bootstrap on your system.

We now properly preserve the bootstrap version when recreating the venv when `--no-bootstrap` is included on the command line.

bmw:
This is ready for another review.

Issue #4434, opened 2017-03-29
Make a test_util function for generating a config


bmw:
All over the place in our tests we create a config object to pass to the code we're testing. We should make a common function/method/class in `certbot.tests.util` that creates a config object with the default values so it can be used throughout our tests.

Related to #4078.

nickthetait:
I'm working on this one

nickthetait:
What sort of default values are you thinking?

SwartzCr:
@bmw that might be a cue to better document this issue

bmw:
I've updated my original post with more information.

bmw:
A version of this defining common values was added in #4974.

SwartzCr:
@bmw should this be closed now that #4974 is merged?

bmw:
I don't think so. That only defined 4 values rather than default values for the entire config. There's still work being done towards this though (see #5037).

Issue #4435, opened 2017-03-29
Set all CLI defaults in certbot/constants.py


bmw:
This issue is mainly motivated to make #4434 easier, but it's nice to have a common place for setting these values rather than setting some in `cli.py` and some in `constants.py`. `HelpfulArgumentParser` in `cli.py` could maybe even be modified to no longer take a default value when an argument is added and instead take it from the config.

bmw:
To build on this, notice how [some command line arguments](https://github.com/certbot/certbot/blob/a7a8e060e37fad21a9702ac0f143fd39609c8c2d/certbot/cli.py#L846) use [flag_default()](https://github.com/certbot/certbot/blob/a7a8e060e37fad21a9702ac0f143fd39609c8c2d/certbot/cli.py#L270) while [others do not](https://github.com/certbot/certbot/blob/a7a8e060e37fad21a9702ac0f143fd39609c8c2d/certbot/cli.py#L850). This is inconsistent and limits how useful `certbot.constants.CLI_DEFAULTS` can be in testing. If all defaults were defined in this dictionary, solving #4434 could be as simple as making a class with this in `setUp()`:
```
self.config = mock.MagicMock(**constants.CLI_DEFAULTS)
```
and changing a couple values such as file paths to a temporary directory.

bmw:
This is done.

PR #4974, opened 2017-07-28
#4434 Test Config Base Class

cj-dev:
Opening for review. If I can get confirmation this approach meets expectations outlined in the issue, I'll proceed with refactoring other tests that can take advantage of the `ConfigTestCase`.

### DIFF
--- a/certbot/tests/account_test.py
+++ b/certbot/tests/account_test.py
@@ -61,9 +61,6 @@ class AccountTest(unittest.TestCase):
 class ReportNewAccountTest(ConfigTestCase, unittest.TestCase):
     """Tests for certbot.account.report_new_account."""
 
-    def setUp(self):
-        super(ReportNewAccountTest, self).setUp()
-
     def _call(self):
         from certbot.account import report_new_account
         report_new_account(self.config)
--- a/certbot/tests/account_test.py
+++ b/certbot/tests/account_test.py
@@ -58,7 +58,7 @@ class AccountTest(unittest.TestCase):
         self.assertTrue(repr(self.acc).startswith(
           "<Account(i_am_a_regr, bca5889f66457d5b62fbba7b25f9ab6f, Meta("))
 
-class ReportNewAccountTest(ConfigTestCase, unittest.TestCase):
+class ReportNewAccountTest(ConfigTestCase):
     """Tests for certbot.account.report_new_account."""
 
     def _call(self):
@@ -95,7 +95,7 @@ class AccountMemoryStorageTest(unittest.TestCase):
         self.assertEqual([account], self.storage.find_all())
 
 
-class AccountFileStorageTest(ConfigTestCase, TempDirTestCase):
+class AccountFileStorageTest(ConfigTestCase):
     """Tests for certbot.account.AccountFileStorage."""
 
     def setUp(self):
--- a/certbot/tests/account_test.py
+++ b/certbot/tests/account_test.py
@@ -16,7 +16,7 @@ from certbot import errors
 
 from certbot.tests import util
 
-from certbot.tests.util import TempDirTestCase
+from certbot.tests.util import TempDirTestCase, ConfigTestCase
 
 
 KEY = jose.JWKRSA.load(util.load_vector("rsa512_key_2.pem"))
@@ -58,11 +58,11 @@ class AccountTest(unittest.TestCase):
         self.assertTrue(repr(self.acc).startswith(
           "<Account(i_am_a_regr, bca5889f66457d5b62fbba7b25f9ab6f, Meta("))
 
-class ReportNewAccountTest(unittest.TestCase):
+class ReportNewAccountTest(ConfigTestCase, unittest.TestCase):
     """Tests for certbot.account.report_new_account."""
 
     def setUp(self):
-        self.config = mock.MagicMock(config_dir="/etc/letsencrypt")
+        super(ReportNewAccountTest, self).setUp()
 
     def _call(self):
         from certbot.account import report_new_account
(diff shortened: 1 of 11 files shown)


bmw:
This is largely on the right track with a couple of caveats:

1. Rather than using multiple inheritance in the test cases, you should just have the tests inherit from `ConfigTestCase` unless absolutely necessary. 
2. You probably want to use subdirectories of the temporary directory for the config, log, and work directories. I can see the directories interacting in ugly ways otherwise.
3. You may want to consider solving #4435 first. To build on my description in #4434, the config object is [built at runtime](https://github.com/certbot/certbot/blob/7d0e4d7bad7f602c63af01526d8b4d316e05e32e/certbot/main.py#L741) largely by just parsing the command line arguments. For testing, it'd be nice to have a config with all the default values (except a couple which may be modified to point to things like temporary directories). The reason this is nice is many tests duplicate the work of setting up these values so the tests run properly. If they don't, code may fail due to an attribute error or being unexpectedly passed a `mock` object somewhere down the line. Some tests may changes a value or two to not be the default, but the test changing it's copy of the config shouldn't be a problem and allowing the test to focus on what it's doing differently rather than defining boilerplate makes both reading and writing tests easier. #4435 which I just added more information to describes a way to create this default config. While I'd take a PR that duplicates the keys and values in the config here, it'd be cleaner and more maintainable in the long run to use the approach specified in #4435 to help solve this PR.

Happy to answer any questions if you have them.

cj-dev:
1. Oversight on my part. Will do.
2. Absolutely, makes sense.
3. I'll take a jab at that too and keep you posted on progress.

certbot/tests/util.py
```diff
@@ -24,6 +24,7 @@
 from certbot import util
 
 from certbot.display import util as display_util
+from certbot.configuration import NamespaceConfig
```
bmw: nit: Our [style guide](https://certbot.eff.org/docs/contributing.html#coding-style) recommends [only importing packages/modules](https://google.github.io/styleguide/pyguide.html#Imports) so it'd be better to just import `configuration` and use ` configuration.NamespaceConfig`.

certbot/tests/account_test.py
```diff
@@ -98,14 +93,13 @@ def test_it(self):
...
-class AccountFileStorageTest(TempDirTestCase):
+class AccountFileStorageTest(test_util.ConfigTestCase):
     """Tests for certbot.account.AccountFileStorage."""
 
     def setUp(self):
         super(AccountFileStorageTest, self).setUp()
 
-        self.config = mock.MagicMock(
-            accounts_dir=os.path.join(self.tempdir, "accounts"))
+        self.config.accounts_dir = os.path.join(self.tempdir, "accounts")
```
bmw: We actually don't need this line as `NamespaceConfig` defines `accounts_dir` relative to `config_dir` ([source](https://github.com/certbot/certbot/blob/edcfc49303d4ddd2e69ba6ac6933af01d00e332c/certbot/configuration.py#L67)).

certbot/tests/cert_manager_test.py
```diff
@@ -18,58 +18,52 @@
...
-        os.makedirs(os.path.join(self.tempdir, "renewal"))
-
-        self.cli_config = configuration.NamespaceConfig(mock.MagicMock(
-            config_dir=self.tempdir,
-            work_dir=self.tempdir,
-            logs_dir=self.tempdir,
-            quiet=False,
-        ))
+        self.config.quiet = False
+        # TODO: This and others should use the NamespaceConfig properties, no?
```
bmw: Yes that's a good idea. For example, it'd be `self.config.renewal_configs_dir` below.

certbot/tests/cert_manager_test.py
```diff
@@ -354,89 +340,92 @@ def _call(self, *args, **kwargs):
...
-            self._call, mock_config)
+            self._call, self.config)
 
     @test_util.patch_get_utility()
     @mock.patch("certbot.storage.RenewableCert._check_symlinks")
     def test_rename_cert(self, mock_check, unused_get_utility):
         mock_check.return_value = True
-        mock_config = self.mock_config
-        self._call(mock_config)
+        # mock_config = self.mock_config
```
bmw: Can we delete this line?

certbot/tests/cert_manager_test.py
```diff
@@ -354,89 +340,92 @@ def _call(self, *args, **kwargs):
...
-        self.assertEqual(updated_lineage.lineagename, mock_config.new_certname)
+        self.assertEqual(updated_lineage.lineagename, self.config.new_certname)
 
     @test_util.patch_get_utility()
     @mock.patch("certbot.storage.RenewableCert._check_symlinks")
     def test_rename_cert_interactive_certname(self, mock_check, mock_get_utility):
         mock_check.return_value = True
-        mock_config = self.mock_config
-        mock_config.certname = None
+        # mock_config = self.mock_config
```
bmw: Can we delete this line?

certbot/tests/cert_manager_test.py
```diff
@@ -354,89 +340,92 @@ def _call(self, *args, **kwargs):
...
         self.assertTrue(updated_lineage is not None)
-        self.assertEqual(updated_lineage.lineagename, mock_config.new_certname)
+        self.assertEqual(updated_lineage.lineagename, self.config.new_certname)
 
     @test_util.patch_get_utility()
     @mock.patch("certbot.storage.RenewableCert._check_symlinks")
     def test_rename_cert_bad_new_certname(self, mock_check, unused_get_utility):
         mock_check.return_value = True
-        mock_config = self.mock_config
+        # mock_config = self.mock_config
```
bmw: Can we delete this line?

certbot/tests/main_test.py
```diff
@@ -811,36 +804,37 @@ def test_quiet_renew(self):
...
-        args = mock.MagicMock(account=None, config_dir=self.config_dir,
-                              logs_dir=self.logs_dir, work_dir=self.work_dir,
-                              email=None, webroot_path=None)
-        config = configuration.NamespaceConfig(args)
-        lineage = storage.RenewableCert(rc_path, config)
-        renewalparams = lineage.configuration["renewalparams"]
-        # pylint: disable=protected-access
-        renewal._restore_webroot_config(config, renewalparams)
-        self.assertEqual(config.webroot_path, ["/var/www/"])
+    # Duplicated in renewal_test.py
```
bmw: Nice catch! Instead of commenting these lines out, can we delete them?

cj-dev:
Thanks for the feedback! I still do intend to come back around to 4435, but I need to put a ribbon on that other issue we've been working on.

Issue #2857, opened 2016-04-17
Command "/root/.local/share/letsencrypt/bin/python2.7 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-wKPWZS/cffi/setup.py';exec(compile(getattr(tokenize, 'open', ope


strugglingyouth:
![qq 20160417102755](https://cloud.githubusercontent.com/assets/10249666/14584738/35de5de8-0487-11e6-8bdf-b0a46da74bcb.png)

strugglingyouth:
./letsencrypt-auto certonly --standalone --email 630441839@qq.com -d 17linux8.com


run this command then  。。。

robertsky:
same issue? #2804

sorcer1122:
I am having exactly the same issue and nothing helps. Is there any fix?

robertsky:
i managed to get it running... by adding a swapfile to make the memory allocation up to at least 2GB. not sure if it is a fix that works for sure, but it does.

sorcer1122:
Unfortunately, I am not able to add any swaps on my VPS. Is there another solution?

bmw:
Your version of `pip` is unlikely to be the source of your problems. For security/stability reasons, `letsencrypt-auto`, installs specific versions of Python packages that we have tested. With that said, if you want to upgrade your `pip` version, run:

```
${XDG_DATA_HOME:-~/.local/share}/letsencrypt/bin/pip install -U pip
```

SwartzCr:
Could the people who commented upthread let us know if they're still having issues? @sorcer1122 @robertsky @strugglingyouth

bmw:
This should be fixed in our next release with #4978.

Issue #2804, opened 2016-04-11
cannot run letsencrypt in Ubuntu 14.04.4 x64 


lawrencehui:
Previously I just run letsencrypt-auto and will be able to create the cert effortlessly. However today I have encounter the follow issue:

Command "/root/.local/share/letsencrypt/bin/python2.7 -u -c "import setuptools, tokenize;**file**='/tmp/pip-build-mTEeBV/cryptography/setup.py';exec(compile(getattr(tokenize, 'open', open)(**file**).read().replace('
', '
'), **file**, 'exec'))" install --record /tmp/pip-M1jUNg-record/install-record.txt --single-version-externally-managed --compile --install-headers /root/.local/share/letsencrypt/include/site/python2.7/cryptography" failed with error code 1 in /tmp/pip-build-mTEeBV/cryptography
/root/.local/share/letsencrypt/local/lib/python2.7/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:120: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
You are using pip version 8.0.3, however version 8.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

anyone can show me some lights on how to upgrade the pip inside the virtualenv? I'm really stuck on this at the moment...

robertsky:
try adding a swapfile file. might work.

bmw:
Your version of `pip` is unlikely to be the source of your problems. For security/stability reasons, `letsencrypt-auto`, installs specific versions of Python packages that we have tested. With that said, if you want to upgrade your `pip` version, run:

```
${XDG_DATA_HOME:-~/.local/share}/letsencrypt/bin/pip install -U pip
```

StanBright:
I just hit the same issue. Unfortunately, I can't give more details, except, it worked correctly on a fresh v14.04.04 installation and didn't on an older v14.04

Moonomo:
@bmw that doesn't work. `~/.local/share}/letsencrypt/bin/pip` are replaced by the `letsencrypt-auto` and gives similar error every time.

bmw:
@Moonomo, you're right. What I posted won't work here.

[...]

Issue #1680, opened 2015-12-03
installation error on amazon linux


barshinoff:
I ran 
./letsencrypt-auto --help --debug
to install lets encrypt.

All packages successfully installed but then the following error occurs:
Creating virtual environment...
Updating letsencrypt and virtual environment dependencies......Command "/root/.local/share/letsencrypt/bin/python2.7 -c "import setuptools, tokenize;**file**='/dev/shm/pip-build-OslHtf/cffi/setup.py';exec(compile(getattr(tokenize, 'open', open)(**file**).read().replace('\r\n', '\n'), **file**, 'exec'))" install --record /tmp/pip-lx9pFX-record/install-record.txt --single-version-externally-managed --compile --install-headers /root/.local/share/letsencrypt/include/site/python2.7/cffi" failed with error code 1 in /dev/shm/pip-build-OslHtf/cffi  

Any idea? I realize amazon linux support is experimental and I don't intend to use all the automation tools (yet) but it would be great if I could at least get certificates generated manually.

kiwimancy:
From [#1458](https://github.com/letsencrypt/letsencrypt/issues/1458),
>I was able to run the client with a modified letsencrypt-auto on Amazon Linux 2015.03 by manually installing the python27-virtualenv package first, then running the RedHat bootstrap path.

I did that and it seems to be working. You need to run sudo yum install python27-virtualenv and sudo ./bootstrap/_rpm_common.sh

levskaya:
@barshinoff I'm curious if this is at all related to #1677 (an out of memory error when compiling cffi dependencies on other redhat-family distros)  I've checked that ./letsencrypt-auto works on clean installs of amazon linux 2015.09.1 and 2015.03.1 on t2.micro instances with nothing else running.  What instance type and AMI are you trying this on?  If you run it verbosely with '-v' ./letsencrypt-auto --help --debug -v it might shed more information on what's going on.

barshinoff:
Hi, thanks for the advice.
It is not a memory issue, to make sure I made sure nothing else was running and hooked up a 40gb swap drive.
I am running 2015.09 distro with all packages updated.

Installing python27-virtualenv package first, then running the RedHat bootstrap path as suggested by @kiwimancy got me past the initial error. 
The error I now receive is:
`Updating letsencrypt and virtual environment dependencies.......
Running with virtualenv: /root/.local/share/letsencrypt/bin/letsencrypt
Traceback (most recent call last):
[...]

Issue #1458, opened 2015-11-11
AWS Linux errors


yobyot:
Attempting a stand-alone install on [AWS Linux 2015.03](https://aws.amazon.com/amazon-linux-ami/2015.03-release-notes/) generates this error:

`Command "/home/ec2-user/.local/share/letsencrypt/bin/python2.7 -c "import setuptools, tokenize;__file__='/tmp/pip-build-xMWPFC/cffi/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r \n', '\n'), __file__, 'exec'))" install --record /tmp/pip-BRUl0h-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/ec2user/.local/share/letsencrypt/include/site/python2.7/cffi" failed with error code 1 in/tmp/pip-build-xMWPFC/cffi`

I suspect its because I am missing something on this instanance, though I did make sure to 

`alternatives --set python /usr/bin/python2.7`

as the default is 2.6 and letsencrypt-auto said support for Python 2.6 is "experimental."

I also looked for logs that might tell me what was missing, but didn't find anything.

Any ideas are appreciated.

pde:
It looks like this is `pip` failing to build `python-cffi`, which is needed to build `python-cryptography`. Unfortunately we need the latest version of `python-cryptography`.

pde:
You can may be able to get a bit more information by running `letsencrypt-auto --verbose`.

pde:
I was able to run the client with a modified `letsencrypt-auto` on Amazon Linux by manualling installing the `python27-virtualenv` package first, then running the RedHat bootstrap path.

pde:
OK, I've made an amazonlinux branch of the main git repo; supporting this OS isn't going to be a high priority for us until we've closed out some other urgent tickets, but feel free to try the branch and send results to #1465.

yobyot:
Thanks for the quick response, pde. I do hope this will make it into 1.0 as I suspect many business and individuals will do what I did: throw up an AWS Linux instance in order to test the system. It's a quick and fast way to get an Apache system running; you can do it in 15 minutes or less. People might not try again if AWS Linux doesn't work on the first try. ;-) I don't think you need to do automated cert installation in AWS Linux -- but it would be great just to be able to generate the certs.

I've checked out the branch and switched to it. Unfortunately, `./bootstrap/amazon_linux.sh` is missing: 
[...]

PR #4978, opened 2017-07-31
Update pipstrap to version 1.3.0

### LINKED ISSUE #1081: Running with 192MB RAM is not enough
`python-cryptography` apparently tries to compile a pretty big C binding or something like that, and gcc proceeds to ICE due to lack of free memory.

The `letsencrypt-auto` tool should know its memory requirements and ask whether you really want to proceed if you haven’t enough.

bmw:
Fixes #1081, #1458, #1680, #2804, #2857.

This allows us to take advantage of cffi and cryptography's manylinux1 wheels to avoid compilation on most systems.


bmw:
I'll update this PR once erikrose/pipstrap#15 lands.

bmw:
Good idea. I just updated the PR.

bmw:
Awesome. Thanks for the quick reviews @erikrose!

Issue #4917, opened 2017-07-10
Organize + document `certbot/tests/testdata` directory


ynasser:
We have a horribly organized `certbot/tests/testdata/` directory, and it's not clear which of our testing certificates go with which keys, and this has caused some confusion. Ideally, this would be documented somewhere around here: https://certbot.eff.org/docs/contributing.html#testing.

Here's a small sample of what's involved in part of this (e.g. checking that a private key goes with a certificate):

```
$ openssl x509 -noout -modulus -in cert.pem | openssl md5
(stdin)= c1cc74880b0f3366db7c9fc48a359dff
$ openssl rsa -noout -modulus -in rsa256_key.pem | openssl md5
(stdin)= f7f1dda5f208c1718c3c18189114ddd6
```

Since the hashes don't match, they don't go together (but those two are used together in a lot of the unit tests).

Relevant preliminary reading on the differences between `fullchain`, `cert`, and `chain` files for whoever takes this on: https://certbot.eff.org/docs/using.html#where-are-my-certificates

Baeumla:
Since I have already worked on this for my last PR, I will take care of this issue. I will have to take a closer look into the `_test` files so I don't know when I will be able to create the PR for this. Will do it as soon as possible.

@ynasser you can assign this to me if possible.

ynasser:
Oh, wow thanks @Baeumla! I'll just assign it to myself so we know who's managing it, since GH won't let me assign people not in the organization.

Baeumla:
BTW: When I was generating the Sphinx documentation on my macOS with the `make -C docs clean html man` command, I was having import errors. I had to manually install `repoze.sphinx.autointerface` and `sphinx_rtd_theme`. I think this should be mentioned in the Developer Guide. 

To install those packages I used the pip installer.
```
pip install repoze.sphinx.autointerface
pip install sphinx_rtd_theme 
```

ynasser:
Thanks for pointing that out @Baeumla. I've made the following issue to address this: #4946.

Baeumla:
Okay. @ynasser do you actually know why there is a second testdata directory `certbot/acme/acme/testdata`? This one has a README on how the certificates and keys got created. Wouldn't it be better if there would be only one testdata-directory, or is it better to separate the acme from the certbot testdata?

[...]

PR #4983, opened 2017-08-02
Organize + document `certbot/tests/testdata` directory

Baeumla:
@ynasser PR for #4917

I compared all keys, certs and csr's, deleted the ones which didnt get used or which were redundant. 

I thought that the README in the folder is enough and that it doesn't need to be documented in https://certbot.eff.org/docs/contributing.html#testing.


ynasser:
Thanks for the PR @Baeumla. I'm traveling right now so my schedule is a bit hectic, so I'll get to this in a week or so.

ynasser:
This looks good to me! Thanks for the PR @Baeumla!

bmw:
Merging with @ynasser's review. Thanks @Baeumla!

PR #4987, opened 2017-08-03
Revocation reason

SwartzCr:
addresses this community forum comment: https://community.letsencrypt.org/t/revoking-certificates-is-it-possible-useful-to-specify-the-reason-of-the-revocation/39452/4

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1156,9 +1156,10 @@ def _create_subparsers(helpful):
                 " Currently --csr only works with the 'certonly' subcommand.")
     helpful.add("revoke",
                 "--reason", dest="reason",
-                choices=CaseInsensitiveList(constants.REVOCATION_REASONS.keys()),
+                choices=CaseInsensitiveList(sorted(constants.REVOCATION_REASONS,
+                                                   key=constants.REVOCATION_REASONS.get),
                 action=_EncodeReasonAction, default=0,
-                help="Specify reason for revoking certificate.")
+                help="Specify reason for revoking certificate. (default: unspecified)")
     helpful.add("rollback",
                 "--checkpoints", type=int, metavar="N",
                 default=flag_default("rollback_checkpoints"),
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1157,7 +1157,7 @@ def _create_subparsers(helpful):
     helpful.add("revoke",
                 "--reason", dest="reason",
                 choices=CaseInsensitiveList(sorted(constants.REVOCATION_REASONS,
-                                                   key=constants.REVOCATION_REASONS.get),
+                                                   key=constants.REVOCATION_REASONS.get)),
                 action=_EncodeReasonAction, default=0,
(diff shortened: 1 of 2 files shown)


certbot/cli.py
```diff
@@ -1156,9 +1156,10 @@ def _create_subparsers(helpful):
                 " Currently --csr only works with the 'certonly' subcommand.")
     helpful.add("revoke",
                 "--reason", dest="reason",
-                choices=CaseInsensitiveList(constants.REVOCATION_REASONS.keys()),
+                choices=CaseInsensitiveList(sorted(constants.REVOCATION_REASONS,
```
bmw: Why call `sorted` here?

docs/using.rst
```diff
@@ -391,6 +391,10 @@ relevant files can be removed from the system with the ``delete`` subcommand::
 
   certbot delete --cert-name example.com
 
+You can also specify the reason for revocations using the ``reason`` flag::
+
+  certbot revoke --cert-path /etc/letsencrypt/live/CERTNAME/cert.pem --reason keycompromise
+
```
bmw: I personally think this section flows better if you introduce `--reason` right after the revoke subcommand rather than including it after we start talking about delete. Also, it might be good to list the revocation reasons the user can specify.

tdelmas:
Shouldn't that PR include changes for https://github.com/certbot/certbot/blob/master/docs/cli-help.txt#L306 ?
> (default: 0)

should probably be written as:

> (default: unspecified)

if it's the current behavior. 0 suggest the first of the list, which is "keycompromise", but the RFC says that 0 is "unspecified" :  https://tools.ietf.org/html/rfc5280#section-5.3.1

SwartzCr:
@tdelmas so the line that I added to the help output of the function does what you're asking for. If we change the actual default in argparse to the key rather than the value listed in CONSTANTS, then it won't work correctly (unfortunate I know)

tdelmas:
@SwartzCr I was thinking about something like that: https://github.com/tdelmas/certbot/commit/cf031a65b25618dd671d3453d321cbcef7c09881

certbot/cli.py
SwartzCr: because I wanted them to be in order according to their numerical value (as specified in the rfc)
It's not mega important, but I think it's nice

SwartzCr:
@tdelmas I'm fairly certain that that file gets auto build from the help output and updated at release time, so changing it in the code should be enough. @bmw right?

certbot/cli.py
bmw: Oh. I'm assuming you wanted them in this order for when they are displayed in error output. I didn't realize they were used there.

This is fine then. Just need to fix test failures which I think are due to the reason I described [here](https://github.com/certbot/certbot/pull/4987#pullrequestreview-54217053).

tdelmas:
> that that file gets auto build from the help output

Thanks @SwartzCr , I didn't know :-)

Issue #4774, opened 2017-06-03
Domain (or any other prefix of "domains") in cli.ini breaks renew


andreas-wolf:
## My operating system is (include version):
openSUSE Leap 42.2

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto

## I ran this command and it produced this output:
certbot-auto renew --dry-run

Currently, the renew verb is capable of either renewing all installed certificates that are due to be renewed or renewing a single certificate specified by its name. If you would like to renew specific certificates by their domains, use the certonly command. The renew verb may provide other options for selecting certificates to renew in the future.

## Certbot's behavior differed from what I expected because:
It should simulate to renew the certs.

## Here is a Certbot log showing the issue (if available):
2017-06-03 22:13:31,963:DEBUG:certbot.log:Exiting abnormally:
Traceback (most recent call last):
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/main.py", line 742, in main
    return config.func(config, plugins)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/main.py", line 692, in renew
    renewal.handle_renewal_request(config)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/renewal.py", line 370, in handle_renewal_request
    raise errors.Error("Currently, the renew verb is capable of either "
Error: Currently, the renew verb is capable of either renewing all installed certificates that are due to be renewed or renewing a single certificate specified by its name. If you would like to renew specific certificates by their domains, use the certonly command. The renew verb may provide other options for selecting certificates to renew in the future.

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:
N/A

Further information
The reason for this error is a line in cli.ini:
domain = d1.example.com,d2.example.com

I added this line because of this "workaround" https://community.letsencrypt.org/t/set-default-cn-name-for-multiple-certificate/21625/4 mentioned in this issue: https://github.com/certbot/certbot/issues/3698
I also found this issue about domains in cli.ini https://github.com/certbot/certbot/issues/4096

From https://github.com/certbot/certbot/issues/4096 it seems that the proper key in cli.ini is "domains" instead of "domain". The odd thing is, that I could get a cert with all my domains using "domain".
[...]

PR #4991, opened 2017-08-03
add warnings and clarity to config documentation

SwartzCr:
In progress PR for #4152

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -538,14 +538,15 @@ commands into your individual environment.
 Modifying the Renewal Configuration File
 ----------------------------------------
 
-When a certificate is issued Certbot creates a renewal configuration file that
+When a certificate is issued, by default Certbot creates a renewal configuration file that
 tracks the options that were selected when Certbot was run. This allows Certbot
 to use those same options again when it comes time for renewal. These renewal
 configuration files are located at ``/etc/letsencrypt/renewal/CERTNAME``.
 
 For advanced certificate management tasks, it is possible to manually modify the certificate's
 renewal configuration file, but this is discouraged since it can easily break Certbot's
-ability to renew your certificates.
+ability to renew your certificates. If you choose to modify the renewal configuration file
+we advise you to test it's validity with the ``certbot renew --dry-run`` command.
 
 .. warning:: Modifying any files in ``/etc/letsencrypt`` can damage them so Certbot can no longer properly manage its certificates, and we do not recommend doing so.
 
@@ -816,10 +817,11 @@ By default, the following locations are searched:
   set).
 
 Since this configuration file applies to all invocations of certbot it is incorrect
-to list domains in it. Listing domains in cli.ini will prevent renewal from working.
+to list domains in it. Listing domains in cli.ini may prevent renewal from working.
 Additionally due to how arguments in cli.ini are parsed, options which wish to
-not be set should not be listed. Options set to ``= False`` will instead be read
-as being set to true, since they have been listed in the config file.
+not be set should not be listed. Options set to false will instead be read
+as being set to true by older versions of Certbot, since they have been listed
+in the config file.
 
 .. keep it up to date with constants.py
 
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -546,7 +546,7 @@ configuration files are located at ``/etc/letsencrypt/renewal/CERTNAME``.
 For advanced certificate management tasks, it is possible to manually modify the certificate's
 renewal configuration file, but this is discouraged since it can easily break Certbot's
 ability to renew your certificates. If you choose to modify the renewal configuration file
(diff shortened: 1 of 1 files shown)


SwartzCr:
I would like to add a few things to this PR:
- an explanation on how renew-hooks are handled between config files and the command line
- a list of all things that renewal confs can manage
- comments to the renewal .conf files giving example flags that can be added

It might also be good to explicitly mention preferred-challenges somewhere since there was a bug that stripped that from config files, and people may need to add it back in on their own, but I would like some feedback from @schoen or @bmw on how they feel about these changes

docs/using.rst
```diff
@@ -538,8 +538,14 @@ commands into your individual environment.
 Modifying the Renewal Configuration File
 ----------------------------------------
 
+When a certificate is issued Certbot creates a renewal configuration file that
```
bmw: nit: This isn't always true. If you create a certificate with a custom CSR using `--csr`, a renewal conf file isn't created. I have a slight preference towards mentioning this somewhere here for correctness and to avoid confusion for the few `--csr` users out there, but I understand if you want to leave this out to simplify our documentation. I wanted to make sure you were aware of this caveat though.

docs/using.rst
```diff
@@ -804,6 +815,12 @@ By default, the following locations are searched:
   ``~/.config/letsencrypt/cli.ini`` if ``$XDG_CONFIG_HOME`` is not
   set).
 
+Since this configuration file applies to all invocations of certbot it is incorrect
+to list domains in it. Listing domains in cli.ini will prevent renewal from working.
```
bmw: Another nit:

> Listing domains in cli.ini will prevent renewal from working.

This is no longer true since 0.14.0 and #4479. This is a hack though and has caused problems like #4774 but technically we "properly" handle this case now.

Up to you what you'd like to do with this information.

docs/using.rst
```diff
@@ -804,6 +815,12 @@ By default, the following locations are searched:
...
+Additionally due to how arguments in cli.ini are parsed, options which wish to
+not be set should not be listed. Options set to ``= False`` will instead be read
+as being set to true, since they have been listed in the config file.
```
bmw: This is no longer accurate with `configargparse 0.12.0`. #4995 aims to moves to pin this version in `certbot-auto`. Perhaps you should update this sentence to say:

"...read as being set to true by older versions of Certbot"

or

"...read as being set to true by older versions of Certbot's depdencies"

docs/using.rst
```diff
@@ -538,8 +538,14 @@ commands into your individual environment.
...
 
+When a certificate is issued Certbot creates a renewal configuration file that
+tracks the options that were selected when Certbot was run. This allows Certbot
+to use those same options again when it comes time for renewal. These renewal
+configuration files are located at ``/etc/letsencrypt/renewal/CERTNAME``.
+
 For advanced certificate management tasks, it is possible to manually modify the certificate's
-renewal configuration file, located at ``/etc/letsencrypt/renewal/CERTNAME``.
+renewal configuration file, but this is discouraged since it can easily break Certbot's
+ability to renew your certificates.
```
bmw: It might be good to say here (or at least somewhere) that if you do manually modify this file, you should test it using `certbot renew --dry-run` to ensure it can be parsed correctly.

docs/using.rst
```diff
@@ -804,6 +815,12 @@ By default, the following locations are searched:
   ``~/.config/letsencrypt/cli.ini`` if ``$XDG_CONFIG_HOME`` is not
   set).
 
+Since this configuration file applies to all invocations of certbot it is incorrect
+to list domains in it. Listing domains in cli.ini will prevent renewal from working.
+Additionally due to how arguments in cli.ini are parsed, options which wish to
+not be set should not be listed. Options set to ``= False`` will instead be read
```
bmw: super nit: I personally prefer simply "Options set to false..." as:

1. "set to ``= value``" feels awkward to me.
2. Any casing of the words "false" or "no" caused this problem (although "no" working at all isn't documented upstream).

You can ignore this though if you prefer your current approach.

docs/using.rst
```diff
@@ -538,14 +538,15 @@ commands into your individual environment.
...
+When a certificate is issued, by default Certbot creates a renewal configuration file that
 tracks the options that were selected when Certbot was run. This allows Certbot
 to use those same options again when it comes time for renewal. These renewal
 configuration files are located at ``/etc/letsencrypt/renewal/CERTNAME``.
 
 For advanced certificate management tasks, it is possible to manually modify the certificate's
 renewal configuration file, but this is discouraged since it can easily break Certbot's
-ability to renew your certificates.
+ability to renew your certificates. If you choose to modify the renewal configuration file
+we advise you to test it's validity with the ``certbot renew --dry-run`` command.
```
bmw: nit: its

SwartzCr:
This PR is for #4152

PR #5001, opened 2017-08-07
Avoid crashing on SIGWINCH

paultag:
When terminals are resized, the terminal will usually send a SIGWINCH -
the Window Change signal. This usually gives interactive applications a
chance to render the UI with the new width and height. However, if this
sinal isn't handled or trapped, the default behavior tends to be
aborting the process.

This change hooks SIGWINCH up to SIG_IGN, which ignores the signal. This
is safe, because we don't really care that the terminal has been
resized.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -1,6 +1,7 @@
 """Certbot main entry point."""
 from __future__ import print_function
 import logging.handlers
+import signal
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


SwartzCr:
Thanks @paultag! I'm going to tag @bmw to review this

bmw:
@paultag, were you able to reproduce the problem so you could test this?

paultag:
@bmw Yep! It's super easy - just run certonly, and resize it when it's running, and it'll crash. After ignoring the signal, I'll continue to wait for user input.

paultag:
Ah, makes sense. I don't hit this issue, @SwartzCr just mentioned it was an issue, and sent a PR. I'll let someone else pick this up. Thanks for the review, totally sensible feedback!

PR #5004, opened 2017-08-08
Pin dependencies in compatibility tests.

bmw:
Travis tests will fail until this is merged.

We now use tools/pip_install_editable.sh which installs our packages using the
pinned versions from certbot-auto.

We also use letsencrypt-auto-source/letsencrypt-auto instead of certbot-auto in
the root to:

1. Make sure OS bootstrappers are up to date with master.
2. Copy letsencrypt-auto-source into our tree so it can be used by
tools/pip_install_editable.sh later.


bobstrecansky:
I don't have merge rights, but +1 from me. 👍

Issue #4988, opened 2017-08-03
Include plugin selection at the INFO logging level


bmw:
As suggested by @schoen, it'd be useful if Certbot included the plugin it was using at the `INFO` logging level so it was printed to the terminal. This would often help with debugging problems when the user provides only Certbot's output and not it's logfile(s).

We should do this for all major subcommands including certonly, run, and renew.

bmw:
Fixed in #5010.

PR #5010, opened 2017-08-09
Include plugin selection at the info logging level

Baeumla:
Quick fix for #4988

@bmw I included the classnames of the used plugins (`.__class__`) for all the major subcommands  certonly, run, renew and install


Baeumla:
@bmw thank you for your quick feedback. I think it looks better this way now.

PR #5021, opened 2017-08-10
Switched on Gentoo from '/usr/sbin/apache2ctl' to '/usr/sbin/apache2'

### LINKED ISSUE #3411: On Gentoo with systemd, Certbot invokes the OpenRC control scripts (and thus fails)
On Gentoo with systemd, calling `./certbot-auto --debug --apache certonly` (in the checkout dir, fresh apache 2.4 installation) results in a screen asking me for my domain (so far, so god). After that certbot runs into an error when restarting apache via apache2ctl. Thats not very surprising, since apache2ctl calls openrc which is not present on systemd systems.

```
  File "/root/.local/share/letsencrypt/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/main.py", line 744, in main
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/main.py", line 555, in obtain_cert
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/main.py", line 91, in _auth_from_domains
    renewal.renew_cert(config, domains, le_client, lineage)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/renewal.py", line 237, in renew_cert
    new_certr, new_chain, new_key, _ = le_client.obtain_certificate(domains)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/client.py", line 247, in obtain_certificate
    self.config.allow_subset_of_names)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/auth_handler.py", line 70, in get_authorizations
    resp = self._solve_challenges()
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot/auth_handler.py", line 108, in _solve_challenges
    resp = self.auth.perform(self.achalls)
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot_apache/configurator.py", line 1637, in perform
    self.restart()
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot_apache/configurator.py", line 1547, in restart
    self._reload()
  File "/root/.local/share/letsencrypt/lib/python2.7/site-packages/certbot_apache/configurator.py", line 1558, in _reload
    raise errors.MisconfigurationError(str(err))
MisconfigurationError: Error while running apache2ctl graceful.

 * No permission to apply cgroup settings
 * You are attempting to run an openrc service on a
 * system which openrc did not boot.
 * You may be inside a chroot or you may have used
 * another initialization system to boot this system.
 * In this situation, you will get unpredictable results!
 * If you really want to do this, issue the following command:
 * touch /run/openrc/softlevel
```

Jarodiv:
Fixes #3411

As suggested in [my comment](https://github.com/certbot/certbot/issues/3411#issuecomment-258682792) CLI_DEFAULTS_GENTOO is modified to use `/usr/sbin/apache2` instead of `/usr/sbin/apache2ctl` since apache2ctl does not work on systems with systemd.

Although this isn't the exactly same as using apache2ctl (especially the configtest command does a little more than just apache2 -t), it should be enough to make the certbot running on OpenRC as well as systemd Gentoo installations.

I've successfully tested the modified code on my Gentoo server running with systemd and apache.


certbot-apache/certbot_apache/constants.py
```diff
@@ -62,9 +62,9 @@
     vhost_files="*.conf",
     logs_root="/var/log/apache2",
     version_cmd=['/usr/sbin/apache2', '-v'],
-    define_cmd=['apache2ctl', 'virtualhosts'],
-    restart_cmd=['apache2ctl', 'graceful'],
-    conftest_cmd=['apache2ctl', 'configtest'],
+    define_cmd=['/usr/sbin/apache2', '-S'],
```
bmw: I think this breaks the work @joohoi did on #2701. On a Gentoo system with OpenRC, the output from each command is:
```
# apache2ctl virtualhosts
 * No permission to apply cgroup settings
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
VirtualHost configuration:
*:443                  localhost (/etc/apache2/vhosts.d/00_default_ssl_vhost.conf:11)
*:80                   localhost (/etc/apache2/vhosts.d/00_default_vhost.conf:35)
ServerRoot: "/usr/lib64/apache2"
Main DocumentRoot: "/usr/htdocs"
Main ErrorLog: "/var/log/apache2/error_log"
Mutex ssl-cache: dir="/run/apache_ssl_mutex" mechanism=fcntl 
Mutex default: dir="/var/run/" mechanism=default 
Mutex mpm-accept: using_defaults
Mutex rewrite-map: using_defaults
Mutex ssl-stapling-refresh: using_defaults
Mutex ssl-stapling: using_defaults
PidFile: "/run/apache2.pid"
Define: DEFAULT_VHOST
Define: INFO
Define: SSL
Define: SSL_DEFAULT_VHOST
Define: LANGUAGE
Define: DUMP_VHOSTS
Define: DUMP_RUN_CFG
User: name="apache" id=81
Group: name="apache" id=81
```
```
# apache2 -S
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
VirtualHost configuration:
ServerRoot: "/usr/lib64/apache2"
Main DocumentRoot: "/usr/htdocs"
Main ErrorLog: "/var/log/apache2/error_log"
Mutex rewrite-map: using_defaults
Mutex default: dir="/var/run/" mechanism=default 
Mutex mpm-accept: using_defaults
PidFile: "/run/apache2.pid"
Define: DUMP_VHOSTS
Define: DUMP_RUN_CFG
User: name="apache" id=81
Group: name="apache" id=81
```
Notice the missing `Define` output.

Jarodiv: The reason for the different output is that `apache2ctl` additionally sources `/etc/conf.d/apache2` and then adds the `APACHE2_OPTS` defined there to the `apache2` call:

Example for `APACHE2_OPTS` in `/etc/conf.d/apache2`:
`APACHE2_OPTS="-D INFO -D LANGUAGE -D SSL -D PHP -D SECURITY -D PROXY"`

Function that gets executed when calling `apache2ctl virtualhosts`:
```
virtualhosts() {
        checkconfig || return 1
        ${APACHE2} ${APACHE2_OPTS} -S
}
```
The options in `APACHE2_OPTS` are used in Gentoo to define includes in vhosts, like in `/etc/apache2/vhosts.d/00_default_ssl_vhost.conf`:
```
<IfDefine SSL>
<IfDefine SSL_DEFAULT_VHOST>
<IfModule ssl_module>

# see bug #178966 why this is in here

# When we also provide SSL we have to listen to the HTTPS port
# Note: Configurations that use IPv6 but not IPv4-mapped addresses need two
# Listen directives: "Listen [::]:443" and "Listen 0.0.0.0:443"
Listen 443

<VirtualHost _default_:443>
        ...
</VirtualHost>
</IfModule>
</IfDefine>
</IfDefine>
```
For me, it worked perfectly without having the flags included but there may be indeed some side effects. But while using `apache2ctl` renders certbot completely useless for Gentoo systemd users (automatic installation AND certonly are broken), using `apache2` "only" breaks the automatic installation, leaving workarounds using certonly functional.

To fix this issue completely, certbot requires support for different commands depending on the init system. Until that is done I would say generally enabling all users to use certonly is better then breaking the support for one group (especially since systemd seems to be becoming industry standard) while enabling full support for the other.

osirisinferi: @Jarodiv OpenRC is still Gentoos default, so I would prefer the default for `certbot` also to be OpenRC compatible. Until `certbot` has a system to detect the init system used, the Gentoo maintainer might add a use flag in combination with a patch for systemd users.

Jarodiv: In contrast to other distributions Gentoo does not install any "default" by default. The user always has to choose what he uses and not if he wants to exchange a default system with something different.

As I say, it currently is completely useless for all systemd users while only a few OpenRC users make use of the auto installation feature (especially on Gentoo most will very likely manage their vhosts on their own anyway). Currently there is no way on systemd installations to use certbot, while with this change at least everyone is able to use the certonly option.

osirisinferi: @Jarodiv I'm a OpenRC Gentoo user, using `certbot` fully automatically. It's not rocket science and works flawlessly if one keeps every virtual host in a separate file (which isn't Gentoo specific).

Also, OpenRC really is the default init system ([source 1](https://wiki.gentoo.org/wiki/Comparison_of_init_systems) [source 2](https://wiki.gentoo.org/wiki/Handbook:AMD64/Installation/Base#Updating_the_.40world_set)).

Jarodiv: I know that it is labeled the default but as I said ans you should know (since you have installed Gentoo yourself), there is nothing installed by default, you chose to install ObenRC.

I also know, that it is no rocket science (I too have each vhost in a separate file) but by insisting on using the autoconfigure feature (instead of doing the configuration yourself), you deny everyone with systemd the possibility to use certbot at all.

Jarodiv: Still the current status is, that only OpenRC users can use certbot (with the apache plugin) and systemd users can't.
Whereas with this change both can use it and just the autoconfigure feature isn't available to a few of them.

I'm open for alternatives that have equal benefits but lesser disadvantages.

Jarodiv: A possibly working but extremely dirty hack may be using:
```
    version_cmd=['/bin/sh', '-c', '. /etc/conf.d/apache2; apache2 ${APACHE2_OPTS} -v'],
    define_cmd=['/bin/sh', '-c', '. /etc/conf.d/apache2; apache2 ${APACHE2_OPTS} -S'],
    restart_cmd=['/bin/sh', '-c', '. /etc/conf.d/apache2; apache2 ${APACHE2_OPTS} -k graceful'],
    conftest_cmd=['/bin/sh', '-c', '. /etc/conf.d/apache2; apache2 ${APACHE2_OPTS} -t'],
```
In short, this "config" plays inception with `subprocess.Popen` (which, in the end, executes what is defined in `CLI_DEFAULTS_GENTOO`) by making it start a shell that itself starts a shell that sources `/etc/conf.d/apache2` and then executes the apache2 command with the sourced `APACHE2_OPTS` (just like `/etc/init.d/apache2` and `/usr/lib/systemd/system/apache2.service` do).

It passed my tests in a VM but since I have no complete test setup for this project running I cannot confirm that it will work in every situation so if you (@osirisinferi) can take over the formal testing, then we may have a solution that will work for both of us AND enables everyone to use the automatic installation.

But still, it would be a dirty hack that should be removed as soon as certbot supports `CLI_DEFAULTS` depending on the used init system.

joohoi: Sorry for being late in the game. I'm starting the work on better distribution specific abstraction in the Apache plugin in September, giving you better chance to dig in. Issues like this one are more convinient to handle after the changes.

I suggest leaving the current functionality in, and I'll ping @Jarodiv and @osirisinferi when we have something more concrete at hand. Your input is invaluable to make Certbot work flawlessly on as large amount of Gentoo installations as possible.

If you are interested in submitting a PR for the new functionality, that would be more than welcome, but unfortunately we're not quite there yet.

bmw:
After @joohoi and @osirisinferi's input, I don't think we want to make this change right now.

Regardless, I appreciate you looking into the problem and writing up a PR @Jarodiv. If you'd like to continue working on this, we'd accept a PR that makes these values configurable on the command line like many of the other values defined in `constants.py`. If you have any questions about this, don't hesitate to ask!

PR #5025, opened 2017-08-12
Point simp_le link to an actively maintained fork

buchdag:
Kuba's simp_le has been unmaintained for more than a year and is starting to have some breakage.

Zenhack's fork is, as far as I know, the most actively maintained and is available through PyPI.


cpu:
For what its worth I think this is the right change. I [merged a similar redirection](https://github.com/letsencrypt/website/pull/108) on the Let's Encrypt website in Dec 2016 and haven't heard any complaints. I've been following the zenhack fork since then and it's considerably more active.

Issue #3720, opened 2016-10-31
Stop using print


bmw:
In a few places in our code, we use `print` to bypass the line wrapping that `IDisplay` does. The problem with this is it requires us to duplicate code for things like `-q/--quiet`, making maintenance harder and increasing the chance for bugs.

Instead, we should either add an optional flag to the `notification` method to disable line wrapping or add a new method to `IDisplay` that doesn't format output.

ohemorange:
This will let us do the machine-readable output for the `certificates` verb as well!

ohemorange:
Actually I'd even like to get this in for 0.10.0 to go with the human-readable version of `certificates`. What are the "we've promised an interface and now we're changing things" gotchas to be worried about?

bmw:
None here since nobody but us seem to be using the `IDisplay` interface. In general, we should make sure that changes to our interfaces don't break any of our [third party plugins](https://certbot.eff.org/docs/using.html#third-party-plugins).

ohemorange:
The first step of this is fixed by #3728, which is now merged. I'm going to have `certificates` use it, and leave removing the rest of the prints to someone else / a future release.

ohemorange:
Example:
```
disp = zope.component.getUtility(interfaces.IDisplay)
disp.notification("string to print", pause=False, wrap=False)
```

ynasser:
Note to whoever works on this: when grepping for `print` statements to replace, you can ignore ones that appear in tests. Those don't need to be replaced.

dhuang:
Working on this

ECastro10:
working on this

ECastro10:
@ohemorange I replaced the print with IDisplay in main.py in the certbot direcory. Is this the only location that you wanted the print command replaced? I am asking because I noticed that print does appear in other locations in the certbot directory but they don't have zope.component imported and they don't have the TODO comment.

dhuang:
Oops, sorry, I got stuck fixing tests and totally forgot about this. Here's the diff of my branch if it's at all helpful. Good luck!

```diff
diff --git a/certbot/cli.py b/certbot/cli.py
index 1d8952d..950d2c8 100644
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -11,6 +11,7 @@ import sys
 
 import configargparse
 import six
+import zope.component
 
 from acme import challenges
 
@@ -498,12 +499,13 @@ class HelpfulArgumentParser(object):
         else:
[...]

PR #5028, opened 2017-08-15
Stop using print in `certbot.cli` module.

r5d:
PR Notes:

   - In `certbot.main.main` function,
     `certbot.cli.prepare_and_parse_args` is called before setting the
     displayer (via `certbot.main.set_displayer`) .

       - The `certbot.cli.prepare_and_parse_args` function instantiates
         `certbot.cli.HelpfulArgumentParser` class which calls its
         "private" method `_usage_string`.

          - The `_usage_string` method has a call to
            `zope.component.getUtility(interfaces.IDisplay).notification`
            method (`self.notify`).

            Since the displayer is not set yet, call to
            `zope.component.getUtility(interfaces.IDisplay)` throws a
            `ComponentLookupError` exception.

            To handle this case, try-except block was added in
            `certbot.cli.HelpfulArgumentParser.__init__` that sets the
            displayer to `certbot.display.NoninteractiveDisplay` on
            ComponentLookupError.

    - The `certbot.tests.cli_test.ParseTest.test_help` test has two
      asserts that check the output of a command.

      The output comes from
      `certbot.cli.HelpfulArgumentParser._usage_string` through a call
      to `zope.component.getUtility(interfaces.IDisplay).notification`
      (`self.notify`).

      Since, the `zope.component.geUtility` call is patched in the
      test. There is no output; this makes the asserts in
      `certbot.tests.cli_test.ParseTest.test_help` fail.

      To handle this, new argument `stdout_notification` was added to
      `certbot.tests.util.patch_get_utility` function.

Addresses #3720.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -13,7 +13,7 @@ import configargparse
 import six
 import zope.component
 
-from zope.interface.interfaces import ComponentLookupError
+from zope.interface import interfaces as zope_interfaces
 
 from acme import challenges
 
@@ -447,7 +447,7 @@ class HelpfulArgumentParser(object):
         try:
             self.notify = zope.component.getUtility(
                 interfaces.IDisplay).notification
-        except ComponentLookupError:
+        except zope_interfaces.ComponentLookupError:
             self.notify = display_util.NoninteractiveDisplay(
                 sys.stdout).notification
 
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -11,6 +11,9 @@ import sys
 
 import configargparse
 import six
+import zope.component
+
+from zope.interface.interfaces import ComponentLookupError
 
 from acme import challenges
 
@@ -23,6 +26,7 @@ from certbot import hooks
 from certbot import interfaces
 from certbot import util
 
+from certbot.display import util as display_util
 from certbot.plugins import disco as plugins_disco
 import certbot.plugins.selection as plugin_selection
 
@@ -439,6 +443,15 @@ class HelpfulArgumentParser(object):
             "delete": main.delete,
         }
 
+        # Get notification function for printing
+        try:
+            self.notify = zope.component.getUtility(
+                interfaces.IDisplay).notification
+        except ComponentLookupError:
+            self.notify = display_util.NoninteractiveDisplay(
+                sys.stdout).notification
+
+
         # List of topics for which additional help can be provided
         HELP_TOPICS = ["all", "security", "paths", "automation", "testing"]
         HELP_TOPICS += list(self.VERBS) + self.COMMANDS_TOPICS + ["manage"]
@@ -510,10 +523,10 @@ class HelpfulArgumentParser(object):
 
         usage = SHORT_USAGE
         if help_arg == True:
-            print(usage + COMMAND_OVERVIEW % (apache_doc, nginx_doc) + HELP_USAGE)
+            self.notify(usage + COMMAND_OVERVIEW % (apache_doc, nginx_doc) + HELP_USAGE)
             sys.exit(0)
         elif help_arg in self.COMMANDS_TOPICS:
-            print(usage + self._list_subcommands())
+            self.notify(usage + self._list_subcommands())
             sys.exit(0)
         elif help_arg == "all":
             # if we're doing --help all, the OVERVIEW is part of the SHORT_USAGE at
--- a/certbot/tests/cli_test.py
+++ b/certbot/tests/cli_test.py
@@ -51,22 +51,34 @@ class ParseTest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         reload_module(cli)
 
     @staticmethod
-    def parse(*args, **kwargs):
+    def _unmocked_parse(*args, **kwargs):
         """Get result of cli.prepare_and_parse_args."""
         return cli.prepare_and_parse_args(PLUGINS, *args, **kwargs)
+
+    @staticmethod
+    def parse(*args, **kwargs):
+        """Mocks zope.component.getUtility and calls _unmocked_parse."""
+        with test_util.patch_get_utility() as mocked_get_utility: # pylint: disable=unused-variable
+            return ParseTest._unmocked_parse(*args, **kwargs)
 
     def _help_output(self, args):
         "Run a command, and return the output string for scrutiny"
 
         output = six.StringIO()
+
+        def write_msg(message, *args, **kwargs): # pylint: disable=missing-docstring,unused-argument
+            output.write(message)
+
         with mock.patch('certbot.main.sys.stdout', new=output):
-            with mock.patch('certbot.main.sys.stderr'):
-                self.assertRaises(SystemExit, self.parse, args, output)
+            with test_util.patch_get_utility() as mock_get_utility:
+                mock_get_utility().notification.side_effect = write_msg
+                with mock.patch('certbot.main.sys.stderr'):
+                    self.assertRaises(SystemExit, self._unmocked_parse, args, output)
+
         return output.getvalue()
 
     @mock.patch("certbot.cli.flag_default")
-    @test_util.patch_get_utility()
-    def test_cli_ini_domains(self, unused_mock_get_utility, mock_flag_default):
+    def test_cli_ini_domains(self, mock_flag_default):
         tmp_config = tempfile.NamedTemporaryFile()
         # use a shim to get ConfigArgParse to pick up tmp_config
         shim = lambda v: constants.CLI_DEFAULTS[v] if v != "config_files" else [tmp_config.name]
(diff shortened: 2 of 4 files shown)


certbot/tests/util.py
```diff
@@ -154,19 +155,28 @@ def make_lineage(config_dir, testfile):
     return conf_path
 
 
-def patch_get_utility(target='zope.component.getUtility'):
+def patch_get_utility(target='zope.component.getUtility',
+                        stdout_notification=False):
```
bmw: I appreciate you factoring this out allowing for potential code reuse in the future, however, I'm not sure if we should take this approach. First, this parameter only affects the `notification` method. Are we going to add similar parameters for other methods in the future as necessary? Every `IDisplay` method prints something to stdout.

Second, while we could remove this parameter and always do this for every `IDisplay` method or have this flag control the behavior of all methods, I'm not sure this is necessary. There is only one place in the entire code base where we currently use this functionality and it is seems like a lot of machinery for this one use case. It also reminds me of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

My personal preference here is to revert these changes and handle the one use case using the approach I described there.

certbot/tests/cli_test.py
```diff
@@ -63,7 +65,8 @@ def _help_output(self, args):
         return output.getvalue()
 
     @mock.patch("certbot.cli.flag_default")
-    def test_cli_ini_domains(self, mock_flag_default):
+    @test_util.patch_get_utility()
```
bmw: Rather than decorating and adding a `unused_mock_get_utility` parameter to every method, why don't we just modify the `parse` and `_help_output` methods. The way I'd do this is the following:

1. Rename the current `parse` method to `_unmocked_parse`.
2. Add a new static `parse` method that calls `_unmocked_parse` after mocking `getUtility` using `patch_get_utility`.
3. Modify `_help_output` so `notification` messages are written to the mock stdout created in `_help_output`. You could do this by modifying the `side_effect` attribute of the `notification` method of the mock object.

certbot/tests/cli_test.py
```diff
@@ -406,19 +431,22 @@ class SetByCliTest(unittest.TestCase):
     def setUp(self):
         reload_module(cli)
 
-    def test_webroot_map(self):
+    @test_util.patch_get_utility()
```
bmw: Similarly, it'd be nice if we didn't have to decorate every method and add an unused parameter. I'd probably fix this by modifying `_call_set_by_cli`.

certbot/tests/main_test.py
```diff
@@ -519,7 +519,8 @@ def test_update_symlinks(self, mock_cert_manager):
         self.assertEqual(1, mock_cert_manager.call_count)
 
     @mock.patch('certbot.cert_manager.certificates')
-    def test_certificates(self, mock_cert_manager):
+    @test_util.patch_get_utility()
```
bmw: I'm assuming you modified the 4 methods in this file because they are the ones that previously printed to `stdout`? If it's not too difficult, it'd be nice if we could modify the `_call_no_clientmock` method which should result in fewer code changes and future methods can benefit without knowing the intricacies here.

If making this change causes a few test failures, perhaps we could make a method called `_call_no_get_utility_mock` that works like `_call_no_clientmock`. If making this change causes a lot of test failures though, I think the approach you took here is perfectly fine.

certbot/tests/cli_test.py
r5d: @bmw, I'm not sure if I understand step `3.`

`_help_output` calls `parse` (where `zope.component.getUtility` gets
mocked (step `2.`)); so, if I'm understanding it correctly, the
mocked `notification` is not available in `_help_output`.

How about this:

```diff
diff --git a/certbot/tests/cli_test.py b/certbot/tests/cli_test.py
index c4f9ad07..d19c15c2 100644
--- a/certbot/tests/cli_test.py
+++ b/certbot/tests/cli_test.py
@@ -2,6 +2,7 @@
 import argparse
 import unittest
 import os
+import sys
 import tempfile
 
 import mock
@@ -51,10 +52,23 @@ class ParseTest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         reload_module(cli)
 
     @staticmethod
-    def parse(*args, **kwargs):
+    def _unmocked_parse(*args, **kwargs):
         """Get result of cli.prepare_and_parse_args."""
         return cli.prepare_and_parse_args(PLUGINS, *args, **kwargs)
 
+    @staticmethod
+    def parse(*args, **kwargs):
+        """Mocks zope.component.getUtility and calls _unmocked_parse."""
+        with test_util.patch_get_utility() as mocked_get_utility:
+            mocked_get_utility().notification.side_effect = \
+              ParseTest._stdout_notification
+            return ParseTest._unmocked_parse(*args, **kwargs)
+
+    @staticmethod
+    def _stdout_notification(*args, **kwargs):
+        """Send notification to stdout."""
+        sys.stdout.write(args[0] if args else kwargs['message'])
+
     def _help_output(self, args):
         "Run a command, and return the output string for scrutiny"
 
@@ -104,8 +118,8 @@ class ParseTest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         self.assertEqual(namespace.chain_path, os.path.abspath(chain))
         self.assertEqual(namespace.fullchain_path, os.path.abspath(fullchain))
 
-    @test_util.patch_get_utility(stdout_notification=True)
-    def test_help(self, unused_mock_get_utility):
+
+    def test_help(self):
         self._help_output(['--help'])  # assert SystemExit is raised here
         out = self._help_output(['--help', 'all'])
```
`ParseTest.test_help` passes:

```zsh
(venv)  ~/[...]/certbot/certbot/tests
↳ python cli_test.py ParseTest.test_help
.
Ran 1 test in 0.306s

OK

Comment (bmw) [inline certbot/tests/cli_test.py]: If that still causes `_stdout_notification` to write to the correct stdout (e.g. the `StringIO` object and not the real stdout), something like that is totally fine.

I accidentally left out the crucial detail of my plan earlier to have `_help_output` call `_unmocked_parse`. The thought was `_help_output` is the only one that actually has to set the `side_effect` of the `notification` method as it's specifically testing the program output. After steps 1 and 2, step 3 would look something like:
```
     def _help_output(self, args):
         "Run a command, and return the output string for scrutiny"
-
         output = six.StringIO()
+
+        def write_msg(message, *args, **kwargs):
+            output.write(message)
+
         with mock.patch('certbot.main.sys.stdout', new=output):
-            with mock.patch('certbot.main.sys.stderr'):
-                self.assertRaises(SystemExit, self.parse, args, output)
+            with test_util.patch_get_utility() as mock_get_utility:
+                mock_get_utility().notification.side_effect = write_msg
+                with mock.patch('certbot.main.sys.stderr'):
+                    self.assertRaises(SystemExit, self._unmocked_parse, args, output)
         return output.getvalue()
```
I have a slight preference for this approach, but if what you have above works, that's fine with me. Sorry for the misleading post!

File: certbot/tests/util.py (hunk 1 above)
Comment (r5d) [inline certbot/tests/util.py]: Reverted the changes.

File: certbot/tests/cli_test.py (hunk 2 above)
Comment (r5d) [inline certbot/tests/cli_test.py]: Thanks. Done.

File: certbot/tests/cli_test.py (hunk 3 above)
Comment (r5d) [inline certbot/tests/cli_test.py]: Done.

File: certbot/tests/main_test.py (hunk 4 above)
Comment (r5d) [inline certbot/tests/main_test.py]: Tests pass without mocking `getUtility`. Reverted changes.

File: certbot/cli.py [hunk 5]
```diff
@@ -11,6 +11,9 @@
 
 import configargparse
 import six
+import zope.component
+
+from zope.interface.interfaces import ComponentLookupError
```
r5d: Done.

certbot/tests/cli_test.py
```diff
@@ -49,17 +51,30 @@ def setUp(self):
...
     @staticmethod
-    def parse(*args, **kwargs):
+    def _unmocked_parse(*args, **kwargs):
         """Get result of cli.prepare_and_parse_args."""
         return cli.prepare_and_parse_args(PLUGINS, *args, **kwargs)
 
+    @staticmethod
+    def parse(*args, **kwargs):
+        """Mocks zope.component.getUtility and calls _unmocked_parse."""
+        with test_util.patch_get_utility() as mocked_get_utility: # pylint: disable=unused-variable
```
bmw: nit: You can get remove the need for the `pylint: disable` if you change this line to:
```
with test_util.patch_get_utility():
```

certbot/tests/cli_test.py
```diff
@@ -450,9 +466,10 @@ def _test_report_config_interaction_common(self):
 
 def _call_set_by_cli(var, args, verb):
     with mock.patch('certbot.cli.helpful_parser') as mock_parser:
-        mock_parser.args = args
-        mock_parser.verb = verb
-        return cli.set_by_cli(var)
+        with test_util.patch_get_utility() as mocked_get_utility: # pylint: disable=unused-variable
```
bmw: Same comment as above.

certbot/tests/cli_test.py
r5d: Done.

certbot/tests/cli_test.py
r5d: Done.

PR #5029, opened 2017-08-16
expand nginx no name error

### LINKED ISSUE #4792: Nginx should give a prompt when it can't find a server_name
When the nginx plugin can't find a server name it doesn't give you any
advice on what to do. It should tell you to add a server_name directive
to your nginx config and hint at where that might be.

SwartzCr:
fixes #4792

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -236,7 +236,10 @@ class NginxConfigurator(common.Plugin):
         if not vhost:
             # No matches. Raise a misconfiguration error.
             raise errors.MisconfigurationError(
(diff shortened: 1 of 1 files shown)


ohemorange:
As discussed out of band, let's reword to the form of "here is what we would have needed" so that they go look up information about it in a higher-bandwidth source, instead of phrasing it as "go make this particular change."

Also, it's missing a trailing period.

SwartzCr:
@ohemorange want to review again?

ohemorange:
I thought we were planning to change the language instead of pointing directly to a link that might rot?

SwartzCr:
I think it's unlikely that the official nginx docs link will rot. I think this is likely the highest bandwidth source of information for a user

SwartzCr:
@ohemorange how does that look?

PR #5037, opened 2017-08-18
#4435. CLI Argument Default Organization

cj-dev:


cj-dev:
I'm proceeding in a similar vein with the other parsing initializers defined in cli.py: `_create_subparsers`, `_paths_parser`, and `_plugins_parsing`.

cj-dev:
Some implementation notes:

`argparse`'s `add_argument` defaults to `None` when `action` is not specified and defaults to the inverse of store_true or store_false when either of those is the action, thus many defaults are initialized in `CLI_DEFAULTS` but not explicitly assigned in `cli.py`. I think it's redundant, but let me know if it'd be stylistically preferred to assign them in the `add_argument` call too.

Before I remove [WIP], I want to be sure about reliable defaults for:
- `cert_path`: If the CLI verb is "certonly", the `auth_cert_path` default is used. If the verb is "revoke", an argument is required. If it's neither, it's sometimes required, and if not, it'll default to `None`. I need to figure out which default to use for `cert_path` when not taking CLI args.
- `fullchain_path` and `chain_path`: similar duality as `cert_path` above.
- `domains`: During tests, the "domains" argument would carry arguments from previous argument parsings only if using `flag_defaults('domains')`. Several tests fail, but `ParseTest.test_parse_domains` demonstrates it best. Both example.com and trailing.period.com appear in the second parsing.

If the dream is to boil down boilerplate to a line like `self.config = mock.MagicMock(**constants.CLI_DEFAULTS)`, the above have to be addressed. I think it'd be easiest to resolve those path arguments during the `__init__` of `NamespaceConfig`, instead of trying to shoehorn them into static defaults. If anyone wants to comment on the bullet points above over the weekend, that'd be super. Regardless, I'll be back on this issue on Monday.

certbot/constants.py
```diff
@@ -18,23 +18,91 @@
...
-    http01_address="",
     tls_sni_01_port=challenges.TLSSNI01Response.PORT,
     tls_sni_01_address="",
+    http01_port=challenges.HTTP01Response.PORT,
+    http01_address="",
+    break_my_certs=False,
+    rsa_key_size=2048,
+    must_staple=False,
+    redirect=None,
+    hsts=False,
```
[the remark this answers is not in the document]
cj-dev: You're referring to `hsts`? Yeah, I thought that was a little goofy. Meant to mention it, but fixed now.

cj-dev:
> I agree with your thought. I don't have too strong of a preference here, but my thought is since these flags don't have a single default, let's not try to force one into constants.py which is technically part of production code. Instead, let's limit defining a single constant to our tests. We can either define a CLI_DEFAULTS in certbot/tests/util.py or just manually add these values in the ConfigTestCase.__init__ method as you suggested. I think the value that probably makes the most sense is ./<filename>.pem

Going with the latter strategy, I manually set the defaults for ConfigTestCase (`cert_path`, `fullchain_path`, `chain_path`) as if the verb was 'certonly', which is the case in which those two CLI_DEFAULT paths (`auth_cert_path`, `auth_chain_path`) are used. I don't think defining these constants this way is worse than letting them be magicmock'd.

cj-dev:
> I don't think defining these constants this way is worse than letting them be magicmock'd.

I say that, but in other cases, args that were MagicMock'd (therefore, True-y) may now be `None` or False-y. In the name of maintaining consistency and predictability in the future, this is good, but it's something to be aware of, given the wide scope of a change like this.

certbot/cli.py
```diff
@@ -848,29 +848,33 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+        default=flag_default("force_interactive"),
         help="Force Certbot to be interactive even if it detects it's not "
              "being run in a terminal. This flag cannot be used with the "
              "renew subcommand.")
     helpful.add(
         [None, "run", "certonly", "certificates"],
         "-d", "--domains", "--domain", dest="domains",
-        metavar="DOMAIN", action=_DomainsAction, default=[],
+        metavar="DOMAIN", action=_DomainsAction,
+        default=flag_default('domains')[:], # Copy, not reference
```
bmw: I'm slightly worried about other mutable defaults such as `pref_challs`. What if we made `flag_default` return a copy of the object using the `copy` module?

certbot/cli.py
```diff
@@ -1105,14 +1128,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         help="Command to be run in a shell once for each successfully"
         " issued certificate. For this command, the shell variable"
         " $RENEWED_LINEAGE will point to the config live subdirectory"
-        ' (for example, "/etc/letsencrypt/live/example.com") containing'
+        " (for example, \"/etc/letsencrypt/live/example.com\") containing"
         " the new certificates and keys; the shell variable"
         " $RENEWED_DOMAINS will contain a space-delimited list of"
-        ' renewed certificate domains (for example, "example.com'
-        ' www.example.com"')
+        " renewed certificate domains (for example, \"example.com"
```
bmw: nit: quotes

certbot/cli.py
```diff
@@ -1105,14 +1128,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+        " www.example.com\"")
```
bmw: nit: quotes

certbot/cli.py
```diff
@@ -1138,48 +1162,53 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+        None, "--user-agent", default=flag_default("user_agent"),
         help="Set a custom user agent string for the client. User agent strings allow "
              "the CA to collect high level statistics about success rates by OS, "
              "plugin and use case, and to know when to deprecate support for past Python "
              "versions and flags. If you wish to hide this information from the Let's "
-             'Encrypt server, set this to "". '
-             '(default: {0}). The flags encoded in the user agent are: '
-             '--duplicate, --force-renew, --allow-subset-of-names, -n, and '
-             'whether any hooks are set.'.format(sample_user_agent()))
+             "Encrypt server, set this to \"\". "
```
bmw: nit: quotes

certbot/cli.py
```diff
@@ -1245,53 +1274,68 @@ def _plugins_parsing(helpful, plugins):
...
+                "--apache", action="store_true", default=flag_default("apache"),
                 help="Obtain and install certificates using Apache")
     helpful.add(["plugins", "certonly", "run", "install", "config_changes"],
-                "--nginx", action="store_true", help="Obtain and install certificates using Nginx")
+                "--nginx", action="store_true", default=flag_default("nginx"),
+                help="Obtain and install certificates using Nginx")
     helpful.add(["plugins", "certonly"], "--standalone", action="store_true",
-                help='Obtain certificates using a "standalone" webserver.')
+                default=flag_default("standalone"),
+                help="Obtain certificates using a \"standalone\" webserver.")
```
bmw: nit: quotes

certbot/tests/client_test.py
```diff
@@ -75,6 +76,8 @@ def test_email_retry(self, _rep, mock_get_email):
     @mock.patch("certbot.account.report_new_account")
     def test_email_invalid_noninteractive(self, _rep):
         from acme import messages
+        self.config.email = mock.MagicMock()
```
bmw: Why do we set this to a mock object?

certbot/tests/util.py
```diff
@@ -254,13 +254,16 @@ class ConfigTestCase(TempDirTestCase):
...
+            mock.MagicMock(**constants.CLI_DEFAULTS)
         )
+        self.config.verb = "certonly"
+        self.config.config_dir = os.path.join(self.tempdir, 'config')
+        self.config.work_dir = os.path.join(self.tempdir, 'work')
+        self.config.logs_dir = os.path.join(self.tempdir, 'logs')
+        self.config.cert_path = constants.CLI_DEFAULTS['auth_cert_path']
+        self.config.fullchain_path = constants.CLI_DEFAULTS['auth_chain_path']
+        self.config.chain_path = constants.CLI_DEFAULTS['auth_chain_path']
+        self.config.server = "example.com"
```
bmw: This was from your other PR, but after thinking about it, why are we setting this value rather than just leaving it at the normal server default?

certbot/cli.py
cj-dev: That makes sense. As I was incorporating your feedback, I found that `ParseTest.test_cli_ini_domains` is causing the modification to CLI_DEFAULTS. It uses a mocked `flag_default`, so I'm having the mock return a copy as well.

<sleepy braindump>
In the commit you just reviewed, `test_cli_ini_domains` mocked `flag_default` in a way that didn't return a copy, and I just verified with pdb that the value in `CLI_DEFAULTS` does change after parsing domains from the temp file. When I modified `parse_default` to return a copy, given your feedback, the corrupted list was returned.

The default given to the argparser as you just reviewed was a copy of `domains`, as shown in the snippet above. Shouldn't that have likewise returned a copy of a corrupted list, containing artifacts from `test_cli_ini_domains`? Whack.
</braindump>

certbot/cli.py
```diff
@@ -1026,62 +1044,67 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
-             'listed first (Eg, "dns" or "tls-sni-01,http,dns"). '
-             'Not all plugins support all challenges. See '
-             'https://certbot.eff.org/docs/using.html#plugins for details. '
-             'ACME Challenges are versioned, but if you pick "http" rather '
-             'than "http-01", Certbot will select the latest version '
-             'automatically.')
+        action=_PrefChallAction, default=flag_default("pref_challs"),
+        help="A sorted, comma delimited list of the preferred challenge to "
+             "use during authorization with the most preferred challenge "
+             "listed first (Eg, \"dns\" or \"tls-sni-01,http,dns\"). "
```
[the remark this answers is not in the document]
cj-dev: What a blunder. I'll use single quotes around entire help blocks that need it.

certbot/tests/client_test.py
cj-dev: Good question. This test was failing during my work and this mock was my way of returning this test to its previous defaults so I could understand its nature. This mock is actually impotent for this test, so I should remove it. Thanks for catching that.

certbot/tests/util.py
cj-dev: The default points to the letsencrypt ACME server and I usually wouldn't give tests any way to accidentally throw garbage data at a production service. What are your thoughts?

cj-dev:
Marking WIP again. Take a few more minutes to understand that domains list quirk.

certbot/cli.py
bmw: Hmm. Happy to help here but I'm not sure I understand.

If I checkout the commit I just reviewed (955ebb8) and break after the calls to `self.parse` after writing the temporary file, `constants.CLI_DEFAULTS['domains']` is still an empty list.

certbot/cli.py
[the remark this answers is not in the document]
bmw: Not a blunder at all! Not many people follow PEP8 exactly and what you had is functionally equivalent. [PEP8 is just listed as one of our style guides](https://certbot.eff.org/docs/contributing.html#coding-style) and I think it looks a little cleaner and easier to read this way.

certbot/tests/util.py
bmw: My thought was that we should do a pretty good job of mocking out the calls that send data to the production server and it's nice to use the real default when possible so any tests that may rely on that value now or in the future automatically work properly, however, you have a good point. It's unlikely that the value used here will matter in a test and it's an extra safe guard against unnecessary server traffic so you can leave this as is.

Thanks for pointing that out!

certbot/cli.py
cj-dev: There are two cases I'm interested in. In both cases, `test_cli_ini_domains` does what we'd consider the "wrong" thing (not copying the retrieval). In one case, `cli.flag_defaults` returns a copy of the retrieval. In the second case, `flag_defaults` returns a reference which is only copied when it's provided as an argument to `add`. The second case works and the first doesn't when I'd have thought they were equivalent.

A branch demonstrating this in two commits is at https://github.com/cj-dev/certbot/compare/cj-dev:4435-cli-defaults...4435-but-broken. Commit [2e30494](`https://github.com/cj-dev/certbot/commit/2e3049417d7a9a13cd3f8c19414baa19d14626b9`) works, commit [b9df87c](https://github.com/cj-dev/certbot/commit/b9df87ccfb78f9af9455377fd22657b53944f252) doesn't. Only `python certbot/tests/cli_test.py ParseTest.test_cli_ini_domains` is necessary to see the effect.

cj-dev:
Sorry I fell off the face of the planet. I didn't spend the time I wanted to exploring Python object references, but as consolation, I wrote a small test to ensure the code doesn't regress to a state where it's easy to accidentally change a cli default.

I'm satisfied with this if you are.

PR #5060, opened 2017-08-29
python3 renamings

BenediktS:
I needed to do these changez to compile letsencrypt-auto in python36.

Only the line  `in metadata['releases'].keys()`  changed the python2.7 behavior a little bit. 
In python2 this line will return a copy of the list.
In python3 it is still an iteration on the existing list.
The list is not very big so i don't think that it is a problem. But i am open for better solutions.

### DIFF
--- a/letsencrypt-auto-source/pieces/fetch.py
+++ b/letsencrypt-auto-source/pieces/fetch.py
@@ -20,7 +20,11 @@ from os.path import dirname, join
 import re
 from subprocess import check_call, CalledProcessError
 from sys import argv, exit
-from urllib2 import build_opener, HTTPHandler, HTTPSHandler, HTTPError
+try:
+    from urllib2 import build_opener, HTTPHandler, HTTPSHandler, HTTPError
+except ImportError:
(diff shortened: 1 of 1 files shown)


bmw:
Tests failed because `letsencrypt-auto-source/letsencrypt-auto` wasn't updated. You can fix this by running `letsencrypt-auto-source/build.py`. This is a bit of a convoluted process, but you can learn more about it [here](https://certbot.eff.org/docs/contributing.html#updating-certbot-auto-and-letsencrypt-auto).

@erikrose, do you have time to review this or should I?

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -70,6 +70,10 @@ def write(contents, dir, filename):
     with open(join(dir, filename), 'w') as file:
         file.write(contents)
 
+def writebytes(contents, dir, filename):
```
erikrose: I would try to eliminate the repetition between this and write(). We could just add an `is_bytes` kwarg to write(), for example.

bmw:
An alternate version of this PR landed in #5329 so I'm closing this one.

PR #5062, opened 2017-08-30
certbot: Update `renew` command output in quiet mode.

r5d:
* certbot/main.py (renew_cert): Update function.
* certbot/renewal.py (_renew_describe_results): Update function.
* certbot/tests/main_test.py (_test_renewal_common): Update method.
  Add optional arg `stdout`; Modify `mock_get_utilitiy`, `stdout`.
  (test_quiet_renew): Update method.

Addresses #5016

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -652,15 +652,18 @@ def renew_cert(config, plugins, lineage):
 
     notify = zope.component.getUtility(interfaces.IDisplay).notification
     if installer is None:
-        notify("new certificate deployed without reload, fullchain is {0}".format(
-               lineage.fullchain), pause=False)
+        if not config.quiet:
+            notify("new certificate deployed without reload, "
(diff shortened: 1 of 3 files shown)


certbot/main.py
```diff
@@ -652,15 +652,18 @@ def renew_cert(config, plugins, lineage):
 
     notify = zope.component.getUtility(interfaces.IDisplay).notification
     if installer is None:
-        notify("new certificate deployed without reload, fullchain is {0}".format(
-               lineage.fullchain), pause=False)
+        if not config.quiet:
```
bmw: I'm sorry I missed this earlier, but these `if not config.quiet` lines here and below aren't needed because the notification method is essentially a noop if `config.quiet` is set.

r5d: Revert changes.

PR #5075, opened 2017-09-05
Specify the minimum six version in ACME

bmw:
Specifying this here prevents confusing error messages when someone tries to use an older version of `six`.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -20,7 +20,7 @@ install_requires = [
     # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
     # will tolerate; see #2599:
     'setuptools>=1.0',
-    'six',
+    'six>=1.9.0',  # needed for python_2_unicode_compatible
 ]
 
 # env markers cause problems with older pip and setuptools


ynasser:
Assuming the last two tests pass, LGTM! I should have thought to include this when I introduced `six.python_2_unicode_compatible`. Thanks for fixing this.

bmw:
This missed the release train. Travis has actually taken the majority of my workday to complete and I've been wanting to start building the release with only 48+ hour old commits again anyway so devs and people running from `master` have a chance to report bugs.

Since this only helps packagers on older distros (which have already worked around the problem) and people not following our installation instructions using `pip`, I think it's safe to wait until 0.19.0 to ship this.

bmw:
Also,

> I should have thought to include this when I introduced six.python_2_unicode_compatible. Thanks for fixing this.

No problem! I reviewed the PR and I think I'm even the one who suggested using `six.python_2_unicode_compatible`.

bmw:
@ynasser, now that tests have finished, can you give this PR an approving review so it can be merged?

PR #5086, opened 2017-09-07
Fix permissions error when upgrading certbot-auto.

### LINKED ISSUE #5085: certbot-auto permissions error

Debian Stretch (although every OS is affected)

certbot-auto

```
$ ./certbot-auto                  
Upgrading certbot-auto 0.17.0 to 0.18.0...
Replacing certbot-auto...
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 2375, in <module>
    main()
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 724, in main
    symlink=options.symlink)
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 946, in create_environment
    site_packages=site_packages, clear=clear, symlink=symlink))
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 1157, in install_python
    mkdir(lib_dir)
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 329, in mkdir
    os.makedirs(path)
  File "/usr/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/opt/eff.org'
```

It crashes.

The problem that occurs when upgrading is older versions of `certbot-auto` download a new version of themselves during what's called in the script "Phase 1". After this, the script is rerun with `--le-auto-phase2` and the new version of the script assumes that if you're in "Phase 2", we've already ensured we have root access. This isn't true when "Phase 1" is completed by an older version of the script.

bmw:
Fixes #5085.

Now we always check if we have root access if `--cb-auto-has-root` is not given on the command line. This allows `certbot-auto` to properly acquire root when upgrading from an older version. People upgrading from 0.18.0 to 0.18.1+ may check for root access twice, however, if root's user ID is 0, this check is essentially a noop. If root's user ID is not 0, we'll request root access a 2nd time during this upgrade.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -187,8 +187,7 @@ SetRootAuthMechanism() {
 
 if [ "$1" = "--cb-auto-has-root" ]; then
   shift 1
-elif [ "$1" != "--le-auto-phase2" ]; then
-  # if $1 is --le-auto-phase2, we've executed this branch before
+else
   SetRootAuthMechanism
   if [ -n "$SUDO" ]; then
     echo "Requesting to rerun $0 with root privileges..."
@@ -196,6 +195,14 @@ elif [ "$1" != "--le-auto-phase2" ]; then
     exit 0
   fi
 fi
+
+# Runs this script again with the given arguments. --cb-auto-has-root is added
+# to the command line arguments to ensure we don't try to acquire root a
+# second time. After the script is rerun, we exit the current script.
+RerunWithArgs() {
+    "$0" --cb-auto-has-root "$@"
+    exit 0
+}
 
 BootstrapMessage() {
   # Arguments: Platform name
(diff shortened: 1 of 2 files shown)


pde:
Am I correct in concluding that, during upgrades, we would acquire root twice, but not more than twice?

bmw:
Only twice and only when upgrading from exactly 0.18.0 to a newer version. When we upgrade in the future `--cb-auto-has-root` will be included on the command line when the script is rerun.

pde:
LGTM I think but ideally we would test all the cases we now have. Did we get it right for: 
* 0.18.0 -> 0.18.1
* 0.18.1 -> future X
* 0.17.0/0.18.0 with a broken attempt at updating -> 0.18.1

pde:
I'm going to mark this ready to merge. It would be best to have explicit tests for all of these cases, but I know they may be quite hard to create so I'll leave it to your judgement about the trade-offs there.

bmw:
I hacked together some test farm tests per @pde's suggestion and all tests passed.

PR #5097, opened 2017-09-08
add info about -d

### LINKED ISSUE #1376: the help message should be more explicit about the effect of the -d option
#### Hypothetical language

A suggestion, possibly mostly correct:

If the -d option is mentioned more than once on the command line, a _single_ certificate will be created to cover all the domains that were mentioned.  In particular:
- The first-mentioned domain will be the Subject CN of the certificate.
- All the domains (including the first) will be version-3 Subject Alternative Names on the certificate.
- The certificate will be stored with a filename derived from the first-mentioned domain.
#### Question

Suppose there are multiple inequivalent certificates having the same first domain but different additional domains.  There is the potential for conflict with the filename.  How is this conflict resolved?
#### Suggestion

The stated goal is to make the process easy.  To this end, it would be helpful for somebody who understands the process to document it in the help message.

SwartzCr:
fixes #1376

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -891,11 +891,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         metavar="DOMAIN", action=_DomainsAction, default=[],
         help="Domain names to apply. For multiple domains you can use "
              "multiple -d flags or enter a comma separated list of domains "
-             "as a parameter. The first provided domain will be used in "
-             "some software user interfaces and file paths for the "
+             "as a parameter. The first domain provided will be the "
+             "subject CN of the certificate, and all domains will be"
+             "Subject Alternative Names on the certificate. "
+             "The first domain will also be used in "
+             "some software user interfaces and as the file paths for the "
              "certificate and related material unless otherwise "
-             "specified or you already have a certificate for the same "
-             "domains. (default: Ask)")
+             "specified or you already have a certificate with the same "
+             "name. In that case it will append a number like 0001 "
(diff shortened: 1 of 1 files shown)


certbot/cli.py
```diff
@@ -891,11 +891,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+             "as a parameter. The first domain provided will be the "
+             "subject CN of the certificate, and all domains will be"
+             "Subject Alternative Names on the certificate. "
+             "The first domain will also be used in "
+             "some software user interfaces and as the file paths for the "
              "certificate and related material unless otherwise "
-             "specified or you already have a certificate for the same "
-             "domains. (default: Ask)")
+             "specified or you already have a certificate with the same "
+             "name. In that case it will append a number like 0001 "
```
bmw: nit: I personally think it's clearer to make the last sentence something like "In case of name collisions..." to explicitly state what case causes this.

certbot/cli.py
```diff
@@ -891,11 +891,15 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         metavar="DOMAIN", action=_DomainsAction, default=[],
         help="Domain names to apply. For multiple domains you can use "
              "multiple -d flags or enter a comma separated list of domains "
-             "as a parameter. The first provided domain will be used in "
-             "some software user interfaces and file paths for the "
+             "as a parameter. The first domain provided will be the "
+             "subject CN of the certificate, and all domains will be"
```
bmw: You're missing a space at the end of this line.

PR #5101, opened 2017-09-11
fix dns-rfc2136 plugin not respecting cnames

webczat:
The plugin does not work if the domain of a certificate is found to have a cname record in dns.
That is because when plugin tries to find zone boundary, it searches from the domain up for the SOA record, and each DNS response is checked for the answer being empty, assuming that empty answer means no SOA record is present and the higher level domain has to be checked, and non empty answer section means that this domain is a zone root.
However, if the initial domain, or any upper level domain except the zone root has a cname record pointing to the zone root, then the server will, instead of returning an empty answer, return one containing two records, first a cname pointing to the zone root, then the SOA record of zone root, and that will make the check fail and use a wrong domain as a zone name during update.
Fix that by replacing a check for empty answer with explicitly searching in response's answer section for a SOA record matching the domain that is being checked.

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -208,7 +208,7 @@ class _RFC2136Client(object):
             rcode = response.rcode()
 
             # Authoritative Answer bit should be set
-            if (rcode == dns.rcode.NOERROR and len(response.answer) > 0 and
+            if (rcode == dns.rcode.NOERROR and response.get_rrset(response.answer, domain, dns.rdataclass.IN, dns.rdatatype.SOA) and
                     response.flags & dns.flags.AA):
                 logger.debug('Received authoritative SOA response for %s', domain_name)
                 return True
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -208,8 +208,8 @@ class _RFC2136Client(object):
             rcode = response.rcode()
 
             # Authoritative Answer bit should be set
-            if (rcode == dns.rcode.NOERROR and response.get_rrset(response.answer, domain, dns.rdataclass.IN, dns.rdatatype.SOA) and
-                    response.flags & dns.flags.AA):
+            if (rcode == dns.rcode.NOERROR and response.get_rrset(response.answer,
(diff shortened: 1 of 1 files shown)


joohoi:
Thanks for the PR! 
There seems to be a linter error; 
https://travis-ci.org/certbot/certbot/jobs/274016941#L1177-L1178

I'll start reviewing this PR nevertheless.

joohoi:
I'm not sure if I'm misunderstanding the issue, but having CNAME coexisting in same domain with SOA record is a RFC violation: https://tools.ietf.org/html/rfc2181#section-10

webczat:
not same domain, but subdomain of it.
For example if there is a domain webczatnet.pl (my own domain) that is a zone boundary with a SOA record, and you have a domain for example bin.webczatnet.pl with record
bin.webczatnet.pl IN CNAME webczatnet.pl.
And then you do a lookup like the dns-rfc2136 plugin does to find the soa
bin.webczatnet.pl IN SOA
Then you don't get an empty response as for no soa record for bin.webczatnet.pl, you get redirected. The answer contains a cname record pointing to webczatnet.pl, then the soa record of webczatnet.pl. That violates the zone root finding logic of the plugin. It tries to find a root of the zone by locating it's soa, but it assumes it found the zone root when it gets a non empty answer section. but in this case it will be non empty even though it is not a zone root. it should then either go up (that is what this plugin does), or replace the domain of the lookup with the owner of the SOA record returned.

joohoi:
Ah, I see. That's true, my apologies. I was testing that piece of functionality against Cloudflare DNS servers, and they don't return the CNAME response for SOA query so had trouble with reproducing the error. Good old BIND9 however was able to reproduce both the error, and the fix in this PR. 

LGTM! 
+1 needed for merge however.

webczat:
to be honest, my fix is suboptimal as it could actually, instead of treating a mismatched soa as no soa and going up, take the soa owner domain and just replace the input domain with this one. But it works. If cloudflare doesn't return a cname and upper soa, doesn't it itself violate the rfc?

joohoi:
There's always space for improvement, one other use case that comes to mind is that if user has static CNAME _acme-challenge.example.com pointing to another DNS zone in order to not to use & save credentials that allow control of the actual zone (Let's Encrypt follows these CNAMES). That CNAME is not followed by Certbot in order to resolve the SOA of the zone that's actually in control of the record. But these are of course matter of another issue and PR. 

Thanks again for the PR!

webczat:
I know nsupdate can work in all those conditions, but I believe it would follow cnames, as it does only one lookup in such a case and immediately takes the owner of soa record.

Issue #4350, opened 2017-03-15
"certbot plugins" should not throw Permission denied


jsha:
Steps to reproduce:

1. As non-root, run "certbot plugins"

Expected result:

Get a list of plugins.

Actual result:

```
The following error was encountered:
[Errno 13] Permission denied: '/var/log/letsencrypt/letsencrypt.log'
If running as non-root, set --config-dir, --work-dir, and --logs-dir to writeable paths.
```

SwartzCr:
This sounds a lot like what we did with `certbot help` to allow it to run without venv-ing itself.
@ynasser Do you want to think about adding this to things people can work on at the sprint?

SwartzCr:
Still no progress on this, but it seems like an easy fix, @bmw wanna put it in a milestone?

SwartzCr:
similarly `certbot --version` should get the same treatment as per #4491

bmw:
`certbot --version` should work without root.

r5d:
@SwartzCr,

If I'm correctly understanding the issue, it looks like, the `plugins_cmd` must be called before the call to `make_or_verify_needed_dirs(config)` in `certbot.main.main` function. Where would be good place to do this?

bmw:
@SwartzCr asked me to respond to this.

@r5d, the problematic functions are `certbot.main.make_or_verify_needed_dirs` (as you correctly pointed out) and `certbot.log.post_arg_parse_setup`.

As for how to accomplish this, it might be simplest to just do something like:
```
try:
    log.post_arg_parse_setup(config)
    make_or_verify_needed_dirs(config)
except errors.Error:
    if config.func != plugins_cmd:
        raise
```
While this feels a little hacky, the "plugins" subcommand is the exception here so moving this code into the other individual subcommands seems like a recipe for bugs when new subcommands are added in the future. Additionally, I think we should always try to run this code because when it's not run successfully, logs are written to a temporary file instead of the normal log directory. We should try to run it for the times when we do have the necessary permissions.

I'm open to other ideas if you have them though.

r5d:
@bmw, I've submitted a PR. It uses your idea.

bmw:
Fixed in #5103.

SwartzCr:
Awesome job @r5d ! Any interest in working on another issue? Maybe #4879?

jsha:
Nice, thanks for fixing this bug!

r5d:
@SwartzCr, I'll look into #4879.

PR #5103, opened 2017-09-11
certbot: Let plugins_cmd be run as un-priviliged user.

r5d:
* certbot/main.py (main): Update function.

Addresses issue #4350.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -743,8 +743,14 @@ def main(cli_args=sys.argv[1:]):
     config = configuration.NamespaceConfig(args)
     zope.component.provideUtility(config)
 
-    log.post_arg_parse_setup(config)
-    make_or_verify_needed_dirs(config)
+    try:
+        log.post_arg_parse_setup(config)
+        make_or_verify_needed_dirs(config)
+    except errors.Error:
+        # Let plugins_cmd be run as un-privileged user.
+        if config.func != plugins_cmd:
+            raise
+
     set_displayer(config)
 
     # Reporter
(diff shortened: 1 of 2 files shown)


certbot/main.py
```diff
@@ -743,8 +743,14 @@ def main(cli_args=sys.argv[1:]):
...
     zope.component.provideUtility(config)
 
-    log.post_arg_parse_setup(config)
-    make_or_verify_needed_dirs(config)
+    try:
+        log.post_arg_parse_setup(config)
+        make_or_verify_needed_dirs(config)
+    except errors.Error:
+        # Let plugins_cmd be run as un-privileged user.
+        if config.func != plugins_cmd:
```
SwartzCr: I don't see plugins_cmd defined anywhere - have you tested this code?
Also it would be good to write tests for this.

r5d: @SwartzCr, `plugins_cmd` function is defined in `certbot.main`.

I've added a test, see if it is ok.

certbot/tests/main_test.py
```diff
@@ -547,6 +547,26 @@ def test_plugins_no_args(self, _det, mock_disco):
 
     @mock.patch('certbot.main.plugins_disco')
     @mock.patch('certbot.main.cli.HelpfulArgumentParser.determine_help_topics')
+    def test_plugins_no_args_unprivileged(self, _det, mock_disco):
```
r5d: Notes:

When `certbot plugins` is run as an unprivileged user:

* `certbot.util.set_up_core_dir` is called via `log.post_arg_parse_setup(config)` -> `setup_log_file_handler`.
* `certbot.util.set_up_core_dir` then throws `certbot.errors.Errors`.
*  `certbot.errors.Errors` is caught in `certbot.main` and is not raised when the `config.func` is `plugins_cmd`. Consequently `plugins_cmd` runs.

The test tries to simulate this by mocking `certbot.util.set_up_core_dir` and making it throw `certbot.errors.Errors.`

PR #5117, opened 2017-09-15
certbot-dns-google: automatic credential lookup

lobeck:
When running on Google Cloud, credentials can be automatically provided using the metadata service. The current changes allow for that. We only need to query for the project-id, the credentials itself are already handled by `googleapiclient` internally.

We don't need anything from the user, when the metadata service is available and providing a token we can immediately use it and don't require credential files.

### DIFF
--- a/certbot-dns-google/certbot_dns_google/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/dns_google.py
@@ -2,6 +2,7 @@
 import json
 import logging
 
+import httplib2
 import zope.interface
 from googleapiclient import discovery
 from googleapiclient import errors as googleapiclient_errors
@@ -15,6 +16,8 @@ logger = logging.getLogger(__name__)
 
 ACCT_URL = 'https://developers.google.com/identity/protocols/OAuth2ServiceAccount#creatinganaccount'
 PERMISSIONS_URL = 'https://cloud.google.com/dns/access-control#permissions_and_roles'
+METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1/'
+METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
 
 
 @zope.interface.implementer(interfaces.IAuthenticator)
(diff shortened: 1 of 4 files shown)


certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -183,3 +204,22 @@ def _find_managed_zone_id(self, domain):
 
         raise errors.PluginError('Unable to determine managed zone for {0} using zone names: {1}.'
                                  .format(domain, zone_dns_name_guesses))
+
+    @staticmethod
+    def get_project_id():
+        """
+        Query the google metadata service for the current project ID This only works on
+        Google Cloud Platform and raises a ServerNotFoundError if metadata.google.internal
+        is not available
```
bmw: Thanks for writing meaningful documentation here! [PEP 257](https://www.python.org/dev/peps/pep-0257) (which is included in both of our [style guides](https://certbot.eff.org/docs/contributing.html#coding-style)) says that multi-line docstrings should have a one line summary, a blank line, and then a longer description. Ideally we should follow this convention here as well.

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -183,3 +204,22 @@ def _find_managed_zone_id(self, domain):
...
+        :raises: ValueError, ServerNotFoundError
```
bmw: Unfortunately, this line is causing formatting issues in the documentation. If you followed your our [developer installation instructions](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client) and have activated your virtual environment, you can build the documentation yourself by running `make -C docs clean html` from the `certbot/certbot-dns-google` directory.

To fix the error, the syntax here should be:
```
:raises ExceptionType: description of when it is raised
```
This is syntax is described [here](http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists). There should be one line for each exception type.

certbot-dns-google/certbot_dns_google/dns_google_test.py
```diff
@@ -50,6 +52,22 @@ def test_cleanup(self):
...
+
+        super(AuthenticatorTest, self).setUp()
+        self.config = mock.MagicMock(google_credentials=None,
+                                     google_propagation_seconds=0)  # don't wait during tests
+
+        self.auth = Authenticator(self.config, "google")
+
+        self.mock_client = mock.MagicMock()
+        # _get_google_client | pylint: disable=protected-access
+        self.auth._get_google_client = mock.MagicMock(return_value=self.mock_client)
```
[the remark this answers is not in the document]
lobeck: I simplified it a bit, is this better?

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -2,6 +2,7 @@
 import json
 import logging
 
+import httplib2
```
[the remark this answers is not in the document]
lobeck: done.

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -183,3 +204,23 @@ def _find_managed_zone_id(self, domain):
...
                                  .format(domain, zone_dns_name_guesses))
+
+    @staticmethod
+    def get_project_id():
+        """
+        Query the google metadata service for the current project ID
+
+        This only works on Google Cloud Platform
+        :raises: ServerNotFoundError: Not running on Google Compute or DNS not available
+        :raises: ValueError: Server is found, but response code is not 200
```
bmw: Whoops! There shouldn't be a colon directly after `raises` here.

certbot-dns-google/certbot_dns_google/dns_google_test.py
```diff
@@ -50,6 +52,16 @@ def test_cleanup(self):
         expected = [mock.call.del_txt_record(DOMAIN, '_acme-challenge.'+DOMAIN, mock.ANY, mock.ANY)]
         self.assertEqual(expected, self.mock_client.mock_calls)
 
+    @mock.patch('httplib2.Http.request', side_effect=ServerNotFoundError)
+    def test_without_auth(self, unused_mock):
+        from certbot_dns_google.dns_google import Authenticator
+
+        super(AuthenticatorTest, self).setUp()
+        self.config = mock.MagicMock(google_credentials=None)
```
bmw: What I personally had in mind here was just changing this to:
```
    @mock.patch('httplib2.Http.request', side_effect=ServerNotFoundError)
    def test_without_auth(self, unused_mock):
        self.config.google_credentials = None
        self.assertRaises(PluginError, self.auth.perform, [self.achall])
```
This works because `self.config` refers to the same object in the authenticator. You can also do `self.auth.config.google_credentials = None` which may make this cleaner/clearer.

I just think we can reuse the `self.auth` object already created for us by the `setUp` function.

lobeck:
@bmw Thanks for your support and feedback! I'm already working on it and updating the documentation. However `get_project_id()` won't show up in the Sphinx doc and i can't figure out why, can you give me a hint?

PR #5118, opened 2017-09-15
Certbot auto timeout

erikrose:
A continuation of #4929

### DIFF
--- a/letsencrypt-auto-source/pieces/fetch.py
+++ b/letsencrypt-auto-source/pieces/fetch.py
@@ -20,7 +20,9 @@ from os.path import dirname, join
 import re
 from subprocess import check_call, CalledProcessError
 from sys import argv, exit
-from urllib2 import build_opener, HTTPHandler, HTTPSHandler, HTTPError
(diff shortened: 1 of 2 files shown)


erikrose:
We'll need 2 review for this, since it's certbot-auto, of course.

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -56,7 +57,9 @@ def get(self, url):
 
         """
         try:
-            return self._opener.open(url).read()
+            // socket module docs say default timeout is None: that is, no
```
bmw: *gasps*

Erik's been cheating on Python! That doesn't look like Pythonic comment syntax to me.

erikrose: How embarrassing! I thought it didn't look right, but I couldn't put my finger on it!

erikrose:
Of course, rebasing will dissociate this PR from the earlier one, but if it makes you happy, I'm happy to make you happy.

PR #5090, opened 2017-09-08
Stop using print in `certbot.main` module.

r5d:
* Update certbot.main.plugins_cmd` function.

* Update test methods `test_plugins_no_args`, `test_plugins_init` and
`test_plugins_prepare` in `cerbot.tests.MainTest` class.

Addresses #3720.

PR #5121, opened 2017-09-19
Stop using print in `certbot.main` module.

r5d:
* Update certbot.main.plugins_cmd` function.

* Update test methods `test_plugins_no_args`,
`test_plugins_no_args_unprivileged`, `test_plugins_init` and
`test_plugins_prepare` in `cerbot.tests.MainTest` class.

Addresses #3720.

### DIFF
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -278,16 +278,16 @@ def _create_get_utility_mock_with_stdout(stdout):
     def _write_msg(message, *unused_args, **unused_kwargs):
         """Write to message to stdout.
         """
-        stdout.write(message)
+        if message:
+            stdout.write(message)
 
     def mock_method(*args, **kwargs):
         """
         Mock function for IDisplay methods.
         """
-        message = args[0] if args else kwargs.get('message', None)
-        if message:
-            _write_msg(message, args, kwargs)
         _assert_valid_call(args, kwargs)
+        _write_msg(*args, **kwargs)
+
 
     display = FreezableMock()
     for name in interfaces.IDisplay.names():  # pylint: disable=no-member
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -14,6 +14,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 import mock
 import OpenSSL
+import six
 from six.moves import reload_module  # pylint: disable=import-error
 
 from acme import jose
@@ -169,6 +170,26 @@ def patch_get_utility(target='zope.component.getUtility'):
     return mock.patch(target, new_callable=_create_get_utility_mock)
 
 
+def patch_get_utility_with_stdout(target='zope.component.getUtility',
+                                  stdout=six.StringIO()):
+    """Patch zope.component.getUtility to use a special mock IDisplay.
+
+    The mock IDisplay works like a regular mock object, except it also
+    also asserts that methods are called with valid arguments.
+
+    :param str target: path to patch
+    :param object stdout: object to write standard output to; it is
+        expected to have a `write` method
+
+    :returns: mock zope.component.getUtility
+    :rtype: mock.MagicMock
+
+    """
+
+    freezable_mock = _create_get_utility_mock_with_stdout(stdout)
+    return mock.patch(target, new=freezable_mock)
+
+
 class FreezableMock(object):
     """Mock object with the ability to freeze attributes.
 
@@ -246,6 +267,27 @@ def _create_get_utility_mock():
             frozen_mock = FreezableMock(frozen=True, func=_assert_valid_call)
             setattr(display, name, frozen_mock)
     display.freeze()
+    return FreezableMock(frozen=True, return_value=display)
+
+
+def _create_get_utility_mock_with_stdout(stdout):
+    def _write_msg(message, *unused_args, **unused_kwargs):
+        """Write to message to stdout.
+        """
+        stdout.write(message)
+
+    display = FreezableMock()
+    for name in interfaces.IDisplay.names():  # pylint: disable=no-member
+        if name == 'notification':
+            frozen_mock = FreezableMock(frozen=True,
+                                        func=_write_msg)
+            setattr(display, name, frozen_mock)
+        else:
+            frozen_mock = FreezableMock(frozen=True,
+                                        func=_assert_valid_call)
+            setattr(display, name, frozen_mock)
+    display.freeze()
+
     return FreezableMock(frozen=True, return_value=display)
 
 
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -171,11 +171,14 @@ def patch_get_utility(target='zope.component.getUtility'):
 
 
 def patch_get_utility_with_stdout(target='zope.component.getUtility',
-                                  stdout=six.StringIO()):
+                                  stdout=None):
     """Patch zope.component.getUtility to use a special mock IDisplay.
 
     The mock IDisplay works like a regular mock object, except it also
     also asserts that methods are called with valid arguments.
+
+    The `message` argument passed to the IDisplay methods is passed to
+    stdout's write method.
 
     :param str target: path to patch
     :param object stdout: object to write standard output to; it is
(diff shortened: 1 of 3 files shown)


r5d:
From old PR (#5090):

@bmw writes:
> We ended up using code like this a lot more than I thought we would when I wrote [this comment](https://github.com/certbot/certbot/pull/5028#discussion_r134325650). Sorry for the mixed signals on this, but it'd be good if we factored it out to avoid duplication.

> What if we made another function in `certbot/tests/util.py` called something like `patch_get_utility_and_stdout` than returned both the mock `getUtility` and a `StringIO` object that all `IDisplay` methods will write to? My understanding based on [this](http://www.voidspace.org.uk/python/mock/patch.html#where-to-patch) is to ensure writing messages works in all cases, we need to either have the function return the `StringIO` object or have the caller `mock certbot.tests.util.sys.stdout` which feels much more awkward.

Does this look acceptable?

```diff
diff --git a/certbot/tests/util.py b/certbot/tests/util.py
index 69896251..a61d6321 100644
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -14,6 +14,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 import mock
 import OpenSSL
+import six
 from six.moves import reload_module  # pylint: disable=import-error
 
 from acme import jose
@@ -169,6 +170,14 @@ def patch_get_utility(target='zope.component.getUtility'):
     return mock.patch(target, new_callable=_create_get_utility_mock)
 
 
+def patch_get_utility_with_stdout(target='zope.component.getUtility',
+                                  stdout=None):
+    stdout = stdout if stdout else six.StringIO()
+
+    callable_ = _create_get_utility_mock_with_stdout(stdout)
+    return mock.patch('zope.component.getUtility', new_callable=callable_)
+
+
 class FreezableMock(object):
     """Mock object with the ability to freeze attributes.
 
@@ -249,6 +258,24 @@ def _create_get_utility_mock():
     return FreezableMock(frozen=True, return_value=display)
 
 
+def _create_get_utility_mock_with_stdout(stdout):
+    def _write_msg(message, *args, **kwargs):
+        _, _ = args, kwargs
+        stdout.write(message)
+
+    display = FreezableMock()
+    for name in interfaces.IDisplay.names():  # pylint: disable=no-member
+        if name == 'notification':
+            setattr(display, name, _write_msg)
+        else:
+            frozen_mock = FreezableMock(frozen=True,
+                                        func=_assert_valid_call)
+            setattr(display, name, frozen_mock)
+    display.freeze()
+
+    return lambda: FreezableMock(frozen=True, return_value=display)
+
+

```

With the above changes, the `certbot.tests.main_test.MainTest.test_plugins_no_args` will look like:

```python
...
    @mock.patch('certbot.main.plugins_disco')
    @mock.patch('certbot.main.cli.HelpfulArgumentParser.determine_help_topics')
    def test_plugins_no_args(self, _det, mock_disco):
        ifaces = []
        plugins = mock_disco.PluginsRegistry.find_all()

        stdout = six.StringIO()
        with test_util.patch_get_utility_with_stdout(stdout=stdout) \
          as mock_get_utility:
            _, stdout, _, _ = self._call(['plugins'], stdout)

        plugins.visible.assert_called_once_with()
        plugins.visible().ifaces.assert_called_once_with(ifaces)
        filtered = plugins.visible().ifaces()
        self.assertEqual(stdout.getvalue().strip(), str(filtered))
...
```

```bash
(venv)  ~/v/git/others/certbot/certbot/tests 
↳ python main_test.py MainTest.test_plugins_no_args                                                                                                                                                       s@cygnus 
.                                                  
Ran 1 test in 0.078s

OK

```

bmw:
I have a couple of small changes I'd personally like to see, but otherwise, something like that looks good! I like the approach you came up with of using `sys.stdout` or an argument passed into the function better than what I suggested of having the function return both the patch and stdout.

The small changes I'd like to see are:

1. We merged the same code in another PR, but instead of using `_, _ = args, kwargs` to stop warnings from our linter, let's either rename `args` and `kwargs` to `unused_args` and `unused_kwargs`, or add `# pylint: disable=unused-argument`.
2. We only write the message to stdout for the notification method. Rather than possibly adding additional parameters in the future to change this behavior (or just suddenly making the change), I think we should do it for all `IDisplay` methods. The way I'd do this is make another local function that takes `*args` and  `**kwargs`, passes them to `_assert_valid_call`, and then passes them to `write_msg`.
3. I think the approach of setting `write_msg` as the `notification` method directly prevents people from using normal `mock` functionality on that method such as making assertions about call count, parameters, etc. Instead, I'd use the `func` parameter to the `FreezableMock` like we do for other methods.
4. I personally don't like the asymmetry between what `_create_get_utility_mock` and `_create_get_utility_mock_with_stdout` return. One returns a `FreezableMock` object and one returns a function that returns a `FreezableMock` object. Instead of this, I'd make `patch_get_utility_with_stdout` call `_create_get_utility_mock_with_stdout` and pass the return value as the `new` argument to [mock.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).

r5d:
@bmw, I pushed changes based on your feedback in `1` and `3` and `4`. See if it's ok.

`2`, `_assert_valid_call` fails for the `notification` method. Should the `certbot.display.util.assert_valid_call` (that  `_assert_valid_call` calls) be updated so that it passes for the `notification` method? Relatedly, `_create_get_utility_mock` does not call `_assert_valid_call` for the `notification` method.

certbot/tests/util.py
```diff
@@ -169,6 +170,26 @@ def patch_get_utility(target='zope.component.getUtility'):
     return mock.patch(target, new_callable=_create_get_utility_mock)
 
 
+def patch_get_utility_with_stdout(target='zope.component.getUtility',
+                                  stdout=six.StringIO()):
```
bmw: I liked your previous approach of using `None` here better. The problem with this is this is the default is set to a specific `StringIO` object when the function is defined so different calls to the function using this default will get the same object causing interactions between tests.

certbot/tests/util.py
```diff
@@ -169,6 +170,26 @@ def patch_get_utility(target='zope.component.getUtility'):
...
+    """Patch zope.component.getUtility to use a special mock IDisplay.
+
+    The mock IDisplay works like a regular mock object, except it also
+    also asserts that methods are called with valid arguments.
```
bmw: nit: I'd also add a sentence describing the behavior of messages written to stdout here.

r5d: Added a line about argument `stdout` in the docstring.

certbot/tests/util.py
r5d: Updated it to default to `None`.

r5d:
@bmw, pushed changes related to `2`. See if it is ok.

certbot/tests/util.py
```diff
@@ -276,6 +276,15 @@ def _write_msg(message, *unused_args, **unused_kwargs):
         """
         stdout.write(message)
 
+    def mock_method(*args, **kwargs):
+        """
+        Mock function for IDisplay methods.
+        """
+        message = args[0] if args else kwargs.get('message', None)
```
bmw: I think you should be able to just call `_write_msg(*args, **kwargs)` and Python should properly handle assigning parameters to variables.

certbot/tests/util.py
```diff
@@ -276,6 +276,15 @@ def _write_msg(message, *unused_args, **unused_kwargs):
...
+        if message:
```
bmw: If we want to avoid writing the message if it's empty, maybe we should do this in `_write_msg` itself?

certbot/tests/util.py
```diff
@@ -276,6 +276,15 @@ def _write_msg(message, *unused_args, **unused_kwargs):
...
+            _write_msg(message, args, kwargs)
+        _assert_valid_call(args, kwargs)
```
bmw: Unless there's another reason to do it this way that I'm missing, I think we should call `_assert_valid_call` first as this is the behavior of the methods in `certbot/display/util.py`.

Issue #4245, opened 2017-02-22
certbot fails with "AttributeError: 'NoneType' object has no attribute 'strip'"


scubanarc:
I did the following:
<pre>
sudo yum update
sudo yum install httpd mod_ssl python-certbot-apache
sudo systemctl start httpd
systemctl status httpd
sudo certbot --apache -d example.com -d www.example.com
</pre>

certbot --version
  certbot 0.9.3

rpm --query centos-release
  centos-release-7-3.1611.el7.centos.x86_64

Command line doesn't seem to matter. I tried:
  certbot
  certbot --apache
  sudo certbot --apache -d example.com -d www.example.com

In all cases certbot fails almost immediately with the error:

"AttributeError: 'NoneType' object has no attribute 'strip'"

It shows the blue window and immediately fails.

Here's the letsencrypt.log:

<pre>
2017-02-22 20:21:45,928:DEBUG:certbot.main:Root logging level set at 20
2017-02-22 20:21:45,928:INFO:certbot.main:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2017-02-22 20:21:45,932:DEBUG:certbot.main:certbot version: 0.9.3
2017-02-22 20:21:45,932:DEBUG:certbot.main:Arguments: ['--apache', '-d', 'domain.com', '-d', 'www.domain.com']
2017-02-22 20:21:45,932:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#webroot,PluginEntryPoint#null,PluginEntryPoint#manual,PluginEntryPoint#standalone)
2017-02-22 20:21:45,934:DEBUG:certbot.plugins.selection:Requested authenticator apache and installer apache
2017-02-22 20:21:46,202:DEBUG:certbot.main:Exiting abnormally:
Traceback (most recent call last):
  File "/bin/certbot", line 9, in <module>
    load_entry_point('certbot==0.9.3', 'console_scripts', 'certbot')()
  File "/usr/lib/python2.7/site-packages/certbot/main.py", line 776, in main
    return config.func(config, plugins)
  File "/usr/lib/python2.7/site-packages/certbot/main.py", line 502, in run
    installer, authenticator = plug_sel.choose_configurator_plugins(config, plugins, "run")
  File "/usr/lib/python2.7/site-packages/certbot/plugins/selection.py", line 185, in choose_configurator_plugins
[...]

PR #5135, opened 2017-09-26
Handle NoneType from Augeas better in Apache parser get_arg

joohoi:
Handles `None` from aug.get() in a sane way.

I'm still unsure why it happens, but from the Certbot point of view, this is the correct check to add.

Fixes: #4245


joohoi:
Empty values do not get added to modules anymore and a message is logged in such case. I'd love to be able to dig in to the underlying problem if only I could reproduce the actual error in a sane way. 

However I think this solves the rare edge case and is the correct thing to do to be able to continue execution.

Issue #4371, opened 2017-03-17
Automatically use a mirror if pypi.python.org is blocked


pde:
If people try to run `certbot-auto` in certain countries, they may encounter #2516. Perhaps certbot-auto could detect that, and try some mirrors?

This should be safe, since we're checking hashes on all of our python dependencies.

kevinkle:
Being discuessed in https://github.com/certbot/certbot/pull/4711

SwartzCr:
@kevinkle did you end up opening an issue in pipstrap for this?

kevinkle:
Yup, it's at https://github.com/erikrose/pipstrap/issues/8

SwartzCr:
@erikrose does https://github.com/erikrose/pipstrap/commit/19824c6011ccd083c7f182089004cf0d283779c3 close this?

erikrose:
@SwartzCr Partly. I lost momentum when [I realized](https://github.com/erikrose/pipstrap/pull/10#issuecomment-314830591) that it solves only half the problem: the half which doesn't use pip itself. So some work remains to be done.

bmw:
Some work on this was done in #5158 and the code was saved in the [branch pypi-mirrors](https://github.com/certbot/certbot/tree/pypi-mirrors).

PR #5158, opened 2017-10-02
Let certbot-auto work behind the Great Firewall of China. Fix #3222. Ref #4371.

### LINKED ISSUE #3222: Allow custom "pip.conf" for certbot-auto
Those doomed to deal with corporate proxies usually setup a local PyPi mirror that goes configured in "/etc/pip.conf".

Problem is certbot-auto creates a virtualenv that seems to ignore it. There is no way to add the file manually to virtualenv because certbot-auto recreates it everytime.

Could certbot-auto accept a parameter (say, "--pipconf") that could point to a pip.conf file that could be copied into the virtualenv it creates?

erikrose:
Let certbot-auto work in China without needing to get pip >= 8 by hand first and without needing to muck with pip.conf.

We do it by upgrading to a new version of pipstrap which understands pip's own PIP_INDEX_URL env var for specifying a mirror. That done, certbot-auto simply sets that var to a known Chinese mirror in response to a CLI flag, and off we go.

This could be made automatic in the future: see 2nd-last paragraph of https://github.com/erikrose/pipstrap/pull/10#issuecomment-333637736. But there's no reason not to land this simple version first.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1025,6 +1025,11 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
              " installing OS-level dependencies (default: Prompt to install "
              " OS-wide dependencies, but exit if the user says 'No')")
     helpful.add(
+        "automation", "--pypi-mirror", action="store_true",
+        default=flag_default("pypi_mirror"),
+        help="(certbot-auto only) use an alternative Python package server "
+             "that is reachable from China")
+    helpful.add(
         ["automation", "renew", "certonly", "run"],
         "-q", "--quiet", dest="quiet", action="store_true",
         default=flag_default("quiet"),
--- a/certbot/constants.py
+++ b/certbot/constants.py
@@ -44,6 +44,7 @@ CLI_DEFAULTS = dict(
     os_packages_only=False,
     no_self_upgrade=False,
     no_bootstrap=False,
+    pypi_mirror=False,
     quiet=False,
     staging=False,
     debug=False,
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -47,6 +47,7 @@ Help for certbot itself cannot be provided until it is installed.
   --no-bootstrap                            do not install OS dependencies
   --no-self-upgrade                         do not download updates
   --os-packages-only                        install OS dependencies and exit
+  --pypi-mirror                             use alternative Python package server
   -v, --verbose                             provide more output
   -q, --quiet                               provide only update/error output;
                                             implies --non-interactive
@@ -66,6 +67,8 @@ for arg in "$@" ; do
       NO_SELF_UPGRADE=1;;
     --no-bootstrap)
       NO_BOOTSTRAP=1;;
+    --pypi-mirror)
+      export PIP_INDEX_URL="https://pypi.doubanio.com/simple";;
     --help)
       HELP=1;;
     --noninteractive|--non-interactive|renew)
@@ -1118,6 +1121,7 @@ anything goes wrong, it will exit with a non-zero status code.
 from __future__ import print_function
 from distutils.version import StrictVersion
 from hashlib import sha256
+from os import environ
 from os.path import join
 from pipes import quote
 from shutil import rmtree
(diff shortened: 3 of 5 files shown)


erikrose:
Other true miscellany determined in the course of my research:
* Saying `--index-url` (even without disabling update checking) prevents hitting pypi.python.org.
* Having an unreachable PyPI address hard-coded in pip (as pypi.python.org is) or specified as `--index-url` will totally scuttle any success; pip will not go on to try anything specified with `--extra-index-url`. So we can't just stack up a bunch of mirrors and expect pip to go down the list.

bmw:
I tested this PR on the Centos 6 Docker image. I saw:
```
$ ./letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade --pypi-mirror
...
pip prints the following errors: 
=====================================================
DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
Requirement already satisfied: argparse==1.4.0 in /opt/eff.org/certbot/venv/lib/python2.6/site-packages (from -r /tmp/tmp.Zi1r38wsAz/letsencrypt-auto-requirements.txt (line 11))
Collecting pycparser==2.14 (from -r /tmp/tmp.Zi1r38wsAz/letsencrypt-auto-requirements.txt (line 17))
/opt/eff.org/certbot/venv/lib/python2.6/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
/opt/eff.org/certbot/venv/lib/python2.6/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
  Certificate did not match expected hostname: pypi.doubanio.com. Certificate: {'notAfter': 'Jun 11 23:59:59 2018 GMT', 'subjectAltName': ((u'DNS', '*.ucdl.pp.uc.cn'),), 'subject': ((('countryName', u'CN'),), (('stateOrProvinceName', u'Guangdong'),), (('localityName', u'Guangzhou'),), (('organizationName', u'\u5e7f\u5dde\u5e02\u52a8\u666f\u8ba1\u7b97\u673a\u79d1\u6280\u6709\u9650\u516c\u53f8'),), (('organizationalUnitName', u'IT DEPT'),), (('commonName', u'*.ucdl.pp.uc.cn'),))}
  Could not fetch URL https://pypi.doubanio.com/simple/pycparser/: There was a problem confirming the ssl certificate: hostname 'pypi.doubanio.com' doesn't match '*.ucdl.pp.uc.cn' - skipping
  Could not find a version that satisfies the requirement pycparser==2.14 (from -r /tmp/tmp.Zi1r38wsAz/letsencrypt-auto-requirements.txt (line 17)) (from versions: )
No matching distribution found for pycparser==2.14 (from -r /tmp/tmp.Zi1r38wsAz/letsencrypt-auto-requirements.txt (line 17))
=====================================================
...
```
TLS is hard.

How did you pick this mirror? Is there another one we could use that may be more reliable?

erikrose:
I had the same problem, but it works under Stretch (which I munged one of the Dockerfiles to use). It's the fault of the old Python on the old distros, which doesn't understand SNI.

https://pypi.tuna.tsinghua.edu.cn used to be my favorite, but it has been unreachable from my location for 2 days. That's how I ended up on doubanio. https://mirrors.ustc.edu.cn/pypi/web/ [is laggier](https://pypi-mirrors.org/), both in getting up to date and in ping from here, but it doesn't require SNI. Maybe that's the better choice.

erikrose:
Note to self: when this is merged, release pipstrap 1.4.

erikrose:
People are actually testing it out in #2516.

erikrose:
Where I left it is that it "should work", but some people were having communication hangs from Python but not with wget. I was going to add some logging to help figure out what was going on, but it fell down the priority heap. I think it would also be reasonable to merge in the proposed fix and get a larger sample from Chinese users.

ohemorange:
@bmw want to review?

bmw:
Since `certbot-auto` desperately needs to be refactored if we're going to keep it, we're looking into getting rid of it in favor of a different packaging solution, and I haven't heard anything about this problem in a while (possibly because the change to pipstrap to use the environment variable `DEFAULT_INDEX_BASE`), I'm closing this for now. We can revisit this in the future if we decide we want to work on it.

I've put this code in the [branch pypi-mirrors](https://github.com/certbot/certbot/tree/pypi-mirrors) and referenced that in #4371.

Issue #4879, opened 2017-06-28
Messages are displayed differently if run on the commandline and CRON (stderr,stdout)


jobst:
When you run the command 
`/PATH/certbot-auto renew --non-interactive --no-bootstrap --no-self-upgrade --post-hook "apachectl graceful"`
it produces two different outputs (see below) depending if you run it from cron or the commandline, this is probably due to a mix-match of STDERR and STDOUT.

I have not had the pleasure of a real renewal to see what happens then as I only just started using certbot.

So it displays the messages "Cert not yet due for renewal" when running as CRON (in file renewal.py) from function "should_renew()" first and then "Processing ..." from function "handle_renewal_request()" second. If running from commandline all is fine, the order is correct.

This is a annoying especially if you have a longer list of domains as "Cert not yet due for renewal" does NOT contain the domain name and you can't figure out what belongs to what.

## My operating system is (include version):
CentOS 6.9

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto, no yum packages for CentOS

## I ran this command and it produced this two different outputs:
/PATH/certbot-auto renew --non-interactive --no-bootstrap --no-self-upgrade --post-hook "apachectl graceful"

## Certbot's behavior differed from what I expected because:
**When run from commandline:**

    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN1.com.conf
    -------------------------------------------------------------------------------
    Cert not yet due for renewal
    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN2.com.conf
    -------------------------------------------------------------------------------
    Cert not yet due for renewal
    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN3.com.conf
    -------------------------------------------------------------------------------
    Cert not yet due for renewal
    

**When run from CRON:**

    Cert not yet due for renewal
    Cert not yet due for renewal
    Cert not yet due for renewal
    
    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN1.com.conf
    -------------------------------------------------------------------------------
    
    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN2.com.conf
    -------------------------------------------------------------------------------
    
    -------------------------------------------------------------------------------
    Processing /etc/letsencrypt/renewal/DOMAIN3.com.conf
[...]

Issue #4741, opened 2017-05-25
Add --install(ation)-only flag to certbot-auto


bmw:
This allows people to separate installing Certbot and running it for the first time. It does the same thing as `certbot-auto --version`, however, this flag makes it explicit this functionality is available and doesn't actually invoke Certbot.

cc @nuxwin

nuxwin:
@bmw 


@bmw 

Really, your involvement here is much appreciated ;)
Keep up your good work ;)

Thank again.

bmw:
As I've described it, this flag would also update Certbot to the latest version. I think there's benefit of having this flag work this way as well (and those who don't want this behavior can use `--no-self-upgrade`) so we should make this clear in the flag name/documentation.

sydneyli:
ill take this on for the PR demo tomorrow!

nuxwin:
@sydneyli Thank youf or merging this ;) That is much appreciated.

PR #5164, opened 2017-10-04
certbot: Flush output after write in IDisplay methods.

r5d:
- Update `notification`, `yesno`, `checklist`, `_print_menu`, and
  `_get_valid_int_ans` methods in `certbot.display.util.FileDisplay`.
- Update `notification` method in
  `certbot.display.util.NoninteractiveDisplay`.

Addresses issue #4879.

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -117,6 +117,7 @@ class FileDisplay(object):
         self.outfile.write(
             "{line}{frame}{line}{msg}{line}{frame}{line}".format(
                 line=os.linesep, frame=side_frame, msg=message))
+        self.outfile.flush()
         if pause:
             if self._can_interact(force_interactive):
                 input_with_timeout("Press Enter to Continue")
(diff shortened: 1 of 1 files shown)


bmw:
LGTM.

Thanks @r5d!

jobst:
Thank you for the inclusion, I am the reporter of the original #4879.

I just had a look how to get the next version ... IMHO the flags are a little confusing - this maybe just me but I like things to be clear leading to less errors and less queries.

I think there should be a flag or an option called "--update" with an explanation that this flag will initiate an update of all required scripts/programs including certbot-auto to the latest version and then stop. Maybe there should even be a section called "Update"?

bmw:
I'm glad we were able to fix this problem for you, however, this patch hasn't been included in a release yet. It will in our 0.20.0 release at the beginning of next month though.

As for updating `certbot-auto`, I'm sorry for the confusion. We have an issue for a `--install-only` flag at #4741 here which as I've described it, would also update you to the latest version and stop if you already have Certbot installed. I'll make a note about this there to make sure it's clear that flag can be used for both purposes.

Issue #5115, opened 2017-09-14
--deploy-hook value on command line not being used during renewal


bmw:
## I ran this command and it produced this output:
```
$ certbot certonly -d example.org --standalone --deploy-hook "touch foo"
...
$ certbot renew --force-renewal --deploy-hook "touch bar"
...
$ grep bar /etc/letsencrypt/renewal/example.org.conf || echo "Why wasn't my deploy hook set?"
Why wasn't my deploy hook set?
$
```

## Certbot's behavior differed from what I expected because:
The deploy hook value should be set. I think this is due to renew hook to deploy hook rename and default detection. I think `certbot.cli.VAR_MODIFIERS` needs to be changed for this interaction to be handled proerply.

PR #5172, opened 2017-10-05
Integration test to prevent regressions of #5115

bmw:
The next two calls to `CheckDirHooks` should fail if we regress on #5115.


bmw:
Whoops. Nice catch. Thanks for knowing what issue I meant when GitHub didn't. :smile:

PR #5173, opened 2017-10-05
Retry boulder-fetch.sh to prevent test failures

bmw:
Prevents issues like the one seen at https://travis-ci.org/certbot/certbot/jobs/282923098.


bmw:
Darn. This function is available to our own scripts. New solution incoming.

Issue #3640, opened 2016-10-15
[nginx] IPv6 support?


emilstahl:
Currently the nginx plugin adds:
`listen 443 ssl; # managed by Certbot`

It should add the following for IPv6 support:

```
listen 443 ssl; # managed by Certbot
listen [::]:443 ssl; # managed by Certbot
```

~An option for HTTP 2 would be nice too, either by specifying `--http2` flag or just check nginx for compatibility.;~

```
listen 443 ssl http2;
listen [::]:443 ssl http2;
```

bmw:
@emilstahl, thanks for making this issue. I created #3646 so we can trick the issues of IPv6 and HTTP2 separately.

kirelagin:
Actually, it is not entirely safe. In older versions of nginx the `ipv6only` was `off` by default, so having these two lines you propose at the same time would break the config.

I’d say that `ipv6only=on` should be added to the `listen [::]:443` line, but, AFAIU, this option is Linux-only and I’m not sure how nginx will respond to it on other OSes.

jkldgoefgkljefogeg:
`ipv6only=on` is the default.   https://nginx.org/en/docs/http/ngx_http_core_module.html#listen
For a server config with `listen [::]:80;`, certbox does not generate `listen [::]:443;` At least user should be warned IPv6 SSL was not enabled.

jplatte:
@bmw So what's the current state of this? You closed this issue in #5178, but then seem to have reverted your changes. Should this issue be re-opened?

ohemorange:
@jplatte: We then unreverted the changes. The latest version of Certbot supports IPv6; see some caveats that we plan to fix in https://github.com/certbot/certbot/issues/5438.

PR #5178, opened 2017-10-08
Nginx IPv6 support

joohoi:
Adds support for IPv6 to Nginx plugin.
This PR adds IPv6 address handling in similar way that IPv4 is currently handled (wildcard listen directives if needed etc).

IPv6 listens are written to server block and the temporary challenge vhost only if the selected vhost in question is configured to use IPv6 in the first place.

ipv6only=on is added if it doesn't exist in any server block beforehand. It can be defined only once per addr:port pair in the configuration. This is needed to avoid socket bind errors caused by IPv6 wildcard `[::]:443` opening both IPv4 and IPv6 sockets and conflicting with already opened IPv4 socket.

Fixes: #3640


bmw:

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -443,15 +470,29 @@ def _make_server_ssl(self, vhost):
...
+        ipv6_block = ['']
+        if vhost.ipv6_enabled():
+            ipv6_block = ['\n    ',
+                            'listen',
+                            ' ',
+                            '[::]:{0} ssl'.format(self.config.tls_sni_01_port)]
+            if not ipv6info[1]:
+                # ipv6only=on is absent in global config
+                ipv6_block.append(' ')
+                ipv6_block.append('ipv6only=on')
```
ohemorange: Is there any reason not to always append `ipv6only=on`? It will sometimes be unnecessary, but it seems more robust to always explicitly include it instead of relying on existing things in the server block.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -443,15 +470,29 @@ def _make_server_ssl(self, vhost):
...
+
         snakeoil_cert, snakeoil_key = self._get_snakeoil_paths()
 
         ssl_block = ([
+            ipv6_block,
             ['\n    ', 'listen', ' ', '{0} ssl'.format(self.config.tls_sni_01_port)],
```
ohemorange: Actually, why add it as a separate directive at all, instead of just a single `listen [::]:443 ssl ipv6only=off` if appropriate and `listen 443 ssl` otherwise? Does that interact poorly with FreeBSD networking or something?

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -73,8 +88,10 @@ def fromstring(cls, str_addr):
                 ssl = True
             elif nextpart == 'default_server':
                 default = True
+            elif nextpart == "ipv6only=on":
```
ohemorange: What do we want this variable to represent? "Was ipv6only=on explicitly set" is what it currently is. That's not the same as "is ipv6only on for this address", because the default varies by Nginx version.

certbot-nginx/certbot_nginx/tls_sni_01.py
```diff
@@ -51,6 +51,9 @@ def perform(self):
         default_addr = "{0} ssl".format(
             self.configurator.config.tls_sni_01_port)
 
+        ipv6info = self.configurator.ipv6_info(
```
ohemorange: can you name the resulting booleans for clarity?

certbot-nginx/certbot_nginx/tls_sni_01.py
```diff
@@ -63,7 +66,17 @@ def perform(self):
             if vhost.addrs:
                 addresses.append(list(vhost.addrs))
             else:
-                addresses.append([obj.Addr.fromstring(default_addr)])
+                if ipv6info[0]:
```
ohemorange: Actually, what happens if we listen on both IPv6 and IPv4 even if IPv6 isn't enabled in Nginx or for the site? Does it crash? Or does it still work? Why only listen on IPv6 if the config is already doing so, for the default?

certbot-nginx/certbot_nginx/configurator.py
joohoi: ipv6only=on can unfortunately exist only once per IP:PORT pair, duplicate entries result into nginx not being able to start.

certbot-nginx/certbot_nginx/configurator.py
joohoi: Yeah, dual stack socket is a Linux only thing, unfortunately.

certbot-nginx/certbot_nginx/obj.py
joohoi: ipv6only=on can exist only once throughout Nginx configuration, having it explicitly set on two different server blocks would cause Nginx to error out.

certbot-nginx/certbot_nginx/configurator.py
ohemorange: awesome, classic

certbot-nginx/certbot_nginx/tls_sni_01.py
joohoi: Yeah, improves readability quite a bit, done.

certbot-nginx/certbot_nginx/tls_sni_01.py
joohoi: Nginx will not start if listen [::] is defined and it's built without IPv6 support. However we're only listening IPv6 if the originating server block had IPv6 configuration (so we would have caught the problem already at the configtest). We need to add it here for port 443 too to be able to answer the TLS-SNI challenge.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -257,6 +257,31 @@ def choose_vhost(self, target_name, raise_if_no_match=True):
 
         return vhost
 
+
+    def ipv6_info(self, port):
```
ohemorange: would you mind moving this to `parser.py`? since it's used both here and in `tls_sni_01`.

certbot-nginx/certbot_nginx/tls_sni_01.py
```diff
@@ -51,14 +51,28 @@ def perform(self):
...
+                    ipv6_addr = "[::]:{0} ssl".format(
+                        self.configurator.config.tls_sni_01_port)
+                    if not ipv6only:
+                        # If ipv6only=on is not already present in the config
+                        ipv6_addr = ipv6_addr + " ipv6only=on"
+                    addresses.append([obj.Addr.fromstring(default_addr),
+                                      obj.Addr.fromstring(ipv6_addr)])
+                else:
+                    addresses.append([obj.Addr.fromstring(default_addr)])
+                    logger.info("Using default address %s for TLSSNI01 authentication.",
```
ohemorange: We should also log this when doing ipv6

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -484,15 +512,29 @@ def _make_server_ssl(self, vhost):
...
+            if not ipv6info[1]:
+                # ipv6only=on is absent in global config
+                ipv6_block.append(' ')
+                ipv6_block.append('ipv6only=on')
+
         snakeoil_cert, snakeoil_key = self._get_snakeoil_paths()
 
         ssl_block = ([
+            ipv6_block,
             ['\n    ', 'listen', ' ', '{0} ssl'.format(self.config.tls_sni_01_port)],
```
ohemorange: Hm, we should probably not be adding `listen 443` if there is an ipv6 80 listen but no ipv4 80 listen in the server block.

certbot-nginx/certbot_nginx/configurator.py
joohoi: Wouldn't we be calling `parser` through `configurator` anyway from `tls_sni_01`?

certbot-nginx/certbot_nginx/configurator.py
joohoi: That's right, fixed!

certbot-nginx/certbot_nginx/tls_sni_01.py
joohoi: Fixed

PR #5182, opened 2017-10-10
Update brew before installing packages

bmw:
This hopefully fixes the problem seen [here](https://travis-ci.org/certbot/certbot/jobs/285601018).


bmw:
Unless something else in the Travis environment suddenly changes, master is broken until this (or another PR like it) lands.

PR #5183, opened 2017-10-11
Skip menu in webroot plugin when there's nothing to choose from

joohoi:
Do not present user a webroot map menu when the only thing to select is "Enter a new webroot". This might confuse the user to try and input webroot path here, which then again will result into somewhat uninformative error message: `** Invalid input **`. 

I stumbled upon this while helping on IRC, and figured it'd be a trivial fix.

Old behaviour:
```
$ certbot certonly --webroot -d example.com
Plugins selected: Authenticator webroot, Installer None
Obtaining a new certificate
http-01 challenge for example.com

Select the webroot for example.com:
1: Enter a new webroot
Press 1 [enter] to confirm the selection (press 'c' to cancel): 1
Input the webroot for example.com: (Enter 'c' to cancel): /path/to/webroot
...
```

New behaviour:
```
$ certbot certonly --staging --webroot -d example.com
Plugins selected: Authenticator webroot, Installer None
Obtaining a new certificate
http-01 challenge for example.com
Input the webroot for example.com: (Enter 'c' to cancel): /path/to/webroot
...
```

### DIFF
--- a/certbot/plugins/webroot.py
+++ b/certbot/plugins/webroot.py
@@ -102,10 +102,14 @@ to serve all files under specified web root ({0})."""
         webroot = None
 
         while webroot is None:
-            webroot = self._prompt_with_webroot_list(domain, known_webroots)
-
-            if webroot is None:
-                webroot = self._prompt_for_new_webroot(domain)
+            if known_webroots:
+                # Only show the menu if we have options for it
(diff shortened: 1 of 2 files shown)


joohoi:
Added the missing test case. And thanks for the issue link, I wasn't aware we had one for this.

PR #5203, opened 2017-10-23
Use josepy instead of acme.jose.

jezdez:
Ping @bmw?


jezdez:
@bmw One item on my todo list left is figure out what we do with josepy, my recommendation is to move it to the certbot or letsencrypt GitHub organization to simplify collaboration and maintain it close to the rest of certbot. I'm happy to transfer both the PyPI and ReadTheDocs project to whoever you think would be best.

bmw:
Thanks for doing this work @jezdez! I'm sorry this has sat here so long without comments. Most of the Certbot team was unavailable working on other stuff last week and I'm just getting caught up.

I've pinged our major distro package maintainers to make sure that us doing this isn't going to cause any problems on their end. (If you're curious, I'm mainly worried about distros that are in some ways frozen and the need to add this new package causing problems for updates to our packages we'd otherwise get.) Assuming it won't be an issue, we can transfer the josepy repo to the Certbot org, transfer PyPI and ReadTheDocs to me, and I'll review and merge this PR.

I'll report back when I've heard from everyone which should happen in the next day or so. Thanks again!

jezdez:
@bmw No problem at all, I knew you'll get there eventually. Thanks for taking a look!

As with regard to the maintainers, in some cases it has been useful to just ship an external Python dependency as part of the certbot package -- basically vendoring -- for those distros that can't add new dependencies after the fact. But I realize that it may not be compatible with every distro's policy.

bmw:
And it's been another month...

I'm sorry this is taking so long. While I was able to get in touch with most of our package maintainers who were in favor of this change, our main Ubuntu contact has been travelling and we've had a hard time getting a hold of him. He recently reached out to me and I hope to talk to him about this in the next few days.

I've been waiting to talk to all of our major maintainers as I'd like to avoid distro specific changes to our packaging here, but if I can't get an answer from Ubuntu, I'll move forward with this change and deal with problems there as they come up.

I just wanted to give you a very delayed update on what's going on here and I'll report back no later than next week.

jezdez:
@bmw Thank you for the feedback and there is no need for apologies about the review timeline. I knew this would take a while given the amount of 3rd parties distributing certbot and appreciate the speed nevertheless.

I've transferred the [PyPI project](https://pypi.python.org/pypi/josepy) to @bmw, the [repo into the certbot GitHub org](https://github.com/certbot/josepy) (add assigned the [EFF devs](https://github.com/orgs/certbot/teams/eff-devs/members) team to it) and added @bmw to the [Readthedocs project](https://readthedocs.org/projects/josepy/) as a maintainer. For RTD I wasn't able to remove myself as maintainer like I was on PyPI so you'll need to do that on your own.

Please let me know if there are questions about the repo layout and test setup etc. I won't be out of the world as well, don't hesitate to ping me.

bmw:
Awesome. Thanks again @jezdez!

angsuman:
I made the following changes in /usr/local/lib/python3.4/dist-packages/free_tls_certificates/client.py to fix the issue locally:
def request_certificate_issuance(client, challgs, csr, logger):
    import josepy
    # Convert the OpenSSL.crypto.X509Req to a ComparableX509 expected by request_issuance.
    csr = josepy.util.ComparableX509(csr)
------------

def load_or_generate_private_key(keyfile, log):
    import josepy
------------------
key = josepy.JWKRSA(key=josepy.ComparableRSAKey(key))
================================
Tip: Search for jose

PR #5221, opened 2017-11-05
Feature/azure dns

askvictor:
Implemented DNS Auth for Azure DNS. Based on Google DNS.

Apologies for lack of tests - I'm really out of my depth writing them, and not too certain about how useful they can be without a real cloud env to test them in. I've tested it in my own Azure tenancy and it works, though I haven't tested anything with multiple zones as I haven't got multiple zones.


joohoi:
Thank you for the PR, and sorry for the long wait. It's exiting to get Azure DNS support for Certbot as well!

For the tests, it'd be useful to mock responses from azure.mgmt.dns library to catch corner case bugs and to prevent regression further on in the development. For that, you could follow the same scheme that's used for example for Cloudflare DNS plugin: mocking the client library, and setting the expected return values to test Certbot side of the operations. Here's an example from certbot-dns-cloudflare plugin how to achieve that: https://github.com/certbot/certbot/blob/master/certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare_test.py#L53-L81

Using the mocked Azure client you will be able to add tests for things like error cases (TODO in _find_managed_zone()) and so forth.

On a quick read-through, the code looks clean and good! Let me know if I can help with the tests in any way.

jgillula:
Hi @askvictor! Since we haven't seen any activity on this PR for a while, I just wanted to check and see if you still planned to work on it, or if you'd be OK with someone else picking up the work.

askvictor:
Hi; sorry, I'm no longer working on this as I've changed jobs and I no longer need it :( More than happy for someone else to take over (I though I had already said as much, but it might have been in the issue rather than the PR). Good luck!

Issue #4736, opened 2017-05-25
Add + fix up missing API docs


ynasser:
If you look at the sidebar on https://certbot.eff.org/docs/api.html, you'll see the modules which have documentation available (e.g. `certbot.crypto_util`). Several of our modules don't appear (e.g. `certbot.cert_manager`), as they don't have entries in `docs/`.

There are two main parts to this issue:
1. Find modules that are missing from `docs/` and add them in. See the existing examples to figure out how to do this.
2. Review the Sphinx-style documentation in the modules you are going to add (or ones that are already included), and add any extra information if you can. Sometimes we didn't have time to include anything beyond a short comment explaining what the function does, but other kinds of documentation can be added (e.g. parameter types, return types, errors raised, etc ... )

See here for more info on how to work with our documentation: https://certbot.eff.org/docs/contributing.html#updating-the-documentation

This task would be a great way for a first-time contributor to be introduced to certbot's codebase!

ynasser:
This is kind of a duplicate of #3224, but there's a lot more volunteer-friendly info here.

jonasbn:
Hello, 

I would like to volunteer my assistance. I looked at the [Coding with EFF page](https://www.eff.org/about/opportunities/volunteer/coding-with-eff), but I am not sure it points to the right Google group. I have signed up. The toolchain is somewhat new to me, but I would like to give it a shot and perhaps learn something new in the proces. 

Have a look at my profile and let me know what you think.

ynasser:
Hi @jonasbn! We'd love to have your help :)

We don't really use the Google Group you signed up for (it's mostly announce-only), and most discussion happens in GitHub issues (here) or the #letsencrypt-dev channel on Freenode (irc). Also, we have an open source coordinator who can jump on a phone/video call with you and explain how to get setup working on this, and support you while you're working on it: @SwartzCr.

Feel free to ask us questions on #letsencrypt-dev!

jonasbn:
Hi @ynasser 

Thanks for the warm welcome.

I will read up on the documentation on contribution etc.

Then I will see if I can get the proces of adding the lacking documentation going and hopefully some PR will come out of it for your review.

jonasbn:
Just a quick update - and sorry about the silence.

I got the toolchain (Sphinx) installed and running. Got the documentation generated locally and will try to navigate the file structure to get at sense of the concept and my head around the structure. 

[...]

PR #4728, opened 2017-05-24
[WIP] Alternative whitelisting approach to mypy

toolness:
This is an alternative approach to #4244 that attempts to incrementally add mypy typings over time by only running mypy on files specified in a whitelist, as described in https://github.com/certbot/certbot/issues/4244#issuecomment-303418574.

It also adds mypy to tox's default `testenv`, so that it's run on travis.

To do:
- [ ] Document how to add new files to the whitelist.
- [ ] Determine if the settings in `mypy.ini` ware acceptable.

PR #5228, opened 2017-11-07
Documentation update to certbot/main.py

jonasbn:
In relatation to issue #4736 this is the first shot at addition of missing documentation to a single file. There are a few things I have not yet understood about certbot and some Python aswell, so the work is not complete and the PR require a review.

I need to know if I am pulling in the right or wrong / direction and I have asked some questions on issue #4736. I might have addressed some of these based on assumptions. Please provide some feedback on this PR and let me know if you want me to continue the documentation work - and if so I just need a push in the right direction.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -716,7 +716,7 @@ def _install_cert(config, le_client, domains, lineage=None):
     :param plugins: list of domains
     :type plugins: `list` of `str`
 
-    :param lineage: certificate lineage object
+    :param lineage: certificate lineage object. Defaults to `None`
     :type lineage: storage.RenewableCert
 
     :returns: `None`
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -84,8 +84,11 @@ def _get_and_save_cert(le_client, config, domains=None, certname=None, lineage=N
     checks, and requests for user input.
 
     :param interfaces.IConfig config: Configuration object
-    :param list domains: domains to get a certificate. This argument is optional, if not supplied it will default to `None`
-    :param str certname: Name of new cert. This argument is optional, if not supplied it will default to `None`
+
+    :param list domains: domains to get a certificate. Defaults to `None`
+
+    :param str certname: Name of new cert. Defaults to `None`
+
     :param storage.RenewableCert lineage:
 
     :returns: the issued certificate or `None` if doing a dry run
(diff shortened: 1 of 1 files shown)


jonasbn:
Questions addressed, peer review most welcome

SwartzCr:
I think this looks good, but want to wait for @bmw to look at it to give a final okay.
The one thing I'd point out is that it seems like in some of the previous doc strings we have things of the form:
`:param str cert_path: path to cert`
rather than of the form
```
:param cert_path: path to cert
:type cert_path: str
```
I have a mild preference that these be uniform. Additionally it looks like some of the previous doc strings could use expanding, it might be worth it to look at the ones you didn't touch and make sure they look as thorough as the ones you wrote

jonasbn:
Hi @SwartzCr 

Thanks for the feedback, much of the existing documentation is written on the short form

`:param str cert_path: path to cert`

I do however prefer the long form

```
:param cert_path: path to cert
:type cert_path: str
```

Makes it easier to scan using `ack`, `grep` or similar and with a shorter definition, et leaves more room for the descriptive string - the above example does not demonstrate this, but for classes etc it is more clear, but it is your call - I can change all definitions to the short form.

SwartzCr:
Hmm - I wonder if @bmw has a preference
Also, sorry for having this sit in limbo for so long - hopefully we can get it merged soon, but feel free to pick up another issue while we're waiting to get this one merged

jonasbn:
Hi @SwartzCr 

I have [other projects](https://github.com/jonasbn?tab=repositories) to work on so I am in no hurry. I am doing a PR per file as requested by @ynasser in #4736 and I almost have the next PR branch completed, but I will not make the PR until this one is addressed.

SwartzCr:
Looks good to me! Thanks for doing all this work.
I think it'd be awesome if along with this you made type annotations that mypy could use for type checking

jonasbn:
Hi @SwartzCr 

Thanks I will have a look at `mypy` annotations, please inform me of any relevant resources, since I am very much a Python _n00b_

SwartzCr:
Sure!
Here's a good blog post by the zulip team introducing the concept: http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/
The official project website: http://mypy-lang.org/
And here's the official documentation: https://mypy.readthedocs.io/en/latest/

You can also look at the parent issue for this #4244 as well as the PRs that @toolness worked on (#4728 and #4707)
Also it's possible that either @toolness or @dokazaki would be interested in helping you get started with it since it seems that they were both pretty involved in the earlier mypy thread.
That should be a good start, but let me know if you want me to find you other resources!

PR #5230, opened 2017-11-08
Install dnsmadeeasy extras from dns-lexicon

bmw:
Fixes https://travis-ci.org/certbot/certbot/jobs/298857322#L6606.

Now with the possible exception of a couple core Python packaging libraries whose installation is handled by [pipstrap](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/pipstrap.py), all of our Python dependencies should be pinned in the dev environment and during tests.

This also specifes the dnsmadeeasy extras as requirements for our dnsmadeeasy plugin.

### DIFF
--- a/certbot-dns-dnsmadeeasy/setup.py
+++ b/certbot-dns-dnsmadeeasy/setup.py
@@ -10,7 +10,9 @@ version = '0.20.0.dev0'
 install_requires = [
     'acme=={0}'.format(version),
     'certbot=={0}'.format(version),
-    'dns-lexicon',
+    # new versions of lexicon require that we install dnsmadeeasy extras and
+    # 2.1.11 is the first version that defines them.
+    'dns-lexicon[dnsmadeeasy]>=2.1.11',
     'mock',
     # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
     # will tolerate; see #2599:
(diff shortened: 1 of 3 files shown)


ohemorange:
IRC conversation summary: doing this means we extra definitely should be running package breakage tests. Eventually, we should automate them, but for now let's do it during/ right after monthly triage.

AnalogJ:
woah, didn't realize that you guys were using `lexicon` as a dependency. 
In general I try to make my changes backwards compatible, but I recently changed build/CI systems and I was running into some really nasty python locale issues.

Long story short, I added a dependency on `babel` for the `dnsmadeeasy` provider, and bumped up the patch version since it only effected that single provider. 

Hopefully it wasn't too much work to fix on your side. I'm always happy to find `lexicon` usage in the wild. If theres any other ways I can help, don't hesitate to get in touch 😃

bmw:
Thanks for the update @AnalogJ. Certbot uses `lexicon` for 5 of its DNS plugins and we'll be pushing for more Linux distros to package these plugins (and therefore `lexicon`) in the next couple months. `lexicon` made supporting a number of these DNS providers quite easy :smile:

As far as I know, the change didn't negatively affect any of our users and it pushed us to make some changes to our development process we've wanted to do for a while anyway.

PR #5258, opened 2017-11-23
ACMEv2: Allow "uri" or "url" in challenge

jsha:
In ACMEv2, challenges have "url" instead of "uri". To handle this smoothly, Challenge's uri field becomes private (_uri), and is joined by _url. Serialization and deserialization will preserve whichever one was set. The `uri` name is taken over by an `@property` that returns whichever of the two is set. I chose not to enforce that they shouldn't both be present because it would just add unnecessary code and brittleness with no stability benefit.

I had to update a bunch of tests because they relied on manually setting the "uri" field of Challenge during construction. This is a minor API break, but I strongly suspect that most users of Challenge only initialize through JSON parsing, and only use it through Client.answer_challenge.

### DIFF
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -368,7 +368,7 @@ class ChallengeResource(Resource):
     def uri(self):  # pylint: disable=missing-docstring,no-self-argument
         # bug? 'method already defined line None'
         # pylint: disable=function-redefined
-        return self.body.uri  # pylint: disable=no-member
+        return self.body._url or self.body._uri  # pylint: disable=no-member
 
 
 class Authorization(ResourceBody):
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -348,7 +348,7 @@ class ChallengeBody(ResourceBody):
         return jobj_fields
 
     @property
-    def uri(self):
+    def uri(self):  # pylint: disable=missing-docstring
         return self._url or self._uri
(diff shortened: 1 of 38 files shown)


acme/acme/messages.py
```diff
@@ -360,8 +368,8 @@ class ChallengeResource(Resource):
     @property
     def uri(self):  # pylint: disable=missing-docstring,no-self-argument
         # bug? 'method already defined line None'
-        # pylint: disable=function-redefined
-        return self.body.uri  # pylint: disable=no-member
+        # pylint: disable=function-redefined,protected-access
+        return self.body._url or self.body._uri  # pylint: disable=no-member
```
bmw: Can't we just use `self.body.uri` here because of the property in `ChallengeBody` above?

jsha:
Ah, the __init__ trick was a nice suggestion, cleaned this up a lot. There was one test diff I had to leave in because it used `.update(...)` which presumably doesn't call init.

bmw:
Since tests are going to fail until #5291 is merged, I'm cancelling the Travis build but will kick it when they're expected to pass again.

bmw:
I opened #5300 which contains the work I mentioned today which allows the user to use uri/url in all public methods to hide this small implementation detail. Let me know what you think!

bmw:
These changes were merged in #5300.

Issue #5045, opened 2017-08-23
Show expiration and renew dates in certbot renew output


jkufner:
When running `certbot renew` it shows following info (paths truncated):

`````
The following certs are not due for renewal yet:
  /etc/letsencrypt/live/.../fullchain.pem (skipped)
  /etc/letsencrypt/live/.../fullchain.pem (skipped)
  /etc/letsencrypt/live/.../fullchain.pem (skipped)
No renewals were attempted.
`````

It would be muchmore useful if it said when the certificates are going to expire and when they are due to renewal. For example:

`````
The following certs are not due for renewal yet:
  /etc/letsencrypt/live/.../fullchain.pem (skipped; exp. 2017-12-01, renew 2017-11-01)
  /etc/letsencrypt/live/.../fullchain.pem (skipped; exp. 2017-12-05, renew 2017-11-05)
  /etc/letsencrypt/live/.../fullchain.pem (skipped; exp. 2017-11-01, renew 2017-10-01)
No renewals were attempted.
`````

SwartzCr:
Interesting!
That would be useful, let me add it to the wishlist milestone, but mark it as a bite sized task that someone could work on. Maybe @ynasser would want to tackle this when she does her livestream?

ynasser:
Yes, this would be a great issue to implement in the livestream. Thanks @SwartzCr!

ppKrauss:
Hi,  #5065 is perhaps a variant... But about user needs, is not only dates, but a clue about "What Do I Do Now??" 

`certbot`  stop to work (!) and no clue about how to workaround or redo something.

ynasser:
@ppKrauss When I implement this, I'll try to make it more clear in the user interface how users should use the dates we provide them with, that way they'll understand how to use the data we're giving them. So, instead of just saying `Cert not yet due for renewal`, it'll say something like, `Cert not yet due for renewal yet. See recommended renewal dates listed above. For documentation, see: https://certbot.eff.org/docs/using.html#renewing-certificates`.

jkufner:
@ynasser That is awful lot of words. Keep it short.

SwartzCr:
@jkufner do you have a recommendation for a potential wording?

SwartzCr:
@jkufner do you have a proposed wording?

jkufner:
@SwartzCr See the first comment ;-)

SwartzCr:
[...]

PR #5259, opened 2017-11-23
Show expiration dates for cert when skipping its renewal

SwartzCr:
a reasonable fix for #5045 
I don't think it's necessary to include the length of the renewal window to the output, since users shouldn't get concerned about lack of renewal until the cert is within days of expiration, and maybe not even then. Additionally listing the renewal window length would require considerable parsing.
This will at least give the user a sense of why a certificate was not renewed.
Note that this has not been tested by anything but unit tests.

### DIFF
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -425,7 +425,10 @@ def handle_renewal_request(config):
                     main.renew_cert(lineage_config, plugins, renewal_candidate)
                     renew_successes.append(renewal_candidate.fullchain)
                 else:
-                    renew_skipped.append(renewal_candidate.fullchain)
+                    expiry = crypto_util.notAfter(renewal_candidate.version(
+                        "cert", renewal_candidate.latest_common_version()))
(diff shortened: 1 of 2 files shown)


bmw:
Looks like a lot of tests are failing here. Let me know if you'd like me to take a look.

bmw:
Looks like integration tests are failing. I think you probably want to use `renewal_candidate` which is a `certbot.storage.RenewableCert` object instead of `lineage_config` which is a `certbot.configuration.NamespaceConfig`

sydneyli:
Functionality looks good to me. Adding a unit test for cert-skipping, since there isn't one yet.

bmw:
Sorry. Free to work on this and let me know when it's ready for review. Bonus points for squashing all of your commits into one for a slightly cleaner git history since we want to merge rather than squash this PR.

sydneyli:
Just squashed my own commits-- ready now.

PR #5270, opened 2017-11-28
Remove all but one BOULDER_INTEGRATION, and macOS

jsha:
These tests are retained in the test-everything branch, which has a Travis cron
job to run nightly.

Removing these speeds up the Certbot Travis builds dramatically for two reasons:
 - The Boulder integration tests are slow (10-12 minutes), and it's exceedingly
   rare for them to fail on one Python environment but not another.
 - The macOS tests take a very long time to run, because they need to wait for
   build slots on the limited number of macOS instances, which are often in high
   demand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ before_install:
 before_script:
   - 'if [ $TRAVIS_OS_NAME = osx ] ; then ulimit -n 1024 ; fi'
 
-# using separate envs with different TOXENVs creates 4x1 Travis build
-# matrix, which allows us to clearly distinguish which component under
-# test has failed
 matrix:
   include:
     - python: "2.7"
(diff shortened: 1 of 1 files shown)


bmw:
It looks like there are some issues with the test-everything branch when it tries to bring in master. See https://travis-ci.org/certbot/certbot/builds/309100097. I'd like to make sure that's up and working before we merge this PR.

jsha:
The issue was that we also changed .travis.yml on master, so when this branch tried to pull master, it would have to merge. By default, git merge pops open an editor; in this case, that hung forever.

I was able to fix by changing to:

```
git pull origin master --strategy=recursive --strategy-option=theirs --no-edit
```

However, we may want to go another path, and explicitly fail if there is ever any merge conflict. The theory being that the only file that should differ in the test-everything branch is .travis.yml, and if .travis.yml on master changes, we want to make sure we merge those changes into test-everything so the branches stay in sync with regards to what tests are run. What do you think?

jsha:
Done. I snuck in an additional change to comment on cover in the build matrix:

```
TOXENV=cover FYI="this also tests py27"
```

Lemme know if you want me to pull that out. :-)

bmw:
:smile:

I'm fine with adding something to convey that information, but is it possible to do it with a comment rather than an environment variable? It doesn't really matter but a comment seems cleaner. If Travis' syntax doesn't allow to put a comment near/on the that line, the environment variable is fine.

bmw:
Per offline comments, the environment variable causes it to show up on the Travis build page which is a good enough reason to keep it that way for me. Approved!

jsha:
FYI the plan is to merge this, then rebuild the `test-everything` branch as master + a commit that adds back these tests. That way test-everything will be ahead of master, instead of behind master, which will make future merges work correctly.

Once that's done, we'll make test-everything a protected branch and require reviews for merging changes to it.

PR #5271, opened 2017-11-28
Less verbose output from install_and_test.sh

jsha:
This removes the -v (verbose) flag from nosetests, which printed a line for every test case. Now it prints a dot for every test case instead. Being less verbose reduces noise in the output and makes it easier to see failures and warnings.

Also, add the -x flag to install_and_test.sh itself, so people running tox can more easily see which commands are being run. In particular, seeing nosetests being run may help people realize that they can run nosetests directly for a tighter test loop.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,2 @@
 [report]
-# show lines missing coverage in output
-show_missing = True
+omit = */setup.py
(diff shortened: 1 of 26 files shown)


jsha:
This has been overtaken by events; we now use pytests, and the set -x change is over in #5288.

Issue #4825, opened 2017-06-11
Travis may not report failure if a test fails


zjs:
Tests in `certbot-dns-digitalocean` appeared to be passing on Python 3 because Travis was reporting those runs as green.

However, until #4821 one test contained calls to a method that does not exist in Python 3.

These failures were being hidden by nose errors which prevented all tests from being run (without reporting failure?!).

Example: https://travis-ci.org/certbot/certbot/jobs/241386053

```
Obtaining file:///home/travis/build/certbot/certbot/certbot-dns-digitalocean
Requirement already satisfied: zope.interface==4.1.3 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 28))
Requirement already satisfied: mock==2.0.0 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 29))
Requirement already satisfied: acme==0.16.0.dev0 in ./acme (from certbot-dns-digitalocean==0.16.0.dev0)
Requirement already satisfied: certbot==0.16.0.dev0 in /home/travis/build/certbot/certbot (from certbot-dns-digitalocean==0.16.0.dev0)
Collecting python-digitalocean>=1.11 (from certbot-dns-digitalocean==0.16.0.dev0)
Requirement already satisfied: setuptools>=1.0 in ./.tox/py36/lib/python3.6/site-packages (from certbot-dns-digitalocean==0.16.0.dev0)
Requirement already satisfied: six==1.10.0 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 23))
Requirement already satisfied: pbr==1.8.1 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 16))
Requirement already satisfied: cryptography==1.8.2 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 7))
Requirement already satisfied: pyOpenSSL==16.2.0 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 17))
Requirement already satisfied: pyRFC3339==1.0 in ./.tox/py36/lib/python3.6/site-packages (from -c /tmp/tmp.nOoN6GPb3H (line 19))
[...]

PR #5282, opened 2017-11-29
Switch from nose to pytest

### LINKED ISSUE #4377: Bug running unit tests in parallel on Python 2.6
I filed this at nose-devs/nose#1041 but the project seems pretty inactive. We should document this behavior of our tests, fix the problem, or stop running tests in parallel on Python 2.6.

bmw:
Fixes #4377 and #4825. `nose` is no longer maintained so let's switch to a framework that is.

A couple things to note here:

* `_multiprocess_can_split_` lines were for `nose`'s multiprocessing plugin as documented [here](https://nose.readthedocs.io/en/latest/doc_tests/test_multiprocess/multiprocess.html#controlling-distribution) and have been removed.
* Our coverage changed a bit due to `nose` improperly reporting the coverage percentage and `pytest` reporting coverage to the nearest hundreth of a percent rather than the nearest whole percentage point. `route53` was particularly affected because `nose` wasn't reporting coverage on an entire file.

### DIFF
--- a/letsencrypt-auto-source/Dockerfile.wheezy
+++ b/letsencrypt-auto-source/Dockerfile.wheezy
@@ -6,11 +6,11 @@ FROM debian:wheezy
 # Add an unprivileged user:
 RUN useradd --create-home --home-dir /home/lea --shell /bin/bash --groups sudo --uid 1000 lea
 
-# Install pip, sudo, openssl, and pytest:
+# Install pip, sudo, and openssl:
 RUN apt-get update && \
     apt-get -q -y install python-pip sudo openssl && \
     apt-get clean
-# Use pipstrap to install a stable and tested version of pip
+# Use pipstrap to update to a stable and tested version of pip
 COPY ./pieces/pipstrap.py /opt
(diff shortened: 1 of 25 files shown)


letsencrypt-auto-source/Dockerfile.wheezy
```diff
@@ -6,13 +6,15 @@ FROM debian:wheezy
 # Add an unprivileged user:
 RUN useradd --create-home --home-dir /home/lea --shell /bin/bash --groups sudo --uid 1000 lea
 
-# Install pip, sudo, openssl, and nose:
+# Install pip, sudo, openssl, and pytest:
```
jsha: The lines immediately below this don't install pytest, so just remove pytest from this line. Same edit for anything starting with "Install pip, " in this change.

bmw: I made the `pytest` changes.

I think the "Install pip" comments are accurate as we do install `pip` from the package manager, but then update it through `pipstrap` so I changed comments to clarify this.

Issue #5278, opened 2017-11-29
Deprecate Python 3.3


bmw:
This requires a little research into best practices here, but we should start reporting that our ACME module is dropping Python 3.3 support using best practices for a library and we should start logging at the warning level in Certbot about the same problem.

Python 3.3 is no longer supported by CPython and there are no (major) distros using it. I want this for our next release so we can remove support in the release after that.

cc @joohoi and @SwartzCr another good target

hugovk:
Have a look at the practicalities section of http://www.python3statement.org/practicalities/ for some ideas which can be applied specifically for 2.6 and 3.3.

SwartzCr:
Thanks for the comment & link! Please let us know if you think there's anything else we should be doing!
(p.s. I really liked your NaNoGenMo Meowby Dick!)

bmw:
Fixed in #5283.

Issue #5275, opened 2017-11-29
Update tests/letstest/targets.yaml to more relevant AMIs


bmw:
This is something I'd like to sooner rather than later as many of these tests are now irrelevant as they are testing old, sometimes EOL'd distros, and not testing more modern and popular ones. The lack of more modern tests has caused us to ship bugs.

@joohoi or especially @SwartzCr, this would be something useful for you to take off my plate.

hugovk:
https://github.com/certbot/certbot/blob/master/tests/letstest/targets.yaml

Current list:

* ubuntu15.10
* ubuntu15.04LTS
* ubuntu14.04LTS
* ubuntu14.04LTS_32bit
* ~ubuntu12.04LTS~
* debian8.1
* debian7.8.aws.1
* debian7.8.aws.1_32bit
* amazonlinux-2015.09.1
* amazonlinux-2015.03.1
* RHEL7
* ~fedora23~
* centos7
* centos6

Crossed-out ones are EOL.

https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions
https://en.wikipedia.org/wiki/Debian_version_history#Release_table
https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Product_life_cycle
https://en.wikipedia.org/wiki/Fedora_(operating_system)#Releases

Testing on LTS, where available, would be sensible.

Here are the non-EOL from the handy [Support Life Cycles for Enterprise Linux Distributions](https://linuxlifecycle.com/) not already listed:

* RHEL6
* RHEL5
* Ubuntu 16.04 LTS
* Debian 9
* SUSE 12
* SUSE 11

Perhaps add these, ditch the EOL and remove Ubuntu 15.10 (because 16.04 LTS is added)?

I didn't find support dates for Amazon Linux, but 2017.09.1 is the latest, released on 2017-11-22. Perhaps one of amazonlinux-2015.09.1 and amazonlinux-2015.03.1 could be replaced with this.

bmw:
Awesome. Thanks @hugovk.

Despite it being used in our tests, I'm not very familiar with AWS. If you are, what is the easiest way to get the AMI values for the distros you mentioned above so we can update our configuration file?

hugovk:
[...]

PR #5283, opened 2017-11-30
print warnings for 3.3 users

SwartzCr:
responds to #5278

### DIFF
--- a/acme/acme/__init__.py
+++ b/acme/acme/__init__.py
@@ -10,3 +10,12 @@ supported version: `draft-ietf-acme-01`_.
   https://github.com/ietf-wg-acme/acme/tree/draft-ietf-acme-acme-01
 
 """
+import sys
+import warnings
+
+if sys.version_info[:2] == (3, 3):
+    warnings.warn(
+            "Python 3.3 support will be dropped in the next release of "
+            "acme. Please upgrade your Python version.",
+            PendingDeprecationWarning,
+    ) #pragma: no cover
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -850,6 +850,9 @@ def main(cli_args=sys.argv[1:]):
         # Let plugins_cmd be run as un-privileged user.
         if config.func != plugins_cmd:
             raise
+    if sys.version_info[:2] == (3, 3):
+        logger.warning("Python 3.3 support will be dropped in the next release "
+                    "of Certbot - please upgrade your Python version.")
 
     set_displayer(config)
 


certbot/client.py
```diff
@@ -240,6 +241,9 @@ def __init__(self, config, account_, auth, installer, acme=None):
                 auth, self.acme, self.account, self.config.pref_challs)
         else:
             self.auth_handler = None
+        if sys.version_info[:2] == (3, 3):
+            logger.warn("Python 3.3 support will be dropped in the next release"
+                        "of Certbot. Please upgrade your Python version")
```
hugovk: Tiny niggle: the \_\_init__.py warning ends in a fullstop.

See also http://www.python3statement.org/practicalities/ for some ideas.

bmw:
Tests are going to fail until #5291 is merged. I'm cancelling the Travis build but will kick it when they're expected to pass again.

bmw:
I haven't reviewed the PR yet, but it looks like coverage tests are failing. Let me know if you'd like help with this.

hugovk:
The coverage drop is in:
```
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
acme/acme/__init__.py                   4      1    75%   17
```

Which is the file changed in this PR:
```
+if sys.version_info[:2] == (3, 3):
 +    warnings.warn(
 +            "Python 3.3 support will be dropped in the next release of"
 +            "Certbot. Please upgrade your Python version.",
 +            PendingDeprecationWarning,
 +    )
```

The problem is that coverage is only tested on Python 2.7, therefore these lines are not hit.

To allow an individual, non-3.3, build retain 100%, those lines should be excluded. See https://coverage.readthedocs.io/en/coverage-4.2/excluding.html

Alternatively, the coverage build could be switched to Python 3.3.

Ideally, all Python versions should be run with coverage, and all sent to Coveralls. Then they will all be combined and the pass/fail can be done at the build-level, not an individual build job-level.

SwartzCr:
@hugovk any chance we can interest you in contributing to Certbot :)
I've really appreciated your comments on this issue, but it also seems like you'd be well equipped to help us on this and other issues

hugovk:
Yes, I'll be happy to contribute where (and when) I can!

Any preference on the approach to take for this?

hugovk:
Maybe just excluding these lines? It's easiest, the excluded code is simple enough, and will be removed after the next release.

SwartzCr:
I'd be willing to exclude them if @bmw was comfortable with that.
As for other issues to look at, maybe #5275? I think that's the most pressing volunteer level task we have open right now. If that doesn't interest you, I'd be happy to suggest other things.
Also if you want to hang out in #letsencrypt-dev on freenode, that's where we chat about what we're working on :)

acme/acme/__init__.py
```diff
@@ -10,3 +10,12 @@
   https://github.com/ietf-wg-acme/acme/tree/draft-ietf-acme-acme-01
 
 """
+import sys
+import warnings
+
+if sys.version_info[:2] == (3, 3):
+    warnings.warn(
+            "Python 3.3 support will be dropped in the next release of"
+            "Certbot. Please upgrade your Python version.",
```
bmw: We need a space between "of" and "Certbot".

certbot/client.py
```diff
@@ -240,6 +241,9 @@ def __init__(self, config, account_, auth, installer, acme=None):
                 auth, self.acme, self.account, self.config.pref_challs)
         else:
             self.auth_handler = None
+        if sys.version_info[:2] == (3, 3):
+            logger.warning("Python 3.3 support will be dropped in the next release"
+                        "of Certbot - please upgrade your Python version.")
```
bmw: We need a space between "release" and "of"

certbot/client.py
```diff
@@ -240,6 +241,9 @@ def __init__(self, config, account_, auth, installer, acme=None):
                 auth, self.acme, self.account, self.config.pref_challs)
         else:
             self.auth_handler = None
+        if sys.version_info[:2] == (3, 3):
```
bmw: I don't think we should do this in this file and I instead think we should do it in code that Certbot (almost) always runs such as code in `main.py`. The reason for this is some user's might not get this warning. For instance, someone running `certbot renew` only gets this message when their certificate is attempted to be renewed which only happens every 60 days by default. Since we have been doing a release every month, some user's in this common case may miss the warning entirely.

acme/acme/__init__.py
bmw: Also, we should say "acme" here as there are other ACME clients using this module.

acme/acme/__init__.py
```diff
@@ -10,3 +10,12 @@
   https://github.com/ietf-wg-acme/acme/tree/draft-ietf-acme-acme-01
 
 """
+import sys
+import warnings
+
+if sys.version_info[:2] == (3, 3):
+    warnings.warn(
+            "Python 3.3 support will be dropped in the next release of "
+            "Certbot and acme. Please upgrade your Python version.",
```
bmw: I'm sorry that my previous comments were ambiguous. While I'm glad you included acme, I don't think we should mention Certbot at all here. [Many](https://github.com/zenhack/simp_le) [other](https://github.com/costela/wile) [unrelated](https://github.com/adfinis-sygroup/acme-powerdns/blob/master/setup.py) [projects](https://github.com/alex/letsencrypt-aws) use our ACME library and Certbot users will get the warning from Certbot itself.

certbot/main.py
```diff
@@ -482,6 +482,9 @@ def _init_le_client(config, authenticator, installer):
         #crypto_util.validate_key_csr(acc.key)
     else:
         acc, acme = None, None
+    if sys.version_info[:2] == (3, 3):
```
bmw: Doing this here has the same problem as I mentioned before. People running `certbot renew` won't get notifications until their certificate is renewed. I think we should do this in `certbot/main.py` in `main()` sometime after the `try` `except` blocks around `log.post_arg_parse_setup`.

PR #5204, opened 2017-10-26
Add initial ACME v2 support

rolandshoemaker:
Most of this work was done by @jsha. This PR adds baseline support to the `acme` module for a ACME server running the 'v2' (i.e. most recent draft) version of the specification. This doesn't include finalization support yet (i.e. how you actually issue a certificate). It attempts to be completely backward compatible without a user of the module to have to do anything fancy to switch between API versions.

Currently no tests, I'm working on them and will push shortly... honest!

PR #5287, opened 2017-12-01
client.py changes for ACMEv2

jsha:
- Client and ClientNetwork both take an acme_version and account parameter.
  - If acme_version is 2, ClientNetwork will add the account URL and the request URL in the header of signed JWS requests, as required by ACMEv2.
  - If acme_version is 2, Client will use the "new-account" field from the directory for register accounts, instead of "new-reg".

- Registration now has a "terms-of-service-agreed" field. In ACMEv2, you must set this to true at registration time if you agree to the terms as listed in the directory.


bmw:
py27-oldest tests will be fixed when #5291 is merged, but it looks like there are some lint failures as well.

jsha:
Fixed, thanks.

bmw:
Since tests are going to fail until #5291 is merged, I'm cancelling the Travis build but will kick it when they're expected to pass again.

acme/acme/client.py
```diff
@@ -527,7 +540,7 @@ def __del__(self):
         except Exception:  # pylint: disable=broad-except
             pass
 
-    def _wrap_in_jws(self, obj, nonce):
+    def _wrap_in_jws(self, obj, nonce, url):
```
bmw: We should document the `url` parameter below.

acme/acme/client.py
```diff
@@ -56,16 +58,20 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
     """
 
-    def __init__(self, directory, key, alg=jose.RS256, verify_ssl=True,
-                 net=None):
+    # pylint: disable=too-many-arguments
```
bmw: My understanding is putting this comment here disables this check for the rest of the class. Instead, we should put it below the docstring in `__init__` for the same effect to help warn us when classes and functions might be getting too complicated in the future.

acme/acme/client.py
```diff
@@ -509,15 +518,19 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
     JSON_ERROR_CONTENT_TYPE = 'application/problem+json'
     REPLAY_NONCE_HEADER = 'Replay-Nonce'
 
-    def __init__(self, key, alg=jose.RS256, verify_ssl=True,
-                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT):
+    # pylint: disable=too-many-arguments
```
bmw: Same comment as above. We should place this in the body of the `__init__` method.

acme/acme/client.py
```diff
@@ -509,15 +518,19 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
...
+    def __init__(self, key, account=None, alg=jose.RS256, verify_ssl=True,
+                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT,
+                 acme_version=2):
```
bmw: Why make the default here 2 rather than 1? Why not keep existing behavior unless you ask for something different?

acme/acme/client.py
```diff
@@ -509,15 +518,19 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
     JSON_ERROR_CONTENT_TYPE = 'application/problem+json'
     REPLAY_NONCE_HEADER = 'Replay-Nonce'
 
-    def __init__(self, key, alg=jose.RS256, verify_ssl=True,
-                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT):
+    # pylint: disable=too-many-arguments
+    def __init__(self, key, account=None, alg=jose.RS256, verify_ssl=True,
```
bmw: While I care less about the usage of `acme_version` in this class than I do in `Client`, if we take the separate class approach for `Client`, it might make sense to do that here as well for consistency. We could put these in a file called `client2.py` if we wanted.

acme/acme/client_test.py
```diff
@@ -432,9 +449,22 @@ def test_revoke_bad_status_raises_error(self):
...
+        return {'foo': self.value}
+
+    @classmethod
+    def from_json(cls, value):
+        pass  # pragma: no cover
+
 
 class ClientNetworkTest(unittest.TestCase):
     """Tests for acme.client.ClientNetwork."""
+    # pylint: disable=too-many-public-methods
```
bmw: I don't think this comment is needed.

jsha: Pylint errors without it:

```
$ pylint --reports=n --rcfile=.pylintrc acme/acme/client_test.py
************* Module acme.client_test
R:465, 0: Too many public methods (21/20) (too-many-public-methods)
```

acme/acme/client.py
```diff
@@ -48,6 +48,8 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
 
     :ivar messages.Directory directory:
     :ivar key: `.JWK` (private)
+    :ivar account: `.Account` (private)
+    :ivar acme_version: `int` (private)
```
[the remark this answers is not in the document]
jsha: > Looking forward to the changes in #5204 though, these changes are going to be more and more widespread.

I have one more v2 change coming down the line. In my opinion, the `Client` class won't be dramatically changed by it; there will be two new public methods, and one new private method.

> What happens if you call new_order on a client for ACMEv1 or request_challenges on a client for ACMEv2?

For new_order on an ACMEv1 client, we can pretty easily do the right thing: Turn it into a set of request_challenges. For request_challenges on ACMEv2 we could arguably synthesize an order out of them (now that you don't need a CSR to create an order), but it is probably better to just blow up.

In my opinion, on balance, the changes are not obtrusive enough to justify a whole new class.

acme/acme/client.py
jsha: Whoops, this was a mistake, accidentally left in. Fixed now.

acme/acme/client.py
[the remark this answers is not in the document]
bmw: > For new_order on an ACMEv1 client, we can pretty easily do the right thing: Turn it into a set of request_challenges. For request_challenges on ACMEv2 we could arguably synthesize an order out of them (now that you don't need a CSR to create an order), but it is probably better to just blow up.

Do we want to do this work though? My main concern here is making it clear to both existing and future ACME clients what methods they can and should use for each ACME version. I think separate classes (with a common base class containing most of the functionality) makes this self documenting. If you're using an ACMEv1 client, you can't call `new_order` because the method doesn't exists. Another example that presumably doesn't make sense is calling `agree_to_tos` on an ACMEv2 client.

If we don't go with the separate classes approach, we can try to document in the docstring for the class or each method what you can and can't use for each ACME version, but this to me feels awkward and less clear with the only major downside to the separate class approach being some initial refactoring up front.

What are the benefits in your mind to keeping everything in the same class?

acme/acme/client_test.py
bmw: Huh. I certainly tried to test this before I added this comment, but obviously I didn't do so correctly. Sorry for the misleading noise!

jsha:
@bmw can you describe the exact refactoring you're suggesting in https://github.com/certbot/certbot/pull/5287#discussion_r156540010? Right now I'm blocked on this.

bmw:
Sure!

What I'm imagining is we have a class named something like `ClientBase` that contains all of the functionality common to ACMEv1 and ACMEv2. The majority of the current contents of the `Client` class will go here. I think the only public methods of the current `Client` class in `master` that don't belong here are `agree_to_tos`, `request_challenges`, and `request_domain_challenges`, but a second check verifying I'm not missing anything is probably a good idea.

Protected methods can all be copied over. In theory, it would be nice for anything that is really only for a specific version of ACME to be placed into the class for that version of the client, but I'm less worried about this as we can change these methods later to our hearts content.

Inheriting from `ClientBase`, we have two new classes. There's `Client` which contains all of the ACMEv1 client that doesn't belong in `ClientBase` and then another class named something like `Client2` for all of the ACMEv2 functionality.

I think all of the tests for `Client` should continue to work. Tests would just have to be written for the new `Client2` class.

acme/acme/client.py
```diff
@@ -288,6 +199,134 @@ def poll(self, authzr):
...
+    :ivar alg: `.JWASignature`
+    :ivar bool verify_ssl: Verify SSL certificates?
+    :ivar .ClientNetwork net: Client network. Useful for testing. If not
+        supplied, it will be initialized using `key`, `alg` and
+        `verify_ssl`.
+
+    """
+
+    def __init__(self, directory, key, account=None, acme_version=1, alg=jose.RS256,
```
bmw: The parameters here that aren't currently in the `master` version of the client should probably be removed.

acme/acme/client.py
```diff
@@ -481,25 +520,39 @@ def fetch_chain(self, certr, max_length=10):
...
+    """ACME client for a v2 API.
 
-        :raises .ClientError: If revocation is unsuccessful.
+    :ivar messages.Directory directory:
+    :ivar .ClientNetwork net: Client network. Useful for testing. If not
+        supplied, it will be initialized using `key`, `alg` and
+        `verify_ssl`.
+    """
+
+    def __init__(self, directory, net):
```
bmw: nit: If we're not actually doing anything in `__init__` other than calling the parent class, we can omit it entirely.

bmw:
I haven't had a chance to review this again since it was updated a few days ago, but most of the ACMEv2 work (including #5374) will be built on the changes here. @joohoi or @ohemorange, please review this if you get a chance and feel comfortable doing so.

The major design discussion around these changes can be found [here](https://github.com/certbot/certbot/pull/5287#discussion_r155663152).

acme/acme/client.py
```diff
@@ -481,25 +517,30 @@ def fetch_chain(self, certr, max_length=10):
...
-        :param .ComparableX509 cert: `OpenSSL.crypto.X509` wrapped in
-            `.ComparableX509`
 
-        :param int rsn: Reason code for certificate revocation.
+class ClientV2(ClientBase):
+    """ACME client for a v2 API.
 
-        :raises .ClientError: If revocation is unsuccessful.
+    :ivar messages.Directory directory:
+    :ivar .ClientNetwork net: Client network. Useful for testing. If not
```
ohemorange: nit: update description, since we're requiring this now

acme/acme/client.py
```diff
@@ -288,6 +199,131 @@ def poll(self, authzr):
...
+    """ACME client for a v1 API.
+
+    .. todo::
+       Clean up raised error types hierarchy, document, and handle (wrap)
+       instances of `.DeserializationError` raised in `from_json()`.
+
+    :ivar messages.Directory directory:
+    :ivar key: `.JWK` (private)
+    :ivar account: `.Registration` (private)
+    :ivar acme_version: `int` (private)
```
ohemorange: I don't think we're using this anymore

acme/acme/client.py
```diff
@@ -288,6 +199,131 @@ def poll(self, authzr):
...
+                 net=None):
+        """Initialize.
+
+        :param directory: Directory Resource (`.messages.Directory`) or
+            URI from which the resource will be downloaded.
+
+        """
+        # pylint: disable=too-many-arguments
+        self.key = key
+        self.net = ClientNetwork(key, alg=alg, verify_ssl=verify_ssl) if net is None else net
```
ohemorange: wouldn't hurt to explicitly set `acme_version` here

acme/acme/client.py
```diff
@@ -509,15 +550,19 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
     JSON_ERROR_CONTENT_TYPE = 'application/problem+json'
     REPLAY_NONCE_HEADER = 'Replay-Nonce'
 
-    def __init__(self, key, alg=jose.RS256, verify_ssl=True,
-                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT):
+    def __init__(self, key, account=None, alg=jose.RS256, verify_ssl=True,
+                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT,
+                 acme_version=1):
```
ohemorange: I think it will be pretty easy to mess this up by forgetting to set `acme_version` here when using `Client2`. Maybe we could put some sort of assertion over in `Client2` that `acme_version` is 2, or something.

acme/acme/client.py
```diff
@@ -527,21 +572,31 @@ def __del__(self):
...
-            payload=jobj, key=self.key, alg=self.alg,
-            nonce=nonce).json_dumps(indent=2)
+        kwargs = {
+            "alg": self.alg,
+            "nonce": nonce
+        }
+        if self.acme_version == 2:
+            # new ACME spec
+            kwargs["url"] = url
+            if self.account is not None:
```
ohemorange: Is `account` only set for acmev2 uses? Is it required for acmev2 uses?

ohemorange:
Discussed offline:

Let's try making `ClientBase` be initialized with `acme_version` that's passed into `ClientNetwork.post`, so that `ClientNetwork` can be initialized (and `get` called) without needing an acme version.

This way, we can use `ClientNetwork` to fetch the directory, inspect it, then initialize the right sort of `Client` class (see #5491).

acme/acme/client.py
jsha: It is needed for acmev2 uses. It is not needed otherwise.

acme/acme/client.py
```diff
@@ -46,14 +46,16 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
...
+    def __init__(self, directory, net, acme_version):
         """Initialize.
 
         :param .messages.Directory directory: Directory Resource
         :param .ClientNetwork net: Client network.
+        :param int acme_version: ACME version. 1 or 2.
         """
         self.directory = directory
         self.net = net
+        self.acme_version = 1
```
ohemorange: is this intentional?

jsha: Nope! fixed.

acme/acme/client.py
```diff
@@ -288,6 +199,130 @@ def poll(self, authzr):
...
+        :param str new_authzr_uri: Deprecated. Do not use.
+
+        :returns: Authorization Resource.
+        :rtype: `.AuthorizationResource`
+
+        """
+        if new_authzr_uri is not None:
+            logger.debug("request_challenges with new_authzr_uri deprecated.")
+        new_authz = messages.NewAuthorization(identifier=identifier)
+        response = self.net.post(self.directory.new_authz, new_authz)
```
ohemorange: we might want to set `acme_version` to 1 here in case someone goes and changes the default

acme/acme/client.py
ohemorange: update: we probably want to remove acme_version=1 and self.acme_version = acme_version

acme/acme/client.py
```diff
@@ -527,21 +579,30 @@ def __del__(self):
...
-            payload=jobj, key=self.key, alg=self.alg,
-            nonce=nonce).json_dumps(indent=2)
+        kwargs = {
+            "alg": self.alg,
+            "nonce": nonce
+        }
+        if acme_version == 2:
+            kwargs["url"] = url
+            if self.account is not None:
+                kwargs["kid"] = self.account["uri"]
```
ohemorange: we either need to make sure this errors or pass account in directly

acme/acme/client.py
ohemorange: (still need to do this)

acme/acme/client.py
```diff
@@ -509,15 +565,19 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
...
     REPLAY_NONCE_HEADER = 'Replay-Nonce'
 
-    def __init__(self, key, alg=jose.RS256, verify_ssl=True,
-                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT):
+    def __init__(self, key, account=None, alg=jose.RS256, verify_ssl=True,
+                 user_agent='acme-python', timeout=DEFAULT_NETWORK_TIMEOUT,
+                 acme_version=1):
+        # pylint: disable=too-many-arguments
         self.key = key
+        self.account = account
```
ohemorange: Let's document that `account` must be set if this object might end up being used for acmev2

acme/acme/client.py
jsha: Can you remind me in more detail what we settled on for "this"?

ohemorange: Remove the now-dead code of `acme_version=1` and `self.acme_version = acme_version`

Because the `ClientNetwork` object doesn't need to know about acme version anymore, since it's all detailed in `post`

bmw:
I think we might want to make add a convenience `post` method to `ClientBase` that just calls `self.net.post` with the same arguments and adds a `acme_version` parameter, but we can do that later.

LGTM. Thanks for working on this.

acme/acme/client.py
```diff
@@ -527,21 +598,29 @@ def __del__(self):
...
-        return jws.JWS.sign(
-            payload=jobj, key=self.key, alg=self.alg,
-            nonce=nonce).json_dumps(indent=2)
+        kwargs = {
+            "alg": self.alg,
+            "nonce": nonce
+        }
+        if acme_version == 2:
+            kwargs["url"] = url
+            kwargs["kid"] = self.account["uri"]
```
ohemorange: Hold up, just noticed -- `messages.Registration` objects don't have a `uri`. What is this `uri`, and can we get it somewhere else?

bmw: I think `account` needs to be a `messages.RegistrationResource` object not a `messages.Registration` object. Luckily, this is what the `register` method already returns, so I think we just need to update the documentation for `ClientNetwork`.

PR #5293, opened 2017-12-04
Remove slash from document root path in Webroot example

michael-coleman:
It seems the document root path to the `--webroot-path`, `-w` option
can't have a trailing slash.  
Here is an example of a user who followed this example and had their
certificate signing request error out.  
https://superuser.com/questions/1273984/why-does-certbot-letsencrypt-recieve-a-403-forbidden

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -106,7 +106,7 @@ specified ``--webroot-path``. So, for instance,
 
 ::
 
-    certbot certonly --webroot -w /var/www/example/ -d www.example.com -d example.com -w /var/www/other -d other.example.net -d another.other.example.net
+    certbot certonly --webroot -w /var/www/example -d www.example.com -d example.com -w /var/www/other -d other.example.net -d another.other.example.net
 
 would obtain a single certificate for all of those names, using the
 ``/var/www/example`` webroot directory for the first two, and


michael-coleman:
great,  
I can tell you I experienced this problem myself earlier this year. not sure if you are interested in reproducing  the  problem in older  versions of certbot, but here is the best information I can give you to reproduce the error I experienced:  
From the `git blame` info on my own codebase, I experienced this *"nginx error if --webroot-path has a trailing slash"* problem around 21 June this year (2017).
  
The `certbot-auto` I installed was direct from github sources, i.e. I ran:

    git clone "https://github.com/certbot/certbot.git"  "/usr/local/src/certbot"

So assuming the default branch for the certbot repo back then was still `master` I would guess I was running the `certbot-auto`  somewhere around this commit? 

```
commit ed717d6bc483dff49f3d6fe1f9d7ee973127ac29
Author: Alexandre de Verteuil <alexandre@deverteuil.net>
Date:   Mon Jun 19 12:39:14 2017 -0400
```
I used the non interactive command - so I could run  it via a cron job and `certonly` so it didn't interfere with the running nginx.  i.e. I did something like this:

    ./certbot-auto certonly --webroot --webroot-path </doc/root/path> -d <DOMAIN> --non-interactive --agree-tos --email <EMAIL>

bmw:
Strange. I even tried a few `certbot-auto` releases before and after that and was unable to reproduce the problem. I believe we may have had this problem in the past but I think it's fixed now.

PR #5299, opened 2017-12-05
Nginx reversion

bmw:
Don't touch this PR unless your name is ohemorange.

The reason for this PR is many bug fixes in the nginx plugin for changes we haven't released yet are included in #5220 which may not make our next release. If it doesn't, we will (mostly) revert the nginx plugin back to its previous state to avoid releasing these bugs and will revert this PR after the release.


ohemorange:
#5299 should be left in, if there's no conflicts. That's just an existing bug that apparently no one's run into outside of my testing things when working on the other PRs.

bmw:
#5299 is this PR. Which one did you mean?

ohemorange:
oops 'Revert "Correctly test for existing Certbot redirect when adding an N…'

bmw:
Ah got it. There were no merge conflicts.

PR #5300, opened 2017-12-05
ACMEv2: Allow "uri" or "url" in challenge part 2

bmw:
This builds on #5258. This PR should be merged rather than squashed to preserve authorship.

In addition to the work from #5258, this PR modifies the methods necessary to hide the fact that uri and url are protected fields. The user can access them without the leading underscore in all public methods.

EDIT: Since the url attribute needs to be accessible through the public API for things like `__init__` and `encode`, I went ahead and made a `url` property that only returns the `_url` attribute. I didn't make it automatically use `_uri` or `_url` based on what was set because if you're asking for the `url`, you're explicitly requesting something from ACMEv2 rather than something from ACMEv1 where we're trying to keep backwards compatibility.

### DIFF
--- a/acme/acme/client_test.py
+++ b/acme/acme/client_test.py
@@ -181,7 +181,7 @@ class ClientTest(unittest.TestCase):
 
         # TODO: split here and separate test
         self.assertRaises(errors.UnexpectedUpdate, self.client.answer_challenge,
-                          self.challr.body.update(uri='foo'), chall_response)
+                          self.challr.body.update(_uri='foo'), chall_response)
 
     def test_answer_challenge_missing_next(self):
         self.assertRaises(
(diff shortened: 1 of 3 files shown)


jsha:
This looks fine but needs more explicit comments for each of the magic methods we're overriding.

Also, I think we should leave out the `def url()` for now, since it's likely to confuse people. Once more people are using ACMEv2, if we think it's valuable to harmonize the internal properties of the challenge object, we might add that in, but for now I'm fine with `acme` users continuing to think of the property as `uri`.

bmw:
I had to add the `url` attribute for consistency otherwise some methods such as `update` would do the wrong thing due to both `url` and `_url` being used to refer to the same value.

The new commit just does name mangling on `_uri` so it remains accessible through `uri` and does nothing with `_url`. We can add something like that if we want in the future.

What do you think?

PR #5302, opened 2017-12-06
Linode DNS Authenticator

trinopoty:
Implement an Authenticator which can fulfill a dns-01 challenge using the Linode API. Most of the code copied from DigitalOcean DNS Authenticator.

Testing Done:
- tox -e py27
- tox -e lint
- Manual Testing
  - Used certbot certonly --dns-linode -d, specifying a credentials file as a command line argument. Verified that a certificate was successfully obtained without user interaction.
  - Used certbot certonly --dns-linode -d, without specifying a credentials file as a command line argument. Verified that the user was prompted and that a certificate was successfully obtained.
  - Used certbot certonly -d. Verified that the user was prompted for a credentials file after selecting Linode interactively and that a certificate was successfully obtained.
  - Used certbot renew --force-renewal. Verified that certificates were renewed without user interaction.


bmw:
Oh wow. Thanks for the PR @trinopoty! I'll try and do a full review of this either tomorrow or sometime next week.

trinopoty:
One thing I should mention is, I copied `linode_api.py` from [https://github.com/tjfontaine/linode-python](https://github.com/tjfontaine/linode-python) in order to fix an import issue. As such, I had to disable pylint for `linode_api.py`.

trinopoty:
Since I've added support for Linode DNS to [Lexicon](https://github.com/AnalogJ/lexicon), should I switch to that instead?

bmw:
> Since I've added support for Linode DNS to Lexicon, should I switch to that instead?

That would be fantastic. In general, the approach we've taken so far is to use any Python libraries offered by the DNS provider and fall back to Lexicon if those libraries don't exist. This should result in a smaller PR and less code for us to maintain going forward.

Be sure to take a look at our other DNS plugins using Lexicon as there should be a lot of code you can reuse. Don't hesitate to ask if you have any questions!

trinopoty:
There is no official python SDK for the current version of the Linode API. I'll switch to Lexicon and update this PR.

trinopoty:
I've switched to lexicon for the certbot-dns-linode plugin.

bmw:
Thanks for doing that work @trinopoty. I'm planning on giving this a full review tomorrow.

trinopoty:
No problem.

bmw:
Another week has come and gone and I still haven't reviewed this. I'm sorry (again) @trinopoty. At a high level this PR looks great, we really appreciate you writing it, and I feel bad it's still sitting here without a review. I've just been working on things that have a technical deadline for us (like #2208) or removing things that are blocking people from doing more work on Certbot.

I'll try to get to this in the first couple weeks of January. @joohoi, if you feel comfortable reviewing this and think you can get to this before then, I'd appreciate a review here.

MrBMT:
@bmw Any chance of this PR being reviewed?

As someone that uses Linode DNS across numerous domains, it would be great to see support for this added, as I'm looking forward to being able to obtain Wildcard certificates for a number of them soon - and unless I'm mistaken DNS validation will be the only way to go about this?

bmw:
@MrBMT, we'll definitely try and review this sooner rather than later, but it unfortunately probably won't be merged and packaged before Let's Encrypt launches their new endpoint with wildcard support. I'd expect this to land shortly after that though.

leonardder:
I'd also be interested in this pr being reviewed and integrated.

leonardder:
@bmw: i was wondering whether this is still on your radar?

bmw:
@zjs, do you have time to review this?

certbot-dns-linode/setup.py
```diff
@@ -0,0 +1,69 @@
...
+
+from setuptools import setup
+from setuptools import find_packages
+
+version = '0.23.0.dev0'
+
+# Please update tox.ini when modifying dependency version requirements
+install_requires = [
+    'acme=={0}'.format(version),
+    'certbot=={0}'.format(version),
```
zjs: It seems like this pattern may have changed. (@bmw: Could you comment on this? It seems like other plugins now express a specific version number here.)

certbot-dns-linode/setup.py
```diff
@@ -0,0 +1,69 @@
...
+    'dns-lexicon>=2.1.16',
+    'mock',
+    # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
+    # will tolerate; see #2599:
+    'setuptools>=1.0',
```
zjs: It looks like #5617 removed dependence on a minimum version of `setuptoools`.

certbot-dns-linode/setup.py
```diff
@@ -0,0 +1,69 @@
...
+    license='Apache License 2.0',
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Environment :: Plugins',
+        'Intended Audience :: System Administrators',
+        'License :: OSI Approved :: Apache Software License',
+        'Operating System :: POSIX :: Linux',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.6',
```
zjs: It looks like #5508 dropped support for specific versions of Python (incl. 2.6 and 3.3).

We also seem to express that with an additional line between the license and the classifiers like: `python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',`.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self, api_key):
+        super(_LinodeLexiconClient, self).__init__()
+        self.provider = linode.Provider({
+            'auth_token': api_key
+        })
+
+    def _handle_general_error(self, e, domain_name):
+        if not str(e).startswith('Domain not found'):
+            return errors.PluginError('Unexpected error determining zone identifier for {0}: {1}'
```
zjs: I wasn't able to hit this code path. Is there a specific type of bad input that leads here?

certbot-dns-linode/certbot_dns_linode/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+---------------
+
+==========================================  ===================================
+``--dns-linode-credentials``          Linode credentials_ INI file.
+                                            (Required)
+``--dns-linode-propagation-seconds``  The number of seconds to wait for
+                                            DNS to propagate before asking the
+                                            ACME server to verify the DNS
+                                            record.
+                                            (Default: 930)
```
zjs: This is a very long time compared to the other DNS plugins (which mostly take 10-60 seconds). It seems unlikely to confuse an interactive user (as we print `Waiting 930 seconds for DNS changes to propagate` right before we wait). However, I wonder what impact this might have on someone with many distinct certificates which need renewal (it seems like we combine the waiting process for each certificate, but do not combine the process _across_ certificates).

I'm not sure there's any change needed as a part of this PR, but wanted to mention it for completeness. We could perhaps call this out more explicitly in the documentation, although it may already be clear enough.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
```diff
@@ -0,0 +1,70 @@
...
+
+    description = 'Obtain certs using a DNS TXT record (if you are using Linode for DNS).'
+
+    def __init__(self, *args, **kwargs):
+        super(Authenticator, self).__init__(*args, **kwargs)
+        self.credentials = None
+
+    @classmethod
+    def add_parser_arguments(cls, add):  # pylint: disable=arguments-differ
+        super(Authenticator, cls).add_parser_arguments(add, default_propagation_seconds=930)
```
zjs: 930 seconds was not _always_ sufficient in my testing. Perhaps I just got unlucky, but once in 5 attempts it took between 930 and 960.

Are you consistently seeing propagation times under 930 seconds? If so, I think this is fine as-is. If you see a higher time occasionally, it might be worth bumping this up. Waiting an extra 30 seconds every time is probably better than having to retry even 5% of the time.

trinopoty: Linode updates it's DNS records from the dashboard (which is updated by the api) every 15 minutes (900 seconds), so I set the value to 930 to ensure all the linode nameservers are updated at this point. At this point, linode servers are guaranteed to respond with the new value. It should not take longer than this unless some DNS resolver caches the values. I'll change the value to 960 just in case.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
trinopoty: This was supposed to hit when the domain wasn't added to linode. I'll check it out.

certbot-dns-linode/setup.py
zjs: #5655 may be a helpful reference.

trinopoty:
I've completed all the changes.
Added a Dockerfile (copied and modified from nson) though I'm not sure it'll work at the moment.

trinopoty:
Guess I forgot to change the min `default_propagation_seconds`.

trinopoty:
I'm done with the changes and ready for another review.

bmw:
I think using the official Linode library would be best. Looks like it's still in beta at the moment so if it's too unstable or buggy using Lexicon would be fine, but in general we've been preferring official Python libraries when they exist.

trinopoty:
I did some research. Seems like the new v4 API requires a key generated from the new console at `https://cloud.linode.com/`. The v4 API does not work with key generated from the existing console. Also, the new console is still under development, most specifically, the domain feature is unavailable in the new console though the v4 API does work with domains added in the old console.

I think it's best not to move to the new API at this point. The old API will still be available for a long time. We can migrate to the v4 API after both the new API and the new console has been finalized.

bmw:
> Also, the new console is still under development, most specifically, the domain feature is unavailable in the new console though the v4 API does work with domains added in the old console.

Can you tell me more about the domains feature that doesn't exist in the new console or link me to more info?

trinopoty:
@bmw If you login to the new [linode console](https://cloud.linode.com/), you can find that almost everything except the VPS feature is under development at the moment, including the domains feature.
You can also see this in my [screenshot](https://prnt.sc/jd21mn).

bmw:
Makes sense. Thanks for digging into that. Sticking with the v3 API using Lexicon seems like the right approach for now then.

@zjs, do you want to review the follow up commits here or would you like someone else to do it?

@trinopoty, please don't rebase this PR to make reviewing what you've changed since zjs last looked at it easier.

certbot-dns-linode/certbot_dns_linode/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+---------------
+
+==========================================  ===================================
+``--dns-linode-credentials``                Linode credentials_ INI file.
+                                            (Required)
+``--dns-linode-propagation-seconds``        The number of seconds to wait for
+                                            DNS to propagate before asking the
+                                            ACME server to verify the DNS
+                                            record.
+                                            (Default: 930)
```
zjs: Thanks for updating the default value below. I believe this should be updated to match.

https://github.com/certbot/certbot/blob/96aeb0b5479d869d9d6176f1e033ef88ddd41bdd/certbot-dns-linode/certbot_dns_linode/dns_linode.py#L32

certbot-dns-linode/setup.py
```diff
@@ -0,0 +1,65 @@
+import sys
+
+from setuptools import setup
+from setuptools import find_packages
+
+version = '0.24.0.dev0'
```
zjs: This should probably be `0.25.0.dev0` now. (Sorry for the slow turn-around time on this re-review.)

certbot-dns-linode/setup.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Please update tox.ini when modifying dependency version requirements
```
zjs: It looks like cc344bfd1 changed the equivalent comment in other `setup.py` files:

```
# acme/certbot version.
```

It looks like that same commit introduced a `local-oldest-requirements.txt` file in the root of each plugin. I'm not familiar with that change, but I assume you'll need to add one.

trinopoty:
Review changes completed.

s3gfaultx:
Would it be possible to provide a status update on this?

leonardder:
Not trying to be harsh here, but it really feels like this is never going to be merged eventually. Would it be possible to set something like a milestone for this?

PR #5313, opened 2017-12-08
Relax setuptools dependency

bmw:
I feel like I've been saying this a lot lately: We did this due to pip's lack of dependency resolution.

The problem as described in #2599 was `cryptography` specified a minimum `setuptools` version in their package which was being overridden Certbot's unrestricted dependency on `setuptools`. Since these changes landed, `cryptography` has stopped specifying a minimum version in their requirements, both in [master](https://github.com/pyca/cryptography/blob/9aec3f918a1c89e86732303a0cfaf50de930ce74/setup.py) and in the [version pinned in certbot-auto](https://github.com/pyca/cryptography/blob/2.0.2/setup.py).

This pinning of `setuptools>=1.0` caused some minor problems for package maintainers as described in #2599 and is causing me problems in [our tests](https://travis-ci.org/certbot/certbot/jobs/310342309#L9925) where we're testing against `setuptools 0.9.8` to keep compatibility for CentOS/RHEL 7. Since it's no longer needed to satisfy `pip`, let's remove it.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -17,9 +17,7 @@ install_requires = [
     'pyrfc3339',
     'pytz',
     'requests[security]>=2.4.1',  # security extras added in 2.4.1
-    # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
-    # will tolerate; see #2599:
-    'setuptools>=1.0',
+    'setuptools',  # pkg_resources
     'six>=1.9.0',  # needed for python_2_unicode_compatible
 ]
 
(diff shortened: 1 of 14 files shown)


bmw:
Tests failed because in our "oldest" tests, we have a version of `cryptography` that requires `setuptools>=1.0` and we're explicitly testing with `setuptools 0.9.8` since it's the oldest version of `setuptools` we have to support. With these changes, `pip` doesn't automatically upgrade `setuptools` to a newer version.

I'm working on pinning everything down in our oldest tests and will revisit this PR when that has landed.

PR #5314, opened 2017-12-08
Make Travis builds faster in master

bmw:
The changes here were:

1. Remove all but one `letsencrypt-auto` test. They all do the same thing except run on different OSes in Docker and I cannot recall ever seeing one fail and the others succeed (besides the fact that `tests/modification-check.sh` runs in the `le_auto_trusty` environment and sometimes catches bugs so I've left that one in).
2. Remove the `dockerfile-dev` test which is just to make sure that our developer dockerfile still works.
3. Rather than testing on all versions of Python 3, let's test on the newest and the oldest. If it works on both, it's likely to work on all Python 3 versions.
4. I reorderd the Travis jobs based on runtime, keeping `lint` slightly higher than normal just because it fails more often.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,22 +39,7 @@ matrix:
       before_install:
       addons:
     - sudo: required
-      env: TOXENV=le_auto_precise
-      services: docker
-      before_install:
-      addons:
-    - sudo: required
       env: TOXENV=le_auto_trusty
-      services: docker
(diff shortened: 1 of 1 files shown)


ohemorange:
I assume these are all going to run on the nightly branch?

bmw:
Yep!

PR #5315, opened 2017-12-09
Update the test everything branch

bmw:
The important thing here is `.travis.yml` remained unchanged after additional tests were removed from `master`.


bmw:
This PR should be merged and not squashed.

ohemorange:
Why is there so much in here?

bmw:
The tests merge `master` into this branch every night for testing, but it's not actually pushing those changes to our GitHub repo as we're not giving it permissions to do that so these changes are everything that's landed in `master` since this branch was made.

A saner way to look at this is to compare this branch and `master` locally.

PR #5316, opened 2017-12-09
Pin dependencies in oldest tests

bmw:
This PR pins dependencies in our "oldest" test which is the last place where Python dependencies aren't fully pinned. There are a couple pieces to this.

First, I moved our "oldest" dependencies into a file we can use as constraints for `pip`. Previously, the way this test worked was `tox` would install all of the dependencies listed in `tox.ini` and then `pip install` our packages. This resulted in `pip` installing newer versions of some packages than we specified in `tox.ini` which is resolved by providing these pinned versions as constraints for `pip`. While splitting this into its own file, I pinned back some more packages and added comments explaining where these versions came from.

Second, when `pip` is provided multiple constraints files, it appears to be undocumented what `pip` does if conflicting constraints are specified in different files. This is probably due to `pip`'s lack of real dependency resolution, but in practice, it appears that `pip` uses whichever version it sees first and ignores the constraints in another file. To give us more control over this and to not rely on strange behavior from `pip`, I added a simple script to merge requirements files controlling what is done when there are multiple requirements for the same dependency.


tools/merge_requirements.py
```diff
@@ -0,0 +1,61 @@
...
+    :returns: mapping from a project to its pinned version
+    :rtype: dict
+
+    """
+    d = {}
+    with open(file_path) as f:
+        for line in f:
+            line = line.strip()
+            if line and not line.startswith('#'):
+                project, _, version = line.partition('==')
```
ohemorange: You probably want `project, version = line.split('==')`.

bmw:
I ripped out the `CERTBOT_NO_PIN` stuff because with this PR it's dead code, but we'll just be adding something similar/identical when we set up tests using the latest version of everything so we should probably keep it.

tools/merge_requirements.py
bmw: There's a difference in what happens in these two versions if there are multiple `==` substrings in the line, but this version is probably better. I made the change!

PR #5321, opened 2017-12-11
Pin josepy in letsencrypt-auto

bmw:


ohemorange:
I thought we don't check in `letsencrypt-auto-source/letsencrypt-auto` until launch? Or is that just the copy of it in the directory one higher?

bmw:
It's just the copy one level higher. There's some documentation on this at https://certbot.eff.org/docs/contributing.html#updating-certbot-auto-and-letsencrypt-auto but I'm happy to go into more detail with you about all our `*-auto`s if you have questions.

ohemorange:
Nah I just always forget which is which.

Issue #1823, opened 2015-12-08
CentOS 7 install misses python pieces


sorvani:
I have a [detailed post here](http://mangolassi.it/topic/7127/setting-up-letsencrypt-on-a-centos-7-nginx-proxy) on the issue.

If you are on a clean CentOS 7 minimal install the setup has problems because it does not include `python-tools` and `python-pip`

erikrose:
We'll need to also update the letsencrypt-auto-source/pieces/bootstrappers/.

PR #5329, opened 2017-12-15
Deprecate Python2.6 by using Python3 on CentOS/RHEL 6

### LINKED ISSUE #2208: Install Python3 via EPEL on CentOS6 with letsencrypt-auto?

ohemorange:
Fixes #2208.


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -296,11 +314,24 @@ elif [ -f /etc/mageia-release ]; then
...
   BOOTSTRAP_VERSION="BootstrapMageiaCommon $BOOTSTRAP_MAGEIA_COMMON_VERSION"
 elif [ -f /etc/redhat-release ]; then
-  Bootstrap() {
-    BootstrapMessage "RedHat-based OSes"
-    BootstrapRpmCommon
-  }
-  BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
+  PREV_LE_PYTHON="$LE_PYTHON"
+  DeterminePython2Version "NOCRASH"
+  if [ "$PYVER" -eq 26 ] || [ "$PYVER" -eq 0 ]; then
```
bmw: Did you find Red Hat systems that might not have Python installed? Is our Python 3 approach expected to work on them?

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -296,11 +314,24 @@ elif [ -f /etc/mageia-release ]; then
...
+    BOOTSTRAP_VERSION="BootstrapRpmPython3 $BOOTSTRAP_RPM_PYTHON3_VERSION"
+  else
+    Bootstrap() {
+      BootstrapMessage "RedHat-based OSes"
+      BootstrapRpmCommon
+    }
+    BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
+  fi
+  LE_PYTHON="$PREV_LE_PYTHON"
+  export LE_PYTHON
```
bmw: super nit: You can do this in one line like `export LE_PYTHON="$PREV_LE_PYTHON"`.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -245,23 +245,41 @@ DeprecationBootstrap() {
 }
 
 
+DeterminePython2Version() {
+  DeterminePythonVersionBase $1 "$LE_PYTHON python2.7 python27 python2 python"
+}
+
+# TODO: use this to not upgrade if we already have python3 installed
```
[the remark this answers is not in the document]
ohemorange: Yes. I thought having something like this might make it feasible, but on second thought we'd actually need all the other dependencies also, so unless we're switching entirely to checking for things before attempting to install them (which I am not proposing we do), we should probably just rebootstrap anyway.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -245,23 +245,41 @@ DeprecationBootstrap() {
...
 DeterminePythonVersion() {
-  for LE_PYTHON in "$LE_PYTHON" python2.7 python27 python2 python; do
+  # TODO: check that this is the only executable we might have installed
+  if [ "$USE_PYTHON_3" = true ] ; then
```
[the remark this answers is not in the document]
ohemorange: glorious

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -245,23 +245,41 @@ DeprecationBootstrap() {
...
   export LE_PYTHON
 
   PYVER=`"$LE_PYTHON" -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//'`
-  if [ "$PYVER" -lt 26 ]; then
+  if [ "$PYVER" -lt 26 ] && [ "$1" != "NOCRASH" ]; then
     error "You have an ancient version of Python entombed in your operating system..."
     error "This isn't going to work; you'll need at least version 2.6."
     exit 1
   fi
+  export PYVER
```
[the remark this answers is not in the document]
ohemorange: I'd actually forgotten that that's how it accomplishes things, thanks for the reminder.

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: We didn't previously assume that python would be installed, given that we try to install python as a dependency, so I was trying to keep that property. Unless we did assume that implicitly somewhere that I'd missed?

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  else
+    error "Neither yum nor dnf found. Aborting bootstrap!"
+    exit 1
+  fi
+}
+
+BootstrapRpmCommonBase() {
+  FindInstallTool
+  tool=$?
```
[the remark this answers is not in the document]
ohemorange: ach, global variables as default is so icky.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,74 @@
...
+    openssl
+    openssl-devel
+    libffi-devel
+    redhat-rpm-config
+    ca-certificates
+  "
+
+  # Add the python packages
+  pkgs = "$pkgs
+    $1
```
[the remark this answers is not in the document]
ohemorange: but the previous packages don't end in a newline?

or are you saying that we just don't actually care about newlines vs spaces and they're just there for programmer benefit?

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
+# If new packages are installed by BootstrapRpmPython3 below, this version
+# number must be increased.
+BOOTSTRAP_RPM_PYTHON3_VERSION=1
+
+BootstrapRpmPython3() {
+  # Tested with:
+  #   - CentOS 6 (EPEL must be installed manually)
```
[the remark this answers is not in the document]
ohemorange: I just copied this comment from the previous script. Is the comment made untrue by something changed in this PR, or was it just never updated previously? And this code runs at the same point as previously, so why would it be true for this PR but not true previously?

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
...
+  if $tool list python34 >/dev/null 2>&1; then
+    python_pkgs="python34
+      python34-devel
+      python-virtualenv
+      python34-tools
+      python-pip
+    "
+  fi
+  # TODO: add some elifs and elses for other distros that
+  # might have 2.6 or no python installed to get here
```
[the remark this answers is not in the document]
ohemorange: Yes, adding an else seems like the right call.

I don't have access yet, I'd love those stats. And also if you have any easier ways to check package lists on those systems than installing each one with docker.

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
ohemorange: done

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
+# If new packages are installed by BootstrapRpmPython3 below, this version
+# number must be increased.
+BOOTSTRAP_RPM_PYTHON3_VERSION=1
+
+BootstrapRpmPython3() {
+  # Tested with:
+  #   - CentOS 6 (EPEL must be installed manually)
+
+  FindInstallTool
+  tool=$?
```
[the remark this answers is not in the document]
ohemorange: done. by the way, how do I include the contents of `rpm_common_base.sh` in these scripts?

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
...
+      python34-devel
+      python-virtualenv
+      python34-tools
+      python-pip
+    "
+  fi
+  # TODO: add some elifs and elses for other distros that
+  # might have 2.6 or no python installed to get here
+
+  BootstrapRpmCommonBase $python_pkgs
```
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
```diff
@@ -9,59 +9,12 @@ BootstrapRpmCommon() {
...
-    gcc
-    augeas-libs
-    openssl
-    openssl-devel
-    libffi-devel
-    redhat-rpm-config
-    ca-certificates
-  "
+  FindInstallTool
+  tool=$?
```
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common.sh
```diff
@@ -82,23 +34,13 @@ BootstrapRpmCommon() {
...
-    pkgs="$pkgs
-      mod_ssl
-    "
-  fi
-
-  if ! $tool install $yes_flag $QUIET_FLAG $pkgs; then
-    error "Could not install OS dependencies. Aborting bootstrap!"
-    exit 1
-  fi
+  BootstrapRpmCommonBase $python_pkgs
```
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,74 @@
+# If new packages are installed by BootstrapRpmCommonBase below, version
+# numbers in rpm_common.sh and rpm_python3.sh must be increased.
```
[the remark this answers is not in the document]
ohemorange: :+1:

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,74 @@
...
+  then
+    tool=yum
+
+  else
+    error "Neither yum nor dnf found. Aborting bootstrap!"
+    exit 1
+  fi
+}
+
+BootstrapRpmCommonBase() {
```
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
...
+}
```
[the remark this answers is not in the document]
ohemorange: (done)

letsencrypt-auto-source/letsencrypt-auto.template
bmw: We didn't previously assume that, however, we do assume that [yum](https://en.wikipedia.org/wiki/Yellow_Dog_Updater,_Modified) is installed which (at least nowadays) is written in Python so it may be safe to make that assumption.

I'm just not sure if the `"$PYVER" -eq 0` path will ever be used and if it is, if it's even the right thing to do. Do we expect a red hat system without a Python to be CentOS/RHEL 6 where we can install python34 from EPEL?

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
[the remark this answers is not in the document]
bmw: > but the previous packages don't end in a newline?

I think it does by putting the closing quote on a separate line from the last package. For example:
```
#!/bin/sh
a="
  foo
  bar
"
b="
  baz
  qux
"
c="
  $a
  $b
"
echo "$c"
```
outputs:
```

  
  foo
  bar

  
  baz
  qux


```
If `yum` doesn't care though, I wouldn't worry about it too much.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: I forgot to update it previously apparently. See #3259.

If you don't fix my mistake, we should at least not duplicate it here.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: I'm pulling the stats now for distinct `certbot-auto` installs in the past 2 months using Python 2.6. I'll post results here when the query is done.

Unfortunately I really don't have a better way to query versions for Red Hat based systems. All browser search tools I've seen are really buggy/inaccurate. If persistence is at all useful, I think you should feel free to spin up droplets/EC2 instances while you're working on this issue.

bmw:
I responded to specific threads, but since you're still working on this, I'd like to hold off giving it another full review until you think it's ready. If there's specific parts you'd like me to look at though, let me know.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: Here's the count of distinct IPs using Python 2.6 and `certbot-auto` by OS.

I've only included OSes with at least 5 users as below that, some of the values get pretty specific and start to identify the user. There are 61 more different OS strings with less than 5 users.

Also, keep in mind that we don't officially support many of the OSes below and the user had to do some tinkering on their end to make things work.

| "os"                                  | "_count_distinct" | 
|---------------------------------------|-------------------| 
| "CentOS 6.9"                          | "22480"           | 
| "CentOS 6.8"                          | "8180"            | 
| "CentOS 6.7"                          | "3236"            | 
| "CentOS 6.5"                          | "2885"            | 
| "CentOS 6.6"                          | "2088"            | 
| "CentOS 6.4"                          | "815"             | 
| "Red Hat Enterprise Linux Server 6.9" | "455"             | 
| "CentOS 6.3"                          | "390"             | 
| "CentOS 6.2"                          | "220"             | 
| "Scientific Linux 6.9"                | "160"             | 
| "Oracle Linux Server 6.9"             | "108"             | 
| "CentOS Linux 6.0"                    | "83"              | 
| "Red Hat Enterprise Linux Server 6.8" | "73"              | 
| "Scientific Linux 6.8"                | "64"              | 
| "Oracle Linux Server 6.8"             | "53"              | 
| "Red Hat Enterprise Linux Server 6.7" | "51"              | 
| "Red Hat Enterprise Linux Server 6.6" | "47"              | 
| "Red Hat Enterprise Linux Server 6.5" | "45"              | 
| "SHMZ 6.6"                            | "42"              | 
| "CloudLinux Server 6.9"               | "31"              | 
| "Red Hat Enterprise Linux Server 6.4" | "29"              | 
| "SHMZ 6.5"                            | "23"              | 
| "ClearOS Community 6.9.0"             | "21"              | 
| "Scientific Linux 6.5"                | "18"              | 
| "debian 6.0.10"                       | "17"              | 
| "Oracle Linux Server 6.7"             | "15"              | 
| "CentOS 6.1"                          | "15"              | 
| "CentOS 5.11"                         | "14"              | 
| "Scientific Linux 6.7"                | "14"              | 
| "Oracle Linux Server 6.6"             | "11"              | 
| "Scientific Linux 6.1"                | "11"              | 
| "Red Hat Enterprise Linux Server 6.2" | "10"              | 
| "Ubuntu 10.04"                        | "10"              | 
| "Scientific Linux 6.4"                | "10"              | 
| "Scientific Linux 6.6"                | "10"              | 
| "Red Hat Enterprise Linux Server 6.3" | "9"               | 
| "Amazon Linux AMI 2014.09"            | "8"               | 
| "Oracle Linux Server 6.4"             | "8"               | 
| "Scientific Linux 6.3"                | "7"               | 
| "Scientific Linux CERN SLC 6.9"       | "7"               | 
| "Scientific Linux 6.2"                | "6"               | 
| "Oracle Linux Server 6.5"             | "6"               | 
| "Red Hat Enterprise Linux Server 6.0" | "6"               | 
| "Scientific Linux 6.0"                | "6"               | 
| "Red Hat Enterprise Linux Server 6.1" | "5"               | 
| "CentOS 5.8"                          | "5"               |

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: Yes, that is exactly what I was imagining. I didn't know that yum has python as a dependency.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
ohemorange: How are we now requiring the user to manually install EPEL here in a way that we weren't before? This script uses the same logic at the same points in code execution, with respect to when we run `$tool list python34`.

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: I'm happy to remove the check. Worst thing that happens is that we accidentally fall through, install python2.6, then next time around we rebootstrap.

ohemorange:
Yes, given that I intended this as a WIP that wasn't ready for a full review anyway, very happy to not have another full review yet.

I would appreciate responses to https://github.com/certbot/certbot/pull/5329#discussion_r157622044 and https://github.com/certbot/certbot/pull/5329#discussion_r157863085.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: Sorry. I somehow missed this one yesterday.

To do this, modify `letsencrypt-auto-source/letsencrypt-auto.template`. `letsencrypt-auto-source/build.py` replaces instances of `{{ path/to/file }}` found in that template with the contents of the file. File paths should be relative to the `letsencrypt-auto-source/pieces` directory.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: Because the result of `$tool list python34` depends on whether or not EPEL is installed.

`$tool list <package>` searches for a package with that name using the installed repositories. `$tool list python34` will fail until EPEL is installed because `python34` isn't available in the standard repositories, only in EPEL. If this bootstrap function is run on a CentOS 6 system without EPEL already installed, the `if $tool list python34` branch will not be taken.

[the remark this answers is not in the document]
bmw: For more info here, the code that installs EPEL if necessary is the block that starts with `if ! $tool list *virtualenv >/dev/null 2>&1; then`.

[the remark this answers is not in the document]
ohemorange: I feel like we're having a miscommunication here, so maybe we should talk in person tomorrow.

I understand why EPEL is needed to run this code, that's not what I'm asking. But don't we hit the code that installs EPEL before we get here? The way the code was before I ran this, we needed EPEL to check about python 2 packages, now we still need it to check about python 3 packages. In both cases, the code runs after we've installed EPEL for them. What am I missing that I've changed here to make it so we need EPEL earlier than we previously needed it?

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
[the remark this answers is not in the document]
ohemorange: Ah, yeah, yum definitely doesn't care because that's what we were doing before. Looking back now, I put it in on a newline to be precisely consistent with what it was before. Happy to fix this up while I'm here if it's more readable.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
ohemorange: Ah now I see. It's because we were previously running the enable EPEL code before checking for python packages, and now we're running it after checking for python packages because my refactoring changed the order.

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: Actually, if I change this to `if [ "$PYVER" -le 26 ]` and fix up the other function, we could theoretically just as easily support people who have no python, python 2.6, or like some weird python 2.5 or whatever if that's even a thing, all just as easily, by installing python3 for them. Why not do that?

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
ohemorange: As I understand it, using `epel-release` means package names are consistent, right? They're all pulling down the same package list once they have that? And they'll all have to, since everything here that we support that's redhat is a version of RHEL 6, essentially.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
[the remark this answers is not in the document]
bmw: If `yum` doesn't care and we were doing that before, it's fine with me.

bmw:
Currently this is blocked by the problem described at https://github.com/certbot/certbot/issues/2208#issuecomment-353217603.

letsencrypt-auto-source/letsencrypt-auto.template
bmw: If there's a way we can easily do that, I'm certainly not against it, however, I'm not sure it's that simple because we're installing Python 3.4 using Red Hat tooling, package names, and a repository that only works on systems based on RHEL 6.

ohemorange: yeah ok supporting more things somehow sounds less fun, I'm ripping this nocrash stuff out.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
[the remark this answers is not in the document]
bmw: To be a little pedantic to share some background and make sure we're on the same page, `yum install epel-release` installs a CentOS specific package that automates setting up the [EPEL](https://fedoraproject.org/wiki/EPEL) repository on that system. It doesn't work on RHEL for example.

With that said, yes, the package names and contents in EPEL 6 are the same for all the RHEL 6 variants supported by the repo.

bmw:
I think I responded to all the comments, but if there's anything I missed or new that you'd like my thoughts on, let me know.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -296,11 +307,21 @@ elif [ -f /etc/mageia-release ]; then
...
-    BootstrapRpmCommon
-  }
-  BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
+  DeterminePythonVersion "NOCRASH"
+  if [ "$PYVER" -eq 26 ]; then
+    Bootstrap() {
+      BootstrapMessage "RedHat-based OSes that will use Python3"
+      BootstrapRpmPython3
+    }
+    export LE_PYTHON=python3
```
ohemorange: Ok, changed some things around here and also inside `DeterminePythonVersion`. These are the cases I considered:

- their python as specified by `LE_PYTHON` doesn't exist 
- they didn't set le_python and they have no python installed
- le_python is 2.6
- le_python is 3.x
- le_python is 2.7

I can go through why I think this is the right decision for each of them, just not right now.

bmw:
And while still a workaround, I think I suggested an even easier approach in https://github.com/certbot/certbot/issues/2208#issuecomment-353251816.

bmw:
If you agree this new approach is simpler and cleaner, notice the CLI is slightly different and the flags we currently pass to `virtualenv` aren't supported by `venv`.

bmw:
The way I'd probably do this is to use `"$LE_PYTHON" -m venv` if the python version is 3.3+.

Also, I have a TODO item for myself to talk to you about how to write tests for this. Writing meaningful automated tests for `certbot-auto` is really hard due to all the pieces here, the Docker tests using mock/dummy packages for a lot of stuff, auto update mechanisms, and the versions of all Python packages being pinned inside the script.

With this in mind, I'm personally not too worried about you submitting tests here. I'd test manually and maybe hack together a test farm test for local use if there's something there you want to test. I'll be doing the same when I review it.

letsencrypt-auto-source/letsencrypt-auto.template
bmw: #### LE_PYTHON affecting bootstrap selection

I'm unsure if we want to be taking `LE_PYTHON` set by the user into account when determining the bootstrap script to use.

My main reason for thinking this is the bootstrap script changing based on the user's environment can lead to some strange/obnoxious behavior. For example, you're running certbot-auto in cron as root and you get root email that says that certbot-auto needs to be run interactively to get permission to install new packages. You log in as your normal user and run certbot-auto. If you have `LE_PYTHON` set in one environment and not the other, a different bootstrap script may be selected in the two environments. This is similar to the issues we were having when there wasn't one Certbot install location on the system and it depended on the user's home directory.

If `LE_PYTHON` is set for the non-root user, it may not rebootstrap despite certbot-auto saying you have to. If `LE_PYTHON` is set for the root user, you may rebootstrap when running as a non-root user, but when certbot-auto is run again from cron, a different bootstrap script is selected so certbot-auto thinks you have to rebootstrap again!

If we do this, I think the only downside is we may install packages we may end up not needing, but doing this for the very few users who have been setting `LE_PYTHON` seems like it might be less harmful, confusing, and complex than the above behavior. If users know nothing else needs to be installed, there's a `--no-bootstrap` flag.

#### Overwriting LE_PYTHON

When setting `LE_PYTHON` here though, I think we should respect the value the user set it to and not just overwrite it. Setting `LE_PYTHON=python2.6` and having it suddenly run in Python 3 is a little confusing and unexpected. I think we should only set this value to `python3` if it wasn't set by the user.

My thought here is in the next release, you'll still be able to force Certbot to use Python 2.6 if you set `LE_PYTHON`. In the next release, we'll also start printing deprecation warnings if you're running in Python 2.6. In a subsequent release, we'll update `DeterminePythonVersion` to complain about your entombed Python version if it's less than 27 which should force you to stop using 2.6 entirely.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,25 @@
...
+  # Tested with:
+  #   - CentOS 6
+
+  InitializeRPMCommonBase
+
+  # EPEL uses python34
+  if $TOOL list python34 >/dev/null 2>&1; then
+    python_pkgs="python34
+      python34-devel
+      python-virtualenv
```
bmw: This is largely copied from the existing bootstrap script so you don't have to change it if you want, but I don't think we need `python34-tools` and `python-pip`. If we're using `python3 -m venv`, I don't think we need this package either.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -0,0 +1,23 @@
...
+  # Tested with:
+  #   - CentOS 6
+
+  InitializeRPMCommonBase
+
+  # EPEL uses python34
+  if $TOOL list python34 >/dev/null 2>&1; then
+    python_pkgs="python34
+      python34-devel
+      python34-tools
```
bmw: Did you determine that this package is necessary?

letsencrypt-auto-source/tests/auto_test.py
```diff
@@ -349,6 +350,7 @@ def test_openssl_failure(self):
                     self.assertTrue("Couldn't verify signature of downloaded "
                                     "certbot-auto." in exc.output)
                 else:
+                    print(out)
```
bmw: Did you mean to leave this print statement here?

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -112,6 +120,17 @@ def verified_new_le_auto(get, tag, temp_dir):
                             "certbot-auto.", exc)
 
 
+def create_CERT_NONE_context():
+    """Create a SSLContext object to not check hostname."""
+    # PROTOCOL_TLS isn't available before 2.7.13 but this code is for 2.7.9+, so use this.
+    context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+
+    context.check_hostname = False
```
bmw: nit: I don't think this line is necessary if `verify_mode` is set to `CERT_NONE`.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -244,15 +244,29 @@ DeprecationBootstrap() {
...
-
+# Sets LE_PYTHON to Python version string and PYVER to the first two
+# digits of the python version
 DeterminePythonVersion() {
-  for LE_PYTHON in "$LE_PYTHON" python2.7 python27 python2 python; do
-    # Break (while keeping the LE_PYTHON value) if found.
-    $EXISTS "$LE_PYTHON" > /dev/null && break
-  done
+  # Arguments: "NOCRASH" if we shouldn't crash if we don't find a good python
+  if [ "$USE_PYTHON_3" -eq 0 ]; then
```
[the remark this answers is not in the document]
ohemorange: I agree. I wrote this before I realized there's a way to check for test if variables are set and set to something without crashing.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
ohemorange: nope, but I don't know what we were using it for in the first place, in the non-3 python version.

letsencrypt-auto-source/tests/auto_test.py
ohemorange: yes.

ohemorange:
For posterity, I was using these commands for installing python and testing manually and such:

```
sudo su

# from https://danieleriksson.net/2017/02/08/how-to-install-latest-python-on-centos/

# Start by making sure your system is up-to-date:
yum update
# Compilers and related tools:
yum groupinstall -y "development tools"
# Libraries needed during compilation to enable all features of Python:
yum install -y zlib-devel bzip2-devel openssl-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel expat-devel
# If you are on a clean "minimal" install of CentOS you also need the wget tool:
yum install -y wget

# Python 2.7.14:
wget http://python.org/ftp/python/2.7.14/Python-2.7.14.tar.xz
tar xf Python-2.7.14.tar.xz
cd Python-2.7.14
./configure --prefix=/usr/local --enable-unicode=ucs4 --enable-shared LDFLAGS="-Wl,-rpath /usr/local/lib"
make && make altinstall

cp /usr/local/bin/python2.7 /usr/bin/
cd -
exit


pytest -v -s certbot/letsencrypt-auto-source/tests

sudo certbot/letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade

sudo rm -f /usr/bin/python2.7
sudo certbot/letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade
```

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,76 @@
+# If new packages are installed by BootstrapRpmCommonBase below, version
+# numbers in rpm_common.sh and rpm_python3.sh must be increased.
+
+# Sets TOOL to the name of the package manager
```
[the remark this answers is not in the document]
bmw: This function also sets `QUIET_FLAG` and `yes_flag` that are used outside of the function. It's not a big deal, but documenting this seems like a good idea to me and we should probably change `yes_flag` to uppercase while we're at it.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
bmw: Looking into this, I found #1823. There and in the linked blogpost, one person claims `python-tools` and `python-pip` need to be installed without really saying why. Initially, the post talks about CentOS 7 and [later](https://mangolassi.it/topic/7127/setting-up-letsencrypt-on-a-centos-7-nginx-proxy) says they are necessary to fix `InsecurePlatformWarning`. A few things:

1. The installation of these packages do not fix `InsecurePlatformWarning` per my testing.
2. `InsecurePlatformWarning` is not a fatal error.
3. The error isn't shown in normal output (unless you asked for increase verbosity) because the error is irrelevant as we're verifying the hashes of the downloaded packages.

With that said, I looked into the contents of the `python34-tools` package using `repoquery` as described [here](https://stackoverflow.com/questions/104055/how-to-list-the-contents-of-a-package-using-yum). This package contains:

1. [2to3](https://docs.python.org/2/library/2to3.html) and [IDLE](https://docs.python.org/2/library/idle.html) placed in the user's PATH.
2. A bunch of Python developer tools and demos that aren't placed in the user's PATH or in a location importable by Python.

I think this package is cruft and can be removed, especially in the Python 3.4 case which doesn't (currently) have `InsecurePlatformWarning` problems. I think this cleans up `certbot-auto` and reduces our dependency count a bit, but the installed package is only 1.2 MB. If you'd prefer to leave it in despite what I've found, that's fine with me.

ohemorange: It sounds like this isn't a python3-specific problem, which is what this PR is about fixing. If we're removing it, it belongs in a separate PR that can be separately reverted if something goes wrong. Happy to also get that in for this release.

ohemorange:
Expected output of the test I just added:
```
user@machine:~/certbot/letsencrypt-auto-source$ ./build.py && docker build -f Dockerfile.centos6 -t lea . && docker run --rm -t -i lea 
Sending build context to Docker daemon 200.7 kB
Step 1/16 : FROM centos:6
 ---> df3764b1d215
Step 2/16 : RUN yum install -y epel-release
 ---> Using cache
 ---> b49f8dcbb76d
Step 3/16 : RUN yum install -y python-pip sudo
 ---> Using cache
 ---> 034ae3f17580
Step 4/16 : COPY ./pieces/pipstrap.py /opt
 ---> Using cache
 ---> 5fef3b4722ea
Step 5/16 : RUN /opt/pipstrap.py
 ---> Using cache
 ---> 8df8335f85ac
Step 6/16 : RUN pip install pytest==3.2.5
 ---> Using cache
 ---> b4ec309aa6ed
Step 7/16 : RUN useradd --create-home --home-dir /home/lea --shell /bin/bash --groups wheel --uid 1000 lea
 ---> Using cache
 ---> 94cf48ffe842
Step 8/16 : RUN sed -i.bkp -e       's/# %wheel\(NOPASSWD: ALL\)\?/%wheel/g'       /etc/sudoers
 ---> Using cache
 ---> 5dd3e730f731
Step 9/16 : RUN mkdir -p /home/lea/certbot
 ---> Using cache
 ---> 1ce4cfafca9c
Step 10/16 : COPY ./tests/certs/ca/my-root-ca.crt.pem /usr/local/share/ca-certificates/
 ---> Using cache
 ---> f4c64eaecf86
Step 11/16 : RUN update-ca-trust
 ---> Using cache
 ---> e9c210bdf50d
Step 12/16 : COPY . /home/lea/certbot/letsencrypt-auto-source
 ---> 0249d6180351
Removing intermediate container 647ce3a0ac85
Step 13/16 : USER lea
 ---> Running in efc4c2d9dae7
 ---> 2af281e22d92
Removing intermediate container efc4c2d9dae7
Step 14/16 : WORKDIR /home/lea
 ---> 89cec1c128a7
Removing intermediate container 372f4cbb8e65
Step 15/16 : RUN sudo chmod +x certbot/letsencrypt-auto-source/tests/centos6_tests.sh
 ---> Running in f07f1390a3f7
 ---> abe1de68b910
Removing intermediate container f07f1390a3f7
Step 16/16 : CMD sudo certbot/letsencrypt-auto-source/tests/centos6_tests.sh
 ---> Running in 864a4080d505
 ---> de9a946aad41
Removing intermediate container 864a4080d505
Successfully built de9a946aad41

PASSED: Did not upgrade to Python3 when Python2.7 is present.
PASSED: Successfully upgraded to Python3 when only Python2.6 is present.

============================================================================== test session starts ==============================================================================
platform linux2 -- Python 2.6.6, pytest-3.2.5, py-1.4.34, pluggy-0.4.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/lea, inifile:
collected 5 items                                                                                                                                                                

certbot/letsencrypt-auto-source/tests/auto_test.py::tests_dir SKIPPED
certbot/letsencrypt-auto-source/tests/auto_test.py::AutoTests::test_openssl_failure PASSED
certbot/letsencrypt-auto-source/tests/auto_test.py::AutoTests::test_phase2_upgrade PASSED
certbot/letsencrypt-auto-source/tests/auto_test.py::AutoTests::test_pip_failure PASSED
certbot/letsencrypt-auto-source/tests/auto_test.py::AutoTests::test_successes PASSED

===================================================================== 4 passed, 1 skipped in 42.00 seconds ======================================================================
```

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -296,11 +312,24 @@ elif [ -f /etc/mageia-release ]; then
   }
   BOOTSTRAP_VERSION="BootstrapMageiaCommon $BOOTSTRAP_MAGEIA_COMMON_VERSION"
 elif [ -f /etc/redhat-release ]; then
-  Bootstrap() {
-    BootstrapMessage "RedHat-based OSes"
-    BootstrapRpmCommon
-  }
-  BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
+  prev_le_python="$LE_PYTHON"
```
erikrose: I think I finally figured out what's going on here, but I could use a nice big comment, something like the lines of "We run DeterminePythonVersion to decide, on the basis of available Python versions, whether to use 2.x or 3.x on RedHat-like systems. Then we revert its return values to their previous state."

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -244,15 +244,29 @@ DeprecationBootstrap() {
...
-
+# Sets LE_PYTHON to Python version string and PYVER to the first two
+# digits of the python version
 DeterminePythonVersion() {
-  for LE_PYTHON in "$LE_PYTHON" python2.7 python27 python2 python; do
-    # Break (while keeping the LE_PYTHON value) if found.
-    $EXISTS "$LE_PYTHON" > /dev/null && break
-  done
+  # Arguments: "NOCRASH" if we shouldn't crash if we don't find a good python
+  if [ -n "$USE_PYTHON_3" ]; then
```
erikrose: While this is technically correct, we might dodge a future surprise if we use `[ -n "$USE_PYTHON_3" != "1" ]` instead, so somebody doesn't come along later, set USE_PYTHON_3=0, and expect the obvious thing to happen.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -296,11 +312,24 @@ elif [ -f /etc/mageia-release ]; then
...
+  prev_le_python="$LE_PYTHON"
+  unset LE_PYTHON
+  DeterminePythonVersion "NOCRASH"
+  if [ "$PYVER" -eq 26 ]; then
+    Bootstrap() {
+      BootstrapMessage "RedHat-based OSes that will use Python3"
+      BootstrapRpmPython3
+    }
+    USE_PYTHON_3=1
+    BOOTSTRAP_VERSION="BootstrapRpmPython3 $BOOTSTRAP_RPM_PYTHON3_VERSION"
```
erikrose: I think our assumption before was than a given box would never change the bootstrapper it's using. Let me try to construct a scenario that would break.... No, I think we're okay. If the bootstrapper OR the bootstrap version changes, we simply rebootstrap entirely. We never implemented the "differential" bootstrapping, where we install, say, only the new packages that are required.

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
```diff
@@ -0,0 +1,78 @@
...
+    if ! $TOOL list epel-release >/dev/null 2>&1; then
+      error "Enable the EPEL repository and try running Certbot again."
+      exit 1
+    fi
+    if [ "$ASSUME_YES" = 1 ]; then
+      /bin/echo -n "Enabling the EPEL repository in 3 seconds..."
+      sleep 1s
+      /bin/echo -ne "\e[0K\rEnabling the EPEL repository in 2 seconds..."
+      sleep 1s
+      /bin/echo -e "\e[0K\rEnabling the EPEL repository in 1 seconds..."
```
erikrose: If we're going to repeat ourselves, we might as well say "1 second". :-)

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -112,6 +120,14 @@ def verified_new_le_auto(get, tag, temp_dir):
                             "certbot-auto.", exc)
 
 
+def create_CERT_NONE_context():
```
erikrose: For consistency with the noun-like names of other functions that have return values, this would be better named "cert_none_context".

letsencrypt-auto-source/tests/auto_test.py
```diff
@@ -202,6 +202,7 @@ def run_le_auto(le_auto_path, venv_dir, base_url, **kwargs):
 Z5o/NDiQB11m91yNB0MmPYY9QSbnOA9j7IaaC97AwRLuwXY+/R2ablTcxurWou68
 iQIDAQAB
 -----END PUBLIC KEY-----""",
+             NO_CERT_VERIFY='1',
```
erikrose: I'm curious: what changed to make this needed?

letsencrypt-auto-source/pieces/fetch.py
```diff
@@ -66,21 +74,21 @@ def get(self, url):
...
         environ.get('LE_AUTO_JSON_URL',
-                    'https://pypi.python.org/pypi/certbot/json')))
+                    'https://pypi.python.org/pypi/certbot/json')).decode('UTF-8'))
     # metadata['info']['version'] actually returns the latest of any kind of
     # release release, contrary to https://wiki.python.org/moin/PyPIJSON.
     # The regex is a sufficient regex for picking out prereleases for most
     # packages, LE included.
     return str(max(LooseVersion(r) for r
-                   in metadata['releases'].iterkeys()
+                   in iter(metadata['releases'].keys())
```
erikrose: I think iter() is pretty much a no-op in this situation.

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -0,0 +1,65 @@
...
+# ensure python2.7 is available
+python2.7 --version 2> /dev/null
+RESULT=$?
+if [ $RESULT -ne 0 ]; then
+  error "Python3 is not available."
+  exit 1
+fi
+
+# bootstrap, but don't install python 3.
+certbot/letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade -n > /dev/null 2> /dev/null
```
erikrose: I recoil from this a bit. It seems we're splitting the tests (for CentOS only) across the Dockerfile and the Python tests in auto_test.py. None of the other Dockerfiles make assertions or emit pass/fail messages. Unless there's a strong motivation I'm missing, I urge us to incorporate these into the Python-driven tests. If this is very awkward to spell in Python, we should at least kick off a shell script from auto_test.

erikrose:
A couple of cleanliness things and 2 or 3 things of more import. No huge showstoppers. That's all I have to say about that! :-) Thanks for your hard work on a twisty maze of shell commands, @ohemorange!

letsencrypt-auto-source/letsencrypt-auto.template
ohemorange: Adding a comment is a good idea. Not reverting PYVER isn't a problem.

ohemorange: (I can explain more if you want; it's related to #5388)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
ohemorange: I would like to state for posterity that the line was there before (this part was just moved to a new name) and I am not personally responsible for this grammatical heresy.

letsencrypt-auto-source/tests/auto_test.py
ohemorange: We run the tests using a different version of python in some cases. At some point Python changed the default behavior from "don't bother actually checking if a cert is valid" to "actually do". But we need it to not do that for the test.

letsencrypt-auto-source/pieces/fetch.py
ohemorange: I thought it was needed for Python3, but I'll double check that.

letsencrypt-auto-source/tests/centos6_tests.sh
ohemorange: The Python tests are run on all the systems that we test. We don't want these tests to be run on anything other than CentOS. We could add a flag to the Python tests to run certain tests, but given that these tests are mostly about installing packages then rerunning certbot-auto, it's more straightforward to just do it this way. The Python test would essentially be calling out to the shell to install things, which seems even weirder.

letsencrypt-auto-source/letsencrypt-auto.template
erikrose: As long as you have the PYVER reasoning clear in your head, that's good enough for me. :-)

letsencrypt-auto-source/pieces/bootstrappers/rpm_common_base.sh
erikrose: Ha! Okay, you are absolved.

letsencrypt-auto-source/tests/auto_test.py
erikrose: Innnnteresting. I'm still not sure if we're on the same page. All that stuff under tests/certs/ is the gymnastics I did to ensure that the certs would in fact be valid. And I *think* I was using 2.7.9 or higher when I wrote it; that's the version when the cert validation started happening. Where were we having validation errors?

letsencrypt-auto-source/tests/centos6_tests.sh
erikrose: > We don't want these tests to be run on anything other than CentOS.

Ah, that's the strong motivation I was missing. Thanks! In that case, what I could use is a few comments: one demarcating the division between the Dockerfile tests and the Python ones saying pretty much what you said above, and another explaining how Python 2.7 disappears by line 42. (Does le-auto remove a repo or something, and it goes *poof*?)

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ $RESULT -eq 0 ]; then
+  error "Python3 is already installed."
+  exit 1
+fi
+
+# ensure python2.7 is available
+python2.7 --version 2> /dev/null
+RESULT=$?
+if [ $RESULT -ne 0 ]; then
+  error "Python3 is not available."
```
erikrose: Do you mean Python 2.7?

PR #5334, opened 2017-12-17
Interfaces for TLS updates and plugin specific updates

joohoi:
This PR is based on the design proposal by @bmw 

Three new interfaces: `ServerTLSUpdater`, `GenericUpdater` and `RenewDeployer` are being introduced.

## ServerTLSUpdater
`ServerTLSUpdater` allows updating the TLS configuration using the selected installer plugin. This provides way for us to keep the user configuration secure if and when new vulnerabilities in TLS configuration are being discovered.

Implementing `ServerTLSUpdater` involves installer providing method:
```
    def server_tls_updates(self, domain, *args, **kwargs):
        """Set the server's TLS config to latest recommended version.

        This function will only be called if the user hasn't disabled TLS
        server updates.

        :param str domain: domain in the lineage being updated

        """
```
The installer will also need to register the interface by using method
`interfaces.ServerTLSUpdater.register(InstallerPluginClass)`  in the end of the module of installer class.

The method `server_tls_updates` will be called on every time Certbot is run with `renew` verb, and for each lineage will be called once for every included domain.

## GenericUpdater
`GenericUpdater` provides the functionality needed to deploy security enhancements in dynamic manner. This is provided for installer specific updates, that do not fall into category of security enhancements that will be added later as their own interfaces.

Implementing `GenericUpdater` involves installer providing the following method:
```
    def generic_updates(self, domain, *args, **kwargs):
        """Perform any update types defined by the installer.

        If an installer is a subclass of the class containing this method, this
        function will always be called when "certbot renew" is run. If the
        update defined by the installer should be run conditionally, the
        installer needs to handle checking the conditions itself.

        This method is called once for each domain.

        :param str domain: domain in the lineage being updated

        """
```
The installer will also need to register the interface by using method
`interfaces.GenericUpdater.register(InstallerPluginClass)` in the end of the module of installer class.

The method `generic_updates` works exactly as the `server_tls_updates` for `ServerTLSUpdater`, described above.

## RenewDeployer
`RenewDeployer` provides the functionality needed when a lineage is renewed. This interface method will be called if the installer used for the lineage implements it.

Implementing `RenewDeployer` involves installer providing the following method:
```
    def renew_deploy(self, domain, lineage, *args, **kwargs):
        """Perform any update types defined by the installer.

        If an installer is a subclass of the class containing this method, this
        function will always be called when "certbot renew" is run. If the
        update defined by the installer should be run conditionally, the
        installer needs to handle checking the conditions itself.

        This method is called once for each lineage renewed

        :param lineage: Certificate lineage object that is set if certificate
            was renewed on this run.
        :type lineage: storage.RenewableCert

        """
```
The installer will also need to register the interface by using method
`interfaces.RenewDeployer.register(InstallerPluginClass)` in the end of the module of installer class.


bmw:
#2208 is turning out to be a nightmare with us finding more and more problems that need to be solved, so I only had time to write down high level comments here before the holiday break. This is long and loosely formatted so I apologize and try to bear with me.

I think you should look over my explanation of why I chose this approach that I emailed you and if you think we should go with something else, let me know what you prefer and why. Given that we'll be living with this interface for a while, I think it's worth designing it as carefully as we can.

One of my main goals when adding to the plugin interfaces is to make it easy to implement as we can. I think the reason hooks exist is because writing Certbot plugins is too hard and requires too much code. I bring this up now because I think this motivation affects a number of my other comments here.

As for specific things in this PR, what do you think the benefit of `verify_installer_specific_updates` is? The main example you mentioned in your original post is `--disable-installer-updates`. I'm not sure if we should have this flag. My main purpose when adding this interface was to provide something that third party plugins could use to provide enhancements without blocking on changes landing in Certbot's core. Adding this flag introduces the problem you described and forces the plugins to write code checking if this flag and any of their installer specific updates flags were set. I also think the notion of an installer specific update is largely internal and not clear to most users.

The only other reason I could come up with for this method was to allow plugins to use something like `IDisplay` to ask questions about the enhancements selected early in execution. Perhaps this is valuable, however, a plugin may be able to do this during a function like `prepare()`. I really haven't thought about doing this much, but as this was the only other reason I could come up with. If there are things I am missing, let me know.

Also, I feel like with the approach here, every "updater" method we ever create will take a `lineage` parameter to signal whether or not the cert was renewed. This feels clunky. If certificate renewal is something we should signal to the plugin, perhaps we should do so generally outside of this interface. The way I'd do this is yet another simple abstract class that defines a method that is called on renewal. Using this approach, plugins that implement enhancements that want to do something special during renewal can do so and we don't have to duplicate this parameter across all of our updater methods.

As you can tell, I quite the fan of these simple ABCs. We can provide meaningful implementations of them in `certbot.plugins.common` and plugins can pick and choose the functionality they want/need and ignore (and not implement) those they don't.

A more minor point here, both you and Erica were confused by my class names and documentation here. If you like this general approach, but think this can be improved, please do so. I think we should have a common name like "Updater" to group these related classes together, but otherwise, I'm pretty open to the approach here.

Again, sorry for this ramble and let me know what you think!

joohoi:
> I think you should look over my explanation of why I chose this approach that I emailed you and if you think we should go with something else, let me know what you prefer and why. Given that we'll be living with this interface for a while, I think it's worth designing it as carefully as we can.

I agree with that design, and I believe this current basis for the functionality implements exactly that, albeit the "global" enhancement specific interface classes are still missing.

> As for specific things in this PR, what do you think the benefit of verify_installer_specific_updates is? The main example you mentioned in your original post is --disable-installer-updates. I'm not sure if we should have this flag. My main purpose when adding this interface was to provide something that third party plugins could use to provide enhancements without blocking on changes landing in Certbot's core. Adding this flag introduces the problem you described and forces the plugins to write code checking if this flag and any of their installer specific updates flags were set. I also think the notion of an installer specific update is largely internal and not clear to most users.

The `--disable-installer-updates` was used just as an example. I think the flag is needed in general, even though it perhaps shouldn't be used in a way that I used in the example. I see this necessary for user to be able to force `certbot renew` to skip the additional installer magic in case they specifically only want the certificate renewed or there's some kind of unforseen or temporary problem in an custom enhancement a installer provides.

> Also, I feel like with the approach here, every "updater" method we ever create will take a lineage parameter to signal whether or not the cert was renewed. This feels clunky. If certificate renewal is something we should signal to the plugin, perhaps we should do so generally outside of this interface. The way I'd do this is yet another simple abstract class that defines a method that is called on renewal. Using this approach, plugins that implement enhancements that want to do something special during renewal can do so and we don't have to duplicate this parameter across all of our updater methods.

I like this idea, and will create a new updater interface for this!

joohoi:
The original description is now updated.

certbot/interfaces.py
```diff
@@ -591,3 +599,97 @@ def add_message(self, msg, priority, on_crash=True):
...
+        """
+
+
+class ServerTLSUpdater(object):
+    """Interface for updating a server's TLS configuration.
+
+    An installer that wants to perform TLS server updates according to this
+    interface must not only be a subclass but must respect the disablement of
+    server side TLS updates if done so by the user when performing other
+    installer functions. For example, if server TLS updates are disabled and
```
ohemorange: As in, must not modify it elsewhere? We may want to restate this as "only this method can be used to modify server tls configurations", because we're disabling the entire method outside of it; they don't have to check the flag themselves.

certbot/interfaces.py
```diff
@@ -591,3 +599,97 @@ def add_message(self, msg, priority, on_crash=True):
...
+
+    This class allows plugins to perform types of updates that need to run at
+    lineage renewal that Certbot hasn't defined (yet).
+
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def renew_deploy(self, lineage, *args, **kwargs):
```
ohemorange: We should give an example here to explain when this should be used -- "for example, if a plugin wants to copy over a certificate to a new location only when it is renewed".

certbot/interfaces.py
```diff
@@ -591,3 +599,97 @@ def add_message(self, msg, priority, on_crash=True):
...
+        This method is called once for each lineage renewed
+
+        :param lineage: Certificate lineage object that is set if certificate
+            was renewed on this run.
+        :type lineage: storage.RenewableCert
+
+        """
+
+
+class ServerTLSUpdater(object):
```
ohemorange: I think the key word here isn't "server" or "TLS" but "configuration".

certbot/cli.py
```diff
@@ -1173,6 +1183,13 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         default=flag_default("directory_hooks"), dest="directory_hooks",
         help="Disable running executables found in Certbot's hook directories"
         " during renewal. (default: False)")
+    helpful.add(
+        "renew", "--disable-renew-updates", action="store_true",
+        default=flag_default("disable_renew_updates"), dest="disable_renew_updates",
+        help="Disable automatic updates to your server configuration that"
+        " would otherwise be done by the selected installer plugin, and"
+        " triggered by the \"renew\" verb. This setting does not apply to"
```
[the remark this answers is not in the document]
joohoi: It actually is the former. Literally triggered by the renew verb. Do you think we should be more verbose here?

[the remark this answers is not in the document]
ohemorange: Yeah, can't hurt to be more explicit.

certbot/interfaces.py
joohoi: I do feel that configuration itself would be a too broad term. And this is intended to be able to handle important changes to TLS configuration in case of vulnerability or broken cipher or so.

certbot/interfaces.py
joohoi: Only if the `--disable-server-tls-updates` is defined on the command line or in the configuration file. This interface method is intended to be run on `renew` verb, so it shouldn't be expanded for use in every TLS configuration change.

If this is defined in the command line by the user, an installer implementing this shouldn't add our TLS configuration include to the `VirtualHost` / `server` block, or configure additional security enhancements.

It acts as a general "Don't worry Certbot, I got this" - handle for the TLS configuration from behalf of the user.

certbot/cli.py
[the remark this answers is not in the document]
ohemorange: The updates to the documentation say the opposite of this -- that we run 'when a certficate has been renewed by running "certbot renew"'. If it's the former, then in fact we run "when the user executes `certbot renew`, regardless of if the cert is renewed".

certbot/interfaces.py
ohemorange: Is `ServerTLSConfigurationUpdater` too long?

certbot/interfaces.py
ohemorange: Is what you're trying to say that `--disable-server-tls-updates` should also apply to `deploy_cert`? I don't think that's what we want. And even if it was, that comment should go there, not here.

If `--disable-server-tls-updates` is `True`, will this method be run at all? We should put the answer to that in this comment explicitly.

certbot/interfaces.py
joohoi: It really isn't, but it approaches the length that starts to give me Java-shivers. It however is much more informative, and now in use :)

certbot/cli.py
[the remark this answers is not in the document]
joohoi: This is now fixed.

certbot/interfaces.py
joohoi: Yes, that was the original plan, to give user a possibility to tell Certbot that they will handle the TLS configuration (and updates to it) manually. We can discuss about this functionality further. I didn't yet change the behavior.

Currently if `--disable-server-tls-updates` (now changed to `--disable-tls-configuration-updates`) is `True`, the updater interface method doesn't get run at all, but `deploy_cert` will. The implementing Installer should keep configuring the certificate paths, but for the more detailed TLS configuration should check this CLI parameter value, and act accordingly.

certbot/interfaces.py
joohoi: Explanation added.

certbot/interfaces.py
ohemorange: I can't for the life of me find the original plan anywhere.

It seems like since we're talking about "updates" here, that's a separate thing from `deploy_cert`, which is only called initially.

bmw: So my initial proposal was worded like this. It was in an email called "TLS Updates Proposal", let me know if you'd like me to send it again.

I think my main concerns were making Certbot useful to more people and to make it so those who disable our TLS updates for renewal (keeping in mind there are valid reasons to do so) don't suddenly have an update forced upon them through normal Certbot usage.

If a sysadmin wants to deploy a site with a TLS config they know shouldn't be modified, but they want to use a Certbot installer plugin for automated restarts on renewal, security enhancements like slowly increasing HSTS age and OCSP prefetching, etc. forcing them to take our TLS config first and revert it back seems unnecessary. This would make automating that sites deployment very hard without a deep understanding of the changes Certbot's going to make which I wouldn't expect them to have or consider stable and unchanging at this point. I also think adding/removing domains isn't an uncommon part of Certbot usage. If you have disabled TLS security updates and need to do this, do you have to manually revert pieces of your config in this case too?

In theory, there could be a separate flag for this, but it's hard for me to imagine when you'd want one without the other and it causes problems when people forget one of the flags.

joohoi:
Brought this branch up to speed with master, and resolved conflicts.

certbot/interfaces.py
ohemorange: Alternatively, they might want to manually check changes that we make. So, they'd use certbot to install the cert because they'll go check it, but wouldn't want us changing the around their config automatically thereafter. But at the same time they wouldn't want to not have us deploy it in the first place, because otherwise they have to go figure out where the certs live and what the best way to set all that up is.

Also, thunderbird seems to have eaten your original email; I only have the replies to it somehow.

bmw: With this approach they could still do that. They could not provide `--disable-server-tls-updates` and later add it with `certbot enhance` or to a configuration file. The same isn't true if it cannot be disabled outside of `certbot renew`.

My initial email is just a git patch that doesn't include much more than what's documented in this PR, but I've sent that to you again. If you're missing any of the follow up discussion and would like to read that, let me know.

ohemorange: Here are my thoughts:

This entire thing is a complicated use case that only expert Certbot users could set up successfully in the first place.

Given that, we should make the mental model as consistent as possible. It is weird that something called update applies to something that is not an update.

If they want to add/remove domains to a cert, they can run `certonly`. For any use case, they could individually run any command with or without the flag and add or remove it later.

bmw: I want to get this and the other PRs @joohoi has been building off of merged so he's not blocked. His other work should not depend on the `ServerTLSConfigurationUpdater` class so let's land a PR based on this with all of that code removed.

We should keep the `ServerTLSConfigurationUpdater` code around somewhere in a branch though so it can be reused and maybe modified because I think getting this Updater landed is important. Without it, Apache and Nginx only update the user's TLS config when a cert is actually renewed and Postfix is unable to do so at all.

@joohoi, can you do this surgery? I'd also be interested to hear your thoughts on the discussion here, but that can wait until later if you want.

joohoi:
I created a new PR #5879 from a new branch forked from this one, only this time without the `  `. This was done in order to keep this branch and the discussion for archive for possible later use.

certbot/interfaces.py
joohoi: I think that while being somewhat complicated use case, it's something that the regular Certbot users would never even stumble upon of, and the installers would just do their thing in keeping the configuration secure. 

That said, even without this interface, it's still possible for installers to handle the task using the `GenericUpdater` and `RenewDeployer`, it just can't be explicitly allowed while disabling possible other tasks that the installer might be updating.

For now, I created a new PR that is forked from this one, the only difference being the removal of the `ServerTLSConfigurationUpdater` and its functionality. The PR is #5879

ohemorange:
You probably meant to push the changes to the branch in the other PR.

joohoi:
I'm going to close this PR, as fork of this, #5879 was merged. This branch should be kept if we want to return to the `interfaces.ServerTLSConfigurationUpdater` interface later on, as this was removed from the forked branch.

Issue #5337, opened 2017-12-18
Nginx plugin crashes when non-ascii configuration file read


g6123:
## My operating system is (include version):

Arch Linux ARM 4.9.69-1-ARCH

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

OS package manager (`pacman -S certbot-nginx`)

## I ran this command and it produced this output:

I ran this command:

`sudo certbot --nginx`

and it produced this:

```
Saving debug log to /var/log/letsencrypt/letsencrypt.log
An unexpected error occurred:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 10453: ordinal not in range(128)
Please see the logfiles in /var/log/letsencrypt for more details.
```

## Certbot's behavior differed from what I expected because:

I found that the error comes from `/etc/nginx/mime.types` which was included in the default nginx package. If there is any non-ascii encoded file in the nginx configuration path, the same error occurs.

## Here is a Certbot log showing the issue:

<details>
<summary>Click to expand</summary>

```
2017-12-18 20:21:34,224:DEBUG:certbot.main:certbot version: 0.20.0
2017-12-18 20:21:34,235:DEBUG:certbot.main:Arguments: ['--nginx']
2017-12-18 20:21:34,243:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#manual,PluginEntryPoint#nginx,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
2017-12-18 20:21:34,417:DEBUG:certbot.log:Root logging level set at 20
2017-12-18 20:21:34,427:INFO:certbot.log:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2017-12-18 20:21:34,434:DEBUG:certbot.plugins.selection:Requested authenticator nginx and installer nginx
2017-12-18 20:21:34,742:DEBUG:certbot.log:Exiting abnormally:
Traceback (most recent call last):
  File "/usr/bin/certbot", line 11, in <module>
    load_entry_point('certbot==0.20.0', 'console_scripts', 'certbot')()
  File "/usr/lib/python3.6/site-packages/certbot/main.py", line 861, in main
    return config.func(config, plugins)
  File "/usr/lib/python3.6/site-packages/certbot/main.py", line 685, in run
    installer, authenticator = plug_sel.choose_configurator_plugins(config, plugins, "run")
[...]

PR #5341, opened 2017-12-20
Use UTF-8 encoding for nginx plugin

g6123:
Updated nginx plugin parser to always use UTF-8 encoding with Python ([3.x](https://docs.python.org/3/library/codecs.html), [2.x](https://docs.python.org/2/library/codecs.html)) standard library `codecs`.

This fixes the issue #5337.

### DIFF
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -1,5 +1,6 @@
 """NginxParser is a member object of the NginxConfigurator class."""
 import copy
+import codecs
 import functools
 import glob
 import logging
@@ -202,7 +203,7 @@ class NginxParser(object):
             if item in self.parsed and not override:
                 continue
             try:
-                with open(item) as _file:
+                with codecs.open(item, "r", "utf-8") as _file:
                     parsed = nginxparser.load(_file)
                     self.parsed[item] = parsed
                     trees.append(parsed)
@@ -378,7 +379,7 @@ class NginxParser(object):
 def _parse_ssl_options(ssl_options):
     if ssl_options is not None:
         try:
-            with open(ssl_options) as _file:
+            with codecs.open(ssl_options, "r", "utf-8") as _file:
                 return nginxparser.load(_file)
         except IOError:
             logger.warn("Missing NGINX TLS options file: %s", ssl_options)


greencopper:
Is there a manual workaround? I have tried keeping the Nginx config file as small as possible, but still get the error on Arch Linux.

g6123:
@greencopper It is not really clean way, though you can manually apply the changes to the package. The path would be `/usr/lib/python3.6/site-packages/certbot_nginx/parser.py` or so.

GLaDTheresCake:
I was running into this problem for weeks, got the workaround thrown at me to find unicode characters in files. Used the command ```grep -r -P '[^\x00-\x7f]' /etc/nginx /etc/letsencrypt``` to see where the unicode characters might be, turns out theres a unicode character in the mime-types file (³ to be exact). So remove the  line with that character on (by finding it with the above command, don't remember what it was exactly) and you should be golden.

bmw:
@ohemorange, would you like to review this instead?

deprito:
Just installed via instruction

But, get the same error...


2018-06-10 21:03:05,538:DEBUG:certbot.error_handler:Calling registered functions
2018-06-10 21:03:05,538:INFO:certbot.auth_handler:Cleaning up challenges
2018-06-10 21:03:06,688:DEBUG:certbot.log:Exiting abnormally:
  File "/opt/eff.org/certbot/venv/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/main.py", line 1323, in main
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/main.py", line 1086, in run
    certname, lineage)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/main.py", line 120, in _get_and_save_cert
    lineage = le_client.obtain_and_enroll_certificate(domains, certname)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/client.py", line 383, in obtain_and_enroll_certificate
    cert, chain, key, _ = self.obtain_certificate(domains)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/client.py", line 326, in obtain_certificate
    orderr = self._get_order_and_authorizations(csr.data, self.config.allow_subset_of_names)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/client.py", line 362, in _get_order_and_authorizations
    authzr = self.auth_handler.handle_authorizations(orderr, best_effort)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/auth_handler.py", line 75, in handle_authorizations
    resp = self._solve_challenges(aauthzrs)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot/auth_handler.py", line 126, in _solve_challenges
    resp = self.auth.perform(all_achalls)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/configurator.py", line 1045, in perform
    http_response = http_doer.perform()
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/http_01.py", line 62, in perform
    self.configurator.save("HTTP Challenge", True)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/configurator.py", line 975, in save
    self.parser.filedump(ext='')
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/parser.py", line 243, in filedump
    out = nginxparser.dumps(tree)
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/nginxparser.py", line 134, in dumps
    return str(RawNginxDumper(blocks.spaced))
  File "/opt/eff.org/certbot/venv/local/lib/python2.7/site-packages/certbot_nginx/nginxparser.py", line 98, in __str__
    return ''.join(self)
**UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 34: ordinal not in range(128)**
2018-06-10 21:03:06,690:ERROR:certbot.log:An unexpected error occurred:


=========

How to upgrade the certbot to 0.26.0?

schoen:
I'm a bit nervous about the possibility that some people's nginx files might also contain binary data or textual data in an encoding *other than* UTF-8, in which case the Python UTF-8 codec might still crash when trying to read these files.

For example, using Markus Kuhn's stress test file at

https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt

if we say

```
>>> import codecs
>>> f = codecs.open("UTF-8-test.txt", "r", "utf-8")`
>>> x = f.read()
```

we will get a crash because, indeed, the file *isn't* valid UTF-8.

We might want to say that nginx configuration files must be valid UTF-8 in order for Certbot to parse them (which seems like a perfectly valid kind of requirement), but in that case we'd probably still want to give a more useful error in this context, maybe like telling the user `nginx configuration file {} is not a valid UTF-8  text file and couldn't be parsed` or something?

I could imagine that there might be people using other character sets in comments. For instance, suppose a Chinese speaker uses BIG5 encoding in comments. In that case the line

`# 共`

would be encoded with bytes [35, 32, 166, 64, 10]. Unfortunately this is not only not UTF-8 but also has no interpretation as UTF-8.

`>>> (b"# \xa6\x40\n").decode("UTF-8")`

ends up giving

`UnicodeDecodeError: 'utf8' codec can't decode byte 0xa6 in position 2: invalid start byte`

:cry:

schoen:
@g6123, thank you for doing this work. Do you think you could revise this to catch errors when the file contains non-UTF-8 data? I think we can assume that we can reject such files (rather than adding further complexity about detection of encodings) but we will probably want to give a useful error message to the user in such cases.

I will also ask my colleagues to confirm that they agree with this path.

ohemorange:
@schoen does nginx itself crash on that test file? we just want to match nginx's behavior, not catch everything

schoen:
@ohemorange, sad to say, I've just created an nginx configuration file with a comment containing a Chinese character in BIG5 encoding instead of UTF-8, and nginx happily parsed and accepted this file. But Python3 refuses to read the content of that same file into a UTF-8 string and instead crashes with a `UnicodeDecodeError` exception.

So I think we do need to deal with this case after all, or else at least some users may continue to get Unicode-related errors even with the improvement made by this PR!

schoen:
So, we've concluded that we don't necessarily need to match nginx's behavior but we should at least give a friendly error message rather than an error traceback when a non-UTF-8-valid string is encountered in a configuration file. After all, nginx *does* accept these strings at least inside comments (and maybe they can also be served in web headers or something).

The error message would explain that Certbot is only able to modify nginx configuration files that are valid UTF-8, and indicate which particular file failed to be parsed.

Maybe we could also open a wishlist issue to ask for more closely matching nginx's behavior in this regard (especially if nginx turns out to accept other string encodings *outside* of comment lines).

@g6123, are you still interested in working on this issue?

Zopieux:
Fully agree with @schoen analysis. We cannot reasonably create a fully compatible nginx conf parser+writer in Python, this would be too much work to handle all the corner cases of the original nginx parser in C. A pragmatic fallback though, is to fail gracefully with a useful explanation for the user. Having a line number would be a neat addition.

g6123:
I also agree with @schoen.

@ohemorange I'm working on it, but it would take some time, since I'm not actively developing Python programs. I added commits that displays warning message and returns empty result with some naive tests.

ohemorange:
Ok, be sure to let us know if there's anything we could do to help move it along, including questions or thoughts or anything.

ohemorange:
@g6123, @Zopieux do you think you'll work on this more? If not we might take it over so it gets in.

g6123:
@ohemorange okay, I guess I'm not. You can take it over.

Zopieux:
Sorry if I let you folks under the wrong impression but it wasn't my intention to contribute on this. Hope this can be fixed someday though!

Issue #5263, opened 2017-11-27
Add Apache plugin support for Gentoo with systemd


joohoi:
The current default constants for Gentoo fail on systems that are using systemd as init system.

This is because control script in Apache package expects OpenRC based init scripts. It might be feasible to handle the httpd restarts and configuration dumps using the systemd service.

@Jarodiv you have previously been active on this front, so maybe you have some insight on the following points:
 - What would be the sensible way to fingerprint the systemd installation? Just finding out if systemctl exists?
 - What does the systemd service for Apache in Gentoo look like? Does it use the values from `/etc/conf.d/apache2` variables in command line like it does on OpenRC systems?
 - Do we have a way to run `configtest` on the httpd with the correct command line?

cc @osirisinferi as you have been active on Gentoo front also.

Jarodiv:
Wow, I totally missed out this one.

> What would be the sensible way to fingerprint the systemd installation? Just finding out if systemctl exists?
Systemd, if used, always has the PID `1`, so the easiest way is to simply check if the process `systemd` is running and if it has the PID `1`.

> What does the systemd service for Apache in Gentoo look like? Does it use the values from /etc/conf.d/apache2 variables in command line like it does on OpenRC systems?
Nope, it sadly does not. I wasn't able to locate the file neither in the Gentoo repos nor in the Apache Git repo but here is how it looks on my host (vanilla):
```
[Unit]
Description=The Apache HTTP Server
After=network.target remote-fs.target nss-lookup.target

[Service]
EnvironmentFile=/etc/conf.d/apache2
ExecStart=/usr/sbin/apache2 $APACHE2_OPTS -DFOREGROUND
ExecReload=/usr/sbin/apache2 $APACHE2_OPTS -k graceful
ExecStop=/usr/sbin/apache2 $APACHE2_OPTS -k graceful-stop
# We want systemd to give httpd some time to finish gracefully, but still want
# it to kill httpd after TimeoutStopSec if something went wrong during the
# graceful stop. Normally, Systemd sends SIGTERM signal right after the
# ExecStop, which would kill httpd. We are sending useless SIGCONT here to give
# httpd time to finish.
KillSignal=SIGCONT
PrivateTmp=true
#Hardening
PrivateTmp=true
CapabilityBoundingSet=CAP_CHOWN CAP_SETGID CAP_SETUID CAP_DAC_OVERRIDE CAP_KILL CAP_NET_BIND_SERVICE CAP_IPC_LOCK
[...]

PR #5349, opened 2017-12-22
Use apache2ctl modules for Gentoo systems.

### LINKED ISSUE #5344: Can't install new certificates with certbot-apache under Gentoo since version 0.20.0
It seems that version 0.20.0 has a major architecture revision from 0.19.0, in particular regarding OS specific overrides. But the Gentoo overrides have a flaw when trying to install a new certificate : at one point it tries to run "apache2ctl -t -D DUMP_MODULES", which is no good at all under Gentoo because apache2ctl is a dummy script redirecting to /etc/init.d/apache2. It generates this error :

`Error in checking parameter list: /etc/init.d/apache2: invalid option -- 't'`

Gentoo

OS package manager (emerge) with testing (~amd64) flag

```
certbot -d test.example.com
Plugins selected: Authenticator apache, Installer apache
Obtaining a new certificate
tls-sni-01 challenge for test.example.com
Waiting for verification...
Cleaning up challenges
Error in checking parameter list: /etc/init.d/apache2: invalid option -- 't'

Apache is unable to check whether or not the module is loaded because Apache is misconfigured.
```

Certbot did not install the certificate because it failed to run apache2ctl -t -D DUMP_MODULES

joohoi:
Fixes #5344

### DIFF
--- a/certbot-apache/certbot_apache/override_gentoo.py
+++ b/certbot-apache/certbot_apache/override_gentoo.py
@@ -49,6 +49,7 @@ class GentooParser(parser.ApacheParser):
     def update_runtime_variables(self):
         """ Override for update_runtime_variables for custom parsing """
         self.parse_sysconfig_var()
+        self.update_modules()
 
     def parse_sysconfig_var(self):
         """ Parses Apache CLI options from Gentoo configuration file """
@@ -58,6 +59,8 @@ class GentooParser(parser.ApacheParser):
             self.variables[k] = defines[k]
 
     def update_modules(self):
-        """ NOOP for update_modules as Gentoo does not provide a way to dump
-        module list from Apache binary"""
-        pass
+        """Get loaded modules from httpd process, and add them to DOM"""
+        mod_cmd = [self.configurator.constant("apache_cmd"), "modules"]
+        matches = self.parse_from_subprocess(mod_cmd, r"(.*)_module")
+        for mod in matches:
+            self.add_mod(mod.strip())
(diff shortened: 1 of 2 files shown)


bmw:
This seems simple enough but two questions:

1. What's the behavior of `apache2ctl` on systems with systemd? 
2. Could we use `apache2ctl modules` on Gentoo? I'm not sure how new this functionality is, but at least on an up to date system, this works. The relevant snippet from `/etc/init.d/apache2` is:
```
modules() {
        checkconfig || return 1
        ${APACHE2} ${APACHE2_OPTS} -M 2>&1
}
```

joohoi:
Nice catch! Now this PR is doing exactly that, using `apache2ctl modules`, for the modules call. 

Unfortunately I think that systemd version is completely unable to use apache2ctl, as it pretty much calls the OpenRC script directly, the whole contents of `apache2ctl` being:
```
#!/bin/sh
exec /etc/init.d/apache2 "$@"                          
```

And this has also been the main issue for Gentoo systems with systemd. I'm trying to figure out the correct things to do for these systems in #5263

bmw:
Then my only concern with this PR is will trying to use `apache2ctl modules` make things worse for systemd users?

I tried to test this myself, but after waiting an hour for things to compile, I'm not sure I did it right. OpenRC was still installed and `/etc/init.d/apache2` still works. After Googling around, maybe it OpenRC is always installed unless the user does a lot of work on their end to rip it out of Gentoo. See https://forums.gentoo.org/viewtopic-t-994660.html?sid=2d4170a3eda7addd8e2883e402af733d.

Since I struggled to make this work, I'm trusting your judgement here.

joohoi:
We don't currently have a reasonable support for Gentoo systems running systemd at all. Even though systemd might be officially supported by Gentoo, the Apache package unfortunately isn't. This is discussed in #3411 . 

We need to clean up the constants throughout Apache plugin at some point, and this will also include making the user able to provide the Apache control commands on the command line which in turn will help to alleviate this problem.

PR #5362, opened 2018-01-04
Dovecot configuration parser

TheNavigat:
This is a work in progress PR for an installer plugin for Dovecot.

### DIFF
--- a/certbot-dovecot/certbot_dovecot/dovecotparser.py
+++ b/certbot-dovecot/certbot_dovecot/dovecotparser.py
@@ -1,7 +1,6 @@
 """Very low-level dovecot config parser based on pyparsing."""
 from pyparsing import (
-    printables, restOfLine,
-    Literal, Forward, Group, LineEnd, Optional,
+    restOfLine, Literal, Forward, Group, LineEnd, Optional,
     OneOrMore, ParserElement, White, Word, ZeroOrMore
 )
 
@@ -15,17 +14,11 @@ class DovecotParser(object):
     # configuration itself
     ParserElement.setDefaultWhitespaceChars('')
 
-    # This is printables excluding the following characters: # { } ,
+    # This is printables excluding the following characters: # { } , =
     allowed_characters = (
         "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!\"$%&"
-        "\\'()*+-./:;<=>?@[\\]^_`|~"
+        "\\'()*+-./:;<>?@[\\]^_`|~"
     )
-
-    allowed_characters = printables.replace("=", "")
-    allowed_characters = allowed_characters.replace("{", "")
-    allowed_characters = allowed_characters.replace("}", "")
-    allowed_characters = allowed_characters.replace(",", "")
-    allowed_characters = allowed_characters.replace("#", "")
 
     newline = LineEnd()
     same_line_space = Optional(Word(' \t').leaveWhitespace())
--- a/certbot-dovecot/certbot_dovecot/dovecotparser.py
+++ b/certbot-dovecot/certbot_dovecot/dovecotparser.py
@@ -1,9 +1,8 @@
 """Very low-level dovecot config parser based on pyparsing."""
-
 from pyparsing import (
     alphanums, restOfLine,
-    Literal, Dict, Forward, Group, LineEnd, Optional,
-    OneOrMore, ParserElement, Word, ZeroOrMore
+    Literal, Forward, Group, LineEnd, Optional,
+    OneOrMore, ParserElement, White, Word, ZeroOrMore
 )
 
 class DovecotParser(object):
@@ -14,38 +13,49 @@ class DovecotParser(object):
     # Here, we're setting default whitespace characters to ` ` and `\t`,
     # excluding `\n` since a new line is used as a delimeter for dovecot
     # configuration itself
-    ParserElement.setDefaultWhitespaceChars(' \t')
+    ParserElement.setDefaultWhitespaceChars('')
 
-    newline = LineEnd().suppress()
+    newline = LineEnd()
+    same_line_space = Optional(Word(' \t').leaveWhitespace())
+    space = Optional(White().leaveWhitespace())
 
     # Defining comments
-    comment = Group(Literal('#') + restOfLine)
-    comment_line = comment + OneOrMore(newline)
+    comment = space + Literal('#') + restOfLine
 
     # Defining key and value pair
-    key = Word(alphanums) + (Literal('=').suppress())
+    key = Word(alphanums) + same_line_space + (Literal('='))
 
-    value_group = (OneOrMore(Optional(Literal(',')).suppress()
-                   + Word(alphanums)) + Optional(comment))
+    value_group = OneOrMore(
+        same_line_space
+        + Optional(Literal(',') + same_line_space)
+        + Word(alphanums)
+    )
 
-    key_value_pair = Dict(Group(key + value_group)) + newline
+    key_value_pair = (
+        space + key + value_group + same_line_space
+    )
 
     # Defining includes
     # Order matters. It's a first item match.
-    include = Dict(Group(
-        (Literal('!include_try') | Literal('!include'))
-        + OneOrMore(Word(alphanums + "_/-?!")) + newline
-    ))
+    include = (
+        space
+        + (Literal('!include_try') | Literal('!include'))
+        + OneOrMore(same_line_space
+        + Word(alphanums + "_/-?!"))
+        + same_line_space
+    )
 
     # Defining a block
     block = Forward()
 
-    item = comment_line | block | key_value_pair | include
+    item = Group(comment | block | key_value_pair | include) + newline.suppress()
 
-    block_title = Word(alphanums)
+    block_title = Group(
+        space + Word(alphanums) + same_line_space
+        + Literal('{') + same_line_space + newline.suppress()
+    )
 
-    left_brace = Literal('{').suppress() + Optional(comment)
-    right_brace = Literal('}').suppress() + Optional(comment)
+    right_brace = Group(space + Literal('}') + same_line_space + Optional(comment))
 
     # The use of the OneOrMore(newline) here is actually not an assumption.
     # Dovecot requires every { to be followed by a new line, and every } to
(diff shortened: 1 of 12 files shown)


TheNavigat:
This PR is for the dovecot configuration parser. It will be followed by another PR that introduces dumping the parsed list into a string.

sydneyli:
Hey, thanks for working on this! Just tracking the PR.

TheNavigat:
@sydneyli Thanks! :) If you have some context, can you please guide me a bit through this? (Not current test failure, this is easy to fix, but am I doing this correctly? Did I miss something?)

sydneyli:
@TheNavigat Oops, I missed the comment you made earlier! Yeah sure, I have some context on the Dovecot plugin and can take a quick high-level look at this soon. For some background: What's the difference between this PR and #5518 ? For now, both look like Dovecot config parsing utilities?

TheNavigat:
@sydneyli I have split the work into multiple PRs to ease reviewing. This PR includes only the parser. The other one is built on this, and added to it is the dumper code.

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,102 @@
...
+    # excluding `\n` since a new line is used as a delimeter for dovecot
+    # configuration itself
+    ParserElement.setDefaultWhitespaceChars('')
+
+    newline = LineEnd()
+    same_line_space = Optional(Word(' \t').leaveWhitespace())
+    space = Optional(White().leaveWhitespace())
+
+    # Defining comments
+    comment = space + Literal('#') + restOfLine
```
sydneyli: Note that comments can come at the end of any line (they don't have to be isolated on their own line). ([reference](https://github.com/dovecot/core/blob/master/src/config/config-parser.c#L593))

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Defining a block
+    block = Forward()
+
+    item = Group(comment | block | key_value_pair | include) + newline.suppress()
+
+    block_title = Group(
+        space + Word(alphanums) + same_line_space
+        + Literal('{') + same_line_space + newline.suppress()
+    )
```
sydneyli: Sections (or blocks) can be defined/named by two space-delimited tokens ([reference](https://wiki.dovecot.org/ConfigFile#Sections))

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Defining includes
+    # Order matters. It's a first item match.
+    include = (
+        space
+        + (Literal('!include_try') | Literal('!include'))
+        + OneOrMore(same_line_space
+        + Word(alphanums + "_/-?!"))
+        + same_line_space
```
sydneyli: `include` statements seem to only accept a single token-- the filepath. ([reference1](https://github.com/dovecot/core/blob/master/src/config/config-parser.c#L961), [reference2](https://wiki.dovecot.org/ConfigFile#Including_config_files)) Although I may have missed a use case-- so I'm not 100% sure.

certbot-dovecot/certbot_dovecot/tests/dovecotparser_test.py
```diff
@@ -0,0 +1,158 @@
...
+class DovecotParserTest(unittest.TestCase):
+    def setUp(self):
+        self.parser_instance = DovecotParser()
+
+    def test_key_value(self):
+        string = "x = 1 a b c,d ,e, f\n"
+        parsed = self.parser_instance.parse_string(string).asList()
+
+        result = [list(string[0:-1])]
+        self.assertEqual(parsed, result)
```
sydneyli: One thing we should be careful about is preserving whitespace, so we don't accidentally strip things from existing config files. For instance, here, `x = y` should parse differently than `x = y\n`, and `x = y` should parse differently from `x=y`.

From a first glance, most cases should be ok. I'd recommend adding more tests to make this behavior explicit.

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,102 @@
...
+
+    newline = LineEnd()
+    same_line_space = Optional(Word(' \t').leaveWhitespace())
+    space = Optional(White().leaveWhitespace())
+
+    # Defining comments
+    comment = space + Literal('#') + restOfLine
+
+    # Defining key and value pair
+    key = Word(alphanums) + same_line_space + (Literal('='))
```
sydneyli: Be careful with using `alphanums`-- it doesn't look like [Dovecot's parser](https://github.com/dovecot/core/blob/master/src/config/config-parser.c) restricts tokens to use alphanumeric characters.

TheNavigat: From what I see here, https://github.com/dovecot/core/blob/master/src/config/config-parser.c#L572, it basically accepts all characters except whitespace. Thanks for that note!

certbot-dovecot/certbot_dovecot/dovecotparser.py
TheNavigat: Oops, my bad. I totally missed that.

TheNavigat:
@sydneyli Thanks a lot for the review, that was a very helpful one :)

I updated the code accordingly. Can you please give it another shot reviewing this?

sydneyli:
@TheNavigat So, there's a lot of not-nice things about the nginx parser and it would be unfortunate if the Dovecot plugin inherited all of those problems. @ohemorange and I will probably be refactoring the nginx parsing code soon, so we can get rid of UnspacedList (which I noticed you are using in #5751).

Thanks for your work on this so far-- I'll try to review and merge in the parsing code from this PR, but hold off on your other PRs until we clean up the nginx parser code. (And since the deadline for this project is coming up, I may start building on top of your PRs as well!)

TheNavigat:
Sounds good :)

Out of curiosity and willingness to commit further into this, what do you and @ohemorange have in mind to replace `UnspacedList`? A dictionary like structure?

sydneyli:
@TheNavigat I wrote down some brief notes here: #5823 
and should hopefully have some WIP stuff up in a branch tomorrow, if you're curious! The gist of it is that we'll be explicitly deserializing the parsed tokens into immutable object trees, instead of making implicit assumptions about the structure/format of lists of lists of lists.

Let's get this PR merged-- the pyparsing stuff is v helpful! Sorry for the delay in reviews-- looking over it now.

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,115 @@
...
+from pyparsing import (
+    restOfLine, Literal, Forward, Group, LineEnd, Optional,
+    OneOrMore, ParserElement, White, Word, ZeroOrMore
+)
+
+class DovecotParser(object):
+    # pylint: disable=expression-not-assigned
+    """A class that parses dovecot configuration with pyparsing."""
+
+    # pyparsing automatically uses whitespaces as delimeters for tokens.
```
sydneyli: nit: s/delimeters/delimiters

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,115 @@
...
+    # followed by a new line. This should maintain that  } is also always
+    # preceded by a new line.
+    block_body = (Group(
+                        ZeroOrMore(item)
+                  ) + right_brace)
+
+    # Blocks contain of a list containing 3 lists. The first item is the
+    # block name, while the second item is a list of the block contents,
+    # and the third is the right brace and spaces surrounding it
+    block << OneOrMore(block_title + block_body)
```
sydneyli: I think I'd still prefer that you parse blocks into a structure similar to the nginx parsing code ( `[ "block_name", [ "block", "contents" ] ]` rather than `[["block_name", " ", "{"], ["block", "contents"], ["}"]]`. This is mostly because we want to share as much code as possible between the nginx and dovecot parsers!

certbot-dovecot/certbot_dovecot/dovecotparser.py
```diff
@@ -0,0 +1,115 @@
...
+    ParserElement.setDefaultWhitespaceChars('')
+
+    # This is printables excluding the following characters: # { } , =
+    allowed_characters = (
+        "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!\"$%&"
+        "\\'()*+-./:;<>?@[\\]^_`|~"
+    )
+
+    newline = LineEnd()
+    same_line_space = Optional(Word(' \t').leaveWhitespace())
```
sydneyli: nit: does `White(ws=' \t')` work?

certbot-dovecot/certbot_dovecot/dovecotparser.py
TheNavigat: If we do this, how will we be able to preserve spacing around the braces?

In other words, how will we be able to differentiate between `block name { ` and `block name {`, and `} ` and `}`?

sydneyli: From your current implementation, you might have to preserve newlines. For instance, here's how our nginx parser parses an oddly spaced block (`#` used to denote EOL)
```
 block  {   #
    subblock   statement#
  }  #
```
into:
```
[
[ [' ', 'block', '  '],                 # block name, + spacing before left brace
  [ '   \n   ', 'subblock', '   ', 'statement' ] # spacing after left brace
  '\n  '                          # spacing before right brace
],
'  '                   # spacing after right brace
]
```

TheNavigat: Are you sure about this?

I understand how we want parsing to be as similar as possible, but the delimiters differ. Dovecot configuration delimiter is a new line.

It will require some special case handling to parse and dump blocks correctly if we don't include the new line following the braces with the block content itself, requiring us to also add the spacing preceding the `\n`.

If you're sure about this I'll give it a go.

sydneyli: That's reasonable! I think we can make the difference in list-structures work nicely with the new parsing refactor.

TheNavigat: So, should we merge this now? :)

sydneyli:
@TheNavigat This looks ok to me. Adding @ohemorange for some final `pyparsing` review! We'll try to get this reviewed and landed during the next sprint.

TheNavigat:
@sydneyli @ohemorange Ping :)

TheNavigat:
@sydneyli @ohemorange Another ping :(

sydneyli:
Closing for the time being; see [comment here](https://github.com/certbot/certbot/issues/6009#issuecomment-431535887)!

PR #5371, opened 2018-01-05
Have letsencrypt-auto do a real upgrade in leauto-upgrades

### LINKED ISSUE #5370: fetch.py accepts HTTP URLs?
I only played with this briefly, but [this script](https://github.com/certbot/certbot/files/1605310/test_leauto_upgrades.txt) passes as a test farm test when it probably shouldn't. It's possible there's just a bug in my test, but I wanted to make a note about this for us to look into.

bmw:
Rather than replacing the file with `git`, let's have `letsencrypt-auto` download a new version of itself from a dummy server.

This PR takes advantage of #5370, but:

1. The version we're upgrading from is as old as possible so even if we fix #5370 in `master`, this will still work.
2. To use HTTPS with Python 2.7.9+ we need something like `NO_CERT_VERIFY` from #5329 which obviously isn't available in older versions of `certbot-auto`.

I tested this successfully with both `master` and the #5329 branch.

### DIFF
--- a/tests/letstest/scripts/test_leauto_upgrades.sh
+++ b/tests/letstest/scripts/test_leauto_upgrades.sh
@@ -35,7 +35,7 @@ SERVER_PID=$!
 trap 'kill "$SERVER_PID" && rm -rf "$MY_TEMP_DIR"' EXIT
 cd ~-
 
-# Next, we set up the files to be served.
+# Then, we set up the files to be served.
 FAKE_VERSION_NUM="99.99.99"
 echo "{\"releases\": {\"$FAKE_VERSION_NUM\": null}}" > "$MY_TEMP_DIR/json"
 LE_AUTO_SOURCE_DIR="$MY_TEMP_DIR/v$FAKE_VERSION_NUM"
@@ -45,7 +45,7 @@ cp letsencrypt-auto-source/letsencrypt-auto "$LE_AUTO_SOURCE_DIR/letsencrypt-aut
 SIGNING_KEY="letsencrypt-auto-source/tests/signing.key"
 openssl dgst -sha256 -sign "$SIGNING_KEY" -out "$NEW_LE_AUTO_PATH.sig" "$NEW_LE_AUTO_PATH"
 
-# Finally, we set the necessary certbot-auto environment variables.
+# Next, we set the necessary certbot-auto environment variables.
 export LE_AUTO_DIR_TEMPLATE="http://localhost:$SERVER_PORT/%s/"
 export LE_AUTO_JSON_URL="http://localhost:$SERVER_PORT/json"
 export LE_AUTO_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----
@@ -58,6 +58,9 @@ Z5o/NDiQB11m91yNB0MmPYY9QSbnOA9j7IaaC97AwRLuwXY+/R2ablTcxurWou68
 iQIDAQAB
 -----END PUBLIC KEY-----
 "
+
+# Finally, we wait for the server to start.
+sleep 5s
 
 if ! ./letsencrypt-auto -v --debug --version || ! diff letsencrypt-auto letsencrypt-auto-source/letsencrypt-auto ; then
     echo upgrade appeared to fail
--- a/tests/letstest/scripts/test_leauto_upgrades.sh
+++ b/tests/letstest/scripts/test_leauto_upgrades.sh
@@ -59,9 +59,7 @@ iQIDAQAB
 -----END PUBLIC KEY-----
 "
 
-EXPECTED_VERSION=$(grep -m1 LE_AUTO_VERSION certbot-auto | cut -d\" -f2)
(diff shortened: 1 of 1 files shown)


tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -15,19 +15,54 @@ if ! command -v git ; then
...
+git checkout -f v0.5.0 letsencrypt-auto
 if ! ./letsencrypt-auto -v --debug --version --no-self-upgrade 2>&1 | grep 0.5.0 ; then
     echo initial installation appeared to fail
     exit 1
 fi
 
-git checkout -f "$BRANCH"
-EXPECTED_VERSION=$(grep -m1 LE_AUTO_VERSION letsencrypt-auto | cut -d\" -f2)
-if ! ./letsencrypt-auto -v --debug --version --no-self-upgrade 2>&1 | grep $EXPECTED_VERSION ; then
+# Now that python and openssl have been installed, we can set up a fake server
```
ohemorange: Noting that this might break eventually when future systems stop supporting older pythons.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -15,19 +15,54 @@ if ! command -v git ; then
...
+# to provide a new version of letsencrypt-auto. First, we start the server and
+# directory to be served.
+MY_TEMP_DIR=$(mktemp -d)
+SERVER_PORT=12345
```
ohemorange: Should do something like https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/tests/auto_test.py#L84 if we don't want this test to be flaky.

ohemorange: Or, if SimpleHTTPServer can get a port for you and tell you what it is.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -15,19 +15,54 @@ if ! command -v git ; then
...
+LsIVPBuy9IcgHidUQ96hJnoPsDCWsHwX62495QKEarauyKQrJzFes0EY95orDM47
+Z5o/NDiQB11m91yNB0MmPYY9QSbnOA9j7IaaC97AwRLuwXY+/R2ablTcxurWou68
+iQIDAQAB
+-----END PUBLIC KEY-----
+"
+
+# Finally, we wait for the server to start.
+sleep 5s
+
+if ! ./letsencrypt-auto -v --debug --version || ! diff letsencrypt-auto letsencrypt-auto-source/letsencrypt-auto ; then
```
ohemorange: `! diff` --> they are different. We want them to be different, so this should just be `diff`.

tests/letstest/scripts/test_leauto_upgrades.sh
bmw: I don't understand. What might break?

ohemorange: oh, if we just use whatever old python is installed by the old version.

tests/letstest/scripts/test_leauto_upgrades.sh
bmw: If you tell `SimpleHTTPServer` to use port 0, the OS gives it a random free port. I had a version that did this but encountered some issues where Python wasn't flushing the output stream when I had it redirected somewhere, causing it to never print the port it chose.

I can either add a simple recreation of this server to `tools/` that manually flushes the output stream or we can pick a random, unprivileged port like 12345 (or another number) and assume/hope there won't be any collisions on our simple EC2 instances. Preference?

tests/letstest/scripts/test_leauto_upgrades.sh
bmw: `if ! diff a b` branch is run if files `a` and `b` differ. Since we just upgraded `letsencrypt-auto` in the repo root to the version in `letsencrypt-auto-source`, we expect these files to be identical.

tests/letstest/scripts/test_leauto_upgrades.sh
ohemorange: Any reason not to do what https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/tests/auto_test.py#L84 does? Pick a range of ports, and try ports in that range until it works.

bmw: I think there's an awkward race condition. We start `SimpleHTTPServer` in the backgrpound and since it's running in parallel, we're just have to pick an arbitrary amount of time to wait and see if it's going to fail. Also, I'm not aware of a good way to check the exit status of a background process (without blocking on the process exiting).

bmw: I put a version that has the OS pick the port in #5390. Pick whichever you like better or let me know if there's another approach you prefer that I'm missing.

bmw:
Closing in favor of #5390.

PR #5372, opened 2018-01-06
Remove Default Detector log line.

jsha:
This produces a super-long log line that wraps to 30-60 lines, depending on
screen width. Even though it's just at debug level, it clutters up the integration
test output without providing proportional debugging value.


jsha:
Pushed an update. BTW, splitting these out makes it clear that the set_by_cli checks are actually very slow. If you run with high verbosity you can watch them tick by about one per second. With the set of flags in `certbot_test` (from `_common.sh`), it takes about 10 seconds to evaluate for all the flags.

PR #5374, opened 2018-01-06
ACMEv2: Add Order support

jsha:
This pull request is built off of v2-client. I'd like to get some feedback about the API design while v2-client is still in review.

This adds two new classes in messages: Order and OrderResource. It also adds methods to ClientV2 to create orders, and poll orders then request issuance.

The CSR is stored on the OrderResource so it can be carried along and submitted when it's time to finalize the order.


acme/acme/messages.py
```diff
@@ -172,8 +172,9 @@ class Directory(jose.JSONDeSerializable):
     class Meta(jose.JSONObjectWithFields):
         """Directory Meta."""
         terms_of_service = jose.Field('terms-of-service', omitempty=True)
+        terms_of_service_v2 = jose.Field('termsOfService', omitempty=True)
         website = jose.Field('website', omitempty=True)
-        caa_identities = jose.Field('caa-identities', omitempty=True)
+        caa_identities = jose.Field('caaIdentities', omitempty=True)
```
ohemorange: Is this what it is in v2? We presumably still need the v1 version around, no?

acme/acme/messages.py
```diff
@@ -251,7 +252,7 @@ class Registration(ResourceBody):
     contact = jose.Field('contact', omitempty=True, default=())
     agreement = jose.Field('agreement', omitempty=True)
     status = jose.Field('status', omitempty=True)
-    terms_of_service_agreed = jose.Field('terms-of-service-agreed', omitempty=True)
+    terms_of_service_agreed = jose.Field('termsOfServiceAgreed', omitempty=True)
```
ohemorange: same here -- do we not need the v1 version?

acme/acme/messages.py
jsha: This is what it is in v2. Boulder doesn't set this field in either v1 or v2, so we can be confident no one is relying on it and we can change it.

acme/acme/messages.py
jsha: Neither `terms-of-service-agreed` nor `termsOfServiceAgreed` is in v1 (it uses `agreement` instead).

ohemorange: ah I see now, this is updating something that was added in #5287.

acme/acme/messages.py
ohemorange: Are you planning to ever set it for v1 servers?

jsha: Nope.

acme/acme/client.py
```diff
@@ -549,14 +550,82 @@ def new_account(self, new_account):
...
         :rtype: `.RegistrationResource`
-
         """
         response = self.net.post(self.directory['newAccount'], new_account,
             acme_version=2)
         # "Instance of 'Field' has no key/contact member" bug:
         # pylint: disable=no-member
         return self._regr_from_response(response)
 
+    def new_order(self, csr_pem):
```
bmw: Any reason to require the CSR now rather than allowing the user to provide it later? All we need now is the list of domains.

jsha: An earlier ACME draft specified the CSR up front rather than at the end. We've changed the spec, but I figured I'd leave the CSR argument at the beginning of the process, with the idea that that gives us more flexibility under the hood. I figured changing the flow to take the CSR at the end would have been sort of overfitting the current protocol rather than abstracting from it.

bmw: > An earlier ACME draft specified the CSR up front rather than at the end.

Yeah I actually chimed in on the acme list in favor of that approach, but I don't think it will matter much for us in practice.

> We've changed the spec, but I figured I'd leave the CSR argument at the beginning of the process, with the idea that that gives us more flexibility under the hood. I figured changing the flow to take the CSR at the end would have been sort of overfitting the current protocol rather than abstracting from it.

What flexibility does this buy us? I might be missing something you thought about. To be honest, I don't have a strong preference here, but I think only requiring what you actually need simplifies the function and makes it more flexible for the user.

acme/acme/client.py
```diff
@@ -549,14 +550,82 @@ def new_account(self, new_account):
...
+            if certificate_response.ok:
+                fullchain_pem = certificate_response.text
+        return messages.OrderResource(
+            body=body,
+            uri=response.headers.get('Location', uri),
+            fullchain_pem=fullchain_pem,
+            authorizations=authorizations,
+            csr_pem=csr_pem)
+
+    def poll_order_and_request_issuance(self, orderr, max_time=datetime.timedelta(seconds=90)):
```
bmw: We can keep this method if its useful for others, but we actually don't use the `poll_and_request_issuance` method in the current ACMEv1 client in Certbot. Other code handles making sure the authzs are all valid so we could get by with method that just posts to the finalize URL and polls that and avoid polling all authzs again.

acme/acme/messages.py
```diff
@@ -483,3 +484,37 @@ class Revocation(jose.JSONObjectWithFields):
...
+class OrderResource(ResourceWithURI):
+    """Order Resource.
+
+    :ivar acme.messages.Order body:
+
+    """
+    body = jose.Field('body', decoder=Order.from_json)
+    csr_pem = jose.Field('csr_pem', omitempty=True)
+    authorizations = jose.Field('authorizations')
+    fullchain_pem = jose.Field('fullchain_pem', omitempty=True)
```
bmw: Do all of these attributes make sense on an `OrderResource` when compared to the other *Resources in this file?

acme/acme/client.py
jsha: Talked about this offline; I think my preference for up-front was a bit stronger than Brad's preference for not, so keeping it this way.

acme/acme/messages.py
jsha: Yes? I'm not sure how to evaluate that.

bmw: Sorry. I just made this as a quick comment without diving into it since it was getting late on Friday night, but I'm not sure if it does match the other resources.

I certainly could be off base as the vast majority of this file hasn't changed since before I was involved in the project, but my understanding of all of the classes here is for each object in the ACME protocol you have a class that defines that object and a `ObjectResource` class that contains the object as the `body` and the other fields are the various `link` headers provided at that endpoint.

The only exception to this which I couldn't figure out is `CertificateResource` where the inclusion of `authzs` doesn't make sense to me, although looking through the git history, it was added when you the ACME spec said to send the CSR and list of authz to the `new-cert` endpoint.

bmw: We talked about this offline and I'm fine with the changes here.

bmw:
Jacob/Erica let me know if I should review this.

Similarly, Jacob we can use all the help we can get, but obviously getting this landed is a high priority for us. If someone should take this from you and build off it, let us know.

bmw:
This changes landed in #5578 so I'm closing this PR.

PR #5378, opened 2018-01-06
Fix macOS builds for Python2.7 in Travis

joohoi:
macOS image in travis was running Python2.7 from homebrew but for some reason it wasn't linked correctly regardless of the `brew link python` output saying that it already was.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,12 +56,6 @@ matrix:
       sudo: required
     - python: "2.7"
       env: TOXENV=nginxroundtrip
-    - language: generic
-      env: TOXENV=py27
-      os: osx
-    - language: generic
-      env: TOXENV=py36
-      os: osx
 
 
 # Only build pushes to the master branch, PRs, and branches beginning with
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -56,6 +56,12 @@ matrix:
       sudo: required
     - python: "2.7"
       env: TOXENV=nginxroundtrip
+    - language: generic
```
bmw: I created #5381 to document our tests in the `test-everything` branch, but I don't think want to readd these tests here. Since Travis macOS tests take so long to run, we've been doing this and other testing in the `test-everything` branch which is run nightly.

We should just merge the fix above and once that has landed, merge the fix into `test-everything`.

joohoi:
> Thanks a lot for fixing this. Obviously this works, but I'm curious how do you know that Python 2.7 is installed through Homebrew in the macOS image?

Through black box testing ;) The process can be seen here #5377 
Perhaps we could add `python` to the argument list of `brew install` to make sure it stays that way though. I'll add that.

joohoi:
This should be fixed now. Seems like homebrew returns a non-zero exit value when one of the packages listed in `brew install` is already installed. The complete test suite was run in commit HEAD^1 and the macOS tests were removed in the last commit to avoid re-adding them to master.

Issue #5388, opened 2018-01-08
Properly handling user set LE_PYTHONs for RHEL/CentOS 6


bmw:
LE_PYTHON is always set during phase 2 because this variable is exported. Even worse, at this point we can't differentiate between the case where this value was set by the user and one where it was set by a previous version of the script. This is a problem because we won't know which Python executable to use to run Certbot and whether an unsupported value was manually set by the user or not. As the script is right now, this can cause us to create the venv using Python 2.6 or continue to use Python 2.6 to run fetch.py.

A couple potential fixes are:

Fix 1:
* Before we ever set --cb-auto-has-root or --le-auto-phase2, if LE_PYTHON is set, complain about deprecation and tell people to use a new var like CB_PYTHON.
* When bootstrapping for Python 3, overwrite/unset LE_PYTHON.

Fix 2:
* Quietly overwrite/unset LE_PYTHON

The properties I would like are:
 * If in the future, when someone manually sets LE_PYTHON to Python 2.6, we tell them their python is too old like we do for other versions.
* If we can't tell whether the value was user set or not, we ignore the value. There is a tradeoff to be made between printing a warning about this and just doing it quietly. By printing a warning, how many users will get a warning/cron email when most of them haven't set LE_PYTHON as it isn't formally documented anyway.

ohemorange:
Fix 3:

- Stop exporting LE_PYTHON
- Right before calling phase 2, export a version number.
- If in phase 2 this version number is unset, we came from a previous version; override/ignore LE_PYTHON
- If the version number is 0.21.0+, respect LE_PYTHON.

bmw:
Concern: I think we should try and limit the scope of where we do this. Depending on how this is implemented, this could cause `LE_PYTHON` to always be overridden/ignored on every system when upgrading to a newer version. Really, we just want to stop `certbot-auto` from keeping people on older versions of Python.

Can we limit this to values of `LE_PYTHON` we don't support?

ohemorange:
When I said "override/ignore LE_PYTHON", I specifically, meant "for python3". That should limit us to this particular case.

bmw:
Great! This sounds good to me then.

I thought of one additional minor concern that could influence the implementation here. Imagine the following:

1. CentOS 6 user runs certbot-auto < 0.21.0.
2. It upgrades to certbot-auto >= 0.21.0 and runs it with `--le-auto-phase`. `LE_PYTHON` is set to Python 2.6 and the version number environment variable is not set.
[...]

PR #5389, opened 2018-01-08
Handle need to rebootstrap before fetch.py

### LINKED ISSUE #5387: Dropping support for Pythons in fetch.py
If we drop support for a version of Python previously supported by fetch.py, existing certbot-auto installations can fail. If we're unable to reboostrap because we don't have the user's permission, when certbot-auto is run again during phase 1 it will call `DeterminePythonVersion` and if we can't find an appropriate Python, we exit with an error. Instead, we should print a message about being unable to check for upgrades.

bmw:
Fixes #5387.

The
```
# bootstrap, this time installing python3
 certbot/letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade -n > /dev/null 2> /dev/null
```
lines in `centos6_tests.sh` after the test I added may or may not be necessary depending on how #5388 is resolved, but in the worst case `letsencrypt-auto` should effectively be a noop and will just run Certbot.

### DIFF
--- a/letsencrypt-auto-source/tests/centos6_tests.sh
+++ b/letsencrypt-auto-source/tests/centos6_tests.sh
@@ -4,6 +4,8 @@ yum update > /dev/null
 yum install -y centos-release-scl > /dev/null
 yum install -y python27 > /dev/null 2> /dev/null
 
+LE_AUTO="certbot/letsencrypt-auto-source/letsencrypt-auto"
+
 # we're going to modify env variables, so do this in a subshell
 (
 source /opt/rh/python27/enable
@@ -25,7 +27,7 @@ if [ $RESULT -ne 0 ]; then
 fi
 
 # bootstrap, but don't install python 3.
-certbot/letsencrypt-auto-source/letsencrypt-auto --no-self-upgrade -n > /dev/null 2> /dev/null
+"$LE_AUTO" --no-self-upgrade -n > /dev/null 2> /dev/null
 
 # ensure python 3 isn't installed
 python3 --version 2> /dev/null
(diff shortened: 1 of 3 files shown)


letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -47,6 +47,12 @@ if [ $RESULT -eq 0 ]; then
   exit 1
 fi
 
+# Skip self upgrade due to LE_PYTHON value
+if ! LE_PYTHON=nonexistant certbot/letsencrypt-auto-source/letsencrypt-auto 2>&1 | grep -q "WARNING: couldn't find Python"; then
```
ohemorange: nonexistent

ohemorange: unless you're worried nonexistent exists

bmw: The purpose of this was to set `LE_PYTHON` to a nonexistent value, however, the empty string is also a good nonexistent value so I removed setting the variable.

bmw: I read this on my phone last night, didn't notice the spelling difference, and thought you had a problem with this approach, however, I think testing with the real (empty) `LE_PYTHON` value is probably better than testing with a silly value like this so I'm going to leave this PR as is.

ohemorange: Good with either test for now; if something in #5388 I'll fix the test there. Do we not care about line length in leauto?

PR #5391, opened 2018-01-09
Add deprecation warning for Python 2.6

### LINKED ISSUE #5373: Add a message saying we plan to deprecate python2.6
Similarly to how we did it in #5283.

Notes:
- the biggest difference is PendingDeprecationWarning is shown by default in py 2.6 and isn't in py 3.3
- also, while I didn't worry about this for python 3.3 since there were like 7 users, it'd be nice if we printed the warnings in a way that were able to be disabled by the user
- (as for this, I think it would be nice and may be doable (you can set up logging to capture warnings if we wanted to be consistent in where they were sent), but if it's much effort don't worry about it
- we're talking about the LE_PYTHON people and people installing from pip)
- so if someone upgrades, they can see the warning but continue to use it in cron without getting email every 12 hours when certbot renew runs
- probably the warning system is the best way to do this
- it shouldn't be much code, but might take a little research

ohemorange:
Fixes #5373.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -1217,9 +1217,10 @@ def main(cli_args=sys.argv[1:]):
         # Let plugins_cmd be run as un-privileged user.
         if config.func != plugins_cmd:
             raise
-    if sys.version_info[:2] == (3, 3):
-        logger.warning("Python 3.3 support will be dropped in the next release "
-                    "of Certbot - please upgrade your Python version.")
+    for (major, minor) in [(2, 6), (3, 3)]:
+        if sys.version_info[:2] == (major, minor):
+            logger.warning("Python {0}.{1} support will be dropped in the next release "
+                        "of Certbot - please upgrade your Python version.")
 
     set_displayer(config)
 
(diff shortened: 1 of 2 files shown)


certbot/main.py
```diff
@@ -1217,9 +1218,17 @@ def main(cli_args=sys.argv[1:]):
...
-                    "of Certbot - please upgrade your Python version.")
+    deprecation_fmt = (
+        "Python %s.%s support will be dropped in the next "
+        "release of Certbot - please upgrade your Python version.")
+    # We use the warnings system for Python 2.6 and logging for Python 3
+    # because DeprecationWarnings are only reported by default in Python <= 2.6
+    # and warnings can be disabled by the user.
+    if sys.version_info[:2] == (2, 6):
+        warning = deprecation_fmt % sys.version_info[:2]
+        warnings.warn(warning, DeprecationWarning)
```
ohemorange: This should be a PendingDeprecationWarning, unless that didn't used to exist.

Issue #4947, opened 2017-07-19
Developer Guide should include instructions on writing/running tests for certbot-auto


void-elf:
They should go at the bottom of this section: https://certbot.eff.org/docs/contributing.html#updating-certbot-auto-and-letsencrypt-auto

bmw:
I assigned this to myself to make sure Erica had what she needed to work on #2208. I talked to her in person and we came up with a plan for testing that issue.

Personally, I'd like to see testing for `certbot-auto` redone. There are a couple approaches to testing certbot-auto (`letsencrypt-auto-source/tests` and test farm tests) and both are complicated with a number of gotchas. Because of this, tests for `certbot-auto` are often not written.

Rather than trying to explain the complexity of the current approach, I'd rather eliminate it by making certbot-auto tests easier to write and then documenting that. I'm going to try and make this happen. If we don't end up redoing these tests soon, documenting what we currently have should still be done, but personally I'm going to focus on simplifying what we currently have first.

bmw:
Here's a `certbot-auto` testing brain dump. If I forget anything please add it to this issue or if anyone working on this stuff has any questions, let me know. Some of the pain points I describe below should be taken into account when redesigning these tests.

There are two kinds of `certbot-auto` tests. First are the tests in `letsencrypt-auto-source/tests/auto_test.py` and the other are test farm tests in `tests/letstest/scripts/`. Both of these methods test `certbot-auto` as a whole rather than testing individual pieces.

#### letsencrypt-auto-source/tests/auto_test.py

This modifies pieces of `certbot-auto` to make testing easier. It uses a dummy PyPI server, GitHub server, signatures, and has `certbot-auto` install a simple Python package that spits out a version number rather than the real version of Certbot and all of its Python dependencies. All OS dependencies are still installed (if expected by the test).

For each test, a new directory is used for `certbot-auto`'s installation, but any installed OS packages persist between tests. We have Dockerfiles to run these tests on different systems in `letsencrypt-auto-source`. Feel free to add new ones to the `master` branch and then configure it to be run in the `.travis.yml` in `test-everything` branch as appropriate.

#### tests/letstest/scripts/

All of our "test farm" tests are found here but the ones specifically for testing `certbot-auto` that we currently run are `test_leauto_upgrades.sh` and `test_letsencrypt_auto_certonly_standalone.sh`. Feel free to modify these (as long as you're not deleting existing valuable tests) or add new ones.

[...]

PR #5392, opened 2018-01-09
Fix rebootstraping with old venvs

bmw:
If the user has a venv from before we moved the location to `/opt`, `certbot-auto` doesn't properly detect we're rebootstrapping and thinks the script is being run for the first time. This fixes that.

The tests here rely on #5389 also being merged.

If I requested a review from you but you don't feel comfortable reviewing it, just let me know so I can make sure I get a review from someone else in time for the release.

EDIT: I explained this PR in more detail https://github.com/certbot/certbot/pull/5392#issuecomment-356503127 if it helps the reviewer.


tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -64,6 +64,38 @@ iQIDAQAB
 -----END PUBLIC KEY-----
 "
 
+if [ $(python -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//') -eq 26 ]; then
```
ohemorange: Note for posterity that I don't actually want to see changed: this test is only correct in practice, because we only actually upgrade to python3 when we're on redhat based systems, whereas this test checks on any system.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -64,6 +64,38 @@ iQIDAQAB
...
+    fi
+    if ! command -v python3; then
+        echo "Python3 wasn't properly installed"
+        exit 1
+    fi
+    if [ "$(/opt/eff.org/certbot/venv/bin/python -V 2>&1 | cut -d" " -f 2 | cut -d. -f1)" != 3 ]; then
+        echo "Python3 wasn't used in venv!"
+        exit 1
+    fi
+elif ! ./letsencrypt-auto -v --debug --version || ! diff letsencrypt-auto letsencrypt-auto-source/letsencrypt-auto ; then
```
ohemorange: Are we intentionally running this test twice if we're not using python2.6?

bmw:
@erikrose, if you have time to give this a (hopefully 2nd) review early tomorrow morning, that'd be really useful. Most people don't feel comfortable reviewing this and I think we need it for the release.

tests/letstest/scripts/test_leauto_upgrades.sh
bmw: Nope.

bmw:
This PR is needed due to a bug I introduced with the combination of #4970 and #4971. Despite being written by me at similar times, those PRs were written separately and this patch is needed when the PRs are combined together.

The need for this can be seen by playing with the test written here on CentOS 6. You can run a test farm test, but you can also just run it manually by running `yum install epel-release -y`, cloning this repo to a directory named `letsencrypt`, checking out the code you want to test, and running this script from the directory containing the `letsencrypt` directory.

Without these changes to `letsencrypt-auto`, [this test](https://github.com/certbot/certbot/pull/5392/files#diff-83d32e155b116db8dfad1d6177025c9eR73) will fail. What happens on this line is `cb-auto` upgrades from 0.5.0 to the version in `letsencrypt-auto-source` and "phase 2" of that script runs. (This is accomplished by the old version of the script providing `--le-auto-phase2` on the command line.) When phase 2 runs, we check if there are new OS dependencies that need to be installed before installing the latest version of Certbot. We determine this by looking at files in the existing Certbot installation to see what version of our dependencies we installed.

The problem that occurs is we only check the new venv location in `/opt` and not the old location in `~/.local/share/letsencrypt`. (In #4971 I added this check here but in #4970  I added the need to check a second location.) As a result, we incorrectly think we just installed OS dependencies as this is what happens at the beginning of "phase 1" if there is no existing installation and try to install Certbot. This installation fails because we try to use Python 3 which hasn't been installed yet and the tests exit with an error saying we shouldn't have tried to upgrade. Instead, we should have printed a message about needing to install new packages to upgrade and continue running the version of Certbot installed in the existing venv.

Hopefully this helps the reviewer. Feel free to ping me however makes you happy if you have any questions.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -898,13 +898,18 @@ if [ "$1" = "--le-auto-phase2" ]; then
...
   INSTALLED_VERSION="none"
-  if [ -d "$VENV_PATH" ]; then
+  if [ -d "$VENV_PATH" -o -d "$OLD_VENV_PATH" ]; then
     # If the selected Bootstrap function isn't a noop and it differs from the
     # previously used version
     if [ -n "$BOOTSTRAP_VERSION" -a "$BOOTSTRAP_VERSION" != "$PREV_BOOTSTRAP_VERSION" ]; then
       # if non-interactive mode or stdin and stdout are connected to a terminal
       if [ \( "$NONINTERACTIVE" = 1 \) -o \( \( -t 0 \) -a \( -t 1 \) \) ]; then
-        rm -rf "$VENV_PATH"
+        if [ -d "$VENV_PATH" ]; then
```
joohoi: If, for some reason both, the `$VENV_PATH` and `$OLD_VENV_PATH` exist here, we're never able to get to a sane state.

bmw: While this case is pretty unlikely, there are a few ways it could happen so it's definitely worth thinking about. Nice catch.

My understanding of what currently happens in this case is we hit this line, see that `$VENV_PATH` exists, delete it, and then rerun the script. In this next run, we hit "phase 1", see that `$OLD_VENV_PATH` exists [here](https://github.com/certbot/certbot/blob/v0.20.0/certbot-auto#L1324), decide not to bootstrap, and run the script again with `--le-auto-phase2` to re-enter "phase 2".

In "phase 2", we check if `$VENV_PATH` exists (which it doesn't), but `$OLD_VENV_PATH` exists, so we delete it, symlink it, and run this script yet again, however, this time, when hitting "phase 1", neither installation location exists so we rebootstrap and everything works.

This is pretty ugly and excessive though. I'll change this up to delete any installation location that exists here, but I'm also going to change this up to check `$OLD_VENV_PATH` more deeply (by specifically looking for `$OLD_VENV_PATH/bin/letsencrypt` as I think one of the cases where both exist is `~/.local/share/letsencrypt` is a location made by the user and not `certbot-auto`.

If you disagree, please let me know!

bmw:
Holy hell that took loner than I expected. I need more sleep/caffeine.

I landed the changes I described [here](https://github.com/certbot/certbot/pull/5392#discussion_r160726626) with a test of that behavior. If you want to run them manually as I described in [this post](https://github.com/certbot/certbot/pull/5392#issuecomment-356503127), you also have to install `sudo` now before running the test. This is handled automatically in the test farm tests as you can see by the `sudo` usage [here](https://github.com/certbot/certbot/blob/00634394f2e077ec5e621acd869e370f8619cd08/tests/letstest/scripts/test_leauto_upgrades.sh#L13).

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -462,6 +474,10 @@ if [ "$1" = "--le-auto-phase2" ]; then
         error "install any required packages."
         # Set INSTALLED_VERSION to be the same so we don't update the venv
         INSTALLED_VERSION="$LE_AUTO_VERSION"
+        # Continue to use OLD_VENV_PATH if the new venv doesn't exist
+        if [ ! -d "$VENV_PATH" ]; then
+          VENV_BIN="$OLD_VENV_PATH/bin"
```
erikrose: It's not like it's happening suddenly with this PR, but this whole script is becoming very brittle, with state changing everywhere and all of it global. It is difficult to reason about. And it's error prone: sure, we update VENV_BIN here, but we don't update VENV_PATH. But that's okay because all the subsequent uses of it are wrapped in `if [ "$LE_AUTO_VERSION" != "$INSTALLED_VERSION" ]`, which should always be true in this case. *For now.* The assumptions are unstated and invisible, and they keep piling up. As we just talked about on IRC, we should do something about that.

And I *think* VENV_PATH is intended to mean "The modern venv path, as opposed to the OLD_VENV_PATH" vs. "The venv that we're going to run stuff from." I think we should comment that explicitly where VENV_PATH is first initialized, in this PR. That will help hold the semantics together until we do a cleanup job.

joohoi:
LGTM, even though we're creating a temporary broken symlink in some cases, it will be fixed on the subsequent call initiated by `RerunWithArgs`

letsencrypt-auto-source/letsencrypt-auto.template
bmw: `VENV_PATH` normally means both of those things:

* The modern venv path, as opposed to the OLD_VENV_PATH
* The venv that we're going to run stuff from

But this PR changes that to allow us to continue to use the old venv.

I can add comments about this to the top of the script where these variables are set if you'd like.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -64,10 +64,38 @@ iQIDAQAB
 -----END PUBLIC KEY-----
 "
 
-if ! ./letsencrypt-auto -v --debug --version || ! diff letsencrypt-auto letsencrypt-auto-source/letsencrypt-auto ; then
+if [ $(python -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//') -eq 26 ]; then
```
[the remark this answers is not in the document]
bmw: I think it is very important that we test that we do the right thing on every OS in `targets.yml` where Python 2.6 is installed. Our plan after this release is to stop supporting Python 2.6 entirely so for any OSes ever specified in `targets.yml` where Python 2.6 is used, we should be upgrading them to a newer version of Python. This is the case for all RHEL 6 based distros such as CentOS 6 and some versions of Amazon Linux.

Also, unfortunately I think these tests have been split for a long time. `test_leauto_upgrades.sh` predates `auto_test.py`. The tests that Yomna wrote in #5232 also use the test farm rather than `auto_test.py` because she determined that it was easier for her to deal with all the OS specific differences there rather than in `auto_test.py`. The test farm tests are also currently run on more systems than `auto_test.py`, `auto_test.py` doesn't actually run a real version of Certbot, and it's currently easier to test an upgrade from a real released version of `certbot-auto` in the test farm.

With that said, these are things we can change and we've actually been talking about that lately in #4947 where I'm pushing for us to rethink how we're doing testing `letsencrypt-auto` and all of it's embedded scripts. Since this PR is blocking the release and neither the test farm tests or `auto_test.py` are included in the release, unless there is a functional problem with the tests here, I'd like to leave this as is and move the conversation about rethinking our testing for certbot-auto to #4947.

letsencrypt-auto-source/letsencrypt-auto.template
erikrose: I move we retitle #4947 with your excellent observation: "Most people don't feel comfortable reviewing this." ;-)

tests/letstest/scripts/test_leauto_upgrades.sh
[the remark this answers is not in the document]
erikrose: I'm not sure how testing on multiple OSs is relevant to this patch; we're just checking and acting on the presence of various directories; spinning up a bunch of OSs (which themselves don't even have anything to do with making those dirs) seems like gross overkill. Also, testing for the presence of Python 3 seems a coincidental side effect to test against, one which could easily change down the road. At the very least, the test needs its intentions documented so we know what to do when that assumption breaks. I still really don't like it, but if you want to merge it to get the user-facing functionality out, I won't stand in the way.

bmw:
To repost (a slightly paraphrased) version here of what I put in IRC:

Thanks for the +1 Erik. I think I have a lot of the same concerns as you do and normally I think/hope I'd definitely argue in favor of doing the much needed refactoring rather than rushing out a release but we're under some deadlines from other projects that aren't likely to be pushed back so I want to prioritize that stuff and after that's done, come back to these conversations and regain our sanity.

PR #5399, opened 2018-01-09
Only respect LE_PYTHON inside USE_PYTHON_3 if we know a user must have set it

### LINKED ISSUE #5388: Properly handling user set LE_PYTHONs for RHEL/CentOS 6

ohemorange:
Fixes #5388.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -255,7 +255,11 @@ DeterminePythonVersion() {
   #
   # If no Python is found, PYVER is set to 0.
   if [ -n "$USE_PYTHON_3" ]; then
-    for LE_PYTHON in "$LE_PYTHON" python3; do
+    if [ -n "$PRE_UPGRADE_LE_VERSION" ]; then
+      TEST_PYTHON="$LE_PYTHON"
+    fi # We stopped exporting LE_PYTHON, so if LE_PYTHON exists, a user set it.
+       # Respect the user-set value.
+    for LE_PYTHON in "$TEST_PYTHON" python3; do
       # Break (while keeping the LE_PYTHON value) if found.
       $EXISTS "$LE_PYTHON" > /dev/null && break
     done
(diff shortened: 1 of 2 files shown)


bmw:
I don't think this meets this property that I wrote:

* If in the future, when someone manually sets LE_PYTHON to Python 2.6, we tell them their python is too old like we do for other versions.

In the `USE_PYTHON_3` case, unless we're upgrading, we never respect `LE_PYTHON`.

ohemorange:
I interpreted that as "let's plan to make that change for future versions"; is that not the case?

bmw:
My plan to make that happen though was to just change the min version number to 2.7. This approach requires us doing more work later than that, but if you can describe the changes we'd make to do it, I'm fine with it.

ohemorange:
How do you feel about saving $PRE_UPGRADE_LE_VERSION to a file?

bmw:
Proposal: At the top of phase 1, have code like:
```
if [ -z "$PRE_UPGRADE_LE_VERSION" ]; then
  export PRE_UPGRADE_LE_VERSION="$LE_AUTO_VERSION"
fi
```
and remove setting it elsewhere. If we rebootstrap during phase 2, we have to unset the value, but otherwise, I think with this and your `DeterminePythonVersion` changes, this works. I might also change the name to `PHASE_1_VERSION` but I don't care too much.

If you don't like this proposal though, I'm potentially fine with us putting a file somewhere, with my main concern being where to put the file (especially if haven't created the venv yet).

bmw:
Closing in favor of #5402.

PR #5402, opened 2018-01-10
Only respect LE_PYTHON inside USE_PYTHON_3 if we know a user must have set it version 2

bmw:
Tests coming, but because to really test this you need to use an older version of the script and upgrade to a new one, I want to use `test_leauto_upgrades` which has unmerged changes in #5392.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -274,7 +274,6 @@ DeterminePythonVersion() {
       return 0
     fi
   fi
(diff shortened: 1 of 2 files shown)


bmw:
After this and #5392 are merged, I'd like `test_leauto_upgrades.sh` to have these [contents](https://github.com/certbot/certbot/files/1617290/test_leauto_upgrades.txt).

bmw:
Merging this now because we definitely need it for the release and it makes it easier to understand why we need #5392.

PR #5414, opened 2018-01-11
Implement HTTP-01 challenge for Nginx

### LINKED ISSUE #5411: HTTP01 support for Nginx
Part of #5409.

ohemorange:
Fixes #5411.


ohemorange:
Tests and factoring out duplicated code to come.

ohemorange:
End-of-night update:

- Manual test of http-01 with current boulder works.
- I think my refactoring broke tls-sni-01; see failing tests.
- http-01 could use manual testing a box with IPv6 set up properly.

ohemorange:
TODO: Decide how we feel about the interaction of `choose_vhost` logic and listen directives and our new server block only serving on http.

ohemorange:
Test to run in the morning: put back the separate, largely-duplicated files to see if that fixes it (and it was therefore a problem with the refactoring)

certbot-nginx/certbot_nginx/challenges.py
```diff
@@ -0,0 +1,330 @@
...
+
+        """
+        addresses = []
+        ipv6, ipv6only = self.configurator.ipv6_info(port)
+
+        for achall in self.achalls:
+            vhost = self.configurator.choose_vhost(achall.domain, create_if_no_match=True)
+
+            if vhost is not None and vhost.addrs:
+                non_ssl_addrs = (addr for addr in vhost.addrs if not addr.ssl)
```
bmw: I was looking through this while giving Joona to read the stuff I wrote and I think this line is the problem with TLS-SNI on this branch. Before the refactor, this line was https://github.com/certbot/certbot/blob/9e952081014b9545ce6ddb8b6ecc86a51bf94131/certbot-nginx/certbot_nginx/tls_sni_01.py#L61.

Applying the below diff to this branch fixed the compatibility tests when running them locally:
```
$ git diff
diff --git a/certbot-nginx/certbot_nginx/challenges.py b/certbot-nginx/certbot_nginx/challenges.py
index aff1c57..ee1c17e 100644
--- a/certbot-nginx/certbot_nginx/challenges.py
+++ b/certbot-nginx/certbot_nginx/challenges.py
@@ -68,8 +68,9 @@ class NginxChallengePerformer(common.ChallengePerformer):
             vhost = self.configurator.choose_vhost(achall.domain, create_if_no_match=True)
 
             if vhost is not None and vhost.addrs:
-                non_ssl_addrs = (addr for addr in vhost.addrs if not addr.ssl)
-                addresses.append(list(non_ssl_addrs))
+                #non_ssl_addrs = (addr for addr in vhost.addrs if not addr.ssl)
+                #addresses.append(list(non_ssl_addrs))
+                addresses.append(list(vhost.addrs))
             else:
                 if ipv6:
                     # If IPv6 is active in Nginx configuration
```
Obviously the real fix will be something different, but this should mean we're on the right track!

ohemorange: Ah thanks!! I knew there had to be something I'd missed that I'd changed in the refactor...

ohemorange: Not that I'm still awake

ohemorange: this is a good reminder that more-awake erica should go over all the changes with a fine-tooth comb

ohemorange:
Update: this is all wrong. The location block needs to be added to the existing server block because we don't want to stop serving their site while we renew.

Also, check to make sure that ctrl+c during `--debug-challenges` always failed to rollback.

ohemorange:
Update: new version of http01 works manually.

Need to add unit tests, do more robust testing, and update integration tests*.

\* Integration tests relied on the access_log and error_log. I'd like to only add those only when we're testing. Otherwise, we'll need to play with the `replace=True` parameter to `add_server_directives` to have an option to not overwrite if it exists and also not error out. An alternative is to have those directives only apply inside a location block. Will look into if that's possible.

ohemorange:
Tests I plan to do/add:

- [ ] test with existing location block
- [ ] test without existing location block

certbot-nginx/certbot_nginx/http_01.py
```diff
@@ -0,0 +1,124 @@
...
+        larger array. NginxHttp01 is capable of solving many challenges
+        at once which causes an indexing issue within NginxConfigurator
+        who must return all responses in order.  Imagine NginxConfigurator
+        maintaining state about where all of the http-01 Challenges,
+        TLS-SNI-01 Challenges belong in the response array.  This is an
+        optional utility.
+
+    """
+
+    def __init__(self, configurator):
```
bmw: nit: This method can be removed.

ohemorange:
- [ ] choose_vhost should probably be choose_redirect_vhost (or something) because we want the http one
- [ ] add tests around various ways it could be set up with redirects and such

bmw:
I'm going to take a look at integration test logging. I think it's just cruft.

bmw:
Especially since we currently don't test http01 with nginx in integration tests.

bmw:
> Also, check to make sure that ctrl+c during --debug-challenges always failed to rollback.

Did this happen? This shouldn't happen. (Unless you pressed ctrl+c multiple times because Certbot will only catch so many.)

ohemorange:
Manual tests that passed for choose_redirect_vhost change (choose block listening on 80):

- listen 80, has name match
- no listen 80
- listen 80 listen 443
- no name match

ohemorange:
> Did this happen? This shouldn't happen. (Unless you pressed ctrl+c multiple times because Certbot will only catch so many.)
It's not happening now, so I must have.

bmw:
I didn't test removing it from `tls_sni_01.py` (not adding more cruft is valuable but let's not spend time right now removing existing cruft), but for HTTP, you can remove all of the `access` and `error` log stuff. Successful Travis branch was https://travis-ci.org/certbot/certbot/builds/327871050.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -382,12 +382,28 @@ def choose_redirect_vhost(self, target_name, port):
...
+            MisconfigurationError.
+
         :returns: vhost associated with name
         :rtype: :class:`~certbot_nginx.obj.VirtualHost`
 
         """
         matches = self._get_redirect_ranked_matches(target_name, port)
-        return self._select_best_name_match(matches)
+        vhost = self._select_best_name_match(matches)
+        if not vhost:
```
bmw: This might just be temporary while you're still working on it, but I don't think we should do this here. Testing on a nginx config without a HTTP server block for a domain and requesting a redirect I got:

```
MisconfigurationError: Cannot find a VirtualHost matching domain example.org. In order for Certbot to correctly perform the challenge please add a corresponding server_name directive to your nginx configuration: https://nginx.org/en/docs/http/server_names.html
```

ohemorange: We should do this. That was a bug because I was testing against an integer when something needed a string. Can you check if this is still happening?

ohemorange: (This doesn't change the redirect behavior at all, it just reuses this method for http01.)

ohemorange: (I'm assuming that failure was happening because it wasn't finding the server block to do a challenge with. I didn't catch that in my manual testing because I have a default block.)

bmw: I can still reproduce. See https://github.com/certbot/certbot/blob/9e952081014b9545ce6ddb8b6ecc86a51bf94131/certbot-nginx/certbot_nginx/configurator.py#L600.

bmw: Previously we were letting `choose_redirect_vhost` return `None` and checking that down below. Now we raise an exception out of `choose_redirect_vhost`.

bmw:
I have integration tests on the test-http01-nginx2 branch. Previously they were failing due to me, but the last two have been failures in the challenge, first because weren't listening on the right port (which is a nonstandard port in the tests) and second because we couldn't find the matching server block.

ohemorange:
> I have integration tests on the test-http01-nginx2 branch. Previously they were failing due to me, but the last two have been failures in the challenge, first because weren't listening on the right port (which is a nonstandard port in the tests) and second because we couldn't find the matching server block.

Is this up-to-date with this branch?

ohemorange:
Also tested:

- Works once we've added a redirect
- Works once we've added a redirect, in the old style

certbot-nginx/certbot_nginx/configurator.py
ohemorange: Ah cool that's easy enough to fix

ohemorange: should be fixed now

bmw:
Tests failed again. `server_name` directive sure looks like it's there. Look at https://travis-ci.org/certbot/certbot/builds/327901082#L1697 and the lines right after it where we reinvoke Certbot.

ohemorange:
Yes, we expect that to fail. ACME only talks to port 80.

bmw:
So Certbot respects `--http-01-port` and `--tls-sni-01-port` to allow people who have their config behind a proxy/load balancer and listening on a different port to still work. This works properly for tls-sni-01 and nginx where we're using port 5001.

bmw:
And new tests running at https://travis-ci.org/certbot/certbot/builds/327903318.

ohemorange:
> So Certbot respects --http-01-port and --tls-sni-01-port to allow people who have their config behind a proxy/load balancer and listening on a different port to still work. This works properly for tls-sni-01 and nginx where we're using port 5001.

It should be respecting the flag. Will look into that.

ohemorange:
The tests don't look like they're being called with the flag, is that happening somewhere I'm missing?

ohemorange:
Yeah, that's the problem:

```
/home/travis/build/certbot/certbot/.tox/py27_install/bin/certbot --server http://localhost:4000/directory --no-verify-ssl --tls-sni-01-port 5001 --http-01-port 5002 --manual-public-ip-logging-ok --config-dir /tmp/leitpU0i/conf --work-dir /tmp/leitpU0i/work --logs-dir /tmp/leitpU0i/logs --non-interactive --no-redirect --agree-tos --register-unsafely-without-email --debug -vv --renew-by-default --authenticator nginx --installer nginx --nginx-server-root /tmp/leitpU0i/nginx rollback --checkpoints 9001
```

But you're listening on 8081 in the linked line.

ohemorange:
Pushed the fix to the integration test branch; build here: https://travis-ci.org/certbot/certbot/builds/327911272

ohemorange:
I think it's just failing on coverage for some reason now. Going to stick it on this branch (minus the ripping out everything else) and see what happens.

pde:
- [ ] confirm behaviour is correct when there is a preexisting `location` block that overlaps with the new one (eg, a `proxy_pass` block)
- [ ] read the nginx source code to double check that?
- [ ] add a test to check that?

PR #5417, opened 2018-01-11
HTTP01 support in Apache

### LINKED ISSUE #5410: HTTP01 support for Apache
Part of #5409.

bmw:
Fixes #5410. This PR should be merged and not squashed to preserve authorship.

@joohoi, the biggest change here that we didn't talk about much is the `-f` to `a2dismod`. If we determine one of the modules we need isn't enabled, we enable it and correctly try to disable it during challenge cleanup. The problem that happens is on newer Debian based systems, we get a prompt like the following when trying to disable some of these core modules:
```
# a2dismod alias     
WARNING: The following essential module will be disabled.
This might result in unexpected behavior and should NOT be done
unless you know exactly what you are doing!
 alias

To continue type in the phrase 'Yes, do as I say!' or retry by passing '-f':
```
This prompt causes challenge cleanup to hang indefinitely. I fixed this by always adding `-f` to the command line when using `a2dismod`. It doesn't seem needed on older systems, but through testing and looking at the `a2dismod` code, the script accepts it on older systems so I went with the easy fix rather than adding more branches to our code.

I tested this on:

* Debian 7
* Ubuntu 16.04
* Debian sid


bmw:
New commit LGTM.

bmw:
I resisted the urge to make the module management tests also use metaclasses :smile:.

Issue #5409, opened 2018-01-10
HTTP01 support tracking issue


ohemorange:
General things to remember/ edge cases/ considerations go here.

~Apache issue: #5410~
~Nginx issue: #5411~
~Nginx follow-up PR: #5409~

- [x] Update `certbot.eff.org` to show a short message about what's going on with useful links.
- [x] Try and properly handle redirects and proxying for HTTP01 in Apache. Failing that, include logging about the error.
- [x] Ensure HTTP01 challenge succeeds in nginx when the server is not currently listening on the HTTP01 port. (#5421)
- [x] Serve 404s from dummy nginx block ~or convince ourselves it will never serve sensitive content~. (#5437)
- [x] Write nginx unit/system tests.
- [x] Add a rewrite directive for the .well-known location so we don't hit existing rewrites (#5436)

ohemorange:
@bmw's main concerns:

1) Redirects, proxypass directives, etc. taking precedence
2) Making sure Apache/Nginx is happy to serve files out of that directory and doesn't complain about permissions

bmw:
To expand slightly:

1. We should make sure the HTTP-01 response is served regardless of any redirects, proxying, etc. the user may have configured for that domain.
2. Apache (and I think Nginx) allow you to set permissions for different paths on where the server can and cannot serve files from. We should make sure this doesn't stop the challenge from being served.
3. Nginx (and maybe Apache) sometimes don't like serving files from hidden directories. We should make this doesn't break the challenge.
4. In `get_chall_pref`, we should return TLS-SNI before HTTP. HTTP will be used if TLS-SNI isn't available, but TLS-SNI has been tested WAY more in these plugins so we should probably use TLS-SNI if the server supports both.

ohemorange:
Is TLS-SNI off in a way that will have us automatically fall through to HTTP?

bmw:
Yep. You can test that with standalone if you want. It's set up the way I describe in 4.

5. For the HTTP challenge, the CA will always request the file over plain HTTP rather than HTTPS. We should try to do the right thing when the vhost we're doing the challenge for is either HTTP or HTTPS.

ohemorange:
HTTP is only available over HTTP, not HTTPS.

bmw:
For certbot.eff.org we want, a short note on certbot.eff.org with links to what's going on. You may want to link to some combination of:

[...]

PR #5421, opened 2018-01-12
Handle more edge cases for HTTP-01 support in Nginx

ohemorange:
Fix nginx default checking so only checks by port.
If we still can't find a default (or found too many), make a dummy block.
Don't change the behavior for anything that's not HTTP01.

Part of #5409.

### DIFF
--- a/certbot-nginx/certbot_nginx/http_01.py
+++ b/certbot-nginx/certbot_nginx/http_01.py
@@ -8,6 +8,9 @@ from acme import challenges
 from certbot import errors
 from certbot.plugins import common
 
+from certbot_nginx import obj
+from certbot_nginx import nginxparser
+
 
 logger = logging.getLogger(__name__)
 
@@ -32,6 +35,13 @@ class NginxHttp01(common.ChallengePerformer):
 
     """
 
+    def __init__(self, configurator):
+        super(NginxHttp01, self).__init__(configurator)
+        self.challenge_conf = os.path.join(
+            configurator.config.config_dir, "le_http_01_cert_challenge.conf")
+        self._ipv6 = None
+        self._ipv6only = None
+
     def perform(self):
         """Perform a challenge on Nginx.
 
@@ -52,8 +62,16 @@ class NginxHttp01(common.ChallengePerformer):
 
         return responses
 
-    def _add_bucket_directive(self):
-        """Modifies Nginx config to include server_names_hash_bucket_size directive."""
+    def _mod_config(self):
+        """Modifies Nginx config to include server_names_hash_bucket_size directive
+           and server challenge blocks.
+
+        :raises .MisconfigurationError:
+            Unable to find a suitable HTTP block in which to include
+            authenticator hosts.
+        """
+        included = False
+        include_directive = ['\n', 'include', ' ', self.challenge_conf]
         root = self.configurator.parser.config_root
 
         bucket_directive = ['\n', 'server_names_hash_bucket_size', ' ', '128']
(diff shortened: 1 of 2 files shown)


ohemorange:
Things I've tested manually (making sure that location is added where it belong not added where it doesn't belong, including in nginx config files and the `/etc/letsencrypt/le_http_01_cert_challenge.conf` file):
- normal
- change port
- change port on default but add flag for deafult port
- change port on default
- no matches, turn off ipv6 on all ports
- multiple domains (check that both names are on cert)
- multiple domains one matching one not with default server
- multiple domains one matching one not without matching default server
- multiple domains neither has match

certbot-nginx/certbot_nginx/http_01.py
```diff
@@ -71,30 +90,98 @@ def _add_bucket_directive(self):
...
+        validation = achall.validation(achall.account_key)
+        validation_path = self._get_validation_path(achall)
 
-    def _get_validation_path(self, achall):
-        return os.sep + os.path.join(challenges.HTTP01.URI_ROOT_PATH, achall.chall.encode("token"))
+        block.extend([['server_name', ' ', achall.domain],
+                      [['location', ' ', '=', ' ', validation_path],
+                        [['default_type', ' ', 'text/plain'],
+                         ['return', ' ', '200', ' ', validation]]]])
+        # TODO: do we want to return something else if they otherwise access this block?
```
bmw: What's nginx's default behavior here?

bmw:
I didn't do any testing of this but I read through the code and it looks really good. Again, you're the best for turning this around so quickly.

While reviewing, I thought of a few things that are a corner cases of corner cases so it's probably not worth investing much (or maybe any) engineering effort in, but I think in theory we could do better with the address we use on the `listen` statement in some cases, mainly when creating our own dummy server block.

One case is Certbot takes an `--http-01-address` flag that matches the port flag. If someone set this, we should ideally respect it and when creating a dummy server block, we should have nginx use that address rather than listening on all interfaces. It's worth noting I didn't do this right in Apache either.

The second case is what if someone sets this and we find an existing server block to use. In theory, I think the behavior would be to use the server block only if it's listening on the same address, but this would likely get complicated when trying to compare the different variations of "listen on all interfaces".

The final case is when searching the config to see if there any IPv6 listens, we ignore (to different degrees) the address and port. We completely ignore the address and for the port, we're only looking for `ipv6only`. Even if nginx isn't listening with IPv6 on that port, we'll still use IPv6 here because it's listening on IPv6 for some address/port combination.

This is not likely to be a problem for (m)any Certbot users, but some people with valid setups have specific requirements for what addresses and ports a services should be listening to. For example, they may have multiple IP addresses allocated to a box with different services running on each. Certbot configuring nginx to use additional addresses/ports beyond what is currently in the nginx config or specified by the user can cause the challenge to fail. (It also will likely make the challenge succeed in cases it wouldn't previously.)

I think it's probably worth resolving the `--http-01-address` case for the dummy server block. Not so sure about the other cases.

ohemorange:
Thanks for the review! I'd like our guiding star to be "are we broken for users that we would have analogously have been able to handle were we using tls-sni-01".

certbot-nginx/certbot_nginx/http_01.py
ohemorange: 404

bmw:
While I think my long post above about listen addresses discuss real problems people may hit, since we've never handled it right in TLS-SNI for nginx (and never got any complaints about it), I don't think we should block shipping HTTP-01 support on these features.

Issue #5427, opened 2018-01-14
HTTP01 fails with default CentOS 7 config due to PrivateTmp


ryanjaeb:
## My operating system is:

I'm using CentOS 7.3, but this is likely applicable to RHEL 7 too.

## I installed Certbot with:

I installed Certbot HEAD (2cb9d9e2aa332e4fd53aab32ad1d3946438a0dd3) using [these instructions](https://community.letsencrypt.org/t/help-test-certbot-apache-and-nginx-fixes-for-tls-sni-01-outage/50207).

## I ran this command:

```
venv/bin/certbot --apache --preferred-challenges http -d sub.example.com
```
and it produced this output:

```
Performing the following challenges:
http-01 challenge for sub.example.com
Waiting for verification...
Cleaning up challenges
Failed authorization procedure. sub.example.com (http-01): urn:acme:error:unauthorized
```

My server is giving back a 403 forbidden error.

## Certbot's behavior differed from what I expected because:

The `httpd` service on CentOS 7 and, I assume, RHEL 7 use [SystemD's PrivateTmp support](https://access.redhat.com/blogs/766093/posts/1976243) by default.  When the `httpd` process reads `/tmp`, it's actually reading from a private folder similar to `/tmp/systemd-private-*`.   As far as the `httpd` process is concerned, the `/tmp/*` directory temporarily used to alias `/.well-known/acme-challenge` doesn't exist, so an error is returned rather than what's expected.

## Workaround

It's possible to use a [SystemD drop-in](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sect-Managing_Services_with_systemd-Unit_Files#brid-Managing_Services_with_systemd-Extending_Unit_Config) to override the `PrivateTmp` feature of the `httpd` service.  However, this is **not recommended** by the author of the [RHEL blog article](https://access.redhat.com/blogs/766093/posts/1976243) linked above.  The author recommends the use of `/run` or `/var/lib` instead.

```
sudo mkdir /etc/systemd/system/httpd.service.d

sudo tee /etc/systemd/system/httpd.service.d/no-privatetmp.conf > /dev/null << 'EOF'
[Service]
PrivateTmp=false
EOF

sudo systemctl daemon-reload
sudo systemctl restart httpd
```

## Other Notes

It's possible to get to the real tmp directory for httpd via the `/proc` directory, but I don't know if it's useful:

```
systemctl status httpd | grep "Main PID"
```

```
[...]

PR #5428, opened 2018-01-14
Use static directory under workdir for HTTP challenges

joohoi:
Instead of using randomly generated temporary directory for HTTP challenge files, use a directory under Certbot work directory. This fixes issues with systemd PrivateTmp.

Fixes: #5427

### DIFF
--- a/certbot-apache/certbot_apache/http_01.py
+++ b/certbot-apache/certbot_apache/http_01.py
@@ -1,8 +1,6 @@
 """A class that performs HTTP-01 challenges for Apache"""
 import logging
 import os
-import shutil
-import tempfile
 
 from certbot.plugins import common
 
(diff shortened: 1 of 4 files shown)


certbot-apache/certbot_apache/http_01.py
```diff
@@ -105,8 +100,10 @@ def _set_up_challenge(self, achall):
         response, validation = achall.response_and_validation()
 
         name = os.path.join(self.challenge_dir, achall.chall.encode("token"))
+
         with open(name, 'wb') as f:
             f.write(validation.encode())
+            self.configurator.reverter.register_file_creation(True, name)
```
bmw: We should register the file with the reverter before creating it. While unlikely, if we get a signal/exception between writing the file and finishing registering it, it will never be deleted.

PR #5435, opened 2018-01-16
Cleanup dockerfile-dev

bmw:
I'm using this currently and hit a number of snags. Changes are:

* Upgrade to Xenial which has Python libraries with real SSL support.
* Stop storing default Certbot directories as volumes as this is rarely (if ever) necessary in development.
* Expose port 80 as well for testing internal web servers.
* Install apache and nginx (this causes Augeas to be installed and makes Apache tests work).
* Copy all files from the repo so tools continue to work as they begin to rely on more files.
* Use `tools/venv.sh` which pins all dependency versions for increased stability and installs all plugins.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -9,7 +9,7 @@ WORKDIR /opt/certbot/src
 # TODO: Install Apache/Nginx for plugin development.
 COPY . .
 RUN apt-get update && \
-    apt-get install apache2 git nginx-light python3-dev -y && \
+    apt-get install apache2 git nginx-light -y && \
     letsencrypt-auto-source/letsencrypt-auto --os-packages-only && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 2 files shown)


Dockerfile-dev
```diff
@@ -1,70 +1,21 @@
...
-# If <dest> doesn't exist, it is created along with all missing
-# directories in its path.
-
 # TODO: Install Apache/Nginx for plugin development.
-COPY letsencrypt-auto-source/letsencrypt-auto /opt/certbot/src/letsencrypt-auto-source/letsencrypt-auto
-RUN /opt/certbot/src/letsencrypt-auto-source/letsencrypt-auto --os-packages-only && \
-    apt-get install python3-dev git -y && \
+COPY . .
+RUN apt-get update && \
+    apt-get install apache2 git nginx-light python3-dev -y && \
```
joohoi: Is python3-dev needed here as we're effectively installing python2 venv?

bmw: You're right. Nice catch. I thought we needed it to run Python 3 tests in tox, but cryptography's precompiled wheel saves us here. I've removed the package.

Issue #5440, opened 2018-01-17
Nginx HTTP-01 tests


ohemorange:
Part of #5409.

- [ ] We should make sure the HTTP-01 response is served regardless of any redirects, proxying, etc. the user may have configured for that domain.
- ~Apache (and I think Nginx) allow you to set permissions for different paths on where the server can and cannot serve files from. We should make sure this doesn't stop the challenge from being served.~
- ~Nginx (and maybe Apache) sometimes don't like serving files from hidden directories. We should make this doesn't break the challenge.~
- ~In get_chall_pref, we should return TLS-SNI before HTTP. HTTP will be used if TLS-SNI isn't available, but TLS-SNI has been tested WAY more in these plugins so we should probably use TLS-SNI if the server supports both.~
- [ ] For the HTTP challenge, the CA will always request the file over plain HTTP rather than HTTPS. We should try to do the right thing when the vhost we're doing the challenge for is either HTTP or HTTPS.
- [x]  test with existing location block
- ~test without existing location block~
- [x]  test with multiple domains
- [ ]  add tests around various ways it could be set up with redirects and such
- [x] listen 80, has name match
- [ ] no listen 80
- [x] listen 80 listen 443
- [x] no name match
- [x] Works once we've added a redirect
- [x] Works once we've added a redirect, in the old style:
```
if ($scheme != https) {
  return 301 https://$host$request_uri;
}
```
- [x] Works with `rewrite 301 https://$host$request_uri` at `server` level
- [ ] rewrite directive is added at top, including with newline
- [x] works with `--http-01-port`
- [x] confirm behaviour is correct when there is a preexisting `location` block that overlaps with the new one (eg, a `proxy_pass` block)
- [ ] nginx default checking should only check by port during http-01 and only during http-01
- [x] default block is selected if there is an appropriate one
- [x] if there's no matching on the right port, we still work
- [x] change port
- [x] change port on default but add flag for deafult port
- [x] change port on default
- [ ] no matches, turn off ipv6 on all ports
- [ ] multiple domains (check that both names are on cert)
- [ ] multiple domains one matching one not with default server
- [ ] multiple domains one matching one not without matching default server
- [ ] multiple domains neither has match
- [ ] document root is set to a nonexistent location so we 404

Issue #5438, opened 2018-01-17
Follow-ups from enabling HTTP-01


ohemorange:
Things we noticed need fixing while we were there (#5409), and code that can be cleaned up.

- `certbot-nginx/certbot_nginx/tls_sni_01.py` and `certbot-nginx/certbot_nginx/http_01.py` could use some deduping -- fix this when we need to
- (nginx) tls sni should have `root` not inside a `location /` block -- probably not worth fixing
- [x] (#5787) (nginx) location is useless as a repeatable directive -- our repeatable directive infrastructure doesn't support blocks as the first argument
- [x] (#5794) code could use deduping internally to `certbot-nginx/certbot_nginx/http_01.py` -- for example, `rewrite_directive` is defined identically twice
- [x] (#5788) Nginx default checking: allow prefix matching of `default_server`, check by port more often, use the first block found listening on that port like Nginx does, be IPv6 sensitive
- [x] (#5793) When copying a default block, remove ipv6only=on because this completely breaks it haha
- "While reviewing, I thought of a few things that are a corner cases of corner cases so it's probably not worth investing much (or maybe any) engineering effort in, but I think in theory we could do better with the address we use on the listen statement in some cases, mainly when creating our own dummy server block.

One case is Certbot takes an --http-01-address flag that matches the port flag. If someone set this, we should ideally respect it and when creating a dummy server block, we should have nginx use that address rather than listening on all interfaces. It's worth noting I didn't do this right in Apache either.

The second case is what if someone sets this and we find an existing server block to use. In theory, I think the behavior would be to use the server block only if it's listening on the same address, but this would likely get complicated when trying to compare the different variations of "listen on all interfaces".

The final case is [now located at #5795].

I think it's probably worth resolving the --http-01-address case for the dummy server block. Not so sure about the other cases." -- add this if anyone ever requests or complains about it

PR #5441, opened 2018-01-17
Add integration tests for nginx plugin

ohemorange:
Towards #5440. Part of #5409.

This is built on top of cherry-picked commits from #5438 and #5436, so may need to do some git wrangling before merging. Actual changes for this PR are to `certbot-nginx/tests/boulder-integration.conf.sh` and `certbot-nginx/tests/boulder-integration.sh`.


ohemorange:
This isn't failing on a test, it's failing on "did Nginx pick up the change we thought it would" and I think there's a chance it just hasn't loaded yet. Will look in a bit.

bmw:
Should we delay the release a bit so you can look into this?

ohemorange:
I don't think it's necessary.

PR #5453, opened 2018-01-19
Proper webroot directory cleanup

### LINKED ISSUE #1003: Webroot plugin should delete .well-known/acme-challenge on cleanup

sydneyli:
Fixes #1003 .

the webroot authenticator keeps track of successfully created dirs in a stack, and tries to delete them on cleanup.

### DIFF
--- a/certbot/plugins/util.py
+++ b/certbot/plugins/util.py
@@ -10,6 +10,10 @@ def get_prefixes(path):
     """Retrieves all possible path prefixes of a path, in descending order
     of length. For instance,
         /a/b/c/ => ['/a/b/c/', '/a/b/c', '/a/b', '/a', '/']
+    :param str path: the path to break into prefixes
+
+    :returns: all possible path prefixes of given path in descending order
+    :rtype: `list` of `str`
     """
     prefix = path
     prefixes = []
--- a/certbot/plugins/util.py
+++ b/certbot/plugins/util.py
@@ -6,6 +6,20 @@ from certbot import util
 
 logger = logging.getLogger(__name__)
 
+def get_prefixes(path):
+    """Retrieves all possible path prefixes of a path, in descending order
+    of length. For instance,
+        /a/b/c => ['/a/b/c/', '/a/b/c', '/a/b', '/a', '/']
+    """
+    prefix = path
+    prefixes = []
+    while len(prefix) > 0:
+        prefixes.append(prefix)
+        prefix, _ = os.path.split(prefix)
+        # Break once we hit '/'
+        if prefix == prefixes[-1]:
+            break
+    return prefixes
 
 def path_surgery(cmd):
     """Attempt to perform PATH surgery to find cmd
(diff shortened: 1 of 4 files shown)


certbot/plugins/util.py
```diff
@@ -6,6 +6,20 @@
 
 logger = logging.getLogger(__name__)
 
+def get_prefixes(path):
+    """Retrieves all possible path prefixes of a path, in descending order
+    of length. For instance,
+        /a/b/c/ => ['/a/b/c/', '/a/b/c', '/a/b', '/a', '/']
```
bmw: nit: We should document the parameter and return value using Sphinx syntax so if we build the documentation for this file, it's nicely formatted. You can find an example of what this syntax looks like and a link to the general Sphinx documentation [here](https://certbot.eff.org/docs/contributing.html#coding-style). There are many other examples in the code base.

You can build the documentation locally by activating the virtual environment and running the command described [here](https://certbot.eff.org/docs/contributing.html#updating-the-documentation).

certbot/plugins/webroot.py
```diff
@@ -161,27 +164,26 @@ def _create_challenge_dirs(self):
...
-                if exception.errno != errno.EEXIST:
-                    raise errors.PluginError(
-                        "Couldn't create root for {0} http-01 "
-                        "challenge responses: {1}", name, exception)
+                stat_path = os.stat(path)
+                for prefix in sorted(util.get_prefixes(self.full_roots[name]), key=len):
+                    try:
+                        # This is coupled with the "umask" call above because
+                        # os.makedirs's "mode" parameter may not always work:
+                        # https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python
```
bmw: nit: We should update this comment since `makedirs` isn't being used anymore. We should at least change the referenced function in the comment, but we may also want to change the link to one directly relevant to [mkdir](https://docs.python.org/3/library/os.html#os.mkdir).

certbot/plugins/webroot.py
```diff
@@ -217,16 +219,14 @@ def cleanup(self, achalls):  # pylint: disable=missing-docstring
                 os.remove(validation_path)
                 self.performed[root_path].remove(achall)
 
-        for root_path, achalls in six.iteritems(self.performed):
-            if not achalls:
```
bmw: I think we actually still need a check like this `if` statement here as we should only be cleaning up the challenges given in the `achalls` parameter. In practice, I don't think this matters right now, but the [interface](https://certbot.eff.org/docs/api/interfaces.html#certbot.interfaces.IAuthenticator.cleanup) only guarantees `achalls` is a non-empty subset of previously performed challenges. We could make use of this in the future so we should make sure there are no active challenges in a directory before trying to delete it.

sydneyli: fixed this and added a test for it, though it's done implicitly because `rmdir` doesn't execute if the directory is not empty.

PR #5457, opened 2018-01-20
Certbot installer for Zimbra

trinopoty:
Implement an Installer which can install certificate to a Zimbra installation.

Testing Done:
- tox -e py27
- tox -e lint

### DIFF
--- a/certbot-installer-zimbra/certbot_installer_zimbra/__init__.py
+++ b/certbot-installer-zimbra/certbot_installer_zimbra/__init__.py
@@ -1,0 +1,1 @@
+"""Certbot zimbra installer plugin."""
--- a/certbot-installer-zimbra/certbot_installer_zimbra/installer.py
+++ b/certbot-installer-zimbra/certbot_installer_zimbra/installer.py
@@ -19,10 +19,26 @@ logger = logging.getLogger(__name__)
 @zope.interface.implementer(interfaces.IInstaller)
 @zope.interface.provider(interfaces.IPluginFactory)
 class ZimbraInstaller(common.Installer):
+    """Zimbra installer
+    """
+
     description = "Zimbra Mail Server plugin - Alpha"
 
     def __init__(self, *args, **kwargs):
         super(ZimbraInstaller, self).__init__(*args, **kwargs)
+
+        self._zimbra_temp_path = None
+        self._zimbra_temp_crt = None
+        self._zimbra_temp_key = None
+        self._zimbra_temp_ca = None
+
+        self._zimbra_cert_path = None
+        self._zimbra_cert_crt = None
+        self._zimbra_cert_key = None
+        self._zimbra_cert_ca = None
+
+        self._zimbra_user = None
+        self._deploy_cert = None
 
     @classmethod
     def add_parser_arguments(cls, add):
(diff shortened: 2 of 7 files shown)


trinopoty:
Zimbra requires the complete certificate chain including the root certificate to be provided. Is it possible to get the root certificate in the installer?

sydneyli:
Wow, really sorry for the lag on this response @trinopoty. I somehow missed your question when we first triaged this PR. Is the [`fullchain_path`](https://certbot.eff.org/docs/api/interfaces.html?highlight=fullchain_path#certbot.interfaces.IInstaller.deploy_cert) arg in `deploy_cert` what you're looking for?

trinopoty:
I've come up with a script for deploying to zimbra [here](https://gist.github.com/trinopoty/508b3e4a2856366f90bcf828ba514ec5). Basically, this plugin needs to replicate the process but there's some problems that I'm not sure will work out.
1. Newer versions of zimbra requires that the deploy command be run as the `zimbra` user.
2. Zimbra needs the root certificate to be available.

sydneyli:
We had a discussion about whether we want to take on the task of maintaining an installer for Zimbra, since few people on the team know much about it. According to [their website](https://www.zimbra.com/), "More Than 500 Million People Trust Zimbra Email Collaboration"-- and on first glance it seems decently popular in enterprises outside the US?

The largest concern is that Zimbra requires the trust anchor certificate. In accordance with the [TLS RFC](https://tools.ietf.org/html/rfc8446#section-4.4.2), most TLS servers don't include the trust anchor cert when sending the chain...
> Because certificate validation requires that trust anchors be distributed independently, a certificate that specifies a trust anchor MAY be omitted from the chain, provided that supported peers are known to possess any omitted certificates.

@trinopoty, do you happen to know why Zimbra requires the trust anchor cert? This seems very atypical, and IIRC they use Postfix under the hood, which doesn't have this requirement.

As @bmw mentioned:
> I do not want us to hardcode LE's cert in Certbot and ACME has no mechanism to fetch it. I wonder if we could programatically figure it out using the certificates given to us from the CA and our locally trusted certificates, however, which is the "right" certificate chain to give Zimbra if there's multiple paths?

trinopoty:
I'm not entirely sure **why** it needs it but the tools used to deploy the certificate to zimbra refuses to verify a partial certificate chain. Documentation about `zmcertmgr` is [here](https://wiki.zimbra.com/wiki/Administration_Console_and_CLI_Certificate_Tools#zmcertmgr).
I'll try a few different ways to see if I can get zmcertmgr to accept a cert without it's root and let you guys know.

trinopoty:
I did some investigations without any promising results.
The `zmcertmgr` command is a perl script that verifies the cert chain by calling `openssl verify`. That's why it needs the ca root cert.
Bypassing this will require copying the logic for deployment into the plugin or modifying `zmcertmgr`.

sydneyli:
@trinopoty Thank you for taking a look! It sounds like there's no way around it, so if we want to maintain this installer we'll have to find some way to programatically obtain the root cert.

sydneyli:
@trinopoty After some internal discussion, we still don't necessarily want core Certbot code to rely on something like this. Let's Encrypt certs should be [trusted by all major root programs](https://letsencrypt.org/2018/08/06/trusted-by-all-major-root-programs.html) for now, and I think the correct way to do things is to 1) ask Zimbra to fix their TLS implementation upstream, and 2) in the meantime, perhaps provide a hook script-- like the one you linked-- to do this transformation (either from the Mozilla bundle you linked or keeping it up-to-date with the existing Let's Encrypt cert in use).

So I'm going to close this PR for the time being-- we may build off of it in the future, though, so I've preserved your work on a branch called `zimbra-installer`.

Thank you for your work on this-- and bringing this to our attention!

orangepizza:
this cloud be work without hard coding a root by downloading root certificate form  Authority Information Access of intermediate certificate:  as resent influx of zimbra request from forum this may worth working on

trinopoty:
Actually, from my recent observations, certbot actually downloads and stores the root certificate in `chain.pem` file.
So we don't even need to download the certificate.

trinopoty:
Although, last time I checked, the root certificate provided by certbot was signed by DST X3 which caused validation to fail. So I had to manually provide the proper one. Not sure if this has been fixed.

orangepizza:
> Although, last time I checked, the root certificate provided by certbot was signed by DST X3 which caused validation to fail. So I had to manually provide the proper one. Not sure if this has been fixed.

that's because by default uses old-android-compatible chain, which uses isrg-signed by DST root x3 so it can chained up to DST root x3 which only old clients have: you should download isrg r1 from AIA of r3 certificate, (http://x1.i.lencr.org/)

PR #5460, opened 2018-01-23
Documentation on cron renewal

hlieberman:
Here's a first pass, @SwartzCr!

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -410,13 +410,6 @@ Renewing certificates
    days). Make sure you renew the certificates at least once in 3
    months.
 
-.. seealso:: Many of the certbot clients obtained through a
-   distribution come with automatic renewal out of the box,
-   such as Debian and Ubuntu versions installed through `apt`,
-   CentOS/RHEL 7 through EPEL, etc.  See `Automated Renewals`_
-   for more details, or consult your distribution's
-   documentation.
-
 As of version 0.10.0, Certbot supports a ``renew`` action to check
 all installed certificates for impending expiry and attempt to renew
 them. The simplest form is simply
@@ -522,6 +515,13 @@ are only renewed when they're determined to be near expiry, the command
 can run on a regular basis, like every week or every day). In that case,
 you are likely to want to use the ``-q`` or ``--quiet`` quiet flag to
 silence all output except errors.
+
+.. seealso:: Many of the certbot clients obtained through a
+   distribution come with automatic renewal out of the box,
+   such as Debian and Ubuntu versions installed through `apt`,
+   CentOS/RHEL 7 through EPEL, etc.  See `Automated Renewals`_
+   for more details, or consult your distribution's
+   documentation.
 
 If you are manually renewing all of your certificates, the
 ``--force-renewal`` flag may be helpful; it causes the expiration time of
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -410,6 +410,13 @@ Renewing certificates
    days). Make sure you renew the certificates at least once in 3
    months.
 
+.. seealso:: Many of the certbot clients obtained through a
+   distribution come with automatic renewal out of the box,
+   such as Debian and Ubuntu versions installed through `apt`,
+   CentOS/RHEL 7 through EPEL, etc.  See `Automated Renewals`_
+   for more details, or consult your distribution's
+   documentation.
+
```
SwartzCr: I have a vague preference to move this to the part of the section where we talk about cron renewals.
As such I'd move it to be after the paragraph that ends on line 524

docs/using.rst
```diff
@@ -860,6 +891,12 @@ in the config file.
 Log Rotation
 ============
 
+.. note:: Some distributions, including Debian and Ubuntu, disable
+   certbot's internal log rotation in favor of a more traditional
+   logrotate script.  If you are using a distribution's packages and
+   want to alter the log rotation, check `/etc/logrotate.d/` for a
+   certbot rotation script.
+
```
SwartzCr: Similarly I'd move this to be after the initial paragraph in this section. I think it makes more sense to explain the default log behavior and then have a note about a special case.

SwartzCr:
I fixed my own nits - @bmw wanna review this?

hlieberman:
👍

SwartzCr:
@jsha would you be willing to review this?

docs/using.rst
```diff
@@ -516,6 +516,13 @@ can run on a regular basis, like every week or every day). In that case,
 you are likely to want to use the ``-q`` or ``--quiet`` quiet flag to
 silence all output except errors.
 
+.. seealso:: Many of the certbot clients obtained through a
+   distribution come with automatic renewal out of the box,
+   such as Debian and Ubuntu versions installed through `apt`,
+   CentOS/RHEL 7 through EPEL, etc.  See `Automated Renewals`_
+   for more details, or consult your distribution's
```
jsha: Is it accurate to say that most distribution-packaged Certbots have documentation somewhere about their cron renewals? I don't see any in the Ubuntu package. I think if the current state is "mostly undocumented" we should leave out this mention of the distribution's documentation.

docs/using.rst
```diff
@@ -611,6 +618,30 @@ The following commands could be used to specify where these files are located::
...
+------------------
+
+Many Linux distributions provide automated renewal when you use the
+packages installed through their system package manager.  The
+following table is an *incomplete* list of distributions which do so,
+as well as their methods for doing so.
+
+If you are not sure whether or not your system has this already
+automated, refer to your distribution's documentation, or check your
+system's crontab and systemd timers.
```
jsha: We should provide instructions on how to do this, or link to them.

SwartzCr: Good point - maybe Harlan can take a stab at that? I can't seem to even find the debian documentation

docs/using.rst
hlieberman: The Ubuntu package is the Debian package, so they ship both cron and systemd.  Apparently CentOS and RHEL both do as well, and the combination of those distributions and their derivatives is the bulk of Linux installations out there.

docs/using.rst
hlieberman: Well... to check systemd timers, you'd do `systemctl ilst-timers`.  For cron, you'd check `/etc/crontab` and `/etc/cron.*`.  And, I guess, `/etc/anacrontab`.  That's the standard locations, at least.  But distributions could customize it; it's true for Debian/Ubuntu/CentOS/RHEL, but, for all *nix?

docs/using.rst
jsha: Right, but do the Debian/Ubuntu packages ship documentation? This PR is saying "check the OS documentation," but I think it doesn't exist.

docs/using.rst
jsha: I think providing the typical examples here is good enough.

hlieberman:
Bumping this @SwartzCr.  Can we get this merged?

docs/using.rst
jsha: I think this is still unresolved. Could you just remove the part that says "or consult your distribution's documentation"?

jsha:
Thanks @hlieberman !

PR #5462, opened 2018-01-23
Alternative: Work around Basic Authentication for challenge dir

joohoi:
This is an alternative to #5461

Unfortunately, the way that Apache merges the configuration directives is different for mod_rewrite and <Location> / <Directory> directives.

To work around basic auth in VirtualHosts, the challenge override `Include` now has an added `<Location>` block, and is included twice in the `<VirtualHost>`: once in the beginning to tackle `mod_rewrite` issues, and once in the end to fix issues with `<Directory>` and `<Location>`.


bmw:
Closing in favor of #5461.

PR #5463, opened 2018-01-23
feat(nginx plugin): add HSTS enhancement

### LINKED ISSUE #3648: HSTS support in the Nginx plugin
Currently only the Apache plugin supports this.

### LINKED ISSUE #5296: Nginx remove_server_directives should also removing a trailing '# managed by Certbot' if one exists

sydneyli:
Fixes #3648 -- adds HSTS support for nginx plugin.
Also fixes #5296.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -105,6 +105,7 @@ class NginxConfigurator(common.Installer):
         self.parser = None
         self.version = version
         self._enhance_func = {"redirect": self._enable_redirect,
+                              "ensure-http-header": self._set_http_header,
                               "staple-ocsp": self._enable_ocsp_stapling}
 
         self.reverter.recovery_routine()
@@ -621,7 +622,7 @@ class NginxConfigurator(common.Installer):
     ##################################
     def supported_enhancements(self):  # pylint: disable=no-self-use
         """Returns currently supported enhancements."""
-        return ['redirect', 'staple-ocsp']
+        return ['redirect', 'ensure-http-header', 'staple-ocsp']
 
     def enhance(self, domain, enhancement, options=None):
         """Enhance configuration.
@@ -646,6 +647,57 @@ class NginxConfigurator(common.Installer):
     def _has_certbot_redirect(self, vhost, domain):
         test_redirect_block = _test_block_from_block(_redirect_block_for_domain(domain))
         return vhost.contains_list(test_redirect_block)
+
+    def _should_break_block(self, vhost):
+        # if host has both SSL and non-SSL addrs, we should break this block!
+        return vhost.ssl and any([not addr.ssl for addr in vhost.addrs])
+
+    def _set_http_header(self, domain, header_substring):
+        """Enables header identified by header_substring on domain.
+
+        If the vhost is listening plaintextishly, separates out the relevant
+        directives into a new server block, and only add header directive to
+        HTTPS block.
+
+        :param str domain: the domain to enable header for.
+        :param str header_substring: String to uniquely identify a header.
+                        e.g. Strict-Transport-Security, Upgrade-Insecure-Requests
+        :returns: Success
+        :raises .errors.PluginError: If no viable HTTPS host can be created or
+            set with header header_substring.
+        """
+        vhost = self.choose_vhost(domain)
+        if vhost.has_header(header_substring):
+            raise errors.PluginEnhancementAlreadyPresent(
+                "Existing %s header" % (header_substring))
+
+        if vhost is None:
+            # this should never happen, since enhancements occur after
+            # setting certs!
+            raise errors.PluginError(
+                "Unable to find corresponding HTTP host for enhancement.")
+
+        # if there is no separate SSL vhost, break the vhost into two blocks:
+        if self._should_break_block(vhost):
+            http_vhost = self.parser.duplicate_vhost(vhost,
+                             only_directives=['listen', 'server_name'])
+
+            def _ssl_match_func(directive):
+                return 'ssl' in directive
+
+            def _no_ssl_match_func(directive):
+                return 'ssl' not in directive
+
+            # remove all ssl addresses from the new block
+            self.parser.remove_server_directives(http_vhost, 'listen', match_func=_ssl_match_func)
+
+            # remove all non-ssl addresses from the existing block
+            self.parser.remove_server_directives(vhost, 'listen', match_func=_no_ssl_match_func)
+
+        header_directives = [
+            ['\n    ', 'add_header', header_substring] + constants.HEADER_ARGS[header_substring],
+            ['\n']]
+        self.parser.add_server_directives(vhost, header_directives, replace=False)
 
     def _add_redirect_block(self, vhost, domain):
         """Add redirect directive to vhost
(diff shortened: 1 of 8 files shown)


bmw:
I'm going to ask @ohemorange to review this as she knows the Nginx plugin better than I do.

Erica, you should probably read the things Sydney and I talked about for this PR which I wrote up [here](https://github.com/certbot/certbot/issues/3648#issuecomment-359118606).

sydneyli:
Fixed the above! Though looks like i botched the branch/PR somehow by trying to merge in master... looking into it now

sydneyli:
Yay fixed!

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -193,6 +193,12 @@ def __eq__(self, other):
 
         return False
 
+    def has_header(self, header_name):
+        """Determine if this vhost has a redirecting statement
```
ohemorange: update description

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +566,45 @@ def _has_certbot_redirect(self, vhost, domain):
         test_redirect_block = _test_block_from_block(_redirect_block_for_domain(domain))
         return vhost.contains_list(test_redirect_block)
 
+    def _should_break_block(self, vhost):
```
ohemorange: This particular set of conditions seems to be particularly for hsts, let's either make that more clear so it's not accidentally used elsewhere, or just put the `if` inline.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +566,45 @@ def _has_certbot_redirect(self, vhost, domain):
...
+        :raises .errors.PluginError: If no viable HTTPS host can be created or
+            set with header header_substring.
+        """
+        vhost = self.choose_vhost(domain)
+        if vhost.has_header(header_substring):
+            raise errors.PluginEnhancementAlreadyPresent(
+                "Existing %s header" % (header_substring))
+
+        if vhost is None:
+            # this should never happen, since enhancements occur after
```
ohemorange: We're looking to make `enhance` a verb so this comment might not be applicable forever. Change it or not, up to you.

ohemorange: But this check should probably happen before we try to call `has_header` on a `None` vhost anyway.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +566,45 @@ def _has_certbot_redirect(self, vhost, domain):
...
+            # setting certs!
+            raise errors.PluginError(
+                "Unable to find corresponding HTTP host for enhancement.")
```
ohemorange: `choose_vhost` looks for HTTPS hosts

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +566,45 @@ def _has_certbot_redirect(self, vhost, domain):
...
+
+        # if there is no separate SSL vhost, break the vhost into two blocks and
+        # choose the SSL vhost.
+        if self._should_break_block(vhost):
+            _, vhost = self._split_block(vhost)
```
ohemorange: I know I just said we can actually reuse code, but we maybe actually can't quite as much. `_split_block` for redirect keeps only the `server_name` and `listen` directives in the new http host; here, we probably want to copy over all the directives, keeping `default_server` if it's there (it should do that automatically but I'd double-check), and then just removing the appropriate `listen` directives. It shouldn't be more of a change than passing in an optional `only_directives` parameter from the outside.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -573,6 +613,31 @@ def _add_redirect_block(self, vhost, domain):
         self.parser.add_server_directives(
             vhost, redirect_block, replace=False, insert_at_top=True)
 
+    def _split_block(self, vhost):
+        """Splits this "virtual host" (i.e. this nginx server block) into
```
ohemorange: you can just call it a server block

certbot-nginx/certbot_nginx/configurator.py
sydneyli: does it affect the HTTP block if it's got other ssl configuration directives set but not `listen ssl`?

ohemorange: the fun thing about nginx is that there's not so much on the documentation and you usually just have to try it to see if it works; do you have a test config setup on a box somewhere?

ohemorange: just tested -- seems to work fine. it might be more aesthetically pleasing to remove the ssl directives, what do you think?

sydneyli: agreed-- it makes more sense not to include ssl directives. just did this-- let me know if i missed anything!

sydneyli:
Done! This PR is getting kinda off-topic-- in hindsight I should have created a new issue for removing the SSL directives from the new HTTP vhost but oh well it's here now \o/

sydneyli:
@ohemorange Thanks for the review and sorry for the delay on this PR-- will work on it today!

You're right-- this solution unfortunately makes a lot of implicit assumptions about how the parser will be run (and only keeps state around for a single session). The correct way to do this would probably be to use the comments-- I remember starting down this path and this PR got even more out of control.

Will roll back the past couple of changes and open a new issue for removing Certbot-managed directives.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +569,39 @@ def _has_certbot_redirect(self, vhost, domain):
...
+        HTTPS block.
+
+        :param str domain: the domain to enable header for.
+        :param str header_substring: String to uniquely identify a header.
+                        e.g. Strict-Transport-Security, Upgrade-Insecure-Requests
+        :returns: Success
+        :raises .errors.PluginError: If no viable HTTPS host can be created or
+            set with header header_substring.
+        """
+        vhost = self.choose_vhost(domain)
```
ohemorange: ack, this is now `self.choose_vhosts(domain)` because I took a while to get back to you.

ohemorange: and this'll return `vhosts` not `vhost`

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +569,39 @@ def _has_certbot_redirect(self, vhost, domain):
...
+        if vhost is None:
```
ohemorange: and this would be an empty list so `if not vhosts`

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -565,6 +569,39 @@ def _has_certbot_redirect(self, vhost, domain):
...
+            raise errors.PluginError(
+                "Unable to find corresponding HTTPS host for enhancement.")
+
+        if vhost.has_header(header_substring):
```
ohemorange: and then probably `for vhost in vhosts` because if `domain` is a wildcard domain we allow multiple vhosts to be chosen

sydneyli:
rebased on wildcard changes! Also, at some point in the past I (incorrectly) removed the spaces in the directives. Now they are back, and can confirm that this added HSTS correctly to one of my personal sites.

sydneyli:
Also it looks like this closes #5296 as well.

ohemorange:
Added "fixes #5296" to the original post for github magic.

PR #5468, opened 2018-01-24
PluginStorage to store variables between invocations.

joohoi:
~The base class for plugins, `certbot.plugins.common.Plugin`~ The base class for Installer plugins `certbot.plugins.common.Installer` now provides functionality of `PluginStorage` to all installer plugins. This allows a plugin to save and retrieve variables in between of invocations.

The on disk storage is basically a JSON file at `config_dir`/`.pluginstorage.json`, usually `/etc/letsencrypt/.pluginstorage.json`. The JSON structure is automatically namespaced using the internal plugin name as a namespace key. Because the actual storage is JSON, the supported data types are: dict, list, tuple, str, unicode, int, long, float, boolean and nonetype.

To add a variable from inside the plugin class:
`self.storage.put("my_variable_name", my_var)`

To fetch a variable from inside the plugin class:
`my_var = self.storage.fetch("my_variable_key")`

The storage state isn't written on disk automatically, but needs to be called:
`self.storage.save()`


certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare_test.py
```diff
@@ -26,15 +28,20 @@ def setUp(self):
         path = os.path.join(self.tempdir, 'file.ini')
         dns_test_common.write({"cloudflare_email": EMAIL, "cloudflare_api_key": API_KEY}, path)
 
+        self.config_dir = tempfile.mkdtemp()
         self.config = mock.MagicMock(cloudflare_credentials=path,
-                                     cloudflare_propagation_seconds=0)  # don't wait during tests
+                                     cloudflare_propagation_seconds=0,
+                                     config_dir=self.config_dir)  # don't wait during tests
```
bmw: nit: The comment about not waiting should be on the previous line.

joohoi:
>What values do we want to set that affect all lineages rather than a specific one?

This could be anything, and I'd like to leave this decision for the installer itself, ie: not tying the storage to a lineage, as I don't see many benefits in that.

> Do we want/expect users to be modifying this JSON object? If not, perhaps we should make it hidden or put it in our working directory.

I think letting users to modify it if they wish could be valuable for some advanced integrations as well as debugging problems in the future.

> This is a pretty minor concern, but is there something easy we can do to make sure third party plugins won't suddenly fail like our own. I expect many of them won't be using this mechanism at least initially and making them understand it and debug their tests for a change that doesn't affect them seems a little unfortunate.

Not an easy way. I looked into this when writing the code, and there sure are ways, but I believe writing exceptions to the actual code just to help with the tests would be an antipattern. The main issue here was that some of the mocked config objects were missing the plugin name variable, which is then used by the pluginstorage when initiating the plugin.

bmw:
> This could be anything, and I'd like to leave this decision for the installer itself, ie: not tying the storage to a lineage, as I don't see many benefits in that.

I think the main benefit of tying the storage to the lineage is that's how most Certbot subcommands currently work where the given configuration options only affects that lineage and it allows the user to have different configurations for their different virtual hosts.

You had a particular use case in mind for this with HSTS right? What was that use case?

> I think letting users to modify it if they wish could be valuable for some advanced integrations as well as debugging problems in the future.

So the user modifiable config files will be `/etc/letsencrypt/cli.ini`, `/etc/letsencrypt/renewal/<lineage>.conf`, and `/etc/letsencrypt/pluginstorage.json`. Values for plugins can go into all three configuration files. To help me understand what you had in mind here, can you give me an idea of the types of things that should go into pluginstorage rather than `<lineage>.conf`? (I understand later in Certbot's execution there's no good way for a plugin to modify its values in `<lineage>.conf` but that's a solvable problem.)

> The main issue here was that some of the mocked config objects were missing the plugin name variable, which is then used by the pluginstorage when initiating the plugin.

My understanding of the problem and your fix of it is the problem wasn't the plugin name, but the lack of a meaningful `config.config_dir` value which is a new attribute for authenticators to be using. It's not a big deal, but I may take a look into different options here just to have a 2nd set of eyes on the problem once I'm on board with the high level design.

I'm sincerely sorry we're having these design conversations now after you've spent time implementing it. I had envisioned this file going into Certbot's work dir and configuration values being global by default seems like a significant deviation from how Certbot usually works to me. I'm not saying I can't be convinced this is the best approach after explaining it to me a little more though.

To try and avoid this in the future, I'll start taking notes during our 1:1s which I'll share with you so you can correct anything I got wrong. We also may want to consider (informal) design docs for larger features, but I'm not really convinced we should require/encourage this yet. If you have any other ideas for things I can do, please let me know!

joohoi:
> I think the main benefit of tying the storage to the lineage is that's how most Certbot subcommands currently work where the given configuration options only affects that lineage and it allows the user to have different configurations for their different virtual hosts.

The plugin name is used only as the root of the namespace. This would give the Installer full control over stored variables, some of which may be global and some per-lineage basis. Lineage based configuration storage would be a `dict()` of different lineages, and their desired variables.

> You had a particular use case in mind for this with HSTS right? What was that use case?

For HSTS the stored values will be on lineage basis. The required variables would be: current maxAge of HSTS, and when it was last increased (the specific incrementation steps and intervals are still undecided), and to ease things, possibly the augeas path of the `VirtualHost` (to avoid going through the somewhat cumbersome selection logic).

For OCSP stapling prefetching the stored variables will be somewhat similar. The last date of successful update and so on.

> Values for plugins can go into all three configuration files. To help me understand what you had in mind here, can you give me an idea of the types of things that should go into pluginstorage rather than <lineage>.conf?

Values of dynamic in nature, or semi-temporary states, like in the examples above. Anything that's usable for only the installer in question. PluginStorage could also be used to keep track of the state of the active security enhancements in the future, easing the implementation of a possible enhancements UI in the long run and so forth. 

Even better example would be the acme-dns plugin I wrote over the last weekend:
for each domain in a lineage, a new set of credentials and a new subdomain is created to the ACME-DNS. These values are mapped to a domain, and a `dict()` of domain-to-credentials-and-subdomains mapping is saved to pluginstorage. This way we can prompt the user to manually add a single CNAME record per domain, and this will be a one-time only prompt. The renewals in the future will then just read the pluginstorage contents, and use the fetched credentials. 

As an quick example, here's the complete `_perform()` method of the acme-dns plugin:
```
    def _perform(self, domain, _validation_name, validation):
        # Try to fetch existing account from pluginstorage
        acc = self.get_account(domain)
        if not acc:
            # Handle new account creation and storage
            acmedns_client = self._get_acmedns_client(self.acmedns_url)
            acc = acmedns_client.register_account()
            acc['acmedns_url'] = self.acmedns_url
            # domain_map also needs to know the URL of the acme-dns server
            # instance as different domains can be using different acme-dns servers
            self.domain_map[domain] = acc
            self.storage.put("domain_map", self.domain_map)
            self.display_instructions(domain, acc)
        else:
            acmedns_client = self._get_acmedns_client(acc['acmedns_url'])
        acmedns_client.update_txt_record(acc, validation)
        self.storage.save()
```

> My understanding of the problem and your fix of it is the problem wasn't the plugin name, but the lack of a meaningful config.config_dir value which is a new attribute for authenticators to be using. It's not a big deal, but I may take a look into different options here just to have a 2nd set of eyes on the problem once I'm on board with the high level design.

You are right, I did reply to this based on my memory, and it served me wrong yet again. These configuration values are not new to Certbot of course, but they are now used in the actual plugin initialization phase. How I see this issue with the existing tests is that now the tests have (albeit forcefully) more real life like config object mocks.

bmw:
Honestly, my main concern here is having another Certbot configuration file for people to worry about. Hiding it like I initially thought we were going to might help, but after thinking about it more, we'd still have state about the user's config that affects Certbot behavior somewhere. This is another place for bugs and things for a user to worry about with what I feel has already become a fairly complicated piece of software.

> The plugin name is used only as the root of the namespace. This would give the Installer full control over stored variables, some of which may be global and some per-lineage basis. Lineage based configuration storage would be a dict() of different lineages, and their desired variables.

What variables would you imagine a plugin wanting to store globally? I can come up with a couple that could be nice, but I'm not sure if the ones I have in mind are useful enough to necessitate this.

How were you planning on giving the plugin the ability to determine the lineage?

> Values of dynamic in nature, or semi-temporary states, like in the examples above. Anything that's usable for only the installer in question. PluginStorage could also be used to keep track of the state of the active security enhancements in the future, easing the implementation of a possible enhancements UI in the long run and so forth.

Is there any reason these values couldn't go in the lineage config?

> Even better example would be the acme-dns plugin I wrote over the last weekend

It might not be possible without some hackery, but any values you add to the config in the authenticator will be saved and restored in the lineage config if they're prefixed with the name of the plugin. What's the major benefit of pluginstorage here?

bmw:
That's a little surprising. Our integration tests measure test coverage as well and it looks like they fell below the minimum percentage. If there's no easy way to add a new test to fix this, you can reduce the percentage [here](https://github.com/certbot/certbot/blob/v0.21.1/tests/boulder-integration.sh#L440).

joohoi:
It was apparently something with the Travis image. The Nginx errored out as it was trying to load error_log from a standard path, so all of our tests failed there, and resulted in to next-to-none coverage.

bmw:
@ohemorange, this PR adds the ability for installers to save JSON data in a hidden file in `/etc/letsencrypt`. The purpose of this is to allow plugins to save things like the last time they updated HSTS age or ways to identify a vhost that should be updated later. These files are not meant to be modified by a user and instead are a way for plugins to persist state between runs.

I'm on board with this at a high level, but before I dig into the details, I wanted to give you a chance to chime in. Would you like to review this at a high or low level? If not, that's fine. If so, please try to do so quickly as Joona is currently building off this and his other PRs.

ohemorange:
Happy to review this at a high level. My main request is that it is extremely, abundantly clear that these files are not for users to touch, and if they do then gosh darnit we are not responsible for what happens.

bmw:
Great! If it's OK with the two of you, I'm going to put off doing a low level review until @ohemorange +1's the high level design so I don't spend time critiquing low level bits that could change and @joohoi doesn't spend time changing them. I'm planning on trying to quickly review this after everyone's on board here.

ohemorange:
Generally happy with the overall design!

Some suggestions:
- Might be cleaner to put `PluginStorage` in its own file
- If we only expect clients to call `put`, `fetch`, and `save`, then we may want to consider underscoring the other ones. I can imagine wanting to reinitialize or reload though, in which case a comment at the top like "the only ones you probably need to worry about are these" could be helpful.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+        :raises .errors.PluginStorageError: when unable to open or read the file
+        """
+
+        self.config = config
+        self.classkey = classkey
+        self.initialized = False
+        self._data = None
+        self.storagepath = None
+
+    def initialize_storage(self):
```
bmw: Is there ever a time we call this function and `os.path.isdir(self.config.config_dir)` not being true isn't an error? Looking at where it's called, we always immediately try to use `self._data` which would otherwise be `None` and raise an error. Perhaps we should remove the `else` branch here.

Similar to Erica's comments, I'm not sure if a user of this class would/should ever call this function, in which case it should be private.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+        :rtype: dict
+
+        :raises .errors.PluginStorageError: when unable to open or read the file
+        """
+
+        self.config = config
+        self.classkey = classkey
+        self.initialized = False
+        self._data = None
+        self.storagepath = None
```
bmw: You may want to consider adding an underscore to the names of variables we don't want people outside the class to use. I think that's all of them here.

I personally don't care too much about this though.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+        if the storage json exists."""
+
+        if hasattr(self.config, "config_dir") and os.path.isdir(self.config.config_dir):
+            self.storagepath = os.path.join(self.config.config_dir, ".pluginstorage.json")
+            self._data = self.load()
+        else:
+            self.storagepath = None
+        self.initialized = True
+
+    def load(self):
```
bmw: Like Erica said, I think we should consider making this private, especially because modifications to the returned dictionary won't be saved and you have to use `put`, `fetch`, and `save`.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+        except IOError as e:
+            errmsg = "Could not read PluginStorage data file: {0} : {1}".format(
+                self.storagepath, str(e))
+            if os.path.isfile(self.storagepath):
+                # Only error out if file exists, but cannot be read
+                logger.error(errmsg)
+                raise errors.PluginStorageError(errmsg)
+        except TypeError:
+            # This happens if storagepath is not set, eg. None
+            raise errors.PluginStorageError("Could not open PluginStorage")
```
bmw: Similar to the `else:` branch in `initialize_storage`, should this ever reasonably happen? If not, I think this should be deleted.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+        try:
+            data = json.loads(filedata)
+        except ValueError:
+            if len(filedata) == 0:
```
bmw: nit: `if not filedata:`

certbot/plugins/common.py
bmw: If you don't want to make it private, please explain why, but I would  make this method return nothing and instead just set `self._data`. If you keep it private, I have no preference about whether or not you make this change.

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+                raise errors.PluginStorageError(errmsg)
+        return data
+
+    def save(self):
+        """Saves PluginStorage content to disk
+
+        :raises .errors.PluginStorageError: when unable to serialize the data
+            or write it to the filesystem
+        """
+        if not self.initialized or not self.storagepath:
```
bmw: If you remove the `else` in `initialize_storage`, you don't need to check `storagepath` here and the error message can be changed to reflect that the user is calling `save` without having first done anything!

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+            serialized = json.dumps(self._data)
+        except TypeError as e:
+            errmsg = "Could not serialize PluginStorage data: {0}".format(
+                str(e))
+            logger.error(errmsg)
+            raise errors.PluginStorageError(errmsg)
+        try:
+            with os.fdopen(os.open(self.storagepath,
+                                   os.O_WRONLY | os.O_CREAT, 0o600), 'w') as fh:
+                fh.truncate()
```
bmw: What is the purpose of `truncate` here?

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+    def put(self, key, value):
+        """Put configuration value to PluginStorage
+
+        :param str key: Key to store the value to
+        :param value: Data to store
+        """
+        if not self.initialized:
+            self.initialize_storage()
+
+        if not self.classkey in self._data.keys():
```
bmw: If you wanted, we could make this one line like:
```
self._data.setdefault(self.classkey, dict())[key] = value
```

I don't have a strong preference though so if you like it better this way, keep it as is!

certbot/plugins/common.py
```diff
@@ -99,7 +100,131 @@ def dest(self, var):
...
+
+        :param str key: Key to get value from the storage
+        """
+        if not self.initialized:
+            self.initialize_storage()
+
+        try:
+            return self._data[self.classkey][key]
+        except KeyError:
+            return None
```
bmw: I'd personally prefer raising an exception here rather than just quietly returning `None`. I think in general making the interface more strict can help prevent bugs and with the current approach you can't differentiate a value set to `None` from a nonexistent key.

certbot/plugins/pluginstorage_test.py
```diff
@@ -0,0 +1,140 @@
...
+                               ".pluginstorage.json"), 'r') as fh:
+            psdata = fh.read()
+        psjson = json.loads(psdata)
+        self.assertTrue("mockplugin" in psjson.keys())
+        self.assertEqual(len(psjson), 1)
+        self.assertEqual(psjson["mockplugin"]["testkey"], "testvalue")
+
+
+
+
```
bmw: Let's remove some of these blank lines.

certbot/errors.py
```diff
@@ -97,6 +101,7 @@ def __init__(self, socket_error, port):
         self.port = port
 
 
+
```
bmw: nit: We don't need three blank lines here.

certbot/plugins/pluginstorage_test.py
```diff
@@ -0,0 +1,140 @@
...
+from certbot.plugins import common
+
+
+class PluginStorageTest(unittest.TestCase):
+    """Test for certbot.plugins.common.PluginStorage"""
+
+    def setUp(self):
+
+        class MockPlugin(common.Installer):
```
bmw: Why not just use `common.Installer`?

certbot/plugins/pluginstorage_test.py
```diff
@@ -0,0 +1,140 @@
...
+import shutil
+import tempfile
+import unittest
+
+from certbot import errors
+
+from certbot.plugins import common
+
+
+class PluginStorageTest(unittest.TestCase):
```
bmw: I think this class should inherit from `certbot.tests.util.ConfigTestCase` and inherit the config and automatic teardown. You'd just want to create the `self.config.config_dir` in `setUp` here and the `tearDown` in `ConfigTestCase` will clean it up.

certbot/plugins/pluginstorage_test.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.config_dir)
+
+    def test_initialize_no_configdir(self):
+        delattr(self.plugin.config, "config_dir")
+        self.plugin.storage.initialize_storage()
+        self.assertTrue(self.plugin.storage.storagepath == None)
+
+    def test_load_errors(self):
```
bmw: nit: It might be nice to split each error here into it's own test method with a nice name. I don't care too much if you make this change or not.

certbot/plugins/pluginstorage_test.py
```diff
@@ -0,0 +1,140 @@
...
+                              "value")
+            self.assertTrue("is corrupted" in mock_log.call_args[0][0])
+
+    def test_save_no_storagepath(self):
+        with mock.patch("certbot.plugins.common.logger.error") as mock_log:
+            self.assertRaises(errors.PluginStorageError,
+                              self.plugin.storage.save)
+            self.assertTrue("Unable to save" in mock_log.call_args[0][0])
+
+    def test_save_errors(self):
```
bmw: Same comment about splitting up testing of different errors.

certbot/plugins/common.py
joohoi: Probably a leftover from a earlier stage of development, deleted as unnecessary.

certbot/plugins/common.py
joohoi: Decided to keep it as it is.

certbot/plugins/common.py
joohoi: This now simply raises KeyError, which should be informative for the developers.

joohoi:
All the comments should now be addressed, I just simply responded with a +1 for every one I implemented as suggested.

joohoi:
The review comments are addressed as proposed, and the files renamed to `storage.py` and `storage_test.py`

PR #5471, opened 2018-01-24
add let's encrypt status to footer and fix link

### LINKED ISSUE #5449: Link to Let's Encrypt status page from documentation
I'm trying to get https://letsencrypt.status.io/ to show up in a Google search for [[letsencrypt status](https://www.google.com/search?q=letsencrypt+status)]. Right now it tries to correct to our "stats" page.

One thing that would help is to get more links to it. Would Certbot be willing to add a link in the footers of the documentation saying "Let's Encrypt status" and linking to the status page? Thanks!

SwartzCr:
resolves #5449


SwartzCr:
The footer template is copy/pasted from the sphinx_rtd_theme - which it overwrites. I've deleted the copyright section so that we can have a hyperlink in the right place and added a link to the let's encrypt status. Feel free to suggest text changes if you think it should be different

PR #5476, opened 2018-01-25
Fix --no-bootstrap on CentOS/RHEL 6

bmw:
If you take certbot-auto 0.21.0 on CentOS 6 and run
```
./certbot-auto --os-packages-only -n
./certbot-auto --no-bootstrap
```
one of two bad things will happen.

If `virtualenv` is installed, it will create the virtual environment using Python 2.6. If `virtualenv` is not installed, it exit saying that it couldn't find the `virtualenv` command.

This PR fixes that by letting the normal bootstrap selection process happen (allowing variables like `USE_PYTHON_3` to be set) and afterwards respecting `--no-bootstrap` overwriting and unsetting variables as necessary.

I think this PR is pretty straightforward, but Erik, if you can get to this PR before Erica, a review would help us out.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -761,13 +761,8 @@ BootstrapMageiaCommon() {
 # Set Bootstrap to the function that installs OS dependencies on this system
 # and BOOTSTRAP_VERSION to the unique identifier for the current version of
 # that function. If Bootstrap is set to a function that doesn't install any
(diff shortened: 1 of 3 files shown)


erikrose:
Sorry, just saw this. Glad you got it merged!

bmw:
@erikrose, no worries! It went up fairly late in the day yesterday. I just tagged you in the hopes you could get to it this morning in case Erica didn't get to it last night.

Issue #4065, opened 2017-01-16
Certbot library usage


jaddison:
Would it be possible to refactor `certbot` to be more of a library+tool than just a tool? Or two packages `python-certbot` and `certbot`, with the latter having the current functionality but using the former at its core?

Referencing #1793, #358, #1299 as semi-related items, although they are more focused on refactoring the `certbot` the tool. I'm interested in a refactor into a libary+tool in one package, or separate packages for each.

I think this would lead to greater uptake from hosting providers (easier implementation), as it becomes much simpler to integrate into a fully automated certificate generation process. 

I created a wrapper around `certbot`[1], but I'd rather see a native Python implementation than running `certbot` in a separate process within Python as I've done. A short conversation on the topic here [2].

[1] https://github.com/jaddison/certbot_py
[2] https://community.letsencrypt.org/t/python-wrapper-around-certbot-auto-0-10-0/25582/2

bmw:
Have you seen our `acme` package?

https://acme-python.readthedocs.io/en/latest/
https://pypi.python.org/pypi/acme/0.10.1

jaddison:
Thanks - I have seen `acme`. As I mention on those referenced pages, it doesn't appear to be as user-friendly as something like a `generate_certificates(domains=domain_list, auth_callback=auth_callback_func, **options)` function, however. (obviously just an example)

As I also mention, it's nice to maintain compatibility with `certbot renew`.

A fair bit of boilerplate is needed and as things change moving forward, an implementation using `acme` could break where if there was a stable interface in `certbot`, the likelihood is lower (but not removed, I will admit). Let's Encrypt engineers can change the code behind the interface without affecting users.

Just putting the thought out there - if it isn't desirable, then by all means, close this issue.

bmw:
No this is definitely desirable and actually something we've talked about internally. Certbot/acme is rarely used by [hosting providers who support Let's Encrypt](https://community.letsencrypt.org/t/web-hosting-who-support-lets-encrypt/6920) and we believe the problems you describe in this issue are largely the reason. These companies need a library and `acme` is too low level.

To be honest, I doubt we'll see this change for a while, but we'll certainly keep it in mind. We're currently working on getting things polished up and increasing stability so we can get Certbot 1.0 out the door. We do, however, have big plans for Certbot 2.0 as we've learned a lot while writing and deploying Certbot to a large user base with a changing protocol. I think this is something important for us to keep in mind then.

cc @ohemorange

[...]

Issue #2947, opened 2016-05-09
Need better python acme client docs/examples


optix2000:
The current documentation for the [Python acme client](http://letsencrypt.readthedocs.io/projects/acme/en/latest/index.html) is fairly lacking. The example client doesn't demonstrate how to set up challenges, and the docs for [`acme.challenges`](http://letsencrypt.readthedocs.io/projects/acme/en/latest/api/challenges.html) also does not give an example on how to set up challenges, nor does it provide enough description on how each class should be used.

This makes it difficult for someone to use the Python acme client to mint their own certs without either having to dig through the code for letsencrypt (certbot) or shelling out to certbot itself. This in turn makes the acme protocol harder to adopt.

Compare this with the example client in for the [Ruby acme client](https://github.com/unixcharles/acme-client/blob/master/README.md), which gives a full working client and an (arguably) easier to use API.
### TLDR

There should be a full end-to-end example for the Python acme client to make it easier to learn.

kuba:
https://github.com/certbot/certbot/blob/master/acme/examples/example_client.py

optix2000:
> The example client doesn't demonstrate how to set up challenges, and the docs for acme.challenges also does not give an example on how to set up challenges

As far as I can tell, the example client you linked is the same on in the docs. It's not a complete example as it doesn't show how to use challenges, nor is it capable of minting certs as it mentions on line 47.

I'm asking for a complete example end-to-end that shows how to use challenges.

kuba:
I suggest you have a look at https://github.com/kuba/simp_le for reference implementation.

bmw:
I agree that we can significantly improve things here so I'm adding this to our wishlist.

bmw:
One thing this example should do is show how to reuse an existing account.

bmw:
I think this has gotten a lot better with PRs like https://github.com/certbot/certbot/pull/5494, but many of the suggestions in this thread have not been done yet.

Ornias1993:
I can definately confirm the examples are still lacking. Primarily in the DNS-01 side of things, which is slowly becoming somewhat of the standard/default due to support for wildcards (which HTTP-01 does not).

Ornias1993:
Needs updates by the devs, not us keeping it open endlessly.

PR #5494, opened 2018-01-26
Add acme library usage example (http-01)

jrodcla:
Here is a usage example of the ACME API exploring HTTP-01 challenge validation using a local Boulder server. This can be used as part of the documentation and is related to https://github.com/certbot/certbot/issues/4065 and https://github.com/certbot/certbot/issues/2947.

Create, edit and deactivate account.
Setup and perform http-01 challenge.
Issue, renew and revoke certificate.

### DIFF
--- a/acme/examples/http01_example.py
+++ b/acme/examples/http01_example.py
@@ -1,0 +1,460 @@
+"""Example ACME API for HTTP-01 challenge.
+
+Copyright 2017 Juliana Rodrigueiro
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+Brief:
+
+This example requires a Boulder CA server running, possibly locally.
+Instructions to set up: https://github.com/letsencrypt/boulder
+Run server with the argument FAKE_DNS, so any domain name will be accepted.
+# docker-compose run -e FAKE_DNS=<client-ip> --service-ports boulder ./start.py
+
+Limitations of this example:
+    - Works for only one Domain name.
+    - Performs only HTTP-01 challenge.
+
+Workflow:
+    - Create account key.
+    - Register account and accept TOS.
+    - Select HTTP-01 within offered challenges by the CA server.
+    - Set up standalone web server.
+    - Create domain private key and CSR.
+    - Issue certificate.
+    - Change contact information
+    - Renew Certificate
+    - Revoke Certificate
+    - Deactivate Account
+"""
+import logging
+import os
+
+from contextlib import contextmanager
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import rsa
+import josepy as jose
+import OpenSSL
+
+from acme import challenges
+from acme import client
+from acme import crypto_util
+from acme import errors
+from acme import fields
+from acme import messages
+from acme import standalone
+
+# Constants:
+
+# IP of the machine running Boulder.
+BOULDER_SERVER_IP = '172.16.1.111'
+
+# Boulder binds by default to port 4000.
+BOULDER_SERVER_PORT = 4000
+
+SERVER_URL = 'http://' + BOULDER_SERVER_IP + ':' + repr(BOULDER_SERVER_PORT)
+DIRECTORY_URL = SERVER_URL + '/directory'
+
+USER_AGENT = 'python-acme-example'
+
+# Account key size
+ACC_KEY_BITS = 2048
+
+# Certificate private key size
+CERT_PKEY_BITS = 2048
+
+# Domain name for the certificate.
+DOMAIN = 'client.example.com'
+
+# The Boulder fake CA tries to connect by default through the port number 5002,
+# so this is where the standalone web server has to bind to. Real CA servers
+# will always use port 80, though.
+PORT = 5002
+
+ACC_FILEPATH = os.path.join('/tmp', DOMAIN + '-account.json')
+REGR_FILEPATH = os.path.join('/tmp', DOMAIN + '-regr.json')
+PKEY_FILEPATH = os.path.join('/tmp', DOMAIN + '-pkey.pem')
+CERT_FILEPATH = os.path.join('/tmp', DOMAIN + '-cert.pem')
+CHAIN_FILEPATH = os.path.join('/tmp', DOMAIN + '-fullchain.pem')
+
+# ACME API can be quite verbose.
+logging.basicConfig(level=logging.INFO)
+
+
+# Useful methods and classes:
+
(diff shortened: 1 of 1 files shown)


cpu:
Hi @jrodcla,

Thanks for submitting this! Two thoughts from a non-Certbot dev doing a drive-by :-)

1. This targets a Boulder instance running on the developer's machine. That's a great setup to have working for long term development aid but it might be a slightly more accessible example if you targeted the Let's Encrypt staging environment. One downside is having to bind port 80 for the challenge request but I think it might be better overall. What do you think?
1. The parts of the example code that persist/load data makes your example closer to a real ACME client but it also lengthens the code a fair amount. I think if you left that code out and provided an example that was in-memory only (perhaps with a note/comment about how you would want to persist this data for a real client) then the example would be more focused on the problem at hand: the acme bits.

ohemorange:
Hey @jrodcla! This is a great example! What do you think about making the changes that @cpu suggested?

ohemorange:
Hi @jrodcla, are you no longer able to work on this?

jrodcla:
Hi!!

Sorry for the delay in replying.

This pull request was not a priority for me, but this week I will finally have time to work on it.

I'll also adapt everything for ACME-v2. Should I create a whole new pull-request or amend this one?

Cheers!

ohemorange:
Thanks! Whatever's easiest for you, either way.

jrodcla:
Here we go. Done.

ohemorange:
Still need to double check the code, but as for the license: if you want to contribute the code under the license for the `acme` package by removing the license at the top, we'd love to have it, but otherwise, we don't accept additional licenses for specific files in the project.

jrodcla:
No worries.

I'd be happy if you could double check the code.

acme/examples/http01_example.py
```diff
@@ -0,0 +1,271 @@
...
+    with challenge_server({resource}):
+
+        # This is the domain used for the local simple_verify. This example
+        # uses a fake domain name in DOMAIN, so this verification would fail
+        # because the DNS would never resolve it to a real IP address. In order
+        # to bypass it and only verify if the standalone server would
+        # successfully reply the correct resource, we use 'localhost' as domain
+        # instead.
+        domain = 'localhost'
+        if not response.simple_verify(challb.chall,
```
ohemorange: What's the purpose of this check? We only use it in testing.

acme/examples/http01_example.py
```diff
@@ -0,0 +1,271 @@
...
+    challb = select_http01_chall(orderr)
+
+    # The certificate is ready to be used in the variable "fullchain_pem".
+    fullchain_pem = perform_http01(client_acme, challb, orderr)
+
+    logging.info('Certificate issued: \n%s', fullchain_pem)
+
+    regr = verify_registration(client_acme, regr)
+
+    # Change contact information
```
ohemorange: Let's reorder so that account updates and certificate updates are grouped together

acme/examples/http01_example.py
```diff
@@ -0,0 +1,271 @@
...
+
+    logging.info('Revoke and Deactivate')
+
+    fullchain_com = jose.ComparableX509(
+        OpenSSL.crypto.load_certificate(
+            OpenSSL.crypto.FILETYPE_PEM, fullchain_pem))
+
+    try:
+        client_acme.revoke(fullchain_com, 0)  # revocation reason = 0
+    except errors.ConflictError:
```
ohemorange: Very helpful to show the errors you get, love that you do this here and in a couple other places!

acme/examples/http01_example.py
jrodcla: Yes, it is a local test. Actually it might just confuse people, so I removed.

adferrand:
With #6668 merged, I move my comment on this PR to make a new example compatible with ACMEv2.

Would it be interested to use `BackwardsCompatibleClientV2` instead of `ClientV2`, to have an example that cover most of the use cases ?

cc @ohemorange @jrodcla @cpu

cpu:
> Would it be interested to use BackwardsCompatibleClientV2 instead of ClientV2, to have an example that cover most of the use cases ?

I defer to you folks here - I don't have a preference :-)

jrodcla:
I believe 'ClientV2' is more pertinent since this piece of code is supposed to be an example for future implementations.

ohemorange:
Agreed, as we're hoping to deprecate ACMEv1 in Certbot code anyway. Plus, new implementations should all use ACMEv2, which is really just spec-compliant ACME (vs pre-spec ACME).

acme/examples/http01_example.py
```diff
@@ -0,0 +1,226 @@
...
+    - Uses ACME-v2
+
+Workflow:
+    - Create account key.
+    - Register account and accept TOS.
+    - Select HTTP-01 within offered challenges by the CA server.
+    - Set up standalone web server.
+    - Create domain private key and CSR.
+    - Issue certificate.
+    - Renew Certificate
```
ohemorange: nit: certificate inconsistently capitalized, should be lowercase

acme/examples/http01_example.py
```diff
@@ -0,0 +1,226 @@
...
+    - Change contact information
+    - Revoke Certificate
```
ohemorange: nit: certificate inconsistently capitalized, should be lowercase

acme/examples/http01_example.py
```diff
@@ -0,0 +1,226 @@
...
+ACC_KEY_BITS = 2048
+
+# Certificate private key size
+CERT_PKEY_BITS = 2048
+
+# Domain name for the certificate.
+DOMAIN = 'client.example.com'
+
+# If you are running Boulder locally, it is possible to configure any port
+# number to execute the challenge, but real CA servers will obviously always
```
ohemorange: Turns out this isn't obvious, so let's replace "obviously" with "will always use port 80, as described in the ACME specification"

acme/examples/http01_example.py
ohemorange: No, I meant change the order to something like this:

(Account actions)
    - Create account key.
    - Register account and accept TOS.
    - Change contact information.
    - Deactivate Account
(Certificate actions)
    - Select HTTP-01 within offered challenges by the CA server.
    - Set up http challenge resource.
    - Set up standalone web server.
    - Create domain private key and CSR.
    - Issue certificate.
    - Renew Certificate
    - Revoke Certificate

But of course then if someone runs the code as is, they'll be deactivating an account they still need...

So maybe something like:
(Account creation)
    - Create account key.
    - Register account and accept TOS.
(Certificate actions)
    - Select HTTP-01 within offered challenges by the CA server.
    - Set up http challenge resource.
    - Set up standalone web server.
    - Create domain private key and CSR.
    - Issue certificate.
    - Renew Certificate
    - Revoke Certificate
(Account update actions)
    - Change contact information.
    - Deactivate Account

jrodcla: Agreed, seems more organized now.,

acme/examples/http01_example.py
jrodcla: Ok, sounds good.

bmw:
@jrodcla, test failures aren't related to your PR.

I will rerun tests here once they're working again. Sorry for the trouble!

adferrand:
Tests relaunched, with success this time.

TheTesla:
Why are two different libraries used to generate public/private key pairs?

Issue #5485, opened 2018-01-25
Certbot keeps adding multiple options-ssl-apache.conf includes


nilsm:
## My operating system is (include version):
Debian 8, Apache 2.4

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto, gives me v0.21.0

## I ran this command and it produced this output:
```
./certbot-auto --force-renew --webroot -d domain.com -w /var/www/domain.com --installer apache
```
I chose option 2, to redirect HTTP to HTTP2.

## Certbot's behavior differed from what I expected because:
Each time I run the command, certbot keeps adding another line to include the ssl options, even though it's already present in the file:

```
Include /etc/letsencrypt/options-ssl-apache.conf'
```

I've now ended up with vhost config with 6 of these include lines, in different places. The file already had the line in once, and certbot added another 5 after consecutively running the command:

```
Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
```

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:

```
<VirtualHost *:443>
        ServerName domain.com
        DocumentRoot /var/www/domain.com
        ErrorLog "/var/log/apache2/error.log"
        CustomLog "/var/log/apache2/access.log" common

	Include /etc/letsencrypt/options-ssl-apache.conf

	<Directory "/var/www/domain.com">
  		AllowOverride All
 	 	Options MultiViews FollowSymLinks
   		Require all granted
	</Directory>
	        Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
Include /etc/letsencrypt/options-ssl-apache.conf
SSLCertificateFile /etc/letsencrypt/live/domain.com/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/domain.com/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
</VirtualHost>
```

nilsm:
Update: even a newly generated SSL vhost has multiple include lines. Here is an example of a site that only had a :80 HTTP vhost set up:

```
<IfModule mod_ssl.c>
<VirtualHost *:443>
        ServerName domain.com
[...]

PR #5498, opened 2018-01-26
Only add Include for TLS configuration if not already there

joohoi:
Currently if Apache plugin is used as an installer for a `VirtualHost` that already is handled by Certbot, the `Include /path/to/our_tls_config.conf` is always appended to the existing `VirtualHost`.

This PR adds a check for the `Include` statement.

Fixes: #5485


joohoi:
Tests added

Issue #5439, opened 2018-01-17
Passing arguments to httpd using apachectl is no longer supported.


jerwinrodriguez:
## My operating system is (include version):
CentOS Linux 7.3.1611

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto

## I ran this command and it produced this output:

certbot --authenticator standalone --installer apache --pre-hook "apachectl -k stop" --post-hook "apachectl -k start"

Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator standalone, Installer apache
Starting new HTTPS connection (1): acme-v01.api.letsencrypt.org
No names were found in your configuration files. Please enter in your domain
name(s) (comma and/or space separated)  (Enter 'c' to cancel): my.domain.name
Running pre-hook command: apachectl -k stop
Output from apachectl:
Passing arguments to httpd using apachectl is no longer supported.
You can only start/stop/restart httpd using this script.
If you want to pass extra arguments to httpd, edit the
/etc/sysconfig/httpd config file.

Obtaining a new certificate
Performing the following challenges:
http-01 challenge for my.domain.name
Waiting for verification...
Cleaning up challenges
Running post-hook command: apachectl -k start
Output from apachectl:
Passing arguments to httpd using apachectl is no longer supported.
You can only start/stop/restart httpd using this script.
If you want to pass extra arguments to httpd, edit the
/etc/sysconfig/httpd config file.

We were unable to find a vhost with a ServerName or Address of my.domain.name.
Which virtual host would you like to choose?
(note: conf files with multiple vhosts are not yet supported)
-------------------------------------------------------------------------------
1: ssl.conf                       |                       | HTTPS | Enabled
-------------------------------------------------------------------------------
Press 1 [enter] to confirm the selection (press 'c' to cancel): 1
Deploying Certificate for my.domain.name to VirtualHost /etc/httpd/conf.d/ssl.conf
Error while running apachectl graceful.

Job for httpd.service invalid.

Rolling back to previous server configuration...
Error while running apachectl graceful.

Job for httpd.service invalid.

Encountered exception during recovery
Error while running apachectl graceful.

Job for httpd.service invalid.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/certbot/error_handler.py", line 100, in _call_registered
    self.funcs[-1]()
  File "/usr/lib/python2.7/site-packages/certbot/client.py", line 557, in _rollback_and_restart
[...]

PR #5500, opened 2018-01-26
If restart fails, try alternative restart command if available

joohoi:
`apachectl` script in some of the distributions does not start Apache if it's stopped. This has existed before, but has been highlighted now with the issues with TLS-SNI challenge. To mitigate these issues we instructed users to use standalone as authenticator, and this required the Apache to be stopped in order to serve the challenge over port 80, so this issue surfaced.

This PR adds a distribution override variable `restart_cmd_alt` which is run in the `restart()` method, if the initial run of `restart_cmd` fails.

Affected:
 - CentOS
 - RHEL
 - Fedora
 - Gentoo

Not affected:
 - Debian
 - Ubuntu
 - Arch Linux
 - OpenSUSE

Fixes: #5439


joohoi:
> If we want to fix this, what if on affected systems we try apachectl restart if graceful fails? We'd want to make sure the behavior of failed graceful commands vs failed restart commands is sane though.

Great idea, I'll implement that in the distribution based overrides of the affected systems.

joohoi:
This is now implemented, and the title and description updated accordingly.

certbot-apache/certbot_apache/override_gentoo.py
```diff
@@ -21,6 +21,7 @@ class GentooConfigurator(configurator.ApacheConfigurator):
         version_cmd=['/usr/sbin/apache2', '-v'],
         apache_cmd="apache2ctl",
         restart_cmd=['apache2ctl', 'graceful'],
+        restart_cmd_alt=['apachectl', 'restart'],
```
bmw: Should this be `apache2ctl` like the other commands?

Issue #4809, opened 2017-06-08
Document when we update docs/cli-help.txt


bmw:
Answer is only during releases but we should put this in our developer guide and/or add a comment to the top of the file that is automatically stripped out of our built docs.

Issue #4124, opened 2017-01-26
Documentation for older Certbot versions


bmw:
As Certbot gets frozen into more and more distros which won't accept updates, we will have people on a variety of Certbot versions. Currently we only have documentation for the most recent version of Certbot and we need to do a better job of documenting older versions of the client.

georgehank:
I just now run into this... CRAP!  That's why I absolutely loathe online-only docs.  EITHER keep older docs (look at Python for how to properly do this... sheesh, it's a SOLVED PROBLEM) or make it easy (and recommend it) to roll up docs into the package.

Yes, I'm angry.  The heat is making it worse, but as I said, I loathe this shit.

SwartzCr:
Hmm this is an interesting topic
I think we currently ship build docs along with Certbot, so people should have a version of the docs that's appropriate for their version on their machine already.
It wouldn't be hard to build old versions of the docs and include them as options to look at on certbot.eff.org/docs, but there are a few issues with that.
First I'd be really surprised if anyone would know to look for them, they're not very discoverable and people would already have to know that they're on a specific version of certbot that isn't the newest one.
Second, we do a very bad job of keeping the documentation up to date. So old documentation might be missing references to commands that that version has, but we failed to document until later.
Additionally @bmw mentioned that we've been putting a lot of work recently into making sure everyone has the most recent certbot versions, and felt that if we just solved that problem then we'd have to worry about this less.

@georgehank I'm sorry about the heat, it's very warm here in San Francisco as well and it's getting quite annoying. If I may ask, what OS are you using certbot on? What was the thing you needed to know that was wrong in the latest documentation? How did you figure it out?

georgehank:
I didn't figure anything out, it still fails, and it's about how DNS auth for *RENEWAL* actually works.  Because it fails for this cert that has two SANs (jaerhard.com and www.jaerhard.com, a pretty standard form I'd say).  I verified that both auth tokens get added to BIND's zone files (and BIND gets reloaded properly too)... but one SAN fails.  Every.  Time.

Then again, I'm sorta stuck on 0.11.1 because the Debian maintainer has any newer (and 0.14 only anyway) stuck in experimental.

Any other previous cert creation and install was *WAY* simpler and easier... and it worked, too!

bmw:
I don't think we're ever going to do this. Instead, I think we're going to take more control of our own packaging and try to keep people on recent versions so I'm closing this issue.

PR #5501, opened 2018-01-26
Add sudo to certbot-auto instructions.

jsha:
This makes it slightly clearer that the user will be prompted for sudo access on running it.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -32,7 +32,7 @@ a new plugin is introduced.
 .. code-block:: shell
 
    cd certbot
-   ./certbot-auto --os-packages-only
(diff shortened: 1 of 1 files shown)


ohemorange:
Going to wait to merge this, because I don't remember the details about how and when these docs are generated.

bmw:
I appreciate you double checking.

The only concerns with modifying a file in `docs/` is:

1. If it's documenting unreleased functionality, we may want to wait to make sure it doesn't make its way to the live site. That happening involves updating the git submodule on https://github.com/certbot/website, but it's easy for someone to not realize doing that's a problem. #4124 is related.
2. `cli-help.txt` is modified by the release script and we shouldn't be touching it manually (see #4809).

Issue #5489, opened 2018-01-26
Remove support for Python 2.6


bmw:
Python 2.6 version of #5312.

Issue #5312, opened 2017-12-08
Remove support for Python 3.3


bmw:
After #5283 ships in 0.21.0, we can remove support for Python 3.3. To do this, we should:

1. Appropriately update `setup.py` so our packages aren't listed as supporting Python 3.3.
2. Remove all Python 3.3 tests.
3. Remove Python 3.3 deprecation warnngs

bmw:
I'm not that familiar with the process, but I believe if we modify `setup.py` right, `pip` is smart enough to notice that a package isn't listed as supporting the Python version and won't try to install it. We should look into this a bit and try to do it as right as we can.

hugovk:
Yes, by adding `python_requires` to `setup()`. This is for 2.7 and 3.4+:

```python
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
```

Links to more info: https://stackoverflow.com/q/44660448/724176

PR #5508, opened 2018-01-27
Drop support for EOL Python 2.6 and 3.3

hugovk:
Fixes https://github.com/certbot/certbot/issues/5489 and fixes https://github.com/certbot/certbot/issues/5312.

---

TODO:

* [x] Deal with `py26_packages` in tox.ini
* [ ] Add `python_requires` to setup.py

---

There's still `py26_packages` and `non_py26_packages` in tox.ini and are installed at different times:

```
[testenv]
commands =
    {[base]install_and_test} {[base]py26_packages}
    python tests/lock_test.py
    {[base]install_and_test} {[base]non_py26_packages}
```

Do they need to be kept separate (and if so, how named?) or can they be merged into `all_packages`?

---

Need to add `python_requires` to setup.py (see https://github.com/certbot/certbot/issues/5312#issuecomment-360972764). Should it be added to all the setup.py files?

```
./certbot-dns-luadns/setup.py
./acme/setup.py
./certbot-dns-rfc2136/setup.py
./certbot-dns-dnsimple/setup.py
./letsencrypt-auto-source/tests/fake-letsencrypt/setup.py
./certbot-dns-cloudflare/setup.py
./certbot-dns-route53/setup.py
./certbot-dns-digitalocean/setup.py
./certbot-compatibility-test/setup.py
./setup.py
./letshelp-certbot/setup.py
./certbot-dns-cloudxns/setup.py
./examples/plugins/setup.py
./certbot-nginx/setup.py
./certbot-dns-nsone/setup.py
./certbot-apache/setup.py
./certbot-dns-google/setup.py
./certbot-dns-dnsmadeeasy/setup.py
```


hugovk:
image

I _think_ the TOXENV=le_auto_trusty job is failing because fetch.py is downloading a file from the repo and comparing it to this PR's modified version:

```
Root *-auto have unexpected changes.
```

certbot-auto
```diff
@@ -781,20 +781,11 @@ elif [ -f /etc/redhat-release ]; then
   prev_le_python="$LE_PYTHON"
   unset LE_PYTHON
   DeterminePythonVersion "NOCRASH"
-  if [ "$PYVER" -eq 26 ]; then
```
bmw: I don't think these lines should be changed. Unfortunately, certbot-auto has become pretty convoluted and difficult to work on over the years, but what's going on here is the code first determines your Python version without looking for Python 3. On RHEL 6 based systems, this will usually only find Python 2.6 which we no longer support and this branch will install Certbot using Python 3 from EPEL 6. The other branch installs Certbot using Python 2.7.

certbot-auto
```diff
@@ -965,18 +956,10 @@ if [ "$1" = "--le-auto-phase2" ]; then
     say "Creating virtual environment..."
     DeterminePythonVersion
     rm -rf "$VENV_PATH"
-    if [ "$PYVER" -le 27 ]; then
```
bmw: I don't think these lines should be changed because we still need the `virtualenv` version for Python 2.7 support.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -4,7 +4,7 @@
 Embed this in your project, and your VCS checkout is all you have to trust. In
 a post-peep era, this lets you claw your way to a hash-checking version of pip,
 with which you can install the rest of your dependencies safely. All it assumes
-is Python 2.6 or better and *some* version of pip already installed. If
+is Python 2.7 or better and *some* version of pip already installed. If
```
bmw: So this file is taken from https://github.com/erikrose/pipstrap and I'd like to keep it in sync with what's in that repo. We should revert these changes for now. If you'd like to make this change, I'd create a PR against that repo and if it lands we can copy it here, although Erik may want to keep Python 2.6 support in pipstrap.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -64,41 +64,7 @@ iQIDAQAB
 -----END PUBLIC KEY-----
 "
 
-if [ $(python -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//') -eq 26 ]; then
```
bmw: We should keep these lines as they are testing our behavior of installing Python 3 in certbot-auto on systems that still use Python 2.6.

.travis.yml
```diff
@@ -58,16 +50,6 @@ matrix:
       env: TOXENV=nginxroundtrip
 
 
-# Only build pushes to the master branch, PRs, and branches beginning with
```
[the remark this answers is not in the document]
hugovk: Sure, restored.

It is a pain for contributors though, it means you cannot test your feature branch on the CI *before* creating a PR. I prefer not to submit a PR before it's green. Ideally these restrictions would not apply to forks, but this can be left for now.

acme/acme/__init__.py
```diff
@@ -13,7 +13,7 @@
 import sys
 import warnings
 
-for (major, minor) in [(2, 6), (3, 3)]:
+for (major, minor) in [(3, 3)]:
```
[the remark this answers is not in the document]
hugovk: Removed!

hugovk:
CI still not green, but it's something else now. I'll be away from a computer for a couple of weeks, so can check back then. Otherwise feel free to modify this, the "Allow edits from maintainers" checkbox is checked.

.travis.yml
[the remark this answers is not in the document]
bmw: That's fair. If there's a way to do this in a way that doesn't affect forks, I'd happily accept another PR containing these changes.

bmw:
We want this for our release in the next couple weeks so I'll probably build off this to get it merged, but thanks so much for the help here.

bmw:
@erikrose, that'd fine with me. Do you think we should set `python_requires` like this in general to . go along with us dropping support for Python 2.6 and 3.3?

erikrose:
I don't see how it could hurt, apart from the aforementioned potential setuptools incompatibilities in our supported range. Somebody could set up a quick tox job to try all the intervening setuptools versions just to make sure no fatal errors pop out.

bmw:
> Somebody could set up a quick tox job to try all the intervening setuptools versions just to make sure no fatal errors pop out.

While I didn't do this in `tox`, I did this in a simple script and other than 12.0.2 where there was an [AttributeError problem in setuptools](https://setuptools.readthedocs.io/en/latest/history.html#id321) that prevented it from installing packages, every version worked just fine.

erikrose:
Wow, I'm impressed with your thoroughness!

bmw:
These changes landed in #5605. Thanks again for the help here @hugovk.

hugovk:
You're welcome, and thanks for finishing it off!

PR #5531, opened 2018-02-04
Added install-only flag

sydneyli:
Here's a first pass at #4741 ! Let me know if I did anything weird.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -47,6 +47,7 @@ Help for certbot itself cannot be provided until it is installed.
   --no-bootstrap                            do not install OS dependencies
   --no-self-upgrade                         do not download updates
   --os-packages-only                        install OS dependencies and exit
+  --install-only                            install certbot, upgrade if needed, and exit
   -v, --verbose                             provide more output
   -q, --quiet                               provide only update/error output;
                                             implies --non-interactive
@@ -60,6 +61,8 @@ for arg in "$@" ; do
       DEBUG=1;;
     --os-packages-only)
       OS_PACKAGES_ONLY=1;;
+    --install-only)
+      INSTALL_ONLY=1;;
     --no-self-upgrade)
       # Do not upgrade this script (also prevents client upgrades, because each
       # copy of the script pins a hash of the python client)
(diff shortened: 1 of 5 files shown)


schoen:
I think the functionality is fine, but I'm a bit worried that users will get confused about "install Certbot only" vs. "install a certificate only", since we have the existing verb `certonly` which means "only run an authenticator, don't run an installer". Users who understand our terminology might then wrongly think that `--install-only` means "only run an installer, don't run an authenticator".

Also, this is intended for use in conjunction with `certbot-auto` (because it still allows the autoupgrader to run its initial passes), right? It won't have any benefit for people who've installed from an OS package, I assume?

sydneyli:
I forgot to link the original issue: #4741 
Yeah, it's intended to be used with `certbot-auto`, and i think the wording means to mirror `os-deps-only`. I think we can make that clearer in the help text.

sydneyli:
Cool, thanks @bmw for the guidance! now `--install-only` lives entirely in the `auto` script.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -1428,6 +1431,12 @@ UNLIKELY_EOF
 
     say "Installation succeeded."
   fi
+
+  if ["$INSTALL_ONLY" = 1 ]; then
```
bmw: So I'm looking into why the letsencrypt-auto Travis test we run for PRs passed, but you need a space between `[` and `"$INSTALL_ONLY"`. If you run the following script:
```
#!/bin/sh

if ["$INSTALL_ONLY" = 1 ]; then
    exit 0
fi
```
You'll get output like:
```
./test.sh: line 3: [: =: unary operator expected
```

bmw: Well that's mildly terrifying. We have `set -e` in `letsencrypt-auto` which means to exit if any commands fail. Apparently this comparison isn't considered a failure, but it never seems to be true. It just always prints a warning and skips the body of the `if` statement.

Since this is new functionality and everything else in `letsencrypt-auto` works the same (sometimes with new error output), our tests pass.

sydneyli: my bad-- had it changed locally and forgot to commit. i'll open a separate issue for the Travis issue!

Issue #5188, opened 2017-10-13
Fix the install verb


pde:
This shouldn't happen:

```
certbot install --cert-name example.com
usage: 
  certbot [SUBCOMMAND] [options] [-d DOMAIN] [-d DOMAIN] ...

Certbot can obtain and install HTTPS/TLS/SSL certificates.  By default,
it will attempt to use a webserver both for obtaining and installing the
certificate. 
certbot: error: argument --cert-path is required
```

We've been putting this off because `certbot` or `certbot run` basically works fine as an installer. But with wildcards becoming available via dns plugins, we might want to give instructions like:

```
certbot certonly -a some-dns-plugin 
certbot install -i nginx --cert-name x
```
or similar.

pde:
This may be an 0.21 or 0.22 ticket, but putting it in the next milestone for now.

Issue #1930, opened 2015-12-17
install verb demands --cert-path even if it has --fullchain-path


pde:
cli.py errors out if given --fullchain-path and --key-path without --cert-path:

```
sudo venv/bin/letsencrypt install  --key-path /etc/letsencrypt/live/isnot.org/privkey.pem --fullchain-path /etc/letsencrypt/live/isnot.org/fullchain.pem --hsts --uir --redirect --apache -d isnot.org,is.isnot.org
usage: 
  letsencrypt [SUBCOMMAND] [options] [-d domain] [-d domain] ...

The Let's Encrypt agent can obtain and install HTTPS/TLS/SSL certificates.  By
default, it will attempt to use a webserver both for obtaining and installing
the cert. Major SUBCOMMANDS are:

  (default) run        Obtain & install a cert in your current webserver
  certonly             Obtain cert, but do not install it (aka "auth")
  install              Install a previously obtained cert in a server
  revoke               Revoke a previously obtained certificate
  rollback             Rollback server configuration changes made during install
  config_changes       Show changes made to server config during installation
  plugins              Display information about installed plugins
letsencrypt: error: argument --cert-path is required
```

jpluimers:
Of these permutations, only the last one works:

```
# ./certbot-auto install --noninteractive --apache --domains continuaci.4delphi.com
# ./certbot-auto install --noninteractive --apache --domains continuaci.4delphi.com --cert-path /etc/letsencrypt/live/continuaci.4delphi.com/fullchain.pem
# ./certbot-auto install --noninteractive --apache --domains continuaci.4delphi.com --cert-path /etc/letsencrypt/live/continuaci.4delphi.com/fullchain.pem --key-path /etc/letsencrypt/live/continuaci.4delphi.com/privkey.pem
# ./certbot-auto install --noninteractive --apache --domains continuaci.4delphi.com --fullchain-path /etc/letsencrypt/live/pcm.4delphi.com/fullchain.pem
# ./certbot-auto install --noninteractive --apache --domains continuaci.4delphi.com --cert-path /etc/letsencrypt/live/continuaci.4delphi.com/fullchain.pem --fullchain-path /etc/letsencrypt/live/pcm.4delphi.com/fullchain.pem
[...]

PR #5536, opened 2018-02-05
Fix install verb

joohoi:
Let user to define a certificate managed by Certbot with `--cert-name` instead of providing direct paths to `--cert-path` and `--key-path`.

If `--cert-name` and either or both of `--cert-path` and `--key-path` are provided on command line, the more precise parameters take precedence.

Unfortunately main.py is growing out of hand, and could use another round of splitting, I had to raise the `.pylintrc` line count limits.

Fixes: #5188

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -806,6 +806,8 @@ def _populate_from_certname(config):
         config.namespace.key_path = lineage.key_path
     if not config.cert_path:
         config.namespace.cert_path = lineage.cert_path
+    if not config.chain_path:
+        config.namespace.chain_path = lineage.chain_path
     if not config.fullchain_path:
         config.namespace.fullchain_path = lineage.fullchain_path
     return config
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -1,4 +1,5 @@
 """Certbot main entry point."""
+# pylint: disable=too-many-lines
 from __future__ import print_function
 import functools
 import logging.handlers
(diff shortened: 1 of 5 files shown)


certbot/main.py
```diff
@@ -779,11 +779,45 @@ def install(config, plugins):
...
+    if config.key_path and config.cert_path:
+        _check_certificate_and_key(config)
+        domains, _ = _find_domains_or_certname(config, installer)
+        le_client = _init_le_client(config, authenticator=None, installer=installer)
+        _install_cert(config, le_client, domains)
+    else:
+        raise errors.ConfigurationError("Path to certificate or key was not defined. "
+            "If your certificate is managed by Certbot, please use --cert-name "
+            "to define which certificate you would like to install. "
+            "Alternatively you can use both --key-path and --cert-path to install a "
```
pde: Normally people would want to use `--fullchain-path` in addition to `--cert-path`, right?

pde: (See also #1930)

pde: It might even be best to not document that alternative way of using `--install`, because it's so weird, confusing and broken?

pde:
Note that in addition to #5188 there's also #1391, which this half fixes (though it can probably be another PR to do the matching from `--domains`).

certbot/main.py
pde: In summary, either add `--fullchain-path` to this CLI documentation, or remove the sentence starting "Alternatively..."?

joohoi: I decided to remove the later part of the message to keep the message short and concise. And there certainly is broken parts in the later execution regarding manually defining all the different values.

certbot/main.py
```diff
@@ -779,11 +780,43 @@ def install(config, plugins):
...
+    defined by --cert-name."""
+
+    lineage = cert_manager.lineage_for_certname(config, config.certname)
+    if not lineage:
+        return config
+    if not config.key_path:
+        config.namespace.key_path = lineage.key_path
+    if not config.cert_path:
+        config.namespace.cert_path = lineage.cert_path
+    if not config.fullchain_path:
```
bmw: Should we do this for `--chain-path` too? This is needed some servers such as [Apache < 2.4.8](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L316).

PR #5542, opened 2018-02-06
Make Certbot depend on josepy

### LINKED ISSUE #5519: Certbot should depend on josepy
As part of the transition of the JOSE library out of acme into josepy, certbot switched to directly importing from josepy -- however, no dependency on it was added in setup.py.  Even though it should always be co-installed by virtue of the transient dependency from acme, if certbot is directly using it, it should be required for install.

bmw:
Fixes #5519.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ install_requires = [
     'ConfigArgParse>=0.9.3',
     'configobj',
     'cryptography>=1.2',  # load_pem_x509_certificate
+    'josepy',
     'mock',
     'parsedatetime>=1.3',  # Calendar.parseDT
     'pyrfc3339',


setup.py
```diff
@@ -41,6 +41,7 @@ def read_file(filename, encoding='utf8'):
     'ConfigArgParse>=0.9.3',
     'configobj',
     'cryptography>=1.2',  # load_pem_x509_certificate
+    'josepy',
```
erikrose: Looks good, but it seems to me we should also pin <2.0.

bmw: I thought we were in agreement in #4433 not to pin major versions back like this and instead do so in `venv.sh`?

erikrose: Ah, yes, I forgot about that. r+

PR #5554, opened 2018-02-07
Add mechanism to detect acme version

### LINKED ISSUE #5491: Add automatic ACME protocol detection mechanism to the acme library
This should work by querying the `directory` endpoint of the server and looking for new fields and/or the new names for existing fields to correctly determine the protocol version that should be used.

This should be done in a way where we store the value of the `directory` so `acme.client.Client` doesn't have to request it again. This should probably be accomplished by adding a new optional parameter to the ACME client which is the value of the directory.

There are many options on the return value here. The current plan is to have two different ACME client classes for each protocol (see #5287). With this in mind, the return value could be int/enum and directory value, ACME client type and directory, initialized ACME client of the right type, etc. I'm fine with any approach, but we have to live with the changes we make here for a long time so they should be designed carefully.

ohemorange:
Fixes #5491.

Detects acme version by checking for `newNonce` field in the directory, since it's mandatory. Also updates `ClientNetwork.account` on `register` and `update_registration`.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -576,20 +576,20 @@ class BackwardsCompatibleClientV2(object):
 
     def new_account_and_tos(self, regr=None, tos_cb=None):
         def assess_tos(tos):
-            if tos_cb is not None and not tos_cb(regr.terms_of_service):
+            if tos_cb is not None and not tos_cb(tos):
                 raise errors.Error(
                     "Registration cannot proceed without accepting "
                     "Terms of Service.")
         if self.acme_version == 1:
-            # if tos in directory
-            # else if tos not in directory
             regr = self.client.register(regr)
             if regr.terms_of_service is not None:
                 assess_tos(regr.terms_of_service)
                 return self.client.agree_to_tos(regr)
         else:
-            assess_tos(self.directory['termsOfService'])
-            regr.update(terms_of_service_agreed=True)
+            assert regr is not None
+            if "terms_of_service_v2" in self.directory.meta:
+                assess_tos(self.directory.meta.terms_of_service_v2)
+                regr.update(terms_of_service_agreed=True)
             return self.client.new_account(regr)
 
     def _acme_version_from_directory(self, directory):
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -559,6 +559,12 @@ class ClientV2(ClientBase):
 
 
 class BackwardsCompatibleClientV2(object):
+    """ACME client wrapper that tends towards V2-style calls, but
+       supports V1 servers.
+
+       :ivar int acme_version:
+       :ivar .ClientBase client: either Client or ClientV2
+    """
 
     def __init__(self, net, key, server):
         directory = messages.Directory.from_json(net.get(server).json())
@@ -580,19 +586,24 @@ class BackwardsCompatibleClientV2(object):
             raise AttributeError
 
     def new_account_and_tos(self, regr=None, check_tos_cb=None):
-        # check_tos_cb should raise an error if we want to error
-        def assess_tos(tos):
+        """Combined register and agree_tos for V1, new_account for V2
+
+        :param .NewRegistration new_account:
+        :param function check_tos_cb: callback that raises an error if
+            the check does not work
+        """
+        def _assess_tos(tos):
             if check_tos_cb is not None:
                 check_tos_cb(tos)
         if self.acme_version == 1:
             regr = self.client.register(regr)
             if regr.terms_of_service is not None:
-                assess_tos(regr.terms_of_service)
+                _assess_tos(regr.terms_of_service)
                 return self.client.agree_to_tos(regr)
         else:
             assert regr is not None
             if "terms_of_service_v2" in self.client.directory.meta:
-                assess_tos(self.client.directory.meta.terms_of_service_v2)
+                _assess_tos(self.client.directory.meta.terms_of_service_v2)
                 regr.update(terms_of_service_agreed=True)
             return self.client.new_account(regr)
 
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -562,7 +562,7 @@ class BackwardsCompatibleClientV2(object):
     """ACME client wrapper that tends towards V2-style calls, but
        supports V1 servers.
 
-       :ivar int acme_version:
+       :ivar int acme_version: 1 or 2, corresponding to the Let's Encrypt endpoint
        :ivar .ClientBase client: either Client or ClientV2
     """
 
@@ -575,7 +575,7 @@ class BackwardsCompatibleClientV2(object):
             self.client = ClientV2(directory, net=net)
 
     def __getattr__(self, name):
-        if name in vars(self.client).keys():
+        if name in vars(self.client):
             return getattr(self.client, name)
         elif name in dir(ClientBase):
             return getattr(self.client, name)
@@ -585,11 +585,11 @@ class BackwardsCompatibleClientV2(object):
         else:
             raise AttributeError
 
-    def new_account_and_tos(self, regr=None, check_tos_cb=None):
+    def new_account_and_tos(self, regr, check_tos_cb=None):
         """Combined register and agree_tos for V1, new_account for V2
 
-        :param .NewRegistration new_account:
-        :param function check_tos_cb: callback that raises an error if
+        :param .NewRegistration regr:
+        :param callable check_tos_cb: callback that raises an error if
             the check does not work
         """
         def _assess_tos(tos):
@@ -601,10 +601,9 @@ class BackwardsCompatibleClientV2(object):
                 _assess_tos(regr.terms_of_service)
                 return self.client.agree_to_tos(regr)
         else:
-            assert regr is not None
             if "terms_of_service_v2" in self.client.directory.meta:
                 _assess_tos(self.client.directory.meta.terms_of_service_v2)
-                regr.update(terms_of_service_agreed=True)
+                regr = regr.update(terms_of_service_agreed=True)
             return self.client.new_account(regr)
 
     def _acme_version_from_directory(self, directory):
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -95,7 +95,6 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         update = regr.body if update is None else update
         body = messages.UpdateRegistration(**dict(update))
         updated_regr = self._send_recv_regr(regr, body=body)
-        self.net.account = updated_regr
         return updated_regr
 
     def deactivate_registration(self, regr):
@@ -556,9 +555,25 @@ class ClientV2(ClientBase):
             acme_version=2)
         # "Instance of 'Field' has no key/contact member" bug:
         # pylint: disable=no-member
-        regr = self._regr_from_response(response)
-        self.net.account = regr
-        return regr
+        return self._regr_from_response(response)
+
+
+class MultiVersionClient(object):
+
+    def __init__(self, net, key, server):
+        directory = messages.Directory.from_json(net.get(server).json())
+        acme_version = self._acme_version_from_directory(directory)
+        if acme_version == 1:
+            self.client = Client(directory, key=key, net=net)
+        else:
+            self.client = ClientV2(directory, net=net)
+
+    def _acme_version_from_directory(self, directory):
+        try:
+            nonce_field = directory['newNonce']
+        except KeyError:
+            return 1
+        return 2
 
 
 class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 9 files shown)


jsha:
FYI, tests are failing.

ohemorange:
yes, I haven't updated tests yet, wanted to put up a rough draft of how to implement this to see what people think.

bmw:
This became a long post so try and bear with me, but it describes defining four methods (other than `__init__`) which would allow us to handle all of the ACMEv1 vs. ACMEv2 differences that affect Certbot that I've identified.

I agree with Jacob that a class with all methods where the user still has to check the ACME version and call the right method probably doesn't buy us much. This could mean we skip my 2nd post of #5491 and just do the first, but if we wanted to try and do the class that tries to handle these differences, I'd do something like this:

Make this class be a simple wrapper for all methods in `ClientBase`. I'd probably do this through something like:
```
def __getattr__(self, name):
    if name in dir(ClientBase):
        return getattr(self.client, name)
    else:
        raise AttributeError
```
and document this behavior in the docstring for the class.

For the methods that try and handle the differences, we first add a method like `request_challenges` or `request_authorizations` that takes either a CSR or list of domains and returns a list of authorizations. A CSR is only necessary if `new_order` takes a CSR. For ACMEv1, it takes the domains (possibly after extracting them from the CSR) and calls `request_domain_challenges` for each domain and returns a list of obtained authz. For ACMEv2, this calls `new_order` with either the CSR/domains and returns the authorizations in the Order. This order object is stored as an attribute of the class to be handled internally

Next we add a method like `request_issuance` which takes a CSR and returns the cert and its chain. For ACMEv1, it forwards the arguments to `request_issuance` and then retries calling `fetch_chain` (see `certbot.client.obtain_certificate_from_csr`). For ACMEv2 it adds the CSR to the internal order object (if necessary) and calls `poll_order_and_request_issuance`. We then need to split the certificate and chain, probably by looking for the first line containing "END CERTIFICATE". If the `request_challenges`/`request_authorizations` method takes a CSR, we could keep this as an attribute of the class too and remove the parameter here if we really wanted.

The last piece of this is the ugliest. We need a method like `register`/`new_account` that also allows the user of the class to prompt the user to agree to the TOS. This method takes a `NewRegistration` object and a callback function. Luckily for us, we already use a callback function for agreeing to terms of service (see `certbot.client.register` and `certbot.main._determine_account`). If we're on ACMEv1, we call `register` with the `NewRegistration` object, pass the TOS into the callback which should either return a bool or raise an exception if the user doesn't agree, and then call `agree_to_tos`. For ACMEv2, we get the TOS out of the directory (perhaps by modifying `ClientV2` to have a property/method that does this for us), calls the callback with the TOS, sets `termsOfServiceAgreed` on the `NewRegistration`, and calls `new_account`. This method returns the up to date `Registration` object.

jsha:
Broadly: I'm in favor of a wrapper client (`EasyClient`) that abstracts away some of the details of each API and provides a very simple "I want a certificate, please get me one" API, which I think is what @bmw's proposing. I think the MultiVersionClient in this PR doesn't get us there, since you have to choose which methods to call based on what API version you're using.

>  then need to split the certificate and chain, probably by looking for the first line containing "END CERTIFICATE".

Certbot should actually parse the PEM chunks and ensure they are of the correct type (CERTIFICATE) before passing them along to the filesystem.

> We need a method like register/new_account that also allows the user of the class to prompt the user to agree to the TOS.

In ACMEv1, Boulder already provides `"meta": {"terms-of-service": "https://...."}`. In ACMEv2, this is called `termsOfService`, but we can provide a simple shim to pick whichever is present. So our new `EasyClient` would have a method to get the terms of service URL before you register, in order to present it to the user. There would be another method, `register`, that would either perform the two-step dance from ACMEv1 or the one-step dance from ACMEv2.

There's a bit of a corner case for the people who have done private non-Boulder implementations of ACMEv1, in case they don't provide `"meta": {"terms-of-service": "..."}`, but we can reach out to them; I think there are only a couple.

bmw:
> There's a bit of a corner case for the people who have done private non-Boulder implementations of ACMEv1, in case they don't provide "meta": {"terms-of-service": "..."}, but we can reach out to them; I think there are only a couple.

My only concern here is if we don't know who they all are. I know I don't.

jsha:
> My only concern here is if we don't know who they all are. I know I don't.

I know a couple, which are probably the only ones. If we miss one or two, having them add a field to the directory should be pretty easy.

bmw:
> If we miss one or two, having them add a field to the directory should be pretty easy.

I suppose and our version is technically still 0.x.y, but I hate sneaking in breaking changes like this. We've essentially treated Certbot as 1.0+ since we launched which means we shouldn't be making breaking changes without changing the version number (which we unfortunately can't do without claiming 1.0). If we were actually 1.0+, I'd come down hard that we should not be making a breaking change.

Another possibility if people really don't like the callback I proposed above is to have Certbot use the MultiversionClient slightly differently as necessary to maintain compatibility here. If there is no terms of service in the directory, Certbot could use `multiversion_client.client.{register,agree_to_tos}` or something.

I'm happy to chat about this more, but I'm also fine with Erica making the call about this minor detail.

ohemorange:
@jsha re: "Also, I'd prefer to do autodetection on `newAccount`."

`new_account` won't necessarily be called, if the account already exists from a previous run. In that case, we'll still need to know which version we're using. No?

ohemorange:
(please don't review yet, I just like github's comparison UI)

ohemorange:
Discussed offline: leaving as detecting `newNonce` for now, unless the acmev2 spec is updated so that `newAccount` is required.

This is now ready for review; I'd ideally like to get this in piece-by-piece so it's not blocked on the v2 orders PR, and add functionality to switch over as we go.

ohemorange:
Coverage is failing but I'd rather get that in after a round of review

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
         return self._regr_from_response(response)
 
 
+class BackwardsCompatibleClientV2(object):
+    """ACME client wrapper that tends towards V2-style calls, but
+       supports V1 servers.
+
+       :ivar int acme_version:
```
bmw: nit: You should write a small description here. Probably pointing out it's either 1 or 2 and these numbers correspond to the LE endpoint. (In terms of the ACME spec, v1 and v2 aren't accurate. We're on version 9 of the unfinished spec now.)

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+        # temporary, for breaking changes into smaller pieces
+        elif name in dir(Client):
+            return getattr(self.client, name)
+        else:
+            raise AttributeError
+
+    def new_account_and_tos(self, regr=None, check_tos_cb=None):
+        """Combined register and agree_tos for V1, new_account for V2
+
+        :param .NewRegistration new_account:
```
bmw: nit: The name of the param above is regr.

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+    def __init__(self, net, key, server):
+        directory = messages.Directory.from_json(net.get(server).json())
+        self.acme_version = self._acme_version_from_directory(directory)
+        if self.acme_version == 1:
+            self.client = Client(directory, key=key, net=net)
+        else:
+            self.client = ClientV2(directory, net=net)
+
+    def __getattr__(self, name):
+        if name in vars(self.client).keys():
```
bmw: Nice catch and solution for making instance variables available.

nit: You don't need to call `keys()` here.

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+        elif name in dir(Client):
+            return getattr(self.client, name)
+        else:
+            raise AttributeError
+
+    def new_account_and_tos(self, regr=None, check_tos_cb=None):
+        """Combined register and agree_tos for V1, new_account for V2
+
+        :param .NewRegistration new_account:
+        :param function check_tos_cb: callback that raises an error if
```
bmw: nit: Sphinx doesn't understand function but it should understand `callable`.

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+            if check_tos_cb is not None:
+                check_tos_cb(tos)
+        if self.acme_version == 1:
+            regr = self.client.register(regr)
+            if regr.terms_of_service is not None:
+                _assess_tos(regr.terms_of_service)
+                return self.client.agree_to_tos(regr)
+        else:
+            assert regr is not None
+            if "terms_of_service_v2" in self.client.directory.meta:
```
bmw: TODO: I think it might be nice for ClientV2 to provide a terms of service attribute/method so it contains the info about how to get the TOS rather forcing the user of the class to have that knowledge, but this is totally fine for now.

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+        def _assess_tos(tos):
+            if check_tos_cb is not None:
+                check_tos_cb(tos)
+        if self.acme_version == 1:
+            regr = self.client.register(regr)
+            if regr.terms_of_service is not None:
+                _assess_tos(regr.terms_of_service)
+                return self.client.agree_to_tos(regr)
+        else:
+            assert regr is not None
```
bmw: TODO: We should provide a somewhat meaningful error message here and probably not use assert since they can be disabled on the Python command line. This is also fine for now though.

acme/acme/client.py
```diff
@@ -558,6 +558,62 @@ def new_account(self, new_account):
...
+            if "terms_of_service_v2" in self.client.directory.meta:
+                _assess_tos(self.client.directory.meta.terms_of_service_v2)
+                regr.update(terms_of_service_agreed=True)
```
bmw: I believe you need to assign the return value of update to `regr` here since these objects are immutable and update creates a new instance of the object.

acme/acme/messages.py
```diff
@@ -172,6 +172,7 @@ class Directory(jose.JSONDeSerializable):
     class Meta(jose.JSONObjectWithFields):
         """Directory Meta."""
         terms_of_service = jose.Field('terms-of-service', omitempty=True)
+        terms_of_service_v2 = jose.Field('termsOfService', omitempty=True)
```
bmw: This can be something we delay until closer until the release if you want, but I wonder if we shouldn't try and make the `terms_of_service` field automatically handle the difference in naming conventions between v1 and v2 for the user of the class and remove `terms_of_service_v2`. We did this for challenges in #5300 which was more complicated since we also send challenges back to the server which isn't something we do for directories.

acme/acme/client.py
ohemorange: nice

acme/acme/messages.py
ohemorange: since only the directory fields in `Meta` are explicitly defined, the one class works for both; I'd rather not have to have `Directory` also be aware of acme version if possible.

bmw: I just think it's a chance to keep the API cleaner/simpler and push details further down so the higher levels don't have to worry about them. If you didn't already, take a look at the approach in #5300 at a high level (protected fields with a property that just returns whichever value was set by the server). I personally favor that approach, but if you don't we can leave this as is.

acme/acme/client.py
ohemorange: better idea, I'm going to make `regr` not optional

ohemorange: we are aiming for V2-style

acme/acme/client.py
ohemorange: going to skip for now because nowhere else in cb accesses the tos

acme/acme/client_test.py
```diff
@@ -21,10 +21,25 @@
...
+        'https://www.letsencrypt-demo.org/acme/revoke-cert',
+    messages.NewAuthorization:
+        'https://www.letsencrypt-demo.org/acme/new-authz',
+    messages.CertificateRequest:
+        'https://www.letsencrypt-demo.org/acme/new-cert',
+})
+
+DIRECTORY_V2 = messages.Directory({
+    'newAccount': 'https://www.letsencrypt-demo.org/acme/new-account',
+    'newNonce': 'https://acme-staging-v02.api.letsencrypt.org/acme/new-nonce'
```
bmw: nit: Can we use a `letsencrypt-demo` URL here like we do for the other endpoints?

acme/acme/client_test.py
```diff
@@ -84,6 +84,127 @@ def setUp(self):
         # Reason code for revocation
         self.rsn = 1
 
+
+class BackwardsCompatibleClientV2Test(ClientTestBase):
+    """Tests for  acme.client.BackwardsCompatibleClientV2."""
+
+    def setUp(self):
```
bmw: nit: If all `setUp` is doing is calling `setUp` for the parent class, you can omit it entirely.

ohemorange: refactoring is fun

acme/acme/client_test.py
ohemorange: sure

ohemorange:
> I don't really think we should spend time doing this now, but in the future or when you're writing tests for new methods on this class, I personally think it's acceptable (and more like a unit test of this class) to just mock out the client object that this is wrapping and assert the right calls are made.

This is what I tried to do; what are you referring to?

bmw:
> This is what I tried to do; what are you referring to?

Oh you're right. I only reviewed your new commits and was surprised to see you were using mock responses, but I forgot you have to fetch the directory yourself. Sorry for the noise!

PR #5557, opened 2018-02-08
Change "Attempting to parse" warning to info.

jsha:
This message shows up on every renewal run when the config was updated
by a newer version of Certbot than the one being run. For instance, if a
user has the certbot packages installed from PPA (currently 0.18.2), but
runs certbot-auto once to try out the latest version (0.21.1), they will
start getting this message via email every 12 hours.


bmw:
Looks like a test that's expecting `warning` to be called is failing.

PR #5576, opened 2018-02-15
Postfix plugin changes

sydneyli:
- Finishing refactor of postconf/postfix command-line utilities
- Plugin uses starttls_policy to specify per-domain policies

More detailed writeup of changes incoming.

### DIFF
--- a/certbot-postfix/certbot_postfix/installer.py
+++ b/certbot-postfix/certbot_postfix/installer.py
@@ -14,7 +14,7 @@ from certbot.plugins import util as plugins_util
 from certbot_postfix import util
 from certbot_postfix import postconf
 
-import starttls_policy
+# import starttls_policy
 
 POLICY_FILENAME = "starttls_everywhere_policy"
 
@@ -48,6 +48,19 @@ DEFAULT_CLIENT_VARS = {
 }
 
 logger = logging.getLogger(__name__)
+
+def _report_master_overrides_less_secure(acceptable):
+    def _report_master_overrides(name, overrides):
+        for override in overrides:
+            if override not in acceptable:
+                print("Warning: Parameter %s is overridden as less-secure option %s " + 
+                      "for service %s in master configuration file!" %
+                      (name, override[1], override[0]))
+
+def _report_master_overrides(name, overrides):
+    for override in overrides:
+        print("Warning: Parameter %s is overridden as %s for service %s in master configuration file!" %
+                (name, override[1], override[0]))
 
 @zope.interface.implementer(interfaces.IInstaller)
 @zope.interface.provider(interfaces.IPluginFactory)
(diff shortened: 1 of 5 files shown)


sydneyli:
@jgillula 
## Postfix plugin
So there are two main things that this Installer does:
 1. Configures TLS Postfix parameters in `main.cf`. The parameters for this come [recommended](http://www.postfix.org/TLS_README.html) for a balance of security & compatibility.
 2. Write the per-domain TLS policies from `starttls-policy` into a [format that Postfix understands](http://www.postfix.org/postconf.5.html#smtp_tls_policy_maps).

The code for [1] is in [`deploy_cert`](https://github.com/certbot/certbot/pull/5576/files#diff-62ce583f47d3952d57797e5bf0a01da2L159), which
  - Sets the key/cert for `smtpd`, the MTA server (mail receiver),
  - Sets reasonable TLS params for both the MTA server and client.
  - Sets the path to the policy file (from [2]) 

The code for [2] is in [`write_domainwise_tls_policies`](https://github.com/certbot/certbot/pull/5576/files#diff-62ce583f47d3952d57797e5bf0a01da2R130), which loops through the domains
and [writes their policy](http://www.postfix.org/postconf.5.html#smtp_tls_policy_maps) to the appropriate file.

## Other stuff
The other large portion of diffs in this pull request pertain to refactoring a lot of the code around running command-line utilities like `postfix` and `postconf`. Brad started this refactor in `postconf.py` and I finished it up (Look at all this [deleted code!](https://github.com/certbot/certbot/pull/5576/files#diff-62ce583f47d3952d57797e5bf0a01da2L228)), and also added tests in `postconf_test.py`. There's also wrappers around the `postfix` command-line util in `util.py`. As a result, I also moved a lot of the tests from `installer_test.py` into `postconf_test.py` (since the code for running configuration edits was no longer in `installer.py`.

## Things that are left
There are always more tests to write, and I have to write an integration test for the per-domain policy things, so I can catch regressions easily. The most pressing thing I'm working on right now is standardizing and thoroughly documenting the API for fetching policies in the `starttls-policy` package.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -44,15 +55,14 @@ class Installer(plugins_common.Installer):
     """Certbot installer plugin for Postfix.
 
     :ivar str config_dir: Postfix configuration directory to modify
-    :ivar dict proposed_changes: configuration parameters and values to
-        be written to the Postfix config when save() is called
     :ivar list save_notes: documentation for proposed changes. This is
         cleared and stored in Certbot checkpoints when save() is called
-
+    :ivar postconf: Wrapper for Postfix configuration command-line tool.
+    :ivar policy: A Policy object using which we can query
```
jgillula: Word missing here? A Policy object using what? (Or maybe "using" should be deleted.)

PR #5578, opened 2018-02-15
Add order support and tests

bmw:
Built on #5374. Fixes a couple bugs and adds tests.

One thing I noticed here which I didn't fix is we don't extract the common name from the CSR. We should do this before the release, especially since users can provide their own CSR on the command line.

EDIT: This PR should be merged rather than squashed to preserve authorship.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -1,6 +1,7 @@
 """ACME client API."""
 import base64
 import collections
+import cryptography
 import datetime
 from email.utils import parsedate_tz
 import heapq
@@ -119,11 +120,11 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         """
         return self._send_recv_regr(regr, messages.UpdateRegistration())
 
-    def _authzr_from_response(self, response, identifier, uri=None):
+    def _authzr_from_response(self, response, identifier=None, uri=None):
         authzr = messages.AuthorizationResource(
             body=messages.Authorization.from_json(response.json()),
             uri=response.headers.get('Location', uri))
-        if authzr.body.identifier != identifier:
+        if identifier is not None and authzr.body.identifier != identifier:
             raise errors.UnexpectedUpdate(authzr)
         return authzr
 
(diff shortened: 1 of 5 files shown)


acme/acme/client.py
```diff
@@ -609,33 +620,42 @@ def poll_authorizations(self, orderr, deadline):
...
             response = self.net.get(orderr.uri)
             body = messages.Order.from_json(response.json())
             if body.error is not None:
-                raise IssuanceError(body.error)
+                raise errors.IssuanceError(body.error)
             if body.certificate is not None:
                 certificate_response = self.net.get(body.certificate).text
-                return orderr.update(fullchain_pem=certificate_response)
-        raise TimeoutError()
+                return orderr.update(body=body, fullchain_pem=certificate_response)
```
ohemorange: Finalize doesn't change anything in `body`?

bmw: Doesn't it? Earlier in the function, we post to the `finalize` endpoint and begin repeatedly making GET requests for the order until the `certificate` field is updated.

bmw:
> As I mentioned OOB, this duplicates #5334 only you called it CommonClientTest and I called it ClientTestBase, oops.
>
> They're pretty much identical though so it shouldn't be a problem?

Yeah it'll be alright. There'll be some small merge conflicts, but it's nbd.

acme/acme/client.py
ohemorange: Read this the other way around, thought you were removing body.

ohemorange:
Cool, I'd like to double check the merge then, but otherwise looks good!

bmw:
Oh you want to wait until we merge your PR and have me address merge conflicts here? That's fine. I can fix them.

bmw:
Since there are multiple authors here, I'll want to fix up the commit history though so you won't have a clean commit to look at. Would you like me to push the non-rebased/squashed version somewhere for you to see?

ohemorange:
I don't think it will be a problem, as long as the eventual PR is based on master.

bmw:
PR rebased on master.

Issue #5431, opened 2018-01-15
Link to webroot documentation from error messages


jsha:
If you follow the autogenerated instructions to use the webroot plugin, you're likely to wind up with this error in your Certbot output:

```
Attempting to renew cert (www.hoffman-andrews.com) from /etc/letsencrypt/renewal/www.hoffman-andrews.com.conf produced an unexpected error: Missing command line flag or config entry for this setting
:
Select the webroot for www.hoffman-andrews.com:
Choices: ['Enter a new webroot']

(You can set this with the --webroot-path flag). Skipping.
The following certs could not be renewed:
  /etc/letsencrypt/live/crud.net/fullchain.pem (failure)
  /etc/letsencrypt/live/jacob.hoffman-andrews.com-0002/fullchain.pem (failure)
  /etc/letsencrypt/live/hoffman-andrews.com/fullchain.pem (failure)
  /etc/letsencrypt/live/www.hoffman-andrews.com/fullchain.pem (failure)
```

This error doesn't define "webroot", and webroot is not a well-defined term (the first page of Google results does not explain it). We shold amend these error messages to link to https://certbot.eff.org/docs/using.html#webroot. We should also update that documentation to define what "webroot" means in the first sentence. Right now it's defined in a parenthetical at the end of the first paragraph:

> In addition, you’ll need to specify --webroot-path or -w with the top-level directory (“web root”) containing the files served by your webserver. For example, --webroot-path /var/www/html or --webroot-path /usr/share/nginx/html are two common webroot paths.

klebervirgilio:
@jsha something like that: https://github.com/certbot/certbot/pull/5583

MarkMura:
Please also add to the documentation, that it's spelled `--webroot-path` in the CLI, but `webroot_path` in the config file

PR #5583, opened 2018-02-16
Add webroot url to warn message

klebervirgilio:
trying to address this issue #5431

### DIFF
--- a/certbot/constants.py
+++ b/certbot/constants.py
@@ -108,6 +108,7 @@ CLI_DEFAULTS = dict(
 
 )
 STAGING_URI = "https://acme-staging.api.letsencrypt.org/directory"
+WEBROOT_URL = "https://certbot.eff.org/docs/using.html#webroot"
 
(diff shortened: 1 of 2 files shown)


jsha:
Thanks for working on this! I think this would link to the webroot docs for *all* renewal failures. We want to add this link only when there is a problem specific to the webroot plugin.

If you check #5431, you'll see that the message I'm talking about is actually a prompt provided when the `--webroot` flag is not provided. You should look for where that prompt is generated, and try to add the URL to that prompt. Once you've done that, try to run Certbot such that it generates the updated prompt, and paste the output here so we can see that it's expected. Thanks!

jgillula:
Hi @klebervirgilio ! Since we haven't seen any activity on this PR for a while, I just wanted to check and see if you still planned to work on it, or if you'd be OK with someone else picking up the work.

bmw:
Since this isn't quite what we're looking for and we haven't seen any updates in over a month, I'm closing this PR, but please open a new one if you keep working on it.

PR #5585, opened 2018-02-16
Return str rather than bytes

mlaferrera:
Project id is returned as bytes, which causes issues when constructing the google cloud API url, converting `b'PROJECT_ID'` to `b%27PROJECT_ID%27` causing the request to fail.


mlaferrera:
@bmw ahh, sorry about the tests. I can fix them...not a problem.

mlaferrera:
The tests should be fixed now. It looks like they were returning an `int` rather than a `str` or `bytes`. I added a test to be able to handle both `str` and `bytes` to ensure things don't break for both py2 and py3.

PR #5586, opened 2018-02-16
Import x509 module

bmw:
```
$ python -c 'import cryptography; cryptography.x509'
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'x509'
```
Like all of our files in `acme` or `certbot`, `x509` is it's own module that needs to be imported. I think the reason this currently works in this file is one of our dependencies is importing the module for us. Let's not rely on this behavior.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -1,7 +1,6 @@
 """ACME client API."""
 import base64
 import collections
-import cryptography
 import datetime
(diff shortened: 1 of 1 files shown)


ohemorange:
No need, I'm removing dependence on `cryptography` in this class.

PR #5587, opened 2018-02-16
Get common name from CSR in new_order in ClientV2

### LINKED ISSUE #5579: Get common name from CSR in new_order in ClientV2
https://github.com/certbot/certbot/pull/5578#issue-169507815.

We have code in `certbot.crypto_util` that does this with the help of some ACME code. The code in `new_order` for extracting SANs is nicer than the code in `acme.crypto_util._pyopenssl_cert_or_req_san`. What I'd do is make (a) function(s) in `acme.crypto_util` that takes a string CSR and uses the SANs extracting code from `new_order` with the common name extracting code from `certbot.crypto_util` and returns the list of domains. If we're currently removing duplicates, we should do this here as well.

ohemorange:
Fixes #5579.


certbot/crypto_util.py
```diff
@@ -340,14 +340,7 @@ def _get_names_from_cert_or_req(cert_or_req, load_func, typ):
...
 def _get_names_from_loaded_cert_or_req(loaded_cert_or_req):
-    common_name = loaded_cert_or_req.get_subject().CN
-    # pylint: disable=protected-access
-    sans = acme_crypto_util._pyopenssl_cert_or_req_san(loaded_cert_or_req)
-
-    if common_name is None:
-        return sans
-    else:
-        return [common_name] + [d for d in sans if d != common_name]
+    return acme_crypto_util._pyopenssl_cert_or_req_all_names(loaded_cert_or_req)
```
bmw: We'll still need to disable protected access here.

acme/acme/crypto_util.py
```diff
@@ -186,6 +186,16 @@ def make_csr(private_key_pem, domains, must_staple=False):
     return OpenSSL.crypto.dump_certificate_request(
         OpenSSL.crypto.FILETYPE_PEM, csr)
 
+def _pyopenssl_cert_or_req_all_names(loaded_cert_or_req):
+    common_name = loaded_cert_or_req.get_subject().CN
+    # pylint: disable=protected-access
```
bmw: nit: I don't think this is needed because it's in the same file.

acme/acme/client.py
```diff
@@ -17,6 +17,7 @@
 import requests
```
bmw: I think cryptography is an unused import now that lint will yell about.

certbot/crypto_util.py
ohemorange: yes, sorry I should have said this wasn't ready for review yet, already fixed

acme/acme/client.py
ohemorange: also already fixed sorry

Issue #5580, opened 2018-02-16
Automatically handle ACMEv1 vs. ACMEv2 issuance


ohemorange:
Part of #5367.

From https://github.com/certbot/certbot/pull/5554#issuecomment-364621848: 

- [x] ~For the methods that try and handle the differences, we first add a method like `request_challenges` or `request_authorizations` that takes either a CSR or list of domains and returns a list of authorizations. A CSR is only necessary if `new_order` takes a CSR. For ACMEv1, it takes the domains (possibly after extracting them from the CSR) and calls `request_domain_challenges` for each domain and returns a list of obtained authzs. For ACMEv2, this calls `new_order` with either the CSR/domains and returns the authorizations in the Order. This order object is stored as an attribute of the class to be handled internally.~ Use `new_order` and a dummy `OrderResource` object.

- [ ] Next we add a method like `request_issuance` which takes a CSR and returns the cert and its chain. For ACMEv1, it forwards the arguments to `request_issuance` and then retries calling `fetch_chain` (see `certbot.client.obtain_certificate_from_csr`). For ACMEv2 it adds the CSR to the internal order object (if necessary) and calls `poll_order_and_request_issuance`. We then need to split the certificate and chain, probably by looking for the first line containing "END CERTIFICATE". If the `request_challenges`/`request_authorizations` method takes a CSR, we could keep this as an attribute of the class too and remove the parameter here if we really wanted.

bmw:
With `new_order` taking a CSR, we need to extract the names from it for ACMEv1. I had some thoughts on how to write a function we could use in multiple places that does this in #5579.

bmw:
Another potential design of this could be to create dummy Order objects in the ACMEv1 case.

ohemorange:
Planning to have `BackwardsCompatibleClientV2` only accept CSRs for its `request_challenges/request_authorizations` function. It will extract domains internally for V1 ACME servers.

bmw:
Yeah that works great.

I think it may be worth considering using dummy order objects here. You could have `new_order` and `finalize_order`. This is better keeping with your `BackwardsCompatibleClientV2` name and means you don't have to somewhat awkwardly store Order objects internally.

I don't have any strong preferences though.

ohemorange:
`answer_challenge` is still a thing, too.

PR #5588, opened 2018-02-17
Support new_order-style in Certbot

ohemorange:
Part of #5580.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -325,8 +325,8 @@ class Client(object):
 
         if successful_domains != domains:
             if not self.config.dry_run:
-                # TODO: delete keys
-                pass
+                os.remove(key.file)
+                os.remove(csr.file)
             return self.obtain_certificate(successful_domains)
         else:
             certr, chain = self.obtain_certificate_from_csr(csr, orderr)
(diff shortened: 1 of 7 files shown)


ohemorange:
Currently adding tests for the new code paths. This should otherwise be ready for review. Also added support for `best_effort`.

certbot/client.py
```diff
@@ -326,10 +318,20 @@ def obtain_certificate(self, domains):
...
             csr = crypto_util.init_save_csr(key, domains, self.config.csr_dir)
 
-        certr, chain = self.obtain_certificate_from_csr(
-            domains, csr, authzr=authzr)
+        orderr = self.acme.new_order(csr.data)
+        authzr = self.auth_handler.handle_authorizations(orderr, self.config.allow_subset_of_names)
+        auth_domains = set(a.body.identifier.value for a in authzr)
+        successful_domains = [d for d in domains if d in auth_domains]
+
+        if successful_domains != domains:
```
bmw: I don't think you can compare set and list objects like this.

bmw:
LGTM if tests pass.

certbot/client.py
ohemorange: I made the same mistake when testing with ipdb.

PR #5591, opened 2018-02-17
Fix domain name recognition

NickMitin:
Domains with underscore are also valid. So I modified enforce_le_validity to mark those to be valid.

### DIFF
--- a/certbot/util.py
+++ b/certbot/util.py
@@ -520,10 +520,10 @@ def enforce_le_validity(domain):
 
     """
     domain = enforce_domain_sanity(domain)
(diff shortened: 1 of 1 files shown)


bmw:
Actually, Let's Encrypt does not allow domains with underscores. See https://github.com/letsencrypt/boulder/issues/1225#issuecomment-175963980.

Certbot will try and issue for domains like this if provided by user, but our plugins won't suggest them as they will not work unless you're using another ACME CA. With this in mind, I'm closing this PR, but please comment if you'd like me to reconsider.

Issue #5551, opened 2018-02-07
Should install/enhance take --domains?


bmw:
Proposed for the install subcommand by Peter at https://github.com/certbot/certbot/pull/5536#issuecomment-363185524. The enhance subcommand works in a similar way by making use of an existing cert so Joona is considering this behavior for that command as well.

Do we want to do this? Should the behavior be identical?

pde:
My default argument would be that all of Certbot's verbs should allow things to be specified in the same ways. Since `certonly` and `run` do this, it will be less confusing if all the verbs do. Though I could also see the case for erroring out if there are multiple plausible (valid, matching) lineages, and asking the user to provide a disambiguating `--cert-name` in that case.

ohemorange:
If this takes any level of annoying work or refactoring, I don't think it's worth it because:
a) our subcommands already don't match each other on the specifiers they take
b) it makes sense for certonly/run to take domains because you need domains to request a cert, but if an operation is actually being done on a cert then it should take the cert, and not try to figure out which cert you actually meant based on domains. this is especially true if we're using the domains to find a cert, then updating the list of domains with the list of domains in the cert

PR #5596, opened 2018-02-19
Enhance verb

joohoi:
Enhance verb to configure security enhancements after installation. Small modifications were required to be able to pass custom dialog messages through the plumbing. 

`enhance()` method for Apache plugin had to be slightly changed to avoid creating a dummy SSL VirtualHost in case the domain didn't have one already. In this situation an error message is printed for the user.


certbot-apache/certbot_apache/configurator.py
```diff
@@ -1377,7 +1377,20 @@ def enhance(self, domain, enhancement, options=None):
...
-            func(self.choose_vhost(domain), options)
+            # Temp value tells choose_vhost to not to create a new SSL vhost if
+            # one is not found.
+            vhost = self.choose_vhost(domain, temp=True)
+            if not vhost.ssl:
+                msg_tmpl = ("Certbot was not able to find SSL VirtualHost for a "
+                       "domain {0} for enabling enhancement \"{1}\". The requested "
+                       "enhancement was not configured.")
+                if options:
+                    msg = msg_tmpl.format(domain, options)
```
bmw: I think whether `options` is a better description than `enhancement` depends on the enhancement type. If it's HSTS, `options` is better, if it's OCSP stapling, `enhancement` is better. What if when `options` is set, we put both `enhancement` and `options` in the error message?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1377,7 +1377,20 @@ def enhance(self, domain, enhancement, options=None):
...
+                else:
+                    msg = msg_tmpl.format(domain, enhancement)
+                logger.warning(msg)
```
bmw: Why did you choose to just log a message here rather than raising an exception?

certbot/main.py
```diff
@@ -860,6 +862,37 @@ def plugins_cmd(config, plugins):
...
+    :rtype: None
+
+    """
+    supported_enhancements = ["hsts", "redirect", "staple", "uir"]
+    # Check that at least one enhancement was requested on command line
+    if not any([getattr(config, enh) for enh in supported_enhancements]):
+        msg = ("Please specify one or more enhancement types to configure. To list "
+               "the available enhancement types, run:\n\n%s --help enhance\n")
+        logger.warning(msg, sys.argv[0])
+        raise errors.MisconfigurationError("No enhancements requested, exiting.")
```
bmw: Why do this here instead of letting `enhance_config` prompt the user so this command works interactively?

certbot/main.py
```diff
@@ -860,6 +862,37 @@ def plugins_cmd(config, plugins):
...
+
+    try:
+        installer, _ = plug_sel.choose_configurator_plugins(config, plugins, "enhance")
+    except errors.PluginSelectionError as e:
+        return str(e)
+    le_client = _init_le_client(config, authenticator=None, installer=installer)
+    domain_question = ("Which domain names would you like to enable the selected "
+                       "enhancements for")
+    domains, _ = _find_domains_or_certname(config, installer, domain_question)
+    le_client.enhance_config(domains, None, ask_redirect=False)
```
bmw: `certbot enable --nginx --staple-ocsp` won't work without the chain path. I think this should be fixed before we ship this. I'd even be fine with us being unable to modify configs we're not managing a cert for, at least for now.

If this makes it significantly harder to support `--domains`, I'd suggest not doing this now. We can always do it later when we have spare cycles and/or there's a demand for it.

certbot/plugins/selection.py
```diff
@@ -160,6 +161,9 @@ def choose_configurator_plugins(config, plugins, verb):
 
     req_auth, req_inst = cli_plugin_requests(config)
 
+    installer_question = ("How would you like to authenticate and install "
```
bmw: Is it correct to mention authentication here?

certbot-apache/certbot_apache/configurator.py
joohoi: This sounds good. For the current enhancements the case was that options was only used for `ensure-http-header` enhancement, where `options` would hold the name of the actual header being added. Printing them both is probably the most informative way to handle this.

certbot-apache/certbot_apache/configurator.py
joohoi: There could be multiple domains to get enhanced, printing a clear error message tells the user that one of them has to be get fixed, but we'll be able to handle the rest of them.

certbot/main.py
joohoi: I wish to have that at later point. However I think that we should spend some time to plan out the interactive UI for the enhancement verb later on. It's easy to add a simple one, but there's a lot we could do with few upcoming features.

certbot/plugins/selection.py
joohoi: Oh, right. It's not :)

certbot/main.py
joohoi: We might want to skip it for now. I'll enable it when I'll be working on the prefetch-OCSP functionality.

joohoi:
The added commits brings this branch up to `master`, and makes `enhance` to behave as discussed and implemented in https://github.com/certbot/certbot/pull/5608#pullrequestreview-99937605 if no HTTPS vhosts were found. I also changed the `choose_vhost` method keyword argument to match with one in `choose_vhosts` for consistency.

certbot/main.py
bmw: I asked @ohemorange about this because I think she has a stronger opinion here than I do and she has a strong preference for us only taking `--cert-name`. Her main argument was we should limit the number of ways there are to accomplish a task in Certbot. This makes both the code and documentation simpler.

If we take `--domains`, we'll need a way to get the chain path in the future. With the way this code is now, if you request `--staple-ocsp`, it will error and say that `--chain-path` needs to be provided, but even if you provide it, we'll still error because it's not being passed through to `enhance_config`. We could fix that by providing the value of `chain_path` from the `config`, but we might instead want to try and find the matching certificate for the given domains. Doing this opens a massive cans of worms which I don't think we want.

I don't have that strong of an opinion, so you and Erica can debate it if you want, but I'll say my preference would be to err on the side of providing fewer bells and whistles initially by not supporting `--domains` since support can always be added later if there is significant demand/need for the feature.

certbot/cli.py
```diff
@@ -1276,7 +1288,7 @@ def _paths_parser(helpful):
         verb = helpful.help_arg
 
     cph = "Path to where certificate is saved (with auth --csr), installed from, or revoked."
-    section = ["paths", "install", "revoke", "certonly", "manage"]
+    section = ["paths", "install", "revoke", "certonly", "manage", "enhance"]
```
bmw: Why do we want `--cert-path` in the `enhance` section?

bmw:
Sorry for the strange review (actually two different reviews and [a comment](https://github.com/certbot/certbot/pull/5596#discussion_r176591690) on an old review and neither review had a high level comment like what I'm writing here), but I think we're pretty close. I think the biggest thing is the question about taking `--domains`, but once we've agreed on a design, it shouldn't be too hard to implement.

joohoi:
The PR comments have now been addressed, and `--cert-name` is now mandatory. If it's missing we're presenting an interactive dialog to choose one. After this step, the user is asked about which domains in the lineage they would like to apply the enhancements for.

certbot/main.py
```diff
@@ -859,6 +861,45 @@ def plugins_cmd(config, plugins):
...
+    except errors.PluginSelectionError as e:
+        return str(e)
+
+    if not config.certname:
+        config.certname = cert_manager.get_certnames(config, "enhance",
+                                                     allow_multiple=False)[0]
+    cert_domains = cert_manager.domains_for_certname(config, config.certname)
+    domain_question = ("Which domain names would you like to enable the selected "
+                       "enhancements for?")
+    domains = display_ops.choose_values(cert_domains, domain_question)
```
bmw: `IDisplay` has become horribly complex over the years (see #3915), but this call causes `IDisplay.checklist` to be called with `force_interactive=True`. I don't think this is what you want here as rather than requiring interactivity, I believe [the plan](https://github.com/certbot/certbot/issues/5551#issuecomment-377004693) was to use all the domains in the certificate when run non-interactively. Currently if you run this with `--non-interactive`, you get:
```
MissingCommandlineFlag: Missing command line flag or config entry for this setting:
Which domain names would you like to enable the selected enhancements for?
```
I think you want to remove the `force_interactive` parameter (or set it to `False`) and include `default` set to `cert_domains` in the call to `IDisplay.checklist`. This may mean you don't add the generic `choose_values` function or you add these values as additional parameters to the function.

certbot/main.py
```diff
@@ -859,6 +861,45 @@ def plugins_cmd(config, plugins):
...
+    if not domains:
+        # To be consistent with the error messages (similar to get_certnames)
+        raise errors.Error("User ended interaction.")
```
bmw: If you keep the `choose_values` function, I think it may make sense to include this in that function. While some of the other functions in `certbot.display.ops` don't do that, I believe something similar is usually being done later by the caller since Certbot almost always errors out when you cancel a prompt so we might as well include it in the generic function.

certbot/main.py
```diff
@@ -859,6 +861,45 @@ def plugins_cmd(config, plugins):
...
+               "the available enhancement types, run:\n\n%s --help enhance\n")
+        logger.warning(msg, sys.argv[0])
+        raise errors.MisconfigurationError("No enhancements requested, exiting.")
+
+    try:
+        installer, _ = plug_sel.choose_configurator_plugins(config, plugins, "enhance")
+    except errors.PluginSelectionError as e:
+        return str(e)
+
+    if not config.certname:
```
bmw: nit: I think we can remove this `if` as the same thing is done inside of `get_certnames`.

certbot/main.py
```diff
@@ -859,6 +861,45 @@ def plugins_cmd(config, plugins):
...
+        config.certname = cert_manager.get_certnames(config, "enhance",
```
bmw: This causes the prompt to be "Which certificate would you like to enhance?". Perhaps it should be something like "Which certificate would you like to use to enhance your configuration?".

This may mean we may want to change the name of the `verb` parameter to `get_certnames`.

certbot/main.py
```diff
@@ -859,6 +861,45 @@ def plugins_cmd(config, plugins):
...
+                                                     allow_multiple=False)[0]
+    cert_domains = cert_manager.domains_for_certname(config, config.certname)
+    domain_question = ("Which domain names would you like to enable the selected "
+                       "enhancements for?")
+    domains = display_ops.choose_values(cert_domains, domain_question)
+    if not domains:
+        # To be consistent with the error messages (similar to get_certnames)
+        raise errors.Error("User ended interaction.")
+    le_client = _init_le_client(config, authenticator=None, installer=installer)
+    le_client.enhance_config(domains, config.chain_path, ask_redirect=False)
```
bmw: `config.chain_path` isn't set unless it's included on the command line. I think we need to use `certbot.cert_manager.lineage_for_certname` to get the `RenewableCert` matching the cert name and use its `chain_path` property.

It may be worth adding logic to handle the case where `--chain-path` is also set.

certbot/main.py
joohoi: It's now using `cert_manager.domains_for_certname` if noninteractive mode is set.

certbot/main.py
joohoi: For the generic `choose_values` I prefer to leave it to the calling functions responsibility to handle the erroring out. That said, this message is now more informative.

certbot/main.py
joohoi: `get_certnames` now takes an optional kwarg if the prompt message needs to be changed completely.

certbot/main.py
joohoi: The `chain_path` from lineage is used if the user didn't provide `--chain-path` on CLI. The command line value overrides the one from lineage.

joohoi:
Ready for the next round :)

certbot/main.py
```diff
@@ -887,16 +887,23 @@ def enhance(config, plugins):
...
                        "enhancements for?")
-    domains = display_ops.choose_values(cert_domains, domain_question)
+    if config.noninteractive_mode:
+        domains = cert_manager.domains_for_certname(config, config.certname)
+    else:
+        domains = display_ops.choose_values(cert_domains, domain_question)
     if not domains:
-        # To be consistent with the error messages (similar to get_certnames)
-        raise errors.Error("User ended interaction.")
+        raise errors.Error("No domains found to enhance.")
```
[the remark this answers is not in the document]
joohoi: Not really, no. This is now changed.

joohoi:
The missing tests are now in

certbot/tests/display/ops_test.py
```diff
@@ -481,5 +490,42 @@ def test_directory_select_validation_with_bad_default(self, mock_util):
...
+    """Test choose_values."""
+    @classmethod
+    def _call(cls, values, question):
+        from certbot.display.ops import choose_values
+        return choose_values(values, question)
+
+    @test_util.patch_get_utility("certbot.display.ops.z_util")
+    def test_choose_names_success(self, mock_util):
+        items = ["first", "second", "third"]
+        mock_util().checklist.return_value = ('ok', [items[2]])
```
bmw: We should use `display_util.OK` here instead of hardcoding this value.

certbot/tests/display/ops_test.py
```diff
@@ -481,5 +490,42 @@ def test_directory_select_validation_with_bad_default(self, mock_util):
...
+        result = self._call(items, None)
+        self.assertEquals(result, [items[2]])
+        self.assertTrue(mock_util().checklist.called)
+        self.assertEquals(mock_util().checklist.call_args[0][0], None)
+
+    @test_util.patch_get_utility("certbot.display.ops.z_util")
+    def test_choose_names_success_question(self, mock_util):
+        items = ["first", "second", "third"]
+        question = "Which one?"
+        mock_util().checklist.return_value = ('ok', [items[1]])
```
bmw: We should use `display_util.OK` here instead of hardcoding this value.

certbot/tests/display/ops_test.py
```diff
@@ -481,5 +490,42 @@ def test_directory_select_validation_with_bad_default(self, mock_util):
...
+        result = self._call(items, question)
+        self.assertEquals(result, [items[1]])
+        self.assertTrue(mock_util().checklist.called)
+        self.assertEquals(mock_util().checklist.call_args[0][0], question)
+
+    @test_util.patch_get_utility("certbot.display.ops.z_util")
+    def test_choose_names_user_cancel(self, mock_util):
+        items = ["first", "second", "third"]
+        question = "Want to cancel?"
+        mock_util().checklist.return_value = ('cancel', [])
```
bmw: We should use `display_util.CANCEL` here instead of hardcoding this value.

joohoi:
I successfully resisted the urge to do this in such an obscure way that would have accounted for yet another round ;)

PR #5601, opened 2018-02-21
Add finalize_order to shim object, update Certbot to use it

### LINKED ISSUE #5580: Automatically handle ACMEv1 vs. ACMEv2 issuance

ohemorange:
Fixes #5580.

### DIFF
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -8,6 +8,8 @@ import socket
 import sys
 
 import OpenSSL
+import josepy as jose
+
 
 from acme import errors
 
@@ -280,3 +282,23 @@ def gen_ss_cert(key, domains, not_before=None,
     cert.set_pubkey(key)
     cert.sign(key, "sha256")
     return cert
+
+def dump_pyopenssl_chain(chain, filetype=OpenSSL.crypto.FILETYPE_PEM):
+    """Dump certificate chain into a bundle.
+
+    :param list chain: List of `OpenSSL.crypto.X509` (or wrapped in
+        :class:`josepy.util.ComparableX509`).
+
+    """
+    # XXX: returns empty string when no chain is available, which
+    # shuts up RenewableCert, but might not be the best solution...
+
+    def _dump_cert(cert):
+        if isinstance(cert, jose.ComparableX509):
+            # pylint: disable=protected-access
+            cert = cert.wrapped
+        return OpenSSL.crypto.dump_certificate(filetype, cert)
+
+    # assumes that OpenSSL.crypto.dump_certificate includes ending
+    # newline character
+    return b"".join(_dump_cert(cert) for cert in chain)
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -1,4 +1,5 @@
 """Certbot client API."""
+import datetime
 import logging
 import os
 import platform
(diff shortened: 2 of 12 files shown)


ohemorange:
todo

- [ ] remove clientv1 passthrough
- [ ] fullchain_pem in the order is bytes
- [ ] tests for changes to certbot (mostly in client, probably)
- [ ] finish tests for changes to acme.client

acme/acme/crypto_util.py
```diff
@@ -280,3 +282,23 @@ def gen_ss_cert(key, domains, not_before=None,
     cert.set_pubkey(key)
     cert.sign(key, "sha256")
     return cert
+
+def dump_pyopenssl_chain(chain, filetype=OpenSSL.crypto.FILETYPE_PEM):
```
bmw: We may want to consider modifying this function before the release or making it private, but I don't care too much. The reason for this is it forces the user of the module to use `OpenSSL` (which is slowly going away in favor of `cryptography`) and it returns bytes instead of strings.

certbot/client.py
```diff
@@ -298,10 +274,9 @@ def obtain_certificate(self, domains):
 
         :param list domains: domains to get a certificate
 
-        :returns: `.CertificateResource`, certificate chain (as
-            returned by `.fetch_chain`), and newly generated private key
-            (`.util.Key`) and DER-encoded Certificate Signing Request
-            (`.util.CSR`).
+        :returns: :returns: certificate as PEM string, chain as PEM string,
```
bmw: Too many returns.

acme/acme/crypto_util.py
ohemorange: It's like that because `fetch_chain` returns `OpenSSL` objects, which isn't going to change.

bmw: > The reason for this is it forces the user of the module to use OpenSSL (which is slowly going away in favor of cryptography)

While this applied to `chain` as you pointed out, we could remove the need to use it for `filetype`, but I think it is a pretty minor point and making separate functions or types to hide this from the user doesn't buy us much and makes our code more complicated.

> it returns bytes instead of strings.

I'm confident this is correct for a function like this. See https://github.com/certbot/certbot/issues/5606#issuecomment-369768427.

PR #5608, opened 2018-02-22
Apache plugin wildcard support for ACMEv2

joohoi:
Functional example for wildcard certificate support in Apache plugin.

In `deploy_cert()` and `enhance()`, the user will be presented with a dialog to choose from the VirtualHosts that can be covered by the wildcard domain name. The (multiple) selection result will then be handled in a similar way that we previously handled a single VirtualHost that was returned by the `_find_best_vhost()`.

Additionally the selected VirtualHosts are added to a dictionary that maps selections to a wildcard domain to be reused in the later `enhance()` call and not forcing the user to select the same VirtualHosts again.

The code here should be almost completely compatible with Nginx plugin as well, so when we get this merged, we can bring the changes over to Nginx plugin also.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -155,7 +155,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         self._chall_out = set()
         # List of vhosts configured per wildcard domain on this run.
         # used by deploy_cert() and enhance()
-        self.wildcard_vhosts = dict()
+        self._wildcard_vhosts = dict()
         # Maps enhancements to vhosts we've enabled the enhancement for
         self._enhanced_vhosts = defaultdict(set)
 
@@ -266,7 +266,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             self.aug, self.conf("server-root"), self.conf("vhost-root"),
             self.version, configurator=self)
 
-    def wildcard_domain(self, domain):
+    def _wildcard_domain(self, domain):
         """
         Checks if domain is a wildcard domain
 
@@ -299,21 +299,35 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             a lack of directives
 
         """
-        if self.wildcard_domain(domain):
-            # Ask user which VHosts to support.
-            # Returned objects are guaranteed to be ssl vhosts
-            deploy_vhosts = self.choose_vhosts_wildcard(domain, create_ssl=True)
-            for vhost in deploy_vhosts:
-                self._deploy_cert(vhost, cert_path, key_path,
-                                  chain_path, fullchain_path)
-                if domain not in self.wildcard_vhosts.keys():
-                    self.wildcard_vhosts[domain] = list()
-                self.wildcard_vhosts[domain].append(vhost)
-        else:
-            vhost = self.choose_vhost(domain)
+        vhosts = self.choose_vhosts(domain)
+        for vhost in vhosts:
             self._deploy_cert(vhost, cert_path, key_path, chain_path, fullchain_path)
 
-    def vhosts_for_wildcard(self, domain):
+    def choose_vhosts(self, domain, cached=False, create_if_no_ssl=True):
+        """
+        Finds VirtualHosts that can be used with the provided domain
+
+        :param str domain: Domain name to match VirtualHosts to
+        :param bool cached: Return VHosts from a list cached previously on this
+            execution.
+        :param bool create_if_no_ssl: If found VirtualHost doesn't have a HTTPS
+            counterpart, should one get created
+
+        :returns: List of VirtualHosts or None
+        :rtype: `list` of :class:`~certbot_apache.obj.VirtualHost`
+        """
+
+        if self._wildcard_domain(domain):
+            if cached and domain in self._wildcard_vhosts:
+                # Vhosts for a wildcard domain were already selected
+                return self._wildcard_vhosts[domain]
+            # Ask user which VHosts to support.
+            # Returned objects are guaranteed to be ssl vhosts
+            return self._choose_vhosts_wildcard(domain, create_if_no_ssl)
+        else:
+            return [self.choose_vhost(domain)]
+
+    def _vhosts_for_wildcard(self, domain):
         """
         Get VHost objects for every VirtualHost that the user wants to handle
         with the wildcard certificate.
@@ -330,7 +344,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
 
     def _in_wildcard_scope(self, name, domain):
         """
-        Helper method for vhosts_for_wildcard() that makes sure that the domain
+        Helper method for _vhosts_for_wildcard() that makes sure that the domain
         is in the scope of wildcard domain.
 
         eg. in scope: domain = *.wild.card, name = 1.wild.card
@@ -340,11 +354,11 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             return fnmatch.fnmatch(name, domain)
 
 
-    def choose_vhosts_wildcard(self, domain, create_ssl=True):
+    def _choose_vhosts_wildcard(self, domain, create_ssl=True):
         """Prompts user to choose vhosts to install a wildcard certificate for"""
 
         # Get all vhosts that are covered by the wildcard domain
-        vhosts = self.vhosts_for_wildcard(domain)
+        vhosts = self._vhosts_for_wildcard(domain)
 
         # Go through the vhosts, making sure that we cover all the names
         # present, but preferring the SSL vhosts
@@ -354,7 +368,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                 if vhost.ssl:
                     # Always prefer SSL vhosts
                     filtered_vhosts[name] = vhost
-                elif name not in filtered_vhosts.keys():
+                elif name not in filtered_vhosts:
                     # Add if not in list previously
                     filtered_vhosts[name] = vhost
 
@@ -373,6 +387,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
             else:
                 return_vhosts.append(vhost)
 
+        self._wildcard_vhosts[domain] = return_vhosts
         return return_vhosts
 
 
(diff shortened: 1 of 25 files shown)


certbot-apache/certbot_apache/configurator.py
```diff
@@ -1377,7 +1478,18 @@ def enhance(self, domain, enhancement, options=None):
             raise errors.PluginError(
                 "Unsupported enhancement: {0}".format(enhancement))
         try:
-            func(self.choose_vhost(domain), options)
+            if self.wildcard_domain(domain):
```
ohemorange: can we refactor this so there's less inside the try block?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,91 @@ def deploy_cert(self, domain, cert_path, key_path,
...
-        # vhost choice namespace similar with the pre-validation one.
-        vhost = self.choose_vhost(domain)
+        if self.wildcard_domain(domain):
+            # Ask user which VHosts to support.
+            # Returned objects are guaranteed to be ssl vhosts
+            deploy_vhosts = self.choose_vhosts_wildcard(domain, create_ssl=True)
+            for vhost in deploy_vhosts:
+                self._deploy_cert(vhost, cert_path, key_path,
+                                  chain_path, fullchain_path)
+                if domain not in self.wildcard_vhosts.keys():
```
ohemorange: is this for if we ended up creating one?

ohemorange: oh nvm because we might not match

ohemorange: don't think you need `.keys()` here

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,91 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+        is in the scope of wildcard domain.
+
+        eg. in scope: domain = *.wild.card, name = 1.wild.card
+        not in scope: domain = *.wild.card, name = 1.2.wild.card
+        """
+        if len(name.split(".")) == len(domain.split(".")):
+            return fnmatch.fnmatch(name, domain)
+
+
+    def choose_vhosts_wildcard(self, domain, create_ssl=True):
```
ohemorange: this can probably be internal

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,91 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+        # present, but preferring the SSL vhosts
+        filtered_vhosts = dict()
+        for vhost in vhosts:
+            for name in vhost.get_names():
+                if vhost.ssl:
+                    # Always prefer SSL vhosts
+                    filtered_vhosts[name] = vhost
+                elif name not in filtered_vhosts.keys():
+                    # Add if not in list previously
+                    filtered_vhosts[name] = vhost
```
ohemorange: what happens if there aren't any? do we want to create a new one?

certbot-apache/certbot_apache/configurator.py
ohemorange: also, why not do this inside `choose_vhosts_wildcard`?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,91 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+                self._deploy_cert(vhost, cert_path, key_path,
+                                  chain_path, fullchain_path)
+                if domain not in self.wildcard_vhosts.keys():
+                    self.wildcard_vhosts[domain] = list()
+                self.wildcard_vhosts[domain].append(vhost)
+        else:
+            vhost = self.choose_vhost(domain)
+            self._deploy_cert(vhost, cert_path, key_path, chain_path, fullchain_path)
+
+    def vhosts_for_wildcard(self, domain):
```
ohemorange: also can be internal?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -1377,7 +1478,18 @@ def enhance(self, domain, enhancement, options=None):
             raise errors.PluginError(
                 "Unsupported enhancement: {0}".format(enhancement))
         try:
-            func(self.choose_vhost(domain), options)
+            if self.wildcard_domain(domain):
+                # Handle virtualhosts configured for a wildcard domain
+                if domain in self.wildcard_vhosts.keys():
```
ohemorange: can this if (`if domain in self.wildcard_vhosts:`) go inside `choose_vhosts_wildcard`?

ohemorange: oh yeah, if both this and the other thing go into the function, that cleans everything else up a bunch!


these are my thoughts as I do the nginx one, it may be clearer to look at the PR I put up for that

ohemorange:
What happens if apache is authenticator and it's given a wildcard domain?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -152,6 +153,9 @@ def __init__(self, *args, **kwargs):
         self.assoc = dict()
         # Outstanding challenges
         self._chall_out = set()
+        # List of vhosts configured per wildcard domain on this run.
+        # used by deploy_cert() and enhance()
+        self.wildcard_vhosts = dict()
```
ohemorange: underscores everywhere

certbot-apache/certbot_apache/configurator.py
joohoi: Yeah, we do create one when appending the vhosts to the final list in the method.

joohoi:
> What happens if apache is authenticator and it's given a wildcard domain?

We had a short discussion about this yesterday, but Certbot should raise an exception when Apache plugin is connecting to ACME server that doesn't allow wildcard challenges over HTTP or TLS-SNI challenges.

certbot-apache/certbot_apache/obj.py
```diff
@@ -167,6 +167,19 @@ def __str__(self):
                 active="Yes" if self.enabled else "No",
                 modmacro="Yes" if self.modmacro else "No"))
 
+    def display_repr(self):
+        """Return a representation of VHost to be used in dialog"""
+        return (
+            "File: {filename}\n"
+            "Addresses: {addrs}\n"
+            "Names: {names}\n"
+            "HTTPS: {https}\n\n".format(
```
bmw: nit: The trailing empty line on the last entry looks a little awkward to me:
```
1: File: /etc/apache2/sites-enabled/bmatch.conf
Addresses: *:80
Names: geqgeqoc.example.org
HTTPS: No

2: File: /etc/apache2/sites-enabled/amatch.conf
Addresses: *:80
Names: a.example.org
HTTPS: No

3: File: /etc/apache2/sites-enabled/wildcard.conf
Addresses: *:80
Names: abc.def.example.org, *.example.org
HTTPS: No

```
Unless you prefer it that way, can we remove it?

certbot/tests/display/ops_test.py
```diff
@@ -300,7 +300,7 @@ def test_get_valid_domains(self):
         from certbot.display.ops import get_valid_domains
         all_valid = ["example.com", "second.example.com",
                      "also.example.com", "under_score.example.com",
-                     "justtld", "*.wildcard.com"]
+                     "justtld"]
```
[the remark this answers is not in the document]
joohoi: You are right. This was removed at the early phase of the development, and is now brought back.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,106 @@ def deploy_cert(self, domain, cert_path, key_path,
...
 
         """
-        # Choose vhost before (possible) enabling of mod_ssl, to keep the
-        # vhost choice namespace similar with the pre-validation one.
-        vhost = self.choose_vhost(domain)
+        vhosts = self.choose_vhosts(domain)
+        for vhost in vhosts:
+            self._deploy_cert(vhost, cert_path, key_path, chain_path, fullchain_path)
+
+    def choose_vhosts(self, domain, cached=False, create_if_no_ssl=True):
```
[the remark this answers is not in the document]
joohoi: This is correct, we shouldn't hit such scenario. This parameter is now removed.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,106 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+    def _in_wildcard_scope(self, name, domain):
+        """
+        Helper method for _vhosts_for_wildcard() that makes sure that the domain
+        is in the scope of wildcard domain.
+
+        eg. in scope: domain = *.wild.card, name = 1.wild.card
+        not in scope: domain = *.wild.card, name = 1.2.wild.card
+        """
+        if len(name.split(".")) == len(domain.split(".")):
+            return fnmatch.fnmatch(name, domain)
```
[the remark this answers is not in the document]
joohoi: Yeah, `vhost.get_names()` returns all the names associated to the `VirtualHost`, both `ServerNames` and `ServerAlias`es. Furthermore, I don't think wildcard names are allowed in the `ServerName` in the first place.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -280,9 +299,106 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+        dialog_output = display_ops.select_vhost_multiple(list(dialog_input))
+
+        # Make sure we create SSL vhosts for the ones that are HTTP only
+        # if requested.
+        return_vhosts = list()
+        for vhost in dialog_output:
+            if not vhost.ssl and create_ssl:
+                return_vhosts.append(self.make_vhost_ssl(vhost))
+            else:
+                return_vhosts.append(vhost)
```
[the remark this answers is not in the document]
joohoi: This is a more sophisticated way to handle it, I changed the behavior accordingly. This also helps to reduce the static from the dialog in the first place.

certbot-apache/certbot_apache/display_ops.py
```diff
@@ -13,10 +13,42 @@
...
+    :param vhosts: Available Apache VirtualHosts
+    :type vhosts: :class:`list` of type `~obj.Vhost`
+
+    :returns: List of VirtualHosts
+    :rtype: :class:`list`of type `~obj.Vhost`
+    """
+    if not vhosts:
+        return list()
+    tags_list = [vhost.display_repr() for vhost in vhosts]
+    while True:
```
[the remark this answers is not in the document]
joohoi: I presumed there being an undefined reason behind this, as `select_vhost()` function has the same behavior. It's been around since original implementation by James, introduced in 99fcb4f23 and now looking at the original commit, it looks like it was in place for a help selection that has since been removed. I cleaned this up from the `select_vhost()` too.

certbot-apache/certbot_apache/obj.py
joohoi: It looks a lot cleaner without the extra newline, I agree. Even though handling that requires a few more lines in the code. This is now implemented.

bmw:
Looks good except I don't think my comment at https://github.com/certbot/certbot/pull/5608#pullrequestreview-99937605 got addressed.

To expand a little more on why I think it's important, Certbot will currently exit with a zero status and no warnings if there's no matching vhosts so it's not clear it didn't do what the user asked.

While we may want to change the behavior around deploy/enhance if we fail for a domain in the future (what's being proposed is we continue execution so we work for subsequent domains), I don't think we should make any of those changes here as it requires some UI changes to make sure the domains we failed on is 100% clear and not hidden in the scrollback in the user's terminal. Currently if we fail on one, we exit with an error and rollback changes to all which is while arguably less useful, I think make things pretty clear to the user: your request to deploy/enhance the cert failed.

PR #5612, opened 2018-02-22
Fix ACMEv2 issues

bmw:
I think most of the changes in this PR are pretty self explanatory with the help of comments and commit messages, however, I should probably expand on the thought process for handling the lack of combinations in ACMEv2 authorizations.

I talked about this a bit with Erica and I think our only options are either to not try and handle this problem in our ACME shim (the approach I took) or complicate and/or muddle the shim's otherwise clean interface. If we wanted to try and handle this difference, we'd need to either do something like always add combinations to the authz (which deviates from the idea of it being a backwards compatible V2 client) or push the logic of choosing challenges based on plugins and user preferences into ACME (which significantly complicates the interface and makes what users of the class can do less flexible).

We're only talking about ACME servers based on the protocol from over a year and a half ago and we're not even sure any of these exist. Rather than complicating the interface that we have to live with going forward, let's keep this clean and deal with this in Certbot for the sake of preserving our current behavior. If anyone else really needs or wants to support this, I tried to make this limitation clear and they should be able to do so with minimal changes like those I made in Certbot's `auth_handler.py`.

### DIFF
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -24,7 +24,7 @@ class AuthHandler(object):
         :class:`~acme.challenges.Challenge` types
     :type auth: :class:`certbot.interfaces.IAuthenticator`
 
-    :ivar acme.client.Client acme: ACME client API.
+    :ivar acme.client.BackwardsCompatibleClientV2 acme: ACME client API.
 
     :ivar account: Client's Account
     :type account: :class:`certbot.account.Account`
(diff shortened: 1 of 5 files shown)


certbot/auth_handler.py
```diff
@@ -100,10 +100,16 @@ def _choose_challenges(self, domains):
         """Retrieve necessary challenges to satisfy server."""
         logger.info("Performing the following challenges:")
         for dom in domains:
+            dom_challenges = self.authzr[dom].body.challenges
+            if self.acme.acme_version == 1:
+                combinations = self.authzr[dom].body.combinations
+            else:
+                combinations = tuple((i,) for i in range(len(dom_challenges)))
```
ohemorange: the documentation below says combinations is a tuple, but for some reason I thought it was a list. obviously it should work either way, but were you basing this off the docs or manually checking?

bmw: Manually checking against the V1 endpoint. It was a tuple of tuples.

Issue #5367, opened 2018-01-04
Get ACMEv2 support in Certbot


ohemorange:
Step 2 of #5365 

#### Certbot core

- [x] Delete lines from `certbot.util.enforce_domain_sanity` that error on wildcards. We should still have code somewhere that errors if a wildcard is requested from a v1 endpoint because according to the spec "Wildcard domain names (with "*" as the first label) MUST NOT be included in authorization requests" and we should provide a clean error message to the user.
- [x] Improve error message if a wildcard is requested with an authenticator that can't do DNS challenges.
- [x] Don't make cert names with wildcards by default.
- [x] Fix `--allow-subset-of-names` for ACMEv2. This could be done in `certbot.client.obtain_certificate`, by having it repeatedly create a CSR, request a `new_order`, and perform authorizations until all authorizations pass. If any fail, we should create a new CSR with only the domains that succeeded. The order/authorizations would then be passed to `obtain_certificate_from_csr`.
- [x] Need to show link to TOS before calling `register`/`new_account` on the client from the acme module when ACMEv2 is used. Doing this requires being able to get TOS from ACMEv2 directory, ideally without making the user of the ACME library reach into the directory object themselves. We also need to skip this code for ACMEv1 and not try and show TOS again after registering for ACMEv2. **(Suggests acme change)**
- [x] Call `new_account`/`register` based on ACME version **(Could be mitigated by #5491 or changes to `acme.client.ClientV2`)**
- [x] If current design of #5374 stays, we have to make CSR before performing challenges in `certbot.client.obtain_certificate` because the CSR is needed earlier by the client in ACME. **(Could be mitigated by #5491 or changes to #5374)**
- [x] `certbot.auth_handler.get_authorizations` (or closely related function) should take a dict/list of authorization objects instead of domains. We can have another function on the `auth_handler` take domains and create authzs if we wanted for convenience of supporting ACMEv1 **(Could be mitigated by #5491 by keeping order object internal to acme client)**
- [x] We now get back the fullchain from the ACME server rather than separate leaf and chain certs. These need to be split (probably by just looking for the first line containing "END CERTIFICATE"). **(Could be mitigated by #5491)**

#### Apache and Nginx

- [x] Open question: Should Apache/Nginx suggest wildcards found in the config with `get_all_names`?
[...]

Issue #5550, opened 2018-02-07
Error: duplicate listen options for [::]:443


hackuun:
## My operating system is (include version):
Ubuntu 16.04

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
`sudo apt-get install python-certbot-nginx`

## I ran this command and it produced this output:
`nginx: [emerg] duplicate listen options for [::]:443 in /etc/nginx/sites-enabled/example.online:29
`

## Certbot's behavior differed from what I expected because:
It should be no errors

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:

```
server {
  listen 80;
  listen [::]:80;

  server_name example.online;

  root /home/example/deploy;
  index index.html;

  location / {
    try_files $uri $uri/ =404;
  }
}

server {
  listen 80;
  listen [::]:80;
  server_name www.example.online;
  return 301 $scheme://example.online$request_uri;
}

```

neoadventist:
@iamdubx did you figure this out? I'm having the same issue.

quen2404:
Same issue... It works for my default site, but not for custom sub domain

varunyellina:
I'm having the same issue.

I do have a config to catch multiple domains though.

```
server {
  listen 80;
  listen [::]:80;

  root /home/primarydomain/public;
  index index.html index.htm;

  server_name domain1.com *.domain1.com domain2.com *.domain2.com domain3.com *.domain3.com domain4.com *.domain4.com;

  return 302 $scheme://primarydomain.com$request_uri;

  access_log /var/log/nginx/others.access.log;
  error_log /var/log/nginx/others.error.log;

  location / {
    try_files $uri $uri/ /index.html =404;
  }
}
```

I get `nginx: [emerg] duplicate listen options for [::]:443 in /etc/nginx/sites-enabled/others:19` for this config.

Any help?

amattn:
Same problem.  

I ran the command: `certbot --redirect --nginx -d readacted.com -d www.redacted.com`

my original conf file looks like: 

```
    server {
        server_name redacted.com;
        location / {
        root   /home/redacted/www;
        index  index.html;
      }
    }

    server {
        listen 80;
        listen [::]:80;
        server_name www.redacted.com;
        return 301 $scheme://redacted.com$request_uri;
    }
```

according to /var/log/letsencrypt/letsencrypt.log I see certbot is trying to do this:

```
    server {
        server_name redacted.com;
        location / {
[...]

Issue #5419, opened 2018-01-12
Nginx fails to deploy cert


bmw:
## My operating system is (include version):

Ubuntu 16.04

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

Dev instructions (I was on commit 5d58a3d847f5a93b8eaaaea482063fae69d2b2d7 so this is unrelated to HTTP01 support in Nginx)

## I ran this command and it produced this output:

Starting with the default Nginx config on this system I ran:

```
sudo certbot -a webroot -w /var/www/html/ -d example.org -i nginx --no-redirect
sudo certbot -a webroot -w /var/www/html/ -d abc.example.org  -d def.example.org -i nginx --no-redirect
```
which gave me:
```
nginx: [emerg] duplicate listen options for [::]:443 in /etc/nginx/sites-enabled/default:200
Encountered exception:
Traceback (most recent call last):
  File "/home/bmw/src/certbot/certbot/client.py", line 462, in deploy_certificate
    self.installer.restart()
  File "/home/bmw/src/certbot/certbot-nginx/certbot_nginx/configurator.py", line 687, in restart
    nginx_restart(self.conf('ctl'), self.nginx_conf)
  File "/home/bmw/src/certbot/certbot-nginx/certbot_nginx/configurator.py", line 919, in nginx_restart
    "nginx restart failed:\n%s\n%s" % (out.read(), err.read()))
MisconfigurationError: nginx restart failed:



Calling registered functions
Rolling back to previous server configuration...
Reporting to user: We were unable to install your certificate, however, we successfully restored your server to its prior configuration.
```

elkhan:
What helped me was deleting `ipv6only=on` from `listen [::]:443 ssl` - `nginx/sites-available/whatever.com`

owalerys:
Also having this issue, certs issued with v0.19.0 Ubuntu client are fine because they are missing the listen [::]:443 line, but certs issued with v0.21.0 certbot-auto client have problems due to the ipv6only=on line.

owalerys:
It seems like `ipv6only=on` is the default configuration for this setting now for nginx in its current version (even as of 2 years ago), so omitting the line from the certbot nginx plugin would solve this issue.

ohemorange:
This looks like a duplicate of https://github.com/certbot/certbot/issues/5550, reopen if it's not.

PR #5619, opened 2018-02-24
Nginx plugin wildcard support for ACMEv2

ohemorange:
Part of #5367. Based largely on #5608.

If the user wants a redirect, they'll have to reselect server blocks.

Todo:
- [x] Add tests for deploying and enhancing wildcard domains
- [x] What happens when Nginx is authenticator and it's given a wildcard domain? 
- [x] Manually test staple-ocsp


bmw:
One piece of this (or another Nginx PR we need before we release ACMEv2) that we didn't need for Apache is fixing redirects for wildcards.
```
if ($host = *.example.org)
```
is never true (unless Nginx has an usual interpretation of `=` that I'm not aware of).

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -377,15 +502,19 @@ def choose_redirect_vhost(self, target_name, port, create_if_no_match=False):
...
 
         """
-        matches = self._get_redirect_ranked_matches(target_name, port)
-        vhost = self._select_best_name_match(matches)
-        if not vhost and create_if_no_match:
-            vhost = self._vhost_from_duplicated_default(target_name, port=port)
-        return vhost
+        if _wildcard_domain(target_name):
+            # Ask user which VHosts to enhance.
+            vhosts = self._choose_vhosts_wildcard(target_name, prefer_ssl=False)
```
bmw: Two things:

1. This method ignores the port number given to this function. Are we OK with this?
2. With `prefer_ssl`, we'll still return SSL vhosts if no plaintext vhosts exist. Is this correct?

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -166,25 +173,139 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+    def _vhosts_for_wildcard(self, domain):
+        """
+        Get VHost objects for every VirtualHost that the user wants to handle
+        with the wildcard certificate.
+        """
+
+        # Collect all vhosts that match the name
+        matched = set()
+        for vhost in self.vhosts:
+            for name in vhost.get_names():
```
bmw: vhost objects don't have `get_names()` in Nginx.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -889,14 +1043,39 @@ def _test_block_from_block(block):
...
+        return domain.startswith(wildcard_marker)
+
+
 def _redirect_block_for_domain(domain):
+    updated_domain = domain
+    match_symbol = '='
+    if _wildcard_domain(domain):
+        match_symbol = '~'
+        updated_domain = updated_domain.replace('.', '\.')
+        updated_domain = updated_domain.replace('*', '.+')
```
bmw: This needs to be restricted a bit more. For example, if `domain` is `*.example.org`, this regex will match `a.b.example.org` which isn't covered by a cert for the given domain. I think using `[^.]+` instead of `.+` here should work.

certbot-nginx/certbot_nginx/display_ops.py
```diff
@@ -0,0 +1,44 @@
...
+    """Select multiple Vhosts to install the certificate for
+    :param vhosts: Available Nginx VirtualHosts
+    :type vhosts: :class:`list` of type `~obj.Vhost`
+    :returns: List of VirtualHosts
+    :rtype: :class:`list`of type `~obj.Vhost`
+    """
+    if not vhosts:
+        return list()
+    tags_list = [vhost.display_repr() for vhost in vhosts]
+    while True:
```
bmw: https://github.com/certbot/certbot/pull/5608#discussion_r171141004

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -216,3 +216,15 @@ def ipv4_enabled(self):
...
             if not a.ipv6:
                 return True
+
+    def display_repr(self):
+        """Return a representation of VHost to be used in dialog"""
+        return (
+            "File: {filename}\n"
+            "Addresses: {addrs}\n"
+            "Names: {names}\n"
+            "HTTPS: {https}\n\n".format(
```
bmw: https://github.com/certbot/certbot/pull/5608#discussion_r171144239

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -889,14 +1043,39 @@ def _test_block_from_block(block):
     parser.comment_directive(test_block, 0)
     return test_block[:-1]
 
+
+def _wildcard_domain(domain):
```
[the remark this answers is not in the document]
ohemorange: done

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -166,25 +173,139 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+        return_vhosts = display_ops.select_vhost_multiple(list(dialog_input))
+
+        for vhost in return_vhosts:
+            if target_name not in vhosts_cache:
+                vhosts_cache[target_name] = []
+            vhosts_cache[target_name].append(vhost)
+
+        return return_vhosts
+
+    def _choose_redirect_vhosts_wildcard(self, domain):
```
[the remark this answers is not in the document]
ohemorange: done

certbot-nginx/certbot_nginx/configurator.py
ohemorange: done

certbot-nginx/certbot_nginx/display_ops.py
ohemorange: fun fact these links never go to the right spot

ohemorange: but I can manually check, so, got it.

certbot-nginx/certbot_nginx/obj.py
ohemorange: done

certbot-nginx/certbot_nginx/display_ops.py
ohemorange: done

certbot-nginx/certbot_nginx/configurator.py
ohemorange: 1. Not ideal because they could theoretically pick a bad one. I'll fix it.
2. Yes -- the test is "is it listening on 80", not "is it not listening on 443 ssl". And it's really just a filter to cut down the number of vhosts the poor users have to sift through; too many is fine.

and, done.

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -166,25 +173,139 @@ def deploy_cert(self, domain, cert_path, key_path,
...
+        # Collect all vhosts that match the name
+        matched = set()
+        for vhost in self.vhosts:
+            for name in vhost.get_names():
+                if self._in_wildcard_scope(name, domain):
+                    matched.add(vhost)
+
+        return list(matched)
+
+    def _in_wildcard_scope(self, name, domain):
```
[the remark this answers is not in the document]
ohemorange: Looks like it is [literally possible](https://stackoverflow.com/questions/1849447/how-can-you-detect-if-two-regular-expressions-overlap-in-the-strings-they-can-ma), but I'm going to go ahead and get rid of this.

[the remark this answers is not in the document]
ohemorange: done

certbot-nginx/certbot_nginx/configurator.py
ohemorange: I agree, done.

certbot-nginx/certbot_nginx/obj.py
```diff
@@ -193,6 +193,11 @@ def __eq__(self, other):
 
         return False
 
+    def __hash__(self):
```
bmw: I'm not that worried about this, we should almost certainly leave this for now, and I know we do the same thing in Apache to work around the same problems, BUT from the documentation for [\_\_hash\_\_](https://docs.python.org/3/reference/datamodel.html#object.__hash__):

> If a class defines mutable objects and implements an \_\_eq\_\_() method, it should not implement \_\_hash\_\_(), since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

We don't change vhosts after they're created often, but we do in a few places and this could lead to bugs like the ones Python is warning us about. Might be worth making these objects immutable or writing our own code to eliminate duplicates from a collection of vhosts at some point.

certbot-nginx/certbot_nginx/configurator.py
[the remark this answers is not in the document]
bmw: Oh wow. Thanks for the link. Yeah, let's not do this now (or ever).

ohemorange:
Sounds like you hit https://github.com/certbot/certbot/issues/5550, which you previously reported at https://github.com/certbot/certbot/issues/5419. Let me know if that's not the case.

certbot-nginx/certbot_nginx/obj.py
ohemorange: https://github.com/certbot/certbot/issues/5647

PR #5620, opened 2018-02-24
Fix wildcard issuance

### LINKED ISSUE #5613: Authz for wildcard vs base domain
In a number of places, we store authorizations as a mapping from domains to its authorization. This is a problem because when requesting a single certificate for *.example.org and example.org, boulder returns an order with two authorizations for example.org. The authorization for the wildcard includes `wildcard: true` in the authz, but this doesn't seem to be in the ACME spec.

I'm sure we can find a way around it, but @jsha, is there anything I'm missing in the spec that we can rely on to map authzs back to their domains for an arbitrary ACMEv2-style server?

bmw:
Fixes #5613.

By far the biggest changes here are in auth handler. Previously it was assumed that the domain was unique in each authorization and this was used to associate challenges back to their authorizations. Since this is no longer true, I'm now storing authorizations and the challenges we're performing for the authorization together and using their index in a list to refer back to them.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -298,7 +298,7 @@ class Client(object):
         auth_domains = set(a.body.identifier.value for a in authzr)
         successful_domains = [d for d in domains if d in auth_domains]
 
-        if successful_domains != domains:
+        if self.config.allow_subset_of_names and successful_domains != domains:
             if not self.config.dry_run:
                 os.remove(key.file)
                 os.remove(csr.file)
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -298,6 +298,11 @@ class Client(object):
         auth_domains = set(a.body.identifier.value for a in authzr)
         successful_domains = [d for d in domains if d in auth_domains]
 
+        # allow_subset_of_names is currently disabled for wildcard
+        # certificates. The reason for this and checking allow_subset_of_names
+        # below is because successful_domains == domains is never true if
+        # domains contains a wildcard because the ACME spec forbids identifiers
+        # in authzs from containing a wildcard character.
         if self.config.allow_subset_of_names and successful_domains != domains:
             if not self.config.dry_run:
                 os.remove(key.file)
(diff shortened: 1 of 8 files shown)


certbot/client.py
```diff
@@ -298,7 +298,7 @@ def obtain_certificate(self, domains):
         auth_domains = set(a.body.identifier.value for a in authzr)
         successful_domains = [d for d in domains if d in auth_domains]
 
-        if successful_domains != domains:
+        if self.config.allow_subset_of_names and successful_domains != domains:
```
ohemorange: `# This seems redundant, but is necessary because a.body.identifier.value will not be the domain for wildcard certificates` and if that's not true, then a comment explaining what is the case.

certbot/client.py
```diff
@@ -298,6 +298,11 @@ def obtain_certificate(self, domains):
         auth_domains = set(a.body.identifier.value for a in authzr)
         successful_domains = [d for d in domains if d in auth_domains]
 
+        # allow_subset_of_names is currently disabled for wildcard
+        # certificates. The reason for this and checking allow_subset_of_names
+        # below is because successful_domains == domains is never true if
+        # domains contains a wildcard because the ACME spec forbids identifiers
+        # in authzs from containing a wildcard character.
```
ohemorange: todo(ebp): read this PR over again with this new knowledge

bmw:
> The failure error message will be inaccurate for wildcards. I'm fine with that for now.

Yeah ☹️ 

I brought this up with Jacob in the issue this PR fixes linked above. Initially I was pretty disappointed about this, but it isn't so bad. If the user wants to go the route of fix things so the challenge passes, the actual domain attempting to be validated is useful to have and that's what's contained in the authz/achall.

If they want to remove the domain causing the problem, the domain in the authz is less useful though. Ideally we could provide both, but unfortunately, I don't see a way for us to do this in ACMEv2 (without relying on things particular to Let's Encrypt).

PR #5622, opened 2018-02-25
make pip_install.sh compatible with POSIX sh(1) again

delan:
Version 0.21.0 introduced an implicit dependency on bash(1), which is often not included by default outside of Linux/macOS/Cygwin.


tools/pip_install.sh
```diff
@@ -23,4 +23,6 @@ fi
 set -x
 
 # install the requested packages using the pinned requirements as constraints
-pip install -q --constraint <("$merge_reqs" "$dev_constraints" "$test_constraints") "$@"
+constraint_path="$(mktemp)"
+"$merge_reqs" "$dev_constraints" "$test_constraints" > "$constraint_path"
+pip install -q --constraint "$constraint_path" "$@"
```
schoen: I guess that the tempfile should be deleted afterwards (like `[ -f "$constraint_path" ] && rm -f "$constraint_path"` or similar).

schoen:
@bmw says that we can't promise to make this kind of tool never depend on bash in the future, but we're happy to take this PR to remove the unnecessary dependency. Do you think you could update the PR to remove the temporary file afterward?

delan:
No worries — avoiding the dependency is nice to have but not essential — and done.

delan:
This pull request is now just a shebang change. While `/bin/sh` isn’t perfect — POSIX [doesn’t have an opinion](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#tag_20_117_16) on where a compliant sh(1) ought to be installed, and platforms such as [Tru64](http://openshell.org/service/man/?qf=sh&af=0&sf=0&of=Tru64&tf=0) and [Solaris < 10](https://docs.oracle.com/cd/E19253-01/816-5165/6mbb0m9rm/index.html) don’t use `/bin/sh` — it should be compatible with most Linux, macOS, BSD, and Cygwin systems.

delan:
No worries @bmw — happy to help!

PR #5605, opened 2018-02-22
Drop Python 2.6 and 3.3 support

bmw:
Built on #5508. Fixes #5489 and fixes #5312. This PR should be merged rather than squashed to preserve authorship.

@erikrose, do you have time to review this? Despite the size, it's fairly straight forward and you'll have the most context about setup.py/pipstrap changes.

PR #5625, opened 2018-02-26
Drop Python 2.6 and 3.3 support in test-everything

bmw:
This PR merges in `master` to `test-everything` to resolve merge conflicts due to `.travis.yml`  changes from #5605. This PR should be merged and not squashed.


bmw:
Test failures will be fixed by #5627. Once that lands, Travis can be kicked and the tests should pass.

bmw:
I was wrong about this branch always being automatically updated unless there were `.travis.yml` conflicts.

We do this merging during the [before_install step](https://github.com/certbot/certbot/blob/349643c9b8dfb0cb97154ca5310617c7ade4ae21/.travis.yml#L7) of the [build process](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle), however, it looks like some of our environments removed the `before_install` step because it's [not needed for the builds in master](https://github.com/certbot/certbot/commit/efd4f357829e538dc0928ba8efe3084ae4d859dd). Latest commit adds this back so we always test with the latest files for each Travis job.

Doing this installs some packages we don't need which might slow down the tests slightly, but because they're run nightly, it doesn't matter much.

bmw:
Ugh. Looks like `before_install` code doesn't work in this environment. I'll fix this soon, but let's just get test-everything working again.

Once again, this PR just removes Python 2.6 and 3.3 tests from `.travis.yml`. It's easiest to compare most changes against `master` (they're all in `.travis.yml`) and the `.travis.yml` changes against the `test-everything` branch.

bmw:
Closing in favor of #5645 which has cleaner commits, handles even more merge conflicts, and has a short essay for people not familiar with what I'm trying to do here.

PR #5626, opened 2018-02-26
Fix revocation in ACMEv2

### LINKED ISSUE #5366: Get ACMEv2 support in acme module
Step 1 of #5365

### LINKED ISSUE #5505: Properly handle modified directory URLs in ACME
Earlier it was suggested just making these URLs optional parameters to the affected methods in `ClientBase` and having the derived classes provide them to the base class methods. Another possible approach is to have the `Directory` class try either naming convention (ACMEv1 version or camel case) and only fail if neither directory field exists. If both exist, we can either favor ACMEv1 or ACMEv2 or raise an error.

bmw:
Fixes #5366 as this is the last major piece of ACMEv2 support in the ACME module I'm aware we're missing. Fixes #5505.

I went with this approach because it fixes the problem without making any API changes (so we can change it later) and still allows `revoke` to work on `BackwardsCompatibleClientV2` because `revoke` is still defined on `ClientBase`.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -222,11 +222,26 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         :raises .ClientError: If revocation is unsuccessful.
 
         """
-        response = self._post(self.directory[messages.Revocation],
-                                 messages.Revocation(
-                                     certificate=cert,
-                                     reason=rsn),
-                                 content_type=None)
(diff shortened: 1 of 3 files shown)


bmw:
Now this should be ready for review. Account based revocation in ACMEv2 works a little differently so I made a 2 line change in Certbot to make that work.

PR #5627, opened 2018-02-27
Don't run tests with Python 2.6

bmw:
Fixes the failures in #5625. These tests were previously being run with Python 2.6 but now the tests don't work on Python 2.6. I tried switching to running the tests in Python 3, but encountered new failures. Let's run the tests in Python 2.7 where we know they work for now.

### DIFF
--- a/letsencrypt-auto-source/Dockerfile.centos6
+++ b/letsencrypt-auto-source/Dockerfile.centos6
@@ -5,13 +5,7 @@ FROM centos:6
 
 RUN yum install -y epel-release
 
-# Install pip and sudo:
-RUN yum install -y python-pip sudo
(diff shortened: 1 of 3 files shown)


ohemorange:
Can you say more about what's going on here? It looks like #5625 removes two tests that we run, why would no longer running tests cause other tests to fail? And weren't we previously passing tests run with python3? Why is that no longer the case?

bmw:
While this makes me sad, getting the test script to work on Python 3 doesn't seem trivial and after fixing a couple imports, I started getting errors from the various tools around `certbot-auto` like `build.py` and `version.py`.

Let's just revert these changes for now and fix it when we have more free time.

bmw:
I created #5632 for tracking this.

Issue #5490, opened 2018-01-26
Break lockstep between our packages


bmw:
As part of getting distros to package more of our plugins, we've agreed to stop releasing all of our packages in lockstep. This means that instead of requiring that all of our packages have the same version number, people are able to mix and match versions. The benefit of this is distros (and us) don't have to repackage everything anytime we change a single package.

To do this, we need to:

1. Modify each `setup.py` to depend on the oldest version of its Certbot dependencies that is needed to work. Some digging needs to be done to figure out what this version is.
2. Set up tests in Travis to ensure the minimum versions required stays accurate. This can be a mix and match between the test-everything branch depending on what makes the most sense.

ohemorange:
My current thought is that rather than digging, we should set everything to the current version. It's faster and I don't currently see any drawbacks.

erikrose:
I agree with @ohemorange. Add to it the complication that pip doesn't yet have a proper dependency resolver, so, if we required >=1.1 in one of our packages and >=1.2 in another, it's a toss-up which one gets obeyed.

erikrose:
Just sat down and sketched out an approach with Erica. Basically amounts to…

* a tox config with pinned shallow versions of the minimum allowed dep versions
* installing the certbot packages with `--no-deps`
* declaring deps like `>=1.1.4,<=2.0` in the setup.py files

Then we'd have a second tox env that tests the latest versions of everything together, because that's a common real-world setup. We should probably just let pip do its thing for that rather than trying to keep up with a pinning; we will *want* to know when somebody does a release of a dep that makes things go bad.

bmw:
Did you guys think about how to handle cases where different Certbot plugins require different minimum versions? We don't have that problem yet, but if the Apache plugin requires Certbot 1.0 and the Nginx plugin requires Certbot 1.1, do we just keep reordering our tests as this changes to test the plugins with the minimum required version first?

erikrose:
No, but, on a brief minute of thought, it seems like that should be handled in the test suites of the individual plugins: Apache should test with certbot 1.0 and nginx with 1.1. Or were you thinking of still coupling the packages together under the auspices of one gigantic test suite? I'd suggest making a list of pros and cons, if so; there is probably value in decoupling "all the way".

bmw:
The way that the unit tests currently work is for each Python version we have a tox environment that iterates over our packages installing and testing them. This looks something like:

1. Install acme.
2. Test acme.
3. Install Certbot.
[...]

Issue #5328, opened 2017-12-14
Make docker images for each DNS plugin


bmw:
This can be another way for people to get DNS plugins once people want them for wildcard certs.

- [x] Write Dockerfiles.
- [ ] Set up Docker Hub repos.

bmw:
Proposal:

Put a Dockerfile under the top level directory for each DNS plugin that looks as simple as something like:
```
FROM certbot/certbot

COPY . src/certbot-dns-cloudflare

RUN pip install --no-cache-dir --editable /opt/certbot/src/certbot-dns-cloudflare
```
Set up Docker Hub repos for each plugin at either certbot/dns-cloudflare or certbot/certbot-dns-cloudflare. I personally prefer the former, but I'd appreciate feedback here.

These images will build for release tags and I'll configure Docker Hub to update the build for each plugin if the Certbot image it's based on changes.

We could offer a rolling tag for each plugin if we want, but the problem is there's no good way that I've found to automate a build for the rolling image so that it depends on the rolling Certbot image. This is desirable because if the most recent version of the DNS plugin relies on unreleased changes in Certbot, the build will fail (as long as we remember to update the plugin's setup.py so it depends on the right version). This might not be the end of the world though.

Changing the base of the Dockerfile either requires having a 2nd Dockerfile (which causes strange behavior with the Dockerfile listed on Docker Hub) or trying to use build vars which isn't officially supported on Docker Hub (see https://github.com/docker/hub-feedback/issues/508). I don't see a clear path forward here or a strong need for an automated rolling image like this, so I'm not sure it's worth the hassle.

Can I get your thoughts on this plan @joohoi?

joohoi:
This all sounds reasonable. As we discussed on a call, it could be worth it to move the plugins to their own repositories at some point of time. 

As the Dockerfiles for DNS plugins will be so short, having only two additional layers, we probably could also ship with `docker-compose.yml` and / or in documentation give a choice of instructions for the user to build the image themselves for transparency.

bmw:
Automated repos are now on Docker Hub. See https://hub.docker.com/u/certbot/.

PR #5628, opened 2018-02-27
Add DNS Dockerfiles

bmw:
Fixes the pieces of #5328 we need to do before the release. Landing this before the release makes sure these Dockerfiles are in our next release tag which we can use on Docker Hub.

Currently, building these yourself is pretty awkward. With the version in `setup.py` in `master` pinned to 0.22.0.dev0 and Certbot on Docker Hub being 0.21.1, these files won't work without some help. You either need to modify `setup.py` or put these files on the 0.21.1 git tag.

This will get better over time though with #5490 and the suggestions [here](https://github.com/certbot/certbot/issues/5328#issuecomment-366896636) about how to build a rolling release.

### DIFF
--- a/certbot-dns-cloudflare/Dockerfile
+++ b/certbot-dns-cloudflare/Dockerfile
@@ -1,0 +1,5 @@
+FROM certbot/certbot
+
+COPY . src/certbot-dns-cloudflare
+
+RUN pip install --no-cache-dir --editable src/certbot-dns-cloudflare
--- a/certbot-dns-cloudxns/Dockerfile
+++ b/certbot-dns-cloudxns/Dockerfile
@@ -1,0 +1,5 @@
+FROM certbot/certbot
+
+COPY . src/certbot-dns-cloudxns
+
+RUN pip install --no-cache-dir --editable src/certbot-dns-cloudxns
--- a/certbot-dns-digitalocean/Dockerfile
+++ b/certbot-dns-digitalocean/Dockerfile
@@ -1,0 +1,5 @@
+FROM certbot/certbot
+
+COPY . src/certbot-dns-digitalocean
+
+RUN pip install --no-cache-dir --editable src/certbot-dns-digitalocean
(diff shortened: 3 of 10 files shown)


joohoi:
LGTM! 

One thing that comes to mind, is that if we would like to override the base image `ENTRYPOINT` with cli parameters defining the plugin that the `Dockerfile` is for. The main benefit being shorter commands for the user. The negative would be that the image couldn't be used to run other plugins. And the command line parameters getting diverged from the common documentation.

If, for example we'd have the following entrypoint: `ENTRYPOINT [ "certbot", "certonly", "--dns-cloudflare" ]` for the Cloudflare DNS Dockerfile.

The users CLI would be:
`docker run --rm -v /path/to/cloudflare_creds:/root/.cloudflare -it certbot/certbot-cloudflare -d wild.cat -d *.wild.cat`
Instead of:
`docker run --rm -v /path/to/cloudflare_creds:/root/.cloudflare -it certbot/certbot-cloudflare certonly --dns-cloudflare -d wild.cat -d *.wild.cat`

If you think that we should keep the CLI as it is, this PR is good to go.

bmw:
That's an interesting idea. What do you think? What is your preference?

bmw:
I don't think we should modify the `ENTRYPOINT`.

Adding a subcommand means that they can't use the image with `renew` or useful commands related to cert management like `certificates`, `delete`, or `revoke` without overriding the `ENTRYPOINT`.

I don't want to add `--dns-<provider>` because if you run `certbot certonly` without a plugin flag, it will interactively walk you through picking one, including standalone and webroot which someone might want to use for a different domain on the system. Even worse, it could cause problems for people who use the image with `certbot renew` because it will ignore the plugin that was initially being used to obtain the certificate and try to use the DNS plugin instead by default. With these things in mind, I don't think adding the flag buys us much and can make other tasks significantly harder.

While you gave me the go ahead to merge here, I'd like for us to be on the same page before merging this as it's not something we can (easily) change later.

joohoi:
I totally agree. Let's leave the entrypoint as it is. So LGTM, feel free to merge.

PR #5635, opened 2018-02-28
Add basic ACMEv2 integration tests

### LINKED ISSUE #5584: Set up ACMEv2 integration tests
We should largely be able to reuse the ACMEv1 tests with just changing the server URL and/or checked out version of boulder. We should add some tests for wildcards in addition to this.

bmw:
~~Start of #5584. Rest of it to come once #5620 is merged and we can test wildcard issuance.~~

Fixes #5584.

To provide a little more info since I don't think everyone who works on Certbot knows the horrors of shell scripting, [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) around `BOULDER_INTEGRATION` and `SERVER` is necessary because [tests/boulder-integration.sh](https://github.com/certbot/certbot/blob/fbace69b5e8fed5ee32cb029478d6b11aea84842/tests/boulder-integration.sh#L11) uses [set -u](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#The-Set-Builtin). This helps prevent bugs, but causes more complicated code to be written sometimes.

### DIFF
--- a/tests/boulder-integration.sh
+++ b/tests/boulder-integration.sh
@@ -432,7 +432,7 @@ done
 
 # Test ACMEv2-only features
 if [ "${BOULDER_INTEGRATION:-v1}" = "v2" ]; then
-    common -a manual -d *.le4.wtf,le4.wtf --preferred-challenges dns \
+    common -a manual -d '*.le4.wtf,le4.wtf' --preferred-challenges dns \
         --manual-auth-hook ./tests/manual-dns-auth.sh
 fi
 
--- a/tests/boulder-integration.sh
+++ b/tests/boulder-integration.sh
@@ -430,6 +430,12 @@ for path in $archive $conf $live; do
     fi
 done
 
+# Test ACMEv2-only features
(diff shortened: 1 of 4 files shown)


tests/boulder-integration.sh
```diff
@@ -430,6 +430,12 @@ for path in $archive $conf $live; do
     fi
 done
 
+# Test ACMEv2-only features
+if [ "${BOULDER_INTEGRATION:-v1}" = "v2" ]; then
+    common -a manual -d *.le4.wtf,le4.wtf --preferred-challenges dns \
```
schoen: I would suggest quoting the asterisk here (even though it will expand to an asterisk if there are no matching files, and there will probably never be any matching files)

bmw:
Thanks for the reviews!

I'll need one more approving review on GitHub for it to let me merge it.

bmw:
Oh Travis.

I put changes to fix this in https://github.com/certbot/certbot/pull/5641 for clarity because they are not related to the changes here.

PR #5636, opened 2018-02-28
Add user friendly wildcard error for ACMEv1

bmw:
Part of #5367.

Checking for wildcards in ACME doesn't have to worry about bytes, because the relevant documentation says the value must be unicode/str and none of Certbot's code paths violates this. (`acme.crypto_util._pyopenssl_cert_or_req_all_names` which is used in `BackwardsCompatibleClientV2` before calling this function returns unicode/str in Python 2/3 respectively.)

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -318,8 +318,8 @@ class Client(ClientBase):
 
         if identifier.value.startswith("*"):
             raise errors.WildcardUnsupportedError(
-                "Requesting an authorization for a wildcard"
-                " name is forbidden by the ACME protocol.")
+                "Requesting an authorization for a wildcard name is"
+                " forbidden by this version of the ACME protocol.")
 
         new_authz = messages.NewAuthorization(identifier=identifier)
         response = self._post(self.directory.new_authz, new_authz)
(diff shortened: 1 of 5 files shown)


certbot/client.py
```diff
@@ -313,6 +310,25 @@ def obtain_certificate(self, domains):
 
             return cert, chain, key, csr
 
+    def _get_order_and_authorizations(self, csr_pem, best_effort):
+        """Request an new order and complete its authorizations.
```
sydneyli: nit: an => a

certbot/client.py
```diff
@@ -313,6 +310,25 @@ def obtain_certificate(self, domains):
...
+
+        :param str csr_pem: A CSR in PEM format.
+        :param bool best_effort: True if failing to complete all
+            authorizations should raise an exception
```
sydneyli: True if failing to complete all authorizations should **not** raise an exception?

acme/acme/client.py
```diff
@@ -310,9 +310,17 @@ def request_challenges(self, identifier, new_authzr_uri=None):
...
+        :raises errors.WildcardUnsupportedError: if a wildcard is requested
+
         """
         if new_authzr_uri is not None:
             logger.debug("request_challenges with new_authzr_uri deprecated.")
+
+        if identifier.value.startswith("*"):
+            raise errors.WildcardUnsupportedError(
+                "Requesting an authorization for a wildcard"
+                " name is forbidden by the ACME protocol.")
```
ohemorange: nit: by this version of the ACME protocol

Issue #5564, opened 2018-02-10
Clearly document how to do DNS challenges with Certbot


bmw:
Suggested at https://community.letsencrypt.org/t/article-certbot-auto-with-dns-based-renewal/52695 and this will become a popular question with the ACMEv2 endpoint.

PR #5638, opened 2018-02-28
Suggest DNS authenticator when it's needed

bmw:
Fixes the 2nd bullet point of #5367.

For context, Let's Encrypt will currently only issue certificates for wildcard domains if people do their domain validation challenges over DNS. If we're about to error out because we can't satisfy the challenges and the only available challenge is the DNS challenge, suggest that they may need to use a plugin that supports DNS challenges.

The reason I didn't hardcode anything more specific here is that wildcard authorization having to be done over DNS is not a requirement of the ACME spec or the CA Baseline Requirements. Since the CAs Certbot is used with can change (either by using a new CA or a CA changing their policy), I just have a general suggestion about doing the challenge over DNS if it was the only option you were given.

### DIFF
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -433,7 +433,7 @@ def _find_smart_path(challbs, preferences, combinations):
         combo_total = 0
 
     if not best_combo:
-        _report_no_chall_path()
+        _report_no_chall_path(challbs)
 
     return best_combo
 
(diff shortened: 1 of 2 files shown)


bmw:
Thanks for the review!

> What do you think about linking to documentation / suggesting --manual in the error message?

Good suggestion!

I'd be up for the former in the long run, but I unfortunately don't know what we'd link people to right now. We have #5564 tracking improving (or creating) some documentation online around our DNS plugins, but we don't really have any at the moment as they really haven't been something most people can install and use.

As for manual, it is a way to do challenges over DNS that ships with Certbot, but I don't think it's appropriate for most people. Either they have to write a script to automate their renewal or they have to manually do the challenge every 3 months, neither of which is a good solution for the average user in my opinion.

My personal preference would be to leave this as is for now and after #5564 is resolved, look into adding a link. I'm happy to create a new issue to track doing this.

PR #5639, opened 2018-03-01
Remove leading *. from default cert name.

bmw:
Part of #5367.

We handle stripping `*.` in `client.py` rather than in `storage` and its utilities so we know whether we're using `domains` or a user provided cert name. If someone runs Certbot with `--cert-name *.example.org`, we should respect their insane wish rather than suddenly changing it on them.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -338,7 +338,14 @@ class Client(object):
                 "Non-standard path(s), might not work with crontab installed "
                 "by your operating system package manager")
 
-        new_name = certname if certname else domains[0]
+        if certname:
+            new_name = certname
+        elif util.is_wildcard_domain(domains[0]):
+            # Don't make files and directories starting with *.
+            new_name = domains[0][2:]
+        else:
+            new_name = domains[0]
+
         if self.config.dry_run:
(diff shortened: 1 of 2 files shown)


certbot/client.py
```diff
@@ -338,7 +338,14 @@ def obtain_and_enroll_certificate(self, domains, certname):
                 "Non-standard path(s), might not work with crontab installed "
                 "by your operating system package manager")
 
-        new_name = certname if certname else domains[0]
```
pde: this is fine though maybe keeping the work in a util function like

`new_name = certname if certname else util.strip_wildcards(domains[0])`

might be cleaner?

bmw: My personal feeling is stuff in `util` should generally be useful to multiple pieces of the project. Do we currently have another use case in mind for a function like this? Unless we have one now, my vote would be to leave it out of `util` until we do.

I'm happy to add a separate function to do this in `client.py` if that seems cleaner though.

pde: Seems slightly cleaner but it's your call about whether it's worth the time :)

bmw:
Going to merge this now, but I'll submit a follow up PR which updates the documentation you noticed and refactors things a bit.

PR #5650, opened 2018-03-01
Improve style around choosing lineagename

bmw:
@pde had some stylistic concerns around #5639 which this PR addresses. I updated the docstring and split some of the code into a separate function. Rather making the code look something like:
```
if certname:
    new_name = certname
else:
    new_name = _strip_wildcard(domains[0])
...
def _strip_wildcard(domain):
    return domain[2:] if util.is_wildcard_domain(domains) else domain
```
I further split up the function moving the logical piece of determining the name for the lineage into its own function. This splits up `obtain_and_enroll_certificate` even more and better documents at a high level what is being done.

### DIFF
--- a/certbot/client.py
+++ b/certbot/client.py
@@ -337,9 +337,10 @@ class Client(object):
         authenticator and installer, and then create a new renewable lineage
         containing it.
 
-        :param list domains: Domains to request.
-        :param plugins: A PluginsFactory object.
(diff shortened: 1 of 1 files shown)


bmw:
Also, while I personally prefer `lineage_name` to `lineagename`, I used the latter for consistency because that's what's used in `storage.py`

Issue #5472, opened 2018-01-24
Can you request a single cert with a wildcard and the base domain with our DNS plugins


bmw:
There was some concern from @joohoi about whether or not we currently handle this correctly in any/all of our DNS plugins. @joohoi, if you could add some additional context here that'd be great. 

We should look into this issue soon as we're currently in the process of getting our DNS plugins packaged everywhere and we should try to avoid making distros immediately repackage them.

joohoi:
Yeah, for this to work, we would need to add two `TXT` records to a single subdomain in the DNS zone, ie: your certificate needs names `yourdomain.tld` and `*.yourdomain.tld`, the record that the CA will check is same for both of these: `_acme-challenge.yourdomain.tld` hence a query to that subdomain should return two `TXT` records.

The actual credit for raising this issue goes to @sahsanu for raising the issue up in another context.

joohoi:
Our DNS plugins do this intelligently, and everything should be in order. This includes cleanup as well, the records are cleaned based on their actual content.

bmw:
I just hit this problem when using the Google DNS plugin when testing #5620. Log is: 
[letsencrypt.log.txt](https://github.com/certbot/certbot/files/1753963/letsencrypt.log.txt)

trinopoty:
I think it's better to keep the two TXT records for with and without wildcard of different name. Changing the name of the wildcarded version to something like `_acme-challenge-wildcard` would indicate to the user that it's for a wildcard domain and keep problems like these (with Google DNS, etc) to a minimum.

joohoi:
> I just hit this problem when using the Google DNS plugin when testing

I originally thought this was a limitation of Google Cloud DNS API, but it turns out that Google _does_ in fact support adding multiple TXT records with the same name, they just have to be created in a single API request. This conflicts with the way that DNS plugins currently handle the record creation.

With our DNS plugin API, the only way to manage this would be to first read TXT records for the validation domain, then deleting the old one(s), and finally writing the list of pre-existing TXT records with the addition of the current validation token. The API call for addition takes a list of strings, which in our case would of course be the list of validation tokens. This is done here https://github.com/certbot/certbot/blob/master/certbot-dns-google/certbot_dns_google/dns_google.py#L117

bmw:
> With our DNS plugin API, the only way to manage this would be to first read TXT records for the validation domain, then deleting the old one(s), and finally writing the list of pre-existing TXT records with the addition of the current validation token.

[...]

PR #5652, opened 2018-03-02
Allow Google DNS plugin to write multiple TXT record values

joohoi:
Google API requires all the records in a RRset to be set as a list on a single call. To achieve this, this PR adds automatic check of pre-existing records, deletes the old ones, and adds them to the list with the new one created before making the API call.

Fixes an issue mentioned in comments of #5472

### DIFF
--- a/certbot-dns-google/certbot_dns_google/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/dns_google.py
@@ -223,7 +223,7 @@ class _GoogleClient(object):
         :rtype: `list` of `string`
 
         """
-        rrs_request = self.dns.resourceRecordSets()
+        rrs_request = self.dns.resourceRecordSets()  # pylint: disable=no-member
         request = rrs_request.list(managedZone=zone_id, project=self.project_id)
         response = request.execute()
         # Add dot as the API returns absolute domains
--- a/certbot-dns-google/certbot_dns_google/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/dns_google.py
@@ -70,7 +70,7 @@ class Authenticator(dns_common.DNSAuthenticator):
         self._get_google_client().add_txt_record(domain, validation_name, validation, self.ttl)
 
     def _cleanup(self, domain, validation_name, validation):
-        self._get_google_client().del_txt_record(domain, validation_name, validation, self.ttl)
+        self._get_google_client().del_txt_record(domain, validation_name, self.ttl)
 
     def _get_google_client(self):
         return _GoogleClient(self.conf('credentials'))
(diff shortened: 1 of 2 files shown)


certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -107,14 +107,22 @@ def add_txt_record(self, domain, record_name, record_content, record_ttl):
 
         zone_id = self._find_managed_zone_id(domain)
 
+        record_contents = self.get_existing_txt_rrset(zone_id, record_name)
+
+        # We need to remove old records and re-add them as a part of the list
+        if record_contents:
+            self.del_txt_record(domain, record_name, record_ttl)
```
bmw: nit: This update isn't atomic and there's a period of time where all of the TXT records have been deleted. It looks like [the API supports atomic updates](https://cloud.google.com/dns/api/v1/changes#resource). If this is the case, can we modify this request to add deletions to `data` below?

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -154,14 +161,18 @@ def del_txt_record(self, domain, record_name, record_content, record_ttl):
...
+
         data = {
             "kind": "dns#change",
             "deletions": [
                 {
                     "kind": "dns#resourceRecordSet",
                     "type": "TXT",
                     "name": record_name + ".",
-                    "rrdatas": [record_content, ],
+                    "rrdatas": record_contents,
```
bmw: nit: Similar to my above comment, this deletes all records rather than the one that Certbot is asking it to. I believe this could be fixed fairly easily by adding `additions` containing `record_contents` with `record_content` removed.

joohoi: This is now done

certbot-dns-google/certbot_dns_google/dns_google.py
joohoi: Fixed

joohoi:
Ready for the next round. Updates are now atomic.

joohoi:
Do you know what's going on with the tests? I'm not able to reproduce that linter problem locally.

bmw:
Going to review latest commit shortly, but it looks like lint is failing.

bmw:
Unfortunately, I'm also unable to reproduce locally and I have no idea why, but I understand the problem.

Google is building these objects and determining their attributes and methods dynamically based on metadata about the endpoint. If you search Google's Python libraries source code, you'll find no reference to `resourceRecordSets` and instead this is being returned by the Google server which the library interprets to create the object. `pylint` has no way of understanding this so it thinks the attribute doesn't exist. If you look at all the other attributes we use on this object, such as `changes` and `managedZones`, we also disable `no-member` there as well to make `pylint` happy.  Disabling this check in `pylint` for `resourceRecordSets` seems safe with this in mind and should fix the problem.

PR #5654, opened 2018-03-03
Ensure fullchain_pem in the order is unicode/str

### LINKED ISSUE #5606: fullchain_pem should be set to a string in ClientV2 and BackwardsCompatibleClientV2
Part of #5367.
`fullchain_pem` in the order is bytes; set the `fullchain_pem` property of the order resource to a string

bmw:
Fixes #5606.

For context, our concern here was with `acme` having a public API that is used by other projects, we shouldn't change types like this after we ship it. With my understanding of best practices as described in #4516, we should be giving the user `unicode` in Python 2 and a `str` in Python 3, however, much of Certbot's code currently requires `bytes`. The quick fix for this that Erica and I had in mind that allows Certbot to work was to make sure the `acme` API is right, but immediately convert back to `bytes` for Certbot.

I did this after the function in `certbot.crypto_util` because what's part of the public API in Certbot isn't exactly clear and something like `crypto_util` could be used by 3rd party plugins. Our own plugins currently use it. Because of this, I let it take the "proper" types as well and threw the conversion into `certbot.client.`

You can see integration tests passing with this code for every version of Python we support for both ACME versions at https://travis-ci.org/certbot/certbot/builds/348519650. Once this lands, I plan on adding tests like this to the `test-everything` branch.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -795,8 +795,8 @@ class BackwardsCompatibleClientV2(object):
                     'certificate, please rerun the command for a new one.')
 
             cert = OpenSSL.crypto.dump_certificate(
-                    OpenSSL.crypto.FILETYPE_PEM, certr.body.wrapped)
-            chain = crypto_util.dump_pyopenssl_chain(chain)
+                    OpenSSL.crypto.FILETYPE_PEM, certr.body.wrapped).decode()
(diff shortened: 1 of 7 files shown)


joohoi:
LGTM!

PR #5655, opened 2018-03-03
Break lockstep between our packages

### LINKED ISSUE #5490: Break lockstep between our packages

bmw:
Fixes #5490.

There's a lot of possibilities discussed in #5490, but I'll try and explain what I actually did here as succinctly as I can. Unfortunately, there's a fair bit to explain. My goal was to break lockstep and give us tests to ensure the minimum specified versions are correct without taking the time now to refactor our whole test setup.

To handle specifying each package's minimum acme/certbot version, I added a requirements file to each package. This won't actually be included in the shipped package (because it's not in the `MANIFEST`).

After creating these files and modifying `tools/pip_install.sh` to use them, I created a separate tox env for most packages (I kept the DNS plugins together for convenience). The reason this is necessary is because we currently use a single environment for each plugin, but if we used this approach for these tests we'd hit issues due to different installed plugins requiring different versions of acme/certbot. There's a lot more discussion about this in #5490 if you're interested in this piece. I unfortunately wasted a lot of time trying to remove the boilerplate this approach causes in `tox.ini`, but to do this I think we need negations described at [complex factor conditions](https://tox.readthedocs.io/en/latest/config.html#complex-factor-conditions) which hasn't made it into a `tox` release yet.

The biggest missing piece here is how to make sure the oldest versions that are currently pinned to master get updated. Currently, they'll stay pinned that way without manual intervention and won't be properly testing the oldest version. I think we should solve this during the larger test/repo refactoring after the release because the tests are using the correct values now and I don't see a simple way around the problem.

Once this lands, I'm planning on updating the `test-everything` tests to do integration tests with the "oldest" versions here.


bmw:
Done! Should be ready for a final review.

joohoi:
LGTM!

PR #5661, opened 2018-03-04
ACMEv2 support for Route53 plugin

joohoi:
Route53 plugin needed similar changes than Google Cloud DNS plugin, namely including the full contents of to-be-updated challenge RRset. Also when cleaning up the records, we can't use `DELETE` verb if there are multiple resource records in the RRset, but need to update the record instead.


bmw:
Closing in favor of #5672.

PR #5668, opened 2018-03-05
Fix merge conflicts and add more ACMEv2 integration tests

bmw:
`test-everything` tests are failing again due to the `.travis.yml` changes in #5655. This PR fixes this and adds more ACMEv2 integration tests.

See #5645 for more information on how to review PRs to this branch in general. We switch to doing "oldest" integration tests against the `py27-nginx-oldest` branch because the nginx plugin needs to be installed for our integration tests to pass. See:

https://github.com/certbot/certbot/blob/559220c2eff90975fb671945d780c2a757d0e167/tests/boulder-integration.sh#L441

**IMPORTANT:** This PR must be merged and not squashed or we will still have merge conflicts with `master`.

PR #5669, opened 2018-03-05
Don't report coverage on Apache during integration tests

bmw:
The integration tests don't actually test any Apache functionality and trying to test this for our "oldest" tests without Apache installed caused failures. See #5668.

EDIT: I used 67% as the minimum as I sometimes got [68% during testing](https://travis-ci.org/certbot/certbot/jobs/349535456) and I'd like to avoid issues like those in #5641.


bmw:
Once this has landed, you should be able to kick #5668 to have tests pass.

PR #5677, opened 2018-03-06
Disable not-callable for iter_entry_points

ohemorange:

PR #5678, opened 2018-03-06
Fallback without dns.resourceRecordSets.list permission

bmw:
Tests may fail until #5677 is merged. Been fighting to make Coveralls report everything passed.

Since `dns.resourceRecordSets.list` is a new permission used by the plugin, we now try to continue without it if the request fails. I had to change things around a bit because if the request fails while trying to add a record, we should just assume there are currently no records and if the request fails when trying to delete a record, we should assume the record we're trying to delete still exists.

Also, I choose not to keep track of records internally like we do for route53 because it's only needed for wildcards so it isn't a regression and doing so adds complexity and is unlikely to matter because Google Cloud only recently started allowing people to set fine grained permissions like this.

### DIFF
--- a/certbot-dns-google/certbot_dns_google/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/dns_google.py
@@ -108,6 +108,8 @@ class _GoogleClient(object):
         zone_id = self._find_managed_zone_id(domain)
 
         record_contents = self.get_existing_txt_rrset(zone_id, record_name)
+        if record_contents is None:
+            record_contents = []
         add_records = record_contents[:]
 
         if "\""+record_content+"\"" in record_contents:
@@ -176,6 +178,8 @@ class _GoogleClient(object):
             return
 
         record_contents = self.get_existing_txt_rrset(zone_id, record_name)
+        if record_contents is None:
+            record_contents = [record_content]
 
         data = {
             "kind": "dns#change",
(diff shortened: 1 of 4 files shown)


bmw:
Adding the `pylint: disable` commit here may not be the most correct, but I'd like both of these PRs to land sooner rather than later and Coveralls wasn't cooperating.

bmw:
Looks like we're having the same coveralls problem...

You can see all the lines I changed still have coverage by looking at https://travis-ci.org/certbot/certbot/jobs/349990410#L1048. The lines missing coverage are also doing so in `master` as you can see in https://travis-ci.org/certbot/certbot/jobs/349864564#L1040 (although the line number has shifted).

Unless anyone has a strong objection, for the sake of getting this in, I'll just force the PR through without waiting for coveralls to get fixed once I get a +1 on the PR.

certbot-dns-google/certbot_dns_google/dns_google.py
```diff
@@ -216,23 +220,32 @@ def get_existing_txt_rrset(self, zone_id, record_name):
...
-            for rr in response["rrsets"]:
-                if rr["name"] == record_name and rr["type"] == "TXT":
-                    return rr["rrdatas"]
-        return []
+        try:
+            response = request.execute()
+        except googleapiclient_errors.Error:
+            logger.info("Unable to list existing records. If you're "
+                        "requesting a wildcard certificate, this might not work.")
+            logger.debug("Error was:", exc_info=True)
```
joohoi: This error gets a bit "lost" in the very verbose output of the library we're using. It's a noop when not requesting a certificate with wildcard + base domain, so `logger.warn` is not suitable for use here. However if the user does request a wildcard + base domain certificate, what would you think about raising a flag here, and if the flag is present when failing, including this message on a `logger.warn` level with the actual API error?

joohoi:
LGTM, the log message comment was dismissed in OOB discussion as bringing too much complexity to address the error messages for each individual domain separately, and the file logging has a cleaner output for this.

PR #5687, opened 2018-03-07
fix(acme): client._revoke sends default content_type

### LINKED ISSUE #5684: `acme` module's `client._revoke` sends incorrect `ContentType` header.
[ACME draft-10, section 6.2](https://tools.ietf.org/html/draft-ietf-acme-acme-10#section-6.2) explicitly requires POSTs be sent with a `Content-Type` of `application/jose+json`: 

> Because client requests in ACME carry JWS objects in the Flattened
> JSON Serialization, they must have the "Content-Type" header field
> set to "application/jose+json".  If a request does not meet this
> requirement, then the server MUST return a response with status code
> 415 (Unsupported Media Type).

The good news is that Certbot & the `acme` module already do this in 99.99% of cases :tada: :+1: :balloon: 

The bad news is that `client._revoke` overrides the default (correct) `ContentType` with `none`: https://github.com/certbot/certbot/blob/77fdb4d7d6194989dcc775f2e0ad81b6147c2359/acme/acme/client.py#L231

This will make revocation requests fail in a world where this restriction is enforced (See https://github.com/letsencrypt/boulder/pull/3532) 

There doesn't seem to be a reason to want to override `ContentType` in this case so I believe the fix is to change `_revoke` to use the default.

Thanks!

sydneyli:
Fixes #5684.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -227,8 +227,7 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         response = self._post(url,
                               messages.Revocation(
                                 certificate=cert,
-                                reason=rsn),
-                                content_type=None)
+                                reason=rsn))
         if response.status_code != http_client.OK:
             raise errors.ClientError(
                 'Successful revocation must return HTTP OK status')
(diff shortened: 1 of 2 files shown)


cpu:
Thank you @sydneyli :1st_place_medal: :racing_car:

PR #5688, opened 2018-03-07
Update the test-everything branch

bmw:
This fixes the current [docker_dev failures](https://travis-ci.org/certbot/certbot/builds/350462813).

EDIT: This PR should be merged and not squashed to prevent merge conflicts.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -227,8 +227,7 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         response = self._post(url,
                               messages.Revocation(
                                 certificate=cert,
-                                reason=rsn),
-                                content_type=None)
+                                reason=rsn))
         if response.status_code != http_client.OK:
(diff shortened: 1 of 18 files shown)


bmw:
> Do we actually want to merge this to master?

Not at all. Thanks for catching that.

> Also: I caught an error in our Dockerfile-dev that basically affects the le-auto tests. Namely the Docker image is missing curl, and this makes tests/modification-check.sh fail. However our travis tests don't seem to complain about that, but I can't figure out why. I stumbled upon this when running docker-compose run --rm --service-ports development bash -c tox -e lint locally. This probably should be done in a separate PR though.

Adding `curl` to `Dockerfile-dev` is probably a good idea, but I think the reason it doesn't fail for the `le_auto` tests is each of those tests have their own Dockerfile in `letsencrypt-auto-source`.

joohoi:
LGTM!

PR #5690, opened 2018-03-08
Fix --allow-subset-of-names

### LINKED ISSUE #5685: --allow-subset-of-names no longer works

bmw:
Fixes #5685 and adds integration tests to prevent regressions.

The issue is now we sometimes use an `AuthHandler` instance multiple times when previously it was only used once. Continuously adding to the list of authorizations and trying to solve the challenges again makes boulder mad (it errors saying we're trying to modify a failed challenge), so let's limit the scope of `aauthzrs` to the `handle_authorizations` method.

@ohemorange, if you have time to quickly review this, it's valuable as you're the most familiar with the changes here, but if not @joohoi, it'd be great if you could take a look. All we should be changing is we now pass the variable around rather than keeping it on the class and adding tests so it should be pretty straightforward.

### DIFF
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -34,8 +34,6 @@ class AuthHandler(object):
     :ivar account: Client's Account
     :type account: :class:`certbot.account.Account`
 
-    :ivar aauthzrs: ACME Authorization Resources and their active challenges
-    :type aauthzrs: `list` of `AnnotatedAuthzr`
     :ivar list pref_challs: sorted user specified preferred challenges
         type strings with the most preferred challenge listed first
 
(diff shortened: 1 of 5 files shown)


certbot/auth_handler.py
```diff
@@ -134,40 +131,43 @@ def _solve_challenges(self):
...
+            self._poll_challenges(aauthzrs, chall_update, best_effort)
         finally:
-            self._cleanup_challenges(active_achalls)
+            self._cleanup_challenges(aauthzrs, active_achalls)
 
-    def _send_responses(self, resps, chall_update):
+    def _send_responses(self, aauthzrs, resps, chall_update):
         """Send responses and make sure errors are handled.
 
+        :param aauthzrs: authorizations and the selected annotated challenges
```
joohoi: Not really in the scope of this PR, but the `resps` arg goes (also previously) undocumented for this internal method. This probably should be fixed in a subsequent PR.

PR #5692, opened 2018-03-08
Allow _acme-challenge in a separate zone

noci2012:
Like described here:
https://github.com/lukas2511/dehydrated/wiki/example-dns-01-nsupdate-script

Not using this patch may be an issue if the parent zone has been (where a wildcard certificate has been requested.) signed by DNSSEC.

Please consider this also for inclusion before dns-01 will be allowed for wildcards.


noci2012:
@bodgit 
Maybe you want to look into this one?

noci2012:
can do.

noci2012:
See new pull request: https://github.com/certbot/certbot/pull/5707

PR #5694, opened 2018-03-08
Update README.rst

noci2012:
Add description of credentials file.


bodgit:
https://github.com/certbot/certbot/blob/master/certbot-dns-rfc2136/certbot_dns_rfc2136/__init__.py

noci2012:
ah did try to find that ....
took good part of a day to find where the troubles might be.
As this info wasn't shown in the past.

Ok the documentatio is in the __init__.py

Thanks, that should make this change obsolete.

noci2012:
This is a related issue: https://github.com/certbot/certbot/pull/5692

noci2012:
this is not needed

PR #5697, opened 2018-03-09
Update dns_rfc2136.py to allow hostnames as well as IPv4 and IPv6.

noci2012:
Add hostname translation..


noci2012:
Closed in favour of #5664 
which has better error checks now.

PR #5707, opened 2018-03-12
Allow _acme-challenge as a zone

noci2012:
Like described here:
https://github.com/lukas2511/dehydrated/wiki/example-dns-01-nsupdate-script

Not using this patch may be an issue if the parent zone has been (where a wildcard certificate has been requested.) signed by DNSSEC.

Please consider this also for inclusion before dns-01 will be allowed for wildcards.

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -106,7 +106,7 @@ class _RFC2136Client(object):
         :raises certbot.errors.PluginError: if an error occurs communicating with the DNS server
         """
 
-        domain = self._find_domain(domain_name)
+        domain = self._find_domain(record_name)
 
         n = dns.name.from_text(record_name)
         o = dns.name.from_text(domain)
@@ -142,7 +142,7 @@ class _RFC2136Client(object):
         :raises certbot.errors.PluginError: if an error occurs communicating with the DNS server
         """
 
-        domain = self._find_domain(domain_name)
+        domain = self._find_domain(record_name)
 
         n = dns.name.from_text(record_name)
         o = dns.name.from_text(domain)
@@ -167,17 +167,17 @@ class _RFC2136Client(object):
             raise errors.PluginError('Received response from server: {0}'
                                      .format(dns.rcode.to_text(rcode)))
 
-    def _find_domain(self, domain_name):
+    def _find_domain(self, record_name):
         """
         Find the closest domain with an SOA record for a given domain name.
 
-        :param str domain_name: The domain name for which to find the closest SOA record.
+        :param str record_name: The record name for which to find the closest SOA record.
         :returns: The domain, if found.
         :rtype: str
         :raises certbot.errors.PluginError: if no SOA record can be found.
         """
 
-        domain_name_guesses = dns_common.base_domain_name_guesses(domain_name)
+        domain_name_guesses = dns_common.base_domain_name_guesses(record_name)
 
         # Loop through until we find an authoritative SOA record
         for guess in domain_name_guesses:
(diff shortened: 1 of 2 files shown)


noci2012:
@joohoi, a new Pull request with request changes from  #5692

noci2012:
Now the add/del txt record routines have an unused parameter domain_name.
Not something i can fix without changing the API

noci2012:
@joohoi ,
So fixing changing to only record_name  leads to another problem with unused parameters.

joohoi:
> So fixing changing to only record_name leads to another problem with unused parameters.
What would be an acceptable method for that?

You can remove the arg (and accompanied pydoc) from both `add_txt_record` and `del_txt_record`methods. You'll most likely have to fix a few tests as well as result of the changed args.

noci2012:
Done that, that just moves the issue up one level to _perform & _cleanup.

noci2012:
@joohoi : new update.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -71,10 +71,12 @@ def _setup_credentials(self):
         )
 
     def _perform(self, domain, validation_name, validation):
-        self._get_rfc2136_client().add_txt_record(domain, validation_name, validation, self.ttl)
+        self._get_rfc2136_client().add_txt_record(validation_name, validation, self.ttl)
+        dummy = domain
```
joohoi: Instead of this, you can prepend the `domain` argument in the `perform()` with underscore to mark it as unused.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -71,10 +71,12 @@ def _setup_credentials(self):
...
 
     def _cleanup(self, domain, validation_name, validation):
-        self._get_rfc2136_client().del_txt_record(domain, validation_name, validation)
+        self._get_rfc2136_client().del_txt_record(validation_name, validation)
+        dummy = domain
```
joohoi: Same thing here, `_domain` will do the trick.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
joohoi: Another ways to make pylint ignore a rule for a method would be for example:
```
def _perform(self, domain, validation_name, validation):
    #  pylint: disable=unused-argument
    ...
```
or
```
def _perform(self, domain, validation_name, validation):  #  pylint: disable=unused-argument
    ...
```

but adding the underscore is the correct approach here.

noci2012:
Done. suggested changes.

zjs:
Not to hijack this PR, but as a question directed to no one in particular: Do similar issues exist with other DNS plugins? If so, it would be good to file issue(s) to ensure they don't get forgotten.

Potentially problematic code used by most other DNS plugins:
https://github.com/certbot/certbot/blob/1d0e3b1bfa5afaa861adc1a4157e6a94d34320a2/certbot/plugins/dns_common.py#L319-L335

(Note that updating this function would likely warrant re-testing each plugin that uses it since it's conceivable that someone's API doesn't play nicely with the leading underscore in `_acme-challenge`.)

noci2012:
@zjs: probably the issue is with all others as well, i can only test/use the rfc2136 one.
This is a issue because of a described dehydrated configuration, which also is about rfc2136,

noci2012:
@joohoi any other changes needed/

joohoi:
This LGTM!

While I feel that the point by @zjs is on spot, I think the `base_domain_name_guesses()` handles the thing correctly, and that we should leave the decision to the actual plugin calling it instead of hardcoding stuff in the `dns_common.py`. These changes for other plugins should be implemented separately.

PR #5710, opened 2018-03-12
Add documentation about DNS plugins and Docker

bmw:


docs/install.rst
```diff
@@ -115,13 +115,17 @@ these make much sense to you, you should definitely use the
...
+to, `install Docker`_, then issue a command like the one found below. If
+you are using Certbot with the :ref:`Standalone` plugin, you will need
+to make the port it uses accessible from outside of the container by
+including something like ``-p 80:80`` on the command line before
+``certbot/certbot``.
 
 .. code-block:: shell
 
-   sudo docker run -it --rm -p 443:443 -p 80:80 --name certbot \
+   sudo docker run -it --rm --name certbot \
```
ohemorange: why remove `443`?

docs/install.rst
```diff
@@ -131,6 +135,16 @@ Running Certbot with the ``certonly`` command will obtain a certificate and plac
 within Docker, you must install the certificate manually according to the procedure
 recommended by the provider of your webserver.
 
+There are also Docker images for each of Certbot's DNS plugins available
+at https://hub.docker.com/u/certbot which automate doing domain
+validation over DNS for popular providers. To use one, just replace
+``certbot/certbot`` in command above with the name of the image you want
```
ohemorange: in the command above

docs/using.rst
```diff
@@ -54,12 +54,18 @@ standalone_ Y    N    | Uses a "standalone" webserver to obtain a certificate.
                       | Requires port 80 or 443 to be available. This is useful on    tls-sni-01_ (443)
                       | systems with no webserver, or when direct integration with
                       | the local webserver is not supported or not desired.
+|dns_plugs| Y    N    | This category of plugins automates obtaining a certificate by dns-01_ (53)
+                      | modifying DNS records to prove you have control over a
+                      | domain. Doing domain validation in this way is
+                      | necessary to obtain certificates from Let's Encrypt.
```
ohemorange: necessary to obtain wildcard certificates I assume

ohemorange: but maybe you actually mean "...is the only way to obtain wildcard certificates..."

docs/using.rst
```diff
@@ -164,6 +172,34 @@ the Internet on the specified port using each requested domain name.
...
+-----------
+
+If you'd like to obtain a wildcard certificate from Let's Encrypt or run
+``certbot`` on a machine other than your target webserver, you can use one of
+Certbot's DNS plugins.
+
+These plugins are still in the process of being packaged
+by many distributions and cannot currently be installed with ``certbot-auto``.
+If, however, you are comfortable installing the certificates yourself and/or
+setting up a more complex automated renewal process, you can run these plugins
```
ohemorange: "setting up a different renewal process than using `certbot renew`" ? it's not actually so much harder to script it yourself.

docs/install.rst
bmw: Because `tls-sni` is dead.

I also tried to be a little general about the ports needed which I think is fine due to all of the warnings about using Certbot in Docker in the beginning of this section.

docs/using.rst
bmw: Thanks for noticing the missing word "wildcard".

What's the difference between "is necessary to obtain wildcard certificates" and "is the only way to obtain wildcard certificates"?

docs/using.rst
bmw: In my mind I was thinking about the renewal craziness around restarting your webserver, but I didn't add anything about that in this PR so I should probably just simplify this sentence too.

docs/using.rst
ohemorange: more down to earth real people talk

ohemorange: also only vs. necessary aren't exactly the same

ohemorange: I don't care that much though

docs/install.rst
ohemorange: but it's not dead for existing users

bmw: I don't think existing users are reading how to set it up again though. I can include "or ``-p 443:443``" in the sentence above this if you prefer.

ohemorange: yeah that sounds good

bmw: done

docs/using.rst
bmw: http://www.reactiongifs.com/wp-content/uploads/2013/11/real.gif

Issue #5110, opened 2017-09-13
Connection failure to letsencrypt servers after doing a long challenge with Route53


hundt:
As the log below shows, when doing a Route53 verification with 18 different domains, a connection to https://acme-staging.api.letsencrypt.org fails with `ECONNRESET`. In the same environment, I can successfully get a certificate for two domains. I suspect this has something to do with the Keep-Alive connection to acme-staging.api.letsencrypt.org established at the beginning of the process but then not used for almost 17 minutes--the next time it tries to use that connection is when it fails.

## My operating system is (include version):
    $ uname -a
    Linux ip-10-9-0-232 3.10.0-327.10.1.el7.x86_64 #1 SMP Tue Feb 16 17:03:50 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
    $ cat /etc/redhat-release
    CentOS Linux release 7.2.1511 (Core) 
    $ certbot --version
    certbot 0.18.1

Python packages installed (2.7):

    ConfigArgParse-0.12.0 PyOpenSSL-17.2.0 acme-0.18.1 asn1crypto-0.22.0 boto-2.48.0 boto3-1.4.7 botocore-1.7.9 certbot-0.18.1 certbot-dns-route53-0.18.1 certifi-2017.7.27.1 cffi-1.10.0 chardet-3.0.4 configobj-5.0.6 cryptography-2.0.3 docutils-0.14 enum34-1.1.6 funcsigs-1.0.2 future-0.16.0 futures-3.1.1 idna-2.6 ipaddress-1.0.18 jmespath-0.9.3 mock-2.0.0 parsedatetime-2.4 pbr-3.1.1 pycparser-2.18 pyrfc3339-1.0 python-dateutil-2.6.1 pytz-2017.2 pyyaml-3.12 requests-2.18.4 s3transfer-0.1.11 six-1.10.0 urllib3-1.22 zope.component-4.4.0 zope.event-4.3.0 zope.interface-4.4.2

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
    pip

## I ran this command and it produced this output:

[...]

PR #5712, opened 2018-03-12
Restore parallel waiting to Route53 plugin

### LINKED ISSUE #5671: route53: Bring back parallel waiting for propagation
The route53 plugin used to be able to send several update requests, and wait on them all in parallel. It looks like we lost that ability in https://github.com/certbot/certbot/pull/4771, possibly due to a bad rebase. We should bring back that capability. Waiting for the server to declare that it's finished propagating is very slow, and so parallelizing that waiting speeds things up a *lot*.

jsha:
Currently, when you issue for multiple domains using the Route53 plugin, Certbot
waits for each update to propagate before moving onto the next. This change
makes Certbot send all updates, then wait for them all at the same time, which
speeds up issuance for multiple domains dramatically.

Note that this means the Route53 plugin uses slightly less of the dns_common
functionality. In particular, it implements the base Authenticator method `perform`,
rather than the DNSAuthenticator method `_perform` (which performs a single
challenge).

Fixes #5671

### DIFF
--- a/certbot-dns-route53/certbot_dns_route53/dns_route53.py
+++ b/certbot-dns-route53/certbot_dns_route53/dns_route53.py
@@ -45,15 +45,16 @@ class Authenticator(dns_common.DNSAuthenticator):
     def perform(self, achalls):
         self._attempt_cleanup = True
 
-        change_ids = [
-            self._change_txt_record("UPSERT",
-              achall.validation_domain_name(achall.domain),
-              achall.validation(achall.account_key))
-            for achall in achalls
-        ]
+        try:
+            change_ids = [
+                self._change_txt_record("UPSERT",
+                  achall.validation_domain_name(achall.domain),
+                  achall.validation(achall.account_key))
+                for achall in achalls
+            ]
 
-        for change_id in change_ids:
-            self._wait_for_change(change_id)
+            for change_id in change_ids:
+                self._wait_for_change(change_id)
         except (NoCredentialsError, ClientError) as e:
             logger.debug('Encountered error during perform: %s', e, exc_info=True)
             raise errors.PluginError("\n".join([str(e), INSTRUCTIONS]))
--- a/certbot-dns-route53/certbot_dns_route53/dns_route53.py
+++ b/certbot-dns-route53/certbot_dns_route53/dns_route53.py
@@ -42,7 +42,7 @@ class Authenticator(dns_common.DNSAuthenticator):
     def _setup_credentials(self):
         pass
 
-    def _perform(self, achalls): # pylint: disable=missing-docstring
+    def _perform(self, domain, validation_domain_name, validation): # pylint: disable=missing-docstring
         pass
 
     def perform(self, achalls):
--- a/certbot-dns-route53/certbot_dns_route53/dns_route53.py
+++ b/certbot-dns-route53/certbot_dns_route53/dns_route53.py
@@ -18,7 +18,6 @@ INSTRUCTIONS = (
     "https://boto3.readthedocs.io/en/latest/guide/configuration.html#best-practices-for-configuring-credentials "  # pylint: disable=line-too-long
     "and add the necessary permissions for Route53 access.")
 
-# pylint: disable=abstract-method
 @zope.interface.implementer(interfaces.IAuthenticator)
 @zope.interface.provider(interfaces.IPluginFactory)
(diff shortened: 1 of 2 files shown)


certbot-dns-route53/certbot_dns_route53/dns_route53.py
```diff
@@ -42,15 +42,30 @@ def more_info(self):  # pylint: disable=missing-docstring,no-self-use
...
+                self._wait_for_change(change_id)
         except (NoCredentialsError, ClientError) as e:
             logger.debug('Encountered error during perform: %s', e, exc_info=True)
             raise errors.PluginError("\n".join([str(e), INSTRUCTIONS]))
 
+        # Sleep for at least the TTL, to ensure that any records cached by
+        # the ACME server after previous validation attempts are gone. In
+        # most cases we'll need to wait at least this long for the Route53
+        # records to propagate, so this doesn't delay us much.
+        time.sleep(self.ttl)
```
zjs: Disclaimer: it has a been a while since I've looked at this, and I don't have a test system handy anymore.

I'm not sure this sleep is necessary. In most cases, I'd expect validation to occur very infrequently relative to the TTL (on the order of months vs. the order of minutes). Because records are deleted at the end of each execution, I wouldn't expect records from a previous challenge to still be cached (and if the absence of a record were being cached, that would be [determined by the TTL on the SOA record](https://tools.ietf.org/html/rfc2308), not our `self.ttl` value here, which seems like something that would cause major problems for our workflow since SOA TTLs might be very high).

Once `_wait_for_change` has returned for each `change_id`, it seems like we should be good to ask the server to validate the challenges.

There are some edge cases around, for example, having a pool of machines behind a load balancer which each order certificates for the same domain on a similar schedule. However, that seems like a less common use case and retrying in cases like that is something we should be handling in a generic way (since the issue could occur with any DNS provider).

jsha: Updated to remove the unnecessary sleep, thanks.

sydneyli:
@jsha The linter is failing here-- It looks like `DNSAuthenticator` expects `_perform` to be defined?

Timbus:
Just checking in to say this is probably more important than you might think.
I had to manually apply this patch because certbot is unusable on our AWS infrastructure without it. See issue #5110 for details.. (Although this fix mostly just sidesteps the problem).

jsha:
Thanks @sydneyli! Fixed.

jsha:
Ping!

certbot-dns-route53/certbot_dns_route53/dns_route53.py
```diff
@@ -18,6 +18,7 @@
     "https://boto3.readthedocs.io/en/latest/guide/configuration.html#best-practices-for-configuring-credentials "  # pylint: disable=line-too-long
     "and add the necessary permissions for Route53 access.")
 
+# pylint: disable=abstract-method
```
sydneyli: Maybe define a dummy `_perform` similar to `_setup_credentials` instead of suppressing the warnings? @bmw  pointed out that if we ever decide to have `common.Plugin` properly extend `ABCMeta`, then [we wouldn't be able to instantiate this class](https://docs.python.org/3/library/abc.html#abc.abstractmethod).

jsha: Done. I feel a little weird about private methods being abstract, now that I think about it; I'll pick some brains in person when I get a chance.

sydneyli:
There's one more `disable=abstract-method` outstanding, and also the linter is unhappy that `_perform` doesn't have the correct number of arguments. You'll probably need a `disable=unused-arguments` there to keep pylint from complaining there too.
Thanks @jsha !!

certbot-dns-route53/certbot_dns_route53/dns_route53.py
```diff
@@ -42,14 +42,26 @@ def more_info(self):  # pylint: disable=missing-docstring,no-self-use
     def _setup_credentials(self):
         pass
 
-    def _perform(self, domain, validation_domain_name, validation):
-        try:
-            change_id = self._change_txt_record("UPSERT", validation_domain_name, validation)
+    def _perform(self, achalls): # pylint: disable=missing-docstring
```
sydneyli: The parameters here should be `self`, `domain`, `validation_domain_name`, and `validation` to make the linter happy.

PR #5713, opened 2018-03-13
Add docs about --server

bmw:


bmw:
We can update this to include LE's ACMEv2 URL when it is public.

docs/using.rst
```diff
@@ -801,6 +801,27 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not
...
+
+By default, Certbot uses Let's Encrypt's initial production server at
+https://acme-v01.api.letsencrypt.org/. You can tell Certbot to use a
+different CA by providing ``--server`` on the command line or in a
+:ref:`configuration file <config-file>` with the URL of the server's
+ACME directory. For example, to test against Let's Encrypt's ACMEv2
+staging server, you would add ``--server
+https://acme-staging-v02.api.letsencrypt.org/directory`` to the command
+line. Certbot will automatically change the version of the ACME protocol
+used based on the contents served at the provided URL.
```
ohemorange: Certbot will automatically ~change the~**select which** version of the ACME protocol to use based on the contents served at the provided URL.

docs/using.rst
```diff
@@ -801,6 +801,27 @@ Example usage for DNS-01 (Cloudflare API v4) (for example purposes only, do not
...
+
+If you use ``--server`` to specify an ACME CA that implements a newer
+version of the spec, you may want to obtain a certificate for a wildcard
```
ohemorange: you may want to --> you will be able to

bmw: I used "you may be able to" because it depends on the CA's policy.

bmw:
I added a couple new changes such as documenting `--server` when talking about running the DNS plugins in Docker and using the production ACMEv2 as the example rather than the staging one.

bmw:
@ohemorange, I'm not blocking the initial website update on this, but if you could prioritize reviewing this when you get a chance, I'd appreciate it as not setting the `--server` flag has been a source of problems on the community forum.

docs/using.rst
ohemorange: cool

ohemorange: cool

Issue #5729, opened 2018-03-14
In manual authenticator, explain that earlier challenges shouldn't be replaced by later ones


schoen:
Per https://community.letsencrypt.org/t/wildcard-base-domain-two-txts-timeout-problem/56105/3, it's not obvious to users that a later DNS challenge isn't meant to *replace* an earlier DNS challenge. This makes sense because Certbot doesn't explicitly explain at what point the CA is going to validate the challenges; a user could easily assume that the CA validated the old challenge right away and that it's no longer needed. (In fact, Certbot could probably even have been implemented that way!)

So, the text in the manual authenticator should probably explicitly say something like "Don't remove, disable, or replace the earlier challenges yet" when telling the user to set up each new challenge for a subsequent name.

schoen:
Fixed by #6370.

PR #5739, opened 2018-03-16
Edit UI documentation in manual authenticator

eicksl:
For manual authentication, this clarifies that earlier challenges
should not be replaced by later ones.

Fixes (#5729).

### DIFF
--- a/certbot/plugins/manual.py
+++ b/certbot/plugins/manual.py
@@ -68,14 +68,21 @@ class Authenticator(common.Plugin):
         'secret key needed to complete the TLS handshake is located at '
         '$CERTBOT_KEY_PATH. An additional cleanup script can also be '
         'provided and can use the additional variable $CERTBOT_AUTH_OUTPUT '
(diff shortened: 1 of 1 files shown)


eicksl:
@schoen Let me know if this is sufficient. I also updated the instructions for the HTTP-01 challenge with a similar message. I'm unsure how the TLS-SNI challenge method is implemented for multiple domains, and considering that this method is currently [disabled for most new issuances](https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811) I decided to leave it alone for now.

schoen:
Hi @eicksl , sorry for the duplicated effort -- I have a PR of my own that I'm working on on this topic, which I would probably prefer to use because I made it print a different message on subsequent challenges than on the first one, which I think will be slightly clearer for the user. But I might use some of your explanatory text in my PR if that's all right.

eicksl:
@schoen Sounds good. I'll go ahead and close this PR in that case.

Issue #2552, opened 2016-02-25
Non standard path(s), might not work with crontab installed by your operating system package manager


pde:
It's really weird to get this message if you aren't using OS packages...

mitar:
Yes. How can I make this quiet? Because now I am getting in my own cron all the time to my email. :-)

pde:
@mitar running with `--quiet` /  `-q` *should* suppress that message?

mitar:
@pde: [This is what I use](https://github.com/tozd/docker-nginx-proxy/blob/master/letsencrypt/dockergen.sh#L23) and it does not suppress it.

mitar:
But it is true. It seems to be output only when a key is generated, if I have it in cron and nothing has to be done (no keys are to be renowed) it is quiet. Only when a key is generated this additional warning comes out.

ezekiel:
Would really like to see this message squelched by `--quiet`.

ezekiel:
One simple way to handle this is adjusting the message level from "Warning" to "Info".
To this end, I have submitted pull request #5743 

If this simple method is not suitable, I would really appreciate further guidance, and I will be happy to adjust/enhance the implementation.

schoen:
@pde, would you be willing to opine on @ezekiel 's proposed solution?

ezekiel:
output from `--help` for various commands gives the following info for `--quiet`:

> Silence all output except errors. Useful for automation via cron. Implies --non-interactive.

As such, perhaps it is a better choice for me to track down the code to make that accurate and to silence all messages below "Error" level, rather than decrease a "Warning" message to an "Info" message to silence it.

ezekiel:
Found it: the constant called `QUIET_LOGGING_LEVEL` is currently set to `WARNING`, I have another branch prepared with a very simple change for this constant (to `ERROR`). I can submit this as a new pull request, which may be a better choice overall.

ohemorange:
Hey @pde, any opinions here?

pde:
My sense is that many of those are serious, fatal or almost-always-fatal conditions that should be flagged despite running in cron. So if we were going to take #5773 we should probably also define semantics for these levels formally and refactor many of the `WARNING`s to `ERROR`s.

PR #5743, opened 2018-03-16
Adjust file paths message from Warning to Info.

ezekiel:
In regards to issue #2552, allow `--quiet` to squelch the message.


schoen:
Thanks @ezekiel , I've asked @pde what he thinks about your solution over in the issue discussion.

ezekiel:
I would suggest that PR #5773 is likely a better choice overall, as brings the logging functionality in-line with the output from `--help`, and would still resolve issue #2552.

pde:
Let's take this for now.

pde:
Closes: #2552

PR #5750, opened 2018-03-17
Fix acme.Client.__init__ (#5738)

### LINKED ISSUE #5738: [PATCH] >=0.22.0: Bug in acme.Client.__init__ breaks acme.Client.register
Hi!

There is a bug in `acme.Client.__init__`. When called with `net=None`, `self.net` is first initialited with a new instance of `ClientNetwork` but then `net` (rather than `self.net`) is pass to `ClientBase.__init__` which sets `self.net` back to None. The breaking code was introduced in 9baf75d6c887a9b69bd92c7bffe585f091c9b08e and is this:
```
class ClientBase(object):  # pylint: disable=too-many-instance-attributes
    [..]
    def __init__(self, directory, net, acme_version):
        [..]
        self.net = net
[..]
class Client(ClientBase):
    [..]
    def __init__(self, directory, key, alg=jose.RS256, verify_ssl=True,
                 net=None):
        [..]
        self.net = ClientNetwork(key, alg=alg, verify_ssl=verify_ssl) if net is None else net
        [..]
        super(Client, self).__init__(directory=directory,
            net=net, acme_version=1)
```
As a result, calling `.register(...)` on the client will raise `AttributeError: 'NoneType' object has no attribute 'post'` which breaks wile, see https://github.com/costela/wile/issues/34.

One way to fix the issue is to change `net=net` to `self.net=net` in the line using `super`. I could provide a pull request to do that if that's the way you want to patch it.

What do you think?

hartwork:
First version to fix #5738 for review. I'm happy to make further adjustments if needed.

Best, Sebastian

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -259,11 +259,13 @@ class Client(ClientBase):
         """
         # pylint: disable=too-many-arguments
         self.key = key
-        net = ClientNetwork(key, alg=alg, verify_ssl=verify_ssl) if net is None else net
+        if net is None:
+            net = ClientNetwork(key, alg=alg, verify_ssl=verify_ssl)
 
         if isinstance(directory, six.string_types):
             directory = messages.Directory.from_json(
                 net.get(directory).json())
+
         super(Client, self).__init__(directory=directory,
             net=net, acme_version=1)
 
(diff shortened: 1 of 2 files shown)


hartwork:
Seen #5748 too late, didn't have my mail client open. No problem.

PR #5754, opened 2018-03-18
Add note to developer guide docs about installing docs extras. (#4946)

edaleeta:
This PR updates the Developer Guide docs with a brief note about how to install the docs extras in case a contributor skips the Getting Started setup.

This should be helpful for those that want to contribute to docs without installing all of the dev requirements through `./tools/venv.sh`.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -376,6 +376,9 @@ commands:
 This should generate documentation in the ``docs/_build/html``
 directory.
 
+.. note:: If you skipped the "Getting Started" instructions above,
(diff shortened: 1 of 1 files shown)


edaleeta:
Thanks for pointing that out, @schoen!

Updated the command to include the quotes to prevent bash from interpreting the syntax.
Also scanned the rest of the docs to see if similar bracketed forms were used, and the rest look okay!

edaleeta:
Noticed this was still open - anything else needed here, @schoen?

schoen:
Whoops, sorry about that delay and thanks for the PR!

PR #5755, opened 2018-03-18
Add a note about installing certbot DNS plugins

gtd:
Nowhere in the documentation does it explain that plugins are simply pip packages, the implication here was that you must use Docker to use these plugins, but that is not the case.

I actually went through quite a bit of pain trying to get docker up and running on Mac OS, but once I looked at the Dockerfile I realized installing the plugin was trivial and fit my workflow very nicely.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -185,7 +185,8 @@ Certbot's DNS plugins.
 These plugins are still in the process of being packaged
 by many distributions and cannot currently be installed with ``certbot-auto``.
 If, however, you are comfortable installing the certificates yourself,
-you can run these plugins with :ref:`Docker <docker-user>`.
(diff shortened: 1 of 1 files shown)


schoen:
I think this is a good idea; @bmw, are you OK with it?

bmw:
No, I do not think we should recommend this and [installing Certbot in this way is not supported](https://certbot.eff.org/docs/install.html#installing-from-source).

@gtd, I'm glad it worked for you, but probably the single most common problem people get when trying to install Certbot is they have a mix of Python files from their OS package manager and `pip` leading to mismatched versions of packages and broken modules. We're actively working on better ways to get Certbot's DNS plugins to people and I agree the current process is pretty painful, but I don't think we should open people up to this new set of problems.

I'm closing this issue, but @schoen, if you think I should reconsider, feel free to reopen.

PR #5769, opened 2018-03-20
Postfix: enhancement update

sydneyli:
installing the policies is now an enhancement to the postfix plugin.


sydneyli:
@jgillula i'll probably just merge this directly since @bmw will start his review later this week.

PR #5773, opened 2018-03-20
Adjust QUIET_LOGGING_LEVEL to ERROR, matching --help output.

ezekiel:
This will squelch the WARNING-level message about Non-standard paths.

issue #2552


bmw:
Warning level messages have always been shown by Certbot for a long time and there's a lot of code that uses this assumption. I think the behavior trying to be changed here was fixed in #5743 so I'm going to close this PR, but if you wanted it for another reason, please comment and I'll reopen.

Issue #4310, opened 2017-03-07
document how different display modes interact with force_interactive


void-elf:
... because `force_interactive` doesn't actually force interactivity in non-interactive mode (it just uses the default instead), which is unintuitive and isn't written down anywhere.

kevinkle:
If I remember what @bmw said correctly, the `force_interactive` flag isn't absolute as it checks for an actual console.

bmw:
Correct. The high level goal of `force_interactive` has a one line description in `certbot/interfaces.py`. To briefly expand on that, it's to prevent workflow regressions for people who are getting away with running certbot non-interactively without `--non-interactive`.

To see how the implementation of this and what exactly it's doing, `certbot.display.util.FileDisplay._can_interact`.

bmw:
This (or https://github.com/certbot/certbot/issues/3915) still needs to be done.

PR #5777, opened 2018-03-21
Postfix plugin

### LINKED ISSUE #4672: Port STARTTLS Everywhere into a Postfix plugin
This is work we've started, and are now committed to finishing:

https://github.com/EFForg/starttls-everywhere/blob/master/letsencrypt-postfix/PostfixConfigGenerator.py#L293

sydneyli:
Fixes #4672. First pass at a certbot plugin for postfix. Adds an Installer for Postfix, with an enhancement for translating the starttls-everywhere policy into postfix configs.

Going to be adding some more tests (and appeasing travis), but definitely ready for a first pass.


sydneyli:
Note: There's a TON of commits here! They're mostly all from the OG [EFForg/starttls-everywhere](https://github.com/EFForg/starttls-everywhere/) repo, from where the original work was migrated.

sydneyli:
@bmw this is ready for a first pass review. Let me know if you have any questions about it!

bmw:
I've started my review on this. Please do not push to this branch or even worse rebase it until I've finished my review.

I've reviewed all of the Python code for `certbot-postfix` including tests except `constants.py` and have ~40 inline comments so far. I still need to review the changes to `certbot`, integration tests, and I have a list of ~10 higher level or functional things I still want to look into.

I'll pick this back up again next week.

certbot-postfix/certbot_postfix/postconf.py
```diff
@@ -0,0 +1,155 @@
...
+        """Reads initial parameter state from main.cf
+        """
+        out = self._get_output()
+        for name, value in _parse_main_output(out):
+            if not value:
+                value = ""
+            self._db[name] = value
+        out = self._get_output('-P') # get master parameters
+        for name, value in _parse_main_output(out):
+            service, param_name = name.rsplit("/")
```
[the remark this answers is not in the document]
sydneyli: Ah. It looks like I meant to call `rsplit("/", 1)` here, to get "service/type", "parameter". Adding test cases for this.

sydneyli:
Thanks for the in-depth review @bmw! I'm :+1:-ing comments as I complete them as a bookkeeping measure, and writing up longer responses to some of the higher-level concerns you had.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,305 @@
...
+
+        """
+        # pylint: disable=unused-argument
+        self.save_notes.append("Configuring TLS for {0}".format(domain))
+        self.postconf.set("smtpd_tls_cert_file", cert_path,
+                          check_override=util.report_master_overrides)
+        self.postconf.set("smtpd_tls_key_file", key_path,
+                          check_override=util.report_master_overrides)
+        self._set_vars(constants.DEFAULT_SERVER_VARS)
+        self._set_vars(constants.DEFAULT_CLIENT_VARS)
```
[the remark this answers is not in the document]
sydneyli: I think we definitely want to have some of these settings to be optional. I asked Ben for what kind of granularity an organization or admin might expect.

The client/server split might be important-- but he also pointed out in particular that ciphersuites are highly org-dependent, so orgs might not want those to be overwritten. Would love to go over the UX-flow in person, but here are my initial thoughts.
Something like:
```
Here are the parameter changes we're going to apply to `main.cf`. Is that okay? Y/N
```
If N, display all the flags you can apply to exclude certain params from being set. These flags might include:
```
--tls-only      # Only sets certs & enables opportunistic TLS.
--server-only   # Only changes params on smtpd_*
```

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,305 @@
...
+        self.policy = None
+        self.postfix = None
+        self.policy_file = None
+        self._enhance_func = {"starttls-policy": self._enable_policy_list}
+
+    def _ensure_ca_certificates_exist(self):
+        # TODO (sydneyli): Ensure `ca-certificates` is installed correctly, or that
+        # /etc/ssl/certs/ even has certificates in it, probably via a sanity check using
+        # `openssl` command?
+        pass
```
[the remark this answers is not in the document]
sydneyli: Yep. In fact, if client encryption isn't set to `verify` or `secure` (which we don't do for them by default), this only affects the STARTTLS policy enhancement. I'm thinking of kicking the enhancement to a separate PR, so we can get the preliminary installer landed, at least!

Good to know re:`openssl`-- I'm open to alternatives to depending on `ca-certificates`.

certbot-postfix/setup.py
```diff
@@ -0,0 +1,59 @@
...
+install_requires = [
+    'acme=={0}'.format(version),
+    'certbot=={0}'.format(version),
+    # For pkg_resources. >=1.0 so pip resolves it to a version cryptography
+    # will tolerate; see #2599:
+    'setuptools>=1.0',
+    'six',
+    'zope.interface',
+]
+
```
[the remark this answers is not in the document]
sydneyli: I looked into doing this via `dependency_links` for some time, it seems that it can't accept a Github link unless `setup.py` is in the root folder. Once it's on `pypi` proper, there should be less of an issue.

I'll try out option 1 for now.

sydneyli:
@bmw I'm going to pull out the `STARTTLS Everywhere` enhancement into its own PR, spawn some child issues to track it and other discussions, and try to land the Postfix certificate installer code first. Most of the high-level discussions we might want to have are around the STARTTLS Everywhere enhancement, especially due to its dependency on `ca-certificates`/`openssl`. Let me know if you have any issues with this.

certbot-postfix/certbot_postfix/constants.py
```diff
@@ -0,0 +1,42 @@
...
+# Should NOT use SSLv2/3.
+ACCEPTABLE_TLS_VERSIONS = ("TLSv1", "TLSv1.1", "TLSv1.2")
+
+# Default variables for a secure MTA server [receiver].
+DEFAULT_SERVER_VARS = {
+    "smtpd_tls_mandatory_protocols": "!SSLv2, !SSLv3",
+    "smtpd_tls_protocols": "!SSLv2, !SSLv3",
+    "smtpd_tls_security_level": ACCEPTABLE_SECURITY_LEVELS,
+    "smtpd_tls_ciphers": ACCEPTABLE_CIPHER_LEVELS,
+    "smtpd_tls_eecdh_grade": "strong",
```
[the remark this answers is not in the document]
sydneyli: Yep, it is. Thanks for the catch! I noted this at some point, but forgot to follow up on it.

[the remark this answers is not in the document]
bmw: Thanks for verifying it is an issue.

Without further research, I'm not positive what size param we should use (although I suspect it's 2048 bit) and I'm not sure the format the file needs to be in for Postfix, but the code I mentioned in my initial comment is in the `certbot.common.Installer` class in case it's useful.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,305 @@
...
+        except ImportError:
+            raise ImportError('STARTTLS Everywhere policy Python module not installed!')
+        if options is None:
+            policy = policy.Config()
+        else:
+            policy = policy.Config(options)
+        policy.load()
+        util.write_domainwise_tls_policies(policy, self.policy_file)
+        policy_cf_entry = "texthash:" + self.policy_file
+        self.postconf.set("smtp_tls_policy_maps", policy_cf_entry)
```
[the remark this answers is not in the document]
bmw: I modified the `tls_policy_maps` file after Certbot exited. `postfix check` not failing doesn't mean there's a problem. I just expected Postfix to complain about the syntax errors when parsing the file. I didn't try sending an email to any of the domains in the policy file to see if it was properly taking effect.

bmw:
We talked about this out of band but just to say it publicly as well, I have no issues with ripping out the enhancement code for now and landing it in another PR.

certbot-postfix/setup.py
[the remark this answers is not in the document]
bmw: Option 1 with a requirements file until `starttls-policy` is shipped sounds the best to me. If you're removing the STARTTLS Everwhere enhancement though, I think you can kick this to that PR.

I spent a little time looking at `dependency_links` and apparently using it requires special flags to `pip` and installing packages in this way is deprecated. See https://github.com/pypa/pip/issues/3939.

certbot-postfix/certbot_postfix/installer.py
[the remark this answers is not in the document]
bmw: Good point about client encryption.

re: re: openssl, I'd like us to spend a little time looking at alternatives, but if after a bit of research there isn't another easy option, shelling out to `openssl` is fine.

certbot-postfix/certbot_postfix/constants.py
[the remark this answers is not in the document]
bmw: Whoops. Correct path is `certbot.plugins.common.Installer`. The Postfix Installer in this PR already inherits from this class.

sydneyli:
@bmw There's an odd import failure in py34/py36, but this is ready for re-review otherwise.

bmw:
That's some pretty unhelpful error output, but I think the problem is the call to `print` missing parentheses on line 247 of `util.py` causing a syntax error. This is a lot more clear in the Python 3.6 error output, but fixing this resolved errors for me locally.

Testing on Python 3.4 I got this error though:
```
    @mock.patch('certbot_postfix.util.PostfixUtilBase._get_output')
    def test_flush(self, mock_out):
        self.config.set('default_parameter', 'new_value')
        self.config.set('extra_param', 'another_value')
        self.config.flush()
        mock_out.assert_called_with(
>           ['-e', 'default_parameter=new_value', 'extra_param=another_value'])

certbot-postfix/certbot_postfix/tests/postconf_test.py:79: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/py34/lib/python3.4/site-packages/mock/mock.py:937: in assert_called_with
    six.raise_from(AssertionError(_error_message(cause)), cause)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = AssertionError("Expected call: _get_output(['-e', 'default_parameter=new_value', 'extra_param=another_value'])\nActual call: _get_output(['-e', 'extra_param=another_value', 'default_parameter=new_value'])",), from_value = None

>   ???
E   AssertionError: Expected call: _get_output(['-e', 'default_parameter=new_value', 'extra_param=another_value'])
E   Actual call: _get_output(['-e', 'extra_param=another_value', 'default_parameter=new_value'])

<string>:2: AssertionError
```
Might just be a hesienbug, but if the order of the params to `_get_output` isn't well defined, we should probably modify this test slightly to make sure it doesn't cause spurious errors for people.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,289 @@
...
+        """Rollback saved checkpoints.
+
+        :param int rollback: Number of checkpoints to revert
+
+        :raises .errors.PluginError: If there is a problem with the input or
+            the function is unable to correctly revert the configuration
+
+        """
+        super(Installer, self).rollback_checkpoints(rollback)
+        self.postconf = postconf.ConfigMain(self.conf('config-utility'), self.conf('config-dir'))
```
bmw: We need to pass in a value for `handle_overrides` here.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,289 @@
...
+        self.postconf.flush()
+
+        del self.save_notes[:]
+
+        if title and not temporary:
+            self.finalize_checkpoint(title)
+
+    def recovery_routine(self):
+        super(Installer, self).recovery_routine()
+        self.postconf = postconf.ConfigMain(self.conf('config-utility'), self.conf('config-dir'))
```
bmw: We need to pass in a value for `handle_overrides` here.

certbot-postfix/certbot_postfix/util.py
```diff
@@ -0,0 +1,254 @@
...
+def _has_acceptable_tls_versions(parameter_string):
+    """
+    Checks to see if the comma-separated list of TLS protocols to exclude is acceptable.
+    Sample string: "!SSLv2, !SSLv3"
+    """
+    bad_versions = list(constants.TLS_VERSIONS)
+    for version in constants.ACCEPTABLE_TLS_VERSIONS:
+        del bad_versions[bad_versions.index(version)]
+    for bad_version in bad_versions:
+        if "!" + bad_version not in parameter_string:
```
bmw: What if `parameter_string` lists protocols to include rather than those to exclude? This seems allowed according to docs like http://www.postfix.org/postconf.5.html#smtpd_tls_protocols.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,289 @@
...
+        """Confirming outstanding updates for configuration parameters.
+
+        :raises errors.PluginError: when user rejects the configuration changes.
+        """
+        updates = self.postconf.get_changes()
+        output_string = "Postfix TLS configuration parameters to update in main.cf:\n"
+        for name, value in six.iteritems(updates):
+            output_string += "{0} = {1}\n".format(name, value)
+        output_string += "Is this okay?\n"
+        if not zope.component.getUtility(interfaces.IDisplay).yesno(output_string, default=False):
```
bmw: I think we probably want `default=True` here because otherwise when you're running Certbot with -n, --non-interactive, or --noninteractive on the command line we raise the error below.

You can also set `force_interactive=True` if you want. The name of this flag is horrible (I swear it made sense in my head when I wrote it...see #4310), but its main purpose and default value of `False` is to be extra conservative when adding new prompts to existing workflows. Since this is a new plugin, we don't need to be conservative and can always try to prompt when non-interactivity wasn't requested.

bmw:
Just making sure you saw this, it looks like the mypy and lint tests are unhappy.

For mypy, we made some changes to make it check and error on more things. I'd encourage you to try and fix these errors as I don't think it should take very long and some of them might be bugs, but if you want to go back to the old behavior for the Postfix plugin for now, you can add:
```
[mypy-certbot_postfix.*]
check_untyped_defs = False
```
to `mypy.ini` in the root of the repo.

sydneyli:
@bmw other than a small mypy error I have to look into, this is ready for re-review.
Sorry if merging in master mungles the commit history a bit-- needed it to test out the new mypy updates properly.

The primary changes:
 - Instead of passing in a Callable to the postconf wrapper for handling overridden parameters from master.cf, it accepts a flag.
 - The TLS protocol parameter parsing/detection (accounting for included protocols as well as excluded, and accounting for mixing them together)
 - docs seem to actually work now

certbot-postfix/certbot_postfix/tests/installer_test.py
```diff
@@ -0,0 +1,316 @@
...
+        mock_util().yesno.return_value = True
+        from certbot_postfix import constants
+        with create_installer(self.config) as installer:
+            installer.prepare()
+
+            # pylint: disable=protected-access
+            installer.deploy_cert("example.com", "cert_path", "key_path",
+                                  "chain_path", "fullchain_path")
+            changes = installer.postconf.get_changes()
+            expected = {} # type: Dict[str, Union[Tuple[str, str], str]]
```
bmw: I think the `Tuple` here should be `Tuple[str, ...]` because the tuple can have an arbitrary number of values.

Also, are you planning on always using tuples in `constants.py` for single values? I think that's fine, but if you are I think you can remove the `Union` here.

certbot-postfix/certbot_postfix/tests/installer_test.py
```diff
@@ -0,0 +1,316 @@
...
+            expected.update(constants.TLS_SERVER_VARS)
+            expected.update(constants.DEFAULT_SERVER_VARS)
+            expected.update(constants.DEFAULT_CLIENT_VARS)
+            self.assertEqual(changes["smtpd_tls_key_file"], "key_path")
+            self.assertEqual(changes["smtpd_tls_cert_file"], "cert_path")
+            for name, value in six.iteritems(expected):
+                if isinstance(value, tuple):
```
bmw: Are you planning on always using tuples in `constants.py` for single values? I think that's fine, but if you are I think the body of this loop can simply be `self.assertEqual(changes[name], value[0])` because all the values are tuples now.

certbot-postfix/certbot_postfix/installer.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        return certbot_util.get_filtered_names(self.postconf.get(var)
+                   for var in ('mydomain', 'myhostname', 'myorigin',))
+
+
+    def _set_vars(self, var_dict):
+        """Sets all parameters in var_dict to config file.
+        """
+        for param, acceptable in six.iteritems(var_dict):
+            if not util.is_acceptable_value(param, self.postconf.get(param), acceptable):
```
bmw: Are you planning on always using tuples in `constants.py` for single values? I think that's fine, but if you are I think the body of this if statement can simply be `self.postconf.set(param, acceptable[0], acceptable)` because all the values are tuples now.

certbot-postfix/certbot_postfix/util.py
```diff
@@ -0,0 +1,285 @@
...
+        message = "Cannot find executable '{0}'.".format(exe)
+    if not (certbot_util.exe_exists(exe) or plugins_util.path_surgery(exe)):
+        raise errors.NoInstallationError(message)
+
+def report_master_overrides(name, overrides, acceptable_overrides=None):
+    """If the value for a parameter |name| is overridden by other services,
+    report a warning to notify the user.
+
+    :param str name: The name of the parameter that is being overridden.
+    :param list overrides: The values that other services are setting for |name|.
```
bmw: We should replace the uses of `|name|` in this docstring with \`name\`.

certbot-postfix/certbot_postfix/util.py
```diff
@@ -0,0 +1,285 @@
...
+
+    :raises .NoInstallationError: when the executable isn't found
+
+    """
+    if message is None:
+        message = "Cannot find executable '{0}'.".format(exe)
+    if not (certbot_util.exe_exists(exe) or plugins_util.path_surgery(exe)):
+        raise errors.NoInstallationError(message)
+
+def report_master_overrides(name, overrides, acceptable_overrides=None):
```
bmw: This functions behavior when `name` refers to TLS protocols is a little unusual. `acceptable_overrides` still needs to be set for it for it to consider other values, but `acceptable_overrides` isn't used.

Maybe we could refactor things a bit, however, I think we should at least add a comment describing this behavior.

certbot-postfix/certbot_postfix/util.py
```diff
@@ -0,0 +1,285 @@
...
+            continue
+        error_string += "  {0}: {1}\n".format(service, value)
+    if error_string:
+        raise errors.PluginError("{0} is overridden with less secure options by the "
+             "following services in master.cf:\n".format(name) + error_string)
+
+def is_acceptable_value(parameter, value, acceptable):
+    """ Returns whether the `value` for this `parameter` is acceptable,
+    given a string or tuple `acceptable`
+
```
bmw: If `tls_protocols` is in `parameter`, `acceptable` isn't used but it's still required to be set.

Maybe we could refactor things a bit and move this `tls_protocols` check outside of this function, however, I think we should at least add a comment describing this behavior.

sydneyli: Yeah, I agree it's weird. I thought about changing the variable mappings in `constants.py` so that they explicitly define acceptability (maybe as a `callable(str) => bool` ) and default value separately for each parameter, since the tuples are being used for both. This would involve a larger refactor.

So I just added a comment, here and for `report_master_overrides`.

certbot-postfix/local-oldest-requirements.txt
```diff
@@ -0,0 +1,2 @@
+-e acme[dev]
```
bmw: Now that acme 0.25.0 is out, this can be `acme==0.25.0`. I should have been more descriptive in my comment as this change just landed for the other files in master.

sydneyli:
> We should merge rather than squash this PR to preserve authorship, but to keep our git history clean, I'd like us to try and rebase this PR to be the history of the starttls-everywhere repo, a single commit from me, and then a single commit from you.

Sure. I've gotten it down to:
 - history from starttls-everywhere repo (assuming the merge from this is the `starttls-everywhere/certbotify` merge)
 - a commit from you
 - a commit from me
 - a merge commit

I could also continue trying to rebase the entire thing and be rid of the final merge commit, but while trying to rebase this history onto current master, I've running into merge conflicts from the original repo rather than Certbot. Do you have a preference here?

bmw:
I think that's totally fine. Thanks for doing that!

sydneyli:
Ok. Opening a separate PR for that to not lose the comment/PR history in this very long correspondence, so closing this one.

PR #5779, opened 2018-03-21
fix(auth_handler): cleanup is always called

### LINKED ISSUE #5429: Authenticator cleanup not always called
In `auth_handler.py`, all code starting with the call to `perform` on the plugin and ending when the challenges can be cleaned up in the normal case should be wrapped in a `with error_handler.ErrorHandler(self._cleanup_challenges):` (or an equivalent) block. With the way this currently works, `_cleanup_challenges` would be called just outside the end of the `with` statement for the case when an exception/signal doesn't occur, but this could/should probably be refactored to have this handled by the `with` statement.

### LINKED ISSUE #5539: Certbot failure leaves standalone server running
Originally reported at https://community.letsencrypt.org/t/renewal-process-crashed-my-production-server/51920/8, we should make sure that we always properly stop the standalone server. This is probably related to #5429. We should reproduce the problem and verify the solution to #5429 fixes it in the simple case.

We may also want look into adding code that more aggressively kills threads and child processes when Certbot tries to exit.

We should at least look into this problem and see if it's a regression for 0.22.0.

sydneyli:
Fixes #5429 
Fixes #5539 
Tests incoming!

### DIFF
--- a/certbot/auth_handler.py
+++ b/certbot/auth_handler.py
@@ -67,9 +67,9 @@ class AuthHandler(object):
         config = zope.component.getUtility(interfaces.IConfig)
         notify = zope.component.getUtility(interfaces.IDisplay).notification
 
-        with error_handler.ExitHandler(self._cleanup_challenges, aauthzrs):
-            # While there are still challenges remaining...
-            while self._has_challenges(aauthzrs):
+        # While there are still challenges remaining...
+        while self._has_challenges(aauthzrs):
+            with error_handler.ExitHandler(self._cleanup_challenges, aauthzrs):
                 resp = self._solve_challenges(aauthzrs)
                 logger.info("Waiting for verification...")
                 if config.debug_challenges:
@@ -79,8 +79,8 @@ class AuthHandler(object):
                 # Send all Responses - this modifies achalls
                 self._respond(aauthzrs, resp, best_effort)
 
-            # Just make sure all decisions are complete.
-            self.verify_authzr_complete(aauthzrs)
+        # Just make sure all decisions are complete.
+        self.verify_authzr_complete(aauthzrs)
 
         # Only return valid authorizations
         retVal = [aauthzr.authzr for aauthzr in aauthzrs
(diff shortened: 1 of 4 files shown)


sydneyli:
Added tests. In particular, `test_respond_error` reproduces the bug in https://community.letsencrypt.org/t/renewal-process-crashed-my-production-server/51920/8 .

certbot/error_handler.py
```diff
@@ -70,8 +70,12 @@ def __exit__(self, exec_type, exec_value, trace):
         self.body_executed = True
         retval = False
         # SystemExit is ignored to properly handle forks that don't exec
-        if exec_type in (None, SystemExit):
+        if exec_type is SystemExit:
             return retval
+        elif exec_type is None:
+            if not self.call_on_regular_exit:
+                return retval
+            retval = True
```
bmw: I'm not sure it matters, but if an exception wasn't raised, why do we want to we want to return `True` if we call the registered functions and `False` if we don't? If it doesn't matter, I'd personally leave `retval` alone and not set it here.

certbot/tests/error_handler_test.py
```diff
@@ -47,6 +47,10 @@ def setUp(self):
         self.handler = error_handler.ErrorHandler(self.init_func,
                                                   *self.init_args,
                                                   **self.init_kwargs)
+        self.exit_handler = error_handler.ExitHandler(self.init_func,
```
bmw: This is definitely a nit, but it might be kind of nice to:

1. Have a separate test class for `ExitHandler` to simplify the classes and make it clear what functionality is tested where.
2. Ensure all of the other functionality of `ExitHandler` works now and in the future.

Doing this might be a bit tricky due to the way tests are automatically found and run, but one way would be to:

1. Make a `ExitHandlerTest` class that inherits from this class.
2. In `ExitHandlerTest`, set `self.handler` to an `ExitHandler`.
3. Change the name of the methods you're adding to something like `test_regular_exit` and override this in `ExitHandler`.

This is only a suggestion so feel free to push back and/or ignore it, but I think it should be pretty easy and might be worth it.

certbot/auth_handler.py
```diff
@@ -67,9 +67,9 @@ def handle_authorizations(self, orderr, best_effort=False):
         config = zope.component.getUtility(interfaces.IConfig)
         notify = zope.component.getUtility(interfaces.IDisplay).notification
 
-        with error_handler.ExitHandler(self._cleanup_challenges, aauthzrs):
-            # While there are still challenges remaining...
-            while self._has_challenges(aauthzrs):
+        # While there are still challenges remaining...
+        while self._has_challenges(aauthzrs):
+            with error_handler.ExitHandler(self._cleanup_challenges, aauthzrs):
```
sydneyli: Cleaned up remaining `_cleanup_challenges` call-- good catch! As a result, had to swap the ordering here (so that each successful auth is cleaned up individually).

bmw: Nice catch!

Issue #5602, opened 2018-02-21
Apache HTTP-01 directory owner issues


bmw:
See https://community.letsencrypt.org/t/create-certificate-error-403-since-january/53489/10.

cc @joohoi

bmw:
My understanding is the only way to hit this issue is if the user manually modified the directory permissions, however, we could do a better job handling errors here.

bmw:
Closing in favor of https://github.com/certbot/certbot/issues/6561.

PR #5785, opened 2018-03-23
Verify permissions of conf root in Apache

eicksl:
* Add LocalPermissionsError to errors.py. This may also be used when addressing issues such as #5602.
* Verify permissions of root configuration directory in Apache configurator.
* Edit configurator_test.py to cover changes.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -224,7 +224,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         else:
             self.vhostroot = os.path.abspath(self.conf("vhost-root"))
 
-	self._check_permissions(self.vhostroot)
+        self._check_permissions(self.vhostroot)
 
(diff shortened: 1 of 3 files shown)


sydneyli:
@joohoi Take a look at this if you get a chance?

joohoi:
Thanks for the PR, few early notes before a more detailed review:

- The check should probably be done elsewhere in the code, as Certbot has fallback behavior for cases where the defined `vhost-root` does not exist at all, so it would be beneficial to conduct the check on the effective path. This check could also be used to actually *result in* the fallback, if the defined `vhost-root` exists, but doesn't have permissions. The final path is determined in https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/configurator.py#L1146-L1169
- The error message should probably be more generic, as there's multiple things that could cause the path to be unwritable in addition to the file permissions, for example file attributes (`man chattr`).

bmw:
@eicksl, are you still interested in working on this?

PR #5786, opened 2018-03-23
Refactor _add_directive into separate functions

ohemorange:
There are multiple things broken in the Nginx codebase that can be fixed by making it less ridiculously easy to write incorrect code. Here is a step towards fixing those things.

### DIFF
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -277,7 +277,7 @@ class NginxParser(object):
         return False
 
     def add_server_directives(self, vhost, directives, insert_at_top=False):
-        """Add or replace directives in the server block identified by vhost.
+        """Add directives to the server block identified by vhost.
 
         This method modifies vhost to be fully consistent with the new directives.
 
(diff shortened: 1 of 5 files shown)


certbot-nginx/certbot_nginx/parser.py
```diff
@@ -276,29 +276,48 @@ def has_ssl_on_directive(self, vhost):
 
         return False
 
-    def add_server_directives(self, vhost, directives, replace, insert_at_top=False):
+    def add_server_directives(self, vhost, directives, insert_at_top=False):
         """Add or replace directives in the server block identified by vhost.
```
bmw: nit: "or replace" should be removed here.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -276,29 +276,48 @@ def has_ssl_on_directive(self, vhost):
...
+            functools.partial(_add_directives, directives, insert_at_top))
+
+    def update_or_add_server_directives(self, vhost, directives, insert_at_top=False):
+        """Add or replace directives in the server block identified by vhost.
+
+        This method modifies vhost to be fully consistent with the new directives.
+
+        ..note :: When a directive with the same name already exists in the
+        config block, the first instance will be replaced. Otherwise, the directive
+        will be appended to the config block as in append_server_directives.
```
bmw: `append_server_directives` doesn't exist.

and

nit: It's not always appended. You could either omit this detail or say something like "prepended or appended depending on the value of `insert_at_top`"

Issue #5624, opened 2018-02-26
certbot_dns_google tests require internet connection


TheNavigat:
`certbot_dns_google` package runs tests that require internet connection. The connections to the APIs should be mocked.

igy254:
Hi, @TheNavigat  if no one is working on this one, I would gladly take it. It seems like a good issue to get started.

PR #5791, opened 2018-03-25
Google DNS: Mock API discovery to run tests without internet connection.

andrewsomething:
In the Debian packaging, we've had to disable the tests for the Google DNS plugin due to https://github.com/certbot/certbot/issues/5624 Build-time tests are run in a clean environment without an internet connection. This PR mocks the call to the Google discovery API.


schoen:
@bmw, would you be up for taking a look at this?

PR #5800, opened 2018-03-28
Update oldest tests to test against 0.22.0 versions

bmw:
@joohoi, you probably have the most context to potentially review this. Anyone else who wants to take a crack at it, see the description in #5655.

PR #5590, opened 2018-02-17
Fix print() and xrange() for Python 3

cclauss:
* __print()__ is a function in modern Python
* __from six.moves import xrange__ because __xrange()__ was removed in Python 3

$ __python3 -m flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./tools/simple_http_server.py:17:44: E999 SyntaxError: invalid syntax
    print 'Serving HTTP on {0} port {1} ...'.format(*server.server_address)
                                           ^
./letsencrypt-auto-source/tests/auto_test.py:84:17: F821 undefined name 'xrange'
    for port in xrange(4443, 4543):
                ^
./certbot-compatibility-test/certbot_compatibility_test/validator.py:69:40: F821 undefined name 'xrange'
        return response.status_code in xrange(300, 309)
                                       ^
./certbot-compatibility-test/certbot_compatibility_test/test_driver.py:59:14: F821 undefined name 'xrange'
    for i in xrange(len(responses)):
             ^
./certbot-compatibility-test/nginx/roundtrip.py:11:15: E999 SyntaxError: invalid syntax
        print t
              ^
2     E999 SyntaxError: invalid syntax
3     F821 undefined name 'xrange'
5
```

PR #5806, opened 2018-03-29
Add before_install to all tests running on test-everything

### LINKED ISSUE #5634: Make before_install code run in all test-everything branches
https://github.com/certbot/certbot/pull/5625#issuecomment-369086346

ohemorange:
Fixes #5634.

This PR now successfully gets `before_install` running in `test-everything`. Unfortunately, several of these tests that are now running are failing.

We have:
- `py27-nginx-oldest BOULDER_INTEGRATION=v1` and `v2` failing in `test-everything` before this PR
- `le_auto_precise`, `le_auto_wheezy`, and `le_auto_centos6` failing to import `six`
- `docker_dev` failing for a different reason

Questions that I assume other people know the answers to that I'm going to ask in the morning before investigating:
- why is `test-everything` currently failing?
- why are some of the `le_auto`s failing to import `six`? where do we specify that dependency, why is it passing in `trusty`? why isn't it pulling in the dependency from wherever dependencies normally come from?
- what's up with `docker_dev`? what even is `docker_dev`? here is the error:
```
docker_dev runtests: commands[0] | docker-compose run --rm --service-ports development bash -c tox -e lint
WARNING:test command found but not installed in testenv
  cmd: /usr/local/bin/docker-compose
  env: /home/travis/build/certbot/certbot/.tox/docker_dev
Maybe you forgot to specify a dependency? See also the whitelist_externals envconfig setting.
ERROR: Cannot start service development: driver failed programming external connectivity on endpoint certbot_development_run_2 (3e43285974a2c4d34e994c035014dc0d0527c3edab2d2e1e9f8ecae1394adbe4): Error starting userland proxy: listen tcp 0.0.0.0:80: bind: address already in use
ERROR: InvocationError: '/usr/local/bin/docker-compose run --rm --service-ports development bash -c tox -e lint'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,113 +21,89 @@ matrix:
   include:
     - python: "2.7"
       env: TOXENV=cover FYI="this also tests py27"
-      before_install:
     - python: "2.7"
       env: TOXENV=lint
-      before_install:
     - python: "2.7"
       env: TOXENV=py27-nginx-oldest BOULDER_INTEGRATION=v1
       sudo: required
       services: docker
-      before_install:
     - python: "2.7"
       env: TOXENV=py27-nginx-oldest BOULDER_INTEGRATION=v2
       sudo: required
       services: docker
-      before_install:
     - python: "2.7"
       env: TOXENV=py27_install BOULDER_INTEGRATION=v2
       sudo: required
       services: docker
-      before_install:
     - python: "2.7"
       env: TOXENV=py27_install BOULDER_INTEGRATION=v1
       sudo: required
       services: docker
-      before_install:
     - python: "2.7"
       env: TOXENV='py27-{acme,apache,certbot,dns}-oldest'
-      before_install:
     - sudo: required
       env: TOXENV=apache_compat
       services: docker
-      before_install:
       addons:
     - sudo: required
       env: TOXENV=nginx_compat
       services: docker
-      before_install:
       addons:
     - sudo: required
       env: TOXENV=le_auto_precise
       services: docker
-      before_install:
       addons:
     - sudo: required
       env: TOXENV=le_auto_trusty
       services: docker
-      before_install:
       addons:
     - sudo: required
       env: TOXENV=le_auto_wheezy
       services: docker
-      before_install:
       addons:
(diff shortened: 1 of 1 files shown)


bmw:
> why is test-everything currently failing?

That started yesterday afternoon. Failure is due to #5800 which have the "oldest" tests for Certbot and Nginx test against 0.22.0 for their dependencies now that that's been released rather than using the version in master. The problem is the integration tests are relying on functionality added after 0.22.0 (specifically that `--allow-subset-of-names` works). I was planning on looking into it this afternoon (although I'd happily let someone else take it if they like to volunteer). We/I can revert my PR that caused the problem if you'd like in the meantime.

> why are some of the le_autos failing to import six? where do we specify that dependency?

Looks like that dependency and import was added in #5590. The `test-everything` branch doesn't have this change so this PR fixing the tests so it merges in `master` is running the test with the `six` dependency for the first time.

>  why is it passing in trusty?

[This command](https://travis-ci.org/certbot/certbot/jobs/359666975#L819) is pulling it in as a dependency of the other packages we install.

> why isn't it pulling in the dependency from wherever dependencies normally come from?

The letsencrypt-auto Dockerfile tests don't do the normal Certbot installation. This is at least partially because we're testing that `certbot-auto` can do this properly. Dependencies are installed explicitly in the relevant Dockerfiles.

> what's up with docker_dev?

I expect that removing the clearing of `addons` has caused Apache/Nginx to be installed and started by `apt` and now port 80 is not available.

> what even is docker_dev?

`Dockerfile-dev` is our Dockerfile for developers documented [here](https://certbot.eff.org/docs/contributing.html#running-the-client-with-docker). This test was added during the SF Python sprint because our [developer instructions](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client) tell macOS users to use this Dockerfile but we somewhat embarrassingly found out it hadn't been maintained and didn't work at all! This test builds the Dockerfile and runs a simple test in it to try and prevent this from happening in the future.

ohemorange:
All the problematic tests are now passing (others are still running).

Issue #5810, opened 2018-03-29
Fix certbot and nginx "oldest" tests


bmw:
#5800 caused [failures](https://travis-ci.org/certbot/certbot/jobs/359981505) in the `test-everything` branch.

The purpose of that PR was to update our "oldest" tests whose purpose is to test components with the oldest versions of their dependencies they support. That PR caused the `certbot` "oldest" tests to use `acme==0.22.0` and the `certbot-nginx` oldest tests to use `acme==0.22.0` and `certbot=0.22.0`. Currently these tests are still using the versions from `master` (which was needed before we released 0.22.0).

While these components should still work with the older versions, the reason for these failures is the test-everything branch is trying to run the latest boulder integration test (found in `tests/boulder-integration.sh`) after installing `certbot-nginx` as defined for the `certbot-nginx` oldest tests (see `.travis.yml` on the `test-everything` branch). This integration test script currently tests both Certbot and Certbot with the Nginx plugin, but changes have landed in that script that test functionality in Certbot that was added after 0.22.0.

There are a few ways to fix this, but what I would like to see happen is to split the oldest integration tests for `certbot` and `certbot-nginx`. I'd like the normal integration tests to still test both packages. To do this, I'd do something like:

* Rename `tests/boulder-integration.sh` to something like `_boulder-integration.sh` or `certbot-boulder-integration.sh`.
* At the bottom of the script, remove the call to `certbot-nginx/tests/boulder-integration.sh` and update the `coverage` percentage so it passes without testing nginx.
* Add a new `boulder-integration.sh` script with logic like the following bashy pseudo-code:
```
if [ "$TOXENV" != "py27-nginx-oldest" ];
   tests/_boulder-integration.sh
fi
if [ "$TOXENV" != "py27-certbot-oldest" ];
    certbot-nginx/tests/boulder-integration.sh
fi
```
* Add two new jobs to `test-everything`'s `.travis.yml` for the environments:

    * `TOXENV=py27-certbot-oldest BOULDER_INTEGRATION=v1`
    * `TOXENV=py27-certbot-oldest BOULDER_INTEGRATION=v2`

* Rename v1 and v2 to acmev1 and acmev2 to make it more clear for people looking at tests

The changes to `test-everything`'s `.travis.yml` need to be made in a separate PR against the `test-everything` branch once all of the other changes have landed.

After all of this, #5800 should be able to be unreverted and should be done before the next release and this issue is closed.

bmw:
[...]

PR #5814, opened 2018-03-31
Separate integration

bmw:
Main piece of #5810. You should read that issue before reviewing this PR.

The actual number of lines changed here is relatively small. I renamed `boulder-intergration.sh` to `certbot-boulder-intergration.sh` and wrote a small `boulder-integration.sh` script in its place.


bmw:
I'm certainly open to alternatives here, but I'm not sure I explained the problem very well because I think we have different understandings of it. Let me know what is the best way for me to discuss things with you so we can get on the same page and come up with a plan, but to address things in your post as a starting point:

> The problem is that the version in the py27-certbot-oldest test is too old for the nginx boulder integration test.

`py27-certbot-oldest` installs the latest version of Certbot and the oldest versions of its dependencies. The Nginx plugin is not installed. Currently if we tried to run a travis environment like `TOXENV=py27-certbot-oldest BOULDER_INTEGRATION=...`, it would fail because it will try to run the Nginx tests without the Nginx plugin.

> We should just run that in the py27-nginx-oldest test then; why run the nginx test with every other plugin?

I think we should continue to run integration tests with the latest versions of all of our packages as well. This way this is done in `master` is by simply running `boulder-integration.sh` which runs all integration tests. This gives people a simple command to run and significantly helps test efficiency because the same Certbot and Boulder installation can be reused for the tests. Without this, we either have to add 1 or 2 more Travis environments to be run against PRs or move some of the integration tests to `test-everything` only.

> Plus, why does the current code not run the certbot-boulder-integration.sh test in py27-nginx-oldest?

This is what was causing the failure. `py27-nginx-oldest` installs the latest version of the Nginx plugin with the oldest version of its dependencies. This caused an older version of Certbot to be installed but the latest version of Certbot's integration tests to be run which would fail because it was trying to test features that don't exist in the older Certbot.

What if instead of relying on the `TOXENV` to determine the tests, I added another environment variable that lists the name of the packages whose integration tests should be run? If the environment variable is not set, it runs all tests.

ohemorange:
my response to this is in mattermost

ohemorange:
My response to this continues to be somewhere in the past of mattermost.

bmw:
Yeah. My response to that continues to be somewhere in the future that I can never quite get to. I'll respond to the comments you put in mattermost when I pick this back up though.

bmw:
On Mattermost, Erica said:

> A
py27-certbot-oldest
current certbot version
current certbot integration test
no nginx integration test, because nginx is not a dependency so it won't be installed and also that doesn't make sense
oldest version of all of certbot dependencies
this is perfectly fine. we don't mind that we're not running the nginx integration test because why would we.

To be a little pedantic to try and make sure we're on the same page here, we currently always try to run the Nginx integration tests when running the Certbot integration tests in Travis. See: https://github.com/certbot/certbot/blob/cc24b4e40af5841c8dfddeecd9bde7d1acce62e8/tests/boulder-integration.sh#L456 This is something I want to change with this PR though and with the change, your understanding here is correct.

> B
py27-nginx-oldest
current certbot-nginx version
current certbot integration test
oldest version of all certbot-nginx dependencies, including an older certbot version
this is a problem because current certbot integration test fails with older certbot version
C
py27-nginx-oldest
current certbot-nginx version
current nginx integration test
oldest version of all certbot-nginx dependencies, including an older certbot version
this is perfectly fine, except we're sad because we don't have current certbot integration test running. or are we not saddened by this? and if not, why not?

I'm not saddened by us not running the latest Certbot integration tests with an older version of Certbot. I think trying to get that to work is an unnecessary headache with little to no benefit.

I would be sad if we didn't setup tests like you described in A though after the Certbot and Nginx tests are split. I put my plan to do this in #5810, but this PR has to land first before we can do that.

> D (theoretical)
py27-everything-else-oldest
current certbot-everything-else version
current certbot integration test
current nginx integration test
oldest version of all certbot-nginx dependencies, including an older certbot version
this is perfectly fine. nice to run nginx integration test because plugin interactions, even if it's a bit overkill.

I think doing this with the oldest tests is hard as plugins depend on different oldest versions and we can easily get conflicting requirements. We do run the nginx integration tests with other plugins installed when we're not running any "oldest" tests and are instead using the latest version of everything. I personally think this is enough. Do you disagree?

> here's my question: is it useful to run the certbot integration test with anything that's not the py27-certbot-oldest environment? is there a point to that?

We also want to run them outside of any oldest tests, but outside of that, I don't think so.

> and, is there more of a point to running the nginx integration test with other package oldests than I realize?

No. 

> and given both of those, how can we solve this problem in this complex environment that does not increase how complex it is, and ideally decreases it 

I think we're inherently increasing complexity here. Until now, we had one set of integration tests we always ran that included tests for both certbot and certbot-nginx. We're trying to split them out and I/we would like a way to run:

* Just the Certbot integration tests
* Just the Nginx integration tests
* Both the Certbot and Nginx integration tests

This adds branches and I think complexity. I of course agree we should aim to reduce the amount of complexity we're adding here though.

> for that last bit, I think the confusing part is that the py27-star-oldest bit determines the dependencies, and it shouldn't also determine which tests are run. I think conflating those is the problem.
a problem, anyway

To be a little pedantic again, the `py27-*-oldest` bit does determine what unit tests are run. This is the conventional purpose of setting the `TOXENV` environment variable. You're correct though that this PR makes them also control which integration tests are run.

> I think that specifying boulder environment should not secretly under the hood determine which tests to actually run
I think that everything that's done should be done clearly
so, perhaps, splitting into BOULDER_INTEGRATION and NGINX_BOULDER_INTEGRATION
because those are now both additional tests that we might want to run

We're currently using the value of `BOULDER_INTEGRATION` to determine whether ACMEv1 or ACMEv2 is used. This affects boulder setup and some of our Certbot integration tests are hidden behind a check that `BOULDER_INTEGRATION` is set to ACMEv2.

I also would like to avoid relying on a lot of environment variables which makes it harder to see what tests are running in Travis' UI. We currently only have 2 integration tests, but we will hopefully have integration tests for Apache and the DNS plugins in the near future.

If you don't like `TOXENV` controlling what integration tests are run, what if we had three environment variables:

* `TOXENV` controls which environment is used by `tox`.
* `ACME_VERSION` which is set to 1 or 2.
* `INTEGRATION_TEST` which is a comma/space delimited list of the packages to run integration tests on. We could also have a value like "all" for convenience and readability.

I don't want to require people set all of these variables when running integration tests locally so I'd want to set some defaults here, but this gives each variable a single purpose.

bmw:
@sydneyli, a review on this would be useful. Erica told me she's OK with you taking it if you're interested.

There's a lot to read here (but it will teach you more about our test setup). I'd start with the issue this starts addressing at https://github.com/certbot/certbot/issues/5810. I tried to give enough background there to explain my plan here. The conversation on the PR here is me responding to Erica's questions and concerns.

bmw:
I'd like to see this merged in the next week or so. @sydneyli, do you have time to look at this or should I ask Erica to take a look again?

sydneyli:
@bmw Thanks for reminding me on this. I'll take a look this week.

sydneyli:
Sorry for the lateness on this. I'm jumping in the middle of a longer thread here, so just wanted to clarify. @bmw Do you want a review on the above, given your explicit design proposition in #5810, or my opinion on the proposition of introducing a separate environment variable to specify which integration test to run? I'm unclear on whether this discussion was resolved.

sydneyli:
Just to leave my opinion here in case this wasn't resolved:
> `INTEGRATION_TEST` which is a comma/space delimited list of the packages to run integration tests on. We could also have a value like "all" for convenience and readability.

I like using a separate environment variable telling us which integration test file to run. I don't think we need this environment variable to be on the granularity of "what subset of all of the integration tests should I run" because it seems that we only need to be able to run:
  - Just one of the existing integration test suites (in case we're on one of `py-*-oldest` and want to run the appropriate test suite)
  - All of the integration test suites (in case we're on `py27_install`. or we're running the int. tests locally and this is a sensible default)

And this holds in the case that we've added other integration tests. Is this correct? In that case, can we have `tests/boulder-integration.sh` just run all the integration tests in succession?

bmw:
> In that case, can we have tests/boulder-integration.sh just run all the integration tests (currently, only includes tests/certbot-boulder-integration.sh and certbot-nginx/tests/boulder-integration.sh) in succession?

Yes we can and I think we should!

I updated the PR adding an `INTEGRATION_TEST` environment variable. You can see this variable successfully being used to control which tests run at https://travis-ci.org/certbot/certbot/builds/416944671.

I made the decision that if the variable isn't set (or is set to a value we don't recognize), all tests are run. I think this is nice for things like running integration tests locally. I also added `INTEGRATION_TEST=all` to `.travis.yml` to make what Travis is doing more explicit. If you don't like any of these decisions, let me know.

Finally, I had to reduce the minimum Certbot coverage by a percentage point because code coverage is reduced slightly when the Nginx plugin isn't installed.

bmw:
New commit resolves merge conflicts with #6306.

bmw:
> Do you want a review on this PR, given your explicit design proposition in #5810, or my opinion on the proposition of introducing a separate environment variable to specify which integration test to run? I'm unclear on whether this discussion was resolved.

@sydneyli, I'd like your review on this PR and it should be ready for another review. Erica never got a chance to respond to my latest messages to her in the discussion here and she told me out of band she was fine with you reviewing this PR.

bmw:
Since we usually squash PRs which changes their commits, we have to do that first to avoid merge conflicts.

Usually we have GH squash the PR into a single commit on `master`. If we go that route, I can quickly make the `test-everything` PR after this is merged which should be easy to review.

Alternatively, I can squash the commits here myself in this or in another PR, we have GitHub merge the squashed PR (which creates a merge commit but we'll live), and build the `test-everything` PR off that.

Do you have a preference?

sydneyli:
> Usually we have GH squash the PR into a single commit on `master`. If we go that route, I can quickly make the `test-everything` PR after this is merged which should be easy to review.

Yes, this is fine. Feel free to squash & merge this when you're ready, and I'll review the `test-everything` PR asap afterwards.

Issue #5543, opened 2018-02-06
Unquote parameters in Nginx


ohemorange:
We should support cases like this: https://community.letsencrypt.org/t/nginx-installer-misconfigurationerror-cannot-find-a-virtualhost-matching-domain/

Issue #5774, opened 2018-03-20
Support multiline concatenated quotes in Nginx


ohemorange:
https://community.letsencrypt.org/t/certbot-on-debian-9-with-nginx-could-not-automatically-find-a-matching-server-block/56974/

Issue #5185, opened 2017-10-13
Put new Nginx server blocks we create in a new file when appropriate.


ohemorange:
When copying from `sites-available/default`, put it in a new file under `sites-enabled` and symlink to that from `sites-available`, if that's the setup that we have and `sites-available/*` is imported in `nginx.conf`. Plus any other details that aren't immediately obvious.

Originally mentioned at https://github.com/certbot/certbot/pull/5153/files#r143084200

Issue #5157, opened 2017-10-02
add_server_directives should pick up whitespace from context rather than having the caller set it, and definitely for replace it should just replace the contents.


ohemorange:

lwm:
Right, I started to get setup and lost into the source. I want to write a test first, I guess that this would go in `certbot-nginx/certbot_nginx/tests/parser_test.py` and might look like this:

``` python
    def test_add_server_directives_indent_from_context(self):
        server_block = nginxparser.loads("""
            server {
                listen 80;
            }""")
        # TODO: add some directive
        # TODO: test that we got the same indentation

        server_block = nginxparser.loads("""
            server {
                    listen 80;
            }""")
        # TODO: add some directive
        # TODO: test that we got the same indentation

        server_block = nginxparser.loads("""
            server {
                listen 80;
                    index star.html;
            }""")
        # TODO: add some directive
        # TODO: test that we got the same indentation
        # TODO: Different indentations!?
```

Is this the right approach? I just want to write a fresh configuration and check that what I put in gets the right indentation.

One question I thought about - what happens when there are various levels of indentation (see third example in the test). What should be done about that?

ohemorange:
Yes, this is definitely the right track!

I could see the tests going either there, in `certbot-nginx/certbot_nginx/tests/nginxparser_test.py`, or in both, depending. `parser.py` has our directive-and-higher level code, whereas `nginxparser.py` is more about parsing into lists and such.

For different indentations, you can make that decision as the implementer -- match the previous line, match the next line, and match whatever seems to be the dominant indentation in that block all seem defensible to me.

Some additional test ideas:
- What happens when there's indentation in the server directive being added?
- What happens when the previous or next line is just a newline?
- What happens when the new directive is just whitespace?

Thanks for your work on this!

bmw:
It looks like we're making progress on this in issues like #5823.

ohemorange:
while not necessarily necessary to have this as a standalone issue, I think it's nice to keep open as a reminder as part of a larger nginx parser rehaul.

martindale:
Any progress on this?

nocturn9x:
It's useless to close issues as duplicates and link to this one if the problem wasn't fixed :)

PR #5834, opened 2018-04-06
Nginx parsing rehaul

### LINKED ISSUE #5823: Revamp Nginx parsing

sydneyli:
Fixes #5823 (eventually)

OOP-y revamp of nginx parsing utilities, so we can hopefully stop depending implicitly on structures of nested lists.

Code is all in new files but will eventually cannibalize `parser.py` entirely.

Still very WIP. There are very few comments, many typos, error messages are largely uninformative, I need to break things out into different files, and some of these things i'll want to break out into `plugins.common`.

Next courses of action:
 - Hook into `configurator.py` as primary parser, and fix any test issues
 - Run short experiment to see how much work implementing a dovecot parser would be in this world
    - I suspect we'll need a way to add custom parsing hooks.
 - Test coverage for new object code
 - General cleanup (comments, error handling)


sydneyli:
After messing around with things, this object-tree approach seemed to be the nicest way to do things, but I'm open to other suggetsions.

The new nginx parser actually constructs an object tree.
Assumptions about parsed list structure should be mostly hidden away in `parser_obj.py`, although certain decisions do leak to objects which consume the parsing objects. For instance, since the parsed objects take care of adding whitespace to a directive (such that it matches existing context), it expects an unspaced list to be passed into its functions.

In essence, the tree looks like:

```
                 root_file
                     |
                 Statements
                list of ...
                 /      \
ServerBloc---Bloc       Sentence---Include
              /                       \
      [name, contents]                file (Statements)
         /        \
    Sentence    Statements
```

The basic building block for all of this is the `Sentence`, which is simply a collection of tokens and whitespace. It acts a lot like the `UnspacedList` object from the previous parsing design-- `Sentence.words` magically hides the associated whitespace and quotation marks.

For future block/statement-like parsing (cough Dovecot cough), the basic Statements/Block/Sentence structures will stay the same. In order to add configurability, you'll be able to introduce parsing hooks (i.e. if hook returns True => parse as some type) in order to parse and define configuration-specific objects (like ServerBlocs in nginx).

`Statements` is a collection of Blocs and Sentences (blocks and directives). With a little bit of type introspection, recursive tree-searching operations become pretty easy to build and query. The API for this has to be fleshed out a little bit more, but the gist is that you can query for either Blocks or Sentences that pass a given filter, either shallowly or recursively.

`Include` directives actually load in and parse the requested file. In order to avoid re-parsing files that may have been included twice, we use the global map in `Context`.

## parse and get_data

`parse` parses these objects from the raw parsed lists, from `nginxparser`. `get_data` does the reverse, and recursively dumps the class into the original parsed raw-lists. Functions that modify these objects should modify the underlying raw-list structures as well, so that they play nicely with existing code and dump correctly into files.

## Context

Each object also tracks a `Context`, which contains contextual information like
 - the object's parent, used when duplicate blocks to automatically add that block as a sibling
 - the file where the object is from
 - the current working directory, used to discover files at relative paths that are included
 - a global map of other parsed files, used to prevent object duplication when following `Include`s
 - tabbing information

## Nginx-specific parsing

Most of the nginx-specific logic should occur in `ServerBloc` (TODO: I haven't finished making this abstraction super clean, but it should be), an Nginx-specific types of directives that necessarily change the way that particular statement is parsed. 
 - `ServerBloc` parsing and changes need to track a parallel `VirtualHost` object. Eventually, we might be able to consolidate the `VirtualHost` and `ServerBloc` objects, rather than keeping the two constantly in parallel.

## How does this solve our other problems?
Besides removing explicit parsed-list indexing from most of the nginx codebase, this also:

Put new Nginx server blocks we create in a new file when appropriate: #5185
 - Change the `context.filename` of the appropriate block before dumping.

Unquote parameters in Nginx: #5543
Support multiline concatenated quotes in Nginx: #5774
 - These can be done automatically with the `Sentence` abstraction.

nginx _add_directive should figure out spacing automatically: #3417
add_server_directives should pick up whitespace from context rather than having the caller set it, and definitely for replace it should just replace the contents: #5157
 - Keeping track of tabbing guesses in objects solves both of these.

## Notes

Some of the abstractions are not perfectly fleshed out/mumbled in the current code, but this can be ironed out.

With some work, I've gotten the existing code to play nicely with `configurator` and `obj`, although it involved more changes to those files than I would have liked (mostly involved removing all explicit parsed-list indexing). **Before continuing to massage the rest of the nginx codebase to play nicely with the new parser, I'm looking for comments/iteration on the high-level design.** The nginx code feels a lot more intuitive without the explicit list-indexing; however, this does touch *a lot* of previous assumptions throughout the codebase and I'm currently trying to make these changes less intrusive (so we don't rehaul everything at once).

If you have any questions or maybe thoughts on smoothing over the transition to the new parser, things that seem not-so-nice, ideas on how to improve the existing high-level design, additional edge-cases that need to be worked into the design, please let me know.

ohemorange:
Here are some thoughts and questions:

1. do changes need to be made to nginxparser?

2. "Functions that modify these objects should modify the underlying raw-list structures as well, so that they play nicely with existing code and dump correctly into files."
as in, in parser or in configurator?

3. "Most of the nginx-specific logic should occur in ServerBloc (TODO: I haven't finished making this abstraction super clean, but it should be), an Nginx-specific types of directives that necessarily change the way that particular statement is parsed."
what does this mean?

4. so what does add_server_directives look like? which nodes need to be updated, and does that propagation happen manually or automatically?

5. "Unquote parameters in Nginx: #5543
Support multiline concatenated quotes in Nginx: #5774
These can be done automatically with the Sentence abstraction."
The easiest way to fix #5543 is changing the options in nginxparser.py. Are we instead going to leave those, so we can better write things back out?

sydneyli:
> do changes need to be made to nginxparser?

Nope!

> "Functions that modify these objects should modify the underlying raw-list structures as well, so that they play nicely with existing code and dump correctly into files."
> as in, in parser or in configurator?

I mean the parser. Might re-word this to avoid confusion (this confused me upon rereading, also)

> "Most of the nginx-specific logic should occur in ServerBloc (TODO: I haven't finished making this abstraction super clean, but it should be), an Nginx-specific types of directives that necessarily change the way that particular statement is parsed."
> what does this mean?

ServerBloc is an nginx-specific parsed object type, so functions that understand what Nginx directives are (beyond just a collection of tokens) should go in ServerBloc.

> so what does add_server_directives look like? which nodes need to be updated, and does that propagation happen manually or automatically?

`add_server_directives` on a vhost first updates the corresponding `ServerBloc` object. The `ServerBloc` handles parsing these directives and adding the resulting objects to its contents, and then updating the corresponding vhost.

> The easiest way to fix #5543 is changing the options in nginxparser.py. Are we instead going to leave 
 those, so we can better write things back out?

Since we still want to preserve the original configuration formatting, we should handle this kind of syntax at a level above `nginxparser`. The lowest-level `Sentence` object handles these nicely, by magically hiding spaces and quotations (similar to how `UnspacedList` does)

ohemorange:
Ok actually I'm going to request that we break this PR into small pieces. Perhaps committing the building blocks first, then changing around higher-level pieces to use them, like we did with ACMEv2. Otherwise I'm going to keep trying to find a large-hour chunk of time to review this and continue to never manage to.

sydneyli:
Splitting up.

PR #5836, opened 2018-04-08
Yandex DNS Authenticator

wrnx:
Implement an Authenticator which can fulfill a dns-01 challenge using the Yandex v2 API. Most of the code was copied from the DNSimple DNS Authenticator.


schoen:
@zjs, would you be available to review this by any chance?

certbot-dns-yandex/certbot_dns_yandex/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+                                          (Default: 30)
+========================================  =====================================
+
+
+Credentials
+-----------
+
+Use of this plugin requires a configuration file containing Yandex API
+credentials, obtained from your Yandex
+`account page <https://tech.yandex.com/domain/doc/concepts/access-docpage/#access-admin>`_.
```
zjs: This page also talks about several different types of authentication. Because Yandex supports these, I think it'd be helpful to clearly document which method this plugin uses.

Usually we link the the specific page of the DNS provider's dashboard where the user can generate a token/key, but Yandex seems to have a more complicated process than most.

Perhaps we need to adjust the wording here to say something more like:
> Use of this plugin requires a configuration file containing Yandex API credentials. Follow the instructions in the [Yandex API documentation](https://tech.yandex.com/domain/doc/concepts/access-docpage/#access-admin) to obtain a PPD token.

certbot-dns-yandex/certbot_dns_yandex/dns_yandex.py
```diff
@@ -0,0 +1,82 @@
...
+            'auth_token': token,
+            'ttl': ttl,
+        })
+
+    def _handle_general_error(self, e, domain_name):
+        if str(e) == '':
+            return
+
+        hint = None
+        if str(e).startswith('No domain found'):
```
zjs: I don't have a Yandex account to test this, but based on inspection of the Lexicon provider, I don't _think_ you want this here.

If you have a domain registered with Yandex (e.g., example.com) and you try to issue a certificate for a nested sub-domain (e.g., foo.bar.baz.example.com), what happens?

(Usually, the `_find_domain_id` method which calls this usually relies on "No domain found" to realize that it needs to check bar.baz.example.com, baz.example.com, and then example.com — which is what the DNS provider recognizes. This check would interrupt that functionality and treat the "No domain found" as an error to report instead of ignoring it while we look for a domain the provider recognizes. It's possible that the Yandex API behaves a bit differently than others here, but the Lexicon provider doesn't convey that.)

certbot-dns-yandex/certbot_dns_yandex/dns_yandex_test.py
```diff
@@ -0,0 +1,49 @@
...
+        self.auth = Authenticator(self.config, "yandex")
+
+        self.mock_client = mock.MagicMock()
+        # _get_yandex_client | pylint: disable=protected-access
+        self.auth._get_yandex_client = mock.MagicMock(return_value=self.mock_client)
+
+
+class YandexLexiconClientTest(unittest.TestCase, dns_test_common_lexicon.BaseLexiconClientTest):
+
+    DOMAIN_NOT_FOUND = Exception('')
```
zjs: What is the error message Yandex returns when it can't find a domain? It would be helpful to include an example of that here to allow the plugin to be realistically unit tested.

tox.cover.sh
```diff
@@ -41,6 +41,8 @@ cover () {
     min=99
   elif [ "$1" = "certbot_dns_route53" ]; then
     min=92
+  elif [ "$1" = "certbot_dns_yandex" ]; then
+    min=92
```
zjs: It seems like we should be able to do better here; most DNS plugins can cover all but 1 line without much difficulty. If we don't end up removing lines 77-81, let's make sure we adjust a test to cover them.

bmw:
This hasn't seen any progress since April so I'm closing the PR. If you continue working on it, please don't hesitate to open a new one!

PR #5837, opened 2018-04-08
Common Name support has been added for acme.crypto_util.make_csr()

ZenSecurity:
With help of this commit we can set "common_name" manually in make_csr(), or just use first domain from "subject_alt_names" list if "common_name" is empty.

### DIFF
--- a/acme/acme/crypto_util.py
+++ b/acme/acme/crypto_util.py
@@ -155,23 +155,27 @@ def probe_sni(name, host, port=443, timeout=300,
             raise errors.Error(error)
     return client_ssl.get_peer_certificate()
 
-def make_csr(private_key_pem, domains, must_staple=False):
-    """Generate a CSR containing a list of domains as subjectAltNames.
+def make_csr(private_key_pem, subject_alt_names, must_staple=False, common_name=None):
+    """Generate a CSR containing a list of subject_alt_names as subjectAltNames
+    and common_name string as Common Name subject. If common_name is not passed,
+    first domain from the subject_alt_names will be a Common Name subject by default.
 
     :param buffer private_key_pem: Private key, in PEM PKCS#8 format.
-    :param list domains: List of DNS names to include in subjectAltNames of CSR.
+    :param list subject_alt_names: List of DNS names to include in subjectAltNames of CSR.
     :param bool must_staple: Whether to include the TLS Feature extension (aka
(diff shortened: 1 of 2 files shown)


schoen:
Hi @ZenSecurity ,

Thanks!

Do you have a particular way that you'd like to see this functionality get used in Certbot, or another client where you would like to use it? I know users sometimes ask us about how to cause a particular name to be displayed in the `CN` X.509 field, so it looks like you're aiming at giving people more convenient control of that. But how do you envision that users would take advantage of this feature?

ZenSecurity:
This feature will be useful for python developers, who use bare acme library without certbot client, like i do. It will help to integrate letsencrypt more easily into complicated infrastructure.

ohemorange:
@schoen, looks like you're on top of reviewing this. Just wanted to point out that different clients sometimes have different behavior around when certs do an do not have a common name set and when SANs are or aren't set, so we might want to consider that when making the change to always set the common name.

ZenSecurity:
Ok, i will be waiting your decision !

ZenSecurity:
Any updates ?

ZenSecurity:
What do you think about backward compatibility (https://www.digicert.com/subject-alternative-name-compatibility.htm), maybe we need more accurate checks, especially for wildcards ?

IdoHaritan:
Any updates?

ohemorange:
Common name is no longer recommended; SAN should be used instead. I'm closing this PR as I do not think we should or will ever implement it. For more info:

https://datatracker.ietf.org/doc/html/rfc2818
https://binfalse.de/2017/05/19/common-name-vs-subject-alt-names/
https://community.letsencrypt.org/t/is-common-name-automatically-included-in-san/214002

IdoHaritan:
I understand the deprecation of the Common Name (CN) in favor of Subject Alternative Names (SAN).
However, we have a specific use case where we rely on a client certificate with a CN to verify user identity. Some of the services we use require this for compatibility and functionality reasons.

If there are alternative solutions that could address this requirement, I would greatly appreciate your guidance. Otherwise, we would still need this feature to meet our operational needs.

Thank you for your consideration.

ohemorange:
@IdoHaritan I would recommend posting on the let's encrypt community forum to see what other users who might be in similar situations have done: https://community.letsencrypt.org/

PR #5844, opened 2018-04-10
Add port option for rfc2136 plugin

connection-reset:
The dns-rfc2136 plugin uses port 53 with no way to change it. This PR adds an optional directive `dns_rfc2136_port` to the credentials.ini file.

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/__init__.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/__init__.py
@@ -21,8 +21,9 @@ Credentials
 -----------
 
 Use of this plugin requires a configuration file containing the target DNS
-server that supports RFC 2136 Dynamic Updates, the name of the TSIG key, the
-TSIG key secret itself and the algorithm used if it's different to HMAC-MD5.
+server and optional port that supports RFC 2136 Dynamic Updates, the name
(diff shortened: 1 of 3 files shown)


bmw:
Travis failures aren't caused by your PR. We're looking into the problem.

bmw:
Test failures have been fixed and I kicked Travis so they pass here as well.

certbot-dns-rfc2136/certbot_dns_rfc2136/__init__.py
```diff
@@ -21,15 +21,18 @@
...
+of the TSIG key, the TSIG key secret itself and the algorithm used if it's
+different to HMAC-MD5.
 
 .. code-block:: ini
    :name: credentials.ini
    :caption: Example credentials file:
 
    # Target DNS server
    dns_rfc2136_server = 192.0.2.1
+   # Target DNS port
```
joohoi: I would probably add a note that this is optional here.

connection-reset: It does say it's optional in the little text above.
dns_rfc2136_algorithm is optional too, do you want a note there too?

joohoi: Oh, you are correct. No need to add it to the comment.

PR #5852, opened 2018-04-11
Use older boulder version for tests

bmw:
Use older boulder version for tests until https://github.com/letsencrypt/boulder/pull/3614#issuecomment-380615172 is resolved.


cpu:
@bmw We reverted the broken commit and you should be able to back this change out again. Thanks for catching that!

PR #5853, opened 2018-04-12
Revert fix-macos-pytest

bmw:
This has been on my TODO list for a while but I'm just now getting around to it.

In #5360 I was trying to fix macOS tests in the test-everything branch but it didn't work. The actual fix came in #5378. Since my PR added a lot of unnecessary complexity, let's revert it. You can see tests on the test-everything branch passing with the changes here at https://travis-ci.org/certbot/certbot/builds/365383143.


bmw:
The 2nd commit is needed because I expanded on the changes in #5360 by having the other "oldest" tox environments reuse the `py27-oldest`'s `passenv` setting in #5655. Now that `py27-oldest` doesn't set `passenv`, we need to remove it from the other environments as well.

bmw:
For more context, I thought the cause of the Travis failures was https://github.com/pytest-dev/pytest-xdist/issues/9 because `pytest` was spinning up hundreds of worker processes but what was actually happening was the Python installation (and therefore `pytest`) was messed up, worker processes kept crashing, and `pytest` would start another one to replace it.

PR #5862, opened 2018-04-12
Update constants.py

mabayhan:
On FreeBSD or MacOS, "certbot --nginx" fails. The reason is, at these op. systems, nginx directory is different than linux.

### DIFF
--- a/certbot-nginx/certbot_nginx/constants.py
+++ b/certbot-nginx/certbot_nginx/constants.py
@@ -8,7 +8,7 @@ else:
     server_root_tmp = "/etc/nginx"
     
 CLI_DEFAULTS = dict(
-    server_root=server_root_tmp
+    server_root=server_root_tmp,
     ctl="nginx",
 )
 """CLI defaults."""
--- a/certbot-nginx/certbot_nginx/constants.py
+++ b/certbot-nginx/certbot_nginx/constants.py
@@ -1,9 +1,14 @@
 """nginx plugin constants."""
 import pkg_resources
+import platform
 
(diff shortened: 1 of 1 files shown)


bmw:
@ohemorange, can you add this to your review queue?

ohemorange:
yup

certbot-nginx/certbot_nginx/constants.py
```diff
@@ -1,9 +1,14 @@
...
 
-
+if(platform.system() == ('FreeBSD' or 'Darwin')):
+    server_root_tmp = "/usr/local/etc/nginx"
+else:
+    server_root_tmp = "/etc/nginx"
+    
 CLI_DEFAULTS = dict(
-    server_root="/etc/nginx",
+    server_root=server_root_tmp
```
ohemorange: tests are failing because it's missing a comma here

mabayhan: I tried your command but it failed, 

certbot: error: unrecognized arguments: --server-root /usr/local/etc/nginx

And sorry for comma, it should be like 

server_root=server_root_tmp,

I tested like this and worked.

ohemorange: You're right, it should have been `--nginx-server-root`.

Great! Could you add another commit to fix up the PR?

mabayhan:
I tried to understand errors but i didnt. I dont have so much knowledge. But I alread tested at FreeBSD system with commit that i made , and it worked.

PR #5873, opened 2018-04-13
Remove unneeded sys import

bmw:
Once upon a time we needed this in some of these `setup.py` files because we were using `sys` in the file, but we aren't anymore so let's remove the import.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -1,5 +1,3 @@
-import sys
-
 from setuptools import setup
 from setuptools import find_packages
 
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -1,4 +1,3 @@
-"""Setup script for acme"""
 from setuptools import setup
 from setuptools import find_packages
 
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -1,3 +1,4 @@
+"""Setup script for acme"""
 from setuptools import setup
 from setuptools import find_packages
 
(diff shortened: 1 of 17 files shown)


ohemorange:
Does lint not check setup.py files for unused imports? Is it easy to fix it so it does?

bmw:
Good idea about configuring lint to check our `setup.py` files.

After doing this, `lint` insisted I added a bunch of docstrings. I also switched a usage of `distutils` to `setuptools` because lint was struggling to understand `distutils`. For context, `distutils` is packaging code in the Python standard library while `setuptools` is a widely adopted 3rd party wrapper around that with a lot of niceties added. Since we were already using and requiring `setuptools` by importing it in the file and we use it everywhere else, let's just use `setuptools` in this file as well.

bmw:
I went ahead and reverted linting these files. It's awkward because a `setup.py` file isn't actually an importable part of a module, it just defines the package. When you run mypy on these files, it errors because there are multiple `setup.py` files it has imported.

I have an unreverted version if we want it for some reason on the `leave-sys-out-of-this2` branch. Getting lint/mypy to work could be nice, but this PR as it is cleans things up a bit and I don't think blocking these small improvements on reworking some of our testing is necessary or a good use of our time right now. I'm happy to have an issue created for us to look into later though.

@sydneyli, if you have time, this should be a simple review.

PR #5889, opened 2018-04-20
Updated base image to python:2-alpine3.7

HadesArchitect:
Updated base image from python:2-alpine to python:2-alpine3.7. Python:2-alpine internally utilises alpine version 3.4 which end-of-life date is the first of May 2018.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2-alpine
+FROM python:2-alpine3.7
 
 ENTRYPOINT [ "certbot" ]
 EXPOSE 80 443


HadesArchitect:
image

joohoi:
Thanks for the PR! 

One would easily assume that `python:2-alpine` would be a metapackage utilizing the current stable version of Alpine, but apparently this isn't the case!

The PR looks good to me, I tested it over and everything is in order. I'm going to leave it to be merged by @bmw as I'm not 100% sure if there is something else to take into account when merging, affecting our CI tests or official container releases.

HadesArchitect:
@joohoi Pleasure for me to support such a great project! I wish I could do more but my python skills aren't good enough.

bmw:
LGTM. Thanks for the PR @HadesArchitect.

PR #5891, opened 2018-04-20
Update good volunteer task to good first issue

bmw:
Before GitHub had "good first issue" as a [default label](https://help.github.com/articles/about-labels/#using-default-labels), we used "Good Volunteer Task" to designate issues good for new contributors. When GitHub added the new label, we removed the "Good Volunteer Task" label, but it looks like the documentation was never updated. This fixes that.


bmw:
Thanks for the review Zach.

Issue #4493, opened 2017-04-12
Rewrite argument parsing


bmw:
We currently use the largely unmaintained `configargparse` and a number of hacks like magic default values and reaching into the internals of `configargparse`/`argparse` to get the behavior we want. We should rewrite our argument parsing to use another library like `click` or `fire`.

cc @erikrose, @ohemorange

erikrose:
I've had excellent experiences with click. See https://github.com/mozilla/dxr/blob/master/dxr/cli/__init__.py for the top-level entrypoint in DXR's click-based CLI. click is particularly good at subcommands.

ohemorange:
I don't know anything about `click`, but my friend wrote `fire` so we could probably get good support. I'm all for cleaner code everywhere!

erikrose:
Then we should probably make a pro/con list and see what pops out!

anarcat:
i don't know if [click](http://click.pocoo.org/) will cover the use cases currently covered by configargparse. first, it doesn't support config files out of the box - that might be done by a third party module like [click_config_file](https://github.com/phha/click_config_file). that config file support doesn't support writing config files the way certbot currently does either.

then click is way more than a commandline parser - it's a whole TUI (text user interface) driver. it hijacks the main control flow of your program and makes testing ... different, for example. it supports progress bars and ANSI art and whatnot. it is *very* opiniated about how things should be and you might find it difficult to work with if you deviate from canon.

[fire](https://github.com/google/python-fire) is much simpler, but also has the limitation of not working with config files.

another option people often favor is [docopt](http://docopt.org/) which also doesn't support config files (let alone writing them) but there is some [sample code to parse config files](https://github.com/docopt/docopt/blob/master/examples/config_file_example.py) that is surprisingly clean.

there aren't that many options here - rewriting the commandline parser is a huge deal in certbot. argparse, while it's weird and complicated, is one of the best things we have right now. i wonder if the best approach might not be to fix the issues in configargparse instead of rewriting half of certbot. ;) and there's been some [action](https://github.com/bw2/ConfigArgParse/issues/133) on the issue of finding a new maintainer on that side, so I wouldn't give up on it already...

bmw:
Thanks for the thoughts @anarcat! For what it's worth, we don't use configargparse for writing configuration files and instead use configobj.

ronhanson:
[...]

PR #5892, opened 2018-04-20
Allow `certbot subcommand [options] --help`

pklapperich:
Looking at other some tools which use a `toolname [subcommand]` structure, it looks like it's common to either support `toolname [subcommand] [options] --help` (ip, btrfs, pacman), support both that method **and also** `toolname --help [subcommand]` (hg, svn, git), or support _neither_ (systemd, aptitude). certbot seems somewhat alone in supporting only the `toolname --help [subcommand] variant`.

**But why??**
The --help stantax of certbot has been a HUGE minor-annoyance for me since the launch of Let's Encrypt. I'll type out a command like:

`certbot renew --foo /long/path --p`

and then realize that I've forgotten the spelling of the option I want. So I quickly key `^hhelp` leaving me with:

`certbot renew --foo /long/path --help`

and I excitedly hit enter only to find it's not at all what I expected or wanted! So I hit the up arrow to recall the previous command, press _several thousand_* keys to manipulate the command so it reads 

`certbot --help renew`

and I now have the needed information, whereafter I can pop up 2 levels of history, delete the word help, and fill in the option I wanted (ex `--preferred-challenges`)

**Too long, didn't care**
This patch brings certbot more in-line with other tools by supporting both options. I haven't updated the documentation yet; I wanted to get feedback first and see if this is something you'd consider including.

*approximate count

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -482,11 +482,12 @@ class HelpfulArgumentParser(object):
             self.args[0] = '--help'
 
         self.determine_verb()
-        help1 = self.prescan_for_flag("-h", self.help_topics)
-        help2 = self.prescan_for_flag("--help", self.help_topics)
+        help1 = self.find_help_topic("-h", self.help_topics)
+        help2 = self.find_help_topic("--help", self.help_topics)
         if isinstance(help1, bool) and isinstance(help2, bool):
             self.help_arg = help1 or help2
         else:
+            #when is this even possible?
             self.help_arg = help1 if isinstance(help1, six.string_types) else help2
 
         short_usage = self._usage_string(plugins, self.help_arg)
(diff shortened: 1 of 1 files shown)


bmw:
Thanks for the PR. I completely agree with you that Certbot's behavior here could be improved.

With that said, I don't really want to add more custom code and features on top of argparse. As you can see, we've done that a lot and it's gotten kind of messy.

Rather than continuing to go down this road, I think a better approach for Certbot in the long term is to switch to another parsing library that has more of the features we want which is tracked by https://github.com/certbot/certbot/issues/4493. That's obviously a major project, but if you're interested in working on that or have suggestions there, let's talk in that issue.

I'm closing this PR, but hopefully some progress is made on the larger issue of rewriting Certbot's CLI parsing and will resolve the issue you describe in the future.

PR #5894, opened 2018-04-23
Implement TLS-ALPN-01 challenge and standalone TLS-ALPN server 

mdebski:
The new challenge is described in https://github.com/rolandshoemaker/acme-tls-alpn.


bmw:
Thanks for the PR @mdebski.

We'll try and fully review this in the next couple weeks, but it looks like some tests are failing when trying to use older versions of our dependencies. We unfortunately have to keep support for some pretty old packages to allow for distro packaging in places like [EPEL](https://fedoraproject.org/wiki/EPEL). In particular, we need to support `pyOpenSSL` as old as 0.13.1. The full list of these versions can be found in `tools/oldest_constraints.txt`.

As for how to accomplish this, we don't plan on using the TLS-ALPN challenge in Certbot anytime soon and I doubt other users of our ACME library are going to either. With this in mind, I think it's OK if you make the implementation of the TLS-ALPN-01 challenge pretty bare bones here. Implementing functionality similar to the other challenge types would be nice, but assuming this is pretty hard with the ancient version of `pyOpenSSL`, I think just adding the basic types so the library can parse the challenge type without error is good enough.

bmw:
Another option would be to put the code that tries to perform/verify the TLS-ALPN challenge behind a check that PyOpenSSL is a new enough version and/or has the features we need. If the check fails, you can raise an exception.

mdebski:
Ok, I'll try to make it work with older pyOpenssl. I prefer the second solution, as the feature that needs it is in fact in the core of tls-alpn challenge - it's for generating the challenge cert with acme extension.

Is cryptography>=1.0 ok?

bmw:
Thanks @mdebski. Bumping the acme requirement to cryptography>=1.0 is fine.

mdebski:
I tried working around this, but unfortunately it failed:
https://travis-ci.org/certbot/certbot/jobs/371503303#L744
for the very same reason I decided to go with cryptography.x509 instead of openssl.x509. I don't think I fully understand why is this failing (and how does it depend on versions of everything), I need to dig deeper into that.

Anyway, how should I implement your proposal to fail alpn on old pyopenssl? Should I just make sure everything that worked before still works, and possibly raise an exception in the new stuff? What about tests? Make alpn tests conditional on version, and simply do not run them on older?

bmw:
> I don't think I fully understand why is this failing (and how does it depend on versions of everything)

I think what's going on is cryptography didn't support making certs with custom extensions until cryptography 1.3. This isn't documented in their changelog, but this seems to be suggested by https://github.com/pyca/cryptography/issues/2747.

I'd like to avoid bumping our cryptography requirement higher than 1.2.3 though. This is the version in Ubuntu Xenial and we've been working to get a new version of Certbot shipped there for a long time now. Getting an update to a crypto library used by many other projects is a much harder sell than just updating Certbot itself.

> Anyway, how should I implement your proposal to fail alpn on old pyopenssl? Should I just make sure everything that worked before still works, and possibly raise an exception in the new stuff?

Pretty much. I haven't dug into it so it might already work this way, but we shouldn't raise an exception if the CA includes a TLS-SNI-01 challenge in the authorization object regardless of the version of our dependencies. Instead, I'd like us to only raise an exception when trying to perform/verify the challenge.

> What about tests? Make alpn tests conditional on version, and simply do not run them on older?

That works. If you're feeling ambitious, you could also add tests that mock out the libraries so the code is still tested when run with the older libraries. I don't think that's totally necessary though.

mdebski:
Hi,

I've managed to work around missing support for custom extensions, and in fact only use pyopenssl for that. For the other problematic issue - alpn support - I just disabled the test if required feature in pyopenssl is missing.

Travis should be all green, I'd say this is fully ready now.

mdebski:
@bmw ^

mdebski:
The single failure (timeout) in travis seems to be flakiness, it worked fine a commit before. Could someone with write access retry it?

bmw:
> The single failure (timeout) in travis seems to be flakiness, it worked fine a commit before. Could someone with write access retry it?

Done. Tests passed this time.

We'll try and give this a full review soon.

bmw:
@sydneyli, this is a larger PR, but if you think you would have time to get to it in the next couple days, taking it would help me out. If not, I can take it.

If you decide to review it, please take care when reviewing the public interfaces added here (which we have to live with for the foreseeable future) and that we handle TLS-ALPN when run with old versions of our dependencies gracefully. You can see the discussion above about the latter.

Yoni116:
@bmw small question - will this implementation replaces the old tls challenge that was removed and allowed to do verification with using only port 443?

mdebski:
tls-alpn is designed to be a replacement for tls-sni. This PR is not supposed to modify any behaviours around tls-sni in certbot though.

As for 443 port - I believe no changes to that, this is still constrained to 443.

acme/acme/challenges.py
```diff
@@ -507,6 +508,147 @@ def validation(self, account_key, **kwargs):
...
+        Verify ``validation`` using ``account_public_key``, optionally
+        probe tls-alpn-01 certificate and check using `verify_cert`.
+
+        :param .challenges.TLSALPN01 chall: Corresponding challenge.
+        :param str domain: Domain name being validated.
+        :param JWK account_public_key:
+        :param OpenSSL.crypto.X509 cert: Optional certificate. If not
+            provided (``None``) certificate will be retrieved using
+            `probe_cert`.
+        :param int port: Port used to probe the certificate.
```
sydneyli: the interface matches TLS-SNI-01's exactly, which is nice :) however, `kwargs` is a bit clunky, and since we really only expect `host` or `port` to be passed in, can we limit the set of arguments from `*` to just  those two? Same goes for `probe_cert`.

acme/acme/crypto_util.py
```diff
@@ -39,9 +39,11 @@ class SSLSocket(object):  # pylint: disable=too-few-public-methods
     :ivar method: See `OpenSSL.SSL.Context` for allowed values.
 
     """
-    def __init__(self, sock, certs, method=_DEFAULT_TLSSNI01_SSL_METHOD):
+    def __init__(self, sock, cert_selection, alpn_selection=None,
```
bmw: I appreciate you setting up the plumbing here so we can only serve the cert when it's requested using ALPN when openssl and pyopenssl support it and while this change is probably OK, we've been avoiding making breaking changes in `acme` since we entered beta in 2015. With this in mind, I'm thinking we should revert the `cert -> cert_selection` changes here for now and we can revisit this if desired when this feature is added to our dependencies.

Does this seem reasonable to you?

acme/acme/crypto_util.py
```diff
@@ -123,6 +129,7 @@ def probe_sni(name, host, port=443, timeout=300,
     :param tuple source_address: Enables multi-path probing (selection
         of source interface). See `socket.creation_connection` for more
         info. Available only in Python 2.7+.
+    :param list bytes alpn_protocols: Protocols to request using ALPN.
```
bmw: Syntax for this should be:
```
:param alpn_protocols: Protocols to request using ALPN.
:type alpn_protocols: `list` of `bytes`
```

acme/acme/crypto_util.py
```diff
@@ -241,12 +250,13 @@ def _pyopenssl_cert_or_req_san(cert_or_req):
...
 
 def gen_ss_cert(key, domains, not_before=None,
-                validity=(7 * 24 * 60 * 60), force_san=True):
+                validity=(7 * 24 * 60 * 60), force_san=True, extensions=None):
     """Generate new self-signed certificate.
 
     :type domains: `list` of `unicode`
     :param OpenSSL.crypto.PKey key:
     :param bool force_san:
+    :param extensions `list` of `OpenSSL.crypto.X509Extension`:
```
bmw: Syntax here should be:
```
:param extensions: <short description if desired>
:type extensions: `list` of `OpenSSL.crypto.X509Extension`
```

acme/acme/crypto_util.py
mdebski: You're perfectly right about breaking changes. I think it's reasonable though to provide this callback feature, so I just refactored it to not break anything. I also changed the order so that the added parameters are the last ones (in case sb specified them as positional).

Is that ok for you?

acme/acme/challenges.py
```diff
@@ -507,6 +508,147 @@ def validation(self, account_key, **kwargs):
...
+
+        :returns: Whether the certificate was successfully verified.
+        :rtype: bool
+
+        """
+        # pylint: disable=protected-access
+        names = crypto_util._pyopenssl_cert_or_req_all_names(cert)
+        logger.debug('Certificate %s. SANs: %s', cert.digest('sha256'), names)
+        if domain not in names:
+            return False
```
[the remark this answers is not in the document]
mdebski: Right, thanks! I also added case folding to mirror boulder impl:
https://github.com/letsencrypt/boulder/pull/3654/files#diff-4c51d1d7ca3ec3022d14b42809af0d7eR719

acme/acme/challenges.py
```diff
@@ -507,6 +508,147 @@ def validation(self, account_key, **kwargs):
...
+            provided (``None``) certificate will be retrieved using
+            `probe_cert`.
+        :param int port: Port used to probe the certificate.
+
+
+        :returns: ``True`` iff client's control of the domain has been
+            verified.
+        :rtype: bool
+
+        """
```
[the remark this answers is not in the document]
mdebski: OpenSSL will terminate connection on failure to terminate connection, so probe_sni will raise an exception.

mdebski:
@bmw , @sydneyli : done

acme/acme/crypto_util.py
bmw: That works for me! It certainly makes it easier for us to do cert selection based on ALPN when it's an option for us.

Thanks for catching the parameter order and moving the new ones to the end.

Only suggested change with this approach is maybe we should make `certs` an optional parameter now with a default value of `None` as it makes the interface a little nicer when you want to use `cert_selection`. If we did this, we should also check that one of `certs` or `cert_selection` is set. You're welcome to ignore this suggestion if you prefer.

acme/acme/challenges.py
```diff
@@ -507,6 +508,148 @@ def validation(self, account_key, **kwargs):
...
+        if port is None:
+            port = self.PORT
+
+        return crypto_util.probe_sni(host=host, port=port, name=domain,
+                alpn_protocols=[self.ACME_TLS_1_PROTOCOL])
+
+    def verify_cert(self, domain, cert):
+        """Verify tls-alpn-01 challenge certificate.
+
+        :param OpensSSL.crypto.X509 cert: Challenge certificate.
```
sydneyli: nit: document `domain`

acme/acme/challenges.py
```diff
@@ -507,6 +508,148 @@ def validation(self, account_key, **kwargs):
...
+        acme_extension = crypto.X509Extension(self.ID_PE_ACME_IDENTIFIER_V1,
+                critical=True, value=der_value)
+
+        return crypto_util.gen_ss_cert(key, [domain], force_san=True,
+                extensions=[acme_extension]), key
+
+    def probe_cert(self, domain, host=None, port=None):
+        """Probe tls-alpn-01 challenge certificate.
+
+        :param unicode domain: domain being validated, required.
```
sydneyli: nit: document `host` and `port`

acme/acme/standalone.py
```diff
@@ -43,7 +43,14 @@ def __init__(self, *args, **kwargs):
 
     def _wrap_sock(self):
         self.socket = crypto_util.SSLSocket(
-            self.socket, certs=self.certs, method=self.method)
+            self.socket, {}, cert_selection=self.cert_selection,
+            alpn_selection=getattr(self, 'alpn_selection', None),
+            method=self.method)
+
+    def cert_selection(self, connection):
+        """Callback selecting certificate for connection."""
```
sydneyli: nit: document `connection`

acme/acme/standalone.py
```diff
@@ -147,6 +154,45 @@ def __init__(self, *args, **kwargs):
...
+        # negotiation is done after cert selection.
+        # Therefore, currently we always return challenge cert, and terminate
+        # handshake in alpn_selection() if ALPN protos are not what we expect.
+        # [0] https://github.com/openssl/openssl/issues/4952
+        server_name = connection.get_servername()
+        logger.debug("Serving challenge cert for server name %s", server_name)
+        return self.challenge_certs.get(server_name, None)
+
+    def alpn_selection(self, _connection, alpn_protos):
+        """Callback to select alpn protocol."""
```
sydneyli: nit: document `alpn_protos`

acme/acme/standalone.py
```diff
@@ -43,7 +43,14 @@ def __init__(self, *args, **kwargs):
...
         self.socket = crypto_util.SSLSocket(
-            self.socket, certs=self.certs, method=self.method)
+            self.socket, {}, cert_selection=self.cert_selection,
+            alpn_selection=getattr(self, 'alpn_selection', None),
+            method=self.method)
+
+    def cert_selection(self, connection):
+        """Callback selecting certificate for connection."""
+        server_name = connection.get_servername()
+        return self.certs.get(server_name, None)
```
sydneyli: Let's prefix `cert_selection` and `alpn_selection` with `_` to indicate they're for internal use only?

mdebski: I don't think they should be for internal use only.

acme/acme/crypto_util.py
mdebski: done

acme/acme/standalone.py
sydneyli: how do you mean? the selection routines can be overriden by subclasses, but what use might someone have for consuming them in the API for `TLSServer`?

mdebski: Oh, right, they definitely should be private. I was reading this on a phone and misread the context.

PR #5901, opened 2018-04-26
Add --reuse-key feature

### LINKED ISSUE #3709: Add full support for using an existing private key

### LINKED ISSUE #3788: please let me reuse a key during cert renewal

schoen:
Intended to supersede #4610 (with a different syntax per discussion with @ohemorange ).

Fixes #3788; fixes #3709.


jschlyter:
Not tested, but feature-wise I agree this is a better solution than #4610.

schoen:
This appears to be working well now and I tested it quite a bit.

I had a red herring because when I renewed with `certbot certonly --force-renewal` (without explicit `--reuse-key`) I lost the `--reuse-key` setting from the renewal configuration. By contrast, `certbot renew` worked as expected. This is a behavior that @bmw has discussed a number of times and that I've even sometimes helped people out on the forum with.

Anyway, issuing with `--reuse-key` and then renewing with `certbot renew` currently works as expected in terms of using the same private key for the renewal.

The only remaining problem that I know about is that I don't have test coverage for all four combinations of `--dry-run --reuse-key`, `--dry-run --no-reuse-key`, `--reuse-key`, and `--no-reuse-key` (I think effectively three of them are covered by the current unit tests here). I tried all of them manually and they did the right thing, but maybe they should all be covered explicitly in unit tests.

Fixes #3788 and #3709.

schoen:
> I tried all of them manually and they did the right thing, but maybe they should all be covered explicitly in unit tests.

Per discussion with @ohemorange, I'm going to add one more unit test related to this.

schoen:
This is ready for review!

bmw:
@ohemorange, can you add this to your review queue?

ohemorange:
already on it!

innovia:
awesome work thank you - i'm using this branch

certbot/client.py
```diff
@@ -266,7 +266,7 @@ def obtain_certificate_from_csr(self, csr, orderr=None):
         cert, chain = crypto_util.cert_and_chain_from_fullchain(orderr.fullchain_pem)
         return cert.encode(), chain.encode()
 
-    def obtain_certificate(self, domains):
+    def obtain_certificate(self, domains, oldkeypath=None):
```
ohemorange: `old_key_path` or `old_keypath` for pythonicness?

certbot/client.py
```diff
@@ -279,16 +279,36 @@ def obtain_certificate(self, domains):
...
+                keypem = f.read()
+            key = util.Key(file=keypath, pem=keypem)
+            logger.info("New key obtained by reusing existing private key "
+                        "from %s.", oldkeypath)
+
         # Create CSR from names
         if self.config.dry_run:
-            key = util.Key(file=None,
-                           pem=crypto_util.make_key(self.config.rsa_key_size))
+            if oldkeypath is None:
```
ohemorange: `key = key or ...`

certbot/client.py
```diff
@@ -279,16 +279,36 @@ def obtain_certificate(self, domains):
...
+                key = util.Key(file=None,
+                               pem=crypto_util.make_key(self.config.rsa_key_size))
             csr = util.CSR(file=None, form="pem",
                            data=acme_crypto_util.make_csr(
                                key.pem, domains, self.config.must_staple))
         else:
-            key = crypto_util.init_save_key(
-                self.config.rsa_key_size, self.config.key_dir)
+            if oldkeypath is None:
```
ohemorange: `key = key or ...`

certbot/client.py
```diff
@@ -279,16 +279,36 @@ def obtain_certificate(self, domains):
...
+
+        if oldkeypath is not None:
+            # We've been asked to reuse a specific existing private key.
+            # Therefore, we'll read it now and not generate a new one in
+            # either case below.
+            with open(oldkeypath, "r") as f:
+                keypath = oldkeypath
+                keypem = f.read()
+            key = util.Key(file=keypath, pem=keypem)
+            logger.info("New key obtained by reusing existing private key "
```
ohemorange: Perhaps clearer to drop the "new key obtained by"

certbot/client.py
```diff
@@ -279,16 +279,36 @@ def obtain_certificate(self, domains):
...
+                        "from %s.", oldkeypath)
+
```
ohemorange: if you stick an `else: key = None` here, the rest gets a little cleaner

certbot/cli.py
```diff
@@ -1003,6 +1003,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
              "but does not match the requested domains, renew it now, "
              "regardless of whether it is near expiry.")
+    helpful.add(
+        "automation", "--reuse-key", dest="reuse_key",
+        action="store_true", default=flag_default("reuse_key"),
+        help="When renewing, use the same private key as the existing "
+             "certificate.")
+    helpful.add(
+        "automation", "--no-reuse-key", dest="reuse_key",
+        action="store_false", default=flag_default("reuse_key"),
```
ohemorange: did we ever figure out why this works? as in where the magic code lives. just out of curiosity

certbot/renewal.py
```diff
@@ -295,7 +295,13 @@ def renew_cert(config, domains, le_client, lineage):
     _avoid_invalidating_lineage(config, lineage, original_server)
     if not domains:
         domains = lineage.names()
-    new_cert, new_chain, new_key, _ = le_client.obtain_certificate(domains)
+    if config.reuse_key:
```
ohemorange:
```
old_key = lineage.privkey if config.reuse_key else None
new_cert, new_chain, new_key, _ = le_client.obtain_certificate(domains, old_key)
```

certbot/tests/main_test.py
```diff
@@ -1023,8 +1023,9 @@ def test_certonly_new_request_failure(self):
 
     def _test_renewal_common(self, due_for_renewal, extra_args, log_out=None,
                              args=None, should_renew=True, error_expected=False,
-                                 quiet_mode=False, expiry_date=datetime.datetime.now()):
-        # pylint: disable=too-many-locals,too-many-arguments
+                             quiet_mode=False, expiry_date=datetime.datetime.now(),
+                             reuse_key=False):
+        # pylint: disable=too-many-locals,too-many-arguments,too-many-branches
```
ohemorange: Is there a way to add the tests that doesn't stick the checks into `_test_renewal_common`? We already overload this with too much stuff.

tests/boulder-integration.sh
```diff
@@ -340,6 +340,26 @@ if common certificates | grep "fail\.dns1\.le\.wtf"; then
...
 fi
 
+# reuse-key
+common --domains reusekey.le.wtf --reuse-key
+common renew --cert-name reusekey.le.wtf
+CheckCertCount "reusekey.le.wtf" 2
+ls -l "${root}/conf/archive/reusekey.le.wtf/privkey"*
+# The final awk command here exits successfully if its input consists of
+# exactly two lines with identical first fields, and unsuccessfully otherwise.
+sha256sum "${root}/conf/archive/reusekey.le.wtf/privkey"* | awk '{a[$1] = 1}; END {exit(NR !=2 || length(a)!=1)}'
```
ohemorange: let's stick this `awk` command into a named parametrized function

certbot/client.py
schoen: I didn't think so because it does create a new file, but maybe users will be able to figure it out!

certbot/cli.py
schoen: Nope, I'm not really sure!

certbot/tests/main_test.py
schoen: I agree, but I think it might look even more complex by defining a whole new set of ways to test renewals!

certbot/cli.py
bmw: I'm not sure if this exactly answers your question, but we have a number of flags like `--foo` and `--no-foo`. For example, take a look at [\-\-redirect](https://github.com/certbot/certbot/blob/36dfd065037678bf38b7adb95ab16ed951ebbd18/certbot/cli.py#L1110).

`--redirect` sets `redirect` on the config object to `True` while `--no-redirect` sets it to `False`. Both `--redirect` and `--no-redirect` share the same default which isn't `True` or `False`, but [None](https://github.com/certbot/certbot/blob/36dfd065037678bf38b7adb95ab16ed951ebbd18/certbot/constants.py#L59). This allows us to differentiate between the user saying yes or no to a redirect and not answering the question.

Do we want to do anything special in the case where the user doesn't provide either `--reuse-key` or `--no-reuse-key`? If not, I think we can probably get rid of `--no-reuse-key` and keep the default `reuse_key` value to `False`. If we do want this for some reason, we should probably change the default value from `False` to `None`.

schoen: After talking to @bmw, I more clearly understand the point that for options that don't relate to prompts, we don't need a separate `--no` version because simply leaving the option out when reissuing with `certonly` is already sufficient to set it to `False`.

ArchangeGabriel:
Maybe I’m missing something, but I don’t understand when/how old_keypath is set.

ohemorange:
It's passed in as an argument to `obtain_certificate` when called from `renewal.py`. Are you working on a PR that you're looking for help with?

ArchangeGabriel:
No, I’m trying to understand how to get certbot to reuse the same key for each request on my server.

ArchangeGabriel:
(For existing lineage as well as new ones)

pgporada:
Seth,

This just came in extremely handy to test a system we've been working on. Thank you.

PR #5902, opened 2018-04-26
Reuse ACMEv1 accounts for ACMEv2

### LINKED ISSUE #5820: Reuse ACMEv1 accounts for ACMEv2
Boulder implemented this and making use of it is a better experience for our users and their servers.

ohemorange:
Fixes #5820.

Do not merge yet.

### DIFF
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -157,7 +157,7 @@ class AccountFileStorage(interfaces.AccountStorage):
     def _metadata_path(cls, account_dir_path):
         return os.path.join(account_dir_path, "meta.json")
 
-    def find_all(self, first_pass=True):
+    def _find_all_inner(self, first_pass):
         try:
             candidates = os.listdir(self.config.accounts_dir)
         except OSError:
@@ -182,9 +182,12 @@ class AccountFileStorage(interfaces.AccountStorage):
                                    self.config.strict_permissions)
             os.symlink(prev_account_dir, self.config.accounts_dir)
             # try again
-            return self.find_all(first_pass=False)
+            return self._find_all_inner(False)
         else:
             return accounts
+
+    def find_all(self):
+        return self._find_all_inner(True)
 
     def load(self, account_id):
         account_dir_path = self._account_dir_path(account_id)
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -143,7 +143,11 @@ class AccountFileStorage(interfaces.AccountStorage):
                                    self.config.strict_permissions)
 
     def _account_dir_path(self, account_id):
-        return os.path.join(self.config.accounts_dir, account_id)
+        return self._account_dir_path_for_server_path(account_id, self.config.server_path)
+
+    def _account_dir_path_for_server_path(self, account_id, server_path):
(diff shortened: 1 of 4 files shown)


certbot/constants.py
```diff
@@ -156,6 +156,13 @@
 ACCOUNTS_DIR = "accounts"
 """Directory where all accounts are saved."""
 
+LE_REUSE_SERVERS = {
+    'acme-v02.api.letsencrypt.org/directory': 'acme-v01.api.letsencrypt.org/directory',
+    'acme-staging-v02.api.letsencrypt.org/directory':
+        'acme-staging-v01.api.letsencrypt.org/directory'
```
ohemorange: this should be `acme-staging.api...`

cpu:
@ohemorange This should be unblocked in Boulder master but if you're looking to test against staging then it will be a little longer (likely next week).

bmw:
> FYI, account reuse now works on staging.

Turns out I was wrong about this, but it should be soon.

When testing this, we should make sure that a cert issued using ACMEv1 and an ACMEv1 account can be revoked on the ACMEv2 endpoint, especially when doing so by account key.

certbot/account.py
```diff
@@ -156,22 +161,41 @@ def _key_path(cls, account_dir_path):
...
+        if not accounts and server_path in constants.LE_REUSE_SERVERS:
+            # find all for the next link down
+            prev_server_path = constants.LE_REUSE_SERVERS[server_path]
+            prev_accounts = self._find_all_for_server_path(prev_server_path)
+            # if we found something, link to that
+            if prev_accounts:
+                if os.path.islink(accounts_dir):
+                    os.unlink(accounts_dir)
+                else:
+                    os.rmdir(accounts_dir)
```
bmw: This will fail if the directory isn't empty which could happen if the directory got corrupted somehow (which we handle gracefully above when loading accounts).

I'm open to other ideas, but if this fails, how about we return an empty list? In this corner case, the account won't be reused, but at least we won't blow up.

certbot/constants.py
```diff
@@ -156,6 +156,13 @@
 ACCOUNTS_DIR = "accounts"
 """Directory where all accounts are saved."""
 
+LE_REUSE_SERVERS = {
+    'acme-v02.api.letsencrypt.org/directory': 'acme-v01.api.letsencrypt.org/directory',
```
bmw: This feature is in staging now but won't be in prod until after our release next week. Would you prefer:

1. We temporarily remove this line and merge this PR, enabling account reuse in staging (which will affect people using flags like `--dry-run`).
2. We delay this entire PR until it's enabled in prod as well.

I prefer the former.

certbot/account.py
ohemorange: I like that

certbot/constants.py
ohemorange: yes, former is great

ohemorange: nice to have staging test account reuse anyway

certbot/account.py
ohemorange: the other option is to just remove the entire tree, but I'd rather keep it around and create a new account.

PR #5911, opened 2018-05-02
Add autorenew option to `renew` subcommand

r5d:
Notes:

  - The `autorenew` flag check is done in
    `certbot.renewal.handle_renewal_request`; not sure if this the right
    place to do it.

  - The `webroot_path` and `installer` fields are getting removed from
  renewalparams (in renewal conf file) after the first `certbot renew
  --cert-name ... --no-autorenew` or `certbot renew --cert-name ... --autorenew`.

Addresses issue #4504.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1192,6 +1192,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         default=flag_default("directory_hooks"), dest="directory_hooks",
         help="Disable running executables found in Certbot's hook directories"
         " during renewal. (default: False)")
+    helpful.add(
+        "renew", "--autorenew", action="store_true",
+        dest="autorenew", help="Enable auto renewal of certificates."
+        )
+    helpful.add(
+        "renew", "--no-autorenew", action="store_false",
+        dest="autorenew", help="Disable auto renewal of certificates."
+        )
 
     helpful.add_deprecated_argument("--agree-dev-preview", 0)
     helpful.add_deprecated_argument("--dialog", 0)
--- a/certbot/constants.py
+++ b/certbot/constants.py
@@ -37,6 +37,7 @@ CLI_DEFAULTS = dict(
     expand=False,
     renew_by_default=False,
     renew_with_new_domains=False,
+    autorenew=True,
     allow_subset_of_names=False,
     tos=False,
     account=None,
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -4,6 +4,7 @@ import copy
 import itertools
 import logging
 import os
+import sys
 import traceback
 
 import six
@@ -33,7 +34,7 @@ STR_CONFIG_ITEMS = ["config_dir", "logs_dir", "work_dir", "user_agent",
                     "pre_hook", "post_hook", "tls_sni_01_address",
                     "http01_address"]
 INT_CONFIG_ITEMS = ["rsa_key_size", "tls_sni_01_port", "http01_port"]
-BOOL_CONFIG_ITEMS = ["must_staple", "allow_subset_of_names"]
+BOOL_CONFIG_ITEMS = ["must_staple", "allow_subset_of_names", "autorenew"]
 
 CONFIG_ITEMS = set(itertools.chain(
     BOOL_CONFIG_ITEMS, INT_CONFIG_ITEMS, STR_CONFIG_ITEMS, ('pref_challs',)))
@@ -362,7 +363,7 @@ def _renew_describe_results(config, renew_successes, renew_failures,
     disp.notification("\n".join(out), wrap=False)
 
 
-def handle_renewal_request(config):
+def handle_renewal_request(config): # pylint: disable=too-many-locals,too-many-branches
     """Examine each lineage; renew if due and report results"""
 
     # This is trivially False if config.domains is empty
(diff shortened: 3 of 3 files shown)


bmw:

ohemorange:
Yes, I plan to.

certbot/renewal.py
```diff
@@ -408,6 +409,16 @@ def handle_renewal_request(config):
         try:
             if renewal_candidate is None:
                 parse_failures.append(renewal_file)
+            elif lineage_config.certname  and cli.set_by_cli('autorenew'):
```
ohemorange: Let's move the overall check into `storage.should_autorenew`.

certbot/renewal.py
```diff
@@ -408,6 +409,16 @@ def handle_renewal_request(config):
...
+                # Update renewal config file and exit.
+                symlinks = dict((kind, renewal_candidate.configuration[kind])
```
ohemorange: shouldn't have to manually be setting this in the renewal config file, once it's added to the list of bools above

certbot/renewal.py
```diff
@@ -408,6 +409,16 @@ def handle_renewal_request(config):
...
+                                for kind in storage.ALL_FOUR)
+                storage.update_configuration(lineagename,
+                        renewal_candidate.archive_dir, symlinks, lineage_config)
+                disp.notification("{} auto renewal for {}".format(
+                    "Enabled" if config.autorenew else "Disabled",
+                    lineagename))
+                sys.exit(0)
```
ohemorange: moving the check to `storage` will automatically handle catching certs we don't renew, so shouldn't need to `sys.exit`.

certbot/cli.py
```diff
@@ -1200,6 +1200,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " when the user executes \"certbot renew\", regardless of if the certificate"
         " is renewed. This setting does not apply to important TLS configuration"
         " updates.")
+    helpful.add(
```
ohemorange: you'll also want a `default=flag_default("autorenew")` for both of these

r5d: Done.

certbot/renewal.py
r5d: Moved the check to `certbot.storage.RenewableCert.should_autorenew`.

certbot/renewal.py
r5d: I'm slightly confused here. I thought the `BOOL_CONFIG_ITEMS` was used to set the data type of the value of a field in renewalparams when certbot reads the field from the renewal config file.

I added this
```python
symlinks = dict((kind, renewal_candidate.configuration[kind])
                     for kind in storage.ALL_FOUR)
     storage.update_configuration(lineagename,
             renewal_candidate.archive_dir, symlinks, lineage_config)
```
to set `autorenew` under `[renewalparams]` in the renewal config file like:

```conf
# renew_before_expiry = 30 days
version = 0.25.0.dev0
archive_dir = /tmp/leitKHTl/conf/archive/ara.ricketyspace.net
cert = /tmp/leitKHTl/conf/live/ara.ricketyspace.net/cert.pem
privkey = /tmp/leitKHTl/conf/live/ara.ricketyspace.net/privkey.pem
chain = /tmp/leitKHTl/conf/live/ara.ricketyspace.net/chain.pem
fullchain = /tmp/leitKHTl/conf/live/ara.ricketyspace.net/fullchain.pem

# Options used in the renewal process
[renewalparams]
account = 73398b4278fc1b489d3ab10dc8b35bb5
work_dir = /tmp/leitKHTl/work
server = https://acme-staging-v02.api.letsencrypt.org/directory
authenticator = webroot
logs_dir = /tmp/leitKHTl/logs
config_dir = /tmp/leitKHTl/conf
autorenew = True
[[webroot_map]]
ara.ricketyspace.net = /home/s/pub
```

Will certbot automatically set the `autorenew` field in the renewal config file when `--autorenew` / `--no-autorenew` option is given in the commandline?

certbot/renewal.py
r5d: Removed `sys.exit`.

Due to this, when the user does

```bash
# certbot renew --cert-name foo.bar.pub  --autorenew
```
certbot  will enable autorenew (set the `autorenew = True` in renewal config file) and will try to renew the certificate for `foo.bar.pub`. Is this the desired behavior or do we want certbot to 
 just enable autorenew and exit?

certbot/storage.py
```diff
@@ -934,16 +936,25 @@ def should_autorenew(self, interactive=False):
...
-            regardless of whether the renewal configuration allows
-            automated renewal (for interactive use). Default False.
-
         :returns: whether an attempt should now be made to autorenew the
             most current cert version in this lineage
         :rtype: bool
 
         """
-        if interactive or self.autorenewal_is_enabled():
+        if cli.set_by_cli('autorenew'):
```
ohemorange: you can delete this entire `if` block

certbot/cli.py
ohemorange: actually, let's remove this option for simplicity, since we don't need to know whether or not it was set on the command line, we just need to know if the value is true or false.

certbot/constants.py
```diff
@@ -37,6 +37,7 @@
     expand=False,
     renew_by_default=False,
     renew_with_new_domains=False,
+    autorenew=None,
```
ohemorange: and then this default will be `True`

PR #5918, opened 2018-05-02
Turn off cancel notifications

bmw:
By default, Travis cancels builds on branches and PRs if another commit lands in favor of running tests on that. See https://blog.travis-ci.com/2017-09-21-default-auto-cancellation.

I think this behavior is good as it stops us spending resources testing out of date code, but with our current setup, it causes Travis to send us notifications when builds are automatically canceled on master.

This PR changes that by not sending out notifications when a build gets canceled.


bmw:
@joohoi or @sydneyli, want to take a crack at reviewing this?

PR #5933, opened 2018-05-09
Use new boulder IP

jsha:
Boulder changed its docker-compose.yml to no longer publish its ports to localhost by default; instead it sets up on 10.77.77.77 on the local Docker network. This updates the Certbot integration tests to work with those changes.


sydneyli:
Closing this since #5936 has been resolved. @jsha, let us know if you still need something like this and we can re-open and create a new GH issue for it.

PR #5934, opened 2018-05-09
Silence spammy integration test cases.

jsha:


ohemorange:
Thanks for making this change! 

Few things:
- assigning a reviewer means we see PRs, and not just months later when we do a cleanup
- since then, we've moved the relevant code to a different file. probably easier to redo than try to do some merging.
- space before and after operator plz: `command > /dev/null`

jsha:
Updated to use the new file, and made the whitespace change. Thanks for the review!

PR #5936, opened 2018-05-09
Use new boulder IP v2

bmw:


bmw:
Supersedes #5933 by making a one line change to the PR. There is more discussion about this on that PR.

bmw:
My single commit here fixes ACMEv2 tests by actually making them use ACMEv2. The `if` statement around setting up ACMEv2 tests checks that `SERVER` isn't already set so we have to set it after that check.

My hypothesis for why boulder was returning the wrong value in the `link` header is in #5933.

PR #5937, opened 2018-05-09
Temporarily disable apacheconftest

bmw:
This allows tests to begin working again while staging is down for maintenance.


bmw:
I created #5938 to suggest a way to improve these tests and remove the dependency on the staging server.

Thanks for the review Jacob.

PR #5979, opened 2018-05-14
Check_untyped_defs in mypy with clean output for certbot-dns-luadns

### LINKED ISSUE #5971: Check unchecked types in certbot-dns-luadns
This issue is for checking untyped definitions in the `certbot-dns-luadns` module.

To start, add the following to `mypy.ini` in the root of the repo:
```
[mypy-certbot_dns_luadns.*]
check_untyped_defs = True
```

Then run `tox -e mypy`. If passes, submit a PR. If not, fix errors and submit a PR when tests pass. You can read more about how all of this works below.

See explanation here: https://docs.google.com/document/d/1fp0tPP9bzYkUViTnOXAyoT_KfgTfs8HSpNMoU8Tm_SY/edit?usp=sharing

From http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/:

Add the `--check-untyped-defs` option to your mypy arguments, and get that running with no errors on the codebase. This option causes mypy to check every def in the codebase for internal consistency; mypy can detect many classes of bugs and mistakes in your codebase, without your having written a single type annotation!

In most cases, you’ll want to fix bugs and bad code as you go, but you can also use #type: ignore annotations or exclude files to defer issues. For example, we excluded all of Zulip’s tests at first, since they’re both lower value to type-check and had most of the monkey-patching and other questionable Python in the project. For Zulip, getting clean `--check-untyped-defs` output took me about 2 days of hard work, including merging fixes for about 40 issues in the Zulip codebase.

I spent another day or two generating nice reproducers for the mypy bugs I’d encountered and improving typeshed. Now that mypy is no longer in its infancy, mypy bugs are increasingly rare. But a large project should expect to encounter and fix typeshed issues (just submit a PR!).

Once `mypy --check-untyped-defs` passes on your codebase, you should lock down your progress by running the mypy checker in your CI environment.

Because mypy type annotations are optional, once you’ve done the setup above, you can start annotating your codebase at whatever pace you like. Over time, you’ll get the benefits of static types in the parts of the codebase that you’ve annotated, without needing to do anything special to the rest of the codebase (the wonders of gradual typing!). In the next section, I’ll discuss strategy for how to move the codebase towards being fully annotated.

douganger:
* check_untyped_defs in mypy with clean output for certbot-dns-luadns

Resolves #5971

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,3 +7,6 @@ check_untyped_defs = True
 
 [mypy-certbot_dns_rfc2136.*]
 check_untyped_defs = True
+
+[mypy-certbot_dns_luadns.*]
+check_untyped_defs = True


bmw:
I canceled tests for now because I'll need to resolve merge conflicts after another PR is merged and run tests on the result, but it looked like tests were passing fine. Tests will start again when the merge commit is added.

bmw:
I built on this PR in #5982 which includes your commits, but handles merge conflicts between the various other similar PRs.

Issue #5908, opened 2018-05-01
Improve failed to bind error in standalone


bmw:
The way this code works is we first try to bind using IPv6 and then IPv4. On most UNIX-like systems, IPv4 traffic is mapped to IPv6 by default so listening on IPv6 gets traffic from both protocols and we fail to bind using IPv4. When this happens, the following message is logged:
```
Failed to bind to :80 using IPv4
```
This often confuses people trying to debug why they failed a challenge. They immediately think the problem is Certbot failed to bind to the port which it sometimes is, but it also could be a firewall, proxy, or misconfigured DNS preventing the traffic from getting to Certbot.

A couple options for fixing this could be that we expand on this log message or also log a DEBUG level message when we successfully bind to the port.

bmw:
There's progress on this in #5985.

bmw:
Fixed in #5985.

dmugtasimov:
where can I find `--standalone-address` argument format?

is it something like `--standalone-address 127.0.0.1:8000`?

PR #5985, opened 2018-05-14
Log cases when standalone fails to bind a port.

kevinkle:
Hi @bmw , per #5908 I've added 2 more log messages wrt if Certbot bound successfully using IPv6/IPv4.

### DIFF
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -88,10 +88,9 @@ class BaseDualNetworkedServers(object):
             except socket.error:
                 if self.servers:
                     # Already bound using IPv6.
-                    logger.debug("""Certbot wasn't able to bind to %s:%s using
-                    %s, this is often expected due to the dual stack nature of
-                    IPv6 socket implementations.""", new_address[0],
-                    new_address[1], "IPv6" if ip_version else "IPv4")
+                    logger.debug("Certbot wasn't able to bind to %s:%s using %s, this is often " +
+                    "expected due to the dual stack nature of IPv6 socket implementations.",
+                    new_address[0], new_address[1], "IPv6" if ip_version else "IPv4")
                 else:
                     logger.debug("Failed to bind to %s:%s using %s", new_address[0],
                         new_address[1], "IPv6" if ip_version else "IPv4")
--- a/acme/acme/standalone.py
+++ b/acme/acme/standalone.py
@@ -88,7 +88,10 @@ class BaseDualNetworkedServers(object):
             except socket.error:
                 if self.servers:
                     # Already bound using IPv6.
-                    logger.debug("Failed to bind to %s:%s using %s, but continuing as we've bound using %s",           new_address[0], new_address[1], "IPv6" if ip_version else "IPv4", "IPv6")
+                    logger.debug("""Certbot wasn't able to bind to %s:%s using
+                    %s, this is often expected due to the dual stack nature of
+                    IPv6 socket implementations.""", new_address[0],
+                    new_address[1], "IPv6" if ip_version else "IPv4")
                 else:
                     logger.debug("Failed to bind to %s:%s using %s", new_address[0],
                         new_address[1], "IPv6" if ip_version else "IPv4")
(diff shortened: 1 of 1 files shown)


acme/acme/standalone.py
```diff
@@ -83,9 +83,15 @@ def __init__(self, ServerClass, server_address, *remaining_args, **kwargs):
...
                 new_args = (new_address,) + remaining_args
                 server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
-            except socket.error:
-                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                logger.debug("Successfully bound to %s:%s using %s", new_address[0],
                     new_address[1], "IPv6" if ip_version else "IPv4")
+            except socket.error:
+                if self.servers:
+                    # Already bound using IPv6.
+                    logger.debug("Failed to bind to %s:%s using %s, but continuing as we've bound using %s",           new_address[0], new_address[1], "IPv6" if ip_version else "IPv4", "IPv6")
```
joohoi: Linter fails here because of the missing newline and indentation. In addition to that, I think we could be more informative here with error message like:

```
Certbot wasn't able to bind to %s:%s using %s, this is often expected due to the dual stack nature of IPv6 socket implementations.
```
I speak only from the personal experience, but seeing `Failed` anywhere in the log entries often takes full focus.

kevinkle: The suggestion sounds good! It makes sense to not have `failed` if it'll continue okay.
I'll make the changes tomorrow and ping you.

acme/acme/standalone.py
```diff
@@ -83,9 +83,18 @@ def __init__(self, ServerClass, server_address, *remaining_args, **kwargs):
                 new_address = (server_address[0],) + (port,) + server_address[2:]
                 new_args = (new_address,) + remaining_args
                 server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
+                logger.debug("Successfully bound to %s:%s using %s", new_address[0],
+                    new_address[1], "IPv6" if ip_version else "IPv4")
             except socket.error:
-                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                if self.servers:
+                    # Already bound using IPv6.
+                    logger.debug("""Certbot wasn't able to bind to %s:%s using
```
joohoi: We have used the following notation for multiline strings elsewhere in the code, and would prefer it for consistency as well as readability:
```
if self.servers:
    # Already bound using IPv6.
    logger.debug("Certbot wasn't able to bind to %s:%s using %s, this is often " +
                 "expected due to the dual stack nature of IPv6 socket implementations.",
                 new_address[0], new_address[1], "IPv6" if ip_version else "IPv4")
```

My apologies for not pointing that out earlier on.

kevinkle: No worries, thanks for the tip! 👍

acme/acme/standalone.py
```diff
@@ -83,9 +83,17 @@ def __init__(self, ServerClass, server_address, *remaining_args, **kwargs):
...
                 server = ServerClass(*new_args, **kwargs) # pylint: disable=star-args
-            except socket.error:
-                logger.debug("Failed to bind to %s:%s using %s", new_address[0],
+                logger.debug("Successfully bound to %s:%s using %s", new_address[0],
                     new_address[1], "IPv6" if ip_version else "IPv4")
+            except socket.error:
+                if self.servers:
+                    # Already bound using IPv6.
+                    logger.debug("Certbot wasn't able to bind to %s:%s using %s, this is often " +
+                    "expected due to the dual stack nature of IPv6 socket implementations.",
```
joohoi: Indentation is a bit off. 

Our developer guide has a section: [Coding Style](https://certbot.eff.org/docs/contributing.html#coding-style) which in turn points to [PEP-8](https://www.python.org/dev/peps/pep-0008/#indentation) and [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#Indentation), both of which describe preferred way to handle indentation. The links should take you directly to the correct part of the documents.

joohoi:
LGTM! Thanks for working on this!

PR #5989, opened 2018-05-14
Issue 5951/check untyped defs apache

dschlessman:
* Add types where `tox -e mypy` reported errors
* Addressed mypy error where logger.fatal function not found: logger.critical used instead.
* Removed unused conf argument in get_apache_configurator due to mypy error, "error: Cannot assign to a method”, in line `config.conf = conf`

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -159,7 +159,7 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
         # used by deploy_cert() and enhance()
         self._wildcard_vhosts = dict()  # type: Dict[str, List[obj.VirtualHost]]
         # Maps enhancements to vhosts we've enabled the enhancement for
-        self._enhanced_vhosts = defaultdict(set)  # type: DefaultDict[str , Set[obj.VirtualHost]]
+        self._enhanced_vhosts = defaultdict(set)  # type: DefaultDict[str, Set[obj.VirtualHost]]
 
         # These will be set in the prepare function
         self.parser = None
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -19,7 +19,7 @@ from certbot import errors
 from certbot import interfaces
 from certbot import util
 
-from certbot.achallenges import KeyAuthorizationAnnotatedChallenge
+from certbot.achallenges import KeyAuthorizationAnnotatedChallenge  # pylint: disable=unused-import
 from certbot.plugins import common
 from certbot.plugins.util import path_surgery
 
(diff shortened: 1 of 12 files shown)


certbot-apache/certbot_apache/configurator.py
```diff
@@ -150,14 +152,14 @@ def __init__(self, *args, **kwargs):
         super(ApacheConfigurator, self).__init__(*args, **kwargs)
 
         # Add name_server association dict
-        self.assoc = dict()
+        self.assoc = dict()  # type: Dict[str, obj.VirtualHost]
         # Outstanding challenges
-        self._chall_out = set()
+        self._chall_out = set()  # type: Set[Tuple[KeyAuthorizationAnnotatedChallenge, ...]]
```
dschlessman: Due to enumerate(iterable) used later on self._chall_out in another function, it appears that a list may be valid as well. In this case the type could be Set[Union[List[KeyAuthorizationAnnotatedChallenge], Tuple[KeyAuthorizationAnnotatedChallenge, ...]]]. However, the automated tests use a tuple. And the aforementioned type annotation is about 109 characters by itself, which leads to issues passing pylint. I don't think there is a valid way to break up that type annotation to pass pylint and be a valid type annotation.

certbot-apache/certbot_apache/tests/util.py
```diff
@@ -133,10 +132,6 @@ def mock_os_constant(key, vhost_path=vhost_path):
                         config_class = configurator.ApacheConfigurator
                     config = config_class(config=mock_le_config, name="apache",
                         version=version)
-                    # This allows testing scripts to set it a bit more
-                    # quickly
-                    if conf is not None:
-                        config.conf = conf  # pragma: no cover
```
dschlessman: This change could use extra attention.

bmw:
Sorry for the cryptic and unexpected test failures here. What's going on is the different pieces packages in this repo are not in required lockstep. For example, you can have a one version of `certbot` and another version of `certbot-apache`. We have tests to make sure that the minimum version we say we require is correct which is failing here because `acme.magic_typing` is new for us.

To fix this, you should modify `certbot-apache/setup.py` to require `acme>0.24.0` and the `acme` line in `certbot-apache/local-oldest-requirements.txt` to `-e acme[dev]`. You can take my word on the latter or I can explain to you how this works if you're curious.

certbot-apache/certbot_apache/configurator.py
dschlessman: You're right. Thank you.

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -348,19 +348,16 @@ def mock_find_dir(directive, argument, _):
         for a in mock_add.call_args_list:
             if a[0][1] == "Include" and a[0][2] == self.config.mod_ssl_conf:
                 tried_to_add = True
-        # Include should be added, find_dir is not patched, and returns falsy
+        # Include should be added, find_dir is not patched, and returns false
```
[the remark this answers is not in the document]
dschlessman: Oh! Thanks!

Issue #4242, opened 2017-02-22
Amount of email address to register


fpeterson194:
Hi everyone,

Is there a limitation in regards with the amount of email addresses I can register with the parameter -m? My goal is tu put more than one address. If we have this possibility, should I separate them by which character? Comma?

Thanks in advance!

SwartzCr:
Good question - it seems like you got some good responses on the community forums about how many emails you could register (https://community.letsencrypt.org/t/amount-of-email-addresses-to-register/28508)
Though I'm not sure how you can specify them using Certbot, maybe @bmw can help?

fpeterson194:
@SwartzCr thanks for your response. Yes, I get some responses from the community. The limit is 10 email addresses. Now my question is what you have already mention. How can I specify them in certbot and in acmesharp.

SwartzCr:
I think @bmw would be the best equipped to answer that

bmw:
Certbot unfortunately doesn't currently support registering multiple e-mail addresses for the same account. We should certainly add support for this in the future though.

fpeterson194:
Thanks very much @bmw ! we can close this issue now :)

bmw:
We should leave it open for tracking purposes. Thanks for bringing this issue up!

SwartzCr:
I updated the name to make it more clear what this is about
I'm gonna throw it in wishlist and we can try to bring it back into a milestone when we're ready
I added documentation to this so that maybe someone will document this (maybe when we document lineages)

shubham8agar:
Is the issue of adding multiple email ids has been resolved? I am still facing the same.

bmw:
https://tools.ietf.org/html/draft-ietf-acme-acme-10

bmw:
If you want to write an integration test once you've got this working, it'd go in `tests/boulder-integration.sh`.

vhquang:
I will be working on this one as part of the Pycon sprint.

PR #5994, opened 2018-05-14
#4242 Support multi emails register

vhquang:
This change will allow registering/updating account with multi emails. 
Detail is enclosed in #4242

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -381,7 +381,7 @@ VERB_HELP = [
     ("register", {
         "short": "Register for account with Let's Encrypt / other ACME server",
         "opts": ("Options for account registration & modification"
-                 ". To register multi email, separate by comma,"
+                 ". To register multi emails, separate by comma,"
                  " ex: u1@example.com,u2@example.com"),
         "usage": "\n\n  certbot register --email user@example.com [options]\n\n"
     }),
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -380,9 +380,7 @@ VERB_HELP = [
     }),
     ("register", {
         "short": "Register for account with Let's Encrypt / other ACME server",
-        "opts": ("Options for account registration & modification"
-                 ". To register multi emails, separate by comma,"
-                 " ex: u1@example.com,u2@example.com"),
+        "opts": "Options for account registration & modification",
         "usage": "\n\n  certbot register --email user@example.com [options]\n\n"
     }),
     ("unregister", {
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -380,7 +380,9 @@ VERB_HELP = [
     }),
     ("register", {
         "short": "Register for account with Let's Encrypt / other ACME server",
-        "opts": "Options for account registration & modification",
+        "opts": ("Options for account registration & modification"
+                 ". To register multi email, separate by comma,"
+                 " ex: u1@example.com,u2@example.com"),
         "usage": "\n\n  certbot register --email user@example.com [options]\n\n"
     }),
     ("unregister", {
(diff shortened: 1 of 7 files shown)


bmw:
`certbot` doesn't like that there's an existing account. Try running `certbot unregister` before running `certbot register --update-registration ...` and it should fix the problem.

certbot/main.py
```diff
@@ -483,6 +483,21 @@ def _determine_account(config):
...
+               "server at {1}".format(
+                   terms_of_service, config.server))
+        obj = zope.component.getUtility(interfaces.IDisplay)
+        result = obj.yesno(msg, "Agree", "Cancel",
+                         cli_flag="--agree-tos", force_interactive=True)
+        if not result:
+            raise errors.Error(
+                "Registration cannot proceed without accepting "
+                "Terms of Service.")
+
```
vhquang: This is just a refactor of moving the internal function from the middle of the function body to top, for better code style. The internal function remains the same.

certbot/cli.py
```diff
@@ -380,7 +380,9 @@ def _get_help_string(self, action):
     }),
     ("register", {
         "short": "Register for account with Let's Encrypt / other ACME server",
-        "opts": "Options for account registration & modification",
+        "opts": ("Options for account registration & modification"
+                 ". To register multi emails, separate by comma,"
+                 " ex: u1@example.com,u2@example.com"),
```
bmw: We should put the documentation for multiple email addresses under the `--email` flag rather than here. The help text for this flag is defined in `certbot.interfaces.IConfig`.

certbot/main.py
```diff
@@ -731,8 +731,9 @@ def register(config, unused_plugins):
     acc, acme = _determine_account(config)
     cb_client = client.Client(config, acc, None, None, acme=acme)
     # We rely on an exception to interrupt this process if it didn't work.
-    acc.regr = cb_client.acme.update_registration(acc.regr.update(
-        body=acc.regr.body.update(contact=('mailto:' + config.email,))))
+    acc_contacts = ['mailto:' + email for email in config.email.split(',')]
+    acc.regr.body.update(contact=acc_contacts)
```
bmw: I think there's actually a bug here. These `acc.regr` and `acc.regr.body` objects are immutable and `update` returns a newly created object leaving the old object unmodified. We need to pass this new objects returned by `updated` into `update_registration`.

bmw: If you want to take a shot at writing a unit test that would have caught this bug, it's probably a good idea, but you're welcome to skip this if you want. Doing so will require a fair bit of mocking. See `test_update_registration_with_email` in `certbot/tests/main_test.py` for an example.

certbot/cli.py
bmw: I'd also expand to "multi" to the full word "multiple".

certbot/interfaces.py
```diff
@@ -201,7 +201,9 @@ class IConfig(zope.interface.Interface):
     """
     server = zope.interface.Attribute("ACME Directory Resource URI.")
     email = zope.interface.Attribute(
-        "Email used for registration and recovery contact. (default: Ask)")
+        "Email used for registration and recovery contact. (default: Ask). "
+        "Use comma to register multiple emails, "
+        "ex: u1@example.com,u2@example.com")
```
bmw: In `argparse`'s help output, it adds "(default: <value>)" to the end of parameter to show it's default value. To consistently keep this convention, we should add the "Use comma..." text before (default: Ask).

vhquang: ah I see. Thanks for pointing that out. I will make an update.

vhquang:
@bmw the code is updated. If there is anything that you would like me to change, please let me know. Thanks!

PR #5997, opened 2018-05-15
Enable checking of type annotation in Nginx plugin

### LINKED ISSUE #5950: Check untyped defs in the Nginx plugin
This issue is for checking untyped definitions in the `certbot-nginx` module.

See explanation here: https://docs.google.com/document/d/1fp0tPP9bzYkUViTnOXAyoT_KfgTfs8HSpNMoU8Tm_SY/edit?usp=sharing

From http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/:

Add the `--check-untyped-defs` option to your mypy arguments, and get that running with no errors on the codebase. This option causes mypy to check every def in the codebase for internal consistency; mypy can detect many classes of bugs and mistakes in your codebase, without your having written a single type annotation!

In most cases, you’ll want to fix bugs and bad code as you go, but you can also use #type: ignore annotations or exclude files to defer issues. For example, we excluded all of Zulip’s tests at first, since they’re both lower value to type-check and had most of the monkey-patching and other questionable Python in the project. For Zulip, getting clean `--check-untyped-defs` output took me about 2 days of hard work, including merging fixes for about 40 issues in the Zulip codebase.

I spent another day or two generating nice reproducers for the mypy bugs I’d encountered and improving typeshed. Now that mypy is no longer in its infancy, mypy bugs are increasingly rare. But a large project should expect to encounter and fix typeshed issues (just submit a PR!).

Once `mypy --check-untyped-defs` passes on your codebase, you should lock down your progress by running the mypy checker in your CI environment.

Because mypy type annotations are optional, once you’ve done the setup above, you can start annotating your codebase at whatever pace you like. Over time, you’ll get the benefits of static types in the parts of the codebase that you’ve annotated, without needing to do anything special to the rest of the codebase (the wonders of gradual typing!). In the next section, I’ll discuss strategy for how to move the codebase towards being fully annotated.

jameshiebert:
This PR enables ``mypy`` type checking and adds type annotations to the Nginx plugin and fixes #5950 

One questionable part of this PR was that I had to [put an *ignore* line](https://github.com/certbot/certbot/compare/master...jameshiebert:issue-5950?expand=1#diff-961797d34f1f467f621e3414463f304bR251) `nginxparser.py`. It's in a class that inherits from ``list`` and adds a few properties (including `is_dirty`). The type checker sees the variable as a ``list`` proper and can't determine that it has the `is_dirty` property. If there's a way to make this work, I am open to it.

Also, [at one point](https://github.com/certbot/certbot/compare/master...jameshiebert:issue-5950?expand=1#diff-b2dab7e6fe39d3a1bf642d561d045633R733) I had to use a ``Dict[str, Any]`` which *could* be more specific. E.g.

~~~python
from mypy_extensions import TypedDict
ParsedServer = TypedDict('addrs': Set, 'ssl': bool, 'names': Set)
parsed_server = ... # type: ParsedServer
~~~

But having to add a dependency on ``mypy_extensions`` seemed undesirable. But it's an alternative option.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -28,7 +28,8 @@ from certbot_nginx import nginxparser
 from certbot_nginx import parser
 from certbot_nginx import tls_sni_01
 from certbot_nginx import http_01
-from acme.magic_typing import List, Dict, Set
+from acme.magic_typing import List, Dict, Set # pylint: disable=unused-import, no-name-in-module
+
 
 
 logger = logging.getLogger(__name__)
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -28,6 +28,7 @@ from certbot_nginx import nginxparser
 from certbot_nginx import parser
 from certbot_nginx import tls_sni_01
 from certbot_nginx import http_01
+from certbot_nginx import obj # pylint: disable=unused-import, no-name-in-module
 from acme.magic_typing import List, Dict, Set # pylint: disable=unused-import, no-name-in-module
 
 
@@ -100,8 +101,8 @@ class NginxConfigurator(common.Installer):
 
         # List of vhosts configured per wildcard domain on this run.
         # used by deploy_cert() and enhance()
-        self._wildcard_vhosts = {} # type: Dict[str, str]
-        self._wildcard_redirect_vhosts = {} # type: Dict[str, str]
+        self._wildcard_vhosts = {} # type: Dict[str, List[obj.VirtualHost]]
+        self._wildcard_redirect_vhosts = {} # type: Dict[str, List[obj.VirtualHost]]
 
         # Add number of outstanding challenges
         self._chall_out = 0
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -28,7 +28,7 @@ from certbot_nginx import nginxparser
 from certbot_nginx import parser
 from certbot_nginx import tls_sni_01
 from certbot_nginx import http_01
-from certbot_nginx import obj # pylint: disable=unused-import, no-name-in-module
+from certbot_nginx import obj # pylint: disable=unused-import
 from acme.magic_typing import List, Dict, Set # pylint: disable=unused-import, no-name-in-module
 
 
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -28,6 +28,7 @@ from certbot_nginx import nginxparser
 from certbot_nginx import parser
 from certbot_nginx import tls_sni_01
 from certbot_nginx import http_01
+from acme.magic_typing import List, Dict, Set
 
 
 logger = logging.getLogger(__name__)
@@ -98,8 +99,8 @@ class NginxConfigurator(common.Installer):
 
         # List of vhosts configured per wildcard domain on this run.
         # used by deploy_cert() and enhance()
-        self._wildcard_vhosts = {}
-        self._wildcard_redirect_vhosts = {}
+        self._wildcard_vhosts = {} # type: Dict[str, str]
+        self._wildcard_redirect_vhosts = {} # type: Dict[str, str]
 
         # Add number of outstanding challenges
         self._chall_out = 0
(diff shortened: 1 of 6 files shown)


certbot-nginx/certbot_nginx/parser.py
```diff
@@ -28,7 +28,7 @@ class NginxParser(object):
     """
 
     def __init__(self, root):
-        self.parsed = {}
+        self.parsed = {} # type: Dict
```
bmw: Not sure how much this buys us, but I think we can be a little more specific here. `type: Dict[str, Union[List, nginxparser.UnspacedList]]` should be correct.

@jameshiebert, I'd use the type in my comment for now, but I'm curious what you think @ohemorange.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -90,7 +90,7 @@ def _build_addr_to_ssl(self):
         """
         servers = self._get_raw_servers()
 
-        addr_to_ssl = {}
+        addr_to_ssl = {} # type: Dict
```
bmw: I think the more specific type here is `Dict[Tuple[str, str], Bool]`.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -104,9 +104,10 @@ def _build_addr_to_ssl(self):
 
     def _get_raw_servers(self):
         # pylint: disable=cell-var-from-loop
+        # type: () -> Dict
         """Get a map of unparsed all server blocks
         """
-        servers = {}
+        servers = {} # type: Dict
```
bmw: `Dict[str, Union[List, nginxparser.UnspacedList]]`

@jameshiebert, I'd use the type in my comment for now, but I'm curious what you think @ohemorange.

certbot-nginx/certbot_nginx/parser.py
```diff
@@ -729,7 +730,7 @@ def _parse_server_raw(server):
     """
     parsed_server = {'addrs': set(),
                      'ssl': False,
-                     'names': set()}
+                     'names': set()} # type: Dict[str, Any]
```
bmw: `Set[str]`?

certbot-nginx/certbot_nginx/tests/parser_test.py
```diff
@@ -99,7 +100,7 @@ def test__do_for_subarray(self):
                    ([[[0], [3], [4]], [[5], [3], [0]]], [])]
 
         for mylist, result in mylists:
-            paths = []
+            paths = [] # type: List[str]
```
bmw: I think this is `List[List[int]]`?

certbot-nginx/certbot_nginx/nginxparser.py
```diff
@@ -248,7 +248,7 @@ def is_dirty(self):
         """Recurse through the parse tree to figure out if any sublists are dirty"""
         if self.dirty:
             return True
-        return any((isinstance(x, list) and x.is_dirty() for x in self))
+        return any((isinstance(x, list) and x.is_dirty() for x in self)) # type: ignore
```
bmw: Changing the code here to `...isinstance(x, UnspacedList)...` and removing the `type: ignore` seems to work. This should be safe because if `x` is really a `list` and not `UnspacedList`, it won't have the `is_dirty` attribute and blow up.

jameshiebert: Hrm, I *thought* that I had tried that, but yes, that *does* seem to work. Thanks!

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -28,6 +28,7 @@
 from certbot_nginx import parser
 from certbot_nginx import tls_sni_01
 from certbot_nginx import http_01
+from certbot_nginx import obj # pylint: disable=unused-import, no-name-in-module
```
bmw: nit: We shouldn't need `no-name-in-module` here.

jameshiebert:
Should be good for another round if/when you have a few spare minutes.

certbot-nginx/certbot_nginx/configurator.py
jameshiebert: Good catch; done!

certbot-nginx/certbot_nginx/parser.py
ohemorange: yes, that's correct. hopefully with sydney's changes we can update this to not need a union.

certbot-nginx/certbot_nginx/parser.py
ohemorange: yes, that's correct. hopefully with sydney's changes we can update this to not need a union.

PR #6000, opened 2018-05-15
switch signature verification to use pure cryptography

reaperhulk:
On systems that prevent write/execute pages this prevents a segfault that is caused by pyopenssl creating a dynamic callback in the verification helper.


reaperhulk:
@bmw I can make it work on the older cryptography releases by not using the one shot functions. I'll swap that as well.

PR #6002, opened 2018-05-15
Fix crash when email submission endpoint unavailable

### LINKED ISSUE #5858: Certbot crashes if EFF e-mail submission endpoint is unavailable
Certbot will crash if the user's e-mail information can't be submitted to the EFF endpoint:

https://community.letsencrypt.org/t/issue-when-trying-to-create-certificate/59307/1

https://github.com/certbot/certbot/blob/master/certbot/eff.py#L74

It crashes with a somewhat relevant error if the endpoint is unavailable, but a confusing error if the endpoint returns something other than JSON. It should probably not crash at all in either case. If the endpoint specifically says that the user's e-mail address is malformed, that's probably worth stopping to report to the user, but if the endpoint simply doesn't reply usefully, it's probably preferable to continue since I wouldn't think we really want to make users wait in this situation.

(Adding the "bug" tag at least for the case where the endpoint returns a non-JSON reply because the error shown to the user in this case is unhelpful.)

douganger:
Handle KeyError and ValueError so that if the email submission endpoint
goes down, Certbot can still run.

Add tests to eff_test.py:
 - simulate non-JSON response as described in issue #5858
 - simulate JSON response without 'status' element

Non-JSON response throws an uncaught ValueError when attempting to
decode as JSON. A JSON response missing the 'status' element throws an
uncaught KeyError when checking whether status is True or False.

Teach _check_response to handle ValueError and KeyError and report an
issue to the user.

Rewrite if statement as assertion with try-except block to make error
handling consistent within the function. Update test_not_ok to make
mocked raise_for_status function raise a requests.exceptions.HTTPError.

Resolves #5858

### DIFF
--- a/certbot/eff.py
+++ b/certbot/eff.py
@@ -71,11 +71,17 @@ def _check_response(response):
 
     """
     logger.debug('Received response:\n%s', response.content)
-    if response.ok:
-        if not response.json()['status']:
-            _report_failure('your e-mail address appears to be invalid')
-    else:
+    try:
+        response.raise_for_status()
+        assert response.json()['status'] == True
+    except requests.exceptions.HTTPError:
         _report_failure()
+    except ValueError:
+        _report_failure('there was a problem with the server response')
+    except KeyError:
+        _report_failure('the server response was missing some information')
+    except AssertionError:
+        _report_failure('your e-mail address appears to be invalid')
 
 
 def _report_failure(reason=None):
--- a/certbot/eff.py
+++ b/certbot/eff.py
@@ -73,15 +73,12 @@ def _check_response(response):
     logger.debug('Received response:\n%s', response.content)
     try:
         response.raise_for_status()
-        assert response.json()['status'] == True
+        if response.json()['status'] == False:
+            _report_failure('your e-mail address appears to be invalid')
     except requests.exceptions.HTTPError:
         _report_failure()
-    except ValueError:
+    except (ValueError, KeyError):
         _report_failure('there was a problem with the server response')
-    except KeyError:
-        _report_failure('the server response was missing some information')
-    except AssertionError:
-        _report_failure('your e-mail address appears to be invalid')
 
 
 def _report_failure(reason=None):
--- a/certbot/tests/eff_test.py
+++ b/certbot/tests/eff_test.py
@@ -117,11 +117,30 @@ class SubscribeTest(unittest.TestCase):
 
     @test_util.patch_get_utility()
     def test_not_ok(self, mock_get_utility):
+        from certbot.eff import requests
         self.response.ok = False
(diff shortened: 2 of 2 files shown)


certbot/eff.py
```diff
@@ -71,11 +71,17 @@ def _check_response(response):
 
     """
     logger.debug('Received response:\n%s', response.content)
-    if response.ok:
-        if not response.json()['status']:
-            _report_failure('your e-mail address appears to be invalid')
-    else:
+    try:
+        response.raise_for_status()
+        assert response.json()['status'] == True
```
bmw: This is really a nitpick, but you can actually disable [assert statements](https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement) with a command line flag to `python` so I don't think we should rely on them for production behavior.

I'd personally fix this by not raising an exception when `status` is `False` and just use the value in an `if` statement to call `_report_failure`.

certbot/eff.py
```diff
@@ -71,11 +71,17 @@ def _check_response(response):
...
+    except requests.exceptions.HTTPError:
         _report_failure()
+    except ValueError:
+        _report_failure('there was a problem with the server response')
+    except KeyError:
+        _report_failure('the server response was missing some information')
```
bmw: I'd personally handle `ValueError` and `KeyError` in the same way and say there was a problem with the server response. I personally think this is a cleaner, user facing error message because exactly what the problem was isn't something the user can do anything about.

I don't have a strong preference though so if you like things better this way, it's fine with me.

certbot/tests/eff_test.py
```diff
@@ -117,12 +117,31 @@ def test_bad_status(self, mock_get_utility):
 
     @test_util.patch_get_utility()
     def test_not_ok(self, mock_get_utility):
+        from certbot.eff import requests
```
bmw: Let's just import this using `import requests` at the top of the module.

certbot/tests/eff_test.py
```diff
@@ -117,12 +117,31 @@ def test_bad_status(self, mock_get_utility):
...
+        self.response.json.side_effect = ValueError()
+        self._call()  # pylint: disable=no-value-for-parameter
+        actual = self._get_reported_message(mock_get_utility)
+        expected_part = 'problem'
+        self.assertTrue(expected_part in actual)
+
+    @test_util.patch_get_utility()
+    def test_response_json_missing_status_element(self, mock_get_utility):
+        self.json = {}
+        self.response.json.return_value = self.json
```
bmw: nit: You can replace this and the above line by calling `self.json.clear()`.

douganger:
Thanks for the code review @bmw. I appreciate the feedback!

Issue #3576, opened 2016-10-03
Add --no-binary flag to pip for packages without wheels


bmw:
`certbot-auto`/`letsencrypt-auto` require specific versions of all our Python dependencies and verifies the hashes of all downloaded packages. Doing this increases both the security and the stability of the scripts. To do this, however, we need to include a hash of each installation mechanism uploaded to PyPI (sdists, wheels, etc.).

Over the weekend, `pycparsing` uploaded a new installation mechanism to PyPI. Previously, they only had a sdist available, but over the weekend they uploaded a wheel. When using `pip` to install packages from PyPI, it prefers wheels over sdists. The result of all of this is when using `certbot-auto` to install `pycparsing`, `pip` would download the wheel, check it against the hash(es) we have for `pycparsing`, and exit with a message about an invalid hash.

`pycparsing` has since removed this wheel due to other problems with it resolving our issue. In general, I think it is uncommon to add a distribution mechanism for an already released version of a package. In my experience, things like this are usually added when a new version of the package is released.

With that said, there's nothing stopping us from having this problem in the future. To try and help mitigate this problem, perhaps we should add `--no-binary` to our usage of `pip` for packages that currently do not have a wheel available. Using this approach, if a wheel is added to one of our dependencies without bumping the package version, `certbot-auto` users are unaffected.

bmw:
@erikrose, I'd like your opinion on this. Do you think this is something we should spend time doing? Do you have a better solution? How common is it for a package to add new distribution mechanisms without bumping the version number?

pde:
Is there a chance that `pycparsing` is going to fix their wheel and re-upload it in the near future?

bmw:
Good [question](https://github.com/eliben/pycparser/issues/147#issuecomment-251264814).

erikrose:
>How common is it for a package to add new distribution mechanisms without bumping the version number?

I've never seen that happen before, but adding `--no-binary` is not a bad idea.

bmw:
`pycparser` maintainer's [answer](https://github.com/eliben/pycparser/issues/147#issuecomment-251465062) to @pde's question.

reubano:
I just ran into this issue and had to add `--no-binary` to `enum34`, `zope.interface`, and `mock` as well.

bmw:
@reubano, really? What OS are you on? What version of Python are you using?

reubano:
Mac OSX 10.9.5
Python 3.5.2
pip 8.0.3
[...]

PR #6010, opened 2018-05-16
fixed issue #3576

pdamodaran:


erikrose:
Nope! The only alternative is to do nothing and swat these down with quick updates as they crop up. I checked, and pip doesn't have any options like "Ignore files released after a given date."

PR #6013, opened 2018-05-16
Support RFC2136 DNS update without TSIG key

### LINKED ISSUE #5778: Optional TSIG key for certbot-dns-rfc2136
My DNS server authenticates RFC2136 DNS update requests based on the IP address of the requester. So, I don't have a TSIG key, secret and algorithm. It would be nice if certbot-dns-rfc2136 could treat them as optional.

Parabola GNU/Linux (rolling release)

cerbot 0.22.2 installed with OS package manager (pacman)

I ran the following command.
`certbot certonly --dns-rfc2136 --dns-rfc2136-credentials /path/to/credentials.ini --cert-name example.com -d "*.example.com" -d example.com --server https://acme-v02.api.letsencrypt.org/directory`

My credentials file has the following.
`dns_rfc2136_server=ip.address.of.server`

I got the following output.
```
Missing properties in credentials configuration file /root/scripts/rfc2136-credentials.ini:
 * Property "dns_rfc2136_name" not found (should be TSIG key name).
 * Property "dns_rfc2136_secret" not found (should be TSIG key secret).
```

certbot errors out because I didn't provide TSIG information. I expected TSIG to be optional.

arunisaac:
* Resolves #5778

Prior to this change, the dns_rfc2136 plugin would error out if no
TSIG key name and secret was provided.

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -59,14 +59,16 @@ class Authenticator(dns_common.DNSAuthenticator):
         if algorithm:
             if not self.ALGORITHMS.get(algorithm):
                 raise errors.PluginError("Unknown algorithm: {0}.".format(algorithm))
+            credentials.require({
+                'name': 'TSIG key name',
+                'secret': 'TSIG key secret',
+            })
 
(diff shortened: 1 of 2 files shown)


joohoi:
Oh, you are correct. Of course the credentials argument is needed to point out other variables as well. I'll continue with the final review soon.

PR #6020, opened 2018-05-18
Set correct Nginx server root on FreeBSD and Darwin

ohemorange:
Fork of #5862.

This patch should be merged not squashed to preserve authorship.


ohemorange:
Anyway turns out running FreeBSD in Docker on Linux is maybe impossible but definitely hard so I set up a new droplet to test but DNS is still propagating and I have to run but it will probably work? The result of `platform.system()` and the root path checks out.

ohemorange:
Works on FreeBSD

PR #6021, opened 2018-05-18
Get mypy passing with check_untyped_defs everywhere

bmw:
This finishes getting mypy to pass with `--check-untyped-defs` for everything in this repo. To make it easy to distribute, `letshelp-certbot` has no dependencies outside of the standard library so rather than adding some so rather than add some, I just copied `acme.magic_typing` to that package as well.

### DIFF
--- a/letshelp-certbot/letshelp_certbot/apache.py
+++ b/letshelp-certbot/letshelp_certbot/apache.py
@@ -16,7 +16,7 @@ import textwrap
 
 import six
 
-from letshelp_certbot.magic_typing import List
+from letshelp_certbot.magic_typing import List  # pylint: disable=unused-import, no-name-in-module
 
 _DESCRIPTION = """
 Let's Help is a simple script you can run to help out the Certbot


certbot-compatibility-test/certbot_compatibility_test/util.py
```diff
@@ -26,12 +26,12 @@ def create_le_config(parent_dir):
...
-    config["logs_dir"] = os.path.join(le_dir, "logs_dir")
-    os.makedirs(config["config_dir"])
-    os.mkdir(config["work_dir"])
-    os.mkdir(config["logs_dir"])
+    os.mkdir(le_dir)
+    for dir_name in ("config", "logs", "work"):
+        full_path = os.path.join(le_dir, dir_name)
+        os.mkdir(full_path)
+        full_name = dir_name + "_dir"
+        config[full_name] = full_path
```
ohemorange: this changes `config` to `config_dir`; that's definitely ok?

letshelp-certbot/letshelp_certbot/magic_typing.py
```diff
@@ -0,0 +1,16 @@
...
+class TypingClass(object):
+    """Ignore import errors by getting anything"""
+    def __getattr__(self, name):
+        return None
+
+try:
+    # mypy doesn't respect modifying sys.modules
+    from typing import *  # pylint: disable=wildcard-import, unused-wildcard-import
+    # pylint: disable=unused-import
+    from typing import Collection, IO  # type: ignore
```
ohemorange: what's up with this?

certbot-compatibility-test/certbot_compatibility_test/util.py
bmw: I'm not sure what you mean. The only change in the end result here should be the basename of the log directory is `logs` rather than `logs_dir`.

Overall change was caused by `mypy` being unsure about the type returned from `config[name]`, even if we set it earlier in the function.

letshelp-certbot/letshelp_certbot/magic_typing.py
bmw: This came up during the sprints.

[typing defines \_\_all\_\_](https://github.com/python/cpython/blob/3.6/Lib/typing.py#L24) which controls which names are imported when you do `from typing import *`. See https://docs.python.org/2/tutorial/modules.html#importing-from-a-package. We needed `Collection` and `IO` when fixing untyped defs for Certbot which aren't included in `__all__`.

I just copied the existing `acme.magic_typing` file here when creating this PR. I think there is some benefit in trying to keep this file somewhat consistent in the different places we have to copy it (there's a copy in `josepy` now too), but I can remove this line here if you prefer.

certbot-compatibility-test/certbot_compatibility_test/util.py
ohemorange: yes, that is the direction I meant. is that ok to change the name.

letshelp-certbot/letshelp_certbot/magic_typing.py
ohemorange: if we add Collection and IO explicitly in `acme.magic_typing`, can we just have the one file then? or do we still need this? is there a problem with importing them explicitly there?

certbot-compatibility-test/certbot_compatibility_test/util.py
bmw: Should be. Only references to `logs_dir` in this package is in this function and Certbot doesn't expect the directory to end in `logs_dir`.

letshelp-certbot/letshelp_certbot/magic_typing.py
bmw: > if we add Collection and IO explicitly in acme.magic_typing, can we just have the one file then?

I think we need at least one copy of `magic_typing` which is the copy here. While we're not currently using this package, I think it's beneficial to not have any dependencies to make it trivial for someone to download and use this.

If we really wanted, we can just make use of the `josepy.magic_typing` and remove `acme.magic_typing` so the file in `letshelp_certbot` is the only copy, but I believe you didn't want to do this before.

> is there a problem with importing them explicitly there?

Importing them explicitly hurts nothing. It just makes them available when `from typing import *` does not.

PR #6030, opened 2018-05-23
Change GenericUpdater parameter to lineage

joohoi:
In order to give more flexibility for plugins using `interfaces.GenericUpdater` interface, lineage needs to be passed to the updater method instead of individual domains. All of the (present and potential) installers do not work on per domain basis, while the lineage does contain a list of them for installers which do. 

This also means that we don't unnecessarily run the updater method multiple times, potentially invoking expensive tooling up to $max_san_amount times.


certbot/interfaces.py
```diff
@@ -623,17 +623,19 @@ class GenericUpdater(object):
...
         function will always be called when "certbot renew" is run. If the
         update defined by the installer should be run conditionally, the
         installer needs to handle checking the conditions itself.
 
-        This method is called once for each domain.
+        This method is called once for each lineage.
 
-        :param str domain: domain to handle the updates for
+        :param lineage: Certificate lineage object that is set if certificate
+            was renewed on this run.
```
bmw: Is this documentation correct? Isn't the `lineage` always set?

joohoi: It's not! Nice catch, same stands true for both of the interfaces.

PR #6032, opened 2018-05-23
Clean up boulder-fetch a bit.

jsha:
The value for FAKE_DNS is now always the same, so skip some sed.
Set a time limit on how long we'll wait for boulder to come up.


jsha:
We recently switched up the Docker networking config in docker-compose.yml so Boulder's always assigned to the 10.77.77.77 in the 10.77.77.0/24 subnet. So the host gets the .1.

PR #6038, opened 2018-05-25
Do not call updaters and deployers when run with --dry-run

### LINKED ISSUE #5927: Updaters run when --dry-run is set
Both the generic updater and the renew deployer run when `--dry-run` is set on the command line. We shouldn't do this as a dry run shouldn't be making any (permanent) changes to the user's config.

joohoi:
When Certbot is run with `--dry-run`, skip running `GenericUpdater` and `RenewDeployer` interface methods.

This PR also makes the parameter order of `updater.run_generic_updaters` and `updater.run_renewal_deployer` consistent.

Fixes #5927

### DIFF
--- a/certbot/tests/main_test.py
+++ b/certbot/tests/main_test.py
@@ -1464,7 +1464,8 @@ class MainTest(test_util.ConfigTestCase):  # pylint: disable=too-many-public-met
                           None, None, None)
 
         with mock.patch('certbot.updater.logger.warning') as mock_log:
-            updater.run_generic_updaters(None, None, None)
+            mock_config = mock.MagicMock(dry_run=False)
+            updater.run_generic_updaters(mock_config, None, None)
             self.assertTrue(mock_log.called)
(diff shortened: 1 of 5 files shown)


certbot/tests/main_test.py
```diff
@@ -1464,7 +1464,8 @@ def test_plugin_selection_error(self, mock_choose):
                           None, None, None)
 
         with mock.patch('certbot.updater.logger.warning') as mock_log:
-            updater.run_generic_updaters(None, None, None)
+            mock_config = mock.MagicMock(dry_run=False)
+            updater.run_generic_updaters(mock_config, None, None)
```
bmw: nit: We should use `self.config` here which is already set up with default config values including `dry_run` equal to `False`.

joohoi:
The `ConfigTestCase` is now used. This is not the first time I failed to use it initially :)

bmw:
Whoops. Looks like there are merge conflicts with your other PR that I just merged.

Issue #6045, opened 2018-05-28
certbot-apache should dump the apachectl -S when it can't renew


jsoref:
If you're having trouble using Certbot and aren't sure you've found a bug or
request for a new feature, please first try asking for help at
https://community.letsencrypt.org/. There is a much larger community there of
people familiar with the project who will be able to more quickly answer your
questions.

## My operating system is (include version):
Red Hat Enterprise Linux Server release 6.9 (Santiago)
Python 2.7 (redhat scl python27)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto (0.24.0)

## I ran this command and it produced this output:
/root/bin/certbot-auto -q renew --post-hook /root/bin/certbot-renew-hook

FailedChallenges: Failed authorization procedure. site.example.com (tls-sni-01): urn:acme:error:unauthorized :: The client lacks sufficient authorization :: Incorrect validation certificate for tls-sni-01 challenge. Requested whatever.acme.invalid from IP.AD.DR.ESS:443. Received 2 certificate(s), first certificate had names "other.example.com"

## Certbot's behavior differed from what I expected because:
It could have included `apachectl -S` output (which would help debugging)

###### Logs are stored in `/var/log/letsencrypt` by default. Feel free to redact domains, e-mail and IP addresses as you see fit.

```
FailedChallenges: Failed authorization procedure. site.example.com (tls-sni-01): urn:acme:error:unauthorized :: The client lacks sufficient authorization :: Incorrect validation certificate for tls-sni-01 challenge. Requested whatever.acme.invalid from IP.AD.DR.ESS:443. Received 2 certificate(s), first certificate had names "other.example.com"
```

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:

Fwiw, here's my normal -S output (minus a number of other hosts that aren't implicated by the output from certbot:
```
127.0.0.1:443          is a NameVirtualHost
         default server other.example.com (/etc/httpd/conf.d/001-other-ssl-localhost.conf:1)
         port 443 namevhost site.example.com (/etc/httpd/conf.d/103-site-ssl-localhost.conf:1)
wildcard NameVirtualHosts and _default_ servers:
*:443                  is a NameVirtualHost
         default server other.example.com (/etc/httpd/conf.d/001-other-ssl.conf:1)
         port 443 namevhost site.example.com (/etc/httpd/conf.d/003-site-ssl.conf:1)
[...]

PR #6048, opened 2018-05-30
Hint about --debug-challenges (#6045)

jsoref:


schoen:
I'm a little nervous about whether this is the right option for most people because they won't necessarily know how to check whether the challenge was, in fact, set up properly (and they also might end up quitting Certbot while the challenge is still set up). Does anyone have any recent experience that bears on this issue?

I might prefer referring people to something like https://letsdebug.net/ because it will perform more concrete "debugging" in comparison to the `--debug-challenges` option, which is more suited for people who are prepared to debug their own challenges.

schoen:
I should also have said: thanks for the PR, @jsoref!

jsoref:
Sorry, in my case, that's useless.
> ValidationMethodDisabled
> FATAL
> The validation method provided (tls-sni-01) has been disabled by Let's Encrypt. For more information, please visit the url in the details.
> https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811 

> ValidationMethodDisabled
> FATAL
> The validation method provided (tls-sni-02) has been disabled by Let's Encrypt. For more information, please visit the url in the details.
> https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811 

As is, I'd be absolutely opposed to suggesting that service. `--debug-challenges` _might_ work, whereas, it's clear that https://letsdebug.net/ won't.

Issue #6046, opened 2018-05-28
apache certbot configured a challenge to 127.0.0.1:443


jsoref:
## My operating system is (include version):
Red Hat Enterprise Linux Server release 6.9 (Santiago)
Python 2.7 (redhat scl python27)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto (0.24.0)

## I ran this command and it produced this output:
```sh
/root/bin/certbot-auto --no-self-upgrade renew --debug-challenges
```
```sh
apachectl -S
Warning: DocumentRoot [/var/lib/letsencrypt/tls_sni_01_page/] does not exist
VirtualHost configuration:
...
127.0.0.1:443          is a NameVirtualHost
         port 443 namevhost site.example.com (/etc/httpd/conf.d/103-site-ssl-localhost.conf:1)
         port 443 namevhost XXXX.YYYY.acme.invalid (/etc/httpd/conf.d/le_tls_sni_01_cert_challenge.conf:2)
wildcard NameVirtualHosts and _default_ servers:
*:443                  is a NameVirtualHost
         default server other.example.com (/etc/httpd/conf.d/001-other-ssl.conf:1)
         port 443 namevhost site.example.com (/etc/httpd/conf.d/003-site-ssl.conf:1)
         port 443 namevhost site.example.com (/etc/httpd/conf.d/103-site-ssl.conf:1)
*:80                   is a NameVirtualHost
         default server other.example.com (/etc/httpd/conf.d/001-other.conf:1)
         port 80 namevhost site.example.com (/etc/httpd/conf.d/001-site.conf:1)
```
```sh
cat /etc/httpd/conf.d/le_tls_sni_01_cert_challenge.conf
<IfModule mod_ssl.c>
<VirtualHost 127.0.0.1:443>
    ServerName XXXX.YYYY.acme.invalid
    UseCanonicalName on
    SSLStrictSNIVHostCheck on

    LimitRequestBody 1048576

    Include /etc/letsencrypt/options-ssl-apache.conf
    SSLCertificateFile /var/lib/letsencrypt/ZZZZ.crt
    SSLCertificateKeyFile /var/lib/letsencrypt/ZZZZ.pem

    DocumentRoot /var/lib/letsencrypt/tls_sni_01_page/
</VirtualHost>

</IfModule>
```

[...]

PR #6052, opened 2018-05-31
Apache: list everywhere for challenges (#6046)

jsoref:
A site can have multiple hosts sharing a certificate, e.g.:
```
<VirtualHost localhost:443>
ServerName x.example.com
</VirtualHost>
<VirtualHost x.example.com:443>
ServerName x.example.com
</VirtualHost>
<VirtualHost 10.0.0.1:443>
ServerName x.example.com
</VirtualHost>
```

The goal of the SNI challenge is to answer requests for specially defined servernames.
We really do not care how they are delivered to apache, as long as we can get apache to
return our specially crafted certificate.

As such, the simplest approach seems to be to just use *.

There are some potentially bizarre ways for traffic to actually come into localhost
(I actually have done this for other systems), but in general, it is more likely that
it will come in via some other address. That said, there does not appear to be any
harm/risk/reason not to just listen for any possible entrypoint.

It is definitely better to listen to all possible entrypoints than it is to only listen
to localhost, which will on average never be able to answer the request.

---
I'm having trouble figuring out if this works ...

### DIFF
--- a/certbot-apache/certbot_apache/tls_sni_01.py
+++ b/certbot-apache/certbot_apache/tls_sni_01.py
@@ -124,8 +124,11 @@ class ApacheTlsSni01(common.TLSSNI01):
             self.configurator.config.tls_sni_01_port)))
 
         try:
-            vhost = self.configurator.choose_vhost(achall.domain,
-                                                   create_if_no_ssl=False)
+            try:
+                vhosts = self.configurator.vhosts
+            except AttributeError:
+                vhost = self.configurator.choose_vhost(achall.domain,
+                                                       create_if_no_ssl=False)
         except (PluginError, MissingCommandlineFlag):
             # We couldn't find the virtualhost for this domain, possibly
             # because it's a new vhost that's not configured yet
@@ -134,13 +137,14 @@ class ApacheTlsSni01(common.TLSSNI01):
             addrs.add(default_addr)
             return addrs
 
-        for addr in vhost.addrs:
-            if "_default_" == addr.get_addr():
-                addrs.add(default_addr)
-            else:
-                addrs.add(
-                    addr.get_sni_addr(
-                        self.configurator.config.tls_sni_01_port))
+        for vhost in vhosts:
+            for addr in vhost.addrs:
+                if "_default_" == addr.get_addr():
+                    addrs.add(default_addr)
+                else:
+                    addrs.add(
+                        addr.get_sni_addr(
+                            self.configurator.config.tls_sni_01_port))
 
         return addrs
 


jsoref:
For reference, here's what the log file has to say for this version (which succeeded!):

```
2018-06-01 02:20:37,806:DEBUG:certbot_apache.parser:Adding Include /etc/httpd/conf.d/le_tls_sni_01_cert_challenge.conf to /files/etc/httpd/conf/httpd.conf
2018-06-01 02:20:37,807:DEBUG:certbot_apache.tls_sni_01:writing a config file with text:
 <IfModule mod_ssl.c>
<VirtualHost a.b.example.com:443 127.0.0.1:443 a.c.example.com:443 *:443 a.d.example.com:443 a.e.example.com:443>
    ServerName XXXX.YYYY.acme.invalid
    UseCanonicalName on
    SSLStrictSNIVHostCheck on

    LimitRequestBody 1048576

    Include /etc/letsencrypt/options-ssl-apache.conf
    SSLCertificateFile /var/lib/letsencrypt/ZZZZ.crt
    SSLCertificateKeyFile /var/lib/letsencrypt/ZZZZ.pem

    DocumentRoot /var/lib/letsencrypt/tls_sni_01_page/
</VirtualHost>

</IfModule>
```

I'm aware that what I'm doing to get the domains may not be valid -- I think it's possible that I'm crossing object boundaries. The code would be no worse than today (as there's a try fallback). Given that this code is a dead end, I don't think we want to make huge infrastructure changes to handle it... But, if there's a more correct way to do what I'm trying to do, I'm happy to do it.

jsoref:
@joohoi: I believe round 2 does what you've described. I haven't updated the tests, but I could, assuming that this is in the right spirit.

joohoi:
Your approach is correct, but I still wish to reiterate, that the whole TLS-SNI challenge type is deprecated, and the complete challenge type was removed from the ACME protocol RFC draft a few iterations ago after it was discovered that some of the larger hosting providers handled SNI in a way that could result in to issuing certificates for entities that shouldn't be able to validate the challenges.

This means that the code you are working on is effectively deprecated, and no new certificates can be issued using it. Apache plugin is using HTTP challenge type instead. The TLS-SNI code still exists in Certbot because Let's Encrypt has this challenge type on sunset period, allowing users that initially used this challenge type are able to renew their existing certificates.

I don't want you to go through the effort of working on something that will become dead code soon.

joohoi:
RFC references:
- [ACME draft 9](https://tools.ietf.org/html/draft-ietf-acme-acme-09#section-8.4) still had the TLS-SNI challenge.
- It got removed in [draft 10](
https://tools.ietf.org/html/draft-ietf-acme-acme-10) posted March 05, 2018

sydneyli:
Closing this PR since TLS-SNI has been deprecated. Feel free to open another PR if you'd like to fix #6046  for the HTTP challenge.

PR #6059, opened 2018-06-01
WIP External Account Binding

robaman:


robaman:
Thanks a lot for your comments @bmw. Will start work in this now.

robaman:
I've refactored a lot of the EAB functionality into an ExternalAccountBinding class now @bmw .
Thought about if I should send in the whole config instance into the from_data method, but decided to send in the more specific data from config inside perform_registration method instead. 

What do you think about that? And of course the general design approach.

Also been thinking about your comment about only using key.public_key().
I am only sending in the public_key() part of the key from perform_registratration method now.
Further, do you mean to skip the JSON encoding inside of the External Account Binding class and let josepy handle it implicitly in some way?

Was this what you meant?

Thanks!

seandilda:
What's the current status on this PR?  I'd love to see this functionality in certbot.

Thanks!

ohemorange:
This PR has been moved to my queue, planning to take a look.

certbot/log.py
```diff
@@ -313,6 +313,7 @@ def post_arg_parse_except_hook(exc_type, exc_value, trace, debug, log_path):
     :param str log_path: path to file or directory containing the log
 
     """
+    debug = True
```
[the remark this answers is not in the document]
robaman: Removed it.

acme/acme/messages.py
```diff
@@ -258,6 +259,27 @@ class ResourceBody(jose.JSONObjectWithFields):
...
 
 
+class ExternalAccountBinding:
+    """ACME External Account Binding"""
+
+    @classmethod
+    def from_data(cls, key=None, kid=None, hmac=None):
+        """Create External Account Binding Resource from contact details, kid and hmac."""
+
+        key_json = json.dumps(key.to_partial_json())
```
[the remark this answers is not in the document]
robaman: Not sure I understand here. Doesn't the payload argument to JWS.sign() have to be a string?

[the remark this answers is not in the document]
ohemorange: The conversion to a string is handled automagically by different classes in josepy. The same `json.dumps(x.to_partial_json())` is being done behind the scenes.

Thanks to @bmw for doing the digging here!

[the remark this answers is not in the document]
bmw: Sorry. I wasn't very clear in my explanation to @ohemorange here. I think this line without changes is correct.

For context, I was initially unsure because we pass this same value in the `jwk` header all the time and don't have any code like this in acme/certbot, but that's because we have a special [Header class in josepy](https://github.com/certbot/josepy/blob/v1.1.0/src/josepy/jws.py#L38) that handles this for us. Passing this value in the `payload` is a one time thing and ACME specific rather than being generic to JOSE, so I think this approach works just fine.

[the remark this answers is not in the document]
robaman: Ok, thanks guys! Resolving this one now then.

acme/acme/messages.py
```diff
@@ -258,6 +259,27 @@ class ResourceBody(jose.JSONObjectWithFields):
...
+        decoded_hmac = jose.b64.b64decode(hmac)
+        jws = jose.JWS.sign(
+            payload=key_json,
+            key=jose.jwk.JWKOct(key=decoded_hmac),
+            alg=jose.jwa.HS256,
+            kid=kid,
+            include_jwk=False,
+            protect=frozenset(['alg', 'kid'])
```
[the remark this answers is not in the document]
robaman: I've now implemented the functionality that makes Certbot error out if the server requires external account binding and the user didn't specify the kid and hmac key on command line. It's in commit: https://github.com/certbot/certbot/pull/6059/commits/854804f87d10cb6d35a7b29c1942f2346cffd6f0

By raising errors.Error in the code it seems that the exception only gets logged in the logfile and not outputted to stdout. Instead the generic error message "Unable to register an account with ACME server". This is the code: https://github.com/certbot/certbot/blob/36ebce4a5fa30a0e73bd8656aea54ded2e9d982e/certbot/main.py#L527

Would this be in line with the Certbot logging philosophy? Or do we need to create a separate exception for this error to make the real error message show up on stdout?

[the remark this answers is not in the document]
ohemorange: To translate: that assertion is saying that there should be a line in https://github.com/certbot/josepy/blob/b0ad174768af90efbf6c98e9d68fa201108e9386/src/josepy/jws.py#L38 that says something like `url = json_util.Field('url', omitempty=True)`

What is a mystery to me is how the other code you reference is working at all without that line... any thoughts?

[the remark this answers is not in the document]
ohemorange: ah yes, there we go

`Header` is subclassed here: https://github.com/certbot/certbot/blob/master/acme/acme/jws.py#L10

instead of using `jose.JWS` directly, do what `client.py` does and `from acme import jws` and then use `jws.JWS`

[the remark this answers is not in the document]
robaman: Thanks a lot! Will look into this now.

acme/acme/messages.py
```diff
@@ -274,19 +296,27 @@ class Registration(ResourceBody):
...
+                  directory=None, **kwargs):
         """Create registration resource from contact details."""
         details = list(kwargs.pop('contact', ()))
         if phone is not None:
             details.append(cls.phone_prefix + phone)
         if email is not None:
             details.append(cls.email_prefix + email)
         kwargs['contact'] = tuple(details)
+
+        if kid is not None and hmac_key is not None:
```
ohemorange: You're totally right about the directory having to be passed in to construct the object. Since the `ExternalAccountBinding` object needs so many parameters that are only for it, let's move the `from_data` call into `client.py` so that we only have to pass one object that might be `None`, instead of four separate things. Then we could clean up that `from_data` call signature.

robaman: Just pushed a change to PR: https://github.com/certbot/certbot/pull/6059/commits/8963fae5dcf99b96c65c2d3d5ea7b1f9a403138a

This what you meant?

robaman: Didn't do cleanup of old stuff. But will do it if this is what we're after.

ohemorange: Should have been more specific, when I said `client.py` above I mean `certbot/certbot/client.py`. The paradigm we have here is mostly constructing objects in the client code or using `from_data` methods, rather than a `build_*` paradigm. So in `certbot/certbot/client.py:perform_registration`, that would look something like:

```
if conditions_checking_if_appropriate_to_construct_eab:
    eab = ExternalAccountBinding.from_data(account_public_key, kid, hmac_key, directory)
else:
    eab = None
try:
    acme.new_account_and_tos(
        messages.NewRegistration.from_data(email=config.email, external_account_binding=eab),
        tos_cb)
```

robaman: Aah, will fix!

robaman: New version pushed.

certbot/client.py
```diff
@@ -196,13 +196,17 @@ def perform_registration(acme, config, tos_cb):
         if config.eab_kid is None or config.eab_hmac_key is None:
             raise errors.Error("Server requires external account binding. Please use --eab-kid and --eab-hmac-key.")
 
-    account_public_key = acme.client.net.key.public_key()
+        account_public_key = acme.client.net.key.public_key()
+        eab = messages.ExternalAccountBinding.from_data(account_public_key=account_public_key,
```
ohemorange: This should probably be outside the "is eab required by the server" if, and inside a separate "are the necessary things available" if.

robaman: Just pushed an updated version with two ifs. Is this what you had in mind?

ohemorange: yes, very nice!

robaman: Great, will proceed with tests and docs.

ohemorange:
Hey, running into any snags that I could help you work through while writing tests and docs?

robaman:
I broke a couple of tests with my EAB changes. So I'm trying to find a  way to get them to pass now.

This is what I've come up with so far, after a lot of wrestling with pythons mock-lib. :)
After I've got that in place it now fails with:

../../acme/acme/messages.py:275: in from_data
    decoded_hmac_key = jose.b64.b64decode(hmac_key)
>           raise TypeError('argument should be a str or unicode')
E           TypeError: argument should be a str or unicode
../../venv/lib/python2.7/site-packages/josepy/b64.py:56: TypeError

at https://github.com/robaman/certbot/blob/fdf9376ae5d2485e0692eec80ab2fabf77e3dd92/acme/acme/messages.py#L273

Right now  I'm thinking about hardcodeding an encoded hmac key in the tests.


https://gist.github.com/robaman/2ea0211ce0725ec1824e27ac59095086

robaman:
All right!

Got the tests to pass with these changes:
https://github.com/certbot/certbot/pull/6059/commits/6d92f445e354f3104008cc87b5b72693ed449b69

Comments?

robaman:
Also just committed a draft of some testing of messages.py.
What do you think?

acme/acme/messages.py
```diff
@@ -258,6 +262,24 @@ class ResourceBody(jose.JSONObjectWithFields):
...
 
 
+class ExternalAccountBinding(object):
+    """ACME External Account Binding"""
+
+    @classmethod
+    def from_data(cls, account_public_key, kid, hmac_key, directory):
+        """Create External Account Binding Resource from contact details, kid and hmac."""
+
+        key_json = json.dumps(account_public_key.to_partial_json())
```
ohemorange: this needs a `.encode()`

certbot/tests/client_test.py
```diff
@@ -25,15 +27,29 @@ def setUp(self):
         self.config.register_unsafely_without_email = False
         self.config.email = "alias@example.com"
         self.account_storage = account.AccountMemoryStorage()
+        self.config.eab_hmac_key = "hmac-key-for-testing"
```
ohemorange: We want these to sometimes not be available, so they should be set in `config` only when explicitly testing that an EAB is constructed.

certbot/tests/client_test.py
```diff
@@ -25,15 +27,29 @@ def setUp(self):
...
+        return m
+
+    @staticmethod
+    def _directory_getitem_mock(input):
+        return "/acme/new-account"
+
     def test_no_tos(self):
         with mock.patch("certbot.client.acme_client.BackwardsCompatibleClientV2") as mock_client:
             mock_client.new_account_and_tos().terms_of_service = "http://tos"
+            mock_client().client.net.key.public_key = mock.Mock(side_effect=self._public_key_mock)
```
ohemorange: Why do we now need these? Can we find a way to test, such that we don't?

acme/acme/messages.py
robaman: Pushed.

certbot/tests/client_test.py
robaman: I set them to None in certbot.constants.CLI_DEFAULTS instead now. Seems to have solved the reason I originally set them. 
Still working through some test logic problems in certbot/client.py that was due to a Mock always returning True: "if acme.client.directory.meta.external_account_required".

certbot/tests/client_test.py
robaman: The reason I put it there, after some discussions with @joohoi and @adferrand on #letsencrypt-dev, is because of:

https://github.com/robaman/certbot/blob/b3e75257a7bc2ea53983efb60ee9baae1ca74bc4/certbot/client.py#L209

Without it, it returns a mock that does not support to_partial_json() and need to be an interfaces.JSONDeSerializable because of a isinstance() check.

The mock below it is because of:
https://github.com/robaman/certbot/blob/b3e75257a7bc2ea53983efb60ee9baae1ca74bc4/acme/acme/messages.py#L274

...not returning an URL correctly.



I've been thinking about other approaches to do it, but I can't find out another way.

Do you have any pointers for me?

ohemorange: The easiest way is to not have `config.eab_kid and config.eab_hmac_key` exist, by removing them from `setUp` and making sure that if `config` is mocked, they return `None`, again in `setUp`. And then setting them to have values explicitly in functions that are testing EAB.

robaman: Hmm. Was this answer supposed to be on another thread? Where you started saying "We want these to sometimes not be available, so they should be set in config only when explicitly testing that an EAB is constructed." perhaps?

I still haven't figured out how to proceed with the mock stuff.

ohemorange: It's relevant to both -- if you do that, it should also fix this problem, because it won't hit the EAB stuff in the first place.

robaman: Think I understood now. :)

adferrand:
@robaman I see you are struggling about using the new code you have in acme project into certbot project. What you need to modify is `local-oldest-requirement` at the git repository root, in order to run tests. Set it to:
```
-e acme[dev]
```

To use the local project instead of trying to get it from pip, and benefit from code of your current PR.

And also move declaration dependency of `acme` in the `setup.py` of `certbot` project to the next stable version, `0.29.0`. It is not necessary for the CI itself, but for the deployment: indeed at this time it will be needed that `certbot` distribution requires the latest `acme` distribution. Better to do it as soon as it is necessary.

acme/acme/client.py
```diff
@@ -874,6 +874,12 @@ def revoke(self, cert, rsn):
         """
         return self.client.revoke(cert, rsn)
 
+    def external_account_required(self):
```
ohemorange: EAB is acmev2-only, right? If so, this should really be on `ClientV2`.

ohemorange: The tests should still work.

robaman: Yep, the EAB feature is V2 only.
But I was thinking that this code line[0] will be executed regardless of, if the server is a ACMEv1 or ACMEv2 server. 
Am I misunderstanding something?

[0]. https://github.com/robaman/certbot/blob/8adf23283c77dca2b3ecb7b2b5f9d0cc58c6a57d/certbot/client.py#L217

ohemorange: ah, good point. so the way the code's set up here is that the `Backwards...` class just translates between v1 and v2 styles, and calls into the appropriate method. I know that seems like a lot of structure for what's only a few lines, but the idea is that some day v1 will be deprecated and `ClientV2` will be a perfect replacement for `Backwards...`. Also, the ACME library is used on its own by external clients, and we want those clients to be able to use the `ClientV2` class directly.

robaman: Working on moving this into ClientV2.

ohemorange: You're totally right, that was a previous version that just automatically passed things through. Now, it's probably best to make it so it's both in ClientV2 and explicitly passed through in BackwardsCompatible, with just returning False if we're acmev1 to avoid complications.

robaman: Fixed now.

robaman:
Now everything seems to be in order but the linting error:
************* Module acme.client_test
C:  1, 0: Too many lines in module (1280/1250) (too-many-lines)

Is the only way to refactor the tests here? Or is there another way around this? 
Like up'ing the line-limit of client_test.py.

Also, do you think we need more docs than the docstrings I added in the code?

adferrand:
I hit the limit on this test module also with POST-as-GET PR. I said to pylint to ignore it, with `pylint: disable=too-many-lines` after module docstring.

Other test modules have also this rule ignored, I think it is acceptable in a test module, because of the specific verbosity required to write unit tests, and as it does not harm the code quality of Certbot itself.

robaman:
Thanks! I’ll try that.

robaman:
All tests pass.

local-oldest-requirements.txt
```diff
@@ -1 +1 @@
-acme[dev]==0.26.0
+-e acme[dev]
```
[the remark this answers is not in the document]
robaman: Fixed.

acme/setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools.command.test import test as TestCommand
 import sys
 
-version = '0.29.0.dev0'
+version = '0.29.0'
```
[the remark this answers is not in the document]
robaman: Fixed.

certbot/client.py
```diff
@@ -4,7 +4,6 @@
 import os
 import platform
 
-
```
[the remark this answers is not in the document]
robaman: Done!

robaman:
Change log updated. I can see the finish line. :)

setup.py
```diff
@@ -31,7 +31,7 @@ def read_file(filename, encoding='utf8'):
 # specified here to avoid masking the more specific request requirements in
 # acme. See https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
-    'acme>=0.26.0',
+    'acme>=0.29.0',
```
ohemorange: Actually sorry, the real correct way to fix this is to change it to 0.29.0.dev0, and the dev0 will get automatically removed at release.
```suggestion
    'acme>=0.29.0.dev0',
```

local-oldest-requirements.txt
```diff
@@ -1 +1 @@
-acme[dev]==0.26.0
+acme[dev]==0.29.0
```
ohemorange: Ok FINAL ANSWER really this time: it DOES have to be `-e acme[dev]` and then changed back after the release. I know, I know.
```suggestion
-e acme[dev]
```

robaman: :)

robaman:
Yiha! Thanks a lot for your help and patience @ohemorange @bmw @adferrand !

robaman:
Btw, will this be included in the 0.29 release?

adferrand:
Well, it reached master before the freeze on Wednesday, so... yes ^^

robaman:
And thanks a lot to you @cpu !  :)

clintwilson:
Awesome! Thank you!

ohemorange:
@themsaid how are you running certbot? easiest thing to do is probably to `rm -rf` the `venv` and rerun `tools/venv.py`.

PR #6070, opened 2018-06-04
Fix Sphinx

### LINKED ISSUE #4686: Certbot docs don't build in Sphinx 1.6.1+
In their 1.6.1 release, Sphinx started requiring the package `latexmk` when trying to make PDFs of docs using the Makefile target `latexpdf`. This would be fine, but it also came with a change in behavior for us.

Apparently, Sphinx does not support SVGs when building PDFs (see sphinx-doc/sphinx#1907). Previously, this would be silent error and our documentation would just build without the SVG images. Now it's a hard error and causes the build to fail.

While it's nice that we're now aware of the problem, it has broken our tests on the https://github.com/certbot/website repo. I'm going to pin `Sphinx<=1.5.6` until this issue has either been resolved by us or upstream. This problem only affects the Certbot package and the problem is the badges in the Certbot README.

bmw:
Fixes #4686.

In Sphinx 1.6, they changed how they handle images in latex and PDF files. You can learn more about this by reading the linked issue (or I can answer any questions), but the shortish version is we now need to use the extension `sphinx.ext.imgconverter`. This is only available in Sphinx 1.6+.

I also updated our pinned versions to use the latest Sphinx and a new dependency it pulled in called `sphinxcontrib-websupport`. To build the latex and PDF docs, you must first run:
```
apt-get install imagemagick latexmk texlive texlive-latex-extra
```
Afterwards, if you create the normal Certbot dev environment using this branch, activate the virtual environment, and from the root of the repo run `make -C docs clean latex latexpdf`, you'll successfully build the PDF docs.

@sydneyli, this is low priority, but do you want to review this when you get the chance?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,6 +40,7 @@ needs_sphinx = '1.0'
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx.ext.imgconverter',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
--- a/setup.py
+++ b/setup.py
@@ -71,8 +71,8 @@ dev3_extras = [
 
 docs_extras = [
     'repoze.sphinx.autointerface',
-    # autodoc_member_order = 'bysource', autodoc_default_flags, and #4686
-    'Sphinx >=1.0,<=1.5.6',
+    # autodoc_member_order = 'bysource', autodoc_default_flags
+    'Sphinx >=1.0',
     'sphinx_rtd_theme',
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -71,8 +71,8 @@ dev3_extras = [
 
 docs_extras = [
     'repoze.sphinx.autointerface',
-    # autodoc_member_order = 'bysource', autodoc_default_flags
-    'Sphinx >=1.0',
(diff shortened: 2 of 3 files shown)


hlieberman:
@bmw is this dep required just for the PDF generation, or is it also needed for the manpage / html build?

bmw:
@hlieberman, it's not needed for the manpage or html docs.

It is needed for the PDF and might also be needed for LaTeX, but I can't remember without testing. If this matters for you, let me know and I'll look into it.

sydneyli:
@bmw what is the `Sphinx==1.7.5` constraint for?

bmw:
> @bmw what is the Sphinx==1.7.5 constraint for?

We pin all our dependencies in our tests and development environment for increased stability. This avoids Certbot development grinding to a halt if one of our dependencies uploads a bad package or one that no longer works with Certbot. This is a bad experience for open source contributors for instance.

We test with the latest version of all dependencies during releases, but want to set this test up to run nightly. I thought there was an issue for that, but I couldn't find it so I made #6283.

`Sphinx==1.7.5` was just the version I used while working on this and is known to work.

(Closing and reopening to get codecov to pass.)

bmw:
I created #6288 to track better documenting these dependencies.

PR #6078, opened 2018-06-06
Upgrade pinned twine version

bmw:
For the past couple of releases, twine has errored while trying to upload
packages and this is fixed by upgrading to a newer version of twine. This
commit updates our pinned version installed when using tools/venv.sh to the
latest available version. pkginfo had to be upgraded as well to support the
latest version of twine.

I quickly tried to find the fix for this in `twine`'s changelog but failed. Regardless, upgrading has fixed my problem twice now and upgrading our development environment pinnings shouldn't hurt anything.

### DIFF
--- a/tools/dev_constraints.txt
+++ b/tools/dev_constraints.txt
@@ -36,7 +36,7 @@ oauth2client==2.0.0
 pathlib2==2.3.0
 pexpect==4.2.1
 pickleshare==0.7.4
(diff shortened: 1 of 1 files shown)


bmw:
That's true. Other than knowing from experience that running `pip install -U twine` has fixed problems multiple times in the past, I tested that a newly created venv uses these versions.

Merging. Thanks for the review.

PR #6082, opened 2018-06-06
Unrevert #6000 and silence deprecation warnings

bmw:
This silences the deprecation warnings from cryptography. I looked into only
silencing the cryptography warning specifically in the function, however,
CryptographyDeprecationWarning doesn't seem to be publicly documented, so we
probably shouldn't depend on it. See https://cryptography.io/en/latest/search/?q=CryptographyDeprecationWarning&check_keywords=yes&area=default.


bmw:
While there's certainly some benefit to logging any warnings, I'm not sure we want to do that here. The cryptography maintainer told me that we don't have to worry about them removing this functionality anytime soon and I worry that logging it creates visual noise that can confuse users thinking there is (or is about to be) a problem with their installed versions.

joohoi:
Perfectly reasonable, LGTM

PR #6089, opened 2018-06-11
Wrap TLS-ALPN extension with ASN.1

rolandshoemaker:
Instead of just using the raw value of the hash. Since the tag and OCTET STRING length will never change just prepend the tag and length bytes instead of using an ASN.1 library to do this work.


rolandshoemaker:
Note: adding the prefix could be done at `TLSALPN01Response.h` (or renamed to something that doesn't imply the hash) but I wasn't really sure what the common approach would be. Happy to do either.

rolandshoemaker:
Whoops, sorry about doing that twice, rebasing is hard :|

Issue #5824, opened 2018-04-04
Offer interactive selection for --cert-name for install verb


joohoi:
If `--cert-name` or key and cert paths were not defined on the command line, Certbot should offer interactive selection of certificate, as implemented in enhance verb PR #5596

sbraden:
I'm going to work on this one.

bmw:
Looks like tests for `install` start at https://github.com/certbot/certbot/blob/802fcc99ee2150989e5a1e860b6aad490691d72f/certbot/tests/main_test.py#L699 and there's a few consecutive ones. They probably should be written in a more easily findable way and not in one giant test class.

Trying to base a test for this off of these tests and some of the enhance tests which you mentioned in https://github.com/certbot/certbot/pull/5596/files#diff-8d7d4822efd312492f20da699e3b062a is a good idea I think.

bmw:
@sbraden, are you still planning on working on this or should one of us take over? If the latter, do you have a start on this somewhere we can build off of?

PR #6097, opened 2018-06-12
Interactive certificate selection with install verb

joohoi:
If either `--cert-name` or both `--key-path` and `--cert-path` (in which case the user requests installation for a certificate not managed by Certbot) are not provided, prompt the user with managed certificates and let them choose.

Fixes: #5824


certbot/main.py
```diff
@@ -790,6 +790,13 @@ def install(config, plugins):
     except errors.PluginSelectionError as e:
         return str(e)
 
+    custom_cert = (config.key_path and config.cert_path)
```
ohemorange: Should also change line 805 to match

joohoi: The current implementation on line 805 catches both of the cases - user providing both `--key-path` and `--cert-path` and configuration object values being populated from either interactively selected or user provided `--cert-name`.

Are you suggesting something in line with `if custom_cert or config.certname:` ?

ohemorange: I meant this as a nit -- you already create a variable that captures this boolean, so might as well use `custom_cert` on line 805 in place of `config.key_path and config.cert_path`

joohoi: Ah, right. Maybe I should elaborate a bit; these values get populated from `config.certname` obtained from CLI or by interactive selection in line 804 to normalize the handling from there on, so the initial variable value might not reflect the actual state at this point.

ohemorange:
I'll let you merge because of the enhancements stuff that's now in there

joohoi:
Ready for the next review round, I had to make the config mock more detailed in two of the `main_test.py` tests.

Issue #6098, opened 2018-06-12
Don't use strictly greater than requirements


bmw:
Now that we've broken lockstep, we need a way to specify that Certbot or a plugin relies on unreleased changes in acme or Certbot.

Suppose the latest released version is 0.24.0 and a package relies on unreleased changes in acme. To do this, we've been using version requirements like `acme>0.24.0`. This is because acme `0.25.0.dev0` satisfies this requirement and assuming we don't release `0.24.1` (in which case we'd have to change the requirement to `acme>0.24.1`), the only packages that satisfy this are `acme>=0.25.0`.

This is unconventional though, has caused confusion with multiple package maintainers, and could cause issues in the extremely unlikely event we release a bug fix release for `0.24` after `0.25.0` comes out.

@erikrose, do you have any thoughts on this? Is this something you think is worth changing and any suggestions on how to do it?

erikrose:
I've never seen anyone even try to solve this problem before. Isn't the usual approach just to wait until a new certbot version is released and only then to release the dependent plugin version? Are people using unreleased versions of plugins in great number?

erikrose:
If there is a pressing need to solve this problem—for instance, I could see setuptools getting grumpy during development while scanning for entrypoints—what about having the plugin require `acme>=0.25.0` and just making sure to bump acme's version number on master as soon as we start adding features rather than bugfixes to it?

bmw:
The issue here is only during our development. We're still releasing every Certbot component at once and sometimes one of our plugins is requiring features that have only landed in `master` and haven't been released yet.

> what about having the plugin require acme>=0.25.0 and just making sure to bump acme's version number on master as soon as we start adding features rather than bugfixes to it?

What we currently do is as part of the release process for releasing version X, we bump the version number for every package in `master` to X+1.dev0. So when doing the release for 0.24.0, everything becomes 0.25.0.dev0.

We could remove the `.dev0` and just always have the next version without any indication that it's a dev version. Only potential downsides is sometimes people show up with logs where Certbot is reporting it's `X.dev0` which makes it clear they were running from our tree rather than a stable and tested release.

elyscape:
It looks like pip will properly handle `acme >= 0.25.0.dev0`. Then, as part of the release step, any requirements that specify `.dev0` can have that removed.

erikrose:
Sounds great to me, @elyscape!

bmw:
This has been fixed.

PR #6116, opened 2018-06-15
Remove .dev0 from version numbers during releases.

bmw:
Partial fix for #6098. Changing existing requirements to `>=` rather than `>` is coming in another PR.

@erikrose, do you feel comfortable and able to review this?

This allows us to depend on packages like acme>=0.26.0.dev0 during development and automatically change it to acme>=0.26.0 during the release. We use `git add -p` to be safe, but if .dev0 is used at all in our released setup.py files, we're probably doing something wrong.

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -87,6 +87,14 @@ if [ "$RELEASE_BRANCH" != "candidate-$version" ] ; then
 fi
 git checkout "$RELEASE_BRANCH"
 
+for pkg_dir in $SUBPKGS_NO_CERTBOT certbot-compatibility-test .
+do
+  sed -i 's/\.dev0//' "$pkg_dir/setup.py"
+done
+# We only add Certbot's setup.py here because the other files are added in the
+# call to SetVersion below.
+git add -p setup.py
+
 SetVersion() {
     ver="$1"
     # bumping Certbot's version number is done differently


elyscape:
Just a reminder to @erikrose that this needs to get merged in addition to #6117.

tools/release.sh
```diff
@@ -87,6 +87,14 @@ if [ "$RELEASE_BRANCH" != "candidate-$version" ] ; then
 fi
 git checkout "$RELEASE_BRANCH"
 
+for pkg_dir in $SUBPKGS_NO_CERTBOT certbot-compatibility-test .
+do
+  sed -i 's/\.dev0//' "$pkg_dir/setup.py"
```
erikrose: You run this only on Linux, with GNU sed, correct? The FreeBSD version on OS X uses `-i` for something else and crashes.

bmw:
Currently, if the released versions of our packages is `X`, the version in master is `X+1.dev0`. This script removes `.dev0` and then bumps the version number and adds `.dev0` back again at [the end](https://github.com/certbot/certbot/blob/master/tools/release.sh#L234).

tools/release.sh
bmw: > You run this only on Linux, with GNU sed, correct? The FreeBSD version on OS X uses -i for something else and crashes.

Yeah. I recently switched to a Mac too so I've been bitten by this, but it's a problem throughout the script that I've worked around it using GNU sed from Homebrew.

elyscape: BSD sed uses `-i` to mean the same thing (edit in-place) but expects that you'll provide it with an argument. The argument, if not the empty string, is appended as an extension to the original filename and the result is created as a backup of the original file. The script is already using `sed -i` elsewhere, though, so I'd say to leave it as is for now and then maybe deal with all the uses in a separate PR.

PR #6119, opened 2018-06-15
fix(display): alternate spaces and dashes

### LINKED ISSUE #6050: Tweak output to not trigger markdown heading status
We have a lot of horizontal lines of dashes in the Certbot output, e.g.

```
1: example.com
2: www.example.com
```

Unfortunately, Markdown interprets any text above one of these lines as a heading, leading to distracting and weird formatting when someone pastes it into the forum:

https://community.letsencrypt.org/t/certbot-apache-succeeded-now-site-times-out/63168/3

Arguably they should wrap things in triple-backticks, but this happens quite a lot: People who need help on the forum often don't take time to learn about Markdown. It would be nice to make a small tweak so that formatting just works correctly. Unfortunately, lines of underscores, plus signs, colons, or equal signs all have the same effect. One simple tweak that works is inserting a space (or multiple spaces) in the line of hyphens:

    - -------------------------------------------------------------------------------

or

    - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

sydneyli:
fixes #6050 

Let me know if this looks ok!

### DIFF
--- a/certbot/display/util.py
+++ b/certbot/display/util.py
@@ -29,8 +29,10 @@ HELP = "help"
 ESC = "esc"
 """Display exit code when the user hits Escape (UNUSED)"""
 
+# Display constants
 SIDE_FRAME = ("- " * 39) + "-"
-
+"""Display boundary (alternates spaces, so when copy-pasted, markdown doesn't interpret
+it as a heading)"""
 
 def _wrap_lines(msg):
     """Format lines nicely to 80 chars.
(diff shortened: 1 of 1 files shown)


certbot/display/util.py
```diff
@@ -29,6 +29,8 @@
 ESC = "esc"
 """Display exit code when the user hits Escape (UNUSED)"""
 
+SIDE_FRAME = ("- " * 39) + "-"
```
bmw: Can we add a comment here explaining why we're using this value?

PR #6125, opened 2018-06-16
dns_route53: Allow to set profile name and AWS API max retries in Route53

arcivanov:


arcivanov:
I'd like this to go into 0.25.2 if possible as well if this is in the charts.

arcivanov:
Py 3.4 failed due to environment and no idea why Lint fails

arcivanov:
@jsha could you review please?

jsha (commented):
For profile-name, I think that might be too much complexity to add for the .ini use case. AWS best practice is to use roles. The .ini file is a handy quick fix for when roles are too much work. If you really need to have multiple .ini files, there's a boto3 environment variable you can use to override the path to your .ini.

For the max_attempts: This is sort of related to https://github.com/certbot/certbot/pull/6071/files. What I'd really like to see is an overall deadline / timeout that can be applied across all the operations: Updating the record, polling for it to be in sync, and then deleting it. What would be really cool is if this could be a cross-plugin property, since many plugins presumably need this sort of behavior. What do you think, @bmw @ohemorange ?

arcivanov:
@jsha Unfortunately, when you have to maintain access to isolated AWS accounts with no trusts established multiple profiles are a must. 

Max attempts is the standards boto3/AWS way of handling API rate limits with progressive random backoff, that makes it transparent to the caller. The default of 3 is often insufficient for access-intensive accounts.

jsha:
When I talk about ".ini", I'm talking about `~/.aws/credentials`. That's how the [boto3 docs](https://boto3.readthedocs.io/en/latest/guide/configuration.html#shared-credentials-file) describe it.

Does setting the `AWS_SHARED_CREDENTIALS_FILE` environment variable work for you instead of a flag?

Speaking of which, I can't find the `boto3-max-attempts` field documented. Could you link to docs about it and see if there's a way to set it via environment variable?

arcivanov:
`AWS_SHARED_CREDENTIALS_FILE` selects the file. It doesn't allow picking a sub-section/profile of the config file. I don't know if requiring each profile to be in a separate file is wise. Profile selection is a standard feature of awscli: https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html

Boto3 uses botocore for comms layer, the max-retries is described here: https://botocore.readthedocs.io/en/latest/reference/config.html

```
A dictionary for retry specific configurations. Valid keys are:

'max_attempts' -- An integer representing the maximum number of retry attempts that will be made on a single request. For example, setting this value to 2 will result in the request being retried at most two times after the initial request. Setting this value to 0 will result in no retries ever being attempted on the initial request. If not provided, the number of retries will default to whatever is modeled, which is typically four retries.
```

jsha:
> I don't know if requiring each profile to be in a separate file is wise.

Can you provide a more detailed objection? We try very hard to keep the complexity of Certbot as low as we can, and adding new flags always increases complexity. If we can meet the use case without adding new flags, I have a strong preference for that.

The retries part seems fine. Thanks for linking to the docs.

There are a number of spurious diffs - added and removed lines unrelated to this change. Could you remove those?

arcivanov:
The more detailed objection is as follows: 

It is a standard feature of AWS clients to store credentials in a single credentials file under various profiles. This is, in fact, the only way reasonable people store configuration on development machines (nobody keeps a credential file per profile, all profiles are aggregated in one file in standard home directory location), and it also happens with automation where a single system juggles multiple isolated accounts that do not trust each other.

Currently the only way to accommodate the lack of profile selection on a system with a single credentials file is to pull profiles into a separate credentials file and either maintain profiles in two places or breaking the utilities that expect profile option and single-location credentials file. The solves this problem without taking anything away from Certbot.

The profile selection is entirely optional. Default behavior (without the option specified) is exactly as before and the presence of the option doesn't interfere with any other behaviors. The option is localized to dns-route53 plugin because it uses boto3 and does not affect certbot globally.

Those are not spurious diffs, those are fixes to the standard formatting violations.

arcivanov:
>  I currently feel that the additional complexity added to Certbot in order to support profiles is unwarranted.
 
I strain to understand the characterization of `additional complexity`. Nobody is *required* to use the option. There are no additional steps that anybody has to follow to achieve the same result they are already achieving. All the functionality will continue working as before. It's an option that is available to people who already need it, and since they need it they understand it, i.e. for such people it's not complexity by definition.

Furthermore, its also not a complexity of implementation, as it's a single argument passed to the boto3 session and literally nothing more.

I'll move the formatting to a separate PR.

arcivanov:
> Complexity is generally used to characterize something with many parts where those parts interact with each other in multiple ways, culminating in a higher order of emergence greater than the sum of its parts. 

There is no higher order emergence. There is no cross interaction between multiple components. Current code executes, effectively with `profile=None`. with this change it will execute with `profile=self.conf('profile-name')`. Nothing more, there are no side-effects of any kind.

jsha:
Thanks! I appreciate your input. I still stand by my experience: Increased number of flags leads to higher maintenance complexity and a wider variety of deployment environments to support. It also adds to the conceptual complexity for a new user. I appreciate the effort you have put into this PR, but we have to keep in mind long-term maintenance burden.

arcivanov:
Very well, thanks for your time.

duckie:
@jsha 

> higher maintenance complexity and a wider variety of deployment environments to support

Huh, yes, that is what is happening when you add features to a product. Certbot is consistently growing with new deployment environements, I fail to see how it is a reason not to implement it. If you dont want additional maintenance, then stop developing the product.

The question is : is the additional maintenance cost worth the benefit ? In that case it definitely looks like a yes to me.

Because without this feature, certbot cannot be properly used when you have several AWS subscriptions, which neither roles or high privileges credentials could fix.

fred9176:
Hello, 

I'm sorry to reopen such an old discussion,  but I have the same request : we need to generate certficates for domains located on multiple AWS accounts (1 certificate contains domains from only one account, no mix, hopefully).

I managed to to tell certbot which account to use as follow : 

```
export AWS_SHARED_CREDENTIALS_FILE=$CONFDIR/route53.cred
export AWS_PROFILE=paris
/opt/apps/certbot/certbot-auto certonly --dns-route53 ....
```
This works fine for creating a new certificate, but I didn't see any mention of the used profile in the renewal config file and I'm afraid that in 90 days, when the certificate will be renewed, certbot will try to connect to the default credential.

Can you confirm this point ? 
I think adding --dns-route53-profile-name and using it this way could be a good alternative.

Thanks for your help,

Fred

patricklucas:
I stumbled across this and was also surprised to find out that certbot doesn't support specifying a profile, which is the preferred way of using specific credentials using Boto. An IAM role of course makes sense if the server is running in EC2, but I'm sure there are many users of Route 53 that aren't using EC2 simply because it's a convenient DNS provider.

GeertHauwaerts:
I concur, I have multiple AWS accounts I need to manage through route53. This plugin is unusable without support to select the a profile for a domain.

speller:
I've been hit by the same issue. Being able to specify the profile name is essential. I'm quite surprised by the rejection reason...

speller:
@jsha @certbotbot is the maintenance complexity reason still valid here? Many users are hit by this issue and have to invent scripting workarounds, increasing their infrastructure complexity and decreasing the stability of their systems. Is it possible to implement this very useful feature eventually?

> it is not a large burden for Certbot users to manage those credentials in a separate file

Unfortunately, it **IS** a large burden to make certificates from different AWS accounts to be automatically renewed by `certbot renew`. Currently, users can't use `certbot renew` at all and have to invent dirty workarounds that substitute specific environment variables depending on the certificate name. The proposed simple addition will solve this trouble like a charm.

ohemorange:
I suspect the upcoming `dns-persist` option will happily eliminate all of these woes! You set up the DNS record once, and then there's no need to touch them again, certbot will just renew. You can track the progress here: https://github.com/certbot/certbot/pull/10633

speller:
@ohemorange sounds good, thank you for the reference. BTW, are there any expectations of when it'll be released?

ohemorange:
we're targeting by the 6/2 release or earlier, but are not rushing to make sure we get it right.

speller:
@ohemorange thank you for the prompt feedback. I'm glad to know the work is in progress and is planned to be released soon.

PR #6135, opened 2018-06-21
Add certbot-dns-godaddy

Shinrai:
Similar to https://github.com/certbot/certbot/pull/4748 But updated for the current master


schoen:
Thanks @Shinrai, is this building on #4748 or starting over from scratch?

@zjs, would you have time to take a look at this?

sydneyli:
GoDaddy as a provider is actually now supported by Lexicon-- it may be worth porting this to use lexicon.

adferrand:
@sydneyli, @Shinrai I coded the godaddy provider in Lexicon. Do not hesitate to ping me if you need something.

m0namon:
Just a note that we're not accepting DNS plugins at the moment -- we want to make a decision on #6464, #6503, and #6504 first.

I'm going to close this PR since I think if we did want one, we'd probably prefer a version based on Lexicon. But realistically, we probably won't be able to give any feedback on it for a while. In the meantime, you are welcome to release this as 3rd party plugin like others have done. See https://github.com/coldfix/certbot-dns-netcup for one example of that.

PR #6137, opened 2018-06-21
Show both possible Nginx default server root values in docs

ohemorange:
See https://github.com/certbot/website/pull/348#issuecomment-399257703.

```
$ certbot --help all | grep -C 3  nginx-server-root
nginx:
  Nginx Web Server plugin - Alpha

  --nginx-server-root NGINX_SERVER_ROOT
                        Nginx server root directory. (default: /etc/nginx)
  --nginx-ctl NGINX_CTL
                        Path to the 'nginx' binary, used for 'configtest' and
 ```

```
$ CERTBOT_DOCS=1 certbot --help all | grep -C 3  nginx-server-root
nginx:
  Nginx Web Server plugin - Alpha

  --nginx-server-root NGINX_SERVER_ROOT
                        Nginx server root directory. (default: /etc/nginx or
                        /usr/local/etc/nginx)
  --nginx-ctl NGINX_CTL
```

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -69,10 +69,7 @@ class NginxConfigurator(common.Installer):
 
     @classmethod
     def add_parser_arguments(cls, add):
-        if os.environ.get("CERTBOT_DOCS") == "1":
-            default_server_root = "%s or %s" % (LINUX_SERVER_ROOT, FREEBSD_DARWIN_SERVER_ROOT)
-        else:
-            default_server_root = constants.CLI_DEFAULTS["server_root"]
+        default_server_root = _determine_default_server_root()
         add("server-root", default=default_server_root,
             help="Nginx server root directory.")
         add("ctl", default=constants.CLI_DEFAULTS["ctl"], help="Path to the "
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: certbot-nginx/certbot_nginx/tests/configurator_test.py [hunk 1]
```diff
@@ -949,5 +949,30 @@ def test_current_file_hash_in_all_hashes(self):
...
+        self._test(expect_both_values=False)
+
+    def _test(self, expect_both_values):
+        server_root = self._call()
+
+        if expect_both_values:
+            self.assertIn("/usr/local/etc/nginx", server_root)
+            self.assertIn("/etc/nginx", server_root)
+        else:
+            self.assertTrue("/etc/nginx" in server_root or "/usr/local/etc/nginx" in server_root)
```
bmw: This will also pass if both values are included in the `server_root`. Can we change this?

certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -69,7 +69,8 @@ class NginxConfigurator(common.Installer):
 
     @classmethod
     def add_parser_arguments(cls, add):
-        add("server-root", default=constants.CLI_DEFAULTS["server_root"],
+        default_server_root = _determine_default_server_root()
+        add("server-root", default=default_server_root,
```
bmw: Up to you if you want to fix this, but by actually setting the default to `default_server_root`, if you use the nginx plugin with `CERTBOT_DOCS` set and without using `--nginx-server-root`, it tries to use the value "/etc/nginx or /usr/local/etc/nginx" as the server root with `nginx`.

If we wanted to fix this, we can take advantage of some magic added to `certbot/cli.py` and do something like https://github.com/certbot/certbot/commit/3cb02c08394d98c0fd65993c06e5c2e96d139777.

certbot-nginx/certbot_nginx/tests/configurator_test.py
ohemorange: oops

certbot-nginx/certbot_nginx/configurator.py
ohemorange: I like it

PR #6138, opened 2018-06-22
run Isort on imported packages

GreatBahram:
HI fellow
I've just run Isort on `setup.py` because I though that was ugly. why two line when you can make one line?
Thanks


schoen:
@bmw @ohemorange, our style seems a bit inconsistent here—do you have a preference or is there an established norm about this?

PR #6145, opened 2018-06-26
Don't use hardcoded port in tests

bmw:
I was running Python 2 and Python 3 unit tests at the same time locally and they failed during `acme.standalone` tests because both tried to grab port 1234. This removes this hardcoded port and uses port 0 instead so a free port can be allocated by the OS.

@joohoi, can you review this?

### DIFF
--- a/acme/acme/standalone_test.py
+++ b/acme/acme/standalone_test.py
@@ -336,7 +336,7 @@ class TestSimpleTLSSNI01Server(unittest.TestCase):
     def test_it(self, mock_logger):
         max_attempts = 5
         self.thread.start()
-        for attempt in range(max_attempts):
+        for _ in range(max_attempts):
             if mock_logger.info.called:
                 # Get the port selection which was logged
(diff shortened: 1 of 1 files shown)


bmw:
This should be ready for review now.

bmw:
This should be ready for review now.

bmw:
OK. Tests actually passed now. Sorry for all the false starts here!

bmw:
Thanks for the review @joohoi.

When I was about to merge this, it hit me that my changes to `standalone_test` could cause issues if `mock` objects aren't thread safe. I did some googling and it seems like they're [not thread safe](https://twitter.com/nailor/status/402479869130317824). To avoid spurious failures, I refactored the standalone test a bit to work around this.

bmw:
Also,

> Not in the scope of this PR, but these could probably use a complete revamp at some point in the future.

I agree. Since I'm not making this situation worse and fixing other problems, I agree it should happen in another PR too.

PR #6149, opened 2018-06-26
Fix server_root default tests on macOS

bmw:
Tests are failing on macOS. See https://travis-ci.org/certbot/certbot/jobs/397040447#L2510.

This is because if `"/usr/local/etc/nginx" in server_root` is true, `"/etc/nginx" in server_root` is also true so `self.assertFalse("/etc/nginx" in server_root and "/usr/local/etc/nginx" in server_root)` fails.

This fixes that by using `==` rather than `in`.


ohemorange:
maybe closing and reopening will help?

bmw:
Maybe. I think the issue is related to https://github.com/lemurheavy/coveralls-public/issues/1149.

If we don't hear back or get a resolution soon, maybe we should just stop requiring coveralls tests to pass until they are fixed again.

PR #6152, opened 2018-06-27
Update default to ACMEv2 server

### LINKED ISSUE #5369: Change the default ACME server to the v2 endpoint
Step 4 of #5365 
- [ ] tools/deactivate.py
- [ ] acme/examples/example_client.py
The backend data structures are the same for the different endpoints, but we might want to either determine the client to use based on the server like we do in certbot, or at least use acmev2 styles.

ohemorange:
Fixes #5369.


bmw:
And #6155.

PR #6156, opened 2018-06-28
Use account reuse symlink logic when loading an account

### LINKED ISSUE #6154: Use account reuse symlink logic when loading an account
To reproduce, start with an empty Certbot config directory and use any released version of Certbot to obtain an ACMEv1 prod cert. You shouldn't set `--server` anywhere.

Then checkout the `acmev2-default` branch and run `certbot renew --force-renewal`. You should see an error like:
```
  File "/home/bmw/src/certbot/certbot/renewal.py", line 430, in handle_renewal_request
    main.renew_cert(lineage_config, plugins, renewal_candidate)
  File "/home/bmw/src/certbot/certbot/main.py", line 1190, in renew_cert
    le_client = _init_le_client(config, auth, installer)
  File "/home/bmw/src/certbot/certbot/main.py", line 641, in _init_le_client
    acc, acme = _determine_account(config)
  File "/home/bmw/src/certbot/certbot/main.py", line 508, in _determine_account
    acc = account_storage.load(config.account)
  File "/home/bmw/src/certbot/certbot/account.py", line 224, in load
    return self._load_for_server_path(account_id, self.config.server_path)
  File "/home/bmw/src/certbot/certbot/account.py", line 204, in _load_for_server_path
    "Account at %s does not exist" % account_dir_path)
AccountNotFound: Account at /etc/letsencrypt/accounts/acme-v02.api.letsencrypt.org/directory/8f7e408abe6ff5832f20529fb812abad does not exist
```
This happens because the default server value changed to ACMEv2 and Certbot is looking for the account specified in the renewal configuration file in the ACMEv2 account directory which doesn't exist and hasn't been symlinked yet.

ohemorange:
Fixes #6154.

### DIFF
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -175,33 +175,44 @@ class AccountFileStorage(interfaces.AccountStorage):
             except errors.AccountStorageError:
                 logger.debug("Account loading problem", exc_info=True)
 
-
         if not accounts and server_path in constants.LE_REUSE_SERVERS:
             # find all for the next link down
             prev_server_path = constants.LE_REUSE_SERVERS[server_path]
             prev_accounts = self._find_all_for_server_path(prev_server_path)
             # if we found something, link to that
             if prev_accounts:
-                if os.path.islink(accounts_dir):
-                    os.unlink(accounts_dir)
-                else:
-                    try:
-                        os.rmdir(accounts_dir)
-                    except OSError:
-                        return []
-                prev_account_dir = self.config.accounts_dir_for_server_path(prev_server_path)
-                os.symlink(prev_account_dir, accounts_dir)
+                try:
+                    self._symlink_to_accounts_dir(prev_server_path, server_path)
+                except OSError:
+                    return []
             accounts = prev_accounts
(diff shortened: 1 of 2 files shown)


certbot/account.py
```diff
@@ -175,33 +175,44 @@ def _find_all_for_server_path(self, server_path):
...
+                except OSError:
+                    return []
             accounts = prev_accounts
         return accounts
 
     def find_all(self):
         return self._find_all_for_server_path(self.config.server_path)
 
+    def _symlink_to_accounts_dir(self, prev_server_path, server_path):
+        server_path = self.config.server_path
```
bmw: This line immediately overwrites the `server_path` parameter passed to this function. I think we should just delete this line as it should allow these methods to just work if there's ever something like an ACMEv3 server that's backwards compatible with both v2 and v1 accounts.

ohemorange: Nice catch, this is a mistake caused by too much moving code around. The entire point of this function is to pass in the `server_path` parameter...

Issue #6056, opened 2018-05-31
Fix test farm tests cleanup


bmw:
These tests in `tests/letstest/multitester.py` often fails to cleanup after itself and leaves around EC2 instances and EBS volumes that we try and remember to manually remove. This cleanup should be much more reliable so we don't have to remember to manually do anything.

bmw:
I think this is as done as it's ever going to get.

PR #6160, opened 2018-06-28
Automate EBS cleanup

bmw:
Part of #6056.

I have another piece to this locally which makes us terminating the instances more reliable, but I thought I'd split up the PR and first submit this which causes Amazon to automatically handle deleting EBS volumes when instances are shutdown.

The general approach here is described at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/RootDeviceStorage.html#Using_RootDeviceStorage. Doing something like this significantly speeds up cleanup and ensures we always cleanup storage because [EC2 only supports EBS and instance volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html#block-device-mapping-def) and [instance volumes cannot outlive the associated instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-lifetime).

Probably the easiest way to tests this is to run a test with `--killboulder` on the command line and wait for the test to finish or hit ctrl+c after you see "Waiting on Boulder Server". If you do that and wait a few minutes, you can then run `aws ec2 --profile <profile> describe-volumes` and if no one else is running tests on our account, the output should be empty.

### DIFF
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -173,33 +173,8 @@ def terminate_and_clean(instances):
     Some AMIs specify EBS stores that won't delete on instance termination.
     These must be manually deleted after shutdown.
     """
-    volumes_to_delete = []
-    for instance in instances:
-        for bdmap in instance.block_device_mappings:
-            if 'Ebs' in bdmap.keys():
-                if not bdmap['Ebs']['DeleteOnTermination']:
-                    volumes_to_delete.append(bdmap['Ebs']['VolumeId'])
-
     for instance in instances:
         instance.terminate()
-
-    # can't delete volumes until all attaching instances are terminated
-    _ids = [instance.id for instance in instances]
-    all_terminated = False
-    while not all_terminated:
-        all_terminated = True
-        for _id in _ids:
-            # necessary to reinit object for boto3 to get true state
-            inst = EC2.Instance(id=_id)
-            if inst.state['Name'] != 'terminated':
-                all_terminated = False
-        time.sleep(5)
-
-    for vol_id in volumes_to_delete:
-        volume = EC2.Volume(id=vol_id)
-        volume.delete()
-
-    return volumes_to_delete
 
 
 # Helper Routines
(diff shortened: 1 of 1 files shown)


tests/letstest/multitester.py
```diff
@@ -151,38 +152,21 @@ def make_instance(instance_name,
...
-                all_terminated = False
-        time.sleep(5)
-
-    for vol_id in volumes_to_delete:
-        volume = EC2.Volume(id=vol_id)
-        volume.delete()
-
-    return volumes_to_delete
+    # Not all devices use EBS, but the default value for DeleteOnTermination
+    # when the device does use EBS is true. See:
```
ohemorange: When would this value not be set to true, if it's the default?

bmw: Each image has its own default settings for its block devices which is what this function is checking here with `EC2.Image(ami_id).block_device_mappings`. If the image configures a device to use EBS and doesn't set `DeleteOnTermination`, the default value is true, but images can set its value to false. Some images, including ones that we use, do this.

ohemorange: Oh, as in images that other people made that we're using?

bmw: exactly

tests/letstest/multitester.py
```diff
@@ -151,38 +152,21 @@ def make_instance(instance_name,
...
+    # * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html
+    # * https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html
+    return [{'DeviceName': mapping['DeviceName'],
```
ohemorange: This would overwrite any existing values in the mapping. Are we 100% sure that no other values exist or might come to exist in the images we use?

bmw: The values we set here get merged into the settings for the instance and any values we do not set, including other properties for `Ebs`, keep the values defined by the instance. This is alluded to in the first paragraph at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html#instance-block-device-mapping, but I think that wording is a little ambiguous. To double check, I set a breakpoint, walked through the code, and verified this is the case.

Thanks for having me check this though. I had assumed things work this way, but upon looking into it, I wasn't actually sure what the behavior was.

Issue #6153, opened 2018-06-27
certbot renew - Certbot not completely/immediately skipping domains not yet due for renewal


atp1122:
Hello,

Since upgrading Certbot from 0.22.2 to 0.25.0, the "certbot renew" command does not seem to be completely/immediately skipping domains not yet due for renewal.

## My operating system is (include version):

Ubuntu 16.04.4 LTS
Certbot version: 0.25.0
Nginx version: 1.10.3

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

apt-get

## I ran this command and it produced this output:

certbot renew

## Certbot's behavior differed from what I expected because:

The "certbot renew" command used to immediately skip domains not due for renewal.  Instead, Certbot is still selecting the plugin (which is taking about 30 seconds per domain).  During this time, the Certbot command is utilizing between 75%-100% of the CPU.  We have a large number of domains so the renew command ends up taking a very long time (during which the CPU is near 100%).

NOTE: this behavior began after upgrading Certbot from version 0.22.2 to 0.25.0.  Previously, the process would skip domains not due for renewal immediately and move to the next domain.

## Here is a Certbot log showing the issue (if available):

2018-06-27 19:00:53,775:DEBUG:certbot.main:certbot version: 0.25.0
2018-06-27 19:00:53,775:DEBUG:certbot.main:Arguments: []
2018-06-27 19:00:53,776:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#manual,PluginEntryPoint#nginx,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
2018-06-27 19:00:53,787:DEBUG:certbot.log:Root logging level set at 20
2018-06-27 19:00:53,788:INFO:certbot.log:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2018-06-27 19:00:53,801:DEBUG:certbot.plugins.selection:Requested authenticator <certbot.cli._Default object at 0x7f8417863470> and installer <certbot.cli._Default object at 0x7f8417863470>
2018-06-27 19:00:53,801:DEBUG:certbot.cli:Var rsa_key_size=4096 (set by user).
2018-06-27 19:00:53,809:INFO:certbot.renewal:Cert not yet due for renewal
2018-06-27 19:00:53,810:DEBUG:certbot.plugins.selection:Requested authenticator nginx and installer nginx
[...]

PR #6167, opened 2018-07-02
Do not call IPlugin.prepare() for updaters when running renew

joohoi:
`interfaces.GenericUpdater` and new enhancement interface updater functions get run on every invocation of Certbot with "renew" verb for every lineage. This causes performance problems for users with large configurations, because of plugin plumbing and preparsing happening in `prepare()` method of installer plugins. This PR moves the responsibility to call `prepare()` to the plugin (possibly) implementing a new style enhancement interface.

Fixes: #6153

### DIFF
--- a/certbot/plugins/enhancements.py
+++ b/certbot/plugins/enhancements.py
@@ -87,8 +87,8 @@ class AutoHSTSEnhancement(object):
     increase over time.
 
     The plugins implementing new style enhancements are responsible of handling
-    the saving of configuration checkpoints as well as calling possible restarts
-    of managed software themselves.
+    the saving of configuration checkpoints as well as the other calls to
+    interface methods of `interfaces.IInstaller` such as prepare() and restart()
 
     Methods:
         enable_autohsts is called when the header is initially installed using a
--- a/certbot/plugins/enhancements.py
+++ b/certbot/plugins/enhancements.py
@@ -87,8 +87,9 @@ class AutoHSTSEnhancement(object):
     increase over time.
 
     The plugins implementing new style enhancements are responsible of handling
-    the saving of configuration checkpoints as well as the other calls to
-    interface methods of `interfaces.IInstaller` such as prepare() and restart()
+    the saving of configuration checkpoints as well as calling possible restarts
+    of managed software themselves. For update_autohsts method, the installer may
+    have to call prepare() to finalize the plugin initialization.
 
     Methods:
         enable_autohsts is called when the header is initially installed using a
@@ -112,6 +113,10 @@ class AutoHSTSEnhancement(object):
 
         :param lineage: Certificate lineage object
         :type lineage: certbot.storage.RenewableCert
+
+        .. note:: prepare() method inherited from `interfaces.IPlugin` might need
+            to be called manually within implementation if this interface method
+            to finalize the plugin initialization.
         """
 
     @abc.abstractmethod
--- a/certbot/plugins/enhancements.py
+++ b/certbot/plugins/enhancements.py
@@ -115,7 +115,7 @@ class AutoHSTSEnhancement(object):
         :type lineage: certbot.storage.RenewableCert
 
         .. note:: prepare() method inherited from `interfaces.IPlugin` might need
-            to be called manually within implementation if this interface method
+            to be called manually within implementation of this interface method
             to finalize the plugin initialization.
         """
 
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -12,6 +12,8 @@ from certbot import interfaces
 
 from certbot.display import util as display_util
 
+from zope.interface.exceptions import DoesNotImplement
+
 logger = logging.getLogger(__name__)
 z_util = zope.component.getUtility
 
@@ -55,6 +57,11 @@ def get_unprepared_installer(config, plugins):
     if not req_inst:
         return None
     installers = plugins.filter(lambda p_ep: p_ep.name == req_inst)
+    installers.init(config)
+    try:
+        installers = installers.verify((interfaces.IInstaller,))
+    except DoesNotImplement:
+        return None
     if len(installers) > 1:
         raise errors.PluginSelectionError(
             "Found multiple installers with the name %s, Certbot is unable to "
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -47,27 +47,25 @@ def get_unprepared_installer(config, plugins):
     :param certbot.plugins.disco.PluginsRegistry plugins:
         All plugins registered as entry points.
 
-    :returns: Unprepared installer plugin.
-    :rtype: IPlugin
+    :returns: Unprepared installer plugin or None
+    :rtype: IPlugin or None
     """
 
     _, req_inst = cli_plugin_requests(config)
     if not req_inst:
-        raise errors.MissingCommandlineFlag(
-            "Missing command line flags. For non-interactive execution, "
-            "you will need to specify an installer plugin on the command line. "
-            "You can do this by running Certbot with '--installer PLUGIN_NAME ")
-    _installers = plugins.filter(lambda p_ep: p_ep.name == req_inst)
-    if _installers:
-        inst = list(_installers.values())[0]
+        return None
+    installers = plugins.filter(lambda p_ep: p_ep.name == req_inst)
+    if len(installers) > 1:
+        raise errors.PluginSelectionError(
+            "Found multiple installers with the name %s, Certbot is unable to "
+            "determine which one to use. Skipping." % req_inst)
+    if installers:
+        inst = list(installers.values())[0]
         logger.debug("Selecting plugin: %s", inst)
-        if inst.misconfigured:
-            raise errors.PluginSelectionError(
-                "Configuration error in plugin %s", req_inst)
         return inst.init(config)
     else:
-        raise errors.MisconfigurationError("Could not select or initialize the "
-                                           "requested installer %s.")
+        raise errors.PluginSelectionError(
+            "Could not select or initialize the requested installer %s." % req_inst)
 
 def pick_plugin(config, default, plugins, question, ifaces):
     """Pick plugin.
--- a/certbot/plugins/selection.py
+++ b/certbot/plugins/selection.py
@@ -39,6 +39,35 @@ def pick_authenticator(
     return pick_plugin(
         config, default, plugins, question, (interfaces.IAuthenticator,))
 
+def get_installer(config, plugins):
+    """
+    Get an unprepared interfaces.IInstaller object.
+
+    :param certbot.interfaces.IConfig config: Configuration
+    :param certbot.plugins.disco.PluginsRegistry plugins:
+        All plugins registered as entry points.
+
+    :returns: Unprepared installer plugin.
+    :rtype: IPlugin
+    """
+
+    _, req_inst = cli_plugin_requests(config)
+    if not req_inst:
+        raise errors.MissingCommandlineFlag(
+            "Missing command line flags. For non-interactive execution, "
+            "you will need to specify an installer plugin on the command line. "
+            "You can do this by running Certbot with '--installer PLUGIN_NAME ")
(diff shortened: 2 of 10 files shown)


certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+    :param certbot.interfaces.IConfig config: Configuration
+    :param certbot.plugins.disco.PluginsRegistry plugins:
+        All plugins registered as entry points.
+
+    :returns: Unprepared installer plugin.
+    :rtype: IPlugin
+    """
+
+    _, req_inst = cli_plugin_requests(config)
+    if not req_inst:
```
bmw: We don't necessarily have to do things this way, but why not make an optional parameter to `pick_installer` and `pick_plugin` named something like `prepared`, add an `if` block around preparing the plugin in `pick_plugin`, and then just check that that function returns something other than `None`? We certainly don't need all of the logic in `pick_plugin`, but I think a lot of the code here is duplicated.

Another approach would be to split the common code in `pick_plugin` into helper functions both functions can use.

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
     return pick_plugin(
         config, default, plugins, question, (interfaces.IAuthenticator,))
 
+def get_installer(config, plugins):
```
bmw: nit: Can we make this name a little more descriptive? Maybe by adding "unprepared"? Or maybe "required" since this unlike many of the other functions, this should never return `None`.

We have all these `{pick,get,choose]_*` functions in this file and it's not clear until reading the docstrings which does what.

certbot/updater.py
```diff
@@ -29,8 +29,10 @@ def run_generic_updaters(config, lineage, plugins):
         return
     try:
         # installers are used in auth mode to determine domain names
-        installer, _ = plug_sel.choose_configurator_plugins(config, plugins, "certonly")
-    except errors.PluginSelectionError as e:
+        installer = plug_sel.get_installer(config, plugins)
+    except (errors.MissingCommandlineFlag,
+            errors.MisconfigurationError,
+            errors.PluginSelectionError) as e:
```
bmw: Unless you're sure you listed all the error types here and there are errors we may throw which you don't want to catch, I'd just use `errors.Error` here.

certbot/plugins/enhancements.py
```diff
@@ -87,8 +87,8 @@ class AutoHSTSEnhancement(object):
     increase over time.
 
     The plugins implementing new style enhancements are responsible of handling
-    the saving of configuration checkpoints as well as calling possible restarts
-    of managed software themselves.
+    the saving of configuration checkpoints as well as the other calls to
+    interface methods of `interfaces.IInstaller` such as prepare() and restart()
```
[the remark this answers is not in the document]
joohoi: That's right, I changed both of these.

certbot/plugins/selection.py
joohoi: This was the approach I initially took, it however became apparent, that `pick_plugin` itself is tightly knit into the interactive selection process, being the helper function that's called by other selection functions. It's bascally meant to be called mostly interactively - helping with the selection where needed, and I think that calling `prepare()` is a required part of the process with other verbs where interactive selection might be in question. Note that the `prepare()` is in some cases called to a whole group of applicable plugins where a single one wasn't requested. 

The current (semi-interactive) selection process is split between `choose_configurator_plugins` and `pick_plugin` that the former calls. This whole process is incorporated in this new function of this PR.

That said, I'd prefer to keep it as it is, as a robust selection process for interactive selection needs the `prepare()` called in the middle.

certbot/plugins/selection.py
joohoi: I changed the name to `get_unprepared_installer()`

certbot/updater.py
joohoi: `errors.Error` is much more readable, using that instead as all of the errors are handled in a similar (non-interrupting) fashion here anyway.

certbot/plugins/enhancements.py
```diff
@@ -112,6 +113,10 @@ def update_autohsts(self, lineage, *args, **kwargs):
 
         :param lineage: Certificate lineage object
         :type lineage: certbot.storage.RenewableCert
+
+        .. note:: prepare() method inherited from `interfaces.IPlugin` might need
+            to be called manually within implementation if this interface method
```
bmw: s/if/of

certbot/plugins/selection.py
bmw: That's fair. We're protected from any interactivity because `renew` implies `noninteractive` mode which `IDisplay` knows about, but there's enough differences in behavior it's probably not worth shoving it all into the same function. Thanks for pushing back.

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+    :param certbot.interfaces.IConfig config: Configuration
+    :param certbot.plugins.disco.PluginsRegistry plugins:
+        All plugins registered as entry points.
+
+    :returns: Unprepared installer plugin.
+    :rtype: IPlugin
+    """
+
+    _, req_inst = cli_plugin_requests(config)
+    if not req_inst:
```
bmw: If you obtain a cert without an installer (e.g. `certonly`) and then run `certbot renew`, the following is output:
```
Could not choose appropriate plugin for updaters: Missing command line flags. For non-interactive execution, you will need to specify an installer plugin on the command line. You can do this by running Certbot with '--installer PLUGIN_NAME
```
We shouldn't do this. How about we just return `None` here?

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+        raise errors.MissingCommandlineFlag(
+            "Missing command line flags. For non-interactive execution, "
+            "you will need to specify an installer plugin on the command line. "
+            "You can do this by running Certbot with '--installer PLUGIN_NAME ")
+    _installers = plugins.filter(lambda p_ep: p_ep.name == req_inst)
```
bmw: Why the underscore prefix here?

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+    if _installers:
+        inst = list(_installers.values())[0]
```
bmw: In the extremely unlikely case there's two+ installers with the same name, we're just arbitrarily picking one. I think the most common case of this would a name collision with third party plugins. Since picking the wrong plugin could cause problems down the line, how about we fail fast here and check that `len(_installers) == 1` in the `if` statement above? If it's not 1, we'll error out below.

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+        logger.debug("Selecting plugin: %s", inst)
+        if inst.misconfigured:
```
bmw: I don't think we need this check because `misconfigured` relies on `prepared` having been called. See https://github.com/certbot/certbot/blob/ab9851d97e7f58290d0299da8fec49f072226750/certbot/plugins/disco.py#L144.

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+            raise errors.PluginSelectionError(
+                "Configuration error in plugin %s", req_inst)
```
bmw: This raises an exception with two strings rather than expanding `%s` to `req_inst`.

certbot/updater.py
```diff
@@ -29,8 +29,8 @@ def run_generic_updaters(config, lineage, plugins):
         return
     try:
         # installers are used in auth mode to determine domain names
```
bmw: I think this comment is outdated as I believe it was referring to the use of "certonly".

certbot/plugins/selection.py
```diff
@@ -39,6 +39,35 @@ def pick_authenticator(
...
+        return inst.init(config)
+    else:
+        raise errors.MisconfigurationError("Could not select or initialize the "
+                                           "requested installer %s.")
```
bmw: We're not putting anything into `%s` here.

Also, nit: If we raise this error, we didn't try to initialize the plugin, we just failed to select it. With this in mind, maybe the error message should be updated and `errors.PluginSelectionError` used.

certbot/plugins/selection_test.py
```diff
@@ -170,5 +173,52 @@ def test_new_interaction_avoidance(self, mock_util):
 
         self.assertTrue("default" in mock_util().menu.call_args[1])
 
+class GetInstallerTest(test_util.ConfigTestCase):
```
bmw: nit: Can we change this to `GetUnpreparedInstallerTest`?

certbot/plugins/selection_test.py
```diff
@@ -170,5 +173,52 @@ def test_new_interaction_avoidance(self, mock_util):
...
+    """Tests for certbot.plugins.selection.get_unprepared_installer."""
+
+    def setUp(self):
+        super(GetInstallerTest, self).setUp()
+        self.mock_apache_fail_ep = mock.Mock(
+            description_with_name="afail", misconfigured=True)
+        self.mock_apache_fail_ep.name = "afail"
+        self.mock_apache_ep = mock.Mock(
+            description_with_name="apache", misconfigured=False)
```
bmw: I think we can stop setting `misconfigured` here and above.

certbot/plugins/selection_test.py
```diff
@@ -170,5 +173,52 @@ def test_new_interaction_avoidance(self, mock_util):
...
+                          self._call)
+
+    def test_get_plugin(self):
+        self.config.configurator = "apache"
+        installer = self._call()
+        self.assertTrue(installer is self.mock_apache_plugin)
+
+
+
+
```
bmw: super nit: too many blank lines

certbot/plugins/selection.py
joohoi: For no reason, in the initial implemention this was a temp variable, and the notation was leftover from there. Underscore removed for consistency.

certbot/plugins/selection.py
joohoi: Nice catch. This is now implemented.

certbot/plugins/selection.py
joohoi: Oh, that's right, it's not checked anymore.

certbot/updater.py
joohoi: That's right, removed.

certbot/plugins/selection.py
joohoi: Changed to proper exception type and fixed formatting.

certbot/plugins/selection.py
```diff
@@ -39,6 +39,33 @@ def pick_authenticator(
...
+        All plugins registered as entry points.
+
+    :returns: Unprepared installer plugin or None
+    :rtype: IPlugin or None
+    """
+
+    _, req_inst = cli_plugin_requests(config)
+    if not req_inst:
+        return None
+    installers = plugins.filter(lambda p_ep: p_ep.name == req_inst)
```
bmw: I'm sorry I didn't catch this in my previous review, but there's actually no check in this function that the plugin returned here is an installer. Can we add this?

I think we should do this by calling the `verify` method on the `PluginsRegistry` object like we do in `pick_plugin`. We should do this before calling `len(installers)` for consistency and to avoid erroring out if there's only one installer plugin with that name.

joohoi: Oh, nice catch! This is now added.

certbot/plugins/selection_test.py
```diff
@@ -215,6 +215,13 @@ def test_multiple_installers_returned(self):
         self.mock_apache_fail_ep.name = "apache"
         self.assertRaises(errors.PluginSelectionError, self._call)
 
+    def test_does_not_implement(self):
+        from zope.interface.exceptions import DoesNotImplement
```
bmw: nit: We should import this at the top of the file rather than in this method.

PR #6169, opened 2018-07-02
Upgrade to the latest cryptography version

bmw:
This allows certbot-auto and our development setup to work with Python 3.7.

Things I checked were:

* The hashes obtained from a couple different network vantage points.
* No new dependencies were added: https://paste.ubuntu.com/p/zzNsNbBN6c/
* Nothing we depend on was deprecated: https://github.com/pyca/cryptography/blob/2.2.2/CHANGELOG.rst

PR #6170, opened 2018-07-02
Allow py37 testing

bmw:
Allows us to test on Python 3.7 until https://github.com/yaml/pyyaml/issues/126 is resolved. Once that issue is resolved, this PR can be reverted.

With the changes here, those in #6169, and changing the macOS tests to use Python 3.7, the tests pass as shown at https://travis-ci.org/certbot/certbot/builds/399337786.


bmw:
This can be closed if https://github.com/yaml/pyyaml/issues/126 is closed before this gets reviewed. Otherwise, I'd like this to land and have us revert it later so we can [fix macOS tests](https://travis-ci.org/certbot/certbot/builds/399732654) and begin Python 3.7 testing.

Issue #5920, opened 2018-05-03
The nginx plugin is not working;The error was NoInstallationError()


ray0324:
## My operating system is (include version):
CentOS 7 x86_64

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

certbot-nginx

## I ran this command and it produced this output:

```
crontab -e

01 00 * * * certbot renew --dry-run &> /var/log/certbot.cronlog

```

![fail](https://user-images.githubusercontent.com/5204198/39557163-2f665ab0-4eb8-11e8-9df7-7197d3b8c4f0.jpg)

## I ran this command and it  produced this output:

```
certbot renew --dry-run
```
![ok](https://user-images.githubusercontent.com/5204198/39557213-7d6c512e-4eb8-11e8-9f7a-59fad0d59313.jpg)

bmw:
The problem is the `nginx` executable isn't in your `PATH` in your cron environment.

bmw:
I'm going to close this issue because resolving that should solve your problem, but feel free to comment if you'd like me to reopen.

PR #6175, opened 2018-07-03
More detailed error logging for nginx plugin

davidjb:
This makes errors more useful when Nginx can't be found or when Nginx's
configuration can't be found.  Previously, a generic `NoInstallationError` isn't descriptive enough to explain _what_ wasn't installed or what failed without going digging into the source code.

An error like:
```
 The nginx plugin is not working; there may be problems with your existing configuration.
The error was: NoInstallationError()
```
seems to imply there's a problem with "configuration" which I've previously interpreted as Certbot config problem rather than the Nginx _installation_ or in other common cases the `PATH` isn't right when say run in `cron` (#5920).

This PR just makes the error messages more verbose, helping the user to know what to check:

```
Could not choose appropriate plugin for updaters: The nginx plugin is not working; there may be problems with your existing configuration.
The error was: NoInstallationError("Could not find a usable 'nginx' binary. Ensure nginx exists, the binary is executable, and your PATH is set correctly.",)
```

The error formatting isn't perfect, but that's beyond the scope of this PR; now the user knows what they need to go and check.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -136,7 +136,9 @@ class NginxConfigurator(common.Installer):
         """
         # Verify Nginx is installed
         if not util.exe_exists(self.conf('ctl')):
-            raise errors.NoInstallationError
+            raise errors.NoInstallationError(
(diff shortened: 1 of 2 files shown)


ohemorange:
Closing and reopening to get codecov to work

PR #6176, opened 2018-07-03
Delete empty directories after deleting an account, including symlinks up and down the chain, as appropriate

### LINKED ISSUE #6172: Handle symlinked account deletion
From discussion at https://github.com/certbot/certbot/issues/6155.

> As a more minor point, we should probably check that symlinked accounts are deleted in the way we expect.

Deleting the accounts deletes the particular account under the `directory/` folder, which remains symlinked. If you then create an ACMEv2 account, it puts the ACMEv2 account into the ACMEv1 folder, with a symlink to it from ACMEv2.

The easiest way to deal with this is probably to delete directories if they're empty.

ohemorange:
Fixes #6172.


certbot/account.py
```diff
@@ -250,12 +250,48 @@ def delete(self, account_id):
...
+        reused_servers = {}
+        for k in constants.LE_REUSE_SERVERS:
+            reused_servers[constants.LE_REUSE_SERVERS[k]] = k
+
+        if server_path in reused_servers:
+            next_server_path = reused_servers[server_path]
+            next_accounts_dir_path = self.config.accounts_dir_for_server_path(next_server_path)
+            if os.path.islink(next_accounts_dir_path) \
+                and os.readlink(next_accounts_dir_path) == accounts_dir_path:
+                self._delete_accounts_dir_for_server_path(next_server_path)
```
bmw: In the case you delete the account using ACMEv1 which is symlinked from ACMEv2, I think this call also deletes the current `accounts_dir_path` directory causing us to blow up below when we try to delete it again.

bmw:
@ohemorange, do you have time to finish this PR before the release? I think we should try to get it in.

ohemorange:
Yes, that's the plan. It's easier to test this with the other one in first.

PR #6184, opened 2018-07-06
Upgrade pinned josepy version

bmw:
We released josepy 1.1.0 a while ago to work around newer versions of cryptography deprecating some of the functionality we were using. We haven't yet upgraded our pinned josepy version though and since #6169 has landed, we're now [seeing these deprecation warnings in our tests](https://travis-ci.org/certbot/certbot/jobs/400951626#L721). This would be shown to `certbot-auto` users as well.

This PR removes these warnings by upgrading our pinned version of josepy.


bmw:
> Related, but separate question: do we have an upgrade path or plan for acme dependencies?

This is tricky. The old versions of our dependencies work just fine with older versions of Python and some distros are relying on this behavior. On top of that, conditionally changing our dependencies based on the Python version isn't supported in the old versions of setuptools used in some distros. Because of this, we've been ignoring problems like this in our Python packages and just handling them in `certbot-auto` and distro packages.

See #3455 and https://setuptools.readthedocs.io/en/latest/history.html#id283 for more info. I'm happy to chat about this more if you'd like.

Issue #6164, opened 2018-07-01
HelpfulArgumentParser not parsing arguments the same way when detect_defaults=True


sTywin:
Hi folks,

I am trying to write an installer plugin. I need an optional argument (parameter) from the user that has one or more values (i.e. `nargs='+'`). In the config namespace passed to my plugin's `__init__`, I can see the parameter I specify, and it works as expected:

- [x] When not specified, the config value is `None`
- [x] When specified with no values, I get an error "expected at least one argument"
- [x] When specified with a single value, the config is `["value0"]`
- [x] When specified with e.g. 2 values, the config is `["value0","value1"]`

These are all the correct behavior that I would expect from argparse.

However, if I combine this with the `--dns-route53` authenticator plugin, I run into some problems. The config my plugin get when certbot starts up is still correct in all cases, the `dns-route53` authenticator runs successfully, but then in the last case above (multiple values), I am getting `certbot: error: unrecognized arguments: value1` (along with a usage printout) before my plugin is run, and then certbot quits. 

If the `dns-route53` authenticator is skipped (i.e. because an existing certificate is found), then my installer plugin is executed correctly.

Suggestions? I'm not really set up for other authentication methods so I haven't been able to try those.

macOS 10.13.5
Python 3.6.5
certbot==0.25.1
certbot-dns-route53==0.25.1

sTywin:
OK, I'm not sure this is related to `dns-route53` at all, so much as `certbot`'s handling of parameters, possibly when trying to write out the renewal configuration file? I'm kind fo drowning here. I've dug down and found the stack traces for where the error message pops up, both for new configurations and renewed configurations.

New:
```
main.py:main
main.py:run
main.py:_get_and_save_cert
client.py:obtain_and_enroll_certificate
storage.py:new_lineage
storage.py:relevant_values
cli.py:option_was_set
cli.py:set_by_cli
cli.py:prepare_and_parse_args
cli.py:parse_args
```

Renewal:
```
main.py:main
main.py:run
main.py:_get_and_save_cert
renewal.py:renew_cert
storage.py:save_successor
storage.py:update_configuration
storage.py:relevant_values
cli.py:option_was_set
cli.py:set_by_cli
cli.py:prepare_and_parse_args
cli.py:parse_args
```

[...]

PR #6185, opened 2018-07-07
Fix detect defaults when nargs present

sTywin:
If a plugin adds a command line argument with the `nargs` parameter, e.g. `nargs='+'`, this parameter gets filtered from the `HelpfulArgumentParser` created/used in default detection. As a result, this parser does not parse the command line in the expected way, instead using the default value for `nargs`.

Leaving `nargs` in place resolves this issue. See #6164.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -120,7 +120,7 @@ More detailed help:
 
 
 # These argparse parameters should be removed when detecting defaults.
-ARGPARSE_PARAMS_TO_REMOVE = ("const", "nargs", "type",)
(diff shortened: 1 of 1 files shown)


sTywin:
Hmm, Travis-CI seems to have choked on its own setup. Will try re-running.

sTywin:
Yeah, Travis-CI choked on itself again.. different build than the first time. Between both jobs, all tests pass.

sTywin:
Trying once more..

sTywin:
There is a second-order problem in `renewal._restore_plugin_configs` where non-string options are being serialized into strings in the renewal configuration file. I am testing a fix for this now.

PR #6188, opened 2018-07-09
Fix detect defaults and config serialization in presence of nargs

sTywin:
If a plugin adds a command line argument with the nargs parameter, e.g. nargs='+', this parameter gets filtered from the HelpfulArgumentParser created/used in default detection. As a result, this parser does not parse the command line in the expected way, instead using the default value for nargs.

Leaving nargs in place resolves this issue. See #6164.

Additionally, in `renewal._restore_plugin_configs`, plugin renewal parameters of list type (as generated by nargs='+', among others) were being serialized as strings. A wrapping function has been added to handle list-based parameters.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -120,7 +120,7 @@ More detailed help:
 
 
 # These argparse parameters should be removed when detecting defaults.
-ARGPARSE_PARAMS_TO_REMOVE = ("const", "nargs", "type",)
+ARGPARSE_PARAMS_TO_REMOVE = ("const", "type",)
 
 
 # These sets are used when to help detect options set by the user.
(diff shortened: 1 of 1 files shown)


sTywin:
Hi folks, what do I need to do to get someone to review this? I am developing an installer plugin that depends on this fix for list-type arguments.

bmw:
I'm closing this PR because it hasn't seen any response in a couple months. Feel free to address my review comments and open a new PR if you'd like.

I also moved this code to the [nargs-default](https://github.com/certbot/certbot/tree/nargs-defaults) branch and referenced it in #6164 in case anyone else wants to take this on.

PR #6191, opened 2018-07-10
Properly handle None value for installer in updater

joohoi:
Our configuration parser reads all of the configuration values as strings. This causes `installer = None` to become 'None' in the configuration object, and the installer plugin selection fails because of it. This PR adds a simple check to the code determining which installer or authenticator plugin was requested, and returns a proper `None` type in the case described above. This allows us to return early from the updater plugin selection code.


bmw:
Since #6195 has landed, should this be closed?

joohoi:
Yeah, indeed!

PR #6192, opened 2018-07-10
certbot-dns-exoscale plugin

greut:
This is a first attempt to provide a plugin for the Exoscale DNS api (which is already in xenolf/lego). As it's very similar to the old DNSimple API and using dns-lexicon. It was easy to build it re-using the code from the DNSimple plugin.

**disclaimer:** Exoscale is my employer.


certbot-dns-exoscale/certbot_dns_exoscale/dns_exoscale.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(self, key, secret, ttl):
+        super(_ExoscaleLexiconClient, self).__init__()
+
+        self.provider = exoscale.Provider(
+            {"auth_key": key, "auth_secret": secret, "ttl": ttl}
+        )
+
+    def _handle_http_error(self, e, domain_name):
+        hint = None
+        if str(e).startswith("401 Client Error: Unauthorized for url:"):
```
zjs: It's not clear from the Exoscale Lexicon plugin or Exoscale API documentation, so I just want to confirm: Is this the only error the Exoscale API returns when you provide an invalid API credentials? (Some providers return different errors for different kinds of authentication issues, such as valid key with invalid secrets, invalid key with invalid secret, etc.)

certbot-dns-exoscale/certbot_dns_exoscale/dns_exoscale.py
```diff
@@ -0,0 +1,101 @@
...
+            hint = "Is your API token value correct?"
```
zjs: Given that Exoscale refers to "key" and "secret", the hint should probably use those terms (not "token").

docs/cli-help.txt
```diff
@@ -551,6 +551,17 @@ dns-dnsmadeeasy:
   --dns-dnsmadeeasy-credentials DNS_DNSMADEEASY_CREDENTIALS
                         DNS Made Easy credentials INI file. (default: None)
 
+dns-exoscale:
```
zjs: I don't think this file is typically updated as a part of this sort of commit; I think it's updated as a part of the release process. (@bmw?)

bmw: > I think it's updated as a part of the release process.

That's correct. We do it this way to help make sure that documentation for unreleased features don't make their way to our site.

That fact that we do this is not very clear though.

greut: fair enough

greut:
@zjs a friendly ping, if you have time to review the changes made. Cheers,

PR #6198, opened 2018-07-11
Remove linode and ovh links which aren't valid yet.

bmw:
These links were added in #5302, but should be removed until they are valid.


bmw:
OVH values were added incorrectly in #5423.

PR #6208, opened 2018-07-12
Doc fixes for certbot-dns-linode

trinopoty:
1) Add linode docs to `docs/cli-help.txt`.
2) Fix API Keys page url in `certbot-dns-linode/certbot_dns_linode/__init__.py` docs

This should be merged before next release if possible.


bmw:
> Add linode docs to docs/cli-help.txt.

This will be handled in https://github.com/certbot/certbot/pull/6206, but if you want to update the PR to just change the API keys URL, I'll merge it.

Thanks for fixing this @trinopoty!

trinopoty:
I've updated the PR. Turns out I put the URL for the new console before I was aware of the fact that API keys from the new console does not work with v3 of the API.

bmw:
No worries. Thanks again for following up with a fix!

PR #6220, opened 2018-07-13
Switch to codecov

bmw:


bmw:
Playing with turning off PR comments and adding commit status checks. Seeing if closing and reopening will retrigger the checks.

bmw:
Codecov is saying "No report found to compare against" because it hasn't run in master yet which will change when this PR is merged.

ohemorange:
Does the fact that it didn't repost mean that commenting is now turned off?

bmw:
> Does the fact that it didn't repost mean that commenting is now turned off?

Yep! I have it set up so all repos in the Certbot GitHub organization share the same settings which can be seen and modified at https://codecov.io/account/gh/certbot/yaml. Let me know if you want to chat about what I have configured and our options there.

Thanks for the review!

ohemorange:
Probably don't need to chat, but I would love to check out the settings -- I don't seem to have access to that page though?

bmw:
> I don't seem to have access to that page though?

Even after clicking the "Login with GitHub" link at the bottom? I figured you'd have access after reading https://docs.codecov.io/docs/administration.

ohemorange:
Ah yup that would be it

PR #6223, opened 2018-07-16
Exclude one-time use parameters. Fixes #6118

### LINKED ISSUE #6118: socket-related parameters to the listen directive should not be copied over for an already used address:port pair
CentOS 7.5, certbot 0.24.0 from epel, NGINX plugin.

http-01 challenge (for a renew) failed, the modified nginx.conf resulted in a `nginx: [emerg] duplicate listen options for` error.

The original `listen` directive for port 80 had the `deferred` parameter set and this was copied over to the `listen` directive inserted by `certbot-nginx`. The `deferred` parameter can be, similarly to the `ipv6only` parameter, only specified once for a given address:port pair.

There's a whole bunch of `listen` directive parameters like this, see https://nginx.org/en/docs/http/ngx_http_core_module.html#listen. I see there's a "blacklist" in `certbot_nginx/parser.py` that (optionally) takes care of the `default_server` and `ipv6_only` parameters. I'm not aware of all the contexts in which a new server block for an already used address:port pair needs to be created by certbot-nginx, but for temporary server blocks for challenge response serving, perhaps ignoring all the listen parameters altogether, apart from the address:port pair, could be an option? (Just a speculation on my part, perhaps shortsighted...)

Dreamsorcerer:


schoen:
This looks plausible to me and I'll suggest that @ohemorange take a look at it.

One idle question: are all of these directives strictly single-use, or could some of them potentially be multiple-use in the context of different listen ports or IP versions? (But maybe that possibility is completely irrelevant in this specific context since we nonetheless don't want to copy them when creating new server blocks based on existing server blocks?)

Dreamsorcerer:
@ohemorange How's this? It's slightly awkward iterating over a list in-place, but still fairly clean, and remains efficient and maintainable.

Dreamsorcerer:
OK, I'm sure that code is wrong, but the tests are passing. I can't quite get my head around how the unit tests work though. So, would it be possible for you to quickly add a new test that fails if "ipv6=on" gets duplicated in a config. I can't see any tests currently testing anything with that setting.

ohemorange:
Yeah, you're right, `index` won't work directly because of just checking for `ipv6only` vs `ipv6only=on`, will probably have to either iterate through the directive or transform the directive using an `item.split('=')[0] for item in directive` then finding the index in that list and deleting it from the other directive. 

Unit tests in Nginx are a bit hard. Happy to add that on in a follow-up PR to preserve squashed authorship, would appreciate results of a test run on a sample config though.

Dreamsorcerer:
OK, going back to the previous idea. I've just reset the index when we remove an element, so it searches the full directive again. Obviously, less efficient than the previous attempts, but that shouldn't make a noticeable difference considering the code is only run on listen directives, and that a typical directive is only 2-5 elements long.

Travis-ci seems to be broken. Half of the jobs have not started, and the lint test has failed (but, I don't think that's caused by me).

Dreamsorcerer:
OK, managed to actually find the lint error in that output. Fixed that.

Also, I take back the comment about it being inefficient, as most directives won't have these keywords, so most times it will only iterate through the list once (compared to the original code which was already iterating 3 times, once for each 'in' keyword).

Dreamsorcerer:
@ohemorange Is this OK to merge now?

Dreamsorcerer:
@ohemorange Like this?

Issue #6236, opened 2018-07-17
Don't use --numprocesses auto in sudo: false builds


bmw:
Building off of #6197, it looks like `pytest-xdist` has a hard time determining the number of cores when it is run in any Travis instance with `sudo` set to false in the Travis config. See https://travis-ci.org/certbot/certbot/jobs/404581691#L813 for an example.

To fix this, I think we should configure `tox.ini` to pass through either `TRAVIS_SUDO` or `TRAVIS` for all environments running pytest and automatically use `--numprocesses 2` if the variable is set to the appropriate value. While `TRAVIS_SUDO` may be more accurate and allows us to take advantage of additional cores if they are added to other environments, all [Travis environments](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System) use two cores, have for years, and doing this for all environments protects us from hitting this issue again if changes in `pytest-xdist` or Travis prevent it from detecting the number of cores in another environment type. I don't have a strong preference which approach is taken.

@greut, thanks for your initial PR working on this. Are you at all interested in working on this as well?

greut:
at this point, it's probably easier to fix pytest-xdist.

greut:
https://github.com/pytest-dev/pytest-xdist/pull/309

bmw:
Nice work in pytest-dev/pytest-xdist#309 @greut.

Once that lands, are you interested in making a PR to revert your previous change to our tests and update our pinnings in `tools/dev_constraints.txt` to use the new version of `pytest-xdist`? Pinnings for `pytest-xdist`'s dependencies may need to be updated as well.

greut:
I'll do that tomorrow morning then, cheers,

bmw:
Fixed in #6253.

PR #6253, opened 2018-07-24
deps: bump pytest-xdist to 1.22.4

greut:
Latest pytest-xdist knows how to properly read the number of processes given by Travis.

cf #6236, reverts #6211

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -64,6 +64,7 @@ source_paths =
     tests/lock_test.py
 
 [testenv]
+passenv = TRAVIS
 commands =
     {[base]install_and_test} {[base]all_packages}
     python tests/lock_test.py
@@ -121,7 +122,6 @@ commands =
 
 [testenv:cover]
 basepython = python2.7
-passenv = NUMPROCESSES
 commands =
     {[base]install_packages}
     ./tox.cover.sh
@@ -166,7 +166,9 @@ commands =
     docker run --rm -it apache-compat -c apache.tar.gz -vvvv
 whitelist_externals =
     docker
-passenv = DOCKER_*
+passenv =
+    DOCKER_*
+    TRAVIS
 
 [testenv:nginx_compat]
 commands =
@@ -175,7 +177,9 @@ commands =
     docker run --rm -it nginx-compat -c nginx.tar.gz -vv -aie
 whitelist_externals =
     docker
-passenv = DOCKER_*
+passenv =
+    DOCKER_*
+    TRAVIS
 
 [testenv:le_auto_precise]
 # At the moment, this tests under Python 2.7 only, as only that version is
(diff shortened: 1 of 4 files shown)


greut:
:confused: 

```
plugins: xdist-1.22.3, forked-0.2, cov-2.5.1
gw0 [877] / gw1 [877] / gw2 [877] / gw3 [877] / gw4 [877] / gw5 [877] / gw6 [877] / gw7 [877] / gw8 [877] / gw9 [877] / gw10 [877] / gw11 [877] / gw12 [877] / gw13 [877] / gw14 [877] / gw15 [877] / gw16 [877] / gw17 [877] / gw18 [877] / gw19 [877] / gw20 [877] / gw21 [877] / gw22 [877] / gw23 [877] / gw24 [877] / gw25 [877] / gw26 [877] / gw27 [877] / gw28 [877] / gw29 [877] / gw30 [877] / gw31 [877] / gw32 [877] / gw33 [877] / gw34 [877] / gw35 [877] / gw36 [877] / gw37 [877] / gw38 [877] / gw39 [877] / gw40 [877] / gw41 [877] / gw42 [877] / gw43 [877] / gw44 [877] / gw45 [877] / gw46 [877] / gw47 [877]
```

greut:
Python 2.7, of course.

bmw:
Ah. I didn't realize `os.sched_getaffinity` was Python 3 only. We could move some tests over to Python 3, but we can't eliminate Python 2.7 tests right now.

We may have a solution when https://github.com/pytest-dev/pytest-xdist/pull/317 is released though.

tox.ini
```diff
@@ -121,7 +121,7 @@ commands =
 
 [testenv:cover]
 basepython = python2.7
-passenv = NUMPROCESSES
+passenv = TRAVIS
```
bmw: In addition to setting this here, I think we should set it in the `[testenv]` environment and inherit it in all of the `[testenv:py27-*oldest]` environments like we do for `setenv`. This should cause the variable to be passed through for all relevant Python 2.7 tests.

greut:
bumping pytest leads to all sorts of trouble with `tox -e py27-oldest`

bmw:
> bumping pytest leads to all sorts of trouble with tox -e py27-oldest

Looks like they require a newer version `six` than is available for us for our CentOS 7 packaging. Can we just bump `pytest-xdist` and not `pytest`?

greut:
@bmw it requires `pytest>=3.3`

bmw:
I think this requirement was removed in `pytest-xdist` 1.22.5. See https://github.com/pytest-dev/pytest-xdist/pull/321.

greut:
yay :+1:

PR #5940, opened 2018-05-09
Feature/azure dns

mattkenn4545:
This PR is https://github.com/certbot/certbot/pull/522 with the acme version bumped to work with the latest version on master.

If there is anything I can do to get this merged let me know.  I've tested it as much as I can and its working like a champ.

PR #5560, opened 2018-02-09
Add Azure DNS auth [WIP]

elmariofredo:
This is Work In Progress PR inherited from @eduvik I have to update/add tests same way as are in [dns cloudflare](https://github.com/certbot/certbot/blob/master/certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare_test.py)

Issue #5521, opened 2018-01-31
Certbot doesnt create acme-challenge file


xgaia:
Hello, I tried to renew my certificate with certbot-auto, but it failed. It seems to not create the acme files.

## My operating system is (include version):

 Raspbian GNU/Linux 8 (jessie)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

 certbot-auto

## I ran this command and it produced this output:

command: 

```bash
certbot-auto renew 
```
output:

```
Saving debug log to /var/log/letsencrypt/letsencrypt.log

-------------------------------------------------------------------------------
Processing /etc/letsencrypt/renewal/<mydomain>.conf
-------------------------------------------------------------------------------
Cert is due for renewal, auto-renewing...
Plugins selected: Authenticator apache, Installer apache
Renewing an existing certificate
Performing the following challenges:
http-01 challenge for <mydomain>
Waiting for verification...
Cleaning up challenges
Attempting to renew cert (<mydomain>) from /etc/letsencrypt/renewal/<mydomain>.conf produced an unexpected error: Failed authorization procedure. <mydomain> (http-01): urn:acme:error:connection :: The server could not connect to the client to verify the domain :: Fetching http://<mydomain>/.well-known/acme-challenge/zebrJ7qQ5k9lixT0Cwg-9wEaPSY-mL4usZkHeZIT8_k: Timeout. Skipping.
All renewal attempts failed. The following certs could not be renewed:
  /etc/letsencrypt/live/<mydomain>/fullchain.pem (failure)

-------------------------------------------------------------------------------

All renewal attempts failed. The following certs could not be renewed:
  /etc/letsencrypt/live/<mydomain>/fullchain.pem (failure)
-------------------------------------------------------------------------------
1 renew failure(s), 0 parse failure(s)

IMPORTANT NOTES:
 - The following errors were reported by the server:

   Domain: <mydomain>
   Type:   connection
   Detail: Fetching
   http://<mydomain>/.well-known/acme-challenge/zebrJ7qQ5k9lixT0Cwg-9wEaPSY-mL4usZkHeZIT8_k:
   Timeout

   To fix these errors, please make sure that your domain name was
   entered correctly and the DNS A/AAAA record(s) for that domain
   contain(s) the right IP address. Additionally, please check that
   your computer has a publicly routable IP address and that no
   firewalls are preventing the server from communicating with the
   client. If you're using the webroot plugin, you should also verify
[...]

PR #6261, opened 2018-07-26
Add Azure DNS Authentication

dlapiduz:
There are currently 3 Azure DNS auth PRs open (#5940, #5560, and #5521) which were proposed but never got traction because of a lack of tests.
This PR builds on top of #5940 (which builds on top #5221) and adds tests based on the cloudflare suite @joohoi suggests in #5221.

A couple things to note:
- I updated `requests` to 2.16.5 to make it compatible with `msrest`.
- I tested the plugin manually just to make sure it worked with Azure live and not only using unit tests

Please let me know if you want any changes to the code or the tests but I'd love to see this merged so we can have Azure DNS as an auth mechanism.


joohoi:
I'm answering here for question in #6262

> I updated requests to 2.16.5 to make it compatible with msrest

The dependency mismatch is a very unfortunate issue, and I personally don't have ideas how to mitigate this. If I recall, the code in original PR was working, so could there be any way to not support the version of `msrest` that requires another requests version, or can we pin another version of `azure-mgmt-dns` which doesn't have this issue by any chance?

dlapiduz:
@joohoi this is an unfortunate issue indeed and I am not 100% sure how it worked in the past.

I believe that previous PRs worked fine because they didn't have test and Tox wasn't installing the package in CI. Probably contributors just updated `requests` locally and it made it work.

A couple data points that I hope help:
- I asked `msrest` to downgrade to 2.12 and they basically said no because 2.16 helped them with another dependency issue (go figure...): https://github.com/Azure/msrest-for-python/issues/118
- The previous version from `msrest` that is compatible from 2.12 is over a year old and I believe incompatible with the existing Azure APIs (https://github.com/Azure/msrest-for-python/commit/166413e3dffcd740719bc4c270e1f4e4605c6ad7)
- `requests` 2.16 has been around for a bit over a year and the current version is 2.19 (https://pypi.org/project/requests/2.16.0/). Do you have a backward compatibility limit?

Is there any chance we can just update to 2.16?

Thanks!

dlapiduz:
@bmw that sounds like a great idea. Let me take a crack at it here and I'll ping you back to see what you think...

dlapiduz:
@bmw I am taking a deeper dive into it and I see 2 issues:
- Changing the way that we create the requirements doesn't change the way Tox runs the tests. Shouldn't I build some condition in [tox.ini](https://github.com/certbot/certbot/blob/master/tox.ini) that if `$CERTBOT_OLDEST` is set that we don't run the azure-dns tests?
- If we want to have `certbot-dns-azure` to remain unpackaged, What does it mean exactly? I am not sure I know where the DNS are packaged other than when installing them in the dev environment...

bmw:
> Changing the way that we create the requirements doesn't change the way Tox runs the tests. Shouldn't I build some condition in tox.ini that if $CERTBOT_OLDEST is set that we don't run the azure-dns tests?

Changing the requirements should change the packages used when tox runs tests on this plugin. By changing the requirements to versions that satisfy all dependencies, tests should pass when tox runs tests.

> If we want to have certbot-dns-azure to remain unpackaged, What does it mean exactly? I am not sure I know where the DNS are packaged other than when installing them in the dev environment...

This doesn't mean much technically for your PR. It's just the reason breaking our normal practices and requiring a newer version of `requests` in `local-oldest-requirements.txt` is OK.

What it means outside of this repo though is many Linux distros won't be able to package this plugin because of how strict its dependency requirements are. We can still put the plugin up on PyPI, Docker Hub, and the more bleeding edge distros will be able to package it.

dlapiduz:
> Changing the requirements should change the packages used when tox runs tests on this plugin. By changing the requirements to versions that satisfy all dependencies, tests should pass when tox runs tests.

I am not a tox expert but isn't this https://github.com/certbot/certbot/blob/master/tox.ini#L76 going to run https://github.com/certbot/certbot/blob/master/tools/install_and_test.sh#L20 which would install the package and run `pip install certbot-dns-azure` and `pytest ... certbot_dns_azure`. That is what I think is happening here: https://travis-ci.org/certbot/certbot/jobs/408774623#L699

Even if we remove the package in the requirements file, this will still try to run because it is part of the tox.ini file.

Am I missing something? Let me send a quick commit on what I am thinking...

bmw:
Sorry. I wasn't very clear on how to modify the contents of `local-oldest-requirements.txt`.

Your latest commit just removes Azure from the `py27-oldest-tests` which I don't think we want to do. While we don't need to worry about keeping compatibility with the versions of our dependencies available in LTS Linux distros, these tests also helps make sure the versions specified in `setup.py` are accurate. Usually this file just specifies the oldest supported `acme` and `certbot` version which is then installed and used during tests. If the plugin no longer works with these older versions, the tests should fail.

For this plugin, I think we also want to specify the oldest version of libraries like `requests` we work with. By changing constraints/requirements files in the way I described above, we can cause the test to use whatever version of `requests` we want and pass.

bmw:
Also to be clear, we don't need to pin every dependency of this plugin in the oldest tests, but we should pin any dependencies whose minimum version we specify and those we need newer versions of than is specified in `tools/oldest_constraints.txt`.

dlapiduz:
@bmw I think I got it now. Does this make sense? https://github.com/certbot/certbot/pull/6261/files#diff-933ede83aa60635da918f2fe6d64f265R33

The idea is to add a `local-oldest-constraints.txt` file instead (or in addition) of the `local-oldest-requirements.txt` to specify constraints to use when running this test suite. The problem with using the requirements file was that `merge_requirements.py` is not fully compatible with all the possible syntax of the requirements.txt file and it would throw error on some other packages.

Please let me know what you think and if I should add something else to get this through.

bmw:
> The problem with using the requirements file was that merge_requirements.py is not fully compatible with all the possible syntax of the requirements.txt file and it would throw error on some other packages.

Nice catch! Sorry for the bad suggestion there. `merge_requirements` exists because to my knowledge, `pip`'s behavior here isn't documented so I wrote this simple script to ensure consistency, but I didn't take the time to make it parse all the different specifiers and flags understood by `pip`.

The only thing I'd potentially like to see changed around this oldest stuff is the packages listed in `local-oldest-constraints.txt`. Since we kept `local-oldest-requirements.txt`, I think we should remove `acme` and `certbot` as it'll otherwise just be another file we'll have to remember to keep in sync.

I'd also tweak the other requirements a bit. For `azure-mgmt-dns`, I think we should either specify `>=2.0.0` in `setup.py` or remove it from the constraints.

For the other packages, I'd test with the absolute oldest version supported. For instance, for `requests`, I'd either use `2.16.0` in the constraints file or put `2.16.5` in `setup.py`. I'd do the same thing for `urllib3` and `chardet` and maybe add a comment saying that these constraints are coming from `requests`.

dlapiduz:
@bmw thanks for the tips! Here is what I did:
- Specified `>=2.0.0` for `azure-mgmt-dns` in `setup.py`
- Use `2.16.0` for requests and the oldest version for `urlib3` and `chardet`
- Added the comment regarding requests constraints
- Moved `acme` to `local-oldest-requirements` so it matches all the other packages (without this the test didn't run properly)

Please let me know if that makes sense!

bmw:
Looks like I overlooked something again on my end. In my previous comment when I was worried about duplicating the `acme` and `certbot` constraints, I didn't notice you had previously completely eliminated `local-oldest-requirements.txt`. Whether you move them back to `local-oldest-constraints.txt` or not is up to you, but wherever we specify these versions, I think we should be pinning `acme` and `certbot` `0.21.1` to match the plugin's `setup.py`. Sorry about that!

Otherwise, the oldest tests stuff looks good to me. Thanks for fixing it up @dlapiduz!

dlapiduz:
> but wherever we specify these versions, I think we should be pinning `acme` and `certbot` `0.21.1` to match the plugin's `setup.py`

Argh, I should had caught that. I just updated the commit with those versions.

dlapiduz:
@bmw how do we proceed now?

bmw:
This PR needs a full review. @joohoi or @zjs, do you have time to take a look?

dlapiduz:
@joohoi @zjs please let me know if you need anything to review the PR! thx

schoen:
@zjs, would you have time and be willing to look at a few Certbot DNS authenticator PRs, including this one? I think there are a total of four that people have sent in recently.

dlapiduz:
@schoen @zjs Is there anything we can do to help with the process? maybe cross review authenticators?

dlapiduz:
@zjs @joohoi @bmw is there anything we can do push this? I'd love to see it through. Thanks!

sydneyli:
Hi @dlapiduz, sorry for the delay here! @bmw has agreed to review this PR, so I've assigned it to him. He's going to be out for the next couple of weeks, but will be able to take a look when he gets back!

bmw:
It's waiting on review from us. [Testing](https://github.com/certbot/certbot/issues/6464) and [reviewing and maintaining](https://github.com/certbot/certbot/issues/6504) all of the DNS plugins is a lot for the small Certbot team to take on.

We haven't forgotten about this or any of our open PRs and will try to get to it as soon as we can.

xtremeperf:
Hey guys...      @bmw @dlapiduz

I just want to let you know that I've been running this in production on a small, multi-cloud network for a year now and it's been absolutely flawless thus far.  I ran the original author's code for about 9 months, and I've been running the updates provided by @dlapiduz for about 3 months now. I reviewed the code pretty closely and didn't find any major issues with it myself.  We run it daily, scripted, in a docker container using deploy hooks to distribute any renewed certs to the various hosts.  

I know it's only one case, but I've noticed concerns posted in other conversations regarding the difficulties in being able to confirm working status of DNS plugins against the provider's actual live APIs... so I figured I would take a moment to share my results with you guys, in case it may be of some value during your review process.  If there's anything I can do to assist, feel free to let me know.

Thanks,
Doug

dlapiduz:
> It's waiting on review from us.

@bmw is there anything else I can do to help with the review? I'd love to see this merged.
Thanks!

mstansberry:
bump for this feature

ClaudiaBaur:
@bmw, any chance to get this feature in in near future?
Thanks!

bmw:
While I'd still like us to do something in the long term, I don't think there will be any progress on this PR in the short term. Reviewing, testing, and maintaining plugins for all of these DNS providers has become quite hard for the few people working on Certbot and we need to redesign how we approach these problems in general as described in issues like #6464, #6503, and #6504.

In the meantime, people are welcome to release this as 3rd party plugin like others have done. See https://github.com/coldfix/certbot-dns-netcup for one example of that.

Mizipzor:
I just tried, and failed, to get this running. It seems the acme dependency has changed, tried downgrading but that didnt work (1.6.0 was installed, tried 1.5.0 and 1.4.0). 

```
$ python3 -m virtualenv .env
$ . .env/bin/activate
$ pip3 install .
$ certbot --help
  File "/home/afast/git/certbot/.env/bin/certbot", line 5, in <module>
    from certbot.main import main
  File "/home/afast/git/certbot/.env/lib/python3.7/site-packages/certbot/main.py", line 18, in <module>
    from certbot import account
  File "/home/afast/git/certbot/.env/lib/python3.7/site-packages/certbot/account.py", line 20, in <module>
    from certbot import constants
  File "/home/afast/git/certbot/.env/lib/python3.7/site-packages/certbot/constants.py", line 53, in <module>
    tls_sni_01_port=challenges.TLSSNI01Response.PORT,
AttributeError: module 'acme.challenges' has no attribute 'TLSSNI01Response'
```

Could try to merge this feature branch with master, but the conflicts are a few too many.

@dlapiduz I would prefer to have this split off as a standalone module, as suggested in the last comment. Is this possible? It will be my next attemp, but I want to give proper credit for the source.

dlapiduz:
@bmw @Mizipzor I pushed this code to the [`certbot-azure`](https://github.com/dlapiduz/certbot-azure) plugin that I had as an installer. I was looking forward to have this in the canonical certbot client but I guess it is not possible anymore.

I am going to go ahead and close the PR now, thanks for your work and your patience!

PR #6267, opened 2018-07-30
Add malformed body check

### LINKED ISSUE #6092: Use the ACMEv2 newNonce endpoint when a new nonce is needed
In our testing, sometimes Akamai fails to reach Let's Encrypt, and returns a 500 with malformed (non-JSON) contents. Certbot reports this as a missing nonce; instead it should report the malformed body and ignore the lack of a nonce.

doug-skinner:
Re-opening under a new name to re-trigger travis-ci which failed due to an external error. This pull request aims to fix #6092

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -1103,6 +1103,9 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
                 raise errors.BadNonce(nonce, error)
             logger.debug('Storing nonce: %s', nonce)
             self._nonces.add(decoded_nonce)
+        # Handles the case where Akamai fails to reach Let's Encrypt
+        elif response.status_code == 500:
+            raise errors.MalformedBodyError()
         else:
             raise errors.MissingNonce(response)
 
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -973,6 +973,8 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
 
         :raises .messages.Error: If server response body
             carries HTTP Problem (draft-ietf-appsawg-http-problem-00).
+        :raises .MalformedBodyError: If the server sends back a 503
+            indicating that the Akamai server failed to reach Let's Encrypt.
         :raises .ClientError: In case of other networking errors.
 
         """
@@ -986,6 +988,9 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
 
         if response.status_code == 409:
             raise errors.ConflictError(response.headers.get('Location'))
+
+        if response.status_code == 503:
+            raise errors.MalformedBodyError()
 
         if not response.ok:
             if jobj is not None:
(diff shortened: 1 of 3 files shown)


schoen:
I guess this is challenging to live-test because we can't really generate this condition on the live CA, although perhaps for the integration test we could force the test instance of Boulder to generate an error condition...?

@ohemorange, would you like to look at this? Shall we wait until @jsha gets back from vacation?

doug-skinner:
Yes, it was challenging to think of a way to live test this. If you have any ideas, just let me know and I'll update the PR

doug-skinner:
Okay, I'll be leaving this until there's a bit more information

doug-skinner:
@ohemorange Is there any timeline on when there will be an update to this?

cpu:
>  @rolandshoemaker or @cpu, do you have any insight into this while jsha's out?

Sorry, missed this mention initially. Is there more context here? What is the problem being addressed? 

I see in the diff "# Handles the case where Akamai fails to reach Let's Encrypt" - is there an exact HTTP response you can share from this case that sent you down this road? Are you specifically trying to handle a case where an Akamai edge node isn't able to communicate to the Let's Encrypt origin servers?

doug-skinner:
So I was working off of #6092, where it states that after failing to reach the server a 500 is returned with malformed contents. So this PR is to add another check that errors on that and gives a better message to the user, rather than error that there isn't a nonce in the body of the response as there won't be with malformed contents.

cpu:
> So I was working off of #6092, where it states that after failing to reach the server a 500 is returned with malformed contents.
@doug-skinner Ahhhh!!! Ok, I think I remember what was going on when Jsha filed #6092. Thanks for your patience.

> So this PR is to add another check that errors on that and gives a better message to the user, rather than error that there isn't a nonce in the body of the response as there won't be with malformed contents.

:+1: - makes sense.

> Also, this may require a full log from somewhere, but is it exactly a 500 we receive?

I think we might have logs for this. I'll ask. I think it's likely an exact HTTP Status 500 response code but I'm not sure what additional response headers are present or what the body is.

> And if it is, do we get that error code for any other type of response? I don't want to be returning a Certbot error that's more specific than the html error code we got from the server. 

@ohemorange in the normal 500 server internal cases where Boulder itself has had a problem we would return a more structured 500 that would include a nonce in the response headers & a JSON problem document in the response's body. I think this kind of 500 returned by a generic webserver with no ACME semantics in the response is likely to be pretty unique/rare.

cpu:
> Also, this may require a full log from somewhere, but is it exactly a 500 we receive?

@ohemorange @doug-skinner Good news - we had logs from an instance on June 19th. I've [updated issue 6092 with the log](https://github.com/certbot/certbot/issue_comments#issuecomment-411828011) and updated my earlier comment to reflect its a 503 specifically.

doug-skinner:
Thanks for the info! I'll look it over and update the branch either tonight or tomorrow morning

doug-skinner:
The first Jenkins error looks like just a networking thing, unable to download a dependency through apt. Ran into that the first time I submitted to. Looking into the second now though, running ```tox --skip-missing-interpreters``` again. I'll have an update later today

doug-skinner:
@cpu sorry for not tagging you originally, but I think the latest commit handles the previously brought up concerns.

cpu:
> @cpu sorry for not tagging you originally, but I think the latest commit handles the previously brought up concerns.

I think you'll want to tag @ohemorange or @bmw for this. I'm not the best person to give this PR a review. Sorry!

bmw:
Thanks for working on this @doug-skinner.

ohemorange is a bit swamped with other work at the moment, but this is on her radar and she's planning take another look at this next week.

ohemorange:
Hey @doug-skinner, are you no longer able to continue working on this?

ohemorange:
Closing due to inactivity, feel free to reopen.

PR #6268, opened 2018-07-30
Enable Apache VirtualHost for HTTP challenge validation if not enabled already

joohoi:
If user provides a custom `--apache-vhost-root` path that's not parsed by Apache per default, Certbot fails the challenge validation. While the `VirtualHost` on custom path is correctly found, and edited, it's still not seen by Apache. This PR adds a temporary `Include` directive to the root Apache configuration when writing the challenge tokens to the `VirtualHost`.

### DIFF
--- a/certbot-apache/certbot_apache/http_01.py
+++ b/certbot-apache/certbot_apache/http_01.py
@@ -172,4 +172,10 @@ class ApacheHttp01(common.TLSSNI01):
             self.configurator.parser.add_dir(
                 vhost.path, "Include", self.challenge_conf_post)
 
+            if not vhost.enabled:
+                from certbot_apache.parser import get_aug_path
+                self.configurator.parser.add_dir(
+                    get_aug_path(self.configurator.parser.loc["default"]),
+                    "Include", vhost.filep)
+
             self.moded_vhosts.add(vhost)
--- a/certbot-apache/certbot_apache/http_01.py
+++ b/certbot-apache/certbot_apache/http_01.py
@@ -6,6 +6,7 @@ from acme.magic_typing import Set  # pylint: disable=unused-import, no-name-in-m
 from certbot import errors
 from certbot.plugins import common
 from certbot_apache.obj import VirtualHost  # pylint: disable=unused-import
+from certbot_apache.parser import get_aug_path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


certbot-apache/certbot_apache/http_01.py
```diff
@@ -172,4 +172,10 @@ def _set_up_include_directives(self, vhost):
             self.configurator.parser.add_dir(
                 vhost.path, "Include", self.challenge_conf_post)
 
+            if not vhost.enabled:
+                from certbot_apache.parser import get_aug_path
```
bmw: Can we move this import to the top of the file?

certbot-apache/certbot_apache/tests/http_01_test.py
```diff
@@ -134,6 +135,17 @@ def test_perform_3_achall_apache_2_2(self):
...
 
+    def test_activate_disabled_vhost(self):
+        vhosts = [v for v in self.config.vhosts if v.name == "certbot.demo"]
+        achalls = [
+            achallenges.KeyAuthorizationAnnotatedChallenge(
+                challb=acme_util.chall_to_challb(
+                    challenges.HTTP01(token=((b'a' * 16))),
+                    "pending"),
+                domain="certbot.demo", account_key=self.account_key)]
+        vhosts[0].enabled = False
```
bmw: I'm not sure what's going on here, but if I set this value to `True`, tests still pass. This seems bad. Do you know why this is happening?

certbot-apache/certbot_apache/http_01.py
```diff
@@ -172,4 +172,10 @@ def _set_up_include_directives(self, vhost):
...
+                self.configurator.parser.add_dir(
```
bmw: While not directly related to this PR, is the reason you used this rather than calling `self.configurator.parser.add_include` because we don't want to add this path to the `existing_paths` list in the parser? If so, is our usage of it in TLS-SNI a bug?

bmw:
Closing and reopening for science.

certbot-apache/certbot_apache/http_01.py
joohoi: Yes, not adding the path to `existing_paths` is the reason. And you are right, it is a bug in our TLS-SNI validation. However I'm pretty confident that it doesn't manifest itself in any way in case of TLS-SNI, as the validation `ServerName` is very unlikely matching anything we search.

certbot-apache/certbot_apache/tests/http_01_test.py
joohoi: The actual test assert is now moved to this test function.

joohoi:
Ready for the next round

Issue #5857, opened 2018-04-12
ACMEv2 client includes "status" field in newOrder JWS payload


cpu:
## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

Cloned from git:
```
$> git rev-parse HEAD
6b29d159a2f221c3437770bdb43924ee6f953c4b
```

## I ran this command and it produced this output:

`certbot_test --server http://localhost:4001/directory certonly --standalone -d one.wtf --preferred-challenges http-01 `

## Certbot's behavior differed from what I expected because:

The JWS payload sent to the `newOrder` endpoint contained two fields not mentioned in modern ACME drafts (e.g. draft-10, draft-11, "ACME v2"): "resource" and "status".

```
{
  "status": "pending", 
  "identifiers": [
    {
      "type": "dns", 
      "value": "one.wtf"
    }
  ], 
  "resource": "new-order"
}
```

The first unexpected field, "resource", is already captured in https://github.com/certbot/certbot/issues/5708

The second unexpected field, "status", is something assigned by the server in the returned order object and not a field that should appear in the client's newOrder request. [Section 7.4 of draft-11](https://tools.ietf.org/html/draft-ietf-acme-acme-11#section-7.4) only specifies the following fields for a newOrder request: "identifiers", "notBefore" and "notAfter". Because Let's Encrypt does not support the latter two fields I would expect a newOrder JWS body to look like:
```
{
  "identifiers": [
    {
      "type": "dns", 
      "value": "one.wtf"
    }
  ], 
}
```

## Here is a Certbot log showing the issue (if available):
```
<snip>

JWS payload:
{
  "status": "pending", 
  "identifiers": [
    {
      "type": "dns", 
      "value": "one.wtf"
    }
  ], 
  "resource": "new-order"
}
Sending POST request to http://localhost:4001/acme/new-order:
{
  "protected": "eyJub25jZSI6ICJEYjVYUWNIWGJ3MmVwdHNUVVJJZkpwN0ZScUl2RlVsQWJlQm80VW1qb25rIiwgInVybCI6ICJodHRwOi8vbG9jYWxob3N0OjQwMDEvYWNtZS9uZXctb3JkZXIiLCAia2lkIjogImh0dHA6Ly9sb2NhbGhvc3Q6NDAwMS9hY21lL2FjY3QvMjE0MSIsICJhbGciOiAiUlMyNTYifQ", 
[...]

Issue #5708, opened 2018-03-12
ACMEv2 client includes resource field in JWS payload


bmw:
Relevant log is:
```
JWS payload:
{
  "termsOfServiceAgreed": true, 
  "resource": "new-reg"
}
Sending POST request to https://acme-staging-v02.api.letsencrypt.org/acme/new-acct:
{
  "protected": "eyJub25jZSI6ICJUMDN3d195UjcxV2haN2ZsbGFHVlVQUF8tejkxcVhKQ1FFaFFwQWp6eTNNIiwgInVybCI6ICJodHRwczovL2FjbWUtc3RhZ2luZy12MDIuYXBpLmxldHNlbmNyeXB0Lm9yZy9hY21lL25ldy1hY2N0IiwgImp3ayI6IHsiZSI6ICJBUUFCIiwgImt0eSI6ICJSU0EiLCAibiI6ICI2QnlTWHVwbnpHU1Y0ME53UHhqRTdSdEp3cFlvNnJyUWNEX0xOMm94RFIzTFQ0N0JjZ1AySHhXR2xSTDV1YWtjWHJWaHlScjd4aUJkNDUzSnBvU2VxR1RnS0tKVVp2eDVnd09tNzg4SE9pSEJDdklva0ljNzBjN3JQTmFrR0tZdXp6SVlUTmR4b2w3dWdSaVhxdUZOQnRZeXlLeDRnVUx6Y1QzYmh4dVdfeldISC1HSGVoNXA4Qjg1TE9KZEx5WlV2djNjQWxYLUYtYmJJX3VqRm5rc0ZhVTI3Z2s0RFlldllpRjN0SEtqZXQ4UXlWeTYwRk1wRFdkaUZhZTJVeklFNXZiczRPX1FvVTBZVWZfbTVMbHFQVmF0TWk0VDVFdFVOZkJGYWp2U0FBcTdoVkg4bmJhbzdVMVVQVVJsQVFZc2otSVFaNlFoblBVWExRZjMtVGozTXcifSwgImFsZyI6ICJSUzI1NiJ9", 
  "payload": "ewogICJ0ZXJtc09mU2VydmljZUFncmVlZCI6IHRydWUsIAogICJyZXNvdXJjZSI6ICJuZXctcmVnIgp9", 
[...]

PR #6271, opened 2018-07-31
Do not send status or resource fields in newOrder payloads for ACMEv2

joohoi:
Remove unnecessary fields `status` and `resource` from ACMEv2 newOrder endpoint payloads.

Fixes: #5708
Fixes: #5857


bmw:

joohoi:
Oh, right. I thought it was still used for something internally for v1 because a `messages.Order` object is still used from [messages.OrderResource](https://github.com/certbot/certbot/blob/master/acme/acme/messages.py#L547) for parsing the json, called from [BackwardsCompatibleClientV2.new_order](https://github.com/certbot/certbot/blob/master/acme/acme/client.py#L820) when the server we're communicating with is ACMEv1. I'll change the `messages.Order` class directly.

acme/acme/messages.py
```diff
@@ -522,8 +521,6 @@ class Order(ResourceBody):
     :ivar .Error error: Any error that occurred during finalization, if applicable.
     """
     identifiers = jose.Field('identifiers', omitempty=True)
-    status = jose.Field('status', decoder=Status.from_json,
```
ohemorange: Although we don't use the returned status, sounds like an `Order` object can come back with a status inside it, which we should retain the ability to parse. So we just want to get the default out I think so that it doesn't fill it in for a `NewOrder`. This is based on my understanding of #5857; if you interpreted that differently or think that we will be able to successfully parse a response that contains a status even without that line, let me know.

joohoi: This is what the initial changes were supposed to handle. Do you feel like populating the `messages.Order` object with the `status` field  when called for parsing JSON payload would be the correct way to take care of this?

ohemorange: We should populate it on parse so we can find out the status. We shouldn't be posting it back with a status.

bmw:
@ohemorange is going to take this over.

Issue #6276, opened 2018-08-01
certbot certificates should explain why it thinks configurations are invalid


jsoref:
If you're having trouble using Certbot and aren't sure you've found a bug or
request for a new feature, please first try asking for help at
https://community.letsencrypt.org/. There is a much larger community there of
people familiar with the project who will be able to more quickly answer your
questions.

## My operating system is (include version):
Ubuntu 16.04.4 LTS (xenial)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
apt

## I ran this command and it produced this output:
```sh
$ certbot certificates
...
The following renewal configuration files were invalid:
  /etc/letsencrypt/renewal/mail.test.example.com.conf
```

## Certbot's behavior differed from what I expected because:
1. It should have told me *why* it was invalid -- the debug output for the renewal clearly explains that the files were expected to be symlinks.
1. It shouldn't say that the `configuration file is invalid`, it should say that the `configuration is invalid` -- the file itself is totally reasonable and anyone using `diff` to compare configuration files will be barking up the wrong tree.

###### Logs are stored in `/var/log/letsencrypt` by default. Feel free to redact domains, e-mail and IP addresses as you see fit.
```
certbot.errors.CertStorageError: expected /etc/letsencrypt/live/mail.test.example.com/cert.pem to be a symlink
2018-07-23 09:37:06,826:WARNING:certbot.renewal:Renewal configuration file /etc/letsencrypt/renewal/mail.test.example.com.conf is broken. Skipping.
```

see also #2550

Kelderic:
I just ran into this issue. I upgraded hardware and moved files wholesale. Apparently, the symlinks got broken. I spent a half hour trying to figure out why the "Configuration" was invalid.

A better error message would be to say that the symlinks aren't valid.

PR #6277, opened 2018-08-01
Clarify that configurations can be invalid

jsoref:
A configuration file will typically point to files...
If those files are supposed to be symbolic links but are not,
then the configuration is invalid.

Any normal person upon being told that a configuration file is
invalid would compare the configuration file to another
configuration file, and not spot anything particularly wrong with it.

The problem in that case is the configuration, which includes the
configuration file and everything it references.
-- NOT the configuration file itself.

#6276

This only addresses half of the problem...
I've looked a little bit into the main part (reporting to the user in the certificates listing), and that appears to require significantly more effort.

### DIFF
--- a/certbot/cert_manager.py
+++ b/certbot/cert_manager.py
@@ -353,7 +353,7 @@ def _describe_certs(config, parsed_certs, parse_failures):
             notify("Found the following {0}certs:".format(match))
             notify(_report_human_readable(config, parsed_certs))
         if parse_failures:
-            notify("\nThe following renewal configuration files "
+            notify("\nThe following renewal configurations "
                "were invalid:")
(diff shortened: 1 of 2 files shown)


bmw:
@ohemorange, do you have an opinion here?

ohemorange:
Yes, my opinion is that the correct solution involves making it clear that these files are not meant to be modified, along with better offline documentation for people who want to try to do custom setups.

bmw:
@ohemorange, is your comment about this PR in particular or this problem in general? I probably wasn't very clear, but I was asking about the former.

ohemorange:
Honestly I looked at this page and thought this was an issue not a PR. I like this PR and think it improves the message.

Issue #5582, opened 2018-02-16
Feature request: certbot-dns-gandi


toto-jy:
The existing Gandi plugin at https://github.com/Gandi/letsencrypt-gandi only supports their v4 API and is deprecated and archived, with Gandi claiming it's not needed because they offer free SSL certificates with their hosting package. This leaves people hosting elsewhere but using Gandi DNS in the cold when it comes to setting up certbot to automate DNS verification.

However, Gandi now has a v5 API (docs at http://doc.livedns.gandi.net) and it would be great if certbot could update its DNS entries through this.

(Sadly, my Python is not currently up to writing the plugin myself and submitting a PR :)

Thanks!

joohoi:
If someone wants to pick this up, we certainly can provide pointers where to look at and how to get going.

maport:
I've just built a plugin for this which is available at: https://gitlab.com/cspublic/certbot-plugin-gandi.

This calls the LiveDNS API directly although it appears support for the new API is coming in Lexicon (see https://github.com/AnalogJ/lexicon/issues/171).

tehfink:
@maport : Very cool, thanks! Would you mind making it available on PyPi?

maport:
@tehfink: Hopefully at some point, but not sure when currently. Out of interest, do you have a particular requirement for it to be on PyPi (i.e. can you not use `pip install 'git+https://gitlab.com/cspublic/certbot-plugin-gandi.git'`)?

reassured-jy:
Unfortunately the maintainers of the nginx/LE Docker image I use only support plugins that are part of the certbot repo — a pain for me, but I guess they want to keep maintenance simple.

@maport Would it be possible for you to submit this as a PR for certbot?

joohoi:
Can confirm. Using lexicon makes it easier to implement, review and eventually to package by our various package maintainers.

adferrand:
Yep, my PR for Gandi LiveDNS on lexicon has been merged few days ago. Be aware that the Gandi provider supports both APIs, and the old one is used by default. The parameter api_protocol need to be passed as 'rest' to use LiveDNS. I do not know how this kind of parameters are passed through certbot, but maintainers should be aware of that.

tehfink:
@maport: does your plugin support LE's wildcard certificates?

maport:
> @maport: does your plugin support LE's wildcard certificates?

@tehfink: No, afraid not, it was written before LE added support for those.

bmw:
[...]

PR #6290, opened 2018-08-07
Gandi DNS Authenticator

martinplaner:
This PR implements an authenticator for the Gandi LiveDNS API. Code is mostly based on the existing DNSimple authenticator. 

Some remarks:
- This should (partially) satisfy #5582.
- ~~Currently only the newer [Gandi LiveDNS API](https://doc.livedns.gandi.net/) is supported. The older [RPC API](https://doc.rpc.gandi.net/) is _not_, since I no longer have any domains managed through this API. Maybe someone else can implement this in the future, possibly with an appropriate command-line switch.~~
- I had to bump lexicon to 2.4.1 for it to support the LiveDNS API.


liaohuqiu:
Hi, who can check this?

jbauson:
Wooohoooo!! Waiting to get this implemented! 👍

adferrand:
@martinplayer, just to let you know that I have an incoming PR on Lexicon for my Gandi provider: AnalogJ/lexicon#274. When merged, you should use auth_api_protocol instead of api_protocol. The parameter api_protocol will continue to work, but will generate a deprecation warning when used.

Furthermore I think I still have a domain on RPC, so we could do something to implement the old API.

martinplaner:
@adferrand Thanks for the heads up! I wasn't quite sure how this project handles external dependencies, so I didn't bump the required lexicon version to the latest. But if we end up with the latest version and your PR is merged, I'll gladly change it to work with future lexicon versions.

Regarding the RPC API, it shouldn't be that complicated to support both APIs (except testing would have to be done by someone else, as mentioned before), but there are 2 smalls decisions to be made that I'd like the maintainers to comment on before implementing:

- How to switch between protocols, additional flag or some switch in the credentials file? ( Since 

- Which API should be the default one, if none is set by flag/config? Just mirror Lexicon behavior and use RPC by default?

adferrand:
Just to keep informed here. With @AnalogJ we discussed a lot on AnalogJ/lexicon#274. We are going to an approach implying to keep --api-protocol, so there will be no `--auth-api-protocol`.

For you questions @martinplaner:
* About the protocol switch, you can check OVH plugin, also backed by Lexicon, were the parameter `endpoint` is exposed. It toggles which endpoint will be used to reach the OVH API. I think it is quite similar to problematic of choosing the protocol. So maybe you could use a parameter here also.
* In fact your second question is closely related to the first one: the procotol (RPC or Live) defines the endpoint to use and all the logic behing in the Lexicon provider. So if you have a parameter to choose between RPC or Live, you have everything you need.
* About which protocol to use by default, I choose in Lexicon the RPC one, with the idea to change someday to Live, when RPC support is dropped or usage becomes negligeable. For now it is not the case, so I would advise to use RPC by default.

FauxFaux:
I rebased this PR, touching only `tox`/`venv` conflicts (`.sh` -> `.py` move).

brandond:
What's blocking this from moving forward at this point?

adferrand:
Currently most of the certbot team is on hollydays. So I think work will start again on January. There is also big discussions about how to integrate and maintain dns plugins in Certbot, so plans may change about PRs concerning DNS plugins.

In the mean time, if you need right now certificates validated against dns challenges using gandi, and have access to Docker, you may be interested by the container I maintain for that purpose by orchestrating Certbot and Lexicon together: https://github.com/adferrand/docker-letsencrypt-dns

brandond:
Thanks @adferrand I'll take a look at that! I'd found the one that @martinplaner published a while back at martinplaner/certbot-dns-gandi on docker hub, which seems to be working so far.

luyseyal:
:100: Thank you for making this.

MTCoster:
Does anyone have an ETA for when this will be merged?

adferrand:
As I said in my previous comment here, where are currently working to find the most relevant solution that will allow Certbot to interact with various DNS providers, while ensuring a sustainable support for the Certbot team.

So currently we are freezing integration of new DNS plugins, without an ETA for when the situation will change so far. We are really sorry about that.

In the mean time, you can have a look at the Docker project that I actively maintain. It could match your needs. See my previous post here for the link.

wxcafe:
Hey! Is DNS plugins integration still frozen? Some of our customers have asked us why we (gandi) are not supported by the official client and they need to install a plugin to automate their certificate renewal

vielmetti:
It's the end of the year, @adferrand - has there been movement on the planning for updating this?

wxcafe:
Hey! Is DNS plugins integration still frozen? Some of our customers have asked us why we (gandi) are not supported by the official client and they need to install a plugin to automate their certificate renewal

wxcafe:
Sorry to ping this again, but it's been four months. Any news?

m0namon:
Apologies @wxcafe, we are still not accepting DNS plugins at the moment. We want to make a decision on #6464, #6503, and #6504 first.

@martinplaner -- you're welcome to leave your PR open, but we probably won't be able to give any feedback on it for a while. In the meantime, you're welcome to release it as 3rd party plugin if you want like others have done. See https://github.com/coldfix/certbot-dns-netcup for one example of that.

martinplaner:
I'll close this PR for the time being. I've personally moved away from using certbot since creating this request, so I'm no longer in active need for this plugin.

Also I haven't touched it since then and I doubt it will still work as-is, or at least will require some rework after the blocking issues are resolved.

If anyone still requires this functionality and feels up to the task, please feel free to take this PR as your basis.

PR #6295, opened 2018-08-11
Test everything

jennyb2911:


bmw:
The test-everything branch remains separate from master to run more and longer running tests than we want to run on every PR. For that reason, I'm closing this PR.

Issue #4351, opened 2017-03-15
3rd party plugin flags


bmw:
The name of 3rd party plugins is generated in `certbot/plugins/disco.py`. Currently, all third party plugin names are `<package>:<entry point>`. Because we use the plugin name to prefix the CLI flag name, this generates some really awkward flags `--certbot-foo:bar-baz`. We should do better, but make sure this doesn't break existing plugins by allowing the old form.

bmw:
I think we can stop adding the name of the package to the plugin flag and just use the name of the entry point which the plugin author controls. The big thing here is we don't want to break backwards compatibility. The `package_name:entry_point_name` version of the flags still need to work. This affects both adding the plugin flags and fetching  the values from the parsed config.

Perhaps a simple solution is to pass multiple flags to `add_argument` [here](https://github.com/certbot/certbot/blob/master/certbot/plugins/common.py#L79) if `option_namespace` contains the package name. The `option_namespace` should probably still comes first as that's what affects the path in the parsed config, however, plugins can access these values using [conf()](https://github.com/certbot/certbot/blob/master/certbot/plugins/common.py#L103) so they don't have to see this ugliness. Only downside with this approach I see is the `package_name:entry_point_name` version will be used in renewal configuration files, but I can't see any other problems here. I definitely could be overlooking something though so this will require some experimentation and testing.

bmw:
I still think we should do this. It's easily the most common complaint I see about writing 3rd party Certbot plugins.

Fluepke:
Due to the colon it is not possible to provide configurations for plugins in the `cli.ini` configuration file.

PR #6310, opened 2018-08-21
Add netcup DNS authenticator

coldfix:
This PR adds a DNS authenticator plugin for [netcup](https://www.netcup.eu/)'s [CCP API](https://www.netcup-wiki.de/wiki/CCP_API).

Things to note:
- apart from unittest, I have tested the plugin manually on my account to work in a live environment
- I have independently [released the plugin on PyPI](https://pypi.org/project/certbot-dns-netcup/) to already make this plugin available to others in case it will take longer to be considered for inclusion, or if in case netcup is not considered a big enough backend for upstream support. If this is merged I will gladly yield control over the PyPI entry.

Best, Thomas


bmw:
Thanks for the PR @coldfix!

I assume netcup doesn't have an official Python library for their DNS API. Is that correct?

If that's the case and you'd like to see this plugin taken over by us, I would like to see support for netcup added to https://github.com/AnalogJ/lexicon and this plugin modified to use that.

Alternatively, you're of course welcome to continue to support this yourself as a third party plugin. If you go this route and have any issues due to being a third party plugin, please let me know!

coldfix:
Hi, thanks a lot for your pointers! I will update the PR to go the lexicon route and report back when that's done.

> If you go this route and have any issues due to being a third party plugin, please let me know!

In fact, I would like to point out that I find the prefix that certbot imposes on all parameter names for all external plugins to be a somewhat big incentintive to not make/use external plugins at all:

- it's one step harder to find out (as a developer) what the actual parameter names are, since they are inconsistent with all the documented ones from included plugins and you will have to look at `certbot/plugins/disco.py` for that
- they are more inconvenient to use
- there is an inconvenience if the plugin is merged later, since then both the command line options as well as the config file parameters change suddenly

If it weren't for the fact of prefixes for external plugins, I would go so far that I would prefer to maintain many DNS plugins externally - which makes it much more dynamically.

bmw:
I agree that the way we mangle third party plugin flags is a major inconvenience. If you'd prefer to fix that instead of or in addition to other work for this specific plugin, see https://github.com/certbot/certbot/issues/4351. That issue is on our TODO list, but the Certbot team hasn't had the time to get to it ourselves.

bmw:
@coldfix, we're going through and cleaning up our PR queue a bit. Did you decide on the approach you wanted to take here?

coldfix:
Hi, I'll close for now and as soon as I have time to do the lexicon based PR, I will open a new one. Thanks!

PR #6311, opened 2018-08-21
OCSP response prefetching for Apache plugin

joohoi:
This PR adds OCSP staple prefetching with `certbot renew`.

### General info

The PR uses dbm module for storing the OCSP cache, while a non-prefetch functionality we have in place uses the in-memory socache module. If a preexisting staple cache directive is found, it will be replaced by the enable method.

OCSP prefetching functionality uses the pluginstorage to store the last refresh timestamps and certificate paths for the managed certificates.

When restarting, Apache deletes the DBM cache file. When calling `restart()`, we check the pluginstorage if we're supposed to manage OCSP prefetching, and backup the dbm file, restoring it after calling restart if necessary.

### Apache DBM file structure

The Apache DBM file is practically a simple key-value store. The key that Apache uses for each entry is the certificate SHA1 fingerprint in its binary form. The value is the internal expiry time value of the cached entry in milliseconds prepended by the raw OCSP response.

The internal renewal of OCSP response is done by Apache only after the internal TTL is hit, and the actual client request is blocked until it is able to fetch it. This PR uses Apache TTL of 5 days for the cache values, while Certbot will attempt to refresh the value if it's older than a day.

### Tests

I was unable to write meaningful tests for the functionality that actually shells out to `openssl` and performs actual network queries towards the OCSP server.

For manual testing I have been using a busyloop that opens the DBM file, and checks for changes in the keys and values. Something along the lines of (note that dbm automatically appends the file extension `.db` to the path):
```
import dbm
import time
import binascii

cur_vals = dict()

while True:
    time.sleep(1)
    x = dbm.open('/etc/letsencrypt/ocsp/ocsp_cache', 'c')
    for k in x.keys():
        try:
            if cur_vals[k] != x[k]:
                print("Key {} changed, this should not happen because of OCSP caching".format(
                    binascii.hexlify(k)))
                cur_vals[k] = x[k]
        except KeyError:
            print("New key {} in DBM".format(binascii.hexlify(k)))
            cur_vals[k] = x[k]
    print("DBM holding {} keys.".format(len(x.keys())))
    x.close()
```


bmw:
Looks like tests are failing. Let me know if you'd like me to take a look.

joohoi:
Well, it seems like this isn't going to happen. 

I did some digging in on how to reliably use the correct dbm module (of which there are four variants of) for various environments, I figured out that we'd have to roll our own. Looking more deeply into Apache `mod_socache_dbm` source code made it clear that there's no way to reliably predict the backend dbm engine flavor (that is decided in compile time), and as such I feel that we can't ship this feature. This would probably work in two thirds of the environment and installations as it is, but it's not nearly enough to justify a fragile implementation with user promises that we're not reliably able to meet.

It has taken quite a bit of effort, and I feel sad that this is the resolution, but it really doesn't seem like a great fit for Certbot. Because of this, I'm closing the PR.

joohoi:
Looks like this could happen after all! Though only for Debian & Ubuntu for now, as these are the only distributions we're able to tell the dbm engine flavor and Python dbm module support with good enough certainty.

The tests now mock the whole dbm engine, with a fully functional alternative in order to provide meaningful tests for the rest of the codebase.

joohoi:
There was a race condition caused by the way I had mocked `ImportError`s previously. This is now fixed, but this PR still needs a few more tests for `cert_util.py` and `ocsp.py` that I missed earlier.

bmw:
@joohoi, have you done any refactoring you wanted to do here?

I also think we should break up this PR a bit. While I've reviewed and wrote plenty of large PRs in the past, I think there's a lot of value in getting in the habit of submitting smaller PRs like we talked about on the call a while ago. An easy split here is the first PR only makes modifications to the `certbot` package while the 2nd one works on `certbot-apache`. If there are better or additional ways to break this up as you understand the code in this PR better than I do right now, please do that instead.

joohoi:
I'm closing this in favor of the refactored PR #7649

PR #6320, opened 2018-08-26
certbot_dns_linode: increase the default propagation interval

stefantalpalaru:
Using the default value of 16 minutes (960 seconds) for
--dns-linode-propagation-seconds leads to DNS failures when the randomly
selected Linode DNS is not the first one out of six, due to an additional
delay before the other five are updated.

The problem can be easily solved by increasing the wait interval, so
this commit increases the default value to 20 minutes.

More details: https://community.letsencrypt.org/t/dns-servers-used-by-letsencrypt-for-challenges/32127/16

### DIFF
--- a/certbot-dns-linode/certbot_dns_linode/__init__.py
+++ b/certbot-dns-linode/certbot_dns_linode/__init__.py
@@ -14,7 +14,11 @@ Named Arguments
                                             DNS to propagate before asking the
                                             ACME server to verify the DNS
                                             record.
-                                            (Default: 960)
+                                            (Default: 1200 because Linode
+                                            updates its first DNS every 15
+                                            minutes and we allow 5 more minutes
+                                            for the update to reach the other 5
+                                            servers)
 ==========================================  ===================================
 
 
@@ -74,13 +78,15 @@ Examples
      -d www.example.com
 
 .. code-block:: bash
-   :caption: To acquire a certificate for ``example.com``, waiting 60 seconds
-             for DNS propagation
+   :caption: To acquire a certificate for ``example.com``, waiting 1000 seconds
+             for DNS propagation (Linode updates its first DNS every 15 minutes
+             and we allow some extra time for the update to reach the other 5
+             servers)
 
    certbot certonly \\
      --dns-linode \\
      --dns-linode-credentials ~/.secrets/certbot/linode.ini \\
-     --dns-linode-propagation-seconds 60 \\
+     --dns-linode-propagation-seconds 1000 \\
      -d example.com
 
 """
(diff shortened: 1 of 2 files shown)


schoen:
Yikes, does Linode have any kind of status API that might be able to make this faster? I fear for users' patience with this a little bit (although if there's no other alternative, we should make this change).

stefantalpalaru:
I'm not aware of any API to notify us of their internal DNS propagation.

Yes, I'm afraid we extend the user's wait in the manual tests and initial certificate creation, but I got DNS errors after waiting the default 16 minutes during my wildcard cert setup and I can't afford to gamble at this lottery with renewal cron jobs.

It's better to be on the safe side with that propagation interval.

sydneyli:
@trinopoty, care to weigh in on this?

trinopoty:
According to official Linode docs, all 5 nameservers pull changes from their API every 15 minutes. Additionally all of the linode nameservers should update at the same time. I set up 1 extra minute to ensure everything's updated across all linode nameservers.
I'm guessing this could be due to a stale cache somewhere. We can increase it to 20 minutes to ensure any cache also refresh themselves but there's no telling how a cache might be configured.

PS: The timeout is starting to get extremely long at this point though.

trinopoty:
I asked around the linode forums and couldn't get a definitive answer to this issue.

certbot-dns-linode/certbot_dns_linode/__init__.py
```diff
@@ -14,7 +14,7 @@
                                             DNS to propagate before asking the
                                             ACME server to verify the DNS
                                             record.
```
sydneyli: Comment here why the default is 20 min?

certbot-dns-linode/certbot_dns_linode/__init__.py
```diff
@@ -74,13 +74,13 @@
...
 .. code-block:: bash
-   :caption: To acquire a certificate for ``example.com``, waiting 60 seconds
+   :caption: To acquire a certificate for ``example.com``, waiting 1000 seconds
              for DNS propagation
 
    certbot certonly \\
      --dns-linode \\
      --dns-linode-credentials ~/.secrets/certbot/linode.ini \\
-     --dns-linode-propagation-seconds 60 \\
+     --dns-linode-propagation-seconds 1000 \\
```
sydneyli: I'm okay with the change in the example to reflect how long people should wait, but maybe we should also motivate why (cite linode's quarter-hour zone republishing)

stefantalpalaru: Commit amended to add comments explaining both the default value and the example value.

Issue #6325, opened 2018-08-29
Certbot not recognizing individual .conf files from include directive in nginx.conf


Briscoooe:
If you're having trouble using Certbot and aren't sure you've found a bug or
request for a new feature, please first try asking for help at
https://community.letsencrypt.org/. There is a much larger community there of
people familiar with the project who will be able to more quickly answer your
questions.

## My operating system is (include version):
Ubuntu 14.04

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
OS package manager [using official instructions](https://certbot.eff.org/lets-encrypt/ubuntutrusty-nginx)
```
$ sudo apt-get update
$ sudo apt-get install software-properties-common
$ sudo add-apt-repository ppa:certbot/certbot
$ sudo apt-get update
$ sudo apt-get install python-certbot-nginx 
```

## I ran this command and it produced this output:
`sudo certbot --nginx --nginx-server-root /opt/bitnami/nginx/conf/`
```
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator nginx, Installer nginx
No names were found in your configuration files. Please enter in your domain
name(s) (comma and/or space separated)  (Enter 'c' to cancel):
```

## Certbot's behavior differed from what I expected because:
I have a number of server blocks configured in individual `.conf` files that are being included in my `nginx.conf` from a specific directory. [See here](#nginxconf)

A typical `.conf` [can be seen here](#site1conf)

I know these `.conf` files are being successfully loaded by Nginx as I can access all of them using their server names, in this case `www.site1.com` or `site1.com` will successfully navigate to `http://site1.mydomain.com`. 

The problem is that certbot does not recognize any of the domains configured in the server blocks in the directory. The only way I can get a domain to show up when I run the `sudo certbot --nginx --nginx-server-root /opt/bitnami/nginx/conf/` is to include the contents of an individual `.conf` file directly in the `nginx.conf` file. This is not a viable option for me as I have many domains and including each of their configurations in the `nginx.conf` file would be messy.

### <a name="nginxconf">`nginx.conf`</a>
```
user  daemon daemon;
worker_processes  auto;

error_log  "/opt/bitnami/nginx/logs/error.log";

pid        "/opt/bitnami/nginx/logs/nginx.pid";


events {
    worker_connections  1024;
}


http {
    server_names_hash_bucket_size 128;

    /* typical nginx configurations */

    server {
        listen       80 default_server;
        server_name  localhost;
[...]

PR #6330, opened 2018-08-30
Added support for double-quoted file paths in nginx.conf include directives

Briscoooe:
## The issue
This is a fix for issues [#5543](https://github.com/certbot/certbot/issues/5543) and [#6325](https://github.com/certbot/certbot/issues/6325).

The issues are related to the fact that Nginx supports using either double quotes or no quotes at all when declaring a path in an `include` directive. Both of the following are syntactically correct in an `nginx.conf` file:

- `include sites-enabled1/*;`
- `include "sites-enabled2/*";`

In the above example, the absolute path of the `sites-enabled1` directory, as returned by the `abs_path` method, will be something like: 
- `/etc/nginx/sites-enabled1/*`

Whereas the absolute path of the `sites-enabled2` directory will be something like:
- `/etc/nginx/"sites-enabled2/*"`

The second path is invalid and will not be considered in any of the scripts subsequent file/directory reading operations.

## Summary of changes
- Removing double quotes `"` from file paths using Pythons `replace()` method at the earliest point relevant. [See code](https://github.com/certbot/certbot/compare/master...Briscoooe:master#diff-b2dab7e6fe39d3a1bf642d561d045633R84)
- Another test case to the `abs_path` test method. [See code](https://github.com/certbot/certbot/compare/master...Briscoooe:master#diff-cf22a62b9f3f06d4c413904d321a4b4bR73)

### DIFF
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -81,7 +81,7 @@ class NginxParser(object):
 
         """
         if not os.path.isabs(path):
-            return os.path.join(self.root, path).strip('"')
+            return os.path.join(self.root, path).replace('"', '')
         else:
             return path
 
--- a/certbot-nginx/certbot_nginx/parser.py
+++ b/certbot-nginx/certbot_nginx/parser.py
@@ -81,7 +81,7 @@ class NginxParser(object):
 
         """
         if not os.path.isabs(path):
-            return os.path.join(self.root, path)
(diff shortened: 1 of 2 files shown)


ohemorange:
Hey @Briscoooe, thanks for your work here! We're hoping to fix this in a way that it accepts quoted parameters in general. This might involve changes to `nginxparser.py`. That might end up being a trickier change. Do you feel comfortable looking into that?

ohemorange:
See discussion on the issue

PR #6332, opened 2018-09-03
Raise error if certname contains special chars

### LINKED ISSUE #6127: Don't allow special characters in lineage name
Ubuntu 16.04.4

PPA, certbot 0.22.2

```bash
root@tom-vps:/etc/letsencrypt# certbot certonly --staging --webroot -d test.tomwassenberg.nl --webroot-path /var/www/acme-challenges/ --cert-name "/example"
Plugins selected: Authenticator webroot, Installer None
Obtaining a new certificate
http-01 challenge for test.tomwassenberg.nl
Using the webroot path /var/www/acme-challenges for all unmatched domains.
Waiting for verification...
Cleaning up challenges

IMPORTANT NOTES:
 - Congratulations! Your certificate and chain have been saved at:
   /etc/letsencrypt/live/example/fullchain.pem
   Your key file has been saved at:
   /etc/letsencrypt/live/example/privkey.pem
   Your cert will expire on 2018-09-14. To obtain a new or tweaked
   version of this certificate in the future, simply run certbot
   again. To non-interactively renew *all* of your certificates, run
   "certbot renew"
```

Certbot should halt and return an error message stating that you can't use a slash in a lineage name. Instead, it continues normally and saves the regular output in `/etc/letsencrypt/{archive,live}/example`. However, when saving the renewal configuration, it becomes apparent the lineage name is parsed as a path because it is saved on the actual path `/example.conf`, so on the root of the filesystem. This shouldn't happen.

Also other permutations of names with special characters, like names with slashes between regular characters, or names starting with a hyphen, cause unexpected behavior.

troy-lamerton:
If the `--cert-name` contains `/` or `\`, or starts with `-`, an error is raised and certbot exits.

Fixes #6127

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -5,6 +5,7 @@ import functools
 import logging.handlers
 import os
 import sys
+import re
 
 import configobj
 import josepy as jose
@@ -413,6 +414,9 @@ def _find_domains_or_certname(config, installer, question=None):
     # if we can't do that but we have a certname, get the domains
     # with that certname
     elif certname:
+        if bool(re.search('[\/]', certname)) or bool(re.search('^-', certname)):
+            raise errors.Error("A certname cannot contain slashes (/ or \) "
+                               "or start a with hyphen (-).")
         domains = cert_manager.domains_for_certname(config, certname)
 
     # that certname might not have existed, or there was a problem.
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -414,7 +414,7 @@ def _find_domains_or_certname(config, installer, question=None):
     # if we can't do that but we have a certname, get the domains
     # with that certname
     elif certname:
-        if bool(re.search('[\/]', certname)) or bool(re.search('^-', certname)):
+        if bool(re.search(r'[\\/]', certname)) or bool(re.search('^-', certname)):
             raise errors.Error("A certname cannot contain slashes (/ or \) "
                                "or start a with hyphen (-).")
         domains = cert_manager.domains_for_certname(config, certname)


certbot/main.py
```diff
@@ -413,6 +414,9 @@ def _find_domains_or_certname(config, installer, question=None):
     # if we can't do that but we have a certname, get the domains
     # with that certname
     elif certname:
+        if bool(re.search(r'[\\/]', certname)) or bool(re.search('^-', certname)):
```
ohemorange: No need for `re` here! Python has the `in` operator, and for the start one a `certname[0]` will do just fine. More pythonic and readable that way. Plus that'll probably fix much of the tests -- escaping and `r''` and all are complicated.

certbot/main.py
```diff
@@ -413,6 +414,9 @@ def _find_domains_or_certname(config, installer, question=None):
...
+            raise errors.Error("A certname cannot contain slashes (/ or \) "
```
ohemorange: I think the backslash in here is likely to give trouble, probably want to `\\` that.

certbot/main.py
schoen: `certname[0] == '-'` is a good idiom; another fun option is `certname.startswith('-')`.

troy-lamerton:
> Hey @troy-lamerton, are you no longer interested in working on this PR?

Unfortunately I'm busy with another things.

ohemorange:
Ok, feel free to reopen this if you have more time in the future; closing now to keep our PR queue clean.

PR #6334, opened 2018-09-04
Route53 assume role

lurcio:
Add a `--dns-route53-assume-role` command line option to assume a role in a different account before performing DNS validation.

I'd appreciate some advice on how to add tests for this please.

### DIFF
--- a/certbot-dns-route53/certbot_dns_route53/__init__.py
+++ b/certbot-dns-route53/certbot_dns_route53/__init__.py
@@ -12,6 +12,10 @@ Named Arguments
                                           to propagate before asking the ACME
                                           server to verify the DNS record.
                                           (Default: 10)
+
+``--dns-route53-assume-role``             Assume a role in a different AWS
+                                          account. Specify the account ID
+                                          and role separated with a colon.
 ========================================  =====================================
 
 
@@ -116,5 +120,14 @@ Examples
    certbot certonly \\
      --dns-route53 \\
      --dns-route53-propagation-seconds 30 \\
+     -d example.com
+
+.. code-block:: bash
+   :caption: To aquire a certificate for ``example.com``, assuming a role in
+             a different account
+
+   certbot certonly \\
+     --dns-route53 \\
+     --dns-route53-assume-role 123456:dns-role \\
      -d example.com
 """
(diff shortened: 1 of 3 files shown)


bmw:
@jsha, do you have the interest and cycles to weigh in on this?

jsha:
@lurcio can you explain what use case this solves, and why it can't be solved in a different way? Thanks!

lurcio:
@jsha thanks for taking a look at this. My use case is someone with a multi account AWS infrastructure who has Route53 zones in one master account, but with services deployed into other member accounts. If there's a different way to achieve this then I'd love to know, but I haven't found the magic IAM incantations to allow DNS updates from different accounts.
Thanks!

hussfelt:
I have the exact same usecase right now, multiple account setup - and I need to change zones in multiple accounts and then deploy those from one of the accounts to ACM in all the others.

As I am currently doing this from a Lambda function I get into a state where I need Lambdas in all accounts where I have active delegated domains. And for wildcard domains as to where the Route53 zone is in one account and the usecase in multiple others I need to do the same.

It gets even messier when the Lambda then need to upload the certificates to all these ACM repositories.

This would clear things up significantly and I would be able to run one lambda for all my given domains and then have logic to upload them to the right account and ACM separately.

jsha:
It seems like you could just provide the Certbot instance with the credentials for an account that has permissions to update Route53 without having to assume additional roles. Is there a reason that doesn't work?

hussfelt:
@jsha When it comes to AWS the preferred way is to assume roles. In this specific case we host our Route53 zones in *different accounts* than we do the actual Lambda running the certbot instance, issuing the certification. I run this validation to multiple domains at once, located in multiple different accounts - which means I'd have to create a separate IAM user that I store the credentials in secrets-manager and then assert to Certbot.

It would make much more sense to just assume one role that I can give any permission I want to, simplifying the setup a great deal.

jsha:
Sorry I've let this lag so long. I don't feel comfortable merging in support for the assume_role functionality without knowing enough about AWS practices to be sure this is the correct way to go. For now I think multiple credential files to go with different accounts, or a single account that has appropriately-scoped permissions on all relevant Route53 zones, are ideal.

You may also be interested to know you can CNAME the _acme-challenge domain, so you could have one account with permission to modify a single "validation" zone, and have all your other zones' _acme-challenge subdomains CNAMEd to that validation zone.

PR #6358, opened 2018-09-07
Update DNS plugin docs

bmw:
This will be more valuable once https://github.com/certbot/website/issues/353 is resolved, but this makes it clear these plugins aren't installed by default and points people to https://certbot.eff.org to learn the best way to install them. If Docker is the user's best option right now, https://certbot.eff.org will point them to https://certbot.eff.org/docs/install.html#running-with-docker which is the target of the `<docker-user>` link I'm deleting.

@schoen, can you review this?


schoen:
I'll work on certbot/website#353 today.

bmw:
Awesome. Thanks Seth!

PR #6370, opened 2018-09-13
Warn manual authenticator users not to remove/undo previous challenges

schoen:

I did this work quite a while ago and it works fine but is still missing tests. I just had a reminder on the forum that this does in fact continue to confuse people, so I wanted to create the PR in the hope that that will lead to my addressing this sooner.

The issue is that when you use `--manual`, you should *not* remove previous challenges when satisfying new ones. But many users don't realize this and it's easy to be unsure of whether it's OK to remove the previous challenges or not.


bmw:
Thanks for this Seth! Let me know if you need any help fixing tests and please ping someone for a review when it's ready.

schoen:
I think this is ready to be reviewed now. There is a minor problem in the tests that we do get coverage for all of the subsequent challenge cases (I added a 2nd DNS-01 challenge so that two of them will occur during the testing of the manual plugin!), but we don't explicitly check the information passed to the UI to see whether it contains the intended message in each situation. I'm not sure where that check would go if we wanted to include it.

schoen:
Kiitos, @joohoi!

joohoi:
Thanks for pointing out the current way to handle changelog! LGTM (Minusta tämä näyttää hyvältä)!

PR #6383, opened 2018-09-18
Add and test new nginx parsing abstractions

sydneyli:

Part 1/? of #5834. Check that PR thread out for more context/explanation of these objects.

`parser_obj` should not contain any nginx-specific parsing directives.

@ohemorange, lmk what you think here! It was from a while ago. The primary purpose of the classes in parser_obj are to replace `UnspacedList`.

### DIFF
--- a/certbot-nginx/certbot_nginx/parser_obj.py
+++ b/certbot-nginx/certbot_nginx/parser_obj.py
@@ -7,11 +7,13 @@ import six
 
 from certbot import errors
 
+from acme.magic_typing import List # pylint: disable=unused-import, no-name-in-module
+
 logger = logging.getLogger(__name__)
 COMMENT = " managed by Certbot"
 COMMENT_BLOCK = ["#", COMMENT]
 
-class WithLists(object):
+class Parsable(object):
     """ Abstract base class for "Parsable" objects whose underlying representation
     is a tree of lists.
 
@@ -22,14 +24,14 @@ class WithLists(object):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self, context):
-        self._data = []
+        self._data = [] # type: List[object]
         self._tabs = None
         self.context = context
 
     @staticmethod
     @abc.abstractmethod
     def should_parse(lists):
-        """ Returns whether the contests of `lists` can be parsed into this object.
+        """ Returns whether the contents of `lists` can be parsed into this object.
 
         :returns: Whether `lists` can be parsed as this object.
         :rtype bool:
@@ -109,7 +111,7 @@ class WithLists(object):
         """
         return [elem.dump(include_spaces) for elem in self._data]
 
-class Statements(WithLists):
+class Statements(Parsable):
     """ A group or list of "Statements". A Statement is either a Block or a Sentence.
 
     The underlying representation is simply a list of these Statement objects, with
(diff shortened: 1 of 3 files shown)


certbot-nginx/certbot_nginx/parser_obj.py
```diff
@@ -0,0 +1,407 @@
...
+import six
+
+from certbot import errors
+
+logger = logging.getLogger(__name__)
+COMMENT = " managed by Certbot"
+COMMENT_BLOCK = ["#", COMMENT]
+
+class WithLists(object):
+    """ Abstract base class for "Parsable" objects whose underlying representation
```
ohemorange: Why isn't it called `Parsable` then?

certbot-nginx/certbot_nginx/parser_obj.py
```diff
@@ -0,0 +1,407 @@
...
+    """ Context information held by parsed objects.
+
+    :param .WithLists parent: The parent object containing the associated object.
+    :param tuple parsing_hooks: Parsing order for `parse_raw` to use.
+    """
+    def __init__(self, parent=None, parsing_hooks=DEFAULT_PARSING_HOOKS):
+        self.parsing_hooks = parsing_hooks
+        self.parent = parent
+
+    def child(self, parent):
```
ohemorange: `create_child`?

ohemorange: actually, why is `parent` on the context, rather than in the object? is `ParseContext` just the metadata for a `WithLists`? If so, let's put it inside the class, and if not, what's going on here?

ohemorange: oh, it is so we can just pass around the contexts for the helper methods?

certbot-nginx/certbot_nginx/tests/parser_obj_test.py
```diff
@@ -0,0 +1,263 @@
...
+        from certbot_nginx.parser_obj import Block
+        self.assertFalse(Block.should_parse([]))
+        self.assertFalse(Block.should_parse(['']))
+        self.assertFalse(Block.should_parse(['two', 'words']))
+        self.assertFalse(Block.should_parse([[[]], []]))
+        self.assertFalse(Block.should_parse([['block_name'], ['hi', []], []]))
+        self.assertFalse(Block.should_parse([['block_name'], 'lol']))
+        self.assertTrue(Block.should_parse([['block_name'], ['hi', []]]))
+        self.assertTrue(Block.should_parse([['hello'], []]))
+        self.assertTrue(Block.should_parse([['block_name'], [['many'], ['statements'], 'here']]))
```
ohemorange: `self.assertTrue(Block.should_parse([['if', ' ', '(whatever)'], ['hi']]))`

ohemorange:
Also, mypy is sad

sydneyli:
The boulder integration tests are also sad, since this file isn't being used yet. Is there a way I can selectively disable coverage reporting for that file for the integration tests but not for the unit tests?

certbot-nginx/certbot_nginx/parser_obj.py
sydneyli: Great question! I introduced contexts in the original giant PR to keep track of things like filesystem location (so we could write these objects to the appropriate files afterwards), but am now questioning whether we need them at all-- At the very least, we definitely don't need them in this PR, so I ripped them out. If it turns out something like this would be helpful for the follow-up PRs, I will introduce them in those PRs :)

sydneyli:
This is ready for another round!

ohemorange:
`--omit` as a flag to coverage to ignore that file for now

ohemorange:
lint has a few things, too

PR #6386, opened 2018-09-18
Certbot dns authenticator to update DNS in plesk.

noci2012:
This is a new dns authenticator.
This is a stab creating a new certbot authenticater,

I am looking for comments (and CI checks).


ohemorange:
Non-lexicon

noci2012:
@ohemorange  What do you mean with non-lexicon?
I am not a python coder by profession (obviously) i am looking for comments where this code can be improved.

ohemorange:
Sorry @noci2012, that was a note for reviewing on our end -- we get a bunch of DNS plugin submissions, and we follow a different process for ones that are based on [lexicon](https://github.com/AnalogJ/lexicon) and ones that use the provider's API (the path to acceptance is smoother for lexicon ones).

noci2012:
ah ok. I took the digitalocean dns authenticator as base and replaced the code to access the external API to the one of  plesk to update their DNS from the system that actually runs the SSL session.  (outside of the ISP running plesk).

ohemorange:
Looks like lexicon supports plesk! In our repo, dnsimple is a good example of a lexicon-based plugin.

noci2012:
No problem. What is needed to implement the plesk one.

Protocol (http/https)
Sitename
Portnumber
and either:
  Username
  Password
or 
  Token...  ( although i see no means of how to create a token on on the plesk i have access to, the API allows for one).

who does read the .ini file?

Should the dnsimple one be adjusted?   or is it meant to be handled differently.

noci2012:
Oh btw, i just tried the lexicontool, the Lexicon plesk plugin cannot retrieve the SOA record.

noci2012:
I think i have recoded it for lexicon i just need to check that somehow....
First i need to get lexicon installed correctly it might take some more time.

noci2012:
Well here the version based on lexicon....

noci2012:
@ohemorage i could successfully get a certificate from --test-cert so ready for review again.

noci2012:
when can i expect the review to be done?

ohemorange:
Thanks for all your hard work here, we really appreciate outside contributions. The team all just got back and we're working through our backlogs, be assured that this is on our radar.

ohemorange:
`--omit` as a flag to coverage to ignore that file for now

bmw:
@noci2012, we'll try to get to it as soon as we can but it will probably be at least a couple weeks.

In the meantime, can you tell me about the use case for using this plugin over https://ext.plesk.com/packages/f6847e61-33a7-4104-8dc9-d26a0183a8dd-letsencrypt?

noci2012:
Use case 1:   
ISP only uses plesk to provide DNS services, no other services (like hosting) are provided through Plesk.
The Client is expected to use the web interface to add/remove names to one or more domains.

Using the Plesk API to add/remove names gives the option automate certbot for this case.
(The ISP has no intention to supply hosting services).

The referenced plesk package will not available for this ISP. (which primarily does Internet & VOIP serving.) (Also if useful, it would only serve to export all data every 3 months).

Use Case 2:
Someone has systems outside of a Plesk environment DNS is managed through Plesk
(The plesk plugin can be used for the website on plesk) the other systems would also need a certificate for haproxy frontend, alfresco server, etc. 
The outside systems need a way to prolong certificates.

noci2012:
ping?

noci2012:
@bmw @ohemorange   
Still mergable with current release, is still viable?
If this is not going to happen then please close the pull request.

kolcon:
This is now quite old issue - any progress?

bmw:
We're hoping to set aside some time to rethink our approach to DNS plugins in the next year. I can't say anything more definitive than that right now though.

In the meantime, people are welcome to release code like this as a 3rd party plugin. See https://github.com/coldfix/certbot-dns-netcup for one example of that.

noci2012:
This may be obsolete.  The certbot-dns-standalone might be a better approach.
See: https://github.com/siilike/certbot-dns-standalone
At least that works now and is documented.

PR #6389, opened 2018-09-23
Respect --quiet when reporting sudo invocation

DigitalBrains1:
My cron entry for certbot-auto is not completely silent even though it has --quiet. The only thing echoing on every invocation is the line changed in this commit.

Thanks for the whole LetsEncrypt thing!


schoen:
Thanks, that looks like an improvement to me!

Issue #6209, opened 2018-07-12
Duplicate Route53 options in `docs/cli-help.txt`


trinopoty:
There are 2 entries of `certbot-route53` in `docs/cli-help.txt`.
One around [line 496](https://github.com/certbot/certbot/blob/master/docs/cli-help.txt#L496):
`
certbot-route53:auth:
  Obtain certificates using a DNS TXT record (if you are using AWS Route53
  for DNS).
`
and another around [line 615](https://github.com/certbot/certbot/blob/master/docs/cli-help.txt#L615):
`
dns-route53:
  Obtain certificates using a DNS TXT record (if you are using AWS Route53
  for DNS).
`

Should it be like that or should one be removed?

bmw:
> Should it be like that or should one be removed?

That's a good question. The reason for this is `certbot-route53:auth` was a 3rd party plugin that we took over maintenance of and made a shim around `certbot-dns-route53`. We keep `certbot-route53:auth` around for backwards compatibility.

@hugopeixoto, do you have a preference about whether `certbot-route53:auth` shows up in Certbot's help output?

hugopeixoto:
@bmw I don't think it's worth keeping both lines, it can be removed

Oshawk:
Fixed https://github.com/certbot/certbot/pull/6405

PR #6405, opened 2018-10-04
Remove duplicate route53 options

Oshawk:
Duplicate route53 option (certbot-route53:auth) removed.

Fixes https://github.com/certbot/certbot/issues/6209

Not sure if it is worth updating the changelog for such a small change.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -749,11 +749,6 @@ class HelpfulArgumentParser(object):
         :param dict **kwargs: various argparse settings for this argument
 
         """
-
-        # Remove duplicate route53 options from help.
-        if args[0] == '--certbot-route53:auth-propagation-seconds':
-            kwargs["help"] = argparse.SUPPRESS
-
         if isinstance(topics, list):
             # if this flag can be listed in multiple sections, try to pick the one
             # that the user has asked for help about
@@ -840,9 +835,8 @@ class HelpfulArgumentParser(object):
 
         """
         for name, plugin_ep in six.iteritems(plugins):
-            if name != 'certbot-route53:auth':  # Remove duplicate route53 options from help.
-                parser_or_group = self.add_group(name,
-                                                 description=plugin_ep.long_description)
+            parser_or_group = self.add_group(name,
+                                             description=plugin_ep.long_description)
             plugin_ep.plugin_cls.inject_parser_options(parser_or_group, name)
 
     def determine_help_topics(self, chosen_topic):
@@ -861,7 +855,8 @@ class HelpfulArgumentParser(object):
         if chosen_topic == "everything":
             chosen_topic = "run"
         if chosen_topic == "all":
-            return dict([(t, True) for t in self.help_topics])
+            # Addition of condition closes #6209 (removal of duplicate route53 option).
+            return dict([(t, True) if t != 'certbot-route53:auth' else (t, False) for t in self.help_topics])
         elif not chosen_topic:
             return dict([(t, False) for t in self.help_topics])
         else:
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -749,6 +749,7 @@ class HelpfulArgumentParser(object):
         :param dict **kwargs: various argparse settings for this argument
 
         """
+
         if isinstance(topics, list):
             # if this flag can be listed in multiple sections, try to pick the one
             # that the user has asked for help about
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -750,6 +750,9 @@ class HelpfulArgumentParser(object):
 
         """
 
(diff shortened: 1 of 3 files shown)


Oshawk:
I have been trying to find where it is generated for about 2 hours but have had no luck. The only mention of rout53:auth seems to be in the `setup.py` and removing that does not fix anything. I'm afraid I am stumped.

trinopoty:
Check out `certbot-dns-route53/certbot_dns_route53/authenticator.py`

Oshawk:
Think that may have solved it. I was having trouble due to the egg info.

ohemorange:
Hey @Oshawk, how's progress going?

Oshawk:
Hi @ohemorange I have discovered that I need to somehow pass in argparse.SUPRESS or use add_deprecated_argument() instead of the normal add_argument() although I am not sure how I would do this as arguments are added in bulk I believe. My current plan is to add a line to the cli.py file to test if the argument is the shim and if so supress it. I sadly don't see any cleaner solution at the moment.

ohemorange:
That sounds reasonable enough.

trinopoty:
Removing `'certbot-route53:auth = certbot_dns_route53.authenticator:Authenticator'` from `certbot-dns-route53/setup.py` line 54 removes the output in help text but I'm not sure if that prevents certbot from detecting the plugin entirely.

trinopoty:
There doesn't seem to be any other way to remove it from help text without adding a specific condition for the plugin in cli.py

ohemorange:
> Removing `'certbot-route53:auth = certbot_dns_route53.authenticator:Authenticator'` from `certbot-dns-route53/setup.py` line 54 removes the output in help text but I'm not sure if that prevents certbot from detecting the plugin entirely.

Almost certainly, but you can test this by removing that line, then trying to select the auth plugin from the command line

Oshawk:
@ohemorange The latest commit should do it.

certbot/cli.py
```diff
@@ -836,8 +839,9 @@ def add_plugin_args(self, plugins):
 
         """
         for name, plugin_ep in six.iteritems(plugins):
-            parser_or_group = self.add_group(name,
-                                             description=plugin_ep.long_description)
+            if name != 'certbot-route53:auth':  # Remove duplicate route53 options from help.
+                parser_or_group = self.add_group(name,
+                                                 description=plugin_ep.long_description)
             plugin_ep.plugin_cls.inject_parser_options(parser_or_group, name)
```
ohemorange: This should also be in the `if`

certbot/cli.py
```diff
@@ -750,6 +750,9 @@ def add(self, topics, *args, **kwargs):
 
         """
 
+        if args[0] == '--certbot-route53:auth-propagation-seconds':  # Remove duplicate route53 options from help.
```
ohemorange: Why is adding it in the other place insufficient?

certbot/cli.py
ohemorange: `parser_or_group` is set to the previous loop's iteration if the first `if` isn't hit.

Oshawk: Yes it is, I believe it is set to apache. If groups are not used for anything other than help then does this matter as it is always hidden? I don;t know if it is possible to add it to no group.

certbot/cli.py
ohemorange: Adding it to the Apache group isn't ideal.

Found a cleaner way to do this: in `determine_help_topics`, can just always set this topic to false:
```
if chosen_topic == "all":
    return dict([(t, True) if t != 'certbot-route53:auth' else (t, False) for t in self.help_topics])
```

Then it's only one place to keep track of things.

Oshawk: Yep, that is certainly better. Have made the commit.

docs/cli-help.txt
```diff
@@ -503,15 +503,6 @@ apache:
                         Full path to Apache control script (default:
                         apache2ctl)
 
-certbot-route53:auth:
```
ohemorange: This file is automatically generated at release time and shouldn't be updated manually.

certbot/cli.py
```diff
@@ -749,7 +749,6 @@ def add(self, topics, *args, **kwargs):
         :param dict **kwargs: various argparse settings for this argument
 
         """
-
```
ohemorange: nit: re-add line

docs/cli-help.txt
```diff
@@ -506,7 +506,6 @@ apache:
 certbot-route53:auth:
   Obtain certificates using a DNS TXT record (if you are using AWS Route53
   for DNS).
-
```
ohemorange: Add this back and we're good to go!

PR #6411, opened 2018-10-11
Stop caching the results of ipv6_info in http01.py

ohemorange:
A single call might change the results of later calls. Add tests for this and `_default_listen_addresses` more broadly.

From the ashes of #6366.


sydneyli:
fix makes sense to me-- a lint & a couple tests in http_01 are failing though

ohemorange:
Tests are passing!

PR #6412, opened 2018-10-11
Fix ranking of vhosts in Nginx so that all port-matching vhosts come first

ohemorange:
To more closely match how Nginx ranks things.

From the ashes of #6366.

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -422,6 +422,7 @@ class NginxConfigurator(common.Installer):
 
     def _rank_matches_by_name(self, vhost_list, target_name):
         """Returns a ranked list of vhosts from vhost_list that match target_name.
+        This method should always be followed by a call to _select_best_name_match.
 
         :param list vhost_list: list of vhosts to filter and rank
         :param str target_name: The name to match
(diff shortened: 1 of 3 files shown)


certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -437,21 +441,37 @@ def _rank_matches_by_name_and_ssl(self, vhost_list, target_name):
...
                 matches.append({'vhost': vhost,
                                 'name': name,
-                                'rank': 4 if vhost.ssl else 5})
+                                'rank': END_WILDCARD_RANK})
             elif name_type == 'regex':
                 matches.append({'vhost': vhost,
                                 'name': name,
-                                'rank': 6 if vhost.ssl else 7})
+                                'rank': REGEX_RANK})
         return sorted(matches, key=lambda x: x['rank'])
```
sydneyli: The length of the name match is used to break wildcard ties in `parser.get_best_match`, but not here. Is the expectation that if it matters, you should call `_select_best_name_match` to disambiguate?

ohemorange: Yes, the two functions are meant to work together. This function should never be called directly. OR at least that's what I'm assuming based on the code, which I did not originally write.

sydneyli: Ok, while we're here let's add a comment in `_rank_matches_by_name` to clarify that it should be called with `_select_best_name_match`? Otherwise the PR LGTM!

PR #6413, opened 2018-10-11
obj.py was failing to write out all parameters. let's fix that.

ohemorange:

From the ashes of #6366.


ohemorange:
For some reason, this is the PR that breaks tls-sni. No one's been complaining about this, it's just a correctness thing. So let's wait until then and not bothering fixing tls-sni.

sydneyli:
@ohemorange any idea about the funky integration test failures?

ohemorange:
Yes. It's because tls-sni copies over addresses wholesale. The easiest way to fix this is to remove tls-sni and not have to worry about it (see comment above).

m0namon:
Closing this out and throwing it into an issue!

PR #6415, opened 2018-10-12
[Windows] Handle file renaming when the destination path already exists

adferrand:
On Linux, you can invoke `os.rename(src, dst)` even if `dst` already exists. In this case, destination file will be atomically replaced by the source file.

On Windows, this will lead to an `OSError` because changes are not atomic. This cause `certbot renew` to fail in particular, because the old certificate configuration needs to be replace by the new when a certificate is effectively renewed.

One could use the cross-platform function `os.replace`, but it is available only on Python >= 3.3.

This PR add a function in `compat` to handle correctly this case on Windows, and delegating everything else to `os.rename`.

### DIFF
--- a/certbot/compat.py
+++ b/certbot/compat.py
@@ -64,6 +64,23 @@ def os_geteuid():
         # Windows specific
         return 0
 
+def os_rename(src, dst):
+    """
+    Rename a file to a destination path and handles situations where the destination exists.
+
+    :param str src: The current file path.
+    :param str dst: The new file path.
+    """
+    try:
+        os.rename(src, dst)
+    except OSError as err:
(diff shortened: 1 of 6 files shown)


adferrand:
Yes, my implementation is not satisfactory.

I have a better solution: `os.replace`. From the specification discussion (https://bugs.python.org/issue8828), it is the best atomic rename that can be done on Windows, at least for renaming occurring on the local system.

However it is available only on Python 3.3. But hey!, minimal Python supported for Windows will be 3.4 ^^

So I will fallback to `os.replace`.

adferrand:
All right @bmw, I committed a new implementation.

PR #6420, opened 2018-10-16
Add route53 cname follow

wimnat:
This PR allows for the route53 plugin to follow CNAME records for an acme-challenge.  This allows users to auth a domain in a different domain account if they don't have permissions for the original domain account.

This is discussed in the following post:

https://community.letsencrypt.org/t/renew-using-dns-01-challenge/53498/10


wimnat:
Build is failing, probably because I've not written any tests to cover my new code.

I'm happy to try writing tests if someone can provide assistance on what I need to do.

PR #6422, opened 2018-10-16
Note about running on web server, not PC

### LINKED ISSUE #2224: Notify user they should be running command on target server
First I read this:

`http://letsencrypt.readthedocs.org/en/latest/using.html#installation`

and then I ran this on my Mac dev machine:

`./letsencrypt-auto`

which told me to run this:

`./letsencrypt-auto certonly`

which resulted in this error:

```
IMPORTANT NOTES:

   Domain: mydomain.com
   Type:   urn:acme:error:tls
   Detail: Failed to connect to host for DVSNI challenge
```

After further reading, I'm guessing it's because I'm not running the command on `mydomain.com`?

I think the ideal UX would be:
- When `./letsencrypt-auto` is run, ask user whether they are on the server they want a cert for.
- If yes, continue
- If no, say "This process is much easier if you run this command on the server you want a cert for. Can you run this command on that server?"
  - If yes, print installation options – `git clone` or `yum/apt/etc install letsencrypt` , then exit
  - If no, start the manual process

schoen:
Documentation change only: two reminders about where Certbot is meant to be used, to avoid any confusion caused by people installing locally on laptops.

Fixes #2224.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@ Anyone who has gone through the trouble of setting up a secure website knows wha
 
 How you use Certbot depends on the configuration of your web server. The best way to get started is to use our `interactive guide <https://certbot.eff.org>`_. It generates instructions based on your configuration settings. In most cases, you’ll need `root or administrator access <https://certbot.eff.org/faq/#does-certbot-require-root-administrator-privileges>`_ to your web server to run Certbot.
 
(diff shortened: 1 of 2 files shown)


schoen:
Fixes #2224.

Issue #6430, opened 2018-10-19
Consolidate tox.ini and tox-win.ini


bmw:
The reason they are different files are all of the custom bash scripts we use in our tests. I think we should reconsider our approach here and try to get by with just Python and `tox`.

adferrand:
Already done, in #6435.

PR #6435, opened 2018-10-22
[Windows|Unix] Rewrite bash scripts for tests into python

adferrand:
Certbot relies heavily on bash scripts to deploy a development environment and to execute tests. This is fine for Linux systems, including Travis, but problematic for Windows machines.

This PR converts all theses scripts into Python, to make them platform independant. 

As a consequence, `tox-win.ini` is not needed anymore, and `tox` can be run indifferently on Windows or on Linux using a common `tox.ini`. AppVeyor is updated accordingly to execute tests for `acme`, `certbot` and all dns plugins. Other tests are not executed as they are for Docker, unsupported Apache/Nginx/Postfix plugins (for now) or not relevant for Windows (explicit Linux distribution tests or pylint).

Another PR will be done on certbot website to update how a dev environment can be set up.


adferrand:
In response to #6430

adferrand:
Another note: I did not convert the bash scripts used to make release deployments. It should not be necessary, as it is done on a Linux systems. But for the consistency, maybe they should, in a future PR.

adferrand:
All settled.

@bmw, if you want to prioritize reviews on my PRs, take this one first. It is quite structuring for the CI, and everything else will benefit from the CI convergence done here between Linux and Windows.

adferrand:
Also updated the documentation. Project certbot/website seems to get its content from certbot/certbot as a submodule, so I assume the submodule is updated in place during the website deployment associated to a new certbot release. Let me know if I need to do further modifications on this.

tools/pip_install.py
```diff
@@ -0,0 +1,91 @@
...
+        fd.write(merged_requirements)
+
+def call_with_print(command, cwd=None):
+    print(command)
+    subprocess.call(command, shell=True, cwd=cwd or os.getcwd())
+
+def main(args):
+    tools_path = find_tools_path()
+    try:
+        working_dir = tempfile.mkdtemp()
```
bmw: nit: I'd move this out of the `try` block.

tools/pip_install.py
```diff
@@ -0,0 +1,91 @@
...
+import tempfile
+
+import merge_requirements as merge_module
+import readlink
+
+def find_tools_path():
+    return os.path.dirname(readlink.main(__file__))
+
+def certbot_oldest_processing(tools_path, args, test_constraints):
+    print(args)
```
bmw: Why do we print the arguments here?

tests/modification-check.py
```diff
@@ -0,0 +1,122 @@
...
+except ImportError:
+    from urllib import urlretrieve
+
+def find_repo_path():
+    return os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
+# We do not use filecmp.cmp to not be sensitive to CRLF/LF during comparison
+def compare_files(path_1, path_2):
+    l1 = l2 = True
+    with open(path_1, 'rU') as f1, open(path_2, 'rU') as f2:
```
bmw: nit: It appears that [open](https://docs.python.org/3/library/functions.html#open) uses "universal newline mode" by default in Python 3. If it has this behavior in Python 3 and we're not planning to support Windows with Python 2, can the `U` here be removed? I'm largely only interested because using `U` is deprecated in newer versions of Python.

tox.cover.py
```diff
@@ -0,0 +1,85 @@
...
+    'certbot_nginx': 97,
+    'certbot_postfix': 100,
+    'letshelp_certbot': 100
+}
+
+def cover(package):
+    threshold = COVER_THRESHOLDS[package]
+
+    if not threshold:
+        raise ValueError('Unrecognized package: {0}'.format(package))
```
bmw: I don't think we'll ever hit this line. If `package` isn't in `COVER_THRESHOLDS`, won't we blow up with a `KeyError`? Maybe we should use `COVER_THRESHOLDS.get` above.

tox.cover.py
```diff
@@ -0,0 +1,85 @@
...
+
+    pkg_dir = package.replace('_', '-')
+
+    skip_projects_on_windows = [
+        'certbot-apache', 'certbot-nginx', 'certbot-postfix', 'letshelp-certbot']
```
bmw: Same comment about maybe making this a global setting.

tools/_venv_common.py
```diff
@@ -0,0 +1,56 @@
...
+
+    subprocess_with_print(' '.join([
+        python_executable, os.path.normpath('./letsencrypt-auto-source/pieces/pipstrap.py')]))
+    command = [python_executable, os.path.normpath('./tools/pip_install.py')]
+    command.extend(args)
+    subprocess_with_print(' '.join(command))
+
+    print('-------------------------------------------------------------------')
+    print('Please run the following command to activate developer environment:')
+    print('source {0}/bin/activate'.format(venv_name))
```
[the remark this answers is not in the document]
adferrand: I forgot to push a local commit that was handling that. It is done now.

tests/modification-check.py
adferrand: Yes you are right. But it means that you should never have Python 2.7 as the default environment in your `PATH` on Windows, otherwise it will fail.

What bother me is that the requirements we are talking about here are for the tests workflow, not Certbot itself. In theory on Linux, you can run `tox` on any Python version, and envs requiring a specific version (like testing Certbot) specifies this directly in `tox.ini`. Then it will not be the case on Windows.

But for the sake of not using a deprecated option, I can live with that ^^. I just put a comment on the function to explain this.

tools/pip_install.py
adferrand: Debug code. Removed.

tools/pip_install.py
adferrand: Done.

tests/modification-check.py
```diff
@@ -0,0 +1,122 @@
...
+                errors = True
+
+    return errors
+
+def main(args):
+    repo_path = find_repo_path()
+    errors = False
+
+    try:
+        temp_cwd = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
adferrand: Done.

tools/install_and_test.py
```diff
@@ -0,0 +1,58 @@
...
+        current_command.append(requirement)
+        call_with_print(' '.join(current_command))
+        pkg = re.sub(r'\[\w+\]', '', requirement)
+        pkg = pkg.replace('_', '-')
+
+        if pkg == '.':
+            pkg = 'certbot'
+
+        try:
+            temp_cwd = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
adferrand: Done.

tools/install_and_test.py
```diff
@@ -0,0 +1,58 @@
...
+                .format(arg)))
+        else:
+            new_args.append(arg)
+
+    for requirement in new_args:
+        current_command = command[:]
+        current_command.append(requirement)
+        call_with_print(' '.join(current_command))
+        pkg = re.sub(r'\[\w+\]', '', requirement)
+        pkg = pkg.replace('_', '-')
```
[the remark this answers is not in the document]
adferrand: Yes, some mixed up things when I converted from bash. Entries are like `certbot-apache[dev]`, and needs to be reformated to `certbot_apache` when passed to `--pyargs` to match the actual module name.

I made the correction. But in general, I think we can do better than just replacing characters. What if a distribution exposes several modules, or a module that is not a direct conversion of the distribution name. Not needed for now, but a possible improvement to make the test workflow more resilient.

tox.cover.py
adferrand: Indeed. Corrected.

tox.cover.py
adferrand: Done.

tools/install_and_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+def main(args):
+    if os.environ.get('CERTBOT_NO_PIN') == '1':
+        command = [sys.executable, '-m', 'pip', '-q', '-e']
+    else:
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        command = [sys.executable, os.path.join(script_dir, 'pip_install_editable.py')]
+
+    skip_projects_on_windows = [
+        'certbot-apache', 'certbot-nginx', 'certbot-postfix', 'letshelp-certbot']
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -45,7 +45,7 @@ def check_output(*popenargs, **kwargs):
                 cmd = popenargs[0]
             raise CalledProcessError(retcode, cmd)
         return output
-from sys import exit, version_info
```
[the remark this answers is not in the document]
adferrand: Done: erikrose/pipstrap#19

tools/install_and_test.py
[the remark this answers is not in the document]
bmw: I agree. Do you want to make an issue to track work on that?

tests/modification-check.py
bmw: That's true.

For people exactly following our [current developer instructions](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client), that shouldn't be a problem because we have people create and activate a virtual environment containing Certbot and developer tools like `tox`, but because we don't plan on Certbot supporting Python 2 on Windows, Windows developers will have to be using a Python 3 environment. People more familiar with Python development could expect they can just run `tox` they have installed elsewhere without activating the environment though.

With your argument, I don't have a strong opinion here. If you have a significant preference for leaving `U` here, feel free to add it back.

In general, if you disagree with a change we recommend, please push back! It's certainly possible you thought of something we didn't which was the case here.

bmw:
Most recent commit looks good to me!

I'd like to wait briefly on merging this to see if Erik has a chance to quickly review erikrose/pipstrap#19. To date, we've never deviated from Erik's copy. If he merges that PR and bumps the pipstrap version number, can you update this PR accordingly?

Alternatively, if all tests work without the changes to pipstrap, we could undo those changes for now and copy them over once they land in the pipstrap repo. Either approach is fine with me.

tools/install_and_test.py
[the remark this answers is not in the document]
adferrand: Done: #6453

tests/modification-check.py
adferrand: Not a disagreement, more a work to find a compromise for various pro/con seen by different people around a design decision ^^ I will not hesitate on that, for sure !

I think you are right. You code a piece of software for a Python version, you use it for everything that is related to it. We just need to be very clear on the developer documentation about Python 3 and Windows when it will be official.

adferrand:
No, the modifications are really needed to make `_venv_common.py` behave correctly. Before, PATH was modified on-the-fly for the pipstrap call, in order to target the correct `pip` executable. But that is something you can do in bash, not in windows batch/powershell. At least, not from a platform-independant `subcommand` call.

So we will wait for upstream response ^^

adferrand:
@bmw, Erik made a new version of pipstrap. The upstream pipstrap code is synchronous with this PR.

bmw:
Awesome. Can you copy over the exact file from his repo (with the version number bump and everything)? With that change, I'll merge this PR.

adferrand:
Done.

PR #6436, opened 2018-10-22
Remove unnecessary setup in .travis.yml

bmw:
From the `dpkg` man page:
```
-s, --status package-name...
                  Report status of specified package.
```
We don't need to print the status of `augeas` on every run. This was more useful when we were installing `augeas` from outside of OS repos to get a recent enough version years ago.

We also have no macOS builds in this Travis config so let's remove the macOS setup. It's still in [test-everything .travis.yml](https://github.com/certbot/certbot/blob/c99688f175aa29d012f0663dc060bc17d96fc8ce/.travis.yml) where we actually have these tests.


bmw:
This PR is no longer needed due to #6661.

PR #6437, opened 2018-10-23
Fix macOS tests

bmw:
`test-everything` tests are failing again. See https://travis-ci.org/certbot/certbot/builds/444765864.

This time it's due to Homebrew conflicting with packages already installed by Travis. To fix this, we no longer upgrade packages if they've already been installed by Homebrew. With the current macOS image, this causes us to skip Python installation entirely and avoid the problem.

As a side effect, I had to make the Python 3 tests for any Python 3 version rather than explicitly Python 3.7. I personally think this is OK.

Other changes in this PR are:

* Cleaning up some cruft. `[ $TRAVIS_OS_NAME == linux ] && dpkg -s libaugeas0)` isn't needed as explained in #6436.
* I split installation into a separate shell script rather than trying to cram it all into a single line of the Travis config.


bmw:
Nice! Thanks for the review.

Does that mean we no longer need to run `brew link`?

joohoi:
I think it's good to leave the `brew link` in place, but we were able to remove the `brew upgrade python` that was previously required as a workaround to address some consistency issues with default Travis Python installation.

Issue #6302, opened 2018-08-15
Update our vendored Augeas httpd.lens for Apache plugin


joohoi:
We have a few outstanding issues with Apache configuration parsing, that have been fixed in upstream httpd lens, we should update it to fix issues like #6247

Issue #6247, opened 2018-07-22
Parsing error apache vhost.conf with nested "(" / ")" over several lines


datiscum:
My operating system is ( Ubuntu 18.04 LTS ):

I installed Certbot with ( OS package manager  ): certbot  0.23.0-1

I ran this command and it produced this output:
letsencrypt renew 

Certbot's behavior differed from what I expected because:
Cert not due for renewal, but simulating renewal for dry run
Could not choose appropriate plugin: The apache plugin is not working; there may be problems with your existing configuration.
The error was: PluginError('There has been an error in parsing the file /etc/apache2/sites-enabled/XXXXX.conf on line 248: Syntax error',)

<ElseIf  "(( %{REQUEST_METHOD} =  ) && ( =~ () )\
          || (  %{HTTP_REFERER} =~ ())\
          || ( (%{HTTP_REFERER} =~ ) && ( =~ ( ) ))\
          || ( (%{QUERY_STRING} =~ () = ) && (  =~  ))\
          || ( (%{HTTP_REFERER} =~ ) && (  =~  ))\
          || ( (-z %{HTTP_REFERER} || %{HTTP_REFERER} = ) &&  =~ )\
          || ( (-z %{HTTP_REFERER} || %{HTTP_REFERER} = ) &&  =~ ) **)**\
          || (( %{REQUEST_METHOD} = ) &&  ( =~ ))">

In the penultimate line at the end, I remove the")" the parser is happy, which is wrong.

schoen:
@joohoi, this looks like a possible Augeas grammar bug (including a to me somewhat obscure aspect of Apache configuration syntax), potentially an upstream problem. Could you please take a look at it?

joohoi:
@datiscum could you add the actual configuration snippet so I could try to reproduce the issue? If you don't want to make it public, you can send it to me via email: joona.hoikkala@eff.org

joohoi:
This is definitely a bug in the grammar. I got the configuration file in email, and was able to reproduce the issue. Certbot was able to correctly parse the file, after removing a empty comment line following the `<IfElse>` statement described in the issue description. I'm going to continue investigating fix for the issue, but I believe I was able to provide a way to mitigate the issue for now.

PR #6438, opened 2018-10-24
Update Augeas lens to fix some Apache configuration parsing issues

joohoi:
This PR updates the Augeas httpd lens used by Apache plugin, fixing some parsing errors and making the parsing more robust. My local tests against it were all successful, but I would like to see test farm tests ran against it.

Fixes: #6302
Fixes: #6247


joohoi:
Yeah, I'll add the changelog entry. Good to hear that the test farm was happy as well, as I personally wasn't able to find any flaws with manual testing.

PR #6442, opened 2018-10-25
Use the ACMEv2 newNonce endpoint when a new nonce is needed

### LINKED ISSUE #6092: Use the ACMEv2 newNonce endpoint when a new nonce is needed

ohemorange:
Also, add checking to the newNonce HEAD request, and check responses in general before attempting to save a nonce, for a better error message.

Fixes #6092.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -978,7 +978,11 @@ class ClientNetwork(object):  # pylint: disable=too-many-instance-attributes
         :raises .ClientError: In case of other networking errors.
 
         """
-        response_ct = response.headers.get('Content-Type')
+        try:
+            response_ct = response.headers.get('Content-Type')
+        except:
+            response_ct = None
+
         try:
             # TODO: response.json() is called twice, once here, and
             # once in _get and _post clients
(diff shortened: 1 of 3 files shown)


acme/acme/client.py
```diff
@@ -569,6 +569,7 @@ def __init__(self, directory, net):
         :param .messages.Directory directory: Directory Resource
         :param .ClientNetwork net: Client network.
         """
+        net.new_nonce_url = getattr(directory, "newNonce")
```
jsha: Modifying the `ClientNetwork` object as a side-effect of initializing `ClientV2` seems sub-optimal. Could we make this a constructor argument for `ClientNetwork`?

acme/acme/client.py
```diff
@@ -977,6 +979,7 @@ def _check_response(cls, response, content_type=None):
 
         """
         response_ct = response.headers.get('Content-Type')
+
```
jsha: Unrelated diff?

acme/acme/client.py
```diff
@@ -1106,10 +1109,15 @@ def _add_nonce(self, response):
...
-    def _get_nonce(self, url):
+    def _get_nonce(self, url, acme_version):
         if not self._nonces:
             logger.debug('Requesting fresh nonce')
-            self._add_nonce(self.head(url))
+            if acme_version == 1:
+                response = self.head(url)
+            else:
+                # request a new nonce from the acme newNonce endpoint
+                response = self._check_response(self.head(self.new_nonce_url), content_type=None)
```
jsha: I think this would throw a weird error if `new_nonce_url` is `None`. Can we check that case and raise an error that says what's wrong?

Issue #2500, opened 2016-02-18
letsencrypt-auto should perform version comparison before "upgrading" itself


pde:
Then we'll be safe from downgrade attacks. And people can run the copy from the `letsencrypt-auto-source` directory before its released.

bmw:
This actually came up in a discussion recently about features that other upgrade systems have and `certbot-auto` doesn't.

If we stick with `certbot-auto`, I think we should definitely consider doing this.

erikrose:
I think it's literally as simple as changing the comparison operator.

adferrand:
Hum. Line 1644 of `certbot-auto`/`letsencrypt-auto` on master:
`elif [ "$LE_AUTO_VERSION" != "$REMOTE_VERSION" ]; then`

For me, current version of `certbot-auto` is already compared to latest version available, and self-upgrade is triggered only if versions differ. At least, on a Ubuntu 14.04 LTS, I see the expected behavior.

Or did I miss something ?

bmw:
The change Peter suggested here is to prevent an attacker from being able to cause certbot-auto to downgrade itself to an older version.

The way this would work is someone would intercept your connection to PyPI. This connection uses HTTPS which thwarts trivial attempts, however, there are a few ways a more sophisticated attacker could accomplish this in theory. If they are able to do this, they could report to `certbot-auto` that an older version of Certbot is the latest one available. When this simple inequality check of `[ "$LE_AUTO_VERSION" != "$REMOTE_VERSION" ]` runs, if `certbot-auto` is not the same version as the version reported by the attacker, `certbot-auto` will happily replace itself with an older version of `certbot-auto` and Certbot. Ideally, `certbot-auto` should only accept upgrades to a newer version.

This is unlikely to happen and the damage caused by downgrading Certbot is likely to be relatively minimal (as all I can think of off the top of my head is causing renewal to fail), however, it's an attack that `certbot-auto` is not protected against.

adferrand:
Yes I understood that after my first response ^^

It makes sense also in particular when you want to test a built `certbot-auto` from the sources, where current version is `0.28.0.dev0` for instance, and will lead to an undesired downgrade. You cannot test directly new logic in `certbot-auto` basically.

But for attackers, well, I am not convinced: if an attacker has the ability to fake your connection to PyPI, it will be able also to advertize a false new version and provide from the fake URL a forged `certbot-auto` script. In this case, downgrade approach is not sufficient.

[...]

PR #6448, opened 2018-10-27
Protect certbot-auto against automated downgrades

adferrand:
Following discussion in #2500.

With current code, the certbot-auto self-upgrade process can make it actually to downgrade itself, because the comparison done is an equality test between local certbot-auto version and the remote one. This is a flaw for attackers, that could make certbot-auto break itself by falsely advertising it about an old version as the latest one available.

A function is added to make a more advanced comparison between version. Certbot-auto will upgrade itself only if the local version is strictly inferior to the latest one available. For instance, a version 0.28.0 will not upgrade itself if the latest one available on internet is 0.27.1. Similarly, non-official versions like 0.28.0.dev0 will never trigger a self-upgrade, to help development workflows.

This implementation relies only on the Python distribution installed by certbot-auto (supporting 2.7+) and basic shell operations, to be compatible with any UNIX-based system.


bmw:
@erikrose said he'd take a look at this.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -635,7 +663,12 @@ UNLIKELY_EOF
       error "WARNING: couldn't find Python $MIN_PYTHON_VERSION+ to check for updates."
     elif ! REMOTE_VERSION=`"$LE_PYTHON" "$TEMP_DIR/fetch.py" --latest-version` ; then
       error "WARNING: unable to check for updates."
-    elif [ "$LE_AUTO_VERSION" != "$REMOTE_VERSION" ]; then
+    fi
+
+    LE_VERSION_STATE=`CompareVersions "$LE_PYTHON" "$LE_AUTO_VERSION" "$REMOTE_VERSION"`
+    if [ "$LE_VERSION_STATE" = "UNOFFICIAL" ]; then
+      echo "Unofficial certbot-auto version detected, self-upgrade is disabled: $LE_AUTO_VERSION"
```
erikrose:
```suggestion
      say "Unofficial certbot-auto version detected, self-upgrade is disabled: $LE_AUTO_VERSION"
```

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -912,6 +912,34 @@ OldVenvExists() {
...
+# Otherwise, print "OUTDATED" if version 1 is outdated, or "UP_TO_DATE" if not.
+# This function relies only on installed python environment (2.x or 3.x) by certbot-auto.
+CompareVersions() {
+    "$1" - "$2" "$3" << "UNLIKELY_EOF"
+import sys
+from distutils.version import StrictVersion
+
+try:
+    current = StrictVersion(sys.argv[1])
+except:
```
erikrose: What exceptions are you trying to catch here? At the very least, we should probably say `except Exception`, just so you don't catch and ignore control-C presses. I suspect we could tighten it even more so as to not hide actual unanticipated errors.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -912,6 +912,34 @@ OldVenvExists() {
...
+    sys.stdout.write('UNOFFICIAL')
+    sys.exit()
+
+try:
+    target = StrictVersion(sys.argv[2])
```
erikrose: "new" or "remote" might be a clearer name here. The latter is consistent with the name we give it at the callsite (`$REMOTE_VERSION`).

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -912,6 +912,34 @@ OldVenvExists() {
     [ -n "$OLD_VENV_PATH" -a -f "$OLD_VENV_PATH/bin/letsencrypt" ]
 }
 
+# Given python path, version 1 and version 2, check if version 1 is outdated compared to version 2.
+# Unofficial versions (eg. 0.28.0.dev0) will be treated specifically by printing "UNOFFICIAL".
```
erikrose: Well, an unofficial version of $2 will say UNOFFICIAL. An unofficial version of $3 will say UP_TO_DATE. I think those are reasonable choices, but we should bring this comment up to date.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -912,6 +912,34 @@ OldVenvExists() {
...
+# This function relies only on installed python environment (2.x or 3.x) by certbot-auto.
+CompareVersions() {
+    "$1" - "$2" "$3" << "UNLIKELY_EOF"
+import sys
+from distutils.version import StrictVersion
+
+try:
+    current = StrictVersion(sys.argv[1])
+except:
+    sys.stdout.write('UNOFFICIAL')
```
erikrose: It's idiomatic to say merely `print('UNOFFICIAL')`, unless you have a reason not to. If you're dodging the parenthesis requirement of Python 3, instead `from __future__ import print_function` at the top.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -451,6 +451,34 @@ OldVenvExists() {
     [ -n "$OLD_VENV_PATH" -a -f "$OLD_VENV_PATH/bin/letsencrypt" ]
 }
 
+# Given python path, version 1 and version 2, check if version 1 is outdated compared to version 2.
+# Unofficial versions (eg. 0.28.0.dev0) will be treated specifically by printing "UNOFFICIAL".
+# Otherwise, print "OUTDATED" if version 1 is outdated, or "UP_TO_DATE" if not.
+# This function relies only on installed python environment (2.x or 3.x) by certbot-auto.
+CompareVersions() {
+    "$1" - "$2" "$3" << "UNLIKELY_EOF"
```
erikrose: This is a really clever way of delegating to another language without having to make a file!

letsencrypt-auto-source/letsencrypt-auto
adferrand: I am doing it intentionally, to avoid an endline that I would need to strip after when output of this function is assigned to the version variable.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Thanks a lot! I was tempted to do the same for the others temporary files that are python scripts, including your pipstrap, before remembering that I should not overcharge a PR with unrelated work.

But if you think it is quite elegant, I will be happy to do it in another PR!

letsencrypt-auto-source/letsencrypt-auto
adferrand: Yes `remote` is a good word.

letsencrypt-auto-source/letsencrypt-auto
adferrand: I update the inline comment.

PR #6455, opened 2018-10-31
Warnings are errors

### LINKED ISSUE #6080: Treat warnings as failures during testing
We should use something like https://docs.pytest.org/en/latest/warnings.html to have warnings treated as errors when running our tests with pytest so we can fix anything like deprecation warnings before it's a problem.

ohemorange:
Fixes #6080


adferrand:
In the class `TempDirTestCase` located in `certbot/tests/util.py`, there is a teardown process that tries to delete the temporary directory generated by tests suite that subclass `TempDirTestCase`. On Windows, this process will often fail because a lot of file descriptors are still opened in this temporary directory when teardown kicks in. 

I covered this situation with an onerror handler that allow the directory to not be deleted, and generates a warning instead. This kind of warning should become recognizable by making a specific subclass of `UserWarning` (or any other relevant subclass of warning builtin classes). And the pytest worflow should put these warning subclass in the whitelist, to not convert it into errors.

Otherwise, hundreds of errors will spawn during the Windows CI. This is a temporary situation, until I implement all the code design and refactoring that ensures that temporary directories can be deleted at the end of a test. But until then, theses warnings should be treated as warnings, not errors.

ohemorange:
Having done all this, I actually think that we should only run this in a nightly test-everything test. That way, workflow is not impeded when working on other issues. Also, then we can just run it under Python 3, and not have to worry about python 2 compatibility for the warning types.

acme/acme/client.py
```diff
@@ -32,11 +32,8 @@
 # for SSL, which does allow these options to be configured.
 # https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
 if sys.version_info < (2, 7, 9):  # pragma: no cover
-    try:
-        requests.packages.urllib3.contrib.pyopenssl.inject_into_urllib3()  # type: ignore
-    except AttributeError:
-        import urllib3.contrib.pyopenssl  # pylint: disable=import-error
-        urllib3.contrib.pyopenssl.inject_into_urllib3()
+    import urllib3.contrib.pyopenssl  # pylint: disable=import-error
```
ohemorange: Do we need a heavier assurance that `urllib3` will be properly available now?

ohemorange:
Bringing `requests` requirement back down; made https://github.com/certbot/certbot/issues/6467

acme/acme/crypto_util_test.py
```diff
@@ -10,11 +10,16 @@
 
 import josepy as jose
 import OpenSSL
+import pytest
 
 from acme import errors
 from acme import test_util
 from acme.magic_typing import List # pylint: disable=unused-import, no-name-in-module
 
+# turns all ResourceWarnings into errors for this module
```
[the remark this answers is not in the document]
ohemorange: warnings.warn("this comment is deprecated")

acme/acme/crypto_util_test.py
```diff
@@ -10,11 +10,16 @@
...
+if six.PY3:
+    pytestmark = pytest.mark.filterwarnings("ignore::ResourceWarning") # pragma: no cover
```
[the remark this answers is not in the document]
ohemorange: Almost all of the legitimate ResourceWarnings I cleaned up were in tests, with the exception of a couple one line `open().read()` situations in certbot. If we were diligent about that when reviewing code, I would be comfortable turning off the warning across the system. These acme tests are flaky about this, I think it might be a garbage collection thing but I didn't look too closely because they're inconsistent and also tests.

But, bad news, `ResourceWarning` only exists in Python 3, not 2. So we can't mention it in `pytest.ini`, because that's used across tests. If we wanted to ignore it across the board, we'd have to stick the ignore in explicitly for only python3 tests, which doesn't seem great.

Another thing fixed most easily by dropping 2.7 support!

certbot/crypto_util.py
```diff
@@ -458,8 +458,12 @@ def sha256sum(filename):
     :rtype: str
     """
     sha256 = hashlib.sha256()
-    with open(filename, 'rU') as file_d:
-        sha256.update(file_d.read().encode('UTF-8'))
```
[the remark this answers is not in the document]
ohemorange: nice

letsencrypt-auto-source/pieces/dependency-requirements.txt
```diff
@@ -150,9 +150,9 @@ pytz==2015.7 \
     --hash=sha256:fbd26746772c24cb93c8b97cbdad5cb9e46c86bbdb1b9d8a743ee00e2fb1fc5d \
     --hash=sha256:99266ef30a37e43932deec2b7ca73e83c8dbc3b9ff703ec73eca6b1dae6befea \
     --hash=sha256:8b6ce1c993909783bc96e0b4f34ea223bff7a4df2c90bdb9c4e0f1ac928689e3
-requests==2.12.1 \
-    --hash=sha256:3f3f27a9d0f9092935efc78054ef324eb9f8166718270aefe036dfa1e4f68e1e \
-    --hash=sha256:2109ecea94df90980be040490ff1d879971b024861539abb00054062388b612e
+requests==2.20.0 \
```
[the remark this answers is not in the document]
ohemorange: nice catch, thanks

acme/acme/client.py
ohemorange: Can't leave it as is in master -- the reason `requests` now has dependencies is that it removed its vendorized packages. The line as it is in master errors with the version of `requests` that we pin.

ohemorange:
Happy to block on #6435.

ohemorange:
Also, whatever we do, before merging I do think errors should only be turned on for test-everything tests. Regular PRs shouldn't block on DeprecationWarnings.

bmw:
Now that #6435 has landed, do the test failures go away if `master` is merged into the PR?

adferrand:
Yes, they will go away. 

On a side note however, #6487 should be merged first on master to re-establish the original fail-fast behavior.

ohemorange:
This is ready for review

acme/acme/client.py
ohemorange: ah. right. that's what it was. lint complains:
```
E: 36, 8: Module 'requests.packages' has no 'urllib3' member (no-member)
```
Technically true, but only in theory, not in practice. You comfortable with me squelching the error?

bmw: I think so! The attribute exists but `lint` doesn't understand the code that magically adds it.

Do you have concerns about doing it?

ohemorange: Nope, good enough for me.

PR #6456, opened 2018-10-31
Dummy AWS credentials for Route53 tests to prevent outbound connections

joohoi:
Boto3 / botocore library has a feature that tries to fetch AWS credentials from IAM if a set of credentials isn't available otherwise. This happens when boto loops through different credential providers in order to find the keys. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=912103

This PR simply adds dummy environmental variables for the tests that will be picked up by the credential provider iterator in order to prevent making outbound connections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Certbot adheres to [Semantic Versioning](http://semver.org/).
 * Fix ranking of vhosts in Nginx so that all port-matching vhosts come first
 * Correct OVH integration tests on machines without internet access.
 * Stop caching the results of ipv6_info in http01.py
+* Test fix for Route53 plugin to prevent boto3 making outgoing connections.
 
 ## 0.27.1 - 2018-09-06
 
--- a/certbot-dns-route53/certbot_dns_route53/dns_route53_test.py
+++ b/certbot-dns-route53/certbot_dns_route53/dns_route53_test.py
@@ -1,5 +1,6 @@
 """Tests for certbot_dns_route53.dns_route53.Authenticator"""
 
+import os
 import unittest
 
 import mock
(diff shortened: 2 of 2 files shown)


bmw:
Thanks for working on this @joohoi. I'll review this PR soon.

What AWS setup do I need to reproduce the bug in Debian?

joohoi:
The funny thing here is that I wasn't able to reproduce the issue, but was able to figure out what's going on in there by following the stack trace, a short explanation follows:

If credentials are not defined when initializing boto3 client in python code, it attempts to figure them out, this is the list of different credential providers (sorry for ugly paste, they are directly from ipdb shell):

```
[<botocore.credentials.EnvProvider object at 0x7f11403fb358>, <botocore.credentials.AssumeRoleProvider object at 0x7f114039ce48>, <botocore.credentials.SharedCredentialProvider object at 0x7f114039ce10>, <botocore.credentials.ProcessProvider object at 0x7f114039cdd8>, <botocore.credentials.ConfigProvider object at 0x7f114039cda0>, <botocore.credentials.OriginalEC2Provider object at 0x7f114039ccf8>, <botocore.credentials.BotoProvider object at 0x7f114039cd30>, <botocore.credentials.ContainerProvider object at 0x7f1140392860>, <botocore.credentials.InstanceMetadataProvider object at 0x7f114039ceb8>]
```

Apparently on Debian build system it reaches the last one (InstanceMetadataProvider), which tries to connect to AWS to fetch the credentials from IAM.

On initialization InstanceMetadataFetcher gets proxy settings from env vars:
    https://github.com/boto/botocore/blob/develop/botocore/utils.py#L265

And that's my assumption what's going on in the Debian test environment which this PR will fix.

joohoi:
This should be good to go now!

PR #6457, opened 2018-10-31
 Increase the default propagation time from 1200sec to 1800sec.

gopalraha:
According to official Linode DNS documentation the recommended time for complete DNS propagation is 30min or 1800seconds for more info see the official Linode DNS documentation https://www.linode.com/docs/platform/manager/dns-manager/#edit-records)

Be sure to edit the `master` section of `CHANGELOG.md` with a line describing this PR before it gets merged.

PR #6458, opened 2018-10-31
 Increase the default propagation time from 1200sec to 1800sec.

gopalraha:
According to official Linode DNS documentation the recommended time for complete DNS propagation is 30min or 1800seconds for more info see the official Linode DNS documentation https://www.linode.com/docs/platform/manager/dns-manager/#edit-records)

### DIFF
--- a/certbot-dns-linode/certbot_dns_linode/dns_linode.py
+++ b/certbot-dns-linode/certbot_dns_linode/dns_linode.py
@@ -29,7 +29,7 @@ class Authenticator(dns_common.DNSAuthenticator):
 
     @classmethod
     def add_parser_arguments(cls, add):  # pylint: disable=arguments-differ
-        super(Authenticator, cls).add_parser_arguments(add, default_propagation_seconds=1200)
+        super(Authenticator, cls).add_parser_arguments(add, default_propagation_seconds=1800)
         add('credentials', help='Linode credentials INI file.')
 
     def more_info(self):  # pylint: disable=missing-docstring,no-self-use


bmw:
@sydneyli, looks like you reviewed a similar patch before in #6457. Are you able to take a look at this as well?

stefantalpalaru:
> According to official Linode DNS documentation the recommended time for complete DNS propagation is 30min or 1800seconds for more info see the official Linode DNS documentation

That's a ballpark guesstimate for DNS change propagation to ISP-specific domain name servers - those that most people use when browsing the web.

Certbot queries the Linode name servers directly, so that approximation does not apply. I would be interested if you ever had a DNS challenge fail with the default 1200s interval.

mnordhoff:
Linode uses Cloudflare DNS Firewall, so there is an additional edge cache delay of 30 seconds to 15 minutes.

https://www.linode.com/community/questions/11459/linodes-cloudflare-dns-firewall-settings#answer-52625

stefantalpalaru:
> Linode uses Cloudflare DNS Firewall, so there is an additional edge cache delay of 30 seconds to 15 minutes.

It doesn't sound like it's additional. That must be where the 15 minutes delay comes from in the first place, since Linode's default TTL is 24 hours and CF cuts that to its configured maximum of 15 minutes.

mnordhoff:
No, they update the zones on the backend servers every 15 minutes. The Cloudflare edge delay is separate from / on top of that.

stefantalpalaru:
Don't you think we'd notice rather quickly if the maximum delay was 30 minutes?

sydneyli:
These docs aren't entirely clear on what they mean by "propagation"-- in another section, it recommends waiting up to [48 hours](https://www.linode.com/docs/platform/manager/dns-manager/#wait-for-propagation)-- which is how long you might wait for the records to propagate across the internet, like @stefantalpalaru mentioned.

@gopalraha, do you find that waiting 20 minutes is leading to renewal failures? I merged a similar PR (#6320) since folks were experiencing failures. Let's get a definitive answer from the Linode folks/forums about this that we can point to and be happy with.

trinopoty:
Linode staff also confirms that it's 30 min. See the discussion on [Linode forum](https://www.linode.com/community/questions/17296/linode-dns-propagation-time).
In my personal experience, it takes 15 min though.

stefantalpalaru:
You were wrong to ask about "propagation time to the public internet" because we query the Linode name servers directly and the staff member was wrong to give you a 30 minutes interval that makes no sense.

trinopoty:
I performed some experiments when I saw this PR and, personally, I've yet to see any change take more than 15 minutes to reflect publicly.
How about we not increase the default timeout and let the user adjust it manually as needed.

kaceo:
I would like to add my insight as my experience using Linode's DNS.  

I have my own Bind running on one Linode as SOA and use the Linode DNS as secondaries.  Everytime I update TXT records in my Bind, it takes more than 15-30 mins to update the secondaries, unless I manually trigger the dashboard.  Since I have 30 domains to renew, these delays are tedious. 
But propagation to the rest of the Internet is quick, e.g. to 8.8.8.8, 8.8.4,4 and 1.1.1.1 happens within 2 mins, although many other DNS does not seem to fetch the correct data until at least 10 mins later.

I think the uncertainty of how long it takes TXT record to be updated across the Internet is the most annoying problem of using Letsencrypt SSL.  When I renew my domains, I start by setting the delays to 5 mins, then this usually renews about a third of the domains; then I increase it to 20mins, and get another third; and finally I have to set it to 40mins to get the rest of the domains.  Since the renew script processes one domain at a time, it takes me up to 3 days to complete all my domains.

I think if the Letsencrypt verification protocol updates the TXT record, and checks only the SOA server, it would have a much more deterministic strategy to get the verification done within a few seconds.  I have used Certbot for a year now and I really don't enjoy the ordeals of renewing certs every 90 days.

mnordhoff:
If Linode is being used as a secondary, their authoritative nameservers will update as soon as they receive a NOTIFY or at the refresh interval given in your SOA record. If you make a change and send notifies, you can have all of the backend servers updated within a few seconds.

But the Cloudflare DNS Firewall caching still applies.

m0namon:
I'm going to close this PR since we couldn't determine whether people are encountering this issue in practice. If you are, please submit an issue for it!

PR #6465, opened 2018-11-02
Indicate DNS plugin package names may vary

gerrard00:
I was confused when trying to install the Digital Ocean DNS plugin because the name is actually prefixed with the python version.

If my suggested text is acceptable this should resolve https://github.com/certbot/website/issues/362

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -193,7 +193,13 @@ Certbot's DNS plugins.
 These plugins are not included in a default Certbot installation and must be
 installed separately. While the DNS plugins cannot currently be used with
 ``certbot-auto``, they are available in many OS package managers and as Docker
-images. Visit https://certbot.eff.org to learn the best way to use the DNS
+images.
+
+Note that the names of the OS packages may not match the names of the
+plugins as listed below due to package naming standards. For example the package
+for the CloudflareDNS plugin is actually ``python3-certbot-dns-cloudflare``.
+
+Visit https://certbot.eff.org to learn the best way to use the DNS
 plugins on your system.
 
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -193,7 +193,13 @@ Certbot's DNS plugins.
 These plugins are not included in a default Certbot installation and must be
 installed separately. While the DNS plugins cannot currently be used with
 ``certbot-auto``, they are available in many OS package managers and as Docker
-images. Visit https://certbot.eff.org to learn the best way to use the DNS
```
osirisinferi: I'm not sure if removing the link to the official `certbot` homepage (i.e. where all the basic "how to's" are) is wise. From there, users will be presented with the guide which will either let them use a PPA for example. Or not. Or `certbot-auto` / Docker.

docs/using.rst
```diff
@@ -193,7 +193,13 @@ Certbot's DNS plugins.
...
+images.
+
+Note that the names of the OS packages may not match the names of the
+plugins as listed below due to package naming standards. For example the package
+for the CloudflareDNS plugin is actually ``python3-certbot-dns-cloudflare``.
```
osirisinferi: "the package". Which distribution? In the Ubuntu PPA? Or official package? Perhaps add a little bit of extra information to the example.

sydneyli:
Hi @gerrard00, thanks for helping out!
We landed a more complete fix for this in certbot/website#387, which should be rolled into the next website release. I'm closing this PR for now, but if there's something you still find confusing that wasn't addressed by certbot/website#387 feel free to open an issue on that repository.

gerrard00:
Not a problem!

Issue #6319, opened 2018-08-24
Stop preferring TLS-SNI in the Apache, Nginx, and standalone plugins


bmw:
We should still offer the challenge for now, but we should change the order they are returned in `get_chall_pref` so HTTP-01 comes before TLS-SNI-01. TLS-SNI-01 is going away.

DimaVIII:
Yes absolutely necessary! Just run into this issue.
``` The client lacks sufficient authorization :: Incorrect validation certificate for tls-sni-01 challenge. Requested XXXXXXXXXX.acme.invalid from XXX.XXX.XXX.XXX:443. Received 1 certificate(s), first certificate had names "hostmaster", blog.XXX.XX (tls-sni-01):```

If there is a self-signed Cert installed on the IP of the the host machine, challenges will fail. 
This forced me to include "--preferred-challenges http" in every certbot call.

bmw:
We're planning on blocking this on https://github.com/certbot/certbot/issues/5763 and https://github.com/certbot/certbot/issues/5764.

PR #6468, opened 2018-11-03
Log warning about TLS-SNI deprecation in Certbot

ohemorange:

For  #6319.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -2276,6 +2276,9 @@ class ApacheConfigurator(augeas_configurator.AugeasConfigurator):
                 http_doer.add_chall(achall, i)
             else:  # tls-sni-01
                 sni_doer.add_chall(achall, i)
+                logger.warning("TLS-SNI-01 is deprecated, and will stop working in February 2019. "
+                    "Remove tls-sni from the --preferred-challenges flag to test issuance with "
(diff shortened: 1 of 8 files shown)


bmw:
I was initially thinking of something like https://github.com/certbot/certbot/commit/20bcb1181094cf127e4fb4a73d0a15b1637c9242. We could instead do it in [challb_to_achall](https://github.com/certbot/certbot/blob/v0.27.1/certbot/auth_handler.py#L358) where it logs each challenge we're doing, but I'm not sure if we want to log it for every challenge rather than just the first occurrence of it.

What do you think? Do you prefer your approach?

ohemorange:
I really don't think it matters either way.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -2276,6 +2276,9 @@ def perform(self, achalls):
                 http_doer.add_chall(achall, i)
             else:  # tls-sni-01
                 sni_doer.add_chall(achall, i)
+                logger.warning("TLS-SNI-01 is deprecated, and will stop working in February 2019. "
```
bmw: I don't have a strong preference, but I'd personally like us to remove the date here as things like that can change but a released version of Certbot cannot.

ohemorange:
I'm seeing that we do check if `pref_challs` is set before using it, where are you talking about?

PR #6480, opened 2018-11-07
Change default privkey permissions while preserving group permissions

### LINKED ISSUE #1473: archive/domain/privkeyN.pem is set to 0644 instead of 0600 (or 0440)

sydneyli:
Fixes #1473.
 * writes `privkey.pem` to 0600 by default for new lineages
 * on renewals where a new privkey is generated, preserves group mode and gid

Things this PR does not do:
 * we talked about forcing 0600 on privkeys when a Certbot upgrade is detected. Instead, this PR only creates new lineages with the more restrictive permission to prevent renewal breakages.
 * this doesn't solve many of the problems mentioned in #1473 that are not directly related to the title issue!

#### Be sure to edit the `master` section of `CHANGELOG.md` with a line describing this PR before it gets merged.

ok


docs/using.rst
```diff
@@ -717,6 +717,10 @@ The following files are available:
      put it into a safe, however - your server still needs to access
      this file in order for SSL/TLS to work.
 
+.. note:: As of Certbot version 0.29.0, private keys for new certificate
```
bmw: At least when building the docs locally, the formatting here is a little funky. It looks like:
screen shot 2018-11-28 at 3 45 58 pm

certbot/storage.py
```diff
@@ -1120,24 +1122,32 @@ def save_successor(self, prior_version, new_cert,
...
-            with open(target["privkey"], "wb") as f:
+            with util.safe_open(target["privkey"], "wb", chmod=BASE_PRIVKEY_MODE) as f:
                 logger.debug("Writing new private key to %s.", target["privkey"])
                 f.write(new_privkey)
+            # If the previous privkey in this lineage has an existing gid or group mode > 0,
+            # let's keep those.
+            group_mode = stat.S_IMODE(os.stat(old_privkey).st_mode) & \
+                (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP)
+            mode = BASE_PRIVKEY_MODE | group_mode
+            os.chown(target["privkey"], -1, os.stat(old_privkey).st_gid)
```
Camusensei: This breaks certbot on Windows: there is no os.chown on windows and this results in:

> AttributeError: module 'os' has no attribute 'chown'

PR #6485, opened 2018-11-08
[URGENT] Fix the CI system

adferrand:
Ok. I broke the Certbot CI system with #6435.

It is about the exit codes that are returned from the various scripts in `tools` during `tox` execution.
Indeed, `tox` relies on the non-zero exit code from a given script to know that something failed during the execution.

Previously, theses scripts were in bash, and a bash script returns an exit code that is the higher code returned from any of the command executed by the script. So if any command return a non-zero (in particular `pylint` or `pytest`), then the script return also non-zero.

Now that these scripts are converted into python, `pylint` and `pytest` are executed via `subprocess`, that returns the exit code as variables. But if theses codes are not handled explicitly, the python script itself will return zero if no python exception occured. As a consequence currently, Certbot CI system is unable to detect any test error or lint error, because there is no exception in this case, only exit codes from the binaries executed.

This PR fixes that, by handling correctly the exit code from the most critical scripts, `install_and_test.py` and `tox.cover.py`, but also all the scripts that I converted into Python and that could be executed in the context of a shell (via `tox` or directly for instance).

This PR should be merged as soon as possible.

Really sorry for the regression.

### DIFF
--- a/certbot/tests/compat_test.py
+++ b/certbot/tests/compat_test.py
@@ -19,3 +19,7 @@ class OsReplaceTest(test_util.TempDirTestCase):
 
         self.assertFalse(os.path.exists(src))
         self.assertTrue(os.path.exists(dst))
+
+    def test_fail(self):
+        # And a trailing white space 
+        raise Exception('Test')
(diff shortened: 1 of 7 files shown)


adferrand:
This first commit will fail. Because there is an error on `master`. This will enlight the fact that CI system is corrected, as it will detect again regressions.

Then I will make at least three more commits:
* one to correct the detected errors that are present on `master`
* one to put on purpose a pylint error to enlight the capability of the CI system to fail on pylint errors
* one to remove this pylint error, and theorically make this PR mergeable.

PR #6493, opened 2018-11-10
Update tools/venv3.py to support py launcher on Windows

adferrand:
I put the old PR description below, for information purpose, because I rewrote entirely the PR in the last commits, on this old description is not relevant anymore.

Following some inconsistencies occurred during by developments, and in the light of #6508, it decided to wrote a PR that will take fully advantage of the conversion from bash to python to the development setup tools.

This PR adresses several issues when trying to use the development setup tools (`tools/venv.py` and `tools/venv3.py`:
* on Windows, `python` executable is not always in PATH (default behavior)
* even if the option is checked, the `python` executable is not associated to the usually symlink `python3` on Windows
* on Windows again, really powerful introspection of the available Python environments can be done with `py`, the Windows Python launcher
* in general for all systems, `tools/venv.py` and `tools/venv3.py` ensures that the respective Python major version will be used to setup the virtual environment if available. However, no check is done do the compatible supported minor version. For instance, Python 2.6 or Python 3.3 will not raise any exception during the setup, even if the virtual environment is effectively not compatible and not supported by Certbot
* finally, the best and first candidate to test should be the Python executable used to launch the `tools/venv*.py` script. It was not relevant before because it was shell scripts, but do it is.

The logic is shared in `_venv_common.py`, and will be called appropriately for both scripts. In priority decreasing order, python executable will be search and tested:
* from the current Python executable, as exposed by `sys.executable`
* from any python or pythonX (X as a python version like 2, 3 or 2.7 or 3.4) executable available in PATH
* from the Windows Python launched `py` if available

> On Windows, standard installation procedure is to not put python executable in `PATH` (either `python` or `python3`), but the Python launcher instead (`py`).
> 
> This PR modifies `tools/venv3.py` to test also presence of the Python launcher in `PATH`, then get the actual Python 3.x executable path to use it as a parameter in `virtualenv`.
> 
> I did not put any logic for `tools\venv.py` because Python 2.x is not supported for Certbot on Windows. However, the code is designed in the way that even if `py` is launched with Python 2.x as default, calling `tools/venv3.py` ensures that it is Python 3.x (if exists) that will be used to generate the virtual environment.
> 
> This PR allows also python paths that contain spaces (Linux/Windows, py2x/py3x).

### DIFF
--- a/tools/_venv_common.py
+++ b/tools/_venv_common.py
@@ -59,7 +59,7 @@ def main(venv_name, venv_args, args):
         # Windows specific
         print('---------------------------------------------------------------------------')
         print('Please run one of the following commands to activate developer environment:')
-        print('{0}\\bin\\activate.bat (for Batch)'.format(venv_name))
+        print('{0}\\Scripts\\activate.bat (for Batch)'.format(venv_name))
         print('.\\{0}\\Scripts\\Activate.ps1 (for Powershell)'.format(venv_name))
         print('---------------------------------------------------------------------------')
     else:
--- a/tools/_venv_common.py
+++ b/tools/_venv_common.py
@@ -30,7 +30,7 @@ def find_python_executable(python_major):
     :param int python_major: the Python major version to target (2 or 3)
     :rtype: str
     :return: the relevant python executable path
-    :raise: RuntimeError if no relevant python executable path could be found
+    :raise RuntimeError: if no relevant python executable path could be found
     """
     python_executable_path = None
 
@@ -43,22 +43,25 @@ def find_python_executable(python_major):
     versions_to_test = ['2.7', '2', ''] if python_major == 2 else ['3', '']
     for one_version in versions_to_test:
         try:
-            output = subprocess.check_output('python{0} --version'.format(one_version),
-                                             universal_newlines=True, shell=True,
-                                             stderr=subprocess.STDOUT)
+            one_python = 'python{0}'.format(one_version)
+            output = subprocess.check_output([one_python, '--version'],
+                                             universal_newlines=True, stderr=subprocess.STDOUT)
             if _check_version(output.strip().split()[1], python_major):
-                return 'python{0}'.format(python_major)
+                return subprocess.check_output([one_python, '-c',
+                                                'import sys; sys.stdout.write(sys.executable);'],
+                                               universal_newlines=True)
         except (subprocess.CalledProcessError, OSError):
             pass
 
     # Last try, with Windows Python launcher
     try:
-        output_version = subprocess.check_output('py -{0} --version'.format(python_major),
-                                                 universal_newlines=True, shell=True,
-                                                 stderr=subprocess.STDOUT)
+        env_arg = '-{0}'.format(python_major)
+        output_version = subprocess.check_output(['py', env_arg],
+                                                 universal_newlines=True, stderr=subprocess.STDOUT)
         if _check_version(output_version.strip().split()[1], python_major):
-            command = 'py -{0} -c "import sys; sys.stdout.write(sys.executable);"'.format(python_major)
-            return subprocess.check_output(command, universal_newlines=True, shell=True)
+            return subprocess.check_output(['py', env_arg, '-c',
+                                            'import sys; sys.stdout.write(sys.executable);'],
+                                           universal_newlines=True)
     except (subprocess.CalledProcessError, OSError):
         pass
 
--- a/tools/_venv_common.py
+++ b/tools/_venv_common.py
@@ -8,6 +8,79 @@ import glob
 import time
 import subprocess
 import sys
+import re
+
+VERSION_PATTERN = re.compile(r'^(\d+)\.(\d+).*$')
+
+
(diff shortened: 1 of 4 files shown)


tools/_venv_common.py
```diff
@@ -8,11 +8,91 @@
...
+    Will test, in decreasing priority order:
+    * the current Python interpreter
+    * 'pythonX' executable in PATH (with X the given major version) if available
+    * 'python' executable in PATH if available
+    * Windows Python launcher 'py' executable in PATH if available
+    Incompatible python versions for Certbot will be evicted (eg. Python < 3.5 on Windows)
+    :param int python_major: the Python major version to target (2 or 3)
+    :rtype: str
+    :return: the relevant python executable path
+    :raise: RuntimeError if no relevant python executable path could be found
```
bmw: It's a little odd, but the `:` after `raise` here should be moved after `RuntimeError`. See http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists for examples.

tools/_venv_common.py
```diff
@@ -8,11 +8,91 @@
...
+    if _check_version('{0}.{1}.{2}'.format(
+            sys.version_info[0], sys.version_info[1], sys.version_info[2]), python_major):
+        return sys.executable
+
+    # Second try, with python executables in path
+    versions_to_test = ['2.7', '2', ''] if python_major == 2 else ['3', '']
+    for one_version in versions_to_test:
+        try:
+            output = subprocess.check_output('python{0} --version'.format(one_version),
+                                             universal_newlines=True, shell=True,
```
bmw: It doesn't matter much, but other than being able to pass the command a string rather than a list, is there any reason to set `shell=True` here? If not, my personal preference is to remove it.

tools/_venv_common.py
```diff
@@ -8,11 +8,91 @@
...
+                                             stderr=subprocess.STDOUT)
+            if _check_version(output.strip().split()[1], python_major):
+                return 'python{0}'.format(python_major)
+        except (subprocess.CalledProcessError, OSError):
+            pass
+
+    # Last try, with Windows Python launcher
+    try:
+        output_version = subprocess.check_output('py -{0} --version'.format(python_major),
+                                                 universal_newlines=True, shell=True,
```
bmw: It doesn't matter much, but other than being able to pass the command a string rather than a list, is there any reason to set `shell=True` here? If not, my personal preference is to remove it.

tools/_venv_common.py
```diff
@@ -8,11 +8,91 @@
...
+                                                 stderr=subprocess.STDOUT)
+        if _check_version(output_version.strip().split()[1], python_major):
+            command = 'py -{0} -c "import sys; sys.stdout.write(sys.executable);"'.format(python_major)
+            return subprocess.check_output(command, universal_newlines=True, shell=True)
```
bmw: It doesn't matter much, but other than being able to pass the command a string rather than a list, is there any reason to set `shell=True` here? If not, my personal preference is to remove it.

tools/_venv_common.py
```diff
@@ -8,11 +8,91 @@
...
+
+    # Second try, with python executables in path
+    versions_to_test = ['2.7', '2', ''] if python_major == 2 else ['3', '']
+    for one_version in versions_to_test:
+        try:
+            output = subprocess.check_output('python{0} --version'.format(one_version),
+                                             universal_newlines=True, shell=True,
+                                             stderr=subprocess.STDOUT)
+            if _check_version(output.strip().split()[1], python_major):
+                return 'python{0}'.format(python_major)
```
bmw: I think we instead want to return `'python{0}'.format(one_version)` here.

We can maybe help avoid errors like this by storing the name of the executable in a variable in the body of this loop.

adferrand: Corrected

tools/_venv_common.py
adferrand: Corrected

tools/_venv_common.py
adferrand: Corrected

tools/_venv_common.py
adferrand: Corrected

tools/_venv_common.py
adferrand: Corrected

tools/_venv_common.py
```diff
@@ -8,11 +8,94 @@
...
+                return subprocess.check_output([one_python, '-c',
+                                                'import sys; sys.stdout.write(sys.executable);'],
+                                               universal_newlines=True)
+        except (subprocess.CalledProcessError, OSError):
+            pass
+
+    # Last try, with Windows Python launcher
+    try:
+        env_arg = '-{0}'.format(python_major)
+        output_version = subprocess.check_output(['py', env_arg],
```
bmw: It looks like we lost `--version` in the most recent commit here. When testing this, the script hangs indefinitely.

adferrand: Oups ! Indeed. I corrected the second step and broke the third... Corrected.

PR #6497, opened 2018-11-12
[Windows] (WIP) Security model for files permissions

### LINKED ISSUE #6356: Fix the permissions on files and directories on Windows
See the discussion at https://github.com/certbot/certbot/pull/6296#discussion_r212695925.

adferrand:
This PR implements a security model to handle correctly the files permissions on Windows.

The complete specification is discussed in #6356, so I will not repeat its details here.

Essentially, the goal is to provide appropriate DACLs on files generated by Certbot to ensure that sensible files and directories cannot be seen on Windows by everyone, but only by specific users. More generally, the implementation here translates POSIX modes (like 0644 or 0700) into DACLs that make sense for Windows.

For instance, 0700 will lead to file that is readable/editable/executable only by the user who runs Certbot and the Administrators.

What I will add here is the impact of this security model for Certbot development.

Indeed, we are talking about critical matter, as private keys for instance should never be accessible by normal users. However, Python, and in particular its standard `os` module, has a very poor support of Windows file permissions mechanism. The direct usage of `os` may silently make files exposable to anyone. In particular, wrong usages of `os.mkdir` or `os.chmod` could have disastrous consequences on the security model.

So I created a module, `certbot.compat.os`, that is intended to replace `os` in order to protect the security model using relevant wrappers around critical functions. Appart that, `certbot.compat.os` exposes any method from `os` and can be used as a valid replacement.

Of course, any call to `os` has been replaced by `certbot.compat.os` in this PR. But it is not sufficient. Indeed in the future a contributor that is not aware of Windows particularities may use inappropriatly the `os` module, and break the security without the tests or the code reviewers notice it.

As a consequence, after discussions with @bmw, I decided to enforce during the CI to never use the standard `os` module in certbot projects (except `acme`). To do that, a new script `tools/enforce-coding-style.py` replaces the current actions on `lint` tox environment. This script will launch `pylint` and fail if errors are detected as usual. But it will also check if `os` is imported in any relevant python file, and fail if this kind of import is detected.

So any developer that would import directly `os` will make the tests fail and see with an appropriate report what is wrong.

Appart that, `pywin32` dependency is added to certbot on Windows. This dependency provides binary builds starting with Python 3.5. Even if it is possible to compile `pywin32` for Python 3.4, it requires a full-fledged .NET environment. So for convenience and security (where `pywin32` is officially supported and tested), we should advertise that Certbot is only available with Python >= 3.5.

Finally, certbot requires now administrative privileges for any operation except `--help`.

Fixes #6356


setup.py
```diff
@@ -48,6 +48,15 @@ def read_file(filename, encoding='utf8'):
     'zope.interface',
 ]
 
+# Add pywin32 on Windows platform to handle low-level system calls.
+# This dependency needs to be added dynamically to avoid to be here
+# when certbot-oldest tests are launched.
+# Indeed, associated with --constraint, pip will ignore platform_system directive and try
```
ohemorange: What do you mean "associated with --constraint"?

setup.py
```diff
@@ -48,6 +48,15 @@ def read_file(filename, encoding='utf8'):
...
+# to install pywin32 on Unix systems.
+# his would fail and break certbot-oldest tests.
```
ohemorange:
```suggestion
# This would fail and break certbot-oldest tests.
```

setup.py
```diff
@@ -48,6 +48,15 @@ def read_file(filename, encoding='utf8'):
     'zope.interface',
 ]
 
+# Add pywin32 on Windows platform to handle low-level system calls.
+# This dependency needs to be added dynamically to avoid to be here
```
ohemorange:
```suggestion
# This dependency needs to be added dynamically to avoid being here
```

certbot/compat/os.py
```diff
@@ -0,0 +1,154 @@
...
+
+import errno
+import os as std_os
+import sys
+
+from acme.magic_typing import Callable, Union  # pylint: disable=unused-import, no-name-in-module
+from certbot.compat import security
+
+# Monkey patch ourselves to get os attributes that are not in __all__ (so not from os import *).
+ourselves = sys.modules[__name__]
```
ohemorange: why not just install ourselves in sys.modules as `os`? then we don't have to replace the call everywhere, and don't need to add the lint changes.

certbot/compat/security.py
```diff
@@ -0,0 +1,307 @@
+"""Compat module to handle files security on Windows and Linux"""
+from __future__ import absolute_import
+
+import os
+import stat
+
+try:
+    import pwd  # pylint: disable=import-error
+except ImportError:  # pragma: no cover
+    pwd = None  # type: ignore
```
ohemorange: mypy has an optional type, does that not work here?

adferrand: I used the same approach that is done on several locations in the Certbot codebase, But indeed, it is smarter, I will try that.

adferrand: But it does not work... you cannot assign a type on an import statement. Then, any try to add a type at the second statement (like `pwd` = None) that would solve the issue with mypy will fail, and of course mypy will fail without `# type: ignore`.

setup.py
adferrand: pyYes it is not obvious from this comment. When `pip_install.py` executes to build a oldest environment, it will invoke pip using a specific constraint file specified by `--constraint`. This as a consequence to ignore every environment marker, and try to install `pywin32` on a Unix platform.

This code fix that.

adferrand: I refactored a little the comment.

certbot/compat/os.py
adferrand: Well, it is a really good question. It was the first thing in my mind: reinjecting a corrected os module in the modules pool of current Python instance. Then it would be used seamlessly by any Certbot module.

There is a lot of problems with that approach eventually.

First, the corrected os module would be used not only by certbot modules, but also by every dependency of Certbot. As you can see, the functionalities added are very specific to Certbot needs. Because in fact, Unix and Windows security models are just too different to have a global and bijective representation in both platforms (otherwise it would be implemented in the standard library). So it means that this model could not be suited for the dependencies, and add unexpected behavior to them. This security model, and so the associated modules, should only be used for the sole Certbot purpose.

Second, the os module is swapped only when the relevant code to manipulate sys.modules is called. Obviously, it should be added to the `__init__.py` of `certbot` package. But also to any other package, like the ones from dns plugins and authenticators, to avoid to break any permission. And ensure this, is the same problem as ensuring that standard `os` is not usedm and should be corrected by a lint process. So lint rules may or may not be used to enforce this policy, but using an injection in sys.modules, do not remove on its own the requirement to check the code.

Finally, even if implicit approach fit quite well in Python, I think there is a limit, related to good or bad design decisions. Making implicit modifications of one of the most used standard library feels wrong to me. I largely prefer to be explicit about using a different `os` module, and make it clear from the code.

ohemorange: > First, the corrected os module would be used not only by certbot modules, but also by every dependency of Certbot. As you can see, the functionalities added are very specific to Certbot needs. Because in fact, Unix and Windows security models are just too different to have a global and bijective representation in both platforms (otherwise it would be implemented in the standard library). So it means that this model could not be suited for the dependencies, and add unexpected behavior to them. This security model, and so the associated modules, should only be used for the sole Certbot purpose.

Is it ever the case that for any of our dependencies that create files, we would want files to be created correctly in Certbot but incorrectly in their own code?

> Second, the os module is swapped only when the relevant code to manipulate sys.modules is called. Obviously, it should be added to the __init__.py of certbot package. But also to any other package, like the ones from dns plugins and authenticators, to avoid to break any permission. And ensure this, is the same problem as ensuring that standard os is not usedm and should be corrected by a lint process. So lint rules may or may not be used to enforce this policy, but using an injection in sys.modules, do not remove on its own the requirement to check the code.

If it's true that our dependencies get the patched version of `os`, then our plugins will also get it and won't need to patch it themselves. This keeps the changed code to the smallest possible footprint, which is good for maintenance (if we need to change it in the future) and for packaging (if a unix packager doesn't want the windows support code in there at all).

> Finally, even if implicit approach fit quite well in Python, I think there is a limit, related to good or bad design decisions. Making implicit modifications of one of the most used standard library feels wrong to me. I largely prefer to be explicit about using a different os module, and make it clear from the code.

This isn't something that callers of `os` within Certbot should have to worry about. We should expect that people using the functions shouldn't have to understand how they work to be able to call them, and leave the behavior similar enough that it works as expected. Requiring callers to import a different class will just lead to changing the line, not using the functions any differently, which is just an implementation blocker.

certbot/compat/security.py
ohemorange: enough of a reason to ignore it for me

certbot/compat/os.py
adferrand: OK I took the wrong examples with plugins, forgotting that plugins do not call certbot, is it certbot that call plugins through the dynamic plugins discovery mechanism of python.

That was not the point I wanted to enlight. 

Monkeypatching a standard library is not a question about design, it is a question about wrong code. If you monkeypatch the standard library, it does not make it effective only for your own code, but for any code that use yours.

It means that any external developper, for any project and any context, import certbot package in one of its own module, will see a standard module be silently swapped to a custom on, starting from the first import of the certbot package, and for the rest of the runtime execution.

The consequences are widely unknown, completely out of our range. It is just not an option. We can discuss ways to ease usage of this specific module, but monkeypatching is not a solution.

Apart that, developers are not required to understand the compat.os module. It is an implementation detail that claim externally as a contract to be a valid replacement of the standard os module for certbot. I do not think that it is overcomitting, as it does not add a single line compared to a call to the standard os, and that a CI process is able to do the job to check this specific subtility. Compared with the thousands of unit test out there, and all the dynamic mocks called from them, for me it is not a big deal ^^

ohemorange:
Hey @adferrand, thanks for the link to the issue, that helps to explain the high-level decisions here and Windows permissions model. As I'm going through reviewing, I'm finding myself wishing for a description of the approach to translate into Windows-ese at the function level. That probably looks like an explanation of why the approach for each `os` function is appropriate. Any chance you could help me out by writing those up?

(And as a bonus, I imagine those will make great function documentation for posterity.)

adferrand:
@ohemorange, I updated the documentation in the `compat.os` module, to clarify what is going on in the overriden functions to protect the security model on Windows.

adferrand:
I extracted the first part to #6833

adferrand:
The plan is the following:
* PR1: `certbot.compat.security` and unit tests, that contains the core functionnalities to apply the Windows security model, and the existing Linux one
* PR2: other modules from `certbot.compat` and unit tests, in particular `certbot.compat.os`, that subclass standard `os` with improved security model for Windows from `certbot.compat.security`
 (eg, `os.chmod` uses `security.apply_mode`)
* PR3: in all certbot code, replace imports of standard `os` to `certbot.compat.os`, and call directly specific functions from `certbot.compat.security` when appropriate
* PR4: build a quality code process in the CI, ensuring that standard `os` module is not used in certbot code

bmw:
In the vein of Erica's comment at https://github.com/certbot/certbot/pull/6821#pullrequestreview-218609294, I wonder if there's not a better way for us to break this up. I think reviewing a bunch of low level code without context or significant tests makes things a little more difficult than they need to be.

(I also want to push back a little now on the idea that this low level code doesn't need unit tests because it happens to have coverage from its use at a higher level. I personally think functions like many of the ones in this PR benefit most from unit tests. We can address this issue more later once the higher level issues are resolved if you like.)

With this in mind, I'd like to propose an alternative way to break this code up which is:

1. Do any refactoring with `certbot.compat` and existing code. This PR should not include any new code.
2. Add `certbot.compat.os`, make other code in Certbot use it, fix up packaging so the oldest supported Certbot version in each plugin is correct, and add tooling to ensure `certbot.compat.os` is used over `os`. This PR should not modify any of the functions in `os`.
3. In separate PRs, either add a top level function in `certbot.compat.security` that is used outside of the `compat` module (although see what I wrote below about this) or replace a function in `certbot.compat.os`. Each PR should should include the low level functions in `certbot.compat.security` that it needs and in the case of top level `certbot.compat.security` functions, fix up any code that is supposed to call the function so it is actually used. Each high level function added may not need its own PR, but unless there's a good reason to include multiple functions in the same PR, why not break it up? The first of these PRs will add the dependency on `pywin32` so it is beneficial to make the actual change in the `os` module here one of the simpler ones.

I know moving around code isn't the most exciting thing, but let's take the time to make sure you, me, and @ohemorange are all on the same page now and we shouldn't have to revisit this decision in the future.

What do you think?

(One aspect of the higher level design here that Erica and I talked about is whether or not code outside of `certbot.compat` should be using the `security` module. A few of the functions in `compat.os` tell you to go use code in `security`, but we could alternatively put this code in `certbot.compat.os` directly.

The benefit of doing this is the `os` module is self contained and there's a clear architecture in the code where `compat.os` is the consumer of `security` and all other callers go through `compat.os` instead of sometimes calling `security` directly. On the other hand, it breaks the abstraction that `compat.os` is more or less the `os` module by adding functions not found in the standard library's `os` module (although that abstraction is already somewhat broken by functions like `chown` yelling at you) and makes it harder for readers of the code to see the added functions they can use because they are in a larger file mixed with wrappers around the `os` code that they ideally should be able to ignore.

I don't think we should block all work on this PR on this question, but it's something we should think/talk about before getting to the later PRs.)

adferrand:
OK @bmw, I agree with you about the top-to-bottom approach, that feels more natural, and will be easier to review, as all code is always effectively used.

For os vs security external call, I really do not know what I prefer. Both approaches have their pros and cons. I decided to write things in the scope of the best readability, with explicit import of os, explicit use of security. This is because we are talking about security here, and I would like to avoid in the future that a contributor breaks something, and the reviewer does not catch it. This is helped by been as readable as possible.

So with that approach in mind, I would tend to keep explicit calls to security module when the method API does not fit an existing one in os module.

bmw:
> OK @bmw, I agree with you about the top-to-bottom approach, that feels more natural, and will be easier to review, as all code is always effectively used.

Thanks for understanding. I think it will help all of us in the long run by making reviews much easier and faster.

Again, I think we can start working on the initial PRs before figuring if we want to restructure things a bit with the `security` module, but I think we should continue that conversation. I have an idea which I can propose, but I'm more interested in hearing what others think.

@ohemorange, do you want to share your thoughts or should we set up a quick call to chat about it sometime soon?

ohemorange:
My main thought about always calling into `os` vs sometimes also calling directly into `security` (which `os` calls into) is that I want the most clarity possible for the user. I tend to think that having well-defined, separate layers is the cleanest way to do that. Like with python's TCPServer vs calling directly into sockets; we want users to not need to know about the underlying layer, except in cases where the higher layer we provide (`os`) really doesn't work for them for some reason.

adferrand:
In the extent of what you said @ohemorange, and following my slight preference to call security module directly when needed, I had also an alternative in mind. 

Instead of delegating some calls from certbot.compat.os to certbot.compat.security, maybe we could just overload them in os to raise an exception, with a message/documentation indicating what is the relevant method from security to use. 

This way, the os module would contain only the logic to delegate to the real os module, or raise an exception when the method is forbidden. And every effective call to the specific security logic in certbot would be done by directly invoking the security module.

bmw:
Isn't what you described @adferrand already in this PR? I'm also not sure it addresses Erica's idea to make a clear architecture between the modules. You still have both `certbot.compat.os` and other code calling into the `security` module.

My personal opinion is I see the value in a clear hierarchy, but I also don't like putting more stuff in `compat.os`. I want to propose another structure that could hopefully help us address the different concerns we have. I briefly proposed this to @ohemorange before, but I think I can better articulate why I like this design better than putting everything in `compat.os`.

The general structure is to make three modules like:
```
certbot code
---------calls into-------
certbot.compat.os or certbot.compat.extended_os
---------calls into-------
certbot.compat._security
```
`certbot.compat.os` only contains functions that are found in the standard library `os` module (and private helper functions only used in that module).`certbot.compat.extended_os` contains new functions such as those `certbot.compat.os` tells you to use instead (and private helper functions only used in that module). `certbot.compat._security` contains functions used by both `compat.os` and `compat.extended_os` and should not be used outside of the `compat` module.

The reason I like this better is I think `compat.os` has to include a lot of code the user of the module should be able to ignore. We have to make all of `os` accessible through it and override all problematic functions. If we write this module well, its user shouldn't have to look at any of it. As far as they know, they're using the `os` module. If you try to use a function you shouldn't, you'll get an error when running tests which will tell you the function to use instead. You should never have to read the code.

With this in mind, I think putting new functions in `compat.os` makes the functions available for you to use harder to find because it's surrounded by other code that you should be able to ignore. Sure, we could try and structure the file in a way to make this distinction clear, but I think the easiest way to make this separation is by putting it in a new file.

Thoughts?

adferrand:
The thing is, @bmw, current implementation of `certbot.compat.os` in my PR is already what you described: it does not add anything to the standard os module, and you can just use it without knowing its internal. The only two functions that will throw an exception on Linux (and Windows, but it was throwing an `AttributeError` anyway from standard `os`) are `geteuid` on `chown`. In this case, there is a message indicating what is the relevant method to use from certbot.compat.security`.

So in this situation, what is the difference between calling `certbot.compat.security` and `certbot.compat.os_extended`? Yes behind there is `certbot.compat._security` in your proposition, but it is about internal organisation, it does not change the API exposed to the user.

About my alternative, it is not what I wrote in this PR. It is about making all methods that needs a specific logic, to behave like `chown`: throwing an exception indicating which method should be used instead (from `security`/`os_extended`).

To be clear also in my also on my current implementation, your proposition or my alternative, it is always to have a `certbot.compat.os` module that expose the same API than the standard `os`.

What differs in my alternative, is that you do not ever use a modified method comparing to standard `os` when calling from `certbot.compat.os`. Either you have it working untouched, or it throws indicating an alternative from the other compat module. As of, this contract put a different separation of concerns, that may be interesting.

And as it, I think it is a response to @ohemorange, where in her example, you would call `TCPServer` methods except when it is not suitable. The non suitability in my alternative is indicated by the method raising an meaningful exception.

bmw:
The difference between what I proposed and what you currently have is that in this PR both code outside of the `compat` package and `compat.os` use `security` while in what I proposed there is a clear distinction between the functions in `security` used to implement `compat.os` and the API exposed to the rest of Certbot.

> What differs in my alternative, is that you do not ever use a method with alternate logic compared to standard os when calling from certbot.compat.os. Either you have it working untouched, or it throws and error indicating an alternative from the other compat module. As of, this contract put a different separation of concerns than current PR, that may be interesting.

I didn't understand before but I had the same idea and was going to propose it once we agreed on a structure we liked. I personally like this idea because I think changing the functionality of the standard library which programmers become used in custom ways to is a bad thing to do. You're forcing people to unlearn or incorrectly learn how the library works.

I think my ideal solution would be to not have `compat.os` and instead just ensure the problematic functions in `os` are never used but I think writing tools to help enforce that is really hard if you let other code use the unwrapped `os` module. Since I don't think that's feasible, I personally like the idea of using `compat.os` with the linter you wrote just to ensure problematic functions are not used and then put all new functions in a separate module.

The last I heard from @ohemorange she was still in favor of extending the API of the `os` module rather than adding a new module, but if she's been swayed by our arguments here and she also likes the idea of never changing the functionality of `os` and instead putting the functionality in a new module, it seems like we can happily avoid adding the three modules I suggested above in favor of just a simplified `compat.os` module and `security` whose public API is meant to be used by the rest of Certbot and not to implement other functions in `compat`.

ohemorange:
> The last I heard from @ohemorange she was still in favor of extending the API of the os module rather than adding a new module, but if she's been swayed by our arguments here and she also likes the idea of never changing the functionality of os and instead putting the functionality in a new module, it seems like we can happily avoid adding the three modules I suggested above in favor of just a simplified compat.os module and security whose public API is meant to be used by the rest of Certbot and not to implement other functions in compat.

I don't actually have a strong preference between extending the API of `os` in the same module vs extending it in a separate module, as long as that separate module isn't the same module as the one used to implement the functions in `os` (ie `security`). If `os` always calls into `_security` instead of `security`, that would likely enforce the abstraction levels that are my main underlying motivation here.

bmw:
@ohemorange, what do you think of Adrien's proposal which to recap was:

1. `certbot.compat.os` has no significant implementation. It's just a wrapper around the `os` module that raises an exception if you try to use a method whose behavior is inadequate for our needs. The message in the raised exception tells you the method you should use instead.
2. `certbot.compat.security` adds the custom functions to fulfill our needs with a public API and private helper functions to help it do that.

ohemorange:
Are the methods in `security` going to have the same signature as the original functions in `os` and just different names? (I assume the answer is "some will and some won't".) The thing I'm worried about is creating a whole new paradigm of custom os-like functions, rather than having the tried-and-true `os` API that python devs know and love when possible. Making people need to learn a whole new OS interface to write Certbot code isn't ideal. So if, when possible, we stick to the existing paradigm, that makes sense to me.

The only other thing is if we do it that way, not sure that `security` is the most accurate name for that module, since it's conceptually more about providing os-level functionality, we just happen to have to reimplement mainly security-focused functions.

bmw:
I argued against changing the behavior of `os` functions above by saying:

> I think changing the functionality of the standard library which programmers become used in custom ways to is a bad thing to do. You're forcing people to unlearn or incorrectly learn how the library works.

In my opinion this includes adding code that hide the standard library's limitations. We might write a function like `os.mkdir` (but call it `make_directory`) whose `mode` parameter is meaningful on Windows, but this not the behavior people should expect from the standard library. We could argue that's what its behavior should be, but since it's not, I personally prefer not to change it in subtle ways even if they seem nice to us who at the moment have less experience with Python on Windows.

As for the name of the new module, I'm personally pretty open. The [standard library os module](https://docs.python.org/3/library/os.html) is described as "miscellaneous operating system interfaces" and it talks about how many OS specific functions are implemented in other standard library modules with more specific names. I'd personally think we should try to come up with a more specific name that we like, but if we can't, I'm fine with something referring back to the `os` module.

bmw:
I talked to Erica out of band and she says she's OK with the decisions in my previous post.

bmw:
Perhaps I should have reraised concerns and suggestions for the module name `security` before you started splitting this PR, but I was going to propose `access_control` as all functions either deal with file system permissions or process user IDs.

In fact, everything is about file system permissions except for `get_current_user` so we could move that function elsewhere if we can come up with a name we like better with the reduced scope, but I personally like keeping the contents as is and just changing the name to `access_control`.

What do you all think?

adferrand:
The method `get_current_user` could go to `certbot.compat.misc`. I am not convinced by `security`, but neither I am for `access_control`. It seems to generalist, as all of this is about files permissions, and I prefer to avoid two-words module, that are very uncommon in Python.

In Java, you find this kind of files operations in the `java.io.File` class. So I propose to name the module `certbot.compat.file`.

bmw:
What about `filesystem`? (Both spellings as a single word or two words are accepted.)

This is somewhat nicer than `access_control` because while the reason we have functions like `mkdir` are permissions problems, making directories isn't exactly related to access control. I also personally like it better than `file` because `file.mkdir` feels a little off to me as well.

If we do this, in addition to moving `get_current_user` out of `security`, I think we should probably also move some functions into it from `certbot.compat.misc` such as `os_rename`.

ohemorange:
`permissions`, maybe?

adferrand:
The word `permissions` is still too generic in my opinion. However I could go for `filesystem`, that has the advantage of allowing the reintegration of `rename` method indeed.

adferrand:
This draft PR is now synchronized with #7233, and so can be closed.

PR #6500, opened 2018-11-13
Workaround for old pip versions that are not exposed as importable modules.

adferrand:
This PR uses the modifications done in erikrose/pipstrap#21 to work around very old versions of pip that cannot be used with a `python -m pip` call.


adferrand:
As Certbot still uses a release version of Pipstrap, I think the PR is ready for review.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -57,7 +57,7 @@ def check_output(*popenargs, **kwargs):
     from urllib.parse import urlparse  # 3.4
 
 
-__version__ = 2, 0, 0
+__version__ = 1, 5, 0
```
bmw: It looks like this is reverting `pipstrap` slightly too far as I updated it to 1.5.1 in e0ae356aa35adf22d154113e06dd01409df93bba. You can find that version of pipstrap at https://github.com/erikrose/pipstrap/blob/1.5.1/pipstrap.py.

tools/_venv_common.py
```diff
@@ -91,18 +92,18 @@ def _check_version(version_str, major_version):
...
-def get_venv_python(venv_path):
+def get_venv_bin_path(venv_path):
     python_linux = os.path.join(venv_path, 'bin/python')
     if os.path.isfile(python_linux):
-        return python_linux
+        return os.path.join(os.getcwd(), os.path.dirname(python_linux))
     python_windows = os.path.join(venv_path, 'Scripts\\python.exe')
     if os.path.isfile(python_windows):
-        return python_windows
+        return os.path.join(os.getcwd(), os.path.dirname(python_windows))
```
bmw: nit: Is the goal here to get an absolute path to the Python binary? If so, maybe we should just use [os.path.abspath](https://docs.python.org/3/library/os.path.html#os.path.abspath).

tools/_venv_common.py
```diff
@@ -91,18 +92,18 @@ def _check_version(version_str, major_version):
...
+    print('+ {0}'.format(subprocess.list2cmdline(cmd)) if isinstance(cmd, list) else cmd)
+    subprocess.check_call(cmd, env=env, shell=shell)
 
 
-def get_venv_python(venv_path):
+def get_venv_bin_path(venv_path):
     python_linux = os.path.join(venv_path, 'bin/python')
     if os.path.isfile(python_linux):
-        return python_linux
+        return os.path.join(os.getcwd(), os.path.dirname(python_linux))
```
bmw: nit: Is the goal here to get an absolute path to the Python binary? If so, maybe we should just use [os.path.abspath](https://docs.python.org/3/library/os.path.html#os.path.abspath).

tools/_venv_common.py
```diff
@@ -119,18 +120,19 @@ def main(venv_name, venv_args, args):
...
-        os.path.normpath('./letsencrypt-auto-source/pieces/pipstrap.py')))
-    subprocess_with_print('"{0}" {1} {2}'.format(
-        python_executable,
-        os.path.normpath('./tools/pip_install.py'),
-        ' '.join(args)))
+    command = [sys.executable, '-m', 'virtualenv', '--no-site-packages', '--setuptools', venv_name]
+    command.extend(shlex.split(venv_args))
+    subprocess_with_print(command)
+
+    # We execute the two following command in the context of the virtual environment, to install
```
bmw: nit: s/command/commands

letsencrypt-auto-source/pieces/pipstrap.py
adferrand: Indeed. Fallback to 1.5.1.

bmw:
Looks like my small nit at https://github.com/certbot/certbot/pull/6500#discussion_r234707839 was missed in the recent commit.

It also seems like something happened to some of the spacing in pipstrap. None of the changes affect the scripts behavior so I'm not too worried about it, but after checking out this branch, see:
```
$ diff letsencrypt-auto-source/pieces/pipstrap.py <(curl https://raw.githubusercontent.com/erikrose/pipstrap/1.5.1/pipstrap.py 2>/dev/null)
2a3
> 
7a9
> 
133a136
> 
135a139
> 
180c184
<     exit(main())
\ No newline at end of file
---
>     exit(main())
```

tools/_venv_common.py
adferrand: You are right.

tools/_venv_common.py
adferrand: You are right.

adferrand:
Sorry, it's late, I think I am a little tired. Second set of corrections with typo correction, I also sync with upstream on pipstrap about newline.

tools/_venv_common.py
adferrand: Also corrected

PR #6510, opened 2018-11-15
closes #6384: Handle dot-style wildcard domains like star-style domains

### LINKED ISSUE #6384: [Feature request] nginx plugin doesn't auto detect wildcard domains
Ubuntu 18.04.1

apt initially, then pip installed all certbot packages to see if issue was resolved

    sudo certbot --authenticator=dns-route53 --installer=nginx

    No names were found in your configuration files. Please enter in your domain

I have a single _index_ nginx conf containing a wildcard `server_name` which routes to all other 'location' confs based on path

Why can't the nginx plugin just detect that its a wildcard cert?

```
server {
    listen 80;
    server_name .example.com;

    root /var/www;
    index index.html;

    include conf.d/*.location;
}
```

ciarancourtney:
* e.g. mutate .example.com into *.example.com as early as possible

Not sure if this is the correct approach, but seems reasonable 

closes #6384

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-*
+* Handle dot-style wildcard domains like star-style domains
 
 ### Changed
 
--- a/certbot-apache/certbot_apache/tests/configurator_test.py
+++ b/certbot-apache/certbot_apache/tests/configurator_test.py
@@ -169,7 +169,7 @@ class MultipleVhostsTest(util.ApacheTest):
         self.assertEqual(names, set(
             ["certbot.demo", "ocspvhost.com", "encryption-example.demo",
              "nonsym.link", "vhost.in.rootconf", "www.certbot.demo",
-             "duplicate.example.com"]
+             "duplicate.example.com", "*.blue.purple.com"]
         ))
 
     @certbot_util.patch_get_utility()
@@ -188,7 +188,7 @@ class MultipleVhostsTest(util.ApacheTest):
         self.config.vhosts.append(vhost)
 
         names = self.config.get_all_names()
-        self.assertEqual(len(names), 9)
+        self.assertEqual(len(names), 10)
         self.assertTrue("zombo.com" in names)
         self.assertTrue("google.com" in names)
         self.assertTrue("certbot.demo" in names)
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -555,7 +555,12 @@ class NginxConfigurator(common.Installer):
         all_names = set()  # type: Set[str]
 
         for vhost in self.parser.get_vhosts():
-            all_names.update(vhost.names)
+            for server_name in vhost.names:
+                if util.is_wildcard_dot_domain(server_name):
+                    # expand .example.com into *.example.com and example.com
+                    all_names.update(('*' + server_name, server_name[1:]))
+                else:
+                    all_names.add(server_name)
 
             for addr in vhost.addrs:
                 host = addr.get_addr()
--- a/certbot-nginx/certbot_nginx/tests/configurator_test.py
+++ b/certbot-nginx/certbot_nginx/tests/configurator_test.py
@@ -83,11 +83,25 @@ class NginxConfiguratorTest(util.NginxTest):
     def test_get_all_names(self, mock_gethostbyaddr):
         mock_gethostbyaddr.return_value = ('155.225.50.69.nephoscale.net', [], [])
         names = self.config.get_all_names()
-        self.assertEqual(names, {
-            "155.225.50.69.nephoscale.net", "www.example.org", "another.alias",
-             "migration.com", "summer.com", "geese.com", "sslon.com",
-             "globalssl.com", "globalsslsetssl.com", "ipv6.com", "ipv6ssl.com",
-             "headers.com"})
+        self.assertEqual({
+            '*.example.com',
+            '*.www.example.com',
+            '*.www.foo.com',
+            '155.225.50.69.nephoscale.net',
+            'www.example.org',
+            'another.alias',
+            'example.com',
+            'migration.com',
+            'summer.com',
+            'geese.com',
+            'sslon.com',
+            'globalssl.com',
+            'globalsslsetssl.com',
+            'ipv6.com',
+            'ipv6ssl.com',
+            'headers.com'
+        },
+        names)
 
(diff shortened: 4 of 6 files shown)


bmw:
@ohemorange, it looks like you were talking to @ciarancourtney about this issue. Are you able to this PR?

ciarancourtney:
Hi @ohemorange just pushed 983653ffe84b0f9ab25dc47f8810651838e75acb, is that what you had in mind? If so, I think there was an implicit bug in some tests relating to the example.com case

ohemorange:
Changing it there would actually change how the configurator sees the server names in various different places. This leads to some different behavior like copying over server names incorrectly. It might be fine, but it's probably best to keep the changes as small as possible to fix this one particular thing, and we also already handle this case specifically in other places when we need to. I recommend modifying the domains in the `get_all_names` methods of `certbot-nginx/certbot_nginx/configurator.py`.

ciarancourtney:
Hi @ohemorange  I pushed a change in approach, does this seem better? One travis test is failing, doesn't make sense to me

```
            for server_name in vhost.names:
>               if util.is_wildcard_dot_domain(server_name):
E               AttributeError: 'module' object has no attribute 'is_wildcard_dot_domain'
```

ohemorange:
To fix that test, bump the version of certbot required by certbot-nginx to 0.29.0; an similar example of how to do that is here (no need to change `local-oldest-requirements.txt` or `setup.py`, just the first two files, but for `certbot`): https://github.com/certbot/certbot/pull/6250/files

ohemorange:
But actually, since the wildcard dot domain is Nginx-specific, there probably don't have to be any code changes in `certbot` at all; if you move the util function to `certbot-nginx` and do what you're doing in `_get_all_names` (which looks great!), there probably don't need to be any other changes in `certbot`.

ciarancourtney:
Hi @ohemorange just rebased but 2 builds still failing for some reason, can't figure out why, maybe you know?

ohemorange:
Certbot's latest version is no longer `0.29.0`; it should now be `0.31.0`, which you can find here: https://github.com/certbot/certbot/releases

That should account for at least the Apache test failures.

ohemorange:
Also, we squash commits, so it's actually a little cleaner to follow the history if you merge in master rather than rebasing.

certbot-nginx/local-oldest-requirements.txt
```diff
@@ -1,2 +1,2 @@
 acme[dev]==0.26.0
-certbot[dev]==0.22.0
+certbot[dev]==0.31.0
```
ohemorange:
```suggestion
-e .[dev]
```

certbot-nginx/setup.py
```diff
@@ -8,7 +8,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.26.0',
-    'certbot>=0.22.0',
+    'certbot>=0.31.0',
```
ohemorange:
```suggestion
    'certbot>=0.32.0.dev0',
```

certbot-nginx/local-oldest-requirements.txt
ciarancourtney: Hi @ohemorange Tried this but it still fails. You mentioned apache, should the changes be made there also or instead? Feel free to mod this PR, I have no clue of the testing deps of certbot.

ohemorange: Yes, you're right, Apache needs the change as well. I can't push to your branch, the change is the same as in `certbot-nginx/setup.py`.

ohemorange:
The Nginx compatibility failure looks like it's a real failure due to a change in behavior; you'll probably want to look at `_get_names` in `certbot-compatibility-test/configurators/nginx/common.py`.

ohemorange:
Hey @ciarancourtney, what is your current blocker? Could you use more clarity about the Nginx compatibility test failures?

ciarancourtney:
> Hey @ciarancourtney, what is your current blocker? Could you use more clarity about the Nginx compatibility test failures?

I tried that failing tox test locally and it passed, the travis result doesn't say what failed https://travis-ci.com/certbot/certbot/jobs/179620921

ohemorange:
It doesn't pass locally for me. The issue probably has to do with these configs:

```
nginx: [warn] conflicting server name "random1413.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "www.random1413.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "random25266.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "www.random25266.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "random28524.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "www.random28524.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "random27579.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "www.random27579.example.org" on 0.0.0.0:443, ignored
nginx: [warn] conflicting server name "random25863.example.org" on 0.0.0.0:443, ignored
nginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size
nginx: [warn] conflicting server name "random28524.example.org" on 0.0.0.0:80, ignored
nginx: [warn] conflicting server name "www.random28524.example.org" on 0.0.0.0:80, ignored
nginx: [warn] conflicting server name "" on 0.0.0.0:80, ignored
nginx: [warn] conflicting server name "www.random25863.example.org" on 0.0.0.0:80, ignored
nginx: [warn] conflicting server name "random25863.example.org" on 0.0.0.0:80, ignored
nginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size
```

ohemorange:
Hey @ciarancourtney, how's investigation on this going?

ciarancourtney:
Hi @ohemorange I rebased and it seems some failures related to certbox package versions were fixed upstream, by `nginx_compat` seems to have one failure I can repro locally with tox, but I can't see how you could identify where the failure is coming from.

See [here](http://www.mergely.com/xCFZEC6S/) for a diff between a good travis run (left) and this PR (right) and search `get_all_names test failed for config `

ohemorange:
Hey @ciarancourtney, looks like there's still some test failures going on here. Unfortunately we don't have the bandwidth to handle the debugging on our end, so I'm going to have to close the issue. If you do end up being able to resolve them please do let us know.

PR #6515, opened 2018-11-16
Update outdated tests

bmw:
These tests were running on Ubuntu Precise and Debian Wheezy which have reached their end of life and are no longer maintained by the respective distros. This updates the tests to a newer version of Debian and Ubuntu.

I also removed some unused `TRAVIS` environment variables. These were added in https://github.com/certbot/certbot/pull/6253 as a workaround for `pytest` but because we don't use `--numprocesses` in these tests or pass the `TRAVIS` environment variable through to the Docker image, they are unneeded.


bmw:
After this lands, `test-everything` will need to be updated to reflect the new test names.

adferrand:
@bmw, be warned that recent changes on master added the numprocesses option in pytest.ini, so it will be applied by default on any pytest execution.

bmw:
That's true! These tests run in Docker with the build environment being the `letsencrypt-auto-source` directory so the `pytest.ini` shouldn't affect these tests in their current form. Perhaps we should future proof it a bit though and always pass `TRAVIS` through just in case things change?

PR #6516, opened 2018-11-16
Whitelist docker-compose in tox.ini

bmw:
When the `docker_dev` test runs, it [prints a warning](https://travis-ci.org/certbot/certbot/jobs/452082456#L614) because it's running a command not found in the test's virtual environment.

This removes that warning by whitelisting `docker-compose`. `docker` does not need to be whitelisted because it is never called.


bmw:
Thanks for the review! While maybe a little silly for such a simple change, our workflow right now is that all changes need reviews.

Issue #6452, opened 2018-10-29
Use POST-as-GET for ACMEv2


bmw:
Since this change is [live for Let's Encrypt](https://community.letsencrypt.org/t/acme-v2-scheduled-deprecation-of-unauthenticated-resource-gets/74380), we should make this change as well.

It might be worth thinking about how to do this gracefully for people using Certbot with other ACME CAs.

PR #6522, opened 2018-11-18
Implement POST-as-GET requests

adferrand:
Relevant issue: #6452

This PR implements POST-as-GET requests for ACME v2 protocol.

With ACME v1, GET requests will be used directly. Furthermore with ACME v2, a GET request will be used if POST-as-GET failed, to handle gracefully ACME CA servers that do not implement yet the new ACME specification.

Also this PR adds scripts to execute the tests against Pepple, the integration ACME CA server of Let'sEncrypt for new features. Nothing is added to the CI yet, as the tests will fail because tls-sni-01 challenges are not accepted anymore by Pebble. But maybe in a future request, with adapted integration tests that would be also translated into Python.

cc @bmw @ohemorange


acme/acme/client.py
```diff
@@ -84,6 +84,32 @@ def _send_recv_regr(self, regr, body):
...
+        :param kwargs:
+        :return:
+        """
+        if self.acme_version >= 2:
+            # We add an empty payload for POST-as-GET requests
+            new_args = args[:1] + (None,) + args[1:]
+            try:
+                return self._post(*new_args, **kwargs)  # pylint: disable=star-args
+            except messages.Error as error:
+                if error.code == 'malformed':
```
ohemorange: `malformed` is for sure what you get once post-as-get has been implemented and you send something bad to that endpoint, but I can't find where in the spec it says what to give when you post to a get, pre-post-as-get. Do you have a pointer to that section?

acme/acme/client.py
```diff
@@ -84,6 +84,32 @@ def _send_recv_regr(self, regr, body):
             response, uri=regr.uri,
             terms_of_service=regr.terms_of_service)
 
+    def _post_as_get(self, *args, **kwargs):
```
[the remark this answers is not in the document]
adferrand: Done.

acme/acme/client.py
```diff
@@ -476,7 +502,7 @@ def _get_cert(self, uri):
 
         """
         content_type = DER_CONTENT_TYPE  # TODO: make it a param
-        response = self.net.get(uri, headers={'Accept': content_type},
+        response = self._post_as_get(uri, headers={'Accept': content_type},
```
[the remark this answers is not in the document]
adferrand: Done.

tests/pebble-fetch.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# Download and run Pebble instance for integration testing
+set -xe
+
+PEBBLE_VERSION=2018-11-02
+
+# We reuse the same GOPATH-style directory than for Boulder.
+# Pebble sto not need it, but it will make the installation consistent with Boulder's one.
```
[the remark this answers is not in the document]
adferrand: Corrected.

tests/certbot-pebble-integration.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
adferrand: Done.

tests/certbot-pebble-integration.sh
```diff
@@ -0,0 +1,13 @@
...
+    EXIT_STATUS=$?
+    unset SERVER
+    exit $EXIT_STATUS
+}
+
+trap cleanup_and_exit EXIT
+
+export SERVER=https://localhost:14000/dir
+
+./tests/certbot-boulder-integration.sh
```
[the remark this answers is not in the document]
adferrand: Corrected.

acme/acme/client.py
```diff
@@ -210,7 +236,7 @@ def poll(self, authzr):
         :rtype: (`.AuthorizationResource`, `requests.Response`)
```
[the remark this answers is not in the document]
adferrand: Done.

acme/acme/client.py
adferrand: There is no explicit declaration in the spec that `malformed` will be returned for requests that uses post-as-get on servers that do not support it.

However, `malformed` is referred to represent errors about invalid request messages in section 5.5. A request that uses an invalid verb would naturally fall in this category. And it is what I observe for Boulder v1 and Boulder v2 without post-as-get support.

So I think it is a reasonable condition to trigger a retry as GET, and ensure retro-compatibility.

acme/acme/client.py
```diff
@@ -236,6 +220,7 @@ def _revoke(self, cert, rsn, url):
             raise errors.ClientError(
                 'Successful revocation must return HTTP OK status')
 
+
```
ohemorange: nit: added whitespace

acme/acme/client.py
```diff
@@ -739,6 +756,32 @@ def revoke(self, cert, rsn):
...
+                return self._post(*new_args, **kwargs)  # pylint: disable=star-args
+            except messages.Error as error:
+                if error.code == 'malformed':
+                    logger.debug('Error during a POST-as-GET request, '
+                                 'your ACME CA may not support it:\n%s', error)
+                    logger.debug('Retrying request with GET.')
+                else:  # pragma: no cover
+                    raise
+
+        # If ACME v1 or POST-as-GET not supported, we use a GET instead.
```
ohemorange: no need to mention ACME v1 anymore

acme/acme/client.py
```diff
@@ -768,12 +811,7 @@ def __init__(self, net, key, server):
             self.client = ClientV2(directory, net=net)
 
     def __getattr__(self, name):
-        if name in vars(self.client):
-            return getattr(self.client, name)
-        elif name in dir(ClientBase):
-            return getattr(self.client, name)
-        else:
-            raise AttributeError()
+        return getattr(self.client, name)
```
ohemorange: Is this change because something was erroring? Note that it checks `vars` and `dir` on different classes -- it's meant to only call directly into shared methods, for our own checking.

tests/certbot-pebble-integration.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
ohemorange: (Copied from last review because it's easy to miss full-review comments)

Larger question: why is Pebble integration in this PR? Were you using it to test against by commenting out parts of certbot-boulder-integration? If we're going to include it, let's put something helpful there, like some test that does run properly.

acme/acme/client.py
adferrand: Well, the wrapper could not find the poll method when the delegate is ClientV2. And when I saw this method, I could not understand the advantage toward a proper delegate.

If I understand correctly, it makes necessary to implement the method on ClientBase, then override it on every child of ClientBase. And the purpose I suppose is to ensure that ClientBase always represents the common skeleton of the childs.

But it is not really Pythonic in my opinion, and not easy to understand. The result is a weird AttributeError that gives no clue of what is going on, while you would expect from a wrapper to just delegate to the underlying delagate object. 

A more Pythonic approach would be to implement on ClientBase the method as raising a NotImplementedError, that would lead straight forward the developper to the real error (we need to implement the method on all childs). Or using abc decorators.

I do not want to add a dummy poll method on ClientBase. The other solution would be to implement poll on ClientBase to use get, then override it on ClientV2 to use POST-as-GET. But it would be an infringement of what you said previously, as it would add a logic on ClientBase that is not relevant for ClientV2, without implementing the specific for Client.

So I think it is preferable to make a complete delegate like it did. Less code, on the right place, with Pythonic duck type approach.

acme/acme/client.py
adferrand: According to PEP8, top-level functions or classes should be separated by two blank lines. This convention is followed in the other definitions in this module, it make it in sync.

adferrand: Corrected.

acme/acme/client.py
adferrand: Corrected.

tests/certbot-pebble-integration.sh
ohemorange: Gotcha. Seems good either way.

acme/acme/client.py
ohemorange: The point of this was to make an explicit forcing function to make sure that all the differences between V1 and V2 were captured, so that our testing didn't automatically pass through and just test one of the versions of some method. Making it a complete delegate is saying that we think this code is stable enough that we're unlikely to make that sort of mistake in the future. That may be the case.

PR #6536, opened 2018-11-28
Conditionally depend on imgconverter for newer versions of Sphinx

### LINKED ISSUE #6343: Conditionally depend on sphinx imgconverter?
This is needed in Sphinx>=1.6, but Sphinx this new may not be available in older distros. Can we configure Sphinx to work with both versions by conditionally depending on imgconverter? Is there another approach we can take entirely?

The relevant PR here is https://github.com/certbot/certbot/pull/6070/files.

sydneyli:
Fixes #6343.


sydneyli:
@bmw, I've been having trouble getting `make -c docs html` to work with versions of Sphinx before 1.5, even before #6070 was landed. I think it's due to the version of `docutils` being pinned in `tools/dev_constraints.txt`, but I'll have to investigate some more. What versions of Sphinx are you seeing in older distros?

sydneyli:
> Let me know when this PR is ready for review or if it already is!

It's ready as-is, as I don't think it would cause any regressions for older versions of Sphinx.

sydneyli:
Actually, would you prefer that I lump in the fix for Sphinx < 1.5 in this PR, or create a new one for it? [This](https://github.com/sphinx-doc/sphinx/issues/3212) is one of the relevant bugs.

docs/conf.py
```diff
@@ -40,14 +42,16 @@
...
     'sphinx.ext.autodoc',
-    'sphinx.ext.imgconverter',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
     'sphinx.ext.viewcode',
     'repoze.sphinx.autointerface',
 ]
 
+if pkg_resources.get_distribution('sphinx').version >= '1.6':
```
bmw: I wasn't aware of this before, but I think we may be able to use [sphinx.version_info](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.version_info) here.

sydneyli:
Ok, so [annotations](https://het.as.utexas.edu/HET/Software/Sphinx/ext/autodoc.html), which are used [here](https://github.com/certbot/certbot/blob/master/docs/api/constants.rst) aren't supported until Sphinx 1.2, so that's the oldest version of Sphinx we can depend on.

sydneyli:
Ready for re-review.

setup.py
```diff
@@ -68,9 +68,11 @@ def read_file(filename, encoding='utf8'):
 ]
 
 docs_extras = [
+    # Bug in Sphinx < 1.5.1
+    # https://github.com/sphinx-doc/sphinx/issues/3212
+    'docutils<0.13.1',
```
bmw: I don't think we should require an older version of docutils here. Newer versions of Sphinx work just fine with newer versions of docutils and this will cause problems on up to date distros that only have newer versions of both packages available. Really what we want to say is if you are using an older version of Sphinx then we require an older version of docutils, but I don't think there's a good way to do that.

In general, this is a bit of an ugly problem that we've hit before with other dependencies. In those cases, we've done nothing and left it up to the user and/or their package manager/maintainers to ensure compatible versions of our dependencies are installed. This certainly isn't elegant, but it hasn't caused any issues. Distro package maintainers (usually) test packages before pushing them live and we wouldn't be the only Python package that would have issues here. For people installing from `pip`, `pip` prefers the latest version of everything which usually has all the latest bug fixes.

It's probably a good idea to mention in a comment here that `docutils<0.13.1` is required when used with `Sphinx<1.5.1` and provide a link to the GitHub issue, but I think the actual dependency should be on an unpinned back version of docutils.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ https://github.com/certbot/certbot/milestone/62?closed=1
 * The grammar used by Augeas parser in Apache plugin was updated to fix various parsing errors.
 * The CloudXNS, DNSimple, DNS Made Easy, Gehirn, Linode, LuaDNS, NS1, OVH, and
   Sakura Cloud DNS plugins are now compatible with Lexicon 3.0+.
+* Only depend on imgconverter extension for Sphinx >= 1.6
```
bmw: It looks like we had a bad merge here. This should be added to the 0.29.0 section above.

CHANGELOG.md
```diff
@@ -79,6 +80,7 @@ package with changes other than its version number was:
 
 More details about these changes can be found on our GitHub repo:
 https://github.com/certbot/certbot/milestone/59?closed=1
+>>>>>>> master
```
bmw: And this should be deleted.

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ Despite us having broken lockstep, we are continuing to release new versions of
 all Certbot components during releases for the time being, however, the only
 package with changes other than its version number was:
 
-*
+* Only depend on imgconverter extension for Sphinx >= 1.6
```
bmw: Whoops. We should move this up to the fixed section. This is where we list the packages that changed in the release.

sydneyli: oops. should be fixed now, I hope >.<

Issue #6466, opened 2018-11-02
HTTPS Redirect gets added to existing HTTPS redirect.


fake-name:
Basically, I'm running `nginx` with a already existing http -> https redirect:

```
server
{
	listen 80;
	server_name my_domain.com www.my_domain.com;
	return 301 https://www.my_domain.com$request_uri; # enforce https
}
```

Running `certbot-auto` apparently does not recognize this, and then offers to generate a pointless addition to the nginx config:

```
server
{
    if ($host = www.my_domain.com) {
        return 301 https://$host$request_uri;
    } # managed by Certbot


	listen 80;
	server_name my_domain.com www.my_domain.com;
	return 301 https://www.my_domain.com$request_uri; # enforce https


}
```

It seems that if the server config entry has a `return 301 https://<anything>` entry in it's root, certbot shouldn't even offer to add a https redirect of it's own, since it's already done?

It also appears that `certbot-auto` offers to add the redirect even if it's already present, but it apparently does remove the existing redirect if present.

This is definitely not a major issue.

sydneyli:
Thanks for the report @fake-name! Yeah, we don't try to detect whether a redirect is already set before prompting the user about it. Wishlisting for now since we still do ask the user about it beforehands. If we do decide to redirect by default then I'd bump up the priority on this. If anyone's interested in implementing a fix for this in the meantime, we can give you some guidance!

jonastheis:
I'd like to help out on this one. Can you please give me some guidance?

fake-name:
As far as I know, this issue is as present now as it was when I reported it.

RasPat1:
Related PR: #6539

eddiewebb:
this IS a major issue when:
- using multiple domains with ALIAS
- trying to go direct from http-alias --> https-primary


Even with `--no-redirect` the installer will add the pointless blocks.

WHY that is a major issue for me is it violates google webmaster migration guidelines as the intermediate redirect does not match the expected (primary) domain.


### Before `certbot install -n --nginx --no-redirect -d mydomain -d aliasdomain`


```
server {
    listen 443;
    root /var/www/domain;
    index index.php index.html index.htm;
    server_name domain www.domain;
  ...
}
server {
    server_name domain;
    rewrite ^/(.*)$ https://domain/$1 permanent;
    listen 80;
}


#
# Aliases are a redirect, straight to https://domain
#  

  server {
    server_name aliasdomain  ;
[...]

PR #6539, opened 2018-11-30
https://github.com/certbot/certbot/issues/6466

b3n4kh:
Test for existing 301 redirect in vhost as described in https://github.com/certbot/certbot/issues/6466.

Fixed two comments since it isnt any rewrite directive (anymore?) but a return 301


bmw:
On IRC @b3n4kh asked:
> The test would be pretty much the same as the already exsiting test_split_for_redirect() from configurator_test.py but I would either have to "add_server_directives" to the generated http block or write my own "sites-enabled" config
> Or is an test for an edge case like this nonsense? Aany feedback would be really appreciated

I suggested they open a PR and let the reviewer decide if a test is needed.

@ohemorange or @sydneyli, are you able to review this small nginx PR?

ohemorange:
Thanks for the PR @b3n4kh! My personal opinion is that it's not worth it since it doesn't break existing configurations, but maybe @sydneyli thinks otherwise.

certbot-nginx/certbot_nginx/_internal/configurator.py
```diff
@@ -719,7 +719,11 @@ def enhance(self, domain, enhancement, options=None):
 
     def _has_certbot_redirect(self, vhost, domain):
         test_redirect_block = _test_block_from_block(_redirect_block_for_domain(domain))
-        return vhost.contains_list(test_redirect_block)
+        redirect_block_for_server = [['\n    ', 'return', '301', 'https://$host$request_uri']]
```
sydneyli: This seems specific to the config in #6466. We might want something more generic that can detect the same directive with any tabbing.

And since this is a regular list, I think you need the whitespace information between tokens. Might be able to reuse some existing code in `_redirect_block_for_domain(domain)` here!

b3n4kh: The tabbing is actually obselete now since I use `nginxparser.UnspacedList`.

I also added a unit Test if `_has_certbot_redirect` returns True if a custom block exists.

Issue #6542, opened 2018-11-30
acme module: bad nonce error request retry logic regression


cpu:
:wave: Hi folks,

I think https://github.com/certbot/certbot/commit/0b5468e992ab57fa028ddf33ca2351cb37c8e1ee from https://github.com/certbot/certbot/pull/6522 introduced a regression that causes the retry logic for badNonce errors in the `acme` module to fail.

Pebble uses Certbot master for its acme module in CI:
https://github.com/letsencrypt/pebble/blob/dadfdf1571ab506718a5ac7face55778c3e2ee31/.travis.yml#L15-L20

It uses our small ACME test client ([`chisel2.py`](https://github.com/letsencrypt/pebble/blob/dadfdf1571ab506718a5ac7face55778c3e2ee31/test/chisel2.py)) to test issuance against a `pebble` instance that by default rejects 15% of all valid nonces as invalid so that client retries are exercised. 

Prior to 0b5468e992ab57fa028ddf33ca2351cb37c8e1ee Certbot and the `acme` module handled this as-expected and would retry requests that failed from a badNonce error. After this commit the badNonce errors seem to consistently cause an overall failure. Here's an [example build](https://travis-ci.org/letsencrypt/pebble/builds/461873896) of Pebble `master` with a log.

To confirm this was a regression I pushed 10 builds of Pebble master using Certbot master (named `test-certbot-regression-1` .. `test-certbot-regression-10`) and 10 builds of Pebble master using Certbot pinned [one commit before tip of master](https://github.com/letsencrypt/pebble/commit/63090c585740f66757c4cdb938a40350d454f7e7) at 8b5ac9e2573de9dfd4182bdcd98d7262a2bdc8f0 (named `test-certbot-regression-bisect-1` .. `test-certbot-regression-bisect-10`). The results are visible [here](https://travis-ci.org/letsencrypt/pebble/branches). All 10/10 of the builds using `acme` from Certbot master failed and all 10/10 of the builds pinned a commit before passed.

bmw:
@adferrand, @ohemorange: Are you able to take a look at this?

We can always revert #6522 for now until we've been able to fix any problems here.

ohemorange:
Let's revert now so tests pass

adferrand:
So I put here what has been discussed on IRC. In fact this is an expected behavior. Certbot will try only once the request on a badNonce error.

[...]

PR #6543, opened 2018-11-30
Revert "Implement POST-as-GET requests (#6522)"

ohemorange:
This reverts commit 0b5468e992ab57fa028ddf33ca2351cb37c8e1ee.

See #6542.


adferrand:
This PR can be canceled, the fallback is not needed  as discussed in #6542

PR #6541, opened 2018-11-30
[Unix] (WIP) Create a framework for certbot integration tests

adferrand:
This PR proposes an industrialization of Certbot integration tests.

On top of the unit tests, there are several integration tests that test certbot and certbot-nginx plugin against an ad-hoc instance of Boulder, the software that run the ACME CA server of LetsEncrypt.

Theses scripts are written in bash, and are located respectively on `tests/certbot-integration-tests.sh` and `certbot-nginx/tests`. While theses tests do their job, I see several flaws in their construction.
* Bash is a scripting language, not a programming language like Python. If you can make already several abstractions with Bash, it is nothing in comparison with what can be done with Python, as of structured data, collection comprehension, context manager, generators, or whatever you can imagine. In fact, using Python lead to more robust codes, and better maintenance capabilities for tests. In particular when tests concern a project written also in Python. So the technical knowledge is settled.
* A really good effort is done to provide some typical test material, like pseudo-fixtures with `tests/integration/common.sh`. But again, this is Bash, and there is not match against the galaxy of features offered by the `pytest` project. Setup/teardown, fixtures, module instrospection, hook and so on. Not only to talk about the excellent pytest plugins available and used by the community, test reruns or code coverage. And like for Python, every unit test is written with unittest or pytest, so here again, the technical knowledge is here.
* Current integration tests are not run in parallel, and we know that integration tests are slow: more than 18 mins in fact, on TravisCI. Adding more tests would lead to an unacceptable time to get feedback from the CI.
* The association of theses tests with TravisCI in fact, is poor: a lot of specific code is required, out of `tox`, and the logs generated are, honestly, unusable. As everything is logged, no matter a test succeeded or failed, you have typically more than 23000 lines of logs. If a test fail in this, you will have a hard time to find what is going on.

In general, I believe more in the integration tests than in the unit tests to provide valuable insight for a CI. So I want to make them faster, more maintainable, better designed in general, and make them as easy to execute than the unit tests.

So this is PR is a complete rewrite of theses tests, using `pytest`, and all python techniques I know to ease the test env setup and the tests execution. Here are the general features:
* tests are now written in Python with `pytest`, giving all the good values of this framework: fixtures, assertion insights, modularization and so on
* migration to execute the integration tests on Windows is now possible (future PR) 
* thanks to all the `pytest` power, each test is know executed independently from each other: fixtures will handle the setup/teardown of all the context required to run Certbot
[...]

PR #6544, opened 2018-11-30
Preserve other-read bit on private keys too

sydneyli:

Additional fix for #1473 to prevent breakages.

Not updating the guidelines in `using.rst`-- I don't want to encourage people to use this. now that Certbot preserves gid, the better way to set permissions is to change the group permisisons!

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -1144,7 +1144,8 @@ class RenewableCert(object):
             # If the previous privkey in this lineage has an existing gid or group mode > 0,
             # let's keep those.
             group_mode = stat.S_IMODE(os.stat(old_privkey).st_mode) & \
-                (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP)
+                (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | \
+                 stat.S_IROTH)
             mode = BASE_PRIVKEY_MODE | group_mode
             os.chown(target["privkey"], -1, os.stat(old_privkey).st_gid)
             os.chmod(target["privkey"], mode)
--- a/certbot/tests/storage_test.py
+++ b/certbot/tests/storage_test.py
@@ -565,12 +565,12 @@ class RenewableCertTests(BaseRenewableCertTest):
         # If new key, permissions should be rest to 600 + preserved group
         self.test_rc.save_successor(2, b"newcert", b"new_privkey", b"new chain", self.config)
         self.assertTrue(compat.compare_file_modes(
-            os.stat(self.test_rc.version("privkey", 3)).st_mode, 0o640))
+            os.stat(self.test_rc.version("privkey", 3)).st_mode, 0o644))
         # If permissions reverted, next renewal will also revert permissions of new key
         os.chmod(self.test_rc.version("privkey", 3), 0o404)
         self.test_rc.save_successor(3, b"newcert", b"new_privkey", b"new chain", self.config)
         self.assertTrue(compat.compare_file_modes(
-            os.stat(self.test_rc.version("privkey", 4)).st_mode, 0o600))
+            os.stat(self.test_rc.version("privkey", 4)).st_mode, 0o604))
 
     @test_util.broken_on_windows
     @mock.patch("certbot.storage.relevant_values")
--- a/tests/certbot-boulder-integration.sh
+++ b/tests/certbot-boulder-integration.sh
@@ -281,7 +281,7 @@ CheckCertCount() {
 }
 
 CheckGroup() {
-    group_mode() { echo $((0`stat -c %a $1` & 070)); }
+    group_mode() { echo $((0`stat -c %a $1` & 074)); }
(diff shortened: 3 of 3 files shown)


tests/certbot-boulder-integration.sh
```diff
@@ -281,7 +281,7 @@ CheckCertCount() {
 }
 
 CheckGroup() {
-    group_mode() { echo $((0`stat -c %a $1` & 070)); }
+    group_mode() { echo $((0`stat -c %a $1` & 074)); }
```
adferrand: Name of the method and error message should change, as it check know both permission on group and read permission on world.

certbot/storage.py
```diff
@@ -1144,7 +1144,8 @@ def save_successor(self, prior_version, new_cert,
             # If the previous privkey in this lineage has an existing gid or group mode > 0,
             # let's keep those.
             group_mode = stat.S_IMODE(os.stat(old_privkey).st_mode) & \
```
bmw: nit: I think we should change the name of this variable and the comment above since it's no longer just the group mode.

certbot/tests/storage_test.py
```diff
@@ -565,12 +565,12 @@ def test_save_successor_maintains_group_mode(self, mock_rv):
         # If new key, permissions should be rest to 600 + preserved group
```
bmw: nit: Can we update this comment?

certbot/tests/storage_test.py
```diff
@@ -565,12 +565,12 @@ def test_save_successor_maintains_group_mode(self, mock_rv):
...
         self.test_rc.save_successor(2, b"newcert", b"new_privkey", b"new chain", self.config)
         self.assertTrue(compat.compare_file_modes(
-            os.stat(self.test_rc.version("privkey", 3)).st_mode, 0o640))
+            os.stat(self.test_rc.version("privkey", 3)).st_mode, 0o644))
         # If permissions reverted, next renewal will also revert permissions of new key
         os.chmod(self.test_rc.version("privkey", 3), 0o404)
         self.test_rc.save_successor(3, b"newcert", b"new_privkey", b"new chain", self.config)
         self.assertTrue(compat.compare_file_modes(
-            os.stat(self.test_rc.version("privkey", 4)).st_mode, 0o600))
+            os.stat(self.test_rc.version("privkey", 4)).st_mode, 0o604))
```
bmw: Rather than changing this to `0o604`, I'd personally prefer we chmod the key to `0o400` and see the other permission is properly removed. This also matches the comment for these lines of the test.

tests/certbot-boulder-integration.sh
```diff
@@ -280,23 +280,32 @@ CheckCertCount() {
     fi
 }
 
-CheckGroup() {
-    group_mode() { echo $((0`stat -c %a $1` & 070)); }
-    group_owner() { echo `stat -c %G $1`; }
-    if [ `group_mode $1` -ne `group_mode $2` ] ; then
-        echo "Expected group permission `group_mode $1`, got `group_mode $2` on file $2"
+CheckPermissions() {
```
bmw: I'm not sure this function is working as we expect. See https://travis-ci.org/certbot/certbot/jobs/462933676#L4862. It's complaining that `$3` is unbound which I think is happening because `$3` is getting reset in the scope of the `masked_mode` function.

tests/certbot-boulder-integration.sh
```diff
@@ -280,23 +280,32 @@ CheckCertCount() {
...
+    masked_mode() { echo $((0`stat -c %a $1` & 0`$3`)); }
+    if [ `masked_mode $1` -ne `masked_mode $2` ] ; then
+        echo "With $3 mask, expected mode `masked_mode $1`, got `masked_mode $2` on file $2"
         exit 1
     fi
+}
+
+CheckGID() {
+# Args: <filepath_1> <filepath_2>
+# Checks group owner of two files match
```
bmw: Thanks for adding comments like this!

tests/certbot-boulder-integration.sh
bmw: Now it's complaining that it can't find the command `074` and `masked_mode` is always returning 0. The fix might be as simple as changing the backticks to quotes, but I didn't play around with it.

sydneyli: I'll play around with it. Thanks for your patience on this-- i'm pretty green when it comes to shell scripting.

bmw: No worries! I learned bash by working on Certbot.

I just quickly made a note of the problem because I was trying to get a few things done, but if you're spending much time trying to fix and would like some help, please let me know!

adferrand: Do not push too hard on bash tests, they are rewritten in Python with #6541. At worst, you could just describe what you want as a comment, and I will write it in Python once your PR is merged in master and that I sync my PR with master.

FelixSchwarz:
Btw: Thank you for implementing this more complicated approach. Now I can safely upgrade my certbot installs :-)

PR #6547, opened 2018-11-30
During the release, reset local-oldest-requirements.txts that might have been changed during development.

ohemorange:


tools/_release.sh
```diff
@@ -264,6 +264,14 @@ if [ "$RELEASE_BRANCH" = candidate-"$version" ] ; then
     SetVersion "$nextversion".dev0
     letsencrypt-auto-source/build.py
     git add letsencrypt-auto-source/letsencrypt-auto
+    for pkg_dir in $SUBPKGS_NO_CERTBOT certbot-compatibility-test .
+    do
+      sed -i 's/-e acme[dev]/acme[dev]==$version/' "$pkg_dir/local-oldest-requirements.txt"
```
bmw: Can we check for the existence of this file first? It doesn't exist for each package and the absence of the file causes the release script to fail.

bmw:
> By the way, while not caused by this PR, while testing this I also had issues due to pytest.ini. Some of the command line options aren't recognized because not all pytest plugins we normally use are installed and some newer versions of our dependencies are throwing deprecation warnings.

> We need to fix this before Wednesday. Are you able to take take this or should I?

I can do this piece.

tools/_release.sh
ohemorange: why doesn't it exist for each package

bmw: `acme` for example doesn't depend on any local packages in the repo.

ohemorange: is it just acme that's excluded?

bmw: I think just that and `certbot-compatibility-test`. We don't release `certbot-compatibility-test` so trying to make sure it works with new and old versions of our packages seems unnecessary to me as it's just a developer tool.

tools/_release.sh
```diff
@@ -264,6 +264,16 @@ if [ "$RELEASE_BRANCH" = candidate-"$version" ] ; then
     SetVersion "$nextversion".dev0
     letsencrypt-auto-source/build.py
     git add letsencrypt-auto-source/letsencrypt-auto
+    for pkg_dir in $SUBPKGS_NO_CERTBOT .
+    do
+      if [ -f "$pkg_dir/local-oldest-requirements.txt" ]; then
+        sed -i 's/-e acme[dev]/acme[dev]==$version/' "$pkg_dir/local-oldest-requirements.txt"
+        sed -i 's/-e acme/acme[dev]==$version/' "$pkg_dir/local-oldest-requirements.txt"
+        sed -i 's/-e .[dev]/certbot[dev]==$version/' "$pkg_dir/local-oldest-requirements.txt"
+        sed -i 's/-e ./certbot[dev]==$version/' "$pkg_dir/local-oldest-requirements.txt"
```
bmw: I just tested this and I think you want to use double quotes here as the literal string `$version` was added to the file.
```
eff-bmw:tmp.Qvtu1bDpnM bmw$ version=0.29.0
eff-bmw:tmp.Qvtu1bDpnM bmw$ echo '$version'
$version
eff-bmw:tmp.Qvtu1bDpnM bmw$ echo "$version"
0.29.0
eff-bmw:tmp.Qvtu1bDpnM bmw$
```
Aren't shell scripts fun?

bmw: The good news is Adrien has taken the task of rewriting everything in Python. The bad news is he wasn't planning on doing this script because it's never likely to be run on Windows. We could always do it anyway in the future though.

adferrand: The good news, it that I will do it eventually ^^

adferrand: Oh also on sed commands, dots and brackets needs to be escaped. They means something in the context of regular expressions.

bmw: Adrien, I'm impressed. I didn't even mention you with @ and you still responded within 2 minutes at 11:35PM UTC :smile:

adferrand: And luckily, the interpolation of $version comes in the replace string part, not in the pattern to match part, otherwise, the dots in the variable would also be interpretated as regular expressions: here, is the real fun with shell.

adferrand:
Really sorry... It is not enough, you have space in the command, it will be interpretated by the shell has a separator between arguments, and escape needs to happen only in the pattern part, not the replacement part.

Here:
```
sed -i "s/-e acme\[dev\]/acme[dev]==$version/" "$pkg_dir/local-oldest-requirements.txt"
```

adferrand:
OK, sed looks good to me now ^^

tools/_release.sh
adferrand: Well, @bmw, lucky you, I was checking my PRs on Certbot. When you have a full suite of them, it happens quite often :p

PR #6553, opened 2018-12-04
[Windows|Unix] Use double quote compatible both for windows and linux

adferrand:
File `_venv_common.py` uses single quotes to ask `pip` to install `setuptools>=30.3`. Using single quotes to enclose a string is not supported on Windows shell (Batch). This PR replaces theses single quotes by double quotes supported both on Windows and Linux.


adferrand:
cc @ohemorange

PR #6557, opened 2018-12-05
List packages in changelog.

bmw:


bmw:
I'm just going to make these changes locally before the release. We never used to review changes to the changelog before we sent it out anyway.

PR #6559, opened 2018-12-05
Correct Unix paths for Certbot

adferrand:
PR #6416 introduced a regression on release 0.29.0, by modifying paths for workdir and log on Unix from `/var/lib/letsencrypt` and `/var/log/letsencrypt` to `/var/letsencrypt/lib` and `/var/letsencrypt/log` respectively.

This PR fixes that by reverting to the original paths.

Really sorry for the inconvenience.


bmw:
Thanks for the PR @adferrand and no worries about the trouble.

I think I might take #6560 instead of this though which preps the changelog for the 0.29.1 release and adds a silly test.

adferrand:
Superseeded by #6560

PR #6560, opened 2018-12-05
Fix default directories on Linux

bmw:
This PR is built on #6558.


bmw:
While silly, I took the changes to `cli_test.py` and ran them against Certbot 0.28.0 (before the change causing problems here) and the tests still passed.

bmw:
I thought merging `master` into this branch would automatically fix up the diff but it didn't. Just rebased on `master`.

PR #6565, opened 2018-12-06
Speed up AppVeyor execution

adferrand:
CI workflow on AppVeyor is slow: there is only one executor at a time, with a heavy warm up. As a consequence, any build currently will take between 20 to 25 mins with four jobs, not to mention time to wait for the executor to be available.

This PR reduce the number of jobs to two, by executing only against the lowest and highest version of Python supported by Certbot on Windows, and by executing everything in Windows Server 2012 R2: if it works on this platform, we can safely rely on the Python stability for the newest Windows versions, to assert that Certbot will work also for them.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,17 +1,9 @@
+image: Visual Studio 2015
+
 environment:
   matrix:
-    - FYI: Python 3.4 on Windows Server 2012 R2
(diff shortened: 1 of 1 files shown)


adferrand:
cc @bmw

PR #6568, opened 2018-12-06
Remove -q/--quiet from pip invocations

bmw:
While reducing noise in test output is valuable, this flag has made a couple aspects of Certbot's development difficult:

1. We test with different sets of dependencies and running pip in quiet mode removes all output about the packages being installed which has made reviewing changes to these tests more difficult.
2. When pip fails, it provides significantly less output about the failure in quiet mode than it does normally. The output is reduced so much that in the two times I've hit this issue in the last month, I was only able to see that installing package X failed rather than what the cause of that failure was which could be seen with `--quiet` removed.

Also, since running pip without `--quiet` is the tox default, I expect Python developers to be familiar with what they see here.

@jsha, this reverts the change you made in 7319cc975a and if you really care, we could reconsider, but landing this PR solves issues that some of the Certbot team and I have been having.

### DIFF
--- a/tools/install_and_test.py
+++ b/tools/install_and_test.py
@@ -23,7 +23,7 @@ def call_with_print(command, cwd=None):
 
 def main(args):
     if os.environ.get('CERTBOT_NO_PIN') == '1':
-        command = [sys.executable, '-m', 'pip', '-q', '-e']
+        command = [sys.executable, '-m', 'pip', '-e']
     else:
         script_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


adferrand:
Also, we can still get concise logs on Travis, thanks to the folded logs option. It is quite hidden, but we can see the way to do it here: https://github.com/travis-ci/travis-ci/issues/2285

Logs stay folded except if the global exit coded of the folded part is not 0.

jsha:
If the `-q` flag is actively causing you problems, by all means roll back. I have no objections.

And thanks for the link about folded logs, @adferrand ! I didn't know about that and it looks quite useful.

PR #6571, opened 2018-12-07
Update pinned version of pyOpenSSL

bmw:
We got a notification of a security vulnerability from GitHub a while ago telling us to update our pyOpenSSL version here. I looked into it at the time and found it didn't affect us, but made a note to myself to update the version of pyOpenSSL in the future anyway. This PR does that.

I verified:

* There are no backwards incompatible features we care about listed at https://github.com/pyca/pyopenssl/blob/master/CHANGELOG.rst.
* pyOpenSSL's dependencies don't also need to be updated according to https://github.com/pyca/pyopenssl/blob/18.0.0/setup.py#L95.
* The hashes match when obtained from different network vantage points.


bmw:
Closing and reopening to run on travis-ci.com.

PR #6572, opened 2018-12-07
Remove IValidator

bmw:
I came across `IValidator` when explaining the code base to someone and it's completely unnecessary. The validator code in `certbot-compatibility-test` was in the `certbot` package once upon a time, but that was long ago so the otherwise unused `zope.interface` definition seems entirely unnecessary.


bmw:
Closing and reopening to run on travis-ci.com.

PR #6578, opened 2018-12-10
[Unix] Create a framework for certbot integration tests: PART 1

adferrand:
Following discussions in #6541, I splitted my PR into several parts to make it reviewable. 

This is part 1, and concerns the paralleled setup of several acme server instances, in order to drastically speed up the integration tests. This is done by orchestrating the pytest execution with an ad-hoc cli, that will make the work before passing the distributed configuration of acme servers to pytest and launch its execution.

The code is functional, in a sense where all the setup is done, and only dummy tests without any fixtures are executed. This will come in the future parts. 

Integrations tests can be executed using the tox target env `integration`. Customizations can be done using directly the cli `launch-integration-tests.py`: it uses argparse to process arguments, and so contains a complete online help. 

cc @ohemorange @bmw


ohemorange:
Thanks for breaking it up! If you were to break this code into two equal parts, which files/pieces would go in which part? I'd like to review in chunks of no more than 400 lines.

adferrand:
No equal part, but one part will have 500 lines, but with a lot of documentation lines. Conceptually you can review `acme.py` + `misc.py` as part B, and everything else as part A. Part A is about all the general configuration of the project and the CLI, while part B is about the acme server setup itself. Part A calls part B from a unique entrypoint, the contextmanager `acme.setup_acme_servers`.

ohemorange:
Thanks, I'll take a look in a bit!

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+from _pytest import config as config_pytest
+
+
+def pytest_configure(config):
+    # type: (config_pytest.Config) -> None
+    """
+    Standard pytest hook used to add a configuration logic for each node of a pytest run.
+    :param pytest_config.Config config: the current pytest configuration
+    """
+    if not os.environ.get('CERTBOT_ACME_TYPE'):
```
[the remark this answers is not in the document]
adferrand: Nope it is not the same. Here we are testing the existence of `CERTBOT_ACME_TYPE`, not `CERTBOT_ACME_XDIST`. Environment variable `CERTBOT_ACME_TYPE` is not used in part 1, but it will be useful in part 3 to skip seamlessly some incompatible tests with current acme server type.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -0,0 +1,25 @@
+import requests
+import urllib3
+
+
+def test_hello_1(request, worker_id):
```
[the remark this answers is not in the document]
adferrand: Fixture `request` is created automatically by `pytest`, it describes the current test environment. Fixture `worker_id` is created automatically by `pytest-xdist` module, it gives the worker id that execute the current test.

certbot-ci/certbot_integration_tests/nginx_tests/test_main.py
```diff
@@ -0,0 +1,2 @@
+def test_hello(request):
```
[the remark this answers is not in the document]
adferrand: See my response about `request` fixture above.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -0,0 +1,25 @@
...
+    try:
+        response = requests.get(request.config.acme_xdist[worker_id]['directory_url'], verify=False)
+        response.raise_for_status()
+        assert response.json()
+        response.close()
+    except urllib3.exceptions.InsecureRequestWarning:
+        pass
+
+
+def test_hello_2(request, worker_id):
```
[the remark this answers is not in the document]
adferrand: See my response about `request` and `worker_id` fixture above.

certbot-ci/launch-integration-tests.py
```diff
@@ -0,0 +1,232 @@
...
+    # By calling certbot_ci_workspace, the tempfile.tempdir value of standard tempfile module is
+    # modified for the context execution time, to ensure that any temporary assets will be created
+    # under the cerbot-ci workspace.
+    with certbot_ci_workspace():
+        acme_config = {}
+        # Prepare the acme config server. Data is specific to an acme type. Module
+        # utils.acme_server will handle theses specifics.
+        if 'pebble' in args.acme_server:
+            acme_config['type'] = 'pebble'
+            acme_config['option'] = 'nonstrict' if 'nonstrict' in args.acme_server else 'strict'
```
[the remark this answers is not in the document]
adferrand: It is two different runtime modes for Pebble. Because Pebble is quite on the edge of ACME protocol specifications, it can be difficult to use it as an integration tool to build software that will use real ACME CA servers. Because theses servers have options that are removed from Pebble in `strict` mode (deprecated options, security problem and so on.).

So the `nonstrict` mode is here to allow some reasonnable deprecated options (things that are not deprecated because of a security flaw) to be used in Pebble. It should be the standard mode for Certbot, as it mimic almost all Boulder ACME v2 capabilities, apart obviously TLS-SNI-01 challenges.

certbot-ci/launch-integration-tests.py
```diff
@@ -0,0 +1,232 @@
...
+        else:
+            acme_config['type'] = 'boulder'
+            acme_config['option'] = 'v1' if 'v1' in args.acme_server else 'v2'
+        # By calling setup_acme_server we ensure that all necessary acme servers instances will be
+        # fully started once the context is entered. This runtime is reflected by the acme_xdist
+        # provided by the contextmanager.
+        with acme.setup_acme_server(acme_config, workers, repositories_path) as acme_xdist:
+            os.environ['CERTBOT_ACME_TYPE'] = args.acme_server
```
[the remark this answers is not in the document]
adferrand: We are passing command line arguments understood by pytest. We need something to pass other contexts. The most easiest way is to set en environment variable. And, you can also use a hook on pytest to extend its capabilities towards command line arguments, and add a new one.

It would work well for `CERTBOT_ACME_XDIST`. However it will be a problem for `CERTBOT_ACME_TYPE`, that will be used in part 3 in a area that have no direct access to the pytest context execution.

Instead of splitting in two ways the context I want to pass to pytest, I use only one, and choose what is working for both contexts.

certbot-ci/launch-integration-tests.py
```diff
@@ -0,0 +1,232 @@
...
+    call to the Python tempdir functions during the execution of this context.
+    :return: The new base tempdir
+    :rtype: str
+    """
+    original_tempdir = tempfile.tempdir
+    try:
+        tempfile.tempdir = os.path.join(tempfile.gettempdir(), 'certbot_ci_tmp')
+        try:
+            os.mkdir(tempfile.tempdir)
+        except OSError as error:
```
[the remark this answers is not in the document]
adferrand: It is not the workspace of the tests, but the workspace of `certbot-ci` itself. It is used before parallelization to gather heavy required assets, like the ACME CA server GIT repositories, and in the future, pre-compiled binaries. Things that we would like to keep across multiple invocations of test integrations.

But workspaces for tests are fully independent.

adferrand:
Now the high level questions. About your first remark.

It is not possible, or extremely inefficient, to have separated launcher for each integration tests campaign. First because from the code point of view, several blocks would be duplicated. Second because it would not use the global and managed setup of all acme server distribution for all tests that are meant to be run. Boulder is a 1 min setup. You do not repeat that several times. And you do not dispatch efficiently the tests across nodes.

In theory, you could share the logic by doing things like, in a tox env:
- setup acme servers
- call a certbot core tests launcher
- call a nginx certbot tests launcher
- teardown acme servers

Well, it is defying all the integration test logic purpose I want to build here. Basically, tox is just a python venv builder + a shell script. It is not a test framework (it shines however to deploy contained test context however).

With a shell, you do not have automated teardown with Exception handling for instance.

What I want here, is a self-contained framework that does everything on its own to setup the integration tests and run them. From this point of view, `tox` is just a fast way to run a predefined set of this integration tests framework for a daily development workflow, or in CIs.

That leads me to the second point, the interest of having a CLI. Well, in my opinion, any first level Python script should be constructed as a CLI. Because setting a CLI with ArgParse is easy, and offer to user a online help, control of input, feedback on input errors. Here, it takes 26 lines of code + 1 line to call the parser, I think it is not overweight. As I said before, the purpose of `launch-integration-tests.py` is not to leverage the work of `tox`, but to superseed the pytest CLI to launch integration tests. So it is a test framework on its own, and should be constructed accordingly.

Also, something that cannot be seen from current PR and their future parts: I am planning to migrate every not unit test (letstest, compatibility-tests, certbot-apache tests, future dns plugin integration tests) into this framework.

adferrand:
For the second part of your general comment, I am not sure to understand it correctly.

If it is about: "Why not staying with bash scripts called from tox ?", then I hope that my explanation above, the description in #6541, and the code you review here, will convince you about the advantages that Python gives over Bash in term of integration tests.

I will just also add a final note: eventually theses integration tests will need to be run on Windows, to allow a secured official support of Certbot on Windows. So here also the Python language to allow portability.

ohemorange:
So, it's becoming clear to me that this integration test change is part of a larger project to redesign how we run non-unit tests.

I'd love to see a design document for the project, exploring the different stages that it could take, weighing the options for implementation. Particularly, I'm interested in why no existing test-running framework could work for us here. Between the load management and the docker instances, this really doesn't seem like the sort of thing that belong to a program that gets people certs, so I want to make sure that it really does make sense for us to be maintaining this code as part of Certbot.

In general, design docs are a good approach for anything other than small changes or bug fixes, especially so we can iron out choices before implementing them.

ohemorange:
On the topic of exploring design decisions: what is the benefit of using fixtures/ dependency injection rather than a util library approach? Dependency injection hides where the code comes from, making it harder to understand how it's working.

adferrand:
So now about your comment before. It is right that as an architect, I am tending to build blocks with a larger picture in my mind, and making a clear state of that for the team is a good thing to challenge my design decisions and make the best out of it.

And in fact your observation for the integration tests could be extended also for the two other large pictures I have for certbot, that are windows support and dns plugins development workflow. 

So I will do that definitively. 

However, do not ask to much about this, by keeping in mind that I am not in a situation of a full-time commitment for certbot, but as an external contributor doing contributions in its sparse time. In particular, building a full plan corresponding more to a waterfall approach is a little to heavy. If I have ideas, it does not correspond to a fixed and structure specs, for which the work would correspond more to a full commitment. In my situation and for the certbot situation as an open source project, it should stay as a focus to a particular need, with some second thought about the generalization of this need, then a continuous specification and its refactoring all along the incremental sprints. More a scrum approach in fact.

adferrand:
That said, about the specific needs of improving the current integration tests executed continuously, you need to consider that this work is not a refactoring, but a full new piece of software. As so, bad design decisions will be made, inevitably, because of future needs that no one is aware of a this time.

So you will need to accept, for a reasonable amount of course, to leak theses unknown bad design decisions into the source code, and trust me about the fact I will stay committed to make the necessary modifications in future PRs. To sum up, yes, this code will not be perfect, and will contain questionable design decisions, but it will improve over the time.

adferrand:
Now for the specific discussion of the code lying here.

I disagree with you about the fact that an existing test framework is not used here. First you state this about `launch-integration-tests.py`, that make some work out of any test framework to setup a test environment. But this code is only 200 lines of code over the 2000 lines of code that will constitute the entire test system. And for everything else, it is build on top of `pytest`, so I am reusing well-known and strong patterns, and avoid, I hope, to reinvent the wheel.

In fact, about making a superset of pytest as `launch-integration-tests.py` is, it is something advised by the `pytest` dev team itself, like here: https://docs.pytest.org/en/latest/usage.html#calling-pytest-from-python-code

For the huge versatility of situations of test environment setup that exists, `pytest` cannot provide a complete coverage. Instead, some specific adaptations can be done, and it is the case here. 

To be complete, `pytest` has hooks that can be used to inject complex setup directly in `pytest`, and this would make `pytest` be the main entrypoint of the test system, and not `launch-integration-test.py`, removing the need to have a superset CLI (as I said that the entrypoint of a Python logic should always be a CLI.

I would really like to do that, but I found it to be extremely difficult to use it correctly for now. I can ensure you that for now, it would add more complexity in the code. But once a complete first and working iteration will be done, I will be glad to retake this work, and move the logic directly in pytest using hooks.

adferrand:
OK @ohemorange, regarding what I said about pytest hooks that would allow to use the pytest cli directly: I figured out how to debug pytest and that matter, and reimplemented today the acme ca setup logic as a pytest hook.

So now, there is no `launch-integration-test.py`, the entrypoint is pytest CLI itself. A convenient set of parameters, corresponding to the previous logic of `launch-integration-test.py`, is used in the tox `integration` env. All of this saves more than one hundred lines.

ohemorange:
Hey Adrien! Checking back in on this. Did you ever get a chance to write up a design doc?

adferrand:
I am on it currently. Draft is on the certbot drive share. I hope to complete it in the next days.

adferrand:
Merging my current version of the work, associated to the specifications given on https://docs.google.com/document/d/1G1b1psbvjg0mLdFXzo70eXQuyv3WsVVB9hv5ob4H7tg/edit?usp=sharing

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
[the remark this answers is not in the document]
ohemorange: if I want to find out more about what these objects are and how they're used, where would I look? and is the comment for this function the best place to link to that information?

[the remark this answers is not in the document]
adferrand: Standard fixtures are referenced here: https://docs.pytest.org/en/latest/reference.html#fixtures

Fixture worker_id is described here: https://pypi.org/project/pytest-xdist/

In general yes a comment on the function would be a good place for that, really close to where the fixture is declared. However at the end, only the certbot context fixture that I will add later will be consumed on exposed in the tests method, so the documentation can go there.

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,113 @@
+"""
+General conftest for pytest execution of all integration tests lying
+in the certbot_integration tests package.
+As stated by pytest documentation, conftest module is used to set on
```
ohemorange: link?

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,113 @@
...
+    acme_config = {}
+    if 'pebble' in config.option.acme_server:
+        acme_config['type'] = 'pebble'
+        acme_config['option'] = 'nonstrict' if 'nonstrict' in acme_server else 'strict'
+    else:
+        acme_config['type'] = 'boulder'
+        acme_config['option'] = 'v1' if 'v1' in acme_server else 'v2'
+    # By calling setup_acme_server we ensure that all necessary acme servers instances will be
+    # fully started. This runtime is reflected by the acme_xdist returned.
+    acme_xdist = acme.setup_acme_server(acme_config, workers)
```
ohemorange: why are we referring to this as xdist? is there some semantic overlap with the pytest-xdist package? does `xdist` mean "dictionary containing these specific values"?

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,202 @@
...
+            # Allow Boulder to ignore usual limit rate policies, useful for tests.
+            os.rename(join(instance_path, 'test/rate-limit-policies-b.yml'),
+                      join(instance_path, 'test/rate-limit-policies.yml'))
+
+        # This configure Pebble using precompiled containers.
+        if acme_type == 'pebble':
+            os.mkdir(instance_path)
+            with open(join(instance_path, 'docker-compose.yml'), 'w') as file_h:
+                file_h.write('''\
+version: '3'
```
ohemorange: why is this inlined instead of in a different file? why is there no file like this already in the pebble repo?

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,202 @@
...
+services:
+  pebble:
+    image: letsencrypt/pebble
+    command: pebble -config /test/config/pebble-config.json {strict} -dnsserver 10.30.50.3:8053
+    ports:
+      - 14000:14000
+    networks:
+      acmenet:
+        ipv4_address: 10.30.50.2
```
ohemorange: where are these addresses coming from?

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,202 @@
...
+        with open(join(instance_path, 'docker-compose.yml'), 'w') as file_h:
+            file_h.write('''\
+version: '3'
+services:
+  traefik:
+    image: traefik
+    command: --api --rest
+    ports:
+      - "5002:80"
+      - "8056:8080"
```
ohemorange: why 8056? can that number be named and put in a variable, especially if we are going to inline config files?

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,202 @@
...
+        response = requests.put('http://localhost:8056/api/providers/rest', data=json.dumps(config))
+        response.raise_for_status()
+
+        print('=> Finished traefik instance deployment.')
+    except BaseException:
+        print('Error while setting up traefik instance.')
+        raise
+
+
+def _launch_command(command, cwd=os.getcwd()):
```
ohemorange: very nice

certbot-ci/certbot_integration_tests/conftest.py
adferrand: Indeed, I named it after `pytest-xdist`. Because this pytest module makes a distribution of the tests, and `acme_xdist` is a dict that gives the distribution of acme server config over these distributed tests.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: They are arbitrary adresses, in a similar way than Boulder Docker config. I took them from the original `docker-compose.yml` in Pebble repo, but their global value does not matter. What matter is that they form together a contained network where you can predictibly find pebble, challtestsrv and the Docker host. 

This is not standard from the Docker world. Normally you autodiscover services through their FQDN, that the Docker system maintain automatically, abstracting you the need to know the actual network architecture. But this is not possible at this moment, due to specifics in Boulder/Pebble, and the current Docker implementation for Linux (on Windows, a special FQDN is set for the Docker host).

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: Arbitrary number again. What matters is that it is coherent over all parts that rely on its value.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: Thanks!

certbot-ci/certbot_integration_tests/conftest.py
adferrand: https://docs.pytest.org/en/latest/reference.html#hook-reference

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: The inline is because I need to control whether strict mode is enabled or not. It is kind of weird to make a search and replace on an existing file. For readability sake, I prefer to have the file in place, in particular as it is very short.

There is a docker-compose.yml in Pebble repo, but it is not suitable for production. It uses the local git repo to rebuild everything from the Dockerfiles. This implies 2 minutes of recompilation while the versioned dockers are already available on Docker Hub.

Even if the docker-compose file was suitable for production, I am quite not in fond with what is done with Boulder. It clone/checkout branch master and use directly the docker-compose file set their. This is weird in my opinion: it makes your stack depends on a third party. I prefer to use my own config stack, and control exactly what is setup in Travis, or every machine a developer would use to setup integration tests.

I sticked for Boulder on the current approach, to limit the modifications. But ultimately I would like also to setup Boulder with our own docker-compose file. Also, both Boulder and Pebble should be pinned from these on-the-fly generated docker-compose file, but this for another story we discussed before.

certbot-ci/certbot_integration_tests/conftest.py
adferrand: Added to the module doc.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: On the contrary to the challtestsrv port (8055) that will be used in various places, and so is in `acme_xdist`, this one is used only to the current function. So I put in a local variable.

adferrand: See my general comment below also.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: See also my comment below.

adferrand:
I updated `acme.py` to use variables for all ports and subnets used by the stack. Boulder and Pebble are quite different in how they can be configured: Boulder is mostly implicit, Pebble is mostly explicit.

The variables tries to make the two runtime as close as possible, and I put in comments wether each value is imported implicitly with Boulder, or set explicitly with Pebble.

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,215 @@
...
+def setup_acme_server(acme_config, nodes):
+    """
+    This method will setup an ACME CA server and a HTTP reverse proxy instances, to allow parallel
+    execution of integration tests against the unique http-01 port expected by the ACME CA server.
+    Instances are properly closed and cleaned when the Python process exits using atexit.
+    Typically all pytest integration tests will be executed in this context.
+    This method returns an object describing ports and directory url to use for each pytest node
+    with the relevant pytest xdist node.
+    :param dict acme_config: a dict describing the current acme server characteristics to setup.
+    :param str[] nodes: list of nodes name that will be setup by pytest xdist
```
ohemorange:
```suggestion
    :param str[] nodes: list of node names that will be setup by pytest xdist
```

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,215 @@
...
+ACME_V1_PORT = 4000   # Implicitly configured in boulder, not used in pebble
+ACME_V2_PORT = 4001  # Implicitly configured in boulder, explicitly in pebble
+ACME_SUBNET = '10.77.77'  # Implicitly configured in boulder, explicitly in pebble
+CHALLTESTSRV_PORT = 8055  # Implicitly configured in boulder, explicitly in pebble
+HTTP_01_PORT = 5002  # Implicitly configured both in boulder and pebble
+
+
+def setup_acme_server(acme_config, nodes):
+    """
+    This method will setup an ACME CA server and a HTTP reverse proxy instances, to allow parallel
```
ohemorange:
```suggestion
    This method will setup an ACME CA server and an HTTP reverse proxy instance, to allow parallel
```

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,215 @@
...
+    execution of integration tests against the unique http-01 port expected by the ACME CA server.
+    Instances are properly closed and cleaned when the Python process exits using atexit.
+    Typically all pytest integration tests will be executed in this context.
+    This method returns an object describing ports and directory url to use for each pytest node
+    with the relevant pytest xdist node.
+    :param dict acme_config: a dict describing the current acme server characteristics to setup.
+    :param str[] nodes: list of nodes name that will be setup by pytest xdist
+    :return: a dict describing the challenges ports that have been setup for the nodes
```
ohemorange:
```suggestion
    :return: a dict describing the challenge ports that have been setup for the nodes
```

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,215 @@
...
+    acme_xdist = {'challtestsrv_port': CHALLTESTSRV_PORT}
+
+    if acme_type == 'pebble':
+        acme_xdist['directory_url'] = 'https://localhost:{0}/dir'.format(ACME_V2_PORT)
+    else:  # boulder
+        port = ACME_V2_PORT if acme_option == 'v2' else ACME_V1_PORT
+        acme_xdist['directory_url'] = 'http://localhost:{0}/directory'.format(port)
+
+    acme_xdist['http_port'] = {node: port for (node, port)
+                               in zip(nodes, range(5200, 5200 + len(nodes)))}
```
ohemorange: why 5100 and 5200? and I assume we're sure we'll have fewer than 100 nodes? :)

certbot-ci/certbot_integration_tests/utils/acme.py
ohemorange: Well, we're already relying on a third party if we're pulling down images from docker hub. I do like having the speed up, but I worry about this getting out of sync with any changes made in the boulder/pebble repo, which is the main reason to use the docker-compose file in the repos.

certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -0,0 +1,45 @@
...
+import ssl
+import time
+import contextlib
+
+from six.moves.urllib.request import urlopen
+
+
+def check_until_timeout(url):
+    """
+    Wait and block until given url responds with status 200, or raise and exception
```
ohemorange:
```suggestion
    Wait and block until given url responds with status 200, or raise an exception
```

certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -0,0 +1,45 @@
...
+        except:
+            pass
+
+    raise ValueError('Error, url did not respond after 150 attempts: {0}'.format(url))
+
+
+@contextlib.contextmanager
+def execute_in_given_cwd(cwd):
+    """
+    Context manager that will execute any command in the given cwd after context entering,
```
ohemorange:
```suggestion
    Context manager that will execute any command in the given cwd after entering context,
```

certbot-ci/setup.py
```diff
@@ -0,0 +1,50 @@
+from setuptools import setup
+from setuptools import find_packages
+
+
+version = '0.29.0.dev0'
+
+install_requires = [
```
ohemorange: we're not using all of these yet

tox.ini
```diff
@@ -250,3 +250,16 @@ commands =
...
+    {[base]pip_install} acme . certbot-nginx certbot-ci
+    pytest {toxinidir}/certbot-ci/certbot_integration_tests \
+        --acme-server={env:ACME_SERVER:pebble-nonstrict} \
+        --cov=acme --cov=certbot --cov=certbot_nginx --cov-report= \
+        --cov-config={toxinidir}/certbot-ci/certbot_integration_tests/.coveragerc \
+        -W 'ignore:Unverified HTTPS request'
+    coverage report --fail-under=65 --show-missing
+passenv =
+    TRAVIS
+    DOCKER_*
```
ohemorange:
```suggestion
    DOCKER_*
    
```

tox.ini
```diff
@@ -250,3 +250,16 @@ commands =
...
+commands =
+    {[base]pip_install} acme . certbot-nginx certbot-ci
+    pytest {toxinidir}/certbot-ci/certbot_integration_tests \
+        --acme-server={env:ACME_SERVER:pebble-nonstrict} \
+        --cov=acme --cov=certbot --cov=certbot_nginx --cov-report= \
+        --cov-config={toxinidir}/certbot-ci/certbot_integration_tests/.coveragerc \
+        -W 'ignore:Unverified HTTPS request'
+    coverage report --fail-under=65 --show-missing
+passenv =
+    TRAVIS
```
ohemorange: where is this being used in the code?

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: I am following the official documentation to build the config, with it we can use pinned versions like in Boulder while the one in the git repo compile master, and I need to customize several ports in the config.

Also, the concern you express is based on the complexity in Boulder. Pebble is designed to be simple. 

Can we let this as it, and modify only if it becomes a problem?

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -0,0 +1,27 @@
+import requests
+import urllib3
+
+
+def test_hello_1(request, worker_id):
+    assert request.config.acme_xdist['http_port'][worker_id]
+    assert request.config.acme_xdist['https_port'][worker_id]
```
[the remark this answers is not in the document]
adferrand: The test here is about the existence of an http_port and an https_port for a given worker_id, not about an assertion about the worker_id value.

But this is a temporary test, I put here to illustrate the correct behavior of the test framework. It will be replaced by real tests later.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -0,0 +1,27 @@
...
+    try:
+        response = requests.get(request.config.acme_xdist['directory_url'], verify=False)
+        response.raise_for_status()
+        assert response.json()
+        response.close()
+    except urllib3.exceptions.InsecureRequestWarning:
+        pass
+
+
+def test_hello_2(request, worker_id):
```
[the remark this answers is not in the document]
adferrand: Yes, and that the tests are effectively distributed among the workers.

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+                          'defaulting to pebble-nonstrict')
+
+
+def pytest_configure(config):
+    """
+    Standard pytest hook used to add a configuration logic for each node of a pytest run.
+    :param config: the current pytest configuration
+    """
+    if not hasattr(config, 'slaveinput'):
+        with _print_on_err():
```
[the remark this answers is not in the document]
adferrand: During pytest-xdist setup, stdout is used to coordinate the primary node and the workers, and stderr is for communicating debug information to the user.

When the tests are under execution, the output capturing of Pytest kicks in to handle test output, and the redirection here is not needed anymore.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: Because I like this symetry ^^
5001 => 5100 - 5199
5002 => 5200 - 5299

Indeed, "only" 100 slots each, I dare you to find a machine that has enough power to handle that :p

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+    workers = ['master'] if not config.option.numprocesses\
+        else ['gw{0}'.format(i) for i in range(config.option.numprocesses)]
+
+    acme_server = config.option.acme_server
+    # Prepare the acme config server. Data is specific to an acme type. Module
+    # utils.acme_server will handle theses specifics.
+    acme_config = {}
+    if 'pebble' in config.option.acme_server:
+        acme_config['type'] = 'pebble'
+        acme_config['option'] = 'nonstrict' if 'nonstrict' in acme_server else 'strict'
```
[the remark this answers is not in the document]
adferrand: In fact, this can be solved even simpler. Pebble 2.0.0 has been released this week. All behaviors previously triggered by strict mode are now default. PR #6758 will make Certbot compatible with this version.

We can just remove all this logic about strict/nonstrict, and just propose a unified pebble server type.

certbot-ci/setup.py
adferrand: No, but next PR will. Do you want me to remove them until that?

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def pytest_configure(config):
+    """
+    Standard pytest hook used to add a configuration logic for each node of a pytest run.
+    :param config: the current pytest configuration
+    """
+    if not hasattr(config, 'slaveinput'):
+        with _print_on_err():
+            _setup_integration_tests(config)
```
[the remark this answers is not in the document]
adferrand: The hook `pytest_configure` is triggered for each node. Line 38 ensures that this branch is executed only on primary node. Then function on line 40 is designed to be run only on primary node.

What name to you have in mind for this method in this context?

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+        raise ValueError('Error: docker is required in PATH to launch the integration tests, '
+                         'but is not installed or not available for current user.')
+
+    try:
+        subprocess.check_output(['docker-compose', '-v'], stderr=subprocess.STDOUT)
+    except (subprocess.CalledProcessError, OSError):
+        raise ValueError('Error: docker-compose is required in PATH to launch the integration tests, '
+                         'but is not installed or not available for current user.')
+
+    workers = ['master'] if not config.option.numprocesses\
```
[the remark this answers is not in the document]
adferrand: Yes, it will work. None is a falsy value, so ['master'] will be taken in this case. This allow tests to run even if pytest-xdist is not present or enabled.

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+def pytest_configure(config):
+    """
+    Standard pytest hook used to add a configuration logic for each node of a pytest run.
+    :param config: the current pytest configuration
+    """
+    if not hasattr(config, 'slaveinput'):
+        with _print_on_err():
+            _setup_integration_tests(config)
+
+    if not os.environ.get('CERTBOT_ACME_TYPE'):
```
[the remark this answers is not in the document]
adferrand: I am not really proud of this code actually. It works, but it is not very elegant. Yes, it should be a fixture, or an object passed in the configuration of every worker node.

certbot-ci/certbot_integration_tests/utils/acme.py
```diff
@@ -0,0 +1,215 @@
+"""Module to setup an ACME CA server environment able to run multiple tests in parallel"""
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _setup_integration_tests(config):
+    """
+    Setup the environment for integration tests.
+    Will:
+        - check runtime compatiblity (Docker, docker-compose, Nginx)
+        - create a temporary workspace and the persistent GIT repositories space
+        - configure and start paralleled ACME CA servers using Docker
+        - transfer ACME CA servers configurations to pytest nodes using env variables
+    :param config: Configuration of the pytest master node
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: Sorry this needs to match a built-in value: if numprocesses in not set on the command line, then the primary node executes the tests, and its worker_id = 'master'.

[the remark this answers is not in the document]
adferrand: Done.

adferrand:
Since there is some concerns about the temporary tests as demo I put here, I moved to the final implementation I want for the fixtures. It is about exposing a unique fixture, `context`, that contains everything needed to interact with the certbot tests context. It is also from there that the other built-in fixtures are consumed, and not directly from the tests anymore, allowing a proper and unique way to describe them.

certbot-ci/certbot_integration_tests/nginx_tests/test_main.py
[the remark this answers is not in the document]
adferrand: Done through the fixture `context`.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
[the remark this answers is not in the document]
adferrand: Not needed anymore, everything is done through the fixture `context`.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: Fixed. Now it is passed directly in the worker nodes configuration.

certbot-ci/certbot_integration_tests/.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
+disable_warnings = module-not-imported,no-data-collected
```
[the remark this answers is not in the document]
adferrand: Because Coverage will complain about `acme`, `certbot` and `certbot_nginx` packages not been imported from the thread that executed the coverage for theses packages. Indeed, certbot is called from the command line using a subprocess.

The coverage works correctly however, and theses warnings are harmless messages, coming from a not standard execution, and then can be disabled.

certbot-ci/certbot_integration_tests/utils/acme.py
ohemorange: Well, why not just fix it in the docker-compose in Pebble since we're fixing it anyway? Like Boulder does? Since the code is written anyway, it should live in the best possible place; why not have all Pebble users get the benefit of it?

certbot-ci/certbot_integration_tests/.coveragerc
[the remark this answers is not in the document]
ohemorange: Great explanation! Let's stick it in there as a comment so future people know why the code is like that.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
ohemorange: `_setup_primary_node`, perhaps?

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
ohemorange: And pytest-xdist is in a separate context of some manner that won't be affected by this redirection?

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
ohemorange: I meant "this all" as in "the whole setup", rather than this particular line. What's the best way to test that? Manually set `numprocesses` to 0?

certbot-ci/setup.py
ohemorange: That would be ideal, to make it easier to track the history of what becomes relevant where.

certbot-ci/certbot_integration_tests/utils/acme.py
adferrand: That is right, I think the docker-compose.yml that is written in my code has a better usability than the one currently in Pebble repo. I will open a PR about that. But this is a commitment for Pebble maintainers, because the docker-compose.yml file needs to be updated at each release, like in Boulder.

So we will see. If the change is accepted, then we can checkout the GIT repo and use the docker-compose.yml. But if it is not, I really would like to stay with the current approach. 

If you agree, then we could consider this conversation as resolved, either by an update when the PR is accepted, or by sticking to the current code.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: Oh OK. It is the default situation in fact. If you remember, we did the change to avoid to set numprocesses to auto in certbot for a pytest execution. So if you do not set it, you are in the situation of only primary around, doing the tests. Or you set it explicitly, or execute the tox target that set it also.

I tested all cases if success of course.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: Yes, this function is a context manager, it ends with the configure hook, so before pytest-xdist start to distribute the tests among the nodes.

certbot-ci/certbot_integration_tests/.coveragerc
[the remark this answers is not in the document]
adferrand: Done.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: Good name. I take it.

certbot-ci/setup.py
adferrand: I cleared the dependencies.

tox.ini
adferrand: Old fix because because pytest-xdist needs to know when it is on Travis, because Travis CI is declaring an insane number of VCPUs (32) instead of the real number (2). Yes, it is hardcoded in pytest-xdist.

But this is not needed anymore, since another general fix I did, to set `TRAVIS` and `APPVEYOR` as passing by default to any tox env when executed from theses CI.

I remove

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+    Standard pytest-xdist hook used to configure a worker node.
+    :param node: current worker node
+    """
+    node.slaveinput['acme_xdist'] = node.config.acme_xdist
+
+
+@contextlib.contextmanager
+def _print_on_err():
+    """
+    With pytest-xdist, stdout is used for nodes communication, so print is uneffective.
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-ci/certbot_integration_tests/conftest.py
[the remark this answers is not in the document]
adferrand: And implemented.

certbot-ci/setup.py
```diff
@@ -0,0 +1,44 @@
+from setuptools import setup
+from setuptools import find_packages
+
+
+version = '0.29.0.dev0'
```
ohemorange: Should update before we merge

certbot-ci/certbot_integration_tests/utils/acme.py
ohemorange: Based on the comments there, looks like that'll get into the pebble repo! So looks promising for getting into this PR even.

certbot-ci/certbot_integration_tests/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+
+def pytest_addoption(parser):
+    """
+    Standard pytest hook to add options to the pytest parser.
+    :param parser: current pytest parser that will be used on the CLI
+    """
+    parser.addoption('--acme-server', default='pebble',
+                     choices=['boulder-v1', 'boulder-v2', 'pebble'],
+                     help='select the ACME server to use (boulder-v1, boulder-v2, '
+                          'pebble-nonstrict or pebble-strict), '
```
ohemorange: This comment is no longer accurate

certbot-ci/setup.py
adferrand: Done.

certbot-ci/certbot_integration_tests/conftest.py
adferrand: Done.

adferrand:
Yes, I think the two PR are in their final review stage and will be merged tomorrow. We can block on them.

certbot-ci/setup.py
```diff
@@ -0,0 +1,45 @@
...
+from setuptools import find_packages
+
+
+version = '0.32.0.dev0'
+
+install_requires = [
+    'pytest',
+    'pytest-cov',
+    'pytest-xdist',
+    'pytest-sugar',
```
ohemorange: We'll want to pin a version of this in `tools/dev_constraints.txt`.

adferrand: Done.

adferrand:
Yeeeeees!!!!

And thanks for your review, @ohemorange, this for sure was not an easy one!

Issue #6178, opened 2018-07-04
Can we please just have one Lexicon DNS authentication plugin?


chris-morgan:
*Or: The approach being used for DNS authentication plugins is madness.*

Currently there are ten DNS providers, five of which are using Lexicon.¹ There are eleven issues open for creating Lexicon-backed authenticators for specific DNS providers.²

Lexicon currently supports thirty-five providers—so should there be twenty-odd more tracking issues? And it has requests for almost twenty more, and doubtless more exist that could be added.

Because I’d like to use it, I just went and created a Vultr plugin, simply copying the DNSimple one and making the required changes—which were all trivial save one, just renaming “DNSimple” to “Vultr”, “dnsimple” to “vultr”, updating the URL to instruct the user on where get the API key from, and adjusting the check for suggesting to the user that the API key might be wrong (because DNSimple uses HTTP status 401, while Vultr uses 403). The total diff is +808 −5, having created 18 files and touched 12 more, and if I submit it and it is accepted (which the existence of #4600 suggests it would be) it will definitely add to the maintenance burden. Before I started, I was kind of hoping—though not expecting—that it might be +1 −0, given that it’s just piggybacking on top of Lexicon.

This situation seems quite absurd to me: surely the whole idea of Lexicon should be to have *one* authenticator plugin, Lexicon, and let the config file which is passed in specify all necessary details for whichever provider may be desired?

Only two arguments³ get passed for each authenticator plugin—a file to look in, and a propagation-seconds figure. Currently, usage might be like this:

```sh
certbot certonly \
    --dns-vultr \
    --dns-vultr-credentials ~/.secrets/certbot/vultr.ini \
    --dns-vultr-propagation-seconds 10 \
    -d example.com \
    -d www.example.com
```

With this ~/.secrets/certbot/vultr.ini:

```ini
dns_vultr_token = THIRTYSIXRANDOMCHARACTERSOFKEYPLEASE
```

I suggest it should look rather like this:

```sh
certbot certonly \
    --dns-lexicon \
    --dns-lexicon-credentials ~/.secrets/certbot/vultr.ini \
    --dns-lexicon-propagation-seconds 10 \
    -d example.com \
    -d www.example.com
```

With this ~/.secrets/certbot/vultr.ini (feel free to replace `[vultr]` with `provider = vultr` if the former is inconvenient):

```ini
[vultr]
auth_token = THIRTYSIXRANDOMCHARACTERSOFKEYPLEASE
```

[...]

PR #6588, opened 2018-12-11
Added support for linode version 4 tokens

trinopoty:

This PR adds support for the linode api token version 4 created by the new linode console.

### DIFF
--- a/certbot-dns-linode/certbot_dns_linode/__init__.py
+++ b/certbot-dns-linode/certbot_dns_linode/__init__.py
@@ -36,6 +36,7 @@ API Tokens page v4 <https://cloud.linode.com/profile/tokens>`_.
 
    # Linode API credentials used by Certbot
    dns_linode_key = 0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ64
+   dns_linode_version = [<blank>|3|4]
 
 The path to this file can be provided interactively or using the
 ``--dns-linode-credentials`` command-line argument. Certbot records the path
--- a/certbot-dns-linode/certbot_dns_linode/__init__.py
+++ b/certbot-dns-linode/certbot_dns_linode/__init__.py
@@ -27,7 +27,8 @@ Credentials
 
 Use of this plugin requires a configuration file containing Linode API
 credentials, obtained from your Linode account's `Applications & API
-Tokens page <https://manager.linode.com/profile/api>`_.
+Tokens page v3 <https://manager.linode.com/profile/api>`_ or `Applications &
+API Tokens page v4 <https://cloud.linode.com/profile/tokens>`_.
 
 .. code-block:: ini
    :name: credentials.ini
--- a/certbot-dns-linode/certbot_dns_linode/__init__.py
+++ b/certbot-dns-linode/certbot_dns_linode/__init__.py
@@ -27,8 +27,8 @@ Credentials
 
 Use of this plugin requires a configuration file containing Linode API
 credentials, obtained from your Linode account's `Applications & API
-Tokens page v3 <https://manager.linode.com/profile/api>`_ or `Applications &
-API Tokens page v4 <https://cloud.linode.com/profile/tokens>`_.
+Tokens page (legacy) <https://manager.linode.com/profile/api>`_ or `Applications
+& API Tokens page (new) <https://cloud.linode.com/profile/tokens>`_.
 
 .. code-block:: ini
    :name: credentials.ini
--- a/certbot-dns-linode/certbot_dns_linode/dns_linode.py
+++ b/certbot-dns-linode/certbot_dns_linode/dns_linode.py
@@ -57,13 +57,18 @@ class Authenticator(dns_common.DNSAuthenticator):
 
     def _get_linode_client(self):
         api_key = self.credentials.conf('key')
-        api_version = 3
+        api_version = self.credentials.conf('version')
 
-        # Match for v4 api key
-        regex_v4 = re.compile('^[0-9a-f]{64}$')
-        regex_match = regex_v4.match(api_key)
-        if regex_match:
-            api_version = 4
+        if not api_version:
+            api_version = 3
+
+            # Match for v4 api key
+            regex_v4 = re.compile('^[0-9a-f]{64}$')
+            regex_match = regex_v4.match(api_key)
+            if regex_match:
+                api_version = 4
+        else:
+            api_version = int(api_version)
 
         return _LinodeLexiconClient(api_key, api_version)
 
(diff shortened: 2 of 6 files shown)


trinopoty:
I made this change because the new linode console is now fully featured and linode is pushing it over the old one. Also, the v3 api has been officially marked as deprecated.
This PR adds support for dynamically choosing the v3 or v4 of the api based on the api key used.
The v3 key from the old console uses `^[0-9a-zA-Z]{64}$` and the v4 uses `^[0-9a-f]{64}$` format. This is what is used to distinguish between the v3 and v4 key and the api version to use.

ohemorange:
@bmw, do we have a test setup for linode?

bmw:
We do not :(

trinopoty:
This PR went through manual testing using both the v3 and v4 api keys.
A linode account is easy to set up. I believe having at least 1 active server unlocks the dns features.

stefantalpalaru:
I've used this patch in production with a v4 API key and it works as advertised.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
```diff
@@ -52,19 +56,44 @@ def _cleanup(self, domain, validation_name, validation):
...
+                'provider_name': 'linode',
+                'auth_token': api_key
+            })
+        elif api_version == 4:
+            self.provider = linode4.Provider({
+                'provider_name': 'linode4',
+                'auth_token': api_key
+            })
+        else:
+            raise errors.PluginError('Invalid api version specified: {0}.'
```
zjs: To make this a bit more future-proof (as Linode may eventually introduce a v5, at which point this error may be confusing), it might be helpful to list the supported API versions:
```suggestion
             raise errors.PluginError('Invalid api version specified: {0}. (Supported: 3, 4.)'
```

certbot-dns-linode/certbot_dns_linode/__init__.py
```diff
@@ -27,14 +27,16 @@
 
 Use of this plugin requires a configuration file containing Linode API
 credentials, obtained from your Linode account's `Applications & API
-Tokens page <https://manager.linode.com/profile/api>`_.
+Tokens page v3 <https://manager.linode.com/profile/api>`_ or `Applications &
+API Tokens page v4 <https://cloud.linode.com/profile/tokens>`_.
```
zjs: I'm not sure exactly what change to make here, but I think I might find this a bit confusing if I didn't already know what "v3" and "v4" were. A few more words here might be plenty to help new users understand.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
```diff
@@ -52,19 +56,44 @@ def _cleanup(self, domain, validation_name, validation):
...
 
     def _get_linode_client(self):
-        return _LinodeLexiconClient(self.credentials.conf('key'))
+        api_key = self.credentials.conf('key')
+        api_version = self.credentials.conf('version')
+
+        if not api_version:
+            api_version = 3
+
+            # Match for v4 api key
```
zjs: It seems like the main difference between v3 and v4 API keys is that v3 API keys are from a larger character set (`[a-zA-Z0-9]`). Thus, a v3 API key _could_ contain only numbers and lower case characters and match this regular expression.

What happens if a user who is unlucky enough to have a v3 API key that _looks like_ a v4 API key upgrades Certbot and tries to renew? It seems like this will fail.

For backwards compatibility (and predictability), would it be reasonable to _require_ that the API version be specified for API version 4? It's slightly less convenient for people creating credentials files for v4, but this would allow us to always use v3 if the API version is not set — which may be safer.

(If we make this change, there will also be some documentation changes required.)

trinopoty: Since I thought the chances of accidentally getting a v3 key that looks like a v4 key is quite low, I used the v3 as default and checked it for v4 key.
If a user does indeed get a v3 key that look likes a v4 key, only they'll have to add the override. Everyone else can not worry about extra configuration.

certbot-dns-linode/certbot_dns_linode/__init__.py
trinopoty: I'll try to change my wordings on this one.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
trinopoty: I think when v5 does come around, v3 will be long gone but I'll make the changes.

certbot-dns-linode/certbot_dns_linode/dns_linode.py
zjs: On second thought, the odds here seem... safe.

(More-likely-that-I'll-get-hit-by-a-meteor-tomorrow levels of safety, I think... `((10+6)/(10+26+26))^64 = 0.00000000000000000000000000000000000224187174%`.)

zjs:
@adferrand, will the approach here (where 1 Certbot plugin maps to 2 Lexicon providers) complicate the design you have in mind for a unified multi-provider Lexicon plugin (https://github.com/certbot/certbot/issues/6178#issuecomment-449513126)?

adferrand:
@zjs : not significantly I think, as there is already existing customizations that need to be done for other plugins backed by lexicon, ovh for instance.

So I need anyway to find a way to customize in the generic plugin specific logics for specific dns providers.

trinopoty:
@zjs another review maybe when you get the time?

trinopoty:
It seems that 'oldest' test is failing because there's no 'linode4' plugin.
Any advice on that?

adferrand:
Yes. Add `dns-lexicon==2.2.3` to `certbot-dns-linode/local-oldest-requirements.txt`.

trinopoty:
@adferrand That seems to have fixed the build. Thanks for the help.

displague:
@trinopoty Would it be possible to add a user-agent to these requests? (as a separate PR)
I've been adding this in other API client tools as a means to gauge usage on the Linode backend side.

displague:
Oh - It looks like this is already handled.  

https://github.com/certbot/certbot/blob/1f4297d0edc46801176bc072276db59126483f08/certbot/client.py#L67

trinopoty:
@displague I'm not sure if that user agent string is used by the lexicon library.
@adferrand Any thoughts on this?

alekam:
This patch works good! When it will be merged with master?

trinopoty:
@adferrand Will that do or do I need to change the wording?

mathewparet:
I am still facing this issue. I installed the latest plugin using "sudo apt-get install python3-certbot-dns-linode", but I still face the issue. Is it possible the changes aren't published yet?

bmw:
The changes were included in our 0.34.0 release, but your OS may not have packaged a newer version. For instance, [0.31.0 is the version in Debian stable](https://packages.debian.org/buster/python3-certbot-dns-linode).

If you need a newer version, I'd recommend asking your distro to offer newer versions of our packages or [run Certbot in Docker](https://certbot.eff.org/docs/install.html#running-with-docker).

If you have version 0.34.0+ of certbot-dns-linode and you're still having trouble here, please open an issue.

Issue #6582, opened 2018-12-10
Add simple unit tests that don't pin dependencies in test everything


bmw:
This would allow us to catch problems with the latest versions of our dependencies such as unexpected API breakage.

I think this could be done either by configuring `tox` to not use our install scripts for these tests or by setting an environment variable or flag that skips the normal pinning in our scripts.

I think this should just be done in the test-everything branch so failures here don't block all development.

adferrand:
There is an environment variable already available, if unit tests are executed through `install_and_test.py`: it is `CERTBOT_NO_PIN`, that needs in this case to be set to `1`.

adferrand:
Done, in #6595

PR #6590, opened 2018-12-11
Allow to execute a tox target without pinned dependencies

adferrand:
In response to #6582.

This PR passes the `CERTBOT_NO_PIN` environment variable to the unit tests tox envs. By setting `CERTBOT_NO_PIN` to `1` before executing a given tox env, certbot dependencies will be installed at their latest version instead of the usual pinned version.

I also moved the unpin logic one layer below to allow it to be used potentially more widely, and avoid unnecessary merging constraints operation in this case.

As warnings are errors now, latest versions of Python will break now the tests, because collections launch a warning when some classes are imported from `collections` instead of `collections.abc`. Certbot code is patched, and warning is ignored for now, because a lot of third party libraries still depend on this behavior.

An another PR will follow for `test-everything` branch.


adferrand:
cc @bmw

bmw:
I think my review on this above may have been missed.

adferrand:
Sorry, did not catch it, I do it know.

tools/pip_install.py
```diff
@@ -75,19 +75,29 @@ def main(args):
...
+            if requirements:
+                call_with_print('"{0}" -m pip install --constraint "{1}" --requirement "{2}"'
+                                .format(sys.executable, all_constraints, requirements))
 
-        merge_requirements(tools_path, test_constraints, all_constraints)
-        if requirements:
-            call_with_print('"{0}" -m pip install --constraint "{1}" --requirement "{2}"'
-                            .format(sys.executable, all_constraints, requirements))
+            with open(all_constraints, 'r') as file:
+                print(file.read())
```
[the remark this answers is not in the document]
adferrand: Yes, debug code. I remove it.

adferrand:
I integrated the changes from #6595.

tox.ini
```diff
@@ -64,6 +64,10 @@ source_paths =
     tests/lock_test.py
 
 [testenv]
+passenv = 
+    TRAVIS
+    APPVEYOR
```
bmw: The first two environment variables should no longer be needed here.

adferrand: Yes you are right. For the CI, either we are on travis and `export TOX_TESTENV_PASSENV=TRAVIS` applies, or we are on appveyor and `set TOX_TESTENV_PASSENV=APPVEYOR` applies.

And outside the CIs, `TRAVIS` and `APPVEYOR` are not set.

PR #6593, opened 2018-12-11
Update Certbot dependency to Lexicon to 3.x

adferrand:
This PR updates Lexicon dependency to the latest version available, 3.0.6, for every lexicon-based DNS plugins. It updates also the provider construction to use the new `ConfigResolver`object, and to remove the legacy configuration process.


adferrand:
cc @bmw

certbot-dns-cloudxns/setup.py
```diff
@@ -9,7 +9,7 @@
 install_requires = [
     'acme>=0.21.1',
     'certbot>=0.21.1',
-    'dns-lexicon>=2.2.1', # Support for >1 TXT record per name
+    'dns-lexicon>=3.0.6', # Support for >1 TXT record per name
```
zjs: It's probably safe to remove the ` # Support for >1 TXT record per name` comments now. I don't remember why it was added, but it doesn't seem useful at this point, and almost certainly isn't meaningful following this upgrade.

certbot/plugins/dns_common_lexicon.py
```diff
@@ -7,6 +7,11 @@
 from certbot import errors
 from certbot.plugins import dns_common
 
+try:
+    from lexicon.config import ConfigResolver
+except ImportError:
+    ConfigResolver = None  # type: ignore
```
zjs: When would an `ImportError` be raised here? Why do we want to ignore this?

certbot/plugins/dns_common_lexicon.py
```diff
@@ -100,3 +105,14 @@ def _handle_general_error(self, e, domain_name):
...
             return errors.PluginError('Unexpected error determining zone identifier for {0}: {1}'
                                       .format(domain_name, e))
+
+
+def build_lexicon_config(config_dict):
+    # type: (dict) -> ConfigResolver
+    """
+    Convenient function to build a Lexicon 3.x config object.
+    :param dict config_dict: the configuration specifics to apply
+    :return: an instanciated ConfigResolver object
```
zjs: nit-pick: spelling
```suggestion
    :return: an instantiated ConfigResolver object
```

certbot/plugins/dns_common_lexicon.py
adferrand: An `ImportError` could happen in theory because this code is in the certbot core part, and the core part does not require Lexicon.

We can ignore it because in effect, this code is called by a plugin that requires Lexicon, so the `ImportError` will not occur.

And yes definitively I should add a comment about this.

certbot-dns-cloudxns/setup.py
adferrand: Yes, it corresponds to the version where the support of multiple txt records for the same name has been formerly added in Lexicon (I mean with dedicated tests to ensure that it is possible).

It was some time ago, and not related to current version. All right I remove it.

bmw: Can we make this change in the other `setup.py` files as well?

certbot/plugins/dns_common_lexicon.py
```diff
@@ -7,6 +7,18 @@
 from certbot import errors
 from certbot.plugins import dns_common
 
+# Lexicon is not declared as a dependency in Certbot itself,
+# but in the Certbot plugins backed by Lexicon.
+# So we catch import error here to allow this module to be
+# always importable, even if it does not make sens to use it
```
bmw: s/sens/sense

certbot/plugins/dns_common_lexicon.py
```diff
@@ -7,6 +7,18 @@
...
+# if Lexicon is not available, obviously.
+# This import is not done at runtime to allow mypy type to
+# be handled correctly in any situation.
```
bmw: I don't understand this last sentence. We may not want to do it for other reasons, but bad thing would happen with mypy if we moved this import into the body of `build_lexicon_config`?

adferrand: Yes, in this case `ConfigResolver` would not be available to mypy and type check would fail. But indeed it the natural way to make import anyway, so I remove this phrase.

adferrand:
Alright @bmw. That is unfortunate, but I have dealt with it. For you record, the only Lexicon-based DNS plugin available in official repositories is dnssimple.

I modified the code to be able to run against Lexicon 2.x and 3.x. I reverted the oldest requirements to their original values for Lexicon. I also pushed the required version of acme to 0.25.0, in order to enable mypy. It is still lower that the oldest requirement claimed by certbot core itself.

Finally, relevant `setup.py` files are updated to be in sync with latest certbot version + oldest supported version of acme, and changelog is updated with packages impacted.

certbot/plugins/dns_common_lexicon.py
```diff
@@ -1,12 +1,23 @@
 """Common code for DNS Authenticator Plugins built on Lexicon."""
-
 import logging
 
 from requests.exceptions import HTTPError, RequestException
 
+# pylint: disable=unused-import, no-name-in-module
+from acme.magic_typing import Union, Dict, Any
```
bmw: nit: Can we reenable these globally disabled pylint checks after this line. Syntax is `# pylint: enable=unused-import, no-name-in-module`.

certbot-dns-dnsimple/local-oldest-requirements.txt
```diff
@@ -1,2 +1,3 @@
-acme[dev]==0.21.1
-certbot[dev]==0.21.1
+acme[dev]==0.25.0
+-e .[dev]
+dns-lexicon==2.2.1
```
bmw: I don't think this works unfortunately. The idea behind these `local-oldest-requirements.txt` files were to list the "oldest" dependencies on packages found "locally" in this repo. When used this new way, it appears the constraints in `tools/dev_constraints.txt` overrides the requirements here. See https://travis-ci.org/certbot/certbot/jobs/477426491#L1341.

While I think we should use this version of Lexicon in the oldest tests for all of our DNS plugins where we are claiming to support 2.2.1, we can't do it everywhere. For instance, the OVH plugin requires `dns-lexicon>=2.7.14`.

We haven't had this problem yet where we're testing different pieces of Certbot with different oldest constraints. Do you want to propose a solution for this? This could (and maybe should) be done in a different PR, however, if we want to delay fixing the problem, I think we should put `dns-lexicon==2.7.14` in `tools/oldest_constraints` in a new section for Ubuntu 18.04 Bionic to ensure we keep compatibility with Lexicon 2.x.

certbot-dns-dnsmadeeasy/setup.py
```diff
@@ -8,8 +8,8 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.21.1',
```
bmw: This should be updated to match the oldest requirements.

certbot-dns-cloudxns/local-oldest-requirements.txt
```diff
@@ -1,2 +1,2 @@
-acme[dev]==0.21.1
-certbot[dev]==0.21.1
+acme[dev]==0.25.0
```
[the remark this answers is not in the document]
adferrand: I updated it because to use `acme.magic_typing` in `certbot.plugins.dns_common_lexicon`, `acme>=0.25.0` is required. So it is not to update against Certbot in general, but only for Lexicon-based DNS plugins, that will rely on `certbot.plugins.dns_common_lexicon`. This change is also reflected in the relevant `setup.py` of course.

I did that to avoid to depend unnecessarily to too recent version of `acme`, only to the minimal one in the local oldest spirit. It avoid also the risk that you exposed in solution 1.

However if you prefer, we can go to solution 2., and put `-e acme` as local oldest.

CHANGELOG.md
```diff
@@ -22,6 +23,15 @@ all Certbot components during releases for the time being, however, the only
 package with changes other than its version number was:
 
 * acme
+* certbot-dns-cloudxns
+* certbot-dns-dnssimple
```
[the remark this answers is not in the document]
adferrand: Corrected.

certbot/plugins/dns_common_lexicon.py
adferrand: Corrected. I put it as inline comment, I saves two lines, and apply the disable only on this specific line.

certbot-dns-dnsmadeeasy/setup.py
adferrand: Corrected.

certbot-dns-dnsimple/local-oldest-requirements.txt
adferrand: Ah, I did not caught that. Indeed `local-oldest-requirements.txt` is of low priority not necessarily fulfilled, and local oldest requirement is not designed to isolate specific requirements for each project. I will think of a more general solution, it is not the first time I hit its limitations.

In the meantime, yes, I think putting a little higher `dns-lexicon` is a correct temporary solution.

adferrand: Oldest requirement updated.

bmw: Great! Can you make an issue to track this work on separate oldest requirements?

adferrand: PR is on its way ^^

certbot-dns-cloudxns/local-oldest-requirements.txt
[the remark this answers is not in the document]
adferrand: Ok. I update to latest dev version

PR #6597, opened 2018-12-11
Reduce build log verbosity on Travis

adferrand:
PR #6568 removed the `--quiet` option in pip invocations, because this option deletes a lot of extremely useful logs when something goes wrong. However, when everything goes right, or at least when pip install is correctly executed, theses logs add hundreds of lines that are only noise, making hard to debug errors that can be in only one or two lines.

We can have best of both worlds. Travis allows to fold large blocks of logs, that can be expanded directly from the UI if needed. It only requires to print in the console some specific code, that this PR implements in the `pip_install.py` script when the build is run in Travis (known by the existence of `TRAVIS` environment variable).

I also take the occasion to clean up a little `tox.ini`.

Note that AppVeyor does not have this fold capability, but it can be emulated using a proper capture of stdout/stderr delivered only when an error is detected.


adferrand:
cc @bmw

Issue #6596, opened 2018-12-11
Add flag to disable sleep at renewal


bmw:
The automatic renewal that comes with our Debian, Ubuntu, and Fedora/EPEL packages all already have a random sleep before renewal. As pointed out by @hlieberman, OS package maintainers can often handle this more elegantly. In the case of systemd (which almost all of our users use), rather than starting a Certbot process and letting it sleep, systemd can just delay starting the process for a random amount of time.

Harlan wanted to turn this off for the Debian packages, but doing so requires maintaining a patch to our code which is error prone and requires additional work. Instead, he asked if we could add a flag to disable this sleep. Only potential downside I see is copied and pasted command lines and Certbot getting yet another command line flag. I think we can maybe mitigate these issues though by giving it a good (and probably long) descriptive name and burying it in a help subtopic no one cares about.

hlieberman:
If you wanted, you could also put it in the renewal config only, instead of
a cli flag. We're already shipping one in Debian because of our disabling
of the log system.

bmw:
That's a good idea. We could check if our dependencies support only putting the flag in Certbot's configuration file.

@ohemorange also suggested hiding the flag in Certbot's help output and then we could just mention it in our [packaging guide](https://certbot.eff.org/docs/packaging.html).

adferrand:
The random sleep imon renew is also a problem for me in #6541. As the tests are run in a non interactive stdout, the random sleep is triggered and block the test execution. I solved this by adding a hidden cli option, but I am open to better approach if you have, to disable this behavior during tests.

ohemorange:
We should also make it only sleep when it actually needs to renew a cert.

bmw:
I'm actually having second thoughts about hiding this flag.

In addition to packagers, I think there are other valid use cases for this flag. Adrien brought up one from his own project and he's not the only one writing code around Certbot. I've talked to hosting providers who have done this such as https://i-mscp.net/filebase/file/33-letsencrypt/. There are also setups for things like [SaltStack](https://github.com/saltstack-formulas/letsencrypt-formula) and [Ansible](https://github.com/geerlingguy/ansible-role-certbot). Finally, we've proposed changing our approach to plugins so that people script around Certbot rather than the other way around.

[...]

PR #6599, opened 2018-12-13
Tag to disable random sleep upon a renew task

adferrand:
Following #6596, I extracted out my existing solution from #6541.

It adds a flag to the CLI, `--no-random-sleep-on-renew` to disable the random sleep when renew action is called from non interactive shells. The default behavior stays to make the random sleep if the flag is not provided.

The random sleep logic is also moved on layer down, to avoid to trigger it when no renew is going to be executed. Of course, this logic will be executed only once for a set of certificates to be renewed.

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1219,10 +1219,7 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
     helpful.add(
         "renew", "--no-random-sleep-on-renew", action="store_false",
         default=flag_default("random_sleep_on_renew"), dest="random_sleep_on_renew",
-        help='By default, renew action called from non interactive shells will be delayed'
-        ' of a random time between 1s to 8min, in order to spread the load to certificate'
-        ' authority servers that would arise from bunch of automated cron tasks at specific'
-        ' times of the day. Setting this flag will disable this delay for the current renew task.')
+        help=argparse.SUPPRESS)
     helpful.add(
         "renew", "--deploy-hook", action=_DeployHookAction,
         help='Command to be run in a shell once for each successfully'
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1213,9 +1213,8 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " run if an attempt was made to obtain/renew a certificate. If"
         " multiple renewed certificates have identical post-hooks, only"
         " one will be run.")
-    helpful.add(
-        "renew", "--renew-hook",
-        action=_RenewHookAction, help=argparse.SUPPRESS)
+    helpful.add("renew", "--renew-hook",
+                action=_RenewHookAction, help=argparse.SUPPRESS)
     helpful.add(
         "renew", "--no-random-sleep-on-renew", action="store_false",
         default=flag_default("random_sleep_on_renew"), dest="random_sleep_on_renew",
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1215,6 +1215,9 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " one will be run.")
     helpful.add("renew", "--renew-hook",
                 action=_RenewHookAction, help=argparse.SUPPRESS)
+    helpful.add("renew", "--no-random-sleep-on-renew", action="store_false",
+                default=flag_default("random_sleep_on_renew"), dest="random_sleep_on_renew",
+                help=argparse.SUPPRESS)
     helpful.add(
         "renew", "--deploy-hook", action=_DeployHookAction,
         help='Command to be run in a shell once for each successfully'
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -1213,11 +1213,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " run if an attempt was made to obtain/renew a certificate. If"
         " multiple renewed certificates have identical post-hooks, only"
         " one will be run.")
-    helpful.add("renew", "--renew-hook",
-                action=_RenewHookAction, help=argparse.SUPPRESS)
-    helpful.add("renew", "--no-random-sleep-on-renew", action="store_false",
-                default=flag_default("random_sleep_on_renew"), dest="random_sleep_on_renew",
-                help=argparse.SUPPRESS)
+    helpful.add(
+        "renew", "--renew-hook",
+        action=_RenewHookAction, help=argparse.SUPPRESS)
+    helpful.add(
+        "renew", "--no-random-sleep-on-renew", action="store_false",
+        default=flag_default("random_sleep_on_renew"), dest="random_sleep_on_renew",
+        help='By default, renew action called from non interactive shells will be delayed'
+        ' of a random time between 1s to 8min, in order to spread the load to certificate'
+        ' authority servers that would arise from bunch of automated cron tasks at specific'
+        ' times of the day. Setting this flag will disable this delay for the current renew task.')
     helpful.add(
         "renew", "--deploy-hook", action=_DeployHookAction,
         help='Command to be run in a shell once for each successfully'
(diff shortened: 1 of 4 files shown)


certbot/cli.py
```diff
@@ -1213,8 +1213,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         " run if an attempt was made to obtain/renew a certificate. If"
         " multiple renewed certificates have identical post-hooks, only"
         " one will be run.")
-    helpful.add("renew", "--renew-hook",
-                action=_RenewHookAction, help=argparse.SUPPRESS)
+    helpful.add(
```
ohemorange: Looks like there's a bunch of code cleaning going on in this PR, love seeing those! Let's put those in a separate PR to make tracing the history of code changes easier on us in the future.

certbot/main.py
```diff
@@ -68,6 +66,7 @@ def _suggest_donation_if_appropriate(config):
            "Donating to EFF:                    https://eff.org/donate-le\n\n")
     reporter_util.add_message(msg, reporter_util.LOW_PRIORITY)
 
+
```
ohemorange: This can go with the newline added above!

certbot/cli.py
```diff
@@ -1213,8 +1213,16 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
...
+        "renew", "--renew-hook",
+        action=_RenewHookAction, help=argparse.SUPPRESS)
+    helpful.add(
+        "renew", "--no-random-sleep-on-renew", action="store_false",
```
ohemorange: Let's suppress this instead of displaying it. It's meant for our own tests and for our packagers; we don't want people using it in their cron jobs instead of, say, using hooks. Even worse, we don't want people putting it in tutorials and it getting copied and pasted around, removing the benefit of the random sleep on LE's servers in the first place.

certbot/renewal.py
```diff
@@ -318,10 +321,11 @@ def renew_cert(config, domains, le_client, lineage):
 
 
 def report(msgs, category):
-    "Format a results report for a category of renewal outcomes"
+    """Format a results report for a category of renewal outcomes"""
```
ohemorange: This can go with all the added newlines!

certbot/renewal.py
```diff
@@ -387,15 +391,21 @@ def handle_renewal_request(config):
                            "instead. The renew verb may provide other options "
                            "for selecting certificates to renew in the future.")
 
-    if config.certname:
-        conf_files = [storage.renewal_file_for_certname(config, config.certname)]
-    else:
-        conf_files = storage.renewal_conf_files(config)
+    conf_files = [storage.renewal_file_for_certname(config, config.certname)] if config.certname \
```
ohemorange: This change makes the code harder to understand, since it can't fit on one line anyway.

certbot/renewal.py
```diff
@@ -424,7 +434,16 @@ def handle_renewal_request(config):
...
                 if should_renew(lineage_config, renewal_candidate):
-                    # domains have been restored into lineage_config by reconstitute
+                    # Apply random sleep upon first renewal if needed
+                    if apply_random_sleep:
+                        sleep_time = random.randint(1, 60 * 8)
+                        logger.info("Non-interactive renewal: random delay of %s seconds",
+                                    sleep_time)
+                        time.sleep(sleep_time)
+                        # We will sleep only once this day, folks.
+                        apply_random_sleep = True
```
ohemorange: This should be `False`

certbot/renewal.py
```diff
@@ -424,7 +434,16 @@ def handle_renewal_request(config):
...
+
+                    # Domains have been restored into lineage_config by reconstitute
```
ohemorange: it kills me but if you leave domains as lowercase git history will be nicer

certbot/renewal.py
```diff
@@ -437,14 +456,12 @@ def handle_renewal_request(config):
                     renew_skipped.append("%s expires on %s" % (renewal_candidate.fullchain,
                                          expiry.strftime("%Y-%m-%d")))
                 # Run updater interface methods
-                updater.run_generic_updaters(lineage_config, renewal_candidate,
-                                             plugins)
+                updater.run_generic_updaters(lineage_config, renewal_candidate, plugins)
```
ohemorange: another one for the cleanup pr!

certbot/renewal.py
```diff
@@ -437,14 +456,12 @@ def handle_renewal_request(config):
...
 
         except Exception as e:  # pylint: disable=broad-except
             # obtain_cert (presumably) encountered an unanticipated problem.
             logger.warning("Attempting to renew cert (%s) from %s produced an "
-                           "unexpected error: %s. Skipping.", lineagename,
-                               renewal_file, e)
+                           "unexpected error: %s. Skipping.", lineagename, renewal_file, e)
```
ohemorange: here too

certbot/renewal.py
```diff
@@ -455,5 +472,5 @@ def handle_renewal_request(config):
     if renew_failures or parse_failures:
         raise errors.Error("{0} renew failure(s), {1} parse failure(s)".format(
             len(renew_failures), len(parse_failures)))
-    else:
-        logger.debug("no renewal failures")
+
+    logger.debug("no renewal failures")
```
ohemorange: another thing for the cleanup PR!

adferrand:
Yes, I was charmed by Pycharm... and put more code cleaning that code logic. I will remove that, and make some PRs with only code cleaning, no changes of logic.

certbot/cli.py
adferrand: Is it the consensus or your opinion? Because it was my opinion also, to make a invisible flag, but @bmw was more on a visible flag in the issue mentionned, and @schoen is concerned about the possibility to disable the logic at all, iirc.

adferrand: More on that on the general discussion.

adferrand:
So currently the logic in this PR is in sync with the latest comment of @bmw in #6596.

However there is still an open discussion about this. Here as the options so far:
* visible flag
* invisible flag
* environment variable

I also expressed another proposition on IRC.

Because personally, I am more concerned about the hard coded time range than about enabling or not the sleep. Indeed I think now that the sleep should not be disabled, it is a good thing to dispatch load on ACME CA. However, I think that having a fixed time range, that will block a process for at most this time range "for no reason", and out of the context, is a weird approach.

What about the time range value, 8 min? Why not 10 min, 1 h or even 24 h to be sure to dispatch other the day?

So I would prefer to allow to modify the value of the range. Preferably as an hidden flag, as it is still more an option for certbot developers and integrators.

certbot/cli.py
adferrand: All right, then we all agree! I modify my code accordingly.

adferrand:
All right @ohemorange. All changes are implemented, ready for new review.

certbot/cli.py
ohemorange: looks like this one got missed

ohemorange:
Thanks for making the changes!

Also, ooh, a shiny new tool that looks like it automatically PEP8ifies things...... maybe I'll try pycharm...

As for time range vs disable:

The two uses of this flag that are the reason we're putting it in are:
1. for our packagers
2. for our tests

For both of these, completely disabling it solves the problem. So in the spirit of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), I'm in favor of keeping it as is.

adferrand:
FYI, after trying sublime text, atom, eclipse, visual studio code, I found out PyCharm to be the most advanced and efficient IDE I ever used for Python. But it makes non PEP8 compliant code extremely etchy ^^

certbot/cli.py
ohemorange: I'm excited to fix the formatting as well! When someone goes to trace the history of this line's changes, they will get confused if they see this commit, which seems like it might be related, but in fact isn't. If it's in a different PR, it will be clear that they need to go one more back, and since there's an upcoming cleanup PR anyway, might as well group it with those.

adferrand: All right, refactor reverted. But I will take my revenge on theses files with my PEP8 machine gun in a future PR ^^

ohemorange: I cannot wait

PR #6600, opened 2018-12-13
Correct boulder integration tests using the latest challtestsrv version

adferrand:
Urgent PR, to be merged as soon as possible to repair the CI.

Latest version of boulder test stack uses the latest challtestsrv instance built from pebble project. In this new version, the default IP to respond for any A record DNS request is not set anymore with the `FAKE_DNS` environment variable, but using a specific POST request against the running challtestsrv instance.

Currently all CI processes will be broken on integratino tests until fix, because the boulder instance will not be able to resolve the docker host, where certbot set the HTTP/HTTPS server to validate a given TLS/HTTP challenge.

This PR add required logic to `boulder-fetch.sh` to let integration tests be executed correctly.

### DIFF
--- a/tests/boulder-fetch.sh
+++ b/tests/boulder-fetch.sh
@@ -11,7 +11,6 @@ if [ ! -d ${BOULDERPATH} ]; then
 fi
 
 cd ${BOULDERPATH}
-sed -i "s/FAKE_DNS: .*/FAKE_DNS: 10.77.77.1/" docker-compose.yml
 
 docker-compose up -d boulder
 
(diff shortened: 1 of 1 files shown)


adferrand:
@bmw

cpu:
Sorry folks, my fault: https://github.com/letsencrypt/boulder/issues/3983 I'll also fix this on the Boulder side today.

cpu:
This should be fixed on the boulder side by https://github.com/letsencrypt/boulder/pull/3984

PR #6601, opened 2018-12-13
Update pinned urllib3

bmw:
GitHub notified us about a security vulnerability in our pinned version of `urllib3` earlier this week. It doesn't affect us, but we might as well upgrade anyway. I checked:

* There are no backwards incompatible features we care about listed at https://github.com/urllib3/urllib3/blob/master/CHANGES.rst.
* urllib3's dependencies don't also need to be updated according to https://github.com/urllib3/urllib3/blob/1.24.1/setup.py.
* The hashes match when obtained from different network vantage points.


adferrand:
Even extra secure? Because it requires cryptography 1.3, and currently we are pinned to 1.2. But it will not matter anymore soon.

bmw:
> Even extra secure? Because it requires cryptography 1.3, and currently we are pinned to 1.2. But it will not matter anymore soon.

I think that's just our oldest tests where we also have an [older version of urllib3 pinned](https://github.com/certbot/certbot/blob/6c8652a0a68df61a59cc7edf2f813c69ed68df29/tools/oldest_constraints.txt#L24). Outside of that, we're currently using [cryptography 2.2](https://github.com/certbot/certbot/blob/6b145a480ef19309c84bfecc61465571172a4f3f/letsencrypt-auto-source/pieces/dependency-requirements.txt#L63).

adferrand:
My bad, I mixed 1.x and 2.x...

bmw:
No worries!

bmw:
@ohemorange, are you able to review this simple change?

PR #6603, opened 2018-12-14
Use built-in support for OCSP in cryptography >= 2.5

### LINKED ISSUE #6594: Only conditionally use openssl command line
In [cryptography 2.4](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#24---2018-11-11) added support for making OCSP queries which I requested at https://github.com/pyca/cryptography/pull/4261#issuecomment-410908923.

We currently [shell out to openssl](https://github.com/certbot/certbot/blob/master/certbot/ocsp.py#L52) for this. Doing this makes things like Windows support and our new packaging harder because we depend on the openssl command line which must be installed in an OS specific way.

I think we should change this code to only conditionally shell out to `openssl` when the necessary features in `cryptography` aren't available. In the future, we can hopefully remove the code that shells out all together, but I think we need it while we are relying on packaging from Linux distros.

cc @adferrand

adferrand:
In response to #6594. [Fixes #6594.]

To execute OCSP requests, certbot relies currently on a openssl binary execution. If openssl is not present in the PATH, the OCSP check will be silently ignored. Since version 2.4, cryptography has support for OCSP requests, without the need to have openssl binary available locally.

This PR takes advantage of it, and will use the built-in support of OCSP in cryptography for versions >= 2.4. Otherwise, fallback is done do a direct call to openssl binary, allowing oldest requirements to still work with legacy cryptography versions.

Update: requirement is now cryptography >= 2.5, to avoid to rely on a private method from cryptography.


certbot/ocsp.py
```diff
@@ -1,34 +1,45 @@
...
-       # New versions of openssl want -header var=val, old ones want -header var val
-        test_host_format = Popen(["openssl", "ocsp", "-header", "var", "val"],
-                                 stdout=PIPE, stderr=PIPE, universal_newlines=True)
-        _out, err = test_host_format.communicate()
-        if "Missing =" in err:
-            self.host_args = lambda host: ["Host=" + host]
-        else:
-            self.host_args = lambda host: ["Host", host]
-
+        self.use_openssl_binary = not ocsp or enforce_openssl_binary_usage
```
schoen: Can I suggest `enforce_openssl_binary_usage or ocsp is None` here as potentially clearer for a reader of this code?

schoen:
(I added the `Fixes` keyword in the PR description above so that that issue will be closed automatically when this is merged.)

certbot/ocsp.py
adferrand: Indeed, clearer to read the positive assertions before the negative ones.

adferrand:
Thanks for the tip about Fixes keyword, I did not know about this mechanism!

schoen:
Overall, this looks very credible to me! Is there any chance that @ohemorange or @bmw would be up for giving it a complete review?

bmw:
I won't be able to get to it until next year as I'll be on vacation starting tomorrow. @ohemorange, if you're able to get to it before then, a review would help me out, but otherwise, I can take it when I get back.

ohemorange:
Haven't taken a look at the code yet, did do a quick check on 2.4 availability, and it's not great: none of our top distros have it yet, unfortunately. So we'd only be targeting our certbot-auto users for now.

bmw:
Yeah. The reason I personally wanted this feature is for installation methods where we do control our Python dependencies such as `certbot-auto` and our proposals for packaging going forward so we can drop the dependency on openssl (on x86 systems that can use cryptography's precompiled wheels).

adferrand:
Yes the code I did do not enforce cryptography 2.4. Just take advantage of it if it is available.

certbot/ocsp.py
```diff
@@ -62,12 +79,35 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        response = requests.post(url, data=request_binary,
+                                 headers={'Content-Type': 'application/ocsp-request'})
+        if not response.status_code == 200:
+            logger.info("OCSP check failed for %s (are we offline?)", cert_path)
+            return False
+        response_ocsp = ocsp.load_der_ocsp_response(response.content)
+        try:
+            logger.debug("OCSP certificate status for %s is: %s",
+                         cert_path, response_ocsp.certificate_status)
+            return response_ocsp.certificate_status == ocsp.OCSPCertStatus.REVOKED
```
alex: This is not actually checking the signature of the OCSP response against the issuer's key, so this is vulnerable to MITM attacks.

certbot/ocsp.py
```diff
@@ -77,22 +117,38 @@ def determine_ocsp_server(self, cert_path):
...
-            url, _err = util.run_script(
-                ["openssl", "x509", "-in", cert_path, "-noout", "-ocsp_uri"],
-                log=logger.debug)
-        except errors.SubprocessError:
-            logger.info("Cannot extract OCSP URI from %s", cert_path)
-            return None, None
+        if self.use_openssl_binary:
+            try:
+                url, _err = util.run_script(
+                    ["openssl", "x509", "-in", cert_path, "-noout", "-ocsp_uri"],
```
alex: You should be able to able to extract this data from the cert with cryptography, even if it's an older release.

certbot/ocsp.py
adferrand: Indeed, I did not catch it, thanks ! It is corrected: @alex, is it ok for you ?

certbot/ocsp.py
adferrand: Minimal cryptography version to get the OCSP OIDs is 0.8, and Certbot requires at least 1.2. So we are covered: code is simplified to use always cryptography here.

certbot/ocsp.py
```diff
@@ -62,37 +84,104 @@ def ocsp_revoked(self, cert_path, chain_path):
...
         host = url.partition("://")[2].rstrip("/")
+
         if host:
             return url, host
         else:
             logger.info("Cannot process OCSP host from URL (%s) in cert at %s", url, cert_path)
             return None, None
 
+
+def _check_ocsp_response_signature(response_ocsp, issuer_cert):
```
bmw: @reaperhulk, is there a better way for us to write this function? In particular, relying on the internal `_SIG_OIDS_TO_HASH` makes me nervous.

certbot/tests/ocsp_test.py
```diff
@@ -71,21 +82,18 @@ def test_ocsp_revoked(self, mock_run, mock_determine):
         self.assertEqual(self.checker.ocsp_revoked("x", "y"), False)
         self.assertEqual(mock_run.call_count, 2)
 
+    def test_determine_ocsp_server(self):
+        tmpdir = tempfile.mkdtemp()
+        try:
+            cert_path = os.path.join(tmpdir, 'cert.pem')
+            with open(cert_path, 'w') as file_handler:
+                file_handler.write(google_certificate)
```
bmw: Rather than writing out the certificates here and in `OSCPTestCryptography.setUp` below, we usually put stuff like this in the [testdata](https://github.com/certbot/certbot/tree/master/certbot/tests/testdata) directory. Once it's there, you can use `certbot.tests.util.load_vector` with the filename to get the path to the certificate on disk.

certbot/ocsp.py
```diff
@@ -1,36 +1,51 @@
...
+            test_host_format = Popen(["openssl", "ocsp", "-header", "var", "val"],
+                                     stdout=PIPE, stderr=PIPE, universal_newlines=True)
+            _out, err = test_host_format.communicate()
+            if "Missing =" in err:
+                self.host_args = lambda host: ["Host=" + host]
+            else:
+                self.host_args = lambda host: ["Host", host]
 
     def ocsp_revoked(self, cert_path, chain_path):
+        # type: (str, str) -> bool
```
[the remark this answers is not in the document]
adferrand: It will always return a boolean. Docstring is updated.

certbot/ocsp.py
```diff
@@ -62,37 +84,104 @@ def ocsp_revoked(self, cert_path, chain_path):
...
             output, err = util.run_script(cmd, log=logger.debug)
-        except errors.SubprocessError:
+        except errors.SubprocessError:  # pragma: no cover
             logger.info("OCSP check failed for %s (are we offline?)", cert_path)
             return False
-
         return _translate_ocsp_query(cert_path, output, err)
 
+    @staticmethod
+    def _ocsp_revoke_cryptography(cert_path, chain_path, url):
```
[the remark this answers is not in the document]
adferrand: Indeed. I put this as a method, because its equivalent for pyopenssl was a method. But it is better suited as module function.

certbot/ocsp.py
```diff
@@ -62,37 +84,104 @@ def ocsp_revoked(self, cert_path, chain_path):
         logger.debug(" ".join(cmd))
         try:
             output, err = util.run_script(cmd, log=logger.debug)
-        except errors.SubprocessError:
+        except errors.SubprocessError:  # pragma: no cover
```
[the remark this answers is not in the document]
adferrand: Seems to pass code coverage gates now. I remove the exclusion.

certbot/ocsp.py
```diff
@@ -62,37 +84,104 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+            issuer = x509.load_pem_x509_certificate(file_handler.read(), default_backend())
+        with open(cert_path, 'rb') as file_handler:
+            cert = x509.load_pem_x509_certificate(file_handler.read(), default_backend())
+        builder = x509.ocsp.OCSPRequestBuilder()
+        builder = builder.add_certificate(cert, issuer, hashes.SHA1())
+        request = builder.build()
+        request_binary = request.public_bytes(serialization.Encoding.DER)
+        response = requests.post(url, data=request_binary,
+                                 headers={'Content-Type': 'application/ocsp-request'})
+        if not response.status_code == 200:
```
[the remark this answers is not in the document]
adferrand: Corrected.

certbot/ocsp.py
```diff
@@ -44,10 +59,17 @@ def ocsp_revoked(self, cert_path, chain_path):
...
-        url, host = self.determine_ocsp_server(cert_path)
-        if not host:
+        url, host = RevocationChecker._determine_ocsp_server(cert_path)
+        if not host or not url:
             return False
+
+        if self.use_openssl_binary:
+            return self._ocsp_revoke_openssl_bin(cert_path, chain_path, host, url)
+        else:
+            return RevocationChecker._ocsp_revoke_cryptography(cert_path, chain_path, url)
```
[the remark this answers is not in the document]
adferrand: I like `_check_ocsp_*`. Names modified.

certbot/ocsp.py
```diff
@@ -62,37 +84,104 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+            logger.debug("OCSP certificate status for %s is: %s",
+                         cert_path, response_ocsp.certificate_status)
+            return response_ocsp.certificate_status == ocsp.OCSPCertStatus.REVOKED
+        except ValueError:  # pragma: no cover
+            logger.info("Invalid OCSP response status for %s: %s",
+                        cert_path, response_ocsp.response_status)
+            return False
+
+    @staticmethod
+    def _determine_ocsp_server(cert_path):
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/tests/ocsp_test.py
adferrand: Indeed. Way better to store and manipulate test assets. It is `vector_path` by the way to get the path, `load_vector` returns its content.

certbot/ocsp.py
adferrand: It saves a lot of boilerplate code, and seems strongly anchored in `cryptography` codebase. Anyway, waiting for opinions on that matter.

reaperhulk: This isn't wrong, but it's actually a missing API for `cryptography`. On Certificates, CRLs, and CSRs we have `signature_hash_algorithm`, which returns the value you want here. We should add that to cryptography.

adferrand: Method `signature_hash_algorithm` has been added on upstream, will be available with Cryptography 2.5. I added a TODO in the code to remember that.

certbot/ocsp.py
```diff
@@ -65,33 +87,103 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        #  response_ocsp.signature_hash_algorithm once cryptography 2.5 is released.
+        #  See https://github.com/pyca/cryptography/issues/4680
+        chosen_hash = x509._SIG_OIDS_TO_HASH[response_ocsp.signature_algorithm_oid]  # pylint: disable=protected-access
+    except KeyError:
+        raise UnsupportedAlgorithm(
+            "Signature algorithm OID:{0} not recognized"
+            .format(response_ocsp.signature_algorithm_oid)
+        )
+
+    issuer_cert.public_key().verify(
```
bmw: Something is going wrong here. When I test this against LE staging (or prod), I get:
```
Invalid signature for OCSP response on /etc/letsencrypt/live/example.com/cert.pem
```
which doesn't happen when using the openssl command line.

certbot/ocsp.py
```diff
@@ -65,33 +87,103 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+    builder = ocsp.OCSPRequestBuilder()
+    builder = builder.add_certificate(cert, issuer, hashes.SHA1())
+    request = builder.build()
+    request_binary = request.public_bytes(serialization.Encoding.DER)
+    response = requests.post(url, data=request_binary,
+                             headers={'Content-Type': 'application/ocsp-request'})
+    if response.status_code != 200:
+        logger.info("OCSP check failed for %s (are we offline?)", cert_path)
+        return False
+    response_ocsp = ocsp.load_der_ocsp_response(response.content)
```
bmw: I think we may need to change some things around after getting this response.

1. Looking at [RFC 6960](https://tools.ietf.org/html/rfc6960#section-3.2), we should probably make some other checks here.
2. We fail in an ugly way if the OCSP responder returns a failure response. The way I did this was try using a hash other than `SHA1()` above. When I did this with `--debug --verbose` on the command line, I got:
```
Exiting abnormally:
  File "/usr/local/bin/certbot", line 11, in <module>
  File "/home/bmw/src/certbot/certbot/main.py", line 1364, in main
  File "/home/bmw/src/certbot/certbot/main.py", line 1038, in certificates
    cert_manager.certificates(config)
  File "/home/bmw/src/certbot/certbot/cert_manager.py", line 89, in certificates
    _describe_certs(config, parsed_certs, parse_failures)
  File "/home/bmw/src/certbot/certbot/cert_manager.py", line 355, in _describe_certs
    notify(_report_human_readable(config, parsed_certs))
  File "/home/bmw/src/certbot/certbot/cert_manager.py", line 340, in _report_human_readable
    certinfo.append(human_readable_cert_info(config, cert))
  File "/home/bmw/src/certbot/certbot/cert_manager.py", line 266, in human_readable_cert_info
    if checker.ocsp_revoked(cert.cert, cert.chain):
  File "/home/bmw/src/certbot/certbot/ocsp.py", line 69, in ocsp_revoked
    return _check_ocsp_cryptography(cert_path, chain_path, url)
  File "/home/bmw/src/certbot/certbot/ocsp.py", line 169, in _check_ocsp_cryptography
    _check_ocsp_response_signature(response_ocsp, issuer)
  File "/home/bmw/src/certbot/certbot/ocsp.py", line 99, in _check_ocsp_response_signature
    chosen_hash = x509._SIG_OIDS_TO_HASH[response_ocsp.signature_algorithm_oid]  # pylint: disable=protected-access
  File "/home/bmw/src/certbot/venv/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/ocsp.py", line 30, in wrapper
    "OCSP response status is not successful so the property "
ValueError: OCSP response status is not successful so the property has no value
```

certbot/ocsp.py
adferrand: Indeed. The signature is failing, and I cannot understand why, I am doing as in the documentation from `cryptography`.

@reaperhulk or @alex, would you mind helping me to find what is wrong with my code ?

certbot/ocsp.py
adferrand: I resolved the issue 2., I am looking on the RFC to implement further checks.

certbot/ocsp.py
alex: You're using PSS, but I'm pretty sure the signature is PKCS1v15. Is there a reason you went for PSS?

adferrand: Only because it is written as it in the documentation for general RSA signature verification, but no specific walkthrough on OCSP. It is only noted that the OCSPResponse contains the DER encrypted payload used to build the signature, and that it could be used for verifying the signature, but nothing more.

https://cryptography.io/en/latest/x509/ocsp/#cryptography.x509.ocsp.OCSPResponse.tbs_response_bytes

adferrand: It worked ! Indeed PKCS1v15 is used for OCSP.

Really sorry, everything is explained in the relevant test from cryptography project:
https://github.com/pyca/cryptography/blob/master/tests/x509/test_ocsp.py

May I suggest to maybe link, from `tbs_response_bytes` of `OCSPResponse`, to the example provided in https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.tbs_certificate_bytes, as it is relevant for this situation ?

certbot/ocsp.py
adferrand: Ok, it added several checks from RFC 6960:
* certificate corresponds,
* issuer corresponds,
* thisUpdate/nextUpdate correspond

Except for checking responder is authorized, all optional checks from section 3.2 are covered.

adferrand:
@bmw, everything is ok, I tested against staging/prod and signature + optional checks are correct. Thanks a lot @alex for your help!

certbot/ocsp.py
```diff
@@ -65,33 +88,128 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+    except KeyError:
+        raise UnsupportedAlgorithm(
+            "Signature algorithm OID:{0} not recognized"
+            .format(response_ocsp.signature_algorithm_oid)
+        )
+
+    issuer_cert.public_key().verify(
+        response_ocsp.signature,
+        response_ocsp.tbs_response_bytes,
+        padding.PKCS1v15(),
```
bmw: If the CA is using an ECDSA intermediate cert, I think this code will have to be changed a bit. See [crypto_util.py](https://github.com/certbot/certbot/blob/693bac4b4943898ced3397fa0d6e436ef2343159/certbot/crypto_util.py#L243) or [cryptography's test_ocsp.py](https://github.com/pyca/cryptography/blob/master/tests/x509/test_ocsp.py#L384). To avoid problems in the future, can we make sure this code works with both approaches now?

If it makes sense, we should try to break out and reuse the code in `crypto_util.py`.

certbot/ocsp.py
```diff
@@ -65,33 +88,128 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+
+
+def _check_ocsp_response_signature(response_ocsp, issuer_cert):
+    """Verify an OCSP response signature against certificate issuer"""
+    try:
+        # TODO: (adferrand 2019-11-01) Following line can be improved using a direct call to
+        #  response_ocsp.signature_hash_algorithm once cryptography 2.5 is released
+        #  (watch out for retro-compatibility with 2.4 though).
+        #  See https://github.com/pyca/cryptography/issues/4680
+        chosen_hash = x509._SIG_OIDS_TO_HASH[response_ocsp.signature_algorithm_oid]  # pylint: disable=protected-access
```
bmw: After thinking about this a little more, I'm hesitant for us to ship code relying on private functionality like this. I just think it can break too easily and cause problems.

How about we disable this code from being used for production, but have a way to enable using `cryptography` during tests? Ideally these tests should check specifically for the version we're pinning to avoid potential problems for packagers running our test suite. After `cryptography` 2.5 is released, we can enable this feature in production using `signature_hash_algorithm`.

adferrand: I do not see how to easily make a specific behavior on this code for tests only. And adding code for tests in the software itself is bad in my opinion.

Let's consider doing this logic: if there is signature_hash_algorithm method, use it, otherwise use _SIG_OIDS_TO_HASH. And as soon as cryptography 2.5 lands, we escalate the minimal version for cryptography to 2.5 on Certbot, and remove the execution on _SIG_OIDS_TO_HASH to use signature_hash_algorithm only. Then we will have a unique version of Certbot, the next one, that in case a unique version of cryptography is installed in the OS, 2.4, will rely on the internal field, that, at the time of 2.4, exists and behaves correctly.

For any other situation, either openssl will be used, or cryptography will be used through signature_hash_algorithm.

I think it is a pretty safe situation.

certbot/ocsp.py
```diff
@@ -65,33 +88,128 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+    # Assert OCSP response corresponds to the certificate we are talking about
+    if response_ocsp.serial_number != request_ocsp.serial_number:
+        raise AssertionError('the certificate in response does not correspond '
+                             'to the certificate in request')
+
+    # Assert signature is valid
+    _check_ocsp_response_signature(response_ocsp, issuer_cert)
+
+    # Assert issuer in response is the expected one
+    if response_ocsp.issuer_name_hash != request_ocsp.issuer_name_hash:
```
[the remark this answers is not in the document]
adferrand: I agree for issuer_key_hash. But hash_algorithm is overkill. If hashes algorithms does not correspond, there is no way that hashes can correspond.

certbot/ocsp.py
```diff
@@ -65,33 +88,128 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        raise AssertionError('the issuer does not correspond to issuer of the certificate.')
+
+    # Assert nextUpdate is in the future, and that thisUpdate is not too old
+    if response_ocsp.next_update:
+        if response_ocsp.next_update < datetime.now():
```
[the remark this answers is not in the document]
adferrand: Agree. next_update can be in the "foreseeable future" minus 5 minutes.

certbot/ocsp.py
adferrand: Definitively! I refactored the code to be able to reuse `crypto_util`.

adferrand:
@bmw, I added several tests to check most of the logic added in OCSP, including the various checks against OCSPResponse.

certbot/crypto_util.py
```diff
@@ -239,29 +239,45 @@ def verify_renewable_cert_sig(renewable_cert):
...
                 Details: {1}".format(renewable_cert.cert, e)
         logger.exception(error_str)
         raise errors.Error(error_str)
 
 
+def verify_signed_payload(public_key, signature, payload, signature_hash_algorithm):
+    """Check the signature of a payload.
+
+    :param RSAPublicKey/EllipticCurvePublicKey public_key: the public_key to check signature
+    :param byte signature: the signature bytes
```
bmw: nit: s/byte/bytes

certbot/crypto_util.py
```diff
@@ -239,29 +239,45 @@ def verify_renewable_cert_sig(renewable_cert):
...
+    :param payload: the payload bytes
+    :param signature_hash_algorithm: algorithm used to hash the payload
```
bmw: Can we add the types of `payload` and `signature_hash_algorithm` too?

certbot/ocsp.py
```diff
@@ -65,33 +88,126 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        _check_ocsp_response(response_ocsp, request, issuer)
+    except UnsupportedAlgorithm as e:
+        logger.error(str(e))
+    except errors.Error as e:
+        logger.error(str(e))
+    except InvalidSignature:
+        logger.error('Invalid signature on OCSP response for %s', cert_path)
+    except AssertionError as error:
+        logger.error('Invalid OCSP response for %s: %s.', cert_path, str(error))
+    else:
```
bmw: Nice use of `else` :)

certbot/ocsp.py
```diff
@@ -65,33 +88,126 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+    if response_ocsp.serial_number != request_ocsp.serial_number:
+        raise AssertionError('the certificate in response does not correspond '
+                             'to the certificate in request')
+
+    # Assert signature is valid
+    _check_ocsp_response_signature(response_ocsp, issuer_cert)
+
+    # Assert issuer in response is the expected one
+    if (response_ocsp.issuer_key_hash != request_ocsp.issuer_key_hash
+            or response_ocsp.issuer_name_hash != request_ocsp.issuer_name_hash):
```
bmw: I fought GH to let me continue the thread at https://github.com/certbot/certbot/pull/6603#discussion_r248453185 and failed so here we go...

> I agree for issuer_key_hash. But hash_algorithm is overkill. If hashes algorithms does not correspond, there is no way that hashes can correspond.

I disagree. While I agree the chance of hash collisions in two different hash functions that are supported both on the client and server side for two different values is extremely unlikely, I don't think we get any cryptographic guarantees like that. I strongly prefer we also check `hash_algorithm` and err on the safe side.

certbot/ocsp.py
```diff
@@ -65,33 +88,126 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        raise AssertionError('the issuer does not correspond to issuer of the certificate.')
+
+    # Assert nextUpdate is in the future, and that thisUpdate is not too old
+    if response_ocsp.next_update:
+        if response_ocsp.next_update < datetime.now() - timedelta(minutes=5):
```
bmw: Shouldn't this be:
```suggestion
        if response_ocsp.next_update < datetime.now() + timedelta(minutes=5):
```

certbot/ocsp.py
```diff
@@ -65,33 +88,126 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+            raise AssertionError('next update is in the past.')
+        interval = response_ocsp.next_update - response_ocsp.this_update
+        if datetime.now() - response_ocsp.this_update > interval:
```
bmw: Can we add 5 minutes of leeway to this check too?

certbot/ocsp.py
```diff
@@ -65,33 +88,126 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+            raise AssertionError('this update is too old.')
+
+
+def _check_ocsp_response_signature(response_ocsp, issuer_cert):
+    """Verify an OCSP response signature against certificate issuer"""
+    # TODO: (adferrand 2019-11-01) A fallback for cryptography==2.4 is implemented here.
+    #   Once cryptography==2.5 is released, Certbot can require this new version and
```
bmw: Luckily, `cryptography` 2.5 was released last week.

Can we simplify this code and update this file to only use `cryptography` for OCSP if `response_ocsp.signature_hash_algorithm` is available?

certbot/crypto_util.py
```diff
@@ -239,29 +239,45 @@ def verify_renewable_cert_sig(renewable_cert):
...
+            verify_signed_payload(pk, cert.signature, cert.tbs_certificate_bytes,
+                                  cert.signature_hash_algorithm)
     except (IOError, ValueError, InvalidSignature) as e:
         error_str = "verifying the signature of the cert located at {0} has failed. \
                 Details: {1}".format(renewable_cert.cert, e)
         logger.exception(error_str)
         raise errors.Error(error_str)
 
 
+def verify_signed_payload(public_key, signature, payload, signature_hash_algorithm):
```
bmw: Thanks for doing this refactoring!

certbot/ocsp.py
adferrand: This is highly improbable, I would bet more confidently on the Armageddon movie scenario to happen in real life next yeqr. But it is mathematically not impossible, see https://crypto.stackexchange.com/questions/5304/can-there-be-two-hash-functions-without-common-collisions. And computationally almost free for us to check also the hash algorithm.

So I add it ^^

certbot/crypto_util.py
adferrand: Done.

certbot/crypto_util.py
adferrand: Done.

certbot/ocsp.py
adferrand: Yes. And here with an interval it is the contrary we want: if the interval is 30 min, we will not start to raise an exception if this_update is 30 min older than now, but if it is 35 min older than now.

certbot/ocsp.py
adferrand: Great! I update the logic.

certbot/crypto_util.py
adferrand: Your welcome^^

certbot/ocsp.py
```diff
@@ -1,53 +1,80 @@
...
 import re
-
+from datetime import datetime, timedelta
 from subprocess import Popen, PIPE
 
+try:
+    # Only cryptography>=2.5 has ocsp module
+    # and signature_hash_algorithm attribute in OCSPResponse class
+    from cryptography.x509 import ocsp  # pylint: disable=import-error
+    from cryptography.x509.ocsp import OCSPResponse  # pylint: disable=import-error,unused-import
```
bmw: nit: Can we remove this line and just use `ocsp.OCSPResponse` in the line below?

certbot/tests/ocsp_test.py
```diff
@@ -1,18 +1,34 @@
...
 
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes  # type: ignore
+from cryptography.exceptions import UnsupportedAlgorithm, InvalidSignature
+from cryptography import x509
+try:
+    # Only cryptography>=2.5 has ocsp module
+    # and signature_hash_algorithm attribute in OCSPResponse class
+    from cryptography.x509 import ocsp as ocsp_lib  # pylint: disable=import-error
+    from cryptography.x509.ocsp import OCSPResponse  # pylint: disable=import-error,unused-import
```
bmw: nit: Can we remove this line and just use `ocsp_lib.OCSPResponse` in the line below?

certbot/ocsp.py
bmw: Thanks. I agree with everything you said :)

certbot/ocsp.py
bmw: You're right! I was thinking about it backwards. What you have here is correct.

certbot/ocsp.py
```diff
@@ -65,33 +92,117 @@ def ocsp_revoked(self, cert_path, chain_path):
...
+        interval = response_ocsp.next_update - response_ocsp.this_update
+        if datetime.now() - response_ocsp.this_update > interval + timedelta(minutes=5):
+            raise AssertionError('this update is too old.')
+
+
+def _check_ocsp_response_signature(response_ocsp, issuer_cert):
+    """Verify an OCSP response signature against certificate issuer"""
+    # Following line may raise UnsupportedAlgorithm
+    chosen_hash = response_ocsp.signature_hash_algorithm
+    crypto_util.verify_signed_payload(issuer_cert.public_key, response_ocsp.signature,
```
bmw:
```suggestion
    crypto_util.verify_signed_payload(issuer_cert.public_key(), response_ocsp.signature,
```

Without this change, I'm getting:
```
Unsupported public key type
```
when doing manual testing.

certbot/ocsp.py
adferrand: Done.

certbot/ocsp.py
adferrand: Indeed. Corrected.

adferrand:
@bmw I updated the changelog.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 * Avoid to process again challenges that are already validated
   when a certificate is issued.
+* If possible, Certbot uses built-in support for OCSP from recent cryptography
+  versions instead of OpenSSL binary: as a consequence Certbot does not need
+  OpenSSL binary to be installed anymore if cryptography>=2.5 is installed.
```
bmw: nit: `s/OpenSSL binary/the OpenSSL binary/g`

There are two occurrences of this in this changelog entry. Without the "the" it totally makes sense, but with the "the", it sounds a little more natural/correct in my opinion.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 * Avoid to process again challenges that are already validated
```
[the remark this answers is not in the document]
adferrand: No problem, fixed!

CHANGELOG.md
adferrand: Corrected!

Issue #5805, opened 2018-03-29
proposal - additional environment vars populated for manual plugin


jvanasco:
I've been trying to figure out how/why multiple cloud DNS systems have been failing on certificates that  contain both a wildcard and root domain. Originally I thought the verification of a first domain was caching a record, which blocked the second domain for a TTL (see #5801). Now I think the issue may be a push of the first record at the DNS provider is being cached, and blocking the second record from being read. 

To help code around this, I'd like to propose 2 adjustments to the manual plugin:

1. (simple) Augment the environment variables populated in `plugins/manual.py:Authenticator._perform_achall_with_script` with 
* `CERTBOT_BATCH_LEN` which track how many items will be called by the auth process
* `CERTBOT_BATCH_IDX` which is the current index.  This would be accomplished with about 5 clean and short lines of code.  Here's an example (I wasn't sure how to extend tests for environment vars, they might not exist?) https://github.com/jvanasco/certbot/commit/8f005bc7c678f1bdcacad64825aab772efd73708
* hook scripts would then be able to detect progress, including being on the last item, by operations like `"$(($CERTBOT_BATCH_LEN <= ($CERTBOT_BATCH_IDX + 1)))"`

2. (complex) Adding versions of the above that are locked to the second-level domain. e.g. `CERTBOT_2LD_BATCH_LEN` and `CERTBOT_2LD_BATCH_IDX`.  This isn't hard to do, but will add a chunk of code that pre-analyzes the achalls.

With that information provided to the existing manual-auth-hooks, scripts can better determine when they need to 'sleep' or execute a request to flush dns (if possible).  The first option would allow for scripts to easily detect they are the last overall invocation of the hook; the second option would note that a particular domain is finished.

bmw:
I'm pretty hesitant to add complexity like this to our manual hooks. Hooks for the manual plugin are a nice way to automate simple tasks and that don't necessitate [writing your own plugin](https://certbot.eff.org/docs/contributing.html#writing-your-own-plugin), but as your needs get more specific and complex, I think that writing a simple authenticator plugin is a better option. Doing so only requires implementing a few functions in Python.

As a potentially even easier alternative, why not just have your hooks always include the longer sleep? It's not as efficient, but especially when it'll usually be run non-interactively by something like cron, does the longer wait time really matter in your use case?

jvanasco:
First, thanks for your time and feedback.

[...]

Issue #5484, opened 2018-01-25
Feature request: Add option to know what is the last domain passed to dns hook script


sahsanu:
Hi,

certbot doesn't provide an official hook to work with the API of my DNS provider so I'm using my own script. Well, we know every time we create/modify a record on our DNS provider it needs to propagate to all the authoritative DNS servers, in my case it takes 1 minute and 30 seconds approx., right now, as far as I know, there is no option to know what is the last domain that is being issued so to be bullet proof, every time the script creates the _acme_challenge TXT record, I wait 1:30 minutes, well I use the API to know when the propagation is finished but as I said, it takes 1:30 minutes. If I want to issue a certificate with 100 domains it will take approx. 2 hours and 30 minutes to complete the request. If I had a way to know what is the last domain I could write all the TXT records and when the last domain is created, I could add the check to know when the propagation has finished (or just a simple sleep...) so it takes only 1:30 minutes instead of 2 hours and 30 minutes.

It would be great if certbot provided a variable like `CERTBOT_LAST_DOMAIN=true` or something like that so the script could have the logic to sleep the right amount of time to let finish the DNS  propagation or to perform the needed checks only when there are no more domains to add to our DNS provider.

As a workaround I pass the last domain to my script:

`./certbot-auto certonly --manual --preferred-challenges=dns --manual-auth-hook "/path/to/hook/dnsapi.sh yyy.domain.tld" --manual-cleanup-hook /path/to/hook/dnsapi-cleanup.sh -d zzz.domain.tld,aaa.domain.tld,yyy.domain.tld`

and it works perfectly but would be great to have this feature added directly to certbot.

Thank you in advance.

Cheers,
sahsanu

SwartzCr:
Sounds like a question for @bmw

bmw:
We're currently rethinking a bit about how our plugins work, but if we don't end up making major changes there, I think can and should do something like this.

sahsanu:
That sounds good, we'll appreciate it.

Thank you very much. 

-----Original Message-----

bmw:
We'd accept a PR for this which adds an environment variable like `CERTBOT_REMAINING_CHALLENGES` that is set for both `--manual-auth-hook` and `--manual-cleanup-hook`. The value would be the number of additional times Certbot will call the hook for the current certificate. When the hook is called for the last challenge, `CERTBOT_REMAINING_CHALLENGES` would be 0.

This was discussed in https://github.com/certbot/certbot/pull/6607.

[...]

Issue #4628, opened 2017-05-08
Allow Certbot to run in parallel


bmw:
In 0.14.0 we added simple lockfiles to prevent concurrently running Certbot instances from corrupting your files. As you can see in https://github.com/certbot/certbot/pull/4449#issuecomment-299802507, some people had been running Certbot concurrently though. While doing this was unsafe (and undocumented), it'd be nice if things like certificate renewal could be run in parallel to speed up things for large integrations.

Doing this would be a significant increase in complexity to Certbot's locking mechanism though.

vrubiella:
Yes please!, no-lock option or lock-domain option.

mattlqx:
I'd really love to see the ability for Certbot to spawn multiple instances of the `--manual-auth-hook`/`--manual-cleanup-hook` scripts so that multiple DNS challenges in the same certificate request (ala SANs) can be processed in parallel. This would speed up individual requests significantly if you're using SANs and are using a DNS provider that has some settling time on each transaction.

bmw:
While I certainly see the benefits of this, to be honest, I really doubt the small Certbot team will ever have time to get to this so I'm closing this issue and marking it as wontfix.

PELock:
Well it's a good reason to allow it to run in parallel.

I have 3 domains expiring at the same time, I was out of home and it expired, all 3 of them.

- I had to run certbot-auto with DNS validation on each of them
- It failed 3 times (like it always does, because since the DNS gets refreshed, certbot-auto fails to read it and generates a new code each time, it's been like that on certbot-auto v1.0.x for over a year)
- I have to wait for like 45 minutes for each domain to get a proper cert
- So it's 3x 45 min when I need it the most!

bmw:
@PELock, if Certbot failing the challenge three times is a different issue than https://github.com/certbot/certbot/issues/7679, I'd encourage you to make a post at https://community.letsencrypt.org describing the problem. If it's a problem in Certbot, they'll probably send you back here to create an issue, but if it's a problem with your setup, they should be able to help you through it.

I'd also encourage you to automate your renewal by following the instructions at https://certbot.eff.org/lets-encrypt/pip-other. If you are using the manual plugin, you'll either need to switch to another plugin or write scripts to automate performing the challenges as described at https://certbot.eff.org/docs/using.html#hooks. If you need any help with this, I'd also recommend making a post at https://community.letsencrypt.org.

PR #6607, opened 2018-12-15
Manual delay hook

LeviticusMB:
DNS verification with the manual plugin can be painfully slow if multiple domains or hostnames are to be included in the certificate.

I see that various suggestions have been made to improve this, like #4628, #5484 and #5805.

This PR proposes a new hook that can be used to wait for the DNS propagation. The auth hooks can now just trigger the DNS updates in the background and when all updates have been performed, the delay hooks wait for the updates to propagate before the actual verification continues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Certbot adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 * Added the `update_account` subcommand for account management commands.
+* Added a --manual-delay-hook argument that can be used to wait for DNS updates.
 
 ### Changed
 
--- a/certbot/plugins/manual.py
+++ b/certbot/plugins/manual.py
@@ -120,6 +120,8 @@ permitted by DNS standards.)
     def add_parser_arguments(cls, add):
         add('auth-hook',
             help='Path or command to execute for the authentication script')
+        add('delay-hook',
+            help='Path or command to execute for the DNS propagation delay script')
         add('cleanup-hook',
             help='Path or command to execute for the cleanup script')
         add('public-ip-logging-ok', action='store_true',
@@ -135,7 +137,7 @@ permitted by DNS standards.)
 
     def _validate_hooks(self):
         if self.config.validate_hooks:
-            for name in ('auth-hook', 'cleanup-hook'):
+            for name in ('auth-hook', 'delay-hook', 'cleanup-hook'):
                 hook = self.conf(name)
                 if hook is not None:
                     hook_prefix = self.option_name(name)[:-len('-hook')]
@@ -166,8 +168,13 @@ permitted by DNS standards.)
                 self.tls_sni_01 = ManualTlsSni01(self)
                 self.tls_sni_01.add_chall(achall)
                 self.tls_sni_01.perform()
-            perform_achall(achall)
+            perform_achall(achall, False)
             responses.append(achall.response(achall.account_key))
+
+        if self.conf('delay-hook'):
+            for achall in achalls:
+                self._perform_achall_with_script(achall, True)
+
         return responses
 
     def _verify_ip_logging_ok(self):
@@ -184,7 +191,7 @@ permitted by DNS standards.)
             else:
                 raise errors.PluginError('Must agree to IP logging to proceed')
 
-    def _perform_achall_with_script(self, achall):
+    def _perform_achall_with_script(self, achall, isdelay):
         env = dict(CERTBOT_DOMAIN=achall.domain,
                    CERTBOT_VALIDATION=achall.validation(achall.account_key))
         if isinstance(achall.chall, challenges.HTTP01):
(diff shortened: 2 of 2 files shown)


LeviticusMB:
Note that this is just a POC. If there’s interest in adding such a hook, I could fix up a proper implementation.

adferrand:
I think your proposition is very interesting. As a contributor in Lexicon, that backs a lot Certbot DNS plug-in, and as the maintainer of a Docker that is handling DNS challenges from Certbot with Lexicon, it is not the first time I am confronted to this problematic.

Since wildcard certificates are out there, DNS challenges became a biggest part of the challenges that needs to be handled. But on the contrary to HTTP challenges, that can be treated as synchronous operations as it is basically put locally a file in a specific path, DNS challenges require most of the time asynchronous calls to Rest APIs, making this process highly asynchronous, and heavily slow if processed in a synchronous way. 

Your idea is focused on the main need, that is to wait DNS records modifications to be correctly propagated by the DNS provider, before asking the ACME CA server to trigger the challenge verification. I think your asynchronous approach can be generalized.

Conceptually, the point is to have four logical blocks that are synchronously executed, but where every operation in a given block can be executed in the order you want, including as massively in parallel. This is first, challenges committing, then challenges propagation validation, then challenges check by ACME CA, and finally challenges cleanup. In theory, what you say about DNS challenges could be also applied to HTTP challenges, even if it is less critical: after the files have been written in the HTTP server, you could want to check first that theses files are effectively served before asking the ACME CA to launch its logic.

adferrand:
So if I recall correctly, currently the provided authenticator hook (and same for clean up hook), is executed in a blocking loop for each domain described in the SANs of the current certificate.

To take fully advantage of your delay hook (that I would call more check propagation hook), theses authenticator/cleanup hooks should be executed in a non blocking loop, certainly as processes, and in the middle, the check propagation hook would run, it also as non blocking, to trigger the ACME CA call when all processes returned green.

adferrand:
However in this case, an independent check propagation hook may not be necessary. If we consider that an authenticator hook async process returning green means "challenge is written AND propagation has been checked", then everything can be done in the authenticator hook.

This does not allow a specific and global logic to be applied for checking the propagation. An example I have in my mind, is that for DNS challenges for domains all in the same DNS zone, you do not need to check each individual domain. You can just poll the entire TXT records of the zone, and wait for all expected records to appear in it.

adferrand:
And of course, retro-compatibility needs to be taken care of. What about existing hooks that effectively relies on synchronous operations?

Here it would be another argument for a separated class of hooks. Without any check propagation hook specified, the synchronous process is triggered. With it specified, then the asynchronous process is triggered.

OK, finished to put my thinking here ^^

bmw:
Instead of this design, what do people think of an environment variable like `CERTBOT_REMAINING_CHALLENGES` that is set for both `--manual-auth-hook` and `--manual-cleanup-hook`. The value would be the number of additional times Certbot will call the hook for the current certificate. When the hook is called for the last challenge, `CERTBOT_REMAINING_CHALLENGES` would be 0.

The reason I personally prefer this design is it avoids adding a 7th type of hook to Certbot to potentially confuse people on what they should use and while I admit I do not know what a user would want to do with this functionality, it's a little more flexible to know where you are in the list of challenges you're performing than getting called only on the last challenge.

Inspiration for this proposal came from https://github.com/certbot/certbot/issues/5805.

adferrand:
So if I understand correctly, with that design, every auth hook (and similarly clean up hook) would make non blocking requests to give hand back to certbot, except the last one that could make a blocking loop to check correct DNS propagation?

It could work, if a given hook has access to the list of all domains that are challenged, to be able to check DNS propagation for all tokens.

bmw:
To make sure we're talking about the same kind of blocking, I'm not proposing letting Certbot run hooks in parallel as was discussed earlier in this thread. The change I'm proposing here is quite small. We simply would add a decrementing counter in an environment variable. Each hook would be executed sequentially and Certbot will block on the hook to finish execution before continuing like it does now.

As for what the hook itself does though, yes, it wouldn't have to wait separately for each DNS change to propagate and it could only do after making the last one. I suspect most people would just have the script sleep for a long time. That's what all of Certbot's DNS plugins do.

Alternatively, they could in theory check each record. By the time `CERTBOT_REMAINING_CHALLENGES` is 0, the script will have been called once with each domain and token value. It would be up to the script to save these values somewhere to be reused during the last execution though.

adferrand:
Yes, with your design I had not a parallel execution in mind. Just a sequential execution that is fast, because the hook does not wait for anything except the last one.

It can work. I am OK with it. However, even if it not blocking with that design, I think it is beneficial in general to get the list of all domains in an env variable.

bmw:
I think there is value to a variable like that, but I think most people won't use it, it's another variable to learn about and understand when writing hook scripts, and you can get the same functionality by writing a little code yourself.

You OK leaving that out until a request for that feature arises from multiple users?

bmw:
Adrien and I talked about this out of band and he's OK with my design which I wrote up in https://github.com/certbot/certbot/issues/5484#issuecomment-546086974. @LeviticusMB, if you're interested in implementing this, I'll review the PR.

bmw:
I'm closing this PR because we'd like a slightly different design for this and this PR hasn't seen any response in a little over a month. If you're still interested in implementing this, feel free to open a PR with the new design at https://github.com/certbot/certbot/issues/5484#issuecomment-546086974. If you have any questions/concerns about the design, feel free to bring them up in that issue.

varun-da:
@bmw @adferrand I would live `CERTBOT_REMAINING_CHALLENGES`, this would speed up cert generation for us. Currently with manual-auth-hook for 28 subdomain wildcard SAN, 1 root domain wildcard SAN, 1 root domain, we are at 45 - 50 mins to generate a cert.

A implementation like this will greatly help us. Can you please consider opening this issue?

https://community.letsencrypt.org/t/parallel-run-of-certbot-manual-auth-hook-manual-cleanup-hook-for-validation-of-multidomain-cert/115352

varun-da:
moving the comment to #5484.

Issue #6605, opened 2018-12-14
Look into using pytest-xdist to push code to external servers to speed up local tests


bmw:
I just ran `tox` tests locally for the first time in a while (as it's the first time I've written a significant PR in a while) and they were frustratingly slow. `pytest-xdist` which we already use for running our tests in parallel has the ability to push the code to a remote server and run tests there which could significantly speed up how long it takes to run tests locally. This functionality is described at https://github.com/pytest-dev/pytest-xdist#sending-tests-to-remote-ssh-accounts.

I haven't deeply looked into this and there may be reasons it won't work or is a bad idea, but I think it's probably worth playing with.

EFF has some servers we could probably use for this. Unfortunately, we wouldn't be able to give external developers access to these servers and any options specifically related to them probably shouldn't be merged into this repo, however, there may be some generic config/options/scripts that external developers could easily use to set this up themselves on their own servers.

ohemorange:
Yeah they take a while, not sure they're so much faster on the remote server tbh? I usually only run one or two locally (py27, lint) anyway.

ohemorange:
Personally, both 6 minutes and 10 minutes are over the threshold for "find something else to do while waiting" so I end up switching tasks anyway. If we could get it down to, like, 2 minutes maybe it'd be worth it for me.

adferrand:
I did some benchmark on Ubuntu. For the serie of tox env you mentionned @bmw. After warm up and, servers are closed to me (Germany), 

My machine, a 3 years old gaming laptop: 5 min
Standard Digital Ocean droplet (40$/month): 8 min
CPU optimized Digital Ocean droplet (80$/month): 6 min
High-end CPU optimized Digital Ocean droplet (160$/month): 5 min

bmw:
> If we could get it down to, like, 2 minutes maybe it'd be worth it for me.

Not quite. Running `time tox -e py27,lint` on the server was 2m33.492s.

Another thing we could do that might speed things up both in this case and for all the environments I linked earlier is to run tests and lint at the same time like we do in josepy. We might have to switch from pylint to flake8 first, but I've personally been wanting to do that anyway.

This would result in fewer environments needing to be set up, but that single environment would run more tests. I think this is probably worth doing though, especially if there's an easy way for people to specify to `tox` that they just want to run unit tests or the linter without any of the other tests.

bmw:
[...]

PR #6609, opened 2018-12-17
Make tox default to any py3 test

bmw:
[envlist](https://tox.readthedocs.io/en/latest/config.html#conf-envlist) defines the default test environments that should be run if someone runs `tox` without specifying them on the command line or in an environment variable.

When playing with things for #6605, I noticed that we're not testing on Python 3.7 here. We could just also specify `py37` here, however, I personally think it's better to just test with any Python 3 version. My reasons for this are:

* All of our tests in things like Travis do (and I think should) specify which test environment that should be run so this setting isn't used there.
* When devs test locally, most of them do not all have the 4 supported versions of Python 3 installed and instead usually just have 1 so let's just tell `tox` to test with the Python 3 interpreter it can find and test with multiple versions in CI.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 
 [tox]
 skipsdist = true
-envlist = modification,py{34,35,36},py27-cover,lint
+envlist = modification,py3,py27-cover,lint
 
 [base]
 # pip installs the requested packages in editable mode


bmw:
> what about environment with Python < 3.4?

Good question! Maybe there's a cleaner way to do this, but our tests would currently fail with the output:
```
acme requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 3.3.whatever
```
This happens after installing packages but before running tests. If you couldn't even install the packages in Python 3.3, the error would be worse. (Although I'd be very curious to know what OS the dev is using where `python3` defaults to Python 3.3.)

It would be nice if `tox` could detect this earlier, but I'm not sure if there's an easy way to do that while getting the behavior I'd like.

PR #6612, opened 2018-12-18
Hostname length

### LINKED ISSUE #5311: Not a FQDN because label is too long
 
Ubuntu 14.04

git checkout of version 0.19.0, letsecnrypt-auto 

/letsencrypt-auto certonly --post-hook=service nginx reload --no-self-upgrade --webroot -w /var/www/letsencrypt -d production.ldt.mashbodev.com -d reallyabsurdlystupidlylongdomainnamethatisdefinitelygoingtobeoverthe63characterlimitbloodyhellthisislongchillyourbeansman.mashbodev.com

Requested domain reallyabsurdlystupidlylongdomainnamethatisdefinitelygoingtobeoverthe63characterlimitbloodyhellthisislongchillyourbeansman.mashbodev.com is not a FQDN because label reallyabsurdlystupidlylongdomainnamethatisdefinitelygoingtobeoverthe63characterlimitbloodyhellthisislongchillyourbeansman is too long.

Other tickets discussions have indicated that long domain names +64 characters can be validated as domains by including them as alternative subject names instead of the common name.  Also, isn't certbot supposed to take the first domain as the common name?  In this case production.ldt.mashbodev.com ? Which is under 64 characters and should be fine?

https://github.com/certbot/certbot/issues/1915
https://community.letsencrypt.org/t/ssl-for-a-63-character-max-number-of-characters-domain-name-s/36387/20

### LINKED ISSUE #5958: Python 3 socket.getfqdn() crashes when local hostname is more than 63 characters
https://community.letsencrypt.org/t/attempting-to-generate-cert-getting-unicodeerror-encoding-with-idna-codec-failed/61675

This prevents account registration in this case when Certbot is run with Python 3. It wasn't a problem in Python 2, which doesn't crash under these circumstances.

Do we want to stop using `socket.getfqdn()`, or do we want to ask the Python developers to change this behavior? (The hostnames in question *are* invalid according to the DNS specifications, so it's not quite clear what the proper behavior should be from the point of view of the `socket` library.)

We could also catch the specific exception and provide a more useful error to the user.

Knight-Ops:
This added a new error type of `HostnameTooLong` to certbot. It also implemented the catch for `UnicodeError` that is thrown from `socket.getfqdn()` in python 3 when the domain name does not abide by RFC3490 for a max length of 63 characters.

This should create a more accurate error for solving issues:
Fixes #5958
Fixes #5311

### DIFF
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -53,11 +53,14 @@ class Account(object):  # pylint: disable=too-few-public-methods
     def __init__(self, regr, key, meta=None):
         self.key = key
         self.regr = regr
-        self.meta = self.Meta(
-            # pyrfc3339 drops microseconds, make sure __eq__ is sane
-            creation_dt=datetime.datetime.now(
-                tz=pytz.UTC).replace(microsecond=0),
-            creation_host=socket.getfqdn()) if meta is None else meta
+        try:
+            self.meta = self.Meta(
+                # pyrfc3339 drops microseconds, make sure __eq__ is sane
+                creation_dt=datetime.datetime.now(
+                    tz=pytz.UTC).replace(microsecond=0),
+                creation_host=socket.getfqdn()) if meta is None else meta
+        except UnicodeError:
+            raise errors.HostnameTooLong("Expected hostname to be <= 63 characters according to RFC3490.")
 
         self.id = hashlib.md5(
             self.key.key.public_key().public_bytes(
--- a/certbot/account.py
+++ b/certbot/account.py
@@ -54,13 +54,14 @@ class Account(object):  # pylint: disable=too-few-public-methods
         self.key = key
         self.regr = regr
         try:
-            self.meta = self.Meta(
-                # pyrfc3339 drops microseconds, make sure __eq__ is sane
-                creation_dt=datetime.datetime.now(
-                    tz=pytz.UTC).replace(microsecond=0),
-                creation_host=socket.getfqdn()) if meta is None else meta
+            creation_hostname = socket.getfqdn()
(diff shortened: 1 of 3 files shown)


jsha:
I don't know, sorry!

schoen:
I don't know either but my hypothesis is that it was something that was originally going to go in the local account file in order to make it clearer where it came from. If so, it's probably not necessary anymore.

I don't know of any indication that it was ever used in ACME itself.

ohemorange:
Cool great ok then I say rip it out!

certbot/account.py
```diff
@@ -53,11 +53,15 @@ class Meta(jose.JSONObjectWithFields):
...
         self.key = key
         self.regr = regr
-        self.meta = self.Meta(
-            # pyrfc3339 drops microseconds, make sure __eq__ is sane
-            creation_dt=datetime.datetime.now(
-                tz=pytz.UTC).replace(microsecond=0),
-            creation_host=socket.getfqdn()) if meta is None else meta
+        try:
+            creation_hostname = socket.getfqdn()
+        except UnicodeError:
```
zjs: (Disclaimer: I'm not remotely familiar with this area of the code; I'm just skimming reviews.)

Are there other cases where a `UnicodeError` could be thrown? (E.g., would this same error also be thrown — and replaced with a message about hostname length — if the hostname contains invalid characters?)

ohemorange:
Hey @r0ck3tAKATrashPanda, are you still interested in working on this?

ohemorange:
No response since April, so I'm closing.

Issue #6464, opened 2018-11-01
Integration/system tests for DNS plugins


bmw:
I think we should do more testing beyond our simple unit tests, however, I've been thinking a bit about what that'd look like. Here's a relevant conversation I had with @adferrand about the topic:
```
[2018-11-01 11:02:29] <bmw> adferrand[m]: One thing I've been thinking about lately is how to better test Certbot's DNS plugins.
[2018-11-01 11:02:49] <bmw> I was curious how your project and Lexicon handle this. Do you maintain accounts with all these different providers?
[2018-11-01 11:30:15] <adferrand[m]> In Lexicon, we use vcrpy. It works with cassettes, which are yaml files that describes a serie of http request and their response. First, a dns provider developer executes the tests against the real dns api, because at this time, the developer has a valid account on it.
[2018-11-01 11:30:46] <adferrand[m]> During this test, cassettes are recorded, and committed in the repo.
[2018-11-01 11:31:57] <adferrand[m]> In all subsequent tests execution, in particular CI, vcrpy will mock transparently any requests emitted by python (vcrpy supports the most used http libraries).
[2018-11-01 11:32:57] <adferrand[m]> If for a given test method, with its related cassette, the request corresponds to the recorded one, vcrpy will return the recorded response.
[2018-11-01 11:46:05] <bmw> Interesting.
[2018-11-01 11:46:28] <bmw> One thing we've struggled with is how to initially test these plugins as well as to regularly run tests to ensure they still work with the provider's current API.
[2018-11-01 11:47:31] <bmw> We were planning to set up accounts, but doing this for every plugin someone writes based on Lexicon isn't very sustainable.
[2018-11-01 12:53:34] <adferrand[m]> Indeed, you cannot. And if I construct a universal lexicon plug-in, it is dozens of accounts to maintain potentially. Here with Lexicon, it is possible.
[2018-11-01 12:55:04] <adferrand[m]> The only limitation, is if an actual API changes its specifications, because cassettes are not relevant anymore, and you do not know it until someone executes again a real request. But APIs of this kind
[2018-11-01 12:55:46] <adferrand[m]> are really stable. It should occurs only from time to time, I think it is doable to correct cassettes when needed.
[2018-11-01 13:08:50] <bmw> Sure.
[2018-11-01 13:09:42] <bmw> The only thing I don't like about it is it would likely be our users who run the real request and hit the bug when things change, but maybe that's not the end of the world.
[...]

PR #6619, opened 2018-12-23
Proof-of-concept: Implement end-to-end tests for cloudflare plugin

zjs:
Implement [BATS](https://github.com/bats-core/bats-core) tests for the `dns-cloudflare` plugin to verify obtaining and renewing a certificate, obtaining a certificate for multiple domains, and a variety of negative cases involving incorrect, invalid, or missing credentials.

These test cases are skipped if the appropriate environment variables are not set. This will allow for execution of the subset of the end-to-end tests for DNS plugins for which credentials are available.

Towards: #6464

### DIFF
--- a/tests/dns/cloudflare.bats
+++ b/tests/dns/cloudflare.bats
@@ -1,0 +1,107 @@
+#!/usr/bin/env bats
+
+load 'common'
+
+function setup() {
+    # Skip if any required environment variable is unset or empty
+    [ -z "$TEST_CLOUDFLARE_DOMAIN" ]  && skip '$TEST_CLOUDFLARE_DOMAIN variable not set'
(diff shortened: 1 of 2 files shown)


adferrand:
I think however that the tests should be written in Python. I have ongoing work about that in fact, for the integration tests. See #6578 and #6541.

First because in the perspective of declarative code that is typically part of a test code, I think Python is far superior than Bash in what you can do, and you can reuse easily the logic of other tests. Second, because I think it is always a good idea to right test in the same language than the piece of software you want to test: the learning curve is less rough, as you can reuse as a developer your technical knowledge on the software to write the tests. Finally, because with Python you can use Pytest, that is one of the best test framework I know.

It makes a particular sense if the software you test is in fact a library, as you can execute directly any logic that would be relevant for a library imported in a given code. I know that it is not the case for Certbot yet, but it is something that could be done.

PR #6639, opened 2019-01-04
Remove some irrelevant history in using.rst

slokhorst:
The documentation only applies to the current version of Certbot.
Move Apache plugin details out of the table to make it more readable.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -44,8 +44,7 @@ a combination_ of distinct authenticator and installer plugins.
 =========== ==== ==== =============================================================== =============================
 Plugin      Auth Inst Notes                                                           Challenge types (and port)
 =========== ==== ==== =============================================================== =============================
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -83,8 +82,7 @@ the circumstances in which each plugin can be used, and how to use it.
 Apache
 ------
 
-The Apache plugin currently requires an OS with augeas version 1.0; currently `it
-supports
+The Apache plugin requires Apache 2.4+ and Augeas 1.0+; currently `it supports
```
schoen: This is a useful change, but while we're at it, I suggest that we get rid of the version references entirely! There is code in the Apache plugin to support 2.2 as well, so this information isn't current. We could just say "The Apache plugin currently supports...".

slokhorst:
@schoen agreed! I've made the change.

PR #6642, opened 2019-01-04
Fix oldest nginx integration tests

bmw:
#6636 broke [test-everything tests](https://travis-ci.org/certbot/certbot/builds/475173804) because `_common.sh` is a common file shared between Certbot and Nginx integration tests and `--no-random-sleep-on-renew` isn't defined for the version of Certbot used in the "oldest" integration tests.

This PR adds code to `_common.sh` to check the Certbot version and if it's new enough, add `--no-random-sleep-on-renew` to the command line. I repurposed `$store_flags` and stopped exporting it because it's not used anywhere outside of this file.

Other approaches I considered and decided against were:

1. Adding this flag in `certbot-boulder-integration.sh`. I decided against it because it's setting us up for the same problem in the future if the oldest version of Certbot is upgraded in the Nginx tests and we call `certbot renew`.
2. Just upgrading the oldest version of Certbot required by Nginx to avoid these issues. While this would work (with perhaps some unnecessary burden for our packagers), I think it's avoiding the real problem here which should now be able to addressed easily with the addition of `$other_flags` and `version_at_least`.

@adferrand or @ohemorange, are one of you able to review this?

### DIFF
--- a/tests/integration/_common.sh
+++ b/tests/integration/_common.sh
@@ -3,16 +3,15 @@
 root=${root:-$(mktemp -d -t leitXXXX)}
 echo "Root integration tests directory: $root"
 config_dir="$root/conf"
-store_flags="--config-dir $config_dir --work-dir $root/work"
-store_flags="$store_flags --logs-dir $root/logs"
 tls_sni_01_port=5001
 http_01_port=5002
 sources="acme/,$(ls -dm certbot*/ | tr -d ' \n')"
-export root config_dir store_flags tls_sni_01_port http_01_port sources
+export root config_dir tls_sni_01_port http_01_port sources
 certbot_path="$(command -v certbot)"
 # Flags that are added here will be added to Certbot calls within
 # certbot_test_no_force_renew.
-other_flags=""
+other_flags="--config-dir $config_dir --work-dir $root/work"
+other_flags="$other_flags --logs-dir $root/logs"
 
 certbot_test () {
     certbot_test_no_force_renew \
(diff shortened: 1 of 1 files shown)


tests/integration/_common.sh
```diff
@@ -3,24 +3,49 @@
...
+other_flags="--config-dir $config_dir --work-dir $root/work"
+other_flags="$other_flags --logs-dir $root/logs"
 
 certbot_test () {
     certbot_test_no_force_renew \
         --renew-by-default \
         "$@"
 }
 
+# Checks that the Certbot version is at least the given version number. This is
```
adferrand: As bash function does not return anything, it fails or succeed, and its return code represents this state, with 0 being a succeed. To avoid confusion here, and reduce the comment size, you could say "Succeeds if the Certbot..., fails otherwise." as the first sentence, and remove the last one.

tests/integration/_common.sh
```diff
@@ -3,24 +3,49 @@
...
+    # and minor version is greater than or equal to.
+    [ \( "$major" -gt "$1" \) -o \( "$major" -eq "$1" -a "$minor" -ge "$2" \) ]
+}
+
 # Use local ACMEv2 endpoint if requested and SERVER isn't already set.
 if [ "${BOULDER_INTEGRATION:-v1}" = "v2" -a -z "${SERVER:+x}" ]; then
     SERVER="http://localhost:4001/directory"
 fi
 
+if version_at_least 0 30; then
```
adferrand: You could put a reference to the PR that introduced the option, or this PR, to improve clarity for future readers about what is special about this option.

bmw:
Thanks for the review! I updated comments to match your suggestions.

PR #6644, opened 2019-01-05
 Open Allow updating domain pointed to by CNAME in DNS plugins

### LINKED ISSUE #6566: Allow updating domain pointed to by CNAME in DNS plugins
Say you have the domain `important.example.org` and want to obtain a wildcard certificate. One way you could do that is make a CNAME record at `_acme-challenge.important.example.org` that points to `dnsprovider.example.com`. This approach allows you to only put the DNS credentials for `dnsprovider.example.org` on your server for use with Certbot while keeping your credentials for `important.example.org` secure. This is a better approach from a security perspective and one we even recommended in https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation.

Unfortunately, Certbot doesn't let you do this because it tries to modify the domain you requested the certificate for instead of the domain pointed to by the CNAME. We cannot just always modify the domain pointed to by a CNAME, but giving the users the option to do this seems valuable to me.

There was some progress made on this in https://github.com/certbot/certbot/pull/5350 which is now closed, but the work is preserved in the branch [quinot/topic/dns-follow-cnames](https://github.com/certbot/certbot/tree/quinot/topic/dns-follow-cnames).

adferrand:
This PR integrates the work of @quinot from #5350, and updates it towards current master. Use case solved by this PR is described in #6566. 

The state of PR #5350 was a code review validated by @joohoi, except end-to-end tests for Google DNS plugin, and potentially a refactoring of the text help for `--dns-*-override-challenge`.

On top of that, I confirm from live tests that the Google DNS plugin works as expected. Text help seems complete for me, so I let reviewers to propose modifications if it seems to be necessary. 

Fixes #6566. 

cc @bmw @joohoi @quinot


bmw:
@adferrand, someone needs to review your changes, but are you happy with the code here? If so, I think we can just review any changes you made and treat the previous code as code reviewed by you.

From IRC, this still needs testing on:

- [ ] cloudxns
- [ ] gehirn
- [ ] linode

adferrand:
Besides merging master, and correcting a threshold for code coverage, I did not do any functional changes here. So from the code written by @quinot, I am happy with it.

However @joohoi was a little concerned by the verbosity of the help for the new cli option. So he may review here. 

Appart that, yes 3 providers to test left. I will update the PR when tests are done.

bmw:
Proposal:

cloudxns, gehrin, and linode are all based on Lexicon. The changes that are needed for one of our Lexicon based plugins to support the features added in this PR are extremely small. If the changes made to these three plugins match the ones made to other Lexicon based DNS plugins which we have successfully tested, they should also work, the chance of a bug is little to non-existent, and we should merge this PR.

Thoughts?

Daegalus:
I have been following this PR as its very needed for a project. And IF they are all Lexicon based, and other Lexicon based DNS plugins are indeed tested, I would agree with bmw that it should be pretty safe to merge.

adferrand:
And as part of the maintainers team for Lexicon, I can positively assert that these providers are fully covered for tests about TXT records creation/deletion, that are required to make functional the relevant Certbot DNS plugins. So I think also that the feature is covered and I am ready to take the responsability to merge it.

certbot/plugins/dns_common.py
```diff
@@ -29,13 +54,68 @@ def __init__(self, config, name):
...
     @classmethod
     def add_parser_arguments(cls, add, default_propagation_seconds=10):  # pylint: disable=arguments-differ
         add('propagation-seconds',
             default=default_propagation_seconds,
             type=int,
-            help='The number of seconds to wait for DNS to propagate before asking the ACME server '
-                 'to verify the DNS record.')
+            help='The number of seconds to wait for DNS to propagate before '
+                 'asking the ACME server to verify the DNS record.')
+        add('override-challenge',
```
bmw: What do people (e.g. @adferrand, @joohoi, etc.) think of the name `validation-domain` rather than `override-challenge`. I personally think the former is a little more descriptive of what is being set here but I'm curious what other people think.

certbot/plugins/dns_common.py
```diff
@@ -29,13 +54,68 @@ def __init__(self, config, name):
...
+            help='Override default challenge for following domains on the '
+                 'command line. Validation depends on a CNAME having been '
+                 'provisioned at the standard location '
+                 '(_acme-challenge.<domain>), pointing to the alternate '
+                 'location. The specified value may include "{domain}" '
+                 'and "{acme}" format strings, which will respectively '
+                 'expand to the domain being requested, and the corresponding '
+                 'standard ACME challenge location. For instance, when '
+                 'requesting "example.com", {domain} is "example.com", and '
+                 '{acme} is "_acme-challenge.example.com".')
```
bmw: I see the concern about the help text now. Not only is how to use this feature with CNAMEs somewhat complex, but this flag works like `--webroot-path` and includes a templating system.

I personally think we can't just say something like:
```
This flag is complex. Go read <url> to learn how to use it.
```
I think we should try to:

1. Provide a succinct but sufficient explanation about what this flag does for an experienced, technical user.
2. Maybe provide a link at the bottom to learn more.

@schoen, if you agree with me, are you interested in trying to propose text for (1) here? I think you would do a much better job at this than me.

certbot/plugins/dns_common.py
adferrand: I also think `validation_domain` is more explicit.

certbot/plugins/dns_common.py
adferrand: It is also what I proposed to @joohoi on the last discussion we had: just put a link to the web documentation in the CLI help, because I do not see how to make a short summary of this feature that would be understandable.

The main concern in the web documentation on which we are agree also, is that the documentation must not give to users a false feeling of security just because they use a delegate CNAME. There is more to do to build a really secured way that protect DNS credentials, and some work and study need to be done on the users side.

certbot/plugins/dns_common.py
joohoi: Yes, agreed!

adferrand: I change all assets to validation-domain.

certbot/plugins/dns_common.py
bmw: I think I'd personally rather have long help text about this flag than no help text other than a link. What harm is there in including an explanation of whatever length is necessary?

I think this help text should focus on what the flag **does** and we can put a link if we want explaining who, what, where, when, why, and how someone might want to use this feature.

Do people disagree? If so, why?

bmw:
@adferrand, @joohoi: As noted above, this flag as complex. Have one of the two of you tested various cases on it:

- Flag isn't set.
- Flag is before domain on CLI.
- Flag is after domain on CLI.
- Flag is set once for multiple domains.
- Flag is set multiple times for multiple domains.

certbot/plugins/dns_common.py
adferrand: Oh I misunderstood your last response. You want a description. Still, I do not see how to write something moderately short, as it involves a lot of context to understand what is its usage. Even by limiting to what it is doing.

adferrand: I think a native english speaker should handle the job. I would be not sufficiently clear, or too verbose. @schoen or @bmw, one of you would be willing to make a proposition?

adferrand:
@joohoi, concerning the latest comment of @bmw here, did you studied one or all situations described during your review? What would be the remaining ones?

joohoi:
Unfortunately the initial rounds of review were conducted so long time ago, that I honestly cannot remember all the possible corner cases I tested. I did a lot of testing on this however.

bmw:
Most of the blockers on this PR in my mind are due to how complicated the flag being added is. It's perhaps a bit unfortunate to revisit this decision now, but I wonder if we shouldn't reconsider the idea of automatically figuring out where to set the TXT record by following CNAMEs as was initially done in #5350. Doing this removes the complicated flag completely and avoids duplicating information that is already present in DNS.

Is the main concern here adding a new dependency? Are there any other reasons people thing we shouldn't do this?

adferrand:
Regarding the initial review @joohoi made, I think indeed that we can mitigate some points.

I think the biggest problem about adding a new dependency is that for OS packages, it requires to add another OS dependency. That said, `dnspython` is almost a standard package, similarly to `requests`, for its wide usage. So every Linux distribution has the proper OS package. And for the long term, the packaging redesign of Certbot will solve the issue.

Concerning the credentials, it is true that finally, the implicit CNAME resolution will make another credential be required despite the "visible" provider of the hostname requested. That said, this feature is an advanced one, and a given user needs to know what it is doing anyway. Also, credentials is already a problem if, for instance, you try to make a certificate composed of hostnames from different provider.

On the complexity on itself, well, it is the case anyway. Either with the original implementation, or the current one. At least with the original one, the explanation to give on the flag is easier to sum up.

bmw:
While we should definitely factor in the non-zero cost of adding a new dependencies (not doing this resulted in the creation of issues like https://github.com/certbot/certbot/issues/1301 which was created in response to backlash from some users), it's always a tradeoff and sometimes adding new dependencies is worth it for us. If we're worried about existing OS packages, the thing to check is if new enough versions of the package we want to depend on is present in the various distros where we're packaged and receiving updates.

It seems to me like the same credentials problem exists regardless of the choice made here.

We can certainly solve the relatively small UX/testing problems in this PR, but quickly revisiting this issue, automatic CNAME resolution seems like a better approach to me. It results in less things to explain so docs and `--help` output have less noise and it stops the user from having to write out, duplicate, and maintain the information already stored in DNS.

I'm curious what other people think though.

nerijus:
Yes, automatic CNAME resolution is a better approach and no need for duplication of information already in DNS.

mew1033:
Any updates here?

jsirex:
Will this PR be ever finished and merged?

adferrand:
I am not sure it will. We are currently working on #7244 that proposes basically a superior solution because it is automated, and is able to resolve more cases. However, it is for rfc2136 plugin only.

Once it is merged, I would like to make it generic for all dns plugins, if possible.

PR #6646, opened 2019-01-06
Add standalone DNS authenticator

siilike:
This is a plugin that uses an integrated DNS server to respond to the `_acme-challenge` records.

A subdomain with the relevant NS and A records needs to be set up, e.g. for `acme.example.com`:

```
acme     IN  NS  ns.acme.example.com.
ns.acme  IN  A   1.2.3.4
```

and then `_acme-challenge` for the domain that the certificate is requested for configured as a CNAME record to `domain.acme.example.com`.


bmw:
we're no longer accepting new first party plugins and i don't foresee that changing anytime soon so i'm going to close this

thanks for working on this though. it looks like this has already been added to the [list of 3rd party plugins](https://eff-certbot.readthedocs.io/en/stable/using.html#third-party-plugins)

PR #6647, opened 2019-01-08
Stop using staging in apacheconftests

### LINKED ISSUE #5938: Port apache conf tests to compatibility tests
Apache conf tests iterate over Apache configs checking that Certbot can parse them without blowing up or that it fails due to mod_macro. These tests also require making accounts on the staging server meaning that when the server is down for maintenance, our tests don't pass.

I think we should definitely stop using the staging server in our tests. As a way to do that and improve on the tests, we can add them to the Apache compatibility tests. This not only has Apache parse the files, but tests that it can perform challenges, set up certs, and perform enhancements and doesn't rely on boulder.

If we do this, we should make sure that we have a unit test that checks the mod_macro case (which we probably already do because we have 100% test coverage) and delete any configs testing this.

### LINKED ISSUE #6585: Remove reliance on staging from apacheconftests
https://github.com/certbot/certbot/blob/73458daac92c56bb211517a333e52fc1fef09715/tox.ini#L153 uses Let's Encrypt's staging server. I believe it is only used to register an account. I think we can remove this reliance by setting up a simple `/etc/letsencrypt` directory before running Certbot with a staging account. The way I'd do this is by creating a real staging account and deactivating it. The file that would need to be modified is https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/tests/apache-conf-files/apache-conf-test.

bmw:
Fixes #6585.

I wrote up three suggestions for fixing this at https://github.com/certbot/certbot/issues/6585#issuecomment-448054502. I took the middle approach of requiring the user to provide an ACME server to use. I like this better than the other approaches which were:

> Resolve #5938 instead of this issue.

There is value in these tests as is over the compatibility tests in that they don't use Docker and run on different OSes.

> Spin up a local Python server to return the directory object.

Trying to set up a dummy ACME server seemed hacky and error prone.

Other notes about this PR are:

* ~~I chose to put the Boulder setup for Travis in `.travis.yml` rather than `tox.ini` to keep it all in a single file. (We set up boulder for integration tests in `.travis.yml`.)~~
* I put the Pebble setup in `tox.ini` rather than `.travis.yml` as this seems much cleaner and more natural.
* I created a new `tox` environment called `apacheconftest-with-pebble` that reuses the code from `testenv:apacheconftest` so `apacheconftest` can continue to be used with servers other than Pebble like is done in our test farm tests.
* I chose the environment variable `SERVER` for consistency with our integration tests. I chose to not give this environment variable a default but to fail fast when it is not set.
* ~~I gave the environment variable the default of localhost to simplify things when running Boulder locally (like in Travis). You can still set `SERVER` to staging, however, I didn't make this the default as it kind of defeats the purpose of this PR.~~
* ~~I used Python rather than `curl` to check the value of `SERVER` because `curl` wasn't available on all test farm machines. `requests` is always installed in the virtual environment set up by `tox`.~~
* I ran test farm tests on this PR and they passed.

Finally, after this PR is merged, the `test-everything` branch will have to be updated to resolve changes in `.travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ matrix:
       addons:
     - python: "2.7"
       env:
-        - ACME_SERVER_URI="http://localhost:4000/directory"
         - BOULDER_FOR_TOX=1
         - TOXENV=apacheconftest
       sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,10 @@ matrix:
       before_install:
       addons:
     - python: "2.7"
-      env: TOXENV=apacheconftest
+      env:
+        - ACME_SERVER_URI="http://localhost:4000/directory"
+        - BOULDER_FOR_TOX=1
+        - TOXENV=apacheconftest
       sudo: required
     - python: "2.7"
       env: TOXENV=nginxroundtrip
@@ -92,8 +95,10 @@ addons:
 
 install: "travis_retry $(command -v pip || command -v pip3) install codecov tox"
 script:
+    - '[ -z "${BOULDER_FOR_TOX+x}" ] || travis_retry tests/boulder-fetch.sh'
     - travis_retry tox
-    - '[ -z "${BOULDER_INTEGRATION+x}" ] || (travis_retry tests/boulder-fetch.sh && tests/tox-boulder-integration.sh)'
+    # If BOULDER_INTEGRATION is set, run integration tests. If BOULDER_FOR_TOX isn't set, we still need to start boulder.
+    - '[ -z "${BOULDER_INTEGRATION+x}" ] || ( ( [ -n "${BOULDER_FOR_TOX+x}" ] || travis_retry tests/boulder-fetch.sh ) && tests/tox-boulder-integration.sh)'
 
 after_success: '[ "$TOXENV" == "py27-cover" ] && codecov'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ matrix:
       before_install:
       addons:
     - python: "2.7"
-      env: TOXENV=apacheconftest
+      env: TOXENV=apacheconftest-with-pebble
       sudo: required
     - python: "2.7"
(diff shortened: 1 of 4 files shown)


adferrand:
I put the following observation as a general comment, and not a suggested change under review, because even if I think it is a potential improvement, it does not worth to apply it now.

Indeed I apply it in the more general refactoring of integration tests in `certbot-ci`, so it will be reviewed when it is its time.

That said, so you put the test setup logic in `.travis.yml` to be consistent with what is done for the other integration tests.

However, I think `tox` is a way better mechanism to host setup logic and requirements about numerous isolated test environments, as it is in fact its what it has been designed for. It is the Boulder setup in integration tests that should move in `tox.ini`, leaving only the tests environment we want for Travis in `travis.yml` through the `TOXENV` variable. It would remove the hackish conditions block from the `script` section. 

But as said, this approach is in my agenda, so we will have the occasion to challenge it.

bmw:
I'm responding out of order to make my response flow better.

> In this case, you can substantially reduce the warm up of apacheconftest using Pebble instead of Boulder. Setup script is tests/pebble-fetch.sh.

Nice! I actually didn't realize this script had already landed. I made this change.

> However, I think tox is a way better mechanism to host setup logic and requirements about numerous isolated test environments, as it is in fact its what it has been designed for.

I agree, however, I didn't want to either:

a) Take the time to fix up the Boulder setup logic now or bundle it with the changes in this PR.
b) Duplicate that logic in multiple places which complicates how boulder setup works and makes correct maintenance more difficult.

With that said, we don't currently set up Pebble in Travis, so I think it makes sense to just put this setup in `tox` which I did.

@adferrand, are you able to finish the review here? I updated my original post to reflect the new changes.

bmw:
Travis failed reporting coverage to codecov so this PR is stuck waiting on codecov. Running the individual coverage test again didn't fix it, hopefully closing and reopening the PR will.

bmw:
Changed codecov settings and trying again.

PR #6651, opened 2019-01-09
Update test-everything

bmw:
This PR bundles a couple of changes. Normally we should try to split PRs like that up but all changes need to be made for tests to pass.

First, this PR merges `master` into the `test-everything` branch to resolve test failures created by #6597 and #6647.

Second, installing packages through Homebrew started failing. I fixed this by making use of the Homebrew options in `.travis.yml` which didn't exist when our current Homebrew scripts were written.

A summary of the changes there are:

* I removed `tests/travis-macos-setup.sh` (which only existed in the `test-everything` branch).
* I switched the packages installed by Homebrew to only install what's actually needed by the test.
* I started being explicit about the Python version we wanted because Homebrew has flipped back and forth on what the `python` package refers to.
* We stop calling `brew link python`. I think this is OK because `brew link <package>` just makes the package available in your `PATH` but we can see this is already the case with Python because the [tests are using 2.7.15](https://travis-ci.org/certbot/certbot/jobs/477576201#L265) which is available through Homebrew but not macOS.

To review this PR notice:

1. If you diff this branch against `95557fa9b` (which was `master` when these tests were written), all the changes are in `.travis.yml`.
2. The diff of `.travis.yml` against the `test-everything` branch is small and reflects the changes in the PRs mentioned above.

This PR should be merged and not squashed to avoid further merge conflicts.

@adferrand, are you able to give this a quick review?


bmw:
`brew` is failing which I don't believe is related to this PR, but I'm looking into it.

bmw:
Tests should pass now and I updated the PR description to match the new changes.

PR #6661, opened 2019-01-15
Setup cron build with specific jobs

### LINKED ISSUE #6659: Can we replace the test-everything branch with conditional jobs?
See https://docs.travis-ci.com/user/conditional-builds-stages-jobs. Could we just add conditional jobs that check `if: type = cron` (whose availability is specified in https://docs.travis-ci.com/user/conditions-v1#integration) and set up `master` to always run nightly? I don't see why this wouldn't work.

adferrand:
Fixes #6659 

It appears that the conditional statement `if: cron = type` works as expected:
* on a push event, jobs marked with this statement are completely ignored
* on a cron event, these jobs are executed

This PR takes advantages of it, by integrating the specific environments from `test-everything` branch to be merged into master, and add the conditional statement to execute them only during cron builds.

Once this PR is merged, a cron job can be set to build master at the appropriate time of the day, and `test-everything` will not be needed anymore.

For execution examples from this branch, see:
* during a push event: https://travis-ci.org/adferrand/certbot/builds/480127828
* during a cron event: https://travis-ci.org/adferrand/certbot/builds/480130236

No changelog, as this PR concerns only development stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_script:
 
 matrix:
   include:
+    # These environments are executed on push and cron events
     - python: "2.7"
       env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=all TOXENV=py27_install
       sudo: required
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,7 @@ before_script:
 
 matrix:
   include:
+    # These environments are executed on push and cron events
```
bmw: They're also executed on pull requests? Maybe we can just say "These environments are always executed."?

.travis.yml
```diff
@@ -64,6 +65,113 @@ matrix:
...
+      services: docker
+      if: type = cron
+    - python: "3.7"
+      dist: xenial
+      env: TOXENV=py37 BOULDER_INTEGRATION=v2
+      sudo: required
+      services: docker
+      if: type = cron
+    - sudo: required
+      env: TOXENV=nginx_compat
```
bmw: This TOXENV is listed twice.

adferrand: Thanks. Kind of mixed up with the numerous TOXENV.

.travis.yml
adferrand: Yes I am influenced by Gitlab that does not make the distinction at this level. Formely you always push to a reference, pull_request or not. It is better to say always, or only in cron jobs.

adferrand:
The before_install seemed to be useless with master configuration, as no tests were executed on osx, right? The test-everything branch seems to live correctly without it, add-ons are sufficient and better scoped, so I remove the before_install.

PR #6667, opened 2019-01-17
Call atexit handlers before test tearDown to reduce errors on Windows

adferrand:
When certbot is executing, several resources are opened. It is typically file handles and locks on them. Of course, theses resources need to be cleanup. It is done in Certbot by registering cleanup functions through `atexit` module, that ensures theses functions will be called when Certbot exit. This allow to not care about resource cleanup everywhere in the code, as it is processed globally.

The problem with `atexit` is it cleanup functions are called when the Python program exit. If the program is Certbot itself when used, this is Pytest in unit test execution. So during a unit test execution, cleanup is not called after a test and before its tearDown, but when Pytest exit, so way after tests and their respective tearDown.

But many tearDown implies to delete folders where this kind of resources are hold.

This is never a problem on Linux, thanks to its non-blocking file handling. It is usually not a problem on Windows, despite its blocking approach. But if the tearDown requires folder cleanup, exceptions are raised, and currently hidden as warnings. There is currently 504 exceptions of this type in Certbot core tests on Windows.

This PR starts to correct this situation. To do so, some of the functions cleanup normally called through `atexit`, are explicitly called as part of the tearDown process of relevant test classes, before directory removal is done. Theses situations come all from the `certbot.tests.util.TempDirTestCase`, so the code is in this specific tearDown process.

As a consequence, exceptions drop from 504 to 208.

Then there are still a significant part of them, that will be handled in later mitigation.

### DIFF
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -3,6 +3,7 @@
 .. warning:: This module is not part of the public API.
 
 """
+import logging
 import multiprocessing
 import os
 import pkg_resources
@@ -328,8 +329,14 @@ class TempDirTestCase(unittest.TestCase):
 
     def tearDown(self):
         """Execute after test"""
+        # Cleanup opened resources after a test. This is usually done through atexit handlers
+        # in Certbot, but in a pytest execution, atexit will kicks in after all tests execution.
+        # It is a problem on Windows, that does not accept to clean resources before closing them.
         # On Windows we have various files which are not correctly closed at the time of tearDown.
-        # For know, we log them until a proper file close handling is written.
+        logging.shutdown()
+        util._release_locks()  # pylint: disable=protected-access
+
+        # Still, some resources are not closed properly. For know, we log them until proper fix.
         # Useful for development only, so no warning when we are on a CI process.
         def onerror_handler(_, path, excinfo):
             """On error handler"""
@@ -340,10 +347,9 @@ class TempDirTestCase(unittest.TestCase):
                 warnings.warn(message)
         shutil.rmtree(self.tempdir, onerror=onerror_handler)
 
-class ConfigTestCase(TempDirTestCase):
-    """Test class which sets up a NamespaceConfig object.
 
-    """
+class ConfigTestCase(TempDirTestCase):
+    """Test class which sets up a NamespaceConfig object."""
     def setUp(self):
         super(ConfigTestCase, self).setUp()
         self.config = configuration.NamespaceConfig(


adferrand:
Further mitigations in relevant tests remove the last errors. Windows tests have no tearDown errored processes anymore.

certbot/tests/util.py
```diff
@@ -328,22 +329,21 @@ def setUp(self):
...
-        def onerror_handler(_, path, excinfo):
-            """On error handler"""
-            if not os.environ.get('APPVEYOR'): # pragma: no cover
-                message = ('Following error occurred when deleting the tempdir {0}'
-                           ' for path {1} during tearDown process: {2}'
-                           .format(self.tempdir, path, str(excinfo)))
-                warnings.warn(message)
-        shutil.rmtree(self.tempdir, onerror=onerror_handler)
+        # Cleanup opened resources after a test. This is usually done through atexit handlers in
+        # Certbot, but during pytest, atexit will kicks in only after each tearDown execution.
```
bmw: nit: I'd rephrase this to say "but during tests, atexit will not run registered functions before tearDown is called and instead will run them right before the entire test process exits."

certbot/tests/util.py
```diff
@@ -328,22 +329,21 @@ def setUp(self):
...
+        # It is a problem on Windows, that does not accept to clean resources before closing them.
+        logging.shutdown()
+        util._release_locks()  # pylint: disable=protected-access
 
-class ConfigTestCase(TempDirTestCase):
-    """Test class which sets up a NamespaceConfig object.
+        def handle_rw_files(_, path, __):
+            """Handle read-only files, that will fail to be removed on Windows."""
```
bmw: This is probably fine, by where are these read-only files being created?

certbot/tests/util_test.py
```diff
@@ -191,7 +191,10 @@ def test_ok_mode(self):
 
     def test_wrong_mode(self):
         os.chmod(self.tempdir, 0o400)
-        self.assertFalse(self._call(0o600))
+        try:
+            self.assertFalse(self._call(0o600))
+        finally:
+            os.chmod(self.tempdir, 0o700)
```
bmw: I'm assuming this is for Windows cleanup. Can you add an inline comment explaining that?

certbot/tests/util_test.py
```diff
@@ -277,15 +280,17 @@ def test_multiple(self):
...
+    def test_failure(self, mock_fdopen, _):
         err = OSError("whoops")
         err.errno = errno.EIO
         mock_fdopen.side_effect = err
         self.assertRaises(OSError, self._call, "wow")
 
+    @mock.patch("certbot.util.os.open")  # Avoid orphan file descriptor
     @mock.patch("certbot.util.os.fdopen")
-    def test_subsequent_failure(self, mock_fdopen):
+    def test_subsequent_failure(self, mock_fdopen, _):
```
bmw: I'd appreciate a sanity check, but I feel like it's more realistic these errors come from `os.open` rather than `os.fdopen` anyway. If you agree, how about we just mock `os.open` and raise the exception there?

certbot/tests/util_test.py
```diff
@@ -277,15 +280,17 @@ def test_multiple(self):
         for f, _ in items:
             f.close()
 
+    @mock.patch("certbot.util.os.open")  # Avoid orphan file descriptor
     @mock.patch("certbot.util.os.fdopen")
-    def test_failure(self, mock_fdopen):
+    def test_failure(self, mock_fdopen, _):
```
bmw: I'd appreciate a sanity check, but I feel like it's more realistic these errors come from `os.open` rather than `os.fdopen` anyway. If you agree, how about we just mock `os.open` and raise the exception there?

adferrand: In fact, it was what I did in the next commit, but it seems I forgot to push it. Will be resolved tomorrow then.

certbot/tests/util_test.py
adferrand: See the other similar issue.

certbot/tests/util_test.py
adferrand: Done.

certbot/tests/util.py
adferrand: Corrected.

certbot/tests/util.py
adferrand: Typically ACME account keys, that have `0o400`. If this is fine in Linux when you are root, it is not on Windows. From TempDirTestCase point of view, whose objective is to delete the temporary directory and all its content to matter what, if think it is a sane approach to put the abstraction here.

certbot/tests/util_test.py
```diff
@@ -277,20 +282,13 @@ def test_multiple(self):
...
-        self._call("wow")
-        err = OSError("whoops")
-        err.errno = errno.EIO
-        mock_fdopen.side_effect = err
-        self.assertRaises(OSError, self._call, "wow")
+    def test_failure(self):
+        with mock.patch("certbot.util.os.open", side_effect=OSError(errno.EIO)):
+            self.assertRaises(OSError, self._call, "wow")
+
+    def test_subsequent_failure(self):
```
bmw: I think the meaning of this test has been changed. Previously it was successfully using `_call` and was mocked to fail on the 2nd attempt. Now it's identical to `test_failure()`.

I don't understand the purpose of the previous test though and code coverage seems to be unchanged, so perhaps we should just delete this test altogether.

certbot/tests/util.py
bmw: Thanks for the explanation!

certbot/tests/util_test.py
adferrand: If code coverage is unchanged, then a higher level test is already relying on any behavior tested. Without specific purpose, yes I think the test can be deleted.

PR #6670, opened 2019-01-17
Pin dependency-requirements.txt in Dockerfile

### LINKED ISSUE #6580: Pin Python dependencies in Dockerfile in repo root
This would have prevented https://github.com/certbot/certbot/issues/6563.

My thought would be to just use the same versions that we pin in certbot-auto which can be found at https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/dependency-requirements.txt.

My thought would be to continue to install acme/certbot from the local files though. This allows us to do things like offer our [`rolling` tag based on master](https://hub.docker.com/r/certbot/certbot/tags/) and seems most conventional as otherwise the Dockerfile basically ignores the build context.

The problem this causes is the packages in our repo root don't have hashes and pip gets mad when you enforce hashes on some packages but not all. To resolve this I think we either need to:

1. Install package dependencies first in a separate `pip install` command and then install acme/certbot.
2. Remove hashes from the file. We probably want to do this at runtime because we do want the file with hashes in `certbot-auto`. This is the approach taken at https://github.com/certbot/certbot/blob/353d09258575a5a8bfcfe7cb145834f787f2b7bb/tools/pip_install.py#L51.

I think I personally prefer the 1st option.

sydneyli:
Fixes #6580

having been changed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ VOLUME /etc/letsencrypt /var/lib/letsencrypt
 WORKDIR /opt/certbot
 
 COPY CHANGELOG.md README.rst setup.py src/
+COPY letsencrypt-auto-source/pieces/dependency-requirements.txt .
 COPY acme src/acme
 COPY certbot src/certbot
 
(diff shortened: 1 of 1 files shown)


bmw:
Closing and reopening because Travis forgot to run.

bmw:
> If rerunning Travis & AppVeyor doesn't fix codecov, ...

It did! Merging.

PR #6671, opened 2019-01-17
Pin Python dependencies for each DNS plugin

### LINKED ISSUE #6581: Pin Python dependencies installed in DNS plugins
While #6580 suggests a simple way to pin Python dependencies in our main Docker image, doing it each for each of our DNS plugins is slightly more work but also valuable to avoid the same problem.

The way we pin these dependencies elsewhere is by using https://github.com/certbot/certbot/blob/master/tools/dev_constraints.txt which is a catch all list of all dependencies. Maybe we use this catch all file that we initially made just for our dev setup, but I'm not sure it will work because it's not in the build context of the Dockerfile which is its directory and everything it contains. Maybe we can change the build context on Docker Hub? Even so, using this dev file feels a little hacky to me.

Another way we could approach is this is creating new files in each DNS plugin directory pinning the dependencies it has. These dependencies can be found by running `pip freeze` and I think that all of our packages and dependencies already pinned in #6580 should be removed these files.

While this works, I'd like to avoid us having multiple sets of dependencies we have to remember to update so after doing this, I think we should also rework the scripts in `tools` to use these new files as well. https://github.com/certbot/certbot/blob/master/tools/merge_requirements.py could be used to merge this into a single file and handle shared dependencies.

sydneyli:
Fixes #6581 

Created by `pip freeze`ing each plugin and removing the requirements that were
already in `letsencrypt-auto-source/pieces/dependency-requirements.txt`

Let me know if this is what you expected, @bmw.
Here's the janky script I used to generate the Dockerfiles:
```
# extract package names
cat letsencrypt-auto-source/pieces/dependency-requirements.txt | sed '/^#/d' | sed '/^$/d' \
    | sed '/^\s*--/d' | cut -d "=" -f 1 > common_dep_names.txt
for dir in certbot-dns-*
do
    python ./tools/_venv_common.py -e acme -e . -e $dir
    source venv/bin/activate
    cd $dir
    # pip freeze all the deps that aren't in common-dep_names and aren't "-e" directories
    pip freeze | grep -vwF -f ../common_dep_names.txt \
        | sed '/^-e git+/d' > plugin_dependencies.txt
    cat > Dockerfile <<EOF
FROM certbot/certbot

COPY . src/$dir

RUN pip install -r src/$dir/plugin_dependencies.txt \\
    && pip install --no-cache-dir --editable src/$dir
EOF
    cd .. && deactivate && rm -rf venv
done
rm common_dep_names.txt
```

having been changed.


sydneyli:
Ah, I missed the part on consolidating these dependencies with the pins used for dev/testing.

the only reason i can think for keeping it this way is keeping Docker image size as minimal as possible (iirc boto3 is pretty massive compared to all the other requirements)-- but i'm not sure this is enough of a draw to untangle the mess of resolving conflicting pinned dependency versions during testing.

Your third idea seems simplest to me. let's just freeze `dev_constraints.txt` with the union of all plugin requirements, and rename it to something that aligns more with how the file is now being used. As for resolving conflicts between `dev_constraints.txt` and the le-auto pins, can we reuse `tools/merge_requirements`?

bmw:
> the only reason i can think for keeping it this way is keeping Docker image size as minimal as possible (iirc boto3 is pretty massive compared to all the other requirements)-- but i'm not sure this is enough of a draw to untangle the mess of resolving conflicting pinned dependency versions during testing.

I think we can work around this problem by using the `--constraint` flag with `pip install` rather than `--requirement`. See https://pip.pypa.io/en/stable/user_guide/#constraints-files for more info about this.

> As for resolving conflicts between dev_constraints.txt and the le-auto pins, can we reuse tools/merge_requirements?

We can! The one additional problem I can think of with this is the le-auto pins contain hashes which turns on pip's hash verification and blocks installing packages without a hash available such as our plugins which haven't been built into a single package file.

We work around this problem in [tools/pip_install.py](https://github.com/certbot/certbot/blob/b58d0e722d4d3b54ede48a67aeb25d4fa6d6fc45/tools/pip_install.py#L51) by stripping out the hashes before merging the two files. Perhaps we can factor this code out of this script so we can use it here as well?

tools/docker_constraints.py
```diff
@@ -0,0 +1,23 @@
...
+
+import sys
+
+import pip_install
+import merge_requirements
+
+def main(hashed_reqs_in, unhashed_reqs_out,
+         constraints_in, constraints_out):
+    with open(unhashed_reqs_out, 'w') as fd:
+        fd.write(pip_install.remove_requirements_hashes(hashed_reqs_in))
```
bmw: nit: I'm a little worried about us breaking this without realizing it.

I believe none of our tests in Travis currently run this file and I don't think we're really maintaining a stable API for our various scripts in tools so I'm worried someone might refactor `pip_install.py`, not realizing that this file was using `remove_requirements_hashes` and break everything.

I wonder if we might be able to avoid this by making a simple script whose sole purpose was to strip hashes. Files who needed that functionality could then call into the main function or shell out to the entire script.

What do you think?

Dockerfile
```diff
@@ -22,8 +29,9 @@ RUN apk add --no-cache --virtual .build-deps \
         openssl-dev \
         musl-dev \
         libffi-dev \
-    && pip install -r /opt/certbot/dependency-requirements.txt \
     && pip install --no-cache-dir \
+        --requirement /opt/certbot/unhashed_requirements.txt \
```
bmw: I don't think we should switch to the unhashed requirements here. While it probably doesn't matter much, the hashed requirements is an additional security check that the packages we obtained over HTTPS are the ones we expect. Since `certbot-auto` uses these requirements, all needed packages should be specified in `letsencrypt-auto-source/pieces/dependency-requirements.txt` and we shouldn't be installing anything that is (only) in `dev_constraints.txt`.

What if we:

1. Kept the `pip install -r /opt/certbot/dependency-requirements.txt` line.
2. Changed the command installing `acme` and `certbot` to include `--no-deps`.

The inclusion of `--no-deps` may cause things to break in the future, but it's a sanity check that all dependencies were properly installed from the hashed requirements and the second `pip` invocation wasn't secretly installing other packages.

Thoughts?

tools/dev_constraints.txt
```diff
@@ -1,4 +1,4 @@
-# Specifies Python package versions for development.
+# Specifies Python package versions for development and building Docker images.
```
bmw: Thanks for adding this comment! I probably would have forgot about it.

tools/strip_hashes.py
```diff
@@ -0,0 +1,16 @@
...
+def main(args):
+    out_lines = []
+    for line in args:
+        search = re.search(r'^(\S*==\S*).*$', line)
+        if search:
+            out_lines.append(search.group(1))
+    return os.linesep.join(out_lines)
+
+if __name__ == '__main__':
+    print(main(sys.argv[1:]))
```
bmw: This is requiring the user provide the file as CLI arguments to this script:
```
tools/strip_hashes.py argparse==1.4.0 --hash=sha256:c31647edb69fd3d465a847ea3157d37bed1f95f19760b11a47aa91c04b666314 ...
```
I think this is a little awkward and doesn't work with the `RUN sh -c 'cat dependency-requirements.txt | /opt/certbot/tools/strip_hashes.py > unhashed_requirements.txt'` line in the Dockerfile.

Given how this script is being used, maybe it should just take the filename to read from as its first and only CLI argument?

With that said, if you also want to give it the option to read from stdin, I don't think the unix fans of the world would complain.

Dockerfile
```diff
@@ -8,11 +8,10 @@ WORKDIR /opt/certbot
...
 
 # Generate constraints file to pin dependency versions
-COPY letsencrypt-auto-source/pieces/dependency-requirements.txt hashed_requirements.txt
+COPY letsencrypt-auto-source/pieces/dependency-requirements.txt .
 COPY tools /opt/certbot/tools
-RUN /opt/certbot/tools/docker_constraints.py \
-    hashed_requirements.txt unhashed_requirements.txt \
-    tools/dev_constraints.txt docker_constraints.txt
+RUN sh -c 'cat dependency-requirements.txt | /opt/certbot/tools/strip_hashes.py > unhashed_requirements.txt'
+RUN sh -c 'cat tools/dev_constraints.txt unhashed_requirements.txt | /opt/certbot/tools/merge_requirements.py > docker_constraints.txt'
```
bmw: I don't think this works with `merge_requirements.py`. It currently takes a list of filenames as CLI arguments.

bmw:
Joona said he could take this over.

@joohoi, assuming you're happy with the overall approach here, in my opinion you should make a new branch based on this and squash Sydney's commits down to a single commit. Before we merge your PRs, I think we should do the same with any commits you add so we have a single commit from Sydney followed by a single commit from you, but I don't care whether we squash your commits initially or just before we merge.

adferrand:
Sorry, my browser went full berserk...

joohoi:
I'm closing this in favor of #6786

PR #6683, opened 2019-01-22
Forcibly reactivate tls-sni-01 challenges until complete removal.

adferrand:
This PR reactivate tls-sni-01 challenges on recent Boulder versions checkout for integration tests. This allows to continue testing this challenge until it is officially dropped from server (Boulder) and client (Certbot).


adferrand:
No its OK I think, it just needs to be set until TLS-SNI-01 is removed from Certbot. So we do not need a more complex long-term solution here.

Issue #6676, opened 2019-01-18
acme module: support client.answer_challenge for TLS-ALPN-01


cpu:
After https://github.com/certbot/certbot/pull/6144 landed Boulder had to [switch away from using an up to date Certbot clone for our integration tests](https://github.com/letsencrypt/boulder/blob/ed01d6bc14307ca5e361417a6b68dfac94db8c6e/test/boulder-tools/build.sh#L54-L62). 

We don't rely on Certbot's `acme` module to create or serve TLS-ALPN-01 challenge response certificates (we use [`pebble-challtestsrv`](https://github.com/letsencrypt/pebble/tree/master/cmd/pebble-challtestsrv) for this), but we do rely on the `acme` module for initiating challenges. E.g. in [`do_tls_alpn_challenges`](https://github.com/letsencrypt/boulder/blob/ed01d6bc14307ca5e361417a6b68dfac94db8c6e/test/chisel.py#L191-L202) in our `chisel.py` test client we call:
```
 client.answer_challenge(c, c.response(client.key))
```
where `c` is a `acme.ChallengeBody` with type `TLSALPN01`.

If we remove this commit pin then `answer_challenge` begins to fail:
```
Traceback (most recent call last):
  File "test/integration-test.py", line 888, in <module>
    main()
  File "test/integration-test.py", line 821, in main
    run_chisel(args.test_case_filter)
  File "test/integration-test.py", line 845, in run_chisel
    value()
  File "test/integration-test.py", line 328, in test_tls_alpn_challenge
    auth_and_issue(domains, chall_type="tls-alpn-01")
  File "/go/src/github.com/letsencrypt/boulder/test/chisel.py", line 224, in auth_and_issue
    cleanup = do_tlsalpn_challenges(client, authzs)
  File "/go/src/github.com/letsencrypt/boulder/test/chisel.py", line 198, in do_tlsalpn_challenges
    client.answer_challenge(c, c.response(client.key))
  File "/certbot/acme/acme/challenges.py", line 179, in response
    key_authorization=self.key_authorization(account_key))
TypeError: 'NotImplementedType' object is not callable
```

We could potentially work around this by figuring out the right way to use the `acme` module to prepare a signed JWS body using the account key that we can POST to the challenge URI ourselves but I think that's a little yucky: we'd be duplicating a lot of the `acme` module's existing code. It also means other projects that use the `acme` module may have to do the same in the future.

[...]

PR #6689, opened 2019-01-23
acme: add TLSALPN01Response for initiating tls-alpn-01.

cpu:
The existing `acme.TLSALPN01` challenge class did not have a `response_cls`, meaning it was not possible to use a tls-alpn-01 challenge with `client.answer_challenge`.

To support the above a simple `TLSALPN01Response` class is added and used as the `response_cls`. This response class doesn't provide the ability to solve a tls-alpn-01 challenge end to end (e.g. generating and serving the correct response certificate) but it does allow the challenge to be initiated. This is sufficient for users that have set up the challenge response independent of the `acme` module code. I tested this successfully with a hacked up version of Boulder's `chisel2.py` client and a `pebble` instance.

Resolves https://github.com/certbot/certbot/issues/6676

### DIFF
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -513,6 +513,27 @@ class TLSSNI01(KeyAuthorizationChallenge):
         return self.response(account_key).gen_cert(key=kwargs.get('cert_key'))
 
 
+@ChallengeResponse.register
+class TLSALPN01Response(KeyAuthorizationChallengeResponse):
+    """ACME TLS-ALPN-01 challenge response.
+
+    This class only allows initiating a TLS-ALPN-01 challenge returned from the
+    CA. Full support for responding to TLS-ALPN-01 challenges by generating and
+    serving the expected response certificate is not currently provided.
+    """
+    typ = "tls-alpn-01"
+
+    def simple_verify(self, chall, domain, account_public_key):
+        """simple_verify for TLS-ALPN-01 challenges passes through to the base
+        KeyAuthorizationChallengeResponse verify implementation.
+        """
+        # pylint: disable=unused-argument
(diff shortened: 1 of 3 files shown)


acme/acme/challenges.py
```diff
@@ -513,6 +513,27 @@ def validation(self, account_key, **kwargs):
...
+class TLSALPN01Response(KeyAuthorizationChallengeResponse):
+    """ACME TLS-ALPN-01 challenge response.
+
+    This class only allows initiating a TLS-ALPN-01 challenge returned from the
+    CA. Full support for responding to TLS-ALPN-01 challenges by generating and
+    serving the expected response certificate is not currently provided.
+    """
+    typ = "tls-alpn-01"
+
+    def simple_verify(self, chall, domain, account_public_key):
```
bmw: Unless you need or want this method on your side, I'd prefer we removed it and the tests for it. The only thing in the class definition would be the docstring and `typ`.

(If you're interested, the reason why I prefer this is `simple_verify` in our challenges has not worked well for us. The initial idea was we could do some sanity checking of challenges on our end before reaching out to the CA but there were a number of scenarios like weird firewall rules that would cause `simple_verify` to fail but  validation by the CA to succeed. For this reason, Certbot no longer uses it and we recommend other clients don't either. We've kept the methods around in some form of another to prevent API breakage, but this is a new challenge type so I'd prefer to skip on adding the method entirely until we have a need to do otherwise.)

cpu: > Unless you need or want this method on your side, I'd prefer we removed it and the tests for it. The only thing in the class definition would be the docstring and typ.

WFM! Removed in 8b011a4. Thanks @bmw

Issue #6685, opened 2019-01-23
Pin pip version during virtualenv creation through certbot-auto


adferrand:
Issue #6682 enlighted a breakage on certbot-auto since version 19.0 of pip has been released that contains a bug. Issue on upstream as been filled here pypa/pip#6158 and bug fix version is incoming.

What I want to study here, is how to avoid that this kind of breakage breaks certbot-auto. It impacted us because by default, virtualenv will try upon its creation to update its fundamental components like pip or setuptools to the latest version available on PyPI. So a temporary fix on our side is to set the `--no-download` to virtualenv to disable the automated update, and use the modules packages in the OS distribution instead.

I think this is not a complete solution, as it makes certbot-auto be sticked with potentially very old modules, that would gain some benefit to be updated nonetheless. For pip at least, we are covered to avoid too old pip versions, but if its logic is triggered, it will still get the latest pip version, so we are again on the same situation.

In fact, we use pinned versions for every other certbot dependencies exactly for this purpose, avoid to have unexpected breakage in recent releases in some modules. 

I propose to do the same for pip and pin its version. This could be done through the pipstrap process.

What do you think?

cc @bmw @ohemorange

adferrand:
It seems that there is two independent problems, related to the wheel module of pip:
* Failing assertion `building_is_possible` #6682
* Failing parsing on files with comma #6687

@cailen, you confirm that neither issues are corrected by pip 19.0.1 ?

bmw:
I agree and I think doing this should be as simple as removing the branch https://github.com/certbot/certbot/blob/1dd7db12e05c7a52f1b1092d76a14caf0bf4c4cc/letsencrypt-auto-source/pieces/pipstrap.py#L152

Erik originally offered to help here, however, because it looks like we may need to put out a bug fix release with this change, I personally think someone from the core Certbot team should probably take this to make sure it can be done quickly.

I also think we should run our test farm tests before landing changes here to make sure the pinned version of pip and friends is working well on various distros.

@ohemorange, are you interested in taking this?

ohemorange:
Reposting here: #6682 is addressed by 19.0.1, but certbot-auto still hits #6687. Currently looking into getting that addressed.

ohemorange:
@bmw yes

ohemorange:
My plan here is to try three separate things, and put them up as three separate PRs so we can decide separately about them.

1. add `--no-download`. this will for sure fix everything right now for most users.
[...]

PR #6690, opened 2019-01-24
Add VIRTUALENV_NO_DOWNLOAD=1 to all calls to virtualenv

ohemorange:
This will immediately address the breakage reported in #6682 and tracked at #6685. Virtualenv downloads the latest pip, which causes issues, so tell virtualenv to not download the latest pip.

I added the flag preemptively to other files as well, they're in separate commits so it will be easy to revert any spots we don't want.

I've confirmed that this fixes the issue on a machine that fails with the version of certbot-auto currently in master: recent version of virtualenv, python 2.7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 * Fixed accessing josepy contents through acme.jose when the full acme.jose
   path is used.
 * Clarify behavior for deleting certs as part of revocation.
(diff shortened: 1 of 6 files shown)


bmw:
When I run this on Debian Jessie, I get:
```
Usage: virtualenv [OPTIONS] DEST_DIR

virtualenv: error: no such option: --no-download
```

ohemorange:
Ah right, forgot about that. Updating.

PR #6691, opened 2019-01-24
Always download the pinned version of pip in pipstrap

ohemorange:
This will immediately address the breakage reported in #6682 and tracked at #6685. Virtualenv downloads the latest pip, which causes issues, so after virtualenv upgrades pip, downgrade to the pinned version.

I've confirmed that this fixes the issue on a machine that fails with the version of certbot-auto currently in master: recent version of virtualenv, python 2.7.


letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -149,8 +149,6 @@ def main():
     pip_version = StrictVersion(check_output(['pip', '--version'])
                                 .decode('utf-8').split()[1])
     min_pip_version = StrictVersion(PIP_VERSION)
```
erikrose: Might as well delete this line too, with its now-unread variable.

bmw:
I'm fine either just merging this as is or deleting the unused variable first. Up to you @ohemorange.

ohemorange:
give me one minute, I'll fix it

PR #6694, opened 2019-01-24
Prep 0.30.1

bmw:
Cherry picks the changes for the 0.30.1 release.

I did this all in one PR to avoid repeated conflicts in the CHANGELOG.

I think I'd personally prefer if we merged this PR so the commits more or less match those on master, but squashing should be fine too.


bmw:
Boulder's master branch removed support for the TLS-SNI-01 challenge by default. Certbot's current master branch has tests for TLS-SNI-01 disabled.

For the purposes of testing the point release, I modified `boulder-fetch.sh` to check out the most recent commit to Boulder that still has support for TLS-SNI-01 enabled by default.

Issue #6698, opened 2019-01-25
certbot-auto doesn't work on Fedora 29


bmw:
## My operating system is (include version):

Fedora 29

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

certbot-auto

## I ran this command and it produced this output:

```
# git checkout v0.30.0
# ./certbot-auto -n
<bunch of stuff>
Creating virtual environment...
./certbot-auto: line 1003: virtualenv: command not found
```

## Certbot's behavior differed from what I expected because:

`virtualenv` should probably be found.

bmw:
Kind of? `certbot-auto` did used to work on Fedora and we're still passing tests on an old version in our test farm tests. I think what probably happened here is a more recent Fedora version changed the name of the virtualenv package and we haven't yet updated certbot-auto to reflect that.

I think we should fix this and I think doing so is probably pretty easy, however, I don't think it's a super high priority because we have very few Fedora users. As far as I know, no users have reported this bug yet.

ohemorange:
I mean, is this a regression based on our recent changes in response to the `pip` issues.

bmw:
Oh. No it is not.

ohemorange:
> /usr/bin/virtualenv is now in the python3-virtualenv package
> The virtualenv command now comes from the python3-virtualenv package, as opposed to earlier releases where the command was in the python2-virtualenv. This effectively switches the command to Python 3; if you run virtualenv without any additional options, it will create Python 3 environments. Use virtualenv -p python2.7 to get the previously default behavior.

https://docs.fedoraproject.org/en-US/fedora/f29/release-notes/developers/Development_Python/

bmw:
Nice find!

If all non-EOL'd Fedora versions made that change, I think this fix would be trivial. Unfortunately, it looks like Fedora 28 doesn't have a definitive deprecation date yet and looking at the pattern in https://en.wikipedia.org/wiki/Fedora_version_history, it will probably be at least a few months.

With that in mind, I think we need to modify certbot-auto to work with both package names.

@ohemorange, do you have the interest/cycles to work on this?

ohemorange:
I don't think I can get to it this sprint, but happy to do it in a future one.

PR #6700, opened 2019-01-25
Update test farm targets

### LINKED ISSUE #6106: Test against newer systems in the test farm tests
The list at `tests/letstest/targets.yaml` is very old and hasn't been updated in a long time.

bmw:
Fixes #6106.

AMIs were taken from https://wiki.debian.org/Cloud/AmazonEC2Image/Stretch and https://cloud-images.ubuntu.com/locator/ec2/.

I didn't update the AMI for Fedora due to #6698.

These new AMIs pass on all test farm tests we run during the release process except Ubuntu 18.04 and 18.10 fail on `test_apache2.sh`. This is tracked at #6706. If this PR lands before this issue is resolved, we should list these systems as expected failures in the release notes.

Adding these AMIs slows down our tests significantly. I didn't measure it, but it feels 50-100% slower at least on my setup. I think it's worth it though.

### DIFF
--- a/tests/letstest/targets.yaml
+++ b/tests/letstest/targets.yaml
@@ -11,7 +11,7 @@ targets:
     type: ubuntu
     virt: hvm
     user: ubuntu
-  - ami: ami-70dad51a
+  - ami: ami-09677e0a6b14905b0
     name: ubuntu16.04LTS
     type: ubuntu
     virt: hvm
--- a/tests/letstest/targets.yaml
+++ b/tests/letstest/targets.yaml
@@ -1,6 +1,21 @@
 targets:
   #-----------------------------------------------------------------------------
   #Ubuntu
+  - ami: ami-064bd2d44a1d6c097
+    name: ubuntu18.10
+    type: ubuntu
+    virt: hvm
+    user: ubuntu
+  - ami: ami-012fd5eb46f56731f
+    name: ubuntu18.04LTS
+    type: ubuntu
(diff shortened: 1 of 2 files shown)


tests/letstest/targets.yaml
```diff
@@ -1,6 +1,21 @@
 targets:
   #-----------------------------------------------------------------------------
   #Ubuntu
+  - ami: ami-064bd2d44a1d6c097
```
ohemorange: Why did you choose the us-east targets? Why not west? Do the test farm tests always run on east? Is that something users set up manually?

bmw: Yes, the test farm tests currently always run on US east. Which region you use is something you configure when you set up your AWS credentials with `aws configure`. Our instructions in the release notes for setting up these tests have you choose us-east-1 because that's where all AMIs in this file are located.

ohemorange:
Also let's add the expected failures to the release process notes.

bmw:
> We should probably replace 18.10 with 19.04 when it comes out, don't think we need both.

Because Ubuntu is so popular, I'd personally prefer we test on both. I wouldn't think an additional test instance would affect speed too much and it helps ensure we catch bugs before they're released.

I created #6720 to track that issue. Happy to talk more about whether or not we should keep 18.10 and 19.04 tests around at the same time whenever.

> Also let's add the expected failures to the release process notes.

Done!

PR #6702, opened 2019-01-25
 Add FreeBSD specific paths

sunpoet:
This is originally [a local patch for FreeBSD ports](https://svnweb.freebsd.org/ports/head/security/py-certbot/files/patch-certbot_compat.py).


schoen:
@adferrand, could you please check whether this is an AppVeyor problem or whether this PR really does decrease test coverage? Thanks!

adferrand:
@schoen, yes this PR really decreases the code coverage for Windows, and for Linux by the way. But for Windows, we pass under a unit of the percentage, from 94 to 93.

In the long term, it is inevitable that the threshold for code coverage in both Linux and Windows will need to be lowered, as more platform specific code  that cannot be tested in the other platform will be added.

Here however, it can be fixed without touching threshold. There is a test, `certbot.tests.cli_test:FlagDefaultTest:test_linux_directories` that test default directories correctness. It executes exclusively for Linux conventions. With appropriate mocking, it could test for Linux, Windows and FreeBSD independently from the running platform.

adferrand:
Hello @sunpoet, are you willing to fix the merge conflict on your PR? It will consists essentially to move the logic you added from `certbot.compat` to `certbot.compat.misc`, where the default folders methods have been relocated. 

We improved the code coverage about windows and Linux, so the PR may have a build succeed out of the box.

sunpoet:
Hi @adferrand, the patch is updated and the build failure on Travis CI seems unrelated.

certbot/compat/misc.py
```diff
@@ -142,8 +146,13 @@ def get_default_folder(folder_type):
 
     """
     if os.name != 'nt':
-        # Linux specific
-        return LINUX_DEFAULT_FOLDERS[folder_type]
+        # Unix-like
+        if sys.platform.startswith('freebsd') or sys.platform.startswith('dragonfly'):
+            # FreeBSD specific
+            return FREEBSD_DEFAULT_FOLDERS[folder_type]
+        else:
```
adferrand: The failure in the build is due to a linting error here. The `else` is not necessary, since we return in L152 if condition on L150 is `True`. So L154 and L155 can be executed only if the condition is `False`.

sunpoet: Should be fixed in d0a50ef92f6085f14a36c8fc076f2caa7af711c4.

PR #6709, opened 2019-01-25
Candidate 0.30.2 point release

ohemorange:
This PR should be merged and not squashed to preserve the release commit.


bmw:
I noticed this after I merged this and it probably doesn't matter, but why isn't the "Release 0.30.2" commit the [signed and tagged release commit](https://github.com/certbot/certbot/pull/6707/commits/6cba691c1954137e1baaef3c901a6ce5d9b6b2ee) that was used in #6707?

PR #6711, opened 2019-01-27
[Windows] Fix account paths on Windows when colons are involved

adferrand:
The account path used to store user credentials is calculated from the domain used to contact the relevant ACME CA server.

For instance, if the directory URL is `https://my.domain.net/directory`, then the account path will be `$CONFIG_DIR/accounts/my.domain.net`.

However, if non standard HTTP/HTTPS port need to be used, colons will be involved. For instance, `https://my.domain.net:14000/directory` will give `$CONFIG_DIR/accounts/my.domain.net:14000`.

Colons in paths are supported on POSIX systems, but not on Windows (it is reserved for the root drive letter).

This PR replaces colons by underscores for account paths on Windows, and leaves them untouched on Linux.


certbot/account.py
```diff
@@ -141,8 +141,9 @@ class AccountFileStorage(interfaces.AccountStorage):
     """
     def __init__(self, config):
         self.config = config
-        util.make_or_verify_dir(config.accounts_dir, 0o700, compat.os_geteuid(),
-                                   self.config.strict_permissions)
+        account_dir = compat.normalize_path(config.accounts_dir)
```
bmw: With the change in `configuration.py`, why do we need this?

certbot/compat.py
```diff
@@ -199,3 +199,19 @@ def get_default_folder(folder_type):
         return LINUX_DEFAULT_FOLDERS[folder_type]
     # Windows specific
     return WINDOWS_DEFAULT_FOLDERS[folder_type]
+
+
+def normalize_path(path):
```
bmw: I'm a little on the fence about adding this function here. Do we expect to have other paths that are incompatible with Windows that need fixing up and do we expect that we'll want to make the same decision about how to modify them everywhere?

Unless we're confident about both, I think I'd personally prefer to either put this code in the same file that's modifying the account path or rename this function to specifically mention accounts paths. I think I prefer the former for now as we can always move code into `compat` if we have multiple uses for it later.

What do you think?

certbot/compat.py
```diff
@@ -199,3 +199,19 @@ def get_default_folder(folder_type):
...
+    # type: (str) -> str
+    """
+    Replace unsupported characters in path for current OS by underscores.
+    :param path: the path to normalize
+    :return: the normalized path
+    """
+    if 'fcntl' in sys.modules:
```
bmw: Can we use `os.name` here?

certbot/compat.py
```diff
@@ -199,3 +199,19 @@ def get_default_folder(folder_type):
         return LINUX_DEFAULT_FOLDERS[folder_type]
     # Windows specific
     return WINDOWS_DEFAULT_FOLDERS[folder_type]
+
+
+def normalize_path(path):
+    # type: (str) -> str
+    """
+    Replace unsupported characters in path for current OS by underscores.
+    :param path: the path to normalize
```
bmw: nit: Can we change this to `:param str path: ...`

certbot/compat.py
```diff
@@ -199,3 +199,19 @@ def get_default_folder(folder_type):
...
+    :return: the normalized path
```
bmw: Can we specify the return type here as well by using `rtype`?

Specifying these types in multiple places is redundant, but it makes sure the type is added to our API docs.

certbot/compat.py
adferrand: For this code specifically, yes you are right, the method could be replaced by an inline logic, and I am not expecting to correct paths somewhere else.

But in fact my motivation to put the logic as a separate function follows a design convention that I tried to respect so far: for better maintainancy, I put every platform specific code in the compat module, to let only platform agnostic code everywhere else (except tests, to avoid to clutter the module with non-functional logic).

In fact with the windows permission PR, this compat module will become a package, to improve modularization of platform specific code. And I am also thinking to move the lockfile module in this package when the relevant PR will be integrated and compat package module is available.

Even if it adds some lines in particular implementations, I think it is really important to keep this platform specific code in one place in the long term.

adferrand: I think however that the function name can be changed to something more specific about replacing platform specific unsupported characters in paths.

bmw: That sounds good to me. I was initially not worried about moving code like this out of `compat` because we're already doing that in the locking PR, but if you're planning to move that code back in later, that makes sense to me and I agree a more explicit name here is good enough.

certbot/compat.py
adferrand: Done. I took the occasion to refactor a little `compat` module to be more consistent.

PR #6713, opened 2019-01-28
No numprocesses in pytest.ini

bmw:
Setting --numprocesses doesn't work with `pdb`.

The issue is dated and closed, but here's some documentation for that: https://github.com/pytest-dev/pytest/issues/390.

Removing `--numprocesses` from `pytest.ini` allows devs to invoke `pytest` and use their debugger normally, but you still get the speed improvements for using multiple processes when you invoke `tox`. We could instead only set `--numprocesses` in CI, but I think more devs will appreciate the speed improvements in `tox` than being able to invoke a debugger.

cc @joohoi

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,7 +2,10 @@
 # settings we want to also change there must be added to the release script
 # directly.
 [pytest]
-addopts = --numprocesses auto --pyargs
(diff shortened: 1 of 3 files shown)


bmw:
Closing in favor of #6715.

PR #6717, opened 2019-01-29
(WIP) Boost pytest execution

adferrand:
Work in progress to see if more pytest nodes boost the tests execution.


adferrand:
The test is not conclusive.

PR #6732, opened 2019-02-02
/var/logs/ -> /var/log/

Shifterovich:


adferrand:
Nice catch! Thanks a lot!

PR #6741, opened 2019-02-06
Test all on push events to tested non-master branches.

bmw:
We always run a full set of CI tests before beginning the release process. The way this would work previously is we would either trigger tests on the `test-everything` branch to run through Travis' web UI or if it was a point release, create a new branch based on `test-everything` but modify `.travis.yml` so the branch that was pulled in to be tested was the point release branch instead of `master`.

This no longer works because the former `test-everything` tests are now only run when Travis automatically runs our tests nightly.

We could create and maintain a separate branch for the purpose of manually running all tests or remove the conditionals from the latest `.travis.yml` file every time before we want to run these tests, but there must be A Better Way™.

This PR makes the change that in addition to running all tests nightly, they would also run on pushes to tested branches other than master. These changes do not affect the tests run on PRs or on commits to `master`.

What is affected is commits to point release branches and branches named `test-*`. (See [.travis.yml](https://github.com/certbot/certbot/blob/2ddaf3db043ea8526ae3f9ab2ef120b194b2e506/.travis.yml#L177) for what branches we run tests on.) Running all tests on point release branches automates the step of running our full test suite before doing a point release.

The changes to `test-*` could be a mixed bag, however, since we switched to travis-ci.com over 3 weeks ago, I'm the only one who has used this functionality and I personally prefer things this way. At the very least, since these branches don't seem to be widely used, I think we can make this change and reevaluate if it becomes a problem.

If this PR lands, the release instructions will need to be updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,85 +62,86 @@ matrix:
     - python: "2.7"
       env: TOXENV=nginxroundtrip
 
-    # These environments are executed on cron events only
+    # These environments are executed on cron events and pushes to tested
+    # branches other than master.
     - python: "3.7"
       dist: xenial
       env: TOXENV=py37 CERTBOT_NO_PIN=1
-      if: type = cron
+      if: type = cron OR (type = push AND branch != master)
     - python: "2.7"
       env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=certbot TOXENV=py27-certbot-oldest
       sudo: required
       services: docker
-      if: type = cron
+      if: type = cron OR (type = push AND branch != master)
     - python: "2.7"
       env: BOULDER_INTEGRATION=v2 INTEGRATION_TEST=certbot TOXENV=py27-certbot-oldest
       sudo: required
       services: docker
-      if: type = cron
+      if: type = cron OR (type = push AND branch != master)
     - python: "2.7"
       env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=nginx TOXENV=py27-nginx-oldest
       sudo: required
       services: docker
-      if: type = cron
+      if: type = cron OR (type = push AND branch != master)
     - python: "2.7"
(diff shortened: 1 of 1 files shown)


bmw:
@adferrand or @ohemorange, are you able to review this?

ohemorange:
What's the mechanism for distinguishing these pushes from the tests run on travis for any open PR? How does it know that `push` only applies to the branches specified in `.travis.yml`?

adferrand:
In fact it the association of `branches.only` + the condition `branch != master` that ensure that theses extra tests are executed only on push for `test-*` branches and release version branches.

But it is not obvious from `.travis-ci.yml` reading, in particular as the `branches.only` assertion come after the definition of all execution environments.

I think `branches.only` should be moved on top of the file, and a comment explaining how the two parameters interact to do what we want, should be written on the first extra tests environment encountered in `.travis-ci.yml` during a reading.

bmw:
> What's the mechanism for distinguishing these pushes from the tests run on travis for any open PR?

I'm not sure exactly how this works, but I think that's handled by the information GH sends to Travis when an event happens.

Regardless of how Travis offers this functionality, pull requests are a different event type (see https://docs.travis-ci.com/user/conditions-v1) and Travis runs different tests in these two cases. If you look at the status of the checks for this (or any) PR, you'll see "Travis CI - Pull Request" but if you look at the status checks of https://github.com/certbot/certbot/pull/6740, you'll also see "Travis CI - Branch" which was run because the branch was named "test-conditions".

> I think `branches.only` should be moved on top of the file, and a comment explaining how the two parameters interact to do what we want, should be written on the first extra tests environment encountered in `.travis-ci.yml` during a reading.

Done.

adferrand:
The only other concern would be unexpectedly full travis build on `test-*` push or PR, but as you said @bmw, it is exclusively used by you for now, without any conflicting use cases.

bmw:
I've updated the release instructions describing how this works.

PR #6751, opened 2019-02-07
Fix cryptography OCSP support

adferrand:
This PR, following #6603, is fixing a divergence in how OCSPResponse is checked in OpenSSL versus our implementation. This divergence led some tests to fail, and implied to revert the integration of #6603 through #6747.

The hypothesis is that nextUpdate/thisUpdate are not tested the same way than in OpenSSL. After checking the source code, I indeed catched some differences. This PR removed these differences to use strictly the same behavior than OpenSSL on that matter.

Link to the relevant source code of OpenSSL is given directly as an inline comment in the `ocsp` module, here: https://github.com/adferrand/certbot/blob/5e4eaf2841f66f71eb573e14466d779bd0c8ffac/certbot/ocsp.py#L189-L197


adferrand:
Note that on the contrary to OpenSSL that consider only nextUpdate/thisUpdate inconsistencies as warning, Certbot consider them as errors and ignore the OCSP response in this case. This PR do the same for cryptography: it aligns the tests to what is done in OpenSSL, but still consider nextUpdate/thisUpdate as failures.

Next step in this PR is to implement requirements done in #6760.

adferrand:
Output when connection to the OCSP server is not meaningful:
* `TEST_CERT` display depends only on which URL is registered in the certificate
* `EXPIRED` is calculated on the expiration date in the certificate
* `REVOKED` will never be displayed in this case, as the default behavior is to consider the certificate not as expired if a valid OCSP respond could not be retrieved.

However, OCSP implementation on cryptography was not protected against a connection error when trying to contact the OCSP server. It is now.

Tests have been migrated to Boulder integration tests.

PR is ready to be review @bmw.

tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh
```diff
@@ -1,44 +0,0 @@
...
-
-cd letsencrypt
-export PATH="$PWD/letsencrypt-auto-source:$PATH"
-letsencrypt-auto --os-packages-only --debug --version
-letsencrypt-auto certonly --no-self-upgrade -v --standalone --debug \
-                   --text --agree-dev-preview --agree-tos \
-                   --renew-by-default --redirect \
-                   --register-unsafely-without-email \
-                   --domain $PUBLIC_HOSTNAME --server $BOULDER_URL
-
```
bmw: I do not think we should delete this whole file. Just from here down.

tests/letstest/scripts/test_letsencrypt_auto_venv_only.sh
```diff
@@ -5,3 +5,8 @@
 cd letsencrypt
 # help installs virtualenv and does nothing else
 ./letsencrypt-auto-source/letsencrypt-auto -v --debug --help all
+
+if ! ./letsencrypt-auto-source/letsencrypt-auto --help --no-self-upgrade | grep -F "letsencrypt-auto [SUBCOMMAND]"; then
```
bmw: Why are we adding this, especially in this PR?

certbot/ocsp.py
```diff
@@ -67,6 +67,7 @@ def ocsp_revoked(self, cert_path, chain_path):
         if not host or not url:
             return False
 
+        logger.error(url)
```
bmw: Why add this?

certbot/ocsp.py
```diff
@@ -137,11 +138,16 @@ def _check_ocsp_cryptography(cert_path, chain_path, url):
...
     request = builder.build()
     request_binary = request.public_bytes(serialization.Encoding.DER)
-    response = requests.post(url, data=request_binary,
-                             headers={'Content-Type': 'application/ocsp-request'})
-    if response.status_code != 200:
+    try:
+        response = requests.post(url, data=request_binary,
+                                 headers={'Content-Type': 'application/ocsp-request'})
+    except requests.exceptions.RequestException:
         logger.info("OCSP check failed for %s (are we offline?)", cert_path)
```
bmw: nit: Can we use `logger.debug(..., exc_info=True)` to log the exception that occurred here?

tests/certbot-boulder-integration.sh
```diff
@@ -527,3 +529,62 @@ if [ "${BOULDER_INTEGRATION:-v1}" = "v2" ]; then
 fi
 
 coverage report --fail-under 64 --include 'certbot/*' --show-missing
+
+# Test OCSP status
+
+## OCSP 1: Check staled certificate OCSP status
```
bmw:
```suggestion
## OCSP 1: Check stale OCSP status
```

tests/certbot-boulder-integration.sh
```diff
@@ -527,3 +529,62 @@ if [ "${BOULDER_INTEGRATION:-v1}" = "v2" ]; then
...
+    echo "Did not find two test certs as expected ($TEST_CERTS)"
+    exit 1
+fi
+
+if [ "$EXPIRED" != 2 ] ; then
+    echo "Did not find two test certs as expected ($EXPIRED)"
+    exit 1
+fi
+
+if [ "$REVOKED" != 0 ] ; then
```
bmw: I think this will fail in our nightly tests where we run integration tests with our oldest dependencies. When doing that, we will be shelling out to openssl and one of these certificates will be listed as revoked.

I personally think this minor inconsistency is fine and we should remove this check.

certbot/ocsp.py
adferrand: Debug trace. I remove it.

certbot/ocsp.py
adferrand: Done.

tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh
adferrand: I added again the test script.

tests/letstest/scripts/test_letsencrypt_auto_venv_only.sh
adferrand: See above.

tests/certbot-boulder-integration.sh
adferrand: Indeed. We test the revoke on a live certificate just after. I remove that.

tests/certbot-boulder-integration.sh
```diff
@@ -527,3 +529,57 @@ if [ "${BOULDER_INTEGRATION:-v1}" = "v2" ]; then
...
+
+# Test OCSP status
+
+## OCSP 1: Check stale OCSP status
+pushd ./tests/letstest/testdata
+
+OUT=`common certificates --config-dir sample-config`
+TEST_CERTS=`echo "$OUT" | grep TEST_CERT | wc -l`
+EXPIRED=`echo "$OUT" | grep EXPIRED | wc -l`
+REVOKED=`echo "$OUT" | grep REVOKED | wc -l`
```
bmw: This is not used. Can we remove it?

tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh
```diff
@@ -17,27 +17,6 @@ letsencrypt-auto certonly --no-self-upgrade -v --standalone --debug \
                    --register-unsafely-without-email \
                    --domain $PUBLIC_HOSTNAME --server $BOULDER_URL
 
-# we have to jump through some hoops to cope with relative paths in renewal
```
bmw: It looks like the permissions on this file were changed and the execute permission was lost. Can we add it back?

adferrand: Corrected. As a side note: I would not trust the GIT client to always setup the executable permissions for me, as this kind of modification are quite hard to see. You should add a `chmod +x` at the relevant place for letstest/scripts/*.sh during the setup of letstest.

tests/certbot-boulder-integration.sh
adferrand: Removed.

tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh
bmw: Sounds good! Want to make an issue/PR?

adferrand:
I move `sample-config` in `test/integration`, and removed the empty folder `test/letstest/testdata`.

tests/letstest/testdata/sample-config/live/a.encryption-example.com/cert.pem
```diff
@@ -1 +0,0 @@
-../../archive/a.encryption-example.com/cert1.pem
```
bmw: It looks like these files got changed from symlinks to files containing the path to the relevant file somehow.

The scary thing is, our tests still passed! It looks like our cleanup code in the integration tests is masking the real nonzero exit code.

adferrand: Sorry, I made the move on Windows: on this platform POSIX symlinks are represented as real files containing the path to the linked file. Fix that by repushing from Linux.

Yes, this is not very good. I already saw the tests failing, so I think in a way or another, it catches several potential errors, but clearly not all of them.

Soon certbot-ci will be there with its rewriting of these tests in Python. Then it will not be possible anymore to have some weird shell behavior hiding non zero exit codes.

adferrand: I checked the integration tests log directly, it is working now.

tests/certbot-boulder-integration.sh
```diff
@@ -15,8 +15,10 @@ command -v python > /dev/null || (echo "Error, python executable is not in the P
 
 . ./tests/integration/_common.sh
 export PATH="$PATH:/usr/sbin"  # /usr/sbin/nginx
+CURRENT_DIR="$(pwd)"
 
 cleanup_and_exit() {
+    cd $CURRENT_DIR
```
bmw: Sorry to keep adding minor things in separate reviews, but I thought the unreported failures in our tests was really bad so I went digging. It turns out the source of the problem is this line. It should be moved below `EXIT_STATUS=$?` so we're capturing the exit status of the last command ran rather than `cd`.

I tested this at https://travis-ci.com/certbot/certbot/builds/102535040 with the prior config directory using files and not symlinks and the tests properly failed.

adferrand: No problem. In particular if it is my modifications that started this ^^

I changed the order of the statements, to get the correct exit code.

PR #6758, opened 2019-02-11
Remove keyAuthorization field from the challenge response JWS token

### LINKED ISSUE #6755: Remove keyAuthorization field when POSTing challenges
The latest ACME spec doesn't want a keyAuthorization field when POSTing a challenge; now you just POST a signed "{}".

This is enforced by Pebble in `-strict` mode, and we're planning to make that the default on Feb 15: https://github.com/letsencrypt/pebble/issues/192. After that you'll be able to check out an older tagged version of Pebble to make sure tests don't break.

If there's any concern about breaking with non-public servers that aren't up-to-date with the spec, it would be reasonable to make this a v2-only change if that's not too hard. Boulder should support "{}" for both v1 and v2.

adferrand:
Fixes #6755.

POSTing the `keyAuthorization` in a JWS token when answering an ACME challenge, has been deprecated for some time now. Indeed, this is superfluous as the request is already authentified by the JWS signature.

Boulder still accepts to see this field in the JWS token, and ignore it. Pebble in non strict mode also. But Pebble in strict mode refuses the request, to prepare complete removal of this field in ACME v2.

Certbot still sends the `keyAuthorization` field. This PR removes it, and makes Certbot compliant with current ACME v2 protocol, and so Pebble in strict mode.

The implementation chosen here is done to reduce at most the surface of the API change, to mitigate risk of breakage in third-party applications that rely on `acme`. Indeed, the objects themselves see their API untouched: only the JSON serialization, done to build the JWS token, filter out the `keyAuthorization` field.

See also https://github.com/letsencrypt/pebble/issues/192 for implementation details server side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
   warnings described at https://github.com/certbot/josepy/issues/13.
 * Apache plugin now respects CERTBOT_DOCS environment variable when adding
   command line defaults.
+* `acme` module avoids to send the `keyAuthorization` in the JWS token while
+  POSTing to a challenge, as it is deprecated on the current ACME specification
+  draft. To ease the migration path for ACME CA servers, Certbot will use a
+  temporary fallback consisting in sending `keyAuthorization` if a `malformed`
+  error is received: this fallback will be removed on Certbot 0.34.0.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
   warnings described at https://github.com/certbot/josepy/issues/13.
 * Apache plugin now respects CERTBOT_DOCS environment variable when adding
   command line defaults.
-* `acme` module avoids to send the `keyAuthorization` in the JWS token while
-  POSTing to a challenge, as it is deprecated on the current ACME specification
-  draft. To ease the migration path for ACME CA servers, Certbot will use a
-  temporary fallback consisting in sending `keyAuthorization` if a `malformed`
-  error is received: this fallback will be removed on Certbot 0.34.0.
+* The `acme` module avoids sending the `keyAuthorization` field in the JWS
+  payload when responding to a challenge as the field is not included in the
+  current ACME protocol. To ease the migration path for ACME CA servers,
+  Certbot and its `acme` module will first try the request without the
+  `keyAuthorization` field but will temporarily retry the request with the
+  field included if a `malformed` error is received. This fallback will be
+  removed in version 0.34.0.
 
 ### Fixed
 
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -107,6 +107,7 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
     """
     key_authorization = jose.Field("keyAuthorization")
     thumbprint_hash_function = hashes.SHA256
+    authorization_key_config = {'dump': False}
 
     def verify(self, chall, account_public_key):
         """Verify the key authorization.
@@ -139,6 +140,21 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
             return False
 
         return True
+
+    def dump_authorization_key(self, dump):
+        # type: (bool) -> None
+        """
+        Set if keyAuthorization is dumped in the JSON representation of this ChallengeResponse.
+        :param bool dump: True to dump the keyAuthorization, False otherwise
+        """
+        self.authorization_key_config['dump'] = dump
+
+    def to_partial_json(self):
+        jobj = super(KeyAuthorizationChallengeResponse, self).to_partial_json()
+        if not self.authorization_key_config.get('dump', False):
+            jobj.pop('keyAuthorization', None)
+
+        return jobj
 
 
 @six.add_metaclass(abc.ABCMeta)
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -99,13 +99,27 @@ class _TokenChallenge(Challenge):
         return b'..' not in self.token and b'/' not in self.token
 
 
+class _KeyAuthorizationField(jose.Field):
+    """
+    JSON field 'keyAuthorization', that is not displayed in a JSON serialization,
+    as it is deprecated to have this field in the JWS token of a challenge request.
+    See https://github.com/letsencrypt/pebble/issues/192
+    """
+    def __init__(self):
+        super(_KeyAuthorizationField, self).__init__('keyAuthorization')
+
+    def omit(self, value):
+        """Always ignore this field in a JSON serialization"""
+        return True
+
+
 class KeyAuthorizationChallengeResponse(ChallengeResponse):
     """Response to Challenges based on Key Authorization.
 
     :param unicode key_authorization:
 
     """
-    key_authorization = jose.Field("keyAuthorization")
+    key_authorization = _KeyAuthorizationField()
     thumbprint_hash_function = hashes.SHA256
 
     def verify(self, chall, account_public_key):
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -110,7 +110,7 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
 
     def __init__(self, *args, **kwargs):
         super(KeyAuthorizationChallengeResponse, self).__init__(*args, **kwargs)
-        object.__setattr__(self, "authorization_key_config", {"dump": False})
+        object.__setattr__(self, "_authorization_key_config", {"dump": False})
 
     def verify(self, chall, account_public_key):
         """Verify the key authorization.
@@ -150,11 +150,11 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
         Set if keyAuthorization is dumped in the JSON representation of this ChallengeResponse.
         :param bool dump: True to dump the keyAuthorization, False otherwise
         """
-        self.authorization_key_config['dump'] = dump
+        self._authorization_key_config['dump'] = dump  # pylint: disable=no-member
 
     def to_partial_json(self):
         jobj = super(KeyAuthorizationChallengeResponse, self).to_partial_json()
-        if not self.authorization_key_config.get('dump', False):
+        if not self._authorization_key_config.get('dump', False):  # pylint: disable=no-member
             jobj.pop('keyAuthorization', None)
 
         return jobj
--- a/acme/acme/challenges.py
+++ b/acme/acme/challenges.py
@@ -110,7 +110,7 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
 
     def __init__(self, *args, **kwargs):
         super(KeyAuthorizationChallengeResponse, self).__init__(*args, **kwargs)
-        object.__setattr__(self, "_authorization_key_config", {"dump": False})
+        self._dump_authorization_key(False)
 
     def verify(self, chall, account_public_key):
         """Verify the key authorization.
@@ -144,17 +144,18 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
 
         return True
 
-    def dump_authorization_key(self, dump):
+    def _dump_authorization_key(self, dump):
         # type: (bool) -> None
         """
         Set if keyAuthorization is dumped in the JSON representation of this ChallengeResponse.
+        NB: This method is declared as private because it will eventually be removed.
         :param bool dump: True to dump the keyAuthorization, False otherwise
         """
-        self._authorization_key_config['dump'] = dump  # pylint: disable=no-member
+        object.__setattr__(self, '_dump_auth_key', dump)
 
     def to_partial_json(self):
         jobj = super(KeyAuthorizationChallengeResponse, self).to_partial_json()
-        if not self._authorization_key_config.get('dump', False):  # pylint: disable=no-member
+        if not self._dump_auth_key:  # pylint: disable=no-member
             jobj.pop('keyAuthorization', None)
 
         return jobj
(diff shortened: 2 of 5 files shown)


adferrand:
I choose this implementation because it has a very low impact on API, while making in sync the `KeyAuthorizationResponseBody` in sync with the ACME spec, that avoids also to leak the keyAuthorization accidentally in the log, in case of JSON dump, similarly to what is done on Boulder and Pebble.

However, if you have concerns on the retro-compatibility (but think on it twice in term of the impact on a deprecated protocol vs security related improvement), I have another proposition with API low impact.

It is to do like POST-as-GET. From the lowest level of POST implementation in `acme` (both for V1 and  V2):
- check if the body to be dumped is of type `KeyAuthorizationResponseBody`
- if so, try first with dump expurged of `keyAuthorization`
- expect a `malformed` error
- retry with dump including `keyAuthorization`
- die for any further error

For other bodies, normal process. And in any circumstances, do not log the `keyAuthorization` from the body dump (either because it is not present, or it is edited if so).

Formely, the spec says to send an empty `{}`, which is not the case here (we have still the others fields, like `resource`). But sending an empty JSON is refused by Boulder v1 mode, while juste removing `keyAuthorization` is accepted.

If it becomes a problem on Boulder V2, we could pass to a three steps try, instead of two (empty, `keyAuthorization` expurged, with `keyAuthorization`, but hopefully at this time the ACME v1 will be totally removed, and we will not care about it anymore.

adferrand:
However again on the security related issue, this process exposes Certbot to leak the `keyAuthorization` if by MITM attack it receives a `malformed` request error.

So considering that:
- the deprecation is one year old, like TLS-SNI-01
- TLS-SNI-01 removal next month will nonetheless force the migration to a recent ecosystem for acme clients and servers
- most used server (Boulder) accepts a request without `keyAuthorization` for both v1 and v2 protocol
- all of this is about mitigating security issues

I will still strongly militate for a clean removal without fallback.

cpu:
> There are other ACME CAs out there using Certbot and I worry that this change might break things for them.

There's no spec for "ACME v1", just a reference implementation (Boulder). I'm not sure when in the history of Boulder we removed processing a `keyAuthorization` in a POST body to a challenge but it has been at least 2 years.

Personally I think anyone that implemented an "ACME v1" server made a poor choice and there are two solutions available to them that wouldn't be a burden on Certbot: they could implement a protocol with a specification (ACMEv2), or to invest the engineering effort required to track and adopt Boulder's changes as they happen. 

Server operators that do neither of these things are guaranteed to have endless compatibility problems and I would hate to see the client ecosystem encourage more "ACME v1" development by working too hard to remain compatible.

bmw:
@adferrand, you wrote a lot about your concerns about leaking the key authorization. What is the attack here?

Rather than falling back, I think my personally preferred behavior is to change Certbot's behavior based on the ACME version as suggested by jsha at https://github.com/certbot/certbot/issues/6755#issue-408033490. I'd argue that "ACME v1" (however you want to define it) requires the `keyAuthorzation` while including for "ACME v2" is wrong.

@cpu,

> There's no spec for "ACME v1", just a reference implementation (Boulder).

Right, but as you know, there has been a lot of client software implemented against this reference implementation. The other ACME CAs listed at https://en.wikipedia.org/wiki/Automated_Certificate_Management_Environment#CAs_&_PKIs_that_offer_ACME_certificates whose directory endpoint I've been able to query implement "ACME v1".

> I'm not sure when in the history of Boulder we removed processing a `keyAuthorization` in a POST body to a challenge but it has been at least 2 years.

If that's the case, simply suddenly changing the behavior is honestly tempting to me, however, this does not seem to be documented at https://github.com/letsencrypt/boulder/blob/master/docs/acme-divergences.md so I wouldn't expect CAs other than boulder who implemented "ACME v1" to know this.

> Personally I think anyone that implemented an "ACME v1" server made a poor choice and there are two solutions available to them that wouldn't be a burden on Certbot: they could implement a protocol with a specification (ACMEv2), or they could invest the engineering effort required to track and adopt Boulder's changes as they happen.

I agree this would be great but I don't think this happening.

> Server operators that do neither of these things are guaranteed to have endless compatibility problems and I would hate to see the client ecosystem encourage more "ACME v1" development by working too hard to remain compatible.

I think this is likely particularly painful for Certbot. For better or worse, many CAs recommend people use Certbot as well. I've heard stories about ACME servers being implemented with the goal being that "it works with Certbot".

With that said, I personally think we're making progress here. We've been talking about deprecating and then removing our support for "ACME v1" to strongly discourage CAs from continuing to rely on this dated "spec".

adferrand:
@bmw I wrote about security, because it seems to be an initial concern to remove it server side, but I may be wrong. Maybe @jsha or @cpu could confirm or not that it is the case.

For the other proposition I made, I prefer it over to decide depending on the acme version, because Boulder seems to support it with acme v1, so we will end up to remove this deprecated key most of the time, for any acme version used.

adferrand:
On my previous post, for the concerns, I was thinking of this: https://github.com/letsencrypt/boulder/pull/3526, that prevent to display the `keyAuthorization` value in the logs.

jsha:
I haven't followed this whole thread, but it sounds like you're asking: Is the keyAuthorization a confidential value? If so, no it's not. The removal in https://github.com/letsencrypt/boulder/pull/3526 was just to get closer to matching the spec, not to avoid writing anything to logs.

adferrand:
OK, then the fallback is reasonable, and not insecure. I will do that. Are we agree on my approach, to rely on the `malformed` error instead of the acme version used? It will have also the benefit to cover servers that do not implement a recent draft of the acme v2 draft.

Once it becomes a RFC, we should remove the fallback.

cpu:
> The other ACME CAs listed at https://en.wikipedia.org/wiki/Automated_Certificate_Management_Environment#CAs_&_PKIs_that_offer_ACME_certificates whose directory endpoint I've been able to query implement "ACME v1".

I don't think these other CAs implement ACME. They implement something unspecified.

> If that's the case, simply suddenly changing the behavior is honestly tempting to me, however, this does not seem to be documented at https://github.com/letsencrypt/boulder/blob/master/docs/acme-divergences.md so I wouldn't expect CAs other than boulder who implemented "ACME v1" to know this.

This isn't a specification and shouldn't be used by other CAs in place of one - it isn't intended for server developers. Arguably it isn't a divergence because we won't error if you do POST the keyAuthorization, its just ignored. What I'm trying to say is that if another CA implemented "ACME v1" then the way they are supposed to know about these changes is to closely follow Boulder/Certbot development. That's the trade-off of choosing ACME v1.

> I've heard stories about ACME servers being implemented with the goal being that "it works with Certbot".

Right, and this is a chance to indicate that's not a strategy the team supports. It's your call if you do want to send a message that this is an acceptable way to implement ACME but I'm trying to convince you that it's a bad idea for the long term health of the ecosystem. One of the major goals of ACME is to promote a vendor neutral protocol that would let users switch server providers and clients without risking compatibility. Ossifying on what Certbot supported at a specific point in time for an unspecified protocol is the opposite of this.

> We've been talking about deprecating and then removing our support for "ACME v1" to strongly discourage CAs from continuing to rely on this dated "spec".

I'm a big +1 for this because again, there is no spec! There's not even a single draft # that can be pointed to that approximates ACMEv1. It's a reference implementation spanning multiple drafts as a product of the co-evolution that occurred.

It sounds like I'm in the minority here so I won't keep beating on this dead horse. If you want to maintain the `keyAuthorization` challenge POST that's your call. Thanks for engaging with me on the discussion!

adferrand:
@bmw you have heard of other ACME CA servers to develop by following the Certbot implementation. @cpu you are talking about not giving a bad signal to developer about maintaining deprecated and unspecified features, and also state that implementing ACME "v1" is in fact following Boulder/Certbot development.

As it is not a security update, there is no urge to process immediately. But it is clearly an occasion to be proactive and engage a migration path to incite other ACME CA server maintainers to clean their implementation.

Instead of waiting official RFC for ACME, or support forever `keyAuthorization` in the fear of breaking stuff, could we implement my proposition, advert in changelog that `keyAuthorization` is deprecated and should not be used anymore, then completely remove it in N+2 releases of Certbot ?

bmw:
> Thanks for engaging with me on the discussion!

I'm happy to respond to some/all of what you wrote or answer any questions if you'd like to talk about this more @cpu. I largely agree with you with my main hesitation here being suddenly making these changes without sufficient notice.

@adferrand, your approach sounds good to me.

cpu:
@bmw If everyone is on board with @adferrand's plan I don't think we need to talk more. It sounds like a deprecation notice and delaying the removal a release or two would address your hesitation about a sudden change/sufficient notice and also result in it being removed in the near future so I'm :+1:

adferrand:
@bmw I implemented the new behavior with fallback. Changelog is also updated.

acme/acme/challenges.py
```diff
@@ -107,6 +107,7 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
     """
     key_authorization = jose.Field("keyAuthorization")
     thumbprint_hash_function = hashes.SHA256
+    authorization_key_config = {'dump': False}
```
bmw: This is a class level attribute so changing it affects all `KeyAuthorizationChallengeResponse` objects. I'm not sure if this is the behavior you wanted, but I think it's one step away from mutable global state which I'd personally like to avoid.

Can we make changes here so if there's any saved state in this object, it only affects the current instance? If we want to remember this setting for other objects, can we save the state in the `acme.client` object?

acme/acme/challenges_test.py
```diff
@@ -94,7 +94,8 @@ def setUp(self):
         self.response = self.chall.response(KEY)
```
bmw: I think we should also add simple unit tests for `to_partial_json` with the `keyAuthorization` to this file.

acme/acme/client.py
```diff
@@ -155,7 +156,23 @@ def answer_challenge(self, challb, response):
         :raises .UnexpectedUpdate:
 
         """
-        response = self._post(challb.uri, response)
+        # Because sending keyAuthorization in a response challenge is deprecated,
```
bmw: nit: I'd argue that it's worse than deprecated, it has been completely ripped out of the spec.

acme/acme/client.py
```diff
@@ -155,7 +156,23 @@ def answer_challenge(self, challb, response):
...
+        # we let KeyAuthorizationResponseChallenge not dumping it (default behavior).
```
bmw:
```suggestion
        # it is not included in the KeyAuthorizationResponseChallenge JSON by default.
```

acme/acme/client.py
```diff
@@ -155,7 +156,23 @@ def answer_challenge(self, challb, response):
...
+        except messages.Error as error:
+            if (error.code == 'malformed'
+                    and isinstance(response, challenges.KeyAuthorizationChallengeResponse)):
+                logger.debug('Error while responding to a challenge without keyAuthorization '
+                             'in the JWS, your ACME CA server may not support it:\n%s', error)
+                logger.debug('Retrying request with keyAuthorization set.')
+                response.dump_authorization_key(True)
+                response = self._post(challb.uri, response)
+            else:  # pragma: no cover
+                raise
```
bmw: nit: Adding a test to get coverage on this line seems trivial. Any reason not to add it?

acme/acme/challenges.py
adferrand: It was not my intention indeed. I was a little puzzled by the behavior of `josepy` package, that change a lot of default Python object behavior, including the immutability, so I forgot to fix that.

I fix it now.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
   warnings described at https://github.com/certbot/josepy/issues/13.
 * Apache plugin now respects CERTBOT_DOCS environment variable when adding
   command line defaults.
+* `acme` module avoids to send the `keyAuthorization` in the JWS token while
```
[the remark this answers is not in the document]
adferrand: Yes, it is better ^^

acme/acme/client.py
adferrand: I even did the negative test to check that branch! Enabling the coverage here.

acme/acme/client.py
adferrand: I will use a little more moderated term in the comment, but that is the idea ^^ Modified.

acme/acme/challenges.py
bmw: If you spend much time fighting this, let me know and I'm happy to help you. `acme` and `josepy` have a lot of (largely unnecessary) magic going on.

acme/acme/challenges_test.py
adferrand: Done.

acme/acme/challenges.py
adferrand: That is OK. The most difficult think was to understand that declaring a class parameter an assign it to the result of a josepy.Field result was creating effectively an instance parameter suitable for a controlled deserialization. And that most of the natural mutability of a python object was disabled in this case.

acme/acme/client.py
```diff
@@ -155,7 +156,23 @@ def answer_challenge(self, challb, response):
         :raises .UnexpectedUpdate:
 
         """
-        response = self._post(challb.uri, response)
+        # Because sending keyAuthorization in a response challenge has been removed from the ACME
+        # specs, it is not included in the KeyAuthorizationResponseChallenge JSON by default.
```
bmw:
```suggestion
        # spec, it is not included in the KeyAuthorizationResponseChallenge JSON by default.
```

acme/acme/challenges.py
```diff
@@ -108,6 +108,10 @@ class KeyAuthorizationChallengeResponse(ChallengeResponse):
     key_authorization = jose.Field("keyAuthorization")
     thumbprint_hash_function = hashes.SHA256
 
+    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
adferrand: Yes, it is cleaner. I updated the code accordingly.

acme/acme/challenges.py
```diff
@@ -140,6 +144,21 @@ def verify(self, chall, account_public_key):
 
         return True
 
+    def dump_authorization_key(self, dump):
```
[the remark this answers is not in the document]
adferrand: Yes, it will stay only two months. I also added in the docstring a nota bene about why the method is private.

PR #6759, opened 2019-02-12
[Docker] Update to Python 3.6

J0WI:
Not every module seems to be 3.7 compatible, but I think 3.6 is fine?
https://github.com/certbot/certbot/blob/7b7f7b25fb25414156c9e80d4829531faa78a05e/certbot-dns-ovh/setup.py#L37-L45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
-*
+* Certbot's official Docker images are now based on Python 3.6 rather than 2.7.
 
 ### Fixed
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2-alpine3.9
+FROM python:3.6-alpine3.10
 
 ENTRYPOINT [ "certbot" ]
 EXPOSE 80 443


J0WI:
@bmw would you like to review this?

bmw:
Sorry this PR has sat around for so long @J0WI.

While we need to switch the Dockerfile to Python 3 pretty soon and I would like us to do so, I'm slightly worried about us just suddenly changing this behavior which could break things for people using the unpinned version of the Dockerfile.

Do you know if sudden breakage like this is expected? Are there any best practices here?

Also, as I said on #7250, once certbot/certbot-docker#1 lands, I think we'll be deleting the Dockerfile in this repo. Want to open a similar PR on that repo once that PR is merged?

J0WI:
> Do you know if sudden breakage like this is expected? Are there any best practices here?Are there any best practices here?

It builds and executes successfully, but I haven't tested all cases and plugins.
Best practice would be to provide two additional tags for `python3` and `python2`. IMHO the `latest` tag should default to `python3` then.

bmw:
My main concern personally would be mitigating problems for things like 3rd party plugins when the base image is suddenly changed from Python 2 to Python 3. I've asked around this problem a bit and haven't heard of any suggestions for how to resolve this.

In the absence of any standard practices to avoid this problem or anything we can easily to do to help mitigate it, I think we can just do the sudden upgrade. While having additional tags could be nice in theory, Python 2 is going away soon and I personally don't think it's worth it to set up another tag. If people really want to stick with Python 2 to keep things working until they have time to upgrade, they can use an older image.

bmw:
I'm closing this PR because these Dockerfiles have changed a bit and moved to https://github.com/certbot-docker/certbot-docker, but in the continued absence of any easy suggestions to ease the migration for people, if you'd like to open a PR on that repo changing the Python version to Python 3, I'll merge it.

PR #6765, opened 2019-02-13
Fix the pebble fetch script

adferrand:
This PR updates and fixes `pebble-fetch.sh` considering latest improvements done on Pebble, to start a working instance.

### DIFF
--- a/tests/pebble-fetch.sh
+++ b/tests/pebble-fetch.sh
@@ -2,7 +2,7 @@
 # Download and run Pebble instance for integration testing
 set -xe
 
-PEBBLE_VERSION=2018-11-02
+PEBBLE_VERSION=v1.0.1
 
 # We reuse the same GOPATH-style directory than for Boulder.
 # Pebble does not need it, but it will make the installation consistent with Boulder's one.
@@ -15,7 +15,7 @@ cat << UNLIKELY_EOF > "$PEBBLEPATH/docker-compose.yml"
 version: '3'
 services:
   pebble:
-    image: letsencrypt/pebble
+    image: letsencrypt/pebble:${PEBBLE_VERSION}
     command: pebble -dnsserver 10.30.50.3:8053
     environment:
     - PEBBLE_VA_NOSLEEP=1
@@ -25,7 +25,7 @@ services:
       acmenet:
         ipv4_address: 10.30.50.2
   challtestsrv:
-    image: letsencrypt/pebble-challtestsrv
+    image: letsencrypt/pebble-challtestsrv:${PEBBLE_VERSION}
     command: pebble-challtestsrv -defaultIPv6 "" -defaultIPv4 10.30.50.1
     ports:
       - 8055:8055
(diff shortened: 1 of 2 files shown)


tests/pebble-fetch.sh
```diff
@@ -13,15 +13,31 @@ mkdir -p ${PEBBLEPATH}
 
 cat << UNLIKELY_EOF > "$PEBBLEPATH/docker-compose.yml"
 version: '3'
-
 services:
- pebble:
-  image: letsencrypt/pebble:${PEBBLE_VERSION}
-  command: pebble -strict ${PEBBLE_STRICT:-false} -dnsserver 10.77.77.1
```
bmw: Did the behavior around `-strict` change or did this environment variable never work? If we're not going to allow people to set this variable to enable strict mode, can we remove `PEBBLE_STRICT` from `tox.ini`?

tests/pebble-fetch.sh
```diff
@@ -13,15 +13,31 @@ mkdir -p ${PEBBLEPATH}
...
-  ports:
-    - 14000:14000
-  environment:
+  pebble:
+    image: letsencrypt/pebble
+    command: pebble -dnsserver 10.30.50.3:8053
+    environment:
+    - PEBBLE_VA_ALWAYS_VALID=1
```
bmw: Why should we set this environment variable?

tests/pebble-fetch.sh
```diff
@@ -13,15 +13,31 @@ mkdir -p ${PEBBLEPATH}
...
+        ipv4_address: 10.30.50.2
+  challtestsrv:
+    image: letsencrypt/pebble-challtestsrv
+    command: pebble-challtestsrv -defaultIPv6 "" -defaultIPv4 10.30.50.1
+    ports:
+      - 8055:8055
+    networks:
+      acmenet:
+        ipv4_address: 10.30.50.3
+networks:
```
bmw: Why don't we set the `driver` values like is done at https://github.com/letsencrypt/pebble/blob/master/docker-compose.yml?

tests/pebble-fetch.sh
adferrand: Tomorrow, pebble will use the strict mode by default, and I saw that in some situation, pebble in non strict mode was more permissive than Boulder, that is not useful.

So yes, this environnement variable is not useful anymore, and we can remove `PEBBLE_STRICT` from `tox.ini`, without consequences because `apacheconftest` does not issue certificates using pebble.

tests/pebble-fetch.sh
adferrand: I did not see an effect to add it or not, not harmful to keep it so I will.

tests/pebble-fetch.sh
adferrand: Glitch from my tests, to be removed.

adferrand: Done

tests/pebble-fetch.sh
adferrand: Done

tests/pebble-fetch.sh
adferrand: Done

adferrand:
Also pin again the Pebble/Challtestsrv version.

PR #6766, opened 2019-02-13
Implement Retry-After, and refactor authorization polling

### LINKED ISSUE #5789: Certbot should respect Retry-After when polling authzs

adferrand:
Fixes #5789 

This PR is about allowing Certbot to respect the `Retry-After` HTTP header that an ACME CA server can return to a client POSTing to a challenge, to instruct him and retry the request later.

However, this feature was not easily implementable in the current code of `certbot.auth_handler`, because the code became really hard to read. In fact, @bmw was thinking that the code was really deceiving, and a lot of supposed functionalities declared in the comments were in fact not implemented or not functional.

So I took the time to understand what was going on, and effectively, most of the code is in fact not usable or not used. Then I did a refactoring against the bare ACME spec about what to do to prepare challenges, instruct the ACME CA server to perform them, then polling regularly the authorization resources until they are decided (valid or invalid).

And of course this implementation takes care of Retry-After ^^

I added a lot of comments in the new implementation, to explain what is going on for a future developer. The workflow I used is relying on the relationships between authorizations and challenges states as described in section 7.1.6 of the ACME spec draft: https://datatracker.ietf.org/doc/draft-ietf-acme-acme/


certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
 
-                # Send all Responses - this modifies achalls
-                self._respond(aauthzrs, resp, best_effort)
+            # Inform the ACME CA server that challenges are available for validation.
+            for achall, resp in zip(achalls, resps):
+                self.acme.answer_challenge(achall.challb, resp)
 
-        # Just make sure all decisions are complete.
-        self.verify_authzr_complete(aauthzrs)
+            # Wait for authorizations to be decided.
```
bmw: Instead of "decided" we should say something like "checked" or "verified".

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            authzrs_still_pending = self._poll_authorizations(authzrs, max_retries, best_effort)
 
-        # Only return valid authorizations
-        ret_val = [aauthzr.authzr for aauthzr in aauthzrs
-                   if aauthzr.authzr.body.status == messages.STATUS_VALID]
+            # All authorizations should be decided now. If not, we reached the max retries.
```
bmw: Instead of "decided" we should say something like "checked" or "verified".

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            authzrs_validated = [authzr for authzr in authzrs
+                                 if authzr.body.status == messages.STATUS_VALID]
+            if not authzrs_validated:
+                raise errors.AuthorizationError('All challenges have failed.')
 
-        return ret_val
+            return authzrs_validated
 
-    def _choose_challenges(self, aauthzrs):
+    def _poll_authorizations(self, authzrs, max_retries, best_effort):
```
bmw: It works, but the return value of this function seems a little awkward to me. Do you think we can use something else instead?

certbot/auth_handler.py
```diff
@@ -306,43 +197,18 @@ def _get_chall_pref(self, domain):
...
 
-    def _cleanup_challenges(self, aauthzrs, achalls=None):
+    def _cleanup_challenges(self, achalls):
         """Cleanup challenges.
 
-        :param aauthzrs: authorizations and their selected annotated
+        :param authzrs: authorizations and their selected annotated
             challenges
-        :type aauthzrs: `list` of `AnnotatedAuthzr`
+        :type authzrs: `list` of `AnnotatedAuthzr`
```
bmw: This parameter is no longer passed to this method.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+
+            # Handle failed authorizations: with best effort this is only a warning,
+            # otherwise an exception.
+            authzrs_failed = [authzr for index, (authzr, _) in authzrs_to_check.items()
+                              if authzr.body.status == messages.STATUS_INVALID]
+            if authzrs_failed and best_effort:
+                logger.warning('Following authorizations have failed: %s', authzrs_failed)
+            elif authzrs_failed:
+                raise errors.AuthorizationError('Some challenges have failed: {0}.'
+                                                .format(authzrs_failed))
```
bmw: I personally don't think we should be logging or raising the list of the full authzr objects like this. It's very hard to read and the full object has already been logged by acme.

bmw:
@jsha, one change being made in this PR is we no longer sleep before beginning to poll authorizations.

Simplified prior pseudocode was:
```
Respond to challenges
while authorizations are pending:
    sleep 3s
    poll authorizations
```
Pseudocode for this PR is:
```
Respond to challenges
while true:
    poll authorizations
    if no pending authorizations:
        break
    if retry-after is set:
        sleep maximum retry-after value
    else:
        sleep 3s
```
Do you have concerns about us removing this initial sleep?

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
-        :param acme.messages.OrderResource orderr: must have
-            authorizations filled in
-        :param bool best_effort: Whether or not all authorizations are
-            required (this is useful in renewal)
-
-        :returns: List of authorization resources
-        :rtype: list
-
-        :raises .AuthorizationError: If unable to retrieve all
-            authorizations
```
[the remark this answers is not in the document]
adferrand: No, I added it back.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+        if not achalls:
+            return authzrs
+
+        # Starting to now, challenges will be cleaned at the end no matter what.
+        with error_handler.ExitHandler(self._cleanup_challenges, achalls):
+            # To begin, let's ask the authenticator plugin to perform all challenges.
+            try:
+                resps = self.auth.perform(achalls)
+
+                # If debug is on, wait for user call before starting the verification process.
```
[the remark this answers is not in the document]
adferrand: Done

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+                logger.info('Attempting to clean up outstanding challenges...')
+                raise error
+            # All challenges should have been processed by the authenticator.
+            assert len(resps) == len(achalls), 'Some challenges have not been performed.'
 
-                # Send all Responses - this modifies achalls
-                self._respond(aauthzrs, resp, best_effort)
+            # Inform the ACME CA server that challenges are available for validation.
+            for achall, resp in zip(achalls, resps):
+                self.acme.answer_challenge(achall.challb, resp)
```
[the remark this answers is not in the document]
adferrand: Indeed, this is the idea. I update the documentation in `interface.py`.

certbot/auth_handler.py
adferrand: I replace by checked.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+        authzrs = orderr.authorizations[:]
+        if not authzrs:
+            raise errors.AuthorizationError('No authorization to handle.')
+
+        # Retrieve challenges that need to be performed to validate authorizations.
+        achalls = self._choose_challenges(authzrs)
+        if not achalls:
+            return authzrs
+
+        # Starting to now, challenges will be cleaned at the end no matter what.
```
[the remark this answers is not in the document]
adferrand: Corrected.

certbot/auth_handler.py
adferrand: Done

certbot/auth_handler.py
adferrand: I replaced by an exception handling, the method does not return something anymore.

certbot/auth_handler.py
adferrand: Corrected.

certbot/auth_handler.py
adferrand: Allright.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+                                if authzr.body.status == messages.STATUS_PENDING}
+            if not authzrs_to_check:
+                break
+
+            # Be merciful with the ACME server CA, check the Retry-After header returned,
+            # and wait this time before next polling. From all the pending authorizations
+            # pending, we take the greatest one, and avoid this way to poll an authorization
+            # before its relevant Retry-After value.
+            retry_after = max(self.acme.retry_after(resp, 3)
+                              for index, (_, resp) in authzrs_to_check.items())
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            authzrs_to_check = {index: self.acme.poll(authzr) for index, (authzr, _)
+                                in authzrs_to_check.items()}
+            # Update the original list of authzr with the updated authzrs from server.
+            for index, (authzr, _) in authzrs_to_check.items():
+                authzrs[index] = authzr
+
+            # Handle failed authorizations: with best effort this is only a warning,
+            # otherwise an exception.
+            authzrs_failed = [authzr for index, (authzr, _) in authzrs_to_check.items()
+                              if authzr.body.status == messages.STATUS_INVALID]
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            # Extract out the authorization already decided for next poll iteration.
+            # Poll may stop here because there is no pending authorizations anymore.
+            authzrs_to_check = {index: (authzr, resp) for index, (authzr, resp)
+                                in authzrs_to_check.items()
+                                if authzr.body.status == messages.STATUS_PENDING}
+            if not authzrs_to_check:
+                break
+
+            # Be merciful with the ACME server CA, check the Retry-After header returned,
+            # and wait this time before next polling. From all the pending authorizations
```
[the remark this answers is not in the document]
adferrand: `polling again`

certbot/auth_handler.py
```diff
@@ -47,242 +43,137 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            # pending, we take the greatest one, and avoid this way to poll an authorization
+            # before its relevant Retry-After value.
+            retry_after = max(self.acme.retry_after(resp, 3)
+                              for index, (_, resp) in authzrs_to_check.items())
+            time.sleep((retry_after - datetime.datetime.now()).total_seconds())
```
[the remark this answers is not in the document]
adferrand: Done.

adferrand:
I added back the formatted report for failed authorizations.

jsha:
> Do you have concerns about us removing this initial sleep?

It would probably be better to stick it back in since it will almost never result in a "valid" result. You're welcome to make it shorter if you like -- I think most validations should complete in under 1s in ideal circumstances.

adferrand:
I added a first second before the first polling.

certbot/tests/auth_handler_test.py
```diff
@@ -95,23 +94,24 @@ def _test_name1_tls_sni_01_1_common(self, combos):
...
 
-        self.assertEqual(mock_poll.call_count, 1)
-        chall_update = mock_poll.call_args[0][1]
-        self.assertEqual(list(six.iterkeys(chall_update)), [0])
-        self.assertEqual(len(chall_update.values()), 1)
+            self.assertEqual(self.mock_net.poll.call_count, 2)  # Because there is one retry
+            self.assertEqual(mock_time.sleep.call_count, 2)
+            # Retry-After header is 1 second, but at the time sleep is invoked, several
+            # instructions are executed, and next pool is in less than a second.
+            self.assertTrue(mock_time.sleep.call_args_list[0][0][0] <= 1)
```
bmw: It's probably pretty unlikely, but I'm worried about on a machine with limited resources, we could actually take longer than a second here and spuriously fail this test. Can we add an optional parameter to `_gen_mock_on_poll` to set the `Retry-After` value, set it to something fairly large here and then check it is less than or equal to that value but greater than 3?

If this is done, the comment in the code here will need to be updated as well.

certbot/tests/auth_handler_test.py
```diff
@@ -231,6 +213,21 @@ def test_perform_failure(self):
...
+        # We will return STATUS_PENDING twice before returning STATUS_VALID.
+        self.mock_net.poll.side_effect = _gen_mock_on_poll(retry=2)
+
+        with self.assertRaises(errors.AuthorizationError) as error:
+            # We retry only once, so retries will be exhausted before STATUS_VALID is returned.
+            self.handler.handle_authorizations(mock_order, False, 1)
+        self.assertTrue('All challenges could not be checked on time' in str(error.exception))
+
+        self.assertRaises(
+            errors.AuthorizationError, self.handler.handle_authorizations, mock_order, False, 1)
```
bmw: I don't understand why we're testing this again. Is there something I'm missing?

certbot/tests/auth_handler_test.py
```diff
@@ -290,32 +286,46 @@ def test_perform_error(self):
...
-            errors.AuthorizationError, self.handler.handle_authorizations, mock_order)
+        with test_util.patch_get_utility():
+            with self.assertRaises(errors.AuthorizationError) as error:
+                self.handler.handle_authorizations(mock_order, False)
+        self.assertTrue('Some challenges have failed' in str(error.exception))
         self.assertEqual(self.mock_auth.cleanup.call_count, 1)
         self.assertEqual(
             self.mock_auth.cleanup.call_args[0][0][0].typ, "tls-sni-01")
 
+    def test_best_effort(self):
```
bmw: It looks like we didn't have this previously either but since we changed a lot of logic around here, can we add a test where we use `best_effort`, some challenges fail, but we don't raise an exception?

certbot/auth_handler.py
```diff
@@ -136,14 +140,16 @@ def _poll_authorizations(self, authzrs, max_retries, best_effort):
...
             retry_after = max(self.acme.retry_after(resp, 3)
-                              for index, (_, resp) in authzrs_to_check.items())
-            time.sleep((retry_after - datetime.datetime.now()).total_seconds())
+                              for _, resp in authzrs_to_check.values())
+            sleep_seconds = (retry_after - datetime.datetime.now()).total_seconds()
+            time.sleep(sleep_seconds if sleep_seconds > 0 else 3)
 
-        return authzrs_to_check
+        if authzrs_to_check:
+            raise acme_errors.TimeoutError()
```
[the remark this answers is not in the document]
adferrand: You are right. Done.

certbot/interfaces.py
```diff
@@ -161,9 +161,9 @@ def perform(achalls):
         :returns: `collections.Iterable` of ACME
             :class:`~acme.challenges.ChallengeResponse` instances corresponding to each provided
             :class:`~acme.challenges.Challenge`. Not returning all expected
```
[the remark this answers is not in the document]
adferrand: Yes, raising an exception is more pythonic. I removed the comment about not giving a complete array in the parameters description. I precised "all or some" in the exception, to guide developers in the situation of handling some incompatible challenges.

certbot/auth_handler.py
```diff
@@ -136,14 +140,16 @@ def _poll_authorizations(self, authzrs, max_retries, best_effort):
...
-            # pending, we take the greatest one, and avoid this way to poll an authorization
+            # and wait this time before polling again. From all the pending authorizations
+            # pending, we take the greatest one to avoid polling an authorization
             # before its relevant Retry-After value.
             retry_after = max(self.acme.retry_after(resp, 3)
-                              for index, (_, resp) in authzrs_to_check.items())
-            time.sleep((retry_after - datetime.datetime.now()).total_seconds())
+                              for _, resp in authzrs_to_check.values())
+            sleep_seconds = (retry_after - datetime.datetime.now()).total_seconds()
+            time.sleep(sleep_seconds if sleep_seconds > 0 else 3)
```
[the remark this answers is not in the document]
adferrand: That is right. I moved the wait on top of the loop, and do not wait for negative values. My initial concern about minimal value to wait was from a previous assumption that Retry-After contains a date, but it is just a number of seconds. The actual date is calculated from the local time of the client machine.

certbot/auth_handler.py
```diff
@@ -119,15 +122,16 @@ def _poll_authorizations(self, authzrs, max_retries, best_effort):
 
             # Handle failed authorizations: with best effort this is only a warning,
             # otherwise an exception.
-            authzrs_failed = [authzr for index, (authzr, _) in authzrs_to_check.items()
+            authzrs_failed = [authzr for authzr, _ in authzrs_to_check.values()
                               if authzr.body.status == messages.STATUS_INVALID]
             if authzrs_failed and best_effort:
                 logger.warning('Following authorizations have failed: %s', authzrs_failed)
```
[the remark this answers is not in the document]
adferrand: Indeed. I think the output here will be modified compared to the previous implementation. if I remember well, I took the logger.warning from it. But it is good also when best_effort is True, to have an intelligible report of failed authorizations. Now, the report is always displayed if there is some errors, then an exception is raised if not best_effort.

certbot/auth_handler.py
```diff
@@ -47,242 +42,144 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            # Inform the ACME CA server that challenges are available for validation.
+            for achall, resp in zip(achalls, resps):
+                self.acme.answer_challenge(achall.challb, resp)
+
+            # Wait for authorizations to be checked.
+            try:
+                self._poll_authorizations(authzrs, max_retries, best_effort)
+            except acme_errors.TimeoutError:
+                # Exceeding the max polling attempts, and some authentication are still not checked.
+                raise errors.AuthorizationError('All challenges could not be checked on time.')
```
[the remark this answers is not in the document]
adferrand: Yes I was not sure because I am not an english native speaker, but "checked" in french is "vérifié", and this word in french has a positive connotation: it usually means that the subject as made actions to be sure that the object fulfills the expectations. This is not about this here, we want to say that all authorizations are not in a definitive situation, so either valid or invalid, situation resolved by the ACME CA server.

I update accordingly to your suggestion.

certbot/auth_handler.py
```diff
@@ -136,14 +140,16 @@ def _poll_authorizations(self, authzrs, max_retries, best_effort):
                 break
```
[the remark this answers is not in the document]
adferrand: Done, and ...

certbot/auth_handler.py
```diff
@@ -136,14 +140,16 @@ def _poll_authorizations(self, authzrs, max_retries, best_effort):
...
             # before its relevant Retry-After value.
             retry_after = max(self.acme.retry_after(resp, 3)
-                              for index, (_, resp) in authzrs_to_check.items())
-            time.sleep((retry_after - datetime.datetime.now()).total_seconds())
+                              for _, resp in authzrs_to_check.values())
+            sleep_seconds = (retry_after - datetime.datetime.now()).total_seconds()
+            time.sleep(sleep_seconds if sleep_seconds > 0 else 3)
 
-        return authzrs_to_check
+        if authzrs_to_check:
```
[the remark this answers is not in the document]
adferrand: ... nice catch!

certbot/auth_handler.py
```diff
@@ -396,22 +399,20 @@ def _report_no_chall_path(challbs):
...
-        :class:`certbot.achallenges.AnnotatedChallenge`.
-
-    """
-    problems = collections.defaultdict(list)\
-        # type: DefaultDict[str, List[achallenges.KeyAuthorizationAnnotatedChallenge]]
     for achall in failed_achalls:
-        if achall.error:
-            problems[achall.error.typ].append(achall)
+        problems[achall.error.typ] = problems.get(achall.error.typ, [])
+        problems[achall.error.typ].append(achall)
```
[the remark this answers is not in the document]
adferrand: I prefer to stay with the bare dict, because it removes a third-party dependency. So nice catch about setdefault. In theory, then I can merge it with the list comprehension above, but I think it would become hard to read.

certbot/tests/auth_handler_test.py
adferrand: No, I think it is something from a previous implementation of this test. I remove it.

certbot/tests/auth_handler_test.py
adferrand: I already test that indirectly with this method. On the previous one, I check that the exception raised is "Some challenges have failed", indicating that without best_effort, we have an exception directly in the loop. Here, with best_effort, I test that we raise "All challenges have failed": this is an exception raised outside of the loop.

So this test checks the contract already.

certbot/tests/auth_handler_test.py
adferrand: Done. And in fact I was not checking the correct time call. Corrected also.

certbot/auth_handler.py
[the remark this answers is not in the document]
bmw: > it usually means that the subject as made actions to be sure that the object fulfills the expectations.

This is how the word is used in English as well. In this case, I was thinking of "checked" in the sense that the CA still hasn't checked/verified that the challenges were performed correctly, but that's not really describing the immediate problem.

I think the new wording is a lot better and hopefully you do too. Feel free to push back when I suggest wording changes (or any other changes). I'm certainly wrong sometimes.

certbot/auth_handler.py
```diff
@@ -47,242 +41,142 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+        Poll the ACME CA server, to wait for confirmation that authorizations have their challenges
+        all verified. The poll may occur several times, until all authorizations are checked
+        (valid or invalid), or after a maximum of retries.
+        """
+        authzrs_to_check = {index: (authzr, None)
+                            for index, authzr in enumerate(authzrs)}
+        # Give an initial second to the ACME CA server to check the authorizations
+        sleep_seconds = 1
+        for _ in range(max_retries):
+            # Wait during appropriate time (from Retry-After, initial wait, or no wait)
```
bmw:
```suggestion
            # Wait for appropriate time (from Retry-After, initial wait, or no wait)
```

certbot/tests/auth_handler_test.py
bmw: It's pretty clear reading the current code and seeing the raised exception that best_effort should work.

I still personally think a unit test where `handle_authorizations` is called with `best_effort=True` and actually returns some but not all of the initial authorizations would be valuable though and may prevent regressions in the future as the code is refactored. I think it's only apparent that `best_effort` works with the given code structure which can change at any time.

I do see your point and we didn't have this test before either so I'll leave it up to you. If you don't think it's worth adding the test, we can skip it.

adferrand: I would lose my DevOps chocolate medal by not responding on the call for a proper unit test. Here is a complete and explicit test for best_effort, with assertions for situations of all authzrs failing and only a part of them failing.

adferrand:
And changelog updated.

certbot/auth_handler.py
```diff
@@ -47,242 +41,142 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+                authzrs[index] = authzr
+
+            # Handle failed authorizations: with best effort this is only a warning,
+            # otherwise an exception.
+            authzrs_failed = [authzr for authzr, _ in authzrs_to_check.values()
+                              if authzr.body.status == messages.STATUS_INVALID]
+            if authzrs_failed:
+                # Display report to the user
+                _report_failed_authzrs(authzrs_failed, self.account.key)
+                if not best_effort:
```
bmw: I think this is a useful addition to Certbot's output when `best_effort` is used.

One change being made here which I didn't notice before is that if Certbot is run in `--quiet` mode and you fail a nonempty subset of challenges with `--allow-subset-of-names`, on `master` you see:
```
Challenge failed for domain example.org
```
while on this branch nothing about the failure is printed to the terminal.

Can we add this message back? It might be worth logging the message unconditionally so people see it when `best_effort` isn't used as well.

certbot/auth_handler.py
```diff
@@ -47,242 +41,142 @@ def __init__(self, auth, acme_client, account, pref_challs):
...
+            for index, (authzr, _) in authzrs_to_check.items():
+                authzrs[index] = authzr
+
+            # Handle failed authorizations: with best effort this is only a warning,
+            # otherwise an exception.
+            authzrs_failed = [authzr for authzr, _ in authzrs_to_check.values()
+                              if authzr.body.status == messages.STATUS_INVALID]
+            if authzrs_failed:
+                # Display report to the user
+                _report_failed_authzrs(authzrs_failed, self.account.key)
```
bmw: nit: If you look at the output generated by this function, it creates one formatted message with a header and a footer like:
```

   Domain: example.org
   Type:   unauthorized
   Detail: Invalid response from
   http://example.org/.well-known/acme-challenge/y3jq7cdUngIddjenKtN-6_613t1YUrvceS-WNCRBJRM
   [2606:2800:220:1:248:1893:25c8:1946]: "<!doctype
   html>\n<html>\n<head>\n    <title>Example Domain</title>\n\n
   <meta charset=\"utf-8\" />\n    <meta http-equiv=\"Content-type"

   Domain: www.example.org
   Type:   unauthorized
   Detail: Invalid response from
   http://www.example.org/.well-known/acme-challenge/Dtch-6y2bYGI_T4Wir-N0H5Oi08ySaYLzVB587CGn1c
   [2606:2800:220:1:248:1893:25c8:1946]: "<!doctype
   html>\n<html>\n<head>\n    <title>Example Domain</title>\n\n
   <meta charset=\"utf-8\" />\n    <meta http-equiv=\"Content-type"

```
In the `best_effort` case, it's possible we'll call this method multiple times leading to rather unnecessarily verbose and cluttered output.

While this is a fairly minor problem, I think we can fairly easily fix this. What do you think?

adferrand: Yes, it is a good idea. I accumulate now all failed authzrs in case of best_effort, and send one unique report when leaving the loop. I move back to a `break` instead of a `return` for leaving the loop, in order to let the report be displayed even if we reached the max polling retry.

certbot/auth_handler.py
adferrand: It is not exactly the same logic than before. The warning would be superfluous in two situations:
* without best effort, because in this case, you have an exception already
* with best effort without quiet, because you have a way better report just after

So it is useful only for quiet mode with best_effort=true. But if you ask for quiet, and choose an option where challenges failures are not critical, why would we display theses warnings ?

bmw: Despite it being the current behavior, you could certainly make an argument that these warnings shouldn't be displayed when using best_effort in quiet mode.

I think for me what feels like "correct" behavior depends on how you're using `--allow-subset-of-names`. The initial motivation behind the flag was damage mitigation. If you're running a server hosting multiple domains and you're failing challenges for some reason, obtain an updated cert for as many domains as possible rather than letting it expire for all domains in the cert.

If you're using the flag in this way, I think I'd expect to be clearly told when domains are being dropped. If, however, you're using it because you want to remove a domain from a cert, the warning is probably unnecessary. (I might argue that doing the latter is a bad way to do things though because you may drop more domains than you expect. In this case, explicitly saying what domains were dropped may still be helpful).

I'm certainly biased, but I personally prefer we continue to print the warning when using `--allow-subset-of-names`. I'm not that worried about it though so if you disagree, we can leave it out.

adferrand: I see your point. `--allow-subset-of-names` is more a mitigation than a feature. And `quiet` does not mean `silent`. It makes sense, in the context of a mitigation, to inform the user about the fact that this mitigation has been triggered.

I update the whole process of reporting like this:
* during the loop, log warn for each failed authzr received
* at the end of the loop, build the consolidated report, and show it except if `quiet`
* then if best_effort is not `true`, raise an exception about the fact that some challenges are failed

Then with this I think we provide sufficient information in all scenarios.

adferrand:
I have a failure on a test unrelated on this PR, on for Python 3 tests only (on 2.7 it is passing). I do not know what is going on.

bmw:
The test that failed is testing some particularly nasty code which uses global state. I'm surprised your PR is consistently reproducing the error but I'll write a quick patch which should hopefully fix it.

bmw:
See #6799.

adferrand:
OK @bmw, your fix is fixing my PR!

certbot/auth_handler.py
bmw: Another change that was made as part of this was now when `--allow-subset-of-names` is not used, we still continue to poll all authorizations even if we have failed some. I think this is a good change as I'd personally be slightly annoyed if I failed some challenges, fixed up my config for those domains, and then was told about additional failed challenges for other domains.

PR #6753, opened 2019-02-07
Fix atexit handlers

adferrand:
This PR fixes problems that have occured after #6667 integration on farm tests, and required the revert #6752.

This fix is simply to trigger the tearDown logic only for Windows:

PR #6769, opened 2019-02-14
Fix test sdists with atexit handlers

adferrand:
This PR superseeds #6753.

The failure of `test_sdists.py` with the changes of #6667 is not due to test failures. It is due to the size of the output. Indeed, an execution on my laptop is succeeding, but generates an enormous amount of output about non critical teardown errors, or exceptions generated and checked during the tests. My hypothesis here, as @bmw saw the test not executed and failed, but staled with the necessity to kill it, is that the AWS instance get too much output to handle.

The reason of the size of this output, is that `test_sdists.py` uses `setuptools`, through `python setup.py test`, and this does not capture output, like `pytest` would do.

Hopefully, it is possible to delegate from `setuptools` to `pytest`, and it is already done for `acme`. This PR applies this configuration to `certbot`, `certbot-apache` and `certbot-nginx`.

With this, output is now clean, and should let AWS instances successfully execute `test_sdists.py`.

Of course, this PR reapplies #6667.

### DIFF
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -3,14 +3,15 @@
 .. warning:: This module is not part of the public API.
 
 """
+import logging
 import multiprocessing
 import os
 import pkg_resources
 import shutil
+import stat
 import tempfile
 import unittest
 import sys
-import warnings
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
@@ -328,22 +329,22 @@ class TempDirTestCase(unittest.TestCase):
 
     def tearDown(self):
         """Execute after test"""
-        # On Windows we have various files which are not correctly closed at the time of tearDown.
-        # For know, we log them until a proper file close handling is written.
-        # Useful for development only, so no warning when we are on a CI process.
-        def onerror_handler(_, path, excinfo):
-            """On error handler"""
-            if not os.environ.get('APPVEYOR'): # pragma: no cover
-                message = ('Following error occurred when deleting the tempdir {0}'
-                           ' for path {1} during tearDown process: {2}'
-                           .format(self.tempdir, path, str(excinfo)))
-                warnings.warn(message)
-        shutil.rmtree(self.tempdir, onerror=onerror_handler)
+        # Cleanup opened resources after a test. This is usually done through atexit handlers in
+        # Certbot, but during tests, atexit will not run registered functions before tearDown is
+        # called and instead will run them right before the entire test process exits.
+        # It is a problem on Windows, that does not accept to clean resources before closing them.
+        logging.shutdown()
+        util._release_locks()  # pylint: disable=protected-access
+
+        def handle_rw_files(_, path, __):
+            """Handle read-only files, that will fail to be removed on Windows."""
+            os.chmod(path, stat.S_IWRITE)
+            os.remove(path)
+        shutil.rmtree(self.tempdir, onerror=handle_rw_files)
+
 
 class ConfigTestCase(TempDirTestCase):
-    """Test class which sets up a NamespaceConfig object.
-
-    """
+    """Test class which sets up a NamespaceConfig object."""
     def setUp(self):
         super(ConfigTestCase, self).setUp()
         self.config = configuration.NamespaceConfig(
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -335,6 +335,8 @@ class TempDirTestCase(unittest.TestCase):
         # called and instead will run them right before the entire test process exits.
         # It is a problem on Windows, that does not accept to clean resources before closing them.
         logging.shutdown()
+        # Remove logging handlers that have been closed so they won't be
+        # accidentally used in future tests.
         logging.getLogger().handlers = []
         util._release_locks()  # pylint: disable=protected-access
 
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -334,6 +334,7 @@ class TempDirTestCase(unittest.TestCase):
         # called and instead will run them right before the entire test process exits.
         # It is a problem on Windows, that does not accept to clean resources before closing them.
         logging.shutdown()
+        logging.getLogger().handlers = []
         util._release_locks()  # pylint: disable=protected-access
 
         def handle_rw_files(_, path, __):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
   warnings described at https://github.com/certbot/josepy/issues/13.
 * Apache plugin now respects CERTBOT_DOCS environment variable when adding
   command line defaults.
+* Tests execution for certbot, certbot-apache and certbot-nginx packages now relies on pytest.
 
 ### Fixed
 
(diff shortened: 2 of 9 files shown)


adferrand:
First correction was not sufficient: test output was captured by pytest, but lead to a segfault.

In appears that tests calling the certbot logging system, are adding each two new handlers in the global logging system, that are never cleaned up. Upon `logging.shutdown()` called at each tearDown as added to #6667, python tries to close all registered handlers, and all of them throw an exception because they are already closed.

It means that on the Nth test, 2N handlers needs to be closed. It also means that if 200 tests are adding two handlers, the complete test suite will try and fail a close action grossly 10000 times over 400 handlers. This is exhausting the memory of light runners like AWS instance.

The correction is to reinit the handlers list on each tearDown, to stop the growing of handlers.

After tests, this correction seems to fix the segfault observed on AWS.

certbot/tests/util.py
```diff
@@ -329,23 +330,23 @@ def setUp(self):
...
-                           ' for path {1} during tearDown process: {2}'
-                           .format(self.tempdir, path, str(excinfo)))
-                warnings.warn(message)
-        shutil.rmtree(self.tempdir, onerror=onerror_handler)
+        # Cleanup opened resources after a test. This is usually done through atexit handlers in
+        # Certbot, but during tests, atexit will not run registered functions before tearDown is
+        # called and instead will run them right before the entire test process exits.
+        # It is a problem on Windows, that does not accept to clean resources before closing them.
+        logging.shutdown()
+        logging.getLogger().handlers = []
```
bmw: Can you add a comment like:

```suggestion
        # Remove logging handlers that have been closed so they won't be
        # accidentally used in future tests.
        logging.getLogger().handlers = []
```

bmw:
I did some more digging here because the explanations we had so far for this error we were seeing didn't seem right to me. Calling `logging.shutdown()` when there are previously closed handlers should be fine looking at the `logging` module.

If you look at the output when we see these errors, a traceback looks like:
```
  File "/usr/lib/python2.7/logging/handlers.py", line 76, in emit
    if self.shouldRollover(record):
  File "/usr/lib/python2.7/logging/handlers.py", line 154, in shouldRollover
    self.stream = self._open()
  File "/usr/lib/python2.7/logging/__init__.py", line 943, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 2] No such file or directory: '/tmp/tmpOXRIWf/logs/letsencrypt.log'
Logged from file account.py, line 116 
```
What's happening on `account.py` line 116? https://github.com/certbot/certbot/blob/8f7b2801061bb36a9b16b4a82193af700574182d/certbot/account.py#L116

We're trying to use the handlers we already closed.

adferrand:
So merging the study from @bmw and me, here is what happened.

Each invocation of `certbot.logger.post_arg_parse_setup` create a file handler on `letsencrypt.log`. This function also set an atexit handler invoking `logger.shutdown()`, that have the effect to close all logger file handler not already closed at this point. This method is supposed to be called when a python process is close to exit, because it makes all logger unable to write new logs on any handler.

Before #6667 and this PR, for tests, the atexit handle would be triggered only at the end of the pytest process. It means that each test that launches `certbot.logger.post_arg_parse_setup` add a new file handler. These tests were typically connecting the file handler on a `letsencrypt.log` located in a temporary directory, and this directory and content was wipped out at each test tearDown. As a consequence, the file handles, not cleared from the logger, were accumulating in the logger, with all of them connected to a deleted file log, except the last one that was just created by the current test. Considering the number of tests concerned, there were ~300 file handler at the end of pytest execution.

One can see that, on prior #6667, by calling `print(logger.getLogger().handlers` on the `tearDown` of these tests, and see the array growing at each test execution.

Even if this represent a memory leak, this situation was not really a problem on Linux: because a file can be deleted before it is closed, it was only meaning that a given invocation of `logger.debug` for instance, during the tests, was written in 300 log files. The overhead is negligeable. On Windows however, the file handlers were failing because you cannot delete a file before it is closed.

It was one of the reason for #6667, that added a call to `logging.shutdown()` at each test tearDown, with the consequence to close all file handlers. At this point, Linux is not happy anymore. Any call to `logger.warn` will generate an error for each closed file handler. As a file handler is added for each test, the number of errors grows on each test, following an arithmetical suite divergence.

On `test_sdists.py`, that is using the bare setuptools test suite without output capturing, we can see the damages. The total output takes 216000 lines, and 23000 errors are generated. A decent machine can support this load, but a not a small AWS instance, that is crashing during the execution. Even with pytest, the captured output and the memory leak become so large that segfaults are generated.

On the current PR, the problem is solved, by resetting the file handlers array on the logging system on each test tearDown. So each fileHandler is properly closed, and removed from the stack. They do not participate anymore in the logging system, and can be garbage collected. Then we stay on always one file handler opened at any time, and tests can succeed on AWS instances.

For the record, here is all the places where the logging system is called and fail if there is still file handlers closed but not cleaned (extracted from the original huge output before correction):
```
Logged from file account.py, line 116
Logged from file account.py, line 178
Logged from file client.py, line 166
Logged from file client.py, line 295
Logged from file client.py, line 415
Logged from file client.py, line 422
Logged from file client.py, line 480
Logged from file client.py, line 503
Logged from file client.py, line 540
Logged from file client.py, line 601
Logged from file client.py, line 622
Logged from file client.py, line 750
Logged from file cli.py, line 220
Logged from file cli.py, line 226
Logged from file crypto_util.py, line 101
Logged from file crypto_util.py, line 127
Logged from file crypto_util.py, line 147
Logged from file crypto_util.py, line 261
Logged from file crypto_util.py, line 283
Logged from file crypto_util.py, line 307
Logged from file crypto_util.py, line 336
Logged from file disco.py, line 116
Logged from file disco.py, line 124
Logged from file disco.py, line 134
Logged from file disco.py, line 138
Logged from file disco.py, line 141
Logged from file dns_common_lexicon.py, line 45
Logged from file dns_common_lexicon.py, line 61
Logged from file dns_common_lexicon.py, line 67
Logged from file dns_common.py, line 316
Logged from file dns_common.py, line 64
Logged from file eff.py, line 60
Logged from file eff.py, line 73
Logged from file error_handler.py, line 105
Logged from file error_handler.py, line 110
Logged from file error_handler.py, line 87
Logged from file hooks.py, line 248
Logged from file main.py, line 1071
Logged from file main.py, line 1075
Logged from file main.py, line 1189
Logged from file ops.py, line 122
Logged from file ops.py, line 325
Logged from file ops.py, line 338
Logged from file reporter.py, line 55
Logged from file selection.py, line 110
Logged from file selection.py, line 118
Logged from file selection.py, line 123
Logged from file selection.py, line 176
Logged from file selection.py, line 231
Logged from file selection.py, line 310
Logged from file selection.py, line 66
Logged from file standalone.py, line 101
Logged from file standalone.py, line 88
Logged from file standalone.py, line 97
Logged from file standalone.py, line 98
Logged from file storage.py, line 52
Logged from file storage.py, line 59
Logged from file storage.py, line 75
Logged from file util.py, line 56
Logged from file webroot.py, line 165
Logged from file webroot.py, line 186
Logged from file webroot.py, line 187
Logged from file webroot.py, line 204
Logged from file webroot.py, line 223
Logged from file webroot.py, line 234
Logged from file webroot.py, line 235
Logged from file webroot.py, line 237
Logged from file webroot.py, line 91
```

adferrand:
@bmw I update the comments + changelog.

Issue #6499, opened 2018-11-13
pipstrap change causing failures in older OSes


bmw:
This change hasn't been included in a release yet so no users are seeing failures, however, it's causing our some of our nightly tests to fail for older OSes which can be seen at https://travis-ci.org/certbot/certbot/builds/454634684?utm_source=github_status&utm_medium=notification.

@adferrand, are you able to take a look at this?

PR #6775, opened 2019-02-16
[Windows] Fix pipstrap

adferrand:
So pipstrap is broken on Windows. This impact is limited, because pipstrap is only used during the dev environment setup in Windows (`py tools\venv3.py`). The breakage was invisible before #6691, because pipstrap was doing nothing if the pip version in a new venv was sufficiently new, and this was the case for the supported Python versions for Windows.

Now pipstrap is effectively pin `pip`, `setuptools` and `wheel` at each execution.

The breakage is due to two things:
1) One of the command is invoked with paths that are wrapped using single quotes, to handle paths with spaces. However, single quotes are not a valid identifier for strings on Windows. So `'test'` is the string `'test'`, not `test`
2) Because on Windows you cannot delete a file before all ressources on it are closed, `pip` cannot self-update. You must invoke explicitly the python interpreter, like with `python -m pip --upgrade pip`.

First problem is solved by using the recommended way to execute subprocceses, with an array command out of a shell, that make unecessary to wrap parameters with spaces, like paths, between quotes.

Second problem is solved specifically for Windows: on this platform, `python -m pip install pip.tar.gz` is used instead of `pip install pip.tar.gz`.

Some discussions on the second problem resolution. I did not make a change to use globally (Linux+Windows) `python -m pip ...`, despite it is the recommended way to upgrade pip, because of some concerns explained by @erikrose in mozilla/pipstrap#21. But it was in the scope of pipstrap as a general purpose tool, and since then, we vendored our own version for Certbot specific needs. And for Certbot, pipstrap is supposed to be used in an enabled venv to fix its basic bricks. It is expected in a venv to have a `python` executable in the PATH, representing the interpreter available for this particular venv. So for this case `python -m pip ...` is a safe execution in my opinion.

I propose on top of the current PR changes, to use `python -m pip ...` unconditionnally in pipstrap.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -1219,9 +1219,8 @@ anything goes wrong, it will exit with a non-zero status code.
 from __future__ import print_function
 from distutils.version import StrictVersion
 from hashlib import sha256
-from os import environ, name
+from os import environ
 from os.path import join
-from pipes import quote
 from shutil import rmtree
 try:
     from subprocess import check_output
@@ -1241,7 +1240,7 @@ except ImportError:
                 cmd = popenargs[0]
             raise CalledProcessError(retcode, cmd)
         return output
-from sys import exit, version_info, executable as python_executable
+import sys
 from tempfile import mkdtemp
 try:
     from urllib2 import build_opener, HTTPHandler, HTTPSHandler
@@ -1263,7 +1262,7 @@ maybe_argparse = (
     [('18/dd/e617cfc3f6210ae183374cd9f6a26b20514bbb5a792af97949c5aacddf0f/'
       'argparse-1.4.0.tar.gz',
       '62b089a55be1d8949cd2bc7e0df0bddb9e028faefc8c32038cc84862aefdd6e4')]
-    if version_info < (2, 7, 0) else [])
+    if sys.version_info < (2, 7, 0) else [])
 
 
 PACKAGES = maybe_argparse + [
@@ -1344,8 +1343,8 @@ def get_index_base():
 
 
 def main():
-    executable = python_executable or 'python'
-    pip_version = StrictVersion(check_output([executable, '-m', 'pip', '--version'])
+    python = sys.executable or 'python'
+    pip_version = StrictVersion(check_output([python, '-m', 'pip', '--version'])
                                 .decode('utf-8').split()[1])
     has_pip_cache = pip_version >= StrictVersion('6.0')
     index_base = get_index_base()
@@ -1356,7 +1355,7 @@ def main():
                                      digest)
                      for path, digest in PACKAGES]
         # On Windows, pip self-upgrade is not possible, it must be done through python interpreter.
-        command = [executable, '-m', 'pip', 'install', '--no-index', '--no-deps', '-U']
+        command = [python, '-m', 'pip', 'install', '--no-index', '--no-deps', '-U']
         # Disable cache since it is not used and it otherwise sometimes throws permission warnings:
         command.extend(['--no-cache-dir'] if has_pip_cache else [])
         command.extend(downloads)
@@ -1373,7 +1372,7 @@ def main():
 
 
 if __name__ == '__main__':
-    exit(main())
+    sys.exit(main())
 
 UNLIKELY_EOF
     # -------------------------------------------------------------------------
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -1241,7 +1241,7 @@ except ImportError:
                 cmd = popenargs[0]
             raise CalledProcessError(retcode, cmd)
         return output
-from sys import exit, version_info
+from sys import exit, version_info, executable as python_executable
 from tempfile import mkdtemp
 try:
     from urllib2 import build_opener, HTTPHandler, HTTPSHandler
@@ -1344,7 +1344,8 @@ def get_index_base():
 
 
 def main():
-    pip_version = StrictVersion(check_output(['pip', '--version'])
+    executable = python_executable or 'python'
+    pip_version = StrictVersion(check_output([executable, '-m', 'pip', '--version'])
                                 .decode('utf-8').split()[1])
     has_pip_cache = pip_version >= StrictVersion('6.0')
     index_base = get_index_base()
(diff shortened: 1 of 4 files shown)


bmw:
Do you remember what failed when we just changed it to be unconditionally `python -m pip ...`? I'm curious if we can get away with keeping things simple and always invoking pip that way. It's been supported since September 2012 (see https://pip.pypa.io/en/stable/news/#id132).

The issue was #6499, but I think I kicked the build after we fixed it so we lost the log. We've removed tests on some EOL'd OSes since then though.

erikrose:
For my reference, it's pip 1.2 that added support for `python -m pip`. Interestly, as late as 7.0.2 (and maybe to the present day, for all I know), they were recommending `python -m pip` only on Windows for self-updating. I'm not sure why.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -156,12 +156,13 @@ def main():
...
-                     # sometimes throws permission warnings:
-                     ('--no-cache-dir ' if has_pip_cache else '') +
-                     ' '.join(quote(d) for d in downloads),
-                     shell=True)
+        # On Windows, pip self-upgrade is not possible, it must be done through python interpreter.
+        command = ['pip'] if name != 'nt' else ['python', '-m', 'pip']
+        command.extend(['install', '--no-index', '--no-deps', '-U'])
+        # Disable cache since it is not used and it otherwise sometimes throws permission warnings:
+        command.extend(['--no-cache-dir'] if has_pip_cache else [])
+        command.extend(downloads)
```
erikrose: I certainly like using the array syntax better than glomming a string together. I like it so much (and always have) that I wonder whether I had a good reason for doing it the other way. Let me check the commit logs.

erikrose: Nope, I did it the worse way from the beginning! Glad you changed it.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -156,12 +156,13 @@ def main():
...
                                      digest)
                      for path, digest in PACKAGES]
-        check_output('pip install --no-index --no-deps -U ' +
-                     # Disable cache since we're not using it and it otherwise
-                     # sometimes throws permission warnings:
-                     ('--no-cache-dir ' if has_pip_cache else '') +
-                     ' '.join(quote(d) for d in downloads),
-                     shell=True)
+        # On Windows, pip self-upgrade is not possible, it must be done through python interpreter.
+        command = ['pip'] if name != 'nt' else ['python', '-m', 'pip']
```
erikrose: I think we can move to `python -m pip` for all OSes. pip 1.2, which is the first that supported that invocation style, came out 7 years ago. Let's try that and make sure we run test_everything before we release, just to be sure we aren't supporting some extremely conservative OS that still ships an ancient pip.

Also, if we use `sys.executable` instead of hard-coding "python", it should work all the time, on every platform, venvs or not, even if somebody calls their interpreter something odd. Actually, let's do `sys.executable or 'python'`, because `executable` can be `None` or `''` in some corner cases.

letsencrypt-auto-source/pieces/pipstrap.py
adferrand: Your welcome!

letsencrypt-auto-source/pieces/pipstrap.py
adferrand: Yes, it was my approach in the previous iteration several months ago. But we did a fallback, as @bmw said, because the test-everything detected a regression if nearly EOL Linux distributions shipping really old pip versions.

However, we removed several of these distributions since then. So it may be actually working!

I will make the update here, and start an equivalent of test-everything.

adferrand:
For the record, I think `python -m pip` became very quickly the official approach to self-upgrade pip on Windows, because of my so beloved "YOU SHALL NOT DELETE THIS FILE BEFORE CLOSING IT" Windows specificity about file handlers. And obviously, an active process that wants to move himself and all its resources during a self-update, may have a very bad time with this platform ^^

adferrand:
And so @bmw here is my response about your comment: yes, it is test-everything that broke with a particular OS distribution, but I could not remember which one. Anyway, we will know if the approach is viable in few minutes here.

adferrand:
Allright. The test-everything approach is passing with @erikrose proposition.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -1219,7 +1219,7 @@ anything goes wrong, it will exit with a non-zero status code.
 from __future__ import print_function
 from distutils.version import StrictVersion
 from hashlib import sha256
-from os import environ
+from os import environ, name
```
erikrose: No longer need `name`

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -1344,7 +1344,8 @@ def get_index_base():
 
 
 def main():
-    pip_version = StrictVersion(check_output(['pip', '--version'])
+    executable = python_executable or 'python'
```
erikrose: The meaning might be clearer if we shallow-import `executable` and call the result "python", making this `python = sys.executable or 'python'`. Idiomatic Python readers familiar with the stdlib will readily identify sys.executable, and `python` will stand in for ".../python", its usual value.

letsencrypt-auto-source/letsencrypt-auto
adferrand: Cleaned. And also `quotes`.

letsencrypt-auto-source/letsencrypt-auto
adferrand: Good idea. I did that.

letsencrypt-auto-source/pieces/pipstrap.py
adferrand: test-everything worked successfully

letsencrypt-auto-source/letsencrypt-auto
erikrose: Oops, forgot to rebuild le-auto

adferrand:
Yes I launched it in parallel of you response. Thanks for the review @erikrose!

adferrand:
Oh yes, can you approve, to unblock the merging process?

letsencrypt-auto-source/letsencrypt-auto
adferrand: Done.

letsencrypt-auto-source/pieces/pipstrap.py
```diff
@@ -43,7 +42,7 @@ def check_output(*popenargs, **kwargs):
                 cmd = popenargs[0]
             raise CalledProcessError(retcode, cmd)
         return output
-from sys import exit, version_info
+import sys
```
erikrose: Just in case you didn't know, it's perfectly possible to import some things deep from a module and other things shallow.

adferrand: Yes it was just to save a line, but if you think it is a problem, as for instance by exposing the sys module as part of this module could be a problem, I can stay with a deep import for the other members.

erikrose: I mentioned it just because it made you change lines of code that otherwise wouldn't have needed to change. As for "security", it makes no difference whether you expose `sys` directly or not, as you can get to it by following the `__module__` attribute of anything defined in `sys`.

adferrand: Thanks!

PR #6778, opened 2019-02-17
[Windows] Refactor lock_and_call using queues

adferrand:
The PR #6772 created a new implementation that was not broken for Windows, using files to coordinate the parallel processes. However, I observed that sometimes, on AppVeyor for Windows with Python 3.5, lock_and_call is deadlocking, blocking the entire Windows build until the global one hour timeout is reached.

This does not impact Linux on Travis CI however. And I cannot reproduce it in a local development Windows machine.

So I reimplemented lock_and_call with the recommended way: using together multiprocessing.Process, and multiprocessing.Queue to coordinate the parallel processes. This implementation works both on Linux and Windows.

This implementation is also protected against deadlock, throwing exceptions when timeouts are reached.

### DIFF
--- a/certbot/tests/util.py
+++ b/certbot/tests/util.py
@@ -10,8 +10,7 @@ import tempfile
 import unittest
 import sys
 import warnings
-import subprocess
-import time
+from multiprocessing import Process, Queue
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
@@ -25,6 +24,7 @@ from certbot import constants
 from certbot import interfaces
 from certbot import storage
 from certbot import configuration
+from certbot import lock
 from certbot import util
 
 from certbot.display import util as display_util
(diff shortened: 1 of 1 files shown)


bmw:
I proposed an alternate approach to this in #6783.

certbot/tests/util.py
```diff
@@ -361,56 +361,44 @@ def setUp(self):
         self.config.server = "https://example.com"
 
 
+def _handle_lock(queue_in, queue_out, path):
```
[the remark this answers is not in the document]
adferrand: On current implementation, using queue is overkill compared to queue. It was because before, I was checking for specific signal values during the coordination (START/LOCKED/STOP), but it was also overkill for a test method.

I move to an event, and add back docstring.

certbot/tests/util.py
```diff
@@ -361,56 +361,44 @@ def setUp(self):
...
+    queue_in.get(timeout=10)
```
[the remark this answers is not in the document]
adferrand: Same reason than above. Yes, only waiting for the worker to signal file is locked is sufficient.

certbot/tests/util.py
```diff
@@ -361,56 +361,44 @@ def setUp(self):
...
-            callback()
-        finally:
-            # This will trigger the lock release in subprocess.
-            os.remove(trigger)
-            process.communicate()
-        assert process.returncode == 0
-    finally:
-        shutil.rmtree(workspace)
+    emit_queue = Queue()  # type: ignore
+    receive_queue = Queue()  # type: ignore
```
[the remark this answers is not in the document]
adferrand: With queues, mypy was not satisfied. I suppose it is because `queue` module is not available in Python 2.7, so `multiprocessing` module needs to make some dynamic operations to expose it. But it is not the case with `threading.Event`, so it is not needed anymore.

adferrand:
As a general comment, this is working both on my personal computer and on AppVeyor. But not on my professional machine. Maybe a story of baaaaaaad antivirus (thanks McAfee). Anyway, the behavior is not perfectly consistent, and as said more informaly: we should find a way to remove `lock_and_call` that would clean the code and provide a consistent behavior.

I think it is possible. More on this on a future PR.

certbot/tests/util.py
```diff
@@ -360,57 +360,63 @@ def setUp(self):
...
+    :param multiprocessing.Event event_out: event object to signal when the lock is acquired
+    :param path: the path to lock
+    """
+    if os.path.isdir(path):
+        my_lock = lock.lock_dir(path)
+    else:
+        my_lock = lock.LockFile(path)
+    try:
+        event_out.set()
+        event_in.wait(timeout=20)
```
bmw: I think we have to check the return value here. See https://docs.python.org/3.7/library/threading.html#threading.Event.wait.

I think this can be as simple as adding an `assert`.

certbot/tests/util.py
```diff
@@ -361,56 +361,49 @@ def setUp(self):
...
-        assert process.returncode == 0
-    finally:
-        shutil.rmtree(workspace)
+    emit_event = Event()
+    receive_event = Event()
+    process = Process(target=_handle_lock, args=(emit_event, receive_event, path_to_lock))
+    process.start()
+
+    # Wait confirmation that lock is acquired
+    receive_event.wait(timeout=10)
```
bmw: I think we have to check the return value here. See https://docs.python.org/3.7/library/threading.html#threading.Event.wait.

I think this can be as simple as adding an `assert`.

certbot/tests/util.py
adferrand: Only available since 3.1, before that it returns nothing. It is true that if the timeout is reached, there is no exception. But the callback is likely to fail in this case. 

If you think it is a problem, we should use queues again. They throw exception when timeout is reached.

certbot/tests/util.py
adferrand: See above.

certbot/tests/util.py
bmw: It was [added in 2.7](https://docs.python.org/2.7/library/threading.html#threading.Event.wait) and 3.1. Since we only support 2.7 and 3.4+ this seems OK to me?

I do not think we should allow tests to potentially continue with the lock not properly obtained/released.

adferrand: I did not saw it for 2.7. In this case, yes it is OK. I implement that.

adferrand: Fixed.

certbot/tests/util.py
adferrand: Fixed.

PR #6783, opened 2019-02-19
Revert to old locking code and reenable tests

bmw:
#6772 started causing our tests to sometimes deadlock on Windows.

#6778 proposes to fix this by partially reverting #6772 and adding two queues that pass around dummy objects, but instead, why not switch back to our previous code which simply uses a single condition variable for synchronization? I'd argue that this approach is a little simpler and it's been tested longer since it was the previous behavior.

I think that the broken tests being reenabled here were fixed in an earlier PR working on locking on Windows.

@adferrand, are you able to review this?

### DIFF
--- a/certbot/tests/lock_test.py
+++ b/certbot/tests/lock_test.py
@@ -23,7 +23,6 @@ class LockDirTest(test_util.TempDirTestCase):
         from certbot.lock import lock_dir
         return lock_dir(*args, **kwargs)
 
-    @test_util.broken_on_windows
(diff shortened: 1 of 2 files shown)


bmw:
Unfortunately, adding a timeout to all blocking code isn't possible with this approach. `Condition.wait()` doesn't give you a way to distinguish between a timeout and another thread calling `notify()` in Python 2.

Adding this timeout is valuable so I'm closing this PR.

Issue #6606, opened 2018-12-14
Self signed certificate continues to be served with Apache on CentOS 6


bmw:
I've heard about two reports of the Apache plugin thinking it obtained and installed a certificate for Apache on CentOS 6, but when the user visits their domain, Apache's self signed certificate is served. This is a pretty serious problem and should be fixed.

The one report I've seen myself is at https://community.letsencrypt.org/t/connection-not-secure-on-centos-6/80054. @schoen has seen more.

We're thinking this probably has something to do with vhost precedence rules, especially on CentOS 6 where if I remember correctly, [NameVirtualHost](https://httpd.apache.org/docs/2.2/mod/core.html#namevirtualhost) is disabled by default.

We should definitely try and get a configuration that reproduces this problem as it will make debugging much easier.

bmw:
I'm moving this to the current sprint because I think its one of our more important tasks and it makes it more visible.

schoen:
I tried testing this today and I found that it was quite easy to reproduce (fortunately and unfortunately, we might say).

I installed CentOS 6.9 on a new VM, installed Apache, and created a single HTTP virtual host in `/etc/httpd/conf.d` for a name which was pointed at the CentOS machine. The machine didn't immediately support HTTPS at all.

When `certbot-auto` installed packages including `mod_ssl`, the `/etc/httpd/conf.d/ssl.conf` was added. The `certbot-auto` process then obtained a cert and created an `-le-ssl` variant of my existing HTTP virtual host.

However, the new cert was not actually used; instead, the default system cert specified with `<VirtualHost _default_:443>` in `ssl.conf` continued to be used. :slightly_frowning_face: That virtual host (just as in a number of users' similar experiences) seems to override the more-specific one in the Certbot-created `-le-ssl` file!

This is a relatively plausible sequence of events for a new user on CentOS 6.9, so I guess this bug is pretty easy to run across.

joohoi:
I have looked into this as well, and the CentOS 6 HTTPS VirtualHost precedence seems to be decided by the `Include` order, and the _**last**_ one taking over the control. One of the common configuration structure that users of CentOS have is having all `VirtualHost`s in the main `httpd.conf` file. In this case everything works as expected, as the HTTPS configuration file created by Certbot gets `Include`d in the end of `httpd.conf`. 

[...]

PR #6784, opened 2019-02-19
Fix CentOS 6 installer issue

joohoi:
In CentOS 6 default httpd configuration, the `LoadModule ssl_module ...` is handled in `conf.d/ssl.conf`. As the `VirtualHost` configuration files in `conf.d/` are loaded in alphabetical order, this means that all files that have `<IfModule mod_ssl.c>` and are loaded before `ssl.conf` are effectively ignored. This PR moves the `LoadModule ssl_module` to the main `httpd.conf` while leaving a conditional `LoadModule` directive in `ssl.conf`.

Features
 - Reads the module configuration from `ssl.conf` in case some modifications to paths have been made by the user.
 - Falls back to default paths if the directive doesn't exist.
 - Moves the `LoadModule` directive in `ssl.conf` inside `<IfModule !mod_ssl.c>` to avoid printing warning messages of duplicate module loads.
 - Adds `LoadModule ssl_module` inside of `<IfModule !mod_ssl.c>` to the top of the main `httpd.conf`.
 - Ensures that these modifications are not made multiple times.

Fixes: #6606


certbot-apache/certbot_apache/tests/centos6_test.py
```diff
@@ -0,0 +1,125 @@
...
+            os.path.join(aug_pre, "ssl.conf/VirtualHost"),
+            set([obj.Addr.fromstring("_default_:443")]),
+            True, True, None)
+    ]
+    return vh_truth
+
+class CentOS6Tests(util.ApacheTest):
+    """Tests for CentOS 6"""
+
+    _multiprocess_can_split_ = True
```
bmw: Now that we've switched to pytest from nose, we don't need to set values like this.

certbot-apache/certbot_apache/tests/centos6_test.py
```diff
@@ -0,0 +1,125 @@
...
+            "example/cert_chain.pem", "example/fullchain.pem")
+        self.config.save()
+
+        post_loadmods = self.config.parser.find_dir(
+            "LoadModule", "ssl_module", exclude=False)
+
+        # We should now have LoadModule ssl_module in root conf and ssl.conf
+        self.assertEqual(len(post_loadmods), 2)
+        for lm in post_loadmods:
+            arguments = self.config.parser.get_all_args(lm[:-7])
```
bmw: Can you add a comment about what we're stripping off the end of `lm` here?

Alternatively, maybe we could make a simple function with a name that does this if we're doing it in multiple places.

certbot-apache/certbot_apache/tests/centos6_test.py
```diff
@@ -0,0 +1,125 @@
...
+        root_loadmods = self.config.parser.find_dir(
+            "LoadModule", "ssl_module",
+            start=parser.get_aug_path(self.config.parser.loc["default"]),
+            exclude=False)
+
+        mods = [lm for lm in root_loadmods if self.config.parser.loc["default"] in lm]
+
+        self.assertEqual(len(mods), 1)
+        self.assertEqual(
+            self.config.parser.get_all_args(mods[0][:-7]),
```
bmw: Can you add a comment about what we're stripping off the end of `mods[0]` here?

Alternatively, maybe we could make a simple function with a name that does this if we're doing it in multiple places.

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -47,6 +47,60 @@ def get_parser(self):
...
+        """
+        Add "LoadModule ssl_module <pre-existing path>" to main httpd.conf if
+        it doesn't exist there already.
+        """
+
+        loadmods = self.parser.find_dir("LoadModule", "ssl_module", exclude=False)
+
+        # We should have one in ssl.conf
+        sslconf_loadmod_path = None
+        loadmod_args = ["ssl_module", "modules/mod_ssl.so"]
```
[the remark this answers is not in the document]
bmw: We talked about a complication to doing what I described here that we do not properly parse `IfModule !<module>` statements. I do not think we should try and resolve this issue now, but after thinking about this more, I think I personally have the same opinion I did before. My reasons for this are.

1. `certbot --help apache` says `--apache-handle-modules` is `False`.
2. To date, we have never automatically enabled `mod_ssl` on user's systems and as far as I know, there have been no requests for us to do so.
3. I think a small subset of people may want to user alternative ssl modules such as those compiled against libraries other than openssl.

For these reasons, I think we should search for `LoadModule ssl_module` directives with `exclude=False` and if the path to the SSL module isn't shared between all directives, error out with a clear error message for what the user can do to fix the problem themselves.

What do you think?

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -47,6 +47,60 @@ def get_parser(self):
...
+
+        for m in loadmods:
+            if "/conf.d/ssl.conf/" in m:
+                # Strip "arg[1]" off from the end
+                sslconf_loadmod_path = m[:-7]
+                # Use the preconfigured LoadModule values from ssl.conf
+                loadmod_args = self.parser.get_all_args(sslconf_loadmod_path)
+            elif self.parser.loc["default"] in m:
+                return
```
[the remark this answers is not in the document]
joohoi: As we spoke yesterday on call, we don't have a reliable way to detect this, and I wouldn't think that this would affect too many users.

certbot-apache/certbot_apache/override_centos.py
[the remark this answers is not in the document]
joohoi: If multiple `LoadModule ssl_module ...` directives with different arguments are found across the configuration, we now print an info message telling the user to do the change manually if they encounter issues with the configuration.

certbot-apache/certbot_apache/parser.py
```diff
@@ -458,6 +468,24 @@ def find_dir(self, directive, arg=None, start=None, exclude=True):
 
         return ordered_matches
 
+    def get_all_args(self, match):
+        """Tries to fetch all arguments for a directive. See get_arg."""
+
+        if match[-1] != "/":
+            match = match+"/"
+        results = []
```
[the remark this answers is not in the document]
joohoi: Yeah, that's much more elegant. It will, however when run against a parent node (for example: `.../IfModule[1]/`) return the child directive names as well. I made sure to include a note about this in the pydoc.

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -66,40 +72,50 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
-                # Strip "arg[1]" off from the end
-                sslconf_loadmod_path = m[:-7]
-                # Use the preconfigured LoadModule values from ssl.conf
-                loadmod_args = self.parser.get_all_args(sslconf_loadmod_path)
-            elif self.parser.loc["default"] in m:
+            noarg_path = m.rpartition("/")[0]
+            path_args = self.parser.get_all_args(noarg_path)
+            if loadmod_args:
+                if loadmod_args != path_args:
+                    logger.info("Multiple different LoadModule directives for mod_ssl "
```
bmw: I slightly misunderstood the changes you described on the call earlier this week. What do you think of the following pseudocode:

```
for m in loadmods:
    if m is in config root:
        return
for m in loadmods:
    if loadmod_args is different:
        raise "Multiple different LoadModule..."
add "LoadModule ssl_module" with loadmod_args to the top of the config root
```

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -66,40 +72,50 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
+                                "were found. If you encounter issues with resulting "
+                                "configuration, it's suggested to move the LoadModule "
+                                "ssl_module directive to the beginning of main httpd.conf.")
```
bmw: nit: Can we include the path to this file in the message?

certbot-apache/certbot_apache/tests/centos6_test.py
```diff
@@ -116,10 +117,51 @@ def test_loadmod_rootconf_exists(self):
...
+        orig_loadmods = self.config.parser.find_dir("LoadModule",
+                                                    "ssl_module",
+                                                    exclude=False)
+        for mod in orig_loadmods:
+            noarg_path = mod.rpartition("/")[0]
+            self.config.aug.remove(noarg_path)
+        self.config.save()
+        # get_all_args() is called for each LoadModule that was found
+        getall = "certbot_apache.override_centos.CentOSParser.get_all_args"
+        with mock.patch(getall) as mock_getall:
```
bmw: nit: Instead of or in addition to mocking out this function, could we test that we didn't add the directives to the main configuration file?

certbot-apache/certbot_apache/override_centos.py
joohoi: I'm not sure if we should raise an exception here. If the user has different `LoadModule ssl_module` directives in their configuration, I would believe they have manually tuned the configuration, and most likely at that situation know what they are doing.

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -47,6 +53,71 @@ def get_parser(self):
...
+
+        # Wrap LoadModule mod_ssl inside of <IfModule !mod_ssl.c> if it's not
+        # configured like this already.
+        if loadmod_path and "ifmodule" not in loadmod_path.lower():
+            sslconf_path = loadmod_path.split("/directive")[0]
+            # Remove the old LoadModule directive
+            self.aug.remove(loadmod_path)
+
+            # Create a new IfModule !mod_ssl.c
+            ssl_ifmod = self.parser.get_ifmod(sslconf_path, "!mod_ssl.c", beginning=True)
```
bmw: This is certainly a corner case, but I encountered a minor issue with the code here. In `ssl.conf`, I had:
```
LoadModule ssl_module modules/mod_ssl.so

<IfModule !mod_ssl.c>
LoadModule ssl_module modules/mod_ssl.so
</IfModule>
```
When this code was run, it resulted in:
```
<IfModule !mod_ssl.c>
LoadModule ssl_module modules/mod_ssl.so
LoadModule ssl_module modules/mod_ssl.so
</IfModule>
```
Can we always add a new `IfModule` block specifically for the unwrapped `LoadModule `directive?

joohoi: I changed this in a way that paths that already have an existing `<IfModule !mod_ssl.c>` `LoadModule ssl_module ...` under them, don't get a new `LoadModule` added within. We basically care about having the `ssl_module` loaded in the main `httpd.conf`, and about not confusing the user too much.

joohoi:
Ok, this is ready for another round of review. We now throw an exception with a new error message instructing the user to remove the extra (non-matching) `LoadModule ssl_module` directives from their configuration.

If a `LoadModule` directive exists on a same level than one that's already within `<IfModule !mod_ssl.c>` block, remove the old one, but don't add new.

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -72,51 +74,66 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
-            # Create a new IfModule !mod_ssl.c
-            ssl_ifmod = self.parser.get_ifmod(sslconf_path, "!mod_ssl.c", beginning=True)
-
-            self.parser.add_dir(ssl_ifmod[:-1], "LoadModule", loadmod_args)
-            self.save_notes += ("Wrapped pre-existing LoadModule ssl_module "
-                                "inside of <IfModule !mod_ssl> block.\n")
+        for loadmod_path in loadmod_paths:
+            nodir_path = loadmod_path.split("/directive")[0]
+            # Only create if !mod_ssl.c ifmodule not in path
+            if not self.parser.not_modssl_ifmodule(loadmod_path):  # pylint: disable=no-member
```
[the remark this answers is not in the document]
bmw: Didn't we effectively already check this above? If so, can we save the result and only do this for paths that we know aren't in an `ifmodule !mod_ssl.c` block?

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -47,6 +55,86 @@ def get_parser(self):
...
+                loadmod_args = path_args
+
+            loadmod_paths.append(noarg_path)
+
+            if self.parser.not_modssl_ifmodule(noarg_path):  # pylint: disable=no-member
+                # Populate the list of known !mod_ssl.c IfModules
+                nodir_path = noarg_path.rpartition("/directive")[0]
+                correct_ifmods.append(nodir_path)
+
+            if self.parser.loc["default"] in noarg_path:
```
bmw: nit: Can we indent this block and remove the `if self.parser.not_modssl_ifmodule(noarg_path)` call below?

certbot-apache/certbot_apache/tests/centos6_test.py
```diff
@@ -95,6 +94,26 @@ def test_loadmod_default(self):
...
+        self.assertFalse(any(["ifmodule" in m.lower() for m in pre_loadmods]))
+        self.config.assoc["test.example.com"] = self.vh_truth[0]
+        self.config.deploy_cert(
+            "random.demo", "example/cert.pem", "example/key.pem",
+            "example/cert_chain.pem", "example/fullchain.pem")
+        post_loadmods = self.config.parser.find_dir(
+            "LoadModule", "ssl_module", exclude=False)
+
+        for mod in post_loadmods:
+            self.assertTrue(self.config.parser.not_modssl_ifmodule(mod))  #pylint: disable=no-member
```
bmw: Did you check if adding a line before this like:
```
assert isinstance(self.config.parser, CentOSParser)
```
satisfies `pylint`?

certbot-apache/certbot_apache/override_centos.py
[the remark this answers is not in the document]
joohoi: assert unfortunately didn't help with the pylint issue.

I was able to clean up the function a bit, in order to not to do redundant checks for the `not_modssl_ifmodule`.

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -72,51 +74,66 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
+                           "for LoadModule ssl_module. Please remove or comment out the "
+                           "one(s) that are not in use.")
+                    raise ConfigurationError(msg)
             else:
                 loadmod_args = path_args
-                loadmod_path = noarg_path
+
+            loadmod_paths.append(noarg_path)
+
+            if self.parser.not_modssl_ifmodule(noarg_path):  # pylint: disable=no-member
```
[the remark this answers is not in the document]
joohoi: Unfortunately it doesn't :/

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -72,51 +74,66 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
+                # Populate the list of known !mod_ssl.c IfModules
+                nodir_path = noarg_path.rpartition("/directive")[0]
+                correct_ifmods.append(nodir_path)
+
             if self.parser.loc["default"] in noarg_path:
-                # LoadModule already in the main configuration file, NOOP
-                return
+                if self.parser.not_modssl_ifmodule(noarg_path):  #pylint: disable=no-member
```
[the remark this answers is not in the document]
joohoi: Unfortunately it doesn't :/

certbot-apache/certbot_apache/tests/centos6_test.py
joohoi: Unfortunately it doesn't :/

certbot-apache/certbot_apache/parser.py
```diff
@@ -297,7 +297,7 @@ def add_dir_to_ifmodssl(self, aug_conf_path, directive, args):
             for i, arg in enumerate(args):
                 self.aug.set("%s/arg[%d]" % (nvh_path, i + 1), arg)
 
-    def get_ifmod(self, aug_conf_path, mod, beginning=False):
+    def get_ifmod(self, aug_conf_path, mod, beginning=False, force_create=False):
```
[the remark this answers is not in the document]
joohoi: This is now split to two separate functions

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -72,51 +74,66 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
+                    # LoadModule already in the main configuration file
+                    if "ifmodule/" in noarg_path or "ifmodule[1]" in noarg_path.lower():
+                        # It's the first or only IfModule in the file
+                        return
```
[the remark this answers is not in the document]
joohoi: Yeah, I'd leave it as it is :)

joohoi:
This is now ready for the next round of review. Do we have some way to kick AppVeyor the build seems to have failed for unrelated issue (error in temp directory handling in AppVeyor).

certbot-apache/certbot_apache/override_centos.py
```diff
@@ -72,51 +74,66 @@ def _deploy_loadmodule_ssl_if_needed(self):
...
+                # Populate the list of known !mod_ssl.c IfModules
+                nodir_path = noarg_path.rpartition("/directive")[0]
+                correct_ifmods.append(nodir_path)
+
             if self.parser.loc["default"] in noarg_path:
-                # LoadModule already in the main configuration file, NOOP
-                return
+                if self.parser.not_modssl_ifmodule(noarg_path):  #pylint: disable=no-member
+                    # LoadModule already in the main configuration file
+                    if "ifmodule/" in noarg_path or "ifmodule[1]" in noarg_path.lower():
```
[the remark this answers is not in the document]
bmw: I think this comment was missed.

It's also not a coherent sentence. It's supposed to say:

Should `ifmodule/` also be searching `noarg_path.lower()`?

certbot-apache/certbot_apache/parser.py
```diff
@@ -292,22 +292,45 @@ def add_dir_to_ifmodssl(self, aug_conf_path, directive, args):
...
-                                      (aug_conf_path, mod)))
+        if not if_mods:
+            return self.create_ifmod(aug_conf_path, mod, beginning)
+
         # Strip off "arg" at end of first ifmod path
-        return if_mods[0][:len(if_mods[0]) - 3]
+        return if_mods[0].rpartition("arg")[0]
+
+    def create_ifmod(self, aug_conf_path, mod, beginning=False):
+        """Creates a new <IfMod mod> and returs it path.
```
bmw:
```suggestion
        """Creates a new <IfMod mod> and returns its path.
```

certbot-apache/certbot_apache/override_centos.py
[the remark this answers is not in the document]
bmw: I think [this comment](https://github.com/certbot/certbot/pull/6784/files#r267590622) was missed.

My comment is also not a coherent sentence. It's supposed to say:

Should `ifmodule/` also be searching `noarg_path.lower()`?

certbot-apache/certbot_apache/parser.py
```diff
@@ -303,21 +303,34 @@ def get_ifmod(self, aug_conf_path, mod, beginning=False, force_create=False):
...
+            c_path_arg = "{}/IfModule[1]/arg".format(aug_conf_path)
+            # Insert IfModule before the first directive
+            self.aug.insert("{}/directive[1]".format(aug_conf_path),
+                            "IfModule", True)
+            retpath = "{}/IfModule[1]/".format(aug_conf_path)
+        else:
+            c_path = "{}/IfModule[last() + 1]".format(aug_conf_path)
+            c_path_arg = "{}/IfModule[last()]/arg".format(aug_conf_path)
+            self.aug.set(c_path, "")
+            retpath = "{}/IfModule[last()]/".format(aug_conf_path)
```
bmw: Is there a way to normalize an Augeas path?

In the current PR we compare two Augeas paths to see if they are the same, but will `match` ever return a path using `last()` like this? It's probably not an issue for our current PR, but I'm curious if this change to the return value to `get_ifmod` will cause any problems anywhere else in the code base where we may try to compare paths like this.

certbot-apache/certbot_apache/parser.py
```diff
@@ -303,21 +303,34 @@ def get_ifmod(self, aug_conf_path, mod, beginning=False, force_create=False):
...
+        return if_mods[0].rpartition("arg")[0]
+
+    def create_ifmod(self, aug_conf_path, mod, beginning=False):
+        """Creates a new <IfMod mod> and returs it path.
+
+        :param str aug_conf_path: Augeas configuration path
+        :param str mod: module ie. mod_ssl.c
+        :param bool beginning: If the IfModule should be created to the beginning
+            of augeas path DOM tree.
+
```
bmw: nit: can we document the return value?

certbot-apache/certbot_apache/parser.py
joohoi: Match never returns a `last()` or any other dynamic XPATH value. Currently the get_ifmod is only used in `parser.add_dir_to_ifmodssl()` (in addition to the overrride_centos.py) and that doesn't compare paths, and while we compare paths in the CentOS override, the paths are always added as `beginning=True`, resulting to index 1. 

Previously this method added the new directive, then did a search, and returned the first value of the matches. I see this previous functionality potentially causing more issues than the current approach. We could figure our way around this by returning the last object in `if_mods` list of matches if `beginning=False`, and the first one otherwise. That said, I'm not 100% convenient that Augeas always returns the paths in the same order (re: spurious Apache test failures in VirtualHost matching).

I think that we are fine with this current approach, after I document the return value (as requested in the next review comment)

joohoi:
This is ready for review again. My apologies, this must have been a pain to review :/

certbot-apache/certbot_apache/parser.py
bmw: That all makes sense. Thanks for diving into it a bit more to prove there's no problems caused by the changes here.

I don't think we need to do anything else now, but this is an ugly problem. It's not completely unrelated to the problems in #6256 which also is due to problems trying to use xpaths.

It's not on our roadmap or part of our OTF grant, but if we end up sticking with Augeas, I personally think we should take the time to do some serious refactoring of the code here as soon as we can. It feels like we're repeatedly hitting issues like this which I think we can solve by rethinking how we use Augeas and limiting its scope to a subset of the Apache plugin rather than letting it propagate everywhere. That will be a lot of work, but I feel like we're spending a lot of time fighting these issues on any significant Apache PR.

Do you agree? Feel free to say no, I'm just genuinely curious if you are feeling and thinking the same way about this.

certbot-apache/certbot_apache/parser.py
bmw: Thanks for adding this to `get_ifmod` but can you also add it to `create_ifmod`? I actually put this comment on `create_ifmod` because I didn't feel great asking you to add missing documentation to functions you didn't write in this PR, but with the change we made to `get_ifmod`'s return value, I'm glad you did so!

bmw:
> My apologies, this must have been a pain to review :/

No worries!

Reviewing this did take more rounds than most PRs, but I think that was mostly caused by [technical debt/Augeas issues](https://github.com/certbot/certbot/pull/6784#discussion_r270114302) and this being an awkward problem to solve due to strange SSL setup on CentOS 6.

bmw:
Given my only outstanding comment is https://github.com/certbot/certbot/pull/6784#discussion_r270115557, I think we should try and resolve this for the release on Wednesday.

There's also https://github.com/certbot/certbot/pull/6784#discussion_r270114302, but that's largely unrelated to this PR.

certbot-apache/certbot_apache/parser.py
joohoi: Yeah, I was supposed to do that but somehow it slipped through. It's now added however.

certbot-apache/certbot_apache/parser.py
joohoi: I absolutely and wholeheartedly agree, we had a discussion about this offline, and it looks like a good way forward in all aspects - also those relating to other development goals we have in our roadmap.

joohoi:
The changes are now in, and this should be ready to merge.

PR #6786, opened 2019-02-20
Pin Python dependencies for each DNS plugin - continued

### LINKED ISSUE #6581: Pin Python dependencies installed in DNS plugins

joohoi:
This continues from the work of @sydneyli in PR #6671
I didn't do much here. Basically added support for reading data from `sys.stdin` to both `tools/merge_requirements.py` and `tools/strip_hashes.py` as well as support for reading files from paths passed as cli parameters to `strip_hashes.py`.

Reading the filepaths was not strictly required, but I thought would be a good thing to do in order to keep the tooling usage options consistent.

Fixes #6581

### DIFF
--- a/tools/merge_requirements.py
+++ b/tools/merge_requirements.py
@@ -10,27 +10,36 @@ from __future__ import print_function
 import sys
 
 
-def read_file(file_path):
-    """Reads in a Python requirements file.
+def process_entries(entries):
+    """
     Ignore empty lines, comments and editable requirements
 
-    :param str file_path: path to requirements file
+    :param list entries: List of entries
 
     :returns: mapping from a project to its pinned version
     :rtype: dict
-
     """
     data = {}
-    with open(file_path) as file_h:
-        for line in file_h:
-            line = line.strip()
-            if line and not line.startswith('#') and not line.startswith('-e'):
-                project, version = line.split('==')
-                if not version:
-                    raise ValueError("Unexpected syntax '{0}'".format(line))
-                data[project] = version
+    for e in entries:
+        e = e.strip()
+        if e and not e.startswith('#') and not e.startswith('-e'):
+            project, version = e.split('==')
+            if not version:
+                raise ValueError("Unexpected syntax '{0}'".format(e))
+            data[project] = version
     return data
 
+def read_file(file_path):
+    """Reads in a Python requirements file.
+
+    :param str file_path: path to requirements file
+
+    :returns: list of entries in the file
+    :rtype: list
+
+    """
+    with open(file_path) as file_h:
+        return file_h.readlines()
 
 def output_requirements(requirements):
     """Prepare print requirements to stdout.
@@ -53,7 +62,15 @@ def main(*paths):
     """
     data = {}
     for path in paths:
-        data.update(read_file(path))
+        data.update(process_entries(read_file(path)))
+
+    # Need to check if interactive to avoid blocking if nothing is piped
+    if not sys.stdin.isatty():
+        stdin_data = []
+        for line in sys.stdin:
+            stdin_data.append(line)
+        data.update(process_entries(stdin_data))
+
     return output_requirements(data)
 
 
--- a/tools/merge_requirements.py
+++ b/tools/merge_requirements.py
@@ -55,21 +55,24 @@ def main(*paths):
     """Merges multiple requirements files together and prints the result.
 
     Requirement files specified later in the list take precedence over earlier
-    files.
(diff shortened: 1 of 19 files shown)


tools/strip_hashes.py
```diff
@@ -1,16 +1,45 @@
...
+            deps += process_entries(file_h.readlines())
+
+    # Need to check if interactive to avoid blocking if nothing is piped
+    if not sys.stdin.isatty():
+        stdin_data = []
+        for line in sys.stdin:
+            stdin_data.append(line)
+        deps += process_entries(stdin_data)
+
+    return os.linesep.join(deps)
```
bmw: nit: Python docs say [not to use os.linesep when writing to text files](https://docs.python.org/3/library/os.html#os.linesep) and [print](https://docs.python.org/3/library/functions.html#print) treats stdout as a text file.

tools/merge_requirements.py
```diff
@@ -53,7 +62,15 @@ def main(*paths):
     """
     data = {}
     for path in paths:
-        data.update(read_file(path))
+        data.update(process_entries(read_file(path)))
+
+    # Need to check if interactive to avoid blocking if nothing is piped
+    if not sys.stdin.isatty():
```
bmw: I'm a little worried about the mix of file order mattering and taking files from both argv and stdin. What comes first? argv or stdin? We could arbitrarily pick one, but there doesn't seem to be an intuitively correct order to me so I'd personally rather not pick at all.

I think we should either:

1. Stick with just parsing files given on the command line. We actually don't even need this functionality because the files already exist in the Dockerfile. We're just running `cat <file> <file> | merge_requirements` when we can simply run `merge_requirements <file> <file>`. This keeps the code simpler, avoids this problem, and is arguably the better engineering decision, but it makes the UNIX fanboy in me somewhat unhappy...
2. Alternatively, we could only read stdin if no files were provided on the command line. If we take this approach, I think we should make sure to update the docstring of this function/file to explain this behavior.

You're welcome to make the call here between these two approaches (or another one you have in mind?) if you like.

tools/strip_hashes.py
```diff
@@ -1,16 +1,45 @@
...
+def main(*paths):
+    """Reads dependency definitions from a (list of) file(s) or stdin and
+    removes hashes from returned entries"""
+
+    deps = []
+    for path in paths:
+        with open(path) as file_h:
+            deps += process_entries(file_h.readlines())
+
+    # Need to check if interactive to avoid blocking if nothing is piped
```
bmw: I'm less worried about this because the order is less likely to matter here, but I think we should probably make a similar decision to the one we do above here about reading from both files and stdin.

tools/pip_install.py
```diff
@@ -47,10 +48,8 @@ def certbot_normal_processing(tools_path, test_constraints):
     with open(certbot_requirements, 'r') as fd:
         data = fd.readlines()
     with open(test_constraints, 'w') as fd:
-        for line in data:
-            search = re.search(r'^(\S*==\S*).*$', line)
-            if search:
-                fd.write('{0}{1}'.format(search.group(1), os.linesep))
+        data = os.linesep.join(strip_hashes.process_entries(data))
```
bmw: nit: Python docs say [not to use os.linesep when writing to text files](https://docs.python.org/3/library/os.html#os.linesep).

tools/merge_requirements.py
joohoi: We should go with option 2.

PR #6794, opened 2019-02-25
Cover case of OpenSUSE Leap 15+ in certbot-auto

### LINKED ISSUE #6228: certbot-auto initial bootstrap fails on openSUSE; attempts to install invalid 'python-virtualenv' *src* package
With release certbot-auto
```
	wget https://dl.eff.org/certbot-auto
	chmod a+x ./certbot-auto
```

1st exec
```
	./certbot-auto --version --verbose
```

on openSUSE* bootstraps
```
	...
	Bootstrapping dependencies for openSUSE-based OSes... (you can skip this with --no-bootstrap)
	....
```

and tries to install package
```
	python-virtualenv
```

as spec'd in
```
	cat certbot-auto
		...
		BootstrapSuseCommon() {
		  # SLE12 don't have python-virtualenv

		  if [ "$ASSUME_YES" = 1 ]; then
		    zypper_flags="-nq"
		    install_flags="-l"
		  fi

		  if [ "$QUIET" = 1 ]; then
		    QUIET_FLAG='-qq'
		  fi

		  zypper $QUIET_FLAG $zypper_flags in $install_flags \
		    python \
		    python-devel \
>>>		    python-virtualenv \
		    gcc \
		    augeas-lenses \
		    libopenssl-devel \
		    libffi-devel \
		    ca-certificates
		}
		...
```

on latest openSUSE Leap15 & Tumbleweed that's a *source* package, causing the install to fail,

```
zypper in python-virtualenv
	Loading repository data...
	Reading installed packages...
	'python-virtualenv' not found in package names. Trying capabilities.
	...
```

as seen here, on the OBS,

https://build.opensuse.org/package/binaries/openSUSE:Factory/python-virtualenv/standard

the actual binary packages are
```
python2-virtualenv
python3-virtualenv
```

and are, obviously, split-packaged for python v2 & v3, respectively.

adferrand:
Fixes #6228.

Since OpenSUSE Leap 15, `python-virtualenv` became a source package, breaking certbot-auto bootstrap on this version. Then `python2-virtualenv` must be used to create Python 2.x virtual environments.

This PR makes certbot-auto compatible to prior and after Leap 15, by testing the existence of `python-virtualenv` on current OpenSUSE system, and then use appropriate packages.


letsencrypt-auto-source/pieces/bootstrappers/suse_common.sh
```diff
@@ -1,6 +1,6 @@
 # If new packages are installed by BootstrapSuseCommon below, this version
 # number must be increased.
-BOOTSTRAP_SUSE_COMMON_VERSION=1
+BOOTSTRAP_SUSE_COMMON_VERSION=2
```
bmw: We haven't documented this anywhere, but bumping this version number is what causes `certbot-auto` to refuse to upgrade itself until it is running interactively and can prompt the user for approval to install new packages.

In this case, I think it's overkill and we should leave this at 1 because if you've successfully created a venv in the past, you should be able to do so again without installing new Python packages.

adferrand: Ok, I did not got it the first time. So the increment is only relevant when in the exact same OS system, some packages need to be changed. I revert the increment.

adferrand:
Almost the same error than another user saw on Ubuntu 16.08, 3 days ago, see #6797. There may be a connection!

letsencrypt-auto-source/pieces/bootstrappers/suse_common.sh
bmw: That's correct. The only time we've ever made use of this functionality is to switch CentOS 6 from Python 2.6 to Python 3.4. In that case, rather than bumping the version for all Red Hat OSes, we created a [new bootstrap script and version for CentOS 6](https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh) to limit the need to "rebootstrap" to those OSes.

bmw:
May be! It occurs when running the `virtualenv` executable after installing the python2 variants of the package names in this PR.

`pkg_resources` is part of `setuptools` and installing `python2-setuptools` fixed the problem on at least one OS I quickly tested on.

adferrand:
This new implementation works correctly for Leap pre-15 and 15+.

We should fill a bug to upstream about `python2-virtualenv` not declaring `python2-setuptools` as a dependency. In the mean time, this version of certbot-auto cover that.

letsencrypt-auto-source/pieces/bootstrappers/suse_common.sh
```diff
@@ -14,10 +14,18 @@ BootstrapSuseCommon() {
     QUIET_FLAG='-qq'
   fi
 
+  if zypper search -x python-virtualenv >/dev/null 2>&1; then
+    OPENSUSE_VIRTUALENV_PACKAGES="python-virtualenv"
+  else
+    # Since Leap 15.0 (and associated Tumbleweed version), python-virtualenv
+    # is a source package, and python2-virtualenv must be used instead.
+    OPENSUSE_VIRTUALENV_PACKAGES="python2-virtualenv python2-setuptools"
```
bmw: nit: Can you add a comment about why we're including `python2-setuptools` so we remember in the future?

adferrand: Done.

adferrand:
Yes, I created #6806 to track it.

letsencrypt-auto-source/pieces/bootstrappers/suse_common.sh
```diff
@@ -14,10 +14,20 @@ BootstrapSuseCommon() {
     QUIET_FLAG='-qq'
   fi
 
+  if zypper search -x python-virtualenv >/dev/null 2>&1; then
+    OPENSUSE_VIRTUALENV_PACKAGES="python-virtualenv"
+  else
+    # Since Leap 15.0 (and associated Tumbleweed version), python-virtualenv
+    # is a source package, and python2-virtualenv must be used instead.
+    # Also currently python2-setuptools is not a dependency of python2-virtuqlenv,
```
bmw:
```suggestion
    # Also currently python2-setuptools is not a dependency of python2-virtualenv,
```

bmw:
Looks like we need to run `build.py`.

PR #6795, opened 2019-02-26
Warn sysadmins about privilege escalation risk

### LINKED ISSUE #6242: Sysadmins should not let unprivileged users run certbot as root
Because Certbot can be given scripts and commands to run as hooks, allowing a user to run Certbot allows them to run arbitrary code as root on the machine. We should make this very clear in our documentation.

schoen:
Fixes #6242.

Maybe this isn't the best place in the documentation for this warning, since it's extremely prominent but relates to a risk that might not be very important to most users.


schoen:
Just checking whether this is a false alarm (otherwise I guess we have to exclude documentation from coverage tests).

PR #6799, opened 2019-02-28
Fix spurious test_relevant_values* failures

bmw:
Tests were failing on #6766. I believe this was due to the global state used in `cli.py`.

We could try to fix up this global state for each test, but `storage_test.py` seems like the wrong place to have tests that are strongly coupled to `cli.py`. I think such tests should be in `cli_test.py`. In fact, we already have tests for the problematic function at https://github.com/certbot/certbot/blob/11e0fa52d0f2e7d8780465f5b2202e0cb717e065/certbot/tests/cli_test.py#L470

Instead, I just mocked out the function in `cli.py` using the problematic global state allowing us to focus on testing pieces specific to `storage.py`.

I tested the failing environments in #6766 with this PR merged on top of it at https://travis-ci.com/certbot/certbot/builds/102590400 and tests passed.

### DIFF
--- a/certbot/tests/cli_test.py
+++ b/certbot/tests/cli_test.py
@@ -340,6 +340,8 @@ class ParseTest(unittest.TestCase):  # pylint: disable=too-many-public-methods
         config_dir_option = 'config_dir'
         self.assertFalse(cli.option_was_set(
             config_dir_option, cli.flag_default(config_dir_option)))
+        self.assertFalse(cli.option_was_set(
+            'authenticator', cli.flag_default('authenticator')))
 
     def test_encode_revocation_reason(self):
         for reason, code in constants.REVOCATION_REASONS.items():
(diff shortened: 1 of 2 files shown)


adferrand:
Reopening the request to trigger a clean of Travis state.

adferrand:
Pushing empty commit to trigger a clean of Travis state. Our CI is not in a super good shape those days.

bmw:
While I completely agree that refactoring the code here is valuable, I think it's not an entirely trivial refactor and that it's hard for me to justify personally spending much time on it right now.

Instead I:

1. Created #6802 to track the issue of doing this work.
2. Modified the tests again to do less mocking. I did this by reloading the module between tests to clean global state and creating a dummy plugin object to use in one of the existing mocks which I think results in a stronger test that we have the correct behavior here.

adferrand:
Indeed with this implementation, you make the unit test closer to the real world, instead of mocking everything. However, it is at the price of a higher coupling (with disco plugin logic), a module reloading, and more complexity in the test method.

Is the price worth it? More specifically, does it increase the number of logic units tested, compared to mocking everything, and so makes the test more likely to fail if something goes wrong?

bmw:
Fair enough. I'll offer you one more implementation 😄

bmw:
In the new implementation, I went with always mocking out the function we were calling in `cli.py` and moved the only missing test of the code there to `cli_test.py` where it probably belongs.

PR #6800, opened 2019-02-28
Disable rerun feature of Travis

adferrand:
The rerun capability in a test campaign can be a nice feature. It can also a bad design.

It is right that with high level tests, like performance or end-to-end tests, a given test runtime can depend on an external component, outside the scope of the developers, that would spuriously fail. In theses situations, having the capacity to rerun several time a test can be a great benefit. Indeed, as these tests are inherently flaky, the rerun greatly reduces their failure because of external reasons, reducing also the `Pierre et le  Loup` effect (from a well-known french book for children): because tests constantly fail for no reason, you stop to listen to them and to not see when they fail for a real reason.

However this not apply to unit tests, or operations about code quality. For theses executions, the flakiness should not exist: a unit test is supposed to have no external dependency. A rerun approach would just hide a situation that is not desirable for this kind of tests

Also effectively I see that our tests are usually not flaky: so the only effect of the rerun is to give the failure state about a test three times slower. If a test becomes flaky, this should be fixed, and so be visible immediately in our CI.

For these reasons, I remove the `travis_retry` in the  `script` section of `.travis-ci.yml`, to call directly `tox` and let the pipeline fails on the first error.

Instead, I set a travis_retry in the install step, as this could theorically solve temporary network failures while executing pip. Arguably it could be removed also, as I do not think that 3 executions within a minute will be sufficient for a network failure, that is most likely be a really critical error with complex solving from the pip ecosystem maintainers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -200,7 +200,7 @@ addons:
 install: "travis_retry $(command -v pip || command -v pip3) install codecov tox"
 script:
     - tox
-    - '[ -z "${BOULDER_INTEGRATION+x}" ] || (tests/boulder-fetch.sh && tests/tox-boulder-integration.sh)'
+    - '[ -z "${BOULDER_INTEGRATION+x}" ] || (travis_retry tests/boulder-fetch.sh && tests/tox-boulder-integration.sh)'
 
 after_success: '[ "$TOXENV" == "py27-cover" ] && codecov'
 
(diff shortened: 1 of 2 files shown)


adferrand:
I also never saw pip installation failing in the CI because of external causes. I will remove that.

PR #6801, opened 2019-02-28
Fix the Nginx configuration during integration tests

adferrand:
If you execute `tests/lock_test.py` or `tox -e integration` on a fairly recent machine, you will get the following error during tests executing against a live Nginx instance:
```
no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking, client: x.x.x.x, server: y:y:y:y:z
```

Indeed, having no defined ssl certificate for a ssl port would inevitably lead to an error during the handshake SSL process between a client and this mis-configured nginx instance.

However it was not a problem one year before, because the handshake was not occurring in practice: the test just need to have a nginx started, and then immediately proceed to modify the configuration with a correct SSL setup. And nginx was able to start with a mis-configuration on SSL. 

But then this fix has been done: https://trac.nginx.org/nginx/ticket/178

Basically with this, validation of the configuration is done during nginx startup, that will refuse to start with invalid configuration on SSL. Consequently, all related tests are failing with a sufficiently up-to-date nginx. For now, it is not seen on Travis because Ubuntu Trusty is used, with an old Nginx.

The PR fixes that, by generating on the fly self-signed certificates in the two impacted tests, and pushing the right parameters in the Nginx configuration.


adferrand:
I added these ignores because otherwise, mypy complains about the attribute not existing in the target object.

I think this is because these attributes are added dynamically at runtime. This is something mypy cannot see. There are several of these ignores assertion in crypto_util, that relies heavily on cryptography.

So I do not really know what to do more here.

bmw:
The attributes being dynamic might indeed be the cause but I believe the current way to work around the problem is getting type annotations added to https://github.com/python/typeshed for cryptography. See https://github.com/certbot/certbot/blob/9c405a3cd12588ba27cdd8f79c8a101831bd6ad3/certbot/crypto_util.py#L21-L22

Can we add comments explaining this by the two `type: ignore` comments so we can check back in on this in the future?

adferrand:
I tested with the latest version of mypy. Still the same errors, the dynamic typing definitions did not make it yet into mypy.

adferrand:
I added the comments.

PR #6804, opened 2019-03-01
Disable build for commits pushed on master

### LINKED ISSUE #6746: Turn off tests on commits to master
Erica pointed that since we're running tests on master nightly running tests on each commit to the branch may not be necessary. Sure, there may be edge cases where the combination of multiple pull requests cause a problem in `master` and break something, but this is extremely rare (in almost 4 years of working on Certbot full time I'm not sure if it's ever happened?), would be caught by our nightly tests in at most 24 hours, and dramatically saves us on Travis resources.

I personally think we should do this, but I'm not sure the best way to accomplish it. I think if we remove `master` from the list of branches we test, we'll stop testing PRs against `master` and the Travis cron job on `master` might stop working as well. The best idea I have is to add conditionals so that none of our tests run on push events to `master` which is probably fine, but if other people have ideas, please share!

adferrand:
Fixes #6746.

Every commit on master is always the result of a merged PR, that has been tested by Travis. So retesting the merge commit on master is superfluous. This PR uses build conditions to avoid to launch a build for a commit push on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ branches:
     - /^\d+\.\d+\.x$/
     - /^test-.*$/
 
+# Since master can receive only commits from PR that have already been tested, we avoid with the
+# the following condition to launch again a pipeline when the merge commit is pushed to master.
+# However master still needs to be set in branches section to allow PR for master to be built.
+if: NOT (type = push AND branch = master)
+
 matrix:
   include:
     # These environments are always executed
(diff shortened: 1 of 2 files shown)


adferrand:
I also added the equivalent logic for AppVeyor. Builds cannot received conditions, so it needs to be done on init using `Exit-AppVeyorBuild`. This command does not fail the build, it finishes it prematurely with success.

See an example here: https://ci.appveyor.com/project/adferrand/certbot-tv55t/builds/22757870/job/mk9k1vwu4y28ja3o

PR #6811, opened 2019-03-03
Do not run the full CI pipeline on master

adferrand:
This PR proposes a better implementation of #6804. This idea is to save the CI working capacity by avoiding to run the CI pipeline on commits to master. Indeed, any commit to this branch comes from a pull request that has already been processed by the CI pipeline. 

First implementation was disabling totally Travis and AppVeyor jobs for pushes to master. However, this approach was breaking the quality gate executed on a PR through Coveralls, because Coveralls needs a report generated on the latest master commit on which a given PR may be merged. And obviously, this report is generated by the pipeline executed on pushes to master... 

On this PR, I leverage available logic on Travis and AppVeyor in order to run the complete pipeline on PRs and pushes to non-master branches, and only the job with coverage on pushes to master:
* on Travis, this is done with conditional stages each of them having the relevant jobs from the matrix execution be declared
* on AppVeyor, the command `Exit-AppVeyorBuild` that exits immediately and successfully a job, is launched on pushes to master branch if the job is not about coverage.

I took also this occasion to classify all jobs that are run only on cron execution to a specific stage. This allows to group the execution condition on the stage level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,141 +18,163 @@ branches:
     - /^\d+\.\d+\.x$/
     - /^test-.*$/
 
-stages:
-  # Normal stage always executed (including PR) except pushes on master
-  - name: test
-    if: NOT (type = push AND branch = master)
-  # Restricted stage executed only for pushes on master
-  - name: test-master
-    if: type = push AND branch = master
-  # Extended stage executed on cron jobs an push on non-master branches
-  - name: test-extended
-    if: type = cron OR (type = push AND branch != master)
+# Normal job always executed (including PR) except pushes on master
+not-on-master: &not-on-master
+  if: NOT (type = push AND branch = master)
+
+# Extended job executed on cron jobs and pushes on non-master branches
+extended-pipeline: &extended-pipeline
+  if: type = cron OR (type = push AND branch != master)
 
 matrix:
   include:
-    # Jobs attached to the test stage
-    - stage: test
-      python: "2.7"
+    # Main pipeline
+    - python: "2.7"
       env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=all TOXENV=py27_install
       sudo: required
       services: docker
+      <<: *not-on-master
     - python: "2.7"
       env: BOULDER_INTEGRATION=v2 INTEGRATION_TEST=all TOXENV=py27_install
       sudo: required
       services: docker
+      <<: *not-on-master
+
+    # This job is always executed, including on master
     - python: "2.7"
       env: TOXENV=py27-cover FYI="py27 tests + code coverage"
+
     - sudo: required
       env: TOXENV=nginx_compat
       services: docker
       before_install:
       addons:
+      <<: *not-on-master
     - python: "2.7"
       env: TOXENV=lint
+      <<: *not-on-master
     - python: "3.4"
       env: TOXENV=mypy
+      <<: *not-on-master
     - python: "3.5"
       env: TOXENV=mypy
+      <<: *not-on-master
     - python: "2.7"
       env: TOXENV='py27-{acme,apache,certbot,dns,nginx,postfix}-oldest'
       sudo: required
       services: docker
+      <<: *not-on-master
     - python: "3.4"
       env: TOXENV=py34
       sudo: required
       services: docker
+      <<: *not-on-master
     - python: "3.7"
       dist: xenial
       env: TOXENV=py37
       sudo: required
       services: docker
+      <<: *not-on-master
     - sudo: required
       env: TOXENV=apache_compat
       services: docker
       before_install:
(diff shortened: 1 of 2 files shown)


bmw:
While using stages allows us to avoid duplicating the `if` condition in each job, the downsides are:

* We have to duplicate the py27-cover job configuration.
* Our full test suite which runs on point release branches, when requested by creating a `test-` branch, or nightly no longer fully executes in parallel. Instead, the first stage runs to completion and then the second stage runs.

I think this 2nd point is a somewhat significant downside. When tests are running nightly it doesn't really matter, but when doing things like trying to test a change for a bug fix release, it could be kind of significant.

I'm currently running the full test suite with these changes at https://travis-ci.com/certbot/certbot/builds/103086838 to get an idea how much longer it takes, but personally, I think I'm leaning towards just duplicating the `if` conditions and unconditionally running our coverage tests in our Travis config.

adferrand:
Indeed. It would make sense if all jobs on the first stage would run in an equivalent time, but three of them are 10 min longer than the others. So I expect the overall time added of 10 min, so from 30 min to 40 min.

However, in such a long run time, 30 min or 40 min, a fail fast strategy start to make sense. Running a full suite of tests, when the first ones are failing anyway, giving no chance for the latter ones to give significant result, is a waste of ressources. This waste is the subject of this issue in fact.

I think that since these tests are mainly executed in cron, and minority of them are executed on demand, plus in this case the ability to give the failure after only 15 min in this case, make the fail fast approach quite suitable in my opinion.

bmw:
Your estimate was pretty close. It took 15 minutes longer. That doesn't sound so bad but it is a 50% increase in the time it takes. I'm not worried about this for nightly tests, but I think it's a bit of an unnecessary burden when testing for a point release or manually running tests.

I'm also not sure I agree we always want to not run all tests if some are failing. I argued before and still think that an individual test should fail fast. It simplifies the output and code when you don't have to worry about potential problems from earlier in execution.

In this case though, neither of those things are concerns. Different tests can fail for different reasons and I think fully reporting on all problems can be valuable. Since we're talking about only saving resources when our full test suite is run and failures occur in the first stage, I don't think we're talking about that many jobs here.

adferrand:
OK. If I still think that a fail fast approach makes sense in a pipeline, I agree that the test-everything approach is not a real pipeline. It is more a tests farm constituting scattered tests over several OS to have a global overview.

Also I would definitively like rethinking the CI pipeline for PR in the scope of stages. But here I am using stages more for grouping the conditions to execute each tests, not to have a grouping depending of the logic of each job (unit tests, quality gate, integration tests, end-to-end tests).

I move back to conditions applied on each job, under the default `test` stage.

bmw:
Woah! It's pretty nice that you can put the conditional logic in a variable to simplify making changes to it in the future.

Do you have a link to the documentation about how that works? I haven't seen it before.

adferrand:
It is not specific to Travis in fact. It is part of the YAML format. See for instance https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

.travis.yml
```diff
@@ -18,149 +18,171 @@ branches:
...
       services: docker
+      <<: *not-on-master
     - python: "2.7"
       env: TOXENV=nginxroundtrip
+      <<: *not-on-master
 
-    # These environments are executed on cron events and commits to tested
-    # branches other than master. Which branches are tested is controlled by
-    # the "branches" section earlier in this file.
+    # Extended pipeline on cron jobs and push to test-* branches
```
bmw: nit: I'd change this to something like "and pushes to tested branches other than master". Another important category here are point release branches.

adferrand:
Yes it is not a real pipeline. A pipeline is a series of stages, containing each several jobs, that together serve a purpose for a CI or CD process. It would more correspond to what I described before: unit tests -> integration tests -> coverage and quality -> ...

I updated to test suite in the Travis config.

.travis.yml
```diff
@@ -18,149 +18,171 @@ branches:
     - /^\d+\.\d+\.x$/
     - /^test-.*$/
 
+# Job for main test suite, always executed (including PR) except for pushes on master.
```
bmw:
```suggestion
# Jobs for the main test suite are always executed (including on PRs) except for pushes on master.
```

.travis.yml
```diff
@@ -18,149 +18,171 @@ branches:
...
+not-on-master: &not-on-master
+  if: NOT (type = push AND branch = master)
+
+# Job for extended test suite, executed for cron jobs and pushes on non-master branches.
```
bmw:
```suggestion
# Jobs for the extended test suite are executed for cron jobs and pushes on non-master branches.
```

PR #6812, opened 2019-03-03
Correct certbot-auto for Fedora 29+

### LINKED ISSUE #6698: certbot-auto doesn't work on Fedora 29

adferrand:
Fixes #6698 

Fedora maintainers engaged a deprecation path for Python 2.x with Fedora 29. As a first step, `python2-virtualenv` does not install the `virtualenv` binary anymore, in favor of `python3-virtualenv`, and so the installation of Python 3 virtual environments by default.

However, certbot-auto installs `python2-virtualenv` for all recent RPM distributions, and relies of the execution of `virtualenv`, and this is failing the process.

Since the plan in the future is to remove Python 2.x from Fedora, this PR follows this logic to fix certbot-auto: started to Fedora 29, certbot-auto will install and execute certbot on Python 3. This implies to detect that we are on Fedora 29+, install `python3-virtualenv` that will install also Python 3 dependencies and `virtualenv` binary, then instruct the process to use Python 3. This is in fact similar to EOL distributions shipping with Python 2.6, and for which Python 3.4 from EPEL is installed and used.

Older versions of Fedora continue to use Python 2.x, and their process is untouched. Four scenarios are covered here:
* fresh Fedora 28: old process is used, nothing changes
* fresh Fedora 29: new process is used, Python 3 is installed, certbot runs on it
* update Fedora 29 from 28, already installed certbot-auto without rebootstrapping required: existing venv continue to be used, certbot runs on it
* update Fedora 29 from 28, already installed certbot-auto with rebootstrapping required: new process is used, installing `python3-virtualenv`, `python3-devel` and `python3-rpm-macros`, Python 3 is installed, certbot runs on it


bmw:
I just want to make sure we're on the same page for expected behavior here.

> update Fedora 29 from 28, already installed certbot-auto without rebootstrapping required

How would someone update from 28 to 29 and not need to rebootstrap?

adferrand:
Because no other certbot-auto version has been released during the user upgrade? And so the installed venv is still valid. It is what I observe in this situation.

bmw:
Ah OK.

I think that once a version of certbot-auto with these changes is released though, those users will have to rebootstrap because the bootstrap script that was previously run was for Python 2 dependencies and certbot-auto will want to run the Python 3 bootstrap code to make sure it has the Python 3 dependencies.

This isn't a problem, but is this the behavior you expect?

adferrand:
Yes for the rebootstrapping on Fedora 29 for new certbot version, yes, it is what I expect.

Concerning `/etc/os-release`, yes it is a standardization, and seems to become the recommended way to get info about current distribution.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -323,7 +323,10 @@ elif [ -f /etc/redhat-release ]; then
   prev_le_python="$LE_PYTHON"
   unset LE_PYTHON
   DeterminePythonVersion "NOCRASH"
-  if [ "$PYVER" -eq 26 ]; then
+  # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+  RPM_DIST_NAME=`(source /etc/os-release && echo $ID) || echo "unknown"`
```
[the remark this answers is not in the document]
adferrand: Yes you are right.

letsencrypt-auto-source/pieces/bootstrappers/rpm_python3.sh
```diff
@@ -5,11 +5,17 @@ BOOTSTRAP_RPM_PYTHON3_VERSION=1
...
   # Tested with:
   #   - CentOS 6
+  #   - Fedora 29
 
   InitializeRPMCommonBase
 
+  # Fedora 29 must use python3-virtualenv
+  if $TOOL list python3-virtualenv >/dev/null 2>&1; then
+    python_pkgs="python3
+      python3-virtualenv
```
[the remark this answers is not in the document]
adferrand: No impact on current set of dependencies installed, allright!

adferrand:
I added the changelog.

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-*
+* Fedora 29+ is know supported by certbot-auto. Since Python 2.x starts to be
```
bmw:
```suggestion
* Fedora 29+ is now supported by certbot-auto. Since Python 2.x starts to be
```

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
...
+  deprecated with Fedora 29, certbot-auto will install and use Python 3.x instead.
```
bmw: Is it true that Python 2 is deprecated in Fedora 29? They're definitely going that direction, but looking at https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3, it seems to me like that's coming later.

If that's the case or you're not sure, I'd make this more generic and say something like:

> Since Python 2.x is on a deprecation path in Fedora, certbot-auto will install and use Python 3.x on Fedora 29+.

adferrand: Yes you are right, I am a little too much assertive here.

PR #6815, opened 2019-03-05
Route53 Plugin: Add throwaway domain options.

foxdan:
Allows use of throwaway domains using CNAME records as described in:
https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation


bmw:
I'm closing this PR in favor of #6644 which does the same thing but for all DNS plugins rather than just route53.

PR #6817, opened 2019-03-05
Replace deprecated Gandi plugin link

obynio:
The [previous certbot plugin for Gandi](https://github.com/Gandi/letsencrypt-gandi) is deprecated and not maintained anymore. I updated the documentation [with a new plugin](https://github.com/obynio/certbot-plugin-gandi) that support the new Gandi LiveDNS v5 API.


schoen:
Hi @obynio, thanks for this. Do you think you could mention on the GitHub page for your plugin that it's not usually relevant to people who are using Gandi's web hosting services, since that seems to be the difference in audience between the old Gandi-maintained plugin and your version?

obynio:
@schoen indeed good suggestion, I will add this right now :+1:

Issue #6506, opened 2018-11-14
Tracking Amazon Linux 2 support


sydneyli:
Tracking issue for Amazon Linux 2 support.

If you stumbled onto this issue, try out our [CentOS/RHEL 7](https://certbot.eff.org/lets-encrypt/centosrhel7-other) instructions or one of the guides written by Amazon about this: [\[1\]](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/SSL-on-an-instance.html#letsencrypt) [\[2\]](https://aws.amazon.com/blogs/compute/extending-amazon-linux-2-with-epel-and-lets-encrypt/)

If you're having trouble getting Certbot to work on Amazon Linux 2, try asking for help [at the Let's Encrypt Community Forums](https://community.letsencrypt.org/c/help).

RichardBradley:
This issue shouldn't have been closed. `certbot` does not yet install properly on Amazon Linux 2.

I got it working recently by installing from source, but it wasn't straightforward. The currently distributed binary package of certbot does not work. See my comment at #6505.

I really hate "stalebot". Just because no-one has commented for a little while doesn't mean that it's at all appropriate to close an active issue.

Issue #6505, opened 2018-11-14
Amazon Linux 1 support


sydneyli:
Tracking issue for getting Certbot packaged into Amazon Linux, or just talking about how officially we want to support it if we can't figure that out.

If you're having trouble getting Certbot to work on Amazon Linux, try asking for help [at the Let's Encrypt Community Forums](https://community.letsencrypt.org/c/help).

jpbedoya:
Any chance Amazon Linux support will be coming soon?

bundgaard:
Adding these lines in the certbot-auto script, helps to support Amazon2 on an experimental level.

````elif [ -f /etc/system-release ] && grep -iq "Amazon Linux" /etc/system-release ; then
  Bootstrap() {
    ExperimentalBootstrap "Amazon Linux" BootstrapRpmCommon
  }
  BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"```

ohemorange:
Looks like urllib gives us some trouble on Amazon linux as well https://community.letsencrypt.org/t/doesnt-match-a-supported-version/105246

x29a:
> Looks like urllib gives us some trouble on Amazon linux as well https://community.letsencrypt.org/t/doesnt-match-a-supported-version/105246

A workaround is stated there which worked for me, to explicitly install the last supported version of urllib via

`pip install urllib3==1.23`

This does not solve the root cause, but makes certbot operational again for me. Thanks for looking into this!

RichardBradley:
I recently managed to get `certbot` working on Amazon Linux 2 in `venv`, but I had to patch `parsedatetime`.

I wrote it up here: https://serverfault.com/a/1021490/140833
See also: https://github.com/bear/parsedatetime/issues/251

There is a packaged version of `certbot` in epel on Amazon Linux 2, but it doesn't work as it has several version conflicts with the packaged version of `awscli` which is installed by default on AL2

steinybot:
@RichardBradley your comment deserves a mention on the Amazon Linux 2 issue (https://github.com/certbot/certbot/issues/6506) in case anyone comes across that issue and doesn't see this one.

RichardBradley:
Thanks, that would have been a better issue yes. Unfortunately that issue was closed by the vandal "stalebot", so I didn't find it.

PR #6822, opened 2019-03-06
Remove amazon linux test farm targets.

bmw:
These images are very old and we don't claim to support Amazon Linux. See https://github.com/certbot/certbot/issues/6505, https://github.com/certbot/certbot/issues/6506, and the following output when `certbot-auto` is run on Amazon Linux:
```
FATAL: Amazon Linux support is very experimental at present...
Alternatively, you can install OS dependencies yourself and run this script
again with --no-bootstrap.
```
I think we should fix our support for Amazon Linux. They are a major distro and we have a lot of users on the platform. With that said, I don't think it makes sense for us to keep testing with these very old images in the meantime.

After this lands, the release instructions will need to be updated so Amazon Linux failures from our tests aren't expected.

### DIFF
--- a/tests/letstest/targets.yaml
+++ b/tests/letstest/targets.yaml
@@ -44,16 +44,6 @@ targets:
     #     - [ apt-get, install, -y, curl ]
   #-----------------------------------------------------------------------------
   # Other Redhat Distros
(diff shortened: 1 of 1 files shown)


bmw:
Thank you!

PR #6830, opened 2019-03-08
Move venv symlink check out of leauto_upgrades.

bmw:
Since #6812 has landed, I tried to update the version of Fedora we use in our test farm tests but I hit a few snags.

One of them was that I had to update the initial version of certbot-auto used in [test_leauto_upgrades.sh](https://github.com/certbot/certbot/blob/master/tests/letstest/scripts/test_leauto_upgrades.sh) that we try to upgrade from because old versions of certbot-auto don't work on new versions of Fedora.

Unfortunately, bumping the initial version any higher causes our tests to fail. This is because in the next version of Certbot (0.18.0), we moved the virtual environment from the user's home directory to `/opt` and the script tests that a symlink was properly created from the old virtual environment to the new one which we only do if the old virtual environment exists.

To fix this, I moved the test to test_letsencrypt_auto_certonly_standalone.sh and created a dummy virtual environment. Alternatives I considered were:

1. Creating this dummy virtual environment before running the most recent version of certbot-auto in `test_leauto_upgrades.sh` but then we'd have virtual environments in both `/opt` and the user's home directory which isn't a realistic scenario and should never happen.
2. Creating this dummy virtual environment in `test_leauto_upgrades.sh` before running the initial version of certbot-auto, but I think we should test that we're creating this symlink properly on the most recent version of the script instead of an older version.


ohemorange:
Where's the code in the old version of certbot-auto that would have created the virtual environment in the user's home directory?

bmw:
The code that determined the venv path in the last version of certbot-auto to do this is at https://github.com/certbot/certbot/blob/v0.17.0/certbot-auto#L23-L29.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -79,17 +79,6 @@ if [ $(python -V 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//') -eq 26 ];
         echo "Certbot shouldn't have updated to a new version!"
         exit 1
     fi
-    if [ -d "/opt/eff.org" ]; then
-        echo "New directory shouldn't have been created!"
-        exit 1
-    fi
-    # Create a 2nd venv at the new path to ensure we properly handle this case
-    export VENV_PATH="/opt/eff.org/certbot/venv"
-    if ! sudo -E ./letsencrypt-auto -v --debug --version --no-self-upgrade 2>&1 | tail -n1 | grep "^certbot $INITIAL_VERSION$" ; then
```
ohemorange: What was this even testing?

bmw: Good question. I had to go digging to find the answer.

Full answer is at https://github.com/certbot/certbot/pull/5392/files#r160726626 but the short(er) version is when we wrote the code to have certbot-auto rebootstrap on CentOS 6, the script's behavior wasn't great if somehow separate virtual environments existed at both the new and old virtual environment paths. This code was setting up a 2nd virtual environment at the new path to make sure our behavior was sane when both exist and we rebootstrap down below.

How about I add this code back, but set `VENV_PATH` to the old virtual environment path so we can keep this test? I'll make this change and rerun test farm tests to make sure everything still works.

bmw: I did this. In my opinion, the test seems largely unnecessary, but we currently have it and it may catch unexpected behavior so I feel like we might as well keep it.

I also added back a check for the symlink but only when we're running on CentOS 6 and created the 2nd venv like this.

bmw: Oh. And I ran test farm tests on these changes and they passed.

PR #6833, opened 2019-03-08
[Windows] Security model for files permissions - PART 1

adferrand:
This PR is the first part of #6497 to ease the integration.

We start with the core module of the PR, `certbot.compat.security`.

NB: Package name is temporarily `certbot.new_compat` and will be changed when the code is connected to the rest of certbot, and old `certbot.compat` module is relocated to `certbot.compat.misc`.


adferrand:
I will do for some specifics, but I am not planning to do extensive unit tests here. Most of the logic is already covered by higher level tests.

adferrand:
I added some unit tests for security module.

adferrand:
In favor of #6893

PR #6839, opened 2019-03-10
Construct the sanitized, pinned and hashed requirements file for certbot-auto

### LINKED ISSUE #6583: Update dependencies pinned in certbot-auto
The relevant file here is https://github.com/certbot/certbot/blob/master/letsencrypt-auto-source/pieces/dependency-requirements.txt. I think that many of these dependencies haven't been updated in a very long time.

Doing this unfortunately is a little tricky because:

* Dependencies can be added conditionally using project's arbitrary `setup.py` files. The dependencies could change based on the Python version or the OS it is run on and this file must contain any possible dependencies we need, otherwise, `certbot-auto` will blow up if it needs a dependencies we don't have pinned.
* We need to make sure we don't drop support for any non-EOL'd OSes we currently support.
* This will likely uncover new warnings which will cause our tests that error on warnings to fail. This would require fixing up any code in a way that removes the warning but also works with older versions of our dependencies used in our "oldest" tests.

The way I might do this is running `pip install certbot-apache certbot-nginx` on many different OSes using their version of both Python 2 and Python 3, running `pip freeze` and recording the result. Ideally, you should also make sure that the installed copy of Certbot works either by running tests manually or running our integration tests.

I'd probably keep `python-augeas` pinned at it's current version because it can have issues finding Augeas in newer versions that don't exist at it's current version.

After that all of output of `pip freeze` would need to be combined into a single file with duplicate dependencies removed. Then you can use `hashin` to get hashes as described at the top of the linked `dependency-requirements.txt` file.

Our test farm tests that use `certbot-auto` should also be run on any changes here which might help us catch problems. If anyone needs help with this piece, please ping me.

Bonus points should be given if whoever takes this:

* Investigates the changed dependencies' changelogs to look for dropped support we may be relying on.
* Writes a script to automate all of this in the future. Regardless of the approach to packaging we take, this is something we will have to do repeatedly in the future.

If the latter isn't done now, I think another issue should be open to track doing that going forward.

@erikrose, you may have some thoughts on this issue and if so, I'd love to hear them.

adferrand:
Fixes #6583.

This PR proposes an implementation of the process described in #6583.

Its goal is to calculate and generate the up-to-date requirements file that certbot-auto use to setup the venv that will host certbot. This list must be compatible with all environments supported by Certbot. All pinned versions must be hashed to control the authenticity of installed packages. Finally, any customization needed by Certbot must be included.

The result is the script `letsencrypt-auto-source\rebuild_dependencies.py`. This python script will generate this file on the path given as its first argument. 

To do so, the script will start dockers for various distributions, and install the current development version of certbot on it. By calling pip with unpinned versions, all required dependencies are installed in their most up-to-date state for each distribution. All requirements are retrieved from the dockers and merged, then hashed using hashin to give the final requirements file. 

Version conflicts are properly detected, and displayed both on stderr and in the generated file. Authoritative constraints can be added to solve these dependencies, and to apply specific versions for some packages if required. 

Also in this PR the current result of this script execution is added to the current development version of certbot-auto.


bmw:
@erikrose, do you have the interest/cycles to review this? This PR is meant to resolve #6583 whose goal was to give us a way to easily update the dependencies pinned in certbot-auto.

erikrose:
If it can wait till next week, I'm happy to!

bmw:
Thanks that'd be great!

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _process_one_distribution(distribution, verbose):
+    print('===> Gathering dependencies for {0}.'.format(distribution))
+    workspace = tempfile.mkdtemp()
+    script = os.path.join(workspace, 'script.sh')
+    authoritative_requirements = os.path.join(workspace, 'requirements.txt')
+    cid_file = os.path.join(workspace, 'cid')
+
+    try:
+        with open(script, 'w') as file_h:
```
erikrose: Certainly not a must-do, but it might be worth factoring out the file-reading and -writing routines. A few times, you write a string to a file and close it. A few times, you read an entire file and store it. Making those into routines would at least keep a few (reused) filehandle vars from blowing around in the statespace of the function.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+    with open(dest_file, 'r') as file_h:
+        print(file_h.read())
+
+
+def _gather_dependencies(dest_file, verbose):
+    dependencies_map = {}
+
+    for distribution in DISTRIBUTION_LIST:
+        results = _process_one_distribution(distribution, verbose)
```
erikrose: I'm allergic to the term "process" in programs. It doesn't really tell the reader anything, unless you've carefully and explicitly defined "process" somewhere. Might I suggest "_requirements_from_one_distribution"?

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+/opt/eff.org/certbot/venv/bin/python letsencrypt-auto-source/pieces/create_venv.py /tmp/venv "$PYVER" "1"
+
+/tmp/venv/bin/python letsencrypt-auto-source/pieces/pipstrap.py
+/tmp/venv/bin/pip install -e acme -e . -e certbot-apache -e certbot-nginx -c /tmp/requirements.txt
+/tmp/venv/bin/certbot plugins
+/tmp/venv/bin/pip freeze >> /tmp/workspace/results
+"""
+
+
+def _process_one_distribution(distribution, verbose):
```
erikrose: I could use a docstring here to delimit the responsibilities of this function, especially since it's just called "process". Docstrings on the other functions would be helpful as well.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+#   - finally current set of dependencies is extracted out of the docker using pip freeze
+SCRIPT = """\
+#!/bin/sh
+set -e
+
+cd /tmp/certbot
+letsencrypt-auto-source/letsencrypt-auto --install-only -n
+PYVER=`/opt/eff.org/certbot/venv/bin/python --version 2>&1 | cut -d" " -f 2 | cut -d. -f1,2 | sed 's/\.//'`
+
+/opt/eff.org/certbot/venv/bin/python letsencrypt-auto-source/pieces/create_venv.py /tmp/venv "$PYVER" "1"
```
erikrose: No sense in quoting a static `1`

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+/tmp/venv/bin/python letsencrypt-auto-source/pieces/pipstrap.py
+/tmp/venv/bin/pip install -e acme -e . -e certbot-apache -e certbot-nginx -c /tmp/requirements.txt
```
erikrose: I believe it's conventional to call it constraints.txt rather than requirements.txt. That would also open up the name "requirements.txt" for the output, rather than the less descriptive "results". Plus, it's consistent with your name `AUTHORITATIVE_CONSTRAINTS`.

letsencrypt-auto-source/rebuild_dependencies.py
erikrose: `results` isn't very illuminating either.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+        if os.path.isfile(cid_file):
+            with open(cid_file, 'r') as file_h:
+                cid = file_h.read()
+            subprocess.call(['docker', 'kill', cid], stdout=None, stderr=None)
+        shutil.rmtree(workspace)
+
+
+def _insert_results(dependencies_map, results, distribution):
+    refined_results = []
+    for result in results.split(os.linesep):
```
erikrose: Did you know about str.splitlines()? It should be fine even if we were running Windows as one of the docker instances, unless we were trying to armor against truly pathological output from `pip freeze`. Perhaps even better, you could use the implicit iteration offered by file objects when you first read the results in, something like this:

```python
return file_h.readlines()
```

That would have the advantage of using the container's OS's native line endings only.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+        match = re.match(r'(.*)==(.*)', result)
```
erikrose: Maybe speed doesn't matter here, but this is going to do quite a bit of backtracking: the `.*` will read all the way to the end of the string, find it can't match the pattern, then try every substring starting from the end all the way until it finally finds the "==". Perhaps excluding "=" from consideration would work, like `([^=]+)==([^=]+)`. (I've also changed `*` to `+`, since we do expect something to be there.)

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+        if match:
+            package = match.group(1)
```
erikrose: `package, version = match.groups()` might be more fun. :-)

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+            version = match.group(2)
+            if not any(dep in package for dep in ['acme', 'certbot', 'pkg-resources']):
```
erikrose: Is `dep in package` deliberate? I'm envisioning some future dependency being called "t**acme**thod" or something and throwing this off.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+            return file_h.read()
+    finally:
+        if os.path.isfile(cid_file):
+            with open(cid_file, 'r') as file_h:
+                cid = file_h.read()
+            subprocess.call(['docker', 'kill', cid], stdout=None, stderr=None)
+        shutil.rmtree(workspace)
+
+
+def _insert_results(dependencies_map, results, distribution):
```
erikrose: Part of this function's docstring should be an example of its return dict's shape.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+    refined_results = []
+    for result in results.split(os.linesep):
+        match = re.match(r'(.*)==(.*)', result)
+        if match:
+            package = match.group(1)
+            version = match.group(2)
+            if not any(dep in package for dep in ['acme', 'certbot', 'pkg-resources']):
+                dependencies_map.setdefault(package, []).append((version, distribution))
+
+    return refined_results
```
erikrose: The return value is unused.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+            requirements.append((package, list(reduced_versions)[0]))
+
+    requirements.sort(key=lambda x: x[0])
+    return requirements, conflicts
+
+
+def _reduce_versions(versions):
+    version_map = {}
+    for version in versions:
+        version_map.setdefault(version[0], []).append(version[1])
```
erikrose: There's just too much anonymity here for me to keep track of. Can we introduce some intermediate names, like `for version, distro in versions:`? Also, a docstring would help me figure out the purpose of this function.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+    return version_map
+
+
+def _write_requirements(dest_file, requirements, conflicts):
+    print('===> Calculating hashes for the requirement file.')
+    if os.path.exists(dest_file):
+        os.remove(dest_file)
```
erikrose: There shouldn't be a need to do this, as opening a file for writing will implicitly truncate it.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+                               '--requirements-file', dest_file])
+
+    if conflicts:
+        with open(dest_file, 'a') as file_h:
+            file_h.write('\n## ! SOME ERRORS OCCURRED ! ##\n')
+            file_h.write('\n'.join('# {0}'.format(conflict) for conflict in conflicts))
+            file_h.write('\n')
+
+    with open(dest_file, 'r') as file_h:
+        print(file_h.read())
```
erikrose: Perhaps it's not important for this program yet, but I generally find it advantageous to keep side effects (like non-debug prints) out of internal functions. It allows for reuse and easy testing. The same applies to the print at the end of _gather_dependencies().

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+    for distribution in DISTRIBUTION_LIST:
+        results = _process_one_distribution(distribution, verbose)
+        _insert_results(dependencies_map, results, distribution)
+
+    requirements, conflicts = _process_dependency_map(dependencies_map)
+
+    _write_requirements(dest_file, requirements, conflicts)
+
+    dest_file_abs = dest_file if os.path.isabs(dest_file) else os.path.abspath(dest_file)
```
erikrose: I bet you could always just call abspath.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -483,6 +483,14 @@ else:
 UNLIKELY_EOF
```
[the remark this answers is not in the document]
adferrand: Pycharm tricked me, by silently transforming this folder into a python package. Removing it.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+    # Using an older version of mock here prevents regressions of #5276.
+    'mock': '1.3.0',
+    # Too touchy to move to a new version. And will be removed soon
+    # in favor of pure python parser for Apache.
+    'python-augeas': '0.5.0',
+}
+
+
+# ./certbot/letsencrypt-auto-source/rebuild_dependencies.py (2 levels from certbot root path)
+CERTBOT_REPO_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+
+# The script will be used to gathering dependencies for a given distribution.
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+#   - certbot-auto is used to install relevant OS packages, and setup an initial venv
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+#   - then this venv is used to consistently construct an empty new venv
+#   - once pipstraped, this new venv pip install certbot runtime (including apache/nginx),
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+#     without pinned dependencies, and in respect with input authoritative requirements
```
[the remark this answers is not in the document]
adferrand: Yes, not always easy for non native English speakers to use appropriately idiomatic expressions.

Done.

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+#   - `certbot plugins` is called to check we have an healthy environment
+#   - finally current set of dependencies is extracted out of the docker using pip freeze
```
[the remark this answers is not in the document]
adferrand: Your welcome!

letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -0,0 +1,221 @@
...
+                        help='path for the generated requirements file')
+    parser.add_argument('--verbose', '-v', action='store_true',
+                        help='verbose will display all output during docker execution')
+
+    namespace = parser.parse_args()
+
+    try:
+        subprocess.check_output(['hashin', '--version'])
+    except subprocess.CalledProcessError:
+        raise RuntimeError('Python library hashin is not installed in the current environment.')
```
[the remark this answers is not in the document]
adferrand: Try to fail fast here indeed, otherwise you could process during a long time before failing at the very end ...

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Your are right. Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Yes, I improved the names using your suggestions.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: You are write. These functions make a lot of things. I add a docstring for each of them.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Indeed. Constraints is better as a name. Updated, and also the meaningless `results`.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Totally. I change the logic to use `readlines`. Thanks for the tip about `splitlines`!

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: It is a more focused regex. I test it.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Deconstructing arrays is life. Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Yes it is deliberate, because depending of pip version, using editable installation, the package name can be contained in some horrible url involving the git remote currently set for the repository from which the package is installed. And sometimes not, it is just `package`, even in editable mode.

Considering that these specific packages are very unlikely to be included in other packages, considering there specificity (it is not `tool`, `lib` and so on), I think it is relatively safe to do that.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Removed

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Indeed. The header written just after will unconditionnally truncate the file if it already exists.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Done.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: Long live to idempotent functions!

bmw:
FWIW, I ran our full suite of test farm tests with the updated dependencies found in this PR and everything passed!

adferrand:
That is great! I am testing the new implementation after correction to be sure that we keep the same results, and give update about here.

adferrand:
I confirm that the modifications done here, after review, does not break the script, and generates exactly the same file.

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
erikrose: Dastardly! ;-)

letsencrypt-auto-source/rebuild_dependencies.py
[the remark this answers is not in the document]
erikrose: Your English is far better than my French! :-)

letsencrypt-auto-source/rebuild_dependencies.py
erikrose: Hmm, that makes me think 2 thoughts:

1. There are entire URLs now in which there could be a false match to something short like "acme".
2. URLs can technically contain `=`, so my suggested regex isn't great (though at least it will crash if surprised).

erikrose: In fact, old-timey egg URLs like -e git://github.com/mozilla/elasticutils.git#egg=elasticutils will certainly have equal signs.

erikrose: So maybe a negative lookahead assertion that forbids "==" is the best approach here.

erikrose: I'd also still like to tighten up the test for acme/certbot/pkg-resources. Won't we know the editable URL for our stuff, leaving only pkg-resources?

adferrand: Anyway, since the modules installed in editable mode can only be modules from certbot projects, I should first filter out any editable entry, old style or new style.

Then we can extract with the regex you gave, that is technically correct in this case and more efficient, and make an exact test match for the packages we want to remove.

erikrose:
That sounds good.

adferrand:
Added the logic:
* filter out entries starting with `-e` (same approach than in merge-requirements.py`)
* then match against regex the entries to get the package and the version, separated by `==`

It covers both situations when editable requirements are output by `pip freeze`, successfully removes the packages related to certbot, and still gives the correct requirements file.

letsencrypt-auto-source/rebuild_dependencies.py
adferrand: See main conversation for resolution.

Issue #6849, opened 2019-03-12
Remove TLS-SNI-01 tests and code


bmw:
Let's Encrypt is turning off TLS-SNI-01 tomorrow, this challenge is no longer part of the ACME spec, and we have deprecated our support for a while. I think it's time to delete a lot of code.

I think we should:

- [x] Remove all tests using TLS-SNI-01
- [x] Remove any special code configuring boulder to still offer TLS-SNI-01 during tests
- [x] Remove code in Certbot and its plugins for performing TLS-SNI-01
- [x] Remove unit tests in Certbot and its plugins for performing TLS-SNI-01

And that may be it for now.

I think we should be a little careful in `acme`. Do we really want code like:
```
if isinstance(foo, acme.challenges.TLSSNI01):
```
to suddenly blow up? Code like this has never printed deprecation warnings.

I'm happy to talk about the `acme` stuff, but I think we may want to talk about how to remove that separately.

bmw:
Also, this should probably be done incrementally in separate PRs because the number of lines deleted here is going to be well over 1000.

adferrand:
I will happily take this charge, if you want.

bmw:
The first piece of this we should do is integration tests because boulder is planning on removing their TLS-SNI-01 code as well.

bmw:
Also also, another thing I think we should be careful about is references to TLS-SNI in configuration files. In particular, I don't think we should suddenly blow up if `tls-sni` is included in the preferred challenges list.

adferrand:
Considering your comments, here is the plan I am thinking.

PR1: Remove integration tests
PR2: Remove unit tests
PR3: Make tls-sni-01 execute a http-01 instead on Certbot, with a deprecation warning
PR4: Make any call of tls-sni from acme generate a deprecation warning 

Then, with one or two releases after 
PR5: Make any call of tls-sni a no-op with failure from certbot

Finally, one or two releases more:
PR6: Remove tls-sni-01 from acme. No code referring to tls-sni-01 on certbot and acme anymore.

adferrand:
I separated PR5 and PR6 to let acme, as a library, behave with a smooth deprecation path, because in theory, other acme CA servers could still offer tls-sni-01.

But we can be faster about our droping support because it is a security issue, and make PR6 on the same time than PR5 if you want.

bmw:
> PR1: Remove integration tests

[...]

PR #6856, opened 2019-03-13
Remove tls-sni challenge in standalone plugin

adferrand:

It removes the capability to make tls-sni-01 challenges in the standalone plugin.


adferrand:
Yes I did not want to remove too much things, as this particular plugin is very likely to get this support.

PR #6854, opened 2019-03-13
Remove tls-sni in compatibility tests

adferrand:
This PR is a part of the tls-sni-01 removal plan described in #6849.

It replaces tls-sni-01 challenges use in the compatibility tests for Apache and Nginx by http-01 challenges.

PR #6858, opened 2019-03-14
Remove tls sni in apache plugin

adferrand:

It removes the capability to make tls-sni-01 challenges in the apache plugin.


adferrand:
Blocking on #6854

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -812,31 +812,24 @@ def test_add_name_vhost_if_necessary(self):
         self.assertEqual(self.config.add_name_vhost.call_count, 2)
 
     @mock.patch("certbot_apache.configurator.http_01.ApacheHttp01.perform")
-    @mock.patch("certbot_apache.configurator.tls_sni_01.ApacheTlsSni01.perform")
     @mock.patch("certbot_apache.configurator.ApacheConfigurator.restart")
-    def test_perform(self, mock_restart, mock_tls_perform, mock_http_perform):
+    def test_perform(self, mock_restart, mock_http_perform):
```
bmw: Can we simplify this test a bit more since the Apache plugin only supports 1 challenge type (and that's not likely to change anytime soon)?

certbot/cli.py
```diff
@@ -1133,6 +1133,10 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):  # pylint: dis
         ["testing", "standalone"], "--http-01-address",
         dest="http01_address",
         default=flag_default("http01_address"), help=config_help("http01_address"))
+    helpful.add(
```
bmw: Why are these changes in both this PR and #6857? This value or `--tls-sni-01-port` doesn't seem to be used in the Apache plugin so I think these changes should be removed from this PR and left in the Nginx one.

adferrand: Yes I did not remove it yet, after seeing that it was not useful for Apache.

certbot-apache/certbot_apache/tests/configurator_test.py
adferrand: Done:

Issue #6850, opened 2019-03-12
Fix remaining broken tests on Windows


adferrand:
When I started to port Certbot on Windows, a lot of unit tests or the logic tested by them were not working on this platform.

In order to still have a working CI during the port, all failing tests where marked with a temporary decorator, `@broken_on_windows` to skip their execution. At this time, there is 40 units tests left in this state.

This issue is done to track the complete removal of broken tests on Windows, either by fix the test or code logic, or by skipping definitively some of theses tests for a good reason with `@skip_on_windows`.

adferrand:
Once mentioned PRs above are merged, plus logic from #6497 and #6859, all unit tests will be compatible with Windows, or skipped for a good reason.

bmw:
The last pieces of this are (at least partially) blocked on landing #6895.

PR #6865, opened 2019-03-16
[Windows] Fix some unit tests

adferrand:
This PR is a part of the effort to remove the last broken unit tests in certbot codebase for Windows, as described in #6850.

This PR fixes various unit tests on Windows, whose resolution was only to modify some logic in the tests, or minor changes in certbot codecase impacting Windows only (like handling correctly paths with DOS-style).


certbot/tests/hook_test.py
```diff
@@ -37,7 +37,7 @@ def _call(cls, *args, **kwargs):
         from certbot.hooks import validate_hook
         return validate_hook(*args, **kwargs)
 
-    @util.broken_on_windows
+    @util.skip_on_windows(reason='Windows does not use mode to control if a file is executable.')
```
bmw: > Windows does not use mode to control if a file is executable.

Will it once your changes in the Windows security PR land? If not, should it?

If we don't take that route, I think we may want to fix up this functionality so it works on Windows. Otherwise I think we'll fail much later in execution trying to run nonexecutable hooks on Windows.

adferrand: No it will not change. The executability in Windows is not defined by the mode, but by the file extension. Here is a demonstration (to be run on Windows of course):
```
#!/usr/bin/env python
import os
import subprocess
import tempfile

CONTENT = '''\
echo Hello World!
'''

with tempfile.TemporaryDirectory() as path:
    test1 = os.path.join(path, 'test1.bat')
    test2 = os.path.join(path, 'test2.txt')

    with open(test1, 'w') as file_h:
        file_h.write(CONTENT)

    with open(test2, 'w') as file_h:
        file_h.write(CONTENT)

    print('Executable mode for test1 is: {0}'.format(os.access(test1, os.X_OK)))
    print('Executable mode for test2 is: {0}'.format(os.access(test2, os.X_OK)))

    try:
        subprocess.check_call([test1])
        print('test1 is effectively executable!')
    except OSError:
        print('test1 not so executable finally...')

    try:
        subprocess.check_call([test2])
        print('test2 is effectively executable!')
    except OSError:
        print('test2 not so executable finally...')
```

So there is not implementation planed for now to fix this functionality on Windows. The idea would be to check the file extension in this case. Would like to create an issue to track this down?

bmw: Interesting. Doesn't file system permissions also play a part though? https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc783530(v=ws.10) describes execute permissions.

adferrand: Updated example, in a venv with Certbot installed using my PR for Windows file permissions:
```python
#!/usr/bin/env python
import os
import subprocess
import tempfile

from certbot.compat import security

CONTENT = '''\
echo Hello World!
'''

with tempfile.TemporaryDirectory() as path:
    test1 = os.path.join(path, 'test1.bat')
    test2 = os.path.join(path, 'test2.txt')

    with open(test1, 'w') as file_h:
        file_h.write(CONTENT)

    with open(test2, 'w') as file_h:
        file_h.write(CONTENT)

    print('Executable mode for test1 is: {0}'.format(os.access(test1, os.X_OK)))
    print('Executable mode for test2 is: {0}'.format(os.access(test2, os.X_OK)))

    try:
        subprocess.check_call([test1])
        print('test1 is effectively executable!')
    except OSError:
        print('test1 not so executable finally...')

    try:
        subprocess.check_call([test2])
        print('test2 is effectively executable!')
    except OSError:
        print('test2 not so executable finally...')

    security.apply_mode(test1, 0o600)

    print('Executable mode for test1 is: {0}'.format(os.access(test1, os.X_OK)))
    print('Executable mode with security module for test1 is: {0}'.format(
        any([security.check_mode(test1, 0o700), security.check_mode(test1, 0o400), security.check_mode(test1, 0o000)])))

    try:
        subprocess.check_call([test1])
        print('test1 still executable without executable bit!')
    except OSError:
        print('test1 is not not executable...')
```
```shell
(venv3) # py test.py
Executable mode for test1 is: True
Executable mode for test2 is: True

(venv3) C:\Dev\test>echo Hello World!
Hello World!
test1 is effectively executable!
test2 not so executable finally...
Executable mode for test1 is: True
Executable mode with security module for test1 is: False
test1 is not not executable...
```

Indeed, in this case, if you remove the executable flag from a file that has a executable file extension, it cannot be run anymore.

So both the file extension and executable bit will be involved.

bmw: I personally think that eventually checking the executable bit on Windows is valuable. The purpose of this code is to make running Certbot where any of the `--*hook` flags are paths to non-executable files fail early in execution rather than much later in the process when it actually goes to run the hook.

If there's a sane way we can query Windows to see if the given file extension is executable, I may be interested in adding that check too.

Since checking the execute bit is blocked on your Windows security work, I'd personally like to see us leave `broken_on_windows` on this method and readdress the problem when that work has landed. What do you think?

adferrand: Ok, we do that.

PR #6867, opened 2019-03-16
Fix main logic for windows

adferrand:
This PR is a part of the effort to remove the last broken unit tests in certbot codebase for Windows, as described in #6850.

This PR proposes a solution to fix the unit tests failing in `certbot.tests.main_test`.

The problem is raised when the `main` method of `certbot.main` is executed not as the entrypoint of the current python process (like when certbot is normally called from the CLI), but multiple times in a particular point of the python process, like it is done during unit tests execution by pytest.

Indeed, the method `main` leaves side-effects after its execution, through `certbot.log`, that are supposed to be cleaned up by the end of the python process, in the assumed design that certbot is called as a main python process, not as a library in another process. Indeed, the `main` function modifies handlers of current logging system, without closing them properly upon exit, and monkeypatch `sys.excepthook` to ensure that a proper formatting is done when some output is written to stderr. In current process, cleaning is made by registering an `atexit` handler, that will be called on the python process finishes.

This is working on Linux even during tests, because it is quite lenient with opened file descriptors, but this is not passing on Windows.

It is the third time that `main` method in `certbot.main` creates problems. First was suprious tests failures on Windows when temporary certbot configs directory are involved, second was for the Letstest execution on the AWS instances.

So I would like to propose a more definitive solution: make `main` cleaning its own context upon exit, allowing both unit tests to call it multiple time without issue on Windows and Linux, and allowing an existing Python process to call it without hidden side affects for the remaining of its execution.

The idea is to use context managers. Context managers are superior to atexit handlers, because they ensure, on the same conditions (current process is not `SIGKILL`ed), to clean a context when leaving it, not when the python process finishes.

So faulty methods in `certbot.log`, `pre_arg_parse_setup` and `post_arg_parse_setup`, becomes context manager, and `main` call them as it (`with` statement), to ensure a proper clean when the function exit. Consequently, existing mitigations in `certbot.tests.util.TempDirTestCase` and `atexit` handlers become useless, and so are removed.

There are two tricky part on which I made a specific effort:
* `post_arg_parse_setup` may fail because of lack of permissions, but can be non-critical if `certbot plugin` is called. This case is handled by yielding from `post_arg_parse_setup` context entering the error reference, if an exception happened, or `None`. Then `main` can decide depending of the command issue to raise the exception, or ignore it.
* specific monkeypatching of `sys.excepthook` are handled, and properly restored upon context exit of `pre_arg_parse_setup` and `post_arg_parse_setup`.

### DIFF
--- a/certbot/log.py
+++ b/certbot/log.py
@@ -54,6 +54,9 @@ def pre_arg_parse_setup():
     `sys.excepthook` to properly log/display fatal exceptions.
 
     """
+    original_handlers = set(logging.getLogger().handlers)
+    original_except_hook = sys.excepthook
+
     temp_handler = TempHandler()
     temp_handler.setFormatter(logging.Formatter(FILE_FMT))
     temp_handler.setLevel(logging.DEBUG)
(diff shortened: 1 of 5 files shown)


bmw:
I've had in my notes for a while now to take a look at this PR when I had the time and I'm just getting around to it.

While I definitely see the value in something like this, I'm not initially totally onboard with the idea. I have yet to see any Python application do anything like this to try and cleanup its logging config before it exits. For stuff unrelated to ACME and TLS configuration, I personal try to avoid doing out of the ordinary things with Certbot as much as possible and just do what everyone else does. We have enough work to do there without reinventing solutions to common problems.

Googling around I didn't find much, but the suggestions I did find recommended handling this in testing code and I largely agree with that approach. Making the change here would add complexity to our production code for the sole purpose of testing and I'd rather keep that complexity in the tests.

I'm happy to talk more about the different approaches we can take and we could probably find a compromise, but to be honest, given our resources and the fact you found another solution to this, I'm tempted to leave things as is.

With that said, I think another thing that would help is rewriting tests on modules like `cli.py` and `main.py`. (I swear we had an issue for that but quickly searching, I couldn't find it.) The tests are overly complex and in many cases should be system tests. By reducing the amount we call into main to simple unit tests on `certbot.main.main()` (if we even want any tests like that), I think this problem also goes away.

Issue #6813, opened 2019-03-04
Fix spurious Apache test failure


bmw:
Over the weekend, our Apache coverage tests spuriously failed. See https://travis-ci.com/certbot/certbot/jobs/181789080.

The lines it failed on were:https://github.com/certbot/certbot/blob/6ff101dcbbe66943487633391a4f03f27bcc6aaa/certbot-apache/certbot_apache/configurator.py#L876-L887

I think depending on the order the vhost directives are found by Augeas, these lines might not be executed.

Running tests locally, we seem to only hit these lines during while running `setUp` during various other tests. I did not see a test explicitly try to test these lines directly.

I think we should add an explicit test for these lines in a way that doesn't depend on things like the order the vhosts are found.

bmw:
We saw this issue again last night at https://travis-ci.com/certbot/certbot/jobs/182971280.

I think no longer rerunning our tests 3 times in Travis may have been the change causing us to start seeing this. I think we should fix this soon.

adferrand:
It is interesting to see how the rerun feature was hiding a problem here. Indeed it seems that it is due to the inconsistent parsing on Apache config that @joohoi found from an unrelated troubleshooting.

With the rerun the inconsistency is mitigated. Without it, it would have participated to understand more quickly what was going on.

So to be clear, my preference on the resolution here would be not to reactivate the rerun, but to fix this inconsistency.

bmw:
> So to be clear, my preference on the resolution here would be not to reactivate the rerun, but to fix this inconsistency.

I agree with that, but I also don't like our tests failing. It can affect unrelated PRs and this test failing when our tests are run at night creates unnecessary noise for us to look into.

I'm planning on opening a PR to temporarily drop the required coverage on the Apache plugin until this is fixed. Let me know if anyone has any strong objections to that.

PR #6873, opened 2019-03-20
Add a test to ensure test coverage regardless of the vhost order

joohoi:
Add a new test to make sure that we are covering all the branches of `get_virtual_hosts()` regardless of the order that Augeas returns the found `VirtualHost` paths.

Fixes: #6813


bmw:
Now that #6880 has been resolved, we should be able to bump the minimum coverage in the Apache plugin back to 100% in this PR as well.

joohoi:
The coverage requirement is back at 100% now. I wasn't sure if we should add changelog entry for a purely test related PR, so I skipped doing so.

PR #6884, opened 2019-03-25
Document dropped tls-sni-01 support in plugins.

bmw:
We forgot to add changelog entries about this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


bmw:
Whoops. I forgot about the manual plugin.

adferrand:
In fact, maybe you could say `from all plugins`, and the list as an indication. We just completely ripped off tls-sni-01. Are you planning a specific changelog entry for the tls-sni flags removal ?

bmw:
> In fact, maybe you could say from all plugins, and the list as an indication. We just completely ripped off tls-sni-01.

Done.

> Are you planning a specific changelog entry for the tls-sni flags removal ?

I was going to suggest you include this in the relevant PR before I merged it.

PR #6890, opened 2019-03-27
Bump the min Certbot version for nginx plugin.

bmw:
nginx oldest tests started failing in https://travis-ci.com/certbot/certbot/builds/105958557 because the integration test script shared between the Certbot and nginx integration tests is now using `--https-port` which isn't supported in the older versions of Certbot used in the oldest nginx integration tests.

This fixes the problem by simply increasing the minimum version of Certbot required. This is more strict than we need to be, but it's the easiest fix to this problem and shouldn't hurt anything.


bmw:
I'm running tests to make sure this actually fixes the problem at https://travis-ci.com/certbot/certbot/builds/106041505.

PR #6893, opened 2019-03-27
[Windows] Security model for files permissions - STEP 1

adferrand:
This PR is the first part of #6497 to ease the integration, following the new plan propose by @bmw here: https://github.com/certbot/certbot/pull/6497#issuecomment-476912120

This step 1 refactor existing `certbot.compat` module into `certbot.compat.misc`, without any logic changed. Package `certbot.compat` will host the new modules that constitute the security model for Windows.


certbot/client.py
```diff
@@ -4,27 +4,24 @@
...
-import josepy as jose
-import OpenSSL
-import zope.component
 
 from acme import client as acme_client
 from acme import crypto_util as acme_crypto_util
 from acme import errors as acme_errors
 from acme import messages
-from acme.magic_typing import Optional  # pylint: disable=unused-import,no-name-in-module
+from acme.magic_typing import Optional  # type: ignore  # pylint: disable=unused-import,no-name-in-module
```
bmw: I don't think this `type: ignore` is needed. I ran tests locally without it and they passed.

adferrand: I heavily relies on the amazing refactor capabilities of Pycharm here, to make wide refactoring, but sometimes, it makes some little weird glitches. Anyway, it is fixed.

PR #6895, opened 2019-03-29
[Windows] Security model for files permissions - STEP 2

adferrand:
This PR is the second part of #6497 to ease the integration, following the new plan propose by @bmw here: https://github.com/certbot/certbot/pull/6497#issuecomment-476912120

This PR creates the module `certbot.compat.os`, that delegates everything to `os`, and that will be the safeguard against problematic methods of the standard module. On top of that, a quality check wrapper is called instead of pylint in the lint tox environment. This wrapper calls pylint and ensure that standard `os` module is no used directly in the certbot codebase.

Finally local oldest requirements are updated to ensure that tests will take the new logic when running.


bmw:
We very well may end up going this route, but I think you should hold off on doing any more work on this PR for a moment. If we're not going to modify the `os` module, it'd be so nice if we could get away without wrapping it all.

I/we need to think about it more, but I wonder if it'd be good enough to just modify your linter to simply check for strings referencing problematic functions like `os.mkdir`, `os.chown`, etc. The linter could then tell you what function you should be using instead in its error message. There are ways we could write code that wouldn't be caught by the linter in theory, but I'm wondering if it would ever happen in practice and if we can pull it off I think it results in a lot cleaner and simpler code. (I think we could write code to get around your current linter too, although it admittedly would be substantially more awkward than code to get around simple checks for strings like `os.mkdir`.)

I'd be interested to hear your thoughts on this, but I'd personally like to think about this a bit more before we go one direction or another.

adferrand:
I understand your concern. We the ongoing discussion, we are on the road to have a module that is design only for code quality purpose, and without effective logic. It is arguably weird.

If I go back to the core spec here, what I want is a code quality process, integrated in the CI, that prevent to use some functions in the certbot code. 

If it is a question of preventing or not a specific module, yes my linter does the job. But this approach will not be satisfactory for me on any more complex code quality process, in particular to what you say. 

Yes, I could track down any function name in the codebase that corresponds to a forbidden one from `os` module. However, this is subject to very brittle processing, and makes very 'artisanal'. One corner case I could immediately think of, is how to ignore references on the forbidden methods in the inline methods. I think that it could happen now, or later, as we are explicitly making some logics to not use them, so it is very likely that a corresponding inline comment will talk about the method!

So I have a more industrialized option in mind that follows the idea to have a code quality process outside of the code itself.

Indeed pylint is designed for code quality processing. It would be nice if pylint itself could detect usages of forbidden methods as errors. It would result in minimal impacts in the codebase, and also in the CI, since pylint continues to be called directly, and not through a wrapper. And also outside of the CI, as you can just call pylint directly. 

In fact it is possible, you can write custom rules. And in this case, the logic relies on traversing the AST python, which is the only reliable way to analyze a python code. 

What I have in mind is a simple tree traversing to detect any AST node whose name would be one of the forbidden method. Out of the box, you gain to ignore comments referring theses methods. You also gain advantage to see the methods names without any noise provoked by the python syntax itself (eg. `test = os.forbidden`, `forbidden()`, `from os import forbidden as hidden` are easily seen). And you do not need to make a very advanced analysis of the Python code logic to avoid false positives, because if you have some of them, you can just `# pylint: disable=` the faulty lines.

If the custom checker find an usage of one of the forbidden process, it will report the error with a message indicating that the method is forbidden, and that the relevant method from compat must be used instead. The CI will fail on `tox -e lint`, informing the developer about the bad practice. 

What do you think of that?

bmw:
@adferrand, I'll respond soon, but to keep this conversation in one place, I pinged Erik about this last night in IRC and our conversation was:
```
6:00:00 PM <@bmw> ErikRose_ (or other Pythonistas): I have a question for you when you get a minute.
6:00:32 PM <@bmw> To prep for Certbot on Windows, we're trying to prevent ourselves from calling problematic functions in the standard library such as `os.mkdir` which ignores any mode parameters on Windows.
6:22:58 PM <@bmw> The best idea we have is to wrap the os module in our own module, have it raise an exception if you try to use any problematic functions, and then write a linter to check for any imports of the `os` module outside of our wrapped version to force all calls to go through that.
12:53:12 PM <ErikRose> I was really hoping to say shutil has higher-level routines which are more portable, but nope.
12:53:18 PM <ErikRose> Now I'm checking pathlib.
12:53:21 PM <ErikRose> pathlib does feature mkdir and other such writing routines.
12:53:24 PM <ErikRose> It's pretty new. Maybe it's better-behaved on Windows?
12:53:27 PM <ErikRose> I don't have a Windows instance to test on, but, reading the pathlib source code, it looks like it just falls through to os.mkdir. So never mind.
12:53:30 PM <ErikRose> So yes, your strategy seems unorthodox but reasonable. Certbot has unorthodox needs, so that's hardly surprising.
12:53:33 PM <ErikRose> In general, we have no choice but to write against some better abstraction, and those combined runtime and (I would say optional, if the test suite is good) lint-time checks will keep us on the straight and narrow.
12:53:37 PM <ErikRose> The only thing to watch out for is that we'll probably have to mutate the os module itself, not just wrap it after the fact. Some 3rd-party modules will possibly import the original one before we have a chance to do our magic, and they will retain a reference to the original.
12:53:40 PM <ErikRose> We used to try to enforce import ordering stuff in Plone, and it turned into a hellscape of Nope.
12:53:43 PM <ErikRose> Hmm, is Zope simply Nope with the first letter rotated?, he muses.
12:56:41 PM <@bmw> We weren't planning on doing anything to try and fix the problem in our dependencies.
12:57:44 PM <@bmw> We were worried monkey patching a standard library module like that would cause strange problems somewhere, so we were just planning on fixing it for ourselves and then trying to make sure any of our dependencies have similarly fixed problems on their end.
12:58:00 PM <@bmw> In general, it shouldn't be too bad because it's really just interacting with the filesystem which most (or maybe all) of our dependencies don't do.
12:59:12 PM <@bmw> The other idea that adferrand[m] and I have been throwing around is rather than wrapping the os module and making devs use our wrapped version just to catch uses of bad functions is to try to solve that problem entirely through linting.
12:59:56 PM <@bmw> I think it'd always be possible for people to write insane code that gets around the linter, but presumably we wouldn't merge something that goes way out of its way to sneakily reference to a function.
1:01:10 PM <@bmw> If you feel like getting in the weeds, adferrand[m] and I wrote a bit about this in the comments on https://github.com/certbot/certbot/pull/6895, but really, if you have any high level thoughts about that alternative, I'd be interested to hear them.
1:03:20 PM <ErikRose> Squirreling away references to functions isn't so uncommon. pathlib does it extensively, to basically all of os. :-)
1:03:32 PM <ErikRose> So yeah, I think runtime catching is important.
1:04:24 PM <ErikRose> Don't be afraid to patch things. I once replaced sys: https://pypi.org/project/tracefront/.
1:04:55 PM <ErikRose> Or maybe it was traceback. Memory fails.
1:05:10 PM <ErikRose> I feel like I replaced sys at one point, to try to get something loaded as early as possible.
1:05:33 PM <@bmw> Interesting. Then do you think we should try to stop our dependencies from using any functions we don't like as well?
1:05:52 PM <ErikRose> You'll get that for free if you patch os.
1:05:53 PM <@bmw> I'm a bit worried about cases where they've already done something on their end to work around the problem and conditionally call functions when they know it's safe being an issue,.
1:06:07 PM <@bmw> Like, that's probably what we'll do.
1:06:17 PM <ErikRose> You're right to be concerned about that, but, as you say, not many libs are writing to the FS.
1:07:49 PM <@bmw> The other argument I made it against it in the past is messing up things like identities of objects (e.g. `os.chown is posix.chown`) which probably isn't a problem, but Certbot is built on a lot of code.
1:08:36 PM <@bmw> I'm a little hesitant to go down that road because I think many of the times we'll blow up at runtime due to our dependencies isn't really an issue and it's very hard to test due to all the versions of everything we support.
1:08:49 PM <ErikRose> That seems like too much of a corner case to worry about. You have your own venv that nobody else is using. As long as your tests are passing on all OSes, you should by definition be fine.
1:09:24 PM <ErikRose> (On second thought, the venv doesn't matter. You have your own *interpreter instance* that nobody else is using. :-) Except plugins, I suppose.)
1:10:58 PM <@bmw> Only 50% of our users use certbot-auto. Most of the rest use OS packages where Certbot depends on whatever versions of our dependencies they have available.
1:11:16 PM <@bmw> We regularly test on the oldest and newest versions of dependencies we support, but we don't test every combination in between.
1:11:24 PM <@bmw> Or the combinations used on popular OSes.
1:11:45 PM <ErikRose> I think you'd be okay either way, but, if I were designing it, I'd make my own cross-platform FS abstraction and write to that.
1:12:05 PM <ErikRose> Use a linter to make sure you don't accidentally import os directly in the future.
1:12:24 PM <ErikRose> Basically like Apache's apxs.
1:13:23 PM <@bmw> Yeah that's basically what we're doing, however, much of the `os` functionality is fine.
1:13:41 PM <ErikRose> Then just be careful, and have good tests.
1:13:58 PM <ErikRose> I thought the linter was a nice touch but was going above and beyond.
1:14:00 PM <@bmw> We're just trying to figure out the best way to ensure problematic functions aren't called whether that be by forcing all calls through some interface, monkey patching all of `os`, searching for problematic functions, etc.
1:14:16 PM <ErikRose> +1 for interface for me.
1:14:44 PM <@bmw> OK. Thanks for the input. That's what we currently have but I was hesitant to rip people away from their `os` module but maybe it's for the best.
1:15:01 PM <@bmw> rip people away from their os module when it's really just a few bad functions*
1:15:22 PM <ErikRose> If you're calling through some other abstraction, it's explicit, and nobody will get the wrong idea reading the code.
```

bmw:
Stepping back for a minute to talk about this at a higher level, there's a few things that have been proposed:

1. Make all of our code use `certbot.compat.os` and blow up at runtime if any bad functions are used.
2. Monkey patch `os` and have it do the same thing as 1.
3. Rely entirely on linting to ensure no bad functions in `os` is used.

I personally still think 2 is a bad idea especially with the current state of packaging. Certbot is built on an absolute ton of code. I think erroring out whenever any supported version of any dependency calls a function that sometimes has unsafe behavior will frequently break Certbot unnecessarily.

When this breakage does occur, what are we to do when we investigate the problem and realize the calling code was safe for some other reason? (For instance, imagine someone creates a temporary directory safely with the `tempfile` module and then calls `mkdir` inside it on Windows.) I think trying to keep Certbot working with this approach while supporting many dependency versions on many OSes will be a nightmare for little gain.

Then for me, it's between 1 and 3. While 3 is so tempting to me, forgetting the code for a minute, what behavior would we want from Certbot if somehow a bad function is called at runtime? I personally think it should blow up. It doesn't silently allow potentially unsafe behavior and erroring out makes the problem more apparent to the dev/user and we can quickly get it fixed.

With this in mind, while I don't love option 1 of forbidding the `os` module in all Certbot, I think it makes it significantly easier for us to find our mistakes and it's my preferred approach. With a combination of that and simple linting to ensure we never import the `os` module outside of `certbot.compat`, I think we can be pretty confident we won't ship code using these functions and if we ever do, we'll hear about it pretty quickly so we can fix it.

adferrand:
I agree also that option 3 is the most elegant thing to do, but option 1 will be the most efficient. A runtime error cannot be avoided, while a linting can fail to see a potential error. With option 1 the linting is limited to its bare minimum, the import of `os` package, making it resilient.

So I would go on option 1.

However, with what I discovered with option 3, and the ability to write custom checkers, I would like to give a try and make a better linting that a wrapper of pylint, concerning the detection of `os` import.

Your call here: I can do that for this PR, or on a later time. Since the current implementation does its job, I would be more for doing that later.

bmw:
I personally think it's worth [timeboxing](https://en.wikipedia.org/wiki/Timeboxing) yourself for a few hours to see if you can't write the `pylint` plugin. I played with writing such a plugin a bit on Friday and I think it should be pretty easy and will result in less code and configuration changes in this PR.

In fact, `pylint` already has the functionality we need. See [this](https://github.com/certbot/certbot/blob/232e0ea50fa7472803510260b8368aac33478ff8/.pylintrc#L269). I think the only thing we need to change is the error message. `os` isn't deprecated, it just shouldn't be used in Certbot and we should point people towards `certbot.compat.os`. 

If it proves difficult for some reason, we can potentially do this later but given our crusade against technical debt, I'm a little hesitant for us to merge the current approach when we're aware of something we think will be better.

adferrand:
You strike a cord here! I will implement that right away.

adferrand:
Block on #6905 to integrate the new code quality approach using a custom pylint checker.

adferrand:
Unblocking on #6905

adferrand:
With pylint upgrade done on #6905, we can again add correctly custom checkers.

So following the latest decisions we took about how to implement the windows file permissions, I created a new checker in `linter_plugin.py`. This checker ensures that the `os` standard module is not imported in the Certbot codebase, except of course on `certbot.compat.os`.

If a faulty import is done, the `lint` env tox execution will fail, with the CI, making it visible to the developer and the reviewer. The error generated invite the developer to use `certbot.compat.os`.

@bmw, the PR is ready for review.

certbot/compat/os.py
```diff
@@ -0,0 +1,20 @@
+"""
+This compat modules extends native capabilities of core os module to handle correctly platform
+specific operations (eg. chown, chmod, geuid).
+This module is intended to replace standard os module throughout certbot projects (except acme).
+"""
+from __future__ import absolute_import
+
+# Expose everything from standard os package to make current package a complete replacement of os.
+from os import *  # type: ignore  # pylint: disable=wildcard-import,unused-wildcard-import,redefined-builtin,os-module-forbidden
```
bmw: Is there a reason to do this when we also have the loop below?

certbot/compat/os.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Monkey patch ourselves to get os attributes that are not in __all__ (so not from os import *).
+import os as std_os  # pylint: disable=os-module-forbidden
+import sys as std_sys
+ourselves = std_sys.modules[__name__]
```
bmw: nit: This will almost certainly not be a problem, but can we pick a slightly more obscure name that we don't think could be used (internally) in the `os` module in the future. What about `certbot_os`?

certbot/compat/os.py
```diff
@@ -0,0 +1,20 @@
...
+for attribute in dir(std_os):
+    if not hasattr(ourselves, attribute):
```
bmw: If we get rid of the `from os import *` at the top, I was initially unsure if we needed this line, but I think we do. I don't think we want to overwrite attributes like `__doc__`, `__file__`, and `__name__`. Since I was initially unsure, can you add a comment explaining why we need this?

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
+        'E5001': (
+            'Forbidden use of os module, certbot.compat.os must be used instead',
+            'os-module-forbidden',
+            'Some methods from standard os modules cannot be used for security reasons on Windows: '
+            'the safe wrapper certbot.compat.os must be used instead in Certbot.'
+        )
+    }
+    priority = -1
+
+    # TODO: exclude acme module from the check
```
bmw: I think this TODO can be removed.

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
+    @check_messages('os-module-forbidden')
+    def visit_import(self, node):
+        if 'os' in [name[0] for name in node.names] and not _check_disabled(node):
+            self.add_message('os-module-forbidden', node=node)
+
+    # TODO: exclude acme module from the check
```
bmw: I think this TODO can be removed.

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
+from pylint.checkers.utils import check_messages
+
+
+# Module in theses packages can import os module.
+WHITELIST_PACKAGES = ['acme', 'certbot_compatibility_test', 'letshelp_certbot', 'lock_test']
+
+
+class ForbidStandardOsModule(BaseChecker):
+    """
+    This checker ensures that standard os module is not imported by certbot classes.
```
bmw: I'd either delete the word classes or change it to modules.

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
-        cls.locals['_fields'] = [nodes.EmptyNode()]
-
-
-MANAGER.register_transform(nodes.Class, _transform)
+from pylint.checkers import BaseChecker
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers.utils import check_messages
+
+
+# Module in theses packages can import os module.
```
bmw:
```suggestion
# Module in theses packages can import the os module.
```

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
+WHITELIST_PACKAGES = ['acme', 'certbot_compatibility_test', 'letshelp_certbot', 'lock_test']
+
+
+class ForbidStandardOsModule(BaseChecker):
```
bmw: Is this code based on checker for `deprecated-modules`? If so, can you state that in a comment here and add a link to the pylint code?

linter_plugin.py
```diff
@@ -1,31 +1,52 @@
...
+    """
+    This checker ensures that standard os module is not imported by certbot classes.
+    Otherwise a 'os-module-forbidden' error will be registered for the faulty lines.
+    """
+    __implements__ = IAstroidChecker
+
+    name = 'forbid-os-module'
+    msgs = {
+        'E5001': (
```
bmw: How was this number chosen? Are we sure it doesn't conflict with any message IDs currently used in pylint?

certbot/compat/os.py
adferrand: Initially it was the way I used to import all os module content into this module. It makes pylint, mypy, IDEs like pycharm very happy, because it allows them to understand, through this static import, that all public members of os are also here.

However, it appears that the `__all__` of `os` is incomplete in some python version (3.5 if I recall correctly), making this static import insufficient in this case.

So I added a dynamic import to cover all the cases, and kept this one to continue benefiting of the advantages of static import I described above. 

I think all of this worths a comment ^^

certbot/compat/os.py
adferrand: Yes, I will write about this, additionally to the comment needed as explained in the above discussion.

linter_plugin.py
adferrand: Done.

linter_plugin.py
adferrand: Done.

certbot/compat/os.py
```diff
@@ -0,0 +1,20 @@
+"""
+This compat modules extends native capabilities of core os module to handle correctly platform
+specific operations (eg. chown, chmod, geuid).
```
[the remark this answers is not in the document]
adferrand: I updated it.

linter_plugin.py
adferrand: Done (/classes/modules/).

certbot/compat/os.py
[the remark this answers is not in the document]
adferrand: Updated.

certbot/compat/os.py
adferrand: Comment added.

certbot/compat/os.py
adferrand: Updated.

linter_plugin.py
adferrand: No, this code is based on official documentation: http://pylint.pycqa.org/en/stable/how_tos/custom_checkers.html

linter_plugin.py
adferrand: I am, by checking the existing code of pylint. Core checkers use error message `E0***` of `E1***`. As a side note, this is the way to select the value in all tuto I could find: "do it manually".

adferrand:
I updated all relevant `setup.py` files, and added a specific paragraph in the official developer guide for `certbot.compat.os`.

linter_plugin.py
bmw: To help me verify this code, I went and looked at the [deprecated-module checker](https://github.com/PyCQA/pylint/blob/b20a2984c94e2946669d727dbda78735882bf50a/pylint/checkers/imports.py#L287). One thing the code you have here doesn't prevent is `import os.path` or `from os.path import foo`. If we wanted to check for this, doing so should be as simple as calling `startswith("os.")` on the string you're comparing with the constant "os" below.

I personally think we should do this. I don't think we have to as I'm not aware of any problematic functionality in the `os.path` module, however, I think it's cleaner and more Pythonic to have 1 way to do things which would be to use `certbot.compat.os.path` instead of also having the option to just use `os.path` (or telling people to not use `certbot.compat.os.path` and to make an exception for the `path` module and to use that directly instead.)

I think this means we should also fix up `certbot.compat.os` so `import certbot.compat.os.path` works like it does for the `os` module, but I think it's probably worth it to have a single path to access the code there.

What do you think?

If we do this, I think it's worth adding a comment that this class is loosely based on the checker for the deprecated-module warning in `pylint` and maybe adding a link to it. I think this gives us an implementation of very similar checks we can refer back to if we ever need to change the code here in the future.

linter_plugin.py
bmw: That's unfortunate but I verified there's no conflicts here.

linter_plugin.py
```diff
@@ -1,31 +1,50 @@
...
-
+"""Certbot PyLint plugin.
 http://docs.pylint.org/plugins.html
-
 """
-from astroid import MANAGER
-from astroid import nodes
+from pylint.checkers import BaseChecker
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers.utils import check_messages
```
bmw: I think we should remove this. It doesn't seem to be documented for use in custom checkers and looking at the code, I'm not sure how using the decorator helps us. If you have a different understanding of its purpose, please let me know, but it seems the [decorator](https://github.com/PyCQA/pylint/blob/293a7f5e6add850e2f127dd1680a6e43fb1a6949/pylint/checkers/utils.py#L430) just gives the functions a `checks_msgs` attribute which is used [here](https://github.com/PyCQA/pylint/blob/293a7f5e6add850e2f127dd1680a6e43fb1a6949/pylint/utils/ast_walker.py#L19). This allows pylint to skip running functions for disabled messages, however, I think we always want this check enabled so I think using the decorator adds slightly more complexity for no real gain.

linter_plugin.py
```diff
@@ -1,31 +1,50 @@
...
-        cls.locals['_fields'] = [nodes.EmptyNode()]
+def register(linter):
+    """Register this module as a Pylint plugin."""
+    linter.register_checker(ForbidStandardOsModule(linter))
 
 
-MANAGER.register_transform(nodes.Class, _transform)
+def _check_disabled(node):
+    module = node.root()
+    return any([package for package in WHITELIST_PACKAGES
```
bmw: super nit: I think we can avoid building a list and just use a generator here by removing the brackets.

linter_plugin.py
```diff
@@ -1,31 +1,50 @@
...
 
-    #if cls.slots() is not None:
-    #    for slot in cls.slots():
-    #        cls.locals[slot.value] = [nodes.EmptyNode()]
 
-    if cls.name == 'JSONObjectWithFields':
-        # _fields is magically introduced by JSONObjectWithFieldsMeta
-        cls.locals['_fields'] = [nodes.EmptyNode()]
+def register(linter):
+    """Register this module as a Pylint plugin."""
```
bmw: nit: We're not registering this module as a plugin are we?

linter_plugin.py
```diff
@@ -1,31 +1,50 @@
...
-
 """
-from astroid import MANAGER
-from astroid import nodes
+from pylint.checkers import BaseChecker
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers.utils import check_messages
+
+
+# Module in theses packages can import the os module.
```
bmw:
```suggestion
# Modules in theses packages can import the os module.
```

linter_plugin.py
```diff
@@ -1,31 +1,50 @@
...
+    Otherwise a 'os-module-forbidden' error will be registered for the faulty lines.
+    """
+    __implements__ = IAstroidChecker
 
+    name = 'forbid-os-module'
+    msgs = {
+        'E5001': (
+            'Forbidden use of os module, certbot.compat.os must be used instead',
+            'os-module-forbidden',
+            'Some methods from standard os modules cannot be used for security reasons on Windows: '
```
bmw:
```suggestion
            'Some methods from the standard os module cannot be used for security reasons on Windows: '
```

certbot-dns-cloudflare/setup.py
```diff
@@ -8,7 +8,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.21.1',
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-dns-digitalocean/setup.py
```diff
@@ -8,7 +8,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.21.1',
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-dns-rfc2136/setup.py
```diff
@@ -8,7 +8,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.21.1',
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-dns-route53/setup.py
```diff
@@ -7,7 +7,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.25.0',
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-dns-google/setup.py
```diff
@@ -8,7 +8,7 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.21.1',
```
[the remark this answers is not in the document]
adferrand: Done.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
 .. _PEP 8 - Style Guide for Python Code:
   https://www.python.org/dev/peps/pep-0008
 
+Use ``certbot.compat.os`` instead of standard ``os``
```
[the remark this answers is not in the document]
adferrand: Yes you are right. I removed the word standard. As it is a header, I think it should be as short as possible, so I choose to not add anything more, as everything is explained in the paragraphs below.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
...
+====================================================
+
+Python standard ``os`` module lacks of a full support for several Windows
```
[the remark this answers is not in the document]
adferrand: Done.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
...
+security features about file permissions (eg. DACLs). However several sensible
+files handled by Certbot (eg. private keys) need a strongly restricted access
+on both Linux and Windows.
+
+This problem is asserted by the ``certbot.compat.os`` module: it wraps the standard
+``os`` module, and forbid usage of methods that lacks support for these Windows
+security features.
+
+As a developper, you must use ``certbot.compat.os`` in every place you would need
```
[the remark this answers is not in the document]
adferrand: French "développeur" toggled here ^^ Corrected.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
 .. _PEP 8 - Style Guide for Python Code:
   https://www.python.org/dev/peps/pep-0008
 
+Use ``certbot.compat.os`` instead of standard ``os``
+====================================================
+
+Python standard ``os`` module lacks of a full support for several Windows
+security features about file permissions (eg. DACLs). However several sensible
```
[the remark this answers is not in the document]
adferrand: Done.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
...
+files handled by Certbot (eg. private keys) need a strongly restricted access
+on both Linux and Windows.
+
+This problem is asserted by the ``certbot.compat.os`` module: it wraps the standard
+``os`` module, and forbid usage of methods that lacks support for these Windows
+security features.
+
+As a developper, you must use ``certbot.compat.os`` in every place you would need
+``os`` (eg. ``from certbot.compat import os`` instead of ``import os``). Otherwise the
```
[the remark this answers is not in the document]
adferrand: Clarified.

bmw:
One more comment here. Especially if we're going to try and automate the list of changed packages in the changelog (see https://github.com/certbot/certbot/issues/6911), I think we should add a brief changelog entry describing what was done so we don't have changed packages with unspecified changes.

In the "changed" section of the changelog, I'd simply say something like:

> Updated Certbot and its plugins to improve the handling of file system permissions on Windows as a step towards adding proper Windows support to Certbot.

And since #6911 isn't finished yet, I think we should list all packages changed in this PR in the changelog.

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
...
+code quality process will fail when your PR is submitted.
```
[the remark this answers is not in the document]
adferrand: Indeed, it is sufficient to know that "Your PR will be failed"

docs/contributing.rst
```diff
@@ -312,6 +312,25 @@ Please:
...
+
+All forbidden methods in ``certbot.compat.os`` have a Windows-safe equivalent
+in ``certbot.compat.security``.
```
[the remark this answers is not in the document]
adferrand: I hesitated also to put this here now. I remove it.

linter_plugin.py
adferrand: No indeed. It is hook. Updated the documentation accordingly.

linter_plugin.py
adferrand: Indeed! Fixed.

linter_plugin.py
adferrand: You are right. I remove it.

linter_plugin.py
adferrand: Ah, you are right! I forgot this situation. Of course, I totally agree, and implemented the additional control toward `os.path`-like imports.

I adapted also `certbot.compat.os` to automatically register a `certbot.compat.os.path`, to be used in a similar way than `os.path`.

Add a comment in the linter to have a similar example to study, as you said.

adferrand:
I updated the changelog.

docs/contributing.rst
```diff
@@ -312,6 +312,23 @@ Please:
 .. _PEP 8 - Style Guide for Python Code:
   https://www.python.org/dev/peps/pep-0008
 
+Use ``certbot.compat.os`` instead of ``os``
+===========================================
+
+
+Python's standard library ``os`` lacks of a full support for several Windows
+security features about file permissions (eg. DACLs). However several files
+handled by Certbot (eg. private keys) need a strongly restricted access
```
bmw:
```suggestion
handled by Certbot (eg. private keys) need strongly restricted access
```

docs/contributing.rst
```diff
@@ -312,6 +312,23 @@ Please:
...
+on both Linux and Windows.
+
+To help with this, the ``certbot.compat.os`` module wraps the standard
+``os`` module, and forbid usage of methods that lack support for these Windows
```
bmw:
```suggestion
``os`` module, and forbids usage of methods that lack support for these Windows
```

CHANGELOG.md
```diff
@@ -20,7 +21,24 @@ Despite us having broken lockstep, we are continuing to release new versions of
...
+* certbot-dns-google
+* certbot-dns-linode
+* certbot-dns-luadns
+* certbot-dns-nsone
+* certbot-dns-ovh
+* certbot-dns-rfc2136
+* certbot-dns-route53
+* certbot-dns-sakuracloud
+* certbot-nginx
+* certbot-postfix
```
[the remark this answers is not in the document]

certbot-apache/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
 acme[dev]==0.25.0
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-postfix/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
 acme[dev]==0.25.0
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/compat/os.py
```diff
@@ -0,0 +1,31 @@
...
+    # Check if the attribute does not already exist in our module. It could be internal attributes
+    # of the module (__name__, __doc__), or attributes from standard os already imported with
+    # `from os import *`.
+    if not hasattr(ourselves, attribute):
+        setattr(ourselves, attribute, getattr(std_os, attribute))
+
+# Clean all remaining importables that are not from the core os module.
+del ourselves, std_os, std_sys
+
+# Similarly to os.path, allow certbot.compat.os.path to behave as a module
```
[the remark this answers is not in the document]
adferrand: Done.

Issue #6897, opened 2019-03-30
Has certbot-dns-rfc2136 changed and nobody's updated the docs ?


udf2457:
I was happily using Certbot + certbot-dns-rfc2136 on openSUSE LEAP 15.

Until today when I updated it:
```
rpm -qa --last | fgrep certbot
python3-certbot-0.30.2-lp150.2.1.noarch       Sat 30 Mar 2019 10:12:42 GMT
python2-certbot-dns-rfc2136-0.30.2-lp150.2.1.noarch Sat 30 Mar 2019 10:12:38 GMT
python2-certbot-0.30.2-lp150.2.1.noarch       Sat 30 Mar 2019 10:12:34 GMT
```

Now when I run:
 `certbot certonly --deploy-hook /etc/foobar/certbot/certbot_deploy_hook.sh --server https://acme-v02.api.letsencrypt.org/directory --dns-rfc2136 --no-eff-email --dns-rfc2136-credentials /etc/foobar/certbot/certbot.ini -d '*.example.com'`  

I get the following error:

`certbot: error: unrecognized arguments: --dns-rfc2136-credentials /etc/foobar/certbot/certbot.ini`

The docs (https://certbot-dns-rfc2136.readthedocs.io/en/stable/) still seem to suggest `--dns-rfc2136-credentials` is still a perfectly acceptable flag ?

schoen:
Hi @udf2457,

There haven't been any changes to our upstream version of this plugin since last April (almost a year ago).

@bmw, do you know what could cause this problem? I think it's a Python version skew problem, because probably Certbot is now running with py3 by default and unable to find the py2 plugin module?

udf2457:
@schoen @bmw 

I think @schoen is onto something. I changed my script to call `/usr/bin/certbot-2.7` instead and the error no longer occurs.   So I suspect `dns-rfc2136` needs to be brought into the 21st century (a.k.a. made py3 friendly).

bmw:
Installing the `python3-certbot-dns-rfc2136` package with `zypper` should solve your problem.

isiko:
Hey, any Updates here? I'm currently getting the same Error

udf2457:
@isiko Sorry, can't help, I gave up on certbot a long time ago. I thoroughly recommend using LEGO instead.  Written in Go, no silly Python dependencies, just a single binary deployment.  Works beautifully. https://go-acme.github.io/lego/

isiko:
[...]

PR #6898, opened 2019-03-30
Changed the text of -h to add details regarding unregister 

### LINKED ISSUE #6879: Unregister command not listed in help
The command `certbot --help` only lists the `register` and `update_account`
commands.

```
manage your account with Let's Encrypt:
    register        Create a Let's Encrypt ACME account
    update_account  Update a Let's Encrypt ACME account
  --agree-tos       Agree to the ACME server's Subscriber Agreement
   -m EMAIL         Email address for important account notifications
```

The help command should list the `unregister` under manage account.

Additionally it is not completely clear from the `--help` output that a more complete help is available with `--help all`, which does list the `unregister`.

aditj:
Added unregister command details to --help command.
See #6879 .
[Fixes #6879.]

### DIFF
--- a/certbot/cli.py
+++ b/certbot/cli.py
@@ -101,6 +101,7 @@ manage certificates:
 
 manage your account with Let's Encrypt:
     register        Create a Let's Encrypt ACME account
+    unregister      Deactivate a Let's Encrypt ACME account
     update_account  Update a Let's Encrypt ACME account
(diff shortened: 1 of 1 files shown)


schoen:
I'm a little bit uneasy about how much detail is going in here, but I think there's a separate project to reassess that so I don't know of a specific policy reason not to be more complete here for the time being.

@aditj, thanks for taking this on!

Fixes #6897.

PR #6008, opened 2018-05-15
Update Pylint to 1.8.4

jepayne1138:
Upgrade pylint and resolve or disables resulting warnings

I made some subjective decisions on where to just disable new linter errors instead of updating the code where I thought that changes would result in worse code clarity.  Let me know if you have a different opinion on how to resolve any of the necessary changes and I'll see what I can do.

Resolves #5904

PR #6905, opened 2019-04-03
Update Pylint to 1.9.4

adferrand:
This PR updates Pylint to latest 1.x version available, on top of the work done in #6008. To do so:
* I reviewed and validated changes from #6008
* then, original PR is updated to master (almost one year ago)
* merge conflicts are fixed
* pylint (and relevant dev requirements) is updated to 1.9.4
* new linting errors are fixed (either by disabling or fixing code, with the limit for fix to avoid making the code harder to read)

The original PR is the work of two contributors beside me. In order to preserve the ownership, this PR should be merged, not squashed. In order to keep a clean history, after PR has been validated, and before merging, I will squash all my commits into one.


certbot/main.py
```diff
@@ -809,16 +809,11 @@ def install(config, plugins):
...
         le_client = _init_le_client(config, authenticator=None, installer=installer)
         _install_cert(config, le_client, domains)
+        return None
     else:
         raise errors.ConfigurationError("Path to certificate or key was not defined. "
             "If your certificate is managed by Certbot, please use --cert-name "
             "to define which certificate you would like to install.")
-    return None
-
-    if enhancements.are_requested(config):
```
bmw: Why are we deleting this block?

.pylintrc
```diff
@@ -189,6 +189,9 @@ init-import=no
 # not used).
 dummy-variables-rgx=(unused)?_.*|dummy
 
+# A regular expression matching ignored argument variable names
+ignored-argument-names=(unused)?_.*|dummy
```
bmw: What happened here? This PR adds ~50 `disable=unused-argument` lines. Can we not continue to have parameters named `unused_` automatically understood as known unused arguments?

acme/acme/client_test.py
```diff
@@ -138,7 +138,7 @@ def test_init_acme_version(self):
     def test_query_registration_client_v2(self):
         self.response.json.return_value = DIRECTORY_V2.to_json()
         client = self._init()
-        self.response.json.return_value = self.regr.body.to_json()
+        self.response.json.return_value = self.regr.body.to_json()  # pylint: disable=no-member
```
bmw: I don't like having to disable checks like this. It's not a huge deal because there aren't many instances of this in this PR, but `pylint` is wrong and making noise because it can't properly understand our code. `pylint` is giving devs false information, causing tests to fail, and making us spend time reassuring it because it's not able to properly detect what's going on here.

Are there any better options? Is there any way to do something like add comments to objects in `messages.py` explaining to `pylint` the type/attributes of these objects?

certbot/main.py
adferrand: Dead code.

.pylintrc
adferrand: Seems that the regex in pylintrc is taken into account for `unused-variable` only, not `unused-argument` here.

However, using `_argument` instead of `unused_argument` effectively avoid pylint to complain about it. So I can also do that instead of adding `disable=unused-argument`.

As a general note, but it is not a part of this PR: I do not like this kind of decorator usage for mock here, as it ends up with weird unused injected arguments. A `with` context manager approach would be way better here.

certbot/main.py
bmw: I think it's only dead code because of the addition of `return None` at https://github.com/certbot/certbot/pull/6905/files#diff-0b4a662e2c7b44bd47156f3d976ed698R812.

I think this code should be added back and `return None` added to the end of this function after this `if` block.

.pylintrc
bmw: That's disappointing. Previously we were able to name variables `unused_` to make pylint happy which is a convention that has been heavily used in the code.

What I'd personally like to see in order of preference is:

1. Verify that there's no way to modify the `pylint` config to keep the current behavior we have where `unused_` variables are treated the same as `_` prefixed variables. If there's a way to configure `pylint` to do this, I think we should do it and remove all added `pylint: disable=` lines about this.
2. If 1 isn't possible, change all parameters whose name is `unused_` and the function isn't part of a public API (there probably isn't any overlap here but you never know...) to use a `_` prefix instead. If some of our `.pylintrc` for setting up `unused_` is no longer necessary, let's remove it.

acme/acme/client_test.py
adferrand: Alternatively, we can set the option `ignored-classes` in `.pylintrc`

.pylintrc
adferrand: Have you specific reasons to not consider the `with` approach?

bmw: I think making the changes you described is fine in the cases the unused parameter is a mock object.

acme/acme/client_test.py
bmw: What other checks on these objects does that disable?

certbot/main.py
adferrand: Indeed. I added again the code block, with premature return

.pylintrc
adferrand: Finally, it appears that the relevant option was here: `ignored-argument-names`. It was just at the wrong place for recent versions of pylint. Now the arguments of type `unused_...` are effectively ignored by pylint.

I removed most of the `#pylint: disable=unused-argument` directives.

acme/acme/client_test.py
adferrand: Documentation of pylint states:


```
ignored-classes:
--
  | List of class names for which member attributes should not be checked (useful for classes with dynamically set attributes). This supports the use of qualified names. Default: optparse.Values,thread._local,_thread._local
```

https://docs.pylint.org/en/1.6.0/features.html#id5

So it seems to be really crafted for the situation of objects with dynamic parameters, and does not seems to removed any other check.

adferrand: However, regarding the low number of inline disables that are done to handle this issue, I think it is not necessary to remove every parameter check on these classes. It is still relevant to raise an error and be sure that the parameter effectively exists, before adding the inline comment to disable the check.

bmw: Setting `ignored-classes` seems perfect to me.

I personally think it's worth looking through the `disable=no-member` lines (or [other messages from that checker](https://docs.pylint.org/en/1.6.0/features.html#id6) if it also affects that) and if `pylint` is wrong, adding it to the list of `ignored-classes`. I think setting it for the `Field` class in `josepy` and `contextlib.closing` gets most of them.

While this allows us to remove a few more disables from this PR, I'm mainly interested in doing this for future development. There's certainly an argument to be made for having `pylint` complain to you about things it's not sure about so the developer can double check, but I personally don't like this approach. I think it's another hurdle for new devs to try and understand and wastes all developers time. I'd rather have `pylint` check the things it can well and we'll rely on things like unit and integration tests to help and sure that these dynamic properties exist when `pylint` can't see them.

adferrand: @bmw I configured `ignore-classes` to remove JSON-related classes from the check, and remove most of the `no-member` suppress.

acme/acme/standalone_test.py
```diff
@@ -266,9 +266,9 @@ def setUp(self):
                     os.path.join(localhost_dir, 'key.pem'))
 
         with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
-            sock.bind(('', 0))
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-            self.port = sock.getsockname()[1]
+            sock.bind(('', 0))  # pylint: disable=no-member
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)  # pylint: disable=no-member
+            self.port = sock.getsockname()[1]  # pylint: disable=no-member
```
bmw: Can we add `closing` to the list of ignored classes and remove these lines?

acme/acme/client.py
```diff
@@ -1171,7 +1163,7 @@ def _add_nonce(self, response):
         if self.REPLAY_NONCE_HEADER in response.headers:
             nonce = response.headers[self.REPLAY_NONCE_HEADER]
             try:
-                decoded_nonce = jws.Header._fields['nonce'].decode(nonce)
+                decoded_nonce = jws.Header._fields['nonce'].decode(nonce)  # pylint: disable=no-member
```
bmw: I think this `pylint: disable` can be deleted.

adferrand: Done.

acme/acme/standalone_test.py
adferrand: Done. And for that matter, I removed also all other `no-member`, that are useless now.

bmw: Thanks for doing that!

adferrand:
OK, I do that.

bmw:
This was merged in #6932.

PR #6913, opened 2019-04-04
Fix certbot-auto on some RPM distributions

### LINKED ISSUE #6912: OS detection - integer expression expected
CloudLinux 7.6

 wget https://dl.eff.org/certbot-auto
 chmod a+x ./certbot-auto

 /bin/certbot-auto renew
/bin/certbot-auto: line 754: [: 7.6: integer expression expected
/bin/certbot-auto: line 754: [: 7.6: integer expression expected

There is some problem with OS detection.

adferrand:
Fixes #6912

Bash evaluate all condition in a predicate statement, eg. `"$SOMEVAR" = "test" -a "$ANOTHERVAR" = "test2"`, even if it is not necessary, for instance if the first condition is false in the example here.

As a consequence, on non-Fedora distributions, an evaluation of the distribution version could be done on non numeric value, eg. `"6.7" -eq "29"`, making certbot-auto failing in this case.

This PR fixes that, by evaluating the version on RPM distributions only if we are on Fedora. Otherwise, version will be "0".


bmw:
In the likely event we do a point release to fix this, can you update the changelog describing the fix here?

bmw:
I did this in #6914.

PR #6915, opened 2019-04-04
Prepare for the 0.33.1 release.

bmw:
The changelog should still say `<version> - master` because it will be fixed up automatically by the release script at https://github.com/certbot/certbot/blob/master/tools/_release.sh#L69.


bmw:
More codecov problems...

Issue #5412, opened 2018-01-11
Speed up or eliminate set_by_cli


jsha:
In https://github.com/certbot/certbot/pull/5372#issuecomment-356781859, I noticed that set_by_cli can be very slow. It takes about one second per flag to evaluate. With a lot of flags, that really adds up.

bmw:
The way this should work is the first time `set_by_cli` is run in a given instance of Certbot or a test, it is very slow. This is because it is creating a new argument parser class, having all the different pieces of Certbot add the arguments they take to it (including Certbot's plugins), and reparsing all command line arguments. After that initial run though, subsequent runs should be very fast because these reparsed arguments are saved in something like a static variable of the function [here](https://github.com/certbot/certbot/blob/0a4f926b160cf118c81974e91df7f196bc7a379d/certbot/cli.py#L205). Did you see different behavior?

jsha:
Yeah, I saw each run taking about a second.

bmw:
How were you running it?

bmw:
I've seen Certbot print output about set_by_cli with a long pauses in between, but I'm not sure this is necessarily due to set_by_cli. It certainly could be, but this function often isn't called in a tight loop so I think it's entirely possible it's other parts of Certbot that are too slow or Certbot just isn't being very clear about what it's doing.

Erica's planning on cleaning up the UI soon and while we could leave this open to profile this area of Certbot, it doesn't seem like the highest priority to me personally. If we're going to try and improve performance, I think it makes more sense to look at the big picture which is being tracked by issues like https://github.com/certbot/certbot/issues/6027.

For that reason, I'm closing this issue, but if you disagree, feel free to reopen it explaining why!

jsha:
Hi! set_by_cli actually is called in a tight loop in storage.py when writing out the renewal config:

```
        for option, value in six.iteritems(all_values)
        if _relevant(option) and cli.option_was_set(option, value))
```

(option_was_set calls set_by_cli)

I still regularly see this add several seconds to an issuance, and think it's worth fixing.

bmw:
That's fair! I agree speeding up this section of Certbot is valuable. I put this in wishlist because in the grand scheme of Certbot priorities, it doesn't seem as critical as some of the other problems we have, but with that said, this nerdsniped me a bit while I was waiting on tests for the release to run and I think I figured this out.

PR incoming...

bmw:
[...]

Issue #6027, opened 2018-05-21
Make Certbot issue certs faster


bmw:
I haven't done any measurements myself, but I've been told it takes Certbot 10 - 15 seconds to obtain a certificate and was asked if we can do this more in the <5 second range. This would be particularly useful in large deployments using Certbot.

I suspect this won't be very hard to do because we've spent almost no time trying to optimize Certbot.

The one exception to this might be the Apache and Nginx plugins where we sleep for 3 seconds after setting up the DV challenge. The reason for this is the command you use to ask both webservers to reload their config simply sends a signal to the webserver and immediately returns. This command returning does not mean the server has finished loading their config though and I'm unaware of a reliable way for us to know when this process is complete. As a result, we currently just sleep for a few seconds and afterwards assume the config has been reloaded.

Erutan409:
Just reading this information taught me a lot about the process and is very good to know from a devops standpoint.

margana:
One thing I have found is that issuing a new cert without specifying `--cert-name` gets noticeably slower when you have thousands of certificates already.

Also `renew` takes about a minute per thousand certificates while none of them are due to expire, this is with certs stored on an nfs share.

bmw:
I think we should still do this.

PR #6917, opened 2019-04-05
Don't search for plugins once for each config item

bmw:
I came across #5412 during my triage and after jsha reopened the issue, I couldn't stop myself from taking a quick look. Upon looking at code he linked to, I noticed that Certbot is searching for plugins once for every value in Certbot's configuration. This is completely unnecessary and very slow so I moved searching for plugins outside of the loop.

On a small VPS, I timed Certbot's performance obtaining a new cert with standalone before and after this change and was pretty happy with the results.

Before:
```
real 0m26.625s
user 0m22.228s
sys 0m0.144s
```

After:
```
real 0m7.458s
user 0m1.620s
sys 0m0.156s
```

### DIFF
--- a/certbot/storage.py
+++ b/certbot/storage.py
@@ -238,16 +238,17 @@ def _write_live_readme_to(readme_path, is_base_dir=False):
                                     "certificates.\n".format(prefix=prefix))
 
 
-def _relevant(option):
+def _relevant(namespaces, option):
     """
     Is this option one that could be restored for future renewal purposes?
+
+    :param namespaces: plugin namespaces for configuration options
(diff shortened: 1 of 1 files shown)


bmw:
This is progress on #6027 too but we're still not quite there.

adferrand:
Holly hell of a doped rabbit! I did not see the huge improvement on the CI: integration tests are dropping from 20 min to 8 min! Cannot wait what it will give using the parallelization of Certbot-CI!

jsha:
Thanks for this! Very exciting to see the test speedup.

PR #6923, opened 2019-04-05
Don't run pip tools/venv.py

bmw:
It won't work. Instead, follow the instructions at the top of this document to set up a virtual environment and activate it.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -353,7 +353,7 @@ Steps:
 
 1. Write your code!
 2. Make sure your environment is set up properly and that you're in your
-   virtualenv. You can do this by folliwng the instructions in the
+   virtualenv. You can do this by following the instructions in the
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -353,8 +353,8 @@ Steps:
 
 1. Write your code!
 2. Make sure your environment is set up properly and that you're in your
-   virtualenv. You can do this by running ``pip tools/venv.py``.
-   (this is a **very important** step)
+   virtualenv. You can do this by folliwng the instructions in the
```
adferrand: Typo here.

adferrand:
Codecov again is crazy.

docs/contributing.rst
bmw: Thanks for catching this.

Issue #6927, opened 2019-04-08
Lexicon update causes urls to be malformed when adding TXT records


jstaehle:
## My operating system is (include version):
Linux (amazon aws fargate v 1.3.0 container)

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
I am depending on the `certbot/dns-dnsimple` docker image in my Dockerfile.

## I ran this command and it produced this output:
When attempting to generate certificates for qa environments in my startup script (full package here: https://github.com/askwonder/wonqa) I run `docker run -i --name wonqa-certbot wonqa-certbot:latest certonly --dns-dnsimple --dns-dnsimple-credentials dnsimple.ini --agree-tos --no-eff-email -m eng@askwonder.com -d fix-qa.wondev.net,api.fix-qa.wondev.net,start.fix-qa.wondev.net,research.fix-qa.wondev.net,storybook.fix-qa.wondev.net,pusher.fix-qa.wondev.net`.
I get the following output: 

```
Obtaining a new certificate

Performing the following challenges:

dns-01 challenge for api.fix-qa.wondev.net
dns-01 challenge for fix-qa.wondev.net
dns-01 challenge for pusher.fix-qa.wondev.net
dns-01 challenge for research.fix-qa.wondev.net

dns-01 challenge for start.fix-qa.wondev.net
dns-01 challenge for storybook.fix-qa.wondev.net
Unsafe permissions on credentials configuration file: dnsimple.ini

Cleaning up challenges

Error adding TXT record: 400 Client Error: Bad Request for url: https://api.dnsimple.com/v281695/zones/wondev.net/records
```

## Certbot's behavior differed from what I expected because:
I expect it to add the TXT records successfully. Instead the url is malformed and there is no slash in between `v2` and `81695` which is my account id.

## Investigation
After doing some digging I have determined that the issue was caused by a change made by lexicon (https://github.com/AnalogJ/lexicon), likely this commit: https://github.com/AnalogJ/lexicon/commit/f46a072fc119d069bc276f7fd1977ab548fcd364#diff-89f3c260164ceef017a121144c790f42. 
[...]

PR #6928, opened 2019-04-08
dnsimple 3.2.1

jstaehle:
Hello
In order to pick up a fix to a lexicon issue (resolved via https://github.com/AnalogJ/lexicon/pull/389) which fixes a url construction issue that is affecting us over at https://github.com/askwonder/wonqa, certbot can update its lexicon dependency to the latest version.

I understand if there are additional considerations in upgrading this dependency, I have tested that it resolves my issue but obviously there may be other parts of certbot-dns-dnsimple that are affected as well.

See corresponding issue here: https://github.com/certbot/certbot/issues/6927
------
having been changed.


bmw:
Before merging this, we want to see what packagers at various distros are doing in response to this problem. I suspect many of them will just backport the fix which means we may want to do more here to ensure we maintain compatibility with the versions of Lexicon's APIs available in older distros where Certbot may still be getting updates.

Assuming we bump this version number eventually though, we'll also have to fix up `tools/oldest_constraints.txt` or `certbot-dns-dnsimple/local-oldest-requirements.txt`. I wouldn't worry about this piece for now though.

bmw:
Closing in favor of #7008.

Issue #6937, opened 2019-04-09
Make tools/venv.py prefer Python 3


bmw:
In the next year, Certbot should stop supporting Python 2 but if someone gets to this issue before then...

There were some concerns raised about `tools/venv.py` continuing to use Python 2 despite Python 2 approaching its deprecation date. To try and restate these concerns myself, they were:

* Python 3 is what Certbot will use in the future while Python 2 is going away so we should be running on Python 3 wherever possible to increase the chance of finding bugs.
* Most devs will already have Python 3 installed (however, devs on Debian based systems may not have the `python3-venv` package installed which is needed for our `tools/venv3.py` script)
* Tools such as `pip` are starting to print deprecation warnings on Python 2.7 in newer versions. Seeing warnings like this should not be a regular part of the development process in Certbot.
* People may notice that we're continuing to use Python 2 and not switch to Python 3 themselves which hurts the Python ecosystem.

Despite these concerns, I was hesitant to make this change because we tell devs to run `certbot-auto --os-packages-only` which automates installing Python 2 and not Python 3 (on most systems). This isn't entirely trivial as different systems use different names for the Python 3 package and I personally prefer this extra automation for devs over solving the problems above.

One idea I had to try and address all of these concerns is to rename `tools/venv.py` to `tools/venv2.py` and make `tools/venv.py` an automagical script that will use either Python 2 or Python 3 but prefers Python 3 and only falls back to Python 2 if Python 3 is not available. `tools/venv.py` should also check that the Python 3 `venv` module is available before agreeing to use Python 3.

Another benefit of a script like this is we can probably use it to solve test failures in https://github.com/certbot/certbot/blob/master/tests/letstest/scripts/test_sdists.sh which I believe are failing due to not trying to use Python 3.

adferrand:
I quite disagree with the idea to do more magic in `venv.py`, or its associates, including detecting if python3-venv is installed in a Debian system, or toggle installation of Python 2 or Python 3 automatically based on crafted rules.

[...]

PR #6938, opened 2019-04-09
Add --disable-pip-version-check to pip calls.

bmw:
To address some of the concerns in #6937, this makes `pip` stop printing warnings about a newer version of `pip` being available. We know we're specifically pinning an older version of `pip`. It is unnecessary to tell every developer about it on every run.

### DIFF
--- a/tools/pip_install.py
+++ b/tools/pip_install.py
@@ -74,6 +74,12 @@ def call_with_print(command, cwd=None):
     subprocess.check_call(command, shell=True, cwd=cwd or os.getcwd())
 
 
+def pip_install_with_print(args_str):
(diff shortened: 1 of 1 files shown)


bmw:
I haven't dug into the `pip` code to understand why, but I don't believe pip's warning here is ever printed from `pipstrap.py`. Have you seen it?

adferrand:
You are right. It is because in `pipstrap`, `pip` is invoked through a `check_output` capturing in the process all its stdout. And in fact, adding the flag could be harmful, because `pipstrap` is designed to upgrade very old `pip` version, where the flag could be unsupported.

PR #6946, opened 2019-04-12
[Unix] Create a framework for certbot integration tests: PART 3a

adferrand:
Following #6821, this PR continues to convert certbot integration tests into certbot-ci.

This PR add tests covering on L185-222 in `tests/certbot-boulder-integration.sh`.

### DIFF
--- a/certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
+++ b/certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
@@ -128,23 +128,6 @@ def test_manual_dns_auth(context):
     assert_save_renew_hook(context.config_dir, certname)
 
 
-def test_invalid_domain_with_dns_challenge(context):
-    """Test certificate issuance failure with DNS-01 challenge."""
-    # Manual dns auth hooks from misc are designed to fail if the domain contains 'fail-*'.
-    certs = ','.join([context.get_domain('dns1'), context.get_domain('fail-dns1')])
-    context.certbot([
-        '-a', 'manual', '-d', certs,
-        '--allow-subset-of-names',
-        '--preferred-challenges', 'dns',
-        '--manual-auth-hook', context.manual_dns_auth_hook,
-        '--manual-cleanup-hook', context.manual_dns_cleanup_hook
-    ])
-
-    output = context.certbot(['certificates'])
-
-    assert context.get_domain('fail-dns1') not in output
-
-
 def test_renew_files_permissions(context):
     """Test proper certificate file permissions upon renewal"""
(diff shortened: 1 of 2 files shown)


certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -23,6 +24,72 @@ def context(request):
...
+        context.certbot([
+            '--domains', certname, '--preferred-challenges', 'http-01', 'run',
+            '--cert-name', certname,
+            '--pre-hook', 'echo wtf.pre >> "{0}"'.format(context.hook_probe),
+            '--post-hook', 'echo wtf.post >> "{0}"'.format(context.hook_probe),
+            '--deploy-hook', 'echo deploy >> "{0}"'.format(context.hook_probe)
+        ])
+
+    assert_hook_execution(context.hook_probe, 'deploy')
+    assert_save_renew_hook(context.config_dir, certname)
```
bmw: I suspect we just want to change the code in `assert_save_renew_hook`, but this isn't testing the same thing as [CheckSavedRenewHook](https://github.com/certbot/certbot/blob/d9880721b31ab4e9e9d2fdc38d83ff2bd9078378/tests/certbot-boulder-integration.sh#L142). The shell code is trying to check that there is an entry for `renew_hook` saved into the renewal configuration file.

adferrand: Indeed, I did not interpretate this assertion correctly. This is fixed.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -120,12 +121,12 @@ def test_manual_dns_auth(context):
         '--manual-cleanup-hook', context.manual_dns_cleanup_hook,
         '--pre-hook', 'echo wtf.pre >> "{0}"'.format(context.hook_probe),
         '--post-hook', 'echo wtf.post >> "{0}"'.format(context.hook_probe),
-        '--renew-hook', 'echo renew >> "{0}"'.format(context.hook_probe)
+        '--renew-hook', 'echo deploy >> "{0}"'.format(context.hook_probe)
```
bmw: Can we leave this at "renew"?

If for some reason we want to change this to "deploy", I think we should also change the `assert_hook_execution` call down below to "deploy" as well.

adferrand: Original tests were using the `deploy` keyword. But it is just a naming convention, without functional impact. I change the word, it makes more sense

PR #6951, opened 2019-04-12
[Unix] Create a framework for certbot integration tests: PART 3e

adferrand:
Following #6821, this PR continues to convert certbot integration tests into certbot-ci.

This PR add tests covering checks on L430-447 in `tests/certbot-boulder-integration.sh`. Previous lines are covered with existing tests, or by #6946, #6947, #6948, #6949.

PR #6949, opened 2019-04-12
[Unix] Create a framework for certbot integration tests: PART 3d

adferrand:
Following #6821, this PR continues to convert certbot integration tests into certbot-ci.

This PR add tests covering checks on L397-429 in `tests/certbot-boulder-integration.sh`. Previous lines are covered with existing tests, or by #6946, #6947 and #6948.

PR #6948, opened 2019-04-12
[Unix] Create a framework for certbot integration tests: PART 3c

adferrand:
Following #6821, this PR continues to convert certbot integration tests into certbot-ci.

This PR add tests covering about renew, on L283-396 in `tests/certbot-boulder-integration.sh` (by including existing `test_renew_files_permissions` and `test_renew_with_hook_scripts`). Previous lines are covered with existing tests, or by #6946 and #6947.

PR #6953, opened 2019-04-12
[Unix] Create a framework for certbot integration tests: PART 3g

adferrand:
Following #6821, this PR continues to convert certbot integration tests into certbot-ci.

This PR add tests covering checks on L531 to the end on `tests/certbot-boulder-integration.sh`. Previous lines are covered with existing tests, or by #6946, #6947, #6948, #6949, #6951, #6952.


certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -209,3 +209,35 @@ def get_certbot_version():
     # Typical response is: output = 'certbot 0.31.0.dev0'
     version_str = output.split(' ')[1].strip()
     return LooseVersion(version_str)
+
+
+def find_certbot_root_directory():
```
bmw: Instead of doing this, I'd rather make `certbot-ci` self contained.

To do this, I propose making something like a "testdata" directory within the `certbot-ci` package. After doing this, we can copy over `sample-config` in this directory and add it to `MANIFEST.in` for the package. We can then use `pkg_resources.resource_filename` to help us find the package. See our other packages for examples of all of this. Once all certbot-ci PRs have landed, we'll delete the `sample-config` found in `tests/integration`. 

I'm open to changing any of the details there, but I personally think keeping `certbot-ci` a little more self contained has value.

certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -209,3 +209,35 @@ def get_certbot_version():
...
+        raise ValueError('Error, could not find certbot sources root directory')
+
+    return current_dir
+
+
+def load_sample_data_path(workspace):
+    """
+    Load the certbot configuration example designed to make OCSP tests, and return its path
+    :param str workspace: current test workspace directory path
+    :return str: the path to the loaded sample data directory
```
bmw:
```suggestion
    :returns: the path to the loaded sample data directory
    :rtype: str
```

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -98,3 +98,33 @@ def test_renew_with_hook_scripts(context):
 
     assert_cert_count_for_lineage(context.config_dir, certname, 2)
     assert_hook_execution(context.hook_probe, 'deploy')
+
+
+def test_ocsp_status(context):
+    """Test retrieval of OCSP statuses."""
+    if context.acme_server == 'pebble':
+        pytest.skip('Pebble does not support OCSP status requests.')
```
bmw: Can we either:

1. Move this skip after checking the stale OCSP status.
2. Split this into two functions and only skip the code under "OCSP 2".

certbot-ci/certbot_integration_tests/utils/misc.py
adferrand: In totally agree. I did not remember this solution at the time I wrote the test. I will do it now.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
adferrand: Solution 2. Otherwise the whole test would be marked as skipped even if one part as been executed.

adferrand: Implemented.

certbot-ci/certbot_integration_tests/utils/misc.py
adferrand: Implemented. The config folder is put it `certbot_integration_tests/assets`, and marked in the manifest of certbot-ci to be included in the distribution.

certbot-ci/certbot_integration_tests/utils/misc.py
adferrand: Done.

adferrand:
I added the minimal coverage control that I was planning to do. It is located in the integration env in tox config. This limit can be increased when all tests are added.

PR #6956, opened 2019-04-12
Update Fedora AMI

### LINKED ISSUE #6831: Update the Fedora AMI used in test farm tests
To update to Fedora 29, we probably need to wait until 0.33.0 is released because  test_leauto_upgrades.sh tests upgrading from an old version of certbot-auto to the current version but the current version of certbot-auto is the only version that supports this OS.

To upgrade to Fedora 28, we may want to resolve #6828 and #6830 first so ideally all tests will pass on the image.

bmw:
Fixes #6831.

I ran all test farm tests we run as part of the release process and they passed. The AMI ID was taken from https://web.archive.org/web/20180626045008/https://alt.fedoraproject.org/cloud/.

### DIFF
--- a/tests/letstest/scripts/test_leauto_upgrades.sh
+++ b/tests/letstest/scripts/test_leauto_upgrades.sh
@@ -15,13 +15,8 @@ if ! command -v git ; then
         exit 1
     fi
 fi
-# 0.17.0 is the oldest version of letsencrypt-auto that has precompiled
-# cryptography and the tagged commit is in master. 0.16.0 was the first version
-# to use precompiled cryptography, but the release PR was squashed losing the
-# commit. We want to use a precompiled version of cryptography for stability.
-# Previous versions that have to compile against OpenSSL on installation
-# started failing on newer distros with newer versions of OpenSSL.
-INITIAL_VERSION="0.17.0"
+# 0.18.0 is the oldest version of letsencrypt-auto that works on Fedora 26+.
+INITIAL_VERSION="0.18.0"
 git checkout -f "v$INITIAL_VERSION" letsencrypt-auto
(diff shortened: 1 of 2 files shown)


tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -15,13 +15,8 @@ if ! command -v git ; then
...
     fi
 fi
-# 0.17.0 is the oldest version of letsencrypt-auto that has precompiled
-# cryptography and the tagged commit is in master. 0.16.0 was the first version
-# to use precompiled cryptography, but the release PR was squashed losing the
-# commit. We want to use a precompiled version of cryptography for stability.
-# Previous versions that have to compile against OpenSSL on installation
-# started failing on newer distros with newer versions of OpenSSL.
-INITIAL_VERSION="0.17.0"
+# 0.18.0 is the oldest version of letsencrypt-auto that works on Fedora 26+.
```
adferrand: Fedora 26 is EOL, and so 27 in fact (end of last year). Why we do not set the minimum version here in regards to the last non EOL, Fedora 28?

bmw: 0.18.0 is also needed for Fedora 28 which is the Fedora version being used in this PR.

I said Fedora 26+ because that marks the version where Fedora changed package names requiring a change to certbot-auto which is documented at https://github.com/certbot/certbot/blob/master/CHANGELOG.md#fixed-26.

adferrand: OK !

adferrand:
@ohemorange I did not see that @bmw requested a review from you before reviewing and merging. I did it since I understand better letstest with already several corrections and a design doc on my own.

But if you want to check by yourself since you can launch the farm tests, do not hesitate to double check ^^

PR #6957, opened 2019-04-12
Remove slash from path

bmw:
I tried running certbot-ci in master and tests failed. The reason for this can be seen by:
```
>>> from os.path import join
>>> join('config_dir', 'archive', 'certname', '/privkey2.pem')
'/privkey2.pem'
```


adferrand (approved):
Yes, I added this correction in part3a, but we can add this right away. LGTM!

bmw:
> Yes, I added this correction in part3a, but we can add this right away. LGTM!

Ah, I did not know that. I was just looking through and playing with what we had so far before I started reviewing new stuff.

Thanks for the review though!

PR #6965, opened 2019-04-17
[Windows] Security model for files permissions - STEP 3b

adferrand:
This PR implements some required modifications in `certbot.compat.misc` from #6497. In particular, all operations on certbot will require administrative access, very similarly to the root permissions scope for Linux.


certbot/compat/misc.py
```diff
@@ -2,42 +2,31 @@
 This compat module handles various platform specific calls that do not fall into one
 particular category.
 """
-import ctypes
-import errno
+from __future__ import absolute_import
+
+import os  # pylint: disable=os-module-forbidden
```
bmw: Can we continue to use `certbot.compat.os` here? If not, why not?

certbot/compat/misc.py
```diff
@@ -2,42 +2,31 @@
...
 import select
 import stat
 import sys
 
-from certbot import errors
-from certbot.compat import os
+try:
+    from win32com.shell import shell as shellwin32  # pylint: disable=import-error
+except ImportError:  # pragma: no cover
+    shellwin32 = None  # type: ignore
```
bmw: This is my usual (probably overly paranoid) nit about not wanting to disable `mypy` when possible.

certbot/compat/misc.py
```diff
@@ -2,42 +2,31 @@
...
 
-UNPRIVILEGED_SUBCOMMANDS_ALLOWED = [
-    'certificates', 'enhance', 'revoke', 'delete',
-    'register', 'unregister', 'config_changes', 'plugins']
```
bmw: This is probably fine, but what is the motivation of removing this list?

certbot/compat/os.py
```diff
@@ -29,3 +29,14 @@
...
+
+
+# Because of the blocking strategy on file handlers on Windows, rename to not behave as expected
+# with POSIX systems: an exception will be raised if dst already exists. Hopefully there is
+# os.replace on Windows for Python 3, that will do the same than on POSIX. Hopefully also, only
+# Python 3 is supported for Certbot. So we can rely on os.rename on Linux, and os.replace
+# on Windows.
+def rename(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
+    """Method os.rename() is forbidden"""
+    raise RuntimeError('Usage of os.rename() is forbidden. '  # pragma: no cover
```
bmw: Same comment as I put at https://github.com/certbot/certbot/pull/6964#discussion_r283070863.

certbot/compat/misc.py
adferrand: This because on top of the symlinks that required already administrative privileges, several operations in `filesystem` will require them also, making effectively almost all operations in certbot unusable with them.

That, added to the fact that Certbot usually requires root on Linux, as stated by the documentation, that communicating with IIS is also an administrative operation, and that Windows implements metadata to programmatically require administrative privileges to run marked executable, make this distinction between certbot commands useless.

bmw: That makes sense. I think some commands like `plugins` could still work without administrative privileges, but we can keep things simple and require it across the board until there's a good reason for us to do something differently.

certbot/tests/compat/filesystem_test.py
```diff
@@ -1,21 +1,21 @@
-"""Tests for certbot.compat."""
+"""Tests for certbot.compat.misc"""
```
bmw:
```suggestion
"""Tests for certbot.compat.filesystem"""
```

certbot/compat/misc.py
adferrand: It is temporary, until a future PR in the part3 plan of windows file permissions makes the complete modifications for `misc`.

certbot/compat/os.py
adferrand: Done. As a side not, if part3a is merged first, I will update from this PR the test that trigger exceptions from `os` module. Otherwise, I will update it from part3a.

certbot/compat/misc.py
adferrand: Fixed as usual with a global boolean.

certbot/compat/misc.py
bmw: Then can we use `certbot.compat.os` for now in this PR and change it later when it's necessary?

certbot/compat/filesystem.py
```diff
@@ -0,0 +1,29 @@
+"""Compat module to handle files security on Windows and Linux"""
+from __future__ import absolute_import
+
+import errno
+import os  # pylint: disable=os-module-forbidden
+
+
+def rename(src, dst):
```
bmw: Actually one more comment. I've been thinking about our conversation in 3c and think that this function is more accurately called `replace` as it has the same behavior as that function on Python 3 and `os.rename` on Windows has different behavior.

(Bonus points for changing the implementation to match that by calling `replace` first if it exists and otherwise `rename`, but I'm less worried about that because the implementation isn't being added in this PR.)

certbot/compat/misc.py

certbot/compat/filesystem.py
adferrand: Done. With bonus point.

certbot/compat/os.py
```diff
@@ -29,3 +29,18 @@
 
 # Clean all remaining importables that are not from the core os module.
 del ourselves, std_os, std_sys
+
+
+# Because of the blocking strategy on file handlers on Windows, rename do not behave as expected
```
bmw:
```suggestion
# Because of the blocking strategy on file handlers on Windows, rename does not behave as expected
```

certbot/compat/os.py
```diff
@@ -29,3 +29,18 @@
...
+# with POSIX systems: an exception will be raised if dst already exists.
+def rename(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
```
bmw: nit: Given the purpose of this module, I'd personally just move a single `pylint: disable=function-redefined` above these function definitions at the top level of the file.

certbot/compat/os.py
```diff
@@ -29,3 +29,18 @@
...
+    """Method os.rename() is forbidden"""
+    raise RuntimeError('Usage of os.rename() is forbidden. '
+                       'Use certbot.compat.filesystem.replace() instead.')
+
+
+# To be consistent across Windows and Linux, filesystem.replace should be used.
```
bmw: I personally think we should remove this method.

If `os.replace` exists, I think it's fine for people to call it as it has the same behavior on both OSes.

If `os.replace` does not exist because we're on Python 2, would we rather have the behavior that `os.replace` does not exist or have it raise this `RuntimeError`? I personally lean toward the former because it acts more like the standard `os` module and `pylint` will warn about there not being a `replace` attribute while it won't warn about this function raising an exception.

certbot/tests/compat/filesystem_test.py
```diff
@@ -0,0 +1,21 @@
...
+import certbot.tests.util as test_util
+from certbot.compat import os
+from certbot.compat import filesystem
+
+
+class OsReplaceTest(test_util.TempDirTestCase):
+    """Test to ensure consistent behavior of rename method"""
+
+    def test_os_replace_to_existing_file(self):
+        """Ensure that rename will effectively rename src into dst for all platforms."""
```
bmw:
```suggestion
        """Ensure that replace will effectively rename src into dst for all platforms."""
```

certbot/tests/reverter_test.py
```diff
@@ -347,7 +347,7 @@ def test_finalize_checkpoint_cannot_title(self, mock_move):
         self.assertRaises(
             errors.ReverterError, self.reverter.finalize_checkpoint, "Title")
 
-    @mock.patch("certbot.reverter.misc.os_rename")
+    @mock.patch("certbot.reverter.filesystem.replace")
     def test_finalize_checkpoint_no_rename_directory(self, mock_rename):
```
bmw: nit: I think we should change this variable name to `mock_replace`.

certbot/compat/os.py
adferrand: I would prefer to keep it. It makes this consistent with the purpose of `certbot.compat.os`, that is to raise exceptions on function that cannot be used. Here `os.replace` cannot be used in Certbot, since it will break on Python 2.

Considering your second paragraph, I would evaluate it towards two opinions you expressed in the past:
* first that we are not committed to follow the `os` Python core module API anymore
* second that pylint process cannot be fully trusted, and abandon for instance the idea of linting specifically each forbidden function, in favor of a global lint of `os` import

I think that raising the same exception than in the other functions will give it a lot of visibility. We should not have any `os.replace` in Certbot codebase, like `os.rename`, and so I think it is more coherent.

bmw: Like everything else, I think there's pros and cons. My approach means that a standard exception will be raised when running the code on Python 2 and our linter has a chance to catch it. Your approach means that the linter will not catch it, but it will always raise the same custom error. I personally think it's debatable which approach is better.

I'm fine keeping this, but if we do, can you update the comment here? I personally don't think it accurately describes the purpose of raising an exception here. `os.replace` is consistent on Windows and Linux.

adferrand: I tried to balance again the two approaches, but when I consider the development workflow, I still really think it makes things clearer.

Let's consider things like this: you cannot use os.rename, because it does not work on windows, you cannot use os.replace because it does not work on python 2. You need to use filesystem.replace.

In the case of os.rename without forbidding the method by an exception, you would get the error from the CI when code is tested I. Appveyor. For os.replace without forbidding, it would be because lint is not happy, or when executing tests in the CI on Python 2.

So two very different errors should deliver the same message: do not use them, use filesystem instead.

Here I think we simplify the development workflow by giving the same behavior (a runtime exception) with an explicit message about what should be used.

If you are still ok with this, I will update the comments.

certbot/compat/os.py
adferrand: Done.

certbot/tests/reverter_test.py
adferrand: Done.

certbot/compat/os.py
adferrand: Updated the comment nonetheless.

adferrand:
@bmw, all conversations are answered in this PR.

bmw:
Closing and reopening to see if codecov works now that their [incidents](https://status.codecov.io/#past-incidents) have been resolved.

bmw:
Codecov is still broken, but at least coverage was successfully uploaded this time which looks fine.

PR #6967, opened 2019-04-17
[Windows] Security model for files permissions - STEP 3c

adferrand:
This PR implements the `security.open` method from #6497.


bmw:
Did you look into what it would take to create the file with secure permissions up front rather than creating the file with default permissions and running `chmod`?

We might be able to convince ourselves that what we have here is OK, however, I think this approach has security problems in some contexts (for example, see http://tldp.org/HOWTO/Secure-Programs-HOWTO/avoid-race.html) and an approach to set the permissions at file creation allows us to avoid this speculation altogether and keeps the behavior closer to what we have on UNIX.

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
 
 # Clean all remaining importables that are not from the core os module.
 del ourselves, std_os, std_sys
+
+
+# The os.open function on Windows will have the same effect than a bare os.chown towards the given
```
ohemorange: What does this mean? What effect, and on what? Which given mode? What does a "bare" os.chown mean? Where are those flaws described?

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
...
+# mode, and will create a file with the same flaws that what have been described for os.chown.
+# So upon file creation, security.take_ownership will be called to ensure current user is the owner
```
ohemorange: Does that mean, "after creating a file, you'll have to call take_ownership" or "hen you call the function to create a file, that function calls take_ownership"?

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
...
+# of the file, and security.chmod will do the same thing than for the modified os.chown.
```
ohemorange: security.chmod will do what thing?

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
...
+# Internally, take_ownership will update the existing metdata of the file, to set the current
```
ohemorange:
```suggestion
# Internally, take_ownership will update the existing metadata of the file, to set the current
```

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
...
+# username (resolved thanks to win32api module) as the owner of the file.
```
ohemorange:
```suggestion
# username (determined by the win32api module) as the owner of the file.
```

certbot/compat/os.py
ohemorange: should this be "then", as in "after"?

certbot/compat/os.py
ohemorange: Internally to what?

Should all of this documentation be here? This function is just raising an error. Maybe it should say what to call instead, and have that documentation there?

certbot/lock.py
```diff
@@ -208,22 +209,22 @@ class _WindowsLockMechanism(_BaseLockMechanism):
     A Windows lock file mechanism.
     By default on Windows, acquiring a file handler gives exclusive access to the process
     and results in an effective lock. However, it is possible to explicitly acquire the
-    file handler in shared access in terms of read and write, and this is done by os.open
+    file handler in shared access in terms of read and write, and this is done by security.open
```
ohemorange:
```suggestion
    file handler in shared access in terms of read and write, and this is done by filesystem.open
```

certbot/lock.py
```diff
@@ -208,22 +209,22 @@ class _WindowsLockMechanism(_BaseLockMechanism):
...
     and io.open in Python. So an explicit lock needs to be done through the call of
     msvcrt.locking, that will lock the first byte of the file. In theory, it is also
     possible to access a file in shared delete access, allowing other processes to delete an
     opened file. But this needs also to be done explicitly by all processes using the Windows
     low level APIs, and Python does not do it. As of Python 3.7 and below, Python developers
     state that deleting a file opened by a process from another process is not possible with
-    os.open and io.open.
+    security.open and io.open.
```
ohemorange:
```suggestion
    filesystem.open and io.open.
```

certbot/util.py
```diff
@@ -206,24 +207,20 @@ def check_permissions(filepath, mode, uid=0):
...
 
 
-def safe_open(path, mode="w", chmod=None, buffering=None):
+def safe_open(path, mode="w", chmod=None):
     """Safely open a file.
 
     :param str path: Path to a file.
     :param str mode: Same os `mode` for `open`.
-    :param int chmod: Same as `mode` for `os.open`, uses Python defaults
+    :param int chmod: Same as `mode` for `security.open`, uses Python defaults
```
ohemorange:
```suggestion
    :param int chmod: Same as `mode` for `filesystem.open`, uses Python defaults
```

certbot/compat/filesystem.py
```diff
@@ -0,0 +1,23 @@
...
+    is correctly applied.
+    :param str file_path: The file path to open
+    :param int flags: Flags to apply on file while opened
+    :param int mode: POSIX mode to apply on file when opened,
+        Python defaults will be applied if ``None``
+    :returns: the file descriptor to the opened file
+    :rtype: int
+    """
+    file_descriptor = os.open(file_path, flags, mode)
+    # TODO: Change to security.chmod once all logic of windows files permissions has been merged
```
ohemorange:
```suggestion
    # TODO: Change to filesystem.chmod once all logic of windows files permissions has been merged
```

certbot/compat/os.py
ohemorange: Also presumably all the mentions of `security` in this explanation should be `filesystem`.

certbot/compat/os.py
```diff
@@ -29,3 +29,15 @@
...
+def open(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
+    """Method os.open() is forbidden"""
+    raise RuntimeError('Usage of os.open() is forbidden. '  # pragma: no cover
```
ohemorange: Let's add a trivial test that this raises a runtime error and remove this comment

certbot/util.py
```diff
@@ -206,24 +207,20 @@ def check_permissions(filepath, mode, uid=0):
...
         if ``None``.
-    :param int buffering: Same as `bufsize` for `os.fdopen`, uses Python
-        defaults if ``None``.
 
     """
     open_args = ()  # type: Union[Tuple[()], Tuple[int]]
     if chmod is not None:
         open_args = (chmod,)
     fdopen_args = ()  # type: Union[Tuple[()], Tuple[int]]
-    if buffering is not None:
```
ohemorange: Why don't we need this anymore? What was this doing?

adferrand: It was doing nothing, in fact. This option is never used in certbot.

certbot/compat/os.py
adferrand: Here we are on a side effect of splitting the logic in several PRs. This comment is written using a lot of reference to `os.chmod`, that is implemented in #6964.

So it should be read as it will be once the logic is merged: first the comment on `os.chown` gives the context about limitation of file modes in Windows, then this comment and the function `os.open` is written just after, reusing this context for the reader.

certbot/compat/os.py
adferrand: Second option. I forgot that "will" can be seen as an intention in the future in english.

I make the comment more clear.

certbot/compat/os.py
adferrand: Indeed. I simplified the comment here.

certbot/compat/os.py
adferrand: See https://github.com/certbot/certbot/pull/6965#discussion_r287553843

certbot/compat/os.py
adferrand: Not relevant anymore, since I rewrote the comment.

adferrand:
@bmw, regarding your comment about potential race conditions, I think it is not possible to mitigate this. I tried several approaches to do it from the open function.

Without rewriting our own open function from the Windows kernel API, any attempt lead to break certbot. The idea would be to have directly the correct DACL right when open is invoked. 

As it is not possible due to the implementation of open, the idea would be to let Windows use the correct DACL, thanks to the NTFS filesystem that inherits the DACL on a new file from the directory that contains it.

It is the way tempfile.mkstemp() do things for instance. And I tried also a similar strategy, by creating the file first in a secure location, to get a restrictive DACL, then apply chmod, then move the file to the intended location.

But this does not play nicely with the flags passed to open, typically os.O_CREAT and os.O_EXCL. Or I get a DACL that locks me out of the file.

So, now, I consider the risk of race conditions limited: it would happen on a new file, between the open call and the chmod call. But between the two calls, the file is still empty, as the file descriptors has not been returned yet to the caller. So there will be no sensible data that would be, even during a short period of time, exposed to everybody.

Regarding the complexity of implementing a mitigation VS the effective non-criticality of the race condition, I think we can live with it.

bmw:
> So, now, I consider the risk of race conditions limited: it would happen on a new file, between the open call and the chmod call. But between the two calls, the file is still empty, as the file descriptors has not been returned yet to the caller. So there will be no sensible data that would be, even during a short period of time, exposed to everybody.
>
> Regarding the complexity of implementing a mitigation VS the effective non-criticality of the race condition, I think we can live with it.

I don't think it's a major security problem, but I do think it potentially a problem. For another warning about this, see this text from http://timgolden.me.uk/python/win32_how_do_i/add-security-to-a-file.html

> Since it's far easier to create a file using Python's usual open ("blah", "w") than to use Windows' CreateFile, we'll create the file first with whatever default permissions it inherits and then replace them with our own. Obviously if this file needed to be truly secure then this would be a security hole and you'd do better to create it secure.

Again, we may be able to carefully analyze this and come to the conclusion that this is never a problem in any context, however, I have the same opinion as I had before and I'd personally rather err on the side of being overly cautious in this code, make sure we get this right, and not have to worry about these kind of problems down the road.

Did you look into the `CreateFile` and `CreateDirectory` functions at http://timgolden.me.uk/pywin32-docs/win32file.html? Given the code in part3a of creating the DACL we want, it seems like these functions wouldn't be too hard to use would it? Keep in mind that we do not have to commit ourselves to the `os` APIs and we can change the functions in `certbot.compat.filesystem` however we want to make our lives easier.

adferrand:
Keep in mind that we do not have to commit ourselves to the os APIs and we can change the functions in certbot.compat.filesystem however we want to make our lives easier.

I disagree with this statement. The design I follow in the windows file permissions, and in general in the windows port of Certbot, is to be as less intrusive as possible to the existing logic written for Posix. My reasoning on it is that even if adding the full support of Windows is a great improvement, calling Certbot from Linux will stay the major case.

On the specific topic of Windows file permissions then, I think it is already sufficiently intrusive, as we forbid `os` and do not allow direct alternatives for some of its functions. So breaking the API of the remaining ones is not desirable in my opinion, and even if it can help us now, it could be in the long term, by removing some possibilities of coding.

adferrand:
That said, I think that I can keep a not too complex reimplementation of `os.open` for Windows based on its current call on Certbot. Then we can improve it in the future as new needs appear in the future.

But I will stay with the existing API, and so direct and straightforward delegate to `os.open` when we are on Linux, according to what I said in the previous comment.

So, I need the DACL created by part3a, so I block here on #6964.

bmw:
In my opinion, keeping the API the same lost its long term value when we moved the functions out of `certbot.compat.os`. We're no longer trying to silently change the behavior of these functions such that the caller can use the module without knowing the changes exist. If any function with inadequate behavior on Windows is called in `certbot.compat.os`, an exception will be raised pointing them to a function in another module instead. At this point, it's clear the function referenced in the exception works differently than the standard `os` function. I think we should embrace this fact and change the APIs in whatever way better suits how the function is used and works within Certbot rather than trying to after the fact pretend it's still the normal `os` function.

While a bit extreme, I think one could argue that pretending to be the standard library `os` functions at this point is harmful. It's both forcing developers to learn that some standard library `os` functions cannot be used because they are inadequate as well as trying to hide the details of the changes we've made behind the standard `os` interfaces.

adferrand:
My point here is not to make `filesystem` pretend that it behaves like `os`. I think that the very name of the module is sufficient to express the difference. My point was about keeping if possible the general design of python standard function, because I think there is some value in that design.

I think our discussion here is a little too much high level for the specific purpose of this PR. We can continue this conversation on another channel if you agree.

For the context of this PR, my question here is which API to expose in the `filesystem.open` method. My intention was to continue to use flags (so `def open(path, flags, mode)`) instead of a specific API design (like `def open(path, mode, create=False, exclusive=False, readWrite=False)`).

The first design allows to pass all flags as they are in Linux, and keep the full capacities of `os.open`. Then, the implementation on Windows is limited at the beginning (by handling only some flags combinations), and will see its behavior been elaborated in the future as new needs appear.

The second design limits behavior on both platform, so losing existing capacities on Linux, and will see its overall behavior been elaborated in the future as new needs appear.

What is your opinion on these designs @bmw?

bmw:
I'm fine moving the high level conversation elsewhere, however, if we're struggling to come to an agreement on how this function should work, I think we may want to have that conversation first before continuing to get into the details here.

My potential concerns with the first approach are:

1. If someone calls `filesystem.open` on Windows with flags that haven't been implemented yet, I wouldn't want the function to silently ignore them.
2. Flags to [POSIX open](https://pubs.opengroup.org/onlinepubs/009695399/functions/open.html) and [CreateFile](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea) do not work the same way and `CreateFile` only has 5 options.

Due to these concerns, I wonder if it might be better for us to change the function a bit to only serve the limited subset of shared functionality. While there are some combinations that will be nontrivial to implement like `os.CREAT | os.TRUNC` on Windows, I think there's really only a few options we care about. I don't see us setting flags like `O_DSYNC` anytime soon.

adferrand:
Looking at the code, the following flags need to be handled on various associations: `os.O_WRONLY`, `os.O_CREAT`, `os.O_TRUNC`, `os.O_EXCL`, `os.O_RDWR`. As you said, `os.CREAT | os.TRUNC` is already tricky.

I am really worried that even if we build a subset of `open`, this will lead to a significantly complex code. I am also concerned by the fact that we are trying to reimplement `os.open` from the Windows API, will the existing `os.open` does already a pretty good job. What it does not is to set an appropriate ACL list.

If we go back to the primary requirement here, the idea is that: add the proper ACL list upon `os.open`. And you want to do this without race condition risks, so with a ACL list on the file that is never too relaxed.

I propose to focus on that requirement.

If the file already exists, things are trivial: just call `chmod` before opening the file and return the descriptor.

Things are more complex if the file does not exists. In this case, indeed I think like you said, that the proper way is to use `CreateFile` from `win32file` module. But I want to use it here solely for the requirement I said: apply the target ACL list, and ensure that the ACL list is never too relaxed. In fact, two situations of set of flags need to be handled in the creation context:
* `os.O_CREAT`: create the file if not exists
* `os.O_CREAT | os.O_EXCL`: create the file if not exists and raise an exception if already exists

This two situations can be handled by the following `creationPolicy` in `CreateFile`: `OPEN_ALWAYS` and `CREATE_NEW` respectively.

Once the creation is handled, the file exists and so `chmod` can be applied as before.

Finally now `os.open` can be called, handling all the other situations, as the ACL list is now properly set. When I say situation is handled, I say in particular that in the context of `os.O_CREAT | os.O_EXCL`, an exception has been already raised if the file already exists, before even `os.open` is called.

Finally, the code I have in mind is the following: 
```
def open(file_path, flags, mode=0o777):
    if POSIX_MODE:
        # Handle creation of the file atomically with proper permissions
        if flags & os.O_CREAT:

            # If os.O_EXCL is set, we will use the "CREATE_NEW", that will raise an exception if file exist,
            # matching the API contract of this bit flag. Otherwise, we use "OPEN_ALWAYS" that will create
            # the file if not exist.
            creation_disposition = win32con.CREATE_NEW if flags & os.O_EXCL else win32con.OPEN_ALWAYS

            attributes = win32security.SECURITY_ATTRIBUTES()
            security = attributes.SECURITY_DESCRIPTOR
            user = _get_current_user()
            dacl = _generate_dacl(user, mode)
            security.SetSecurityDescriptorDacl(1, dacl, 0)

            handle = win32file.CreateFile(file_path, win32file.GENERIC_READ, 0, attributes, creation_disposition, 0, None)
            handle.Close()

        # At this point, the file has been created with proper permissions if os.O_CREAT is set,
        # or an exception has been raised if file already exists and os.O_CREAT | os.O_EXCL is set.
        # We can safely invoke chmod for the general case.
        if os.path.exists(file_path):
            chmod(file_path, mode)

    # On Linux, invoke directly os.open. On Windows, file creation with proper permissions in
    # respect with os.O_CREAT and os.O_EXCL has already been handled.
    return os.open(file_path, flags, mode)
```

adferrand:
My proposition still implies to block on #6964, and also on a future PR to handle ownership on the files (alternatively it can be integrated in this PR).

bmw:
I didn't take the time to carefully review that code right now, but at a high level, I like it! Good idea @adferrand.

adferrand:
@bmw this PR is ready for a new round of review, despite the CI failure on Codecov.

See https://github.com/certbot/certbot/pull/7173 for why Codecov is mad, again.

adferrand:
Flags are all green for PR review now.

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+            handle.Close()
+
+            # At this point, the file that was not existing has been created, with proper
+            # permissions, so os.O_CREAT and os.O_EXCL are not needed anymore. We remove them
+            # from the flags to avoid a FileExists exception when os.open will be called.
+            flags = flags ^ os.O_CREAT ^ os.O_EXCL
+        elif os.path.exists(file_path):
+            chmod(file_path, mode)
+
+    # On Linux, invoke directly os.open. On Windows, file creation with proper
```
ohemorange:
```suggestion
    # On Linux, invoke os.open directly. On Windows, file creation has proper
```

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+    # permissions in respect with os.O_CREAT so we can call os.open.
```
ohemorange:
```suggestion
    # permissions with respect to os.O_CREAT so we can call os.open.
```

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+
+            handle = win32file.CreateFile(file_path, win32file.GENERIC_READ, 0,
+                                          attributes, win32con.CREATE_NEW, 0, None)
+            handle.Close()
+
+            # At this point, the file that was not existing has been created, with proper
+            # permissions, so os.O_CREAT and os.O_EXCL are not needed anymore. We remove them
+            # from the flags to avoid a FileExists exception when os.open will be called.
+            flags = flags ^ os.O_CREAT ^ os.O_EXCL
+        elif os.path.exists(file_path):
```
ohemorange: where did all the `os.O_EXCL` handling go? if the file already exists but `os.O_EXCL` was set, won't we chmod then open the file anyway?

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+    if not POSIX_MODE:
+        # Handle creation of the file atomically with proper permissions
+        if not os.path.exists(file_path) and flags & os.O_CREAT:
+            attributes = win32security.SECURITY_ATTRIBUTES()
+            security = attributes.SECURITY_DESCRIPTOR
+            user = _get_current_user()
+            dacl = _generate_dacl(user, mode)
+            security.SetSecurityDescriptorDacl(1, dacl, 0)
+
+            handle = win32file.CreateFile(file_path, win32file.GENERIC_READ, 0,
```
ohemorange: there's still a race condition here, no? the file could have been created between the `if not os.path.exists` call and here. So we don't know for sure at this line that the file doesn't exist. And we're not checking `os.O_EXCL` here, so how do we know for sure that we should error out if the file already exists? we're just letting the program crash if there's a race condition, even if they've only set `os.O_CREAT`?

adferrand: Indeed you are right. See my general comment below.

certbot/compat/filesystem.py
adferrand: Indeed, see my general comment below.

certbot/compat/os.py
```diff
@@ -45,12 +45,22 @@
...
 # group, appropriate permissions for the "Everyone" group, and all permissions to the
 # "Administrators" group + "System" user, as they can do everything anyway.
-def chmod(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
+def chmod(*unused_args, **unused_kwargs):
     """Method os.chmod() is forbidden"""
     raise RuntimeError('Usage of os.chmod() is forbidden. '
                        'Use certbot.compat.filesystem.chmod() instead.')
 
 
+# The os.open function on Windows has the same effect than a call to os.chown concerning the file
```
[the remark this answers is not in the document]
adferrand: I update the comment.

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+                                          attributes, win32con.CREATE_NEW, 0, None)
+            handle.Close()
+
+            # At this point, the file that was not existing has been created, with proper
```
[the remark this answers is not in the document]
adferrand: Applied.

certbot/compat/filesystem.py
```diff
@@ -36,6 +41,43 @@ def chmod(file_path, mode):
...
+            # permissions, so os.O_CREAT and os.O_EXCL are not needed anymore. We remove them
+            # from the flags to avoid a FileExists exception when os.open will be called.
```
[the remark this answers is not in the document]
adferrand: Not relevant anymore.

certbot/compat/filesystem.py
adferrand: Not relevant anymore.

certbot/compat/filesystem.py
adferrand: Applied.

adferrand:
So I proposed initially https://github.com/certbot/certbot/pull/6967#issuecomment-498592010 to implement open without race conditions. Then I tried to simplify it. However @ohemorange showed that this simplification was missing some situations (like `os.O_CREAT` without `os.O_EXCL` if the file already exists). It made me also realized that is `os.open` was failing for other reasons, the file will still be `chmod`-ed.

Then I went back to my initial implementation, that consists basically:
1) If no creation operation is involved, invoke `os.open`, then `chmod`, then return the descriptor
2) If creation is involved, properly map `os.O_CREAT` and `os.O_EXCL` to their counterpart in the Windows API using `CreateFile` to ensure permissions are atomically applied to the file upon its creation.

With this, we cover the following cases:
1) `os.O_CREAT | os.O_EXCL` and file does not exist:  handled as `CreateFile` is invoked with `win32con.CREATE_NEW`
2) `os.O_CREAT | os.O_EXCL` and file exist: handled as `CreateFile` is invoked with `win32con.CREATE_NEW`, that will raise an exception, re-thrown as `errno.EEXIST
3) `os.O_CREAT` and file does not exist: handled as `CreateFile` is invoked with `win32con.CREATE_ALWAYS`
4) `os.O_CREAT` and file exist: handled as `CreateFile` is invoked with `win32con.CREATE_ALWAYS` that will silently re-create the file, still with atomically permissions applied
5) `os.O_CREAT` and file exist and locked: handled as `CreateFile` is invoked, and will raise inconditionnally an error, re-thrown as `errno.EACCES`

certbot/compat/os.py
```diff
@@ -45,12 +45,23 @@
...
```
ohemorange:
```suggestion
# The os.open function on Windows has the same effect as a call to os.chown concerning the file
```

certbot/compat/os.py
```diff
@@ -45,12 +45,23 @@
...
+# modes: these modes lack of a correct control over the permissions given to the file. Instead,
```
ohemorange:
```suggestion
# modes: these modes lack the correct control over the permissions given to the file. Instead,
```

certbot/compat/os.py
```diff
@@ -45,12 +45,23 @@
...
+# filesystem.open invokes the Windows native API `CreateFile` to ensure that permissions are
+# atomically set in case of file creation, or invoke filesystem.chmod to set properly the
```
ohemorange:
```suggestion
# atomically set in case of file creation, or invokes filesystem.chmod to properly set the
```

certbot/compat/filesystem.py
```diff
@@ -173,3 +235,14 @@ def _compare_dacls(dacl1, dacl2):
...
     return ([dacl1.GetAce(index) for index in range(0, dacl1.GetAceCount())] ==
             [dacl2.GetAce(index) for index in range(0, dacl2.GetAceCount())])
+
+
+def _get_current_user():
+    """
+    Return the pySID corresponding to the current user.
+    """
+    account_name = win32api.GetUserNameEx(win32api.NameSamCompatible)
+    # Passing None to systemName instruct the lookup to start from the local system,
```
ohemorange:
```suggestion
    # Passing None to systemName instructs the lookup to start from the local system,
```

certbot/compat/filesystem.py
```diff
@@ -173,3 +235,14 @@ def _compare_dacls(dacl1, dacl2):
...
+    # then continue the lookup to associated domain.
```
ohemorange: what does this mean? continues the lookup...where? how? what is looking up what domain? I suspect this is a perfectly reasonable choice and it's just the language here.

certbot/compat/filesystem.py
```diff
@@ -36,6 +44,60 @@ def chmod(file_path, mode):
...
+
+        try:
+            handle = win32file.CreateFile(file_path, win32file.GENERIC_READ,
+                                          win32file.FILE_SHARE_READ & win32file.FILE_SHARE_WRITE,
+                                          attributes, disposition, 0, None)
+            handle.Close()
+        except pywintypes.error as err:
+            # Handle native windows errors into python errors to be consistent with the API
+            # of os.open in the situation of a file already existing or locked.
+            if err.winerror == winerror.ERROR_FILE_EXISTS:
```
ohemorange: will we hit this error on windows when the file is a symbolic link as well? "If O_EXCL and O_CREAT are set, and path names a symbolic link, open() shall fail and set errno to [EEXIST], regardless of the contents of the symbolic link."

certbot/compat/filesystem.py
```diff
@@ -36,6 +44,60 @@ def chmod(file_path, mode):
...
+        disposition = win32con.CREATE_NEW if flags & os.O_EXCL else win32con.CREATE_ALWAYS
+
+        attributes = win32security.SECURITY_ATTRIBUTES()
+        security = attributes.SECURITY_DESCRIPTOR
+        user = _get_current_user()
+        dacl = _generate_dacl(user, mode)
+        security.SetSecurityDescriptorDacl(1, dacl, 0)
+
+        try:
+            handle = win32file.CreateFile(file_path, win32file.GENERIC_READ,
```
ohemorange: just to check -- does this behave the same way as it would on linux if the file exists, but is a directory?

certbot/compat/filesystem.py
```diff
@@ -36,6 +44,60 @@ def chmod(file_path, mode):
...
+        # If os.O_EXCL is set, we will use the "CREATE_NEW", that will raise an exception if
+        # file exists, matching the API contract of this bit flag. Otherwise, we use
+        # "CREATE_ALWAYS" that will always create the file whether it exists or not.
+        disposition = win32con.CREATE_NEW if flags & os.O_EXCL else win32con.CREATE_ALWAYS
+
+        attributes = win32security.SECURITY_ATTRIBUTES()
+        security = attributes.SECURITY_DESCRIPTOR
+        user = _get_current_user()
+        dacl = _generate_dacl(user, mode)
+        security.SetSecurityDescriptorDacl(1, dacl, 0)
```
ohemorange: what do these numbers mean?

certbot/compat/filesystem.py
adferrand: Indeed, my comment was not clear. I improved it.

certbot/compat/filesystem.py
adferrand: Windows does make a distinction between files and symbolic links in the documentation. Also, symbolic links does not behave as in Linux: they are real files, and in fact, you have file-type symlinks and directory-type symlinks.

Finally, I can confirm from my tests that a pre-existing symlink will raise the same error than for files (`EEXIST`) if we invoke `filesystem.open` with `os.O_CREAT | os.O_EXCL`.

certbot/compat/filesystem.py
adferrand: No, in this case you have a `PermissionError` while in Linux it is a `FileExists` (`EEXIST`), since on Windows, files and directories are two different objects.

However the behavior of `filesystem.open` is the same than standard `os.open` on Windows in this situation.

certbot/compat/filesystem.py
adferrand: We set first parameter to 1 (`True`) to say that this security descriptor contains a DACL. Otherwise second and third parameters are ignored.

We set third parameter to 0 (`False`) to say that this security descriptor is **not** constructed from a default mechanism, but explicitly set by the user.

See https://docs.microsoft.com/en-us/windows/desktop/api/securitybaseapi/nf-securitybaseapi-setsecuritydescriptordacl

I added this in the comments.

certbot/compat/filesystem.py
ohemorange: as in, we'll hit the PermissionError here that we otherwise would have hit later, in the call to `os.open`?

adferrand: Could you be more precise about what you have in mind here, in order to answer more accurately?

Indeed the PermissionError occurs in CreateFile call instead of os.open, and so ensures to have the same behavior and the same result than the original os.open method.

ohemorange: Ok, so that does mean we should be catching PermissionError, and raising it as EEXIST instead? Is there a list of errors CreateFile might raise?

adferrand: You want to make `filesystem.open` not behave like `os.open` on Windows? What is the reason?

adferrand: Also about your question for the list of errors. No, there is not. We are on Windows here, not some sort of open source software with people willing to write exhaustive and comprehensive documentation ^^

ohemorange: Isn't the whole point of this to get consistent behavior out of `filesystem.open` across platforms?

adferrand: Really good question!

In fact it is not exactly this goal. The point is to make Certbot behaves correctly, and this imply most of the time to get consistent behavior across platforms. But not always, for instance the requirement to run Certbot as an administrator, while you can run Certbot as a normal user on Linux. 

It means that I do make converge behaviors across platforms not by principle, but by necessity.

I prefer to avoid converting this error into a `EEXIST` for the following reasons:
* Certbot works correctly without it
* Python developers did not convert this windows exception into a python one, which constitute a hint here in my opinion
* and in fact, the error is very generic: it is a `PermissionError`, not something like "a directory already exists at this path"

This makes me worry to accidentally shadow an unrelated exception into an EEXIST one, and create unexpected side effects in the future. Also as you said, it is not documented. It seems that no Windows developer thought it would be a good idea to document, and maybe handle specifically this case on a so much critical function as `CreatFile` is.

Instead you just get a "You are not authorized to do the action you are trying to do"...

ohemorange: Ah, I see now. The situation I was imagining is that somewhere else in certbot, we wrap the function call that was previously `os.open` and is now `filesystem.open` in a `try`, and catch `EEXIST` only, instead of setting flags in the method call. In which case, I was worried that we'd crash there when really we should have caught it in the hypothetical `except`.

But since `PermissionError` is more generic, we probably do want to be letting it raise, since it could mean other things as well.

I still feel a little weird about leaving it to stumble across in the future... at minimum, documenting with a `:raise:`would be helpful, open to other suggestions as well.

adferrand: Yes, this behavior is really unfortunate on Windows. But the segregation between files and directories is at the core design of the Windows filesystem.

To enlight this a little more, I tried to invoke `win32file.CreateFile` using the `win32con.CREATE_ALWAYS` flag on a situation where a directory of the same name than the file that we try to create. To recall, `win32con.CREATE_ALWAYS` instruct Windows to always re-create the file, even if it already exists. But if it is a directory, you end again with a FilePermission error.

ohemorange: and on Linux you get an `OSError`. Looks like `PermissionError` subclasses `OSError`. So the most specific thing we can promise is that we'll throw an `OSError`, which is fine imo. But we should note that in the docstring.

adferrand: I added a `:raise:` entry in the docstring.

certbot/compat/filesystem.py
```diff
@@ -55,6 +55,7 @@ def open(file_path, flags, mode=0o777):  # pylint: disable=redefined-builtin
         Python defaults will be applied if ``None``
     :returns: the file descriptor to the opened file
     :rtype: int
+    :raise: OSError(errno.EEXIST) if the file already exists and os.O_CREAT & os.O_EXCL are set.
```
ohemorange: "or on Windows, when the file exists, is a directory, and os.O_CREAT is set".

adferrand: But in this case this is not a EEXIST.

adferrand:
@ohemorange I added the tests you asked regarding create file while invoking `filesystem.open`

certbot/compat/filesystem.py
adferrand: Added in the docstring.

certbot/lock.py
```diff
@@ -208,26 +209,30 @@ class _WindowsLockMechanism(_BaseLockMechanism):
     A Windows lock file mechanism.
     By default on Windows, acquiring a file handler gives exclusive access to the process
     and results in an effective lock. However, it is possible to explicitly acquire the
-    file handler in shared access in terms of read and write, and this is done by os.open
+    file handler in shared access in terms of read and write, and this is done by filesystem.open
```
ohemorange:
```suggestion
    file handler in shared access in terms of read and write, and this is done by os.open
```

Looks like this was probably meant to be the original `os`

certbot/lock.py
```diff
@@ -208,26 +209,30 @@ class _WindowsLockMechanism(_BaseLockMechanism):
...
     and io.open in Python. So an explicit lock needs to be done through the call of
     msvcrt.locking, that will lock the first byte of the file. In theory, it is also
     possible to access a file in shared delete access, allowing other processes to delete an
     opened file. But this needs also to be done explicitly by all processes using the Windows
     low level APIs, and Python does not do it. As of Python 3.7 and below, Python developers
     state that deleting a file opened by a process from another process is not possible with
-    os.open and io.open.
+    filesystem.open and io.open.
```
ohemorange:
```suggestion
    os.open and io.open.
```
Python devs aren't making statements about filesystem.open

certbot/lock.py
adferrand: Indeed, this is pycharm that replace this in comments during a safe replace operation.

certbot/lock.py
adferrand: Same as above.

PR #6968, opened 2019-04-17
[Windows] Security model for files permissions - STEP 3d

adferrand:
This PR implements the `security.mkdir` and `security.makedirs` methods from #6497.


bmw:
Similar to my comment at https://github.com/certbot/certbot/pull/6967#issuecomment-495007461, did you look into what it would take to create the directory with secure permissions up front rather than creating the directory with default permissions and running `chmod`?

certbot/compat/filesystem.py
```diff
@@ -0,0 +1,48 @@
...
+    orig_mkdir_fn = os.mkdir
+    try:
+        def wrapper(one_path, one_mode=0o777):  # pylint: disable=missing-docstring
+            # Note, we need to provide the origin os.mkdir to our mkdir function,
+            # or we will have a nice infinite loop ...
+            mkdir(one_path, mode=one_mode, mkdir_fn=orig_mkdir_fn)
+
+        os.mkdir = wrapper
+
+        os.makedirs(file_path, mode)
```
ohemorange: I don't love that we're relying on the internal details of `os.makedirs`' implementation for this to work. The implementation has to only be a few lines of recursive code that we could easy reuse here, right?

certbot/compat/os.py
```diff
@@ -29,3 +29,25 @@
...
+
+
+# Very similarly to os.open, os.mkdir has the same effect on Windows, to create an unsecured
+# folder. Same mitigation is provided using security.take_ownership and security.chmod.
+# On top of that, we need to handle the fact that os.mkdir is called recursively by os.makedirs.
+# This is done by protecting the original os.mkdir to have the real logic, call it during the
+# recurrence and apply immediately the security model on every processed folder.
+def mkdir(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
+    """Method os.mkdir() is forbidden"""
+    raise RuntimeError('Usage of os.mkdir() is forbidden. '  # pragma: no cover
```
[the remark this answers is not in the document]
adferrand: See https://github.com/certbot/certbot/pull/6965#discussion_r287553843

certbot/compat/os.py
```diff
@@ -29,3 +29,25 @@
...
+
+
+# As said above, os.makedirs would call the original os.mkdir function recursively, creating the
+# same flaws for every actual folder created. This method is modified to ensure that our
+# modified os.mkdir is called, by monkey patching temporarily the mkdir method on the
+# original os module, executing the modified logic to protect corecrtly newly created folders,
+# then restoring original mkdir method in the os module.
+def makedirs(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
+    """Method os.makedirs() is forbidden"""
+    raise RuntimeError('Usage of os.makedirs() is forbidden. '  # pragma: no cover
```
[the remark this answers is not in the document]
adferrand: See https://github.com/certbot/certbot/pull/6965#discussion_r287553843

certbot/compat/filesystem.py
```diff
@@ -0,0 +1,48 @@
+"""Compat module to handle files security on Windows and Linux"""
+from __future__ import absolute_import
+
+import os  # pylint: disable=os-module-forbidden
+
+from acme.magic_typing import Callable  # pylint: disable=unused-import,no-name-in-module
+
+
+def makedirs(file_path, mode=0o777):  # pylint: disable=function-redefined
```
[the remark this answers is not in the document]
adferrand: Indeed, it came from the copy of code in `certbot.compat.os`. Corrected.

certbot/compat/filesystem.py
```diff
@@ -0,0 +1,48 @@
...
+    is correctly applied.
+    :param str file_path: The file path to open
+    :param int mode: POSIX mode to apply on file when opened,
+        Python defaults will be applied if ``None``
+    """
+    # As we know that os.mkdir is called internally by os.makedirs, we will swap the function in
+    # os module for the time of makedirs execution.
+    orig_mkdir_fn = os.mkdir
+    try:
+        def wrapper(one_path, one_mode=0o777):  # pylint: disable=missing-docstring
```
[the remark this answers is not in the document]
adferrand: Corrected.

adferrand:
@bmw you can see my response here about open https://github.com/certbot/certbot/pull/6967#issuecomment-495938856 that is still relevant for the case of mkdir.

certbot/compat/filesystem.py
adferrand: I agree that I could reimplement the makedirs logic here, but it is not the design I want to follow.

The idea here is to make filesystem module as close as possible to the original os module capabilities. I do that by following a design consisting in reusing at most the existing logic in os, and change only what needs to be changed to apply the security model. 

Here the documentation says that makedirs use mkdir internally, so it is a documented entrypoint that I think suitable to use. However, I do not want to make any other assumption on how makedirs work for the several python versions that we can encounter. I prefer to consider most of the logic of makedirs as a black box, and stay consistent to the actual behavior of it in a given python version, instead of fixing a global unique logic for all of them.

adferrand: I will add that the specific case that I want to cover with this design, is that if for any reason python maintainers wants to change some behavior in makedirs, I want to take advantage of it flawlessly.

bmw:
> @bmw you can see my response here about open #6967 (comment) that is still relevant for the case of mkdir.

Let's keep the conversation about this in #6967 so it's in one place unless we get to the point we're talking about the specifics of creating files vs. directories.

adferrand:
According to the discussion on #6967, I block also on #6964 here.

adferrand:
@bmw this PR is ready for a new round of review, despite the CI failure on Codecov.

See https://github.com/certbot/certbot/pull/7173 for why Codecov is mad, again.

adferrand:
@bmw this PR is ready for a new round of review, despite the CI failure on Codecov.

See https://github.com/certbot/certbot/pull/7173 for why Codecov is mad, again.

adferrand:
Flags are all green now.

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,67 @@ def chmod(file_path, mode):
...
+    """
+    Rewrite of original os.mkdir function, that will ensure on Windows that given mode
+    is correctly applied.
+    :param str file_path: The file path to open
+    :param int mode: POSIX mode to apply on directory when created, Python defaults
+                     will be applied if ``None``
+    """
+    if POSIX_MODE:
+        _os_mkdir(file_path, mode)
+    else:
```
ohemorange: nit: if we return instead of `else`ing, everything else can be one less indent...

certbot/compat/filesystem.py
```diff
@@ -173,3 +241,14 @@ def _compare_dacls(dacl1, dacl2):
...
     return ([dacl1.GetAce(index) for index in range(0, dacl1.GetAceCount())] ==
             [dacl2.GetAce(index) for index in range(0, dacl2.GetAceCount())])
+
+
+def _get_current_user():
+    """
+    Return the pySID corresponding to the current user.
+    """
+    account_name = win32api.GetUserNameEx(win32api.NameSamCompatible)
+    # Passing None to systemName instruct the lookup to start from the local system,
```
ohemorange: just a note that we'll want to keep the version in 3c rather than this version

certbot/compat/os.py
```diff
@@ -51,6 +51,28 @@ def chmod(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
...
+def mkdir(*unused_args, **unused_kwargs):
+    """Method os.mkdir() is forbidden"""
+    raise RuntimeError('Usage of os.mkdir() is forbidden. '
+                       'Use certbot.compat.filesystem.mkdir() instead.')
+
+
+# As said above, os.makedirs would call the original os.mkdir function recursively, creating the
+# same flaws for every actual folder created. This method is modified to ensure that our
+# modified os.mkdir is called, by monkey patching temporarily the mkdir method on the
+# original os module, executing the modified logic to protect corecrtly newly created folders,
```
ohemorange:
```suggestion
# original os module, executing the modified logic to correctly protect newly created folders,
```

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,67 @@ def chmod(file_path, mode):
...
+        dacl = _generate_dacl(user, mode)
+        security.SetSecurityDescriptorDacl(1, dacl, 0)
+
+        try:
+            win32file.CreateDirectory(file_path, attributes)
+        except pywintypes.error as err:
+            # Handle native windows error into python error to be consistent with the API
+            # of os.mkdir in the situation of a directory already existing.
+            if err.winerror == winerror.ERROR_ALREADY_EXISTS:
+                raise OSError(errno.EEXIST, err.strerror)
```
ohemorange: sure do wish we could drop 2.7 and raise a py3 `FileExistsError` rather than `OSError`... probably best to stick with this though, I suppose.

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,67 @@ def chmod(file_path, mode):
...
+                           already exists.
+    """
+    # As we know that os.mkdir is called internally by os.makedirs, we will swap the function in
+    # os module for the time of makedirs execution.
+    orig_mkdir_fn = os.mkdir
+    try:
+        os.mkdir = mkdir  # type: ignore
+        try:
+            os.makedirs(file_path, mode)
+        except (IOError, OSError) as err:
```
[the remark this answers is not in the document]
adferrand: Because `exists_ok` parameter does not exist on Python 2. Here is the usual way to get a consistent behavior on makedirs for both python versions. 

I followed here the API of Python 3, since it is the future. However you could argue that `exists_ok` is not used in Certbot codebase, so maybe you prefer that I just remove this feature?

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,67 @@ def chmod(file_path, mode):
...
+    is correctly applied.
+    :param str file_path: The file path to open
+    :param int mode: POSIX mode to apply on leaf directory when created, Python defaults
+                     will be applied if ``None``
+    :param bool exists_ok: If set to ``True``, do not raise exception if leaf directory
+                           already exists.
+    """
+    # As we know that os.mkdir is called internally by os.makedirs, we will swap the function in
+    # os module for the time of makedirs execution.
+    orig_mkdir_fn = os.mkdir
```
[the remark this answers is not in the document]
adferrand: Indeed, done.

certbot/compat/filesystem.py
adferrand: Done.

certbot/compat/filesystem.py
adferrand: I would be tempted to use `FileExistsError` since it subclasses correctly `OSError`, and that this portion of code should run only on Python 3 so we can suppose that the error subclass exists at this point. However I agree with you that it is less obvious when reading the code, so it seems that current implementation is shorter and cleaner for now.

Maybe a refactoring to do once we drop completely Python 2.7.

certbot/compat/filesystem.py
adferrand: Even if GIT would shout at us a merge conflict reminding us to synchronize the two implementations, I took the version in part3c to put it here and so deal with it right now.

certbot/compat/filesystem.py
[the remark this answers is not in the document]
ohemorange: I actually meant to stick that `if` in right before this line... we'll still need to cache `os.mkdir` at the global level on windows.

certbot/compat/filesystem.py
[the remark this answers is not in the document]
ohemorange: For both this and the discussion below, it's actually a good point that we should be hitting this code only on py3, since it's only running on windows. In which case I'm actually perfectly happy using py3-only code, especially since we're going to have to drop py2.7 soon anyway. So might as well pass `exists_ok` through.

certbot/compat/filesystem.py
ohemorange: as I mentioned in the conversation above, since this code will only ever be run on py3, I'm excited about moving to py3-style as much as possible now.

adferrand: See general comment below.

certbot/compat/filesystem.py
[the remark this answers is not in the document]
adferrand: See general comment below.

certbot/compat/filesystem.py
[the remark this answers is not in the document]
adferrand: See general comment below.

adferrand:
Regarding the question about using Python 3 specific code when we are on the Windows part only run by Python 3.

I would like to do it also, in particularly in the scope that we will have to drop Python 2.7 soon.

However at this point, since pylint and mypy are run under Python 2.7, it makes them very very angry when trying to add specific Python 3 code. Calming them down require to add a lot of comments like `#pylint: disable=` or `#type: ignore` that makes the code in fact more complex and maintainable in my opinion.

So I prefer to stick to a simpler implementation, that is cross 2.x/3.x compatible. We should makes Python 3 specific moves when our execution of Pytest and MyPy is properly migrated to Python 3.

In the mean time, the current version of `filesystem.mkdir` cannot end with an infinite loopback reference under Windows anymore, so I remove the specific global reference to `os.mkdir` in `filesystem` module.

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,63 @@ def chmod(file_path, mode):
...
+    :param int mode: POSIX mode to apply on leaf directory when created, Python defaults
+                     will be applied if ``None``
+    :param bool exists_ok: If set to ``True``, do not raise exception if leaf directory
+                           already exists.
+    """
+    orig_mkdir_fn = os.mkdir
+    try:
+        # As we know that os.mkdir is called internally by os.makedirs, we will swap the function in
+        # os module for the time of makedirs execution on Windows.
+        os.mkdir = mkdir if not POSIX_MODE else os.mkdir  # type: ignore
```
ohemorange: `if not POSIX_MODE else os.mkdir` is doubly unnecessary if `mkdir` has its own `if POSIX_MODE`, and if we bail early as I suggest above.

certbot/compat/os.py
```diff
@@ -51,6 +51,28 @@ def chmod(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
                        'Use certbot.compat.filesystem.chmod() instead.')
 
 
+# Very similarly to os.open, os.mkdir has the same effect on Windows, to create an unsecured
+# folder. Same mitigation is provided using security.take_ownership and security.chmod.
+# On top of that, we need to handle the fact that os.mkdir is called recursively by os.makedirs.
+# This is done by protecting the original os.mkdir to have the real logic, call it during the
```
ohemorange: we no longer call the original `os.mkdir`, making this code much simpler in fact!

adferrand:
To elaborate a bit more on after https://github.com/certbot/certbot/pull/6968#issuecomment-506690974, the current implementation is necessary to keep a clean code regarding the parameter `exist_ok` that is available on Python 3 only.

However, I see that this gives still a too much complex code, and in fact `exist_ok` is never call on our codebase, since it still needs to be compatible with Python 2.x. Good candidate for a refactoring when drop Python 2.x, but until that I will stick to Python 2.x API.

certbot/compat/filesystem.py
adferrand: See https://github.com/certbot/certbot/pull/6968#issuecomment-506988582

certbot/compat/filesystem.py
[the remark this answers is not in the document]
adferrand: See https://github.com/certbot/certbot/pull/6968#issuecomment-506988582

certbot/compat/os.py
adferrand: Comment simplified.

certbot/compat/filesystem.py
```diff
@@ -36,6 +43,58 @@ def chmod(file_path, mode):
...
+                     will be applied if ``None``
+    """
+    if POSIX_MODE:
+        return os.makedirs(file_path, mode)
+
+    orig_mkdir_fn = os.mkdir
+    try:
+        # As we know that os.mkdir is called internally by os.makedirs, we will swap the function in
+        # os module for the time of makedirs execution on Windows.
+        os.mkdir = mkdir  # type: ignore
```
ohemorange: why do we need to ignore the type?

certbot/plugins/webroot.py
```diff
@@ -173,10 +174,10 @@ def _create_challenge_dirs(self):
                 # as it does not correspond to a folder path ('/' or 'C:')
                 for prefix in sorted(util.get_prefixes(self.full_roots[name])[:-1], key=len):
                     try:
-                        # This is coupled with the "umask" call above because
-                        # os.mkdir's "mode" parameter may not always work:
-                        # https://docs.python.org/3/library/os.html#os.mkdir
-                        os.mkdir(prefix, 0o0755)
+                        # This is coupled with the "umask" call above because filesystem.mkdir's
+                        # "mode" parameter may not always work under Linux:
+                        # https://docs.python.org/3/library/os.html#security.mkdir
```
ohemorange:
```suggestion
                        # https://docs.python.org/3/library/os.html#os.mkdir
```

certbot/plugins/webroot.py
```diff
@@ -173,10 +174,10 @@ def _create_challenge_dirs(self):
                 # as it does not correspond to a folder path ('/' or 'C:')
                 for prefix in sorted(util.get_prefixes(self.full_roots[name])[:-1], key=len):
                     try:
-                        # This is coupled with the "umask" call above because
-                        # os.mkdir's "mode" parameter may not always work:
-                        # https://docs.python.org/3/library/os.html#os.mkdir
-                        os.mkdir(prefix, 0o0755)
+                        # This is coupled with the "umask" call above because filesystem.mkdir's
```
ohemorange:
```suggestion
                        # This is coupled with the "umask" call above because os.mkdir's
```

certbot/compat/os.py
```diff
@@ -51,6 +51,25 @@ def chmod(*unused_args, **unused_kwargs):  # pylint: disable=function-redefined
                        'Use certbot.compat.filesystem.chmod() instead.')
 
 
+# Very similarly to os.open, os.mkdir has the same effects on Windows and creates an unsecured
+# folder. So a similar mitigation than security.chmod is provided on this platform.
```
ohemorange:
```suggestion
# folder. So a similar mitigation to security.chmod is provided on this platform.
```

certbot/compat/filesystem.py
adferrand: Because our mkdir function does not have exactly the same type signature. And fixing that requires unpleasant tricks to make it work both on python 2.7 and python 3.x, that I do not think it worth the complexity for this specific internal call to mkdir.

adferrand:
@ohemorange I added the test you asked.

Issue #6960, opened 2019-04-13
Implement redirect by default


ohemorange:
https://docs.google.com/document/d/1aKnQYEzCrZgYX-iuE-ErOdIKERUuLmZSX_BDb7d0M6I/edit

# Transitioning users

I think we need to properly announce the change and transition here as running Certbot in non-interactive mode defaults to no redirect.

> Which users will see this message? 

In addition to there being various ways that they could see that output, I'd want the change to be mentioned in our changelog and release announcement. It's not perfect, but I think it's the simple and standard way to try and give people a heads up.

> Of those, which will have problems if we don't do it?

A very small minority. With that said, I think it costs us very little and I think it's good practice to try and help people relying on our software.

> Which users will get confused and erroneously add the flag?

I'm pretty open to the wording on the message and we don't have to mention --no-redirect. It was just the first wording that came to mind.

# Consistent behavior

I think our current behavior is "it works this way except when". I think that because we say "certbot automatically sets up https for you except it doesn't actually unless you select the correct answer to this question". redirect should be part of install, not an enhancement. the fact that it's currently implemented as an enhancement is an implementation detail that shouldn't govern our decisions about certbot's behavior.

> How would this work if the installer doesn't support redirects either because support for it was just never written or because it's for something other than a webserver like postfix where HTTP redirects don't make sense?

The action certbot's default run should take is "turn on https", whatever that means for the server. Honestly, that could even mean switching redirect to be part of deploy for nginx and apache.

> When I initially wrote that question I was worried about having consistent behavior and communicating the server's current state to the user.
>
> Now I think we can pretty easily solve this though by making our own plugins consistent and encouraging third party installer plugins (if they exist by the time we make these changes) to do the same.

bmw:
I would like to bump the priority of this. At the very least, I think changes to the non-interactive flow would be nice to land before Certbot 1.0.

bmw:
Superseded by #7594.

PR #6976, opened 2019-04-19
Warn install users that future versions of certbot will automatically redirect

ohemorange:
First step of #6960.

Warning is printed for all install users, even those who run interactively, so that they know the change is coming and can adjust their usage now. This is because we don't know who the people are who this will affect, so getting the message to the widest audience possible.


certbot/client.py
```diff
@@ -498,6 +498,10 @@ def deploy_certificate(self, domains, privkey_path,
                            "the certificate")
             raise errors.Error("No installer available")
 
+        logger.warning("Future versions of Certbot will automatically configure "
```
bmw: I personally don't think we should unconditionally print warnings. I think it can create problems with automated alerts if they can't be disabled and it discourages people from paying attention to them.

Can we print this only in the case where we interactively ask the user about a redirect and they say (or default to) no redirect? Alternatively we could do it if we ask at all, but people who want the redirect have nothing to worry about.

This could be done in [enhance_config](https://github.com/certbot/certbot/blob/2812f054a3e9bf87a799c9d153393a9d3593fed0/certbot/client.py#L551) or [redirect_by_default](https://github.com/certbot/certbot/blob/2812f054a3e9bf87a799c9d153393a9d3593fed0/certbot/display/enhancements.py).

We may also want to add simple tests about this behavior.

certbot/client.py
```diff
@@ -498,6 +498,10 @@ def deploy_certificate(self, domains, privkey_path,
...
+            "the webserver so that all requests redirect to secure HTTPS access. "
+            "If you do not want this feature, run Certbot with the --no-redirect flag.")
```
bmw: The way this is worded almost seems like it's discouraging people to set up redirects which I don't think we should do. I'm pretty open on the exact wording but what about something like:

> You can control this behavior and disable this warning with the --redirect and --no-redirect flags.

(That's assuming we implement my above suggestion.)

Alternatively, we could actively push people towards setting up redirects.

certbot/client.py
ohemorange: I put it here because we previously talked about users who either have `--quiet` or `--non-interactive` set; are we ok with not notifying these users through Certbot?

ohemorange:
> In addition to my inline comments, can we add a changelog entry about this?

Yes, I was planning to; I like adding changelog entries as the very last thing because the file is merge-conflict-prone.

certbot/client.py
ohemorange: I was actually trying to push them *towards* setting up redirects here, by saying "if you don't want this feature". Your wording works for me though.

certbot/client.py
bmw: I may be missing something, but I think my proposal would still work for people with `--quiet` or `--non-interactive` set. If we can't prompt the user, the [IDisplay method in redirect_by_default](https://github.com/certbot/certbot/blob/2812f054a3e9bf87a799c9d153393a9d3593fed0/certbot/display/enhancements.py#L51) should still be called, it will just return the default response which is no redirect.

certbot/client.py
bmw: Oh. I didn't interpret it that way because it's telling people how to disable the feature, but not how to get it now. I personally prefer my wording over what you initially wrote in this PR, but if there's something else you like better, please propose it!

certbot/client.py
ohemorange: > Can we print this only in the case where we interactively ask the user about a redirect

We don't prompt in `--non-interactive`

certbot/client.py
ohemorange: Nope, I'm cool with whatever.

certbot/client.py
bmw: I see! I should have worded that better. How about:

Can we log this warning only when the user was prompted for a redirect and they said no or they're running non-interactively and they default to no redirect?

This could be by checking for a response from `IDisplay` for no redirect either in [enhance_config](https://github.com/certbot/certbot/blob/2812f054a3e9bf87a799c9d153393a9d3593fed0/certbot/client.py#L551) or [redirect_by_default](https://github.com/certbot/certbot/blob/2812f054a3e9bf87a799c9d153393a9d3593fed0/certbot/display/enhancements.py).

ohemorange: > Can we log this warning only when the user was prompted for a redirect and they said no or they're running non-interactively and they default to no redirect?

Sure

PR #6978, opened 2019-04-19
Fix typo and add instructions for changing a single dependency

ohemorange:


letsencrypt-auto-source/rebuild_dependencies.py
```diff
@@ -217,7 +217,13 @@ def _write_requirements(dest_file, requirements, conflicts):
 # To generate this, do (with docker and package hashin installed):
 # ```
 # letsencrypt-auto-source/rebuild_dependencies.py \\
-#   letsencrypt-auto-sources/pieces/dependency-requirements.txt
+#   letsencrypt-auto-source/pieces/dependency-requirements.txt
+# ```
+# If you want to update a single dependency, run commands similar to these:
+# ```
+# pip install --no-cache-dir -e acme -e . -e certbot-apache -e certbot-nginx
```
adferrand: Why is it required to install certbot ? In my opinion, it is not strictly required to update a dependency (even if strongly advised to test it for certbot, but this is a more general requirement).

ohemorange: good question! these are the original instructions that I just pasted back in. not sure why they were written like that in the first place.

adferrand: Hello @ohemorange ! So about the question here, did you get some information about why there was these instructions to install certbot first ? @bmw maybe you have an idea about an hidden requirement here ?

bmw: I'm not 100% sure, but my guess is that the idea was in order rebuild this file, you could use `pip install` to find all the dependencies of our packages and after that you could use `hashin` to get their hashes.

adferrand: OK then in this case, this is not relevant anymore to update a single requirement for corner cases. Main case is a global update respecting all the dependency tree, and this should be done using `rebuild-dependencies.py`

ohemorange: changed

PR #6986, opened 2019-04-23
Refer to ubuntu in install.rst

### LINKED ISSUE #5758: <https://certbot.eff.org/docs/install.html> should contain the word "ubuntu"
According to wikipedia, ubuntu is "the most popular server Linux distribution".  It's odd then that the page at <https://certbot.eff.org/docs/install.html> doesn't contain the word "ubuntu" at all.

Worse, following the debian instructions:

```
sudo apt-get install certbot
```

results in "Unable to locate package certbot" with the current ubuntu LTS (16.04).

Seems like a big documentation fail to me. :(

adferrand:
Fixes #5758


docs/install.rst
```diff
@@ -218,6 +218,31 @@ repo, if you have not already done so. Then run:
 
    sudo apt-get install certbot python-certbot-apache -t jessie-backports
 
+**Ubuntu**
+
+If you run Ubuntu Trusty, Xenial or Bionic, certbot is available through the official PPA,
```
jsoref:
```suggestion
If you run Ubuntu Trusty, Xenial, or Bionic, certbot is available through the official PPA,
```

devec0:
@schoen, @bmw - this LGTM, can we merge this?

PR #6989, opened 2019-04-23
 [Unix] Create a framework for certbot integration tests: PART 5-FINAL

adferrand:
This is the final part to integrate #6541. This PR connect certbot-ci to Travis, replacing the old execution of integration tests. This PR removes also all bash scripts related to these tests.

To arguable modifications are done to other tests to make all work:
1) A specific bash script is added to Apache plugin source code, in order to allow correct execution of `apacheconftests`
2) The script `tests/lock_test.py` is modified in a not so elegant way to reuse the logic of certbot-ci to configure Nginx.

My approach here is to make the modifications with the lowest impact, while planning to migrate eventually these tests into certbot-ci (where they better fits, along `nginxcompat`) in a future PR.


adferrand:
Need to migrate test-everything before merging.

adferrand:
Ok, I made the required modifications to run all tests in the nightly cron job using certbot-ci. An example of successful build in this situation can be found here: https://travis-ci.com/certbot/certbot/builds/109499348

@bmw, PR is ready for review.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -79,7 +80,11 @@ def cleanup():
                 pass
             print('=> Finished tear down of {0} instance.'.format(acme_type))
 
-        shutil.rmtree(workspace)
+        try:
+            shutil.rmtree(workspace)
+        except (OSError, IOError):
+            # Not critical
```
bmw: I agree this isn't critical, but do you know why this is happening? I've only seen it happen in the oldest tests.

tests/lock_test.py
```diff
@@ -108,13 +108,16 @@ def set_up_nginx_dir(root_path):
     """
     # Get the root of the git repository
     repo_root = check_call('git rev-parse --show-toplevel'.split()).strip()
-    conf_script = os.path.join(
-        repo_root, 'certbot-nginx', 'tests', 'boulder-integration.conf.sh')
-    # Prepare self-signed certificates for Nginx
+    # We add manually nginx_config module, because certbot_integration_tests may not be installed.
```
bmw: You're right that this isn't exactly elegant. How about we leave this file and `boulder-integration.conf.sh` alone for now and in a subsequent PR move this test into `certbot-ci`?

Most of the code in this file would no longer be needed if it was running inside `certbot-ci`. We could have one test in the Certbot tests and one in the Nginx tests.

As an aside, I wonder if we should make another set of tests in `certbot-ci` in the future which would be system tests that don't require an ACME server. This file, most of our test for `cli.py` and `main.py`, and some of the current integration tests fall into this category.

tox.ini
```diff
@@ -257,9 +252,25 @@ commands =
...
-passenv =
-    DOCKER_*
+passenv = DOCKER_*
+
+[testenv:integration-certbot-oldest]
+commands =
+    {[base]pip_install} .
+    {[base]pip_install} certbot-ci
+    pytest {toxinidir}/certbot-ci/certbot_integration_tests \
+        --acme-server={env:ACME_SERVER:pebble} --campaign=certbot -p no:sugar
```
bmw: Why no sugar?

tox.ini
```diff
@@ -257,9 +252,25 @@ commands =
...
+passenv = DOCKER_*
+setenv = CERTBOT_OLDEST=1
```
bmw: This may depend on your answers to the above questions, but why not use `{[testenv:py27-oldest]setenv}` here?

.travis.yml
```diff
@@ -30,12 +30,17 @@ matrix:
   include:
     # Main test suite
     - python: "2.7"
-      env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=all TOXENV=py27_install
+      env: ACME_SERVER=boulder-v1 PYTEST_ADDOPTS=-n4 TOXENV=integration
```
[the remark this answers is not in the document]
adferrand: 1) I found from testing various values that this setting was the most efficient one when integration tests are run in Travis

2) Because we care less about the execution time during nightly tests, so I let the default behavior, that is to take the number of CPU (so 2), and avoid unnecessarily settings to Travis in this case

3) I think it is doable to set this up in tox.ini, for the integration target env, when `TRAVIS` env is detected

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Well failure to delete here is due to the OCSP test configuration stored in the assets that contains some files with `0744` permissions, not owned by current user. So current user in Travis fail to delete them. Obviously a `sudo` call would help here, but I wanted to avoid to rely on this, since it is not very portable (for various reasons in a given test env, `sudo` may not be usable/available.

tox.ini
adferrand: Because in the context of oldest tests, `pytest-sugar` does not work with the pinned old versions of pytest component installed by `pip_install.py`

tox.ini
adferrand: In the context of separating unit oldest tests and integration oldest tests, I think it would add unnecessarily an indirection between two tests suite that are (beside the oldest requirements), are not about the same subject.

.travis.yml
```diff
@@ -102,64 +107,64 @@ matrix:
       env: TOXENV=py37 CERTBOT_NO_PIN=1
       <<: *extended-test-suite
     - python: "2.7"
-      env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=certbot TOXENV=py27-certbot-oldest
+      env: ACME_SERVER=boulder-v1 TOXENV=py27-certbot-oldest,integration-certbot-oldest
```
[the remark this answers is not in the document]
adferrand: Certainly!

.travis.yml
```diff
@@ -30,12 +30,17 @@ matrix:
...
     # Main test suite
     - python: "2.7"
-      env: BOULDER_INTEGRATION=v1 INTEGRATION_TEST=all TOXENV=py27_install
+      env: ACME_SERVER=boulder-v1 PYTEST_ADDOPTS=-n4 TOXENV=integration
       sudo: required
       services: docker
       <<: *not-on-master
     - python: "2.7"
-      env: BOULDER_INTEGRATION=v2 INTEGRATION_TEST=all TOXENV=py27_install
+      env: ACME_SERVER=boulder-v2 PYTEST_ADDOPTS=-n4 TOXENV=integration
```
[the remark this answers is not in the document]
adferrand: Yes integration tests against `boulder-v1` and `boulder-v2` can move to nightly. The only thing you miss using pebble is the OCSP status live tests.

cpu:
> @cpu, as a heads up, this PR could break some things boulder side. Most of our integration testing code is moving into this new framework that Adrien has been working on.
>
> I noticed https://github.com/letsencrypt/boulder/blob/30d155911bc58386da910e3be2e43f961dee3c26/test/integration-test.py#L60 and https://github.com/letsencrypt/boulder/blob/0c700143bb7a2f7dbea2b1bbc9973875e0b8bd24/README.md#working-with-certbot. Do you all still use either of these features?

:wave: @bmw Thanks for the heads-up! I don't think we use the `run_client_tests` stuff at all. It can probably be removed.

I do still use the "working with certbot" instructions, the `_common.sh` script and `certbot_test` on occasion and would like to keep a documented process in place for that in the Boulder README. Is it just a matter of adapting to @adferrand's work?

adferrand:
Yes @cpu, it is just a matter of some adaptation. The logic that was hold by `_common.sh` is still here, it is just rewritten into python. The adaptation to do is that the logic is not callable anymore as a standalone script, since it is included as a method of a class representing a fixture.

I think we should extract this out in a both callable and importable python script, to be used either by the pytest fixture of from the cli.

I will have some time tonight (UTC), so we can discuss about it in mattermost, and find the best way to implement the solution.

.travis.yml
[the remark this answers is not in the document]
bmw: 1. Yeah? What's the time difference? Are our other tests also faster with PYTEST_ADDOPTS=4?

2. If there's a significant speed increase, I'd personally also like to see it added to the nightly tests. Makes doing tested for things like releases just a little bit faster.

certbot-ci/certbot_integration_tests/utils/acme_server.py
bmw: Really? Why doesn't the current user own the files?

While this isn't a big deal, I'd ideally like to fix this so the tests properly clean up after themselves.

tests/lock_test.py
bmw: Keeping files like `tests/integration/_common.sh` around to help with running a local Pebble/Boulder and configuring Certbot to use it so you can run tests manually isn't a bad idea. I think most of the core team never does this (the staging server and cheap domains and VPSes are great), but I think it's nice to have for people who don't have those options.

If we keep `tests/integration/_common.sh`, I think we should:

1. Rename it.
2. Keep some of the boulder/pebble start scripts around.

I'd also like to see `tests/integration/_common.sh` simplified and the boulder/pebble start scripts refactored to share as much logic with `certbot-ci` that makes sense, but that could be done in a later PR if you prefer.

tox.ini
bmw: Deciding when it makes sense to couple things is subjective and kind of tricky in my opinion. Regardless of the approach taken, I think we should also be setting `PYTEST_ADDOPTS` to make use of multiple processes.

I think setting `PYTHONHASHSEED` which is the only other value in `{[testenv:py27-oldest]setenv}` isn't a bad idea either for reproducibility (see https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed), but I personally don't really care one way or another here.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Because it is owned by uid=0 if I remember correctly, so root. I need to check again, but I think it was about sensible files with specific permissions, so it may match some security logic in certbot.

.travis.yml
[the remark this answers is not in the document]
adferrand: 1) It is about a 10% less, so 1min or 1min 30s. The gain is marginal for unit tests. It is because in integration tests you have some steps where nothing is done, just waiting (for instance to return a http response, so increasing the parallelization "fills the gaps".

2) OK! I think that if I am able to put that in tox.ini, all integration tests can take advantage of it.

tests/lock_test.py
adferrand: So as discussed, I will keep for now:
1) `common.sh`
2) Pebble and Boulder setup bash scripts
3) nginx config bash script

They will be removed in a future PR to be replaced by appropriate solutions in Certbot-CI.

adferrand: Done.

.travis.yml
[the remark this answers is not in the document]

.travis.yml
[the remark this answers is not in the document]

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Ok, it seems that we can forget about this, the local copy done by `shutil.copytree` is entirely cleanable by the current user from the tests I did locally and on Travis. So I removed the try/except.

.travis.yml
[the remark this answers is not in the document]
adferrand: 2. and 3. done through setenv inheritance (see relevant conversation below) + a `before_script` entry to export conditionally `PYTEST_ADDOPTS="--numprocesses 4"`.

tox.ini
adferrand: Here I see the interest to inherit `setenv` configs from more generic tox env. I added it, and use it to also configure specifically `PYTEST_ADDOPTS` during integration tests.

adferrand:
On top of the modifications required in conversations, I also transformed the package specific plugin `conftest.py` in `certbot-ci/certbot_integration_tests` into a fully-fledged setuptools plugin, auto-discovered using the `pytest11` entrypoint. It has the desirable following advantages:
* all pytest hooks are available
* loading of the plugin is not tied anymore to correct resolution of `rootdir` in Pytest: it means that there weird workaround to limit tests on certbot or nginx only is not needed anymore, and the target tests can be passed as a complete package namespace using `--pyargs` like for every other tests in Certbot
* so no `--campaign` flag anymore
* plugin is triggered by passing `--acme-server` flag to pytest, that is available as soon as `certbot-ci` is installed, and so can be used in any pytest execution
* in the future this plugin can become an extensible class or composable plugin with other pytest plugins

However an drawback is that pytest plugin auto-discovery does not work in oldest tests, for the same reason than `pytest-sugar`: it require `pyparsing>=2.0.2`, while only `pyparsing==1.5.6` is installed because of centos 6. I added a workaround in these oldest tests, since a prefer to put a workaround in oldest tests than in main tests.

As an alternative, `pyparsing` could be pinned to `2.0.2` in oldest tests. Indeed, 2.X branch is not available in CentOS 6 because 2.X branch is not compatible anymore with Python < 2.6. However, there is no other breaking change from 1.5.6 (https://github.com/pyparsing/pyparsing/blob/f0264bd8d1a548a50b3e5f7d99cfefd577942d14/CHANGES#L985), `pyparsing` has no requirement, and oldest tests are executed on Python 2.7.

@bmw, PR ready for review.

adferrand:
Associated equivalent nightly tests for the current state of this PR can be found here: https://travis-ci.com/certbot/certbot/builds/111378938

adferrand:
Oups, `integration-certbot-oldest-tests` was not entirely fixed. Done now, and new equivalent nightly tests execution is here: https://travis-ci.com/certbot/certbot/builds/111383938

adferrand:
Finally, workaround is not sufficient, as the nodes fails on the same issue about pyparsing version.

Regarding the limited effect of pinning pyparsing to 2.0.2 described above, I added a specific `local-oldest-requirements.txt` on certbot-ci to install `pyparsing==2.0.2`. This is effective in oldest tests only when integration tests are invoked.

tox.ini
```diff
@@ -249,28 +249,29 @@ passenv = DOCKER_*
...
     {[base]pip_install} acme . certbot-nginx certbot-ci
-    pytest {toxinidir}/certbot-ci/certbot_integration_tests \
+    pytest --pyargs certbot_integration_tests \
         --acme-server={env:ACME_SERVER:pebble} \
         --cov=acme --cov=certbot --cov=certbot_nginx --cov-report= \
         --cov-config={toxinidir}/certbot-ci/certbot_integration_tests/.coveragerc
     coverage report --include 'certbot/*' --show-missing --fail-under=67
     coverage report --include 'certbot-nginx/*' --show-missing --fail-under=74
 passenv = DOCKER_*
+setenv = {[testenv]setenv}
```
bmw: I don't think we need this because the value should be inherited from `[testenv]`. See https://tox.readthedocs.io/en/latest/config.html#tox-environment-settings.

.travis.yml
```diff
@@ -102,64 +99,84 @@ matrix:
...
       dist: xenial
-      env: TOXENV=py37 BOULDER_INTEGRATION=v1
+      env: ACME_SERVER=boulder-v1 TOXENV=py37,integration
       sudo: required
       services: docker
       <<: *extended-test-suite
     - python: "3.7"
       dist: xenial
-      env: TOXENV=py37 BOULDER_INTEGRATION=v2
+      env: ACME_SERVER=boulder-v2 TOXENV=py37,integration
```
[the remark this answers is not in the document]
adferrand: Ah yes, it makes sense. Done.

.travis.yml
```diff
@@ -102,64 +99,84 @@ matrix:
...
+      <<: *extended-test-suite
+    - python: "2.7"
+      env: TOXENV=py27-certbot-oldest
+      sudo: required
+      services: docker
+      <<: *extended-test-suite
+    - python: "2.7"
+      env: TOXENV=py27-nginx-oldest
+      sudo: required
+      services: docker
```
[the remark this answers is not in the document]

tox.ini
adferrand: Indeed, fixed.

certbot-ci/local-oldest-requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Even if only pyparsing 1.5.6 is available on CentOS 6, version 2.0.2 is required
+# to make pytest plugins auto-discovery working. No breaking change is done from 1.5.6
+# besides removal of support to Python < 2.6, that is irrelevant for oldest tests.
+pyparsing==2.0.2
```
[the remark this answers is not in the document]
adferrand: Why the nginx plugin tests are not accurate anymore with this ?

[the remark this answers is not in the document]
adferrand: Ah. Nginx plugin is using directly pyparsing. OK, I reconsider my approach.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -567,6 +568,9 @@ def test_ocsp_status_live(context):
 
     # OSCP 2: Check live certificate OCSP status (REVOKED)
     context.certbot(['revoke', '--cert-name', cert, '--no-delete-after-revoke'])
+    # Sometimes in oldest tests (using openssl binary and not cryptography), the OCSP status is
+    # not seen immediately by Certbot as invalid. Waiting few seconds solves this transient issue.
+    time.sleep(5)
```
bmw: This change came in while I was reviewing the PR and I missed it.

Instead of doing this, I'd rather fix the problem in Certbot? Are we not specifying enough wiggle room with validity times to OpenSSL?

adferrand: The problem is transient, only observed on Travis so far, not locally. I not know how to recreate a reliable pattern and find precisely what needs to be done.

I would like to avoid to block this PR on that, since it is unrelated. We could open an issue to track finding a way to troubleshoot and fix.

certbot-ci/local-oldest-requirements.txt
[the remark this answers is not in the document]
adferrand: Done, see below.

certbot-ci/certbot_integration_tests/utils/pytest_plugin.py
```diff
@@ -18,33 +18,18 @@ def pytest_addoption(parser):
     Standard pytest hook to add options to the pytest parser.
     :param parser: current pytest parser that will be used on the CLI
     """
-    parser.addoption('--acme-server', default='pebble',
+    parser.addoption('--acme-server',
```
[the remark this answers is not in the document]
adferrand: Done, see below.

adferrand:
So I reimplemented again a local directory plugin in `contest.py`, but this time alongside with an empty `pytest.ini` that makes the rootdir finding algorithme of pytest to set correctly rootdir to be ̀certbot-ci` directory. From that the `conftest.py` file is correctly found beneath `certbot_integration_tests` module, and trigger the ACME server deployment logic with the need to add back the `--campaign` flag

The only necessary condition is to invoke pytest on the module path, not through `--pyargs` package name.

Associated pseudo nightly build is: https://travis-ci.com/certbot/certbot/builds/111423483

Also during execution of nightly build, I found a transient error when executing `integration-certbot-oldest`: sometimes, tests are failing because the OCSP status of a certificate revoked, and immediately checked after that, is not yet adverted as revoked by the local Boulder instance. Adding some seconds before checking the OCSP status solve the transient issue. We should find a way to remove this spurious failure in a future PR.

adferrand:
Updated nightly tests: https://travis-ci.com/certbot/certbot/builds/111427881

adferrand:
So in the end I removed pytest-sugar. On top of the error associated to pyparsing version, that requires a workaround on oldest tests, the output generated by this plugin did not fit that well in the context of Travis CI (in term of readeness and usefulness of the addition information added in the console output).

Now, finally, all jobs are green in nightly tests!

https://travis-ci.com/certbot/certbot/builds/111440485

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
bmw: Do you have a log of this failure?

adferrand:
Indeed, it seems to work without `pytest.ini` now. I remove it.

New nightly tests: https://travis-ci.com/certbot/certbot/builds/111651489

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
adferrand: https://travis-ci.com/certbot/certbot/builds/111407674

adferrand:
I also test the oldest integration tests without the wait in OCSP, maybe it was related to slow network issues in Docker encountered in Travis the last days: https://travis-ci.com/certbot/certbot/builds/111652568

bmw:
That seemed to have worked. If it wasn't failing on every run for you, it may be worth testing a couple of times, but I think Docker network issues certainly could have caused this. I triggered the tests again at https://travis-ci.com/certbot/certbot/builds/111662041.

adferrand:
@bmw, I added #7058 to track resolution of spurious failures during OCSP live tests.

Issue #6258, opened 2018-07-24
Allow querying an account without calling new_account in ACMEv2


bmw:
If you do this with boulder, you get:
```
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 130, in query_registration
    return self._send_recv_regr(regr, messages.UpdateRegistration())
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 75, in _send_recv_regr
    response = self._post(regr.uri, body)
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 92, in _post
    return self.net.post(*args, **kwargs)
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 1109, in post
    return self._post_once(*args, **kwargs)
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 1123, in _post_once
    return self._check_response(response, content_type=content_type)
  File "/Users/bmw/Development/certbot/certbot/acme/acme/client.py", line 983, in _check_response
    raise messages.Error.from_json(jobj)
acme.messages.Error: urn:ietf:params:acme:error:malformed :: The request message was malformed :: No Key ID in JWS header
```
This is because [querying an account requires kid to be set](https://tools.ietf.org/html/draft-ietf-acme-acme-13#section-6.2), but if [self.net.account isn't set, kid isn't included](https://github.com/certbot/certbot/blob/c129ab2965f7ec4edf173b1cd900bd5add13cd2a/acme/acme/client.py#L938).

An easy fix for this would be to set `self.net.account = regr` in the beginning of `query_registration`, at least for ACMEv2. Doing it at the end of `query_registration` to store the account as given by the ACME server isn't a bad idea either.

This is useful for the scenario in #6246.

bmw:
Fixed in #6251.

evilaliv3:
@bmw @dovreshef: is there any easy fix that one can do in 0.25.1?

It would be great if you could suggest one as there are so many users that will continue to use suchversion given that is present in many debian/ubuntu versions.

Which is the correct code that a client that hasalready performed the registration should execute in order to renew it?

PR #6993, opened 2019-04-26
Send a POST-as-GET request to query registration in ACME v2

### LINKED ISSUE #6962: client.query_registration uses POST instead of POST as GET
[Here](https://github.com/certbot/certbot/blob/d7610c1ae73699cbac7d3433d8ff286d201ea018/acme/acme/client.py#L134) acme module uses empty UpdateRegistration message (POST with payload `{}`), instead of using standard POST as GET (POST with `null` content), resulting in some servers (`pebble` in strict mode) rejecting these requests.

adferrand:
Fixes #6962

This PR modify `query_registration` method in `ClientV2` to use a POST-as-GET request, as specified by the RFC. This allows Pebble, that enforce this behavior, to work correctly with our `acme` module. `Client` (ACME v1) keep the old behavior.

Steps to reproduce the error, and confirm correct behavior in this PR:
```
cd /dev
git clone https://github.com/letsencrypt/pebble.git
cd /dev/pebble
docker-compose up -d
curl --request POST --data '{"ip":"10.30.50.1"}' http://localhost:8055/set-default-ipv4
cd /dev
git clone https://github.com/adferrand/certbot.git --branch post-as-get-query-registration
# Prepare certbot dev venv as usual
# In acme/examples/http01_example, set DIRECTORY_URL to: https://localhost:14000/dir
# In acme/examples/http01_example, set PORT to: 5002
(venv) REQUESTS_CA_BUNDLE=../pebble/test/certs/pebble.minica.pem python acme/examples/http01_example.py
```

Without the PR, pebble will send a `malformed` error when `query_registration` is invoked. With it, the certificate issuance finishes without errors.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -606,7 +606,9 @@ class ClientV2(ClientBase):
         self.net.account = regr  # See certbot/certbot#6258
         # ACME v2 requires to use a POST-as-GET request (POST an empty JWS) here.
         # This is done by passing None instead of an empty UpdateRegistration to _post().
-        self.net.account = self._send_recv_regr(regr, None)
+        response = self._post(regr.uri, None)
+        self.net.account = self._regr_from_response(response, uri=regr.uri,
+                                                    terms_of_service=regr.terms_of_service)
         return self.net.account
 
     def update_registration(self, regr, update=None):
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -603,6 +603,9 @@ class ClientV2(ClientBase):
             Resource.
 
         """
+        self.net.account = regr  # See certbot/certbot#6258
+        # ACME v2 requires to use a POST-as-GET request (POST an empty JWS) here.
+        # This is done by passing None instead of an empty UpdateRegistration to _post().
         self.net.account = self._send_recv_regr(regr, None)
         return self.net.account
 
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -123,15 +123,6 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         """
         return self.update_registration(regr, update={'status': 'deactivated'})
 
-    def query_registration(self, regr):
-        """Query server about registration.
-
-        :param messages.RegistrationResource: Existing Registration
-            Resource.
-
-        """
-        return self._send_recv_regr(regr, messages.UpdateRegistration())
-
     def _authzr_from_response(self, response, identifier=None, uri=None):
         authzr = messages.AuthorizationResource(
             body=messages.Authorization.from_json(response.json()),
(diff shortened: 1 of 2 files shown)


acme/acme/client.py
```diff
@@ -603,10 +603,8 @@ def query_registration(self, regr):
             Resource.
 
         """
-        self.net.account = regr
```
ohemorange: Looks like this line is here for a reason: https://github.com/certbot/certbot/issues/6258

acme/acme/client.py
```diff
@@ -603,10 +603,8 @@ def query_registration(self, regr):
...
-        updated_regr = super(ClientV2, self).query_registration(regr)
-        self.net.account = updated_regr
-        return updated_regr
+        self.net.account = self._send_recv_regr(regr, None)
```
ohemorange: Let's add a line of comment explaining why it's important to have it pass `None` instead of an empty `UpdateRegistration` message

acme/acme/client.py
adferrand: Added the line back.

acme/acme/client.py
adferrand: Added a comment here.

ohemorange: Actually, I just noticed that there's a whole separate POST-as-GET function. Once we've removed the retry and deprecated ACMEv1 codepaths, we'll probably want this code to have been calling directly into the POST-as-GET function so we can keep those grouped together, no? Does the spec say that these calls should be done with POST-as-GET specifically, or does it say "post but with an empty body?

adferrand: Yes certainly we will want to do that. The spec says something more closer to the first assertion than the second one. It says that POST-as-GET is supposed to be used everywhere you are supposed to fetch a ressource, without saying it explicitly for the account. I suppose the migration path here is to make fetching accounts be done in a consistent way with the other resources.

ohemorange: Is there a reason to not fetch accounts in a consistent way with other resources in this batch of changes?

adferrand: I do not understand your question. When I was talking about a consistent way in my previous response, I was refering to the fact that the specification in its current state turns out to require to fetch all ressources with a POST-as-GET request.

If your question is about why `_post_as_get()` function is not use here, you answered yourself previously: to be done in the future when there is no fallback to GET anymore an ACMEv1.

ohemorange: Ah, I see there's some grammatical confusion. My earlier comments meant that once we've cleaned everything up, we will want to not need to make further changes; that is, to right now put the code in a state such that when we make that future change, everything will fit nicely together, and we won't have to go back and fix this. So to do this PR in a way that means we won't have to touch this codepath again in the future. Is that not possible?

adferrand: Ok I see. Indeed, with this in mind, the best design is to use directly a `_post()` then a `_regr_from_response()`. It is then consistent with the other calls where a `_post_as_get()` is done then a `_*_from_response()` (like `_authzr_from_response()`.

This way, when the `_post_as_get()` will stop to fallback to a `_get()`, we can simply replace the `_post()` by a `_post_as_get()` in `query_registration`.

I did that.

ohemorange: That's a great idea, thank you!

PR #6994, opened 2019-04-26
Remove POST-as-GET fallback to GET

adferrand:
In #6522, I added support for POST-as-GET requests. In that PR there was also a fallback: if a given POST-as-GET request is returned as `malformed`, then a second request is done using GET. This fallback was designed to give some time to some ACME servers that would not have implemented yet the POST-as-GET protocol.

We can now considering two things:
* this fallback has been here for 5 months, giving time to developers to adapt
* ACME v2 is now an RFC, and so its specifications can legitimately being enforced

Because of theses two reasons, I think we can now remove this fallback and simplify the code. This PR does that.


ohemorange:
I'm not sure that all ACME servers are actually ready; I can send some emails around, but I'd want to confirm with people before making this change.

ohemorange:
I never heard back about this. I think it's fine to just go ahead with it at this point.

ohemorange:
Actually, I probably approved too soon -- we almost certainly want to mention this in the changelog. Mind making a PR to make that change, @adferrand?

adferrand:
No problem, I do that right now.

PR #6995, opened 2019-04-27
Print warning when certbot-auto has insecure permissions.

bmw:
This PR attempts to better inform people about the problem identified at https://community.letsencrypt.org/t/certbot-auto-deployment-best-practices/91979/.

I'd like to include this PR in our release on May 1st and since making any significant change to `certbot-auto` in its current state stresses me out, I'd ideally like to have two reviews on this. 

@erikrose and @joohoi, are you able to review this by early Tuesday morning PST at the latest so I have time to address review comments? If not, no problem, but please let me know ASAP so I can try and find another reviewer in time.

I was hesitant to add the flag `--no-permissions-check`, however, if there's some obscure distro out there (or custom user setup) that has a strange users and groups, I didn't want us to either:

* Have to put out a bug fix release
* Refuse to fix the problem and let them deal with warnings on every run

In addition to running our "test farm" tests (on the OSes specified [here](https://github.com/certbot/certbot/blob/master/tests/letstest/targets.yaml)) using the modified `test_letsencrypt_auto_certonly_standalone.sh` test script, I successfully tested on these changes on macOS and FreeBSD.

### DIFF
--- a/letsencrypt-auto-source/pieces/check_permissions.py
+++ b/letsencrypt-auto-source/pieces/check_permissions.py
@@ -1,9 +1,9 @@
 """Verifies certbot-auto cannot be modified by unprivileged users.
 
 This script takes the path to certbot-auto as its only command line
-argument. It then checks that the file can only be modified by uid/gid 0
-and if other users can modify the file, it prints a warning with a
-suggestion on how to solve the problem.
+argument.  It then checks that the file can only be modified by uid/gid
+< 1000 and if other users can modify the file, it prints a warning with
+a suggestion on how to solve the problem.
 
 If the absolute path of certbot-auto contains a symlink, it is not
 handled specially and the symlink is followed. Due to this, there could
@@ -30,7 +30,21 @@ def has_safe_permissions(path):
     """Returns True if the given path has secure permissions.
 
     The permissions are considered safe if the file is only writable by
-    uid/gid 0.
+    uid/gid < 1000.
+
+    The reason we allow more IDs than 0 is because on some systems such
+    as Debian, system users/groups other than uid/gid 0 are used for the
+    path we recommend in our forum post which is /usr/local/bin.  1000
+    was chosen because on Debian 0-999 is reserved for system IDs[1] and
+    on RHEL 0-500 is reserved[2]. Debian recommends normal uids start at
+    uid 1000 and RHEL recommends uid 5000 to allow them to increase the
+    range used for system IDs in the future. It's possible that the
+    threshold of 1000 is too high, however, this seems unlikely and
+    avoids printing warnings when we're not confident there is a
+    problem.
+
+    [1] https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes
+    [2] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-managing_users_and_groups
 
     :param str path: filesystem path to check
     :returns: True if the path has secure permissions, otherwise, False
@@ -40,9 +54,9 @@ def has_safe_permissions(path):
     stat_result = os.stat(path)
     if stat_result.st_mode & stat.S_IWOTH:
         return False
-    if stat_result.st_mode & stat.S_IWGRP and stat_result.st_gid != 0:
+    if stat_result.st_mode & stat.S_IWGRP and stat_result.st_gid >= 1000:
         return False
-    if stat_result.st_mode & stat.S_IWUSR and stat_result.st_uid != 0:
+    if stat_result.st_mode & stat.S_IWUSR and stat_result.st_uid >= 1000:
         return False
     return True
 
--- a/letsencrypt-auto-source/pieces/check_permissions.py
+++ b/letsencrypt-auto-source/pieces/check_permissions.py
@@ -5,11 +5,11 @@ argument.  It then checks that the file can only be modified by uid/gid
 < 1000 and if other users can modify the file, it prints a warning with
 a suggestion on how to solve the problem.
 
-If the absolute path of certbot-auto contains a symlink, it is not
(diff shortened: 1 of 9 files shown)


bmw:
I'd still like two reviews on this if possible, but @adferrand, do you think you'd be able to give this a quick review?

adferrand:
Sure.

letsencrypt-auto-source/pieces/check_permissions.py
```diff
@@ -0,0 +1,80 @@
...
+        return False
+    if stat_result.st_mode & stat.S_IWGRP and stat_result.st_gid >= 1000:
+        return False
+    if stat_result.st_mode & stat.S_IWUSR and stat_result.st_uid >= 1000:
+        return False
+    return True
+
+
+def main(certbot_auto_path):
+    current_path = os.path.abspath(certbot_auto_path)
```
adferrand: OK. I tried very hard, and finally found an insane corner case. I am not sure that it worth to study it, but let's expose it.

Let's consider that the real path of certbot-auto is `/root/certbot/certbot-auto`. Let's consider also that we have a symlink `/tmp/certbot` pointing to `/root/certbot` AND some insane admin sys executed the command `chown -h myuser:myuser /tmp/certbot` (or he created the symlink as a normal user, not using sudo ...).

Let's consider now:

1) That certbot-auto is called through `/tmp/certbot/certbot-auto`.
In this case, the input from `"$0"` is `/tmp/certbot/certbot-auto`, and path resolution here will be `/tmp/certbot/certbot-auto`. When checking `/tmp/certbot`, so on the symlink owned by `myuser` pointing to `/root/certbot` owned by `root`, will raise a `PermissionDenied` (even running as root, as the permissions between the symlink and its target are not corresponding.

2) That certbot-auto is called from `/tmp/certbot` using `./certbot-auto`
In this case, the input from `"$0"` is `./certbot-auto` and path resolution is `/root/certbot/certbot-auto` because `abspath` will resolve the symlinks. It means that the permissions will be seen as safe. But `/tmp/certbot` is owned by `myuser`, that can delete the symlink and point it to somewhere else.

In the first situation, I think that we should check if the current path is in fact a symlink, and check its ownership. In the second situation, I do not know how to get the path including symlinks from bash `"$0"` variable.

letsencrypt-auto-source/pieces/check_permissions.py
```diff
@@ -0,0 +1,80 @@
...
+    [1] https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes
+    [2] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/ch-managing_users_and_groups
+    [3] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-managing_users_and_groups
+
+    :param str path: filesystem path to check
+    :returns: True if the path has secure permissions, otherwise, False
+    :rtype: bool
+
+    """
+    stat_result = os.stat(path)
```
adferrand: Can we make a comment here to say that `os.stat` is resolving the symlinks before fetching ownership ?

letsencrypt-auto-source/tests/auto_test.py
```diff
@@ -192,7 +192,7 @@ def install_le_auto(contents, install_path):
     chmod(install_path, S_IRUSR | S_IXUSR)
 
 
-def run_le_auto(le_auto_path, venv_dir, base_url, **kwargs):
+def run_le_auto(le_auto_path, venv_dir, base_url=None, le_auto_args_str=None, **kwargs):
```
adferrand: Since a string is immutable, it is safe to put it as a default argument value in a method. So for the readability of the method, we can set `le_auto_args_str='--version'` in the `run_le_auto` signature.

letsencrypt-auto-source/pieces/check_permissions.py
bmw: Nice catch! Case (1) is a type of problem I hadn't considered. Luckily the code at https://github.com/certbot/certbot/pull/6995/files#diff-a4881cc6ba1c9dcaeb3962d88a0c369dR1598 seems to properly work to silence all error output and keep certbot-auto from crashing, but this exception certainly isn't the behavior I had in mind.

To fix this, I changed the call to `os.path.abspath` to `os.path.realpath` to eliminate any symlinks in the path upfront to avoid problems like this.

For the permission problem described in (2), I think we should ignore problems like this. I wrote more about this at https://github.com/certbot/certbot/pull/6995/files#diff-932f0067f1e5ad8700ea2c5b1a14a043R8.

adferrand: I agree!

letsencrypt-auto-source/tests/auto_test.py
```diff
@@ -395,3 +405,95 @@ def test_pip_failure(self):
                 else:
                     self.fail("Pip didn't detect a bad hash and stop the "
                               "installation.")
+
+    def test_permissions_warnings(self):
+        """Make sure letsencrypt-auto properly warns about permissions problems."""
+        # This test assumes that only the parent of the directory containing
+        # letesncrypt-auto (usually /tmp) may have permissions letsencrypt-auto
```
joohoi:
```suggestion
        # letsencrypt-auto (usually /tmp) may have permissions letsencrypt-auto
```

bmw:
Thanks for the quick reviews @adferrand and @joohoi!

I do need one more review from one of you since I fixed the typo Joona noticed.

> ...not sure about how the GH UI works, especially if the change is done within the UI)

There's a setting for protected branches on GitHub to dismiss stale pull request reviews which we have turned on. I personally think it's useful. As far as I know, changes done in the UI aren't treated specially.

PR #7001, opened 2019-04-29
Move nginx_compat to nightly tests

adferrand:
With the various optimizations already done and upcoming (certbot-ci), the time execution of integration tests have significantly decreased, allowing potentially a complete execution of a Travis PR job to be done within 5min30s.

However, one job is significantly longer that the other ones after this migration: this is `nginx_compat`, that takes more that 11min to finish. I tried to split the `nginx_compat` in terms of tested configuration and of tests to execute (auth, install, enhance). Both are not satisfactory:
* splitting by configuration may work, but add a significant complexity in the tests
* splitting by tests type is supported almost out-of-the-box, but fails to make two fast tests (see https://travis-ci.org/adferrand/certbot/builds/525892885?utm_source=github_status&utm_medium=notification for instance)

Since these tests are designed to check corner cases on the nginx parser, this is mostly useless to execute them on each PR, as the nginx parser is rarely updated.

After some discussion with @bmw, I think that we can just move the `nginx_compat` from the PR tests to the nightly tests. This PR does that.

cc @bmw @ohemorange


bmw:
It's funny, since we talked about this @adferrand, I noticed that these tests actually did prevent failures landing in `master` in https://github.com/certbot/certbot/pull/6510.

With that said, I'm still personally in favor of this change. I think we can try this and reevaluate if many broken nginx PRs are slipping through to `master`. If you resolve merge conflicts, I'll merge this.

adferrand:
@bmw I solved the merge conflicts. Agree with you, and if these tests were never catching anything, then it would be useless to keep them at all ^^

PR #7002, opened 2019-04-30
Update test farm version of boulder to current master

ohemorange:
Recent changes are no longer compatible with the old version of boulder used in the test farm tests. This PR updates the version of boulder used, and runs it with the new way of running boulder.

A new ami was created and is used here that uses Ubuntu 18.04, so that docker-compose can be installed more properly.

Removed commented-out section about rabbitmq that was already deprecated.

Switched to using the public DNS resolver 8.8.8.8 for the tests because the way to find the correct local resolver changed.

### DIFF
--- a/tests/letstest/scripts/boulder_config.sh
+++ b/tests/letstest/scripts/boulder_config.sh
@@ -1,32 +1,26 @@
 #!/bin/bash -x
 
 # Configures and Launches Boulder Server installed on
-# us-east-1 ami-5f490b35 bouldertestserver (boulder commit 8b433f54dab)
+# us-east-1 ami-072a9534772bec854 bouldertestserver3 (boulder commit b24fe7c3ea4)
 
 # fetch instance data from EC2 metadata service
 public_host=$(curl -s http://169.254.169.254/2014-11-05/meta-data/public-hostname)
 public_ip=$(curl -s http://169.254.169.254/2014-11-05/meta-data/public-ipv4)
 private_ip=$(curl -s http://169.254.169.254/2014-11-05/meta-data/local-ipv4)
 
-# get local DNS resolver for VPC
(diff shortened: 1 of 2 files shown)


tests/letstest/scripts/boulder_config.sh
```diff
@@ -1,32 +1,26 @@
...
-# start rabbitMQ
-#go run cmd/rabbitmq-setup/main.go -server amqp://localhost
-# start acme services
-#nohup ./start.py >& /dev/null < /dev/null &
-#./start.py
+sed -i '/httpPort/ s/5002/80/' ./test/config/va.json
+sed -i '/httpsPort/ s/5001/443/' ./test/config/va.json
+sed -i '/tlsPort/ s/5001/443/' ./test/config/va.json
+# set dns resolver
+sed -i 's/"127.0.0.1:8053",/"'$resolver'"/' ./test/config/ra.json
```
jsha: You don't really need these changes to the RA (see https://github.com/letsencrypt/boulder/pull/4185), but they're harmless.

bmw:
Which of the test farm tests that we run as part of the release did you get a chance to test the changes here (and in #7003) with?

ohemorange:
So far, with both PRs applied, tests pass for all images running `scripts/test_letsencrypt_auto_certonly_standalone.sh`.

tests/letstest/scripts/boulder_config.sh
bmw: If these changes aren't needed, I think we should remove them.

What do you think @ohemorange? If you're worried about potential problems, we could do this after the release tomorrow.

ohemorange:
(Just a note for posterity that when testing this, you'll want to terminate existing instances first to make sure the boulder server is recreated)

PR #7003, opened 2019-04-30
 Update Debian Jessie AMI to continue being able to use apt

### LINKED ISSUE #6907: Find new Debian Jessie AMI
Debian removed `jessie-updates` from its mirrors which is causing tests on the AMI at https://github.com/certbot/certbot/blob/3830c0f9005ba8c301a6e0bedc129a3a4be723a6/tests/letstest/targets.yaml#L36 to fail.

We could update our test scripts to fix up the apt config, but I think ideally we should just find a new Debian 8 AMI that has this problem resolved out of the box.

However we fix this, we should make sure the changes cause all tests we run as the part of the release process to pass on Debian 8 before merging.

ohemorange:
This branch is based on #7002, which should be merged first. Fixes #6907.

### DIFF
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -98,7 +98,7 @@ PROFILE = cl_args.aws_profile
 
 # Globals
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ohemorange:
All test farm tests pass on this image.

Issue #6996, opened 2019-04-27
`certbot-auto --version` prints unprintable ASCII character


morxa:
Running `certbot-auto --version` prints an unprintable ASCII character ` [?1034h`.

This leads to problems when piping that output into `mail`, because mailx interprets the input as binary data and thus sends emails that have no text body but only an attachment. You can test this with:
```
$ certbot-auto --version | mail -s "certbot-test" you@example.com
```

## My operating system is (include version):
CentOS 6.10

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto

## I ran this command and it produced this output:
```
$ certbot-auto --version | xxd
0000000: 6365 7274 626f 7420 302e 3333 2e31 201b  certbot 0.33.1 .
0000010: 5b3f 3130 3334 680a                      [?1034h.
```

## Certbot's behavior differed from what I expected because:
The second line should not be there, that's a non-printable character (escape sequence?). [This blog post](https://incenp.org/notes/2012/python-term-smm-fix.html) seems to provide some context and may hint to a fix.

morxa:
This issue does not occur on CentOS 7 with certbot installed from the epel repository:
```
$ certbot --version | xxd
certbot 0.29.1
```
After updating certbot:
```
$ certbot --version | xxd
certbot 0.31.0
```

schoen:
@bmw, do you know how this could happen? Could it be a vestige of our prior use of ncurses somehow?

schoen:
Or perhaps a behavior of some command-line program that we call in `certbot-auto` but not in `certbot` itself?

bmw:
Googling around, this seems to be a problem in older versions of the Python standard library. See https://bugs.python.org/issue19884.

I was able to reproduce this on CentOS 6 after installing Python 3.4 from EPEL and the `vim-common` package by running:
```
# TERM=xterm python3 -c 'import readline' | xxd
0000000: 1b5b 3f31 3033 3468                      .[?1034h
```

Unsetting `TERM` solves the problem. I could not reproduce this problem on CentOS 7 regardless of how Python (or Certbot) was installed.

schoen:
So, I imagine this might be difficult for us to work around in Certbot itself.

bmw:
I think that's right so I'm going to close this but please let me know if you think we should reconsider!

The only workaround I found was unsetting `TERM` (or maybe changing its value). Unfortunately, it appears the same problem is present in all of the versions of Python from SCL as well.

PR #7016, opened 2019-05-02
Stop certbot-auto from printing blank lines

### LINKED ISSUE #7012: certbot-auto produces a stray newline in the standard output
Running certbot-auto on Debian Jessie, since the automatic upgrade to 0.34.0 each calls to certbot-auto outputs a new line

```
/usr/local/sbin/certbot-auto -q renew | hexdump -C
00000000  0a                                                |.|
00000001
```

The problem is that this in turn generates an empty cronjob mail. This was not happening with previous versions.

Thanks

bmw:
Fixes #7012.

Apparently, the previous test we had here doesn't catch the case when certbot-auto prints blank lines. (I don't yet understand why so if someone does, please let me know!)

Regardless, I fixed up the test and verified it fails with the version of `letsencrypt-auto` in `master` and then fixed `letsencrypt-auto` so the test passes.

I ran test farm tests on the changes here and they passed on all instances.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -1593,12 +1593,14 @@ UNLIKELY_EOF
     # ---------------------------------------------------------------------------
     # If the script fails for some reason, don't break certbot-auto.
     set +e
-    # Suppress unexpected error output and only print the script's output if it
-    # ran successfully.
+    # Suppress unexpected error output.
     CHECK_PERM_OUT=$("$LE_PYTHON" "$TEMP_DIR/check_permissions.py" "$0" 2>/dev/null)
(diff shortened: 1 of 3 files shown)


jsoref:
Fwiw prior to this change, this was my output when run w/ an interactive shell:
```sh
/root/bin/certbot-auto --no-permissions-check -q renew --post-hook /root/bin/certbot-renew-hook </dev/null 2>&1 | hexdump -C
00000000  1b 5b 3f 31 30 33 34 68                           |.[?1034h|
00000008
```

I'm not sure why something decided to spit out a color code...

bmw:
Unfortunately, this is a known issue in old versions of Python. See https://github.com/certbot/certbot/issues/6996.

Issue #7015, opened 2019-05-02
Update INITIAL_VERSION in test_leauto_updates.sh


bmw:
[This test](https://github.com/certbot/certbot/blob/master/tests/letstest/scripts/test_letsencrypt_auto_certonly_standalone.sh) no longer works on Debian 9 due to issues running the old, [INITIAL_VERSION](https://github.com/certbot/certbot/blob/60e734c9690dc84bcea968d1d2041d17cbdeec27/tests/letstest/scripts/test_leauto_upgrades.sh#L19) of `certbot-auto` which is set up to then test the upgrade to the latest version.

We can fix the problem by updating `INITIAL_VERSION` to 0.33.1 which also allows us to test on Fedora 29+.

bmw:
This was fixed in #7017.

PR #7017, opened 2019-05-02
Fix test_leauto_upgrades on Debian 9 and Fedora 29+

bmw:
We made this change locally yesterday while preparing the release.

I tested this change on all AMIs currently in the test farm as well as Fedora 29 and this test passed on all instances.


bmw:
This fixes #7015.

Issue #6961, opened 2019-04-15
HTTP header Accept/ContentType


vproot:
Hi,
recently finished development of an Java based ACME server and found one obvious bug (even in the latest version) and one controversial “feature”.
The bug is located in certbot/acme/acme/client.py at line 497:
content_type = DER_CONTENT_TYPE # TODO: make it a param
The problem is that ContentType is bound to request and should be ‘application/jose+json’ and not ‘application/pkix-cert’ as defined at line 45 in the same file.
I understand that boulder may ignore ContentType, but correct server implementation should obey the value in this attribute. Accept header tells what client accepts as a response and server can ignore it (by ACME spec)
Controversial feature is decision to drop CN in CSR subject. Even though (in theory) everything but signature is optional in CSR and that Chrome and other browser ignores CN, use of the CN is security feature in enterprise environments. Since having CN in CSR Subject attribute doesn’t harm but widen possible acceptance of the ACME in enterprises, I really don’t see the problem having it back, something like:
csr.get_subject().CN = domains[0] (after the line 179 in certbot/acme/acme/crypto_util.py)

Just my five cents

bmw:
The CN issue has been addressed multiple times before (see https://github.com/certbot/certbot/issues/6463 for one example), but thanks for pointing out the problem with the content type.

We'd accept a PR to fix this.

bmw:
On closer inspection, this issue only affects our client for "ACMEv1" which only supports an unstandardized version of the protocol.

Now that there is a final, standardized version of ACME (see https://tools.ietf.org/html/rfc8555), we're planning on completely removing support for ACMEv1 in the near future (see https://github.com/certbot/certbot/issues/6844) so I'm closing this issue as wontfix.

I'd strongly encourage you to update your ACME server to be in line with the ACME spec.

PR #7018, opened 2019-05-02
Send HTTP Accept: JOSE

jsoref:
#6961

I know this is a bit more complicated than the simplest change. The previous code was structured in a way that seemed really hard to follow.

Note that I'm changing some places from using `cls` (which is sort of an arg) to a static mime type, that isn't necessary if things are really expecting to let the caller specify an alternate mime type, but if that's the case, then it's probably best not to claim that it's `cls.JSON_ERROR_CONTENT_TYPE` when in fact it's `application/vnd.someone.wonky+xml` because with a name like `JSON_ERROR_CONTENT_TYPE` or `JSON_CONTENT_TYPE` one would legitimately expect it to be what it says instead of some arbitrary user supplied mime type.

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -41,7 +41,12 @@ if sys.version_info < (2, 7, 9):  # pragma: no cover
 
 DEFAULT_NETWORK_TIMEOUT = 45
 
-DER_CONTENT_TYPE = 'application/pkix-cert'
+class ContentTypes(object):
+    """Mime Types"""
+    DER = 'application/pkix-cert'
+    JSON = 'application/json'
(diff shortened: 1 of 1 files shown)


bmw:
Thanks for the PR jsoref but I am going to close this. This arguably changes the API of this module, [ACMEv1 is being shut down by Let's Encrypt](https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430), and I want Certbot to be more aggressive about removing support for ACMEv1 ourselves. With this in mind, I'm personally not interested in making changes/improvements to our ACMEv1 support.

PR #7019, opened 2019-05-02
Fix oldest tests when local dependencies are used

### LINKED ISSUE #7014: Fix "oldest" tests on point release branches
In https://github.com/certbot/certbot/pull/6921, we made a change to properly update `local-oldest-requirements.txt` for point release branches, but it broke the tests. See https://travis-ci.com/certbot/certbot/jobs/197027330.

Until now, we've never had the case in the "oldest" tests where the pinned version like `certbot==0.34.0` matches the local files. Before either an older, non-local version was pinned or it was something like `-e acme` which instructs pip to install the local package into the virtual environment.

When we go to install the package in the oldest tests in the 0.34.x branch, we ask for `certbot==0.34.0`, which is found to be already available in the current working directory so pip does nothing.

Then when we change directories to run tests to prevent local files from impacting tests (which has been a problem in the past), the local package that was found before is no longer there, can't be imported, and the tests fail.

adferrand:
Fixes #7014.

Using a `--force-reinstall` (only for oldest tests), dependencies are properly reinstalled. Since this action significantly increases the execution time of oldest tests, I split them into two parts to allow their parallel execution by Travis.

We will need to find a better way to solve this in the future.

An example of successful execution of oldest tests in the situation of a point release can be found here: https://travis-ci.org/adferrand/certbot/builds/527475532

### DIFF
--- a/tools/pip_install.py
+++ b/tools/pip_install.py
@@ -99,8 +99,15 @@ def main(args):
         else:
             # Otherwise, we merge requirements to build the constraints and pin dependencies
             requirements = None
+            reinstall = False
             if os.environ.get('CERTBOT_OLDEST') == '1':
                 requirements = certbot_oldest_processing(tools_path, args, test_constraints)
+                # We need to --force-reinstall the tested distribution during oldest requirements
+                # because of an error in these tests in particular situations described in
+                # https://github.com/certbot/certbot/issues/7014
+                # However this slows down considerably the oldest tests (5 min -> 10 min),
+                # so we need to find a better mitigation in the future.
+                reinstall = True
             else:
                 certbot_normal_processing(tools_path, test_constraints)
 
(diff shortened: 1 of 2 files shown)


adferrand:
If we do that, the tests will be even more impacted in their execution time, because the two parts of pip install will reinstall all the dependencies for each dns plugin. Here only the second part, where the error about certbot dependence can occur, is impacted.

tools/pip_install.py
```diff
@@ -99,8 +99,15 @@ def main(args):
         else:
             # Otherwise, we merge requirements to build the constraints and pin dependencies
             requirements = None
+            reinstall = False
             if os.environ.get('CERTBOT_OLDEST') == '1':
                 requirements = certbot_oldest_processing(tools_path, args, test_constraints)
+                # We need to --force-reinstall the tested distribution during oldest requirements
```
bmw: Ah I see. In that case, can we just change this comment from "during oldest requirements" to "when using oldest requirements".

adferrand: Done.

PR #6372, opened 2018-09-13
3rd party plugins can be used with and without the pip package name

jfeil:
This is a fix for Issue #4351 
Furthermore the `certbot plugins` command is extended to display for 3rd party plugins the old name as alternative name.

PR #7026, opened 2019-05-03
Refer to thirdparty plugins by their short name

coldfix:
Hey,

this contains a patch for #4351. I saw that there is already an open PR #6372 addressing this issue, however noticed that it is incomplete in some regards so I decided to post this alternative PR.

Things to note:

- I believe the old form can be deprecated because it probably leaves everybody worse off and leads to some additional complexity. However, if you don't like this, I can remove the corresponding commit.
- I have not yet had the time to work on tests for the new behaviour. Can you give me a quick hint what would be needed?
- I found two additional locations that might need to be updated, see https://github.com/certbot/certbot/blob/master/certbot/plugins/selection.py#L324 (`if requested not in plugins` looks up by name), and https://github.com/certbot/certbot/blob/master/certbot/storage.py#L267 (`dest_namespace(plugin)` is based on its name, but I think it should be okay)

Best, Thomas


joohoi:
Thanks for the PR! We're currently in the middle of our packaging and distribution redesign, and this might prove very useful in order to help people to run third party plugins more easily.

coldfix:
Hi @adferrand, it might be better for me to rebase this on current master. What do you think?

adferrand:
I agree, it will be cleaner. Sorry for the two commits, I tried to update your PR, but did it from Windows, which completely screw up the symbolic links. Do not hesitate to remove them from the history of your branch. You can take the rebase as an occasion to move up your changelog entry to the latest master section.

coldfix:
Hi @adferrand, I have now rebased on master and fixed the broken test.

coldfix:
Closed in favor of #8131.

PR #7030, opened 2019-05-06
Fix check_permissions.py output and avoid writing script to root directory

AndiDog:
One of the two fixes was already merged into letsencrypt-auto. Since I don't know what happens with the different scripts (certbot-auto vs. letsencrypt-auto[-source]) and I have another fix, I'm posting this anyway so you can handle the patch as desired.

### DIFF
--- a/certbot-auto
+++ b/certbot-auto
@@ -1017,7 +1017,7 @@ if [ "$1" = "--le-auto-phase2" ]; then
     fi
 
     say "Installing Python packages..."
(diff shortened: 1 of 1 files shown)


bmw:
Thanks for the PR @AndiDog.

We landed an equivalent fix in https://github.com/certbot/certbot/pull/7034 and `certbot-auto` in the root of the repo is only updated at release time. We're planning on doing a release tomorrow morning including the changes in #7034.

PR #7037, opened 2019-05-06
Fix test_leauto_upgrades.sh on CentOS 6.

bmw:
I chose this approach because "rebootstrapping" is one of the more complicated pieces of certbot-auto and this test makes sure that things work properly.

The main value in the "Python 3 tests" is not in the initial `if [ "$RUN_PYTHON3_TESTS" = 1 ]; then` branch which is mainly testing that the initial installation worked as expected. The value is in the second branch like this which is testing that we properly rebootstrapped using the latest version of the script.

Alternatively, we could move these tests to another script, but I don't think making a 6th script to run at release time is worth it.

### DIFF
--- a/tests/letstest/scripts/test_leauto_upgrades.sh
+++ b/tests/letstest/scripts/test_leauto_upgrades.sh
@@ -15,8 +15,17 @@ if ! command -v git ; then
         exit 1
     fi
 fi
-# 0.33.x is the oldest version of letsencrypt-auto that works on Fedora 29+.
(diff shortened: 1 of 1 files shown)


bmw:
I actually tested that these changes work in the test farm this time. I have logs to prove it :)

bmw:
Yeah :(

We're actively working on redoing both packaging and these tests so hopefully stuff like this will go away soon. If you have other ideas for things you think we should do in the meantime, please let me know!

PR #7052, opened 2019-05-11
dns-cloudflare: update URL for obtaining API keys

judge2020:
Updated the ACCOUNT_URL in the Cloudflare-DNS plugin.
This uses the new "dash.cloudflare.com" scheme and future-proofs this URL for an upcoming change to Cloudflare API keys (this is not public yet, so no other changes related to this).

### DIFF
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/dns_cloudflare.py
@@ -10,7 +10,7 @@ from certbot.plugins import dns_common
 
 logger = logging.getLogger(__name__)
 
-ACCOUNT_URL = 'https://www.cloudflare.com/a/account/my-account'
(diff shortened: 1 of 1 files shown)


judge2020:
The old link is to `www.cloudflare.com/a`, which is the URL of the old dashboard. 

The current "log in" link on https://www.cloudflare.com goes to `dash.cloudflare.com/login`, which is a link to the new dashboard (introduced ~May 2018).

I can't give documentation for the "api-tokens" part of the new URL because that change is currently in closed beta, but going to the URL currently on a non-privileged account will correctly show the "global API key" page.

schoen:
Thanks! Can anyone help to confirm that this change will still work properly for those who aren't part of the beta?

alexzorin:
@schoen Yes, for us regular users, it seems like `/profile/<anything>` falls back to the normal profile page where the Global API Key is shown.

PR #7054, opened 2019-05-12
Validate OCSP responses in case an explicit responder is designated

### LINKED ISSUE #7053: Certbot can't verify OCSP responses signed by an explicitly designated responder
If you're having trouble using Certbot and aren't sure you've found a bug or
request for a new feature, please first try asking for help at
https://community.letsencrypt.org/. There is a much larger community there of
people familiar with the project who will be able to more quickly answer your
questions.

Debian stretch/sid

certbot-auto

./certbot-auto certificates --cert-name buypasstest1

Invalid signature on OCSP response for /etc/letsencrypt/live/buypasstest1/cert.pem

The OCSP response is correctly signed, not by the issuer itself but by a properly designated responder.

2019-05-12 13:32:13,645:DEBUG:certbot.main:certbot version: 0.34.2
2019-05-12 13:32:13,647:DEBUG:certbot.main:Arguments: ['--cert-name', 'buypasstest1']
2019-05-12 13:32:13,647:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#manual,PluginEntryPoint#nginx,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
2019-05-12 13:32:13,719:DEBUG:certbot.log:Root logging level set at 20
2019-05-12 13:32:13,721:INFO:certbot.log:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2019-05-12 13:32:13,770:DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): ocsp.buypass.com:80
2019-05-12 13:32:13,994:DEBUG:urllib3.connectionpool:http://ocsp.buypass.com:80 "POST / HTTP/1.1" 200 1702
2019-05-12 13:32:13,998:ERROR:certbot.ocsp:Invalid signature on OCSP response for /etc/letsencrypt/live/buypasstest1/cert.pem

<doesn't matter.>

adferrand:
Fixes #7053 

This PR implements the second case of OCSP response signee as described in RFC6960 - 4.2.2.2](https://tools.ietf.org/html/rfc6960#section-4.2.2.2), that is when an explicit authorized responder, different from the issuer, is designated to sign OCSP responses for a given certificate.

Let'sEncrypt CA is not using it (issuer and responder are the same), but other CA like BuyPass separates the two concepts,  and so require the full RFC to be implemented.

This PR has been tested successfully in both cases:
* a certificate generated with Let'sEncrypt CA
* a certificate generated with BuyPass Go SSL


eabalea:
That should work most of the time.
Will fail if the issuer uses another key to sign the response (different scenarios with self-issued certs, or key rollover), but those aren't common in the wild, haven't found any such CA implementing ACME.

adferrand:
Yes in fact, it is stated in the RFC that a client that uses OCSP must recognize the delegation certificate only if it is issued with the same key than the certificate to check. It is not prohibited server side to use different keys for retro-compatibility purpose, but it is strongly discouraged.

certbot/ocsp.py
```diff
@@ -207,11 +207,43 @@ def _check_ocsp_response(response_ocsp, request_ocsp, issuer_cert):
...
+
+        responder_cert = responder_certs[0]
+
+        if responder_cert.issuer != issuer_cert.subject:
+            raise AssertionError('responder certificate is not signed '
+                                 'by the certificate\'s issuer')
+
+        try:
+            extension = responder_cert.extensions.get_extension_for_class(x509.ExtendedKeyUsage)
+            authorized = x509.oid.ExtendedKeyUsageOID.OCSP_SIGNING in extension.value
```
joohoi: nit: I was confused about this variable name for a while. Maybe `delegated` would fit better :)

certbot/ocsp.py
```diff
@@ -207,11 +207,43 @@ def _check_ocsp_response(response_ocsp, request_ocsp, issuer_cert):
...
+        except (x509.ExtensionNotFound, IndexError):
+            authorized = False
+        if not authorized:
+            raise AssertionError('responder is not authorized by issuer to sign OCSP responses')
+
+        # Following line may raise UnsupportedAlgorithm
```
joohoi: nit: This and the following signature check could use a comment line to make it more clear and easier to read what we're checking in both of the cases. Makes the code a whole lot more readable.

certbot/tests/ocsp_test.py
```diff
@@ -150,6 +150,38 @@ def test_revoke(self, mock_ocsp_response, mock_post):
...
+        # On first call, responder and issuer are not the same. Two signatures will be checked
+        # then, first to verify the responder cert (using the issuer public key), second to
+        # to verify the OCSP response itself (using the responder public key).
+        self.assertEqual(mock_check.call_count, 2)
+        self.assertEqual(mock_check.call_args_list[0][0][0].public_numbers(),
+                         issuer.public_key().public_numbers())
+        self.assertEqual(mock_check.call_args_list[1][0][0].public_numbers(),
+                         responder.public_key().public_numbers())
+
+        response.responder_name = issuer.subject
```
joohoi: Although these tests require a bit of boilerplate for setup, it might be beneficial to split the test cases up to different tests while moving the setup to a separate function. (also see the comment about coverage).

certbot/ocsp.py
```diff
@@ -207,11 +207,43 @@ def _check_ocsp_response(response_ocsp, request_ocsp, issuer_cert):
...
+    else:
+        logger.debug('OCSP response for certificate %s is delegated to an external responder.',
+                     cert_path)
+
+        responder_certs = [cert for cert in response_ocsp.certificates
+                           if cert.subject == response_ocsp.responder_name]
+        if not responder_certs:
+            raise AssertionError('no matching responder certificate could be found')
+
+        responder_cert = responder_certs[0]
```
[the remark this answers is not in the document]
adferrand: I added a comment about this.

certbot/ocsp.py
adferrand: I merge the two, because it is not only that it is a delegate, it needs also to be explicitly authorized. So `delegate_authorized`.

certbot/ocsp.py
adferrand: Added the comments here.

certbot/tests/ocsp_test.py
adferrand: Done.

adferrand:
@joohoi, I added the unit tests you asked. I took the occasion to abstract more setup logic to make the tests more readable.

certbot/tests/ocsp_test.py
```diff
@@ -139,116 +140,131 @@ def test_ensure_cryptography_toggled(self, mock_revoke, mock_determine):
...
+            mocks['mock_response'].return_value.certificates[0] = mock.Mock(
+                issuer='fake', subject=cert.subject)
+            revoked = self.checker.ocsp_revoked(self.cert_path, self.chain_path)
+        self.assertFalse(revoked)
 
+        with _ocsp_mock(ocsp_lib.OCSPCertStatus.REVOKED, ocsp_lib.OCSPResponseStatus.SUCCESSFUL):
+            # This mock is necessary to avoid the first call contained in _determine_ocsp_server
+            # of the method cryptography.x509.Extensions.get_extension_for_class.
+            with mock.patch('certbot.ocsp._determine_ocsp_server') as mock_server:
+                mock_server.return_value = ('https://i.do.not.care', 'i.do.not.care')
```
joohoi: Even though these don't do anything. I think we should stick with `example.com` domains in here to avoid misunderstandings and such (`not.care` is an actual, registered domain name).

adferrand: My creativity is frustrated! But done, with the usual dns example... :)

joohoi:
LGTM!

Issue #6748, opened 2019-02-07
Test farm tests can silently drop test instances


bmw:
The tests in https://github.com/certbot/certbot/tree/master/tests/letstest can silently drop test instances causing neither success or failure to be reported from them with no mention of the problem of the dropped instance.

This is really bad. The error message that seems to occur when this happens is `ResponseParserError: Unable to parse response (no element found: line 1, column 0), invalid XML received:`. Maybe we can just fix that problem.

Alternatively or in addition to that, the tests create a file at the end called `results` containing the result from each test instance. We can assert that the number of instances we're reporting is equal to the number of instances we started (not counting boulder) and if they're not equal, error out.

ohemorange:
If for some reason we do decide to fix this instead of deprecating the entire framework, this may be relevant: https://github.com/ceph/ceph/pull/13461/files

PR #7057, opened 2019-05-13
Stop sharing state between processes in test farm tests

bmw:
### Problem

In looking into putting our current test farm tests in Travis, I took a quick look at trying to resolve the error briefly mentioned at https://github.com/certbot/certbot/issues/6748 where you'd get a traceback like:
```
  File "multitester.py", line 232, in block_until_instance_ready
    _state = _instance.state['Name']
  File "/path/to/my/venv/lib/python2.7/site-packages/boto3/resources/factory.py", line 339, in property_loader
    self.load()
  File "/path/to/my/venv/lib/python2.7/site-packages/boto3/resources/factory.py", line 505, in do_action
    response = action(self, *args, **kwargs)
  File "/path/to/my/venv/lib/python2.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/client.py", line 648, in _make_api_call
    operation_model, request_dict, request_context)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/client.py", line 667, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/endpoint.py", line 135, in _send_request
    request, operation_model, context)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/endpoint.py", line 167, in _get_response
    request, operation_model)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/endpoint.py", line 218, in _do_get_response
    response_dict, operation_model.output_shape)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/parsers.py", line 242, in parse
    parsed = self._do_parse(response, shape)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/parsers.py", line 501, in _do_parse
    root = self._parse_xml_string_to_dom(xml_contents)
  File "/path/to/my/venv/lib/python2.7/site-packages/botocore/parsers.py", line 437, in _parse_xml_string_to_dom
    "invalid XML received:\n%s" % (e, xml_string))
ResponseParserError: Unable to parse response (no element found: line 1, column 0), invalid XML received:
```
when running test farm tests. Resolving this would be nice for stability because our tests currently spuriously fail with this problem.

### Solution

I think the problem is the shared EC2 state between processes. See https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing. I've run the test farm tests over ~~15~~ 20 times with these changes and I haven't seen this problem once.

### Notes for reviewer

Unfortunately, the diff here is pretty awful to look at, but the changes being made are hopefully straightforward.

It might be easiest to look at all commits but the last one and then notice that the last commit just:

1. Deletes the `SCRIPT BEGINS` comments.
2. Puts all code after that in a `main` function with no other modifications.
3. Adds the standard `if __name__ == '__main__':` to the bottom of the file.

Also, I'm planning on opening a few PRs like this and creating a single PR/branch that incorporates all of the changes here with the tests running in Travis. Feel free to wait to review this until it's testable in Travis if you like.


bmw:
Unfortunately, there is still a lot of other global state in this script such as the results of command line parsing, but all global AWS state has been removed and is no longer passed between processes.

tests/letstest/multitester.py
```diff
@@ -341,22 +338,28 @@ def create_client_instances(targetlist, security_group_id, subnet_id):
...
     cur_proc = mp.current_process()
     for inreq in iter(inqueue.get, SENTINEL):
-        ii, target = inreq
+        ii, instance_id, target = inreq
+
+        # Each client process is given its own session due to the suggestion at
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing.
+        aws_session = boto3.session.Session(profile_name=PROFILE)
+        ec2_client = aws_session.resource('ec2')
+        instance = ec2_client.Instance(id=instance_id)
```
adferrand: So if I understood well the logic here, these three lines are the core modification of this PR, and everything else is either to make theses lines possibles, or to adapt the logic from theses lines.

Am I right?

tests/letstest/multitester.py
```diff
@@ -99,11 +99,9 @@
 # Globals
 #-------------------------------------------------------------------------------
 BOULDER_AMI = 'ami-072a9534772bec854' # premade shared boulder AMI 18.04LTS us-east-1
-LOGDIR = "" #points to logging / working directory
-# boto3/AWS api globals
-AWS_SESSION = None
-EC2 = None
+LOGDIR = "letest-%d"%int(time.time()) #points to logging / working directory
```
adferrand: Here so we have a dedicated LOGDIR for each execution of multitester, am I right ?

Since the name is non-predictable anymore, could we print at the beginning of the multitester execution the LOGDIR value, to help find the appropriate folder for this given execution during troubleshooting?

I have no particular opinion on the implementation, but I would personally replace `LOGDIR` by a function that generate the name and print its value, and then this function would be call appropriately at the beginning of multitester execution.

tests/letstest/multitester.py
```diff
@@ -99,11 +99,9 @@
...
 SECURITY_GROUP_NAME = 'certbot-security-group'
+SENTINEL = None #queue kill signal
```
adferrand: I see no code here that change `SENTINEL` value to something else than `None`. Is it really used ?

In parallel, setting a global variable to `None` here indicate a mutable global state. I think it can be improved, by creating it lazily during the multitester execution.

tests/letstest/multitester.py
```diff
@@ -225,20 +224,17 @@ def block_until_ssh_open(ipstring, wait_time=10, timeout=120):
...
-    while _state != 'running' or _ip is None:
+    state = booting_instance.state['Name']
+    ip = booting_instance.public_ip_address
+    while state != 'running' or ip is None:
         time.sleep(wait_time)
-        _instance = EC2.Instance(id=_id)
-        _state = _instance.state['Name']
-        _ip = _instance.public_ip_address
-    block_until_ssh_open(_ip)
+        # The instance needs to be reloaded to update its local attributes.
```
adferrand: Could we add here a link to the documentation addressing this problem?

tests/letstest/multitester.py
bmw: Pretty much. A higher level summary of the changes in this PR which I probably should have given before is I created a new AWS session and EC2 resource in each client process, don't pass any `boto3` objects through the shared queues, and removed the global `boto3` state by putting it in a `main()` function. This last change caused a lot of other changes to have to be made.

tests/letstest/multitester.py
bmw: > Here so we have a dedicated LOGDIR for each execution of multitester, am I right ?

Yep!

> Since the name is non-predictable anymore, could we print at the beginning of the multitester execution the LOGDIR value, to help find the appropriate folder for this given execution during troubleshooting?

If I understand you correctly, this is done at https://github.com/certbot/certbot/pull/7057/files#diff-821c98fdcbcf065345fc89980bc6609eR410 which is the first use of `LOGDIR` when `main` is called.

tests/letstest/multitester.py
bmw: You're right the value of `SENTINEL` is never changed.

`SENTINEL` is used to tell client processes that their work is done. See https://github.com/certbot/certbot/pull/7057/files#diff-821c98fdcbcf065345fc89980bc6609eR343.

The way this works is the parent process loads the queue with a bunch of client instances that need tests run on them at https://github.com/certbot/certbot/pull/7057/files#diff-821c98fdcbcf065345fc89980bc6609eR542. The child processes work through this list until it reaches the `SENTINEL` value at which point they exit.

`SENTINEL` is a global constant for this purpose. For what it's worth, this aspect never changed. `SENTINEL` is currently a global constant that works that way in `master`, I just moved setting it's value to the top of the script so it is outside of the `main()` function.

tests/letstest/multitester.py
adferrand: Yes indeed, I missed it!

tests/letstest/multitester.py
adferrand: OK I see! So it is irrelevant for this PR, as it does not change a logic that was working the similarly before.

tests/letstest/multitester.py
bmw: No worries! It's pretty hard to see because the diff shown here is so bad.

bmw:
If you'd like to sanity check that merge commit, I recommend comparing this branch to `test-letstest-all-changes`.

adferrand:
The only thing you added here is a comment, and the merged logic is also the one done on `test-letstest-all-changes`.

We will make a final check on the final PR that adds the tox target, when the other ones are merged.

adferrand:
Oups, conflicts now ^^

PR #7061, opened 2019-05-14
Add test farm tests requirements file

bmw:
This is another piece of getting test farm tests set up in Travis. I've ran the full suite of test farm tests with these pinned packages and they passed.


adferrand:
LGTM !

PR #7065, opened 2019-05-14
Exit with a nonzero status when test farm tests fail

bmw:
Yet another small piece of prepping the test farm tests to run in Travis.

### DIFF
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -106,6 +106,11 @@ EC2 = None
 SECURITY_GROUP_NAME = 'certbot-security-group'
 SUBNET_NAME = 'certbot-subnet'
 
+class Status(object):
+    """Possible statuses of client tests."""
+    PASS = 'pass'
+    FAIL = 'fail'
+
 # Boto3/AWS automation functions
 #-------------------------------------------------------------------------------
 def should_use_subnet(subnet):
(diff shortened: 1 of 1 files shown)


tests/letstest/multitester.py
```diff
@@ -106,6 +106,11 @@
 SECURITY_GROUP_NAME = 'certbot-security-group'
 SUBNET_NAME = 'certbot-subnet'
 
+class Status(object):
```
adferrand: Why not just using a boolean ? `True` -> PASSED, `False` -> FAILED

bmw: We could do that, but these values currently in `master` aren't actually used for signaling pass or failure to the logic to the script, but are used as output to the terminal and logfiles so they are strings.

This PR adds a second use for these values of knowing whether the tests passed or failed internally so if we wanted I could instead:

1. Convert the current strings to booleans.
2. In the main function, convert them back to the strings we want for output.

I think this results in more code changes without a real significant benefit, so I'd prefer to keep things as is, but if you still feel the same way with this additional explanation I can make the change.

(The actual internal values used don't really matter in my opinion, they just need to be some consistent value that can be used as a point of comparison. I suppose there is some additional memory and CPU overhead from using strings vs. booleans, but I don't think this matters in this case.)

adferrand: OK, if it is used also as an output, then it makes sense that they are strings.

PR #7070, opened 2019-05-16
Fix cleanup on failure.

bmw:
An annoying problem in test farm tests is if a problem occurs while starting instances, you get a traceback like:
```
  File "multitester.py", line 591, in <module>
    main()
  File "multitester.py", line 584, in main
    cleanup(cl_args, instances, targetlist)
UnboundLocalError: local variable 'instances' referenced before assignment
```
This can be seen at https://travis-ci.com/certbot/certbot/builds/112000424.

This fixes that problem.

### DIFF
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -314,31 +314,27 @@ def grab_certbot_log():
     sudo('if [ -f ./certbot.log ]; then \
     cat ./certbot.log; else echo "[nolocallog]"; fi')
 
-def create_client_instances(targetlist, security_group_id, subnet_id):
-    "Create a fleet of client instances"
-    instances = []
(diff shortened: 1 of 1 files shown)


tests/letstest/multitester.py
```diff
@@ -488,9 +487,13 @@ class FabricException(Exception):
                                    security_group_id=security_group_id,
                                    subnet_id=subnet_id)
 
+instances = []
```
adferrand: Since `instance` becomes a global state, can we name it using uppercase letters, and define it at the top of the script?

adferrand: Cancel my comment above, #7057 will move all this logic in a `main` method.

Issue #7071, opened 2019-05-16
Dump failed test farm tests logs in Azure


bmw:
After all of my current PRs are merged to get test farm tests running in Travis, one remaining pain point in my opinion is the fact that when tests fail on an instance, the only relevant output will be something like:
```
1 ubuntu18.04LTS fail
```
This will tell us that a test is failing, but we'll have no idea why until we run the tests again locally.

All output from the test instances are saved in log files, but these aren't included in the Travis output. I think they probably should be in a way that ensures the logfile contents aren't overlapping in Travis' output.

adferrand:
Still relevant.

PR #7072, opened 2019-05-16
Enable test farm tests in Travis

bmw:
You can see the full test suite running at https://travis-ci.com/certbot/certbot/builds/112291892.

A few noteworthy things:

* `--fast` is included because without, the tests would sometimes reach Travis' 50 minute timeout even with 1 test script per Travis build.
*  The only script that is run at release time which is not being run here is https://github.com/certbot/certbot/blob/master/tests/letstest/scripts/test_tests.sh because that script runs tests on the packages installed by `certbot-auto` which won't be updated until midway through a release.
* We check `TRAVIS_PULL_REQUEST` and error out if it is not false for simplicity which should be fine because these tests are never run on PRs. The reason it's more complex to run test farm tests on PRs is the test farm tests need a named branch to pull from and Travis effectively merges the PR into the target branch before running tests complicating this.

I don't think this should block this PRs, but the one final change we may want to make to the current setup is https://github.com/certbot/certbot/issues/7071.


bmw:
@adferrand, if you have the time and interest, I think you can start reviewing these PRs. If I were you, I'd read the above description first and then start working through my PRs chronologically.

tox.ini
```diff
@@ -274,3 +274,34 @@ commands =
...
+commands =
+    {[testenv:travis-test-farm-tests-base]commands}
+    python multitester.py apache2_targets.yaml travis-test-farm.pem none scripts/test_apache2.sh --fast
+    python multitester.py targets.yaml travis-test-farm.pem none scripts/test_leauto_upgrades.sh --fast
+deps = {[testenv:travis-test-farm-tests-base]deps}
+passenv = {[testenv:travis-test-farm-tests-base]passenv}
+setenv = {[testenv:travis-test-farm-tests-base]setenv}
+whitelist_externals = {[testenv:travis-test-farm-tests-base]whitelist_externals}
+
+[testenv:travis-test-farm-tests-part2]
```
adferrand: The part2 execution time is still extremely high, 48 minutes in the last run. We are not very far from the hard limit of 1 hour, when Travis just kill the entire job, and so potentially leaving here I suppose some AWS instances uncleaned.

Can we either:
1) Put some of the work here to part 1
2) Split part 2 again to have globally three parts ?

I think, for the global run efficiency and to avoid potential hard kill as I said just before, that all theses runs should have an execution time below 30 minutes.

bmw: Unfortunately the times of these tests in general vary wildly. For instance, here's another run where part 2 is fast and part 1 is dangerously slow: https://travis-ci.com/certbot/certbot/builds/112169625

I haven't seen the builds time out with this configuration yet though. I've ran them a half a dozen times or so.

What I propose we do is do some more testing (which can happen either before or after this is merged) and if we (regularly?) timeout split the 4 tests into 4 individual jobs and deal with the consequences of doing that which I described above.

bmw:
I agree with the comment at https://github.com/certbot/certbot/pull/7057#issuecomment-493540772 that we should do another test of this PR before merging it once all other test farm tests PRs have been merged.

tox.ini
adferrand: What you describe about the big changes observed in execution time makes me think that in fact, the two parts are in a race condition to get AWS instances. I would bet that in fact one of the part is blocking until it gets the AWS instance because the other one is already holding all the nodes that can be run in parallel.

Could you move one part at the beginning of the Travis job list and the other one at the end? This way we will would be sure that the two parts are not running in parallel.

bmw:
Full tests are running again at https://travis-ci.com/certbot/certbot/builds/112272810.

tox.ini
bmw: Hm. If we think that's the problem, I'd be tempted to put them all in one Travis build. That build may take ~30 minutes, but it should minimize that problem.

I'm not entirely confident that's what's going on though. Here's a build where there were 4 tests running in parallel and they all were fast: https://travis-ci.com/certbot/certbot/builds/112154559

adferrand: Do you want to avoid having 4 jobs in parallel? Because it seems to be very efficient and keep, when every thing is good, the execution time very low. If in bad situations it takes 4 or 5 times more, we are still good.

bmw: I did. My explanation was:

> The reason I put 2 scripts in the same build is there is a limit on the number of concurrent EC2 instances you can have. Our limit is 87 so to avoid problems due to two full test suites running at once and/or someone trying to do manual testing, I put them in two builds.

Looking at this problem more closely, this seems unlikely to be a problem though (as long as we do a good job making sure instances are properly cleaned up in the case of an error while running these tests such as them timing out). It seems unlikely that two nightly builds run at once fast enough to allow the test farm tests in different Travis builds to overlap. When running the client tests locally, running the tests in parallel on the same machine significantly slows down tests so Erica and I usually only run at most two tests at once.

Looking at those numbers, currently the 4 tests being run in parallel is 42 test instances plus 1 boulder instance which is shared. We have issues in the current milestone to potentially add 1 or 2 more instances to each run which could increase this to 50 client instances with the non-Apache test farm tests using 13 client instances. So even if we were running two more of these tests locally, that's only 50 + 13 * 2 + 1 = 77 instances so we should be fine.

bmw:
Full tests are running again again at https://travis-ci.com/certbot/certbot/builds/112281301.

One other change I think we should make to these tests in either this or another PR is these tests are currently only testing the `master` branch. Instead, I think I should add `--repo --branch` flags to the script to have it always test the local files. This is good for things like point releases.

adferrand:
Definitely! I think this PR is the good place since it is already adding and configuring the tox targets.

bmw:
This is done and I updated my original post to reflect the current state of things.

PR #7073, opened 2019-05-16
Add executable scripts to start certbot and acme server in certbot-ci

adferrand:
During review of #6989, we saw that some of our test bash scripts were still used in the Boulder project in particular. It is about `tests/integration/_common.sh` in particular, to expose the `certbot_test` bash function,  that is an appropriate way to execute a local version of certbot in test mode: define a custom server, remove several checks, full log and so on.

This PR is an attempt to assert this goal: exposing a new `certbot_test` executable for test purpose. More generally, this PR is about giving well suited scripts to quickly make manual tests against certbot without launching the full automated pytest suite.

The idea here is to leverage the existing logic in certbot-ci, and expose it as executable scripts. This is done thanks to the `console_scripts` entry of setuptools entrypoint feature, that install scripts in the `PATH`, when `pip install` is invoked, that delegate to specific functions in the installed packages.

Two scripts are defined this way:
* `certbot_test`: it executes certbot in test mode in a very similar way than the original `certbot_test` in `_common.sh`, by delegating to `certbot_integration_tests.utils.certbot_call:main`. By default this execution will target a pebble directory url started locally. The url, and also http-01/tls-alpn-01 challenge ports can be configured using ad-hoc environment variables. All arguments passed to `certbot_test` are transferred to the underlying certbot command.
* `acme_server`: it set up a fully running instance of an ACME server, ready for tests (in particular, all FQDN resolves to localhost in order to target a locally running `certbot_test` command) by delegating to `certbot_integration_tests.utils.acme_server:main`. The choice of the ACME server is given by the first parameter passed to `acme_server`, it can be `pebble`, `boulder-v1` or `boulder-v2`. The command keeps running on foreground, displaying the logs of the ACME server on stdout/stderr. The server is shut down and resources cleaned upon entering CTRL+C.

This two commands can be run also through the underlying python modules, that are executable.

Finally, a typical workflow on certbot side to run manual tests would be:
```
cd certbot
tools/venv.py
source venv/bin/activate
acme_server pebble &
certbot_test certonly --standalone -d test.example.com
```

On boulder side it could be:
```
# Follow certbot dev environment setup instructions, then ...
cd boulder
docker-compose run --use-aliases -e FAKE_DNS=172.17.0.1 --service-ports boulder ./start.py
SERVER=http://localhost:4001/directory certbot_test certonly --standalone -d test.example.com
```

cc @bmw @cpu


cpu:
> On boulder side it could be:
>
> # Follow certbot dev environment setup instructions, then ...
> cd boulder
> docker-compose run --use-aliases -e FAKE_DNS=172.17.0.1 --service-ports boulder ./start.py
> SERVER=http://localhost:4001/directory certbot_test certonly --standalone -d test.example.com

That's great :+1: Thanks @adferrand.

certbot-ci/certbot_integration_tests/certbot_tests/context.py
```diff
@@ -60,71 +53,24 @@ def cleanup(self):
...
+        Base command to execute certbot in a distributed integration test context.
         """
-        command = ['--renew-by-default']
-        command.extend(args)
-        return self._common_test_no_force_renew(command)
+        return certbot_call.certbot_test(
+            args, self.directory_url, self.http_01_port, self.tls_alpn_01_port,
+            self.config_dir, self.workspace, force_renew, True)
 
-    def certbot_no_force_renew(self, args):
```
bmw: optional nit: Because calls like:
```
context.certbot(['renew', '--deploy-hook', 'echo deploy >> "{0}"'.format(context.hook_probe)],
                            False)
```
look pretty mysterious when `force_renew` isn't set as a keyword argument, I think I prefer keeping `certbot_no_force_renew` and `certbot` around. You can make the call on that though.

certbot-ci/certbot_integration_tests/nginx_tests/context.py
```diff
@@ -29,15 +29,16 @@ def cleanup(self):
         self._stop_nginx()
         super(IntegrationTestsContext, self).cleanup()
 
-    def certbot_test_nginx(self, args):
+    def certbot_test_nginx(self, args, force_renew=True):
```
bmw: Same comment about `force_renew` being set this way as above.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,73 @@
...
+                           cwd=workspace, env=new_environ)
+
+
+def main():
+    args = sys.argv[1:]
+
+    # Default config is pebble
+    directory_url = os.environ.get('SERVER', 'https://localhost:14000/dir')
+    http_01_port = int(os.environ.get('HTTP_01_PORT', '5002'))
+    tls_alpn_01_port = int(os.environ.get('TLS_ALPN_01_PORT', '5001'))
```
bmw: nit: Can we reuse these URL and port values in a global so if they ever change, we're not making modifications to multiple places?

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Execution of certbot in a self-contained workspace
+    workspace = os.environ.get('WORKSPACE', os.path.join(os.getcwd(), '.certbot_test_workspace'))
+    if not os.path.exists(workspace):
+        print('--> Create a workspace for certbot_test: {0}'.format(workspace))
```
bmw:
```suggestion
        print('--> Creating a workspace for certbot_test: {0}'.format(workspace))
```

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,73 @@
...
+        os.mkdir(workspace)
+    else:
+        print('--> Use an existing workspace for certbot_test: {0}'.format(workspace))
```
bmw:
```suggestion
        print('--> Using an existing workspace for certbot_test: {0}'.format(workspace))
```

certbot-ci/setup.py
```diff
@@ -45,4 +45,11 @@
     packages=find_packages(),
     include_package_data=True,
     install_requires=install_requires,
+
+    entry_points={
+        'console_scripts': [
+            'certbot_test=certbot_integration_tests.utils.certbot_call:main',
+            'acme_server=certbot_integration_tests.utils.acme_server:main',
```
bmw: nit: This is messing with my muscle memory. I regularly type `venv/bin/a<tab>` but now my shell sees both `acme_server` and `activate`. What about changing this to something like `start_acme_server` or `run_acme_server`?

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -20,27 +20,31 @@
 HTTP_01_PORT = 5002
 
 
-def setup_acme_server(acme_server, nodes):
+def setup_acme_server(acme_server, nodes, proxy=True):
```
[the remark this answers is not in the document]
adferrand: Yes it would continue to work. But it is very limited, since the teardown occurs only when the main python process finishes. I have some use cases in mind that could require the teardown to occur before. 

Here for now, the teardown is connected to the pytest general teardown, so calling programmatically pytest (as described in the official documentation https://docs.pytest.org/en/latest/usage.html#calling-pytest-from-python-code) will work as expected, with a teardown occurring at the end of the pytest invocation, not at the end of the python script.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -195,3 +196,36 @@ def _launch_command(command, cwd=os.getcwd()):
...
+    process = None
+
+    try:
+        start()
+        print('--> Instance of {0} is running, directory URL is {0}'
+              .format(acme_xdist['directory_url']))
+        print('--> Press CTRL+C to stop the ACME server.')
+
+        docker_name = 'pebble_pebble_1' if 'pebble' in server_type else 'boulder_boulder_1'
+        process = subprocess.Popen(['docker', 'logs', '-f', docker_name])
```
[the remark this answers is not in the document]
adferrand: Both approaches are very useful. Getting used to the Docker way, I appreciate a lot processes than can be run in foreground. What about proposing a flag to run either in foreground or in background?

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
bmw: > Yes it would continue to work. But it is very limited, since the teardown occurs only when the main python process finishes. I have some use cases in mind that could require the teardown to occur before.

What's the use case you had in mind? If it requires further changes to our code, what do you think about waiting to make this change until that use case comes up? I just think sticking with `atexit` is cleaner and simpler given the current needs I'm aware of.

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
bmw: I'd personally rather us just implement one way to do this now and implement another behind a flag if the demand/need really arises. I like having it run the server in the background because I think it's more flexible (you can run `docker logs` yourself if you want) and it's been years since I had to look at server logs at all. Why do you like running it in the foreground?

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
adferrand: This https://github.com/certbot/certbot/pull/7073#discussion_r289206742 for instance ^^

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
adferrand: Well, the last several times I needed to code on acme, it was very useful to have the log in parallel, and see how the server reacted to the modifications.

In general, I prefer to run in foreground because it is quite useful to get a blocking point in a script, as I said for Docker a process needs to be in foreground, and because in shell, it is easier to transform a foreground process into background one than the contrary.

For the specific case of Boulder and Pebble in fact, there is more than on Docker that is started, and also a dedicated network. Shutting down for Boulder requires at least 4 Docker commands. It is easier and less error prone to see it properly cleaned up upon SIGTERM on the foreground command, than to run the Docker command or have a separate dedicated command provided by certbot-ci to do it.

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
bmw: Right, but you're pushing back on making that change and like I said, I'd rather not add options unless they're really needed.

Any other real, current use cases you have in mind?

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
bmw: > In general, I prefer to run in foreground because it is quite useful to get a blocking point in a script, as I said for Docker a process needs to be in foreground, and because in shell, it is easier to transform a foreground process into background one than the contrary.

I buy this argument. Let's stick with it in the foreground.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -195,3 +196,36 @@ def _launch_command(command, cwd=os.getcwd()):
...
+
+        while True:
+            time.sleep(1)
```
bmw: I think we should dramatically increase this value. ctrl+c will interrupt the sleep and a higher value causes less busy waiting.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -195,3 +196,36 @@ def _launch_command(command, cwd=os.getcwd()):
...
+    except KeyboardInterrupt:
+        stop()
+        if process:
+            process.poll()
+        pass
```
bmw: I don't think this is needed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
adferrand: OK, last argument on my side. If you do not agree after it, I revert to atexit.

It is true that there is no actual use in certbot codebase, server teardown will work correctly with atexit. Two things pushed me to make this change. First one is, as I said before, potential uses of this logic in other contexts in certbot.

Second one is that the reason we did this PR, so to offer a way to have an alternative to the bash scripts that will be removed, made me realized that Certbot integration tests could be run by other projects. I would expect for instance that other ACME server projects are using some tests against certbot, and maybe reuse parts of our own tests. 

So in the scope of external usages, I do not like a lot the atexit approach, as we do not offer a way to control when the teardown occurs: it is at the end of the python process, independantly from what is happening during the script itself. I already talked about pytest programmatical call on that matter.

Here I give a way to control this teardown. Future useful API could be also to provide a context manager, even if with these modifications, it is really easy to do it by yourself. 

And finally, the modifications done to implement that are very lightweight, I think it does not add some complexity. So the low usability added on our sole interest comes with a low price, it stays balanced. 

What do you think of that?

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
adferrand: Acknowleded

certbot-ci/certbot_integration_tests/certbot_tests/context.py
adferrand: Are you talking about a potential weird use here? Because in the current code I always pass the value as a keyword argument. As passing an optional value positionnally is generally not a good practice, I would expect that this kind of usage would be catched during a review if introduced.

My main concern here is to remove the old behavior, inherited from the bash scripts, to have 5 methods to call certbot, and have a situation where `--renew-by-default` then `--no-force-renew` are passed to Certbot when `certbot_no_force_renew` was invoked, and expect things are working as expected because magic.

certbot-ci/certbot_integration_tests/nginx_tests/context.py
adferrand: See above

adferrand: In fact not, not as above. Indeed I am passing positionally the optional argument on nginx tests. Bad me.

Checking more closely, in fact I think we can always renew by default, because each second call to Certbot is about rolling back the nginx config, not creating a new certificate. But never renew by default will also work, since each first call to Certbot is for a new domain. 

What about just removing the parameter in `certbot_test_nginx`, and always renew by default (this instead of no renew by default to be consistent with the major case in certbot tests).

Also, why not calling directly `certbot_test` in this function instead of `_common_test` from parent class? Then we could remove `_common_test` completely.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,73 @@
...
+        '-vv'
+    ]
+
+    command.extend(certbot_args)
+    command.extend(additional_args)
+
+    subprocess_call = subprocess.check_output if capture_stdout else subprocess.check_call
+
+    print('--> Invoke command:\n=====\n{0}\n====='.format(subprocess.list2cmdline(command)))
+    return subprocess_call(command, universal_newlines=True,
```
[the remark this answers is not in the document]
adferrand: I admit this can be really different values. Ideally, I would like this function to do both: returning the output, and piping it to standard stdout. Because first case to cover here is from certbot tests, where the returned output need to be asserted, and second case is call from shell, where you expect to see dynamically the output of the command.

However, this is hard to code. I mean, you can capture everything, then print the output before returning it, but then you see the all output in one bunch only at the end of the command. Not great for a shell command. 

Note that it would be also interesting during the tests themselves. Currently, pytest displays correctly the output if a test fails, because in fact all output is redirected in stderr, that is not captured by subprocess, but directly piped to stderr, so captured by pytest. 

All of this to say:
1) We can keep things like this
2) We can have two dedicated methods, one using `subprocess.check_call, another using `subprocess.check_output` and call appropriately
3) Implement the logic to do everything in the same method (returning output and piping it dynamically to stdout)
4) Your idea?

certbot-ci/certbot_integration_tests/utils/certbot_call.py
adferrand: I wanted to do that initially, but these variables are set only when the script is invoked from the shell, not during integration tests, and if they are global, they would be evaluated all the time. 

If your concern is about avoiding duplication, why not put this logic in a dedicated function with a meaningful name?

certbot-ci/setup.py
adferrand: OK for `run_acme_server` as run is more associated to foreground processes than start.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,73 @@
...
+from distutils.version import LooseVersion
+import subprocess
+import sys
+import os
+
+from certbot_integration_tests.utils import misc
+
+
+def certbot_test(certbot_args, directory_url, http_01_port, tls_alpn_01_port,
+                 config_dir, workspace, force_renew, capture_stdout):
```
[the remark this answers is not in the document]
adferrand: Done.

certbot-ci/setup.py
adferrand: Done.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Increased sleep to 3600 seconds.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Removed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -195,3 +196,36 @@ def _launch_command(command, cwd=os.getcwd()):
...
+
+        docker_name = 'pebble_pebble_1' if 'pebble' in server_type else 'boulder_boulder_1'
+        process = subprocess.Popen(['docker', 'logs', '-f', docker_name])
+
+        while True:
+            time.sleep(1)
+    except KeyboardInterrupt:
+        stop()
+        if process:
+            process.poll()
```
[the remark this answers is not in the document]
adferrand: Oh, it was `wait` I wanted to use. Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
bmw: A big part of the reason I'm not a huge fan of this change is having the two functions as return values feels a bit awkward. Are there any alternatives you like better such as a context manager or a simple class with a `workspace` attribute and start and stop methods?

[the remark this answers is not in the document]
adferrand: Context manager is, I think, my preferred feature in Python, for sure I am willing to implement that.

Done ^^

adferrand:
To respond on the high-level comments on second review @bmw:
1. Done.
2. Done.
3. Once this PR is merged, and Boulder/Pebble dev teams have confirmed documentation and tests are migrated to these new commands on their side.

cpu:
> Once this PR is merged, and Boulder/Pebble dev teams have confirmed documentation and tests are migrated to these new commands on their side.

I'm keeping an :eye: on this PR's progress to make sure the Boulder/Pebble changes are made quickly. I will test the WIP branch sometime today to get the diff ready ahead of it being merged.

cpu:
>  I will test the WIP branch sometime today to get the diff ready ahead of it being merged.

Everything worked the way @adferrand described in the PR description :tada: I was able to follow the updated instructions to issue a certificate using `certbot_test` against a dev Boulder instance.

I opened a draft PR on Boulder to update the "Working with Certbot" section where we previously described using `test/integration/_common.sh`: https://github.com/letsencrypt/boulder/pull/4247 That should be ready to be merged when this branch is :+1: 

> Once this PR is merged, and Boulder/Pebble dev teams have confirmed documentation and tests are migrated to these new commands on their side.

I checked the Pebble repo and this branch shouldn't affect anything there. We didn't use the `certbot_test` alias there or describe how to get it.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
bmw: This is a response to https://github.com/certbot/certbot/pull/7073/files#r289605994.

To clarify, what I wanted to add constants for were the default values like `https://localhost:14000/dir` which is used in multiple places. I don't have any complaints about the environment variable stuff here.

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
 
 .. _`virtualenv docs`: https://virtualenv.pypa.io
 
+Running certbot against a locally deployed Pebble CA
```
bmw: I'd move this to be a new section under "Testing" after "Integration testing with the Pebble CA". My thinking here is that while this is undoubtedly useful, I think this is the least important part of testing since it's just manual testing so it doesn't need to come so early in the doc.

In this new context, it changes the titles a bit. "Integration testing with the Pebble CA" could be something like "Running automated integration tests" and this section could be called "Running manual integration tests".

What do you think?

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By following the instructions above, you can execute Certbot against a local instance
```
bmw: I think this is what you mean, but you should double check.

I also added the word "manually" to differentiate it from the automated integration tests.

```suggestion
By following the instructions below, you can manually execute Certbot against a local instance
```

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+of Pebble CA. This is useful to verify that the modifications done to the code makes
```
bmw:
```suggestion
of the Pebble CA. This is useful to verify that the modifications done to the code makes
```
I also personally like "Pebble ACME server" over "Pebble CA" but either is fine.

Also, optional nit: I think it'd be useful to make the first reference to "Pebble" in each of these sections link to https://github.com/letsencrypt/pebble. If you're a new dev, what's a Pebble?

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+You need Docker installed, and a user with access to the Docker client.
+
+The virtual environment set up with `python tools/venv.py` contains two commands
+that can be used once the virtual environment is activated:
+
+.. code-block:: shell
+
+    run_acme_server
+
+- Start a local instance of Pebble CA, and keep displaying its log on foreground.
```
bmw:
```suggestion
- Starts a local instance of Pebble and runs in the foreground printing its logs.
```

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+- Press CTRL+C to stop this instance.
+- This instance is configured to validate challenges against a certbot executed locally.
```
bmw:
```suggestion
- This instance is configured to validate challenges against certbot executed locally.
```

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+
+.. code-block:: shell
+
+    certbot_tests [ARGS...]
+
+- Execute certbot with the provided ARGS in test mode: verbose output, debug enabled...
```
bmw: I'd personally get rid of the phrase "test mode" and say something like:

> Execute Certbot with the provided arguments and other arguments useful for testing such as verbose output, full tracebacks in case Certbot crashes, etc.

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+- Execution is preconfigured to interact with the Pebble CA started with ``run_acme_server``.
+- Any arguments can be passed as they would be to Certbot (eg. ``certonly -d test.example.com``).
```
bmw: nit: I'd say `certbot_tests certonly -d test.example.com` to complete the example.

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+
+Here is a typical workflow to verify that Certbot successfully issued a certificate
+using an HTTP-01 challenge on a machine with Python 3:
+
+.. code-block:: shell
+
+    python tools/venv3.py
+    source venv3/bin/activate
+    run_acme_server &
+    certbot_tests certonly --standalone -d test.example.com
```
bmw:
```suggestion
    certbot_test certonly --standalone -d test.example.com
```

docs/contributing.rst
```diff
@@ -69,6 +69,45 @@ found in the `virtualenv docs`_.
...
+    # To stop Pebble, launch `bg` to get back the background job, then press CTRL+C
```
bmw: Did you mean fg?

certbot-ci/certbot_integration_tests/utils/certbot_call.py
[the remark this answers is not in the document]
adferrand: Done, with option 2. In fact, I remove the specific branch of calling certbot without output capturing, to put it directly in the `main` function, since this is the only place where is is useful.

certbot-ci/certbot_integration_tests/nginx_tests/context.py
adferrand: I did the two propositions (always force renew in nginx, remove `_common_test`.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -10,6 +10,21 @@
 
 def certbot_test(certbot_args, directory_url, http_01_port, tls_alpn_01_port,
                  config_dir, workspace, force_renew, capture_stdout):
+    """
+    Invoke the certbot executable available in PATH in a test context for the given args.
+    The test context consists in running certbot in debug mode, with various flags suitable
+    for tests (eg. no ssl check, customizable ACME challenge ports and config directory ...).
+    :param str[] certbot_args: the arguments to pass to the certbot executable
```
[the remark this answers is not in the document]
adferrand: Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -20,27 +20,45 @@
...
-    with the relevant pytest xdist node.
+    with the relevant pytest xdist node, and appropriate method to start and stop the stack.
     :param str acme_server: the type of acme server used (boulder-v1, boulder-v2 or pebble)
     :param str[] nodes: list of node names that will be setup by pytest xdist
-    :return: a dict describing the challenge ports that have been setup for the nodes
-    :rtype: dict
+    :param bool proxy: set to False to not start the Traefik proxy
+    :return: a tuple with a dict describing the challenge ports that have been setup for the nodes,
+             a function to start the stack, and one to clean up everything
+    :rtype: tuple
```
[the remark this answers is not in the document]
adferrand: I like `ACMEServer`. I implemented that.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
adferrand: Ok, you meant constants to share in all the project. Added a specific module for that, took the occasion also to put boulder directory urls and traefik api port in it.

docs/contributing.rst
adferrand: Did the modification, and put a link ref for Pebble.

docs/contributing.rst
adferrand: Indeed !

docs/contributing.rst
adferrand: Done.

docs/contributing.rst
adferrand: Improved the sentence with your advices.

docs/contributing.rst
adferrand: What you propose was my first move. Then thinking of it again, I asked myself this question: as a contributor, willing to add a functionality, what would be the first thing I would do after writing the code?

I think it is: I would execute certbot to check my functionality is working.

In fact, we could consider that calling certbot to check the functionality is not part of the Quality Assurance process, it is part of the development workflow, and what I add here is a tool to help you doing this step. 

In that spirit, running the unit tests and the integration tests is part of the QA, and will maybe be executed by the contributor, or inevitably by the CI that will yell at you back when the PR is opened ^^

So in this case, the tool added here should be part of the dev section, right where I placed it, and unit/integration tests are in the testing part.

Does it make sense to you?

docs/contributing.rst
adferrand: In fact I wanted to refer to the previous section to ensure that the user is in a properly set up venv, because you need that before calling the commands exposed by certbot-ci. I improved the clarity of this paragraph.

docs/contributing.rst
bmw: I understand where you're coming from, but I still would like to see this change for a couple reasons.

The first reason is that it breaks the flow of the document, at least in some cases. For example, when we run Certbot hackathons, we usually tell people to go to this doc. In master, the changes are described in this order:

1. Set up the development environment.
2. Find an issue to work on.
3. Test it.

I think this is the correct workflow as there's not much to test against Pebble until you have something to work on.

The second reason is I suspect most core Certbot devs won't do this and therefore we shouldn't make these instructions so prominent. I suspect they'll rely more heavily on automated testing or manually testing against staging.

For those reasons, I'll suggest again you make the changes I described above, but as a compromise, I think you could add links to this section at the end of "Running a local copy of the client" and/or you could add a brief intro to the top of the "Testing" section with links to each section.

adferrand: Indeed, I can confirm your suspicion about core developers, at least on my side ^^

I agree with your compromise. Then I will put an intro on top of the Testing section. I think in this case it is the clearest thing to do.

adferrand: Done.

certbot-ci/certbot_integration_tests/utils/certbot_call.py
```diff
@@ -0,0 +1,98 @@
...
+from certbot_integration_tests.utils.constants import *
+
+
+def certbot_test(certbot_args, directory_url, http_01_port, tls_alpn_01_port,
+                 config_dir, workspace, force_renew=True):
+    """
+    Invoke the certbot executable available in PATH in a test context for the given args.
+    The test context consists in running certbot in debug mode, with various flags suitable
+    for tests (eg. no ssl check, customizable ACME challenge ports and config directory ...).
+    This command capture stdout and returns it to the caller.
```
bmw:
```suggestion
    This command captures stdout and returns it to the caller.
```

docs/contributing.rst
```diff
@@ -179,6 +148,52 @@ With a user allowed to access your local Docker daemon, run:
...
+
+- Starts a local instance of Pebble and runs in the foreground printing its logs.
+- Press CTRL+C to stop this instance.
+- This instance is configured to validate challenges against certbot executed locally.
+
+.. code-block:: shell
+
+    certbot_tests [ARGS...]
+
+- Execute certbot with the provided arguments and other arguments useful for testing purpose,
```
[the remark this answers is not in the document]
bmw: You can also say "useful for testing purposes".

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -14,13 +14,15 @@
...
-HTTP_01_PORT = 5002
 
-
-class _Handler(object):
+class ACMEServer(object):
+    """
+    Handler to start and stop the ACME server, and get its configuration (eg. challenges ports).
+    Upon stop the server assets are properly cleaned up. ACMEServer is also a context manager,
+    and so can be used as it to ensure implicit safe stop/cleanup.
+    """
```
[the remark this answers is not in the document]
adferrand: Done, by focusing the respective docstring of `ACMEServer` and `setup_acme_server` to their specific actions.

docs/contributing.rst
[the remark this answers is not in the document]
adferrand: Then I choose the minimal correction ^^

cpu:
@adferrand @bmw Thanks folks! I've merged the [associated Boulder README update](https://github.com/letsencrypt/boulder/commit/3918714bc1aa97ed4d8e4b1b6eb8c197c4f2055c).

PR #7081, opened 2019-05-17
Fix unpinned dependencies tests towards botocore and urllib3

adferrand:
Currently unpinned dependencies tests are failing, see latest Travis nightly run: https://travis-ci.com/certbot/certbot/builds/112182519

This is because:
1) New version of `requests` has been released yesterday (`2.22.0`), and it relax its transitive dependency from `urllib<1.25` to `urllib<1.26`
2) Current version of `botocore` in the route53 certbot dns plugin still requires `urllib<1.25`

Several issues have been raised in `botocore` project about this, with one issue tracking down the update of `urllib3` dependency in `botocore`: https://github.com/boto/botocore/issues/1733

So hopefully `botocore` will be update before our next release, but nonetheless I make a PR here to fix the situation until that, by capping our own transitive dependency for `urllib3` to `<1.25`. I prefer doing that instead of capping `requests` version, because it still compatible with `urllib3` 1.24.x branch, and we should keep the benefit of the latest improvements in `requests`.

Once `botocore` is updated, this PR should be reverted.

### DIFF
--- a/certbot-dns-route53/setup.py
+++ b/certbot-dns-route53/setup.py
@@ -9,7 +9,7 @@ install_requires = [
     # boto3 requires urllib<1.25 while requests 2.22+ requires urllib<1.26
     # Since pip lacks of real dependency graph resolver, it will peak the constraint only from
     # requests, and install urllib==1.25.2. Setting explicit dependency here solves the issue.
-    # Check https://github.com/boto/botocore/issues/1733 for resolution on boto3 side.
+    # Check https://github.com/boto/botocore/issues/1733 for resolution in botocore.
     'urllib3<1.25',
     'acme>=0.29.0',
     'certbot>=0.34.0',
--- a/certbot-dns-route53/setup.py
+++ b/certbot-dns-route53/setup.py
@@ -7,7 +7,7 @@ version = '0.35.0.dev0'
 # acme/certbot version.
 install_requires = [
     'acme>=0.29.0',
-    'certbot>=0.35.0.dev0',
+    'certbot>=0.34.0',
     'boto3',
     'mock',
     'setuptools',
--- a/certbot-dns-route53/setup.py
+++ b/certbot-dns-route53/setup.py
@@ -7,7 +7,7 @@ version = '0.35.0.dev0'
 # acme/certbot version.
 install_requires = [
     # boto3 requires urllib<1.25 while requests 2.22+ requires urllib<1.26
-    # Since pip lacks of real dependency graph resolver, it will peak the constraint only from
+    # Since pip lacks a real dependency graph resolver, it will peak the constraint only from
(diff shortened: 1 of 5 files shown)


adferrand:
Relevant nightly build can be seen here: https://travis-ci.com/certbot/certbot/builds/112210832

bmw:
Whoever ends up resolving https://github.com/pypa/pip/issues/988 instantly becomes one of my favorite people. (@adferrand, try not to let this comment nerdsnipe you 😝 I think it's a big and fairly political task.)

What I think we should do here is this:

1. Pin back `urllib3` in `certbot-dns-route53/setup.py` with a comment explaining that this is needed to make installation work with `pip` due to https://github.com/pypa/pip/issues/988, https://github.com/boto/botocore/issues/1733, and `pip` preferring the newest version of all dependencies. I'd also say that if https://github.com/boto/botocore/issues/1733 has been resolved, the dependency on `urllib3` specified here can be safely removed. I think the fix should go in the route53 plugin since that is the only component affected and the comment is useful to describe what is going on to package maintainers who may need to fix this up on their end.
2. Create an issue in the current milestone for us to remove this pinning with a link to the boto3 issue.

What do you think?

certbot-dns-route53/setup.py
```diff
@@ -6,6 +6,11 @@
 # Remember to update local-oldest-requirements.txt when changing the minimum
 # acme/certbot version.
 install_requires = [
+    # boto3 requires urllib<1.25 while requests 2.22+ requires urllib<1.26
+    # Since pip lacks of real dependency graph resolver, it will peak the constraint only from
```
bmw:
```suggestion
    # Since pip lacks a real dependency graph resolver, it will peak the constraint only from
```

certbot-dns-route53/setup.py
```diff
@@ -6,6 +6,11 @@
...
+    # requests, and install urllib==1.25.2. Setting explicit dependency here solves the issue.
```
bmw:
```suggestion
    # requests, and install urllib==1.25.2. Setting an explicit dependency here solves the issue.
```

certbot-dns-route53/setup.py
```diff
@@ -6,6 +6,11 @@
...
+    # Check https://github.com/boto/botocore/issues/1733 for resolution on boto3 side.
```
bmw:
```suggestion
    # Check https://github.com/boto/botocore/issues/1733 for resolution in botocore.
```

adferrand:
I added #7083 to track revert on this PR.

certbot-dns-route53/setup.py
bmw: After this change, this PR LGTM.

PR #7092, opened 2019-05-21
Turn off session tickets for versions of Nginx that support it

### LINKED ISSUE #6902: Disable TLS session tickets in Nginx?
Inspired by https://jhalderm.com/pub/papers/forward-secrecy-imc16.pdf, we might want to look into changing our nginx plugin's behavior around TLS session tickets. The [Mozilla site](https://mozilla.github.io/server-side-tls/ssl-config-generator/) where we get our TLS config from has TLS session tickets disabled, but this setting is not present in our config file at https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/options-ssl-nginx.conf.

This is likely due to the minimum version Nginx that supports these settings, however, I don't believe this has been investigated for a while. We should see what version these settings were added to Nginx and what versions are available in major distros.

Even if support is not available everywhere yet, it is probably worth adding these values conditionally as I suspect it will work on the vast majority of systems.

sydneyli:
fixes #6902 in line with Mozilla's security recommendations.

having been changed.

Will do!

One potential concern is that it's still possible here to downgrade the config to *not* disable tickets for newer versions of nginx. Is this something we should be concerned about?


bmw:
Let me know if you'd like me to take a look at the Windows test failures.

ohemorange:
Gave this a quick look but plan to look more closely later. The two main paths I plan to trace through the code:

previously installed certbot, new nginx, does the file update correctly?
previously installed certbot, old nginx, does the file update correctly?

This is mostly a note to myself. I think the first case should be covered by existing tests, but we should ideally add a test for the second case.

sydneyli:
> Ok, cool, satisfied with these paths. I'd still love a test for that second case.
While looking into this I don't think it currently tests for the first case correctly either-- working on a fix!

> How are the tests on Windows going?
I have access to a Windows machine briefly this morning so I'll take a look soon.

I'll let you know when it's ready for anohter round!

sydneyli:
Windows finally fixed. Ready for another review! The issue with the existing test for the first path [here](https://github.com/certbot/certbot/blob/f8c1facc267a77002e6e786671cabcd4d446fe66/certbot-nginx/certbot_nginx/tests/configurator_test.py#L911) was that by mocking sha256 entirely, it was hitting [this](https://github.com/certbot/certbot/blob/f8c1facc267a77002e6e786671cabcd4d446fe66/certbot/plugins/common.py#L444) case instead of the next one (which prompts an update).

Let me know how this looks. Happy to change around the naming, too.

sydneyli:
Thanks for merging the changelog! You beat me to it by a couple seconds :)

PR #7095, opened 2019-05-21
Avoid to delete both webroot_map and webroot_path

### LINKED ISSUE #7048: Certbot drops webroot options from renewal configuration file
To reproduce this, I think you need Certbot 0.31.0+ and an ACME server that reuses authorizations.

Here is each command and the renewal configuration file it produces.

The plugin used first here is irrelevant, you just have to get a cert with a valid authz.
```
sudo certbot certonly --staging --webroot -w /var/www/html -d example.com
```
produces
```
# renew_before_expiry = 30 days
version = 0.34.0
archive_dir = /etc/letsencrypt/archive/example.com
cert = /etc/letsencrypt/live/example.com/cert.pem
privkey = /etc/letsencrypt/live/example.com/privkey.pem
chain = /etc/letsencrypt/live/example.com/chain.pem
fullchain = /etc/letsencrypt/live/example.com/fullchain.pem

# Options used in the renewal process
[renewalparams]
authenticator = webroot
account = 0652f690fd6e48af380ad6d6a1b4e792
webroot_path = /var/www/html,
server = https://acme-staging-v02.api.letsencrypt.org/directory
[[webroot_map]]
example.com = /var/www/html
```
This next command is the start of the problem. Notice the empty value from `webroot_map` when compared to the previous value.
```
sudo certbot certonly --staging --webroot -w /var/www/html -d example.com --renew-by-default
```
produces
```
# renew_before_expiry = 30 days
version = 0.34.0
archive_dir = /etc/letsencrypt/archive/example.com
cert = /etc/letsencrypt/live/example.com/cert.pem
privkey = /etc/letsencrypt/live/example.com/privkey.pem
chain = /etc/letsencrypt/live/example.com/chain.pem
fullchain = /etc/letsencrypt/live/example.com/fullchain.pem

# Options used in the renewal process
[renewalparams]
authenticator = webroot
account = 0652f690fd6e48af380ad6d6a1b4e792
webroot_path = /var/www/html,
server = https://acme-staging-v02.api.letsencrypt.org/directory
[[webroot_map]]
```
Finally, this command removes all references to the webroot from the renewal configuration file.
```
sudo certbot renew --force-renewal
```
produces
```
# renew_before_expiry = 30 days
version = 0.34.0
archive_dir = /etc/letsencrypt/archive/example.com
cert = /etc/letsencrypt/live/example.com/cert.pem
privkey = /etc/letsencrypt/live/example.com/privkey.pem
chain = /etc/letsencrypt/live/example.com/chain.pem
fullchain = /etc/letsencrypt/live/example.com/fullchain.pem

# Options used in the renewal process
[renewalparams]
authenticator = webroot
account = 0652f690fd6e48af380ad6d6a1b4e792
server = https://acme-staging-v02.api.letsencrypt.org/directory
```

Certbot dropped the reference to the webroot path. This is bad because renewal will fail with:
```
Missing command line flag or config entry for this setting: Input the webroot for example.com
```
when authz are no longer available to be reused.

## What's going on?

While I haven't verified this, I think what's going on here is a combination of authz reuse (which Certbot understands as of version 0.31.0) and full parsing of the webroot options not happening until the challenge is preformed by the webroot plugin.

I'm imagining the 2nd command causes the value for `webroot_path` to be preserved because it is set on the command line. `webroot_map` is preserved because `webroot_path` is set on the command line and `webroot_map` is considered modified if `webroot_path` is set due to the code [here](https://github.com/certbot/certbot/blob/d391fb8876dd97d5f446bcd1667ecdce0c4b187c/certbot/cli.py#L141), however, `webroot_map` is empty because the webroot plugin never ran.

When `certbot renew --force-renewal` runs, I think `webroot_map` is initially set to it's default value and `webroot_path` is dropped due to the code [here](https://github.com/certbot/certbot/blob/d391fb8876dd97d5f446bcd1667ecdce0c4b187c/certbot/renewal.py#L103). After that, I think `webroot_map` is also dropped and not written to the renewal configuration file because it has its default value.

adferrand:
Fixes #7048.

This PR "partially" correct #7048. By "partially" here, I mean that the issue is effectively corrected, but not revert to the previous behavior, as this would require more in-depth modifications of the plugins API and behavior.

The idea is that in the specific corner case where `certbot certonly` is invoked two times using `webroot` plugin in a row for a given domain, the second call does not in fact perform a new challenge, but instead reuse the still valid authorization from ACME server. Indeed, authorizations remain valid for a couple of time.

However, this means that the `perform` method of `webroot` here is not invoked. But this method contains some logic to fill the `webroot_map` values in the renewal configuration file, that is not invoked then. If it is not invoked, it means that the last value in `webroot_path` is not associated to the given domains that are not already in `webroot_map`. In particular, this means in the case of a single provided domain that the webroot_map is empty `{}`.

Here it becomes tricky. If we invoke `certbot renew` after that, the method `_restore_webroot_config` is puzzled: it sees that a `webroot_map` is set, and so save its empty value. Then it does not copy `webroot_path`, as there was a value for `webroot_map`, and `webroot_path` is itself a corner case from ancient values ...

Finally, we just lost both `webroot_path` values and `webroot_map`, and the renewal configuration is broken from now on.

In the case of more than one given domain, this broken behavior leads to at least one missing association webroot -> domain in the `webmap_root`.

This PR fixes that by copying the `webroot_path` value if it exists and not explicitly set by CLI. With this, even if `webroot_map` is lost, the renewal configuration remains functional with `webroot_path` using the latest value in it for all missing domains in `webroot_map`, and matches functionally the previous behavior, even if the renewal configuration file remains impacted.

I pass the lool to @bmw.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Renewal parameter `webroot_path` is always saved, avoiding some regressions
+  when `webroot` authenticator plugin is invoked with no challenge to perform.
 
 Despite us having broken lockstep, we are continuing to release new versions of
 all Certbot components during releases for the time being, however, the only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Despite us having broken lockstep, we are continuing to release new versions of
 all Certbot components during releases for the time being, however, the only
 package with changes other than its version number was:
 
+* certbot
 * certbot-dns-rfc2136
 
 More details about these changes can be found on our GitHub repo.
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -109,6 +109,8 @@ def _restore_webroot_config(config, renewalparams):
     if "webroot_map" in renewalparams:
         if not cli.set_by_cli("webroot_map"):
             config.webroot_map = renewalparams["webroot_map"]
+    # To understand why webroot_path and webroot_map processing are not mutually exclusive, see:
+    # https://github.com/certbot/certbot/pull/7095
     if "webroot_path" in renewalparams and not renewalparams.get('webroot_map', {}):
         logger.debug("Ancient renewal conf file without webroot-map, restoring webroot-path")
         wp = renewalparams["webroot_path"]
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -106,13 +106,11 @@ def _restore_webroot_config(config, renewalparams):
     restoring logic is not able to correctly parse it from the serialized
     form.
     """
-    if "webroot_map" in renewalparams:
-        if not cli.set_by_cli("webroot_map"):
-            config.webroot_map = renewalparams["webroot_map"]
-    # To understand why webroot_path and webroot_map processing are not mutually exclusive, see:
-    # https://github.com/certbot/certbot/pull/7095
-    if "webroot_path" in renewalparams and not renewalparams.get('webroot_map', {}):
-        logger.debug("Ancient renewal conf file without webroot-map, restoring webroot-path")
+    if "webroot_map" in renewalparams and not cli.set_by_cli("webroot_map"):
+        config.webroot_map = renewalparams["webroot_map"]
+    # To understand why webroot_path and webroot_map processing are not mutually exclusive,
+    # see https://github.com/certbot/certbot/pull/7095
+    if "webroot_path" in renewalparams and not cli.set_by_cli("webroot_path"):
         wp = renewalparams["webroot_path"]
         if isinstance(wp, six.string_types):  # prior to 0.1.0, webroot_path was a string
             wp = [wp]
--- a/certbot/renewal.py
+++ b/certbot/renewal.py
@@ -109,7 +109,7 @@ def _restore_webroot_config(config, renewalparams):
     if "webroot_map" in renewalparams:
         if not cli.set_by_cli("webroot_map"):
             config.webroot_map = renewalparams["webroot_map"]
-    elif "webroot_path" in renewalparams:
+    if "webroot_path" in renewalparams and not renewalparams.get('webroot_map', {}):
         logger.debug("Ancient renewal conf file without webroot-map, restoring webroot-path")
         wp = renewalparams["webroot_path"]
         if isinstance(wp, six.string_types):  # prior to 0.1.0, webroot_path was a string
--- a/certbot/plugins/webroot.py
+++ b/certbot/plugins/webroot.py
@@ -74,27 +74,28 @@ to serve all files under specified web root ({0})."""
         self._created_dirs = []  # type: List[str]
 
     def prepare(self):  # pylint: disable=missing-docstring
-        # To understand why set_webroots is in the early prepare phase, and not in the perform phase,
-        # see https://github.com/certbot/certbot/pull/7095
-        self._set_webroots(self.config.domains)
+        pass
 
     def perform(self, achalls):  # pylint: disable=missing-docstring
+        self._set_webroots(achalls)
+
         self._create_challenge_dirs()
 
         return [self._perform_single(achall) for achall in achalls]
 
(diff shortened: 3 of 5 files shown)


certbot/renewal.py
```diff
@@ -109,7 +109,9 @@ def _restore_webroot_config(config, renewalparams):
     if "webroot_map" in renewalparams:
         if not cli.set_by_cli("webroot_map"):
             config.webroot_map = renewalparams["webroot_map"]
-    elif "webroot_path" in renewalparams:
+    # To understand why webroot_path and webroot_map processing are not mutually exclusive, see:
+    # https://github.com/certbot/certbot/pull/7095
+    if "webroot_path" in renewalparams and not renewalparams.get('webroot_map', {}):
```
bmw: Can we simplify this old code by removing the `and not renewalparams...` condition (and the `logger.debug` call below) and always restore the `webroot_path` if it exists? If not, why not?

adferrand: Someone took great care here to make this branch be executed on specific conditions, and for legacy support only. I do not know why, but since I have no spec here, I prefer to limit at the maximum the deviation from the previous behavior.

Adding the condition is sufficient to solves the bug, so I would prefer to keep it and reduce the risk of other regressions (maybe from logic outside of Certbot) that could be created by always propagating both parameters.

If we remove the logger, we change from a legacy situation to a standard. Why changing that?

adferrand:
If it solely for regression testing, I think a unit test will be sufficient. On that aspect, it is still a corner case and I would like to avoid invest time to build an integration test for corner cases that is also consuming resources (these tests are slow).

If it is about testing the capability of Certbot to reuse authorizations when it is supported by the acme server, then I agree about writing an integration tests. 

So, what do you have in mind?

certbot/renewal.py
bmw: I believe the previous behavior was unnecessary, overly complex, and when combined with other changes, created this problem in the first place. We could potentially leave this to be fixed "later", but since we're here I think we should take the time to investigate and do it properly. I already have looked into it and think what I propose here should work, but I'd want a 2nd set of eyes on that decision. Do you feel comfortable doing that?

Regardless of any arguments for code health though, I now think this change this isn't sufficient. Follow my instructions at https://github.com/certbot/certbot/issues/7048 but use two webroot paths. (If you haven't seen this before, you can run `certbot certonly -w /first/webroot/path -d domain.at.first.path -w /second/webroot/path -d domain.at.second.path`.) Combing that with my instructions at https://github.com/certbot/certbot/issues/7048 caused information on the 2nd webroot path to be dropped with this patch.

bmw:
Adding some integration tests around authz reuse is probably a good idea, but I'm specifically interested in preventing regressions here.

certbot/renewal.py
adferrand: Implemented, as discussed.

adferrand:
So, finally I do not implement the reconstruction of webroot_map in the webroot plugin during the prepare phase. Indeed, the cure is worse than the decease, as the process to move is potentially interactive, and change too much thinks in the workflow. I prefer to stay simple, and the copy of `webroot_path` alone is sufficient to fix the bug.

adferrand:
@bmw, relevant unit tests have been added to renewal and webroot.

bmw:
Whoops. Looks like the code from https://github.com/certbot/certbot/pull/7054 accidentally got pushed to this branch.

bmw:
Also, don't forget to add a changelog entry about this.

adferrand:
Fixed about the unwanted merge of #7054, and changelog updated.

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Renewal parameter `webroot_path` is always saved, avoiding some regressions
+  when `webroot` authenticator plugin is invoked with no challenge to perform.
 
 Despite us having broken lockstep, we are continuing to release new versions of
 all Certbot components during releases for the time being, however, the only
```
bmw: Can we list `certbot` as a modified package below?

PR #7096, opened 2019-05-23
Added support for Pre/Post hooks to CSR (certbot certonly)

PVi1:
Hi all,

we have added pre/post hooks support to certbot certonly when using csr, like in case you have FIPS compliant hardware which does not support importing private key - eg. HP blade chassis and network devices.

### DIFF
--- a/certbot/main.py
+++ b/certbot/main.py
@@ -1165,15 +1165,20 @@ def _csr_get_and_save_cert(config, le_client):
     :rtype: `tuple` of `str`
 
     """
+    hooks.pre_hook(config)
     csr, _ = config.actual_csr
     cert, chain = le_client.obtain_certificate_from_csr(csr)
     if config.dry_run:
         logger.debug(
             "Dry run: skipping saving certificate to %s", config.cert_path)
         return None, None
-    cert_path, _, fullchain_path = le_client.save_certificate(
-        cert, chain, os.path.normpath(config.cert_path),
-        os.path.normpath(config.chain_path), os.path.normpath(config.fullchain_path))
+    try:
+        cert_path, _, fullchain_path = le_client.save_certificate(
+            cert, chain, os.path.normpath(config.cert_path),
+            os.path.normpath(config.chain_path), os.path.normpath(config.fullchain_path))
+    finally:
+        hooks.post_hook(config)
+
     return cert_path, fullchain_path
 
 def renew_cert(config, plugins, lineage):
--- a/docs/cli-help.txt
+++ b/docs/cli-help.txt
@@ -279,7 +279,8 @@ certonly:
 
   --csr CSR             Path to a Certificate Signing Request (CSR) in DER or
                         PEM format. Currently --csr only works with the
-                        'certonly' subcommand. (default: None)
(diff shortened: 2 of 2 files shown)


bmw:
Hi @PVi1! Thanks for the PR.

I don't entirely understand why this feature is needed though. You should be able to get the same behavior with a simple shell script like:
```
pre-hook.sh
certbot certonly --csr ...
post-hook.sh
```

PVi1:
Hi @bmw,

It will help to keep the consistency of code used to manage and deploy the certs. Eg. DELL chassis HW supports import of private keys while HP not, as stated before. This way certbot scripts works on the same logic and are easier to read(we can use --post-hooks in both cases altough in one case with renew and in other cases with certonly subcommand).
Or is there other idea behind pre/post hooks that cannot be handled the way you have described?

bmw:
> This way certbot scripts works on the same logic and are easier to read(we can use --post-hooks in both cases altough in one case with renew and in other cases with certonly subcommand).

Hm. This benefit seems pretty small since the command line already has to differ a significant amount (e.g. I assume you're not running `certbot renew --csr`). Would you agree with that or am I missing something?

> Or is there other idea behind pre/post hooks that cannot be handled the way you have described?

At least for use with `certbot certonly`, the only potentially difference in the running of hooks that I see would be that once the pre-hook runs, Certbot will try to run the post-hook even if issuance fails. Whether or not you'd get that same behavior depends on the language you're using, but it seems like it'd be pretty straightforward to implement regardless.

PVi1:
Hi @bmw, 

that's correct. It would be better to run --deploy-hook, so I don't need to check if crt had been really renewed, but I cannot use "certbot renew --csr" nor --deploy-hook because certificate lineage does not exist after I get crt from csr with certonly subcommand (_csr_get_and_save_cert in main.py cannot create it). 

The main motivation behind our PR, was just to add this functionality to certonly chain, because the same functionality already existed in renew chain. 

The way I understand what you have written is that all concept around pre/post hooks is redundant and will be probably removed in the future, because it can be replaced with script calls outside of certbot. Am I correct?

Peter

bmw:
I think pre and post hooks are redundant when used with `certbot certonly --csr` for the reason I described.

It's a little different when using `certbot certonly` without `--csr` or `certbot renew` though. One difference here is hooks are only run if the certificate is attempted to be issued. Another difference is any hooks you provide on the command line get saved when a certificate is successfully issued and will be rerun automatically when the certificate is reissued with `certbot renew`. Hooks you provide on the command line take precedence though so if you're running `certbot renew --post-hook <something>` that will be run instead of any hooks saved from prior runs.

A final difference is that with `certbot renew`, identical pre and post hooks will only run once even if multiple certificates are renewed. The way that works is each unique pre-hook is run before an attempt is made to issue that certificate, but the post-hook is only run once all certificates for that run have attempted to be issued.

So in short, I do think adding support for hooks to `certbot certonly --csr` is redundant because you can do the same thing outside of Certbot. I do not think this is the case for our current support for hooks though and we have no plans to remove them in the near future.

For this reason, I'm closing this PR, but thanks again for writing it and I hope that makes sense. If you have any questions about all this, I'd be happy to answer them.

PR #7098, opened 2019-05-23
Move IRC notifications to #certbot-devel.

bmw:
I tested this at https://travis-ci.com/certbot/certbot/builds/113496172 which successfully sent a notification to the new channel. @joohoi, I believe you were in the channel to see this message, but if not, we can trigger the build again.

Because I had to look up how to do this again, I added a comment describing how this value was created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -276,7 +276,11 @@ notifications:
   email: false
   irc:
     channels:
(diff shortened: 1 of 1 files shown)


bmw:
Thanks for the review! I went ahead and just deleted `use_notice` because its default is false according to https://docs.travis-ci.com/user/notifications/#configuring-irc-notifications.

I also retested the failures and they worked and updated my original post with a link to the new Travis build.

joohoi:
LGTM!

Issue #6955, opened 2019-04-12
Update Fedora AMI to Fedora 29+


bmw:
This has the added benefit of doing additional testing on Python 3.

If we want to do this and have all test farm tests pass, this is at least blocked on #6940, however, I have not tried running all tests with a Fedora 29 AMI and suspect that other tests need to be fixed as well. We'll need to update the initial version of certbot-auto used in `test_leauto_upgrades.sh` and probably fix `test_sdists.sh` to work on Python 3.

Fedora AMIs can be found at https://alt.fedoraproject.org/cloud/.

PR #7102, opened 2019-05-24
Update Fedora AMI

bmw:
Fixes https://github.com/certbot/certbot/issues/6955.

This updates the Fedora version used in our test farm tests to Fedora 30. The AMI ID comes from https://alt.fedoraproject.org/cloud/ where it is listed as their standard HVM AMI for the region we use `us-east-1` (US East (N. Virginia)).

Unfortunately, there were a lot of small changes required for this. The big reason for this is on Fedora, there isn't a Python 2 executable installed. In fact, there's not even an executable named `python`. It's just `python3`. Rather than installing another Python in each test, I wrote a script that the test scripts can share to figure out the different paths and names that should be used in their script. (This isn't used in `test_sdists.sh` because the logic is a little different.)

Other changes here worth flagging are:

* I changed the name of the variable `RUN_PYTHON3_TESTS` in `test_leauto_upgrades.sh` to `RUN_RHEL6_TESTS`. The tests that are run when this variable is set test the upgrade from Python 2 to Python 3 on RHEL 6. I think this new name is much better now that we also have Fedora running Python 3.
* I made `tools/simple_http_server.py` work on Python 3.

You can see tests passing with these changes at https://travis-ci.com/certbot/certbot/builds/113821476. I also ran `test_tests.sh` and they passed.

### DIFF
--- a/tests/letstest/apache2_targets.yaml
+++ b/tests/letstest/apache2_targets.yaml
@@ -40,8 +40,8 @@ targets:
     user: admin
   #-----------------------------------------------------------------------------
   # Fedora
-  - ami: ami-5c69df23
-    name: fedora28
+  - ami: ami-00bbc6858140f19ed
+    name: fedora30
     type: centos
     virt: hvm
     user: fedora
--- a/tests/letstest/scripts/set_python_envvars.sh
+++ b/tests/letstest/scripts/set_python_envvars.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+# This is a simple script that can be sourced to set Python environment
+# variables for use in Certbot's letstest test farm tests.
+
+# Some distros like Fedora may only have an executable named python3 installed.
+if command -v python; then
+    PYTHON_NAME="python"
+    VENV_SCRIPT="tools/venv.py"
+    VENV_PATH="venv"
+else
+    # We could check for "python2" here, however, the addition of "python3"
+    # only systems is what necessitated this change so checking for "python2"
+    # isn't necessary.
+    PYTHON_NAME="python3"
+    VENV_PATH="venv3"
+    VENV_SCRIPT="tools/venv3.py"
+fi
--- a/tests/letstest/scripts/test_apache2.sh
+++ b/tests/letstest/scripts/test_apache2.sh
@@ -45,8 +45,10 @@ if [ $? -ne 0 ] ; then
     exit 1
 fi
 
-python tools/venv.py -e acme[dev] -e .[dev,docs] -e certbot-apache
-sudo venv/bin/certbot -v --debug --text --agree-dev-preview --agree-tos \
+. tests/letstest/scripts/set_python_envvars.sh
+
+"$VENV_SCRIPT" -e acme[dev] -e .[dev,docs] -e certbot-apache
+sudo "$VENV_PATH/bin/certbot" -v --debug --text --agree-dev-preview --agree-tos \
                    --renew-by-default --redirect --register-unsafely-without-email \
                    --domain $PUBLIC_HOSTNAME --server $BOULDER_URL
 if [ $? -ne 0 ] ; then
@@ -55,7 +57,7 @@ fi
 
 if [ "$OS_TYPE" = "ubuntu" ] ; then
     export SERVER="$BOULDER_URL"
-    venv/bin/tox -e apacheconftest
+    "$VENV_PATH/bin/tox" -e apacheconftest
 else
     echo Not running hackish apache tests on $OS_TYPE
 fi
--- a/tests/letstest/scripts/test_apache2.sh
+++ b/tests/letstest/scripts/test_apache2.sh
@@ -45,6 +45,9 @@ if [ $? -ne 0 ] ; then
     exit 1
 fi
 
+# This script sets the environment variables PYTHON_NAME, VENV_PATH, and
+# VENV_SCRIPT based on the version of Python available on the system. For
+# instance, Fedora uses Python 3 and Python 2 is not installed.
 . tests/letstest/scripts/set_python_envvars.sh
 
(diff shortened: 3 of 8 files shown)


bmw:
I also tested this branch combined with #7099 at https://travis-ci.com/certbot/certbot/builds/113487633 and ran `test_tests.sh` locally and everything passed.

bmw:
> On the other side, I think that the progressive additions in letstest bash scripts make them harder and harder to read, understand and certainly maintain. I had to read for instance several time the scripts here to understand what was the effective version of Python used for each Fedora version.

Do you have a suggestion for something else I should do instead?

I personally think the general problem here is one of the thing we're testing in some of these scripts is `certbot-auto`'s ability to install the dependencies it needs such as Python and then we're relying on that same Python installation in these tests.

I think a solution could be to add a 2nd Python installation not in our `PATH` to try to minimize the chance certbot-auto uses it, and use that Python installation for running test code, however, this doesn't seem like a totally trivial change to make.

adferrand:
Honestly I would have liked to propose a solution together with my comment, but I do not have one. I can just state about the complexity here. As you said, there is no trivial solution here.

bmw:
@ohemorange, are you free to be a 2nd set of eyes on this given Adrien's concerns?

I spent some time rethinking the approach here to see if I could come up with something simpler that didn't involve largely rewriting these tests and failed to do so. My opinion is we should try to address stuff like this in the test farm test rewrite we're planning on doing soon.

ohemorange:
I will take a look.

ohemorange:
Did previous Fedora also only have python3 installed, or is that a new change? Should we be upgrading users to python3 like we do on RHEL 6, and testing that it's been upgraded correctly?

> I think a solution could be to add a 2nd Python installation not in our PATH to try to minimize the chance certbot-auto uses it and use that Python installation for running test code, however, this doesn't seem like a totally trivial change to make.

I agree that we shouldn't do this.

> I had to read for instance several time the scripts here to understand what was the effective version of Python used for each Fedora version.

What if we didn't use a general, reusable script here? Maybe it would be more readable if we hard-coded fedora using a different python where needed. That being said, while this approach makes the overall script more complex, I do think it's pretty traceable, especially given the comments, to see how `$PYTHON_NAME` is being set. @adferrand, I'm wondering which particular part was the tricky part -- maybe we could fix that specifically.

Also, what did you mean "for each Fedora version"? We're only testing one version of Fedora, no?

bmw:
> Did previous Fedora also only have python3 installed, or is that a new change?

It was a change we made back in March in https://github.com/certbot/certbot/pull/6812.

>  Should we be upgrading users to python3 like we do on RHEL 6, and testing that it's been upgraded correctly?

I personally don't think we should. The only time there's an upgrade path is if you had `certbot-auto` installed and you upgraded your entire OS from Fedora 28 or older to a newer version, but [Fedora 28 reached its EOL two days ago](https://fedoramagazine.org/fedora-28-end-of-life/) and if there was a bug in this, I don't think we'd fix it because we don't tell Fedora users to use `certbot-auto`.

> What if we didn't use a general, reusable script here? Maybe it would be more readable if we hard-coded fedora using a different python where needed.

I'm imagining basically replacing running `set_python_envvars.sh` with a copy of its code each time its needed right? I could do that if both of you like that approach better.

Another somewhat similar thing I could do which may help is add comments above where the script runs to better explain what it is doing.

adferrand:
> @adferrand, I'm wondering which particular part was the tricky part -- maybe we could fix that specifically.

The is no tricky part on their own. The problem I have is that the context of a given test is scattered in several parts distributed in different modules, and that makes harder to understand it. 

So in general, even if it would lead to more duplication, I will be more for an approach as @bmw proposed: write in each script the logic needed to make the correct python version be used.

> Also, what did you mean "for each Fedora version"? We're only testing one version of Fedora, no?

Well it is all about the point I said about the context being hard to understand. Is the script designed to execute on several Fedora distributions, or RHEL, or CentOS? You need to gather both the list of AMI defined, the common script to select correct python version, and the actual test script, to understand what will be done in an execution of letstest.

adferrand:
So to directly respond on your last comment @bmw, I think some explicit documentation is good, but having the complete  test context been understood by reading one unique file will be better, so I would go to duplicate the good, favoring the readability over maintainability.

bmw:
> So in general, even if it would lead to more duplication, I will be more for an approach as @bmw proposed: write in each script the logic needed to make the correct python version be used.

I think this is what @ohemorange suggested with her comment:

> What if we didn't use a general, reusable script here? Maybe it would be more readable if we hard-coded fedora using a different python where needed.

If she's OK with or even better prefers this approach, I'll make this change.

ohemorange:
I kind of like Brad's suggestion, actually:

> Another somewhat similar thing I could do which may help is add comments above where the script runs to better explain what it is doing.

I think this gets both the benefit of explaining what it happening inline, and the benefit of not duplicating code.

bmw:
For what it's worth, I personally like that approach better too. The comment I'd add is something like:
```
# This script sets the environment variables PYTHON_PATH, VENV_PATH, and
# VENV_SCRIPT based on the version of Python available on the system.
. tests/letstest/scripts/set_python_envvars.sh
```

ohemorange:
Maybe specifically add "Fedora uses python3"

adferrand:
Even if it is not my preferred approach, I am OK with it, as it is still improving readability. As said from the review, I am not searching to find a way  to change things right now, I think it should be studied during a more in-depth redesign. For now what matters is to be able to test Fedora 30 and keep the tests understandable.

bmw:
Just to make sure I didn't make a typo and break everything by adding a comment, I reran the tests on these changes at https://travis-ci.com/certbot/certbot/builds/113821476 and they passed.

bmw:
Closing and reopening to try and trigger AppVeyor.

PR #7105, opened 2019-05-29
Update docs/cli-help.txt -- typo and formatting

petecooper:
'areusing' -> 'are using'


bmw:
Thanks for the PR @petecooper!

This file is regenerated during releases and the problem in the code is at https://github.com/certbot/certbot/blob/d1330efe41c4939098fd37c5746c68919fe166b7/certbot/cli.py#L1456-L1457

We should fix the problem there as well.

PR #7108, opened 2019-05-30
Configure the stale bot

bmw:
This PR configures a conservative [stale](https://github.com/marketplace/stale) bot for this repo. To learn more about the options we can configure, see https://github.com/probot/stale#usage.

Please review this PR, but do not merge it. I think we should wait until the beginning of a week so we can better initially monitor things.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,28 @@
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 180
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  To help us better see what issues are still affecting our users, this issue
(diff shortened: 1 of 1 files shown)


ohemorange:
One thing we could do is do a round where we ignore issues assigned to team members, at least at first, using `exemptAssignees`. This isn't many issues, but could save us some work responding to any of those, if they exist.

bmw:
Good idea! I made that change.

.github/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  To help us better see what issues are still affecting our users, this issue
+  has been automatically marked as stale. If you still have this issue with an
+  up-to-date version of Certbot and are interested in seeing it resolved,
+  please add a comment letting us know.
```
ohemorange: We should probably tell them that it'll be closed if they don't -- something like "If there is no further activity, this issue will be closed."

bmw: I made the change with the same wording as you suggested except I added "automatically".

I initially left it out because I was worried about people adding a comment just to keep the issue open without actually checking if it's still a problem, but adding it probably prevents other problems by better explaining upfront what is going on.

ohemorange: A valid worry, but yeah, I expect being transparent will help.

Issue #7107, opened 2019-05-30
Ignore editor backups in renewal hooks


lechner:
Hi,

Would it be safer to exclude editor backups (*~) when renewal-hooks are run?

Per your request, it was posted [here](https://community.letsencrypt.org/t/editor-backup-files-executed-as-renewal-hooks/94750) first.  Thank you!

Kind regards,
Felix Lechner

## My operating system is (include version):

Debian testing

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

0.31.0-1 via apt

## I ran this command and it produced this output:

certbot renew --force-renewal

## Certbot's behavior differed from what I expected because:

Editor backup files are run when executing renewal hooks. I expected them to be ignored.

## Here is a Certbot log showing the issue (if available):

```
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Plugins selected: Authenticator nginx, Installer nginx
Renewing an existing certificate
Running deploy-hook command: /etc/letsencrypt/renewal-hooks/deploy/slapd-wallace
Running deploy-hook command: /etc/letsencrypt/renewal-hooks/deploy/slapd-wallace~
```

bmw:
Thanks for the well written issue @lechner! I agree this is a good idea.

If someone wants to take this on, the function where this should probably be done is https://github.com/certbot/certbot/blob/926c8c198cfb6158b7ad59988702c30e33f13d53/certbot/hooks.py#L260

lechner:
I filed #7109 but, as a contributor otherwise unfamiliar with the project's customs, I did not include an entry in `CHANGELOG.md`.

This pull request was tested on a live system, where it produced the following output even though a backup file was present:
```
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Plugins selected: Authenticator nginx, Installer nginx
Renewing an existing certificate
Running deploy-hook command: /etc/letsencrypt/renewal-hooks/deploy/slapd-wallace

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```

bmw:
This was fixed in #7109.

PR #7109, opened 2019-05-30
Ignore editor backups when running hooks.

lechner:
**Sorry, new contributor; no changelog entry provided.**

When processing hooks, certbot also runs editor backups even though such files are outdated, clearly warranted correction and may quite possibly be defective.

That behavior could lead to unexpected breakage, and perhaps even pose security risks—for example, if a previous script was careless with file permissions. As an aggravating factor, the backup runs after the corrected version and could unintentionally override a fix the user thought was properly implemented.

This commit causes editor backup files ending in tilde (~) to be excluded when running hooks.

Additional information can be found here:

https://github.com/certbot/certbot/issues/7107
https://community.letsencrypt.org/t/editor-backup-files-executed-as-renewal-hooks/94750

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 * Renewal parameter `webroot_path` is always saved, avoiding some regressions
   when `webroot` authenticator plugin is invoked with no challenge to perform.
+* Scripts in Certbot hook directories are no longer executed when their
+  filenames end in a tilde.
 
 Despite us having broken lockstep, we are continuing to release new versions of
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 * Renewal parameter `webroot_path` is always saved, avoiding some regressions
   when `webroot` authenticator plugin is invoked with no challenge to perform.
```
bmw: Can we delete this blank line?

lechner:
Are there outstanding "changes requested"? If so, a quick pointer would be appreciated. Thank you!

PR #7111, opened 2019-05-31
Remove scripts that are never run.

bmw:


bmw:
Nice catch! I've made that change as well.

PR #7113, opened 2019-05-31
Add HTTP2 Support

### LINKED ISSUE #3646: HTTP2 support in Nginx plugin
Split from #3640.

TheBrokenRail:
Fixes #3646

### DIFF
--- a/certbot-nginx/certbot_nginx/configurator.py
+++ b/certbot-nginx/certbot_nginx/configurator.py
@@ -114,9 +114,11 @@ class NginxConfigurator(common.Installer):
         # These will be set in the prepare function
         self.parser = None
         self.version = version
+        self.http2 = False;
         self._enhance_func = {"redirect": self._enable_redirect,
                               "ensure-http-header": self._set_http_header,
-                              "staple-ocsp": self._enable_ocsp_stapling}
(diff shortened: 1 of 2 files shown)


kolcon:
This is now quite old issue - any progress?

emilv:
What needs to happen with this PR to get it reviewed?

Wirone:
@bmw You added `priority: significant` label to #3646 few days ago but this PR is more than one year old... Maybe it can be merged and issue can be closed? :thinking:

bmw:
This PR is a start, but as far as I can tell, the code added here is unreachable without further modifications to Certbot.

To flag a few other high level things quickly:

* Flags like `--http2` need to be added. This code references them but doesn't actually add them to Certbot's CLI.
* The interactive UI code added here is unused and should be deleted for now. I think the initial version of this should default to HTTP2 not being enabled and you can only enable it through CLI flags. We change things in the future to push it to more users if it seems to be working well.
* Due to how [Certbot's plugin system](https://certbot.eff.org/docs/contributing.html#installer) works, enabling HTTP2 should probably be done through a proper call to `enhance` with the value and flags the plugin should expect properly documented.
* Tests need to be added

If someone wants to do this work, we'll take a look when we can.

PR #7120, opened 2019-06-04
Add docker-compose as a requirement of certbot-ci

### LINKED ISSUE #7110: Add dependency on docker-compose to certbot-ci
It's written in Python and it not being installed regularly bites me. I often install it in my venv so I have a newer version available than that provided by the OS without making global changes to my system, but when I recreate the venv, I often forget.

I'm not sure what other devs setups look like, but making this change would make mine easier.

What do you think @adferrand?

adferrand:
Fixes #7110 

This PR declares docker-compose as a requirement for certbot-ci. This way, a recent version of docker-compose is installed in the standard virtual environment set up by `tools/venv.py` and `tools/venv3.py`, and so is available to pytest integration tests from `tox` or in the virtual environment enabled.


adferrand:
Last time, release of docker-compose took 3 months. I am also a little bothered to downgrade requests in certbot-auto just for a dev dependency.

Alternatively, we could keep this requirement unfulfilled until new version of docker-compose: indeed as seen from the PR you mentionned, the incompatibility is virtual, due to the strict control made by docker-compose maintainers on the request version because the package bite them several times in the past. However this version did not pushed incompatibility problem, and the PR just updated the requirement without code modification, and all test passed.

I also successfully used the docker-compose in the situation of our PR.

bmw:
That might be fine, but I don't completely understand when `setuptools` verifies that dependencies are satisfied. It does sometimes and when it catches it, it crashes the program.

Another option would be to pin `docker-compose==1.25.0rc1` where this constraint on `requests` has been lifted.

bmw:
`docker-compose 1.25.0` came out last month and resolves the problems here. The package is no longer a requirement when running pebble though. I think we can either just make docker-compose an unconditional dependency or make something like `boulder` extras and make sure that's installed in the dev environment created by the `tools/venv*` scripts and relevant integration tests in CI.

I prefer the former for simplicity.

adferrand:
I agree, having docker-compose as a base dev requirement is the easiest solution. I upgraded the requirement to 1.25.0, no other transitive dependency bump is needed from what I see from the sources and pip behavior.

PR #7137, opened 2019-06-08
Add reference to the Standalone DNS Authenticator

siilike:
Updated documentation to add a reference to the Standalone DNS Authenticator, https://github.com/siilike/certbot-dns-standalone


siilike:
Thank you for your reply and sorry for not testing the change. Now the table is fixed.

Regarding #6646, it's easy enough to install and use it as a plugin, so don't worry about that, although would be great if it eventually will get merged.

siilike:
Thank you.

PR #7141, opened 2019-06-10
Fix dns-rfc2136 validation

jensalmer:
There is a bug in the dns-rfc2136 code that causes all renewals and issues to fail because of parameter order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Fixed issue with dns-rfc2136 validator
 
 Despite us having broken lockstep, we are continuing to release new versions of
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Fixed issue with dns-rfc2136 validator
```
bmw: Can we clarify this comment just a bit saying something like:

> Fixed bug causing the dns-rfc2136 plugin to fail that was introduced with support for specifying an authoritative base domain in 0.35.0.

bmw:
Since we completely reverted the commit adding the bug being fixed here, I'm going to close this PR, but again, feel free to open a PR adding the buggy code back with your patch if you'd like!

Thanks again for brining this issue to our attention so we could put out a release to fix it.

PR #7149, opened 2019-06-11
Bump needs-update threshold to 1 year.

bmw:
While I expect stale bot will close out 150 - 250 issues, that'll still leave us with 400+ open issues. My concern is that with a threshold of 6 months, most of these 400 issues will be in the same state 6 months from now and stale bot will annoy people by asking them if their issue is still valid too frequently.

Doubling the stale threshold to 1 year should mitigate this problem a bit I think.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Configuration for https://github.com/marketplace/stale
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
-daysUntilStale: 180
+daysUntilStale: 365
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.


bmw:
> what is the typical time for an issue that has not been responded to become irrelevant for the issuer or the maintainers (defining here an upper bound) ?

I think this should be our lower bound. Given the concerns about upsetting the Certbot community and stale bots in general, [my](https://opensource.eff.org/eff-open-source/pl/3izzzg1wcffx5munr7kdwp8dco) and [Max's](https://opensource.eff.org/eff-open-source/pl/wemarrhrqbri5myg1ob7quksdy) thoughts for this bot was to be quite conservative, especially initially. I think we should err on the side of being too conservative and correct as needed in the future.

Pinging an issue before we typically expect the issue to become irrelevant seems unnecessarily aggressive and typically would result in wasting the time of those involved in the issue. As the author of 177 open Certbot issues and being involved in many more than that, I thing that too frequent pings wastes developer's time too.

> In that extent, is there some rationale to say that one year is a better balance than 6 months?

Yes. To build off of my original post, I do not think that Certbot is moving fast enough that the majority of those 400+ issues will become invalidated in 6 months. You've been working on Certbot for 10 months and during that time, I do not think anything has changed that would cause significant sets of issues to become invalid. A year seems to me like a more realistic threshold where if we haven't seen any updates in that time, the issue may no longer be relevant.

> Maybe some other similar projects that also use a stale bot could give us a good value here?

Unfortunately, I haven't seen any repositories for bots anywhere near as conservative as ours. I don't think that necessarily means we're doing anything wrong, just that we're trying to be more conservative right now than most other projects using a stale bot are.

Finally, I'll add that we talked about this change on the weekly Certbot call this week before you joined and Max asked if people had any objections. Everyone said no.

PR #7154, opened 2019-06-13
Fix integration tests errors against Boulder in nightly tests

adferrand:
Currently integration tests against Boulder fail during nightly tests. See https://travis-ci.com/certbot/certbot/builds/115373954.

This is due to a failure to cleanup the workspace associated to the Boulder docker started during the integration tests. Indeed this docker compile several artifacts whose owner is root, and permissions are `0744`. These files are persisted in the workspace folder attached to the Docker.

Since tox is run as a non-root user (but this user still have access to the Docker daemon), everything works fine until the end of the test suite, when all resources are cleaned up. At this point, pytest fires a `PermissionError` when failing to delete these artifacts, return with a non-zero exit code, and so fail the build.

Since this situation could happen outside of the CI, I made appropriate corrections to allow the integration tests to be run as a non-root user, instead of changing Travis to execute tests as root user.

The correction is to add a step to the cleanup process: the deletion of these artifacts through an ad-hoc docker instance.

### DIFF
--- a/certbot-ci/certbot_integration_tests/utils/acme_server.py
+++ b/certbot-ci/certbot_integration_tests/utils/acme_server.py
@@ -99,6 +99,14 @@ def _construct_workspace(acme_type):
                 pass
             print('=> Finished tear down of {0} instance.'.format(acme_type))
 
+        if acme_type == 'boulder' and os.path.exists(os.path.join(workspace, 'boulder')):
+            # Boulder docker generates build artifacts owned by root user with 0o744 permissions.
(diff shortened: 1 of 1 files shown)


bmw:
Did the changes we made in https://github.com/certbot/certbot/pull/7073 cause this problem? If so, why? Removing that commit and running tests again, everything seems to be passing: https://travis-ci.com/certbot/certbot

adferrand:
No the changes in #7073 just made it visible. Before, the error was happening during the atexit handler call, and so was not participating in the exit_code value returned by pytest. Now it is.

PR #7156, opened 2019-06-14
[Windows|Linux] Use builtin Python proxy capabilities for Certbot-CI

adferrand:
This PR is a part of the actions necessary to make Certbot-CI work on Windows, in order to execute the integration tests on this platform.

I initially used the fully-fledged HTTP proxy [Traefik](https://docs.traefik.io/) to distribute HTTP challenges among several pytest nodes, and so parallelize the integration tests. Traefik for this purpose is overkill. We just want to redirect the ACME server to a pytest node depending on the `Host` header, and we use here a production-grade HTTP proxy for that.

However it was not a problem on Linux, as soon as you can have Docker, because this instance is deployed through it.

But this becomes a problem for Windows, where Docker is not available everywhere, very compelling on its setup, and limited by the implemented network drivers. See my comments here https://github.com/letsencrypt/pebble/pull/240 for more details.

Hopefully Python ships with everything needed to implement a simple HTTP proxy, with strictly what we need for the parallelization of integration tests.

This PR implements this kind of HTTP proxy, and remove the coupling to Traefik.

This PR has been tested successfully with integration tests on Pebble under Linux for Python 2.x and Python 3.x, and the proxy alone has been also tested successfully on Windows (no integration tests can be run for now on this platform).

PR #7157, opened 2019-06-14
[Windows|Linux] Launch integration tests on Pebble without Docker

adferrand:
This PR is a part of the actions necessary to make Certbot-CI work on Windows, in order to execute the integration tests on this platform.

Following #7156, this PR changes how the integration tests are setup against Pebble to not need Docker anymore.

As a reminder, one can check #7156 and letsencrypt/pebble#240 to see the rationale about why using Docker is a problem to run the integration tests on Windows.

Basically, this PR executes directly Pebble using its executable, since it is build using Go, and Go produces self-contained executable that can run without any installation on Linux and on Windows. During the integration tests setup, Certbot-CI will get the Pebble (and Challtestsrv) executables for the defined target version on the GitHub releases. The binaries are persisted on the filesystem, so it is not needed to download them again on the second integration tests execution. Nonetheless, we are talking about 20MB of executables.

Since the setup needs to hold a state, I also took this occasion to refactor the `acme_server`, in order to use on object oriented approach and improve the readability/maintainability.

Once this PR and #7156 are merged, Docker will not be needed anymore for the main integration tests usecase, that is to use Pebble.


adferrand:
This PR blocks on letsencrypt/pebble#240 integration, and the release of a new version of Pebble occuring after it.

For now the code is linked to my own fork of `letsencrypt/pebble`, that already integrates the GitHub releases, as a showcase for this PR.

adferrand:
Unblocking on letsencrypt/pebble#240, pebble v2.1.0 is working great here. Just need to restart the coverage jobs ...

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -11,176 +11,168 @@
...
-        sys.stderr.write(e.output)
-        raise
+    def _launch_process(self, command, cwd=os.getcwd(), env=None):
+        """Launch silently an subprocess OS command"""
+        if not env:
+            env = os.environ
+        with open(os.devnull, 'w') as null:
+            process = subprocess.Popen(command, stdout=null, stderr=subprocess.STDOUT, cwd=cwd, env=env)
+            self._processes.append(process)
+            return process
```
bmw: Aren't we closing the file descriptor that still may be being used by the process here? I'm open to other approaches, but maybe it makes sense to make a single `/dev/null` file descriptor in the class that can be shared between all processes and closed when `stop()` is called. What do you think?

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -11,176 +11,168 @@
...
-    def __init__(self, acme_xdist, start, server_cleanup):
-        self._proxy_process = None
-        self._server_cleanup = server_cleanup
-        self.acme_xdist = acme_xdist
-        self.start = start
+    def __init__(self, acme_server, nodes, proxy=True):
+        """
+        Create an ACMEServer instance.
+        :param acme_server: the type of acme server used (boulder-v1, boulder-v2 or pebble)
+        :param str[] nodes: list of node names that will be setup by pytest xdist
```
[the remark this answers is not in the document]
adferrand: Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -11,176 +11,168 @@
...
+        :param bool proxy: set to False to not start the Traefik proxy
```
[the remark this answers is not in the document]
adferrand: Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -11,176 +11,168 @@
...
-    (eg. challenges ports). ACMEServer is also a context manager, and so can be used to
-    ensure ACME server is started/stopped upon context enter/exit.
+    ACMEServer configure and handle lifecycle of an ACME CA server and an HTTP reverse proxy
+    instance, to allow parallel execution of integration tests against the unique http-01 port
+    expected by the ACME CA server.
+    Typically all pytest integration tests will be executed in this context.
+    ACMEServer gives access the acme_xdist parameter, listing the ports and directory url to use
+    for each pytest node. It exposes also start and stop methods in order to start the stack, and
+    stop it with proper resources cleanup.
+    An ACMEServer instance will be returned, giving access to the ports and directory url to use
```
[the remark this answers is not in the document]
adferrand: Indeed, it is a part left from the documentation refactoring. Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Yes, you are right. Not elegant at all.

In fact I though of this again, and there are two typical cases here, about what you want to do with the output generated by the subprocesses (acme server, http proxy ...):
1) In the context of integration tests using pytest, we want to silence this out. What is interesting is the output of Certbot, not the noise of the third parties, that we assume working correctly
2) In the context of `run_acme_server`, either to test Certbot or troubleshoot these third parties. In this case, we want the stdout of these processes to be displayed in the standard stdout, to be consumed as it is necessary

So, since `ACMEServer` is a context manager with everything needed for setup/teardown the context, I implemented a stdout alongside its lifecycle, to be used by all suprocesses. Its nature depends of a `stdout` parameter given to the `ACMEServer` constructor. If `True`, this stdout is sys.stdout, and we see the processes activity in the standard stdout. If `False`, we use a descriptor on `/dev/null` that is opened/closed accordingly in the context manager at entry and exit context.

For 1), `stdout` is False, for 2) it is `True`.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -183,7 +183,7 @@ def main():
         raise ValueError('Invalid server value {0}, should be one of {1}'
                          .format(server_type, possible_values))
 
-    acme_server = ACMEServer(server_type, [], False)
+    acme_server = ACMEServer(server_type, [], http_proxy=False, stdout=True)
     process = None
```
bmw: Can we remove this unused variable and just change the body of `except KeyboardInterrupt` below to `pass`?

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -192,9 +192,6 @@ def main():
                   .format(acme_xdist['directory_url']))
             print('--> Press CTRL+C to stop the ACME server.')
 
-            docker_name = 'pebble_pebble_1' if 'pebble' in server_type else 'boulder_boulder_1'
-            process = subprocess.Popen(['docker', 'logs', '-f', docker_name])
```
bmw: What do you think about adding code like this to the case when `ACMEServer` is used with boulder and stdout is printed? Currently in this PR, `run_acme_server` produces no additional output once boulder has been started.

adferrand: Are you sure you are on the latest version of my PR? I pass now the `docker-compose up` through `self._launch_process`, and so see the GIT, docker-compose and docker processes output in my terminal when running `run_acme_server boulder-v2` for instance.

adferrand: Oh, I got it. In fact you will not get more log from it by running `docker-compose logs`. All dockers logs are already displayed by `docker-compose up` since it is not run in detached mode.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Fixed.

certbot-ci/certbot_integration_tests/utils/acme_server.py
bmw: Is that right?

In one terminal, I ran `run_acme_server boulder-v2`. The last output it ever printed while it was running was:
```
boulder_1    | GOBIN=/go/src/github.com/letsencrypt/boulder/bin GO111MODULE=on go install -mod=vendor  ./...
=> Finished boulder instance deployment.
--> Instance of http://localhost:4001/directory is running, directory URL is http://localhost:4001/directory
--> Press CTRL+C to stop the ACME server.
```
In a second terminal I ran:
```
docker logs -f boulder_boulder_1
```
which continued to produce output as I ran Certbot against Boulder.

bmw:
When playing with this a little more I hit a small issue. it seems the `docker-compose` process returns before it has finished shutting down the Docker instances. As a result, the output I got was:
```
^C=> Tear down the test infrastructure...
Gracefully stopping... (press Ctrl+C again to force)
  File "/home/bmw/src/certbot/venv/bin/run_acme_server", line 11, in <module>
    load_entry_point('certbot-ci', 'console_scripts', 'run_acme_server')()
  File "/home/bmw/src/certbot/certbot-ci/certbot_integration_tests/utils/acme_server.py", line 195, in main
    time.sleep(3600)
  File "/home/bmw/src/certbot/certbot-ci/certbot_integration_tests/utils/acme_server.py", line 85, in __exit__
    self.stop()
  File "/home/bmw/src/certbot/certbot-ci/certbot_integration_tests/utils/acme_server.py", line 75, in stop
    shutil.rmtree(self._workspace)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 250, in rmtree
    os.remove(fullname)
OSError: [Errno 13] Permission denied: '/tmp/tmpBYsy4Q/boulder/.gocache/2e/2e396bd74e880099564450c53fc063100001dddd92c1be7c68bd45a8e89392d6-d'
Stopping boulder_boulder_1   ... done
Stopping boulder_bhsm_1      ... done
Stopping boulder_bmysql_1    ... done
```
and `/tmp/tmpBYsy4Q/boulder/` still exists when the process exits.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: This is very surprising. I always use docker-compose up without detached mode to get the logs when developing, I never saw a difference with docker-compose logs.

I will check tomorrow.

adferrand:
This also surprise me a lot, for two reasons. First, wait should not return before the process exit. Second, there is an action, in case of Boulder, that delete the files with write access only for root directly from a Docker. So then shutil should only find files that can be suppressed by a normal user. 

Will check that also tomorrow.

adferrand:
About https://github.com/certbot/certbot/pull/7157#issuecomment-506070952, I figured out what was happening.

I forgot that under Linux, `SIGINT` are not handled by context manager out of the box. See https://www.python.org/dev/peps/pep-0419/. We need to add an handler explicitly for the signal, in order to invoke normal exiting of the Python process, that will trigger the context manager exiting in the while.

This is done with https://github.com/certbot/certbot/pull/7157/commits/3fa9622d7f055ac25ee445b159ce76b98cd0a241

On Windows, the break signal is properly handled by the context manager out of the box.

certbot-ci/certbot_integration_tests/utils/acme_server.py
bmw: Just to make sure we're not both sitting here waiting for the other to respond, you're still planning on looking into this right?

adferrand: Yes, I am planning to do this.

adferrand: So, I confirm that was is displayed by a `docker-compose up` in foreground is the same than a `docker logs -f boulder_boulder_1̀`. Which make sense, since both are supposed to display the stdout + stderr generated by the PID 1 of each Docker.

In fact Boulder is generated some logs after it started. Here is for instance, from `docker-compose up` through the `run_acme_server`, the output starting to be generated in the terminal while running a `certonly` certbot command against this local Boulder instance.

```
boulder_1    | All servers running. Hit ^C to kill.
boulder_1    | pebble-challtestsrv - 2019/06/30 10:55:22 Set default IPv4 address for DNS A queries to "10.77.77.1"
=> Finished boulder instance deployment.
--> Instance of http://localhost:4001/directory is running, directory URL is http://localhost:4001/directory
--> Press CTRL+C to stop the ACME server.
boulder_1    | I105947 boulder-va [AUDIT] Attempting to validate HTTP-01 for "test.example.com" with GET to "http://test.example.com/.well-known/acme-challenge/pqtx32L6Q2asy6EJmCuuXbpgOIRSsKRqK9y37jhaT2E"
boulder_1    | I105947 boulder-va [AUDIT] Checked CAA records for test.example.com, [Present: false, Account ID: 1, Challenge: http-01, Valid for issuance: true] Records=null
boulder_1    | I105947 boulder-va [AUDIT] Validation result JSON={"ID":"47jmyIhiPq74M6T6Ce5j1zHONEcKzCHUArr5SpSpaFQ","Requester":1,"Hostname":"test.example.com","Challenge":{"id":3,"type":"http-01","status":"valid","token":"pqtx32L6Q2asy6EJmCuuXbpgOIRSsKRqK9y37jhaT2E","keyAuthorization":"pqtx32L6Q2asy6EJmCuuXbpgOIRSsKRqK9y37jhaT2E.57dQskmRZbPyp-DKTD6t9pgj_Gx09oib1NlvwP2qY5Q","validationRecord":[{"url":"http://test.example.com/.well-known/acme-challenge/pqtx32L6Q2asy6EJmCuuXbpgOIRSsKRqK9y37jhaT2E","hostname":"test.example.com","port":"5002","addressesResolved":["10.77.77.1"],"addressUsed":"10.77.77.1"}]},"ValidationLatency":0.012}
boulder_1    | I105947 boulder-va Validations: {ID:47jmyIhiPq74M6T6Ce5j1zHONEcKzCHUArr5SpSpaFQ Identifier:{Type: Value:} RegistrationID:1 Status: Expires:<nil> Challenges:[] Combinations:[] Wildcard:false V2:false}
```

bmw: Huh. That is indeed what [the docs](https://docs.docker.com/compose/reference/up/) say. When using docker-compose 1.17.1, this works for me, however, when I try using the latest version from PyPI I get no output from `docker-compose up`.

Do you see the same behavior? I don't see anything about this in [Docker Compose's changelog](https://github.com/docker/compose/blob/master/CHANGELOG.md), but is there a bug/undocumented change in behavior here?

adferrand: I did not have the time to check today, but maybe handling and solving this issue would be more suitable in #7120?

bmw: I created https://github.com/certbot/certbot/issues/7216 to track this.

adferrand:
@bmw yes you are right. I was not clear here: what is not handled properly is the entering and exiting context in a context manager. If a `SIGINT` is received during `__enter__`, then `__exit__` will not be called, and so the dockers and workspace are not properly cleaned.

So yes, I see the behavior described in https://github.com/certbot/certbot/pull/7157#issuecomment-506070952 if I press CTRL+C during the starting of the Boulder docker. But once it is started up and so the `__enter__` function is complete, pressing CTRL+C will trigger properly the context exit.

Thinking of it again to how properly handle that, I came back to `try/except` logic instead of adding a new techno here, with `signal`. So now if any exception (so including the `KeyboardInterrupt` raised by the SIGINT signal) will trigger `stop`. If no exception is raised, we will finish the enter context, so `__exit__` will be called appropriately.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -58,7 +63,10 @@ def stop(self):
         print('=> Tear down the test infrastructure...')
         try:
             for process in self._processes:
-                process.terminate()
+                try:
+                    process.terminate()
+                except OSError:
```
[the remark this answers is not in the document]
adferrand: In some pathological situations (if you press CTRL+C just after docker-compose has been invoked), you can end in a situation where the `process` Python object is created, but the associated system process has never been started, so there is no PID and so `terminate()` fail.

I narrowed the scope of the `OSError` catched and added a comment about this.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -72,7 +80,11 @@ def stop(self):
                                                 'alpine', 'rm', '-rf', '/workspace/boulder'])
                 process.wait()
         finally:
-            shutil.rmtree(self._workspace)
+            try:
+                def _onerror(*_args): pass
+                shutil.rmtree(self._workspace, onerror=_onerror)
+            except OSError:
```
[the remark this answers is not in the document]
adferrand: Indeed, it is superfluous. I kept only the `onerror` handler.

adferrand:
Indeed @bmw, I forgot to push things. It is done now.

certbot-ci/certbot_integration_tests/utils/acme_server.py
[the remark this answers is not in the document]
adferrand: Testing again with the new approach I just pushed, this is not needed anymore since the teardown is always invoked.

PR #7159, opened 2019-06-17
Removing insecure protocols

SubaruSama:
Updating the option ssl_protocols and removing:
- TLSv1.1
and adding:
- TLSv1.3
keeping the TLSv1.2 as a option.

Be sure to edit the `master` section of `CHANGELOG.md` to include a description
of the change being made in this PR.

You are also welcome to add your name to `AUTHORS.md` if you like.


bmw:
Thanks for the PR, but we follow Mozilla's recommendations at https://mozilla.github.io/server-side-tls/ssl-config-generator/ which has not made this change yet. I expect they will soon though at which point we will update our config to follow their new recommendations.

PR #7163, opened 2019-06-17
Update Nginx conf file to match Mozilla's security recommendations

### LINKED ISSUE #7089: Update SSL session cache size to match Mozilla recommendations
This is a followup from the research issue at #6903.

Ideally, https://github.com/mozilla/server-side-tls/issues/198 is resolved and Mozilla updates their recommendations. If not, I think we should update our value in https://github.com/certbot/certbot/blob/master/certbot-nginx/certbot_nginx/options-ssl-nginx.conf.

Exactly what these values should be is up for discussion, however, nginx's default timeout of 5 minutes seems like a reasonable place to start to me. I don't know of the top of my head how I think the cache should be configured.

sydneyli:
Fixes #7089 .

Be sure to edit the `master` section of `CHANGELOG.md` to include a description
of the change being made in this PR.

You are also welcome to add your name to `AUTHORS.md` if you like.

### DIFF
--- a/certbot-nginx/certbot_nginx/constants.py
+++ b/certbot-nginx/certbot_nginx/constants.py
@@ -24,7 +24,7 @@ UPDATED_MOD_SSL_CONF_DIGEST = ".updated-options-ssl-nginx-conf-digest.txt"
 
 SSL_OPTIONS_HASHES_NEW = [
     '63e2bddebb174a05c9d8a7cf2adf72f7af04349ba59a1a925fe447f73b2f1abf',
-    '5f094d6f3f99401865e8a48e5c43e98647a77174dc451fa59e332ad25bd030e7',
+    '2901debc7ecbc10917edd9084c05464c9c5930b463677571eaf8c94bffd11ae2',
 ]
 """SHA256 hashes of the contents of versions of MOD_SSL_CONF_SRC for nginx >= 1.5.9"""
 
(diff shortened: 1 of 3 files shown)


bmw:
So while reviewing this, I started to think about ways this could go wrong and started wondering what would happen on resource constrained systems like our tens of thousands of Raspberry Pi users. Playing with Nginx a bit (and values much larger than 50MB), Nginx crashes with a message like:
```
nginx: [alert] mmap(MAP_ANON|MAP_SHARED, 5242880000) failed (12: Cannot allocate memory)
```
These extra 49MB certainly wouldn't be a problem for almost all users, but I worry that it may be a problem for some and that it'll be hard to debug.

Luckily, Mozilla has a preview of their new SSL config generator live at https://ssl-config.mozilla.org/#server=nginx&server-version=1.17.0&config=intermediate and they've reduced this value to 10M. As far as I know, the real site should be updated any day now.

If you disagree, let's chat, but I think we should wait for Mozilla to update their recommendations which will likely result in a lower cache size and make this change then. Sorry for the false start here!

sydneyli:
Nice catch, and thanks for doing the research!

So let's put this PR on hold for now until upstream updates.

sydneyli:
Closing for the time being, until Mozilla updates their recs.

sydneyli:
Re-opened now that new version has been released: https://wiki.mozilla.org/Security/Server_Side_TLS

and cache size has been reduced to `10m`!

Issue #7125, opened 2019-06-06
certbot-auto and cPanel httpdconf


CanadianJeff:
I use cPanel/WHM and ran certbot. All my certs installed and are working
I confirmed with ```sudo apachectl -S``` 

certbot-auto put ```Include conf/httpd-le-ssl.conf``` in ```conf/httpd.conf``` however if on a cPanel server you run ```/scripts/rebuildhttpdconf``` the Include gets wiped out causing hours of frustration they suggest you put your Include statements in ```conf.d/includes/pre_main_global.conf``` so I did this on my server and it works perfectly

My ```pre_main_global.conf``` looks like this
```
# Supplemental configuration
#
# The configuration files in the conf/extra/ directory can be 
# included to add extra features or to modify the default configuration of 
# the server, or you may simply copy their contents here and change as 
# necessary.

# Server-pool management (MPM specific)
#Include conf/extra/httpd-mpm.conf

# Multi-language error messages
#Include conf/extra/httpd-multilang-errordoc.conf

# Fancy directory listings
#Include conf/extra/httpd-autoindex.conf

# Language settings
#Include conf/extra/httpd-languages.conf

# User home directories
#Include conf/extra/httpd-userdir.conf

# Real-time info on requests and configuration
#Include conf/extra/httpd-info.conf

# Virtual hosts
#Include conf/extra/httpd-vhosts.conf

# Local access to the Apache HTTP Server Manual
#Include conf/extra/httpd-manual.conf

# Distributed authoring and versioning (WebDAV)
#Include conf/extra/httpd-dav.conf

# Various default settings
#Include conf/extra/httpd-default.conf

# Secure (SSL/TLS) connections
# Listen for virtual host requests on all IP addresses
#NameVirtualHost *:443

# Go ahead and accept connections for these vhosts
# from non-SNI clients
#SSLStrictSNIVHostCheck off
Include conf/httpd-le-ssl.conf
#Include conf/extra/httpd-ssl.conf
#
# Note: The following must must be present to support
#       starting without SSL on platforms with no /dev/random equivalent
#       but a statically compiled-in mod_ssl.
#
```

alexzorin:
The short story is that you should never use the Certbot Apache plugin (authenticator or installer) on a cPanel server. It is dangerous and does not mesh with cPanel's management of virtual hosts and their certificates.

Also relevant: https://github.com/certbot/certbot/issues/6576

[...]

Issue #7171, opened 2019-06-20
Sending non-empty JSON body on validation check requests


dopey:
https://community.letsencrypt.org/t/acme-spec-asks-for-empty-json-blob-in-challenge-validation-request/96252/2

The acme spec (https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.7.5.1) requires that clients send an empty JSON body ({}) for client validation checks. Certbot appears to be sending a non-empty payload 

```
{
  "resource": "challenge",
  "type": "http-01"
}
``` 
which is not in accordance with the acme spec and will illicit malformed request errors from servers that are strictly adhering to the ACME spec.

bmw:
Thanks for the issue @dopey. I agree we should fix this.

Currently it's slightly complicated by the fact we're maintaining our support for "ACMEv1", but we could still do this for "ACMEv2". We'd accept a well written PR for this if you're interested in working on it.

This is related to https://github.com/certbot/certbot/issues/5708.

cpu:
@bmw FWIW As of [Pebble v2.1.0](https://github.com/letsencrypt/pebble/releases/tag/v2.1.0) you can use `pebble -strict` to verify the fix for this issue.

adferrand:
I increase this issue priority, since actually, we are not compliant against Pebble v2.1.0.

bmw:
This was fixed in #7176.

PR #7176, opened 2019-06-23
Fix acme compliance to RFC 8555

adferrand:
This PR is an alternative to #7125.

Instead of disabling the strict mode on Pebble, this PR fixes the JWS payloads regarding RFC 8555 to be compliant, and allow certbot to work with Pebble v2.1.0+.


bmw:
I'm going to be in meetings for the next few hours, but we should get our tests fixed soon. Thanks for writing multiple solutions for this so quickly.

To provide an initial comment on this PR though, I'm curious if there's not a more fundamental change we can make here. https://github.com/certbot/certbot/issues/7171 is the same issue as https://github.com/certbot/certbot/issues/5708 except #5708 is about every ACME object and #7171 also has the problem of the `type` field (which I don't remember how it works off the top of my head). Is there a change we could make to remove the `resource` field for all ACME objects? The `type` field for challenge objects may have to be treated separately.

If so, I'm personally leaning towards #7175 which we can then revert once we've had time to address the deeper issue without rushing to get tests fixed. What do you think?

adferrand:
TL;DR

Yes, this was my intention by opening the two pull requests: offering a short term solution and discuss about the long term solution. I agree you should merge #7175 right now.

adferrand:
So for the long-term solution.

I was seeing basically two options:
1) What I did here, meaning restructure afterwards the JWS token in an isolated method that is unit testable and that follows strictly the controls done on Pebble-side.
2) What I initially tried, similarly to `keyAuthorization` in https://github.com/certbot/certbot/pull/6758, that consists in changing the josepy objects beforehand to dump the JWS token in conformity with RFC 8555.

Then I ended to understand that removing some fields from JSON objects would require to modify in dozen places implementations of these objects. I though it was utterly complex, and came across #1297.

Indeed I thought it last time and agree with what you said in this issue. Challenges, and also Messages related classes are too much complex. I think we should start by refactoring this, before trying to have a separate representation of the JWS tokens between ACME v1 and standardized ACME v2. Otherwise we will just continue to increase the spaghetti code in `acme` module.

However, this refactoring is significant. In the mean time, I think the solution I propose here is an acceptable middle-term solution, that provides to Certbot a full compliance to the spec, while keeping the code isolated, unit testable and documented in one place, as I said at the beginning of this comment.

bmw:
Taking another look at this, I still prefer the 2nd approach you describe if we can reasonably pull it off. I agree it's more difficult and results in more modifications, but it fixes the individual problems (https://github.com/certbot/certbot/issues/5708 and https://github.com/certbot/certbot/issues/7171) in the classes themselves rather than writing a separate catch all method elsewhere that cleans things up in different ways depending on the object it is given.

I think I may be able to help design how to do this though. What do you think of something like this?

First, in a separate file named something like `mixins.py`, we create something like `LEACMEMixIn`. This mixin defines a getter and setter property like `le_acme_version` that works similar to https://github.com/certbot/certbot/pull/6758. The value defaults to 1.

Second, in the same file inheriting from that (or if a mixin being part of an inheritance tree drives you nuts, the above functionality could be written into this class directly instead), we define `ResourceMixIn` that provides a `fields_to_partial_json` method that looks like:
```
jobj = super(ResourceMixIn, self).fields_to_partial_json()
if self.le_acme_version == 2:
    jobj.pop('resource', None)
return jobj
```

Third, every class that defines a resource field inherits from this mixin. It needs to be listed as the first parent class to ensure this new `fields_to_partial_json` object is used.

Fourth, in `acme.challenges.ChallengeResponse`, we provide `to_partial_json` like:
```
jobj = super(ChallengeResponse, self).to_partial_json()
if self.le_acme_version == 2:
    jobj.pop('type', None)
return jobj
```

Fifth, in `acme/client.py`, write code like:
```
if isinstance(obj, LEACMEMixIn):
    obj.le_acme_version = acme_version
```

Sixith, profit...or more responsibly write tests.

What do you think? It's certainly more complex, but I think it solves the problem on the objects themselves and provides a pattern for how we could address similar problems in the future.

adferrand:
cc @bmw 

As a technology enthusiast I am quite motivated to use mixins/traits, since it is a pattern that is not easy to use in general. Downside: it is a pattern that is not easy to use in general ... Using this can make your code very hard to maintain very quickly, so it has to be considered very carefully.

Here I think it has the advantage, indeed, to implement the corrections right into the classes instead of externally, and so in the most elegant way (few LOC, low cyclomatic complexity).

So I modified this PR to use your proposition. It works well, and a passing full test suite is available here: https://travis-ci.com/certbot/certbot/builds/142551566

However I consider this approach, and mine before, as hacks to circumvent limitations around josepy principles and their usage in acme module. Current design is tending to create spaghetti code, and we should think of an overall refactoring when possible.

acme/acme/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+        if key == 'le_auto_version':
+            # Required for @property to operate properly. See comment above.
+            object.__setattr__(self, key, value)
+        else:
+            super(VersionedLEACMEMixin, self).__setattr__(key, value)  # pragma: no cover
+
+
+class ResourceMixin(VersionedLEACMEMixin):
+    """
+    This mixin allows to generate a RFC8555 compliant JWS payload
```
bmw:
```suggestion
    This mixin generates a RFC8555 compliant JWS payload
```
or
```suggestion
    This mixin allows generation of a RFC8555 compliant JWS payload
```

acme/acme/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+    by removing the `resource` field if needed (eg. ACME v2 protocol).
+    """
+    def to_partial_json(self):
```
bmw: nit: While maybe not intentionally part of the public API `josepy.JSONDeserializable` defines the method `fields_to_partial_json`. In my comment at https://github.com/certbot/certbot/pull/7176#issuecomment-567718597, I suggested we overrode `fields_to_partial_json` since `to_partial_json` calls it, but that's making assumptions about the internals of  `josepy.JSONDeserializable.to_partial_json`.

Instead, what do you think about overriding both methods here and ensuring `resource` is popped in both cases. Bonus points for any code reuse if it's sane to do so.

acme/acme/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+
+        raise AttributeError('Method to_partial_json() is not implemented.')  # pragma: no cover
+
+
+class TypeMixin(VersionedLEACMEMixin):
+    """
+    This mixin allows to generate a RFC8555 compliant JWS payload
+    by removing the `type` field if needed (eg. ACME v2 protocol).
+    """
+    def to_partial_json(self):
```
bmw: Same comment as above about `fields_to_partial_json`.

certbot/setup.py
```diff
@@ -36,7 +36,7 @@ def read_file(filename, encoding='utf8'):
 # specified here to avoid masking the more specific request requirements in
 # acme. See https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
-    'acme>=0.40.0',
+    'acme>=1.1.0.dev0',
```
bmw:
```suggestion
    'acme>=1.4.0.dev0',
```

certbot-nginx/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
-acme[dev]==1.0.0
+-e acme[dev]
```
bmw: We should also update the nginx plugin's `setup.py` file since we're not going to be testing against 1.0 anymore.

acme/tests/jws_test.py
```diff
@@ -62,5 +64,33 @@ def test_jwk_serialize(self):
         self.assertEqual(jws.signature.combined.jwk, self.pubkey)
 
 
+class JWSPayloadRFC8555Compliant(unittest.TestCase):
```
bmw: nit: Since these tests don't test the code in `acme.jws`, can we move the tests to `challenge_test.py` and `messages_test.py` respectively?

acme/acme/mixins.py
```diff
@@ -0,0 +1,55 @@
+"""Useful mixins for Challenge and Resource objects"""
+
+
+class VersionedLEACMEMixin(object):
+    """This mixin allows to store the current ACME version as a property"""
+    @property
+    def le_auto_version(self):
```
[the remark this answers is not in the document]
adferrand: Weeeell ... yes. I have totally been influenced by my work on Python 3.4->3.6 migration that occured in certbot-auto.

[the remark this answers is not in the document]
adferrand: So, renamed into `le_acme_version`, and the same for the internal property.

acme/acme/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+        """Define the version of ACME protocol to use"""
+        return getattr(self, '_le_auto_version', 1)
```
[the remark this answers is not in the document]
adferrand: Done.

acme/acme/mixins.py
adferrand: I tried to do so, but modifying `fields_to_partial_json` has a lot more implications. In particular, every call that implies `josepy.JSONDeserializable._check_required()` will fail because `resource` or `type` is missing in the generated json object while it is defined as a property at class level.

Since it is sufficient for the purpose of this PR (fixing compliance in regards to the ACME server), I would like to avoid to modify `fields_to_partial_json` and avoid to add more workarounds here.

acme/acme/mixins.py
adferrand: See above.

certbot-nginx/local-oldest-requirements.txt
adferrand: Done.

acme/tests/jws_test.py
adferrand: Done. About the message, I use a real class instead of a mock class to be closer to real cases.

acme/acme/mixins.py
bmw: > In particular, every call that implies `josepy.JSONDeserializable._check_required()` will fail because `resource` or `type` is missing in the generated json object while it is defined as a property at class level.

Is this the problem? I believe `_check_required` is only called when converting JSON to the class and not when dumping the class to JSON.

Playing with your code before the revert commit, I was able to get all unit tests to pass by fixing a small typo:
```
diff --git a/acme/acme/mixins.py b/acme/acme/mixins.py
index d4d78636c..1d76280c1 100644
--- a/acme/acme/mixins.py
+++ b/acme/acme/mixins.py
@@ -58,7 +58,7 @@ class TypeMixin(VersionedLEACMEMixin):
 def _safe_jobj_compliance(instance, jobj_method, uncompliant_field):
     if hasattr(instance, jobj_method):
         jobj = getattr(instance, jobj_method)()
-        if instance.le_acme_version:
+        if instance.le_acme_version == 2:
             jobj.pop(uncompliant_field, None)
         return jobj
```
If you still don't like this approach for some reason though, another thing we could do which would satisfy me is to deprecate `fields_to_partial_json` in `josepy` by having the method send a deprecation warning. Probably the easiest way to do that is to just make a private `_fields_to_partial_json` method which everything else (including the deprecated `fields_to_partial_json` method) can still call.

My overall concern with my comment here is just leaving the problems fixed by this PR unresolved for some methods on the objects. Removing the problematic methods also fixes that. If we take this approach though, I'd kind of like the `josepy` deprecation to land before we land this PR to make sure we actually do that work.

adferrand: You are right. I misinterpretated the error.

adferrand:
Full tests suite is here: https://travis-ci.com/github/certbot/certbot/builds/153099124

acme/acme/mixins.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def fields_to_partial_json(self):
+        """See josepy.JSONDeserializable.fields_to_partial_json()"""
+        return _safe_jobj_compliance(super(ResourceMixin, self),
+                                     'fields_to_partial_json', 'resource')
+
+
+class TypeMixin(VersionedLEACMEMixin):
+    """
+    This mixin allows to generate a RFC8555 compliant JWS payload
```
bmw: Equivalent to https://github.com/certbot/certbot/pull/7176#discussion_r391834371. Sorry I missed it in my last review.
```suggestion
    This mixin generates a RFC8555 compliant JWS payload
```
or
```suggestion
    This mixin allows generation of a RFC8555 compliant JWS payload
```

acme/tests/messages_test.py
```diff
@@ -467,5 +468,18 @@ def test_to_partial_json(self):
...
+class JWSPayloadRFC8555Compliant(unittest.TestCase):
+    """Test for RFC8555 compliance of JWS generated from resources/challenges"""
+    def test_message_payload(self):
+        from acme.messages import NewAuthorization
+
+        new_order = NewAuthorization()
+        new_order.le_acme_version = 2
+
+        jobj = new_order.json_dumps(indent=2).encode()
+        # RFC8555 states that challenge requests must have an empty payload.
```
bmw: I think you accidentally switched the comments in the two tests you added as they both seem more appropriate in the other location.

If so, I'd change this comment to say "...challenge responses..." when it's moved to the other file.

acme/acme/mixins.py
```diff
@@ -25,18 +25,18 @@ def __setattr__(self, key, value):
...
-            jobj = super(ResourceMixin, self).to_partial_json()  # type: ignore
-            if self.le_auto_version == 2:
-                jobj.pop('resource', None)
-            return jobj
+        return _safe_jobj_compliance(super(ResourceMixin, self),
+                                     'to_partial_json', 'resource')
 
-        raise AttributeError('Method to_partial_json() is not implemented.')  # pragma: no cover
+    def fields_to_partial_json(self):
+        """See josepy.JSONDeserializable.fields_to_partial_json()"""
```
bmw:
```suggestion
        """See josepy.JSONObjectWithFields.fields_to_partial_json()"""
```

acme/acme/mixins.py
```diff
@@ -46,10 +46,20 @@ class TypeMixin(VersionedLEACMEMixin):
...
-        if hasattr(super(TypeMixin, self), 'to_partial_json'):
-            jobj = super(TypeMixin, self).to_partial_json()  # type: ignore
-            if self.le_auto_version == 2:
-                jobj.pop('type', None)
-            return jobj
+        return _safe_jobj_compliance(super(TypeMixin, self),
+                                     'to_partial_json', 'type')
+
+    def fields_to_partial_json(self):
+        """See josepy.JSONDeserializable.fields_to_partial_json()"""
```
bmw:
```suggestion
        """See josepy.JSONObjectWithFields.fields_to_partial_json()"""
```

acme/tests/messages_test.py
adferrand: Fixed.

adferrand:
Aaaand changelog added.

PR #7181, opened 2019-06-25
Refactor augeas_configurator.py functionality to configurator.py and parser.py accordingly.

joohoi:
This pull request moves the functionality within `AugeasConfigurator` that previously existed as a parent class of `ApacheConfigurator` to `ApacheConfigurator` and `ApacheParser` accordingly.

Most of the methods were moved as-is, and one (`recovery_routine()`) was completely removed. Few of the methods had to be split between the configurator and parser, good example of this is `save()`.

The Augeas object now lives completely within the `ApacheParser`.

I wasn't sure if I should have added changelog entry because no functionality has been affected.


certbot-apache/certbot_apache/parser.py
```diff
@@ -41,11 +43,21 @@ def __init__(self, aug, root, vhostroot=None, version=(2, 4),
         # issues with aug.load() after adding new files / defines to parse tree
         self.configurator = configurator
 
+        # Initialize augeas
+        self.aug = None
+        self.save_notes = ""
```
bmw: Since `save_notes` are used with the reverter for the purposes of describing a checkpoint, what do you think about moving this string to the configurator object?

certbot-apache/certbot_apache/augeas_configurator.py
```diff
@@ -1,207 +0,0 @@
...
-        """Revert all previously modified files.
-
-        Reverts all modified files that have not been saved as a checkpoint
-
-        :raises .errors.PluginError: If unable to recover the configuration
-
-        """
-        super(AugeasConfigurator, self).recovery_routine()
-        # Need to reload configuration after these changes take effect
-        self.aug.load()
```
bmw: For the sake of robustness, I think we still want to call `aug.load()` after calling `recovery_routine()`. Since `recovery_routine()` is part of the IInstaller interface, I wouldn't want us making a change to Certbot core down the road where we expect continuing to use the installer after calling `recovery_routine()` to break things here.

If we want to avoid parsing the config twice on initial startup, we could make a separate method to use in `configurator.prepare()` or change the behavior of the function if `self.parser is None`.

certbot-apache/certbot_apache/tests/augeas_configurator_test.py
```diff
@@ -63,23 +63,9 @@ def test_finalize_save(self):
 
         self.assertTrue(mock_finalize.is_called)
 
-    def test_recovery_routine(self):
```
bmw: If we add an implementation of `recovery_routine()` to the configurator, I think we should add at least this test back if not both.

certbot-apache/certbot_apache/tests/augeas_configurator_test.py
```diff
@@ -33,12 +33,12 @@ def test_bad_parse(self):
         self.config.parser.parse_file(os.path.join(
```
bmw: I think we should rename this file and update comments and names. Maybe something like `configurator_reverter_test`?

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -80,15 +81,14 @@ def test_prepare_version(self, mock_exe_exists, _):
         self.assertRaises(
             errors.NotSupportedError, self.config.prepare)
 
-    @mock.patch("certbot_apache.parser.ApacheParser")
     @mock.patch("certbot_apache.configurator.util.exe_exists")
-    def test_prepare_old_aug(self, mock_exe_exists, _):
+    def test_prepare_old_aug(self, mock_exe_exists):
```
bmw: I think we should move this test to `parser_test.ParserInitTest` and check that the class blows up during `__init__` if `check_aug_version` fails. It'd simplify this test a bit.

certbot-apache/certbot_apache/tests/augeas_configurator_test.py
```diff
@@ -33,12 +33,12 @@ def test_bad_parse(self):
...
             self.config.parser.root, "conf-available", "bad_conf_file.conf"))
         self.assertRaises(
-            errors.PluginError, self.config.check_parsing_errors, "httpd.aug")
+            errors.PluginError, self.config.parser.check_parsing_errors, "httpd.aug")
```
bmw: I think we should move this test to the parser tests.

certbot-apache/certbot_apache/tests/augeas_configurator_test.py
```diff
@@ -33,12 +33,12 @@ def test_bad_parse(self):
...
 
     def test_bad_save(self):
         mock_save = mock.Mock()
         mock_save.side_effect = IOError
-        self.config.aug.save = mock_save
+        self.config.parser.aug.save = mock_save
```
bmw: If we want this to be a test on the configurator, I think this should be:
```suggestion
        self.config.parser.save = mock_save
```

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -50,7 +50,7 @@ def mocked_deploy_cert(*args, **kwargs):
         self.config.deploy_cert = mocked_deploy_cert
         return self.config
 
-    @mock.patch("certbot_apache.configurator.ApacheConfigurator.init_augeas")
+    @mock.patch("certbot_apache.parser.ApacheParser.init_augeas")
```
bmw: Can't this mock be removed now that parser initialization happens lower?

certbot-apache/certbot_apache/tests/configurator_test.py
```diff
@@ -59,13 +59,14 @@ def test_prepare_no_install(self, mock_surgery, _init_augeas):
             self.assertRaises(errors.NoInstallationError, self.config.prepare)
             self.assertEqual(mock_surgery.call_count, 1)
 
-    @mock.patch("certbot_apache.augeas_configurator.AugeasConfigurator.init_augeas")
+    @mock.patch("certbot_apache.parser.ApacheParser.init_augeas")
```
bmw: If we make the change I suggested of moving the handling of this exception to `init_augeas`, I think we should move this test to the parser tests.

certbot-apache/certbot_apache/parser.py
```diff
@@ -77,6 +89,140 @@ def __init__(self, aug, root, vhostroot=None, version=(2, 4),
...
+        # set to remove this possibility. This is a known augeas 0.10 error.
+        save_paths = self.aug.match("/augeas/events/saved")
+
+        save_files = set()
+        if save_paths:
+            for path in save_paths:
+                save_files.add(self.aug.get(path)[6:])
+        return save_files
+
+    def ensure_augeas_state(self):
```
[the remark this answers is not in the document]
joohoi: Yeah, this is a specific need of Augeas parser and will be going away when we finalize the migration process. 

Additionally the calls to reverter are methods of `common.Installer` base class, and I think trying to wrap `ApacheParser` around it would cause even more confusion and redundancy.

certbot-apache/certbot_apache/parser.py
```diff
@@ -77,6 +89,140 @@ def __init__(self, aug, root, vhostroot=None, version=(2, 4),
...
+            self.configurator.save()
+
+    def save(self):
+        """Saves all changes to the configuration files.
+
+        save() is called from ApacheConfigurator to handle the parser specific
+        tasks of saving.
+
+        """
+        save_files = self.unsaved_files()
```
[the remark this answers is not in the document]
joohoi: Good idea! I think we can skip adding the snippet too - the parser.save() is not called from outside of configurator.save() so we can always pass the save_files list along!

certbot-apache/certbot_apache/parser.py
joohoi: Yeah, you are right. The `save_notes` is now moved to ApacheConfigurator.

certbot-apache/certbot_apache/augeas_configurator.py
joohoi: Yeah, it's now changed to handle that. 

Currently the `recovery_routine()` was only called within initialization (prepare) phase, so I thought of leaving it out. Making the check and acting accordingly is probably the best way forward.

certbot-apache/certbot_apache/tests/augeas_configurator_test.py
joohoi: Even though it tests and calls things through configurator, I moved this to parser_test as it's Augeas implementation specific and we'd end up moving it there eventually anyway. 

Also `self.config.parser.aug.save` was actually still correct because the `PluginError` is thrown by the save preflight check `ApacheParser.unsaved_files()`.

certbot-apache/certbot_apache/parser.py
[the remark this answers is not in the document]
bmw: That's another good point about their being inherited functionality from `common.Installer`. Leaving the reverter in the configurator probably makes more sense.

certbot-apache/certbot_apache/parser.py
```diff
@@ -188,18 +192,19 @@ def ensure_augeas_state(self):
         """
 
         if self.unsaved_files():
-            self.save_notes += "(autosave)"
+            self.configurator.save_notes += "(autosave)"
             self.configurator.save()
 
-    def save(self):
+    def save(self, save_files=None):
```
bmw: I think not calling `unsaved_files()` in this function is fine, but if we do that, I think `save_files` should be a required parameter.

certbot-apache/certbot_apache/tests/parser_test.py
```diff
@@ -21,6 +22,18 @@ def tearDown(self):
...
+    def test_bad_parse(self):
+        self.parser.parse_file(os.path.join(self.parser.root,
+                                            "conf-available", "bad_conf_file.conf"))
+        self.assertRaises(
+            errors.PluginError, self.parser.check_parsing_errors, "httpd.aug")
+
+    def test_bad_save(self):
+        mock_save = mock.Mock()
+        mock_save.side_effect = IOError
+        self.config.parser.aug.save = mock_save
```
bmw: nit: If we're moving this test to the parser, can we just call `parser.unsaved_files` directly?

certbot-apache/certbot_apache/tests/parser_test.py
```diff
@@ -30,6 +43,19 @@ def test_aug_version(self):
         self.parser.aug.match.side_effect = RuntimeError
         self.assertFalse(self.parser.check_aug_version())
 
+    def test_recovery_routine_reload(self):
```
bmw: I think we should keep this test in `configurator_reverter_test.py` as I think it's testing that the configurators implementation of `recovery_routine` properly asks the parser to reload the config.

certbot-apache/certbot_apache/tests/parser_test.py
```diff
@@ -30,6 +43,19 @@ def test_aug_version(self):
...
+        mock_load = mock.Mock()
+        self.config.parser.aug.load = mock_load
+        self.config.recovery_routine()
+        self.assertEqual(mock_load.call_count, 1)
+
+    @mock.patch("certbot_apache.parser.ApacheParser.init_augeas")
+    def test_prepare_no_augeas(self, mock_init_augeas):
```
bmw: Can we move this to `ParserInitTest` and test that `errors.NoInstallationError` is raised out of the parser's `__init__` method instead of calling back into the configurator?

certbot-apache/certbot_apache/parser.py
```diff
@@ -91,7 +90,12 @@ def __init__(self, root, vhostroot=None, version=(2, 4),
 
     def init_augeas(self):
         """ Initialize the actual Augeas instance """
-        import augeas
+
+        try:
+            import augeas
+        except ImportError:
```
bmw: I personally think it's OK to just disable coverage here. Actually causing the `ImportError` during the tests is more trouble than it's worth I think.

certbot-apache/certbot_apache/tests/parser_test.py
joohoi: My approach to this is that it specifically checks that augeas state is reloaded, and because of this I'd like to keep it within the parser tests. When we get to test against the new parser interface, another similar test to `configurator_test.py` should be created, checking that the interface `load()` is called.

certbot-apache/certbot_apache/parser.py
joohoi: Oh yeah, I'm starting to remember...

certbot-apache/certbot_apache/tests/parser_test.py
bmw: At the risk of harping on this very minor point too much, I might agree with keeping it in the parser tests if the fact we're calling `load` on Augeas had to do with the code in `parser.py` but I don't think it does. In `configurator.py`, we're reaching into the parser object and calling `parser.aug.load()` directly. I think the only code in `parser.py` that this function is indirectly testing is that the parser has an `aug` attribute which is tested plenty of other places.

I completely agree with you that we want a configurator test asserting that we properly reload the config using the correct interface after a recovery routine, but I think this is that test. We're checking that the implementation of `recovery_routine` in the configurator is properly reloading the parser with its currently existing interfaces. When the parser's interface changes, we'll need to update the code in `configurator.py` to use those interfaces as well as the tests for that configurator code. We'll also probably want a parser test in the future that tests that the common `load()` interface properly calls `load()` on Augeas, but I don't think this test needs to be coupled to the configurator code.

For those reasons, I personally think it makes sense to just have a single test for reloading during `recovery_routine` and for this test to exist in `configurator_reverter_test.py`.

If you disagree with me, I'd kind of like to know why, but I don't think it's worth blocking this PR on this minor point. If you still prefer to keep the test here after my additional explanation of how I'm thinking about this, let me know and I'll merge this PR as is.

joohoi: Yeah, I agree. I was too focused on keeping the `configurator_reverter_test.py` parser implementation agnostic (for the changes to come). It is the logical place for this test, and we'll have to refactor tests a lot when we're finally removing Augeas anyway. This test is now moved to the reverter test.

bmw:
I should have responded to this sooner but:

> I wasn't sure if I should have added changelog entry because no functionality has been affected.

The way I personally think about this is that if it's something Certbot users/packagers should know, we should include a changelog entry, otherwise, it's not necessary. Before when we were listing the packages changed in the changelog, listing internal changes was more important in my opinion to avoid saying that a package changed with no explanation of what was done.

PR #7182, opened 2019-06-25
[Windows] Security model for files permissions - STEP 3e

adferrand:
This PR implements the `filesystem.copy_ownership_and_apply_mode` method from #6497.

This method is used in two places in Certbot, replacing `os.chown`, to copy the owner and group owner from a file to another one, and apply to the latter the given POSIX mode.


certbot/compat/filesystem.py
```diff
@@ -129,6 +155,16 @@ def _analyze_mode(mode):
     }
 
 
+def _copy_win_ownership(src, dst):
+    security_src = win32security.GetFileSecurity(src, win32security.OWNER_SECURITY_INFORMATION)
+    user_src = security_src.GetSecurityDescriptorOwner()
+
+    security_dst = win32security.GetFileSecurity(dst, win32security.OWNER_SECURITY_INFORMATION)
+    security_dst.SetSecurityDescriptorOwner(user_src, False)
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/compat/filesystem.py
```diff
@@ -36,6 +36,32 @@ def chmod(file_path, mode):
         _apply_win_mode(file_path, mode)
 
 
+def copy_ownership_and_apply_mode(src, dst, mode, user=True, group=False):
```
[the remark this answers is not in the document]
adferrand: You are right, in particular considering that on the only two calls to this method, we override each time the default values. Corrected and docstring updated.

However, this raised another interrogation, see below.

adferrand:
@bmw while reviewing your comment in https://github.com/certbot/certbot/pull/7182#discussion_r297875247, I came across this portion of code again:
https://github.com/certbot/certbot/blob/141ad0e593018fdc8354bda31594981394e806a1/certbot/storage.py#L1107-L1112

This made me thing of a potential outcome that I think it is not a problem, but I wanted to check with you.

So on Linux this portion of code will apply to the new private key the most restrictive permissions between previous private key mode, and `0o740`. However on Windows as we say, mode returned by `os` gives largely unrelated mode with the DACL of a file. So this logic will end applying `0o740` independently from the actual DACL of the previous private key file.

I think this is not a problem, because this logic was mainly a fallback for Linux about a migration from a previous security pattern in certbot directories to a new one, so it does not apply to Windows since there is no implementation on production at this time. Also, since group is ignored on Windows and Certbot will be run as administrator, the mode `0o740` is actually always secure, as it gives no right to `Everybody`, and full rights to the actual user, that he would have anyway since he is an administrator.

At the minimum, I think an inline comment would be suitable to be clear about what is going on in this code when we are on Windows, and I am waiting you to give me a feedback about my analysis here :).

certbot/compat/filesystem.py
```diff
@@ -36,6 +36,32 @@ def chmod(file_path, mode):
...
+    # type: (str, str, int, bool, bool) -> None
+    """
+    Copy ownership (user and optionally group) from the source to the destination,
+    then apply given mode in compatible way for Linux and Windows.
+    NB: The copy_ownership() function does not exist, because on Windows, DACLs would to be
```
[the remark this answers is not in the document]
adferrand: It is all about `copy_ownership()` and why it is not suitable to have this kind of function, of the reasons you explained here. The fact that we forbid `os.chown` is for more general reason, and explained appropriately in the `certbot.compat.os` module.

Your opinion about `copy_ownership_and_apply_mode()` vs `copy_ownership()` is right. I am working on a more expanded comment here.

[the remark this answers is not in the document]
adferrand: Done.

adferrand:
About https://github.com/certbot/certbot/pull/7182#pullrequestreview-255512091
This is unfortunate.

Indeed, the mode returned by `stat` is largely useless, it is either `0o666` it the read-only attribute is not set, and `0o444` if it is.

To confirm what this portion of code is doing, here is a script example:
```python
from __future__ import print_function
import stat

base_mode = 0o600
modes_to_test = [0o740, 0o744, 0o447, 0o600, 0o777]

for mode_to_test in modes_to_test:
    old_mode = mode_to_test & (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | stat.S_IROTH)
    mode = base_mode | old_mode
    print('{0} => {1}'.format(oct(mode_to_test), oct(mode)))
```
```bash
0o740 => 0o640
0o744 => 0o644
0o447 => 0o644
0o600 => 0o600
0o777 => 0o674
```

Indeed, what you say is true, the code is keeping read permission for everyone, all permissions for group, and rw for the user. So on Windows, the read permission will always been set for everyone.

I am hesitating between your option, and just never give permissions to everyone (and to group since it is useless on Windows). If people are complaining, then we can think of a solution.

What do you prefer ?

certbot/plugins/webroot.py
```diff
@@ -168,19 +168,19 @@ def _create_challenge_dirs(self):
...
+                        # For Linux, this is coupled with the "umask" call above because
                         # os.mkdir's "mode" parameter may not always work:
                         # https://docs.python.org/3/library/os.html#os.mkdir
-                        os.mkdir(prefix, 0o0755)
-                        self._created_dirs.append(prefix)
-                        # Set owner as parent directory if possible
                         try:
-                            os.chown(prefix, stat_path.st_uid, stat_path.st_gid)
+                            filesystem.copy_ownership_and_apply_mode(
+                                path, prefix, 0o755, True, True)
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/storage.py
```diff
@@ -1105,12 +1105,11 @@ def save_successor(self, prior_version, new_cert,
...
-            old_mode = stat.S_IMODE(os.stat(old_privkey).st_mode) & \
-                (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | \
-                 stat.S_IROTH)
+            old_mode = (stat.S_IMODE(os.stat(old_privkey).st_mode) &
+                        (stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | stat.S_IROTH))
             mode = BASE_PRIVKEY_MODE | old_mode
-            os.chown(target["privkey"], -1, os.stat(old_privkey).st_gid)
-            filesystem.chmod(target["privkey"], mode)
+            filesystem.copy_ownership_and_apply_mode(
+                old_privkey, target["privkey"], mode, False, True)
```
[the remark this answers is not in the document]
adferrand: Done.

bmw:
> I am hesitating between your option, and just never give permissions to everyone (and to group since it is useless on Windows). If people are complaining, then we can think of a solution.
>
> What do you prefer ?

For me I think it depends on how you plan to accomplish this. What would the code look like to accomplish this? I personally don't think we should be changing the behavior on Linux in this PR.

adferrand:
My idea would be to apply this behavior only on Windows, not on Linux. To implement that, we would replace `stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | stat.S_IROTH` by a call to a function from `compat`, that would return `stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | stat.S_IROTH` on Linux, and `0` on Windows, explaining in a comment why we do not propagate the previous permissions on group and everyone in this situation.

This would avoid to make a complex reimplementation of `os.stat` or other facilities to transfer the DACL, while making a security model (everyone has no right on private keys) that I consider more satisfying.

bmw:
That works for me!

Outside the scope of this PR though, I think we should create an issue to look into forbidding `os.stat` which I did at https://github.com/certbot/certbot/issues/7212.

adferrand:
@bmw I added the workaround to transfer permissions on private key correctly on Linux and Windows.

certbot/compat/misc.py
```diff
@@ -17,6 +17,19 @@
...
 
+
+# UMASK_FOR_PRIVATE_KEY_PERMISSIONS defines what are the permissions flags to keep
+# when transferring the permissions from an old private key to a new one.
+if POSIX_MODE:
+    # On Linux, we keep read/write/execute permissions
+    # for group and read permissions for everybody.
+    UMASK_FOR_PRIVATE_KEY_PERMISSIONS = stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP | stat.S_IROTH
+else:
+    # On Windows, the mode returned by os.stat is not reliable (all bits are always set.
```
bmw: nit: It's not exactly true that all bits are always set. I'd remove the parenthetical here and just say:

> ...is not reliable so we do not...

certbot/compat/misc.py
```diff
@@ -17,6 +17,19 @@
 from certbot import errors
 from certbot.compat import os
 
+
+# UMASK_FOR_PRIVATE_KEY_PERMISSIONS defines what are the permissions flags to keep
```
bmw: nit: Is this really a umask for any definition of umask given at https://en.wikipedia.org/wiki/Umask? I think we should just change this and all other references to `MASK_FOR_PRIVATE_KEY_PERMISSIONS`.

certbot/compat/filesystem.py
```diff
@@ -44,6 +44,39 @@ def chmod(file_path, mode):
         _apply_win_mode(file_path, mode)
 
 
+# One could ask why there is no copy_ownership() function, or even a reimplementation
+# of os.chown() that would modify the ownership of file without touching the mode itself.
+# This is because on Windows, it would require recalculating the existing DACL against
+# the new owner, since DACL is composed of ACLs that targets a specific user, not dynamically
```
bmw:
```suggestion
# the new owner, since the DACL is composed of ACEs that targets a specific user, not dynamically
```

certbot/compat/misc.py

certbot/compat/misc.py

adferrand:
@bmw I noted your request about a unit test. I will do it tomorrow. However it will be an entirely mocked test, since Windows forbid to associate as a user any well know sid registered in the tests, except the admin group, which is already the default owner of created files.

adferrand:
I added the unit test.

bmw:
> However it will be an entirely mocked test, since Windows forbid to associate as a user any well know sid registered in the tests, except the admin group, which is already the default owner of created files.

There's not a single SID in https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems we can use? I'm in agreement with you that a fully mocked out test like this has little value (if the code already has coverage), but I think it'd be nice to have a test somewhere that the code functionally does what we expect. Given the somewhat complex Windows APIs which at least I have no experience with, I think it'd be good to verify that the code works now and continues to do so in the future.

adferrand:
I am sorry, but I think it will be impossible. 

I tried a lot of well known sid, on top of the three already defined in the tests, and always received the same error when trying to give ownership to the target sid.

This error is explained here: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc978898(v=technet.10)

Actually the sid that should receive ownership should be explicitly allowed to do so. The only usable sids described in the mentionned error are not suitable for the tests, as they are the owners of any file generated during the tests. 

The best I can do is what I did in the latest commit: mock the ownership transfer at the lowest level, so on the call to `SetFileSecurity` from pywin32. If a failure occurs beyond this level, it will be the problem of pywin32 implementation or the Windows API itself, but not certbot.

It is suboptimal, but coherent with the unit tests approach, that is not supposed to test third party dependencies, or the system, but instead is supposed to isolate the target code (here certbot codebase) and try to detect any defect in it.

At a higher level, so checking the correct execution of Certbot as a complete piece of software, it will be the integration tests purpose.

certbot/compat/filesystem.py
```diff
@@ -44,6 +44,39 @@ def chmod(file_path, mode):
...
+# the POSIX mode applied to the file and the current owner of this file.
+# Since copying and editing arbitrary DACL is very difficult, and since we actually know
+# the mode to apply at the time the owner of a file should change, it is easier to just
+# change the owner, then reapply the known mode, as copy_ownership_and_apply_mode() does.
+def copy_ownership_and_apply_mode(src, dst, mode, copy_user, copy_group):
+    # type: (str, str, int, bool, bool) -> None
+    """
+    Copy ownership (user and optionally group) from the source to the destination,
+    then apply given mode in compatible way for Linux and Windows. This replaces
+    the os.chown command.
```
bmw: I was just manually testing this function and noticed that `copy_group` has no effect on Windows. This is totally understandable, but I think we should document this here and/or in the documentation for the parameter down below.

certbot/compat/filesystem.py
```diff
@@ -44,6 +44,39 @@ def chmod(file_path, mode):
...
+    if POSIX_MODE:
+        stats = os.stat(src)
+        user_id = stats.st_uid if copy_user else -1
+        group_id = stats.st_gid if copy_group else -1
+        os.chown(dst, user_id, group_id)
+        os.chmod(dst, mode)
+    elif copy_user:
+        # There is no group handling in Windows
+        _copy_win_ownership(src, dst)
+        _apply_win_mode(dst, mode)
```
bmw: I just noticed that `_apply_win_mode` is conditional on the value of `copy_user` while the equivalent call to `os.chmod` is unconditional on Linux. What do you think about this?

With our current use of the function, it doesn't matter, but for simplicity/consistency, what do you think about removing the calls to `os.chmod` and `_apply_win_mode` directly and putting a single call to `chmod` (`filesystem.chmod`) at the bottom of this function unconditionally?

certbot/compat/filesystem.py
adferrand: Yes, I added it to the doc.

certbot/compat/filesystem.py
adferrand: Nice catch. Indeed, better to reuse factorization from `filesystem.chmod`. It is expected that the mode is always applied, so we execute it unconditionally.

PR #7189, opened 2019-06-25
Add Debian ARM AMI

bmw:
Inspired by the [number of ARM users we have](https://gist.github.com/c36e9bf6171f0b953e116dd4dc2c9558) (and because I want to rip out the only 32 bit test we have which without this PR would remove all tests we have on non-x86_64 architectures), this test adds an ARM image to the test farm tests. The image ID was taken from https://wiki.debian.org/Cloud/AmazonEC2Image/Stretch, you can see tests passing at https://travis-ci.com/certbot/certbot/builds/116857897, and I ran `test_tests.sh` locally and it passed.

### DIFF
--- a/tests/letstest/apache2_targets.yaml
+++ b/tests/letstest/apache2_targets.yaml
@@ -33,6 +33,12 @@ targets:
     type: ubuntu
     virt: hvm
     user: admin
+  - ami: ami-0ed54dd1b25657636
+    name: debian9_arm64
+    type: ubuntu
+    virt: hvm
+    user: admin
+    machine_type: a1.medium
   - ami: ami-077bf3962f29d3fa4
     name: debian8.1
     type: ubuntu
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -307,7 +307,9 @@ def grab_certbot_log():
 
 def create_client_instance(ec2_client, target, security_group_id, subnet_id):
     """Create a single client instance for running tests."""
-    if target['virt'] == 'hvm':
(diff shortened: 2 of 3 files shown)


tests/letstest/apache2_targets.yaml
```diff
@@ -33,6 +33,12 @@ targets:
     type: ubuntu
     virt: hvm
     user: admin
+  - ami: ami-0ed54dd1b25657636
+    name: debian9_arm64
+    type: ubuntu
+    virt: hvm
+    user: admin
+    machine_type: a1.medium
```
adferrand: What is the meaning of machine_type?

bmw: It's the instance type: https://aws.amazon.com/ec2/instance-types/

Maybe `instance_type` is a better name?

adferrand: Yes it is a good name.

Also, so you want to toggle two different types of A1 instances depending on the `--fast` flag?

bmw: I can change the name.

We could make a separate instance type for the fast flag, but it'd require farther reaching changes. My thought was to skip it until a need comes up since a1.medium is in between t2.micro and t2.medium and seems fast enough.

adferrand: Seems fare to me.

bmw:
Did you want me to change `machine_type` to `instance_type` or are you OK with the original name?

adferrand:
No, I think it is OK finally. The values are quite obvious.

PR #7198, opened 2019-06-26
Simplify and deprecate viewing config changes

bmw:
The `--apache` and `--nginx` flags have no effect on this command.

EDIT: In addition to removing these flags which have no effect, this PR implements the plan described in https://github.com/certbot/certbot/pull/7198#issuecomment-506519884.


ohemorange:
I suspect there was originally an idea to do something in the plugins about this, see https://github.com/certbot/certbot/blob/e9bcaaa576eb3bb29a71252f64c6cbde03d230d5/certbot/plugins/common.py#L193. Pretty sure we don't call that from anywhere either though?

bmw:
Nice catch.

I've never seen any documentation, design docs, or had anyone explain to me the plan here, but I suspect what's going on is that this was supposed to be [required functionality of installer plugins](https://github.com/certbot/certbot/blob/1c6210ee00b372484c0156f06e13605ed923d1a6/certbot/interfaces.py#L358) and our reverter and code in `plugins/common.py` was a possible way for installer plugins to implement that functionality but wasn't required they do it that way. If that's correct, this code should be creating an installer object and calling its `view_config_changes` method rather than assuming that the reverter should be used.

We could make this change, but I don't think anyone is using the `config_changes` subcommand. A [search of "certbot config_changes" on the community forums](https://community.letsencrypt.org/search?q=%22certbot%20config_changes%22) only results in posts from you and me. A [similar search on Google](https://google.com/search?hl=en&q=%22certbot%20config_changes%22) has 13 results as of writing this and only [one is real use of the subcommand](https://blog.tiger-workshop.com/lets-encrypt-for-apache-on-ubuntu-16-04/). All the others are copies of our changelog or `--help` output. This subcommand and all the `view_config_changes` methods just complicate our code and plugin interfaces.

For this reason, what do you think about taking the extreme route here of:

1. Having `certbot config_changes` start printing a deprecation warning.
2. Remove `view_config_changes` as a required method of `IInstaller`.
3. Remove the `view_config_changes` code from the Apache and Nginx plugins.
4. Have `certbot.plugins.common.Plugin.view_config_changes` and `certbot.reverter.Reverter.view_config_changes` start causing Python deprecation warnings.

After a couple releases in this state we delete the subcommand and shared code.

ohemorange:
I love this plan.

ohemorange:
There's a tiny chance that someone will be upset about the deprecation -- maybe no one talked about it because it was so clear and easy to use. But I doubt that.

bmw:
Awesome! I'll update this PR with those changes when I find some spare time. It should be pretty quick to do.

We also could be creating problems for ourselves in the future with official OS packaging which I just discussed at https://opensource.eff.org/eff-open-source/channels/certbot a bit, but I personally don't think we should let various distro's policies hamstring our ability to improve Certbot anymore. Let's make the changes we want and help deal with any OS packaging problems as we see fit later. I think we should be taking control of our own packaging soon so issues with distros that both want the latest Certbot and want us to never change anything become less of a problem for us.

bmw:
I added the changes i described at https://github.com/certbot/certbot/pull/7198#issuecomment-506519884.

PR #7209, opened 2019-07-01
Checks for delegated acme and no check root domain

bdowling:
These changes allow one to have a delegated DNS domain for the most-specific part to reduce the level of DNS modification access that needs to be provided to certbot container.

e.g. One could delegate `_acme-challenge.example.com` to a new zone, and create the access keys that only allow updates within that zone.

It also prevents returning the root DNS domain from `base_domain_name_guesses()` as looking for that zone is likely not useful.

Tests have been updated for both.


bdowling:
I attempt to fix one test, break another... will need some input on this one.

PR #7211, opened 2019-07-01
Pin/upgrade virtualenv in our tests

bmw:
This PR fixes our nightly test failures which can be seen at https://travis-ci.com/certbot/certbot/jobs/212188745.

The cause of these failures is that [older versions of virtualenv provide their own copy of distutils' `__init__.py`](https://github.com/pypa/virtualenv/blob/16.1.0/virtualenv_embedded/distutils-init.py#L1). This file imports the imp module which is deprecated in Python 3.7. When we import distutils in our Certbot tests, this causes a deprecation warning to be given and our tests to fail.

To fix this, I stopped using the virtualenv package from OS packages/CI and instead grab a pinned copy from PyPI in both AppVeyor and Travis. As part of doing this, we now also pin codecov, tox, and their dependencies (except when `CERTBOT_NO_PIN` is set which can be see [here](https://travis-ci.com/certbot/certbot/jobs/212441226#L283) using newer versions of those packages).

I also had to update the version of py in our tests to fix failures in AppVeyor seen at https://ci.appveyor.com/project/certbot/certbot/builds/25670201/job/0860r2utpryovrax#L117. I checked both [py's](https://py.readthedocs.io/en/latest/changelog.html#changelog) and [virtualenv's](https://virtualenv.pypa.io/en/stable/changes/) changelog to make sure they didn't drop support for any systems we care about.

I'm running our full test suite on these changes at https://travis-ci.com/certbot/certbot/builds/117578552.

Unfortunately, I wasn't able to figure out why the `CERTBOT_NO_PIN` tests were failing but our normal py37 unit tests were not ☹️ Regardless of that though, this PR fixes the problem and I think it's an improvement to pin/upgrade more of Certbot's dependencies while running tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -268,7 +268,10 @@ addons:
 
 # tools/pip_install.py is used to pin packages to a known working version
 # except in tests where the environment variable CERTBOT_NO_PIN is set.
-install: "tools/pip_install.py -U codecov tox"
+# virtualenv is listed here explicitly to make sure it is upgraded when
+# CERTBOT_NO_PIN is set to work around failures we've seen when using an older
+# version of virtualenv.
+install: "tools/pip_install.py -U codecov tox virtualenv"
 script: tox
 
 after_success: '[ "$TOXENV" == "py27-cover" ] && codecov -F linux'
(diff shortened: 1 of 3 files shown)


joohoi:
LGTM!

PR #7213, opened 2019-07-02
Update to the latest josepy.

bmw:
I released a new version of josepy this morning. Let's use it.


bmw:
Done.

Sorry for the ugly commit history. To flag a couple things about https://github.com/certbot/certbot/pull/7213/commits/ef4dd53e2a6cbe469d179bc1b28b426f02baebe0 .

1. This commit was made locally but not pushed to this branch before your review. I actually didn't realize that GitHub ordered their UI based upon the commit date in git rather than the order things appeared on GitHub.
2. We need to revert this commit for now because until we upgrade to [pytest 3.5.1+](https://docs.pytest.org/en/latest/changelog.html#pytest-3-5-1-2018-04-23) where they fixed this warning internally, the tests will spuriously pass and then fail in our unpinned tests because we still have problems here in `certbot`. See https://travis-ci.com/certbot/certbot/builds/117701111 (the boulder failures are unrelated).  I talk a bit about the surprising behavior that I don't fully understand caused by warnings that also occur inside pytest at https://opensource.eff.org/eff-open-source/pl/4b61xwajf38yzmay4xqstecoho.

bmw:
I think this PR is no longer needed because the initial change I wanted was made in https://github.com/certbot/certbot/pull/7337 and requiring `josepy>1.2` is no longer necessary because [CPython pushed back the requirement to use `collections.abc` to Python 3.9](https://github.com/python/cpython/pull/13409).

PR #7231, opened 2019-07-10
Run tests on apache-parser-v2

bmw:
We're planning on using the branch `apache-parser-v2` allowing us to incrementally work on the new Apache parser and feel comfortable landing temporary test code that we don't really want in `master`.

The `apache-parser-v2` branch is created and locked down, but neither Travis or AppVeyor are configured to run tests on it. See https://github.com/certbot/certbot/pull/7230. This PR fixes that problem.

This could probably just land in the `apache-parser-v2` branch, but why unnecessarily deviate the branch from `master`? It doesn't hurt anything there. Once it lands, I'll get this added to the `apache-parser-v2` branch too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,10 @@ branches:
 not-on-master: &not-on-master
   if: NOT (type = push AND branch = master)
 
-# Jobs for the extended test suite are executed for cron jobs and pushes on non-master branches.
+# Jobs for the extended test suite are executed for cron jobs and pushes to
+# non-development branches. See the explanation for apache-parser-v2 above.
 extended-test-suite: &extended-test-suite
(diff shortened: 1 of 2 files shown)


adferrand:
Also, here the extended test suite will be executed on Travis for any push on the `apache-parser-v2`.

Maybe you want to narrow the scope of the jobs to execute by modifying the `extended-tests-suite` yaml anchor?

bmw:
I did both. I also started to play around with the configs to skip more tests on pushes to the apache-parser-v2 branch, but since it will only happen on PRs which shouldn't be that frequent, I think we can probably skip this for now. This is a temporary setup and if we're ever feeling starved for CI resources because of this, we can easily revisit this.

PR #7232, opened 2019-07-10
Remove test symlink.

bmw:
I just noticed a symlink called `test` in the root of the Certbot repo that was added in https://github.com/certbot/certbot/pull/6967 and in particular in cea3ed940c62c4b97f3d72bc54ea054315c1534b.

I assume this was added by mistake so this PR deletes it. If it was added on purpose, it's probably fine, but I'd love to know why.


adferrand:
Thanks for catching this!

PR #7233, opened 2019-07-11
[Windows] Security model for files permissions - STEP 3f

adferrand:
This is (normally) the last PR needed to implement the complete windows files security model.

Now that all functions required to modify the DACL and ownership of files are settled, this PR adds the functions `check_mode` and `check_owner` in order to verify, in the scope of Windows, that the permissions are correctly set for sensible files.

The main impact is for the unit tests, on every place where controls are done to ensure that files are generated by Certbot with the correct permissions.

The second impact is more subtle: because of the large difference in how an account, or a user, is defined in Linux and in Windows, this PR drops any reference to a user representation in the Certbot codebase. I mean here that Certbot does not manipulate explicitly UID, GID or any sort of username/groupname, since everything is done inside `certbot.compat.filesystem`. As a consequence, the central method `certbot.utils.safe_open` does not accept anymore an UID as parameter (and so propagate this impact to every method that depends on that).

You will see that in reality, passing the UID to `safe_open` was useless, since it was always corresponding to the UID of the current user, as returned by `os.geteuid()` and affiliates. Now, passing ownership to the current user is done internally to `filesystem`, using the appropriate representation for each platform (UID for Linux, SID for Windows).

As a desirable side-effect, the partial and hacky functions `certbot.util.check_permissions` and `certbot.compat.misc.os_geteuid` are removed.


certbot/log.py
```diff
@@ -240,9 +238,12 @@ class TempHandler(logging.StreamHandler):
 
     """
     def __init__(self):
-        stream = tempfile.NamedTemporaryFile('w', delete=False)
+        fd, path = tempfile.mkstemp()  # To get a proper tempfile path
```
ohemorange: Could you explain what was wrong with the original code here?

certbot/plugins/webroot.py
```diff
@@ -207,7 +208,7 @@ def _perform_single(self, achall):
         old_umask = os.umask(0o022)
 
         try:
-            with open(validation_path, "wb") as validation_file:
+            with safe_open(validation_path, mode="wb", chmod=0o644) as validation_file:
```
ohemorange: I see a couple changes from `open` to `safe_open` in this PR; what was your decision-making process for which ones to change and which ones to leave?

certbot/tests/cert_manager_test.py
```diff
@@ -277,13 +277,12 @@ class SearchLineagesTest(BaseCertManagerTest):
     @mock.patch('certbot.storage.renewal_conf_files')
     @mock.patch('certbot.storage.RenewableCert')
     def test_cert_storage_error(self, mock_renewable_cert, mock_renewal_conf_files,
-        mock_make_or_verify_dir):
+                                mock_make_or_verify_dir):
```
ohemorange: It looks like this file only has lint-like changes; was something substantive changed here that I'm missing? I'm fine with the lint changes, but I do want to be sure I'm not missing something here.

certbot/tests/crypto_util_test.py
```diff
@@ -29,6 +30,9 @@ class InitSaveKeyTest(test_util.TempDirTestCase):
     def setUp(self):
         super(InitSaveKeyTest, self).setUp()
 
+        self.workdir = os.path.join(self.tempdir, 'workdir')
```
ohemorange: What was the reason behind this change?

certbot/tests/log_test.py
```diff
@@ -260,8 +260,7 @@ def tearDown(self):
         self.handler.close()
 
     def test_permissions(self):
-        self.assertTrue(
-            util.check_permissions(self.handler.path, 0o600, misc.os_geteuid()))
+        self.assertTrue(filesystem.check_mode(self.handler.path, 0o600))
```
ohemorange: We don't want to check the owner here?

certbot/compat/filesystem.py
```diff
@@ -77,6 +77,54 @@ def copy_ownership_and_apply_mode(src, dst, mode, copy_user, copy_group):
...
+    if POSIX_MODE:
+        return stat.S_IMODE(os.stat(file_path).st_mode) == mode
+
+    return _check_win_mode(file_path, mode)
+
+
+def check_owner(file_path):
+    # type: (str) -> bool
+    """
+    Check if given file is owner by current user.
```
ohemorange:
```suggestion
    Check if given file is owned by current user.
```

certbot/compat/filesystem.py
```diff
@@ -112,6 +160,7 @@ def open(file_path, flags, mode=0o777):  # pylint: disable=redefined-builtin
         # We set third parameter to 0 (`False`) to say that this security descriptor is
         # NOT constructed from a default mechanism, but is explicitly set by the user.
         # See https://docs.microsoft.com/en-us/windows/desktop/api/securitybaseapi/nf-securitybaseapi-setsecuritydescriptordacl  # pylint: disable=line-too-long
+        security.SetSecurityDescriptorOwner(user, False)
```
ohemorange: 1. nit: this line should go before the comment, because it explains what the next line does
2. do I understand correctly that this is something we probably wanted to add previously that was caught by fixing up the checks here, and not a new change?

certbot/compat/filesystem.py
```diff
@@ -213,6 +263,24 @@ def _apply_win_mode(file_path, mode):
...
+
+    # Apply the new DACL
+    security.SetSecurityDescriptorDacl(1, dacl, 0)
+    win32security.SetFileSecurity(file_path, win32security.DACL_SECURITY_INFORMATION, security)
+
+
+def _resolve_symlinks(file_path):
+    """
+    Find the real path for the given path. This method resolves symlinks, including
+    recursive symlinks, and is protected against symlinks that creates an infinite loop.
```
ohemorange:
```suggestion
    recursive symlinks, and is protected against symlinks that create an infinite loop.
```

certbot/tests/util_test.py
```diff
@@ -145,65 +143,29 @@ def setUp(self):
...
-        self.assertRaises(errors.Error, self._call, self.path, 0o400)
+        self.assertRaises(errors.Error, self._call, self.path, 0o644)
 
     def test_reraises_os_error(self):
         with mock.patch.object(filesystem, "makedirs") as makedirs:
             makedirs.side_effect = OSError()
             self.assertRaises(OSError, self._call, "bar", 12312312)
 
 
-class CheckPermissionsTest(test_util.TempDirTestCase):
```
ohemorange: We probably want a unit test for the new methods.

certbot/tests/util_test.py
```diff
@@ -145,65 +143,29 @@ def setUp(self):
...
 
     def test_existing_correct_mode_does_not_fail(self):
         self._call(self.path, 0o600)
-        self.assertTrue(misc.compare_file_modes(os.stat(self.path).st_mode, 0o600))
+        self.assertTrue(filesystem.check_mode(self.path, 0o600))
 
-    @test_util.skip_on_windows('Umask modes are mostly ignored on Windows.')
     def test_existing_wrong_mode_fails(self):
-        self.assertRaises(errors.Error, self._call, self.path, 0o400)
+        self.assertRaises(errors.Error, self._call, self.path, 0o644)
```
ohemorange: Right, meant to ask about this one as well, what's going on here?

certbot/compat/filesystem.py
adferrand: 1. Done
2. Yes, you are right, I forgot to add this before.

certbot/tests/crypto_util_test.py
adferrand: Without it, Windows tests can fail because the owner of the directory containing the key is not the correct one (it can be the actual user or Administrators group depending on the machine). Using a dedicated directory created with `filesystem.mkdir` ensures that the folder containing the keys is properly constructed accordingly to the Windows security model, like it would be done in a real execution of Certbot, and so make the tests behave correctly on this platform.

certbot/tests/cert_manager_test.py
adferrand: No, nothing more. My PyCharm was angry while working on `cert_manager` module.

certbot/plugins/webroot.py
adferrand: In fact, the builtin method `open` is not creating files that are correctly set up in terms of DACL and owner for Windows. I changed from builtin `open` to `certbot.util.safe_open` each time a test related to the concerned logic was asserting things about the security of its content (and basically was failing without `safe_open`).

certbot/tests/log_test.py
adferrand: Well, yes in fact. Corrected.

certbot/log.py
adferrand: Wrong permissions on Windows, since `NamedTemporaryFile` is calling internally `os.open`, that is creating wrong DACL and owner for a file. Since this class explicitly requires secured permissions through its unit tests, I changed its implementation to use the `filesystem` methods.

certbot/tests/util_test.py
adferrand: Sorry. It is a thing from the past, of a less capable version of the Windows model. Not needed anymore, and corrected.

certbot/tests/util_test.py
adferrand: Already done in #6964. This test suite is in fact an limited and old version used for `util_test.check_permissions` function that is removed by this PR.

certbot/log.py
ohemorange: Hold up, how was `NamedTemporaryFile` able to call `os.open` internally without throwing an error? Is our "replace `os.open` in `sys.modules`" plan not working properly?

certbot/plugins/webroot.py
ohemorange: Well, glad the tests were catching it! Makes me wonder, if the tests just caught those cases, maybe we're just lacking tests for other cases but should be doing the same thing. Do we need to disallow builtin `open`?

certbot/tests/util_test.py
ohemorange: I did actually mean unit tests for `check_owner`/`check_mode`/`check_permissions`, which it doesn't look like there are unit tests for. But if I'm correct, these are only used in testing code anyway. I see they reuse internal code from the `filesystem.py` module, which I suspect is why they're there in the first place. I don't love the dissonance of testing-only code in the same module as production code; some other options include adding unit tests for them, or placing them somewhere else. What do you think?

certbot/log.py
adferrand: But, it is not the plan. We discussed that and decided together not monkey patching os module.

certbot/tests/util_test.py
adferrand: No they are used in `safe_open`. So some unit tests then.

certbot/plugins/webroot.py
adferrand: As a result of the approach described in my previous comment, there is effectively not call to builtin `open` or `os.open` in certbot production code anymore. It may be interesting to lock this out using a pylint rule. I will check what is possible here.

adferrand: There is a lot of use of `open` in Certbot codebase. I agree with you that we should propose a complete replacement for `open` in Certbot in the future, but I think also it is not necessary now.

Indeed, these calls to `open` are not setting any permission on Linux, so one can expect that it is not a problem also to not control the permissions on Windows. It would be an overfit for Windows compared to Linux, and would impact significantly Certbot behavior on Linux.

Here `safe_open` and `os.open` are used in Certbot explicitly to open files in a safe way (in term of concurrency and in term of permission). For everywhere it is relevant, theses methods are used with a proper handle through `filesystem` module, and there are tested.

certbot/tests/util_test.py
adferrand: I added the tests.

certbot/plugins/webroot.py
ohemorange: > Indeed, these calls to open are not setting any permission on Linux, so one can expect that it is not a problem also to not control the permissions on Windows.

This can't be quite true -- if a file is created, it has to be created with *some* permissions.

The thing that surprises me is that there are calls to `open` that turn out to need `safe_open`. I would have thought that anything that needs `safe_open` would have already been using `os.open`. For the ones that were changed here, do you already know why it would have been fine on linux before? If not I'll check myself, I don't want to be duplicating work you may have already done.

adferrand: A file has obviously always a set of permissions defined. My sentence here had to be interpreted as "these calls to open are not setting any _specific_ permission".

So for the calls to `open` that was fine on Linux, but not on Windows:
* in `log.py`, it was because `NamedTemporaryFile`, despite the fact that it is claimed to create a temporary file in the most secured manner, is not on Windows. The bare call to `os.open` just let the DACL be inherited from upper level. On Linux, mode is set to 0o700 which is fine.
* in `webroot.py`, it was because `os.umask` is invoked before creating the file. So on Linux this umask is honored and file is created with the correct permissions. But Windows is not using this umask, so we need to use the `filesystem` functions to make this work correctly.

The last situation I anticipated, but it is not encountered in certbot codebase, would be a call to `os.open` with `mode` parameter explicitly set.

certbot/log.py
adferrand: Still not clear to me, I do not understand what you are asking me here.

I replaced here a call to a function from builtin module `tempfile` to another function from this same builtin module. Any call from functions of `tempfile` cannot it our `certbot.compat.os` function, since `tempfile` is not from our codebase.

I could use `filesystem.open` directly here, but I would loose the capability of `tempfile.mkstemp` to use the OS APIs to create consistently a new temporary file with random name, in particular on Linux. Instead, I first call `tempfile.mkstemp` to get the path, then delete it, then re-create it correctly using `util.safe_open` both for Windows and Linux.

If you are not satisfied with this approach, I propose to create first a temporary _directory_ using `tempfile.mkdtemp()`, and not care about the permissions on this directory, then create a safe `log` file in it using `util.safe_open`, and use this file as the stream facility for the `TempHandler`.

adferrand: I implemented the logic described in the last paragraph in by comment above.

ohemorange: To ask in a different way: isn't the point of replacing `os.open` in `sys.modules` with `certbot.compat.os.open` that calls to `os.open` error out? Is there a different `sys.modules` for different namespaces? Why isn't code in `tempfile` using our modified `sys.modules`?

adferrand: No it is not the point. We decided to not try to make a perfect replacement of `os.open` or any other function from `os` module, and not replace the `os` module to `certbot.compat.os` in the python runtime of Certbot.

Instead we build functions suitable and crafted explicitly for Certbot use, and replace only the call to `os` done in our own codebase only. We do not alterate in any way the behavior of our third party dependencies, or the Python distribution itself. If this leads to undesired results for Certbot, we replace these faulty calls by calls to functions from our own codebase. This is the case here with `tempfile` for this implementation of `TempHandler` that requires a strictly secure access.

ohemorange: In that case, could you explain what these lines do? https://github.com/certbot/certbot/blob/master/certbot/compat/os.py#L21-L27

adferrand: They are copying every member of the `os` module into the `certbot.compat.os` module (including private members and everything not in the special `__all__` dict of the module), in order to use the `certbot.compat.os` as a fully functional `os`-like module in our codebase.

adferrand:
Thanks a lot @ohemorange!

PR #7239, opened 2019-07-12
Unify license filename (LICENSE.txt)

sunpoet:
All license files are LICENSE.txt except the one in certbot-dns-route53. Change it to LICENSE.txt.


sunpoet:
Thanks!

PR #7240, opened 2019-07-12
Update certbot-auto script to work with RHEL 8

blackknight36:
/usr/bin/python no longer exists in RHEL 8.  This patch updates
the certbot-auto script to use python3 on nodes running RHEL 8.

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -755,13 +755,31 @@ elif [ -f /etc/redhat-release ]; then
   prev_le_python="$LE_PYTHON"
   unset LE_PYTHON
   DeterminePythonVersion "NOCRASH"
-  # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
-  RPM_DIST_VERSION=0
-  if [ "$RPM_DIST_NAME" = "fedora" ]; then
-    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) || echo "0"`
+
+  # Set RPM_DIST_VERSION to VERSION_ID from /etc/os-release after splitting on
+  # '.' characters (e.g. "8.0" becomes "8"). If the command exits with an
+  # error, RPM_DIST_VERSION is set to "unknown".
+  RPM_DIST_VERSION=$( (. /etc/os-release 2> /dev/null && echo "$VERSION_ID") | cut -d '.' -f1 || echo "unknown")
+
+  # If RPM_DIST_VERSION is an empty string or it contains any nonnumeric
+  # characters, the value is unexpected so we set RPM_DIST_VERSION to 0.
+  if [ -z "$RPM_DIST_VERSION" ] || [ -n "$(echo "$RPM_DIST_VERSION" | tr -d '[0-9]')" ]; then
+     RPM_DIST_VERSION=0
   fi
+
+  # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+  # RHEL 8 also uses python3 by default.
   if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$PYVER" -eq 26 ]; then
+    RPM_USE_PYTHON_3=1
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
+    RPM_USE_PYTHON_3=1
+  else
+    RPM_USE_PYTHON_3=0
+  fi
+
+  if [ "$RPM_USE_PYTHON_3" = 1 ]; then
     Bootstrap() {
       BootstrapMessage "RedHat-based OSes that will use Python3"
       BootstrapRpmPython3
@@ -775,6 +793,7 @@ elif [ -f /etc/redhat-release ]; then
     }
     BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
   fi
+
   LE_PYTHON="$prev_le_python"
 elif [ -f /etc/os-release ] && `grep -q openSUSE /etc/os-release` ; then
   Bootstrap() {
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -330,13 +330,31 @@ elif [ -f /etc/redhat-release ]; then
   prev_le_python="$LE_PYTHON"
   unset LE_PYTHON
   DeterminePythonVersion "NOCRASH"
-  # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
-  RPM_DIST_VERSION=0
-  if [ "$RPM_DIST_NAME" = "fedora" ]; then
-    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) || echo "0"`
+
+  # Set RPM_DIST_VERSION to VERSION_ID from /etc/os-release after splitting on
+  # '.' characters (e.g. "8.0" becomes "8"). If the command exits with an
+  # error, RPM_DIST_VERSION is set to "unknown".
+  RPM_DIST_VERSION=$( (. /etc/os-release 2> /dev/null && echo "$VERSION_ID") | cut -d '.' -f1 || echo "unknown")
+
+  # If RPM_DIST_VERSION is an empty string or it contains any nonnumeric
+  # characters, the value is unexpected so we set RPM_DIST_VERSION to 0.
+  if [ -z "$RPM_DIST_VERSION" ] || [ -n "$(echo "$RPM_DIST_VERSION" | tr -d '[0-9]')" ]; then
+     RPM_DIST_VERSION=0
   fi
+
+  # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+  # RHEL 8 also uses python3 by default.
   if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$PYVER" -eq 26 ]; then
+    RPM_USE_PYTHON_3=1
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
+    RPM_USE_PYTHON_3=1
+  else
+    RPM_USE_PYTHON_3=0
+  fi
+
+  if [ "$RPM_USE_PYTHON_3" = 1 ]; then
     Bootstrap() {
       BootstrapMessage "RedHat-based OSes that will use Python3"
       BootstrapRpmPython3
@@ -350,6 +368,7 @@ elif [ -f /etc/redhat-release ]; then
     }
     BOOTSTRAP_VERSION="BootstrapRpmCommon $BOOTSTRAP_RPM_COMMON_VERSION"
   fi
+
   LE_PYTHON="$prev_le_python"
 elif [ -f /etc/os-release ] && `grep -q openSUSE /etc/os-release` ; then
   Bootstrap() {
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -161,6 +161,7 @@ Authors
 * [Michael Schumacher](https://github.com/schumaml)
 * [Michael Strache](https://github.com/Jarodiv)
 * [Michael Sverdlin](https://github.com/sveder)
+* [Michael Watters](https://github.com/blackknight36)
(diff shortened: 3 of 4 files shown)


joohoi:
Hi, and thanks for the PR! 

While a bit confusing, the `certbot-auto` and `letsencrypt-auto` files in the repository _root_ are updated only during release process. In order to make changes to `certbot-auto`, please change `letsencrypt-auto-source/letsencrypt-auto.template` file instead, and run the `build.py` script in that directory afterwards - this builds a new `letsencrypt-auto-source/letsencrypt-auto` script, that will be copied to the root `certbot-auto` and `letsencrypt-auto` files during the release.

Overall the code in this PR itself looks fine.

blackknight36:
@joohoi Thanks for the feedback.  I've applied the changes to the template file which should result in a working certbot-auto script on RHEL 8.

adferrand:
Hello @blackknight36, you need also to commit the `letsencrypt-auto` script that `build.py` has generated.

blackknight36:
@adferrand File has been added.

bmw:
To give a belated update here, I've been building off this branch, updating [our automated tests](https://github.com/certbot/certbot/tree/master/tests/letstest) to also test against RHEL 8.

I'm not quite done with this, but so far none of the tests have found any problems in the code here which looks good to me.

I plan to finish up this work soon and open a new PR including your commits here as well as the additional tests which I hope for us to merge before our next release.

Thanks again for the well-written PR @blackknight36!

blackknight36:
@bmw No problem.  Glad I could contribute.

letsencrypt-auto-source/letsencrypt-auto
```diff
@@ -768,6 +768,14 @@ elif [ -f /etc/redhat-release ]; then
     }
     USE_PYTHON_3=1
     BOOTSTRAP_VERSION="BootstrapRpmPython3 $BOOTSTRAP_RPM_PYTHON3_VERSION"
+  # RHEL 8 uses python3.
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
```
adferrand: Following discussions in https://github.com/certbot/certbot/pull/7282#discussion_r308478388, I would like to avoid to repeat two times the same logic about python 3 bootstrapping for RPM distributions, but @bmw would like to avoid also a big and unreadable if condition when this bootstrap is triggered.

So, could you first set a truthy variable `RPM_USE_PYTHON_3` for the each following branches in a if/elif statement:
* fedora 29+
* rhel 8+
* rpm distrib with only python 2.6 (rhel6/centos 6)

Then, if `RPM_USE_PYTHON_3` is truthy, apply the python 3 rpm bootstrap, otherwise apply the python 2 rpm bootstrap ?

Once done, @bmw will update #7282 accordingly.

blackknight36: Would a case statement work?  There may be other distros such as SLES 15 that use python 3 as well.

blackknight36: I've updated the script with the changes suggested.  I also changed how RPM_DIST_VERSION gets set as I don't think it makes sense to set it to 0 and then immediately change it two lines later.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -330,13 +330,20 @@ elif [ -f /etc/redhat-release ]; then
...
+  RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) | cut -d '.' -f1 || echo "0"`
   if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$PYVER" -eq 26 ]; then
+    RPM_USE_PYTHON_3=1
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
+    RPM_USE_PYTHON_3=1
+  else
+    RPM_USE_PYTHON_3=0
+  fi
+
+  if [ $RPM_USE_PYTHON_3 == 1 ]; then
```
bmw: Thanks for the update @blackknight36.

Only new change I'd personally like to see is to use a single `=` here. I believe `==` isn't POSIX complaint and certbot-auto isn't always run on shells like bash.

blackknight36: I've changed that line to use a single = sign.

bmw: Thanks a lot!

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -330,13 +330,20 @@ elif [ -f /etc/redhat-release ]; then
...
   DeterminePythonVersion "NOCRASH"
+
   # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+  # RHEL 8 also uses python3 by default.
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
-  RPM_DIST_VERSION=0
-  if [ "$RPM_DIST_NAME" = "fedora" ]; then
-    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) || echo "0"`
-  fi
+  RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) | cut -d '.' -f1 || echo "0"`
```
bmw: Testing this script on CentOS 6, I get the output:
```
./letsencrypt-auto: line 763: [: : integer expression expected
./letsencrypt-auto: line 765: [: : integer expression expected
You have an ancient version of Python entombed in your operating system...
This isn't going to work; you'll need at least version 2.7.
```
I think the problem here is in the case of a failure parsing `/etc/os-release` like on CentOS 6 where it doesn't exist, `RPM_DIST_VERSION` is being set to a string.

adferrand: Hell, it means that my "fail-safe" `|| echo "0"` instruction was not fail-safe at all, and would set `RPM_DIST_VERSION` to a string. It was working because this line was only executed on fedora, and fedora has an existing and correct `/etc/os-release`.

I think the fail-safe instruction should be `|| echo 0`.

blackknight36: I've changed the failsafe to echo 0.  I can test on CentOS 6 when I return to work tomorrow.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -330,13 +330,20 @@ elif [ -f /etc/redhat-release ]; then
...
   if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$PYVER" -eq 26 ]; then
+    RPM_USE_PYTHON_3=1
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
+    RPM_USE_PYTHON_3=1
+  else
+    RPM_USE_PYTHON_3=0
+  fi
+
+  if [ $RPM_USE_PYTHON_3 = 1 ]; then
```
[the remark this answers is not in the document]
blackknight36: I have updated this line as suggested.

letsencrypt-auto-source/letsencrypt-auto.template
blackknight36: Just ran the script on a CentOS 6 VM, everything appears to be working.

blackknight36: I see the problem.  The failsafe is never being evaluated because of the pipe to cut which returns as true even when the file being read doesn't exist.

```
[root@c6 ~]# cat /tmp/foo.txt 2>/dev/null | cut -d "." -f1
[root@c6 ~]# echo $?
0
```

I added a condition to check if /etc/os-release exists which now allows the script to run on CentOS 6.

blackknight36: Another thing I just thought of, these conditions apply to more than just RHEL and CentOS.  There's also ScientificLinux and OracleLinux which are basically identical to RHEL.  Would it be acceptable to add a regex on line 347 to cover these other distros?  For example:

```
elif [ "$RPM_DIST_NAME" =~ "rhel|scientific|oracle|centos" -a "$RPM_DIST_VERSION" -ge 8 ]; then
```

adferrand: Technically it would. However, we saw that modifications done on `certbot-auto` were responsible of almost all point releases we had to do for the last 6 months.

Also the fact that `certbot-auto` should be removed in the next months in favor of a more robust and universal packaging approach is to be taken into accounts.

For these reasons, we want to avoid to do any modification to the script if it is not necessary to prevent a regression.

It is this case for that PR, since RHEL 7 is supported by `certbot-auto`, so having it failing when someone upgrade to RHEL 8 is definitely a regression.

However, adding new distributions, even really similar to the supported ones by `certbot-auto`, is really subject to discussions.

I think it should be not included in that PR for the least. But we can discuss about doing it or not for a future PR.

blackknight36: Is there a list of what distros are officially supported?  According to the  system requirements page certbot only requires "Python 2.7 or 3.4+ running on a UNIX-like operating system."

I'll hold off on that change but it's definitely something to be considered for future releases.

bmw: > Is there a list of what distros are officially supported?

Not really. There's a list of systems we offer explicit instructions in the dropdown menus at https://certbot.eff.org/instructions, however, in the same vein as your quoted text, there's an option there for "Other UNIX" which recommends that you use certbot-auto.

> I'll hold off on that change but it's definitely something to be considered for future releases.

I agree.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -330,13 +330,27 @@ elif [ -f /etc/redhat-release ]; then
...
   # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
+  # RHEL 8 also uses python3 by default.
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
-  RPM_DIST_VERSION=0
-  if [ "$RPM_DIST_NAME" = "fedora" ]; then
-    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) || echo "0"`
+
+  # CentOS 6 does not have an /etc/os-release file.  On these systems RPM_DIST_VERSION will be set to 0.
+  if [ -f /etc/os-release ]; then
+    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) | cut -d '.' -f1`
```
bmw: My understanding based on https://www.freedesktop.org/software/systemd/man/os-release.html is that there is no requirement that `VERSION_ID` exists in `/etc/os-release` in which case I think `RPM_DIST_VERSION` still is not set to an integer that can be compared below. With this in mind, I think it's still possible for us to have the [failures we had before on CentOS](https://github.com/certbot/certbot/pull/7240#discussion_r310284518).

Is there a more robust way to write this? What do you think @schoen?

bmw: Heck, it looks like we technically can't even count on `VERSION_ID` containing an integer according to the freedesktop link above which says:

> A lower-case string (mostly numeric, no spaces or other characters outside of 0–9, a–z, ".", "_" and "-") identifying the operating system version, excluding any OS name information or release code name, and suitable for processing by scripts or usage in generated filenames. This field is optional. Example: "VERSION_ID=17" or "VERSION_ID=11.04".

bmw: @schoen and I brainstormed about this out of band and here's what we came up with:
```
  # Set RPM_DIST_VERSION to VERSION_ID from /etc/os-release after splitting on
  # '.' characters (e.g. "8.0" becomes "8"). If the command exits with an
  # error, RPM_DIST_VERSION is set to unknown.
  RPM_DIST_VERSION=$( (. /etc/os-release 2> /dev/null && echo "$VERSION_ID") | cut -d '.' -f1 || echo "unknown")
  # If RPM_DIST_VERSION is an empty string or it contains any nonnumeric
  # characters, the value is unexpected so we set RPM_DIST_VERSION to 0.
  if [ -z "$RPM_DIST_VERSION" ] || [ -n "$(echo "$RPM_DIST_VERSION" | tr -d '[0-9]')" ]; then
     RPM_DIST_VERSION=0
  fi
```

blackknight36: That looks fine to me.  I'll update the PR tomorrow.  On a side note, this stuff would be a lot easier using facter.  :)

bmw: Thanks again @blackknight36.

What's facter?

blackknight36: @bmw facter is an application that provides node facts such as the operating system, OS release, and many other things.  It's used by Puppet but you can also run it standalone.

blackknight36: I've updated the script with the changes suggested.

bmw: Thank you.

I updated my PR which includes your latest changes at https://github.com/certbot/certbot/pull/7302.

Issue #5775, opened 2018-03-21
Clearly document what APIs we consider public


bmw:
For `acme` (and `josepy`), I'd say everything is public (besides objects prefixed with an underscore).

For the plugins, I'd say nothing is public besides the plugin API they implement.

For `certbot`, this is less clear. Everything in `achallenges.py`, `errors.py`, and `interfaces.py` should probably be considered public. Additionally, maybe some combination of `certbot/plugins/common.py`, `certbot/crypto_util.py`, and `certbot/util.py`?

We should clearly document this for people wanting to hack on/around the probject.

adferrand:
Following discussions in #6841, I would like to note the inconsistency of test modules from `certbot.plugins`. They are unit tests for certbot CI process, and so are for internal use and should not be used. But they are also part of the public API, and so likely to be used by third party plugins for what `certbot.plugins` as been designed for.

In theory, and following the design for the other tests, tests modules should be a part of a `*.tests` package that by convention is private and must not be imported in any way.

Solving the inconsistency could be to have a public API for plugin testing, exposing "abstract" modules, that would be in `certbot.plugins` or a submodule. Then for internal use certbot would subclass them in a `certbot.tests.plugins` to be effectively used by the certbot CI.

bmw:
I wrote up a proposal for how to resolve this at https://docs.google.com/document/d/1ozJvEDq2qTtKMeicCUyxH6BZ80_7XZSQDMHMo8uA1WA/edit?usp=sharing.

bmw:
I think the only pieces left here not tracked by other issues are:

- [ ] Update the submodule on the website repo and fix any paths to the docs. See the 2nd paragraph in https://github.com/certbot/certbot/pull/7544#issuecomment-557727307.
- [ ] Fix import orders. (Maybe check out [isort](https://isort.readthedocs.io/en/stable/)?)

bmw:
@adferrand is planning on fixing up import orders.

bmw:
@ohemorange, is fixing up the website repo still on your radar? I didn't expect this to happen, but website builds are going to fail until this resolved because links back to the source files are broken since the docs directory moved.

ohemorange:
> @ohemorange, is fixing up the website repo still on your radar? I didn't expect this to happen, but website builds are going to fail until this resolved because links back to the source files are broken since the docs directory moved.

Thanks for the reminder, I'll do this today

bmw:
This is done!

PR #7242, opened 2019-07-12
Implement a consistent realpath function in certbot.compat.filesystem

### LINKED ISSUE #7115: Implement a `realpath` method in `certbot.compat.filesystem` for Windows
The method `os.path.realpath` is broken on some Python versions. See https://bugs.python.org/issue9949.

Even if the situation is covered by `certbot.compat.filesytem` for its internal usage, it would be beneficial in the long term to have a proper `realpath` function exposed by this module to ensure a consistent behavior across platforms and python versions. 

See https://github.com/certbot/certbot/pull/6964#discussion_r287544214 for the relevant discussion.

adferrand:
Fixes #7115 

This PR creates a `realpath` method in `filesystem`, whose goal is to replace any call to `os.path.realpath` in Certbot. The reason is that `os.path.realpath` is broken on some versions of Python for Windows. See https://bugs.python.org/issue9949. The function created here works consistently across Linux and Windows.

As for the other forbidden functions in `os` module, our `certbot.compat.os` will raise an exception if its `path.realpath` function is invoked, and using the `os` module from Python is forbidden from the pylint check implemented in our CI.

Every call to `os.path.realpath` is corrected in `certbot` and `certbot-apache` modules.


certbot/compat/os.py
```diff
@@ -26,7 +26,8 @@
     if not hasattr(ourselves, attribute):
         setattr(ourselves, attribute, getattr(std_os, attribute))
 
-# Similar to os.path, allow certbot.compat.os.path to behave as a module
+# Import our path module, then allow certbot.compat.os.path to behave as a module (like os.path)
+from certbot.compat import path  # type: ignore  # pylint: disable=wrong-import-position
```
bmw: If we add this `type: ignore`, is all type checking disabled on all calls to all functions in `certbot.compat.os.path`? I suspect this is the case which doesn't seem great. Is there an easy way to avoid this `type: ignore`?

certbot/compat/path.py
```diff
@@ -0,0 +1,31 @@
+"""This compat module wraps os.path to forbid some functions."""
+# pylint: disable=function-redefined
+from __future__ import absolute_import
+
+# First round of wrapping: we import statically all public attributes exposed by the os.path
```
bmw: There's a cruel, nerdy, overachieving/overengineering part of me that wants to try and generalize this code of wrapping a module so it's not duplicated but it's not worth it.

certbot/compat/path.py
```diff
@@ -0,0 +1,31 @@
...
+# module. This allows in particular to have pylint, mypy, IDEs be aware that most of os.path
+# members are available in certbot.compat.path.
+from os.path import *  # type: ignore  # pylint: disable=wildcard-import,unused-wildcard-import,redefined-builtin,os-module-forbidden
```
bmw: I have the same concern here about `type: ignore` which I missed in my review of `certbot.compat.os`. I'm curious if there's anything we can do here. Disabling type checking on all functions in `os.path` doesn't seem ideal.

certbot/compat/path.py
```diff
@@ -0,0 +1,31 @@
+"""This compat module wraps os.path to forbid some functions."""
```
bmw: Do we want to let people import this module directly? I see no reason why we should.

What do you think about calling this file `_path.py`? We haven't done anything like this in Certbot yet, although I think we should as part of a resolution of https://github.com/certbot/certbot/issues/5775.

certbot/tests/compat/os_test.py
```diff
@@ -7,8 +7,12 @@
 class OsTest(unittest.TestCase):
     """Unit tests for os module."""
     def test_forbidden_methods(self):
+        # Checks for os module
         for method in ['chmod', 'chown', 'open', 'mkdir', 'makedirs', 'rename', 'replace']:
             self.assertRaises(RuntimeError, getattr(os, method))
+        # Checks for os.path module
+        for method in ['realpath']:
+            self.assertRaises(RuntimeError, getattr(os.path, method))
```
bmw: If we don't make `compat.path` private, for consistency I think we should make a `path_test.py` file for this test and then just add a test to this file that `compat.os.path` is `compat.path`.

certbot/compat/os.py
adferrand: https://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker, see in particular the note in this section.

I think mypy will continue to typecheck members of path even if ignore is set on this line.

adferrand: Let's give a demo of this behavior.

```python
# util.py
from os import *  # type: ignore
def chmod():
    pass
```

```python
# main.py
from util import chmod
from util import mkdir
from util import nonexistent
```

```bash
$ mypy main.py
main.py:3: error: Module 'util' has no attribute 'nonexistent'
```

Despite `# type: ignore` is set on the star import from `os` in `util`, mypy successfully sees that `mkdir` exists in `util` thanks to the star import, and that `nonexistent` does not exist in `util`.

certbot/compat/path.py
adferrand: See my response in https://github.com/certbot/certbot/pull/7242#discussion_r304701443

certbot/compat/path.py
adferrand: Well I would have really liked to do so, but on top of the complexity of abstracting this, you would lose a great feature: Python IDE would not be able to autocomplete anything, since the wrapper modules would become completely dynamic. Pylint and MyPy would become really angry also.

certbot/compat/path.py
adferrand: Using underscored name modules for internal uses was already in my suggestion list for the team. I will gladly rename the module.

adferrand: Done.

certbot/tests/compat/os_test.py
adferrand: Since I made the `path` module internal, this comment is not relevant anymore.

certbot/compat/os.py
bmw: Thanks for the explanation! Typechecking indeed still seems to work fine so I think we can leave things as is.

certbot/compat/path.py
bmw: I'd love to see that suggestion list sometime!

Issue #7243, opened 2019-07-15
dns-rfc2136 plugin: once again fix CNAME/DNAME behavior


hpax:
The fix for bug #7142 once again introduces bug #6787. Note that the patch proposed by myself for #6787 is automatic, rather than expecting manual configuration as in the botched #7029 

I have refreshed the patch on top of the current master and will submit a pull request shortly.

schoen:
@bmw, could you take a look at this because you created the fix @hpax is referring to?

bmw:
I'm going to close this in favor of the PR implementing this at https://github.com/certbot/certbot/pull/7244 and the general issue tracking this for all of our DNS plugins at https://github.com/certbot/certbot/issues/6566.

PR #7244, opened 2019-07-15
dns-rfc2136: find the correct zone/name when CNAME/DNAMEs are used

hpax:
Dynamic zones have significant problems with DNSSEC and with redundant
servers (which, of course is highly desirable for DNS.) The obvious
solution to that is to use a CNAME or DNAME record to point the
_acme-challenge to a different zone which can have different NS and
TTL properties. In particular, breaking DNSSEC support breaks exactly
the chain of trust on which ACME depends, and is thus extremely
undesirable.

In order to find the correct base zone and name-in-zone when
CNAME/DNAMEs might be present, search from the top down instead of the
bottom up, and allow non-authoritative answers for anything other than
the final SOA. There is no guarantee that the authentication server is
authoritative for anything but the zone into which the TXT record is
to be placed.

If the authentication server disallows recursion, this code will this
do the right thing as long as the server is authoritative for the
dynamic zone and any zone which contains a CNAME or DNAME record. If
that is not the case, then the server must support recursion for its
dynamic clients; it obviously does not need to offer that service to
the general public. If even this turns out to be unacceptable, then
the solution would be to query the normal nameservers (using the
system resolver), at least if an !AA !RA response is returned. The
dns.resolver module has a zone_for_name() function, but unfortunately
it does not return the name-in-zone, and to me its algorithm appears
to be incorrect (at least for our purposes) in a way that is similar
to the previous dns-rfc2136 code.

This patch changes several levels of the interface to use
dns.name.Name objects instead of strings, and passes dns.rdata.Rdata
objects between _query_soa() and _find_domain(). This turns out to
significantly simplify a fair number of things, but requires a fair
number of changes to the test suite. Clean up the test suite by
implementing a mock resolver with a mapping instead of a simple
sequence of return values, and by precomputing dns.name.Name objects
for (sub)domains and prefixes used.

Signed-off-by: H. Peter Anvin <hpa@zytor.com>

## Pull Request Checklist

  the change being made.
- [ ] Add [mypy type

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -15,6 +15,7 @@ import zope.interface
 from certbot import errors
 from certbot import interfaces
 from certbot.plugins import dns_common
+from collections import defaultdict
 
 logger = logging.getLogger(__name__)
 
@@ -109,11 +110,9 @@ class _RFC2136Client(object):
         :raises certbot.errors.PluginError: if an error occurs communicating with the DNS server
         """
 
-        domain = self._find_domain(record_name)
+        logger.debug('Adding TXT record: %s %d "%s"', record_name, record_ttl, record_content)
 
-        n = dns.name.from_text(record_name)
-        o = dns.name.from_text(domain)
-        rel = n.relativize(o)
+        (rel, domain) = self._find_domain(record_name)
 
         update = dns.update.Update(
             domain,
@@ -144,11 +143,7 @@ class _RFC2136Client(object):
         :raises certbot.errors.PluginError: if an error occurs communicating with the DNS server
         """
 
-        domain = self._find_domain(record_name)
-
-        n = dns.name.from_text(record_name)
-        o = dns.name.from_text(domain)
-        rel = n.relativize(o)
+        (rel, domain) = self._find_domain(record_name)
 
         update = dns.update.Update(
             domain,
@@ -174,49 +169,153 @@ class _RFC2136Client(object):
         Find the closest domain with an SOA record for a given domain name.
 
         :param str record_name: The record name for which to find the closest SOA record.
-        :returns: The domain, if found.
-        :rtype: str
-        :raises certbot.errors.PluginError: if no SOA record can be found.
+        :returns: tuple of (`entry`, `zone`) where
+                `entry` - canonical relative entry into the target zone;
+                `zone` - canonical absolute name of the zone to be modified.
+        :rtype: (`dns.name.Name`, `dns.name.Name`)
+        :raises certbot.errors.PluginError: if the search failed for any reason.
         """
 
-        domain_name_guesses = dns_common.base_domain_name_guesses(record_name)
+        # Note: an absolute dns.name.Name ends in dns.name.root, which
+        # is non-empty. Therefore the first prefix.split(1) splits off
+        # dns.name.root, i.e. example.com. -> (example.com, .), not
+        # example.com. -> (example, com.).  dns.name.empty, however,
+        # is an actual empty name, has a truth value of False, and is
+        # an identity element for the append operation; thus
+        # dns.name.root + dns.name.empty == dns.name.root.
+        #
+        # This code relies on these properties.
 
-        # Loop through until we find an authoritative SOA record
-        for guess in domain_name_guesses:
-            if self._query_soa(guess):
-                return guess
+        domain = dns.name.from_text(record_name)
+        prefix = domain
+        suffix = dns.name.empty
+        found  = None
+        domstr = str(domain)    # For messages, may have a DNAME/CNAME added
 
-        raise errors.PluginError('Unable to determine base domain for {0} using names: {1}.'
-                                 .format(record_name, domain_name_guesses))
+        # The domains already queried and the corresponding results
+        domain_names_searched = dict()
 
-    def _query_soa(self, domain_name):
+        while prefix:
+            (prefix, next_label) = prefix.split(1)
+            suffix = next_label + suffix
+
+            # Don't re-query if we have already been here (normal
+            # during DNAME/CNAME re-walk)
+            if suffix in domain_names_searched:
+                result = domain_names_searched[suffix]
+            else:
+                result = self._query_soa(suffix)
+                domain_names_searched[suffix] = result
+
+            (auth, rr) = result
+            if rr is None:
+                # Nothing to do, just descend the DNS hierarchy
+                pass
+            elif rr.rdtype == dns.rdatatype.SOA:
+                # We found an SOA, authoritative or not
+                found = (auth, prefix, suffix)
+            else:
+                # We found a DNAME or CNAME. We need to start the walk over
+                # from the common point of departure.
+                target = rr.target
+                if target in domain_names_searched:
+                    # DNAME/CNAME loop!
+                    errors.PluginError('%s %s loops seeking SOA for %s',
+                                       suffix, repr(rr), domstr)
+
+                # Restart from the root, replacing the current suffix
+                prefix = prefix + target
+                suffix = dns.name.empty
+                found  = None
+                domstr = str(domain)+' ('+str(prefix)+')' # For messages
+
+        if not found:
+            raise errors.PluginError('No SOA of any kind found for %s',
+                                     domstr)
+
+        (auth, prefix, suffix) = found
+        if not auth:
+            raise errors.PluginError('SOA %s for %s not authoritative',
+                                     suffix, domstr)
+        return (prefix, suffix)
+
+    def _query_soa(self, domain):
         """
         Query a domain name for an authoritative SOA record.
 
-        :param str domain_name: The domain name to query for an SOA record.
-        :returns: True if found, False otherwise.
-        :rtype: bool
+        :param dns.name.Name domain: The domain name to query for an SOA record.
+        :returns: (`authoritative`, `rdata`) if found
+                autoritative bool if response was authoritative
+                rdata dns.rdata.Rdata or None the returned record
+        :rtype: (`bool`, `dns.rdata.Rdata` or `None`)
         :raises certbot.errors.PluginError: if no response is received.
         """
 
-        domain = dns.name.from_text(domain_name)
+        # In order to capture any possible CNAMEs, we have to do the
+        # search upward from the root. On the way, any time we find a
+        # SOA record, save it; the final SOA record captured is the
+        # target. If that SOA record is not authoritative, then
+        # we have a fatal error.
+        #
+        # As we want to know about either type, we request recursion
+        # from the target name server. If the target nameserver does
+        # not provide recursion services, it will still work for
+        # finding an authoritative SOA, DNAME or CNAME record
+        # in a zone for which the nameserver is authoritarive; this is
+        # expected to be the common case, although it is not 100%
+        # guaranteed. The only ways to avoid that, ultimately, is to use
+        # a trusted recursive nameserver instead if we get a !RA response
+        # (e.g. using dns.resolver?) or actually query the authoritative name
+        # servers all the way from the top.
+        #
+        # We intentionally only look in the answer section, not in
+        # the authority or additional sections, and only for records
+        # which match the requested domain name exactly.
+        #
+        # If we get more than one SOA, DNAME, or CNAME record of the
+        # same type and exactly matching the requested domain in the
+        # *answer* section we are really in an error situation (these
+        # are all singleton RRs), but try to make the best of the
+        # situation.
 
         request = dns.message.make_query(domain, dns.rdatatype.SOA, dns.rdataclass.IN)
-        # Turn off Recursion Desired bit in query
-        request.flags ^= dns.flags.RD
 
         try:
+            logmsg = 'Query '+str(domain)
             response = dns.query.udp(request, self.server, port=self.port)
             rcode = response.rcode()
+            logmsg += ': '+dns.rcode.to_text(rcode)
 
-            # Authoritative Answer bit should be set
-            if (rcode == dns.rcode.NOERROR and response.get_rrset(response.answer,
-                domain, dns.rdataclass.IN, dns.rdatatype.SOA) and response.flags & dns.flags.AA):
-                logger.debug('Received authoritative SOA response for %s', domain_name)
-                return True
+            auth = (response.flags & dns.flags.AA) != 0
+            if auth:
+                logmsg += ', authoritative'
+            else:
+                logmsg += ', non-authoritative'
 
-            logger.debug('No authoritative SOA record found for %s', domain_name)
-            return False
+            found = dict()
+            for rrset in response.answer:
+                if rrset.name != domain: continue
+                if rrset.rdclass != dns.rdataclass.IN: continue
+                for rr in rrset:
+                    if not rr.rdtype in found:
+                        found[rr.rdtype] = [rr]
+                    elif not rr in found[rr.rdtype]:
+                        # Explicitly ignore exact duplicate RRs
+                        found[rr.rdtype].append(rr)
+
+            for rdtype in found:
+                logmsg += ' %s %d' % (dns.rdatatype.to_text(rdtype), len(found[rdtype]))
+
+            retrr = None
+            for rdtype in dns.rdatatype.SOA, dns.rdatatype.DNAME, dns.rdatatype.CNAME:
+                if rdtype in found:
+                    retrr = found[rdtype][0]    # Use the first one returned
+                    break
+
+            logmsg += ', returning '+repr(retrr)
+            logger.debug(logmsg)
+            return (auth, retrr)
+
         except Exception as e:
-            raise errors.PluginError('Encountered error when making query: {0}'
+           raise errors.PluginError('Encountered error when making query: {0}'
                                      .format(e))
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136_test.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136_test.py
@@ -3,7 +3,17 @@
 import unittest
 
 import dns.flags
+import dns.message
+import dns.name
+import dns.namedict
 import dns.rcode
+import dns.rdata
+import dns.rdataclass
+import dns.rdatatype
+import dns.rdtypes
+import dns.rdtypes.ANY.CNAME
+import dns.rdtypes.ANY.DNAME
+import dns.rdtypes.ANY.SOA
 import dns.tsig
 import mock
 
@@ -13,12 +23,13 @@ from certbot.plugins import dns_test_common
 from certbot.plugins.dns_test_common import DOMAIN
 from certbot.tests import util as test_util
 
+from collections import defaultdict
+
 SERVER = '192.0.2.1'
 PORT = 53
 NAME = 'a-tsig-key.'
 SECRET = 'SSB3b25kZXIgd2hvIHdpbGwgYm90aGVyIHRvIGRlY29kZSB0aGlzIHRleHQK'
 VALID_CONFIG = {"rfc2136_server": SERVER, "rfc2136_name": NAME, "rfc2136_secret": SECRET}
-
 
 class AuthenticatorTest(test_util.TempDirTestCase, dns_test_common.BaseAuthenticatorTest):
 
@@ -72,27 +83,59 @@ class AuthenticatorTest(test_util.TempDirTestCase, dns_test_common.BaseAuthentic
 
 class RFC2136ClientTest(unittest.TestCase):
 
+    def _stub_query_soa(self, domain):
+        (rrname, retval) = self._query_soa_return.get_deepest_match(domain)
+        if domain != rrname:
+            return (retval[0], None) # Just the authoritative flag
+        else:
+            return retval
+
     def setUp(self):
         from certbot_dns_rfc2136.dns_rfc2136 import _RFC2136Client
 
         self.rfc2136_client = _RFC2136Client(SERVER, PORT, NAME, SECRET, dns.tsig.HMAC_MD5)
+        self.domain = dns.name.from_text(DOMAIN)
+        self.prefix = dict()
+        self.subdom = dict()
+        for pfx in 'foo', 'bar', 'foo.bar', 'baz', 'quux', 'cname', 'dname', 'subzone', 'bad':
+                self.prefix[pfx] = dns.name.from_text(pfx, dns.name.empty)
+                self.subdom[pfx] = self.prefix[pfx] + self.domain
+
+        # _find_domain stub -> (bar, DOMAIN.)
+        self._mock_find_domain = mock.MagicMock(return_value=(self.prefix['bar'], self.domain))
+
+        # Mock DNS records
+        soa_rr = dns.rdtypes.ANY.SOA.SOA(dns.rdataclass.IN, dns.rdatatype.SOA,
+                                         SERVER, SERVER, 1, 2, 3, 4, 5)
+        cname_rr = dns.rdtypes.ANY.CNAME.CNAME(dns.rdataclass.IN, dns.rdatatype.CNAME,
+                                               self.subdom['subzone'])
+        dname_rr = dns.rdtypes.ANY.DNAME.DNAME(dns.rdataclass.IN, dns.rdatatype.DNAME,
+                                               self.subdom['subzone'])
+        self._query_soa_return = \
+            dns.namedict.NameDict({dns.name.root : (False, soa_rr),
+                                   self.domain : (True, soa_rr),
+                                   self.subdom['subzone'] : (True, soa_rr),
+                                   self.subdom['cname'] : (True, cname_rr),
+                                   self.subdom['dname'] : (False, dname_rr),
+                                   self.subdom['bad'] : (False, soa_rr)})
+        self._mock_query_soa = mock.MagicMock(side_effect=self._stub_query_soa)
 
     @mock.patch("dns.query.tcp")
     def test_add_txt_record(self, query_mock):
         query_mock.return_value.rcode.return_value = dns.rcode.NOERROR
         # _find_domain | pylint: disable=protected-access
-        self.rfc2136_client._find_domain = mock.MagicMock(return_value="example.com")
+        self.rfc2136_client._find_domain = self._mock_find_domain
 
-        self.rfc2136_client.add_txt_record("bar", "baz", 42)
+        self.rfc2136_client.add_txt_record("bar"+DOMAIN, "baz", 42)
 
         query_mock.assert_called_with(mock.ANY, SERVER, port=PORT)
-        self.assertTrue("bar. 42 IN TXT \"baz\"" in str(query_mock.call_args[0][0]))
+        self.assertTrue("bar 42 IN TXT \"baz\"" in str(query_mock.call_args[0][0]))
 
     @mock.patch("dns.query.tcp")
     def test_add_txt_record_wraps_errors(self, query_mock):
         query_mock.side_effect = Exception
         # _find_domain | pylint: disable=protected-access
-        self.rfc2136_client._find_domain = mock.MagicMock(return_value="example.com")
+        self.rfc2136_client._find_domain = self._mock_find_domain
 
         self.assertRaises(
             errors.PluginError,
@@ -103,7 +146,7 @@ class RFC2136ClientTest(unittest.TestCase):
     def test_add_txt_record_server_error(self, query_mock):
         query_mock.return_value.rcode.return_value = dns.rcode.NXDOMAIN
         # _find_domain | pylint: disable=protected-access
-        self.rfc2136_client._find_domain = mock.MagicMock(return_value="example.com")
+        self.rfc2136_client._find_domain = self._mock_find_domain
 
         self.assertRaises(
             errors.PluginError,
@@ -114,18 +157,18 @@ class RFC2136ClientTest(unittest.TestCase):
     def test_del_txt_record(self, query_mock):
         query_mock.return_value.rcode.return_value = dns.rcode.NOERROR
         # _find_domain | pylint: disable=protected-access
-        self.rfc2136_client._find_domain = mock.MagicMock(return_value="example.com")
+        self.rfc2136_client._find_domain = self._mock_find_domain
 
         self.rfc2136_client.del_txt_record("bar", "baz")
 
         query_mock.assert_called_with(mock.ANY, SERVER, port=PORT)
-        self.assertTrue("bar. 0 NONE TXT \"baz\"" in str(query_mock.call_args[0][0]))
+        self.assertTrue("bar 0 NONE TXT \"baz\"" in str(query_mock.call_args[0][0]))
 
     @mock.patch("dns.query.tcp")
     def test_del_txt_record_wraps_errors(self, query_mock):
         query_mock.side_effect = Exception
         # _find_domain | pylint: disable=protected-access
-        self.rfc2136_client._find_domain = mock.MagicMock(return_value="example.com")
+        self.rfc2136_client._find_domain = self._mock_find_domain
 
         self.assertRaises(
             errors.PluginError,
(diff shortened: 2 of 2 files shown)


hpax:
Fix for issue #7243.

bmw:
@adferrand, are you interested in reviewing this when you magically find spare cycles? @joohoi, can probably help you with testing if you want.

adferrand:
Hello @hpax! Your PR is extremely promising.

Before we start the review itself, I would like to talk about few points.

First could you clean the errors detected by pylint and mypy in the CI build ? It seems to be several parts of unused code, and it may be even fix by itself the coverage drop we see with your PR (and which makes the coverage job fail).

Second, the mechanism you propose is very powerful. I am asking myself if it even covers the specific situation described here https://github.com/certbot/certbot/issues/6566, with a PR to solve it here https://github.com/certbot/certbot/pull/5350, over which we took ownership with https://github.com/certbot/certbot/pull/6644. Because if your PR is also covering the case of automatically resolving a CNAME entry of `_acme-challenge.domain.net` pointing to an external zone entry, we could just drop the other PR and continue with yours.

Third, again, your mechanism is very powerful ^^ It allows basically to always resolve the correct zone to use to insert the TXT entry. It seems to not being tied to any specific functionality of `dns-rfc2136` plugin. How it would be complex/doable to generalize the approach for any Certbot DNS plugin ? If for instance `domain.net` zone is hosted by Cloudflare, but `subdomain.domain.net` is delegated to DigitalOcean, this would allow `certbot certonly --dns-digitalocean -d test.subdomain.domain.net` work correctly, by calling DigitalOcean API for `subdomain.domain.net`, not `domain.net`.

bmw:
The Certbot team just talked about this PR a bit in our weekly standup. I'll echo what Adrien said and say that we are very excited about the potential of this PR!

Two questions that came out of that conversation that we don't immediately know the answer to are:

1. Are there any situations where landing this PR would break a currently working setup?
2. Should this code definitively find the correct zone/name in all situations or is it using heuristics to make the best approximation? In other words, in what scenarios would this automatic resolution be insufficient and you may still need a PR like https://github.com/certbot/certbot/pull/6644 to manually tell Certbot which domain needs to be modified?

One scenario we thought about for question (2) that may be a problem was split-horizon DNS.

hpax:
[Answering two threads]

On 7/17/19 1:10 AM, Adrien Ferrand wrote:
> Hello @hpax <https://github.com/hpax>! Your PR is extremely promising.
> 
> Before we start the review itself, I would like to talk about few points.
> 
> First could you clean the errors detected by pylint and mypy in the CI
> build ? It seems to be several parts of unused code, and it may be even
> fix by itself the coverage drop we see with your PR (and which makes the
> coverage job fail).

My only problem is that I'm currently buried very very deep. I posted
this because I had to do the most trivial possible forward port after my
own system broke from an update.

It would have been far better if I had gotten a positive response the
first time I submitted it.

> Second, the mechanism you propose is very powerful. I am asking myself
> if it even covers the specific situation described here #6566
> <https://github.com/certbot/certbot/issues/6566>, with a PR to solve it
> here #5350 <https://github.com/certbot/certbot/pull/5350>, over which we
> took ownership with #6644
> <https://github.com/certbot/certbot/pull/6644>.

It ought to. I really isn't any difference than individual host CNAME
records.

> Because if your PR is
> also covering the case of automatically resolving a CNAME entry of
> |_acme-challenge.domain.net| pointing to an external zone entry, we
> could just drop the other PR and continue with yours.

This is in fact exactly my own use case.

> Third, again, your mechanism is very powerful ^^ It allows basically to
> always resolve the correct zone to use to insert the TXT entry. It seems
> to not being tied to any specific functionality of |dns-rfc2136| plugin.
> How it would be complex/doable to generalize the approach for any
> Certbot DNS plugin ? If for instance |domain.net| zone is hosted by
> Cloudflare, but |subdomain.domain.net| is delegated to DigitalOcean,
> this would allow |certbot certonly --dns-digitalocean -d
> test.subdomain.domain.net| work correctly, by calling DigitalOcean API
> for |subdomain.domain.net|, not |domain.net|.

It is sort-of-true. Currently it relies on the RFC 2136 DNS resolver to
either allow recursion (for clients allowed to push) or it falls back on
(fairly strong) heuristics.

In order to make it generic, it would need to either walk authoritative
servers all the way from the root (which would break behind some
firewalls) or it would need to do these queries via the system-provided
caching resolvers. The best way to do that would probably be to use
dns.resolver.query() for all the searching instead of piggy backing on
the RFC 2136 code.

One major reason I didn't mess with the generic code is that I don't
know how the various DNS providers do their thing; for example, if the
zone boundaries exposed to the Internet actually match what they
internally use.  I would not have either the facilities nor bandwidth to
address that probem.

However, at least for the subcase of a CNAME/DNAME pointing into one of
those zones one would expect that it would Just Work[TM].

> The Certbot team just talked about this PR a bit in our weekly standup. I'll echo what Adrien said and say that we are very excited about the potential of this PR!
> 
> Two questions that came out of that conversation that we don't immediately know the answer to are:
> 
>     Are there any situations where landing this PR would break a currently working setup?

I do not believe so, even in the case of an internal nameserver, as it
would still encounter the top-level SOA.

>     Should this code definitively find the correct zone/name in all situations or is it using heuristics to make the best approximation? In other words, in what scenarios would this automatic resolution be insufficient and you may still need a PR like #6644 to manually tell Certbot which domain needs to be modified?
>
> One scenario we thought about for question (2) that may be a problem was split-horizon DNS.

Basically, if the zone name that has to be modified has a different name
than the terminal SOA from a domain walk, there is simply no way to
discovering it automatically. It isn't even certain that the certbot
client can even see the zone in which the _acme-challenge record needs
to be added in that case. I do believe this falls under the category of
"pathological setups."

	-hpa

systemcrash:
Oh wow. Finally.  Working code for what EFF recommended [here](https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation).

Anyway, tested the version in this PR. Works. It has my blessing. 

```
# certbot certonly --dry-run --dns-rfc2136 -d rt.example.org --dns-rfc2136-credentials /root/.secrets/certbot/rfc2136.ini
Plugins selected: Authenticator dns-rfc2136, Installer None
Starting new HTTPS connection (1): acme-staging-v02.api.letsencrypt.org
Obtaining a new certificate
dns-01 challenge for rt.example.org
Waiting 60 seconds for DNS changes to propagate
Waiting for verification...
Cleaning up challenges

IMPORTANT NOTES:
 - The dry run was successful.
```

Dig logs:
```
[root@dnssrv named]# dig TXT _acme-challenge.rt.example.org

; <<>> DiG 9.9.4-RedHat-9.9.4-74.el7_6.2 <<>> TXT _acme-challenge.rt.example.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 26945
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;_acme-challenge.rt.example.org.	IN	TXT

;; ANSWER SECTION:
_acme-challenge.rt.example.org. 3555 IN	CNAME	_acme-challenge.example.org.
_acme-challenge.example.org. 3555	IN	CNAME	_acme-challenge.challenge.example.org.

;; AUTHORITY SECTION:
challenge.example.org.	15	IN	SOA	ns0.example.org. dnsdomains.example.org. 31 300 120 600 60

;; Query time: 11 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Wed Sep 04 03:21:46 CEST 2019
;; MSG SIZE  rcvd: 178

[root@dnssrv named]# dig _acme-challenge.challenge.example.org TXT

; <<>> DiG 9.9.4-RedHat-9.9.4-74.el7_6.2 <<>> _acme-challenge.challenge.example.org TXT
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 28624
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;_acme-challenge.challenge.example.org. IN	TXT

;; ANSWER SECTION:
_acme-challenge.challenge.example.org. 119 IN TXT	"2mwuCvyBLgoxZyp643rU-QwPmsCliP1H2lFhkpk7mp4"

;; Query time: 22 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Wed Sep 04 03:21:51 CEST 2019
```

Bind logs
```
Sep  4 03:21:30 dnssrv named[6498]: client 10.48.254.12#33560/key domainupdate-key: view internal: updating zone 'challenge.example.org/IN': adding an RR at '_acme-challenge.challenge.example.org' TXT
Sep  4 03:21:30 dnssrv named[6498]: zone challenge.example.org/IN/internal: sending notifies (serial 34)
Sep  4 03:21:30 dnssrv named[6498]: client 192.0.2.1#57933/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR started: TSIG internal-key
Sep  4 03:21:30 dnssrv named[6498]: client 192.0.2.1#57933/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR ended
Sep  4 03:21:31 dnssrv named[6498]: client 193.180.23.242#56325/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR started: TSIG internal-key
Sep  4 03:21:31 dnssrv named[6498]: client 193.180.23.242#56325/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR ended
Sep  4 03:22:35 dnssrv named[6498]: client 10.48.254.12#33562/key domainupdate-key: view internal: updating zone 'challenge.example.org/IN': deleting an RR at _acme-challenge.challenge.example.org TXT
Sep  4 03:22:35 dnssrv named[6498]: zone challenge.example.org/IN/internal: sending notifies (serial 35)
Sep  4 03:22:35 dnssrv named[6498]: client 192.0.2.1#41738/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR started: TSIG internal-key
Sep  4 03:22:35 dnssrv named[6498]: client 192.0.2.1#41738/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR ended
Sep  4 03:22:35 dnssrv named[6498]: client 193.180.23.242#36155/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR started: TSIG internal-key
Sep  4 03:22:35 dnssrv named[6498]: client 193.180.23.242#36155/key internal-key (challenge.example.org): view internal: transfer of 'challenge.example.org/IN': IXFR ended
```

systemcrash:
Hmm, while the pull request is valid for this particular problem, I run into problems when I attempt to run commands with multiple (sub)domains whose each _acme-challenge TXT pointer points to the same challenge.example.com (so DNS updates simply produce two TXT pointers, otherwise indistinguishable, but for their content):

```
certbot certonly --dns-rfc2136 -d mail.example.com -d webmail.example.com --dns-rfc2136-credentials /root/.secrets/certbot/rfc2136.ini --agree-tos --dns-rfc2136-propagation-seconds 9
```

So running dig
```
;; ANSWER SECTION:
_acme-challenge.challenge.example.com. 120 IN TXT	"XeelfKYTRD0wtM0q9coPLkz7evQWQhRJSn6UoTaFG-0"
_acme-challenge.challenge.example.com. 120 IN TXT	"e9_CVNy0C3njReesqqqdXPiUyc3mgVGL2l_xlBam_H8"
```

Certbot probably cannot handle two unique keys yet. Cerbot outputs:
```

   Domain: mail.example.com
   Type:   unauthorized
   Detail: Incorrect TXT record
   "qgoa5jJJ7TkArEcrPqCURefBFsYUZrC4Lwqh-_-_2bU" (and 1 more) found at
   _acme-challenge.mail.example.com
```

Giving cerbot up to 3 'chances' to encounter a correct key (is that the amount of subdomains possible in a single cert request?) for a request when looking for such DNS TXT records while still mandating that that amount of correct keys = amount of requests would still work.

hpax:
I wanted to add something with regards to pathological cases. I personally believe that the best way to deal with those is to allow the _acme-challenge record to contain a URI record (RFC 7553) which would allow the challenge to be posted to an arbitrary location on an arbitrary server. The "arbitrary location" is important, because there is no guarantee, nor is there a need in this case, for the domain owner to have control over the http root.

This is obviously a change to the ACME protocol, as opposed to a localized change in certbot, but I suspect that consider it sooner rather than later is probably a good idea.

The advantage of using a URI record is that the URI record can remain static, and the dynamic aspect is completely divorced from DNS (including potential issues with propagation delays, DNSSEC, users who might not be able to create a dynamic zone, and so on) and just requires access to any part of the namespace of a world-facing http server, which could even be special-cased for this purpose and possibly run on a nonstandard port (one use for this could be the certbot standalone webserver run on a nonstandard port.)

The existing web server protocol is not adequate, because:

1. It requires that the challenge is put on a web server with exactly the same name as the certificate being sought.
2. It requires the use of a well-known URL.

This makes it impossible to execute a challenge for a hostname which has no web server running, e.g. for a dedicated mail server, and in the case of an outsourced web service, only part of the domain name might be accessible to the user.

hpax:
There was a comment in here which I can't seem to find anymore (deleted?) about having multiple domains CNAME to the same record breaking the letsencrypt server.

The ACME spec for dns-01 explicitly say that multiple TXT records for the same name are permitted, and that the server should accept the request if any one of them is correct, so this would be a bug in the server. That being said, this is probably not the best idea.

duck-rh:
@hpax what you're suggesting is interesting but IMHO is entirely another method of solving the challenge. What you worked on in this PR would already help a lot of people and has much better chance to be accepted quickly.

I have tested your patch on top of 0.36 (0.35.1 for dns-rfc2136) packages from Debian unstable, and rebuilt for Buster, and except for a minor error in the test which I worked around by commenting the `logmsg += ': '+dns.rcode.to_text(rcode)` line (see attached file for the log) it built fine with all tests OK. I tested it on my infra for Postfix certificates, one with a single name and two with an alias and all went fine. Thanks a lot for your work.

[7244_tests_failure.txt](https://github.com/certbot/certbot/files/3674718/7244_tests_failure.txt)

hpax:
You are quite correct that it is a completely different solution, at the protocol level, and which has totally different properties. I think the Right Thing[TM] is to do this client only solution first.

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -174,49 +170,152 @@ def _find_domain(self, record_name):
...
+            elif rr.rdtype == dns.rdatatype.SOA:
+                # We found an SOA, authoritative or not
+                found = (auth, prefix, suffix)
+            else:
+                # We found a DNAME or CNAME. We need to start the walk over
+                # from the common point of departure.
+                target = rr.target
+                if target in domain_names_searched:
+                    # DNAME/CNAME loop!
+                    errors.PluginError('{0} {1} loops seeking SOA for {2}'
```
bmw: Looks like we're missing a `raise` here.

bmw:
Per our [postmortem a few months ago](https://docs.google.com/document/d/1CDFazf8MfHa5NA4BxrU0mLMrg1TD03TDdg8cO1Kjb8w/edit?usp=sharing) this PR will need a 2nd review before being merged.

adferrand:
I will note here also that two different Certbot users confirmed that this PR is working as expected.

systemcrash:
plz merge

duck-rh:
@bmw if community reviews does not count, then could a core dev take care of this extra review then please?

bmw:
Yep. The few people who are paid to work on Certbot (most of whom only work on the project part time) are stretched pretty thin but we'll take a look at this when we can!

systemcrash:
@hpax suggestion for `CHANGELOG.md`: I'm rad, and I finally made certbot_dns_rfc2136 work. Done.

systemcrash:
Does this fix need to hang in limbo indefinitely because @hpax is possibly too busy/awol to write something up for `CHANGELOG.md`? Surely this is trivial and can be fixed later?

bmw:
It's waiting on a 2nd review from a core Certbot maintainer and we've had other commitments keeping us busy such as the [work funded by OTF](https://www.opentech.fund/results/supported-projects/certbot-improvements/).

We're hoping to spend the first couple months of next year working through our PR queue though so hopefully this will be reviewed and merged then. Sorry for the delay!

duck-rh:
@bmw any news?

bmw:
> @bmw any news?

No good news unfortunately. We had some things come up and as you can see, the cleanup of outstanding community PRs did not happen during the first couple months of this year. Even worse, we don't have a plan to do it in the near future as the Certbot team is stretched especially thin.

This and other community PRs are on our radar and discussed often, but I honestly wouldn't expect them to land in the next few months. If you'd like the feature sooner, you could package the code as a [third party plugin](https://certbot.eff.org/docs/using.html#third-party-plugins) and use it there.

duck-rh:
Just a follow-up on testing this PR.

As I need this patch on Red Hat systems and this PR is blocked I worked on adapting the python-certbot-dns-rfc2136 package which is based on 1.3.0. I still have the same problem reported earlier with the logmsg line to make it build. I commented it and it now builds on Fedora (F30, F31, F32, Rawhide) as well as EL8. Bumping dnspython to 1.16.0 solved the build on EL7.

I have tested this package including this PR successfully on EL7 and EL8.

systemcrash:
Oh cool. This is, as far as I can tell, the *one* thing holding me back
from upgrading from el7 to 8. Well, the fact I cannot manage certs via DNS.
Certbot is still blind on 8.


Stay well, everyone.

duck-rh:
@systemcrash I am a Red Hat employee but I'm not in charge of this package, so I made a build for community infrastructures I'm working on. You can nevertheless use our public repository: https://copr.fedorainfracloud.org/coprs/duck/osas-infra-team-rpm-repo/

The package is python{2,3}-certbot-dns-rfc2136 but I would advise to also grab the rebuild of certbot (to be sure to have them both in sync at the same version) as well as python-dns (for some fixes).

duck-rh:
I forgot to say I did the same port for Debian Buster for DuckCorp: https://repository.duckcorp.org/debian/

I guess that may help people in need and also have more testers.

anarcat:
i confirm this patch works.

duck-rh:
@bmw Since we now have a second review, what would be missing to get things going forward?

systemcrash:
An accept, I guess. Is the review approval from the team?

bmw:
Yeah that second review is not from a member of the Certbot team.

Unfortunately, the status of this and many other PRs from open source contributors is the same as what I wrote a couple months ago [above](https://github.com/certbot/certbot/pull/7244#issuecomment-602850873).

systemcrash:
The review is then a clear signal from the user base that... Where things
didn't work before now, they're working out nominally with this PR. This
goes for me also.

anarcat:
> Yeah that second review is not from a member of the Certbot team.

I confirm that I am not a member of the certbot team. :p I still hope the contribution is valuable.

> Unfortunately, the status of this and many other PRs from open source contributors is the same as what I wrote a couple months ago [above](https://github.com/certbot/certbot/pull/7244#issuecomment-602850873).

At this stage, we're in this uncomfortable situation that we have a manually deployed certbot in a virtualenv on a production server. It's basically a fork, with this patch, so it means extra churn for us. So either we really fork the rfc2136 plugin or all of certbot or just write a new hook for dehydrated or something, all of which seem unsatisfactory. :(

bmw:
Yeah I appreciate external contributors taking the time to test this as it helps catch potential problems.

I understand the situation here and in many other open issues and PRs is frustrating and I am genuinely working on resolving it. If another solution works better for you, I would encourage you to use it. The Certbot team is just stretched a bit thin at the moment, but we're working on expanding the team and catching up and for both of us, I hope that happens sooner rather than later.

duck-rh:
@bmw thanks for your reply

I don't think forking for good is a good idea, especially since several other plugins use DNS and would probably benefit from such behavior. To me this is a first step and more generic methods could be introduced in the core to provide a common framework for them If we fork now that's going to be a mess to reconcile I fear.

anarcat:
> I don't think forking for good is a good idea, especially since several other plugins use DNS and would probably benefit from such behavior.

I want a pony too, but at this point, I do need to scratch that itch first. ;) Maintainers here did suggest the dns-rfc2136 plugin be split out of the certbot core - maybe it would make sense to fork just that? Or maybe more plugins could simply moved out to a separate repository?

certbot-dns-rfc2136/certbot_dns_rfc2136/_internal/dns_rfc2136.py
```diff
@@ -174,53 +170,156 @@ def _find_domain(self, record_name):
...
+            elif rr.rdtype == dns.rdatatype.SOA:
+                # We found an SOA, authoritative or not
+                found = (auth, prefix, suffix)
+            else:
+                # We found a DNAME or CNAME. We need to start the walk over
+                # from the common point of departure.
+                target = rr.target
+                if target in domain_names_searched:
+                    # DNAME/CNAME loop!
+                    errors.PluginError('{0} {1} loops seeking SOA for {2}'
```
anarcat, line 224:
```suggestion
                    raise errors.PluginError('{0} {1} loops seeking SOA for {2}'
```

duck-rh: good catch

duck-rh:
For people using the Red Hat OSPO repository: 1.4.0 landed in EPEL and 1.5.0 is in EPEL testing, so the patch vanished with the upgrade. I recently updated the packages with a patched 1.5.0 and it is working fine so far.

systemcrash:
So maybe a year later this PR will get merged and fix all of our problems. Anyway, the code is proven to be good. I recommend just mashing this merge in there, because... well, a year.

systemcrash:
ping (`pr-test-suite Expected — Waiting for status to be reported `)

mikma:
I had a look at the CNAME test case (RFC2136ClientTest.test_find_domain_cname) and it doesn't seem right.

According to the test the call  _find_domain('foo.bar.cname.'+DOMAIN) should return prefix='foo.bar' and domain='subzone'+DOMAIN.

But there is no CNAME record that matches 'foo.bar.cname.'+DOMAIN. The only CNAME record is 'cname' + DOMAIN, which doesn't match and it won't be returned by the DNS server when looking up 'foo.bar.cname.'+DOMAIN.

systemcrash:
Pretty plz 😄

certbot-dns-rfc2136/certbot_dns_rfc2136/_internal/dns_rfc2136.py
adferrand: Fixed in https://github.com/certbot/certbot/pull/7244/commits/2b5f7ceed2c7115540e33ff06f243220dd8cb0fb

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
adferrand: Fixed in https://github.com/certbot/certbot/pull/7244/commits/2b5f7ceed2c7115540e33ff06f243220dd8cb0fb

adferrand:
@mikma, you are right, a CNAME is not always set for the complete name. I have myself this kind of situation in some of my DNS zones. I added your contribution on this PR, thanks a lot!

About using the first system resolver vs the target RFC2136 server, it is subject to discussion. On one side it could increase the resiliency if the target RFC2136 server does not support recursion. On the other side, it adds another DNS server in the picture, that could itself be flawed, and may be out of reach of user configuration. Given that the plugin is already sensible to good behavior of the target RFC2136 server, I prefer to keep it for SOA queries and add another risk by adding another external asset.

adferrand:
Another note. With #8448 merged, we can now add actual integration tests for this plugin.

adferrand:
Another note (2). I discussed with @alexzorin about this feature. He raised concerns about what could happen with badly configured RFC2136 DNS servers, that could alter Certbot behavior for some people without warn if the feature is integrated.

He suggested to turn it into an opt-in feature with an appropriate CLI flag.

systemcrash:
> Another note. With #8448 merged, we can now add actual integration tests for this plugin.

Does this concern the actual dns-rfc2136 plugin, or specifically this PR?

systemcrash:
> Another note (2). I discussed with @alexzorin about this feature. He raised concerns about what could happen with badly configured RFC2136 DNS servers, that could alter Certbot behavior for some people without warn if the feature is integrated.
> 
> He suggested to turn it into an opt-in feature with an appropriate CLI flag.

Down to the admin, really. As long as they're cognizant of what's necessary when they run this plugin. 

Is your change request still valid, given that your rebase fixed it?

adferrand:
> 
> 
> > Another note. With #8448 merged, we can now add actual integration tests for this plugin.
> 
> Does this concern the actual dns-rfc2136 plugin, or specifically this PR?

On master there is already a first integration test for a straight-forward case. Now this can be enriched in the scope of this PR to cover the recursion model proposed here.

adferrand:
> 
> 
> > Another note (2). I discussed with @alexzorin about this feature. He raised concerns about what could happen with badly configured RFC2136 DNS servers, that could alter Certbot behavior for some people without warn if the feature is integrated.
> > He suggested to turn it into an opt-in feature with an appropriate CLI flag.
> 
> Down to the admin, really. As long as they're cognizant of what's necessary when they run this plugin.
> 
> Is your change request still valid, given that your rebase fixed it?

Which fix are you referring to? In general the opt-in through a flag is still relevant in my opinion, given that if the admin of the DNS server knows exactly what they are doing, users of this server may not.

systemcrash:
> > > Another note (2). I discussed with @alexzorin about this feature. He raised concerns about what could happen with badly configured RFC2136 DNS servers, that could alter Certbot behavior for some people without warn if the feature is integrated.
> > > He suggested to turn it into an opt-in feature with an appropriate CLI flag.
> > 
> > 
> > Down to the admin, really. As long as they're cognizant of what's necessary when they run this plugin.
> > Is your change request still valid, given that your rebase fixed it?
> 
> Which fix are you referring to? In general the opt-in through a flag is still relevant in my opinion, given that if the admin of the DNS server knows exactly what they are doing, users of this server may not.

Your 'changes requested' on this PR. Which are done...?

AlexanderplUs:
Any news? I have to apply this changes by hand after each Certbot release which is annoying.

systemcrash:
Same.....

duck-rh:
@adferrand what is the status here? I think you can imagine the level of frustration. It really does not feel one can contribute to this project.

> Another note (2). I discussed with @alexzorin about this feature. He raised concerns about what could happen with badly configured RFC2136 DNS servers, that could alter Certbot behavior for some people without warn if the feature is integrated.
> 
> He suggested to turn it into an opt-in feature with an appropriate CLI flag.

I'm fine with a CLI flag but I would have been interested to know about which situations you're talking about. I have no clue what kind of situation would only break when using this plugin while having the user's service working fine.

bmw:
I wouldn't expect this PR to be merged for at least a few more months.

In the next 6 months or so, we're planning on rethinking our approach to DNS plugins in Certbot since the current situation is not maintainable by the small team here. Until then, we're not planning on adding any new plugins or significant features to existing ones.

I know this isn't what you wanted to hear and that it'd be easier for you all if this change landed in our official plugin, but for now if you need this feature I recommend maintaining a 3rd party fork of the plugin. You can find examples of other 3rd party plugins as examples to build off of if you want at https://certbot.eff.org/docs/using.html#third-party-plugins.

hpax:
This once again brings up something I have pointed out in the past: runtime modification of the DNS entries should not be necessary requiring all these plugins.

I would once again like to propose a new challenge protocol added to ACME where DNS ownership is indicated by an URL RR at _acme-challenge indicating an https server to which the verification is delegated. That can be done statically, and the rest reverts to http-01 protocol, but does not need an https server on the actual host the name of which needs to be certified, nor does it require control of the /.well-known URL, or a server on port 80 (as opposed to, say, 443, or a custom port for a transient certbot buildin server.)

It is really too sad that this has been blocked for this long, given that this is in fact stated explicitly on LetsEncrypt as a supported mechanism:

"Since Let’s Encrypt follows the DNS standards when looking up TXT records for DNS-01 validation, you can use CNAME records or NS records to delegate answering the challenge to other DNS zones. This can be used to delegate the _acme-challenge subdomain to a validation-specific server or zone. It can also be used if your DNS provider is slow to update, and you want to delegate to a quicker-updating server."

everlanes:
Seriously? Two years now, a working pull request and no intention to merge it?

The normal Open Source way is: "If you want that feature, then implement it yourself!" Here someone has implemeneted it and ... oh .. nothing happes. "We are too busy to merge it, while thinking about problems that fix could cause..."

How else should others help you? What support do you need, ie. what has to be done, that this pull request will be merged soon? If you are not interested in anyone helping you, please state it clearly!

```
If you like Certbot, please consider supporting our work by:
 * Donating to ISRG / Let's Encrypt:   https://letsencrypt.org/donate
 * Donating to EFF:                    https://eff.org/donate-le
```
At the moment, I don't like it - as it is not working.

systemcrash:
Can we plz get this to work? Just replace the current whatever of a plugin with @hpax version. Works.

hpax:
FWIW, for me personally I have long since given up and switched to using external scripts.

segdy:
Won’t make a difference as it seems but I’m super sad as well that such an important bug can’t be fixed within 5 years (2018 is the first reference) especially that everything is done and ready.

The current DNS challenge with either manual updates or requiring to make the whole zone dynamic/updateable is not at all a solution.

mdella-nutanix:
Did a bunch of research and for an enterprise with hundreds of "internal only" TLS sites that need to use something other than HTTP-01 updates, this is one of the only ways to accomplish certs that will never be seen outside of the company.  Its unfortunate that we had to commit resources to basically pull a 4 year old PR, and merge it with the current certbot just to get this method working.  This also means that we have to dedicate someone to constantly merge the code in and regenerate "custom" RPM packages to deploy this code.

adferrand:
In the light of what I did with #9838, I would like to reassess what we should do about this PR. In previous discussions we said that this approach was extremely useful, and could even be generalized to all DNS plugins.

There are two things to cover:
1) what is the actual name of the DNS zone to target (ex. `sub.domain.net` is actually a DNS zone on its own whose delegation is declared in the DNS zone `domain.net` _vs_ `www.domain.net` is defined by the `www` entry in the DNS zone `domain.net`.
2) what is the canonical name of the FQDN of a challenge taking into account any CNAME/DNAME (eg. `_acme_challenge.www.domain.net` is canonically `challenge.other-domain.net` because of a CNAME declared in the relevant DNS zone for `www.domain.net`).

Both approaches are covered by dedicated functions in `dnspython`.
* For 1), one can use `dns.resolver.zone_for_name()`.
* For 2), one can use `dns.resolver.canonical_name()` (since `dnspython` > 2.1.0).
* Chaining 1) and 2):  `dns.resolver.zone_for_name(dns.resolver.canonical_name(name))`.

I said in #9838 that 1) has to be done for Lexicon-based DNS plugins because of the need of Lexicon to know the actual DNS zone name (and 1) is actually done by Lexicon directly), and that 2) should be the responsibility of the ACME client because a CNAME delegation of `_acme_challenge` has its purpose defined for the DNS-01 challenge, not the underlying client & API DNS provider.

As a side note, I also does 2) in my own project `dnsrobocert` for the same reasons.

So I think we should close both #7244 and #6644 in favor of a new approach to use at least `dns.resolver.canonical_name()` and potentially `dns.resolver.zone_for_name()` and resolve the actual FQDN and/or zone at a higher level, most likely in `certbot.plugins.dns_common.DNSAuthenticator`. My preference goes to do only CNAME resolution, and leaves zone resolution to the specific needs of a DNS provider implementation (eg. covered for all Lexicon-based plugins with #9838). Also the new behavior should be optional, to avoid to conflict with any workaround that users have certainly developed over the years this feature has stayed unimplemented.

cc @bmw

timriker:
Questions about this PR. I'd like to setup something like:

```
validate.example.com is the domain with remote dns keys
_acme-challenge.example.com CNAME to example.com.validate.example.com.
_acme-challenge.example.net CNAME to example.net.validate.example.com.
_acme-challenge.example.org CNAME to example.org.validate.example.com.
_acme-challenge.dev.example.net CNAME to dev.example.net.validate.example.com.
```
such that the TXT records for each domain can be created in validate.example.com and used for domain validation for a number of zones and sub-zones.

"validate" could probably be something like _rfc2136 to make it less likely to conflict with any name desired for use.

The code would need to detect a CNAME then follow the hostnames back till it finds a SOA record and add the TXT record there with the remaining names. ie: dev.example.net in the zone validate.example.com and then remember and delete that entry after validation.

From what I read of the spec, I think these entries will validate ok, it's just a matter of certbot understanding what's needed and doing it based on the pre-existing _acme-challenge CNAME.

Will this PR, or something like it, show up in certbot anytime soon? I'm on Ubuntu running:

- certbot 1.21.0-1build1
- python3-certbot 1.21.0-1build1
- python3-certbot-dns-rfc2136 1.21.0-1

timriker:
I pulled down this PR, patched it for current, and tested things out. I does not seem to handle a CNAME like:

updated the PR in my fork. [timriker:timriker/rfc-2136-cname-fix](https://github.com/timriker/certbot/tree/timriker/rfc-2136-cname-fix)

lelutin:
I don't know anything to add wrt the more generalized approach for dns plugins (I guess it could make sense?)

but I do want to add that I've just used the code from this PR and it did fix the issue I was having with a CNAME that intended to point certbot to a sub-zone for the dynamic records.

dkg:
Also disappointing that this code has not been merged yet.   I don't really understand the hesitation.

If the concern is that this is somehow too risky, perhaps you'd be willing to accept a change request that would at least provide a visible error when a CNAME is detected when looking for the SOA?

timriker:
Merged updates from master on my fork with HPA's updates. Still working. Please merge.

[timriker:timriker/rfc-2136-cname-fix](https://github.com/timriker/certbot/tree/timriker/rfc-2136-cname-fix)

Should I open this as a separate PR? @hpax are you still using this? Are you planning to merge updates on your fork/PR?

hpax:
Oh, I gave up on it years ago and replaced it with external scripts.

geektoor:
So i ran into the same issue. I configured my domain with a CNAME to redirect it to a special subdomain. And it failed. Then i checked the code, found that basically nothing would respect the CNAME. After a short google search, i ended up here.

So this request is now 5 years old (behold, next year we need to send it to School), i was really surprised that it never made it.

I just tested it with the branch from @timriker and it just worked as expected. So i wonder why we don't just finish this PR? Whats exactly missing?

PR #7246, opened 2019-07-15
[Apache v2] New ParserNode interface abstraction

joohoi:
This is the first PR of many to come. It contains the interface abstraction that the ParserNode implementation will be built upon.


certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+    management methods.
+    """
+
+    @abc.abstractmethod
+    def add_child_block(self, name, arguments=[], position=-1):
+        """
+        Adds a new BlockNode child node with provided values and marks the callee
+        BlockNode dirty. This is used to add new children to the AST.
+
+        :param str name: The name of the child node to add
```
bmw: What is the meaning of this `name` parameter for each of block, directive, and comments? Does the `name` also appear in the list of arguments? If so, is it ever not the first argument?

certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+    and other blocks as its children. A BlockNode can have zero or more arguments
+    attached to it.
+
+    Configuration blocks typically consist of one or more child nodes of all possible
+    types. Because of this, the BlockNode class has various discovery and structure
+    management methods.
+    """
+
+    @abc.abstractmethod
+    def add_child_block(self, name, arguments=[], position=-1):
```
bmw: This is definitely a nitpick, but I don't love how the position parameter deviates from `list.insert` in the standard library. As an example:
```
In [1]: l = [0,1,2]                                                                                                                                                                                                                                                                                                                                                                                                                   

In [2]: l.insert(-1, 3)                                                                                                                                                                                                                                                                                                                                                                                                               

In [3]: l                                                                                                                                                                                                                                                                                                                                                                                                                             
Out[3]: [0, 1, 3, 2]
```
What if we keep the behavior the same but make the default value of `position` `None` which appends to the end of the list?

certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+        """
+        Adds a new BlockNode child node with provided values and marks the callee
+        BlockNode dirty. This is used to add new children to the AST.
+
+        :param str name: The name of the child node to add
+        :param list arguments: list of arguments for the node
```
bmw: I think we should fully specify the expected contents of `arguments` here and for the other object types.

certbot/interfaces.py
bmw: Also to satisfy our linters concerns (which I think are valid), we should make the default value of `arguments` `None` as well.

certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+    different files in the configuration structure.
+
+    While for the most parts the usage and the child types are obvious, "include"-
+    and similar directives are an exception to this rule. This is because of the
+    nature of include directives - which unroll the contents of another file or
+    configuration block to their place. While we could unroll the included nodes
+    to the parent tree, it remains important to keep the context of include nodes
+    separate in order to write back the original configuration as it was.
+    """
+
```
bmw: Will a `ParserNode` (or any of its subclasses) have any public attributes? If so, I think we should define them using an approach like: https://stackoverflow.com/questions/2736255/abstract-attributes-in-python

certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+    def find_blocks(self, name, arg=None):
+        """
+        Find a configuration block with name and (optionally) an argument.
+        This method walks the child tree of ParserNodes under the instance it was
+        called from. This way it is possible to search for the whole configuration
+        tree, when starting from root node or to do a partial search when starting
+        from a specified branch.
+
+        :param str name: The name of the directive to search for
+        :param str arg: Return only directives that have argument with provided value
```
bmw: Should the `arg` parameter here and in `find_directives` below be a string or a list? Will we ever want to specify more than one argument to search for?

certbot/interfaces.py
```diff
@@ -604,3 +604,204 @@ def renew_deploy(self, lineage, *args, **kwargs):
...
+    Nodes keep track of their differences to on-disk representation of configuration
+    by marking modified ParserNodes as dirty to enable partial write-to-disk for
+    different files in the configuration structure.
+
+    While for the most parts the usage and the child types are obvious, "include"-
+    and similar directives are an exception to this rule. This is because of the
+    nature of include directives - which unroll the contents of another file or
+    configuration block to their place. While we could unroll the included nodes
+    to the parent tree, it remains important to keep the context of include nodes
+    separate in order to write back the original configuration as it was.
```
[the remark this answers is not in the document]
joohoi: My initial thought was to have `Include*` acting as a BlockNode, the children of which would be the contents of the included file(s).

certbot/interfaces.py
joohoi: Name is the directive "key", for `<VirtualHost *:80>` the name would be `VirtualHost`, for `Listen 80` the name would be `Listen`. It's not in the argument list.

certbot/interfaces.py
joohoi: Using `None` as the default value for both makes sense, these are changed now.

certbot/interfaces.py
joohoi: I wonder if the definition could be done once in the main docstring for the class. The list of arguments is a list of strings that can be basically anything depending on the configuration structure. I find it hard to word this in the method documentation in a way that's not overly vague or too elaborate.

certbot/interfaces.py
joohoi: Good point. I'll add those. Examples of public attributes would be `ancestor` and `children`.

certbot/interfaces.py
joohoi: I can't think of a case where we would in Apache land. But to make this more generic, a list works better here.

joohoi: From the interface design perspective however, I might be leaning towards removing the arg argument completely, and directing users to do post processing for the directives in order to find the correct ones. This would give more freedom for the user. For example:

```
for vh in rootnode.find_blocks("VirtualHost"):
    if "*:80" in vh.arguments():
        do_a_thing()
```

certbot/interfaces.py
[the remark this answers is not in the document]
bmw: Works for me! I think we should explicitly document this either here or in the Google doc.

certbot/interfaces.py
bmw: Do comments have a meaningful `name`? If not, I think we should remove the parameter from `add_child_comment` and only have a `name` attribute for block and directive nodes.

certbot/interfaces.py
bmw: SGTM! We can always add this functionality back if the need/desire arises.

certbot/interfaces.py
bmw: Sure. I'm not currently not too worried about how it's documented, I just think we should carefully document how we're imagining things will work for each node type so we can all be on the same page when implementing the interface and writing code that uses it.

If you want to document this elsewhere and we're not planning on doing any processing on this list before creating the resulting node, maybe we could say something here like "list of arguments which will be on the resulting node" to try and make it more explicit that it's up to the caller to get this list right.

certbot/interfaces.py
joohoi: SGTM

certbot-apache/certbot_apache/parserassert.py
```diff
@@ -0,0 +1,69 @@
+"""Runtime assertions for two different parser implementations"""
```
bmw: nit: Can we add this in a second PR after deciding on an interface for now?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,281 @@
...
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def save(self, msg):
+        """
+        Save traverses the children, and attempts to write the AST to disk for
+        all the objects that are marked dirty. The actual operation of course
+        depends on the underlying implementation. save() shouldn't be called
+        from the Configurator outside of it's designated save() method in order
```
bmw: English insanity nit:
```suggestion
        from the Configurator outside of its designated save() method in order
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,281 @@
...
+    @abc.abstractmethod
+    def add_child_block(self, name, arguments=None, position=None):
+        """
+        Adds a new BlockNode child node with provided values and marks the callee
+        BlockNode dirty. This is used to add new children to the AST.
+
+        :param str name: The name of the child node to add
+        :param list arguments: list of arguments for the node
+        :param int position: Position in the list of children to add the new child
+            node to. Defaults to None, which appends the newly created node to the list.
```
bmw: nit: I'd like to add a sentence to each `add*` method for the parameter here like "If an integer is given, the child is inserted before that index in the list just like `list.insert`."

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,281 @@
...
+    configuration structure. Because of the nature of comments, it is not able
+    to have child nodes and hence it is always treated as a leaf node.
+
+    CommentNode stores its contents in class variable 'arguments' and does not
+    have a specific name.
+    """
+
+    @property
+    @abc.abstractmethod
+    def comment(self):
```
bmw: Is this different from `arguments?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,281 @@
...
+        :returns: list of file paths of files that have been changed but not yet
+            saved to disk.
+        """
+
+    @property
+    @abc.abstractmethod
+    def children(self):
+        """
+        This property contains a list ParserNode objects that are the children
+        for this node.
```
bmw: Couple nits:

1. I assume this list has the same order as the directives found in the block. I think it'd be good to add a line somewhere saying this explicitly.
2. For the properties we added, I think it'd be good to document the return type. I'm mainly thinking of this for cases like this where we may want to make this a `tuple` to help prevent incorrect modifications.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,281 @@
...
+
+    The node type would be BlockNode, name would be 'VirtualHost' and arguments
+    would be: ['*:80'].
+
+    While for the following example:
+
+        LoadModule alias_module /usr/lib/apache2/modules/mod_alias.so
+
+    The node type would be DirectiveNode, name would be 'LoadModule' and arguments
+    would be: ['alias_module', '/usr/lib/apache2/modules/mod_alias.so']
```
bmw: I personally think this is ideal if we can make this work sanely with whitespace, but if not, we should remember to update this comment if our plan changes.

certbot-apache/certbot_apache/interfaces.py
joohoi: It's basically the same, but returning a string instead of list. I find this more natural for comments. If you disagree, we can definitely just handle the contents as arguments here too.

joohoi:
> My proposal as a starting point of this conversation is that the case in the ParserNode objects is the same as it appears on disk but when searching for directives, we do a case insensitive comparison. This may be done by doing something like converting everything to lowercase, however, the returned results should be in the original case.

This sounds good, as Apache is case insensitive, at least in case of directive and block names and such.

certbot-apache/certbot_apache/interfaces.py
joohoi: What do you think? If we want to use `arguments` property instead, especially with the current plan of writing a whitespace-hiding list extension, I feel that `comment` would suit it better. If we decide to do this, I'll move the `arguments` property to `DirectiveNode` and `BlockNode` interfaces instead of the `ParserNode`.

bmw: The plan in your most recent comment SGTM!

My only question becomes how do we update the comment?

Proposal: Allow the user to change the `comment` attribute directly. This should be pretty easy to implement by giving the comment a setter property (which we maybe can/should define in the ABC here for clarity?) so we can set the dirty bit when the comment is modified.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,334 @@
...
+
+    @property
+    @abc.abstractmethod
+    def children(self):
+        """
+        This property contains a list ParserNode objects that are the children
+        for this node. The order of children is the same than that of the parsed
+        configuration block.
+
+        :returns: A tuple of this blocks children
```
bmw:
```suggestion
        :returns: A tuple of this block's children
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+@six.add_metaclass(abc.ABCMeta)
+class CommentNode(ParserNode):
+    """
+    CommentNode class is used for representation of comments within the parsed
+    configuration structure. Because of the nature of comments, it is not able
+    to have child nodes and hence it is always treated as a leaf node.
+
+    CommentNode stores its contents in class variable 'arguments' and does not
```
bmw: If we remove `arguments` from comments, we should remember to update this comment.

bmw:
If you want to add tests for these interfaces, feel free, but I'm also fine personally with just not requiring test coverage on the interfaces file.

certbot-apache/certbot_apache/tests/parsernode_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def set_parameters(self, parameters):  # pragma: no cover
+        pass
+
+
+class DummyBlockNode(interfaces.BlockNode):
+    """ A dummy class implementing BlockNode interface """
+    ancestor = None
+    parameters = tuple()  # type: Tuple[str, ...]
+    children = tuple()  # type: Tuple[str, ...]
```
bmw: Shouldn't this be a tuple of ParserNode objects?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -135,6 +125,28 @@ def name(self):
...
+        :rtype: tuple
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def set_parameters(self, parameters):
+        """
+        Sets the sequence of parameters for this ParserNode object without
+        whitespaces, and marks this object dirty.
```
bmw: I think we should describe in the docstrings of the `set_parameters` functions here or in the Google doc about what whitespace is preserved when this method is called. I think we decided that whitespace preceding the directive name is unaffected.

certbot-apache/certbot_apache/tests/parsernode_test.py
joohoi: It absolutely should. That somehow got lost while battling the mypy syntax.

certbot-apache/certbot_apache/interfaces.py
joohoi: The way I see it, this information belongs to the docstring of `ParserNode` itself. I'm imagining the whitespacing and metadata of the directives, blocks and comments to exist in the domain of the "main object" instead of its parameters.

I'll add it to the `ParserNode` docstring for now, and we can move it later if you disagree with this.

joohoi:
Are you referring to adding more documentation to the docstrings, or as a comment or description of this PR?

certbot-apache/certbot_apache/interfaces.py
bmw: What you added to the `ParserNode` docstring seems fine. (Although while you don't have to change it, it's a little general for my liking. I thought we agreed we weren't going to worry about trying to generalize this interface for more than Apache right now?)

I personally would still like a sentence saying explicitly that `set_parameters` doesn't affect the whitespace preceding the directive itself though. I think that makes most sense to include here because we're talking about the behavior of this function, but it doesn't matter to me as long as it is documented in the file.

bmw:
I think there should be a single location where we can look to get the full description of how this interface should work.

Unless we want to delete the ABC here entirely (which I am fine with), I have a preference for doing it in the interfaces file in this PR. This because if we have both the code here and an explanation somewhere else, we're taking on the problem of keeping the information in sync which we won't have to do otherwise.

While I have a strong preference of merging all information into a single location, if we don't do that (and you don't want to delete the ABC), I think the source of truth should be the Google Doc. I do not think it should just be written in the comments on GitHub.

bmw:
Weird. Rerunning AppVeyor tests on #7240 worked (because [AppVeyor/Travis merge the PR into the target branch before running tests](https://ci.appveyor.com/project/certbot/certbot/builds/26434046/job/6w373jmo5tq3oglh#L7)) but that doesn't seem to be working here.

Merging `master` into this branch should solve the problem though.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,458 @@
...
+        their contents. This property returns the state of this configuration block.
+        In case of unmatched conditional statement in block, this block itself should
+        be set enabled while its children should be set disabled.
+
+        :returns: True if the BlockNode is parsed and enabled in the configuration.
+        :rtype: bool
+        """
+
+    @abc.abstractmethod
+    def find_blocks(self, name, exclude=True):
```
bmw: You talked about at a higher level, but I think we should say explicitly in the docstring here and in `find_directives` that the search is case insensitive.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,458 @@
...
+        :param bool exclude: If the search results should exclude the contents of
+            ParserNode objects that reside within conditional blocks and because
+            of current state are not enabled.
+
+        :returns: A list of found DirectiveNode objects.
+
+        """
+
+    @abc.abstractmethod
+    def find_comments(self, comment, exact=False):
```
bmw: Is `find_comments` case sensitive? I personally think it's worth saying this explicitly here.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,458 @@
...
+
+    The node type would be DirectiveNode, name would be 'LoadModule' and its
+    parameters would be: ['alias_module', '/usr/lib/apache2/modules/mod_alias.so']
+
+    The applicable parameters are dependent on the underlying configuration language
+    and its grammar.
+    """
+
+    @abc.abstractmethod
+    def add_child_block(self, name, parameters=None, position=None):
```
bmw: If you still think it's valuable for the `add_*` functions to not add whitespace based on context to its children so that `apacheconfig` matches the behavior of Augeas until we can get rid of it, I think we should document this explicitly in each of these functions.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -1,4 +1,87 @@
...
+
+Whitespaces
+
+Each ParserNode object is responsible of storing its prepending whitespace characters
+in order to be able to write the AST back to filesystem like it was, preserving the
+format, this applies for parameters of BlockNode and DirectiveNode as well.
+When parameters of ParserNode are changed, the pre-existing whitespaces are discarded
+however, as the general reason for storing them is to maintain the ability to write
+the configuration back to filesystem exactly like it was. This loses its meaning when
+we have to change the directives or blocks parameters for other reasons.
```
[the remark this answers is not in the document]
joohoi: The usage is now clarified in both - the `set_parameters` docstring as well as the generic doc level docstring as well.

certbot-apache/certbot_apache/interfaces.py
joohoi: I think it's the best to not to differentiate the behavior through different lookup methods here. It has been now documented here as well.

certbot-apache/certbot_apache/interfaces.py
joohoi: This definition has been added to docstrings.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,476 @@
...
+While find_* interface methods allow including the disabled parts of the configuration
+tree in searches a special care needs to be taken while parsing the structure in
+order to reflect the active state of configuration.
+
+Whitespaces
+
+Each ParserNode object is responsible of storing its prepending whitespace characters
+in order to be able to write the AST back to filesystem like it was, preserving the
+format, this applies for parameters of BlockNode and DirectiveNode as well.
+When parameters of ParserNode are changed, the pre-existing whitespaces  in the
```
bmw:
```suggestion
When parameters of ParserNode are changed, the pre-existing whitespaces in the
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -0,0 +1,476 @@
...
+        :returns: BlockNode instance of the created child block
+
+        """
+
+    @abc.abstractmethod
+    def add_child_directive(self, name, parameters=None, position=None):
+        """
+        Adds a new DirectiveNode child node with provided values and marks the
+        callee BlockNode dirty. This is used to add new children to the AST. The
+        preceeding whitespaces should not be added based on the ancestor or siblings i
```
bmw:
```suggestion
        preceeding whitespaces should not be added based on the ancestor or siblings
```

joohoi:
I can't make much out of the [erroring tests report](https://travis-ci.com/certbot/certbot/jobs/224223075), but I assume the work in #7321 fixes this (albeit it having its own errors). As there's no changes landed to `apache-parser-v2` branch, I would personally be fine with rebasing this on `master` branch again, and I can modify the PR branches accordingly. That would seem like the least painful way forward.

bmw:
The errors on #7321 were on codecov because one of the AppVeyor builds had a network problem submitting coverage. I reran the test though and everything passed.

In the future, I think we'll want PRs like #7321 to bring over necessary fixes from `master` once the branches have diverged. Do you agree?

If so, I personally prefer just merging #7321 since the work is done, but for this time, if you prefer, I can temporarily turn off branch protections and bring `apache-parser-v2` up-to-date with `master` without the merge commit.

PR #7254, opened 2019-07-17
acme: Implement authz deactivation

### LINKED ISSUE #4945: Implement authorization deactivation in acme
The acme package should offer a method to deactivate authorizations. I don't think this needs to be part of the Certbot CLI, but it would be useful for the Boulder integration tester (chisel) and possible other users of the acme package.

### LINKED ISSUE #5116: --dry-run should prevent authz reuse
Some users get themselves into a situation where `--dry-run` works but production does not (see #3231 and #5040). This is due to the user having a valid authorization from staging which is reused, but not from production. This results in people having to use the production server to test their configuration.

We should do something to work around this problem. This probably involves deactivating valid authorizations so the challenge is actually checked by the CA. This requires obtaining the URL of a valid authorization and currently the only way I'm aware of to get the URL of a valid authz from Let's Encrypt that will be reused is to request new authzs for the same domains.

Perhaps we should implement a subcommand that requests a new authz from Let's Encrypt and deactivates it and/or deactivate any valid authorization returned during a `--dry-run` and automatically request a new one.

Related #4945 

cc @jsha, @SwartzCr

alexzorin:
Resolves #4945. First PR in order to address #5116.

~~@bmw I'm not sure what's going on with unit tests in the acme module. I tried to go by the example of `test_deactivate_account`, but it seems to be some kind of no-op test. Can I test the status change here somehow, or does it belong in integration?~~  I think I figured something out for the test.

## Pull Request Checklist

- [x] Edit the `master` section of `CHANGELOG.md` to include a description of the change being made.
- [x] Add [mypy type annotations](https://certbot.eff.org/docs/contributing.html#mypy-type-annotations) for any functions that were added or modified.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,6 +15,7 @@ Authors
 * [Alex Gaynor](https://github.com/alex)
 * [Alex Halderman](https://github.com/jhalderm)
 * [Alex Jordan](https://github.com/strugee)
+* [Alex Zorin](https://github.com/alexzorin)
 * [Amjad Mashaal](https://github.com/TheNavigat)
 * [Andrew Murray](https://github.com/radarhere)
 * [Anselm Levskaya](https://github.com/levskaya)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-*
+* acme: Authz deactivation added to `acme` module.
 
 ### Changed
 
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -126,10 +126,10 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
     def deactivate_authorization(self, authzr):
         """Deactivate authorization.
 
-        :param messages.AuthorizationResources authzr: The Authorization Resources
+        :param messages.AuthorizationResources authzr: The Authorization resource
             to be deactivated.
 
-        :returns: The Authorization resources that was deactivated.
+        :returns: The Authorization resource that was deactivated.
         :rtype: `.AuthorizationResources`
 
         """
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -133,7 +133,7 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         :rtype: `.AuthorizationResource`
 
         """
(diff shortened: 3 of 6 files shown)


alexzorin:
Should `deactivate_authorization` also raise an Error if the returned authz does not have the expected status? RFC8555 seems to imply that the response will come back with HTTP 200 if the deactivation was accepted.

bmw:
Thanks for the PR @alexzorin. I'm planning on reviewing this as soon as I can find some spare time. I'm hoping that will be later this week.

To answer your question though, `deactivate_authorization` should at least raise an error if it gets an error response which you should get for free by calling `self._post`. If you want to explicitly check for 200 as well, that's fine with me, however, I'm also fine with just relying on the generic check in `self._post`.

alexzorin:
>or only support authz deactivation with ACMEv2/RFC8555 which will complicate the Certbot code slightly because we can't drop support for ACMEv1 just yet

That was my original intent, since authz deactivation in Certbot would only occur with one of `--dry-run` or `--staging`, both implying ACME v2.

But turns out that actually adding an `UpdateAuthorization` class had the effect of fixing of the ACME v1 breakage, which I should have perhaps done to begin with: .

Thanks for the great hand holding :+1:.

bmw:
The `UpdateAuthorization` class has the slight downside that the resource field is sent with ACMEv2/RFC8555 which is not compliant, however, Let's Encrypt currently accepts this, we have this problem elsewhere, and I remain somewhat hopeful that we can find a general solution in #7176.

I don't have a strong opinion, but I think the approach you have here is my preferred solution. `UpdateAuthorization` follows the current pattern in the `acme` module and allows users of the library to call `deactivate_authorization` without worrying about the method not existing/working depending on the client object. You make a good point about `--dry-run`/`--staging` implying ACMEv2, however, we'll probably want a way to test this functionality outside of Let's Encrypt's staging servers and an always reliable `deactivate_authorization` function simplifies this a bit.

With all that said, I'm happy with the code here. If you are, feel free to mark it as ready for review and I'll merge it.

Thanks again @alexzorin!

alexzorin:
Alright. My thinking was to avoiding blowing up the pebble integration tests, but I'm sure we can skip the tests or something until pebble is fixed.

cpu:
Just wanted to add another round of thanks to this PR since it will help us on the boulder-side of life. :clinking_glasses: :tada: Thanks again!

Issue #6888, opened 2019-03-26
Can you explain this codecov report?


bmw:
See https://codecov.io/gh/certbot/certbot/pull/6784.

Why is there a drop in coverage on the manual, nginx, and standalone plugins? Those plugins aren't at all affected by the PR.

Is there a way to see what codecov is comparing the coverage against? I spent a few minutes poking around and couldn't quickly figure it out.

adferrand:
At this point, the status of this codecov report is correct in term of coverage, and in term of diff. I mean that there is no drop in the coverage, and the diff corresponds to what has been modified in this PR. The only strange result is about the list of changes. Several of them, untouched by the PR, are referenced in this list.

I suspect that the drop in coverage/diff was transient, due to missing reports from Windows or Linux tests that was not yet processed. After everything is aggregated, the coverage is normal. This specific issue is corrected by #6934

About the list of files, one can see that the real URL of https://codecov.io/gh/certbot/certbot/pull/6784 is https://codecov.io/gh/certbot/certbot/compare/491d6c8f4538c28f7c05d780db9a700982460242...3d811a784dba6f6a13fb773e35efc79a9aa0f867.

This URL gives use the two commits that are compared:
* `3d811a784dba6f6a13fb773e35efc79a9aa0f867` is the last commit of  `fix_centos6_ssl`, the merged branch by the PR (commit from 02/04/2019)
* `491d6c8f4538c28f7c05d780db9a700982460242` is the last ancestor on the master branch for the `fix_centos6_ssl` branch (commit from 27/03/2019)

Now, let's consider the last commit on master that was present when the codecov report has been triggered (so on 02/04/2019): it is `1daa3ca07634168945fc35ff44c413c6c94b795d` (the immediate commit before the one that integrated #6784.

And now so, let's compare `491d6c8f4538c28f7c05d780db9a700982460242` (the last ancestor for PR, so the based reference) and `1daa3ca07634168945fc35ff44c413c6c94b795d` (the latest commit in master at the time of the report generation):
https://github.com/certbot/certbot/compare/491d6c8f..1daa3ca0

The list of files given by this GitHub diff **is** the list of changes shown in the Codecov report changes https://codecov.io/gh/certbot/certbot/pull/6784/changes.

adferrand:
So, to sum up.

1) The base commit on which codecov is constructing a comparison to build the report is the last ancestor on the target branch for the PR.
[...]

PR #7258, opened 2019-07-18
Remove list of packaging efforts.

bmw:
I think this list maybe had value when distros were first starting to package Certbot, but now I don't think it does. What function does this list serve? The instruction generator at https://certbot.eff.org/instructions does a much better job telling users how to use these packages. On the packaging side, I think anyone capable of packaging Certbot at the various distros would be able to search their repositories to see if a Certbot package is available.

Since this list is hard to maintain as links semi-regularly break and keeping it up to date with all distros and all Certbot components is a fair bit of work, let's just remove it.

This PR was motivated by the Travis failures at https://travis-ci.com/certbot/website/builds/119588518 due to GNU Guix changing the layout of their site.


bmw:
We don't measure coverage on our docs.

I think what's going on here is one of the limitations with codecov. It's comparing coverage between:

1. The latest commit shared between this branch and master
2. This branch merged into master as is done in Travis

This second commit contains other changes causing coverage to drop. If you're really curious, you can read more about this problem at https://github.com/certbot/certbot/issues/6888.

Due to issues like this, we currently don't require codecov to pass to merge PRs, but being safe and making sure it passes (or at least we understand why it failed) isn't a bad idea.

I merged in master and tests pass now.

PR #7265, opened 2019-07-25
Fix test farm test cleanup logic

bmw:
This problem was introduced in https://github.com/certbot/certbot/pull/7070.

As described in the code comment, the idea here was to always cleanup after ourselves if we failed to start all instances by checking that the two lists are **NOT** equal in length. This code currently does the opposite.

I started a run of test farm tests with this change at https://travis-ci.com/certbot/certbot/builds/120535621.

### DIFF
--- a/tests/letstest/multitester.py
+++ b/tests/letstest/multitester.py
@@ -375,7 +375,7 @@ def cleanup(cl_args, instances, targetlist):
     # If lengths of instances and targetlist aren't equal, instances failed to
     # start before running tests so leaving instances running for debugging
     # isn't very useful. Let's cleanup after ourselves instead.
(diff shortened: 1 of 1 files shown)


bmw:
Yeah. It was hitting the second condition of this if statement of `not cl_args.saveinstances`. `cl_args.saveinstances` defaults to False according to https://github.com/certbot/certbot/blob/06a0dae67f7497075e6da1b0e51d4fcab2c679c6/tests/letstest/multitester.py#L75-L76 so `not cl_args.saveinstances` defaults to True.

I hit this problem trying to use `--saveinstances` to debug problems on RHEL 8.

bmw:
Sorry I wrote the bug in the first PR!

PR #7266, opened 2019-07-25
Prevent authorization re-use for --dry-run/--staging/--deactivate-authorizations (#5116)

alexzorin:
First go at implementing authz deactivation in Certbot, now that #7254 is merged.

For most users, they will automatically be opted in when they use `--dry-run` or `--staging`.

Although I had not planned on it initially, it seemed necessary to also add `--[no-]deactivate-authorizations`. Firstly, for the ability to test it (though I guess an env variable could have also done the same thing without growing the collection of CLI flags). Secondly, there are now other production RFC8555 CAs, and that functionality should be available to them as well, since `--dry-run ` is Let's Encrypt specific.

As discussed in #5116, this implementation deactivates authorizations on both order creation and after order finalization. This *should* make it foolproof for users, assuming the deactivations themselves don't fail. 

However, one defect is that authorization errors raised by `_get_order_and_authorizations` prevent the deactivations from running. The design around that and `allow_subset_of_names` is a little complicated so I'm not sure how to address it. Practically speaking, it just means those authzs will get deactivated on the next order.

I have tested with Boulder and Let's Encrypt ACMEv1 and ACMEv2. Testing with Pebble did not appear possible since it does not seem to implement authz reuse. 

I do not know how to create a good integration test for this - I'm not sure whether the one I added in certbot-ci even does anything.

I am not sure what Certbot should do if authz deactivation raises any `acme.Error`. Right now, I am just swallowing them and warning about it. This way, if something goes wrong, it will hopefully not not get in the way of the user.

The other concern I have right now is the impact on the ACME server. For example, there are surely users with 100-SAN certificates, and a `--dry-run` on them is going to start blasting the server, quite probably. Depending how fast the ACME server is, we might even hit the 20 req/sec rate limit. It also necessarily makes Certbot slower - the user must wait for the 100 deactivations to finish before control is returned to them.

Edit: I'm not sure what the build failure means. Seems to me like `deactivate_authorization` should be in the mock object ... it is in Python 3. Is it because Python 2's mock does not support looking up methods in super classes? What's the fix?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ version = meta['version']
 # specified here to avoid masking the more specific request requirements in
 # acme. See https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
-    'acme>=0.29.0',
+    'acme[dev]',
     # We technically need ConfigArgParse 0.10.0 for Python 2.6 support, but
     # saying so here causes a runtime error against our temporary fork of 0.9.3
     # in which we added 2.6 support (see #2243), so we relax the requirement.
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ version = meta['version']
 # specified here to avoid masking the more specific request requirements in
 # acme. See https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
-    'acme[dev]',
+    'acme>=0.37.0.dev0',
     # We technically need ConfigArgParse 0.10.0 for Python 2.6 support, but
     # saying so here causes a runtime error against our temporary fork of 0.9.3
     # in which we added 2.6 support (see #2243), so we relax the requirement.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
+* CLI flag `--deactivate-authorizations` has been added. It deactivates existing valid authorizations before and after issuing new certificates, ensuring that the results of all challenges are fresh. It is automatically enabled by `--dry-run` and `--staging`, but can be disabled with `--no-deactivate-authorizations`. 
 * acme: Authz deactivation added to `acme` module.
 
 ### Changed
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -136,7 +136,8 @@ class ClientBase(object):  # pylint: disable=too-many-instance-attributes
         """
         body = messages.UpdateAuthorization(status='deactivated')
         response = self._post(authzr.uri, body)
-        return self._authzr_from_response(response)
+        return self._authzr_from_response(response,
+            authzr.body.identifier, authzr.uri)
 
     def _authzr_from_response(self, response, identifier=None, uri=None):
         authzr = messages.AuthorizationResource(
--- a/certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
+++ b/certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
@@ -575,3 +575,17 @@ def test_ocsp_status_live(context):
 
     assert output.count('INVALID') == 1, 'Expected {0} to be INVALID'.format(cert)
     assert output.count('REVOKED') == 1, 'Expected {0} to be REVOKED'.format(cert)
+
+def test_deactivate_authzs(context):
+    """Test deactivation of valid authzs after a successful order"""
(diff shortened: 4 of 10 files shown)


adferrand:
Hello @alexzorin, the error you see in the oldest tests is quite subtle.

_TL;DR_

You need to update the file `local-oldest-requirement.txt` at the root of the GIT repository, to replace `acme[dev]==0.29.0` by `-e acme[dev]`.

_Explanation_

So `self.mock_net` is in instance of `mock.MagicMock`. Normally this kind of mock transparently create child `mock.MagicMock` and attach it to any member of the parent mock you try to access. Basically it means that doing `test = self.mock_net.deactivate_authorization` would initialize `test` to a magic instance mock, that is now also referenced by `self.mock_net.deactivate_authorization`. Obviously you can then play with `side_effect` on the mock to control what it will do.

However here, `self.mock_net` is initialized using a spec `mock.MagicMock(spec=acme_client.Client)`. Here it means that the mock will create child mocks only for members name referenced in the spec class, here `acme_client.Client`.

Obviously you created the `deactivate_authorization` in `acme_client.Client` so everything should work, right? But oldest tests have an additional logic on top of the other tests: the dependencies of each project (`acme`, `certbot`, `certbot-apache` ...) is strictly controlled by the content of the relevant `local-oldest-requirement.txt`. This is done to reproduce the situation of old Linux systems, where all component of Certbot may not be available at their latest version.

In the case of oldest tests, pip will not install the version of `acme` module available locally in your branch, but the version specified in `local-oldest-requirement.txt` from PyPI. Currently, this is the version `0.29.0`, which has obviously not the class `acme_client.Client` with method `deactivate_authorization` available, and so the test fail.

You need to say to the test system to use the local version of `acme`, following the modifications I describe in the TL;DR section.

setup.py
```diff
@@ -34,7 +34,7 @@ def read_file(filename, encoding='utf8'):
 # specified here to avoid masking the more specific request requirements in
 # acme. See https://github.com/pypa/pip/issues/988 for more info.
 install_requires = [
-    'acme>=0.29.0',
+    'acme[dev]',
```
adferrand: In fact in `setup.py`, you still need to set a "real" dependency, with version, and out of dev scope. Here, to rely on the development version of `acme`, put `acme>=0.37.0.dev0`. It will be updated automatically during the release.

alexzorin:
@adferrand thanks a lot for your help!

 I need to rebase against master to solve a conflict (and trigger a build), but I'm not sure what the convention in this project is - the contributing guidelines say "please do not force push to the branch containing your pull request". Likewise, merging master into the PR seems like it would become unmergeable. What's the best approach here?

adferrand:
I do not like rebasing. At least as soon as a branch becomes "public", meaning anyone beside me potentially looking at it. It is also what means the request in contributing guide. Because rebasing replaces all existing commits by new ones, so it screw up any review history a reviewer would have on your PR.

Merge master into your branch. It does not make it unmergeable, quite the opposite in fact. Git is smart enough to merge your PR back into master.

On top of it, our PR merge strategy is to do a squashed commit merge. What it means is that in fact it is not the branch PR that is merged: github creates a unique commit containing all modification done by your PR in comparison with master, and push this commit on master. Finally, your branch and its history is for you and the reviewers. What is left for master is a unique commit associated to your identity.

bmw:
Well I had planned on reviewing this today but ran out of time. I'm unassigning myself for now just in case someone else on the Certbot team can get to it before me, but I'm planning on making sure this PR gets reviewed soon.

bmw:
It's been a busy few weeks... Here's one high level comment to get things started though.

What do you think about allowing people to use `--dry-run` with other servers rather than adding `--[no-]deactivate-authorizations` flags? The benefit of this is the CLI stays a little simpler and you don't have to worry as much about people using `--deactivate-authorizations` against Let's Encrypt's production server because doing so also means Certbot won't save their certificates. This would be done by fixing up the processing of arguments after `argparse` in the function [here](https://github.com/certbot/certbot/blob/4eaa06d58e9d148bf2cb89231889e2709e86b71b/certbot/cli.py#L591).

alexzorin:
I think allowing `--server` to be combined with `--dry-run` would be better, yes. Do you think that should go into a separate PR?

bmw:
I think a separate PR that lands before this one would be preferred, but I don't have that strong of a preference.

alexzorin:
Closing in favor of #7442

PR #7268, opened 2019-07-25
Migrate certbot-auto users on CentOS 6 to Python 3.6

adferrand:
Fixes #7007

Python 3.4 is [EOL](https://www.python.org/dev/peps/pep-0429/), and only Python 3.x version available for CentOS 6 through EPEL is this version, and so is used by `certbot-auto`, the only official way to install Certbot on this platform.

This unpleasant situation becomes a little more uncomfortable, considering that the newest `pip` version (19.2) [just dropped Python 3.4 support](https://github.com/pypa/pip/issues/6685) and will refuse to start on this Python version. We can expect a lot of dependencies to follow this path now.

One direct result of this situation is that a fix to support correctly the ARM platforms requires to upgrade `pip` to 19.2 for `certbot-auto`. So this is not possible right now.

Then, let's upgrade Certbot instances on CentOS 6 to a supported version of Python 3.

This PR proposes a new bootstrap approach for CentOS 6 platform, `BootstrapRpmPython3Legacy`, that will install Python 3.6 from [SCL](https://www.softwarecollections.org) (the latest one available for now on CentOS 6). In term of Python 3 specific bootstrap methods, I take the occasion here to completely separate the bootstrap of CentOS 6 as a legacy system, from the RPM-based newest systems (like Fedora 29+) that are simply dropping support for Python 2.x. This is in prevision of future migration for all systems on Python 3.x, that is a different problematic than supporting old systems.

Tested for now:
* installation of `certbot-auto` on a fresh CentOS 6 system
* upgrade of `certbot-auto` on a CentOS 6 system with a previous version of the script (so with Python 3.4 from EPEL)

Farm tests coming soon.

PR #7269, opened 2019-07-25
Support hashed wheels on ARM platforms

### LINKED ISSUE #6933: Raspberry Pi wheels hash mismatch
As raised in #6918 and discussed on [community](https://community.letsencrypt.org/t/certbot-auto-certificates-fails-while-installing-phyton-packages-with-these-packages-do-not-match-the-hashes/90363), users installing certbot on Raspberry Pi have hash mismatches because Raspbian is configured to use [piwheels.org](https://www.piwheels.org) as an additional index.

I'm the maintainer of piwheels.org - it provides wheels built for Raspberry Pi (so users don't have to build from source which can be time consuming on the Pi).

Would it be possible to solve this problem somehow? Some options to explore:

- Add hashes of piwheels files to requirements file in certbot
- File a bug with pip to suggest it uses the hash to find a match rather than fail when it picks up the wrong file

I'm not sure it's possible to specify "only use pypi", as using the `-i` flag sets the index url, but doesn't unset `--extra-index-url`.

I would rather the solution to this not be to delete `/etc/pip.conf` as suggested in the community thread.

Thanks!

P.S. piwheels.org uses certbot, obvs. :100:

adferrand:
Fixes #6933.

This PR upgrades the version of pip installed by pipstrap (and so by certbot-auto) to [19.2.1](https://pip.pypa.io/en/stable/news/#id1). This version includes https://github.com/pypa/pip/issues/5874, that allows to validate hashed versions of dependencies against an alternate packages repository for PyPi.

This approach is used in particular on ARM platforms, where https://www.piwheels.org/ is used. This packages repositories offer pre-build wheels for this platform, which greatly facilitate installation of Certbot, and is usually pre-installed on ARM distributions.

This PR allows ARM users to take advantage of these pre-built wheels when `certbot-auto` is installed without deviating from the standard configuration of their system.

**Warning: https://github.com/certbot/certbot/pull/7268 must be merged before this PR. Otherwise, users on CentOS 6 will fail to run `certbot-auto`, because this version of pip will refuse to run on Python 3.4.**


bmw:
Since this makes a not strictly necessary change to certbot-auto, I'm leaning towards closing this PR and closing the issue as won't fix since certbot-auto is on a deprecation path.

What do you think @adferrand?

adferrand:
Indeed, it makes sense.

PR #7270, opened 2019-07-26
Fix unit tests on Windows

### LINKED ISSUE #6850: Fix remaining broken tests on Windows

adferrand:
Fixes #6850

This PR makes the last corrections needed to run all unit tests on Windows:
* add a function to check if a hook is executable in a cross-platform compatible way
* handle correctly the PATH surgery for Windows during hook execution
* handle correctly an account compatibility over both ACMEv1 and ACMEv2
* remove (finally!) the `@broken_on_windows` decorator.

### DIFF
--- a/certbot/compat/misc.py
+++ b/certbot/compat/misc.py
@@ -30,6 +30,10 @@ else:
     MASK_FOR_PRIVATE_KEY_PERMISSIONS = 0
 
 
+# For Linux: define OS specific standard binaries directories
+STD_BINARIES_DIRS = ("/usr/sbin", "/usr/local/bin", "/usr/local/sbin") if POSIX_MODE else ()
+
+
 def raise_for_non_administrative_windows_rights():
     # type: () -> None
     """
--- a/certbot/compat/misc.py
+++ b/certbot/compat/misc.py
@@ -31,7 +31,7 @@ else:
 
 
 # For Linux: define OS specific standard binaries directories
-STD_BINARIES_DIRS = ("/usr/sbin", "/usr/local/bin", "/usr/local/sbin") if POSIX_MODE else ()
+STD_BINARIES_DIRS = ["/usr/sbin", "/usr/local/bin", "/usr/local/sbin"] if POSIX_MODE else []
 
 
 def raise_for_non_administrative_windows_rights():
--- a/certbot/compat/misc.py
+++ b/certbot/compat/misc.py
@@ -30,8 +30,8 @@ else:
     MASK_FOR_PRIVATE_KEY_PERMISSIONS = 0
 
 
-# For Linux: define OS specific standard binaries directories
-STD_BINARIES_DIRS = ["/usr/sbin", "/usr/local/bin", "/usr/local/sbin"] if POSIX_MODE else []
+# For Linux: define OS specific standard binary directories
+STANDARD_BINARY_DIRS = ["/usr/sbin", "/usr/local/bin", "/usr/local/sbin"] if POSIX_MODE else []
 
 
 def raise_for_non_administrative_windows_rights():
--- a/certbot/plugins/util_test.py
+++ b/certbot/plugins/util_test.py
@@ -30,12 +30,14 @@ class PathSurgeryTest(unittest.TestCase):
                 self.assertEqual(mock_debug.call_count, 0)
                 self.assertEqual(os.environ["PATH"], all_path["PATH"])
         no_path = {"PATH": "/tmp/"}
-        with mock.patch.dict('os.environ', no_path):
-            path_surgery("thingy")
-            self.assertEqual(mock_debug.call_count, 2)
-            self.assertTrue("Failed to find" in mock_debug.call_args[0][0])
-            self.assertTrue("/usr/local/bin" in os.environ["PATH"])
-            self.assertTrue("/tmp" in os.environ["PATH"])
+        if os.name != 'nt':
+            # This part is specific to Linux since on Windows no PATH surgery is ever done.
+            with mock.patch.dict('os.environ', no_path):
+                path_surgery("thingy")
+                self.assertEqual(mock_debug.call_count, 2 if os.name != 'nt' else 1)
+                self.assertTrue("Failed to find" in mock_debug.call_args[0][0])
+                self.assertTrue("/usr/local/bin" in os.environ["PATH"])
+                self.assertTrue("/tmp" in os.environ["PATH"])
 
 
 if __name__ == "__main__":
--- a/certbot/plugins/util_test.py
+++ b/certbot/plugins/util_test.py
@@ -16,6 +16,7 @@ class GetPrefixTest(unittest.TestCase):
         self.assertEqual(get_prefixes('/'), [os.path.normpath('/')])
         self.assertEqual(get_prefixes('a'), ['a'])
 
+
 class PathSurgeryTest(unittest.TestCase):
     """Tests for certbot.plugins.path_surgery."""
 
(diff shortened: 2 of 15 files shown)


certbot/tests/hook_test.py
```diff
@@ -30,19 +30,37 @@ def test_it(self, mock_validate_hook):
...
+
+        # On Windows a file created within Certbot will always have all permissions to the
+        # Administrators group set. Since the unit tests are typically executed under elevated
+        # privileges, it means that current user will always have effective execute rights on the
+        # hook script, and so the test will fail. To prevent that and represent a file created
+        # outside Certbot as typically a hook file is, we mock the _generate_dacl function in
+        # certbot.compat.filesystem to give rights only to the current user. This implies removing
+        # all ACEs except the first one from the DACL created by original _generate_dacl function.
+
+        from certbot.compat.filesystem import _generate_dacl
```
bmw: Since this test relies on mocking the internals of `filesystem.is_executable`, I propose moving the majority of the code in this test to a new class in `certbot/tests/compat/filesystem_test.py` and updating the code here to just mock out `filesystem.is_executable` entirely with a comment about it being hard to test without a mock on Windows.

What do you think?

certbot/tests/util_test.py
```diff
@@ -52,26 +53,42 @@ def _call(cls, exe):
         from certbot.util import exe_exists
         return exe_exists(exe)
 
-    @mock.patch("certbot.util.os.path.isfile")
```
bmw: I have the same high level comments about testing here as I do above. I personally think these tests should be moved to `filesystem_test.py` and we can keep simple tests here that just mock out `filesystem.is_executable` if we want.

certbot/compat/misc.py
```diff
@@ -30,6 +30,10 @@
     MASK_FOR_PRIVATE_KEY_PERMISSIONS = 0
 
 
+# For Linux: define OS specific standard binaries directories
+STD_BINARIES_DIRS = ["/usr/sbin", "/usr/local/bin", "/usr/local/sbin"] if POSIX_MODE else []
```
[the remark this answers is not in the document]

certbot/plugins/util_test.py
```diff
@@ -30,12 +30,14 @@ def test_path_surgery(self, mock_debug):
                 self.assertEqual(mock_debug.call_count, 0)
                 self.assertEqual(os.environ["PATH"], all_path["PATH"])
         no_path = {"PATH": "/tmp/"}
```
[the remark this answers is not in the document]
adferrand: Done.

certbot/tests/hook_test.py
adferrand: I agree. Done.

certbot/tests/util_test.py
adferrand: Done.

certbot/tests/util_test.py
```diff
@@ -52,26 +53,26 @@ def _call(cls, exe):
...
+        with mock.patch("certbot.util.filesystem.is_executable", return_value=True):
+            self.assertTrue(self._call("/path/to/exe"))
+
+    def test_exe_not_exists(self):
+        with mock.patch("certbot.util.filesystem.is_executable", return_value=False):
+            self.assertFalse(self._call("/path/to/exe"))
+
+
+@contextlib.contextmanager
+def _fix_windows_runtime():
```
bmw: This should be deleted.

adferrand: Done.

adferrand:
> Other than two more tiny nits, when manually testing this, I was surprised by the behavior I saw. In a command prompt not running as administrator, I checked out this code, activated the venv, and ran:
> 
> ```
> import os
> import tempfile
> 
> import win32security
> 
> from certbot.compat import filesystem
> from certbot.util import safe_open
> 
> with tempfile.TemporaryDirectory() as dir_path:
>     file_name = 'foo'
>     file_path = os.path.join(dir_path, file_name)
>     safe_open(file_path, chmod=0o666).close()
>     print(win32security.LookupAccountSid(None, filesystem._get_current_user()))
>     assert not filesystem.is_executable(file_path)
> ```
> 
> The output I got was:
> 
> ```
> ('IEUser', 'MSEDGEWIN10', 1)
>   File "test.py", line 15, in <module>
>     assert not filesystem.is_executable(file_path)
> AssertionError
> ```
> 
> Can you explain what's going on here?

This is because you are member of the Administrators group, and this group has always all the permissions. So effectively, the file is executable for your user.

To give more details here, `is_executable` does not inspect the DACL itself, but call `dacl.GetEffectiveRightsFromAcl` instead. This is an API call that gives you the effective permissions for a user on the current DACL, regarding all ACEs applied in it.

I use this high-level approach since `filesystem.is_executable` is primarily designed to check hooks validity, and these hooks, created outside certbot, are most likely not following our security model.

bmw:
That makes sense for the behavior from `GetEffectiveRightsFromAcl`, but to change my example slightly, it doesn't seem like Windows is using that same behavior.

Is there a way we can prevent the behavior shown here?
```
import os
import tempfile

import win32security


from certbot.compat import filesystem
from certbot.hooks import execute
from certbot.util import safe_open

with tempfile.TemporaryDirectory() as dir_path:
    file_name = 'foo.py'
    file_path = os.path.join(dir_path, file_name)
    with safe_open(file_path, chmod=0o666) as f:
        f.write('print("Hello world!")')
    print(win32security.LookupAccountSid(None, filesystem._get_current_user()))
    if filesystem.is_executable(file_path):
        print(execute('test-hook', file_path))
```
Output was:
```
('IEUser', 'MSEDGEWIN10', 1)
test-hook command "C:\Users\IEUser\AppData\Local\Temp\tmpehg661r9\foo.py" returned error code 1
Error output from test-hook command foo.py:
Access is denied.

('Access is denied.\n', '')
```
If you call `safe_open` with 0o777, it runs fine.

adferrand:
Your are encountering this error because you run this code from a non privileged shell, and the effective rights here are accessible only when you are in a privileged shell. However, Windows documentation about GetEffectiveRightsFromAcl says that the privileges hold by the user at call time are not taken into account.

However, the situation you described will not happen: certbot refuses to run on a non-privileged shell. So you will always get the maximum permissions you can get from the file, and `is_executable` result will be consistent with the effective execution of the hook script.

bmw:
> However, Windows documentation about GetEffectiveRightsFromAcl says that the privileges hold by the user at call time are not taken into account to get the result, so you do not see it.

Windows why? Are you familiar with an easy way to take the current privileges into account? I think that'd be ideal, but if not, I think we should add a simple comment in the code describing this behavior.

Also, I just noticed our avoidance of `os.access`. I assume that's yet another function we can't use on Windows? Grepping through the code, we don't seem to use it anywhere else, but I personally think it's worth preventing it by adding a simple function in `compat.os` that errors out saying it doesn't work on Windows.

adferrand:
For your first remark, no, I think there is no easy way do to it that, and found no clue how it could be properly handled on internet. In fact, I think it is an approach from Posix that is not applicable on Windows. On Windows, you can mark an executable or a script as requiring privileged rights. It appears as so in the interface, and if you try to run it, it will trigger automatically the privileges escalation, and fail if your account cannot acquire them.

So I suppose that you design and write your program with the assumption it will be run under elevated privileges, and mark the executable as so. This is basically what we did for certbot internally, this is also what I will do for the executable created by the installers if possible.

For `os.access`, yes the official documentation https://docs.python.org/3/library/os.html#os.access talk only about Unix things, and warn about inaccuracy of the outcome for os that goes beyond the POSIX approach. So I expect it to be useless for Windows.

Yes, I think it can be forbidden. I can give a reference to `is_executable` in the exception about execution rights specifically.

adferrand:
OK @bmw, I forbid `os.access` and added a comment on `is_executable`.

bmw:
Other than the test failures I just noticed which are stopping me from merging the PR 😛

bmw:
Looks like lint is still angry

adferrand:
Pylint is happy again

Issue #7152, opened 2019-06-12
Don't send OCSP requests for expired certificates


bmw:
Let's Encrypt doesn't update OCSP for expired certificates. They are not required to do so. When someone runs `certbot certificates` with expired certificates and they're using our Python support for making OCSP requests rather than shelling out to `openssl`, they get output like:
```
Invalid OCSP response for /etc/letsencrypt/live/example.org/cert.pem: param nextUpdate is in the past…
```
There's not actually a problem here so we shouldn't be printing these warnings to users. 

The code that does this is at https://github.com/certbot/certbot/blob/d75908c645e0d13827b60c4c94491e0c41859ad4/certbot/cert_manager.py#L264-L267

Fixing that is simple enough, however, I think [one of our tests is relying on this behavior](https://github.com/certbot/certbot/blob/9a7f774706484dcec10cf5247d31bb547a5e1160/certbot-ci/certbot_integration_tests/certbot_tests/test_main.py#L545) (and if so, the test should be deleted) and we may also want to add code in `certbot.ocsp.ocsp_revoked` to check if the certificate is still valid before querying OCSP to help prevent this problem in the future.

This was reported at https://community.letsencrypt.org/t/invalid-ocsp-response-nextupdate-is-in-the-past/95640.

ohemorange:
> Fixing that is simple enough, however, I think one of our tests is relying on this behavior (and if so, the test should be deleted)

After changing the relevant `if` to an `elif`, `tox -e integration-certbot` and `tox -e integration` both still passed. Am I not running the right thing to hit that test?

bmw:
No I think you did the right thing. Looks like the test isn't relying on the behavior.

PR #7281, opened 2019-07-29
Create a mock OCSP server for Pebble integration tests

adferrand:
Our PR builds and nightly builds execute integration tests against [Pebble](https://github.com/letsencrypt/pebble), the lightweight ACME server dedicated for testing.

On Windows specifically, this is the only solution available to run integration tests.

Pebble tries to be a simple and maintainable as possible, and so focus only on the sole purpose of testing the ACME protocol. In particular, it does not integrate an OCSP server, on the contrary to Boulder: consequently, the integration test `test_ocsp_status_live` was not run against Pebble.

There is another OCSP test, `test_ocsp_status_stale`, but it is based on an out-of-date Certbot configuration sample, that is somewhat broken (see https://github.com/certbot/certbot/issues/7152). And so Windows does not have any relevant OCSP status test.

This PR handles this issue, by giving a capable OCSP server, to run alongside Pebble and execute the `test_ocsp_status_live` against it. This is done by crafting an HTTP server, using `cryptography` module, to communicate with the management interface of Pebble and respond to OCSP statuses in its name.

Several assets were needed on Pebble side to make this possible: a management interface, an OCSP status entrypoint, entrypoints for the CA key and certificate, and the crafting of OCSP responder URL in certificates. These assets have been developed the last weeks by @felixfontein and myself, and I would like to thanks a lot @felixfontein for the big work he did here.

Since `test_ocsp_status_stale` is not relevant anymore, I also remove it in this PR.


certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -542,24 +542,11 @@ def test_wildcard_certificates(context):
     assert exists(join(context.config_dir, 'live', certname, 'fullchain.pem'))
 
 
-def test_ocsp_status_stale(context):
```
bmw: I don't think we should entirely delete this test. I think it's useful for testing for "EXPIRED" and to a lesser extent "TEST_CERT" in `certbot certificates` output.

Neither of these things have to do with OCSP though so the function could use a better name and maybe a better approach for testing for this behavior. Assuming this test still works, I'd recommend leaving it as is for now.

certbot-ci/setup.py
```diff
@@ -11,6 +11,7 @@
     'pytest',
     'pytest-cov',
     'pytest-xdist',
+    'python-dateutil',
```
bmw: We should pin this in `tools/dev_constraints.txt`.

certbot-ci/certbot_integration_tests/utils/pebble_ocsp_server.py
```diff
@@ -0,0 +1,71 @@
...
+import datetime
+import re
+
+import requests
+from dateutil import parser
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography import x509
+from cryptography.x509 import ocsp
```
bmw: Won't this fail in our nightly tests where the version of `cryptography` installed doesn't have OCSP support?

certbot-ci/setup.py
adferrand: It is already, on master: https://github.com/certbot/certbot/blob/56f609d4f582f61274f93a807ec4b8a0bba43828/tools/dev_constraints.txt#L62

certbot-ci/certbot_integration_tests/utils/pebble_ocsp_server.py
adferrand: It would, but the situation were it would occur is the oldest integration tests `integration-certbot-oldest` and `integration-nginx-oldest`, and these tests are using Boulder.

However I moved the import of pebble_ocsp_server inside the dedicated method that setup Pebble environment, to not raise a useless `ImportError` on oldest tests when Boulder is used.

certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
adferrand: Well, I definitetly think that the current setup of this test is now overkill regarding its purpose. However I agree this should be done in a separate PR. We would certainly be able to make the same tests, but without embedding a full certbot config in the asset tests.

For now I revert the removal.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -134,6 +134,9 @@ def _prepare_pebble_server(self):
             [challtestsrv_path, '-management', ':{0}'.format(CHALLTESTSRV_PORT), '-defaultIPv6', '""',
              '-defaultIPv4', '127.0.0.1', '-http01', '""', '-tlsalpn01', '""', '-https01', '""'])
 
+        from certbot_integration_tests.utils import pebble_ocsp_server
```
bmw: I think we should add a brief comment explaining why this module is imported here.

certbot-ci/setup.py
bmw: Indeed it is! Sorry for the noise.

certbot-ci/certbot_integration_tests/utils/acme_server.py
adferrand: Done.

Issue #7241, opened 2019-07-12
certbot-auto doesn't work on RHEL 8


bmw:
This was reported by @haidarvm at https://github.com/certbot/certbot/issues/7202#issuecomment-510455526.

## My operating system is (include version):

RHEL 8

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

certbot-auto

## I ran this command and it produced this output:
```
sudo /usr/local/bin/certbot-auto  --debug --nginx --verbose
Bootstrapping dependencies for RedHat-based OSes... (you can skip this with --no-bootstrap)
dnf is /bin/dnf
dnf is hashed (/bin/dnf)
Updating Subscription Management repositories.
Last metadata expiration check: 2:13:18 ago on Jum 12 Jul 2019 12:03:04  +08.
Package gcc-8.2.1-3.5.el8.x86_64 is already installed.
Package augeas-libs-1.10.1-8.el8.x86_64 is already installed.
Package openssl-1:1.1.1-8.el8.x86_64 is already installed.
Package openssl-devel-1:1.1.1-8.el8.x86_64 is already installed.
Package libffi-devel-3.1-18.el8.x86_64 is already installed.
Package redhat-rpm-config-116-1.el8.noarch is already installed.
Package ca-certificates-2018.2.24-6.el8.noarch is already installed.
Package python2-libs-2.7.15-22.module+el8.0.0+2961+596d0223.x86_64 is already installed.
Package python2-setuptools-39.0.1-11.module+el8.0.0+2961+596d0223.noarch is already installed.
Package python2-devel-2.7.15-22.module+el8.0.0+2961+596d0223.x86_64 is already installed.
Package python2-virtualenv-15.1.0-18.module+el8.0.0+2961+596d0223.noarch is already installed.
Package python2-tools-2.7.15-22.module+el8.0.0+2961+596d0223.x86_64 is already installed.
Package python2-pip-9.0.3-13.module+el8.0.0+2961+596d0223.noarch is already installed.
Dependencies resolved.
Nothing to do.
Complete!
Upgrading certbot-auto 0.35.1 to 0.36.0...
Replacing certbot-auto...
Creating virtual environment...
Traceback (most recent call last):
  File "<stdin>", line 27, in <module>
  File "<stdin>", line 19, in create_venv
  File "/usr/lib64/python2.7/subprocess.py", line 185, in check_call
    retcode = call(*popenargs, **kwargs)
[...]

Issue #7273, opened 2019-07-26
certbot-apache fails on a fresh install of httpd on RHEL 8


bmw:
## My operating system is (include version):

RHEL 8

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):

certbot-auto

###### Logs are stored in `/var/log/letsencrypt` by default. Feel free to redact domains, e-mail and IP addresses as you see fit.
```
2019-07-26 19:05:42,605:DEBUG:certbot.main:certbot version: 0.37.0.dev0
2019-07-26 19:05:42,607:DEBUG:certbot.main:Arguments: ['-v', '--debug', '--text', '--agree-dev-preview', '--agree-tos', '--renew-by-default', '--redirect', '--register-unsafely-without-email', '--domain', 'ec2-3-83-206-109.compute-1.amazonaws.com', '--server', 'http://10.0.0.232:4000/directory']
2019-07-26 19:05:42,608:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#apache,PluginEntryPoint#manual,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
2019-07-26 19:05:42,617:WARNING:certbot.util:Use of --agree-dev-preview is deprecated.
2019-07-26 19:05:42,634:DEBUG:certbot.log:Root logging level set at 10
2019-07-26 19:05:42,634:INFO:certbot.log:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2019-07-26 19:05:42,636:DEBUG:certbot.plugins.selection:Requested authenticator None and installer None
2019-07-26 19:05:42,791:ERROR:certbot.util:Error while running apachectl configtest.

AH00526: Syntax error on line 85 of /etc/httpd/conf.d/ssl.conf:
SSLCertificateFile: file '/etc/pki/tls/certs/localhost.crt' does not exist or is empty

2019-07-26 19:05:42,792:DEBUG:certbot.plugins.disco:Misconfigured PluginEntryPoint#apache: Error while running apachectl configtest.

AH00526: Syntax error on line 85 of /etc/httpd/conf.d/ssl.conf:
SSLCertificateFile: file '/etc/pki/tls/certs/localhost.crt' does not exist or is empty
Traceback (most recent call last):
  File "/home/ec2-user/letsencrypt/certbot-apache/certbot_apache/configurator.py", line 2239, in config_test
    util.run_script(self.option("conftest_cmd"))
[...]

PR #7282, opened 2019-07-29
Fix certbot-auto on RHEL 8

bmw:
This PR builds off of #7240 to fix https://github.com/certbot/certbot/issues/7241.

The code in certbot-auto is unchanged which I +1. Someone else should give it a 2nd review.

For the code in the tests, you can see all tests passing (including `test_tests.sh`) at https://travis-ci.com/certbot/certbot/jobs/220681085.

One noteworthy thing here is I did not add the RHEL 8 AMI to the Apache tests due to https://github.com/certbot/certbot/issues/7273. This problem is not related to support in certbot-auto though, is an edge case, and I do not personally believe it should block this PR.

@adferrand, are you interested in giving this PR a review?

### DIFF
--- a/letsencrypt-auto-source/letsencrypt-auto.template
+++ b/letsencrypt-auto-source/letsencrypt-auto.template
@@ -333,10 +333,18 @@ elif [ -f /etc/redhat-release ]; then
   # Starting to Fedora 29, python2 is on a deprecation path. Let's move to python3 then.
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
   RPM_DIST_VERSION=0
-  if [ "$RPM_DIST_NAME" = "fedora" ]; then
-    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) || echo "0"`
+  if [ "$RPM_DIST_NAME" = "fedora" ] || [ "$RPM_DIST_NAME" = "rhel" ]; then
+    RPM_DIST_VERSION=`(. /etc/os-release 2> /dev/null && echo $VERSION_ID) | cut -d '.' -f1 || echo "0"`
   fi
   if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$PYVER" -eq 26 ]; then
+    Bootstrap() {
+      BootstrapMessage "RedHat-based OSes that will use Python3"
+      BootstrapRpmPython3
+    }
+    USE_PYTHON_3=1
+    BOOTSTRAP_VERSION="BootstrapRpmPython3 $BOOTSTRAP_RPM_PYTHON3_VERSION"
+  # RHEL 8 uses python3.
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
     Bootstrap() {
       BootstrapMessage "RedHat-based OSes that will use Python3"
       BootstrapRpmPython3
--- a/tests/letstest/scripts/test_leauto_upgrades.sh
+++ b/tests/letstest/scripts/test_leauto_upgrades.sh
@@ -26,6 +26,17 @@ else
     # 0.33.x is the oldest version of letsencrypt-auto that works on Fedora 29+.
     INITIAL_VERSION="0.33.1"
 fi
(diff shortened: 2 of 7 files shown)


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -343,6 +343,14 @@ elif [ -f /etc/redhat-release ]; then
     }
     USE_PYTHON_3=1
     BOOTSTRAP_VERSION="BootstrapRpmPython3 $BOOTSTRAP_RPM_PYTHON3_VERSION"
+  # RHEL 8 uses python3.
+  elif [ "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 ]; then
```
adferrand: Is it possible to merge this condition with the previous one? It will reduce code duplication since the exact same logic is applied in case of old centos, rhel8 and fedora 29+.

Note that in my PR about migrating centos 6 users to python 3.6, I will extract this situation in a specific branch with a specific bootstrap.

tests/letstest/scripts/test_leauto_upgrades.sh
```diff
@@ -26,6 +26,17 @@ else
     # 0.33.x is the oldest version of letsencrypt-auto that works on Fedora 29+.
     INITIAL_VERSION="0.33.1"
 fi
+
+# If we're on RHEL 8, the initial version of certbot-auto will fail until we do
```
adferrand: We should create issue to mark the fact we need to remove this code once the next version of Certbot is released.

tests/letstest/scripts/test_sdists.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh -xe
 
 cd letsencrypt
-./certbot-auto --install-only -n --debug
+letsencrypt-auto-source/letsencrypt-auto --install-only -n --debug
```
adferrand: Why using the local newest letsencrypt-auto here instead of the released certbot-auto script? Does it mean the previous logic was wrong?

letsencrypt-auto-source/letsencrypt-auto.template
bmw: I personally don't think we should make a massive line like:
```
if [ "$RPM_DIST_NAME" = "fedora" -a "$RPM_DIST_VERSION" -ge 29 -o "$RPM_DIST_NAME" = "rhel" -a "$RPM_DIST_VERSION" -ge 8 -o "$PYVER" -eq 26 ]; then
```
(plus any parens to make that work)

We could have separate if branches that set some variable like `RPM_USE_PYTHON_3` and then a separate if branch that checks this variable and then sets the relevant bootstrap variables.

If you think we should do this, I'd encourage you to suggest it on https://github.com/certbot/certbot/pull/7240 otherwise we need to get a 3rd person in here to review any changes we make.

If you don't like my suggestion to reuse code, I think we should get on the same page ourselves before making requests in #7240.

tests/letstest/scripts/test_sdists.sh
bmw: We need to use letsencrypt-auto in letsencrypt-auto-source here because otherwise we fail to bootstrap and create a venv so we know whether to use Python 2 or Python 3 in this test script. We do the same kind of thing in `test_apache2.sh` and `test_test.sh`.

I wouldn't say the previous logic was necessarily wrong, just that it didn't allow the test to run on systems where bootstrapping with the latest released certbot-auto didn't work.

adferrand: OK it makes sense. And the new approach is more consistent, since the current letsencrypt-auto script becomes a part of the test, while we now that the released one is already working.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Yes my concern is to avoid duplication of the code, to improve readability and maintainability, but I agree that a big chunk of if condition does not reach this purpose.

I am OK with your proposition, it will make the code more explicit, while avoiding duplication of the code. 

I create a request on #7240.

tests/letstest/scripts/test_leauto_upgrades.sh
bmw: Created https://github.com/certbot/certbot/issues/7301.

PR #7306, opened 2019-08-07
[Apache v2] Change _get_vhost_names to take virtualhost object instead path

joohoi:
First refactoring PR with minimal changes.


joohoi:
Closing, this change isn't needed. The changes to the method in order to support ParserNode is better handled elsewhere.

PR #7307, opened 2019-08-07
certbot-dns-dreamhost. adopted from certbot-dns-linode.

chhsiao1981:
DNS Authenticator for Dreamhost


chhsiao1981:
https://github.com/AnalogJ/lexicon/blob/master/examples/certbot.default.sh

Found that the manual hook with dns-lexicon may be an easier option to integrate with.
Closing this PR for now.

PR #7321, opened 2019-08-09
Fix apache-parser-v2 tests

bmw:
Integration tests are failing on the apache-parser-v2 branch due to missing coverage on Certbot. See https://travis-ci.com/certbot/certbot/builds/122633529. I don't immediately understand the sequence of commits and Travis runs that caused/allowed the branch to get into this state, but regardless, this PR should fix it.

**This PR should be merged and not squashed to help us keep our sanity when merging the apache-parser-v2 branch back into master**


joohoi:
I suggested rebasing the `apache-parser-v2` branch on the current `master` in #7246 as there's no changes landed yet, and it would probably be the easiest way to handle everything here. What do you think?

PR #7326, opened 2019-08-14
[Windows] Fix closing files descriptors during unit tests

adferrand:
Some scars were still there from the dark age of unit tests failures on Windows. They are ignoring exceptions that occurs during the tearDown phase of a test when some file handles are not properly closed.

This PR fixes that, ensuring that all file handles are really closed at the end of each test, and so these workarounds are not needed anymore, so removed.

### DIFF
--- a/certbot-nginx/certbot_nginx/tests/util.py
+++ b/certbot-nginx/certbot_nginx/tests/util.py
@@ -1,9 +1,9 @@
 """Common utilities for certbot_nginx."""
 import copy
+import logging
 import shutil
 import tempfile
 import unittest
-import warnings
 
 import josepy as jose
 import mock
@@ -11,6 +11,7 @@ import pkg_resources
 import zope.component
 
 from certbot import configuration
+from certbot import util
 from certbot.compat import os
 from certbot.plugins import common
 from certbot.tests import util as test_util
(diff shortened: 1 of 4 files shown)


certbot-nginx/certbot_nginx/tests/util.py
```diff
@@ -34,20 +35,20 @@ def setUp(self):
...
-        shutil.rmtree(self.work_dir, onerror=onerror_handler)
-        shutil.rmtree(self.logs_dir, onerror=onerror_handler)
+        # Cleanup opened resources after a test. This is usually done through atexit handlers in
+        # Certbot, but during tests, atexit will not run registered functions before tearDown is
+        # called and instead will run them right before the entire test process exits.
+        # It is a problem on Windows, that does not accept to clean resources before closing them.
+        logging.shutdown()
+        # Remove logging handlers that have been closed so they won't be
+        # accidentally used in future tests.
+        logging.getLogger().handlers = []
```
bmw: I get the addition of the call to `util._release_lock()` but needing to clean up loggers here surprises me. Do we definitely need to do this or did you just copy it from the `tearDown` code in Certbot's tests? If the former, do you understand why?

adferrand: Oh, I just copy the code. It is not necessary.

adferrand: Fixed that. By the way, why this class does not extend TempDirTestCase from `certbot.tests.util`?

bmw: I didn't look through the git history to verify this but I think it's that the code here predates `TempDirTestCase` and maybe there was a bit of concern about what code in Certbot the plugins should be using in a clean design.

PR #7327, opened 2019-08-14
Use travis_retry for farm tests

adferrand:
The farm tests are flaky, and fail at least once each two days. Personally it starts to annoy me, and by that I mean that their red flag effect on failures vanishes, since they just fail for any reason.

This is not good, and in the past some real failures have been ignored on Apache tests that were also failing all the time.

In order to improve the situation, until we setup a new approach to improve their stability, I propose to use the `travis_retry` bash method exposed by Travis, in order to rerun (at most 2 times) the failed farm tests jobs during nightly builds.

This PR does that. The retry is triggered conditionally on farm tests only, since the other jobs are failing or passing consistently.

You can see an execution of these tests here: https://travis-ci.com/certbot/certbot/builds/123250252

All tests are failing in this build, because I think I exhausted the available AMI instances (problem of clean up here ?), but you will notice in the logs that each farm test has been run 3 times before the relevant Travis job fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -276,7 +276,7 @@ addons:
 # CERTBOT_NO_PIN is set to work around failures we've seen when using an older
 # version of virtualenv.
 install: 'tools/pip_install.py -U codecov tox virtualenv'
-script: '"$TRAVIS_RETRY" tox'
+script: '$TRAVIS_RETRY tox'
 
 after_success: '[ "$TOXENV" == "py27-cover" ] && codecov -F linux'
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -273,8 +275,8 @@ addons:
 # virtualenv is listed here explicitly to make sure it is upgraded when
 # CERTBOT_NO_PIN is set to work around failures we've seen when using an older
 # version of virtualenv.
-install: "tools/pip_install.py -U codecov tox virtualenv"
-script: tox
+install: 'tools/pip_install.py -U codecov tox virtualenv'
+script: '$TRAVIS_RETRY tox'
```
bmw: nit: Can you add a brief comment here explaining that TRAVIS_RETRY is set for flakey tests and for most tests it is an empty string and has no effect?

PR #7329, opened 2019-08-15
added nslookup command info to show if txt record is deployed

monotek:
Signed-off-by: André Bauer <monotek23@gmail.com>

Adds a note to the dns validation output, how to use dig command to see if the txt record is already available.

## Pull Request Checklist

- [x] Edit the `master` section of `CHANGELOG.md` to include a description of
  the change being made.
- [x] Add [mypy type


bmw:
@schoen would you be interested in reviewing this PR which slightly expanding our manual output to tell people how to check if their TXT record has been deployed?

schoen:
Yes, I'll take a look!

certbot/plugins/manual.py
```diff
@@ -43,7 +43,8 @@ class Authenticator(common.Plugin):
 
 {validation}
 
-Before continuing, verify the record is deployed."""
+Before continuing, verify the record is deployed.
+This can be done via cli: watch -n1 dig -t txt {domain}"""
```
schoen: I'm not sure that all of our users know what "via CLI" means, especially if it's not capitalized to appear like an acronym. It seems that we have quite a few users who aren't used to using CLIs for many tasks, and in some cases who are just beginning to learn them specifically in order to run Certbot!

Maybe we could spell it out a little more explicitly, like

`One way to do this is running the following command at a terminal prompt:`

or

`This can be monitored by running the following command:`

or something similar?

The choice of command itself seems very good to me.

schoen: @adferrand, should this be wrapped with some kind of logic to prevent it from being displayed on Windows, since I assume the default Windows CLI environment doesn't have `watch` and `dig`? Is there a straightforward alternative that would work on most Windows systems to check whether the TXT record exists?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ## 0.38.0 - master
 
+* added dig command info to show if txt record is deployed
```
schoen: I think that `TXT` should be capitalized in documentation; compare

https://en.wikipedia.org/wiki/TXT_record

(I don't think this is necessary in the command, though.)

monotek:
Maybe nslookup would work on Linux and windows. I'll check it out today.

certbot/plugins/manual.py
adferrand: Yes, `nslookup`, that is installed on Windows, and in most Linux environments. Following command will display TXT entries in both system: `nslookup -q=TXT example.com`.

adferrand:
> Maybe nslookup would work on Linux and windows. I'll check it out today.

Yes it will work, see my response here: https://github.com/certbot/certbot/pull/7329#discussion_r314620163

monotek:
I've updated the command to nslookup and changed the changelog to use capital TXT.

certbot/plugins/manual.py
```diff
@@ -43,7 +43,9 @@ class Authenticator(common.Plugin):
 
 {validation}
 
-Before continuing, verify the record is deployed."""
+Before continuing, verify the DNS record is deployed.
+One way to do this is running the following command at a terminal prompt: nslookup -type=TXT _acme-challenge.{domain}
```
schoen: Doesn't `{domain}` already include the `_acme-challenge` prefix?

monotek: true. removed it.

schoen:
I double-checked that `{domain}` in this context already includes the `_acme-challenge` prefix, and so it is indeed correct not to add that prefix.

monotek:
Ok. I'll do it tomorrow.

monotek:
done :-)

monotek:
update pr to current master...

monotek:
@schoen 
Is something missign to get this merged?

PR #7330, opened 2019-08-15
[Apache v2] Modifications to ParserNode interfaces

joohoi:
This PR contains the changes requested in initial pre-review comments of #7308 

- Move properties to class pydocs in `interfaces.py`
- Prefer class ABC register() functionality instead of class inheritance for interface classes
- Add apache implementation specific functions to interfaces

### DIFF
--- a/certbot-apache/certbot_apache/interfaces.py
+++ b/certbot-apache/certbot_apache/interfaces.py
@@ -124,7 +124,7 @@ class ParserNode(object):
 
     # Filepath of the file where the configuration element for this ParserNode
     # object resides.
-    filepath: str
+    filepath: Optional[str]
     """
 
     @abc.abstractmethod
--- a/certbot-apache/certbot_apache/interfaces.py
+++ b/certbot-apache/certbot_apache/interfaces.py
@@ -121,7 +121,20 @@ class ParserNode(object):
 
     # True if this node has been modified since last save.
     dirty: bool
+
+    # Filepath of the file where the configuration element for this ParserNode
+    # object resides.
+    filepath: str
     """
+
+    @abc.abstractmethod
+    def metadata(self, key):
+        """
+        Gets an element from the metadata dictionary for this ParserNode object.
+
+        :param str key: Element key name
+        :returns: Requested metadata element
+        """
 
     @abc.abstractmethod
     def save(self, msg):
@@ -142,7 +155,7 @@ class ParserNode(object):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class CommentNode(ParserNode):
+class CommentNode(object):
     """
     CommentNode class is used for representation of comments within the parsed
     configuration structure. Because of the nature of comments, it is not able
@@ -161,7 +174,7 @@ class CommentNode(ParserNode):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class DirectiveNode(ParserNode):
+class DirectiveNode(object):
     """
     DirectiveNode class represents a configuration directive within the configuration.
     It can have zero or more parameters attached to it. Because of the nature of
@@ -195,7 +208,7 @@ class DirectiveNode(ParserNode):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class BlockNode(ParserNode):
+class BlockNode(object):
     """
     BlockNode class represents a block of nested configuration directives, comments
     and other blocks as its children. A BlockNode can have zero or more parameters
--- a/certbot-apache/certbot_apache/interfaces.py
+++ b/certbot-apache/certbot_apache/interfaces.py
@@ -123,9 +123,18 @@ class ParserNode(object):
     dirty: bool
 
     # Filepath of the file where the configuration element for this ParserNode
-    # object resides.
+    # object resides. This can be None if a configuration directive is defined in
+    # for example the httpd command line.
     filepath: Optional[str]
     """
+
+    @abc.abstractmethod
+    def __init__(self, ancestor=None, filepath="", dirty=False):  # pragma: no cover
+        """
+        Initializes the ParserNode instance, and sets the ParserNode specific
+        instance variables. This is not meant to be used directly, but through
+        specific classes implementing ParserNode interface.
+        """
 
     @abc.abstractmethod
     def save(self, msg):
@@ -163,6 +172,20 @@ class CommentNode(ParserNode):
 
     """
 
+    # pylint: disable=super-init-not-called
+    @abc.abstractmethod
+    def __init__(self, comment, ancestor, filepath, dirty=False):  # pragma: no cover
+        """
+        Initializes the CommentNode instance and sets its instance variables.
+
+        :param str comment: Contents of the comment.
+        :param BlockNode ancestor: BlockNode ancestor for this CommentNode.
+        :param str filepath: Filesystem path for the file where this CommentNode
+            does or should exist in the filesystem.
+        :param bool dirty: Boolean flag for denoting if this CommentNode has been
+            created or changed after the last save.
+        """
+
 
 @six.add_metaclass(abc.ABCMeta)
 class DirectiveNode(ParserNode):
@@ -171,6 +194,10 @@ class DirectiveNode(ParserNode):
     It can have zero or more parameters attached to it. Because of the nature of
     single directives, it is not able to have child nodes and hence it is always
     treated as a leaf node.
+
+    If a this directive was defined on the httpd command line, the ancestor instance
+    variable for this DirectiveNode should be None, and it should be inserted to the
+    beginning of root BlockNode children sequence.
 
     DirectiveNode objects should have the following attributes:
 
@@ -185,6 +212,26 @@ class DirectiveNode(ParserNode):
     parameters: Tuple[str, ...]
 
     """
+    # pylint: disable=too-many-arguments, super-init-not-called
+    def __init__(self, name, parameters=(), ancestor=None, filepath="",
+                 dirty=False, enabled=True):  # pragma: no cover
+        """
+        Initializes the DirectiveNode instance and sets its instance variables.
+
+        :param name: Name or key of the DirectiveNode object.
+        :param tuple parameters: Tuple of str parameters for this DirectiveNode.
+        :param ancestor: BlockNode ancestor for this DirectiveNode, or None for
+            directives introduced at the httpd command line.
+        :param str filepath: Filesystem path for the file where this DirectiveNode
+            does or should exist in the filesystem, or None for directives introduced
+            in the httpd command line.
+        :param bool dirty: Boolean flag for denoting if this DirectiveNode has been
+            created or changed after the last save.
+        :param bool enabled: True if this DirectiveNode object is parsed in the active
+            configuration of the httpd. False if the DirectiveNode exists within a
+            unmatched conditional configuration block.
+
+        """
 
     @abc.abstractmethod
     def set_parameters(self, parameters):
@@ -238,8 +285,9 @@ class BlockNode(ParserNode):
     # parsed, but its children should be flagged as disabled.
     enabled: bool
 
-    # Name, or key of the configuration directive
-    name: str
+    # Name, or key of the configuration directive. If the BlockNode is the root
+    # configuration node, the name should be None.
+    name: Optional[str]
 
     # Tuple of parameters of this ParserNode object, excluding whitespaces.
     parameters: Tuple[str, ...]
@@ -248,6 +296,28 @@ class BlockNode(ParserNode):
     # of the children is the same s that of the parsed configuration block.
     children: Tuple[ParserNode, ...]
     """
+
+    # pylint: disable=too-many-arguments, super-init-not-called
+    @abc.abstractmethod
+    def __init__(self, name="", parameters=(), children=(), ancestor=None,
+                 filepath="", dirty=False, enabled=True):  # pragma: no cover
+        """
+        Initializes the BlockNode instance and sets its instance variables.
+
+        :param name: Name or key of the BlockNode object, or None for root
+            configuration node.
+        :param tuple parameters: Tuple of str parameters for this BlockNode.
+        :param tuple children: Tuple of ParserNode children for this BlockNode.
+        :param ancestor: BlockNode ancestor for this BlockNode, or None for root
+            configuration node.
+        :param str filepath: Filesystem path for the file where this BlockNode does
+            or should exist in the filesystem.
+        :param bool dirty: Boolean flag for denoting if this BlockNode has been
+            created or changed after the last save.
+        :param bool enabled: True if this BlockNode object is parsed in the active
+            configuration object by the httpd. False if the BlockNode exists within
+            a unmatched conditional configuration block.
+        """
 
     @abc.abstractmethod
     def add_child_block(self, name, parameters=None, position=None):
--- a/certbot-apache/certbot_apache/interfaces.py
+++ b/certbot-apache/certbot_apache/interfaces.py
@@ -135,6 +135,12 @@ class ParserNode(object):
         Initializes the ParserNode instance, and sets the ParserNode specific
         instance variables. This is not meant to be used directly, but through
         specific classes implementing ParserNode interface.
+
+        :param ancestor: BlockNode ancestor for this CommentNode.
+        :param filepath: Filesystem path for the file where this CommentNode
+            does or should exist in the filesystem.
+        :param bool dirty: Boolean flag for denoting if this CommentNode has been
+            created or changed after the last save.
         """
 
     @abc.abstractmethod
@@ -182,8 +188,8 @@ class CommentNode(ParserNode):
         Initializes the CommentNode instance and sets its instance variables.
 
         :param str comment: Contents of the comment.
-        :param BlockNode ancestor: BlockNode ancestor for this CommentNode.
-        :param str filepath: Filesystem path for the file where this CommentNode
+        :param ancestor: BlockNode ancestor for this CommentNode.
+        :param filepath: Filesystem path for the file where this CommentNode
             does or should exist in the filesystem.
         :param bool dirty: Boolean flag for denoting if this CommentNode has been
             created or changed after the last save.
@@ -209,7 +215,9 @@ class DirectiveNode(ParserNode):
     # inactive conditional block. Default: True
     enabled: bool
 
-    # Name, or key of the configuration directive. Required.
+    # Name, or key of the configuration directive. If BlockNode subclass of
+    # DirectiveNode is the root configuration node, the name should be None.
+    # Required.
     name: str
 
     # Tuple of parameters of this ParserNode object, excluding whitespaces.
@@ -217,7 +225,7 @@ class DirectiveNode(ParserNode):
     parameters: Tuple[str, ...]
 
     """
-    # pylint: disable=too-many-arguments, super-init-not-called
+    # pylint: disable=super-init-not-called
     def __init__(self, **kwargs):
         """
         Initializes the DirectiveNode instance and sets its instance variables.
@@ -225,8 +233,8 @@ class DirectiveNode(ParserNode):
         :param name: Name or key of the DirectiveNode object.
         :param tuple parameters: Tuple of str parameters for this DirectiveNode.
         :param ancestor: BlockNode ancestor for this DirectiveNode, or None for
-            directives introduced at the httpd command line.
-        :param str filepath: Filesystem path for the file where this DirectiveNode
+            root configuratio node.
+        :param filepath: Filesystem path for the file where this DirectiveNode
             does or should exist in the filesystem, or None for directives introduced
             in the httpd command line.
         :param bool dirty: Boolean flag for denoting if this DirectiveNode has been
@@ -250,7 +258,7 @@ class DirectiveNode(ParserNode):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class BlockNode(ParserNode):
+class BlockNode(DirectiveNode):
     """
     BlockNode class represents a block of nested configuration directives, comments
     and other blocks as its children. A BlockNode can have zero or more parameters
@@ -281,26 +289,9 @@ class BlockNode(ParserNode):
     The applicable parameters are dependent on the underlying configuration language
     and its grammar.
 
-    BlockNode objects should have the following attributes in addition to
-    the ones described in ParserNode:
-
-    # True if this BlockNode is enabled and False if it is inside of an
-    # inactive conditional block. If a BlockNode contains an unmatched
-    # conditional statement, it should itself be flagged as enabled as it's
-    # parsed, but its children should be flagged as disabled. Default: True
-    enabled: bool
-
-    # Name, or key of the configuration directive. If the BlockNode is the root
-    # configuration node, the name should be None. Required.
-    name: Optional[str]
-
-    # Tuple of parameters of this ParserNode object, excluding whitespaces.
-    # Default: ()
-    parameters: Tuple[str, ...]
-
     """
 
-    # pylint: disable=too-many-arguments, super-init-not-called
+    # pylint: disable=super-init-not-called
     @abc.abstractmethod
     def __init__(self, **kwargs):
         """
@@ -441,17 +432,6 @@ class BlockNode(ParserNode):
 
         :param ParserNode child: Child ParserNode object to remove from the list
             of children of the callee.
-        """
-
-    @abc.abstractmethod
-    def set_parameters(self, parameters):
-        """
-        Sets the sequence of parameters for this ParserNode object without
-        whitespaces. While the whitespaces for parameters are discarded when using
-        this method, the whitespacing preceeding the ParserNode itself should be
-        kept intact.
-
-        :param list parameters: sequence of parameters
         """
 
     @abc.abstractmethod
--- a/certbot-apache/certbot_apache/interfaces.py
+++ b/certbot-apache/certbot_apache/interfaces.py
@@ -116,16 +116,16 @@ class ParserNode(object):
     ParserNode objects should have the following attributes:
 
     # Reference to ancestor node, or None if the node is the root node of the
-    # configuration tree. Required.
+    # configuration tree.
     ancestor: Optional[ParserNode]
 
-    # True if this node has been modified since last save. Default: False
+    # True if this node has been modified since last save.
     dirty: bool
 
     # Filepath of the file where the configuration element for this ParserNode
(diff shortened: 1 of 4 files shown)


certbot-apache/certbot_apache/interfaces.py
```diff
@@ -242,7 +208,7 @@ def set_parameters(self, parameters):
 
 
 @six.add_metaclass(abc.ABCMeta)
-class BlockNode(ParserNode):
+class BlockNode(object):
```
bmw: Why remove the inheritance from `ParserNode`?

joohoi: I thought the suggestion for using ABC `register()` functionaity was aimed towards the interfaces as well. I hit some linter errors about not implementing the superclass methods and thought that if we'd be using the register() regardless, it would not matter.

I can return the inheritance and add the linter exceptions there as well though.

bmw: My suggestion to use `register` [here](https://github.com/certbot/certbot/pull/7308#discussion_r313650641) was to define an `AugeasParserNode` _implementation_ so that we weren't duplicating the same `ParserNode` code 3 times and use it without resorting to multiple inheritance. This looks something like:
```
class _AugeasParserNode(interfaces.ParserNode):
    def save():
    ....
class AugeasCommentNode(_AugeasParserNode):
    def __init__():
    ....
AugeasCommentNode.register(interfaces.CommentNode)
```
You could also probably have `_AugeasParserNode` inherit from `object` and register it if you prefer.

As for the interfaces modified in this PR, the three real node interfaces are now defined in a way where to implement the class you do not need to defined the `ParserNode` methods which seems like a problem that should be fixed to me instead of making implementations explicitly register themselves with multiple interfaces. I don't think we ever want a `CommentNode` that is not a `ParserNode`.

I think continuing to use inheritance between the interfaces and fixing the `pylint` issue which is a [known bug in `pylint`](https://github.com/PyCQA/pylint/issues/179) is the best approach.

bmw: My suggestion to use `register` [here](https://github.com/certbot/certbot/pull/7308#discussion_r313650641) was to define an `AugeasParserNode` _implementation_ so that we weren't duplicating the same `ParserNode` code 3 times and use it without resorting to multiple inheritance. This looks something like:
```
class _AugeasParserNode(interfaces.ParserNode):
    def save():
    ....
class AugeasCommentNode(_AugeasParserNode):
    def __init__():
    ....
AugeasCommentNode.register(interfaces.CommentNode)
```
You could also probably have `_AugeasParserNode` inherit from `object` and register it if you prefer.

As for the interfaces modified in this PR, the three real node interfaces are now defined in a way where to implement the class you do not need to defined the `ParserNode` methods which seems like a problem that should be fixed to me instead of making implementations explicitly register themselves with multiple interfaces. I don't think we ever want a `CommentNode` that is not a `ParserNode`.

I think continuing to use inheritance between the interfaces and fixing the `pylint` issue which is a [known bug in `pylint`](https://github.com/PyCQA/pylint/issues/179) is the best approach.

joohoi: Thanks for the pointer! I was getting pretty frustrated there. Unfortunately only a file-wide pylint exclude worked here for a reason or another. It's now fixed however.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -110,35 +113,30 @@ class ParserNode(object):
...
+    # configuration tree.
+    ancestor: Optional[ParserNode]
+
+    # True if this node has been modified since last save.
+    dirty: bool
 
-        raise NotImplementedError
+    # Filepath of the file where the configuration element for this ParserNode
+    # object resides.
+    filepath: str
```
bmw: What should filepath be for the root node?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -110,35 +113,30 @@ class ParserNode(object):
...
+    """
 
-    @property
     @abc.abstractmethod
-    def dirty(self):  # pragma: no cover
+    def metadata(self, key):
```
bmw: Just to make sure we're on the same page, you're planning on deleting this method based on https://github.com/certbot/certbot/pull/7308#discussion_r314388990 right?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -167,20 +165,14 @@ class CommentNode(ParserNode):
...
+    CommentNode objects should have the following attributes:
 
-    @property
-    @abc.abstractmethod
-    def comment(self):  # pragma: no cover
-        """
-        Comment property contains the contents of the comment without the comment
-        directives (typically # or /* ... */).
+    # Contains the contents of the comment without the directive notation
+    # (typically # or /* ... */
```
bmw:
```suggestion
    # (typically # or /* ... */).
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -272,6 +239,24 @@ class BlockNode(ParserNode):
...
+    # inactive conditional block. If a BlockNode contains an unmatched
+    # conditional statement, it should itself be flagged as enabled as it's
+    # parsed, but its children should be flagged as disabled.
+    enabled: bool
+
+    # Name, or key of the configuration directive
+    name: str
+
+    # Tuple of parameters of this ParserNode object, excluding whitespaces.
+    parameters: Tuple[str]
```
bmw:
```suggestion
    parameters: Tuple[str, ...]
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -189,45 +181,20 @@ class DirectiveNode(ParserNode):
...
+    name: str
 
-    @property
-    @abc.abstractmethod
-    def parameters(self):  # pragma: no cover
-        """
-        This property contains a tuple of parameters of this ParserNode object
-        excluding whitespaces.
+    # Tuple of parameters of this ParserNode object, excluding whitespaces.
+    parameters: Tuple[str]
```
bmw:
```suggestion
    parameters: Tuple[str, ...]
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -272,6 +239,24 @@ class BlockNode(ParserNode):
...
+
+    # Name, or key of the configuration directive
+    name: str
+
+    # Tuple of parameters of this ParserNode object, excluding whitespaces.
+    parameters: Tuple[str]
+
+    # Tuple of ParserNode objects that are the children for this node. The order
+    # of the children is the same s that of the parsed configuration block.
+    children: Tuple[ParserNode]
```
bmw:
```suggestion
    children: Tuple[ParserNode, ...]
```

certbot-apache/certbot_apache/tests/parsernode_test.py
```diff
@@ -8,28 +8,41 @@
 
 
 
-class DummyCommentNode(interfaces.CommentNode):
+class DummyCommentNode(interfaces.ParserNode):
     """ A dummy class implementing CommentNode interface """
     ancestor = None
     comment = ""
     dirty = False
+    filename = ""
```
bmw: Can we define these attributes in `__init__` for this and all other classes in this file?

certbot-apache/certbot_apache/tests/parsernode_test.py
```diff
@@ -41,38 +54,57 @@ class DummyBlockNode(interfaces.BlockNode):
...
         pass
 
     def unsaved_files(self):  # pragma: no cover
+        """Unsaved files"""
         pass
 
 
+interfaces.CommentNode.register(DummyCommentNode)
+interfaces.DirectiveNode.register(DummyDirectiveNode)
+interfaces.BlockNode.register(DummyBlockNode)
```
bmw: For the classes in this file I think we should either:

1. Define a `DummyParserNode` and have a setup similar to what I described at https://github.com/certbot/certbot/pull/7330#discussion_r314443836 in this file as well. This is my preference as I would have complained about the dummy duplicated methods in this file if I would have noticed.
2. Delete these class to `register` and have each class inherit directly from the interface it implements as is currently in `apache-parser-v2`. I don't think separately inheriting from `interfaces.ParserNode` in each of these classes here does anything significant.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=abstract-method
```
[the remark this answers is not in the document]
joohoi: Sure! Sounds good!

certbot-apache/certbot_apache/interfaces.py
joohoi: I think the root node should point to the main config which we start parsing from, even though it doesn't have a key name.

certbot-apache/certbot_apache/interfaces.py
joohoi: That's right!

certbot-apache/certbot_apache/interfaces.py
joohoi: Oh, right, as it'll be empty if no parameters are defined.

certbot-apache/certbot_apache/tests/parsernode_test.py
joohoi: We can, do you think we should do that in context of this PR, or the VirtualHost creation one?

certbot-apache/certbot_apache/tests/parsernode_test.py
joohoi: I went with the first option as I feel it's a cleaner one.

certbot-apache/certbot_apache/interfaces.py
joohoi: I made it `Optional[str]` as while if we go with the root node having the file path of the Apache2 root config file, if directives are added on Apache command line, those will not have a filepath.

bmw: Works for me. Can you document this stuff in the comment for `filepath`?

certbot-apache/certbot_apache/tests/parsernode_test.py
bmw: I have a slight preference for this one since it's the one making the change that these no longer need to be class variables, but I don't really care.

bmw:
Yikes. Reviewing a PR in multiple tabs gets funky sometimes. My other comment part of this review was https://github.com/certbot/certbot/pull/7330#discussion_r314859228.

joohoi:
The initialization methods have now been defined in this PR. I'm having some doubts about the parameters that should be optional, as I think that it's counterintuitive if the parameter order changes between different ParserNode classes. If you have any thoughts about this, I'd love to hear them.

bmw:
> The initialization methods have now been defined in this PR. I'm having some doubts about the parameters that should be optional, as I think that it's counterintuitive if the parameter order changes between different ParserNode classes. If you have any thoughts about this, I'd love to hear them.

We talked about this out of band and the approach we agreed on is keyword only arguments. This forces us to write more readable code when creating the objects and avoids the problem of argument order.

I'm open to changing this but here's my opinion as a starting point. I think we should land the code described here in this PR or another PR before #7338 because it requires all class implementations change their approach to `__init__`.

We could allow some arguments to be passed by position where the purpose of the parameter is clear such as `comment`, `name`, and `parameters`, but honestly, I'm leaning towards just making everything a keyword argument and we can change the approach later if we want for some reason. With this approach, each `__init__` method would look like:
```
def __init__(self, **kwargs):
```
To implement this, I recommend making a `util.py` or `parser_node_util.py` file. This file would contain a function named something like `validate_kwargs`. My thought is its only arguments should be the `kwargs` dict and an iterable of strings containing the names of the expected keyword arguments.

This function iterates over the list of `kwargs`, popping each item in the list. If an argument isn't found, it raises a `TypeError` containing a simple error message about the missing keyword. If after iterating over all expected arguments, kwargs isn't empty, it also raises a `TypeError` about the extra keyword. I think this function should return a new dict with the validated keyword arguments.

We'd then have a function building on that like `validate_parser_node_kwargs` which only takes `kwargs` as an argument. This function already knows the expected keywords. This function first calls `kwargs.setdefault` for each parameter that has a default value. It then calls `validate_kwargs` with the full list of keyword arguments. Instead of returning the dict though, I think it should return a tuple of the expected keyword arguments. The return code would be something like:
```
return validated_kwargs['ancestor'], validated_kwargs['dirty'], validated_kwargs['filepath']
```
I think this approach is useful later.

Finally, we'd have methods like `validate_comment_node_kwargs`. I don't think the code in `validate_parser_node_kwargs` can be reused because things like defaults may differ. Instead, it does the same thing as `validate_parser_node_kwargs` but only split the specific class's arguments from the dict leaving the arguments for the base class in the dict. For a comment node, this would look like:
```
...
comment = validated_kwargs.pop('comment')
return comment, validated_kwargs
```
This is nice because the use of this method in an implementation of `CommentNode` would be:
```
def __init__(self, **kwargs):
    comment, kwargs = util.validate_comment_node_kwargs(kwargs)
    super(DummyCommentNode, self).__init__(**kwargs)
    ...
```
What do you think?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -123,10 +123,19 @@ class ParserNode(object):
...
+    # for example the httpd command line.
     filepath: Optional[str]
     """
 
+    @abc.abstractmethod
+    def __init__(self, ancestor=None, filepath="", dirty=False):  # pragma: no cover
+        """
+        Initializes the ParserNode instance, and sets the ParserNode specific
+        instance variables. This is not meant to be used directly, but through
+        specific classes implementing ParserNode interface.
```
bmw: If we take the approach I described at https://github.com/certbot/certbot/pull/7330#issuecomment-522795095, we should be sure to document the parameters and their defaults for each class in their docstring.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -123,10 +123,19 @@ class ParserNode(object):
...
 
     # Filepath of the file where the configuration element for this ParserNode
-    # object resides.
+    # object resides. This can be None if a configuration directive is defined in
+    # for example the httpd command line.
     filepath: Optional[str]
     """
 
+    @abc.abstractmethod
+    def __init__(self, ancestor=None, filepath="", dirty=False):  # pragma: no cover
```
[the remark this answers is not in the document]
bmw: I have my same feeling as I did at https://github.com/certbot/certbot/pull/7308#discussion_r314111032 where I'm somewhat tempted to not give `ancestor` a default since not having one is uncommon, but you can ignore that if you prefer.

I feel the same way about `filepath`, but if you prefer to give it one, I think it should be `None` and not the empty string.

These comments apply to some of the derived classes as well.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -186,6 +212,26 @@ class DirectiveNode(ParserNode):
     parameters: Tuple[str, ...]
 
     """
+    # pylint: disable=too-many-arguments, super-init-not-called
+    def __init__(self, name, parameters=(), ancestor=None, filepath="",
```
bmw: I'm probably getting ahead of myself so you can ignore this comment for now if you prefer, but I was thinking we'll have to change these parameters a bit in the future because we need a way to pass in things like whitespace. With this in mind, maybe we don't need/want `name` and `parameter` options and it'll instead be passed through metadata. If we accept both, it's redundant and leaving ourselves open to the parameters differing.

This applies to the `BlockNode` as well.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -250,6 +297,28 @@ class BlockNode(ParserNode):
     children: Tuple[ParserNode, ...]
     """
 
+    # pylint: disable=too-many-arguments, super-init-not-called
+    @abc.abstractmethod
+    def __init__(self, name="", parameters=(), children=(), ancestor=None,
```
bmw: I think we should remove this `children` parameter and instead just use the `add_child*` methods. The reason for this is:

1. Proper children need to have `ancestor` set which is impossible before this object is created.
2. During parsing, I'm struggling to imagine a scenario where we've parsed a block's children before the block itself.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -186,6 +212,26 @@ class DirectiveNode(ParserNode):
     parameters: Tuple[str, ...]
 
     """
+    # pylint: disable=too-many-arguments, super-init-not-called
+    def __init__(self, name, parameters=(), ancestor=None, filepath="",
+                 dirty=False, enabled=True):  # pragma: no cover
```
bmw: There's probably a balance to be struck here, but I'm personally leaning towards removing a lot of/all of these default values. I think default values makes it easy for you to forget to add a parameter which is especially true where these functions take so many parameters.

What do you think?

certbot-apache/certbot_apache/tests/parsernode_test.py
```diff
@@ -13,14 +13,29 @@ class DummyParserNode(interfaces.ParserNode):
     dirty = False
     filepath = None
 
+    def __init__(self, ancestor=None, filepath=str(), dirty=False):
```
bmw: Since we're adding `__init__` in this PR now too, can we delete the class variables in this file?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -189,45 +194,44 @@ class DirectiveNode(ParserNode):
...
-    @abc.abstractmethod
-    def name(self):  # pragma: no cover
-        """
-        Name property contains the name of the directive.
-
-        :returns: Name of this node
-        :rtype: str
-        """
+    # Name, or key of the configuration directive
+    name: str
```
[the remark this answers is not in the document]
joohoi: While I do like this approach, it can make life harder in other parts of the code, as BlockNodes would also match `isinstance(DirectiveNode, ...)`. This is _mainly_ a concern regarding assertion tests, but maintaining the correct order to check the types might prove unnecessarily hard.

certbot-apache/certbot_apache/interfaces.py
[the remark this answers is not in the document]
joohoi: I agree and made both of these required now.

certbot-apache/certbot_apache/interfaces.py
joohoi: I agree. The only ones that I think might make sense to keep optional would be `dirty` and `enabled`.

certbot-apache/certbot_apache/tests/parsernode_test.py
joohoi: Yeah, I forgot to remove them previously!

certbot-apache/certbot_apache/interfaces.py
joohoi: ... then again, making all the arguments required, makes the implementation of https://github.com/certbot/certbot/pull/7330#issuecomment-522795095 much easier, as all the expected parameters are also required.

certbot-apache/certbot_apache/interfaces.py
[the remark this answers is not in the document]
bmw: I don't think it'll be that bad. I don't think it'll come up when using these objects normally because you're normally dealing with a node for specific directives and I only see two instances of `isinstance(var, DirectiveNode)` in #7338.

I'm happy to chat about this here, Mattermost, or during a 1-1, but I honestly feel pretty strongly about this. I think there's a significant benefit of code reuse and I think it results in a more accurate design. Of course a `BlockNode` is a `DirectiveNode`. It's just a special type of directive that can contain other directives.

As a potential compromise, what if we made a `BaseDirectiveNode` and `DirectiveNode` and `BlockNode` that inherit from it? `DirectiveNode` adds nothing but a distinct class name that is unrelated to BlockNodes.

certbot-apache/certbot_apache/interfaces.py
bmw: Right. I don't care too much about this and default parameters are fairly common in Python code, but I think:

1. There are times when we'll want to remember to set directives even those like `dirty` and `enabled` to a different value (such as adding a new directive to the config and parsing a disabled block respectively) and writing the code such that we are forced to set them could help prevent bugs.
2. Creation of these objects will happen almost entirely inside the implementations themselves. The pain of doing this won't really escape outside of the class I think.

certbot-apache/certbot_apache/interfaces.py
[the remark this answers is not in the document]
joohoi: I do like the compromise option but at the same time I'm not very happy with the added complexity that could possibly be avoided. Because of this, I think I'm ok with going with the original proposal, and we'll change this later if a need arises.

certbot-apache/certbot_apache/interfaces.py
joohoi: Oh wow, indeed. Sorry about this.

certbot-apache/certbot_apache/interfaces.py
joohoi: I think we should definitely address this, but in a later PR. However I've been thinking of the metadata as _everything else than what's already defined in the interface_ , and this is largely based on the idea I was fixed on, but now realize isn't relevant: storing the exact tokenized results of parsing within the ParserNode object. This would have meant that we change the AST bits only if the object is `dirty` upon `save()`. I see that this doesn't (most likely) stand true with the current plans we have.

However to keep everything simple and easy to access. I think what I'd personally prefer would be an additional method to `BlockNode` (the root node needs to be created "manually" anyway), something along the lines of `add_raw_child_block` that would take in the metadata instead of values for `name` and `parameters`. This would keep the interface easy and idiomatic to use, while providing a way for the internal parser functionality to initialize the objects in way that would preserve all the required information. I think this would be a rather small addition, but I'd prefer to do it in a separate PR in order to avoid growing this one larger.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -116,24 +116,31 @@ class ParserNode(object):
     ParserNode objects should have the following attributes:
 
     # Reference to ancestor node, or None if the node is the root node of the
-    # configuration tree.
+    # configuration tree. Required.
```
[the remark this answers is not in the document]
joohoi: Yeah, that's right. these are now documented in both places. Do you think they should be removed from the class docstrings?

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -189,45 +203,47 @@ class DirectiveNode(ParserNode):
...
-        raise NotImplementedError
+    # Tuple of parameters of this ParserNode object, excluding whitespaces.
+    # Default: ()
+    parameters: Tuple[str, ...]
 
-    @property
-    @abc.abstractmethod
-    def parameters(self):  # pragma: no cover
+    """
+    # pylint: disable=super-init-not-called
```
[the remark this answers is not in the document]
joohoi: This is now done, it required another linter exclusion, but I'm not able to come up with any scenario where this would be problematic.

certbot-apache/certbot_apache/interfaces.py
joohoi: The alternative approach here would be initializing the instances with sequences including whitespaces - an approach I would personally prefer over having all this in metadata. 

My opinion mostly boils down to name and parameters being first class citizens that form the core of the whole node, while other information would and could belong to the metadata category.

joohoi: The alternative approach here would be initializing the instances with name and parameter sequences including whitespaces - an approach I would personally prefer over having all this in metadata. 

My opinion mostly boils down to name and parameters being first class citizens that form the core of the whole node, while other information would and could belong to the metadata category.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -136,11 +136,11 @@ def __init__(self, **kwargs):
         instance variables. This is not meant to be used directly, but through
         specific classes implementing ParserNode interface.
 
-        :param ancestor: BlockNode ancestor for this CommentNode.
+        :param ancestor: BlockNode ancestor for this CommentNode. Required.
```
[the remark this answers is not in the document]
joohoi: I agree, and I have tried to do this where plausible. The ones that are missing the type definition are ones that might also be `None` for a reason or another, as the type definition there would feel like it's potentially a wrong one.

bmw:
This is so close! I think the only outstanding comment is https://github.com/certbot/certbot/pull/7330#discussion_r318789117.

certbot-apache/certbot_apache/interfaces.py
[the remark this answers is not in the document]
joohoi: Oh of course. I have not used this style of notation myself previously. Thanks!

joohoi:
The missing types were added, and everything should now be ready

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -110,34 +112,41 @@ class ParserNode(object):
...
+        :type ancestor: BlockNode or None
 
-        raise NotImplementedError
+        :param filepath: Filesystem path for the file where this CommentNode
+            does or should exist in the filesystem. Required.
+        :type filepath: str or None
+
+        :param dirty: Boolean flag for denoting if this CommentNode has been
+            created or changed after the last save. Default: False.
+        :type dirty: bool or None
```
bmw: If `dirty` is set, it should always be set to a boolean.
```suggestion
        :type dirty: bool
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -189,45 +219,61 @@ class DirectiveNode(ParserNode):
...
+    DirectiveNode objects should have the following attributes in addition to
+    the ones described in ParserNode:
+
+    # True if this DirectiveNode is enabled and False if it is inside of an
+    # inactive conditional block.
+    enabled: bool
+
+    # Name, or key of the configuration directive. If BlockNode subclass of
+    # DirectiveNode is the root configuration node, the name should be None.
+    name: str
```
bmw: For the case of the root node
```suggestion
    name: Optional[str]
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -272,6 +318,7 @@ class BlockNode(ParserNode):
 
     The applicable parameters are dependent on the underlying configuration language
     and its grammar.
+
```
bmw: Looks like we need a description of the `children` here. We deleted the property below but didn't document it here.

PR #7336, opened 2019-08-16
Update Debian instructions in docs

bmw:
In response to https://community.letsencrypt.org/t/debian-jessie-certbot-installation/100038/2.

@schoen, are you interested in reviewing this PR?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -200,7 +200,7 @@ Operating System Packages
 
 **Debian**
 
-If you run Debian Buster or Debian testing/sid, you can easily install certbot
+If you run Debian Buster or Debian testing/Sid, you can easily install certbot
 packages through commands like:
 
 .. code-block:: shell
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -200,23 +200,36 @@ Operating System Packages
...
+.. code-block:: shell
 
-If you don't want to use the Apache plugin, you can omit the
-``python-certbot-apache`` package. Or you can install ``python-certbot-nginx`` instead.
+   sudo apt-get install certbot -t stretch-backports
 
-Packages exist for Debian Jessie via backports. First you'll have to follow the
-instructions at http://backports.debian.org/Instructions/ to enable the Jessie backports
-repo, if you have not already done so. Then run:
+In all of these cases, there also packages available to help Certbot integrate
```
schoen: Maybe we should emphasize this more in order to get users to pay attention to the fact that they will probably need at least one of these additional packages. (At a minimum, we could say "Almost all users will want to install packages that help Certbot integrate with other software" or something like that.)

docs/install.rst
```diff
@@ -200,23 +200,36 @@ Operating System Packages
...
 
 .. code-block:: shell
 
-   sudo apt-get install certbot python-certbot-apache -t jessie-backports
+    apt search python-certbot*
+
+They can be installed by running the same installation command above but
+replace ``certbot`` with the name of the desired package.
+
+There are no Certbot packages available for Debian Jessie and you should
```
schoen: How about "and Jessie users should"?

bmw:
I made every change exactly as suggested except for the comment at https://github.com/certbot/certbot/pull/7336#discussion_r317264898. Let me know if you think we should make additional changes here.

Issue #7106, opened 2019-05-29
Stop using platform.linux_distribution in Python 3.8+


bmw:
See https://bugzilla.redhat.com/show_bug.cgi?id=1705300.

The relevant code is in certbot/util.py and its tests. The code is currently a fallback if other mechanisms don't work and is being used both for our user-agent string and in the Apache plugin. We could either remove the fallback completely or don't offer it on Python 3.8+. I've only briefly looked at the issue, but I'm personally leaning towards the latter to prevent regressions on current installations.

cc @joohoi

elyscape:
The Python documentation mentions that [the distro package][distro] can be used to get this information for Python 3.8+. That's an option as well.

[distro]: https://pypi.org/project/distro/

bmw:
I'm tentatively assigning this to Joona who said he would work on this if he found the time to help us track who is doing what.

joohoi:
Looks like we could get away with just removing the `platform.linux_distribution()` fallback, as we're using `/etc/os-release` in the first place anyway. The main concern was keeping supporting non-systemd distributions as the file is a systemd feature. Fortunately it seems to be adopted by other distros as well:

Non-systemd distributions confirmed having `/etc/os-release` file:
* Devuan
* Void Linux
* Alpine Linux
* Slackware

That's basically every non-systemd distribution I checked.

Additional notes:
Apache plugin relies heavily on the distribution flavor detection, so just to be sure, we should add a CLI option to override the [override selection code](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/entrypoint.py) (that's a lot of overriding).

joohoi:
Scientific Linux 6 does **not** have an `/etc/os-release` file, while the next version - 7 does. While [Scientific Linux is being discontinued](https://listserv.fnal.gov/scripts/wa.exe?A2=SCIENTIFIC-LINUX-ANNOUNCE;11d6001.1904) the current versions will be supported as they are based on CentOS and the SL 6 shares CentOS 6 EOL date - 2020-11-30.

The current information seems to suggest that SL is the only distribution without the file - and now looking at CentOS 6 again, it looks like it doesn't ship with the file either. In my previous checks the file has been there for CentOS 6 - probably installed by some other package. If these are the only distributions not having the `/etc/os-release` I think we should default to `centos` override template in case the file isn't found. Hard requirement for this is for the user to be able to override the override template selection as noted in the previous comment.

bmw:
[...]

PR #7337, opened 2019-08-16
Replace platform.linux_dependencies with distro.linux_dependencies

ohemorange:
Smallest possible fix for #7106.

## Pull Request Checklist

  the change being made.
- [ ] Add [mypy type


ohemorange:
Full tests running here: https://travis-ci.com/certbot/certbot/builds/123571821

I suspect this small change will mostly just work, because we only try to use it on platforms where we believe it's available. But rather than investigating each distro manually, that's what a test matrix is for! If someone has reason to believe the test matrix won't catch this, please chime in.

adferrand:
Maybe we could block on https://github.com/certbot/certbot/pull/7268 here? Assume Python 3.4 anywhere will become more and more difficult to support I think.

bmw:
I don't think that's a good idea. While I want to land #7268 as soon as I (or someone else) has cycles to review it, I don't think blocking this on that PR is necessary and doing so prevents us from fixing an issue that definitely will cause packaging problems for us in a couple months.

While I think my suggestion of pinning `requests` is fine and allows us to use `rebuild_dependencies.py` until the much more complex #7268 is merged, another option is to not use `rebuild_dependencies.py` for this PR at all and instead follow the instructions for a single dependency at https://github.com/certbot/certbot/pull/7337/files#diff-f357548877ed4fd3a64223f9fe586512R7. The `distro` package doesn't have any dependencies. See https://github.com/nir0s/distro/blob/v1.4.0/setup.py.

adferrand:
Python 3.8 is planned to be released by the end of October. Do you expect the PR that removes Python 3.4 will take two months to be settled?

It is just that pinning `requests` (so blocking any potential necessary upgrade here) is not the more elegant solution, and not needed in fact before two months. In the mean time, we could deploy the elegant one, which is just about dropping the version of Python that have been deprecated for a while and that is the root cause, and necessary anyway in the short term now. 

However if it is about pinning requests, I would largely prefer indeed to do it in `AUTHORITATIVE_CONSTRAINTS`, that I designed for this specific purpose: it keeps track of why it is pinned, and allow it to be reused at any future call to `rebuild_dependencies.py`, instead of letting this customization been removed unexpectedly and so potentially merged if we are not careful.

bmw:
> Python 3.8 is planned to be released by the end of October. Do you expect the PR that removes Python 3.4 will take two months to be settled?

No, but @maximillianh and I planned for this to be done this month to help give packagers time to fix things up on our end while we weren't planning on doing #7268 until September.

I'm done responding to this thread. If you'd like to talk about it more, let's talk during our 1-1.

bmw:
Adrien and I talked about this more and decided not to block this PR on #7268 and that if we update `rebuild_dependencies.py` to pin back `requests`, we should create an issue to remember to unpin it once we no longer support Python 3.4.

ohemorange:
Oldest versions:
- stretch: 1.0.1
- bionic: 1.0.1
- epel: 1.0.3

Unfortunately, the package doesn't exist at all in either jessie or xenial or their backports (in either python 2 or 3 version).

For jessie, we use certbot-auto, so theoretically it should be installing from pip, right? Although if that's correct, I don't understand why oldest-constraints has a jessie backports constraints section.

For xenial, I suspect we could include whatever version we want in the PPA, unless I've completely misunderstood how PPAs work.

ohemorange:
So, there's also another note at the bottom about `urllib3` support:
> `# package urllib3 is declared with several versions: 1.24.3 (centos:6), 1.25.3 (ubuntu:18.04,ubuntu:16.04,debian:stretch,debian:jessie,centos:7,opensuse/leap:15,fedora:29)`

I'm interpreting this to mean "you need to set `urllib3` version to 1.24.3 in `AUTHORITATIVE_CONSTRAINTS`. Is that accurate?

adferrand:
Not necessarily. The goal is to pin the least amount of dependencies to fix all errors.

Here `urllib3` version duplication is not a root cause, it is triggered by the `request` version duplication, for which `urllib3` is a transitive dependency.

Fix `request`, and `urllib3` will be fixed as well.

ohemorange:
Ah, thanks!

ohemorange:
New full test suite available here: https://travis-ci.com/certbot/certbot/builds/124442232

PR #7338, opened 2019-08-19
[Apache v2] Implement dual ParserNode implementation abstraction

joohoi:


certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,266 @@
+""" Tests for ParserNode interface """
+from certbot_apache import assertions
+from certbot_apache import augeasparser
+
+
+class DualNodeBase(assertions.InterfaceAssertions):
```
bmw: I think we should have a place to specify the classes that should be used for both primary and secondary inside these objects. This allows us to easily plug in the apacheconfig implementation and swap which is the primary without having to modify every place the class is referenced in the config.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,266 @@
...
+    """ Dual parser interface for in development testing. This is used as the
+    base class for dual parser interface classes. This class handles runtime
+    attribute value assertions."""
+
+    def save(self, msg):
+        """ Call save for both parsers """
+        self.primary.save(msg)
+        self.secondary.save(msg)
```
bmw: I think this should probably come in (much) later PRs, but just to flag it now, before we ship this code, I think we need to figure out more of the details of our migration plan and implement things like good reporting and error handling around all of the calls here.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,266 @@
...
+        if not primary:
+            self.primary = augeasparser.AugeasCommentNode(comment, ancestor,
+                                                          filepath, dirty)
+        else:
+            self.primary = primary
+        if not secondary:
+            self.secondary = augeasparser.AugeasCommentNode(comment, ancestor,
+                                                            filepath, dirty)
+        else:
+            self.secondary = secondary
```
bmw: I think at the bottom of these `__init__` methods we should also assert equality and doing so simplifies some of the code elsewhere such as creating children.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,266 @@
...
+        instances that encapsulate the pairs of returned DirectiveNode objects.
+        """
+        primary_directives = self.primary.find_directives(name, exclude)
+        secondary_directives = self.secondary.find_directives(name, exclude)
+
+        new_dirs = list()
+        # Check for assertion pass for results, as zip loses data otherwise.
+        # unimplemented find_* methods should return a single object with
+        # assertion pass.
+        if len(primary_directives) != len(secondary_directives):
```
bmw: nit: I think this `if/else` block can be simplified a lot and maybe even shared between different `find_*` implementations.

PR #7346, opened 2019-08-21
Dns rfc2136 zone

drdivano:
Zone detection logic _find_domain didn't work for me, so I added an option to explicitly provide zone name. Please review the code. If the general approach is ok, I can add the docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ before_script:
   - 'if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 1024 ; fi'
   # On Travis, the fastest parallelization for integration tests has proved to be 4.
   - 'if [[ "$TOXENV" == *"integration"* ]]; then export PYTEST_ADDOPTS="--numprocesses 4"; fi'
+  # Use Travis retry feature for farm tests since they are flaky
+  - 'if [[ "$TOXENV" == "travis-test-farm"* ]]; then export TRAVIS_RETRY=travis_retry; fi'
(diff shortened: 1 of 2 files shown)


bmw:
I think this is a duplicate of https://github.com/certbot/certbot/pull/6644 and https://github.com/certbot/certbot/pull/7244 right?

drdivano:
> I think this is a duplicate of #6644 and #7244 right?

It looks similar, but I think my case is distinct.

I have BIND zone file for my domain (example.com) under version control.

I add a separate zone file containing a TXT record for ACME challenge (_acme-challenge.example.com) and set up keys and permissions in BIND to be able to modify it via rfc2136 protocol.

Since domain name in SOA record in BIND can't have underscores, I can't put _acme-challenge.example.com in SOA, and therefore certbot can't correctly deduce the domain from SOA.

bmw:
Thanks for the additional info.

@adferrand or @joohoi, I think you both know more about this than me. When you get a few minutes, can you comment on this use case and whether or not it would be supported by the two PRs I reference above?

adferrand:
I think this situation is one of the pathological cases described by @hpax in #7244.

Here `_acme-challenge.example.com` is a zone on its own described out of the scope of the DNS server declared as SOA for the `example.com` zone. Typically a situation where recursive discovery will not work. 

So I think having a flag to explicitly set the zone name is a good complement to override any auto discovery result if needed, or circumvent the inability to use the auto discovery. 

Finally, if #7244 lands as a more generic implementation for any dns provider, this one should be generalized as well. If not, we can keep it for the sole purpose of rfc2136 plugin.

PR #7353, opened 2019-08-26
Execute certbot version in a different directory

ohemorange:
Python prefers files in its CWD over what's actually installed, so let's change directories before executing the tests so that things like https://travis-ci.com/certbot/certbot/jobs/227892868 pass.

### DIFF
--- a/certbot-ci/certbot_integration_tests/utils/misc.py
+++ b/certbot-ci/certbot_integration_tests/utils/misc.py
@@ -209,19 +209,6 @@ shutil.rmtree(well_known)
         shutil.rmtree(tempdir)
 
 
-def get_certbot_version(workspace):
-    """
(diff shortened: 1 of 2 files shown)


certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -209,13 +209,14 @@ def manual_http_hooks(http_server_root, http_port):
         shutil.rmtree(tempdir)
 
 
-def get_certbot_version():
+def get_certbot_version(workspace):
```
adferrand: Initially this logic was in a dedicated function because it was used in several places. But it is not the case anymore.

Also as a public method, passing a workspace to get the certbot version does not have an obvious meaning. 

So, what do you think about moving this logic right into the `certbot_call` module, since it is used only there?

PR #7355, opened 2019-08-27
Disable TLS session tickets in Nginx

### LINKED ISSUE #7351: Disable TLS session tickets in Nginx
With the resolution of #7322, we now seem to have a way to do this safely. The solution is not only to check that Nginx supports disabling TLS session tickets, but also to only do so if the version of OpenSSL used by Nginx is 1.0.2l or greater.

The best way to get the version of OpenSSL used by Nginx, as found by Erica, is:

> Parsing openssl version from `nginx -V`. Printing openssl version info was added in Nginx 1.8.0, so that should be fine, as long as we remember to check for the "running with" [parenthetical](https://github.com/nginx/nginx/blob/master/src/core/nginx.c#L435).

For error handling, I talked about this with people and right now we're thinking that:

1. If for some reason we fail to determine the OpenSSL version, we should keep session tickets at their default of enabled. Doing anything else isn't worth potentially breaking people's servers.
2. This could potentially be done in a later PR, but we should set up test farm tests that check that we can determine the OpenSSL version used by Nginx on all systems the tests are run on.

The combination of these two things should stop us from breaking people's servers if something changes around us detecting these versions and will let us know if this detection is broken on a major system that we support.

ohemorange:
Fixes #7351.

## Pull Request Checklist

  the change being made.


certbot-nginx/certbot_nginx/configurator.py
```diff
@@ -124,11 +131,33 @@ def __init__(self, *args, **kwargs):
...
+        recommendations. But TLS1.3 is only supported by newer versions of Nginx.
+        And as for session tickets, our ideal is to turn them off across the board.
+        But! Turning them off at all is only supported with new enough versions of
+        Nginx. And older versions of OpenSSL have a bug that leads to browser errors
+        given certain configurations. While we'd prefer to have forward secrecy, we'd
+        rather fail open than error out. Unfortunately, Nginx can be compiled against
+        many versions of OpenSSL. So we have to check both for the two different features,
+        leading to four different combinations of options.
+        For a complete history, check out https://github.com/certbot/certbot/issues/7322
+        """
```
sydneyli: smallest final thing, linter's unhappy that this is on its own line instead of a part of the function comment above

sydneyli:
linter still unhappy :( that the string is on its own line 
```
************* Module certbot_nginx.configurator
W:144,-1: String statement has no effect (pointless-string-statement)
```

ohemorange:
@sydneyli linter is happy

Issue #7165, opened 2019-06-18
OSError: [Errno 2] No such file or directory: in webroot.py on SmartOS with Python 2.7


martin-c:
## Operating System: SmartOS instance(base-64-lts 16.4.1) ##
 * Python version 2.7.12
 * certbot install via `wget https://dl.eff.org/certbot-auto` ...
## running command ##
`sudo /usr/local/bin/certbot-auto --debug -d [removed] --webroot --webroot-path /usr/local/www/[removed] certonly`

## command output ##
`OSError: [Errno 2] No such file or directory: '/usr/local/www/[removed]/public/.well-known/acme-challenge/[removed]'`

## traceback ##
`Traceback (most recent call last):
  File "/opt/eff.org/certbot/venv/bin/letsencrypt", line 11, in <module>
    sys.exit(main())
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/main.py", line 1379, in main
    return config.func(config, plugins)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/main.py", line 1262, in certonly
    lineage = _get_and_save_cert(le_client, config, domains, certname, lineage)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/main.py", line 120, in _get_and_save_cert
    lineage = le_client.obtain_and_enroll_certificate(domains, certname)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/client.py", line 406, in obtain_and_enroll_certificate
    cert, chain, key, _ = self.obtain_certificate(domains)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/client.py", line 349, in obtain_certificate
    orderr = self._get_order_and_authorizations(csr.data, self.config.allow_subset_of_names)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/client.py", line 385, in _get_order_and_authorizations
    authzr = self.auth_handler.handle_authorizations(orderr, best_effort)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/auth_handler.py", line 69, in handle_authorizations
    resps = self.auth.perform(achalls)
  File "/opt/eff.org/certbot/venv/lib/python2.7/site-packages/certbot/plugins/webroot.py", line 81, in perform
    self._create_challenge_dirs()
[...]

PR #7356, opened 2019-08-27
Fix issue #7165 in _create_challenge_dirs()

martin-c:
Some operating systems won't reliably raise EEXIST or EISDIR if a directory exists when filesystem.mkdir() is called, instead raising `Read-only file system` errors or similar even when path is in fact writable. To work around this we include a check to see if directory already exists. See Issue #7165 for more information.

## Pull Request Checklist

  the change being made.
- [ ] Add [mypy type


martin-c:
`webroot.py` has been moved to `certbot/_internal/plugins/webroot.py` and I'm not sure how to apply my changes to the original file there. I will make a new pull request.

PR #7359, opened 2019-08-28
Create Apache integration tests

adferrand:
As suggested in https://github.com/basak/certbot-snap-build/pull/10#issuecomment-525251461, and as I raised it few times to @joona and @bmw.

This PR introduces integration tests for Apache. They are integrated in certbot-ci, and follow the very same path than the integration tests for Nginx. This implementation takes a big care to run an isolated Apache instance during the tests: original configurations in `/etc/apache2` are not modified, allowing to run these tests not only in Travis, but also locally without particular isolation requirements.

These tests are mapped in the `integration` tox env.

For now only one test is done, but it is a meaningful one: it submits and installs a certificate through certbot using the Apache plugin authenticator + installer.


adferrand:
Current full tests suite is here: https://travis-ci.com/certbot/certbot/builds/124997111

I expect one last build to fail, `TOXENV=docker_dev`.

adferrand:
Ok, the full tests suite is running with success: https://travis-ci.com/certbot/certbot/builds/125090693

This PR is ready for review.

certbot-ci/certbot_integration_tests/apache_tests/apache_config.py
```diff
@@ -0,0 +1,171 @@
...
+import os
+
+import pkg_resources
+import getpass
+
+
+def construct_apache_config_dir(apache_root, http_port, https_port, key_path=None,
+                                cert_path=None, wtf_prefix='le'):
+    config_path = os.path.join(apache_root, 'config')
+    shutil.copytree('/etc/apache2', config_path, symlinks=True)
```
bmw: This makes the behavior of these tests reliant on the machine they are run on and things like the directories included in the main config file below are Debian specific. I think in an ideal world, we'd remove all this and make the Apache configuration independent of the machine it is run on like is done in the nginx integration tests, but I think doing so is hard/impossible because Apache binaries normally need things like shared object files to enable TLS.

Since this is not feasible and [configuring the Apache plugin to use an alternate directory is not fully supported from inside the plugin](https://github.com/certbot/certbot/issues/7385) and as a result requires a lot of extra complexity in the tests, I think we should run these tests in Docker where we have a consistent environment and just modify `/etc/apache2` as we wish.

Stepping back even further about how to test the Apache plugin though, I think one of the most complicated and necessary to test parts of the plugin is all the overrides on different OSes. I think it'd be nice if in the long term the Apache integration tests and the "test farm tests" that do this shared as much code as possible. I don't see why there needs to be any difference between the Apache integration tests run on one system and the tests run as part of the future "test farm" tests on the same system.

With all this in mind, I think it may make sense for us to block adding Apache integration tests on rewriting the test farm tests. Then as part of testing PRs or snaps, we could run the Apache "test farm tests" but on a smaller set of systems to speed things up. There are other ways we could go about reusing code between these two types of tests too, but this to me initially seems like it might be the simplest.

What do you think?

bmw:
In addition to my inline comments from 6 months ago which also touch on this, my notes from the last time we talked about this PR was that we were going to try to get them to run in Docker so they have a consistent environment rather than being relying on the host system.

Since this is all a fair bit of work that we probably won't get to in the short term, what do you think about closing this PR @adferrand? We could move it to a branch on the Certbot repo and make sure it's linked from an issue to pick up later.

adferrand:
Yes, I agree. I just pushed the branch to `apache-integration-tests` in Certbot repository.

PR #7369, opened 2019-09-07
Add Python3.8 tests

ohemorange:
Fixes #7333.

## Pull Request Checklist

- [ ] Edit the `master` section of `CHANGELOG.md` to include a description of
  the change being made.

PR #7372, opened 2019-09-10
Get integration tests working on python 3.8

### LINKED ISSUE #7333: Test for other Python 3.8 problems after #7106 is resolved
~~I think there are some such as the [use collections.Mapping here](https://github.com/certbot/certbot/blob/2911eda3bda95340b12d6deb096e7c15bdd5f44b/certbot/plugins/disco.py#L183) instead of `collections.abc.Mapping`.~~

There may not be any, but I had trouble getting all of our tests to pass as expected as described at https://github.com/certbot/certbot/issues/7333#issuecomment-525926864.

To run tests in Travis, you can probably make use of the [3.8-dev or nightly Python builds offered by Travis](https://docs.travis-ci.com/user/languages/python/#python-37-and-higher).

ohemorange:
Supersedes #7369.

Fixes #7333.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
   include:
     # Main test suite
     - python: "2.7"
-      env: ACME_SERVER=pebble TOXENV=clean,integration,integration-report
+      env: ACME_SERVER=pebble TOXENV=integration
       <<: *not-on-master
 
     # This job is always executed, including on master
@@ -122,12 +122,12 @@ matrix:
       env: TOXENV=py37 CERTBOT_NO_PIN=1
       <<: *extended-test-suite
     - python: "2.7"
-      env: ACME_SERVER=boulder-v1 TOXENV=clean,integration,integration-report
+      env: ACME_SERVER=boulder-v1 TOXENV=integration
       sudo: required
       services: docker
       <<: *extended-test-suite
     - python: "2.7"
-      env: ACME_SERVER=boulder-v2 TOXENV=clean,integration,integration-report
+      env: ACME_SERVER=boulder-v2 TOXENV=integration
       sudo: required
       services: docker
       <<: *extended-test-suite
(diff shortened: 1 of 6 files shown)


ohemorange:
*sigh* I solved it. So. Coverage had a bug when running reports on python 3.8. This was fixed in `coverage 4.5.4`. We had coverage pinned to `4.4.2`. `4.5` made a major change -- omissions need to be specified in both `[run]` and `[report]` sections of `.coveragerc` files.

By separating out the run and report, it so happened that we used the pinned `4.4.2` for running and the latest `4.5.4` for report, which happened to work.

To fix this, I updated the `.coveragerc` files, pinned coverage to `4.5.4`, and recombined back into a single `toxenv`. Need to do some final checks and changes, but pretty sure this will work.

.travis.yml
```diff
@@ -68,6 +68,10 @@ matrix:
       dist: xenial
       env: TOXENV=py37
       <<: *not-on-master
+    - python: "3.8-dev"
+      dist: xenial
```
adferrand: Xenial environment is the default one for Travis: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

So `dist: xenial` is not needed anymore.

.travis.yml
```diff
@@ -160,6 +164,10 @@ matrix:
       dist: xenial
       env: TOXENV=py37
       <<: *extended-test-suite
+    - python: "3.8-dev"
+      dist: xenial
```
adferrand: See above.

.travis.yml
```diff
@@ -202,6 +210,14 @@ matrix:
       sudo: required
       services: docker
       <<: *extended-test-suite
+    - python: "3.8-dev"
+      dist: xenial
```
adferrand: See above.

.travis.yml
```diff
@@ -202,6 +210,14 @@ matrix:
...
+      env: ACME_SERVER=boulder-v1 TOXENV=integration
+      <<: *extended-test-suite
+    - python: "3.8-dev"
+      dist: xenial
```
adferrand: See above.

tools/dev_constraints.txt
```diff
@@ -76,7 +76,7 @@ Sphinx==1.7.5
 sphinx-rtd-theme==0.2.4
 sphinxcontrib-websupport==1.0.1
 tldextract==2.2.0
-tox==2.9.1
+tox==3.14.0
```
adferrand: Why `tox` needs to be updated?

adferrand: Gotcha!

PR #7374, opened 2019-09-12
[Apache v2] DualParserNode implementation 1/3

joohoi:
This PR includes the first of the three PRs that are split from #7338 

The DualParserNode itself holds two ParserNode implementations and in every turn tries to make sure that they return the same values using various assertions.

### DIFF
--- a/certbot-apache/certbot_apache/assertions.py
+++ b/certbot-apache/certbot_apache/assertions.py
@@ -1,0 +1,70 @@
+"""Dual parser node assertions"""
+from certbot_apache import interfaces
+
+
+PASS = "CERTBOT_PASS_ASSERT"
+
+
+def assertEqual(first, second):
+    """ Equality assertion """
+
+    if isinstance(first, interfaces.CommentNode):
+        assertEqualComment(first, second)
+    elif isinstance(first, interfaces.DirectiveNode):
+        assertEqualDirective(first, second)
+
+    # Skip tests if filepath includes the pass value. This is done
+    # because filepath is variable of the base ParserNode interface, and
+    # unless the implementation is actually done, we cannot assume getting
+    # correct results from boolean assertion for dirty
+    if not isPass(first.filepath, second.filepath):
+        assert first.dirty == second.dirty
+        # We might want to disable this later if testing with two separate
+        # (but identical) directory structures.
+        assert first.filepath == second.filepath
+
+def assertEqualComment(first, second): # pragma: no cover
+    """ Equality assertion for CommentNode """
+
+    assert isinstance(first, interfaces.CommentNode)
+    assert isinstance(second, interfaces.CommentNode)
+
+    if not isPass(first.comment, second.comment):
+        assert first.comment == second.comment
+
+def _assertEqualDirectiveComponents(first, second): # pragma: no cover
+    """ Handles assertion for instance variables for DirectiveNode and BlockNode"""
+
+    # Enabled value cannot be asserted, because Augeas implementation
+    # is unable to figure that out.
+    # assert first.enabled == second.enabled
+    if not isPass(first.name, second.name):
+        assert first.name == second.name
+
+    if not isPass(first.parameters, second.parameters):
+        assert first.parameters == second.parameters
+
+def assertEqualDirective(first, second):
+    """ Equality assertion for DirectiveNode """
+
+    assert isinstance(first, interfaces.DirectiveNode)
+    assert isinstance(second, interfaces.DirectiveNode)
+    _assertEqualDirectiveComponents(first, second)
+
+def isPass(first, second): # pragma: no cover
+    """ Checks if either first or second holds the assertion pass value """
+
+    if isinstance(first, (tuple, list)):
+        if PASS in first:
+            return True
+    if isinstance(second, (tuple, list)):
+        if PASS in second:
+            return True
+    if PASS in [first, second]:
+        return True
+    return False
+
+def assertSimple(first, second):
+    """ Simple assertion """
+    if not isPass(first, second):
+        assert first == second
--- a/certbot-apache/certbot_apache/assertions.py
+++ b/certbot-apache/certbot_apache/assertions.py
@@ -64,7 +64,7 @@ def isPass(first, second): # pragma: no cover
         return True
     return False
 
-def assertSimple(first, second):
+def assertEqualSimple(first, second):
     """ Simple assertion """
     if not isPass(first, second):
         assert first == second
--- a/certbot-apache/certbot_apache/assertions.py
+++ b/certbot-apache/certbot_apache/assertions.py
@@ -53,12 +53,7 @@ def assertEqualDirective(first, second):
 
 def isPass(value): # pragma: no cover
     """Checks if the value is set to PASS"""
-    if isinstance(value, (tuple, list)):
-        if PASS in value:
-            return True
-    if PASS in value:
-        return True
-    return False
+    return PASS in value
 
 def assertEqualSimple(first, second):
     """ Simple assertion """
--- a/certbot-apache/certbot_apache/assertions.py
+++ b/certbot-apache/certbot_apache/assertions.py
@@ -17,7 +17,7 @@ def assertEqual(first, second):
     # because filepath is variable of the base ParserNode interface, and
     # unless the implementation is actually done, we cannot assume getting
     # correct results from boolean assertion for dirty
-    if not isPass(first.filepath, second.filepath):
+    if not isPass(first.filepath) and not isPass(second.filepath):
         assert first.dirty == second.dirty
         # We might want to disable this later if testing with two separate
         # (but identical) directory structures.
@@ -29,7 +29,7 @@ def assertEqualComment(first, second): # pragma: no cover
     assert isinstance(first, interfaces.CommentNode)
     assert isinstance(second, interfaces.CommentNode)
 
-    if not isPass(first.comment, second.comment):
+    if not isPass(first.comment) and not isPass(second.comment):
         assert first.comment == second.comment
 
 def _assertEqualDirectiveComponents(first, second): # pragma: no cover
@@ -38,10 +38,10 @@ def _assertEqualDirectiveComponents(first, second): # pragma: no cover
     # Enabled value cannot be asserted, because Augeas implementation
     # is unable to figure that out.
     # assert first.enabled == second.enabled
-    if not isPass(first.name, second.name):
+    if not isPass(first.name) and not isPass(second.name):
         assert first.name == second.name
 
-    if not isPass(first.parameters, second.parameters):
+    if not isPass(first.parameters) and not isPass(second.parameters):
         assert first.parameters == second.parameters
 
 def assertEqualDirective(first, second):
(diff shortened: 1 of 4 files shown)


certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -0,0 +1,69 @@
...
+            return (self.name == other.name and
+                    self.filepath == other.filepath and
+                    self.parameters == other.parameters and
+                    self.enabled == other.enabled and
+                    self.dirty == other.dirty and
+                    self.ancestor == other.ancestor and
+                    self.metadata == other.metadata)
+        return False
+
+    def set_parameters(self, parameters):
```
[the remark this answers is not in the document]
joohoi: These are any parameters for a directive or a block. There's an attempt to explain this behavior a bit [here](https://github.com/certbot/certbot/blob/apache-parser-v2/certbot-apache/certbot_apache/interfaces.py#L328-L347). The parameters are set in the `__init__`, but the direct modification of these values is prohibited (by using tuple) to help to handle (and hide where unnecessary) the whitespacing information.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,99 @@
...
+            used when creating new DualParser nodes using add_* methods.
+        :param CommentNode secondary: Secondary pre-created CommentNode
+        """
+
+        kwargs.setdefault("primary", None)
+        kwargs.setdefault("secondary", None)
+        primary = kwargs.pop("primary")
+        secondary = kwargs.pop("secondary")
+
+        if not primary:
```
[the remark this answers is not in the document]
joohoi: The plan is to have the DualParserNode implementation behave identically to the underlying interface. This allows us to handle the refactoring completely using DualParserNode implementation, and when later we're ready with the changes, it's trivial to substitute its use with the apacheconfig parser implementation.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,99 @@
...
+from certbot_apache import assertions
+from certbot_apache import augeasparser
+
+
+class DualNodeBase(object):
+    """ Dual parser interface for in development testing. This is used as the
+    base class for dual parser interface classes. This class handles runtime
+    attribute value assertions."""
+
+    def save(self, msg):  # pragma: no cover
```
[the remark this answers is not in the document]
joohoi: This would be good if there would be a bunch of these, but there just really is `save` and `set_parameters`. One of the following PR's will bring along also `delete_child` and `unsaved_files`, but these unfortunately require different test assertions, so it would not be feasible to generalize them all. The current way keeps the code a bit more readable.

certbot-apache/certbot_apache/assertions.py
```diff
@@ -0,0 +1,70 @@
...
+        assert first.parameters == second.parameters
+
+def assertEqualDirective(first, second):
+    """ Equality assertion for DirectiveNode """
+
+    assert isinstance(first, interfaces.DirectiveNode)
+    assert isinstance(second, interfaces.DirectiveNode)
+    _assertEqualDirectiveComponents(first, second)
+
+def isPass(first, second): # pragma: no cover
```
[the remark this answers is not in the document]
joohoi: 1) `PASS` here is used exactly as you read it. It's an placeholder value in unimplemented `ParserNode` variable or a method, that will pass the equality assertions and will start breaking after implementation. A sort of test driven development approach.
2) I think we should keep checking multiple values. This way we can catch errors early when gradually working on the actual interface implementations; we don't have to have a complete implementation before the assertions start pointing out that we did something wrong.

joohoi:
Oh I'm sorry about the missing context. This has been discussed throughout the last few months on various occasions, but I don't think we have a solid single document describing the expected behavior of DualParserNode interface. We can talk about this in more depth in Monday meeting, but I'll try to give a quick recap here.

We're implmeenting DualParserNode interface for three main reasons:
- To be able to refactor the ApacheConfigurator code to work with ParserNode interfaces in a way that's implementation agnostic.
- To be able to conduct runtime assertions of the results of DualParserNode implementations in order to assure that we're not missing anything when implementing the apacheconfig version of ParserNode, as compared to the Augeas implementation.
- To be able to work in parallel with both of the implementations - Augeas and apacheconfig. 

The assertions done in various steps in the implementation usage allow passing the said assertions if one (or both) of the values returned by the underlying implementations returns a pre-defined string that tells the DualParserNode implementation that we should continue execution regardless of (typically the secondary) underlying implementation not returning an identical result.

To keep the DualParserNode results intact throughout the execution, instead of direct results of underlying primary implementation, we should return the results from both implementations again wrapped into DualParserNode instances. This is however more relevant in the `find_*` methods PR (last one of the bunch).

The primary and secondary ParserNode implementation that DualParserNode wraps into itself can be changed around when needed, by changing the initialization code in three `__init__` methods of DualParserNode classes.

certbot-apache/certbot_apache/dualparser.py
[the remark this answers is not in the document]
ohemorange: We probably don't need to initialize missing primary/secondary nodes, since these objects will mostly not be directly created by the user.

certbot-apache/certbot_apache/augeasparser.py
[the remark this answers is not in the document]
ohemorange: (Note: when parameters are set, internal whitespace information is discarded.)

certbot-apache/certbot_apache/assertions.py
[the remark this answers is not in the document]
ohemorange: PASS also has a third meaning, around assertion-checking.

We decided to switch to `isPass(item)` and keep boolean logic external to the method.

certbot-apache/certbot_apache/dualparser.py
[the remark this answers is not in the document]
ohemorange: Based on discussion about the `find` methods, let's assert that both `primary` and `secondary` exist and remove some of this complexity.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -0,0 +1,69 @@
...
+    """ Augeas implementation of CommentNode interface """
+
+    def __init__(self, **kwargs):
+        comment, kwargs = util.commentnode_kwargs(kwargs)  # pylint: disable=unused-variable
+        super(AugeasCommentNode, self).__init__(**kwargs)
+        # self.comment = comment
+        self.comment = assertions.PASS
+
+    def __eq__(self, other): # pragma: no cover
+        if isinstance(other, self.__class__):
```
ohemorange: So presumably this equality check is written for the `assertEqualComment` above. In which case, we'll want to be able to check equality of comments that are implemented by different implementations of the ParserNode interface. Therefore it's probably ok for them to be equal even if `__class__` isn't equal. Also, it seems a little weird to be checking equality of parent instance variables here, especially when we have that whole thing about `filepath` above. So maybe we could call a `super` equality check, and implement `__eq__` in `AugeasParserNode`.
 
Actually thinking about this, it might make sense to have a `CommentNode` interface as well (though that may be overkill). But either way, if the equality check is going to fail in `assertEqualComment` anyway, what is this check for?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -0,0 +1,69 @@
...
+    """ Augeas implementation of DirectiveNode interface """
+
+    def __init__(self, **kwargs):
+        name, parameters, enabled, kwargs = util.directivenode_kwargs(kwargs)
+        super(AugeasDirectiveNode, self).__init__(**kwargs)
+        self.name = name
+        self.parameters = parameters
+        self.enabled = enabled
+
+    def __eq__(self, other): # pragma: no cover
```
ohemorange: Similar comment to the `CommentNode` equality check above.

certbot-apache/certbot_apache/augeasparser.py
joohoi: This is meant for checking the actual equality of two different instances of same implementation. In some cases we might need to omit comparing some details (first thing that comes to mind is whitespacing within metadata). So I think it should stay as it is.

ohemorange: In that case, would you mind explaining what is for?

https://github.com/certbot/certbot/blob/62fd2d5753785247b201560c83b31a502b581c30/certbot-apache/certbot_apache/assertions.py#L33

joohoi: We're expecting the two implementations to be different ones (the other being AugeasParserNode, while other one being ApacheConfigParserNode when it's created).

joohoi: We also want the actual equality assertion kick in only if the `PASS` string is not present in either of the instances.

ohemorange: > The current situation where both; primary and secondary are instances of the same implementation is only temporary placeholder.

So is the plan to change this eq method later, once we have `ApacheConfigParserNode` implemented?

certbot-apache/certbot_apache/tests/dualnode_test.py
```diff
@@ -0,0 +1,88 @@
...
+        self.directive.primary.set_parameters = pparams
+        self.directive.secondary.set_parameters = sparams
+        self.directive.set_parameters(("param", "seq"))
+        self.assertTrue(pparams.called)
+        self.assertTrue(sparams.called)
+
+    def test_getattr_equality(self):
+        self.directive.primary.variableexception = "value"
+        self.directive.secondary.variableexception = "not_value"
+        with self.assertRaises(AssertionError):
```
ohemorange: can simplify this to `self.assertRaises(AssertionError, self.directive.variableexception)`

certbot-apache/certbot_apache/augeasparser.py
joohoi: There shouldn't be a need for the change. The assertion in https://github.com/certbot/certbot/blob/62fd2d5753785247b201560c83b31a502b581c30/certbot-apache/certbot_apache/assertions.py#L33
Is comparing the `comment` property of the `AugeasCommentNode` instances instead of the complete instances for this exact reason.

certbot-apache/certbot_apache/tests/dualnode_test.py
joohoi: Unfortunately not. I think the `AssertionError` in this case is a special beast that gets caught in the internals of `unittest` if not within `with` context. This is what I came up with to mitigate the issue.

certbot-apache/certbot_apache/assertions.py
```diff
@@ -0,0 +1,66 @@
...
+    assert isinstance(first, interfaces.DirectiveNode)
+    assert isinstance(second, interfaces.DirectiveNode)
+    _assertEqualDirectiveComponents(first, second)
+
+def isPass(value): # pragma: no cover
+    """Checks if the value is set to PASS"""
+    if isinstance(value, (tuple, list)):
+        if PASS in value:
+            return True
+    if PASS in value:
```
ohemorange: you know, this was probably meant to be `if PASS == value`, but since it's a string this totally works anyway, in which case this whole method can just be `return PASS in value`

PR #7375, opened 2019-09-12
[Apache v2] DualParserNode implementation 2/3

joohoi:
This PR includes the second of the three PRs that are split from #7338 

The PR itself is based on the first PR for clarity, note that this needs to be changed, and this PR rebased after the part 1 is merged.

I had to add dummy methods for `find_*`, because some checks require the interface to be implemented.

### DIFF
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -1,0 +1,69 @@
+""" Augeas implementation of the ParserNode interface """
+from certbot_apache import assertions
+from certbot_apache import interfaces
+from certbot_apache import parsernode_util as util
+
+
+class AugeasParserNode(interfaces.ParserNode):
+    """ Augeas implementation of ParserNode interface """
+
+    def __init__(self, **kwargs):
+        ancestor, dirty, filepath, metadata = util.parsernode_kwargs(kwargs)  # pylint: disable=unused-variable
+        super(AugeasParserNode, self).__init__(**kwargs)
+        self.ancestor = ancestor
+        # self.filepath = filepath
+        self.filepath = assertions.PASS
+        self.dirty = dirty
+        self.metadata = metadata
+
+    def save(self, msg): # pragma: no cover
+        pass
+
+
+class AugeasCommentNode(AugeasParserNode):
+    """ Augeas implementation of CommentNode interface """
+
+    def __init__(self, **kwargs):
+        comment, kwargs = util.commentnode_kwargs(kwargs)  # pylint: disable=unused-variable
+        super(AugeasCommentNode, self).__init__(**kwargs)
+        # self.comment = comment
+        self.comment = assertions.PASS
+
+    def __eq__(self, other): # pragma: no cover
+        if isinstance(other, self.__class__):
+            return (self.comment == other.comment and
+                    self.filepath == other.filepath and
+                    self.dirty == other.dirty and
+                    self.ancestor == other.ancestor and
+                    self.metadata == other.metadata)
+        return False
+
+
+class AugeasDirectiveNode(AugeasParserNode):
+    """ Augeas implementation of DirectiveNode interface """
+
+    def __init__(self, **kwargs):
+        name, parameters, enabled, kwargs = util.directivenode_kwargs(kwargs)
+        super(AugeasDirectiveNode, self).__init__(**kwargs)
+        self.name = name
+        self.parameters = parameters
+        self.enabled = enabled
+
+    def __eq__(self, other): # pragma: no cover
+        if isinstance(other, self.__class__):
+            return (self.name == other.name and
+                    self.filepath == other.filepath and
+                    self.parameters == other.parameters and
+                    self.enabled == other.enabled and
+                    self.dirty == other.dirty and
+                    self.ancestor == other.ancestor and
+                    self.metadata == other.metadata)
+        return False
+
+    def set_parameters(self, parameters):
+        """Sets the parameters for DirectiveNode"""
+        self.parameters = parameters
+
+
+interfaces.CommentNode.register(AugeasCommentNode)
+interfaces.DirectiveNode.register(AugeasDirectiveNode)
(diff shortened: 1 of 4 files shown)


joohoi:
Need to have the base on `apache-parser-v2` for the tests to run, so changing it.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -0,0 +1,134 @@
...
+                    self.children == other.children and
+                    self.enabled == other.enabled and
+                    self.dirty == other.dirty and
+                    self.ancestor == other.ancestor and
+                    self.metadata == other.metadata)
+        return False
+
+    def add_child_block(self, name, parameters=None, position=None):  # pylint: disable=unused-argument
+        """Adds a new BlockNode to the sequence of children"""
+        new_block = AugeasBlockNode(name=assertions.PASS,
```
ohemorange: why isn't it `name=name` and `filepath=self.filepath`?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -0,0 +1,134 @@
...
+                                    ancestor=self,
+                                    filepath=assertions.PASS)
+        self.children += (new_block,)
```
ohemorange: Is there a reason to hold off on respecting the `position` argument?

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        pass
+
+    def delete_child(self, child):  # pragma: no cover
+        """Deletes a child from the ParserNode implementations. The actual
+        ParserNode implementations are used here directly in order to be able
+        to match a child to the list of children."""
+
+        self.primary.delete_child(child.primary)
+        self.secondary.delete_child(child.secondary)
```
ohemorange: Can we assert after we do this that the two children lists are still identical?

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,229 @@
...
+        assertions.assertEqual(primary_new, secondary_new)
+        new_comment = DualCommentNode(primary=primary_new, secondary=secondary_new)
+        return new_comment
+
+    def _create_matching_list(self, primary_list, secondary_list):  # pragma: no cover
+        """ Matches the list of primary_list to a list of secondary_list and
+        returns a list of tuples. This is used to create results for find_
+        methods. """
+
+        matched = list()
```
ohemorange: Would you mind describing the behavior we want here in words? What do `primary_list and `secondary_list` look like, and what should results look like given that?

certbot-apache/certbot_apache/dualparser.py
ohemorange: We can't, because we won't be able to build the parser node tree in memory in Augeas. In practice, we'll just be implementing a dynamic version that can't be iterated through, using `find` instead. This is because of Augeas resource usage.

certbot-apache/certbot_apache/dualparser.py
ohemorange: Joona to add comments explaining behavior.

We expect that we'll get two lists with the same contents, but they might be in different orders. This function reorders them in sets of tuples so they're the same.

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -93,3 +93,140 @@ def set_parameters(self, parameters):
...
+        implementations to a list of tuples.
+        """
+
+        matched = list()
+        for p in primary_list:
+            match = None
+            for s in secondary_list:
+                try:
+                    assertions.assertEqual(p, s)
+                    match = s
```
ohemorange: if we do find a match we should break, no reason to keep going through

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -93,3 +93,140 @@ def set_parameters(self, parameters):
...
+        in a same way. The function pairs the same search results from both
+        implementations to a list of tuples.
+        """
+
+        matched = list()
+        for p in primary_list:
+            match = None
+            for s in secondary_list:
+                try:
+                    assertions.assertEqual(p, s)
```
ohemorange: Ok, seeing this usage now, I wonder about the `assertEqual` implementation. If you have the same node in two places in the same file, what will happen? What do we want to happen? If it finds different instances of the same comment, for example, which will it return? Which do we want it to return? Basically, do we also need to be checking something about "position in the file" or something, or do we not need to worry about that?

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -93,3 +93,140 @@ def set_parameters(self, parameters):
...
+        instances that encapsulate the pairs of returned CommentNode objects.
+        """
+        pass
+
+    def delete_child(self, child):  # pragma: no cover
+        """Deletes a child from the ParserNode implementations. The actual
+        ParserNode implementations are used here directly in order to be able
+        to match a child to the list of children."""
+
+        self.primary.delete_child(child.primary)
```
ohemorange: Can this fail in one or the other of the implementations? Actually, this question probably applies to the `add_*` methods as well. Do we need to be raising and checking exceptions here?

joohoi: The interface doesn't currently define exceptions, so this will depend on the implementations themelves, and doing `except Exception` would in my opinion be overkill here and could hide some actual underlying issues. The execptions will definitely need to be handled here too when implementing them to the actual interface though.

certbot-apache/certbot_apache/dualparser.py
ohemorange: If we think that we will be able to distinguish them by deducing position in the block (or by some other characteristic), I agree it can wait.

joohoi: Yup, there will be some added complexity to this due to the planned nature of the AugeasParserNode interface (doing things dynamically instead of prepopulating the ParserNode tree). But if the need arises, I think we can handle this later on.

certbot-apache/certbot_apache/dualparser.py
ohemorange: ok, I'm fine with leaving it out for now and then adding it as necessary later

PR #7376, opened 2019-09-13
[Apache v2] DualParserNode implementation 3/3

joohoi:
This PR includes the last of the three PRs that are split from #7338

This PR should be rebased on apache-parser-v2 when first two are merged.

While the lines of code in this PR is quite high, most of it is basically somewhat lengthy tests that have a lot of duplicated code.

### DIFF
--- a/certbot-apache/certbot_apache/assertions.py
+++ b/certbot-apache/certbot_apache/assertions.py
@@ -10,12 +10,13 @@ def assertEqual(first, second):
 
     if isinstance(first, interfaces.CommentNode):
         assertEqualComment(first, second)
-    elif isinstance(first, interfaces.BlockNode):
-        # this check needs to be first, as BlockNodes are also instances of
-        # DirectiveNode
-        assertEqualBlock(first, second)
     elif isinstance(first, interfaces.DirectiveNode):
         assertEqualDirective(first, second)
+
+    # Do an extra interface implementation assertion, as the contents were
+    # already checked for BlockNode in the assertEqualDirective
+    if isinstance(first, interfaces.BlockNode):
+        assert isinstance(second, interfaces.BlockNode)
 
     # Skip tests if filepath includes the pass value. This is done
     # because filepath is variable of the base ParserNode interface, and
@@ -55,17 +56,6 @@ def assertEqualDirective(first, second):
     assert isinstance(second, interfaces.DirectiveNode)
     _assertEqualDirectiveComponents(first, second)
 
-def assertEqualBlock(first, second):
-    """ Equality assertion for BlockNode """
-
-    # first was checked in the assertEqual method
-    assert isinstance(first, interfaces.BlockNode)
-    assert isinstance(second, interfaces.BlockNode)
-    _assertEqualDirectiveComponents(first, second)
-    # Children cannot be asserted, because Augeas implementation will not
-    # prepopulate the sequence of children.
-    # assert len(first.children) == len(second.children)
-
 def isPass(value): # pragma: no cover
     """Checks if the value is set to PASS"""
     return PASS in value
@@ -73,20 +63,20 @@ def isPass(value): # pragma: no cover
 def isPassDirective(block):
     """ Checks if BlockNode or DirectiveNode should pass the assertion """
 
-    if block.name == PASS:
+    if isPass(block.name):
         return True
-    if PASS in block.parameters: # pragma: no cover
+    if isPass(block.parameters): # pragma: no cover
         return True
-    if block.filepath == PASS: # pragma: no cover
+    if isPass(block.filepath): # pragma: no cover
         return True
     return False
 
 def isPassComment(comment):
     """ Checks if CommentNode should pass the assertion """
 
-    if comment.comment == PASS:
+    if isPass(comment.comment):
         return True
-    if comment.filepath == PASS: # pragma: no cover
+    if isPass(comment.filepath): # pragma: no cover
         return True
     return False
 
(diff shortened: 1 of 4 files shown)


certbot-apache/certbot_apache/dualparser.py
```diff
@@ -0,0 +1,320 @@
...
+                matched.append((p, match))
+            else:
+                raise AssertionError("Could not find a matching node.")
+        return matched
+
+    def find_blocks(self, name, exclude=True):
+        """
+        Performs a search for BlockNodes using both implementations and does simple
+        checks for results. This is built upon the assumption that unimplemented
+        find_* methods return a list with a single assertion passing object.
```
ohemorange: why are we returning lists in the first place if the assumption is that it'll return a single object?

certbot-apache/certbot_apache/assertions.py
```diff
@@ -0,0 +1,116 @@
...
+
+    try:
+        node = nodelist[0]
+    except IndexError:
+        node = None
+
+    if not node: # pragma: no cover
+        return False
+
+    if isinstance(node, interfaces.BlockNode):
```
ohemorange: since a `BlockNode` is an instance of a `DirectiveNode`, this is unnecessary.

certbot-apache/certbot_apache/dualparser.py
joohoi: List with only a single object applies to unimplemented `find_*` functions. When they are actually implemented, they should return [lists of ParserNode objects](https://github.com/certbot/certbot/blob/apache-parser-v2/certbot-apache/certbot_apache/interfaces.py#L451).

certbot-apache/certbot_apache/assertions.py
joohoi: Nice catch! This is a remnant from earlier state where there was different checks on `BlockNode`. It's now removed.

certbot-apache/certbot_apache/dualparser.py
ohemorange: Oh, I see what you mean by "unimplemented find_* methods" now. Either in the main implementation or in testing, we should make sure to test with lists being returned.

certbot-apache/certbot_apache/assertions.py
```diff
@@ -10,6 +10,10 @@ def assertEqual(first, second):
 
     if isinstance(first, interfaces.CommentNode):
         assertEqualComment(first, second)
+    elif isinstance(first, interfaces.BlockNode):
+        # this check needs to be first, as BlockNodes are also instances of
+        # DirectiveNode
+        assertEqualBlock(first, second)
     elif isinstance(first, interfaces.DirectiveNode):
         assertEqualDirective(first, second)
```
ohemorange: do we want it to talk check `assertEqualDirective` for `BlockNodes` then?

certbot-apache/certbot_apache/assertions.py
```diff
@@ -51,10 +55,59 @@ def assertEqualDirective(first, second):
...
     _assertEqualDirectiveComponents(first, second)
 
+def assertEqualBlock(first, second):
+    """ Equality assertion for BlockNode """
+
+    # first was checked in the assertEqual method
+    assert isinstance(first, interfaces.BlockNode)
+    assert isinstance(second, interfaces.BlockNode)
+    _assertEqualDirectiveComponents(first, second)
+    # Children cannot be asserted, because Augeas implementation will not
```
ohemorange: if we're not asserting about children, this is basically the same check as `assertEqualDirective`; is there any benefit to having this around as well?

certbot-apache/certbot_apache/assertions.py
```diff
@@ -51,10 +55,59 @@ def assertEqualDirective(first, second):
...
+    # assert len(first.children) == len(second.children)
+
 def isPass(value): # pragma: no cover
     """Checks if the value is set to PASS"""
     return PASS in value
 
+def isPassDirective(block):
+    """ Checks if BlockNode or DirectiveNode should pass the assertion """
+
+    if block.name == PASS:
```
ohemorange: We have `isPass`, might as well use it for these checks (ie `isPass(block.name)`, `isPass(block.parameters)`).

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -185,37 +186,124 @@ def _create_matching_list(self, primary_list, secondary_list):  # pragma: no cov
...
         Performs a search for BlockNodes using both implementations and does simple
         checks for results. This is built upon the assumption that unimplemented
         find_* methods return a list with a single assertion passing object.
         After the assertion, it creates a list of newly created DualBlockNode
         instances that encapsulate the pairs of returned BlockNode objects.
         """
-        pass
 
-    def find_directives(self, name, exclude=True):  # pragma: no cover
+        primary_blocks = self.primary.find_blocks(name, exclude)
```
ohemorange: It feels kind of silly to have this much code copy-pasted; how would you feel about creating a helper function that takes the appropriate `find_*` method and the appropriate `Dual*Node` class?

certbot-apache/certbot_apache/assertions.py
joohoi: Yeah, the `assertEqualBlock` had just a single different step; asserting the `BlockNode` interface implementation. I moved this single step to the `assertEqual` itself, and removed the `assertEqualBlock` and the accompanying conditional from here.

certbot-apache/certbot_apache/assertions.py
joohoi: Only the actual interface implementation assertion, it's now moved to `assertEqual`

certbot-apache/certbot_apache/dualparser.py
joohoi: This is a good idea, and now implemented.

PR #7377, opened 2019-09-13
Create a new CI for Certbot on Windows using Azure Pipelines

adferrand:
This PR defines pipelines that can be run on Azure Pipelines. Currently there are two:
* `.azure-pipelines/main.yml` is the main one, executed on PRs for master, and pushes to master,
* `.azure-pipelines/advanced.yml` add installer testing on top of the main pipeline, and is executed for `test-*` branches, release branches, and nightly run for master.

These two pipelines covers all existing stuff done by AppVeyor currently, and so AppVeyor can be decommissioned once Azure Pipelines is operational.

You can see working pipeline in my fork:
* a PR for `master` (so using main pipeline): https://github.com/adferrand/certbot/pull/65
* a PR for `test-something` (so using advanced pipeline): https://github.com/adferrand/certbot/pull/66
* uploaded coverage from Azure Pipelines: https://codecov.io/gh/adferrand/certbot/commit/499aa2cbf25e1e0ab4c93ab64057db92dfec0fba/build

Once this PR is merged, we need to enable Azure Pipelines for Certbot. Instructions are written in `azure-pipelines/INSTALL.md`. This document also references all access rights required to Azure Pipelines onto GitHub to make the CI process work.

Future work for future PRs:
* create a CD pipeline for the releases that will push the installer to GitHub releases
* implement a solution to generate notification on IRC or Mattermost when a nightly build fails


.azure-pipelines/INSTALL.md
```diff
@@ -0,0 +1,112 @@
...
+- Click "Pipelines" tab
+- Click "New pipeline"
+- Where is your code? Select "GitHub"
+
+```
+!!! ACCESS !!!
+Here comes the big requests:
+- Admin access to webhooks and services (not the code)
+- RO access on personal data (email + profile)
+- RO + RW access to repositories (code, issue, PR, wiki, settings, webhooks, services, deploy keys, collaboration invites)
```
bmw: Oh wow. So it basically needs full access to everything? I don't think there's much missing from this list.

Maybe we could give them this access, but given how much time we spent avoiding doing something similar for Docker Hub, I'm not immediately sure we want to do that.

adferrand: This is not the same situation than DockerHub.

First, we are talking here about a CI/CD tool that works on the GitHub project, not a tool working from the GitHub project like DockerHub. These access are quite relevant since its purpose is to proceed actions on the project. In fact, access for Travis CI and for Azure Pipelines are quite similar:
* Travis CI: `Read and write access to checks, commit statuses, deployments, and repository hooks`
* Azure Pipelines: `Read and write access to checks, commit statuses, deployments, issues, and pull requests`

Second, this is not a full access to everything. DockerHub was asking from a full access on the organization itself, but Azure Pipelines (or other CIs) asks for a write access on the repositories only. And this is only a user level access, not an admin one. Controls done on GitHub for protected branches, or merging policies, are out of its reach.

Third, and this is valid for all systems that uses the GitHub Applications: this is a satisfyingly secured approach to delegate authentication. We are not talking about long-live tokens giving full-access to the APIs, but short-live tokens generated dynamically during the authentication process. See https://developer.github.com/apps/migrating-oauth-apps-to-github-apps/

Fourth, versus the risk of migrating, we need to consider the advantages to move to Azure Pipelines. It is not only about gaining a lot of parallel executors, or efficient and stable VMs both for Windows and Linux. It is also to build a full CD solution with the appropriate solution. Modern CI/CD are pipelines: Travis and AppVeyor do not have this, while CircleCI, GitHub Actions and Azure Pipelines are designed around it. In the long term having dedicated pipelines is the best choice to do in my opinion.

Otherwise, the hacks like in `before_install`, `branches.only` or `not-on-master/extended-test-suite` we find in our Travis config will become more and more present as we go deeper in a CD approach (first with Windows installer, after certainly with Snaps or other deployed artifacts).

bmw: > Azure Pipelines: Read and write access to checks, commit statuses, deployments, issues, and pull requests

This is not the line I'm concerned about or commented on.

> Second, this is not a full access to everything. DockerHub was asking from a full access on the organization itself, but Azure Pipelines (or other CIs) asks for a write access on the repositories only. And this is only a user level access, not an admin one. Controls done on GitHub for protected branches, or merging policies, are out of its reach.

I think there may have been a miscommunication here. My understanding of the section of this document I commented on talks about admin access for webhooks and services and write access on things like your GitHub settings.

adferrand: After discussions on 1-1, we agree on the fact that Travis vs Azure Pipelines on actual permissions, when compared from the GitHub settings looks very similar.

@bmw to install this PR on a fork and verify this similarity. We need to confirm in particular what are the `Read write access on GitHub settings` assertion that is encountered in popups to setup Azure Pipelines, since it does not seem to be related to anything in the GitHub settings.

If it is confirmed that Azure Pipelines does not get significant more permissions than Travis, it will unblock this PR merging.

bmw: I just finished playing with this a bit.

When following the instructions here (or the most intuitive flow on Azure DevOps), Azure tries to set up an Oauth app in addition to the GitHub app we initially installed. This Oauth app is where the concerning permissions about write access to all aspects of your repository come in.

I tried to look into what these repository permissions allows Azure to do but couldn't find much. I think they correspond to the repository scope described [here](https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/) but I couldn't find anything about what exactly that entails.

(As a slightly unrelated aside, I looked into the permissions that Travis' Oauth app requires and it is:
```
Organizations and teams
Read-only access
This application will be able to read your organization, team membership, and private project boards.

Commit statuses
Read and write access
This application will be able to read and write commit statuses (no direct code access).

Deployments
Manage deployments and deployment status
This application will be able to manage deployments and deployment status (no direct code access).

Personal user data
Email addresses (read-only)
This application will be able to read your private email addresses.

Repository webhooks and services
Read and write access
This application will be able to read and modify repository webhooks and services (no direct code access).
```
which is much more limited.)

I'm still personally a bit hesitant to grant the permissions on this Oauth app at least without a better understanding of what it can do, but there may be good news. The need to install this second application seems to me like a bug. [This page](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#grant-access-to-your-github-repositories) talks about how the GitHub application should be preferred over the Oauth app for setting up CI. This Oauth app also isn't referenced at all at https://www.azuredevopslabs.com/labs/azuredevops/github-integration/. When you initially create the pipeline following the instructions there, you don't need the Oauth app.

I'm not the only one that thinks this is a bug either, however, it appears Microsoft disagrees as they closed https://developercommunity.visualstudio.com/content/problem/642369/pipelines-creation-falls-back-to-github-oauth-auth.html as "Not a bug". Despite that, they describe a way to work around the problem. You can click the button circled here:

Screen Shot 2019-09-24 at 4 56 19 PM

This button says it is to create a pipeline without a YAML, but it lets you use your GitHub app connection to link to GitHub and then the first option is to use a YAML. I tested this a bit on my fork and this seems to work.

Do you think these permissions are sufficient for us now and in the future or do you think we will also need/want to set up the Oauth app? I'll flag for you that https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#grant-access-to-your-github-repositories which I linked above says:

> The Azure Pipelines GitHub App is the recommended authentication type for continuous integration pipelines. For release pipelines that are triggered by changes to a GitHub repository, use OAuth or personal access token authentication.

adferrand: Despite the advice in the doc you quoted, I think it is indeed totally doable with this approach. I reconfigured my personal fork, and it is working great in term of CI.

In term of CD, the only question I had was how github releases would be handled in this case. The [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops) states that you need a "Service connection". Following the path for this, it appears that you can typically choose 3 options here:
- use the pre-existing connection that corresponds to the GitHub App installed by Azure. This has been configured at the beginning of connection between Azure Pipelines and GitHub. It has the reasonable set of permissions I wrote at the beginning of the `INSTALL.md` file, section "Installing Azure Pipelines to GitHub".
- create a connection using the "Grant access" option. In this case, you have again an OAuth pop-up permission, like when creating a pipeline in non-classic mode, with the "give me all the rights" thing
- create a connection using the "Personal token access". In this case, recommended (as shown in the UI when trying this option) rights associated to the token to use are: `repo, read:user, user:email, admin:repo_hook`. This is also very reasonable (and really similar to Travis here), with admin control _only_ on the repo hooks. Understandable, since you need to setup several hooks around GitHub releases.

adferrand: Update here: the pre-existing connection cannot be used as a Service connection to push GitHub releases. Azure DevOps complains that it needs to be a Service connection created from a token or OAuth.

So I go for the third option in my previous comment.

bmw: A personal access token (which should probably be set up using a bot account rather than own of our own GitHub accounts) or just pushing the release outside of Azure (especially if we're going to be signing it locally anyway) seems workable to me.

I'll keep reviewing this PR.

adferrand:
@bmw I added the last modifications I wanted to do for this PR. I start to answer your questions.

.azure-pipelines/INSTALL.md
adferrand: Alright. So as discussed, we will continue to work on the continuous delivery using the token approach in the next PR, then properly design the final CD approach we want.

.azure-pipelines/INSTALL.md
```diff
@@ -0,0 +1,112 @@
+# Configuring Azure Pipelines with Certbot
+
+Let's begin. All pipelines are defined in `.azure-pipelines`. Currently there are two:
+* `.azure-pipelines/main.yml` is the main one, executed on PRs for master, and pushes to master,
+* `.azure-pipelines/advanced.yml` add installer testing on top of the main pipeline, and is executed for `test-*` branches, release branches, and nightly run for master.
+
+Several templates are defined in `.azure-pipelines/templates`. These YAML files aggregate common jobs configuration that can be reused in several pipelines.
+
+Note that `CODECOV_TOKEN` secured environment variable needs to be set to allow the main pipeline to publish coverage reports to CodeCov.
```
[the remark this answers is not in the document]
adferrand: Yes, it is because of the partial support of Azure Pipeline on Codecov.

In fact there is three level of supports for a CI:
1) flawless support: in this case, you do not need to provide `CODECOV_TOKEN`, and reports integrates flawlessly in Codecov
2) partial support: the reports integrates flawlessly in Codecov if you provide the token through `CODECOV_TOKEN`
3) no support: you can still push coverage reports if you provide the token through `CODECOV_TOKEN`, but because of missing optional other parameters consumed from the CI environment, reports are not integrated nicely

At this point, `python-coverage` is in situation 3), and `coverage-bash` is in situation 2).

That is why I moved to `coverage-bash`. I update the comment here.

.azure-pipelines/INSTALL.md
```diff
@@ -0,0 +1,112 @@
...
+- On GitHub, go to Marketplace
+- Select Azure Pipeline, and "Set up a plan"
+- Select Free, then "Install it for free"
+- Click "Complete order and begin installation"
+
+```
+!!! ACCESS !!!
+Azure Pipeline needs RW on code, RO on metadata, RW on checks, commit statuses, deployements, issues, pull requests.
+Access can be defined for all or only selected repositories, which is nice.
+```
```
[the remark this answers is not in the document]
adferrand: Done.

.azure-pipelines/INSTALL.md
```diff
@@ -0,0 +1,112 @@
...
+- The Visibility is public, to profit from 10 parallel jobs
+
+```
+!!! ACCESS !!!
+Azure Pipelines needs access to the GitHub account (in term of beeing able to check it is valid), and the Resources shared between the GitHub account and Azure Pipelines.
+```
+
+_Done. We can move to pipelines configuration._
+
+## Import an existing pipelines from `.azure-pipelines` folder
```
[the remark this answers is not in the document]
adferrand: Done.

.azure-pipelines/avanced.yml
```diff
@@ -0,0 +1,18 @@
+# Advanced pipeline for isolated checks and release purpose
```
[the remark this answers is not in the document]
adferrand: Fixed.

.azure-pipelines/INSTALL.md
```diff
@@ -0,0 +1,112 @@
...
+
+- (Bonus) Go again to Pipeline, select your pipline, button "..." and choose "Rename/Move": give to the pipeline a nice name!
+
+
+## Add a secret variable to a pipeline (like `CODECOV_TOKEN`)
+
+- On Azure DevOps, go to you organization, project, pipeline tab
+- Select the pipeline, click "Edit" button, then click "Variables" button
+- Set name (eg `codecov_token`), value, tick "Kep this value secret"
+- In YAML, use something like to consume the secret as an enviroment variable
```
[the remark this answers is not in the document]
adferrand: Done.

.azure-pipelines/templates/installer-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+          path: $(Build.ArtifactStagingDirectory)
+          artifact: WindowsInstaller
+      - script: python windows-installer/install.py
+        displayName: Install Certbot
+      - script: |
+          python -m venv venv
+          venv\Scripts\python tools\pip_install.py -e certbot-ci
+        displayName: Prepare Certbot-CI
+      - script: |
+          set PATH=%ProgramFiles(x86)%\Certbot\bin;%PATH%
```
bmw: Do you know why we need to do this? The installer should be putting `certbot` in the default `PATH`. It'd be nice if we were able to rely on this behavior and help test it here.

adferrand: It fails without it. Yes normally the installer adds the relevant entries, and it works on all real systems I tested, except in CI. For your information, the tests I made on Travis end up with the same issue.

There is something weird with the env set by CIs, so far I did not find the underlying reason.

.azure-pipelines/advanced.yml
```diff
@@ -0,0 +1,18 @@
+# Advanced pipeline for isolated checks and release purpose
+trigger:
+  - test-*
+  - '*.x'
+pr:
+  - test-*
+  - '*.x'
+# This pipeline is also nightly run on master
+schedules:
+  - cron: "0 0 * * *"
```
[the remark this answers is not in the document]
adferrand: Done.

windows-installer/install.py
```diff
@@ -0,0 +1,33 @@
...
+
+def main():
+    script_dir_path = os.path.dirname(os.path.realpath(__file__))
+    repo_path = os.path.dirname(script_dir_path)
+    certbot_version = subprocess.check_output([sys.executable, '-c', 'import certbot; print(certbot.__version__)'],
+                                              universal_newlines=True, cwd=repo_path).strip()
+    installer_path = os.path.join(script_dir_path, 'build', 'nsis', 'certbot-{0}-installer-{1}.exe'
+                                  .format(certbot_version, 'win_amd64' if PYTHON_BITNESS == 64 else 'win32'))
+
+    return_code = subprocess.check_call([installer_path, '/S', '/AllUsers'])
```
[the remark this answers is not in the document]
adferrand: No it is not needed. I created this PR before #7371 has been merged.

windows-installer/construct.py
```diff
@@ -70,13 +70,39 @@ def _copy_assets(build_path, repo_path):
...
+
+# Preload pywintypes and pythoncom
+pwt = os.path.join(pkgdir, 'pywin32_system32', 'pywintypes{0}{1}.dll')
+pcom = os.path.join(pkgdir, 'pywin32_system32', 'pythoncom{0}{1}.dll')
+import warnings
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    import imp
+imp.load_dynamic('pywintypes', pwt)
+imp.load_dynamic('pythoncom', pcom)
```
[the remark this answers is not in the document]
adferrand: Quite hard. The only example in `importlib` documentation (https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) is not working, with a `spec` object that is `None`. And reproducing the internal behavior of `imp.load_dynamic` (https://github.com/python/cpython/blob/c02b41b1fb115c87693530ea6a480b2e15460424/Lib/imp.py#L330) will add a lot of code that would require also to reimplement internal methods from `importlib`.

.azure-pipelines/templates/tests-suite.yml
```diff
@@ -0,0 +1,33 @@
...
+    - task: UsePythonVersion@0
+      inputs:
+        versionSpec: $(PYTHON_VERSION)
+        addToPath: true
+    - script: python tools/pip_install.py -U tox coverage
+      displayName: Install dependencies
+    - script: python -m tox
+      displayName: Run tox
+    - bash: |
+        curl -s https://codecov.io/bash -o codecov-bash || echo "Failed to download codecov-bash"
```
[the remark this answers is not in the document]
adferrand: Done.

windows-installer/install.py
```diff
@@ -0,0 +1,33 @@
...
+PYTHON_BITNESS = 32
+
+
+def main():
+    script_dir_path = os.path.dirname(os.path.realpath(__file__))
+    repo_path = os.path.dirname(script_dir_path)
+    certbot_version = subprocess.check_output([sys.executable, '-c', 'import certbot; print(certbot.__version__)'],
+                                              universal_newlines=True, cwd=repo_path).strip()
+    installer_path = os.path.join(script_dir_path, 'build', 'nsis', 'certbot-{0}-installer-{1}.exe'
+                                  .format(certbot_version, 'win_amd64' if PYTHON_BITNESS == 64 else 'win32'))
```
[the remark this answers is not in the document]
adferrand: Current implementation is the best deal I found. Your two propositions implies higher hidden complexity in other places as a consequence.

For b, we loose the signature of Certbot version and architecture in the installer filename, which is important user side when you download a Windows installer to install a program. Reinstate it for the release pipeline will add a lot of complexity in the pipeline itself, so in Yaml, that is harder to maintain than a python script. 

For a, grouping logic needs to respect two constraints: being callable as a script for tox, and as a module for the installer script. Importing a module in a script is a brittle process if the script and the module are not in a fully fledged python project (for instance relative imports do not work from a python script). So it would require to add the usual setuptools configuration around the `windows-installer` project, which is overkill in my opinion and add a lot of complexity again.

Here, if because of the duplicated logic we break the build, it will break the associated pipelines run nightly, and we will know it. Anyway, I do not expect the installer name pattern to change regularly.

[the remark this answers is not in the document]
adferrand: I thought about this during the week end. I have to admit that having a windows installer file name changing for each version implies also, on top of current work, an additional complexity for the CD process. This makes the option b) more appealing in the pro/con balance.

I will go for this option b). I keep however the platform (`win32`/`win_amd64`) since using a `win_amd64` on a pure Windows 32bits would fail with obscure error codes.

It is not ideal, but the installer version is still recognizable in the GitHub release API since the version is contained in the full URL of the artifact binary. Also, version is recognizable during installation as it is displayed by the UI throughout the install process, and properly declared in Windows, so also visible in the Windows Uninstaller UI.

In the future, I think the ideal would be to have an installer not tied to a particular version of Certbot, but that dynamically install the latest version of Certbot, like for `certbot-auto`, and like a lot of program that I use and Windows, and also propose online installer/updater programs.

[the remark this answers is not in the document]
bmw: Thanks for simplifying this a bit. There are a number of tradeoffs here.

I think something like what you have now is probably fine, but just for the sake of exploring our options, I don't think we'd have to make `windows-installer` a package to import a Python script in the same directory. We do this without issue in the `tools/venv*.py` scripts when we import `_venv_common`.

.azure-pipelines/templates/installer-tests.yml
bmw: I couldn't find a better way either. I suspect it's something like the pipeline is being executed in one continuous environment that's not reloaded to pick up changes like this.

I searched around for a way to explicitly reset things like environment variables to their default value on Windows and couldn't find anything.

adferrand: I tried dedicated Powershell and Batch dedicated utility provided by choco for that, https://chocolatey.org/docs/helpers-update-session-environment, and also another explicit command to feed PATH from the Registry that I do not find again at the moment.

But all without success.

bmw: Weird...

Thanks for trying all of that.

windows-installer/construct.py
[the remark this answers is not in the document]
bmw: I don't love this either, but I think it's probably fine for now. We're following the example in the pynsist repo.

`pywin32` has Python files for loading the [pywintypes](https://github.com/mhammond/pywin32/blob/master/win32/Lib/pywintypes.py) and [pythoncom](https://github.com/mhammond/pywin32/blob/master/com/pythoncom.py) dlls which would probably be much easier for us to use in the future if we want to clean this up.

windows-installer/construct.py
```diff
@@ -70,13 +70,39 @@ def _copy_assets(build_path, repo_path):
...
+    # referred in the pynsist configuration as an "extra_preamble".
+    # Reference example: https://github.com/takluyver/pynsist/tree/master/examples/pywebview
+    with open(pywin32_paths_file, 'w') as file_h:
+        file_h.write('''\
+pkgdir = os.path.join(os.path.dirname(installdir), 'pkgs')
+
+sys.path.extend([
+    os.path.join(pkgdir, 'win32'),
+    os.path.join(pkgdir, 'win32', 'lib'),
+])
```
bmw: What made you remove `Pythonwin` here which is present at https://github.com/takluyver/pynsist/blob/master/examples/pywebview/pywin32_paths.py#L4

adferrand: Because it is related to Windows UI utilities, and we do not use that since Certbot is a pure CLI project. All the integrations tests are passing without it, by the way.

bmw: Thanks for the explanation. Just trying to understand the change.

As an aside, did you know that the first 10 or so Certbot releases had and defaulted to a TUI: https://www.youtube.com/watch?v=Ya6t8nMclos

PR #7379, opened 2019-09-17
Specify xenial for broken py3.7 integration tests

ohemorange:
Looks like travis' rollout isn't complete, so let's get tests passing in the meantime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -198,11 +198,13 @@ matrix:
       env: ACME_SERVER=boulder-v1 TOXENV=integration
       sudo: required
       services: docker
+      dist: xenial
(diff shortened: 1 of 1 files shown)


ohemorange:
Full test suite https://travis-ci.com/certbot/certbot/builds/128083086

bmw:
I personally think this is the correct, minimal fix for our test failures, but I went further in https://github.com/certbot/certbot/pull/7380 and locked everything down preventing problems such as our oldest tests failing when the default finally decides to update to Xenial. Because that work is done and tested, I'm closing this PR in favor of that one but if you disagree, I'm happy to chat!

PR #7380, opened 2019-09-17
Use xenial globally

bmw:
As described at https://github.com/certbot/certbot/pull/7372#discussion_r323592366, Travis is transitioning people to Xenial, but it seems this transition still may not be complete as some of our jobs ran on Trusty with all references to `dist` removed as seen at https://travis-ci.com/certbot/certbot/builds/127960999.

This PR sets `dist: xenial` globally and overrides it as needed for the oldest tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 
 cache:
     directories:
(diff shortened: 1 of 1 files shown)


bmw:
You can see the full test suite passing with these changes at https://travis-ci.com/certbot/certbot/builds/128076379.

PR #7387, opened 2019-09-20
Don't send OCSP requests for expired certificates

### LINKED ISSUE #7152: Don't send OCSP requests for expired certificates

ohemorange:
Fixes #7152.

### DIFF
--- a/certbot/tests/ocsp_test.py
+++ b/certbot/tests/ocsp_test.py
@@ -72,13 +72,12 @@ class OCSPTestOpenSSL(unittest.TestCase):
     @mock.patch('certbot.ocsp._determine_ocsp_server')
     @mock.patch('certbot.util.run_script')
     def test_ocsp_revoked(self, mock_run, mock_determine):
-        import datetime
         import pytz
-        now = pytz.UTC.fromutc(datetime.datetime.utcnow())
+        now = pytz.UTC.fromutc(datetime.utcnow())
         cert_obj = mock.MagicMock()
         cert_obj.cert = "x"
         cert_obj.chain = "y"
-        cert_obj.target_expiry = now + datetime.timedelta(hours=2)
+        cert_obj.target_expiry = now + timedelta(hours=2)
 
         self.checker.broken = True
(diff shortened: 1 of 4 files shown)


bmw:
Looks like lint is angry and appveyor thought a more recent build of this PR started so it canceled this one.

If you continue to have issues with appveyor, let me know if you want me to take a look.

certbot/tests/ocsp_test.py
```diff
@@ -72,21 +72,35 @@ def test_init(self, mock_exists, mock_popen, mock_log):
     @mock.patch('certbot.ocsp._determine_ocsp_server')
     @mock.patch('certbot.util.run_script')
     def test_ocsp_revoked(self, mock_run, mock_determine):
+        import pytz
```
bmw: nit: I think we should just import this once at the top level of the file rather than in the various methods that need it.

PR #7388, opened 2019-09-21
acme: socket timeout for HTTP standalone servers

### LINKED ISSUE #7386: Standalone server waits forever to drain clients during shutdown
Originally discovered in [a help thread on the forums](https://community.letsencrypt.org/t/certbot-hangs-on-stopping-server/102450?u=_az).

Ubuntu 19.04

certbot-auto

    sudo certbot-auto certonly -d certbot.eff.org --standalone --dry-run

and simultaneously in a second terminal:

    nc -vvv localhost 80

Causes an infinite hang until the `nc` client disconnects:

    2019-09-20 19:29:28,932:DEBUG:certbot.error_handler:Calling registered functions
    2019-09-20 19:29:28,932:INFO:certbot.auth_handler:Cleaning up challenges
    2019-09-20 19:29:28,932:DEBUG:certbot.plugins.standalone:Stopping server at :::80...

The standalone servers should not wait forever for clients to drain. I don't think they should wait at all and instead should just close the server socket. There is no reason to expect further legitimate traffic to the server. 

If a very patient internet crawler happens to hit the port, we need to dispatch of it quickly.

It may be appropriate to also add read and write timeouts to the socket, since it is practically always going to be exposed to the internet.

alexzorin:
Adds a default 30 second timeout to the StreamRequestHandler for clients
connecting to standalone HTTP-01 servers. This should prevent most cases
of an idle client connection preventing the standalone server from
shutting down.

Fixes #7386

## Pull Request Checklist

- [x] Edit the `master` section of `CHANGELOG.md` to include a description of the change being made.
- [x] Add [mypy type annotations](https://certbot.eff.org/docs/contributing.html#mypy-type-annotations) for any functions that were added or modified.


acme/acme/standalone.py
```diff
@@ -176,10 +176,11 @@ def __init__(self, *args, **kwargs):
 class HTTP01Server(HTTPServer, ACMEServerMixin):
     """HTTP01 Server."""
 
-    def __init__(self, server_address, resources, ipv6=False):
+    def __init__(self, server_address, resources, ipv6=False, timeout=30):
+        # type: (Tuple[str, int], Set[Any], bool, float) -> None
```
alexzorin: Would appreciate a hint on how I can get this mypy forward reference to `Set[HTTP01RequestHandler.HTTP01Resource]` instead of `Any` to work.

adferrand: Dod you try to import it? Mypy expects types to be imported like any other reference you would use in the actual code. For instance, `Any` is imported from `acme.magic_typing` that (conditionnally if available) delegates to the `typing` module.

alexzorin: I did try it originally, yes. `tox -e mypy` doesn't like it:

>acme/acme/standalone.py:179: error: Name 'HTTP01RequestHandler.HTTP01Resource' is not defined

It seems like it's because it's a forward reference. If I move the entire `HTTP01Server` class below that of `HTTP01RequestHandler`, the error goes away ... is that what I should do?

adferrand: Ah yes, I understand. Indeed, you should !

alexzorin: Thanks, you're always a great help!

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Standalone servers now have a default socket timeout of 30 seconds, fixing
```
adferrand: Could you rebase/merge master, and move this entry to master section of this changelog?

bmw:
@adferrand, do you have the time and interest to finish the review here?

I'm tentatively assigning myself for now so I can make sure this gets reviewed if you're unavailable.

alexzorin:
OK, I've got it up to date with master again.

Issue #7393, opened 2019-09-24
Drop support for Python 3.4


bmw:
This can only be done after we've released a version printing a deprecation warning.

To do this though, we need to:

- [ ] Remove Python 3.4 from the [classifiers in setup.py](https://github.com/certbot/certbot/blob/8cb57566c0be9fbed6d86316a0afb18e86d9f481/setup.py#L137) for each package.
- [ ] Update the [python_requires value](https://github.com/certbot/certbot/blob/8cb57566c0be9fbed6d86316a0afb18e86d9f481/setup.py#L125) in each packages setup.py file to exclude Python 3.4.
- [x] Update any tests in `.travis.yml` using Python 3.4 to a newer version of Python or delete the test entirely. (Done in #7394)
- [ ] Remove the deprecation warnings added in https://github.com/certbot/certbot/pull/7378.
- [x] Update any files such as [contributing.rst](https://github.com/certbot/certbot/blob/8cb57566c0be9fbed6d86316a0afb18e86d9f481/docs/contributing.rst#notes-on-os-dependencies) or [install.rst](https://github.com/certbot/certbot/blob/8cb57566c0be9fbed6d86316a0afb18e86d9f481/docs/install.rst#id16) to not claim support for Python 3.4. (Done in #7395)

TheNavigat:
If no one is working on this I'd love to have it assigned to me. It has been a while since I last contributed to Certbot and I'd like to ramp up a bit. This looks like a very simple task I can use to ramp up.

bmw:
Hi @TheNavigat!

You're more than welcome to work on this, but I want to flag that the step "Remove the deprecation warnings added in #7378" won't be possible yet and the PR may not be reviewed/merged for a few weeks while we finish the process of migrating people off of Python 3.4. It'd still be useful to have this PR ready though!

TheNavigat:
Hey @bmw,

I think I should be able to break down the PR into multiple ones. We can start by a PR that drops testing for Python 3.4, followed by another that updates documentation, followed by another that drops support. The order of merging might change, of course.

What do you think?

bmw:
Splitting the work into separate PRs sounds fine to me!

Also, in case you stopped by #letsencrypt-dev in IRC and found it empty, we've moved! We now have a Mattermost channel described at https://certbot.eff.org/docs/contributing.html#asking-for-help which is connected to the #certbot-devel IRC channel on Freenode.

bmw:
[...]

PR #7394, opened 2019-09-25
Drop Travis tests for Python 3.4

TheNavigat:
For https://github.com/certbot/certbot/issues/7393


.travis.yml
```diff
@@ -62,9 +59,6 @@ matrix:
       dist: trusty
       env: TOXENV='py27-{acme,apache,certbot,dns,nginx}-oldest'
       <<: *not-on-master
-    - python: "3.4"
-      env: TOXENV=py34
```
bmw: Sorry I didn't notice this in my last review, but can we change this to test Python 3.5? My logic is on PRs it'll keep us testing both the newest and oldest versions of Python 3 we support. Then in our nightly tests, we test on all versions.

bmw:
LGTM!

Unfortunately we had a hiccup with us dropping Python 3.4 support so we can't merge this just yet but I will once Python 3.4 has been removed.

PR #7398, opened 2019-09-25
[Apache v2] ApacheConfigurator get_virtual_hosts() for ParserNode interfaces

joohoi:
This PR implements the `get_virtual_hosts()` for ParserNode interfaces. This is the very first operation that the parser takes in the execution chain, and most of the operations later on make use of the retrieved list of virtual host objects.

While before the actual Augeas implementation of ParserNode interface, it's not possible to retrieve an actual list of VirtualHosts, the mocked tests are performed in the way that the actual execution flow would look like when parsing the real configuration.

This is a dependency for the legacy assertions.


bmw:
Test failures should be fixed by #7400.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -924,6 +941,89 @@ def get_virtual_hosts(self):
...
+        addrs = set()
+        for param in node.parameters:
+            addrs.add(obj.Addr.fromstring(param))
+
+        is_ssl = False
+        sslengine = node.find_directives("SSLEngine")
+        if sslengine:
+            for directive in sslengine:
+                # TODO: apache-parser-v2
+                # This search should be made wiser. (using other identificators)
```
ohemorange: should it be? why don't we do this now, then?

certbot-apache/certbot_apache/configurator.py
```diff
@@ -924,6 +941,89 @@ def get_virtual_hosts(self):
...
+                    pass
+
+        # "SSLEngine on" might be set outside of <VirtualHost>
+        # Treat vhosts with port 443 as ssl vhosts
+        for addr in addrs:
+            if addr.get_port() == "443":
+                is_ssl = True
+
+        macro = False
+        if node.find_directives("Macro"):
```
ohemorange: I see we're changing the test here; how come?

joohoi:
Closing this in favor of #7564

PR #7414, opened 2019-10-01
Keep compatibility with IE11 in the Nginx plugin

bmw:
As discussed at https://github.com/mozilla/server-side-tls/issues/263, Mozilla's current intermediate recommendations drop support for some non-EOL'd versions of IE. [Their TLS recommendations were updated to suggest a couple possible workarounds for people who need this support](https://github.com/mozilla/server-side-tls/pull/264) and [April suggested that we make this change in Certbot](https://github.com/mozilla/server-side-tls/issues/263#issuecomment-537085728).

We know `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA` translates to `ECDHE-RSA-AES128-SHA` because [nginx uses the same cipher format as OpenSSL](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_ciphers) and the translation is shown in the table at https://github.com/mozilla/server-side-tls/blob/gh-pages/Cipher_Suites.mediawiki.

The risk of regressions making this change is low as we always had this ciphersuite enabled just a few releases ago: https://github.com/certbot/certbot/tree/v0.36.0/certbot-nginx/certbot_nginx

I'm currently doing a bit more testing to make sure there aren't any problems with this ciphersuite and TLS 1.3, but everything is going smoothly so far.


bmw:
> I'm currently doing a bit more testing to make sure there aren't any problems with this ciphersuite and TLS 1.3, but everything is going smoothly so far.

I had no issues testing this with TLS 1.3 and TLS session tickets enabled and disabled.

PR #7433, opened 2019-10-08
Upload coverage for integration tests

bmw:
In AppVeyor, we upload coverage from both [unit tests](https://ci.appveyor.com/project/certbot/certbot/builds/27973759/job/c47a2mojahxrmomd#L831) and [integration tests](https://ci.appveyor.com/project/certbot/certbot/builds/27973759/job/hbecdexqvt7fudik#L275). Currently in Azure Pipelines, we only upload coverage from our unit tests which will cause a drop in coverage.

This fixes that by also uploading coverage from our integration tests.

### DIFF
--- a/.azure-pipelines/templates/tests-suite.yml
+++ b/.azure-pipelines/templates/tests-suite.yml
@@ -30,7 +30,7 @@ jobs:
         curl -s https://codecov.io/bash -o codecov-bash || echo "Failed to download codecov-bash"
         chmod +x codecov-bash || echo "Failed to apply execute permissions on codecov-bash"
         ./codecov-bash -F windows  || echo "Codecov did not collect coverage reports"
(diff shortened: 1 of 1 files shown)


bmw:
Overall status will show as failed because I canceled the full test suite build kicked off in Travis on this commit.

bmw:
My only concern in doing that is I don't think we want to upload coverage from `TOXENV=integration tox` runs in nightly builds because we do many more of them for different versions of Python and different ACME servers than we do on PRs. I think this difference could easily cause codecov to detect a drop in coverage which is misleading.

We could work around this, but I think we'd need to add a little more complexity than we have right now.

bmw:
I think this PR not being merged is causing Windows codecov tests to fail sometimes. See https://github.com/certbot/certbot/pull/7440.

It doesn't seem to fail all of the time as I tested it before disabling AppVeyor.

PR #7442, opened 2019-10-10
Use fresh authorizations in dry runs

alexzorin:
(Continuation of #7266)

This PR is simpler than before - the authz deactivation behavior is just attached to `--dry-run` and only performs the check after order creation.


certbot-ci/certbot_integration_tests/certbot_tests/test_main.py
```diff
@@ -594,3 +594,23 @@ def test_ocsp_status_live(context):
 
     assert output.count('INVALID') == 1, 'Expected {0} to be INVALID'.format(cert)
     assert output.count('REVOKED') == 1, 'Expected {0} to be REVOKED'.format(cert)
+
+
+def test_dry_run_deactivate_authzs(context):
+    """Test that Certbot deactivates authorizations when performing a dry run"""
+    if context.acme_server == 'pebble':
```
alexzorin: I'd like to get this integration test unskipped (or otherwise delete it).

I added authz reuse to Pebble in v2.2.1, but it is random (50%) for any eligible authz. It is possible to raise that to 100% via env var, but I am not sure whether that is safe for Certbot's other integration tests.

adferrand: I encountered the same problem while implementing the POST-as-GET protocol. Pebble by default will also reject a percentage of valid nonce during POST requests. Certbot has a recover mechanism to reissue a new nonce in this case, but not two.

It was working before POST-as-GET, but with it, the number of POST done during a certificate issuance increased of an order of magnitude. Then the probability that two nonces are rejected at a time, and so the recover mechanism fails, increased sufficiently to fail 1 out of 3 integration tests run in the certbot CI.

We hesitated between adding more tries to reduce again the probability to an acceptable rate, or disable the feature Pebble-side with the relevant env variable to get a completely reproducible behavior. 

We went with the second option finally: https://github.com/certbot/certbot/blob/118cb3c9b1e62599b57082b71ab7313b2cdc1632/certbot-ci/certbot_integration_tests/utils/acme_server.py#L127

alexzorin: Thanks. I did try `PEBBLE_AUTHZREUSE=100`, but I have *no idea* whether doing that is causing some existing test to begin always succeeding because it shares an authz identifier with yet another test. And it sounds like a landmine for future contributors. :confused:

adferrand: Oh, I thought it was about disabling totally auth reuse but the problematic is the same.

Our integrations tests are running already with 50% of auth reuse. So the problem is already here: if some tests should fail because reusing auths covers a defect, it would have happened already. Even worse, it would have happen randomly.

Nothing that kind happened. As I rewrote completely the integration tests this year from bash to python, I am also pretty confident in a theoritical perspective that challenges are properly tested: there are several tests dedicated to that, with unique domain names, so on them no auth can be reused.

adferrand: Stable tests that test less are preferable to flaky tests that test more.

The good question is : what does the production server? If it always reuses auths, so here Pebble should do the same.

alexzorin:
Thanks @adferrand, based on your advice I've bumped up Pebble's authz reuse to 100, and finished off the test coverage.

acme/acme/client.py
```diff
@@ -136,7 +136,8 @@ def deactivate_authorization(self, authzr):
         """
         body = messages.UpdateAuthorization(status='deactivated')
         response = self._post(authzr.uri, body)
-        return self._authzr_from_response(response)
+        return self._authzr_from_response(response,
```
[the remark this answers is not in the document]
alexzorin: It should have been in #7254 but wasn't caught. The defect wasn't found back then because the entire thing was dead code and the test coverage didn't check the URI. 

I discovered it when writing this PR because I needed to log any authz URIs that fail to deactivate.

adferrand:
@alexzorin could you fix the merge conflicts? I would like to merge this PR before the release tomorrow.

alexzorin:
@adferrand done

PR #7455, opened 2019-10-18
Run nightly on Azure even if no commits landed

bmw:
No nightly builds have run on Azure in 9 days. In https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers it says that the default is to only run if the source code changes. I think there is value in always running to catch changes problems in our dependencies, build environment, etc.

This PR changes Azure to always run nightly and you can see the file being successfully parsed by Azure at https://dev.azure.com/certbot/certbot/_build/results?buildId=49.

### DIFF
--- a/.azure-pipelines/advanced.yml
+++ b/.azure-pipelines/advanced.yml
@@ -12,6 +12,7 @@ schedules:
     branches:
       include:
       - master
+    always: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


bmw:
I canceled the Travis build on the `test-azure-always` branch that was running on this same commit.

PR #7457, opened 2019-10-21
[Apache v2] AugeasBlockNode find_comments() implementation

joohoi:
This PR depends on #7443 and implements:
- AugeasCommentNode and AugeasCommentNode creation from a Augeas path
- AugeasBlockNode.find_comments()


ohemorange:
Do we know what's going on with the azure tests?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -167,6 +176,19 @@ def unsaved_files(self):  # pragma: no cover
...
         return [assertions.PASS]
 
+    def _create_commentnode(self, path):
+        """Helper function to create a CommentNode from Augeas path"""
+
+        comment = self.parser.aug.get(path)
+        metadata = {"augeasparser": self.parser, "augeaspath": path}
+
+        # Because of the dynamic nature, and the fact that we're not populating
+        # the complete ParserNode tree, we use the search parent as ancestor
```
ohemorange: wait, really? this will work? won't we want to be able to do things like modify comments, or delete them or something? or are we assuming this is essentially a get-only function that just needs to be able to find and read things?

joohoi: Oh, absolutely! We will have the ability to do so. All the mentioned actions will work directly using the `metadata["augeaspath"]`. This is also the reason we don't strictly need to prepopulate the tree, and can use the existing parser functionality for the plumbing.

ohemorange:
for the codecov failure... how often are we merging master into the apache parser branch? maybe something changed with respect to that?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -152,12 +152,21 @@ def find_directives(self, name, exclude=True): # pylint: disable=unused-argument
...
-        new_metadata = {"augeasparser": self.parser}
-        return [AugeasCommentNode(comment=assertions.PASS,
-                                  ancestor=self,
-                                  filepath=assertions.PASS,
-                                  metadata=new_metadata)]
+        """
+        Recursive search of DirectiveNodes from the sequence of children.
+
+        Note that the argument exact is not implemented here, as it's not
+        currently in use in ApacheConfigurator.
```
ohemorange: now that we're actually implemented, the comment should describe the arguments and return values

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -152,12 +152,21 @@ def find_directives(self, name, exclude=True): # pylint: disable=unused-argument
...
+        """
```
ohemorange: also, will we be using exact? when would it not have to match exactly? what was the idea there?

joohoi: `exact = False` allows us to search for partial comments. For example, in ApacheConfigurator `autohsts` enhancement we're adding a generated id for each managed virtualhost, and by doing `exact = False` search allows us to retrieve it. On the contrary, we don't currently have functionality requiring `exact = True`.

joohoi: I added `raise NotImplementedError()` when calling `find_comments()` with `exact = True`

ohemorange: if we don't actually need a search for `exact = True`, can we just rip it out?

joohoi: We decided to rip it off from the interface, as there doesn't seem to be any use cases where we would strictly need to do this in the `ParserNode` context, in contrary to letting user to do this in the application.

certbot-apache/certbot_apache/augeasparser.py
ohemorange: Note: Out of band, we decided we use `assertions.PASS` instead of the behavior described in the comment. On which note, we should probably remove this comment.

joohoi: I changed the comment to reflect the current state.

PR #7461, opened 2019-10-23
[Apache v2] Implement set_parameters()

joohoi:
This PR implements `set_parameters()` for AugeasDirectiveNodes and AugeasBlockNodes.


certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -67,8 +67,30 @@ def __eq__(self, other):
...
+
+        # Clear out old parameters
+        for _ in orig_params:
+            # When the first parameter is removed, the indices get updated
+            param_path = "{}/arg[1]".format(self.metadata["augeaspath"])
+            self.parser.aug.remove(param_path)
+        # Insert new ones
+        for pi, _ in enumerate(parameters):
+            param_path = "{}/arg[{}]".format(self.metadata["augeaspath"], pi+1)
+            self.parser.aug.set(param_path, parameters[pi])
```
ohemorange: Super new to augeas syntax... parameters are set at a particular index? Surprised that we have to remove the old parameters before new ones can be added.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -67,8 +67,30 @@ def __eq__(self, other):
...
+
+        self.parameters = tuple(parameters)
+
+    def _aug_get_params(self, path):
+        """Helper function to get parameters for DirectiveNodes and BlockNodes"""
+
+        arg_paths = self.parser.aug.match(path + "/arg")
```
ohemorange: Really? They just use the same syntax? Maybe someone who knows augeas syntax should pop in for a quick look... @bmw mind double-checking the augeas bits?

certbot-apache/certbot_apache/augeasparser.py
joohoi: Yeah, the parameters are set at a particular index. We don't strictly have to remove and re-add the parameters in order to change them, but if `len(new_parameters) < len(old_parameters)` we have to explicitly clear the postponing, previously set indices. I think this approach helps to keep the code more readable, and less error prone.

certbot-apache/certbot_apache/augeasparser.py
bmw: Sure. I don't think I'll get to this today, but I can take a look tomorrow or early next week at the latest.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -67,8 +67,30 @@ def __eq__(self, other):
...
+        orig_params = self._aug_get_params(self.metadata["augeaspath"])
+
+        # Clear out old parameters
+        for _ in orig_params:
+            # When the first parameter is removed, the indices get updated
+            param_path = "{}/arg[1]".format(self.metadata["augeaspath"])
+            self.parser.aug.remove(param_path)
+        # Insert new ones
+        for pi, _ in enumerate(parameters):
+            param_path = "{}/arg[{}]".format(self.metadata["augeaspath"], pi+1)
```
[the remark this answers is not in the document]
joohoi: It works with a single index (being 1) as well. The Augeas search functions however return just `.../arg` for single item indices. This is not a problem here though. We actually discussed about enforcing indices for all items in other context for another issue (which I cannot find right now), but decided against that for the amount of changes we'd had to make and the potential bugs we might introduce while doing that.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -67,8 +67,30 @@ def __eq__(self, other):
...
+            self.parser.aug.set(param_path, parameters[pi])
+
+        self.parameters = tuple(parameters)
```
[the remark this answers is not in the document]
joohoi: 1. Good catch. I'm not sure about this. I'll look into some real world examples.
2. As you noted, having `parameters` as property would mitigate the out-of-sync issue and would probably work pretty well. I think it was initially a property but we decided to move away from that at some point of the interface design, for a reason I can't remember right now. I'll look into it.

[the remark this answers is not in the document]
joohoi: > If parameters contains a variable, it will not be expanded unlike functions like parser.get_arg. Is this the behavior we want?

I looked this up. The code in `parser.get_arg` is actually introducing a bug. Augeas expands the variables itself, and if the code in that function replaces something, it actually ends up introducing a bug instead.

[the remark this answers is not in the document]
bmw: That sounds like the right approach to me. I think we should make sure that this is done the same way in the `apacheconfig` `ParserNode`.

certbot-apache/certbot_apache/augeasparser.py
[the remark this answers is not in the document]
bmw: Great!

If it works, always using indices seems a lot simpler.

joohoi:
`parameters` is now a property. I had issues with registering the setter for it though, and wasn't able to figure it out. As this is used only for the Augeas parsernode, and as such will be short lived I would be fine leaving the property as it is in the current state (ie. not raising `TypeError` when user tries to set it).

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -192,13 +226,11 @@ def _create_directivenode(self, path):
         """Helper function to create a DirectiveNode from Augeas path"""
 
         name = self.parser.get_arg(path)
-        params = tuple(self._aug_get_params(path))
         metadata = {"augeasparser": self.parser, "augeaspath": path}
 
         # Because of the dynamic nature, and the fact that we're not populating
         # the complete ParserNode tree, we use the search parent as ancestor
         return AugeasDirectiveNode(name=name,
-                                   parameters=params,
```
ohemorange: why are we no longer passing this in?

certbot-apache/certbot_apache/tests/augeasnode_test.py
```diff
@@ -72,3 +74,59 @@ def test_find_comments(self):
...
+
+    def test_set_parameters_atinit(self):
+        from certbot_apache.augeasparser import AugeasDirectiveNode
+        servernames = self.config.parser_root.find_directives("servername")
+        setparam = "certbot_apache.augeasparser.AugeasDirectiveNode.set_parameters"
+        with mock.patch(setparam) as mock_set:
+            AugeasDirectiveNode(
+                name=servernames[0].name,
+                parameters=["test", "setting", "these"],
+                ancestor=assertions.PASS,
```
ohemorange: didn't we decide we're not using ancestor for augeas? what if we just... stop including it?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -67,8 +68,39 @@ def __eq__(self, other):
...
+        :returns: Tuple of parameters for this DirectiveNode
+        :rtype: tuple:
+        """
+        return tuple(self._aug_get_params(self.metadata["augeaspath"]))
+
+    def _aug_get_params(self, path):
+        """Helper function to get parameters for DirectiveNodes and BlockNodes"""
+
+        arg_paths = self.parser.aug.match(path + "/arg")
+        return [self.parser.get_arg(apath) for apath in arg_paths]
```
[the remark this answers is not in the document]
ohemorange: I wonder if this is still true after the refactors. regardless, I'm fine with keeping it.

certbot-apache/certbot_apache/tests/augeasnode_test.py
ohemorange: bumping this because I think it wasn't in the original review so may have gotten lost

joohoi: We decided to use the dummy assertion pass string as ancestor in order to raise an exception early if we mistakenly use it in the user code. It needs to be there however, as the ancestor `None` denotes the root node. So for me this looks like it's doing exactly what was decided.

certbot-apache/certbot_apache/augeasparser.py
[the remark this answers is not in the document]
joohoi: Yeah, still true :)

Issue #7405, opened 2019-09-26
Consistently and reliably use distro detection


bmw:
As described at https://github.com/certbot/certbot/issues/7402, it appears that `distro.linux_distribution()` cannot be used as a replacement for `platform.linux_distribution()` with the way we use the return value in the Apache plugin which is expecting specific strings to be returned.

We fixed this in https://github.com/certbot/certbot/pull/7403 by continuing to use `platform.linux_distribution` when we can, but this means we still have problems when run on Python 3.8 on Linux systems without `/etc/os-release` (which is the only time `*.linux_distribution` is called).

Unfortunately, we cannot completely remove this fallback while still supporting RHEL 6 based systems. There may be other niche systems relying on this too, but it's the only one we're aware of.

I'm open to other ideas, but to propose a way to accomplish this, I think we could:

1. Revert #7403 so we consistently use `distro` rather than having different behavior on different Python versions.
2. When not calling `linux_distribution` to build the user-agent string, call `distro.linux_distribution(full_distribution_name=False)` since it maintains an explicit list of the values it may return at https://github.com/nir0s/distro/blob/cdfe85d15bd366820db6a1cfdc6cf9a0a5de7e37/distro.py#L134-L166.
3. Add any missing strings we want to handle to the [dict in the Apache plugin](https://github.com/certbot/certbot/blob/8cb57566c0be9fbed6d86316a0afb18e86d9f481/certbot-apache/certbot_apache/entrypoint.py#L17).

As a bonus, we could maybe tackle https://github.com/certbot/certbot/issues/7399 at the same time, but I don't think this is absolutely necessary.

bmw:
@joohoi, do you have any concerns about my proposal here?

joohoi:
No, this looks good! In the scope of #7399 we might need to do some additional mapping, but I'll add the details to that issue.

PR #7467, opened 2019-10-25
Use distro library for all OS version detection

joohoi:
This pull request ensures that we use `distro` package in all the distribution version detection. It also replaces the custom systemd `/etc/os-release` parsing and adds a few version fingerprints to Apache override selection. 

Fixes: #7405


certbot/util.py
```diff
@@ -278,76 +277,52 @@ def get_filtered_names(all_names):
...
-        if os_ua:
-            return os_ua
+    return distro.version(pretty=True)
 
-    # Fallback
-    return " ".join(get_python_os_info())
-
-
-def get_systemd_os_info(filepath="/etc/os-release"):
+def get_systemd_os_info():
```
bmw: There are no calls to this function after your refactoring so I think we should get rid of it.

It's probably overkill but since I saw plugins using this module when working on https://github.com/certbot/certbot/issues/5775, I would keep its behavior but add a deprecation warning and test for it like I did for the reverter code in https://github.com/certbot/certbot/pull/7198. We can then create an issue and remove it in a subsequent release.

certbot/util.py
```diff
@@ -367,71 +342,13 @@ def get_var_from_file(varname, filepath="/etc/os-release"):
...
-
 def _normalize_string(orig):
     """
     Helper function for get_var_from_file() to remove quotes
     and whitespaces
     """
     return orig.replace('"', '').replace("'", "").strip()
 
-
-def get_python_os_info():
```
bmw: I think we still need to use this function where it was previously being called. [distro does not support Windows or macOS](https://github.com/nir0s/distro/issues/177) and its [BSD support was added in version 1.3](https://github.com/nir0s/distro/blob/master/CHANGELOG.md#130-20180509) which is newer than is available in distros like Ubuntu 18.04.

The way I'd recommend doing this is adding a boolean parameter like `full_name` to this function and then in the `linux` section of this function pass it to `distro.linux_distribution`. `distro.linux_distribution(full_distribution_name=False)[:2]` is the same as the current call in `get_os_info()` and I think `distro.linux_distribution(full_distribution_name=True)` works nicely for `get_os_info_ua()`.

certbot/util.py
```diff
@@ -278,76 +277,52 @@ def get_filtered_names(all_names):
...
     :rtype: `str`
     """
 
-    if os.path.isfile(filepath):
-        os_ua = get_var_from_file("PRETTY_NAME", filepath=filepath)
-        if not os_ua:
-            os_ua = get_var_from_file("NAME", filepath=filepath)
-        if os_ua:
-            return os_ua
+    return distro.version(pretty=True)
```
bmw: `distro.version` seems pretty different from our current behavior to me. The value in our user agent will change from "'Ubuntu 18.04.3 LTS" to "18.04 (bionic)" for example.

Instead, what about calling `distro.name(pretty=True)` and if that is an empty string call `" ".join(get_python_os_info(full_name=True))`? See my comment about adding that function back if you haven't already.

certbot/tests/util_test.py
```diff
@@ -470,79 +470,6 @@ def test_wildcard(self):
         self.assertTrue(self._call(self.wildcard.encode()))
 
 
-class OsInfoTest(unittest.TestCase):
```
bmw: I think we may want to add some of these tests back depending on your response to my suggestions.

certbot-apache/certbot_apache/entrypoint.py
```diff
@@ -16,13 +16,15 @@
...
+    "cloudlinux": override_centos.CentOSConfigurator,
     "darwin": override_darwin.DarwinConfigurator,
     "debian": override_debian.DebianConfigurator,
     "ubuntu": override_debian.DebianConfigurator,
     "centos": override_centos.CentOSConfigurator,
     "centos linux": override_centos.CentOSConfigurator,
     "fedora_old": override_centos.CentOSConfigurator,
     "fedora": override_fedora.FedoraConfigurator,
+    "linuxmint": override_debian.DebianConfigurator,
     "ol": override_centos.CentOSConfigurator,
```
bmw: I think we should add `oracle` for Oracle Linux 6 systems which do not have systemd so we can't rely on `get_systemd_os_like()`.

certbot-apache/certbot_apache/entrypoint.py
```diff
@@ -16,13 +16,15 @@
 
 OVERRIDE_CLASSES = {
     "arch": override_arch.ArchConfigurator,
+    "cloudlinux": override_centos.CentOSConfigurator,
```
bmw: Nice catch with this. I would have missed it.

certbot/util.py
joohoi: Distro does fallback on different version information sources if `/etc/os-release` isn't available. How strict is the requirement to keep version information intact for UA strings? For example, `distro.name(pretty=True)` on RHEL6 returns: `Red Hat Enterprise Linux Server release 6.10 (Santiago)`, which is not exactly the same (even though close) to the one we compile from `platform.linux_distribution()` output.

certbot/util.py
joohoi: I pushed the changes in. I'm currently installing FreeBSD to test this out (to confirm that the return values are indeed empty in this case amongst other things), so I would like to wait with approving until tested against that as well.

joohoi: Good catch about the BSD support btw, I completely missed it myself.

joohoi: Ok, this needs yet another refactor. Distro raises exception when imported under unsupported platforms.
```
ImportError: Unsupported platform: freebsd12
```

joohoi: This should apply to windows as well. Do you know why aren't we seeing the `ImportError` on our Azure tests?

bmw: Yeah. Most of our tests including those on Windows are running with [distro 1.4.0](https://github.com/certbot/certbot/blob/0f31d9b7ac1bb01a3ff6ac396fe315cb657eef6f/letsencrypt-auto-source/pieces/dependency-requirements.txt#L71). The only tests that are running with an older version are [our "oldest" tests](https://github.com/certbot/certbot/blob/0f31d9b7ac1bb01a3ff6ac396fe315cb657eef6f/tools/oldest_constraints.txt#L56) which currently only run on Linux.

certbot/util.py
bmw: There's no real requirement for them to not change. I think the only things to think about would be:

1. Are we collecting significantly more information?
2. Changing the values may make it a little harder for us to group users based on their UA.

I personally don't think either of these are changing enough with the current plan to make it a problem.

certbot/util.py
```diff
@@ -277,77 +283,58 @@ def get_filtered_names(all_names):
...
+    warnings.warn(
+        "The get_sytemd_os_like() function is deprecated and will be removed in "
+        "a future release.", DeprecationWarning, stacklevel=2)
+    return get_os_info()
 
-    return (os_name, os_version)
 
-
-def get_systemd_os_like(filepath="/etc/os-release"):
+def get_systemd_os_like(filepath="/etc/os-release"):  # pylint: disable=unused-argument
```
bmw: I think we should just delete this `filepath` argument since it's not used anymore.

certbot/util.py
```diff
@@ -277,77 +283,58 @@ def get_filtered_names(all_names):
...
     :returns: (os_name, os_version)
     :rtype: `tuple` of `str`
     """
 
-    os_name = get_var_from_file("ID", filepath=filepath)
-    os_version = get_var_from_file("VERSION_ID", filepath=filepath)
+    warnings.warn(
+        "The get_sytemd_os_like() function is deprecated and will be removed in "
+        "a future release.", DeprecationWarning, stacklevel=2)
+    return get_os_info()
```
bmw: nit:
```suggestion
    return get_os_info()[:2]
```

certbot/util.py
joohoi: I don't think we're collecting more information. The format just might change a bit. The main source for our UA info has been `/etc/os-release` PRETTY_NAME, or if that's not found, NAME. We have this from distro again, but as distro compiles multiple fingerprinting sources we might not have as much control over it.

joohoi:
I was trying to make the tests OS independent previously but the change broke that. The last commit should have fixed this issue.

bmw:
Windows tests are still angry unfortunately.

joohoi:
I decided to go with
```
@unittest.skipUnless(sys.platform.startswith("linux"), "requires Linux")
```
As I'd prefer the test suite to run on non-windows, non-linux systems without errors. Mocking out `sys` proved way too hacky.

certbot/tests/util_test.py
```diff
@@ -474,18 +475,24 @@ class OsInfoTest(unittest.TestCase):
...
 
     @mock.patch("certbot.util.distro")
+    @unittest.skipUnless(sys.platform.startswith("linux"), "requires Linux")
     def test_systemd_os_release_like(self, m_distro):
-        import certbot.util as cbutil
-        cbutil.USE_DISTRO = True
+        with mock.patch("certbot.util.sys.platform") as mock_platform:
+            mock_platform.return_value = "linux"
+            import certbot.util as cbutil
+        cbutil._USE_DISTRO = True  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
joohoi: Unfortunately looks like Python optimizes the imports, leaving the state intact in between tests, so the `_USE_DISTRO` has to be set to ensure the state in the scope of a single test.

certbot/util.py
joohoi: You are right, as the parameter isn't doing anything, I ripped it off. Better to fail early there if it's used in a third party plugin.

certbot/tests/util_test.py
```diff
@@ -473,74 +474,98 @@ def test_wildcard(self):
...
-                self.assertEqual(get_python_os_info(), ("freebsd", "9"))
-
-            with mock.patch('platform.system_alias',
-                            return_value=('windows', '', '')):
-                with mock.patch('platform.win32_ver',
-                                return_value=('4242', '95', '2', '')):
-                    self.assertEqual(get_python_os_info(),
-                                     ("windows", "95"))
+        import certbot.util as cbutil
+        cbutil._USE_DISTRO = False  # pylint: disable=protected-access
```
bmw: I'd argue that the problems implementing my suggestion at https://github.com/certbot/certbot/pull/7467/files/a1c1c3d36a7c343697e549e60b0bfc4258371c71..6654f3b9c7b644ea520d69c3adc33ae754b9e2f7#r340904031 are due to this test not properly cleaning up after itself. To fix this, I think we can change this line to:
```suggestion
        with mock.patch('certbot.util._USE_DISTRO', False):
```
and indent the rest of this function which should allow you to make the changes in https://github.com/certbot/certbot/pull/7467/files/a1c1c3d36a7c343697e549e60b0bfc4258371c71..6654f3b9c7b644ea520d69c3adc33ae754b9e2f7#r340904031.

joohoi: This is definitely the non-sloppy way of handling the issue, thanks!

joohoi:
I lowered Windows coverage test target in order to not to show general fail status for subsequent PR's, even though it doesn't block merging.

adferrand:
Sorry to arrive after the battle ^^

I see that this PR introduce some platform specific code in `certbot.util` module. Usually in my PRs related to Windows, I follow as a general design pattern to put all platform specific code in the `certbot.compat` package, and then call modules of this package from the rest of certbot code.

This is for maintainability and testability concerns.

Would it be possible to move the platform specific code into `certbot.compat.misc` for instance?

bmw:
@adferrand, I think most of the platform specific code here existed before this PR. The use of `_USE_DISTRO` is actually to fix an existing problem Joona found where `distro` which we were previously unconditionally importing errors when not running on Linux in older versions. See https://github.com/certbot/certbot/pull/7467#discussion_r339952203.

I also would like this PR for our release on Tuesday to resolve https://github.com/certbot/certbot/issues/7405 to fix unresolved issues in our Python 3.8 support.

I'm all for hiding platform specific code behind `certbot.compat`, but since the release is in 2 weekdays, how about we open an issue for us to make these changes in the future?

adferrand:
OK, if the code was mostly already there, it is fine for me.

This refactor could typically be done when me or someone else will make the Apache plugin compatible with Windows. An issue to remind us of this is sufficient I think.

bmw:
Thanks Adrien. Issue is https://github.com/certbot/certbot/issues/7488.

PR #7473, opened 2019-10-30
change random sleep to use fractional seconds

jprenken:
If we use fractional instead of whole seconds for the random sleep before renewing, it will reduce bunching of requests to Let's Encrypt's API.


schoen:
I am totally astonished that we're living in a future where server clocks are accurate enough, and Let's Encrypt is popular enough, for this to matter!

bmw:
Seems fine to me!

PR #7491, opened 2019-11-01
Clarify when the changelog should be modified

bmw:


pull_request_template.md
```diff
@@ -1,7 +1,8 @@
 ## Pull Request Checklist
 
-- [ ] Edit the `master` section of `CHANGELOG.md` to include a description of
-  the change being made.
+- [ ] If the change isn't just to unpackaged tests or scripts, edit the
```
adferrand: "isn't just for"? But I may be wrong.

Also, I think the recommendation would be more efficient if we could give an idea to contributors about what is an unpackaged test/script.

bmw: I think "just to" is better. You made the change "to" the unpackaged tests or scripts, not "for" the unpackaged test or scripts.

I agree but I think it's a little tricky. How about I list a somewhat high level the components that are shipped such as `acme`, `certbot`, `certbot-apache`, etc. and `certbot-auto`. I would not explicitly call out files like `setup.cfg`, `setup.py`, `MANIFEST.in`, etc. in the root of the repo because I think it's excessive and the need for it should go away soon.

adferrand: Ok for the `to`, and I agree with the approach you proposed here.

bmw:
Gah this relative link doesn't always work. I'll do this:

1. Open a separate PR for the doc changes.
2. Open a PR for the Certbot website to push those changes live.
3. Update this PR to link to link to the live site.

PR #7494, opened 2019-11-01
[POC] Lexicon plugin provider

trinopoty:
This PR implements a way to provide lexicon providers to certbot as discussed in issue #6178 .
Each provider has to be explicitly added in order to be provided to certbot with a few lines of code.
Also, I've made it so that a single DNS plugin provided to certbot can be mapped to more than one lexicon provider if desired/required.
Almost all possible modifications needed to instantiate a lexicon provider is provided and the implementation details are all hidden away from certbot.

I had to put this in certbot itself because of the way certbot is structured right now. I could not find a way to properly hook into the system from a plugin.

This PR is a POC only at the time being and is here for discussion. That is why there is no tests as of now.

### DIFF
--- a/certbot/certbot/_internal/cli.py
+++ b/certbot/certbot/_internal/cli.py
@@ -26,10 +26,12 @@ from certbot import util
 from certbot._internal import constants
 from certbot._internal import hooks
 from certbot._internal.plugins import disco as plugins_disco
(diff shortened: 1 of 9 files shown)


trinopoty:
I was able to successfully issue a certificate using this PR and the implemented Godaddy plugin.

PR #7497, opened 2019-11-04
[Apache v2] Adding nodes 1/3 : add_child_block()

joohoi:
Implement `add_child_block()` and the functionality required to resolve position within immediate children.

### DIFF
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -1,4 +1,69 @@
-""" Augeas implementation of the ParserNode interfaces """
+"""
+Augeas implementation of the ParserNode interfaces.
+
+Augeas works internally by using XPATH notation. The following is a short example
+of how this all works internally, to better understand what's going on under the
+hood.
+
+A configuration file /etc/apache2/apache2.conf with the following content:
+
+    # First comment line
+    # Second comment line
+    WhateverDirective whatevervalue
+    <ABlock>
+        DirectiveInABlock dirvalue
+    </ABlock>
+    SomeDirective somedirectivevalue
+    <ABlock>
+        AnotherDirectiveInABlock dirvalue
+    </ABlock>
+    # Yet another comment
+
+
+Translates over to Augeas path notation (of immediate children), when calling
+for example: aug.match("/files/etc/apache2/apache2.conf/*")
+
+[
+    "/files/etc/apache2/apache2.conf/#comment[1]",
+    "/files/etc/apache2/apache2.conf/#comment[2]",
+    "/files/etc/apache2/apache2.conf/directive[1]",
+    "/files/etc/apache2/apache2.conf/ABlock[1]",
+    "/files/etc/apache2/apache2.conf/directive[2]",
+    "/files/etc/apache2/apache2.conf/ABlock[2]",
+    "/files/etc/apache2/apache2.conf/#comment[3]"
+]
+
+Regardless of directives name, its key in the Augeas tree is always "directive",
+with index where needed of course. Comments work similarly, while blocks
+have their own key in the Augeas XPATH notation.
+
+It's important to note that all of the unique keys have their own indices.
+
+Augeas paths are case sensitive, while Apache configuration is case insensitive.
+It looks like this:
+
+    <block>
+        directive value
+    </block>
+    <Block>
+        Directive Value
+    </Block>
+    <block>
+        directive value
+    </block>
+    <bLoCk>
+        DiReCtiVe VaLuE
+    </bLoCk>
+
+Translates over to:
+
+[
+    "/files/etc/apache2/apache2.conf/block[1]",
+    "/files/etc/apache2/apache2.conf/Block[1]",
+    "/files/etc/apache2/apache2.conf/block[2]",
+    "/files/etc/apache2/apache2.conf/bLoCk[1]",
+]
+"""
 
 from certbot_apache import apache_util
 from certbot_apache import assertions
@@ -320,7 +385,9 @@ class AugeasBlockNode(AugeasDirectiveNode):
             self.metadata["augeaspath"])
         )
 
-        # Augeas indices start at 1
+        # Calculate resulting_path
+        # Augeas indices start at 1. We use counter to calculate the index to
+        # be used in resulting_path.
         counter = 1
         for i, child in enumerate(all_children):
             if position is not None and i >= position:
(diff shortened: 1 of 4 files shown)


certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -253,6 +286,72 @@ def _aug_get_block_name(self, path):
...
+            if name == childname:
+                counter += 1
+
+        resulting_path = "{}/{}[{}]".format(
+            self.metadata["augeaspath"],
+            name,
+            counter
+        )
+
+        # Inserting the only child child and appending as the last child work
```
```suggestion
        # Inserting the only child and appending as the last child work
```

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -253,6 +286,72 @@ def _aug_get_block_name(self, path):
         name = name.split("[")[0]
         return name
 
+    def _aug_resolve_child_position(self, name, position):
+        """
+        Helper function that iterates through the immediate children and figures
+        out the insertion path for a new AugeasParserNode.
+
+        Augeas also generalizes indices for directives and comments, simply by
```
ohemorange: what does "generalizes indices" mean?

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -253,6 +286,72 @@ def _aug_get_block_name(self, path):
...
+        counter = 1
+        for i, child in enumerate(all_children):
+            if position and i >= position:
+                # We're not going to insert the new node to an index after this
+                break
+            childname = self._aug_get_name(child)
+            if name == childname:
+                counter += 1
+
+        resulting_path = "{}/{}[{}]".format(
```
ohemorange: seriously? we have to calculate this ourselves? augeas doesn't return it??

joohoi: Yup, seriously.

certbot-apache/certbot_apache/augeasparser.py
joohoi: I probably should have opened the context a bit, but I'll do it here.
So if we have a config file `/etc/apache2/apache2.conf` with the following contents:
```
# First comment line
# Second comment line
WhateverDirective whatevervalue
<ABlock>
    DirectiveInABlock dirvalue
</ABlock>
SomeDirective somedirectivevalue
<ABlock>
    AnotherDirectiveInABlock dirvalue
</ABlock>
# Yet another comment
```
It translates to augeas tree nodes as follows, when calling `aug.match("/files/etc/apache2/apache2.conf/*")`:

```
[
"/files/etc/apache2/apache2.conf/#comment[1]",
"/files/etc/apache2/apache2.conf/#comment[2]",
"/files/etc/apache2/apache2.conf/directive[1]",
"/files/etc/apache2/apache2.conf/ABlock[1]",
"/files/etc/apache2/apache2.conf/directive[2]",
"/files/etc/apache2/apache2.conf/ABlock[2]",
"/files/etc/apache2/apache2.conf/#comment[3]"
]
```
So regardless of directive name, its key in the Augeas tree is always "directive", same goes with the comments, and all of them have their own respective indices. Blocks however have their actual name in the Augeas path.

In this comment with generalizing I meant "hiding" the actual name of the directive or a comment.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -253,6 +289,72 @@ def _aug_get_block_name(self, path):
...
+
+        :returns: Tuple of insert path, resulting path and a boolean if the new
+            node should be inserted before it.
+        :rtype: tuple of str, str, bool
+        """
+
+        # Default to appending
+        before = False
+
+        all_children = self.parser.aug.match("{}/*".format(
```
ohemorange: @bmw there's a few lines of augeas going in this function that look reasonable but I know I'll miss any weird nuances so would love an extra set of eyes!

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -253,6 +289,72 @@ def _aug_get_block_name(self, path):
         name = name.split("[")[0]
         return name
 
+    def _aug_resolve_child_position(self, name, position):
```
bmw: Will `self.metadata["augeaspath"]` ever already have a trailing slash? If it's possible, I think we should check for this case and make sure the path we construct doesn't contain two consecutive slashes as Augeas' behavior was different when testing this case locally.

certbot-apache/certbot_apache/augeasparser.py
bmw: Other than my comment at https://github.com/certbot/certbot/pull/7497/files#r343358309, the Augeas code in this function looks correct to me!

certbot-apache/certbot_apache/augeasparser.py
ohemorange: actually having this example as a whole has been pretty helpful for reviewing the other ones and might be useful to have in the comment

certbot-apache/certbot_apache/augeasparser.py
joohoi: It _should_ not. We however used that code in `ApacheParser` and I decided to bring it over to ensure consistency. Do you think it'd be better to just raise an exception instead to catch the occurances (if any) early?

certbot-apache/certbot_apache/augeasparser.py
joohoi: I added the example to the main pydoc of the file. I also added a short description with an example how case sensitivity works within Augeas paths.

certbot-apache/certbot_apache/augeasparser.py
bmw: Are you referring to specific code in `ApacheParser` or just that we make the assumption that the Augeas path won't end in a trailing slash in most places?

I personally think writing an assertion in `AugeasParserNode.__init__` or `AugeasDirectiveNode.__init__` to cause the code to fail fast if the given `augeaspath` ends in a trailing slash is a good idea.

joohoi: It was based on the code that checks the path in [current parser](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/parser.py#L649-L650). According to git log, I myself wrote those lines, and just half a year ago. I can't remember the reason for the check anymore though. I think it was added, because we do a lot of manual Augeas path string manipulation in the current code.

In the ParserNode implementation, the path creation is managed better, and we shouldn't have the same issue. I'll a add the check to `__init__` nevertheless.

bmw: Sounds good to me!

We could alternatively just fix the problem in `__init__` and remove the trailing slash rather than erroring if you prefer, but I personally think both of those approaches are better than possibly letting the problem go unchecked and erroring much later in execution.

Issue #7365, opened 2019-09-03
Certbot Windows installer feature tracker


bmw:
This issue is just to help us track the various features we want to look into for our Windows installer. This list started at https://github.com/certbot/certbot/pull/7324#issue-307052277 which I modified a bit:

- [x] register a scheduled task for certificate renewal
- [ ] auto-update of Certbot when a new release is available
- [x] online documentation for Windows
- [ ] installer distribution plan (continuous deployment + distribution channels)
- [ ] method to check the downloaded installer is untampered
- [x] pin runtime dependencies
- [x] pin build dependencies
- [ ] installation of additional plugins, such as 3rd party plugins
- [ ] Make our installer trusted by Windows

EDIT: Making our installer trusted by Windows is a thing I think we should look into that was flagged by Adrien.

bmw:
I think this is the list of features we need before fully claiming to support Certbot on Windows, but we could offer a beta before a lot of these features are resolved.

adferrand:
In order to meet two more requirements: check that the downloaded installer is untampered + make our installer trusted by Windows.

The solution is Microsoft Authenticode. This is a Windows built-in functionality that is able to check that a file (including executables) is genuine: it allows to authenticate its source and that its content is not tampered.

Without this, any execution of the installer would generate a warning that the user will need to explicitly override to continue installation. And of course an error would be thrown if the installer has been modified in any way.

Microsoft Authenticode is implemented on top of X509 certificates functionalities: content is signed and associated to a certificate, and a chain of trust is used to validate the certificate against CA root certificates accepted by the system. At this point, something like the bells of Notre-Dame should have rung to the reader.

Here is a working example to generate a code-signing certificate from an arbitrary CA, and to use it to sign the Certbot for Windows installer. Compatible with Linux :)

```bash
#!/bin/bash

# Requirements:
#  - OpenSSL is installed
#  - Mono is installed (it includes the signcode binary)
#  - CA key (myCA.key) and cert (myCA.pem) files are available locally
#  - Certbot for Windows installer file is available locally

# Prepare OpenSSL configuration for CSR
cat << EOF > openssl.cfg
[req]
req_extensions = v3_req
distinguished_name = req_distinguished_name
prompt = no

[req_distinguished_name]
C = US
ST = CA
L = San Francisco
O = EFF
OU = Certbot dev team
CN = Electronic Frontier Foundation

[v3_req] 
basicConstraints = CA:FALSE
keyUsage = digitalSignature
[...]

PR #7498, opened 2019-11-04
Build Windows installers with pinned dependencies

adferrand:
This PR modifies `windows-installer\construct.py` in order to use `letsencrypt-auto-source\pieces\dependency-requirements.txt` as a constraint to get/build the wheels to include in the Windows installer.

This way the Windows installer contains pinned dependencies as it is done for installations made by `certbot-auto` or in Certbot dockers.

This PR is a part of #7365 requirements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,16 +54,12 @@ install_requires = [
     'zope.interface',
 ]
 
-# Load minimal pywin32 version from the Windows installer builder
-windows_installer = os.path.join(here, 'windows-installer', 'construct.py')
-pywin32_version_search = re.search(r'PYWIN32_VERSION = (\d+)', read_file(windows_installer))
-pywin32_req = 'pywin32>={0}'.format(pywin32_version_search.group(1))
-
 # Add pywin32 on Windows platforms to handle low-level system calls.
 # This dependency needs to be added using environment markers to avoid its installation on Linux.
 # However environment markers are supported only with setuptools >= 36.2.
 # So this dependency is not added for old Linux distributions with old setuptools,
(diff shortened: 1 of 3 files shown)


adferrand:
This PR also synchronize the `pywin32` requirement defined in `setup.py` with the one defined in the Windows installer builder. This is done in a similar way to how Certbot version is extracted in order to maintain the compatibility with older systems that would build Certbot.

adferrand:
Full tests suite:
* Travis: https://travis-ci.com/certbot/certbot/builds/134931502
* Azure Pipelines: https://dev.azure.com/certbot/certbot/_build/results?buildId=241

adferrand:
> 
> 
> This PR also synchronize the `pywin32` requirement defined in `setup.py` with the one defined in the Windows installer builder. This is done in a similar way to how Certbot version is extracted in order to maintain the compatibility with older systems that would build Certbot.

But ... it may `setup.py` depends on `windows-installer` artifacts, and these packaging utilities have nothing to do in something like a sdist. I do not want to reverse the relationship, because `setup.py` is not designed to define in any way a strict dependency in Certbot project.

So I will simply define separately the version in `setup.py` and in `construct.py`, adding a comment to remember to change accordingly both files. Inconsistencies between the two files would be caught by integration tests, if some breaking change is introduce by a new version of pywin32 for instance.

adferrand:
New full tests suite:
* Travis: https://travis-ci.com/certbot/certbot/builds/134935906
* Azure Pipelines: https://dev.azure.com/certbot/certbot/_build/results?buildId=247

PR #7503, opened 2019-11-05
dns-rfc2136: use TCP to query SOA records

### LINKED ISSUE #7502: dns-rfc2136 is unusable if packet loss present
certbot_dns_rfc2136 authenticator is not robust on a losing network.

During authenticating, the plugin have to send several SOA queries directly to the authoritative nameserver.  The number of queries is depended on the specific configuration. In my case, it sends out 21 queries for a single certification with 6 dns-alt names.

Currently, it sends them out using UDP **without timeout** and without retry mechanism. Thus, any single packet lost on these queries will cause the certbot stuck forever.

https://github.com/certbot/certbot/blob/3c24ff88cc0106ac39e5b0f5bd6bf0f29572201e/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py#L209-L210

In my case, the network between my webserver and the nameserver regularly get a ~20% packet loss rate (since one of them is in China). The chance I pass the authentication is around 0.8 ^ 21 < 1%.

## Proposed solution

1. Add a timeout with proper retry mechanism; or
2. Simply use TCP and let the OS handle it for us.

sorz:
certbot-dns-rfc2136: Use TCP queries to improve the network robust. Fixes #7502.

### DIFF
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -206,7 +206,7 @@ class _RFC2136Client(object):
         request.flags ^= dns.flags.RD
 
         try:
-            response = dns.query.udp(request, self.server, port=self.port)
+            response = dns.query.tcp(request, self.server, port=self.port)
             rcode = response.rcode()
 
             # Authoritative Answer bit should be set
--- a/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
+++ b/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
@@ -206,7 +206,11 @@ class _RFC2136Client(object):
         request.flags ^= dns.flags.RD
 
         try:
-            response = dns.query.tcp(request, self.server, port=self.port)
+            try:
+                response = dns.query.tcp(request, self.server, port=self.port)
+            except OSError as e:
+                logger.debug('TCP query failed, fallback to UDP: %s', e)
+                response = dns.query.udp(request, self.server, port=self.port)
             rcode = response.rcode()
 
             # Authoritative Answer bit should be set
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 * acme.standalone.BaseRequestHandlerWithLogging and
   acme.standalone.simple_tls_sni_01_server have been deprecated and will be
   removed in a future release of the library.
+* certbot-dns-rfc2136 now use TCP to query SOA records.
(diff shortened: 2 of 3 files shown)


alexzorin:
To fix the test failure, you will also need to change the `test_query_soa_` tests to mock out TCP instead of UDP [in this `dns_rfc2136_test.py`](https://github.com/certbot/certbot/blob/b84edfd39a94d34ce51658e5968787807d490a13/certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136_test.py#L165-L194).

sorz:
> a significant part of administrators forget to set up the TCP listening on their server

But this is not the case for all the *current* users of certbot-dns-rfc2136?
Since `add_txt_record()` and `del_txt_record()` are already using TCP, change `_query_soa()` to TCP should not be a regression. They all talk to the same user-configured authoritative nameserver.

adferrand:
> > a significant part of administrators forget to set up the TCP listening on their server
> 
> But this is not the case for all the _current_ users of certbot-dns-rfc2136?
> Since `add_txt_record()` and `del_txt_record()` are already using TCP, change `_query_soa()` to TCP should not be a regression. They all talk to the same user-configured authoritative nameserver.

Well, you are totally right. So the fallback will most likely be useless. However, it does not harm, so thanks anyway :)

bmw:
Has the changes here been tested (outside of our fully mocked out tests)?

Due to a previous simple change to this plugin which completely broke it, PRs for DNS plugins are supposed to have two reviews per our [postmortem notes](https://docs.google.com/document/d/1CDFazf8MfHa5NA4BxrU0mLMrg1TD03TDdg8cO1Kjb8w/edit?usp=sharing).

certbot-dns-rfc2136/certbot_dns_rfc2136/dns_rfc2136.py
```diff
@@ -206,7 +206,11 @@ def _query_soa(self, domain_name):
         request.flags ^= dns.flags.RD
 
         try:
-            response = dns.query.udp(request, self.server, port=self.port)
+            try:
+                response = dns.query.tcp(request, self.server, port=self.port)
+            except OSError as e:
```
bmw: Why are we only catching OSErrors here?

From the [source of dns.query.tcp](http://www.dnspython.org/docs/1.16.0/dns.query-pysrc.html#tcp), it looks like other types can be raised.

bmw:
I appreciate the PR @sorz and I'd bet your code is fine, but since we've had trouble here in the past and our tests for this plugin are a bit lacking, I'm going to revert this PR for now. After that lands, by all means feel free to open the PR again. Sorry for the trouble.

bmw:
@adferrand, told me that @joohoi also gave his +1 on this PR and if so that's fine with me. Sorry for the noise.

PR #7517, opened 2019-11-06
[Apache v2] Adding nodes 2/3 : add_child_directive()

joohoi:
This PR implements add_child_directive() and is built on top of #7497

Note: this PR uses temporarily #7497 as base to help to hide the unnecessary diffs. This also means that the tests won't run before it's changed back to `apache-parser-v2`

### DIFF
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -221,12 +221,26 @@ class AugeasBlockNode(AugeasDirectiveNode):
     # pylint: disable=unused-argument
     def add_child_directive(self, name, parameters=None, position=None):  # pragma: no cover
         """Adds a new DirectiveNode to the sequence of children"""
-        new_metadata = {"augeasparser": self.parser, "augeaspath": assertions.PASS}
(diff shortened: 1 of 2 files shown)


certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -146,12 +148,26 @@ def add_child_block(self, name, parameters=None, position=None):  # pragma: no c
...
+        insertpath, realpath, before = self._aug_resolve_child_position(
+            "directive",
+            position
+        )
+        new_metadata = {"augeasparser": self.parser, "augeaspath": realpath}
+
+        # Create the new directive
+        self.parser.aug.insert(insertpath, "directive", before)
+        # Set the directive key
+        self.parser.aug.set(realpath, name)
```
ohemorange: what's this? why didn't we need to do it with `add_child_block`?

joohoi: That's because blocks use their name as a key in Augeas XPATH notation, while directives always use key "directive", we need to set the name for it.

PR #7518, opened 2019-11-06
[Apache v2] Adding nodes 3/3 : add_child_comment() 

joohoi:
This PR implements add_child_comment()

Note: this PR uses temporarily #7497 as base to help to hide the unnecessary diffs. This also means that the tests won't run before it's changed back to apache-parser-v2

### DIFF
--- a/certbot-apache/certbot_apache/tests/augeasnode_test.py
+++ b/certbot-apache/certbot_apache/tests/augeasnode_test.py
@@ -132,6 +132,16 @@ class AugeasParserNodeTest(util.ApacheTest):
                 found = True
         self.assertTrue(found)
 
+    def test_add_child_comment(self):
+        newc = self.config.parser_root.primary.add_child_comment("The content")
+        comments = self.config.parser_root.find_comments("The content")
+        self.assertEqual(len(comments), 1)
+        self.assertEqual(
+            newc.metadata["augeaspath"],
+            comments[0].primary.metadata["augeaspath"]
+        )
+        self.assertEqual(newc.comment, comments[0].comment)
+
     def test_add_child_block(self):
         nb = self.config.parser_root.primary.add_child_block(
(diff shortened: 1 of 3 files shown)


certbot-apache/certbot_apache/tests/augeasnode_test.py
```diff
@@ -144,6 +144,16 @@ def test_set_parameters_delete(self):
                 found = True
         self.assertTrue(found)
 
+    def test_add_child_comment(self):
+        newc = self.config.parser_root.primary.add_child_comment("The content")
+        comments = self.config.parser_root.find_comments("The content")
+        self.assertEqual(len(comments), 1)
+        self.assertEqual(
+            newc.metadata["augeaspath"],
```
sydneyli: Should this also be `primary.metadata`?

certbot-apache/certbot_apache/tests/augeasnode_test.py
```diff
@@ -144,6 +144,16 @@ def test_set_parameters_delete(self):
                 found = True
         self.assertTrue(found)
 
+    def test_add_child_comment(self):
+        newc = self.config.parser_root.primary.add_child_comment("The content")
```
sydneyli: Remove `primary`?

certbot-apache/certbot_apache/tests/augeasnode_test.py
joohoi: Good catch, and you are right. There's more than meets the eye though: All "read-only" operations towards instance variables already inherently are, as `DualParserNode` returns the value from primary from `__getattr__`.

I'm mainly accessing `primary` directly where I need to count the calls, and in all cases where I need to make "write" operations to not to do the writes twice as you already figured out :)

joohoi:
This should be ready to merge

Issue #7500, opened 2019-11-04
Remove the code deprecated util code


bmw:
We should remove the code deprecated in https://github.com/certbot/certbot/pull/7467.

PR #7526, opened 2019-11-08
Remove get_systemd_os_info

bmw:
Fixes https://github.com/certbot/certbot/issues/7500.


bmw:
Normally I think that'd be nice, but the next release is 1.0 meaning we couldn't remove this until 2.0 and I don't think it's worth waiting or bumping the major version for this change. I don't think anyone is using this code, but we gave it a deprecation warning for a release in https://github.com/certbot/certbot/pull/7467 just in case.

PR #7528, opened 2019-11-08
Make uncomplicated modules private

ohemorange:
First step towards of #5775.

This PR moves modules to `_internal` that aren't called from plugins and removes them from `docs/api`. It was created by manually creating `_internal/__init__.py`, and then running the following script:

```
modify_item () {
    git grep -l "from certbot import $1" | xargs sed -i "s/from certbot import $1/from certbot._internal import $1/g"
    git grep -l "certbot\.$1" | xargs sed -i "s/certbot\.$1/certbot._internal.$1/g"
    git status
    mv certbot/$1.py certbot/_internal
    rm docs/api/$1.rst
    git checkout -- CHANGELOG.md
    git add -A
    git commit -m "Move $1.py to _internal"
}

for module in account auth_handler cert_manager client error_handler lock main notify ocsp renewal reporter storage updater
do
  modify_item $module
done
```

`log.py` and `eff.py` were deemed too complicated for this script and will therefore be in their own PRs. Likewise, anything with a bullet point or in a subfolder of `certbot` will be in another PR.

Happy to split these commits into a PR each if that's easier to review.


certbot/_internal/lock.py
```diff
@@ -33,7 +33,7 @@ def lock_dir(dir_path):
     :raises errors.LockError: if unable to acquire the lock
 
     """
-    return LockFile(os.path.join(dir_path, '.certbot.lock'))
+    return LockFile(os.path.join(dir_path, '.certbot._internal.lock'))
```
bmw: I think we should undo the content changes to this file and the other places `.certbot.lock` was changed into `.certbot._internal.lock`. `sed` was greedy and `.certbot.lock` here refers to the name of the lockfile created on disk.

certbot-apache/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
 acme[dev]==0.29.0
-certbot[dev]==0.39.0
+-e certbot[dev]
```
bmw: This shouldn't be needed if we change back the name of the lockfile, but if you want to keep it for some reason, I think this should be `-e .[dev]` and we should update `certbot-apache/setup.py` as well.

certbot-nginx/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
 acme[dev]==0.29.0
-certbot[dev]==0.36.0
+-e certbot[dev]
```
bmw: I think changing this to `-e .[dev]` should fix the oldest test failures. (Also, we shouldn't have to use `.` anymore in the future after step 2.a under "How do we express this?" in the doc is done.)

Also, don't forget to update `certbot-nginx/setup.py` to match the changes here.

certbot/_internal/lock.py
ohemorange: ohhhhhhhhhh that explains it. I was wondering...

certbot-apache/local-oldest-requirements.txt
ohemorange: interesting -- this is technically true, but apache does refer in comments to `_internal`. shouldn't matter either way, really.

bmw: That's a good point. I'd vote to bump the minimum version then so that's always a valid path to an object but I don't care too much.

(The object being referenced is private which I think is another problem and something I missed in my initial design, but I personally think we can punt this issue down the road.)

ohemorange: I already reverted; if we decide it's actually a problem that the object referenced is private we'll end up bumping later anyway.

certbot-nginx/local-oldest-requirements.txt
ohemorange: I actually think this should ideally be resolved by fixing the lock thing as well, if it's not I'll be surprised.

ohemorange: ah jk this was a different error in the lock test

ohemorange: why is there an external lock test being run under `tests/` by `nginx-oldest` anyway

bmw: The test is a simple system test that we're locking files the way we're supposed to and like our other tests, it's run on both the "newest" and "oldest" setups to make sure everything still works. It's relevant for nginx because it also test the lockfile created by the nginx plugin.

ohemorange: but the nginx plugin doesn't import lock.py...

bmw: But it calls `certbot.util.lock_dir_until_exit` [here](https://github.com/certbot/certbot/blob/4b488614cf7749c8139c11f0983fe4b71e29827f/certbot-nginx/certbot_nginx/configurator.py#L210) which does internally.

If I remember correctly, I think you (rightly) pushed for `tests/lock_test.py` to test that all of our locks were always acquired in the same order.

ohemorange: yep, still think that's correct. still weird that changes to internal naming would make the nginx tests break.

bmw: I think the problem is that `tests/lock_test.py` imports `certbot.lock`. This is awkward, it probably shouldn't, and certbot-nginx doesn't actually require a newer version of Certbot, but I personally think just bumping the minimum version to fix this is fine.

ohemorange: agreed on all counts.

certbot-nginx/local-oldest-requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Remember to update setup.py to match the package versions below.
-acme[dev]==0.29.0
-certbot[dev]==0.36.0
+-e acme[dev]
+-e .[dev]
```
bmw: I think we should update `certbot-nginx/setup.py` to reflect these changes as well since we're no longer testing with older versions.

certbot/tests/client_test.py
```diff
@@ -18,12 +18,13 @@
 KEY = test_util.load_vector("rsa512_key.pem")
 CSR_SAN = test_util.load_vector("csr-san_512.pem")
 
+# pylint: disable=line-too-long
```
bmw: Gah I wish we didn't have to do this but I think this is better than cleaning up all of these lines right now.

certbot/tests/main_test.py
```diff
@@ -916,9 +916,10 @@ def test_csr_with_inconsistent_domains(self):
             'certonly -d example.org --csr {0}'.format(CSR).split())
 
     def _certonly_new_request_common(self, mock_client, args=None):
-        with mock.patch('certbot.main._find_lineage_for_domains_and_certname') as mock_renewal:
+        with mock.patch('certbot._internal.main._find_lineage_for_domains_and_certname') \
+            as mock_renewal:
```
bmw: Thanks for fixing the long lines in this file.

PR #7531, opened 2019-11-09
Move log.py to _internal

ohemorange:
Part of #5775. Methodology similar to #7528, but slightly more manual.


ohemorange:
Closing and reopening in the hopes of making codecov happy

PR #7533, opened 2019-11-09
Move items in certbot/plugins to _internal

ohemorange:
Part of #5775. Methodology similar to #7532.

```
export folder=plugins

mkdir certbot/_internal/$folder
cp certbot/$folder/__init__.py certbot/_internal/$folder

git add -A
git commit -m "Create and initialize _internal/$folder"

modify_item () {
    export item=$1
    git grep -l "from certbot\.$folder import $item" | xargs sed -i "s/from certbot\.$folder import $item/from certbot._internal.$folder import $item/g"
    git grep -l "certbot\.$folder\.$item" | xargs sed -i "s/certbot\.$folder\.$item/certbot._internal.$folder.$item/g"
    git status
    mv certbot/$folder/$item.py certbot/_internal/$folder
    rm docs/api/$folder/$item.rst
    git checkout -- CHANGELOG.md
    git add -A
    git commit -m "Move $folder/$item.py to _internal/"
}

for module in manual disco selection webroot null standalone
do
  modify_item $module
done
```


certbot/_internal/plugins/__init__.py
```diff
@@ -0,0 +1 @@
+"""Certbot client.plugins."""
```
bmw: I think `client.plugins` here refers to when the plugins directory was contained in a `client` directory (see 2d848994f4aec4a640ac6c526d9ea28f68dc8ff2 if you're curious).

I don't think we should we should copy this over to this new file. Bonus points for fixing it in `certbot/plugins/__init__.py` too, but I'm also fine doing that later since it's not technically related to this PR.

PR #7534, opened 2019-11-09
Move constants.py to _internal

ohemorange:
Part of #5775. Methodology similar to #7528, but slightly more manual.


ohemorange:
I agree we should change it.

grep results:
```
$ git grep ENHANCEMENTS
certbot-apache/certbot_apache/configurator.py:            :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot-apache/certbot_apache/configurator.py:            See :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot-apache/certbot_apache/tests/configurator_test.py:    # TEST ENHANCEMENTS
certbot-nginx/certbot_nginx/configurator.py:            :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot-nginx/certbot_nginx/configurator.py:            See :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot/_internal/display/enhancements.py:        :class:`certbot.CONFIG.ENHANCEMENTS` enhancements
certbot/interfaces.py:            :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot/interfaces.py:            :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot/interfaces.py:            :const:`~certbot._internal.constants.ENHANCEMENTS`
certbot/plugins/enhancements.py:ENHANCEMENTS = ["redirect", "ensure-http-header", "ocsp-stapling"]
```

I have one left that I wasn't sure about, should that also be changed? What did it mean by `certbot.CONFIG.ENHANCEMENTS` in the first place?

```
def ask(enhancement):
    """Display the enhancement to the user.

    :param str enhancement: One of the
        :class:`certbot.CONFIG.ENHANCEMENTS` enhancements

    :returns: True if feature is desired, False otherwise
    :rtype: bool

    :raises .errors.Error: if the enhancement provided is not supported

    """
```

bmw:
> I have one left that I wasn't sure about, should that also be changed? What did it mean by certbot.CONFIG.ENHANCEMENTS in the first place?

I _think_ it's either a typo or really outdated. I poked through the `git` history and made it back to 72813ec1f where the configuration object we currently pass around was a global named `CONFIG`. This predates my involvement in the project, but I think enhancements either used to be an attribute of this object or whoever wrote that comment thought it was.

It's not anymore though so I think we should change it too.

ohemorange:
"This predates my involvement in the project" 🤦 ok changing it

PR #7537, opened 2019-11-11
[Windows] Avoid letsencrypt.log permissions error during scheduled certbot renew task

adferrand:
While coding for #7536, I ran into another issue. It appears that Certbot logs generated during the scheduled task execution have wrong permissions that make them almost unusable: they do not have an owner, and their ACL contains nonsense values (non existant accounts name).

The class `logging.handler.RotatingFileHandler` is responsible for these logs, and become mad when it is in a Python process run under a scheduled task owned by `SYSTEM`. This is precisely our case here.

This PR avoids (but not fix) the issue, by changing the owner of the scheduled task from `SYSTEM` to the `Administrators` group, that appears to work fine.

### DIFF
--- a/windows-installer/renew-up.ps1
+++ b/windows-installer/renew-up.ps1
@@ -8,8 +8,10 @@ $action = New-ScheduledTaskAction -Execute 'Powershell.exe' -Argument '-NoProfil
 $delay = New-TimeSpan -Hours 12
 $triggerAM = New-ScheduledTaskTrigger -Daily -At 12am -RandomDelay $delay
 $triggerPM = New-ScheduledTaskTrigger -Daily -At 12pm -RandomDelay $delay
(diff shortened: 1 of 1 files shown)


adferrand:
Note, technically this PR overseeds #7536 since it solves also the problem handled by it, but #7536 is still interesting to merge because it ensures that the user is correctly resolved for other (including `SYSTEM`) exotic accounts.

bmw:
Both of these are probably just mistakes on my end and I can play with it more tomorrow but I struggled to test this. Problems I had:

1. I tried just running the code to get the principal/adminGID in an interactive shell or script to make sure it looked reasonable but the code crashed on the `$adminSID.Translate` line.
2. I tried running the installer and looking in the task scheduler for the task but it wasn't there.

If you have any immediate ideas, let me know!

adferrand:
No, it was not you, I made a mistake. It is fixed now.

PR #7541, opened 2019-11-12
Fix docstring

bmw:
All of Certbot's files used to be in a `client` directory. We've come across a few references to this during our renaming such as https://github.com/certbot/certbot/pull/7533#discussion_r344846647. I noticed this one when reviewing #7528 and wanted to fix it.


certbot/tests/plugins/common_test.py
```diff
@@ -177,7 +177,7 @@ def test_current_file_hash_in_all_hashes(self):
 
 
 class AddrTest(unittest.TestCase):
-    """Tests for certbot._internal.client.plugins.common.Addr."""
+    """Tests for certbot._internal.plugins.common.Addr."""
```
m0namon: I think this is located in `certbot.plugins.common` now!

bmw: Ah you're right! Nice catch.

m0namon:
(thanks for the bump, @adferrand ! i missed this in my notifs)

PR #7543, opened 2019-11-12
Internalize modules called by internal plugins

ohemorange:
Part of #5775. Methodology similar to #7528. These modules, `cli` and `hooks`, are used by the `webroot` and `manual` plugins respectively. Since these plugins are internal to Certbot their implementations can stay the same, and therefore these modules should be private.


bmw:
Ah I can't resolve these new conflicts through the web UI. If you fix them, I'll approve and merge though.

PR #7544, opened 2019-11-12
Refactor certbot/ and certbot/tests/ to use the same structure as the other packages

ohemorange:
Summary of changes in this PR:
- Refactor files involved in the `certbot` module to be of a similar structure to every other package; that is, inside a directory inside the main repo root (see below).
- Make repo root README symlink to `certbot` README.
- Pull tests outside of the distributed module.
- Make `certbot/tests` not be a module so that `certbot` isn't added to Python's path for module discovery.
- Remove `--pyargs` from test calls, and make sure to call tests from repo root since without `--pyargs`, `pytest` takes directory names rather than package names as arguments.
- Replace mentions of `.` with `certbot` when referring to packages to install, usually editably.
- Clean up some unused code around executing tests in a different directory.
- Create public shim around main and make that the entry point.

New directory structure summary:
```
repo root ("certbot", probably, but for clarity all files I mention are relative to here)
├── certbot
│   ├── setup.py
│   ├── certbot
│   │   ├── __init__.py
│   │   ├── achallenges.py
│   │   ├── _internal
│   │   │   ├── __init__.py
│   │   │   ├── account.py
│   │   │   ├── ...
│   │   ├── ...
│   ├── tests
│   │   ├── account_test.py
│   │   ├── display
│   │   │   ├── __init__.py
│   │   │   ├── ...
│   │   ├── ... # note no __init__.py at this level
│   ├── ...
├── acme
│   ├── ...
├── certbot-apache
│   ├── ...
├── ...
```

`certbot-docker` may also need to be updated based on these changes. There's a `.` at https://github.com/certbot-docker/certbot-docker/blob/master/core/Dockerfile#L51 that may need to be a `certbot`, but since the script fetches tags from github this will be tricky to confirm before release time.


adferrand:
Hello @ohemorange !

The biggest piece of work to do to make things work here, is to change in all places where a typical `pip install -e .` is invoked, into a `pip install -e certbot`.

Arguably these changes are tricky to do through a script, since the dot is certainly one of the most used character in a text ;), and some of these invocations are hidden in the code, with dynamic conversion of the `certbot` keyword into `.` when it was necessary.

Here are some regex sed operations that you can apply on all the Git repo to get rid of most of the dots:
* `s/-e(\s+)\./-e\1certbot/g`
* `s/\.\[dev\]/certbot[dev]/g`
* `s/\.\[dev3\]/certbot[dev3]/g`

There is also a replacement of `certbot` into `.` done in `install_and_test.py` to remove:
https://github.com/certbot/certbot/blob/4792e1ee217eda2d7a2e2a1b0e5c4da9a1892841/tools/install_and_test.py#L44-L45

After this, most of the errors would vanish, and you can proceed to the last manual fixes.

I anticipate also a migration to do once this PR is merged. It is for `certbot/website`, since the `docs` folder path of `certbot/certbot` changes, and so it is not in the same location in the submodule checked out by `certbot/website`.

adferrand:
Additional remarks:
* a new README should be created for the GIT repository itself
* certainly the same for LICENSE, since there is still some significant assets outside the python projects themselves
* `certbot/MANIFEST.in` should be changed to reflect the new organization of files: in particular `linter_plugin.py` and `CONTRIBUTING.md` stay at the top level

bmw:
> certbot/MANIFEST.in should be changed to reflect the new organization of files: in particular linter_plugin.py and CONTRIBUTING.md stay at the top level

For this, I think we should just remove the files from `MANIFEST.in`. I talked to Adrien about this out of band and he agreed.

ohemorange:
> a new README should be created for the GIT repository itself

Looks like most of the content in the original README is about the project as a whole. I actually think the original README should be moved back into the main directory, and do something like one of these:
- make one of these locations (main folder, inside `certbot`/) symlink to the other
- have a shorter README inside `certbot/` more like the ones in other packages

adferrand:
> > a new README should be created for the GIT repository itself
> 
> Looks like most of the content in the original README is about the project as a whole. I actually think the original README should be moved back into the main directory, and do something like one of these:
> 
>     * make one of these locations (main folder, inside `certbot`/) symlink to the other
> 
>     * have a shorter README inside `certbot/` more like the ones in other packages
> 
> 
> UPDATE: I did the former, happy to change.
> UPDATE 2: Python does not like the symlink.
> UPDATE 3: The other way around seems to be working better.

Indeed for GitHub let the Readme at its original place make a lot of sense. But the Readme associated to the Certbot python project (at the root directory, under certbot directory now) is used to generate the description of the certbot package in PyPI.

I think also that the original place is better. But in this case, we should craft a Readme under `certbot` directory that is suitable for PyPI.

ohemorange:
next mystery (that I will look at tomorrow):
`py27-cover` fails because tests fail to import `_internal.constants` when run with `-m pytest`

```
$ python -m pytest certbot-apache/certbot_apache/tests/parser_test.py
====================================================================== test session starts ======================================================================
platform linux2 -- Python 2.7.15+, pytest-3.2.5, py-1.8.0, pluggy-0.4.0
rootdir: /home/erica/certbot, inifile: pytest.ini
plugins: xdist-1.22.5, forked-0.2, cov-2.5.1
collected 0 items / 1 errors                                                                                                                                     

============================================================================ ERRORS =============================================================================
______________________________________________ ERROR collecting certbot-apache/certbot_apache/tests/parser_test.py ______________________________________________
ImportError while importing test module '/home/erica/certbot/certbot-apache/certbot_apache/tests/parser_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
certbot-apache/certbot_apache/tests/parser_test.py:11: in <module>
    from certbot_apache.tests import util
certbot-apache/certbot_apache/tests/util.py:13: in <module>
    from certbot.plugins import common
/home/erica/certbot/certbot/plugins/common.py:18: in <module>
    ???
/home/erica/certbot/certbot/crypto_util.py:30: in <module>
    ???
certbot/certbot/util.py:22: in <module>
    from certbot._internal import constants
E   ImportError: cannot import name constants
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
==================================================================== 1 error in 0.36 seconds ====================================================================
$ python  certbot-apache/certbot_apache/tests/parser_test.py
.......No handlers could be found for logger "certbot_apache.parser"
...................
Ran 26 tests in 5.880s

OK
```

I'm clicking "ready for review" so tests will run, in case there's something I'm missing on my local machine. Doesn't mean this is actually ready.

ohemorange:
Full test suite passing here: https://travis-ci.com/certbot/certbot/builds/136598848

ohemorange:
The release script is now working. A note about `--pyargs`, because I had to fix it up here and in other places:

Previously, we were using `pytest --pyargs` to call our tests. When run this way, it takes a module name as argument. The certbot tests are no longer part of a module, so cannot be run this way. In practice this means that we don't get the parallelization that `--pyargs` purportedly gave, though whether this was happening in practice is unclear to me. More importantly, it means that you need to reference the folder containing the module rather than the name of the module etc (e.g. `certbot-apache` rather than `certbot_apache`), and you're calling a path so need to make sure you're actually in the right directory to do so.

The reason for doing this was to be able to easily call certbot tests from other tests. If you have an `__init__.py` in there and make it a module, they don't like calling their sibling tests (see [here](http://doc.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code) for more info about this).

If we really want to switch back to using `--pyargs`, the steps to take are:
1. Replace calls to `--pyargs` and associated module vs. filenames, and probably undo directory changes.
2. Add `__init__.py` to `certbot/tests`
3. then AT LEAST ONE OF:
  a. refactor `cert_manager_test.py` to not call into `storage_test.py`, and/or
  b. figure out how to get tests calling each other in this sort of directory setup (probably possible, may take some amount of editing `PYTHONPATH` trickery)

There's also the separate but related question that we tabled earlier about `storage.py` more broadly; since it's an internal module, should plugins really be taking `RenewableCert`s as arguments? Do we need to fix this before 1.0?

ohemorange:
> On that matter however, you said that you tested the release process: in what extend? Because I suppose the last parts about effectively pushing a new release to PyPI cannot be tested easily. It would be interesting to see also if the Docker release process from `certbot/certbot-docker` is working correctly with this refactoring.

Correct, I tested up to but not including that. I also did not test offline signing, but I looked at the code and will be surprised if that step is broken by this change. Testing the Docker release process is also a good idea.

> That said, it bothers me that `certbot.test` is not a package, on the contrary to the other test packages in the repository, and in violation with the [Pytest good practices](http://doc.pytest.org/en/latest/goodpractices.html).

The link that you sent was the first place I looked, and if you read carefully you'll notice that it actually does not recommend making tests a package as a good practice; rather, it's saying to put modules themselves in subdirectories, and that's particularly important if you choose to set up tests as modules (they use the word "might"). This doesn't mean we necessarily *don't* want to do this, just that we aren't violating good practices if we don't.

Given that this is the first package in the repository to have both code and test with the new directory structure, I don't think we can rely on prior arrangement as an argument.

Which is all to say: while I could go either way, I don't think these specific arguments are particularly compelling.

> As you said, the root cause was that `cert_manager_test.py` uses material from `storage_test.py`: this should be fixed. Indeed in my opinion a test module should not rely to another test module, in principle of isolation in test assets.

This is the root of the question I am asking here, yes. I think there's a particular question about storage.py and how we treat it through the existing codebase that makes this a thorny question. Basically, "refactor certbot and its plugins to not rely as explicitly on `RenewableCert`s" is almost certainly the best answer in terms of technically correct code; the question is about whether it's also the best answer in practical terms to fix right now.

> If you do not want to fix that in this PR, would the tests in this PR still work if you add `__init__.py` to `certbot/tests` directory, and postpone the isolation of `cert_manager_test.py` to a future PR (I would be pleased to do it)?

They didn't when I tried -- I initially had `__init__.py` in there, and the easiest way to get them working was to remove it. There are other ways to get them working that don't involve a refactor, this is just the first/easiest one I found.

adferrand:
> > On that matter however, you said that you tested the release process: in what extend? Because I suppose the last parts about effectively pushing a new release to PyPI cannot be tested easily. It would be interesting to see also if the Docker release process from `certbot/certbot-docker` is working correctly with this refactoring.
> 
> Correct, I tested up to but not including that. I also did not test offline signing, but I looked at the code and will be surprised if that step is broken by this change. Testing the Docker release process is also a good idea.
> 
> > That said, it bothers me that `certbot.test` is not a package, on the contrary to the other test packages in the repository, and in violation with the [Pytest good practices](http://doc.pytest.org/en/latest/goodpractices.html).
> 
> The link that you sent was the first place I looked, and if you read carefully you'll notice that it actually does not recommend making tests a package as a good practice; rather, it's saying to put modules themselves in subdirectories, and that's particularly important if you choose to set up tests as modules (they use the word "might"). This doesn't mean we necessarily _don't_ want to do this, just that we aren't violating good practices if we don't.

You are totally right! Indeed this is not breaking a good practice, I was not sufficiently carefull while reading this.

But still, there is some inconsistencies that we need to resolve.

First, I see in the good practices that making `tests` a package can help to avoid name collision since without it, all scripts in `test` will be imported as top-level problem. I think it is unlikely to be a problem, but it is still a potentiality.

Second, if `test` as no `__init__.py`, its subdirectories `compat` and `display` have it.

Third, what about Python 3 (and this is totally unclear on the best practices)? Indeed, Python 3 does not need `__init__.py` to consider a directory as package. I would assume that it will still see here a `certbot.test` package.

For these various reasons, I still think we should explicitly turn `test` into a package.

ohemorange:
> Python 3 does not need __init__.py to consider a directory as package. I would assume that it will still see here a certbot.test package.

This is surprising! I would not expect the `import storage_test` to be working in that case, thank you for pointing it out. Actually it's surprising on several levels because why would it consider `certbot.test` a package but still call the other thing `certbot` rather than `certbot.certbot`, or conversely why use `certbot.tests` instead of `tests`? Seems like it must be doing something clever here, I will investigate further.

ohemorange:
Hitting some errors that I will continue investigating later; no need to continue the module vs. not module discussion because it may not matter.

Saving error here as a note to myself, this is not a request for help investigating.

```
certbot-apache/certbot_apache/tests/http_01_test.py:12: in <module>
    from certbot.tests import acme_util
certbot/tests/acme_util.py:15: in <module>
    ???
certbot/tests/util.py:43: in load_vector
    ???
venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py:1155: in resource_string
    self, resource_name
venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py:1399: in get_resource_string
    return self._get(self._fn(self.module_path, resource_name))
venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py:1522: in _get
    with open(path, 'rb') as stream:
E   IOError: [Errno 2] No such file or directory: 'certbot/tests/testdata/rsa512_key.pem'
```

ohemorange:
Ok so found the source of many weird inconsistencies and bugs: lots of `.pyc`s and `__pycache__`s sitting around in my local repo 🤦.

Next, to clear up several misconceptions:

- Python 3 supports implicit namespaces packages (i.e., no need for `__init__.py`s) *if you explicitly ask it to using `find_namespace_packages`*. Certbot does not do that; we instead use the old `find_packages`. And this is probably best to keep this way for several reasons that seem obvious to me but I am happy to spell out if needed, but in particular because:
- In the new directory structure, `certbot.tests` is the *public* tests module, located at `certbot/certbot/tests`. The *internal* tests module, located at `certbot/tests`, is the one that I have been talking about removing `__init__.py` from. Whether or not it has an `__init__.py`, we could not refer to `storage_test.py` as `certbot.tests.storage_test` in the proposed layout.
- If we did want `certbot/tests` to be a module, the correct choice is actually to create *another* layer of indirection, as the pytest good practices guide explains in the paragraph containing the words "now this introduces a subtle problem". Specifically, the way to hit this "sublte problem" is to have an `__init__.py` in the **top-level** `certbot/tests` folder. So, I feel strongly that we should not do that, and instead avoid hitting this problem.
- Now, I do agree that we don't want to have the name collision problem you mentioned. Specifically, we want both `certbot/tests` and `certbot/tests/display` to contain a `util_test.py`, as that matches up with our prior practices. Luckily, the current situation supports this! That is, within `certbot/tests`, you can do things like `import display.util_test`, and pytest is all good on its understanding of which tests to include and which are different and everything. And this understanding is limited to within the `certbot/tests` tests, which is exactly what we want.

On a separate but related note, I realized that I missed removing one last call using `--pyargs`. Basically that had the effect of only running a subset of the certbot tests (the public ones); it should be fixed now. As a side effect of this, I undid the changes in #6017 and cleaned up some now-dead code. For basically the same module-finding reason I mentioned above, the problem mentioned in #6017 is no longer an issue now that `certbot` is a directory lower.

I still have some more tests to run, and the test-farm tests to fix. I am posting this update and submitting a PR because I am leaving and do not believe in allowing hardware failure to cause me to lose work, not because I am inviting discussion at this time or because I am requesting assistance. I would prefer to address the remaining items before continuing discussion of `certbot/tests` being a module.

1. [x] remove pyargs
2. [x] run tests
3. [...] run oldest tests
4. [ ] get test farm tests working
5. [ ] test docker build

adferrand:
Juste one correction around the behavior you give for Python 3.

Python 3 do resolves implicit namespace packages (without `__init__.py`) that are in its current `sys.path` without any explicit action. What you are saying concerns the packaging: `setuptools` provides `find_packages` to resolve only `pkgutil`-style packages (with `__init__.py`) and `find_namespace_packages` to resolve both `pkgutil`-style and implicit namespace packages.

ohemorange:
> that are in its current `sys.path`

sounds like this still shouldn't affect us, then.

ohemorange:
Test farm tests are now passing.

Full test suite passing at https://travis-ci.com/certbot/certbot/builds/137479648

As for docker, it fetches the public tags from github, which seems difficult to test before having pushed the tags to github. But I did grep for `acme`, and it seems like the one place that might need a certbot added is here: https://github.com/certbot-docker/certbot-docker/blob/master/core/Dockerfile#L51.

That will obviously need to be a separate PR though, since it's not even in this repo, and I think we shouldn't block this PR on getting that working.

acme/readthedocs.org.requirements.txt
```diff
@@ -1,10 +1,10 @@
 # readthedocs.org gives no way to change the install command to "pip
-# install -e .[docs]" (that would in turn install documentation
+# install -e certbot[docs]" (that would in turn install documentation
```
bmw: I think we should either change these comments back or correct them to `acme[docs]`, `certbot-apache[docs]`, etc. Only  Certbot's version of this file runs `pip install -e .[docs]`. I wonder if `.` was meant as a shorthand for the current package.

ohemorange:
Thanks for the review!

> Let me know if you want to take that stuff or if we should find someone else to take a crack at it.

>  Update the submodule on the website repo and fix building the docs so they don't suddenly fail the next time we try to update the submodule.

I can easily update the submodule, does the second clause of this mean "fix the underlying problem that's causing it to keep breaking when we don't update the submodule right away"?

> Fix the certbot-docker repo before the next release.

I could do this, but might take me a bit to get the testing set up, since it's not entirely straightforward to do without having pushed the released tags to Certbot.

> Include the tests in the built package and make sure they run with python setup.py test. This should be as easy as adding the line recursive-include tests *.py to MANIFEST.in

Just to check, is this still the correct test given #7585?

bmw:
> does the second clause of this mean "fix the underlying problem that's causing it to keep breaking when we don't update the submodule right away"?

No. I believe the problem you're describing is caused by the docs by default including "Edit on GitHub" links back to the source files which break when the source file is deleted in this repo and the website repo tests fail if any links are broken. It's annoying, but I personally am not super worried about that. We should be almost done deleting API documentation.

The problem I was referring to is that the location of Certbot's docs which are hosted at https://certbot.eff.org/docs/ is going to move inside the submodule to be under the `certbot` directory. Quickly taking a look, I think it's just https://github.com/certbot/website/blob/master/_docs.sh and https://github.com/certbot/website/blob/master/_gulp/tasks/docs.js that have to be updated.

> I could do this, but might take me a bit to get the testing set up, since it's not entirely straightforward to do without having pushed the released tags to Certbot.

OK. I created https://github.com/certbot/certbot/issues/7588 and I'll see if someone else can take it.

> Just to check, is this still the correct test given #7585?

I think so. I don't think there's anything wrong with the proposed `MANIFEST.in` change to include our tests in the distribution and despite `python setup.py test` being recently deprecated, I think that's how most of our packagers are going to run tests and I expect it'll be quite a while for everyone to change to a new approach.

bmw:
Another thing I think we should check after this PR lands is that builds still work at https://letsencrypt.readthedocs.io/en/latest/. I added this to the checklist above.

README.rst
```diff
@@ -1,131 +0,0 @@
-.. This file contains a series of comments that are used to include sections of this README in other files. Do not modify these comments unless you know what you are doing. tag:intro-begin
```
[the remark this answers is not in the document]
ohemorange: Great, that's exactly what I would hope the behavior would be -- if I go directly to the file I'd want to see that it's a symlink. Yes perhaps we can improve this in the future by separating out which information we actually want where, but for now, glad to hear you agree it's ok.

certbot/certbot/main.py
```diff
@@ -0,0 +1,20 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def main(*args, **kwargs):
+    """Shim around internal main function"""
+    return internal_main.main(*args, **kwargs)
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
ohemorange: perfect, I love deleting code.

certbot/certbot/main.py
```diff
@@ -0,0 +1,20 @@
...
+import logging.handlers
+import sys
+
+from certbot._internal import main as internal_main
+
+
+logger = logging.getLogger(__name__)
+
+
+def main(*args, **kwargs):
```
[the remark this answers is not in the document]
ohemorange: works for me

certbot/certbot/main.py
```diff
@@ -0,0 +1,20 @@
+"""Certbot main public entry point."""
```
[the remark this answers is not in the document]
ohemorange: done

acme/readthedocs.org.requirements.txt
ohemorange: Weird. My guess is actually that the text was copied from certbot's version, and no one changed it. I think it makes most sense to correct them to the actual package name.

ohemorange: That being said, the latter actually isn't true -- the dns plugins did run `pip install -e .[docs]`. I'm still convinced of my explanation for why it's like this, and think that the correction is still accurate.

ohemorange:
Ah, I think it ate my comments -- when modifying `MANIFEST.in`, I noticed the lines `include CHANGELOG.md` and `recursive-include examples *`. For the latter, I think we do just want to move `examples` inside. For the former, I'm thinking do the same move-and-symlink that we did for the `README`, but I don't have a strong opinion.

ohemorange:
Aaaand just in case, here's a full test farm run: https://travis-ci.com/certbot/certbot/builds/137870262

certbot/certbot/main.py
```diff
@@ -0,0 +1,14 @@
...
+from certbot._internal import main as internal_main
+
+
+def main(cli_args=None):
+    """Command line argument parsing and main script execution.
```
bmw: I'd like to improve these docs. (Should `cli_args` include `sys.argv[0]`? What return values can be expected?)

Since this was just copied from the internal function though, I'm fine doing it later.

ohemorange: I literally don't know the answers to those questions, since, as you correctly identified, I just copied it from the existing comment.

PR #7549, opened 2019-11-12
Remove python2 and certbot-auto references in Certbot dev setup instructions

### LINKED ISSUE #7548: Default developer instructions should work in py3-only environment
Right now the developer instructions say to run:

```
./certbot-auto --debug --os-packages-only
```

Right now on Ubuntu, that will install `python` and `python-dev` (the Python 2 packages). However, latest Ubuntu (19.10) ships with `python3` and `python3-dev`, but no Python 2. Since we fully support Python 3, we shouldn't encourage people to install older Pythons. That's likely to mask issues with Python 3 support that we otherwise might find during development.

Since the list of packages in `--os-packages-only` has been static for a while, I recommend just hoisting the list of packages into the developer instructions, and replacing `python` and `python-dev` with `python3` and `python3-dev`. This avoids risks in changing certbot-auto.

adferrand:
Fixes #7548.

This PR udpdates installation instructions to get rid of python2 and certbot-auto in the how-to explaining the Certbot development environment setup.

Instead, Python 3 is used, and appropriate instructions for APT and RPM based distributions are provided.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -36,29 +36,36 @@ run Certbot in Docker. You can find instructions for how to do this :ref:`here
 install dependencies and set up a virtual environment where you can run
 Certbot.
 
+Install the OS system dependencies required to run Certbot.
+
+.. code-block:: shell
+
+   # For APT-based distributions (e.g. Debian, Ubuntu ...)
+   sudo apt update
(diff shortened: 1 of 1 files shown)


bmw:
LGTM too! Thanks for the nudge here @jsha.

PR #7554, opened 2019-11-13
Configure lint and cover to run on Python 3.8

adferrand:
This PR updates Travis configuration to run linting and coverage on Python 3.8.

NB: This PR is a part of a global action to remove Python 2 from our CI.


adferrand:
OK, this will not work, current set of pylint + astroid is not a good fit for Python 3. We need to upgrade it alongside with running it under Python 3.

PR #7559, opened 2019-11-14
[Apache v2] Initial ApacheParser skeleton

sydneyli:
Adding an empty implementation here so we can start filling it out once the appropriate interfaces land upstream in `apacheconfig`.

Following @joohoi 's parallel footsteps in Augeas land.

Are the names `apacheparser.py`/`ApacheParserNode` confusing? Let me know if you have suggestions here.

### DIFF
--- a/certbot-apache/certbot_apache/tests/augeasnode_test.py
+++ b/certbot-apache/certbot_apache/tests/augeasnode_test.py
@@ -110,7 +110,7 @@ class AugeasParserNodeTest(util.ApacheTest):  # pylint: disable=too-many-public-
                 name=servernames[0].name,
                 parameters=["test", "setting", "these"],
                 ancestor=assertions.PASS,
-                metadata=servernames[0].metadata
+                metadata=servernames[0].primary.metadata
             )
             self.assertTrue(mock_set.called)
             self.assertEqual(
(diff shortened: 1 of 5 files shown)


certbot-apache/certbot_apache/tests/augeasnode_test.py
```diff
@@ -145,11 +145,11 @@ def test_set_parameters_delete(self):
...
 
     def test_delete_child(self):
-        listens = self.config.parser_root.find_directives("Listen")
+        listens = self.config.parser_root.primary.find_directives("Listen")
         self.assertEqual(len(listens), 1)
         self.config.parser_root.primary.delete_child(listens[0])
 
-        listens = self.config.parser_root.find_directives("Listen")
+        listens = self.config.parser_root.primary.find_directives("Listen")
         self.assertEqual(len(listens), 0)
```
sydneyli: I changed these to only interact with the `primary` because the other implementation will not be able to `delete_child` on a non-shallow child, so the checks in `dualparser.py` will fail.

joohoi: That's fine, thanks for clearing that up.

PR #7561, opened 2019-11-14
[Apache v2] Implement find_ancestors

joohoi:
This PR implements find_ancestors that's needed by the Augeas ParserNode implementation to traverse ancestors where possible. While there's quite a few lines changed, majority of it is just moving few helper functions from `AugeasBlockNode` to `AugeasParserNode`, as all the node types need the `find_ancestors` capability.

Note: due to the nature of Augeas, we are not able to trace back the complete ancestry tree, for example in a VirtualHost file that's included from the root configuration, the Augeas path would be `/path/to/vhost.conf/Ancestor/VirtualHost`, leaving ancestor blocks that were defined in the main configuration invisible. This however isn't a regression from our current implementation in `master`, but apacheconfig will be able to do better here.

### DIFF
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -75,7 +75,6 @@ from certbot_apache import parser
 from certbot_apache import parsernode_util as util
 
 
-
 class AugeasParserNode(interfaces.ParserNode):
     """ Augeas implementation of ParserNode interface """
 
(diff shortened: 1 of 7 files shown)


certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -100,6 +99,65 @@ def __init__(self, **kwargs):
     def save(self, msg):
         self.parser.save(msg)
 
+    def find_ancestors(self, name):
+        """
+        Searches ancestor BlockNodes with a given name.
```
```suggestion
        Searches for ancestor BlockNodes with a given name.
```

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -100,6 +99,65 @@ def __init__(self, **kwargs):
...
+        AugeasParserNode.find_ancestors and AugeasBlockNode.
+        and AugeasBlockNode.find_blocks
+
+        """
+
+        name = self._aug_get_name(path)
+        metadata = {"augeasparser": self.parser, "augeaspath": path}
+
+        # Because of the dynamic nature, and the fact that we're not populating
+        # the complete ParserNode tree, we use the search parent as ancestor
```
ohemorange: I thought we removed this comment, but no time like the present to do so.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -100,6 +99,65 @@ def __init__(self, **kwargs):
...
+        :returns: List of matching ancestor nodes.
+        :rtype: list of AugeasBlockNode
+        """
+
+        ancestors = []
+
+        parent = self.metadata["augeaspath"]
+        while True:
+            # Get the path of ancestor node
+            parent = parent.rpartition("/")[0]
```
ohemorange: does augeas always specifically use "/" in an api-defined way, or should we actually be looking for something like `pathsep`?

joohoi: Yeah, Augeas always uses `/` in values returned from the searches as they always point to an absolute location.

PR #7562, opened 2019-11-14
[Apache v2] Implement parsed_files

joohoi:
Add new function to BlockNode interfaces: parsed_files(). This is needed in order for ApacheConfigurator to figure out if a file path (typically a VirtualHost that was added) is included in the main Apache configuration or if Certbot should enable it to make it available.

### DIFF
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -386,7 +386,7 @@ class AugeasBlockNode(AugeasDirectiveNode):
     def parsed_paths(self):
         """
         Returns a list of file paths that have currently been parsed into the parser
-        tree. The returned list  may include paths with wildcard characters, for
+        tree. The returned list may include paths with wildcard characters, for
         example: ['/etc/apache2/conf.d/*.load']
 
         This is typically called on the root node of the ParserNode tree.
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -384,7 +384,15 @@ class AugeasBlockNode(AugeasDirectiveNode):
         return self.parser.unsaved_files()
 
     def parsed_paths(self):
-        """Returns a list of parsed filepaths"""
+        """
+        Returns a list of file paths that have currently been parsed into the parser
+        tree. The returned list  may include paths with wildcard characters, for
+        example: ['/etc/apache2/conf.d/*.load']
+
+        This is typically called on the root node of the ParserNode tree.
+
+        :returns: list of file paths of files that have been parsed
+        """
 
         parsed_paths = self.parser.aug.match("/augeas/load/Httpd/incl")
         return [self.parser.aug.get(path) for path in parsed_paths]
--- a/certbot-apache/certbot_apache/augeasparser.py
+++ b/certbot-apache/certbot_apache/augeasparser.py
@@ -317,6 +317,12 @@ class AugeasBlockNode(AugeasDirectiveNode):
         """Returns a list of unsaved filepaths"""
         return self.parser.unsaved_files()
 
+    def parsed_paths(self):
+        """Returns a list of parsed filepaths"""
+
+        parsed_paths = self.parser.aug.match("/augeas/load/Httpd/incl")
+        return [self.parser.aug.get(path) for path in parsed_paths]
+
     def _create_commentnode(self, path):
         """Helper function to create a CommentNode from Augeas path"""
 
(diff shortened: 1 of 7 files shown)


sydneyli:
On the `apacheconfig` implementation side, how do you envision the root `BlockNode` determine what files are included? That is, on the initial pass, how should `include` `DirectiveNode`s bubble up the information that particular files have been parsed, etc. ?

For context, the current rather janky but working PoC (in branch `apacheconfig_tape`) uses an entry in `metadata` called `parsed_files`, which refers to a single global, mutable dictionary of files to tree roots. This is in order to:
 * prevent the same files from being parsed more than once (and ensure each file has a single tree)
 * perform recursive `find_*` operations: any `include` `DirectiveNode`s can look up its subtree

joohoi:
> On the apacheconfig implementation side, how do you envision the root BlockNode determine what files are included? That is, on the initial pass, how should include DirectiveNodes bubble up the information that particular files have been parsed, etc. ?

I think there's multiple ways to accomplish this. What I would personally think as being idiomatic to the ParserNode interface would be doing a deep iteration through the whole tree, and picking up unique `BlockNode.filepath`s to a set.

For the current implementation you have there, the mutable dictionary itself should be a good source of information as it needs to be kept up-to-date in order for the implementation to work in the first place anyway.

sydneyli:
OK, just checking. Two more clarifications: 
 * Does this search the entire tree or just a single config file, i.e. including `includes` within included files?
 * Is it necessary to return the wildcard path or directory path rather than the true list of files?

joohoi:
It's essentially everything that's parsed by the Apache, so it does include includes within includes :)
It should also return the include paths with wildcards, as it's used to determine if a file we create to a directory is already parsed by the Apache configuration or not.

sydneyli:
> It should also return the include paths with wildcards, as it's used to determine if a file we create to a directory is already parsed by the Apache configuration or not.

What i mean is, would it be possible to expand the wildcards and directories in the returned set? It sounds like that might still work to determine whether a particular filepath has already been parsed or not.

joohoi:
> What i mean is, would it be possible to expand the wildcards and directories in the returned set? It sounds like that might still work to determine whether a particular filepath has already been parsed or not.

I think it would, if we choose to take that apparoach however, we still need a way to differentiate if the includes in the configuration in filesystem covers a file added to the parser tree (when for example creating a new virtualhost config file for HTTPS vhost), or not.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -490,3 +490,15 @@ def unsaved_files(self):
...
         """
+
+    @abc.abstractmethod
+    def parsed_paths(self):
+        """
+        Returns a list of file paths that have currently been parsed into the parser
+        tree. The returned list  may include paths with wildcard characters, for
+        example: ['/etc/apache2/conf.d/*.load']
+
+        This is typically called for the root node of the ParserNode tree.
```
ohemorange:
```suggestion
        This is typically called on the root node of the ParserNode tree.
```

joohoi:
> My preference here is to be very specific about when we expect something to be a list of files and when we expect a wildcard.

With Augeas, I don't think we have a way to fetch the exact filepaths of each file included in the tree. It just stores the wildcard paths that we (or Apache configuration `Include` directive) asked it to include, in verbatim.

If we decide to go with strictly a single behavior here, it'd have to be wildcard paths.

certbot-apache/certbot_apache/augeasparser.py
```diff
@@ -383,6 +383,20 @@ def unsaved_files(self):
         """Returns a list of unsaved filepaths"""
         return self.parser.unsaved_files()
 
+    def parsed_paths(self):
+        """
+        Returns a list of file paths that have currently been parsed into the parser
+        tree. The returned list  may include paths with wildcard characters, for
```
```suggestion
        tree. The returned list may include paths with wildcard characters, for
```

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -502,3 +502,15 @@ def unsaved_files(self):
         :returns: list of file paths of files that have been changed but not yet
             saved to disk.
         """
+
+    @abc.abstractmethod
+    def parsed_paths(self):
+        """
+        Returns a list of file paths that have currently been parsed into the parser
+        tree. The returned list  may include paths with wildcard characters, for
```
ohemorange:
```suggestion
        tree. The returned list may include paths with wildcard characters, for
```

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -289,3 +288,19 @@ def unsaved_files(self):
         assertions.assertEqualSimple(primary_files, secondary_files)
 
         return primary_files
+
+    def parsed_paths(self):
+        """
+        Returns a list of file paths that have currently been parsed into the parser
+        tree. The returned list  may include paths with wildcard characters, for
```
ohemorange:
```suggestion
        tree. The returned list may include paths with wildcard characters, for
```

PR #7564, opened 2019-11-15
[Apache v2] Implement ApacheConfigurator get_virtual_hosts()

joohoi:
This replaces #7398 as there are quite a few changes (mostly to other parts than the actual ApacheConfigurator logic though), so I guessed it'd probably be best to start from scratch. The diff looks a bit messy as this is built on top of #7561 and  #7562 but it'll clear up when those are taken care of.

This PR also implements runtime assertion against the legacy implementation for VirtualHost object creation to ensure that the returned lists match where it matters. The objects created by the old implementation have Augeas paths, and the ones created by the new one have ParserNode references, but those obviously aren't matched.

### DIFF
--- a/certbot-apache/certbot_apache/configurator.py
+++ b/certbot-apache/certbot_apache/configurator.py
@@ -202,7 +202,11 @@ class ApacheConfigurator(common.Installer):
         self._autohsts = {}  # type: Dict[str, Dict[str, Union[int, float]]]
         # Reverter save notes
         self.save_notes = ""
-
+        # Should we use ParserNode implementation instead of the old behavior
+        self.USE_PARSERNODE = False
+        # Saves the list of file paths that were parsed initially, and
+        # not added to parser tree by self.conf("vhost-root") for example.
+        self.parsed_paths = []  # type: List[str]
         # These will be set in the prepare function
         self._prepared = False
         self.parser = None
@@ -261,6 +265,7 @@ class ApacheConfigurator(common.Installer):
                    "augeaspath": self.parser.get_root_augpath(),
                    "ac_ast": None}
         self.parser_root = self.get_parsernode_root(pn_meta)
+        self.parsed_paths = self.parser_root.parsed_paths()
 
         # Check for errors in parsing files with Augeas
         self.parser.check_parsing_errors("httpd.aug")
@@ -897,6 +902,29 @@ class ApacheConfigurator(common.Installer):
         return vhost
 
     def get_virtual_hosts(self):
+        """
+        Temporary wrapper for legacy and ParserNode version for
+        get_virtual_hosts. This should be replaced with the ParserNode
+        implementation when ready.
+        """
+
+        v1_vhosts = self.get_virtual_hosts_v1()
+        v2_vhosts = self.get_virtual_hosts_v2()
+
+        for v1_vh in v1_vhosts:
+            found = False
+            for v2_vh in v2_vhosts:
+                if assertions.isEqualVirtualHost(v1_vh, v2_vh):
+                    found = True
+                    break
+            if not found:
+                raise AssertionError("Equivalent for {} was not found".format(v1_vh.path))
+
+        if self.USE_PARSERNODE:
+            return v2_vhosts
+        return v1_vhosts
+
+    def get_virtual_hosts_v1(self):
         """Returns list of virtual hosts found in the Apache configuration.
 
         :returns: List of :class:`~certbot_apache.obj.VirtualHost`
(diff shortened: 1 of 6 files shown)


certbot-apache/certbot_apache/configurator.py
```diff
@@ -202,6 +202,11 @@ def __init__(self, *args, **kwargs):
         self._autohsts = {}  # type: Dict[str, Dict[str, Union[int, float]]]
         # Reverter save notes
         self.save_notes = ""
+        # Should we use ParserNode implementation instead of the old behavior
+        self.USE_PARSERNODE = False
+        # Saves the list of file paths that were parsed initially, and
```
ohemorange: We're using "parsed initially" here. I assume that means "on augeas' initial passthrough". I assume the intent is to figure out which ones are enabled in the config that will actually be loaded by apache. In that case, we should check that we're setting things appropriately anytime we do some change that might modify this -- essentially, making sure it's set any time we do a full passthrough, whether by augeas or apacheparser.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -890,6 +895,28 @@ def _create_vhost(self, path):
...
+        """
+
+        v1_vhosts = self.get_virtual_hosts_v1()
+        v2_vhosts = self.get_virtual_hosts_v2()
+
+        for v1_vh in v1_vhosts:
+            found = False
+            for v2_vh in v2_vhosts:
+                if assertions.isEqualVirtualHost(v1_vh, v2_vh):
+                    found = True
```
ohemorange: probably want to add a `break` after this

certbot-apache/certbot_apache/configurator.py
```diff
@@ -942,6 +969,78 @@ def get_virtual_hosts(self):
...
+        addrs = set()
+        for param in node.parameters:
+            addrs.add(obj.Addr.fromstring(param))
+
+        is_ssl = False
+        sslengine = node.find_directives("SSLEngine")
+        if sslengine:
+            for directive in sslengine:
+                if directive.parameters[0].lower() == "on":
+                    is_ssl = True
```
ohemorange: probably want to add a `break` after this

certbot-apache/certbot_apache/configurator.py
```diff
@@ -942,6 +969,78 @@ def get_virtual_hosts(self):
...
+
+        # "SSLEngine on" might be set outside of <VirtualHost>
+        # Treat vhosts with port 443 as ssl vhosts
```
ohemorange: I know this is copied from the original. I would just like to note that I hate it, not least since it's missing an idiomatic break.

certbot-apache/certbot_apache/configurator.py
```diff
@@ -942,6 +969,78 @@ def get_virtual_hosts(self):
...
+        for addr in addrs:
+            if addr.get_port() == "443":
+                is_ssl = True
+
+        macro = False
+        # Check if the VirtualHost is contained in a mod_macro block
+        if node.find_ancestors("Macro"):
+            macro = True
+        vhost = obj.VirtualHost(
+            node.filepath, None, addrs, is_ssl, node.enabled, modmacro=macro, node=node
```
ohemorange: We really don't need the path? I know it's now unnecessary, but I'm surprised it doesn't cause test failures in legacy checks. I wonder if it makes sense to do something like putting `USE_PARSERNODE` in `constants.py` and running all the tests with it on. Or do we think we'll have removed any code referencing it before we're ready to run all tests with that set?

certbot-apache/certbot_apache/dualparser.py
```diff
@@ -18,10 +18,62 @@ def __getattr__(self, aname):
...
-            assertions.assertEqualSimple(firstval, secondval)
+        exclusions = [
+            # Metadata will inherently be different, as ApacheParserNode does
+            # not have Augeas paths and so on.
+            aname == "metadata",
+            callable(firstval)
+        ]
+        if not any(exclusions):
+            if not isinstance(firstval, bool):
+                # Assertion check does not work for bool like enabled
```
ohemorange: I don't see the problem here? If it's set to `PASS`, it won't be a boolean. Unless we're not setting booleans to `PASS`? Is this breaking a test somewhere?

certbot-apache/certbot_apache/parser.py
```diff
@@ -840,7 +841,12 @@ def parsed_in_current(self, filep):
         :returns: True if file is parsed in existing configuration tree
         :rtype: bool
         """
-        return self._parsed_by_parser_paths(filep, self.parser_paths)
+        paths = []
+        for directory in self.parser_paths:
```
ohemorange: since this code is duplicated, it seems best to keep it factored out as before, no?

certbot-apache/certbot_apache/configurator.py
joohoi: You are right. At the current state with Augeas parser maintaining a meaningful value here gets a bit cumbersome, and hence only the very first initial parse state is maintained for this purpose. Some of the reasons it hasn't historically been feasible to maintain the state for this variable are:
 - Autosaving anything not-yet-written-to-disk on `find_dir` as it also looks for new `Include` and `IncludeOptional` directives - causing a reload of Augeas parsing tree, and erasing all unsaved changes.
 - "Manually" added parse paths to include `VirtualHost` files from arbitrary location using `--apache-vhost-path`

This  should be changed in the future, but for the sake of sanity during the migration process, should probably be implemented in the current, naive way.

certbot-apache/certbot_apache/configurator.py
joohoi: Currently the `VirtualHost` objects created by `_v2` implementation are not used further down in the execution, but merely just asserted against the list of `VirtualHost`s returned by the legacy implementation, and then stowed away. With the continued refactoring, we get to use them when `self.USE_PARSERNODE` boolean gets flipped.

certbot-apache/certbot_apache/dualparser.py
joohoi: If it's a boolean value, like for `enabled` we don't know if we should let the single value to pass assertion or not. At this point we don't know if the value is set to PASS, and furthermore `anyothervalue` should pass assertion against another `anyothervalue` so we should be testing them regardless of being strings.

certbot-apache/certbot_apache/parser.py
joohoi: Yeah, sounds good. I created a helper function to reuse in both of the cases.

certbot-apache/certbot_apache/configurator.py
ohemorange: As we discussed offline, agreed to leave this for now and possibly fix once we've moved off of augeas.

ohemorange: As we discussed offline, agreed to leave this for now and possibly fix once we've moved off of augeas.

certbot-apache/certbot_apache/dualparser.py
ohemorange: Ah sorry meant to bring this up in discussion earlier -- still not quite sure I understand. Maybe an example might help?

joohoi: Sure, no problem. For most of the values, we want to handle the assertion as per `assertions.assertEqualSimple()` as its able to either pass `assert` for matching values, raise `AssertionError` for mismatches or omit the `assert` if one of the values has `PASS` in it. This works for strings and sequences: 
 - `"a_value"` vs. `"a_value"` - this should pass the `assert`
 - `"a_value"` vs. `"mismatch"` - this should raise `AssertionError`
 - `"a_value"` vs. `"PASS"` - should omit the assert, and not raise `AssertionError`
 - `["a_value", "another_value"]` vs. `["PASS"]` - should omit the assert, and not raise `AssertionError`

For booleans however, we have no means to signal to `assertEqualSimple()` that we intend to omit the `assert`, as the value might not be implemented yet. Furthermore, `assertions.isPass()` that gets called by `assertions.assertEqualSimple()` does `return PASS in value`, which obviously doesn't work for booleans.

That said, we could handle omitting the checks for booleans in `assertions.assertEqualSimple()` as well.

ohemorange: > For booleans however, we have no means to signal to `assertEqualSimple()` that we intend to omit the `assert`, as the value might not be implemented yet. 

This is the part that I'm getting hung up on. What do you mean by "as the value might not be implemented yet"? Isn't that what setting the value to `PASS` means?

That being said, `return PASS in value` not working for booleans is a great point, and I also do think I would be more comfortable omitting the check in `assertions.assertEqualSimple()`.

joohoi: > This is the part that I'm getting hung up on. What do you mean by "as the value might not be implemented yet"? Isn't that what setting the value to PASS means?

Yeah, however, if we had `Implementation_A` that had the `enabled` variable implemented and working, and `Implementation_B` which would set the (expected boolean value) to `PASS`, the check would raise a `TypeError` for `Implementation_A`.

ohemorange: That's just because of the `PASS in value` bit though, right? We could fix that.

joohoi: Yeah, that's the only reason, however we're still unable to signal the potential `assert` if we want it  to pass or not, so I'd probably skip changing the `isPass()` . When we have everything implemented and working in both parsers, we might revisit this decision and change it.

joohoi:
Changing the base branch back and forth to update the PR diff.

joohoi:
That didn't work, trying to close - reopen.

joohoi:
Thanks! I think I'll do it in its own PR before moving towards developing against `master`

PR #7569, opened 2019-11-20
[Apache v2] Move the apachectl parsing to apache_util

joohoi:
This PR moves the `Define`, `Include` and `LoadModule` parsing from Augeas specific `ApacheParser` to `apache_util.py` to allow apacheconfig implementation of ParserNode to use this information.

Additionally the mentioned data is added to ParserNode initialization metadata:

The metadata has a new key `apache_vars` that itself contains a dict with keys:
- `defines` a dict with each defined key - value pair parsed from the apachectl output
- `includes` a list of Included paths parsed from the apachectl output
- `modules` a list of modules parsed from the apachectl output

### DIFF
--- a/certbot-apache/certbot_apache/apache_util.py
+++ b/certbot-apache/certbot_apache/apache_util.py
@@ -4,8 +4,9 @@ import logging
 import re
 import subprocess
 
-from certbot import util
 from certbot import errors
+from certbot import util
+
 from certbot.compat import os
 
 logger = logging.getLogger(__name__)
@@ -142,8 +143,6 @@ def parse_defines(apachectl):
         parts = match.partition("=")
         variables[parts[0]] = parts[2]
 
-    if not variables:
-        return {}
     return variables
 
 
--- a/certbot-apache/certbot_apache/apache_util.py
+++ b/certbot-apache/certbot_apache/apache_util.py
@@ -1,8 +1,14 @@
 """ Utility functions for certbot-apache plugin """
 import binascii
+import logging
+import re
+import subprocess
 
 from certbot import util
+from certbot import errors
 from certbot.compat import os
+
+logger = logging.getLogger(__name__)
 
 
 def get_mod_deps(mod_name):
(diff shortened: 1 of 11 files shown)


certbot-apache/certbot_apache/apache_util.py
```diff
@@ -1,9 +1,15 @@
 """ Utility functions for certbot-apache plugin """
 import binascii
+import logging
+import re
+import subprocess
 
 from certbot import util
+from certbot import errors
```
ohemorange: nit: these imports are misalphabetized

certbot-apache/certbot_apache/apache_util.py
```diff
@@ -105,3 +111,117 @@ def parse_define_file(filepath, varname):
...
+    for match in matches:
+        if match.count("=") > 1:
+            logger.error("Unexpected number of equal signs in "
+                         "runtime config dump.")
+            raise errors.PluginError(
+                "Error parsing Apache runtime variables")
+        parts = match.partition("=")
+        variables[parts[0]] = parts[2]
+
+    if not variables:
```
ohemorange: isn't this redundant?

certbot-apache/certbot_apache/parser.py
```diff
@@ -291,33 +291,11 @@ def parse_modules(self):
 
     def update_runtime_variables(self):
         """Update Includes, Defines and Includes from httpd config dump data"""
-        self.update_defines()
+
+        self.variables = apache_util.parse_defines(self.configurator.option("ctl"))
```
ohemorange: I think these should be consistent. If the `update` methods aren't used anywhere else, we can inline all of them, but otherwise I think we should keep `update_defines` around.

certbot-apache/certbot_apache/apache_util.py
joohoi: Absolutely, removed!

certbot-apache/certbot_apache/parser.py
joohoi: Yeah, I agree. `update_defines()` is now reintroduced, even though as just as a wrapper method.

joohoi: Yeah, I agree. `update_defines()` is now reintroduced, even though it's just a wrapper method for the implementation above.

certbot-apache/certbot_apache/apache_util.py
joohoi: They should be fixed now

PR #7587, opened 2019-11-22
Implement a sunset mechanism in certbot-auto for systems not supported anymore

adferrand:
During review of #7519, we discovered that SCL was not available for 32 bits versions of CentOS 6, and so we could not migrate these uses on Python 3.6.

However, we need to handle that nicely, by giving a proper sunset mechanism that avoid certbot-auto to just break without warning. Also, it will be reused for Python 2.7 in general, then for every users as we will remove certbot-auto, so it is good to invest some time on it.

This PR implements that.

When certbot-auto detects a system that is not supported anymore:
* it will try to upgrade itself if a sufficient version of Python is available
* it will not invoke any bootstrap mechanism
* it will not delete any existing venv
* it will not install a new venv and so it will not upgrade certbot
* it will try to pass the arguments to the current installation of certbot, if it exists
* it will fail with an explanatory message if there is no installation of certbot
* it will print at each invocation a warning message about the situation

Of course this mechanism is enabled for 32 bits CentOS 6 OSes and affiliates.

NB: Modifications in `letsencrypt-auto.template` L750-778 are taken directly from #7519 to make certbot-auto skip self-upgrade instead of breaking when no suitable version of Python is available.


letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -333,6 +333,11 @@ elif [ -f /etc/redhat-release ]; then
 
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
 
+  if [ "$PYVER" -eq 26 -a "$RPM_DIST_NAME" != "rhel" -a $(arch) != 'x86_64' ]; then
```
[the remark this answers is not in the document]
adferrand: I honestly could not find any way to get a 32 bit version of scl on RHEL and affiliates. I it is available, it is confidential. So I am quite OK to include RHEL in the deprecation.

For arch, well, `uname -m` will do the job!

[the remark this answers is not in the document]
bmw: Nice. `uname -m` is indeed in the POSIX standard. Is its output reliable though, at least on RHEL 6 based systems? I tried to find something stating possible outputs but couldn't find anything.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -530,7 +542,19 @@ CheckPathPermissions() {
 UNLIKELY_EOF
 }
 
-if [ "$1" = "--le-auto-phase2" ]; then
+if [ "$1" = "--le-auto-phase2" -a "$DEPRECATED_OS" = 1 ]; then
```
bmw: nit: What do you think about moving this as a case of the existing phase 2 logic? I personally find it a little simpler to understand than adding a 3rd high level case and it deduplicates some code.

If we do this, at the end of the `DEPRECATED_OS = 1` case, we'll need to call `exit`.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -530,7 +542,19 @@ CheckPathPermissions() {
...
+  # Phase 2 damage control mode for deprecated OSes.
+  # In this situation, we bypass any bootstrap or certbot venv setup.
+  shift 1
+
+  error "Your system is not supported by certbot-auto anymore, and so certbot cannot be installed or upgraded."
```
bmw: What do you think about making the rest of the code in this case look something like:
```
  error "Your system is no longer supported by certbot-auto."
  if [ -f "$VENV_BIN/letsencrypt" ]; then 
    error "Certbot will no longer receive updates."
    error "Please visit https://certbot.eff.org/ to check for other alternatives."
    "$VENV_BIN/letsencrypt" "$@" 
    exit 0
  else 
    error "Certbot cannot be installed."
    error "Please visit https://certbot.eff.org/ to check for other alternatives."
    exit 1
  fi
```

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -530,7 +542,19 @@ CheckPathPermissions() {
...
+  error "Please consult official Certbot site (https://certbot.eff.org/) to check other alternatives."
```
bmw: If we point people here (which I think we should), I think we need to make sure we have something to tell these users.

Given the complexity of getting a new version of Python and OpenSSL and how few users are affected, I think we should probably just add a warning to the top of the CentOS and RHEL 6 instructions saying that only x86_64 systems are supported.

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
adferrand: I got `x86_64` for my machine for any system I could think of. All code source dumps I could find are stating `x86_64` for 64 bits, and `ix86` (with `x=3,4,5,6`) for 32 bits.

For a standard POSIX tool, I expect this behavior to be extremely consistent.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Done.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Ah. With your previous comment, I was hesitating on the exit code value depending on the situation (certbot installed or not). Yes, it makes sense, and I add that.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: You mean, in the website?

adferrand:
@bmw, I am very curious about how to trick `uname -m` or if there is other alternatives here to check this logic!

adferrand:
Oh, I have an idea! An alias to uname, set in a subshell like for venv environment variables setup!

adferrand:
> 
> 
> Oh, I have an idea! An alias to uname, set in a subshell like for venv environment variables setup!

No, not working, aliases are not transfered to child scripts...

adferrand:
I added tests to check the sunset mechanism.

Full tests suite is running at https://travis-ci.com/certbot/certbot/builds/139610387

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
bmw: I don't love this. Everything I've seen about `uname` says its output is implementation defined and looking at the source code it seems its output is completely controlled by the kernel where I was unable to find any guarantees about its output across systems/versions.

With that said, I'm not aware of any better options. I considered using Python (which I think we can rely on being there because its used by `yum`), but we can't just check if the machine is 64-bit because [RHEL 6 supports other 64-bit architectures](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/6.0_release_notes/index) and the other options I'm aware of all go back to `uname`.

I think this is _probably_ fine, but I don't like uncertainty like this especially when dealing with certbot-auto.

letsencrypt-auto-source/letsencrypt-auto.template
bmw: Yeah. I created https://github.com/certbot/website/issues/512 and https://github.com/certbot/certbot/issues/7632 to track this and I think we should get that PR ready before releasing this code.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -534,6 +546,24 @@ if [ "$1" = "--le-auto-phase2" ]; then
   # Phase 2: Create venv, install LE, and run.
 
   shift 1  # the --le-auto-phase2 arg
+
+  if [ "$DEPRECATED_OS" = 1 ]; then
+    # Phase 2 damage control mode for deprecated OSes.
+    # In this situation, we bypass any bootstrap or certbot venv setup.
+    error "Your system is not supported by certbot-auto anymore."
+
+    if [ -f "$VENV_BIN/letsencrypt" ]; then
```
bmw: I just noticed that this code breaks the `--install-only` flag on deprecated systems. To fix this, how about we also check `"$INSTALL_ONLY" != 1` here and go into the "Certbot cannot be installed" case.

I think it'd be good to add a simple test for this case where an older version of Certbot is already installed too.

bmw: Also, nit:

I think we should add a comment before the `INSTALL_ONLY` case below saying something like:

> If you're modifying any of the code after this point in this if block, you may need to update the "$DEPRECATED_OS" = 1 case at the beginning of phase 2 as well.

letsencrypt-auto-source/uname_wrapper
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/uname_wrapper
```diff
@@ -0,0 +1,9 @@
...
+
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Done.

letsencrypt-auto-source/letsencrypt-auto.template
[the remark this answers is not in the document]
adferrand: Me neither, in term of solution. But I never saw anything except `x86_64` for 64 bits on kernel architectures.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -688,6 +688,8 @@ UNLIKELY_EOF
   fi
 
   if [ "$INSTALL_ONLY" = 1 ]; then
+    # If you're modifying any of the code after this point in this `if` block, you may
+    # need to update the "$DEPRECATED_OS" = 1 case at the beginning of phase 2 as well.
```
bmw: I should have been more clear in my initial comment, but I was actually imagining we move this comment up a level and put it before the `if [ "$INSTALL_ONLY" = 1 ]; then` line. If we change anything after this point such as how we invoke Certbot, we probably need to update the deprecation case as well.

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -95,11 +95,11 @@ unset VENV_PATH
 # ensure CentOS6 32bits is not supported anymore, and so certbot is not upgraded
 export UNAME_FAKE_32BITS=true
 if ! "$LE_AUTO" 2>&1 | grep -q "Certbot will no longer receive updates."; then
-  echo "On CentOS 32 bits, certbot-auto installed certbot."
+  echo "On CentOS 32 bits, certbot-auto upgraded certbot."
   exit 1
 fi
```
bmw: Can we add the test for the `--install-only` case which I suggested at https://github.com/certbot/certbot/pull/7587#discussion_r354986394?

letsencrypt-auto-source/letsencrypt-auto.template
adferrand: Indeed, it can also impact in non install-only mode.

letsencrypt-auto-source/tests/centos6_tests.sh
adferrand: Done.

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -77,5 +90,22 @@ if [ "$($VENV_PATH/bin/python -V 2>&1 | cut -d" " -f2 | cut -d. -f1)" != 3 ]; th
...
+(
+# ensure CentOS6 32bits is not supported anymore, and so certbot
+# is not upgraded nor reinstalled.
+export UNAME_FAKE_32BITS=true
+if ! "$LE_AUTO" 2>&1 | grep -q "Certbot will no longer receive updates."; then
+  echo "On CentOS 32 bits, certbot-auto upgraded certbot."
+  exit 1
+fi
+if ! "$LE_AUTO" --install-only 2>&1 | grep -q "Certbot cannot be installed."; then
+  echo "On CentOS 32 bits, certbot-auto installed again certbot."
```
bmw:
```suggestion
  echo "On CentOS 32 bits, certbot-auto installed certbot again."
```

letsencrypt-auto-source/Dockerfile.centos6
```diff
@@ -30,6 +30,11 @@ RUN update-ca-trust
 # Copy code:
 COPY . /home/lea/certbot/letsencrypt-auto-source
 
+# Tweak uname binary for tests on fake 32bits
+COPY uname_wrapper.sh /bin
```
bmw: It looks like this path needs to be updated to since the file was moved under `tests`.

bmw:
`tox -e le_auto_centos6` is now failing with:
```
certbot/letsencrypt-auto-source/tests/centos6_tests.sh: line 31: error: command not found
Python upgrade failure warning not printed!
```
I'll take another look at this after a successful run of the nightly tests.

adferrand:
The full tests suite is passing here: https://travis-ci.com/certbot/certbot/builds/140511179

bmw:
I did a mock release of this running ran test farm tests and played with changing the version number before running `tox -e le_auto_centos6` and all tests passed.

This LGTM and is ready for a 2nd review.

bmw:
Oh! One aspect of this I forgot to leave a comment on is we should add a changelog entry saying that support for non-x86_64 variants of RHEL 6 based systems has been dropped and that use Certbot you should upgrade your OS.

Relatedly, I think the fact that the script no longer works on new RHEL 6 systems is a breaking change in terms of versioning. I think/hope we're going to be making changes like this a lot in certbot-auto over the next few months and bumping Certbot's version number and all of its components every time for this seems like overkill. Maybe we should make a separate section in our changelog or even a separate file for certbot-auto changes and I can update https://github.com/certbot/certbot/pull/7611 to say that certbot-auto is in the process of being deprecated and should not be considered stable. Here soon the only people that should be using it are x86_64 RHEL 6 users.

I'm going to think about this a bit more but if anyone has ideas, please share!

ohemorange:
To continue the high-level conversation I alluded to earlier today -- I think there are benefits to defining a last-supported-version for deprecated systems.

- New installs of certbot-auto with the latest version will just work.
- Idempotency
- Universally predictable behavior is easier to test
- If we decide for some esoteric reason in the future to ship an update to a deprecated system, we'll have an easier time (this is obviously unlikely)

The biggest change would be to change the logic of "if we're on a deprecated system, do not update certbot" to "if we're on a deprecated system, either update to the pinned version or do nothing if we're already sufficiently up-to-date".

adferrand:
I see the benefit of it, but I think it gives a wrong signal that on these deprecated systems are somehow still supported.

If it can make sense for already installed versions to push to the latest possible release, it does less for systems where certbot is not installed. Because in this case, certbot-auto still install something.

In general we would create a fork here : most systems are running latest version of Certbot, other are using on older one. What if the latest supported version  for these deprecated systems happens to have a security breach?

In general, I think we have to consider this sunset in the scope of the soon Python 2.7 EOL, and more generally of the certbot-auto decommissioning itself. This sunset, in my opinion, is to keep people progressively away from certbot-auto. The fact that it still runs certbot if installed is more a migration path, to no break systems where everything is automated.

adferrand:
To respond more specifically on your points here:
* about installation, well, as said I think the all point of the sunset is to not do it anymore
* about indempotency, in my opinion, here the idempotency is that certbot-auto is not doing anything on any deprecated system
* about testability, it is easier to test that nothing is done, instead of checking that a given behavior for each deprecated system now and in the future is matched
* about future modified behaviors, the expected decommissioning of this script makes that even less likely

I will add also that any new logic added in certbot-auto increase a lot the risk of a regression, that the previous experiences showed it could happen anywhere, despite all our efforts. The minimal viable logic is certainly the best strategy in these conditions.

ohemorange:
> In general we would create a fork here : most systems are running latest version of Certbot, other are using on older one.

Actually, this is exactly my worry. The change as-is forks into "latest current certbot" and "whatever version happened to be on the machine at the time". So rather than forking into two versions, it essentially forks less predictably into many versions, particularly if we actually expect people to grab an old certbot-auto and run it with `no-self-upgrade`.

adferrand:
> 
> 
> > In general we would create a fork here : most systems are running latest version of Certbot, other are using on older one.
> 
> Actually, this is exactly my worry. The change as-is forks into "latest current certbot" and "whatever version happened to be on the machine at the time". So rather than forking into two versions, it essentially forks less predictably into many versions, particularly if we actually expect people to grab an old certbot-auto and run it with `no-self-upgrade`.

I would not call that a fork, that would imply two or more coordinated states, but a fragmentation. This fragmentation of certbot version already exists because of systems that are not regularly update certbot. This is in fact these non-upgrading systems that worried you. Here the statu quo does not reduce the fragmentation nor it increases it. This is what I imply by using the minimal viable logic strategy.

adferrand:
But I think in general that you are misunderstanding the intention of this sunset. As we discussed with @bmw and @joohoi starting from here https://github.com/certbot/certbot/pull/7519#issuecomment-555695496, the point of this sunset is less about dropping in a safe zone and more to cease all operations in decent elegant manner.

Because there is no safe zone. This sunset is not because we decide to drop some functional environment, it is because we cannot get anymore a supported an testable environment for certbot and certbot-auto: here, for CentOS/RHEL 6 32 bits, no non-EOL version of Python, and OpenSSL version 1.0.1 support drop.

What you are asking here is to make certbot-auto install Certbot on a unsupported version of Python, and unsupported version of OpenSSL.

But things can be even funnier. Let's suppose that we allow upgrade of Certbot to a given version if it is already installed. Here in certbot-auto it implies to call several python code, in particular `fetch.py`. A given version of `fetch.py`, running on Python 3.4, could fail for any reason. In this case, and because certbot-auto is deleting the venv before starting the logic of reinstalling certbot, the previous certbot instance would be wiped out, making systems even less functional than what they would be with the current state of the sunset mechanism.

letsencrypt-auto-source/letsencrypt-auto.template
```diff
@@ -333,6 +333,11 @@ elif [ -f /etc/redhat-release ]; then
 
   RPM_DIST_NAME=`(. /etc/os-release 2> /dev/null && echo $ID) || echo "unknown"`
 
+  if [ "$PYVER" -eq 26 -a $(uname -m) != 'x86_64' ]; then
+    # 32 bits CentOS 6 and affiliates (except RHEL) are not supported anymore by certbot-auto.
```
bmw:
```suggestion
    # 32 bits CentOS 6 and affiliates are not supported anymore by certbot-auto.
```

letsencrypt-auto-source/Dockerfile.centos6
adferrand: Fixed.

adferrand:
> Oh! One aspect of this I forgot to leave a comment on is we should add a changelog entry saying that support for non-x86_64 variants of RHEL 6 based systems has been dropped and that to use Certbot you should upgrade your OS.
> 
> Relatedly, I think the fact that the script no longer works on new RHEL 6 systems is a breaking change in terms of versioning. I think/hope we're going to be making changes like this a lot in certbot-auto over the next few months and bumping Certbot's major version number and all of its components every time for this seems like overkill. Maybe we should make a separate section in our changelog or even a separate file for certbot-auto changes and I can update #7611 to say that certbot-auto is in the process of being deprecated and should not be considered stable. Here soon the only people that should be using it are x86_64 RHEL 6 users.
> 
> I'm going to think about this a bit more but if anyone has ideas, please share!

I agree that if certbot-auto is part of the assets that defines the version, then it is a 2.x. Your solution is OK for me, certbot-auto can be removed from it, and we can start to explain that certbot-auto is about to be deprecated.

adferrand:
I added a changelog entry.

certbot/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
-*
+* certbot-auto now supports only x86_64 architectures on CentOS/RHEL 6.
```
bmw: I think we should expand this. What about:
```
* certbot-auto no longer supports architectures other than x86_64 on RHEL 6
  based systems. Existing certbot-auto installations affected by this will
  still work, but they will no longer receive updates. To install a newer
  version of Certbot on these systems, you should update your OS.
```

adferrand: Done.

adferrand:
> Looks good to me. One thing I want to double-check is that we're intentionally (rather than accidentally) disregarding the existing `DeprecationBootstrap` mechanism, presumably to keep around the code that sets the Python version so we can continue to self-upgrade.

The `DeprecationBootstrap` on its own is not skipped explicitly, it is the entire bootstrap/rebootstrap feature that is intentionally skipped. Indeed we want to avoid this process to be triggered for two reasons:
* allow certbot-auto to continue to update itself if a suitable version of Python is available
* do not destroy the existing certbot installation (that is done normally during rebootstrap) and leave with a non-functional state, while before certbot (in its old version) was still operational

Specifically on `DeprecationBootstrap`, it may have some advantages to move the concerned distribution to the sunset mechanism, since it allows at least certbot-auto script to be upgrade. It will not change anything for the certbot installation, since there was none.

I suppose we could add this in a future PR, but our general approach of "let's touch certbot-auto as less as possible" would lead us to not change anything for the existing `DeprecationBootstrap` logic.

bmw:
> One thing I want to double-check is that we're intentionally (rather than accidentally) disregarding the existing DeprecationBootstrap mechanism

I had honestly forgotten about `DeprecationBootstrap` so nice catch there. I personally don't think we should use it though. I believe it works a little differently in that you can still bootstrap by setting `--debug` which certbot-auto won't be able to do anymore on non-x86_64 RHEL 6 systems and probably shouldn't do on other systems in the (near) future if the bootstrap script is going to install Python 2 but Certbot starts requiring Python 3. I also think we don't want to change this behavior for existing systems using `DeprecationBootstrap` at the same time as we drop non-x86_64 RHEL 6 support to limit the damage if we got something wrong and I personally don't think making the behavior conditional is worth it.

> Specifically on DeprecationBootstrap, it may have some advantages to move the concerned distribution to the sunset mechanism, since it allows at least certbot-auto script to be upgrade. It will not change anything for the certbot installation, since there was none.
>
> I suppose we could add this in a future PR, but our general approach of "let's touch certbot-auto as less as possible" would lead us to not change anything for the existing DeprecationBootstrap logic.

Yeah. I was imagining that in a future release we'll set `DEPRECATED_OS=1` in the various branches where `Bootstrap` and `BOOTSTRAP_VERSION` are set, but otherwise won't touch anything.

adferrand:
I am not confortable with the idea of trying now to modify another existing logic, and potentially introduced new unexpected errors. Current implementation has the advantage to execute in a dedicated branch, where we already thought a lot of potential errors, and so we are certainly closer to have a bug-free implementation.

In theory we have three situations: certbot not install, certbot installed in old location, certbot installed in new location. Is there any other possible situation, regarding history of certbot-auto?

bmw:
I think reusing the existing code to print a message and reuse the venv is more maintainable in the long run and has the benefit of being tested by hundreds of thousands of users, but I understand your point too. I don't really have a strong preference personally.

> Is there any other possible situation, regarding history of certbot-auto?

Not that I'm aware of, but I didn't remember this problem, I found it by reading through the code to see if I missed anything. Other people doing the same is probably a good idea.

> For instance, is it possible to have a user-defined location for certbot?

Yes, but this should be handled automatically by using the environment variable `VENV_BIN`:
https://github.com/certbot/certbot/blob/4a906484eeec0748f480dc4dca9ed7557d11ae56/certbot-auto#L26-L32

bmw:
As usual, I think it'd be good to add a test of this behavior to make sure we properly fixed it and don't break it in the future. This can probably be easily done by setting `VENV_PATH` to `${XDG_DATA_HOME:-~/.local/share}/letsencrypt`

adferrand:
During the development of this PR, I used the following pattern several times:
```bash
#!/bin/bash
# Protect against environment variable modifications
(
export TEST="test"
if [ "foo" != "bar" ]; then
  echo "Error!"
  exit 1
fi
echo "Success"
)
```

This pattern was already present in the test script, so it made sense to reuse it to enrich the tests here. Its intention is to make ephemeral modifications thanks to a subshell, execute some logic in it, and fails the script using a non zero exit code if the logic does not fullfill some expectation.

However this pattern is not working as expected: invoking `exit` in a subshell finishes the subshell, not the englobing shell (the test script here). Since `set -e` is not set in the englobing shell, the non zero exit code of the exited subshell does not fail the englobing shell.

In practice it was not a problem until now for the test logic itself, because a failed assertion was equivalent to a missing requirement for the success of any following assertions. But with my modifications it is not the case (we test several parallel and independent situations) and nonetheless it is not a clean approach.

To fix that, I did not put `set -e`, because I think it would make the test script less readable, since there is several commands that are expected to fail, and subshells to handle properly regarding what are the options set in them. Instead, I deal with this using a trigger file that makes the script fail on explicit checkpoint if it exists.

adferrand:
Full test suite passing here: https://travis-ci.com/certbot/certbot/builds/142492386

adferrand:
Relevant tests from the full tests suite are running here: https://travis-ci.com/certbot/certbot/builds/142931508

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -95,17 +88,38 @@ unset VENV_PATH
...
+  echo "On CentOS 6 32 bits, certbot-auto installed certbot again."
   exit 1
 fi
+)
 
-echo "PASSED: On CentOS6 32 bits, certbot-auto refuses to install/upgrade certbot."
+# we're going to modify env variables, so do this in a subshell
+(
+# Prepare a certbot installation in the old venv path
+export VENV_PATH=~/.local/share/letsencrypt
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -16,48 +17,42 @@ if ! "$LE_AUTO" 2>&1 | grep -q "Certbot cannot be installed."; then
...
-RESULT=$?
-if [ $RESULT -eq 0 ]; then
+if python3 --version 2> /dev/null; then
   echo "Python3 is already installed."
   exit 1
 fi
 
 # ensure python2.7 is available
 python2.7 --version 2> /dev/null
 RESULT=$?
```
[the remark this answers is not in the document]
adferrand: Done.

letsencrypt-auto-source/tests/centos6_tests.sh
```diff
@@ -95,17 +88,38 @@ unset VENV_PATH
 # ensure CentOS6 32bits is not supported anymore, and so certbot
 # is not upgraded nor reinstalled.
 export UNAME_FAKE_32BITS=true
-if ! "$LE_AUTO" 2>&1 | grep -q "Certbot will no longer receive updates."; then
-  echo "On CentOS 32 bits, certbot-auto upgraded certbot."
+if ! "$LE_AUTO" --version > /dev/null 2> /dev/null; then
+  echo "On CentOS 6 32 bits, certbot-auto failed to run installed certbot instance."
+  exit 1
+fi
+if ! "$LE_AUTO" --version 2>&1 | grep -q "Certbot will no longer receive updates."; then
```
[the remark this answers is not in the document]
adferrand: Indeed, it is better to test everything together. I used the `set -o pipefail` approach.

adferrand:
Full tests suite passing here: https://travis-ci.com/certbot/certbot/builds/142976996

ohemorange:
Did you change this back after making this comment? https://github.com/certbot/certbot/pull/7587#issuecomment-569348984

bmw:
> Did you change this back after making this comment? #7587 (comment)

Yes he did. Sorry. To speed things up, Adrien and I talked about this synchronously during our 1-1.

I thought using `set -e` rather than setting/checking custom files like this was cleaner, simpler, resulted in less code, and we probably should have been using `set -e` regardless since I think it's a good practice and is what we do in almost all of our other shell scripts.

bmw:
While reviewing #7519, I noticed this slightly awkward behavior on 32-bit CentOS 6:
```
# ./letsencrypt-auto --os-packages-only
Skipping bootstrap because certbot-auto is deprecated on this system.
OS packages installed.
# echo $?
0
```
This will happen on any deprecated system when `--os-packages-only` set. I think ideally certbot-auto would be a little more informative and exit with a non-zero status like we do when you run `certbot-auto --install-only`.

My initial reaction is this isn't worth fixing though.

Issue #7594, opened 2019-11-25
Remove interactive redirect ask


ohemorange:
Follow-up to #6960.

More information available here: https://docs.google.com/document/d/1aKnQYEzCrZgYX-iuE-ErOdIKERUuLmZSX_BDb7d0M6I/edit

PR #7595, opened 2019-11-25
Implement redirect by default

ohemorange:
Change redirect default to yes so that it happens automatically in noninteractive mode and remove error message.

As a follow-up, we should remove the interactive ask, controlling this feature only by a flag. Issue for this has been created [here](https://github.com/certbot/certbot/issues/7594), but doesn't need to be done for 1.0.

### DIFF
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -28,6 +28,9 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
   `certbot.reverter.Reverter.view_config_changes`, and
   `certbot.util.get_systemd_os_info` have been removed
 * Certbot's `register --update-registration` subcommand has been removed
+* When possible, default to automatically configuring the webserver so all requests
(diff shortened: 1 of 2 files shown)


certbot/certbot/_internal/client.py
```diff
@@ -557,11 +557,6 @@ def enhance_config(self, domains, chain_path, ask_redirect=True):
                 if ask_redirect:
                     if config_name == "redirect" and config_value is None:
                         config_value = enhancements.ask(enhancement_name)
-                        if not config_value:
```
bmw: I think we should leave this code. My understanding is the only time we will enter this branch is if the result of `enhancements.ask` is to not provide a redirect which I think is only possible if the user interactively asks for no redirect.

If we do this, we should probably keep the test.

ohemorange: Oh, good call, agreed.

ohemorange:
Thanks for the review! I think we should hold off on merging until test farm tests pass (currently doing a test release with this change, then will run test farm tests).

Issue #7540, opened 2019-11-12
Fix references to RenewableCert in storage/plugins


bmw:
One of our plugin interfaces takes a `RenewableCert`, however, we recently made the file containing that class private. This is awkward as there there is no guidance on how to use this parameter.

I personally don't want to make all of storage.py or even `RenewableCert` public because there's a lot of code there that I don't think will be used in plugins and making it public makes it harder for us to refactor it.

Instead, I was thinking we could make a simple zope interface in `interfaces.py` describing the attributes/methods one could expect this parameter would have. I would base this off of how the parameter is used in the Apache plugin. We'd then register `RenewableCert` as implementing this interface and always refer to the parameter in the plugin docs through this zope interface. We do something similar to this with `certbot.interfaces.IConfig`.

bmw:
I found one 3rd party plugin which uses these interfaces at https://github.com/dlapiduz/certbot-s3front/blob/master/certbot_s3front/installer.py#L165.

PR #7603, opened 2019-11-26
Defines the RenewableCert API

bmw:
This is my proposed fix for #7540. I would ideally like this to be included in our 1.0 release.

I came up with this design by adding all attributes used either in our own plugins, 3rd party plugins listed at https://certbot.eff.org/docs/using.html#third-party-plugins, or our public API code.

Despite me thinking that zope is unneeded nowadays, I initially tried to use it to define this interface since we have it and it gives us a way to define expected attributes, but it doesn't work because zope interface objects also have a method called `names` which conflict with the API.

~~For defining attributes on the object, I took the approach Joona and I came up with in https://github.com/certbot/certbot/pull/7246. The `*_path` attributes could be more naturally defined as properties which works, but I don't like doing this because I think whether it's an attribute or property is an implementation detail.~~

@adferrand or @joohoi, are you able to review this?

### DIFF
--- a/certbot/certbot/interfaces.py
+++ b/certbot/certbot/interfaces.py
@@ -534,28 +534,49 @@ class IReporter(zope.interface.Interface):
 
 @six.add_metaclass(abc.ABCMeta)
 class RenewableCert(object):
-    """Interface to a certificate lineage.
+    """Interface to a certificate lineage."""
 
-    RenewableCert objects have the following attributes:
+    @abc.abstractproperty
+    def cert_path(self):
+        """Path to the certificate file.
 
-    .. code-block:: python
+        :rtype: str
 
-        # Path to the certificate file.
-        cert_path: str
+        """
 
-        # Path to the private key file.
-        key_path: str
+    @abc.abstractproperty
+    def key_path(self):
+        """Path to the private key file.
 
-        # Path to the certificate chain file.
-        chain_path: str
+        :rtype: str
 
-        # Path to the full chain file (cert plus chain).
-        fullchain_path: str
+        """
 
-        # Name given to the certificate lineage
-        lineagename: str
+    @abc.abstractproperty
+    def chain_path(self):
+        """Path to the certificate chain file.
 
-    """
+        :rtype: str
+
+        """
+
+    @abc.abstractproperty
+    def fullchain_path(self):
+        """Path to the full chain file.
+
+        The full chain is the certificate file plus the chain file.
+
+        :rtype: str
+
+        """
+
+    @abc.abstractproperty
+    def lineagename(self):
+        """Name given to the certificate lineage.
+
+        :rtype: str
+
+        """
 
     @abc.abstractmethod
     def names(self):
--- a/certbot/certbot/interfaces.py
+++ b/certbot/certbot/interfaces.py
@@ -532,6 +532,41 @@ class IReporter(zope.interface.Interface):
         """Prints messages to the user and clears the message queue."""
 
 
+@six.add_metaclass(abc.ABCMeta)
+class RenewableCert(object):
+    """Interface to a certificate lineage.
+
+    RenewableCert objects have the following attributes:
+
+    .. code-block:: python
+
+        # Path to the certificate file.
+        cert_path: str
+
+        # Path to the private key file.
+        key_path: str
+
+        # Path to the certificate chain file.
+        chain_path: str
+
+        # Path to the full chain file (cert plus chain).
+        fullchain_path: str
+
+        # Name given to the certificate lineage
+        lineagename: str
+
+    """
+
+    @abc.abstractmethod
+    def names(self):
+        """What are the subject names of this certificate?
+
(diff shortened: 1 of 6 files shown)


bmw:
I talked about this with Adrien out of band and did some of my own research and there are some minor benefits with this new approach of using properties:

1. It's more conventional.
2. If you also change the implementation to inherit from the class, Python will error if all properties aren't defined.
3. The PEP 526 style type annotations with mypy seem to (currently) only be used to validate code using the class, not the class implementation itself. You can add a type annotation saying the class needs to have this attribute, never define it, and mypy won't complain.

With this new approach, I had to fix `names` because pylint was complaining that the arguments differed, however, we never used the optional parameter to `names` outside of tests so I just deleted the code altogether.

certbot/tests/crypto_util_test.py
```diff
@@ -181,15 +181,15 @@ def setUp(self):
         super(VerifyCertSetup, self).setUp()
 
         self.renewable_cert = mock.MagicMock()
-        self.renewable_cert.cert = SS_CERT_PATH
-        self.renewable_cert.chain = SS_CERT_PATH
-        self.renewable_cert.privkey = RSA2048_KEY_PATH
-        self.renewable_cert.fullchain = test_util.vector_path('cert_fullchain_2048.pem')
+        self.renewable_cert.cert_path = SS_CERT_PATH
```
adferrand: It is certainly not a big deal, but we are not testing exactly the same logic than before here and in the next modifications for this test module.

Previously we were mocking a `RenewableCert` object instance by assigning string values to internal fields `cert` and so on. Then these internal fields would be accessed directly in the `verify_` methods.

Now we are inserting a mock that slightly diverging from the `RenewableCert` interface instance, since `*_path` are ready-only properties, taking their values from the internal fields. Here the mock contains `*_path` as pure string fields, not read-only properties.

The effect is exactly the same for the test, so it is certainly not worth an additional complexity, but for the record, here a `PropertyMock` (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock) could be used.

certbot/certbot/interfaces.py
```diff
@@ -532,6 +532,62 @@ def print_messages(self):
         """Prints messages to the user and clears the message queue."""
 
 
+@six.add_metaclass(abc.ABCMeta)
+class RenewableCert(object):
+    """Interface to a certificate lineage."""
+
+    @abc.abstractproperty
```
adferrand: Just to be sure.

Before, the result of these properties could be modified by assigning new values to the internal fields that back them. Now, from the interface point of view, there is no constraint nor indication about the capability to modify them (no setter contract for instance).

It is fine for me, since having a mutable or immutable concrete class behind the interface is an implementation detail in my opinion, but do you confirm that is also what you want to express here?

certbot/tests/crypto_util_test.py
bmw: > The effect is exactly the same for the test, so it is certainly not worth an additional complexity, but for the record, here a PropertyMock (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock) could be used.

Cool. I agree a PropertyMock produces objects with more authentic behavior, but for these tests I don't think it matters or that the extra complexity of multiple PropertyMocks is worth it.

certbot/certbot/interfaces.py
bmw: These properties should be treated as read only.

I think this is the normal behavior people expect from a Python class. How often do you create a class and start modifying attributes of that class without being explicitly told you can do so? I'd argue that the absence of being told you can modify these values and there only being read only properties defined loosely but reasonably expresses that you shouldn't modify these values.

I don't think there's really a (sane) way we can enforce this behavior in the concrete class as any subclass can provide their own setter methods overriding any in this class.

The best idea I have if you think this should be more explicitly expressed is to add "read only" to the docstring of each property, but I personally think this is overly verbose and largely just restates what we have in code about there only being read only properties defined.

adferrand: > I think this is the normal behavior people expect from a Python class. How often do you create a class and start modifying attributes of that class without being explicitly told you can do so? I'd argue that the absence of being told you can modify these values and there only being read only properties defined loosely but reasonably expresses that you shouldn't modify these values.

Right for a class, less for an interface, where mutability is not applying. I would never mess up with a class that is designed to be immutable. However here it is an interface, and the meaning is slightly different: the interface states the existence of getters, but does not say anything about the setters.

It would be perfectly valid to write a mutable class that implements the interface. It would be not for the logic here since Certbot is globally relying on the fact that these properties do not change after the object creation.

No, I do not think there is any sane way in any language to force a class to be immutable when it implements some interface.

However for Python specifically, where we are not talking about real interfaces, official documentation is strongly implying that using `@abc.abstractproperty` alone is associated to a read-only property (https://docs.python.org/3/library/abc.html#abc.abstractproperty). So I am fine with it ^^

Issue #7425, opened 2019-10-05
luadns dns validation: Failure if "forward" entry in zone


kaifried:
## My operating system is (include version):
I tested it with fedora 29 and ubuntu 18:04

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
I installed certbot and certbot-dns-luadns via pip.

## I ran this command and it produced this output:
Command:

```bash
certbot certonly  --dns-luadns --dns-luadns-credentials ./lua.ini -d ***.de
```
The output was:

```bash
Saving debug log to /home/kai/le-dns/logdir/letsencrypt.log
Plugins selected: Authenticator dns-luadns, Installer None
Obtaining a new certificate
Performing the following challenges:
dns-01 challenge for ***.de
Cleaning up challenges
Encountered exception during recovery: 
Traceback (most recent call last):
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/certbot/auth_handler.py", line 69, in handle_authorizations
    resps = self.auth.perform(achalls)
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/certbot/plugins/dns_common.py", line 58, in perform
    self._perform(domain, validation_domain_name, validation)
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/certbot_dns_luadns/dns_luadns.py", line 52, in _perform
    self._get_luadns_client().add_txt_record(domain, validation_name, validation)
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/certbot/plugins/dns_common_lexicon.py", line 43, in add_txt_record
    self.provider.create_record(type='TXT', name=record_name, content=record_content)
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/lexicon/providers/base.py", line 80, in create_record
    return self._create_record(rtype, name, content)
  File "/home/*/le-dns/env/lib64/python3.7/site-packages/lexicon/providers/luadns.py", line 45, in _create_record
    existing_records = self._list_records(rtype, name, content)
  File "/home/kai/le-dns/env/lib64/python3.7/site-packages/lexicon/providers/luadns.py", line 71, in _list_records
    'id': record['id']
KeyError: 'id'
```

## dirty fix:
remove all "forward" entries from the dns zone.  
After that, the validation work fine.

bmw:
I'm closing this in favor of https://github.com/certbot/certbot/issues/7631.

[...]

Issue #7411, opened 2019-09-28
dnsmadeeasy: authorizing ~50+ domains causes a fatal rate limiting error


alexzorin:
Originally reported [on the forums](https://community.letsencrypt.org/t/dnsmadeeasy-automatic-cert-generation-returns-400/102984?u=_az).

## My operating system is (include version):
Ubuntu Disco

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
certbot-auto

## I ran this command and it produced this output:

I ran (50 domains):

[...]

Issue #7007, opened 2019-04-30
Upgrade CentOS 6 certbot-auto users to Python 3.5+ from SCL


bmw:
Python 3.4 reached its end of life last month (see https://devguide.python.org/devcycle/#end-of-life-branches). Because of this, we may want to upgrade our CentOS 6 users from Python 3.4 to a newer version of Python from [SCL](https://www.softwarecollections.org/en/).

PR #7630, opened 2019-12-05
Fixes #6972 by updating lexicon to latest version

### LINKED ISSUE #6972: Unable to determine zone identifier
Just wanted to leave a note saying that using the dns-linode plugin, my zone identifier started with a capital letter, which caused the whole thing to choke with the "Unable to determine zone identifier" error message. When I went to linode console and changed the zone name to a lowercase letter, it worked correctly. Capitalizing the commands didn't make a difference. Had to do it from the linode dns manager side.

trinopoty:
This PR updates the lexicon dependency to v3.3.11 which has the fix for #6972 .

### DIFF
--- a/certbot-dns-linode/setup.py
+++ b/certbot-dns-linode/setup.py
@@ -9,7 +9,7 @@ version = '1.1.0.dev0'
 install_requires = [
     'acme>=0.31.0',
     'certbot>=0.39.0',
-    'dns-lexicon>=2.2.3',
+    'dns-lexicon>=3.3.11',
     'mock',
     'setuptools',
     'zope.interface',
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -14,7 +14,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Fixed #6972 by updating lexicon dependency.
 
 More details about these changes can be found on our GitHub repo.
 
--- a/tools/dev_constraints.txt
+++ b/tools/dev_constraints.txt
@@ -19,7 +19,7 @@ configparser==3.7.4
 contextlib2==0.6.0.post1
 coverage==4.5.4
 decorator==4.1.2
-dns-lexicon==3.2.1
+dns-lexicon==3.3.11
 dnspython==1.15.0
 docutils==0.12
 execnet==1.5.0


trinopoty:
@sreyemnayr Would it be possible for you to test out this PR?

alexzorin:
This would additionally resolve:

- #7411, fixed in Lexicon v3.3.4.
- #7425, fixed in Lexicon v3.3.6

trinopoty:
@alexzorin If the dependency in dev_constraints.txt overrides the plugin's dependency after deployment, that would be the case. However, I'm not aware if it does so I can't say.

alexzorin:
Ah, I should have looked at the PR more closely. Sorry :) !

trinopoty:
I can integrate #7411 and #7425 into here or in a different PR if updating the dependency is the only thing that needs to be done.

trinopoty:
`py34` can't find `dns-lexicon==3.3.11`. Any fix for that?

alexzorin:
It seems that [v3.3.5 was the last version to support Python 3.4](https://github.com/AnalogJ/lexicon/commit/d0f6b0b5b3bb292075508f545ff1479d55a487a1#diff-2eeaed663bd0d25b7e608891384b7298R80) and it's restricted in newer releases. cc @adferrand

bmw:
This is both a bit tricky for us to fix and for me to explain why.

I don't really want to update the requirement in `setup.py`. I think the main effect it would have is to completely break the plugin with older versions of Lexicon and it would prevent some Linux distros from being able to package this plugin at all. Certbot packages often see updates in distros where they are unable to update libraries because they may be used by other software. I personally think a Linode plugin that doesn't work with capitalized zone names is better than no Linode plugin at all. We could add a comment in `setup.py` explaining that v3.3.11 is preferred though.

The one place where our users would see an effect by updating the pinning in `dev_constraints.txt` is in our [docker image](https://hub.docker.com/r/certbot/dns-linode). Unfortunately, the way our system of pinnings currently works is the same versions are mostly used everywhere and we can't update the lexicon version until we drop Python 3.4 support. We've been trying to do this for a while and are planning on doing it after our next release. This is blocked by https://github.com/certbot/certbot/issues/7007.

So we could tweak the system of pinnings a bit allowing us to do something like only run with an older version of Lexicon when testing on Python 3.4, but I think I'd prefer waiting until we've dropped Python 3.4 support and updating this pinning globally for our setup.

trinopoty:
@bmw Should I close this PR? I can open another one when Python 3.4 support is dropped with the then latest lexicon release if needed.

bmw:
Either way is fine with me! I created https://github.com/certbot/certbot/issues/7631 and put it in our 1.2.0 milestone to make sure we don't forget about this.

PR #7636, opened 2019-12-09
Issue 6095: Issue with parsing error with a blank ssl.conf for nginx.

mozebdi:
## Pull Request Checklist

- [X] If the change being made is to a [distributed component](https://certbot.eff.org/docs/contributing.html#code-components-and-layout), edit the `master` section of `certbot/CHANGELOG.md` to include a description of the change being made.
- [X] Include your name in `AUTHORS.md` if you like.


mozebdi:
Oops. Code got 10/10 on lint, so I missed the "do not use len()" error. Pushing a fixed version momentarily.

mozebdi:
Unfortunately, ZeroOrMore(contents) results in several pre-existing tests failing.  I can definitely write a test that has a sample .conf file that passes with this change!

ohemorange:
I'm interested in why this causes pre-existing tests to fail -- is this an issue with the testing code, or an accurate reproduction of Certbot's and the internal NGINX parser's behavior? The general goal is to match how NGINX would behave with Certbot's behavior, and have the tests simulate that.

bmw:
Since it seems like a different approach was requested and this PR hasn't seen any updates in almost 4 months, I think we should probably close this. A new PR can always be opened in the future that addresses ohemorange's suggestions.

What do you think @ohemorange?

PR #7638, opened 2019-12-10
[Apache v2] Account for the case where there are multiple ancestors

sydneyli:
In the diamond-include case, where a file is included by multiple other
files in the configuration tree, nodes can have multiple ancestors.

Change interfaces to return a `tuple` of ParserNodes for `ancestors` property and fixes existing interface implementations.


certbot-apache/certbot_apache/apacheparser.py
```diff
@@ -28,7 +28,7 @@ def find_ancestors(self, name):  # pylint: disable=unused-variable
         """Find ancestor BlockNodes with a given name"""
         return [ApacheBlockNode(name=assertions.PASS,
                                 parameters=assertions.PASS,
-                                ancestor=self,
+                                ancestors=self,
```
ohemorange: These `ancestors=self`s should probably be `ancestors=(self)`

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -131,9 +131,10 @@ class ParserNode(object):
 
     ParserNode objects should have the following attributes:
 
-    # Reference to ancestor node, or None if the node is the root node of the
-    # configuration tree.
-    ancestor: Optional[ParserNode]
+    # Reference to ancestor node(s), empty if the node is the root node of the
+    # configuration tree. This node can have multiple ancestors in the case
+    # that it is the root of a file included by multiple files.
```
ohemorange: We're using `root` twice here, and as I understand it, it doesn't have quite the same meaning both times. Could we find a different word for the second usage?

certbot-apache/certbot_apache/tests/parsernode_util_test.py
```diff
@@ -48,8 +48,8 @@ def test_parsernode(self):
         params = self._setup_parsernode()
         ctrl = self._setup_parsernode()
 
-        ancestor, dirty, filepath, metadata = util.parsernode_kwargs(params)
-        self.assertEqual(ancestor, ctrl["ancestor"])
+        ancestors, dirty, filepath, metadata = util.parsernode_kwargs(params)
+        self.assertEqual(ancestors, ctrl["ancestors"])
```
ohemorange: This change makes me think that a) maybe we should use a set here instead of a tuple, and b) we might need to fix the equality checks in `assertions`

sydneyli: I think that's right-- we might want hashcodes for these objects anyhow since we essentially construct one in the `__eq__` functions.

certbot-apache/certbot_apache/interfaces.py
```diff
@@ -198,6 +199,11 @@ def find_ancestors(self, name):
         Traverses the ancestor tree up, searching for BlockNodes with a specific
         name.
 
+        .. note:: To maintain consistency with the original parser, this search
+                  function currently only traverses nodes in the same file. Due
+                  to this, we also don't run into the multiple-includes
+                  ambiguity that the `ancestors` property has, for the time being.
```
joohoi: Just a note: This is going to change when #7642 (and a subsequent PR leveraging it for ParserNode interfaces) lands.

certbot-apache/certbot_apache/tests/parsernode_util_test.py
m0namon: As we talked about, this is probably overly complex. We don't use `__eq__` outside of testing, so I changed the tests to use `assertions` instead and removed `__eq__` altogether. That way we don't have to check ancestor parity, as the `assertions` library already doesn't. If it turns out we need to, we can implement that in the future.

certbot-apache/certbot_apache/_internal/interfaces.py
```diff
@@ -131,9 +131,10 @@ class ParserNode(object):
 
     ParserNode objects should have the following attributes:
 
-    # Reference to ancestor node, or None if the node is the root node of the
-    # configuration tree.
-    ancestor: Optional[ParserNode]
+    # Reference to ancestor node(s), empty if the node is the root node of the
+    # configuration tree. This node can have multiple ancestors if it is in
+    # a file included by multiple files.
```
ohemorange: Let's add a note here about how callers can't have any expectation around ordering of the ancestors within the tuple.

certbot-apache/certbot_apache/_internal/assertions.py
```diff
@@ -123,8 +123,7 @@ def isEqualVirtualHost(first, second):
         first.addrs == second.addrs and
         first.ssl == second.ssl and
         first.enabled == second.enabled and
-        first.modmacro == second.modmacro and
-        first.ancestor == second.ancestor
```
ohemorange: Actually, thinking about this more... the most important check that calls this is here: https://github.com/certbot/certbot/blob/apache-parser-v2/certbot-apache/certbot_apache/_internal/configurator.py#L916

Without checking `ancestors`, we could get false positives for identical VirtualHosts that are located in different spots in the tree. That seems bad. If we're comparing across implementations and therefore can't look at `path` or `node`, `ancestors` seems like the only way to check for sure that they're actually the same. Although, as I type this, I realize now that that still wouldn't differentiate between sibling VirtualHosts with the same ancestors. So we may have a larger problem here.

ohemorange: My hope is that someone will jump in and say, "ah, but you forgot about x, therefore this isn't actually a problem."

m0namon: That sounds right, but I'm not sure we necessarily have to use ancestors to disambiguate.What's the minimum set of information we need to disambiguate two otherwise identical VirtualHost blocks? In Augeas, the augeas path seems like it'll do that. In apacheconfig land, I think we might be able to just compare memory addresses, since the tree is generated once at the beginning. Tagging @joohoi here too.

I will also think about this a bit more.

ohemorange: Yes, I agree that full path in augeas and testing if they're the same actual object in apacheconfig should do it. That does make checking across implementation impossible though.

bmw:
Is this also something we should land as part of wrapping up Apache parser work for now or should this be done later?

ohemorange:
I am closing this PR from 2019.

PR #7643, opened 2019-12-13
[Apache v2] Add apacheconfig as a dependency

sydneyli:
https://pypi.org/project/apacheconfig/0.3.0/ release has the things we need. Let's add it as a dependency.

apacheconfig's largest dependency is `ply` and will work with any version above 3.0. The oldest packaged version is 3.4 packaged in CentOS 6. It's also already pinned in `oldest-constraints.txt`.

We chatted about adding the dependency in dev extras while the apache parser undergoes development. Once we think upstream (and the apache parser) has stabilized, we can move the dependency into the regular requirements and remove the `apache-v2` test env.


sydneyli:
Yep! In this PR I've added an extra tox env for running unit-tests with `apacheconfig`, and I think we can add more as we go along.

I'm thinking the bugfix will be landed upstream fairly soon, so it's not very long to wait. If it doesn't land by tomorrow or the day after, I'm fine landing this PR first.

PR #7648, opened 2019-12-18
Small how to uninstall certbot-auto

BarbzYHOOL:
Also added missing sudo

feel free to edit it or discard it

### DIFF
--- a/certbot/docs/install.rst
+++ b/certbot/docs/install.rst
@@ -70,6 +70,7 @@ The ``certbot-auto`` wrapper script installs Certbot, obtaining some dependencie
 from your web server OS and putting others in a python virtual environment. You can
 download and run it as follows::
 
+  cd /tmp
   wget https://dl.eff.org/certbot-auto
   sudo mv certbot-auto /usr/local/bin/certbot-auto
   sudo chown root /usr/local/bin/certbot-auto
--- a/certbot/docs/install.rst
+++ b/certbot/docs/install.rst
@@ -70,7 +70,6 @@ The ``certbot-auto`` wrapper script installs Certbot, obtaining some dependencie
 from your web server OS and putting others in a python virtual environment. You can
 download and run it as follows::
 
-  cd /tmp
   wget https://dl.eff.org/certbot-auto
   sudo mv certbot-auto /usr/local/bin/certbot-auto
   sudo chown root /usr/local/bin/certbot-auto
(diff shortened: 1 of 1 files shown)


certbot/docs/install.rst
```diff
@@ -70,11 +70,14 @@ The ``certbot-auto`` wrapper script installs Certbot, obtaining some dependencie
...
-  user@webserver:~$ chmod 0755 /usr/local/bin/certbot-auto
-  user@webserver:~$ /usr/local/bin/certbot-auto --help
+  cd /tmp
+  wget https://dl.eff.org/certbot-auto
+  sudo mv certbot-auto /usr/local/bin/certbot-auto
+  sudo chown root /usr/local/bin/certbot-auto
+  sudo chmod 0755 /usr/local/bin/certbot-auto
+  /usr/local/bin/certbot-auto --help
+
+To remove certbot-auto, just delete it and the files it places in /opt (aka /opt/eff.org), along with any cronjob you may have created.
```
bmw:
```suggestion
To remove certbot-auto, just delete it and the files it places under /opt/eff.org, along with any cronjob or systemd timer you may have created.
```

certbot/docs/install.rst
```diff
@@ -70,11 +70,14 @@ The ``certbot-auto`` wrapper script installs Certbot, obtaining some dependencie
 from your web server OS and putting others in a python virtual environment. You can
 download and run it as follows::
 
-  user@webserver:~$ wget https://dl.eff.org/certbot-auto
-  user@webserver:~$ sudo mv certbot-auto /usr/local/bin/certbot-auto
-  user@webserver:~$ sudo chown root /usr/local/bin/certbot-auto
-  user@webserver:~$ chmod 0755 /usr/local/bin/certbot-auto
-  user@webserver:~$ /usr/local/bin/certbot-auto --help
+  cd /tmp
```
bmw: I think this can probably be omitted. If we're worried about security, we could do something like `cd $(mktemp -d)`, but I think it's overkill and distracts people from the important parts of the instructions.

BarbzYHOOL: it's just that you don't want to download the file in a random dir but then i realized that the file is moved anyway so yes it is quite useless

BarbzYHOOL:
ok improved PR, squash the commits though :D

PR #7650, opened 2019-12-19
Include header files for compilation

bmw:
https://github.com/certbot/certbot/pull/7120 broke our Windows installer tests (see https://dev.azure.com/certbot/certbot/_build/results?buildId=764&view=results) because `pip` tries to install and compile `bcrypt` when installing `certbot-ci` which failed because we were installing Python without development headers. This PR adds development headers to the installation fixing the problem.

cc @adferrand


bmw:
You can see tests passing with this change at https://dev.azure.com/certbot/certbot/_build/results?buildId=768&view=results.

PR #7682, opened 2020-01-14
Update files from 1.1.0 release

bmw:


certbot-dns-cloudxns/setup.py
```diff
@@ -4,13 +4,13 @@
...
 from setuptools.command.test import test as TestCommand
 
-version = '1.1.0.dev0'
+version = '1.2.0.dev0'
 
 # Remember to update local-oldest-requirements.txt when changing the minimum
 # acme/certbot version.
 install_requires = [
     'acme>=0.31.0',
-    'certbot>=1.0.0.dev0',
```
ohemorange: Weird that the previous release didn't catch this

bmw: It looks like the problem is it landed after the release.

https://github.com/certbot/certbot/pull/7551 was created before the last release, but landed after it and the version numbers weren't fixed. The only reason they were increased was for the "oldest" tests where we wanted the changes to the dependencies in `certbot[dev]`.

It probably doesn't matter, but I'd like to land this PR and then fix it in a separate commit so we can cherry pick it to the point release branch to ensure things are properly tested there.

What do you think?

ohemorange: works for me

PR #7685, opened 2020-01-16
Don't run some tests multiple times

bmw:
This PR stops us from running some unit tests multiple times when running the extended test suite.

I removed Python 3.7 from the tests we run on PRs because we're also running Python 3.8 tests and I personally think running our tests on the oldest and newest version of Python 3 we support is good enough outside of our nightly tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,9 +62,6 @@ matrix:
     - python: "3.4"
       env: TOXENV=py34
       <<: *not-on-master
-    - python: "3.7"
-      env: TOXENV=py37
-      <<: *not-on-master
     - python: "3.8"
       env: TOXENV=py38
       <<: *not-on-master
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -175,9 +169,6 @@ matrix:
     - python: "3.7"
       env: TOXENV=py37
       <<: *extended-test-suite
-    - python: "3.8"
```
adferrand: Isn't it 3.7 that you want to remove?

bmw: No.

I should have probably made it clearer in my comment, but instead I left the Python 3.8 tests in the `not-on-master` tests at https://github.com/certbot/certbot/pull/7685/files#diff-354f30a63fb0907d4ad57269548329e3R65.

adferrand: OK, got it!

PR #7690, opened 2020-01-16
Cleanup disabled warnings list in pytest.ini

bmw:
The comment about TLS-SNI in `pytest.ini` refers to a disable that was deleted in another PR.

The problem with `ipdb` has been fixed in newer versions which you can verify by following the instructions in https://github.com/certbot/certbot/pull/6714.

I forgot to remove the disable about `get_systemd_os_info` in #7526.


bmw:
Once this lands, we should make sure this change is clearly flagged for everyone because people will need to recreate their venv to avoid hitting this error when using pytest/ipdb in this way.

PR #7693, opened 2020-01-16
Fix minimum certbot version in plugins in 1.1.x

bmw:
This is probably overkill, but this cherry picks #7684 to the `1.1.x` branch to properly specify the minimum Certbot version in case of a point release.


bmw:
We're planning on releasing 1.2 on Tuesday so I don't think we need this.

PR #7702, opened 2020-01-21
Downgrade NSIS and upgrade Python

bmw:
Windows installer tests started failing last night because Chocolatey was refusing to downgrade NSIS. See https://dev.azure.com/certbot/certbot/_build/results?buildId=951.

I fixed that problem by passing `--allow-downgrade` to chocolatey, but the Python installation for `certbot-ci` started failing. This happened even if we do not run the Certbot installer until after Python is installed which should mean that these failures were unrelated. See https://dev.azure.com/certbot/certbot/_build/results?buildId=957&view=results. I fixed this by upgrading Python used for testing to 3.8.1.

You can see the tests for the combination of these changes passing at https://dev.azure.com/certbot/certbot/_build/results?buildId=956&view=results.


bmw:
@adferrand, are you free and interested to review this?

PR #7707, opened 2020-01-22
Fix collections.abc imports for Python 3.9

### LINKED ISSUE #7692: Properly use collections.abc
Python 3.9 is planning on enforcing that abstract base classes in the collections module are imported from `collections.abc`. See https://docs.python.org/3/library/collections.html#module-collections.

We currently are not always doing this in Certbot and are ignoring the deprecation warning about it at https://github.com/certbot/certbot/blob/6e07e8b5c0f3c6da2ae03a2a63adc7ce8e15a202/pytest.ini#L16

Properly testing and finding all occurrences of this problem is a little tricky (see https://github.com/certbot/certbot/issues/7691), but we should fix this and remove the `pytest.ini` entry well before Python 3.9 is released.

hugovk:
## Pull Request Checklist

Fixes #7692.

What sort of `certbot/CHANGELOG.md`

### DIFF
--- a/certbot/certbot/_internal/plugins/disco.py
+++ b/certbot/certbot/_internal/plugins/disco.py
@@ -13,6 +13,12 @@ from certbot import errors
 from certbot import interfaces
 from certbot._internal import constants
 
+try:
+    # Python 3.3+
+    from collections.abc import Mapping  # pylint: disable=no-name-in-module
+except ImportError:  # pragma: no cover
+    from collections import Mapping
+
 logger = logging.getLogger(__name__)
 
 
@@ -178,7 +184,7 @@ class PluginEntryPoint(object):
         return "\n".join(lines)
 
 
-class PluginsRegistry(collections.Mapping):
+class PluginsRegistry(Mapping):
     """Plugins registry."""
 
     def __init__(self, plugins):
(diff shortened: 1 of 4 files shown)


pytest.ini
```diff
@@ -10,4 +10,3 @@
 filterwarnings =
     error
     ignore:decodestring:DeprecationWarning
-    ignore:.*collections\.abc:DeprecationWarning
```
tirkarthi: Nit : The text at https://github.com/certbot/certbot/pull/7707/files#diff-513c9b6861fc981fb2c708db477f542cR7 could also be removed.

hugovk: Thanks, removed!

certbot/certbot/_internal/plugins/disco.py
```diff
@@ -13,6 +13,12 @@
 from certbot import interfaces
 from certbot._internal import constants
 
+try:
+    # Python 3.3+
+    from collections.abc import Mapping  # pylint: disable=no-name-in-module
```
adferrand: Disabling `no-name-in-module` is not needed here, since we are running the lint checks on Python 3.

hugovk: Thanks, updated!

Issue #7204, opened 2019-06-28
Update the Apache plugin's TLS config to follow Mozilla's new recommendations


bmw:
This can be found at https://ssl-config.mozilla.org. In general, doing this is a little tricky because we don't want to break people with old versions of Apache/OpenSSL. Looking at the code that is used to generate these config snippets may be useful for this which is found at https://github.com/mozilla/ssl-config-generator.

For now, I don't think we should be enabling HTTP2, HSTS, or OCSP stapling as part of this. We may want to reconsider doing this in the future, especially for HTTP2, but so far we've thought about these things as an optional security enhancement for Certbot and I think deciding to enable these features on all configs should be a separate issue.

bmw:
@adferrand, this is tentatively assigned to you for our next release, but we'll chat more during our 1-1.

ohemorange:
I think this should block on un-reverting the session tickets machinery and implementing #7350.

bmw:
When doing this we should continue allowing `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA` like we did in the nginx plugin in https://github.com/certbot/certbot/pull/7414.

fleish:
Hi @bmw, is there a reason this keeps getting pushed out? If I'm reading correctly it's now slated for 1.2.0 in February 2020 which means current installations of certbot+nginx are still supporting TLSv1.1 & even worse, TLSv1. I understand from previous issues the certbot team is taking direction from Mozilla, but they are no longer recommending this (for many months now) and both of these protocols have been widely considered deprecated since June 30, 2018.

bmw:
Thanks for the nudge on this @fleish.

To be clear, this issue is about Apache, not nginx. We've been following Mozilla's updated TLS recommendations for nginx since August. See https://github.com/certbot/certbot/blob/master/certbot/CHANGELOG.md#0370---2019-08-07.

The reason this issue hasn't been resolved though is due to the combination of this issue being pretty difficult to fully implement (see https://github.com/certbot/certbot/issues/7350) and us having commitments to work on other aspects of Certbot. We know it's important and I honestly think we'll get it to it soon, but the small team of people maintaining Certbot just haven't been able to tackle this issue yet.

[...]

PR #7712, opened 2020-01-22
Disable old SSL versions and ciphersuites to follow Mozilla recommendations in Apache

ohemorange:
Part of #7204.

Makes the smaller changes described at https://github.com/certbot/certbot/issues/7204#issuecomment-571838185 to disable many old ciphersuites and TLS versions < 1.2. Does not add checks for OpenSSL version or modify session tickets.

Since Apache uses TLS protocol blacklisting instead of whitelisting (as in NGINX), we additionally may not need to determine if the server supports TLS1.3 and turn it on or off based on Apache version.


certbot-apache/certbot_apache/_internal/centos-options-ssl-apache.conf
```diff
@@ -7,8 +7,8 @@
 SSLEngine on
 
 # Intermediate configuration, tweak to your needs
-SSLProtocol             all -SSLv2 -SSLv3
-SSLCipherSuite          ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS
+SSLProtocol             all -SSLv2 -SSLv3 -TLSv1 -TLSv1.1
+SSLCipherSuite          ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-SHA
 SSLHonorCipherOrder     on
```
bmw: I wasn't clear about this at https://github.com/certbot/certbot/issues/7204#issuecomment-571838185, but I was imagining that in this PR we could update our Apache TLS configs to use the worst case recommendation given by Mozilla for any Apache/OpenSSL version. This would include things like setting `SSLHonorCipherOrder off` and maybe deleting some of our other configuration options in these files. For the latter, I think we should probably double check that we agree with Mozilla about removing them.

What do you think about doing that in this PR as well?

ohemorange: What does "worst case" mean here? My first guess was "for old versions of Apache" but I don't think that interpretation quite fits.

bmw: The resulting config will probably look like the recommended settings for old versions of Apache and OpenSSL.

Another way to look at it would be to make the TLS config changes we want to make unconditionally regardless of the user's software versions. We can make the config changes which we only want to do conditionally once we have a good way to figure out the OpenSSL version.

ohemorange: Ah, you mean the lowest common denominator of config changes.

bmw: Yeah that's a good way to put it.

ohemorange: Additional considerations/changes that I'm putting here because of Github's UX:
- Removing commented-out lines. Not sure why we had those in the first place.
- According to https://github.com/mozilla/server-side-tls/issues/263, the relevant deprecation date is January 2020. So keeping `ECDHE-RSA-AES128-SHA` around shouldn't be relevant for the next release. My preference is to remove it, and update `SSLHonorCipherOrder` to `off`. If we do include `ECDHE-RSA-AES128-SHA`, `SSLHonorCipherOrder` should actually be `on`, as we'd prefer that clients don't use `ECDHE-RSA-AES128-SHA`, and therefore should honor server ciphersuite order. Whichever way we go, we should also fix NGINX's configuration, because we shouldn't both include `ECDHE-RSA-AES128-SHA` in the ciphersuites list and also let clients choose the order.
- Looks like `SSLOptions +StrictRequire` is still relevant for certbot reasons
- The mozilla recs don't mention `SSLCompression off` (which we have for non-centos systems) at all. The default is now `off` but it was `on` [before 2.4.3](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcompression). We may want to suggest adding the `SSLCompression off` directive to the mozilla recs.

ohemorange: (I wrote Windows 2008 everywhere, but Windows 7 is also EOL now.)

bmw: > Looks like `SSLOptions +StrictRequire` is still relevant for certbot reasons

I'd bet money that you're right, but do you have a link/explanation for our reason for keeping it?

> We may want to suggest adding the SSLCompression off directive to the mozilla recs.

I think they used to have it before the major redesign a few months ago. They may have just left it out for simplicity because it only affects absolutely ancient versions of Apache, but I couldn't quickly find an explanation for its removal and making sure this was done on purpose is probably a good idea.

Do you want to check with them or should I?

Otherwise, this all sounds right to me! Thanks for looking into this and writing up what you found.

ohemorange: > I'd bet money that you're right, but do you have a link/explanation for our reason for keeping it?

Haha, good one! This line has been in Certbot for 8 years, there's no link. But here's [what it does](https://httpd.apache.org/docs/current/mod/mod_ssl.html#ssloptions):
> This forces forbidden access when SSLRequireSSL or SSLRequire successfully decided that access should be forbidden. Usually the default is that in the case where a ``Satisfy any`` directive is used, and other access restrictions are passed, denial of access due to SSLRequireSSL or SSLRequire is overridden (because that's how the Apache Satisfy mechanism should work.) But for strict access restriction you can use SSLRequireSSL and/or SSLRequire in combination with an ``SSLOptions +StrictRequire``. Then an additional ``Satisfy Any'' has no chance once mod_ssl has decided to deny access.

Basically, by having this line, we override users' insecure decisions that they may have copied and pasted from insecure configs. If someone really knows what they're doing and disagrees with what's clearly the correct security decision they can always remove it. That being said, it's likely a rare case and we could take it out for cleanliness, but it seems probably beneficial to keep it around.

> Do you want to check with them or should I?

Here's the [commit that removed it](https://github.com/mozilla/ssl-config-generator/commit/e4fe21c16ad2d0e79776f44614cfe05f738f2b50#diff-573e9e12ace6899c2ccc32e3cc9f2a3c). I'll ask.

bmw: That makes sense about `SSLOptions +StrictRequire`. I agree keeping it probably makes more sense.

> Here's the [commit that removed it](https://github.com/mozilla/ssl-config-generator/commit/e4fe21c16ad2d0e79776f44614cfe05f738f2b50#diff-573e9e12ace6899c2ccc32e3cc9f2a3c). I'll ask.

Thanks!

ohemorange: > Available in httpd 2.4.3 and later, if using OpenSSL 0.9.8 or later; virtual host scope available if using OpenSSL 1.0.0 or later. The default used to be `on` in version 2.4.3.

Based on both this comment and the [apache changelogs](https://archive.apache.org/dist/httpd/), the `SSLCompression` option was first added in 2.4.3, but then backported to 2.2.24. 2.2.26 mentions changing the default to `off`, so only in 2.2.24 and 2.2.25 of that series would it be helpful to bother turning it off. 2.4.4 mentions changing the default to `off`, but the 2.4.3x series (all released after 2.4.4) just don't mention changing the default either way, so it's hard to say if it's default on or off in that situation.

For perspective, RHEL 6 which EOLs in November shipped with 2.2.15, so someone might have theoretically upgraded to some bad version. Ubuntu 14.04 shipped with 2.4.7, which is safe in this respect. Honestly it's probably safest and the least amount of headache to keep this line around until November 2020, especially since we haven't previously run into issues with having it around.

april: I can't remember if it was intentional or not when I did the redesign (as it was a good while ago), but in general the current SSL Configuration Generator tries to leave off any settings where the default is decent. This is for two reasons:

a) it is less stuff to maintain, and
b) even listing `SSLCompression off` might tempt people to change it

I talked to ebp and I think I will continue to leave it out of there. I personally can't seem to find any supported distributions that use 2.4.3, and it seems unlikely that someone is using the configuration generator to generate a new configuration for a 7.5(!) year old versions of Apache.

Either way (defaults or purposefully setting to `off`) is probably fine though.  :)

bmw: Thanks to both of you for the info.

As for next steps for Certbot, I looked into this a bit and we don't set `SSLCompression off` on RHEL 6 based systems because the version of Apache packaged in RHEL 6 doesn't recognize the value (and we make the decision on what file to used based on the distro rather than the Apache version).

Because of this, the fact Apache 2.2.x is EOL'd, and us not being aware of any 2.4.3 packages, I think we should just remove `SSLCompression off` ourselves as well and consolidate our two files. What do you think Erica?

ohemorange: Ah, I missed that we already don't set this on RHEL 6! Especially since this allows us to consolidate the files, I love that. But I think if we're going to remove it we really do need to check distros more extensively.

For example on Debian: jessie has 2.4.10, stretch has 2.4.25, buster has 2.4.38. Which immediately begs the question: if this feature was added in 2.4.3, how is this not erroring in 2.4.10 and 2.4.25? Is apache just bad about updating changelogs? Are we currently using the file labelled "centos" for debian? Questions abound.

ohemorange: Gentoo's good.

Fedora 26 updates has 2.4.33. Fedora 27 updates has 2.4.34. 28 has 2.4.39, 29+ is good.

FreeBSD is good. OpenBSD is good.

Devuan ASCII 2.0 has 2.4.25, Jessie 1.0 has 2.4.10, Beowulf has 2.4.38, after is good.

openSUSE tumbleweed is good, Leap 15.0 and .1 have 2.4.33, but 15.2 is good.

Arch Linux and MacOS aren't on repology.

Summary: there are still many subversions of 2.4.3x out there (which might have the wrong setting), and even earlier versions (which presumably don't recognize the option but maybe just aren't failing, or we're actually using the other version, or everyone on those already ran into the issue and updated, or apache isn't good about updating its changelogs).

ohemorange: Next step would be to manually check what's going on to answer these questions, including testing on the various < 2.4.4 versions.

bmw: I think the last digit of Apache's version number in `2.4.x` is just an incrementing counter. For instance, 2.4.10 comes 7 releases after 2.4.3.

With this in mind, I think the only systems with Apache 2.4 that may be negatively affected by us removing this line are those with the exact version number 2.4.3. I'm not aware of any systems packaged with this [7 year old](https://github.com/apache/httpd/releases/tag/2.4.3) version of Apache.

I found Arch Linux listed at https://repology.org/project/apache/versions under "Arch". For macOS, I found entries for Homebrew and MacPorts. All of them have the latest version of 2.4.41:

* https://formulae.brew.sh/formula/httpd
* https://www.archlinux.org/packages/extra/x86_64/apache/
* https://ports.macports.org/port/apache2/summary

ohemorange: OH that makes so much more sense than what I was imagining (entirely based on the lexicographic ordering on their downloads page), which is another level of major and minor versions. That makes it so much easier and answers every question!

ohemorange: In that case yeah, I'm absolutely totally good with removing the line.

PR #7721, opened 2020-01-24
Drop Python 3.4 support

bmw:
Fixes https://github.com/certbot/certbot/issues/7393.

### DIFF
--- a/certbot/setup.py
+++ b/certbot/setup.py
@@ -136,7 +136,6 @@ setup(
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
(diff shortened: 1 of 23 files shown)


certbot/setup.py
```diff
@@ -88,7 +88,6 @@ def read_file(filename, encoding='utf8'):
     'astroid',
     'mypy',
     'pylint',
-    'typing',  # for python3.4
```
adferrand: I suppose `tools/dev_constaints.txt` can be recalculated since `typing` does not need to be pinned anymore.

bmw: Nice catch, but it looks like `typing` is still installed because our pinned version of [Sphinx requires it for Python 2](https://github.com/sphinx-doc/sphinx/blob/v1.7.5/setup.py#L38) so I think we should keep the pinning for now.

PR #7722, opened 2020-01-24
Add --dns-server option in run_acme_server

### LINKED ISSUE #7717: run_acme_server and DNS plugins
As reported at https://github.com/certbot/certbot/pull/7583#issuecomment-577433664, if you're trying to use `run_acme_server` while working on a DNS plugin, you're going to have some issues.

What can we do to avoid people hitting this problem? My first thought would be to update our docs to say that if people are using a real domain for testing, they should just use Certbot with `--staging` rather than running an ACME server locally.

adferrand:
Fixes #7717

This PR adds a `--dns-server` option to the `run_acme_server` test tool, in order to provide an arbitrary DNS server to Pebble or Boulder for the integration tests.

I also take this occasion to make `run_acme_server` a real CLI tool using argparse, and set the `--server-type` (default `pebble`) option as well.


Tugzrida:
Perhaps add something to the manual integration tests section in certbot/docs/contributing.rst too? That's where I found out about the test server.

bmw:
Closing and reopening this to test something with the new Azure setup.

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -208,14 +226,19 @@ def _launch_process(self, command, cwd=os.getcwd(), env=None, force_stderr=False
...
-
-    acme_server = ACMEServer(server_type, [], http_proxy=False, stdout=True)
+    parser = argparse.ArgumentParser(
+        description='CLI tool to start a local instance of Pebble or Boulder CA server.')
+    parser.add_argument('--server-type', '-s',
+                        choices=['pebble', 'boulder-v1', 'boulder-v2'], default='pebble',
+                        help='type of CA server to start: can be Pebble or Boulder '
+                             '(in ACMEv1 or ACMEv2 mode), Pebble is used if not set.')
+    parser.add_argument('--dns-server', '-d',
+                        help='(Pebble specific) specify the DNS server as `IP:PORT` to use by '
```
bmw, line 236: This isn't Pebble specific anymore is it?
```suggestion
                        help='specify the DNS server as `IP:PORT` to use by '
```

certbot/docs/contributing.rst
```diff
@@ -180,6 +180,9 @@ that can be used once the virtual environment is activated:
 - Press CTRL+C to stop this instance.
 - This instance is configured to validate challenges against certbot executed locally.
 
+.. note:: Some options are available to tweak the local acme server. You can execute
+    ``run_acme_server --help` to see the inline help of ``run_acme_server`` tool.
```
bmw, line 184: The syntax here doesn't seem quite right. Rendered docs looks like:
Screen Shot 2020-11-12 at 12 22 54 PM

certbot/docs/contributing.rst
```diff
@@ -180,6 +180,9 @@ that can be used once the virtual environment is activated:
 - Press CTRL+C to stop this instance.
 - This instance is configured to validate challenges against certbot executed locally.
 
+.. note:: Some options are available to tweak the local acme server. You can execute
```
bmw, line 183: nit:
```suggestion
.. note:: Some options are available to tweak the local ACME server. You can execute
```

bmw (changes requested):
I had a few documentation nitpicks, but this otherwise LGTM!

certbot/docs/contributing.rst
adferrand: Fixed.

bmw (approved):

certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -167,6 +176,14 @@ def _prepare_boulder_server(self):
         os.rename(join(instance_path, 'test/rate-limit-policies-b.yml'),
                   join(instance_path, 'test/rate-limit-policies.yml'))
 
+        if self._dns_server:
+            # Change Boulder config to use the provided DNS server
+            with open(join(instance_path, 'test/config/va.json'), 'r') as file_h:
```
alexzorin, line 181: Oops, I should have piped up yesterday. Wasn't aware this was on the radar for merging :scream:.

At some point after this code was written, Boulder's `test/config` environment was updated to [make use of remote VAs](https://github.com/letsencrypt/boulder/blob/bcca7c2d05e2ced89f726adccd37ab388b6c9c9d/test/config/va.json#L32-L42). They are configured separately in `va-remote-{a,b}.json`.

As written, I think this will fail because both remote VAs will fail to query the right DNS server and the `EnforceMultiVA` feature flag is on as well.

I do not think it is urgent, as I am not planning to integrate Boulder with this flag yet anyway.

bmw: Thanks for letting us know!

I agree it's not urgent, but think we should fix it or rip the boulder code out relatively soon. I think it's easy to do and I think it's better to fix the problem now rather than letting people stumble into it and rediscover it in the future. I created https://github.com/certbot/certbot/issues/8453 to track that.

PR #7724, opened 2020-01-25
Windows installer integration tests

adferrand:
As discussed in #7539, we need proper tests of the Windows installer itself in order to variety that all the logic contained in a production-grade runtime of Certbot on Windows is correctly setup by each version of the installer, and so for a variety of Windows OSes. 

This PR handles this requirement. The new `windows_installer_integration_tests` module in `certbot-ci` will:
* run the given Windows installer
* check that Certbot is properly installed and working
* check that the scheduled renew task is set up
* check that the scheduled task actually launch the Certbot renew logic

The Windows nightly tests are updated accordingly, in order to have the tests run on Windows Server 2012R2, 2016 and 2019.

These tests will evolve as we add more logic on the installer.


certbot-ci/windows_installer_integration_tests/test_main.py
```diff
@@ -0,0 +1,60 @@
...
+    else:
+        raise AssertionError('Expect certbot to not be available in the PATH.')
+
+    try:
+        # Install certbot
+        subprocess.check_call([request.config.option.installer_path, '/S'])
+
+        # Assert certbot is installed and runnable
+        output = subprocess.check_output(['certbot', '--version'], universal_newlines=True)
+        assert 'certbot 1.' in output, 'Flag --version does not output a version.'
```
bmw: nit: Can we use something like a regex to prevent this test from breaking when we release Certbot 2.0?

certbot-ci/windows_installer_integration_tests/test_main.py
```diff
@@ -0,0 +1,60 @@
...
+            time.sleep(1)
+
+        log_path = os.path.join('C:\\', 'Certbot', 'log', 'letsencrypt.log')
+
+        modification_time = os.path.getmtime(log_path)
+        assert now < modification_time, 'Certbot log file has not been modified by the renew task.'
+
+        with open(log_path) as file_h:
+            data = file_h.read()
+        assert 'DEBUG:certbot._internal.renewal:no renewal failures' in data, 'Renew task did not execute properly.'
```
bmw: nit: While this log line will probably never change, I don't love the possibility of nightly tests failing way over here in the installer tests if it does. Are we able to do something like check the exit status of the scheduled task?

certbot-ci/windows_installer_integration_tests/test_main.py
```diff
@@ -0,0 +1,60 @@
...
+
+    finally:
+        # Sadly this command cannot work in non interactive mode: uninstaller will ask explicitly permission in an UAC prompt
```
bmw: That's unfortunate. I assume you didn't find any (sane) way to uninstall the program non-interactively?

If not, maybe we could either allow this code to run interactively or make it harder for a dev to run these tests and unknowingly make lasting changes to their machine? This could be done in future work if you prefer, but I think we should at least make an issue about it.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: We can, but this is not what I want to check here. I want to verify that whatever happens in the scheduled task, and without any assumption of its internal logic, the certbot renew process is effectively called when the scheduled task is triggered.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: This test is mainly designed to be run non interactively on Azure Pipelines. I propose we check for specific env variables when run on Azure, and warn at the initialization of pytest we explicit approval before running the test.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: Done.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: Done, this test now asks for user confirmation. `--confirm` flag is added to execute the test non interactively with an implicit confirmation, and is used in the Azure pipeline.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: As far as I know, checking certbot logs is the only way for all situations to see that certbot really tried to renew the certificates.

bmw: Ah I see. We could reach into the place where Certbot is installed on disk and temporarily move the script and replace it with our own executable, but that's probably more trouble than it's worth.

Instead, how about we change the assertion here to only look for "no renewal failures" and we add a comment in `renewal.py` where that line is logged saying that if this text is changed, these tests need to be updated?

certbot-ci/windows_installer_integration_tests/conftest.py
```diff
@@ -22,3 +24,26 @@ def pytest_addoption(parser):
...
+
+def pytest_configure(config):
+    """
+    Standard pytest hook used to add a configuration logic for each node of a pytest run.
+    :param config: the current pytest configuration
+    """
+    if not config.option.confirm:
+        capture_manager = config.pluginmanager.getplugin('capturemanager')
+        try:
+            capture_manager.suspendcapture(in_=True)
```
bmw: It looks like these [functions have been renamed in newer versions of pytest](https://stackoverflow.com/questions/50638571/in-pytest-how-to-config-capturemanager-plugin-error-capturemanager-object-has) and I'm not sure pytest considers them part of their stable API as they made the change after pytest 3.0 but before pytest 4.0.

Unless there's a more stable way to do this, how about we just raise an assertion saying that the flag must be set to run these tests?

certbot-ci/windows_installer_integration_tests/conftest.py
```diff
@@ -22,3 +24,26 @@ def pytest_addoption(parser):
                                           'nsis', 'certbot-beta-installer-win32.exe'),
                      help='set the path of the windows installer to use, default to '
                           'CERTBOT_ROOT_PATH\\windows-installer\\build\\nsis\\certbot-beta-installer-win32.exe')
+    parser.addoption('--confirm', action='store_true',
```
bmw: nit: Especially if we change the behavior of the flag, what do you think about a name like `--allow-persistent-changes`?

adferrand: I was thinking of a `--confirm` flag as "say yes to whatever interactive question you would encounter", quite similarly to `-y` with apt.

But if you prefer a flag for each interactive behavior, it is your call. Still prefer that ?

certbot-ci/windows_installer_integration_tests/conftest.py
adferrand: Ok. Implemented.

certbot-ci/windows_installer_integration_tests/test_main.py
adferrand: Done.

certbot-ci/windows_installer_integration_tests/conftest.py
bmw: I don't care that much, but while our only use of this flag is with the behavior here, I personally prefer a flag that tries to clearly explain its use. If we get multiple of these flags in the future and people are wanting to run these tests outside of CI, I would maybe prefer a flag like `--force-yes` instead of or in addition to flags specifically for each warning.

adferrand: Ok, done.

PR #7725, opened 2020-01-27
Allow setting DNS TTL

### LINKED ISSUE #7969: Allow setting DNS TTL  in DigitalOcean
This is more of a conceptual bug, but here goes
Running certbot with --test-cert first, getting a test certificate issued, then trying the same command without --test-cert will fail, if ran again within the TTL time of the TXT record made in DO.

It seems that the default TTL of DNS records in DO is 1800, so you cannot reliably get a certificate in testing and then in non-testing within this timeframe.

I know that you can set the TTL on certs in datadog's api on creation time, so maybe using the value provided by "--dns-digitalocean-propagation-seconds" as the record TTL would be a good idea.

dscottboggs:
Fixes #7969.

## Pull Request Checklist

# Purpose
This pull request is to allow setting a custom TTL when using DigitalOcean TXT records for authentication, as well as to lower the default to a more reasonable 60 seconds.

# Rationale
The default used by the dependent `python-digitalocean` library is 1800 seconds (one-half hour) which can cause frustrations when testing out configurations with `--dry-run`.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -70,6 +70,7 @@ Authors
 * [Dominic Cleal](https://github.com/domcleal)
 * [Dominic Lüchinger](https://github.com/dol)
 * [Douglas José](https://github.com/douglasjose)
+* [D. Scott Boggs](https://github.com/dscottboggs)
 * [Erica Portnoy](https://github.com/ohemorange)
 * [Eric Engestrom](https://github.com/1ace)
 * [Eric Rescorla](https://github.com/ekr)
--- a/certbot-dns-digitalocean/certbot_dns_digitalocean/_internal/dns_digitalocean.py
+++ b/certbot-dns-digitalocean/certbot_dns_digitalocean/_internal/dns_digitalocean.py
@@ -7,6 +7,7 @@ import zope.interface
 from certbot import errors
 from certbot import interfaces
 from certbot.plugins import dns_common
+from certbot.compat.os import environ
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


alexzorin:
I agree with the rationale. Consecutive dry-runs within the TTL period can produce confusing results, depending on how the ACME server observes TTLs.

If I can make a suggestion: we may as well lower it all the way to 30s, which is the documented minimum for the DO API.  

It's too bad it can't be lowered to 0s/1s, which would prevent the problem entirely. These are one-shot records, after all.

LoganDark:
It's been 6 months, just because this is unplanned doesn't mean the PR should be left to sit.

It provides a much more useful default, especially for staging, preventing you from having to wait 30 minutes to have a real cert issued.

@bmw please merge

Aidenir:
Any reason this hasn't been merged 1 year later?

dscottboggs:
Yeah I'm genuinely confused as to why this hasn't been merged. Is the change undesired? Or is it simply a matter of no one being available to review and approve it?

LoganDark:
I'm already using a modified version of Certbot with this change and it works perfectly.

alexzorin:
Hello,

The change is complicated by a dependency issue.

The `ttl` kwarg made an appearance in https://github.com/koalalorenzo/python-digitalocean/commit/c2071bc88a9e29e7487eecd0624a86ab0a5cc7da, released in `1.15`. Certbot only requires `>=1.11`. 

On a released stable distro like Debian, that only ships `1.13`, this PR has no effect.

Per [this comment](https://github.com/certbot/certbot/issues/8161#issuecomment-705724583), we could try to bump the required and pinned versions to `1.15` anyway.

Last time I had a crack at doing that, I noticed in https://github.com/koalalorenzo/python-digitalocean/compare/v1.11..v1.15.0 that an extra dependency on `jsonpickle` is introduced, and to avoid that I thought maybe to rewrite this provider using Lexicon, which Certbot already uses for other providers.

Since more users are hitting this, I will try to get an updated answer on how to progress this.

LoganDark:
> On a released stable distro like Debian, that only ships `1.13`, this PR has no effect.

So? It doesn't hurt anything, it only helps people with a recent version. If you merge it you just have to make sure people know to upgrade their digitalocean package using `pip` if they need to use TTL, which is kind of how I setup everything in the first place.

Issue #7701, opened 2020-01-21
DNS plugins don't work if _acme-challenge is a separate zone


patrakov:
## My operating system is (include version):
Debian buster

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
pip

## What's unusual in my setup:

`smb.mydomain.com` is the domain for which I want to get a wildcard. However, my publicly available authoritative PowerDNS server is configured as a slave of a different server that I would rather not touch with certbot. Instead, a separate master zone, `_acme-challenge.smb.mydomain.com`, is configured on PowerDNS.

## I ran this command and it produced this output:
```
certbot ... --non-interactive --authenticator certbot-dns-powerdns:dns-powerdns --certbot-dns-powerdns:dns-powerdns-credentials /etc/letsencrypt/pdns-credentials.ini --expand --domains smb.mydomain.com,*.smb.mydomain.com
```

```
...
2020-01-21 16:45:29,232:DEBUG:urllib3.connectionpool:http://172.31.100.21:8081 "GET /api/v1/servers/localhost/zones/smb.mydomain.com. HTTP/1.1" 200 4094
...
2020-01-21 16:45:29,255:DEBUG:urllib3.connectionpool:http://172.31.100.21:8081 "PATCH /api/v1/servers/localhost/zones/smb.mydomain.com. HTTP/1.1" 204 0
...
Domain: smb.mydomain.com
Type:   unauthorized
Detail: No TXT record found at _acme-challenge.smb.mydomain.com
```

## Certbot's behavior differed from what I expected because:

It made the edit to the wrong zone. It created an `_acme-challenge` TXT record in the `smb.mydomain.com` slave zone, but should have created an `@` TXT record in the `_acme-challenge.smb.mydomain.com` zone.

## Workaround

I have edited the dns_common_lexicon.py file. Replaced this line:

        domain_name_guesses = dns_common.base_domain_name_guesses(domain)

with this one:

        domain_name_guesses = dns_common.base_domain_name_guesses('_acme-challenge.' + domain)

After that, it works as expected.

bmw:
@adferrand or @joohoi, can you comment on this? I'm personally not immediately sure what is the best path forward here.

joohoi:
I think the workaround is good, and handles the use case of pointing the validation domain apex to a server that the user would be comfortable storing the credentials to pretty well. 

[...]

PR #7732, opened 2020-01-30
Support _acme-challenge as a DNS zone apex (#7701)

patrakov:
Multiple DNS plugins tried to determine the correct zone to edit based on
the domain name. However, this fails when _acme-challenge is separated out
to its own zone, perhaps for access control reasons.

Tested on DigitalOcean and with the external Lexicon-based PowerDNS
authenticator.

Maintainers: Please test on Google Cloud DNS and Cloudflare before merging.

## Pull Request Checklist

- [X] If the change being made is to a [distributed component](https://certbot.eff.org/docs/contributing.html#code-components-and-layout), edit the `master` section of `certbot/CHANGELOG.md` to include a description of the change being made.


patrakov:
Well, the testsuite says it is not safe to merge. I will deal with it after FOSDEM.

patrakov:
What's apparently left are pylint errors. They are valid, because the "domain" parameter of add_txt_record() and del_txt_record() methods is now indeed unused, and its role does not correspond to the inline documentation (i.e., it is no longer a "domain to use to look up the zone").

I am not 100% sure what is a public API here and therefore cannot be changed. In some cases (e.g. Cloudflare), the correct solution is to remove the argument and adjust the callers in the tests. But that's a lot of work, so not for today.

bmw:
In general, if the file is under a directory named "_internal" or the relevant code is prefixed with an underscore, it's not part of the public API and can be changed. The one relevant exception to this is we should not modify the API of the "private" abstract methods of `certbot.plugins.dns_common.DNSAuthenticator` like `_perform` as the methods are expected to be overridden in the classes that inherit from `DNSAuthenticator`.

In other cases though, I agree with you that the correct solution is to remove the argument and adjust the callers. For the cases we don't want to modify, we can just add a comment disabling the warning from pylint.

patrakov:
OK, now all the checks pass. But the unused `_domain` parameter in `dns_common_lexicon` is really ugly.

I don't know whether it would be better, instead of documenting it as unused, to remove it and add a decorator for backwards compatibility like this:

```python
def _accept_obsolete_domain_parameter(f):
    """
    Allows a method to accept an optional unused domain parameter as the first argument.
    """
    @functools.wraps(f)
    def wrapper(*args):
        if len(args) == 4:
            args = (args[0], args[2], args[3])
        return f(*args)
    return wrapper
```

bmw:
This is a bit tricky I think. Rather than keeping the unused parameter around forever, I'd like to deprecate it and eventually remove it.

While I don't think we've ever deprecated a function parameter like this, our pattern for deprecating code in our API in general is to use the Python [warnings module](https://docs.python.org/3/library/warnings.html) to issue a `DeprecationWarning` when that functionality is used and have the warning message describe what the programmer should use instead of the deprecated code. After we've been issuing the deprecation warning for a while, in a new major release we'd remove the functionality.

As part of this, all of our DNS plugins that are affected by this would be updated to use the correct functionality and the minimum version of Certbot they require should be updated to make sure that this functionality is available. The way to update the minimum version of Certbot a plugin requires is a bit tricky due to some of our tooling/testing, but you can find an example of how it's done in https://github.com/certbot/certbot/pull/7325.

As for how to accomplish this, I have two ideas:

1. Write code like you described, however, I think we'd need to be confident that this works properly for any valid combination of positional and keyword arguments to the function. This initially seems a bit tricky.
2. If doing that is hard, we could deprecate the function entirely, create new functions like `add_text_record` that accept the new parameters, and have the old functions simply issue a warning and pass the relevant arguments to the new function.

patrakov:
Sorry, will rewrite the new tests in a py27-compatible way now :(

patrakov:
Unfortunately, failed. See e406510 as a commit which preserves the failed attempt.

The problem is with the py27-dns-oldest test, and with certbot plugin version requirements in general: we need a released version which allows omitting the domain parameter. Also we can't issue a DeprecationWarning right away, because of the plugins: they will be tested both against the old certbot version which doesn't understand when one omits the parameter, and against the current certbot master, which would then fail the tests because the plugins hit deprecated functionality.

patrakov:
Oh, and by the way, the reference to #7325 didn't help (TLDR).

patrakov:
OK, I will try once more with "-e dev" inserted in local-oldest-requirements.txt.

patrakov:
The two-step solution doesn't work either, because of lint, which rightfully complains that we don't call add/del_txt_record consistently with its prototype. So let me think more.

patrakov:
The "-e certbot[dev]" was indeed the missing link. Now all the autotests pass. Still, given the amount of changes, I would appreciate some manual testing from devs. I will retest on powerdns and DigitalOcean later today just in case.

patrakov:
Tested with PowerDNS (without any changes to the existing certbot_dns_powerdns plugin), DigitalOcean (with _acme-challenge as a zone) and DNSMadeEasy (using their sandbox).

Issue #7446, opened 2019-10-11
certbot-nginx parse $hostname


cjwijtmans:
## My operating system is (include version):
gentoo

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
portage

## I ran this command and it produced this output:
sudo cerbot --nginx

## Certbot's behavior differed from what I expected because:
certbot-nginx should parse my default server config where i specify $hostname as server_name and list it for renewal.

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:

```
server {
        listen 80 default_server;
        listen [::]:80 default_server;
        server_name $hostname;

        access_log off;
        error_log off;

        return 301 https://$host$request_uri;
}

server {
        listen 443 default_server ssl http2;
        listen [::]:443 default_server ssl http2;
        server_name $hostname;

        include snippets/letsencrypt.conf;
#       ssl_certificate /etc/letsencrypt/live/$hostname/fullchain.pem;
#       ssl_certificate_key /etc/letsencrypt/live/$hostname/privkey.pem;
#       ssl_certificate /etc/ssl/nginx/nginx.pem;
#       ssl_certificate_key /etc/ssl/nginx/nginx.pem;

        access_log off;
        error_log off;

        return 301 werthrtyujwer;
}
```

PR #7738, opened 2020-02-02
[nginx] Parse $hostname in `server_name`

osirisinferi:
Fixes https://github.com/certbot/certbot/issues/7446.

I'm pretty sure I'm properly abusing the try/except statements, but in my mind it was way easier than comparing and ifs and elses when I saw the `remove()` operation :grin:. If there's any objection at all with this programming, please let me know.

Also, I was a little bit tempted to use `socket.getfqdn()`, but didn't change it, because [nginx uses gethostname() too](https://github.com/nginx/nginx/blob/9e07862d6e9d37041a42bbfa34dd1f56ed547505/src/core/ngx_cycle.c#L196). Perhaps that's because C doesn't have a `getfqdn()` function, but hey, if `gethostname()` is good enough for C, it's probably good enough for Python? :stuck_out_tongue: Also, it's a small change if `getfqdn()` is prefered.

### DIFF
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -601,7 +601,6 @@ class NginxConfigurator(common.Installer):
             try:
                 vhost.names.remove("$hostname")
                 vhost.names.add(socket.gethostname())
-                breakpoint()
             except KeyError:
                 pass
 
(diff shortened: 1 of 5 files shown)


osirisinferi:
I can *try*..

osirisinferi:
Existing tests were failing b/c of the added virtualhost.. Didn't realise this locally, as my system for some reason doesn't run all the tests (something about a missing `libaugeas` in the VirtualEnv?) and `python configuration_test.py` didn't pick those errors up..

PR #5467, opened 2018-01-24
Support Openresty in the NGINX plugin

alokinsoft:
@ohemorange Could you review this pull request for possible inclusion in certbot? This is related to issue #4919 

We later found the pull request #4833 and make the necessary adjustments based on your comments.

Issue #4716, opened 2017-05-23
EPEL SELinux problems


bmw:
Opening an issue here because I feel better about opening potentially spurious bug reports here than on Bugzilla:

Some users are reporting that SELinux is causing the Apache plugin to fail domain validation challenges on CentOS 7. ~~Doing my own testing, this doesn't happen with a default configuration.~~ More details can be found at: https://community.letsencrypt.org/t/letsencrypt-suddenly-failing-to-create-temporary-web-root-for-verification/11739

@hogarthj, what do you think?

EDIT: I know very little about SELinux so the testing I thought I did should be ignored as I think I did it wrong.

hogarthj:
I'm configuring selinux to have the certificates in /etc/letsencrypt/(archive|live) correct [here](https://src.fedoraproject.org/cgit/rpms/certbot.git/tree/certbot.spec#n192) but have I missed something happening in /var/lib I need to pay attention to?

I'm not using the apache plugin myself but rather webroot which works fine with this configuration. 

If apache needs to serve out of /var/lib/something we should label correctly there so that it will serve up the files - httpd_sys_content_t if memory serves ...

At any rate they should not be doing apachectl to restart httpd on CentOS7/Fedora/RHEL7 as that's not going to be running in the appropriate selinux context or under the supervision of systemd properly.

I really want to get 0.14.1 to stable as updates have been pending a while, but I'll do some more specific testing of the plugin and get an update prepared for afterwards if I can reproduce.

hogarthj:
Okay a quick test of the plugin and I see this:
```
#============= httpd_t ==============
# audit(1495574784.270:117):
#  scontext="unconfined_u:system_r:httpd_t:s0" tcontext="unconfined_u:object_r:etc_t:s0"
#  class="file" perms="write"
#  comm="httpd" exe="" path="/etc/letsencrypt/.certbot.lock"
#  message="type=AVC msg=audit(1495574784.270:117): avc:  denied  { write } for
#   pid=9306 comm="httpd" path="/etc/letsencrypt/.certbot.lock" dev="xvda1"
#   ino=8414283 scontext=unconfined_u:system_r:httpd_t:s0
#   tcontext=unconfined_u:object_r:etc_t:s0 tclass=file"
# audit(1495574798.901:118):
#  scontext="unconfined_u:system_r:httpd_t:s0" tcontext="unconfined_u:object_r:etc_t:s0"
#  class="file" perms="write"
[...]

PR #7740, opened 2020-02-03
Wait for Nginx restart to complete when using Nginx plugin

### LINKED ISSUE #7422: Nginx installer not properly reloading configuration
Detailed conversation could be found here:
https://community.letsencrypt.org/t/cannot-renew-create-a-new-cert-when-i-had-no-issue-previously/103040/

Ubuntu 18.04.3 LTS
Nginx 1.17.4 (note, this isn't the version of nginx that's part of Ubuntu's repo's)

`sudo certbot renew --cert-name ackis.duckdns.org --debug-challenges --dry-run`

```

Processing /etc/letsencrypt/renewal/ackis.duckdns.org.conf
Plugins selected: Authenticator nginx, Installer nginx
http-01 challenge for ackis.duckdns.org
Waiting for verification...

Challenges loaded. Press continue to submit to CA. Pass "-v" for more info about
challenges.
Cleaning up challenges
Attempting to renew cert (ackis.duckdns.org) from /etc/letsencrypt/renewal/ackis.duckdns.org.conf produced an unexpected error: Failed authorization procedure. ackis.duckdns.org (http-01): urn:ietf:params:acme:error:unauthorized :: The client lacks sufficient authorization :: Invalid response from https://ackis.duckdns.org/.well-known/acme-challenge/7bO_DNxtDDyO_hPdRJcpGWEJHaLTRwtTsMWpWtQREDE [174.3.126.96]: "<html>\r\n<head><title>404 Not Found</title></head>\r\n<body>\r\n<center><h1>404 Not Found</h1></center>\r\n<hr><center>nginx</center>\r\n". Skipping.
All renewal attempts failed. The following certs could not be renewed:
  /etc/letsencrypt/live/ackis.duckdns.org/fullchain.pem (failure)

- - - - - - - - - - - - - - - - - - -  - - - - - - - - - - - - - - - - - - - - -
** DRY RUN: simulating 'certbot renew' close to cert expiry

All renewal attempts failed. The following certs could not be renewed:
  /etc/letsencrypt/live/ackis.duckdns.org/fullchain.pem (failure)
** DRY RUN: simulating 'certbot renew' close to cert expiry
1 renew failure(s), 0 parse failure(s)

IMPORTANT NOTES:

   Domain: ackis.duckdns.org
   Type:   unauthorized
   Detail: Invalid response from
   https://ackis.duckdns.org/.well-known/acme-challenge/7bO_DNxtDDyO_hPdRJcpGWEJHaLTRwtTsMWpWtQREDE
   [174.3.126.96]: "<html>\r\n<head><title>404 Not
   Found</title></head>\r\n<body>\r\n<center><h1>404 Not
   Found</h1></center>\r\n<hr><center>nginx</center>\r\n"

```

Certbot should have renewed the certificate.

```
# HTTP server - redirect to HTTPS
server {
        listen                                  80;
        listen                                  [::]:80;
        server_name                             www.ackis.duckdns.org ackis.duckdns.org;

        location / {
                return                          301 https://ackis.duckdns.org$request_uri;
        }

        # Workaround LE and certbot not working with nginx
        location /.well-known/acme-challenge/ {
                                                root /var/www/letsencrypt;
        }

        access_log                              syslog:server=localhost,tag=nginx_access_internet,severity=info;
        error_log                               syslog:server=localhost,tag=nginx_error_internet;

        add_header                              Strict-Transport-Security "max-age=31536000;";
        add_header                              X-Frame-Options SAMEORIGIN;
        add_header                              X-Content-Type-Options nosniff;
        add_header                              X-XSS-Protection "1; mode=block";
}

# www domain - redirect to domain without www
server {
        listen                                  443 ssl;
        listen                                  [::]:443 ssl;
        server_name                             www.ackis.duckdns.org;

        location / {
                return                          301 https://ackis.duckdns.org$request_uri;
        }

        access_log                              syslog:server=localhost,tag=nginx_access_internet,severity=info;
        error_log                               syslog:server=localhost,tag=nginx_error_internet;

        ssl_certificate                         /etc/letsencrypt/live/www.ackis.duckdns.org/fullchain.pem;
        ssl_certificate_key                     /etc/letsencrypt/live/www.ackis.duckdns.org/privkey.pem;

        server_tokens                           off;
        etag                                    off;

        add_header                              Strict-Transport-Security "max-age=31536000;";
        add_header                              X-Frame-Options SAMEORIGIN;
        add_header                              X-Content-Type-Options nosniff;
        add_header                              X-XSS-Protection "1; mode=block";
}

server {
        listen                                  443 ssl default_server;
        listen                                  [::]:443 ssl;
        server_name                             ackis.duckdns.org;

        access_log                              syslog:server=localhost,tag=nginx_access_internet,severity=info;
        error_log                               syslog:server=localhost,tag=nginx_error_internet;

        ssl_certificate                         /etc/letsencrypt/live/ackis.duckdns.org/fullchain.pem; # managed by Certbot
        ssl_certificate_key                     /etc/letsencrypt/live/ackis.duckdns.org/privkey.pem; # managed by Certbot

        server_tokens                           off;
        etag                                    off;

        add_header                              Strict-Transport-Security "max-age=31536000;";
        add_header                              X-Frame-Options SAMEORIGIN;
        add_header                              X-Content-Type-Options nosniff;
        add_header                              X-XSS-Protection "1; mode=block";

        location ~* /\.\./ {
                deny                            all;
                return                          404;
        }

        location ~* "^(?:.+\.(?:htaccess|make|txt|test|markdown|md|engine|inc|info|install|module|profile|po|sh|.*sql|theme|tpl(?:\.php)?|xtmpl)|code-style\.pl|/Entries.*|/Re$
                return                          404;
        }

        location = /favicon.ico {
                try_files                       /favicon.ico =204;
        }

        location / {
                root                            /var/www/internet;
                index                           index.html;
        }
        location /nginx_status {
                access_log                      syslog:server=localhost,tag=nginx_access_admin,severity=info;
                error_log                       syslog:server=localhost,tag=nginx_error_admin;
                allow                           192.168.0.0/24;
                deny                            all;

                auth_basic                      "Restricted access";
                auth_basic_user_file            /etc/nginx/auth/admin.htpasswd;

                stub_status                     on;
        }

}

```

mustanggb:
Fixes #7422.

### DIFF
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -7,7 +7,6 @@ import socket
 import subprocess
 import tempfile
 import time
-import requests
 
 import OpenSSL
 import pkg_resources
@@ -1111,23 +1110,35 @@ class NginxConfigurator(common.Installer):
             # when they are all complete.
             http_doer.add_chall(achall, i)
 
+        # Add sentinel vhost used to detect successful nginx restart.
+        sentinel_socket_path = "/var/lib/letsencrypt/nginx-sentinel.sock"
+        sentinel_vhost = [["server"],
+                             [["listen", " ", "unix:", sentinel_socket_path],
+                             [["location", " ", "/"],
+                                 [["return", " ", "200", " ", "letsencrypt-sentinel"]]]]]
+        root = self.parser.config_root
+        main = self.parser.parsed[root]
+        for line in main:
+            if line[0] == ["http"]:
+                body = line[1]
+                body.append(sentinel_vhost)
+
         http_response = http_doer.perform()
         # Must restart in order to activate the challenges.
         # Handled here because we may be able to load up other challenge types
         self.restart()
 
-        # Wait for responses to be valid.
+        # Wait for valid sentinel response to know nginx restart is complete.
         timeout = time.time() + 120
-        achalls_to_validate = achalls[:]
         while time.time() < timeout:
-            for achall in achalls_to_validate:
-                url = "http://127.0.0.1/" + http_doer.get_validation_path(achall)
-                headers = {"Host": achall.domain}
-                response = requests.get(url, headers=headers)
-                if response.status_code == 200:
-                    achalls_to_validate.remove(achall)
-            if len(achalls_to_validate) == 0:
-                break
+            if os.path.exists(sentinel_socket_path):
+                client = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+                client.connect(sentinel_socket_path)
+                request = "GET / HTTP/1.1\r\nHost: letsencrypt-sentinel\r\n\r\n"
+                client.send(request.encode('utf-8'))
+                response = client.recv(1024).decode('utf-8').split(' ', 2)
+                if int(response[1]) == 200:
+                    break
             time.sleep(1)
 
         # Go through all of the challenges and assign them to the proper place
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -1121,8 +1121,8 @@ class NginxConfigurator(common.Installer):
         achalls_to_validate = achalls[:]
         while time.time() < timeout:
             for achall in achalls_to_validate:
-                url = 'http://127.0.0.1/' + http_doer.get_validation_path(achall)
-                headers = {'Host': achall.domain}
+                url = "http://127.0.0.1/" + http_doer.get_validation_path(achall)
+                headers = {"Host": achall.domain}
                 response = requests.get(url, headers=headers)
                 if response.status_code == 200:
                     achalls_to_validate.remove(achall)
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -1113,7 +1113,8 @@ class NginxConfigurator(common.Installer):
         http_response = http_doer.perform()
 
         # Store old workers.
-        worker_pids = list(map(int, subprocess.check_output(["pidof", "nginx: worker process"]).split()))
+        output = subprocess.check_output(["pidof", "nginx: worker process"])
+        worker_pids = list(map(int, output.split()))
 
         # Must restart in order to activate the challenges.
         # Handled here because we may be able to load up other challenge types
@@ -1121,13 +1122,14 @@ class NginxConfigurator(common.Installer):
 
         # Wait for old workers to terminate.
         timeout = time.time() + 120
-        while time.time() < timeout:
-            for worker_pid in reversed(worker_pids):
-                if subprocess.call(["kill", "-0", str(worker_pid)], stderr=open(os.devnull, "w")) != 0:
-                    worker_pids.remove(worker_pid)
-            if len(worker_pids) == 0:
-                break
-            time.sleep(1)
+        with open(os.devnull, "w") as devnull:
+            while time.time() < timeout:
+                for worker_pid in reversed(worker_pids):
+                    if subprocess.call(["kill", "-0", str(worker_pid)], stderr=devnull) != 0:
+                        worker_pids.remove(worker_pid)
+                if len(worker_pids) == 0:
+                    break
+                time.sleep(1)
 
         # Go through all of the challenges and assign them to the proper place
         # in the responses return value. All responses must be in the same order
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -7,7 +7,8 @@ import socket
 import subprocess
 import tempfile
 import time
-import psutil
+import requests
+
 
 import OpenSSL
 import pkg_resources
@@ -1112,25 +1113,21 @@ class NginxConfigurator(common.Installer):
             http_doer.add_chall(achall, i)
 
         http_response = http_doer.perform()
-
-        # Store old workers.
-        worker_pids = []
-        for proc in psutil.process_iter():
-            cmdline = proc.cmdline()
-            if cmdline and cmdline[0] == "nginx: worker process":
-                worker_pids.append(proc.pid)
-
         # Must restart in order to activate the challenges.
         # Handled here because we may be able to load up other challenge types
         self.restart()
 
-        # Wait for old workers to terminate.
+        # Wait for responses to be valid.
         timeout = time.time() + 120
+        achalls_to_validate = achalls[:]
         while time.time() < timeout:
-            for worker_pid in reversed(worker_pids):
-                if not psutil.pid_exists(worker_pid):
-                    worker_pids.remove(worker_pid)
-            if len(worker_pids) == 0:
+            for achall in achalls_to_validate:
+                url = 'http://127.0.0.1' + http_doer._get_validation_path(achall)
+                headers = {'Host': achall.domain}
+                response = requests.get(url, headers=headers, allow_redirects=False)
+                if response.status_code == 200:
+                    achalls_to_validate.remove(achall)
+            if len(achalls_to_validate) == 0:
                 break
             time.sleep(1)
 
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -1123,7 +1123,7 @@ class NginxConfigurator(common.Installer):
             for achall in achalls_to_validate:
                 url = 'http://127.0.0.1' + http_doer._get_validation_path(achall)
                 headers = {'Host': achall.domain}
-                response = requests.get(url, headers=headers, allow_redirects=False)
+                response = requests.get(url, headers=headers)
                 if response.status_code == 200:
                     achalls_to_validate.remove(achall)
             if len(achalls_to_validate) == 0:
--- a/certbot-nginx/certbot_nginx/_internal/configurator.py
+++ b/certbot-nginx/certbot_nginx/_internal/configurator.py
@@ -7,6 +7,7 @@ import socket
 import subprocess
 import tempfile
 import time
+import psutil
 
 import OpenSSL
 import pkg_resources
@@ -1113,8 +1114,11 @@ class NginxConfigurator(common.Installer):
         http_response = http_doer.perform()
 
         # Store old workers.
-        output = subprocess.check_output(["pidof", "nginx: worker process"])
-        worker_pids = list(map(int, output.split()))
+        worker_pids = []
+        for proc in psutil.process_iter():
+            cmdline = proc.cmdline()
+            if cmdline and cmdline[0] == "nginx: worker process":
+                worker_pids.append(proc.pid)
 
         # Must restart in order to activate the challenges.
         # Handled here because we may be able to load up other challenge types
@@ -1122,14 +1126,13 @@ class NginxConfigurator(common.Installer):
 
         # Wait for old workers to terminate.
         timeout = time.time() + 120
-        with open(os.devnull, "w") as devnull:
-            while time.time() < timeout:
-                for worker_pid in reversed(worker_pids):
-                    if subprocess.call(["kill", "-0", str(worker_pid)], stderr=devnull) != 0:
-                        worker_pids.remove(worker_pid)
-                if len(worker_pids) == 0:
-                    break
-                time.sleep(1)
+        while time.time() < timeout:
+            for worker_pid in reversed(worker_pids):
+                if not psutil.pid_exists(worker_pid):
+                    worker_pids.remove(worker_pid)
+            if len(worker_pids) == 0:
+                break
+            time.sleep(1)
 
         # Go through all of the challenges and assign them to the proper place
         # in the responses return value. All responses must be in the same order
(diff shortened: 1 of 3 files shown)


schoen:
Hi @mustanggb,

I remember having had some code that depended on `psutil` in the past which we were subsequently able to remove (which I think may have helped improve Certbot's compatibility because `psutil` may be unavailable on some platforms). I'd like to check in with others about whether there's a consensus about whether it's OK to re-introduce that dependency.

schoen:
The history of our removing `psutil` deliberately is found at

https://github.com/certbot/certbot/issues/3106

Is there any other way to do this without reintroducing this dependency? Perhaps we could use `psutil` if it happens to be present, but fall back to a fixed delay or another method if it's not?

mustanggb:
@schoen thanks for commenting, I did try to look for why `psutil` was removed from "core", but didn't notice any particular reason, so assumed it was just to remove some dependencies for easier maintainability/portability.

I deliberately didn't squash my commits so people could have a look at the approaches I tried.

`psutil.pid_exists(worker_pid)` could be replaced with `subprocess.call(["kill", "-0", str(worker_pid)], stderr=subprocess.DEVNULL)` as I had originally, or potentially even something like `os.kill(worker_pid, 0)` (untested), so that one's not too much of an issue.

However the issue I had was the tests complaining that `pidof` wasn't available on Azure/Windows, hence why I gave `psutil` a try.

Also I didn't want to spend too much time on it if the overall approach wasn't going to be acceptable.

If people are happy with the approach a cross-platform method of achieving `pidof` could potentially be sought. I'm no Python expert, so suggestions welcome.

schoen:
Thanks! I realize this is a thorny problem but so far I haven't come across a consensus about what to do about it. It does seem like the maintainers don't want to re-add the `psutil` dependency as discussed back in #3106; that may leave us without a clear path forward for this issue.

adferrand:
Hello! If you want, I can search for Windows native alternatives for `pidof` and associated operations on pids.

If there is a solution, we will need to implement it in the `certbot.compat` package, in order to expose to the rest of the codebase several platform agnostic methods.

schoen:
@adferrand, I think that would probably be helpful. A related question is whether there is a portable Unix equivalent that works on non-Linux Unix systems (which I think was a problem with `psutil` as well, if I'm remembering correctly from a few years ago).

@bmw, do you have a thought on an appropriate solution to #7422 or what characteristics such a solution ought to have?

adferrand:
OK. My feedback so far after some digging. I think that without `psutil`, it will be hard to have something working on every system. `psutil` is doing a hard work to find the native solution, but as @schoen said, failed to do so in the past. Maybe it is better now, but even in this case, we would add a new dependency on the nginx plugin, that will make integration more difficult.

On Windows specifically, we can have a decent solution thanks to `pywin32`, and reproduce what `psutil` does, but it still a significant complexity: it consistents essentially in trying to open the process associated to PID, and assert on if we can retrieve the process handle or not, and what is its status if it is retrieved.

I would like to propose to study another approach, that would avoid to manipulate PIDs and so to require a deep knowledge of the actual platform that is running Nginx.

Why not trying to make HTTP requests to the local Nginx instance ? If the results of these requests are usable to know when Nginx is effectively restarted, I think this solution will be way more maintainable in the long term.

mustanggb:
Appreciate the replies.

> Why not trying to make HTTP requests to the local Nginx instance ?

Mainly just because I thought the original method would be a more robust solution, but I've updated the pull request to do it this way instead, please let me know your feedback.

bmw:
> @bmw, do you have a thought on an appropriate solution to #7422 or what characteristics such a solution ought to have?

I think this is tricky and I'm not sure there is a solution that will work 100% of the time.

While it's probably rare, the server might not be listening on all interfaces and will not be accessible through localhost.

As for the `psutil` approach, the concerns I have are:

* The issues we've had in the past with the library
* The reliability of the string "nginx: worker process" across all nginx packages and forks like [Openresty](https://github.com/certbot/certbot/pull/5467).
* Complexity in making the dependency conditional and/or adding another dependency that requires compilation.

If we thought it was significantly more reliable than connecting to localhost, I'd be potentially open to adding the `psutil` dependency again, but I personally suspect this is not the case.

If we cannot do this 100% reliably, I personally think the most important thing here in general is to make sure we have good behavior when this is not going to work. A couple suggestions I have related to this are:

1. If nginx is not accessible over localhost, the current code raises an exception from `requests.get` which I think it should not do.
2. If we resolve the issue of the exception and nginx is not accessible over localhost, we now wait 2 minutes instead of 3 seconds. I think this is pretty bad UX for someone running Certbot interactively where in almost all cases the 3 second sleep is long enough. What if we checked if the server was accessible over localhost before reloading nginx and only wait for the challenge response to become available if we could previously connect to it? This would mean this logic is skipped for servers that weren't listening on port 80 before we set up the challenge, but I personally think it's worth it to avoid waiting for 2 minutes when we'll never be able to connect.

mustanggb:
> 1. If nginx is not accessible over localhost, the current code raises an exception from requests.get which I think it should not do.

I'm not familiar with such a setup, do you mean a non-default port, listening on a socket (I don't think Nginx has this feature), or something else; you could elaborate? If it's just a different port could we add an argument to set this if it needs to be changed?

> 2. If we resolve the issue of the exception and nginx is not accessible over localhost, we now wait 2 minutes instead of 3 seconds. I think this is pretty bad UX for someone running Certbot interactively where in almost all cases the 3 second sleep is long enough. What if we checked if the server was accessible over localhost before reloading nginx and only wait for the challenge response to become available if we could previously connect to it? This would mean this logic is skipped for servers that weren't listening on port 80 before we set up the challenge, but I personally think it's worth it to avoid waiting for 2 minutes when we'll never be able to connect.

Agree.

bmw:
Most people's listen directives look like:
```
listen 80;
```
This configures nginx to listen to port 80 on all interfaces.

Using the output of a tool like `ifconfig`, if I configure nginx to listen on the IP address of a specific interface (other than the loopback interface of `127.0.0.1`), it will no longer be accessible through `localhost` but still accessible to users of the site. The modified directive would look something like:
```
listen 1.2.3.4:80;
```
I wasn't thinking of a problem of using a non-default port, but we should probably make that work too. The nginx plugin should currently respect the value of `--http-01-port` when setting up the challenge so we should probably direct our request there as well.

I also wasn't thinking of listening on a socket, although that is also supported by nginx: https://nginx.org/en/docs/http/ngx_http_core_module.html#listen

alexzorin:
Just a shower thought: you don't need to verify that the challenge path is accessible, we only need to verify that nginx reloaded.

To that end, one could just get the configurator to add a sentinel vhost as part of the challenge solver, such as:

```nginx
server {
    listen unix:/var/lib/letsencrypt/nginx-sentinel.sock;
    location / { return 200 "<random sentinel value>"; }
}
```

This might dodge some of the issues that @bmw is worried about.

mustanggb:
> one could just get the configurator to add a sentinel vhost as part of the challenge solver

Bingo! This seems like the best of both worlds, I like this idea better as well as it's a closer check of an actual reload, like you say, a vhost could be giving a 200 response even before a reload has completed.

mustanggb:
Well it seems a reload does work fine.

Updated with the sentinel approach; not expecting tests to pass yet.

mustanggb:
Okay, tests did better than expected, however I had no idea why the `integration` one is failing.

mustanggb:
Any further input from maintainers, possible to have a milestone?

schoen:
@bmw, could you opine on this? It seems that @alexzorin has had a clever idea and @mustanggb has implemented it.

(One thing I wonder about is whether this could reduce our Windows compatibility, since I didn't think Windows supported Unix domain sockets.)

bmw:
That is a clever idea from alexzorin.

In addition to making Windows support harder though, I'm a little concerned about file system permissions or mandatory access control issues like those described in https://github.com/certbot/certbot/issues/4716. We used the directory more in the past, but we currently don't configure nginx to use our working directory at all except when the config previously wasn't listening on necessary port for the challenge, we conditionally reference the nonexistent directory [http_01_nonexistent](https://github.com/certbot/certbot/blob/fc7e5e8e6060d9e0df2e704a20103d5c0f456925/certbot-nginx/certbot_nginx/_internal/http_01.py#L155).

I'm sorry I keep popping in here trying to poke holes in the proposed approaches, but we currently have [nearly half a million nginx users](https://gist.github.com/bmw/0ef8f41d6dc70ec76d6545b9af3a9c78) for whom what we currently have is working. That's a lot of nginx configurations doing who knows what and I don't want to break their ability to automatically renew certificates. While it'd be great for us to also support the few configurations who have reported hitting #7422, landing anything here with any significant risk of regressions personally makes me a little nervous.

I don't really care what approach is taken here if we're confident it'll work, but out of what's been proposed so far, I personally think I like using localhost the best with the changes I described at https://github.com/certbot/certbot/pull/7740#issuecomment-584386670.

Two other ideas are:

1. Offer a configurable flag like `--nginx-sleep-seconds`. It's ugly and there's a discoverability problem for new users who hit #7422, but it gives a way for the plugin to work reliably for any nginx configuration.
2. Don't change anything here and close #7422 as wontfix. People can still do things like using the nginx installer with the webroot authenticator if they want.

adferrand:
I would like to give back my reasoning here, trying to avoid further nginx configuration, to include windows compatibility, and various situations. Do not hesitate to explain me why I am wrong.

So, given the port used for the HTTP-01 challenge, as defined by the `--http-01-port`. The nginx plugin is supposed to ensure that a vhost is listening to this port, and define a path in it to the challenge. The vhost may pre-exist (usually the case, because it is the standard `80` HTTP port), or not pre-exist, and the socket may be available for every interface, or only specific ones.

I suppose that we could test the availability of the socket (and so state that nginx is currently running or not) by trying to bind from certbot to `0.0.0.0` on the HTTP-01 challenge port, and see if it possible or not. If it is not, then the socket is currently used at one or more interfaces accessible to the system, and so nginx is running.

Now, the process to see that nginx restarted. We can be in the situation where nginx was listening to this port before the restart or not.
1) (pre-perform) check for socket availability
2) reconfigure nginx
3) restart it
4) if the socket was not available, wait for it to be released
5) wait for the socket to not available again
6) assume nginx has restarted, proceed to perform the challenge

In theory, we avoid any further configuration to Nginx, like the unix socket, increasing compatibility for Windows.

Considering that the current situation (the plugin does not wait) is mostly working, I think the all process should be wrapped by a global short timeout, typically 30s. If timeout is reached, the plugin assume nginx has restarted and continue.

mustanggb:
@bmw 
I'm not a Windows user, but from https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ it looks like unix sockets are supported?

If the directory permissions are a concern could always put it somewhere else, `/var/run` perhaps, or even `/tmp`?

adferrand:
> @bmw
> I'm not a Windows user, but from https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ it looks like unix sockets are supported?
> 
> If the directory permissions are a concern could always put it somewhere else, `/var/run` perhaps, or even `/tmp`?

This is available since Windows 10 1803. I could not find information about the Server flavor of Windows 10, which is Windows Server 2019. And we need to cover all non-EOL versions of Windows, so up to Windows Server 2012.

bmw:
@adferrand, how would you propose that your recent suggestion be implemented in a platform independent way and why do you like it better than your previous suggestion of trying to connect to nginx on localhost?

adferrand:
Well, about trying to connect to Nginx on localhost may not work depending on how Nginx is configured: it may just not listen to the loopback interface (whereas very unlikely in a standard configuration, but possible). Given VirtualHost features, the behavior of an HTTP request could also be very different depending on the actual `Host` header shipped with the request.

However, I may be wrong, but I would expect that trying to bind a port to all interfaces will fail if any interface is already bound for this port: this failure would be a good indication that Nginx has actually restarted.

About how to implement that, the Python `socket.bind()` is very likely platform independant, has it is already used by Certbot for the `standalone` module for instance, and is working both on Linux, Mac OS and Windows.

bmw:
I also could be mistaken, but the two problems I see there are:

1. If nginx was already listening on port 80, because we [reload nginx rather than restart it](https://github.com/certbot/certbot/blob/06599a1e18cd809dc4db26990da67fcdfc3316d4/certbot-nginx/certbot_nginx/_internal/configurator.py#L1172) nginx will never release the port so we won't be able to see any change here.
2. If nginx was not already listening on port 80 and we regularly try to bind to it, we may prevent nginx from binding it when it restarts causing it to crash.

mustanggb:
Won't fix isn't an option, because as it stands the nginx plugin is useless.

So what solution would you like to use?

bmw:
@mustanggb, please keep your comments civil and constructive if you are going to continue posting to this repo. Calling an open source project that has been worked on by hundreds of people and whose nginx plugin is successfully being used by hundreds of thousands of people useless is not appropriate.

As for a path forward, I'm personally just leaning towards an `--nginx-sleep-seconds` flag for now. It's not at all elegant, however, it's a simple fix that provides a way to resolve this problem 100% of the time.

If other people who have offered suggestions in this thread have differing opinions, I'd be interested to hear them.

bmw:
`--nginx-sleep-seconds` was implemented in https://github.com/certbot/certbot/pull/8163 and will be included in our release next week so I'm closing this PR.

Thanks for getting things started here and triggering the discussion around this issue.

PR #7742, opened 2020-02-04
Force non restrictive umask

### LINKED ISSUE #7741: Apache plugin should work with restrictive umask
This problem was initially described at https://github.com/certbot/certbot/issues/1761#issuecomment-581678866.

##  I ran this command and it produced this output:
```
# umask 0077
# certbot -d example.org --apache
Plugins selected: Authenticator apache, Installer apache
Registering without email!
Obtaining a new certificate
http-01 challenge for example.org
Enabled Apache rewrite module
Waiting for verification...
Challenge failed for domain example.org
http-01 challenge for example.org
Cleaning up challenges
Some challenges have failed.

IMPORTANT NOTES:

   Domain: example.org
   Type:   unauthorized
   Detail: Invalid response from
   http://example.org/.well-known/acme-challenge/dYsk27k0uIUgA1QOewleMrShVzv_BLZx7nP5H5eDfww
   [2604:a880:2:d0::578:6001]: "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD
   HTML 2.0//EN\">\n<html><head>\n<title>403
   Forbidden</title>\n</head><body>\n<h1>Forbidden</h1>\n<p"

```

The challenge failed.

## Discussion

The problem is we're not accounting for the possibility of a restrictive umask in the Apache plugin when setting up the challenge. We should fix this like we do in the webroot plugin.

~~I also think we should check if we have the same problem in the nginx plugin.~~

osirisinferi:
Fixes #7741

### DIFF
--- a/certbot-apache/certbot_apache/_internal/http_01.py
+++ b/certbot-apache/certbot_apache/_internal/http_01.py
@@ -1,5 +1,6 @@
 """A class that performs HTTP-01 challenges for Apache"""
 import logging
+import errno
 
 from acme.magic_typing import List  # pylint: disable=unused-import, no-name-in-module
 from acme.magic_typing import Set  # pylint: disable=unused-import, no-name-in-module
--- a/certbot-apache/certbot_apache/_internal/http_01.py
+++ b/certbot-apache/certbot_apache/_internal/http_01.py
@@ -169,8 +169,14 @@ class ApacheHttp01(common.ChallengePerformer):
     def _set_up_challenges(self):
         if not os.path.isdir(self.challenge_dir):
             old_umask = os.umask(0o022)
-            filesystem.makedirs(self.challenge_dir, 0o755)
-            os.umask(old_umask)
+            try:
+                filesystem.makedirs(self.challenge_dir, 0o755)
+            except OSError as exception:
+                if exception.errno not in (errno.EEXIST, errno.EISDIR):
+                    raise errors.PluginError(
+                        "Couldn't create root for http-01 challenge")
+            finally:
+                os.umask(old_umask)
 
         responses = []
         for achall in self.achalls:
--- a/certbot-apache/certbot_apache/_internal/http_01.py
+++ b/certbot-apache/certbot_apache/_internal/http_01.py
@@ -168,7 +168,9 @@ class ApacheHttp01(common.ChallengePerformer):
 
     def _set_up_challenges(self):
         if not os.path.isdir(self.challenge_dir):
+            old_umask = os.umask(0o022)
             filesystem.makedirs(self.challenge_dir, 0o755)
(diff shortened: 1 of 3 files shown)


certbot-apache/certbot_apache/_internal/http_01.py
```diff
@@ -168,7 +168,9 @@ def _relevant_vhosts(self):
 
     def _set_up_challenges(self):
         if not os.path.isdir(self.challenge_dir):
+            old_umask = os.umask(0o022)
             filesystem.makedirs(self.challenge_dir, 0o755)
+            os.umask(old_umask)
```
m0namon: Should we wrap this in a try/finally?

osirisinferi:
If possible, I'd like to get some pointers on where to start with this missing coverage. I'm just terrible at even understanding unittest a little bit. For example, my few lines of codes makes the coverage test fail, apparently because I've got a branching try/except.
So I thought, lets look at the other example of the umask fixing code where I got the code from in the first place, `webroot.py`. However, for some reason the function `_create_challenge_dirs()` isn't covered in any test as far as I can see.. But even so, `webroot.py` gets a 100 % score for coverage?
In any case, I've got no idea where to start on how to test a restrictive umask with unittest in Python.

m0namon:
`_create_challenge_dirs()` isn't tested on its own, but the caller, `webroot.perform` is tested many times. The relevant equivalent tests for the `webroot` plugin seems to be around here:
https://github.com/certbot/certbot/blob/master/certbot/tests/plugins/webroot_test.py#L135
https://github.com/certbot/certbot/blob/master/certbot/tests/plugins/webroot_test.py#L149

Hope that helps!

osirisinferi:
@m0namon It seems `tox -e py27-cover` isn't happy enough with the following:

```
    @mock.patch("certbot_apache._internal.http_01.filesystem.makedirs")
    def test_failed_makedirs(self, mock_makedirs):
        mock_makedirs.side_effect = OSError(errno.EACCES, "msg")
        self.http.perform()
```

I.e., it still fails with lines 175-177 missing, same as before adding those lines of code...

osirisinferi:
@schoen @bmw Could one of you perhaps be of any assistance please?

m0namon:
@osirisinferi For this tests to hit the right code pathways, it looks like you'll have to install some mock challenges for it to perform. You can see this being done for the rest of the http01 tests in `common_perform_tests`.  it might work to just do `self.http.add_chall(mock.Mock())` to set up a challenge-- It can just be a mock object since you're expecting a failure anyways.

osirisinferi:
@m0namon Python wasn't happy about a mock challenge, but just adding one worked like a charm, thanks for the hint!

Issue #6604, opened 2018-12-14
Make mock a test dependency of Certbot


bmw:
Our packages currently add `mock` to `install_requires`. Here is [one example](https://github.com/certbot/certbot/blob/346a4246398b72d27b74b537142ee05073e71cd0/setup.py#L42) but I think this is done in all/most of our packages.

Really, we should only need `mock` when running tests, so why not make it a test dependency like we do for `pytest` [here](https://github.com/certbot/certbot/blob/346a4246398b72d27b74b537142ee05073e71cd0/acme/setup.py#L83)? Us doing this was requested by our PPA maintainers.

We should look through very old commits and the referenced PRs because I believe Kuba had a reason for us doing things that way. I think it's no longer necessary given `test_requires` is working for us, but it'd good to verify there's not something I have forgotten.

FelixSchwarz:
+1 from a Fedora packager :-)

bmw:
Hi @FelixSchwarz. Thanks for the work you've been doing trying to get Certbot packaged in EPEL 8 :smile:

We're stretched pretty thin at the moment, but don't hesitate to ask if you have any questions or suggestions that would make packaging significantly easier for you. You're welcome to communicate with us over GitHub but you can also stop by [our Mattermost instance](https://certbot.eff.org/docs/contributing.html#asking-for-help) or the #certbot-devel IRC on Freenode which it is connected to.

As for this issue, unless it's a significant hurdle for you, we probably won't get to it soon especially since `test_requires` is now deprecated so we'll have to take another approach to solving this problem (see https://github.com/pypa/setuptools/issues/1684), but thanks for expressing your interest in this issue.

FelixSchwarz:
No worries I think we can carry that patch just fine.

ekohl:
Also note that since Python 3.3 mock is part of `unittest`. It would be preferable to use that instead of an external dependency when available.

bmw:
I'm going to close this issue in favor of https://github.com/certbot/certbot/issues/7886 which lays out a plan for removing our dependency on `mock` entirely.

PR #7755, opened 2020-02-07
acme: no mock dependency at runtime

FelixSchwarz:
The first commit moves the `mock` dependency to `test_requires` (`acme` component only). This is basically a template for removing the `mock` runtime dependency in other parts of the code (see also #6604).

The second commit removes the dependency on `mock` altogether for Python >= 3.3. Once Python 2 support is dropped we can also remove the `try/except`. Doing so tree-wide might be bit controversial as `certbot.tests` imports `mock` and is part of certbot's public api (so just switching it with `unittest.mock`might not be desirable).

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -1,5 +1,7 @@
+from distutils.version import StrictVersion
 import sys
 
+from setuptools import __version__ as setuptools_version
 from setuptools import find_packages
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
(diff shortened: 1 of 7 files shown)


FelixSchwarz:
On a second thought I think this PR is not as useful as I thought. `certbot.tests` is public API and I get test failures in dns plugins when mixing `unittest.mock` and `mock

bmw:
Thanks for trying to clean this up @FelixSchwarz. While unfortunate, it's good to know that we can't just swap out `mock` with `unittest.mock` without breaking the API.

FelixSchwarz:
But please don't take my edits as gospel - these tests were done quickly while I took a few minutes trying to build some of the updates in Fedora. I suspect you can actually use `unittest.mock` but mix'n'match may not work.

Issue #2265, opened 2016-01-22
aug.match throws out unsaved augeas changes - temp save before calling?


SwartzCr:
as seen in #2228 we are having some errors where directives are added and then lost before saving due to aug.match being called. One place where aug.match is called is in configurator.find_dir()

This is pretty unexpected behavior, and will likely cause problems down the line. It would be nice if before aug.match is called we check to see if there are unsaved changes and save them.

I've also opened an issue with python-augeas about this behavior: https://github.com/hercules-team/python-augeas/issues/15

bmw:
While saving the changes might be a good solution, we need to be careful here. If the save is temporary, it will be reverted on the next run of the client. If the save is permanent, it will create a new checkpoint causing our current rollback mechanisms not work. These are certainly solvable problems, however.

Another approach is to just check/assert that there are no unsaved changes before calling `match`.

SwartzCr:
otherwise do what? error out?

bmw:
Yeah. I think we could just `assert` that there are no unsaved changes before calling `match`. It'd force us to make sure we're saving everything correctly and not losing changes anywhere else. It'd require a lot of testing though to make sure we're not going to randomly fail for people.

joohoi:
I dug in to this, and looks like it's not actually aug.match, but the parse_file under the hood. In example case of find_dir the call stack is (if the stars align, and the issue is exposed): `find_dir -> _get_include_path -> parse_file -> aug.load` which reloads the Augeas DOM, hence discarding the changes. This got fixed in #4104 , so closing the issue.

PR #7771, opened 2020-02-14
Disable TLS session tickets in Apache

### LINKED ISSUE #7350: Disable TLS session tickets in Apache
With the resolution of #7322, we now seem to have a way to do this safely. The solution is not only to check that Apache supports disabling TLS session tickets, but also to only do so if the version of OpenSSL used by Apache is 1.0.2l or greater.

The best way to get the version of OpenSSL used by Apache (which may not be the same version of OpenSSL output by running `openssl version` or checking the version of OpenSSL used by Python) is a bit of an open question. The ideas that have been proposed to me that seem viable are:

1. Parsing openssl version from apache.  Potential problems here are needing to figure out the path to the logfile because it's configurable and making sure that the version number is in the logfile when we look. Could we potentially read the file before it was written? What if these files are configured with something like logrotate? Might the `openssl` version no longer be in the log?
2. Finding the `mod_ssl` shared object file and seeing what version of ~Apache~OpenSSL it links to. Potential issues I see here are finding the path to the module as it's configurable, reliably figuring out what it's linked to (do some systems statically link while others dynamically link?), and if it's possible, how to do this if any systems include `mod_ssl` in Apache as a built in module rather than one that is loaded externally.

For error handling, I talked about this with people and right now we're thinking that:

1. If for some reason we fail to determine the OpenSSL version, we should keep session tickets at their default of enabled. Doing anything else isn't worth breaking people's servers.
2. This could potentially be done in a later PR, but we should set up test farm tests that check that we can determine the OpenSSL version used by Apache on all systems the tests are run on.

The combination of these two things should stop us from breaking people's servers if something changes around us detecting these versions and will let us know if this detection is broken on a major system that we support.

@adferrand, you're welcome to pick this issue up if you find the spare time.

ohemorange:
Fixes #7350.

This PR changes the parsed modules from a `set` to a `dict`, with the filepath argument as the value. Accordingly, after calling `enable_mod` to enable `ssl_module`, modules now need to be re-parsed, so call `reset_modules`. (Alternatives include parsing the loaded module file in a system-dependent way and add it to the dict, or re-parsing modules within the `openssl_version()` call; this way seemed cleaner.)


ohemorange:
Anything I could do to make this easier to review? Happy to write up any descriptions of decisions, code patterns, tradeoffs, etc.

bmw:
Thanks but I don't think so right now. I've just had a bunch of other Certbot stuff come up lately so I haven't been able to dive into this yet.

I should be able to in the next day or two though and will let you know if I have any questions.

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -225,6 +238,32 @@ def updated_mod_ssl_conf_digest(self):
...
+            return None
+        if not ssl_module_location:
+            return None
+        # Step 2. Grep in the .so for openssl version
+        try:
+            with open(ssl_module_location) as f:
+                contents = f.read()
+        except IOError as error:
+            logger.debug(str(error), exc_info=True)
+            raise errors.PluginError("Unable to open ssl_module file")
```
bmw: I personally think we may just want to log a warning and leave session tickets enabled in this case too. If we hit this case on a user's system, raising an exception will break the plugin there until we patch it while logging a warning will print and log a message on every run, but keep the plugin working.

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -225,6 +238,32 @@ def updated_mod_ssl_conf_digest(self):
...
+        """Lazily retrieve openssl version"""
+        if self._openssl_version:
+            return self._openssl_version
+        # Step 1. Check for LoadModule directive
+        try:
+            ssl_module_location = self.parser.modules['ssl_module']
+        except KeyError:
+            return None
+        if not ssl_module_location:
+            return None
```
bmw: Did you consider logging a warning in this case too?

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -225,6 +238,32 @@ def updated_mod_ssl_conf_digest(self):
         """Full absolute path to digest of updated SSL configuration file."""
         return os.path.join(self.config.config_dir, constants.UPDATED_MOD_SSL_CONF_DIGEST)
 
+    def openssl_version(self):
+        """Lazily retrieve openssl version"""
+        if self._openssl_version:
+            return self._openssl_version
+        # Step 1. Check for LoadModule directive
+        try:
+            ssl_module_location = self.parser.modules['ssl_module']
```
bmw: Today I learned that this path can also be relative to the Apache server root. See https://httpd.apache.org/docs/2.4/mod/mod_so.html#loadmodule.

I noticed this because I ran our Apache test farm tests and they failed on Fedora trying to open this file because a relative path is used. (The good news is the tests on all other systems in `apache2_targets.yaml` used the correct file though!)

certbot-apache/certbot_apache/_internal/configurator.py
ohemorange: If I did, I can't remember why I decided not to; adding warnings now.

certbot-apache/certbot_apache/_internal/configurator.py
ohemorange: My original reasoning here was that if this is an error, there's likely a bigger issue at play that will cause failures so might as well error, but I'm also perfectly fine with changing this to a non-breaking error and letting whatever else might have failed elsewhere fail.

certbot-apache/tests/autohsts_test.py
```diff
@@ -42,8 +42,11 @@ def get_autohsts_value(self, vh_path):
     @mock.patch("certbot_apache._internal.configurator.ApacheConfigurator.restart")
     @mock.patch("certbot_apache._internal.configurator.ApacheConfigurator.enable_mod")
     def test_autohsts_enable_headers_mod(self, mock_enable, _restart):
-        self.config.parser.modules.discard("headers_module")
-        self.config.parser.modules.discard("mod_header.c")
+        try:
+            del self.config.parser.modules["headers_module"]
+            del self.config.parser.modules["mod_header.c"]
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
ohemorange: it shouldn't matter because we explicitly add them in the test setup, but I agree that the other way is more robust to changes so I'll switch to that.

certbot-apache/tests/configurator_test.py
```diff
@@ -620,12 +627,14 @@ def test_prepare_server_https_needed_listen(self):
...
         self.config.parser.add_dir_to_ifmodssl = mock_add_dir
         self.config.enable_mod = mock_enable
+        self.config.parser.parse_modules = mock.Mock()
 
         self.config.prepare_server_https("443")
         self.assertEqual(mock_add_dir.call_count, 1)
 
-    def test_prepare_server_https_mixed_listen(self):
-
+    @mock.patch("certbot_apache._internal.apache_util._get_runtime_cfg")
```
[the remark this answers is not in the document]
ohemorange: works for me; I'd assumed there was a reason that the previous implementation hadn't done it (coverage? specifying particular responses?) but I'm happy to make the change.

certbot-apache/tests/configurator_test.py
```diff
@@ -1752,18 +1762,81 @@ def test_manually_modified_past_file_warns(self):
...
+
+        self.config._openssl_version = None
+        self.config.parser.modules['ssl_module'] = None
+        self.assertEqual(self.config.openssl_version(), None)
+
+        self.config.parser.modules['ssl_module'] = "/fake/path"
+        self.assertRaises(errors.PluginError, self.config.openssl_version)
+
+        contents_missing_openssl = "these contents won't match the regex"
+        with mock.patch("%s.open" % six.moves.builtins.__name__,
```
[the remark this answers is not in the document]
ohemorange: I found this approach somewhere else that I sure do wish I'd kept the reference to now; I'll switch to the mock version.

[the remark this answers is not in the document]
ohemorange: I guess there could theoretically be a different between `configurator` and `configurator_test`'s mocks, but probably  not, and if it there, it's better to use `configurator`'s.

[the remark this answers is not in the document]
ohemorange: we also apparently use the builtins approach in other tests above.

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -225,6 +238,32 @@ def updated_mod_ssl_conf_digest(self):
...
+        except KeyError:
+            return None
+        if not ssl_module_location:
+            return None
+        # Step 2. Grep in the .so for openssl version
+        try:
+            with open(ssl_module_location) as f:
```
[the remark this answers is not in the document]
ohemorange: "decode to strings at the edges of your program" seems to me to imply that we should decode immediately after reading, before applying the regex. the only problem here is the "platform-dependent" bit, which we can resolve by specifying the encoding ourselves.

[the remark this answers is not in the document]
ohemorange: in which case, why not just specify the encoding in the `open` call?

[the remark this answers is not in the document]
ohemorange: although it is a binary file, we care about the text bits.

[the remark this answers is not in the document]
ohemorange: ah. I see the answer to my own question is "this could be a problem on windows".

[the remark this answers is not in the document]
ohemorange: ugh. `mock_open` doesn't support `rb`. I think calling `open('rb')` then `decode('UTF-8')` will work perfectly fine, including on windows, and won't require hand-coding an open mock. if you know of some reason to expect that not to work, lmk.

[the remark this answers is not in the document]
ohemorange: ah scratch that last bit, it's not going to be that simple anyway. btw feel free to ignore these for now, I'll assign you when I'm completely done with this batch!

certbot-apache/tests/configurator_test.py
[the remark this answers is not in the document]
ohemorange: well, this is moot now anyway.

certbot-apache/certbot_apache/_internal/configurator.py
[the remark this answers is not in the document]
ohemorange: went with the last suggestion, and factored out the open call for testing.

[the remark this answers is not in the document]
ohemorange: well, guess you cannot just pass garbage to `decode`.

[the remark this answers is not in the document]
ohemorange: also, I now know exactly how much our codebase is still using python2.

certbot-apache/tests/configurator_test.py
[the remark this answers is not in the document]
ohemorange: it's no longer moot! but our `oldest` tests don't like this, so I'm changing it back because apparently there is some reason that the other way works.

certbot-apache/certbot_apache/_internal/configurator.py
ohemorange: which test exactly was this?

ohemorange: I'm guessing this one (running now): https://travis-ci.com/github/certbot/certbot/jobs/297589213

ohemorange: [all of the test farm tests](https://travis-ci.com/github/certbot/certbot/builds/153094830) are passing now, so unless it was a different test that you ran manually, looks like it's good now.

certbot-apache/tests/configurator_test.py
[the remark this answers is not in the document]
bmw: Sorry for the bad suggestion!

I couldn't stop myself from digging into this so if you're also curious, what's going on here is newer versions of `mock` will automatically let you patch out builtin objects, while older ones require you pass `create=True` to `mock.patch` because they fail to find the builtin object to be patched since the lookups for these objects are done differently. There's a note about this buried in the [mock.patch docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).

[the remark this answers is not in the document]
ohemorange: yeah, this seems to be an idiomatic way to do it anyway.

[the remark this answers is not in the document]
ohemorange: thanks for investigating! I was mildly interested in knowing why but not enough to look it up myself.

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -225,6 +237,42 @@ def updated_mod_ssl_conf_digest(self):
...
+
+    def openssl_version(self):
+        """Lazily retrieve openssl version"""
+        if self._openssl_version:
+            return self._openssl_version
+        # Step 1. Check for LoadModule directive
+        try:
+            ssl_module_location = self.parser.modules['ssl_module']
+        except KeyError:
+            logger.warning("Could not find ssl_module; not disabling session tickets.")
```
bmw: I think you're probably right that we want to warn in this case too if we're going to error if `self.parser.modules['ssl_module']` is `None`, however, this has the effect of printing this warning on fresh Debian/Ubuntu systems when we haven't enabled the SSL module yet. I don't think we should do this.

To avoid this, I think we essentially need a way to check if we should assume of mod SSL is available yet. A couple ideas here are  just checking if `self.option("handle_modules")` is true in this except block or adding another argument to `install_ssl_options_conf` that's passed all the way here. What do you think?

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -238,6 +237,15 @@ def updated_mod_ssl_conf_digest(self):
         """Full absolute path to digest of updated SSL configuration file."""
         return os.path.join(self.config.config_dir, constants.UPDATED_MOD_SSL_CONF_DIGEST)
 
+    def _open_module_file(self, ssl_module_location):
+        """Extract the open lines of openssl_version for testing purposes"""
+        try:
+            with open(ssl_module_location, mode="rb") as f:
+                contents = f.read()
+        except IOError:
+            return None
```
bmw: nit: I think we should send the error here to `logger.debug`.

certbot-apache/tests/configurator_test.py
```diff
@@ -1819,25 +1813,35 @@ def test_current_version(self):
...
             with mock.patch("certbot_apache._internal.configurator.logger.warning") as mock_log:
                 # Check that correct logger.warning was printed
                 self.assertEqual(self.config.openssl_version(), None)
-                self.assertTrue("Could not find" in mock_log.call_args[0][0])
+                self.assertTrue("Could not find OpenSSL" in mock_log.call_args[0][0])
 
+    def test_open_module_file(self):
+        mock_open = mock.mock_open(read_data="testing 12 3")
+        with mock.patch("six.moves.builtins.open", mock_open):
+            self.assertEqual (self.config._open_module_file("/nonsense/"), "testing 12 3")
```
bmw: Today I learned that Python lets you put a space between the name of a function and its parentheses when calling it.

ohemorange:
> What do you think? Do you think adding this test now is worth it?

Yeah, good idea.

certbot-apache/tests/configurator_test.py
ohemorange: I knew this, but that's not why I put a space there, I'll fix it 😂

certbot-apache/certbot_apache/_internal/configurator.py
ohemorange: Ugh, I honestly hate adding contextual checks to utility methods. So annoying to trace what we expect to run when reading it later. Here's an alternate proposal: we don't warn on this. Because `enable_mod` [raises an error](https://github.com/certbot/certbot/blob/4abd81e2186eddc67551d61a8260440bd177d18d/certbot-apache/certbot_apache/_internal/configurator.py#L2190) if `ssl_module` isn't found. So, when he hit this line, either a) we expected it or b) something is [terribly wrong](https://www.sbnation.com/a/17776-football/). So what if we just return here and don't print anything? Basically this check will be a switch to know which of the two states we're in. I'm happy to add an explanation to this effect.

bmw: Good idea! I also like that idea better.

We can also be sure it's in the list of modules after the override for Debian too since we manually add it at https://github.com/certbot/certbot/blob/9e5bca4bbf3c4fd198382dd9d6e05b54014aca30/certbot-apache/certbot_apache/_internal/override_debian.py#L121

I think adding a comment so we remember this stuff in the future is a great idea too.

In looking into this, I think I found another problem though. When the Debian override adds the SSL module to the modules list, it doesn't add the path to the shared object file (see the above link) and as a result the second time `openssl_version` is called we hit the case where `self.parser.modules['ssl_module']` is `None` and as a result still print a warning and do not disable session tickets on a fresh Debian/Ubuntu install. I think we should probably fix this too.

bmw: This comment is a little more hastly written than normal, but I wanted to flag this now in case you kept working on this tonight.

I think I may have thought of a problem with your plan to handle Debian and `enable_mod`. We only call `enable_mod("ssl")` on systems where `self.option("handle_modules")` is true (which I think only Debian off the top of my head). See https://github.com/certbot/certbot/blob/3bfaf41d3df0417a2d58d088af1a893d320aeb55/certbot-apache/certbot_apache/_internal/configurator.py#L1238-L1243

If that's right, then I think there's a realistic option (c) to hitting this case which is you're on a non-Debian system and we're failing to parse your modules for whatever reason, but it's not an error because we essentially assume that your modules are set up correctly.

ohemorange: I knew that it was setting it to `None` and thought I accounted for that but can't remember how.

ohemorange: Oh, I think it was by making sure that `parse_modules` is called again, after the `add_mod` call, but maybe that's not actually the case. Because I had an earlier implementation where I tried to set it to something other than `None`, but then decided that calling `parse_modules` again is easier.

ohemorange: This addition here: https://github.com/certbot/certbot/pull/7771/files#diff-2caba5786389c0d032404485e9936206R1293

ohemorange: > If that's right, then I think there's a realistic option (c) to hitting this case which is you're on a non-Debian system and we're failing to parse your modules for whatever reason, but it's not an error because we essentially assume that your modules are set up correctly.

This is a good point. But I also think that we're fine not logging anything here; if `ssl_module` isn't installed, there's no way certbot works, so we'll error somewhere else to fix it, no? Like, printing that we're not disabling session tickets is nice, but probably unnecessary relative to the other issues that are going to need to be fixed in that case.

ohemorange: Oh, I see what you mean about `self.option("handle_modules")` now; the github interface was cutting that off. Looking again.

ohemorange:
Much frustrating setup later, I'm pretty sure that `apacheconftest` is currently erroneously passing on `two-blocks-one-line-1693.conf`, and the change in this PR is now making us accurately fail on that invalid conf (`</VirtualHost> </IfModule>` should be two lines, based on my testing).

ohemorange:
@bmw a thought -- I'm making the change to call `self.parser.aug.load()` before calling `self.parser.reset_modules()`. But should `self.parser.aug.load()` actually be called from *within* `self.parser.reset_modules()` semantically, not for here but just in general? Not sure about if we need it other places. Looks like we call `self.parser.aug.load()` in several places basically usually just before we know we'll need it reloaded, but maybe if it were in `reset_modules` we wouldn't have to do that so much? Not actually necessary to take the time to look into, mostly wondering if you had a general sense of it.

bmw:
Good question. I think we should leave it out of `reset_modules`.

In addition, one problem with calling `aug.load` that I initially thought was no longer a problem, but when looking into it in response to your question, I think it is that any changes to Augeas state that hasn't been saved to disk are thrown away. This is discussed at https://github.com/certbot/certbot/issues/2265 and particular instances were fixed in https://github.com/certbot/certbot/pull/4104. I think we should call `ensure_augeas_state` ourselves here before calling `aug.load` to make sure we're not throwing away any unwritten changes to the config.

PR #7772, opened 2020-02-14
Simplify and document use by services without root privileges

mcb30:
Certificates are public information by design: they are provided by
web servers without any prior authentication required.  In a public
key cryptographic system, only the private key is secret information.

The private key file is already created as accessible only to the root
user with mode 0600, and these file permissions are set before any key
content is written to the file.  There is no window within which an
attacker with access to the containing directory would be able to read
the private key content.

Set the relevant default directory permissions to 0755 rather than
0700.  This allows certificates to be read by non-root users without
requiring further manipulation of the directory permissions.

Leave the private key file permissions as 0600 by default, and add
documentation explaining how to use chgrp and chmod to make the
private key file readable by a non-root service user.

This provides guidance on the simplest way to solve the common problem
of making keys and certificates usable by services that run without
root privileges, with no requirement to create a custom (and hence
error-prone) executable hook.

Signed-off-by: Michael Brown <mbrown@fensystems.co.uk>

### DIFF
--- a/certbot/certbot/_internal/storage.py
+++ b/certbot/certbot/_internal/storage.py
@@ -986,7 +986,7 @@ class RenewableCert(interfaces.RenewableCert):
         for i in (cli_config.renewal_configs_dir, cli_config.default_archive_dir,
                   cli_config.live_dir):
             if not os.path.exists(i):
-                filesystem.makedirs(i, 0o700)
+                filesystem.makedirs(i, 0o755)
                 logger.debug("Creating directory %s.", i)
         config_file, config_filename = util.unique_lineage_name(
             cli_config.renewal_configs_dir, lineagename)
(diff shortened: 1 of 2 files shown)


bmw:
Unfortunately I don't think we can accept a PR like this yet.

Prior to Certbot 0.29.0, the permissions of the private key were 0644 and were only protected by the permissions on the directory.

While this was over a year ago, with the way Certbot is currently distributed, it is common for people to downgrade their Certbot version. We have the script [certbot-auto](https://certbot.eff.org/docs/install.html#certbot-auto) which we no longer recommend most people use that always uses the latest released version, while packages like those in Debian and Ubuntu lag behind.

If someone first creates the directory with the change in this PR and then downgrades their Certbot version to be older than 0.29.0, their private key will become world readable. To avoid this, I unfortunately think we should continue to rely on the user manually changing these directory permissions at least until this isn't a common pattern for Certbot users.

Luckily, I don't think they need an executable hook. I think they can just change the permissions of `/etc/letsencrypt/{archive,live}` as those are the two directories with restrictive permissions which Certbot does not reset upon renewal.

mcb30:
@bmw Thanks for the review, and for raising the point about downgrades.  I am surprised to hear that downgrading certbot is a common scenario, but you are correct that it would cause a problem.

I have created #7821 to address your comments.

Issue #7262, opened 2019-07-23
filesystem.realpath doesn't handle parent directories


bmw:
`certbot.compat.filesystem.realpath` only removes symlinks from the path if the symlink is the last element of the path while `os.path.realpath` removes all symlinks.

For example, if the path is `/var/folders` and `/var` is a symlink to `/private/var`, notice the differing result between the two functions:
```
In [10]: path = '/var/folders'

In [11]: os.path.realpath(path)
Out[11]: '/private/var/folders'

In [12]: filesystem.realpath(path)
Out[12]: '/var/folders'
```
I think we should fix this.

PR #7794, opened 2020-02-21
Fix tests on macOS Catalina

bmw:
This PR fixes the failures that can be seen at https://dev.azure.com/certbot/certbot/_build/results?buildId=1184&view=results.

You can see this code running on macOS Catalina at https://dev.azure.com/certbot/certbot/_build/results?buildId=1192&view=results.

### DIFF
--- a/certbot/tests/cli_test.py
+++ b/certbot/tests/cli_test.py
@@ -34,7 +34,9 @@ class TestReadFile(TempDirTestCase):
             # code using relative paths below. See
             # https://bugs.python.org/issue38295 for another example of this.
             # Eliminating any possible symlinks in self.tempdir before passing
-            # it to os.path.relpath solves the problem.
+            # it to os.path.relpath solves the problem. This is done by calling
+            # filesystem.realpath which removes any symlinks in the path on
+            # POSIX systems.
             real_path = filesystem.realpath(os.path.join(self.tempdir, 'foo'))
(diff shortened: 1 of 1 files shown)


certbot/tests/cli_test.py
```diff
@@ -30,15 +30,21 @@ def test_read_file(self):
             # However a relative path between two different drives is invalid. So we move to
             # self.tempdir to ensure that we stay on the same drive.
             os.chdir(self.tempdir)
-            rel_test_path = os.path.relpath(os.path.join(self.tempdir, 'foo'))
+            # The read-only filesystem introduced with macOS Catalina can break
+            # code using relative paths below. See
+            # https://bugs.python.org/issue38295 for another example of this.
+            # Eliminating any possible symlinks in self.tempdir before passing
```
adferrand: I would like to note here that `filesystem.realpath` does not have the exact same behavior than `os.path.realpath`: on Windows `os.path.realpath` is not called, and currently it then not capable of resolving symlinks in parent directories for this platform.

Obviously it does not apply to MacOS X, but could you be a little more precise in your comment, saying that `filesystem.realpath` invokes `os.path.realpath` on Unix-like platform, and so will clean up any possible symlink in the path?

I would like to avoid to imply that `filesystem.realpath` equals `os.path.realpath` until https://github.com/certbot/certbot/issues/7262 is resolved.

Issue #7694, opened 2020-01-17
Certbot forgets key size


bmw:
If you run a command like
```
sudo certbot certonly -d example.org --standalone --rsa-key-size 4096
```
it will result in a renewal configuration file containing the following renewal parameters:
```
rsa_key_size = 4096
authenticator = standalone
```
If you then run
```
sudo certbot certonly -d example.org --standalone --reuse-key --force-renewal
```
`rsa_key_size` is dropped from the renewal configuration file despite it still being a 4096 bit key.

I think we should preserve this value.

osirisinferi:
Not sure if this is related or perhaps should be in its own issue, but the `certbot` output is also confusing:

```
(...)
Reusing existing private key from /etc/letsencrypt/live/example.com/privkey.pem.
Generating 2048 bits RSA key
(...)
```

While I tested with `--rsa-key-size 4096` too.

bmw:
@osirisinferi I wasn't able to reproduce this. Do you have the RSA key size set explicitly to 2048 in something like `/etc/letsencrypt/cli.ini`?

If not, can you open a new issue describing how to reproduce this so we can take a look?

osirisinferi:
@bmw Sorry, my bad, I had my git branched from `master` (trying to look into this issue), but my server was still running my custom ECDSA branch.. :blush: 

I'm having a hard time figuring out where the config item is dropped though :stuck_out_tongue:

PR #7798, opened 2020-02-23
WIP: Rough fix for vanishing renewal config.

dmwilcox:
Addresses #7694 

Findings: Issue was caused by modifying config by reference and subsequently relying on reference in caller being passed to all subsequent users.  Issue did not manifest in 'renew' subcommand but was present in 'certonly' and 'run' code paths due to how object reference was passed.

Suggested permanent fix:  Collapse config object references, store it for each given domain in a canonical object attribute and always use it from there.

Still needs:
- a decision on making renewal.py:_reconstitute public _for real_ or wrapping it in a public function
- is RenewableCertificate a reasonable candidate for holding config object reference?
- tests -- I have some partially working ones but will seek advice in the EFF #certbot mattermost channel

### DIFF
--- a/certbot/certbot/_internal/configuration.py
+++ b/certbot/certbot/_internal/configuration.py
@@ -58,6 +58,15 @@ class NamespaceConfig(object):
     def __setattr__(self, name, value):
         setattr(self.namespace, name, value)
 
+    def __str__(self):
+        s = list()
+        for k, v in self.__dict__.items():
+            s.append("{}='{}'".format(k, v))
+        return "< NamespaceConfig {} >".format(" ".join(s))
+
+    def __repr__(self):
+        return self.__str__()
+
     @property
     def server_path(self):
         """File path based on ``server``."""
--- a/certbot/certbot/_internal/cert_manager.py
+++ b/certbot/certbot/_internal/cert_manager.py
@@ -13,6 +13,7 @@ from certbot import errors
 from certbot import interfaces
 from certbot import ocsp
 from certbot import util
+from certbot._internal import renewal
 from certbot._internal import storage
 from certbot.compat import os
 from certbot.display import util as display_util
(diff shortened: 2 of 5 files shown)


bmw:
To provide an update here, I haven't forgotten about this, but I have a lot on my plate at the moment and @maximillianh was going to find someone else to take a look.

m0namon:
Thanks a bunch @dmwilcox ! Sorry for the delay here. I'm taking a closer look later today, but here are some initial thoughts.

It seems that the main issue is that `_reconstitute` is only called in the `renew` code path, and not in the other code paths-- does that match what you found in your your investigation?

> a decision on making renewal.py:_reconstitute public for real or wrapping it in a public function

I think that would make sense. It also seems like this function (and its helpers) might belong better in `storage.py`.

> tests -- I have some partially working ones but will seek advice in the EFF #certbot mattermost channel

Feel free to ping us in the mattermost or here about testing! All the failures at the moment are in `cert_manager_test.py`-- you should be able to run these tests locally with `pytest` after setting up your virtual environment.

bmw:
Just to flag this to help with the review process here, it looks like there are some merge conflicts preventing CI from being run on this PR.

certbot/certbot/_internal/configuration.py
```diff
@@ -58,6 +58,15 @@ def __getattr__(self, name):
     def __setattr__(self, name, value):
         setattr(self.namespace, name, value)
 
+    def __str__(self):
+        s = list()
```
atombrella, line 62: Mainly a choice of style and speed, `s = []` is faster. Generally, literals are better for speed and probably also more pythonic.

ohemorange:
Given that https://github.com/certbot/certbot/pull/8343 was merged as an alternate fix, is there any reason to keep this open?

PR #7800, opened 2020-02-24
Run hooks with Powershell on Windows

### LINKED ISSUE #7713: Windows: Renewal deploy hook opens in notepad.exe instead of being executed
I placed a .ps1 Powershell script into `C:\Certbot\renewal-hooks\deploy` but Certbot does not execute it. Instead, it opens the script with Windows Notepad.

Windows 10 1809

https://dl.eff.org/certbot-beta-installer-win32.exe

`certbot renew --force-renewal`

```
Saving debug log to C:\Certbot\log\letsencrypt.log

Processing C:\Certbot\renewal\miketest.pwsd76.ab.ca.conf
Plugins selected: Authenticator standalone, Installer None
Running deploy-hook command: C:\Certbot\renewal-hooks\deploy\LM-CertUpload.ps1
```
Then Windows' notepad.exe opened up, showing my .ps1 script.

The [forum post about the Certbot Windows beta](https://community.letsencrypt.org/t/beta-phase-of-certbot-for-windows/105822) explicitly states that .ps1 hooks are supposed to work:
> Similarly to Linux, Certbot for Windows supports any file executable for this platform, if it is available in the PATH. It can be .bat, .ps1 or even executable binaries like .exe.

adferrand:
Fixes #7713.

As discussed in #7713, providing a Powershell script as hook for Certbot is not working currently. This is because hooks are run in a `cmd` environment, that recognizes only `.bat` files as valid scripts that can be run from their bare name on command line.

On the other hand, the Powershell both `.bat` and `.ps1` scripts as valid scripts.

This PR makes hooks command be executed by Powershell, instead of `cmd` as `Popen` does by default when `shell=true` is used. It also modifies the tests to handle this new environment, in particular in term of encoding (UTF-16-LE is the default one in Powershell).


certbot/certbot/compat/misc.py
```diff
@@ -109,3 +112,39 @@ def underscores_for_unsupported_characters_in_path(path):
...
+
+    :returns: `tuple` (`str` stderr, `str` stdout)
+    """
+    logger.info("Running %s command: %s", cmd_name, shell_cmd)
+
+    if POSIX_MODE:
+        cmd = subprocess.Popen(shell_cmd, shell=True, stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE, universal_newlines=True)
+    else:
+        line = ['powershell.exe', '-Command', shell_cmd]
```
bmw, line 135: This won't work if `shell_cmd` is a list will it?

certbot/tests/compat/misc_test.py
```diff
@@ -0,0 +1,44 @@
+"""Tests for certbot.compat.misc"""
+import mock
```
bmw, line 2: This  needs to be updated now to fallback to `unittest.mock` like in https://github.com/certbot/certbot/pull/7911.

certbot-ci/certbot_integration_tests/utils/misc.py
```diff
@@ -140,13 +140,12 @@ def generate_test_file_hooks(config_dir, hook_probe):
             entrypoint_script = '''\
 #!/usr/bin/env bash
 set -e
-"{0}" "{1}" "{2}" "{3}"
+"{0}" "{1}" "{2}" >> "{3}"
 '''.format(sys.executable, hook_path, entrypoint_script_path, hook_probe)
         else:
-            entrypoint_script_path = os.path.join(hook_dir, 'entrypoint.bat')
+            entrypoint_script_path = os.path.join(hook_dir, 'entrypoint.ps1')
```
bmw, line 146: One more question. I'm not against changing these tests to use Powershell instead of CMD, however, running a `.bat` file should still work right? I tried reverting all the changes to `certbot-ci` and the tests failed though:  https://dev.azure.com/certbot/certbot/_build/results?buildId=1576&view=logs&j=3c6fb28b-7007-5aac-a412-e7418ae4acae&t=8d779bb7-a58f-5e48-61aa-ce0ba6566b9a

Any idea why?

certbot/certbot/compat/misc.py
adferrand: Indeed, this would not work. I note that Certbot always uses always a string for `shell_cmd`, but since the package is a part of the public API, I added a branch to handle the case.

certbot/tests/compat/misc_test.py
adferrand: Done.

certbot-ci/certbot_integration_tests/utils/misc.py
adferrand: Ah yes. This is because CMD (and .bat files) is using the local charset for the Windows region (in my case, in Europe, `CP1252`) while Powershell (and .ps1 files) is using `UTF-16` in every region.

So to have consistent results when running string comparison in the tests, I decided to use only Powershell scripts, and fix the charset to use in assertions to `UTF-16` for Windows.

adferrand:
Concerning your general questions:
1) I do not expect a sane person to do anything useful with an inline command written in Batch. However even in this case, Microsoft put a great care to map all CMD  builtins functions to their Cmdlet counterpart in Powershell. So you can use `echo`, `rm`, `mv` and affiliates. Then I do not expect any significant break. If there are some edge case, I think we are covered with the current beta status of Certbot on Windows.
2) Depending what have you have in mind:
- if it is about the differences between environment variables in CMD and in Powershell, I not aware of any changes, `dir env:` (Powershell) and `set` (CMD) outputs are exactly the same.
- if it is about how environment variables are transfered in the shell invoked for `powershell -Command ...`, all environment variables of the caller context are passed to the callee.

For instance, the following python script:
```
import os
import subprocess
os.environ["HELLO"] = "WORLD"
subprocess.call(["powershell.exe", "-Command", "dir env:"])
```

Will display correctly the `HELLO` variable with value `WORLD` in the environment variables list.

certbot/certbot/compat/misc.py
adferrand: Outdated command. See below.

adferrand:
One new great example that mypy is really useful. Given that the documentation said that string and list of string were accepted for `shell_cmd` in `certbot.compat.misc:execute_command`, and given your comment in the review, I added the mypy definition to the function as `# type: (str, Union[str, List[str]]) -> None`.

Then mypy complains that calling `.split()` on a list was not possible as written in the original function here:
https://github.com/certbot/certbot/blob/a06d5ac7a11cb17610389344b6ae8cc10a75a0ef/certbot/certbot/_internal/hooks.py#L251

So since `shell_cmd` as list would have broke anyway, I decided in the migration function to only accept string value for `shell_cmd`, and updated the documentation/types appropriately.

certbot-ci/certbot_integration_tests/utils/misc.py
bmw: I don't understand. On that branch I reverted all changes to `certbot-ci` so changes to how we parse the probe file were also reverted. Does running the `.bat` file through Powershell change its output to be encoded using `UTF-16`?

adferrand: Well as always with charsets, it is a mess. In the reverted branch you tested, you end up with a python script, writing to a file, which is invoked from a batch file itself called from a Powershell command. It ends up with a file created with a declared charset `UTF-8`, but whose content written by python is in `CP1252`. And so the error when we try to compare this unreadable file with the `deploy` string.

Instead in this PR it is Powershell all along, and so `UTF-16` characters inside a declared `UTF-16` charset file, which can be successfully read by the assertion hook function.

bmw: Do you understand why this change occurs? My understanding is that in both `master` and in this PR the probe file is created in Python at https://github.com/certbot/certbot/blob/4cce3458f3e6a38fbb3f39613aca1c9e1ebd86ca/certbot-ci/certbot_integration_tests/certbot_tests/context.py#L34

Then in the tests in `master` which I used in the branch above, we call a `.bat` file that runs a Python script that appends to this file. The change to Certbot's production code to run this `.bat` file through Powershell somehow changes the declared encoding of the hook probe? I tried to look into this myself, but ended up wasting time fighting https://github.com/certbot/certbot/issues/7939 and ran out of time for the day.

It's probably fine, but since this is a user facing change that we'll probably have to live with for a long time, I'm trying to understand how running commands through Powershell may change the behavior of commands in ways that sysadmins might not expect as can be seen in the test failures above. While it's certainly OK to change our tests, I see no reason why the tests as they are in `master` shouldn't pass on the production code changes to Certbot in this PR.

adferrand: Ok let's put things that way.

You have 3 possibles ways to write content in a file in the tests, that leads to three difference encoding:
1) using `open("/path/to/file")` in Python: UTF-8
2) builtin CMD write (`output-command >> /path/to/file`): CP1252 (or other depending of the region)
3) builtin Powershell write (`& output-command >> /path/to/file`): UTF-16-LE

In parallel Pyton will assume `UTF-8` in `open` calls that end up reading a file, if not using specific flags or encoding.

On master method 1) is use in the tests when we invoke hooks by files, like it is done in `test_renew_with_hook_scripts`. And method 2) is used each time we have something like `misc.echo('wtf_pre', context.hook_probe)` in parameters for hook flags during a certbot call, since the `echo` method is defined as below:
https://github.com/certbot/certbot/blob/4cce3458f3e6a38fbb3f39613aca1c9e1ebd86ca/certbot-ci/certbot_integration_tests/utils/misc.py#L290-L302

What changes my PR is since hooks commands are not executed with Powershell, the method 3) is effectively used when `echo` function is involved, and the probe files then have `UTF-16-LE` encoding instead of `CP1252`.

At the end, all probe files are asserted with `assert_hook_execution` method, that is invoking `open` and so read all files in `UTF-8`.

Despite having `UTF-8` or `CP1252` encoding probe files on master, things were working well, because `CP1252` is single-byte encoding, and UTF-8 also for ANSI-like characters, which is only what we use here. And `CP1252`, similarly to `UTF-8` had the elegancy to use the same byte value than the ANSI encoding, so at the end everything is correctly read with just assuming everything is `UTF-8`.

However `UTF-16-LE` is a double-byte encoding. Here forget about ANSI encoding retrocompatibility. Then our tests are failing when `assert_hook_execution` tries to read the files written with method 3) assuming `UTF-8` encoding.

There were two possibilities to fix that issue:
- change the logic as I did in this PR, to always have `UTF-16-LE` encoding and assume that when we are on Windows for the tests
- change `echo` function to write probes from Python using `open`, and continue to assume everything is `UTF-8`

I prefered the first solution because I did not want to rely to much on Python to cover the plateform differences, since these integration tests are precisely here to detect errors on specific plateforms. In particular I wanted to be more "real world" with Certbot hooks, as on Windows, everything is embedded and you cannot assume that a Certbot user has in fact a Python distribution installed on his system. Using the builtin shell file writes felt a decent logic to test. Finally the first solution allows to continue to have a minimal logic written in something else than Python.

bmw: Ah I see! I had forgotten about this `echo` function and thought the`.bat` files (that run Python) from method 1 somehow changed their behavior when ran through Powershell in a way that broke our tests which I was trying to understand as a potential way things might not work as we/our users would expect here.

Your explanation makes sense though and I'm fine with the test changes being made.

bmw:
I think the only thing we need here is a changelog entry. Otherwise, this LGTM!

adferrand:
@bmw, I added a changelog

Issue #7803, opened 2020-02-25
Fix mypy on Windows


bmw:
If you run `mypy` on Windows (or use `--platform` to make `mypy` behave like its on Windows), you'll get the following errors:
```
certbot/certbot/util.py:303: error: Name 'distro' is not defined
certbot/certbot/util.py:319: error: Name 'distro' is not defined
certbot/certbot/util.py:369: error: Name 'distro' is not defined
certbot/certbot/compat/filesystem.py:76: error: Module has no attribute "chown"
```
(The last error only shows up for me after fixing the first errors.)

I think we should probably fix these issues so our `mypy` tests can be run successfully on any OS. I think this is preferred so our code is tested in more scenarios and it really works on any platform. Alternatively, we could specify `platform` in `mypy.ini` which would tell `mypy` to treat the code like its running on that platform regardless of the underlying OS.

bmw:
If anyone ends up working on this issue, you should probably take a look at https://github.com/certbot/certbot/issues/7803 where we stopped running mypy tests on Windows by default and configured mypy in a way that may need some tweaking on Windows.

Issue #7488, opened 2019-11-01
Move platform specific code in certbot.util to certbot.compat


bmw:
See https://github.com/certbot/certbot/pull/7467#issuecomment-548724348.

adferrand:
Still relevant

PR #7804, opened 2020-02-25
Fix mypy on macOS

bmw:
If you run `mypy --platform darwin certbot/certbot/util.py` you'll get:
```
certbot/certbot/util.py:303: error: Name 'distro' is not defined
certbot/certbot/util.py:319: error: Name 'distro' is not defined
certbot/certbot/util.py:369: error: Name 'distro' is not defined
```
This is because mypy's logic for handling platform specific code is pretty simple and can't figure out what we're doing with `_USE_DISTRO` here. See https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks for more info.

Setting `_USE_DISTRO` to the result of `sys.platform.startswith('linux')` solves the problem without changing the overall behavior of our code here though.

This fixes part of https://github.com/certbot/certbot/issues/7803, but there's more work to be done on Windows.

### DIFF
--- a/certbot/certbot/util.py
+++ b/certbot/certbot/util.py
@@ -25,11 +25,9 @@ from certbot._internal import lock
 from certbot.compat import filesystem
 from certbot.compat import os
 
-if sys.platform.startswith('linux'):
+_USE_DISTRO = sys.platform.startswith('linux')
+if _USE_DISTRO:
     import distro
-    _USE_DISTRO = True
-else:
-    _USE_DISTRO = False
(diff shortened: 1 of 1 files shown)


adferrand:
LGTM. As a side note, put of the scope of this PR: ultimately this platform specific logic should go in `certbot.compat` module.

bmw:
Yep! That's tracked by https://github.com/certbot/certbot/issues/7488.

Issue #7795, opened 2020-02-22
[Apache] No check if required module isn't loaded


osirisinferi:
Bug first identified by user on the Community: https://community.letsencrypt.org/t/ah00526-syntax-error-on-line-1-of-etc-apache2-other-le-http-01-challenge-pre-conf/114001

## My operating system is (include version):
Gentoo

## I installed Certbot with (certbot-auto, OS package manager, pip, etc):
Latest (git master) through Portage (OS package manager)

## I ran this command and it produced this output:
I disabled the `LoadModule` for mod_rewrite and ran:

`certbot certonly --staging --apache -d example.com`

which produced:

```
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator apache, Installer apache
Obtaining a new certificate
Generating 2048 bits RSA key
Performing the following challenges:
http-01 challenge for example.com
Error while running apache2ctl configtest.
 * Checking apache2 configuration ...AH00526: Syntax error on line 1 of /etc/apache2/vhosts.d/le_http_01_challenge_pre.conf:
Invalid command 'RewriteEngine', perhaps misspelled or defined by a module not included in the server configuration
 [ !! ]


 * apache2 has detected an error in your setup:

Cleaning up challenges
Error while running apache2ctl configtest.
 * Checking apache2 configuration ...AH00526: Syntax error on line 1 of /etc/apache2/vhosts.d/le_http_01_challenge_pre.conf:
Invalid command 'RewriteEngine', perhaps misspelled or defined by a module not included in the server configuration
 [ !! ]


 * apache2 has detected an error in your setup:
```

## Certbot's behavior differed from what I expected because:
A proper error message, warning the user that a required module wasn't loaded with advice to install the required module and modify the Apache configuration to enable it.

## Here is a Certbot log showing the issue (if available):
No additional useful information is logged.

## Here is the relevant nginx server block or Apache virtualhost for the domain I am configuring:
N/A

osirisinferi:
In `http_01.py` there is the function `prepare_http01_modules`, which checks for and tries to enable the required modules:

https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/_internal/http_01.py#L81-L92

 *However*, it does this **only** for Debian and Ubuntu because of the `self.configurator.conf("handle-modules")` check, which will be False for every distribution except Debian (and Ubuntu).

[...]

PR #7808, opened 2020-02-25
Apache module check fix

osirisinferi:
Fixes https://github.com/certbot/certbot/issues/7795

In https://github.com/certbot/certbot/pull/5202 @joohoi added distribution specific override functionality based on class inheritance to the Apache module. This includes distro-specific `enable_mod()` functions. The [current `enable_mod()` function only raises an error](https://github.com/certbot/certbot/blob/42dda35/certbot-apache/certbot_apache/_internal/configurator.py#L2295-L2314) with the specific overrides providing a functioning enable_mod, if possible.

Therefore, my argument is that checking for the `handle-modules` variable is superfluous with the functionality of the current `enable_mod` function**s**. Removing the check for `handle-modules` fixes situations where required modules aren't loaded and `handle-modules` equals False.

One might even argue the [using of `handle-modules` in configurator.py`](https://github.com/certbot/certbot/blob/42dda35/certbot-apache/certbot_apache/_internal/configurator.py#L1238) is also superfluous with the same argument.

### DIFF
--- a/certbot-apache/certbot_apache/_internal/http_01.py
+++ b/certbot-apache/certbot_apache/_internal/http_01.py
@@ -81,15 +81,14 @@ class ApacheHttp01(common.ChallengePerformer):
     def prepare_http01_modules(self):
         """Make sure that we have the needed modules available for http01"""
 
-        if self.configurator.conf("handle-modules"):
-            needed_modules = ["rewrite"]
-            if self.configurator.version < (2, 4):
-                needed_modules.append("authz_host")
-            else:
-                needed_modules.append("authz_core")
-            for mod in needed_modules:
-                if mod + "_module" not in self.configurator.parser.modules:
-                    self.configurator.enable_mod(mod, temp=True)
+        needed_modules = ["rewrite"]
+        if self.configurator.version < (2, 4):
+            needed_modules.append("authz_host")
+        else:
+            needed_modules.append("authz_core")
+        for mod in needed_modules:
+            if mod + "_module" not in self.configurator.parser.modules:
+                self.configurator.enable_mod(mod, temp=True)
 
     def _mod_config(self):
         selected_vhosts = []  # type: List[VirtualHost]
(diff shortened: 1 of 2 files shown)


bmw:
We are indeed inconsistent with whether or not we check `handle_modules` before calling `enable_mod`.

While improving the error message is great, I think it's possible we may not want to change anything here. The reason is due to the wide variety of ways different distros add modules to Apache, I'm not sure if our checks of `parser.modules` is always reliable. If we get this wrong and the module is indeed enabled but we weren't able to detect it, Certbot wil error out with this change while it would have worked before.

If this is indeed a concern, we could wrap the error message raised when we run `configtest` to list the required modules as something for the user to look into.

@joohoi, do you have the time/interest to weigh in here?

osirisinferi:
> (&hellip;) I think it's possible we may not want to change anything here.

I think that chance is rather large, but I wanted to spike a little bit of discussion :wink:

bmw:
I haven't heard from Joona about this, but I think we should continue to assume that the necessary modules are available when `handle-modules` is false.

To improve this error message though, I'd be willing to accept a PR that wraps the error potentially thrown out of `restart` [here](https://github.com/certbot/certbot/blob/1285297b23e445c10dfac481b02447d4cc2cf6f6/certbot-apache/certbot_apache/_internal/configurator.py#L2516) to suggest that the problem may be that `mod_rewrite` and `mod_authz_host`/`mod_authz_host` are not available and that they are required for Certbot to be able to perform the challenge.

To build the list of required modules, I'd recommend creating a function in `http_01.py` that returns the needed modules that can be used both by that file and when generating this error message.

What do you think @osirisinferi? Are you interested in writing that PR?

osirisinferi:
@bmw Interested sure, be able to any time soon: I'm afraid not. As a medical doctor, I'm pretty busy with the whole pandemic at the moment I'm afraid.

bmw:
Oh wow! I didn't know you were a medical doctor. That work is certainly more important and urgent than any Certbot issues right now.

In that case, since I'd like to take a different overall approach here, I'm going to close this PR for now, but if you find the time and interest in the future, by all means open a new PR.

Try and stay healthy!

Issue #7810, opened 2020-02-26
Try and set up coveralls


bmw:
When I brought up removing codecov, some people fondly remembered the nice reports generated by coveralls. I think it'd be great if we could use it again.

I briefly looked into this and as of now, it seems tricky. The first problem is Coveralls doesn't officially support Azure Pipelines. See https://github.com/lemurheavy/coveralls-public/issues/1406.

The second problem is none of the coveralls Python clients I found support setting build information described at the bottom of https://docs.coveralls.io/supported-ci-services to allow other CI services to be used. I created https://github.com/z4r/python-coveralls/issues/76 and https://github.com/coveralls-clients/coveralls-python/issues/220 to request this feature.

Finally, we've talked about maybe migrating to [GitHub Actions which does have coveralls support](https://github.com/marketplace/actions/coveralls-github-action), however, it requires your coverage to be formatted as an lcov file which `coverage` doesn't seem to support.

bmw:
I think ideally we should try to get nice coverage data from all of our tests measuring test coverage, but if we cannot, maybe having this only for the Linux builds from Travis will be an improvement for the devs not working on Windows support.

bmw:
We didn't have this with codecov for quite a while, but the lack of a way to gently notify the reviewer of a PR that the change is reducing code coverage is making me sad...

I considered tightening up our checks in `tox.cover.py`, but I think making the tests fail until you change those values or adding `# pragma: no cover` lines is too aggressive and I don't think we don't really have a way to signal "failures" in CI other than making the entire test fail.

@adferrand, if we were to hack support into a coveralls client and/or play with build artifacts to manually combine coverage in Azure for different builds, do you remember from your previous testing of this if we could get the behavior we want? https://github.com/lemurheavy/coveralls-public/issues/613 makes it sound like we can't nicely merge the coverage reports from the different OSes, but two separate coverage reports if they are reported properly is probably preferred for us anyway.

PR #7811, opened 2020-02-26
Remove codecov

bmw:
After getting a +1 from everyone on the team, this PR removes the use of `codecov` from the Certbot repo because we keep having problems with it.

Two noteworthy things about this PR are:

1. I left the text at https://github.com/certbot/certbot/blob/4ea98d830bcc3d1b980a4055243c6a6a25d8dc54/.azure-pipelines/INSTALL.md#add-a-secret-variable-to-a-pipeline-like-codecov_token because I think it's useful to document how to set up a secret variable in general.
2. I'm not sure what the text "Option -e makes sure we fail fast and don't submit to codecov." in `tox.cover.py` refers to but it seems incorrect since `-e` isn't accepted or used by the script so I just deleted the line.

As part of this, I said I'd open an issue to track setting up coveralls (which seems to be the only real alternative to codecov) which is at https://github.com/certbot/certbot/issues/7810.


bmw:
I don't think we should continue running once we've failed to meet the coverage threshold for a package. I have this opinion for implementation simplicity, consistency with our other unit tests, and the fact that I've gotten feedback that it's hard to understand what failed in our tests because the relevant information was printed a long time ago.

I can catch the exception and print a message that you failed to meet coverage threshold X for package Y though before exiting.

adferrand:
In term of unit testing, since this job I also doing that, I agree with you about the fail fast strategy. I am specifically focus on the coverage part, and put my reasoning for the consistency in behavior with coveralls on its report in GitHub: the coverage in this case is done in the entire codebase, not a part of it. This is what I think we should reproduce here.

Overall we would keep the fail fast behavior about unit testing, and global coverage about coverage check. In particular about the feedback you mentionned, my idea was to gather the coverage failures to display them at the end of the job. So I expect this fulfill the requirement to not fail the job with a message generated way before the end of the job.

bmw:
> I am specifically focus on the coverage part, and put my reasoning for the consistency in behavior with coveralls on its report in GitHub: the coverage in this case is done in the entire codebase, not a part of it. This is what I think we should reproduce here.

I disagree. This is not how `tox.cover.py` currently works and I think it is a mistake to try and make it work like codecov given the differences in the tools. `coverage report` generates a ton of output and including the output of all packages in a single report I think is a mistake.

`tox.cover.py` also currently allows you to run coverage tests on a subset of our packages which is a nice feature as I regularly hear complaints about our tests being slow and running tests on parts of the code base the dev isn't currently working on. Being able to run coverage on a subset of our packages helps with this which is something we'd lose by measuring coverage globally unless we maintained two thresholds which I think is necessary given the other downsides of this approach.

Finally, I don't have much time to spend on this. I was just trying to improve things for other Certbot devs by removing confusing codecov info, not rework `tox.cover.py`. If you want to do that, you're welcome to submit a PR, but it doesn't feel like the best use of our time or that helpful for developers for the reasons I've described.

For all of these reasons, I still prefer my simple approach of just improving the error message raised here (which you almost never see...) and merging this PR.

adferrand:
OK. I am convinced. Let's do this.

bmw:
Thanks for understanding!

With my change, failure output looks something like:
```
$ tox -e py27-cover
...
Name                                                         Stmts   Miss  Cover   Missing
------------------------------------------------------------------------------------------
certbot/certbot/__init__.py                                      1      0   100%
certbot/certbot/_internal/__init__.py                            0      0   100%
certbot/certbot/_internal/account.py                           191      4    98%   62-63, 206, 337
...
certbot/tests/storage_test.py                                  530      0   100%
certbot/tests/util_test.py                                     374     29    92%   211-213, 480-484, 489-499, 504-511, 545-547, 552-554
------------------------------------------------------------------------------------------
TOTAL                                                        14451    647    96%
Command '['/path/to/certbot/dir/.tox/py27-cover/bin/python', '-m', 'coverage', 'report', '--fail-under', '100', '--include', 'certbot/*', '--show-missing']' returned non-zero exit status 2
Test coverage on certbot did not meet threshold of 100%.
ERROR: InvocationError for command /Users/bmw/Development/certbot/certbot/.tox/py27-cover/bin/python tox.cover.py (exited with code 1)
_________________________________________________________________________________________________________________________________________________________ summary _________________________________________________________________________________________________________________________________________________________
ERROR:   py27-cover: commands failed
```
I printed the exception just so we're not throwing away information.

I think it's also possible we fail for a reason other than the threshold not meeting the percentage, but I've personally never seen this, `coverage report` output is not being captured so hopefully that would inform devs if something else is going on, and saying something like "Test coverage probably did not..." seems like overkill to me personally.

Issue #7756, opened 2020-02-07
Support async/await in the acme client library


babolivier:
It would be great if the client library could provide an `async` client, using Python's new `async`/`await` semantics. The current main solution is to use Twisted's [txacme](https://github.com/twisted/txacme) which doesn't seem to be very actively maintained and currently doesn't support ACMEv2.

bmw:
That would be cool! Realistically, I doubt this is something the Certbot team would get to anytime soon but maybe we could support it if someone else initially wrote it.

Do you know how much code could be shared between the synchronous and asynchronous clients? I haven't played around with `async`/`await` yet.

babolivier:
> Do you know how much code could be shared between the synchronous and asynchronous clients? I haven't played around with `async`/`await` yet.

I'm not exactly sure. I think some part of the `ClientNetwork` would need an async copy, but there might be some clever thing one can do with coroutines to make it async or not depending on the use case. I don't know `async`/`await` enough to give you a definitive answer on this.

bmw:
Got it.

The answer to questions like that will probably dictate whether or not this is something we'd accept a PR for. We don't really have a need for an `async` client in Certbot so if adding one will result in a lot of code that we don't use and would have to maintain, I think it probably makes more sense in a different library. If the maintenance burden is relatively small though, I could see us taking this on.

babolivier:
Yes, that makes sense.

adiroiban:
Hi. txacme was not active as it used to just work with ACME v1 :)

I am using txacme in production.
I have an old account, and I can continue to use ACME v1... so there is no rush for me to upgrade.

I dedicated time to get ACME v2 support in txacme and it now works for me with ACME v2.
It just need testing and feedback from other users, in order to make a final release of txacme with v2 support.

It would be nice to have sans i/o support in acme library... but I think that this is something for an external library... and it much easier to convert txacme library to support asyncio than convert certbot
 
txame already uses the messages (json serialization / deserialization) from certbot

The only thing that is missing is the  network interaction and then the logic of creating RSA keys, the CSR  and retry on errors.

I think that this can be closed and have an external sans i/o library for ACME v2.

bmw:
Thanks for the update @adiroiban.

I'd still potentially be open to accepting a PR to add asyncio support to our ACME library if the amount of code we had to maintain was small, but since I don't think this is something the Certbot team will ever implement themselves, I'm going to close this issue as wontfix.

PR #7825, opened 2020-02-29
Update order finalization to work in pebble strict mode.

adiroiban:
# Scope

This updates the finalize request body to work with pebble in --strict mode.

The error from pebble is

```
JWS body included JSON with a deprecated ACME v1 "resource" field ("new-cert")'
```

# Changes

I left the old CertificateRequest as it is for v1 compatibilty.

I have created a new Finalize message and use that for the v2 certbot implementation

# How to test

Start pebble with `--strict`` mode and check that certbot works.

Feel free to suggest another name for the new `Finalize` message... maybe OrderFinalize .

Thanks!!

### DIFF
--- a/acme/acme/client.py
+++ b/acme/acme/client.py
@@ -744,7 +744,7 @@ class ClientV2(ClientBase):
         """
         csr = OpenSSL.crypto.load_certificate_request(
             OpenSSL.crypto.FILETYPE_PEM, orderr.csr_pem)
-        wrapped_csr = messages.CertificateRequest(csr=jose.ComparableX509(csr))
+        wrapped_csr = messages.Finalize(csr=jose.ComparableX509(csr))
         self._post(orderr.body.finalize, wrapped_csr)
         while datetime.datetime.now() < deadline:
             time.sleep(1)
--- a/acme/acme/messages.py
+++ b/acme/acme/messages.py
@@ -525,12 +525,26 @@ class AuthorizationResource(ResourceWithURI):
 class CertificateRequest(jose.JSONObjectWithFields):
     """ACME new-cert request.
 
+    Deprecated. Use `.Finalize`.
+
     :ivar josepy.util.ComparableX509 csr:
         `OpenSSL.crypto.X509Req` wrapped in `.ComparableX509`
 
     """
     resource_type = 'new-cert'
     resource = fields.Resource(resource_type)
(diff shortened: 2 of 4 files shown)


bmw:
Thanks for the very well written PR @adiroiban.

I think this issue as well as many other ACME compliance issues will be fixed in https://github.com/certbot/certbot/pull/7176 so I'm closing this issue in favor of that one, but if I'm misunderstanding something, please let me know!

adiroiban:
As long as the pebble issue is fixed, I am OK to have it fixed in #7176. Thanks!

I would still prefer to have the FInalize class, as it is closer to the vocabulary used in ACMEv2.

When migrating from v1 to v2, it was a bit hard to see which messages were deprecated and which not..and what message should I use  in v2...as many of them still had the names from v1... but were no longer usable with v1 ACME protoocol

bmw:
> I would still prefer to have the FInalize class, as it is closer to the vocabulary used in ACMEv2.

I can understand the preference for that. On the other hand, we have this problem of `resource` fields for many objects that work with both of Let's Encrypt's endpoints so we came up with a more general solution which allows us to not introduce more objects that only work with one endpoint.

> When migrating from v1 to v2, it was a bit hard to see which messages were deprecated and which not..and what message should I use in v2...as many of them still had the names from v1... but were no longer usable with v1 ACME protoocol

If you're interested, I'd love to see an issue or PR describing/fixing some of the problems here. We could fix the documentation and/or gradually make API changes to clarify things.

Finally, this is a bit off-topic, but we got a request recently to add `async`/`await` support to our `acme` library at https://github.com/certbot/certbot/issues/7756. I had some concerns/questions about that and we honestly wouldn't have much time to devote to it on our end, but if you have any interest in adding something like that for use in `txacme`, I'd be happy to chat!

adiroiban:
Hi @bmw .

> If you're interested, I'd love to see an issue or PR describing/fixing some of the problems here. We could fix the documentation and/or gradually make API changes to clarify things.

I don't think that it is worth the effort. The 'Finalize` rename, was just a drive-by fix.

It was easy to read the certbot code and the RFC to understand which message class should be used.

------------

I am happy with the level of code usage between `certbot` and `txacme`.
There is an initial PR for txacme https://github.com/twisted/txacme/pull/158 for ACME v2 support.

My goal is to be able to use `pebble --strict` with my automated tests :) If #7176 is merge, I am happy. 

I am grateful to all the effort made by the certbot team.
I don't think there is any value in merging this PR.

PR #7829, opened 2020-03-03
Check OCSP as part of determining if the certificate is due for renewal

### LINKED ISSUE #1028: Check OCSP as part of determining if the certificate is due for renewal
There is [logic in `certbot/storage.py` to say that a certificate needs to renewed if it has been revoked](https://github.com/certbot/certbot/blob/d75908c645e0d13827b60c4c94491e0c41859ad4/certbot/storage.py#L936), but the [`storage.py` function that determines if a certificate revoked always returns `False`](https://github.com/certbot/certbot/blob/d75908c645e0d13827b60c4c94491e0c41859ad4/certbot/storage.py#L902). We have code in `certbot/ocsp.py` to obtain the OCSP status of the certificate and we should make use of it.

I think we should be careful here to only say that a certificate needs to be renewed if we can verify that the certificate is revoked. If we fail to obtain/validate the OCSP response for example, we should not say the certificate needs to be renewed. We probably also do not want to crash Certbot in this case.

bmw:
Fixes #1028.

Doing this now because of https://community.letsencrypt.org/t/revoking-certain-certificates-on-march-4/.

The new `ocsp_revoked_by_paths` function  is taken from https://github.com/certbot/certbot/pull/7649 with the optional argument removed for now because it is unused.

This function was added in this PR because `storage.py` uses `self.latest_common_version()` to determine which certificate should be looked at for determining renewal status at https://github.com/certbot/certbot/blob/9f8e4507ad0cb3dbedb726dda4c46affb1eb7ad3/certbot/certbot/_internal/storage.py#L939-L947

I think this is unnecessary and you can just look at the currently linked certificate, but I don't think we should be changing the logic that code has always had now.

### DIFF
--- a/certbot/certbot/ocsp.py
+++ b/certbot/certbot/ocsp.py
@@ -68,8 +68,19 @@ class RevocationChecker(object):
         :rtype: bool
 
         """
-        cert_path, chain_path = cert.cert_path, cert.chain_path
+        return self.ocsp_revoked_by_paths(cert.cert_path, cert.chain_path)
 
+    def ocsp_revoked_by_paths(self, cert_path, chain_path):
+        # type: (str, str) -> bool
+        """Performs the OCSP revocation check
+
+        :param str cert_path: Certificate filepath
+        :param str chain_path: Certificate chain filepath
+
+        :returns: True if revoked; False if valid or the check failed or cert is expired.
+        :rtype: bool
+
+        """
         if self.broken:
             return False
 
--- a/certbot/tests/storage_test.py
+++ b/certbot/tests/storage_test.py
@@ -699,7 +699,8 @@ class RenewableCertTests(BaseRenewableCertTest):
             self.assertFalse(self.test_rc.ocsp_revoked(version))
         self.assertEqual(mock_checker.call_args[0][0], expected_cert_path)
         self.assertEqual(mock_checker.call_args[0][1], expected_chain_path)
-        self.assertTrue(logger.called)
+        log_msg = logger.call_args[0][0]
+        self.assertIn("An error occurred determining the OCSP status", log_msg)
 
     def test_add_time_interval(self):
         from certbot._internal import storage
--- a/certbot/tests/storage_test.py
+++ b/certbot/tests/storage_test.py
@@ -672,10 +672,30 @@ class RenewableCertTests(BaseRenewableCertTest):
             errors.CertStorageError,
             self.test_rc._update_link_to, "elephant", 17)
 
-    def test_ocsp_revoked(self):
-        # XXX: This is currently hardcoded to False due to a lack of an
-        #      OCSP server to test against.
-        self.assertFalse(self.test_rc.ocsp_revoked())
+    @mock.patch("certbot.ocsp.RevocationChecker.ocsp_revoked")
+    def test_ocsp_revoked(self, mock_checker):
+        # Write out test files
+        for kind in ALL_FOUR:
+            self._write_out_kind(kind, 1)
+        version = self.test_rc.latest_common_version()
+        expected_cert_path = self.test_rc.version("cert", version)
+
+        # Test with cert revoked
+        mock_checker.return_value = True
+        self.assertTrue(self.test_rc.ocsp_revoked(version))
+        self.assertEqual(mock_checker.call_args[0][0], expected_cert_path)
+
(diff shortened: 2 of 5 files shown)


certbot/certbot/ocsp.py
```diff
@@ -68,8 +68,19 @@ def ocsp_revoked(self, cert):
...
+        """Performs the OCSP revocation check
+
+        :param str cert_path: Certificate filepath
+        :param str chain_path: Certificate chain filepath
+
+        :returns: True if revoked; False if valid or the check failed or cert is expired.
+        :rtype: bool
+
+        """
         if self.broken:
```
ohemorange: The other PR moves this after the if-cert-is-expired check; is that necessary to do here as well?

certbot/tests/storage_test.py
```diff
@@ -672,10 +672,34 @@ def test_bad_kind(self):
...
+        self.assertEqual(mock_checker.call_args[0][0], expected_cert_path)
+        self.assertEqual(mock_checker.call_args[0][1], expected_chain_path)
+
+        # Test with error
+        mock_checker.side_effect = ValueError
+        with mock.patch("certbot._internal.storage.logger.warning") as logger:
+            self.assertFalse(self.test_rc.ocsp_revoked(version))
+        self.assertEqual(mock_checker.call_args[0][0], expected_cert_path)
+        self.assertEqual(mock_checker.call_args[0][1], expected_chain_path)
+        self.assertTrue(logger.called)
```
adferrand: Could you make an assertion a little more specific here, like checking a part of the message string provided to the function? It is to avoid the assertion become silently meaningless if we add another warning in the surrounding logic here in the future.

certbot/certbot/ocsp.py
bmw: No. The two checks are independent.

I personally reordered them to keep the logic the same as what is currently in `ocsp_revoked`, but I can switch it if you prefer that for some reason.

bmw:
> The factoring out of these lines was just for testing purposes right?
>
>```
> if self.use_openssl_binary:	        if self.use_openssl_binary:
>            return self._check_ocsp_openssl_bin(cert_path, chain_path, host, url)	            return self._check_ocsp_openssl_bin(cert_path, chain_path, host, url)
>        return _check_ocsp_cryptography(cert_path, chain_path, url)	        return _check_ocsp_cryptography(cert_path, chain_path, url)
>  ```
I'm not sure I fully understand your question, but to try and answer it anyway, nothing there was changed in this PR, but the reason this code is written that way in general is indeed for testing as well as the fact we'll use the OpenSSL branch in prod if `cryptography` doesn't have OCSP support. See https://github.com/certbot/certbot/blob/9f8e4507ad0cb3dbedb726dda4c46affb1eb7ad3/certbot/certbot/ocsp.py#L41-L43

> We're just putting this check inline for renewal instead of doing prefetching? That's probably fine, I think, especially with the LE situation. We may want to double check about the server load.

Since many browsers will check OCSP, I doubt Certbot doing it twice a day will be a big issue, but that's a good thing to verify.

ohemorange:
> I'm not sure I fully understand your question, but to try and answer it anyway, nothing there was changed in this PR, but the reason this code is written that way in general is indeed for testing as well as the fact we'll use the OpenSSL branch in prod if cryptography doesn't have OCSP support. See

No, I meant that in joona's PR these lines are refactored into their own `ocsp_times` function. I assume that was just for testing purposes there and doesn't need to be done here as well?

certbot/certbot/ocsp.py
ohemorange: works for me.

bmw:
Oh I see. I think the diff shown on Joona's PR is weird.

In both his PR and mine, the functions being called at the bottom of `ocsp_revoked_by_paths` is
```
        if self.use_openssl_binary:
            return self._check_ocsp_openssl_bin(cert_path, chain_path, host, url, response_file)
        return _check_ocsp_cryptography(cert_path, chain_path, url, response_file)
```
except in this PR we don't pass a `response_file` parameter because we don't have a use for it yet.

The `ocsp_times` function is new functionality with its own functions to call based on the value of `use_openssl_binary` which you're right we don't need here because it's being added in Joona's PR for functionality in the Apache plugin which doesn't exist here yet.

PR #7832, opened 2020-03-03
Remove --no-site-packages outside of certbot-auto.

bmw:
In a recent release, `virtualenv` stopped accepting the flag `--no-site-packages`. See https://github.com/pypa/virtualenv/issues/1681.

This flag has been deprecated and been the default behavior since 2011. See https://virtualenv.pypa.io/en/legacy/changes.html#v1-7-2011-11-30.

The virtualenv package in all of Debian 8+, EPEL 6, RHEL 7+, and Ubuntu 12.04+ all have this change so I personally think checking that `virtualenv` is new enough that it has same behavior as old versions with `--no-site-packages` is unnecessary.

Since we update `virtualenv` to the latest version in the release script, you need this change in order to successfully do new releases. After this lands, I think we should cherry pick this change to the 1.3.x branch so it's ready in case we need it.

While doing so is almost certainly fine, I didn't change certbot-auto to avoid angering that soon to be deprecated monster.

### DIFF
--- a/certbot-compatibility-test/Dockerfile
+++ b/certbot-compatibility-test/Dockerfile
@@ -31,7 +31,7 @@ COPY certbot-nginx /opt/certbot/src/certbot-nginx/
 COPY certbot-compatibility-test /opt/certbot/src/certbot-compatibility-test/
 COPY tools /opt/certbot/src/tools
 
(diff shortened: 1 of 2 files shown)


bmw:
Since this includes changes to the release script, I think it'd be best if you can review this 3 line PR when you get a chance @ohemorange.

PR #7837, opened 2020-03-04
Exposes environment variable to let hooks scripts know when the last challenge is handled

### LINKED ISSUE #5484: Feature request: Add option to know what is the last domain passed to dns hook script

adferrand:
Fixes #5484 

This PRs makes Certbot expose two new environment variables in the auth and cleanup hooks of the `manual` plugin:
* `CERTBOT_REMAINING_CHALLENGES` contains the number of challenges that remain after the current one (so it equals to 0 when the script is called for the last challenge)
* `CERTBOT_ALL_DOMAINS` contains a comma-separated list of all domains concerned by a challenge for the current certificate

With these variables, an hook script can know when it is run for the last time, and then trigger appropriate finalizers for all challenges that have been executed. This will be particularly useful for certificates with a lot of domains validated with DNS-01 challenges: instead of waiting on each hook execution to check that the relevant DNS TXT entry has been inserted, these waits can be avoided thanks to the latest hook verifying all domains in one run.


varun-da:
very nice! thank you @adferrand!!!

certbot/certbot/_internal/plugins/manual.py
```diff
@@ -134,9 +133,11 @@ def _verify_ip_logging_ok(self):
             else:
                 raise errors.PluginError('Must agree to IP logging to proceed')
 
-    def _perform_achall_with_script(self, achall):
+    def _perform_achall_with_script(self, achall, achalls):
         env = dict(CERTBOT_DOMAIN=achall.domain,
-                   CERTBOT_VALIDATION=achall.validation(achall.account_key))
+                   CERTBOT_VALIDATION=achall.validation(achall.account_key),
+                   CERTBOT_ALL_DOMAINS=','.join([one_achall.domain for one_achall in achalls]),
```
schoen: Python does not really require the `[]` here because `join()` can act on a generator, not just a list. It would work equally well to say

`.join(one_achall.domain for one_achall in achalls)`

since at least Python 2.6 and probably longer ago. :slightly_smiling_face:

schoen:
Very good idea!

certbot/certbot/_internal/plugins/manual.py
adferrand: Thanks, I fixed that.

christian-weiss:
Hope to see this merged soon.

PR #7838, opened 2020-03-05
Fix issues with Azure Pipelines

### LINKED ISSUE #7814: Azure Pipelines is removing vs2015-win2012r2 on March 23rd
See https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent.

Unfortunately, I think this means we cannot easily test on this version of Windows Server. We can easily delete these tests, but I think we may also want to update our code/docs to say that we only officially support Windows Server 2016+. If there's a significant demand for Windows Server 2012 support in the future, we could adding it again but I'd rather keep things simple for ourselves until a demand arises.

What do you think @adferrand?

bmw:
This PR fixes two issues.

First, it fixes #7814 by removing our tests on Windows Server 2012. I also added the sentence "Certbot supports Windows Server 2016 and Windows Server 2019." to https://community.letsencrypt.org/t/beta-phase-of-certbot-for-windows/105822.

Second, it fixes the test failures which can be seen at https://dev.azure.com/certbot/certbot/_build/results?buildId=1309&view=results by no longer manually installing our own version of Python and instead using the one provided by Azure.

These small changes are in the same PR because I wanted to fix test failures ASAP and `UsePythonVersion` is not available on Windows 2012. See https://github.com/certbot/certbot/pull/7641#discussion_r358510854.

You can see tests passing with this change at https://dev.azure.com/certbot/certbot/_build/results?buildId=1311&view=results.

### DIFF
--- a/.azure-pipelines/templates/installer-tests.yml
+++ b/.azure-pipelines/templates/installer-tests.yml
@@ -28,8 +28,6 @@ jobs:
           imageName: windows-2019
         win2016:
           imageName: vs2017-win2016
-        win2012r2:
-          imageName: vs2015-win2012r2
     pool:
       vmImage: $(imageName)
     steps:
--- a/.azure-pipelines/templates/installer-tests.yml
+++ b/.azure-pipelines/templates/installer-tests.yml
@@ -31,10 +31,10 @@ jobs:
     pool:
       vmImage: $(imageName)
     steps:
-      - powershell: Invoke-WebRequest https://www.python.org/ftp/python/3.8.1/python-3.8.1-amd64-webinstall.exe -OutFile C:\py3-setup.exe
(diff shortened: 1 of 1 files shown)


bmw:
> ...I hoped that Azure Pipeline would have kept win2012r2 nodes as it is not EOL.

Same. I wonder if Microsoft made this decision strategically. If they make it harder to test on Windows 2012, some software will drop support for it giving Windows sysadmins an incentive to upgrade to a newer version of Windows before the EOL.

Regardless, we can always consider spending the time and money to do the testing on our own Windows 2012 instances if there is a large enough demand for it.

Thanks for the review.

PR #7861, opened 2020-03-24
Remove interactive redirect ask

### LINKED ISSUE #7594: Remove interactive redirect ask

ohemorange:
Fixes #7594.

Removes the code asking interactively if the user would like to add a redirect.


certbot/certbot/_internal/client.py
```diff
@@ -548,14 +547,6 @@ def enhance_config(self, domains, chain_path, ask_redirect=True):
         for config_name, enhancement_name, option in enhancement_info:
             config_value = getattr(self.config, config_name)
             if enhancement_name in supported:
-                if ask_redirect:
```
bmw: I'm not sure if we want to remove this conditional. After fixing things so `certbot run` sets up a redirect by default, this will have the effect of `certbot enhance` now also trying to set up a redirect by default which is something we never warned about and I'm not even immediately sure it's behavior we want.

What do you think? I personally think we should maybe change the name of the parameter here to something like `redirect_by_default` and keep it set to false (at least for now) in `certbot enhance`.

certbot/certbot/_internal/client.py
```diff
@@ -548,14 +547,6 @@ def enhance_config(self, domains, chain_path, ask_redirect=True):
...
-                    if config_name == "redirect" and config_value is None:
-                        config_value = enhancements.ask(enhancement_name)
```
bmw: This change also has the effect of no longer setting up a redirect unless the user manually requested it with `--redirect`.

To fix this, I'd recommend keeping code like this here:
```
if config_name == "redirect" and config_value is None:
    config_value = True
```

bmw:
Oh! I think we should also update the help text at https://github.com/certbot/certbot/blob/e4a0edc7af05443448082a771e5cb981fa6544ed/certbot/certbot/_internal/cli/__init__.py#L323-L324.

certbot/certbot/_internal/client.py
ohemorange: I thought the point of doing this was to specifically explicit start trying to set up a redirect by default now. Was the warning that we've had for months not sufficient? As a reminder, it said 
```
"Future versions of Certbot will automatically "	
"configure the webserver so that all requests redirect to secure "	
"HTTPS access. You can control this behavior and disable this "	
"warning with the --redirect and --no-redirect flags.")"
```

bmw: When you run `certbot enhance`, we don't hit that warning because `ask_redirect` was false. I would prefer that if we're going to make this change, we print a warning when running `certbot enhance` rather than expecting users to extend warnings they may have seen on other runs to this subcommand.

Again, I'm personally not even entirely sure that setting up a redirect for this subcommand by default is what we want. If `certbot run` and `certbot install` already set up the redirect, how is it an enhancement to the config? If both those commands and `certbot enhance` try to set up a redirect, `certbot enhance` will always print `Enhancement redirect was already set` despite the user not requesting a redirect. Would a user be surprised or confused by any of this behavior?

For these two reasons, I personally still prefer leaving `certbot enhance` behavior as is and maybe changing it as part of future work if we decide that's what we want to do.

ohemorange: The more I think about this (ie, I implemented it and was adding a test and thinking about it), the more I don't want to hide the default in `helpful`. I added a comment to `constants.py` for the default, but even then, I think a default should be a default, not a conditional default. The value of the argument should be consistent as a default. I think replacing `ask_redirect` with `redirect_by_default` is the cleanest thing to do, given our existing was-it-set-on-the-command-line detection system.

ohemorange: *I actually think `redirect_default`, to indicate "this is the value that `redirect` should default to" is more clear than `redirect_by_default`, which is a larger statement about behavior, is the more clear name.

certbot/certbot/_internal/cli/__init__.py
```diff
@@ -321,12 +321,12 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
...
         default=flag_default("redirect"),
         help="Automatically redirect all HTTP traffic to HTTPS for the newly "
-             "authenticated vhost. (default: Ask)")
+             "authenticated vhost. (default: False for enhance, True for renew and run)")
     helpful.add(
         "security", "--no-redirect", action="store_false", dest="redirect",
         default=flag_default("redirect"),
         help="Do not automatically redirect all HTTP traffic to HTTPS for the newly "
-             "authenticated vhost. (default: Ask)")
+             "authenticated vhost. (default: False for enhance, True for renew and run)")
```
ohemorange: I decided not to remove these from `security` because they show up next to each other during `all` so I wanted to find a way that makes sense even if they're next to each other. I considered adding this one under `redirect` and `run` but honestly think we have enough flags there and we didn't need it there before.

certbot/certbot/_internal/cli/__init__.py
```diff
@@ -321,12 +321,12 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         "--redirect", action="store_true", dest="redirect",
         default=flag_default("redirect"),
         help="Automatically redirect all HTTP traffic to HTTPS for the newly "
-             "authenticated vhost. (default: Ask)")
+             "authenticated vhost. (default: False for enhance, True for renew and run)")
```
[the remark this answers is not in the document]
ohemorange: ah yes I did mean install here, thanks!

certbot/certbot/_internal/cli/__init__.py
ohemorange: yes, I love that, stateful booleans are inherently more confusing than idempotent statements.

Issue #6504, opened 2018-11-14
Which DNS plugins should the Certbot team take on?


bmw:
I want to start this by saying that I do not want to discourage anyone from writing a Certbot plugin. In fact, I encourage them to do so and if there are things we can do to make that easier for you, I'd love to hear them so please open an issue/PR with your request.

With that said, the Certbot team has been thinking about which DNS plugins we want to take over maintenance of ourselves. Plugins which we do not take can remain as third party plugins, but by accepting a new plugin into our repo, we're taking on the maintenance burden of it.

Our initial plan here was to take two types of plugins:

1. Plugins for major DNS providers (e.g. Cloudflare, Amazon Route 53, etc.)
2. Plugins based on Lexicon

These has been some discussion recently about changing that plan though. Since Let's Encrypt started offering wildcard certificates, we've seen dozens of pull requests for different DNS providers and trying to review, test, package, and maintain all of them ourselves has been a bit overwhelming. For this reason, we've considered rethinking our plan here.

This is especially true when you look at how much these plugins are being used. This data is about a month old, but here is the number of times each official Certbot plugin was used to obtain a certificate from Let's Encrypt over a 2 month period counted by unique IP address:

| Authenticator plugin | Number of users |
|---|---|
| webroot | 513411 |
| apache | 392422 |
| standalone | 252474 |
| nginx | 235436 |
| manual | 56161 |
| dns-cloudflare | 5722 |
| dns-route53 | 5315 |
| dns-digitalocean | 1286 |
| dns-rfc2136 | 902 |
| dns-google | 899 |
| dns-dnsimple | 334 |
| dns-ovh | 215 |
| dns-dnsmadeeasy | 183 |
| dns-linode | 111 |
| dns-nsone | 62 |
| dns-luadns | 40 |
| dns-sakuracloud | 23 |
| dns-cloudxns | 16 |

As you can see, the DNS plugins are used way less than any of our other plugins. With this in mind, is it worth it for the Certbot team to take on these plugins?

There are things we can do to reduce the burden of these plugins such as rethinking how we test these plugins (#6464) or having more/all of these plugins use Lexicon (#6503). Alternatively, we could stop taking on these plugins ourselves and improve support for 3rd party plugins. It's also possible that the wave of DNS plugin PRs is dying down and we just have to work through the backlog.

The goal of this issue is just to be a starting point for more discussion around this topic for people who are interested.

trinopoty:
[...]

Issue #6503, opened 2018-11-14
Should we rewrite DNS plugins to use Lexicon?


bmw:
Not all of Certbot's DNS plugins use Lexicon. The initial reasoning behind this was simple. If a DNS provider has an official Python library for their DNS API, we preferred it over the third party implementation in Lexicon. We figured that the official library was likely to have better support and to work both now and in the future.

There are also features available in the official libraries that we're using that to my knowledge, aren't available in Lexicon. For instance, our Google Cloud DNS plugin is able to automatically use credentials assigned to the Google Cloud instance (see https://github.com/certbot/certbot/pull/5117). I don't think that this feature is available in Lexicon.

The differences between all of these different libraries complicates things on the Certbot end though. Each provider's libraries are structured differently while Lexicon provides a consistent interface that we can use between each provider. For that reason, some people have suggested that we rewrite these plugins to use Lexicon. The purpose of this issue is to be a starting point for more discussion of that idea.

bmw:
> We figured that the official library was likely to have better support and to work both now and in the future.

While this is probably/maybe true, I don't think we've seen many issues with the implementations in Lexicon and the issues we have seen were quickly resolved.

> There are also features available in the official libraries that we're using that to my knowledge, aren't available in Lexicon.

I think Lexicon would be open to adding support for (m)any of them. I think one of the challenges here would be finding them all to try to avoid breaking things for our users. For instance, Lexicon luckily relies `boto3` for Route53. If it didn't I think trying to track down every possible way users can do things like set credentials in that library and then mimic that behavior in Lexicon would be a bit of a challenge. While most of our other DNS libraries aren't as complicated, we will have some of the same issues there as Lexicon usually has it's own implementation for each provider rather than using the official DNS libraries.

bmw:
this issue is mainly me taking to myself about things we could do and i think the more general problem here is being tracked elsewhere. for that reason, imma close this at least for now

PR #7884, opened 2020-04-07
add support for CNAME challenge alias with route53

### LINKED ISSUE #6566: Allow updating domain pointed to by CNAME in DNS plugins

rcarz:
This PR adds support for [challenge aliases](https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode) to the dns-route53 plugin. These changes are a lot more narrow in scope than #6644 and were meant to service an immediate need while a more [generalized solution](https://github.com/certbot/certbot/pull/7244) is discussed.

## Pull Request Checklist

- [ ] If the change being made is to a [distributed component](https://certbot.eff.org/docs/contributing.html#code-components-and-layout), edit the `master` section of `certbot/CHANGELOG.md` to include a description of the change being made.

Fixes #6566

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -33,6 +33,7 @@ Authors
 * [Ben Wolfe](https://github.com/bwolfe)
 * [Bigfish](https://github.com/bwolfe)
 * [Blake Griffith](https://github.com/cowlicks)
+* [Bob Carroll](https://github.com/rcarz)
 * [Brad Warren](https://github.com/bmw)
 * [Brandon Kraft](https://github.com/kraftbj)
 * [Brandon Kreisel](https://github.com/kraftbj)
--- a/certbot-dns-route53/certbot_dns_route53/__init__.py
+++ b/certbot-dns-route53/certbot_dns_route53/__init__.py
@@ -60,6 +60,45 @@ the required permissions <https://docs.aws.amazon.com/Route53/latest
        ]
    }
 
+.. code-block:: json
+   :name: sample-aws-delegated-policy.json
+   :caption: Example AWS policy file for a delegated zone:
+
+   {
+       "Version": "2012-10-17",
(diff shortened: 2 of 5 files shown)


bmw:
Thanks for the PR @rcarz.

It looks like some of our tests are failing which would need to be fixed before this is merged, but to set expectations here, this probably won't be reviewed for a while. There is a definite need for this feature, but we're hoping to rethink our approach to DNS challenges in Certbot in the next year and any code like this complicates the transition to a new system without breaking things. Because of that, I think we'll probably at least finish our design tackling issues like #6464, #6503, and #6504 before making a decision on PRs like this one.

I'm sure that's not what you wanted to hear, but I think it's the best decision for us in the medium/long term. I'll leave this PR open since we may well accept it once we have figured out the approach to DNS challenges we want to take in the long term.

Issue #7303, opened 2019-08-06
Include server logs in certbot-ci output


bmw:
We've been seeing a lot of spurious test failures in our integration tests in Travis where we time out waiting for boulder to start. I initially hoped it was just we're not waiting long enough, but we increased the timeout to 4 minutes and it still regularly fails.

I think the best way to start fixing this is to include server logs in the Travis output at least when the server fails to start so we can get more information about what's going wrong. I think we should make sure that the client and server output is never intermixed.

This could be done by storing the server's output in memory or in a logfile and at the end of execution (probably only if tests fail?) dump that output to stdout either in the `certbot-ci` code or print the logfile in our CI config.

bmw:
With us dumping boulder logs when it fails to start, I think this is fixed well enough for now.

PR #7885, opened 2020-04-07
Print boulder logs when boulder setup fails

bmw:
This is part of https://github.com/certbot/certbot/issues/7303.

You can see tests passing with this change at https://travis-ci.com/github/certbot/certbot/builds/159128662 and you can see this code dumping failed boulder logs at https://travis-ci.com/github/certbot/certbot/builds/159129571.

@adferrand, do you have the time and interest to review this change?

### DIFF
--- a/certbot-ci/certbot_integration_tests/utils/acme_server.py
+++ b/certbot-ci/certbot_integration_tests/utils/acme_server.py
@@ -38,7 +38,7 @@ class ACMEServer(object):
         :param str acme_server: the type of acme server used (boulder-v1, boulder-v2 or pebble)
         :param list nodes: list of node names that will be setup by pytest xdist
         :param bool http_proxy: if False do not start the HTTP proxy
-        :param bool stdout: if True stream subprocesses stdout to standard stdout
+        :param bool stdout: if True stream all subprocesses stdout to standard stdout
         """
         self._construct_acme_xdist(acme_server, nodes)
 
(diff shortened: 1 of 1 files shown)


certbot-ci/certbot_integration_tests/utils/acme_server.py
```diff
@@ -188,11 +195,12 @@ def _prepare_http_proxy(self):
         self._launch_process(command)
         print('=> Finished configuring the HTTP proxy.')
 
-    def _launch_process(self, command, cwd=os.getcwd(), env=None):
+    def _launch_process(self, command, cwd=os.getcwd(), env=None, force_output=False):
         """Launch silently a subprocess OS command"""
         if not env:
             env = os.environ
-        process = subprocess.Popen(command, stdout=self._stdout, stderr=subprocess.STDOUT, cwd=cwd, env=env)
+        stdout = sys.stdout if force_output else self._stdout
```
adferrand, line 202: It is not a very good idea to dump data to stdout in the context of pytest-xdist, since it is used to coordinate the parallel workers in that context.

Since it is for debugging purpose, and stderr is left available by pytest-xdist for that matter, could we just dump to stderr here?

bmw: If I make that change, I'm imagining something like changing the name of the parameter here to `force_stderr` and this line to:
```
        stdout = sys.stderr if force_stderr else self._stdout
```
and leaving everything else the same.

That work for you?

adferrand: Totally!

bmw:
I made the suggested change. New test runs are at https://travis-ci.com/github/certbot/certbot/builds/159156933 and https://travis-ci.com/github/certbot/certbot/builds/159157048.

Issue #7886, opened 2020-04-07
Do not require mock in Python 3


bmw:
As discussed at https://bugzilla.redhat.com/show_bug.cgi?id=1795666 and https://github.com/certbot/website/pull/557#issuecomment-610182044, our unconditional dependency on the 3rd party `mock` library is causing some problems for EPEL 8 packaging.

To fix this, I think we should make our `mock` dependency conditional on using Python 2 in all of our packages. When doing this, a little care needs to be taken to ensure that:

1. Our wheels will be built properly regardless of the Python version used (see https://github.com/certbot/certbot/issues/3455).
2. The way the dependency is specified creates wheels that works with the version of `pip`, `setuptools`, etc. pinned in `certbot-auto` and the sdist works with `setuptools==1.0` which is used by some distro packaging.

For both of these, I think we can probably reuse an approach similar to the one [here](https://github.com/certbot/certbot/blob/e9895d2ec69395cfcf1cfff35e7668b57e1da00f/certbot/setup.py#L59-L74).

In our testing code, I think we should then write code like:
```
try:
    import mock
except ImportError
    from unittest import mock
```
The reason for continuing to prefer the 3rd party mock library can be seen at https://github.com/certbot/certbot/pull/7755. It appears that mixing the 3rd party and built in `unittest.mock` does not go well and since we have a [test API](https://certbot.eff.org/docs/api/certbot.tests.html), I think it's worth taking a little care to not suddenly break these tests for people if we can.

After these changes are made and we've dropped Python 2 support, I think we should:

1. Remove the `mock` requirement from all of our packages.
2. Print a deprecation warning when using the 3rd party `mock` library.
3. When we release a new major version of Certbot, unconditionally use `unittest.mock`.

These last 3 steps are future work and a new issue should probably be created to track it specifically if the other changes described here are made, but I wanted to lay out the full plan for what I was thinking of for a transition to `unittest.mock`.

ohemorange:
> The reason for continuing to prefer the 3rd party mock library can be seen at #7755. It appears that mixing the 3rd party and built in unittest.mock does not go well and since we have a test API, I think it's worth taking a little care to not suddenly break these tests for people if we can.

[...]

PR #7894, opened 2020-04-10
Do not require mock in Python 3 in acme module

ohemorange:
Part of #7886.

This PR conditionally installs mock in `acme/setup.py` based on setuptools version and python version, when possible. It then updates `acme` tests to use `unittest.mock` when `mock` isn't available.

### DIFF
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -30,11 +30,11 @@ install_requires = [
 setuptools_known_environment_markers = (StrictVersion(setuptools_version) >= StrictVersion('36.2'))
 if setuptools_known_environment_markers:
     install_requires.append('mock ; python_version < "3.3"')
-elif sys.version_info < (3,3):
-    install_requires.append('mock')
-else:
+elif 'bdist_wheel' in sys.argv[1:]:
     raise RuntimeError('Error, you are trying to build certbot wheels using an old version '
                        'of setuptools. Version 36.2+ of setuptools is required.')
+elif sys.version_info < (3,3):
+    install_requires.append('mock')
 
 dev_extras = [
     'pytest',
--- a/acme/setup.py
+++ b/acme/setup.py
@@ -1,5 +1,7 @@
+from distutils.version import StrictVersion
 import sys
 
+from setuptools import __version__ as setuptools_version
 from setuptools import find_packages
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
@@ -15,7 +17,6 @@ install_requires = [
     # 1.1.0+ is required to avoid the warnings described at
     # https://github.com/certbot/josepy/issues/13.
     'josepy>=1.1.0',
-    'mock',
     # Connection.set_tlsext_host_name (>=0.13)
     'PyOpenSSL>=0.13.1',
     'pyrfc3339',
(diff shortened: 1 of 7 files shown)


acme/setup.py
```diff
@@ -26,6 +27,12 @@
     'six>=1.9.0',  # needed for python_2_unicode_compatible
 ]
 
+setuptools_known_environment_markers = (StrictVersion(setuptools_version) >= StrictVersion('36.2'))
+if setuptools_known_environment_markers:
+    install_requires.append('mock ; python_version < "3.3"')
+elif sys.version_info < (3,3):
```
bmw, line 33: nit: I think we should probably error out if we're building the wheel and don't have `setuptools>=36.2` like we do [here](https://github.com/certbot/certbot/blob/e9895d2ec69395cfcf1cfff35e7668b57e1da00f/certbot/setup.py#L59-L74).

The reason I think this is if we build the wheel on Python 3.3+ using an old version of `setuptools`, the wheel we push to PyPI will never specify a dependency on `mock`. If people install this wheel on Python 2.7 and don't already happen to have `mock`, our tests will break.

I think this scenario is unlikely to come up for a number of reasons, but I personally think it's worth the two lines to make sure our dependencies are always specified properly.

ohemorange: Great, I love it. I originally considered that and decided to do it this way because I understood the requirements as having it be more important to support a wider range of machines for people building the wheels (who are more likely to be the ones running the tests), but I absolutely am in favor of supporting a more narrow range of build environment setups.

ohemorange: and to be clear, for systems with `sys.version_info < (3,3)` and old setuptools, would you rather just add `mock` or error out as well?

bmw: I personally think we should add `mock` and not error out in the case of `sys.version_info < (3,3)` and old setuptools when we're not building a wheel.

My understanding is erroring out when building the wheel doesn't affect our packagers because they don't use our wheels or build their own, but erroring out in other cases would break things like our EPEL 7 packages which still use Python 2 and an old version of setuptools. I personally think it's worth the two lines to keep things working there for now.

acme/setup.py
```diff
@@ -26,6 +27,15 @@
...
+    install_requires.append('mock')
+else:
+    raise RuntimeError('Error, you are trying to build certbot wheels using an old version '
```
bmw, line 36: Any reason not to use the same logic as [here](https://github.com/certbot/certbot/blob/e9895d2ec69395cfcf1cfff35e7668b57e1da00f/certbot/setup.py#L59-L74)? With the code in this PR, I think if we try to build the wheel with Python < 3.3 (which [we currently do](https://github.com/certbot/certbot/blob/8e4dc0a48c6108311cf68ca1c683d29b15d49c97/tools/_release.sh#L62)) and an old version of `setuptools`, the resulting wheel won't specify the dependency correctly.

ohemorange: wait, isn't this is the opposite of "I personally think we should add mock and not error out in the case of sys.version_info < (3,3) and old setuptools when we're not building a wheel."?

ohemorange: Maybe you meant specifically the `bdist_wheel` thing.

PR #7896, opened 2020-04-10
Do not require mock in Python 3 in apache module

ohemorange:
Part of #7886.

This PR conditionally installs mock in `apache/setup.py` based on setuptools version and python version, when possible. It then updates `apache` tests to use `unittest.mock` when `mock` isn't available.

### DIFF
--- a/certbot-apache/setup.py
+++ b/certbot-apache/setup.py
@@ -22,11 +22,11 @@ install_requires = [
 setuptools_known_environment_markers = (StrictVersion(setuptools_version) >= StrictVersion('36.2'))
 if setuptools_known_environment_markers:
     install_requires.append('mock ; python_version < "3.3"')
-elif sys.version_info < (3,3):
-    install_requires.append('mock')
-else:
+elif 'bdist_wheel' in sys.argv[1:]:
     raise RuntimeError('Error, you are trying to build certbot wheels using an old version '
                        'of setuptools. Version 36.2+ of setuptools is required.')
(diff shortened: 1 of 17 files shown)


certbot-apache/setup.py
```diff
@@ -10,14 +12,19 @@
 # acme/certbot version.
 install_requires = [
     'acme>=0.29.0',
-    'certbot>=1.1.0',
-    'mock',
+    'certbot>=1.4.0.dev0',
```
bmw, line 15: Is the only reason for this change `pip`'s issues with dependency resolution? I think it's totally fine to make this change, but I was initially thinking we wouldn't have to do this.

certbot-apache/setup.py
```diff
@@ -10,14 +12,19 @@
...
     'python-augeas',
     'setuptools',
     'zope.component',
     'zope.interface',
 ]
 
+setuptools_known_environment_markers = (StrictVersion(setuptools_version) >= StrictVersion('36.2'))
+if setuptools_known_environment_markers:
+    install_requires.append('mock ; python_version < "3.3"')
+elif sys.version_info < (3,3):
```
bmw, line 25: Same comment as https://github.com/certbot/certbot/pull/7894#discussion_r407658937.

certbot-apache/setup.py
ohemorange: I forgot to revert this change, I was just testing things. Nice catch!

Issue #7550, opened 2019-11-12
Run our coverage and lint tests on Python 3


bmw:
Some progress on this may have been made in https://github.com/certbot/certbot/pull/5727.

bmw:
Fixed in #7905.

PR #7905, opened 2020-04-14
Consolidate cover envs and default to py3-cover

bmw:
Since Python 2 has reached its end-of-life, it's going to become harder and harder to install. For this reason, I think we should stop trying to run Python 2 tests by default locally and instead rely on Travis.

This PR changes the default list of `tox` environments to use Python 3 for coverage. Since our coverage tests are almost identical to the `py3` tests we were currently running by default, I also removed them. The tests are still run in Travis as before.

I also consolidated the separate cover environments by making them conditionals in the default tox environment. The benefit of this is in addition to supporting the `py27-cover` and `py37-cover` environments whose behavior is unchanged, we also now support `py3-cover` or even just `cover`.


bmw:
Running the full test suite failed due to an unrelated issue in the test farm tests which is fixed in https://github.com/certbot/certbot/pull/7906.

bmw:
While I think we should do that too (see https://github.com/certbot/certbot/issues/7550), I think that kind of unnecessarily expands the scope of this PR to block on that.

Rather than debating this point though, I went ahead and made the change and fixed up tests so the pass.

This PR fixes #7550 now.

PR #7907, opened 2020-04-14
Add all testing paths to lint and mypy tox sources, and fix subsequent errors

ohemorange:
We weren't testing on tests outside of the module before. Let's make sure to do so. We don't expect that tests can run without also adding the `acme` module, based on our previous implementation choices, so no need to quash those errors.


bmw:
It looks like there's another filename collision that `mypy` is not happy with. See https://travis-ci.com/github/certbot/certbot/jobs/319413365.

This otherwise LGTM though.

bmw:
Well this slow drip of mypy errors is kind of annoying: https://travis-ci.com/github/certbot/certbot/jobs/319416463#L531

PR #7914, opened 2020-04-16
Update dev docs to reflect Python 2 EOL

bmw:
Python 2 is going to get harder and harder to install locally so I don't think we should assume/require devs to have it installed.

This PR builds on https://github.com/certbot/certbot/pull/7905 so our developer guide only has people use Python 3.


bmw:
Thanks for the review!

We don't, but we haven't actually started deprecating Python 2 support yet. We can't until we've started migrating people off of certbot-auto.

My goal here was to just stop requiring/expecting devs to have Python 2 locally because it's getting harder and harder to obtain and we can easily run tests using Python 2 in Travis.

Issue #7851, opened 2020-03-17
Add Ubuntu 19.10 to our test farm tests


bmw:
In the past we've talked about trying to keep the list of systems we test on relatively current I still think there's some value in that.

If the tests don't immediately work, especially due to failures in certbot-auto, I think it probably makes sense to close/kick this issue until we've done some combination of deprecate certbot-auto and rewrite the test farm tests.

Issue #7853, opened 2020-03-17
Add Ubuntu 20.04 to our test farm tests


bmw:
In the past we've talked about trying to keep the list of systems we test on relatively current I still think there's some value in that.

If the tests don't immediately work, especially due to failures in certbot-auto, I think it probably makes sense to close/kick this issue until we've done some combination of deprecate certbot-auto and rewrite the test farm tests.

oracle3987:
Help to work 2gthr not against me

oracle3987:
> In the past we've talked about trying to keep the list of systems we test on relatively current I still think there's some value in that.
> 
> If the tests don't immediately work, especially due to failures in certbot-auto, I think it probably makes sense to close/kick this issue until we've done some combination of deprecate certbot-auto and rewrite the test farm tests.

Help fix

bmw:
Closing in favor of https://github.com/certbot/certbot/issues/7915.

PR #7916, opened 2020-04-16
Test on Ubuntu 19.10

### LINKED ISSUE #6720: Once Ubuntu 19.04 comes out, update test farm tests

bmw:
Fixes https://github.com/certbot/certbot/issues/7851.

You can see tests successfully running with this change at https://travis-ci.com/github/certbot/certbot/builds/160641256.

The AMI-ID was taken from https://cloud-images.ubuntu.com/locator/ec2/. As of writing this, it is the only Ubuntu 19.10 AMI on that page in `us-east-1` using EBS for storage that isn't running on ARM. The instance store AMI cannot be used because that storage type is not available for the instance size we use. See https://aws.amazon.com/ec2/pricing/on-demand/ and https://travis-ci.com/certbot/certbot/builds/113021661 for proof of that.

### DIFF
--- a/tests/letstest/apache2_targets.yaml
+++ b/tests/letstest/apache2_targets.yaml
@@ -1,6 +1,11 @@
 targets:
   #-----------------------------------------------------------------------------
   #Ubuntu
+  - ami: ami-0545f7036167eb3aa
+    name: ubuntu19.10
+    type: ubuntu
+    virt: hvm
+    user: ubuntu
   - ami: ami-095192256fe1477ad
     name: ubuntu18.04LTS
     type: ubuntu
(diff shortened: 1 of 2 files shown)


schoen:
Sorry, I see 20.04 support is a separate issue at (#7853) so it will make sense to add that as a separate PR too.

Issue #7902, opened 2020-04-13
Switch release script to Python 3


bmw:
Similar to https://github.com/certbot/certbot/issues/7857, we should update `tools/_release.sh` to use Python 3 instead of Python 2.

This should be pretty easy to fix. I think the only changes necessary are:

1. Update `virtualenv` commands to instead use `python3 -m venv`.
2. Change [this line](https://github.com/certbot/certbot/blob/8e4dc0a48c6108311cf68ca1c683d29b15d49c97/tools/_release.sh#L160)  to `python -m http.server $PORT &`
3. Update [this line](https://github.com/certbot/certbot/blob/8e4dc0a48c6108311cf68ca1c683d29b15d49c97/tools/_release.sh#L205) where I think we just have to add some parentheses for `print`.

Finally, we have to test it to make sure it works.

PR #7918, opened 2020-04-16
Use Python 3 in the release script

bmw:
Fixes https://github.com/certbot/certbot/issues/7902.

@ohemorange, if you have the time to review this PR too, that'd be great!

I did a mock release with the offline signature code commented out and everything worked fine.

### DIFF
--- a/tools/_release.sh
+++ b/tools/_release.sh
@@ -59,7 +59,7 @@ mv "dist.$version" "dist.$version.$(date +%s).bak" || true
 git tag --delete "$tag" || true
 
 tmpvenv=$(mktemp -d)
-VIRTUALENV_NO_DOWNLOAD=1 virtualenv -p python2 $tmpvenv
+python3 -m venv "$tmpvenv"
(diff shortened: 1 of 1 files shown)


bmw:
By the way, I left the use of `virtualenv` inside of the first virtual environment we create to help get an updated version of pip, setuptools, and wheel in the second virtual environment we make and to not make unnecessary changes in this PR. `virtualenv` will use Python 3 because [it defaults to whatever version of Python it's currently running on](https://virtualenv.pypa.io/en/latest/user_guide.html#introduction) and we installed it in a virtual environment explicitly created using Python 3.

bmw:
> Make sure to update the release instructions with python3-dev and python3-venv for debian/ubuntu, but otherwise looks good!

Good idea! I did this.

To avoid duplicating everything all over the place, I just added a link to https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client in our release instructions.

PR #7926, opened 2020-04-21
Deprecate certbot-auto on Gentoo, macOS, and FreeBSD

### LINKED ISSUE #7747: Deprecate certbot-auto on Gentoo, macOS, and FreeBSD
For the sake of killing certbot-auto and doing it gradually, I would like to start deprecating the script on more systems. For systems where snaps are supported, my current inclination is to wait until we think our snaps are production ready and then direct them to use the snaps.

For systems where snaps cannot be used and we have other installation instructions available, I think it may be worth deprecating the script on these systems now. This would currently be Gentoo, macOS, and FreeBSD systems. Our current instructions at https://certbot.eff.org tell people on these systems to use their OS package manager and all of their Certbot packages are up-to-date. (Some haven't packaged our 1.2 release from yesterday yet but they all have Certbot 1.1+.)

I have two small hesitations with this. The first is that I'm not aware of a technical reason that snaps do not work on Gentoo, it's just that there is no snap package available or any other [instructions on how to install it](https://snapcraft.io/docs/installing-snapd). We could maybe wait and try to nudge Gentoo to package `snapd`, but as of 3 months ago, [we have less than 5000 Gentoo users](https://gist.github.com/bmw/4559c2cd50aa2c1fda5ca996a434ea2c#file-out-L25) so I'm unfortunately not sure it's the best use of our time.

My second hesitation is in the long run, to fully resolve ourselves from distro politics, I'm tempted to provide general instructions for smaller systems like this and describe how to install OS dependencies, create a virtual environment, and use `pip` to install Certbot, but realistically I think the current OS packages will work better for these users.

Despite those two concerns, I personally still think we should do this but I'm curious to hear what other people think.

ohemorange:
Fixes #7747.

There are other references to macOS in the code but I see no reason to rip those out and take the risk. I'm also happy to put back most of the rest of the deleted code if we think there's a chance that'll be a problem.

### DIFF
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -20,6 +20,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 * Stop asking interactively if the user would like to add a redirect.
 * `mock` dependency is now conditional on Python 2 in all of our packages.
+* Deprecate certbot-auto on Gentoo, macOS, and FreeBSD.
 
 ### Fixed
 
--- a/letsencrypt-auto-source/letsencrypt-auto
+++ b/letsencrypt-auto-source/letsencrypt-auto
@@ -656,38 +656,6 @@ BootstrapArchCommon() {
   fi
 }
 
-# If new packages are installed by BootstrapGentooCommon below, this version
-# number must be increased.
-BOOTSTRAP_GENTOO_COMMON_VERSION=1
-
-BootstrapGentooCommon() {
-  PACKAGES="
-    dev-lang/python:2.7
-    dev-python/virtualenv
-    app-admin/augeas
-    dev-libs/openssl
-    dev-libs/libffi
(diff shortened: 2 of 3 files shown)


bmw:
It looks like all changes in this PR outside of the changelog were lost when you rebuilt letsencrypt-auto at https://github.com/certbot/certbot/pull/7926/commits/26ed71cc1cc90799986e2acb2c6dee52604ecd43.

You could update the different files and templates to make the changes you did in your initial commits, but I was personally imagining us making the minimal changes here by just setting `DEPRECATED_OS=1` when the bootstrap code is chosen on these systems and leaving everything else untouched. That leaves behind unused, dead code, but I'm personally kind of in favor of only making minimal changes to this script.

ohemorange:
Oops, I previously edited the wrong file. Should be good now.

PR #7925, opened 2020-04-21
Remove references to the apache-parser-v2 branch

ohemorange:
Fixes #7786.

PR #7927, opened 2020-04-21
Fix Travis extended/full test suite logic

bmw:
In https://github.com/certbot/certbot/pull/7925 we accidently changed the logic here. Before it was:
```
type = cron OR (type = push AND branch NOT IN (apache-parser-v2, master))
```
now it's
```
type = cron OR (type = push AND branch = master)
```
We want to be able to run our full test suite on things like `test-*` branches. The reason we had been excluding `master` is it has the full test suite run on it (through what Travis calls `cron`) nightly and not running on every commit helps prevent us on waiting on CI since our nightly tests spin up so many jobs.

This PR changes things back to the intended behavior.

(We could talk about changing the condition to just `type = cron OR type = push` if we want, but I'd rather do that in a separate PR once things like `test-` branches are fixed.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ not-on-master: &not-on-master
 # Jobs for the extended test suite are executed for cron jobs and pushes to
 # non-development branches.
 extended-test-suite: &extended-test-suite
-  if: type = cron OR (type = push AND branch = master)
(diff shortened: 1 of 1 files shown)


bmw:
No worries! Sorry I missed it myself in my review.

PR #7952, opened 2020-04-25
Allow existing but empty archive and live dir to be used when creating new lineage

ntkme:
## Pull Request Checklist

---

The motivation behind this change is to allow a docker container running a tls service to bind mount the archive dir & live dir of a single lineage. For example:

``` yaml
    volumes:
      - "/etc/letsencrypt/archive/www.example.com:/etc/letsencrypt/archive/www.example.com:ro"
      - "/etc/letsencrypt/live/www.example.com:/etc/letsencrypt/live/www.example.com:ro"
```

This way, this tls service container would not have access to other lineages for other containers. For this to work, archive directory and live directory for a lineage have to be created first to be bind mounted. However, certbot currently would not work with existing but empty archive or live directory. Thus result in certificates successfully being issued, but failed to be written to disk. I learned this the hard way with setting docker auto restart on my certbot container, and it resulted hitting limit rate in just a few seconds by keeping failing and restarting.

This PR would allow writing into existing archive or live dir, as long as they are both empty.

---

Alternative ideas for docker use case:

1. Have each tls service container bind mount the certbot deploy hook dir and install a deploy hook. This means that certbot container would also need to mount the deploy location, which is definitely not ideal. Also, this still have a race condition that, if deploy hook is installed by tls service container after certificate was generated, the tls service container would have no certificate because the deploy hook was installed to late.
2. Pre-install a single deploy hook in certbot container which copy the lineage files to a different shared directory. E.g. Deploy hook would resolve symlinks and then copy files from `/etc/letsencrypt/live/www.example.com` to `/var/ssl/www.example.com`. Certbot container would need to mount `/var/ssl` and the tls service container would mount `/var/ssl/www.example.com`. This works but I don't like the complexity.
3. Simply mount the whole `/etc/letsencrypt` directory on the tls service container. This works as well, but expose unnecessary access of certbot credential as well as private key of other certificates to tls service container, thus not ideal.

### DIFF
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -21,6 +21,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 * Stop asking interactively if the user would like to add a redirect.
 * `mock` dependency is now conditional on Python 2 in all of our packages.
 * Deprecate certbot-auto on Gentoo, macOS, and FreeBSD.
+* Allow existing but empty archive and live dir to be used when creating new lineage.
 
 ### Fixed
 
--- a/certbot/certbot/_internal/storage.py
+++ b/certbot/certbot/_internal/storage.py
@@ -1007,18 +1007,18 @@ class RenewableCert(interfaces.RenewableCert):
         lineagename = lineagename_for_filename(config_filename)
         archive = full_archive_path(None, cli_config, lineagename)
         live_dir = _full_live_path(cli_config, lineagename)
-        if os.path.exists(archive):
+        if os.path.exists(archive) and (not os.path.isdir(archive) or os.listdir(archive)):
             config_file.close()
             raise errors.CertStorageError(
                 "archive directory exists for " + lineagename)
-        if os.path.exists(live_dir):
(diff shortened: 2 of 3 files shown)


ntkme:
@schoen Yes, certbot would crash with an exception like `OSError: [Errno 13] Permission denied: '/etc/letsencrypt/live/www.example.com`, which is very self-explanatory. This kind of crash can happen in many places among the whole certbot code base, but should only happen when user is explicitly playing with the directory structure and did something wrong, as certbot itself should created its own files with right permission. In the case of a user error, it is always more clear to crash and let user know that it's user's fault.

A possible alternative (bad) behavior is to change `util.unique_lineage_name`, and have it check all the permission. If permission is not sufficient, just generate a different name with `-000X` postfix. Imagine that user (or docker) created `/etc/letsencrypt/live/www.example.com` with incorrect permission, and certbot tries to be smart and created `/etc/letsencrypt/live/www.example.com-000X`. That would be confusing as the program work out but it did not do what user is expecting.

ntkme:
@schoen Is there anything specific you want to change?

Issue #7933, opened 2020-04-23
dns-google plugin fails to use instance authentication


mloughran:
On GCP, Compute Engine instances have access to a default service account (via a metadata service) which tools are expected to use to authenticate. The [certbot-dns-google documentation](https://certbot-dns-google.readthedocs.io/en/stable/) claims that this is supported:

> If you’re running on Google Compute Engine, you can assign the service account to the instance which is running certbot. A credentials file is not required in this case, as they are automatically obtained by certbot through the metadata service.

However, in my testing this is not the case. Looking through community discussions (see below), I am not alone, and the only offered solution I've see is to use the `--dns-google-credentials` flag. I get the following error on attempting to renew:

```
$ sudo certbot certonly --dns-google -m <snip> --agree-tos -n --dry-run -d '<snip>'
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator dns-google, Installer None
Obtaining a new certificate
Performing the following challenges:
dns-01 challenge for <snip>
URL being requested: GET https://www.googleapis.com/discovery/v1/apis/dns/v1/rest
URL being requested: GET https://dns.googleapis.com/dns/v1/projects/<snip>/managedZones?dnsName=<snip>.&alt=json
Cleaning up challenges
URL being requested: GET https://www.googleapis.com/discovery/v1/apis/dns/v1/rest
URL being requested: GET https://dns.googleapis.com/dns/v1/projects/<snip>/managedZones?dnsName=<snip>.&alt=json
Error finding zone. Skipping cleanup.
Encountered error finding managed zone: <HttpError 401 when requesting https://dns.googleapis.com/dns/v1/projects/<snip>/managedZones?dnsName=<snip>.&alt=json returned "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.">
```

It appears to fail on the call to mangedZones, which can be easily attempted using the `gcloud` tool (which uses the default service account):

```
gcloud dns managed-zones list
```

If I create a JSON key for the default Compute Engine service account (using the same service account to rule out IAM issues) and provide this via `--dns-google-credentials key.json`, everything works.

## Related community discussions

* https://community.letsencrypt.org/t/dns-google-dns-01-gce-authentication-problems/85929
* https://community.letsencrypt.org/t/how-to-use-dns-01-auth-on-gce-by-certbot-dns-google/85380/8
[...]

PR #7959, opened 2020-04-30
fix for issue #7933 (dns-google plugin fails to use instance authentication)

venim:
Fix updates the google-api-python-client to a current version, which uses an updated backend for creating an api service with googleapiclient.discovery.

Examples from Google Documentation:
https://cloud.google.com/docs/authentication/production#obtaining_and_providing_service_account_credentials_manually
https://cloud.google.com/docs/authentication/production#obtaining_credentials_on_compute_engine_kubernetes_engine_app_engine_flexible_environment_and_cloud_functions

### DIFF
--- a/certbot-dns-google/certbot_dns_google/_internal/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/_internal/dns_google.py
@@ -90,7 +90,13 @@ class _GoogleClient(object):
             with open(account_json) as account:
                 self.project_id = json.load(account)['project_id']
         else:
-            credentials = AppAssertionCredentials()
+            try:
+                # try it infer credentials from Google Compute Engine.
+                # support for other GCP environments is restricted due to the
+                # limitations of the deprecated oauth2client library
+                credentials = AppAssertionCredentials()
+            except:
+                credentials = None
             self.project_id = self.get_project_id()
 
         if not dns_api:
--- a/certbot-dns-google/certbot_dns_google/_internal/dns_google.py
+++ b/certbot-dns-google/certbot_dns_google/_internal/dns_google.py
@@ -6,6 +6,7 @@ from googleapiclient import discovery
 from googleapiclient import errors as googleapiclient_errors
 import httplib2
 from oauth2client.service_account import ServiceAccountCredentials
+from oauth2client.contrib.gce import AppAssertionCredentials
 import zope.interface
 
 from certbot import errors
@@ -89,7 +90,7 @@ class _GoogleClient(object):
             with open(account_json) as account:
                 self.project_id = json.load(account)['project_id']
         else:
-            credentials = None
+            credentials = AppAssertionCredentials()
             self.project_id = self.get_project_id()
 
         if not dns_api:


bmw:
Thanks for the PR @venim.

For us to merge this, test failures would need to be fixed.

Additionally, we're currently going out of our way to keep compatibility with `google-api-python-client` 1.5.5. This is because it's the [version available in Debian Buster](https://packages.debian.org/buster/python3-googleapi) and requiring a newer version would complicate/prevent us from being able to update the Certbot packages in Debian Buster. We currently have version 1.5.5 pinned [here](https://github.com/certbot/certbot/blob/1285297b23e445c10dfac481b02447d4cc2cf6f6/tools/oldest_constraints.txt#L38-L39) which is used in some of our tests to help us ensure we keep compatibility with that version.

Is it possible to fix the issue without bumping our minimum required `google-api-python-client` version? If so, how hard would it be?

venim:
I dug into the release history of the `google-python-api-client` pacakage and unfortunately pinning the version to 1.5.5 will not work. `google-auth` support was added in [release 1.6.0](https://github.com/googleapis/google-api-python-client/releases/tag/v1.6.0), the next release after 1.5.5.

If the `google-python-api-client` version can be rolled forward by one release, `google-auth==0.1.0` can be imported to achieve the bugfix.

The code would be the same essentially, except it would only modify `credentials = None` to `credentials = compute_engine.Credentials()` on line 92 (with updated imports).

venim:
Turns out there was an [older module](https://github.com/googleapis/oauth2client/blob/master/oauth2client/contrib/gce.py) within the `oauth2client` package that creates the right class of credentials. 

Just updated my PR with the 2 line change to support the default GCE credentials.

bmw:
Nice! Thanks for looking into that @venim.

I don't have easy access to a Google Cloud setup to test this right now, but I'm working on getting access to one again and when I do I'll test this PR.

I have one question now though. Based on https://cloud.google.com/docs/authentication/production#finding_credentials_automatically and changelog entries like this from https://github.com/googleapis/google-api-python-client/blob/bd581394bb128c1724bc0bd0ca01f8a3da58fdc4/CHANGELOG.md#v160:

> If neither http or credentials is specified to discovery.build and discovery.build_from_document, then Application Default Credentials will be used. 

Do you understand why a change is needed at all in our code here? If credentials aren't provided, my understanding is we should currently automatically fall back to Google's "Application Default Credentials". If this is no longer working, do you know why?

Do people hitting this problem have their default credentials configured in a way that doesn't work with Certbot (and if so, do you know if this the default behavior of GCE or something that must be done manually)? Alternatively, since that changelog section is from `google-python-api-client` 1.6.0, maybe this problem only occurs when using `google-python-api-client` 1.5.5 as is packaged in Debian.

venim:
I do think the core of the issue is the `google-api-python-client version` being stuck at `1.5.5`. The `1.6.0` changelog you linked did seem to add support for `credentials=None` .

I just tried it on a GCE environment and building a discovery resource from the `1.8.2` api without credentials specified uses the correct account.

I modified the PR to include a try/except block to catch scenarios where the user isn't on a supported GCE environment. I'm not entirely sure what happens when running this code on an AppEngine or GKE instance.

bmw:
If the problem is just the version of `google-api-python-client` in Debian Buster (and Ubuntu 18.04) is too old, I think we should just do this:

1. Update `setup.py` to require `google-api-python-client>=1.6` with a comment that it can be downgraded to `1.5.5`, but the plugin will no longer support obtaining credentials from metadata as described at https://cloud.google.com/compute/docs/storing-retrieving-metadata.
2. Update [oldest_constraints.txt](https://github.com/certbot/certbot/blob/5536c912238d96b15295a1b6938472fd24cf40ce/tools/oldest_constraints.txt#L39) to `1.6.0` and move it under the section for "Our setup.py constraints".
3. Mention that `google-api-python-client>=1.6` must be used for Certbot to automatically obtain credentials through the metadata service [here](https://github.com/certbot/certbot/blob/5536c912238d96b15295a1b6938472fd24cf40ce/certbot-dns-google/certbot_dns_google/__init__.py#L35-L43).

This fixes the problem in newer versions of the plugin by properly declaring our dependencies without adding any additional complexity on our end. If in the future, Debian/Ubuntu tries to update the version of this plugin, they can do so either by updating `google-api-python-client` (my preferred solution) or by modifying `setup.py` on their end.

What do you think?

cpcallen:
@bmw's proposed 'fix' isn't ideal for those of us using Debian 10 in production.  If this PR provides a workaround that would allow credential-file-less operation on buster that would be very helpful.

bmw:
With either approach, people using the packages in Debian 10 unfortunately would not see a benefit until the python3-certbot-dns-google package is updated by the Debian Developers. The current package there is over 2 years old so I'm unfortunately not sure we should expect that to happen anytime soon.

If it did though, I think the only downside of my approach is if they choose to not update the python3-googleapi package when doing this and instead change the dependencies from what we say we require upstream knowingly breaking this feature. Maybe they'd do that, but since there were no breaking changes between versions 1.5.5 and 1.6.0 of Google's library, I'd like to think they'd update that library as well so both users of our library and other users of python3-googleapi could benefit.

cpcallen:
That seems like a reasonable explanation, thanks!

Issue #5087, opened 2017-09-07
Feature request: combined privkey and fullchain file for jabber servers


IvanAnishchuk:
At this moment, both [ejabberd](https://github.com/jabberd2/jabberd2/issues/109) and [jabberd2](https://github.com/jabberd2/jabberd2/wiki/InstallGuide-CommonConfigurationTasks#521-assigning-a-certificate-for-use-by-jabber-clients) require private key and certificate to be combined in a single file. Jabberd2 maintainers even explicitly [refused](https://github.com/jabberd2/jabberd2/issues/109) to add an option for separate files and I strongly suspect that with ejabberd it would be the same. I don't have a complete list but I also suspect that there are other pieces of software with similar requirement. So I think it's safe to assume that we do have a use case, albeit slightly specialized one.

It was never a bit deal with certificates valid for a year or longer, just a simple `cat` command very similar to one you need for other servers if you don't have a fullchain in your CA bundle and the same can be used with certbot as a `renew-hook` but I think it would improve learning curve and eliminate human factor better if we had a option (or maybe even default action) instead.

For example, `cat` hook might have a problem with file permissions. Or it can fail for some reason and leave "combined" file not updated without user noticing. Or in case of revokations and deletions it's very likely that user will forget to add a delete-hook (is it even possible?) and invalid certificate will continue to be used. Then cat-ed file will be a real file unlike other four which are usually symlinks which open waay for yet another set of possible inconsistencies. And overall I feel that hooks should be used to reload servers, not for cert files manipulations.

Not to mention that inexperienced users may find it hard to figure out what hooks are and how to use them for this (and option that specifically mentions some of the servers that need it and standard path for the combined file that configuration guides/examples can use are way easier to find and use with less room for error).

So I propose to add an option to certbot that will combine privkey + cert + chain and put it into fifth file alongside the usual privkey, cert, chain, and fullchain. It's a pretty straightforward enhancement that is extremely unlikely to make code worse or break anything. It would be even simpler if we just added a fifth file by default but to that there might be some objections.

I am a python programmer and I can code everything myself and provide a PR but first I want to have a discussion to settle certain key details:

1. What is the best name for the fifth file? combined.pem? keycert.pem? everything.pem? Something else? (Although I need it for jabber servers in particular I think name should be more generic.)

2. What is the best name for the option? `--combine-key-and-cert`?

[...]

PR #7962, opened 2020-04-30
add a new lineage item containing the private key and the full cert chain

jnahmias:
This patch modifies the lineage definition to now include a fifth item: privkey_fullchainNNN.pem. This format is needed by a number of different TLS-capable servers including jabber, lighttpd, courier-imap, monit, and others; which require all the components to be present in a single file in the proper order (which is passed to OpenSSL).

Closes: #5087


jnahmias:
@bmw - I know I'm still fixing tests, but would appreciate feedback on the approach I've taken.

bmw:
Hey @jnahmias. Feel free to keep working on this if you want, but like you suggested in your ping, it might be a good idea for us get on the same page about a high level design here before you spend more time polishing the specific approach you've implemented here.

While I personally don't have a strong opinion or confidence what the best approach is, my most recent thoughts on the issue can be found at https://github.com/certbot/certbot/issues/5087#issuecomment-436836851 where I argued against an approach like this.

She's currently on vacation, but m0namon was the one that was currently assigned this issue. @m0namon, did you end up thinking any more about this? Are you still in favor of a design like you described https://github.com/certbot/certbot/issues/5087#issuecomment-436534061 which seems similar to the one implemented in this PR?

My only potential concerns with it are confusing users by adding more files under `/etc/letsencrypt/live/<domain>/` by default and creating another copy of the user's private key that the majority of Certbot users will not use (although we're already a little sloppy here because a copy of the user's key is also stored in `/etc/letsencrypt/keys`).

@schoen, you've also expressed interest in some of these issues in the past and if you have opinions, I'd be interested in hearing them!

vlasky:
@bmw it's now December 2020. I really think this pull request should be merged. I am using MongoDB in one of my apps and it also needs this combined format certificate/key file.

I don't think the potential for user confusion outweighs the convenience and system administration gains from having this feature.

certbot already places a README file in the live/domain folder that explains the purpose of the various .pem files.

It would be sufficient to add one more line to explain that the privkey_fullchain.pem file is generated to support the many popular apps/services that require the combined format, such as jabber, lighttpd, courier-imap, monit, and mongod.

bmw:
we've decided not to support this directly in certbot so i'm going to close this. see https://github.com/certbot/certbot/issues/5087#issuecomment-2759690295 for more info about our decision

PR #7968, opened 2020-05-01
Platform-agnostic method to copy owner and permissions between files

adferrand:
Related to #7649 since @joohoi needs a method to copy owner and permissions together from a source file to a destination file.

This PR creates the method `copy_ownership_and_mode()` in `certbot.compat.filesystem` module to achieve this goal. Its behavior is consistent across Linux and Windows in respect to the security model that have been defined for Certbot on Windows.

The method behaves globally the same than `copy_ownership_and_apply_mode`, but this time the permissions are extracted from the source file. For Windows it means that the DACL is copied from the source to the destination with the same content.

### DIFF
--- a/certbot/certbot/compat/filesystem.py
+++ b/certbot/certbot/compat/filesystem.py
@@ -86,7 +86,8 @@ def copy_ownership_and_apply_mode(src, dst, mode, copy_user, copy_group):
 def copy_ownership_and_mode(src, dst, copy_user=True, copy_group=True):
     # type: (str, str, bool, bool) -> None
     """
-    Copy ownership (user and optionally group on Linux)
+    Copy ownership (user and optionally group on Linux) and mode/DACL
+    from the source to the destination.
     :param str src: Path of the source file
     :param str dst: Path of the destination file
     :param bool copy_user: Copy user if `True`
--- a/certbot/certbot/compat/filesystem.py
+++ b/certbot/certbot/compat/filesystem.py
@@ -78,6 +78,34 @@ def copy_ownership_and_apply_mode(src, dst, mode, copy_user, copy_group):
     chmod(dst, mode)
 
 
+# Quite similar to copy_ownership_and_apply_mode, but this time the DACL is copied from
+# the source file on Windows. The DACL stays consistent with the dynamic rights of the
+# equivalent POSIX mode, because ownership and mode are copied altogether on the destination
+# file, so no recomputing of the DACL against the new owner is needed, as it would be
+# for a copy_ownership alone method.
+def copy_ownership_and_mode(src, dst, copy_user=True, copy_group=True):
+    # type: (str, str, bool, bool) -> None
+    """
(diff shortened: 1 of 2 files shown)


certbot/certbot/compat/filesystem.py
```diff
@@ -78,6 +78,34 @@ def copy_ownership_and_apply_mode(src, dst, mode, copy_user, copy_group):
...
 
+# Quite similar to copy_ownership_and_apply_mode, but this time the DACL is copied from
+# the source file on Windows. The DACL stays consistent with the dynamic rights of the
+# equivalent POSIX mode, because ownership and mode are copied altogether on the destination
+# file, so no recomputing of the DACL against the new owner is needed, as it would be
+# for a copy_ownership alone method.
+def copy_ownership_and_mode(src, dst, copy_user=True, copy_group=True):
+    # type: (str, str, bool, bool) -> None
+    """
+    Copy ownership (user and optionally group on Linux)
```
bmw, line 89: nit: I'd expand this to say something like "Copy ownership (user and optionally group on Linux) and mode/DACL from the source to the destination."

certbot/tests/compat/filesystem_test.py
```diff
@@ -280,14 +280,14 @@ def test_makedirs_switch_os_mkdir(self):
         self.assertEqual(original_mkdir, std_os.mkdir)
 
 
-class OwnershipTest(test_util.TempDirTestCase):
+class CopyOwnershipAndModeTest(test_util.TempDirTestCase):
     """Tests about copy_ownership_and_apply_mode and has_same_ownership"""
```
bmw, line 284: nit: We should update this comment since we're testing more functions here.

certbot/tests/compat/filesystem_test.py
```diff
@@ -334,6 +334,24 @@ def test_has_same_ownership(self):
...
+        src = self.probe_path
+        dst = _create_probe(self.tempdir, name='dst')
+
+        filesystem.chmod(src, 0o700)
+        self.assertTrue(filesystem.check_mode(src, 0o700))
+        self.assertTrue(filesystem.check_mode(dst, 0o744))
+
+        # Checking an actual change of owner is tricky during a unit test, since we do not know
+        # if any user exists beside the current one. So we mock _copy_win_ownership. It's behavior
+        # have been checked theoretically with test_copy_ownership_and_apply_mode_linux.
```
bmw, line 348: Did you mean this?
```suggestion
        # have been checked theoretically with test_copy_ownership_and_apply_mode_windows.
```

certbot/certbot/compat/filesystem.py
adferrand: Done.

certbot/tests/compat/filesystem_test.py
adferrand: Done.

Issue #7863, opened 2020-03-25
Make externally snapped plugin updates more stable


bmw:
With [the way externally snapped plugins work](https://github.com/certbot/certbot/blob/master/tools/snap/README.md), Certbot and its dependencies mask the version of Certbot installed in the plugin snap. (See https://github.com/certbot/certbot/issues/7946 about removing the 2nd copy of Certbot altogether, but even if that's resolved, the purpose of this issue remains.) As a result, if the plugin snap is upgraded before the Certbot snap and that plugin requires a new version of Certbot, the plugin would break until Certbot is also updated.

When this problem happens, I expect it to be resolved quickly (unless the user did something like manually held Certbot's updates back), but since as far as I can tell there is no well defined ordering for updates, I expect problems like this to occur fairly regularly for users of externally snapped plugins. I personally don't like this at all and unless we can convenience ourselves that it is a very minor problem, I think we should either fix it or not offer externally snapped plugins. It's just doesn't seem stable enough to me for it to be the official way to install an external plugin with Certbot.

Luckily, we may have a solution here. Robie looked into https://forum.snapcraft.io/t/supported-snap-hooks/3795/ which describes a way to define hooks which if they fail, snaps will either not be updated or they will be rolled back. This needs to be tested and designed a little carefully to avoid something like a badly written plugin snap causing the Certbot snap to never upgrade, but it seems like this could fairly easily solve our problems.

This issue is mainly a research/design issue trying to design a system to avoid these problems with our own or third party plugins, although assuming we find a design we like, we'll need to make sure we implement it once we have Certbot/plugin snaps in this repo.

As of writing this, the instructions at https://github.com/certbot/certbot/blob/master/tools/snap/README.md#use-testing-and-development can be used to create proof-of-concept snap plugins.

bmw:
@ohemorange, I've update the proof of concept instructions so things should work for you now.

If you hit any issues, please let me know!

ohemorange:
tl;dr: Sticking a hook named `post-refresh` into the plugin’s `meta/hooks/` directory that fails if certbot is the wrong version for the newly updated plugin's needs should mostly work. There’s a few potential issues we might run into while implementing this, but most of those have backup solutions detailed below. The major design question is if we’re ok with having this solution not work for backwards-incompatible Certbot releases, and instead being sure to update plugins a release ahead of deprecations in those cases.

[...]

PR #7971, opened 2020-05-04
Add warning about ignoring our own warnings

bmw:
Coming out of the conversation at https://github.com/certbot/certbot/issues/7863 in the linked Google Doc, we should always have at least 1 release between updating one of our plugins to stop using a deprecated acme/certbot API and removing it from acme/certbot. Doing this gives the plugin changes time to propagate rather than potentially having the plugin break because Certbot was updated before the plugin had made the necessary changes.

This comment here should help ensure this.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,6 +4,12 @@
 [pytest]
 # In general, all warnings are treated as errors. Here are the exceptions:
 #   1- decodestring: https://github.com/rthalley/dnspython/issues/338
+# Warnings being triggered by our plugins using deprecated features in
+# acme/certbot should be fixed by having our plugins no longer using the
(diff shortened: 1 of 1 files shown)


pytest.ini
```diff
@@ -4,6 +4,12 @@
 [pytest]
 # In general, all warnings are treated as errors. Here are the exceptions:
 #   1- decodestring: https://github.com/rthalley/dnspython/issues/338
+# Warnings being triggered by our plugins using deprecated features in
+# acme/certbot should be fixed by having our plugins no longer using the
+# deprecated code and not ignored here. Fixing things in this way prevents us
```
ohemorange, line 9:
```suggestion
# deprecated code rather than adding them to the list of ignored warnings here. Fixing things in this way prevents us
```

PR #7984, opened 2020-05-12
Error out earlier in apache installer when mod_ssl is not available

### LINKED ISSUE #7612: Handle case where mod_ssl is not available
As flagged by @joohoi, especially when using Certbot packaged as a snap, it is possible that `mod_ssl` is not available in Apache on the host system. We should check for this case early and error out with a useful error message if its not found.

ohemorange:
Fixes #7612.

Testing for `ssl_module` patterned after `openssl_version()` code here: https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/_internal/configurator.py#L274

Previously we were getting tests to pass by setting `ssl_module` to `None`, because `enable_mod` figures out if it needs to enable the mod by [testing inclusion](https://github.com/certbot/certbot/blob/master/certbot-apache/certbot_apache/_internal/override_debian.py#L108). Now we error even if it exists but is `None`, so tests need to be updated. Also, I checked through all the places we expect errors to be raised in Apache tests to make sure we weren't accidentally catching the wrong error now, and I couldn't find any past the ones that are fixed here. I'd also expect coverage to catch those, and the ones it was catching are now taken care of.

### DIFF
--- a/certbot-apache/certbot_apache/_internal/configurator.py
+++ b/certbot-apache/certbot_apache/_internal/configurator.py
@@ -595,6 +595,10 @@ class ApacheConfigurator(common.Installer):
         # cert_key... can all be parsed appropriately
         self.prepare_server_https("443")
 
+        # If we haven't managed to enable mod_ssl by this point, error out
+        if "ssl_module" not in self.parser.modules or not self.parser.modules["ssl_module"]:
+            raise errors.PluginError("Could not find ssl_module; not installing certificate.")
+
         # Add directives and remove duplicates
         self._add_dummy_ssl_directives(vhost.path)
         self._clean_vhost(vhost)
--- a/certbot-apache/certbot_apache/_internal/configurator.py
+++ b/certbot-apache/certbot_apache/_internal/configurator.py
@@ -597,7 +597,8 @@ class ApacheConfigurator(common.Installer):
 
         # If we haven't managed to enable mod_ssl by this point, error out
         if "ssl_module" not in self.parser.modules or not self.parser.modules["ssl_module"]:
-            raise errors.PluginError("Could not find ssl_module; not installing certificate.")
+            raise errors.MisconfigurationError("Could not find ssl_module; "
+                "not installing certificate.")
 
         # Add directives and remove duplicates
         self._add_dummy_ssl_directives(vhost.path)
--- a/certbot-apache/certbot_apache/_internal/configurator.py
+++ b/certbot-apache/certbot_apache/_internal/configurator.py
@@ -614,21 +614,6 @@ class ApacheConfigurator(common.Installer):
             path["chain_path"] = self.parser.find_dir(
                 "SSLCertificateChainFile", None, vhost.path)
 
-        # Handle errors when certificate/key directives cannot be found
-        if not path["cert_path"]:
-            logger.warning(
-                "Cannot find an SSLCertificateFile directive in %s. "
-                "VirtualHost was not modified", vhost.path)
-            raise errors.PluginError(
-                "Unable to find an SSLCertificateFile directive")
-        elif not path["cert_key"]:
-            logger.warning(
-                "Cannot find an SSLCertificateKeyFile directive for "
-                "certificate in %s. VirtualHost was not modified", vhost.path)
-            raise errors.PluginError(
-                "Unable to find an SSLCertificateKeyFile directive for "
-                "certificate")
-
         logger.info("Deploying Certificate to VirtualHost %s", vhost.filep)
 
         if self.version < (2, 4, 8) or (chain_path and not fullchain_path):
--- a/certbot-apache/certbot_apache/_internal/configurator.py
+++ b/certbot-apache/certbot_apache/_internal/configurator.py
@@ -596,7 +596,7 @@ class ApacheConfigurator(common.Installer):
         self.prepare_server_https("443")
 
         # If we haven't managed to enable mod_ssl by this point, error out
-        if "ssl_module" not in self.parser.modules or not self.parser.modules["ssl_module"]:
+        if "ssl_module" not in self.parser.modules:
             raise errors.MisconfigurationError("Could not find ssl_module; "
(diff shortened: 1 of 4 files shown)


certbot-apache/tests/configurator_test.py
```diff
@@ -1349,6 +1314,20 @@ def test_deploy_cert_not_parsed_path(self):
                 self.assertTrue(mock_add.called)
         shutil.rmtree(tmp_path)
 
+    def test_deploy_cert_no_mod_ssl(self):
+        # Create
+        ssl_vhost = self.config.make_vhost_ssl(self.vh_truth[0])
+        self.config.parser.modules["socache_shmcb_module"] = None
+
+        self.assertRaises(errors.MisconfigurationError, self.config.deploy_cert,
```
bmw, line 1322: The error raised here changes depending on the system I run on it on.

On Ubuntu, it's the `"ssl_module" not in self.parser.modules` case in the line you added. On macOS, it's
```
Error running command ['apache2ctl', '-t', '-D', 'DUMP_MODULES'] for runtime parameters!

certbot.errors.MisconfigurationError: Error accessing loaded Apache parameters: ['apache2ctl', '-t', '-D', 'DUMP_MODULES']
```
with the relevant part of the traceback being:
```
> /my/local/path/certbot/certbot-apache/certbot_apache/_internal/configurator.py(596)_deploy_cert()
    595         # cert_key... can all be parsed appropriately
--> 596         self.prepare_server_https("443")
    597 

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/configurator.py(1185)prepare_server_https()
   1184 
-> 1185         self.prepare_https_modules(temp)
   1186         self.ensure_listen(port, https=True)

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/configurator.py(1330)prepare_https_modules()
   1329                 self.parser.aug.load()
-> 1330                 self.parser.reset_modules() # Reset to load the new ssl_module path
   1331                 # Call again because now we can gate on openssl version

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/parser.py(259)reset_modules()
    258         self.modules = {}
--> 259         self.update_modules()
    260         self.parse_modules()

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/parser.py(319)update_modules()
    318 
--> 319         matches = apache_util.parse_modules(self.configurator.option("ctl"))
    320         for mod in matches:

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/apache_util.py(197)parse_modules()
    196                "DUMP_MODULES"]
--> 197     return parse_from_subprocess(mod_cmd, r"(.*)_module")
    198 

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/apache_util.py(210)parse_from_subprocess()
    209     """ 
--> 210     stdout = _get_runtime_cfg(command)
    211     return re.compile(regexp).findall(stdout)

  /my/local/path/certbot/certbot-apache/certbot_apache/_internal/apache_util.py(237)_get_runtime_cfg()
    236             "Error accessing loaded Apache parameters: {0}".format(
--> 237                 command))
    238     # Small errors that do not impede
```
To fix this, what do you think about mocking out `self.config.prepare_server_https` since we're not trying to test any of its behavior in this test.

certbot/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-*
+* Error out in apache installer when mod_ssl is not available.
```
bmw, line 9: nit: We previously errored out, just the error message wasn't very helpful.

What do you think about saying something like "Improved error message in apache installer when mod_ssl is not available." and moving this to the changed section?

certbot-apache/certbot_apache/_internal/configurator.py
```diff
@@ -595,6 +595,11 @@ def _deploy_cert(self, vhost, cert_path, key_path, chain_path, fullchain_path):
         # cert_key... can all be parsed appropriately
         self.prepare_server_https("443")
 
+        # If we haven't managed to enable mod_ssl by this point, error out
+        if "ssl_module" not in self.parser.modules or not self.parser.modules["ssl_module"]:
```
bmw, line 599: One other thing...I think we should just make this line:
```suggestion
        if "ssl_module" not in self.parser.modules:
```
While we also check `not self.parser.modules["ssl_module"]` in `openssl_version()`, I think we have to do that because we need its value. Additionally, this check failing was previously only a warning and now it's a fatal error.

`"ssl_module" not in self.parser.modules` was previously a fatal error when looking for our own dummy SSL directives because of [exclude_dirs](https://github.com/certbot/certbot/blob/bcf33c665932160c2fe446e46e42c15b57c05a81/certbot-apache/certbot_apache/_internal/parser.py#L621) used in `parser.find_dir`.


Due to the way functions like [update_modules](https://github.com/certbot/certbot/blob/bcf33c665932160c2fe446e46e42c15b57c05a81/certbot-apache/certbot_apache/_internal/parser.py#L316) work, I worry that on some systems, we may not have a path here. I'm not certain there is a problem and I may be being a little paranoid, but I think it'd be safer to just check `"ssl_module" not in self.parser.modules` and I don't see any real downsides to us doing so.

If you make this change, you can also probably simplify/remove some of the other changes in this PR if you want.

alexzorin:
Is this going to make #7998 worse? 

I might be misreading, but the reporter's installer seems to work right now. After this change, Certbot would abort in the installer?

certbot-apache/tests/configurator_test.py
ohemorange: Looks like we can't quite just do that, since that function does other things that the brittle tests rely on (I'm getting a list index out of range error). I'll mock something else.

ohemorange: figured it out. please ignore this entire response, I forgot to remove a test.

PR #7989, opened 2020-05-13
ocsp: add support for public key hash ResponderIDs

### LINKED ISSUE #7986: ocsp: support key hash ResponderIDs, not just names.
This was first reported [on the forums](https://community.letsencrypt.org/t/invalid-ocsp-response-for-no-matching-responder-certificate-could-be-found/122497).

Certbot produces a warning when checking the revocation status of any certificate issued by [this Sectigo CA](https://censys.io/certificates?q=parsed.extensions.authority_key_id:%2017d9d6252767f931c24943d93036448c6ca94feb). 

>Invalid OCSP response for sectigo-leaf.pem: no matching responder certificate could be found.

I think the warning is produced because Certbot's (`cryptography`) OCSP client is assuming that the Basic Response's `ResponderID` is always going to be a name. 

https://github.com/certbot/certbot/blob/71e3d82e47696daa24c5f07bd2cb8c553b40babd/certbot/certbot/ocsp.py#L265-L272

[However](https://tools.ietf.org/html/rfc6960#section-4.2.2.3), it may be a hash of the public key:

>   o  either the name of the responder or a hash of the responder's
>      public key as the ResponderID;

This is exposed in `cryptography` as [`OCSPResponse.responder_key_hash`](https://cryptography.io/en/latest/x509/ocsp/#cryptography.x509.ocsp.OCSPResponse.responder_key_hash).

It seems that an OCSP client should implement both ways of identifying the OCSP issuer.

alexzorin:
For both cases where the the response is signed by the issuer, or by a
delegated OCSP signer.

Resolves #7986

### DIFF
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -14,7 +14,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-*
+* Add support for OCSP responses which use a public key hash ResponderID, fixing
+  interoperability with Sectigo CAs.
 
 More details about these changes can be found on our GitHub repo.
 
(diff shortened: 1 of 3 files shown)


ohemorange:
@bmw, I know you've been looking at our OCSP implementation recently

PR #7995, opened 2020-05-14
Add a '--renew-before-expiry' CLI flag which overrides the default

TomAnthony:
This is an attempt at adding a new flag `--renew-before-expiry` which works in the same way as the `renew_before_expiry` renewal config file option. It is often easier to supply a flag rather than find and update a config file, especially in scripted systems.

Notable is how this behaves when both the new CLI flag and config file option are present for a certificate. In that case the longer interval (i.e. earlier renewal date) is selected. If either option is set the same as the default then it is ignored.

This is motivated by the default 30 day period causing difficulty for customers of a project I work on, as they often have alarms that fire at the 30 day mark for vendor certs, so we wanted to renew at 35 days, but disliked the relative complexity of having scripts update renewal config files.

This is my first PR on the project, so hopefully I haven't missed anything! :)

## Pull Request Checklist

- [x ] If the change being made is to a [distributed component](https://certbot.eff.org/docs/contributing.html#code-components-and-layout), edit the `master` section of `certbot/CHANGELOG.md` to include a description of the change being made.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -247,6 +247,7 @@ Authors
 * [Thom Wiggers](https://github.com/thomwiggers)
 * [Till Maas](https://github.com/tyll)
 * [Timothy Guan-tin Chien](https://github.com/timdream)
+* [Tom Anthony](https://github.com/tomanthony)
 * [Torsten Bögershausen](https://github.com/tboegi)
 * [Travis Raines](https://github.com/rainest)
 * [Trung Ngo](https://github.com/Ngo-The-Trung)
--- a/certbot/CHANGELOG.md
+++ b/certbot/CHANGELOG.md
@@ -6,7 +6,7 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-*
+* Added `--renew-before-expiry` CLI flag, which allows overriding the default (currently 30 day) expiry interval, allowing easy renewal of certs at an earlier time. It can be used in conjunction with the `renew_before_expiry` renewal config file setting, in which case the the longer interval (i.e. earlier renewal date) will be accepted (provided it is different to the default value).
 
 ### Changed
 
--- a/certbot/certbot/_internal/cli/__init__.py
+++ b/certbot/certbot/_internal/cli/__init__.py
@@ -204,6 +204,14 @@ def prepare_and_parse_args(plugins, args, detect_defaults=False):
         version="%(prog)s {0}".format(certbot.__version__),
         help="show program's version number and exit")
     helpful.add(
+        ["automation", "renew"], "--renew-before-expiry",
+        metavar="EXPIRY_TIME_INTERVAL", type=str, dest="renew_before_expiry",
+        default=flag_default("renew_before_expiry"),
+        help="If a certificate "
+             "already exists for the requested domains, renew it if "
+             "it is within the specified time period. (Provide "
+             "a value such as '40 days', or '3 weeks').")
+    helpful.add(
         ["automation", "renew"],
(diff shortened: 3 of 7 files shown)


schoen:
Hi @TomAnthony,

Thanks! That's some functionality that we never got around to adding anywhere.

One thing that I'm mindful of is that we originally named some of these Certbot options things that turned out to be rather confusing or counterintutive for users. The worst example is probably `--renew-by-default`, which was later renamed to `--force-renewal`.

I wonder if `renew_before_expiry` is also in that territory to some extent. Do you think it might be clearer to users if we called the command-line flag for this something different? Like `--autorenewal-interval-before-expiry` or something? (The name of the option *in the config file* is probably fixed, and there's a general pattern of trying to make them match up with the command-line option names, but I'm not sure whether or not that would be more or less confusing here.)

schoen:
Oh, and I've demonstrated this by getting confused about whether this option *changes* the renewal configuration file on disk, or simply overrides it for the current invocation of Certbot when running with a verb that could potentially cause a renewal!

TomAnthony:
Hey @schoen ,

Thanks for the feedback!

Yes - I didn't much like the name, but kept with it for consistency, I think adding an alternative new name (in the same way as with `--force-renewal`) would be good.

It is also worth having a chat about the functionality (which plays to your [valid] confusion about the config file interaction).

At this point in the code there are three potential sources for this setting:

 - The default value from `RENEWER_DEFAULTS`, which is set to 30 days (and I doubt will ever change)
 - The value from the renewal config file for this cert. This setting, if unset, adopts the default from `RENEWER_DEFAULTS`, so always has a value
 - The value from this new command line switch, which if not provided also adopts the value from `RENEWER_DEFAULTS`

My question is: how should precedence be decided.

**Option 1**

The value in the command line switch takes precedence over everything else, if it is present (which in reality means if it is not set to the default, which is the only way to tell if it has been set).

**Option 2**

The value in the certs configuration file takes precedence (if it is not the default).

**Option 3 (what PR currently does)**

It examines both the command line and config file settings, and selects the value which leads to the earlier renewal. Once again, it ignores them if they have the default value (assuming they've not been provided).


I went with *Option 3*, because I felt in most instances where both the renewal config file and the command line have been provided, the user is probably trying to renew certificates earlier than the standard 30 days, rather than later. It is also more robust to malformed imports, leaning towards ensuring certs are renewed and valid.

I think if you are trying to renew later, then the onus is on you to get those settings correct. However, I could equally be happy with *Option 1* or *Option 2* if they were preferred (and if there was precedence elsewhere, which I didn't immediately see).

Let me know your thinking!

_Note_: The command line flag never changes the actual config file setting (i.e. never writes to the config). I think that would be confusing 

**Flag name**

I think your point about the name being confusing is valid. The code talks about the *interval* but I didn't feel like that is very helpful for the user, where they typically think of when the cert will `expire`, but the `--renew-before-expiry` is very unclear here.

Suggestions:

 - `--renew-if-expiry-within` - I like this as it is barely any longer but is clear what it does and that it expects a value argument along with it.
 - `--renewal-expiration-threshold`
 - `--expiration-threshold`
 - `--renew-at`

I've tried to bear in mind that the flag could be used to force _later_ renewal than usual rather than just earlier, so the name needs to be neutral on that front.

Let me know what you think about this and the precedence question and I can update the PR accordingly. I can also update the docs to explain the precedence. 🙂

TomAnthony:
Hey @schoen - any thoughts on the above? I still like `--renew-if-expiry-within`.

TomAnthony:
Hey - just wondering on status of this. @schoen - any update? Thanks.

ggruber:
I's also like to see this feature implemented

PR #8000, opened 2020-05-18
Print cause of exit in red text

bmw:
Certbot writes logging messages at the warning level or higher to the terminal in red text. It turns out fatal errors aren't always printed in red though.

I noticed this when testing https://github.com/certbot/certbot/pull/7984. Relevant output looks like:
Screen Shot 2020-05-18 at 10 22 11 AM

The fatal error message here is "Could not find ssl_module; not installing certificate." but it's not given any special significance and I think it's harder to see because of the other red text and the important notes section.

While we have plans to remove the important notes section in the future, I personally think another thing we should do here is to display the fatal error in red as well which is done in this PR.

I started to add tests here, but I think we're pretty well covered by the existing tests checking that `SystemExit` is raised and that the error message is written to the terminal in all cases. I'm happy to add more if there's something in particular the reviewer of this PR thinks should be tested.

### DIFF
--- a/certbot/certbot/_internal/log.py
+++ b/certbot/certbot/_internal/log.py
@@ -322,10 +322,10 @@ def post_arg_parse_except_hook(exc_type, exc_value, trace, debug, log_path):
         logger.error('Exiting abnormally:', exc_info=exc_info)
     else:
         logger.debug('Exiting abnormally:', exc_info=exc_info)
+        # Use logger to print the error message to take advantage of
+        # our logger printing warnings and errors in red text.
         if issubclass(exc_type, errors.Error):
-            # Use logger to print the error message to take advantage of
-            # our logger printing warnings and errors in red text.
-            logger.error(exc_value)
+            logger.error(str(exc_value))
             sys.exit(1)
         logger.error('An unexpected error occurred:')
         if messages.is_acme_error(exc_value):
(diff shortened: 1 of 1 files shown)


certbot/certbot/_internal/log.py
```diff
@@ -322,15 +322,23 @@ def post_arg_parse_except_hook(exc_type, exc_value, trace, debug, log_path):
         logger.error('Exiting abnormally:', exc_info=exc_info)
     else:
         logger.debug('Exiting abnormally:', exc_info=exc_info)
+        # Use logger to print the error message to take advantage of
+        # our logger printing warnings and errors in red text.
         if issubclass(exc_type, errors.Error):
-            sys.exit(exc_value)
+            logger.error(str(exc_value))
+            sys.exit(1)
```
schoen, line 329: Did Certbot previously sometimes exit with a more specific error code depending on the reason for the failure? (If so, people might have scripts that are assuming that.) Is `exc_value` a numeric value? If not, does `sys.exit()` just interpret it as an int after interpreting it as a bool, or something?

bmw: Nope! `exc_value` is the exception that was thrown and is crashing Certbot.

From the [sys.exit docs](https://docs.python.org/3.8/library/sys.html#sys.exit):

> The optional argument arg can be an integer giving the exit status (defaulting to zero), or another type of object. ... If another type of object is passed, None is equivalent to passing zero, and any other object is printed to stderr and results in an exit code of 1.

schoen: Aha! That makes sense.

bmw:
Thanks for the review!

PR #8002, opened 2020-05-19
dns-cloudflare: Allow config-file overriding of Zone IDs

### LINKED ISSUE #7893: dns-cloudflare requires too many permissions
The `dns-cloudflare` module should provide support for manually specifying the `zone_id`. This quote from the documentation is well known:

> However, due to some shortcomings in Cloudflare’s implementation of Tokens, Tokens created for Certbot currently require Zone:Zone:Read and Zone:DNS:Edit permissions for **all** zones in your account. While this is not ideal, your Token will still have fewer permission than the Global key, so it’s still worth doing. Hopefully Cloudflare will improve this in the future.

You do not really need to give the token rights for **ALL** zones. You can successfully generate certificates with cloudflare if you use `manual-auth-hook`:
```
    certbot certonly \
      --manual \
      --non-interactive \
      --preferred-challenges dns \
      --manual-auth-hook "{{ work_folder }}/letsencrypt_auth.sh" \
      --manual-cleanup-hook "{{ work_folder }}/letsencrypt_cleanup.sh" \
      --rsa-key-size 4096 \
      --agree-tos \
      --manual-public-ip-logging-ok \
      --no-eff-email \
      --email {{ letsencrypt_email }} \
      --domains "*.{{ fqdn }}"
```
And having the following in your `letsencrypt_auth.sh`:
```
#!/bin/bash
EMAIL="xxx"
ZONE_ID="xxx"
TYPE="TXT"
NAME="_acme-challenge.$( echo "$CERTBOT_DOMAIN" | cut -d"." -f1-2 )"
CONTENT="$CERTBOT_VALIDATION"
TTL="1"
curl -X POST "https://api.cloudflare.com/client/v4/zones/$ZONE_ID/dns_records/" \
    -H "Authorization: xxxxxxxx" \
    -H "Content-Type: application/json" \
    --data '{"type":"'"$TYPE"'","name":"'"$NAME"'","content":"'"$CONTENT"'","ttl":'"$TTL"'}' 
```

When running with `certbot certonly --dns-cloudflare`, you will get an error as follows:
> Unable to determine zone_id for xxx using zone names: ['xxx', 'xxx', 'xxx', 'xxx']. Please confirm that the domain name has been entered correctly and is already associated with the supplied Cloudflare account. The error from Cloudflare was: 0 Actor 'com.cloudflare.api.token.xxxx' requires permission 'com.cloudflare.api.account.zone.list' to list zones

This happens because certbot forces you to give it access to all the zones under the account, instead of allowing you to manually specify the zone id.

Tugzrida:
Since the introduction of support for Cloudflare's limited-scope API Tokens in #7583, API Tokens for use with certbot have required access to all zones in a user's Cloudflare account due to a quirk of how Cloudflare's API works. This is obviously not ideal from a security perspective.

Since then, myself and @bmw have spoken with people at Cloudflare about fixing the bug in their API, however it is apparently not an easy fix, and they don't have a timeline. This PR implements an (optional) workaround with a slight cost to usability, meaning the three ways to authenticate to the Cloudflare API will now be:

- (existing) Using a Cloudflare API Token with the config option `dns_cloudflare_api_token`. This Token must have DNS Edit access to all zones in a user's Cloudflare account, and so is not ideal, but is easy to setup and is more restrictive than the global key.

- (new) Using a Cloudflare API Token with the config option `dns_cloudflare_api_token` **and providing the Cloudflare zone IDs for the zone(s) for which certs are required with `dns_cloudflare_zone_ids`.** With this option, the Token will only need DNS Edit access to the zone(s) which you need certs for.

- (existing) Using Cloudflare's global API key with the config options `dns_cloudflare_email` and `dns_cloudflare_api_key`. The Global Key grants full access to a user's entire Cloudflare account, and hence it is recommended against in the docs.

These three methods are documented as "Easy setup", "Advanced setup" and "Legacy setup" respectively.

This PR hence closes #7893, however I will keep a tab on Cloudflare's API so that the workaround can be removed from the docs once it's no longer needed. Depending on the way that Cloudflare fixes their bug, removing the actual functionality of this workaround will most likely require any users utilising it to modify their Token permissions, and so probably won't be possible without breaking things.

There's only one thing I feel might need some work: When using the new authentication method, if a zone ID is missing from the config file, the code ends up in the same state as it would when using a Token which is missing zone listing abilities in the first mode listed above. The error is:

certbot.errors.PluginError: Unable to determine zone_id for example.com using zone names: ['example.com', 'com']. Please confirm that the domain name has been entered correctly and is already associated with the supplied Cloudflare account. The error from Cloudflare was: 0 Actor 'com.cloudflare.api.token.<tokenID>' requires permission 'com.cloudflare.api.account.zone.list' to list zones

I feel like this needs reworking to be clearer, as there are probably about 3 cases that will result in the same error(some exclude the "The error from Cloudflare was:..." part).

### DIFF
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
@@ -51,7 +51,8 @@ including for renewal, and cannot be silenced except by addressing the issue
 Easy setup
 ^^^^^^^^^^
 
-Create a Token in your `Cloudflare dashboard <https://dash.cloudflare.com/profile/api-tokens>`_
+Create a Token in your
+`Cloudflare dashboard <https://dash.cloudflare.com/?to=/:account/profile/api-tokens>`_
 with ``Zone:Zone:Read`` and ``Zone:DNS:Edit`` permissions for **all** zones in your account.
 If you wish to restrict token access on a per-zone level, follow the `advanced setup`_.
 
@@ -68,7 +69,8 @@ Copy and save the token in a file like so:
 Advanced setup
 ^^^^^^^^^^^^^^
 
-Create a Token in your `Cloudflare dashboard <https://dash.cloudflare.com/profile/api-tokens>`_
+Create a Token in your
+`Cloudflare dashboard <https://dash.cloudflare.com/?to=/:account/profile/api-tokens>`_
 with ``Zone:DNS:Edit`` permissions for the specific zones for which you need certificates.
 
 You will also need to add the Zone ID for each zone(from the bottom right of each zone page in
@@ -94,8 +96,8 @@ the entire Cloudflare API for all domains in your account, meaning it could caus
 damage if leaked. **If possible, you should use a Cloudflare Token.**
 
 Copy your Global Key from your
-`Cloudflare dashboard <https://dash.cloudflare.com/profile/api-tokens>`_ and save it with your
-Cloudflare account's email address in the configuration file:
+`Cloudflare dashboard <https://dash.cloudflare.com/?to=/:account/profile/api-tokens>`_
+and save it with your Cloudflare account's email address in the configuration file:
 
 .. code-block:: ini
    :name: certbot_cloudflare_key.ini
--- a/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
+++ b/certbot-dns-cloudflare/certbot_dns_cloudflare/__init__.py
@@ -93,8 +93,9 @@ A Global Key was previously used by Cloudflare for authentication, however this 
 the entire Cloudflare API for all domains in your account, meaning it could cause a lot of
 damage if leaked. **If possible, you should use a Cloudflare Token.**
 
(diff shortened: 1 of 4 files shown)


Tugzrida:
Totally coincidentally, I've just read a [blog post](https://blog.cloudflare.com/deeplinks-and-scrollanchor/) about deeplinking to Cloudflare's dash. I've now updated the url used in the plugin, which should make it a bit easier for people with more complicated Cloudflare accounts.

bmw:
Thanks for the PR @Tugzrida.

I have some other work I should really prioritize first, but I'll take a look at this in the next couple weeks.

@adferrand, since this is a PR for our DNS plugins, it'll need a 2nd review. If you have the time and interest to be the other reviewer here, I think it makes the most sense for you to do so. As usual, if not, no worries :)

bmw:
I haven't tested it, but I think Cloudflare just fixed everything! See https://github.com/certbot/certbot/issues/7893#issuecomment-632237045.

Tugzrida:
Hahaha of course Murphy would have it that way! I'll close this and work on a new PR that tidies up the docs and error messages.

PR #8007, opened 2020-05-20
apache: handle statically linked mod_ssl

alexzorin:
In #7771, the Apache configurator gained the ability to identify what
version of OpenSSL Apache's ssl_module is linked against. However, the
detection was only functional if the module was built as a DSO (which is
almost always the case).

This commit covers the case where the ssl_module is statically linked
within the Apache binary. It requires the user to specify the path to
the binary (with --apache-bin) and emits a warning if static linking is
detected but no path has been provided.


alexzorin:
@bmw this is ready for review, unless you want to wait for more discussion on the issue.

Issue #7945, opened 2020-04-24
Make Certbot find externally snapped plugins


bmw:
There's a patch that does this at https://github.com/basak/certbot-snap-build/commit/f9ba645c9ce0fdd42504d3dc3cb53b54f6a6f375 with an alternate approach and pros and cons of it discussed at https://github.com/basak/certbot-snap-build/issues/19.

Regardless of the approach taken, this will need to work some way to use externally snapped plugins in Certbot.

bmw:
This needs to be blocked on https://github.com/certbot/certbot/issues/7667 because we told the snap store maintainers we wouldn't implement this until that system is in place.

ohemorange:
I think adding itself to `PYTHONPATH` and being clear in our example code/any instructions we write that plugins should just append to `PYTHONPATH` and use names similar to ours is the cleanest course of action. In practice, most plugins I've seen are very simple and follow a template, so I wouldn't worry too much about fancy things happening along `PYTHONPATH`.

Issue #7667, opened 2020-01-06
Make external snap plugins permissions more explicit


bmw:
From [conversations with the snap team](https://forum.snapcraft.io/t/certbot-request-for-classic-snap-approval/6240/19), we need to make the fact that connecting externally snapped plugins to the Certbot snap gives them unrestricted root access to the system more explicit.

There are some ideas on how we might accomplish this in the linked thread, but I don't have an exact design in mind or even know what is possible. I think it's probably a good idea to get a 2nd +1 on the design here.

bmw:
I think in the future it would be nice for our own externally packaged plugins to automatically connect as described at the top of https://snapcraft.io/docs/content-interface.

I think we should try and make sure the design here allows for that possibility in the future.

bmw:
The problem described in https://github.com/certbot/certbot/issues/7863 was something that initially had me considering that we don't offer externally snapped plugins at all. Because of that, we may want to make sure we have a resolution for that issue before this one to avoid unnecessary work. Robie seems to have found us a solution to #7863 and this issue could otherwise be done in parallel, but if we have the option, let's prioritize the other issue first.

I'm adding the blocked label to help signal that when looking at issues at a high level.

ohemorange:
> I think in the future it would be nice for our own externally packaged plugins to automatically connect as described at the top of https://snapcraft.io/docs/content-interface.
>
> I think we should try and make sure the design here allows for that possibility in the future.

I'm pretty certain that the connect command cannot be given additional command-line arguments. Given that, the option needs to be a separate command. Taking these into account, I'm pretty sure this means that it won't be possible to just say `snap install certbot certbot-dns-whatever` (or whatever the syntax is to install all of them at the same time. This is because auto-connection happens [when the snap is installed](https://snapcraft.io/docs/interface-management). So, our installation instructions would have to look something like:

1. install certbot snap
2. run command to accept condition
3. install certbot plugin

Technically, if we want multiple plugins to be installed, steps 2 and 3 would need to repeated each time. It's pretty trivial to change this so that step 2 only needs to be run once ever to apply to all certbot plugins (though we might want to double check that it's ok to do this), but I don't see a way to not have step 2 be dependent on step 1 (you can't set attributes for plugins that haven't been installed yet based on my testing). I don't think it's that bad of a solution, but also I don't know of another one.

bmw:
[...]

PR #8009, opened 2020-05-21
Add the content interface to Certbot

bmw:
This PR moves the setup of the content interface which we plan to use to support external plugins from the [snap-plugin branch](https://github.com/certbot/certbot/tree/snap-plugin) to `master`. This being in `master` might make it easier to make progress on issues like https://github.com/certbot/certbot/issues/7667.

I think it's OK to land this in `master` now because:

1. Certbot isn't configured to consume anything shared over the interface yet. That work is tracked in https://github.com/certbot/certbot/issues/7945.
2. While I don't think we currently have any plans to change anything here, since the Certbot snap is still in its beta phase, I think we can change whatever we want here whenever we want.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -85,3 +85,9 @@ parts:
     # After certbot-apache to not rebuild duplicates (essentially sharing what was already staged,
     # like zope)
     after: [certbot-apache]
+
+plugs:
+  plugin:
+    interface: content
+    content: certbot-1
+    target: $SNAP/certbot-plugin


bmw:
You can see the snap being built and tested with this change at https://travis-ci.com/github/certbot/certbot/builds/167656481.

bmw:
On 2nd thought, one problem with landing this now is someone in theory could connect something to the Certbot snap now without the explicit warning and I think that connection will be preserved even after the Certbot snap is updated to load external plugins.

This probably doesn't actually matter, but it may make sense to delay landing this until https://github.com/certbot/certbot/issues/7945 is (almost) ready.

bmw:
Now that #8013 has landed I think we can safely land this too.

Issue #7976, opened 2020-05-05
Upgrade Zope dependencies in Certbot to work with setuptools>=46


adferrand:
In setuptools 46.0.0, a deprecated API have been removed: `setuptools.Feature`. This can break some installations, as explained in https://github.com/pypa/setuptools/issues/2017, and it precisely break ours during the compilation of the Certbot dockers, because the base docker contains now this latest version of setuptools.

It seems also from this same issue that `zope.interface<=5.0.0` is still using this deprecated API during its installation.

We should update our dependencies to zope in order to get rid elegantly of this error, and allow certbot installation in most recent versions of setuptools.

bmw:
Fixed in #8012.

PR #8012, opened 2020-05-22
Upgrade pinned certbot dependencies

adferrand:
This upgrades the pinned dependencies of Certbot using `letsencrypt-auto-source/rebuild_dependencies.py`. It fixes in particular #7976 by upgrading the `zope` packages, that now can be built from source with recent versions of `setuptools`.


bmw:
LGTM.

@ohemorange, since this is technically a change to certbot-auto, to be cautious, can you be the 2nd reviewer on this? You can see the full test suite passing with these changes at https://travis-ci.com/github/certbot/certbot/builds/168251266.

To provide some more context on what's going on here, we have one set of pinned dependencies that is used in certbot-auto, our Docker images, snaps, tests (outside of "oldest" and unpinned tests), etc.

As described at https://github.com/certbot/certbot/issues/7976, the current versions of these dependencies do not work with newer versions of `setuptools`. This is a problem for us because the snap [Python plugin](https://snapcraft.io/docs/python-plugin) doesn't give us a way to pin things like `pip` and `setuptools` and they're always upgraded to the latest version. Things currently work because `pip` uses a prebuilt wheel for zope on amd64, but it breaks when trying to use the sdist to build the snap on other architectures which Adrien is currently working on.

Given all the testing we have using these dependencies, automation we have around updating them (`letsencrypt-auto-source/rebuild_dependencies.py`), and the fact I don't think we've ever shipped a bug due to updating these dependencies, it's hard for me to imagine there being a problem here, but since it's technically certbot-auto, I thought it'd be a good idea to get a 2nd +1.

ohemorange:
Has anyone already done a rest release with these dependencies? Sometimes we hit things there that didn't come up for strange reasons elsewhere.

On a separate note, "the snap Python plugin doesn't give us a way to pin things like pip and setuptools and they're always upgraded to the latest version" is a little worrying for me, given that we've previously had problems with updates to those packages breaking things for us on their own.

bmw:
> Has anyone already done a rest release with these dependencies?

Good idea. I just did this and all test farm tests passed.

> On a separate note, "the snap Python plugin doesn't give us a way to pin things like pip and setuptools and they're always upgraded to the latest version" is a little worrying for me, given that we've previously had problems with updates to those packages breaking things for us on their own.

I agree it's not ideal. In general I think using the latest version of `pip` & co. is a simple and fairly reasonable strategy, however, as we seen in the past 5 years with Certbot, occasionally things temporarily break.

We can ask the `snapcraft` maintainers to allow us to pin it, however, I personally don't think we should block our snap work on getting this feature. If a new version ever causes a problem, I think it'll happen when Travis is building and testing the snap rather than it hitting our users. Builds failing is annoying, but as we know, it happens so I don't think it's that big of a problem right now personally.

ohemorange:
> We can ask the snapcraft maintainers to allow us to pin it, however, I personally don't think we should block our snap work on getting this feature.

I like this plan.

bmw:
If anyone is curious, I went to open an issue to request that feature but I think we may have it in the latest version of `snapcraft` that they haven't quite marked as `stable` yet. See https://github.com/certbot/certbot/issues/8023.

PR #8020, opened 2020-05-26
OS umask Context Manager

Mercerenies:
This PR addresses #7992, creating a context manager in `certbot.util` and implementing its use in the places the `os.umask` / `finally` pattern is used in the current codebase.

## Pull Request Checklist

- [ ] If the change being made is to a [distributed component](https://certbot.eff.org/docs/contributing.html#code-components-and-layout), edit the `master` section of `certbot/CHANGELOG.md` to include a description of the change being made.
- [ ] Include your name in `AUTHORS.md` if you like.

PR #8029, opened 2020-05-28
Support for IP address SAN

orangepizza:
## Pull Request Checklist
tested against pebble,  --standalone works, IPv6 doesn't work currently (as SAN parser's part_separator colliding with ipv6's (both ':') need better parser idea)
ipv4 can be treated as if all-numeric domain so it needed minimal change
(need additional test....)
while let's encrypt (the CA) doesn't support IP address currently, this will update acme package imported by boulder tester, so it could be used to develop boulder to suppport IP address in SAN


orangepizza:
ouch I forgot to remove type hint I used for debug

orangepizza:
code coverage failed with current test set: all about added code about IP addresses

orangepizza:
can use inet.pton for ipv6, but it looks like windows version doesn't have it

orangepizza:
as all autometic check passed: now writing to support ipv6

orangepizza:
support for ipv6 added
by the way I don't think this will work with webserver plugins like --apache or --nginx

bmw:
Thanks for the PR @orangepizza!

To be honest, we probably won't have the cycles to review this for a while, but I'll take a look when I can.

orangepizza:
wonder why py2.7 can't find newly created module when 3.x can?

bmw:
The problem looks similar to the one I just describe a solution for at https://github.com/certbot/certbot/pull/8020#issuecomment-641422547. Essentially:

> To fix the Travis failures, you need to update [this line](https://github.com/certbot/certbot/blob/cc07722b3ebe2c733c20159487acd7c87ce42d79/certbot-apache/setup.py#L15) to `certbot>=1.6.0.dev0` and [this line](https://github.com/certbot/certbot/blob/cc07722b3ebe2c733c20159487acd7c87ce42d79/certbot-apache/local-oldest-requirements.txt#L3) to `-e certbot[dev]`.

orangepizza:
duplicated util function I made to Apache_util so keep same backward compatibility

hosseinsh:
Hi @bmw, I was wondering what would help with reducing the overhead of reviewing and picking up this PR?
Would it help, if it this was a smaller PR, just focusing on adding the IP address SAN support?

bmw:
Can you explain to me what your interest in this feature is? Let's Encrypt closed the equivalent PR on their side at https://github.com/letsencrypt/boulder/pull/4920#issuecomment-832104881 saying that they weren't going to add support for IP SANs in the near future.

hosseinsh:
Brad, thanks for the input. I wish Let's Encrypt was not postponing adding IP cert support.

There are, however, other CAs which rely on the ACME protocol and are working on enabling IP-based certificates.
With that in mind, it would be greatly appreciated, if certbot could already go ahead with adding IP cert support.
Today, the client just considers DNSNames when placing a new order.

bmw:
Interesting! What other ACME CA is currently issuing IP certs? I didn't know there was any.

The Certbot team is honestly stretched pretty thin and had no plans to work on this anytime soon, but depending on the level support in Certbot or our ACME library that you wanted to see and what could be done to simplify the review here, I could try to take a look.

orangepizza:
feels like this PR got needlessly large because it's a year ago and all the commits are pulled into ("merge maseter into PR") 
I'll try to reduce the PR size in weekend

bmw:
@hosseinsh, do you need/want support for IP SANs in Certbot or just the acme library? I assume it's the latter and if so I'd like to only add support there for now.

With that said, I appreciate all the work you've done here @orangepizza and I have every intention of potentially using all of this when/if we implement IP SAN support in Certbot. It's just while demand for support IP SANs in Certbot is low, I'd prefer not to try and support it for the sake of simplicity.

If people are able to open a relatively small PR with unit tests adding IP SAN support to our acme library, I'll try and take a look. Anything the community can do to get such a PR up and polished to speed up the review process on my end would be much appreciated.

hosseinsh:
@bmw, that is correct. We rely on the acme library for issuance. Understandable that initially just adding support in the ACME library would be desirable, though reviewing @orangepizza's commit's on this PR, the changes seem relatively compact, and hopefully allow for a smooth review process.

I hope starting with adding IP address support to the ACME library would also address @orangepizza's use-case.
I believe the changes in `make_csr` and `enforce_domain_sanity` are the relevant changes to certbot, which we could move to a separate PR.

orangepizza:
wrote the new PR, https://github.com/certbot/certbot/pull/8904

PR #8054, opened 2020-06-04
Make Certbot find externally snapped plugins

### LINKED ISSUE #7945: Make Certbot find externally snapped plugins

ohemorange:
Fixes #7945.

### DIFF
--- a/certbot.wrapper
+++ b/certbot.wrapper
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+join() {
+    sep=$1
+    first=$2
+    shift 2
+    echo -n "${first}"
+    for item in "$@"; do echo -n "${sep}${item}"; done
+    echo
+}
+
+paths=$(for plugin_snap in $(snap connections certbot|sed -n '2,$p'|awk '$1=="content[certbot-plugin]"{print $3}'|cut -d: -f1); do echo /snap/$plugin_snap/current/lib/python2.7/site-packages; done)
+export CERTBOT_PLUGIN_PATH=$(join : $paths)
+exec certbot "$@"
--- a/certbot.wrapper
+++ b/certbot.wrapper
@@ -3,12 +3,14 @@
 join() {
     sep=$1
     first=$2
-    shift 2
-    echo -n "${first}"
-    for item in "$@"; do echo -n "${sep}${item}"; done
-    echo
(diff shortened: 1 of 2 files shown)


certbot.wrapper
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
bmw, line 1: nit: `/bin/sh` seems to work, but I think the behavior of commands like `echo -n` depends on the system the script is run on. Explicitly using bash should fix this, I tested it in the snap, and it worked.
```suggestion
#!/bin/bash
```

certbot.wrapper
```diff
@@ -0,0 +1,16 @@
...
+
```
bmw, line 2: nit: I think we should probably make sure we exit if something goes wrong.
```suggestion
set -e
```

certbot.wrapper
```diff
@@ -0,0 +1,16 @@
...
+    if [ "$first" != "" ]; then
+        shift 2
+        echo -n "${first}"
+        for item in "$@"; do echo -n "${sep}${item}"; done
+        echo
+    fi
+}
+
+paths=$(for plugin_snap in $(snap connections certbot|sed -n '2,$p'|awk '$1=="content[certbot-1]"{print $3}'|cut -d: -f1); do echo /snap/$plugin_snap/current/lib/python3.6/site-packages; done)
+export PYTHONPATH=$PYTHONPATH:$(join : $paths)
```
bmw, line 15: I'm not sure if this matters, but if `PYTHONPATH` is initially empty, its value after this line will start with a colon. Can we either make sure this doesn't matter or fix the value in this case? This may not apply here, but [sys.path docs](https://docs.python.org/3/library/sys.html#sys.path) for example talk about how an empty value means to search in the current working directory for Python modules to import.

ohemorange: Not sure if it would matter either, but happy to make this change.

PR #8063, opened 2020-06-09
Stop running snapcraft with sudo

bmw:
It looks like `snapcraft` 4.0 has been pushed to the stable channel which broke our current set up. Test failures can be seen at https://travis-ci.com/github/certbot/certbot/builds/170443208.

This PR fixes the problem and you can see tests passing with this change at https://travis-ci.com/github/certbot/certbot/builds/170565797.

@adferrand or @ohemorange, if you're able to provide a quick review on this, I'd appreciate it!


bmw:
I also updated the proof-of-concept instructions to work around this new behavior from snapcraft.

bmw:
Thanks for the quick review!

Issue #8037, opened 2020-06-01
Delete or update references to Debian 8 Jessie


bmw:
Debian 8 Jessie reaches its end-of-life on June 30th. We should update or delete any references to it we find in our repo after that date.

PR #8065, opened 2020-06-10
Delete or update references to Debian 8 Jessie

kden:
Related to issue #8037


tools/oldest_constraints.txt
```diff
@@ -26,15 +26,6 @@ zope.component==4.1.0
 zope.event==4.0.3
 zope.interface==4.0.5
 
-# Debian Jessie Backports constraints
-PyICU==1.8
```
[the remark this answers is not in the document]
bmw: The full test suite initially failed here (you can see the log [here](https://gist.github.com/71c2b318f29ffc65cb47c7f54727110e)), but it was in our unpinned tests and was a spurious failure.

I agree with Erica's comment though. We need to do something else here.

For what it's worth, I personally think we can either:

1. Leave those dependencies pinned to those versions, but update the comment referencing Debian Jessie to say that since Debian Jessie has reached it's end-of-life, we can probably update these packages to a newer version if needed.
2. Take the time now to figure out the new oldest version of those packages we need to support is and update the pinned version in this file accordingly. The way to check this would be to look at the versions of these packages found in Debian 9, CentOS 7 with EPEL 7, and Ubuntu 16.04 and update the pinning to the oldest version found and move it to the correct section of this file to document where that version constraint is coming from.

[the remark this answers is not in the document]
ohemorange: I think going with option 1 is fine.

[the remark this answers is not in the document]
ohemorange: But if we do that, the comment should also make sure to include the places to check mentioned in option 2, because we will want to do that instead of just deleting things willy-nilly in the hypothetical future.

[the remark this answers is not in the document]
kden: I wasn't sure I understood initially but I checked with Brad and I think I made the desired change.  I'm not as up on the context as you guys are, so please let me know if it needs something more.

tools/oldest_constraints.txt
```diff
@@ -26,6 +26,9 @@ zope.event==4.0.3
 zope.interface==4.0.5
 
 # Debian Jessie Backports constraints
+# Debian Jessie has reached end of life. However:
+# When updating these dependencies, you should only update them to the oldest version of the package found
+# in a non-EOL'd version of CentOS, Debian, or Ubuntu that has Certbot packages in their OS repositories.
```
ohemorange, line 31:
```suggestion
# in a non-EOL'd version of CentOS, Debian, or Ubuntu that Certbot is supported on.
```

tools/oldest_constraints.txt
```diff
@@ -26,6 +26,9 @@ zope.event==4.0.3
 zope.interface==4.0.5
 
 # Debian Jessie Backports constraints
+# Debian Jessie has reached end of life. However:
+# When updating these dependencies, you should only update them to the oldest version of the package found
```
ohemorange, line 30:
```suggestion
# When it becomes necessary to upgrade any of these dependencies, you should only update them to the oldest version of the package found
```

tools/oldest_constraints.txt
bmw: Erica I think whatever you want here is fine, but FYI:

1) The exact wording for this entire sentence came from me.
2) While maybe overly verbose, it really is only distros that have Certbot OS packages. This file only contains Python packages and for other distribution methods such as certbot-auto, Docker, or snaps, we download our own versions from PyPI.

ohemorange: I didn't realize about that latter bit; resolving this conversation.

tools/oldest_constraints.txt
bmw: This change looks good to me!

PR #8078, opened 2020-06-15
Upgrade snap to be based on core20

### LINKED ISSUE #7854: Update Certbot snap to be based on Ubuntu 20.04
[Here](https://github.com/certbot/certbot/blob/5536c912238d96b15295a1b6938472fd24cf40ce/snap/snapcraft.yaml#L18) we define the [base image](https://snapcraft.io/docs/base-snaps) of the Certbot snap to be `core18` which is based on Ubuntu 18.04.

Probably the single most problematic part of packaging Certbot as a snap is going to be when we have to change the base snap in the future which may require all externally packaged plugins to have to upgrade in order to continue to work. More details about this can be found at https://docs.google.com/document/d/1OB5WCDz2mBMgRvtnvo4FSjWHPBDlt1NpccJ5sNrhWIw/edit?usp=sharing.

To delay these problems for as long as possible, I think we should be using the `core20` image based on Ubuntu 20.04 before we publish a stable Certbot snap. I think we may want to consider doing this before adding support for external plugins (see https://github.com/certbot/certbot/issues/7667), but I think that'll depend on the state of the `core20` snap and Ubuntu 20.04. Apparently, it already exists, but I'm not sure how stable it is yet. See
```
$ snap info core20
name:      core20
summary:   Runtime environment based on Ubuntu 20.04
publisher: Canonical✓
license:   unset
description: |
  The base snap based on the Ubuntu 20.04 release.
type:    base
snap-id: DLqre5XGLbDqg9jPtiAhRRjDuPVa5X1q
channels:
  stable:    20 2020-02-10 (347) 62MB -
  candidate: 20 2020-02-10 (347) 62MB -
  beta:      20 2020-03-16 (486) 62MB -
  edge:      20 2020-03-17 (496) 62MB -
```
We should probably look into the state of things with `core20` more when we go to work on this issue.

ohemorange:
This PR makes several changes to be built on top of the core20 base snap. Fixes #7854.

The main changes are to `snapcraft.yaml`. With Snapcraft 4.0/core20 base, the python plugin is a thin wrapper, basically creating a `venv` and installing the packages from the source. The trouble with this is that it runs pycache, creating caches that conflict from the different parts. So to solve that, we put everything in a single part. Other changes include:

- We use classic confinement, so we need to specify a bunch of python packages to `stage-packages`, as mentioned [here](https://forum.snapcraft.io/t/trouble-bundling-python-with-classic-confinement-in-core20-4-0-4/18234/2).
- The certbot executable now lives in `bin`, so specify running `certbot/bin`.
- Since `python-augeas` is now being pulled into the single part, remove the pinning from constraints so we can use the latest version directly from github.
- Precompile our `cryptography` and `cffi` wheels to be based on python3.8.

Separately, we had to upgrade the snapcraft docker image to be based on focal, due to the thin wrapper situation. This was accomplished [here](https://github.com/adferrand/snapcraft/pull/1).

Tests are passing here:
https://travis-ci.com/github/certbot/certbot/builds/171942590


ohemorange:
Once this is merged, I'll update https://github.com/certbot/certbot/blob/snap-plugin/README.md#use-testing-and-development

ohemorange:
Some other checks I want to run, as I think of them:
- [ ] make sure our dns-plugin can connect

ohemorange:
Closing in favor of https://github.com/certbot/certbot/pull/8086.

Issue #7512, opened 2019-11-06
Email address sent to EFF before certificate is successfully created


mfb:
Certbot sends the user's email address to EFF as soon as the user first runs certbot (if --eff-email option is on, or user opts in at the prompt), but before a valid certificate has been created.  I'm concerned that this behavior might be contributing to a higher than expected bounce rate for the subscription confirmation emails EFF sends out.

How hard would it be for certbot to only send the email address to EFF after a valid certificate has been created successfully?  This should cut down on the proportion of bogus email addresses - in some sense acting like a sortof "captcha".

bmw:
It'd be a bit of work honestly. The main difficulty is the user may not have provided their email address on the same run they issue their certificate. We'd have to store the need to offer to register their email with EFF somewhere on disk, check that after issuing certificates, and then delete it after the first certificate is issued. Obviously not impossible, but more work than just moving a function call later in execution.

mfb:
Yes I was imagining that the preference could be stored in the json files (or a new json file if need be)?

I wasn't sure on some of the details for how this would need to work - like is each instance of certbot tied to one email address, or could you create different certs with different email addresses on the same host, and send each of those email addresses to EFF?

bmw:
> Yes I was imagining that the preference could be stored in the json files (or a new json file if need be)?

Yep! We'd just have to shuffle some code around and then write the code to write and read it. We'd also have to decide if we want to ask about sending your email address to EFF when you first give us your email address like we currently do or only before we'd want to send it after you obtain your certificate.

> I wasn't sure on some of the details for how this would need to work - like is each instance of certbot tied to one email address, or could you create different certs with different email addresses on the same host, and send each of those email addresses to EFF?

It's the latter. One of the first steps in the ACME protocol used by Let's Encrypt in Certbot is for the user to create an account. This account can contain 0, 1, or even multiple email addresses. Certbot only allows you to have one account per server at a time, but you can always delete your account and create a new one or change the email addresses associated with your account.

mfb:
> We'd also have to decide if we want to ask about sending your email address to EFF when you first give us your email address like we currently do or only before we'd want to send it after you obtain your certificate.

Either way could work, but I guess it would be preferable to ask users to opt-in at the time of cert generation, since this would be closer in time to when the data is actually sent to EFF (and they receive the subscription confirmation email).

bmw:
[...]

PR #8079, opened 2020-06-16
Print a message about obtaining a certificate.

bmw:
While #8076 does everything asked for in https://github.com/certbot/certbot/issues/7512 and functionally works great, the placement of this prompt feels strange to me in the context of Certbot's other output. It looks like:
```
$ sudo certbot -d example.org --standalone certonly
Plugins selected: Authenticator standalone, Installer None
Obtaining a new certificate
http-01 challenge for example.org
Waiting for verification...
Cleaning up challenges

Would you be willing to share your email address with the Electronic Frontier
Foundation, a founding partner of the Let's Encrypt project and the non-profit
organization that develops Certbot? We'd like to send you email about our work
encrypting the web, EFF news, campaigns, and ways to support digital freedom.
(Y)es/(N)o:
```
This PR adds a new message in `_report_new_cert` which makes the flow a little nicer in my opinion. The new output looks like:
```
$ sudo certbot -d example.org --standalone certonly
Plugins selected: Authenticator standalone, Installer None
Obtaining a new certificate
http-01 challenge for example.org
Waiting for verification...
Cleaning up challenges
You have successfully obtained a certificate!

Would you be willing to share your email address with the Electronic Frontier
Foundation, a founding partner of the Let's Encrypt project and the non-profit
organization that develops Certbot? We'd like to send you email about our work
encrypting the web, EFF news, campaigns, and ways to support digital freedom.
(Y)es/(N)o:
```

### DIFF
--- a/certbot/certbot/_internal/main.py
+++ b/certbot/certbot/_internal/main.py
@@ -448,6 +448,8 @@ def _report_new_cert(config, cert_path, fullchain_path, key_path=None):
 
     assert cert_path and fullchain_path, "No certificates saved to report."
 
+    logger.info('You have successfully obtained a certificate!')
+
     expiry = crypto_util.notAfter(cert_path).date()
     reporter_util = zope.component.getUtility(interfaces.IReporter)
     # Print the path to fullchain.pem because that's what modern webservers


schoen:
I'm concerned with both this and #8076 that users who are already more familiar with Certbot might be confused about having this question move to the end, and may feel unprepared for it.

I think @adferrand has already thought and discussed about this a lot, but how about the possibility of still asking right at the beginning, but not acting on the result until after the certificate issuance?

The question could also be rephrased as

> Would you be willing, once your certificate is successfully issued, to share your email address with the Electronic Frontier Foundation, a founding partner of the Let's Encrypt project and the non-profit organization that develops Certbot? We'd like to send you email about our work encrypting the web, EFF news, campaigns, and ways to support digital freedom.

or

> After your certificate has been issued, would you be willing to share your email address with the Electronic Frontier Foundation, a founding partner of the Let's Encrypt project and the non-profit organization that develops Certbot? We'd like to send you email about our work encrypting the web, EFF news, campaigns, and ways to support digital freedom.

The local account metadata could still be used to actually *do* this action after the certificate issuance, but not move the question to that stage of the process.

(I think this is a complicated question. On the one hand, users will certainly feel more grateful to EFF right at the moment that the certificate has been issued, but on the other hand, they might be confused about whether the question is necessary or relevant at that point, since they may already be mentally moving on to the next step in their server configuration.)

bmw:
We decided to move the question to be during registration as you suggested Seth.

Because of that, I don't think this PR is needed anymore.

PR #8083, opened 2020-06-16
Drop i386 architecture on snap build

adferrand:
This PR remove the i386 architecture in the snap build process, because the base snap `core20` is not available for this architecture.


adferrand:
Just to check, I triggered and stopped immediately (to not waste Travis resources) a full test suite, to check that snaps jobs are created correctly.

Issue #8024, opened 2020-05-26
certbot.ocsp changes for Apache OCSP prefetching


bmw:
For https://github.com/certbot/certbot/pull/7649, we need a way to query OCSP and if the OCSP status is good, get the raw bytes of the OCSP response as well as the nextUpdate timestamp.

Unfortunately, this is a little tricky with the current functions we have in `certbot.ocsp`. The current functions in `certbot.ocsp` only return a `bool` whose value is only `True` if the OCSP status has been validated as revoked. If it has a different status or an error occurs, the value is `False`. We need to fully validate things like the OCSP signature, timestamp, etc. for "good" OCSP responses as well as get more information out of these functions.

To do this, I'd like to propose API changes like those seen at https://github.com/certbot/certbot/compare/ocsp-response-interface.

The reason I'd like to add all these interfaces is to:

1. Minimize passing around raw `bytes`.
2. Allow us to provide an OpenSSL (or other) implementation of this interface in the future without making API changes. (The hard parts of an OpenSSL implementation already exist in #7649. I just want to leave it out for now because I think we may not really need an OpenSSL implementation and to get the initial version of that feature out the door.)

To reuse code, I'm planning on essentially fixing up the current `cryptography` code path to implement the new function and then just checking the `certificate_status` attribute to return `True` or `False` for `ocsp_revoked*`.

Please treat that code as a little pseudocody, but if someone can give any feedback on the high level proposal before I dive deep into an implementation here, I'd really appreciate it!

adferrand:
I think that an OOM approach makes a lot of sense when we are dealing with APIs that needs to be stable. The overall complexity does not seem very high, so I have nothing against, in particular considering that the other parts of OCSP logic in certbot are moving in the same direction, and `acme` module globally uses the same approach with `josepy` classes.

bmw:
@joohoi, thanks for offering to take a look at this. Since you're most familiar with #7649, I think your feedback here is most valuable. To provide some more context for you specifically, the problem here is I don't think this comment is true:
https://github.com/certbot/certbot/blob/ae76f2d3b5baed6ed8c95b47a9f443310215d15c/certbot-apache/certbot_apache/_internal/prefetch_ocsp.py#L156-L157

`ocsp_revoked_by_paths` will also return `False` if there was a network failure, the OCSP response is expired, it's not properly signed, etc.

[...]

PR #8085, opened 2020-06-17
OCSP response interface

bmw:
Fixes https://github.com/certbot/certbot/issues/8024.

I tested the Apache code using the changes from this branch and it works as expected.

@adferrand or @joohoi, do you have the time/interest to review this?


bmw:
Welp. In case you hadn't heard, the Trump administration fired the head of OTF which was funding this work and the organization is at a standstill. See https://www.cnn.com/2020/06/17/media/us-agency-for-global-media-michael-pack/index.html.

Until that is resolved or we have some more time on our hands, I don't think we should worry about making more progress on this or making changes to Certbot's API to support it so I'm closing this PR for now.

PR #8089, opened 2020-06-18
Merge the snap-plugin branch into master

### LINKED ISSUE #8073: Merge the snap-plugin branch into master
- [x] We should probably do a bit of squashing to clean up commits and then merge and not squash the PR to preserve history.

- [x] The current README on that repo should probably be moved under `snap/local` or `certbot-dns-dnsimple/snap/local`.

To make this task easier, we can probably continue to use `snapcraft --use-lxd` since it's hard for me to come up with a concrete benefit to us switching the instructions to use the approach used in CI right now.

- [x] We will need to do things like update`tools/strip_hashes.py letsencrypt-auto-source/pieces/dependency-requirements.txt > constraints.txt` to `tools/strip_hashes.py letsencrypt-auto-source/pieces/dependency-requirements.txt > snap-constraints.txt` and generally make sure the instructions still work on `master` though.

- [x] The changes to `disco.py` should be discarded.

Doing this is valuable because it reduces the number of branches we're maintaining and automatically keeps the snap-plugin branch in sync with master.

ohemorange:
Fixes #8073.

Merge this PR to preserve history.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,38 +53,46 @@ You can now operate the plugin as normal.
 To try this out, you'll need to build the snaps (a patched Certbot snap and a
 plugin snap) manually.
 
- 1. Start with a Xenial VM.
- 2. Install snapcraft with `snap install --classic snapcraft`.
- 3. Run `git clone git://github.com/basak/certbot-snap-build -b snap-plugins/snap/certbot`.
- 4. `cd certbot-snap-build`
- 5. Run `git clone https://github.com/basak/certbot-snap-build -b snap-plugins/certbot certbot` (this is a workaround for #13).
- 6. Run `certbot/tools/strip_hashes.py certbot/letsencrypt-auto-source/pieces/dependency-requirements.txt > certbot/constraints.txt` (this is a workaround for #13).
- 7. Run `snapcraft`.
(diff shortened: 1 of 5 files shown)


certbot/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Certbot adheres to [Semantic Versioning](https://semver.org/).
 * Make Certbot snap find externally snapped plugins
 * Function `certbot.compat.filesystem.umask` is a drop-in replacement for `os.umask`
   implementing umask for both UNIX and Windows systems.
+* Merge the snap-plugin branch into master. This includes a snapcraft.yaml for
```
bmw, line 14: nit: Since this isn't a change that affects any of our distribution methods, I personally don't think we should have a changelog entry about this. I think our changelog is mostly for our users and packagers and this isn't useful information for them.

We can keep it if you disagree for some reason though.

ohemorange:
I will make the post-merge changes.

certbot/CHANGELOG.md
ohemorange: sgtm

ohemorange: this may disappear when I force-push, but: done.

ohemorange:
Actually, it might be helpful to keep around the `snap-plugin` branch for a bit; there are links littered here and there to commits that we might want. It's not super necessary since the things they point to will be in `master`, but might just make things easier at least until we get the first major round of snap support done.

PR #8094, opened 2020-06-18
Do not unstage generated cffi augeas

bmw:
This line seems to refer to [augeas.py](https://github.com/hercules-team/python-augeas/blob/v0.5.0/augeas.py) from the version of Augeas we normally have pinned. This was necessary when we were installing each Certbot component in separate parts and only combining them later to ensure that the Augeas fork (which uses cffi) was used instead of the unmodified pinned version of Augeas.

Since everything is installed in one part and we're removing the Augeas pinning now though, this line is no longer necessary. You can see the snap being built and tested successfully with this change at https://travis-ci.com/github/certbot/certbot/builds/172134518.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -49,8 +49,6 @@ parts:
       - ./certbot-nginx
     stage:
       - -usr/lib/python3.8/sitecustomize.py # maybe unnecessary
-      # Prefer cffi
-      - -lib/python3.8/site-packages/augeas.py
     stage-packages:
       - libaugeas0
       # added to stage python:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -49,6 +49,13 @@ parts:
       - ./certbot-nginx
     stage:
       - -usr/lib/python3.8/sitecustomize.py # maybe unnecessary
+      # Old versions of this file used to unstage
+      # lib/python3.8/site-packages/augeas.py to avoid conflicts between
+      # python-augeas 0.5.0 which was pinned in snap-constraints.txt and
+      # Robie's python-augeas fork which creates an auto-generated cffi file at
+      # the same path. Since we've combined things in one part and removed the
+      # python-augeas pinning, unstaging this file had a different, unintended
+      # effect so we now stage the file to keep the auto-generated cffi file.
     stage-packages:
       - libaugeas0
       # added to stage python:


ohemorange:
I agree this line isn't necessary, but this file isn't nonexistent. When I built and installed the snap, it did in fact install this file, which is not the file you link to. You can see the file it stages attached here (I changed the extension to upload it).
[augeas.txt](https://github.com/certbot/certbot/files/4801683/augeas.txt)

ohemorange:
The version we pull is quite different from the file linked above, actually. https://github.com/basak/python-augeas

bmw:
Oh nice catch! The other day I was very confused where this file is coming from since we don't have an `augeas.py` file ourselves and neither does https://github.com/basak/python-augeas/tree/snap, so I had assumed it must be the `augeas.py` file I linked above.

This exclusion originally comes from a snapcraft maintainer when they forked the Python plugin for us at https://github.com/basak/certbot-snap-build/pull/21.

I'll try building the snap at that state next week and if the file being removed there is the augeas.py file I linked above, I think we should merge this PR. If it's this autogenerated cffi file, I think we shouldn't merge this PR without learning more.

bmw:
Yeah this was originally referring to the `augeas.py` file I linked above.

On a machine set up to build snaps using LXD, if you run:
```
git clone https://github.com/basak/certbot-snap-build.git .
git checkout 4c52664161f52bd5b4da33c0edefbfb3e74c986d
git clone https://github.com/certbot/certbot.git certbot
certbot/tools/strip_hashes.py certbot/letsencrypt-auto-source/pieces/dependency-requirements.txt > certbot/constraints.txt
```
Then delete the lines deleted in this PR and run `snapcraft --use-lxd --debug`, `snapcraft` will error out with:
```
Failed to stage: Parts 'python-augeas' and 'certbot-apache' have the following files, but with different contents:
    lib/python3.6/site-packages/augeas.py
```
If you inspect things with the shell you now have inside the snap, you'll see `/root/parts/python-augeas/install/lib/python3.6/site-packages/augeas.py` is the autogenerated cffi file you linked [here](https://github.com/certbot/certbot/pull/8094#issuecomment-646355615) while `/root/parts/certbot-apache/install/lib/python3.6/site-packages/augeas.py` is the `augeas.py` file from our pinned version of python-augeas.

Since we no long have python-augeas pinned and everything is in one part, I think we're now unstaging this autogenerated cffi file. Things still seem to work, but this wasn't the intention of this line and I'm assuming python-augeas/cffi is creating it for some reason so I think we should continue to stage it like we were before.

ohemorange:
Thanks for running through this investigation, that analysis looks solid to me. Do we want to put a more coherent note in there summing this up or linking to that comment or something?

bmw:
I pushed a commit adding a comment about this.

Issue #7862, opened 2020-03-25
Document how to make external Certbot plugin snaps


bmw:
Once we have made our own external Certbot plugins, we should document how we did it so others can follow suit. I think this issue should be blocked on us doing this work and solving any problems along the way. The issue for making our own external plugin snaps is https://github.com/certbot/certbot/issues/7672.

PR #8106, opened 2020-06-23
Make DNS plugin snaps use core20

### LINKED ISSUE #8103: Make DNS plugin snaps use core20
Pulled out of https://github.com/certbot/certbot/issues/7672, as it looks like that one will be focused on generating the files themselves.

> As part of this, I think we should use core20 as the base for our DNS plugins. Using core18 seems to work for the dnsimple plugin, but I think we'll hit problems with this approach eventually and I think we'll also want to use core20/Python 3.8 everywhere eventually so I figure we might as well do it that way now.

ohemorange:
Fixes #8103.

This PR also modifies `certbot.wrapper` to look at the new path. This begs the question, how will we change the base for our plugins the next time we want to do it? Both snaps will need to be updated before the plugins can be found again. Are we ok with that?

### DIFF
--- a/certbot-dns-cloudflare/snap/snapcraft.yaml
+++ b/certbot-dns-cloudflare/snap/snapcraft.yaml
@@ -3,7 +3,7 @@ summary: Cloudflare DNS Authenticator plugin for Certbot
 description: Cloudflare DNS Authenticator plugin for Certbot
 confinement: strict
 grade: devel
-base: core18
+base: core20
 adopt-info: certbot-dns-cloudflare
 
 parts:
@@ -11,7 +11,6 @@ parts:
     plugin: python
     source: .
     constraints: [$SNAPCRAFT_PART_SRC/snap-constraints.txt]
-    python-version: python3
     override-pull: |
         snapcraftctl pull
         snapcraftctl set-version `grep ^version $SNAPCRAFT_PART_SRC/setup.py | cut -f2 -d= | tr -d "'[:space:]"`
@@ -23,4 +22,4 @@ slots:
     interface: content
     content: certbot-1
     read:
-      - $SNAP/lib/python3.6/site-packages
+      - $SNAP/lib/python3.8/site-packages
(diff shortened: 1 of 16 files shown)


ohemorange:
I tested this manually with `certbot-dns-dnsimple`.

bmw:
> how will we change the base for our plugins the next time we want to do it? Both snaps will need to be updated before the plugins can be found again. Are we ok with that?

I'm ok with that, but I'm curious what you think.

When we change the base and the version of Python used, I don't think we should expect things to still work until both snaps are updated. They might as we saw with our own transition to core20, but I don't think we should rely on that. Dependencies can differ on the Python version and the ABI will change so any compiled code will probably break. We could probably fix up `certbot.wrapper` so it dynamically uses the path provided by the plugin snap, but it doesn't avoid this problem.

Instead, when we want to change the base in the future, I think we need to do it in a coordinated way and we may want to change the name of the interface to `certbot-2` so the errors people will get is about plugins not being found instead of Certbot potentially crashing with a more obscure message at a random point in execution.

This all isn't pretty and I think dealing with changes like this is the ugliest part of the current snap plan, but I don't have any ideas I like better. I think we should be able to deal with this pretty rarely too. `core20` uses [Python 3.8 which will not be EOL'd until October 2024](https://devguide.python.org/#status-of-python-branches). `core20` and Ubuntu 20.04 in general will be supported long after that.

ohemorange:
I agree with your comment about doing things like adding a new interface and making sure the correct error message goes through.

adferrand:
Indeed we should not overdesign for a problem that will occur only each two years, or even less frequently. Changing the interface is a good to have a clear breakage. A migration plan could also be setup to give people time to adapt their dns plugibs before the actual breakage.

May I suggest a more informative interface name, like `certbot-core20` or `certbot-py38`?

bmw:
Keeping in mind I think we should [write some documentation about this](https://github.com/certbot/certbot/issues/7862) either way, do you prefer `certbot-core20` over just `certbot` with a clear, increasing version number? I personally don't have a strong opinion about which I'd like better.

If you do though, I'd personally be fine with `certbot-core20`. I don't like `certbot-py38` because I think at least in theory, the external snap cannot use any version of Python 3.8 since Python doesn't guarantee [ABI compatability between versions](https://docs.python.org/3/c-api/stable.html).

